PR #92, opened 2019-04-18
Clean code

hzxuzhonghu:
Continue removing some redundant code

### DIFF
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -17,12 +17,9 @@ limitations under the License.
 package job
 
 import (
-	"fmt"
-
 	"github.com/golang/glog"
 
 	"k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -38,8 +35,6 @@ import (
 	kbinfo "github.com/kubernetes-sigs/kube-batch/pkg/client/informers/externalversions/scheduling/v1alpha1"
 	kblister "github.com/kubernetes-sigs/kube-batch/pkg/client/listers/scheduling/v1alpha1"
 
-	v1corev1 "volcano.sh/volcano/pkg/apis/bus/v1alpha1"
-	"volcano.sh/volcano/pkg/apis/helpers"
 	vkver "volcano.sh/volcano/pkg/client/clientset/versioned"
 	vkscheme "volcano.sh/volcano/pkg/client/clientset/versioned/scheme"
 	vkinfoext "volcano.sh/volcano/pkg/client/informers/externalversions"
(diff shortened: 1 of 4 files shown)


pkg/controllers/job/job_controller.go
```diff
@@ -127,53 +122,18 @@ func NewJobController(config *rest.Config) *Controller {
...
-			default:
-				runtime.HandleError(fmt.Errorf("unable to handle object %T", obj))
-				return false
-			}
-		},
-		Handler: cache.ResourceEventHandlerFuncs{
-			AddFunc: cc.addCommand,
-		},
+	cc.cmdInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: cc.addCommand,
```
TommyLike: Can we keep the ``ControllerdBy`` filter logic here?

pkg/controllers/job/job_controller.go
```diff
@@ -127,53 +122,18 @@ func NewJobController(config *rest.Config) *Controller {
...
-				return false
-			}
-		},
-		Handler: cache.ResourceEventHandlerFuncs{
-			AddFunc:    cc.addPod,
-			UpdateFunc: cc.updatePod,
-			DeleteFunc: cc.deletePod,
-		},
+	podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    cc.addPod,
```
TommyLike: Ditto

pkg/controllers/job/job_controller_actions.go
```diff
@@ -66,38 +66,20 @@ func (cc *Controller) killJob(jobInfo *apis.JobInfo, nextState state.NextStateFn
...
 
-			switch pod.Status.Phase {
-			case v1.PodRunning:
-				err := cc.deleteJobPod(job.Name, pod)
-				if err != nil {
-					running++
-					errs = append(errs, err)
-					continue
-				}
+			if err := cc.deleteJobPod(job.Name, pod); err != nil {
```
TommyLike: err == nil

pkg/controllers/job/job_controller.go
hzxuzhonghu: Is there any command resource not managed by volcano?

pkg/controllers/job/job_controller.go
hzxuzhonghu: This is not needed,  we already have many other checks in event handlers, like this

```
	taskName, found := pod.Annotations[vkbatchv1.TaskSpecKey]
	if !found {
		glog.Infof("Failed to find taskName of Pod <%s/%s>, skipping",
			pod.Namespace, pod.Name)
		return
	}

	jobName, found := pod.Annotations[vkbatchv1.JobNameKey]
	if !found {
		glog.Infof("Failed to find jobName of Pod <%s/%s>, skipping",
			pod.Namespace, pod.Name)
		return
	}

	version, found := pod.Annotations[vkbatchv1.JobVersion]
	if !found {
		glog.Infof("Failed to find jobVersion of Pod <%s/%s>, skipping",
			pod.Namespace, pod.Name)
		return
	}
```

pkg/controllers/job/job_controller_actions.go
hzxuzhonghu: good catch

pkg/controllers/job/job_controller.go
TommyLike: which is faster? Just asking.

pkg/controllers/job/job_controller.go
TommyLike: Which is faster, just asking.

pkg/controllers/job/job_controller.go
TommyLike: No for now.

pkg/controllers/job/job_controller.go
hzxuzhonghu: The above check is inevitable, so it's enough, if we check twice, then absolutely double time wasting

PR #94, opened 2019-04-18
Boilerplate update

hzxuzhonghu:
Remove `YEAR`, keep the ci valid in a new year.


TommyLike:
Which is failing?

hzxuzhonghu:
same things have happened in k8s, like when time goes to 2020.1.1, the verify will fail.

TommyLike:
They should fix it:)

k82cn:
let keep "YEAR" there.

Issue #93, opened 2019-04-18
Allow multi sync job works run in parallel


hzxuzhonghu:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind feature

**What happened**:

Currently, there is only one goroutine worker syncing jobs. For large scale jobs, this will be a bottle neck.

TommyLike:
Do you mind to re-upload your patch again 😄

hzxuzhonghu:
Sure, will try to find the commit

PR #96, opened 2019-04-19
Support syncing jobs concurrently

hzxuzhonghu:
fixes: #93


pkg/controllers/job/job_controller_actions.go
```diff
@@ -106,14 +107,10 @@ func (cc *Controller) killJob(jobInfo *apis.JobInfo, nextState state.NextStateFn
 	}
 
 	// Update Job status
-	if job, err := cc.vkClients.BatchV1alpha1().Jobs(job.Namespace).UpdateStatus(job); err != nil {
+	if err := cc.updateJobStatus(job); err != nil {
 		glog.Errorf("Failed to update status of Job %v/%v: %v",
 			job.Namespace, job.Name, err)
 		return err
-	} else {
-		if e := cc.cache.Update(job); e != nil {
```
hzxuzhonghu: We donot need update at all, now every time we `processNextReq`, get from the informer lister and do a deepcopy to prevent concurrent status modification lead to unexpected result.

cmd/controllers/app/options/options.go
```diff
@@ -28,11 +28,18 @@ type ServerOption struct {
 	Kubeconfig           string
 	EnableLeaderElection bool
 	LockObjectNamespace  string
+	// ConcurrentJobSyncs is the number of job syncing operations
+	// that will be done concurrently. Larger number = faster job updating,
+	// but more CPU (and network) load.
+	ConcurrentJobSyncs int32
```
TommyLike: worker threads or something similar sounds more clear.

hzxuzhonghu: SGTM

pkg/controllers/job/job_controller.go
```diff
@@ -203,6 +206,15 @@ func (cc *Controller) processNextReq() bool {
 		return true
 	}
 
+	job, err := cc.jobLister.Jobs(req.Namespace).Get(req.JobName)
```
k82cn: there's maybe multiple requests for a job, which will make it have different workers handling the same thing.

hzxuzhonghu: Yes, it may occur, for example when multi-pods event arrives.There is a risk for  causing job.status outdated.  

The safe way is to add a lock not to handle same job concurrently.

PR #98, opened 2019-04-22
Add TF example

TommyLike:
See above


TommyLike:
@asifdxtreme Related image ``volcanosh/example-tf`` has been uploaded to docker hub already, please have a try.

asifdxtreme:
@TommyLike Thanks..I will check it now. Do we need to take the latest volcano or 0.1.0 version of volcano should be fine?

TommyLike:
@asifdxtreme  version is not strict limited, the example only needs the basic volcano feature and the `env` plugins.

asifdxtreme:
I tried this example and it works fine

PR #100, opened 2019-04-23
Add design doc for CronJob

TommyLike:
See above.


docs/design/cronjob.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Admission Controller
+
+The admission controller will ensure the correctness of `JobSpec`, especially the `Schedule` expression.
+
+### Feature Interaction
+
+#### cli
+
+Command line support for `CronJob` is not considered at this first stage, However end user still can manipulate
```
k82cn: we need to design it here, and finish it in different release.

docs/design/cronjob.md
```diff
@@ -0,0 +1,118 @@
...
+    // This is a pointer to distinguish between explicit zero and not specified.
+    // Defaults to 3.
+    // +optional
+    SuccessfulJobsHistoryLimit *int32 `json:"successfulJobsHistoryLimit,omitempty" protobuf:"varint,6,opt,name=successfulJobsHistoryLimit"`
+
+    // The number of failed finished jobs to retain.
+    // This is a pointer to distinguish between explicit zero and not specified.
+    // Defaults to 1.
+    // +optional
+    FailedJobsHistoryLimit *int32 `json:"failedJobsHistoryLimit,omitempty" protobuf:"varint,7,opt,name=failedJobsHistoryLimit"`
```
k82cn: Did not check code in upstream, but how do we trace this history?

docs/design/cronjob.md
```diff
@@ -0,0 +1,118 @@
...
+	// Current status of CronJob
+	// +optional
+	Status CronJobStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+type CronJobSpec struct {
+	// The schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+    Schedule string `json:"schedule" protobuf:"bytes,1,opt,name=schedule"`
+	// Template is a template based on which the Job should run.
+
```
k82cn: please format the code.

docs/design/cronjob.md
```diff
@@ -0,0 +1,118 @@
...
+	// Information when was the last time the job was successfully scheduled.
+	// +optional
+	LastScheduleTime *metav1.Time `json:"lastScheduleTime,omitempty" protobuf:"bytes,4,opt,name=lastScheduleTime"`
+}
+
+```
+
+### CronJob Controller
+
+The `CronJobController` will run periodically, watching all `CronJob` resource as well as their batch job resources.
```
k82cn: what's the different between this implementation and spark-operator/tf-operator?

docs/design/cronjob.md
TommyLike: ok.

docs/design/cronjob.md
TommyLike: By browsing all the jobs in the cluster and deleting some of the jobs according to this attribute.

docs/design/cronjob.md
TommyLike: There is no huge difference between them, the thing I can think of is kubernetes will perform the sync logic periodically.

k82cn: I'd suggest to reference spark/tf-operator scheduled job here.

TommyLike: What do you mean here, replace the whole design/CronJob object within the design that in spark-operator(tf-operator doesn't have a Cron TFjob resource) or just the implementation detail? I personally prefer the later if this is you are asking.

k82cn: if I want to use vk for spark scheduled-application, what should I do?

TommyLike: @k82cn AFAIK, when considering using vk for spark scheduled application, there is no big difference between these two approaches, since key point only lies at how we define the `JobSpec` or `JobTemplate` inside of CronJob, that is to say how can we define a volcano job to execute spark application,  these two approaches we mentioned only has difference on how to schedule the cron jobs.

PR #107, opened 2019-04-28
Move the whole kube-batch codes into volcano

TommyLike:
1. Move the whole kube-batch codes into volcano
2. Remove kubernete-sigs/kube-batch from dep.


TommyLike:
@k82cn  Please review when you free.

TommyLike:
@k82cn  what's the difference between the two PR checks?

Issue #122, opened 2019-05-05
Sync combine input & output volumes into array logic.


TommyLike:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

During main logic sync, the change of combine input & output feature is partially synced due to the code merge conflict, need fix and reenable it.

See https://github.com/volcano-sh/volcano/blob/d9b532c184783d8e718a537880e00ac2eea1ab16/pkg/admission/admit_job.go#L147

TommyLike:
Done in #124

PR #124, opened 2019-05-05
Combine input & output volumes

TommyLike:
Sync feature of combine input & output volumes.

For #122


pkg/apis/batch/v1alpha1/job.go
```diff
@@ -81,8 +78,11 @@ type VolumeSpec struct {
 	// not contain ':'.
 	MountPath string `json:"mountPath" protobuf:"bytes,1,opt,name=mountPath"`
 
+	// defined the PVC name
+	VolumeClaimName string `json:"volumeClaimName,omitempty" protobuf:"bytes,2,opt,name=volumeClaimName"`
```
hzxuzhonghu: Why would u add this? Would prefer auto gen it.

TommyLike: Yeah, there would be a few codes which can be optimized, would like to do it when the initial sync patch get merged.

Issue #121, opened 2019-05-05
Fix state machine issue


TommyLike:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

New state of enqueue has been introduced, but it's unfinished, need keep working on this and fix related testcase issues.

**NOTES**: There are some testcases are expected to have job status: pending->running/xxxxx, which are incorrect within new status of enqueue, please update them all asl well.

TommyLike:
Partial fix for #120

PR #126, opened 2019-05-05
[Issue #121]fix state convert

wangyuqing4:


TommyLike:
@wangyuqing4  It's very rare to see  all of the testcases failing, please ensure this change works correctly for CREATE/RESTART/SUSPEND/RESUME actions, and if that so, I would like to work on updating these testcases to reflect the latest state machine of Job Resource.

cc @k82cn

Issue #16, opened 2019-03-16
Queue controller and related cli


k82cn:
Currently, user can only create a Queue for scheduling; but it's hard to know more info about it, e.g. how many job in the queue, which plugins is used by this queue; and if the Queue is deleted, the job is still there :( It's better to have QueueController to mamange Queue's lifecycle and update its status; and have related command line for uset to get its info.

hzxuzhonghu:
@k82cn I would make queueController in the same process with job controller, is this ok?

k82cn:
> I would make queueController in the same process with job controller, is this ok?

That's ok to me; similar to kube-controller-manager :) BTW, we have a cache in `pkg/controller`; please reuse it if necessary :)

PR #128, opened 2019-05-05
Add queue controller

hzxuzhonghu:
fixes: #16


pkg/apis/helpers/helpers.go
```diff
@@ -63,7 +63,7 @@ func ControlledBy(obj interface{}, gvk schema.GroupVersionKind) bool {
 	return false
 }
 
-func CreateConfigMapIfNotExist(job *vkv1.Job, kubeClients *kubernetes.Clientset, data map[string]string, cmName string) error {
+func CreateConfigMapIfNotExist(job *vkv1.Job, kubeClients kubernetes.Interface, data map[string]string, cmName string) error {
```
k82cn: prefer to have seperate PR for such kind of cleanup.

pkg/controllers/queue/queue_controller.go
```diff
@@ -0,0 +1,266 @@
...
+	glog.V(5).Infoln("Begin sync queue")
+
+	var pending, running, unknown int32
+	c.pgMutex.RLock()
+	podGroups := make([]string, len(c.podGroups[key]))
+	if c.podGroups[key] == nil {
+		glog.V(2).Infoln("queue %s has not been seen or deleted")
+		return nil
+	}
+	for pgKey := range c.podGroups[key] {
```
k82cn: prefer to handle this in Cache: Queue -> Job -> PodGroup.

hzxuzhonghu: I donot want to overlap with job controller.  For queue controller, it just care about the podgroups, not care about Job or anything else. 

podGroups is keyed by queue name and value is the podgroups that are using this queue.

pkg/apis/helpers/helpers.go
hzxuzhonghu: make sense

pkg/controllers/queue/queue_controller.go
k82cn: `pkg/cache` is a common component in controller, and we need to clean up jobs when queue was deleted.

hzxuzhonghu: IC, then we need to record jobs too.

hzxuzhonghu: But have a doubt: why would user delete their queues before jobs? Is there a use case? If there is a  malicious user who does this, we should not delete the running jobs by any reason.

k82cn: "malicious user" is handled by security; if admin want to delete queue, it's a heavy work to delete all jobs before delete queue.

hzxuzhonghu: For this, I can think of a more simple way to do this: I remember there will be an admission controller to validate queue existence before create jobs. Then we can also set the job's owner reference.  Then when the queue is deleted, k8s garbage collector can reclaim all children resources automatically. 

So we have two choice:

1. do it in controller, need to record all dependents relationship.

1. setting ownerreference, do nothing else, leave it to GC.

cmd/controllers/app/server.go
```diff
@@ -72,10 +76,17 @@ func Run(opt *options.ServerOption) error {
 		return err
 	}
 
-	jobController := job.NewJobController(config)
+	// TODO: add user agent for different controllers
```
TommyLike: What is user agent?

pkg/controllers/queue/queue_controller.go
TommyLike: Personally prefer not adding this into cache even we need considering deleting jobs when queue deleted.

cmd/controllers/app/server.go
hzxuzhonghu: It is just for http server to know the client identity.

hzxuzhonghu: A http header

pkg/controllers/queue/queue_controller.go
k82cn: > setting ownerreference, do nothing else, leave it to GC.

owerReference should be reserved for `JobGroup` :)

hzxuzhonghu: OwnerReferences is an array, can be more than one.  

>owerReference should be reserved for JobGroup :)

Sorry, there is no `JobGroup` now.

k82cn: not sure whether GC will check controlled owner reference.

> there is no JobGroup now.

so "reserved" it.

pkg/controllers/queue/queue_controller.go
```diff
@@ -0,0 +1,277 @@
...
+
+	return nil
+}
+
+func (c *Controller) addQueue(obj interface{}) {
+	queue := obj.(*kbv1alpha1.Queue)
+	c.queue.Add(queue.Name)
+}
+
+func (c *Controller) updateQueue(old, new interface{}) {
```
TommyLike: Why adding this while leave it empty?

hzxuzhonghu: will remove

hzxuzhonghu:
I am adding e2e

TommyLike:
@hzxuzhonghu please update the [cluster role](https://github.com/volcano-sh/volcano/blob/9b7e9d7479cca4c170267bd7298666a66ab3b31a/installer/chart/volcano/templates/controllers.yaml#L8) to get appropriate rights to handle queue resource.

hzxuzhonghu:
Thanks

pkg/controllers/queue/queue_controller.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func (c *Controller) syncQueue(key string) error {
+	glog.V(5).Infoln("Begin sync queue")
+
+	var pending, running, unknown int32
+	c.pgMutex.RLock()
+	podGroups := make([]string, len(c.podGroups[key]))
+	if c.podGroups[key] == nil {
+		glog.V(2).Infoln("queue %s has not been seen or deleted")
```
TommyLike: Incorrect of info message:
```
glog.V(2).Infoln("queue %s has not been seen or deleted")
```

PR #132, opened 2019-05-06
Fix log tailed issue

TommyLike:
When used with label filter option, ``Kubectl logs`` would tail logs in 10 lines in default, this patch increases it to 10000


TommyLike:
@k82cn please take a look

hzxuzhonghu:
any reason not the whole logs?

TommyLike:
@hzxuzhonghu since it's only used in CI environment, it would be great to limit the whole size if the **limit** can cover most of the cases. But as you said, it's worth discussing.

hzxuzhonghu:
It's better if these log files can be downloaded.

TommyLike:
@hzxuzhonghu  can be uploaded to external storage, for instance S3.

Issue #129, opened 2019-05-05
Propose not to delete all pods after the job finished


hzxuzhonghu:
xref: https://github.com/kubernetes-sigs/kube-batch/issues/797

Even job fails, the state machine would call `KillJob` to delete all owned resources. I think it is not acceptable for users, they need to look why it fails.

k82cn:
Only delete `Running/Pending` Pod should be ok :)

PR #149, opened 2019-05-09
Cleanup Job after a configured ttl

hzxuzhonghu:


TommyLike:
@hzxuzhonghu great! Is it for #129 ?

hzxuzhonghu:
Should be merged after #128

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -70,6 +70,15 @@ type JobSpec struct {
...
 	MaxRetry int32 `json:"maxRetry,omitempty" protobuf:"bytes,8,opt,name=maxRetry"`
+
+	// ttlSecondsAfterFinished limits the lifetime of a Job that has finished
+	// execution (either Completed or Failed). If this field is set,
+	// ttlSecondsAfterFinished after the Job finishes, it is eligible to be
+	// automatically deleted. If this field is unset,
+	// the Job won't be automatically deleted. If this field is set to zero,
+	// the Job becomes eligible to be deleted immediately after it finishes.
+	// +optional
+	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty" protobuf:"varint,8,opt,name=ttlSecondsAfterFinished"`
```
hzxuzhonghu: Should we set a default value for this ?  Current behavior is when no set, gc will not delete Job.

TommyLike: index -> 9

pkg/controllers/garbagecollector/garbagecollector.go
```diff
@@ -0,0 +1,283 @@
...
+		return err
+	} else if !expired {
+		return nil
+	}
+
+	// The Job's TTL is assumed to have expired, but the Job TTL might be stale.
+	// Before deleting the Job, do a final sanity check.
+	// If TTL is modified before we do this check, we cannot be sure if the TTL truly expires.
+	// The latest Job may have a different UID, but it's fine because the checks will be run again.
+	fresh, err := gb.vkClient.BatchV1alpha1().Jobs(namespace).Get(name, metav1.GetOptions{})
```
TommyLike: In which case, we need to do a double check ?

pkg/controllers/job/state/aborting.go
```diff
@@ -31,17 +33,18 @@ func (ps *abortingState) Execute(action vkv1.Action) error {
 		// Already in Restarting phase, just sync it
 		return KillJob(ps.job, func(status *vkv1.JobStatus) {
 			status.State.Phase = vkv1.Restarting
+			status.State.LastTransitionTime = metav1.Now()
```
TommyLike: There a lots of codes:
```
status.State.LastTransitionTime = metav1.Now()
```
which can be upgraded into a function which used to update status and update the `LastTransitionTime`  when required.

pkg/apis/batch/v1alpha1/job.go
k82cn: > Should we set a default value for this ?  Current behavior is when no set, gc will not delete Job.

Yes, we should set default for this in admission controller.

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -218,6 +227,10 @@ type JobState struct {
 	// +optional
 	Phase JobPhase `json:"phase,omitempty" protobuf:"bytes,1,opt,name=phase"`
 
+	// Last time the condition transit from one phase to another.
+	// +optional
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,4,opt,name=lastTransitionTime"`
```
k82cn: ordered filed by protobuf id

pkg/controllers/garbagecollector/garbagecollector.go
hzxuzhonghu: We can not guarantee the job TTL not changed between gc received the it and processing here.

hzxuzhonghu: Mostly these codes are referred from k8s ttlafterfinished controller

pkg/apis/batch/v1alpha1/job.go
hzxuzhonghu: > index -> 9

Will update kube-batch

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -70,6 +70,15 @@ type JobSpec struct {
...
 	MaxRetry int32 `json:"maxRetry,omitempty" protobuf:"bytes,8,opt,name=maxRetry"`
+
+	// ttlSecondsAfterFinished limits the lifetime of a Job that has finished
+	// execution (either Completed or Failed). If this field is set,
+	// ttlSecondsAfterFinished after the Job finishes, it is eligible to be
+	// automatically deleted. If this field is unset,
+	// the Job won't be automatically deleted. If this field is set to zero,
+	// the Job becomes eligible to be deleted immediately after it finishes.
+	// +optional
+	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty" protobuf:"varint,9,opt,name=ttlSecondsAfterFinished"`
```
SrinivasChilveri: I think we should add the validation as it should not be negative

hzxuzhonghu: validate can be applied in admission controller

SrinivasChilveri: > validate can be applied in admission controller

yeah thats correct we should add the validation in admission controller

hzxuzhonghu: IIRC @wangyuqing4 will work on admission controller.

hzxuzhonghu:
@k82cn This is ready to go?

pkg/apis/batch/v1alpha1/job.go
k82cn: @hzxuzhonghu , please confirm with  @wangyuqing4 ; if she is not working this, please feel free to open a PR for that.

PR #152, opened 2019-05-10
admission to get tls certificate from kubeconfig, if tls config not defined in command line

sivanzcw:
Admission to obtain the cluster tls authentication certificate, firstly try to get from the tls config defined by command line, if not, try to obtain from the cluster's kubeconfig.


TommyLike:
@sivanzcw  Have you tried whether the admission service can work with the default kubeconfig certificate?

sivanzcw:
@TommyLike Certificate of default kubeconfig like admin.conf does not contain the credit for 'volcano-admission-service.default.svc' service of admission webhook. If default admin.conf kubeconfig is used, the admission can access the apiserver normally, but the apiserver callback webhook service will be authentication failed, because the service address was not trusted. The TLS certificate in the kubeconfig file that is attached to the admission needs to be a certificate that has been trusted for the service of admission webhook. If certificate in volcano-admission-secret is configured to kubeconfig, the admission can work normally.

PR #155, opened 2019-05-10
Fix running job state setting

hzxuzhonghu:


hzxuzhonghu:
may be this is not necessary,

PR #156, opened 2019-05-10
Leaves kill job to GC

hzxuzhonghu:
This allows users to investigate the task result.


pkg/controllers/job/state/finished.go
```diff
@@ -26,6 +26,7 @@ type finishedState struct {
 }
 
 func (ps *finishedState) Execute(action vkv1.Action) error {
-	// In finished state, e.g. Completed, always kill the whole job.
-	return KillJob(ps.job, nil)
+	// In finished state, e.g. Completed,Terminated, Failed
```
SrinivasChilveri: does it mean we will be updating the admission webhook  as per ttl controller. otherwise GC will not comeinto picture naaa?? am i missing something else?


Use a mutating admission webhook to set this field dynamically at resource creation time. Cluster administrators can use this to enforce a TTL policy for finished resources.
Use a mutating admission webhook to set this field dynamically after the resource has finished, and choose different TTL values based on resource status, labels, etc.

hzxuzhonghu:
ping @k82cn

PR #162, opened 2019-05-14
Fix calculation issue for extension resource in volcano.

TommyLike:
Use original ResourceList sum up logic to avoid resource amplify issue.  
```
minAvailableTasksRes.Add(kbapi.NewResource(c.Resources.Requests))
```
The code above should not be used outside of kube-batch, since when converting back, the resource would be amplified with 1000 times. see [here](https://github.com/volcano-sh/kube-batch/blob/0f36bc162ea2a351c675b816bc21da0b49723161/pkg/scheduler/api/resource_info.go#L87)


k82cn:
/hold

It's better to make `Convert2K8sResource` and `NewResource` align with each other. BTW, please rename `Convert2K8sResource` to `ToResourceList`.

TommyLike:
@k82cn  I agree, but for this your suggestion, it's the matter of kube-batch's, we don't need to convert it into kube-batch type and then convert it back here.

TommyLike:
Raised the issue in community [here](https://github.com/kubernetes-sigs/kube-batch/issues/851)

k82cn:
we should not duplicate the code; we maintain both repo.

TommyLike:
> we should not duplicate the code; we maintain both repo.

Which are duplicated?

k82cn:
please help to rename Convert2K8sResource to ToResourceList and make sure them align.

TommyLike:
> please help to rename Convert2K8sResource to ToResourceList and make sure them align.

They are different logics, one for the kube batch resource and the other for kubernetes resource.

k82cn:
`Convert2K8sResource` is also a bug to me, please update kube-batch to make them align.

TommyLike:
ha, removed unused codes in order to eliminate issue, see [pr](https://github.com/volcano-sh/kube-batch/pull/12)

PR #164, opened 2019-05-14
Add GPU based tensorflow job example

TommyLike:
See above.


example/integrations/tensorflow/Dockerfile.cpu
```diff
@@ -9,5 +9,5 @@ RUN  apt-get update --fix-missing \
 && apt-get install -y git \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/*
-RUN pip install tf-nightly-gpu \
+RUN pip install tf-nightly \
```
k82cn: if we're not using this dockerfile  in the yaml file, please move this to different director.

TommyLike: hmm, this is a dockerfile that can actually work with the latest master branch.

k82cn: if `volcanosh/example-tf:0.0.1` is not built from this dockerfile, please move it out and give a new name to `volcanosh/example-tf:0.0.1`.

TommyLike: New images now:
1. example-tf-cpu: 0.0.1
2. example-tf-gpu: 0.0.1

k82cn: As discussed before, this dockerfile is an example; and `example-tf:xxx` is retag from tensorflow-benchmark. So we need to make this message clear to others.

TommyLike: `example-tf:xxx` is only used by OpenLab now, and those can be removed in the future, we don't use `example-tf:xxx` image

k82cn: hm.. I'm confused about current status :( please make sure dockerfile & image in yaml align, all the time.

TommyLike: @k82cn , ha let me summarize the history.

1. We used to have `example-tf.yaml` and `dockerfile`, in which `example-tf:0.0.1` image is used. But now they are all have been removed.
2. `dockerfile.cpu` is used to generate` example-tf-cpu.0.0.1` image, which is been used in `tf-example-cpu.yaml`
3. there is another `dockerfile.gpu` is used to generate` example-tf-gpu.0.0.1` image, and which is been used in `tf-example-gpu.yaml`

4. openlab generated another image example-tf:0.0.3 which is used for compare test with kubeflow.

k82cn: > openlab generated another image example-tf:0.0.3 which is used for compare test with kubeflow.

do not use the same name of image; that's a really long history ...

TommyLike: > > openlab generated another image example-tf:0.0.3 which is used for compare test with kubeflow.
> 
> do not use the same name of image; that's a really long history ...

Will be renamed when this cycle of tests finished.

example/integrations/tensorflow/Dockerfile.gpu
```diff
@@ -0,0 +1,12 @@
+# NOTE: the build process would change during developing.
+
+FROM nvidia/cuda:10.0-runtime-ubuntu16.04
+MAINTAINER volcano <maintainer@volcano.sh>
+RUN  apt-get update --fix-missing \
+&& apt-get install -y git python2.7 python-pip libcudnn7=7.4.1.5-1+cuda10.0 libcudnn7-dev=7.4.1.5-1+cuda10.0 \
+&& apt-get clean \
+&& rm -rf /var/lib/apt/lists/*
+RUN pip install tf-nightly-gpu \
```
lminzhw: `tf-nightly-gpu` in pip and master branch of tensorflow/benchmarks may change rapidly.

I think we can use a special version/branch of tensorflow and benchmark to make the image more reliable and more reproducible.

TommyLike: personally I prefer to use the package without specifying version unless the latest version has some issues/bugs. And for the master issue, I think we need to add a timestamp to the image tag.

lminzhw: I think so, we should support at lease one of the timestamp or version/branch.

PR #166, opened 2019-05-15
Repropose support cronjob

TommyLike:
Propose to support cronjob in volcano. Spec as below.

### DIFF
--- a/docs/design/cronjob.md
+++ b/docs/design/cronjob.md
@@ -1,0 +1,147 @@
+# CronJob
+
+[@TommyLike](https://github.com/tommyLike/); April 23, 2019
+
+## Motivation
+
+`CronJob` is designed to help end user to schedule their batch job at some specific time, such as if you want to
+schedule a batch job at a low activity period, or just need the job to be executed periodically, it's especially
+useful for generating data for test or research. 
+
+## Function Specification
+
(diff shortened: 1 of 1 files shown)


docs/design/cronjob.md
```diff
@@ -0,0 +1,147 @@
+# CronJob
+
+[@TommyLike](https://github.com/tommyLike/); April 23, 2019
+
+## Motivation
+
+`CronJob` is designed to help end user to schedule their batch job at some specific time, such as if you want to
+schedule a batch job at a low activity period, or just need the job to be executed periodically, it's especially
```
hzxuzhonghu: I can see the value of delay scheduling/running, but can hardly imagine when  periodically run needed. Can you elaborate a little?

TommyLike: It's a common case for high performance tasks which usually task a long time, for example, daily processing business data, generate report, update warehouse.

xieydd: CronJob is valueable, add the flexibility of scheduler.

k82cn: yes, we need this feature :)

TommyLike:
@k82cn what's the next step of this proposal?

k82cn:
it's low priority for now, let's handle this in next release.

TommyLike:
@k82cn  close and open when next release.

PR #167, opened 2019-05-15
Added ttl validation in admission controller

SrinivasChilveri:
ttl validation in admission controller.

travis result

VolcanoTravis_ 2019-05-15 11-45-14

### DIFF
--- a/pkg/admission/admit_job.go
+++ b/pkg/admission/admit_job.go
@@ -86,6 +86,11 @@ func validateJob(job v1alpha1.Job, reviewResponse *v1beta1.AdmissionResponse) st
 		return fmt.Sprintf("'maxRetry' cannot be less than zero.")
 	}
 
+	if job.Spec.TTLSecondsAfterFinished != nil && *job.Spec.TTLSecondsAfterFinished < 0 {
+		reviewResponse.Allowed = false
+		return fmt.Sprintf("'ttlSecondsAfterFinished' cannot be less than zero.")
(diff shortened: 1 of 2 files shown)


SrinivasChilveri:
@hzxuzhonghu ,added validation in admission controller. pls do have a look.

hzxuzhonghu:
/lgtm

@SrinivasChilveri Could you add a default value for it when it is nil?

TommyLike:
> /lgtm
> 
> @SrinivasChilveri Could you add a default value for it when it is nil?

Would be conflicted to your design:
```
	// ttlSecondsAfterFinished limits the lifetime of a Job that has finished
	// execution (either Completed or Failed). If this field is set,
	// ttlSecondsAfterFinished after the Job finishes, it is eligible to be
	// automatically deleted. If this field is unset,
	// the Job won't be automatically deleted. If this field is set to zero,
	// the Job becomes eligible to be deleted immediately after it finishes.
	// +optional
```

/lgtm

SrinivasChilveri:
> /lgtm
> 
> @SrinivasChilveri Could you add a default value for it when it is nil?

can i add default value as 10 ?

hzxuzhonghu:
@TommyLike  Yeah， should update the desc.

>can i add default value as 10 ?

I think this should be large enough to allow users to check logs / results. I prefer 1 day,  @k82cn @TommyLike WDYT

Issue #168, opened 2019-05-15
Info: Do not just add a e2e test


hzxuzhonghu:
For small features or bug fixes, if it can be tested by a UT, do not add e2e again.

As one e2e case can cause 20+ seconds, it will make the CI takes more time.

TommyLike:
@hzxuzhonghu would you help on setting up the UT framework and moving some of the them?

hzxuzhonghu:
>@hzxuzhonghu would you help on setting up the UT framework and moving some of the them?

sure,  but i have been working on other stuff these days. May have bandwidth in the middle of next week.

thandayuthapani:
Have written UT cases for two files in Job package, ref https://github.com/volcano-sh/volcano/pull/201 & https://github.com/volcano-sh/volcano/pull/213
Files in Job package
  - [x]  job_controller_plugins
  - [x]  job_controller_handler
  - [ ]  job_controller_actions

thandayuthapani:
UT for other packages are to be added, listing out other packages to which UT cases to be added.
 - [ ] pkg/cli
 - [ ] pkg/controllers/queue
 - [ ] pkg/controllers/garbagecollector
 - [ ] pkg/controllers/apis
 - [ ] pkg/controllers/cache
 - [ ] pkg/controllers/job/state

k82cn:
@thandayuthapani , please also help to provide code coverage of each package :)

asifdxtreme:
UT has been added for all the packages and the current coverage for UT is around [61%](https://coveralls.io/github/volcano-sh/volcano?branch=master), In my opinion we can go ahead and close this issue

hzxuzhonghu:
Can we remove some e2e cases, which is already covered by e2e?

asifdxtreme:
> Can we remove some e2e cases, which is already covered by e2e?

We went through all E2E test case, In our opinion all are required as of now, the test scenarios for UT are restricted to a particular functions/methods whereas E2E is tesing the complete flow for each scenario, so the both have different scope while testing. Going ahead we need to add more E2E's to cover all scenario, I don't think so we can reduce the E2E.

TommyLike:
@asifdxtreme  I am wondering whether we need keeping the case of `MPI`, since all of the required features have been (or should) tested separately in different cases. the downloading and loading mpi images also consume some time.
```
• [SLOW TEST:16.254 seconds]
MPI E2E Test
[...]

Issue #84, opened 2019-04-09
Speed up E2E tests


TommyLike:
/kind bug

Currently, Travis would spend almost 26 minutes to finish e2e tests, need to figure it out how to speed up these tests.
```
Ran for 26 min 14 sec
```
```
Ran 33 of 33 Specs in 773.302 seconds
SUCCESS! -- 33 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2E (773.30s)
PASS
ok  	volcano.sh/volcano/test/e2e	773.323s
release "integration" deleted
Running kind: [kind delete cluster --name integration]
Deleting cluster "integration" ...
$KUBECONFIG is still set to use /home/travis/.kube/kind-config-integration even though that file has been deleted, remember to unset it
Volcano logs are currently not supported.
The command "make e2e-test-kind" exited with 0.
```

TommyLike:
Reopen when required

k82cn:
```
Ran 35 of 35 Specs in 849.903 seconds
SUCCESS! -- 35 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2E (849.90s)
```

xref https://travis-ci.com/volcano-sh/volcano/jobs/198124601

TommyLike:
@k82cn To be honest, I think we are testing some of the advanced features which **SHOULD** and **ALREADY** been tested in kube-batch, suggest removing some of them.

k82cn:
> which SHOULD and ALREADY been tested in kube-batch, suggest removing some of them.

kube-batch did not include vk.job :(

TommyLike:
@k82cn that's true, but for some cases such as preemption, statement, gang scheduling and etc. they are all testing the features of kube-batch. If our CI can ensure `vk.job` are correctly translated into kube-batch recognizable resource. I think that's enough. at lease we can skipping them for every PR's check, maybe valid them in cron task?  Here are some of the testcase time consumption:

```
• [SLOW TEST:39.837 seconds]
Job E2E Test
/root/go_projects/src/volcano.sh/volcano/test/e2e/job_scheduling.go:26
  Gang scheduling
  /root/go_projects/src/volcano.sh/volcano/test/e2e/job_scheduling.go:82
------------------------------
• [SLOW TEST:33.809 seconds]
Job E2E Test
/root/go_projects/src/volcano.sh/volcano/test/e2e/job_scheduling.go:26
  Preemption
[...]

PR #171, opened 2019-05-15
moved some of admission e2e test cases as ut test cases

SrinivasChilveri:
moved e2e admission tcs as ut test cases.
added ut trigger as part of CI
Part of  Fix Related to #84 
Part of  Fix Related to #168 

I hope similary we can move which ever e2e tcs are possible or add  UT tcs only for small fetures/issues if e2e is not required.

UT_PR_2019-05-15 18-40-02


pkg/admission/admit_job_test.go
```diff
@@ -0,0 +1,291 @@
+package admission
```
k82cn: please add header there.

TommyLike:
/lgtm
/approve

merge for move on.

Issue #170, opened 2019-05-15
Job Priority


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

Currently, kube-batch already support `PriorityClass`; it's better to also add `PriorityClass` to job in controller.

PR #174, opened 2019-05-16
Support job level priorityClassName

TommyLike:
Add `PriorityClassName` and add it into PD Spec when creating PodGroup. Also add some UT for mutate_job 

For #170


k82cn:
we should not modify Task's priority which is used by Task-Priority feature; we just need to set PodGroup's priorityClassName in job controller.

PR #176, opened 2019-05-16
retain pod with special phase

lminzhw:
xref: https://github.com/volcano-sh/volcano/issues/129

Now, pod would be delete when calling KillJob function.

I try to retain part of them based of pod phase.


TommyLike:
@lminzhw  Is there a overlap with PR #156

lminzhw:
> @lminzhw Is there a overlap with PR #156

No.

#156 change the status of job, but still delete the pod as before. This PR just want to limit the range of pod to be deleted.

Issue #151, opened 2019-05-10
Adding more status/events for measurement


TommyLike:
/kind feature

**What happened**:

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

k82cn:
please also add description for this issue, including which kind of event/status we'd like to update.

TommyLike:
@k82cn  sure

TommyLike:
**Description**:
Now the a lot of job details are missed which makes end user or developer hard to measure and debug, for instance. when running  our tf/mpi examples, users are hard to find out when job starts to scheduler/run/complete/fail. As a developer I also want to know more details about the volcano job such as whether & when plugins have been successfully executed. We need to add more traces on our job API resources.

TommyLike:
**Proposes**:
1. Add more event for tracing such as when job plugin execute, configmap/service/pods are created.
2. Update our job state attribute to condition which can provide more details.

```go
type JobCondition struct {
  // Type of job condition.
  Type JobConditionType `json:"type"`
  // Status of the condition, one of True, False, Unknown.
  Status v1.ConditionStatus `json:"status"`
  // The reason for the condition's last transition.
  Reason string `json:"reason,omitempty"`
  // A human readable message indicating details about the transition.
  Message string `json:"message,omitempty"`
  // The last time this condition was updated.
  LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
  // Last time the condition transitioned from one status to another.
  LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
}
```


@k82cn  & @hzxuzhonghu

hzxuzhonghu:
Sounds good

k82cn:
LGTM :)

PR #179, opened 2019-05-16
Support job condition

TommyLike:
Support condition in job status

#151


TommyLike:
Within this change, `JobState` has been removed, and only `JobPhase` is reserved more likely for our state machine.  At the meantime a new attribute `JobCondition` which mainly consists of some constants states is been added. 

@k82cn  @hzxuzhonghu  Please take a review and leave your concerns.

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -225,63 +225,83 @@ const (
...
 }
 
 // JobStatus represents the current status of a Job
 type JobStatus struct {
 	// Current state of Job.
-	State JobState `json:"state,omitempty" protobuf:"bytes,1,opt,name=state"`
+	Phase JobPhase `json:"phase,omitempty" protobuf:"bytes,1,opt,name=phase"`
+
+	//Jon conditions
+	Conditions []JobCondition `json:"conditions,omitempty" protobuf:"bytes,2,opt,name=conditions"`
```
hzxuzhonghu: Donot change others' index

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -225,63 +225,83 @@ const (
...
-	// Last time the condition transit from one phase to another.
-	// +optional
-	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,4,opt,name=lastTransitionTime"`
+// +k8s:deepcopy-gen=true
+// JobCondition describes the state of the job at a certain point.
+type JobCondition struct {
+	// Type of job condition.
+	Type JobConditionType `json:"type"`
+	// Status of the condition, one of True, False, Unknown.
+	Status v1.ConditionStatus `json:"status"`
```
hzxuzhonghu: Is this necessary?

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -225,63 +225,83 @@ const (
...
-	Message string `json:"message,omitempty" protobuf:"bytes,3,opt,name=message"`
+const (
+	// JobCreated means the job has been accepted by the system, but it hasn't
+	// reached the phase of running, this includes creating the job resource,
+	// waiting for pod resource and scheduling.
+	JobCreated JobConditionType = "JobCreated"
+	// JobRunning means at least 'MinAvailable' pods of this job are running.
+	JobRunning JobConditionType = "JobRunning"
+	// JobSucceed means the job has been successfully executed.
+	JobSucceed JobConditionType = "JobSucceed"
```
hzxuzhonghu: `JobSucceed` redundant Job prefix

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -225,63 +225,83 @@ const (
...
+	// JobTerminated means the job has been suspended
+	JobAborted JobConditionType = "JobAborted"
+)
 
-	// Last time the condition transit from one phase to another.
-	// +optional
-	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,4,opt,name=lastTransitionTime"`
+// +k8s:deepcopy-gen=true
+// JobCondition describes the state of the job at a certain point.
+type JobCondition struct {
```
hzxuzhonghu: This will make status more complicated, and if the condition like `JobSucceed` bacomes true, you need to set all other conditions false, which is not good

condition xxx, status: true/false
condition yyy, status: true/false
condition zzz, status: true/false

pkg/apis/batch/v1alpha1/job.go
TommyLike: Ok

pkg/apis/batch/v1alpha1/job.go
TommyLike: Sure

pkg/apis/batch/v1alpha1/job.go
TommyLike: We still have status/phase in JobStatus, JobConditions used to show some major changes during job lifecycle. Yeah, When jobSucceed the JobRunning condition would be false. But JobSucceed/Failed/Termiated/Aborted will not appear at the same time.

pkg/apis/batch/v1alpha1/job.go
TommyLike: I think so

TommyLike:
I paste the Pod Phase and Pod Condition here to see if we can figure out the detail we all agreed.

```
// PodPhase is a label for the condition of a pod at the current time.
type PodPhase string

// These are the valid statuses of pods.
const (
	// PodPending means the pod has been accepted by the system, but one or more of the containers
	// has not been started. This includes time before being bound to a node, as well as time spent
	// pulling images onto the host.
	PodPending PodPhase = "Pending"
	// PodRunning means the pod has been bound to a node and all of the containers have been started.
	// At least one container is still running or is in the process of being restarted.
	PodRunning PodPhase = "Running"
	// PodSucceeded means that all containers in the pod have voluntarily terminated
	// with a container exit code of 0, and the system is not going to restart any of these containers.
	PodSucceeded PodPhase = "Succeeded"
	// PodFailed means that all containers in the pod have terminated, and at least one container has
	// terminated in a failure (exited with a non-zero exit code or was stopped by the system).
	PodFailed PodPhase = "Failed"
	// PodUnknown means that for some reason the state of the pod could not be obtained, typically due
	// to an error in communicating with the host of the pod.
	PodUnknown PodPhase = "Unknown"
)

type PodConditionType string

// These are valid conditions of pod.
const (
	// PodScheduled represents status of the scheduling process for this pod.
	PodScheduled PodConditionType = "PodScheduled"
	// PodReady means the pod is able to service requests and should be added to the
	// load balancing pools of all matching services.
	PodReady PodConditionType = "Ready"
	// PodInitialized means that all init containers in the pod have started successfully.
	PodInitialized PodConditionType = "Initialized"
	// PodReasonUnschedulable reason in PodScheduled PodCondition means that the scheduler
	// can't schedule the pod right now, for example due to insufficient resources in the cluster.
	PodReasonUnschedulable = "Unschedulable"
	// ContainersReady indicates whether all containers in the pod are ready.
	ContainersReady PodConditionType = "ContainersReady"
)
```

pkg/apis/batch/v1alpha1/job.go
hzxuzhonghu: I checked the pod condition, this make sense

TommyLike:
Within this change the Status attribute would be like:
```
Status:
  Succeeded:        1
  Completion Time:  2019-05-17T07:32:01Z
  Conditions:
    Last Transition Time:  2019-05-17T07:31:57Z
    Last Update Time:      2019-05-17T07:31:57Z
    Reason:                Job created
    Status:                True
    Type:                  JobCreated
    Last Transition Time:  2019-05-17T07:32:01Z
    Last Update Time:      2019-05-17T07:32:01Z
    Reason:                Job successfully scheduled
    Status:                True
    Type:                  JobScheduled
    Last Transition Time:  2019-05-17T07:32:01Z
    Last Update Time:      2019-05-17T07:32:01Z
    Reason:                Job completed
    Status:                True
    Type:                  JobSucceed
  Min Available:           1
  Phase:                   Completed
  Start Time:              2019-05-17T07:31:57Z
  Version:                 1
```

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -225,29 +225,50 @@ const (
...
+	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
+	// Last time the condition transitioned from one status to another.
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
 }
 
 // JobStatus represents the current status of a Job
 type JobStatus struct {
 	// Current state of Job.
-	State JobState `json:"state,omitempty" protobuf:"bytes,1,opt,name=state"`
+	Phase JobPhase `json:"phase,omitempty" protobuf:"bytes,1,opt,name=phase"`
```
lminzhw: We lost the reason/message/lastTransitionTime field in `job.status.state` right? ... 

So, how does we prepare to maintain this information now?

TommyLike: I don't found any usage for these two attributes(reason/message). if we need some additional informations, we can leverage the Condition's message/message or events. and for lastTransitionTime, it has been upgraded into Condition's lastTransitionTime attribute.

lminzhw: seems good  :)

TommyLike:
@k82cn please take a review when you free

TommyLike:
@k82cn  @hzxuzhonghu  please take a look when free

hzxuzhonghu:
@TommyLike My main concern is if there is a compatibility issue for our users. @lminzhw Do you use this api as well?

TommyLike:
> @TommyLike My main concern is if there is a compatibility issue for our users. @lminzhw Do you use this api as well?

cc @lminzhw

lminzhw:
> > @TommyLike My main concern is if there is a compatibility issue for our users. @lminzhw Do you use this api as well?
> 
> cc @lminzhw

This is a break change. Users would be affected when they use `Phase` to judge this job is stopped or not.

So, by using SemVer, could we tolerate this break changes? or leave this changes to a large version? or just keep the `State` and add `Condition` parallelly.

TommyLike:
> > > @TommyLike My main concern is if there is a compatibility issue for our users. @lminzhw Do you use this api as well?
> > 
> > 
> > cc @lminzhw
> 
> This is a break change. Users would be affected when they use `Phase` to judge this job is stopped or not.
> 
> So, by using SemVer, could we tolerate this break changes? or leave this changes to a large version? or just keep the `State` and add `Condition` parallelly.

@lminzhw  Thanks for you reply, to be clear the `phase` is still there, but it has been updated into a single string. But yes, it's an incompatible change, I can keep both `state` and `condition` in my changes.

TommyLike:
@hzxuzhonghu   @lminzhw  Now the `JobState` has been kept while introduce the `JobCondition`

PR #181, opened 2019-05-17
support vkctl job run --limits

lminzhw:
Now, `vkctl job run` support argument `--requests` to set resource request for job to be created.

But, the resource limits could not set by `vkctl`. So, this PR support `--limits` to do this.

We don't need to validate limits with request, because the admission do this for us in `validateTaskTemplate` function from `pkg/admission/admit_job.go`.


lminzhw:
It would be used like this:

`vkctl job run --limits "cpu=1000m,memory=2Gi" --requests "cpu=1"`

PR #187, opened 2019-05-19
move scheduler name out of common args

lminzhw:
`--scheduler` is common argument now, so it can be parsed in any command ignoring that most of them do not use it at all.

So, this PR move this argument out of common argument and add it on demand for command which use it before.

And, add support for `vkctl job list` to list job with specified scheduler name.

### DIFF
--- a/pkg/cli/job/common.go
+++ b/pkg/cli/job/common.go
@@ -22,13 +22,11 @@ import (
 )
 
 type commonFlags struct {
-	Master        string
-	Kubeconfig    string
-	SchedulerName string
+	Master     string
+	Kubeconfig string
 }
 
 func initFlags(cmd *cobra.Command, cf *commonFlags) {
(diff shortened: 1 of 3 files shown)


TommyLike:
Do we have the case that a volcano job is not scheduled by `kube-batch`?

lminzhw:
> Do we have the case that a volcano job is not scheduled by `kube-batch`?

Following our best practice, all the job should use kube-batch.

But we have field `job.spec.schedulerName` to change the default behavior, so I think this logic should be supported all around.

TommyLike:
> > Do we have the case that a volcano job is not scheduled by `kube-batch`?
> 
> Following our best practice, all the job should use kube-batch.
> 
> But we have field `job.spec.schedulerName` to change the default behavior, so I think this logic should be supported all around.

Thanks
/lgtm
/approve

PR #190, opened 2019-05-20
sync job works concurrently

SrinivasChilveri:
sync jobs concurrently

Fix : #93

### DIFF
--- a/cmd/controllers/app/server.go
+++ b/cmd/controllers/app/server.go
@@ -85,12 +85,12 @@ func Run(opt *options.ServerOption) error {
 	kbClient := kbver.NewForConfigOrDie(config)
 	vkClient := vkclient.NewForConfigOrDie(config)
 
-	jobController := job.NewJobController(kubeClient, kbClient, vkClient)
+	jobController := job.NewJobController(kubeClient, kbClient, vkClient, opt.WorkerThreads)
 	queueController := queue.NewQueueController(kubeClient, kbClient)
 	garbageCollector := garbagecollector.New(vkClient)
 
 	run := func(ctx context.Context) {
-		go jobController.Run((int)(opt.WorkerThreads), ctx.Done())
+		go jobController.Run(opt.WorkerThreads, ctx.Done())
 		go queueController.Run(ctx.Done())
 		go garbageCollector.Run(ctx.Done())
 		<-ctx.Done()
--- a/cmd/controllers/app/server.go
+++ b/cmd/controllers/app/server.go
@@ -90,7 +90,7 @@ func Run(opt *options.ServerOption) error {
 	garbageCollector := garbagecollector.New(vkClient)
 
 	run := func(ctx context.Context) {
-		go jobController.Run(ctx.Done())
+		go jobController.Run((int)(opt.WorkerThreads), ctx.Done())
 		go queueController.Run(ctx.Done())
 		go garbageCollector.Run(ctx.Done())
 		<-ctx.Done()
--- a/cmd/controllers/app/server.go
+++ b/cmd/controllers/app/server.go
@@ -90,7 +90,7 @@ func Run(opt *options.ServerOption) error {
 	garbageCollector := garbagecollector.New(vkClient)
 
 	run := func(ctx context.Context) {
-		go jobController.Run(opt.WorkerThreads, ctx.Done())
+		go jobController.Run(ctx.Done())
 		go queueController.Run(ctx.Done())
 		go garbageCollector.Run(ctx.Done())
 		<-ctx.Done()
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 
 	"github.com/golang/glog"
+	"sync"
 	"time"
 
 	"k8s.io/api/core/v1"
@@ -105,6 +106,7 @@ type Controller struct {
 
 	sync.Mutex
 	errTasks workqueue.RateLimitingInterface
+
 }
 
 // NewJobController create new Job Controller
@@ -130,6 +132,7 @@ func NewJobController(
 		errTasks:        newRateLimitingQueue(),
 		recorder:        recorder,
 		priorityClasses: make(map[string]*v1beta1.PriorityClass),
+		
 	}
 
 	cc.jobInformer = vkinfoext.NewSharedInformerFactory(cc.vkClients, 0).Batch().V1alpha1().Jobs()
@@ -233,6 +236,25 @@ func (cc *Controller) processNextReq() bool {
 
 	req := obj.(apis.Request)
 	defer cc.queue.Done(req)
+
+	key := jobcache.JobKeyByReq(&req)
+
+	// prevent multi threads processing the same job simultaneously.
+	cc.jobsLock.Lock()
+	if cc.jobsMap[key] {
+		// the job is being processed by some other thread
+		cc.queue.AddRateLimited(req)
+		cc.jobsLock.Unlock()
+		return true
+	} else {
+		cc.jobsMap[key] = true
+		cc.jobsLock.Unlock()
+		defer func() {
+			cc.jobsLock.Lock()
+			delete(cc.jobsMap, key)
+			cc.jobsLock.Unlock()
+		}()
+	}
 
 	glog.V(3).Infof("Try to handle request <%v>", req)
 
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -18,12 +18,12 @@ package job
 
 import (
 	"fmt"
-	"sync"
-
-	"github.com/golang/glog"
 	"hash"
 	"hash/fnv"
+	"sync"
 	"time"
+
+	"github.com/golang/glog"
 
 	"k8s.io/api/core/v1"
 	"k8s.io/api/scheduling/v1beta1"
@@ -107,10 +107,7 @@ type Controller struct {
 
 	sync.Mutex
 	errTasks workqueue.RateLimitingInterface
-
-	// To protect the queue list by processing multiple workers
-	lock    sync.RWMutex
-	workers uint32
+	workers  uint32
 }
 
 // NewJobController create new Job Controller
@@ -119,7 +116,6 @@ func NewJobController(
 	kbClient kbver.Interface,
 	vkClient vkver.Interface,
 	workers uint32,
-
 ) *Controller {
 
 	//Initialize event client
@@ -205,7 +201,7 @@ func NewJobController(
 }
 
 // Run start JobController
-func (cc *Controller) Run(workers uint32, stopCh <-chan struct{}) {
+func (cc *Controller) Run(stopCh <-chan struct{}) {
 
 	go cc.sharedInformers.Start(stopCh)
 	go cc.jobInformer.Informer().Run(stopCh)
@@ -221,7 +217,7 @@ func (cc *Controller) Run(workers uint32, stopCh <-chan struct{}) {
 
 	go wait.Until(cc.handleCommands, 0, stopCh)
 	var i uint32
-	for i = 0; i < workers; i++ {
+	for i = 0; i < cc.workers; i++ {
 		go func(num uint32) {
 			wait.Until(
 				func() {
@@ -230,7 +226,6 @@ func (cc *Controller) Run(workers uint32, stopCh <-chan struct{}) {
 				time.Second,
 				stopCh)
 		}(i)
-
 	}
 
 	go cc.cache.Run(stopCh)
@@ -242,14 +237,12 @@ func (cc *Controller) Run(workers uint32, stopCh <-chan struct{}) {
 }
 
 func (cc *Controller) worker(i uint32) {
-
 	glog.Infof("worker %d start ...... ", i)
 
 	for cc.processNextReq(i) {
 	}
 }
 
-// TODO we may need to make this sharding more proper if required
 func (cc *Controller) belongsToThisRoutine(key string, count uint32) bool {
 	var hashVal hash.Hash32
 	var val uint32
@@ -266,8 +259,7 @@ func (cc *Controller) belongsToThisRoutine(key string, count uint32) bool {
 	return false
 }
 
-// TODO we may need to make this sharding more proper if required
-func (cc *Controller) getWorkerID(key string) workqueue.RateLimitingInterface {
+func (cc *Controller) getWorkerQueue(key string) workqueue.RateLimitingInterface {
 	var hashVal hash.Hash32
 	var val uint32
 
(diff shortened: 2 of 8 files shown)


SrinivasChilveri:
@hzxuzhonghu , pls do have a look.

pkg/controllers/job/job_controller.go
```diff
@@ -197,7 +198,10 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
 		cc.svcSynced, cc.cmdSynced, cc.pvcSynced, cc.pcSynced)
 
 	go wait.Until(cc.handleCommands, 0, stopCh)
-	go wait.Until(cc.worker, 0, stopCh)
+
+	for i := 0; i < workers; i++ {
+		go wait.Until(cc.worker, time.Second, stopCh)
```
lminzhw: Why we use `time.Second` here but not `0` like the old logic? ...

SrinivasChilveri: @lminzhw  followed k8s controller mgr controllers way.I hope if we have many controllers then its good to have some time instead of zero. may be we can have this parameter also as config parameter. what do you say?

lminzhw: I browse some code of controller in k8s, and realize that most of k8s controller use `time.Second` as interval and multi-worker to solve the large scale problem.

Seems good :)

hzxuzhonghu:
@SrinivasChilveri it is not that simple to concurrently run multi workers, the workers may conflict operating same Job i think.  Have to prevent this.

SrinivasChilveri:
> @SrinivasChilveri it is not that simple to concurrently run multi workers, the workers may conflict operating same Job i think. Have to prevent this.

yeah, as our controller can have multiple entries related to same job in queue. have pushed the commit. pls do  have a look

SrinivasChilveri:
@hzxuzhonghu ,added new commit for your comment. pls do have a look.

pkg/controllers/job/job_controller.go
```diff
@@ -219,6 +230,25 @@ func (cc *Controller) processNextReq() bool {
 	req := obj.(apis.Request)
 	defer cc.queue.Done(req)
 
+	key := jobcache.JobKeyByReq(&req)
+
+	// prevent multi threads processing the same job simultaneously.
+	cc.jobsLock.Lock()
+	if cc.jobsMap[key] {
+		// the job is being processed by some other thread
+		cc.queue.AddRateLimited(req)
```
k82cn: if conflict after a few times, it'll take a really time to retry job. It's better to shard the request to different chanel/worker.

hzxuzhonghu: Also this re-enqueue ops can lead to request processed in unexpected orders, though i  have no detailed scenario now.

SrinivasChilveri: @k82cn , @hzxuzhonghu , Based on my understading sharding with single queue will not solve this issue. so i have added queueList with number of workers & did the commit. please do review the same.

pkg/controllers/job/job_controller.go
```diff
@@ -197,27 +217,96 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
...
 
 	req := obj.(apis.Request)
-	defer cc.queue.Done(req)
+	defer cc.queueList[count].Done(req)
+
+	key := jobcache.JobKeyByReq(&req)
+	// Later we can remove this code if we want
+	if !cc.belongsToThisRoutine(key, count) {
+		glog.Errorf("should not occur The job does not belongs to this routine key:%s, worker:%d...... ", key, count)
+		cc.queueList[count].AddRateLimited(req)
```
k82cn: do not use AddRateLimited in this case; just hash request by job into different queue, and have several workers to handle them.

pkg/controllers/job/job_controller.go
```diff
@@ -197,27 +217,96 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
...
+
+	val = hashVal.Sum32()
+
+	return val % cc.workers
 }
 
-func (cc *Controller) processNextReq() bool {
-	obj, shutdown := cc.queue.Get()
+func (cc *Controller) processNextReq(count uint32) bool {
+	obj, shutdown := cc.queueList[count].Get()
```
hzxuzhonghu: nit: I would like to acquire the queue before this loop

pkg/controllers/job/job_controller.go
```diff
@@ -197,27 +217,96 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
...
 		glog.Errorf("Fail to pop item from queue")
 		return false
 	}
 
 	req := obj.(apis.Request)
-	defer cc.queue.Done(req)
+	defer cc.queueList[count].Done(req)
+
+	key := jobcache.JobKeyByReq(&req)
+	// Later we can remove this code if we want
```
hzxuzhonghu: why add this?

pkg/controllers/job/job_controller.go
```diff
@@ -197,27 +217,96 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
...
+	if !cc.belongsToThisRoutine(key, count) {
+		glog.Errorf("should not occur The job does not belongs to this routine key:%s, worker:%d...... ", key, count)
+		cc.queueList[count].AddRateLimited(req)
+		return true
+	}
+
+	// prevent multi threads processing the same job simultaneously.
+	cc.jobsLock.Lock()
```
hzxuzhonghu: This would not happen if you do sharding in the source.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -241,7 +247,9 @@ func (cc *Controller) updatePod(oldObj, newObj interface{}) {
 		JobVersion: int32(dVersion),
 	}
 
-	cc.queue.Add(req)
+	key := vkcache.JobKeyByReq(&req)
+	i := cc.getWorkerID(key)
```
hzxuzhonghu: I think this can guarantee worker threads will not process same jobs

hzxuzhonghu:
I'd like to see some tests on concurrent accessing the job cache

pkg/controllers/job/job_controller.go
SrinivasChilveri: Basically we are sharing the jobs and adding into different queues. In this function just reading from different queues ,so it never happens to hit success this condition because the while placing  the job request into different  queue based on shard. while reading also it  should belong to same shard / queue/worker.

pkg/controllers/job/job_controller.go
SrinivasChilveri: agreed, will  remove the code

pkg/controllers/job/job_controller_handler.go
SrinivasChilveri: yes

pkg/controllers/job/job_controller.go
SrinivasChilveri: changed

SrinivasChilveri:
> I'd like to see some tests on concurrent accessing the job cache

better todo in separate PR

hzxuzhonghu:
LGTM, leaves to @k82cn for last eye

SrinivasChilveri:
@k82cn  @hzxuzhonghu , pls do have a look and let me know, need to update any more points.

SrinivasChilveri:
@k82cn , the PR rebased.

SrinivasChilveri:
@k82cn  do i need to change anything else.

pkg/controllers/job/job_controller.go
```diff
@@ -215,20 +241,69 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
...
+
+	if val%cc.workers == count {
+		return true
 	}
+
+	return false
+}
+
+// TODO we may need to make this sharding more proper if required
+func (cc *Controller) getWorkerID(key string) workqueue.RateLimitingInterface {
```
TommyLike: This function doesn't work as its name describes

TommyLike:
cc @k82cn @hzxuzhonghu

cmd/controllers/app/server.go
```diff
@@ -85,12 +85,12 @@ func Run(opt *options.ServerOption) error {
...
 	vkClient := vkclient.NewForConfigOrDie(config)
 
-	jobController := job.NewJobController(kubeClient, kbClient, vkClient)
+	jobController := job.NewJobController(kubeClient, kbClient, vkClient, opt.WorkerThreads)
 	queueController := queue.NewQueueController(kubeClient, kbClient)
 	garbageCollector := garbagecollector.New(vkClient)
 
 	run := func(ctx context.Context) {
-		go jobController.Run(ctx.Done())
+		go jobController.Run(opt.WorkerThreads, ctx.Done())
```
k82cn: We already create jobController with `opt.WorkerThreads`, why `Run` still need this parameters?

pkg/controllers/job/job_controller.go
```diff
@@ -21,6 +21,9 @@ import (
 	"sync"
 
 	"github.com/golang/glog"
+	"hash"
```
k82cn: close to other system libs.

pkg/controllers/job/job_controller.go
```diff
@@ -104,13 +107,19 @@ type Controller struct {
...
+	workers uint32
 }
 
 // NewJobController create new Job Controller
 func NewJobController(
 	kubeClient kubernetes.Interface,
 	kbClient kbver.Interface,
 	vkClient vkver.Interface,
+	workers uint32,
+
```
k82cn: remove empty line

pkg/controllers/job/job_controller.go
```diff
@@ -215,20 +241,69 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
 	glog.Infof("JobController is running ...... ")
 }
 
-func (cc *Controller) worker() {
-	for cc.processNextReq() {
+func (cc *Controller) worker(i uint32) {
+
```
k82cn: remove empty line.

pkg/controllers/job/job_controller.go
```diff
@@ -205,7 +220,18 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
...
+	for i = 0; i < workers; i++ {
+		go func(num uint32) {
+			wait.Until(
+				func() {
+					cc.worker(num)
+				},
+				time.Second,
+				stopCh)
+		}(i)
+
```
k82cn: remove empty line.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -65,7 +65,9 @@ func (cc *Controller) addJob(obj interface{}) {
 		glog.Errorf("Failed to add job <%s/%s>: %v in cache",
 			job.Namespace, job.Name, err)
 	}
-	cc.queue.Add(req)
+	key := vkcache.JobKeyByReq(&req)
```
k82cn: introduces a helper function for that.

pkg/controllers/job/job_controller.go
k82cn: let's remove that

pkg/controllers/job/job_controller_plugins_test.go
```diff
@@ -35,7 +35,7 @@ func newFakeController() *Controller {
 	VolcanoClientSet := volcanoclient.NewSimpleClientset()
 	KubeClientSet := kubeclient.NewSimpleClientset()
 
-	controller := NewJobController(KubeClientSet, KubeBatchClientSet, VolcanoClientSet)
+	controller := NewJobController(KubeClientSet, KubeBatchClientSet, VolcanoClientSet, 1)
```
k82cn: add more workers so we may catch some race condition issue.

pkg/controllers/job/job_controller.go
SrinivasChilveri: changed the name

cmd/controllers/app/server.go
SrinivasChilveri: changed

pkg/controllers/job/job_controller.go
SrinivasChilveri: changed

pkg/controllers/job/job_controller.go
SrinivasChilveri: removed

pkg/controllers/job/job_controller.go
SrinivasChilveri: removed

pkg/controllers/job/job_controller.go
SrinivasChilveri: removed

pkg/controllers/job/job_controller_handler.go
SrinivasChilveri: done

pkg/controllers/job/job_controller_plugins_test.go
SrinivasChilveri: changed

SrinivasChilveri:
@hzxuzhonghu  @k82cn  @TommyLike  , Fixed all the comments

pkg/controllers/job/job_controller.go
hzxuzhonghu: @SrinivasChilveri

pkg/controllers/job/job_controller.go
```diff
@@ -215,20 +239,66 @@ func (cc *Controller) Run(stopCh <-chan struct{}) {
...
+	cc.lock.Lock()
+	queue := cc.queueList[val%cc.workers]
+	cc.lock.Unlock()
+
+	return queue
+}
+
+func (cc *Controller) processNextReq(count uint32) bool {
+	cc.lock.Lock()
+	queue := cc.queueList[count]
```
hzxuzhonghu: Is the lock necessary?

SrinivasChilveri: yeah, I felt as the queueList is accessed by multiple threads paralelly

hzxuzhonghu: nope ， concurrent access an array does not cause race.

hzxuzhonghu: there will be no write ops after the start, right?

SrinivasChilveri: yeah, no write ops, may be i miss understood ur previous comment

hzxuzhonghu 23 days ago  Member
nit: I would like to acquire the queue before this loop

hzxuzhonghu: hha

hzxuzhonghu: >nit: I would like to acquire the queue before this loop

Just ignore this now

SrinivasChilveri: changed

SrinivasChilveri:
@TommyLike  @hex108 any other suggestions.

hzxuzhonghu:
ping @k82cn for approval

Issue #184, opened 2019-05-17
Use Job to generate secret for admission service


TommyLike:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

Now we use helm plugin to generate the admission secret and which seems a little complicated. Instead we can use Job to address this.

**What happened**:

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #191, opened 2019-05-20
Use Job instead of helm plugin to generate admission secret

TommyLike:
See above
For #184

### DIFF
--- a/installer/dockerfile/admission/Dockerfile
+++ b/installer/dockerfile/admission/Dockerfile
@@ -15,5 +15,16 @@
 
 FROM alpine:latest
 
+# Install requirements
+ARG KUBE_VERSION="1.13.1"
+RUN apk add --update ca-certificates && \
+    apk add --update openssl && \
+    apk add --update -t deps curl && \
+    curl -L https://storage.googleapis.com/kubernetes-release/release/v$KUBE_VERSION/bin/linux/amd64/kubectl -o /usr/local/bin/kubectl && \
+    chmod +x /usr/local/bin/kubectl && \
+    apk del --purge deps && \
+    rm /var/cache/apk/*
+
 ADD vk-admission /vk-admission
+ADD gen-admission-secret.sh /gen-admission-secret.sh
 ENTRYPOINT ["/vk-admission"]
--- a/README.md
+++ b/README.md
@@ -99,17 +99,9 @@ kubernetes API server.
 #1. Install helm plugin
 helm plugin install installer/chart/plugins/gen-admission-secret
 
-#2. Generate secret within service name
-helm gen-admission-secret --service <specified-name>-admission-service --namespace <namespace>
-
-## For eg: 
-kubectl create namespace volcano-trial
-
-helm gen-admission-secret --service volcano-trial-admission-service --namespace volcano-trial
-
 ```
 
-Finally, install helm chart.
+Secondly, install helm chart.
 
 ```
 helm install installer/chart --namespace <namespace> --name <specified-name>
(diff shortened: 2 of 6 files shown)


installer/chart/volcano/templates/admission.yaml
```diff
@@ -98,3 +108,23 @@ spec:
...
+  backoffLimit: 3
+  template:
+    spec:
+      serviceAccountName: {{ .Release.Name }}-admission
+      restartPolicy: Never
+      containers:
+        - name: main
+          image: {{.Values.basic.admission_image_name}}:{{.Values.basic.image_tag_version}}
+          imagePullPolicy: IfNotPresent
+          command: ["./gen-admission-secret.sh", "--service", "{{ .Release.Name }}-admission-service", "--namespace", "{{ .Release.Namespace }}"]
```
lminzhw: This job should follow the config `.Values.basic.admission_secret_name` when generate secret.

TommyLike: sure

installer/chart/volcano/templates/admission.yaml
```diff
@@ -98,3 +108,24 @@ spec:
   selector:
     app: volcano-admission
   sessionAffinity: None
+
+---
+apiVersion: batch/v1
```
k82cn: how about init container for admission controller?

TommyLike: The difference is we need to run the init-container every time when new admission pod is created, but job would only execute once.

installer/dockerfile/admission/Dockerfile
```diff
@@ -15,5 +15,16 @@
 
 FROM alpine:latest
 
+# Install requirements
```
k82cn: the file name should be `Dockerfil.secret`, and only include `gen-admission-secret.sh`.

installer/dockerfile/admission/Dockerfile.only_binay
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 The Volcano Authors.
```
k82cn: The file name should be `Dockerfile`.

installer/chart/volcano/templates/admission.yaml
k82cn: That should be ok, as the script will remote the old one. If using job, how to make sure secret is generated before admission pod running?

TommyLike: Pod will wait until the secret is ready because it will mount the secret. when used in init-container it will delete&recreate the secret every time when pod recreated.

installer/dockerfile/admission/Dockerfile
TommyLike: The sh file and binary are kept in one container in default since they are both simple logic and work for admission service, adding one more image can be complicated for maintain and the usage of end user.

k82cn: > adding one more image can be complicated for maintain and the usage of end user

If so, why `Dockerfile.only_binay` ?

installer/chart/volcano/templates/admission.yaml
k82cn: that's ok to me. BTW, Do we delete secret when uninstall volcano?

TommyLike: yeah, we need investigate all of the uncleaned resource and remove them when uninstalling.

installer/dockerfile/admission/Dockerfile
TommyLike: ok, will remove this.

PR #193, opened 2019-05-20
doc for drf fair share

lminzhw:
add doc.


lminzhw:
Related: https://github.com/volcano-sh/kube-batch/pull/14

TommyLike:
@lminzhw  why don't you add this into volcanosh/kube-batch repo?

k82cn:
put every discussion here. only code change in kube-batch

docs/design/drf - fairshare.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+The Scheduler will watch the lifecycle of `ResourceQuota` by `quotaInformer`, and refresh the info in `NamespaceInfo`.
+
+In `openSession` function, we should pass the `NamespaceInfo` through function `cache.Snapshot` into `Session` by using a new filed in `Session`/`ClusterInfo` struct.
+
+```go
+type Session struct {
+    ...
+    NamespaceInfo  map[string]*kbapi.NamespaceInfo
```
k82cn: Prefer to `UserInfo`, similar to other place.

k82cn:
LGTM overall, let's wait for other's feedback on that.

docs/design/drf - fairshare.md
```diff
@@ -0,0 +1,179 @@
...
+kind: ResourceQuota
+metadata:
+  namespace: default
+spec:
+  hard:
+    limits.memory: 2Gi
+    volcano.sh/namespace.weight: 1  <- this field represent the weight of this namespace
+```
+
+If many `ResourceQuota` in the same namespace have weight, the weight for this namespace is the highest one of them.
```
Jeffwan: I don't quite understand the assumption here. Do you mean if user create several `ResourceQuota` in this namespace, the one with highest weight wins?

docs/design/drf - fairshare.md
```diff
@@ -0,0 +1,179 @@
...
+
+All these plugin would choose some victims respective, and the intersection of them would be the final victim set. So, the choice made by DRF plugin would never break the requirement of others.
+
+### short hand
+
+1. Preempt may cause killing of some running pod.
+
+### Cases:
+
+- cluster have __16 cpu__, queue and namespace have default weight.
```
Jeffwan: When namespace requests more resources (calculated from pods) than it's resources quota(from v1.ResourceQuota). At the same time, this namespace has higher weight than the other namespaces in the same queue? Is it a possibility some resources allocated by volcano but failed resource quota check?

docs/design/drf - fairshare.md
lminzhw: > Prefer to `UserInfo`, similar to other place.

Related: https://github.com/volcano-sh/kube-batch/pull/14#discussion_r285819379

docs/design/drf - fairshare.md
lminzhw: Sure. Just for corner case.

docs/design/drf - fairshare.md
lminzhw: For this case, the request of more resources than it's resources quota will cause error like this 

`pods \"lm-fairshare-wnchr-tb-3\" is forbidden: exceeded quota: compute-resources, requested: limits.cpu=500m, used: limits.cpu=2, limited: limits.cpu=2`

So, if that pod is `Pending` and wait for allocation from volcano, it is already satisfied the resource quota.

docs/design/drf - fairshare.md
```diff
@@ -0,0 +1,179 @@
...
+
+DRF plugin use preemption and order of job to balance resource among jobs. The `share` in this plugin is defined as resource usage, the higher `share` means this job occupies the more resource now.
+
+#### Namespace Compare
+
+To introduce namespace weight into this plugin, we should define how to compare namespace having weight firstly.
+
+For namespace n1 having weight w1 and namespace n2 having weight w2, we can compute the `share` of resource and recorded as u1 and u2. Now, the resource usage of n1 less than n2 can be defined as (u1 / w1 < u2 / w2)
+
+`e.g.` ns1 having weight w1=2 use 6cpu, ns2 having weight w2=1 use 2cpu. In the scope of cpu, the ns1 use less resource than ns2. (6 / 3 < 3 / 1)
```
TommyLike: How we get this equation:
```
6/3 < 3/1
```

docs/design/drf - fairshare.md
```diff
@@ -0,0 +1,179 @@
...
+    |       | ns4       | 2 cpu     |                | 2 cpu              |
+
+- cluster have __16 cpu__, q1 with weight 1, q2 with weight 3. ns1 with weight 3, ns2 have weight 1, ns3 have weight 2, ns4 have weight 6.
+
+    | queue | namespace | requested | queue assigned | namespace assigned |
+    | ----- | --------- | --------- | -------------- | ------------------ |
+    | q1 w1 | ns1 w3    | 5 cpu     | 4 cpu          | 3 cpu              |
+    |       | ns2 w1    | 10 cpu    |                | 1 cpu              |
+    | q2 w3 | ns3 w2    | 10 cpu    | 12 cpu         | 10 cpu             |
+    |       | ns4 w6    | 2 cpu     |                | 2 cpu              |
```
TommyLike: namespace4 gets more compared with namespace3 why?

docs/design/drf - fairshare.md
hex108: The hierarchy is `Queue` -> `PodGroup` -> `Pod` now. And there is no namespace now. It means that we could create pods in the same namespace for different queues. How do we deal with it?

docs/design/drf - fairshare.md
```diff
@@ -0,0 +1,179 @@
...
+> 3. record in namespace `quota`
+> 4. higher value means more resources after balancing
+
+### where is the weight
+
+```yaml
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  namespace: default
```
hex108: Is `ResourceQuota` same as Kubernetes [`ResourceQuota`](https://kubernetes.io/docs/concepts/policy/resource-quotas/)? If not, could `volcano.sh/namespace.weight` be named to `weight` because the whole type is `volcano.sh` related? 

Does it need a `name`? Then we could create different quotas of same namespace for different queues.

docs/design/drf - fairshare.md
k82cn: > It means that we could create pods in the same namespace for different queues

Yes, we support that; namespace is considered as a user, so we will balance the user's resources cross queue. If there's one queue per namespace, this feature will not take effect.

docs/design/drf - fairshare.md
k82cn: it's Kubernetes `ResourceQuota` as far as I known :)

hex108: If it is same, it is a little inappropriate to put `volcano.sh/namespace.weight` in `.spec.hard`, because `hard is the set of desired hard limits for each named resource` and `weight` is not a hard limit.

k82cn: > hard is the set of desired hard limits for each named resource and weight is not a hard limit.

It's a kind of `hard` as scheduler will make sure namespace will not be overused.

hex108: It is OK if it is a hard limit.

k82cn:
LGTM overall 👍 , let's wait for other comments :)

docs/design/drf - fairshare.md
jiaxuanzhou: According to the design, ns3 should have only 3 * CPU and 9 * CPU for ns4, i think.

docs/design/drf - fairshare.md
jiaxuanzhou: > It is OK if it is a hard limit.

And once the resource reached the limit, there will be no more pods created for the namespace.

k82cn:
Get it merged firstly; let's open other PRs if any enhancements.

PR #202, opened 2019-05-30
Bump volcano-sh/kube-batch to volcano-sh/scheduler

asifdxtreme:


k82cn:
@asifdxtreme , please help to update this PR to include https://github.com/volcano-sh/scheduler/pull/23 which should fix the error.

PR #225, opened 2019-06-11
Lint Fixes

SrinivasChilveri:
some more lint fixes


k82cn:
@SrinivasChilveri , that'll be better if we can not give enough information by comments.

/cc @asifdxtreme

SrinivasChilveri:
@k82cn , I do agree some comments were not proper. updated withe same with the new commit. if you still feel some more things need update pls do let me know.

asifdxtreme:
lgtm overall

PR #220, opened 2019-06-11
Bump Scheduler

asifdxtreme:
https://github.com/volcano-sh/volcano/issues/219

PR #226, opened 2019-06-12
Fix E2E because of Event Change

thandayuthapani:
Fix E2E
Bump PR https://github.com/volcano-sh/volcano/pull/220 fails, because of Event reason change in Scheduler, So this PR should fix that Bump PR.


k82cn:
@thandayuthapani , we need to also cherry pick this one to volcano-sh/scheduler to enable all e2e cases.

PR #236, opened 2019-06-18
Update README

thandayuthapani:
Helm can be installed without cloning code too,
Have added steps in README to install volcano using helm without cloning code.


README.md
```diff
@@ -53,8 +53,27 @@ the open source community.
...
+Add helm repo using following command,
+
+```
+helm repo add volcano https://volcano-sh.github.io/charts
+```
+
+Install Volcano using following command,
+
+```
+helm install volcano/volcano -namespace <namespace> --name <specified-name>
```
TommyLike: --namespace

README.md
```diff
@@ -53,8 +53,27 @@ the open source community.
 
 ## Quick Start Guide
 
-The easiest way to deploy Volcano is to use the Helm chart.
+The easiest way to deploy Volcano is to use the Helm chart.  Volcano can be deployed by cloning code and also by adding helm repo.
 
+## Using Volcano Helm Repo
+## Helm Charts
```
TommyLike: looks like an incorrect level or redundant header
```
## Helm Charts
```

TommyLike:
Now we have two different repos which both keep the volcano charts codes?

README.md
thandayuthapani: Have removed that header

README.md
thandayuthapani: changed to --namespace

thandayuthapani:
> Now we have two different repos which both keep the volcano charts codes?

Currently we have it in volcano-sh/volcano repo too, once decision on whether to remove or keep it made, we can update README accordingly.

Issue #56, opened 2019-04-01
Add event on actions


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

Currently, we only record event for `Command`s; it's better to also record an event for each actions of jobs.

asifdxtreme:
This feature has been added by the above PR, @Rajadeepan  @k82cn can we go ahead and close this issue?

k82cn:
Great! let's close that.

PR #238, opened 2019-06-18
Adding Events for Action

Rajadeepan:
Adding events for all the actions on the job except the syncjob
https://github.com/volcano-sh/volcano/issues/56

image


pkg/controllers/job/job_controller.go
```diff
@@ -239,6 +242,11 @@ func (cc *Controller) processNextReq() bool {
 	glog.V(3).Infof("Execute <%v> on Job <%s/%s> in <%s> by <%T>.",
 		action, req.Namespace, req.JobName, jobInfo.Job.Status.State.Phase, st)
 
+	if action != vkbatchv1.SyncJobAction {
+		cc.recordJobEvent(jobInfo.Job.Namespace, jobInfo.Job.Name, vkbatchv1.ExecuteAction, fmt.Sprintf(
```
k82cn: That's ok to record event for `SyncJobAction`.

Rajadeepan: If the syncAction event is included it generates a lot of events. If included it is observed that the spamfilter of EventBroadcaster then blocks the similar events on that job object. Because of which new events like restartjob/completejob events were not getting send to the api server.

PR #247, opened 2019-06-24
UT cases for garbagecollector package

thandayuthapani:
For https://github.com/volcano-sh/volcano/issues/168
Added UT cases for garbagecollector package


thandayuthapani:
@nikita15p Thanks for noting it down, Would have missed it while writing.

PR #258, opened 2019-06-26
Adding UT for Queue pkg under cli

shivramsrivastava:
Adding Unit test for Queue package under cli.

```
ok  	volcano.sh/volcano/pkg/cli/queue	0.011s	coverage: 80.0% of statements
```


pkg/cli/queue/queue_test.go
```diff
@@ -0,0 +1,175 @@
+package queue
```
asifdxtreme: Add license header

Issue #272, opened 2019-07-01
Task topology


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

When high performance workload running on Kubernetes, the performance is greatly affected by pod affinity/anti-affinity, e.g. ps/worker mode of Tensorflow. The default scheduler support pod affinity/anti-affinity, but there're some weaknesses:

- The default scheduler will check PodAffinity/PodAntiAffinity against all running pods which impacts schedule performance/throughput
- The default scheduler only check current pod, the coming pods may not have enough resources to meet affinity

Reference:

- [Optimus: An Efficient Dynamic Resource Scheduler for Deep Learning Clusters](https://dl.acm.org/citation.cfm?id=3190517)
- [Pod Affinity and anti-affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity)

gaocegege:
Yeah, I know there are some disadvantages. And it is the reason for me to have a try on volcano scheduler.

I'd appreciate it if you have a proposal to improve it

Thanks :clinking_glasses: :beers:

hzxuzhonghu:
move this feature to 0.3

PR #274, opened 2019-07-02
Design doc of task-topology

k82cn:
part of https://github.com/volcano-sh/volcano/issues/272

### DIFF
--- a/docs/design/bucket.md
+++ b/docs/design/bucket.md
@@ -1,0 +1,103 @@
+# Bucket: Task Affinity/Anti-Affinity within Job
+
+[@k82cn](http://github.com/k82cn); April 2, 2019
+
+## Motivation
+
+When high performance workload running on Kubernetes, the performance is greatly affected by pod affinity/anti-affinity, e.g.
+ps/worker mode of Tensorflow. The default scheduler support pod affinity/anti-affinity, but there're some weaknesses:
+
+- The default scheduler will check PodAffinity/PodAntiAffinity against all running pods which impacts schedule performance/throughput
+- The default scheduler only check current pod, the coming pods may not have enough resources to meet affinity
+
+To improve the performance of scheduler and workload, a new algorithm is introduced, named as Bucket. This doc describes the function
+specification of the new algorithm.
+
+## Function Specification
+
+### Scope
+
+* In Scope
+
+  - Provide a way for user to define task affinity/anti-affinity within a job
+  - Support selecting a better node for the tasks based on affinity/anti-affinity
+
+* Out Of Scope
+
+  - Do not support "Network Topology", e.g. switches  
+  
+### API
+
+A new plugin is introduced to avoid Job API changes, named as `task-topology`. The arguments of the plugin are defined
+as follow: 
+
+```yaml
+plugins
+  task-topology: ["--affinity", "[[ps-task, worker-task]]", "--anti-affinity", "[[ps-task]]"]
+```
+
+**affinity:**
+
+The `--affinity` parameter defines which tasks/pods should be placed on the same node. It's a json two-dimension array:
+
+- First dimension defines which tasks affinity, e.g. `[ps-task, worker-task]`; it only defines the affinity between tasks.
+  The array with only one tasks defines affinity to itself.  
+- Second dimension defines multiple affinity configuration, e.g. `[[ps, worker], [worker]]` means `ps` and `worker` affinity
+  to each other, and `worker` affinity to itself, but not self-affinity preference for `ps` 
+
+**anti-affinity:**
+
+The `--anti-affinity` parameter defines which tasks/pods should not be placed on the same node. It's also a json two-dimension array:
+
+- First dimension defines which tasks anti-affinity, e.g. `[ps-task, worker-task]`; it only defines the anti-affinity between tasks.
+  The array with only one tasks defines anti-affinity to itself.  
+- Second dimension defines multiple anti-affinity configuration, e.g. `[[ps, worker], [ps]]` means `ps` and `worker` anti-affinity
+  to each other, and `ps` anti-affinity to itself, but not self-affinity preference for `worker` 
+
+If there's any conflict, the pod will not be placed in any bucket.
+
+### Function Details
+
+#### Job controller
+
+In `OnJobAdd` callback, the `task-topology` plugin will encode the arguments of `task-topology` into a json string and
+put them into `Job`'s annotation if `volcano.sh/task-topology` not found.
+
+When job controller create `PodGroup` for the job, all its annotations and labels will be passed on to the `PodGroup`. 
+
+#### Bucket Plugin
+
+##### OnSessionOpen
+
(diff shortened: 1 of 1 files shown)


k82cn:
/cc @hex108 @Jeffwan @kinglion811 @gaocegege @richardsliu 

The other operators can provide the same annotation in PodGroup to work with v/scheduler; so please also help to review it.

k82cn:
> I am wondering what will happen if the affinity cannot be satisfied.

 if the affinity cannot be satisfied, the score should be zero; it'll dependent on other plugins' score.

gaocegege:
Gotcha, LGTM

docs/design/bucket.md
```diff
@@ -0,0 +1,103 @@
...
+  - Do not support "Network Topology", e.g. switches  
+  
+### API
+
+A new plugin is introduced to avoid Job API changes, named as `task-topology`. The arguments of the plugin are defined
+as follow: 
+
+```yaml
+plugins
+  task-topology: ["--affinity", "[[ps-task, worker-task]]", "--anti-affinity", "[[ps-task]]"]
```
TommyLike: Have some trouble understanding the `task-topology` usage and explanation

docs/design/bucket.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+**affinity:**
+
+The `--affinity` parameter defines which tasks/pods should be placed on the same node. It's a json two-dimension array:
+
+- First dimension defines which tasks affinity, e.g. `[ps-task, worker-task]`; it only defines the affinity between tasks.
+  The array with only one tasks defines affinity to itself.  
+- Second dimension defines multiple affinity configuration, e.g. `[[ps, worker], [worker]]` means `ps` and `worker` affinity
+  to each other, and `worker` affinity to itself, but not self-affinity preference for `ps`
```
hex108: > not self-affinity preference for `ps`

Does `[ps, worker]` mean `ps` is self-affinity to itself? If not, is there any reason we need add this line?

docs/design/bucket.md
hex108: I doubt it is too specific that adding these annotations/labels in plugins. If we need to deal with other annotation/label, we need change it and restart it? 

Need we add a section to tell how to use this feature, using annotation or label?

docs/design/bucket.md
```diff
@@ -0,0 +1,103 @@
...
+**anti-affinity:**
+
+The `--anti-affinity` parameter defines which tasks/pods should not be placed on the same node. It's also a json two-dimension array:
+
+- First dimension defines which tasks anti-affinity, e.g. `[ps-task, worker-task]`; it only defines the anti-affinity between tasks.
+  The array with only one tasks defines anti-affinity to itself.  
+- Second dimension defines multiple anti-affinity configuration, e.g. `[[ps, worker], [ps]]` means `ps` and `worker` anti-affinity
+  to each other, and `ps` anti-affinity to itself, but not self-affinity preference for `worker` 
+
+If there's any conflict, the pod will not be placed in any bucket.
```
hex108: Ah, what is `bucket`?

docs/design/bucket.md
TommyLike: In case someone miss the background of `plugins`, it's a existing feature in [volcano](https://github.com/volcano-sh/volcano/blob/ae619b489407207cc5dea21a76d3caf9ffbafb35/pkg/apis/batch/v1alpha1/job.go#L66)

docs/design/bucket.md
```diff
@@ -0,0 +1,103 @@
...
+#### Bucket Plugin
+
+##### OnSessionOpen
+
+In `OnSessionOpen` of `bucket` plugin, it'll get `task-topology` from `PodGroup`'s annotation and build bucket accordingly.
+The buckets of job is only maintained in `bucket` plugin, similar with other plugins. 
+
+In `bucket` plugin, it'll register a `nodeOrderFn` and `taskOrderFn` to find a better node to place the bucket:
+ 
+ - `taskOrderFn` will sort all tasks by buckets, e.g. ps1, worker1, worker2, ps2, worker3, worker4, the pods without bucket are
```
hzxuzhonghu: Can hardly understand what does it do

docs/design/bucket.md
k82cn: > Does [ps, worker] mean ps is self-affinity to itself?

No, highlight behaviour.

docs/design/bucket.md
k82cn: > Need we add a section to tell how to use this feature, using annotation or label?

Yes, we need to define its format for other operators, e.g. kubeflow/tf-operator.

docs/design/bucket.md
k82cn: for `2ps + 4workers` case, `1ps + 2workers` is a bucket; so there're two bucket, and we need to find a better host to dispatch all pod in each bucket.

docs/design/bucket.md
lminzhw: here, the behavior of `taskOrderFn` is to sort the task by bucket, and finally the tasks with same bucket could be allocated in close time.

docs/design/bucket.md
```diff
@@ -0,0 +1,103 @@
...
+In `OnJobAdd` callback, the `task-topology` plugin will encode the arguments of `task-topology` into a json string and
+put them into `Job`'s annotation if `volcano.sh/task-topology` not found.
+
+When job controller create `PodGroup` for the job, all its annotations and labels will be passed on to the `PodGroup`. 
+
+#### Bucket Plugin
+
+##### OnSessionOpen
+
+In `OnSessionOpen` of `bucket` plugin, it'll get `task-topology` from `PodGroup`'s annotation and build bucket accordingly.
```
SrinivasChilveri: I think we have to consider --affinity and --anti-affinity  differently  while choosing the node for a tasks related to job.  as bucket is group of  related tasks of a job.

docs/design/bucket.md
```diff
@@ -0,0 +1,103 @@
...
+The buckets of job is only maintained in `bucket` plugin, similar with other plugins. 
+
+In `bucket` plugin, it'll register a `nodeOrderFn` and `taskOrderFn` to find a better node to place the bucket:
+ 
+ - `taskOrderFn` will sort all tasks by buckets, e.g. ps1, worker1, worker2, ps2, worker3, worker4, the pods without bucket are
+   placed at the end of task queue.
+ - in `nodeOrderFn`, it'll calculate a score for the node as follow:
+   - if node has enough resource to place bucket, return `bucket / node`
+   - if node does not have enough resource to place the bucket, keeps removing items from the bucket until the node can place the bucket;
+     and then return `bucket / node`. When removing items, balance Pod of `taskSpec`s in the bucket.
```
SrinivasChilveri: does removing items means, suppose we have 1 ps and 3 workers replicas with  --affinity [ps,worker] then the ps & atleastone worker will be on one node(suppose reourcs available on that Node for that much only) and other workers can be on other nodes??? or am i missing something here.

k82cn:
/close

I'll open a new PR later to clearup branch :)

Issue #266, opened 2019-07-01
Support multiple events in lifecycle policy


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

Currently, `spec.policy` only support one event or `*` which is not enough. It's better to support an event list for it. It need to handle conflict, but duplication is not necessary to me.

asifdxtreme:
@k82cn the related PR's for this feature has been merged, Can we go ahead and close this issue?

PR #285, opened 2019-07-03
Support multiple events in the lifecycle policy

Rajadeepan:
Support multiple events in the lifecycle policy. updating events to list of events.
https://github.com/volcano-sh/volcano/issues/266

This PR need to be updated with updated submodule of charts after the PR https://github.com/volcano-sh/charts/pull/5 gets merged.


k82cn:
> This PR need to be updated with updated submodule of charts after the PR volcano-sh/charts#5 gets merged.

We need cherry-pick CRD when this PR is merged.

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -178,7 +178,7 @@ type LifecyclePolicy struct {
 	// The Event recorded by scheduler; the controller takes actions
 	// according to this Event.
 	// +optional
-	Event Event `json:"event,omitempty" protobuf:"bytes,2,opt,name=event"`
+	Event []Event `json:"event,omitempty" protobuf:"bytes,2,opt,name=event"`
```
k82cn: Add a new member; we already publish this CRD; we can not change it.

pkg/admission/admission_controller.go
```diff
@@ -115,32 +115,42 @@ func validatePolicies(policies []v1alpha1.LifecyclePolicy, fldPath *field.Path)
...
+					break
+				}
+
+				if allow, ok := policyActionMap[policy.Action]; !ok || !allow {
+					err = multierror.Append(err, field.Invalid(fldPath, policy.Action, fmt.Sprintf("invalid policy action")))
+					bFlag = true
+					break
+				}
+				if _, found := policyEvents[event]; found {
+					err = multierror.Append(err, fmt.Errorf("duplicate event %v", event))
```
k82cn: that's ok if duplicated in the same array.

pkg/apis/batch/v1alpha1/job.go
Rajadeepan: Added a new member EventList

pkg/admission/admission_controller.go
Rajadeepan: Code updated to ignore duplicates in the same array

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -180,6 +180,11 @@ type LifecyclePolicy struct {
 	// +optional
 	Event Event `json:"event,omitempty" protobuf:"bytes,2,opt,name=event"`
 
+	// The EventList recorded by scheduler; the controller takes actions
+	// according to this EventList.
+	// +optional
+	EventList []Event `json:"eventlist,omitempty" protobuf:"bytes,3,opt,name=eventlist"`
```
k82cn: Just call `Events`

Rajadeepan: Updated to Events

pkg/admission/admission_controller.go
TommyLike: The duplicated events are already been removed in codes of:
```
policyEventsList := getEventlist(policy)
```
?

Rajadeepan: If there are duplicate events within **one policy** it will be removed in the codes of:
  ` golicyEventList := getEventlist(policy)`

But if there are duplicate events **across different policy** then
  Error message will be thrown `("duplicate events %v",event)`

TommyLike: ok, then we may need more clear error message for that since the first case is being handled automatically now.

Rajadeepan: can  the error message be updated  as ` ("duplicate events %v across different policy", event)` ?

TommyLike: That would be good thanks.

TommyLike:
cc @k82cn

k82cn:
/lgtm

Please also open PR for e2e test.

Rajadeepan:
> /lgtm
> 
> Please also open PR for e2e test.

ok will raise a new PR for e2e test

PR #298, opened 2019-07-07
Add design document for reserve action

sivanzcw:


k82cn:
@sivanzcw , please open a separate PR for the design doc. It'll be better to make sure all are on the same page before reviewing PR :)

docs/design/reserve-action.md
```diff
@@ -0,0 +1,152 @@
...
+
+The `reserve` action prevents the large resource requirement task from being starved by reserving the node resources for
+it and not allowing the other tasks to be scheduled on the node.
+
+## Function Detail 
+
+### Job starving
+
+If the `pending` time of job is greater that the set time threshold, the task has the risk of starvation, and needs to
+reserve node resources for it. If there is unscheduled task under the job, but the job has already met the `running`
```
hzxuzhonghu: Is there a risk the unscheduled tasks of the job can never be scheduled?

docs/design/reserve-action.md
```diff
@@ -0,0 +1,152 @@
...
+reserve node resources for it. If there is unscheduled task under the job, but the job has already met the `running` 
+condition and is already `running`, we think that this job does not need to reserve node resource any more.
+
+### Node resource reservation rule
+
+If the allocatable resources on the node is greater that the resource request of the task, then the node satisfies the 
+scheduled requirement of the task now or later. This node can reserve resources for this task. If the idle resource of 
+the current node has met the scheduling condition of the task, resources will be allocated to this task, and the task 
+will be directly assigned to the node. If the idle resource of the node can not meet the resource requirement of the task,
+the node will reserve resources for this task. Once a node reserve resources for a task, no other tasks will be scheduled
```
hzxuzhonghu: If two tasks require affinity, then what will happen?

hzxuzhonghu: L57 should answer me

docs/design/reserve-action.md
```diff
@@ -0,0 +1,152 @@
...
+
+<img src="http://chart.googleapis.com/chart?cht=tx&chl=\Large score=100-\frac{used}{allocatable}" style="border:none;">
+
+The formula for calculating the node score based on the above three resources is
+
+<img src="http://chart.googleapis.com/chart?cht=tx&chl=\Large score=\frac{\frac{cpuScore\pm memoryScore}{2}\pm2*sclarScore}{3}" style="border:none;">
+
+## Data Model
+
+Introduce one node resource `Reserved` resource, used to indicate the size of resources reserved on the node. The size
```
hzxuzhonghu: >Introduce one node resource `Reserved` resource,

one field?

docs/design/reserve-action.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Data Model
+
+Introduce one node resource `Reserved` resource, used to indicate the size of resources reserved on the node. The size 
+of the reserved resource on the node is the sum of the resource requests of all the pods reserved on the node. The 
+reserved resources should be smaller than the allocatable resources.
+```go
+	
+// Reserved is reserved resources for starving task
+Reserved *Resource
```
hzxuzhonghu: What's the struct?

sivanzcw: // Resource struct defines all the resource type
type Resource struct {
	MilliCPU float64
	Memory   float64

	// ScalarResources
	ScalarResources map[v1.ResourceName]float64

	// MaxTaskNum is only used by predicates; it should NOT
	// be accounted in other operators, e.g. Add.
	MaxTaskNum int
}

docs/design/reserve-action.md
sivanzcw: including `cpu` `memory` and `scalar` three resources fields

docs/design/reserve-action.md
sivanzcw: In each round, the node with the most sufficient idle node is selected as the reserved node, so after several rounds of scheduling, the selected nodes will gradually converge, and finally meet the resource scheduling requirements of the task. There will be some fluctuations in this process. Fluctuation will make the task scheduled longer, and in extreme cases it will not converge, but the probability of this scene is low and can be ignored.

docs/design/reserve-action.md
sivanzcw: The `PredicateFn` will be called during selecting the reserved node. For a pair of tasks configured with affinity, after the first task reserves resources on the node successfully, the result of `PredicateFn` for the second task will be the node who reserve resources for the first task. If the node can also reserve resources for the second task, the second task will be reserved on this node, otherwise the sencond task will be failed to reserve resources in this round. Regardless of the results of the above two tasks, for the job of these two tasks, if the state of the job can be ready, the reservation of the node for tasks will be performed, otherwise, the reservation will be discarded.

docs/design/reserve-action.md
```diff
@@ -0,0 +1,254 @@
...
+      - [Node reserve logic](#node-reserve-logic)
+      - [The proportion of nodes participating in reserve action](#the-proportion-of-nodes-participating-in-reserve-action)
+      - [Node selection strategy to participate in reserve](#node-selection-strategy-to-participate-in-reserve)
+      - [Node calculating score logic](#node-calculating-score-logic)
+  * [Data Model](#data-model)
+  
+## Introduction
+
+The `reserve` action is to solve the problem that the task can not be scheduled for a long time. We call these long 
+unscheduled jobs starving job, if a job is not scheduled for a long time, it is a starving job. The `reserve` action
```
gaocegege:
```suggestion
unscheduled jobs starving jobs, if a job is not scheduled for a long time, it is a starving job. The `reserve` action 
```

docs/design/reserve-action.md
```diff
@@ -0,0 +1,254 @@
...
+sorted in descending order of rank.
+
+In the process of calculating node scores based on node idle resource, there will be `cpu` `memory` and `scalar resources` 
+participating in the score calculation. The formula for calculating score for a single resource is  
+
+<img src="http://chart.googleapis.com/chart?cht=tx&chl=\Large score=100-\frac{usedResource}{allocatableResource}" style="border:none;">
+
+The formula for calculating the node score based on the above three resources is
+
+<img src="http://chart.googleapis.com/chart?cht=tx&chl=\Large score=\frac{\frac{cpuScore%2b memoryScore}{2}%2b2*sclarScore}{3}" style="border:none;">
```
gaocegege: What's the meaning of [LargScore?]

docs/design/reserve-action.md
sivanzcw: done

docs/design/reserve-action.md
sivanzcw: `score` and `\Large` are separate here, and `\Large` is the formula syntax, indicating that the font of the following letters becomes larger

image

docs/design/reserve-action.md
hzxuzhonghu: I mean where does Reserved reside in ?  You should explicitly specify in design

sivanzcw: updated

docs/design/reserve-action.md
```diff
@@ -0,0 +1,261 @@
...
+    + [Node resource reservation rule](#node-resource-reservation-rule)
+      - [Node reserve logic](#node-reserve-logic)
+      - [The proportion of nodes participating in reserve action](#the-proportion-of-nodes-participating-in-reserve-action)
+      - [Node selection strategy to participate in reserve](#node-selection-strategy-to-participate-in-reserve)
+      - [Node calculating score logic](#node-calculating-score-logic)
+  * [Data Model](#data-model)
+  
+## Introduction
+
+The `reserve` action is to solve the problem that the task can not be scheduled for a long time. We call these long
```
hex108: Could we solve it by `preempt`?

k82cn: if jobs/pods are in the same "priority", we can not trigger preemption :)

docs/design/reserve-action.md
hex108: It seems that the pic has not been fixed.

docs/design/reserve-action.md
```diff
@@ -0,0 +1,261 @@
...
+try to schedule `pod-4`, but idle resource of node can not satisfy pod's requirement. The `pod-4` can not be scheduled and 
+was skipped. The schedule try to schedule `pod-5` and succeed in `Time3`. In `Time4`, `pod-2` and `pod-3` were `completed`
+and `2C2G` was released from node. Because of that the idle resource of node can not satisfy requirement of pod `pod-4`, `pod-4`
+was skipped. The `pod-6` and `pod-7` were bound to node and begin to run. In `Time6` all pods except pod `pod-4` were completed
+and all resources of node were release. The scheduling condition of pod `pod-4` is satisfied, `pod-4` was bound to node and 
+begin to run. In the above scenario, the `pod-4` can not be scheduled in time because a large number of small resource requirement
+pods occupied the node resources.
+ 
+In order to solve such problems, the reserve action is introduced into the scheduling policy. The `reserve` action can 
+prevents the large resource requirement task from being starved by reserving node resources for it and not allowing the other
```
hex108: How long will tasks be considered starved? Is it configured?

docs/design/reserve-action.md
```diff
@@ -0,0 +1,261 @@
...
+* Job state is not `Job Condition Ready`
+
+If a job has been created for a long time, but it is still not `running`, it is at risk of starving. If nodes under the cluster
+have reserved resource for the starving job, then there is no need to reserve resources for it again.
+
+#### ForceAllocated
+
+If a node tries to reserve resources for the task under the job, when the node has not reserved resources for other tasks, and 
+the idle resources of the node are larger than the requirement resources of the task, this task can be directly allocated to 
+this node with `ForceAllocated` state.  The `ForceAllocated` state is to avoid subsequent `backfill` action to preempt pod
```
hex108: Could `preempt` action preempt `ForceAllocated` and `FakeAllocated` pods?

docs/design/reserve-action.md
```diff
@@ -0,0 +1,261 @@
...
+
+Because of that some of node resources was reserved for tasks and will not be used by other tasks, the resources 
+reservation will affect the task scheduling process of the entire cluster. In theory, the more nodes that participate in
+resource reservation, the better it is to solve the problem of the task being starved to death. However, the fewer nodes 
+participating in regular scheduling, and the greater the impact on regular cluster scheduling. Therefore, the number of 
+nodes participating in resource reservation is configurable.
+
+#### Node selection strategy to participate in reserve
+ 
+How to choose which nodes will be selected as candidate nodes for starving job scheduling. The process of selecting a
```
hex108: After nodes are reserved for a task, could it be scheduled to other nodes if they have enough resource?

docs/design/reserve-action.md
hex108: Is there any reference for this formula?

docs/design/reserve-action.md
```diff
@@ -0,0 +1,261 @@
...
+## Function Detail 
+
+### Explanation of terms
+
+#### Job Ready
+
+If the running condition of a job is satisfied, then the job is in the `Job Ready` state, and after scheduling, the job can be in the 
+`Job Running` state. The running condition of a job may include restrictions such as `gang-scheduler` and so on.
+
+#### Job Condition Ready
```
hex108: It is defined, but never used in this PR?

docs/design/reserve-action.md
sivanzcw: This parameter can be configured, configured by the arguments parameter of the reserve action.

docs/design/reserve-action.md
sivanzcw: The `preempt` action can only evict `running` tasks, will not evict `FakeAllocated` or `ForceAllocated` tasks

hex108: It seems not very reasonable that high priority task could not get resource?

docs/design/reserve-action.md
sivanzcw: In a scheduling round, if the task is reserved for resources, its state will be `ForceAllocated` or `FakeAllocated`. Subsequent actions will only process tasks in the `pending` state, and will not process these tasks. If there is a situation in which the idle resource of the node meets the demand of the task, the `reserve` action needs to ensure that the task is allocated to the node during the `reserve` phase. In the `reserve` phase, nodes with sufficient resources will be selected as candidate nodes. If the node's allocatable resource is larger than the task's required resource, the task will be `ForceAllocated` to the node.

docs/design/reserve-action.md
sivanzcw: This will be used to judge whether the job need to be reserved resources. If there are already nodes that have reserved resources for this job, there is no need to perform a `reserve` operation for them.

docs/design/reserve-action.md
sivanzcw: Yes, in the `reserve` process, when trying to reserving resource for a starving task, we need to ensure that the tasks whose priority is higher than the starving task will be prioritized to schedule and prioritized to get resources. I will make an update here.

docs/design/reserve-action.md
hex108: How to deal with this situation: In the first reserve pass, node N's idle resource is very low, and its score is very low, so it was not reserved for task T, and we reserved other nodes for task T. However after some time(those scheduler phases have passed again and again), node N has enough idle resource for task T, and those reserved nodes still do not have enough idle resource for it. Will we allocate node N for task T?

docs/design/reserve-action.md
hex108: But it is not used in other places in this PR?

docs/design/reserve-action.md
sivanzcw: Reserved node selection is a dynamic process, and the nodes reserved in each scheduling round may be different. When selecting candidate nodes, the score of the nodes is firstly calculated and sorted according to the nodes score, so that the node selected for this task in this time will be better than the node selected in the previous round. As the scheduling round progresses, the final filtered nodes will get better and better until the task resource requirement is met.

docs/design/reserve-action.md
sivanzcw: there is a quote in `Job starving`
···
#### Job starving

The starving job needs to meet the following conditions

* The creation time of job is greater than the set starving job threshold time
* Job state is not `Job Ready`
* Job state is not `Job Condition Ready`

···

hex108: Sorry, just missed it.

docs/design/reserve-action.md
hex108: Get it now, thanks!

docs/design/reserve-action.md
sivanzcw: Added a new chapter `Reserve ligic` to introduce the logic of processing jobs who have higher priority than the starving job.

docs/design/reserve-action.md
hex108: What about this question?

sivanzcw: The formula here uses the syntax of html, calls the api of google chart, generates the mathematical formula of Latex, and can be seen by tool of markdown.
referenced this document
https://developers.google.com/chart/image/docs/making_charts
image

hex108: Is there any reference for this formula?

docs/design/reserve-action.md
```diff
@@ -0,0 +1,273 @@
...
+      - [The proportion of nodes participating in reserve action](#the-proportion-of-nodes-participating-in-reserve-action)
+      - [Node selection strategy to participate in reserve](#node-selection-strategy-to-participate-in-reserve)
+      - [Node calculating score logic](#node-calculating-score-logic)
+    + [Reserve logic](#reserve-logic)
+  * [Data Model](#data-model)
+
+## Introduction
+
+The `reserve` action is to solve the problem that the task can not be scheduled for a long time. We call these long 
+unscheduled jobs starving jobs, if a job is not scheduled for a long time, it is a starving job. The `reserve` action
```
k82cn: s/,/; for `unscheduled jobs starving jobs, if a job is not sc`

docs/design/reserve-action.md
sivanzcw: updated to add reference for GoogleChart quota in `References`

docs/design/reserve-action.md
```diff
@@ -0,0 +1,278 @@
...
+
+If the running condition of a job is satisfied under certain conditions, then the state of the job can be called `Conditional Running`.
+When after the node reserving resources for the tasks under the job, the job can reach the running condition, then the state of 
+the job can be called `Condition Ready`.
+
+#### Job starving
+
+The starving job needs to meet the following conditions
+
+* The creation time of job is greater than the set starving job threshold time
```
k82cn: This should be a callback, and `creation time > threshold` is the default implementation

docs/design/reserve-action.md
```diff
@@ -0,0 +1,278 @@
...
+* Job state is not `Job Ready`
+* Job state is not `Job Condition Ready`
+
+If a job has been created for a long time, but it is still not `running`, it is at risk of starving. If nodes under the cluster
+have reserved resource for the starving job, then there is no need to reserve resources for it again.
+
+#### ForceAllocated
```
k82cn: Why introduces two state? I'm thinking just trace `Reserved` resources should be enough, ignore the relationship between idle resources and request.

docs/design/reserve-action.md
```diff
@@ -0,0 +1,278 @@
...
+nodes participating in resource reservation is configurable.
+
+#### Node selection strategy to participate in reserve
+ 
+How to choose which nodes will be selected as candidate nodes for starving job scheduling. The process of selecting a 
+reserved node will include the following steps.
+
+1. Get all nodes in cluster
+2. PredicateNodes for task
+3. Calculating node score by idle resource and resource requirement of task
```
k82cn: we should delegate this to plugin.

docs/design/reserve-action.md
```diff
@@ -0,0 +1,278 @@
...
+A node may reserve resources only for one task, or may reserve resources for multiple tasks, depending on the size of the 
+allocatable resources of node.
+
+#### The proportion of nodes participating in reserve action
+
+Because of that some of node resources was reserved for tasks and will not be used by other tasks, the resources 
+reservation will affect the task scheduling process of the entire cluster. In theory, the more nodes that participate in
+resource reservation, the better it is to solve the problem of the task being starved to death. However, the fewer nodes 
+participating in regular scheduling, and the greater the impact on regular cluster scheduling. Therefore, the number of 
+nodes participating in resource reservation is configurable.
```
k82cn: how to configure that?

docs/design/reserve-action.md
k82cn: let's save the pic locally :)

docs/design/reserve-action.md
```diff
@@ -0,0 +1,278 @@
...
+
+```
+
+Introduce two task status `ForceAllocated` and `FakeAllocated`. The `ForceAllocated` indicates that the pod is assigned 
+to this node, occupies the corresponding resources of the node, and does not allow other tasks to backFill these resources.
+The `FakeAllocated` indicates that  although the task is added to the node, it does not actually occupy the resources of 
+the node, it just wants the node to reserve resources for it. 
+
+```go
+// FakeAllocated means a task/pod is fake allocated to the node
```
k82cn: can we re-use "Pipelined" state? or introduce only one state, e.g. "Waiting".

docs/design/reserve-action.md
```diff
@@ -0,0 +1,278 @@
...
+// this is the task is added to the node, but not allocated resources for it
+FakeAllocated
+
+// ForceAllocated means a task/pod is allocated to the node
+ForceAllocated
+```
+The `Reserved` resources on the node will be increased to 
+indicate that the node reserves resources for the tasks, at the same time, the idle resources of the node will be reduced
```
k82cn: > the idle resources of the node will be reduced to prevent the node from providing resources for other pods.

IMO, that maybe not necessary to modify idle resources.

k82cn:
btw, when we reserve resource for task, should we also call `Allocated/Deallocated` to each plugins to update their data, e.g. DRF?

docs/design/reserve-action.md
sivanzcw: I will replace it

docs/design/reserve-action.md
sivanzcw: I will add some description for the judgment of starving job: encapsulating the judgment into a function that embeds in the plugin.

docs/design/reserve-action.md
sivanzcw: The introduction of the `ForceAllocated` state is expected to replace the `allocate` state. In the `reserve` phase, for starving jobs, if node idle resources satisfy pod's resource requirement, the pod scheduling may be completed. So, the pod needs an identifier that occupies the actual resource. In order to avoid preempting the resources that have been allocated to the pod in the `backfill` phase, a new state is defined. Introducing the `FakeAllocate` state is to make an identifier for whether the node reserves resources and whether the task is reserved. If task with `FakeAllocate` status is added to the node, idle resource of node will be decreased, so the node with `FakeAllocate` tasks indicates that the idle resources of the node are tight and need to overdraw resources. The `Reserved` resource can also determine whether the node has reserved resources for the task, and whether there are remaining resources to schedule other tasks, but it is possible that this judgment is not as clear as the status identifier.

docs/design/reserve-action.md
sivanzcw: I will update it

docs/design/reserve-action.md
sivanzcw: Configured by `arguments` of `reserve` action by `reserveNodePercent` parameter, which means the percent of node participate in resoure reserve. Configured by service

docs/design/reserve-action.md
sivanzcw: Using `pipelined` may be a bit difficult. Although the concepts of the two are similar, the `pipelined` is basically bound to the node's `releasing` resources, and the `reserve` action needs to operate the node's `reserve` resources.

docs/design/reserve-action.md
sivanzcw: Updating the `idle` resources of node is to avoid subsequent actions to dispatch tasks to this node. However, in order to ensure that the logic of node score calculation based on the idle resource is fair to all the tasks, it is necessary to ensure that the idle resources of the node are stable during the `reserve` process. I will try a new logic to ensure that the node who has reserve resources no longer dispatches other tasks by other logic, instead of operating the `idle` resources of the node.

sivanzcw:
> btw, when we reserve resource for task, should we also call `Allocated/Deallocated` to each plugins to update their data, e.g. DRF?

According to the current logic, during the `reserve` processing, when the task is `ForceAllocated` to the node, the `AllocateFunc` function will be called to update the resources maintained by the plugin.

docs/design/reserve-action.md
sivanzcw: done

docs/design/reserve-action.md
sivanzcw: updated it

docs/design/reserve-action.md
sivanzcw: add some description about `isJobStarving`  function in L137

docs/design/reserve-action.md
sivanzcw: A new `Reserved` state is added to indicate that the task is reserved on the node, and the `ForceAllocate` state is replaced by the `Allocated` state.

docs/design/reserve-action.md
sivanzcw: done

docs/design/reserve-action.md
sivanzcw: replace with `Reserved` state

docs/design/reserve-action.md
sivanzcw: The idle resource of node is no longer updated for the `reserved` task. Using the `Reserved` resource of the node to indicate whether the node can schedule a new non-BestEffort task.

Issue #294, opened 2019-07-05
Scheduler panic happens when the GPU is lost on node


william-wang:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> /kind bug

**What happened**:
Scheduler panic happens on starting whenever the GPU card is lost on node.

**What you expected to happen**:
Scheduler should works even the GPU is lost on some node. In this case, the node
can be mark as NotReady.

**How to reproduce it (as minimally and precisely as possible)**:
This case can not be reproduced inhouse.  The user hitted this issue from time.

**Environment**:
- Volcano Version: latest version

PR #306, opened 2019-07-08
Fix the scheduler panic whenever the GPU is lost on node

william-wang:
The volcano-sh/scheduler panics when startup whever the GPU is lost on nodes.
The cause is that the GPU resource descreaced, the node.idle <task.Req. The panic happened
when ni.Idle.Sub(ti.Resreq). This PR is used to fix the issue.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes volcano-sh/volcano#294


pkg/scheduler/api/node_info.go
```diff
@@ -176,12 +188,16 @@ func (ni *NodeInfo) AddTask(task *TaskInfo) error {
 	if ni.Node != nil {
 		switch ti.Status {
 		case Releasing:
+			if err := ni.allocateIdleResource(ti); err != nil {
+				return err
+			}
 			ni.Releasing.Add(ti.Resreq)
-			ni.Idle.Sub(ti.Resreq)
 		case Pipelined:
 			ni.Releasing.Sub(ti.Resreq)
```
hzxuzhonghu: Does this face the same problem?

k82cn:
@asifdxtreme , please help to cherry-pick this on to kube-batch :)

k82cn:
@asifdxtreme , please help to cherry-pick this one to kube-batch :)

pkg/scheduler/api/node_info.go
william-wang: Yes, So I deleted this line and replace it to line 191-193.

asifdxtreme:
@william-wang Can you please take out some time and sign cncf-cla https://github.com/kubernetes/community/blob/master/CLA.md#the-contributor-license-agreement , this PR's needs to be cherry picked to kube-batch and kube-batch has mandatory check for cncf-cla

k82cn:
@asifdxtreme , I think we also need cncf-cla check for volcano :)

asifdxtreme:
> @asifdxtreme , I think we also need cncf-cla check for volcano :)

okay we will add it

pkg/scheduler/api/node_info.go
william-wang: > @william-wang Could you add more comments why this can cause a panic?
In customer's environment, when the GPU was lost. the ni.Idle decreased and  the value was less than ti.Resreq. Then the panic happened in Sub func. So we need to add a judgement before the execution of ni.Idle.Sub(ti.Resreq).

asifdxtreme:
> @asifdxtreme , I think we also need cncf-cla check for volcano :)

For CNCF Cla check we need webhook secret from CNCF, we contacted the admins to get the secret but they replied back saying they cannot enable webhook for external repo's as of now.
image
so probably we cannot enable CNCF cla check now for Volcano

PR #307, opened 2019-07-08
Some yaml fix and cleanups after kube-batch in tree

hzxuzhonghu:


cmd/kube-batch/app/server.go
```diff
@@ -72,7 +71,7 @@ func buildConfig(opt *options.ServerOption) (*rest.Config, error) {
 	return cfg, nil
 }
 
-// Run the kubeBatch scheduler
+// Run the kube-batch scheduler
```
k82cn: Should be volcano scheduler.

hzxuzhonghu:
cc @TommyLike

cmd/kube-batch/app/server.go
hzxuzhonghu: sorry, i missed this.

As there are many cleanup work left, let me open an issue

PR #313, opened 2019-07-09
Update kube-batch to scheduler

asifdxtreme:


Makefile
```diff
@@ -13,13 +13,13 @@ LD_FLAGS=" \
 
 .EXPORT_ALL_VARIABLES:
 
-all: kube-batch vc-controllers vc-admission vkctl
+all: scheduler vc-controllers vc-admission vkctl
```
k82cn: vc-scheduler

Issue #311, opened 2019-07-09
Volcano controller logs is too noisy


hzxuzhonghu:
```
I0709 09:36:57.282999       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-controllers-84f45c65fc-npfhr>, skipping
I0709 09:36:57.285724       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-controllers-84f45c65fc-npfhr>, skipping
I0709 09:36:57.681311       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-scheduler-585b479f76-8ktls>, skipping
I0709 09:36:57.683399       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-scheduler-585b479f76-8ktls>, skipping
I0709 09:36:58.078774       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-admission-init-zxg86>, skipping
I0709 09:36:58.078776       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-admission-init-zxg86>, skipping
I0709 09:36:58.479090       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-admission-init-zxg86>, skipping
I0709 09:36:58.479126       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-admission-init-zxg86>, skipping
I0709 09:37:01.398778       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-admission-5c4c8dc76c-lvj6z>, skipping
I0709 09:37:01.398998       1 job_controller_handler.go:191] Failed to find taskName of Pod <kube-system/laughing-seal-admission-5c4c8dc76c-lvj6z>, skipping
```
For pods which do not belong to volcano job, donot process it.

hzxuzhonghu:
@Rajadeepan @thandayuthapani  would appreciate if you could improve it.

Rajadeepan:
ok sure will improve it

PR #316, opened 2019-07-09
Removing logs which does not belong to volcano job

Rajadeepan:
Removing logs which does not belong to volcano job :

https://github.com/volcano-sh/volcano/issues/311

### DIFF
--- a/pkg/controllers/job/job_controller_handler.go
+++ b/pkg/controllers/job/job_controller_handler.go
@@ -134,15 +134,11 @@ func (cc *Controller) addPod(obj interface{}) {
 
 	jobName, found := pod.Annotations[vkbatchv1.JobNameKey]
 	if !found {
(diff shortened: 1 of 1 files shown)


pkg/controllers/job/job_controller_handler.go
```diff
@@ -134,15 +134,11 @@ func (cc *Controller) addPod(obj interface{}) {
 
 	jobName, found := pod.Annotations[vkbatchv1.JobNameKey]
 	if !found {
-		glog.Infof("Failed to find jobName of Pod <%s/%s>, skipping",
-			pod.Namespace, pod.Name)
```
hzxuzhonghu: I would rather leave it here, and check the ownerreference of the pod at the begining.
Because this is so silent, we can not know what happened when unexpected result occurs.

PR #320, opened 2019-07-10
set drf dominant resource name

hzxuzhonghu:


pkg/scheduler/plugins/drf/drf.go
```diff
@@ -155,23 +158,25 @@ func (drf *drfPlugin) OnSessionOpen(ssn *framework.Session) {
 }
 
 func (drf *drfPlugin) updateShare(attr *drfAttr) {
-	attr.share = drf.calculateShare(attr.allocated, drf.totalResource)
+	attr.dominantResource, attr.share = drf.calculateShare(attr.allocated, drf.totalResource)
```
hex108: Where will `dominantResource` be used?

hzxuzhonghu: should be logged

PR #326, opened 2019-07-11
Fix issue in makefile

SrinivasChilveri:
root@root1-HP-EliteBook-840-G2:/home/root1/SriniCH/workspace/src/volcano.sh/volcano# make
make: *** No rule to make target 'vc-scheduler', needed by 'all'.  Stop.

this issue is there on latest code with commit

commit c1cbbc1292e1adc4eae3a5f14062396a444960e0

make should  build the binaries properly, but its failing because of recent changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ all: vc-scheduler vc-controllers vc-admission vkctl
 init:
 	mkdir -p ${BIN_DIR}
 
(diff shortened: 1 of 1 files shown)


hex108:
@asifdxtreme The CI could not find this kind of issue?

Issue #310, opened 2019-07-09
Volcano install failed


hzxuzhonghu:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:

 /kind bug

**What happened**:

install volcano as readme, but i meet the following error.

```
$ helm install volcano/volcano --namespace kube-system --name volcano
Error: namespaces "kube-system" is forbidden: User "system:serviceaccount:kube-system:default" cannot get resource "namespaces" in API group "" in the namespace "kube-system
```

**What you expected to happen**:

1. install successfully 

1. allow user to set volcano flags 

1. provide a simple install yaml

**How to reproduce it (as minimally and precisely as possible)**:

install volcano following  https://github.com/volcano-sh/volcano/blob/master/README.md

**Anything else we need to know?**:

my kubernetes cluster is created by `kind`

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

k82cn:
We should make sure the readme/tutorial correct :)

hzxuzhonghu:
I have opened #325 to track contributor experience improvement stuff.

hzxuzhonghu:
We need a more simple way to install, one command `kubectl apply -f  xxx.yaml`. 

Helm is complicated, it requires users to install helm and tiller first.

xieydd:
what is the version of helm ? Tiller of helm v3 is gone, how about maintaining xxx.yaml and helm repo.

asifdxtreme:
With the current master branch code I am getting this error for `jobs.batch.volcano.sh`
```
kubectl create -f batch_v1alpha1_job.yaml 
The CustomResourceDefinition "jobs.batch.volcano.sh" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[plugins].additionalProperties: Forbidden: additionalProperties cannot be set to false
* spec.validation.openAPIV3Schema.properties[status].properties[ControlledResources].additionalProperties: Forbidden: additionalProperties cannot be set to false

```

hzxuzhonghu:
@asifdxtreme What's your k8s version?

asifdxtreme:
> @asifdxtreme What's your k8s version?

```
kubectl version --short
Client Version: v1.15.0
Server Version: v1.10.0
[...]

PR #327, opened 2019-07-11
Update README

thandayuthapani:
Add steps to download helm and setup helm with appropriate serviceaccount

For Issue: https://github.com/volcano-sh/volcano/issues/310
https://github.com/volcano-sh/volcano/issues/310#issuecomment-509873456

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,33 @@ You can watch industry experts talking about Volcano in different International 
 
 The easiest way to deploy Volcano is to use the Helm chart.  Volcano can be deployed by cloning code and also by adding helm repo.
 
-## Using Volcano Helm Repo
+## Install Helm
+
+### Download Helm
+
+Install helm by following official guide - https://github.com/helm/helm/blob/master/docs/install.md
+
+### Setup Helm
+
+Since Tiller needs cluster-admin RBAC access, create `ServiceAccount` and  `ClusterRoleBinding` by following the offical guide - https://github.com/helm/helm/blob/master/docs/rbac.md
+
+
+Create `ServiceAccount` and `ClusterRoleBinding` for tiller
+```bash
+  kubectl create serviceaccount --namespace kube-system tiller
+  kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
+``` 
+
+Then `helm init` using created `ServiceAccount`
+
+```bash
+  helm init --service-account <ServiceAccountName> --kubeconfig ${KUBECONFIG} --wait
+  
+  e.g :
+  helm init --service-account tiller --kubeconfig ${KUBECONFIG} --wait
+```
+
+## 1. Using Volcano Helm Repo
 
 Add helm repo using following command,
 
@@ -80,7 +106,7 @@ e.g :
 helm install volcano/volcano --namespace volcano-trial --name volcano-trial
 ```
 
-## Cloning Code
+## 2. Cloning Code
 ### Pre-requisites
 
 First of all, clone the repo to your local path:
@@ -88,7 +114,7 @@ First of all, clone the repo to your local path:
 ```
 # mkdir -p $GOPATH/src/volcano.sh/
 # cd $GOPATH/src/volcano.sh/
-# git clone --recursive https://github.com/volcano-sh/volcano.git
+# git clone https://github.com/volcano-sh/volcano.git
 ```
 
 ### 1. Volcano Image


README.md
```diff
@@ -63,7 +63,30 @@ You can watch industry experts talking about Volcano in different International
 
 The easiest way to deploy Volcano is to use the Helm chart.  Volcano can be deployed by cloning code and also by adding helm repo.
 
-## Using Volcano Helm Repo
+## Install Helm
```
hzxuzhonghu: how about make this under quick start guide and change it to Install Volcano?

hzxuzhonghu:
Can you also remove the `### 2. Helm charts` under `## Cloning Code`, because the `installer/helm/chart/volcano` does not exist at all.

thandayuthapani:
> Can you also remove the `### 2. Helm charts` under `## Cloning Code`, because the `installer/helm/chart/volcano` does not exist at all.

Since volcano-sh/charts is added as submodule, we will have to clone using --recursive tag, then only we can find those files in installer/helm

README.md
TommyLike: Why not we provide the helm document [link](https://github.com/helm/helm/blob/master/docs/install.md) here and notes about the RBAC issues.

thandayuthapani: @TommyLike Sure, will change that

TommyLike:
> Can you also remove the `### 2. Helm charts` under `## Cloning Code`, because the `installer/helm/chart/volcano` does not exist at all.

It's a submodule

README.md
thandayuthapani: @TommyLike Done

README.md
```diff
@@ -63,7 +63,23 @@ You can watch industry experts talking about Volcano in different International
...
+
+### Download Helm
+
+Install helm by following official guide - https://github.com/helm/helm/blob/master/docs/install.md
```
hzxuzhonghu: I think it is too complicated, as a newer, he would not want to install  `A`, and then `B`, before actually install volcano, i would prefer moving this to a simple scripts.

No one like complicated thing. If possible, can we remove dependency on helm totally？ @k82cn @TommyLike

TommyLike: @hzxuzhonghu maybe you can bring the practice from istio?

k82cn: I'm ok with that. IMO, it's complex because of security, e.g. RBAC; maybe we can provide a quick guidance which ignore security to make it simpler.

hzxuzhonghu: FYI:
istio provides a generated istio-demo.yaml, which allows users to install easily(https://istio.io/docs/setup/kubernetes/install/kubernetes/).  And also provides helm install https://istio.io/docs/setup/kubernetes/install/helm/.

And when it comes to kubernetes, we can run ./hack/local-ip-cluster.sh to start a cluster for developing and testing.

hzxuzhonghu:
/lgtm

Since this is a bug fix, should be in ASAP. But we should make a install improvement later @thandayuthapani .

k82cn:
Please update README according to https://github.com/volcano-sh/volcano/pull/334 :)

TommyLike:
According to the latest response from end user, I guess we need to add some prerequisites for installing volcano in order to reduce confusion. For example, from what I have tested on my environments, we have this requirement **at least**:
```
# Prerequisites
Kubernetes 1.12 or newer cluster.
```
**Why**: Since we enabled subresource in our CRDs, end user will have this issue when using elder version: 
**ISSUE**: https://github.com/kubernetes/kubernetes/pull/65357
**NOTES**: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#api-changes  (65357)

Free feel to add more prerequisites

@hzxuzhonghu @k82cn

hex108:
@TommyLike CRD status could be enabled manually since v1.10(https://blog.openshift.com/kubernetes-custom-resources-grow-up-in-v1-10/).

hzxuzhonghu:
Also golang version, should be at least 1.11+, because we use go module.

TommyLike:
> @TommyLike CRD status could be enabled manually since v1.10(https://blog.openshift.com/kubernetes-custom-resources-grow-up-in-v1-10/).

Hmm, so that is another requirement since we need the feature on?

thandayuthapani:
> Please update README according to #334 :)

Updated

README.md
k82cn: it's better to highlight chart version/branch if using volcano/chart. We're going to update master when releasing.

README.md
```diff
@@ -63,7 +63,23 @@ You can watch industry experts talking about Volcano in different International
...
+
+### Setup Helm
+
+Since Tiller needs cluster-admin RBAC access, create `ServiceAccount` and  `ClusterRoleBinding` by following the offical guide - https://github.com/helm/helm/blob/master/docs/rbac.md 
+
```
k82cn: let's give an example here; I don-t think a new user will go through the whole doc of rbac for volcano installation.

thandayuthapani: Sure! Will add step to add serviceaccount creation and such

thandayuthapani: @k82cn Have added steps to create serviceaccount and clusterrolebinding for tiller in README

k82cn: hm... if the following steps about serviceaccount is the correct solution, why do we need helm rbac installation guidance? If it's only an example, please highlight that.

hzxuzhonghu:
I think with #344, the install guide in this readme can be removed, and just add a link to ` installer/README.md `

k82cn:
> I think with #344, the install guide in this readme can be removed, and just add a link to installer/README.md

so where's the yaml file?

hzxuzhonghu:
The yaml generate should be a step, it is with default options and new  images built.

TommyLike:
Since we have different concerns on this section, I drafted the install section with the latest process of installing. @k82cn @thandayuthapani @hzxuzhonghu  please take a look and share your idea.

===============================================================
## Quick Start Guide

### Prerequisites

- Kubernetes 1.12+ with CRD support
- (Optional) Helm is needed if you'd like use volcano via helm charts. Please follow official
  [guide](https://github.com/helm/helm/blob/master/docs/install.md) to complete installation **<Need Metion RBAC issue here?>**

### Install release from github
The easiest way to install volcano is to use release yaml, please refer to the **README** file inside release folder after
download from [website](https://github.com/volcano-sh/volcano/releases)

### Install release from helm repo
We also support helm chart, install with the command below
```$xslt
# Add volcano offical helm chart
helm repo add volcano https://volcano-sh.github.io/charts
# Install a release wanted
helm install volcano/volcano --namespace <namespace> --name <name>  --version  <release-version>
```

**NOTE**:
Since tiller server will not manage CRD resource, therefore try install command with `--no-crd-hook` option when reinstalling.

### Install from source code

1. Clone the repo to your local path:

```
# mkdir -p $GOPATH/src/volcano.sh/
# cd $GOPATH/src/volcano.sh/
# git clone --recursive https://github.com/volcano-sh/volcano.git
```

2. Build docker images and load them into your k8s cluster

```
cd $GOPATH/src/volcano.sh/volcano
make images

## Verify your images
# docker images
REPOSITORY                      TAG                 IMAGE ID            CREATED             SIZE
volcanosh/vc-admission          latest              a83338506638        8 seconds ago       41.4MB
volcanosh/vc-scheduler          latest              faa3c2a25ac3        9 seconds ago       49.6MB
volcanosh/vc-controllers        latest              7b11606ebfb8        10 seconds ago      44.2MB

```

3. generate the yaml file and install
```$xslt
make generate-yaml    
# The generated yaml file would be located in _output/release/volcano-latest.yaml
kubectl apply -f _output/release/volcano-latest.yaml
# Create default queue resource after installation.
kubectl apply -f installer/helm/chart/volcano/templates/default-queue.yaml
```

### Install verify

To verify your installation run the following commands:

```
#1. Verify the Running Pods
# kubectl get pods --namespace <namespace>
NAME                                                 READY   STATUS    RESTARTS   AGE
<specified-name>-admission-84fd9b9dd8-9trxn          1/1     Running   0          43s
<specified-name>-controllers-75dcc8ff89-42v6r        1/1     Running   0          43s
<specified-name>-scheduler-b94cdb867-89pm2           1/1     Running   0          43s
<specified-name>--admission-init-qbtmb               0/1     Completed 0          43s

#2. Verify the Services
# kubectl get services --namespace <namespace>
NAME                                     TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
<specified-name>-admission-service       ClusterIP   10.105.78.53   <none>        443/TCP   91s

```

hzxuzhonghu:
As install guide is very long, i suggest moving it to [docs repo](https://github.com/volcano-sh/website) and link it here.
If we put a very long install procedure here, makes users feel bad.

k82cn:
@TommyLike @hzxuzhonghu , can you summary current status about installation? I'm lost in your PR and discussion, and I don-t know when it'll be finished :(

thandayuthapani:
Since PR https://github.com/volcano-sh/volcano/pull/346 has been merged, this PR can be closed, @k82cn Please close this PR when discussion this PR if finished.

Issue #329, opened 2019-07-11
Move admission webhook registeration into source code from yaml


TommyLike:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug

 /kind feature

**What happened**:
Now we have separated yaml file to register admission webhook, we should move it into source code which keeps maintaince easier.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hzxuzhonghu:
agree， moves it to `/installer `

TommyLike:
> agree， moves it to `/installer `

I mean use code to register webhookconfiguration resource when starting admission service.

hzxuzhonghu:
IC, SGTM

Issue #325, opened 2019-07-11
Contributor experience improvement


hzxuzhonghu:
In order to improve the contributor experience, we need many more docs:

- [ ] contributor guide: about how to contribute to valocano

- [ ] development guide: how to setup a development environment

- [ ] Testing: How to run kinds of tests, include ut, e2e.

- [ ] Code Conventions: coding style advice

TommyLike:
ADD: volcano usecase document(Maybe add document for our sample document?)

TommyLike:
> 2. volcano install improvement

/assign myself

k82cn:
> volcano usecase document (Maybe add document for our sample document?)

Two improvements is necessary: contributor experience and user experience; that's more about user experiences. Both of them are important, but let's open another issue for user experiences.

TommyLike:
> > volcano usecase document (Maybe add document for our sample document?)
> 
> Two improvements are necessary: contributor experience and user experience; that's more about user experiences. Both of them are important, but let's open another issue for user experiences.

# 335

PR #332, opened 2019-07-12
Register webhook in codes

TommyLike:
For Issue #329, Within this change, admission webhook configs as below are registered in admission codes automatically. Therefore this yaml can be removed.
```
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingWebhookConfiguration
metadata:
  name: {{ .Release.Name }}-validate-job
  annotations:
    "helm.sh/hook": pre-install
    "helm.sh/hook-delete-policy": before-hook-creation
webhooks:
  - clientConfig:
      service:
        name: {{ .Release.Name }}-admission-service
        namespace: {{ .Release.Namespace }}
        path: /jobs
    failurePolicy: Ignore
    name: validatejob.volcano.sh
    namespaceSelector: {}
    rules:
      - apiGroups:
          - "batch.volcano.sh"
        apiVersions:
          - "v1alpha1"
        operations:
          - CREATE
          - UPDATE
        resources:
          - jobs
---
apiVersion: admissionregistration.k8s.io/v1beta1
metadata:
  name: {{ .Release.Name }}-mutate-job
  annotations:
    "helm.sh/hook": pre-install
    "helm.sh/hook-delete-policy": before-hook-creation
webhooks:
  - clientConfig:
      service:
        name: {{ .Release.Name }}-admission-service
        namespace: {{ .Release.Namespace }}
        path: /mutating-jobs
    failurePolicy: Ignore
    name: mutatejob.volcano.sh
    namespaceSelector: {}
    rules:
      - apiGroups:
          - "batch.volcano.sh"
        apiVersions:
          - "v1alpha1"
        operations:
          - CREATE
        resources:
          - jobs
```
Also, this two configurations have been added the owner reference of admission service, which makes them deleted automatically when purge helm chart.


TommyLike:
This CI would fail due to the unchange of helm chart.

k82cn:
@TommyLike @hzxuzhonghu @asifdxtreme , can you help to list the items we need to do to improve contributor experience for this? That's really hard to get the progress of that if we open PR one by one :(

TommyLike:
> @TommyLike @hzxuzhonghu @asifdxtreme , can you help to list the items we need to do to improve contributor experience for this? That's really hard to get the progress of that if we open PR one by one :(

hmm, @k82cn  this patch is not related to contributor experience improvement.  @hzxuzhonghu has opened an issue which lists most of the things we need to do to improve it. We can discuss and talk on that one。https://github.com/volcano-sh/volcano/issues/325

k82cn:
>  this patch is not related to contributor experience improvement.

If contributor can not install Volcano, how to contribute?

TommyLike:
> > this patch is not related to contributor experience improvement.
> 
> If contributor can not install Volcano, how to contribute?

There would be several reasons that leads to failing installing volcano. 
1. kubernetes cluster version requirement.
2. misconfiguration of the kubernetes cluster
3. helm binary version requirement.
4. inconsistency between our charts templates and controller images.
5. .....


This patch intends to reduce the yaml amount and does not belong to any of these above.

cmd/admission/app/configure/configure.go
```diff
@@ -60,17 +62,30 @@ func (c *Config) AddFlags() {
...
 	flag.IntVar(&c.Port, "port", 443, "the port used by admission-controller-server.")
-	flag.StringVar(&c.MutateWebhookConfigName, "mutate-webhook-config-name", "volcano-mutate-job",
-		"Name of the mutatingwebhookconfiguration resource in Kubernetes.")
-	flag.StringVar(&c.MutateWebhookName, "mutate-webhook-name", "mutatejob.volcano.sh",
-		"Name of the webhook entry in the webhook config.")
-	flag.StringVar(&c.ValidateWebhookConfigName, "validate-webhook-config-name", "volcano-validate-job",
-		"Name of the mutatingwebhookconfiguration resource in Kubernetes.")
-	flag.StringVar(&c.ValidateWebhookName, "validate-webhook-name", "validatejob.volcano.sh",
-		"Name of the webhook entry in the webhook config.")
+	flag.StringVar(&c.MutateWebhookConfigName, "mutate-webhook-config-name", "",
```
hzxuzhonghu: I remember there is a method called `MarkDeprecated` , deprecate explicitly.

cmd/admission/app/configure/configure.go
```diff
@@ -79,78 +94,146 @@ func (c *Config) CheckPortOrDie() error {
...
+		return configured
 	}
-	prev, err := json.Marshal(config)
+	return generated
+}
+
+// RegiesterWebhooks register webhooks for admission service
+func RegiesterWebhooks(c *Config, clienset *kubernetes.Clientset, cabundle []byte) error {
+	// Get the admission service
+	admissionService, err := clienset.CoreV1().Services(
```
hzxuzhonghu: I would suggest removing this dependency, think if admission process running here, but no service is created for it.

cmd/admission/app/configure/configure.go
```diff
@@ -79,78 +94,146 @@ func (c *Config) CheckPortOrDie() error {
...
+					Path:      &path,
+				},
+				CABundle: cabundle,
+			},
+			FailurePolicy: &ignorePolicy,
+		}},
 	}
-	patch, err := strategicpatch.CreateTwoWayMergePatch(prev, curr, v1beta1.MutatingWebhookConfiguration{})
-	if err != nil {
+	JobValidateHooks.OwnerReferences = append(JobValidateHooks.OwnerReferences, *serviceRef)
```
hzxuzhonghu: No need setting this.

cmd/admission/app/configure/configure.go
```diff
@@ -79,78 +94,146 @@ func (c *Config) CheckPortOrDie() error {
...
-// PatchValidateWebhookConfig patches a CA bundle into the specified webhook config.
-func PatchValidateWebhookConfig(client admissionregistrationv1beta1client.ValidatingWebhookConfigurationInterface,
-	webhookConfigName, webhookName string, caBundle []byte) error {
-	config, err := client.Get(webhookConfigName, metav1.GetOptions{})
-	if err != nil {
-		return err
-	}
-	prev, err := json.Marshal(config)
-	if err != nil {
+	JobMutateHooks.OwnerReferences = append(JobMutateHooks.OwnerReferences, *serviceRef)
```
hzxuzhonghu: ditto

cmd/admission/main.go
```diff
@@ -52,39 +54,53 @@ func main() {
 	http.HandleFunc(admissioncontroller.MutateJobPath, serveMutateJobs)
 
 	if err := config.CheckPortOrDie(); err != nil {
-		fmt.Fprintf(os.Stderr, "%v\n", err)
+		glog.Fatalf("Configured port is invalid: %v\n", err)
```
hzxuzhonghu: The below os.Exit is redundant

cmd/admission/app/configure/configure.go
TommyLike: It's in spf13 package, can it be used here?

cmd/admission/app/configure/configure.go
TommyLike: The webhook configuration needs work with its backed service. see [here](https://github.com/volcano-sh/volcano/blob/789f95002e6f6f9e887414765e8d8e1cf1c20b90/installer/helm/chart/volcano/templates/admission-config.yaml#L11), or maybe we need waiting sometime until its get exist(ready)?

cmd/admission/app/configure/configure.go
TommyLike: for the purpose of deletion within admission service.

cmd/admission/app/configure/configure.go
hzxuzhonghu: lol

cmd/admission/main.go
TommyLike: why the none zero exit code not need here?

cmd/admission/app/configure/configure.go
hzxuzhonghu: I know that it need the service. Doesnot matter, without service, the call will fail. Once service comes up, it can succeed.

cmd/admission/app/configure/configure.go
hzxuzhonghu: When service is deleted, the configuration is also deleted, But when service is created, the configuration is not created auto.

cmd/admission/main.go
hzxuzhonghu: glog.Fatalf print a msg and exit?

TommyLike: really, where is the code?

hzxuzhonghu:
```
	if s == fatalLog {
		// If we got here via Exit rather than Fatal, print no stacks.
		if atomic.LoadUint32(&fatalNoStacks) > 0 {
			l.mu.Unlock()
			timeoutFlush(10 * time.Second)
			os.Exit(1)
		}
		// Dump all goroutine stacks before exiting.
		// First, make sure we see the trace for the current goroutine on standard error.
		// If -logtostderr has been specified, the loop below will do that anyway
		// as the first stack in the full dump.
		if !l.toStderr {
			os.Stderr.Write(stacks(false))
		}
		// Write the stack trace for all goroutines to the files.
		trace := stacks(true)
		logExitFunc = func(error) {} // If we get a write error, we'll still exit below.
		for log := fatalLog; log >= infoLog; log-- {
			if f := l.file[log]; f != nil { // Can be nil if -logtostderr is set.
				f.Write(trace)
			}
		}
		l.mu.Unlock()
		timeoutFlush(10 * time.Second)
		os.Exit(255) // C++ uses -1, which is silly because it's anded with 255 anyway.
	}
```

hzxuzhonghu: github.com/golang/glog/glog.go

TommyLike: That's great!

PR #345, opened 2019-07-17
Add one click install script

hzxuzhonghu:
Added local-up-volcano.sh, developers can 
Install kind cluster and then install volcano by 
``` bash
$ ./hack/local-up-volcano.sh
```

Install  install volcano into a existing cluster
``` bash
$ export INSTALL_MODE="xx" // set this env to a value not equals kind
$ ./hack/local-up-volcano.sh
```


hack/local-up-volcano.sh
```diff
@@ -0,0 +1,137 @@
...
+  echo kind load docker-image ${IMAGE_PREFIX}-controllers:${TAG}  ${CLUSTER_CONTEXT}
+  kind load docker-image ${IMAGE_PREFIX}-scheduler:${TAG}  ${CLUSTER_CONTEXT}
+  kind load docker-image ${IMAGE_PREFIX}-admission:${TAG}  ${CLUSTER_CONTEXT}
+}
+
+function install-volcano {
+  # TODO: add a graceful way waiting for all crd ready
+  kubectl apply -f ${RELEASE_FOLDER}/${YAML_FILENAME} --kubeconfig ${KUBECONFIG}
+  sleep 5
+  kubectl apply -f ${VK_ROOT}/installer/helm/chart/volcano/templates/default-queue.yaml --kubeconfig ${KUBECONFIG}
```
k82cn: just have a retry should be enough :)

TommyLike:
@hzxuzhonghu  It's better to update/add document along with this change, otherwise we need another one after this.

hack/local-up-volcano.sh
```diff
@@ -0,0 +1,137 @@
...
+if [[ $? -eq 0 ]]; then
+  cleanup
+  exit 0
+fi
+
+check-prerequisites
+
+prepare
+
+if [ "${INSTALL_MODE}" == "kind" ]; then
```
TommyLike: For the clusters other than kind cluster, how can user deploy the images into the cluster?

hzxuzhonghu: good question: we can not know their registry. If they use one node development env, the image is on the host. But if it is a multi-node cluster, we can not do that for users.

hack/local-up-volcano.sh
hzxuzhonghu: I want to make the default-queue install into the scheduler if possible.

hzxuzhonghu:
> It's better to update/add document along with this change, otherwise we need another one after this.

Yes, but i am not sure where to put it. Any suggestion?

hack/local-up-volcano.sh
TommyLike: Need to mention this in the document or just support kind cluster, since for the other cases, they can deploy volcano via the installation guide.

TommyLike:
> > It's better to update/add document along with this change, otherwise we need another one after this.
> 
> Yes, but i am not sure where to put it. Any suggestion?

Add a development section or individual document would be ok.

hack/local-up-volcano.sh
hzxuzhonghu: SG

hzxuzhonghu:
>I can see some functions repeated those in run-e2e-kind.sh, better move them to a common file to reduce duplicated code.

Will do

hzxuzhonghu:
> @hzxuzhonghu It's better to update/add document along with this change, otherwise we need another one after this.


@SrinivasChilveri Could you help

hzxuzhonghu:
ping @TommyLike ptal

pkg/scheduler/cache/cache.go
```diff
@@ -258,6 +260,19 @@ func newSchedulerCache(config *rest.Config, schedulerName string, defaultQueue s
...
+	// create default queue
+	defaultQue := v1alpha2.Queue{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: defaultQueue,
+		},
+		Spec: v1alpha2.QueueSpec{
+			Weight: 1,
+		},
+	}
+	if _, err := kbClient.SchedulingV1alpha2().Queues().Create(&defaultQue); err != nil && !apierrors.IsAlreadyExists(err) {
```
TommyLike: What if I have a existing default-queue from the previous version of volcano?

TommyLike:
> ping @TommyLike ptal

https://github.com/volcano-sh/volcano/blob/master/hack/publish.sh#L52 and https://github.com/volcano-sh/volcano/blob/master/installer/README.md both need update since the default-queue is not required now.

pkg/scheduler/cache/cache.go
hzxuzhonghu: Does not matter

hzxuzhonghu:
This can be a follow up.

PR #355, opened 2019-07-18
Fix preemption errors and add e2e case

### LINKED ISSUE #354: Fix the queue jobs preemption error

> /kind bug

**What happened**:
The error is frequently reported when using job preemption in some cases.

Error should not happen when using job preemption feature. The message is like:

```
E0712 11:23:32.825789       1 event_handlers.go:220] Failed to update pod qj-1-8w2ml in cache: errors:  1: failed to find task <default/qj-1-8w2ml> on host <test-virtualbox>
E0712 11:23:32.829026       1 event_handlers.go:252] Failed to delete pod qj-1-8w2ml from cache: errors:  1: failed to find task <default/qj-1-8w2ml> in job <default/qj-1>, 2: failed to find task <default/qj-1-8w2ml> on host <test-virtualbox>
```

1. enable preempt action.
2. submit a first job with 4 tasks.  Make the job occupy all cluster resource.
3. submit the second job with 4 tasks to preempt the first job.
The cause is that the succeeded status task is not treated in DeletePod function.

**Environment**:
- Volcano Version: master

william-wang:
The PR is target to fix the frequently errors when use queue job preemption feature.
The error message is like:
```
E0712 11:23:32.829026 1 event_handlers.go:252] Failed to delete pod qj-1-8w2ml from cache: errors: 1: failed to find task <default/qj-1-8w2ml> in job <default/qj-1>, 2: failed to find task <default/qj-1-8w2ml> on host
```

Which issue(s) this PR fixes (optional, in fixes #(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #354

The cause is that the succeeded status task is not treated correctlly in DeletePod function.


pkg/scheduler/cache/event_handlers.go
```diff
@@ -145,7 +145,7 @@ func (sc *SchedulerCache) deleteTask(pi *kbapi.TaskInfo) error {
 
 	if len(pi.NodeName) != 0 {
 		node := sc.Nodes[pi.NodeName]
-		if node != nil {
+		if node != nil && !isTerminated(pi.Status) {
```
k82cn: what happen if task finished?

hzxuzhonghu:
It seems there is some deep error, `RemoveTask` is called in many places, i am not sure whether it is right. Need more evidence.

But the more simple and direct way is not to return error when task not found

```
// RemoveTask used to remove a task from nodeInfo object
func (ni *NodeInfo) RemoveTask(ti *TaskInfo) error {
	key := PodKey(ti.Pod)

	task, found := ni.Tasks[key]
	if !found {
		return fmt.Errorf("failed to find task <%v/%v> on host <%v>",
			ti.Namespace, ti.Name, ni.Name)
	}
```

pkg/scheduler/cache/event_handlers.go
william-wang: I used the nginx images in containers. It should never exit as expected. Currently I do not know why it exit and the pod enter into Succeeded phase. I will use another kind of image to simulate the complete successfully.

william-wang:
> It seems there is some deep error, `RemoveTask` is called in many places, i am not sure whether it is right. Need more evidence.
> 
> But the more simple and direct way is not to return error when task not found
Yes,  RemoveTask is called in many places. Some debug logs was added and found this issue. 
According to following logic, when one evicted task is Succeeded(Terminated), the UpdatePod and DeletePod will be called one by one.  There is logical problem here.
**UpdatePod**
1. removeTask
2. addTask
     if **!isTerminated**(pi.Status) {
	     return node.AddTask(pi)
     }
     --> The Succeeded(Terminated) task is not added to node.
**DeletePod**
1. removeTask  
    if node != nil  {
	 nodeErr = node.RemoveTask(pi)
    }
   -->The removeTask should not be called because the Succeeded(Terminated) task is not on node at all.

pkg/scheduler/cache/event_handlers.go
william-wang: Had a test. 
If the task/pod complete successfully, there is not `DeletePod` event. All thing goes well.
If the task/pod is evicted, there is `DeletePod` event. However the task state is `DeletePod/UpdatePod` is releasing. All thing goes well too.
Only with nginx, the Succeded state happens and leads to this issue.

hzxuzhonghu:
good catch

hzxuzhonghu:
Do you mind analyzing the other call sites at the same way？
```
pkg/scheduler/api/node_info.go:211:// RemoveTask used to remove a task from nodeInfo object
pkg/scheduler/api/node_info.go:212:func (ni *NodeInfo) RemoveTask(ti *TaskInfo) error {
pkg/scheduler/api/node_info.go:242:	if err := ni.RemoveTask(ti); err != nil {
pkg/scheduler/framework/statement.go:176:		if err := node.RemoveTask(task); err != nil {
pkg/scheduler/framework/statement.go:295:		node.RemoveTask(task)
pkg/scheduler/cache/event_handlers.go:149:			nodeErr = node.RemoveTask(pi)

```

william-wang:
> Do you mind analyzing the other call sites at the same way？
> 
Not at all. I just went through all these places carefully. There is not any thing missing logical.

hzxuzhonghu:
As a following, i prefer removing the return error from `NodeInfo.RemoveTask`. If the task not exists, it must be something wrong else where. Just warning here is enough.

william-wang:
> As a following, i prefer removing the return error from `NodeInfo.RemoveTask`. If the task not exists, it must be something wrong else where. Just warning here is enough.

It make sensse. like below, right?

if !found {
	glog.Warningf("failed to find task <%v/%v> on host <%v>",
		ti.Namespace, ti.Name, ni.Name)
	return nil
}

hzxuzhonghu:
Yes @william-wang

hzxuzhonghu:
@william-wang on your own branch, run `git pull --rebase upstream master`.

william-wang:
> @william-wang on your own branch, run git pull --rebase upstream master.

@hzxuzhonghu 
Seems there is some isssue with git.
After I exectued "git pull --rebase upstream master",  It prompts there is some conflict.
I fixed the confliction and execute "git add <>" and "git rebase --continue". Following message came out.

$ git rebase --continue
Applying: Fix the build issue
No changes - did you forget to use 'git add'?
If there is nothing left to stage, chances are that something else
already introduced the same changes; you might want to skip this patch.
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".

pkg/scheduler/cache/event_handlers.go
william-wang: @k82cn , please help to review and approve. Thanks.

k82cn: why 82 commits?

william-wang: The 82 commits includes many other member's commits after using 'pull --rebase'.

william-wang: @k82cn any more comments ?

k82cn: please remove other commits.

william-wang: @k82cn Other comments is removed.

k82cn: still 22 commits for 3 files?

pkg/scheduler/api/node_info.go
```diff
@@ -214,8 +214,9 @@ func (ni *NodeInfo) RemoveTask(ti *TaskInfo) error {
 
 	task, found := ni.Tasks[key]
 	if !found {
-		return fmt.Errorf("failed to find task <%v/%v> on host <%v>",
+		glog.Warningf("failed to find task <%v/%v> on host <%v>",
```
k82cn: Why change that? It should be error here.

pkg/scheduler/cache/event_handlers.go
k82cn: > If the task/pod is evicted, there is DeletePod event. However the task state in DeletePod/UpdatePod is releasing. 

What's happen if no grace period?

pkg/scheduler/api/node_info.go
william-wang: Here is hzxuzhonghu's comments. 
> As a following, i prefer removing the return error from `NodeInfo.RemoveTask`. If the task not exists, it must be something wrong else where. Just warning here is enough.

pkg/scheduler/cache/event_handlers.go
william-wang: There is no issue if no grace perod.

william-wang: > still 22 commits for 3 files?
The redundant commit is removed.

k82cn: > There is no issue if no grace perod.

Why?

william-wang: If grace period is set 0, the pod is deleted inmediately and only has DeletePod event.  So the task must be on the node when calling DeletePod func.
If grace period is not set 0, there are two events including UpdatePod and DeletePod. If the task is removed from node in UpdatePod func, the task does not exist on node any more when calling DeletePod func.

pkg/scheduler/api/node_info.go
hzxuzhonghu: Yep, removing a non existing task does not make harm to the scheduler.

william-wang: @k82cn 

Should I need to change back to Errorf again? If no need, please help to approve. Thanks.

k82cn:
If it's terminated, we should only remove it once instead of check it every time.

william-wang:
> If it's terminated, we should only remove it once instead of check it every time.

Yes.  Now the behavior is as your said. If the task is terminated, the UpdatePod delete it once. When the Pod is deleted finally, the DeletePod find the task already not exist and no need delete any more.

Issue #244, opened 2019-06-21
Theme of next release


k82cn:
Volcano'd like to have another release around end of Aug; there're the proposed features that maybe included in next release, please share your comments/suggestion and if any other requirements, please also feel free to share :)

## Job Management

- [x] Delay pod creation improvement (https://github.com/volcano-sh/volcano/issues/135) (@wangyuqing4)

## Queue Management 

- [x] Lifecycle manage of Queue, e.g. Queue's status (@hzxuzhonghu )
- [ ] Improve Queue CRD (@hzxuzhonghu )

## Scheduler/Policy

- [ ] GPU enhancement (https://github.com/kubernetes-sigs/kube-batch/issues/236) (@TommyLike)
- [ ] Reservation to avoid starvation (https://github.com/volcano-sh/volcano/issues/13) (@sivanzcw)
- [x] Fair-sharing cross namespace (https://github.com/volcano-sh/volcano/issues/147) (@lminzhw)
- [x] Proportion enhancement for Queue, e.g. (https://github.com/kubernetes-sigs/kube-batch/issues/492) (@william-wang)
- [ ] Preemption/Reclaim enhancement (@william-wang)

## Stability

- [x] Improve code coverage of controller & scheduler ( @asifdxtreme )
- [ ] Improve e2e test ( @asifdxtreme )

k82cn:
/cc @hex108 @johnugeorge for Queue
/cc @richardsliu @gaocegege for delay pod creation
/cc @ahmad-diaa  for fair-sharing
/cc @wsxiaozhang @kinglion811 for GPU part

Please feel free to hands up if any items you'd like to contribute :)

/cc @jeefy @Jeffwan @mrbobbytables who may also interest; and please feel free to share your requirements/comments if any :)

jeefy:
No hard reqs yet, but couple thoughts:
- This looks like a lot of work, and it's exciting to think of all this getting rounded out.
- It's awesome if the plan is to get into a semi-regular release cadence.

k82cn:
> It's awesome if the plan is to get into a semi-regular release cadence.

Yes, we'd like to have a minor release every quarter, similar to k8s :)

hzxuzhonghu:
There are also many clean code work after `kube-batch` in tree. 

It mainly focus on:

1. the variable naming, like `kbClient` => `vcClient`

1. duplicate import, like `"volcano.sh/volcano/pkg/client/clientset/versioned"`

[...]

PR #358, opened 2019-07-19
support fair share

lminzhw:
Implementation of: https://github.com/volcano-sh/volcano/blob/master/docs/design/fairshare.md

support fair share based on weight among namespaces.

Part of: https://github.com/volcano-sh/volcano/issues/244


pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
...
+const (
+	// QuotaKey is the key in ResourceQuota.spec.hard indicating the weight of thi namespace
+	QuotaKey = "volcano.sh/namespace.weight"
+	// DefaultWeight is the default weight of namespace
+	DefaultWeight = 1
+)
+
+// NamespaceInfo records information of namespace
+type NamespaceInfo struct {
+	Name string
```
hzxuzhonghu: What's Name mean?

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
...
+// NewNamespaceCollection creates new NamespaceCollection object to record all information about a namespace
+func NewNamespaceCollection(name string) *NamespaceCollection {
+	n := &NamespaceCollection{
+		Name:        name,
+		quotaWeight: cache.NewHeap(quotaItemKeyFunc, quotaItemLessFunc),
+	}
+	// add at least one item into quotaWeight.
+	// Because cache.Heap.Pop would be blocked until queue is not empty
+	n.updateWeight(&quotaItem{
+		name:   QuotaKey,
```
hzxuzhonghu: Is this necessary, since it is a fixed value.

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
...
+	B := b.(*quotaItem)
+	return A.weight > B.weight
+}
+
+// NamespaceCollection will record all details about namespace
+type NamespaceCollection struct {
+	Name string
+
+	weightMu    sync.Mutex
+	quotaWeight *cache.Heap
```
hzxuzhonghu: To my understanding, only need to keep the biggest weight.

Even if you want all the weights, there is a `PriorityQueue` in volcano, does it satisfy your requirement?

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
...
+	A := a.(*quotaItem)
+	B := b.(*quotaItem)
+	return A.weight > B.weight
+}
+
+// NamespaceCollection will record all details about namespace
+type NamespaceCollection struct {
+	Name string
+
+	weightMu    sync.Mutex
```
hzxuzhonghu: donnot need the lock

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -43,7 +43,14 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 	defer glog.V(3).Infof("Leaving Allocate ...")
 
 	queues := util.NewPriorityQueue(ssn.QueueOrderFn)
+	// map[queueId]PriorityQueue(*api.JobInfo)
 	jobsMap := map[api.QueueID]*util.PriorityQueue{}
+	// map[queueId]PriorityQueue(namespaceName)
+	namespaceMap := map[api.QueueID]*util.PriorityQueue{}
+	// map[queueId]map[namespaceName]PriorityQueue(*api.JobInfo)
+	jobInNamespaceMap := map[api.QueueID]map[string]*util.PriorityQueue{}
```
hzxuzhonghu: Add more comments

pkg/scheduler/api/cluster_info.go
```diff
@@ -20,9 +20,10 @@ import "fmt"
 
 // ClusterInfo is a snapshot of cluster by cache.
 type ClusterInfo struct {
-	Jobs   map[JobID]*JobInfo
-	Nodes  map[string]*NodeInfo
-	Queues map[QueueID]*QueueInfo
+	Jobs          map[JobID]*JobInfo
+	Nodes         map[string]*NodeInfo
+	Queues        map[QueueID]*QueueInfo
+	NamespaceInfo map[string]*NamespaceInfo
```
hzxuzhonghu: Add comments key-->value

pkg/scheduler/cache/event_handlers.go
```diff
@@ -1129,3 +1129,87 @@ func (sc *SchedulerCache) addPriorityClass(pc *v1beta1.PriorityClass) {
...
+	sc.updateResourceQuota(newR)
+}
+
+// AddResourceQuota add ResourceQuota to scheduler cache
+func (sc *SchedulerCache) AddResourceQuota(obj interface{}) {
+	var r *v1.ResourceQuota
+	switch t := obj.(type) {
+	case *v1.ResourceQuota:
+		r = t
+	case cache.DeletedFinalStateUnknown:
```
hzxuzhonghu: donot need this in add

pkg/scheduler/cache/event_handlers.go
```diff
@@ -1129,3 +1129,87 @@ func (sc *SchedulerCache) addPriorityClass(pc *v1beta1.PriorityClass) {
...
+
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
+
+	glog.V(3).Infof("Update ResourceQuota <%s/%v> in cache, with spec: %v.", newR.Namespace, newR.Name, newR.Spec.Hard)
+	sc.updateResourceQuota(newR)
+}
+
+// AddResourceQuota add ResourceQuota to scheduler cache
+func (sc *SchedulerCache) AddResourceQuota(obj interface{}) {
```
hzxuzhonghu: I can not see where you filterout resource quota that does not set `vaolcano.sh/xx.weight`

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -62,12 +69,37 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
+
+		jobs, found := namespaceJob[job.Namespace]
+		if !found {
+			jobs = util.NewPriorityQueue(ssn.JobOrderFn)
+			namespaceJob[job.Namespace] = jobs
+
+			namespaceMap[job.Queue].Push(job.Namespace)
+		}
+
+		glog.V(4).Infof("Added Job <%s/%s> into Queue <%s> Namespace <%s>", job.Namespace, job.Name, job.Queue, job.Namespace)
```
hzxuzhonghu: nit: duplicate namespace

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -83,18 +89,66 @@ func (drf *drfPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			lNsAlloc := lNsAtt.allocated.Clone().Add(preemptor.Resreq)
+			_, lNsShare := drf.calculateShare(lNsAlloc, drf.totalResource)
+			lNsShareWeighted = lNsShare / float64(lWeight)
+		}
+
 		allocations := map[api.JobID]*api.Resource{}
+		namespaceAllocation := map[string]*api.Resource{}
 
 		for _, preemptee := range preemptees {
+			if namespaceOrderEnabled && preemptor.Namespace != preemptee.Namespace {
```
hzxuzhonghu: more comments

pkg/scheduler/api/namespace_info.go
lminzhw: `Name` is the name of this namespace, I'll add comments above it.

pkg/scheduler/api/namespace_info.go
lminzhw: Now, `name` of `quotaItem` is used in `quotaItemKeyFunc` and it is for `cache.Heap` to distinguish different quota resources. 

`PriorityQueue` in volcano may have another behavior. I'll have a try.

pkg/scheduler/api/namespace_info.go
lminzhw: The only biggest weight will be kept in `NamespaceInfo` after `cache.Snapshot()`.

At here, `NamespaceCollection` store all the information about this namespace and will be able to handle the weight changes in `Quota` resource (such as try to decrease the biggest weight).

I'll have a look at `PriorityQueue` later~

lminzhw: The `PriorityQueue` is a simple implementation of `container/heap`, but it could not support the `Update` action, and delete some item in heap directly is also not supported.

So, It it also the reason why I choose the Heap in k8s cache before.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -193,6 +253,10 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 		} else {
 			stmt.Discard()
 		}
+
+		if namespaceOrderEnabled {
```
TommyLike: We would enter the place that we have a huge **Execute** method that is much too long 😄

pkg/scheduler/api/namespace_info_test.go
```diff
@@ -0,0 +1,88 @@
+package api
```
TommyLike: Missing header here.

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
...
+// NamespaceInfo records information of namespace
+type NamespaceInfo struct {
+	Name string
+	// Weight is the highest weight among many ResourceQuota.
+	Weight int64
+}
+
+// GetWeight returns weight of a namespace, any invalid case would get default value
+func (n *NamespaceInfo) GetWeight() int64 {
+	if n == nil {
```
TommyLike: too if statements can be combined.

pkg/scheduler/api/namespace_info.go
lminzhw: The original idea for me to add the lock is to guarantee the actions in `Snapshot` should be atomic.

Now, I also think this lock is unnecessary.

I'll remove this lock.

pkg/scheduler/api/namespace_info.go
TommyLike: Keep them all would be ok to me.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -1129,3 +1129,87 @@ func (sc *SchedulerCache) addPriorityClass(pc *v1beta1.PriorityClass) {
...
+func (sc *SchedulerCache) DeleteResourceQuota(obj interface{}) {
+	var r *v1.ResourceQuota
+	switch t := obj.(type) {
+	case *v1.ResourceQuota:
+		r = t
+	case cache.DeletedFinalStateUnknown:
+		var ok bool
+		r, ok = t.Obj.(*v1.ResourceQuota)
+		if !ok {
+			glog.Errorf("Cannot convert to *v1beta1.PriorityClass: %v", t.Obj)
```
TommyLike: fix `v1beta1.PriorityClass`

pkg/scheduler/cache/event_handlers.go
```diff
@@ -1129,3 +1129,87 @@ func (sc *SchedulerCache) addPriorityClass(pc *v1beta1.PriorityClass) {
...
+			return
+		}
+	default:
+		glog.Errorf("Cannot convert to *v1beta1.PriorityClass: %v", t)
```
TommyLike: ditto

pkg/scheduler/cache/event_handlers.go
```diff
@@ -1129,3 +1129,87 @@ func (sc *SchedulerCache) addPriorityClass(pc *v1beta1.PriorityClass) {
...
+
+	glog.V(3).Infof("Delete ResourceQuota <%s/%v> in cache", r.Namespace, r.Name)
+	sc.deleteResourceQuota(r)
+}
+
+// UpdateResourceQuota update ResourceQuota to scheduler cache
+func (sc *SchedulerCache) UpdateResourceQuota(oldObj, newObj interface{}) {
+	newR, ok := newObj.(*v1.ResourceQuota)
+	if !ok {
+		glog.Errorf("Cannot convert newObj to *v1beta1.PriorityClass: %v", newObj)
```
TommyLike: ditto

pkg/scheduler/cache/event_handlers.go
TommyLike: The namespace without this label would default to 1

pkg/scheduler/framework/session_plugins.go
```diff
@@ -304,6 +309,50 @@ func (ssn *Session) JobOrderFn(l, r interface{}) bool {
...
+			if j := nof(l, r); j != 0 {
+				return j < 0
+			}
+		}
+	}
+
+	// TODO(lminzhw): if all NamespaceOrderFn treat these two namespace as the same,
+	// we should make the job order have its affect among namespaces.
+	// or just schedule namespace one by one
+	lv := l.(string)
```
TommyLike: Do we need to check if this converting failing?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -304,6 +309,50 @@ func (ssn *Session) JobOrderFn(l, r interface{}) bool {
...
+				continue
+			}
+			return isEnabled(plugin.EnabledNamespaceOrder)
+		}
+	}
+	return false
+}
+
+// NamespaceOrderFn invoke namespaceorder function of the plugins
+func (ssn *Session) NamespaceOrderFn(l, r interface{}) bool {
```
TommyLike: So the order of namespace is decided by the first plugin if it has enabled the namespace order?

pkg/scheduler/plugins/drf/drf.go
TommyLike: Better refactor this part. since the change is little too hardcoded.

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
k82cn: s/kubernetes/Volcano

k82cn: new files belong to volcano community

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	// QuotaKey is the key in ResourceQuota.spec.hard indicating the weight of thi namespace
+	QuotaKey = "volcano.sh/namespace.weight"
+	// DefaultWeight is the default weight of namespace
+	DefaultWeight = 1
```
k82cn: default namespace weight

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"github.com/golang/glog"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	// QuotaKey is the key in ResourceQuota.spec.hard indicating the weight of thi namespace
+	QuotaKey = "volcano.sh/namespace.weight"
```
k82cn: s/QuotaKey/NamespaceWeightKey

pkg/scheduler/framework/session_plugins.go
```diff
@@ -304,6 +309,50 @@ func (ssn *Session) JobOrderFn(l, r interface{}) bool {
...
 }
 
+// NamespaceOrderEnabled returns the NamespaceOrder is enabled in this session or not
+// judging by length of namespaceOrderFns
+func (ssn *Session) NamespaceOrderEnabled() bool {
+	return len(ssn.namespaceOrderFns) > 0
+}
+
+// NamespaceOrderEnabledInPlugin the NamespaceOrder for this plugin is enabled in this session or not
+func (ssn *Session) NamespaceOrderEnabledInPlugin(pluginName string) bool {
```
k82cn: move this to drf plugin; otherwise, there're lots of helper function in `Session`.

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -66,6 +70,8 @@ func (drf *drfPlugin) OnSessionOpen(ssn *framework.Session) {
 		drf.totalResource.Add(n.Allocatable)
 	}
 
+	namespaceOrderEnabled := ssn.NamespaceOrderEnabledInPlugin(drf.Name())
```
k82cn: is there anyway to remove `namespaceOrderEnabled`? That's error prone :)

pkg/scheduler/actions/allocate/allocate.go
lminzhw: add more comments to describe these two maps.

pkg/scheduler/api/cluster_info.go
lminzhw: add type `NamespaceName` to avoid too many comments

pkg/scheduler/cache/event_handlers.go
lminzhw: delete this case block

pkg/scheduler/cache/event_handlers.go
lminzhw: In `itemFromQuota` function, any `ResourceQuota` without `xx.weight` key would  have weight with default value 1

pkg/scheduler/actions/allocate/allocate.go
lminzhw: change duplicate namespace log to prefix `NamespaceEnabled: ` to indicate this add action is different to adding job directly into map belonging to queue.

pkg/scheduler/plugins/drf/drf.go
lminzhw: refactor this part, namespace order policy is now a separate code block out of job order policy.

pkg/scheduler/api/namespace_info_test.go
lminzhw: add headers before package

pkg/scheduler/api/namespace_info.go
lminzhw: combine these two statements

pkg/scheduler/cache/event_handlers.go
lminzhw: fix it using 'v1.ResourceQuota'

pkg/scheduler/cache/event_handlers.go
lminzhw: fixed

pkg/scheduler/cache/event_handlers.go
lminzhw: fixed

pkg/scheduler/framework/session_plugins.go
lminzhw: this actions following the code pattern in other `CompareFn`, like `taskOrder` or `jobOrder` in `priority` plugin.

pkg/scheduler/framework/session_plugins.go
lminzhw: Yes, if some plugin decide namespace A is prior to B, this decision is just the final result.

This logic follows the code pattern in other `CompareFn`, like `JobOrderFn`.

pkg/scheduler/api/namespace_info.go
lminzhw: fixed

pkg/scheduler/api/namespace_info.go
lminzhw: rename it to `DefaultNamespaceWeight`

pkg/scheduler/api/namespace_info.go
lminzhw: rename it

pkg/scheduler/framework/session_plugins.go
lminzhw: move it to `drf.NamespaceOrderEnabled`.

pkg/scheduler/plugins/drf/drf.go
lminzhw: I refactor the code in allocate, there is no `namespaceOrderEnabled` in allocation func now.

And, `namespaceOrderEnabled` in `drf` plugin is also refactored to be more readable.

pkg/scheduler/api/namespace_info.go
lminzhw: move its type from string to `NamespaceName`

pkg/scheduler/actions/allocate/allocate.go
lminzhw: after code refactoring, this log is deleted.

pkg/scheduler/actions/allocate/allocate.go
lminzhw: refactor action Allocate, remove all `namespaceOrderEnabled` in it.

lminzhw:
@hzxuzhonghu  @TommyLike  @k82cn 

TLDR: code and algorithm in `allocate.go` changes a lot.

---

In previous version, I use `PriorityQueue` to pick namespace from choosed queue.

But, when there is many queue, the changes of namespace priority by allocation in this queue would have no chance to affect  `PriorityQueue` in other queue.

So, I change the method to pick <namespace, queue> tuple.

---

And why I choose to pick namespace firstly but not queue?

we have two method to pick them:

- pick namespace firstly, with time complexity log(n) * q
- pick queue firstly, with time complexity log(q) * n

and the first picked one would have a more balanced resource usage.

In my opinion, the number of queue would less than namespace in most case, and the balanced resource usage of namespace should be considered first, so I choose to pick namespace firstly.

lminzhw:
> I think we need more tests, especially for namespace job allocate part.
> 
> Also i noticed a problem: The namespce may not get requested resources even when there is enough quota. Say we have queue1 and queue2, ns1 and ns2, all with weight 1. Jobs in ns1 (requires 50% of all resources) are using queue1 and queue2. Jobs in ns2(requires 50% all resources) are using queue2 only.
> 
> If queue order is queue2 ->queue1, then ns2 can only get 25% of the resources. Is this correct?

After reading code carefully, that is not a problem now, I think.

In `propotion` plugin, the queue with less share would be allocation firstly. And in `allocate` action, after every job allocation, the queue with less share would have change to allocate pod.

So, if ns1 and ns2 have enough pod count (for having enough scheduling loop), queue2 be scheduled firstly or not is not that concerned, because the next scheduling loop would choose to schedule another for its lower share.

In newly pushed version, ns1 be scheduled firstly would lead to schedule ns2 next and the resource is also balanced.

pkg/scheduler/api/namespace_info.go
hzxuzhonghu: IC,  this is to support resourcequota update

hzxuzhonghu:
LGTM, but i would expect e2e tests as well.

lminzhw:
> LGTM, but i would expect e2e tests as well.

We have 2 more e2e test cases and a XXL PR now.  : )

hzxuzhonghu:
@lminzhw Would suggest not squash all commits, it is hard to review.

test/e2e/job_scheduling.go
```diff
@@ -387,4 +392,196 @@ var _ = Describe("Job E2E Test", func() {
 		err = waitJobPhaseReady(context, job)
 		Expect(err).NotTo(HaveOccurred())
 	})
+
```
hzxuzhonghu: Better add a case with both queue and namespace fair share?

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,146 @@
...
+	"fmt"
+
+	"github.com/golang/glog"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+// NamespaceName is name of namespace
+type NamespaceName string
```
hex108: Would `Namespace` be other types other than `string`? If not, it might be better to use `string` directly. Then it avoids many type conversion

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -91,21 +111,47 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
 		}
 
-		queue := queues.Pop().(*api.QueueInfo)
-		if ssn.Overused(queue) {
-			glog.V(3).Infof("Queue <%s> is overused, ignore it.", queue.Name)
-			continue
+		// pick namespace from namespaces PriorityQueue
+		namespace := namespaces.Pop().(api.NamespaceName)
+
+		queueInNamespace := jobsMap[namespace]
```
hex108: A little confused. What is the relationship between queue and namespace? According to the doc [motivation](https://github.com/volcano-sh/volcano/blob/master/docs/design/fairshare.md#motivation), a namespace works like a user, and there might be many users(namespaces) in a queue. Will be many queues in a namespace?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -91,21 +111,47 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
...
+		// pick queue for given namespace
+		//
+		// This block use a algorithm with time complex O(n).
+		// But at least PriorityQueue could not be used here,
+		// because the allocation of job would change the priority of queue among all namespaces,
+		// and the PriorityQueue have no ability to update priority for a special queue.
+		var queue *api.QueueInfo
+		for queueId := range queueInNamespace {
+			currentQueue := ssn.Queues[queueId]
+			if ssn.Overused(currentQueue) {
```
hex108: If the queue is overused, could not the queue get resource even the cluster has idle resource?

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,146 @@
...
+func quotaItemKeyFunc(obj interface{}) (string, error) {
+	item, ok := obj.(*quotaItem)
+	if !ok {
+		return "", fmt.Errorf("obj with type %T could not parse", obj)
+	}
+	return item.name, nil
+}
+
+// for big root heap
+func quotaItemLessFunc(a interface{}, b interface{}) bool {
```
hex108: s/quotaItemLessFunc/quotaItemGreaterFunc/?

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -60,12 +64,27 @@ func (drf *drfPlugin) Name() string {
 	return PluginName
 }
 
+// NamespaceOrderEnabled returns the NamespaceOrder for this plugin is enabled in this session or not
+func (drf *drfPlugin) NamespaceOrderEnabled(ssn *framework.Session) bool {
```
hex108: What is the difference between it and `func (ssn *Session) NamespaceOrderEnabled()`?

pkg/scheduler/actions/allocate/allocate.go
Jeffwan: @hex108 As I understand,  generally speaking, different users (namespace) could schedule jobs in different Queues.  This doc I think concentrates more on the faire resource sharing in one queue between users. This solution is mainly trying to resolve starvation problem in resource sharing.

pkg/scheduler/api/namespace_info.go
```diff
@@ -0,0 +1,146 @@
...
+	// DefaultNamespaceWeight is the default weight of namespace
+	DefaultNamespaceWeight = 1
+)
+
+// NamespaceInfo records information of namespace
+type NamespaceInfo struct {
+	// Name is the name of this namespace
+	Name NamespaceName
+	// Weight is the highest weight among many ResourceQuota.
+	Weight int64
```
Jeffwan: if we use signed integer, what's the best practice here? Do we suggest users to always use weights > 0 or it could be negative numbers?

test/e2e/job_scheduling.go
lminzhw: This RP is too big now, I'll submit a new test next PR.

pkg/scheduler/api/namespace_info.go
lminzhw: I just want to make codes more readable, so choose to mark it as a new type but not mark it in comment.

pkg/scheduler/actions/allocate/allocate.go
lminzhw: Yes, the overused queue will be skipped in master code too.

pkg/scheduler/api/namespace_info.go
lminzhw: done~

pkg/scheduler/plugins/drf/drf.go
lminzhw: The `NamespaceOrderEnabled` in plugin just return the value of that plugin. This func return the result of this session and used in `allocate` action.

And in the newest version code, I delete this func and will open a Issue to adjust the behavior when the NamespaceOrder is disabled or other cases.

pkg/scheduler/api/namespace_info.go
lminzhw: Yes, weight < 1 is meaningless in this case, because we have a default value which is 1. Any weight < 1 will be ignored.

lminzhw:
@all, sorry for the delayed reply.

And this PR leaves two unsolved problem now:
1. add a case with both Queue and Namespace fair share
2. the behavior when disable NamespaceOrder is not so good, I will adjust it.

Waiting for your response~

k82cn:
/lgtm

help to re-label lgtm, and it LGTM to :)

PR #362, opened 2019-07-19
Some cleanups after kube-batch intree

hzxuzhonghu:


TommyLike:
scheduler config file is used to generate the configmap yaml. Therefore we can't delete them all.

hzxuzhonghu:
I remember i renamed it,  will have a look

hzxuzhonghu:
Added back

Issue #347, opened 2019-07-18
README document update follow ups


TommyLike:
Since we changed the guide how to install volcano now. We have to completes these things below:

- [x] Setup a cron task for Travis to build the latest images and push daily.
- [x] Add a Travis Job to verify the file `installer/volcano-development` is identical to the generated file via command `make generate-yaml`.
- [x] Add individual document to guide user how to install volcano via github releases (ref: travis release task) and helm repo.
- [x] Update the helm chart zip file to use the 0.1 image rather than the latest image. https://github.com/volcano-sh/charts/blob/master/package/volcano-0.1.tgz
- [x] Support release helm chart to [repo](https://github.com/volcano-sh/charts) when releasing in script automatically.

TommyLike:
@hzxuzhonghu  @k82cn  please add the todos if you think that is still missing.

@asifdxtreme thanks.

TommyLike:
@asifdxtreme do we created the cronjob for daily build? https://hub.docker.com/r/volcanosh/vc-controllers/tags

asifdxtreme:
![image](https://user-images.githubusercontent.com/9736066/61682193-4b9a1200-ad2e-11e9-94c1-0cf1df7f7a00.png)
it is schedlued to run in another 6 hours, last it was ran 18 hours ago(at that time the PR was not merged), I guess travis is using PST time to run the nightly builds

TommyLike:
@asifdxtreme thanks

asifdxtreme:
> @asifdxtreme thanks

![image](https://user-images.githubusercontent.com/9736066/61722006-1f5eaf80-ad87-11e9-8486-080d5a056e38.png)
The images are getting updated, it will usually update around 18:00 CST everyday

asifdxtreme:
> * [ ]  Support release helm chart to [repo](https://github.com/volcano-sh/charts) when releasing in script automatically.

@TommyLike IMO we should keep this as manual step, these changes might need a manual intervention to decide whether the changes needs to be upstreamed to volcano/charts

TommyLike:
> > * [ ]   Support release helm chart to [repo](https://github.com/volcano-sh/charts) when releasing in script automatically.
> 
> @TommyLike IMO we should keep this as manual step, these changes might need a manual intervention to decide whether the changes needs to be upstreamed to volcano/charts

I am ok with this. Will close this.

PR #372, opened 2019-07-22
Publish Docker Image in Daily Build

asifdxtreme:
Partially fixes https://github.com/volcano-sh/volcano/issues/347

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,14 @@ jobs:
           skip_cleanup: true
           on:
             tags: true
+    - stage: Publish latest Docker Image on Daily Build
+      script:
+        - echo "Publish latest Docker Images to DockerHub"
+      deploy:
+        - provider: script
+          script: make TAG=latest RELEASE_VER=latest release
+          on:
+            condition: $TRAVIS_EVENT_TYPE = cron
 notifications:
   webhooks: https://www.travisbuddy.com/
   on_success: never


.travis.yml
```diff
@@ -58,6 +58,16 @@ jobs:
...
           on:
             tags: true
+    - stage: Publish latest Docker Image on Daily Build
+      before_deploy:
+        - export TRAVIS_TAG=$(git describe --tags)
+      script:
+        - echo "Publish latest Docker Images to DockerHub"
+      deploy:
+        - provider: script
+          script: make TAG=${TRAVIS_TAG} RELEASE_VER=latest release
```
TommyLike: We need always generate the images with `latest` tag, since it's what we used in the all-in-one yaml.

asifdxtreme: got it..updated

TommyLike:
thanks @asifdxtreme  /lgtm

.travis.yml
```diff
@@ -58,6 +58,14 @@ jobs:
           skip_cleanup: true
           on:
             tags: true
+    - stage: Publish latest Docker Image on Daily Build
```
hzxuzhonghu: How can ci control it is run once a day?

asifdxtreme: `condition: $TRAVIS_EVENT_TYPE = cron` it is based on this condition
currently cron job is configured to run daily once

PR #373, opened 2019-07-22
Add Verify Job to check consistency between generate yaml and the yam…

asifdxtreme:
…l in the repo

https://github.com/volcano-sh/volcano/issues/347


hack/check-generated-yaml.sh
```diff
@@ -0,0 +1,33 @@
...
+
+if ! diff ${VK_ROOT}/installer/volcano-development.yaml ${RELEASE_FOLDER}/volcano-latest.yaml ; then
+	{
+		echo
+		echo "The Generated yaml is different from the one in installer/volcano-development.yaml"
+		echo "Suggest to sync both the files"
+		echo
+	} >&2
+	false
+fi
```
TommyLike: Missing newline here

asifdxtreme: updated

PR #380, opened 2019-07-23
support binpack policy

lminzhw:
## Motivation

While running workloads on kubernetes which use accelerator devices the default scheduler spreads pods across nodes resulting in fragmentation of extended resources. This fragmentation prevents scheduling pods with larger device resource requirements, and they remain in the pending state.

### Goals

 - Schedule Pods using BestFit Policy using RequestedToCapacityRatio Priority Function for extended Resources
- Reduce Fragmentation of scarce resources on the Cluster


lminzhw:
@hzxuzhonghu @TommyLike @k82cn 

Now, we have another XL PR. An e2e test case is included.

Please have a look on it.

pkg/scheduler/plugins/binpack/binpack.go
```diff
@@ -0,0 +1,230 @@
...
+
+	score += ResourceBinPackingScore(requested.MilliCPU, allocatable.MilliCPU, used.MilliCPU, weight.BinPackingCPU)
+	weightSum += weight.BinPackingCPU
+	score += ResourceBinPackingScore(requested.Memory, allocatable.Memory, used.Memory, weight.BinPackingMemory)
+	weightSum += weight.BinPackingMemory
+
+	// All resource with weight should be calculated, because the weightSum need it,
+	// even the node have no this resource.
+	for name, weight := range weight.BinPackingResources {
+		weightSum += weight
```
hex108: If score calculated at L202 is zero, whether we need to add `weight` to `weightSumt`?

pkg/scheduler/plugins/binpack/binpack.go
```diff
@@ -0,0 +1,230 @@
...
+	   actions: "enqueue, reclaim, allocate, backfill, preempt"
+	   tiers:
+	   - plugins:
+	     - name: binpack
+	       arguments:
+	         binpack.weight: 10
+	         binpack.cpu: 5
+	         binpack.memory: 1
+	         binpack.resources: nvidia.com/gpu, example.com/foo
+	         binpack.resources.nvidia.com/gpu: 2
```
hex108: s/`binpack.resources.nvidia.com/gpu`/`binpack.nvidia.com/gpu`/? Then it is same as cpu/memory, it is just a suggestion.

pkg/scheduler/plugins/binpack/binpack.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return weight
+}
+
+func (bp *binpackPlugin) Name() string {
+	return PluginName
+}
+
+func (bp *binpackPlugin) OnSessionOpen(ssn *framework.Session) {
+	glog.V(3).Infof("Enter binpack plugin ...")
```
hex108: Other plugins do not log these info. Maybe we could use `glog.V(4)` instead of `glog.V(3)` if needed.

pkg/scheduler/plugins/binpack/binpack.go
lminzhw: Zero is also a valid value for score, so I think add it to sum is also valid.

pkg/scheduler/plugins/binpack/binpack.go
lminzhw: use glog.V(4)

pkg/scheduler/plugins/binpack/binpack.go
lminzhw: In my opinion, all additional resource is listed in `binpack.resources`, so the key for weight of resource should have prefix of it.

pkg/scheduler/plugins/binpack/binpack.go
hex108: Take GPU for example: 
Node A has no GPU, so its GPU score will be 0, Node B has GPU capacity 6 and use 3, then it has a score for GPU resource.  Suppose that Node A and B's other resource's scores are same, if we add `weight` in this case, Node B's total score will be higher than Node A, then it will choose Node B, is it expected? Node A might be a better choice?

k82cn: if the resource is not exist on the node, we should also skip weight for it.

lminzhw: In order to choose NodeA, maybe we should make the GPU score in NodeB to be nagetive in this case?

For example, request need resource Ra/Rb, but the NodeA has resource Ra/Rb, NodeB has resource Ra/Rb/Rc.
Policy Current is the score now. Policy Skip will ignore the Sc/Wc in NodeA but could not guarantee that NodeA have a higher score than NodeB. Policy Negative would treat the Sc in NodeB as negative score.
```
policy      NodeA              NodeB
current     Sa+Sb+0/Wa+Wb+Wc   Sa+Sb+Sc/Wa+Wb+Wc
skip        Sa+Sb/Wa+Wb        Sa+Sb+Sc/Wa+Wb+Wc
negative    Sa+Sb+0/Wa+Wb+Wc   Sa+Sb-Sc/Wa+Wb+Wc
```

hex108: I don't have any better idea for this now. Just some thought:

* If request does not need resource C, then we do not calculate the score of this resource, it means that we compare `Sa+Sb/Wa+Wb`(for NodeA)  with `Sa+Sb/Wa+Wb`(for Node B).

or 

* If NodeA does not has the resource, we regards its RequestedToCapacityRatio to 1 instead of 0, it means that it is full, 0 seems it is empty.

k82cn: Option 1 is better to me :)

k82cn:
LGTM, thanks :)

pkg/scheduler/plugins/binpack/binpack.go
lminzhw: I refactor the code in `BinPackingScore`. Now, only resources that pod requested can affect the binpack score.

hex108:
Thanks! LGTM

PR #382, opened 2019-07-23
E2E admission

Rajadeepan:
E2E admission for invalid cpu and memory values

### DIFF
--- a/test/e2e/admission.go
+++ b/test/e2e/admission.go
@@ -17,9 +17,11 @@ limitations under the License.
 package e2e
 
 import (
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


test/e2e/admission.go
```diff
@@ -51,4 +53,103 @@ var _ = Describe("Job E2E Test: Test Admission service", func() {
...
+                        }
+                     }
+                  ],
+                  "restartPolicy": "Never"
+               }
+            }
+         }
+      ]
+   }
+}`)
```
hzxuzhonghu: I would suggest moving this to a file.
 and if we want to fix something in the future, it would be easy.

Issue #384, opened 2019-07-24
The job task resync logic is not right


hzxuzhonghu:
/kind bug

When i deleted several pods of volcano job, i see too many errors 

```
E0724 01:27:37.980526       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-0>, retry it, err can not find pod <default/test-job3-default-nginx-0> in cache
E0724 01:27:37.984007       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-5>, retry it, err can not find pod <default/test-job3-default-nginx-5> in cache
I0724 01:27:37.985337       1 job_controller_actions.go:273] Created Task <test-job3-default-nginx-4> of Job <default/test-job3>
I0724 01:27:37.985356       1 job_controller_actions.go:273] Created Task <test-job3-default-nginx-5> of Job <default/test-job3>
E0724 01:27:37.986466       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-0>, retry it, err can not find pod <default/test-job3-default-nginx-0> in cache
I0724 01:27:37.986642       1 job_controller_actions.go:273] Created Task <test-job3-default-nginx-0> of Job <default/test-job3>
I0724 01:27:37.989292       1 job_controller_actions.go:349] Finished Job <default/test-job3> sync up
E0724 01:27:37.989859       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-4>, retry it, err can not find pod <default/test-job3-default-nginx-4> in cache
E0724 01:27:37.991320       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-4>, retry it, err can not find pod <default/test-job3-default-nginx-4> in cache
E0724 01:27:37.992678       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-5>, retry it, err can not find pod <default/test-job3-default-nginx-5> in cache
E0724 01:27:37.994137       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-4>, retry it, err can not find pod <default/test-job3-default-nginx-4> in cache
E0724 01:27:37.995630       1 job_controller_resync.go:62] Failed to sync pod <default/test-job3-default-nginx-5>, retry it, err can not find pod <default/test-job3-default-nginx-5> in cache
[...]

PR #385, opened 2019-07-24
Fix error resync logic

hzxuzhonghu:
fixes: #384

There is time latency between `Controller.syncJob` creating a pod and the k8s informer get the notification. So there are cases job controller calling  `Controller.syncJob`  creating same pod twice,

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -265,10 +265,6 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 					pod.Name, job.Name, err)
 				creationErrs = append(creationErrs, fmt.Errorf("failed to create pod %s, err: %#v", pod.Name, err))
 			} else {
-				if err != nil && apierrors.IsAlreadyExists(err) {
-					cc.resyncTask(pod)
-				}
-
 				classifyAndAddUpPodBaseOnPhase(newPod, &pending, &running, &succeeded, &failed, &unknown)
 				glog.V(3).Infof("Created Task <%s> of Job <%s/%s>",
 					pod.Name, job.Namespace, job.Name)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -265,10 +265,6 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 					pod.Name, job.Name, err)
 				creationErrs = append(creationErrs, fmt.Errorf("failed to create pod %s, err: %#v", pod.Name, err))
 			} else {
-				if err != nil && apierrors.IsAlreadyExists(err) {
```
TommyLike: Need provide more detail why we delete this resync logic here rather than fix the "can not find pod <namespace/pod> in cache" issue.

hzxuzhonghu: I think there is a race between the core job controller and the resync controller.

And same race between `JobInfo.AddPod` and `controller.syncJob`

k82cn: > I think there is a race between the core job controller and the resync controller.

Please share your analysis of this issue firstly before PR.

hzxuzhonghu: I will add the analysis to Pr description

TommyLike:
> fixes: #384
> 
> There is time latency between `Controller.syncJob` creating a pod and the k8s informer get the notification. So there are cases job controller calling `Controller.syncJob` creating same pod twice,

Do you mean this will always happen no matter we delete pod or not 

> fixes: #384
> 
> There is time latency between `Controller.syncJob` creating a pod and the k8s informer get the notification. So there are cases job controller calling `Controller.syncJob` creating same pod twice,

This change can help fix this issue, but I am wondering if the resync logic is used to cover the case intentionaly

hzxuzhonghu:
>This change can help fix this issue, but I am wondering if the resync logic is used to cover the case intentionaly

I think so. For eventual consistent model, resync is a regular way to catch up.

Issue #359, opened 2019-07-19
Race detection in test


hzxuzhonghu:
We need race detect in our test

/kind feature

TommyLike:
@hzxuzhonghu  could you please add more detail?

hzxuzhonghu:
`go test -race`

```
-race
	enable data race detection.
	Supported only on linux/amd64, freebsd/amd64, darwin/amd64 and windows/amd64.
```

hzxuzhonghu:
completed

PR #391, opened 2019-07-26
Fix race condition issue

TommyLike:
For #359

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"sort"
 	"sync"
+	"sync/atomic"
 
 	"github.com/golang/glog"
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -253,6 +253,7 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 
 	waitCreationGroup := sync.WaitGroup{}
 	waitCreationGroup.Add(len(podToCreate))
+	stateMutex := sync.Mutex{}
```
k82cn: move mutex into function

TommyLike: `classifyAndAddUpPodBaseOnPhase` is not the only place that we need to protect.

k82cn: the other also should move to function, do not put mutex in more than on place.

hzxuzhonghu: FYI：
```
 // A Mutex must not be copied after first use.

```


So this is not right, you can change to stateMutex := &sync.Mutex{}

TommyLike: @hzxuzhonghu  just curious, for which reason why this is not working? Which code leads to the multiple copies of mutex instance?

hzxuzhonghu: Actually, this var is used by many go routines, and if you print the address of it in the go routines, you can find they are not same.  So can not protect the critical-section using different lock.

TommyLike: But the mutex instance is not used as a parameter for those go routines.

hzxuzhonghu: make sense, my memory messed up.

TommyLike:
@asifdxtreme  what's the command to retest the patch?

/retest

PR #397, opened 2019-07-30
Ignore deleting pod from updating when resync

TommyLike:
For #384

Now volcano would add the deleting pod back into cache when resync, fix it by ignore it.

### DIFF
--- a/pkg/controllers/job/job_controller_resync.go
+++ b/pkg/controllers/job/job_controller_resync.go
@@ -69,20 +69,16 @@ func (cc *Controller) syncTask(oldTask *v1.Pod) error {
 	defer cc.Mutex.Unlock()
 
 	newPod, err := cc.kubeClients.CoreV1().Pods(oldTask.Namespace).Get(oldTask.Name, metav1.GetOptions{})
-	if err != nil {
-		if errors.IsNotFound(err) {
-			if err := cc.cache.DeletePod(oldTask); err != nil {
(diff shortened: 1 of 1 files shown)


TommyLike:
@hzxuzhonghu  please take a look

pkg/controllers/job/job_controller_resync.go
```diff
@@ -69,20 +69,16 @@ func (cc *Controller) syncTask(oldTask *v1.Pod) error {
...
-			glog.V(3).Infof("Pod <%v/%v> was deleted, removed from cache.", oldTask.Namespace, oldTask.Name)
-
-			return nil
-		}
-		return fmt.Errorf("failed to get Pod <%v/%v>: err %v", oldTask.Namespace, oldTask.Name, err)
-	}
 
-	return cc.cache.UpdatePod(newPod)
+	//Delete from cache when pod does not exist or `DeletionTimestamp` not nil
+	if (err == nil && newPod.DeletionTimestamp != nil) || (err != nil && errors.IsNotFound(err)) {
```
hzxuzhonghu: I donot like this hacky way(check DeletionTimestamp here), it is already done in informer event handler.

TommyLike: Let's forget about the issue itself, would this be a more correct behavior for the `resync` logic?

hzxuzhonghu: not that i think, look at my analysis

TommyLike:
Close for #384

PR #404, opened 2019-08-02
Make Gang and Drf preemption work together

### LINKED ISSUE #390: The Gang and Drf preemption work together more reasonable
> /kind feature

**What happened**:
I'd like to use drf plugin to share resource among jobs in the same queue.
However the drf preemptionFn has no chance to be called as it is in Tier 2.

The Gang preemptionFn is always called to execute preemption. The DRF preemptionFn is only be called when Gang preemptionFn find no victims.  This is not reasonable for many use cases.

- Job(min>1) preempt job(min==1) forever
- Job(min==1) preempt job(min==1) each other forever
- Jobs(min==1) are expected to use drf to share resource. However the Gang preemptionFn is called.

When there are variables of jobs( both min>1 & min==1) in cluster. The preemption can be more reasonable while Gang and Drf are both enabled.

**Environment**:
- Volcano Version: latest

william-wang:
The PR is target to resolve the issue #390 to make drf and gang preemption work together well.

Which issue(s) this PR fixes (optional, in fixes #(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #390

Currently when I configure Gang, drf plguin and enable the preempt actions. The drf preemption has no chance to be called. This causes several problems. such as 

- Job(min>1) preempt job(min==1) forever
- Job(min==1) preempt job(min==1) each other forever
- When most of tasks in job are completed, the jobs meet both preemptor and preemptee condition, so the different jobs preempt each other endless.
drf preemption should be used to work together with Gang to resolve above issues.

Solution:
Without making any changes to current Tier design principal, the new solution is let Gang preemptFn to find the victims firstly, If all plugins preemptFn in Tier 1 find no victims, The drf preemptFn in Tier 2 is used to find victims.

**Solution Update:**
In previous solution, the DRF plguin need to take care of Job.MinAvailable. The reviewer suggested that DRF should not care the Job.MinAvaiable. So in the new solution.

Another return value is introduced to preemptableFns. 
**victims, unpreemptableList = pf(preemptor, preemptees)**
Each plugin can decide two values:
- victims:  the preemptable tasks
- upreeemptableList:  unpreemptable tasks. In some cases, Gang plugin can not find the victims from a set of tasks. However it can get a set of tasks that should not be preemptable(Min part tasks). This part of tasks can be useful for later preemption in other plugin. 

Here is an example.
There are two running jobs. DRF preemption might preempt MinAvailabe part tasks according to the share. In this solution, the Gang can detect the upreeemptableList and exclude these tasks from preemptees before later preemption in DRF.

The solution has been tested and works well.

### DIFF
--- a/docs/design/hybird-preemption.md
+++ b/docs/design/hybird-preemption.md
@@ -1,0 +1,117 @@
+# Hybird preemption
+
+[@william-wang](http://github.com/william-wang); Aug 20, 2019
+
+## Table of Contents
+
+   * [Dynamic Plugins Configuration](#dynamic-plugins-configuration)
+      * [Table of Contents](#table-of-contents)
+      * [Motivation](#motivation)
+      * [Function Detail](#function-detail)
+     
+
+## Motivation
+
+Currently when Gang and Drf plugins are both configured and enable the preempt actions. The Drf and Gang preemption can not play together
+well. The following issues have been found.
+- When job’s `min.available` is configured to 1. This Job will be preempted endlessly.
+- When the `min.available` of multiple jobs are configured to 1. These jobs will preempt each other endless.
+- When most of tasks in job are completed, the jobs meet both preemptor and preemptee condition, so different jobs
+  will preempt each other endlessly.
+
+The PR404's target is to enhance the preemption to make make Gang and Drf work together more reasonable and resolve above problems.
+
+## Function Detail
+
+The enhancement complys with the tier design that if it fits plugins in high priority tier, the action will not go through the plugins
+in lower priority tiers And in each tier, it's considered passed when all the plugins are fitted.
+
+The basic preemption behavior is that the pending jobs preempt running jobs with Gang preemption. Once the preemptor jobs turn into 
+running, the Drf preemption will be used to balance resource among running jobs according to the Drf share.
+
+The detail is as below:
+
+When the job of preemptor task is not ready, it will only preempt running tasks from other jobs which occupy more resource than it’s min.Avaiable firstly
+and turns into running. Then the Drf preemption will be involved to preempt the resource among running jobs.
+And the Drf will not preempt `min.available` part resource of job forever.
+
+
+In `Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) []*api.TaskInfo {..}` function, The plugin function `pf` is redesigned.
+
+>candidates, unpreemptableList := pf(preemptor, preemptees)
+>
+>candidates: 
+```
+Tasks that can be preemptable. 
+If the preemptor job is not ready, it will get the candidates. 
+If the preemptor job is running, It will not find candidates. It need Drf plugin to preempt resource for this running job.
+```
+>unpreemptableList:
+```
+Tasks that should not be preemptable. 
+Gang plugin has the `min.Available` info. So it knows those tasks that should not be preemptable. The framework will filter
+out these tasks and pass the left tasks to Drf preemption function. The targe is to avoid Drf preempt `min.available` part resource.
+
+```
+
+```
+For example
+1. candidates: [task1, task2], unpreemptableList:[]
+
+Gang plugin will make decision and complete preemption on task1 and task2 alone.
+
+2. Candidates:[], unpreemptableList[task1]
+
+Gang plugin will not make decision, the Drf will complete preemption alone. The preemption will ignore the task1.
+```
+
+## Use Case
+The cluster has 4 CPUs totally, each task in job-01 and job-02 requests 1 CPU.
+- The Gang and Drf work together to complete preemption
+
+    | Actions        | Job Name  | Task Num  | Min.Available | Task status |  
+    | -------------- | --------- | --------- | --------------| ----------- |
+    | Submit 1st Job | job-01    | 4         | 2             | 4 Running   |
+    | Submit 2st Job | job-02    | 4         | 1             | 4 Pending   |
+    | Result         | job-01    | 4         | 2             | 2 Running   |
+    |                | job-02    | 4         | 1             | 2 Running   |
+    
+    The job-02 preempts one task from job-01 by Gang preemption. And preempts the second task from job-01 by Drf preemption. 
+    
+- The Drf preemption does not preempt `min.available` part resource of job
+
+    | Actions        | Job Name  | Task Num  | Min.Available | Task status |  
+    | -------------- | --------- | --------- | --------------| ----------- |
+    | Submit 1st Job | job-01    | 4         | 3             | 4 Running   |
+    | Submit 2st Job | job-02    | 4         | 1             | 4 Pending   |
+    | Result         | job-01    | 4         | 3             | 3 Running   |
+    |                | job-02    | 4         | 1             | 1 Running   |
+    
+    The job-02 preempts one task from job-01 by Gang preemption. And can not continue to preempt more tasks because all 
+    remaining 3 tasks in job-01 are added into the unpreemptableList. These tasks will be ignored by framework before pass to 
+    Drf preemption.
+
+- If high tier plugin(Gang) can make decision, it will complete preemption alone.
+
+    | Actions        | Job Name  | Task Num  | Min.Available | Task status |  
+    | -------------- | --------- | --------- | --------------| ----------- |
+    | Submit 1st Job | job-01    | 4         | 2             | 4 Running   |
+    | Submit 2st Job | job-02    | 4         | 2             | 4 Pending   |
+    | Result         | job-01    | 4         | 2             | 2 Running   |
+    |                | job-02    | 4         | 2             | 2 Running   |
+    
+   The job-02 preempts two tasks from job-01 by Gang preemption.
+  
+ - When most tasks of job are completed, the jobs does not preempt each other endless any more. 
+   The Drf will balance resource among jobs. 
+   
+    | Actions        | Job Name  | Task Num  | Min.Available | Task status                       |  
+    | -------------- | --------- | --------- | --------------| ----------------------------------|
+    | Submit 1st Job | job-01    | 10        | 3             | 4 Running  2 completed 4 Pending  |
+    | Submit 2st Job | job-02    | 10        | 2             | 10 Pending 2 completed 4 Pending  |
+    | Result         | job-01    | 10        | 3             | 2 Running  2 completed 6 Pending  |
+    |                | job-02    | 10        | 2             | 2 Pending  8 pending              |
+    
+    The job-02 preempts two tasks from job-01 by Gang preemption and turns into running. And then Drf balance 
+    resource between jobs according to the share.
+
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -415,3 +415,9 @@ func (ji *JobInfo) Pipelined() bool {
 
 	return occupied >= ji.MinAvailable
 }
+
+func (ji *JobInfo) Preemptable() bool {
+	occupied := ji.ReadyTaskNum()
+
+	return ji.MinAvailable <= (occupied - 1)
+}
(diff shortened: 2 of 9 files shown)


k82cn:
any description on the problem & solution?

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -75,10 +75,18 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
 	preemptableFn := func(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) []*api.TaskInfo {
 		var victims []*api.TaskInfo
 
+		// If job is already ready, no need to preempt others in Gang plgin
+		preemptorJob := ssn.Jobs[preemptor.Job]
+		readyNum := preemptorJob.ReadyTaskNum()
+		ready := readyNum >= preemptorJob.MinAvailable || preemptorJob.MinAvailable == 1
```
hzxuzhonghu: Why if `preemptorJob.MinAvailable == 1`, do not preempt ?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -154,6 +155,9 @@ func (ssn *Session) Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskI
 				continue
 			}
 			candidates := pf(preemptor, preemptees)
+			if candidates == nil {
+				break
```
hzxuzhonghu: continue?

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -95,6 +95,15 @@ func (drf *drfPlugin) OnSessionOpen(ssn *framework.Session) {
 		allocations := map[api.JobID]*api.Resource{}
 
 		for _, preemptee := range preemptees {
+
+			// Min part tasks in job should not be preempted
+			preempteeJob := ssn.Jobs[preemptee.Job]
+			readyNum := preempteeJob.ReadyTaskNum()
+			preemptable := preempteeJob.MinAvailable <= readyNum-1 || preempteeJob.MinAvailable == 1
```
hzxuzhonghu: Abstract into a common function

pkg/scheduler/plugins/gang/gang.go
william-wang: If preemptorJob.MinAvaiable ==1, it is the scope of drf preemption. We use the drf share to preempt.

pkg/scheduler/framework/session_plugins.go
william-wang: Within one Tier, each plugin preemptFn get victims and finally get intersections.  Here if one plugin get no victim, the intersection must be nil, no need to continue.

hzxuzhonghu: ok, ic.  I didn't understand the tiers concept.

pkg/scheduler/plugins/drf/drf.go
william-wang: >preempteeJob := ssn.Jobs[preemptee.Job]
>readyNum := preempteeJob.ReadyTaskNum()

good suggestion. I thought to do it. The reson are
- only above two lines can be abstracted.
- only two places including drf/gang preemptFn use it. The reusablity is low.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -144,6 +144,7 @@ func (ssn *Session) Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskI
 	var init bool
 
 	for _, tier := range ssn.Tiers {
+		init = false
```
k82cn: why set init to false in every tier?

pkg/scheduler/framework/session_plugins.go
k82cn: I think we define return values as follow:

1. `nil` means this plugin can not make decission
1. `empty list` means this plugin rejects all victims
1. `non-empty list` means this plugin accept victims in the list

So we can not break if it's `nil` :)

pkg/scheduler/plugins/drf/drf.go
k82cn: DRF should not care about `minAvailable`.

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -107,7 +116,7 @@ func (drf *drfPlugin) OnSessionOpen(ssn *framework.Session) {
 			}
 		}
 
-		glog.V(4).Infof("Victims from DRF plugins are %+v", victims)
+		glog.V(3).Infof("Victims from DRF plugins are %+v", victims)
```
k82cn: let keep log level to 4.

pkg/scheduler/framework/session_plugins.go
william-wang: The init might be set to true, however the intersection of tier1 might be nil. 

Take below as example, plugin1 get victims and set init true, plugin2 get nil. The intersection is nil. In this case, we want to run Tier2 plugin3 with new start(init is false)

Tier1 
plugin1
plugin2
Tier2
plugin3

pkg/scheduler/plugins/drf/drf.go
william-wang: OK.  Currently the same log level is inconsistent in different plugin. For example, in Gang plugin the following log is Level3. I will change it to Level4 as well.

> glog.V(3).Infof("Victims from Gang plugins are %+v", victims)

pkg/scheduler/plugins/drf/drf.go
william-wang: Yes. pure DRF should only care about share. This is like adding another prerequisite(do not preempt not ready job) . However, this is the best way I can think of right now if we don't want to preempt min part tasks.

pkg/scheduler/framework/session_plugins.go
william-wang: @k82cn 

The solution has been updated. Please see the "Solution Update" description and help to review.

pkg/scheduler/api/job_info.go
```diff
@@ -415,3 +415,9 @@ func (ji *JobInfo) Pipelined() bool {
 
 	return occupied >= ji.MinAvailable
 }
+
+func (ji *JobInfo) Preemptable() bool {
```
k82cn: move this to related plugin, e.g. `ji.MinAvailable <= (occupied - 1)` does not mean preemptable for drf plugin.

pkg/scheduler/api/node_info.go
```diff
@@ -94,10 +94,14 @@ func NewNodeInfo(node *v1.Node) *NodeInfo {
 // Clone used to clone nodeInfo Object
 func (ni *NodeInfo) Clone() *NodeInfo {
 	res := NewNodeInfo(ni.Node)
-
+	otherRes := EmptyResource()
 	for _, p := range ni.Tasks {
 		res.AddTask(p)
+		if p.pod.Spec.SchedulerName != 'volcano' {
```
k82cn: we can not hard code scheduler name there; user can change it when install it.

pkg/scheduler/api/node_info.go
```diff
@@ -94,10 +94,14 @@ func NewNodeInfo(node *v1.Node) *NodeInfo {
...
+			otherRes.Add(p.Resreq)
+		}
 	}
+	res.Allocatable = res.Allocatable.Sub(otherRes)
```
k82cn: `Allocatable` means all 'idle' resources in the node, we can not change it.

k82cn:
please update [design doc](https://github.com/volcano-sh/volcano/blob/master/docs/design/plugin-conf.md) for this firstly.

william-wang:
@k82cn Please help to review the design doc (Hybird-preempt.md) in changed files.

docs/design/hybird-preemption.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+In `Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) []*api.TaskInfo {..}` function, The plugin function `pf` is redesigned.
+
+>candidates, unpreemptableList := pf(preemptor, preemptees)
+>
+>candidates: 
+```
+Tasks that can be preemptable. 
+If the preemptor job is not ready, it will get the candidates.
```
hzxuzhonghu: It will change the current behavior.

docs/design/hybird-preemption.md
```diff
@@ -0,0 +1,117 @@
...
+   
+    | Actions        | Job Name  | Task Num  | Min.Available | Task status                       |  
+    | -------------- | --------- | --------- | --------------| ----------------------------------|
+    | Submit 1st Job | job-01    | 10        | 3             | 4 Running  2 completed 4 Pending  |
+    | Submit 2st Job | job-02    | 10        | 2             | 10 Pending 2 completed 4 Pending  |
+    | Result         | job-01    | 10        | 3             | 2 Running  2 completed 6 Pending  |
+    |                | job-02    | 10        | 2             | 2 Pending  8 pending              |
+    
+    The job-02 preempts two tasks from job-01 by Gang preemption and turns into running. And then Drf balance 
+    resource between jobs according to the share.
```
hzxuzhonghu: should also show how to implement it.

pkg/scheduler/api/job_info.go
```diff
@@ -415,3 +415,9 @@ func (ji *JobInfo) Pipelined() bool {
...
+	occupied := ji.ReadyTaskNum()
+
+	return ji.MinAvailable <= (occupied - 1)
```
hzxuzhonghu: ji.MinAvailable < ji.ReadyTaskNum()

pkg/scheduler/framework/session_plugins.go
```diff
@@ -96,12 +96,27 @@ func (ssn *Session) AddJobEnqueueableFn(name string, fn api.ValidateFn) {
 	ssn.jobEnqueueableFns[name] = fn
 }
 
+// Exclude part of tasks from a task list
+func excludeList(tasks []*api.TaskInfo, excludedTasks []*api.TaskInfo) []*api.TaskInfo {
```
hzxuzhonghu: A little concerned with the efficiency, use map instead?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -110,9 +125,10 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 			if !found {
 				continue
 			}
-			candidates := rf(reclaimer, reclaimees)
+			candidates, unreclaimableList := rf(reclaimer, reclaimees)
```
hzxuzhonghu: why not candidates, unreclaimableList := rf(reclaimer, candidates)? 
I think this can simplify the following intersection and match.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -153,9 +182,10 @@ func (ssn *Session) Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskI
 			if !found {
 				continue
 			}
-			candidates := pf(preemptor, preemptees)
+			candidates, unpreemptableList := pf(preemptor, preemptees)
```
hzxuzhonghu: same here

pkg/scheduler/plugins/conformance/conformance.go
```diff
@@ -58,7 +58,7 @@ func (pp *conformancePlugin) OnSessionOpen(ssn *framework.Session) {
 			victims = append(victims, evictee)
 		}
 
-		return victims
+		return victims, nil
```
hzxuzhonghu: why not return the unpreemptable

docs/design/hybird-preemption.md
```diff
@@ -0,0 +1,117 @@
+# Hybird preemption
```
k82cn: let's open a separate PR for this enhancement.

PR #401, opened 2019-07-31
add podgroup controller

wangyuqing4:
Fixes #134

if normal pod/job/...... use volcano scheduler, podgroup controller watch pod, can create podgroup.

PR #406, opened 2019-08-03
remove shadowPodgroup in scheduler

wangyuqing4:
in pr #401, podgroup controller can create pg for normal pod, so remove shadowPodgroup part code in scheduler.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -45,7 +45,6 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/volumebinder"
 
 	"volcano.sh/volcano/cmd/scheduler/app/options"
-	"volcano.sh/volcano/pkg/apis/scheduling/v1alpha1"
 	"volcano.sh/volcano/pkg/apis/scheduling/v1alpha2"
 	kbver "volcano.sh/volcano/pkg/client/clientset/versioned"
 	"volcano.sh/volcano/pkg/client/clientset/versioned/scheme"
@@ -507,24 +506,9 @@ func (sc *SchedulerCache) Evict(taskInfo *kbapi.TaskInfo, reason string) error {
 		}
 	}()
 
-	if !shadowPodGroup(job.PodGroup) {
-		if job.PodGroup.Version == api.PodGroupVersionV1Alpha1 {
-			pg, err := api.ConvertPodGroupInfoToV1alpha1(job.PodGroup)
-			if err != nil {
-				glog.Errorf("Error While converting api.PodGroup to v1alpha.PodGroup with error: %v", err)
-				return err
-			}
-			sc.Recorder.Eventf(pg, v1.EventTypeNormal, "Evict", reason)
-		} else if job.PodGroup.Version == api.PodGroupVersionV1Alpha2 {
-			pg, err := api.ConvertPodGroupInfoToV1alpha2(job.PodGroup)
-			if err != nil {
-				glog.Errorf("Error While converting api.PodGroup to v2alpha.PodGroup with error: %v", err)
-				return err
-			}
-			sc.Recorder.Eventf(pg, v1.EventTypeNormal, "Evict", reason)
-		} else {
-			return fmt.Errorf("Invalid PodGroup Version: %s", job.PodGroup.Version)
-		}
+	if err := sc.convertPodGroupInfo(job); err != nil {
+		glog.Errorf("Error While converting api.PodGroup %v", err)
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -507,24 +507,22 @@ func (sc *SchedulerCache) Evict(taskInfo *kbapi.TaskInfo, reason string) error {
...
-		} else if job.PodGroup.Version == api.PodGroupVersionV1Alpha2 {
-			pg, err := api.ConvertPodGroupInfoToV1alpha2(job.PodGroup)
-			if err != nil {
-				glog.Errorf("Error While converting api.PodGroup to v2alpha.PodGroup with error: %v", err)
-				return err
-			}
-			sc.Recorder.Eventf(pg, v1.EventTypeNormal, "Evict", reason)
-		} else {
-			return fmt.Errorf("Invalid PodGroup Version: %s", job.PodGroup.Version)
+	if job.PodGroup.Version == api.PodGroupVersionV1Alpha1 {
```
k82cn: let's move this into a helper function.

pkg/scheduler/cache/util.go
```diff
@@ -17,43 +17,20 @@ limitations under the License.
...
-
-	return found
-}
-
 func createShadowPodGroup(pod *v1.Pod) *api.PodGroup {
-	jobID := api.JobID(utils.GetController(pod))
-	if len(jobID) == 0 {
-		jobID = api.JobID(pod.UID)
-	}
+	pgName := helpers.GeneratePodgroupName(pod)
```
k82cn: why do we still need this function?

wangyuqing4: just want to keep the same to get pi.Job, like vk job NewJobInfo here.

image

k82cn: we should also update pod's in podgroup controller for that.

wangyuqing4: ? ? in podgroup controller pg name is the same as here.

pkg/scheduler/cache/cache.go
wangyuqing4: new a func convertPodGroupInfo() :)

pkg/scheduler/cache/cache.go
```diff
@@ -837,3 +804,25 @@ func (sc *SchedulerCache) UpdateJobStatus(job *kbapi.JobInfo, updatePG bool) (*k
...
 }
+
+func (sc *SchedulerCache) convertPodGroupInfo(job *kbapi.JobInfo, eventtype, reason, message string) error {
+	if job.PodGroup.Version == api.PodGroupVersionV1Alpha1 {
+		pg, err := api.ConvertPodGroupInfoToV1alpha1(job.PodGroup)
+		if err != nil {
+			glog.Errorf("Error While converting api.PodGroup to v1alpha.PodGroup with error: %v", err)
+			return err
+		}
+		sc.Recorder.Eventf(pg, eventtype, reason, message)
```
k82cn: why do we record this event?

pkg/scheduler/cache/cache.go
```diff
@@ -507,24 +507,8 @@ func (sc *SchedulerCache) Evict(taskInfo *kbapi.TaskInfo, reason string) error {
...
-			pg, err := api.ConvertPodGroupInfoToV1alpha2(job.PodGroup)
-			if err != nil {
-				glog.Errorf("Error While converting api.PodGroup to v2alpha.PodGroup with error: %v", err)
-				return err
-			}
-			sc.Recorder.Eventf(pg, v1.EventTypeNormal, "Evict", reason)
-		} else {
-			return fmt.Errorf("Invalid PodGroup Version: %s", job.PodGroup.Version)
-		}
+	if err := sc.convertPodGroupInfo(job, v1.EventTypeNormal, "Evict", reason); err != nil {
```
TommyLike: Move the record event code outside of the convertPodGroupInfo.

pkg/scheduler/cache/util.go
TommyLike: At lease we should rename this function, since the `ShadowPodGroup ` concept has been deprecated.

pkg/scheduler/cache/cache.go
```diff
@@ -778,32 +762,15 @@ func (sc *SchedulerCache) RecordJobStatusEvent(job *kbapi.JobInfo) {
...
-				if err != nil {
-					glog.Errorf("Error while converting PodGroup to v1alpha2.PodGroup with error: %v", err)
-				}
-				sc.Recorder.Eventf(podGroup, v1.EventTypeWarning,
-					string(v1alpha1.PodGroupUnschedulableType), msg)
-			}
+	// If pending or unschedulable, record unschedulable event.
+	if pgUnschedulable || pdbUnschedulabe {
+		if err := sc.convertPodGroupInfo(job); err != nil {
+			glog.Errorf("Error While converting api.PodGroup %v", err)
```
k82cn: We should record event here :)

Issue #403, opened 2019-08-02
ssh plugin adds user flag


chenyangxueHDU:
> /kind feature

**What happened**:
The ssh plugin has --no-root flag for common user ssh, but I need to cp/link .ssh dir to user home dir.

**What you expected to happen**:
I would like to add --user flag to mount ssh configmap to user home dir, like --user=someone, mount .ssh to /home/users/someone/.ssh

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

k82cn:
hm... how to change the ownership of `.ssh`?

hzxuzhonghu:
Currently, the permission is 0600, which is decided by ssh.

@chenyangxueHDU Do you mean you want to use a specific user instead of root in your docker images?

TommyLike:
I guess he is asking whether volcano could mount the ssh config file into a user default folder, for instance `/home/user1/.ssh/`. This is not supported now, see https://github.com/volcano-sh/volcano/blob/1a856a19c44c41257d71de52e0a10c66bc001af9/pkg/controllers/job/plugins/ssh/ssh.go#L97

cc @k82cn  @hzxuzhonghu

chenyangxueHDU:
> hm... how to change the ownership of `.ssh`?

Yes, it's the problem. I think we can pass uid/gid, then we can change the ownership

chenyangxueHDU:
> Currently, the permission is 0600, which is decided by ssh.
> 
> @chenyangxueHDU Do you mean you want to use a specific user instead of root in your docker images?

yes, I use the `runuser` to run the command.

chenyangxueHDU:
> I guess he is asking whether volcano could mount the ssh config file into a user default folder, for instance `/home/user1/.ssh/`. @chenyangxueHDU This is not supported now, see
> 
> https://github.com/volcano-sh/volcano/blob/1a856a19c44c41257d71de52e0a10c66bc001af9/pkg/controllers/job/plugins/ssh/ssh.go#L97
> 
> cc @k82cn @hzxuzhonghu

I know, so it is a feature

hzxuzhonghu:
[...]

PR #407, opened 2019-08-05
Allow setting user name in ssh plugin

hzxuzhonghu:
Fixes： #403

To support specifying ssh user when running volcano job.

```
kind: Job
metadata:
  name: lm-mpi-job
spec:
  minAvailable: 3
  schedulerName: volcano
  plugins:
    ssh: ["--user=volcano"]    # specify volcano username
    svc: []
  tasks:
    ...

```

### DIFF
--- a/pkg/controllers/job/plugins/ssh/ssh.go
+++ b/pkg/controllers/job/plugins/ssh/ssh.go
@@ -21,12 +21,12 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
-	"flag"
 	"fmt"
-
-	"golang.org/x/crypto/ssh"
+	"path"
 
 	"github.com/golang/glog"
+	"github.com/spf13/pflag"
+	"golang.org/x/crypto/ssh"
 
 	"k8s.io/api/core/v1"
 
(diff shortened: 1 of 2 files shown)


k82cn:
@hzxuzhonghu , would you help to print output of `mpirun` under `volcano` user, and show the ownership and permission of `.ssh`?

TommyLike:
@hzxuzhonghu  Does this change mean the user need to start the container via the identical user configured here?

hzxuzhonghu:
>would you help to print output of mpirun under volcano user, and show the ownership and permission of .ssh?

Good point, the ownership of the id_rsa is root, not the user itself. I think this is expected by k8s.

So we canot do this..

TommyLike:
> So we canot do this..

Does this mean we can't print the permission or we can not address feature like this?

hzxuzhonghu:
Yep, i can not think of a perfect solution.

TommyLike:
> Yep, i can not think of a perfect solution.

It should be ok if the `.ssh` folder is created by kubernentes.

TommyLike:
plus, this is what I found in the MPI test:
```
root@mpi-mpimaster-0:~# ls -al
total 20
drwx------ 1 root root 4096 Aug  7 06:25 .
drwxr-xr-x 1 root root 4096 Aug  7 06:25 ..
-rw-r--r-- 1 root root 3106 Oct 22  2015 .bashrc
-rw-r--r-- 1 root root  148 Aug 17  2015 .profile
drwxr-xr-x 2 root root 4096 Aug  7 06:25 .ssh
```

hzxuzhonghu:
Yes, i looked into this and found that the volume mount by kubelet is owned by root. Currently, k8s does not support user setting. So i think it is a blocker for this.

Workaround maybe change the ownership in container. But this requires the image including some tools like sudo, chown.

k82cn:
@hzxuzhonghu , let's move the discussion to issue and close this PR; we can open another PR if necessary.

hzxuzhonghu:
ok

Issue #409, opened 2019-08-08
How to use privilege container with Volcano? 


goversion:
**What happened:**

I need to use privilege level in a volcano job's yaml file, which is to add “privileged: true” to the yaml file.

When creating the job, the following error will be reported.

```
$ kubectl create -f abc.yaml
Error from server: error when creating "abc.yaml":
admission webhook "validatejob.volcano.sh" denied the request:
spec.task[0].template.spec.containers[0].securityContext.privileged:
Forbidden: disallowed by cluster policy.
spec.task[1].template.spec.containers[0].securityContext.privileged:
Forbidden: disallowed by cluster policy.
```

Note that kube-apiserver and kubelet have been set "privilege=true".

Enter the following command to delete the settings related to volcano admission. I can avoid the above error when creating the job, then the job can use privilege level.

```
$ kubectl delete validatingwebhookconfiguration volcano-admission-service-validate-job
```

However, the problem is that volcano admission does not verify the fields in the yaml file at all, which is not good.

What better solution does the community have, please? Thanks!

**What you expected to happen:**
I would like to add “privileged: true” to a volcano job's yaml file, this is not allowed at present.

**How to reproduce it (as minimally and precisely as possible):**
1. enable kube-apiserver with --allow-privileged=true ...
2. enable kubelet with --allow-privileged=true ...
3. Create a sample volcano job yaml file and add "privileged: true" field to the yaml. then create the job with command "kubectl apply". 

**Environment**
Volcano Version : The latest version 
kubernetes version : 1.14.3

hzxuzhonghu:
@goversion  Thanks for reporting, it seems a bug, which we did not consider before.

hzxuzhonghu:
Given to the validation in-consistent in volcano and kube-apiserver. I would suggest introducing max-retry to volcano, so we can prevent endless retry.

hzxuzhonghu:
Maybe i should file a new issue.

PR #411, opened 2019-08-12
Skip verify volcano job container's Privileged mode

hzxuzhonghu:
fixes: #409

### DIFF
--- a/pkg/admission/admit_job.go
+++ b/pkg/admission/admit_job.go
@@ -173,7 +173,8 @@ func validateTaskTemplate(task v1alpha1.TaskSpec, job v1alpha1.Job, index int) s
 	var coreTemplateSpec k8score.PodTemplateSpec
 	k8scorev1.Convert_v1_PodTemplateSpec_To_core_PodTemplateSpec(&v1PodTemplate.Template, &coreTemplateSpec, nil)
 
-	// Skip verify container SecurityContex.Privileged
+	// Skip verify container SecurityContex.Privileged as it depends on
+	// the kube-apiserver `allow-privileged` flag.
 	for i, container := range coreTemplateSpec.Spec.Containers {
 		if container.SecurityContext != nil && container.SecurityContext.Privileged != nil {
 			coreTemplateSpec.Spec.Containers[i].SecurityContext.Privileged = nil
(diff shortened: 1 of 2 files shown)


pkg/admission/admit_job.go
```diff
@@ -173,6 +173,13 @@ func validateTaskTemplate(task v1alpha1.TaskSpec, job v1alpha1.Job, index int) s
 	var coreTemplateSpec k8score.PodTemplateSpec
 	k8scorev1.Convert_v1_PodTemplateSpec_To_core_PodTemplateSpec(&v1PodTemplate.Template, &coreTemplateSpec, nil)
 
+	// Skip verify container SecurityContex.Privileged
+	for i, container := range coreTemplateSpec.Spec.Containers {
+		if container.SecurityContext != nil && container.SecurityContext.Privileged != nil {
+			coreTemplateSpec.Spec.Containers[i].SecurityContext.Privileged = nil
```
k82cn: Where did we validate `privileged`, `ValidatePodTemplate`? BTW, we should enable/disable this validation according to `--allow-privileged` of kube-apiserver.

hzxuzhonghu: Yes. Just leave the validation to apiserver, as we may not know the cluster flag.

TommyLike: > Yes. Just leave the validation to apiserver, as we may not know the cluster flag.

Better to add this into code notes.

k82cn: is there any other validation dependent on api-server parameter?

hzxuzhonghu: Let me  have an investigate

hzxuzhonghu: Should have no other fields. But will ignore some fields validation which depends on the feature gate, like `CustomPodDNS`.   

I think this still does not matter as #412 in.

k82cn: hm... so how to identify those issues before user?

k82cn: @hzxuzhonghu , let's get this merged firstly, and open an issue on how to identify those issues before user;

hzxuzhonghu: I suggest introducing k8s native ut cases into volcano, i think they should already cover corner cases.

PR #412, opened 2019-08-12
Add maxRetry in job controller to prevent endless loop

hzxuzhonghu:
In case there is invalid job container template(like stated https://github.com/volcano-sh/volcano/issues/409#issuecomment-520295818), that cannot be rejected at volcano admission. The job controller can get into an endless loop error retrying.

### DIFF
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -328,8 +328,9 @@ func (cc *Controller) processNextReq(count uint32) bool {
 			queue.AddRateLimited(req)
 			return true
 		}
-
-		glog.V(2).Infof("Dropping job<%s/%s> out of the queue: %v", jobInfo.Job.Namespace, jobInfo.Job.Name, err)
+		cc.recordJobEvent(jobInfo.Job.Namespace, jobInfo.Job.Name, vkbatchv1.ExecuteAction, fmt.Sprintf(
+			"Job failed on action %s for retry limit reached", action))
+		glog.Warningf("Dropping job<%s/%s> out of the queue: %v because max retries has reached", jobInfo.Job.Namespace, jobInfo.Job.Name, err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/job/job_controller.go
```diff
@@ -312,11 +321,15 @@ func (cc *Controller) processNextReq(count uint32) bool {
...
-		glog.Errorf("Failed to handle Job <%s/%s>: %v",
-			jobInfo.Job.Namespace, jobInfo.Job.Name, err)
-		// If any error, requeue it.
-		queue.AddRateLimited(req)
-		return true
+		if queue.NumRequeues(req) < maxRetries {
+			glog.V(2).Infof("Failed to handle Job <%s/%s>: %v",
+				jobInfo.Job.Namespace, jobInfo.Job.Name, err)
+			// If any error, requeue it.
+			queue.AddRateLimited(req)
```
k82cn: We should update job status and record events accordingly.

TommyLike: job status and event inside of the Execute function should be enough for this.

pkg/controllers/job/job_controller.go
```diff
@@ -312,11 +321,15 @@ func (cc *Controller) processNextReq(count uint32) bool {
...
+			return true
+		}
+
+		glog.V(2).Infof("Dropping job<%s/%s> out of the queue: %v", jobInfo.Job.Namespace, jobInfo.Job.Name, err)
```
TommyLike: ``glog.V(2).Warnf`` ?, also better to announce in log that its because of maxRetries exceeded.

pkg/controllers/job/job_controller.go
k82cn: what's kind of event in Execute? How user identify whether it's done?

PR #416, opened 2019-08-16
Improved log output so that successful pods do not exist on node

CuZn13:
A successful pod is not found on node and does not need to be printed with an error


k82cn:
/cc @william-wang 

@CuZn13 would help to fix ci issue :)

hzxuzhonghu:
@CuZn13 FYI: https://github.com/volcano-sh/volcano/pull/355#issuecomment-513124991

PR #370, opened 2019-07-20
add admitPod

wangyuqing4:
Fixes: #135 

Add admission controller(admitPod)

add ValidatingWebhookConfiguration volcano-validate-pod, only limit CREATE pods, allow pods to create when:
- pod.spec.schedulerName is not volcano
- podgroup phase isn't Pending
- normal job, no podgroup

PR #421, opened 2019-08-19
remove inqueue job phase

wangyuqing4:
pr #370 add ValidatingWebhookConfiguration volcano-validate-pod, limit CREATE pods.

so inqueue job phase can be removed.


pkg/controllers/job/job_controller_actions.go
```diff
@@ -210,6 +186,10 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 		return nil
 	}
 
+	if err := cc.createJob(job); err != nil {
```
hzxuzhonghu: confusing createJob in synJob

hzxuzhonghu: better rename it

PR #422, opened 2019-08-20
Updated volcano logo

k82cn:
Signed-off-by: Klaus Ma <mada3@huawei.com>


hzxuzhonghu:
Why you remove these jobs?

k82cn:
it's unstable, sometime the whole tasks got cancelled :(

TommyLike:
@k82cn The process you deleted are not triggered for the common PR builds. And the stable is related to the E2E tests not the publish stages, we may need investigating on how to avoid this.

k82cn:
@TommyLike , I had meet several times that the CI was cancelled but all e2e tests were passed. Let's revert it when it's ready for us.

TommyLike:
@k82cn  Usually we have several **jobs** for one PR **build**, if one specific job failed, for example the E2E job, then the following jobs (publish) would be canceled. In this case we can trigger two different events in travis.ci website. 

1. Restart the job
2. Restart the whole build.

If we only restart the job , the E2E test may pass next time, but the build still marked **canceled**,  But if we restart the build, that would be ok I think.

I trigged a whole rebuild for this https://github.com/volcano-sh/volcano/pull/421, let's see the result.

k82cn:
> If we only restart the job , the E2E test may pass next time, but the build still marked canceled, But if we restart the build, that would be ok I think.

Is there anyway to fix that? I don-t want to put effort on identify which button I should push :(. If we can not fix that, just use a script to include all related tests.

TommyLike:
> > If we only restart the job , the E2E test may pass next time, but the build still marked canceled, But if we restart the build, that would be ok I think.
> 
> Is there anyway to fix that? I don-t want to put effort on identify which button I should push :(. If we can not fix that, just use a script to include all related tests.

If we want to retrigger the build on the website, I guess we need to get used to the Travis's design.
image
But if we want a more convenient way, I think we need add this support in ci bot.

k82cn:
I removed the modification to .trivis, it's seems it does not work too.

> But if we want a more convenient way, I think we need add this support in ci bot.

That's really a bad contributor experience, please open an issue to fix that.

k82cn:
If not more questions/comments, let's get log update merged :)

PR #434, opened 2019-09-04
Add demo about Click-Through-Rate distributed training with PaddlePad…

sivanzcw:
Add demo about distributed training with PaddlePaddle on Volcano, source demo taken from https://github.com/PaddlePaddle/edl/tree/develop/example/ctr

### DIFF
--- a/example/integrations/paddlepaddle/ctr-paddlepaddle-on-volcano.yaml
+++ b/example/integrations/paddlepaddle/ctr-paddlepaddle-on-volcano.yaml
@@ -1,0 +1,183 @@
+apiVersion: batch.volcano.sh/v1alpha1
+kind: Job
+metadata:
+  name: ctr-volcano
+spec:
+  minAvailable: 4
+  schedulerName: volcano
+  policies:
+  - event: PodEvicted
+    action: RestartJob
+  - event: PodFailed
+    action: RestartJob
+  tasks:
+  - replicas: 2
+    name: pserver
+    template:
+      metadata:
+        labels:
+          paddle-job-pserver: fluid-ctr
+      spec:
+        imagePullSecrets:
+        - name: default-secret
+        volumes:
+        - hostPath:
+            path: /home/work/
+            type: ""
+          name: seqdata
+        containers:
+        - image: volcanosh/edlctr:v1
+          command:
+          - paddle_k8s
+          - start_fluid
+          imagePullPolicy: IfNotPresent
+          name: pserver
+          volumeMounts:
+          - mountPath: /mnt/seqdata
+            name: seqdata
+          resources:
+            limits:
+              cpu: 10
+              memory: 30Gi
+              ephemeral-storage: 10Gi
+            requests:
+              cpu: 1
+              memory: 100M
+              ephemeral-storage: 1Gi
+          env:
+          - name: GLOG_v
+            value: "0"
+          - name: GLOG_logtostderr
+            value: "1"
+          - name: TOPOLOGY
+            value: ""
+          - name: TRAINER_PACKAGE
+            value: /workspace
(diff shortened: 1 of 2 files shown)


example/integrations/paddlepaddle/ctr-paddlepaddle-on-volcano.yaml
```diff
@@ -0,0 +1,187 @@
...
+      spec:
+        imagePullSecrets:
+        - name: default-secret
+        volumes:
+        - hostPath:
+            path: /home/work/
+            type: ""
+          name: seqdata
+        containers:
+        - image: sivanzcw/edlctr:v1
```
k82cn: let's move this image to volcanosh

example/integrations/paddlepaddle/ctr-paddlepaddle-on-volcano.yaml
```diff
@@ -0,0 +1,187 @@
...
+          - name: GLOG_v
+            value: "0"
+          - name: GLOG_logtostderr
+            value: "1"
+          - name: TOPOLOGY
+            value: ""
+          - name: TRAINER_PACKAGE
+            value: /workspace
+          - name: PADDLE_INIT_NICS
+            value: eth2
```
hzxuzhonghu: I am not familiar with paddlepaddle, but what's this?

sivanzcw: `PADDLE INIT NICS` is used to pass the `nics` parameter to specify the network card in the `paddle pserver` or `paddle train` command [here](https://github.com/PaddlePaddle/edl/blob/develop/docker/paddle_k8s)

 ```bash
start_pserver() {
    stdbuf -oL paddle pserver \
      --use_gpu=0 \
      --port=$PADDLE_INIT_PORT \
      --ports_num=$PADDLE_INIT_PORTS_NUM \
      --ports_num_for_sparse=$PADDLE_INIT_PORTS_NUM_FOR_SPARSE \
      --nics=$PADDLE_INIT_NICS \
      --comment=paddle_process_k8s \
      --num_gradient_servers=$PADDLE_INIT_NUM_GRADIENT_SERVERS
}
```

 The default value is taken from the relevant configuration in the PaddlePaddle EDL demo [here](https://github.com/PaddlePaddle/edl/blob/develop/example/ctr/ps-train/pserver.yaml)

example/integrations/paddlepaddle/ctr-paddlepaddle-on-volcano.yaml
sivanzcw: Already replaced with volcanosh image

example/integrations/paddlepaddle/ctr-paddlepaddle-on-volcano.yaml
hzxuzhonghu: What if `eth2` not exist in a container?

sivanzcw: After testing, for this demo, setting a non-existent NIC for PaddlePaddle does not affect the training process. Based on the original Baidu demo, this parameter was removed. During the implementation of the demo, each pod filters out the `server` and `trainer` components through the preset `PServer pod lable` and `Trainer pod lable`, thus obtaining the ip list of the `server` and `trainer` components. When obtaining the server ip list, the system assign a port to the `server` at the same time. After the `server` gets the assigned port, it starts listening to the corresponding port and provides services. Finally, each `server` or `trainer` component in the computing cluster knows the detailed communication address of other components under the cluster, so the components can communicate directly without using the network card configuration.

PR #435, opened 2019-09-04
remove the nodes that cannot be used to run workload

suigh:
This logic is used to remove the nodes (E.g. master) that cannot be used.

1. The related logic in precidate will still keep there in case reusing them in the future.

2. This fix can resolve the issue in following link partially, please see it for more details.
https://github.com/kubernetes-sigs/kube-batch/pull/889

3. This fix can resolve an issue of DRF. Append the details here:
Let's say there are 2 nodes in cluster, one is master, another is worker. Master has 4 cpu, 8G memory, worker has 8 cpu, 4G memory.  There are two jobs with many tasks, a task in the first job asks for 1cpu and 0.5G memory, a task in the second job asks for 0.5cpu and 1G memory.
If the master node is included by DRF, there will be 12cpu and 12G memory, and by the current ordering rule (see jobOrderFn in drf.go:203), the code will:
    a. get one task from first job and assign resource.
    b. get one task from second job and assign resource.
    c. repeat a and b until there are no tasks/resource.
If the master node is not included by DRF, there will be 8 cpu and 4G memory, the code will:
    a. get one task from first job and assign resource.
    b. get one task from second job and assign resource.
    c. get one task from first job and assign resource.
    d. get one task from first job and assign resource.
    e. get one task from second job and assign resource.

As we can see, from the step d, the result is different, it is impacted if the master node is included, while we cannot use it at all, it should be removed.

### DIFF
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -25,6 +25,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	schedulercache "k8s.io/kubernetes/pkg/scheduler/cache"
 
 	"volcano.sh/volcano/pkg/apis/scheduling"
 	"volcano.sh/volcano/pkg/scheduler/api"
@@ -129,6 +130,26 @@ func openSession(cache cache.Cache) *Session {
 	}
 
 	ssn.Nodes = snapshot.Nodes
+
+	//remove the nodes that cannot be used to run workload
+	for key, n := range ssn.Nodes {
+		nodeInfo := schedulercache.NewNodeInfo(n.Pods()...)
+		nodeInfo.SetNode(n.Node)
+		taints, _ := nodeInfo.Taints()
+		ignore := false
+		for i := range taints {
+			t := taints[i]
+			if t.Effect == v1.TaintEffectNoSchedule || t.Effect == v1.TaintEffectNoExecute {
+				ignore = true
+				glog.V(3).Infof("Ingore host <%v>,  %v.", n.Name, t)
+				break
+			}
+		}
+		if ignore {
+			delete(ssn.Nodes, key)
+		}
+	}
+
 	ssn.Queues = snapshot.Queues
 	ssn.NamespaceInfo = snapshot.NamespaceInfo
 


k82cn:
@suigh , thanks very much for your PR!

Your case is valid and we should handle that; but I think the solution should be enhanced: a Pod can be scheduled to the tainted node with toleration, maybe we can ask user to identify which nodes are worker nodes :)

suigh:
> @suigh , thanks very much for your PR!
> 
> Your case is valid and we should handle that. But I think the solution should be enhanced, as a Pod can be scheduled to the tainted node with toleration; maybe we can ask user to identify which nodes are worker nodes :)

In the current logic, once the "predicates" plugin is enabled, the master node will not be used, this is the default behavior, user cannot identify the work nodes.  If we need to enhance this solution in this way (ask user to identify the worker node), maybe we should also enhance the current behavior.

Or, can we add a new plugin interface like "PredicateValidNodes" in  "predicates" plugin to identify the worker node, and call it at the beginning of the scheduling to remove the non-worker node from the session? Seems this can fit the current design structure better, and the result the same as the current added code.

k82cn:
> In the current logic, once the "predicates" plugin is enabled, the master node will not be used, this is the default behavior, user cannot identify the work nodes. 

The "master" node can be used by the pod with toleration, `predicates` does not reject that.
 
> Or, can we add a new plugin interface like "PredicateValidNodes" in "predicates" plugin to identify the worker node, and call it at the beginning of the scheduling to remove the non-worker node from the session? Seems this can fit the current design structure better, and the result the same as the current added code.

hm... the master and worker nodes are identified by administrator or installer; I'm not sure whether a plugin is enough to handle that :)

> Add the function "PredicateValidNodes" and remove the function "predicateFns" in the predicates.go should be what we need.

We should not remove `predicateFns`, it provides several features to let pod filter nodes out.

suigh:
> > In the current logic, once the "predicates" plugin is enabled, the master node will not be used, this is the default behavior, user cannot identify the work nodes.
> 
> The "master" node can be used by the pod with toleration, `predicates` does not reject that.
> 

How to enable to pod for toleration?  For the logic in "predicates.PodToleratesNodeTaints" (from from vendor\k8s.io\kubernetes\pkg\scheduler\algorithm\predicates\predicates.go), the pod information is not checked at all. Seems I miss some logic.

k82cn:
Please refer to  https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ on how to enable Toleration in Pod :)

suigh:
> Please refer to https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ on how to enable Toleration in Pod :)

Thanks for the information. I had missed some details.  :(

Then, a possible solution for this issue is to check the toleration of every pod, and remove the ndoes that cannot be used for any pod, the cost is high as the pod will keep coming....  Just clost this PR.

k82cn:
> Then, a possible solution for this issue is to check the toleration of every pod, and remove the ndoes that cannot be used for any pod, the cost is high as the pod will keep coming.... Just clost this PR.

I'm thinking to have a "configuration", e.g. `nodeSelector`, for scheduler to identify worker nodes; the configuration is set/updated by administrator or installer.

suigh:
> > Then, a possible solution for this issue is to check the toleration of every pod, and remove the ndoes that cannot be used for any pod, the cost is high as the pod will keep coming.... Just clost this PR.
> 
> I'm thinking to have a "configuration", e.g. `nodeSelector`, for scheduler to identify worker nodes; the configuration is set/updated by administrator or installer. Please feel free to open other issue for that.

Yes, a solution like "nodeSelector" is better. 

I had thought about removing the nodes by checking job level requirement, while it is still not smooth.

PR #439, opened 2019-09-06
set the json name of exitCode

davidstack:
set the json name of exitCode ,so we can specify the exitcode in the policies.
       policies:
              - exitCode: 137
               action: RestartJob

### DIFF
--- a/pkg/apis/batch/v1alpha1/job.go
+++ b/pkg/apis/batch/v1alpha1/job.go
@@ -193,7 +193,7 @@ type LifecyclePolicy struct {
 	// according to this code.
 	// Note: only one of `Event` or `ExitCode` can be specified.
 	// +optional
-	ExitCode *int32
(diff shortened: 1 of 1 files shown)


pkg/apis/batch/v1alpha1/job.go
```diff
@@ -193,7 +193,7 @@ type LifecyclePolicy struct {
 	// according to this code.
 	// Note: only one of `Event` or `ExitCode` can be specified.
 	// +optional
-	ExitCode *int32
+	ExitCode *int32 `json:"exitCode,omitempty" protobuf:"bytes,3,opt,name=exitCode"`
```
k82cn: Good catch!

It should be `5`, `3` is used by `Events` :)

davidstack: got it,i have changed this.

PR #445, opened 2019-09-11
Queue refactor.

k82cn:

### DIFF
--- a/pkg/controllers/queue/queue_controller.go
+++ b/pkg/controllers/queue/queue_controller.go
@@ -17,6 +17,8 @@ limitations under the License.
 package queue
 
 import (
+	"fmt"
+	"reflect"
 	"sync"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


pkg/controllers/queue/queue_controller.go
```diff
@@ -137,40 +139,50 @@ func (c *Controller) processNextWorkItem() bool {
...
-	var pending, running, unknown, inqueue int32
+func (c *Controller) getPodGroups(key string) ([]string, error) {
 	c.pgMutex.RLock()
+	defer c.pgMutex.RUnlock()
+
 	if c.podGroups[key] == nil {
-		c.pgMutex.RUnlock()
-		glog.V(2).Infof("queue %s has not been seen or deleted", key)
-		return nil
+		return nil, fmt.Errorf("queue %s has not been seen or deleted", key)
```
hzxuzhonghu: It is dangerous, think a queue is deleted before a pg. It will cause endless retry.

k82cn: let's handle this in different PR, refer to the TODO in following codes :)

PR #456, opened 2019-09-18
Refine queue controller

hzxuzhonghu:
1. When podgroup.spec.queue not specified, use `default` by default, which is consistent with scheduler.

1. Do not treat non podgroups as an error, since it is common.

### DIFF
--- a/pkg/controllers/queue/queue_controller.go
+++ b/pkg/controllers/queue/queue_controller.go
@@ -17,7 +17,6 @@ limitations under the License.
 package queue
 
 import (
-	"fmt"
 	"reflect"
 	"sync"
 
@@ -34,6 +33,9 @@ import (
 	kbinformer "volcano.sh/volcano/pkg/client/informers/externalversions/scheduling/v1alpha2"
 	kblister "volcano.sh/volcano/pkg/client/listers/scheduling/v1alpha2"
 )
+
+// The default queue when not specified in podgroup
+const defaultQueue = "default"
 
 // Controller manages queue status.
 type Controller struct {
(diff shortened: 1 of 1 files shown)


k82cn:
IMO, we should keep Queue event handler here; we should not let queue events dependent on podgroups events.

hzxuzhonghu:
We should only update queue status until there are podgroups for it.

TommyLike:
> We should only update queue status until there are podgroups for it.

@hzxuzhonghu Other than the discussion point here, is it possible that we could update the queue to the incorrect status when the controller restarted that only part of the PG has been cached locally?

k82cn:
> We should only update queue status until there are podgroups for it.

Let's keep event handler there; that's really wired that Queue's handler dependent on PodGroup's event :(

hzxuzhonghu:
Then i think this part need a refactor

k82cn:
> Then i think this part need a refactor

Please find another solution for that.

pkg/controllers/queue/queue_controller.go
```diff
@@ -279,12 +280,21 @@ func (c *Controller) deletePodGroup(obj interface{}) {
...
 
-	delete(c.podGroups[pg.Spec.Queue], key)
+	delete(c.podGroups[queue], key)
 
-	c.queue.Add(pg.Spec.Queue)
+	c.queue.Add(queue)
+}
+
+func getPogGroupQueue(pg *schedulingv1alpha2.PodGroup) string {
+	if pg.Spec.Queue == "" {
```
k82cn: prefer to handle this in admission controller.

hzxuzhonghu: We can mutate in admission webhook, but keep in mind admission is an optional component.

We do not have admission for pg now, it is a big cost just for a default value set. Anyway if you really insist on that,   crd definition can set default value.

PR #460, opened 2019-09-19
modify the return value of 'vcctl'

### LINKED ISSUE #454: 'vcctl job run' executes with no arguments

> Uncomment only one, leave it on its own line:
>
 /kind bug
> /kind feature

**What happened**:
when I use the command line `vcctl job run` , I forget to add arguments such as `-N job-name` and enter, then the `vcctl` executes the job `test` automatically as below:
捕获

report my command lacks arguments and stop

**Environment**:
- Volcano Version: release-0.2
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jiangkaihua:
Fixes #454 

in cmd/cli/vkctl.go, return error code '2' instead of '0' when error occurs in 'vcctl', so that cromwell can read the correct 'check-alive' which refers to 'vcctl job view -N <job-name>'.

in pkg/cli/job.go, when job is nil, do not run the default job 'test', print the error output and return nil.

### DIFF
--- a/pkg/cli/job/run.go
+++ b/pkg/cli/job/run.go
@@ -72,23 +72,14 @@ func RunJob() error {
 		return err
 	}
 
-	req, err := populateResourceListV1(launchJobFlags.Requests)
-	if err != nil {
-		return err
-	}
-
-	limit, err := populateResourceListV1(launchJobFlags.Limits)
-	if err != nil {
-		return err
-	}
-
 	job, err := readFile(launchJobFlags.FileName)
 	if err != nil {
 		return err
 	}
 
 	if job == nil {
-		job = constructLaunchJobFlagsJob(launchJobFlags, req, limit)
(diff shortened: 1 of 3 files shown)


pkg/cli/job/run.go
```diff
@@ -72,23 +73,14 @@ func RunJob() error {
 		return err
 	}
 
-	req, err := populateResourceListV1(launchJobFlags.Requests)
```
k82cn: why remove this?

jiangkaihua: this function runs twice to get arg 'req' and 'limit', and the two args are used only when the default job 'test' is started.

k82cn: ok, got that.

k82cn:
please fix ci error and merge commits.

Issue #457, opened 2019-09-19
Podgroup conditions set wrong


hzxuzhonghu:
When i tested MPI, and I used volcano to schedule 3 pods by creating a podgroup. 

After running up, then one pod exited. And there are now only 2 pods running. But i can see the podgroup.status.conditions is wrong.

```
  status:
    conditions:
    - lastTransitionTime: "2019-09-19T04:13:19Z"
      message: '1/2 tasks in gang unschedulable: job is not ready, 2 Running, 3 minAvailable.'
      reason: NotEnoughResources
      status: "True"
      transitionID: d02318a0-da93-11e9-ad67-fadd1cf7d37b
      type: Unschedulable
    phase: Unknown
    running: 2

```

It has already scheduled all the pods, this status is absolutely not what i expect.

PR #465, opened 2019-09-24
Consider PG min resource when enqueue job resource

TommyLike:
Now PG min resource would be considered when calculating the node idle resource.

### DIFF
--- a/pkg/scheduler/actions/enqueue/enqueue.go
+++ b/pkg/scheduler/actions/enqueue/enqueue.go
@@ -48,6 +48,9 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
 
 	jobsMap := map[api.QueueID]*util.PriorityQueue{}
 
+	pgUsedRes := api.EmptyResource()
+	nodeUsedRes := api.EmptyResource()
+
 	for _, job := range ssn.Jobs {
 		if queue, found := ssn.Queues[job.Queue]; !found {
 			glog.Errorf("Failed to find Queue <%s> for Job <%s/%s>",
@@ -61,6 +64,15 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
 				queueMap[queue.UID] = queue
 				queues.Push(queue)
 			}
+		}
+
+		//Reserve pg min resource in advance.
+		if (job.PodGroup.Status.Phase == scheduling.PodGroupInqueue ||
+			job.PodGroup.Status.Phase == scheduling.PodGroupRunning) &&
+			job.PodGroup.Status.Succeeded != job.PodGroup.Spec.MinMember &&
+			job.PodGroup.Spec.MinResources != nil {
+			pgResource := api.NewResource(*job.PodGroup.Spec.MinResources)
+			pgUsedRes.Add(pgResource)
 		}
 
 		if job.PodGroup.Status.Phase == scheduling.PodGroupPending {
@@ -77,7 +89,21 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
 	emptyRes := api.EmptyResource()
 	nodesIdleRes := api.EmptyResource()
 	for _, node := range ssn.Nodes {
-		nodesIdleRes.Add(node.Allocatable.Clone().Multi(1.2).Sub(node.Used))
+		nodesIdleRes.Add(node.Allocatable.Clone().Multi(1.2))
+		nodeUsedRes.Add(node.Used)
+	}
+
+	//deltaRes will equal to the bigger resource consumption
+	deltaRes := api.EmptyResource()
+	if pgUsedRes.Less(nodeUsedRes) {
+		deltaRes = nodeUsedRes
+	} else {
+		deltaRes = pgUsedRes
+	}
+	if deltaRes.Less(nodesIdleRes) {
+		nodesIdleRes.Sub(deltaRes)
+	} else {
+		nodesIdleRes = api.EmptyResource()
 	}
 
 	for {
@@ -85,7 +111,7 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
 			break
 		}
 
-		if nodesIdleRes.Less(emptyRes) {
+		if nodesIdleRes.LessEqual(emptyRes) {
 			glog.V(3).Infof("Node idle resource is overused, ignore it.")
 			break
 		}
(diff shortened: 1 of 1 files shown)


TommyLike:
@k82cn @hzxuzhonghu please take a review

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -63,6 +66,15 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
 			}
 		}
 
+		//Reserve pg min resource in advance.
```
hzxuzhonghu: nit: add a white space after //

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -63,6 +66,15 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
...
+		if (job.PodGroup.Status.Phase == scheduling.PodGroupInqueue ||
+			job.PodGroup.Status.Phase == scheduling.PodGroupRunning) &&
+			job.PodGroup.Status.Succeeded != job.PodGroup.Spec.MinMember &&
+			job.PodGroup.Spec.MinResources != nil {
```
hzxuzhonghu: What if when one pod completed and deleted? FYI: https://github.com/volcano-sh/volcano/issues/457

TommyLike: @hzxuzhonghu what's your suggestion here?

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -77,15 +89,29 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
...
+	deltaRes := api.EmptyResource()
+	if pgUsedRes.Less(nodeUsedRes) {
+		deltaRes = nodeUsedRes
+	} else {
+		deltaRes = pgUsedRes
+	}
+	if deltaRes.Less(nodesIdleRes) {
+		nodesIdleRes.Sub(deltaRes)
+	} else {
+		nodesIdleRes = api.EmptyResource()
```
hzxuzhonghu: Not understand what does this do?

pkg/scheduler/actions/enqueue/enqueue.go
hzxuzhonghu: The pg is running. Why do you add its MinResources?

TommyLike: Considering all of the resources that will / already being used for pod group resources.

pkg/scheduler/actions/enqueue/enqueue.go
TommyLike: use either node used resource or pg used resource that which is greater here. to perform the strict check。

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -63,6 +66,15 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
 			}
 		}
 
+		// Reserve pg min resource in advance.
+		if (job.PodGroup.Status.Phase == scheduling.PodGroupInqueue ||
```
k82cn: let use swithc for `job.PodGroup.Status.Phase`.

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -77,15 +89,29 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
 	emptyRes := api.EmptyResource()
 	nodesIdleRes := api.EmptyResource()
 	for _, node := range ssn.Nodes {
-		nodesIdleRes.Add(node.Allocatable.Clone().Multi(1.2).Sub(node.Used))
+		nodesIdleRes.Add(node.Allocatable.Clone().Multi(1.2))
```
k82cn: 1. the Clone of Allocatable is unnecessary to me.
1. multi 1.2 out of for loop.

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -77,15 +89,29 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
...
+		nodeUsedRes.Add(node.Used)
+	}
+
+	//deltaRes will equal to the bigger resource consumption
+	deltaRes := api.EmptyResource()
+	if pgUsedRes.Less(nodeUsedRes) {
+		deltaRes = nodeUsedRes
+	} else {
+		deltaRes = pgUsedRes
```
k82cn: I don-t think this's correct: `node.Used` resources also include the Pod out of Volcano control.

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -77,15 +89,29 @@ func (enqueue *enqueueAction) Execute(ssn *framework.Session) {
...
+		nodesIdleRes = api.EmptyResource()
 	}
 
 	for {
 		if queues.Empty() {
 			break
 		}
 
-		if nodesIdleRes.Less(emptyRes) {
+		if nodesIdleRes.LessEqual(emptyRes) {
```
k82cn: why not `.IsEmpty()`?

k82cn:
we need e2e test for this fix.

pkg/scheduler/actions/enqueue/enqueue.go
TommyLike: method 'Multi' will change the object itself, Clone required if we put the logic inside of for loop

pkg/scheduler/actions/enqueue/enqueue.go
TommyLike: This is the logic before my changes:
```
		nodesIdleRes.Add(node.Allocatable.Clone().Multi(1.2).Sub(node.Used))
	}

```
what's your suggestion here?

pkg/scheduler/actions/enqueue/enqueue.go
TommyLike: Phase is just part of the conditions here.

TommyLike:
> we need e2e test for this fix.

will do

pkg/scheduler/actions/enqueue/enqueue.go
k82cn: check other condition in the `case`.

PR #466, opened 2019-09-27
Remove invalid job

hzxuzhonghu:


k82cn:
IMO, just comments out the volumes part should be ok; why remove the whole example?

PR #467, opened 2019-09-27
Make controller and admission service optional

TommyLike:
When to be integrated with other service, the controller and admission service could be optional. Let's make it optional.


TommyLike:
@hzxuzhonghu  @k82cn please take a review.

k82cn:
>When to be integrated with other service, the controller and admission service could be optional. Let's make it optional.

We do not provide any quality guarantee or backward compatibility without controller & admission; so we should not enable this in our install script.

PR #472, opened 2019-10-08
Make pod validating in admission service optional

TommyLike:
Make pod validation optional in admission service, for the case of spark operator, when application's podgroup has been initially created, the status would be pending and admission service would block the creation of driver pod which doesn't fit the case.

### DIFF
--- a/cmd/admission/app/options/options.go
+++ b/cmd/admission/app/options/options.go
@@ -50,6 +50,7 @@ type Config struct {
 	AdmissionServiceName      string
 	AdmissionServiceNamespace string
 	SchedulerName             string
+	ValidatePod               bool
 }
 
 // NewConfig create new config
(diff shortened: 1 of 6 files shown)


k82cn:
Can spark-operator retry driver pod creation?

TommyLike:
> Can spark-operator retry driver pod creation?

Yes, there are options for end user to configure how many retries for submit failure, but this would be inconvenient since user has to manually set the retry count and interval. this change can provide another option for user.

k82cn:
> Yes, there are options for end user to configure how many retries for submit failure,

Why ask user to set that? It only need to sent an event or update status.

> this change can provide another option for user.

If accept this change, how to handle the case that driver consumer all resources and not resources for executor to run?

TommyLike:
> > Yes, there are options for end user to configure how many retries for submit failure,
> 
> Why ask user to set that? It only need to sent an event or update status.

I mean if it takes too much time to let running task finishes. the retry count would exceeds.

> 
> > this change can provide another option for user.
> 
> If accept this change, how to handle the case that driver consumer all resources and not resources for executor to run?
the driver will still not get scheduled since the enqueue action will block this driver from creating due to the podgroup resource check.

k82cn:
> I mean if it takes too much time to let running task finishes. the retry count would exceeds.

The job should not in running state as not pod are in running state.

> the driver will still not get scheduled since the enqueue action will block this driver from creating due to the podgroup resource check.

That'll also create lots of driver pod which need to be handled by scheduler.

PR #483, opened 2019-10-15
Implement mpi plugin

hzxuzhonghu:
Add a new plugin `mpi`, with it users do not need to specify other plugins like `ssh` and `svc`, only one `mpi` is enough. 

Also, it add an env `MPI_HOST`, and mount a hostfile in ` /etc/volcano/mpi/hostfile` as  `MPI_HOST_FILE`,.

It will simplify MPI job greatly.

### DIFF
--- a/pkg/admission/admit_job.go
+++ b/pkg/admission/admit_job.go
@@ -146,6 +146,7 @@ func validateJob(job v1alpha1.Job, reviewResponse *v1beta1.AdmissionResponse) st
 			}
 		}
 
+		// TODO: find a way not to hardcode plugin name here
 		if _, found := job.Spec.Plugins["mpi"]; found {
 			if len(job.Spec.Plugins) > 1 {
 				msg = msg + " can not specify other plugins when `mpi` existed"
--- a/pkg/admission/admit_job.go
+++ b/pkg/admission/admit_job.go
@@ -141,8 +141,14 @@ func validateJob(job v1alpha1.Job, reviewResponse *v1beta1.AdmissionResponse) st
 	// invalid job plugins
 	if len(job.Spec.Plugins) != 0 {
 		for name := range job.Spec.Plugins {
-			if _, found := plugins.GetPluginBuilder(name); !found {
+			if P := plugins.GetPluginBuilder(name); P == nil {
 				msg = msg + fmt.Sprintf(" unable to find job plugin: %s", name)
+			}
+		}
+
+		if _, found := job.Spec.Plugins["mpi"]; found {
+			if len(job.Spec.Plugins) > 1 {
+				msg = msg + " can not specify other plugins when `mpi` existed"
 			}
 		}
 	}
--- a/pkg/controllers/job/plugins/factory.go
+++ b/pkg/controllers/job/plugins/factory.go
@@ -17,7 +17,7 @@ limitations under the License.
 package plugins
 
 import (
-	"sync"
+	"volcano.sh/volcano/pkg/controllers/job/plugins/mpi"
 
 	"volcano.sh/volcano/pkg/controllers/job/plugins/env"
 	"volcano.sh/volcano/pkg/controllers/job/plugins/interface"
@@ -29,9 +29,8 @@ func init() {
 	RegisterPluginBuilder("ssh", ssh.New)
 	RegisterPluginBuilder("env", env.New)
 	RegisterPluginBuilder("svc", svc.New)
+	RegisterPluginBuilder("mpi", mpi.New)
 }
-
-var pluginMutex sync.Mutex
 
 // Plugin management
 var pluginBuilders = map[string]PluginBuilder{}
@@ -41,17 +40,10 @@ type PluginBuilder func(pluginsinterface.PluginClientset, []string) pluginsinter
 
 // RegisterPluginBuilder register plugin builders
 func RegisterPluginBuilder(name string, pc func(pluginsinterface.PluginClientset, []string) pluginsinterface.PluginInterface) {
-	pluginMutex.Lock()
-	defer pluginMutex.Unlock()
-
 	pluginBuilders[name] = pc
 }
 
 // GetPluginBuilder returns plugin builder for a given plugin name
-func GetPluginBuilder(name string) (PluginBuilder, bool) {
-	pluginMutex.Lock()
-	defer pluginMutex.Unlock()
-
-	pb, found := pluginBuilders[name]
-	return pb, found
+func GetPluginBuilder(name string) PluginBuilder {
+	return pluginBuilders[name]
 }
--- a/pkg/controllers/job/plugins/factory.go
+++ b/pkg/controllers/job/plugins/factory.go
@@ -17,10 +17,9 @@ limitations under the License.
 package plugins
 
 import (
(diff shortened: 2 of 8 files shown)


example/job.yaml
```diff
@@ -1,41 +0,0 @@
-apiVersion: batch.volcano.sh/v1alpha1
```
k82cn: why delete this example?

pkg/admission/admit_job.go
```diff
@@ -140,10 +140,16 @@ func validateJob(job v1alpha1.Job, reviewResponse *v1beta1.AdmissionResponse) st
 	// invalid job plugins
 	if len(job.Spec.Plugins) != 0 {
 		for name := range job.Spec.Plugins {
-			if _, found := plugins.GetPluginBuilder(name); !found {
+			if P := plugins.GetPluginBuilder(name); P == nil {
 				msg = msg + fmt.Sprintf(" unable to find job plugin: %s", name)
 			}
 		}
+
+		if _, found := job.Spec.Plugins["mpi"]; found {
```
k82cn: we should not check specific plugin here.

pkg/controllers/job/plugins/factory.go
```diff
@@ -29,10 +29,9 @@ func init() {
 	RegisterPluginBuilder("ssh", ssh.New)
 	RegisterPluginBuilder("env", env.New)
 	RegisterPluginBuilder("svc", svc.New)
+	RegisterPluginBuilder("mpi", mpi.New)
 }
 
-var pluginMutex sync.Mutex
```
k82cn: why remove this?

pkg/controllers/job/plugins/factory.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package plugins
 
 import (
-	"sync"
+	"volcano.sh/volcano/pkg/controllers/job/plugins/mpi"
```
k82cn: close to other plugins.

pkg/controllers/job/plugins/mpi/wellknown.go
```diff
@@ -0,0 +1,29 @@
...
+
+package mpi
+
+const (
+	// MPI_HOST env var
+	MPI_HOST = "MPI_HOST"
+
+	// HOST_FILE_PATH and HOST_FILE when used
+	// HOST_FILE_PATH represents mpi hostfile mount path
+	HOST_FILE_PATH = "/etc/mpi"
```
k82cn: should be `/etc/volcano/mpi`

pkg/controllers/job/plugins/mpi/mpi.go
```diff
@@ -0,0 +1,168 @@
...
+type mpi struct {
+	// Arguments given for the plugin
+	pluginArguments []string
+
+	clientset plugininterface.PluginClientset
+
+	ssh plugininterface.PluginInterface
+
+	// mpi launcher index, by default it is 0.
+	launcher int
```
k82cn: why do we need this? we already support multiple pod template.

k82cn:
@hzxuzhonghu , can you give an example on how to use this plugin?

example/job.yaml
hzxuzhonghu: This is done with by mistake

pkg/controllers/job/plugins/factory.go
hzxuzhonghu: unneccessary

pkg/controllers/job/plugins/mpi/mpi.go
hzxuzhonghu: Need this to only set `MPI_HOST` env and mount hostfile to launcher pod

hzxuzhonghu:
Yes,  this still needs an e2e test

pkg/controllers/job/plugins/factory.go
k82cn: please keep this for now.

pkg/controllers/job/plugins/mpi/mpi.go
k82cn: we need to provide info to all pods.

k82cn: similar to svc.

hzxuzhonghu: I donot want to add useless info to mpi workers, unless they are needed

pkg/controllers/job/plugins/factory.go
hzxuzhonghu: For what

pkg/admission/admit_job.go
hzxuzhonghu: Yeah, but we need to prevent this case, otherwise there may be unexpected conflict.

pkg/controllers/job/plugins/factory.go
k82cn: How can you sure we will not use that later? I do not see any benefit, and it may introduce race condition later if anyone use it in multi thread scenario.

pkg/admission/admit_job.go
k82cn: which kind of conflict?

pkg/controllers/job/plugins/mpi/mpi.go
k82cn: refer to `svc` plugin for that.

pkg/admission/admit_job.go
hzxuzhonghu: Think of that we specify ssh and mpi at the same time

pkg/controllers/job/plugins/mpi/mpi.go
hzxuzhonghu: IIUC, mpi worker will never need to connect to the launcher.  A plugin should only add useful info to the container.

pkg/controllers/job/plugins/factory.go
hzxuzhonghu: I can not make sure.  But How does this differ with you add a new struct/interface that maybe used oneday?

hzxuzhonghu:
ping @k82cn

pkg/admission/admit_job.go
k82cn: find some way to do that; we should not highlight plugin name here.

pkg/controllers/job/plugins/factory.go
k82cn: please keep this.

pkg/controllers/job/plugins/mpi/mpi.go
k82cn: please do not have specific setting to mpi.

k82cn: we should not distinguish master/worker here.

hzxuzhonghu: FYI: https://github.com/kubeflow/mpi-operator/blob/815f72098264e1c53743e2cdde3675b207247873/pkg/controllers/v1alpha2/mpi_job_controller.go#L1062-L1149 What mpi operator does is also not mount any host info in the worker.

We should distinguish launcher and worker, as we have to support slots per node as well. IMO, the launcher is the master, it will send commands to workers. But workers will not communicate to the launcher initiative.

PR #484, opened 2019-10-16
fix Resource Less/LessEqual

wangyuqing4:
when i use job.yaml
```
kind: Job
metadata:
  generateName: xxx-
spec:
  minAvailable: 1
  schedulerName: volcano
  tasks:
  - replicas: 10
    name: ta
    template:
      spec:
        containers:
        - image: "nginx:latest"
          name: mpimaster
          resources:
            requests:
              cpu: 5
              memory: 10Gi
            limits:
              cpu: 5
              memory: 10Gi
```
on nodes like this 
```
apiVersion: v1
kind: Node
......
status:
  allocatable:
    cpu: "72"
    ephemeral-storage: "18903225108"
    hugepages-1Gi: "0"
    localdir: 1048600Mi
    memory: 232Gi
    pods: "110"
```
i use kubectl create -f job.yaml, but no pods were scheduled.

pod events:
```
Events:
  Type     Reason         Age                From     Message
  ----     ------         ----               ----     -------
  Warning  Unschedulable  0s (x19 over 18s)  volcano  1/10 tasks in gang unschedulable: job is not ready, 1 minAvailable, 10 Pending.

```

<br />

1. i read func Less/LessEqual very confused, so i edit

2. func Less, if map r.ScalarResources  and rr.ScalarResources are all nil, i think should not compare, should according to cpu and memory.

3. func Less, if map r.ScalarResources nil, rr.ScalarResources is not nil and value <= minMilliScalarResources, i think is not less, should return false. in other words, rr.ScalarResources should large enough if map is not nil.
```
e.g.
ScalarResources request of job is nil, but ScalarResources of node Allocatable is not nil（like hugepages-1Gi: "0"）

in proportion plugin, line 136, if attr.request.Less(attr.deserved) == true {meet}, the deserved of queue can not be larger， the job can not be scheduled any more.
``` 

4. func LessEqual, the same as point 3

### DIFF
--- a/pkg/scheduler/api/resource_info_test.go
+++ b/pkg/scheduler/api/resource_info_test.go
@@ -278,7 +278,7 @@ func TestLessEqual(t *testing.T) {
 				ScalarResources: map[v1.ResourceName]float64{"scalar.test/scalar1": 1},
 			},
 			resource2: &Resource{},
-			expected:  false,
+			expected:  true,
 		},
 		{
 			resource1: &Resource{
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -224,24 +224,38 @@ func (r *Resource) Multi(ratio float64) *Resource {
 
 // Less checks whether a resource is less than other
 func (r *Resource) Less(rr *Resource) bool {
-	if !(r.MilliCPU < rr.MilliCPU && r.Memory < rr.Memory) {
+	lessFunc := func(l, r float64) bool {
+		if l < r {
+			return true
+		}
+		return false
+	}
+
+	if !lessFunc(r.MilliCPU, rr.MilliCPU) {
+		return false
+	}
+	if !lessFunc(r.Memory, rr.Memory) {
 		return false
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/scheduler/api/resource_info_test.go
```diff
@@ -278,7 +278,7 @@ func TestLessEqual(t *testing.T) {
 				ScalarResources: map[v1.ResourceName]float64{"scalar.test/scalar1": 1},
 			},
 			resource2: &Resource{},
-			expected:  false,
+			expected:  true,
```
k82cn: `resource1 < resource2` in this case?

wangyuqing4: yes, resource1.ScalarResources map is not nil, but value <= minMilliScalarResources, so continue, point 4

k82cn: how about the others, e.g. cpu, memory? we need to make sure all resources meet `LessEqual`.

k82cn:
/lgtm

Thanks :)

PR #491, opened 2019-10-28
Control the number of feasible nodes to find and score in scheduling

yuanchen8911:
Added a new feature that allows the scheduler to stop looking for more feasible nodes once it finds a certain number of them. The idea is described in the official document https://kubernetes.io/docs/concepts/scheduling/scheduler-perf-tuning/.

The implementation is similar to the K8S official scheduler https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/core/generic_scheduler.go

The number of feasible nodes is determined by the following parameters. 

- **MinNodesToFind** specifies the minimum number of feasible nodes to find. The default value is 100.
- **PercentageOfNodesToFind** specifies the percentage of feasible nodes to find. The default value is 100%. If the value is less than or equal to zero, an adaptive percentage is calculated according to a simple formula.  
- **MinPercentageOfNodesToFind** specifies the minimum percentage of feasible nodes to find. The default value is 5%.  It ensures that the calculated adaptive percentage is not too small.


cmd/scheduler/app/options/options.go
```diff
@@ -33,6 +33,11 @@ const (
 
 	defaultQPS   = 50.0
 	defaultBurst = 100
+
+    // Default arameters to control the number of feasible nodes to find and score
```
k82cn: s/arameters/parameters ?

k82cn:
/approve

LGTM overall, would you help to fix ci issue? We can get this merged when CI is happy :)

cmd/scheduler/app/options/options.go
yuanchen8911: Fixed the errors. make verify and unit-test passed. Thanks.

PR #507, opened 2019-11-04
fix filter NotReady node 

wangyuqing4:
when openSession get ssn.Node, don't filter notReady node, so calculate totalResource incorrect.

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -126,6 +126,17 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
 		return
 	}
 
+	// If node not ready, e.g. power off
+	for _, cond := range node.Status.Conditions {
+		if cond.Type == v1.NodeReady && cond.Status != v1.ConditionTrue {
+			ni.State = NodeState{
+				Phase:  NotReady,
+				Reason: "NotReady",
+			}
+			return
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/scheduler/api/node_info.go
```diff
@@ -126,6 +128,16 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
 		return
 	}
 
+	var nodeInfo cache.NodeInfo
+	nodeInfo.SetNode(ni.Node)
+	if fit, _, _ := predicates.CheckNodeConditionPredicate(nil, nil, &nodeInfo); !fit {
```
k82cn: How to work together with `TaintNodeByCondition`?

lminzhw: xref: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/taint-node-by-condition.md 
the example at the end of this document may be help.

wangyuqing4: now  in AddPredicateFn we use func CheckNodeConditionPredicate, if conditions have some type and status, like NodeNetworkUnavailable, true, the pod can't be scheduled.

so i move the CheckNodeConditionPredicate for node forward.

wangyuqing4: @k82cn please help to review :)

PR #508, opened 2019-11-04
Add state parameter to queueSpec and queueStatus for queue

sivanzcw:
* In order to subsequently manage queue through the `state` of queue, add `state` parameter to queue struct

* Add valid queue states

* Add update queue logic to queue controller, when queue state changes the `state` of queue can be refreshed


pkg/apis/scheduling/types.go
```diff
@@ -202,12 +202,16 @@ type QueueStatus struct {
...
+	// State is status of queue
+	State string
 }
 
 // QueueSpec represents the template of Queue.
 type QueueSpec struct {
 	Weight     int32
 	Capability v1.ResourceList
+	// State controller the status of queue
+	State string
```
k82cn: s/string/QueueState/g

PR #512, opened 2019-11-06
Add queue controller about state

sivanzcw:
* Add queue action and events

* Add filter function for command watch in `job controller`, only list command whose target is `batch.volcano.sh/v1alpha1 Job`

* Add events recorder for queue controller

* Add command sync during queue sync

* Change queue sync logic, sync queue according to its events and action

### DIFF
--- a/pkg/controllers/queue/queue_controller.go
+++ b/pkg/controllers/queue/queue_controller.go
@@ -18,21 +18,38 @@ package queue
 
 import (
 	"fmt"
-	"reflect"
 	"sync"
+	"time"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
+	busv1alpha1 "volcano.sh/volcano/pkg/apis/bus/v1alpha1"
 	schedulingv1alpha2 "volcano.sh/volcano/pkg/apis/scheduling/v1alpha2"
 	vcclientset "volcano.sh/volcano/pkg/client/clientset/versioned"
+	versionedscheme "volcano.sh/volcano/pkg/client/clientset/versioned/scheme"
 	informerfactory "volcano.sh/volcano/pkg/client/informers/externalversions"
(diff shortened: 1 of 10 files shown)


pkg/controllers/queue/queue_controller_handler.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	// If the `state` value is empty, the status of queue will be set as `Open`
+	// If the `state` value is `Open`, then the status of queue will also be `Open`
+	// If the `state` value is `Closed`, then we need to further consider whether there
+	// is a podgroup under the queue. if there is a podgroup under the queue, the status
+	// of the queue will be set as `Closing`, while if there is no podgroup under the queue,
+	// the status of queue will be set as `Stopped`.
+	queueStatus.State = queue.Spec.State
+	if len(queueStatus.State) == 0 {
```
k82cn: why still update state here?

sivanzcw: For condition of podgroups under the queue will affect the state of the queue, it may determine whether the queue is in the `Closed` state or the `Closing` state, when desired state of queue is `Closed`. So during synchronizing queue, we need to refresh the true `state` of queue. True `state` of queue is `state` in `status`. `state` in `spec` only defines the desired state of queue. The code still follows the logic that `state` of `spec` and condition of podgroup under the queue determine the true `state` of queue, so here to get the true state of the queue according to the `state` of `spec`.

pkg/controllers/queue/queue_controller.go
```diff
@@ -96,22 +137,64 @@ func NewQueueController(
...
 	})
 
+	c.cmdInformer = informerfactory.NewSharedInformerFactory(c.vcClient, 0).Bus().V1alpha1().Commands()
+	c.cmdInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: func(obj interface{}) bool {
+			switch obj.(type) {
+			case *busv1alpha1.Command:
+				cmd := obj.(*busv1alpha1.Command)
+				if cmd.TargetObject != nil &&
+					cmd.TargetObject.APIVersion == schedulingv1alpha2.SchemeGroupVersion.String() &&
```
k82cn: let's create `QueueKind` in apis package and use DeepEqual here.

pkg/controllers/queue/queue_controller_handler.go
k82cn: the queue's state should be update by each `State`, and `queue.Satatus.State` should only copy it.

sivanzcw: updated it

pkg/controllers/queue/queue_controller.go
sivanzcw: Since the `TargetObject` is an instantiated  object that contains the `name` and `uid` of queue reference, it is difficult to use the `DeepEqual` function to compare it with a constant with `QueueKind` type. A filter function is provided to replace the filtering action here.

k82cn:
we need add command line to open/close queue for this :)

sivanzcw:
add it #590

PR #515, opened 2019-11-07
ssh plugin support specifying private/public keys path

hzxuzhonghu:
Added `ssh-key-file-path` to allow users to specify the path of ssh keys. For example if you want to store the keys in `/a/b`, the blowing config can work.

```
kind: Job
metadata:
  name: lm-horovod-job
  labels:
    "volcano.sh/job-type": Horovod
spec:
  minAvailable: 4
  schedulerName: volcano
  plugins:
    ssh: ["--ssh-key-file-path=/a/b"]    # or ["--ssh-key-file-path", "/a/b"]
```

### DIFF
--- a/pkg/controllers/job/plugins/ssh/ssh.go
+++ b/pkg/controllers/job/plugins/ssh/ssh.go
@@ -23,8 +23,6 @@ import (
 	"encoding/pem"
 	"flag"
 	"fmt"
-	"strings"
-
 	"github.com/golang/glog"
 	"golang.org/x/crypto/ssh"
 
@@ -143,7 +141,7 @@ func (sp *sshPlugin) mountRsaKey(pod *v1.Pod, job *batch.Job) {
 
 	for i, c := range pod.Spec.Containers {
 		vm := v1.VolumeMount{
-			MountPath: strings.TrimSuffix(sp.sshKeyFilePath, "/"+SSHRelativePath),
+			MountPath: sp.sshKeyFilePath,
 			SubPath:   SSHRelativePath,
 			Name:      cmName,
 		}
--- a/pkg/controllers/job/plugins/ssh/ssh.go
+++ b/pkg/controllers/job/plugins/ssh/ssh.go
@@ -23,10 +23,10 @@ import (
 	"encoding/pem"
 	"flag"
 	"fmt"
-
-	"golang.org/x/crypto/ssh"
+	"strings"
 
 	"github.com/golang/glog"
+	"golang.org/x/crypto/ssh"
 
 	"k8s.io/api/core/v1"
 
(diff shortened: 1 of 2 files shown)


hzxuzhonghu:
Fixes: https://github.com/volcano-sh/volcano/issues/403

pkg/controllers/job/plugins/ssh/ssh.go
```diff
@@ -23,6 +23,7 @@ import (
 	"encoding/pem"
 	"flag"
 	"fmt"
+	"volcano.sh/volcano/pkg/controllers/job/plugins/env"
```
k82cn: follow import order

pkg/controllers/job/plugins/ssh/ssh.go
```diff
@@ -187,7 +191,10 @@ func (sp *sshPlugin) cmName(job *batch.Job) string {
 
 func (sp *sshPlugin) addFlags() {
 	flagSet := flag.NewFlagSet(sp.Name(), flag.ContinueOnError)
+	// TODO: deprecate no-root
 	flagSet.BoolVar(&sp.noRoot, "no-root", sp.noRoot, "The ssh user, --no-root is common user")
+	flagSet.StringVar(&sp.sshKeyFilePath, "ssh-key-file-path", sp.sshKeyFilePath, "The path used to store "+
+		"ssh private and public keys, by default it is `/root/.ssh`")
```
k82cn: it is `/root/.ssh` by default

k82cn:
please list the value table for `--no-root` and `ssh-key-file-path`.

pkg/controllers/job/plugins/ssh/ssh.go
hzxuzhonghu: ok, seems we should add a ci for this.

hzxuzhonghu:
>please list the value table for --no-root and ssh-key-file-path.

Sorry, i can not catch this.

hzxuzhonghu:
ping @k82cn

dalfos:
@hzxuzhonghu have you example of job how to use ssh & mpi in userspace?

k82cn:
@hzxuzhonghu , please help to list expected result when `--no-root` & `--ssh-key` have different value.

k82cn:
> @hzxuzhonghu have you example of job how to use ssh & mpi in userspace?

@dalfos , a initContainer or script is necessary to copy private/public key into `~/.ssh` to run mpi job in userspace.

dalfos:
@k82cn maybe add param --port $port which adds lines "Port $port" for master and workers in .ssh/config. User cannot start sshd on port 22 (I started for example on port 2222).

dalfos:
> @k82cn maybe add param --port $port which adds lines "Port $port" for master and workers in .ssh/config. User cannot start sshd on port 22 (I started for example on port 2222).

I found in mpirun:
 -mca plm_rsh_args "-p 2222"

hzxuzhonghu:
Yes, we donot care about the port of sshd in volcano.

>@hzxuzhonghu have you example of job how to use ssh & mpi in userspace?

What do u mean by userspace?


>@hzxuzhonghu , please help to list expected result when --no-root & --ssh-key have different value.

if `--ssh-key-file-path`  is specified, `no-root` is not cared about, i want to remove the `no-root` flag in future.

dalfos:
@hzxuzhonghu I mean case then sshd is executed by user (not by root).
All my questions are settled. Thank you!

k82cn:
> if --ssh-key-file-path is specified, no-root is not cared about, i want to remove the no-root flag in future.

ok, so the default value of `--ssh-key-file-path` will be `/root/.ssh` ?

hzxuzhonghu:
Yes

Issue #540, opened 2019-11-20
Remove commit mechanism for pipeline in allocate action


sivanzcw:
/kind feature

Description:

The `allocate` action uses the `commit mechanism`. Only when the job meets the requirements of `gang`, the operations performed in the `allocate` action will be executed. Otherwise, all actions will be rolled back.

The `commit mechanism` in `allocate` action for `pipeline`  action should be canceled:

* When judging whether the job meets the requirements of `gang`, it does not calculate the task with `pipelined` state. That is to say, in the `commit mechanism`, th pipeline action is not considered.

* When the `preempt` action or the `reclaim` action is enabled, in the previous scheduling period, the pod named `poda` with `pending` state successfully evicts a running pod named `podb`, and `poda` was pipelined to the node where the `podb` is running. In the next cycle, when the `allocate` action is performed. Because the job where `poda` is located does not meet the requirements of gang, and cannot perform the pipelined action, in this scheduling period, `poda` is still in the `pending` state, and will evict other pods.

k82cn:
I think we should cancel the pipeline if gang scheduling failed; the releasing resource should be reused by other jobs.

PR #541, opened 2019-11-20
Pipeline task if task's request resource less than the releasing resource of node during performing allocate action

sivanzcw:
#540

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -216,9 +216,9 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 				if task.InitResreq.LessEqual(node.FutureIdle()) {
 					klog.V(3).Infof("Pipelining Task <%v/%v> to node <%v> for <%v> on <%v>",
 						task.Namespace, task.Name, node.Name, task.InitResreq, node.Releasing)
-					if err := stmt.Pipeline(task, node.Name); err != nil {
-						klog.Errorf("Failed to pipeline Task %v on %v",
-							task.UID, node.Name)
+					if err := ssn.Pipeline(task, node.Name); err != nil {
+						klog.Errorf("Failed to pipeline Task %v on %v in Session %v for %v.",
+							task.UID, node.Name, ssn.UID, err)
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -222,9 +222,9 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 				if task.InitResreq.LessEqual(node.Releasing) {
 					glog.V(3).Infof("Pipelining Task <%v/%v> to node <%v> for <%v> on <%v>",
 						task.Namespace, task.Name, node.Name, task.InitResreq, node.Releasing)
-					if err := stmt.Pipeline(task, node.Name); err != nil {
-						glog.Errorf("Failed to pipeline Task %v on %v",
-							task.UID, node.Name)
+					if err := ssn.Pipeline(task, node.Name); err != nil {
```
k82cn: any more info on why change it from `stmt` to `ssn` ?

sivanzcw: If `reclaim` action is enabled after `allocate` action. The cluter situation is as below

| serial   | node name | resource  |
|----------|------------|---------|
| 1        | node1    | 4c8g       |
| 2        | node2    | 4c8g  |

| serial   | queue name | weight  | quota| status
|----------|------------|---------|---|---- |
| 1        | default    | 1       | 0.8c 1.5M | overused |
| 2        | queue1     | 100000  | 7c 8g| active |

| serial   | job name | pods number  |minA | queue|status |
|----------|------------|---------|---|---|---|
| 1        | joba    | 7       | 1|default | all running |
| 2        | jobb     | 7  | 7| queue1 | all pending|

There are two jobs in the cluster, joba and jobb. Joba was placed in `default` queue. Jobb was placed in `queue1` queue. Joba  has 7pods Running. Jobb has 7pods pending. `default` queue was overused. Pods in `queue1` will try to reclaim resource from `defualt` queue. 

* In `reclaim` action, `podb-1` in `jobb` evicted pod `poda-1` in `joba`, `poda-1` was originally at node `node1`,  the scheduling loop ends.

* In the next scheduling loop. In `allocate` action,  `podb-1` want to be pipelined to node `node1`, but gang-restriction of `jobb`  was not meet, the pipeline action will be discard. In `allocate` action no pod was pipelined, though there are releasing resources in cluster.

* In `relcaim` action of this scheduling loop, `podb-1` in `jobb` will try to evicted other pods in `joba`.

* Finally, `podb-1` will evicted 6 pods from `joba`.

* So if there are releasing resources in cluster, pod who has the higher priority, may should be `pipelined` to the node, regardless of whether  the gang restriction of job of the pod was meet, in case that, the pod will evict other pods in subsequent actions.

PR #542, opened 2019-11-20
modify the 'vcctl job run' function

jiangkaihua:
now `vcctl job run` can used as`vcctl job run -N <job name> -n <namespace> -i <image>`
and `vcctl job run` with no arguments will throw error


pkg/cli/job/run.go
```diff
@@ -72,14 +72,23 @@ func RunJob() error {
...
+
 	job, err := readFile(launchJobFlags.FileName)
 	if err != nil {
 		return err
 	}
 
-	if job == nil {
-		fmt.Printf("Error: job script (specified by --filename or -f) is mandatory to run a particular job")
-		return nil
+	if job == nil && launchJobFlags.Name != "" {
```
k82cn: if job name is nil, we should return error here.

PR #544, opened 2019-11-21
Simplify job pvc create process

hzxuzhonghu:
This is to simplify `createJobIOIfNotExist` and fix some unreasonable logic, like `job.Status.ControlledResources["volume-emptyDir-"+vcName] = vcName`.

### DIFF
--- a/test/e2e/job_controlled_resource.go
+++ b/test/e2e/job_controlled_resource.go
@@ -26,47 +26,6 @@ import (
 )
 
 var _ = Describe("Job E2E Test: Test Job PVCs", func() {
-	It("Generate PVC name if not specified", func() {
-		jobName := "job-pvc-name-empty"
-		namespace := "test"
-		taskName := "task"
-		pvcName := "specifiedpvcname"
-		context := initTestContext()
-		defer cleanupTestContext(context)
-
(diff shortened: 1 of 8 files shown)


test/e2e/job_controlled_resource.go
```diff
@@ -26,47 +26,6 @@ import (
 )
 
 var _ = Describe("Job E2E Test: Test Job PVCs", func() {
-	It("Generate PVC name if not specified", func() {
```
k82cn: why delete this case?

hzxuzhonghu: I changed the behavior: previously if only `MountPath` defined under  `JobSpec.Volumes`, job controller will create an emptydir pvc for it.  This is unreasonable behavior, emptydir is not suggested to be used in production.  Also volcano users will not use it, if they really need , they should declare it explicitly. We don't want to set too many default values for it.

hzxuzhonghu: And i will update the api doc in a following pr.

k82cn: > emptydir is not suggested to be used in production

by k8s community?

PR #545, opened 2019-11-21
Added dedicated volume design

hzxuzhonghu:


hzxuzhonghu:
@k82cn @zrss take a look if you have bandwidth

docs/design/exclusive-volume.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Motivation
+
+Volume mount was supported from begin. But there are few limitations:
+
+1. We can specify volumes of Job scope, all the volumes are shared by all pods within a Job.
+
+2. We specify volumes by setting `TaskSpec.PodTemplateSpec.Volumes`, but similarly they are shared by pods within a task.
+
+But in real world, scenarios like DL/BigData, etc requires high performance. Shared storage has some performance issue,
```
k82cn: I think that's not the motivation of this proposal :)

docs/design/exclusive-volume.md
```diff
@@ -0,0 +1,126 @@
...
+
+Job TaskSpec will contain an additional structure called `VolumeClaimTemplates` to control exclusive volumes.
+
+## Proposal
+
+### Implementation Details
+
+#### API Changes
+
+Following changes will be made to the TaskSpec for volcano Job.
```
k82cn: what changes are made?

docs/design/exclusive-volume.md
```diff
@@ -0,0 +1,126 @@
...
+    // +optional
+    VolumeClaimTemplates []v1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty" protobuf:"bytes,5,rep,name=volumeClaimTemplates"`
+}
+```
+
+- By default, this is empty. Will use that defined in `JobSpec` and `TaskSpec.Template`.
+
+- If `VolumeClaimTemplates` specified, these pvcs are referenced by all the pods of the task. But the persistent volume claim name is rebuilt by job controller. In a word, pods will not share any of the persistent volume claims. 
+
+- If the rebuilt persistent does not exist, job controller will create it.
```
k82cn: how about the lifecycle of pvc? will job controller delete it?

hzxuzhonghu: Common as other children resource a job owns, i donot want it being different.   But the pv it binds can be retained by reclaim policy in  storage class.

docs/design/exclusive-volume.md
hzxuzhonghu: I will add

docs/design/exclusive-volume.md
hzxuzhonghu: I am not sure what others ?

docs/design/exclusive-volume.md
hzxuzhonghu: Meybe we should handle the pvc differently as it is stateful app.

hzxuzhonghu: Hmm, it will look weird, inconsistent. as there are pvcs in job spec and pod spec, which are all deleted as volcano job deleted.

hzxuzhonghu:
About the pvc reclaim policy, i suggest we keep consistent with the current defined behavior. And we can change it in another pr.

docs/design/exclusive-volume.md
k82cn: please talk with zrss for more detail about that.

k82cn:
> About the pvc reclaim policy, i suggest we keep consistent with the current defined behavior. And we can change it in another pr.

We need to talk with zrss about target behaviour, and keep compatibility with our previous releases.

zrss:
thx @k82cn @hzxuzhonghu 

ref the https://github.com/kubernetes/kubernetes/issues/30606#issuecomment-240306750

> if retain PV if needed, just don't delete PVC.

i'd like there is an option of vj pvc volume for keeping the pvc when the vj job done

zrss:
btw, exclusize ... maybe dedicated volume is good to know

docs/design/exclusive-volume.md
hzxuzhonghu: @zrss Here we need your input

zrss:
anything new on this topic ?

docs/design/dedicated-volume.md
```diff
@@ -0,0 +1,115 @@
...
+      spec:
+        containers:
+        - name: ps
+          image: ps-img
+          volumeMounts:
+          - mountPath: /data
+            name: train-vol
+    volumeClaimTemplates:
+    - metadata:
+        name: train-vol
```
k82cn: if using `GenerateName`, we'll generate pvc for each pods. if using `name`, only one pvc for all pods.

k82cn:
lgtm overall :)

docs/design/dedicated-volume.md
k82cn: @hzxuzhonghu , is that ok for you?

hzxuzhonghu: Agree, i will update to explicitly indicate it.

PR #552, opened 2019-11-26
Support networkpolicy

hzxuzhonghu:
This is for the sake of the security,  added networkpolicy to allow access from only groups of pods belong to the same job and prevent access from external.

**Note: networkpolicy depends on CNI plugin, only takes effect when cni supports it.**

### DIFF
--- a/pkg/controllers/job/plugins/svc/svc.go
+++ b/pkg/controllers/job/plugins/svc/svc.go
@@ -24,6 +24,7 @@ import (
 	"github.com/golang/glog"
 
 	"k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
I will add test later if this is feasible.

k82cn:
LGTM overall

pkg/controllers/job/plugins/svc/svc.go
```diff
@@ -97,6 +98,11 @@ func (sp *servicePlugin) OnJobAdd(job *batch.Job) error {
 		return err
 	}
 
+	// TODO: maybe add a flag
```
dalfos: @hzxuzhonghu can you pls add flag for this?

hzxuzhonghu: @dalfos Yeah, anyway i want to know your use case.

dalfos: @hzxuzhonghu I already implemented it through calicoctl I want to avoid conflicts

k82cn: hmm... let's add a flag to this plugin :)

dalfos:
Hi @hzxuzhonghu and @k82cn should I open issue for discussed flag?

k82cn:
> Hi @hzxuzhonghu and @k82cn should I open issue for discussed flag?

That's ok to me, please go ahead to open an issue/pr for that :)

PR #583, opened 2019-12-03
Fix preemptableFn of gang plugin

zionwu:
Fix the same issue mentioned in https://github.com/kubernetes-sigs/kube-batch/issues/920


carmark:
any update?

PR #585, opened 2019-12-05
use future idle resources when checking if task can fit node

### LINKED ISSUE #582: node idle resources not considered in preemption
/kind bug

**What happened**:
There is a pod A (memory: 1GB) running on a node (memory: 2GB), then I scheduled a pod B (memory: 2GB). Pod A was not preempted, because sum of victim resources (1GB) is lower than pod B resource requirements.

Pod A being preempted, because idle resources (1GB) plus victims resources (1GB) is enough to fit pod B (2GB).

Modify `TestPreempt` in `pkg/scheduler/actions/preempt/preempt_test.go`.

I do not know if it is intentional, but node idle resources are **ignored** when making preemption decisions. I would expect that we want to check if resources from preempting victims **plus** node idle resources are enough to fit a preemptor.

I noticed that this change might be related to https://github.com/kubernetes-sigs/kube-batch/pull/721/files#diff-b944f6bc39bea34c3a8e145c97c5fc91R75. 

Mentioned in https://github.com/kubernetes-sigs/kube-batch/issues/911

mateuszlitwin:
Future idle resources (current idle resources + resources being released) are considered for pipelining during allocate and preempt action.

Fixes a bug when task was used as preemptor on the first iteration of the preempt loop despite task's job being "pipelined".
Fixes issue #582


k82cn:
/lgtm
/approve

LGTM overall :)

PR #591, opened 2019-12-06
disp job in default queue

jiangkaihua:
if a yaml file do not assign a specified queue for job, then the job will be assigned to default queue.


k82cn:
please help to fix CI issue :)

Issue #308, opened 2019-07-09
Please leave a comment if you're using or evaluating Volcano :)


k82cn:
Please leave a comment if you're using Volcano in following format:

```
Organization: xxxx
Environment: Production/Evaluation
Contact: xxxxxx
Description of Use:  xxxxxx
```

or create a PR to add yourself into adopter list at https://github.com/volcano-sh/volcano/blob/master/docs/community/adopters.md :)

xieydd:
Organization: Unisound
Environment: Evaluation
Contact: [xieydd](https://github.com/xieydd)
Description of Use:  Evaluation in ATLAS AI Plateform

felix5572:
Organization:Peking University, Beijing Institute of Big Data Research.
Environment: Evaluation
Contact: me.   @felix5572
Description of Use: scientific calculations in physics, materials , biology and chemistry. molecular dynamics simulation.

gaocegege:
Organization: Caicloud
Environment: Evaluation
Contact: @gaocegege 
Description of Use:  Scheduler for Distributed DL training jobs

alartin:
Organization: GrandOmics
Environment: Evaluation
Contact: @alartin
Description of Use:  Infrastructure of Hanwell (Huawei Cloud backend of Cromwell which is a Broad Institute implementation of WDL)

tsjsdbd:
Organization:  [Huawei Cloud](https://huaweicloud.com/) 
Environment: **Production** 
Contact: [@tsjsdbd](https://github.com/tsjsdbd)
Description of Use:  Scheduler & Job Management of AI Container Service and CCI

yuanchen8911:
Organization: JD Retail Infrastructure Department
Environment: Evaluation
Contact: @yuanchen8911 
Description of Use: Spark on K8S

minyk:
Organization: kt NexR
Environment: Evaluation
Contact: @minyk, @dieselnexr
Description of Use: spark scheduler of our next cloud native product.

yuzhaojing:
Organization: QTT Bigdata Infra
Environment: Evaluation
Contact: @yuzhaojing
Description of Use: Spark and AI on K8S.

zionwu:
Organization:  vivo
Environment: Evaluation
Contact:  @zionwu 
Description of Use:  Scheduler for Vtraining platform (deep learning training platform)

xiaogaozi:
Organization: Xiaohongshu
Environment: Evaluation
[...]

PR #592, opened 2019-12-07
Add Vivo as adopter

k82cn:
Add Vivo as adopter


k82cn:
xref https://github.com/volcano-sh/volcano/issues/308#issuecomment-562558894

PR #601, opened 2019-12-11
Added localup script.

k82cn:


hzxuzhonghu:
Is it worth building k8s binaries from code and start up the components one by one in cmd line? Suggest using the existing tools kubeadm or minikube.

k82cn:
submit first version, we'll enhance it with more detail.

PR #606, opened 2019-12-14
Update webhook path.

k82cn:


hzxuzhonghu:
Generally, this is a bad idea with backward compatibility issue.

PR #610, opened 2019-12-15
Added hosts into environment.

### LINKED ISSUE #260: Add hosts list into environment

/kind feature

**Description**:

In `svc` plugin, we only provide host files in `/etc/volcano`, and user has to translate those files in to a list of hosts as arguments. It's better to include such an env in all pods of job.

k82cn:

fixed #260 

```
Added hosts into pod's environment.
```

### DIFF
--- a/pkg/controllers/job/plugins/svc/const.go
+++ b/pkg/controllers/job/plugins/svc/const.go
@@ -19,6 +19,10 @@ package svc
 const (
 	// ConfigMapTaskHostFmt key in config map
 	ConfigMapTaskHostFmt = "%s.host"
+	// EnvTaskHostFmt is the key for host list in environment
+	EnvTaskHostFmt = "VC_%s_HOSTS"
+	// EnvHostNumFmt is the key for host number in environment
+	EnvHostNumFmt = "VC_%s_NUM"
 
 	// ConfigMapMountPath mount path
 	ConfigMapMountPath = "/etc/volcano"
--- a/pkg/controllers/job/plugins/svc/svc.go
+++ b/pkg/controllers/job/plugins/svc/svc.go
@@ -19,6 +19,7 @@ package svc
 import (
 	"flag"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"k8s.io/klog"
@@ -88,6 +89,34 @@ func (sp *servicePlugin) OnPodCreate(pod *v1.Pod, job *batch.Job) error {
 		pod.Spec.Subdomain = job.Name
 	}
 
+	var hostEnv []v1.EnvVar
+	var envNames []string
+
+	for _, ts := range job.Spec.Tasks {
+		// TODO(k82cn): The splitter and the prefix of env should be configurable.
+		envNames = append(envNames, fmt.Sprintf(EnvTaskHostFmt, strings.ToUpper(ts.Name)))
+		envNames = append(envNames, fmt.Sprintf(EnvHostNumFmt, strings.ToUpper(ts.Name)))
+	}
+
+	for _, name := range envNames {
+		hostEnv = append(hostEnv, v1.EnvVar{
+			Name: name,
+			ValueFrom: &v1.EnvVarSource{
+				ConfigMapKeyRef: &v1.ConfigMapKeySelector{
+					LocalObjectReference: v1.LocalObjectReference{Name: sp.cmName(job)},
+					Key:                  name,
+				}}},
+		)
+	}
+
+	for i := range pod.Spec.Containers {
+		pod.Spec.Containers[i].Env = append(pod.Spec.Containers[i].Env, hostEnv...)
+	}
+
+	for i := range pod.Spec.InitContainers {
+		pod.Spec.InitContainers[i].Env = append(pod.Spec.InitContainers[i].Env, hostEnv...)
+	}
+
 	sp.mountConfigmap(pod, job)
 
 	return nil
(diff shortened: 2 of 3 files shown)


zrss:
generally, AI workload also need the `current_host` info, it would be better if vj provides the `current_host` env too

zrss:
by the way, from this PR, in my point of view, `the env plugin` should change to a more suitable name ...

k82cn:
> by the way, from this PR, in my point of view, the env plugin should change to a more suitable name ...

`svc` plugin will handle hosts related issue, including environment. `env` plugin will help to handle others :)

k82cn:
> generally, AI workload also need the current_host info, it would be better if vj provides the current_host env too

what does urrent_host` info mean? you can get hostname in container by POSIX api, e.g. hostname.

zrss:
> > generally, AI workload also need the current_host info, it would be better if vj provides the current_host env too
> 
> what does urrent_host` info mean? you can get hostname in container by POSIX api, e.g. hostname.

`hostname` is the podName, `VC_CURRENT_HOST` = the headless svc of current pod, and may be in this format `[VC-Job-Name]-[VC-Task-Name]-[VC-Task-Index].[VC-Job-Name]`

hzxuzhonghu:
>VC_CURRENT_HOST = the headless svc of current pod, and may be in this format [VC-Job-Name]-[VC-Task-Name]-[VC-Task-Index].[VC-Job-Name]

Is this a  common requirement, I am worried this is just for TF job

zrss:
> > VC_CURRENT_HOST = the headless svc of current pod, and may be in this format [VC-Job-Name]-[VC-Task-Name]-[VC-Task-Index].[VC-Job-Name]
> 
> Is this a common requirement, I am worried this is just for TF job

hostname or podname can work well in the container network, but it doesn't work in the host network mode, but the headless svc work

k82cn:
hm.. maybe we can support `podName [VC-Job-Name]-[VC-Task-Name]-[VC-Task-Index].[VC-Job-Name]` style in hostfile.

k82cn:
just checked code here: https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/plugins/svc/svc.go#L274 . It seems we already support that; so I'd like to support host network in other PR :)

hzxuzhonghu:
I am still not sure what do you mean by hostnetwork. IIRC, the hostnetwork mode has nothing special for headless service.

k82cn:
> I am still not sure what do you mean by hostnetwork. IIRC, the hostnetwork mode has nothing special for headless service.

If so, please work with @zrss on this.

hzxuzhonghu:
@zrss please clarify, we need a little bit more. Not just an issue without any context

pkg/controllers/job/plugins/svc/svc.go
```diff
@@ -258,8 +279,9 @@ func (sp *servicePlugin) cmName(job *batch.Job) string {
 	return fmt.Sprintf("%s-%s", job.Name, sp.Name())
 }
 
-func generateHost(job *batch.Job) map[string]string {
-	data := make(map[string]string, len(job.Spec.Tasks))
+func generateHosts(job *batch.Job) (hostFile map[string]string, hostEnv map[string]string) {
+	hostFile = make(map[string]string, len(job.Spec.Tasks))
+	hostEnv = make(map[string]string, len(job.Spec.Tasks))
```
hzxuzhonghu: Where do you set it

k82cn:
> @zrss please clarify, we need a little bit more. Not just an issue without any context

@hzxuzhonghu , please start another thread for this discussion.

pkg/controllers/job/plugins/svc/svc.go
```diff
@@ -88,6 +89,34 @@ func (sp *servicePlugin) OnPodCreate(pod *v1.Pod, job *batch.Job) error {
...
+	for _, ts := range job.Spec.Tasks {
+		// TODO(k82cn): The splitter and the prefix of env should be configurable.
+		envNames = append(envNames, fmt.Sprintf(EnvTaskHostFmt, strings.ToUpper(ts.Name)))
+		envNames = append(envNames, fmt.Sprintf(EnvHostNumFmt, strings.ToUpper(ts.Name)))
+	}
+
+	for _, name := range envNames {
+		hostEnv = append(hostEnv, v1.EnvVar{
+			Name: name,
+			ValueFrom: &v1.EnvVarSource{
```
hzxuzhonghu: Downward api is used when we can not get the value directly, and is it not that efficient. Here can set the value directly.

pkg/controllers/job/plugins/svc/const.go
```diff
@@ -19,6 +19,10 @@ package svc
 const (
 	// ConfigMapTaskHostFmt key in config map
 	ConfigMapTaskHostFmt = "%s.host"
+	// EnvTaskHostFmt is the key for host list in environment
+	EnvTaskHostFmt = "VC_%s_HOSTS"
+	// EnvHostNumFmt is the key for host number in environment
+	EnvHostNumFmt = "VC_%s_NUM"
```
hzxuzhonghu: what's the use case for the num.

pkg/controllers/job/plugins/svc/const.go
```diff
@@ -19,6 +19,10 @@ package svc
 const (
 	// ConfigMapTaskHostFmt key in config map
 	ConfigMapTaskHostFmt = "%s.host"
+	// EnvTaskHostFmt is the key for host list in environment
+	EnvTaskHostFmt = "VC_%s_HOSTS"
```
hzxuzhonghu: I suggest we change to volcano full name from usability view. Abbreviate is hard to remember.

k82cn: no, let's keep `VC_`.

hzxuzhonghu: ok, you are the boss you decide.

PR #614, opened 2019-12-16
Update factory.go

YesterdayxD:
replace parameter type


pkg/controllers/job/plugins/factory.go
```diff
@@ -40,7 +40,7 @@ var pluginBuilders = map[string]PluginBuilder{}
 type PluginBuilder func(pluginsinterface.PluginClientset, []string) pluginsinterface.PluginInterface
 
 // RegisterPluginBuilder register plugin builders
-func RegisterPluginBuilder(name string, pc func(pluginsinterface.PluginClientset, []string) pluginsinterface.PluginInterface) {
+func RegisterPluginBuilder(name string, PluginBuilder) {
```
hzxuzhonghu: i think you need a param

YesterdayxD: yes

PR #616, opened 2019-12-16
Retain job's resources after job finished successfully or failed

hzxuzhonghu:


k82cn:
any backgroup of this PR?

hzxuzhonghu:
Users are unable to see the output, as the pods removed.

k82cn:
> Users are unable to see the output, as the pods removed.

so what's your solution? any pros/cons? It's really a hard work to review code directly.

hzxuzhonghu:
The most urgent requirement is : 1. retain the workload log, we have no way except keeping the pod after job finished(no matter successfully or failed). 2. retain the pvc,  for this we need to add an option in API, which should be a separate pr.

Issue #619, opened 2019-12-17
Propose to set `VK_TASK_INDEX` env by default


hzxuzhonghu:
It is costly to use an env plugin to set the task index for a job.  We can set `VK_TASK_INDEX` or use the full name `VOLCANO_TASK_INDEX` by default.

hzxuzhonghu:
This not make sense

PR #620, opened 2019-12-17
Added VOLCANO_TASK_INDEX env, and remove the env plugin

hzxuzhonghu:
1. Added `VOLCANO_TASK_INDEX`, deprecated `VK_TASK_INDEX`, so in the future recommend using the new env instead.
2.  removed the env plugin, volcano will set the env by default, which reduced the complexity specifying `env` plugin.

Fixes: #619


k82cn:
let's keep `VC_` as reserved key word, `VOLCANO_` is too long for user.

pkg/apis/batch/v1alpha1/labels.go
```diff
@@ -31,4 +31,12 @@ const (
...
 	// PodgroupNamePrefix podgroup name prefix
 	PodgroupNamePrefix = "podgroup-"
+
+	// Env vars
+
+	// LegacyTaskIndex indicates the index of a group of replicas of a task.
+	// Deprecated, remove it in release 0.4, use the TaskIndex instead.
+	LegacyTaskIndex = "VK_TASK_INDEX"
+	// TaskIndex indicates the index of a group of replicas of a task.
+	TaskIndex = "VOLCANO_TASK_INDEX"
```
k82cn: let's keep `VC_` :)

k82cn:
/approve

Please help to address the comments, and them let's merge this PR :)

k82cn:
btw, prefer to enable `svc` and `env` by default instead of removing `env`, as we will add more env vars for a job which is better to isolate into a plugin :)

hzxuzhonghu:
I agree we can set svc or envs by default for a job. But basically i donot  think it is meaningful to keep these plugins.  
Think about how to disable these plugins? For end users, how would they know which plugin to use in order to run a `XXX` framework training job?  Different type of jobs may have different requirements.  Though we can provide as many as possible, but personally i donot think it is secure.
IMO, the eventual target is to provide plugin per framwork. This can really improve UX.

k82cn:
what's different with this PR?

hzxuzhonghu:
i can close it if this makes no sense.

k82cn:
let's add a parameters on which plugins are default.

hzxuzhonghu:
I agree this maybe a good choice.

PR #627, opened 2019-12-19
remove repeated code.

YesterdayxD:
Line 56-58 has same effect.

### DIFF
--- a/pkg/controllers/job/job_controller_util.go
+++ b/pkg/controllers/job/job_controller_util.go
@@ -108,11 +108,6 @@ func createJobPod(job *batch.Job, template *v1.PodTemplateSpec, ix int) *v1.Pod 
 	pod.Labels[batch.JobNameKey] = job.Name
 	pod.Labels[batch.JobNamespaceKey] = job.Namespace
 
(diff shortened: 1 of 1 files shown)


YesterdayxD:
or 
```
if pod.Spec.SchedulerName != job.Spec.SchedulerName {
		pod.Spec.SchedulerName = job.Spec.SchedulerName
	}
```
aim to make the name
@hzxuzhonghu

hzxuzhonghu:
@YesterdayxD Good catch， i prefer https://github.com/volcano-sh/volcano/pull/627#issuecomment-567386295

hzxuzhonghu:
And  one step further, i would like to deprecate the `JobSpec.SchedulerName`, because for volcano, it is meaningless to allow specifying a different scheduler name.

k82cn:
> i would like to deprecate the `JobSpec.SchedulerName`

Please do not do that, that'll make user set schedulerName for each pod-template. We can validate it by admission controller.

YesterdayxD:
Please add lgtm label.

k82cn:
@hzxuzhonghu , is that ok for you?

hzxuzhonghu:
yeah 
/lgtm

PR #633, opened 2019-12-20
fix the the Getting started link in contribute.md

ruiyinchen:
fix the the  Getting started link

### DIFF
--- a/contribute.md
+++ b/contribute.md
@@ -31,8 +31,8 @@ The goal of the community is to develop a volcano system which is useful for run
 
 # Getting started
 
-- Read the [get started](docs/development/perepare-for-development.md) for developing code for Volcano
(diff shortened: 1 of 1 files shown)


contribute.md
```diff
@@ -31,8 +31,8 @@ The goal of the community is to develop a volcano system which is useful for run
 
 # Getting started
 
-- Read the [get started](docs/development/perepare-for-development.md) for developing code for Volcano
-- Read the [setup](docs/development/development.md) for build/deploy instructions.
+- Read the [get started](https://github.com/volcano-sh/volcano/blob/master/docs/development/prepare-for-development.md) for developing code for Volcano
```
k82cn: prefer to keep relative path as we may publish those doc by branch.

Issue #621, opened 2019-12-17
Pod creating is denied by admission with `podgroup phase is Pending`


hzxuzhonghu:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind bug

**What happened**:

When job controller try to create pods, it is denied by admission controller with `as the podgroup phase is Pending`.

This is caused by the podgroup's status phase is set by scheduler and the scheduler is running periodically. There must be a delay before the podgroup `InQueue`.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hzxuzhonghu:
E1216 12:31:28.678835       1 job_controller_actions.go:252] Failed to create pod failed-delete-job-fail-0 for Job failed-delete-job, err &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:""}, Status:"Failure", Message:"admission webhook \"validatepod.volcano.sh\" denied the request: failed to create pod <test/failed-delete-job-fail-0> as the podgroup phase is Pending", Reason:"", Details:(*v1.StatusDetails)(nil), Code:400}}

PR #634, opened 2019-12-23
Do not create jobs until pg inqueue

hzxuzhonghu:
Fixes: #621

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -190,7 +190,7 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 	}
 
 	// Skip job initiation if job is already accepted
-	if job.Status.State.Phase == "" {
+	if job.Status.State.Phase == "" || job.Status.State.Phase == batch.Pending || job.Status.State.Phase == batch.Restarting {
 		var err error
 		if job, err = cc.initiateJob(job); err != nil {
 			return err
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -189,13 +189,14 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 		return nil
 	}
 
-	// Skip job initiation if job is already accepted
-	if job.Status.State.Phase == "" || job.Status.State.Phase == batch.Pending || job.Status.State.Phase == batch.Restarting {
+	// Skip job initiation if job is already initiated
+	if !isInitiated(job) {
 		var err error
 		if job, err = cc.initiateJob(job); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -185,9 +189,46 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
...
+		if e := cc.cache.Update(newJob); e != nil {
+			klog.Errorf("SyncJob - Failed to update Job %v/%v in cache:  %v",
+				newJob.Namespace, newJob.Name, e)
+			return e
+		}
+		return nil
+	}
+
+	// Skip job task sync if it is pending
+	if job.Status.State.Phase == batch.Pending {
```
k82cn: If return, when do we check it again?

hzxuzhonghu: when receives pg update event

hzxuzhonghu:
@k82cn Please let me know if this make sense to you, so i can decide if i should work on fixing ci. Otherwise, i can close it.

hzxuzhonghu:
I am pretty sure the e2e is flaky

pkg/controllers/job/job_controller_actions.go
```diff
@@ -185,9 +189,38 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 		return nil
 	}
 
-	var err error
-	if job, err = cc.createJob(job); err != nil {
-		return err
+	// Skip job initiation if job is already accepted
+	if job.Status.State.Phase == "" || job.Status.State.Phase == batch.Pending || job.Status.State.Phase == batch.Restarting {
```
k82cn: add `isInitialized` func for this check.

pkg/controllers/job/job_controller_actions.go
```diff
@@ -185,9 +189,38 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
...
+		var err error
+		if job, err = cc.initiateJob(job); err != nil {
+			return err
+		}
+	}
+	var syncTask bool
+	if pg, _ := cc.pgLister.PodGroups(job.Namespace).Get(job.Name); pg != nil {
```
k82cn: can we check this in handler?

k82cn:
LGTM overall :)

/approve

pkg/controllers/job/job_controller_actions.go
hzxuzhonghu: done

pkg/controllers/job/job_controller_actions.go
hzxuzhonghu: I am not sure i am understanding what you mean. My thought is to split the SyncJob function: the first phase is initialization and the second phase is sync.

PR #638, opened 2019-12-25
Remove duplicated check in jobEnqueueableFn of proportion

zionwu:
In jobEnqueueableFn of proportion plugin, the following check is duplicated:
```
    	if len(queue.Queue.Spec.Capability) == 0 
```

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -219,15 +219,12 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		// If no capability is set, always enqueue the job.
 		if len(queue.Queue.Spec.Capability) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -218,16 +218,13 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		queue := ssn.Queues[queueID]
 
 		// If no capability is set, always enqueue the job.
-		if len(queue.Queue.Spec.Capability) == 0 {
-			return true
-		}
-
-		pgResource := api.NewResource(*job.PodGroup.Spec.MinResources)
 		if len(queue.Queue.Spec.Capability) == 0 {
```
k82cn: and it seems we did not add capability into crds :(

Issue #646, opened 2019-12-26
The best-effort pod should be immediately dispatched in allocate function of session , without considering the gang constraint


sivanzcw:
/kind bug

Considering the gang constraint in the `backfill` action may cause the job with gang configuration and contains the best-effort pod  to be unscheduled.

In `backfill` action, if task predicated succeed, the scheduler will call `ssn.Allocate` function to dispatch pod

```
for _, node := range ssn.Nodes {
	// TODO (k82cn): predicates did not consider pod number for now, there'll
	// be ping-pong case here.
	if err := ssn.PredicateFn(task, node); err != nil {
		klog.V(3).Infof("Predicates failed for task <%s/%s> on node <%s>: %v",
			task.Namespace, task.Name, node.Name, err)
		fe.SetNodeError(node.Name, err)
		continue
	}

	klog.V(3).Infof("Binding Task <%v/%v> to node <%v>", task.Namespace, task.Name, node.Name)
	if err := ssn.Allocate(task, node.Name); err != nil {
		klog.Errorf("Failed to bind Task %v on %v in Session %v", task.UID, node.Name, ssn.UID)
		fe.SetNodeError(node.Name, err)
		continue
	}

	allocated = true
	break
}
``` 

In `ssn.Allocate` function, only when job of task is ready, can the task be dispatched

```
if ssn.JobReady(job) {
	for _, task := range job.TaskStatusIndex[api.Allocated] {
		if err := ssn.dispatch(task); err != nil {
			klog.Errorf("Failed to dispatch task <%v/%v>: %v",
				task.Namespace, task.Name, err)
			return err
		}
	}
}
```

If others pods under the job are not schedulerd, the best-effort pod can not be scheduled. If all pods under the job are best-effort, the job will not be scheduled.

In addition, for best-effort pods, there are no resource requests, it it not necessary to consider the gang constraint.

carmark:
This may make the MPIJob pending when the launcher does not config resource requests/limits.

sivanzcw:
closed by #647

PR #647, opened 2019-12-26
Considering best-effort pods when calculating ready task number

sivanzcw:
#646

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -343,17 +343,26 @@ func (ji *JobInfo) FitError() string {
 	return reasonMsg
 }
 
-// ReadyTaskNum returns the number of tasks that are ready.
+// ReadyTaskNum returns the number of tasks that are ready or that is best-effort.
 func (ji *JobInfo) ReadyTaskNum() int32 {
-	occupid := 0
+	var occupied int32
 	for status, tasks := range ji.TaskStatusIndex {
 		if AllocatedStatus(status) ||
 			status == Succeeded {
-			occupid = occupid + len(tasks)
+			occupied += int32(len(tasks))
+			continue
+		}
+
+		if status == Pending {
+			for _, task := range tasks {
+				if task.InitResreq.IsEmpty() {
+					occupied++
+				}
+			}
 		}
 	}
 
-	return int32(occupid)
+	return occupied
 }
 
 // WaitingTaskNum returns the number of tasks that are pipelined.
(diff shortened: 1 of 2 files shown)


pkg/scheduler/framework/session.go
```diff
@@ -297,6 +297,15 @@ func (ssn *Session) Allocate(task *api.TaskInfo, hostname string) error {
 				return err
 			}
 		}
+	} else {
+		// For best-effort pod, it it not necessary to consider the gang constraint.
+		if task.InitResreq.IsEmpty() {
```
k82cn: what's happen if there not enough resources for the other Pods?

sivanzcw: If there are six pods in a job, and the `minA` is `6`, two of the the six pods are best-effort pods, and the others are non-best-effort pods. The best-effort pods will be dispatched succeed and refreshed to `running` while the non-best-effort pods will be remain `pending` if there are not enough resources to satisfy the resoure requests of the job in the cluster.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -231,6 +231,10 @@ func (alloc *allocateAction) Execute(ssn *framework.Session) {
 
 		if ssn.JobReady(job) {
 			stmt.Commit()
+		} else if ssn.JobCandidateReady(job) {
```
k82cn: I think that's not necessary to add a new callback; that should ok to check best-effort pods in `JobReadyFn`.

sivanzcw: If the judgement of the best-effort pod is merged into the `JobReady` fuction, when `JobReady` is satisfied, the dispatch of non-best-effort pod will be started, which will cause the non-best-effort pod of the job to be bound to the node first. But at this time, the gang constraint of the job is not satisfied. Maybe the remaining best-effort pods under the job can not be successfully scheduled during the `backfill` action.

k82cn: IMO, we should dispatch all pods of job in allocate action, including best-effort pod :)

pkg/scheduler/api/job_info.go
```diff
@@ -343,17 +343,19 @@ func (ji *JobInfo) FitError() string {
...
-	occupid := 0
-	for status, tasks := range ji.TaskStatusIndex {
-		if AllocatedStatus(status) ||
-			status == Succeeded {
-			occupid = occupid + len(tasks)
+	var occupied int32
+	for _, task := range ji.Tasks {
+		if task.InitResreq.IsEmpty() {
+			occupied++
+		} else if AllocatedStatus(task.Status) ||
```
k82cn: if so, we're going to go through all pending tasks for this. Use `TaskStatusIndex` to check allocated tasks, and find best-effort pods in pending status.

PR #651, opened 2019-12-29
Reclaim Enhancement: Add Reclaimable parameter for queue

sivanzcw:
* Add parameter for queue to support queue unReclaimable

### DIFF
--- a/pkg/apis/scheduling/types.go
+++ b/pkg/apis/scheduling/types.go
@@ -261,6 +261,8 @@ type QueueSpec struct {
 	Capability v1.ResourceList
 	// State controller the status of queue
 	State QueueState
+	// Reclaimable indicate whether the queue can be reclaimed by other queue
+	Reclaimable *bool
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/apis/scheduling/v1alpha2/types.go
+++ b/pkg/apis/scheduling/v1alpha2/types.go
@@ -262,6 +262,8 @@ type QueueSpec struct {
 	Capability v1.ResourceList `json:"capability,omitempty" protobuf:"bytes,2,opt,name=capability"`
 	// State controller the status of queue
 	State QueueState `json:"state,omitempty" protobuf:"bytes,3,opt,name=state"`
+	// Reclaimable indicate whether the queue can be reclaimed by other queue
+	Reclaimable *bool `json:"reclaimable,omitempty" protobuf:"bytes,4,opt,name=reclaimable"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -142,6 +142,10 @@ func (alloc *reclaimAction) Execute(ssn *framework.Session) {
 				if j, found := ssn.Jobs[task.Job]; !found {
 					continue
 				} else if j.Queue != job.Queue {
+					q := ssn.Queues[j.Queue]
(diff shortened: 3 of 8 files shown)


pkg/apis/scheduling/types.go
```diff
@@ -261,6 +261,8 @@ type QueueSpec struct {
 	Capability v1.ResourceList
 	// State controller the status of queue
 	State QueueState
+	// UnReclaimAble indicate whether the queue can be reclaimed by other queue
+	UnReclaimAble bool
```
k82cn: let's use `Reclaimable` directly :)

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -142,6 +142,10 @@ func (alloc *reclaimAction) Execute(ssn *framework.Session) {
 				if j, found := ssn.Jobs[task.Job]; !found {
 					continue
 				} else if j.Queue != job.Queue {
+					q := ssn.Queues[j.Queue]
+					if q != nil && q.Queue != nil && true == q.Queue.Spec.UnReclaimAble {
```
k82cn: that should be fine to check `Reclaimable` directly.

pkg/apis/scheduling/types.go
hzxuzhonghu: I assume you want to make it compatible with old versions. 
make it a pointer type will be ok

pkg/apis/scheduling/v1alpha2/types.go
```diff
@@ -262,6 +262,8 @@ type QueueSpec struct {
 	Capability v1.ResourceList `json:"capability,omitempty" protobuf:"bytes,2,opt,name=capability"`
 	// State controller the status of queue
 	State QueueState `json:"state,omitempty" protobuf:"bytes,3,opt,name=state"`
+	// ReclaimAble indicate whether the queue can be reclaimed by other queue
+	ReclaimAble *bool `json:"reclaimable,omitempty" protobuf:"bytes,4,opt,name=unreclaimable"`
```
hzxuzhonghu: s/unreclaimable/reclaimable

pkg/apis/scheduling/types.go
```diff
@@ -261,6 +261,8 @@ type QueueSpec struct {
 	Capability v1.ResourceList
 	// State controller the status of queue
 	State QueueState
+	// ReclaimAble indicate whether the queue can be reclaimed by other queue
+	ReclaimAble *bool
```
k82cn: s/ReclaimAble/Reclaimable/g

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -142,6 +142,10 @@ func (alloc *reclaimAction) Execute(ssn *framework.Session) {
 				if j, found := ssn.Jobs[task.Job]; !found {
 					continue
 				} else if j.Queue != job.Queue {
+					q := ssn.Queues[j.Queue]
+					if false == q.ReclaimAble() {
```
k82cn:
```
if !q.Reclaimable() {
    ...
}
```

pkg/scheduler/api/queue_info.go
```diff
@@ -56,3 +56,20 @@ func (q *QueueInfo) Clone() *QueueInfo {
 		Queue:  q.Queue,
 	}
 }
+
+// ReclaimAble return whether queue is reclaimAble
+func (q *QueueInfo) ReclaimAble() bool {
```
k82cn: It's better to set default value in admission controller; so user can also check this value in cli :)

sivanzcw: Since there is no queue admission now, this may require another commit to handle queue validating and mutating.

k82cn: yes, that's ok to open another PR for that :)

PR #653, opened 2019-12-30
Gen install yaml without v1alpha queue and poggroup

hzxuzhonghu:


k82cn:
how are we going to handle backward compatibility?

hzxuzhonghu:
There is no backward compatibility issue. Firstly, this yaml is for developers only; secondly, we do not delete the old crd from k8s cluster.

PR #663, opened 2020-01-08
add defaultQPS and defaultBurst for webhook-manager

yuzhaojing:
Use the same QPS and Burst as the scheduler.


k82cn:
Good catch!

/lgtm
/approve

PR #666, opened 2020-01-14
Add command vjobs, vqueues and unit tests.

jiangkaihua:
Signed-off-by: jiangkaihua <54176503+jiangkaihua@users.noreply.github.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,3 +96,5 @@ command-lines:
 	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vcancel ./cmd/cli/vcancel
 	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vresume ./cmd/cli/vresume
 	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vsuspend ./cmd/cli/vsuspend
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -96,3 +96,5 @@ command-lines:
 	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vcancel ./cmd/cli/vcancel
 	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vresume ./cmd/cli/vresume
 	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vsuspend ./cmd/cli/vsuspend
+	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vjobs ./cmd/cli/vjobs
+	go build -ldflags ${LD_FLAGS} -o=${BIN_DIR}/vqueues ./cmd/cli/vqueues
```
hzxuzhonghu: Why do we need so many cli?

jiangkaihua: They are simliar to Slurm cli like `squeue`, `scancel`, so that Slurm user would get used to our cli more easliy.

hzxuzhonghu: I donot think so. If you make each command a binary, it will definitely be inflated. And it will enforce users to install more tools and remember all.

PR #673, opened 2020-01-20
Remove task's nodeName for statement discard

yodarshafrir1:


yodarshafrir1:
When statement's `pipeline` method is being called, the task's nodeName is set to the value of the node that the task was pipelined to:
https://github.com/volcano-sh/volcano/blob/f17c4831aa9d384819d1d98d8ef62f5c455deb36/pkg/scheduler/framework/statement.go#L128

But when `unpipeline` is being called, the value of the nodeName is never reverted back to nothing, this means that the task will not be pipeline or assigned to any node, but yet the `task.NodeName` will have a value of a node.

PR #677, opened 2020-01-22
Remove unknown annotations

k82cn:


pkg/cli/vjobs/view.go
```diff
@@ -165,9 +165,6 @@ func PrintJobInfo(job *v1alpha1.Job, writer io.Writer) {
 		WriteLine(writer, Level2, "Template:\n")
 		WriteLine(writer, Level2+1, "Metadata:\n")
 		WriteLine(writer, Level2+2, "Annotations:\n")
-		WriteLine(writer, Level2+3, "Cri . Cci . Io / Container - Type:          \t%s\n", job.Spec.Tasks[i].Template.ObjectMeta.Annotations["cri.cci.io/container-type"])
```
k82cn: @jiangkaihua , we need to print all annotation by for loop, similar to plugins.

PR #703, opened 2020-02-17
Dedicated volume implementation

hzxuzhonghu:
Depend on #545

TODO: Add implementation


docs/design/dedicated-volume.md
```diff
@@ -0,0 +1,123 @@
+---
```
k82cn: let's separate design doc into another PR.

hzxuzhonghu: #545

k82cn: ok，why still here？

hzxuzhonghu: I was adding api refer to that. Now, i can remove it.

hzxuzhonghu:
Weird, why it is always cancelled

xiaogaozi:
May I ask, when these PVs will be recycled (when use with PVC)? Maybe should allow user to decide how to recycle them, e.g. delete all PVs when the job is succeeded.

k82cn:
> Maybe should allow user to decide how to recycle them, e.g. delete all PVs when the job is succeeded.

Yes, I think we should add a flag for that :)

k82cn:
@hzxuzhonghu , would you help to resolve the conflict :)

PR #712, opened 2020-02-21
Gang PodGroups Feature Implementation

umialpha:
Pls ref ./docs/degin/gang-podgroup for more information.


docs/design/gang-podgroup.md
```diff
@@ -0,0 +1,121 @@
+# Gang PodGroups
```
k82cn: Suggest to create a separate PR for design doc :)

umialpha: NP, it also conflicts with the newest master. I will rebase first.

Issue #713, opened 2020-02-21
Update version to 0.3 in release-0.3 helm chart


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

In branch `release-0.3`, the version of helm chart is still `latest`; we should update it to `v0.3`.

PR #716, opened 2020-02-22
Update images version of deployment.yaml

k82cn:
xref #713

### DIFF
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -121,7 +121,7 @@ spec:
       
       containers:
         - name: volcano-scheduler
-          image: volcanosh/vc-scheduler:latest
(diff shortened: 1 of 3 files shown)


installer/volcano-development.yaml
```diff
@@ -121,7 +121,7 @@ spec:
       
       containers:
         - name: volcano-scheduler
-          image: volcanosh/vc-scheduler:latest
+          image: volcanosh/vc-scheduler:v0.3
```
hzxuzhonghu: Hmm, i don't think we should do this. As master branch is the developing branch now, and this yaml is for developers, not for production use.

k82cn: we should make sure our release work as expected :)

Issue #700, opened 2020-02-10
Support environments or config for common args in CLI


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature
/area cli
/priority important-soon

**Description**:

Currently, we support several common arguments for command lines, e.g. `--scheduler-name`, `--master`. It's better to have a way to avoid adding those common arguments every-time.

PR #729, opened 2020-03-04
Added default namespace env for other CLIs - vcancel, vjobs, vresume, vsuspend

thandayuthapani:
xref https://github.com/volcano-sh/volcano/issues/700

### DIFF
--- a/pkg/cli/vcancel/cancel.go
+++ b/pkg/cli/vcancel/cancel.go
@@ -18,6 +18,7 @@ package vcancel
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


k82cn:
let's build a util function for that as Namespace is a common parameter for CLIs.

hzxuzhonghu:
What happens when we leave it to empty? From my understanding, it works as `default`. If so, why do we need this?

thandayuthapani:
> let's build a util function for that as Namespace is a common parameter for CLIs.

Flag struct variable and struct are not exposed, so creating a util function to handle all those struct will be a problem.

k82cn:
> Flag struct variable and struct are not exposed, so creating a util function to handle all those struct will be a problem.

Let's add struct such as `JobCommonFlags` :)

PR #732, opened 2020-03-08
Added resource predicate.

k82cn:


k82cn:
We can not do that because of preemption & reclaim :)

/cc @jiangkaihua

Issue #734, opened 2020-03-11
Support preempt for batch job


carmark:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug
/kind feature

**What happened**:

batch job cannot be preempted

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`): any
- Cloud provider or hardware configuration: local system
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

carmark:
any update?

k82cn:
/assign

I'll take some time on this one :)

hzxuzhonghu:
implement #738

PR #738, opened 2020-03-12
support basic batch job preempt

carmark:
fixes: #734

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -247,6 +247,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 			stmt.Commit()
 		} else {
 			stmt.Discard()
+			break
 		}
 
 		// Added Namespace back until no job in Namespace.
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -153,7 +153,7 @@ func (de *defaultEvictor) Evict(p *v1.Pod, reason string) error {
 		return nil
 	}
 	if _, err := de.kubeclient.CoreV1().Pods(p.Namespace).UpdateStatus(pod); err != nil {
-		klog.Errorf("Failed to update pod <%v/%v> status: %v", err)
+		klog.Errorf("Failed to update pod <%v/%v> status: %v", pod.Namespace, pod.Name, err)
 		return err
 	}
 	if err := de.kubeclient.CoreV1().Pods(p.Namespace).Delete(p.Name, &metav1.DeleteOptions{}); err != nil {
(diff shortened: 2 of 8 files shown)


hzxuzhonghu:
@carmark Can you update?

k82cn:
@carmark do you have time to make CI happy?

carmark:
@k82cn @hzxuzhonghu 
Finally, the travis passed.

hzxuzhonghu:
cool

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -247,6 +247,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 			stmt.Commit()
 		} else {
 			stmt.Discard()
+			break
```
hzxuzhonghu, line 250: One question: if break here, then it will not try to schedule other jobs in the same period?

pkg/scheduler/cache/cache.go
```diff
@@ -128,16 +128,39 @@ func (db *defaultBinder) Bind(p *v1.Pod, hostname string) error {
...
+		Status:  v1.ConditionFalse,
+		Reason:  "Evict",
+		Message: evictMsg,
+	}
+	if podutil.UpdatePodCondition(&pod.Status, condition) == false {
+		klog.V(1).Infof("UpdatePodCondition: existed condition, not update")
+		klog.V(1).Infof("%+v", pod.Status.Conditions)
+		return nil
+	}
+	if _, err := de.kubeclient.CoreV1().Pods(p.Namespace).UpdateStatus(pod); err != nil {
```
hzxuzhonghu, line 155: Why do we need to update just before delete?

hzxuzhonghu: To give some clients a hint that why this pod is evicted?

carmark: yeah, the operator may find this changes and create another pod for the job.

pkg/scheduler/actions/allocate/allocate.go
carmark: yes.

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -73,21 +73,17 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
...
 
 		for _, preemptee := range preemptees {
 			job := ssn.Jobs[preemptee.Job]
-			if _, found := jobOccupidMap[job.UID]; !found {
-				jobOccupidMap[job.UID] = job.ReadyTaskNum()
-			}
-			occupid := jobOccupidMap[job.UID]
-			preemptable := job.MinAvailable <= occupid-1 || job.MinAvailable == 1
+
+			preemptable := pJob.Priority > job.Priority
```
hzxuzhonghu, line 81: It seems the preemptFn here now is same as `priority`'s

wpeng102:
@carmark 
It seems this fix can not work with `proportion` + `reclaim`. The `ReclaimableFn` in `gang` is base on job priority. If not set priority for jobs and submit job1 into queue1, job2 into queue2.  Job2 in queue2 can not reclaim any resource from queue1, because of job1 and job2 priority are equal, no victims from Gang plugins.

PR #741, opened 2020-03-22
Dedup buildConfig

hzxuzhonghu:
This is to abstract a common function to build kube rest config


pkg/kube/config.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright 2020 The Vulcan Authors.
```
k82cn: s/Vulcan/Volcano

hzxuzhonghu: Good catch, i replaced all

PR #745, opened 2020-03-24
Use gomod to manage dependencies.

tizhou86:
Use gomod to manage dependencies.


k82cn:
@tizhou86 , should we remove dep config files?

tizhou86:
we can keep it for a while.

PR #747, opened 2020-03-25
Post go module

hzxuzhonghu:
Note: we still keep the vendor directory, but added verify to make it keep up


hzxuzhonghu:
@k82cn It is ready now

PR #756, opened 2020-04-03
[controller] Set `BlockOwnerDeletion` to true when create PodGroup

### LINKED ISSUE #755: Bug Report: When delete pod the related "PodGroup" won't be deleted

/kind bug

**What happened**:

When create `Pod` with Volcano as scheduler, Volcano controller will [create](https://github.com/volcano-sh/volcano/blob/v0.3/pkg/controllers/podgroup/pg_controller.go#L134) a `PodGroup` automatically for this pod. The `OwnerReferences` field of this `PodGroup` will set to original pod. But [`BlockOwnerDeletion`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1?tab=doc#OwnerReference) field in `OwnerReference` not set, so its value will be default value (i.e. false). Reference to [official K8s document](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#controlling-how-the-garbage-collector-deletes-dependents), only object with `BlockOwnerDeletion` set to true will be deleted when in cascading deletion. So when delete original pod the related `PodGroup` won't be deleted.

Delete pod will delete related `PodGroup` also.

Create a `Pod` and set `schedulerName` to `volcano`, then delete the pod see if related `PodGroup` deleted also.

N/A

**Environment**:

- Volcano Version: v0.3
- Kubernetes version (use `kubectl version`): v1.14.6
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

xiaogaozi:
Fix #755

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog"
 
 	"volcano.sh/volcano/pkg/apis/helpers"
@@ -109,14 +110,11 @@ func newPGOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
 		}
 	}
 
-	isController := true
-	blockOwnerDeletion := true
-	return []metav1.OwnerReference{{
-		APIVersion:         v1.SchemeGroupVersion.Version,
-		Kind:               "Pod",
-		Controller:         &isController,
-		Name:               pod.Name,
-		UID:                pod.UID,
-		BlockOwnerDeletion: &blockOwnerDeletion,
-	}}
+	gvk := schema.GroupVersionKind{
+		Group:   v1.SchemeGroupVersion.Group,
+		Version: v1.SchemeGroupVersion.Version,
+		Kind:    "Pod",
+	}
+	ref := metav1.NewControllerRef(pod, gvk)
+	return []metav1.OwnerReference{*ref}
 }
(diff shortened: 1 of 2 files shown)


pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -110,11 +110,13 @@ func newPGOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
...
-		Kind:       "Pod",
-		Controller: &isController,
-		Name:       pod.Name,
-		UID:        pod.UID,
+		APIVersion:         v1.SchemeGroupVersion.Version,
+		Kind:               "Pod",
+		Controller:         &isController,
+		Name:               pod.Name,
+		UID:                pod.UID,
+		BlockOwnerDeletion: &blockOwnerDeletion,
```
hzxuzhonghu, line 120: nit: Instead of constructing this one by one. Can use `metav1.NewControllerRef` function.

hzxuzhonghu:
@xiaogaozi The ut failed

```
=== RUN   TestAddPodGroup
--- FAIL: TestAddPodGroup (0.00s)
    pg_controller_test.go:157: Case AddPodGroup: pod has no ownerReferences or priorityClassName failed, expect &{{PodGroup scheduling.volcano.sh/v1beta1} {podgroup-7a09885b-b753-4924-9fba-77c0836bac20  test    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [{v1 Pod pod1 7a09885b-b753-4924-9fba-77c0836bac20 0xc000046250 <nil>}] []  []} {1   <nil>} { [] 0 0 0}}, got &{{ } {podgroup-7a09885b-b753-4924-9fba-77c0836bac20  test    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [{v1 Pod pod1 7a09885b-b753-4924-9fba-77c0836bac20 0xc0000462b1 0xc0000462b2}] []  []} {1   <nil>} { [] 0 0 0}}
FAIL
```

pkg/controllers/podgroup/pg_controller_handler.go
xiaogaozi: Good suggestion! I've updated the PR.

PR #763, opened 2020-04-07
Update release 0.4 version and sha

hzxuzhonghu:


k82cn:
let's build a script later on that for next release :)

PR #766, opened 2020-04-09
Hope to fix e2e flake

hzxuzhonghu:


test/e2e/util.go
```diff
@@ -226,7 +226,7 @@ func createQueues(cxt *context) {
 }
 
 func deleteQueues(cxt *context) {
-	foreground := metav1.DeletePropagationForeground
+	foreground := metav1.DeletePropagationBackground
```
k82cn, line 229: should we wait for Queue deleted?

hzxuzhonghu:
It is more complicated than i expect

test/e2e/util.go
hzxuzhonghu: Yes, this is not the reason. I found queue not deleted in the previous error log.

And found more error logs showing namespace `test` is being deleted when create it. Not sure why, we have already used foreground delete and also wait for namespace disappear

test/e2e/admission.go
```diff
@@ -33,13 +33,12 @@ var _ = Describe("Job E2E Test: Test Admission service", func() {
 
 	It("Default queue would be added", func() {
 		jobName := "job-default-queue"
-		namespace := "test"
-		context := initTestContext(options{})
-		defer cleanupTestContext(context)
+		ctx := initTestContext(options{})
```
k82cn, line 36: let's keep this var name.

hzxuzhonghu: `context` is a golang standard lib, so let's change this confusing name

hzxuzhonghu:
The current behavior is to run each test in a different namespace.

hzxuzhonghu:
And also i find that the job scheduling case may flakes due to the cluster size.

Issue #778, opened 2020-04-15
mpi job not support hostNetwork: true pattern


tongchao199:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>

/kind bug

**What happened**:
When I ran [mpi-example](https://github.com/volcano-sh/volcano/blob/release-0.4/example/integrations/mpi/mpi-example.yaml), I used the `hostNetwork:true` pattern and found that the task could not parse the mpiworker.host
```shell
root@gpu-ser348-06:/etc/volcano# cat mpiworker.host  && echo
lm-mpi-job-mpiworker-0.lm-mpi-job
lm-mpi-job-mpiworker-1.lm-mpi-job
lm-mpi-job-mpiworker-2.lm-mpi-job
root@gpu-ser348-06:/etc/volcano# ping lm-mpi-job-mpiworker-0.lm-mpi-job
ping: unknown host lm-mpi-job-mpiworker-0.lm-mpi-job
```

**What you expected to happen**:
Use the `hostNetwork:true` pattern could  parse the mpiworker.host

**How to reproduce it (as minimally and precisely as possible)**:
* Deploy volcano
* Modify [mpi-example](https://github.com/volcano-sh/volcano/blob/release-0.4/example/integrations/mpi/mpi-example.yaml) and apply it
```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: lm-mpi-job
spec:
  minAvailable: 3
  schedulerName: volcano
  plugins:
    ssh: []
    svc: []
  tasks:
    - replicas: 1
      name: mpimaster
      policies:
        - event: TaskCompleted
          action: CompleteJob
      template:
        spec:
          hostNetwork: true
          containers:
            - command:
                - /bin/sh
                - -c
                - |
                  MPI_HOST=`cat /etc/volcano/mpiworker.host | tr "\n" ","`;
                  mkdir -p /var/run/sshd; /usr/sbin/sshd;
                  mpiexec --allow-run-as-root --host ${MPI_HOST} -np 2 mpi_hello_world > /home/re;
              image: volcanosh/example-mpi:0.0.1
              name: mpimaster
              ports:
                - containerPort: 22
                  name: mpijob-port
              workingDir: /home
          restartPolicy: OnFailure
    - replicas: 2
      name: mpiworker
      template:
        spec:
          hostNetwork: true
          containers:
            - command:
                - /bin/sh
                - -c
[...]

PR #779, opened 2020-04-18
Set DNSPolicy=ClusterFirstWithHostNet when hostnetwork set

hzxuzhonghu:
https://github.com/volcano-sh/volcano/issues/778

### DIFF
--- a/pkg/controllers/job/job_controller_util.go
+++ b/pkg/controllers/job/job_controller_util.go
@@ -53,15 +53,6 @@ func createJobPod(job *batch.Job, template *v1.PodTemplateSpec, ix int) *v1.Pod 
 		Spec: templateCopy.Spec,
 	}
 
-	// If no scheduler name in Pod, use scheduler name from Job.
-	if len(pod.Spec.SchedulerName) == 0 {
-		pod.Spec.SchedulerName = job.Spec.SchedulerName
-	}
-	// If dns policy not set when hostNetwork=true, set it to `ClusterFirstWithHostNet`
-	if pod.Spec.HostNetwork == true && pod.Spec.DNSPolicy == "" {
-		pod.Spec.DNSPolicy = v1.DNSClusterFirstWithHostNet
-	}
-
 	volumeMap := make(map[string]string)
 	for _, volume := range job.Spec.Volumes {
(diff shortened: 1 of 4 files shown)


pkg/controllers/job/job_controller_util.go
```diff
@@ -57,6 +57,10 @@ func createJobPod(job *batch.Job, template *v1.PodTemplateSpec, ix int) *v1.Pod
 	if len(pod.Spec.SchedulerName) == 0 {
 		pod.Spec.SchedulerName = job.Spec.SchedulerName
 	}
+	// If dns policy not set when hostNetwork=true, set it to `ClusterFirstWithHostNet`
+	if pod.Spec.HostNetwork == true && pod.Spec.DNSPolicy == "" {
```
k82cn, line 61: can we add this into admission controller?

pkg/controllers/job/job_controller_util.go
```diff
@@ -57,6 +57,10 @@ func createJobPod(job *batch.Job, template *v1.PodTemplateSpec, ix int) *v1.Pod
 	if len(pod.Spec.SchedulerName) == 0 {
 		pod.Spec.SchedulerName = job.Spec.SchedulerName
```
k82cn, line 58: ditto, admission controller.

pkg/controllers/job/job_controller_util.go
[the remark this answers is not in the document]
hzxuzhonghu: Yes, i must be drunk

hzxuzhonghu: yes, anyway makes sense to me

pkg/webhooks/admission/jobs/mutate/mutate_job.go
```diff
@@ -33,8 +33,10 @@ import (
 )
 
 const (
-	//DefaultQueue constant stores the name of the queue as "default"
+	// DefaultQueue constant stores the name of the queue as "default"
 	DefaultQueue = "default"
+
+	defaultSchedulerName = "volcano"
```
k82cn, line 39: Let's move this as an argument/parameter, similar to `DefaultQueue`.

PR #782, opened 2020-04-24
Proposal: Volcano job support scale up and down

hzxuzhonghu:
/cc @k82cn @zrss


docs/design/job-scale-up-down.md
```diff
@@ -0,0 +1,63 @@
...
+1. job plugins' callbacks：only the `svc` plugin should update the configmap including the job tasks
+
+2. create pods when scale up
+
+3. delete pods when scale down
+
+However, only when the job is not started, the initialization is run. 
+So we need a way to know whether it is a scale up/down event that triggered this round of sync.
+
+The way I propose is to add a new event `JobUpdatedEvent` to indicate that the job is updated(here only cares about the scale up/down).
```
k82cn, line 45: When Pod was created/deleted, how /when to handle configmap? It's better to highlight the time sequence for the user.

docs/design/job-scale-up-down.md
```diff
@@ -0,0 +1,63 @@
...
+And accordingly add a new action `UpdateJobAction` to run `UpdateJob` function. And the overall workflow is:
+![workflow](images/Job-scale-up-down.PNG)
+
+
+### Admission webhook
+
+Should prevent invalid mutating Job Spec on the fly. In this proposal, we only allow replicas update. Any other spec changes will be prohibited.
```
k82cn, line 52: what's our expected behaviour if minMember & replicas does not match?

docs/design/job-scale-up-down.md
```diff
@@ -0,0 +1,63 @@
+# Volcano Job scale up and down
+
+@hzxuzhonghu; April 24, 2020
+
+## Motivation
+
+Currently, Volcano does not support Job update. It is not allowed to update the `Job.Spec` on the fly.
+However, users like ModelArts want to dynamically adjust Job's replicas according to the cluster idle resources
```
k82cn, line 8: That's not only for ModelArts; AFAIK, several ML framework already support elastic model.

docs/design/job-scale-up-down.md
hzxuzhonghu: I would add a new `OnJobUpdate` method to the plugin

docs/design/job-scale-up-down.md
hzxuzhonghu: invalid update, the api calling will fail

docs/design/job-scale-up-down.md
hzxuzhonghu: Will read about that for more context

docs/design/job-scale-up-down.md
k82cn: > I would add a new `OnJobUpdate` method to the plugin

So?

we need to highlight when create pod, when configmap was updated and which pod will be deleted when scale down.

docs/design/job-scale-up-down.md
k82cn: let's document it, if it's the case

k82cn:
LGTM overall :)

@zrss , please help to confirm whether that's behavour cover your cases :)

docs/design/job-scale-up-down.md
```diff
@@ -0,0 +1,103 @@
...
+The Job's scale up and down correlates to reconciling of the resources the job owns, like PVC/PodGroup/Service/HostFile ConfigMap
+so the procedure is kind of similar to the [Job Initialization](#Job Initialization).
+
+The differences are:
+
+1. job plugins' callbacks：only the `svc` plugin should update the configmap including the job tasks
+
+2. create pods when scale up
+
+3. delete pods when scale down
```
Jeffwan, line 44: Does pod of volcano job has corresponding headless service?

docs/design/job-scale-up-down.md
```diff
@@ -0,0 +1,103 @@
...
+
+However, only when the job is not started, the initialization is run. 
+So we need a way to know whether it is a scale up/down event that triggered this round of sync.
+
+The way I propose is to add a new event `JobUpdatedEvent` to indicate that the job is updated(here only cares about the scale up/down).
+And accordingly add a new action `UpdateJobAction` to run `UpdateJob` function. And the overall workflow is:
+![workflow](images/Job-scale-up-down.PNG)
+
+To scale up/down on the fly, Volcano should be responsible to notify the original pods the current status, including the hosts of all the pods.
```
Jeffwan, line 53: I think we probably need more update categories here.

- pod template change
- job replica change
- job manifest change like annotation

I assume only specific change will trigger `UpdateJobAction`

docs/design/job-scale-up-down.md
```diff
@@ -0,0 +1,103 @@
+# Volcano Job scale up and down
```
Jeffwan, line 1: I assume the scope of this proposal is to make sure controller can response to the job scale up and down. no scale up/down decision need to be made here.

hzxuzhonghu: Yeah, correct.

docs/design/job-scale-up-down.md
hzxuzhonghu: Yes, it is similar to the statefulset pods, we can not delete the headless service when deleting the pods. But we will update the accessible hostfile.

docs/design/job-scale-up-down.md
hzxuzhonghu: These kinds of updates are prohibited.

docs/design/job-scale-up-down.md
Jeffwan: Just curious, Any reason headless service can not be deleted?

hzxuzhonghu: Here we just scale down, there are still some tasks maybe ps exist for tensoflow job. BTW, the headless service is deleted when the job completes or deleted.

zrss:
lgtm, sorry for a long delay ... @hzxuzhonghu @k82cn

PR #784, opened 2020-04-28
Add validate job update

hzxuzhonghu:
This is first step for scale up/down

### DIFF
--- a/pkg/webhooks/admission/jobs/validate/admit_job.go
+++ b/pkg/webhooks/admission/jobs/validate/admit_job.go
@@ -224,6 +224,18 @@ func validateJobUpdate(old, new *v1alpha1.Job) error {
 		new.Spec.Tasks[i].Replicas = old.Spec.Tasks[i].Replicas
 	}
 
+	// jon controller will update the pvc name if not provided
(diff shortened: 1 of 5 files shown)


pkg/webhooks/admission/jobs/validate/admit_job.go
```diff
@@ -194,6 +199,50 @@ func validateJob(job *v1alpha1.Job, reviewResponse *v1beta1.AdmissionResponse) s
...
+		return fmt.Errorf("job updates may not add or remove tasks")
+	}
+	// other fields under spec are not allowed to mutate
+	new.Spec.MinAvailable = old.Spec.MinAvailable
+	for i := range new.Spec.Tasks {
+		new.Spec.Tasks[i].Replicas = old.Spec.Tasks[i].Replicas
+	}
+
+	// jon controller will update the pvc name if not provided
+	for i := range new.Spec.Volumes {
```
k82cn, line 228: that means we did not check volumes update here?

hzxuzhonghu: Nope, we do not check the volume name here, it is auto generated by job controller.

PR #787, opened 2020-04-30
Implement job scale up and down

hzxuzhonghu:
Address #782 

Note: 
1. You are responsible to watch the hostfile changes under `/etc/volcano` if you plan to do scale up/down later.
2. The env `VC_{task name}_HOSTS` and `VC_{task name}_NUM` can not be used when you are using scale up/down.

### DIFF
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -215,7 +215,6 @@ func NewJobController(
 	// Register actions
 	state.SyncJob = cc.syncJob
 	state.KillJob = cc.killJob
-	state.UpdateJob = cc.updateJobFn
 
 	return cc
 }
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -215,6 +215,7 @@ func NewJobController(
 	// Register actions
 	state.SyncJob = cc.syncJob
 	state.KillJob = cc.killJob
+	state.UpdateJob = cc.updateJobFn
 
 	return cc
 }
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -146,6 +146,13 @@ func (cc *Controller) initiateJob(job *batch.Job) (*batch.Job, error) {
 	defer klog.V(3).Infof("Finished Job <%s/%s> initiate", job.Namespace, job.Name)
 
 	klog.Infof("Current Version is: %d of job: %s/%s", job.Status.Version, job.Namespace, job.Name)
+	job, err := cc.initJobStatus(job)
+	if err != nil {
+		cc.recorder.Event(job, v1.EventTypeWarning, string(batch.JobStatusError),
+			fmt.Sprintf("Failed to initialize job status, err: %v", err))
+		return nil, err
+	}
+
 	if err := cc.pluginOnJobAdd(job); err != nil {
 		cc.recorder.Event(job, v1.EventTypeWarning, string(batch.PluginError),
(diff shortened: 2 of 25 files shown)


pkg/controllers/job/job_controller.go
```diff
@@ -215,6 +215,7 @@ func NewJobController(
 	// Register actions
 	state.SyncJob = cc.syncJob
 	state.KillJob = cc.killJob
+	state.UpdateJob = cc.updateJobFn
```
k82cn, line 218: can we handle this in SyncJob?

hzxuzhonghu: Good catch. I have tried and it is tricky. Maybe a separate prin future, there can be lots of flags to control different behavior.

hzxuzhonghu:
Note: there is one leftover issue i can not tackle:
In the svc plugin we do inject envs "VC_%s_HOSTS" and "VC_%s_NUM". They can not be changed without restart pod.  
What's lucky is that: I have not heard any user adopted them

pkg/controllers/job/job_controller.go
k82cn: similar to statefulset controller, in syncJob, it just make sure pod & replicas keep sync.

k82cn:
> Note: there is one leftover issue i can not tackle:
> In the svc plugin we do inject envs "VC_%s_HOSTS" and "VC_%s_NUM". They can not be changed without restart pod.
> What's lucky is that: I have not heard any user adopted them

Update this limitation into design doc :)

hzxuzhonghu:
>Update this limitation into design doc :)

How about delete these two envs? IMO, it is a little hacky to do so and error prone.

k82cn:
> > Update this limitation into design doc :)
> 
> How about delete these two envs? IMO, it is a little hacky to do so and error prone.

That's ok to me; please also highlight in the document :)

pkg/controllers/job/job_controller.go
k82cn: any update?

pkg/controllers/job/job_controller_actions.go
```diff
@@ -300,7 +327,11 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 		return fmt.Errorf("failed to create %d pods of %d", len(creationErrs), len(podToCreate))
 	}
 
-	// TODO: Can hardly imagine when this is necessary.
+	// Delete pods when scale down.
+	// sort pods to be deleted from higher order to lower
+	sort.Slice(podToDelete, func(i, j int) bool {
+		return podToDelete[i].Name > podToDelete[j].Name
```
k82cn, line 333: how about 1 vs 12

hzxuzhonghu:
@k82cn take another round of review

pkg/controllers/job/job_controller_actions.go
```diff
@@ -300,7 +329,15 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 		return fmt.Errorf("failed to create %d pods of %d", len(creationErrs), len(podToCreate))
 	}
 
-	// TODO: Can hardly imagine when this is necessary.
+	// Delete pods when scale down.
+	// sort pods to be deleted from higher order to lower
+	sort.Slice(podToDelete, func(i, j int) bool {
+		iName := strings.Split(podToDelete[i].Name, "-")
```
k82cn, line 335: can we use "volcano.sh/task-index" annotation?

k82cn: as we forbide volcano.sh annotation update in admission controller.

PR #799, opened 2020-05-08
add golangci-lint to project

### LINKED ISSUE #792: Enable more linter checks
We should introduce https://github.com/golangci/golangci-lint to make volcano better.

I donot think i have enough bandwidth, if anyone interested, please feel free to take over.

masihtehrani:
fix #792


k82cn:
That's great to improve golint! Let's get it merged when CI happy with that :)

PR #802, opened 2020-05-11
Upgrade: test volcano on k8s 1.16

hzxuzhonghu:
To keep consistent with our k8s dependency


hzxuzhonghu:
The ci failure is related to in 1.16 k8s removed the extensions/v1beta1 API.  Will check all extensions usage, and migrate to apps/v1.

Issue #791, opened 2020-05-07
Delete an success  pod, will cause volcano recreate an new pod to do repeat task


gaopeiliang:
**What happened**:
when job is running , one pod task successful completed, and delete pod some reason, then another new pod will be create to repeat do same task ....

**What you expected to happen**:
ignore successful completed delete pod event ,, and not create new same pod.... 

**How to reproduce it (as minimally and precisely as possible)**:
1.  running job  , with  short running pod successful completed and one long running pod..
2.  delete short running pod when completed ....

it will created an new short running pod to do same work

**Anything else we need to know?**:

**Environment**:

- Volcano Version:
release-0.4

- Kubernetes version (use `kubectl version`):
1.14
- OS (e.g. from /etc/os-release):
ubuntu:16

- Kernel (e.g. `uname -a`):
Linux 4.9.70-040970-generic 

- Install tools:
- Others:

gaopeiliang:
`
func (cc *Controller) deletePod(obj interface{}) {
        ........

        if err := cc.cache.DeletePod(pod); err != nil {
		klog.Errorf("Failed to delete Pod <%s/%s>: %v in cache",
			pod.Namespace, pod.Name, err)
	}

	key := jobhelpers.GetJobKeyByReq(&req)
	queue := cc.getWorkerQueue(key)
	queue.Add(req)
}
// here  received delete pod event => delete cache  ,,  (but this pod is an successful completed and deleted some reason）

func (cc *Controller) processNextReq(count uint32) bool {
           ........
           cloneJOBINFO
           running => syncJob
           ........
}  

func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateStatusFn) error {

          	for i := 0; i < int(ts.Replicas); i++ {
			podName := fmt.Sprintf(jobhelpers.PodNameFmt, job.Name, name, i)
			if pod, found := pods[podName]; !found {
				newPod := createJobPod(job, tc, i)
                               
                               // here will created new , because it not find in cache , even though it has 
                              //  successful completed ..... 
				if err := cc.pluginOnPodCreate(job, newPod); err != nil {
					return err
				}
				podToCreate = append(podToCreate, newPod)
			} else {
				delete(pods, podName)
				if pod.DeletionTimestamp != nil {
					klog.Infof("Pod <%s/%s> is terminating", pod.Namespace, pod.Name)
					atomic.AddInt32(&terminating, 1)
					continue
				}

[...]

PR #803, opened 2020-05-13
Distinguish different pod-delete scenario

vincent-pli:
Try to address issue #791
It's a draft solution, need further discussion.

In my ENV, seems it could work, but the `pg` status not correct, after delete(after success) the original pod will gone and not recreate but the status of `pg` is:
```
status:
  phase: Running
  running: 2
```
not what expect to:
```
status:
  phase: Running
  running: 2
  success: 1
```

### DIFF
--- a/pkg/controllers/job/job_controller_handler.go
+++ b/pkg/controllers/job/job_controller_handler.go
@@ -332,18 +332,21 @@ func (cc *Controller) deletePod(obj interface{}) {
 		return
 	}
 
-	req := apis.Request{
-		Namespace: pod.Namespace,
-		JobName:   jobName,
-		TaskName:  taskName,
-
-		Event:      bus.PodEvictedEvent,
-		JobVersion: int32(dVersion),
+	if pod.Status.Phase == v1.PodSucceeded {
+		return
 	}
 
 	if err := cc.cache.DeletePod(pod); err != nil {
 		klog.Errorf("Failed to delete Pod <%s/%s>: %v in cache",
 			pod.Namespace, pod.Name, err)
+	}
+
+	req := apis.Request{
+		Namespace:  pod.Namespace,
+		JobName:    jobName,
+		TaskName:   taskName,
+		Event:      bus.PodEvictedEvent,
+		JobVersion: int32(dVersion),
 	}
 
 	key := jobhelpers.GetJobKeyByReq(&req)
--- a/pkg/controllers/cache/cache.go
+++ b/pkg/controllers/cache/cache.go
@@ -59,7 +59,15 @@ func JobKey(job *v1alpha1.Job) string {
 }
 
 func jobTerminated(job *apis.JobInfo) bool {
-	return job.Job == nil && len(job.Pods) == 0
+	for _, pods := range job.Pods {
+		for _, pod := range pods {
+			if pod.Status.Phase != v1.PodSucceeded {
+				return false
+			}
+		}
+	}
+
+	return job.Job == nil
 }
 
 func jobKeyOfPod(pod *v1.Pod) (string, error) {
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -66,9 +66,13 @@ func (cc *Controller) killJob(jobInfo *apis.JobInfo, podRetainPhase state.PhaseM
 			_, retain := podRetainPhase[pod.Status.Phase]
 
 			if !retain {
-				err := cc.deleteJobPod(job.Name, pod)
+				existing, err := cc.deleteJobPod(job.Name, pod)
 				if err == nil {
-					terminating++
+					if existing {
+						terminating++
+					} else {
+						cc.cache.DeletePod(pod)
+					}
 					continue
 				}
 				// record the err, and then collect the pod info like retained pod
@@ -334,7 +338,7 @@ func (cc *Controller) syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateSt
 	for _, pod := range podToDelete {
 		go func(pod *v1.Pod) {
 			defer waitDeletionGroup.Done()
-			err := cc.deleteJobPod(job.Name, pod)
+			_, err := cc.deleteJobPod(job.Name, pod)
 			if err != nil {
 				// Failed to delete Pod, waitCreationGroup a moment and then create it again
 				// This is to ensure all podsMap under the same Job created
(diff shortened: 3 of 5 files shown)


vincent-pli:
@k82cn 
@hzxuzhonghu 
Cannot pass e2e test since can not simulate a `evicted` pod in e2e environment after the modification, any comments?

pkg/controllers/job/job_controller_handler.go
```diff
@@ -324,20 +324,36 @@ func (cc *Controller) deletePod(obj interface{}) {
...
 		klog.Errorf("Failed to delete Pod <%s/%s>: %v in cache",
 			pod.Namespace, pod.Name, err)
 	}
 
+	if pod.Status.Phase == v1.PodSucceeded {
+		return
+	}
+
+	var req apis.Request
+	if pod.Status.Phase == v1.PodFailed && pod.Status.Reason == "Evicted" {
```
hzxuzhonghu, line 337: remove this, we should treat every other phase as Evicted. This is the design intention.

hzxuzhonghu: I mean only ignore `Succeeded` phase.

hzxuzhonghu: And also see https://github.com/volcano-sh/volcano/issues/791#issuecomment-625581151

vincent-pli:
@hzxuzhonghu 
Thanks for the comments.
In my ENV(I replace `volcano-controllers` and `volcano-scheduler` based on the `master`), and use this sample for testing.
```
kind: Job
metadata:
  name: test-job
spec:
  minAvailable: 2
  schedulerName: volcano
  policies:
    - event: PodEvicted
      action: RestartJob
  plugins:
    ssh: []
    env: []
    svc: []
  maxRetry: 0
  queue: default
  tasks:
    - replicas: 2
      name: "default-nginx"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: nginx
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure
    - replicas: 1
      name: "sleep"
      template:
        metadata:
          name: sleep
        spec:
          containers:
            - image: docker
              imagePullPolicy: IfNotPresent
              name: sleep
              args: ["sleep", "60"]
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure
```
 Based on my testing, with this pr, when delete the pod when complete, the `PodEvictedEvent` will not emit, so the `syncJob` action will not trigger, and when delete the pod `running` the job will restart, the result is expected.

hzxuzhonghu:
`syncJob` can be triggered by many events, not only pod events. keep in mind it is to reconcile the job.

vincent-pli:
@hzxuzhonghu Thanks for your comments.

Agree with you `syncJob` actually is a `reconcile` of `Job`, based on this I think it's hard to handle this scenario here. since `syncJob` has no idea about a `pod` which both delete from cluster and cache, unless extends the `job.status` like this:
```
succeeded: 
      test-job-default-nginx-0
      test-job-default-nginx-1
running:
      test-job-default-nginx-2
```

So I think could we keep the `success` and `deleted` pod in cache, not clean it. 
Please check the modification, expect to your comments, thanks

hzxuzhonghu:
@vincent-pli Basically looks good, but I need to further look in case I miss something.

hzxuzhonghu:
Apply the below job

```
kind: Job
metadata:
  name: test-job
spec:
  minAvailable: 3
  schedulerName: volcano
  policies:
    - event: PodEvicted
      action: RestartJob
  plugins:
    ssh: []
    env: []
    svc: []
  maxRetry: 5
  queue: default
  tasks:
    - replicas: 2
      name: "default-nginx"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: nginx
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure
    - replicas: 1
      name: "success"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: nginx
              imagePullPolicy: IfNotPresent
              command: ["/bin/sh", "-c", "sleep 10"]
              name: nginx
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure

```
And then do as below:

1. wait until the job is running and the second task succeeded.
2. delete the succeeded pod, it works as intended: no new pod created
3. delete one of the pod in running phase, now it behaves unexpected, the job will be in restarting phase always.

hzxuzhonghu:
@vincent-pli FYI https://github.com/hzxuzhonghu/volcano/commits/distinguish-evicted-and-normal-pod-delete

vincent-pli:
@hzxuzhonghu 
Thanks for comments,
I tried your case, I found the job stick in `restarting` phase because it can not meet the `minAvailable`, since the `task`: "success" is complete and will not start again. Change `minAvailable` to 2 will resolve it.

For the https://github.com/hzxuzhonghu/volcano/commits/distinguish-evicted-and-normal-pod-delete, I think it's necessary fix, could you show me some case which will broken without the fix (the number of terminating?), thanks.

hzxuzhonghu:
>could you show me some case which will broken without the fix (the number of terminating?), thanks.

What do you mean by `the number of terminating`?

vincent-pli:
@hzxuzhonghu 
Thanks for your comments,
I update the code, just one thing:
we should not delete from cache here,  it will cause the complete pod create again after Restart.
image

hzxuzhonghu:
No, this is called in `killJob`, which is to kill pods that need to be killed. And if it is restarting, then all new pods will be created

vincent-pli:
@hzxuzhonghu 
I think we do not need run `cache.DeletePod(pod)` here, since in `KillJob` it delete `pod` from cluster,  and the `podInformer` we listen to will trigger `cache.DeletePod(pod)`(will not remove `success` pod: enhanced in this pr).

Delete here will cause duplicate and break logic in `Eventhandler` for `podInformer`.

hzxuzhonghu:
I will tell you why we need it.

You can verify by creating a job, and wait one task succeeded, and delete it. It will not be deleted from the cache.

And then you delete a running pod to simulate the eviction, without this the result is unexpected.

I am concerned with this, and I think this is not a good solution. Maybe we should come up with a grace way to deal with it.

vincent-pli:
Yeah, the `RestartJob` should restart all the `tasks`, regardless the original status is `success` or not, it make sense, thanks.

I agree with you the solution is not graceful, but seems could satisfy us at least now.
anyway, I will correct code, thanks for comments, it's helpful for me go though the code.

hzxuzhonghu:
Please fix the ut

k82cn:
@vincent-pli , do you have time to continue this PR ?

vincent-pli:
@k82cn 
Sorry for that, I struggled with product stuff recently, maybe I can continue later this week, I remember i can not make the e2e test success, anyway, I will take a try later this week, any comments?

PR #808, opened 2020-05-14
Fix crd validation

hzxuzhonghu:


hzxuzhonghu:
This is caused by manually editing.  The better way is to make use of existing tools to generate CRD automatically. I will file an issue.

Issue #822, opened 2020-05-20
Job validatation does not accept fractional GPU


william-wang:
**What happened**:
Volano job validation does not accept a fractional GPU and report error.

**What you expected to happen**:
Job webhook validation should accept fractional GPU for sharing scenario.

**How to reproduce it (as minimally and precisely as possible)**:
$ cat test-gpu-job-01.yaml 
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: test-gpu-job-01
spec:
  minAvailable: 1
  schedulerName: volcano
  plugins:
    env: []
    ssh: []
    svc: []
  maxRetry: 5
  tasks:
    - replicas: 1
      name: "default-nginx"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: 100.79.1.215:20202/paas_cce_y00416048/nginx:latest
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
                limits:
                  nvidia.com/gpu: 0.4
          imagePullSecrets:
          - name: default-secret
          restartPolicy: OnFailure
$ kubectl apply -f test-gpu-job-01.yaml 
Error from server: error when creating "test-gpu-job-01.yaml": admission webhook "validatejob.volcano.sh" denied the request: spec.task[0].template.spec.containers[0].resources.limits[nvidia.com/gpu]: Invalid value: resource.Quantity{i:resource.int64Amount{value:5, scale:-1}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"", Format:"DecimalSI"}: must be an integer.

**Environment**:
- Volcano Version: 0.4
- Kubernetes version (use `kubectl version`): 1.13
- Cloud provider or hardware configuration: Huawei Cloud
- OS (e.g. from /etc/os-release): Linux amd64 and ARM

hzxuzhonghu:
This is fixed

PR #823, opened 2020-05-20
fix the job validation to allow using fractional GPU

william-wang:
Address issue #822

### DIFF
--- a/pkg/webhooks/admission/jobs/validate/admit_job.go
+++ b/pkg/webhooks/admission/jobs/validate/admit_job.go
@@ -24,6 +24,7 @@ import (
 	whv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	resource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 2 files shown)


pkg/webhooks/admission/jobs/validate/admit_job.go
```diff
@@ -257,6 +258,12 @@ func validateTaskTemplate(task v1alpha1.TaskSpec, job *v1alpha1.Job, index int)
 		if container.SecurityContext != nil && container.SecurityContext.Privileged != nil {
 			coreTemplateSpec.Spec.Containers[i].SecurityContext.Privileged = nil
 		}
+		if _, found := container.Resources.Limits["nvidia.com/gpu"]; found {
```
hzxuzhonghu, line 261: please add a comment, assume this is for gpu share

william-wang: the comment is added.

k82cn: is there any const for "nvidia.com/gpu"? if not, let's add a new one.

k82cn:
> please add a ut

that's ok to have a seperate PR for ut :)

pkg/webhooks/admission/jobs/validate/admit_job.go
hzxuzhonghu: not any const for it now. 

What about other GPU types, like ascend, AMD?

william-wang: Add the ascend-310. Others are not supportted so far.

k82cn:
please also commit to master for next release :)

PR #825, opened 2020-05-21
Lint: fix deadcode issue

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -69,12 +69,11 @@ issues:
       linters:
         - gomnd
         - dupl
-    - path: pkg/apis/scheduling/v1beta1/conversion.go
+    - path: v1beta1
       linters:
         - golint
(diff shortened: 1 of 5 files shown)


pkg/client/clientset/versioned/fake/register.go
```diff
@@ -31,7 +31,6 @@ import (
 
 var scheme = runtime.NewScheme()
 var codecs = serializer.NewCodecFactory(scheme)
-var parameterCodec = runtime.NewParameterCodec(scheme)
```
hzxuzhonghu, line 34: This is a generated file, should be excluded.

Same applies to ` pkg/client`

daixiang0: Done.

PR #835, opened 2020-05-26
Lint: fix stylecheck issue

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>


daixiang0:
If so, we can disable this linter.

PR #843, opened 2020-05-31
Skipped task if no candidate nodes.

### LINKED ISSUE #838: Observed a panic: invalid argument to Intn（when batchNodeOrder function return err)
> /kind bug
**What happened**:
I added a schedule plugin ，and used a function named batchNodeOrder. This funciton return a error when nodes unmeet required. But, the volcano-schedule crash.

Add the job backup to queue, and run ok

```
			node := ssn.BestNodeFn(task, nodeScores)
			if node == nil {
				node = util.SelectBestNode(nodeScores)
			}
```
After these, add the node judge.  

**Environment**:
- Volcano Version:  0.4
- Kubernetes version (use `kubectl version`): 1.17.3
- Cloud provider or hardware configuration: arm64
- OS (e.g. from /etc/os-release): unbuntu 13.2
- Kernel (e.g. `uname -a`):
- Install tools:
volcano-shedule-crash

k82cn:
fix #838 

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -199,6 +199,11 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 				}
 			}
 
+			// If not candidate nodes for this task, skip it.
+			if len(candidateNodes) == 0 {
+				continue
+			}
+
 			nodeScores := util.PrioritizeNodes(task, candidateNodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
 
 			node := ssn.BestNodeFn(task, nodeScores)


hzxuzhonghu:
The real cause is from `node = util.SelectBestNode(nodeScores)`, where nodeScores is empty.

k82cn:
> The real cause is from `node = util.SelectBestNode(nodeScores)`, where nodeScores is empty.

The nodeScores is built from candidateNodes.

hzxuzhonghu:
@zishen added a customized plugin, which functions on `PrioritizeNodes`

k82cn:
> @zishen added a customized plugin, which functions on `PrioritizeNodes`

xref https://github.com/volcano-sh/volcano/issues/838#issuecomment-636323529

hzxuzhonghu:
Please note:  It is possible when `candidateNodes` is not empty but `nodeScores` is.

k82cn:
> Please note: It is possible when `candidateNodes` is not empty but `nodeScores` is.

when? `PrioritizeNodes` should only sort `candidateNodes`.

hzxuzhonghu:
@zishen Please help explain your use case. I remember that batchFn fails, and returns

```
	batchNodeScore, err := batchFn(task, nodes)
	if err != nil {
		klog.Errorf("Error in Calculating batch Priority for the node, err %v", err)
		return nodeScores
	}

```

PR #848, opened 2020-06-04
add job e2e

Thor-wl:

### DIFF
--- a/test/e2e/admission.go
+++ b/test/e2e/admission.go
@@ -221,7 +221,7 @@ var _ = ginkgo.Describe("Job E2E Test: Test Admission service", func() {
 		gomega.Expect(err.Error()).Should(gomega.ContainSubstring(`the podgroup phase is Pending`))
 	})
 
-	ginkgo.It("Valid volcano job", func() {
+	ginkgo.It("Job mutate check", func() {
 		ctx := initTestContext(options{})
 		defer cleanupTestContext(ctx)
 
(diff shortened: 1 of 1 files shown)


test/e2e/admission.go
```diff
@@ -220,4 +220,1103 @@ var _ = ginkgo.Describe("Job E2E Test: Test Admission service", func() {
 		_, err = ctx.kubeclient.CoreV1().Pods(ctx.namespace).Create(pod)
 		gomega.Expect(err.Error()).Should(gomega.ContainSubstring(`the podgroup phase is Pending`))
 	})
+
+	ginkgo.It("Job mutate check", func() {
+		ctx := initTestContext(options{})
+		defer cleanupTestContext(ctx)
+
+		var job v1alpha1.Job
+		jsonData := []byte(`{
```
hzxuzhonghu, line 229: Why not define Job directly here?

test/e2e/admission.go
```diff
@@ -220,4 +220,1103 @@ var _ = ginkgo.Describe("Job E2E Test: Test Admission service", func() {
...
+			 "ttlSecondsAfterFinished": -1
+		 }
+	 }`)
+		err := json.Unmarshal(jsonData, &job)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		_, err = ctx.vcclient.BatchV1alpha1().Jobs(ctx.namespace).Create(&job)
+		gomega.Expect(err).To(gomega.HaveOccurred())
+	})
+
+	ginkgo.It("job validate check: illegal minAvailable when create", func() {
```
hzxuzhonghu, line 542: `minAvailable`？

test/e2e/admission.go
hzxuzhonghu: Applies to all cases

Thor-wl: black box test use json style input may be better?

test/e2e/admission.go
Thor-wl: It is strange that no such bug in my local workspace, and neither is encode mistake. I've modified the comments

hzxuzhonghu:
/lgtm

Need to migrate some cases to ut to reduce the time running.

PR #852, opened 2020-06-08
Add gpu sharing in predicate logic.

tizhou86:
Add gpu sharing in predicate logic, will add gpu sharing in allocate logic in next pr.


pkg/scheduler/api/device_info.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (di *DeviceInfo) GetUsedGPUMemory() uint {
+	res := uint(0)
+	for _, pod := range di.PodMap {
+		if pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed {
+			continue
+		} else {
+			if len(pod.ObjectMeta.Annotations) > 0 {
+				mem, found := pod.ObjectMeta.Annotations["volcano.sh/pod-gpu-memory"]
```
hzxuzhonghu, line 54: declare a const instead

pkg/scheduler/api/node_info.go
```diff
@@ -314,3 +346,60 @@ func (ni *NodeInfo) Pods() (pods []*v1.Pod) {
...
 	return
 }
+
+func (ni *NodeInfo) CheckPredicatePodOnGPUNode(pod *v1.Pod) bool {
+	res := false
+	memReq := uint(0)
+
+	remainMems := ni.GetDevicesRemainGPUMemory()
+	if len(pod.ObjectMeta.Annotations) > 0 {
+		mem, found := pod.ObjectMeta.Annotations["volcano.sh/pod-gpu-memory"]
```
hzxuzhonghu, line 356: Why not use resource request?

pkg/scheduler/api/node_info.go
```diff
@@ -314,3 +346,60 @@ func (ni *NodeInfo) Pods() (pods []*v1.Pod) {
...
+			m, _ := strconv.Atoi(mem)
+			memReq = uint(m)
+		}
+	}
+
+	if len(remainMems) > 0 {
+		for devID := 0; devID < len(ni.Devices); devID++ {
+			availableGPU, ok := remainMems[devID]
+			if ok {
+				if availableGPU >= memReq {
```
hzxuzhonghu, line 367: I understand the devices here is split by gpu core from `SetNodeGPUInfo` above.
And we can only allocate one device for a pod. My question is why can't we allocate more than one device?

pkg/scheduler/api/node_info.go
```diff
@@ -154,6 +168,24 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
...
 }
 
+func (ni *NodeInfo) SetNodeGPUInfo(node *v1.Node) {
+
+	core, ok := node.Status.Capacity["volcano.sh/node-gpu-core"]
+	if ok {
+		ni.GPUTotalCore = int(core.Value())
+	}
+
+	mem, ok := node.Status.Capacity["volcano.sh/node-gpu-memory"]
```
hzxuzhonghu, line 178: Let's declare with a const, same applies to "volcano.sh/node-gpu-core"

pkg/scheduler/api/node_info.go
```diff
@@ -46,8 +46,7 @@ type NodeInfo struct {
 
 	Tasks map[TaskID]*TaskInfo
 
-	// Used to store custom information
-	Others map[string]interface{}
+	GPUDevices map[int]*GPUDevice
```
k82cn, line 49: please do not remove `Others`, some projects are using it.

hzxuzhonghu: reverted

pkg/scheduler/api/node_info.go
```diff
@@ -154,9 +145,35 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
...
+	res, ok := node.Status.Capacity[VolcanoGPUNumber]
+	if !ok {
+		return
+	}
+	gpuNumber := res.Value()
+	if gpuNumber == 0 {
+		klog.Warningf("invalid %s=%s", VolcanoGPUNumber, res.String())
+	}
+
+	memoryPerCard := uint(totalMemory / gpuNumber)
```
carmark, line 167: gpuNumber may be `0` ?

hzxuzhonghu: Correct, need to fix

hzxuzhonghu: done

PR #855, opened 2020-06-10
Bump to k8s 1.18.3

hzxuzhonghu:


pkg/scheduler/framework/framework.go
```diff
@@ -139,7 +138,7 @@ func (f *framework) VolumeBinder() scheduling.SchedulerVolumeBinder {
 }
 
 func NewFrameworkHandle(pods []*v1.Pod, nodes []*v1.Node) v1alpha1.FrameworkHandle {
-	snapshot := k8scache.NewSnapshot(pods, nodes)
+	snapshot := cache.NewSnapshot(pods, nodes)
```
k82cn, line 142: let's open an issue in upstream to make cache public.

hzxuzhonghu: https://github.com/kubernetes/kubernetes/issues/91983

hzxuzhonghu:
@k82cn please take a look at the scheduling part, especially the nodeorders.go and predicate.go

pkg/scheduler/framework/framework.go
k82cn: ok, let's move this PR forward; I don-t think we will have a conclusion in a shot time :(

k82cn: ok, let's move this PR forward; I don-t think we will have a conclusion in a shot time :(

k82cn:
> @k82cn please take a look at the scheduling part, especially the nodeorders.go and predicate.go

How can I find them in 1.4k files and 15 commits? please separate this PR to some meaningful commits, e.g. scheduler, controller, vendor, test ...

PR #862, opened 2020-06-12
Refactor statement

hzxuzhonghu:
Signed-off-by: xuzhonghu <xuzhonghu@huawei.com>

### DIFF
--- a/pkg/scheduler/framework/statement.go
+++ b/pkg/scheduler/framework/statement.go
@@ -25,22 +25,41 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 )
 
+// Operation type
+type Operation int8
+
+const (
+	// Evict op
+	Evict = iota
+	// Pipeline op
+	Pipeline
+	// Allocate op
+	Allocate
+)
+
+type operation struct {
+	name   Operation
+	task   *api.TaskInfo
+	reason string
+}
+
 // Statement structure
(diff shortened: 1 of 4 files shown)


pkg/scheduler/framework/statement.go
```diff
@@ -254,8 +271,9 @@ func (s *Statement) Allocate(task *api.TaskInfo, hostname string) error {
 	// Update status in session
 	klog.V(3).Info("Allocating operations ...")
 	s.operations = append(s.operations, operation{
-		name: "allocate",
-		args: []interface{}{task, hostname},
+		name:   Allocate,
+		task:   task,
+		reason: hostname,
```
k82cn, line 276: allcate reason is hostname?

pkg/scheduler/framework/statement.go
```diff
@@ -25,22 +25,37 @@ import (
...
+const (
+	Evict = iota
+	Pipeline
+	Allocate
+)
+
+type operation struct {
+	name   Operation
+	task   *api.TaskInfo
+	reason string
```
k82cn, line 39: let's use `args []string` to make it more general :)

pkg/scheduler/framework/statement.go
hzxuzhonghu: Removed. Actually only for evict it is used, for other ops useless.

PR #884, opened 2020-06-27
fix default Qeue e and webhook startup sequence

QizhouYang:
fix default Qeue e and webhook startup sequence
Step·：
change pkg/apis/scheduling/v1beta1/types.go, add ”state“ field

Step2：
change pkg/scheduler/cache/cache.go


QizhouYang:
Signed-off-by: QizhouYang  13523582458@163.com

PR #885, opened 2020-06-27
fix default Qeue e and webhook startup sequence

QizhouYang:
Signed-off-by: QizhouYang <13523582458@163.com>

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -267,12 +267,14 @@ func newSchedulerCache(config *rest.Config, schedulerName string, defaultQueue s
 	}
 
 	// create default queue
+	reclaimable := true
 	defaultQue := vcv1beta1.Queue{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: defaultQueue,
 		},
 		Spec: vcv1beta1.QueueSpec{
-			Weight: 1,
+			Reclaimable: &reclaimable,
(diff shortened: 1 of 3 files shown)


pkg/apis/scheduling/v1beta1/types.go
```diff
@@ -261,6 +261,7 @@ type QueueStatus struct {
 type QueueSpec struct {
 	Weight     int32           `json:"weight,omitempty" protobuf:"bytes,1,opt,name=weight"`
 	Capability v1.ResourceList `json:"capability,omitempty" protobuf:"bytes,2,opt,name=capability"`
+	State QueueState `json:"state,omitempty" protobuf:"bytes,1,opt,name=state"`
```
k82cn, line 264: That's required to your environment? We already moved this to `QueueStatus` :)

QizhouYang: cache.go file
```
Spec: vcv1beta1.QueueSpec{
			Reclaimable: &reclaimable,
			State:       "Open",
			Weight:      1,
		},
```
declare default Queue！

QizhouYang: If this field `QueueStatus` is not needed，Delete the default value in webhook？

QizhouYang: If this field `State` is not needed，Delete the default value in webhook？

PR #893, opened 2020-07-01
0701 e2e：add preemption e2e case

Thor-wl:
add a case about invalid queue of incoming preemptor job


Thor-wl:
In essence，this case is same with case "job validate check: invalid queue when create " in admission e2e case. Close pr

PR #900, opened 2020-07-02
add some allocate cases

alcorf-mizar:
Signed-off-by: alcorf-mizar <yihang.von@gmail.com>

### DIFF
--- a/test/e2e/allocate.go
+++ b/test/e2e/allocate.go
@@ -1,0 +1,288 @@
+package e2e
+
+import (
+	"context"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	schedulingv1beta1 "volcano.sh/volcano/pkg/apis/scheduling/v1beta1"
+)
+
+var _ = Describe("Job E2E Test", func() {
+	It("schedule job when resource is enough", func() {
+		ctx := initTestContext(options{})
+		defer cleanupTestContext(ctx)
+
+		req := corev1.ResourceList{
+			corev1.ResourceCPU: *resource.NewQuantity(2, resource.DecimalSI),
+			corev1.ResourceMemory: *resource.NewQuantity(2, resource.BinarySI),
+		}
+
+		available, err := setPlaceHolderForSchedulerTesting(ctx, req, 2)
+		Expect(err).NotTo(HaveOccurred())
+
+		if !available {
+			Skip("no enough resources in cluster")
+		}
+
+		slot := oneCPU
+		job := &jobSpec{
+			tasks: []taskSpec{
+				{
+					img: defaultNginxImage,
+					req: slot,
+					min: 1,
+					rep: 1,
+				},
+			},
+		}
+
+		job.name = "low"
+		lowReqJob := createJob(ctx, job)
+		err = waitJobReady(ctx, lowReqJob)
+		Expect(err).NotTo(HaveOccurred())
+
+		job.name = "high"
+		job.tasks[0].req = twoCPU
+		highReqJob := createJob(ctx, job)
+		err = waitJobReady(ctx, highReqJob)
+		Expect(err).NotTo(HaveOccurred())
+
+		err = compareNodename(ctx, lowReqJob, highReqJob)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("schedule job when resource is NOT enough", func() {
+		ctx := initTestContext(options{})
+		defer cleanupTestContext(ctx)
+
+		req := corev1.ResourceList{
+			corev1.ResourceCPU: *resource.NewQuantity(2, resource.DecimalSI),
+			corev1.ResourceMemory: *resource.NewQuantity(2, resource.BinarySI),
+		}
+
+		available, err := setPlaceHolderForSchedulerTesting(ctx, req, 2)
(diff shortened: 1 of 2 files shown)


test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
+package e2e
```
hzxuzhonghu, line 1: Add copy right

test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
...
+
+import (
+	"context"
+	. "github.com/onsi/ginkgo"
```
hzxuzhonghu, line 5: nit: add a blank line before

test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
...
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/resource"
```
hzxuzhonghu, line 7: move down and well group all `k8s.io` prefix lib

test/e2e/util.go
```diff
@@ -1379,3 +1468,39 @@ func isPodScheduled(pod *v1.Pod) bool {
...
+	})
+
+	if len(pods.Items) == 1 {
+		return pods.Items[0].Spec.NodeName, nil
+	}
+
+	return "", fmt.Errorf("get pods num not equal 1")
+}
+
+func compareNodename(ctx *testContext, jobA, jobB *batchv1alpha1.Job) (err error) {
```
hzxuzhonghu, line 1497: We can not guarantee the given job is of one replica.  May cause misuse by other developers

test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
...
+					req: slot1,
+					min: 1,
+					rep: 1,
+				},
+			},
+		}
+
+		job.name = "j1-n3"
+		node3Job := createJobWithCondition(ctx, job, "", map[string]string{
+			corev1.LabelHostname: "node3",
```
hzxuzhonghu, line 218: how do you get "node3"

test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		err = compareNodename(ctx, lowReqJob, highReqJob)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("schedule job when resource is NOT enough", func() {
+		ctx := initTestContext(options{
+			nodesNumLimit: 2,
+			nodesResourceLimit: corev1.ResourceList{
```
hzxuzhonghu, line 60: does this mean the resource specified is idle?

test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
...
+		err = waitJobReady(ctx, highReqJob)
+		Expect(err).NotTo(HaveOccurred())
+
+		err = compareNodename(ctx, lowReqJob, highReqJob)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("schedule job when resource is NOT enough", func() {
+		ctx := initTestContext(options{
+			nodesNumLimit: 2,
```
hzxuzhonghu, line 59: what does this mean?

test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
...
+		pg := &schedulingv1beta1.PodGroup{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: ctx.namespace,
+				Name:      pgName,
+			},
+			Spec: schedulingv1beta1.PodGroupSpec{
+				MinMember:    1,
+				MinResources: &thirtyCPU,
+			},
+			Status: schedulingv1beta1.PodGroupStatus{
```
hzxuzhonghu, line 127: remove, it is common sense to create without status specified.

test/e2e/allocate.go
```diff
@@ -0,0 +1,289 @@
...
+		err = waitJobReady(ctx, existJob)
+		Expect(err).NotTo(HaveOccurred())
+
+		job.name = "allocate"
+		allocateJob := createJobWithCondition(ctx, job, pgName, nil)
+		err = waitJobPending(ctx, allocateJob)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("allocate don't work when queue NOT exist", func() {
```
hzxuzhonghu, line 161: I donot think this case will test anything on scheduling, the job creation is denied by admission.
So i would suggest removing it

test/e2e/allocate.go
hzxuzhonghu: Do you mean a non-exist node?

test/e2e/util.go
```diff
@@ -333,6 +335,93 @@ func createJob(context *testContext, jobSpec *jobSpec) *batchv1alpha1.Job {
 	return job
 }
 
+func createJobWithCondition(ctx *testContext, jobSpec *jobSpec, pgName string, nodeSelector map[string]string) *batchv1alpha1.Job {
```
hzxuzhonghu, line 338: It is not clear what condition is from the function name

test/e2e/allocate.go
alcorf-mizar: I agree with you. I'll remove it.

test/e2e/allocate.go
alcorf-mizar: Ok

test/e2e/allocate.go
alcorf-mizar: I will use other conditions to test allocate when job not satisify predicates.

test/e2e/allocate.go
alcorf-mizar: This means the resources specified is required(Prepare form cluster nodes).

PR #901, opened 2020-07-03
fix issue 890

Thor-wl:


k82cn:
Is this fix inclued in mainline?

Thor-wl:
> Is this fix inclued in mainline?

I will push another pr to master branch

PR #905, opened 2020-07-04
add drf e2e cases

Thor-wl:

### DIFF
--- a/test/e2e/drf.go
+++ b/test/e2e/drf.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Volcano Authors.
+Copyright 2020 The Volcano Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


test/e2e/drf.go
```diff
@@ -0,0 +1,37 @@
+package e2e
```
hzxuzhonghu, line 1: copyright

test/e2e/drf.go
```diff
@@ -0,0 +1,37 @@
...
+
+		job.name = "j1-reference"
+		referenceJob := createJob(ctx, job)
+		err := waitTasksReady(ctx, referenceJob, int(rep))
+		Expect(err).NotTo(HaveOccurred())
+
+		job.name = "j2-drf"
+		job.tasks[0].req = halfCPU
+		backfillJob := createJob(ctx, job)
+		err = waitTasksReady(ctx, backfillJob, int(rep))
```
hzxuzhonghu, line 34: Can you elaborate how drf work in this case?

Thor-wl: drf works when a task who requires less resource and to be schedulered will preempt resource of another task who require more resource and is running

test/e2e/drf.go
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright 2019 The Volcano Authors.
```
hzxuzhonghu, line 2: s/2019/2020

test/e2e/drf.go
hzxuzhonghu: add comments

PR #917, opened 2020-07-08
e2e: add case 2 to reclaim

alcorj-mizar:
add case 2 to reclaim


hzxuzhonghu:
@alcorj-mizar please rebase

PR #924, opened 2020-07-13
update kind to see if it speed a little bit up

hzxuzhonghu:


hzxuzhonghu:
close, we have split e2e

PR #928, opened 2020-07-14
Support hierarchical dominant resource fairness

ggaaooppeenngg:
Hierarchical dominant resource fairness is configured with
a weighted tree, such that each node in the tree has a positive weight value.
The weights denote relative importance and do not have to sum to 1. The leaves
in the tree denote the jobs that are ultimately to be allocated resources,
whereas the internal nodes represent organizational or hierarchical groups.

In the case that an internal node has a child with a higher dominant share
while other children with another type of resource to be allocated will starve,
HDRF scales the child to the minimum dominant share and sums them up as the share of itself.

The hierarchy and weights of interest are annotated in the PodGroup
or Job annotations.

Signed-off-by: Peng Gao <peng.gao.dut@gmail.com>

### DIFF
--- a/example/hierarchical-jobs/jobs.yaml
+++ b/example/hierarchical-jobs/jobs.yaml
@@ -1,10 +1,9 @@
+
+---
 apiVersion: batch.volcano.sh/v1alpha1
 kind: Job
 metadata:
   name: eng-prod-job
-  annotations:
-    "scheduling.k8s.io/group-hierarchy": "root/eng/prod"
-    "scheduling.k8s.io/group-hierarchy-weight": "1/2/8"
 spec:
   minAvailable: 1
   schedulerName: volcano
@@ -16,7 +15,7 @@ spec:
     env: []
     svc: []
   maxRetry: 5
-  queue: default
+  queue: root-eng-prod
   tasks:
     - replicas: 6
       name: "default-nginx"
@@ -38,9 +37,6 @@ apiVersion: batch.volcano.sh/v1alpha1
 kind: Job
 metadata:
   name: eng-dev-job
-  annotations:
-    "scheduling.k8s.io/group-hierarchy": "root/eng/dev"
-    "scheduling.k8s.io/group-hierarchy-weight": "1/2/2"
 spec:
   minAvailable: 1
   schedulerName: volcano
@@ -52,7 +48,7 @@ spec:
     env: []
     svc: []
   maxRetry: 5
-  queue: default
+  queue: root-eng-dev
   tasks:
     - replicas: 6
       name: "default-nginx"
@@ -76,9 +72,6 @@ apiVersion: batch.volcano.sh/v1alpha1
 kind: Job
 metadata:
   name: sci-prod-job
-  annotations:
-    "scheduling.k8s.io/group-hierarchy": "root/sci/prod"
-    "scheduling.k8s.io/group-hierarchy-weight": "1/8/8"
 spec:
   minAvailable: 1
   schedulerName: volcano
@@ -90,7 +83,7 @@ spec:
     env: []
     svc: []
   maxRetry: 5
-  queue: default
+  queue: root-sci-prod
   tasks:
     - replicas: 6
       name: "default-nginx"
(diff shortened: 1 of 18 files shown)


hzxuzhonghu:
@ggaaooppeenngg Could you add a proposal first? So we can evaluate easily.

example/hierarchical-jobs/jobs.yaml
```diff
@@ -0,0 +1,145 @@
+apiVersion: batch.volcano.sh/v1alpha1
+kind: Job
+metadata:
+  name: eng-prod-job
+  annotations:
+    "scheduling.k8s.io/group-hierarchy": "root/eng/prod"
```
k82cn, line 6: it's better to update `JobSpec` directly for this feature.

and let's use `volcano.sh` instead of `k8s.io` :)

k82cn:
That's great feature ! Prefer to have a design doc to clarify the case and the overall design :)

ggaaooppeenngg:
I made an issue #933 about the related proposal.

k82cn:
hi team, what's our next step of this feature?

stpabhi:
I'm interested in this feature. can this be merged or still wip?

ggaaooppeenngg:
I made some updates on the proposal #933, which moves hierarchical attributes to queues' annotations.

k82cn:
> I made some updates on the proposal #933, which moves hierarchical attributes to queues' annotations.

Great! It's better to create a PR for the design, so we can append comments to the design and share the doc with other community :)

example/hierarchical-jobs/queues.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  weight: 1
+---
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: root-sci-dev
+  annotations:
+    "volcano.sh/hierarchy": "root/sci/dev"
+    "volcano.sh/hierarchy-weights": "1/8/2"
```
hzxuzhonghu, line 37: Is this hierarchy-weight `1/8/xx` must be same with "root/sci/prod". How can we guarantee that?

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -32,31 +36,73 @@ const PluginName = "drf"
...
 var shareDelta = 0.000001
 
+// hierarchicalNode represents the node hierarchy
+// and the corresponding weight and drf attribute
+type hierarchicalNode struct {
+	parent    *hierarchicalNode
+	attr      *drfAttr
+	weight    float64
+	saturated bool
+	hierarchy string
```
hzxuzhonghu, line 46: This is the name of the node, right?

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -305,6 +430,129 @@ func (drf *drfPlugin) updateNamespaceShare(namespaceName string, attr *drfAttr)
...
+		if drf.totalAllocated.Get(rn) < drf.totalResource.Get(rn) {
+			demandingResources[rn] = true
+
+		}
+	}
+	if len(demandingResources) == 0 {
+		return
+	}
+	drf.buildHierarchy(job, attr, hierarchy, hierarchicalWeights)
+	drf.updateHierarchicalShare(drf.hierarchicalRoot, demandingResources)
```
hzxuzhonghu, line 553: Can we optimize this calling, i can see we donot need to update once per job.

hzxuzhonghu (commented):
Can you also test on  UpdateHierarchicalShare and updateHierarchicalShare?

pkg/scheduler/api/queue_info.go
```diff
@@ -32,6 +33,9 @@ type QueueInfo struct {
 
 	Weight int32
 
+	Weights   string
+	Hierarchy string
```
hzxuzhonghu, line 37: Please add some comments

ggaaooppeenngg:
@hzxuzhonghu  I have roughly completed the job. HDRF queues are reclaimable.

hzxuzhonghu:
Great, would you make the ci happy

Thor-wl:
@ggaaooppeenngg hi, how is the process? v1.1 will be released recently, it's a main feature.

ggaaooppeenngg:
@Thor-wl  The feature could be beta.

Thor-wl:
> @Thor-wl This feature could be beta.

OK, Thanks for your contribution!

PR #945, opened 2020-07-21
Support job scale down to zero

hzxuzhonghu:
This is to support job scaling down to 0.


hzxuzhonghu:
/hold

Need to add test

hzxuzhonghu:
ping @k82cn ptal

PR #949, opened 2020-07-24
fix bug: capability validation failed when some running jobs finished or deleted

Thor-wl:
bug fix point:
1.  the right time of capability validation is in enqueue action(required validation), not the jobEqueueableFn of plugins
2.  allocated resource for queue calculated incorrectly. Considering this cenario that some jobs are in the same queue, while parts are running but parts are pending because of capability limit. When one of running job finished or deleted, next scheduler cycle will choose some pending jobs (actually podgroups) inqueue by go through all pending jobs in the queue. In each inspect period for one job, scheduler will attempt to compare the sum of allocated reource and this job request resource with capability. If the inqueue jobs request resource are not added to allocated, every check round will get a same allocate resource. In fact, ahead inqueue jobs has already locked in some resource quota in capability. 

### DIFF
--- a/pkg/scheduler/actions/enqueue/enqueue.go
+++ b/pkg/scheduler/actions/enqueue/enqueue.go
@@ -18,7 +18,6 @@ package enqueue
 
 import (
 	"k8s.io/klog"
-	"strconv"
 	"volcano.sh/volcano/pkg/apis/scheduling"
 	"volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/framework"
@@ -88,7 +87,6 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 		used.Add(node.Used)
 	}
 	idle := total.Clone().Multi(enqueue.getOverCommitFactor(ssn)).Sub(used)
-	klog.V(3).Infof("used cpu: %s, idle cpu: %s", strconv.FormatFloat(used.MilliCPU, 'E', -1, 64), strconv.FormatFloat(idle.MilliCPU, 'E', -1, 64))
 	for {
 		if queues.Empty() {
 			break
@@ -113,9 +111,6 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 			inqueue = true
 		} else {
 			minReq := api.NewResource(*job.PodGroup.Spec.MinResources)
-			klog.V(3).Infof("minResources of podgroup of job %s: %s(cpu)", job.Name, strconv.FormatFloat(minReq.MilliCPU, 'E', -1, 64))
-			klog.V(3).Infof("node idle: %s(cpu)", strconv.FormatFloat(idle.MilliCPU, 'E', -1, 64))
-			klog.V(3).Infof("JobEnqueueable: %t", ssn.JobEnqueueable(job))
 			if isCapabilitySatisfied(ssn, job) && ssn.JobEnqueueable(job) && minReq.LessEqual(idle) {
 				idle.Sub(minReq)
 				inqueue = true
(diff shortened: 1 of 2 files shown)


pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -142,3 +143,47 @@ func (enqueue *Action) getOverCommitFactor(ssn *framework.Session) float64 {
...
+							allocatedResource.Add(t.Resreq)
+						}
+					}
+				}
+			}
+		}
+	}
+	minReq := api.NewResource(*targetJob.PodGroup.Spec.MinResources)
+	klog.V(3).Infof("minResources of podgroup of job %s: %s(cpu)", targetJob.Name, strconv.FormatFloat(minReq.MilliCPU, 'E', -1, 64))
+	klog.V(3).Infof("podgroup allocated: %s(cpu)", strconv.FormatFloat(allocatedResource.MilliCPU, 'E', -1, 64))
```
hzxuzhonghu, line 187: ditto

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -142,3 +143,47 @@ func (enqueue *Action) getOverCommitFactor(ssn *framework.Session) float64 {
...
+	// If no capability is set, always enqueue the job.
+	if len(queue.Queue.Spec.Capability) == 0 {
+		klog.V(4).Infof("Capability of queue <%s> was not set, allow job <%s/%s> to Enqueue.",
+			queue.Name, targetJob.Namespace, targetJob.Name)
+		return true
+	}
+
+	// If capability is set, calculate allocated resource of queue which the target belongs to,
+	// compare the sum of allocated and Job minRequest resource with the capability resource
+	allocatedResource := api.EmptyResource()
```
hzxuzhonghu, line 160: I think this way is very inefficient as for every job, it is calculated once
suggest adding a new field in `queueAttr`

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -142,3 +143,47 @@ func (enqueue *Action) getOverCommitFactor(ssn *framework.Session) float64 {
 
 	return factor
 }
+
+func isCapabilitySatisfied(ssn *framework.Session, targetJob *api.JobInfo) bool {
```
k82cn, line 147: it's better to handle in callback function.

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -110,12 +109,14 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
...
-
 		if job.PodGroup.Spec.MinResources == nil {
 			inqueue = true
 		} else {
 			minReq := api.NewResource(*job.PodGroup.Spec.MinResources)
-			if ssn.JobEnqueueable(job) && minReq.LessEqual(idle) {
+			klog.V(3).Infof("minResources of podgroup of job %s: %s(cpu)", job.Name, strconv.FormatFloat(minReq.MilliCPU, 'E', -1, 64))
+			klog.V(3).Infof("node idle: %s(cpu)", strconv.FormatFloat(idle.MilliCPU, 'E', -1, 64))
+			klog.V(3).Infof("JobEnqueueable: %t", ssn.JobEnqueueable(job))
+			if isCapabilitySatisfied(ssn, job) && ssn.JobEnqueueable(job) && minReq.LessEqual(idle) {
```
k82cn, line 119: how about the performance?

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -81,15 +81,14 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
...
-
 	total := api.EmptyResource()
 	used := api.EmptyResource()
 	for _, node := range ssn.Nodes {
 		total.Add(node.Allocatable)
 		used.Add(node.Used)
 	}
 	idle := total.Clone().Multi(enqueue.getOverCommitFactor(ssn)).Sub(used)
-
+	klog.V(3).Infof("used cpu: %s, idle cpu: %s", strconv.FormatFloat(used.MilliCPU, 'E', -1, 64), strconv.FormatFloat(idle.MilliCPU, 'E', -1, 64))
```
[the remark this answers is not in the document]
Thor-wl: done

pkg/scheduler/actions/enqueue/enqueue.go
Thor-wl: Maybe two filed "inqueueSource" & "allocatedSource" is a good way, i need to implement it after sorting out of state transition and resource computational logic.

Thor-wl:
create a new issue to fix in a new way at issue 959

Thor-wl:
create a new issue to fix in a new way at issue959

PR #959, opened 2020-07-28
fix bug: capability validation failed when some running jobs finished or deleted

Thor-wl:
fix point:
considering the request resource of inqueue job as part of allocated resource


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -18,6 +18,7 @@ package proportion
 
 import (
 	"k8s.io/klog"
+	"volcano.sh/volcano/pkg/scheduler/actions/enqueue"
```
hzxuzhonghu, line 21: donot make actions and plugins import each other

PR #966, opened 2020-07-29
Record Inqueue job resource request in queueAttr

hzxuzhonghu:
Following #959

Session should not be mutated after initialization.

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -18,6 +18,8 @@ package proportion
 
 import (
 	"k8s.io/klog"
+
+	"volcano.sh/volcano/pkg/apis/scheduling"
 	"volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/api/helpers"
 	"volcano.sh/volcano/pkg/scheduler/framework"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -242,7 +249,11 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		minReq := api.NewResource(*job.PodGroup.Spec.MinResources)
 		// The queue resource quota limit has not reached
-		return minReq.Add(attr.allocated).Add(ssn.InqueueJobResource[job.Queue]).LessEqual(api.NewResource(queue.Queue.Spec.Capability))
+		inqueue := minReq.Add(attr.allocated).Add(attr.inqueue).LessEqual(api.NewResource(queue.Queue.Spec.Capability))
+		if inqueue {
+			attr.inqueue.Add(api.NewResource(*job.PodGroup.Spec.MinResources))
```
k82cn, line 254: `JobEnqueueable` is only a check function, we should not modify attr here.

k82cn: btw, if there're other plugins handle enqueue action, how to share this info?

PR #987, opened 2020-08-09
CI：Speed up CI jobs with parallel 

alcorj-mizar:


k82cn:
LGTM to me overall :)

.travis.yml
```diff
@@ -39,8 +45,8 @@ jobs:
       script:
         - make vcctl
         - make images
-        - make e2e-test-job
-    - stage: E2E Scheduling
+        - travis_retry make e2e-test-jobp
```
hzxuzhonghu, line 48: `travis_retry` what does this mean

Makefile
```diff
@@ -75,16 +75,22 @@ manifests: controller-gen
...
 	./hack/run-e2e-kind.sh
 
-e2e-test-scheduling:
-	E2E_TYPE=SCHEDULING ./hack/run-e2e-kind.sh
+e2e-test-schedulingbase:
+	E2E_TYPE=SCHEDULINGBASE ./hack/run-e2e-kind.sh
 
-e2e-test-job:
-	E2E_TYPE=JOB ./hack/run-e2e-kind.sh
+e2e-test-schedulingaction:
```
hzxuzhonghu, line 86: What's the diff between schedulingbase and schedulingaction

.travis.yml
alcorj-mizar: If make e2e-test-jobp failed, it will retry for 3 times.

Makefile
alcorj-mizar: schedulingbase is the basic scheduling test, schedulingaction is just include action plugin.

.travis.yml
alcorj-mizar: It is a small probability if kind create cluster will failed of write configuration, and this command will retry.

PR #996, opened 2020-08-13
Fix ssh authorize key and remove `no-root` flag

hzxuzhonghu:


k82cn:
what issue are we going to handle in this PR?

pkg/controllers/job/plugins/ssh/ssh.go
```diff
@@ -188,8 +183,6 @@ func (sp *sshPlugin) secretName(job *batch.Job) string {
 
 func (sp *sshPlugin) addFlags() {
 	flagSet := flag.NewFlagSet(sp.Name(), flag.ContinueOnError)
-	// TODO: deprecate no-root
-	flagSet.BoolVar(&sp.noRoot, "no-root", sp.noRoot, "The ssh user, --no-root is common user")
```
hzxuzhonghu, line 192: Here it was deprecated long ago, actually meaningless

pkg/controllers/job/plugins/ssh/ssh.go
```diff
@@ -121,7 +115,7 @@ func (sp *sshPlugin) mountRsaKey(pod *v1.Pod, job *batch.Job) {
 				Path: SSHRelativePath + "/" + SSHPublicKey,
 			},
 			{
-				Key:  SSHPublicKey,
+				Key:  SSHAuthorizedKeys,
```
hzxuzhonghu, line 118: And here the authorized_keys file should be from configmap's SSHAuthorizedKeys, though SSHPublicKey and SSHAuthorizedKeys are of same value

Issue #1001, opened 2020-08-17
Volcano scheduler panic if not set minResources in podgroup


DeliangFan:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

Volcano scheduler panic if not specific the minResources in podgroup when scheduling.

goroutine 186 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /opt/tiger/compile_path/src/code.byted.org/tce/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:55 +0x105
panic(0x1692540, 0x27565c0)
        /usr/local/go/src/runtime/panic.go:969 +0x166
volcano.sh/volcano/pkg/scheduler/plugins/proportion.(*proportionPlugin).OnSessionOpen(0xc0005af0e0, 0xc0002c0a00)
        /opt/tiger/compile_path/src/code.byted.org/tce/volcano/pkg/scheduler/plugins/proportion/proportion.go:107 +0x51d
volcano.sh/volcano/pkg/scheduler/framework.OpenSession(0x1b5f340, 0xc0003e6c00, 0xc000558c30, 0x2, 0x2, 0x0, 0x0, 0x0, 0x0)
        /opt/tiger/compile_path/src/code.byted.org/tce/volcano/pkg/scheduler/framework/framework.go:48 +0x190
volcano.sh/volcano/pkg/scheduler.(*Scheduler).runOnce(0xc00030d570)
        /opt/tiger/compile_path/src/code.byted.org/tce/volcano/pkg/scheduler/scheduler.go:76 +0x15b
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0xc0005884d0)
        /opt/tiger/compile_path/src/code.byted.org/tce/volcano/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:155 +0x5f
k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0xc0005884d0, 0x1b04c40, 0xc000558ae0, 0x1, 0x0)
        /opt/tiger/compile_path/src/code.byted.org/tce/volcano/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:156 +0xa3
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc0005884d0, 0x3b9aca00, 0x0, 0x1, 0x0)
[...]

PR #1002, opened 2020-08-17
fix panic if minResource not set

DeliangFan:
related-bug: #1001

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -49,6 +49,15 @@ type queueAttr struct {
 	inqueue *api.Resource
 }
 
+// GetJobMinResources return the min resources of podgroup.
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -247,11 +255,11 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			return true
 		}
 
-		minReq := api.NewResource(*job.PodGroup.Spec.MinResources)
+		minReq := pp.GetJobMinResources(job.PodGroup.Spec)
```
hzxuzhonghu, line 258: I think we should  return directly when minReq = nil

hzxuzhonghu:
/approve

Thanks for the fix, we need to also backport this to release-1.0 and 0.4

hzxuzhonghu:
@DeliangFan Can you help cherry-pick to  release-1.0 and 0.4

Issue #991, opened 2020-08-11
enqueue conflict with preempt


hzxuzhonghu:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind bug

**What happened**:

Create a low priority job to occupy all the cluster resource, then create a high priority job, we can find that the high priority job can never become `inqueue` status.

This is actually a regression since enqueue feature introduced. 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hzxuzhonghu:
cc @william-wang

hzxuzhonghu:
I donot think it is worth coupling enqueue with preempt/reclaim. Currently, i want to declare it in the release note.  If users need preempt, enqueue should be disabled.

Thor-wl:
This bug will be fix in v1.1.1 in late Dec

PR #1003, opened 2020-08-17
Fix enqueue conflicts with preempt and reclaim

hzxuzhonghu:
Fix: #991 

This makes admission controller allow creating pods when podgroup is pending if the `enqueue` action is not enabled.


k82cn:
it's better to have a central configuration for all components :)

hzxuzhonghu:
TBH, i have not come to a better idea. I have thought about allowing admission controller to read scheduler config as an alternative.  Not sure if it is good.

Issue #1007, opened 2020-08-17
cleanup scheduler api


hzxuzhonghu:
we need to remove `pkg/apis/scheduling/types.go`, which was introduced to support api convert between different versions before. 
And as we have removed the old gvk, we can remove it. Which can bring some performance improvement as well

PR #1012, opened 2020-08-20
Scheduler api cleanup

hzxuzhonghu:
Fix: #1007

The internal API was introduced when we have two groups of API, and in that case, CRD can not be used to convert them.

This removes all these codes.


k82cn:
/hold

Why introduce version into scheduler core ?

hzxuzhonghu:
No one else should use an internal version instead of kube-apiserver.  The internal version is used to do validation and convert between two versions. Obviously, we donot want to convert v1beta1 PodGroup/ Queue to an internal version. This step is useless for a client

k82cn:
we decoupled k8s version from volcano version.

hzxuzhonghu:
IMO, it is not decouple, it is duplicate and introduce extra conversion cost.  cc @kevin-wangzefeng

hzxuzhonghu:
@k82cn @william-wang Scheduler as a client of k8s, we donot need to keep a private version and convert between k8s version and our own. And we donot need to list-watch multi-version resources even if we have in the future.  This is much like the way deployment controller does.

PR #1013, opened 2020-08-20
Improve performance: remove priorityclass

hzxuzhonghu:
Remove the mutex simultaneously, which is used to protect priorityclass

### DIFF
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"hash"
 	"hash/fnv"
-	"sync"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


k82cn:
Any data about the improvement?

hzxuzhonghu:
I have no test data, but from the mutex race in `calcPGMinResources`, it should have some improvement

k82cn:
> I have no test data, but from the mutex race in `calcPGMinResources`, it should have some improvement

please provide performance benchmark for this pr.

hzxuzhonghu:
This is a minor improve, i do not think it is worth to do benchmark for every improvement. And BTW we do not have any tools and   cluster to do that yet.

k82cn:
that's suprised to me that we got this PR merged without addressing related comments.

>  And BTW we do not have any tools and cluster to do that yet.

Any issue to trace this requirement?

Issue #1015, opened 2020-08-27
Propose to watch scheduler config file


hzxuzhonghu:
/kind feature

Instead of  loading it per scheduling period uses file watcher to watch it. And only reload when it changed

PR #1016, opened 2020-08-27
Add file watcher for scheduler config file

hzxuzhonghu:
Address #1015 

Have manually tested it since it is difficult to add a test

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -17,12 +17,17 @@ limitations under the License.
 package scheduler
 
 import (
+	"fmt"
+	"path/filepath"
+	"sync"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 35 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -86,19 +103,58 @@ func (pc *Scheduler) runOnce() {
 
 func (pc *Scheduler) loadSchedulerConf() {
 	var err error
+	pc.once.Do(func() {
```
k82cn, line 114: Why `once`, which make it call only once?

hzxuzhonghu: Yeah, this is the default configuration, we need load it only once, and if user provides a config file, then load the provided one. If the provoded one is invalid, fallback to the default one and log warning.

k82cn: Ok, that makes sense to me :)

PR #1020, opened 2020-08-28
rm unnecessary jobcache.JobKeyByReq

xiaoanyunfei:


hzxuzhonghu (approved):

pkg/controllers/job/job_controller.go
```diff
@@ -304,7 +304,7 @@ func (cc *jobcontroller) processNextReq(count uint32) bool {
 
 	klog.V(3).Infof("Try to handle request <%v>", req)
 
-	jobInfo, err := cc.cache.Get(jobcache.JobKeyByReq(&req))
+	jobInfo, err := cc.cache.Get(key)
```
hzxuzhonghu, line 307: good catch

Issue #1036, opened 2020-09-08
Werid Resource Less behavior


zen-xu:
/kind bug

**What happened**:

I want to compare to `Resource`, here is the code

```go
package main

import (

	"volcano.sh/volcano/pkg/scheduler/api"
	"volcano.sh/volcano/pkg/scheduler/util"
)


func main() {
	resource1 := api.NewResource(util.BuildResourceList("0m", "0"))
	resource2 := api.NewResource(util.BuildResourceList("101m", "0"))
	println(resource1.Less(resource2))  // output is `flase`
}
```

**What you expected to happen**:
I think this should be `true`

**How to reproduce it (as minimally and precisely as possible)**:
Just run the code

hzxuzhonghu:
Thanks for reporting

k82cn:
it's better to replace `float64` with `big.Rat` in `Resource`

hzxuzhonghu:
/reopen
This is not fixed yet

Thor-wl:
I've analysed this bug. The root reason as follows:
1. Ignoring the comparison: 0=0
2. Ignoring the comparison: default argument `minMilliScalarResources` with 0

PR #1037, opened 2020-09-09
Fix resource less method

hzxuzhonghu:


hzxuzhonghu:
Fix: #1036

k82cn:
Can you help to add some test case for the differentiation?

hzxuzhonghu:
@k82cn This is ready to go

pkg/scheduler/api/resource_info.go
```diff
@@ -224,40 +224,46 @@ func (r *Resource) Multi(ratio float64) *Resource {
...
+				isLess = true
+			}
 		}
 	}
 
-	return true
+	return isLess
 }
 
 // LessEqualStrict checks whether a resource is less or equal than other
```
shinytang6, line 269: l think for the issue #1036, maybe function `LessEqualStrict ` and `LessEqual ` can be changed together)

zen-xu:
This pr still has bug

```go
func main() {
	allocated := &api.Resource{
		MilliCPU:        1000,
		Memory:          2000,
	}

	deserved := &api.Resource{
		MilliCPU:        2000,
		Memory:          3000,
		ScalarResources: map[v1.ResourceName]float64{
			"nvidia.com/gpu": 4500,
			"hugepages-1Gi": 0,
			"hugepages-2Mi": 0,
		},
	}
	fmt.Println(allocated.Less(deserved))  // false, should be true
}
```

k82cn:
please use Rat (xref: https://golang.org/src/math/big/rat.go ) for that.

PR #1040, opened 2020-09-10
Optimize Snapshot

hzxuzhonghu:
Concurrently working on snapshot and control the go routines number

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -229,7 +229,8 @@ func (ni *NodeInfo) AddTask(task *TaskInfo) error {
 
 	key := PodKey(task.Pod)
 	if _, found := ni.Tasks[key]; found {
-		return nil
+		return fmt.Errorf("task <%v/%v> already on node <%v>",
+			task.Namespace, task.Name, ni.Name)
 	}
 
 	// Node will hold a copy of task to make sure the status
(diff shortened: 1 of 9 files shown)


pkg/scheduler/api/node_info.go
```diff
@@ -229,8 +229,7 @@ func (ni *NodeInfo) AddTask(task *TaskInfo) error {
 
 	key := PodKey(task.Pod)
 	if _, found := ni.Tasks[key]; found {
-		return fmt.Errorf("task <%v/%v> already on node <%v>",
-			task.Namespace, task.Name, ni.Name)
+		return nil
```
k82cn, line 232: why return nil?

hzxuzhonghu: I am not sure the logic before, it looks this happens when calling AddTask multi-times. This should be a warning, and we should fix the call site.

This pr is just trying to verify if this way is viable

k82cn: > This should be a warning, and we should fix the call site.

If we return nil, how to know the issue there?

hzxuzhonghu:
The benchmark result shows a 30% improve

before:

```
BenchmarkSnapShot-4   	      37	 289957413 ns/op
```

After 
```
BenchmarkSnapShot-4   	      54	 197747549 ns/op

```

PR #1043, opened 2020-09-11
Cleanup useless fields of JobInfo

hzxuzhonghu:
For performance, remove below fields of `JobInfo`
```
Allocated    
TotalRequest 
NodesFitDelta 
```

I have searched these fields are set but no read at all.


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -220,9 +212,6 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 						task.UID, node.Name, ssn.UID, err)
 				}
 			} else {
-				//store information about missing resources
-				job.NodesFitDelta[node.Name] = node.Idle.Clone()
```
k82cn, line 224: how about the impact of metrics?

hzxuzhonghu: this is not used at all

PR #1044, opened 2020-09-14
big job resource reservation

Thor-wl:


Thor-wl:
feature development and test has been finished. Please review the PR ignoring the log. Finally, i'll remove  the noisy log.

pkg/scheduler/actions/elect/elect.go
```diff
@@ -0,0 +1,52 @@
...
+// Execute selects the target job which is of the highest priority and waits for the longest time.
+func (alloc *Action) Execute(ssn *framework.Session) {
+	klog.V(3).Infof("Enter Elect ...")
+	defer klog.V(3).Infof("Leaving Elect ...")
+
+	if util.Reservation.TargetJob == nil {
+		klog.V(3).Infof("Start select Target Job...")
+		var pendingJobs []*api.JobInfo
+		for _, job := range ssn.Jobs {
+			klog.V(3).Infof("job: %s, pgStatus: %s", job.Name, job.PodGroup.Status.Phase)
```
wpeng102, line 37: suggest use log `V(4)` to avoid scheduler log  grows too quickly.  Please also check if log level is appropriate in other code block.

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -237,3 +247,17 @@ func ValidateVictims(preemptor *api.TaskInfo, node *api.NodeInfo, victims []*api
 	}
 	return nil
 }
+
+// ResourceReservation is struct used for resource reservation
+type ResourceReservation struct {
```
wpeng102, line 252: If only used in package internal, suggest change to `resourceReservation `

pkg/scheduler/actions/elect/elect.go
Thor-wl: > suggest use log `V(4)` to avoid scheduler log grows too quickly. Please also check if log level is appropriate in other code block.

OK，I'll check all logs and remove noisy log later.

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -128,6 +151,24 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
...
 	}
+	// if target job exists, judge whether it can be inqueue or not
+	targetJob := util.Reservation.TargetJob
+	if targetJob != nil && targetJob.PodGroup.Status.Phase == scheduling.PodGroupPending && len(util.Reservation.LockedNodes) != 0 {
+		klog.V(3).Infof("Start to deal with Target Job")
+		minReq := api.NewResource(*util.Reservation.TargetJob.PodGroup.Spec.MinResources)
+		klog.V(3).Infof("Target Job request CPU: %f", minReq.MilliCPU)
+		lockedIdle := api.EmptyResource()
+		for _, lockNode := range util.Reservation.LockedNodes {
+			lockedIdle.Add(lockNode.Idle)
```
wpeng102, line 162: why not reuse `lockedNodesIdle`

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -83,11 +91,22 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 
 	total := api.EmptyResource()
 	used := api.EmptyResource()
+	lockedNodesIdle := api.EmptyResource()
+	if util.Reservation.TargetJob != nil && len(util.Reservation.LockedNodes) != 0 {
+		for _, node := range util.Reservation.LockedNodes {
+			lockedNodesIdle.Add(node.Idle)
+			klog.V(3).Infof("locked node: %s, idle CPU: %f", node.Name, node.Idle)
```
wpeng102, line 98: why only print cpu resource in this file?

wpeng102:
@Thor-wl 
1) If enable this feature, when start node reservation? Does it mean when there is a big job coming and pending after one scheduler cycle, it will start node reservation? 

2) In this design, the selected big job can only use locked node resource, other jobs use other node resource, right? 

3) I found it handle the selected big job after other jobs in  `enqueue` process, consider the following situation: 
In one schedule cycle, the total idle CPU is 4, reserved cpu is 2, the selected big job request 3 cpus and other jobs request 2 cpus. 
If we handle the big job first, it could be dispatched, but if we handle the big job after other jobs, the big job can not be dispatched in this schedule cycle.  Why not handle other jobs before the big job in schedule cycle?

pkg/scheduler/actions/enqueue/enqueue.go
Thor-wl: This log will be removed because i take cpu as test metric when self test

Thor-wl:
> @Thor-wl
> 
> 1. If enable this feature, when start node reservation? Does it mean when there is a big job coming and pending after one scheduler cycle, it will start node reservation?
> 2. In this design, the selected big job can only use locked node resource, other jobs use other node resource, right?
> 3. I found it handle the selected big job after other jobs in  `enqueue` process, consider the following situation:
>    In one schedule cycle, the total idle CPU is 4, reserved cpu is 2, the selected big job request 3 cpus and other jobs request 2 cpus.
>    If we handle the big job first, it could be dispatched, but if we handle the big job after other jobs, the big job can not be dispatched in this schedule cycle.  Why not handle other jobs before the big job in schedule cycle?

1. Reservation will start work when scheduler starts in current logic. It should be an optimization but a hard decision because it's hard to define how long is a suitable point.
2. Well, not really. The selected big job can use all node resource but other jobs can only use unlocked nodes.
3. Yeah, you got the point. Let me optimize it.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -94,7 +94,18 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	pendingTasks := map[api.JobID]*util.PriorityQueue{}
 
 	allNodes := util.GetNodeList(ssn.Nodes)
-
+	unlockedNodes := allNodes
```
shinytang6, line 97: why not initialize it to empty here and remove latter `unlockedNodes = unlockedNodes[0:0]`

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -151,7 +162,14 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 		}
 
 		job := jobs.Pop().(*api.JobInfo)
-		if _, found := pendingTasks[job.UID]; !found {
+		var nodes []*api.NodeInfo
```
shinytang6, line 165: `nodes` are not used later? `allNodes` should be set to `nodes`?

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -107,6 +126,10 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 			continue
 		}
 		job := jobs.Pop().(*api.JobInfo)
+		if util.Reservation.TargetJob != nil && job.UID == util.Reservation.TargetJob.UID {
```
shinytang6, line 129: l think this judgment can be abstracted into a function, it is used in many places

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -128,6 +151,24 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 		// Added Queue back until no job in Queue.
 		queues.Push(queue)
 	}
+	// if target job exists, judge whether it can be inqueue or not
+	targetJob := util.Reservation.TargetJob
+	if targetJob != nil && targetJob.PodGroup.Status.Phase == scheduling.PodGroupPending && len(util.Reservation.LockedNodes) != 0 {
+		klog.V(3).Infof("Start to deal with Target Job")
+		minReq := api.NewResource(*util.Reservation.TargetJob.PodGroup.Spec.MinResources)
```
shinytang6, line 158: util.Reservation.TargetJob.PodGroup.Spec.MinResources => targetJob.PodGroup.Spec.MinResources

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -128,6 +151,24 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
...
+			lockedIdle.Add(lockNode.Idle)
+		}
+		klog.V(3).Infof("locked idle CPU: %f", lockedIdle.MilliCPU)
+		idle = idle.Add(lockedIdle)
+		if minReq.LessEqual(idle) {
+			klog.V(3).Infof("Turn Target Job phase to Inqueue")
+			util.Reservation.TargetJob.PodGroup.Status.Phase = scheduling.PodGroupInqueue
+			ssn.Jobs[util.Reservation.TargetJob.UID] = util.Reservation.TargetJob
+		}
+	}
```
shinytang6, line 171: have a question about this block of code: now that the locked node is determined, do we need to calculate `lockedNodesIdle` again? Why not use lockedNodesIdle calculated ahead?

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -128,6 +151,24 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
...
+		lockedIdle := api.EmptyResource()
+		for _, lockNode := range util.Reservation.LockedNodes {
+			lockedIdle.Add(lockNode.Idle)
+		}
+		klog.V(3).Infof("locked idle CPU: %f", lockedIdle.MilliCPU)
+		idle = idle.Add(lockedIdle)
+		if minReq.LessEqual(idle) {
+			klog.V(3).Infof("Turn Target Job phase to Inqueue")
+			util.Reservation.TargetJob.PodGroup.Status.Phase = scheduling.PodGroupInqueue
+			ssn.Jobs[util.Reservation.TargetJob.UID] = util.Reservation.TargetJob
```
shinytang6, line 169: util.Reservation.TargetJob => targetJob

pkg/scheduler/actions/factory.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package actions
 
 import (
+	"volcano.sh/volcano/pkg/scheduler/actions/elect"
+	"volcano.sh/volcano/pkg/scheduler/actions/reserve"
```
shinytang6, line 21: it's better to put these packages together with the following plugin packages :)

pkg/scheduler/plugins/reservation/reservation.go
```diff
@@ -0,0 +1,147 @@
...
+	return &reservationPlugin{pluginArguments: arguments}
+}
+
+func (rp *reservationPlugin) Name() string {
+	return PluginName
+}
+
+func (rp *reservationPlugin) OnSessionOpen(ssn *framework.Session) {
+	// select the job which has the highest priority and waits for the longest duration
+	targetJobFn := func(jobs []*api.JobInfo) *api.JobInfo {
```
shinytang6, line 46: There is a question worth discussing, why use `targetJob` the whole PR while not something like `bigJob` instead?

pkg/scheduler/plugins/reservation/reservation.go
```diff
@@ -0,0 +1,147 @@
...
+	ssn.AddReservedNodesFn(rp.Name(), reservedNodesFn)
+}
+
+func (rp *reservationPlugin) OnSessionClose(ssn *framework.Session) {}
+
+func (rp *reservationPlugin) getHighestPriority(jobs []*api.JobInfo) int32 {
+	if len(jobs) == 0 {
+		return -1
+	}
+	highestPriority := jobs[0].Priority
```
shinytang6, line 73: highestPriority := -1

pkg/scheduler/plugins/reservation/reservation.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, job := range jobs {
+		klog.V(3).Infof("job name: %s", job.Name)
+		klog.V(3).Infof("job name: %d", job.Priority)
+		if job.Priority > highestPriority {
```
shinytang6, line 77: if highestPriority  == -1 || job.Priority > highestPriority

pkg/scheduler/plugins/reservation/reservation.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+	return highestPriorityJobs
+}
+
+func (rp *reservationPlugin) getTargetJob(jobs []*api.JobInfo) *api.JobInfo {
+	if len(jobs) == 0 {
+		return nil
+	}
+	maxWaitDuration := rp.getJobWaitingTime(jobs[0])
```
shinytang6, line 102: ditto, I think the above comment‘s practice is more elegant

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -237,3 +247,17 @@ func ValidateVictims(preemptor *api.TaskInfo, node *api.NodeInfo, victims []*api
...
+	LockedNodes map[string]*api.NodeInfo
+}
+
+// NewResourceReservation is used to create global instance
+func NewResourceReservation() *ResourceReservation {
+	return &ResourceReservation{
+		TargetJob:   nil,
+		LockedNodes: map[string]*api.NodeInfo{},
+	}
+}
```
shinytang6, line 263: There is another question here, why not put the `ResourceReservation` information in the session

pkg/scheduler/actions/reserve/reserve.go
```diff
@@ -0,0 +1,50 @@
...
+	util.Reservation.TargetJob = ssn.Jobs[util.Reservation.TargetJob.UID]
+
+	if !util.Reservation.TargetJob.Ready() {
+		ssn.ReservedNodes()
+	} else {
+		klog.V(3).Infof("Target Job has been scheduled. Reset Target Job")
+		util.Reservation.TargetJob = nil
+		for node := range util.Reservation.LockedNodes {
+			delete(util.Reservation.LockedNodes, node)
+		}
```
shinytang6, line 45: why not using util.Reservation.LockedNodes = nil

pkg/scheduler/plugins/reservation/reservation.go
Thor-wl: > There is a question worth discussing, why using `targetJob` the whole PR while not something like `bigJob` instead?

In fact, selected job is at the high priority and waits for a long time. So we consider it as target job. Usually, it is a big job which keeps pending for lack of resource.  But that's not the only case.

pkg/scheduler/actions/enqueue/enqueue.go
Thor-wl: done

pkg/scheduler/actions/factory.go
Thor-wl: done

pkg/scheduler/actions/reserve/reserve.go
Thor-wl: > why not using util.Reservation.LockedNodes = nil

They are equal in effect but GC reserves memory immedicately with delete.

pkg/scheduler/util/scheduler_helper.go
Thor-wl: > If only used in package internal, suggest change to `resourceReservation `

It's only used in internal indeed. I'm not sure whether it will be used externally.

pkg/scheduler/util/scheduler_helper.go
Thor-wl: > There is another question here, why not put the `ResourceReservation` information in the session?

Good question! In the design, ResourceReservation must be shared accross session.

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -38,6 +38,13 @@ const baselinePercentageOfNodesToFind = 50
 
 var lastProcessedNodeIndex int
 
+// Reservation is used to record target job and locked nodes
+var Reservation *ResourceReservation
+
+func init() {
```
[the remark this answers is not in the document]
Thor-wl: That's right. I've tried to find out a good place to store Reservation object for the data must be shared cross sessions. Unfortunally, i was failed.

pkg/scheduler/actions/reserve/reserve.go
```diff
@@ -0,0 +1,50 @@
...
+		klog.V(3).Infof("No target job, skip reserve action")
+		return
+	}
+	// if target job has not been scheduled, select a locked node for it
+	// else reset target job and locked nodes
+	util.Reservation.TargetJob = ssn.Jobs[util.Reservation.TargetJob.UID]
+
+	if !util.Reservation.TargetJob.Ready() {
+		ssn.ReservedNodes()
+	} else {
```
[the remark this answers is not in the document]
Thor-wl: > please change to following style.
> if !util.Reservation.TargetJob.Ready() {
> ssn.ReservedNodes()
> return
> }
> xxxx

I tried but lint failed.

william-wang (commented):
Some changes like util.Reservation.TargetJob need to refatcor in future release and the e2e need to be added as well.

Thor-wl:
> Some changes like util.Reservation.TargetJob need to refatcor in future release and the e2e need to be added as well.

OK

PR #1061, opened 2020-09-23
Remove creationTimestamp comparation in jobOrderFn of gang plugin

zionwu:
The creationTimestamp comparation in jobOrderFn of gang plugin should be removed, otherwise, the scheduler will ignore the jobOrderFn of other plugins.  The creationTimestamp comparation should be the last approach to use if all jobOrderFn return 0.

### DIFF
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -119,17 +119,6 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
 			return -1
 		}
 
-		if !lReady && !rReady {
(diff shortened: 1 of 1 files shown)


Thor-wl (commented):
/lgtm

hzxuzhonghu:
cc @carmark

carmark:
LGTM

zionwu:
@k82cn 
There is a duplicated CI check with status `Queued`. I clicked the `Details` and found all checks are queued, but I found the same jobs are actually passed in travis-ci: https://travis-ci.com/github/volcano-sh/volcano/builds/186153790.

Do you know how to deal with it?

k82cn:
I restart the build.

Issue #1031, opened 2020-09-03
Add maxRetries parameters to controller


merryzhou:
/kind feature

**What happened**:

```
const (
	// maxRetries is the number of times a volcano job will be retried before it is dropped out of the queue.
	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
	// a volcano job is going to be requeued:
	//
	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
	maxRetries = 15
)
```

the maxRetries is hard coded.

And In my scenario, I hope the job controller keep retrying, just like other k8s controllers.

So, is it ok to add  a maxRetries parameter, default value is 15, when set to -1, it means retry endlessly

**Environment**:
- Volcano Version: v1.0.1

hzxuzhonghu:
I am ok with this requirement and welcome a pr addressing this.

liuchintao:
@hzxuzhonghu hello, can I ask a question? I was puzzled that why Volcano Job Controller only tried to execute action 15 times.
Are there any restrictions on Job Controller or job cache?

hzxuzhonghu:
Actually we think 15 is enough, the retry can happen when error occur, maybe `ratelimit`, `transient network issue`. With limited retry, it can recover automatically.


>Are there any restrictions on Job Controller or job cache?

I am not sure i understand your question

liuchintao:
Thanks for your reply~

As you said, 15 is sufficient in most cases, like ratelimit, transient network issue. 

However, what is the expected behavior, when it still failed after 15 retries? I think it is possible.

For example, in my namespace `foo`, there is a `resource quota` for 5 cpus.

```yaml
  apiVersion: v1
  kind: ResourceQuota
  metadata:
    name: quota-cpu
    namespace: foo
  spec:
    hard:
      cpu: "5"
```

And my mpijob just need 5 cpu, a mpi-launcher need 1 cpu, 2 mpi-workers need 2 cpu separately. Everything looks good.

But before I start my mpi job, I start a pod that consumes 2 cpu. 
Now I just have 2 mpi pods, a launcher and a worker, I have to wait for pod complete and release resources.
But now job controller reach 15 retries, and drop this `req` from worker's queue, 
and I never get the last pod any more, even the former pod releases resources.

[...]

PR #1062, opened 2020-09-23
feat: job controller would keep trying to Execute action

liuchintao:
more info: https://github.com/volcano-sh/volcano/issues/1031

Job controller just retry 15 times to execute job's action, it's enough in most cases.

After job controller retry and failed 15 times, worker just drop this req from working queue but do nothing.
If worker's failed times exceeds limited max retry times when executes `syncJob` action,
the job would stay in `pending` forever, and the resources which hold by job's pod would not be released.

### DIFF
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -338,6 +338,10 @@ func (cc *jobcontroller) processNextReq(count uint32) bool {
 		cc.recordJobEvent(jobInfo.Job.Namespace, jobInfo.Job.Name, batchv1alpha1.ExecuteAction, fmt.Sprintf(
 			"Job failed on action %s for retry limit reached", action))
 		klog.Warningf("Dropping job<%s/%s> out of the queue: %v because max retries has reached", jobInfo.Job.Namespace, jobInfo.Job.Name, err)
+		klog.Warningf("Terminating job<%s/%s> and releasing resources", jobInfo.Job.Namespace, jobInfo.Job.Name)
+		if err = st.Execute(busv1alpha1.TerminateJobAction); err != nil {
+			klog.Errorf("Failed to terminate job<%s/%s>, because: %v", jobInfo.Job.Namespace, jobInfo.Job.Name, err)
+		}
 	}
 
 	// If no error, forget it.


config/crd/bases/batch.volcano.sh_jobs.yaml
```diff
@@ -41,6 +41,9 @@ spec:
           description: Specification of the desired behavior of the volcano job, including
             the minAvailable
           properties:
+            keepSchedule:
```
Thor-wl, line 44: Can you describe your usage scene?

liuchintao: For example, in my namespace foo, there is a resource quota for 5 cpus.

```yaml
  apiVersion: v1
  kind: ResourceQuota
  metadata:
    name: quota-cpu
    namespace: foo
  spec:
    hard:
      cpu: "5"
```

And my mpijob just need 5 cpu, a mpi-launcher need 1 cpu, 2 mpi-workers need 2 cpu separately. Everything looks good.
But before I start my mpi job, I start a pod that consumes 2 cpu.

Now I just have 2 mpi pods, a launcher and a worker, I have to wait for pod completes and releases resources.
But now, job controller reaches 15 retries, and drops this req from worker's queue.
I can't get the last pod any more, even the former pod releases resources.
And the job also not releases the resources it holds.

Thor-wl: > For example, in my namespace foo, there is a resource quota for 5 cpus.
> 
> ```yaml
>   apiVersion: v1
>   kind: ResourceQuota
>   metadata:
>     name: quota-cpu
>     namespace: foo
>   spec:
>     hard:
>       cpu: "5"
> ```
> 
> And my mpijob just need 5 cpu, a mpi-launcher need 1 cpu, 2 mpi-workers need 2 cpu separately. Everything looks good.
> But before I start my mpi job, I start a pod that consumes 2 cpu.
> 
> Now I just have 2 mpi pods, a launcher and a worker, I have to wait for pod completes and releases resources.
> But now, job controller reaches 15 retries, and drops this req from worker's queue.
> I can't get the last pod any more, even the former pod releases resources.
> And the job also not releases the resources it holds.

Perhaps it seems reasonable in some scenes as you describe. But it may cause dead lock. For example, there are 10 cpus in the cluster. And another end-user launch a job which is same with your condition. It happens that he is also running part of his tasks and waiting your running tasks to release resouce. If you both can be wait for each other forever, your jobs both will be unfinished forever. Scheduler must take some actions to release some resources to resolve the problem. That's the reason why we set 15 tries.

liuchintao: > If you both can be wait for each other forever, your jobs both will be unfinished forever. Scheduler must take some actions to release some resources to resolve the problem.

Yeah, you are right, this would lead to dead lock, maybe I should resolve this problem from the perspect of "How to release resources after controller retries 15 times".

But now, job controller just drop job out of  the worker's queue and do nothing. 
In the above case:

1. cpus are allocated, but no job works;
2. controller drops the `sync request` out of the queue, IMO job's status can not be synced;
3. cpus only could be released until job is deleted.

How about releasing job's resources after retry 15 times?
Every `xxxState` should `Execute(TerminateJobAction)` after failed times.

Thor-wl: > > If you both can be wait for each other forever, your jobs both will be unfinished forever. Scheduler must take some actions to release some resources to resolve the problem.
> 
> Yeah, you are right, this would lead to dead lock, maybe I should resolve this problem from the perspect of "How to release resources after controller retries 15 times".
> 
> But now, job controller just drop job out of the worker's queue and do nothing.
> In the above case:
> 
> 1. cpus are allocated, but no job works;
> 2. controller drops the `sync request` out of the queue, IMO job's status can not be synced;
> 3. cpus only could be released until job is deleted.
> 
> How about releasing job's resources after retry 15 times?
> Every `xxxState` should `Execute(TerminateJobAction)` after failed times.

Please show me your scheduler configuration. I guess you have not configed Gang plugin.

liuchintao: this is my scheduler configuration

```yaml
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
      - name: conformance
    - plugins:
      - name: drf
      - name: predicates
      - name: proportion
      - name: nodeorder
      - name: binpack
kind: ConfigMap
metadata:
  namespace: volcano-system
```

Thor-wl: > this is my scheduler configuration
> 
> ```yaml
> apiVersion: v1
> data:
>   volcano-scheduler.conf: |
>     actions: "enqueue, allocate, backfill"
>     tiers:
>     - plugins:
>       - name: priority
>       - name: gang
>       - name: conformance
>     - plugins:
>       - name: drf
>       - name: predicates
>       - name: proportion
>       - name: nodeorder
>       - name: binpack
> kind: ConfigMap
> metadata:
>   name: volcano-scheduler-configmap
>   namespace: volcano-system
> ```

How do you prove the behavior above? As i know, if gang plugin is configured, only allocatable resource satisfies at leaset minAvailable tasks can the job be running. Namly, minAvailable tasks will be allocated resource at the same time.

pkg/controllers/job/job_controller.go
```diff
@@ -338,6 +338,8 @@ func (cc *jobcontroller) processNextReq(count uint32) bool {
 		cc.recordJobEvent(jobInfo.Job.Namespace, jobInfo.Job.Name, batchv1alpha1.ExecuteAction, fmt.Sprintf(
 			"Job failed on action %s for retry limit reached", action))
 		klog.Warningf("Dropping job<%s/%s> out of the queue: %v because max retries has reached", jobInfo.Job.Namespace, jobInfo.Job.Name, err)
+		err = st.Execute(busv1alpha1.TerminateJobAction)
```
k82cn, line 341: please also check the return error of this execution.

Thor-wl:
> /approve

Maybe we should not merge this PR for risk of dead lock. We should consider it more.

config/crd/bases/batch.volcano.sh_jobs.yaml
liuchintao: @Thor-wl  Sorry, it took so long to reply.

We have deployed the a new environment with our customed scheduler. Is it useful that I describe the situation I met before?

The follow is my node's capacity, cpu are sufficient.

```log
Capacity:
 cpu:                                 48
 ephemeral-storage:                   2228605748Ki
 hugepages-1Gi:                       0
 hugepages-2Mi:                       0
 memory:                              263602412Ki
 pods:                                500
```

I want to test the behavior of scheduler with k8s resource quota, so I make a namespace that just be allocated 5 cpus.

```yaml
apiVersion: v1
kind: List
items:
- apiVersion: v1
  kind: ResourceQuota
  metadata:
    name: test-quota-cpu
    namespace: foo
  spec:
    hard:
      cpu: "5"
---
apiVersion: v1
kind: Pod
metadata:
  name: consume-2-cpu
  namespace: foo
spec:
  containers:
  - name: test
    image: busybox:latest
    command: ["/bin/sh", "-c", "tail -f /dev/null"]
    resources:
      requests:
        cpu: 2
      limits:
        cpu: 2
---
kind: Job
metadata:
  name: big-cpu-mpi-job
  namespace: foo
spec:
  minAvailable: 3
  schedulerName: volcano
  keepSchedule: true
  plugins:
    ssh: []
    svc: []
  tasks:
    - replicas: 1
      name: mpimaster
      policies:
        - event: TaskCompleted
          action: CompleteJob
      template:
        spec:
          containers:
            - command:
                - /bin/sh
                - -c
                - |
                  MPI_HOST=`cat /etc/volcano/mpiworker.host | tr "\n" ","`;
                  mkdir -p /var/run/sshd; /usr/sbin/sshd;
                  mpiexec --allow-run-as-root --host ${MPI_HOST} -np 2 mpi_hello_world > /home/re;
              image: example-mpi:0.0.1
              name: mpimaster
              ports:
                - containerPort: 22
                  name: mpijob-port
              workingDir: /home
              resources:
                limits:
                  cpu: 1
          restartPolicy: OnFailure
    - replicas: 2
      name: mpiworker
      template:
        spec:
          containers:
            - command:
                - /bin/sh
                - -c
                - |
                  mkdir -p /var/run/sshd; /usr/sbin/sshd -D;
              image: example-mpi:0.0.1
              name: mpiworker
              ports:
                - containerPort: 22
                  name: mpijob-port
              workingDir: /home
              resources:
                limits:
                  cpu: 2
          restartPolicy: OnFailure
```

I guess volcano scheduler maybe just consider node's resource and ignores namespace's resource quota.

Thor-wl: Yeah, so Volcano should consider resource quota. It's an import point indeed.

liuchintao: > Yeah, so Volcano should consider resource quota. It's an import point indeed.

Yeah, from this perspective, this is pr is a workaround to deal with this issue, I find pr https://github.com/volcano-sh/volcano/pull/1100 considers resource quota.

PR #1076, opened 2020-10-07
Remove case sentitive files to avoid collided paths

shinytang6:
My newly cloned project has collided paths (case-sensitive paths on a case-insensitive filesystem)

PR #1078, opened 2020-10-08
Add MaxRequeueNum as a controller-manager param

### LINKED ISSUE #1031: Add maxRetries parameters to controller

shinytang6:
resolves #1031 
Signed-off-by: tangliang <1074461480@qq.com>

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -28,6 +28,7 @@ const (
 	defaultQPS           = 50.0
 	defaultBurst         = 100
 	defaultWorkers       = 3
+	defaultMaxRequeueNum = 15
 	defaultSchedulerName = "volcano"
 )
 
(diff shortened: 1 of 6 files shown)


hzxuzhonghu:

shinytang6:
> It seems you need to rebase and so that the new travis config from #1085 can take effect

Sure, but because of the problem reported here #1076, it is very difficult for me to synchronize upstream with my local master, I have to fork the project again, so please refer to #1087 and merge that one :)

Issue #1014, opened 2020-08-27
delay pod creation feature not consider resourcequota


merryzhou:
/kind feature

As mentioned in [delay pod design](https://github.com/volcano-sh/scheduler/blob/master/doc/design/delay-pod-creation.md),  scheduler should watch resourcequota to decide whether podGroup has enough resource quota to create.
```
Quota
To delay pod creation, both kube-batch and PodGroupMinResources will watch ResourceQuota to decide which PodGroup should be in queue firstly. The decision maybe invalid because of race condition, e.g. other controllers create Pods. In such case, PodGroupMinResources will reject PodGroup creation and keep InQueue state until kube-batch transform it back to Pending. To avoid race condition, it's better to let kube-batch manage Pod number and resources (e.g. CPU, memory) instead of Quota
```
But now in the implementation, the enqueue action controls the quota by queue.Spec.Capability instead of k8s resourcequota, so I am curious about the reason why not use resourcequota to control quota。 Is there any consideration？

**Environment**:
- Volcano Version: v0.4.2

hzxuzhonghu:
The queue is cluster scoped not namespace scoped, and it is a set of podgroups in different namespaces. And different queues can include podgroups of the same namespaces too. 

As the resourcequota, it is the kube-apiserver that use it to intercept the pod creation

k82cn:
> scheduler should watch resourcequota to decide whether podGroup has enough resource quota to create.

+1, it's better to avoid enquing an `PodGrouop` which does not have enough quota :)

Would you like to create a PR for that?

merryzhou:
> Would you like to create a PR for that?

ok，I‘d like to have a try.

k82cn:
Let's keep this issue open :)

PR #1079, opened 2020-10-09
feat(scheduler): support resource quota consideration during pod group enqueue procedure

YuxiJin-tobeyjin:
Try to solve: https://github.com/volcano-sh/volcano/issues/1014

1. Take resourceQuota in the namespace into consideration during podGroup enqueue procedure

For example you can set your resource quota as follows, then your podGroup will stay in pending state if the resource request plus usage is more than quota  

```yaml
spec:
  hard:
    requests.cpu: "32"
    requests.memory: 32Gi
    requests.nvidia.com/gpu: "10"
```

2. Add a flag to control this feature, default is “false“, which means disable

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -57,9 +57,10 @@ type ServerOption struct {
 	HealthzBindAddress string
 
 	// Parameters for scheduling tuning: the number of feasible nodes to find and score
-	MinNodesToFind             int32
-	MinPercentageOfNodesToFind int32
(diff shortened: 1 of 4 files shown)


YuxiJin-tobeyjin:
cc @k82cn

hzxuzhonghu:

Thor-wl:
@YuxiJin-tobeyjin Please fix the DCO and CI failure

YuxiJin-tobeyjin:
/close

Will test locally and submit a new pr based on master.

Issue #1086, opened 2020-10-13
In allocate action, some job which cannot be scheduled will block the scheduling of following jobs


Thor-wl:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug

**What happened**:
Suppose there are 3 jobs(j1, j2, j3) whose podgroup phase is inqueue and they will be scheduled in allocate action. As the current logic, j1/j2/j3 will be put in the priority queue and schedule them in order(suppose the order is j1 > j2 > j3 in priority). First, the request resource of j1 can be satified and bind its tasks to some nodes. Second, try to allocate resource to j2 and predicate it in all nodes. Unfortunately, no node fits its tasks. What we expect is skip j2 and cotinue to deal with j3. But as what i observed is that scheduler will leave allocate action and go to the next action.   
**What you expected to happen**:
If some job cannot be allocated resource, the following jobs wiht lower priority will be dealt with until all jobs in priority queue are dealt with.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version: v1.0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

merryzhou:
hey，can i take this as a try?

Thor-wl:
> hey，can i take this as a try?

Of course!

PR #1089, opened 2020-10-14
fix: not skip lower prioprity job when job with high priority allocated failed

merryzhou:
fix https://github.com/volcano-sh/volcano/issues/1086

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -247,7 +247,6 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 			stmt.Commit()
 		} else {
 			stmt.Discard()
-			break
 		}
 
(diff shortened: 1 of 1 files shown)


Thor-wl (commented):
/lgtm

Thor-wl:
@william-wang This fix should be merge to v1.0&v0.4

william-wang:
The fix might impact current users. For example, there is one high priority big job. 
This kind of job might starve if there is continous small low priority job comming.

Thor-wl:
> The fix might impact current users. For example, there is one high priority big job.
> This kind of job might starve if there is continous small low priority job comming.

IC, but i think it should not be a default managerment process for most users may expect the usage rate and throughput of cluster is as high as possible. That resource reservation requirement is satisfied with specific actions and plugins may be a better solution. https://github.com/volcano-sh/volcano/pull/1044 is working for this.

Issue #1090, opened 2020-10-14
queue weight cannot be 0


Thor-wl:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug

**What happened**:
admission admits queue weight to be 0, but it is not meaningful
**What you expected to happen**:
queue weight must be above 0
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Thor-wl:
related issue:https://github.com/volcano-sh/volcano/issues/1083

PR #1092, opened 2020-10-15
add queue weight validation

Thor-wl:
fix https://github.com/volcano-sh/volcano/issues/1090


Thor-wl:
We should cherry-pick recent bug-fix pr back to former versions including this one. I'll arrange this work next week. @Thor-wl

PR #1100, opened 2020-10-19
feat(sheduler): take resource quota into consideration during pod group enqueue procedure

YuxiJin-tobeyjin:
Try to solve: https://github.com/volcano-sh/volcano/issues/1014

1. Take resourceQuota in the namespace into consideration during podGroup enqueue procedure

For example you can set your resource quota as follows, then your podGroup will stay in pending state if the resource request plus usage is more than quota  

```yaml
spec:
  hard:
    requests.cpu: "32"
    requests.memory: 32Gi
    requests.nvidia.com/gpu: "10"
```

2. Add a flag to control this feature, default is “false“, which means disable


Thor-wl:
@william-wang @k82cn

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -114,7 +118,23 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
...
+				if !reflect.DeepEqual(resource, api.EmptyResource()) {
+					resourceMap[job.Namespace] = resource
+				}
+			}
+			klog.V(3).Infof("Now start to do comparison, for job <%s/%s>, Enqueueable: %t, idle: %t", job.Namespace, job.Name, ssn.JobEnqueueable(job), minReq.LessEqual(idle))
 			if ssn.JobEnqueueable(job) && minReq.LessEqual(idle) {
+				if _, ok := resourceMap[job.Namespace]; ok {
+					klog.V(3).Infof("For job <%s/%s>, ResourceQuota: %t", job.Namespace, job.Name, minReq.LessEqual(resourceMap[job.Namespace]))
+					if minReq.LessEqual(resourceMap[job.Namespace]) {
+						resourceMap[job.Namespace].Sub(minReq)
```
shinytang6, line 132: l think its better to put `idle.Sub(minReq); inqueue = true` here so that the following `else` clause can be removed :)

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -114,7 +118,23 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 			inqueue = true
```
shinytang6, line 118: And l think even though `job.PodGroup.Spec.MinResources == nil`, resourceQuota should also be checked?

shinytang6, line 118: And l think even though job.PodGroup.Spec.MinResources == nil, resourceQuota should also be checked?

YuxiJin-tobeyjin: Considering minResources is nil, the pods of this podgroup won't consume any resoureQuota, so we don't need it to be checked? WDYT

pkg/scheduler/actions/enqueue/enqueue.go
YuxiJin-tobeyjin: Actually `idle.Sub(minReq)；inqueue = true`  doesn't belong to this `if` branch;

If we put it here, `idle.Sub` will not be calculated under circumstances when there is no resourceQuota in the namespace.

Thor-wl:
Please resolve the conflictions

Thor-wl:
Please fix the conflictions and bugs

YuxiJin-tobeyjin:
> Please fix the conflictions and bugs

Okay， would you please give me some clues about bugs， tks

shinytang6:
> > Please fix the conflictions and bugs
> 
> Okay， I will try to handle the conflicts. Would you please give me some clues about bugs，tks.

Can y please resolve the conflicts, l think the original code is almost ready

YuxiJin-tobeyjin:
Since enqueue structure has been changed in master and there are some logic defects in this design, we will give a new proposal with a plugin realization in the near future.

@merryzhou 

This  PR can be closed.

PR #1111, opened 2020-10-26
Some improvements of scheduler

shinytang6:
some minor improvements while scanning the scheduler code
Signed-off-by: tangliang <1074461480@qq.com>

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -283,7 +283,7 @@ func (ni *NodeInfo) RemoveTask(ti *TaskInfo) error {
 			ni.Releasing.Sub(task.Resreq)
 			ni.Idle.Add(task.Resreq)
 			ni.Used.Sub(task.Resreq)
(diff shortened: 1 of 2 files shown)


Thor-wl:
@shinytang6 please resolve the conflictions.

pkg/scheduler/api/node_info.go
```diff
@@ -283,7 +283,7 @@ func (ni *NodeInfo) RemoveTask(ti *TaskInfo) error {
 			ni.Releasing.Sub(task.Resreq)
 			ni.Idle.Add(task.Resreq)
 			ni.Used.Sub(task.Resreq)
-			ni.AddGPUResource(ti.Pod)
+			ni.SubGPUResource(ti.Pod)
```
william-wang, line 286: Why decreace GPU resource?

shinytang6: Because it's removing task from node, how could we add resource here?

PR #1118, opened 2020-10-27
Fix resource Less method

zen-xu:
fix-bug: #1036

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -224,37 +224,38 @@ func (r *Resource) Multi(ratio float64) *Resource {
 
 // Less checks whether a resource is less than other
 func (r *Resource) Less(rr *Resource) bool {
-	lessFunc := func(l, r float64) bool {
-		return l < r
+	greaterFunc := func(l, r float64) bool {
+		return l > r
 	}
 
-	if !lessFunc(r.MilliCPU, rr.MilliCPU) {
-		return false
-	}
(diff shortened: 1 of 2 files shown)


william-wang:
It's just a workaround :) 
Suggest to use Rat (xref: https://golang.org/src/math/big/rat.go ) to improve the accuracy.

@Thor-wl , As far as I know, you have been working on this issue a short time ago.

pkg/scheduler/api/resource_info.go
```diff
@@ -224,39 +224,35 @@ func (r *Resource) Multi(ratio float64) *Resource {
...
+				return false
 			}
+
+			same = same && (rQuant == rrQuant)
 		}
-		return true
 	}
-
-	if rr.ScalarResources == nil {
+	if same {
```
Thor-wl, line 252: How about the following testcase?
`
r = {
cpu: 100,
memory: 0,
scalar_resource:{
   gpu:1
}
}
rr = {
cpu: 100,
memory: 100
}
`

zen-xu: result will be `false`
Why should care about such small resource?

zen-xu: I'm sorry, it's a bug, I'll fix it

Thor-wl: > I'm sorry, it's a bug, I'll fix it

In my opinion, current logic has serious problem for not considering some boundary testcases. Fix on current code is not a good idea.

PR #1131, opened 2020-11-04
Optimize spec def of job

### LINKED ISSUE #1120: Optimize Spec Definitions

> Uncomment only one, leave it on its own line:
>
> /kind bug

Currently all the int parameters in such like `TaskSpec` and `JobSpec` are defined as Interger, they should be optimized to pointers, otherwise it's difficult to distinguish the default value and 0.
for example, the following code may have some problem:
```
// Get the maximum number of retries.
maxRetry := DefaultMaxRetry
if ps.job.Job.Spec.MaxRetry != 0 {
	maxRetry = ps.job.Job.Spec.MaxRetry
}
```

shinytang6:
resolves #1120 

**Note: users may need to update their code before upgrading to the latest version of Volcano if this PR is merged.**


pkg/apis/batch/v1alpha1/job.go
```diff
@@ -52,8 +52,9 @@ type JobSpec struct {
 	SchedulerName string `json:"schedulerName,omitempty" protobuf:"bytes,1,opt,name=schedulerName"`
 
 	// The minimal available pods to run for this Job
+	// Defaults to the summary of tasks' replicas
 	// +optional
-	MinAvailable int32 `json:"minAvailable,omitempty" protobuf:"bytes,2,opt,name=minAvailable"`
+	MinAvailable *int32 `json:"minAvailable,omitempty" protobuf:"bytes,2,opt,name=minAvailable"`
```
k82cn, line 57: will that PR break the backward compatability?

shinytang6: No, users can use it as before

shinytang6:
l think the PR is ready to be reviewed. PTAL:) @k82cn @Thor-wl @william-wang

Thor-wl:
> l think the PR is ready to be reviewed. PTAL:) @k82cn @Thor-wl @william-wang

OK, there seems to be some bug in CI. I'll approve after CI happy

pkg/apis/batch/v1alpha1/job.go
william-wang: Currently there are some customers already call this Job API in their products.  They have to update their code to adapt this changes.

shinytang6: yes, l actually missed that. If user submit the vcjob through code and set/use minAvailable, while not yaml, they need to modify their code indeed.. So what's your opinion about that?

william-wang: We can find another way to resolve the orignal issue and keep the api as it is.

shinytang6: The biggest problem with `int` is that we cannot distinguish between user behavior(set to 0) and the default int value of 0. And I think it can only be solved by setting it as a pointer..

Thor-wl:
please resove the conflictions

Issue #1129, opened 2020-11-03
Is the "volcano-system" namespace supposed to be hardcoded in scheduler.yaml


Eo300:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
When installing the v1.1.0 release via Helm, I got a the following message:
```
Error: namespace "volcano-system" not found
```
I'm attempting to upgrade from v0.3, so I was assessing potential side effects. I noticed the following was introduced between v1.0.1 and v1.1.0 in [scheduler.yaml](https://github.com/volcano-sh/volcano/blob/master/installer/helm/chart/volcano/templates/scheduler.yaml#L131):
```
---
apiVersion: v1
kind: Service
metadata:
  annotations:
    prometheus.io/path: /metrics
    prometheus.io/port: "8080"
    prometheus.io/scrape: "true"
  name: volcano-scheduler-service
  namespace: volcano-system
spec:
  ports:
  - port: 8080
    protocol: TCP
    targetPort: 8080
  selector:
    app: volcano-scheduler
  type: ClusterIP
```
Should `namespace: volcano-system` be `namespace: {{ .Release.Namespace }}` like in the rest of the files?

**What you expected to happen**:
A Helm install deploy without issue.

**How to reproduce it (as minimally and precisely as possible)**:

1. Download release tarball from [v1.1.0 release](https://github.com/volcano-sh/volcano/releases/tag/v1.1.0).
2. Perform a `helm install <RELEASE NAME> <PATH TO CHART>` on the volcano chart

**Anything else we need to know?**:
Specifically, I'm using the install as a dependency, but the issue should manifest regardless.
Additionally, I ran into another similar error, except with the `volcano-monitoring` namespace. Wasn't sure whether it was intentional, so I mainly wished to highlight this one.

**Environment**:
- Volcano Version: **1.1.0**
- Kubernetes version (use `kubectl version`): **1.16.6**

william-wang:
Thanks for your reporting. The namespace format is not right. We will have a investigation as soon as possible.

Thor-wl:
> Additionally, I ran into another similar error, except with the `volcano-monitoring` namespace. Wasn't sure whether it was intentional, so I mainly wished to highlight this one.

@alcorj-mizar Pleae have a look at this issue

alcorj-mizar:
OK, i will fix that.

PR #1132, opened 2020-11-04
Deployment: fix helm upgrade bugs on namespace hardcode

alcorj-mizar:
fix helm upgrade bugs on namespace hardcode


alcorj-mizar:
#1129

william-wang (approved):
/approve

Issue #1098, opened 2020-10-15
when not use  proportion plugin, all job in the queue created first  will be scheduled first


merryzhou:
> /kind feature

**What happened**:

my scenario：

Jobs in different namespaces should be sorted by created time, and jobs in the same namespace should be sorted by priority。

for example:
1. ns A has job a1 which priority  is 2,  job a2 with priority 1,  we expect a1 will be scheduled first； --- volcano can do it 
2. ns A has job a, and ns B has job b, and job a created first, we expect a will be scheduled first;    --- we cannot find a proper config for this scenario

other info:
we bind a queue to each namespace, and unset proportion plugin( https://github.com/volcano-sh/volcano/issues/1083#issuecomment-707443694).

After unset  proportion plugin, all job in the queue created first  will be scheduled first, that means in the above example, all jobs in ns A will be scheduled first。 

**What you expected to happen**:
1. is there any way to achieve example 2
2. is it unreasonable all job in the queue created first  will be scheduled first when unset proportion plugin?

**Environment**:
- Volcano Version: v1.0.1
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Thor-wl:
Accord to the logic of allocate action, the requirement cannot be satisfied currently. I'll take it into consideration of v1.2 feature design. /assign @Thor-wl

shinytang6:
I wonder why the default `QueueOrder` behavior(without plugins) is by CreationTimestamp & UID? I don't think it's meaningful and instead it should be the same as the default `NamespaceOrderFn` behavior, maybe just schedule queue one by one, let jobs dominate the scheduing order?

alcorj-mizar:
> while not having to select a queue first and then decide which job to schedule
There are some misunderstanding, we do not select a queue first.Instead, all jobs are schedule by job's creation timestamp between namespaces and queues will ensure the fair schedule.
Scheduling system is always make a balance between fairness and high efficiency.
The example above is not complex enough, let's think more deeper.
There are 5 jobs to be scheduled by 2 queue(namespace).
1. ns A has job a1 with priority 2, creation timestamp is 8:05AM, job a2 with priority 1 creation timestamp is 8:10AM, job a3 with priority 3 timestamp is 8:30AM.
[...]

PR #1142, opened 2020-11-10
Fair Scheduling For Jobs Of Same Priority And Different Queue

alcorj-mizar:
fair scheduling for jobs of same priority and different queue, add top() to priority queue and add compare job create timestamp towards different queue.


alcorj-mizar:
#1098

william-wang:
The fix broken the existing feature. For this case, I will have a discussion with user today to confirm the scenario firstly.

Thor-wl:
cc @merryzhou

Thor-wl:
If no more discussion about this feature, we settle it as a confliction temporarily and close this PR @merryzhou

Issue #1035, opened 2020-09-07
Customize scheduling algorithms per queue


hzxuzhonghu:
Support customized scheduling algorithms per queue, the main use scenario is: different users may need different algorithms, even in the same clusters.

For example, the `enqueue` is not necessary and may conflict with preempt. So it is better to configure the scheduling according to the user's requirement.

Subham-Pratap:
Hey @hzxuzhonghu .  

I'm interested in working on this project as part of community bridge. I have demonstrated experience working on scheduling algorithms and scheduling in general with projects on the same. I also have decent knowledge of Golang. 

Is there any forum for discussion regarding the project which I can join or a timeline?

Thanks.

Thor-wl:
> Hey @hzxuzhonghu .
> 
> I'm interested in working on this project as part of community bridge. I have demonstrated experience working on scheduling algorithms and scheduling in general with projects on the same. I also have decent knowledge of Golang.
> 
> Is there any forum for discussion regarding the project which I can join or a timeline?
> 
> Thanks.

Warmly Welcome! Volcano community has wechat group and weekly meeting. You can search "k8s2222" for a apply to add friend and ask her to join you into volcano group. Our weekly meeting is every 15:00 PM every Friday. The meeting link can be found in README. Volcano feature design and version arrangement are listed volcano/docs. You can find our current ongoing feature and bug fix in the issue list. We are waiting for your PRs and discussions.

hzxuzhonghu:
@Subham-Pratap Have you applied in the official website?  And this project was selected by sresthasrivastava.

Subham-Pratap:
Yes I have applied on the official website and I'm looking at the issue list in the meantime.

hzxuzhonghu:
@Subham-Pratap Could you select another one?

Subham-Pratap:
Okay do you want me to select another project within volcano or select a different org. I have also applied on the official website, so should I withdraw it?

hzxuzhonghu:
I mean you can select  another project within volcano

sresthas:
@k82cn 
As of now the configuration is loaded into scheduler at the initialisation or when scheduling config is changed,
Now scheduler creates session periodically & goes through each of the actions and executes them. 
This works assuming same config for all queues, hence once actions are loaded, they can be directly traversed in every period. In this design, in one session, jobs corresponding to different queues can be scheduled/traversed together.

For our use case, 
[...]

PR #1143, opened 2020-11-11
design doc for custom scheduling algorithms per queue

sresthas:
https://github.com/volcano-sh/volcano/issues/1035


Thor-wl:
OK, since CI has some problems, i'm trying to fix it.

Thor-wl:
cc @william-wang @Thor-wl @shinytang6 @alcorj-mizar

docs/design/schedule-customisation-design.md
```diff
@@ -0,0 +1,29 @@
...
+
+![Feature Design](./images/reservation_design.png)
+
+## Implementation
+1. Update the CRD structure to take scheduling configmap key in QueueSpec
+		The optional field which will take configmap selector
+		Default configmap is passed as a command-line argument to the scheduler it should be considered for a Queue in which no scheduling configmap is specified.
+2. Load and store different configurations in the scheduler.
+3. Store mapping of configuration to list of Queues in the scheduler. When a new Queue is created, its controller should check if a config map is specified with it. The controller should signal the scheduler to update its mapping of config to a list of queues to include a new queue in the corresponding list. 
+4. Traverse through each configuration periodically and create separate scheduling sessions for them.
```
shinytang6, line 23: Are there any restrictions for doing it in one session?

shinytang6: IMO, if we start different sessions for different configured queues, it is difficult to expand, what if we want to custom scheduling algorithms for per ns or maybe other custom requirements?

Thor-wl: > IMO, if we start different sessions for different configured queues, it is difficult to expand, what if we want to custom scheduling algorithms for per ns or maybe have other customed requirements later?

I think custom scheduling algorithms per ns is not so reasonable, but reasonable per queue.

Thor-wl: The biggest difficulity is to satisfy the requirement without changing existing use experience. It must keep backward capability.

sresthas: @shinytang6 
> Are there any restrictions for doing it in one session?

So a session is traversal of actions, now which actions are invoked depends on configuration right?
Different scheduling configurations will have different actions. So a session should just follow actions & plugins from one configuration. Once a config is loaded and session is started, it shouldn't change config until that session is closed. Hence we need separate sessions for separate configurations, right?

sresthas: There are two options :

1) We follow default config ( loaded through command-line ) in all cases except for custom queues. For custom queues, we could have config within yaml. So we load config and create a separate session for them. So in every iteration, we create one session for all the queues in default bucket, and a separate session per custom queue. 
2) We have a mapping of configuration to list of queues. By default, all queues will be mapped to default config ( loaded through command-line ). Now for custom queues, each custom queue will map to a config. Now in every iteration, scheduler will create a session for each configuration, it will handle all queues belonging to a configuration together.

Suppose we have a case where we have 200 queues. 100 of them follow default config. So there are 100 custom queues. Now in these 100 custom queues, if we have only 4 unique configurations ( say 25 queues per config ). Then
In design 1) we will have one session for default queues and 100 separate sessions for each custom queue. 
But in design 2) we will have only 5 sessions ( 1 for default config with 100, and 4 for custom config with 25 queues each ).

docs/design/schedule-customisation-design.md
```diff
@@ -0,0 +1,29 @@
+# Customize Scheduling Algorithms per Queue
```
william-wang, line 1: please add the interaction with existing features.

sresthas: Updated [Link](https://github.com/volcano-sh/volcano/pull/1143/files#diff-9d6403c433ff965dac666d298542c7a98df50791bd61ba2ab3340ab75ad49db5R27)

sresthas:
@Thor-wl @william-wang  I have updated design doc. The proposed feature is to allow customisation of plugins per Queue. However actions will remain same throughout the cluster.

Thor-wl:
Have you made an appointment with @william-wang to discuss the detail? @sresthas

sresthas:
> Have you made an appointment with @william-wang to discuss the detail? @sresthas

@william-wang  Please let me know when/where we can discuss.

wpeng102:
@sresthas All the queues will use global `action` configure and each queue will use its self `plugin` if customer specify `plugin` configure in queue spec, right?  

Could you help explain the following detail design?
1) At the beginning of `action`, it will use `NamespaceOrderFn`, `QueueOrderFn` which defined in each plugin to sort and pick up a namespace or a queue. So, before pickup a queue, you will use which plugin to load  `NamespaceOrderFn`, `QueueOrderFn` ?
2) Do we support hot update for the plugin configure in queue spec? If yes, we need consider the running workload impact.
3) For the `reclaim` action, configure different plugin for different queue, is it possible to get different `deserved` , `share` for the same queue? If yes, this is maybe lead resource ping-pong for `reclaim` action.

docs/design/queue-customizable-plugins.md
```diff
@@ -0,0 +1,35 @@
...
+Actions will remain the same as the default configuration, but the user can customize different plugins for different queues. 
+* Currently, the scheduling configuration is loaded initially, and it's only updated when the underlying scheduling config is changed. 
+In a scheduling session, the scheduler traverses through actions ( which were loaded from the scheduling config in the beginning ). 
+In each action, it goes through all jobs ( from all Queues ) that are yet to be scheduled and perform actions. 
+The proposed feature won't affect this workflow. However, which plugins are executed for jobs belonging to a particular Queue will depend on whether any plugins are specified in the Queue YAML or not.
+* If plugins are specified with a Queue, they will be loaded and invoked. Otherwise, default plugins specified in the config-map will be used.
+
+## Implementation
+
+1. Introduce an optional field to specify plugins within Queue CRD spec
```
william-wang, line 27: please describe the specific API change in design doc :）

k82cn: is there any update on config file? it's better to define some "type" of scheduling plugins, and refer one of them in Queue spec.

sresthas: We don't need to update the existing config workflow. 
Since we already have `Tier` & `PluginOption` defined. 
We could add the following field to Queue Spec 
```
// optional field to define plugins
Tiers []Tier `yaml:"tiers"`
```

We could also have a `SchedulerConfiguration` instance instead of `Tier`. But we don't need `Actions`, just the plugins.

sresthas:
> @sresthas All the queues will use global `action` configure and each queue will use its self `plugin` if customer specify `plugin` configure in queue spec, right?

Correct

> 
> Could you help explain the following detail design?
> 
> 1. At the beginning of `action`, it will use `NamespaceOrderFn`, `QueueOrderFn` which defined in each plugin to sort and pick up a namespace or a queue. So, before pickup a queue, you will use which plugin to load  `NamespaceOrderFn`, `QueueOrderFn` ?

Since each Queue belongs to one namespace or another, nothing needs to be changed right? At the beginning of `action`, it will pick up a namespace, then pick up a queue, if the queue has plugins defined in the spec, it will use them, instead of global ones.

> 2. Do we support hot update for the plugin configure in queue spec? If yes, we need consider the running workload impact.

We don't need to support hot update, at least in the beginning. 
In the existing workflow, if the scheduling config is changed, then it gets updated in the next cycle. If we support hot update, it should be similar to this. So to avoid any conflicts, we could load a mapping of Queues to plugins at the beginning of each cycle. This way, even if plugins get updated within a cycle, it won't affect the workflow. 

> 3. For the `reclaim` action, configure different plugin for different queue, is it possible to get different `deserved` , `share` for the same queue? If yes, this is maybe lead resource ping-pong for `reclaim` action.

Not sure about this. Doesn't make sense for `reclaim` action to use different plugins for different queues.

Issue #1094, opened 2020-10-15
need more info when job pending


merryzhou:
> /kind feature

**What happened**:

when pod of volcano job pending, there is no actual reason in events/status of job、podgroup and pod.

```
kubectl describe vcjob test
...
Status:
  Min Available:  1
  Pending:        1
  State:
    Last Transition Time:  2020-10-14T02:17:58Z
    Phase:                 Pending
Events:                    <none>

kubectl describe pg test
...
Status:
  Conditions:
    Last Transition Time:  2020-10-15T09:01:46Z
    Message:               1/1 tasks in gang unschedulable: pod group is not ready, 1 Pending, 1 minAvailable.
    Reason:                NotEnoughResources
    Status:                True
    Transition ID:         a6e36d8a-1672-406e-a9f5-f113174114fe
    Type:                  Unschedulable
  Phase:                   Inqueue
Events:
  Type     Reason         Age                    From     Message
  ----     ------         ----                   ----     -------
  Warning  Unschedulable  69s (x78789 over 26h)  volcano  1/1 tasks in gang unschedulable: pod group is not ready, 1 Pending, 1 minAvailable.

kubectl describe pod test-xxx
Events:          <none>
```

expect more detailed info in events/status, for example: xx resources not enough（expect：5， idle：3）

**Anything else we need to know?**:

**Environment**:
- Volcano Version: v1.0.1
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Thor-wl:
I'm ok about the feature. Let me take it into consideration in v1.2 design. /assing @Thor-wl

mikechengwei:
I can do this pr and assign it to me, ok?

Thor-wl:
> I can do this pr and assign it to me, ok? @Thor-wl

Of course! Thanks for your contribution! I'll update the roadmap and assgin this feature to you. According to plan, v1.2 will be released at the middle of 1st 2021, so please note it.

Thor-wl:
@mikechengwei roadmap updated done. https://github.com/volcano-sh/volcano/pull/1133

PR #1148, opened 2020-11-15
job event optimize

mikechengwei:
### What problem does this PR solve? 
#1094 

When podgroup recorder event,job also recorder.
### What is changed and how does it work?
When podgroup recorder event,job also recorder.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -49,7 +49,9 @@ import (
 	vcv1beta1 "volcano.sh/volcano/pkg/apis/scheduling/v1beta1"
 	vcclient "volcano.sh/volcano/pkg/client/clientset/versioned"
 	"volcano.sh/volcano/pkg/client/clientset/versioned/scheme"
+	informerfactory "volcano.sh/volcano/pkg/client/informers/externalversions"
 	vcinformer "volcano.sh/volcano/pkg/client/informers/externalversions"
+	v1alpha12 "volcano.sh/volcano/pkg/client/informers/externalversions/batch/v1alpha1"
 	vcinformerv1 "volcano.sh/volcano/pkg/client/informers/externalversions/scheduling/v1beta1"
 	schedulingapi "volcano.sh/volcano/pkg/scheduler/api"
 )
@@ -88,6 +90,7 @@ type SchedulerCache struct {
 	pcInformer              schedv1.PriorityClassInformer
 	quotaInformer           infov1.ResourceQuotaInformer
 	csiNodeInformer         storagev1.CSINodeInformer
+	jobInformer             v1alpha12.JobInformer
 
 	Binder        Binder
 	Evictor       Evictor
@@ -401,6 +404,8 @@ func newSchedulerCache(config *rest.Config, schedulerName string, defaultQueue s
 		DeleteFunc: sc.DeleteQueueV1beta1,
 	})
 
+	sc.jobInformer = informerfactory.NewSharedInformerFactory(sc.vcClient, 0).Batch().V1alpha1().Jobs()
+
 	return sc
 }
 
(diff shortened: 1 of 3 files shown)


Thor-wl:
Try to resubmit please. There seems some e2e timeout

pkg/scheduler/cache/cache.go
```diff
@@ -851,4 +858,17 @@ func (sc *SchedulerCache) recordPodGroupEvent(podGroup *schedulingapi.PodGroup,
...
+	if err != nil {
+		klog.Errorf("Failed to get VolcanoJob  <%s/%s>: %v",
+			podGroup.Namespace, podGroup.Name, err)
+	}
+	if batchJob == nil {
+		klog.Infof("VolcanoJob <%s/%s> not found",
+			podGroup.Namespace, podGroup.Name)
+	} else {
+		sc.Recorder.Eventf(batchJob, eventType, reason, msg)
+	}
```
Thor-wl, line 873: Adding job event logic into `recordPodGroupEvent` funciton seems a little strange. Can you describe the scenario or consideration?

k82cn: > Adding job event logic into recordPodGroupEvent funciton seems a little strange

Volcano scheduler should not aware any kind of k8s, it only does scheduling based on PodGroup.

mikechengwei: When podgroup can't schedule due to resource limit ,we also need to sync event info to volcano job which help users troubleshoot problems. What do you think?

Thor-wl: > When podgroup can't schedule due to resource limit ,we also need to sync event info to volcano job which help users troubleshoot problems. What do you think?

Sound reasonable in some degree. It' a feaure in v1.2 which has been proposed by @merryzhou at https://github.com/volcano-sh/volcano/issues/1094 @k82cn

pkg/scheduler/cache/cache.go
```diff
@@ -851,4 +858,17 @@ func (sc *SchedulerCache) recordPodGroupEvent(podGroup *schedulingapi.PodGroup,
 		return
 	}
 	sc.Recorder.Eventf(pg, eventType, reason, msg)
+
+	jobLister := sc.jobInformer.Lister()
+	batchJob, err := jobLister.Jobs(podGroup.Namespace).Get(podGroup.Name)
```
k82cn, line 863: we should not bind podgroup to job

pkg/scheduler/cache/cache.go
k82cn: we should not bind podgroup to job; only pod & podgroup should be used in scheduler.

pkg/scheduler/cache/cache.go
mikechengwei: So if the podgroup has an error, should the error information be synchronized to the job?

mikechengwei: User current needs to describe resources one by one

k82cn: The job controller/operator should monitor PodGroup's event.

mikechengwei: So , i should change to monitor podgroup event in controller ,right?

william-wang: Yes, The podgroup is created by controller. The controller monitor podgroup event seems reasonable. 
Job should not be involved in scheduler because we need to keep compatible with variable operator without using volcano job.

mikechengwei: I have move code to job controller. PLTA

mikechengwei:
@k82cn

k82cn:
LGTM overall :)

/approve

PR #1149, opened 2020-11-15
optimize local script

mikechengwei:
### What is changed and how does it work?
Optimze local script,the latest build image will be reinstalled every time.


hack/local-up-volcano.sh
```diff
@@ -97,6 +98,21 @@ if [ "${INSTALL_MODE}" == "kind" ]; then
...
+)
+for n in ${images[@]}; do
+    echo "info: loading image $n"
+    kind load docker-image --name $CLUSTER_NAME $n
+done
+
+kubectl delete deploy volcano-admission -n $NAMESPACE
+kubectl delete deploy volcano-controllers  -n $NAMESPACE
+kubectl delete deploy volcano-scheduler   -n $NAMESPACE
+
```
Thor-wl, line 115: Are these changes working for environments existing Volcano installed and update it?

mikechengwei: Yes,it work in when code change and update image.

Thor-wl: > Yes,it works when we change code and need to update image.

Have you tested it at local environment?

mikechengwei: Sure, when we have code change ,this pr will be very convenient.

mikechengwei: Sorry,there are still some problems with the submission before, I have optimized it and perfected it.

mikechengwei: https://github.com/kubernetes-sigs/cluster-api/issues/2120  kind kubeconfig-path have not supported.

mikechengwei:
@Thor-wl What do you think now ?

mikechengwei:
@Thor-wl PLTA

hack/local-up-volcano.sh
shinytang6: l don't think it's a good idea to hard code `volcano` here, due to the reason that we may deploy volcano using helm with some other prefix.
In addition, if we set `imagepullpolicy` to `Always`, the latest build image will be pulled every time you run this script.

hack/local-up-volcano.sh
```diff
@@ -94,9 +95,13 @@ prepare
 
 if [ "${INSTALL_MODE}" == "kind" ]; then
   kind-up-cluster
-  export KUBECONFIG="$(kind get kubeconfig-path ${CLUSTER_CONTEXT})"
+  export KUBECONFIG="$(kind get kubeconfig ${CLUSTER_CONTEXT})"
 fi
 
+kubectl delete deploy volcano-admission -n $NAMESPACE
```
shinytang6, line 101: l don't think it's a good idea to hard code volcano here, due to the reason that we may deploy volcano using helm with some other prefix.

In addition, if we set `ImagePullPolicy` to `Always`, the latest build image will be pulled every time you run this script.

Thor-wl: We've set daily job to build the latest image and push to dockerhub, so that `ImagePullPolicy` set to `Always` can ensure you get the latest version

mikechengwei: No, i think local-up-volcano.sh script just use for develop.
 I just want to execute this script every time I finish developing the code, it will automatically update the mirror I just developed for debugging.

mikechengwei: And helm can’t know if I have updated, because the name of the image created is the same every time

Thor-wl: Sorry for it cannot be satified now. You can build image locally and load it to your develop environment. Replace the deployment yaml `image` field to debug.

shinytang6: FYI: generally my way to do that is to deploy all the component with `ImagePullPolicy: Always` and delete all the **related pods**, then all the component will be updated.

shinytang6: Even if you want to debug using this script, is `uninstall-volcano` before `install-volcano` work?

mikechengwei: I’m ok.

PR #1157, opened 2020-11-18
make the code easy to understand

divinerapier:
Hi, I find the code is difficult to understand. So I make the pull request.

Thanks.

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -94,7 +94,7 @@ func NewResource(rl v1.ResourceList) *Resource {
 
 // IsEmpty returns bool after checking any of resource is less than min possible value
 func (r *Resource) IsEmpty() bool {
-	if !(r.MilliCPU < minMilliCPU && r.Memory < minMemory) {
+	if r.MilliCPU >= minMilliCPU || r.Memory >= minMemory {
 		return false
 	}
 
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -94,7 +94,12 @@ func NewResource(rl v1.ResourceList) *Resource {
 
 // IsEmpty returns bool after checking any of resource is less than min possible value
 func (r *Resource) IsEmpty() bool {
-	if r.MilliCPU >= minMilliCPU || r.Memory >= minMemory {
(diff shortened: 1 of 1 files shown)


Thor-wl:
Welcome! If find any other code hard to understand, don't hesitate to optimize or make comments.

pkg/scheduler/api/resource_info.go
```diff
@@ -94,7 +94,7 @@ func NewResource(rl v1.ResourceList) *Resource {
 
 // IsEmpty returns bool after checking any of resource is less than min possible value
 func (r *Resource) IsEmpty() bool {
-	if !(r.MilliCPU < minMilliCPU && r.Memory < minMemory) {
+	if r.MilliCPU >= minMilliCPU || r.Memory >= minMemory {
 		return false
```
william-wang, line 98: The minMilliCPU and mini minMemory is float64 type. Do not use equal to compare.

divinerapier: Thanks for your advice. I have two ideas to deal with the problem. PTAL @william-wang , thanks a lot.

The first:
``` go
	// Do not use equal to compare floats.
	const precision = 1e-6
	if r.MilliCPU > (minMilliCPU-precision) || r.Memory > (minMemory-precision) {
		return false
	}

	for _, rQuant := range r.ScalarResources {
		if rQuant >= minMilliScalarResources {
			return false
		}
	}

	return true
}
```
The second:
``` go
	// Do not use equal to compare floats. Compare if a is greater than or equal to b
	geFn := func(a, b float64) bool {
		const precision = 1e-6
		return a > (b - precision)
	}
	if geFn(r.MilliCPU, minMilliCPU) || geFn(r.Memory, minMemory) {
		return false
	}

	for _, rQuant := range r.ScalarResources {
		if rQuant >= minMilliScalarResources {
			return false
		}
	}

	return true
}
```

k82cn:
when will we replace those helper func with Rat :)

divinerapier:
@william-wang  Thanks for your advice. I updated the code, please take a look, thanks a lot.

william-wang:
> when will we replace those helper func with Rat :)

@Thor-wl is working on the helper replacement with RAT. It will be ready in 1.1.1 version on January 2021.

Thor-wl:
Accuracy development is on the way.

PR #1161, opened 2020-11-23
feat: add task minAvailable design

shinytang6:

### DIFF
--- a/docs/design/task-minavailable.md
+++ b/docs/design/task-minavailable.md
@@ -45,7 +45,7 @@ So if we want to support minAvailable for task level, I think the changes involv
    }
    ```
 3. Add a new judgment logic to the current gang scheduling. The current logic is if sumof(valid tasks) >= job.minavailable, we take this job as valid(can be scheduled). We need to add a logic before that. The logic is that if the `minAvailable` field of the task is set, the task under current job must meet the conditions of (valid pod of the task) >= job.task.minAvailable, then we can take the job as valid.
-4. Modify the judgment of job status. The current logic is that if the (successful pod of every task) >= job.task.minAvailable, then we can take the status of the job as `Completed`. This change may need another field in `JobStatus` to record the status of the pod corresponding to the task, currently `JobStatus` only records the number of pods in different states which we cannot distinguish which task belongs to.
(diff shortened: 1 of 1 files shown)


shinytang6:
#1131 (the first step in my design) will cause some compatibility issues indeed, but l think it makes sense in some way.
WDYT? @Thor-wl @william-wang

shinytang6:
@Thor-wl @k82cn @william-wang PTAL.

docs/design/task-minavailable.md
```diff
@@ -0,0 +1,68 @@
...
+        ...
+       
+        // MinTaskMember defines the minimal number of pods to run each task in the pod group;
+        // if there's not enough resources to start each task, the scheduler
+        // will not start anyone.
+        MinTaskMember map[string]int32
+   }
+   ```
+3. Add a new judgment logic to the current gang scheduling. The current logic is if sumof(valid tasks) >= job.minavailable, we take this job as valid(can be scheduled). We need to add a logic before that. The logic is that if the `minAvailable` field of the task is set, the task under current job must meet the conditions of (valid pod of the task) >= job.task.minAvailable, then we can take the job as valid.
+4. Modify the judgment of job status. The current logic is that if the sumof(successful pods) >= job.minAvailable, then we can take the status of the job as `Completed`. This change may need another field in `JobStatus` to record the status of the pods under each task, currently `JobStatus` only records the number of pods in different states which we cannot distinguish which task pod belongs to.
```
Thor-wl, line 48: an optional point: adjust CLI to the feature and support getting pods statistics of task level.

Thor-wl:
Has this feature been implemented? @shinytang6

shinytang6:
> Has this feature been implemented? @shinytang6

[PR 1179](https://github.com/volcano-sh/volcano/pull/1193) implemented this feature, but it's still under test, not fully ready.

Thor-wl:
> > Has this feature been implemented? @shinytang6
> 
> [PR 1179](https://github.com/volcano-sh/volcano/pull/1193) implemented this feature, but it's still under test, not fully ready.

OK, please finish it as soon as possible for it is an important feature in v1.2

Thor-wl:
This feature is very important and we plan to merge it in v1.3.0

shinytang6:
Please also help review that
/cc @zen-xu

zen-xu (approved):

docs/design/task-minavailable.md
Thor-wl, line 48: Another concern: we should consider the releationship between job.minAvailable and the sum of minAvailable at task levels. If job.minAvailable is less, the judgement can be ignored.

Thor-wl (approved):

PR #1164, opened 2020-11-27
fix openSession delete invalid job from session

huone1:
Now, calling  jobValid function before plugin installed, so, it's invalid; the code should execute after plugin installed

create function RemoveInvalidJob to remove invalid job from session, it  works after   plugin installed


huone1:
this modification  leads that pod's gang event doesn't happen

PR #1167, opened 2020-11-30
Bug/fix scheduling duration update after complete

alcorj-mizar:


hack/local-up-volcano.sh
```diff
@@ -44,6 +45,7 @@ function install-volcano {
 }
 
 function uninstall-volcano {
+  kubectl delete job volcano-admission-init -n $VOLCANO_NAMESPACE
```
Thor-wl, line 48: Is this action aimed to prevent that reinstall and bring confliction?

alcorj-mizar: yes, always got this problem when reinstall.

Thor-wl:
@alcorj-mizar Please resolve the confliction

PR #1170, opened 2020-12-02
feat: mutate default of job spec

shinytang6:
Since #1131 needed to modify the API level, some compromises were made. The disadvantage of this is that if the user set the specified field as 0, we will still set the default value, but this will at least not affect the user’s original behavior.


shinytang6:
@william-wang PTAL:)

Issue #1169, opened 2020-12-02
vcjob phase incorrect when job scale up


merryzhou:
> /kind bug

**How to reproduce it (as minimally and precisely as possible)**:
1. create a vcjob: one task, replica:2
2. wait job phase turn to running
3. update job, set replica to 0
4. we found:  vcjob status is running, and pods status are pending

```
status:
  controlledResources:
    plugin-env: env
    plugin-ssh: ssh
    plugin-svc: svc
  minAvailable: 2
  pending: 2
  retryCount: 1
  state:
    lastTransitionTime: "2020-12-02T06:55:31Z"
    phase: Running
```

**Environment**:
- Volcano Version: v1.0.1

Thor-wl:
The logic about job run status execute should be reconsidered, especially in 0 replicas condition.

PR #1174, opened 2020-12-03
fix: update job phase when scale up from jobReplicas==0

merryzhou:
Signed-off-by: merryzhou <merryzhou0882@gmail.com>

https://github.com/volcano-sh/volcano/issues/1169


hzxuzhonghu (commented):
https://github.com/volcano-sh/volcano/blob/42b2cf48a9f686d935a55a34d4135cf04ab8bfc2/pkg/controllers/job/job_controller_actions.go#L207-L216

Take care of this logic to prevent initializing twice

merryzhou:
> https://github.com/volcano-sh/volcano/blob/42b2cf48a9f686d935a55a34d4135cf04ab8bfc2/pkg/controllers/job/job_controller_actions.go#L207-L216
> 
> Take care of this logic to prevent initializing twice

It seems impossible to distinguish the reason for update here...

Could you give me some points?

wpeng102:
just use `job.Status.State.Phase` to check job if initiated maybe not accuracy any more. In preemption case also may lead job status change from running to pending.
```
func isInitiated(job *batch.Job) bool {
	if job.Status.State.Phase == "" || job.Status.State.Phase == batch.Pending {
		return false
	}

	return true
}
```

wpeng102:
Maybe we need introduce other job level flag to mark if job has initiated

k82cn:
> Maybe we need introduce other job level flag to mark if job has initiated

A new phase/state maybe better :)

merryzhou:
> > Maybe we need introduce other job level flag to mark if job has initiated
> 
> A new phase/state maybe better :)

how about restarting?

PR #1179, opened 2020-12-07
generate v1 crds. remove subresource status in pg crd.

stpabhi:
Signed-off-by: Abhilash Pallerlamudi <stp.abhi@gmail.com>

This PR upgrades crds to v1. Also, it fixes an issue where kubebuilder:subresources:status causes PodGroup Status not being updated.

Notes for reviewer: This PR also removes the metadata description from the resource types. Otherwise, it causes a validation error when applying crds. There's an open PR [here](https://github.com/kubernetes-sigs/controller-tools/pull/511) in controller-tools to remove the description from metadata field in generated crds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,7 @@ BIN_DIR=_output/bin
 RELEASE_DIR=_output/release
 REPO_PATH=volcano.sh/volcano
 IMAGE_PREFIX=volcanosh/vc
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+CRD_OPTIONS ?= "crd:crdVersions=v1"
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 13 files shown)


k82cn:
will this PR introduce backward compatability issue?

stpabhi:
> will this PR introduce backward compatability issue?

Did you mean the k8s crd backward compatibility? The apiextensions.k8s.io/v1 API is available in v1.16 which means the crds won't work with k8s api-server  < 1.16. We still have the v1beta1 crds included in volcano-deployment.yaml. If needed we could also keep generating v1beta1 in a different directory. Let me know.

k82cn:
> > will this PR introduce backward compatability issue?
> 
> Did you mean the k8s crd backward compatibility? The apiextensions.k8s.io/v1 API is available in v1.16 which means the crds won't work with k8s api-server < 1.16. We still have the v1beta1 crds included in volcano-deployment.yaml. If needed we could also keep generating v1beta1 in a different directory. Let me know.

Yes, it's better to keep both, and set v1 as default and make v1beta1 as deprecated option.

stpabhi:
Done. Added a note in the readme as well.

william-wang (commented):
/lgtm
/approve

Issue #1177, opened 2020-12-04
volcano job dose not delete network policy when job finished


wpeng102:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:

 /kind bug

**What happened**:
volcano job dose not delete network policy when job finished

**What you expected to happen**:
when job finished, the NetworkPolicy should be deleted just like svc and configmap created by job plugin.

**How to reproduce it (as minimally and precisely as possible)**:
1) submit a volcano job
2) when job running, check the job's NetworkPolicy
3) when job finished, this NetworkPolicy still left.

```
[root@wangpeng yaml]# kubectl get po
NAME           READY   STATUS      RESTARTS   AGE
highjob-t1-0   0/1     Completed   0          32s
highjob-t1-1   0/1     Completed   0          32s
highjob-t1-2   0/1     Completed   0          32s
highjob-t1-3   0/1     Completed   0          32s
highjob-t1-4   0/1     Completed   0          32s
highjob-t1-5   0/1     Completed   0          32s
[root@wangpeng yaml]# kubectl get NetworkPolicy
NAME      POD-SELECTOR                                                   AGE
highjob   volcano.sh/job-name=highjob,volcano.sh/job-namespace=default   35s

```

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #1186, opened 2020-12-09
fix dose not delete network policy when job finished

wpeng102:
Fix issue https://github.com/volcano-sh/volcano/issues/1177

### DIFF
--- a/pkg/controllers/job/plugins/svc/svc.go
+++ b/pkg/controllers/job/plugins/svc/svc.go
@@ -168,6 +168,14 @@ func (sp *servicePlugin) OnJobDelete(job *batch.Job) error {
 	}
 	delete(job.Status.ControlledResources, "plugin-"+sp.Name())
 
+	if !sp.disableNetworkPolicy {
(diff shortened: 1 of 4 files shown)


wpeng102:
SVC job plugin will create network policy automatically but miss to delete it when job finished.

pkg/controllers/job/plugins/svc/svc.go
```diff
@@ -168,6 +168,14 @@ func (sp *servicePlugin) OnJobDelete(job *batch.Job) error {
 	}
 	delete(job.Status.ControlledResources, "plugin-"+sp.Name())
 
+	if !sp.disableNetworkPolicy {
+		if err := sp.Clientset.KubeClients.NetworkingV1().NetworkPolicies(job.Namespace).Delete(context.TODO(), job.Name, metav1.DeleteOptions{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				klog.Errorf("Failed to delete Network policy of Job %v/%v: %v", job.Namespace, job.Name, err)
+				return err
```
k82cn, line 175: what will happen if failed to delete network policy?

wpeng102: network policy will left and controller print error log.  No impact to job finish process.
When job deleted, this network policy also be deleted by GC.

PR #1207, opened 2020-12-20
feat: schedule jobs only on certain nodes

### LINKED ISSUE #1188: Schedule jobs only on certain nodes
/kind feature

**What happened**:
We'd like to schedule jobs only on certain nodes

We would expect to use node selectors to be able to do this through volcano.

ChAnYaNG97:
Signed-off-by: ChAnYaNG97 790194334@qq.com
#1188


ChAnYaNG97:
Something went wrong, I will close this PR.

pkg/scheduler/cache/util.go
```diff
@@ -18,9 +18,22 @@ package cache
...
+	"strings"
 )
 
 // responsibleForPod returns true if the pod has asked to be scheduled by the given scheduler.
 func responsibleForPod(pod *v1.Pod, schedulerName string) bool {
 	return schedulerName == pod.Spec.SchedulerName
 }
+
+// convertNodeSelector converts node selector from string to map
+func convertNodeSelector(selector string) map[string]string {
```
k82cn, line 30: it's better to use util/pkg from upstream on label selector :)

k82cn:
LGTM overall, just some minor comments :)

Thor-wl:
It seems confliction occured. Please fix it

pkg/scheduler/cache/util.go
```diff
@@ -18,9 +18,26 @@ package cache
...
 // responsibleForPod returns true if the pod has asked to be scheduled by the given scheduler.
 func responsibleForPod(pod *v1.Pod, schedulerName string) bool {
 	return schedulerName == pod.Spec.SchedulerName
 }
+
+// convertNodeSelector converts node selector from string to *LabelSelector
+func convertNodeSelector(selector string) metav1.LabelSelector {
+	matchLabels := make(map[string]string)
+	if selector != "" {
+		s := strings.Split(selector, ":")
```
Thor-wl, line 34: IC, just support one label. Seems not so flexible.

pkg/scheduler/cache/cache.go
```diff
@@ -63,8 +63,8 @@ func init() {
 }
 
 // New returns a Cache implementation.
-func New(config *rest.Config, schedulerName string, defaultQueue string) Cache {
-	return newSchedulerCache(config, schedulerName, defaultQueue)
+func New(config *rest.Config, schedulerName string, defaultQueue string, nodeSelector string) Cache {
+	return newSchedulerCache(config, schedulerName, defaultQueue, nodeSelector)
```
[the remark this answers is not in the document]
ChAnYaNG97: Got it, this is a big problem. Do you mean that it's more reasonable to add `nodeSelector` in QueueSpec, and node selectors are specified by user in creating queue yaml, and specify the `queue` in Job and PodGroup yaml. 

But in this way, I am not sure where to filter nodes in code, maybe in actions?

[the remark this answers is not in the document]
Thor-wl: I've no idea about requirement from real scenarios, just private thoughts. In my view, if it works in global level, some nodes must be idle and it's a huge waste. Perhaps user just want this feature for its own scenarios or organzations.

pkg/scheduler/cache/util.go
Thor-wl: Do you have some idea about how to support multiple labels?

ChAnYaNG97: 1. If add `nodeSelector` as a scheduler parameter, maybe the form "<key1>:<value1>,<key2>:<value2>,..." works.
2. If in queue level, maybe just like how pod selects nodes, add a `nodeSelector` map to QueueSpec.

PR #1213, opened 2020-12-21
reconstruct Less func

Thor-wl:
fix https://github.com/volcano-sh/volcano/issues/1036


william-wang:
/ hold

Issue #1215, opened 2020-12-22
Support customer implementing plugins, and the scheduler loads these plugins dynamically


zen-xu:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind feature

**What happened**:

Until now, plugins like `binpack`, `drf`, `gang` are provided by official. 
But some users may want to implement the plugin by themself. So if scheduler can dynamicly load plugins, that will make it more flexible when handling different business scenarios.

**What you expected to happen**:

1. Implement  the plugin

```go
// magic.go

package main  // package must be main

import (
    "volcano.sh/volcano/pkg/scheduler/framework"
)

type magicPlugin struct {}

func (mp *magicPlugin) Name() string {
    return "magic"
}

func New(arguments framework.Arguments) framework.Plugin {  // `New` is PluginBuilder
    return &magicPlugin{}
}

func (mp *magicPlugin) OnSessionOpen(ssn *framework.Session) {}

func (mp *magicPlugin) OnSessionClose(ssn *framework.Session) {}
```

2. Build the plugin to .so
```bash
# Example
go build -buildmode=plugin -o plugins/magic.so magic.go
```

3. Since each plugin builder has the same type `framework.PluginBuilder`,  then we can use [plugin](https://golang.org/pkg/plugin/) package to dynamicly load these .so plugins.

k82cn:
That's interesting feature, I like that so much!!!!
@zen-xu would you help on this feature?

zen-xu:
@k82cn 
No problem

k82cn:
> /assign @zen-xu

Thanks very much :) 

BTW, AFAIK, Golang does not support `.so` well, so please also add a feature gate if someone want to disable it :)

zen-xu:
I will add a cli option `--plugins-dir` to vc-scheduler, so if no custom plugin in this dir, scheduler logic will be consistent with the present

Thor-wl:
related PR: 
Doc https://github.com/volcano-sh/volcano/pull/1219
Implementation https://github.com/volcano-sh/volcano/pull/1218

PR #1218, opened 2020-12-23
feat: support vc-scheduler loading custom plugins (#1215)

zen-xu:


docs/design/custom-plugin.md
```diff
@@ -0,0 +1,72 @@
+# Custom Plugin
```
william-wang, line 1: Please submit the design doc in a seperated pull request :)

docs/design/custom-plugin.md
```diff
@@ -0,0 +1,72 @@
...
+```yaml
+...
+    containers:
+    - name: volcano-scheduler
+      image: volcano.sh/vc-scheduler-pluginable:latest
+      args:
+        - --logtostderr
+        - --scheduler-conf=/volcano.scheduler/volcano-scheduler.conf
+        - -v=3
+        - --plugins-dir=plugins  # specify plugins dir path
```
william-wang, line 65: Currently user configure the plugin in configmap. It's better to let user configure the plugin name and directory there.

docs/design/custom-plugin.md
zen-xu: The reason is ?

docs/design/custom-plugin.md
zen-xu: User shoud still add custom plugin name in configmap

```yaml
    - plugins:
      - name: priority
      - name: gang
      - name: conformance
      - name: magic   # custom plugin name
```

docs/design/custom-plugin.md
Thor-wl: > The reason is ?

It's much more friendly for newers to have a better understanding of this feature

zen-xu: I mean why need to seperate feature and doc?

cmd/scheduler/app/options/options.go
```diff
@@ -30,6 +30,7 @@ const (
 	defaultSchedulerPeriod = time.Second
 	defaultQueue           = "default"
 	defaultListenAddress   = ":8080"
+	defaultPluginsDir      = ""
```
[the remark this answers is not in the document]
zen-xu: if `defaultPluginDir` is empty, custom plugin is disabled

cmd/scheduler/app/server.go
```diff
@@ -63,6 +64,13 @@ func Run(opt *options.ServerOption) error {
 		return err
 	}
 
+	if opt.PluginsDir != "" {
+		err := framework.LoadCustomPlugins(opt.PluginsDir)
+		if err != nil {
+			return err
+		}
+	}
+
```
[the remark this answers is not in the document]
zen-xu: err will be catched by `main` function in `cmd/scheduler/app/main.go`

```go

	if err := app.Run(s); err != nil {
		fmt.Fprintf(os.Stderr, "%v\n", err)
		os.Exit(1)
	}
```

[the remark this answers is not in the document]
Thor-wl: > err will be catched by `main` function in `cmd/scheduler/app/main.go`
> 
> ```go
> 	if err := app.Run(s); err != nil {
> 		fmt.Fprintf(os.Stderr, "%v\n", err)
> 		os.Exit(1)
> 	}
> ```

yeah, i mean add some custom info such as "get plugin path error, err msg: %v\n" instead of output raw os msg. It' more friendly when view log.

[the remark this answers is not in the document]
zen-xu: OK

cmd/scheduler/app/server.go
```diff
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"time"
+	"volcano.sh/volcano/pkg/scheduler/framework"
```
k82cn, line 25: move closer to other volcano pkg import

example/custom-plugin/README.md
```diff
@@ -0,0 +1,22 @@
+# Build plugin
+
+## In docker
+Plugin must use `musl-gcc` to build. The best choose is build it in the docker.
```
k82cn, line 4: hm... IMO, the compiler of plugin and scheduler should be the same; as some users may build scheduler themself.

example/custom-plugin/magic.go
```diff
@@ -0,0 +1,21 @@
...
+
+func (mp *magicPlugin) Name() string {
+	return PluginName
+}
+
+func New(arguments framework.Arguments) framework.Plugin { // `New` is PluginBuilder
+	return &magicPlugin{}
+}
+
+func (mp *magicPlugin) OnSessionOpen(ssn *framework.Session) {}
```
k82cn, line 19: it's better to pring some log here :)

pkg/scheduler/framework/plugins_test.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
k82cn, line 2: s/Kubernetes/Volcano

k82cn:
LGTM overall :)

example/custom-plugin/README.md
zen-xu: because we use alpine as base image，which only has musl-libc

k82cn: yes, that's ok in our case. But if someone want to build scheduler themself, it's better to highlight that; or add add a reference about the limitation of `Plugin`.

Makefile
```diff
@@ -73,11 +74,14 @@ vcctl: init
...
-	CGO_ENABLED=0 $(GOBIN)/gox -osarch=${REL_OSARCH} -ldflags ${LD_FLAGS} -output ${BIN_DIR}/${REL_OSARCH}/vcctl ./cmd/cli
-	for name in controller-manager scheduler webhook-manager; do\
-		CGO_ENABLED=0 $(GOBIN)/gox -osarch=${REL_OSARCH} -ldflags ${LD_FLAGS} -output ${BIN_DIR}/${REL_OSARCH}/vc-$$name ./cmd/$$name; \
+	CC=${CC} CGO_ENABLED=0 $(GOBIN)/gox -osarch=${REL_OSARCH} -ldflags ${LD_FLAGS} -output ${BIN_DIR}/${REL_OSARCH}/vcctl ./cmd/cli
+	for name in controller-manager webhook-manager; do\
+		CC=${CC} CGO_ENABLED=0 $(GOBIN)/gox -osarch=${REL_OSARCH} -ldflags ${LD_FLAGS} -output ${BIN_DIR}/${REL_OSARCH}/vc-$$name ./cmd/$$name; \
 	done
 
+	# scheduler need cgo enabled to support plugin
+	CC=${CC} CGO_ENABLED=1 $(GOBIN)/gox -osarch=${REL_OSARCH} -ldflags ${LD_FLAGS} -output ${BIN_DIR}/${REL_OSARCH}/vc-scheduler ./cmd/scheduler;
```
shinytang6, line 83: hi @zen-xu , it will cause some problem using `gcc` with CGO enabled, l think it will be better to specify the `musl-gcc` in Makefile.

FYI: l just ran `make images` directly to build images and there's something wrong.

zen-xu: what's the wrong message?

shinytang6: `standard_init_linux.go:211: exec user process caused "no such file or directory"`

It should be a cross compilation problem with CGO enabled

zen-xu: yes, I know the reason

zen-xu: Since CI use ubuntu as default image, so I think scheduler should also use `ubuntu`.

shinytang6: `Makefile` is not only used by CI, but users on different platforms may use it. It should not only support a specific platform

zen-xu: If you use `gcc` build the scheduler, you should promise your container has gnu-libc
If you use `musl-gcc` build the scheduler, you should promise your container has musl-libc

shinytang6: But this is a strong assumption. The compilation method needs to be changed according to the environment in which the image is deployed. E.g. I am currently on ubuntu, and I just want to compile vc-scheduler image(with alpine as basic image) through `make images`,  but it doesn't work and you don't even tell me to use `musl-gcc`

zen-xu: I write a [design doc](https://github.com/zen-xu/volcano/blob/custom-plugin-design/docs/design/custom-plugin.md), But has not been merged yet.

shinytang6: okay, now l got it.. but it is still confusing if someone doesn't use the `Custom Plugin` need to know that)

PR #1219, opened 2020-12-24
doc: add custom plugin design doc (#1215)

zen-xu:
ref: #1215 #1218 

### DIFF
--- a/docs/design/custom-plugin.md
+++ b/docs/design/custom-plugin.md
@@ -27,7 +27,7 @@ func (mp *magicPlugin) Name() string {
 }
 
 func New(arguments framework.Arguments) framework.Plugin {  // `New` is PluginBuilder
-    return &magicPlugin{}
+	return &magicPlugin{}
 }
 
 func (mp *magicPlugin) OnSessionOpen(ssn *framework.Session) {}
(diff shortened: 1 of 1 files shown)


docs/design/custom-plugin.md
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (mp *magicPlugin) OnSessionOpen(ssn *framework.Session) {}
+
+func (mp *magicPlugin) OnSessionClose(ssn *framework.Session) {}
+```
+
+### 2. Build the plugin to .so
+
+vc-scheduler binary is build by `musl-gcc`, so we also need to build the plugin by `musl-gcc`.
```
k82cn, line 40: prefer to list known limitations and its options, e.g. different compiler :)

docs/design/custom-plugin.md
```diff
@@ -0,0 +1,128 @@
...
+```yaml
+...
+    containers:
+    - name: volcano-scheduler
+      image: volcano.sh/vc-scheduler:${VERSION}
+      args:
+        - --logtostderr
+        - --scheduler-conf=/volcano.scheduler/volcano-scheduler.conf
+        - -v=3
+        - --plugins-dir=plugins  # specify plugins dir path
```
Thor-wl, line 94: So official plugin will be built and loaded as privous way, and custom plugin can be built and loaded in this way.

zen-xu: yes

Issue #1155, opened 2020-11-17
Request volcano to support k8s 1.19


william-wang:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> /kind feature
As more and more customers upgrade their cluster to k8s  to 1.19 version. We hope volcano can support k8s 1.19 as soon as possible.

**Anything else we need to know?**:

k82cn:
it's better to move api to `volcano-sh/apis` and build client for different version.

Thor-wl:
V1.1.1 is planed to be launched as a bug fix version around late December.

Thor-wl:
This feature has already been supported: https://github.com/volcano-sh/volcano/pull/1444

PR #1249, opened 2021-01-15
update to k8s v1.19

Thor-wl:
update k8s.io to v1.19
related issue: https://github.com/volcano-sh/volcano/issues/1155


k82cn:
let's move related api to `volcano.sh/apis` firstly, and generated client-go by them.

pkg/scheduler/framework/session.go
```diff
@@ -254,8 +258,8 @@ func (ssn *Session) Pipeline(task *api.TaskInfo, hostname string) error {
 }
 
 //Allocate the task to the node in the session
-func (ssn *Session) Allocate(task *api.TaskInfo, hostname string) error {
-	if err := ssn.cache.AllocateVolumes(task, hostname); err != nil {
+func (ssn *Session) Allocate(task *api.TaskInfo, hostname string, podVolumes *volumescheduling.PodVolumes) error {
```
k82cn, line 261: why do you change this interface?

Thor-wl: > why do you change this interface?

k8s v1.19.6 has changed the interface `AssumePodVolume`, arguement `podVolumes` is added and stores pv info at one node

k82cn: we need a design doc for that; the API change may break backward compatibility.

Thor-wl: > we need a design doc for that; the API change may break backward compatibility.

It's invisible for users and has no influence in backward compatibility

k82cn: both Action & Plugins are plugable, we need to keep backward compatibility.

Thor-wl:
@k82cn @william-wang @wpeng102 please take a review

Issue #1101, opened 2020-10-19
v1.2 feature advice: reserve specified resources for specified queue


Thor-wl:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
> /kind feature

**What happened**:
1. support reserving specified resources for specified queue
2. reservation without preemption, no influence with existing running workload
3. support enable and disable resource reservation for specified queue dynamiclly without restarting Volcano
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version: v1.2
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hudson741:
this will be a big improve. !!!!!

hudson741:
@Thor-wl  I will take this issue

Thor-wl:
ref: https://github.com/volcano-sh/volcano/pull/1771

PR #1253, opened 2021-01-19
update crd admission for Queue Resource Reservation

Thor-wl:
related issue: https://github.com/volcano-sh/volcano/issues/1101


Thor-wl:
As queue resource reservation is delaied temporarily, will continue the development later some time.

Issue #1256, opened 2021-01-20
fix: more pods are reclaimed than need


huone1:
**Is this a BUG REPORT or FEATURE REQUEST?**:
> /kind bug

**What happened**:
the cluster spec: 9 cpu 16GB

1. set reclaim action in scheduler configmap, 
2. create queue "test" weight: 3
2. apply a job test-1(replicas:8, minavailable: 3 queue: default each task request cpu: 1 ) 
3. apply another job test-2 (replicas:2, minavailable: 2, queue: test each task request cpu: 1)

result:
test-1 delete 5 pods
**What you expected to happen**:
test-1 delete 2 pods

**How to reproduce it (as minimally and precisely as possible)**:
The above operations

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
  latest
- Kubernetes version (use `kubectl version`):
1.17
- OS (e.g. from /etc/os-release):
-centos

huone1:
fix it in PR #1255

huone1:
last fix has a problemand reset it , the new fix is PR #1260

PR #1260, opened 2021-01-21
fix: more pods are reclaimed than required

huone1:


wpeng102:
https://github.com/volcano-sh/volcano/issues/1256

PR #1263, opened 2021-01-23
feat(controller): update ssh secret when job updated

shinytang6:


shinytang6:
There seems to be some problems with this implementation, which will cause ssh to fail. The reason is that it will create some symlinks("..data" dir) without subpath. But with subpath, configmap will not be updated.

ref:
1. https://github.com/kubernetes/kubernetes/issues/50345
2. https://stackoverflow.com/questions/50685385/kubernetes-config-map-symlinks-data-is-there-a-way-to-avoid-them

shinytang6:
The only way l thought of is to set the ssh config `StrictModes no` explicitly...

Issue #1264, opened 2021-01-25
Where to Use musl-libc in code at branch master？？


huone1:
Where to Use musl-libc in code at branch master？？

currently， make image （scheduler）need  musl-libc， but  musl-libc is NO Exist at the Common compilation environment；
It's very unfriendly for new students.
branch： master

huone1:
@zen-xu

zen-xu:
The default base image of `scheduler` is alpine, which only provides `musl-libc`. 
If your want to build the image on ubuntu system by youself, you can specify the base image like `ubuntu` in Dockerfile.

k82cn:
is there any way to disable that by default? Most of user does not care about the compiler :)

huone1:
OK

PR #1266, opened 2021-01-26
add Makefile flag `SUPPORT_PLUGINS`

zen-xu:
#1264 


zen-xu:
If want to disable scheduler with plugin, add flag `SUPPORT_PLUGINS=no` after make

k82cn:
prefer to disable it by default, and update doc of `plugin` on how to enable it :)

PR #1267, opened 2021-01-26
fix: issue#1261 config "overcommit-factor" less than 1 leads scheduler crash

huone1:
kaihua give better advice： check OverCommitFactor in webhook validate。
so add webhook validate ，just for scheduler configmap；
currently， there are only some check for enqueue action， and add other actions or plugins ‘s check  to it in future

### DIFF
--- a/pkg/scheduler/util.go
+++ b/pkg/scheduler/util.go
@@ -34,6 +34,7 @@ tiers:
 - plugins:
   - name: priority
   - name: gang
+  - name: conformance
 - plugins:
   - name: drf
   - name: predicates
--- a/pkg/webhooks/admission/configmap/validate/admit_cm.go
+++ b/pkg/webhooks/admission/configmap/validate/admit_cm.go
@@ -1,0 +1,143 @@
+package validate
+
+import (
+	"fmt"
+
+	yaml "gopkg.in/yaml.v2"
+
+	"k8s.io/api/admission/v1beta1"
+	whv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+	"strings"
+	"volcano.sh/volcano/pkg/webhooks/schema"
+	"volcano.sh/volcano/pkg/webhooks/util"
+
+	"volcano.sh/volcano/pkg/scheduler/conf"
+	"volcano.sh/volcano/pkg/webhooks/router"
+)
+
+const (
+	configmapName          = "volcano-scheduler-configmap"
+	enqueueName            = "enqueue"
+	overCommitFactor       = "overcommit-factor"
+	overCommitFactorMinVal = 1.0
+)
+
+func init() {
+	router.RegisterAdmission(service)
+}
+
+var service = &router.AdmissionService{
+	Path: "/configmaps/validate",
+	Func: AdmitConfigmaps,
+
+	Config: config,
+
+	ValidatingConfig: &whv1beta1.ValidatingWebhookConfiguration{
+		Webhooks: []whv1beta1.ValidatingWebhook{{
+			Name: "validateconfigmap.volcano.sh",
+			Rules: []whv1beta1.RuleWithOperations{
+				{
+					Operations: []whv1beta1.OperationType{whv1beta1.Create, whv1beta1.Update},
+					Rule: whv1beta1.Rule{
+						APIGroups:   []string{""},
+						APIVersions: []string{"v1"},
+						Resources:   []string{"configmaps"},
+					},
+				},
+			},
+		}},
+	},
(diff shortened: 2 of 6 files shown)


pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -170,5 +171,9 @@ func (enqueue *Action) getOverCommitFactor(ssn *framework.Session) float64 {
 		arg.GetFloat64(&factor, overCommitFactor)
 	}
 
+	if factor < minOverCommitFactor {
+		factor = minOverCommitFactor
+	}
+
```
shinytang6, line 177: Could you add a unit test for that

shinytang6 (changes requested):

shinytang6 (approved):

pkg/scheduler/actions/enqueue/enqueue.go
huone1: I've added ut just now.

pkg/webhooks/admission/configmap/validate/util.go
```diff
@@ -0,0 +1,43 @@
+package validate
```
shinytang6, line 1: These functions should already exist in codebase？

shinytang6 (changes requested):

pkg/scheduler/util.go
```diff
@@ -34,6 +34,7 @@ tiers:
 - plugins:
   - name: priority
   - name: gang
+  - name: conformance
 - plugins:
   - name: drf
   - name: predicates
```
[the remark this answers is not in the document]
huone1: I was attend to add binpack because it is not  fit all scenarios， so it wasn't added.

pkg/webhooks/admission/configmap/validate/util.go
huone1: yes ， but it is in scheduler folder， so I repeat some

huone1: yes ， but it is in scheduler folder， so I repeat some

pkg/webhooks/admission/configmap/validate/admit_cm.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if !reviewResponse.Allowed {
+		reviewResponse.Result = &metav1.Status{Message: strings.TrimSpace(msg)}
+	}
+	return &reviewResponse
+}
+
+/*
+allow config to create and update when
+1. configmap name must be "volcano-scheduler-configmap"
```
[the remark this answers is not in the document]
huone1: I check helm files ， you're right。
configmap name Transfer to  admission webhook through ENV ， is it OK ？？

pkg/scheduler/util.go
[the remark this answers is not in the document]
shinytang6: The default cm under `installer/volcano-development.yaml` contains `binpack`, maybe we should remove `binpack` in that file?

pkg/webhooks/admission/configmap/validate/admit_cm.go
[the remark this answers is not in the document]
shinytang6: emm, l think it's enough to just remove this name judgement?

pkg/webhooks/admission/configmap/validate/admit_cm.go
```diff
@@ -0,0 +1,143 @@
...
+func validateConfigmap(cm *v1.ConfigMap, reviewResponse *v1beta1.AdmissionResponse) string {
+	msg := ""
+	if cm.Name != configmapName {
+		return msg
+	}
+
+	confKey := ""
+	confNum := 0
+	for key := range cm.Data {
+		if strings.HasSuffix(key, ".conf") {
```
shinytang6, line 99: also `.conf` is not a necessity

huone1: 1，it  watch all configmap, we must filter them, so it is necessary to adjust the configmap name
2。i think  it is necessary to  certain restrictions on the format of scheduler configmap  data

shinytang6: Okay got it, I have no opinion on this change, but this introduces some name restrictions indeed.

k82cn:
prefer not to check scheduler's arguments in webhook.

huone1:
> prefer not to check scheduler's arguments in webhook.

Another way ： scheduler process gets the  scheduler's arguments to check them， and if  one  arguments is illegal，reset it to default val;

k82cn:
> Another way ： scheduler process gets the scheduler's arguments to check them， and if one arguments is illegal，reset it to default val;

That's better to me, please also record an event and log for that :)

huone1:
the question is resloved by #1324 , so  closed it

PR #1268, opened 2021-01-26
fix(controller): disable strict mode for mpi examples

shinytang6:
Can not find a better to solve [this issue](https://github.com/volcano-sh/volcano/pull/1263#issuecomment-766567292)

### DIFF
--- a/example/integrations/mpi/mpi-example.yaml
+++ b/example/integrations/mpi/mpi-example.yaml
@@ -21,6 +21,9 @@ spec:
                 - /bin/sh
                 - -c
                 - |
+                  cat /etc/ssh/sshd_config | grep -v StrictModes > /etc/ssh/sshd_config.new && \
+                  echo "StrictModes no" >> /etc/ssh/sshd_config.new && \
+                  mv /etc/ssh/sshd_config.new /etc/ssh/sshd_config;
                   MPI_HOST=`cat /etc/volcano/mpiworker.host | tr "\n" ","`;
                   mkdir -p /var/run/sshd; /usr/sbin/sshd;
                   mpiexec --allow-run-as-root --host ${MPI_HOST} -np 2 mpi_hello_world;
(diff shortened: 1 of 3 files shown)


example/integrations/mpi/mpi-example.yaml
```diff
@@ -21,6 +21,9 @@ spec:
                 - /bin/sh
                 - -c
                 - |
+                  cat /etc/ssh/sshd_config | grep -v StrictModes > /etc/ssh/sshd_config.new && \
```
k82cn, line 24: Why do we need this one? It's a little complex to me :)

shinytang6:
@k82cn After I removed the `subpath` in [PR 1263](https://github.com/volcano-sh/volcano/pull/1263), secret volume relies on symlinks to update secret, ssh secret directory looks something like this:
```
drwxr-xr-x 2 root root 120 Jan 28 06:43 ..2021_01_28_06_43_07.443659693/
lrwxrwxrwx 1 root root  31 Jan 28 06:43 ..data -> ..2021_01_28_06_43_07.443659693/
lrwxrwxrwx 1 root root  22 Jan 25 03:54 authorized_keys -> ..data/authorized_keys
lrwxrwxrwx 1 root root  13 Jan 25 03:54 config -> ..data/config
lrwxrwxrwx 1 root root  13 Jan 25 03:54 id_rsa -> ..data/id_rsa
lrwxrwxrwx 1 root root  17 Jan 25 03:54 id_rsa.pub -> ..data/id_rsa.pub
```
which will cause ssh communication between pods to fail due to directory permissions.

lf use `subpath` as before, these symlinks will not be generated(so that ssh communication works fine), but the flaw is that secret updates cannot be received.

The only way l thought of was to modify the server `sshd_config` to avoid that..

k82cn:
> lf use subpath as before, these symlinks will not be generated(so that ssh communication works fine), but the flaw is that secret updates cannot be received.

Is it possible to enhance k8s for that case?

shinytang6:
> > lf use subpath as before, these symlinks will not be generated(so that ssh communication works fine), but the flaw is that secret updates cannot be received.
> 
> Is it possible to enhance k8s for that case?

The k8s official does not seem to have this plan recently, the documentation has clearly pointed out this problem.
```
Note: A container using a Secret as a subPath volume mount will not receive Secret updates.
```

There are also some discussions here: [https://github.com/kubernetes/kubernetes/issues/50345](https://github.com/kubernetes/kubernetes/issues/50345)

k82cn:
hm… interesting, I think we can share our case to upstream for more input.

shinytang6:
> hm… interesting, I think we can share our case to upstream for more input.

Okay, I will raise an issue to k8s later. But WDYT is a temporary solution by now, to revert subpath changes or modify the sshd configuration? Due to the reason that the current codebase will make mpi-job unavailable.

k82cn:
prefer to revert subpath changes :)

Thor-wl:
I'm ok about the commits. Please resolve the conflicts.

PR #1269, opened 2021-01-26
feat: add TDM（time division multiplexing) plugin

yahaa:


pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+)
+
+const (
+	// PluginName indicates name of volcano scheduler plugin.
+	PluginName = "tdm"
+	// PreemptableTaskKey pod annotation key
+	PreemptableTaskKey = "volcano.sh/preemptable"
```
k82cn, line 33: prefer to move them to global const, so other plugins may also reuse this one :)

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,113 @@
...
+		if !ok {
+			return score, nil
+		}
+
+		if err := bp.availableRevocableZone(rz, ssn); err != nil {
+			klog.V(4).Infof("TDM plugin rz not available %s", err)
+			return score, nil
+		}
+
+		if _, ok := task.Pod.Annotations[PreemptableTaskKey]; !ok {
```
k82cn, line 77: Let's add `Preemptable` to `TaskInfo` :)

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,113 @@
...
+func (bp *tdmPlugin) OnSessionOpen(ssn *framework.Session) {
+	klog.V(4).Infof("Enter tdm plugin ...")
+	if klog.V(4) {
+		defer func() {
+			klog.V(4).Infof("Leaving tdm plugin.")
+		}()
+	}
+
+	// tdm plugin just handle revocable node
+	nodeOrderFn := func(task *api.TaskInfo, node *api.NodeInfo) (float64, error) {
```
k82cn, line 64: hm... in your case, `nodeOrderFn` maybe unnecessary :)

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,257 @@
...
+
+			victims = append(victims, item)
+		}
+
+		klog.V(4).Infof("TDM victims are %+v", victims)
+
+		return victims
+	}
+
+	victimsFn := func() []*api.TaskInfo {
```
wpeng102, line 190: we need support expel step by step, so in victimsFn not return all the victim `taskInfo` in one scheduler period.

pkg/scheduler/plugins/tdm/tdm.go
yahaa: Moved it into `pkg/scheduler/api/well-known-labels. go`

pkg/scheduler/plugins/tdm/tdm.go
yahaa: Ok, it done.

pkg/scheduler/plugins/tdm/tdm.go
yahaa: I think it was necessary, preemptable task can schedule to all kind of node, revocable node just have the high priority.

pkg/scheduler/plugins/tdm/tdm.go
yahaa: You are right.

pkg/scheduler/plugins/tdm/tdm.go
k82cn: ok, great!

pkg/scheduler/api/pod_info.go
```diff
@@ -67,6 +68,18 @@ func GetPodResourceRequest(pod *v1.Pod) *Resource {
 	return result
 }
 
+// GetPodPreemptable return volcano.sh/preemptable value for pod
+func GetPodPreemptable(pod *v1.Pod) bool {
+	if len(pod.Annotations) > 0 {
+		if value, found := pod.Annotations[v1beta1.PodPreemptable]; found {
+			if value == "true" {
```
k82cn, line 75: prefer to use `strconv.ParseBool`

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,315 @@
...
+			if err := bp.availableRevocableZone(rz); err != nil {
+				// rz disactive, then evict preemptable tasks from the revocable node
+				victims = append(victims, bp.revocableNodePreemptableTask(rz, ssn)...)
+			} else {
+				active[rz] = struct{}{}
+			}
+		}
+
+		// if there are some active rz, then evict preemptable tasks from all the none revocable node
+		victims = append(victims, bp.noneRevocableNodePreemptableTask(ssn)...)
```
wpeng102, line 250: It seems there is no check if has active rz. If there is no active rz, the preemtable pod on non revocable zone will be evicted again and again.

yahaa: Yes, fixed.

wpeng102:
suggest use the following configure to test:
```
actions: "enqueue, allocate, backfill, preempt"
    tiers:
    - plugins:
      - name: tdm
        arguments:
          tdm.revocableZone.rz1: 1:00-4:00
          tdm.evict.period: 1m
    - plugins:
      - name: priority
      - name: gang
      - name: conformance
    - plugins:
      - name: drf
      - name: predicates
      - name: proportion
      - name: nodeorder
      - name: binpack

```

yahaa:
> suggest use the following configure to test:
> 
> ```
> actions: "enqueue, allocate, backfill, preempt"
>     tiers:
>     - plugins:
>       - name: tdm
>         arguments:
>           tdm.revocableZone.rz1: 1:00-4:00
>           tdm.evict.period: 1m
>     - plugins:
>       - name: priority
>       - name: gang
>       - name: conformance
>     - plugins:
>       - name: drf
>       - name: predicates
>       - name: proportion
>       - name: nodeorder
>       - name: binpack
> ```

Got it.

wpeng102:
The framework and basic funtion is OK to me. Suggest merge this PR first.

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,333 @@
...
+	ssn.AddPredicateFn(bp.Name(), predicateFn)
+	ssn.AddNodeOrderFn(bp.Name(), nodeOrderFn)
+	ssn.AddPreemptableFn(bp.Name(), preemptableFn)
+	ssn.AddVictimTasksFns(bp.Name(), victimsFn)
+}
+
+func (bp *tdmPlugin) maxVictims(victims []*api.TaskInfo) []*api.TaskInfo {
+	targetNum := 0
+	switch bp.evictMaxNum.Type {
+	case intstr.Int:
```
k82cn, line 281: why check its type? let's convert it to float when got this parameter.

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,333 @@
...
+		targetNum = int(math.Min(float64(bp.evictMaxNum.IntValue()), float64(len(victims))))
```
k82cn, line 282: are we going to select Task/Pod by Job/PodGroup, e.g. not only get pod from one PodGroup.

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	return victims[:targetNum]
+}
+
+func (bp *tdmPlugin) revocableNodePreemptableTask(rz string, ssn *framework.Session) []*api.TaskInfo {
+	tasks := make([]*api.TaskInfo, 0)
+
+	for _, node := range ssn.Nodes {
+		if node.RevocableZone != rz {
```
k82cn, line 298: lets' build an index in cache for revocable nodes.

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,333 @@
...
+		score = float64(v1alpha1.MaxNodeScore)
+
+		klog.V(4).Infof("TDM score for Task %s/%s on node %s is: %v", task.Namespace, task.Name, node.Name, score)
+		return score, nil
+	}
+
+	preemptableFn := func(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) []*api.TaskInfo {
+		if preemptor.Preemptable {
+			klog.V(4).Infof("TDM task %s/%s is preemptable, do nothing skip", preemptor.Namespace, preemptor.Name)
+			return preemptees
```
k82cn, line 203: return nil?

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -0,0 +1,333 @@
...
+				active[rz] = struct{}{}
+			}
+		}
+
+		// if there are some active rz, then evict preemptable tasks from all the none revocable node
+		if len(active) > 0 {
+			victims = append(victims, bp.noneRevocableNodePreemptableTask(ssn)...)
+		}
+
+		victims = bp.maxVictims(victims)
```
k82cn, line 239: It requires to preempt pods in one node, so `maxVictims` is not suitable for this.

pkg/scheduler/plugins/tdm/tdm.go
yahaa: I want evictMaxNum support two kind of parameter, such as
* evictMaxNum: 5    （this mean that user can specify exactly number of tasks)
* evictMaxNum: 5%  

Dose it make sense?

pkg/scheduler/plugins/tdm/tdm.go
yahaa: Yes, it should return nil.

pkg/scheduler/plugins/tdm/tdm.go
wpeng102: OK，I will use a new PR to handle this comments.

pkg/scheduler/plugins/tdm/tdm.go
yahaa: Your mean in this preemptableFn we no need to call  `bp.maxVictims()`?

yahaa: Yes, it really don't need.

pkg/scheduler/plugins/tdm/tdm.go
yahaa: OK，I will use a new PR to handle this comments, after we support evictMaxNum params in workload.

PR #1282, opened 2021-02-01
update job-resource-reservation-design doc

Thor-wl:
* compelete implementation and recommend practice for v1.1.0
* add update points for v1.2.1

### DIFF
--- a/docs/design/job-resource-reservation-design.md
+++ b/docs/design/job-resource-reservation-design.md
@@ -1,10 +1,10 @@
-# Volcano Resource Reservation For Big Jobs
+# Volcano Resource Reservation For Target Jobs
 
 @[Thor-wl](https://github.com/Thor-wl); Aug 19th, 2020
 
 ## Motivation
 As [issue 13](https://github.com/volcano-sh/volcano/issues/13) / [issue 748](https://github.com/volcano-sh/volcano/issues/748) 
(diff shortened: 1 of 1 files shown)


docs/design/job-resource-reservation-design.md
```diff
@@ -81,16 +82,67 @@ task type is unpredictable and complex.
...
-Add new Plugin reservation to implement algorithm detail above.
-
+Add new Plugin reservation to implement algorithm detail about selecting target job and reserving nodes. `targetJobFn`
+selects job whose priority is the highest and waits for the longest time. `reservedNodesFn` reserve node whose idle resource
+is the max in every session.
 ![Workflow](./images/reservation_workflow.png)
+### Recommend practice
+An example how to make use of this feature is to configure scheduler's configuration as follows:
+```yaml
+actions: "enqueue, elect, allocate, backfill, reserve"
```
k82cn, line 104: why do we need `elect` action? A callback/plugin function in `reserve` should be enough.

jiangkaihua: `elect` action is used to elect jobs labeled by user manually or selected by plugin automatically, then `reserve` action reserves nodes according to jobs and related plugins. So in my opinion, `elect` is necessary, but `reserve` could merge into `elect`, or filter nodes directly in `ssn.predicateFn()` by callback functions in `reservation` plugin.

PR #1287, opened 2021-02-03
[tdm plugin] support minAlive and evictMaxNum for job

wpeng102:
support `volcano.sh/pod-min-alive` and `volcano.sh/evict-max-step` in podgroup/pod level. 
we can specify these two config in annotation or label

### DIFF
--- a/pkg/apis/scheduling/v1beta1/labels.go
+++ b/pkg/apis/scheduling/v1beta1/labels.go
@@ -37,3 +37,9 @@ const PodPreemptable = "volcano.sh/preemptable"
 
 //NodeRevocableZone is the key of revocable-zone
 const NodeRevocableZone = "volcano.sh/revocable-zone"
+
+// PodMinAlive is the key of min alive pod number
+const PodMinAlive = "volcano.sh/pod-min-alive"
+
+// PodEvictMaxStep is the key of min alive pod number
+const PodEvictMaxStep = "volcano.sh/evict-max-step"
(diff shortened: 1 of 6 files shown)


pkg/apis/scheduling/v1beta1/labels.go
```diff
@@ -37,3 +37,9 @@ const PodPreemptable = "volcano.sh/preemptable"
 
 //NodeRevocableZone is the key of revocable-zone
 const NodeRevocableZone = "volcano.sh/revocable-zone"
+
+// PodMinAlive is the key of min alive pod number
+const PodMinAlive = "volcano.sh/pod-min-alive"
```
k82cn, line 42: s/pod-min-alive/min-pod-alive

pkg/apis/scheduling/v1beta1/labels.go
```diff
@@ -37,3 +37,9 @@ const PodPreemptable = "volcano.sh/preemptable"
...
+
+// PodEvictMaxStep is the key of min alive pod number
+const PodEvictMaxStep = "volcano.sh/evict-max-step"
```
k82cn, line 45: s/evict-max-step/max-evict-step

pkg/scheduler/api/job_info.go
```diff
@@ -153,7 +155,9 @@ type JobInfo struct {
 
 	ScheduleStartTimestamp metav1.Time
 
-	Preemptable bool
+	Preemptable     bool
+	PodAliveNum     string
+	PodEvictMaxStep string
```
k82cn, line 160: align with new name

PR #1243, opened 2021-01-08
use nodeMap for predicates

njlkj:
#1231 
/assign @william-wang

PR #1292, opened 2021-02-04
[fix#1291]optimize nodeorder plugin

huone1:

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -130,7 +130,7 @@ func calculateWeight(args framework.Arguments) priorityWeight {
 
 func (pp *nodeOrderPlugin) OnSessionOpen(ssn *framework.Session) {
 	weight := calculateWeight(pp.pluginArguments)
-	pl := util.NewPodLister(ssn)
+	pl := util.NewPodListerFromNode(ssn)
 	pods, _ := pl.List(labels.NewSelector())
 	nodeMap, nodeSlice := util.GenerateNodeMapAndSlice(ssn.Nodes)
 
--- a/pkg/scheduler/plugins/util/util.go
+++ b/pkg/scheduler/plugins/util/util.go
@@ -83,6 +83,33 @@ func NewPodLister(ssn *framework.Session) *PodLister {
 	return pl
 }
 
+// NewPodListerFromNode returns a PodLister generate from ssn
+func NewPodListerFromNode(ssn *framework.Session) *PodLister {
+	pl := &PodLister{
+		Session:          ssn,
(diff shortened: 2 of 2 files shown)


k82cn:
how to verify this case?

william-wang:
Please show the case that we experienced and the relation with your fix. @huone1

huone1:
> how to verify this case?
check it by the sorce of nodeorder 
reproduce oper:
there are three node:
node 192.168.121.209 cpu releasing 0, Idle 3310, Used 610, Pipelined 0
node 192.168.121.54 cpu releasing 0, Idle 3060, Used 860, Pipelined 0
node 192.168.121.62 cpu releasing 0, Idle 3520, Used 400, Pipelined 0

submit three jobs with the request of 2 cpu
node 192.168.121.54 cpu releasing 0, Idle 1060, Used 2860, Pipelined 0
node 192.168.121.62 cpu releasing 0, Idle 1520, Used 2400, Pipelined 0
node 192.168.121.209 cpu releasing 0, Idle 1310, Used 2610, Pipelined 0
delete the pod in node 192.168.121.54, then submit one job with the request of 1.5 cpu,
observe the pod info:
you can find that the new pod is pending until the pod in node 192.168.121.54 terminated，because new pod is scheduled in node 192.168.121.54 by volcano

huone1:
by review K8s code and test by default schduler , such as BalancedAllocation.score works  base on the all pod request resource at node

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -130,7 +130,7 @@ func calculateWeight(args framework.Arguments) priorityWeight {
 
 func (pp *nodeOrderPlugin) OnSessionOpen(ssn *framework.Session) {
 	weight := calculateWeight(pp.pluginArguments)
-	pl := util.NewPodLister(ssn)
+	pl := util.NewPodListerFromNode(ssn)
```
shinytang6, line 133: Will [PR 1243](https://github.com/volcano-sh/volcano/pull/1243) resolves this issue? l think in that pr `nodemap` contains all status' pods

huone1: nodeOrderFn function score doesn't base  on  nodemap   but base on handle [handle := k8s.NewFrameworkHandle(pods, nodeSlice) ], and  pods info in handle struct data is create by "pl := util.NewPodLister(ssn)"

shinytang6: yes, but in that PR it modifies the function `NewFrameworkHandle` so that it doesn't depend on `pods` param any more, then the snapshot will contain all the pods.

PR #1293, opened 2021-02-05
build cache for revocable nodes

yahaa:
Build cache for revocable node.


pkg/scheduler/cache/cache.go
```diff
@@ -684,6 +685,10 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 		}
 
 		snapshot.Nodes[value.Name] = value.Clone()
+
+		if value.RevocableZone != "" {
+			snapshot.RevocableNodes[value.Name] = value.Clone()
```
k82cn, line 690: `RevolcanoNodes` should be an index, so just one copy of `value.Clone()` should be enough.

yahaa: Fixed it.

PR #1305, opened 2021-02-09
fix(job_controller): Fix log

gaocegege:
We are using both `klog.V(3).Infof` and `klog.Errorf` when checking the errors.

Like https://github.com/volcano-sh/volcano/compare/master...gaocegege:log#diff-28557d59525e85335ee0eacff9269dcc0dbc0ffcc7ef146ef4e0e59d713df788L537 and https://github.com/volcano-sh/volcano/compare/master...gaocegege:log#diff-28557d59525e85335ee0eacff9269dcc0dbc0ffcc7ef146ef4e0e59d713df788L487

I think it is better to keep using Errorf for errors.

Signed-off-by: cegao <cegao@tencent.com>


hzxuzhonghu (commented):
/lgtm

PR #1316, opened 2021-02-18
add queue resource reservation design v2 doc

Thor-wl:


docs/design/queue-resource-reservation-design-v2.md
```diff
@@ -0,0 +1,35 @@
+# Volcano Resource Reservation For Queue(v2)
+
+@[Thor-wl](https://github.com/Thor-wl); Feb 28th, 2021
```
jiangkaihua, line 3: Feb 18th, 2021

docs/design/queue-resource-reservation-design-v2.md
```diff
@@ -0,0 +1,35 @@
...
+
+Inspired by design optimization of `target job resource reservation`, our goal is to ensure the SLA of jobs in target
+queues. So just allocate resource for jobs by order SLA.
+
+## Design
+Queues with annotation `maxWaitingTime` indicate that this queue should ensure SLA. Jobs in the queue will be scheduled
+first if its SLA
```
jiangkaihua, line 10: Do you mean "if its SLA arrives."?

docs/design/queue-resource-reservation-design-v2.md
```diff
@@ -0,0 +1,35 @@
...
+### API
+```
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: q1
+  annotations:
+    # maxWaitingTime indicates the maximal pending time for jobs in the queue. 
+    # If not scheduled after waiting for maxWaitTime, jobs will be scheduled first.
+    maxWaitingTime: 30m
```
jiangkaihua, line 22: if we add `maxWaitingTime` in queue annotation, then any user can set SLA on queue, which should only be modified by admin in my opinion. So would it be better to set `maxWaitingTime` in some other field like `status`?

docs/design/queue-resource-reservation-design-v2.md
Thor-wl: done

docs/design/queue-resource-reservation-design-v2.md
Thor-wl: > if we add `maxWaitingTime` in queue annotation, then any user can set SLA on queue, which should only be modified by admin in my opinion. So would it be better to set `maxWaitingTime` in some other field like `status`?

Yes. Do you have any other good ideas?

docs/design/queue-resource-reservation-design-v2.md
Thor-wl: > Do you mean "if its SLA arrives."?

Sorry for the uncompletion of the doc. I've supplied it.

docs/design/queue-resource-reservation-design-v2.md
```diff
@@ -0,0 +1,34 @@
...
+## Implementation
+### API
+```
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: q1
+  annotations:
+    # maxWaitingTime indicates the maximal pending time for jobs in the queue. 
+    # If not scheduled after waiting for maxWaitTime, jobs will be scheduled first.
```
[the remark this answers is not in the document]
Thor-wl: > Another possible question is how to deal with jobs if exceed the maxWaitingTime。 stop the jobs? or try to schedule until ...?

Just as what describes in doc https://github.com/volcano-sh/volcano/blob/master/docs/design/sla-plugin.md, `maxWaitingTime` indicates the priority the job can be enqueued and allocated resource. If exceed the `maxWaitTime`,  the priority of the job will be increased and allocated resource as soon as possible.

PR #1340, opened 2021-03-01
Add mpi plugin.

acmore:
The ssh plugin supports mpi, but it requires privileges to launch ssh agent. This PR adds a mpi plugin to launch horovod processes using kubectl.

### DIFF
--- a/pkg/controllers/job/plugins/mpi/mpi.go
+++ b/pkg/controllers/job/plugins/mpi/mpi.go
@@ -1,0 +1,340 @@
+/*
+Copyright 2019 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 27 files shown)


pkg/controllers/job/plugins/mpi/mpi.go
```diff
@@ -0,0 +1,340 @@
+/*
+Copyright 2019 The Volcano Authors.
```
Thor-wl, line 2: Please update the copyright up to date.

Thor-wl:
please resolve the DOC

pkg/controllers/job/plugins/mpi/mpi.go
acmore: done

PR #1344, opened 2021-03-04
fix: prevent SelectBestNode func arise panic

yahaa:
Prevent SelectBestNode func arise panic when the length of nodeScores is zero.


yahaa:
> /lgtm
> /approve

Thanks.

PR #1345, opened 2021-03-04
feat(scheduler): support resource quota consideration during pod group enqueue procedure

merryzhou:
https://github.com/volcano-sh/volcano/issues/1014


zen-xu:
Add `event` will be more  user-friendly

merryzhou:
> Add `event` will be more user-friendly

done

shinytang6:
l think we also need a doc for this plugin ?

pkg/scheduler/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+func (rq *resourceQuotaPlugin) Name() string {
+	return PluginName
+}
+
+func (rq *resourceQuotaPlugin) OnSessionOpen(ssn *framework.Session) {
+	ssn.AddJobEnqueueableFn(rq.Name(), func(obj interface{}) int {
+		job := obj.(*api.JobInfo)
```
Robert-Christensen-visa, line 38:
```suggestion
func (rq *resourceQuotaPlugin) OnSessionOpen(ssn *framework.Session) {
	pendingResources := make(map[string]v1.ResourceList
        
	ssn.AddJobEnqueueableFn(rq.Name(), func(obj interface{}) int {
		job := obj.(*api.JobInfo)
```

pkg/scheduler/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,72 @@
...
+			klog.V(3).Infof("job: %s/%s minQuotas not exist, skip check resource quota", job.Namespace, job.Name)
+			return util.Permit
+		}
+
+		quotas := ssn.NamespaceInfo[api.NamespaceName(job.Namespace)].RQStatus
+		for _, resourceQuota := range quotas {
+			hardResources := quota.ResourceNames(resourceQuota.Hard)
+			requestedUsage := quota.Mask(*resourcesRequests, hardResources)
+			newUsage := quota.Add(resourceQuota.Used, requestedUsage)
+			maskedNewUsage := quota.Mask(newUsage, quota.ResourceNames(requestedUsage))
```
Robert-Christensen-visa, line 51:
```suggestion
			hardResources := quota.ResourceNames(resourceQuota.Hard)
			requestedUsage := quota.Mask(*resourcesRequests, hardResources)
			var resourcesUsed = resourceQuota.Used
			if pendingUse, found := pendingResources[job.Namespace]; found {
			        resourcesUsed = quota.Add(pendingUse, resourcesUsed)
			}
			newUsage := quota.Add(resourcesUsed, requestedUsage)
			maskedNewUsage := quota.Mask(newUsage, quota.ResourceNames(requestedUsage))
```

pkg/scheduler/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,72 @@
...
+					failedRequestedUsage,
+					failedUsed,
+					failedHard,
+				)
+				klog.V(3).Infof("enqueueable false for job: %s/%s, because :%s", job.Namespace, job.Name, msg)
+				ssn.RecordPodGroupEvent(job.PodGroup, v1.EventTypeNormal, string(scheduling.PodGroupUnschedulableType), msg)
+				return util.Reject
+			}
+		}
+		return util.Permit
```
Robert-Christensen-visa, line 67:
```suggestion
		}
		if _, found := pendingResources[job.Namespace]; !found {
		        pendingResources[job.Namespace] = v1.ResourceList{}
		}
		pendingResources[job.Namespace] = quota.Add(pendingResource[job.Namespace], *resourcesRequests)
		
		return util.Permit
```

Robert-Christensen-visa (commented):
This plugin would be useful and would resolve an issue we are encountering in our setup, so I hope it gets merged soon.  This is really great!

It looks like there might be a corner case this is not covered covered.

I assume this plugin will look at pending podgroups and will enqueue them only if there is enough capacity in the namespace according to Kubernetes ResourceQuota.

If many jobs small jobs are submitted to a namespace at the same time, it is possible for all the jobs to be enqueued because this plugin is not keeping state of previous podgroups that have been permitted in the current scheduling round.  Take this example:

```json
apiVersion: v1
kind: ResourceQuota
metadata:
  name: small-quota
spec:
  hard:
    cpu: "2"
```

Say I have many small jobs I want to queue in the namespace `small` that has the above `ResourceQuota`.  Here is a template I used.  I saved this file as `job_template.json`:

```json
kind: Job
metadata:
  name: vcjob-job-<id>
spec:
  minAvailable: 1
  schedulerName: volcano
  tasks:
    - replicas: 1
       name: "worker"
       template:
        spec:
          containers:
            - image: ubuntu
              imagePullPolicy: IfNotPresent
              name: ubuntu
              command: ["sh", "-c", "sleep 60"]
              resources:
                limits:
                 cpu: "1"
                 memory: 64Mi
                requests:
                 cpu: "1"
                 memory: 64Mi
            restartPolicy: OnFailure
```

I submit many of these jobs at once, using this script:

```bash
for id in $(seq 32)
do
  sed "s/<id>/job-$id/g" < job_template.json | \
  kubectl apply -n small -f -
done
```

When each job is considered individually using the `resourcequota` plugin, they will all be permitted.  I don't think this should be the case, I think the plugin will need to consider other podgroups that have already been permitted in the scheduling round to prevent it from enqueueing too many podgroups and exceeding the namespace resource quota.

I have made some suggestions that would fix this issue.  This assumes the order of jobs submitted is in priority order.

Thor-wl:
> This plugin would be useful and would resolve an issue we are encountering in our setup, so I hope it gets merged soon. This is really great!
> 
> It looks like there might be a corner case this is not covered covered.
> 
> I assume this plugin will look at pending podgroups and will enqueue them only if there is enough capacity in the namespace according to Kubernetes ResourceQuota.
> 
> If many jobs small jobs are submitted to a namespace at the same time, it is possible for all the jobs to be enqueued because this plugin is not keeping state of previous podgroups that have been permitted in the current scheduling round. Take this example:
> 
> ```json
> apiVersion: v1
> kind: ResourceQuota
> metadata:
>   name: small-quota
> spec:
>   hard:
>     cpu: "2"
> ```
> 
> Say I have many small jobs I want to queue in the namespace `small` that has the above `ResourceQuota`. Here is a template I used. I saved this file as `job_template.json`:
> 
> ```json
> apiVersion: batch.volcano.sh/v1alpha1
> kind: Job
> metadata:
>   name: vcjob-job-<id>
> spec:
>   minAvailable: 1
>   schedulerName: volcano
>   tasks:
>     - replicas: 1
>        name: "worker"
>        template:
>         spec:
>           containers:
>             - image: ubuntu
>               imagePullPolicy: IfNotPresent
>               name: ubuntu
>               command: ["sh", "-c", "sleep 60"]
>               resources:
>                 limits:
>                  cpu: "1"
>                  memory: 64Mi
>                 requests:
>                  cpu: "1"
>                  memory: 64Mi
>             restartPolicy: OnFailure
> ```
> 
> I submit many of these jobs at once, using this script:
> 
> ```shell
> for id in $(seq 32)
> do
>   sed "s/<id>/job-$id/g" < job_template.json | \
>   kubectl apply -n small -f -
> done
> ```
> 
> When each job is considered individually using the `resourcequota` plugin, they will all be permitted. I don't think this should be the case, I think the plugin will need to consider other podgroups that have already been permitted in the scheduling round to prevent it from enqueueing too many podgroups and exceeding the namespace resource quota.
> 
> I have made some suggestions that would fix this issue. This assumes the order of jobs submitted is in priority order.

Thanks for your advice. The request resources of podgroups, which has already been admitted to be enqueued at current scheduling round, should be considered as used in namespace quota.

merryzhou:
> l think we also need a doc for this plugin ?

https://github.com/volcano-sh/volcano/pull/1583

Thor-wl:
Is this PR still needed? @merryzhou

k82cn:
/hold

Let's discuss this at weekly meeting firstly :)

pkg/scheduler/plugins/resourcequota/resourcequota.go
merryzhou: done

pkg/controllers/job/job_controller_actions.go
```diff
@@ -745,27 +748,29 @@ func (cc *jobcontroller) calcPGMinResources(job *batch.Job) *v1.ResourceList {
...
 				break
 			}
+
 			podCnt++
-			for _, c := range task.Template.Spec.Containers {
-				addResourceList(minAvailableTasksRes, c.Resources.Requests, c.Resources.Limits)
+			pod := &v1.Pod{
+				Spec: task.Template.Spec,
 			}
+			res, _ := quotacore.PodUsageFunc(pod, clock.RealClock{})
```
william-wang, line 768: Would you calc the minReq with volcano resource calculate functions? It's better not import too much k8s func like PodUsageFunc.

pkg/scheduler/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,79 @@
...
+			if allowed, exceeded := quotav1.LessThanOrEqual(maskedNewUsage, resourceQuota.Hard); !allowed {
+				failedRequestedUsage := quotav1.Mask(requestedUsage, exceeded)
+				failedUsed := quotav1.Mask(resourceQuota.Used, exceeded)
+				failedHard := quotav1.Mask(resourceQuota.Hard, exceeded)
+				msg := fmt.Sprintf("resource quota insufficient, requested: %v, used: %v, limited: %v",
+					failedRequestedUsage,
+					failedUsed,
+					failedHard,
+				)
+				klog.V(3).Infof("enqueueable false for job: %s/%s, because :%s", job.Namespace, job.Name, msg)
```
william-wang, line 65: suggest use log level 4 here.

regadas (commented):
This looks great!
I'm really looking forward to this. What's the timeline for this?

shinytang6:
> This looks great! I'm really looking forward to this. What's the timeline for this?

I also hope this feature will be released in v1.5, stay tuned )

pkg/scheduler/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,79 @@
...
+	return PluginName
+}
+
+func (rq *resourceQuotaPlugin) OnSessionOpen(ssn *framework.Session) {
+	pendingResources := make(map[string]v1.ResourceList)
+
+	ssn.AddJobEnqueueableFn(rq.Name(), func(obj interface{}) int {
+		job := obj.(*api.JobInfo)
+
+		resourcesRequests := job.PodGroup.Spec.MinResources
```
shinytang6, line 42: we should check if `resourcesRequests == nil` in the logic behind

pkg/scheduler/api/resource_info.go
```diff
@@ -75,6 +75,8 @@ func NewResource(rl v1.ResourceList) *Resource {
 			r.Memory += float64(rQuant.Value())
 		case v1.ResourcePods:
 			r.MaxTaskNum += int(rQuant.Value())
+		case v1.ResourceName("count/pods"):
```
shinytang6, line 78: what about some other fields contains `/` ?

pkg/scheduler/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,79 @@
...
+		if _, found := pendingResources[job.Namespace]; !found {
+			pendingResources[job.Namespace] = v1.ResourceList{}
+		}
+		pendingResources[job.Namespace] = quotav1.Add(pendingResources[job.Namespace], *resourcesRequests)
+		return util.Permit
+	})
+}
+
+func (rq *resourceQuotaPlugin) OnSessionClose(session *framework.Session) {
+}
```
shinytang6, line 79: BTW, l think we should add this plugin as default in scheduler conf :)

shinytang6 (changes requested):

pkg/controllers/job/job_controller_actions.go
merryzhou: to avoid rewrite the functions when the resourcequota logic in k8s is modified..

pkg/scheduler/api/resource_info.go
merryzhou: other fields are all ScalarResource..

pkg/scheduler/plugins/resourcequota/resourcequota.go
merryzhou: WDYT @william-wang

pkg/scheduler/plugins/resourcequota/resourcequota.go
merryzhou: done

pkg/scheduler/plugins/resourcequota/resourcequota.go
merryzhou: done

Thor-wl:
Pls rebase the latest branch and push again. @merryzhou

pkg/scheduler/api/resource_info.go
shinytang6: l mean it's strange to filter `count/pods` only, what about `count/deployments.apps` and sth like that

pkg/scheduler/plugins/resourcequota/resourcequota.go
william-wang: We can make it as default in scheduler conf after it is stable after a period of time.

pkg/scheduler/api/namespace_info.go
```diff
@@ -40,6 +40,9 @@ type NamespaceInfo struct {
 	Name NamespaceName
 	// Weight is the highest weight among many ResourceQuota.
 	Weight int64
+
+	// RQStatus stores the ResourceQuotaStatus of all ResourceQuotas in this namespace
+	RQStatus map[string]v1.ResourceQuotaStatus
```
k82cn, line 45: Let's rename `RQStatus` to `QuotaStatus` :)

pkg/scheduler/api/resource_info.go
k82cn: +1, 

AFAIK, `ResourceQuota` already support CRD, how to handle such case in Volcano?

merryzhou: > BTW, `ResourceQuota` also support PriorityClass, how to handle that?

you mean the quota scope?

pkg/scheduler/api/namespace_info.go
Thor-wl: +1
I agree with that `QuotaStatus` may be more readable.

merryzhou: done

pkg/scheduler/api/resource_info.go
merryzhou: create a new issue to track quota scope?

william-wang:
@merryzhou Overall, looks good to me. Please make the ci happy.

merryzhou:
> @merryzhou Overall, looks good to me. Please make the ci happy.

done

shinytang6 (commented):
LGTM, l think it's ready to go!
/cc @Thor-wl @william-wang

PR #1360, opened 2021-03-12
fix wrong Pipeline in action allocate

### LINKED ISSUE #1358: Maybe a mismatch in allocate action?
In action allocate, there are two approaches to allocate job resources, according to allocate idle or releasing resource. Namely

https://github.com/volcano-sh/volcano/blob/b09b8aff7ee930028e9515535baec9b2e4e51bf9/pkg/scheduler/actions/allocate/allocate.go#L232

https://github.com/volcano-sh/volcano/blob/b09b8aff7ee930028e9515535baec9b2e4e51bf9/pkg/scheduler/actions/allocate/allocate.go#L246

So is there any reason for not using `stmt.Pipeline` in the later occasion?

And will later `ssn.Pipeline` cause status mismatch when gang is not met and `stmt.Discard` is called, because the `Pipeline` operation has not been recorded in `stmt`?

yzs981130:
fix #1358


k82cn:
/lgtm
/approve

btw, please sign DCO

yzs981130:
> /lgtm
> /approve
> 
> btw, please sign DCO

Have fixed DCO. Sorry for the delay.

Issue #1363, opened 2021-03-16
it is possible to Occur OutOfCpu， when exist some pods  including init container


huone1:
in cluster ， there is a worker node：
```
Allocatable
 cpu:                6
Allocated
cpu                2950m
```
(1)  I create job including init container with 2 CPU
```
[root@h00510777-09369-kxklr huwanxing]# cat init-container.yaml
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
  labels:
    app: myapp
spec:
  nodeSelector:
    kubernetes.io/hostname: 192.168.121.209
  containers:
  - name: myapp-container
    image: busybox
    imagePullPolicy: IfNotPresent
    resources:
      requests:
        cpu: 100m
    command: ['sh', '-c', 'echo The app is running! && sleep 3600']
  initContainers:
  - name: init-myservice
    image: busybox
    imagePullPolicy: IfNotPresent
    resources:
      requests:
        cpu: 2000m
    command: ['sh', '-c', "sleep 2"]

```

check node :
cpu                4950m
![image](https://user-images.githubusercontent.com/71266853/111247318-bb1aee80-8642-11eb-83aa-a2f68aa404d7.png)

(2)   after apply  init-container.yaml , remain CPU resources:   1150m；
   apply job including init container with 1720m CPU

```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: cpu-2-1
spec:
  minAvailable: 1
  schedulerName: volcano
  Queue: default
  tasks:
    - replicas: 1
      name: "hwx"
      template:
        metadata:
          name: hello
        spec:
          containers:
            - image: busybox
              command: ["/bin/sh", "-c", "sleep 1000"]
              imagePullPolicy: IfNotPresent
              name: running
              resources:
                requests:
                  cpu: "1720m"
                  memory: "500Ki"
          restartPolicy: OnFailure

```

**What happened**:
![image](https://user-images.githubusercontent.com/71266853/111247235-94f54e80-8642-11eb-9556-09ac486ecd6a.png)
**What you expected to happen**:

the pod status should be pending

**How to reproduce it (as minimally and precisely as possible)**:
as described above
**Anything else we need to know?**:

**Environment**:
- Volcano Version:
 compile  base master 
- Kubernetes version (use `kubectl version`):
1.20.4
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
[...]

Issue #1333, opened 2021-02-25
Incorrectly calculates node resources with large requests for init containers


Robert-Christensen-visa:
**What happened**:

Volcano scheduler may schedule a job that can't fit on a node if the resource requests of the init container is larger than the resource requests of the pod.  To total resources set aside on a node by Kubernetes will be `max(max(<init container resources>), sum(<containers in pod>))`.  However, it seem that the volcano scheduler is only looking at the resource requests without init containers when determining if a pod can be placed on a node.

In the following example, `sleepy-box` has an init container that requests a lot of resources.  After the pod is running the resources consumed are very small.  The `echo-box` container has high resource requests.

When  I run `sleepy-box` followed by `echo-box` on a small one-node cluster with 8 CPUs, the pods are bound to the node even though the node does not have sufficient resources to run the node.

**What you expected to happen**:

I expect if Volcano schedules a job, the node should have sufficient resources to run the job.

**How to reproduce it (as minimally and precisely as possible)**:

``` yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: sleepy-box
spec:
  template:
    spec:
      schedulerName: volcano
      containers:
      - name: sleep
        image: busybox
        command: ['sh', '-c', 'sleep 3600']
        resources:
          limits:
            cpu: "10m"
      initContainers:
      - name: init-sleep
        image: busybox
        command: ['sh', '-c', "echo init container"]
        resources:
          limits:
            cpu: "6"
```

``` yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: echo-box
spec:
  template:
    spec:
      schedulerName: volcano
      containers:
      - name: hello
        image: busybox
        command: ['sh', '-c', 'echo done']
        resources:
          limits:
            cpu: "6"
```

Here is an example of what I got when I ran these two jobs on my dev box:

``` terminal
$ kubectl get pods
NAME               READY   STATUS     RESTARTS   AGE
echo-box-2z24c     0/1     OutOfcpu   0          73s
echo-box-bpbhr     0/1     OutOfcpu   0          24s
echo-box-m8wjk     0/1     OutOfcpu   0          63s
echo-box-mg5xp     0/1     OutOfcpu   0          74s
sleepy-box-242jj   0/1     OutOfcpu   0          69s
sleepy-box-7gxbs   0/1     OutOfcpu   0          79s
sleepy-box-mcc2z   0/1     OutOfcpu   0          79s
[...]

PR #1364, opened 2021-03-16
fix: it is possible to Occur OutOfCpu， when exist some pods including init container

huone1:
The Pod's effective request/limit:
image

issue: #1363

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -104,7 +104,7 @@ func getJobID(pod *v1.Pod) JobID {
 
 // NewTaskInfo creates new taskInfo object for a Pod
 func NewTaskInfo(pod *v1.Pod) *TaskInfo {
-	req := GetPodResourceWithoutInitContainers(pod)
+	req := GetPodResourceRequest(pod)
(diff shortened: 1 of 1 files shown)


huone1:
fix #1363

pkg/scheduler/api/job_info.go
```diff
@@ -104,7 +104,7 @@ func getJobID(pod *v1.Pod) JobID {
 
 // NewTaskInfo creates new taskInfo object for a Pod
 func NewTaskInfo(pod *v1.Pod) *TaskInfo {
-	req := GetPodResourceWithoutInitContainers(pod)
+	req := GetPodResourceRequest(pod)
```
Thor-wl, line 107: The relate issue is https://github.com/volcano-sh/volcano/issues/1333, which makes clear analysis about the the root reason. We should get the max(<init container resources>), sum(<containers in pod>).

Thor-wl: Well, this fix is right.

k82cn: /lgtm

PR #1365, opened 2021-03-16
fix sla jobOderFn when sla not set

merryzhou:


jiangkaihua:
@merryzhou In my opinion, if job A owns sla & job B do not, scheduler should schedule job A first, then job B. So the return value is set as before. If they are exchanged, job A with sla settings will be scheduled later than job B without sla settings.

merryzhou:
> @merryzhou In my opinion, if job A owns sla & job B do not, scheduler should schedule job A first, then job B. 

I agree

```
            if lJobWaitingTime == nil {
			if rJobWaitingTime == nil {
				return 0
			}
			return -1
		}
```
this's means, the leftJob not set sla, but rightJob owns sla  ==> so the rightjob has higher priority.  ==> return 1?

jiangkaihua:
> > @merryzhou In my opinion, if job A owns sla & job B do not, scheduler should schedule job A first, then job B.
> 
> I agree
> 
> ```
>             if lJobWaitingTime == nil {
> 			if rJobWaitingTime == nil {
> 				return 0
> 			}
> 			return -1
> 		}
> ```
> 
> this's means, the leftJob not set sla, but rightJob owns sla ==> so the rightjob has higher priority. ==> return 1?

You're right, I mistook the return value.

PR #1375, opened 2021-03-19
feat(webhook): add podgroup admission

shinytang6:


pkg/webhooks/admission/podgroups/mutate/mutate_podgroup.go
```diff
@@ -0,0 +1,111 @@
...
+		Allowed:   true,
+		Patch:     patchBytes,
+		PatchType: &pt,
+	}
+}
+
+func createPodGroupPatch(podgroup *schedulingv1beta1.PodGroup) ([]byte, error) {
+	var patch []patchOperation
+
+	defaultQueue := "default"
```
k82cn, line 101: is there any const from `types.go`?

shinytang6: No, I just changed it to const in that file

k82cn: If so, let's add a const for it :)

Issue #1379, opened 2021-03-24
Support set completions/backoffLimit to the task 


chenyangxueHDU:
#### What would you like to be added:
Like natvie job.Completions. As pods successfully complete, we tracks the successful completions. When a specified number of successful completions is reached, the task is complete.
Like natvie job.BackoffLimit
#### Why is this needed:
More flexible

Thor-wl:
That's interesting! Suggest to take it as a topic as the weekly meeting.

zen-xu:
We have implemented it, users can specify `minSuccess` to control when job is completed.

Thor-wl:
> I have implemented it in my company , users can specify `minSuccess` to control when job should be completed.

Great! Do you have a plan to contribute to community?

chenyangxueHDU:
> I have implemented it in my company , users can specify `minSuccess` to control when job should be completed.

你是港仔？

zen-xu:
> > I have implemented it in my company , users can specify `minSuccess` to control when job should be completed.
> 
> 你是港仔？

上海人

chenyangxueHDU:
> > > I have implemented it in my company , users can specify `minSuccess` to control when job should be completed.
> > 
> > 
> > 你是港仔？
> 
> 上海人

不好意思，我有个大学同学跟你名字一样

Thor-wl:
PR: https://github.com/volcano-sh/volcano/pull/1384

PR #1384, opened 2021-03-26
Support minSuccess

zen-xu:
ref: #1379

### DIFF
--- a/config/crd/bases/batch.volcano.sh_jobs.yaml
+++ b/config/crd/bases/batch.volcano.sh_jobs.yaml
@@ -50,6 +50,10 @@ spec:
                   to the summary of tasks' replicas
                 format: int32
                 type: integer
+              minSuccess:
+                description: The minimal success pods to run for this Job
+                format: int32
+                type: integer
               plugins:
                 additionalProperties:
                   items:
--- a/config/crd/bases/batch.volcano.sh_jobs.yaml
+++ b/config/crd/bases/batch.volcano.sh_jobs.yaml
@@ -53,6 +53,7 @@ spec:
               minSuccess:
                 description: The minimal success pods to run for this Job
                 format: int32
+                minimum: 1
                 type: integer
               plugins:
                 additionalProperties:
--- a/config/crd/v1beta1/batch.volcano.sh_jobs.yaml
+++ b/config/crd/v1beta1/batch.volcano.sh_jobs.yaml
@@ -53,6 +53,7 @@ spec:
             minSuccess:
               description: The minimal success pods to run for this Job
               format: int32
+              minimum: 1
               type: integer
             plugins:
               additionalProperties:
--- a/config/crd/v1beta1/batch.volcano.sh_jobs.yaml
+++ b/config/crd/v1beta1/batch.volcano.sh_jobs.yaml
@@ -50,6 +50,10 @@ spec:
                 to the summary of tasks' replicas
               format: int32
               type: integer
+            minSuccess:
+              description: The minimal success pods to run for this Job
+              format: int32
+              type: integer
             plugins:
               additionalProperties:
                 items:
--- a/pkg/apis/batch/v1alpha1/job.go
+++ b/pkg/apis/batch/v1alpha1/job.go
@@ -97,6 +97,10 @@ type JobSpec struct {
 	// If specified, indicates the job's priority.
 	// +optional
 	PriorityClassName string `json:"priorityClassName,omitempty" protobuf:"bytes,10,opt,name=priorityClassName"`
+
+	// The minimal success pods to run for this Job
+	// +optional
+	MinSuccess *int32 `json:"minSuccess,omitempty" protobuf:"varint,11,opt,name=minSuccess"`
 }
 
 // VolumeSpec defines the specification of Volume, e.g. PVC.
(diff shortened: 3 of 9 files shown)


config/crd/bases/batch.volcano.sh_jobs.yaml
```diff
@@ -50,6 +50,10 @@ spec:
                   to the summary of tasks' replicas
                 format: int32
                 type: integer
+              minSuccess:
```
Thor-wl, line 53: It's better to set the replicas as the default value.

config/crd/v1beta1/batch.volcano.sh_jobs.yaml
```diff
@@ -50,6 +50,10 @@ spec:
                 to the summary of tasks' replicas
               format: int32
               type: integer
+            minSuccess:
```
Thor-wl, line 53: Suggent to add a validation to this field in the admission

pkg/apis/batch/v1alpha1/job.go
```diff
@@ -97,6 +97,10 @@ type JobSpec struct {
 	// If specified, indicates the job's priority.
 	// +optional
 	PriorityClassName string `json:"priorityClassName,omitempty" protobuf:"bytes,10,opt,name=priorityClassName"`
+
+	// The minimal success pods to run for this Job
+	// +optional
+	MinSuccess *int32 `json:"minSuccess,omitempty" protobuf:"varint,11,opt,name=minSuccess"`
```
Thor-wl, line 103: Why set this field as *int , not int?

pkg/controllers/job/state/running.go
```diff
@@ -56,6 +56,13 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 				// when scale down to zero, keep the current job phase
 				return false
 			}
+
+			minSuccess := ps.job.Job.Spec.MinSuccess
```
Thor-wl, line 60: Got the reason why you set it as *int instead of int. I perfer to int and set default value to relicas. That may be better. How do you think about it?

pkg/controllers/job/state/running.go
```diff
@@ -56,6 +56,13 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
...
+			if minSuccess != nil && status.Succeeded >= *minSuccess {
+				status.State.Phase = vcbatch.Completed
+				return true
+			}
+
 			if status.Succeeded+status.Failed == jobReplicas {
 				if status.Succeeded >= ps.job.Job.Spec.MinAvailable {
 					status.State.Phase = vcbatch.Completed
```
shinytang6, line 68: If we use `minSuccess`, do we still need `MinAvailable` to determine job status?

pkg/apis/batch/v1alpha1/job.go
zen-xu: I still think this field should be *int. 

`MinSuccess` is an **optional** feature, If the user does not set this field, the logic should same as the original.

pkg/controllers/job/state/running.go
zen-xu: I still think this field should be *int. 

`MinSuccess` is an **optional** feature, If the user does not set this field, the logic should same as the original.

pkg/controllers/job/state/running.go
zen-xu: We can use `MinAvailable` to determine job status when users don't want to use `MinSuccess`

shinytang6: yes, l agree. So maybe we can add another layer of logic inside `if status.Succeeded+status.Failed == jobReplicas ` ? If `minSuccess` is specified, we use `minSuccess` instead of `minAvailable`

zen-xu: Yes

config/crd/v1beta1/batch.volcano.sh_jobs.yaml
zen-xu: done

pkg/controllers/job/state/running.go
Thor-wl: > I still think this field should be *int.
> 
> `MinSuccess` is an **optional** feature, If the user has not set this field, the logic should same as the original.

Yes, it's reasonable.

Thor-wl:
related issue: https://github.com/volcano-sh/volcano/issues/1379

zen-xu:
@Thor-wl  needs approved label

pkg/controllers/job/state/running.go
zen-xu: If `minSuccess` has matched，Job will mark `Completed`. So it's not necessary to handle the next compare logic.

shinytang6: l mean if `minSuccess` is specified while `status.Succeeded < *minSuccess` all the time, then after all the pods completed, we should also replace `if status.Succeeded >= ps.job.Job.Spec.MinAvailable` with `if status.Succeeded >= minSuccess`. WDYT?

zen-xu: Agree

shinytang6 (approved):

Thor-wl:
There seems to be lack of doc design about this feature? @zen-xu

zen-xu:
A little busy these weeks, I will complete it later.

Thor-wl:
> A little busy these weeks, I will complete it later.

/assign @zen-xu

PR #1388, opened 2021-03-30
add `bindingTasks` to judge whether adding node to the snapshot.

### LINKED ISSUE #1383: SchedulerCache Bind use goroutine to bind may cause OutOfcpu/OutOfMemory
**What happened**:
https://github.com/volcano-sh/volcano/blob/173f3fa3de996a7e0a7bf0846fd00d75f4eb2fbb/pkg/scheduler/cache/cache.go#L558-L565

We found bind duration may take more than **1s** while cluster in high load, so in the next session cache is wrong.

image

zen-xu:
If the node has some binding tasks, current session skip this node.

fix #1383

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -417,8 +417,8 @@ func (ni *NodeInfo) SubGPUResource(pod *v1.Pod) {
 func (ni *NodeInfo) GetBindingTasks() []TaskID {
 	var tasks []TaskID
 
-	for taskId := range ni.bindingTasks {
-		tasks = append(tasks, taskId)
+	for taskID := range ni.bindingTasks {
+		tasks = append(tasks, taskID)
 	}
 
 	return tasks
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/node_info.go
```diff
@@ -52,6 +52,8 @@ type NodeInfo struct {
 	// Used to store custom information
 	Others     map[string]interface{}
 	GPUDevices map[int]*GPUDevice
+
+	bindingTasks map[TaskID]string
```
k82cn, line 56: It's better to have a `Binding` resources, similar to `Releasing` above :)

k82cn: Any feedback on this comment?

Thor-wl: /cc @zen-xu

zen-xu: As it's hard to judge the task state of allocating failed, `Releasing` resource is hard to implement by now.

PR #1391, opened 2021-04-01
Support scheduler to select nodes

zen-xu:
implement: #1397

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1,17 +1,17 @@
 /*
-Copyright 2017 The Kubernetes Authors.
+ Copyright 2021 The Volcano Authors.
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
 */
 
 package cache
@@ -19,6 +19,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -26,6 +27,7 @@ import (
 	"k8s.io/api/scheduling/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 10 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -679,11 +684,21 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 		RevocableNodes: make(map[string]*schedulingapi.NodeInfo),
 	}
 
+	// If nodeSelector == "", will match all nodes
+	selector, err := labels.Parse(strings.TrimSpace(sc.nodeSelector))
+	if err != nil {
+		panic(fmt.Errorf("the provided selector %q is not valid: %v", sc.nodeSelector, err))
+	}
+
```
shinytang6, line 692: why not register this block to `nodeInformer` filter event handler >?

pkg/scheduler/scheduler.go
```diff
@@ -67,10 +68,17 @@ func NewScheduler(
 		}
 	}
 
+	// If nodeSelector == "", will match all nodes
```
shinytang6, line 71: pls remove this block :)

shinytang6 (changes requested):

pkg/scheduler/cache/cache.go
zen-xu: Because we must sync all nodes, otherwise the cache may go wrong when adding nodes existing nodes to this scheduler.

We have met this problem in my company.

zen-xu: I think we should sync all cluster nodes, which will more efficient to add nodes to the current scheduler. 

And the current scheduler should still take care of others nodes which hold some pods. 

If we do filter in `nodeInformer`, we will get the wrong node idle resources.

Thor-wl:
Please add the corresonding issue or scenario.

Thor-wl:
@william-wang @wpeng102 @shinytang6 Please review this doc design
@zen-xu Is there any process about this feature?

zen-xu:
I think #1576 has covered my pr. So I am ok to close this pr

Thor-wl:
> I think #1576 has covered my pr. So I am ok to close it

That PR cannot select nodes for volcano cache.

zen-xu:
> > I think #1576 has covered my pr. So I am ok to close it
> 
> That PR cannot select nodes for volcano cache.

I haven't tried this feature yet.
But as what the [doc](https://github.com/volcano-sh/volcano/blob/master/docs/design/multi-scheduler.md) said, after I specify the following config, the job which has labelled `node-zone: zone1` will only be scheduled on the **zone1 nodes** , Right?

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: volcano-admission-configmap
  namespace: volcano-system
data:
  volcano-admission.conf: |
    resourceGroups:
    - resourceGroup: zone1                   
      object:
        key: label                       
        value:
        - "node-zone: zone1"
      schedulerName:  zone1-scheduler      
      labels:
        node-zone: zone1
    - resourceGroup: zone2         
      object:
        key: label                       
        value:
        - "node-zone: zone2"
      schedulerName:  zone2-scheduler      
      labels:
        node-zone: zone2
```

Thor-wl:
> > > I think #1576 has covered my pr. So I am ok to close it
> > 
> > 
> > That PR cannot select nodes for volcano cache.
> 
> I haven't tried this feature yet.
> But as what the [doc](https://github.com/volcano-sh/volcano/blob/master/docs/design/multi-scheduler.md) said, after I specify the following config, the job which has labelled `node-zone: zone1` will only be scheduled on the **zone1 nodes** , Right?
> 
> ```yaml
> apiVersion: v1
> kind: ConfigMap
> metadata:
>   name: volcano-admission-configmap
>   namespace: volcano-system
> data:
>   volcano-admission.conf: |
>     resourceGroups:
>     - resourceGroup: zone1                   
>       object:
>         key: label                       
>         value:
>         - "node-zone: zone1"
>       schedulerName:  zone1-scheduler      
>       labels:
>         node-zone: zone1
>     - resourceGroup: zone2         
>       object:
>         key: label                       
>         value:
>         - "node-zone: zone2"
>       schedulerName:  zone2-scheduler      
>       labels:
>         node-zone: zone2
> ```

Yes, but the cache still contains all nodes in cluster, which is not reasonable.

zen-xu:
Keep all node in each scheduler cache is reasonable.
Because sometimes we may move a node [ *node-0* ]  from `zone1` to `zone2`.
If scheduler cache only keep `zone1` nodes, `zone1` scheduler informer will re-sync [ *node-0* ] which will take a long time and may cause `UB` ( *node-0* has not synced yet, but scheduler want to allocate a pod to it )

Thor-wl:
> Keep all node in each scheduler cache is reasonable.
> Because sometimes we may move a node [ _node-0_ ] from `zone1` to `zone2`.
> If scheduler cache only keep `zone1` nodes, `zone1` scheduler informer will re-sync [ _node-0_ ] which will take a long time and may cause `UB` ( _node-0_ has not synced yet, but `zone1` scheduler wants to allocate a pod to it )

If all nodes cached, It will bring another problem: queue resource calculation. In fact, only part of nodes are mapped to volcano. But deserved resource of queue is calculated based on all resource of cluster.

zen-xu:
Haha, I agree

Thor-wl:
> Haha, I agree
> But my `pr` still keep all cache, but snapshot will use selected nodes

Got it. Let me try another way.

PR #1401, opened 2021-04-06
update dependency to volcano.sh/api

Thor-wl:


vendor/volcano.sh/apis/pkg/apis/batch/v1alpha1/job.go
```diff
@@ -29,16 +29,6 @@ import (
...
-// +kubebuilder:printcolumn:name="Pending",type=integer,JSONPath=`.status.pending`
-// +kubebuilder:printcolumn:name="Running",type=integer,JSONPath=`.status.running`
-// +kubebuilder:printcolumn:name="Succeeded",type=integer,JSONPath=`.status.succeeded`
-// +kubebuilder:printcolumn:name="Failed",type=integer,JSONPath=`.status.failed`
-// +kubebuilder:printcolumn:name="Terminating",type=integer,JSONPath=`.status.terminating`
-// +kubebuilder:printcolumn:name="Unknown",type=integer,JSONPath=`.status.unknown`
-// +kubebuilder:printcolumn:name="RetryCount",type=integer,JSONPath=`.status.retryCount`
-// +kubebuilder:printcolumn:name="MinAvailable",type=integer,JSONPath=`.status.minAvailable`
-// +kubebuilder:printcolumn:name="State",type=string,JSONPath=`.status.state.phase`
-// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```
shinytang6, line 41: These annotations are newly added features, we can update to volcano/apis :)

vendor/modules.txt
```diff
@@ -715,3 +743,4 @@ sigs.k8s.io/yaml
 # k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.18.3
 # k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.18.3
 # k8s.io/sample-controller => k8s.io/sample-controller v0.18.3
+# volcano.sh/apis => volcano.sh/apis v1.3.0-k8s1.18.3-alpha.3
```
shinytang6, line 746: could we remove the vendor directory, l think it is not necessary if we use go mod?

Thor-wl: > could we remove the vendor directory, l think it is not necessary if we use go mod?

Of course.

zen-xu:
What's the benefit of splitting `api` into another repo.

Thor-wl:
> What's the benefit of splitting `api` into another repo.

Downstream community such as kubeflow can get CRD clientset generated by certain K8s version. It is separated with the main repository.

vendor/modules.txt
william-wang: we'd better keep vendor. It is convient and no need download dependent package everytime.

config/crd/bases/batch.volcano.sh_jobs.yaml
```diff
@@ -19,38 +19,7 @@ spec:
     singular: job
   scope: Namespaced
   versions:
-  - additionalPrinterColumns:
```
william-wang, line 22: These changes is commited to volcano.sh/volcano/pkg/api/ after you move api to volcano/api/.  Need to update the chagnes to volcano.sh/api master first and then refresh this pr.

william-wang (changes requested):

Issue #1397, opened 2021-04-06
Support multi schedulers


zen-xu:
#### What would you like to be added:

Support scheduler to select nodes by labels for scheduling

#### Why is this needed:

Our cluster need to be divided into multi sections. 

Example:
1. section A for common scheduling.
2. section B for GPU scheduling.
3. section C for daily job.
4. etc

Thor-wl:
Well, we need a design doc for that.

Thor-wl:
refer：https://github.com/volcano-sh/volcano/pull/1576
refer：https://github.com/volcano-sh/volcano/pull/1565

PR #1403, opened 2021-04-07
doc: add multi schedulers design doc

zen-xu:
ref: #1397 #1391


Thor-wl:
I'll take part in the design and development of this feature. Let's work together. @zen-xu

zen-xu:
> I'll take part in the design and development of this feature. Let's work together. @zen-xu

👌

docs/design/multi-schedulers.md
```diff
@@ -0,0 +1,105 @@
+# Multi schedulers
+
+## Introduction
+
+Sometimes we may want to the cluster be divided into multi sections for different usage.
```
Thor-wl, line 5: As the feedback from the users, another scenario is to schedule workloads under ns `kube-system`  by default-scheduler and others by Volcano. So i suggest to take it into consideration, too.

Thor-wl:
I'll merge this doc and continue to complete it.

PR #1409, opened 2021-04-10
fix CI

Thor-wl:


Thor-wl:
I've taken an investigation last weekend about the frequent CI timeout. It's not caused by the logic of E2E  but the instability of machines of github action.  So i analysesed the fauilre cases and picked out the cases of high frequency.

william-wang (approved):
/lgtm

Issue #1417, opened 2021-04-13
there is a OutOfcpu when applying job yaml if  node sync after pod sync in scheduler deployment 


huone1:
### env
![image](https://user-images.githubusercontent.com/71266853/114528579-b9f9d300-9c7b-11eb-9d49-31df9e8bbbd7.png)

### test

```
func (sc *SchedulerCache) Run(stopCh <-chan struct{}) {
     go sc.podInformer.Informer().Run(stopCh)
     time.Sleep(10 * time.Second)  // add sleep 
     go sc.nodeInformer.Informer().Run(stopCh)
}
```

1. kubectl apply -f job-1.yaml  
2. replace scheduler image to image including the modification, restart the scheduler pod
   check the used resource info:
![image](https://user-images.githubusercontent.com/71266853/114528710-d4cc4780-9c7b-11eb-96e4-b70df53417b2.png)
3. kubectl apply -f job-2.yaml  
    there is a OutOfcpu 
![image](https://user-images.githubusercontent.com/71266853/114528949-0f35e480-9c7c-11eb-83e3-f26abfe53ad4.png)

**job-1 (2  pod  2 core/pod)**
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: job-1
spec:
  schedulerName: volcano
  minAvailable: 2
  tasks:
    - replicas: 2
      name: "test"
      template:
        spec:
          containers:
            - image: alpine
              command: ["/bin/sh", "-c", "sleep 10000"]
              imagePullPolicy: IfNotPresent
              name: running
              resources:
                limits:
                  memory: "2000Mi"
                  cpu: "2"
          restartPolicy: OnFailure
```

**job-2 (1  pod  4 core/pod)**

```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: job-2
spec:
  schedulerName: volcano
  minAvailable: 1
  tasks:
    - replicas: 1
      name: "test"
      template:
        spec:
          containers:
            - image: alpine
              command: ["/bin/sh", "-c", "sleep 10000"]
              imagePullPolicy: IfNotPresent
              name: running
              resources:
                limits:
                  memory: "2000Mi"
                  cpu: "4"
          restartPolicy: OnFailure
```

**What happened**:
there is a OutOfcpu when applying job yaml if  node sync after pod sync in scheduler deployment 

**What you expected to happen**:
the created pod is pending

[...]

PR #1418, opened 2021-04-13
fix: create nodeinfo when add task to the nodeinfo and the nodeinfo is not existed

huone1:
fix issue #1417


pkg/scheduler/cache/event_handlers.go
```diff
@@ -59,7 +59,11 @@ func (sc *SchedulerCache) getOrCreateJob(pi *schedulingapi.TaskInfo) *scheduling
 func (sc *SchedulerCache) addTask(pi *schedulingapi.TaskInfo) error {
 	if len(pi.NodeName) != 0 {
 		if _, found := sc.Nodes[pi.NodeName]; !found {
-			return fmt.Errorf("node <%s> does not exist", pi.NodeName)
+			if sc.nodeInformer.Informer().HasSynced() {
```
k82cn, line 62: AFAIK, we already sync all informers when starting scheduelr.

huone1: The time sequence of pods and node resources cannot be guaranteed.

huone1: the problem is happened in the starting scheduler period to sync informers data

huone1: I deleted the judgment."if sc.nodeInformer.Informer().HasSynced" , and  created the nodeinfo structure directly.

k82cn: what's the impact if there's no node object in node info?

huone1: if there's no node object in node info, the nodeinfo status will be "NotReady" util the imformer receives the node object 
 to update the status to "Ready". So it is no influence to the volcano schduling and just saves the task info in its nodeInfo.

k82cn: that's great!

pkg/scheduler/api/node_info.go
```diff
@@ -122,6 +122,10 @@ func (ni *NodeInfo) Ready() bool {
 }
 
 func (ni *NodeInfo) setRevocableZone(node *v1.Node) {
+	if node == nil {
+		return
```
k82cn, line 126: let's log a message here, e.g. warning. BTW, will scheduler re-update revocable-zone after that?

huone1: I checked it ， and I found it drop the setRevocableZone in function SetNode and I will add it。
And I add some log in function setNodeGPUInfo, setNodeState, setRevocableZone

huone1:
Close it because the  branch has been confused due to  the merge operation error 
And reopen #1662

PR #1426, opened 2021-04-20
manual code update to match k8s v1.19

Thor-wl:
This is the manual code update to match k8sv1.19

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,53 +3,53 @@ module volcano.sh/volcano
 go 1.14
 
 require (
-	github.com/fsnotify/fsnotify v1.4.7
+	github.com/fsnotify/fsnotify v1.4.9
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/onsi/ginkgo v1.11.0
 	github.com/onsi/gomega v1.7.0
-	github.com/prometheus/client_golang v1.0.0
-	github.com/spf13/cobra v0.0.5
+	github.com/prometheus/client_golang v1.7.1
+	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/crypto v0.0.0-20200220183623-bac4c82f6975
-	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
+	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
+	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 	gopkg.in/yaml.v2 v2.2.8
-	k8s.io/api v0.18.3
-	k8s.io/apimachinery v0.18.3
-	k8s.io/client-go v0.18.3
-	k8s.io/component-base v0.18.3
+	k8s.io/api v0.19.6
+	k8s.io/apimachinery v0.19.6
+	k8s.io/client-go v0.19.6
+	k8s.io/component-base v0.19.6
 	k8s.io/klog v1.0.0
 	k8s.io/kube-scheduler v0.0.0
-	k8s.io/kubernetes v1.18.3
+	k8s.io/kubernetes v1.19.6
 	sigs.k8s.io/yaml v1.2.0
-	volcano.sh/apis v1.3.0-k8s1.18.3-alpha.3
+	volcano.sh/apis v1.2.0-k8s1.19.6-alpha.1
 )
 
 replace (
-	k8s.io/api => k8s.io/api v0.18.3
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.18.3
-	k8s.io/apimachinery => k8s.io/apimachinery v0.18.3
-	k8s.io/apiserver => k8s.io/apiserver v0.18.3
-	k8s.io/cli-runtime => k8s.io/cli-runtime v0.18.3
-	k8s.io/client-go => k8s.io/client-go v0.18.3
-	k8s.io/cloud-provider => k8s.io/cloud-provider v0.18.3
-	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.18.3
-	k8s.io/code-generator => k8s.io/code-generator v0.18.3
-	k8s.io/component-base => k8s.io/component-base v0.18.3
-	k8s.io/cri-api => k8s.io/cri-api v0.18.3
-	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.18.3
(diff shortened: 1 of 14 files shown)


go.mod
```diff
@@ -3,53 +3,53 @@ module volcano.sh/volcano
...
+	k8s.io/apimachinery v0.19.6
+	k8s.io/client-go v0.19.6
+	k8s.io/component-base v0.19.6
 	k8s.io/klog v1.0.0
 	k8s.io/kube-scheduler v0.0.0
-	k8s.io/kubernetes v1.18.3
+	k8s.io/kubernetes v1.19.6
 	sigs.k8s.io/yaml v1.2.0
-	volcano.sh/apis v1.3.0-k8s1.18.3-alpha.3
+	volcano.sh/apis v1.2.0-k8s1.19.6-alpha.1
```
william-wang, line 25: Volcano.sh/volcano should require the tag-commit from volcano/api master branch, other than release-1.19 branch.

william-wang (changes requested):

go.mod
Thor-wl: Let me release a normal version

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -272,3 +280,38 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
+	}, fh)
+	if err != nil {
+		klog.V(3).Infof("new volumebinding instance failed for %v", err.Error())
+		return nil
+	}
+
+	p := pl.(*volumebinding.VolumeBinding)
+	var podVolumes *volumescheduling.PodVolumes
+	boundClaims, claimsToBind, _, err := p.Binder.GetPodVolumes(pod)
+	if err != nil {
```
wpeng102, line 307: check info log is enough to handle the err, should we need return the err?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -272,3 +280,38 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
+		klog.V(3).Infof("Get podVolumes for pod %v on node %v failed for %v", pod.Name, node.Name, err.Error())
+		return podVolumes
+	}
+	podVolumes, _, err = p.Binder.FindPodVolumes(pod, boundClaims, claimsToBind, node)
+	if err != nil {
```
wpeng102, line 312: check info log is enough to handle the err, should we need return the err?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -272,3 +280,38 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
+		klog.V(3).Infof("Find podVolumes for pod %v on node %v failed for %v", pod.Name, node.Name, err.Error())
+		return podVolumes
```
wpeng102, line 314: no need return podVolumes in `if` scope, because it just print a log and beyond `if` it return podVolumes finally.

pkg/scheduler/framework/session.go
```diff
@@ -44,11 +45,12 @@ type Session struct {
 	// This should not be mutated after initiated
 	podGroupStatus map[api.JobID]scheduling.PodGroupStatus
 
-	Jobs           map[api.JobID]*api.JobInfo
-	Nodes          map[string]*api.NodeInfo
+	Jobs  map[api.JobID]*api.JobInfo
+	Nodes map[string]*api.NodeInfo
```
william-wang, line 49: the alignment is inconsistent.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -193,7 +195,7 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 			fmt.Errorf("failed to predicates, node info for %s not found", node.Name)
 		}
 
-		if node.Allocatable.MaxTaskNum <= len(nodeInfo.Pods()) {
```
william-wang, line 196: why not use Pod() as before.

pkg/scheduler/plugins/util/k8s/snapshot.go
```diff
@@ -25,39 +25,39 @@ import (
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
-	schedulerlisters "k8s.io/kubernetes/pkg/scheduler/listers"
-	schedulernodeinfo "k8s.io/kubernetes/pkg/scheduler/nodeinfo"
```
william-wang, line 29: the lister and nodeinfo is not available in 1.19?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -272,3 +280,38 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 }
 
 func (alloc *Action) UnInitialize() {}
+
+func setPodVolumesByNode(pod *v1.Pod, node *v1.Node) *volumescheduling.PodVolumes {
+	client := fake.NewSimpleClientset()
```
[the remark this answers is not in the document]
Thor-wl: Yes, i'll repleace it with ssn.kubeclient

pkg/scheduler/actions/allocate/allocate.go
Thor-wl: if err is not nil, the podVolumes is nil, It can be the flag to judge if some operation fails.

pkg/scheduler/actions/allocate/allocate.go
Thor-wl: That's right. Let me optimize it.

pkg/scheduler/actions/allocate/allocate.go
Thor-wl: ditto

pkg/scheduler/framework/session.go
Thor-wl: It's the result after go fmt.

pkg/scheduler/plugins/predicates/predicates.go
Thor-wl: The struct nodeInfo is a map and the key is Pods

pkg/scheduler/plugins/util/k8s/snapshot.go
Thor-wl: Yes, they are removed

pkg/scheduler/actions/allocate/allocate.go
Thor-wl: done

PR #1429, opened 2021-04-21
fix(controller): fix ssh missing when job update

### LINKED ISSUE #1420: I lost ssh files when I updated my job(<pluginOnJobUpdate>: ssh on jo)
**What happened**:
pkg/controllers/job/plugins/ssh/ssh.go
```golang
// TODO: currently a container using a Secret as a subPath volume mount will not receive Secret updates.
func (sp *sshPlugin) OnJobUpdate(job *batch.Job) error {
	data, err := generateRsaKey(job)
	if err != nil {
		return err
	}
	
	if err := helpers.CreateOrUpdateSecret(job, sp.client.KubeClients, data, sp.secretName(job)); err != nil {
		return fmt.Errorf("update secret for job <%s/%s> with ssh plugin failed for %v",
			job.Namespace, job.Name, err)
	}

	return nil
}
```
```golang
	for i, c := range pod.Spec.Containers {
		vm := v1.VolumeMount{
			MountPath: sp.sshKeyFilePath,
			SubPath:   SSHRelativePath,
			Name:      secretName,
		}

		pod.Spec.Containers[i].VolumeMounts = append(c.VolumeMounts, vm)
	}
```
[Mounted Secrets are updated automatically](https://kubernetes.io/docs/concepts/configuration/secret/)
Note: A container using a Secret as a **subPath** volume mount will not receive Secret updates.

-------

when run mpi job,such as horovod elastic job, when i delete pod or update job... ,it's while `Starting to execute plugin at <pluginOnJobUpdate>: ssh on job`,and the ssh secret mount with subPath, when update secret, the ssh file while be remove. This caused the ssh communication to fail and the training exited

`Why do we have to update the ssh secret every time we update the job`

I have to delete this code.
```golang
func (sp *sshPlugin) OnJobUpdate(job *batch.Job) error {

	return nil
}
```

**Environment**:
- Volcano Version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shinytang6:
resolves #1420

### DIFF
--- a/pkg/controllers/job/plugins/ssh/ssh.go
+++ b/pkg/controllers/job/plugins/ssh/ssh.go
@@ -100,16 +100,18 @@ func (sp *sshPlugin) OnJobDelete(job *batch.Job) error {
 }
 
 // TODO: currently a container using a Secret as a subPath volume mount will not receive Secret updates.
+// we may not update the job secret due to the above reason now.
+// related issue: https://github.com/volcano-sh/volcano/issues/1420
 func (sp *sshPlugin) OnJobUpdate(job *batch.Job) error {
-	data, err := generateRsaKey(job)
-	if err != nil {
-		return err
-	}
-
-	if err := helpers.CreateOrUpdateSecret(job, sp.client.KubeClients, data, sp.secretName(job)); err != nil {
-		return fmt.Errorf("update secret for job <%s/%s> with ssh plugin failed for %v",
-			job.Namespace, job.Name, err)
-	}
+	//data, err := generateRsaKey(job)
+	//if err != nil {
+	//	return err
+	//}
+	//
+	//if err := helpers.CreateOrUpdateSecret(job, sp.client.KubeClients, data, sp.secretName(job)); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/job/plugins/ssh/ssh.go
```diff
@@ -100,16 +100,18 @@ func (sp *sshPlugin) OnJobDelete(job *batch.Job) error {
...
-	data, err := generateRsaKey(job)
-	if err != nil {
-		return err
-	}
-
-	if err := helpers.CreateOrUpdateSecret(job, sp.client.KubeClients, data, sp.secretName(job)); err != nil {
-		return fmt.Errorf("update secret for job <%s/%s> with ssh plugin failed for %v",
-			job.Namespace, job.Name, err)
-	}
+	//data, err := generateRsaKey(job)
```
Thor-wl, line 106: Have you tested locally?

shinytang6: yes, l have tested. These blocks will cause the ssh key to be lost when the job updated, it's better to comment out now。

PR #1431, opened 2021-04-22
Fix bug when action config is not found

sniperking1234:
Signed-off-by: chen zhengwei <chenzhengwei@inspur.com>
I think when action is not found, it should continue rather than return an error.


pkg/scheduler/util.go
```diff
@@ -76,7 +77,7 @@ func unmarshalSchedulerConf(confStr string) ([]framework.Action, []conf.Tier, []
 		if action, found := framework.GetAction(strings.TrimSpace(actionName)); found {
 			actions = append(actions, action)
 		} else {
-			return nil, nil, nil, fmt.Errorf("failed to found Action %s, ignore it", actionName)
+			klog.Errorf("failed to found Action %s, ignore it", actionName)
```
Thor-wl, line 80: If no action configured, scheduler will make no sense for no scheduling process. Enqueue and Allocate is necessary.

sniperking1234: There is an action's array loop outside the code. So if only one action config is not found, should all configurations be discarded?

Thor-wl: Yes, i think it is reasonable. Even if one invalid action configuration is not allowed.

sniperking1234: @Thor-wl Thanks for the explanation. Maybe the error message need to be more accurate.

PR #1437, opened 2021-04-25
submit the guide files about numa-aware

huone1:
Signed-off-by: h00510777 <huwanxing@huawei.com>


docs/user-guide/how_to_use_numa_aware.md
```diff
@@ -0,0 +1,158 @@
+# NUMA Aware User Guide
+
+## Environment setup
+
+### Pre-Condition
+
+```
+1. Enable cpu manager and set policy to "static"
```
shinytang6, line 8: Just curious, is setting cpu manager policy to "static" necessary ?

huone1: Because the cpu manager on kubelet handle  the cpu assign action  and  the shceduler just determines which node can be scheduled , but does not determine cpu allocation.
So  for exclusiving  CPUs , set cpu manager policy to "static"

shinytang6: Yes, l know what you mean.
My question is l think topology manager will still work if cpu manager policy is not static(don't want exclusiving CPUs), is it right?

huone1: topology manager will still work for others resources，but CPU resource Limits  just processes  by CFS in linux kerne if the cpu manager policy is  None.

shinytang6: So `Enable cpu manager and set policy to "static"` is optional for numa awareness. It’s just that in some cpu-intensive scenarios, it’s better to set cpu manager policy to static.

docs/design/numa-aware.md
```diff
@@ -0,0 +1,114 @@
...
+  4. none
+```
+
+### Predicate function
+
+for the pods with the topology policy, we need to predicate the matched node list
+
+| policy | action |
+| :----   | :---- |
+| none   | 1. no filter ction |
```
Thor-wl, line 73: %s/ction/action/

docs/design/numa-aware.md
```diff
@@ -0,0 +1,114 @@
...
+| restricted | 1.	filter out the node with the topology policy “restricted” <br> 2.	filter out the node that the cpu topology meets the cpu requirements for "restricted" | 
+| single-numa-node | 1.	filter out the node with the topology policy “single-numa-node”;  <br> 2. filter out the node that the cpu topology meets the cpu requirements for "single-numa-node" |
+
+### Priority function
+
+Regardless of the topology policy, pod is hoped to be scheduled to the optimal node at that time. <br>
+So we select the best node by scoring all filtered nodes
+```
+calculation formula:
+     score = weight * (100 - 100 * numaNodeNum / maxNumaNodeNum)
```
Thor-wl, line 84: add comments about `numaNodeNum` and `maxNumaNodeNumber`

Thor-wl: What's the default weight and how to configure it?

docs/design/numa-aware.md
```diff
@@ -0,0 +1,114 @@
...
+![](./images/numa-aware-process.png) 
+
+
+### Pod-level topology
+In the volcano job, it sets the different policies config for the specific task.
+```
+task:
+  - replicas: 1
+    name: "test-1"
+    topologyPolicy: single-numa-node
```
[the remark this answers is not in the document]
huone1: Chapter  【the Pod-level topology 】has commented all topogyPolicy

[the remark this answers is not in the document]
huone1: Chapter 【the Pod-level topology 】has commented all topogyPolicy

docs/design/numa-aware.md
huone1: added the arguments description

docs/design/numa-aware.md
huone1: fixed it

PR #1447, opened 2021-04-28
update volcano-development.yaml to v1

huone1:
Signed-off-by: h00510777 <huwanxing@huawei.com>


Thor-wl:
ref: https://github.com/volcano-sh/volcano/pull/1483

PR #1450, opened 2021-05-03
same job's diffenrent tasks should compare task's priority

lowang-bh:
priority plugin's preemptableFn has 2 conditions to preempt: Preemption between Jobs and Preemption between Task within Job; so same job's diffenrent tasks should compare task's priority


Thor-wl:
I'll close the PR because of the repeation https://github.com/volcano-sh/volcano/pull/1450

PR #1453, opened 2021-05-06
Hotfix issue 1451: lose preemptor when considering Preemption between…

lowang-bh:
… Tasks within same Job

Signed-off-by: lowang_bh <lhui_wang@163.com>

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -144,6 +144,11 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 		// Preemption between Task within Job.
 		for _, job := range underRequest {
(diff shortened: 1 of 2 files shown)


shinytang6 (commented):
I have a question, after preemption between jobs within queue, some jobs are already running, no longer need to preempt, so should we recalculate the `underRequest` before preemption between task within job?

lowang-bh:
> I have a question, after preemption between jobs within queue, some jobs are already running, no longer need to preempt, so should we recalculate the `underRequest` before preemption between task within job?

I think it doesn't need any more.  The code shows `preemptorTasks` will be empty and break out of the for loop if it met this  situation

shinytang6 (approved):
l am ok with the change.
/lgtm

Thor-wl (approved):

PR #1459, opened 2021-05-10
feat: support minAvailable at task level

shinytang6:


Thor-wl:
Build failed?

shinytang6:
> Build failed?

l will update the dependencies once https://github.com/volcano-sh/apis/pull/40 is merged, please review the code first.

pkg/controllers/job/job_controller_actions.go
```diff
@@ -489,6 +498,16 @@ func (cc *jobcontroller) createOrUpdatePodGroup(job *batch.Job) error {
 				job.Namespace, job.Name, err)
 			return err
 		}
+
+		minTaskMember := map[string]int32{}
+		for _, task := range job.Spec.Tasks {
+			if task.MinAvailable != nil {
+				minTaskMember[task.Name] = *task.MinAvailable
+			} else {
+				minTaskMember[task.Name] = task.Replicas
```
zen-xu, line 507: Will it conflict with `Job.Minabailable = 1`

shinytang6: What do you mean by `conflict` here, can you elaborate more? Normally, the task.minavailable here should not be empty because it is filled in the webhook

zen-xu:
```
Job:
  minAvailable: 1
  tasks:
  - task1
       replicas: 10
  - task2
       replicas: 3
```

If I specify the `vj` which task.minavailable is omitted, the original logic job will be scheduled if one pod minresouce meet.
In this pr, job will never be scheduled until total 13 pods resources can be allocated, right?

zen-xu:
### One question:

If I have a specification like this

```
Job:
  tasks:
  - task1:
       replicas: 10
       minAvailable: 2
       requests:
          cpu: 3
  - task2:
       replicas: 10
       minAvailable: 2
       requests:
          cpu: 3
```

And exists a node which has 12 cpu

Will this pr promise that each task has 2 pods being scheduled rather than only task1 has 4 pods being scheduled?

pkg/controllers/job/job_controller_actions.go
shinytang6: No, I have judged  sumof(task minavailable) and job.minavailable, if job.minavailable<sumof(task minavailable), we will ignore task minavailable.
See:
```
// CheckTaskMinAvailable returns whether each task of job is valid.
func (ji *JobInfo) CheckTaskMinAvailable() bool {
	// if job minAvailable is less than sumof(task minAvailable), skip this check
	if ji.MinAvailable < ji.TaskMinAvailableTotal {
		return true
	}
        ...
}
```

shinytang6:
> ### One question:
> If I have a specification like this
> 
> ```
> Job:
>   tasks:
>   - task1:
>        replicas: 10
>        minAvailable: 2
>        requests:
>           cpu: 3
>   - task2:
>        replicas: 10
>        minAvailable: 2
>        requests:
>           cpu: 3
> ```
> 
> And exists a node which has 12 cpu
> 
> Will this pr promise that each task has 2 pods being scheduled rather than only task1 has 4 pods being scheduled?

currently not, I think it is a bit difficult to achieve because we actually don’t know the state of the cluster when we choose which task to schedule, so it is difficult to make the best choice. WDYT, any advice?

zen-xu:
> currently not, I think it is a bit difficult to achieve because we actually don’t know the state of the cluster when we choose which task to schedule, so it is difficult to make the best choice. WDYT, any advice?

Maybe we can use `max(0, task.minAvailable - scheduledNum) / task.minAvailable` as priority to determine which task will be the candidate

Thor-wl:
Any process about this feature?

shinytang6:
> Any process about this feature?

A bit busy these days, will resolve the CI error this week.
@william-wang @Thor-wl Please help review the code first.

config/crd/bases/batch.volcano.sh_jobs.yaml
```diff
@@ -115,6 +115,11 @@ spec:
                 items:
                   description: TaskSpec specifies the task specification of Job.
                   properties:
+                    minAvailable:
+                      description: The minimal available pods to run for this Task
+                        Defaults to the task replicas
```
Thor-wl, line 120: If the default value is task replicas, does it mean that the minAvailable to the job is the sum of the minAvailable for all tasks?

config/crd/bases/batch.volcano.sh_jobs.yaml
```diff
@@ -6857,6 +6862,9 @@ spec:
                           - containers
                           type: object
                       type: object
+                    topologyPolicy:
```
Thor-wl, line 6865: suggest to list all available policy in the describtion

config/crd/bases/batch.volcano.sh_jobs.yaml
shinytang6: yes

config/crd/bases/batch.volcano.sh_jobs.yaml
shinytang6: these files are auto generated by scripts

config/crd/bases/batch.volcano.sh_jobs.yaml
Thor-wl: > yes

So if a user doesn't set task minAvailable but set job minavailable, which keeps the original user habit, he may wants to get pods at the number of job minAvailable, but he can't get the result for the sum of task replicases is larger than job minAvailable.

shinytang6: Sorry I did not explain clearly, if job minavailable is not set, it will be set to sumof(task minavailable), otherwise it will not be changed and keep user settings.

Thor-wl: > Sorry I did not explain clearly, if job minavailable is not set, it will be set to sumof(task minavailable), otherwise it will not be changed and keep user settings.

That's reasonable.

Thor-wl:
Please fix the bug from CI

config/crd/bases/batch.volcano.sh_jobs.yaml
huone1: the topologyPolicy is added by me in api repo , but the PR hasn't been merged in volcano

shinytang6: emmm yeah, but it has been merged into volcano/apis, so the script generated these definitions..

lowang-bh:
> > currently not, I think it is a bit difficult to achieve because we actually don’t know the state of the cluster when we choose which task to schedule, so it is difficult to make the best choice. WDYT, any advice?
> 
> Maybe we can use `max(0, task.minAvailable - scheduledNum) / task.minAvailable` as priority to determine which task will be the next candidate

Currently, the allocating order will be task1-0, task2-0, task1-1, task2-1, ..., and so on.  So this problem does not exist any more.
https://github.com/volcano-sh/volcano/blob/34836851e6317807267230b1b9508194f81a6f67/pkg/controllers/job/helpers/helpers.go#L51-L66

PR #1477, opened 2021-05-15
vcctl queue support get kubeconfig from env

yahaa:
vcctl queue support get kubeconfig from KUBECONFIG env variable.


shinytang6 (approved):

Thor-wl:
This PR is merged for all review is ok and the CI doesn't work well.

Issue #1449, opened 2021-04-29
"file name too long" bug when install volcano with hack/local-up-volcano.sh


Thor-wl:
**What happened**:
When install volcano with hack/local-up-volcano.sh，error occurs with message "file name too long"

**What you expected to happen**:
Installing Volcano with hack/local-up-volcano.sh works well
**How to reproduce it (as minimally and precisely as possible)**:
bash hack/local-up-volcano.sh

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master branch
- Kubernetes version (use `kubectl version`): v1.19
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release): CentOS 7
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Issue #1445, opened 2021-04-27
CRD validation error


pangbingqiang1:
**What happened**:
The CustomResourceDefinition "jobs.batch.volcano.sh" is invalid:
![截图](https://user-images.githubusercontent.com/48078129/116209601-aae34c80-a774-11eb-9e58-b2b255e1f52d.PNG)

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/master/installer/volcano-development.yaml

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`): 1.18.12
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Thor-wl:
Thank you for your report. We've already discovered this bug recently. Bug fix is on the way.

huone1:
fix it in PR #1446

huone1:
you can resolve it by  “make update-development-yaml CRD_VERSION=v1”  to  updating volcano-development.yaml

Thor-wl:
This bug is fixed and close this issue.

PR #1483, opened 2021-05-21
fix the bug of CRD apiversion and installation volcano with kind

Thor-wl:
ref: https://github.com/volcano-sh/volcano/issues/1449#event-4661573569
ref: https://github.com/volcano-sh/volcano/issues/1445


hack/local-up-volcano.sh
```diff
@@ -95,9 +95,7 @@ prepare
 
 if [ "${INSTALL_MODE}" == "kind" ]; then
   kind-up-cluster
-  export KUBECONFIG="$(kind get kubeconfig ${CLUSTER_CONTEXT})"
+  export KUBECONFIG=${HOME}/.kube/config
```
shinytang6, line 98: is there any reason changing this line?

config/crd/v1beta1/bus.volcano.sh_commands.yaml
```diff
@@ -1,6 +1,6 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
```
huone1, line 3: why do you change the version to V1 in the v1beta1 folder

hack/local-up-volcano.sh
Thor-wl: > is there any reason changing this line? l think if only the user not using `kind`, the default KUBECONFIG can be set as `${HOME}/.kube/config ` ?

Currently, if users install kind cluster and volcano with this shell script, it will be failed because it cannot get kubeconfig path from env KUBECONFIG.

config/crd/v1beta1/bus.volcano.sh_commands.yaml
Thor-wl: > why do you change the version to V1 in the v1beta1 folder, the base folder is v1 version

When install volcano with the script, it will be warning for `v1beta` is invalid.

Makefile
```diff
@@ -115,7 +115,7 @@ generate-code:
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) paths="./vendor/volcano.sh/apis/pkg/apis/scheduling/v1beta1;./vendor/volcano.sh/apis/pkg/apis/batch/v1alpha1;./vendor/volcano.sh/apis/pkg/apis/bus/v1alpha1;" output:crd:artifacts:config=config/crd/bases
-	$(CONTROLLER_GEN) "crd:crdVersions=v1beta1" paths="./vendor/volcano.sh/apis/pkg/apis/scheduling/v1beta1;./vendor/volcano.sh/apis/pkg/apis/batch/v1alpha1;./vendor/volcano.sh/apis/pkg/apis/bus/v1alpha1;" output:crd:artifacts:config=config/crd/v1beta1
+	$(CONTROLLER_GEN) "crd:crdVersions=v1" paths="./vendor/volcano.sh/apis/pkg/apis/scheduling/v1beta1;./vendor/volcano.sh/apis/pkg/apis/batch/v1alpha1;./vendor/volcano.sh/apis/pkg/apis/bus/v1alpha1;" output:crd:artifacts:config=config/crd/v1beta1
```
shinytang6, line 118: l think y don't need to change this? `CRD_OPTIONS ` is already `crd:crdVersions=v1`.

Thor-wl: > l think y don't need to change this? `CRD_OPTIONS ` is already `crd:crdVersions=v1`.

Yes, i'll recover it

shinytang6 (approved):
/lgtm

PR #1490, opened 2021-05-27
sort resource comparation functions, add accurate definitions and fix…

Thor-wl:
The modifications are as follows:
1. set min resource as const
2. resort functions as groups according its function
3. add assurate definitions for resource comparation functions
4. fix some bugs such as lack of validation for keys in map
5. delete `Scale` function for it is all the same with `Multi`

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -27,6 +27,15 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util/assert"
 )
 
+const (
+	// GPUResourceName need to follow https://github.com/NVIDIA/k8s-device-plugin/blob/66a35b71ac4b5cbfb04714678b548bd77e5ba719/server.go#L20
+	GPUResourceName = "nvidia.com/gpu"
+)
+
+const (
+	minResource float64 = 0.1
+)
+
 // Resource struct defines all the resource type
 type Resource struct {
 	MilliCPU float64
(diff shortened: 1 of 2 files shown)


pkg/scheduler/api/resource_info.go
```diff
@@ -208,30 +245,23 @@ func (r *Resource) FitDelta(rr *Resource) *Resource {
...
-			r.ScalarResources = map[v1.ResourceName]float64{}
-		}
+	if r.ScalarResources == nil {
+		r.ScalarResources = make(map[v1.ResourceName]float64)
+	}
 
-		if rrQuant > 0 {
+	for rrName, rrQuant := range rr.ScalarResources {
+		_, ok := r.ScalarResources[rrName]
+		if ok && rrQuant > 0 {
```
wpeng102, line 254: if r does not have `rrName` also need do `r.ScalarResources[rrName] -= rrQuant + minResource`

Thor-wl: > if r does not have `rrName` also need do `r.ScalarResources[rrName] -= rrQuant + minResource`

If that, we should initially set `r.ScalarResources[rrName] == 0`

wpeng102 (approved):
/lgtm

pkg/scheduler/api/resource_info.go
```diff
@@ -68,32 +93,44 @@ func (r *Resource) Clone() *Resource {
...
+		return r.ScalarResources[rn]
 	}
-	return r
 }
 
-// IsEmpty returns bool after checking any of resource is less than min possible value
+// IsEmpty returns false if any kind of resource is not less than min value, otherwise returns true
 func (r *Resource) IsEmpty() bool {
-	if !(r.MilliCPU < minResource && r.Memory < minResource) {
+	if r.MilliCPU >= minResource || r.Memory >= minResource {
```
william-wang, line 133: do not use '>=' for float comparing.  Combining '!' and '<'  is a general way.

pkg/scheduler/api/resource_info.go
```diff
@@ -40,17 +49,33 @@ type Resource struct {
...
-)
-
 // EmptyResource creates a empty resource object and returns
 func EmptyResource() *Resource {
 	return &Resource{}
 }
 
-// Clone is used to clone a resource type
+// NewResource creates a new resource object from resource list
+func NewResource(rl v1.ResourceList) *Resource {
```
william-wang, line 58: If there is no change for the function. We'd better keep it as it is.

Thor-wl: > If there is no change for the function. We'd better keep it as it is.

No changes about these functions, just resort their orders.

pkg/scheduler/api/resource_info.go
Thor-wl: > do not use '>=' for float comparing. Combining '!' and '<' is a general way.

Yes, i will recover it

william-wang (approved):
/lgtm
/approve

shinytang6 (approved):
/lgtm
/approve

PR #1491, opened 2021-05-27
refresh version number in Makefile.def

Thor-wl:


Thor-wl:
Close this PR for we should not modifiy the file on master branch

PR #1493, opened 2021-05-27
submit cputopo feature

huone1:
The CPU Manager and the Topology Manager are all Kubelet components, However There is the following limitation:

1. The scheduler is not topology-aware. so it is possible to be scheduled on a node and then fail on the node due to the Topology Manager. this is unacceptable for Tensorflow job. If any worker or ps failed on node, the job will fail.
2. The managers are node-level that results in an inability to match the best node in a cluster.

the related PR is https://github.com/volcano-sh/resource-exporter/pull/1
the design doc PR is https://github.com/volcano-sh/volcano/pull/1437


wpeng102 (approved):
/lgtm

william-wang (approved):
/lgtm

Thor-wl (approved):

PR #1497, opened 2021-05-28
fix select wrong queue when `proportion` is disable

### LINKED ISSUE #1496: Job will always pending when proportion plugin is disable
**What happened**:

1. Disable the plugin `proportion`.
2. Deploy multi jobs to different queues in one namespace.
3. Some jobs will always pending.

These jobs will be scheduled.

https://github.com/volcano-sh/volcano/blob/44ec8eb28df26fbacd03ab15edabfc5916900c25/pkg/scheduler/framework/session_plugins.go#L510-L515

The `proportion` plugin is the only one which implements `queueOrderFn`, so if we disable it we will use the creation timestamp of queue to order.

https://github.com/volcano-sh/volcano/blob/44ec8eb28df26fbacd03ab15edabfc5916900c25/pkg/scheduler/actions/allocate/allocate.go#L139-L151

From `allocate.go` we can find `ssn.QueueOrderFn` will always return the very first created queue, thus the after created will never be scheduled even the pending jobs in previous queue is empty.

**Environment**:
- Volcano Version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zen-xu:
fix #1496

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -1,17 +1,17 @@
 /*
-Copyright 2017 The Kubernetes Authors.
+ Copyright 2021 The Volcano Authors.
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


Thor-wl (commented):
How about move the queue order into enqueue action for it'a required action?

shinytang6 (approved):
/lgtm

PR #1508, opened 2021-06-01
fix the error for make verify

huone1:


hack/update-gencode.sh
```diff
@@ -22,16 +22,19 @@ echo $(dirname "${BASH_SOURCE[0]}")
...
 
+chmod 755 vendor/k8s.io/code-generator/generate-groups.sh
+chmod 755 vendor/k8s.io/code-generator/generate-internal-groups.sh
+
 # generate the code with:
 # --output-base    because this script should also be able to run inside the vendor dir of
 #                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
 #                  instead of the $GOPATH directly. For normal projects this can be dropped.
-${SCRIPT_ROOT}/hack/generate-groups.sh "deepcopy,client,informer,lister" \
+vendor/k8s.io/code-generator/generate-groups.sh "deepcopy,client,informer,lister" \
```
shinytang6, line 32: it's better to use `bash` here and rm chmod above

huone1: It's better. done

shinytang6 (approved):
/lgtm

PR #1510, opened 2021-06-01
use struct NodeScoreList better than HostPriorityList for score

huone1:

### DIFF
--- a/pkg/scheduler/util/scheduler_helper.go
+++ b/pkg/scheduler/util/scheduler_helper.go
@@ -19,16 +19,16 @@ package util
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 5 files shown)


pkg/scheduler/util/scheduler_helper.go
```diff
@@ -19,6 +19,7 @@ package util
 import (
 	"context"
 	"fmt"
+	k8sframework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
```
shinytang6, line 22: pls move it to the third package

shinytang6 (approved):
/lgtm

PR #1517, opened 2021-06-05
Improve: remove nodename from taskinfo when RemoveTask

lowang-bh:
Signed-off-by: lowang_bh <lhui_wang@163.com>

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -116,8 +116,8 @@ func getTaskID(pod *v1.Pod) TaskID {
 
 // NewTaskInfo creates new taskInfo object for a Pod
 func NewTaskInfo(pod *v1.Pod) *TaskInfo {
-	req := GetPodResourceRequest(pod)
-	initResreq := GetPodResourceRequest(pod)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -116,8 +116,7 @@ func getTaskID(pod *v1.Pod) TaskID {
 
 // NewTaskInfo creates new taskInfo object for a Pod
 func NewTaskInfo(pod *v1.Pod) *TaskInfo {
-	req := GetPodResourceRequest(pod)
-	initResreq := GetPodResourceRequest(pod)
+	resReq := GetPodResourceRequest(pod)
```
k82cn, line 119: I think we should distinguish initReq vs. req.

lowang-bh: changed to 
```go
	initResReq := GetPodResourceRequest(pod)
	resReq := initResReq
```

lowang-bh:
@tizhou86 GetDevicesIdleGPUMemory miss deviceID that has not been used ever, is it as design or not?

shinytang6 (approved):
/lgtm
/approve

PR #1530, opened 2021-06-09
support get queue from namespace

hacker-qian:
Signed-off-by: q00575201 <qianjianmin@huawei.com>

In some scenarios,  the namespace is associated with the user’s level (e.g., v1 namespace is created by v1 user). Therefore, when scheduling jobs with the queue, we  also want the queue associates with the user level and the namespace (e.g., v1-queue). This PR supports get queue information from the namespace. To avoid interfering with other scenes, we also add a switch to controll this function.

### DIFF
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -39,16 +39,12 @@ func OpenSession(cache cache.Cache, tiers []conf.Tier, configurations []conf.Con
 			} else {
 				plugin := pb(plugin.Arguments)
 				ssn.plugins[plugin.Name()] = plugin
+				onSessionOpenStart := time.Now()
+				plugin.OnSessionOpen(ssn)
+				metrics.UpdatePluginDuration(plugin.Name(), metrics.OnSessionOpen, metrics.Duration(onSessionOpenStart))
 			}
 		}
 	}
-
-	for _, plugin := range ssn.plugins {
-		onSessionOpenStart := time.Now()
-		plugin.OnSessionOpen(ssn)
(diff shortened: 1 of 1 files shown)


k82cn:
/hold

> In some scenarios, the namespace is associated with the user’s level (e.g., v1 namespace is created by v1 user). 

I do not think we should bind ns to queue, please move this to your plugin instead of framework.

hacker-qian:
> /hold
> 
> > In some scenarios, the namespace is associated with the user’s level (e.g., v1 namespace is created by v1 user).
> 
> I do not think we should bind ns to queue, please move this to your plugin instead of framework.

hacker-qian:
> /hold
> 
> > In some scenarios, the namespace is associated with the user’s level (e.g., v1 namespace is created by v1 user).
> 
> I do not think we should bind ns to queue, please move this to your plugin instead of framework.

> I think this is a common ability in severless scenario, because in severless scenario, cluster resource is managed by cloud provider, user cannot directly divide resource with queue in the cluster.

hacker-qian:
> /hold
> 
> > In some scenarios, the namespace is associated with the user’s level (e.g., v1 namespace is created by v1 user).
> 
> I do not think we should bind ns to queue, please move this to your plugin instead of framework.

> I think this is a common ability in severless scenario, because in severless scenario, cluster resource is managed by cloud provider, user cannot directly divide resource with queue in the cluster.

k82cn:
> I think this is a common ability in severless scenario, because in severless scenario, cluster resource is managed by cloud provider, user cannot directly divide resource with queue in the cluster.

we do not build volcano for severless scenario **only**, so we did not bind ns to queue.

PR #1537, opened 2021-06-15
Elect Action should also consider PodGroup.Status == PodGroupInqueue as candidates 

umialpha:


umialpha:
@Thor-wl please help to review, thx.

pkg/scheduler/actions/elect/elect.go
```diff
@@ -47,5 +47,9 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	}
 }
 
+func (e *Action) isJobElectable(job *api.JobInfo) bool {
+	return job.PodGroup.Status.Phase == scheduling.PodGroupPending || job.PodGroup.Status.Phase == scheduling.PodGroupInqueue
```
Thor-wl, line 51: Can you give more details about this modification?

Thor-wl:
Please fix the DCO: https://github.com/volcano-sh/volcano/pull/1537/checks?check_run_id=2826267968

pkg/scheduler/actions/elect/elect.go
```diff
@@ -18,23 +18,23 @@ func New() *Action {
...
+func (e *Action) Execute(ssn *framework.Session) {
 	klog.V(3).Infof("Enter Elect ...")
 	defer klog.V(3).Infof("Leaving Elect ...")
 
 	if util.Reservation.TargetJob == nil {
 		klog.V(4).Infof("Start select Target Job")
 		var pendingJobs []*api.JobInfo
 		for _, job := range ssn.Jobs {
-			if job.PodGroup.Status.Phase == scheduling.PodGroupPending {
+			if job.PodGroup.Status.Phase == scheduling.PodGroupInqueue {
```
Thor-wl, line 37: The job, whose podgroup is `inqueue`, is elected as target job. That may be not so reasonable for it will be scheduled at normal process originally.

umialpha: I agree. But some plugins permit inqueue if queue's capacity not set, e.g. proportion. I prefer putting `elect` behind `allocate` and check podgroup in `inqueue` status. At that stage, podgroups in `inqueue` status are starving.

wpeng102: I think `overcommit` plugin maybe block some large job change to `inqueue`.  Please try to config `overcommit` plugin.

k82cn:
Are we going to deprecate elect action? AFAIK, we already have SLA plugin for that.

Thor-wl:
> Are we going to deprecate elect action? AFAIK, we already have SLA plugin for that.

Elect action can be used for all scenarios that users want to elect job satisfing some conditions, not just for job resource reservation. So i think we can reserve it.

k82cn:
> Elect action can be used for all scenarios that users want to elect job satisfing some conditions, 

Any specific case for that?

pkg/scheduler/actions/elect/elect.go
huone1: I think we can add another condition : Whether the job has been into the scheduler session but the job's status is not running

PR #1553, opened 2021-06-22
fix: Deployment with request exceeding queue capacity should not be scheduled

### LINKED ISSUE #1542: K8s Deployment resource request exceed Queue capability still can be scheduled
**What happened**:
when K8s Deployment resource request exceed Queue capability , the deployment still can be scheduled by volcano

1. queue
spec:
  capability:
    cpu: "1"
    memory: "1024Mi"
2. deployment
  request:
    cpu: "2"
    memory: "2048Mi"

deployment exceed the queue capability, should not be scheduled by volcano

create a queue and deployment with specific queue capability and deployment resource request and limit value

no

**Environment**:
- Volcano Version: 1.2.x
- Kubernetes version (use `kubectl version`): 1.17
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ChAnYaNG97:
try to fix #1542

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -18,6 +18,7 @@ package podgroup
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -141,3 +151,50 @@ func newPGOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
...
+
+	if memReq.Cmp(*initMemReq) < 0 {
+		memReq = initMemReq
+	}
+
+	if len(pod.OwnerReferences) != 0 {
+		for _, ownerReference := range pod.OwnerReferences {
+			if ownerReference.Controller != nil && *ownerReference.Controller && ownerReference.Kind == "ReplicaSet" {
+				rsName := ownerReference.Name
+				rs, err := pg.kubeClient.AppsV1().ReplicaSets(pod.Namespace).Get(context.TODO(), rsName, metav1.GetOptions{})
```
shinytang6, line 185: I'm not sure if it is a good approach because it assumes that the pod is from a replicaset. Do we need to support pods under other resources (although no one has mentioned it at the moment except deployment)
/cc @Thor-wl @jasonliu747

ChAnYaNG97: Agree, need further discussion.

jasonliu747 (commented):
/hold
@Thor-wl WDYT

ChAnYaNG97:
@Thor-wl

Thor-wl:
Why close this PR? It's on reviewing. @ChAnYaNG97

ChAnYaNG97:
> /hold cancel
> @ChAnYaNG97 any progress on this PR? 😄

Almost no yet. I'm stuck and will discuss with @Thor-wl later.

PR #1563, opened 2021-06-25
Improve: preempt break out if intersection is null

lowang-bh:
when `candidates` is null in last plugin, there is no need to check any more due to `intersection` will be null.

return it directly to improve perfomance.

### DIFF
--- a/pkg/scheduler/framework/session_plugins.go
+++ b/pkg/scheduler/framework/session_plugins.go
@@ -148,6 +148,11 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 				continue
 			}
 			candidates := rf(reclaimer, reclaimees)
+			// intersection will be nil if length is 0, don't need to do any more check
+			if len(candidates) == 0 {
+				victims = nil
+				break
+			}
 			if !init {
 				victims = candidates
 				init = true
@@ -191,6 +196,12 @@ func (ssn *Session) Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskI
 				continue
 			}
 			candidates := pf(preemptor, preemptees)
+			// intersection will be nil if length is 0, don't need to do any more check
+			if len(candidates) == 0 {
+				victims = nil
+				break
+			}
+
 			if !init {
 				victims = candidates
 				init = true


lowang-bh:
this change will met it to find victims when all plugins met preemptable;
if it just want to find victims between tiers, should use break instead of return.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -209,8 +214,8 @@ func (ssn *Session) Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskI
 				victims = intersection
 			}
 		}
-		// Plugins in this tier made decision if victims is not nil
-		if victims != nil {
+		// Plugins in this tier made decision if victims is nil because intersection will be nil
+		if victims == nil {
```
shinytang6, line 218: if victims is nil in first tier, the decision should be made by next tier, and so on... So l don't think you should change this.

lowang-bh: yeal, I have revet this one.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -148,6 +148,11 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 				continue
 			}
 			candidates := rf(reclaimer, reclaimees)
+			// intersection will be nil if lenth is 0, don't need to do any more check
```
shinytang6, line 151: s/lenth/length

pkg/scheduler/framework/session_plugins.go
```diff
@@ -148,6 +148,11 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
...
+			if len(candidates) == 0 {
+				victims = nil
```
shinytang6, line 153: No need for this assignment, just break

lowang-bh: how about the first plugin return's candidates is not null, and other plugins's candidates is null? That will cause victims has contents and not to be nil.

pkg/scheduler/framework/session_plugins.go
lowang-bh: fixed

pkg/scheduler/framework/session_plugins.go
shinytang6: ok, l see. It's my bad

shinytang6 (approved):
/lgtm
/approve

jasonliu747 (approved):
/lgtm

Thor-wl (approved):

Issue #1322, opened 2021-02-20
Support multi scheduler by admission controller


k82cn:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**Description**:

One option is use admission to balance workload to different resource group, e.g. default scheduler for system daemon, and volcano scheduler for biz workload.

Thor-wl:
@huone1 has implemented that. Ref: https://github.com/volcano-sh/volcano/pull/1576

PR #1565, opened 2021-06-26
add multi-scheduling design doc

huone1:
issue #1322


docs/design/multi-scheduler.md
```diff
@@ -0,0 +1,150 @@
+# multi-scheduling
```
shinytang6, line 1: generally LGTM. Do we need to change the name of the document(maybe hybrid scheduler or something?) to distinguish because there is already a document(multi-volcano-schedulers.md) explaining how to deploy multiple **volcano** schedulers

Tammy-kunyu (commented):
/lgtm

shinytang6 (commented):
By the way, the assumption of this document is that the **management-type pod is scheduled by default scheduler and the job-type pod is scheduled by volcano**. In fact, it only relieves users from tainting/adding labels on the pod with this strong assumption. l think it did not solve the real problem of mixed resources conflict.

huone1:
> By the way, the assumption of this document is that the **management-type pod is scheduled by default scheduler and the job-type pod is scheduled by volcano** and **management-type pods will be scheduled to management ns**. In fact, it only relieves users from tainting/adding labels on the pod with this strong assumption, but l think it will cause mixed resources conflict once user submit management pods with volcano or user want to submit management pods to other ns.



> By the way, the assumption of this document is that the **management-type pod is scheduled by default scheduler and the job-type pod is scheduled by volcano** and **management-type pods will be scheduled to management ns**. In fact, it only relieves users from tainting/adding labels on the pod with this strong assumption, but l think it will cause mixed resources conflict once user submit management pods with volcano or user want to submit management pods to other ns.


> By the way, the assumption of this document is that the **management-type pod is scheduled by default scheduler and the job-type pod is scheduled by volcano** and **management-type pods will be scheduled to management ns**. In fact, it only relieves users from tainting/adding labels on the pod with this strong assumption, but l think it will cause mixed resources conflict once user submit management pods with volcano or user want to submit management pods to other ns.

management pods must be in the namespace list you set。user use taint + nodeSelector to resolve resources conflict ;
the admission would complete based on the admission configmap,so  the admission configmap must adapt to user scenarios .

Today, I will optimize the design and refresh the doc.

docs/design/multi-scheduler.md
```diff
@@ -0,0 +1,213 @@
...
+        key: namespace                             # set the field and the value to be matched
+        value:
+        - mng-ns-1
+      schedulerName: default-scheduler             # set the scheduler for patching
+      tolerations:                                 # set the tolerations for patching
+      - effect: NoSchedule
+        key: management-taint
+        operator: Exists
+      labels:
+        volcano.sh/nodetype: management           # set the nodeSelector for patching
```
shinytang6, line 114: this label is wrong

shinytang6 (commented):
rest lgtm

docs/design/multi-scheduler.md
huone1: > this label is wrong

what's wrong about the label?

shinytang6: in this case, it should be `mng-1-label: value-1`?

huone1: Got it . I modify the node info from "mng-1-label: value-1" to "volcano.sh/nodetype: management"

huone1: done

docs/design/multi-scheduler.md
```diff
@@ -0,0 +1,213 @@
...
+
+2. Submit deployment-A and volcano-job-B
+
+3. Check the deployment Pod 
+ 
+````
+deployment-A:
+....
+nodeSelector:
+  mng-1-label: value-1
```
shinytang6, line 125: the nodeSelector should also be modified

huone1: I'm so sorry and fix it now

shinytang6 (approved):
thx, LGTM now
/lgtm

wpeng102 (approved):
/lgtm

Issue #1348, opened 2021-03-08
Support Oversubscription for mix workload scenarios


k82cn:
#### What would you like to be added:

When some application request more resources than it need, publish those "idle" resources to the scheduler to dispatch preemptable tasks and reclaim them when there's CPU pressure. For now, only CPU is supported.

#### Why is this needed:

That's helpful to improve the resource usage of cluster.

k82cn:
Both scheduler & kubelet should be enhanced to support this feature.

PR #1566, opened 2021-06-28
support oversubscription in volcano framework

wpeng102:
https://github.com/volcano-sh/volcano/issues/1348

Our team is working on cluster resource oversubscription on volcano as a internal feature. When some application request more resources than it need, publish those "idle" resources to the scheduler to dispatch preemptable tasks and reclaim them when there's CPU pressure. 
So, we need volcano framework 1) support to record the oversubscription resource. 2) support plugin can drop preempt/reclaim decision in some cases.

Currently, we use 
`volcano.sh/oversubscription` to record if node is enable oversubscription feature
`volcano.sh/oversubscription-cpu` to record cpu oversubscription number
`volcano.sh/oversubscription-memory` to record memory oversubscription number
`volcano.sh/offline-job-evicting` to record if node has stop oversubscription resoruce due to resource pressure.

Introduce second return value for `reclaimableFns` and `preemptableFns` , if plugin return `abstain (0 value)`, it meas this plugin drop the preempt/reclaim decision.


pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -27,6 +27,7 @@ import (
 
 	"volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/framework"
+	putil "volcano.sh/volcano/pkg/scheduler/plugins/util"
 	tutil "volcano.sh/volcano/pkg/scheduler/plugins/util"
```
jasonliu747, line 31: duplicate import

jasonliu747 (changes requested):

Thor-wl (approved):

william-wang (approved):
/lgtm
/approve

Issue #1567, opened 2021-06-28
lost annotation in task.template


hwdef:
**What happened**:

When I add a annotation to vcjob spec.tasks.template.metadata,But the running pod does not have this annotation.

**What you expected to happen**:

Keep annotation in pod

**How to reproduce it (as minimally and precisely as possible)**:

create a vcjob

```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: test-job
spec:
  schedulerName: volcano
  priorityClassName: high-priority
  policies:
    - event: PodEvicted
      action: RestartJob
  maxRetry: 10
  tasks:
    - replicas: 1
      name: "default-nginx"
      template:
        metadata:
          name: web
          annotations:
            "xxxxxxxxxxxxx": "xxxxxxxxxxxxxx"
        spec:
          containers:
            - image: nginx
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
                requests:
                  cpu: "100m"
          restartPolicy: OnFailure
```

get this vcjob 

```yaml
[centos@master ~]$ kubectl get vcjob test-job -oyaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"batch.volcano.sh/v1alpha1","kind":"Job","metadata":{"annotations":{},"name":"test-job","namespace":"default"},"spec":{"maxRetry":10,"policies":[{"action":"RestartJob","event":"PodEvicted"}],"priorityClassName":"high-priority","schedulerName":"volcano","tasks":[{"name":"default-nginx","replicas":1,"template":{"metadata":{"annotations":{"xxxxxxxxxxxxx":"xxxxxxxxxxxxxx"},"name":"web"},"spec":{"containers":[{"image":"nginx","imagePullPolicy":"IfNotPresent","name":"nginx","resources":{"requests":{"cpu":"100m"}}}],"restartPolicy":"OnFailure"}}}]}}
  creationTimestamp: "2021-06-28T02:15:27Z"
  generation: 1
  name: test-job
  namespace: default
  resourceVersion: "1446893"
  uid: 6e0ddf28-d920-4eae-a0c0-f38f4e35298e
spec:
  maxRetry: 10
  minAvailable: 1
  policies:
  - action: RestartJob
    event: PodEvicted
  priorityClassName: high-priority
  queue: default
  schedulerName: volcano
[...]

PR #1615, opened 2021-07-13
Adds the missing field to the Webhook

hwdef:
Ref: #1567

I'm not sure I'm right about whether these fields are permissible for external use.

Also, about adding aliases, does that mean that some events are the same? For example `podFailed` and `taskFailed`, I am not sure if this is correct,  If only 10% of pods fail, is taskfailed? or only 1% of pod fail?

### DIFF
--- a/pkg/webhooks/admission/jobs/validate/util.go
+++ b/pkg/webhooks/admission/jobs/validate/util.go
@@ -37,6 +37,8 @@ var policyEventMap = map[busv1alpha1.Event]bool{
 	busv1alpha1.TaskCompletedEvent: true,
 	busv1alpha1.OutOfSyncEvent:     false,
 	busv1alpha1.CommandIssuedEvent: false,
+	busv1alpha1.JobUpdatedEvent:    true,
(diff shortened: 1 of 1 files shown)


hwdef:
assign @k82cn :)

k82cn:
/lgtm
/approve

re aliases: hm..., I think we need to clarify the definition of `TaskFailed` for your case :)

jasonliu747 (commented):
Hi, could you fix the DCO problem? Otherwise the merge will be blocked :(

hwdef:
> Hi, could you fix the DCO problem? Otherwise the merge will be blocked :(

sorry about that ,I fixed it.

hwdef:
@k82cn  How about we discuss this with others at the weekly meeting? Wait until we confirm how to handle it before creating a new PR on aliases

k82cn:
> @k82cn How about we discuss this with others at the weekly meeting? Wait until we confirm how to handle it before creating a new PR on aliases

Sure :)

Issue #1617, opened 2021-07-14
unpipeline did not release resources


merryzhou:
Before call DeallocateFuncs, the task.NodeName was set to ""

https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/framework/statement.go#L203
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/predicates/predicates.go#L208

merryzhou:
https://github.com/volcano-sh/volcano/issues/979

Thor-wl:
So this issue can be closed? @merryzhou

PR #1618, opened 2021-07-14
fix: reset task.NodeName after call DeallocateFunc

merryzhou:
https://github.com/volcano-sh/volcano/issues/1617


wpeng102 (approved):
/lgtm

PR #1626, opened 2021-07-18
add task-spec in pod label

python279:
add task-spec in pod label since we want to add service to select the specific task spec pod. like the notebook app, we only want to expose the 22 (sshd) port of the master task pod.


shinytang6 (changes requested):
pls fix the DCO

wpeng102 (approved):
/lgtm

wpeng102:
@python279 Thanks for your contribution!

Issue #1525, opened 2021-06-08
Enhancement about resource comparision functions according to different scenarios


Thor-wl:
According to backfeed from community, current resource comparasion functions at resource_info.go cannot meet all scenarios and introduce some bugs. We should reorganize the scenarios and optimize the resource comparasions

k82cn:
Wow, that's really long time since last time we raised this concern :(

shinytang6:
A collection of typical problems:

Less function logic bug: https://github.com/volcano-sh/volcano/issues/1036
If a resource (Except CPU and mem) doesn't configure in queue capability, it will be seen as Zero in resources comparation: https://github.com/volcano-sh/volcano/issues/1485
when one resource(GPU) is overused in queue, other resources (CPU) instances in the queue cannot be scheduled: https://github.com/volcano-sh/volcano/issues/1425

Thor-wl:
As what i ananlysed last week, bugs of resoure comparision logic mainly lies at the following aspects:
* When comparing resourceList in L and R, the blank item should be regarded as infinity value instead of zero, especially comparing queue.capability with others. For example, L=queue.deserved = {CPU: 1c, memory:2G, GPU: 10}, R=queue.capability={GPU: 2}.
* Lack of resource comparision functions such as LessInSomeDemision, we should supply them.
* Because of lack of definition comments for comparision functions, developer may misunderstand the real meaning and use them by mistake, which introduce some bugs.

Thor-wl:
Based on the above discovery, I give the following suggestions and will put them into practice as bug fix:
* List all resource comparision scenarios and give the correctly recommanded function.
* Add `defaultValue` as a parameter for all resource comparison funciton, its value can only be one between `zero` or `infinity`
* Fix existing resource comparision bugs in all plugins, especially in `proportion` and `preempt`

Thor-wl:
Recommanded Resource Comparision Functions:
| Name | Comment | Example | Original function | Used plugins/actions |
| :----:| :----: | :----: | :----: | :----: |
| l.LessInAllDimension(r *api.Resource, defaultVaule string) | Values in all demisions in `l` are less than that in `r` | L{cpu:1c, memory:2G} < R{cpu:2c, memory:4G} | Less(rr *Resource) | proportion |
[...]

PR #1628, opened 2021-07-19
fix bugs about resource comparison

Thor-wl:
ref: https://github.com/volcano-sh/volcano/issues/1525

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -220,7 +220,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 				allocations[job.Queue] = attr.allocated.Clone()
 			}
 			allocated := allocations[job.Queue]
-			if allocated.Less(reclaimee.Resreq, api.Zero) {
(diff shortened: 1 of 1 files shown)


lowang-bh (commented):
log may changed to follow up the fix

Thor-wl:
> log may changed to follow up the fix

I'm sorry for not get the point. What's the influence about the fix?

lowang-bh:
> > log may changed to follow up the fix
> 
> I'm sorry for not get the point. What's the influence about the fix?

klog.V(3).Infof("Failed to allocate resource for Task <%s/%s> in Queue <%s>, not enough resource.",		reclaimee.Namespace, reclaimee.Name, job.Queue)

In this log info, param should be reclaimer, I think.

Thor-wl:
> > > log may changed to follow up the fix
> > 
> > 
> > I'm sorry for not get the point. What's the influence about the fix?
> 
> klog.V(3).Infof("Failed to allocate resource for Task <%s/%s> in Queue <%s>, not enough resource.", reclaimee.Namespace, reclaimee.Name, job.Queue)
> 
> In this log info, param should be reclaimer, I think.

Thanks. Let me fix it.

PR #1630, opened 2021-07-19
add design docs for task-DAG-scheduling

hwdef:


docs/design/task-DAG-scheduling.md
```diff
@@ -0,0 +1,85 @@
...
+2.2 `trigger` indicates the point in time when the trigger, it has two fields, `status` and `time`, `status` indicates the status of the task, `time` indicates the duration of time since the state was reached,it can be empty, indicating 0 seconds.
+
+```go
+type TriggerSpec struct {
+	Status string
+	Time   time.Duration
+}
+```
+
+2.3 `probe` is used to detect the status of the task in dependOn. This field can be used with the `probe` struct in kubernetes
```
shinytang6, line 63: I don’t quite understand what the field `probe ` is used for, can you elaborate more?

docs/design/task-DAG-scheduling.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+2.1 `dependOn` is a string indicating the name of the task that the current task depends on
+
+2.2 `trigger` indicates the point in time when the trigger, it has two fields, `status` and `time`, `status` indicates the status of the task, `time` indicates the duration of time since the state was reached,it can be empty, indicating 0 seconds.
+
+```go
+type TriggerSpec struct {
+	Status string
+	Time   time.Duration
```
shinytang6, line 59: Does `Time` means the time period after reaching the state or the period after the task starting to run

docs/design/task-DAG-scheduling.md
```diff
@@ -0,0 +1,85 @@
...
+	Probe    *v1.Probe
+}
+```
+2.1 `dependOn` is a string indicating the name of the task that the current task depends on
+
+2.2 `trigger` indicates the point in time when the trigger, it has two fields, `status` and `time`, `status` indicates the status of the task, `time` indicates the duration of time since the state was reached,it can be empty, indicating 0 seconds.
+
+```go
+type TriggerSpec struct {
+	Status string
```
shinytang6, line 58: And l think we should describe how we define task status

shinytang6 (changes requested):
And l think we should also add corresponding check in the webhook

docs/design/task-DAG-scheduling.md
hwdef: Thanks for your suggestions. I will add some documents.

hwdef: I have updated documents and modified some fields, let us discuss on this basis.

docs/design/task-DAG-scheduling.md
hwdef: `time` is the time period after reaching the state

hwdef: Such as the task has been running for 10s

docs/design/task-DAG-scheduling.md
hwdef: I think probe applies to both cases in lines 29 and 30 of the documentation

shinytang6: So do you mean the probe configured here should be applied to each containers in pod template?

docs/design/task-DAG-scheduling.md
shinytang6: ls there any usage scenarios? Generally, l think `Status` is enough, wdyt?

shinytang6 (changes requested):

docs/design/task-DAG-scheduling.md
hwdef: I think for both HTTP and TCP, there is no need to consider the issue of multiple containers. But `exec` method should be selected containers.

hwdef: I can think of two ways:
* Add a field to specify the container name
* dependOn.taskname in the format of `taskName/containerName`

docs/design/task-DAG-scheduling.md
hwdef: I haven't thought of any scenarios yet, but I think it will be more flexible with the addition of the time field.

docs/design/task-DAG-scheduling.md
shinytang6: IMO, we should only care about taskName in task spec, so this field is not so necessary

docs/design/task-DAG-scheduling.md
shinytang6: In terms of realization, if we want to support this, we also need to record the transition time of task (e.g maybe the time when all replicas under one task converting to running), which is not available right now

docs/design/task-DAG-scheduling.md
hwdef: If don't add fields, then how to specify the container? Do you have any suggestions?

docs/design/task-DAG-scheduling.md
hwdef: yes, you are right, But I think maybe we can just use task to meet the minavailable time

docs/design/task-DAG-scheduling.md
```diff
@@ -0,0 +1,92 @@
...
+
+This feature provides the ability to customize the order in which tasks are launched.
+
+## Scenarios
+
+* mpi job. the master needs to wait for the worker to start before starting, If the master is already started, but the worker is not yet started, the master will restart, which will add unnecessary waste of resources ,in this case, mpiworker needs to be in the running state, and then create the pod for mpimaster..
+
+* Suppose there are two tasks and task 2 needs to use the calculation results of task 1,in this case, task1 needs to be in the complete state, and then create the pod for task2.
+
+* some ETL applications
```
Thor-wl, line 15: Can you give more details about ETL applcations?

docs/design/task-DAG-scheduling.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Requirement
+
+Each task can specify which one or more tasks it depends on, and it can specify the time point or interface for triggering.
+
+Examples of time points:
+
+* running
```
Thor-wl, line 23: IC, Are there any other time points? Suggest to list all possbible time points in order to take them into consideration when doing feature design.

docs/design/task-DAG-scheduling.md
```diff
@@ -0,0 +1,92 @@
...
+	Probe    *v1.Probe
+}
+```
+2.1 `taskName` is a string indicating the name of the task that the current task depends on
+
+2.2 `trigger` indicates the point in time when the trigger, it has two fields, `status` and `time`, `status` indicates the status of the task, `time` indicates the duration of time since the state was reached,it can be empty, indicating 0 seconds.
+
+```go
+type TriggerSpec struct {
+	Event string
```
Thor-wl, line 58: I think Event should also be a struct. It includes not only the task status but also the conditions list above.

docs/design/task-DAG-scheduling.md
Thor-wl: > yes, you are right, But I think maybe we can just use task to meet the minavailable time

That's an available solution.

docs/design/task-DAG-scheduling.md
Thor-wl: > If don't add fields, then how to specify the container? Do you have any suggestions?

To simplify the implementation, i think we can only consider one container per pod temporarily for most users take it into practice.

docs/design/task-DAG-scheduling.md
```diff
@@ -0,0 +1,92 @@
...
+    trigger: 
+      event: running
+      time: 10s
+  - taskName: taskD
+    trigger:
+      event: running
+  - taskName: taskE
+    trigger:
+      event: complete
+```
```
Thor-wl, line 88: I think of another scenario. Not sure if it exists. If task A depends on task B and C，it may need to wait for both B and C compeletes or either B or C completes. How do you think about it? @shinytang6 @hwdef

hwdef: Both of these situations really need to be considered, and we should use a method to distinguish them.

docs/design/task-DAG-scheduling.md
hwdef: https://en.wikipedia.org/wiki/Extract,_transform,_load

https://baike.baidu.com/item/ETL/1251949?fr=aladdin

Here are some links about ETL. Simply put, ETL represents the three steps of processing data, and there is a dependency between these three steps, so we need a specific order to start the task.

docs/design/task-DAG-scheduling.md
hwdef: I can't think of another time pointes at the moment.

docs/design/task-DAG-scheduling.md
Thor-wl: > Both of these situations really need to be considered, and we should use a method to distinguish them.

So we need a field at the same level with `dependon` to indicates the way to watch the status of previous tasks

hwdef:
@shinytang6 @Thor-wl 
I updated this docs. Please review it again.

Thor-wl:
> @shinytang6 @Thor-wl
> I have updated this docs. Please review it again.

Got it. I'll review it again.

docs/design/task-leve-advanced-scheduling-policy.md
```diff
@@ -0,0 +1,98 @@
...
+  * Imediately after task completed
+  * Half an hour afater task starts to run or completes
+* Besides the time period, user may specify an certain condition as the triggering policy, for example,
+  * Check if a specified file exists in the /data directory.
+  * Check if the specified port is open, such as 2222.
+  * Check if the HTTP GET request to the specified path was successfully accessed, such as /health.
+  
+## Design
+
+1. Add a field that is an array named `dependsOn` in task spec
```
Thor-wl, line 52: In order to achieve the following goals:
* keep job api brief and focus on its basic functions
* separate task scheduling logic from job
* make scheduling logic abstract and can be reused

I make the following suggestions:
* create a new CRD `TaskFlow`, it indicates the start order of tasks within a job
* divide the scenarios into two categories, namely, Dag and StartOrder. Dag means the following task starts depending on the status of the previous task. StartOrder means the following task starts depending on some condition custom self-defined such as health check/liveness check/target output. New api should take them into consideration

Thor-wl:
CI has been fixed and please pull and rebase from master branch to make CI happy.
You can execute `git pull --rebase origin/master` to do that. @hwdef

docs/design/task-leve-advanced-scheduling-policy.md
```diff
@@ -0,0 +1,98 @@
+# Task-leve advanced scheduling policy
```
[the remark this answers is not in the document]
hwdef: Thanks for your suggestions. I changed the title.

docs/design/task-leve-advanced-scheduling-policy.md
hwdef: I agree with most. But for the definition of DAG, In previous community discussions, most people thought that DAG means that the previous task reaches the complete state before the next task can be executed? So we should not use the concept of DAG, because our definition of state includes `Complete`,`Running`and other states.

Thor-wl: > I agree with most. But for the definition of DAG, In previous community discussions, most people thought that DAG means that the previous task reaches the complete state before the next task can be executed? So we should not use the concept of DAG, because our definition of state includes `Complete`,`Running`and other states.

OK. the definition can be stressed.

Thor-wl (approved):

docs/design/task-launch-order-within-job.md
```diff
@@ -0,0 +1,98 @@
...
+  * task completes
+* User may specify the time period after the triggering event happens, for example
+  * Imediately after task completed
+  * Half an hour afater task starts to run or completes
+* Besides the time period, user may specify an certain condition as the triggering policy, for example,
+  * Check if a specified file exists in the /data directory.
+  * Check if the specified port is open, such as 2222.
+  * Check if the HTTP GET request to the specified path was successfully accessed, such as /health.
+  
+## Design
```
william-wang, line 50: 1.  Add the feature interaction with volcano existing feature.
2.  Is there any impact to existing feature? please add these info to the design doc.
3.  Pleaes add the desgin behavior (usage) for user in doc.

docs/design/task-launch-order-within-job.md
```diff
@@ -0,0 +1,104 @@
...
+      action: ['ls', '/data']
+      iteration: any/all
+```
+
+### Usage
+* create taskflow `task-flow`
+* create volcano job `vc-job`, spec.taskflow is the taskflow name, make sure task names corresponding to taskflow
+* get task status to check if it is correct order
+
+### Implementaion
```
Thor-wl, line 90: I'm ok about the CRD design. We need more discussion details about the implementation. Currently I'm not quite sure about add a new action or implement a new component dealing with taskflow. @william-wang @shinytang6 Do you have any ideas?

docs/design/task-launch-order-within-job.md
```diff
@@ -0,0 +1,104 @@
...
+* generate new CRD **TaskFlow**
+* create admission for TaskFlow, VCJob adds new fields
+* overwrite `allocate` action, name as a new action
+
+### Notice
+* deal with the conflict with gang-scheduling
```
shinytang6, line 96: This should also be clarified, how to avoid conflicts with gang scheduling. And if we need to overwrite `allocate`, what's the basic workflow?

shinytang6 (changes requested):

docs/design/task-launch-order-within-job.md
hwdef: abount gang, we discussed two options: 
1. task level gang scheduling
2. taskflow and gang do not coexist
abount allocate: 
1. create a new action
2. create a new component

We can discuss the above options specifically in our weekly meetings.

docs/design/task-launch-order-within-job.md
```diff
@@ -0,0 +1,113 @@
...
+  
+## Design
+
+### API
+taskflow example
+```
+apiVersion: batch.volcano.sh/v1alpha1
+kind: TaskFlow
+metadata:
+  name: task-flow
```
k82cn, line 58: task or job?

k82cn: if job, how to handle MPI job issue?

hwdef: I think it is `taskflow`.
if mpi, we can create a taskflow like this:

```yaml
kind: TaskFlow
metadata:
  name: task-flow
spec:
  tasks:
  - name: worker
  - name: master
    depends: 
      target: ['worker']
      probe:
        taskStatus: Running

```

k82cn: if so, why not just add `Depends` to volcano job?

hwdef: We think this taskflow can be reused by other vcjob.
@Thor-wl Can you join this discussion?

docs/design/task-launch-order-within-job.md
```diff
@@ -0,0 +1,113 @@
+# Task Launch Order Within Job
+
+@[hwdef](https://github.com/hwdef), @[Thor-wl](https://github.com/Thor-wl); July 19th, 2021
+
+## Introduction
+
+This feature provides the ability to customize the order in which tasks are launched.
```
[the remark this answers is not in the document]
hwdef: I added the scope, please review it again.

docs/design/task-launch-order-within-job.md
k82cn: > We think this taskflow can be reused by other vcjob.

Why reuse? If Job_1 has task A, B, C and Job_2 has task B, C, D, how can you sure those two job works as expected?

docs/design/task-launch-order-within-job.md
k82cn: Please see my comments in the other thread. This CRD may make the execution of different jobs confused.

docs/design/task-launch-order-within-job.md
hwdef: If we want to reuse a taskflow, then we need to satisfy the following two conditions:

* vcjob's taskflow field value is the name of the taskflow.
* The task in vcjob has the same name as the task in taskflow

for example:
```
jobA: 
task: taskA, taskB, taskC
taskflow: taskflowA

jobB: 
task: taskA, taskB,taskC
taskflow: taskflowA

taskflowA: 
task: taskA, taskB, taskC
```
In this case, both jobA and jobB can use taskflowA

hwdef:
@william-wang @Thor-wl @kevin-wangzefeng 

I update the design doc, please review it again.

docs/design/task-launch-order-within-job.md
```diff
@@ -0,0 +1,79 @@
...
+* Task-A depends on task-B, which means A must be started first and then B.
+* Triggering policy, in our cases, there may be only one trigger policy, which is the running state
+
+For the ease of end user's experince, we need to unify the way of composing a complicated jobs with lots of tasks, instead of letting user handle the complexcity themselvs using init-containers or other workflow tools. So we need to  have an more advanced VCJob that has below abilities:
+
+## Design
+
+### Field
+
+Add a field in `vcjob.spec.task.DependentTask`, it represents the name of the task that this task needs to depend on
```
k82cn, line 47: let's add an object instead of a field which is more extendable :)

hwdef: @k82cn 
After yesterday’s discussion, I think that several other reviewers do not want the possibility of expansion of this feature.Currently only the running state is retained. If a field has only one value, then the field should not exist.

```
dependency:
   name: taskA
   status: running
```
For example, in an example like the one above, there is no reason for the status field to exist

k82cn: > After yesterday’s discussion, I think that several other reviewers do not want the possibility of expansion of this feature.

Extensibility is important! It's ok to support only one task right now, but it's better to be extendable, something like that:

```
depends:
   name: taskA
```

hwdef: 1. we should support multi tasks.
2. In yesterday's meeting, I felt that other reviewers did not pay attention to scalability. Please unify internally
3. `depends` just have one field named `name`, this kind of design is an over-design.Yesterday, we made it clear that we will only solve existing problems and not consider future scenarios.

ZhangZhenhua:
Task level DAG scheduling is giant feature for Volcano, I agree that we should give more extra thoughts about it.

From my point of view, I think we may first answer one basic question: Do we need to add task level dependency in VCJob?

If no, then the current behavior is OK, pods of one VCJob are created out of order. 

If yes, we should design it. And the current design of adding only one `dependsOn` field is not enough.

Thor-wl:
> Task level dependency scheduling is giant feature for Volcano, I agree that we should give more extra thoughts about it.
> 
> From my point of view, I think we may first answer one basic question: Do we need to add task level dependency in VCJob?
> 
> If no, then the current behavior is OK, pods of one VCJob are created out of order.
> 
> If yes, we should design it. And the current design of adding only one `dependsOn` field is not enough.

I agree with @ZhangZhenhua most. Obviously, task dependency is valuable in applications. And we should design it. We should leave space for furture expansion such as custom detection means.

ZhangZhenhua:
> > Task level dependency scheduling is giant feature for Volcano, I agree that we should give more extra thoughts about it.
> > From my point of view, I think we may first answer one basic question: Do we need to add task level dependency in VCJob?
> > If no, then the current behavior is OK, pods of one VCJob are created out of order.
> > If yes, we should design it. And the current design of adding only one `dependsOn` field is not enough.
> 
> I agree with @ZhangZhenhua most. Obviously, task dependency is valuable in applications. And we should design it. We should leave space for furture expansion such as custom detection means.

I talked to some customers recently, some of them are using Volcano right now, and almost all of the Volcano users are using Argo(or something else similar) together with Volcano. I think this is a clear evidence that volcano can't handle complex Jobs/tasks alone. 

If this is the right way to go, I think it's OK.

If we just want to add some basic dependency into VCJob, like starting tasks sequentially, we can just add one annotation to handle it and leave the current job.spec unchanged.

hwdef:
And I do not understand. Why reviewers are very cautious about adding a crd, but very open to introducing new projects. The introduction of argo and kubegenen projects will introduce many crd, at least more than one, but this is the recommended approach. I don’t know if there is a unified standard for crd. 

Are we looking at problems from the perspective of Volcano or from the perspective of large cloud vendors?

@william-wang @kevin-wangzefeng

docs/design/task-launch-order-within-job.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Scenarios
+
+* MPI Job. The worker pods need to be started first and then the master pod can run. Elsewise, the master pod can't setup the ssh tunnel and thus failed, this will add unnecessary waste of resources. In this case, MPI worker pods need to be in the running state, and then the master pod can start. Very similary, for the TensorFlow job, the TF master pod must be started first and then the TF worker pods.
+
+## Scene Comparison
+
+| example   | number of tasks | task dependencies | concurrent execution | Solution Status   | Disadvantages of the current solution |
+| ---------- | -------- | ------------ | -------- | -------------------------------------------------- | -------------------------- |
+| MPI        | 2        | linear dependencies | yes    | 1. Add initcontainer to the task<br/>2.Using user-written check scripts in initcontainer | 1.Increase the cost of use for users<br />2.Still using resources while waiting |
```
eggiter, line 30: It'd be better that VCJob provides a convenient image for initContainer and maintains the related codes like MPI-Operator does.

hwdef: There are many limitations to using initcontainer, For example, for windows cluster support, and support for multiple detection methods will not be very convenient. I think add initcontainer for pod is not a good way.

hwdef:
I update the API, please review this again, thanks!
@william-wang @Thor-wl @k82cn @shinytang6

PR #1636, opened 2021-07-21
delete bindingTasks from NodeInfo structure

huone1:
in  #1388 , the nodeinfo add  the member bindingTasks to avoid the resource overuse.
But  before bind action，schedulercache has reduced the pod request resource on the binding node. 
image

So the adjustment whether the binding is still in progress is unnecessary and it has some impact on binding performance due to the need for locks.


wpeng102:
@zen-xu as we talked offline last week. `Node.AddTask(task)` should reduced the resource from cache.  Please help review this fix.

zen-xu:
I am OK with that

jasonliu747 (commented):
/lgtm

PR #1653, opened 2021-07-31
optimize function setOversubscription

huone1:
if argment node is null , There's going to be a crashoff here.


pkg/scheduler/api/node_info.go
```diff
@@ -185,6 +185,10 @@ func (ni *NodeInfo) setRevocableZone(node *v1.Node) {
 // Check node if enable Oversubscription and set Oversubscription resources
 // Only support oversubscription cpu and memory resource for this version
 func (ni *NodeInfo) setOversubscription(node *v1.Node) {
```
Thor-wl, line 187: BTW, if the function mainly focuses on check node, why not return a boolen?

william-wang (approved):
/lgtm
/approve

pkg/scheduler/api/node_info.go
huone1: it is not care of the return

PR #1656, opened 2021-08-02
fix addmission parsing bug

hacker-qian:
Signed-off-by: jianmin qian

background: in bash scripts, ";" also used as a delimiter. Thus if we use ";" to split a string, such as "/jobs/mutate;/jobs/validate;/podgroups/mutate;/pods/validate", we will just get "/jobs/mutate" as a result. In there, we change ";" to ","

### DIFF
--- a/pkg/webhooks/router/admission.go
+++ b/pkg/webhooks/router/admission.go
@@ -49,7 +49,7 @@ func RegisterAdmission(service *AdmissionService) error {
 }
 
 func ForEachAdmission(config *options.Config, handler func(*AdmissionService)) {
-	admissions := strings.Split(config.EnabledAdmission, ";")
+	admissions := strings.Split(strings.TrimSpace(config.EnabledAdmission), ",")
 	for _, admission := range admissions {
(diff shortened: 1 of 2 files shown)


pkg/webhooks/router/admission.go
```diff
@@ -49,7 +49,7 @@ func RegisterAdmission(service *AdmissionService) error {
 }
 
 func ForEachAdmission(config *options.Config, handler func(*AdmissionService)) {
-	admissions := strings.Split(config.EnabledAdmission, ";")
+	admissions := strings.Split(config.EnabledAdmission, ",")
```
shinytang6, line 52: one more suggestion: it's better to use `strings.TrimSpace` remove extra spaces first.

hacker-qian: done

wpeng102 (approved):
/lgtm

shinytang6 (approved):
/lgtm

jasonliu747:
Hi @hacker-qian , please make CI happy.

PR #1659, opened 2021-08-04
fix some problems of golangci-lint

gy95:

modify some codes of golangci-lint
we can use `make lint` command to run golangci-lint tools to check codes.
And too much codes to run all linters, so I only choose `gofmt` `goimports` `golint` `govet` linters to check codes, `goimports` put imports beginning with `volcano.sh` (packages of volcano.sh)after 3rd-party packages;

We can enable more linters to check codes to make codes robuster.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,26 @@
+# options for analysis running
+run:
+  # default concurrency is a available CPU number
+  concurrency: 4
+
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  timeout: 10m
+
+  # exit code when at least one issue was found, default is 1
+  issues-exit-code: 1
+
+  # which dirs to skip: issues from them won't be reported;
+  # can use regexp here: generated.*, regexp is applied on full path;
+  # default value is empty list, but default dirs are skipped independently
+  # from this option's value (see skip-dirs-use-default).
+  # "/" will be replaced by current OS file path separator to properly work
+  # on Windows.
+  skip-dirs:
+    - vendor
+    - test
+  skip-files:
+    - .*_test.go
+
 # output configuration options
 output:
   # colored-line-number|line-number|json|tab|checkstyle|code-climate, default is "colored-line-number"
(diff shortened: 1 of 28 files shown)


gy95:
how to fix this e2e error? I just adjust the import package sequence, and  don't modify code logic. Could you please give me some advice? @shinytang6 @Thor-wl

shinytang6:
> how to fix this e2e error? I just adjust the import package sequence, and don't modify code logic. Could you please give me some advice? @shinytang6 @Thor-wl

Yes, I think CI failure is not caused by your changes. E2E testing had some problems which troubled us a lot

Thor-wl:
> how to fix this e2e error? I just adjust the import package sequence, and don't modify code logic. Could you please give me some advice? @shinytang6 @Thor-wl

CI may fails at some cases due to CI machine performance. I'll rerun those jobs.

gy95:
> > how to fix this e2e error? I just adjust the import package sequence, and don't modify code logic. Could you please give me some advice? @shinytang6 @Thor-wl
> 
> CI may fails at some cases due to CI machine performance. I'll rerun those jobs.

Ok, and all ci checks have passed. 👍

pkg/controllers/job/plugins/factory.go
```diff
@@ -20,7 +20,7 @@ import (
 	"sync"
 
 	"volcano.sh/volcano/pkg/controllers/job/plugins/env"
-	"volcano.sh/volcano/pkg/controllers/job/plugins/interface"
+	pluginsinterface "volcano.sh/volcano/pkg/controllers/job/plugins/interface"
```
Thor-wl, line 23: If alias, is there any related code modification?

hwdef:
I don't understand what bad consequences would result from running all the tests.

By the way, https://github.com/golang/lint has been archived, maybe we can use `staticcheck` replace it.

pkg/controllers/job/plugins/factory.go
gy95: the package name is `pluginsinterface` and the last part of import path `volcano.sh/volcano/pkg/controllers/job/plugins/interface` is `interface`, not equal to package name `pluginsinterface`. So we should use alias. If not, goimports will failed with information File is not `goimports`-ed with -local volcano.sh (goimports)

jasonliu747, line 23: Modifying `pkg/controllers/job/plugins/factory.go#L40` might be a better choice.

gy95: no, this is `goimports` specification, 
the package name is `pluginsinterface` and the last part of import path `volcano.sh/volcano/pkg/controllers/job/plugins/interface` is `interface`, the two are not equal. 
On this senario, we should use alias. 
If not, goimports will failed with information: 
File is not `goimports`-ed with -local volcano.sh (goimports)

shinytang6:
> I don't understand what bad consequences would result from running all the tests.
> 
> By the way, https://github.com/golang/lint has been archived, maybe we can use `staticcheck` replace it.

@gy95 what do you think about it? l think it make sense, btw we can also refer to [test-infra](https://github.com/kubernetes/test-infra/blob/master/.golangci.yml)

gy95:
> > I don't understand what bad consequences would result from running all the tests.
> > By the way, https://github.com/golang/lint has been archived, maybe we can use `staticcheck` replace it.
> 
> @gy95 what do you think about it? l think it make sense, btw we can also refer to [test-infra](https://github.com/kubernetes/test-infra/blob/master/.golangci.yml)

Yes, now that `golang/lint` has been archived, we can use `golangci-lint` to replace it, And there are lots of go projects being using `golangci-lint` tool to do golang code checks, And this tools is much more powerful.

.golangci.yml
```diff
@@ -9,80 +32,35 @@ output:
...
-    - deadcode
-    - errcheck
-    - gocritic
+    # linters maintained by golang.org
     - gofmt
-    - goimports 
-    - golint 
-    - gosimple
+    - goimports
+    - golint
```
hwdef, line 46: why keep this archived tools?

.golangci.yml
```diff
@@ -9,80 +32,35 @@ output:
...
-    - interfacer
-    - lll
-    - misspell
-    - staticcheck 
-    - structcheck 
-    - typecheck 
-    - unconvert 
-    - unparam 
-    - varcheck 
+    # linters default enabled by golangci-lint .
```
hwdef, line 48: If we want to use golangci-lint, why comment out?

gy95:
> > I don't understand what bad consequences would result from running all the tests.
> > By the way, https://github.com/golang/lint has been archived, maybe we can use `staticcheck` replace it.
> 
> Does anyone have an answer to this question? I don't understand the significance of this PR.

just comply with go programming specifications, but some checks may find some fatal problems, like `gosec` check

.golangci.yml
gy95: yes, can be removed

.golangci.yml
gy95: linters are too much, if enable all, we should modify more codes, so just first enable some linters, we could open another PR to enable other linters

hwdef: ok, I see.

.golangci.yml
gy95: done

shinytang6 (approved):
Thanks
/lgtm

Issue #1658, opened 2021-08-03
svc plugin produces invalid envs


d4l3k:
**What happened**:

If you have a task with a dash in the name i.e. `trainer-0` when used with the svc plugin it creates environement variables with dashes in them. Such as `VC_TRAINER-0_HOSTS`. These _can_ be used in some languages but can't be used in standard compliant shells such as bash.

https://stackoverflow.com/questions/36989263/why-cant-environment-variables-with-dashes-be-accessed-in-bash-4-1-2

**What you expected to happen**:

That the environment variables can be used from bash via `$VC_TRAINER_0_HOSTS`.

**How to reproduce it (as minimally and precisely as possible)**:
```yaml
spec:
  plugins:
    env: []
    svc: []
  queue: test
  schedulerName: volcano
  tasks:
    replicas: 1
    template:
      spec:
        containers:
        - command:
          - /bin/sh
          - -c
          - '"echo" "$(env)" "$VC_SH-0_NUM"'
          image: alpine:latest
          name: sh-0
```

```bash
$ kubectl logs sh-28rq2-sh-0-0
HOSTNAME=sh-28rq2-sh-0-0
SHLVL=1
VC_SH-0_NUM=1
VC_TASK_INDEX=0
VC_SH-0_HOSTS=sh-28rq2-sh-0-0.sh-28rq2
VK_TASK_INDEX=0
... -0_NUM
```

This produces `"-0_NUM"` since only the first part is substituted due to - not being a valid env character.

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
    image: volcanosh/vc-controller-manager:latest
    imageID: docker-pullable://volcanosh/vc-controller-manager@sha256:897793abe922641b47acc2c872209078346e855a3453a0d35dfb573812ae64b9

- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.3", GitCommit:"ca643a4d1f7bfe34773c74f79527be4afd95bf39", GitTreeState:"archive", BuildDate:"2021-07-16T17:16:46Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #1660, opened 2021-08-04
format env key

wpeng102:
Fix issue https://github.com/volcano-sh/volcano/issues/1658


wpeng102:
If you have a task with a dash in the name i.e. trainer-0 when used with the svc plugin it creates environement variables with dashes in them. Such as VC_TRAINER-0_HOSTS. These can be used in some languages but can't be used in standard compliant shells such as bash.


task name with "-"
image

pod env:
image

shinytang6 (approved):
/lgtm
/approve

PR #1666, opened 2021-08-10
add setting MinResources to pg for normal pod

huone1:
fix issue #1542 
there are two problem:
1. minresource in podgroup for deployment is empty
2. the podgroup init status is "" and volcano scheduler doesn't adjust whether the status is "", So it isgoing to cause a lot of problems. such as queue capability failure

### DIFF
--- a/installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
+++ b/installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
@@ -6,9 +6,9 @@ tiers:
   - name: conformance
   - name: sla
 - plugins:
+  - name: overcommit
   - name: drf
   - name: predicates
   - name: proportion
   - name: nodeorder
   - name: binpack
-  - name: overcommit
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -92,6 +92,7 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
 			Spec: scheduling.PodGroupSpec{
 				MinMember:         1,
 				PriorityClassName: pod.Spec.PriorityClassName,
+				MinResources:      calcPGMinResources(pod),
 			},
 		}
 		if queueName, ok := pod.Annotations[scheduling.QueueNameAnnotationKey]; ok {
@@ -140,4 +141,43 @@ func newPGOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
 	}
 	ref := metav1.NewControllerRef(pod, gvk)
 	return []metav1.OwnerReference{*ref}
+}
+
+// addResourceList add list resource quantity
+func addResourceList(list, req, limit v1.ResourceList) {
+	for name, quantity := range req {
+
+		if value, ok := list[name]; !ok {
+			list[name] = quantity.DeepCopy()
+		} else {
+			value.Add(quantity)
+			list[name] = value
+		}
+	}
+
+	if req != nil {
+		return
+	}
+
+	// If Requests is omitted for a container,
+	// it defaults to Limits if that is explicitly specified.
+	for name, quantity := range limit {
+		if value, ok := list[name]; !ok {
+			list[name] = quantity.DeepCopy()
+		} else {
+			value.Add(quantity)
+			list[name] = value
+		}
+	}
+}
+
+// calcPGMinResources calculate podgroup minimum resource
+func calcPGMinResources(pod *v1.Pod) *v1.ResourceList {
+	pgMinRes := v1.ResourceList{}
+
+	for _, c := range pod.Spec.Containers {
+		addResourceList(pgMinRes, c.Resources.Requests, c.Resources.Limits)
+	}
+
+	return &pgMinRes
 }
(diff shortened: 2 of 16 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -58,7 +58,9 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	jobsMap := map[api.NamespaceName]map[api.QueueID]*util.PriorityQueue{}
 
 	for _, job := range ssn.Jobs {
-		if job.PodGroup.Status.Phase == scheduling.PodGroupPending {
+		if job.PodGroup.Status.Phase == scheduling.PodGroupPending || job.PodGroup.Status.Phase == "" {
```
shinytang6, line 61: l believe there are more code having this issue, pls help check it out.
and it's better to abstract a func to do that

shinytang6 (changes requested):

pkg/scheduler/actions/allocate/allocate.go
huone1: OK, define the function "jobinfo.IsPending() bool" to check it.

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -66,6 +66,10 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 			queues.Push(queue)
 		}
 
+		if job.PodGroup.Status.Phase == "" {
+			continue
+		}
+
 		if job.PodGroup.Status.Phase == scheduling.PodGroupPending {
```
william-wang, line 73: Use job.IsPending() to replace 'job.PodGroup.Status.Phase == scheduling.PodGroupPending' here.

huone1: don't do it, because the pending job  should be  pushed  to jobsMap.

shinytang6 (approved):
/lgtm

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -65,6 +65,9 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 			queueMap[queue.UID] = queue
 			queues.Push(queue)
 		}
+		if job.PodGroup.Status.Phase == "" {
```
Thor-wl, line 68: Suggest to add some comments here in order not to make confusion for developers later.

pkg/scheduler/api/job_info.go
```diff
@@ -589,3 +589,16 @@ func (ji *JobInfo) Ready() bool {
 
 	return occupied >= ji.MinAvailable
 }
+
+// IsPending returns whether job is in pending status
+func (ji *JobInfo) IsPending() bool {
```
Thor-wl, line 594: Maybe it can be merged with the `if` below.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -263,9 +262,14 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		}
 
 		if job.PodGroup.Spec.MinResources == nil {
+			klog.V(4).Infof("job %s MinResources is null.", job.Name)
 			return util.Permit
 		}
 		minReq := job.GetMinResources()
+
+		klog.V(4).Infof("job %s min resource<%v>", job.Name, minReq)
```
Thor-wl, line 270: These logs can be removed which are just for test

huone1: not only for test，but also for debug when the plugin behavior doesn't meet expectations. And the log level is changed to 5

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -141,3 +142,42 @@ func newPGOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
...
+			list[name] = quantity.DeepCopy()
+		} else {
+			value.Add(quantity)
+			list[name] = value
+		}
+	}
+}
+
+// calcPGMinResources calculate podgroup minimum resource
+func calcPGMinResources(pod *v1.Pod) *v1.ResourceList {
```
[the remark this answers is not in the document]
huone1: the request resource of BestEffort  pod is empty  and minresource is empty, so there is no limit about the resource.

pkg/scheduler/actions/enqueue/enqueue.go
huone1: it  is  modified to `if job.IsPending()` with the if below

pkg/scheduler/api/job_info.go
huone1: merged it

jasonliu747:
/hold
@huone1 let us know when you feel ready~

shinytang6 (changes requested):

installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
```diff
@@ -6,9 +6,9 @@ tiers:
   - name: conformance
   - name: sla
 - plugins:
+  - name: overcommit
```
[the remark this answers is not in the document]
huone1: plugin overcommit just  has the function jobEnqueuedFns , it shouldn't have any conflict with reclaim or preempt

Thor-wl (approved):

jasonliu747 (commented):
/lgtm

PR #1668, opened 2021-08-11
fix(interpodaffinity): make 'existing pods anti-affinity rules' work

eggiter:
1. The "existing pods anti-affinity rules" of interpodaffinity in plugin `predicates` dose not work now;
2. The rules is using `snapshot.HavePodsWithRequiredAntiAffinityList` but this method return nil now;
3. This commit fix this problem by syncing code from https://github.com/kubernetes/kubernetes/blob/v1.19.6/pkg/scheduler/internal/cache/snapshot.go

### DIFF
--- a/pkg/scheduler/plugins/util/k8s/snapshot.go
+++ b/pkg/scheduler/plugins/util/k8s/snapshot.go
@@ -25,6 +25,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+
 	"volcano.sh/volcano/pkg/scheduler/plugins/util"
 )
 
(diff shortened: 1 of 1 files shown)


eggiter:
> /assign @zen-xu

/assign @zen-xu

pkg/scheduler/plugins/util/k8s/snapshot.go
```diff
@@ -22,10 +22,11 @@ package k8s
 import (
 	"fmt"
 
+	"volcano.sh/volcano/pkg/scheduler/plugins/util"
```
shinytang6, line 25: please format the import and fix the DCO

shinytang6 (changes requested):

Thor-wl (commented):
Please fix the DCO

pkg/scheduler/plugins/util/k8s/snapshot.go
eggiter: sure

shinytang6 (approved):
/approve

Thor-wl:
Thanks for your contribution! @eggiter 
/lgtm

jasonliu747 (commented):
/lgtm
Welcome to Volcano!

PR #1673, opened 2021-08-12
fix(scheduler): modify default scheduler cfg

shinytang6:
Supplement to  https://github.com/volcano-sh/volcano/pull/1666
/cc @huone1


huone1 (commented):
/lgtm

PR #1637, opened 2021-07-21
fix overused judgement when deal with allocate and proportion

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/1425
Signed-off-by: Thor-wl <13164644535@163.com>

PR #1676, opened 2021-08-17
add new fn for overused modifications

Thor-wl:
separate the modifications for https://github.com/volcano-sh/volcano/pull/1637

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -187,6 +187,26 @@ func (r *Resource) Add(rr *Resource) *Resource {
 	return r
 }
 
+// AddBaseLeft is used to add two given resources based on the resource list of Resource r
+func (r *Resource) AddBaseLeft(rr *Resource) *Resource {
+	r.MilliCPU += rr.MilliCPU
+	r.Memory += rr.Memory
+
+	if rr.ScalarResources == nil {
+		return r
+	}
+
+	for rName := range r.ScalarResources {
+		if _, ok := rr.ScalarResources[rName]; !ok {
+			continue
+		}
+
+		r.ScalarResources[rName] += rr.ScalarResources[rName]
+	}
+
+	return r
+}
+
 //Sub subtracts two Resource objects.
 func (r *Resource) Sub(rr *Resource) *Resource {
 	assert.Assertf(rr.LessEqual(r, Zero), "resource is not sufficient to do operation: <%v> sub <%v>", r, rr)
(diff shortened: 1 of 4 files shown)


pkg/scheduler/api/resource_info.go
```diff
@@ -187,17 +187,38 @@ func (r *Resource) Add(rr *Resource) *Resource {
 	return r
 }
 
+// AddBaseLeft is used to add two given resources based on the resource list of Resource r
+func (r *Resource) AddBaseLeft(rr *Resource) *Resource {
```
k82cn, line 191: What's the different between `Add` and `AddBaseLeft`?

pkg/scheduler/api/types.go
```diff
@@ -168,3 +168,9 @@ type ReservedNodesFn func()
 
 // VictimTasksFn is the func declaration used to select victim tasks
 type VictimTasksFn func() []*TaskInfo
+
+// QueueOverusedFn is the func declaration used to judge whether queue is overused.If overused, whether it is all dimensions overused.
+type QueueOverusedFn func(*QueueInfo) (bool, bool)
```
k82cn, line 173: We should not return two bool for an API.

pkg/scheduler/api/types.go
```diff
@@ -168,3 +168,9 @@ type ReservedNodesFn func()
 
 // VictimTasksFn is the func declaration used to select victim tasks
 type VictimTasksFn func() []*TaskInfo
+
+// QueueOverusedFn is the func declaration used to judge whether queue is overused.If overused, whether it is all dimensions overused.
```
k82cn, line 172: What does `If overused, whether it is all dimensions overused.` mean?

pkg/scheduler/api/resource_info.go
Thor-wl: @huone1

pkg/scheduler/api/types.go
Thor-wl: Because there exists two different scenario:

pkg/scheduler/api/types.go
Thor-wl: Considering queue.deserved = {cpu:1, memory:1, gpu:1}, queue.allocated={cpu:2, memory:2, gpu:1}, so this queue is overused, but not overused at all dimensions.

pkg/scheduler/api/resource_info.go
huone1: if a resource is not in r but in rr,  AddBaseLeft function will skip the  add operation of the resource; Add function will set the resource in r to zero and add them。

pkg/scheduler/api/types.go
k82cn: find another way for that case; it's hard for user to understand the differentiation between those two return value.

Thor-wl: Another way is to separate it into two functions with different targets, but there exists similar function logic.

pkg/scheduler/api/resource_info.go
william-wang: Got you point. The logic is a little trick. Let's think whether there is a better idea for this.
AddbaseLeft function
input:
r cpu 1
rr cpu1 gpu1
output:
r cpu: 2

Add function
input:
r: cpu 1
rr: cpu1 gpu1
output:
r cpu: 2, gpu: 1

PR #1677, opened 2021-08-17
enable gosec and fix related errors

gy95:

enable gosec linter check
the error output is ` G601: Implicit memory aliasing in for loop. (gosec)`
in the `for range` loop, we should address the value directly, maybe the address points to the same address

ref: [using-reference-to-loop-iterator-variable](https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable)

### DIFF
--- a/example/kubecon-2019-china/scripts/node-info.go
+++ b/example/kubecon-2019-china/scripts/node-info.go
@@ -61,7 +61,8 @@ func main() {
 		if _, found := podMap[nodeName]; !found {
 			podMap[nodeName] = api.EmptyResource()
 		}
-		res := api.GetPodResourceWithoutInitContainers(&p)
+		pod := p
+		res := api.GetPodResourceWithoutInitContainers(&pod)
 		podMap[nodeName].Add(res)
 	}
 
--- a/pkg/scheduler/plugins/numaaware/numaaware.go
+++ b/pkg/scheduler/plugins/numaaware/numaaware.go
@@ -127,7 +127,8 @@ func (pp *numaPlugin) OnSessionOpen(ssn *framework.Session) {
 		taskPolicy := policy.GetPolicy(node, numaNodes[node.Name])
 		allResAssignMap := make(map[string]cpuset.CPUSet)
 		for _, container := range task.Pod.Spec.Containers {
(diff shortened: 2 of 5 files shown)


pkg/scheduler/plugins/numaaware/numaaware.go
```diff
@@ -136,7 +137,7 @@ func (pp *numaPlugin) OnSessionOpen(ssn *framework.Session) {
 
 			klog.V(4).Infof("[numaaware] hits for task %s container '%v': %v on node %s, besthit: %v",
 				task.Name, container.Name, providersHints, node.Name, hit)
-			resAssignMap := policy.Allocate(&container, &hit, node.NumaSchedulerInfo, resNumaSets, pp.hintProviders)
+			resAssignMap := policy.Allocate(&con, &hit, node.NumaSchedulerInfo, resNumaSets, pp.hintProviders)
```
Thor-wl, line 140: So why new thest variables?

gy95: for-range is not recommended reference value directly, so new variable to fix lint checks.

jasonliu747:
Seems a little bit not necessary to me, but...
image

jasonliu747 (commented):
Seems a little bit not necessary to me, but...
image

pkg/scheduler/plugins/numaaware/numaaware.go
Thor-wl: Well, thanks for the advice. Maybe we can pick some other bug fix which is more valuable next time.

example/kubecon-2019-china/scripts/node-info.go
```diff
@@ -61,7 +61,8 @@ func main() {
 		if _, found := podMap[nodeName]; !found {
 			podMap[nodeName] = api.EmptyResource()
 		}
-		res := api.GetPodResourceWithoutInitContainers(&p)
+		pod := p
```
k82cn, line 64: honestly, I don-t think that's necessary fix for this lint :(

hzxuzhonghu (commented):
Overall, i donot think it is worth  to introduce gosec

PR #1678, opened 2021-08-17
add revive linter as drop-in replacement of archived golint

gy95:

ref: https://golangci-lint.run/usage/linters/  , golint has been archived by the owner. Replaced by revive.
So add `revive` as drop-in replacement of archived `golint`


shinytang6 (approved):
/lgtm
/approve

Issue #1606, opened 2021-07-09
Suggest the sshKey access permission to a smaller value


ChuckGithub:
Excessive access permissions may increase security risks.
![image](https://user-images.githubusercontent.com/24219758/125051898-ffd90280-e0d5-11eb-9dc6-991e52d2831e.png)

Related Version:
Kubernetes: 1.17.8
Volcano: 1.2.0

OS:
Ubuntu 18.04.5

PR #1682, opened 2021-08-18
lower ssh key access permission

hwdef:
fix: #1606

### DIFF
--- a/pkg/controllers/job/plugins/ssh/ssh.go
+++ b/pkg/controllers/job/plugins/ssh/ssh.go
@@ -148,7 +148,7 @@ func (sp *sshPlugin) mountRsaKey(pod *v1.Pod, job *batch.Job) {
 	}
 
 	if sp.sshKeyFilePath != SSHAbsolutePath {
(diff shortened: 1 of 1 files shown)


pkg/controllers/job/plugins/ssh/ssh.go
```diff
@@ -148,7 +148,7 @@ func (sp *sshPlugin) mountRsaKey(pod *v1.Pod, job *batch.Job) {
 	}
 
 	if sp.sshKeyFilePath != SSHAbsolutePath {
-		var noRootMode int32 = 0755
+		var noRootMode int32 = 0600
```
shinytang6, line 151: have you tested for that?

shinytang6 (changes requested):

pkg/controllers/job/plugins/ssh/ssh.go
hwdef: image

hzxuzhonghu (approved):

Issue #1620, opened 2021-07-15
Show controller-manager version print “controller is registered”.


zishen:
Maybe this is only a  area of improvement.

![image](https://user-images.githubusercontent.com/7522079/125721201-68b54568-043f-4172-a36b-5b8bffab3014.png)

and this is happened in init()
![image](https://user-images.githubusercontent.com/7522079/125721312-566d79e3-6aaa-42e0-9ea7-22b885811649.png)

Can change him any better?

**Environment**:
- Volcano Version: v1.3.0
- Kubernetes version (use `kubectl version`): 1.17.3
- Cloud provider or hardware configuration: A800/9000
- OS (e.g. from /etc/os-release): kylin
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

william-wang:
@zishen Yes, The binary version, git SHA and built time are supposed to be enhanced to be shown.

PR #1683, opened 2021-08-18
Optimization register controller log

hwdef:
fix: #1620

### DIFF
--- a/pkg/controllers/framework/factory.go
+++ b/pkg/controllers/framework/factory.go
@@ -18,6 +18,7 @@ package framework
 
 import (
 	"fmt"
+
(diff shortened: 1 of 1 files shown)


shinytang6 (changes requested):
Could you also help enhance the version, git SHA, built time? They are `Not provided` now.

hwdef:
@shinytang6 

I used `make vc-controller-manager` to build bin, and I get the following results

image


So, I think the code about show git SHA and build time is correct.

jasonliu747 (approved):
/lgtm

shinytang6 (approved):
/lgtm
/approve

PR #1685, opened 2021-08-19
fix: fix possible panic when 'SetNode' is called

eggiter:
- This commit fixes possible panic when `SetNode` is called, which may happen when gpus of one machine are lost during the restarting of `nvidia-device-plugin`;
- Detailed reproduce procedures are provided in UT `TestNodeInfo_SetNode`;
- This commit is a supplement to MR #306 and issue #294.


pkg/scheduler/api/resource_info.go
```diff
@@ -187,10 +187,14 @@ func (r *Resource) Add(rr *Resource) *Resource {
 	return r
 }
 
-//Sub subtracts two Resource objects.
+// Sub subtracts two Resource objects with assertion.
```
Thor-wl, line 190: Although this modification is clear, I think the original function is also acceptable.

k82cn:
/hold

> This commit fixes possible panic when SetNode is called, which may happen when gpus of one machine are lost during the restarting of nvidia-device-plugin;

Is there any panic stack info for this issue?

pkg/scheduler/api/node_info.go
```diff
@@ -295,17 +295,34 @@ func (ni *NodeInfo) setNodeGPUInfo(node *v1.Node) {
...
 	if !ni.Ready() {
-		klog.Warningf("Failed to set node info, phase: %s, reason: %s",
-			ni.State.Phase, ni.State.Reason)
+		klog.Warningf("Failed to set node info for %s, phase: %s, reason: %s",
+			ni.Name, ni.State.Phase, ni.State.Reason)
+		return
+	}
+
+	// Dry run, make sure all fields other than `State` are in the original state.
+	copy := ni.Clone()
```
hzxuzhonghu, line 306: why clone?

eggiter:
+ Stack traces are listing as follows;
+ Detailed re-produce procedures can be found in UT `TestNodeInfo_SetNode`;
```
panic: resource is not sufficient to do operation: <cpu 5000.00, memory 5000000000.00> sub <cpu 6000.00, memory 6000000000.00> [recovered]
	panic: resource is not sufficient to do operation: <cpu 5000.00, memory 5000000000.00> sub <cpu 6000.00, memory 6000000000.00>

goroutine 28 [running]:
testing.tRunner.func1.1(0x278c940, 0xc00007f290)
	/usr/local/go/src/testing/testing.go:1072 +0x552
testing.tRunner.func1(0xc000182600)
	/usr/local/go/src/testing/testing.go:1075 +0x328
panic(0x278c940, 0xc00007f290)
	/usr/local/go/src/runtime/panic.go:975 +0x4c6
volcano.sh/volcano/pkg/scheduler/util/assert.Assert(0x29e9a00, 0xc00044c200, 0x77)
	/workspace/go/src/volcano.sh/volcano/pkg/scheduler/util/assert/assert.go:33 +0x205
volcano.sh/volcano/pkg/scheduler/util/assert.Assertf(0xc0004d6600, 0x29e9a8c, 0x39, 0xc0004e7898, 0x2, 0x2)
	/workspace/go/src/volcano.sh/volcano/pkg/scheduler/util/assert/assert.go:43 +0xa5
volcano.sh/volcano/pkg/scheduler/api.(*Resource).Sub(0xc0004d67a0, 0xc0004d6620, 0x0)
	/workspace/go/src/volcano.sh/volcano/pkg/scheduler/api/resource_info.go:192 +0x15e
volcano.sh/volcano/pkg/scheduler/api.(*NodeInfo).SetNode(0xc000318e70, 0xc0004ea300)
	/workspace/go/src/volcano.sh/volcano/pkg/scheduler/api/node_info.go:329 +0x5a5
volcano.sh/volcano/pkg/scheduler/api.TestNodeInfo_SetNode(0xc000182600)
	/workspace/go/src/volcano.sh/volcano/pkg/scheduler/api/node_info_test.go:222 +0xda5
testing.tRunner(0xc000182600, 0x2a50500)
	/usr/local/go/src/testing/testing.go:1123 +0x1a3
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1168 +0x648
```

pkg/scheduler/api/node_info.go
eggiter: To make make sure all fields are left untouched and in the original state.

hzxuzhonghu: Can you point where do we mutate this node?

hzxuzhonghu:
@eggiter we need to fix func (r *Resource) Sub(rr *Resource) *Resource {

hzxuzhonghu:
Amazing there is panic in the functiuon, should remove all the stuff

k82cn:
> Detailed re-produce procedures can be found in UT TestNodeInfo_SetNode;

AFAIK, we'll ignore not ready Nodes; is there any e2e case to trigger the panic?

eggiter:
> > Detailed re-produce procedures can be found in UT TestNodeInfo_SetNode;
> 
> AFAIK, we'll ignore not ready Nodes; is there any e2e case to trigger the panic?

- Yes, nodes which are not ready are already ignored. 
- But this situation here is:
  -  the node is **Ready** but the allocatable gpu becomes '0'. 
  - furthermore, this event triggers NodeUpdate event, which will call `SetNode` underneath and the panic occurred.

pkg/scheduler/api/node_info.go
eggiter: - The original codes are all changing fields of nodeInfo.
- Here, the `clone` i used, is merely a dry-run to make sure if all these changes makes this node not ready then will shall not to actually do this operation.

eggiter:
> @eggiter we need to fix func (r *Resource) Sub(rr *Resource) *Resource {

can you guys leave this fix to me? and you can tell me the idea of how to resolve.

eggiter:
@k82cn @hzxuzhonghu what's the current status?

k82cn:
> the node is Ready but the allocatable gpu becomes '0'.

We should not let's this happen :)

eggiter:
> > the node is Ready but the allocatable gpu becomes '0'.
> 
> We should ignore such nodes during scheduling cycle.

It will ignored by setting status of node to `NotReady`

Thor-wl:
I think such nodes can act as CPU nodes and continue to work if device plugin doesn't work well.

eggiter:
> I think such nodes can act as CPU nodes and continue to work if device plugin doesn't work well.

Sorry to tell you but, the machine can lost its memory occasionally. It'd be better to set status to `NotReady` and ignore in scheduling cycle.

Issue #1663, opened 2021-08-07
Install YAML: unknown flag: --admission-conf


5had3z:
**What happened**:

Unable to install volcano using yaml file.
```
kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/master/installer/volcano-development.yaml
```

```
kubectl get pods -n volcano-system
NAME                                   READY   STATUS             RESTARTS      AGE
volcano-admission-f5f4fb5b6-s68k8      0/1     CrashLoopBackOff   6 (44s ago)   6m31s
volcano-admission-init--1-6z7l5        0/1     Completed          0             6m31s
volcano-controllers-5bbcc9c49f-4tkjh   1/1     Running            0             6m30s
volcano-scheduler-5665cdc4d9-zbshx     1/1     Running            0             6m30s
```

```
kubectl logs volcano-admission-f5f4fb5b6-s68k8 -n volcano-system
unknown flag: --admission-conf
Usage of /vc-webhook-manager:
      --add-dir-header                   If true, adds the file directory to the header
      --alsologtostderr                  log to standard error as well as files
      --ca-cert-file string              File containing the x509 Certificate for HTTPS.
      --kube-api-burst int               Burst to use while talking with kubernetes apiserver (default 100)
      --kube-api-qps float32             QPS to use while talking with kubernetes apiserver (default 50)
      --kubeconfig string                Path to kubeconfig file with authorization and master location information.
      --log-backtrace-at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log-dir string                   If non-empty, write log files in this directory
      --log-file string                  If non-empty, use this log file
      --log-file-max-size uint           Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --log-flush-frequency duration     Maximum number of seconds between log flushes (default 5s)
      --logtostderr                      log to standard error instead of files (default true)
      --master string                    The address of the Kubernetes API server (overrides any value in kubeconfig)
      --port int                         the port used by admission-controller-server. (default 8443)
      --scheduler-name string            Volcano will handle pods whose .spec.SchedulerName is same as scheduler-name (default "volcano")
      --skip-headers                     If true, avoid header prefixes in the log messages
      --skip-log-headers                 If true, avoid headers when opening log files
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
[...]

PR #1691, opened 2021-08-23
fix: update volcano admission yaml

youhonglian:
remove "admission-conf" args

### DIFF
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -136,7 +136,6 @@ spec:
             - --tls-cert-file=/admission.local.config/certificates/tls.crt
             - --tls-private-key-file=/admission.local.config/certificates/tls.key
             - --ca-cert-file=/admission.local.config/certificates/ca.crt
-            - --admission-conf=/admission.local.config/configmap/volcano-admission.conf
             - --webhook-namespace=volcano-system
             - --webhook-service-name=volcano-admission-service
             - --logtostderr


installer/volcano-development.yaml
```diff
@@ -136,7 +136,6 @@ spec:
             - --tls-cert-file=/admission.local.config/certificates/tls.crt
             - --tls-private-key-file=/admission.local.config/certificates/tls.key
             - --ca-cert-file=/admission.local.config/certificates/ca.crt
-            - --admission-conf=/admission.local.config/configmap/volcano-admission.conf
```
shinytang6, line 139: This configuration is useful and should not be deleted :)

shinytang6 (changes requested):

installer/volcano-development.yaml
Thor-wl: Why delete admission-conf?

hwdef: image

I encountered this error.

shinytang6: same issue with https://github.com/volcano-sh/volcano/issues/1663, latest image should be updated.
/cc @Thor-wl

Thor-wl: > same issue with #1663, latest image should be updated.
> /cc @Thor-wl

Got it. It seems to be necessary to update recently.

youhonglian: > Why delete admission-conf? BTW, please add self signature when push the commit.

Thanks for reminding, I added my signature.
I also encountered this error. so i  updated YAML, is right? :)

Thor-wl: Well, thanks for the contribution. But the root reason is not here. I'll fix it.

youhonglian: thanks

Issue #1686, opened 2021-08-19
GPU sharing does not preemt or reclaim jobs from from lower weight queue


igormishsky:
**What happened**:
Have two servers with total of 65120 gpu-memory (using Volcano device plugin for Kubernetes).
Running 7 jobs with volcano.sh/gpu-memory: "7168" in the default queue (weight 1) and adding job in main queue (weight 7) with volcano.sh/gpu-memory: "7768". The job in the main queue is stuck in pending state

**What you expected to happen**
Adding job to queue to with higher weight did not reclaim or preempt any job with the lower weight queue.
I expect it to remove job from default queue in favor of the job in main queue. 

**How to reproduce it (as minimally and precisely as possible)**:
Install on kubernets with gpu https://github.com/volcano-sh/devices using the scheduler config provided.

Copy the template to template.yaml

```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: jobName
  namespace: defaullt
  annotations:
    volcano.sh/preemptable: "true"
    volcano.sh/revocable-zone: "*"
  labels:
    volcano.sh/preemptable: "true"
spec:
  minAvailable: 1
  schedulerName: volcano
  policies:
    - event: PodEvicted
      action: RestartJob
  plugins:
    ssh: []
    env: []
    svc: [ "--disable-network-policy=true" ]
  maxRetry: 2
  queue: default
  tasks:
    - replicas: 1
      name: "default-nginx"
      template:
        metadata:
          name: web
          annotations:
            volcano.sh/preemptable: "true"
            volcano.sh/revocable-zone: "*"
          labels:
            volcano.sh/preemptable: "true"
        spec:
          schedulerName: volcano
          containers:
            - image: busybox
              command:
                - sleep
                - "3300"
              name: sleep
              resources:
                limits:
                  volcano.sh/gpu-memory: "7168"
          restartPolicy: OnFailure

```

Run 

`for i in {1..10} ;do sed  "s/jobName/j$i/g" template.yaml | kubectl apply -f - ;done
`

Apply the the job in the main queue 
```
---
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: h1
  namespace: default
  annotations:
    volcano.sh/preemptable: "true"
    volcano.sh/revocable-zone: "*"
  labels:
    volcano.sh/preemptable: "true"
spec:
  minAvailable: 1
  schedulerName: volcano
  policies:
    - event: PodEvicted
[...]

Issue #1440, opened 2021-04-26
【summer2021】Support specify GPU number for pod resource request


Thor-wl:
Currently, Volcano only support specified GPU share memory. Specified GPU number is not supported. Your task is to support 'volcano.sh/gpu-number' just like nvidia gpu plugin to support specify GPU number for pod resource request. You can refer to https://github.com/volcano-sh/volcano/blob/master/docs/user-guide/how_to_use_gpu_sharing.md for more details.

Notes:
* These are competitions set for Summer2021. The Summer2021 official website: https://summer.iscas.ac.cn/
* Anyone who satisfies the requirement of the activity can submit application for the competition you are interested in. Please remind that the activity is only for college and graduate students. More details please refer to https://summer.iscas.ac.cn/help/student/
* The schedule of the activity refers to https://summer.iscas.ac.cn/help/timeline/

ChAnYaNG97:
Hi, I am interested in this issue. I have sent my resume and basic introduction to @Thor-wl @william-wang by e-mail.
Now I am working on the design and plan and will send in the near future.

Thor-wl:
> Hi, I am interested in this issue. I have sent my resume and basic introduction to @Thor-wl @william-wang by e-mail.
> Now I am working on the design and plan and will send in the near future.

I've received your resume. Thanks!

PR #1692, opened 2021-08-23
Add GPU Numbers Predicates

peiniliu:
Support specify GPU numbers for pod resource requests [issue#1440](https://github.com/volcano-sh/volcano/issues/1440)

Currently, Volcano only supports specified GPU share memory. Specified GPU number is not supported. This pr supports defining GPU numbers for pod resource requests. You can check the design doc https://github.com/peiniliu/volcano/blob/dev/docs/user-guide/how_to_use_gpu_number.md for more details.


jasonliu747:
Hi @peiniliu , thanks for your contribution. Please fix the DCO check.

william-wang:
@jasonliu747 Really hope this feature can be involved into v1.4. I have not got a gpu env. do you have gpu environment for verification?

jasonliu747:
@william-wang let me double check and get back to you!

william-wang:
@peiniliu Rebase and submit the pr again.

peiniliu:
@jasonliu747 @william-wang  
Hi, 
the pr has been updated! 
Best,
Peini

zamog:
This PR will fix #1686 ?

peiniliu:
this PR supports users to define GPU number predicates via container level resource requirements using the 'volcano.sh/gpu-numbers'. For provided issue, you may look at the queue and preempt or reclaim actions.

Thor-wl:
This feature is very useful and important! Let's take a review for it. @allReviewers

hwdef:
Can it support splitting a GPU. For example, use one-half of the GPU.

hwdef:
What if the gpu models are different?
If If there are three 2080 and three 3080 on one machine.The gpu number is set to 2. How to specify which gpu to use?

peiniliu:
> Can it support splitting a GPU. For example, use one-half of the GPU.

thanks for this comment! As far as I understand, splitting a GPU should use our gpu-sharing configuration. GPU number now only supports numbers in integer (even if we change it to support float number, I think the backend should always require integer+1 number of GPUs, for example, configure 1.2 will ask 2 GPUs).

peiniliu:
> What if the gpu models are different? If If there are three 2080 and three 3080 on one machine.The gpu number is set to 2. How to specify which gpu to use?

This is a nice question! Indeed now we can not distinguish different type of GPUs. GPUs will be allocated by their id in sequence. In this case, the first two GPUs will be selected.

I think now the node status and GPU device info do not support heterogeneous GPU devices. Volcano knows the numbers/memory of GPUs through the device plugin.

$ kubectl get node {node name} -oyaml
...
status:
  addresses:
  - address: 172.17.0.3
    type: InternalIP
  - address: volcano-control-plane
    type: Hostname
  allocatable:
    cpu: "4"
    ephemeral-storage: 123722704Ki
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    memory: 8174332Ki
    pods: "110"
    volcano.sh/gpu-memory: "89424"
    volcano.sh/gpu-number: "8"    # GPU resource
  capacity:
    cpu: "4"
    ephemeral-storage: 123722704Ki
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    memory: 8174332Ki
    pods: "110"
    volcano.sh/gpu-memory: "89424"
    volcano.sh/gpu-number: "8"   # GPU resource

peiniliu:
> /cc @peiniliu Peini, pls resolve the conflicts.

let's solve all the technical questions first :), I ll do this before merging!

hwdef:
> > Can it support splitting a GPU. For example, use one-half of the GPU.
> 
> thanks for this comment! As far as I understand, splitting a GPU should use our gpu-sharing configuration. GPU number now only supports numbers in integer (even if we change it to support float number, I think the backend should always require integer+1 number of GPUs, for example, configure 1.2 will ask 2 GPUs).

Do we have gpu-sharing related documents? Sorry, I don’t know much about volcano's gpu. I think the combination of these two functions into one function will be more user-friendly.

hwdef:
> > What if the gpu models are different? If If there are three 2080 and three 3080 on one machine.The gpu number is set to 2. How to specify which gpu to use?
> 
> This is a nice question! Indeed now we can not distinguish different type of GPUs. GPUs will be allocated by their id in sequence. In this case, the first two GPUs will be selected.
> 
> I think now the node status and GPU device info do not support heterogeneous GPU devices. Volcano knows the numbers/memory of GPUs through the device plugin.
> 
> ```shell
> $ kubectl get node {node name} -oyaml
> ...
> status:
>   addresses:
>   - address: 172.17.0.3
>     type: InternalIP
>   - address: volcano-control-plane
>     type: Hostname
>   allocatable:
>     cpu: "4"
>     ephemeral-storage: 123722704Ki
>     hugepages-1Gi: "0"
>     hugepages-2Mi: "0"
>     memory: 8174332Ki
>     pods: "110"
>     volcano.sh/gpu-memory: "89424"
>     volcano.sh/gpu-number: "8"    # GPU resource
>   capacity:
>     cpu: "4"
>     ephemeral-storage: 123722704Ki
>     hugepages-1Gi: "0"
>     hugepages-2Mi: "0"
>     memory: 8174332Ki
>     pods: "110"
>     volcano.sh/gpu-memory: "89424"
>     volcano.sh/gpu-number: "8"   # GPU resource
> ```

ok,I see. This is not a big problem.

peiniliu:
> > > Can it support splitting a GPU. For example, use one-half of the GPU.
> > 
> > 
> > thanks for this comment! As far as I understand, splitting a GPU should use our gpu-sharing configuration. GPU number now only supports numbers in integer (even if we change it to support float number, I think the backend should always require integer+1 number of GPUs, for example, configure 1.2 will ask 2 GPUs).
> 
> Do we have gpu-sharing related documents? Sorry, I don’t know much about volcano's gpu. I think the combination of these two functions into one function will be more user-friendly.

Hi,

yes, I think our first version to implement this was to combine them both. But I remember we asked in real cases that users usually use only one type of GPU configuration. @william-wang 
Also I attached here the gpu-sharing -> https://github.com/volcano-sh/volcano/blob/master/docs/user-guide/how_to_use_gpu_sharing.md

Best,
Peini

5had3z:
Before this intergrated, I just wanted to bring up the point of why not just properly intergrate with nvidia's gpu operator? I made a fork of volcano and made modifications to peiniliu's code so that I can just use nvidia.com/gpu requests and nodeSelector to select the type of GPU node I want to run on. I haven't had any issues so far with this.

k82cn:
Let's create a PR for design doc :)

shinytang6 (changes requested):
hi @peiniliu, I have a design question to ask:
If the user only turns on GPUNumberEnable, it seems that the functionally is similar to the nvidia device plugin. I understand the difference is:
* The real gpu allocation of nvidia device plugin is done at the device plugin level
* volcano does gpu id allocation at the scheduling level, and the volcano device plugin is only responsible for delivering the environment variables that scheduler patches?

Am l right? If so, is it possible to register volcano.sh/gpu-number as a resource in the volcano device plugin, just like nvidia device plugin, and do health management there. Then I understand that no special treatment is required in the scheduling layer, have you considered this way?

peiniliu:
Hi @shinytang6,

Thanks for your comments.

You are right, the real GPU allocation is done at the device plugin level(but in volcano case is volcano device plugin instead of the official Nvidia one)

See here the way the volcano device plugin allocates the GPU: [volcano` device plugin GPU allocation]((https://github.com/volcano-sh/devices/blob/b8725b77abef5255a9fc419085f8f194ca687a39/pkg/plugin/nvidia/server.go#L311)). Basically, the plugin is reading the GPU allocation results from what volcano scheduler patched to the pod annotation.
`id := GetGPUIDFromPodAnnotation(candidatePod)`

If to do in way as u said, we may need to change the plugin's allocation function to distinguish one card or multiple cards for gpu-share or gpu-number scenarios in volcano. 
I see the design doc [#1779](https://github.com/volcano-sh/volcano/pull/1779), they implemented this. If everyone vote for this change, I think I could change the implementation to follow this new device plugin. :)

But from my point of view, there are still some values to make the allocation in the scheduling layer. Given the scheduler has already known the GPU status, it is easy to decide if the allocation could work or not in advance.

Best,
Peini

shinytang6:
> Hi @shinytang6,
> 
> Thanks for your comments.
> 
> You are right, the real GPU allocation is done at the device plugin level(but in volcano case is volcano device plugin instead of the official Nvidia one)
> 
> See here the way the volcano device plugin allocates the GPU: [volcano device plugin GPU allocation](https://github.com/volcano-sh/devices/blob/b8725b77abef5255a9fc419085f8f194ca687a39/pkg/plugin/nvidia/server.go#L311). Basically, the plugin is reading the GPU allocation results from what volcano scheduler patched to the pod annotation. `id := GetGPUIDFromPodAnnotation(candidatePod)`
> 
> If to do in way as u said, we may need to change the plugin's allocation function to distinguish one card or multiple cards for gpu-share or gpu-number scenarios in volcano. I see the design doc [#1779](https://github.com/volcano-sh/volcano/pull/1779), they implemented this. If everyone vote for this change, I think I could change the implementation to follow this new device plugin. :)
> 
> But from my point of view, there are still some values to make the allocation in the scheduling layer. Given the scheduler has already known the GPU status, it is easy to decide if the allocation could work or not in advance.
> 
> Best, Peini

@peiniliu Thanks for the clarification, it makes sense to me. But my concern is that if `volcano.sh/gpu-number` is not registered as a resource, it is difficult to do health management, currently volcano device plugin just patches the number of volcano.sh/gpu-number when initialization which is not so robust.

peiniliu:
> > Hi @shinytang6,
> > Thanks for your comments.
> > You are right, the real GPU allocation is done at the device plugin level(but in volcano case is volcano device plugin instead of the official Nvidia one)
> > See here the way the volcano device plugin allocates the GPU: [volcano device plugin GPU allocation](https://github.com/volcano-sh/devices/blob/b8725b77abef5255a9fc419085f8f194ca687a39/pkg/plugin/nvidia/server.go#L311). Basically, the plugin is reading the GPU allocation results from what volcano scheduler patched to the pod annotation. `id := GetGPUIDFromPodAnnotation(candidatePod)`
> > If to do in way as u said, we may need to change the plugin's allocation function to distinguish one card or multiple cards for gpu-share or gpu-number scenarios in volcano. I see the design doc [#1779](https://github.com/volcano-sh/volcano/pull/1779), they implemented this. If everyone vote for this change, I think I could change the implementation to follow this new device plugin. :)
> > But from my point of view, there are still some values to make the allocation in the scheduling layer. Given the scheduler has already known the GPU status, it is easy to decide if the allocation could work or not in advance.
> > Best, Peini
> 
> @peiniliu Thanks for the clarification, it makes sense to me. But my concern is that if `volcano.sh/gpu-number` is not registered as a resource, it is difficult to do health management, currently volcano device plugin just patches the number of volcano.sh/gpu-number when initialization which is not so robust.

Hi, 
yes, currently 'volcano.sh/gpu-number' is registered as resources in the scheduler layer, but not in the device plugin. I agree to improve the device plugin and do health management there. By the way, I ve seen a lot of topics in the community to enhance the GPU, we may consider those designs together in order to avoid conflicts. I would like to join some talks and make contributions there.
Best,
Peini

shinytang6:
> > > Hi @shinytang6,
> > > Thanks for your comments.
> > > You are right, the real GPU allocation is done at the device plugin level(but in volcano case is volcano device plugin instead of the official Nvidia one)
> > > See here the way the volcano device plugin allocates the GPU: [volcano device plugin GPU allocation](https://github.com/volcano-sh/devices/blob/b8725b77abef5255a9fc419085f8f194ca687a39/pkg/plugin/nvidia/server.go#L311). Basically, the plugin is reading the GPU allocation results from what volcano scheduler patched to the pod annotation. `id := GetGPUIDFromPodAnnotation(candidatePod)`
> > > If to do in way as u said, we may need to change the plugin's allocation function to distinguish one card or multiple cards for gpu-share or gpu-number scenarios in volcano. I see the design doc [#1779](https://github.com/volcano-sh/volcano/pull/1779), they implemented this. If everyone vote for this change, I think I could change the implementation to follow this new device plugin. :)
> > > But from my point of view, there are still some values to make the allocation in the scheduling layer. Given the scheduler has already known the GPU status, it is easy to decide if the allocation could work or not in advance.
> > > Best, Peini
> > 
> > 
> > @peiniliu Thanks for the clarification, it makes sense to me. But my concern is that if `volcano.sh/gpu-number` is not registered as a resource, it is difficult to do health management, currently volcano device plugin just patches the number of volcano.sh/gpu-number when initialization which is not so robust.
> 
> Hi, yes, currently 'volcano.sh/gpu-number' is registered as resources in the scheduler layer, but not in the device plugin. I agree to improve the device plugin and do health management there. By the way, I ve seen a lot of topics in the community to enhance the GPU, we may consider those designs together in order to avoid conflicts. I would like to join some talks and make contributions there. Best, Peini

That will be great! 
/cc @Thor-wl  for review

docs/user-guide/how_to_use_gpu_number.md
```diff
@@ -0,0 +1,135 @@
...
+![gpu_number](../images/volcano-gpunumber.png)
+
+1. create a pod with `volcano.sh/gpu-number` resource request,
+
+2. volcano scheduler predicates and allocate gpu cards to the pod. Add the below annotation
+
+```yaml
+annotations:
+  volcano.sh/gpu-index: “0”
+  volcano.sh/predicate-time: “1593764466550835304”
```
Thor-wl, line 125: Is it necessary to provide the `predicate-time`? Besides, it seems to be a time stamp which is not so readable.

docs/user-guide/how_to_use_gpu_number.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+3. kubelet watches the pod bound to itself, and calls allocate API to set env before running the container.
+
+```yaml
+env:
+  NVIDIA_VISIBLE_DEVICES: “0” # GPU card index
+  VOLCANO_GPU_ALLOCATED: “1” # GPU number allocated
+  VOLCANO_GPU_TOTAL: “15109” # GPU memory of the card
```
Thor-wl, line 134: IC, so perhaps it is more readable to name the GPU memory of the card as `VOLCANO_GPU_MEMORY_TOTAL`.

pkg/scheduler/api/device_info.go
```diff
@@ -53,6 +53,14 @@ func (g *GPUDevice) getUsedGPUMemory() uint {
...
 }
 
+// isIdleGPU check if the device is idled.
+func (g *GPUDevice) isIdleGPU() bool {
+	if g.PodMap == nil || len(g.PodMap) == 0 {
+		return true
+	}
+	return false
+}
+
```
Thor-wl, line 63: Just a small advice, the following expression maybe more brief.
```go
// isIdleGPU check if the device is idled.
func (g *GPUDevice) isIdleGPU() bool {
    return g.PodMap == nil || len(g.PodMap) == 0
}
```
It is not something matters.

pkg/scheduler/api/device_info.go
```diff
@@ -70,3 +78,21 @@ func getGPUResourceOfContainer(container *v1.Container) uint {
 	}
 	return mem
 }
+
+// GetGPUNumberOfPod returns the number of GPUs required by the pod.
+func GetGPUNumberOfPod(pod *v1.Pod) int {
+	var gpus int
+	for _, container := range pod.Spec.Containers {
```
Thor-wl, line 85: Please also take initContainer into consideration.

pkg/scheduler/api/pod_info.go
```diff
@@ -160,20 +160,26 @@ func GetPodResourceWithoutInitContainers(pod *v1.Pod) *Resource {
...
+//return the gpu index list
+func GetGPUIndex(pod *v1.Pod) []int {
 	if len(pod.Annotations) > 0 {
 		value, found := pod.Annotations[GPUIndex]
 		if found {
-			id, err := strconv.Atoi(value)
-			if err != nil {
-				klog.Errorf("invalid %s=%s", GPUIndex, value)
-				return -1
+			ids := strings.Split(value, ",")
```
Thor-wl, line 168: Are there some validations about the format of this annotation?

peiniliu:
I changed docs because the device plugin now sets GPU share as default, as well as fixed some env names and the flow image,
I think now this is ready as well.

Thor-wl (commented):
Except for the consideration of initContianer, it is OK to me generally.

docs/user-guide/how_to_use_gpu_number.md
```diff
@@ -0,0 +1,135 @@
...
+  name: gpu-pod1
+spec:
+  containers:
+    - name: cuda-container
+      image: nvidia/cuda:9.0-devel
+      command: ["sleep"]
+      args: ["100000"]
+      resources:
+        limits:
+          volcano.sh/gpu-number: 1 # requesting 2 gpu cards
```
shinytang6, line 98: requesting 1 gpu cards

docs/user-guide/how_to_use_gpu_sharing.md
```diff
@@ -46,6 +46,8 @@ Same as above, after installed, update the scheduler configuration in `volcano-s
 
 Please refer to [volcano device plugin](https://github.com/volcano-sh/devices/blob/master/README.md#quick-start)
 
+* By default volcano device plugin supports shared GPUs, users do not need to config volcano device plugin. Default setting is the same as setting --gpu-strategy=number. For more information [volcano device plugin configuration](https://github.com/peiniliu/devices/blob/dev/doc/config.md)
```
shinytang6, line 49: please use the volcano official link here

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -111,12 +114,19 @@ func enablePredicate(args framework.Arguments) predicateEnable {
...
 		cacheEnable:        false,
 		proportionalEnable: false,
 	}
 
 	// Checks whether predicate.GPUSharingEnable is provided or not, if given, modifies the value in predicateEnable struct.
 	args.GetBool(&predicate.gpuSharingEnable, GPUSharingPredicate)
+	args.GetBool(&predicate.gpuNumberEnable, GPUNumberPredicate)
+
+	if predicate.gpuSharingEnable && predicate.gpuNumberEnable {
+		klog.Errorf("can not define true in both gpu sharing and gpu number")
```
shinytang6, line 127: l think it's better to use `klog.Fatal` here

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -175,31 +185,62 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				return
 			}
 
+			//predicate gpu sharing
 			if predicate.gpuSharingEnable && api.GetGPUResourceOfPod(pod) > 0 {
```
shinytang6, line 189: let's rename GetGPUResourceOfPod to GetGPUMemoryOfPod

shinytang6 (changes requested):

docs/user-guide/how_to_use_gpu_number.md
```diff
@@ -0,0 +1,135 @@
...
+NVIDIA_VISIBLE_DEVICES=0
+VOLCANO_GPU_MEMORY_TOTAL=15109
+VOLCANO_GPU_ALLOCATED=1
+...
+```
+### Understanding How Multiple GPU Cards Requirement Works 
+
+The main architecture is similar as the previous, but the gpu-index results of each pod will be a list of gpu cards index. 
+
+![gpu_number](../images/gpu-number.png)
```
shinytang6, line 116: Kebe-API-Server => Kube-API-Server

peiniliu:
I fixed those comments.

docs/user-guide/how_to_use_gpu_sharing.md
```diff
@@ -46,6 +46,8 @@ Same as above, after installed, update the scheduler configuration in `volcano-s
 
 Please refer to [volcano device plugin](https://github.com/volcano-sh/devices/blob/master/README.md#quick-start)
 
+* By default volcano device plugin supports shared GPUs, users do not need to config volcano device plugin. Default setting is the same as setting --gpu-strategy=number. For more information [volcano device plugin configuration](https://github.com/volcano-sh/devices/blob/dev/doc/config.md)
+
```
shinytang6, line 50: the link is 404

shinytang6 (changes requested):

shinytang6 (commented):
Generally LGTM. 
/cc @Thor-wl @william-wang Please take another look

pkg/scheduler/api/device_info.go
Thor-wl: BTW, is this commnet resolved? @peiniliu

Thor-wl (commented):
BTW, is the comment about `initContainer` resolved?

peiniliu:
> BTW, is the comment about `initContainer` resolved?

as explained, currently device plugin does not fit multiple containers. as well as the initContainer is designed for some start scripts which may not request GPUs. Well, I think the important part is the device plugin, once it supports, the GPU usage for initContainer can be added.

Thor-wl:
> > BTW, is the comment about `initContainer` resolved?
> 
> as explained, currently device plugin does not fit multiple containers. as well as the initContainer is designed for some start scripts which may not request GPUs. Well, I think the important part is the device plugin, once it supports, the GPU usage for initContainer can be added.

IC, that's OK for me.

peiniliu:
I removed some unrelated info in the doc after the presentation.

Thor-wl (approved):

shinytang6 (approved):
/lgtm

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -175,31 +185,62 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			if predicate.gpuSharingEnable && api.GetGPUMemoryOfPod(pod) > 0 {
 				nodeInfo, ok := ssn.Nodes[nodeName]
 				if !ok {
 					klog.Errorf("Failed to get node %s info from cache", nodeName)
 					return
 				}
-
-				id := predicateGPU(pod, nodeInfo)
-				if id < 0 {
+				ids := predicateGPUbyMemory(pod, nodeInfo)
```
WingkaiHo, line 195: share gpu pod only need to one gpu id, but predicateGPUbyMemory reture all gpu id which are suitable for the pod.

Issue #1710, opened 2021-09-01
can not schedule job 


qiankunli:
volcano version： latest  built by 8.31

volcano-scheduler log
```
I0901 02:56:07.473120       1 allocate.go:197] Try to allocate resource to 1 tasks of Job <xdl-system/mljob0901095107577>
I0901 02:56:07.473128       1 proportion.go:267] Queue <default>: deserved <cpu 66000.00, memory 239444426752.00, hugepages-2Mi 0.00, hugepages-1Gi 0.00, nvidia.com/gpu 12000.00, aliyun.com/gpu-count 0.00, aliyun.com/gpu-mem 0.00, tencent.com/vcuda-core 0.00, tencent.com/vcuda-memory 0.00>, allocated <cpu 0.00, memory 0.00>, share <0>, underUsedResName [cpu memory]
I0901 02:56:07.473153       1 allocate.go:209] Queue <default> is overused when considering task <mljob0901095107577-worker-0>, ignore it.
```
remove proportion plugin and restart volcano-scheduler
```
I0901 05:44:20.714920       1 allocate.go:163] Try to allocate resource to Jobs in Namespace <xdl-system> Queue <default>
I0901 05:44:20.714939       1 allocate.go:197] Try to allocate resource to 1 tasks of Job <xdl-system/mljob0901070435609>
I0901 05:44:20.714956       1 allocate.go:209] Queue <default> is overused when considering task <mljob0901070435609-worker-0>, ignore it.
```

tfjob yaml

```
Name:         mljob0901095107577
Namespace:    xdl-system
Labels:       <none>
Annotations:  <none>
API Version:  kubeflow.org/v1
Kind:         TFJob
Metadata:
  Creation Timestamp:  2021-09-01T01:52:19Z
  Generation:          1
    Manager:         tf-operator.v1
    Operation:       Update
    Time:            2021-09-01T01:52:44Z
  Resource Version:  410322266
  Self Link:         /apis/kubeflow.org/v1/namespaces/xdl-system/tfjobs/mljob0901095107577
  UID:               4f51d4f2-12a6-411f-8671-52ac735d39e8
Spec:
  Clean Pod Policy:  Running
  Tf Replica Specs:
    Worker:
      Replicas:        1
      Restart Policy:  Never
      Template:
        Spec:
          Affinity:
            Node Affinity:
              Required During Scheduling Ignored During Execution:
                Node Selector Terms:
                  Match Expressions:
                    Key:       queue
[...]

PR #1712, opened 2021-09-02
fix diff function

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/1710

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -443,23 +443,22 @@ func (r *Resource) Diff(rr *Resource, defaultValue DimensionDefaultValue) (*Reso
 	rightRes := rr.Clone()
 	increasedVal := EmptyResource()
 	decreasedVal := EmptyResource()
-
 	r.setDefaultValue(leftRes, rightRes, defaultValue)
 
 	if leftRes.MilliCPU > rightRes.MilliCPU {
-		increasedVal.MilliCPU += leftRes.MilliCPU - rightRes.MilliCPU
+		increasedVal.MilliCPU = leftRes.MilliCPU - rightRes.MilliCPU
 	} else {
-		decreasedVal.MilliCPU += rightRes.MilliCPU - leftRes.MilliCPU
+		decreasedVal.MilliCPU = rightRes.MilliCPU - leftRes.MilliCPU
 	}
 
 	if leftRes.Memory > rightRes.Memory {
-		increasedVal.Memory += leftRes.Memory - rightRes.Memory
+		increasedVal.Memory = leftRes.Memory - rightRes.Memory
 	} else {
-		decreasedVal.Memory += rightRes.Memory - leftRes.Memory
+		decreasedVal.Memory = rightRes.Memory - leftRes.Memory
 	}
 
-	increasedVal.ScalarResources = map[v1.ResourceName]float64{}
-	decreasedVal.ScalarResources = map[v1.ResourceName]float64{}
+	increasedVal.ScalarResources = make(map[v1.ResourceName]float64, 0)
+	decreasedVal.ScalarResources = make(map[v1.ResourceName]float64, 0)
 	for lName, lQuant := range leftRes.ScalarResources {
 		rQuant, _ := rightRes.ScalarResources[lName]
 		if lQuant == -1 {
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -437,38 +437,43 @@ func (r *Resource) Equal(rr *Resource, defaultValue DimensionDefaultValue) bool 
 }
 
 // Diff calculate the difference between two resource object
-func (r *Resource) Diff(rr *Resource) (*Resource, *Resource) {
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/resource_info.go
```diff
@@ -437,38 +437,43 @@ func (r *Resource) Equal(rr *Resource, defaultValue DimensionDefaultValue) bool
...
+			continue
 		}
-
-		if rQuant > rrQuant {
-			if increasedVal.ScalarResources == nil {
-				increasedVal.ScalarResources = map[v1.ResourceName]float64{}
-			}
-			increasedVal.ScalarResources[rName] += rQuant - rrQuant
+		if rQuant == -1 {
+			decreasedVal.ScalarResources[lName] = -1
```
huone1, line 470: I think it needn't to think about Infinity indiff function， just only set to zero ； if set to “-1” ， the logic using  this function also do the corresponding adjust

huone1:
What's the reason?

pkg/scheduler/api/resource_info.go
Thor-wl: As a tool function, it should be considered as general use for furture scenarios. For example, one of the comparsion object is queue.capability, dimensions not set should be considered as infinity.

Thor-wl:
> What's the reason?

Diff function is not robust and lack of consideration about unequal dimensions between two objects.

pkg/scheduler/api/resource_info.go
```diff
@@ -437,38 +437,43 @@ func (r *Resource) Equal(rr *Resource, defaultValue DimensionDefaultValue) bool
 }
 
 // Diff calculate the difference between two resource object
-func (r *Resource) Diff(rr *Resource) (*Resource, *Resource) {
+// Note: if `defaultValue` equals `Infinity`, the difference between two values will be `Infinity`, marked as -1
+func (r *Resource) Diff(rr *Resource, defaultValue DimensionDefaultValue) (*Resource, *Resource) {
+	leftRes := r.Clone()
```
william-wang, line 442: The Clone() function added recently causes 400ms+ latency in scheduling period while submit 1w pod.  We'd better think out a way without using Clone().

Thor-wl: > The Clone() function added recently causes 400ms+ latency in scheduling period while submit 1w pod. We'd better think out a way without using Clone().

OK. I think it's better to open another issue to optimize performance totally. As to this fix, the input parameter is a pointer and we should not make any modifications at the original object. So clone is needed.

pkg/scheduler/api/resource_info.go
```diff
@@ -437,38 +437,43 @@ func (r *Resource) Equal(rr *Resource, defaultValue DimensionDefaultValue) bool
...
-		rrQuant, ok := rr.ScalarResources[rName]
-
-		if !ok {
-			rrQuant = 0
+	increasedVal.ScalarResources = make(map[v1.ResourceName]float64, 0)
+	decreasedVal.ScalarResources = make(map[v1.ResourceName]float64, 0)
+	for lName, lQuant := range leftRes.ScalarResources {
+		rQuant, _ := rightRes.ScalarResources[lName]
+		if lQuant == -1 {
+			increasedVal.ScalarResources[lName] = -1
```
jinzhejz, line 466: do we need to enhance `Resource.Sub()` and `Resource.Add()` when set `Resource.ScalarResources`  to -1?

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -178,7 +178,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			klog.V(4).Infof("The attributes of queue <%s> in proportion: deserved <%v>, allocate <%v>, request <%v>, share <%0.2f>",
 				attr.name, attr.deserved, attr.allocated, attr.request, attr.share)
 
-			increased, decreased := attr.deserved.Diff(oldDeserved)
+			increased, decreased := attr.deserved.Diff(oldDeserved, api.Zero)
 			increasedDeserved.Add(increased)
 			decreasedDeserved.Add(decreased)
```
jinzhejz, line 183: do we need to compare `increasedDeserved` and `increased` here?

pkg/scheduler/api/resource_info.go
Thor-wl: > do we need to enhance `Resource.Sub()` and `Resource.Add()` when set `Resource.ScalarResources` to -1?

Actually yes, but that should be in PR focusing on the enhancement. That my be better.

pkg/scheduler/plugins/proportion/proportion.go
Thor-wl: > do we need to compare `increasedDeserved` and `increased` here?

I've not analysed the logic in proportion here. Just beacause of the input parameters changed, this call should be changed, too.

pkg/scheduler/api/resource_info.go
william-wang: That's ok to me to open another issue to improve the performance.

william-wang (approved):

PR #1715, opened 2021-09-03
use root context

lowang-bh:
use a root context go control all workflow


pkg/signals/signal.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
k82cn, line 2: Why Kuberentes Authors? If dependents on k8s, let's import it.

lowang-bh: fixed. import it from apiserver

lowang-bh:
> @lowang-bh Hi, any related issue?

No related issue, just a improvement.

PR #1726, opened 2021-09-06
fix(underused): Do not check overused when there is no UnderUsedResourceFn added

### LINKED ISSUE #1724: Queues are always OVERUSED when plugin proportion is disabled
**What happened**:
- Queues are **always overused** when there is no `UnderUsedResourceFn` is added (i.e. plugin `proportion` is disabled).
  - https://github.com/volcano-sh/volcano/commit/012c2a26068a144813866c96579bbe717a0ddf74#diff-3dcbd15d9e1bf118608b4eedcaadc5cc7bdb9f60b482ad0ca7238ada5ad9e32cR208
  ```go
  if !ssn.UnderusedResources(queue).Contains(taskRequest) {
      klog.V(3).Infof("Queue <%s> is overused when considering task <%s>, ignore it.", queue.Name, task.Name)
      continue
  }
  ```
- It was introduced in PR #1637 .

- Queue is always underused when there is no `UnderUsedResourceFn` is added, i.e.: `ssn.UnderusedResources(queue).Contains(taskRequest)` returns true.
- The following scheduling cycle can continue to run.

- Do not enable plugin `proportion` and do not use any feature related to that.
  - Only turn on plugin predicates & gang, and use the default queue.
- Gang scheduling any podgroup, it will always fail due to queue overused.

+ @Thor-wl PTAL.

eggiter:
- fix #1724 .
- Do not check overused when there is no `UnderUsedResourceFn` added.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -204,8 +204,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 			// Check whether the queue is overused on dimension that the task requested
 			taskRequest := task.Resreq.ResourceNames()
-
-			if !ssn.UnderusedResources(queue).Contains(taskRequest) {
+			if underusedResources := ssn.UnderusedResources(queue); underusedResources != nil && !underusedResources.Contains(taskRequest) {
 				klog.V(3).Infof("Queue <%s> is overused when considering task <%s>, ignore it.", queue.Name, task.Name)
 				continue
 			}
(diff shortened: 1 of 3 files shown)


Thor-wl:
Thanks for the fix. 
/approve

eggiter:
> Thanks for the fix.
> /approve
@Thor-wl 
Needs /lgtm : )

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -205,7 +205,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 			// Check whether the queue is overused on dimension that the task requested
 			taskRequest := task.Resreq.ResourceNames()
 
-			if !ssn.UnderusedResources(queue).Contains(taskRequest) {
+			if ssn.UnderusedResourceRegistered() && !ssn.UnderusedResources(queue).Contains(taskRequest) {
```
k82cn, line 208: return `nil` if no fn registered, return `[]` if no under-used resources.

eggiter: updated

shinytang6 (commented):
lgtm overall

PR #1729, opened 2021-09-07
fix(scheduler): improve job/task clone func

shinytang6:

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -190,6 +190,7 @@ func (ti *TaskInfo) Clone() *TaskInfo {
 		Name:           ti.Name,
 		Namespace:      ti.Namespace,
 		Priority:       ti.Priority,
+		PodVolumes:     ti.PodVolumes,
 		Pod:            ti.Pod,
 		Resreq:         ti.Resreq.Clone(),
 		InitResreq:     ti.InitResreq.Clone(),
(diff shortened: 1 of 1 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -190,6 +190,7 @@ func (ti *TaskInfo) Clone() *TaskInfo {
 		Name:           ti.Name,
 		Namespace:      ti.Namespace,
 		Priority:       ti.Priority,
+		PodVolumes:     ti.PodVolumes,
```
k82cn, line 193: should we use deep-copy?

shinytang6: We did not mutate PodVolumes directly, so l think it ok to not use deepcopy here

pkg/scheduler/api/job_info.go
```diff
@@ -485,18 +486,20 @@ func (ji *JobInfo) Clone() *JobInfo {
...
 
-		TaskStatusIndex:  map[TaskStatus]tasksMap{},
-		TaskMinAvailable: ji.TaskMinAvailable,
-		Tasks:            tasksMap{},
-		Preemptable:      ji.Preemptable,
-		RevocableZone:    ji.RevocableZone,
-		Budget:           ji.Budget.Clone(),
+		TaskStatusIndex:       map[TaskStatus]tasksMap{},
+		TaskMinAvailable:      ji.TaskMinAvailable,
+		TaskMinAvailableTotal: ji.TaskMinAvailableTotal,
```
william-wang, line 498: What does "TaskMinAvailableTotal" mean?

shinytang6: the sum of task minAvailable

PR #1733, opened 2021-09-08
upgrade k8s to v1.19.11 because of security notification

Thor-wl:
security


hwdef (commented):
/lgtm

Issue #1734, opened 2021-09-08
record label `volcano.sh/queue-name: default` on pod


qiankunli:
in grafana dashboard `Volcano Queue View` ，there is a Variables whose name is "queue"

![729C4749-757B-41C3-8FC5-B0B8DF749018](https://user-images.githubusercontent.com/8024725/132489290-e4892c98-5bb4-4c38-9347-aa2464cc004f.png)

for example, panel `Running job` use metric `kube_pod_volcano_container_status_running{job="kube-state-metrics",queue="$queue"}`, label=queue come from pod label `volcano.sh/queue-name`

when user create vcjob, it will create pod with label = `volcano.sh/queue-name` normally, but if user create tfjob/pytorchjob/mpijob, it will create pod without the label, leading to no data in grafana

so I suggest that volcano update pod with adding label `volcano.sh/queue-name` when bind pod with node

alcorf-mizar:
> /assign @alcorf-mizar Please take a look. THX.
Confused account. @alcorj-mizar Please take a look.

PR #1735, opened 2021-09-09
add queue label in pod created by tfjob

qiankunli:
fix issue #1734

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -19,6 +19,8 @@ package cache
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Thor-wl:
Please add `--signoff` flag when execute `git commit`

k82cn:
/hold

let's handle this in admission controller instead of schedueler :)

pkg/scheduler/cache/cache.go
```diff
@@ -655,7 +657,11 @@ func (sc *SchedulerCache) Bind(taskInfo *schedulingapi.TaskInfo, hostname string
 			}
 		}()
 	}
-
+	// add QueueNameKey label
+	labelPatch := fmt.Sprintf(`[{"op":"add","path":"/metadata/labels/%s","value":"%s" }]`, strings.Replace(batch.QueueNameKey,"/","~1",-1), string(job.Queue))
+	if _, err := sc.kubeClient.CoreV1().Pods(p.Namespace).Patch(context.TODO(), p.Name, types.JSONPatchType, []byte(labelPatch), metav1.PatchOptions{}); err != nil {
```
huone1, line 662: it is better to add  the data about it in admission/pods/mutate

Issue #1609, opened 2021-07-11
There may be something wrong with "pod affinity" and "pod with volume"


Thor-wl:
**What happened**:
There may be some bugs when dealing with "pod affinity" and "pod with volume" after a set of tests in CI.
The details as follows:
![use exisisting PVC in job](https://user-images.githubusercontent.com/61179420/125180926-578c8080-e232-11eb-91b8-a068c11c4a48.PNG)
![Test SSH Plugin with Pod Affinity](https://user-images.githubusercontent.com/61179420/125180929-5bb89e00-e232-11eb-86da-d3b4a7e4c4b7.PNG)
![Pod Affinity](https://user-images.githubusercontent.com/61179420/125180931-5ce9cb00-e232-11eb-9fd3-911a467898d1.PNG)

**What you expected to happen**:
All CI E2E cases passes nornally.
**How to reproduce it (as minimally and precisely as possible)**:
Retest code in https://github.com/volcano-sh/volcano/pull/1595

**Anything else we need to know?**:
In order to not block other PRs, I'll skip thest three case temporarily and continue analyses the root reasons. Then submit the bug fix.

**Environment**:
- Volcano Version: master branch
- Kubernetes version (use `kubectl version`): 
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #1742, opened 2021-09-13
fix vcjob not work when mount volume

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/1609


hwdef:
Why do we need to update pvc, please explain the reason.

Thor-wl:
> Why do we need to update pvc, please explain the reason.

PVC not existing will be created automatically by volcano such as the `pvSpec` specified in e2e "use exisisting PVC in job". That's the root reason why that case cannot pass ever.

william-wang (approved):
/lgtm

hwdef:
But when vcjob is bound to an existing pvc, there will be this problem.

PR #1649, opened 2021-07-29
feat(core): upgrade controller tools to v0.6.0

shinytang6:
resolves https://github.com/volcano-sh/volcano/issues/1567

PR #1743, opened 2021-09-13
e2e cases about pod affinity skip cancel

Thor-wl:
@shinytang6 has fixed it. PR: https://github.com/volcano-sh/volcano/pull/1649
Ref: https://github.com/volcano-sh/volcano/issues/1609


shinytang6 (approved):
/approve

Issue #1740, opened 2021-09-11
Optimize the throughput of scheduler


william-wang:
#### What would you like to be added:
Optimize the scheduler throughput for high performance workload.

#### Why is this needed:
As more and more high performance workload are migrating to kubernetes. Current throughtput of scheduler can not meet throughput requirement for large scale job submitting scenario.

justadogistaken:
k8s maintains the snapshot in the scheduler, and use generation and double linked list to avoid clone all information every scheduling period. Is that possible in volcano?

k82cn:
> k8s maintains the snapshot in the scheduler, and use generation and double linked list to avoid clone all information every scheduling period. Is that possible in volcano?

That's reasonable to me :)

PR #1745, opened 2021-09-13
refactor the Jobinfo functions to reduce redundant computing

william-wang:
#1740 

 ReadyTaskNum function costs more time in the scheduling period than we thought from the pprof result. 
1.  move the isEmpty() from ReadyTaskNum(scheduling period) to taskinfo creating period.
2.  reduce the loop operation in ReadyTaskNum and WaitingTaskNum.

Around 600ms is reduced in scheduling period after above changes is made while submiting 1w pod.

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -100,6 +100,7 @@ type TaskInfo struct {
 	Priority    int32
 	VolumeReady bool
 	Preemptable bool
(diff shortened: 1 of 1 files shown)


shinytang6 (commented):
l am ok with the changes, btw `CheckTaskMinAvailable` and `ValidTaskNum` can be modified together in this pr.

Issue #1736, opened 2021-09-09
If maxRetry is set to 1, the vcjob will not complete.


hwdef:
**What happened**:

I run a vcjob, set `maxRetry` to 1, when pod is complete, the vcjob will not complete, vcjob will always be in the completing state.

**What you expected to happen**:

The job is in a completed state

**How to reproduce it (as minimally and precisely as possible)**:

1. set `maxRetry` to 1.
2. run vcjob

**Anything else we need to know?**:

vcjob
```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: lm-mpi-job
spec:
  minAvailable: 3
  maxRetry: 1
  queue: default
  schedulerName: volcano
  plugins:
    ssh: []
    svc: []
  tasks:
    - replicas: 2
      name: mpiworker
      template:
        spec:
          containers:
            - command:
                - /bin/sh
                - -c
                - |
                  mkdir -p /var/run/sshd; /usr/sbin/sshd -D;
              image: volcanosh/example-mpi:0.0.1
              name: mpiworker
              ports:
                - containerPort: 22
                  name: mpijob-port
              workingDir: /home
          restartPolicy: OnFailure
    - replicas: 1
      name: mpimaster
      policies:
        - event: TaskCompleted
          action: CompleteJob
      template:
        spec:
          containers:
            - command:
                - /bin/sh
                - -c
                - |
                  MPI_HOST=`cat /etc/volcano/mpiworker.host | tr "\n" ","`;
                  mkdir -p /var/run/sshd; /usr/sbin/sshd;
                  mpiexec --allow-run-as-root --host ${MPI_HOST} -np 2 mpi_hello_world > /home/re;
              image: volcanosh/example-mpi:0.0.1
              name: mpimaster
              ports:
                - containerPort: 22
                  name: mpijob-port
              workingDir: /home
          restartPolicy: OnFailure
```

![图片](https://user-images.githubusercontent.com/13084946/132661730-a8bc8d3a-9101-47c5-bc44-ddeb582d1f89.png)

**Environment**:
- Volcano Version: 1.3.0
- Kubernetes version (use `kubectl version`): 1.21.0
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release): 
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Thor-wl:
[...]

PR #1746, opened 2021-09-13
fix bug that vcjob is not compeleted when maxRetry is 1

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/1736

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -69,7 +69,7 @@ func (cc *jobcontroller) killJob(jobInfo *apis.JobInfo, podRetainPhase state.Pha
 
 			maxRetry := job.Spec.MaxRetry
 			lastRetry := false
-			if job.Status.RetryCount >= maxRetry-1 {
+			if job.Status.RetryCount > maxRetry-1 {
 				lastRetry = true
 			}
 


pkg/controllers/job/job_controller_actions.go
```diff
@@ -69,7 +69,7 @@ func (cc *jobcontroller) killJob(jobInfo *apis.JobInfo, podRetainPhase state.Pha
 
 			maxRetry := job.Spec.MaxRetry
 			lastRetry := false
-			if job.Status.RetryCount >= maxRetry-1 {
+			if job.Status.RetryCount > maxRetry-1 {
```
d4l3k, line 72: on the last retry what's the retry count? my naive reading is that it would make this condition always be false. i.e. on the 2nd retry with maxRetry=3 you get `2 > 3-1` so lastRetry is never true. 

I don't quite understand how this breaks for `maxRetry == 1` but works for `maxRetry > 1`.

d4l3k (commented):
which part of this is the bit that actually completes the job here?

```
if !retain && !lastRetry {
  err := cc.deleteJobPod(job.Name, pod)
  if err == nil {
	  terminating++
	  continue
  }
  // record the err, and then collect the pod info like retained pod
  errs = append(errs, err)
  cc.resyncTask(pod)
}
```
Can we move that part outside of this logic so we only skip deleteJobPod? Alternatively maybe we should create a new pod for each retry and avoid deleting them entirely.

The other thing that would make this a lot more robust would be to add some unit tests for both the retry-fail before and the maxRetry==1 complete behavior. Is that feasible?


`

pkg/controllers/job/job_controller_actions.go
Thor-wl: As what I tested, the retry count is 0 when the job starts. So as to the example you list, on the 2nd retry with maxRetry=3, it is not the last retry but there is another chance to retry. That's the last retry.

Thor-wl:
> which part of this is the bit that actually completes the job here?
> 
> ```
> if !retain && !lastRetry {
>   err := cc.deleteJobPod(job.Name, pod)
>   if err == nil {
> 	  terminating++
> 	  continue
>   }
>   // record the err, and then collect the pod info like retained pod
>   errs = append(errs, err)
>   cc.resyncTask(pod)
> }
> ```
> 
> Can we move that part outside of this logic so we only skip deleteJobPod? Alternatively maybe we should create a new pod for each retry and avoid deleting them entirely.
> 
> The other thing that would make this a lot more robust would be to add some unit tests for both the retry-fail before and the maxRetry==1 complete behavior. Is that feasible?
> 
> `

/cc @wpeng102

wpeng102 (approved):
/lgtm

aivanou:
Hello. It seems that this bug still persists in 1.4.0. Repro:

      kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/v1.4.0/installer/volcano-development.yaml

Run job with more than a single retry.

    vcctl job list 

    Name                          Creation       Phase       JobType     Replicas    Min   Pending   Running   Succeeded   Failed    Unknown     RetryCount
cifar-trainer-tz45941r7u1b    2021-10-11     Failed      Batch       2           2     0         0         0           0         0           3

But `kubectl get pods` does not contain this pod. 


Let me know if you want a detailed information for the use case

PR #1747, opened 2021-09-13
add legency reason metric

qiankunli:
fix issue #1739


Thor-wl:
Please add `--signoff` to mark the signature when execute 'git commit', which is used to fix the DCO

Thor-wl:
Any update for this PR? @qiankunli

Issue #1720, opened 2021-09-04
Premption not happening between high priority and low priority jobs


Robert-Christensen-visa:
**What happened**:

When submitting several jobs, some with high priority and some with low priority, the high priority jobs do not preempt the lower priority jobs.  Both jobs are running in the same queue.

**What you expected to happen**:

If a low priority job is running and a pending high priority job is queued.  There there is not sufficient capacity for both to run, I would expect the lower priority job to stop running so the higher priority job can run.

**How to reproduce it (as minimally and precisely as possible)**:

This is my `volcano-scheduler.conf`

``` yaml
actions: "enqueue, allocate, preempt"
tiers:
- plugins:
  - name: priority
- plugins:
  - name: proportion
```

I run the following script to setup the environment and run the jobs.  I am running on my local machine, which has a capacity of 12CPUs and 8GB memory.

``` bash
cat <<EOF | kubectl apply -f -
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: basic-queue
  annotations:
    "volcano.sh/hierarchy": "root/basic"
    "volcano.sh/hierarchy-weights": "1/1"
spec:
  weight: 1
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 1000000
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: low-priority
value: 1
EOF

job_template_file=$(mktemp)
cat <<EOF > ${job_template_file}
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: vcjob-job-<id>
spec:
  minAvailable: 1
  schedulerName: volcano
  priorityClassName: <priority>
  policies:
    - event: PodEvicted
      action: RestartTask
  maxRetry: 100
  queue: <queue>
  tasks:
    - replicas: 2
      name: "x"
      template:
        metadata:
          name: core
        spec:
          priorityClassName: <priority>
          terminationGracePeriodSeconds: 10
          containers:
            - image: ubuntu
              imagePullPolicy: IfNotPresent
              name: ubuntu
              command: ['sh', '-c', "sleep 600000"]
              resources:
                requests:
                  cpu: "1000m"
                  memory: "256Mi"
          restartPolicy: OnFailure
EOF

for id in $(seq 6)
do
    sed 's/<queue>/basic/g' < ${job_template_file} | \
[...]

PR #1757, opened 2021-09-22
support preempt with priority plugin alone

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/1720 

### DIFF
--- a/pkg/scheduler/plugins/priority/priority.go
+++ b/pkg/scheduler/plugins/priority/priority.go
@@ -18,7 +18,6 @@ package priority
 
 import (
 	"k8s.io/klog"
-
 	"volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/plugins/util"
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/priority/priority.go
```diff
@@ -113,6 +113,26 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			}
+		}
+
+		return nil
+	}
+	ssn.AddJobValidFn(pp.Name(), validJobFn)
+
+	jobStarvingFn := func(obj interface{}) bool {
+		ji := obj.(*api.JobInfo)
+		return ji.ReadyTaskNum() < int32(len(ji.Tasks))
```
william-wang, line 132: Please take a look whether we need to consider the WaitingTaskNum here as well?

pkg/scheduler/plugins/priority/priority.go
```diff
@@ -113,6 +113,26 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 	}
 
 	ssn.AddPreemptableFn(pp.Name(), preemptableFn)
+
+	validJobFn := func(obj interface{}) *api.ValidateResult {
```
william-wang, line 117: If there is not special things to do in validJobFn. It's better to remove this function.

pkg/scheduler/plugins/priority/priority.go
huone1: it is the same as the code in gang plugin。 it is repeat。 I think it is better to modify the default value of the function JobStarving and change from false to true

pkg/scheduler/plugins/priority/priority.go
huone1: > If there is not special things to do in validJobFn. It's better to remove this function.

I also agree to delete it.

pkg/scheduler/plugins/priority/priority.go
Thor-wl: > Please take a look whether we need to consider the WaitingTaskNum here as well?

Yes, It is needed. Let me add it.

Thor-wl: > it is the same as the code in gang plugin。 it is repeat。 I think it is better to modify the default value of the function JobStarving and change from false to true

No, they are not same. `minAvailable` is not considered here.

william-wang (approved):
/lgtm
/approve

Issue #1750, opened 2021-09-15
Request to add run begin/end time for vcjob


HecarimV:
#### What would you like to be added:
I want to get the run beginTime and endTime of the vcjob.
Maybe we can add these info into job annotations like this:
```go
Annotations:  volcano.sh/job-begintime: 2021-09-15 13:24:10
              volcano.sh/job-endtime: 2021-09-15 13:45:55
```

#### Why is this needed:
Sometimes we need to check the actual running time, but not just simply creationTimestamp or deletionTimestamp.

shinytang6:
+1, l think these fields should be added to crd definitions of `Job.Status`

HecarimV:
yeah!  It's more appropriate to put it in the status field. 
Both approaches are acceptable

Thor-wl:
I agree with that, too. It's also useful for monitoring.

k82cn:
Can we add this to job status?

HecarimV:
> Can we add this to job status instead of annotations?

I think it's ok to add this to job status.

HecarimV:
> Can we add this to job status instead of annotations?

I think it's ok to add this to job status.

shinytang6:
@HecarimV Do you have time or are you interested in helping with this ?

HecarimV:
> @HecarimV Do you have time or are you interested in helping with this ?

I'm ok to work on it :)

PR #1764, opened 2021-09-28
Add job conditions (status&lastTransitionTime)

HecarimV:
Signed-off-by: liuhaojie <1187851227@qq.com>

Ref:#1750


shinytang6 (changes requested):
Maybe we also need to add the first condition entity when job init in `initJobStatus` ?

HecarimV:
> Maybe we also need to add the first condition entity when job init in `initJobStatus` ?

You mean the conditions starts with pending status

shinytang6:
> > Maybe we also need to add the first condition entity when job init in `initJobStatus` ?
> 
> You mean the conditions starts with pending status

yes

HecarimV:
> > Maybe we also need to add the first condition entity when job init in `initJobStatus` ?
> 
> You mean the conditions starts with pending status



> > > Maybe we also need to add the first condition entity when job init in `initJobStatus` ?
> > 
> > 
> > You mean the conditions starts with pending status
> 
> yes

Ok, I'll record the pending status when jobcontroller initiate Job

Issue #1782, opened 2021-10-14
gpu error occur, cause scheduling to the wrong node


zhiyuone:
**What happened**:
After pod scheduled, it create fail.
```
Status:         Failed
Reason:         UnexpectedAdmissionError
Message:        Pod Allocate failed due to requested number of devices unavailable for nvidia.com/gpu. Requested: 1, Available: 0, which is unexpected
```
**What you expected to happen**:
Pod Shouldn't schedule to this Node.

**How to reproduce it (as minimally and precisely as possible)**:
1. Node with 4 GPU
```
Capacity:
  nvidia.com/gpu:     4
Allocatable:
  nvidia.com/gpu:     4
```
2. Start Pod With 4 GPU in this Node.
3. One GPU error detect by device-plugin, then.
```
Capacity:
  nvidia.com/gpu:     4
Allocatable:
  nvidia.com/gpu:     3
```
4. Start Volcano Scheduler now.
5. Create Pod With 1 GPU.
6. Pod scheduled to this node, but node node idle gpu resource.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: e83119
- Kubernetes version (use `kubectl version`): 
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #1783, opened 2021-10-14
Catch add pod out of sync error

zhiyu0729:
If add pod to node error, this Node should be mark not ready .
Issue: #1782 
Signed-off-by: Zhiyu Wang <cloudsky.newbis@gmail.com>


zhiyuone:
> When will node be changed to ready?

When update Node, a new Node will replace. And recalculate whether it overflow.

PR #1790, opened 2021-10-15
Fix preemption behavior for multiple plugin tiers

### LINKED ISSUE #1791: Higher priority jobs are preempted by lower priority jobs in certain plugin configurations
**What happened**:
After upgrading from volcano v1.2 to v1.4, we found that some of our higher priority jobs were getting preempted by jobs of the same priority, or even those of lower priority. 

Jobs of the same priority or lower priority should not get preempted.

Added test in [pull request](https://github.com/volcano-sh/volcano/pull/1790)

**Environment**:
- Volcano Version: v1.4
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

kenoung:
This commit fixes a bug in the evaluation of whether tasks are preemptable.

Before this fix, if the first plugin in the first tier returns no candidates,
we will skip to the next tier and use plugins in that tier to evaluate whether
tasks are preemptible.  This commit modifies the behavior such that we return a
victims list for a given tier as long as one of the plugins in the tier has run
its PreemptableFn and initialized the victims list. We only go to the next tier
if there was no valid preemptable function in the previous tier.

Fixes #1791

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt_test.go
+++ b/pkg/scheduler/actions/preempt/preempt_test.go
@@ -32,13 +32,69 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/plugins/conformance"
+	"volcano.sh/volcano/pkg/scheduler/plugins/drf"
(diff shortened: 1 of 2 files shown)


lowang-bh:
The origin means to check another tier when former one has no victims; You can adjust the plugins to one tire is need;

detail comment on https://github.com/volcano-sh/volcano/pull/1563

Issue #1784, opened 2021-10-14
vcctl should work with oidc clusters


igormishsky:
#### What would you like to be added:

I will like to add support to vcctl such that also clusters that use oidc authentication will work.

#### Why is this needed:
Currently vcctl can only use .kube/config configuration cannot be used on oidc defined cluster.
We use multiply cluster all with oidc authentication and the cli cannot not be used locally from our terminal.

PR #1793, opened 2021-10-17
Adding oidc import to enable vcctl work with oidc cluster

igormishsky:
Following open issue #1784


igormishsky:
@shinytang6 this is the issue https://github.com/volcano-sh/volcano/issues/1784

PR #1817, opened 2021-11-02
Remove default quota info in `NewNamespaceCollection`

zen-xu:
Since `NewNamespaceCollection` only used in `updateResourceQuota`, add default quota info is unnecessary.

https://github.com/volcano-sh/volcano/blob/5ebe2d94de86c19d1c3368d7f6a2238519c36665/pkg/scheduler/cache/event_handlers.go#L692-L701


jasonliu747 (commented):
/lgtm

PR #1819, opened 2021-11-02
feat(outOfSync): tracks and re-synchronizes node when it becomes 'OutOfSync'

### LINKED ISSUE #1818: The `Lost of GPUs` problem has not been resolved completely
**What happened**:
1. Allocatable GPU of `node1` is now 6;
2. Create pod `pod1` requests 6 GPUs;
3. Volcano successfully schedules pod `pod1` to node `node1`;
4. However, Pod turned to `Failed` because of `UnexpectedAdmissionError`;

+ `node1` should not be scheduled any other pods;

1. `node1` has 8 allocatable GPUs;
2. Create pod `pod0` requests 8 GPUs;
3. GPU error detected and allocatable GPU of `node1` becomes 6;
4. Label or annotate node so that resource version of node updated;
5. Create pod `pod1` requests 6 GPUs;
6. The bug occurred, volcano successfully schedules pod `pod1` to node `node1`, but `pod1` turned to `Failed` because of `UnexpectedAdmissionError`;

+ It seems that the lost of GPU has not been throughly handled by now, related issues: 
    - #294
    - #1782
+ All the PRs have failed to cover the above bug: 
    - #306
    - #1685 
    - #1783

**Main reason to cause**:
1. As the comment https://github.com/volcano-sh/volcano/pull/1783#issuecomment-945314727 said, node will become `Ready` when its resource is updated. But there is a problem, when `SetNode` is called and node is updated, there might be tasks which were not added back to this node due to previous `AllocateFailError`.
2. As a supplement to the above question(`When will node be changed to ready?`), there are two possible cases that the node will become `Ready`:
    - The resources of node is updated. 
        - But in this case the node needs further re-synchronization to take back missing tasks(due to previous `AllocateFailError`);
    - The removal of tasks which were not added to this node due to previous `AllocateFailError`. 
        - In this case, maybe the lost CPUs didn't come back to the node, but the state of this node is consistent with its actual counterpart, and it's ready to accept Pods.

**Environment**:
- Volcano Version:
  + latest version.

eggiter:
+ fix #1818.
+ A possible fix to the above issue:
    -  Add a `RateLimitingQueue` to holds the not ready nodes and start a asynchronous thread to synchronize  them.
    -  The events of `updateNode` and `removeTask` will trigger the re-synchronization of the related nodes.
    - The synchronization of node will use the latest `v1.Node` and `v1.Pod` list( a list of pods located on the former node) from the informer cache to:
        + updates node;
        + removes unwanted tasks(unlikely) and adds missing tasks;

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -96,7 +96,7 @@ func (ni *NodeInfo) GetNodeAllocatable() *Resource {
 // NodeState defines the current state of node.
 type NodeState struct {
 	Phase  NodePhase
-	Reason string
+	Reason NodePhaseReason
 }
 
 // NodeUsage defines the real load usage of node
@@ -268,22 +268,33 @@ func (ni *NodeInfo) setOversubscription(node *v1.Node) {
 	}
 }
 
+// SetOutOfSync sets state of nodeInfo to `OutOfSync`
+func (ni *NodeInfo) SetOutOfSync() {
+	state := NodeState{
+		Phase:  NotReady,
+		Reason: ReasonOutOfSync,
+	}
+	ni.State = state
+}
+
+// IsOutOfSync determines if state of nodeInfo is `OutOfSync`
+func (ni *NodeInfo) IsOutOfSync() bool {
+	return !ni.Ready() && ni.State.Reason == ReasonOutOfSync
+}
+
 func (ni *NodeInfo) setNodeState(node *v1.Node) {
 	// If node is nil, the node is un-initialized in cache
 	if node == nil {
 		ni.State = NodeState{
 			Phase:  NotReady,
-			Reason: "UnInitialized",
+			Reason: ReasonUnInitialized,
 		}
 		return
 	}
 
 	// set NodeState according to resources
 	if !ni.Used.LessEqual(ni.Allocatable, Zero) {
-		ni.State = NodeState{
-			Phase:  NotReady,
-			Reason: "OutOfSync",
-		}
+		ni.SetOutOfSync()
 		return
 	}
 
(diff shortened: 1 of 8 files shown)


Thor-wl:
@eggiter Please help resolve the conflicts. Thanks!

eggiter:
> @eggiter Please help resolve the conflicts. Thanks!

Resolved. This is an important fix as discussed in https://github.com/volcano-sh/volcano/issues/1818#issuecomment-1102257005. PTAL!
@Thor-wl @k82cn

Thor-wl:
OK. Let me take a review recently. Thanks!

pkg/scheduler/api/node_info.go
```diff
@@ -263,22 +263,33 @@ func (ni *NodeInfo) setOversubscription(node *v1.Node) {
...
+func (ni *NodeInfo) SetOutOfSync() {
+	state := NodeState{
+		Phase:  NotReady,
+		Reason: ReasonOutOfSync,
+	}
+	ni.State = state
+}
+
+// OutOfSync determines if state of nodeInfo is `OutOfSync`
+func (ni *NodeInfo) OutOfSync() bool {
```
Thor-wl, line 276: Suggest to follow the common code of conduct. For this function is to judge the node status, perhaps name it as `IsOutOfSync` is better.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -321,10 +322,82 @@ func (sc *SchedulerCache) deleteNode(node *v1.Node) error {
...
+		return err
+	}
+	nodeInfo.State = schedulingapi.NodeState{Phase: schedulingapi.Ready}
+	nodeInfo.SetNode(newNode)
+	// remove unwanted tasks
+	for taskID, task := range nodeInfo.Tasks {
+		if _, exists := tasks[taskID]; exists {
+			continue
+		}
+		if err := sc.deleteTask(task); err != nil {
```
Thor-wl, line 350: IMO, maybe we can save the errors and continue to delete the following unwanted tasks when fails to delete some task. At the end of loop, throw all the errors to the caller. That may be more reasonable. In this way, we can ensure the most of the synchronization.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -321,10 +322,82 @@ func (sc *SchedulerCache) deleteNode(node *v1.Node) error {
...
+			klog.Errorf("failed to remove task <%s/%s> from node <%s> while synchronizing: %v", task.Namespace, task.Name, nodeInfo.Name, err)
+			return err
+		}
+	}
+	// add missing tasks
+	for taskID, task := range tasks {
+		if _, exists := nodeInfo.Tasks[taskID]; exists {
+			continue
+		}
+		if err := sc.addTask(task); err != nil {
```
Thor-wl, line 360: Same here.

Thor-wl: This enhancement is generally OK to me now. BTW, pls resolve the conflicts.

pkg/scheduler/cache/event_handlers.go
eggiter: "At the end of loop, throw all the errors to the caller" may result in big change of current schedule cycle, so i feel that the current fix may  have less impact on the current logic

eggiter:
@Thor-wl @k82cn

PR #1823, opened 2021-11-09
clean up readAdmissionConf

xiaoanyunfei:
`ioutil.ReadFile` already return `[]byte`, there is no need to convert `dat` to `string` type

```
func readAdmissionConf(confPath string) ([]byte, error) {
	return  ioutil.ReadFile(confPath)
}
```
`readAdmissionConf` has no need to exist if it only has one line `ioutil.ReadFile` function


shinytang6 (approved):
/lgtm
/approve

PR #1824, opened 2021-11-09
all controller should share same informerFactory

xiaoanyunfei:
If all  controllers created  their own `informerFactory`, there were many copies cache for one k8s object (such as volcano job or podgroup).  This would waste resource and memory 

all controller should share  one `informerFactory`

### DIFF
--- a/cmd/controller-manager/app/server.go
+++ b/cmd/controller-manager/app/server.go
@@ -36,6 +36,7 @@ import (
 
 	"volcano.sh/apis/pkg/apis/helpers"
 	vcclientset "volcano.sh/apis/pkg/client/clientset/versioned"
+	vcinformers "volcano.sh/apis/pkg/client/informers/externalversions"
 	"volcano.sh/volcano/cmd/controller-manager/app/options"
 	"volcano.sh/volcano/pkg/controllers/framework"
 	"volcano.sh/volcano/pkg/kube"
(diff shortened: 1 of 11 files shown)


hzxuzhonghu (commented):
/ok-to-test

lgtm

jasonliu747 (approved):
/lgtm

jasonliu747:
> /hold

@william-wang ptal

cmd/controller-manager/app/server.go
```diff
@@ -110,28 +119,31 @@ func Run(opt *options.ServerOption) error {
 	return fmt.Errorf("lost lease")
 }
 
-func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx context.Context) {
-	controllerOpt := &framework.ControllerOption{}
+func startControllers(ctx context.Context, controllerOpt *framework.ControllerOption) {
+	framework.ForeachController(func(c framework.Controller) {
+		if err := c.Initialize(controllerOpt); err != nil {
```
justadogistaken, line 124: I'm curious about this error handling. Is it acceptable for failling to initiate some of controllers (I know it won't go into errors regularly and the logic here is same as it used to be)?

william-wang:
> /cc @william-wang Any suggestion?

With the new pr, i hit some random issues. Let's me have further investigation.

PR #1829, opened 2021-11-12
clean addNumaInfo

xiaoanyunfei:
If `sc.Nodes[info.Name].NumaInfo` has already been nil, there is no need to compare with `newLocalInfo`


shinytang6 (approved):
/lgtm
/approve

jasonliu747 (commented):
/lgtm

Issue #1627, opened 2021-07-19
 Task launch order support


hwdef:
#### What would you like to be added:

Task-level DAG scheduling policy

#### Why is this needed:

This feature provides the ability to customize the order in which tasks are launched

The following scenarios come to mind so far:

* mpi job. the master needs to wait for the worker to start before starting, If the master is already started, but the worker is not yet started, the master will restart, which will add unnecessary waste of resources ,in this case, mpiworker needs to be in the running state, and then create the pod for mpimaster..
* Suppose there are two tasks and task 2 needs to use the calculation results of task 1,in this case, task1 needs to be in the complete state, and then create the pod for task2.

* some ETL  applications

#### design

* Add a field that is an array named `startPolicy` in task spec
* It contains two fields, `dependOn` and `trigger`
* `dependOn` is a string indicating the name of the task that the current task depends on
* `trigger`is used to detect the status of the task in dependOn. This field can be used with the `probe` struct in kubernetes

eg:

```yaml
startPolicy:
  - dependOn: taskA, taskB
    trigger: {probe}
  - dependOn: taskC
    trigger: {probe}
     
```

Thor-wl:
That's interesting. Are you interested in working for that or let's work together?

shinytang6:
maybe similar to https://github.com/volcano-sh/volcano/issues/1117
l am also interested in that feature, let's work together )

hwdef:
@Thor-wl I think both options are ok :)

Thor-wl:
Let's give a design doc first about introduction about the scenarios/apis/algorithms/workflows.etc together. Do you think so? @hwdef @shinytang6

PR #1833, opened 2021-11-16
task level advanced scheduling policy

hwdef:
ref: #1627

There is generally no problem, but there is currently a bug that has not been resolved.
When the vcjob is deleted, a new vcjob with the same name is created without waiting for all pods to be deleted. The pod in the vcjob will be lost and only a part of it will be created.


hwdef:
/hold

Cancel the hold until the bug is resolved

Thor-wl:
Suggest to add UT and e2e for this feature first.

pkg/controllers/job/helpers/helpers.go
```diff
@@ -35,7 +36,7 @@ const (
 )
 
 // GetTaskIndex returns task Index.
-func GetTaskIndex(pod *v1.Pod) string {
+func GetPodIndex(pod *v1.Pod) string {
```
Thor-wl, line 39: Suggest to reserve the original name for the uniformity of the name

pkg/controllers/job/helpers/helpers.go
```diff
@@ -70,3 +71,24 @@ func GenPVCName(jobName string) string {
 func GetJobKeyByReq(req *apis.Request) string {
 	return fmt.Sprintf("%s/%s", req.Namespace, req.JobName)
 }
+
+func GetTaskIndex(taskName string, job *batch.Job) int {
```
Thor-wl, line 75: Pls add comments for functions as format. Same to the whole PR.

shinytang6 (changes requested):
Will this pr contains changes about scheduler ?

pkg/controllers/job/helpers/helpers.go
```diff
@@ -70,3 +71,24 @@ func GenPVCName(jobName string) string {
...
+}
+
+func GetPodsInTask(taskName string, job *batch.Job) []string {
+	var res []string
+	for _, task := range job.Spec.Tasks {
+		if task.Name == taskName {
+			for index := 0; index < int(task.Replicas); index++ {
+				res = append(res, MakePodName(job.Name, taskName, index))
+			}
+		}
```
Thor-wl, line 91: Sugeest to add `break`  to optimize the performance when `task` is found

pkg/controllers/job/job_controller_actions.go
```diff
@@ -307,7 +307,7 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 	var running, pending, terminating, succeeded, failed, unknown int32
 	taskStatusCount := make(map[string]batch.TaskState)
 
-	var podToCreate []*v1.Pod
+	podToCreate := make(map[string][]*v1.Pod) // taskname -> []pod
```
Thor-wl, line 310: Remove nonstandard comments

pkg/controllers/job/job_controller_actions.go
```diff
@@ -349,33 +353,67 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
-					pod.Name, job.Name, err)
-				appendError(&creationErrs, fmt.Errorf("failed to create pod %s, err: %#v", pod.Name, err))
-			} else {
-				classifyAndAddUpPodBaseOnPhase(newPod, &pending, &running, &succeeded, &failed, &unknown)
-				calcPodStatus(pod, taskStatusCount)
-				klog.V(3).Infof("Created Task <%s> of Job <%s/%s>",
-					pod.Name, job.Namespace, job.Name)
+	var dependsOn batch.DependsOn
+
+	for taskName, podToCreateEachTask := range podToCreate {
```
Thor-wl, line 364: Suggest to make some optimization about this piece of code in performance.

pkg/controllers/job/job_controller_actions.go
```diff
@@ -349,33 +353,67 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+		go func(taskName string, podToCreateEachTask []*v1.Pod) {
+			taskIndex := jobhelpers.GetTaskIndex(taskName, job)
+			if job.Spec.Tasks[taskIndex].DependsOn != nil {
+				dependsOn = *job.Spec.Tasks[taskIndex].DependsOn
+				if len(job.Spec.Tasks[taskIndex].DependsOn.Name) > 1 && job.Spec.Tasks[taskIndex].DependsOn.Iteration == batch.IterationAny {
+					breakFlag := false
+					for {
+						time.Sleep(time.Second)
```
Thor-wl, line 372: Why sleep 1 second?

pkg/controllers/job/job_controller_actions.go
```diff
@@ -349,33 +353,67 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+							}
+						}
+						if breakFlag {
+							break
+						}
+					}
+				} else {
+					for _, DependsOnTask := range dependsOn.Name {
+						for {
+							time.Sleep(time.Second)
```
Thor-wl, line 386: Same as above

pkg/controllers/job/helpers/helpers.go
k82cn, line 39: Let's use `Task` and `TaskInstance` :)

hwdef:
> Will this pr contains changes about scheduler ?

No, it will not change the scheduler.

pkg/controllers/job/job_controller_actions.go
hwdef: I think there is no need to judge too fast here.

pkg/controllers/job/helpers/helpers.go
hwdef: @Thor-wl 

I think the function of this function is to find the index of the pod in the task, so maybe the name `GetPodIndex` is more appropriate?

Thor-wl: Indeed, it's an awkward point that `Task` in the code represents `Pod` most through the project. It is easily confused for new comers to make difference.
Do you have some good ideas?
@william-wang @shinytang6 @jasonliu747

pkg/controllers/job/job_controller_actions.go
```diff
@@ -349,33 +353,67 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
-			} else {
-				classifyAndAddUpPodBaseOnPhase(newPod, &pending, &running, &succeeded, &failed, &unknown)
-				calcPodStatus(pod, taskStatusCount)
-				klog.V(3).Infof("Created Task <%s> of Job <%s/%s>",
-					pod.Name, job.Namespace, job.Name)
+	var dependsOn batch.DependsOn
+
+	for taskName, podToCreateEachTask := range podToCreate {
+		go func(taskName string, podToCreateEachTask []*v1.Pod) {
+			taskIndex := jobhelpers.GetTaskIndex(taskName, job)
```
shinytang6, line 366: suggest to make this block into separate function

shinytang6 (commented):
Another tip: it's better to give an obvious reminder that gang need to close if this feature enabled

pkg/controllers/job/job_controller_actions.go
hwdef: How about we write this in webhook?

pkg/controllers/job/helpers/helpers.go
william-wang: Seems "GetTaskIndex" is the best for two places:)  Let's use GetTask(pod *v1.Pod) here and use the GetTaskIndex(taskName, job) for your new code change.

hwdef: @william-wang ok, that loosk good.

hwdef:
@Thor-wl @shinytang6 @william-wang 

Most of the review comments have been resolved, and the following questions remain:

* What is the appropriate setting for the detection interval? It is currently one nanosecond, this time may be too short https://github.com/volcano-sh/volcano/pull/1833/files#diff-28557d59525e85335ee0eacff9269dcc0dbc0ffcc7ef146ef4e0e59d713df788R473
* Prompt need to disable gang plugin.Here we have two options, one is to limit the minavailable value to the  reasonable, let the Gang plugin can be retained, and the other is to disable the Gang plugin. No matter which kind of choice, we are best completed in Webhook.
* We still have a bug here, as I demonstrated at the weekly meeting. Create the same name Job when Job is not completely deleted. After many attempts, I think this bug is caused by this PR, did not exist before. But I haven't found the reason for this bug yet.Maybe fix this bug later, first merge PR, so we can not block other PR,#1874

pkg/controllers/job/job_controller_actions.go
```diff
@@ -452,6 +468,78 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+						return true, nil
+					}
+					return false, nil
+				})
+			}
+		}
+	}
+}
+
+func (cc *jobcontroller) isDenpendentPodsReady(DependsOnTask string, job *batch.Job) bool {
```
shinytang6, line 497: isDenpendentPodsReady  => isDependentPodsReady
DependsOnTask => task

pkg/controllers/job/job_controller_actions.go
```diff
@@ -452,6 +468,78 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 	return nil
 }
 
+func (cc *jobcontroller) waitDependsOnTaskMeetCondition(taskName string, podToCreateEachTask []*v1.Pod, job *batch.Job) {
+	taskIndex := jobhelpers.GetTaskIndex(taskName, job)
+	if job.Spec.Tasks[taskIndex].DependsOn != nil {
+		dependsOn := *job.Spec.Tasks[taskIndex].DependsOn
+		if len(dependsOn.Name) > 1 && dependsOn.Iteration == batch.IterationAny {
+			wait.PollInfinite(100*time.Millisecond, func() (bool, error) {
+				for _, DependsOnTask := range dependsOn.Name {
```
shinytang6, line 477: DependsOnTask => task

pkg/controllers/job/job_controller_actions.go
```diff
@@ -452,6 +468,78 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+					}
+					return false, nil
+				})
+			}
+		}
+	}
+}
+
+func (cc *jobcontroller) isDenpendentPodsReady(DependsOnTask string, job *batch.Job) bool {
+	dependentPod := jobhelpers.GetPodsInTask(DependsOnTask, job)
```
shinytang6, line 498: dependentPod => dependentPods
And it is better to unify using dependon or dependent, not mixed

pkg/controllers/job/job_controller_actions.go
```diff
@@ -452,6 +468,78 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+				}
+				return false
+			}
+			klog.Errorf("Failed to get pod %v/%v %v", job.Namespace, podName, err)
+			return false
+		}
+
+		if pod.Status.Phase != v1.PodRunning {
+			klog.V(4).Infof("Sequential state, pod %v/%v of dependent tasks is not running", pod.Namespace, pod.Name)
+			continue
```
shinytang6, line 519: what if all the pods under that task become succeeded quickly?

pkg/controllers/job/job_controller_actions.go
```diff
@@ -452,6 +468,78 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+				_, err := cc.vcClient.BatchV1alpha1().Jobs(job.Namespace).Get(context.TODO(), job.Name, metav1.GetOptions{})
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						return true
+					}
+				}
+				return false
+			}
+			klog.Errorf("Failed to get pod %v/%v %v", job.Namespace, podName, err)
+			return false
```
shinytang6, line 514: l'm a little confused about this block.
1. why need to get job if this pod not found?
2. if pod is not found and job exists, do we need to return false immediately?

shinytang6 (changes requested):

pkg/controllers/job/helpers/helpers.go
```diff
@@ -34,8 +35,8 @@ const (
 	persistentVolumeClaimFmt = "%s-pvc-%s"
 )
 
-// GetTaskIndex returns task Index.
-func GetTaskIndex(pod *v1.Pod) string {
+// GetTask returns task Index.
+func GetTask(pod *v1.Pod) string {
```
shinytang6, line 39: And l think this func can be named more clearly, maybe `getPodIndexUnderTask`?

pkg/controllers/job/helpers/helpers.go
shinytang6: And l think this func can be named more clearly, maybe `getTasklndexUnderJob `?

pkg/controllers/job/helpers/helpers.go
```diff
@@ -70,3 +71,27 @@ func GenPVCName(jobName string) string {
...
+	for index, task := range job.Spec.Tasks {
+		if task.Name == taskName {
+			return index
+		}
+	}
+	return -1
+}
+
+// GetPodsInTask return names of all pods in the task.
+func GetPodsInTask(taskName string, job *batch.Job) []string {
```
shinytang6, line 86: And l think this func can be named more clearly, maybe `GetPodsNameUnderTask `?

shinytang6 (changes requested):

pkg/controllers/job/job_controller_actions.go
hwdef: I agree with what you said, which one do you think is better?

pkg/controllers/job/job_controller_actions.go
hwdef: The current situation is that the following tasks will stop and will never be run.
This should be dependent on completion status, we can add a field later:

```
dependsOn:
    name: xxxxx
    status: running/complete
```


I think we can also deal with it now, if it is completed, allow it to pass.

```go
pod.Status.Phase != v1.PodRunning || pod.Status.Phase != v1.PodSucceeded
```

pkg/controllers/job/helpers/helpers.go
hwdef: Please discuss the name of this function with @william-wang, I think both are ok

pkg/controllers/job/job_controller_actions.go
```diff
@@ -349,33 +354,44 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+						// Failed to create Pod, waitCreationGroup a moment and then create it again
+						// This is to ensure all podsMap under the same Job created
+						// So gang-scheduling could schedule the Job successfully
+						klog.Errorf("Failed to create pod %s for Job %s, err %#v",
+							pod.Name, job.Name, err)
+						appendError(&creationErrs, fmt.Errorf("failed to create pod %s, err: %#v", pod.Name, err))
+					} else {
+						classifyAndAddUpPodBaseOnPhase(newPod, &pending, &running, &succeeded, &failed, &unknown)
+						calcPodStatus(pod, taskStatusCount)
+						klog.V(3).Infof("Created Task <%s> of Job <%s/%s>",
```
william-wang, line 387: This log is in for loop. it's better use log level 5.

william-wang:
@hwdef is it possible to add e2e case for this new feature?

hwdef:
> @hwdef is it possible to add e2e case for this new feature?

I am trying to add e2e, but due to the characteristics of this feature, I feel a little difficult.(It may also be because I am not familiar with e2e)

pkg/controllers/job/job_controller_actions.go
```diff
@@ -355,33 +360,44 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
-				klog.V(3).Infof("Created Task <%s> of Job <%s/%s>",
-					pod.Name, job.Namespace, job.Name)
+	for taskName, podToCreateEachTask := range podToCreate {
+		if len(podToCreateEachTask) == 0 {
+			continue
+		}
+		go func(taskName string, podToCreateEachTask []*v1.Pod) {
+			taskIndex := jobhelpers.GetTasklndexUnderJob(taskName, job)
+			if job.Spec.Tasks[taskIndex].DependsOn != nil {
+				cc.waitDependsOnTaskMeetCondition(taskName, podToCreateEachTask, job)
```
william-wang, line 376: for each task with DependsOn, GetTaskIndexUnderJob is called twice(Line374 and Line478). It's better take taskIndex as input of waitDependsOnTaskMeetCondition.

pkg/controllers/job/job_controller_actions.go
```diff
@@ -458,6 +474,69 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
...
+				for _, task := range dependsOn.Name {
+					if cc.isDependsOnPodsReady(task, job) {
+						return true, nil
+					}
+				}
+				return false, nil
+			})
+		} else {
+			for _, dependsOnTask := range dependsOn.Name {
+				wait.PollInfinite(100*time.Millisecond, func() (bool, error) {
```
william-wang, line 492: it's better to make 100ms configurable. At least define a const for it if no time :)

pkg/controllers/job/job_controller_actions.go
hwdef: done.

pkg/controllers/job/job_controller_actions.go
hwdef: I implemented configurable parameters. But I don't know if the implementation is elegant, especially the process of variable transmission. In addition, the naming of variables is also very long. If you have any comments, please let me know, thank you!

william-wang (approved):
/lgtm

PR #1821, opened 2021-11-04
support reserved node

qiankunli:
for example, my k8s cluster has 10 nodes (10 gpu per node), 5 for training, 5 for serving, I want use volcano schedule tfjob/pytorchjob/vcjob on training nodes. use default-scheduler  schedule online pod on serving nodes.

it is work by setting nodeSelector or nodeAffinity to pod,  but it is not properly when considering queue.

if there are  two queue: queue1(weight=1) and queue2 (weight=1),  it is overused if queue1 used 40 gpu, but it's weight  resource is 50, so if queue2 want use 20 gpu, reclaimable mechanism will not work.

so it is necessary to limit volcano scheduler that  it is can only work on training nodes, not all nodes in cluster.  

so I add `work-node-label` args for volcano scheduler, in following example,  volcano can only work on the node which has label `node-role:training` or `node-role:dev`. you can use any label name you like.

```yaml
kind: Deployment
apiVersion: apps/v1
metadata:
  name: volcano-scheduler
  namespace: xdl-system
  labels:
    app: volcano-scheduler
spec:
  replicas: 1
  selector:
    matchLabels:
      app: volcano-scheduler
  template:
    metadata:
      labels:
        app: volcano-scheduler
    spec:
      serviceAccount: volcano-scheduler
      containers:
        - name: volcano-scheduler
          image: xx/vc-scheduler:v1.4.0-node
          args:
            - --work-node-label=node-role:training
            - --work-node-label=node-role:dev
            - --logtostderr
            - --scheduler-conf=/volcano.scheduler/volcano-scheduler.conf
            - -v=5
            - 2>&1
          imagePullPolicy: "IfNotPresent"
          volumeMounts:
            - name: scheduler-config
              mountPath: /volcano.scheduler
      volumes:
        - name: scheduler-config
          configMap:
            name: volcano-scheduler-configmap
```

PR #1834, opened 2021-11-16
[fix]add nodeSelector design doc

qiankunli:
Signed-off-by: bert.li <qiankun.li@qq.com>
design doc for #1821


docs/design/node-selector.md
```diff
@@ -0,0 +1,37 @@
+## Introduction
+
+for example, my k8s cluster has 10 nodes (10 gpu per node), 5 for training, 5 for serving, I want to use volcano schedule tfjob/pytorchjob/vcjob on training nodes. use default-scheduler schedule online pod on serving nodes. it is work by setting nodeSelector or nodeAffinity to pod, but it is not properly when considering volcano queue mechanism.
+
+if there are two queue: queue1(weight=1) and queue2 (weight=1), it is overused if queue1 used 40 gpu, but it's weight resource is 50 (100 * 1/2), so if queue2 want use 20 gpu, reclaimable mechanism will not work.
```
Thor-wl, line 5: If there are totally 100 GPUs in the cluster, the deserved resource for queue1 and queue2 are both 50 GPUS. Why is it overused if queue1 used 40 GPUs? BTW, please make the format and grammar of the doc more canonical.

docs/design/node-selector.md
```diff
@@ -0,0 +1,37 @@
...
+
+so it is necessary to limit volcano scheduler that it can only work on training nodes, not all nodes in cluster.
+
+so I add nodeSelector for volcano scheduler.
+
+## Usage
+
+in following example, volcano can only work on the node which has `node-role:training` or `node-role:dev` or `gpu-model: tesla` label. you can use any label name you like.
+
+```yaml
```
Thor-wl, line 15: Sugeest to add comment about the configuration to point out what should be foucsed on.

docs/design/node-selector.md
```diff
@@ -0,0 +1,37 @@
...
+      - name: conformance
+    - plugins:
+      - name: overcommit
+      - name: drf
+      - name: predicates
+      - name: proportion
+      - name: nodeorder
+      - name: binpack
+    nodeSelector:
+      node-role: training,dev
```
Thor-wl, line 30: Pls separate `node-role` to different items

docs/design/node-selector.md
```diff
@@ -0,0 +1,37 @@
...
+      gpu-model: tesla
+```
+
+## Design
+
+volcano use nodeInformer to get all node list in k8s cluster. so we can filter the node list by nodeSelector label.
```
Thor-wl, line 36: This is implementation part and suggest to give more details about what should be done clearly.

shinytang6 (changes requested):
Two suggestions:
1. pods should also be filtered by specified nodes
2. I recommend putting the configuration of node-selector in the scheduler parameters

Thor-wl:
> Two suggestions:
> 
> 1. pods should also be filtered by specified nodes
> 2. I recommend putting the configuration of node-selector in the scheduler parameters

If node-selector is put as scheduer parameters, may it be not so flexible when make modifications?

shinytang6:
> > Two suggestions:
> > 
> > 1. pods should also be filtered by specified nodes
> > 2. I recommend putting the configuration of node-selector in the scheduler parameters
> 
> If node-selector is put as scheduer parameters, may it be not so flexible when make modifications?

The configuration file is hot updated, but it will not resync after the update, which may cause some of the changed nodes not be added to the cache

qiankunli:
> > > Two suggestions:
> > > 
> > > 1. pods should also be filtered by specified nodes
> > > 2. I recommend putting the configuration of node-selector in the scheduler parameters
> > 
> > 
> > If node-selector is put as scheduer parameters, may it be not so flexible when make modifications?
> 
> The configuration file is hot updated, but it will not resync after the update, which may cause some of the changed nodes not be added to the cache
I have changed the `schedule.conf` file watch code to watch the changed nodeSelector config

qiankunli:
I have changed the schedule.conf file watch code to watch the changed nodeSelector config

Thor-wl:
> > > Two suggestions:
> > > 
> > > 1. pods should also be filtered by specified nodes
> > > 2. I recommend putting the configuration of node-selector in the scheduler parameters
> > 
> > 
> > If node-selector is put as scheduer parameters, may it be not so flexible when make modifications?
> 
> The configuration file is hot updated, but it will not resync after the update, which may cause some of the changed nodes not be added to the cache

Well, yes, that may be a problem.

shinytang6 (commented):
lgtm overall

jasonliu747 (commented):
lgtm overall, except some grammar issues.
I'll give more comments on Monday~

PR #1848, opened 2021-11-24
fix: avoid chan block within taintTolerationScore

justadogistaken:
While learning the scheduling process, I found that there maybe a problem which could block the progress of scheduling. The easiest way to fix this problem is scale the size of errch up to len(nodes).


Thor-wl:
/lgtm
FOSSA seems not work today. Don't worry. Will merge after it is fixed.

shinytang6 (changes requested):
`interPodAffinityScore` has the same problem

shinytang6 (approved):
Thanks!
/lgtm
/approve

PR #1849, opened 2021-11-24
support to set healthz address and metrics address

huone1:

it is  insecure to bind address 0.0.0.0， so it is better to support to set the listen address 
it is also supported to set them in k8s arguments

for example
kubectl edit deploy volcano-controllers -n volcano-system
set the pod ip as the listen IP
> spec:
      containers:
      - args:
        - --logtostderr
        - --healthz-address=$(MY_POD_IP):11251
        - -v=4
        - 2>&1
        env:
        - name: MY_POD_IP
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: status.podIP

image

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -25,11 +25,12 @@ import (
 )
 
 const (
-	defaultQPS           = 50.0
-	defaultBurst         = 100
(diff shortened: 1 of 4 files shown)


Thor-wl:
Any background? Suggest to give a brief guidence for users.
/lgtm

Issue #1850, opened 2021-11-24
pod cannot be allocated with sufficient resource


aidaizyy:
**What happened**:
I found a scheduling problem about proportion, I constructed the environment to reproduce it stably.

My scheduler configuration (default conf):
``` yaml
actions: "enqueue, allocate, backfill"
tiers:
- plugins:
  - name: priority
  - name: gang
  - name: conformance
- plugins:
  - name: overcommit
  - name: drf
  - name: predicates
  - name: proportion
  - name: nodeorder
  - name: binpack
```

My cluster has two node: 
``` yaml
node1 
  - cpu: 1
  - memory: 3237024Ki
node2
  - cpu: 20
  - memory: 108459168Ki
```

and two queues: 
``` yaml
queue1
  - weight: 1
  - capability
    - cpu: 15
    - memory: 30Gi
queue2
  - weight: 1
  - capability
    - cpu: 5
    - memory: 70Gi
```

There are already two jobs: 
``` yaml
deploy-with-volcano
  - queue1
  - only one pod (request cpu: 11, request memory: 1Gi)
deploy-with-volcano2
  - queue2
  - only one pod (request cpu: 1, request memory: 60Gi)
```

Next I am going to submit other jobs:
``` yaml
deploy-with-volcano3
  - queue1
  - only one pod (request cpu: 1, request memory: 1Gi)
deploy-with-volcano4
  - queue2
  - only one pod (request cpu: 1, request memory: 1Gi)
```
But no matter deploy-with-volcano3 or deploy-with-volcano4, it cannot be scheduled.

``` bash
NAME                                    READY   STATUS    RESTARTS   AGE
deploy-with-volcano-54d568854c-77fhb    1/1     Running   0          79s
deploy-with-volcano2-85f8cb54d7-55dcn   1/1     Running   0          79s
deploy-with-volcano3-5476c9cd69-dfv57   0/1     Pending   0          54s
deploy-with-volcano4-6f6579f745-8s2f6   0/1     Pending   0          54s
```

**What you expected to happen**:
deploy-with-volcano3/deploy-with-volcano4 only request cpu 1, memory 1Gi.
queue 1: allocated (cpu: 11, memory: 1Gi), capability (cpu:15, memory:30Gi), has sufficient resources to start the deploy-with-volcano3;
queue 2: allocated (cpu: 1, memory: 60Gi), capability (cpu:5, memory:70Gi), has sufficient resources to start the deploy-with-volcano4;
[...]

PR #1851, opened 2021-11-24
fix pod cannot be allocated with sufficient resource

aidaizyy:
background: https://github.com/volcano-sh/volcano/issues/1850

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -480,7 +480,8 @@ func (r *Resource) SetScalar(name v1.ResourceName, quantity float64) {
 // e.g r resource is <cpu 2000.00, memory 4047845376.00, hugepages-2Mi 0.00, hugepages-1Gi 0.00>
 // rr resource is <cpu 3000.00, memory 1000.00>
 // return r resource is <cpu 2000.00, memory 1000.00, hugepages-2Mi 0.00, hugepages-1Gi 0.00>
-func (r *Resource) MinDimensionResource(rr *Resource) *Resource {
+// @param defaultValue "default value for resource dimension not defined in ScalarResources. Its value can only be one of 'Zero' and 'Infinity'"
+func (r *Resource) MinDimensionResource(rr *Resource, defaultValue DimensionDefaultValue) *Resource {
 	if rr.MilliCPU < r.MilliCPU {
 		r.MilliCPU = rr.MilliCPU
 	}
@@ -488,19 +489,31 @@ func (r *Resource) MinDimensionResource(rr *Resource) *Resource {
 		r.Memory = rr.Memory
 	}
 
+	if r.ScalarResources == nil {
+		return r
+	}
+
 	if rr.ScalarResources == nil {
-		if r.ScalarResources != nil {
-			for name := range r.ScalarResources {
-				r.ScalarResources[name] = 0
-			}
+		if defaultValue == Infinity {
+			return r
 		}
-	} else {
-		if r.ScalarResources != nil {
-			for name, quant := range rr.ScalarResources {
-				if quant < r.ScalarResources[name] {
-					r.ScalarResources[name] = quant
-				}
+
+		for name := range r.ScalarResources {
+			r.ScalarResources[name] = 0
+		}
+		return r
+	}
+
+	for name, quant := range r.ScalarResources {
+		rQuant, ok := rr.ScalarResources[name]
+		if ok {
+			r.ScalarResources[name] = math.Min(quant, rQuant)
+		} else {
+			if defaultValue == Infinity {
+				continue
 			}
+
+			r.ScalarResources[name] = 0
 		}
 	}
 	return r
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -17,6 +17,7 @@ limitations under the License.
 package proportion
 
 import (
+	"math"
 	"reflect"
 
 	v1 "k8s.io/api/core/v1"
@@ -90,6 +91,12 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			}
 			if len(queue.Queue.Spec.Capability) != 0 {
 				attr.capability = api.NewResource(queue.Queue.Spec.Capability)
+				if attr.capability.MilliCPU <= 0 {
+					attr.capability.MilliCPU = math.MaxFloat64
+				}
+				if attr.capability.Memory <= 0 {
+					attr.capability.Memory = math.MaxFloat64
+				}
 			}
 
 			pp.queueOpts[job.Queue] = attr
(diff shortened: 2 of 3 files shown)


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -159,9 +159,11 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 
 			oldDeserved := attr.deserved.Clone()
 			attr.deserved.Add(remaining.Clone().Multi(float64(attr.weight) / float64(totalWeight)))
-
-			if attr.capability != nil && !attr.deserved.LessEqual(attr.capability, api.Infinity) {
+			if attr.capability != nil {
 				attr.deserved = helpers.Min(attr.deserved, attr.capability)
+			}
+
+			if attr.capability != nil && attr.capability.LessEqual(attr.deserved, api.Zero) {
```
Thor-wl, line 166: What if some dimensions are lack in some queue's capability? For example, capability of q1 = {cpu:10c}, but there must be `cpu` and `memory` in deserved.

aidaizyy: In the previous code, `helpers.Min` also make the deserved memory is 0.
``` go
if attr.capability != nil && !attr.deserved.LessEqual(attr.capability, api.Infinity) {
        attr.deserved = helpers.Min(attr.deserved, attr.capability)
……
```
In the current semantics, this situation means that the memory deserved is 0 always.
I think this pr didn't make it worse.

But if necessary, I can fix this problem in this pr or another pr soon.

Thor-wl: I get your meaning. In fact, deserved cpu and memory will never be zero for they are necessary dimensions. What I want to express is that lack dimension in capability should be in infinity.

aidaizyy:
@Thor-wl pls review agin.
I reorganized the code here.
Not only solved the scheduling bugs I found. I also fixes the previously unresolved problems you mentioned.
Now, no matter whether it is cpu, memory or scalar resources, If they are not set, they are regarded as infinity.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -160,18 +160,18 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
...
 			attr.deserved.Add(remaining.Clone().Multi(float64(attr.weight) / float64(totalWeight)))
 
-			if attr.capability != nil && !attr.deserved.LessEqual(attr.capability, api.Infinity) {
-				attr.deserved = helpers.Min(attr.deserved, attr.capability)
-				attr.deserved = helpers.Min(attr.deserved, attr.request)
-				meet[attr.queueID] = struct{}{}
-				klog.V(4).Infof("queue <%s> is meet cause of the capability", attr.name)
-			} else if attr.request.LessEqual(attr.deserved, api.Zero) {
-				attr.deserved = helpers.Min(attr.deserved, attr.request)
+			if attr.capability != nil {
```
eggiter, line 163: Sorry, which part of change fixes your problem? i'm not quite following you.

eggiter (changes requested):

pkg/scheduler/plugins/proportion/proportion.go
aidaizyy: refer to https://github.com/volcano-sh/volcano/issues/1850

aidaizyy: The calculated deserved value is wrong. 
In the process of calculating the deserved value, only some resources of deserved exceeds the capability and should not be stopped. Because other resources which smaller than capaliby may increase.

eggiter: Got it. I've noticed that `meet[attr.queueID] = struct{}{}` is removed when `!attr.deserved.LessEqual(attr.capability, api.Infinity)`.

pkg/scheduler/api/resource_info.go
```diff
@@ -480,26 +480,28 @@ func (r *Resource) SetScalar(name v1.ResourceName, quantity float64) {
...
 				r.ScalarResources[name] = 0
 			}
 		}
-	} else {
-		if r.ScalarResources != nil {
-			for name, quant := range rr.ScalarResources {
-				if quant < r.ScalarResources[name] {
-					r.ScalarResources[name] = quant
-				}
+	} else if r.ScalarResources != nil {
```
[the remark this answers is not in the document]
aidaizyy: done

Thor-wl:
Pls fix the DCO @aidaizyy

pkg/scheduler/api/resource_info.go
```diff
@@ -480,26 +480,28 @@ func (r *Resource) SetScalar(name v1.ResourceName, quantity float64) {
 // e.g r resource is <cpu 2000.00, memory 4047845376.00, hugepages-2Mi 0.00, hugepages-1Gi 0.00>
 // rr resource is <cpu 3000.00, memory 1000.00>
 // return r resource is <cpu 2000.00, memory 1000.00, hugepages-2Mi 0.00, hugepages-1Gi 0.00>
-func (r *Resource) MinDimensionResource(rr *Resource) *Resource {
-	if rr.MilliCPU < r.MilliCPU {
-		r.MilliCPU = rr.MilliCPU
+// @param defaultValue "default value for resource dimension not defined in CPU/Memory/ScalarResources. Its value can only be one of 'Zero' and 'Infinity'"
```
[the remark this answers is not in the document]
aidaizyy: There has never been a default value for the main resource (cpu/memory) when there is no definition in the project, but this is meaningful for the capability. I am more embarrassed how to express, please review whether my latest revision is reasonable.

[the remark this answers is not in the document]
eggiter: The better way to achieve that is to set the default value of `capacity.cpu`, `capacity.memory` is the admission web-hook or the `AddQueue` in the `event_handler`. cc @Thor-wl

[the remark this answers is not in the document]
aidaizyy: Good suggestion! I set `math.maxFloat64` for cpu and memory in OnSessionOpen. This modification should be the best with few changes.

pkg/scheduler/api/resource_info.go
```diff
@@ -480,27 +480,40 @@ func (r *Resource) SetScalar(name v1.ResourceName, quantity float64) {
...
+// @param defaultValue "default value for resource dimension not defined in ScalarResources. Its value can only be one of 'Zero' and 'Infinity'"
+func (r *Resource) MinDimensionResource(rr *Resource, defaultValue DimensionDefaultValue) *Resource {
 	if rr.MilliCPU < r.MilliCPU {
 		r.MilliCPU = rr.MilliCPU
 	}
 	if rr.Memory < r.Memory {
 		r.Memory = rr.Memory
 	}
 
+	if r.ScalarResources == nil {
```
eggiter, line 492: no needs for the `== nil` comparation.

aidaizyy: Sorry, I do not understand what you mean.
I think it is necessary. Otherwise, `for name := range r.ScalarResources ` will cause panic.

aidaizyy: This is a kind-hearted redundancy. :)

pkg/scheduler/api/resource_info.go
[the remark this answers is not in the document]
Thor-wl: I agree with @eggiter. Perhaps we can do that at another PR. Can you help for that? @aidaizyy

PR #1856, opened 2021-11-25
improve taintTolerationScore interPodAffinityScore throghput when failure occurs

justadogistaken:
I guess there is a work that could help improve taintTolerationScore & interPodAffinityScore throghput at failure cases.
This could avoid invalid computation when failure occurs, especially in huge cluster.
I think this could help enhance the scheduling throughput
refer to #1740

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -316,13 +316,24 @@ func interPodAffinityScore(
 	}
 
 	nodescoreList := make(k8sframework.NodeScoreList, len(nodes))
-	errCh := make(chan error, len(nodes))
-	workqueue.ParallelizeUntil(context.TODO(), 16, len(nodes), func(index int) {
+	// the default parallelization worker number is 16.
+	// the whole scoring will fail if one of the processes failed.
+	// so just create a parallelizeContext to control the whole ParallelizeUntil process.
+	// if the parallelizeCancel is invoked, the whole "ParallelizeUntil" goes to the end.
+	// this could avoid extra computation, especially in huge cluster.
+	// and the ParallelizeUntil guarantees only "workerNum" goroutines will be working simultaneously.
+	// so it's enough to allocate workerNum size for errCh.
+	// note that, in such case, size of errCh should be no less than parallelization number
+	workerNum := 16
+	errCh := make(chan error, workerNum)
+	parallelizeContext, parallelizeCancel := context.WithCancel(context.TODO())
+	workqueue.ParallelizeUntil(parallelizeContext, workerNum, len(nodes), func(index int) {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -368,13 +379,17 @@ func taintTolerationScore(
 	}
 
 	nodescoreList := make(k8sframework.NodeScoreList, len(nodes))
-	errCh := make(chan error, len(nodes))
-	workqueue.ParallelizeUntil(context.TODO(), 16, len(nodes), func(index int) {
+	// size of errCh should be ge than parallelization number, see interPodAffinityScore.
```
hwdef, line 382: what is `ge`?

justadogistaken: "greater or equal"
I can make it more clear

hwdef: That's the better. I think there is no need to abbreviate here.

justadogistaken: Got it!

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -316,13 +316,24 @@ func interPodAffinityScore(
...
-	workqueue.ParallelizeUntil(context.TODO(), 16, len(nodes), func(index int) {
+	// the default parallelization worker number is 16.
+	// the whole scoring will fail if one of the processes failed.
+	// so just create a parallelizeContext to control the whole ParallelizeUntil process.
+	// if the parallelizeCancel is invoked, the whole "ParallelizeUntil" goes to the end.
+	// this could avoid extra computation, especially in huge cluster.
+	// and the ParallelizeUntil guarantees only "workerNum" goroutines will be working simultaneously.
+	// so it's enough to allocate workerNum size for errCh.
+	// note that, in such case, errCh should be ge than parallelization number
+	workerNum := 16
```
hwdef, line 327: why use this value?

justadogistaken: 16 or workerNum?
16 is the default Parallelization number from volcano.
For "workerNum", I want to express the correlation between size of errCh and Parallelization number. And the min(Parallelization number, len(nodes)) for size of errCh is the least number for main goroutine safety.
But I don't want to make it too complicated, so I did't do min(Parallelization number, len(nodes)) for size of errCh

hwdef: ok, I got it.

hwdef:
Looks good in general.
/lgtm

shinytang6 (approved):
good catch, thx
/lgtm
/approve

PR #1862, opened 2021-11-29
optimize the CA parse Process

huone1:
For some users, they Maybe use their own certification files with specific encryption for security in webhook‘s register.
So add one interface ParseCAFiles to parse CA data and the argument decryptFunc is for the decryption .

### DIFF
--- a/cmd/webhook-manager/main.go
+++ b/cmd/webhook-manager/main.go
@@ -58,6 +58,10 @@ func main() {
 		klog.Fatalf("Configured port is invalid: %v", err)
 	}
 
+	if err := config.ParseCAFiles(nil); err != nil {
+		klog.Fatalf("Failed to parse CA file: %v", err)
+	}
+
 	if err := app.Run(config); err != nil {
 		fmt.Fprintf(os.Stderr, "%v\n", err)
 		os.Exit(1)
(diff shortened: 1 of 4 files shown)


william-wang:
@huone1 Please add the backgroud and needs of this pr. Thanks.

huone1:
> @huone1 Please add the backgroud and needs of this pr. Thanks.

added it

cmd/webhook-manager/main.go
```diff
@@ -58,6 +58,10 @@ func main() {
 		klog.Fatalf("Configured port is invalid: %v", err)
 	}
 
+	if err := config.ReadCAFiles(); err != nil {
+		klog.Fatalf("Failed to read CA files: %v", err)
+	}
```
william-wang, line 63: please clarify what volcano user should do if the cert file is encripted. It's better to add a decrypt func.

cmd/webhook-manager/main.go
```diff
@@ -58,6 +58,10 @@ func main() {
...
+
 	if err := app.Run(config); err != nil {
```
xiaoanyunfei, line 65: I think It's better to call `config.ReadCAFiles()` in `cmd/webhook-manager/app/server.go`

cmd/webhook-manager/main.go
huone1: > 

optimized it. the ParseCAFiles obtain the action to read files and the action to decrypt the data if a decrypt func is as a argument

william-wang (approved):
/lgtm

Issue #1864, opened 2021-11-29
Image Locality Plugin


justadogistaken:
#### What would you like to be added:
I happened to found the TODO left in volcano nodeorder. which is about ImageLocality.
```		
//TODO: Add ImageLocalityPriority Function once priorityMetadata is published
//Issue: #74132 in kubernetes ( https://github.com/kubernetes/kubernetes/issues/74132 )
```

Wonder if the "ImageLocality" is not important in batch-scheduling, so volcano never fills this TODO. That is easy to complete nowadays.

shinytang6:
> TODO: Add ImageLocalityPriority Functio

This is needed, but no one filled it. Would you like to have a try? @justadogistaken

Thor-wl:
It is very useful for accelerating the speed of starting pod without pulling big image. So if scheduler can be aware of the location of images on all nodes, it will be helpful.

justadogistaken:
> without

I'd like to work on this.

PR #1868, opened 2021-11-30
feat: add imagelocality priority to nodeOrder

justadogistaken:
refer to #1864 

k8s default score weight as following:
```
	Score: config.PluginSet{
		Enabled: []config.Plugin{
			{Name: names.NodeResourcesBalancedAllocation, Weight: 1},
			{Name: names.ImageLocality, Weight: 1},
			{Name: names.InterPodAffinity, Weight: 1},
			{Name: names.NodeResourcesFit, Weight: 1},
			{Name: names.NodeAffinity, Weight: 1},
			// Weight is doubled because:
			// - This is a score coming from user preference.
			// - It makes its signal comparable to NodeResourcesLeastAllocated.
			{Name: names.PodTopologySpread, Weight: 2},
			{Name: names.TaintToleration, Weight: 1},
		},
	},
```

note: But I didn't find the e2e test about the nodeorder yet.

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/scheduler/apis/config"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/imagelocality"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/interpodaffinity"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodeaffinity"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/noderesources"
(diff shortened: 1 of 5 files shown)


pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/imagelocality"
```
huone1, line 23: move it to line 28

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -113,6 +119,7 @@ func calculateWeight(args framework.Arguments) priorityWeight {
 		podAffinityWeight:      1,
 		balancedResourceWeight: 1,
 		taintTolerationWeight:  1,
+		imageLocalityWeight:    0,
```
shinytang6, line 122: Is this plugin disabled by default in k8s scheduler?

justadogistaken: Oh, I forgot to check how k8s deals with this. I thought we leave this decision to user, but I guess user will never overwrite the priority config. I will modify the weight.

pkg/scheduler/plugins/nodeorder/nodeorder.go
justadogistaken: Got it!

k82cn:
LGTM overall, thanks for the PR :)

Thor-wl:
LGTM overall

hwdef:
looks good!

shinytang6 (approved):
/lgtm
/approve

PR #1875, opened 2021-12-03
add a argument csi-storage to control the storage capacity resource

huone1:
1. add a argument csi-storage to control the storage capacity resource, because storage capacity is  disable by default in k8s 1.19 and is enable  by default in k8s 1.21
   if volcano is installed in k8s 1.19, it will happen as the following:
   
image

  
2. delete the  repeated rules


Thor-wl:
After the k8s vendor is upgraded to v1.22, it will be open by default. Then I will remove the parameter. It works for v1.19 temporarily.
/lgtm

william-wang (approved):
/lgtm

PR #1880, opened 2021-12-04
rm JobValid in openSession

xiaoanyunfei:
There is not any `jobValidFns` registered during `Session` creating


Thor-wl:
IMO, reserving it is also OK for it is more clear for new contributors. Also, some end user can specify their scenario here.

hwdef:
> IMO, reserving it is also OK for it is more clear for new contributors. Also, some end user can specify their scenario here.

+1

PR #1888, opened 2021-12-06
add healthz and metric switch for deploy controller and scheduler

huone1:
#1889 
For some users,it is  unnecessary and insecure to enable the metrics  service if they needn't the metrics ; And the  healthz check is also so on.


william-wang (approved):
/lgtm

Issue #1853, opened 2021-11-24
The podgroup is not deleted after the pod created by Deployment is deleted 


hansongChina:
**What happened**:
When changing Deployment triggers the creation of a new ReplicaSet and the pod created by the old ReplicaSet is deleted, the podgroup of the old pod is not deleted

**What you expected to happen**:

the podgroup of the old pod is deleted

**How to reproduce it (as minimally and precisely as possible)**:

- step1:Add the default volcano to pod using admission Webhooks, for example

`patches = append(patches, patchOperation{
			Op:    "add",
			Path:  "/spec/schedulerName",
			Value: "volcano",
		})`
- step2:Create a Deployment,for example
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```
```shell
kubectl apply -f nginx-deployment.yaml
```
![image](https://user-images.githubusercontent.com/7590000/143237845-f6fecac1-f7fc-4673-9b72-616f5785cba1.png)
```shell
kubectl edit pod nginx-deployment-6b474476c4-v6k4x
```
![image](https://user-images.githubusercontent.com/7590000/143237933-6709bed5-d3b1-4ae2-b90d-534b2f0ce9ce.png)
```shell
kubectl edit pg  podgroup-6f08b2ae-296e-43a1-825c-84e483c3ea89
```
![image](https://user-images.githubusercontent.com/7590000/143238061-72dc2155-8298-4a43-bf6e-9aa2bc20a917.png)
- step3: Modify "image:nginx:1.14.2" in Deployment to trigger the creation of a new RS and a new POD. The old pod is deleted and the podgroup associated with the old POD still exists in the state of inqueue
![image](https://user-images.githubusercontent.com/7590000/143239373-fcd60614-ba7a-43d5-9f27-0d7875b98376.png)
```shell
kubectl edit pg  podgroup-6f08b2ae-296e-43a1-825c-84e483c3ea89
```
![image](https://user-images.githubusercontent.com/7590000/143240533-ba236ce3-d069-4b2f-a91f-2e35ea65bc36.png)

**Anything else we need to know?**:

<font color=red>kind is workflow runs into the same problem</font>

[...]

PR #1891, opened 2021-12-06
fix pg not deleted after pod created by deployment is deleted

lucming:
#https://github.com/volcano-sh/volcano/issues/1853


pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -49,6 +49,24 @@ func (pg *pgcontroller) addPod(obj interface{}) {
...
+	if _, err := pg.pgLister.PodGroups(pod.Namespace).Get(pgName); err != nil {
+		return
+	}
+
+	if err := pg.vcClient.SchedulingV1beta1().PodGroups(pod.Namespace).Delete(context.TODO(), pgName, metav1.DeleteOptions{}); err != nil {
+		klog.Errorf("Failed to delete volcano job,ns:%s,name:%s,err:", pod.Namespace, pgName, err)
+		return
+	}
+}
+
```
shinytang6, line 69: i'm ok about the changes, but is there a way to filter out the pods under vcjob? Because currently all pods are watched, but when the pods under vcjob are deleted, deleting pg will throw an error

shinytang6 (changes requested):
i'm ok about the changes, but is there a way to filter out the pods under cjob? Because currently all pods are watched, but when the pods under vcjob are deleted, deleting pg will throw an error

pkg/controllers/podgroup/pg_controller_handler.go
Thor-wl: Agree with @shinytang6 .Without the filter, just delete the pg is not so reasonable when some pod is deleted.

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -76,7 +76,8 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
 	pg.podLister = pg.podInformer.Lister()
 	pg.podSynced = pg.podInformer.Informer().HasSynced
 	pg.podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: pg.addPod,
+		AddFunc:    pg.addPod,
+		DeleteFunc: pg.deletePG,
```
hzxuzhonghu, line 80: This is not good to call some time costing logic in eventhandler.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -49,6 +49,24 @@ func (pg *pgcontroller) addPod(obj interface{}) {
...
+		klog.Errorf("Failed to convert %v to v1.Pod", obj)
+		return
+	}
+
+	pgName := helpers.GeneratePodgroupName(pod)
+	if _, err := pg.pgLister.PodGroups(pod.Namespace).Get(pgName); err != nil {
+		return
+	}
+
+	if err := pg.vcClient.SchedulingV1beta1().PodGroups(pod.Namespace).Delete(context.TODO(), pgName, metav1.DeleteOptions{}); err != nil {
```
hzxuzhonghu, line 64: does it mean delete the pg once a pod deleted?

lucming: yes

pkg/controllers/podgroup/pg_controller.go
lucming: Perhaps i can wirite the deleted pod in a queue,just like create pg

lucming: > 

Perhaps i can wirite the deleted pod in a queue,just like create pg,do you think that's ok

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -139,6 +142,8 @@ func (pg *pgcontroller) processNextReq() bool {
 	// If no error, forget it.
 	pg.queue.Forget(req)
 
+	pg.deletePG()
```
shinytang6, line 145: it's better to put this into another new goroutine

shinytang6 (changes requested):

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -139,6 +142,8 @@ func (pg *pgcontroller) processNextReq() bool {
 	// If no error, forget it.
 	pg.queue.Forget(req)
 
+	go pg.deletePG()
```
shinytang6, line 145: l think you misunderstood what l mean, l mean it's better to use another worker to delete pg, while not doing it in `processNextReq `

shinytang6 (changes requested):

pkg/controllers/podgroup/pg_controller.go
lucming: so,you hopes i use first version that delete pg in delete eventhandler,but use a gorutine to do it,is that so?

shinytang6: @lucming No, l mean you can increase a `deleteWorker` goroutine to do that.
Something like that:
```
go wait.Until(pg.deleteWorker, 0, stopCh)

func (pg *pgcontroller) deleteWorker() {
	for pg.processNextDeleteReq() {}
}

func (pg *pgcontroller) processNextDeleteReq() {
        ...
}
```

lucming: @shinytang6 done

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -94,6 +97,7 @@ func (pg *pgcontroller) Run(stopCh <-chan struct{}) {
 	cache.WaitForCacheSync(stopCh, pg.podSynced, pg.pgSynced)
 
 	go wait.Until(pg.worker, 0, stopCh)
+	go wait.Until(pg.deletePG, 0, stopCh)
```
shinytang6, line 100: you need a for-loop here, just like the `worker()` :)
```
func (pg *pgcontroller) worker() {
	for pg.processNextReq() {
	}
}
```

shinytang6 (changes requested):

pkg/controllers/podgroup/pg_controller.go
lucming: @shinytang6 done

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package podgroup
 
 import (
+	"context"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
shinytang6, line 21: pls add one blank line between external and sys package

shinytang6 (changes requested):

Thor-wl:
@lucming Pls rebase all the commits into one.

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -150,3 +157,32 @@ func contains(slice []string, element string) bool {
...
+	if shutdown {
+		klog.Errorf("Fail to pop item from queue")
+		return false
+	}
+
+	req := obj.(pgRequest)
+	defer pg.pgDeletedQueue.Done(req)
+
+	if _, err := pg.pgLister.PodGroups(req.pgNamespace).Get(req.pgName); err != nil {
+		return false
```
shinytang6, line 177: l think it should not return false here

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -150,3 +157,32 @@ func contains(slice []string, element string) bool {
...
+	}
+
+	if err := pg.vcClient.SchedulingV1beta1().PodGroups(req.pgNamespace).Delete(context.TODO(), req.pgName, metav1.DeleteOptions{}); err != nil {
+		klog.Errorf("Failed to delete volcano job,ns:%s,name:%s,err:%v", req.pgNamespace, req.pgName, err)
+		return false
```
shinytang6, line 182: ditto

shinytang6 (changes requested):

pkg/controllers/podgroup/pg_controller.go
lucming: @Thor-wl done

shinytang6 (commented):
Thx, generally lgtm now 👍 
/cc @Thor-wl @hwdef @william-wang pls take another look

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -150,3 +157,32 @@ func contains(slice []string, element string) bool {
...
+	obj, shutdown := pg.pgDeletedQueue.Get()
+	if shutdown {
+		klog.Errorf("Fail to pop item from queue")
+		return false
+	}
+
+	req := obj.(pgRequest)
+	defer pg.pgDeletedQueue.Done(req)
+
+	if _, err := pg.pgLister.PodGroups(req.pgNamespace).Get(req.pgName); err != nil {
```
hwdef, line 176: Did we ignore this error?

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -49,6 +55,44 @@ func (pg *pgcontroller) addPod(obj interface{}) {
 	pg.queue.Add(req)
 }
 
+func (pg *pgcontroller) deletePod(obj interface{}) {
```
k82cn, line 58: Are we going to delete `PodGroup` when one Pod deleted?

k82cn:
IMO, we should handle this case by GC, e.g. only delete PodGroup when its owner was deleted.

pkg/controllers/podgroup/pg_controller_handler.go
lucming: > IMO, we should handle this case by GC, e.g. only delete PodGroup when its owner was deleted.

I agree with you,the essence of this problem is wrong to set the owner of pod,but it's a little hard to find the outest controller by pod,there are too many upstream pods that can be created,and we cannot enumerate explicit outputs.

k82cn: > > IMO, we should handle this case by GC, e.g. only delete PodGroup when its owner was deleted.
> 
> I agree with you,the essence of this problem is wrong to set the owner of pod,but it's a little hard to find the outest controller by pod,there are too many upstream pods that can be created,and we cannot enumerate explicit outputs.

We can find controlled owner reference of Pod to do that :)

lucming: ok, let me change that

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -125,10 +126,14 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
...
 
+// newPGOwnerReferences create ownerreference for pg,the pod that create by other controller use pod as the ownerreference
 func newPGOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
 	if len(pod.OwnerReferences) != 0 {
 		for _, ownerReference := range pod.OwnerReferences {
-			if ownerReference.Controller != nil && *ownerReference.Controller {
+			if ownerReference.Controller != nil &&
+				*ownerReference.Controller &&
+				ownerReference.APIVersion == batchv1alpha1.SchemeGroupVersion.String() &&
+				ownerReference.Kind == "Job" {
```
k82cn, line 136: hm... how this PR fix the problem?

lucming: Only pods created by the controller of volcano use the owner of the pod as the owner of the pg,in other case,pg will use pod as the owner of pg.

Thor-wl:
Any update for this PR? Seems some problems in UT. @lucming

lucming:
> Any update for this PR? Seems some problems in UT. @lucming

i need modify this unit-test,in this case that pod created by rs,we should use pod as pg's owner,because the lifetime of pg may be different with rs,rs maybe craeted by deployment,use pod as pg's owner maybe the probably direct way,and what do you think?

Thor-wl:
> > Any update for this PR? Seems some problems in UT. @lucming
> 
> i need modify this unit-test,in this case that pod created by rs,we should use pod as pg's owner,because the lifetime of pg may be different with rs,rs maybe craeted by deployment,use pod as pg's owner maybe the probably direct way,and what do you thin

> > Any update for this PR? Seems some problems in UT. @lucming
> 
> i need modify this unit-test,in this case that pod created by rs,we should use pod as pg's owner,because the lifetime of pg may be different with rs,rs maybe craeted by deployment,use pod as pg's owner maybe the probably direct way,and what do you think?

Well, I don't think set the pg's owner reference to be pod is reasonable. In fact, we should create pg for deployment, statefulset, job, vcjob or some other resource types. If it's pod alone, I think it's ok to create pg for it.

lucming:
> > > Any update for this PR? Seems some problems in UT. @lucming
> > 
> > 
> > i need modify this unit-test,in this case that pod created by rs,we should use pod as pg's owner,because the lifetime of pg may be different with rs,rs maybe craeted by deployment,use pod as pg's owner maybe the probably direct way,and what do you thin
> 
> > > Any update for this PR? Seems some problems in UT. @lucming
> > 
> > 
> > i need modify this unit-test,in this case that pod created by rs,we should use pod as pg's owner,because the lifetime of pg may be different with rs,rs maybe craeted by deployment,use pod as pg's owner maybe the probably direct way,and what do you think?
> 
> Well, I don't think set the pg's owner reference to be pod is reasonable. In fact, we should create pg for deployment, statefulset, job, vcjob or some other resource types. If it's pod alone, I think it's ok to create pg for it.

we can get resource easy from deloyment->rs->pod,but it's hard to go from bottom to top,eg:pod->rs->deployment,maybe i can try

lucming:
`func getOutestOwner(dyClient dynamic.Interface, kubeClient kubernetes.Interface, object *unstructured.Unstructured, result *[]metav1.OwnerReference) {
	owners := object.GetOwnerReferences()
	if len(owners) > 0 {
		for _, owner := range owners {
			if owner.Controller != nil {
				apiversionList := strings.Split(owner.APIVersion, "/")
				var group, version string
				if len(apiversionList) == 2 {
					group, version = apiversionList[0], apiversionList[1]
				} else if len(apiversionList) == 1 {
					version = apiversionList[0]
				}
				gk := schema.GroupKind{
					Group: group,
					Kind:  owner.Kind,
				}

				groupResources, err := restmapper.GetAPIGroupResources(kubeClient.Discovery())
				if err != nil {
					klog.Errorf("failed to get resouces(%s:%s)in ns(%s),error=%v", gk.String(), owner.Name, object.GetNamespace())
					return
				}
				rm := restmapper.NewDiscoveryRESTMapper(groupResources)
				mapping, err := rm.RESTMapping(gk, version)
				parentObj, err := dyClient.Resource(mapping.Resource).
					Namespace(object.GetNamespace()).
					Get(context.Background(), owner.Name, metav1.GetOptions{})
				if err != nil {
					klog.Errorf("failed to get resouces(%s:%s)in ns(%s),error=%v", gk.String(), owner.Name, object.GetNamespace())
					return
				}
				getOutestOwner(dyClient, kubeClient, parentObj, result)
			}
		}
	} else {
		ref := metav1.NewControllerRef(object, object.GroupVersionKind())
		*result = append(*result, *ref)
	}
}`

lucming:
```
func getOutestOwner(dyClient dynamic.Interface, kubeClient kubernetes.Interface, object *unstructured.Unstructured, result *[]metav1.OwnerReference) {
	owners := object.GetOwnerReferences()
	if len(owners) > 0 {
		for _, owner := range owners {
			if owner.Controller != nil {
				apiversionList := strings.Split(owner.APIVersion, "/")
				var group, version string
				if len(apiversionList) == 2 {
					group, version = apiversionList[0], apiversionList[1]
				} else if len(apiversionList) == 1 {
					version = apiversionList[0]
				}
				gk := schema.GroupKind{
					Group: group,
					Kind:  owner.Kind,
				}

				groupResources, err := restmapper.GetAPIGroupResources(kubeClient.Discovery())
				if err != nil {
					klog.Errorf("failed to get resouces(%s:%s)in ns(%s),error=%v", gk.String(), owner.Name, object.GetNamespace())
					return
				}
				rm := restmapper.NewDiscoveryRESTMapper(groupResources)
				mapping, err := rm.RESTMapping(gk, version)
				parentObj, err := dyClient.Resource(mapping.Resource).
					Namespace(object.GetNamespace()).
					Get(context.Background(), owner.Name, metav1.GetOptions{})
				if err != nil {
					klog.Errorf("failed to get resouces(%s:%s)in ns(%s),error=%v", gk.String(), owner.Name, object.GetNamespace())
					return
				}
				getOutestOwner(dyClient, kubeClient, parentObj, result)
			}
		}
	} else {
		ref := metav1.NewControllerRef(object, object.GroupVersionKind())
		*result = append(*result, *ref)
	}
}
```
I've implemented a way to find the outermost controller recursively, so you can look at it and give me some feedback

PR #1895, opened 2021-12-08
enhance the security of TLS client authentication for webhook

huone1:

currently,  api-server  webhook Authentication use the default authentication policy and  it  is  insecure .
So it is necessary to set a list of cipher suites to enhance the security of TLS client authentication


william-wang (approved):
/lgtm

PR #1896, opened 2021-12-08
support label volcano.sh/task-priority

qiankunli:
a training job( such as tfjob/pytorchjob) always has many pods: one master and some workers. we need tell volcano that  the priority of  master pod is high than worker, so volcano can run master first and preempt master last.

we can set `pod.Priority` by create PriorityClass object in k8s, but it is troublesome for operator(such as tf-operator/pytorch-operator) to maintain the PriorityClass object. 

so it it simple to use pod Label `volcano.sh/task-priority` to mark the pod priority

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -162,6 +162,11 @@ func NewTaskInfo(pod *v1.Pod) *TaskInfo {
 		},
 	}
 
+	if taskPriority, ok := pod.Labels["volcano.sh/task-priority"]; ok {
+		if priority, err := strconv.ParseInt(taskPriority, 10, 32); err == nil {
+			ti.Priority = int32(priority)
+		}
+	}
 	if pod.Spec.Priority != nil {
 		ti.Priority = *pod.Spec.Priority
 	}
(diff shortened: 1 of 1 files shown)


hwdef:
I do not recommend using this method,for your scene in contrast it is better to use the dependsOn field.
I think there are two shortcomings in the way you proposed:
* Users need to check the document to know how to use it.(But you did not provide documentation)
* Even if there is this field, is it good to put it in the label? I think it might be better to put it in the comments.

pkg/scheduler/api/job_info.go
```diff
@@ -166,6 +166,12 @@ func NewTaskInfo(pod *v1.Pod) *TaskInfo {
 		ti.Priority = *pod.Spec.Priority
 	}
 
+	if taskPriority, ok := pod.Labels["volcano.sh/task-priority"]; ok {
```
k82cn, line 169: Can we have a const for `volcano.sh/task-priority`?

And it's better to get this info from annotation instead of labels; AFAIK, we did several check against annotations, e.g. reject updating task priority :)

qiankunli: ok, I have changed `volcano.sh/task-priority` from label to annotation and use `volcano.sh/task-priority` as a const

PR #1898, opened 2021-12-13
default sort task by index

xiaoanyunfei:
The tasks of vc-job is created in goroutine, and the `CreationTimestamp` of tasks is random.
 Sorting tasks by task index is a better choice

### DIFF
--- a/pkg/controllers/job/helpers/helpers.go
+++ b/pkg/controllers/job/helpers/helpers.go
@@ -19,6 +19,7 @@ package helpers
 import (
 	"fmt"
 	"math/rand"
+	"strconv"
 	"strings"
 	"time"
 
@@ -26,6 +27,7 @@ import (
 
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	"volcano.sh/volcano/pkg/controllers/apis"
+	"volcano.sh/volcano/pkg/scheduler/api"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/controllers/job/helpers/helpers.go
```diff
@@ -45,6 +46,24 @@ func GetTaskIndex(pod *v1.Pod) string {
 	return ""
 }
 
+// ComparePodByIndex by pod index
+func ComparePodByIndex(lPod, rPod *v1.Pod) int {
```
k82cn, line 50: Let's compare task directly :)

pkg/scheduler/framework/session_plugins.go
```diff
@@ -633,13 +634,14 @@ func (ssn *Session) TaskOrderFn(l, r interface{}) bool {
 		return res < 0
 	}
 
-	// If no task order funcs, order task by CreationTimestamp first, then by UID.
+	// If no task order funcs, order task by index first, then by CreationTimestamp.
 	lv := l.(*api.TaskInfo)
 	rv := r.(*api.TaskInfo)
-	if lv.Pod.CreationTimestamp.Equal(&rv.Pod.CreationTimestamp) {
-		return lv.UID < rv.UID
+	res := helpers.ComparePodByIndex(lv.Pod, rv.Pod)
```
k82cn, line 640: Prefer to `helpers.CompareTasksByIndex(lv, rv)`; and maybe we can just use `helpers.CompareTasks(lv, rv)` which also incloudes timestamp comparation.

pkg/controllers/job/helpers/helpers.go
```diff
@@ -45,6 +46,24 @@ func GetTaskIndex(pod *v1.Pod) string {
...
+	if lPod == nil || rPod == nil {
+		return 0
```
Thor-wl, line 52: If any pod is `nil` or can not be conveted to integer, then returns `0`, is that reasonable?

pkg/controllers/job/helpers/helpers.go
xiaoanyunfei: > Let's compare task directly :)
done

k82cn: thanks

PR #1906, opened 2021-12-14
optimize: reuse predicate error on same task group

justadogistaken:
### Background
I found that there is a logic in "PredicateNodes" could be improved. 
In current logic, every replica(TaskInfo wrapping pod) of same task will do the same predicate operations at every scheduling action. But those replicas has same resource requirements in common. 
So I guess there is no need to predicate on the node again for another replicas if the "predicate" failure happend for one of the replicas.

```
  tasks:
    - replicas: 6
      name: "default-nginx"
      template:
        metadata:
          name: web
```

### Solution
saving `taskPredicateErrorCache map[string]map[string]error` in PredicateHelper. The map is like `map[TaskGroupID]map[Node.Name]error`.
And `TaskGroupID` is "$JobID/$TaskSpecKey". Could this identify a task group?

ref: #1740 (Optimize the throughput of scheduler)

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -79,6 +79,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 		}
 	}
 
+	ph := util.NewPredicateHelper()
 	// Preemption between Jobs within Queue.
 	for _, queue := range queues {
 		for {
@@ -124,7 +125,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 					}
 					// Preempt other jobs within queue
 					return job.Queue == preemptorJob.Queue && preemptor.Job != task.Job
-				}); preempted {
+				}, ph); preempted {
 					assigned = true
 				}
 			}
@@ -172,7 +173,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 					}
 					// Preempt tasks within job.
 					return preemptor.Job == task.Job
-				})
+				}, ph)
 				stmt.Commit()
 
 				// If no preemption, next job.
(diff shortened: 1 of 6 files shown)


pkg/scheduler/util/predicate_helper.go
```diff
@@ -0,0 +1,108 @@
...
+		// Check the nodes starting from where is left off in the previous scheduling cycle,
+		// to make sure all nodes have the same chance of being examined across pods.
+		node := nodes[(lastProcessedNodeIndex+index)%allNodes]
+		atomic.AddInt32(&processedNodes, 1)
+		klog.V(4).Infof("Considering Task <%v/%v> on node <%v>: <%v> vs. <%v>",
+			task.Namespace, task.Name, node.Name, task.Resreq, node.Idle)
+
+		// Check if the task had "predicate" failure before.
+		// And then check if the task failed to predict on this node before.
+		if taskFailedBefore {
```
shinytang6, line 58: But node resource will be changed anytime, right?
so l think that we cannot assume current task will fail to predicate based on historical failure.

justadogistaken: Yes, node resource will change anytime. Thus I record the failure in every scheduling perioid, not the entire volcano lifetime. 
If the replica is from same task, its "predicate" failure could be reuse for other replicas. And no need to go predicating for  other replicas in this scheduling. I think this could help improve throughput.

shinytang6: yeah l mean in one scheduling period, node resource may change a lot.
e.g when you fail to predicate for task 1 on node, but some pods released on this node immediately, the next task2 should be scheduled to that node successfully.

shinytang6: ok, got you now.

k82cn:
LGTM overall :)

justadogistaken:
I have one question:
Could "$JobID/$TaskSpecKey"  identify a task group?

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -124,7 +125,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 					}
 					// Preempt other jobs within queue
 					return job.Queue == preemptorJob.Queue && preemptor.Job != task.Job
-				}); preempted {
+				}, ph); preempted {
```
huone1, line 128: Why is ph returned here？

shinytang6:
> I have one question: Could "$JobID/$TaskSpecKey" identify a task group?

Yes.

shinytang6 (commented):
/lgtm

PR #1910, opened 2021-12-20
update k8s to v1.23.0

Thor-wl:
ref: https://github.com/volcano-sh/volcano/issues/1665
Signed-off-by: Thor-wl <13164644535@163.com>

PR #1912, opened 2021-12-20
update golang to v1.17

Thor-wl:
As https://github.com/kubernetes-sigs/controller-tools/issues/643 mentions, there may be some bugs in package `encoding` for golang v1.16.x. This PR should be merged before https://github.com/volcano-sh/volcano/pull/1910

### DIFF
--- a/.github/workflows/code_verify.yaml
+++ b/.github/workflows/code_verify.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.x
+          go-version: 1.17.x
 
       - name: Checkout code
         uses: actions/checkout@v2
--- a/.github/workflows/e2e_parallel_jobs.yaml
+++ b/.github/workflows/e2e_parallel_jobs.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.x
+          go-version: 1.17.x
 
       - name: Install musl
         run: |
--- a/.github/workflows/e2e_scheduling_actions.yaml
+++ b/.github/workflows/e2e_scheduling_actions.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.x
+          go-version: 1.17.x
 
       - name: Install musl
         run: |
--- a/.github/workflows/e2e_scheduling_basic.yaml
+++ b/.github/workflows/e2e_scheduling_basic.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.x
+          go-version: 1.17.x
 
       - name: Install musl
         run: |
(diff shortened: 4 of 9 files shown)


Thor-wl:
Not sure which PR has introduces this bug or is there something wrong with this UT case. Ignore it temporarily. Will look into it and then give the fix.

hwdef:
I meet this error,too. 
Very confused

Thor-wl:
> 

I reviewed the UT file allocate_test.go and it was not changed recently. So I guess there may be some wrong update with the allocate.go. @justadogistaken I noticed that only this PR https://github.com/volcano-sh/volcano/pull/1906 was merged with this file. So can you help give a test about this bug? I've tested for serval times and found that the allocations are not stable. Namely, the output allocations are not unique. There are small probability to be sam e with expected.

justadogistaken:
> > 
> 
> I reviewed the UT file allocate_test.go and it was not changed recently. So I guess there may be some wrong update with the allocate.go. @justadogistaken I noticed that only this PR #1906 was merged with this file recently. So can you help give a test about this bug? I've tested for serval times and found that the allocations are not stable. Namely, the output allocations are not unique. There are small probability to be same with expected.

I reset my commit to the pr ahead of PR #1906, this error exists previously. But this may not be an error. Pods of c1/p1, c1/p2, c2/p1, c2/p2 have same requests. This error will happen if c1/p2 gets into tasks(priorityQueue) first or c2/p2 gets into first instead of p1.

justadogistaken:
Let's check this out.
TaskStatusIndex will return a taskMap, which is like map[TaskID]*TaskInfo.
And map does not have a concrete order while being accessed.
So it could happen to p2 is accessed before p1 in a pg.
But the result is correct.
The problem is that the "expected: "c2/p1": "n1", "c1/p1": "n1"" in UT is not appropriate.
It should be like "c1/ p1 or p2, c2/ p1 or p2".

```
			for _, task := range job.TaskStatusIndex[api.Pending] {
				// Skip BestEffort task in 'allocate' action.
				if task.Resreq.IsEmpty() {
					klog.V(4).Infof("Task <%v/%v> is BestEffort task, skip it.",
						task.Namespace, task.Name)
					continue
				}

				tasks.Push(task)
			}
```

justadogistaken:
There was a pr #1898  removing the follwing block.
Tasks will be sort by UID(***-p1, ***-p2) Originally.

```
	if lv.Pod.CreationTimestamp.Equal(&rv.Pod.CreationTimestamp) {
		return lv.UID < rv.UID
	}
	return lv.Pod.CreationTimestamp.Before(&rv.Pod.CreationTimestamp)
```
截屏2021-12-20 下午7 43 31

Thor-wl:
> There was a pr #1898 removing the follwing block. Tasks will be sort by UID(***-p1, ***-p2) Originally.
> 
> ```
> 	if lv.Pod.CreationTimestamp.Equal(&rv.Pod.CreationTimestamp) {
> 		return lv.UID < rv.UID
> 	}
> 	return lv.Pod.CreationTimestamp.Before(&rv.Pod.CreationTimestamp)
> ```
> 
> 截屏2021-12-20 下午7 43 31

/cc @xiaoanyunfei Pls help take a look and make corresponding UT cases changed if the allocate logic changes. THX.

shinytang6 (approved):
/lgtm
/approve

PR #1913, opened 2021-12-20
fix: allocate ut for "two Jobs on one node"

justadogistaken:
fix ut for "two Jobs on one node", which failed sometimes due to some changes of TaskOrderFn.
detail: #1912


Thor-wl:
Thanks.

PR #1915, opened 2021-12-21
non-vendor files changed when update k8s to v1.23.0

Thor-wl:
ref: https://github.com/volcano-sh/volcano/pull/1910
In order to make it clear for reviewers, separate the non-vendor files from the whole PR .

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -36,7 +36,7 @@ import (
 	infov1 "k8s.io/client-go/informers/core/v1"
 	schedv1 "k8s.io/client-go/informers/scheduling/v1"
 	storagev1 "k8s.io/client-go/informers/storage/v1"
-	storagev1alpha1 "k8s.io/client-go/informers/storage/v1alpha1"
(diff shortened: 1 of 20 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -36,7 +36,7 @@ import (
 	infov1 "k8s.io/client-go/informers/core/v1"
 	schedv1 "k8s.io/client-go/informers/scheduling/v1"
 	storagev1 "k8s.io/client-go/informers/storage/v1"
-	storagev1alpha1 "k8s.io/client-go/informers/storage/v1alpha1"
```
william-wang, line 39: the "storagev1alpha1" should be changed. Additionally, is the change compatible for v1alpha1 object?

Thor-wl: > the "storagev1alpha1" should be changed. Additionally, is the change compatible for v1alpha1 object?

Sorry for the clerical error. In fact, `storagev1apha1` has already been changed to `storagev1beta1`. The reference: https://github.com/volcano-sh/volcano/pull/1910/files

Issue #1777, opened 2021-10-12
Add dynamic scheduling and rescheduling functionality based on the usage of node


william-wang:
#### What would you like to be added:
1. Dynamic scheduling based on the usage of node other than the allocation rate of node.
2. Re-scheduling the pod on node based on the usage to another node to avoid that parts of node are are under or over utilized.

#### Why is this needed:
Currently the pod is scheduled based on the resource request and node allocatable resource other than the node usage. This would cause high allocation rate and low utilization rate in some conditions. As a result it's more likely for some nodes to get failure. 

SubTasks
- [x] Usage based scheduling
Support getting metrics from metrics server in cache and add uaged based scheduling plugin 
Design doc(pr #2023 )
Implemetation(pr #2129)  
- [x] Rescheduling plugin
Re-scheduling the pod on node based on the usage
Design doc(pr #1923 )
Implemetation(pr #2098 )
- [x] Shuffle action
Add shuffle action for evicting tasks for TDM and rescheduling scenarios
Design doc(pr #1923)
Implemetation shuffle action(pr #2099)
Implemetation rescheduling plugin (pr #2099)

justadogistaken:
I guess the real-time usage  of application is dynamic. Scheduling job based on real-time usage will improve system throughput no doubt. But I guess such a scheduling strategy (oversell in some way?) will also bring some risks of resource contention

Thor-wl:
> I guess the real-time usage of application is dynamic. Scheduling job based on real-time usage will improve system throughput no doubt. But I guess such a scheduling strategy (oversell in some way?) will also bring some risks of resource contention

Yes, I agree with that. IMO, it is suitable for scenarios such as mix deplpyment of online services and off-line jobs. That means it is reasonable to evict some off-line jobs when oversell works not so accurately.

k82cn:
There're two scenarios for William's proposal: 1. avoid 100% cpu usage, 2. oversubscription; for this issue, prefer to handle the first scenario.

justadogistaken:
> > I guess the real-time usage of application is dynamic. Scheduling job based on real-time usage will improve system throughput no doubt. But I guess such a scheduling strategy (oversell in some way?) will also bring some risks of resource contention
> 
> Yes, I agree with that. IMO, it is suitable for scenarios such as mix deplpyment of online services and off-line jobs. That means it is reasonable to evict some off-line jobs when oversell works not so accurately.

[...]

PR #1917, opened 2021-12-22
support descheduler for volcano

Thor-wl:
This is just fork from https://github.com/kubernetes-sigs/descheduler. Will give specified support for Volcano such as group scheduling consideration when evict pods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,10 +110,32 @@ job.batch/volcano-admission-init   1/1           48s        96s
 
 ```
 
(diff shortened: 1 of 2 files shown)


k82cn:
why install this by default? If only a fork, a document is better.

Thor-wl:
> why install this by default? If only a fork, a document is better.

It's not installed by default, just optional. I'll give a doc for this feature later.

Issue #1822, opened 2021-11-04
More friendly and adjustable chart!


swartz-k:
```
W1104 18:26:18.359321   49865 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W1104 18:26:18.415368   49865 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "prometheus" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "volcano"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "volcano-system"
```
#### What would you like to be added:
When I use `helm` install latest helm chart in `master` branch got errors. As using `1.22.x` kubernetes and existing prometheus I have to edit those chart file. Maybe making those components such as `grafana`、`promethues` be adjustable in `values.yaml` will be nice

#### Why is this needed:

I think a more friendly and adjustable chart will be nice for beginners.

swartz-k:
And same error in previous issue.
```
Error: INSTALLATION FAILED: CustomResourceDefinition.apiextensions.k8s.io "jobs.batch.volcano.sh" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[tasks].items.properties[template].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property, spec.validation.openAPIV3Schema.properties[spec].properties[tasks].items.properties[template].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property]
```

swartz-k:
> protocol

Change to `basic.crd_version` to `v1` will solve CRD error.

william-wang:
Thank you for reporting the issue.  We will fix the issue.

Thor-wl:
[...]

PR #1919, opened 2021-12-23
update CustomResourceDefinition version to v1

Thor-wl:
ref: https://github.com/volcano-sh/volcano/issues/1822
version `CustomResourceDefinition` will be updated to `v1` instead of `v1beta1` after k8s v1.22

### DIFF
--- a/installer/helm/chart/volcano/Chart.yaml
+++ b/installer/helm/chart/volcano/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for Volcano
 name: volcano
-version: 0.1
+version: 1.4
 appVersion: 0.1
 icon: https://raw.githubusercontent.com/volcano-sh/charts/master/docs/images/volcano-logo.png
 home: https://volcano.sh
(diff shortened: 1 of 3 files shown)


installer/helm/chart/volcano/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for Volcano
 name: volcano
-version: 0.1
```
william-wang, line 4: please change it to 1.5.

Thor-wl: OK. Changed.

k82cn:
Does that mean we will not support old k8s cluster?

Thor-wl:
> Does that mean we will not support old k8s cluster?

Yes, the latest k8s version has already been v1.24. We cannot support so many old versions.

william-wang (approved):
/lgtm

k82cn:
> > Does that mean we will not support old k8s cluster?
> 
> Yes, the latest k8s version has already been v1.24. We cannot support so many old versions.

We need to align with k8s release; for example, we should support at least 1.22, 1.23 & 1.24 :)

k82cn:
please also update related document .

PR #1920, opened 2021-12-24
feat(webhook): add task dependson webhook

shinytang6:

prior pr: https://github.com/volcano-sh/volcano/pull/1833

### DIFF
--- a/pkg/webhooks/admission/jobs/validate/util.go
+++ b/pkg/webhooks/admission/jobs/validate/util.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/go-multierror"
-
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/kubernetes/pkg/apis/core/validation"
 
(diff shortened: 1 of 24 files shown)


hwdef (commented):
Do you need to check if the task name in task.depedensOn is in the job?

By the way, Very well written about dag. tql.

shinytang6:
> Do you need to check if the task name in task.depedensOn is in the job?
> 
> By the way, Very well written about dag. tql.

I think it is not necessary. If there is an illegal taskName defined in dependsOn, it will not form a dag and rejected by webhook.

hwdef:
/lgtm
Thanks!

pkg/webhooks/admission/jobs/validate/util.go
```diff
@@ -189,3 +188,62 @@ func validateIO(volumes []batchv1alpha1.VolumeSpec) error {
 	}
 	return nil
 }
+
+// topoSort uses topo sort to sort job tasks based on dependsOn field
+// it will return an array contains all sorted task names and a bool which indicates whether it's a valid dag
+func topoSort(job *batchv1alpha1.Job) ([]string, bool) {
```
Thor-wl, line 194: Nice work for `topoSort` and `makeGraph` functions!

PR #1930, opened 2021-12-31
add UT coverage report for v1.5.0

Thor-wl:


william-wang:
We do not have to maintain such a big file in source code. It is about 1M. We could provide a readme to describe how to generate the report if someone care about it.

Thor-wl:
> We do not have to maintain such a big file in source code. It is about 1M. We could provide a readme to describe how to generate the report if someone care about it.

I think it's ok though the file is 1M. It's also regular to provide a UT report when launching a new release.

shinytang6 (approved):
/lgtm

PR #1949, opened 2022-01-11
Turn down level of enqueue logs to reduce log print in each session.

jiangkaihua:

In volcano default configuration, level-3 logs would be printed by default. So some unnecessary logs in `enqueue` action should be turned down to level-5.

https://github.com/volcano-sh/volcano/blob/14da3e942deb397877c39dd2d7e0870534409b90/pkg/scheduler/actions/enqueue/enqueue.go#L62 
This log would be printed for every queue in every session.

---

https://github.com/volcano-sh/volcano/blob/14da3e942deb397877c39dd2d7e0870534409b90/pkg/scheduler/actions/enqueue/enqueue.go#L73
This log would be printed for every pending job in every session.

### DIFF
--- a/pkg/scheduler/actions/enqueue/enqueue.go
+++ b/pkg/scheduler/actions/enqueue/enqueue.go
@@ -59,7 +59,7 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 				job.Queue, job.Namespace, job.Name)
 			continue
 		} else if _, existed := queueMap[queue.UID]; !existed {
-			klog.V(3).Infof("Added Queue <%s> for Job <%s/%s>",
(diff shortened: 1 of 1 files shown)


hwdef (commented):
/lgtm

k82cn:
hm... how do we define log level, e.g. 4 vs. 5?

hwdef:
> hm... how do we define log level, e.g. 4 vs. 5?

No related documentation found.
If it is an infinite loop, level 5 should be used. Is level 4 or 5 reasonable here?

just by feeling ^.^

Thor-wl:
I agree with that we can set standard about log level. It all depends on the contributor's recognization about the importance of the log message.
/approve

Issue #1946, opened 2022-01-06
When device-plugin redeploy， resources that have been used are not counted.


zishen:
**What happened**:
When resources are used by the vcJob allocation, redeploy deviceplugin. Found  the using resource has not been calculated.
![1](https://user-images.githubusercontent.com/7522079/148354595-ada72573-dd7f-42e0-b3fe-ad707f504656.PNG)

When reshipment deviceplugin volcano restart itself, an error is as follows:

**What you expected to happen**:
 When redeploy resource device-plugin,  the count of resource is right.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:v1.4.0
- Kubernetes version (use `kubectl version`):1.17.3
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):ubuntu 
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

william-wang:
Would you like to provide the device plugin information？ Which device plugin and version are you using?

zishen:
> Would you like to provide the device plugin information？ Which device plugin and version are you using?

Device plugin  is what we aimed at NPU custom, similar to the GPU。

zishen:
This issue has be reply in v1.5.0-beta https://github.com/volcano-sh/volcano/pull/1685/files.

PR #1952, opened 2022-01-12
fix possible panic when 'SetNode' is called

william-wang:
fix issue for #1946


k82cn:
let's use check-pick script for this PR :)

Issue #1901, opened 2021-12-13
SLA plugin doesn't work on `batch/v1` `Job` objects; `sla-waiting-time` from `volcano-scheduler.conf` is ignored


adamnovak:
**What happened**:

As mentioned in #1869 I am using Volcano to schedule Kubernetes Job objects, to try and prevent smaller jobs submitted later from immediately filling any available space and starving larger jobs submitted earlier.

My cluster has a 96-core node with hostname "k1.kube".

I installed Volcano from the Helm chart in tag v1.4.0, using this `values.yaml`:

```
basic:
  image_tag_version: "v1.4.0"
  controller_image_name: "volcanosh/vc-controller-manager"
  scheduler_image_name: "volcanosh/vc-scheduler"
  admission_image_name: "volcanosh/vc-webhook-manager"
  admission_secret_name: "volcano-admission-secret"
  admission_config_file: "config/volcano-admission.conf"
  scheduler_config_file: "config/volcano-scheduler.conf"

  image_pull_secret: ""
  admission_port: 8443
  crd_version: "v1"
custom:
  metrics_enable: "false"
```

And then overriding the scheduler configmap with this and restarting the scheduler pod:

```
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
      - name: conformance
      - name: sla
        arguments:
          # Stop letting little jobs pass big jobs after the big jobs have been
          # waiting this long
          sla-waiting-time: 5m
    - plugins:
      - name: overcommit
      - name: drf
      - name: predicates
      - name: proportion
      - name: nodeorder
        arguments:
          # Maybe this will try to fill already full nodes first?
          leastrequested.weight: 0
          mostrequested.weight: 2
          nodeaffinity.weight: 3
          podaffinity.weight: 3
          balancedresource.weight: 1
          tainttoleration.weight: 1
          imagelocality.weight: 1
      - name: binpack
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: volcano
    meta.helm.sh/release-namespace: volcano-system
  labels:
    app.kubernetes.io/managed-by: Helm
  name: volcano-scheduler-configmap
  namespace: volcano-system
```

So I should be using a global SLA of 5 minutes.

Then, I prepared a test: fill up the node with some jobs, then queue a big job, then queue a bunch of smaller jobs after it:

```
# Clean up
[...]

PR #1961, opened 2022-01-14
Inherit annotations from upper resources to k8s podgroup.

jiangkaihua:
When users applied k8s origin workloads like `daemonset`, `job`, etc., k8s created pods first, then volcano created podgroup from the pods, whose progress was different form volcano jobs. So podgroup created from k8s pods would miss annotations of origin workloads, causing configurations inserted in the form of annotations neglected, like #1901 .

So this pr fetched annotations from upper resources by searching pod `ownerReferences`, and filled in podgroups, so that some configurations in annotations were available for k8s workloads.

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -365,11 +365,19 @@ func (ji *JobInfo) SetPodGroup(pg *PodGroup) {
 	ji.CreationTimestamp = pg.GetCreationTimestamp()
 
 	var err error
-	ji.WaitingTime, err = ji.extractWaitingTime(pg)
+	ji.WaitingTime, err = ji.extractWaitingTime(pg, v1beta1.JobWaitingTime)
 	if err != nil {
 		klog.Warningf("Error occurs in parsing waiting time for job <%s/%s>, err: %s.",
 			pg.Namespace, pg.Name, err.Error())
 		ji.WaitingTime = nil
+	}
+	if ji.WaitingTime == nil {
+		ji.WaitingTime, err = ji.extractWaitingTime(pg, JobWaitingTime)
+		if err != nil {
+			klog.Warningf("Error occurs in parsing waiting time for job <%s/%s>, err: %s.",
+				pg.Namespace, pg.Name, err.Error())
+			ji.WaitingTime = nil
+		}
 	}
 
 	ji.Preemptable = ji.extractPreemptable(pg)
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	k8s.io/kubernetes v1.23.0
 	sigs.k8s.io/yaml v1.2.0
 	stathat.com/c/consistent v1.0.0
-	volcano.sh/apis v0.0.0-20211109104455-7e8d893a9f66
+	volcano.sh/apis v1.5.0-beta.0.0.20220112022954-fd6c584c7994
```
hwdef, line 30: I'm not sure if this version number is spelled correctly, it seems to be different than in another pr

https://github.com/volcano-sh/volcano/pull/1964/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R30

jiangkaihua: Yes, I've updated `apis` vendor to latest master branch.

pkg/scheduler/api/job_info.go
```diff
@@ -365,12 +365,20 @@ func (ji *JobInfo) SetPodGroup(pg *PodGroup) {
...
 	var err error
-	ji.WaitingTime, err = ji.extractWaitingTime(pg)
+	ji.WaitingTime, err = ji.extractWaitingTime(pg, v1beta1.JobWaitingTime)
 	if err != nil {
 		klog.Warningf("Error occurs in parsing waiting time for job <%s/%s>, err: %s.",
 			pg.Namespace, pg.Name, err.Error())
 		ji.WaitingTime = nil
 	}
+	if ji.WaitingTime == nil {
+		ji.WaitingTime, err = ji.extractWaitingTime(pg, JobWaitingTime)
```
shinytang6, line 375: why we have different keys `JobWaitingTime` and `v1beta1.JobWaitingTime` here?

jiangkaihua: The old key `JobWaitingTime` is only available to volcano jobs, whereas new key `v1beta1.JobWaitingTime` can be used in both volcano job and k8s workloads. So new key can substitute the old one basically.

The old key was reserved temporarily to ensure that previous users still could use SLA plugin as usual. After all annotation-keys were available, a doc would be proposed to introduce these keys, and then old keys could be removed.

Thor-wl:
Pls rebase from the master branch and solve the conflicts.
@jiangkaihua

william-wang (commented):
/lgtm

william-wang (approved):
/lgtm
/approve

Issue #1935, opened 2022-01-04
Provide extender framework for volcano


xing0821:
#### What would you like to be added:

  can we add an extender framework such as kube-scheduler's extender. maybe we can send a RPC request or http request on each plugin functions, such as predicateFns.

#### Why is this needed:
  
  we can add some logic of scheduling out of source code. such as an Operator specific requirements of scheduling

k82cn:
+1

xing0821:
l am interested in that feature, maybe i can give a design doc first about extender

shinytang6:
> can we add an extender framework such as kube-scheduler's extender. maybe we can send a RPC request or http request on each plugin functions, such as predicateFns.

That's great, looking forward to that 😃

k82cn:
Yes, we can have an http/grpc plugin to deligate the request; so I think that's simple enough for beginner :)

hejufang:
looking forward to that

xing0821:
PR is https://github.com/volcano-sh/volcano/pull/1951

PR #1963, opened 2022-01-16
Add design doc for volcano extender plugin

xing0821:
Extender plugin design doc in (https://github.com/volcano-sh/volcano/issues/1935)

### DIFF
--- a/docs/design/extender.md
+++ b/docs/design/extender.md
@@ -1,0 +1,64 @@
+# Volcano Framework Extender
+## Summary
+  At present, volcano has not provided a mechanism for non-invasive function realization. If the user needs to implement an internal scheduling function, the source code of volcano must be modified.
+  Such as kube-scheduler, this proposal is to dynamically extend Volcano's scheduling capabilities via an http request.
+## Motivation
+  Currently, the only way to implement a scheduling strategy based on Volcano is to modify Volcano's code and recompile.
+  This method requires developers to deeply understand the working principle of Volcano and the related architecture, which is not friendly to beginners. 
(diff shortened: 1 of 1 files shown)


shinytang6 (commented):
lgtm overall 👍

docs/design/extender.md
```diff
@@ -0,0 +1,64 @@
...
+  Implement a plugin named extender, the plugin will register related functions in the SessionOpen function, and pass the snapshot to the endpoint through the http method.
+  The extender plugin will choose which methods to register according to the configuration file, and is responsible for serializing and deserializing the parameters. When there is an error in the http call, the plugin decides how to deal with it according to the configuration file.
+### Extender Arguments
+```
+- plugins:
+   - name: extender
+     arguments:
+       extender.urlPrefix: http://127.0.0.1
+       extender.httpTimeout: 100ms
+       extender.openSessionOpenVerb: onSessionOpen
```
k82cn, line 25: prefer to have default value for those verbs to make configuration simple :)

k82cn: and have a parameter `extender.defaultVerbs = false` to disable all default verbs if necessary.

k82cn: anyway, we can improve that later; as I'm not sure which one is simpler :)

xing0821: If plugin provide default value of verbs, how to distinguish between default value and null value(null value means that the function has not been defined by the user) 
Maybe we can add an example configuration in example package

k82cn: that's ok :)

xing0821: > and have a parameter `extender.defaultVerbs = false` to disable all default verbs if necessary.

that's a good idea. We can do some user interviews to make configuration more simpler

Issue #1947, opened 2022-01-10
Large memory is used by volcano-scheduler


jazka:
**What happened**:
The volcano-scheduler uses large memory and the function of schedule seems to be unusual. The pod of job will always be pending. But there is no any error logs in pod. After restart it, everything is okay. 

The image link of the result of  "kubectl top pods -n volcano-system":
https://l4x826wg3c.feishu.cn/file/boxcnLnrbgq6CmvjAQWfOlWX9gc

**What you expected to happen**:
The schduler works well. Or know how to check what happened, then I could restart it when monitor some events.

**How to reproduce it (as minimally and precisely as possible)**:
Not sure how to reproduce it, but it happend many times after running some days.

**Anything else we need to know?**:
1. The ETCD in our cluster uses normal Disk, not SSD. 
2. The Node volcano pods are deployed could be scheduled by other computer task like pytorchjob, and there is no Requests to volcano.

**Environment**:
- Volcano Version: https://github.com/volcano-sh/volcano/commit/1b96bdf4de821e1e4af2b4c056f67be7559a880d
- Kubernetes version (use `kubectl version`): 1.22.2
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hwdef:
Please provide more information such as scheduler logs

jazka:
> Please provide more information such as scheduler logs

The images below are the some logs screenshot (I can not parse all the logs here, sorry for that), but there is no other different contents with these logs.

https://l4x826wg3c.feishu.cn/file/boxcnFFuid08JWiylxu0NAw5CZc
https://l4x826wg3c.feishu.cn/file/boxcnLnyvqitaA6bAshUriNUEMc
https://l4x826wg3c.feishu.cn/file/boxcnQmvPxfjGiNv6BglGmDH7Za
https://l4x826wg3c.feishu.cn/file/boxcnYDoTEobLvJZ0bPHx4vMpFb

k82cn:
did you check whether the memory of scheduler pod keep increasing?

jazka:
I am not very sure if it keeps increasing. The large memory case was found after scheduler has been running for a few days, I guess the scheduler has some internal errors in these days. My guess is from the logs of the scheduler which has no more  scheduling logs to output.
[...]

Issue #1863, opened 2021-11-29
Pytorchjob pods hang on pending status with volcano scheduler


jazka:
**What happened**:
The Pytorchjob pods hang on pending status with volcano scheduler. And there is no logs about this job in volcano-scheduler pod. 
The podgroup is created.

**What you expected to happen**:
The Pytorchjob pods should be scheduled normally

**How to reproduce it (as minimally and precisely as possible)**:
It's an accidental phenomenon

**Anything else we need to know?**:
The pytorchjob infos, refer to https://github.com/kubeflow/training-operator/issues/1492#issuecomment-981527879

**Environment**:
- Volcano Version: 1.4
- Kubernetes version (use `kubectl version`): 1.22
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shinytang6:
Could you describe the podgroup info? @jazka

jazka:
Sorry for late reply. 

The podgroup info as below:

https://l4x826wg3c.feishu.cn/file/boxcnEpNDeZfgfeKU9LX9yY6LVe

https://l4x826wg3c.feishu.cn/file/boxcnnvDjWG3NDwX0tMabcPVXvd

shinytang6:
> Sorry for late reply.
> 
> The podgroup info as below:
> 
> https://l4x826wg3c.feishu.cn/file/boxcnEpNDeZfgfeKU9LX9yY6LVe
> 
> https://l4x826wg3c.feishu.cn/file/boxcnnvDjWG3NDwX0tMabcPVXvd

it's strange that the status of PodGroup is empty.. The pods will start to be scheduled once podgroup become inqueue or ready. Any related scheduler log? @jazka

jazka:
> > Sorry for late reply.
> > The podgroup info as below:
> > https://l4x826wg3c.feishu.cn/file/boxcnEpNDeZfgfeKU9LX9yY6LVe
> > https://l4x826wg3c.feishu.cn/file/boxcnnvDjWG3NDwX0tMabcPVXvd
> 
> it's strange that the status of PodGroup is empty.. The pods will start to be scheduled once podgroup become inqueue or ready. Any related scheduler log? @jazka

Yes, it is weired that there is no any logs abount this job on scheduler pod. But the pending pods will became running if I restart the scheduler pod.

Thor-wl:
[...]

PR #1971, opened 2022-01-18
fix the scheduler stuck after delete resourcequota for namespace

william-wang:
#1863 and #1947
The reproduce step and root cause is on #1947.
Signed-off-by: william-wang <wang.platform@hotmail.com>


wpeng102 (approved):
/lgtm

Issue #1962, opened 2022-01-14
Unable to schedule independent pod when RBAC enforced


Robert-Christensen-visa:
**What happened**:

I am trying to use volcano to schedule independent pods and to add those pods to a specific queue.  This is done so volcano can keep track of resources used by the cluster.

I submit this pod:

``` yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: testns
spec:
  schedulerName: volcano
  containers:
  - name: main
    image: ubuntu:latest
    command: ["sh", "-c", "sleep 600000"]
    resources:
      limits:
        cpu: "1000m"
        memory: "256Mi"
```

The pod is submitted, but remains in the "pending" state.  When I look though the logs for the controller, it is reports a podgroup can't be created for the pod.  I also don't see a podgroup created for the pod.  I see something like this in the logs:

```
E0113 20:57:02.278172       1 pg_controller.go:135] Failed to handle Pod <testns/test-pod>: podgroups.scheduling.volcano.sh "podgroup-044f57a5-0484-451c-a1c1-8d9078bc9ef7" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
E0113 20:59:37.613273       1 pg_controller_handler.go:119] Failed to create normal PodGroup for Pod <testns/test-pod>: podgroups.scheduling.volcano.sh "podgroup-0775e3d2-940a-47b8-afc5-92023a135217" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
```

When an independent pod requests to use the volcano scheduler, it will try to create a podgroup for that pod.  When the podgroup is created it sets `blockOwnerDeletion=true` in the `ownerReference` field.  This requries updating the owner's finalizer list.  However, the current RBACs for Volcano do not include the ability to modify pod/finalizer.  Because of this, the podgroup can't be created.

**What you expected to happen**:

I expect if an independent pod is submitted to a cluster with `schedulerName: volcano` the pod will be scheduled.  This works on clusters where RBACs are not being enforced, such as on the local kubernetes cluster installed in docker desktop.

**How to reproduce it (as minimally and precisely as possible)**:

Submit the pod described above to a cluster that enforces RBACs.  The issue will not present itself if submitted to a local kubernetes cluster installed through docker.

**Anything else we need to know?**:

[...]

PR #1976, opened 2022-01-19
add RBACs for pod finalizers for controller

robertchristensen:
This should resolve https://github.com/volcano-sh/volcano/issues/1962.

Add "pod/finalizers" RBACs so the controller will be able to add finalizers to the pod.  This is needed because when an arbitrary pod requests to use volcano for scheduling, the vc-controller will create a podgroup.  The owner of the podgroup is the pod that triggered its creation.  In order to set `blockOwnerDeletion=true`, a finalizer will be set on the owner.

To limit the permissions required for volcano to be used to schedule arbitrary pods, some code that will try to set the owner of the podgroup to be the owner of the pod is removed.  Without removing this section, the vc-controller would require permissions to update the finalizer for other CRDs.  It would be unreasonable to have to add RBACs for finalizers for every CRD that creates a pod and wants the pod scheduled using volcano.


hwdef (commented):
please fix the DCO

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -126,14 +126,6 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
 }
 
 func newPGOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
-	if len(pod.OwnerReferences) != 0 {
```
k82cn, line 129: why remove this?

Thor-wl: > why remove this?

+1

robertchristensen: The function `newPGOwnerReferences` returns what to set as the owner of the podgroup that is being created.  Without the modifications above, what it says is if the pod is independent (it is created using something like `kubectl apply -f pod.yaml`) the owner of the podgroup is set to the pod.  Because `blockOwnerDeletion=true` it will need to modify that pod's finalizers.  These are the permissions being added in this PR.  If the pod is owned by a controller, such as if a replicaset is creating the pods, this block of code will try to set the owner of the podgroup being created to be the owner of the pod.  If a replicaset is creating a pod, this code will try to set the owner of the podgroup to be the replicaset that created that pod.  Because `blockOwnerDeletion=true`, setting the owner of the podgroup to the replicaset will require a modification to the replicaset finalizers.

If we remove that block of code volcano controller will only require RBACs for pod/finalizers when it is scheduling independent pods because the owners of the podgroups created though this code path will only ever be pods.  It will not require permissions for finalizers for other resources.

In our cluster we want to use volcano to schedule the pods for several workloads.  Currently we are using volcano to schedule vcjobs, but we want to expand it to schedule pods for other workflows.  As we add additional workloads with additional CRDs it would be difficult to continue to add additional permissions for the finalizers for each new CRD.

I am not sure if this explanation helped.  As I have tested using volcano to schedule arbitrary pods in a cluster that enforces RBACs, removing this portion of code made it so pods would get scheduled.  Otherwise, many pods would never be scheduled because the podgroup was never created.

k82cn: > I am not sure if this explanation helped. As I have tested using volcano to schedule arbitrary pods in a cluster that enforces RBACs, removing this portion of code made it so pods would get scheduled. Otherwise, many pods would never be scheduled because the podgroup was never created.

I'd suggest to only commit RBAC part in this pr, and share more logs for this case.  We'd like to support all kind of workloads in one cluster, so it's important to clarify the root cause :)

robertchristensen: Sounds good.  I looked into this a little more.  I think the reason why I am seeing this issue and other have not might be because of changes to out Kubernetes is configured.  In the cluster where I am seeing this issue, we enable [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement).  According to the docs, `/update` for the finalizer will be required to set `blockOwnerDeletion`.  However, I can take a little more time to explore options for resolving this issue on our cluster.

k82cn: please feel free to share the log and discuss it in the community :)

Issue #1980, opened 2022-01-22
Some Actions use alloc as its object name while it's not


elinx:
Several actions use `alloc` as their object name when it is actually not an `Allocate` action as in the following examples, which is a bit confusing to read.
- backfill action [L33](https://github.com/volcano-sh/volcano/blob/d706e416763ee2df84efd35ac8e793837d42a382/pkg/scheduler/actions/backfill/backfill.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L33)  `func (alloc *Action) Name() string {` which should be `func (backfill *Action) Name() string {`
- elect
- preempt
- reserve

PR #1981, opened 2022-01-22
fix misuse object name issue

elinx:
#1980

### DIFF
--- a/pkg/scheduler/actions/backfill/backfill.go
+++ b/pkg/scheduler/actions/backfill/backfill.go
@@ -30,13 +30,13 @@ func New() *Action {
 	return &Action{}
 }
 
-func (alloc *Action) Name() string {
+func (backfill *Action) Name() string {
 	return "backfill"
 }
 
-func (alloc *Action) Initialize() {}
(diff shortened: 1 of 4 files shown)


justadogistaken:
I guess the original naming is fine.

elinx:
> I guess the original naming is fine.

I would say yes if you mean the functionality, it works fine and they are all even not used in their methods. But one could reasonably assume that they were all copied and pasted from the `Allocate` action which I would like to change to their real name for readability and scalability(may be used in the future).

shinytang6 (commented):
pls fix the DCO

/lgtm

justadogistaken:
> > I guess the original naming is fine.
> 
> I would say yes if you mean the functionality, it works fine and they are all even not used in their methods. But one could reasonably assume that they were all copied and pasted from the `Allocate` action which I would like to change to their real name for readability and scalability(may be used in the future).

Got it

elinx:
> pls fix the DCO /lgtm

fixed, thanks

shinytang6 (approved):
/approve

PR #1986, opened 2022-01-24
Fix windows build

k82cn:
Signed-off-by: Klaus Ma <mada3@huawei.com>


Thor-wl:
Pls rebase the latest code of master branch. The ginkgo client has updated to v2 and it introduces some problems for CI

PR #1990, opened 2022-01-25
update config args to map[string]interface{} to extend the configs

Thor-wl:
This update will help extend the configurations for the furture. Besides, it will be used in the following PR for the rescheduling feature. The configuration example can be referred to https://github.com/volcano-sh/volcano/blob/3a7007f0e4bfb5b82c349409ecafc49bd09bda9e/docs/design/rescheduling.md

### DIFF
--- a/pkg/scheduler/conf/scheduler_conf.go
+++ b/pkg/scheduler/conf/scheduler_conf.go
@@ -36,7 +36,7 @@ type Configuration struct {
 	// Name is name of action
 	Name string `yaml:"name"`
 	// Arguments defines the different arguments that can be given to specified action
-	Arguments map[string]string `yaml:"arguments"`
+	Arguments map[string]interface{} `yaml:"arguments"`
 }
 
 // PluginOption defines the options of plugin
@@ -80,5 +80,5 @@ type PluginOption struct {
 	// EnabledJobStarving defines whether jobStarvingFn is enabled
 	EnabledJobStarving *bool `yaml:"enableJobStarving"`
 	// Arguments defines the different arguments that can be given to different plugins
-	Arguments map[string]string `yaml:"arguments"`
+	Arguments map[string]interface{} `yaml:"arguments"`
 }
--- a/pkg/scheduler/framework/arguments.go
+++ b/pkg/scheduler/framework/arguments.go
@@ -25,7 +25,7 @@ import (
 )
 
 // Arguments map
-type Arguments map[string]string
+type Arguments map[string]interface{}
 
 // GetInt get the integer value from string
 func (a Arguments) GetInt(ptr *int, key string) {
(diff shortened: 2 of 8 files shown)


pkg/scheduler/framework/arguments.go
```diff
@@ -34,13 +33,14 @@ func (a Arguments) GetInt(ptr *int, key string) {
 	}
 
 	argv, ok := a[key]
-	if !ok || argv == "" {
+	if !ok {
 		return
 	}
 
-	value, err := strconv.Atoi(argv)
+	valueStr, ok := argv.(string)
```
hwdef, line 40: Is this ok used?

pkg/scheduler/framework/arguments.go
```diff
@@ -54,13 +54,14 @@ func (a Arguments) GetFloat64(ptr *float64, key string) {
 	}
 
 	argv, ok := a[key]
-	if !ok || len(argv) == 0 {
+	if !ok {
 		return
 	}
 
-	value, err := strconv.ParseFloat(argv, 64)
+	valueStr, ok := argv.(string)
```
hwdef, line 61: here too

pkg/scheduler/framework/arguments.go
```diff
@@ -74,13 +75,14 @@ func (a Arguments) GetBool(ptr *bool, key string) {
 	}
 
 	argv, ok := a[key]
-	if !ok || argv == "" {
+	if !ok {
 		return
 	}
 
-	value, err := strconv.ParseBool(argv)
+	valueStr, ok := argv.(string)
```
hwdef, line 82: here too

pkg/scheduler/framework/arguments.go
Thor-wl: No, let me remove unused variables.

pkg/scheduler/conf/scheduler_conf.go
```diff
@@ -36,7 +36,7 @@ type Configuration struct {
 	// Name is name of action
 	Name string `yaml:"name"`
 	// Arguments defines the different arguments that can be given to specified action
-	Arguments map[string]string `yaml:"arguments"`
+	Arguments map[string]interface{} `yaml:"arguments"`
```
william-wang, line 39: We have to consider how to keep compatible with user developed plugins.

Thor-wl: > We have to consider how to keep compatible with user developed plugins.

This update will not influence users who make use of the built-in plugins. A little work to match the interface is needed for users who implement custom plugins with arguments. I've no idea how to support that gracefully. Hope to see some advices.

hwdef: Wait until go1.18, we can try to use go's generics

Thor-wl: > Wait until go1.18, we can try to use go's generics

@william-wang means the update from `string` to `interface` will introduces some code modifications in custom plugins, for example, the value should be converted to `string` with assert after getting the value from `Arguments`. I agree with that finding a way with no code modification for users is the best. If we cannot achieve that, the update is also acceptable for two factors: 
1. Users who make use of custom plugins are few as to the feedback from issues. 
2. The modification is of little work.

william-wang (approved):
/lgtm

PR #1992, opened 2022-01-25
add comments for custom plugins for CI code verification

Thor-wl:


kevin-wangzefeng:
It's not recommanded to leave code in comment rather than deleting it.
Can we skip checking this file?

Thor-wl:
> It's not recommanded to leave code in comment rather than deleting it. Can we skip checking this file?

Yes, I find the configuration. Thanks for advices. Let me close this PR and give another one.

PR #1995, opened 2022-01-27
fix enqueue

lucming:
the second cycle have already find the resource should break directly :)

### DIFF
--- a/pkg/scheduler/framework/session_plugins.go
+++ b/pkg/scheduler/framework/session_plugins.go
@@ -403,6 +403,7 @@ func (ssn *Session) JobEnqueueable(obj interface{}) bool {
 			}
 			if res > 0 {
 				hasFound = true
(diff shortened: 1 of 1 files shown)


pkg/scheduler/framework/session_plugins.go
```diff
@@ -403,6 +403,7 @@ func (ssn *Session) JobEnqueueable(obj interface{}) bool {
 			}
 			if res > 0 {
 				hasFound = true
+				break
```
william-wang, line 406: Can not break here. Need loop all fn in each plugin to make sure there is no reject (res <0).
Do you hit some issue in your env?

lucming: i didn't hit any issue in my env,i am ignoring this,i will close this pr,thanks :)

PR #2007, opened 2022-02-10
support EAS service pod request

elinx:
Add `EasAcceleratorFlavorId` field for `TaskInfo` to indicate whether EAS accelerator resource is used, also a special annotation tag `volcano.sh/eas-accelerator-flavor-id` is added accordingly for the pod request.

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -127,9 +127,10 @@ type TaskInfo struct {
 	// * value means workload can use all the revocable node for during node active revocable time.
 	RevocableZone string
 
-	NumaInfo   *TopologyInfo
-	PodVolumes *volumescheduling.PodVolumes
-	Pod        *v1.Pod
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -160,22 +161,24 @@ func NewTaskInfo(pod *v1.Pod) *TaskInfo {
...
+		Namespace:              pod.Namespace,
+		Priority:               1,
+		Pod:                    pod,
+		Resreq:                 resReq,
+		InitResreq:             initResReq,
+		Preemptable:            preemptable,
+		BestEffort:             bestEffort,
+		RevocableZone:          revocableZone,
+		NumaInfo:               topologyInfo,
+		EasAcceleratorFlavorId: easAcceleratorFlavorId,
```
wpeng102, line 181: task also need `EasAccelerator bool`, if EasAcceleratorFlavorId != "", task's EasAccelerator  is true

PR #2013, opened 2022-02-11
update the version in deployment file for arm64

william-wang:


hwdef:
This yaml file should be upgraded as a whole, not just the version number.

hwdef:
The last update was last December

Thor-wl:
Are there docker images for v1.5.0 and arm architecture pushed to dockerhub? @william-wang

PR #2016, opened 2022-02-13
remove mutex during numa-cnt calculation

justadogistaken:
A slight optimization.
Removed the mutex while calculating numa count.
Signed-off-by: david.bao <baojn1998@163.com>

### DIFF
--- a/pkg/scheduler/api/numa_info.go
+++ b/pkg/scheduler/api/numa_info.go
@@ -36,6 +36,8 @@ const (
 	NumaInfoMoreFlag NumaChgFlag = 0b11
 	// NumaInfoLessFlag indicate the received allocatable resource is getting less
 	NumaInfoLessFlag NumaChgFlag = 0b10
+	// DefaultMaxNodeScore indicates the default max node score
+	DefaultMaxNodeScore = 100
 )
 
 // PodResourceDecision is resource allocation determinated by scheduler,
@@ -252,4 +254,10 @@ func (resSets ResNumaSets) Clone() ResNumaSets {
 	}
 
 	return newSets
+}
+
+// ScoredNode is the wrapper for node during Scoring.
+type ScoredNode struct {
+	NodeName string
+	Score    int64
 }
--- a/pkg/scheduler/plugins/numaaware/numaaware.go
+++ b/pkg/scheduler/plugins/numaaware/numaaware.go
@@ -170,11 +170,12 @@ func (pp *numaPlugin) OnSessionOpen(ssn *framework.Session) {
 		}
 
 		scoreList := getNodeNumaNumForTask(nodeInfo, pp.assignRes[task.UID])
-		util.NormalizeScore(100, true, scoreList)
+		util.NormalizeScore(api.DefaultMaxNodeScore, true, scoreList)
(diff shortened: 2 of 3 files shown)


pkg/scheduler/plugins/util/util.go
```diff
@@ -272,29 +274,32 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
...
-		if reverse {
-			for key := range scores {
-				scores[key] = maxPriority
-			}
+	maxMinDiff := maxCount - minCount
+	for idx := range scores {
+		tmpScore := int64(0)
+
+		if maxMinDiff > 0 {
+			tmpScore = maxPriority * (scores[idx].Score - minCount) / maxMinDiff
```
shinytang6, line 297: Can you clarify why we need to change the calculation way here?

pkg/scheduler/plugins/numaaware/numaaware.go
```diff
@@ -225,21 +226,21 @@ func filterNodeByPolicy(task *api.TaskInfo, node *api.NodeInfo, nodeResSets map[
...
-	var mx sync.RWMutex
+func getNodeNumaNumForTask(nodeInfo []*api.NodeInfo, resAssignMap map[string]api.ResNumaSets) []api.ScoredNode {
+	nodeNumaCnts := make([]api.ScoredNode, len(nodeInfo))
 	workqueue.ParallelizeUntil(context.TODO(), 16, len(nodeInfo), func(index int) {
 		node := nodeInfo[index]
 		assignCpus := resAssignMap[node.Name][string(v1.ResourceCPU)]
-		mx.Lock()
-		defer mx.Unlock()
-		nodeNumaNumMap[node.Name] = int64(getNumaNodeCntForcpuID(assignCpus, node.NumaSchedulerInfo.CPUDetail))
+		nodeNumaCnts[index] = api.ScoredNode{
```
shinytang6, line 234: l am ok with the changes here

pkg/scheduler/plugins/util/util.go
shinytang6: l'm ok with both, it seems that both of the calculation methods are used in k8s )

justadogistaken: Yes, So it's better to keep the current normalization.

shinytang6 (commented):
Thx, lgtm overall )

pkg/scheduler/api/numa_info.go
```diff
@@ -253,3 +255,9 @@ func (resSets ResNumaSets) Clone() ResNumaSets {
 
 	return newSets
 }
+
+// ScoredNode is the wrapper for node during Scoring.
+type ScoredNode struct {
+	NodeName string
+	Score    int64
```
huone1, line 262: what is the intention of adding this structure?

justadogistaken: Describe a node and its score. For constructing slice of scoredNode instead of map[string]score

Thor-wl: I think it is more readable comparing with map.

PR #2021, opened 2022-02-16
fix when build with -buildmode=plugin, go complains asm error in v1.5.0

zishen:
in v1.5.0

See cespare/xxhash@v2.1.2...main for more details.

Format with Go 1.17 gofmt to use //go:build constraints
/kind bug
/assig Thor-wl

Signed-off-by: zishen softwarebtg@163.com


hwdef (commented):
please squash commit and fix DCO

PR #2029, opened 2022-02-18
Remove place holder port

hzxuzhonghu:
K8s allows creating svc without port, in volcano we just use this svc to provide dns resolve


hzxuzhonghu:
cc @Thor-wl

hwdef (commented):
/lgtm

PR #1571, opened 2021-06-29
Switch to cross-compiled docker containers & container build

holdenk:
This allows for one container name to support multiple archs, and once pushed we can remove more of the arm64 specific installation stuff since the same container names will support both.

docker does "the right thing" and pulls the container associated with the arch.

You can take a look at the containers I built with this change in my own dockerhub at https://hub.docker.com/repository/docker/holdenk/volcanosh-scheduler , https://hub.docker.com/repository/docker/holdenk/volcanosh-controller-manager , https://hub.docker.com/repository/docker/holdenk/volcanosh-webhook-manager-base , etc. 

This is in response to https://github.com/volcano-sh/volcano/issues/1570 (although it could also solve https://github.com/volcano-sh/volcano/issues/1568 since we wouldn't need volcano-development-arm64.yaml  anymore).

To preserve backwards compatibility with users who might be developing locally in single arch mode I have that the default. If there are release docs I should update as well let me know.

Signed-off-by: Holden Karau <holden@pigscanfly.ca>

PR #2030, opened 2022-02-20
update arm's yaml, and upgrade makefile to automatically generate arm's yaml

hwdef:


Yikun:
@hwdef I guess this might no needed after https://github.com/volcano-sh/volcano/pull/1571 merged, could you help to review on it? Thanks!

hwdef:
> @hwdef I guess this might no needed after #1571 merged, could you help to review on it? Thanks!

ok, I think we should fix CI first.

Issue #2037, opened 2022-02-26
Resource not allocated should be reserved for running jobs whose podgroup contains minResource in proportion plugin


Thor-wl:
**What happened**:
Background:
* Only 1 node whose allocatable resource is {cpu: 8c, memory: 16G} in the cluster
* Besides the default queue which has no jobs inside, create a queue: {name: test-queue, capability: {cpu: 2c, memory: 4G}, weight: 1}
* The `spec` for all the podgroups created for the following Spark jobs: {minResource: {cpu:2c, memory: 2G }, queue: test-queue}
* The request resource of all the pods(drivers and executors) created for the following Spark jobs: {cpu:1c, memory: 1G}

When trying to submit 2 Spark jobs at the same time, there must be that all the driver pods are running while executors are pending, which leads to resource dead lock and all the Spark jobs cannot work well.

**What you expected to happen**:
If one of the Spark jobs is admitted to allocated resource, Volcano should reserve resource for it according to the minResouce.

**How to reproduce it (as minimally and precisely as possible)**:
Pls refer to the above

**Anything else we need to know?**:

**Environment**:
- Volcano Version: latest
- Kubernetes version (use `kubectl version`): v1.23
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release): Ubuntu
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #2038, opened 2022-02-26
Fix the resource calculation error for running jobs whose podgroup contains minResource

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/2037

### DIFF
--- a/pkg/scheduler/plugins/overcommit/overcommit.go
+++ b/pkg/scheduler/plugins/overcommit/overcommit.go
@@ -88,10 +88,21 @@ func (op *overcommitPlugin) OnSessionOpen(ssn *framework.Session) {
 	}
 	op.idleResource = total.Clone().Multi(op.overCommitFactor).Sub(used)
 
-	// calculate inqueue job resources
 	for _, job := range ssn.Jobs {
+		// calculate inqueue job resources
 		if job.PodGroup.Status.Phase == scheduling.PodGroupInqueue && job.PodGroup.Spec.MinResources != nil {
 			op.inqueueResource.Add(api.NewResource(*job.PodGroup.Spec.MinResources))
+			continue
+		}
+		// calculate inqueue resource for running jobs
+		// the judgement 'job.PodGroup.Status.Running >= job.PodGroup.Spec.MinMember' will work on cases such as the following condition:
+		// Considering a Spark job is completed(driver pod is completed) while the podgroup keeps running, the allocated resource will be reserved again if without the judgement.
+		if job.PodGroup.Status.Phase == scheduling.PodGroupRunning &&
+			job.PodGroup.Spec.MinResources != nil &&
+			job.PodGroup.Status.Running >= job.PodGroup.Spec.MinMember {
+			allocated := util.GetAllocatedResource(job)
+			inqueued := util.GetInqueueResource(job, allocated)
+			op.inqueueResource.Add(inqueued)
 		}
 	}
 
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -140,6 +140,17 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		if job.PodGroup.Status.Phase == scheduling.PodGroupInqueue {
 			attr.inqueue.Add(job.GetMinResources())
 		}
+
+		// calculate inqueue resource for running jobs
+		// the judgement 'job.PodGroup.Status.Running >= job.PodGroup.Spec.MinMember' will work on cases such as the following condition:
+		// Considering a Spark job is completed(driver pod is completed) while the podgroup keeps running, the allocated resource will be reserved again if without the judgement.
+		if job.PodGroup.Status.Phase == scheduling.PodGroupRunning &&
+			job.PodGroup.Spec.MinResources != nil &&
+			job.PodGroup.Status.Running >= job.PodGroup.Spec.MinMember {
+			allocated := util.GetAllocatedResource(job)
+			inqueued := util.GetInqueueResource(job, allocated)
+			attr.inqueue.Add(inqueued)
+		}
 	}
 
 	// Record metrics
--- a/pkg/scheduler/plugins/util/util.go
+++ b/pkg/scheduler/plugins/util/util.go
@@ -298,3 +298,48 @@ func NormalizeScore(maxPriority int64, reverse bool, scores []api.ScoredNode) {
 		scores[idx].Score = score
 	}
 }
+
+// GetAllocatedResource returns allocated resource for given job
+func GetAllocatedResource(job *api.JobInfo) *api.Resource {
+	allocated := &api.Resource{}
+	for status, tasks := range job.TaskStatusIndex {
+		if api.AllocatedStatus(status) {
+			for _, t := range tasks {
+				allocated.Add(t.Resreq)
+			}
+		}
+	}
+	return allocated
+}
+
+// GetInqueueResource returns reserved resource for running job whose part of pods have not been allocated resource.
+func GetInqueueResource(job *api.JobInfo, allocated *api.Resource) *api.Resource {
+	inqueue := &api.Resource{}
+	for rName, rQuantity := range *job.PodGroup.Spec.MinResources {
+		switch rName {
+		case v1.ResourceCPU:
+			reservedCPU := float64(rQuantity.Value()) - allocated.MilliCPU
+			if reservedCPU > 0 {
+				inqueue.MilliCPU = reservedCPU
+			}
+		case v1.ResourceMemory:
+			reservedMemory := float64(rQuantity.Value()) - allocated.Memory
+			if reservedMemory > 0 {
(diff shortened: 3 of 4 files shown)


pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -140,6 +143,36 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		if job.PodGroup.Status.Phase == scheduling.PodGroupInqueue {
 			attr.inqueue.Add(job.GetMinResources())
 		}
+		if job.PodGroup.Status.Phase == scheduling.PodGroupRunning && job.PodGroup.Spec.MinResources != nil {
```
shinytang6, line 146: Should we also consider `scheduling.PodGroupInqueue`?

shinytang6 (changes requested):

pkg/scheduler/plugins/proportion/proportion.go
Thor-wl: What do you mean? The code above has already added `minResource` of Inqueue jobs into `queue.inqueue`.

Thor-wl: If the phase of podgroup is `Inqueue`, all the `minResource` has already been taken into `attr.inqueue`.

shinytang6: got it.

pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -88,10 +88,17 @@ func (op *overcommitPlugin) OnSessionOpen(ssn *framework.Session) {
 	}
 	op.idleResource = total.Clone().Multi(op.overCommitFactor).Sub(used)
 
-	// calculate inqueue job resources
 	for _, job := range ssn.Jobs {
+		// calculate inqueue job resources
 		if job.PodGroup.Status.Phase == scheduling.PodGroupInqueue && job.PodGroup.Spec.MinResources != nil {
 			op.inqueueResource.Add(api.NewResource(*job.PodGroup.Spec.MinResources))
```
william-wang, line 94: need to consider below conner case while PodGroup is in `Inqueue`. 
driver pod is running, part or all of executor pod are pending. The already allocated resource might need to be excluded from op.inqueueResource  and then compare with `idle` resource.

Thor-wl: > need to consider below conner case while PodGroup is in `Inqueue`. driver pod is running, part or all of executor pod are pending. The already allocated resource might need to be excluded from op.inqueueResource and then compare with `idle` resource.

It will not be at `inqueue` phase for Spark application because it will be turn into `running` at the end of the first session. The default `minMember` will be 1 and the allocated pod number is at least 1.

pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -88,10 +88,17 @@ func (op *overcommitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			continue
+		}
+		// calculate inqueue resource for running jobs
+		if job.PodGroup.Status.Phase == scheduling.PodGroupRunning && job.PodGroup.Spec.MinResources != nil {
+			allocated := util.GetAllocatedResource(job)
+			inqueued := util.GetInqueueResource(job, allocated)
+			op.inqueueResource.Add(inqueued)
```
william-wang, line 101: Add the consideration for this case, Spark app is driver pod is completed and podgroup is running.
In this case, the inqueued should be `0` . The judgement might be as below.
job.PodGroup.Status.Phase == scheduling.PodGroupRunning && jobInfo.PodGroup.Spec.MinMember > `allocated numbers in job`

pkg/scheduler/plugins/util/util.go
```diff
@@ -298,3 +298,48 @@ func NormalizeScore(maxPriority int64, reverse bool, scores []api.ScoredNode) {
 		scores[idx].Score = score
 	}
 }
+
+// GetAllocatedResource returns allocated resource for given job
```
william-wang, line 302: The jobinfo has `Allocated` field, whether it is ok in this case.

william-wang (changes requested):

pkg/scheduler/plugins/util/util.go
Thor-wl: Yes, but it is initial status when used in `OnSessionOpen`, namely, it's nil.

pkg/scheduler/plugins/overcommit/overcommit.go
Thor-wl: Yes, I think it is more suitable to add a phase `Completed` for podgroup.

Thor-wl: IC, that's right. Let me modify the condition.

william-wang: Yes. new `state` needs to be added for `podgroup`. Right now the hot fix here is ok for me.  Please add detail comments here. We will refactor it later.

pkg/scheduler/plugins/util/util.go
william-wang: got it.

pkg/scheduler/plugins/overcommit/overcommit.go
shinytang6: Yes, podgroup `completed` state is necessary. Currently, in addition to vcjob which deletes the pg when it completes, the pg created by third-party job will not be deleted after the task is completed(though AFAIK spark operator [managed it itself](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/v1beta2-1.3.3-3.1.1/pkg/batchscheduler/volcano/volcano_scheduler.go#L164)), but will remain in inqueue or running, which causes unnecessary inqueue computations.

shinytang6: But there is a detail l want to confirm, I think `job.PodGroup.Status.Running >= job.PodGroup.Spec.MinMember` may should be replaced by `job.TaskStatusIndex[api.Running] >= job.PodGroup.Spec.MinMember`?

shinytang6 (changes requested):

william-wang (approved):
/lgtm

william-wang:
@Thor-wl make the ci happy.

pkg/scheduler/plugins/overcommit/overcommit.go
shinytang6: you are right

shinytang6 (approved):
/lgtm
/approve

Yikun:
I tested this PR with Spark integration test:
- minRes with default queue
```
  test("SPARK-38187: Run SparkPi Jobs with minCPU", k8sTestTag, volcanoTag) {
    val groupName = generateGroupName("min-cpu")
    val cpuAllocatable = getNode().getStatus.getAllocatable.get("cpu").getAmount
    val jobNum = 4
    (1 to jobNum).map { i =>
      Future {
        runJobAndVerify(
          i.toString,
          groupLoc = Option(groupName),
          minCPU = Option(cpuAllocatable))
      }
    }
    // Check Pending jobs completed one by one
    (1 until jobNum).map { completedNum =>
      Eventually.eventually(TIMEOUT, INTERVAL) {
        val pendingPods = getPods(role = "driver", groupName, statusPhase = "Pending")
        assert(pendingPods.size === jobNum - completedNum)
      }
    }
    // All jobs succeeded finally
    Eventually.eventually(TIMEOUT, INTERVAL) {
      val succeededPods = getPods(role = "driver", groupName, statusPhase = "Succeeded")
      assert(succeededPods.size === jobNum)
    }
  }
```

- minRes with resource limit queue (2 cpu)
```scala
  test("SPARK-38187: Run SparkPi Jobs with minCPU", k8sTestTag, volcanoTag) {
    // Create a queue with 2 CPU, 3G memory capacity
    createOrReplaceYAMLResource(VOLCANO_RES_LIMIT_QUEUE_YAML)
    val jobNum = 4
    (1 to jobNum).map { i =>
      Future {
        runJobAndVerify(
          i.toString,
          groupLoc = Option(GROUP_PREFIX),
          queue = Option("queue-limit-resource"),
          minCPU = Option("2"))
      }
    }
    // Check Pending jobs completed one by one
    (1 until jobNum).map { completedNum =>
      Eventually.eventually(TIMEOUT, INTERVAL) {
        val pendingPods = getPods(role = "driver", GROUP_PREFIX, statusPhase = "Pending")
        assert(pendingPods.size === jobNum - completedNum)
      }
    }
    // All jobs succeeded finally
    Eventually.eventually(TIMEOUT, INTERVAL) {
      val succeededPods = getPods(role = "driver", GROUP_PREFIX, statusPhase = "Succeeded")
      assert(succeededPods.size === jobNum)
    }
    deleteYAMLResource(VOLCANO_RES_LIMIT_QUEUE_YAML)
  }
```

Yikun (approved):
LGTM, @Thor-wl @william-wang @shinytang6 Thanks for all your helps!

Issue #2053, opened 2022-03-06
float64 error from arguments.go:65 when adding overcommit-factor to configmap


tFable:
**What happened**:
After changing the `configmap` to add a `overcommit-factor` and installing volcano from [this](https://github.com/volcano-sh/volcano/blob/release-1.5/installer/volcano-development.yaml) file on the `v1.5` branch and changing the configmap getting this log continuously 

```
arguments.go:65] Could not parse argument: %!s(float64=1) for key overcommit-factor to float64, with err strconv.ParseFloat: parsing "": invalid syntax
```

This is how the `overcommit-factor` was defined:

```
# Source: volcano/templates/scheduler.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: volcano-scheduler-configmap
  namespace: volcano-system
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
      - name: conformance
    - plugins:
      - name: overcommit
        arguments:
          overcommit-factor: 1.0
      - name: drf
      - name: predicates
      - name: proportion
      - name: nodeorder
      - name: binpack
```

However, if on the `volcano-development.yaml` we change the docker images to pull a specific version (1.5.0) of the below containers (instead of `:latest` that is the default on the branch) the error doesn't occur.

```
volcanosh/vc-webhook-manager:v1.5.0
volcanosh/vc-controller-manager:v1.5.0
volcanosh/vc-scheduler:v1.5.0
```
That said, I am still not seeing the overcommit-factor be 1.0 (queued jobs are still having `pending` pods created).

**What you expected to happen**:

For no float error message in the logs and for the overcommit-ratio change to actually "take".

**How to reproduce it (as minimally and precisely as possible)**:
Install Volcano from the `volcano-development.yaml` file of the `v1.5` branch. Prior to that, add an argument for the `overcommit-ratio` to the `overcommit` plugin.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: 1.5
- Kubernetes version (use `kubectl version`):1.23
- Cloud provider or hardware configuration:vSphere on-repm
- OS (e.g. from /etc/os-release): 
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Thor-wl:
@tFable Thanks for the feedback.
/assign

Thor-wl:
[...]

PR #2061, opened 2022-03-08
fix that parse the scheduler configuration argument with blank error

Thor-wl:
ref：https://github.com/volcano-sh/volcano/issues/2053
The root reason is that argument is collected with interface{} now and its value is valid type. Namely, it cannot be converted to string first or it will be error. Unfortunely, the original code ignore the error with `_` and the `value` gets the default value "".

### DIFF
--- a/pkg/scheduler/framework/arguments.go
+++ b/pkg/scheduler/framework/arguments.go
@@ -17,8 +17,6 @@ limitations under the License.
 package framework
 
 import (
-	"strconv"
-
 	"k8s.io/klog"
 
 	"volcano.sh/volcano/pkg/scheduler/conf"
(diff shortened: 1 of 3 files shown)


hwdef (commented):
/lgtm

pkg/scheduler/framework/arguments.go
```diff
@@ -38,10 +36,9 @@ func (a Arguments) GetInt(ptr *int, key string) {
 		return
 	}
 
-	valueStr, _ := argv.(string)
-	value, err := strconv.Atoi(valueStr)
-	if err != nil {
-		klog.Warningf("Could not parse argument: %s for key %s to int, with err %v", argv, key, err.Error())
+	value, ok := argv.(int)
+	if !ok {
+		klog.Warningf("Could not parse argument: %s for key %s to int", argv, key)
```
shinytang6, line 41: pls use %v for argv

shinytang6 (changes requested):
Others LGTM

pkg/scheduler/framework/arguments.go
Thor-wl: OK. Let me upate that.

william-wang (approved):
/lgtm

Issue #2068, opened 2022-03-11
Make sure `Open` state queue can be deleted until we support queue close in kubectl


Yikun:
**What happened**:
`Open` state queue can't be deleted in kubectl, this is not very convenient for users.

We'd better to make sure `Open` state queue can be deleted until we support queue operations in kubectl

**What you expected to happen**:
`Open` state queue can be deleted 

**How to reproduce it (as minimally and precisely as possible)**:
create a queue can't be deleted by kubectl

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`): 1.23

hwdef:
https://github.com/volcano-sh/volcano/issues/2058
https://github.com/volcano-sh/volcano/issues/1687

Please combine the content of these two issues and the discussion at the last community meeting to make a comprehensive judgment

Thor-wl:
It may not be so reasonable for **open** queue to be deleted directly. There may be still some running and new coming jobs.

Issue #2058, opened 2022-03-07
Suggest to add a field to queue.spec indicating queue open or closed


hwdef:
#### What would you like to be added:

Add a field to queue.spec, by modifying this field, the queue can be turned on or off.

#### Why is this needed:

After adding this field, the user can open or close the queue through kubectl.

REF: #1687

#### optional name:

* switch: open/close
* state: open/close
* open: true/false
* ...

k82cn:
Can we delete all jobs fristly, and then delete the queue; we can handle those actions in `QueueController`, similar to namespace.

hwdef:
> Can we delete all jobs fristly, and then delete the queue; we can handle those actions in `QueueController`, similar to namespace.

It's really simpler, but I am worried that this will delete the job by mistake.
What if when deleting a queue, check if there are still jobs in the queue, and refuse to delete the queue if there are jobs?
I'm not sure which one to use, the one you said looks good and is more in line with the logic of kubernetes.

Yikun:
From api view, I think it's reasonable, 

then we can use yaml to control queue deletion, and I'm not sure the final UX for this.

I hope just use kubectl apply yaml to set queue spec rather than kubectl volcano plugin or something. 

Because we still need to allow users use API interfaces(golang API/java API) to set spec. The better way is use crd capability directly.

Yikun:
And we might want to think about the rleationship of original state and spec.state.

jyn244636695:
Is this question have been solved?

hwdef:
> Is this question have been solved?

no, this isuse is closed.

Issue #1687, opened 2021-08-19
Deleting a queue with `kubectl` fails with "only queue with state `Closed` can be deleted"


prokher:
**What happened**:

An attempt to delete a queue with `kubectl delete` ends with the following error:
```
Error from server: error when deleting "ex2-volcano.yaml": admission webhook "validatequeue.volcano.sh" denied the request: only queue with state `Closed` can be deleted, queue `one-q` state is `Open`
Error from server (NotFound): error when deleting "ex2-volcano.yaml": jobs.batch.volcano.sh "j1-vj" not found
```

**What you expected to happen**:

I expect queue be deleted.

**How to reproduce it (as minimally and precisely as possible)**:

Just create a new queue and try to delete it with `kubectl delete`.

**Anything else we need to know?**:

I realize I can close the queue first with `vcctl`, but I need a pure Kubernetes solution without extra CLI tools. I am going to eventually create/delete queues through Kubernetes API so I cannot rely on some external CLI utilities. I tried to edit a queue with `kubectl edit` but unfortunately could not close it this way.

**Environment**:
- Volcano Version: 1.3.0
- Kubernetes version (use `kubectl version`):
  ```
  Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T17:56:19Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"darwin/amd64"}
  Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:53:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
  ```

Thor-wl:
As the design of `queue`, only queues with `closed` status, which means it will no longer receive new coming jobs, can be deleted. `queue` is a concept inheriting from traditional HPC scheduling architectures and widely used for cluster resource sharing and division among departments or groups. As to default scheduler of Kubernets, it's a scheduling for common scenarios. So it contains no `queue`.

prokher:
[...]

PR #2069, opened 2022-03-11
Make sure Open state queue can be deleted

Yikun:
Make sure `Open` state queue can't be deleted in kubectl, this is not very convenient for users. We'd better to make sure `Open` state queue can be deleted until we support queue operations in kubectl.

Signed-off-by: Yikun Jiang <yikunkero@gmail.com>

Closes: https://github.com/volcano-sh/volcano/issues/2068

### DIFF
--- a/pkg/webhooks/admission/queues/validate/validate_queue.go
+++ b/pkg/webhooks/admission/queues/validate/validate_queue.go
@@ -198,14 +198,9 @@ func validateQueueDeleting(queue string) error {
 		return fmt.Errorf("`%s` queue can not be deleted", "default")
 	}
 
-	q, err := config.VolcanoClient.SchedulingV1beta1().Queues().Get(context.TODO(), queue, metav1.GetOptions{})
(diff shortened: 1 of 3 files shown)


Yikun:
cc @william-wang

hwdef:
https://github.com/volcano-sh/volcano/issues/2058
https://github.com/volcano-sh/volcano/issues/1687

Please combine the content of these two issues and the discussion at the last community meeting to make a comprehensive judgment

hwdef (commented):
/hold
1. Not sure if this approach is widely accepted.
2. The case where there are jobs in the queue is not processed

william-wang:
The `Open` queue can not be deleted. It is a good feature in some scenarios. However this feature also cause some inconvenience for users due to it is not supported with kubectl and vcctl needs a extra installation.
So the #1687 is working on to support delete queue by kubectl.  I agree disable/remove the queue validation prior to the pr is merged.

Issue #2072, opened 2022-03-11
Change installer imagePullPolicy to `Always`


Yikun:
#### What would you like to be added:
Change installer imagePullPolicy to `Always`

#### Why is this needed:
We doc installer in https://volcano.sh/en/docs/installation/ ,

We'd better use the always download images for users who want to use latest volcano

Yikun:
We'd better add a `volcano-latest.yaml`  with `Always` to make sure users download new latest images.

PR #2073, opened 2022-03-11
Add a new `volcano-latest.yaml` with `Always` imagePullPolicy

Yikun:
Add a new `volcano-latest.yaml` for users want to always use remote new latest image.

#### Why is this needed:
some users want to always use remote new latest image.

Signed-off-by: Yikun Jiang <yikunkero@gmail.com>

#### How to test?
```
make update-development-yaml
# To make sure only policy replaced by Always
diff installer/volcano-latest-arm64.yaml installer/volcano-development-arm64.yaml
diff installer/volcano-latest.yaml installer/volcano-development.yaml
```
Closes: https://github.com/volcano-sh/volcano/issues/2072

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,4 +199,8 @@ update-development-yaml:
 	make generate-yaml TAG=latest RELEASE_DIR=installer
 	cp installer/volcano-latest.yaml installer/volcano-development-arm64.yaml
 	sed -r -i 's#(.*)image:([^:]*):(.*)#\1image:\2-arm64:\3#'  installer/volcano-development-arm64.yaml
-	mv installer/volcano-latest.yaml installer/volcano-development.yaml
+	cp installer/volcano-latest.yaml installer/volcano-development.yaml
+	sed -i 's/PullPolicy: IfNotPresent/PullPolicy: Always/g' installer/volcano-latest.yaml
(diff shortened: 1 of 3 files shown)


Yikun:
This change is not enough, because CI are using `IfNotPresent` image policy to make sure latest are using latest PR to test.

Yikun:
We'd better add a `volcano-latest.yaml` with `Always` to make sure users download latest images.

And keep the `volcano-developer.yaml` as current behavior to keep same UX for developer.

Yikun:
cc @hwdef could you take a look I think you are very familar with this.

Also cc @Thor-wl @william-wang How about this? Add a new yaml, and still keep origin one for developers and CI.

hwdef:
I think it is also good to keep only one file. Because the tag is `latest`, user want to use latest images. And because image is layered, very little traffic is generated if the mirror has not changed.

It would also work if you want to save two files, but I recommend adding documentation explaining how the two files are different

Yikun:
@hwdef 
- dev suffix yaml is for developers and CI, dev can use

```
TAG: latest make images
k apply dev.yml
```

to validate dev work.

- latest suffix yaml is for users, they always want to download latest published images.

Yep, we need doc, latest.yml will be shown in website, dev.yml will be shown in developer doc.

PR #2081, opened 2022-03-13
[cherry-pick]fix the driver pod can not be created due to unreasonable admit

wpeng102:


william-wang (approved):
/lgtm

Issue #2088, opened 2022-03-15
Admission MutatingWebhook will influence  kubectl creating pod performance


berlinsaint:
volcano 1.2-> 1.5

Because of some reason, we can't  use volcano admission default config.

we change the yaml to this:

![image](https://user-images.githubusercontent.com/10629930/158293999-a7de50fe-05d7-43e7-b3e8-1b57a2d91876.png)

and will generate the mutating config:

![image](https://user-images.githubusercontent.com/10629930/158294088-a9aa98e2-4b04-4633-a37b-6d91e926b683.png)

when creating any pod ; it will call the webhook:
admission component have this log:

![image](https://user-images.githubusercontent.com/10629930/158294158-663b76f1-3f8d-4a61-92bf-ded317d185f6.png)

and creating process will have  [kubectl  create pod  very slow ] 30s delay

FengXingYuXin:
It looks like your caBundle is bad? and your MutatingWebhookConfiguration's resources is pods, so it makes effect when you create pods in any ways.

berlinsaint:
> It looks like your caBundle is bad? and your MutatingWebhookConfiguration's resources is pods, so it makes effect when you create pods in any ways.

yes  and i have fix them here: https://github.com/berlinsaint/volcano/commit/ee239b086ad56f33beb0a43ec780bf8e349d7a09

PR #2089, opened 2022-03-15
fix: fix admission not support ip address which will cause ssl error.…

berlinsaint:
Fix [issue](https://github.com/volcano-sh/volcano/issues/2088)

Signed-off-by: maybaby <berlinsaint@126.com>

### DIFF
--- a/installer/dockerfile/webhook-manager/gen-admission-secret.sh
+++ b/installer/dockerfile/webhook-manager/gen-admission-secret.sh
@@ -16,6 +16,7 @@ The following flags are required.
        --service          Service name of webhook.
        --namespace        Namespace where webhook service and secret reside.
        --secret           Secret name for CA certificate and server certificate/key pair.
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
IP is not a good practice as the pod can be down/up dynamically. Do you want to generate certificates dynamically?

berlinsaint:
> IP is not a good practice as the pod can be down/up dynamically. Do you want to generate certificates dynamically?
yes but our ip is vip , there are some cases that we we can't use svc as i point at the issue.  our kube master is on another main kube cluster, so they can't callback the sub kubecluster's svc

berlinsaint:
> IP is not a good practice as the pod can be down/up dynamically. Do you want to generate certificates dynamically?

And i think  this pr is just a small support for the ips, you see that choices is optional, not required

hzxuzhonghu:
Kind of understanding your k8s cluster, that means in your scene, native webhook can not be used.  Looks it make sense.

PR #2094, opened 2022-03-16
add tool functions BuildBestEffortPod and BuildPodWithPriority

Thor-wl:
The tool functions will be used in the UT of shuffle action. 
Ref: https://github.com/volcano-sh/volcano/pull/2099/files

### DIFF
--- a/pkg/scheduler/util/test_utils.go
+++ b/pkg/scheduler/util/test_utils.go
@@ -68,8 +68,8 @@ func BuildNode(name string, alloc v1.ResourceList, labels map[string]string) *v1
 	}
 }
 
-// BuildPod builts Pod object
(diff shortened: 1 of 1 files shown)


shinytang6 (commented):
why not include this pr in https://github.com/volcano-sh/volcano/pull/2099?

Thor-wl:
> why not include this pr in #2099?

I also suggest like that. @k82cn advised to separate the whole PR into different small functions in order to have a good glimpse of the code for reviewer. I think that is also reasonable. These PRs should be merged beferore shuffle action so that that PR can pass the CI.

k82cn:
> I also suggest like that. @k82cn advised to separate the whole PR into different small functions in order to have a good glimpse of the code for reviewer.

My suggstion is to create two PRs, one for action, the other for plugins; and those two PRs should not have dependency with each other :)

PR #2095, opened 2022-03-16
Support executing registered functions registered in plugins regularly

Thor-wl:
This function will judge whether the functions registered in plugins should be executed currently, which is useful in scenarios such as rescheduling. This feature has already been used in rescheduling plugin. 
Ref: https://github.com/volcano-sh/volcano/pull/2098/files

### DIFF
--- a/pkg/scheduler/util/scheduler_helper.go
+++ b/pkg/scheduler/util/scheduler_helper.go
@@ -23,6 +23,7 @@ import (
 	"math/rand"
 	"sort"
 	"sync"
+	"time"
 
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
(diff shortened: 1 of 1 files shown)


pkg/scheduler/util/scheduler_helper.go
```diff
@@ -39,8 +40,23 @@ var lastProcessedNodeIndex int
 // Reservation is used to record target job and locked nodes
 var Reservation *ResourceReservation
 
+// FnsLastExecTime records the last execution time of functions registering in the plugins and executing periodically
+var FnsLastExecTime map[string]time.Time
+
 func init() {
 	Reservation = NewResourceReservation()
+	FnsLastExecTime = make(map[string]time.Time)
```
k82cn, line 48: How to register `an exectTimeFn`?

PR #2096, opened 2022-03-16
add NodeUsage struct to support realtime scheduling

Thor-wl:
This struct is prepared for the implemtation of rescheduling and realtime scheduling. It is already been used in the rescheduling plugin.
Ref: https://github.com/volcano-sh/volcano/pull/2098/files

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -34,6 +34,11 @@ func (o *AllocateFailError) Error() string {
 	return o.Reason
 }
 
+type NodeUsage struct {
+	CPUUsageAvg map[string]float64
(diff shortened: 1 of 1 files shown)


pkg/scheduler/api/node_info.go
```diff
@@ -34,6 +34,11 @@ func (o *AllocateFailError) Error() string {
 	return o.Reason
 }
 
+type NodeUsage struct {
+	CPUUsageAvg map[string]float64
```
wpeng102, line 38: what does the CPUUsageAvg key mean?

william-wang:
/hold
The “ResourceUsage“ will be included in another pr which contains volcnao integrating with metrics (prometheus).

PR #2098, opened 2022-03-16
add rescheduling plugin

Thor-wl:
design doc: https://github.com/volcano-sh/volcano/blob/master/docs/design/rescheduling.md
This PR should work together with https://github.com/volcano-sh/volcano/pull/2094 https://github.com/volcano-sh/volcano/pull/2095 https://github.com/volcano-sh/volcano/pull/2096. So please **DO NOT** merge it until the above PR are merged.

### DIFF
--- a/pkg/scheduler/plugins/factory.go
+++ b/pkg/scheduler/plugins/factory.go
@@ -29,6 +29,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/plugins/predicates"
 	"volcano.sh/volcano/pkg/scheduler/plugins/priority"
 	"volcano.sh/volcano/pkg/scheduler/plugins/proportion"
+	"volcano.sh/volcano/pkg/scheduler/plugins/rescheduling"
 	"volcano.sh/volcano/pkg/scheduler/plugins/reservation"
 	"volcano.sh/volcano/pkg/scheduler/plugins/sla"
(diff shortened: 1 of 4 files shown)


pkg/scheduler/framework/session_plugins.go
```diff
@@ -503,6 +507,30 @@ func (ssn *Session) ReservedNodes() {
 	}
 }
 
+// Victims returns the victims for rescheduling
+func (ssn *Session) Victims(tasks []*api.TaskInfo) map[*api.TaskInfo]bool {
```
k82cn, line 511: This function should be part of action instead of this plugin.

Thor-wl: `Victims` get registered functions from `ssn.reschedulingFns` and then execute them. I think it is better add it in this PR.

k82cn: One PR for framework (shuffle action), one pr for implementation (rescheduling plugin).
`Victims` should be also used by TDM plugin.

Thor-wl: > One PR for framework (shuffle action), one pr for implementation (rescheduling plugin).
> `Victims` should be also used by TDM plugin.

I get you. It's OK for me. Let me make a update.

Issue #1955, opened 2022-01-13
Add shuffle action for descheduler & tdm plugins


k82cn:
#### What would you like to be added:

Let's add a shuffle action for descheduler & tdm plugins; in this action, it'll only evict all tasks return from `victimTasks`.

#### Why is this needed:

Currently, victim tasks are evicted in `preempt` action which combine with preemption; it's better to have a stand alone action for victim tasks :)

PR #2099, opened 2022-03-16
add shuffle action

Thor-wl:
design doc: https://github.com/volcano-sh/volcano/blob/master/docs/design/rescheduling.md

### DIFF
--- a/pkg/scheduler/actions/shuffle/shuffle_test.go
+++ b/pkg/scheduler/actions/shuffle/shuffle_test.go
@@ -1,0 +1,331 @@
+/*
+ Copyright 2022 The Volcano Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package shuffle
+
+import (
+	"testing"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+
+	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/cache"
+	"volcano.sh/volcano/pkg/scheduler/conf"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/plugins/rescheduling"
+	"volcano.sh/volcano/pkg/scheduler/util"
+)
+
+func TestShuffle(t *testing.T) {
+	framework.RegisterPluginBuilder("rescheduling", rescheduling.New)
+	defer framework.CleanupPluginBuilders()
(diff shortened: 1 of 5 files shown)


k82cn:
> This PR should work together with https://github.com/volcano-sh/volcano/pull/2094 https://github.com/volcano-sh/volcano/pull/2095 https://github.com/volcano-sh/volcano/pull/2096. So please DO NOT merge it until the above PR are merged.

Why does this PR have such kind of dependencies?

k82cn:
Please help to build UT for this action instead of reusing rescheduling UT.

pkg/scheduler/api/types.go
```diff
@@ -171,3 +171,6 @@ type VictimTasksFn func() []*TaskInfo
 
 // AllocatableFn is the func declaration used to check whether the task can be allocated
 type AllocatableFn func(*QueueInfo, *TaskInfo) bool
+
+// VictimsFromCandidatesFn is the func declaration used to select victim tasks from candidates and evict them
+type VictimsFromCandidatesFn func([]*TaskInfo) []*TaskInfo
```
k82cn, line 176: Why not reuse `VictimTasksFn`?

pkg/scheduler/actions/shuffle/shuffle_test.go
```diff
@@ -0,0 +1,331 @@
...
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+
+	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/cache"
+	"volcano.sh/volcano/pkg/scheduler/conf"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/plugins/rescheduling"
```
k82cn, line 33: shuffle action should not dependent on rescheduling plugin.

pkg/scheduler/api/types.go
Thor-wl: `VictimTaskFn` is a function without input parameters, which is not so suitable for shuffle action. In shuffle action, the input parameter should be a collection of running tasks.

pkg/scheduler/actions/shuffle/shuffle_test.go
Thor-wl: Just as `allocate_test.go` and `preempt_test.go` works, all the UT of actions should work with some registered plugins and then get an expected output. So this UT registers `rescheduling` plugin.

pkg/scheduler/api/types.go
k82cn: That should be ok to update `victimTasksFn` for that.

pkg/scheduler/actions/shuffle/shuffle_test.go
k82cn: Let's build a mock plugin for action UT, and we should also update allocate_test & preempt_test to have mock plugins.

Thor-wl: It is reasonable in some degree. Perhaps we can submit another PR especially for that.

k82cn: > It is reasonable in some degree. Perhaps we can submit another PR especially for that.

How to handle CI error?

pkg/scheduler/api/types.go
shinytang6: +1 for resuing VictimTasksFn

Thor-wl: OK. Let me give a upate.

william-wang:
> ref: #1777 design doc: https://github.com/volcano-sh/volcano/blob/master/docs/design/rescheduling.md Signed-off-by: Thor-wl [13164644535@163.com](mailto:13164644535@163.com)

ref: #1955
Associate this pr to issue #1955. the shuffle action will be a seperated common action for multiple plugins.

PR #2103, opened 2022-03-18
update helm version in install.sh

hwdef:

### DIFF
--- a/hack/lib/install.sh
+++ b/hack/lib/install.sh
@@ -82,9 +82,8 @@ function install-helm {
   if [[ $? -ne 0 ]]; then
     echo "Installing helm via script"
     HELM_TEMP_DIR=$(mktemp -d)
-    curl https://raw.githubusercontent.com/helm/helm/master/scripts/get > ${HELM_TEMP_DIR}/get_helm.sh
-    # TODO: There are some issue with helm's latest version, remove '--version' when it get fixed.
-    chmod 700 ${HELM_TEMP_DIR}/get_helm.sh && ${HELM_TEMP_DIR}/get_helm.sh --version v3.5.3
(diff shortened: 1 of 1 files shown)


hack/lib/install.sh
```diff
@@ -82,9 +82,8 @@ function install-helm {
   if [[ $? -ne 0 ]]; then
     echo "Installing helm via script"
     HELM_TEMP_DIR=$(mktemp -d)
-    curl https://raw.githubusercontent.com/helm/helm/master/scripts/get > ${HELM_TEMP_DIR}/get_helm.sh
-    # TODO: There are some issue with helm's latest version, remove '--version' when it get fixed.
-    chmod 700 ${HELM_TEMP_DIR}/get_helm.sh && ${HELM_TEMP_DIR}/get_helm.sh --version v3.5.3
+    curl -fsSL -o ${HELM_TEMP_DIR}/get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3
+    chmod 700 ${HELM_TEMP_DIR}/get_helm.sh && ${HELM_TEMP_DIR}/get_helm.sh
```
Thor-wl, line 86: Suggest to only give the executing priviledge for the file.

hwdef: https://helm.sh/docs/intro/install/

Helm officially recommends giving `700` permissions, I think we'd better set it according to the official recommendation

shinytang6: have you tested it with the latest version of helm? The original comments reports the potential risks.

hwdef: @shinytang6 
Yes, I tested it.
The comment said that there is a problem with the latest version. After I checked it, I found that there was already this comment when this file was introduced, and the version of helm has been upgraded twice since then. I guess the problem mentioned originally should no longer exist.

shinytang6 (approved):
/lgtm

PR #2105, opened 2022-03-21
support elastic annotation in preempt/reclaim plugin

qiankunli:
support a new annotation `volcano.sh/elastic` in pod, if `volcano.sh/elastic` = "false" (default value is true), preempt and reclaim plugin will not evict the pod.
in AI/spark case, there are always two role of pod: master/worker or driver/executor,  if master pod is evicted, the worker pod will be fail too, so it may be necessary to make sure the master pod can not to be evicted.

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -119,6 +119,9 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 					if task.Resreq.IsEmpty() {
 						return false
 					}
+					if !task.Preemptable {
+						return false
+					}
 					job, found := ssn.Jobs[task.Job]
 					if !found {
 						return false
--- a/pkg/scheduler/actions/preempt/preempt_test.go
+++ b/pkg/scheduler/actions/preempt/preempt_test.go
@@ -177,9 +177,8 @@ func TestPreempt(t *testing.T) {
 					},
 				},
 			},
-
 			pods: []*v1.Pod{
-				util.BuildPod("c1", "preemptee1", "n1", v1.PodRunning, util.BuildResourceList("1", "1G"), "pg1", make(map[string]string), make(map[string]string)),
+				util.BuildPod("c1", "preemptee1", "n1", v1.PodRunning, util.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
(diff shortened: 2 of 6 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -120,6 +120,7 @@ type TaskInfo struct {
 	VolumeReady bool
 	Preemptable bool
 	BestEffort  bool
+	Elastic     bool
```
Thor-wl, line 123: Why not resure `Preemptable` filed?

qiankunli: Preemptable `volcano.sh/preemptable` had  already be occupied by tdm plugin. and its default value is false. so  it can not be used directly.

k82cn: IMO, we can set the elastic pods to be preemptable :)

k82cn: > Preemptable volcano.sh/preemptable had already be occupied by tdm plugin. and its default value is false. so it can not be used directly.

Preemptable is not only used for TDM plugin, we'd like to make annotation more genral for all plugins :)

qiankunli: @k82cn In preempt/relclaim case, we want the default value of Preemptable is true, in tdm case,  the default value of Preemptable is false, How do we deal with this difference？

wpeng102: preemptable default value is false, which means pod can not be preempted. we can set worker/executor pods `preemptable=true` to describe your workload.

qiankunli: @wpeng102 it means that the controller/operator need to create all the worker pod with `preemptable=true` annotation explicitly. it may be better for operator to create the master pod(just one) with `preemptable=false`.

k82cn: Went through this PR last night; I think it's better to ask controller/operator to set `preemptable` explicitly,as in controller/operator, it only few lines of code in PodTempalte with a new annotation :)

qiankunli: ok, I changed the code

k82cn:
/approve

Please help to update UT & E2E accordingly :)

pkg/scheduler/api/job_info.go
qiankunli: @k82cn but if we do this, it means that preempt/reclaim action will not work unless the operator being ready for the preemptable annotation. it may be a long time

k82cn: > @k82cn but if we do this, it means that preempt/reclaim action will not work unless the operator being ready for the preemptable annotation. it may be a long time

Different operators may have different expectation, it's better to have single behavior in Volcano; let's sync up with other community to align with each other :)

Issue #2112, opened 2022-03-24
Add `E2E Spark Integration Test`


Yikun:
**What happened**:
Add `E2E Spark Integration Test`

**What you expected to happen**:

Volcano support Spark on Kuberntes since Spark 3.3 and Volcano 1.5

**How to reproduce it (as minimally and precisely as possible)**:
CI triggered as expected

**Anything else we need to know?**:
Related: https://github.com/volcano-sh/volcano/issues/1704

PR #2113, opened 2022-03-24
Add `E2E Spark Integration Test`

Yikun:
**What happened**:
Add `E2E Spark Integration Test`

Volcano support Spark on Kuberntes since Spark 3.3 and Volcano 1.5

CI triggered as expected

Closes: https://github.com/volcano-sh/volcano/issues/2112

### DIFF
--- a/.github/workflows/e2e_spark.yaml
+++ b/.github/workflows/e2e_spark.yaml
@@ -1,0 +1,68 @@
+name: E2E Spark Integration Test
+
+on:
+  # Post check in below branches
+  push:
+    branches:
+      - master
+  # PR triggered job
+  pull_request:
+jobs:
+  k8s-integration-tests:
+    name: "E2E about Spark Integration test"
+    runs-on: ubuntu-20.04-spark
+    steps:
+
+    - name: Checkout current Volcano repository
+      if: github.event.inputs.volcano-branch==''
+      uses: actions/checkout@v2
+
+    - name: Checkout Spark repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        repository: apache/spark
+        ref: branch-3.3
+        path: ${{ github.workspace }}/spark
+    - name: Install Java 8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.x
+    - name: start minikube
+      run: |
+        # Use pre-install minikube
+        minikube start --cpus max --memory max
+    - name: Build lastest volcano images
+      run: |
+        eval $(minikube docker-env)
+        make TAG=latest update-development-yaml
+        make TAG=latest images
+        docker images | grep volcano
+        cat ./installer/volcano-development.yaml  | grep image:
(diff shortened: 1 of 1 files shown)


Yikun:
cc @Thor-wl @martin-g @william-wang

.github/workflows/e2e_spark.yaml
```diff
@@ -0,0 +1,68 @@
...
+      with:
+        java-version: 8
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.x
+    - name: start minikube
+      run: |
+        # Use pre-install minikube
+        minikube start
```
hwdef, line 38: Can we us `kind`? In my impression, volcano uses `kind` to test more often.

Yikun: In theory, kind and minikube have similar functions. Spark are using `minikube`, there are also some related advanced test depends on `minikube` (such as PV test), so it's better to keep same to use minikube if no more objection, to ensure we can enable more `minikube` related spark integration test in future.

.github/workflows/e2e_spark.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - name: Build lastest volcano images
+      run: |
+        eval $(minikube docker-env)
+        make TAG=latest update-development-yaml
+        make TAG=latest images
+        docker images | grep volcano
+        cat ./installer/volcano-development.yaml  | grep image:
+        kubectl apply -f ./installer/volcano-development.yaml
```
william-wang, line 46: Is thre any more dependency package needed to be installed such as kubectl?

Yikun: `kubectl`/`minikube` had been pre installed in self hosted runner, we might consider install specific version minikube/kubectl in future.

Yikun: To reduce deps on `kubectl`, we can use `minikube kubectl` instead.

.github/workflows/e2e_spark.yaml
```diff
@@ -0,0 +1,68 @@
...
+        # Use minikube kubectl
+        minikube kubectl -- get pods -A
+        minikube kubectl -- get nodes -oyaml
+    - name: Run K8S integration test
+      run: |
+        eval $(minikube docker-env)
+        minikube kubectl -- create clusterrolebinding serviceaccounts-cluster-admin --clusterrole=cluster-admin --group=system:serviceaccounts || true
+        build/sbt -Pvolcano -Pkubernetes -Pkubernetes-integration-tests -Dtest.include.tags=volcano "kubernetes-integration-tests/test"
+      working-directory: ${{ github.workspace }}/spark
+    - name: Upload spark integration tests log files
```
Yikun, line 58: @Thor-wl @william-wang  Do we also need to collect the running log of volcano as a reference for the failure?

william-wang: @Yikun It's better to keep logs if possible:)

Yikun: sure, we can do it in followup.

william-wang (commented):
/lgtm

PR #2121, opened 2022-03-27
fix: node state will not be ready once it become unready(OutOfSync)

NSBlink:
Signed-off-by: Xinjian Yu <blink@blink.moe>
Fix: node state will not be ready again once it become unready because of out of sync.

An example: I have a node with some running pods which requested nvidia.com/gpu. If I delete the nvidia device plugin pod at this time, the allocatable of nvidia.com/gpu will be 0 for a short time, the node will be unready(reason: OutOfSync). the nvidia device plugin pod recreated, the allocatable of nvidia.com/gpu will be 8, but the node state will not be ready again.


k82cn:
Why current logic can not handle your case?

william-wang:
@NSBlink This case was ever handled in Volcano, the node state would get ready when it gets event when the allocatable of nvidia.com/gpu recovered. Could you provide more detail info like the reproduce step?

PR #2122, opened 2022-03-28
[code refine] if only one candidate node, no need do scoring for it

wpeng102:
In `allocate` action, if only get one  candidate node after PredicateNodes, no need to socreing for that node. scheduler can use this node derectly. 

Just like following kubernetest scheduler logic:
https://github.com/kubernetes/kubernetes/blob/2cb3c7f706dbf266820fbde2e1b23a320e5d3de7/pkg/scheduler/schedule_one.go#L338-L347
image

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -225,10 +225,8 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 			var node *api.NodeInfo
 			switch {
 			case len(candidateNodes) == 0: // If not candidate nodes for this task, skip it.
-				// If not candidate nodes for this task, skip it.
 				continue
 			case len(candidateNodes) == 1: // If only one node after predicate, just use it.
-				// When only one node after predicate, just use it.
 				node = candidateNodes[0]
 			case len(candidateNodes) > 1: // If more than one node after predicate, using "the best" one
 				nodeScores := util.PrioritizeNodes(task, candidateNodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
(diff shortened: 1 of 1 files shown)


k82cn:
LGTM overall; how about `switch` for this PR?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -222,16 +222,20 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
+			} else if len(candidateNodes) == 1 {
+				// When only one node after predicate, just use it.
+				node = candidateNodes[0]
+			} else {
+				nodeScores := util.PrioritizeNodes(task, candidateNodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
 
-			node := ssn.BestNodeFn(task, nodeScores)
-			if node == nil {
-				node = util.SelectBestNode(nodeScores)
+				node := ssn.BestNodeFn(task, nodeScores)
```
k82cn, line 235: it should be `=` instead of `:=`

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -222,16 +222,21 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 				}
 			}
 
-			// If not candidate nodes for this task, skip it.
-			if len(candidateNodes) == 0 {
+			var node *api.NodeInfo
+			switch {
+			case len(candidateNodes) == 0: // If not candidate nodes for this task, skip it.
+				// If not candidate nodes for this task, skip it.
```
k82cn, line 228: duplicated comments?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -222,16 +222,21 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
 				continue
-			}
-
-			nodeScores := util.PrioritizeNodes(task, candidateNodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
-
-			node := ssn.BestNodeFn(task, nodeScores)
-			if node == nil {
-				node = util.SelectBestNode(nodeScores)
+			case len(candidateNodes) == 1: // If only one node after predicate, just use it.
+				// When only one node after predicate, just use it.
```
k82cn, line 231: ditto?

PR #2078, opened 2022-03-12
fix the driver pod can not be created due to unreasonable admit

william-wang:
Resolve the first item of #2040.  Also for #1778
Signed-off-by: william-wang <wang.platform@gmail.com>

PR #2124, opened 2022-03-28
Yolgun/fix admit pod

yolgun:
This PR reverts https://github.com/volcano-sh/volcano/pull/2078

We use Jobs and StatefulSets with podgroups. Creating pending pods cause issues.


yolgun:
Sorry, opened in wrong repo

PR #2132, opened 2022-03-30
fix signal on operating system expcept windows

xiaoanyunfei:
the code can not be compiled on mac


william-wang:
@xiaoanyunfei please make the ci happy.

Issue #2127, opened 2022-03-29
allocate action UT failed for some race condition


wpeng102:
**What happened**:
`go test  -count=1 -v -race volcano.sh/volcano/pkg/scheduler/actions/allocate` to run UT mabe failed by race condition

**What you expected to happen**:
UT pass

For the `pvc with matched pv` case, it has two pods and one pv, expect the first pod will use `StaticBindings` volume, the second pod use `DynamicProvisions` volume. 

But their is a race condition between client create pv and pvinformer in scheduler cache get the pv.
https://github.com/volcano-sh/volcano/blob/19498030cff7b9ff195c12c31d88761049c0e3ea/pkg/scheduler/actions/allocate/allocate_test.go#L446-L448

https://github.com/volcano-sh/volcano/blob/19498030cff7b9ff195c12c31d88761049c0e3ea/pkg/scheduler/actions/allocate/allocate_test.go#L476

When scheduler allocate the first pod, if pvinformer in scheduler cache does not aware the pv it will meet the following error:

![image](https://user-images.githubusercontent.com/10152842/160581426-784a2d7d-a71b-4370-b170-aeaa5c219a88.png)
![image](https://user-images.githubusercontent.com/10152842/160588123-1fd2c772-8fb8-45e3-a0d5-c53a647c3dbd.png)

PR #2133, opened 2022-03-30
fix dynamic provision ut case error

wpeng102:
Fix Issue: https://github.com/volcano-sh/volcano/issues/2127

**What happened**:
When run `go test -count=1 -v -race volcano.sh/volcano/pkg/scheduler/actions/allocate` unit test, you may get the following error results:
```
allocate_test.go:484: expected: map[c1/p3:[GetPodVolumes AllocateVolumes StaticBindings] c1/p4:[GetPodVolumes AllocateVolumes DynamicProvisions]], got map[c1/p3:[GetPodVolumes AllocateVolumes DynamicProvisions] c1/p4:[GetPodVolumes AllocateVolumes DynamicProvisions]] 
```
```
allocate_test.go:484: expected: map[c1/p3:[GetPodVolumes AllocateVolumes StaticBindings] c1/p4:[GetPodVolumes AllocateVolumes DynamicProvisions]], got map[c1/p3:[GetPodVolumes AllocateVolumes DynamicProvisions] c1/p4:[GetPodVolumes AllocateVolumes StaticBindings]] 
```
This happens occasionally.

**Root Cause**:
In the `TestAllocateWithDynamicPVC` unit test, it create a fake `kubeClient` at the beginning and use this client to create a fake volumebinder.
https://github.com/volcano-sh/volcano/blob/19498030cff7b9ff195c12c31d88761049c0e3ea/pkg/scheduler/actions/allocate/allocate_test.go#L356

In this volumebinder, it use `pvinformer` to maintains a cache.  

https://github.com/volcano-sh/volcano/blob/19498030cff7b9ff195c12c31d88761049c0e3ea/vendor/k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumebinding/assume_cache.go#L132-L150

When use fake `kubeclient` to create a pv, there is a time gap when the volumebinder get the created pv. 
https://github.com/volcano-sh/volcano/blob/19498030cff7b9ff195c12c31d88761049c0e3ea/pkg/scheduler/actions/allocate/allocate_test.go#L477-L479

So, the volumebinder has not get the created pv from informer, allocate the first pod will use `DynamicProvisions` for pod volumes, which is not match the expect result.

**Solution:**
1) For each unit test case, it should define its own pv/pvc/pods/node etc.
2) For each unit test case, it use own fake `kubeclient` and create its own resources before `fakeVolumeBinder := util.NewFakeVolumeBinder(kubeClient)`. Because this code will tigger cache sync.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -31,6 +31,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
 	"volcano.sh/volcano/pkg/scheduler/plugins/priority"
 
+	storagev1 "k8s.io/api/storage/v1"
 	schedulingv1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/scheduler/api"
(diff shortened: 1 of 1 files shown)


k82cn:
Great catch!

/lgtm
/approve

Issue #2044, opened 2022-03-01
PodGroup may lost if you delete a pending vcjob and then recreate it immediately.


Megan96:
**What happened**:
If you delete a pending vcjob and then recreate it immediately,  the podgroup of vcjob may lost.

**What you expected to happen**:
Recreate a new vcjob and a new podgrop.

**How to reproduce it (as minimally and precisely as possible)**:
Recreate immediately after deleting the pending vcjob, repeat the above behavior twice, there will always be one time to lose podgroup.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: 1.2.0
- Kubernetes version (use `kubectl version`): 1.20.15 / 1.18.0 
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shinytang6:
Thanks for your report. l am able to reproduce this issue, let me take a look )

Megan96:
您的邮件我已经收到，我会尽快查阅的，谢谢！

tFable:
I wanted to +1 this as I'm also seeing it. re-creating the job with a different name is the only way to get the podgroup to be created.

tFable:
Maybe this is helpful. I noticed that when I ran into this and I was creating jobs but no podgroups were created, the admission-controller was throwing the following error over and over again until the pod was deleted (and was automatically re-created)

```
I0303 18:19:12.594965       1 job_controller.go:324] Execute <SyncJob> on Job <kweiler/kweiler-test-job1d> in <Pending> by <*state.pendingState>.
I0303 18:19:12.594971       1 job_controller_actions.go:223] Starting to sync up Job <kweiler/kweiler-test-job1d>, current version 0
I0303 18:19:12.595001       1 job_controller_actions.go:164] Starting to initiate Job <kweiler/kweiler-test-job1d>
W0303 18:19:12.595011       1 job_controller_actions.go:744] Ignore task test-head priority class : priorityclass.scheduling.k8s.io "" not found
W0303 18:19:12.595018       1 job_controller_actions.go:744] Ignore task test-workers priority class : priorityclass.scheduling.k8s.io "" not found
I0303 18:19:12.599644       1 queue_controller_action.go:81] End sync queue q1.
I0303 18:19:12.599660       1 queue_controller.go:220] Finished syncing queue q1 (4.735513ms).
[...]

PR #2134, opened 2022-03-31
fix: add jobUID into job's podgroup name ensure podgroup's unique, solve issue #2044

FengXingYuXin:
podgroup may carete failed when a job with same jobName exec a series of operations, such as  create/delete/create, you can see it in #2044


k82cn:
please help to resolve conflicts and make CI happy :)

PR #2139, opened 2022-03-31
vgpu support for volcano

archlitchi:
resubmit this pr to merge into master branch of volcano

vgpu device plugin repo:  https://github.com/archlitchi/devices

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ manifests: controller-gen
 
 unit-test:
 	go clean -testcache
-	go test -p 8 -race $$(find pkg -type f -name '*_test.go' | sed -r 's|/[^/]+$$||' | sort | uniq | sed "s|^|volcano.sh/volcano/|")
+	go test -p 8 -race $$(find pkg cmd -type f -name '*_test.go' | sed -r 's|/[^/]+$$||' | sort | uniq | sed "s|^|volcano.sh/volcano/|")
 
 e2e: images
 	./hack/run-e2e-kind.sh
--- a/cmd/controller-manager/app/options/options_test.go
+++ b/cmd/controller-manager/app/options/options_test.go
@@ -46,11 +46,12 @@ func TestAddFlags(t *testing.T) {
 			QPS:        defaultQPS,
 			Burst:      200,
 		},
-		PrintVersion:       false,
-		WorkerThreads:      defaultWorkers,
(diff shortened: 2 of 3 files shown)


k82cn:
let's seperate design doc to stand-alone PR :)

k82cn:
And please seperate this feature into smaller PRs, it's hard for reviewer to review such a huge commits :)

installer/dockerfile/scheduler/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 
-FROM alpine:latest
+FROM ubuntu:18.04
```
jinzhejz, line 16: Is there any background for using `ubuntu` instead of `alpine` here?

installer/volcano-development.yaml
```diff
@@ -143,7 +143,7 @@ spec:
             - --port=8443
             - -v=4
             - 2>&1
-          image: volcanosh/vc-webhook-manager:latest
+          image: volcanosh/vc-webhook-manager:289608dd88a1e43ac93e29f900313503e9b14f62
```
jinzhejz, line 146: I don't think we should use a specific version here

pkg/scheduler/scheduler.go
```diff
@@ -170,3 +189,131 @@ func (pc *Scheduler) watchSchedulerConf(stopCh <-chan struct{}) {
...
+			_ = stream.SendAndClose(&vgpuproto.RegisterReply{})
+			return err
+		}
+		klog.V(3).Infof("device register %v", req.String())
+		nodeID = req.GetNode()
+		nodes := pc.cache.(*schedcache.SchedulerCache).Nodes
+		nodeinfo, ok := nodes[nodeID]
+		if !ok {
+			nodes[nodeID] = api.NewNodeInfo(nil)
+		}
```
jinzhejz, line 257: Adding an item to map here may conflicts with `SchedulerCache.addNode` in scheduler cache. How about start vgpu server and update VGPUDevices in scheduler cache.

jinzhejz:
I thought about this pr again and it may be better to create a new plugin to support vGPU. 

Here is rough idea about this plugin:
* start vGPUServer once
* maintain a local cache of Pod/Node/vGPUDeviceInfo, 
* add a predicate function to filter vGPU nodes for Pod
* add allocate/deallocate function to update vGPU info to local cache

PR #2140, opened 2022-03-31
fix: add jobUID into job's podgroup name ensure podgroup's unique

FengXingYuXin:
podgroup may carete failed when a job with same jobName exec a series of operations, such as create/delete/create, you can see it in https://github.com/volcano-sh/volcano/issues/2044


shinytang6 (approved):
/approve

pkg/controllers/job/job_controller_actions.go
```diff
@@ -147,7 +147,8 @@ func (cc *jobcontroller) killJob(jobInfo *apis.JobInfo, podRetainPhase state.Pha
 	}
 
 	// Delete PodGroup
-	if err := cc.vcClient.SchedulingV1beta1().PodGroups(job.Namespace).Delete(context.TODO(), job.Name, metav1.DeleteOptions{}); err != nil {
+	pgName := job.Name + "-" + string(job.UID)
```
hzxuzhonghu, line 150: This is not safe, the length could exceed k8s limit

FengXingYuXin: yep, thanks for your remind, so if I change the prefix into a constant such as 
1) "pod-group-"(it's from volcano-sh/apis's PodgroupNamePrefix), if uses it, the prefix unify.
2) or "job-pod-group-", add one new constant prefix into apis,  if uses it, we can distinguish job's podgroup from pod's podgroup easily.
so, which one do you think better?

william-wang: It is convinient that the podgroup name contains the job.Name. It is easy to know which job the podgroup belongs to without checking the ownereference.  From my point of view, job.Name + string(job.UID) is a better option. Besides, I am thinking whether the podgroup uid could use the Object ID generator util in #2145.

FengXingYuXin: i agree with your point about the podgroup name format, however, about the uid generator, i think it doesnot make much difference.

k82cn: similar to the pod name of Deployments :)

william-wang:
Please fix the e2e failures :)

FengXingYuXin:
> > 
> 
> The reason of E2E failed is :
> 
> ```shell
> kubectl logs -n volcano-system volcano-controllers-xxxxxxxx-xxxxx
> 
> ## admission webhook \"validatepod.volcano.sh\" denied the request: failed to get PodGroup for pod <qeyeu1fm/scale-up-job-default-1>: podgroups.scheduling.volcano.sh \"scale-up-job\" not found
> ```
> 
> The `validatePod` func will get PG from annotation field. See
> 
> https://github.com/volcano-sh/volcano/blob/fac7d5f05bfe2e8da1d9d5a5ce9275418ec626a4/pkg/webhooks/admission/pods/validate/admit_pod.go#L112
> 
> So here, Add UID in the annotation field : https://github.com/FengXingYuXin/volcano/blob/master/pkg/controllers/job/job_controller_util.go#L103:
> 
> ```go
> pod.Annotations[schedulingv2.KubeGroupNameAnnotationKey] = job.Name     // UID should also be added here
> ```
> 
> Also, add UID in job.Name in E2E code:
> 
> https://github.com/FengXingYuXin/volcano/blob/master/test/e2e/util/job.go#L347 https://github.com/FengXingYuXin/volcano/blob/master/test/e2e/util/job.go#L381 https://github.com/FengXingYuXin/volcano/blob/master/test/e2e/util/job.go#L651

thanks for your pointing out about the problems in details very much

FengXingYuXin:
> Please fix the e2e failures :)

ok, it has been fixed

william-wang (approved):
/lgtm

Issue #2075, opened 2022-03-11
Add cooldown time support for preempt action


flyhighzy:
#### What would you like to be added:
user can set cool down time for preemptible job's pod by add some labels or annotations, to avoid some pods being preempted when they just started for a short time.

#### Why is this needed:
related to [elastic scheduler](https://github.com/volcano-sh/volcano/blob/master/docs/design/elastic-scheduler.md), when we need to enable elastic training or serving, preemptible job's pods can be preempted or back to running repeatedly, if no cool down time set, these pods can be preempted again after they just started for a short time, this may cause service stability dropped.

Thor-wl:
Good idea. I think that is meaningful. Can you help give a design and implementation for that?

flyhighzy:
we already have a localized implementation, it's not complex so I simply summary it here:
1. provide a new label/annotation named "volcano.sh/preempt_stable_time", whose value means the cool down time for preempt with unit second. This label/annotation can be set for entire vcjob or some dedicated tasks, if set to job, we'll transfer to all tasks' pods.
2. add a plugin to participate in preempt action, ensure pods whose scheduled time after `now - preempt_stable_time` will be not in the result victims list

@Thor-wl  Please have a review, if ok, I can submit a pr

k82cn:
> provide a new label/annotation named "volcano.sh/preempt_stable_time", whose value means the cool down time for preempt with unit second. 

This should be handled within scheduler's cache to avoid additional apiserver request :)

k82cn:
IMO, it's better to have a plugin to order victims by start time; and have a min-start-time parameter for that :)

flyhighzy:
> This should be handled within scheduler's cache to avoid additional apiserver request :)

Thanks for the reminding! I'll take care of it.

> IMO, it's better to have a plugin to order victims by start time; and have a min-start-time parameter for that :)

I'm not very sure about what can we do with the ordered victims, can you provide more details? thanks :）

Issue #68, opened 2019-04-04
How do we support Job.Spec update


hzxuzhonghu:
I noticed we want to support Job.Spec update in `Controller.updateJob`

But the generated request is 
```
	req := apis.Request{
		Namespace: newJob.Namespace,
		JobName:   newJob.Name,

		Event: vkbatchv1.OutOfSyncEvent,
	}
```
But in syncJob
if no pods provided in request, it will create new pods for the Job, and so it will fail, and the following status is unknown.

btw, I am not very familiar with the entire state machine , and maybe i miss something.

PR #2149, opened 2022-04-06
add cooldown protection plugin

flyhighzy:
resolve issue [#2075](https://github.com/volcano-sh/volcano/issues/2075)

Add cooldown time support for preempt action:
1. provide a new label/annotation named "volcano.sh/preempt-stable-time", whose value means the cool down time for preempt with unit second. This label/annotation can be set for entire vcjob or some dedicated tasks, if set to job, we'll transfer to all tasks' pods.
2. add a plugin to participate in preempt action, ensure pods whose scheduled time after `now - preempt-stable-time` will be not in the result victims list

### DIFF
--- a/docs/user-guide/how_to_use_cdp_plugin.md
+++ b/docs/user-guide/how_to_use_cdp_plugin.md
@@ -1,0 +1,201 @@
+# Cooldown Protection Plugin User Guide
+
+## Background
+When we need to enable elastic training or serving, preemptible job's pods can be preempted or back to running repeatedly, if no cooldown protection set, these pods can be preempted again after they just started for a short time, this may cause service stability dropped.
+So we add "cdp" plugin to ensure preemptible job's pods can run for at least some time set by user.
+
+## Environment setup
+
+### Install volcano
+
+Refer to [Install Guide](../../installer/README.md) to install volcano.
+
+### Update scheduler configmap
+
+After installed, update the scheduler configuration:
+
+```shell
+kubectl edit configmap -n volcano-system volcano-scheduler-configmap
+```
+
+Register `cdp` plugin in configmap while enable `preempt` action
+
+```yaml
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: volcano-scheduler-configmap
+  namespace: volcano-system
+data:
+  volcano-scheduler.conf: |
+    actions: "enqueue, allocate, preempt, backfill"
+    tiers:
+    - plugins:
+      - name: priority
+      - name: gang
+      - name: conformance
+      - name: cdp
+    - plugins:
+      - name: drf
+      - name: predicates
+      - name: task-topology
+        arguments:
+          task-topology.weight: 10
+      - name: proportion
+      - name: nodeorder
+      - name: binpack
+```
+
+### Running Jobs
+
+Take a simple volcano job as sample.
+
+original job yaml is as below, which has "ps" and "worker" task
+
+```yaml
+apiVersion: batch.volcano.sh/v1alpha1
+kind: Job
+metadata:
+  name: test-job
+spec:
+  minAvailable: 3
+  schedulerName: volcano
+  priorityClassName: high-priority
+  plugins:
+    ssh: []
+    env: []
+    svc: []
+  maxRetry: 5
+  queue: default
+  volumes:
+    - mountPath: "/myinput"
+    - mountPath: "/myoutput"
+      volumeClaimName: "testvolumeclaimname"
+      volumeClaim:
+        accessModes: [ "ReadWriteOnce" ]
+        storageClassName: "my-storage-class"
+        resources:
+          requests:
+            storage: 1Gi
+  tasks:
+    - replicas: 6
+      name: "worker"
+      template:
+        metadata:
+          name: worker
+        spec:
+          containers:
+            - image: nginx
+              imagePullPolicy: IfNotPresent
+              name: nginx
+              resources:
+                requests:
+                  cpu: "1"
+          restartPolicy: OnFailure
+    - replicas: 2
+      name: "ps"
+      template:
+        metadata:
+          name: ps
+        spec:
+          containers:
+            - image: nginx
+              imagePullPolicy: IfNotPresent
+              name: nginx
+              resources:
+                requests:
+                  cpu: "1"
+          restartPolicy: OnFailure
+
+```
+
+#### Edit yaml of vcjob
+
+1. add annotations in volcano job in format below.
+   1. `volcano.sh/preemptable` annotation indicates that job or task is preemptable
+   2. `volcano.sh/cooldown-time` annotation indicates cooldown time for the entire job or dedicated task. Value for the annotation indicates cooldown time, valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". 
+
+        ```yaml
+            volcano.sh/preemptable: "true"
+            volcano.sh/cooldown-time: "600s"
+        ```
+
+**Example 1**
+
+Add annotation to entire job, then "ps" and "worker" task can be preempted and all have cooldown time support.
+
+```yaml
+apiVersion: batch.volcano.sh/v1alpha1
+kind: Job
+metadata:
+  name: test-job
+  annotations:
+    volcano.sh/preemptable: "true"
+    volcano.sh/cooldown-time: "600s"
+spec:
+  ... # below keep the same
+```
+
+**Example 2**
+
+Add annotation to dedicated task, as shown below, only "worker" can be preempted and have cooldown time support.
+
+```yaml
+apiVersion: batch.volcano.sh/v1alpha1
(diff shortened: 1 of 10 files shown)


flyhighzy:
add volcano.sh/api pr [#68](https://github.com/volcano-sh/apis/pull/68), please have a review

Thor-wl:
Please update the vendor locally.

flyhighzy:
thanks for the advice and please have a review :)
since this pr depends on [pr of. volcano.sh/api](https://github.com/volcano-sh/apis/pull/68), please also review api's pull request as well

shinytang6 (commented):
Generally lgtm, can we have a simple doc to claim how to use this plugin?
btw, it seems that this pr contains some api changes that causing ci failed )

shinytang6:
pls help resolve the conflicts.

flyhighzy:
> pls help resolve the conflicts.

updated

shinytang6 (commented):
LGTM overall.
ping @Thor-wl @k82cn @william-wang, pls take a review :)

flyhighzy:
new conflict resolved
 @Thor-wl @k82cn  @william-wang , pls take a review

flyhighzy:
all workflow succeeded except code review now.
Please take a review~  @william-wang @wpeng102 @k82cn

flyhighzy:
ping @william-wang  @shinytang6 @qiankunli :)

william-wang:
@flyhighzy thanks for your contribution. I will review it as soon as possible.

docs/user-guide/how_to_use_stable_preempt_plugin.md
```diff
@@ -0,0 +1,201 @@
...
+                requests:
+                  cpu: "1"
+          restartPolicy: OnFailure
+
+```
+
+#### Edit yaml of vcjob
+
+1. add annotations in volcano job in format below.
+   1. `preemptable` annotation(or label) indicates that job or task is preemptable
```
Thor-wl, line 115: IMO, annotation is better for there exists preemption annotaion `volcano.sh/preemptable: "true"`.

Thor-wl (commented):
Thanks for your contribution! It's generally LGTM for me now.
/lgtm

docs/user-guide/how_to_use_stable_preempt_plugin.md
```diff
@@ -0,0 +1,201 @@
...
+  namespace: volcano-system
+data:
+  volcano-scheduler.conf: |
+    actions: "enqueue, allocate, preempt, backfill"
+    tiers:
+    - plugins:
+      - name: priority
+      - name: gang
+      - name: conformance
+      - name: stablepreempt
```
william-wang, line 37: I strongly suggest change another plugin name instead of `stablepreempt`. 
It seems current preempt is not stable without this plugin:)

docs/user-guide/how_to_use_stable_preempt_plugin.md
```diff
@@ -0,0 +1,201 @@
...
+
+#### Edit yaml of vcjob
+
+1. add annotations in volcano job in format below.
+   1. `preemptable` annotation(or label) indicates that job or task is preemptable
+   2. `preempt-stable-time` annotation(or label) indicates cool down time for entire job or dedicated task. Value for the annotation indicates stable time, valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". 
+
+        ```yaml
+            volcano.sh/preemptable: "true"
+            volcano.sh/preempt-stable-time: "600s"
```
william-wang, line 120: As you described in the background, i think `volcano.sh/cooldown-time` is more strightforward.

pkg/scheduler/plugins/stablepreempt/stable_preempt.go
```diff
@@ -0,0 +1,105 @@
...
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/plugins/util"
+)
+
+const (
+	// PluginName is name of plugin
+	PluginName = "stablepreempt"
+)
+
```
william-wang, line 35: Suggest add a comment here to describe this new plugin.

docs/user-guide/how_to_use_stable_preempt_plugin.md
flyhighzy: How about "preempt-protected"?
or any other suggestions for the new plugin name?

docs/user-guide/how_to_use_stable_preempt_plugin.md
flyhighzy: Job label changes already discussed in [pr#68](https://github.com/volcano-sh/apis/pull/68),  so we still need to change this label name again?

docs/user-guide/how_to_use_stable_preempt_plugin.md
william-wang: Let's name it `cdt` plugin which means cool down time or `cdp` which means cool down period.  It can be used not only in preemption but also in reclaim as well.

docs/user-guide/how_to_use_stable_preempt_plugin.md
william-wang: @flyhighzy No worry, just change it and let's do things right the first time :)

Thor-wl (commented):
I'm generally OK about the modifications. Please rebase all the commits into one and push again.

docs/user-guide/how_to_use_stable_preempt_plugin.md
Jason-Liu-Dream: @william-wang @flyhighzy I don't recommend using this name. There may be multiple ways to cool down victim tasks, not only by time interval. The name limits the functionality of the plugin and makes the functionality of this plugin slightly thin. How about `vcp` which means victim cooldown protection or `ecp` which means evicted cooldown protection.

flyhighzy: @Jason-Liu-Dream I noticed your [issue](https://github.com/Jason-Liu-Dream/volcano/issues/7), but maybe you need to put issue in the right repo :).   I think it sounds reasonable, but may need to discuss more details.

So what's your opinions? @william-wang @Thor-wl

william-wang: @Jason-Liu-Dream Would you clariy some other vailid cool down ways besides  the time/period?

Jason-Liu-Dream: > @Jason-Liu-Dream I noticed your [issue](https://github.com/Jason-Liu-Dream/volcano/issues/7), but maybe you need to put issue in the right repo :). I think it sounds reasonable, but may need to discuss more details.
> 
this issue is put here because our inner version implements a demo of this function, we want to combine your design, and retry to design our implements.

Jason-Liu-Dream: > @Jason-Liu-Dream Would you clariy some other vailid cool down ways besides the time/period?

We think cool down reason maybe need to consider time periods， eviction times， pod height load pressure,  block messages, message events and so on and different business scenarios have different requirements.

william-wang: @Jason-Liu-Dream Thanks for the explaining the different cases for the cool down operation.  If possible you can elaborate on pod height load pressure, block messages, message events. 
 
@flyhighzy I think we can refine the name to `clp` which means cool down protection. It can cover most of use cases so far :)

flyhighzy: sure, I think so :)

flyhighzy:
I've finished rebase all the commits into one and rename plugin name to "cdp", pls take a review again~
@Thor-wl @william-wang

william-wang (approved):
Overall, looks good to me.
/approve

docs/user-guide/how_to_use_stable_preempt_plugin.md
Jason-Liu-Dream: > @Jason-Liu-Dream Thanks for the explaining the different cases for the cool down operation. If possible you can elaborate on pod height load pressure, block messages, message events.
>  @william-wang 
pod height load pressure: Sometimes we need to preempt some pods, but we don't want to preempt it when this pod is at height load pressure, height load pressure means a large scope of influence after stopping service in this pod. And we don't know how long this will last. so cooldown time does not work.
block messages: similar to above, sometimes message block in queue means there is a lot of stress to deal with in this service. Preempting is not a good idea.
message events: In batch tasks, we can record progress. Assuming that a task has N stages, and each stage supports breakpoints to continue working, then the best plan is to complete the goal of the previous stage, send a message event, and then let this task be preempted so that not cause excessive waste of resources. This is very common in pipeline jobs such as video transcoding.

PR #2157, opened 2022-04-08
update readme that Volcano has been accepted as an incubating project

Thor-wl:


hwdef (commented):
/lgtm
Congratulation!

PR #2163, opened 2022-04-13
fix: Add mirror for Spark voclano IT

Yikun:
Switch 3rd party debian mirror to make spark IT ci passed.

Closes: #2165


Yikun:
spark CI passed: https://github.com/volcano-sh/volcano/runs/6003764089?check_suite_focus=true

cc @Thor-wl @william-wang

Thor-wl (approved):

Issue #2160, opened 2022-04-12
vcjob phase can get stuck in "Pending" or "Starting" even after all pods have completed


hapatrick:
**What happened**:
In our system, we have repeatedly seen vcjobs get stuck in phase “Pending” or “Starting” and remain there even after all associated pods have completed successfully. From the Volcano controller logs, it seems Volcano fails to patch the vcjob’s status because the vcjob resource has unexpectedly been changed. In our system, these changes are being caused by an operator we wrote to monitor the status of vcjobs. Our operator uses the kopf framework (https://github.com/nolar/kopf). Kopf uses annotations on the Kubernetes resources it monitors to help it track changes to resources. 

Restarting the Volcano controller causes the phase of stuck vcjobs to be updated.

Sample of the Volcano controller log when this error occurs:
```
Starting to initiate Job <app/workflow-njd4q-job0>
W0406 18:52:54.638604       1 job_controller_actions.go:652] Ignore task 0 priority class : priorityclass.scheduling.k8s.io "" not found
I0406 18:52:54.638852       1 event.go:282] Event(v1.ObjectReference{Kind:"Job", Namespace:"app", Name:"workflow-njd4q-job0", UID:"aec8c14c-64e2-44a3-a928-e62217cafba5", APIVersion:"batch.volcano.sh/v1alpha1", ResourceVersion:"45929598", FieldPath:""}): type: 'Warning' reason: 'PodGroupPending' PodGroup app:workflow-njd4q-job0 unschedule,reason: 1/0 tasks in gang unschedulable: pod group is not ready, 1 minAvailable
E0406 18:52:54.666228       1 job_controller_actions.go:434] Failed to update status of Job app/workflow-njd4q-job0: Operation cannot be fulfilled on jobs.batch.volcano.sh "workflow-njd4q-job0": the object has been modified; please apply your changes to the latest version and try again
I0406 18:52:54.666251       1 job_controller_actions.go:436] Finished Job <app/workflow-njd4q-job0> sync up, current version 0
W0406 18:52:54.666279       1 job_controller.go:342] Terminating Job <app/workflow-njd4q-job0> and releasing resources
I0406 18:52:54.666289       1 job_controller_actions.go:45] Killing Job <app/workflow-njd4q-job0>, current version 0
I0406 18:52:54.666330       1 job_controller_actions.go:113] Running duration is 5m16.666322492s
[...]

PR #2169, opened 2022-04-14
fix controller job cache not sync latest version issue

wpeng102:

Resolve issue:  https://github.com/volcano-sh/volcano/issues/2160

#### What happened:
vcjob phase can get stuck in "Pending" or "Starting" even after all pods have completed.
Detail refer https://github.com/volcano-sh/volcano/issues/2160

#### Root Cause:
When controller update job status, `Kopf` add annotation for this job.  

Consider this situation: 
https://github.com/volcano-sh/volcano/blob/ccda7582e7df156ceddebeeb07fd83aacdf7d04a/pkg/controllers/job/job_controller_actions.go#L462-L468

https://github.com/volcano-sh/volcano/blob/ccda7582e7df156ceddebeeb07fd83aacdf7d04a/pkg/controllers/job/job_controller_handler.go#L93

thread1 at L462, get the job with `version_1`.
thread2 at L93, get `Kopf` add annotation event and put `version_2` job into cache.
thread1 at L468, put `version_1` job into cache again. 

When next call `UpdateStatus` will use the old `version_1` job, then meet [2160](https://github.com/volcano-sh/volcano/issues/2160) issue.

#### Manual  reproduce steps:
1) add `time.Sleep(time.Second*20)`  before https://github.com/volcano-sh/volcano/blob/ccda7582e7df156ceddebeeb07fd83aacdf7d04a/pkg/controllers/job/job_controller_actions.go#L468
2) submit vcjob
3) during controller sleep, add annotation for this job manually.

#### Solution:
In `cache.update` need check if the job resourceversion is bigger than the old one.

After this fix, when trigger the above case the controller log will print:
```
E0414 08:12:30.419699       1 job_controller_actions.go:307] SyncJob - Failed to update Job default/job-38 in cache:  job <default/job-38> has too old resource version: 6358346 (6358347)
```
Then controller will requeue this request.

### DIFF
--- a/pkg/controllers/cache/cache.go
+++ b/pkg/controllers/cache/cache.go
@@ -18,6 +18,7 @@ package cache
 
 import (
 	"fmt"
+	"strconv"
 	"sync"
 	"time"
 
@@ -152,8 +153,20 @@ func (jc *jobCache) Update(obj *v1alpha1.Job) error {
 	if !found {
 		return fmt.Errorf("failed to find job <%v>", key)
 	}
-	job.Job = obj
 
+	var oldResourceversion, newResourceversion uint64
+	var err error
+	if oldResourceversion, err = strconv.ParseUint(job.Job.ResourceVersion, 10, 64); err != nil {
+		return fmt.Errorf("failed to parase job <%v> resource version <%s>", key, job.Job.ResourceVersion)
+	}
+
+	if newResourceversion, err = strconv.ParseUint(obj.ResourceVersion, 10, 64); err != nil {
+		return fmt.Errorf("failed to parase job <%v> resource version <%s>", key, obj.ResourceVersion)
+	}
+	if newResourceversion < oldResourceversion {
+		return fmt.Errorf("job <%v> has too old resource version: %d (%d)", key, newResourceversion, oldResourceversion)
+	}
+	job.Job = obj
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/cache/cache.go
```diff
@@ -152,8 +153,20 @@ func (jc *jobCache) Update(obj *v1alpha1.Job) error {
...
+	var err error
+	if oldResourceversion, err = strconv.ParseUint(job.Job.ResourceVersion, 10, 64); err != nil {
+		return fmt.Errorf("failed to parase job <%v> resource version <%s>", key, job.Job.ResourceVersion)
+	}
+
+	if newResourceversion, err = strconv.ParseUint(obj.ResourceVersion, 10, 64); err != nil {
+		return fmt.Errorf("failed to parase job <%v> resource version <%s>", key, obj.ResourceVersion)
+	}
+	if newResourceversion < oldResourceversion {
+		return fmt.Errorf("job <%v> has too old resource version: %d (%d)", key, newResourceversion, oldResourceversion)
```
william-wang, line 167: If the newResourceversion is less than odResourceversion, the cache keep the same as it was, giving a warning log and returning nil seems more reasonable. 
fox example, in the killJob func, if cc.cache.Update(newJob) return error in this case, the podgroup will not be cleaned in system.

wpeng102: If just log a warning log and return nil will not trigger request rerun, Actually,  `processNextReq` in `job_controller` will check `st.Execute(action)` return result, if has error, it will re-queue this request and next will get the latest job from cache.
killJob func also follow the above rule.

PR #2175, opened 2022-04-19
add additionalPrinterColumns

qiankunli:
add additionalPrinterColumns field in crd yaml file so that we can query vcjob and podgroup like:

77E8F0A8-EC37-41BA-859F-0C68FAC1587D

### DIFF
--- a/installer/volcano-development-arm64.yaml
+++ b/installer/volcano-development-arm64.yaml
@@ -204,7 +204,20 @@ spec:
     singular: job
   scope: Namespaced
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
+    - jsonPath: .status.state.phase
+      name: STATUS
+      type: string
+    - jsonPath: .status.minAvailable
+      name: minAvailable
+      type: integer
+    - jsonPath: .status.running
+      name: RUNNINGS
+      type: integer
+    - jsonPath: .metadata.creationTimestamp
+      name: AGE
+      type: date
+    name: v1alpha1
     schema:
       openAPIV3Schema:
         description: Job defines the volcano job.
@@ -8886,6 +8899,10 @@ spec:
           status:
             description: The status of queue.
             properties:
+              completed:
+                description: The number of `Completed` PodGroup in this queue.
+                format: int32
+                type: integer
               inqueue:
                 description: The number of `Inqueue` PodGroup in this queue.
                 format: int32


hwdef (commented):
I feel that there is some problem with the generated yaml file, there is no arm64 version

Thor-wl:
> I feel that there is some problem with the generated yaml file, there is no arm64 version

Yes, can you help add that?

hwdef:
> > I feel that there is some problem with the generated yaml file, there is no arm64 version
> 
> Yes, can you help add that?

The latest code already can generate the arm64 yaml.
Just need to run `make update-development-yaml`

https://github.com/volcano-sh/volcano/blob/657a0b85aa8e0e8f3448c8d9b91ee56a3aee66c3/Makefile#L201

shinytang6 (commented):
l think it better to add these annotations to `Job` crd under volcano/apis and then generate this file automatically

qiankunli:
> l think it better to add some kubebuilder annotations to `Job` crd under volcano/apis and then generate this file automatically

good idea

hwdef:
Still no arm64 version

qiankunli:
> Still no arm64 version

@hwdef Does it matter a lot? can you create a pr about arm64 separately?

hwdef:
> > Still no arm64 version
> 
> @hwdef Does it matter a lot? can you create a pr about arm64 separately?

I think we should keep arm64 and amd64 with consistent functionality.
If you use the latest code and run `make update-development-yaml`, you will get the arm64 version yaml.

hwdef:
The ci error is also related to this

qiankunli:
> > > Still no arm64 version
> > 
> > 
> > @hwdef Does it matter a lot? can you create a pr about arm64 separately?
> 
> I think we should keep arm64 and amd64 with consistent functionality. If you use the latest code and run `make update-development-yaml`, you will get the arm64 version yaml.

I have updated volcano-development-arm64.yaml,  please help check it @hwdef

hwdef:
There is an arm64 version now, but the previous amd64 version seems to have disappeared? There should still be two yaml files.
You can try to deal with it again, if there is still a problem, we can merge this PR first, I will add the missing yaml later, because now the CI is correct.
What do you think about this?

qiankunli:
> There is an arm64 version now, but the previous amd64 version seems to have disappeared? There should still be two yaml files. You can try to deal with it again, if there is still a problem, we can merge this PR first, I will add the missing yaml later, because now the CI is correct. What do you think about this?

you may need to modify the Makefile @hwdef 

image

hwdef (commented):
/lgtm
Thanks!

PR #2176, opened 2022-04-19
fix task MinAvailable issue

merryzhou:
1. task MinAvailable should greater than 0 when create & update job
2. skip check when task MinAvailable == 0

Signed-off-by: merryzhou <merryzhou0882@gmail.com>

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -704,6 +704,9 @@ func (ji *JobInfo) CheckTaskMinAvailable() bool {
 
 	klog.V(4).Infof("job %s/%s actual: %+v, ji.TaskMinAvailable: %+v", ji.Name, ji.Namespace, actual, ji.TaskMinAvailable)
 	for task, minAvailable := range ji.TaskMinAvailable {
+		if minAvailable == 0 {
+			continue
+		}
 		if act, ok := actual[task]; !ok || act < minAvailable {
 			return false
 		}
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -704,6 +704,9 @@ func (ji *JobInfo) CheckTaskMinAvailable() bool {
 
 	klog.V(4).Infof("job %s/%s actual: %+v, ji.TaskMinAvailable: %+v", ji.Name, ji.Namespace, actual, ji.TaskMinAvailable)
 	for task, minAvailable := range ji.TaskMinAvailable {
+		if minAvailable <= 0 {
```
hwdef, line 707: Why not intercept invalid values ​​via webhook?

Thor-wl: Why can it be an invalid value? I agree with @hwdef that it should be checked in the webhook.

merryzhou: when task replica == 0，then minAvailable will be zero.

and is valid to set task replica to 0

Thor-wl: /cc @william-wang How do you think about it? Is there some scenarios that users may set `task.replicas` to be `0`?

william-wang: task.Replicas is allowed to be `0`.  Please do this kind of check in webhook.

merryzhou: done

PR #2184, opened 2022-04-21
add rescheduling plugin

Thor-wl:


william-wang (approved):
/lgtm

william-wang:
please test the plugin with promethus and make sure it works as expected.

Thor-wl:
> please test the plugin with promethus and make sure it works as expected.

Thanks. I'll do that recently.

Issue #2195, opened 2022-04-26
【UT】Case "pvc with matched pv" in `TestAllocatedWithDynamicPVC` happens to be get an expected result, which often blocks CI


Thor-wl:
**What happened**:
It often happens that UT case `pvc with matched pv` in `TestAllocatedWithDynamicPVC`, which belongs to `allocate` packge, gets an expected result. This often blocks CI and bothers PR submitter a lot. 
**What you expected to happen**:
This case can work well.
**How to reproduce it (as minimally and precisely as possible)**:
Just execute `make unit-test` under the project root directory and it can often be reproduced.
![截图](https://user-images.githubusercontent.com/61179420/165240998-ef46d4d9-bd7d-46fa-811f-79c277253471.PNG)

**Anything else we need to know?**:
None
**Environment**:
- Volcano Version: master branch
- Kubernetes version (use `kubectl version`): v1.23
- Cloud provider or hardware configuration: No limit
- OS (e.g. from /etc/os-release): No limit
- Kernel (e.g. `uname -a`): No limit
- Install tools: make
- Others: None

PR #2196, opened 2022-04-27
remove unresonable UT for allocate action

Thor-wl:
ref: https://github.com/volcano-sh/volcano/issues/2195
UT `pvc with matched pv` is not correct for that it is also possible to bind a pv dynamically for pod `p3`.


wpeng102 (approved):
/lgtm

PR #2198, opened 2022-04-27
fix scheduling failed when different role in tfjob has empty taskID

xiaoanyunfei:
Tfjob has different role with different NodeSelector and NodeAffinity.
such as  `PS` will choose nodes  with LabelA
`Worker` will choose nodes LabelB
`Chief` will choose nodes with LabelC

Current schedule rule will generate empty `TaskID` for tasks belonging to one job
It's ok for jobs with same NodeSelector rules, but not for  `ps`, `worker`, `chief` of tfjob with different NodeSelectors

/priority-high

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -147,7 +148,17 @@ func getTaskID(pod *v1.Pod) TaskID {
...
 	}
 
-	return ""
+	return TaskID(GetDefaultTaskSpec(pod))
+}
+
+// GetDefaultTaskSpec return the pod generate name without pod index
+// if pod name is "tfjob-worker-9", return "tfjob-worker".
+func GetDefaultTaskSpec(pod *v1.Pod) string {
+	specs := strings.Split(pod.Name, "-")
```
k82cn, line 157: We support different operators, e.g. customized operator, it's hard to say what will be return after this PR :)

PR #2199, opened 2022-04-27
add priorityClassName for rescheduler

jiangxiaobin96:
Signed-off-by: jiangxiaobin96 <401236678@qq.com>

Add priorityClassName for rescheduler to avoid evicting volcano pod due to rescheduler will identify critical pods.


jiangxiaobin96:
If it is necessary, I will update YAML for ARM.

PR #2202, opened 2022-04-28
support ephemeral-storage

xiaoanyunfei:
volcano don not support `ephemeral-storage` for now


pkg/scheduler/api/resource_info.go
```diff
@@ -77,9 +76,7 @@ func NewResource(rl v1.ResourceList) *Resource {
 			r.MaxTaskNum += int(rQuant.Value())
 		default:
 			//NOTE: When converting this back to k8s resource, we need record the format as well as / 1000
-			if v1helper.IsScalarResourceName(rName) {
-				r.AddScalar(rName, float64(rQuant.MilliValue()))
-			}
+			r.AddScalar(rName, float64(rQuant.MilliValue()))
```
Thor-wl, line 79: Why remove the validation?

xiaoanyunfei: > Why remove the validation?



> Why remove the validation?

This may be the easiest way to support  `ephemeral-storage`

xiaoanyunfei: > Why remove the validation?



> Why remove the validation?

This may be the easiest way to support  `ephemeral-storage`

Thor-wl: > > Why remove the validation?
> 
> This may be the easiest way to support `ephemeral-storage`

That may be a little bit rough. Besides, Considering `ephemeral-storage` as part of scalar resource is not so reasonable.

pkg/scheduler/api/helpers/helpers.go
```diff
@@ -49,6 +49,7 @@ func Max(l, r *api.Resource) *api.Resource {
 
 	res.MilliCPU = math.Max(l.MilliCPU, r.MilliCPU)
 	res.Memory = math.Max(l.Memory, r.Memory)
+	res.EphemeralStorage = math.Max(l.EphemeralStorage, r.EphemeralStorage)
```
Thor-wl, line 52: Please update other resource comparsion functions as well such as `Min` `Less` and so on.

william-wang:
@xiaoanyunfei scalar resource is designed to store this kind of resource like gpu, storage, homepage and so forth. It eases your code change while adding a new kind of resource in volcano.

Issue #2191, opened 2022-04-25
k8s cluster totally not work after helm uninstall volcano


pengyuan:
**What happened**:
**k8s cluster totally not work after helm uninstall volcano**

when helm install any other charts, then error occurs.

kubectl get events -w -n volcano

Error creating: Internal error occurred: failed calling webhook "mutatepod.volcano.sh": Post "https://volcano-admission-service.volcano.svc:443/pods/mutate?timeout=10s": service "volcano-admission-service" not found

**What you expected to happen**:

k8s cluster should work as before.

**How to reproduce it (as minimally and precisely as possible)**:

(helm version: 1.5.1)
helm upgrade --install volcano ./volcano -f custom.yaml --namespace volcano

helm uninstall volcano -n volcano

**Anything else we need to know?**:

**Environment**:
- Volcano Version: 0.1
- Kubernetes version (use `kubectl version`): 1.20
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

pengyuan:
#2079 
#2102 

it seems this problem does not resolved.

hwdef:
Yes, you have to clean webhook manually.

jiangxiaobin96:
try https://githubhot.com/repo/volcano-sh/volcano/issues/2079

hwdef:
fix in https://github.com/volcano-sh/volcano/pull/2346
/close

Issue #2102, opened 2022-03-18
Uninstall Volcano Doesn't Clean ValidateWebhook Configuration And MutatingWebhook Configuration


berlinsaint:
**How to reproduce it (as minimally and precisely as possible)**:

step1:  `kubectlcreate -f https://raw.githubusercontent.com/volcano-sh/volcano/v1.5.1/installer/volcano-development.yaml`

step2: `kubectl delete -f https://raw.githubusercontent.com/volcano-sh/volcano/v1.5.1/installer/volcano-development.yaml`

The Webhook Configuration is not  Cleaned because they are not in the yaml.

And the remained configuration will cause the cluster podCreateFailed beacuse of no pod validate endpoint , this is a little problem but not easy to find.

![image](https://user-images.githubusercontent.com/10629930/158919580-a5312dd3-208f-4f34-81a3-1ee2d6e87acb.png)

Suggesstion:

Clean them by using ownerrefernce or labels mechanism

hwdef:
I think it is useful

hwdef:
It is not feasible to use the ownerrefernce,
because webhook should be associated with deployment(volcano-admission), but deployment is namespace scoped, webhook is cluster scoped. kubernetes does not allow deletion across namespaces. 

https://kubernetes.io/zh/docs/concepts/overview/working-with-objects/owners-dependents/

```shell
$ kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace
NAMESPACE   LAST SEEN   TYPE      REASON                     OBJECT                                                                    MESSAGE
default     5m22s       Warning   OwnerRefInvalidNamespace   mutatingwebhookconfiguration/volcano-admission-service-jobs-mutate        ownerRef [apps/v1/Deployment, namespace: , name: volcano-admission, uid: 0d14f232-61dc-40d2-a5f4-a24150e76f25] does not exist in namespace ""
default     5m22s       Warning   OwnerRefInvalidNamespace   mutatingwebhookconfiguration/volcano-admission-service-podgroups-mutate   ownerRef [apps/v1/Deployment, namespace: , name: volcano-admission, uid: 0d14f232-61dc-40d2-a5f4-a24150e76f25] does not exist in namespace ""
[...]

Issue #2079, opened 2022-03-12
Add doc to note how to cleanup `Volcano` completely


Yikun:
#### What would you like to be added:
Add doc to note how to cleanup `Volcano` completely

```
I0312 06:48:38.683192       1 event.go:291] "Event occurred" object="volcano-system/volcano-admission-54b4798bff" kind="ReplicaSet" apiVersion="apps/v1" type="Warning" reason="FailedCreate" message="Error creating: Internal error occurred: failed calling webhook \"mutatepod.volcano.sh\": failed to call webhook: Post \"https://volcano-admission-service.volcano-system.svc:443/pods/mutate?timeout=10s\": dial tcp xxx.xxx.xxx.xxx:443: connect: connection refused"
```

#### Why is this needed:
```
kubectl apply -f ./installer/volcano-development-arm64.yaml
kubectl delete -f ./installer/volcano-development-arm64.yaml
```

Use apply and delete reinstall volcano, we still need to cleanup `validatingwebhookconfigurations` and `mutatingwebhookconfigurations`, then re-apply:
```
k delete validatingwebhookconfigurations volcano-admission-service-jobs-validate volcano-admission-service-pods-validate volcano-admission-service-queues-validate
k delete mutatingwebhookconfigurations volcano-admission-service-jobs-mutate volcano-admission-service-podgroups-mutate volcano-admission-service-pods-mutate volcano-admission-service-queues-mutate
```

https://github.com/volcano-sh/volcano/issues/2102#issuecomment-1072295632: Supplyment: volcano-admission-init will create a secret which cannot be cleaned thoroughly.

LeonardAukea:
For some reason I still see some events like: 

```
Error creating: Internal error occurred: failed calling webhook "mutatepod.volcano.sh": Post "https://volcano-admission-service.volcano-system.svc:443/pods/mutate?timeout=10s": x509: certificate has expired or is not yet valid: current time 2022-03-19T07:53:35Z is after 2021-07-03T10:00:41Z
```

Even after deletion of volcano and  `validatingwebhookconfigurations` , `mutatingwebhookconfigurations`

hwdef:
Here are some improvements I need about webhook:
1. Instead of using code to generate webhook, it is written in yaml and deployed with yaml.This makes it easy to change the scope of application of the webhook and to ignore some pods
2. If possible, narrow the scope of application of webhook, currently intercepting all pods, once volcano has a problem, the entire cluster cannot create pods

[...]

PR #2203, opened 2022-04-28
fix(deploy): add helm pre delete hook to delete orphan resources

shinytang6:
This PR is trying to resolve https://github.com/volcano-sh/volcano/issues/2191, it uses helm hooks to uninstall orphan resources before uninstalling the helm package.

It can perfectly solve the above issue when using helm to deploy, but there still exists two problems:
1. A third-party kubectl dependency will be introduced
2. Installing volcano through `installer/volcano-deployment.yaml` will not take effect, which means issue https://github.com/volcano-sh/volcano/issues/2079, https://github.com/volcano-sh/volcano/issues/2102 still exists
(Though I do think we should not encourage users to deploy in this way)

### DIFF
--- a/installer/README.md
+++ b/installer/README.md
@@ -79,6 +79,7 @@ The following are the list configurable parameters of Volcano Chart and their de
 |`basic.admission_app_name`|Admission Controller App Name|`volcano-admission`|
 |`basic.controller_app_name`|Controller App Name|`volcano-controller`|
 |`basic.scheduler_app_name`|Scheduler App Name|`volcano-scheduler`|
+|`basic.pre_delete_hook_image_name`|Pre Delete Hook Image Name|`bitnami/kubectl:latest`|
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
 
--- a/installer/helm/chart/volcano/templates/admission.yaml
+++ b/installer/helm/chart/volcano/templates/admission.yaml
@@ -22,7 +22,7 @@ rules:
     verbs: ["get", "list", "watch"]
   - apiGroups: ["admissionregistration.k8s.io"]
     resources: ["mutatingwebhookconfigurations", "validatingwebhookconfigurations"]
-    verbs: ["get", "list", "watch", "create", "update"]
+    verbs: ["get", "list", "watch", "create", "update", "delete"]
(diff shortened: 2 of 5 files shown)


hwdef:
For the second reason, it doesn't feel like this change is universally applicable.
Preferably there is a way to remove the residual webhook deployed both ways.

hwdef:
@shinytang6  @Thor-wl 
By the way, no matter what solution we choose, I recommend adding the `FailurePolicy` field. When the webhook cannot connect to the server, ignore the webhook.
This way, even if we don't delete the wehook, it won't cause serious impact.

shinytang6:
> For the second reason, it doesn't feel like this change is universally applicable. Preferably there is a way to remove the residual webhook deployed both ways.

Yes, my personal opinion is:
1. helm is usually the more recommended deployment method(personally preferred way), this pr can completely solve the problem of helm deployment.
2. I think it is difficult to solve it in one file when installing by volcano-deployment.yaml. The essential reason is that deleting these configurations requires the permission of service account, but uninstalling resources in one file cannot ensure the deletion order of sa and other resources, while this is guaranteed by the helm hooks.

The two way I think may be solved is:
* Tell the user to manually delete the webhook residual configuration when using volcano-deployment.yaml.
* It may be possible to declare the webhook resource directly in yaml, but I'm not sure if it will work, and it needs a major modification.

BTW, I don't know if this is the final solution, but it solved my helm deployment problem. So looking forward to more voices.
/cc @k82cn @william-wang @Yikun @berlinsaint

hwdef:
> It may be possible to declare the webhook resource directly in yaml, but I'm not sure if it will work, and it needs a major modification.

Sounds good this way. 

We can merge this pr first to solve the helm problem.
But I have a little suggestion, Can we build a kubectl image ourselves?
Just a small suggestion, bitnami's images are fine too.

Thor-wl:
@shinytang6 Please resolve the conflicts and I will merge it.

shinytang6:
let's wait the implementation of https://github.com/volcano-sh/volcano/pull/2346, then this pr can be closed

shinytang6:
https://github.com/volcano-sh/volcano/pull/2346 merged, this pr is not needed any more
/close

Issue #2213, opened 2022-05-07
Calculate inqueue resource for running jobs not up to expectations


zbbkeepgoing:
**What happened**:

Calculating the inqueue resource of the running job in opensession does not meet the expectation

**What you expected to happen**:

Calculate the inqueue resource of the running job in opensession, which meets the expectation

**How to reproduce it (as minimally and precisely as possible)**:

- Create a queue which capability is` cpu:2000m, memory: 10G`
- Create three SparkApplication which minresource is `cpu:2, memory: 4G` one by one though sleep 1s
- You will see the following process
> - PodGroup of first SparkApplication is Running, and driver is running, executor is pending.
> - PodGroup of second SparkApplication is Running, and driver is running.
> - Spark driver is in a deadlock state 

**Anything else we need to know?**:

There are two bug in Calculate inqueue resource for running jobs

1. `job.PodGroup.Status.Running ` has hysteresis. In high concurrency scenarios, it may not be possible to enter the calculation of the inqueue

https://github.com/volcano-sh/volcano/blob/2cfce7a1305e4ad6d3dcb1a11bf3dc528aee0701/pkg/scheduler/plugins/proportion/proportion.go#L153

2. `rQuantity.Value()` need change to `rQuantity.MilliValue()`
https://github.com/volcano-sh/volcano/blob/2cfce7a1305e4ad6d3dcb1a11bf3dc528aee0701/pkg/scheduler/plugins/util/util.go#L322

**Environment**:
- Volcano Version:
master
- Kubernetes version (use `kubectl version`):
v1.20.4 EKS
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #2214, opened 2022-05-07
fix calculate inqueue resource bug in opensession

zbbkeepgoing:
Fix Issue: https://github.com/volcano-sh/volcano/issues/2213

Signed-off-by: Binbin Zou <zbbkeepgoing@163.com>


zbbkeepgoing:
I'm a bit confused about e2e and lint, my code doesn't get these errors and runs fine locally. Is it possible that this error for e2e is because the sleep time of busybox is too short?

- lint error

TEY79Jmg73

e2e error
image

busybox sleep time (e2e error)
image

Thor-wl (approved):
Thanks for your contribution!

PR #2215, opened 2022-05-07
fix id of gpu devices never delete when number gpu decrease

WingkaiHo:
When some gpu drop on node, and restart volcano gpu share device-plugin，and scheduler never delete id of gpu


shinytang6 (approved):
/lgtm

PR #2216, opened 2022-05-09
numa divided by zero bugfix

elinx:
A typical x86 CPU vendor would encode core ids repeatly in different sockets, but some arm vendors encode the core ids unique and uniformly from 0 to max cores across all sockets. The current implementation calculating physical core numbers by sockets num times de-deplicated cores num would cause **more physical cores than virtual CPUs** in some arm platforms which would cause `CPUs per Core` to zero then the program will crash. This pr aims to create a unified approach to get correct physical core numbers in different CPU vendors, have test both on x86 and arm platform.


shinytang6 (commented):
LGTM overall. Can y please help add some unit tests?

william-wang (approved):
/lgtm

PR #2224, opened 2022-05-11
support nodegroup as a plugin

qiankunli:
nodegroup is the nodes with the same specified label, on the basis of nodegroup, different queues and nodegroups have some relationship about affinity and anti-affinity.

case1: different departments (nlp and recommend) use different nodes (nodegroup1 and nodegroup2), tts uses nodegroup1 first, and nodegroup2 can also be used when resources are insufficient, but recommend can only use nodegroup2.

image

case2: recommend can use private cloud nodes or public cloud nodes, but tts can only use private cloud nodes because its task takes a long time, which is not easy to release public cloud nodes (for saving money)

image

configrue example

```
actions: "reclaim, allocate, backfill, preempt"
tiers:
- plugins:
  - name: priority
  - name: gang
  - name: conformance
- plugins:
  - name: drf
  - name: predicates
  - name: proportion
  - name: nodegroup
    arguments:
      groupantiaffinity.required: 
      groupantiaffinity.preferred:
      groupaffinity.required: nlp = nodegroup1,nodegroup2; recommend=nodegroup2
      groupaffinity.preferred: nlp = nodegroup1
      group.label.name = label1
    enablePredicate: true
    enableNodeOrder: true
```

### DIFF
--- a/pkg/scheduler/framework/arguments.go
+++ b/pkg/scheduler/framework/arguments.go
@@ -85,6 +85,26 @@ func (a Arguments) GetBool(ptr *bool, key string) {
 	*ptr = value
 }
 
+// GetString get the bool value from string
+func (a Arguments) GetString(ptr *string, key string) {
+	if ptr == nil {
+		return
+	}
+
+	argv, ok := a[key]
+	if !ok {
+		return
+	}
+
+	value, ok := argv.(string)
+	if !ok {
+		klog.Warningf("Could not parse argument: %v for key %s to string", argv, key)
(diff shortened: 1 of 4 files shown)


william-wang:
It's an interesting and common scenario. We ever planed to enhance queue ability to allow configuring policy to filter/select nodegroup which contains a specified type of node :)

k82cn:
@qiankunli , can we have a design doc for this feature? I have several comments on the behaviour of this feature :)

PR #2225, opened 2022-05-12
bugfix: api-server deny empty admission response with PatchType set

elinx:
Happen to get an api-server error `webhook returned response.patchType but not response.patch` when using volcano then after some digging into the code, I figure out that api-server needs the `PatchType` and `Patch` pair-matched in the admission response:
https://github.com/kubernetes/apiserver/blob/a0fc11eb6ae0f06eb4fb1f158a1e43e0413adae3/pkg/admission/plugin/webhook/request/admissionreview.go#L66-L73
If an empty `Patch` body is generated and the current implementation always sets the `PatchType` to `JsonType` which will cause the error above. The reason why an empty `Patch` body is generated in my case is I'm not using the `admission-conf` flag of the admission command in my project which will not pass the verification check in the following code for your information:
https://github.com/volcano-sh/volcano/blob/2cfce7a1305e4ad6d3dcb1a11bf3dc528aee0701/pkg/webhooks/admission/pods/mutate/mutate_pod.go#L101-L104

### DIFF
--- a/pkg/webhooks/admission/jobs/mutate/mutate_job.go
+++ b/pkg/webhooks/admission/jobs/mutate/mutate_job.go
@@ -97,8 +97,10 @@ func Jobs(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
 		Allowed: true,
 		Patch:   patchBytes,
 	}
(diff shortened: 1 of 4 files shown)


william-wang (approved):
/lgtm
/approve

PR #2230, opened 2022-05-12
clean up metrics of deleted objects

xiaoanyunfei:
When jobs and queues are delete, the metrics of deleted jobs and queues have never been cleaned up
The memory of volcano-scheduler keeps growing with the number of metrics


william-wang (approved):
/lgtm
/approve

PR #2231, opened 2022-05-13
Feat/exclude unhealthy gpu devices

WingkaiHo:
Exclude unhealhy gpu device by list of node annotations `volcano.sh/gpu-unhealthy-ids`, which patch by volcano device plugin

https://github.com/volcano-sh/devices/pull/24

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -330,6 +330,7 @@ func (ni *NodeInfo) setNodeGPUInfo(node *v1.Node) {
 	}
 	unhealthyGPUs := ni.getUnhealthyGPUs(node)
 	for id := range unhealthyGPUs {
+		klog.Info("delete unhealthy gpu id %d from GPUDevices", id)
 		delete(ni.GPUDevices, id)
 	}
 }
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -329,9 +329,9 @@ func (ni *NodeInfo) setNodeGPUInfo(node *v1.Node) {
 		ni.GPUDevices[i] = NewGPUDevice(i, memoryPerCard)
 	}
 	unhealthyGPUs := ni.getUnhealthyGPUs(node)
-	for id := range unhealthyGPUs {
-		klog.V(4).Infof("delete unhealthy gpu id %d from GPUDevices", id)
-		delete(ni.GPUDevices, id)
+	for i := range unhealthyGPUs {
+		klog.V(4).Infof("delete unhealthy gpu id %d from GPUDevices", unhealthyGPUs[i])
+		delete(ni.GPUDevices, unhealthyGPUs[i])
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/api/node_info.go
```diff
@@ -327,6 +328,11 @@ func (ni *NodeInfo) setNodeGPUInfo(node *v1.Node) {
 	for i := 0; i < int(gpuNumber); i++ {
 		ni.GPUDevices[i] = NewGPUDevice(i, memoryPerCard)
 	}
+	unhealthyGPUs := ni.getUnhealthyGPUs(node)
+	for id := range unhealthyGPUs {
+		klog.Info("delete unhealthy gpu id %d from GPUDevices", id)
```
Thor-wl, line 333: Please set a reasonable log level.

pkg/scheduler/api/node_info.go
```diff
@@ -580,3 +586,24 @@ func (ni *NodeInfo) SubGPUResource(pod *v1.Pod) {
 		}
 	}
 }
+
+// getUnhealthyGPUs returns all the unhealthy GPU id.
+func (ni *NodeInfo) getUnhealthyGPUs(node *v1.Node) (unhealthyGPUs map[int]bool) {
```
Thor-wl, line 591: Why returns a map whose value is bool? According to the following logic, although the current data struct also works, it seems a simpler data struct meets the requirement such as array, which may be more memory friendly.

pkg/scheduler/api/node_info.go
```diff
@@ -580,3 +586,24 @@ func (ni *NodeInfo) SubGPUResource(pod *v1.Pod) {
...
+	unhealthyGPUs = map[int]bool{}
+	devicesStr, ok := node.Annotations[UnhealthyGPUIDs]
```
Thor-wl, line 593: BTW, @william-wang @shinytang6 Does device plugin report the unhealthy GPU ids? And these ids will be patched to node object as an annotation?

pkg/scheduler/api/node_info.go
WingkaiHo: ok

pkg/scheduler/api/node_info.go
WingkaiHo: ok, fix

Thor-wl (commented):
Combinded with [PR-24](https://github.com/volcano-sh/devices/pull/24), it's generally OK to me now. Thanks for your contributioin!
/lgtm

pkg/scheduler/api/node_info.go
william-wang: The device plugin patch the unhealthy gpu id.  We need to make clear the following pr in device plugin firstly and then work on this one.
https://github.com/volcano-sh/devices/pull/24

WingkaiHo: @william-wang volcano-sh/devices#24 have been merge。 Anyother questions in this pr.

william-wang (approved):
/lgtm

Issue #2232, opened 2022-05-13
[preempt] High-priority job can not preempt low-priority job when then queue is overused


wpeng102:
**What happened**:
High-priority job can not preempt low-priority job when then queue is overused.

**What you expected to happen**:
High-priority job  preempt low-priority job 

**How to reproduce it (as minimally and precisely as possible)**:
scheduler configmap: 
```
    actions: "enqueue, allocate, backfill, preempt"
    tiers:
    - plugins:
      - name: priority
      - name: gang
        enablePreemptable: false
        enableJobStarving: false
      - name: conformance
    - plugins:
      - name: overcommit
        arguments:
          overcommit-factor: 10.0
      - name: drf
        enablePreemptable: false
      - name: predicates
      - name: proportion
        enableJobEnqueued: false
      - name: nodeorder
      - name: binpack
```
Queue with capability:
```
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: test-queue
spec:
  capability: 
    cpu: 3
  reclaimable: true
  weight: 1
```

Low-priority job:
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: test-job
spec:
  minAvailable: 1
  schedulerName: volcano
  queue: test-queue
  priorityClassName: low-priority
  policies:
    - event: PodEvicted
      action: RestartTask
  plugins:
    ssh: []
    env: []
    svc: []
  maxRetry: 5
  tasks:
    - replicas: 6
      name: "default-nginx"
      template:
        metadata:
          name: web
          annotations:
            volcano.sh/preemptable: "true"
        spec:
          priorityClassName: low-priority
          containers:
            - image: nginx
              command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep 3600']
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
                requests:
                  cpu: 500m
          restartPolicy: OnFailure
```
High priority job:
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: high-job
spec:
  minAvailable: 1
  schedulerName: volcano
  queue: test-queue
  priorityClassName: high-priority
  policies:
    - event: PodEvicted
      action: RestartJob
  plugins:
    ssh: []
    env: []
    svc: []
  maxRetry: 5
  tasks:
    - replicas: 2
      name: "default-nginx"
      template:
        metadata:
          name: web
        spec:
[...]

PR #2234, opened 2022-05-13
high priority task cannot preemt low priority task when queue is over…

wpeng102:
…used and cluster has idle resource

Fix issue: https://github.com/volcano-sh/volcano/issues/2232

Reproduce steps refer https://github.com/volcano-sh/volcano/issues/2232

Root cause:
When low priority tasks occupied all the queue's capability resources, the high priority tasks need preempt low priority tasks. But if the cluster has idle resources, the preempt action will break the preempt logic, which lead high priority task pending.
https://github.com/volcano-sh/volcano/blob/137bf56499a47764e50b4e2ea7f959ed6b511525/pkg/scheduler/actions/preempt/preempt.go#L243-L245

Solution:
In `preempt`action,  if high priority task's queue is overused it means this high priority task can not be allocated in `allocate` action.  Then need trigger preempt low-priority task.

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -244,7 +244,7 @@ func preempt(
 
 		for !victimsQueue.Empty() {
 			// If reclaimed enough resources, break loop to avoid Sub panic.
-			// If preemptor's queue is overused, it means preemptor can not be allcated. So no ned care about the node idle resourace
+			// If preemptor's queue is overused, it means preemptor can not be allcated. So no need care about the node idle resourace
 			if !ssn.Overused(currentQueue) && preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
 				break
 			}
@@ -263,7 +263,7 @@ func preempt(
 		klog.V(3).Infof("Preempted <%v> for Task <%s/%s> requested <%v>.",
 			preempted, preemptor.Namespace, preemptor.Name, preemptor.InitResreq)
 
-		// If preemptor's queue is overused, it means preemptor can not be allcated. So no ned care about the node idle resourace
+		// If preemptor's queue is overused, it means preemptor can not be allcated. So no need care about the node idle resourace
 		if !ssn.Overused(currentQueue) && preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
 			if err := stmt.Pipeline(preemptor, node.Name); err != nil {
 				klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>",
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -110,7 +110,7 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 
 				preemptor := preemptorTasks[preemptorJob.UID].Pop().(*api.TaskInfo)
 
-				if preempted, _ := preempt(ssn, stmt, preemptor, func(task *api.TaskInfo) bool {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -240,7 +244,8 @@ func preempt(
 
 		for !victimsQueue.Empty() {
 			// If reclaimed enough resources, break loop to avoid Sub panic.
-			if preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
+			// If preemptor's queue is overused, it means preemptor can not be allcated. So no ned care about the node idle resourace
```
Thor-wl, line 247: Tips: no **need** care about the node idle resource. But it doesn't matter.

wpeng102: `preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero)` is old logic, I think it want to handle some pod are in `terminating` case. This node is future idle for preemptor then break preempt logic.

Thor-wl: > `preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero)` is old logic, I think it want to handle some pod are in `terminating` case. This node is future idle for preemptor then break preempt logic.

I mean the wrong spell in the comments.

wpeng102: ：） fixed typos

wpeng102:
@william-wang please help review

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -197,6 +197,7 @@ func preempt(
 	ssn *framework.Session,
 	stmt *framework.Statement,
 	preemptor *api.TaskInfo,
+	job *api.JobInfo,
```
william-wang, line 200: @wpeng102 Are you adding a preemptorJob parameter, if it is, it's better to get the preemptorJob from the preemptor without adding a new parameter.

wpeng102: Address comments in new commit

william-wang (approved):
/approve

hwdef:
@wpeng102 @Thor-wl  @william-wang 
I'm having a problem with preemption, not sure if it's related to the code mentioned in this PR

one node has 12 CPU
queue-a capacity is 4 CPU
low-priority-job request 3 CPU
high-priority-job request 3 CPU

I run low-priority-job first, then I run high-priority-job, high does `not` preempt low

I check volcano source code. 

I find preempt already find a victim, the lowspriority-job, but when the code run this line, it break

```
if !ssn.Overused(currentQueue) && preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
break 
}
```

I think the main reason is that it detects the nodes idle resources.

node has many remain resources, but the queue just has 1 CPU,
This is not a situation that should be BREAKED, but a process that should continue to be preempted

hwdef:
@wangyang0616 @lowang-bh 
Cloud you please check this

lowang-bh:
> node has many remain resources, but the queue just has 1 CPU,
> This is not a situation that should be BREAKED, but a process that should continue to be preempted

Yes, I agree with you. The root cause it that the overused judgement doesn't include the preemtor's resource. I thank it need to add preemtor's request then check queue's oveused. So it need Allocatable judgement as used in allocate action but not overused.

And then the finally logic is:
The current task is skiped allocating resource in Allocate action, due to:
1.  cluster has free resource, but queue not allocatable
2. cluster has no free resource, but queue not allocatable
3. cluster has no free resource, but queue allocatable

As for case 1 and 2, high priority task can preempt low priority task in same queue in Preempt action.
As for case 3, current task should reclaim resource from other queue in reclaim action.

hwdef:
@lowang-bh 
Thanks for your reply. I agree with you.

PR #2245, opened 2022-05-17
Adjust webhook failure policy to ignore so cluster will not get stuck on volcano webhook restart

### LINKED ISSUE #2244: Admission controller can't restart after crashing due to volcano webhook failure policy: fail
**What happened**:

The `volcano-admission` Pod crashed and tried to start back up, but was denied admission because there is a self-referential dependency for this Pod to be admitted.

volcano-admission Pod can crash and restart

```
# First scale down the volcano-admission Pod to simulate a Pod crash or OOM event
kubectl scale deployment volcano-admission -n volcano-system --replicas=0

# Now wait for the volcano-admission pod to stop
kubectl scale deployment volcano-admission -n volcano-system --replicas=1

# Notice the volcano-admission pod won't come back into a running state
kubectl get pod -n volcano-system                                                                                                                                                                                                        

NAME                                   READY   STATUS      RESTARTS   AGE
volcano-admission-init-rgdn2           0/1     Completed   0          54m
volcano-controllers-7c6b5c4f4b-cxvg5   1/1     Running     0          54m
volcano-scheduler-7f9d9984fd-zz52w     1/1     Running     0          54m
```

**Environment**:
- Volcano Version: v1.5.1
- Kubernetes version (use `kubectl version`): 1.21
- Cloud provider or hardware configuration: GKE
- Install tools: In a kubeflow env


I have a patch for this: simply set the retrypolicy for webhooks to ignore instead of fail when the webhook is not available. The admission controller should be multi-replica or else the failurePolicy on the webhook should not be set to fail.

djwhatle:
Fixes #2244 

I believe it's unsafe to have `failurePolicy: fail` set on webhooks when the admission controller is not highly available.

### DIFF
--- a/cmd/webhook-manager/app/util.go
+++ b/cmd/webhook-manager/app/util.go
@@ -23,7 +23,7 @@ import (
 	"regexp"
 	"strings"
 
-	"k8s.io/api/admissionregistration/v1"
+	v1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -37,6 +37,7 @@ import (
 
 func registerWebhookConfig(kubeClient *kubernetes.Clientset, config *options.Config, service *router.AdmissionService, caBundle []byte) {
 	sideEffect := v1.SideEffectClassNoneOnDryRun
+	failurePolicyIgnore := v1.Ignore
 	reviewVersions := []string{"v1"}
 	clientConfig := v1.WebhookClientConfig{
 		CABundle: caBundle,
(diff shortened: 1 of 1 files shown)


djwhatle:
This PR addresses #2245

djwhatle:
I noticed this issue on release-1.5.1. Please let me know if I should target this change at a different branch.

Thor-wl:
Thanks for your fix, @djwhatle. Please also cherry-pick this fix to release-1.5 and release-1.6.

hwdef:
Should we ignore pods in `volcano-system` and `kube-system`?

hwdef:
After careful consideration, I feel that this failure policy is unreliable. When the admission is abnormal, objects such as vcjob and pod that have not been verified by webhook will directly enter the controller and scheduler, which may cause panic of the scheduler and the controller.

djwhatle:
@hwdef the solution you mention of ignoring Pods in the volcano-system namespace seems like a good alternative. I wasn't sure of the potential side-effects of this change. 

I'll open an alternate PR that ignores the volcano-system and kube-system ns based on the selector

```
kubernetes.io/metadata.name: volcano-system
kubernetes.io/metadata.name: kube-system
```

I don't see a mechanism on webhook config to ignore namespaces by the name of the namespace, have to use a label.

hwdef:
@djwhatle 
Thank you for your replay and look forward to seeing your new PR.

djwhatle:
Closing in favor of #2251

PR #2253, opened 2022-05-20
fix nil pointer in VictimTasks

Thor-wl:
This patch fix the bug that when `VictimTasks ` is reset to `nil` at the end of a session, it will introduce a panic when called at the next session.


william-wang (approved):
/lgtm

william-wang:
@Thor-wl please make the ci happy.

PR #2255, opened 2022-05-23
support overcommit factor in proportion plugin

qiankunli:
image

I am trying to use volcano schedule pods in constant node pool(pay annually) and  elastic node pool(pay by the hour). it means that the total resources of the cluster are changing. at this time, it is not necessary for the proportion plugin to judge whether the task is enqueued according to the current total resources.

```yaml
   tiers:
   - plugins:
     - name: proportion
       arguments:
         overcommit-factor: 1.5
```

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -32,14 +32,23 @@ import (
 )
 
 // PluginName indicates name of volcano scheduler plugin.
-const PluginName = "proportion"
+const (
+	PluginName = "proportion"
+	// overCommitFactor is resource overCommit factor for enqueue action
+	// It determines the number of `pending` pods that the scheduler will tolerate
+	// when the resources of the cluster is insufficient
+	overCommitFactor = "overcommit-factor"
+	// defaultOverCommitFactor defines the default overCommit resource factor for enqueue action
+	defaultOverCommitFactor = 1.0
+)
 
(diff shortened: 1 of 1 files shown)


qiankunli:
there are several ways to solve this problem
1. tells the proportion plugin a fake total resource 

	```
    tiers:
    - plugins:
      - name: proportion
        arguments:
          overcommit-factor: 1.5
	```

2. allows the status of any queue without capacity to be changed from Pending to Inqueue. 
	1. We may need to customize a new  enqueue action because the existing enqueue action cannot receive parameters.
	2. We may need to customize a new  enqueue action responsible for automatically changing all Pending PodGroup to Inqueue

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -32,14 +32,23 @@ import (
 )
 
 // PluginName indicates name of volcano scheduler plugin.
-const PluginName = "proportion"
+const (
+	PluginName = "proportion"
+	// overCommitFactor is resource overCommit factor for enqueue action
+	// It determines the number of `pending` pods that the scheduler will tolerate
+	// when the resources of the cluster is insufficient
+	overCommitFactor = "overcommit-factor"
```
hwdef, line 40: Variables in this style are deprecated.Does not conform to k8s style.
```suggestion
	overCommitFactor = "overcommitFactor"
```

Issue #2238, opened 2022-05-16
can not update a pending vcjob


daizry:
**What happened**:
According to [pull1137](https://github.com/volcano-sh/volcano/pull/1137), updating priorityClassName is supported for a pending job. However, volcano denied the update request with Error like this:  
 **for: "vcjob.yaml": admission webhook "validatejob.volcano.sh" denied the request: job updates may not change filelds other than `minAvailable`, `tasks[*].replicas under spec`**
First off, this error message doesn't mention `priorityClassName` at all, so this error message is outdated.
Further more, I've tried to change fields like minAvaliable and even apply the same yaml file again, but still got the same Error. That probably means that job update function does not work well.
**What you expected to happen**:
volcano accepts the updates and changes the configuraton of the pending job
**How to reproduce it (as minimally and precisely as possible)**:
1. ```kubectl apply -f vcjob.yaml -n xxxx```apply a simple vcjob and keep it remains "Pending" state
2. edit the `priorityClassName` filed in the yaml file
3. apply the edited file again

**Anything else we need to know?**:

**Environment**:
- Volcano Version: 1.5.1
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zbbkeepgoing:
Because vcjob.spec.task[*].maxRetry has't considered in job update verification, so apply the source yaml file when you do not specify maxRetry, the problem will occur.

**Note:** Workaround before my code merged:   replace `kubectl apply -f file.yaml` with `kubectl edit vcjob xxx `

daizry:
Thx, after adding `vcjob.spec.task[*].maxRetry` into yaml file, `kubectl apply -f ` worked.  
Also, `kubectl edit vcjob xxx` worked without `vcjob.spec.task[*].maxRetry`.   
Whereas, updating `priorityClassName` field still caused the same error

zbbkeepgoing:
> Thx, after adding `vcjob.spec.task[*].maxRetry` into yaml file, `kubectl apply -f ` worked. Also, `kubectl edit vcjob xxx` worked without `vcjob.spec.task[*].maxRetry`. Whereas, updating `priorityClassName` field still caused the same error

Yes, as expected. So `kubectl edit vcjob xxx` cannot remove maxRetry in task

zbbkeepgoing:
[...]

PR #2259, opened 2022-05-26
support update maxretry

zbbkeepgoing:
Support update vcjob.maxRetry & vcjob.tasks[*].maxRetry and  fix Issue: https://github.com/volcano-sh/volcano/issues/2238

Signed-off-by: Binbin Zou <zbbkeepgoing@163.com>

### DIFF
--- a/pkg/webhooks/admission/jobs/validate/admit_job.go
+++ b/pkg/webhooks/admission/jobs/validate/admit_job.go
@@ -277,9 +277,11 @@ func validateJobUpdate(old, new *v1alpha1.Job) error {
 	// other fields under spec are not allowed to mutate
 	new.Spec.MinAvailable = old.Spec.MinAvailable
 	new.Spec.PriorityClassName = old.Spec.PriorityClassName
+	new.Spec.MaxRetry = old.Spec.MaxRetry
 	for i := range new.Spec.Tasks {
 		new.Spec.Tasks[i].Replicas = old.Spec.Tasks[i].Replicas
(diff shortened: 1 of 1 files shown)


Thor-wl:
Thanks for this fix. I'm considering the rationality of letting users to update the maxtry of a vcjob. How do you think about that? @shinytang6 @william-wang @qiankunli

zbbkeepgoing:
> Thanks for this fix. I'm considering the rationality of letting users to update the maxtry of a vcjob. How do you think about that? @shinytang6 @william-wang @qiankunli

Agree it.
In addition, MaxRetry only works at the controller level, not at the scheduler level.

zbbkeepgoing:
@Thor-wl  Some other ideas about this?

Thor-wl:
> @Thor-wl Some other ideas about this?

Hey. I've reviewed the PR and related issue. As what I understand, your original scenario is to support job priorityclass modification in the admission webhook controller. Right? If that, perhaps you can ping @merryzhou for a patch.

zbbkeepgoing:
> > @Thor-wl Some other ideas about this?
> 
> Hey. I've reviewed the PR and related issue. As what I understand, your original scenario is to support job priorityclass modification in the admission webhook controller. Right? If that, perhaps you can ping @merryzhou for a patch.

No, it is fix this issue https://github.com/volcano-sh/volcano/issues/2238. The essence here is, if we not support maxretry in admission webhook, we can't update vcjob though "kubectl apply -f vcjob.yaml". 

@merryzhou Do you have time look it？

PR #2260, opened 2022-05-28
bugfix: avoid panic for query prometheus no data

waiterQ:
it occurred panic when query prometheus no data return, add somecode to avoid it.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1261,27 +1261,12 @@ func (sc *SchedulerCache) GetMetricsData() {
 func (sc *SchedulerCache) setMetricsData(usageInfo map[string]*schedulingapi.NodeUsage) {
 	sc.Mutex.Lock()
 	defer sc.Mutex.Unlock()
-	beforeNodeinfo := make(map[string]string, len(sc.Nodes))
-	for nName, nInfo := range sc.Nodes {
-		beforeNodeinfo[nName] = fmt.Sprintf("%+v", *nInfo.ResourceUsage)
-	}
+
 	for k := range usageInfo {
 		nodeInfo, ok := sc.Nodes[k]
 		if ok {
+			klog.V(4).Infof("node: %s, ResourceUsage: %+v => %+v", k, *nodeInfo.ResourceUsage, *usageInfo[k])
 			nodeInfo.ResourceUsage = usageInfo[k]
 		}
-	}
-	// record node.ResourceUsage changes
-	for nName, nInfo := range sc.Nodes {
-		newResourceUsage := fmt.Sprintf("%+v", *nInfo.ResourceUsage)
-		oldResourceUsage, found := beforeNodeinfo[nName]
-		if !found {
-			klog.V(4).Infof("new node: %s, ResourceUsage: %s", nName, newResourceUsage)
-			continue
-		}
-		if oldResourceUsage == newResourceUsage {
-			continue
-		}
-		klog.V(4).Infof("node: %s, ResourceUsage: %s => %s", nName, oldResourceUsage, newResourceUsage)
 	}
 }
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1225,8 +1225,16 @@ func (sc *SchedulerCache) GetMetricsData() {
 				if len(warnings) > 0 {
 					klog.V(3).Infof("Warning querying Prometheus: %v", warnings)
 				}
+				if res == nil || res.String() == "" {
+					klog.V(4).Infof("Warning querying Prometheus: no data found for %s", queryStr)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -1225,8 +1225,16 @@ func (sc *SchedulerCache) GetMetricsData() {
...
 					klog.V(3).Infof("Warning querying Prometheus: %v", warnings)
 				}
+				if res == nil || res.String() == "" {
+					klog.V(4).Infof("Warning querying Prometheus: no data found for %s", queryStr)
+					continue
+				}
 
+				klog.V(4).Infof("Query prometheus res %s", res.String())
 				rowValues := strings.Split(strings.TrimSpace(res.String()), "=>")
+				if len(rowValues) < 2 {
```
Thor-wl, line 1235: I'm sorry for a little bit confused about this condition. Can you add some comments for that?

waiterQ: done it.

pkg/scheduler/cache/cache.go
```diff
@@ -1225,8 +1225,17 @@ func (sc *SchedulerCache) GetMetricsData() {
...
 				}
+				if res == nil || res.String() == "" {
+					klog.V(4).Infof("Warning querying Prometheus: no data found for %s", queryStr)
+					continue
+				}
 
+				klog.V(4).Infof("Query prometheus res %s", res.String())
 				rowValues := strings.Split(strings.TrimSpace(res.String()), "=>")
+				// ignore invalid prometheus.data to nodeUsage.map
+				if len(rowValues) < 2 {
```
william-wang, line 1236: in which case, the rowValues is less 2?

pkg/scheduler/cache/cache.go
```diff
@@ -1247,10 +1261,27 @@ func (sc *SchedulerCache) GetMetricsData() {
 func (sc *SchedulerCache) setMetricsData(usageInfo map[string]*schedulingapi.NodeUsage) {
 	sc.Mutex.Lock()
 	defer sc.Mutex.Unlock()
+	beforeNodeinfo := make(map[string]string, len(sc.Nodes))
+	for nName, nInfo := range sc.Nodes {
```
william-wang, line 1265: The pr is target to fix the panic issue. do not optimize code in this pr. 
Also please avoid doing the loop for non-important thing. It's might be a large cluster. And multiple go routines share the cache. There is a lock in this function. Make the code simple enough.

waiterQ: i will keep it in mind, simplify the code for logging

william-wang:
LGTM overall
/approve

pkg/scheduler/cache/cache.go
```diff
@@ -1225,8 +1226,21 @@ func (sc *SchedulerCache) GetMetricsData() {
...
 					klog.V(3).Infof("Warning querying Prometheus: %v", warnings)
 				}
+				if res == nil || res.String() == "" {
+					klog.V(4).Infof("Warning querying Prometheus: no data found for %s", queryStr)
+					continue
+				}
 
-				rowValues := strings.Split(strings.TrimSpace(res.String()), "=>")
+				klog.V(4).Infof("Query prometheus res %s", res.String())
+				switch res.Type() {
```
Thor-wl, line 1235: So only `pmodel.ValVector` is necessary,  if it is another type, just skip the metric process? If that, how about use a `if` to judge that?

Thor-wl: Another issue I doubt is that if something goes wrong and Volcano cannot get the target metrcs, would it  process the following logic with pod's request resource? Namely, swtich to the default mode.

waiterQ: in case [prometheus.rules](https://github.com/volcano-sh/volcano/blob/master/docs/design/usage-based-scheduling.md#prometheus-rule-configuration), prometheus.api.query only type pmodel.ValVector can occurred, so ignore anyother type
plugin usage and predicates can process parallelly, they shouldnt touch each other, no situation "switch" is expected to happen, can imports both as backups for each other

pkg/scheduler/cache/cache.go
```diff
@@ -1225,8 +1226,21 @@ func (sc *SchedulerCache) GetMetricsData() {
 				if len(warnings) > 0 {
 					klog.V(3).Infof("Warning querying Prometheus: %v", warnings)
 				}
+				if res == nil || res.String() == "" {
+					klog.V(3).Infof("Warning querying Prometheus: no data found for %s", queryStr)
```
Thor-wl, line 1230: Suggest to make use of `klog.warning` for this process.

waiterQ: done

PR #2263, opened 2022-05-30
gang blocks task preemption

wpeng102:

This is code mistake, in `gang` plugin only high priority job's `jobStarvingFn` return true will trigger high priority job preempt low priority job. 

https://github.com/volcano-sh/volcano/blob/c708952c89a1ecc3b39ce5fa8b517d59bd808310/pkg/scheduler/plugins/gang/gang.go#L155-L162

But in `ji.CheckTaskMinAvailablePipelined()`, high priority job's occupied task is '0', of cause it is less than `minNum`. So this will lead  `jobStarvingFn` return false. 
https://github.com/volcano-sh/volcano/blob/c708952c89a1ecc3b39ce5fa8b517d59bd808310/pkg/scheduler/api/job_info.go#L789-L794

### DIFF
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -155,7 +155,7 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
 	jobStarvingFn := func(obj interface{}) bool {
 		ji := obj.(*api.JobInfo)
 		occupied := ji.WaitingTaskNum() + ji.ReadyTaskNum()
-		if ji.CheckTaskMinAvailablePipelined() && occupied < ji.MinAvailable {
+		if !ji.CheckTaskMinAvailablePipelined() && occupied < ji.MinAvailable {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/gang/gang.go
```diff
@@ -155,7 +155,7 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
 	jobStarvingFn := func(obj interface{}) bool {
 		ji := obj.(*api.JobInfo)
 		occupied := ji.WaitingTaskNum() + ji.ReadyTaskNum()
-		if ji.CheckTaskMinAvailablePipelined() && occupied < ji.MinAvailable {
+		if !ji.CheckTaskMinAvailablePipelined() && occupied < ji.MinAvailable {
```
william-wang, line 158: the fix is good to me. is there any e2e or ut against the case? 
The mistake code pr should trigger the e2e failure.

PR #2268, opened 2022-06-01
Default queue name from namespace annotations, when no queue name in pod spec

### LINKED ISSUE #2261: Default queue name from namespace labels, when no queue name in pod spec
Currently when no queue name is mentioned in pod spec, it will default to "default" queue. But it will be helpful if we default the queue name from namespace label(scheduling.volcano.sh/queue-name) if exist.

Example Namespace:

```
apiVersion: v1
kind: Namespace
metadata:
  labels:
    volcano.sh/queue-name: namespaace-queue
  name: namespace1
```

#### Why is this needed:

When namespace is dedicated to user and queue name can change dynamically based on when team he belongs to. Currently need to change pod spec when user moved from one team to another. Just changing the namespace label will make life easy.

Sharathmk99:
Fixes #2261 

Currently when no queue name is mentioned in pod spec, it will default to "default" queue. But it will be helpful if we default the queue name from namespace annotations(scheduling.volcano.sh/queue-name) if exist.

### DIFF
--- a/pkg/webhooks/admission/podgroups/mutate/mutate_podgroup.go
+++ b/pkg/webhooks/admission/podgroups/mutate/mutate_podgroup.go
@@ -17,6 +17,7 @@ limitations under the License.
 package mutate
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


william-wang (approved):
/lgtm
/approve

PR #2269, opened 2022-06-01
 TDM plugins add timezone suppose

Jason-Liu-Dream:
1. some times we didn't know timezone of default date, especially when running in a container.
2. timezone like UTC is not very readable in our region，so we need to translate time，that not friendly，so let‘s make some things.

### DIFF
--- a/pkg/scheduler/plugins/tdm/tdm.go
+++ b/pkg/scheduler/plugins/tdm/tdm.go
@@ -37,6 +37,7 @@ const (
 	// revocableZoneLayout revocable zone layout
 	revocableZoneLayout      = "15:04"
 	revocableZoneLabelPrefix = "tdm.revocable-zone."
+	timeZoneLabel            = "tdm.timezone"
 	evictPeriodLabel         = "tdm.evict.period"
 	defaultPodEvictNum       = 1
 )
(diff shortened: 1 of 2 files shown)


hwdef:
Why not change the timezone by mounting the file?

william-wang:
@Jason-Liu-Dream thanks for the PR. Please follow the guideline, open a issue firstly and more contributors can fully discuss there.  Thanks:)

Jason-Liu-Dream:
> @Jason-Liu-Dream thanks for the PR. Please follow the guideline, open a issue firstly and more contributors can fully discuss there. Thanks:)
ok，i create a issue，#2277，let‘s discuss at here

Jason-Liu-Dream:
> Why not change the timezone by mounting the file?



> Why not change the timezone by mounting the file?

did you mean to say this way to change：mount /etc/localtime to container？I think we need keep use utc to system，because it's a convention and other this may use it. but in this plugins，we can use this label to help us more intuitive.

Jason-Liu-Dream:
> Why not change the timezone by mounting the file?


did you mean to say this way to change：mount /etc/localtime to container？I think we need keep use utc to system，because it's a convention and other this may use it. but in this plugins，we can use this label to help us more intuitive.

Issue #2273, opened 2022-06-02
[CI Fix]Github actions don't work occasionally


Thor-wl:
**What happened**:
When I submitted [PR-2272](https://github.com/volcano-sh/volcano/pull/2272), the github action didn't work well. After a short investigation, I found something different as follows.
* Only [E2E about Parallel Jobs](https://github.com/volcano-sh/volcano/runs/6703861243?check_suite_focus=true#logs) can't work well. The logs indicates the failure about 'actions/cache@v2' or 'actions/cache@v2'.
* No matter how many times I retried the job, it still failed.
* Other jobs such as [E2E about Basic Scheduling](https://github.com/volcano-sh/volcano/runs/6703558831?check_suite_focus=true#logs), which also contain these actions, work well.

When I searched the github, I found some report problems similar with mine such as [Failed to Download](https://github.com/actions/checkout/issues/672).

**What you expected to happen**:
CI works well
**How to reproduce it (as minimally and precisely as possible)**:
It seems something wrong with github actions and can not reproduce stably. I liste the log with link above to catch something useful.

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Thor-wl:
Github action  recoveries now.
/close

PR #2274, opened 2022-06-02
fix github action bug and set the release of checkout to be fixed v2.3.5

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/2273
According to the [workaround](https://github.com/actions/checkout/issues/672#issuecomment-1010571414), set the release of checkout to be v2.3.5.


Thor-wl:
Well, it seems no effect with the recomanded workaround. I'll try to find other ways.

PR #2290, opened 2022-06-09
cmd: add --managed-namespaces flag to vc-webhook-manager

### LINKED ISSUE #2286: Volcano's webhooks should be more respectful of other namespaces and RBAC permissions should be reduced

I would like to see Volcano default to managing an explicit list of namespaces rather than managing all namespaces by default. To me, this behavior is both unexpected and problematic. It blocks deployment of other applications. It also leads to the need for less secure RBAC configuration where Volcano can manage all Pods in the cluster.

Currently Volcano's webhooks are able to ignore specific namespaces by using the `--ignored-namespaces` flag when calling `vc-webhook-manager`. That is helpful, but I think that Volcano should be more restrictive in terms of what it manages. Rather than needing to list which namespaces to ignore, it should be possible to list the namespaces that should be managed.

I propose that there should be a `--managed-namespaces` flag. It should operate the same as the `--ignored-namespaces` flag, but instead of using the `NotIn` operator in the namespaceSelector it should use `In`. `--managed-namespaces` should be mutually exclusive with `--ignored-namespaces`.

#### Questions
Why does Volcano manage all namespaces by default?
- I'm sure there is a good reason for this, but is it something that can be addressed?

Would there be side effects of this in the Volcano system? 
- If so, can they be addressed? 
- For example, would this allow creating VCJobs outside of managed namespaces (probably not good)?
- Could this be protected against using additional webhooks to block creation of VCJobs in unmanaged namespaces?

#### Why is this needed:
Generally, I think that it is a bad idea to take over the entire kubernetes cluster by default. People use clusters for many purposes and won't expect Volcano to block deployment of other applications. Currently, once Volcano is deployed, it must be patched before a new application is deployed in a new namespace.

This would also allow more restrictive RBAC for Volcano. It would allow reducing the scope of the ClusterRoles to only manage Pods in the managed namespaces rather than requiring cluster-wide access to all Pods. This would be significantly more secure.

jsolbrig:
As described in #2286 this adds the `--managed-namespaces` flag to the vc-webhook-manager binary. The flag is mutually exclusive with the `--ingored-namespaces` flag.

As a note, I personally think that `--ignored-namespaces` is a dangerous idea. It essentially leans towards "most privileged" rather than "least privileged".

#### Syntax question
I'm not really a Go programmer (mostly Python) but am learning. I struggled with one thing. On line 61 of `cmd/webhook-manager/app/util.go` I used `var selectorOperator metav1.LabelSelectorOperator`. I'm not sure if this is the preferred syntax for this project but struggled to get any other syntax to work here.

#### Removal of default ingored namespaces
In adding this flag, it was significantly simpler to remove the `defaultIgnoredNamespaces`. I also think that setting the list of ignored/managed namespaces should be explicit. 

This became an issue for me while working on putting together a Helm chart for Volcano and found that having a default set of ignored namespaces made developing a functional Helm chart difficult. For example, Helm has the convention of specifying which namespace a chart will be deployed do using `helm install --namespace`. With `defaultIgnoredNamespaces` set, it was non-obvious why the chart was managing the wrong namespace.

#### RBAC
While #2286 mentions the issue of over-broad RBAC, in order to keep this PR small, I haven't done anything to address the RBAC issues. I will, however, attempt to address those issues in the helm chart mentioned above. I can open a new issue for the RBAC issues if needed.

Fixes #2286

### DIFF
--- a/cmd/webhook-manager/app/options/options.go
+++ b/cmd/webhook-manager/app/options/options.go
@@ -26,11 +26,10 @@ import (
 )
 
 const (
-	defaultSchedulerName     = "volcano"
-	defaultQPS               = 50.0
-	defaultBurst             = 100
-	defaultEnabledAdmission  = "/jobs/mutate,/jobs/validate,/podgroups/mutate,/pods/validate,/pods/mutate,/queues/mutate,/queues/validate"
-	defaultIgnoredNamespaces = "volcano-system,kube-system"
+	defaultSchedulerName    = "volcano"
+	defaultQPS              = 50.0
+	defaultBurst            = 100
+	defaultEnabledAdmission = "/jobs/mutate,/jobs/validate,/podgroups/mutate,/pods/validate,/pods/mutate,/queues/mutate,/queues/validate"
 )
 
 // Config admission-controller server config.
@@ -52,6 +51,7 @@ type Config struct {
 	ConfigPath        string
 	EnabledAdmission  string
 	IgnoredNamespaces string
+	ManagedNamespaces string
 }
 
 type DecryptFunc func(c *Config) error
@@ -82,7 +82,10 @@ func (c *Config) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.EnabledAdmission, "enabled-admission", defaultEnabledAdmission, "enabled admission webhooks")
 	fs.StringVar(&c.SchedulerName, "scheduler-name", defaultSchedulerName, "Volcano will handle pods whose .spec.SchedulerName is same as scheduler-name")
 	fs.StringVar(&c.ConfigPath, "admission-conf", "", "The configmap file of this webhook")
-	fs.StringVar(&c.IgnoredNamespaces, "ignored-namespaces", defaultIgnoredNamespaces, "Comma-separated list of namespaces to be ignored by admission webhooks")
+	fs.StringVar(&c.IgnoredNamespaces, "ignored-namespaces", "", "Comma-separated list of namespaces to be ignored by admission webhooks"+
+		"(Mutually exclusive with --managed-namespaces)")
+	fs.StringVar(&c.ManagedNamespaces, "managed-namespaces", "", "Comma-separated list of namespaces to be managed by admission webhooks"+
+		"(Mutually exclustive with --ignored-namespaces)")
 }
 
 // CheckPortOrDie check valid port range.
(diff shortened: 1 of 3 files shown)


jsolbrig:
I'm confused about the issue that the bot is reporting with my commit message. I followed the instructions [here](https://github.com/volcano-sh/volcano/blob/master/contribute.md#format-of-the-commit-message). Should I remove the part that says `Fixes: #2286`? Do the instructions or the bot need to be updated?

jsolbrig:
Thanks @hwdef I signed the commits.

hwdef:
please rebase your code. Merge 2 commits into 1.

jsolbrig:
Thanks @hwdef, I've squashed the two commits into one and removed the "Fixes: #2286" portion to get the bot to quit complaining.

hwdef (commented):
I think we should judge more carefully. For example, no matter which namespace vcjob is, it must go through webhook and cannot be ignored.
Not sure if this is easy to do.

jsolbrig:
Yes, you're right. I didn't do this completely. I guess that I got a little excited. 

I'm looking more closely at what I did and what needs to be done. I think I should probably either retract this PR or set it to "draft" and do some more work before resubmitting. Do you have a preference?

Before I keep working on this, if done correctly, is this functionality that you're interested in? Personally, without it, I will probably need to find a different batch processing solution since Volcano currently interferes with deploying other applications by blocking the deployment of Pods.

jsolbrig:
@hwdef Sorry for wasting your time with this PR. I took on something too complicated for my first issue in this project and the work that I did is largely junk since it doesn't begin to get into the actual webhook code. 

I do think that I see a solution that can easily achieve my basic goal for now. Is there a way that we can discuss how this can work?

The basic goal is to just get Volcano to quit blocking admission for resources from other applications. VCJobs add specific labels to the Pods that they create. Maybe simply adding an ObjectSelector to the Pod webhooks would be enough to avoid blocking pods from other applications. Something like this:

```yaml
- apiVersion: admissionregistration.k8s.io/v1
  metadata:
    name: volcano-admission-pods-mutate
  webhooks:
  - admissionReviewVersions:
    objectSelector:
      matchExpressions:
      - {key: volcano.sh/job-name, operator: Exists}
```

hwdef:
@jsolbrig 
I'm not sure if it's feasible to use objectSelector. It's feasible to filter vcjobs through it, but there may be no way to filter deployments, or pods created through deployments.
You're doing a great job now, we just need to refine the scenario.
Is it possible to combine namespaceSelector and matchPolicy to filter by namespace and ensure that vcjobs must go through webhook?

jsolbrig:
@hwdef 

Sorry for the delay on this. I'll get back to it sometime soon. I've needed to move on for now but it is still a priority for me to get this working correctly.

hwdef:
I think after https://github.com/volcano-sh/volcano/pull/2346 merged. We do not need this PR.

jsolbrig:
@hwdef I think you're right but I'll give it a good look sometime in the next week or so.

Sorry it's taking me so long to get back to this. I've had other things climb up my priority list...

PR #2298, opened 2022-06-13
support task dag with gang

HecarimV:
[Task dependency in job](https://github.com/volcano-sh/volcano/blob/master/docs/design/task-launch-order-within-job.md) is supported in Volcano. 
However, the task depends function must remove the gang plugin to run normally.
I'm trying to get the task dag and gang plugin to work together. So I made some simple modifications.
- Special handling for Jobs with task dependencies when check  job valid. 
- keep pg inqueue when task dag is not finished.

Just a simple try, I want to discuss if it is possible.

Signed-off-by: HaoJie Liu <1187851227@qq.com>

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -44,6 +44,10 @@ import (
 
 var calMutex sync.Mutex
 
+const (
+	customLaunch = "volcano.sh/custom-launch"
+)
+
 func (cc *jobcontroller) killJob(jobInfo *apis.JobInfo, podRetainPhase state.PhaseMap, updateStatus state.UpdateStatusFn) error {
 	job := jobInfo.Job
 	klog.V(3).Infof("Killing Job <%s/%s>, current version %d", job.Namespace, job.Name, job.Status.Version)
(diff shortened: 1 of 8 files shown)


hwdef:
Please make CI happy first.

HecarimV:
> Please make CI happy first.

okay, I'll check it.

pkg/controllers/job/job_controller_actions.go
```diff
@@ -44,6 +44,10 @@ import (
 
 var calMutex sync.Mutex
 
+const (
+	customLaunch = "volcano.sh/custom-launch"
```
Thor-wl, line 48: BTW, what's the meaning of this annotation? Pls add some comments about that.

HecarimV: I want to use this annotation to mark jobs that have task dependencies，or just change the key as “has-task-dependencies”

HecarimV:
> @HecarimV Thanks for your pr. It's a feature enhancement. Would you like to provide a doc based on "task dependency in job" firstly?

Okay, I will organize the ideas into documents.

Issue #2040, opened 2022-02-26
refator 'enqueue' scheduling action


william-wang:
#### What would you like to be added:
refator `enqueue` scheduling action

#### Why is this needed:
The `enqueue` action is a mandatory action right now. It would be wonderful to make it optional and configurable.
Firstly the integration code with other community will be more simple and easy.
Secondly the other actions no need to depend on `enqueue` . All the actions are decoupled.

Actions:
1. Remove the pod admit that pod is not allowed to be created while it's podgroup is in `Pending` state for non-vcjob such as tfjob, spark, etc. This will make volcano integrate with upstream more easier. Previously upstream have to create podgroup, waits for podgroup entry `inqueue` and then create pod. This causes confused for upstream developer. What they expect is just creating podgroup which carring all of scheduling info only.
3.  make the enqueue action configurable.

shinytang6:
resolved by https://github.com/volcano-sh/volcano/pull/2078 and https://github.com/volcano-sh/volcano/pull/2309
/close

PR #2309, opened 2022-06-25
change enqueue to optional action

wpeng102:

The enqueue action is a mandatory action right now. It would be wonderful to make it optional and configurable.
Firstly the integration code with other community will be more simple and easy.
Secondly the other actions no need to depend on enqueue . All the actions are decoupled.

https://github.com/volcano-sh/volcano/issues/2040

If not config `Enqueue` action, scheduler will change the `Pending` pg to `Inqueue` at closesession.

### DIFF
--- a/pkg/scheduler/conf/scheduler_conf.go
+++ b/pkg/scheduler/conf/scheduler_conf.go
@@ -16,6 +16,9 @@ limitations under the License.
 
 package conf
 
+// EnabledActionMap check if a action exist in scheduler configmap. If not exist the value is false.
+var EnabledActionMap map[string]bool
+
 // SchedulerConfiguration defines the configuration of scheduler.
 type SchedulerConfiguration struct {
 	// Actions defines the actions list of scheduler in order
--- a/pkg/scheduler/framework/job_updater.go
+++ b/pkg/scheduler/framework/job_updater.go
@@ -11,9 +11,8 @@ import (
 
 	"volcano.sh/apis/pkg/apis/scheduling"
 	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/conf"
 )
-
-var EnqueueExist bool
 
 const (
 	jobUpdaterWorker = 16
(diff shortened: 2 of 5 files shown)


pkg/scheduler/framework/job_updater.go
```diff
@@ -13,6 +13,8 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/api"
 )
 
+var EnqueueExist bool
```
Thor-wl, line 16: Kindly to ask about why hard code enqueue action to make the judgement?

wpeng102: Introduce ActionMap to store which action is configured by customer.

wpeng102:
Backgroud:
Some customers wants volcano do not block pod created when resource is insufficient, just keep same behavior as default scheduler.

pkg/scheduler/conf/scheduler_conf.go
```diff
@@ -16,6 +16,9 @@ limitations under the License.
 
 package conf
 
+// ActionMap check if a action exist in scheduler configmap
+var ActionMap map[string]bool
```
Thor-wl, line 20: Maybe `ExistActionMap` is better?

wpeng102: EnabledActionMap is better

Thor-wl: I'm OK about that.

pkg/scheduler/framework/job_updater.go
```diff
@@ -98,6 +99,10 @@ func (ju *jobUpdater) updateJob(index int) {
 	job := ju.jobQueue[index]
 	ssn := ju.ssn
 
+	if !conf.EnabledActionMap["enqueue"] && job.PodGroup.Status.Phase == scheduling.PodGroupPending {
```
william-wang, line 102: please add comments here or extract a small func for this special treatment

william-wang:
Any e2e for scheduler configuration without `enqueu`?

pkg/scheduler/framework/job_updater.go
wpeng102: add comments for this block

wpeng102:
> Any e2e for scheduler configuration without `enqueue`?

Add e2e case

william-wang (commented):
/lgtm

PR #2324, opened 2022-07-02
Ignore to reschedule big tasks

autumn0207:
Ignore to evict a task that requires too many resources, since it itself requires more than a threshold, it may be evicted on any other node.

### DIFF
--- a/pkg/scheduler/plugins/rescheduling/low_node_utilization.go
+++ b/pkg/scheduler/plugins/rescheduling/low_node_utilization.go
@@ -17,10 +17,11 @@ limitations under the License.
 package rescheduling
 
 import (
+	"reflect"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/rescheduling/low_node_utilization.go
```diff
@@ -202,3 +203,13 @@ func isContinueEvictPods(usage *NodeUtilization, totalAllocatableResource map[v1
 	}
 	return true
 }
+
+// resourceExceedsThrehold judges if resource percent exceeds the threhold.
+func resourceExceedsThrehold(reqPercent float64, rName v1.ResourceName, config interface{}) bool {
```
Thor-wl, line 208: Thanks for the optimization. It makes sense to deal with evicted pods with more cosideration such as resource request, priority and any other factors. I think it may be more reasonable to list all the conditions which should be give more notice and deal with them in a framework. So do you have some ideas or suggestions about that?

Issue #2328, opened 2022-07-04
Skip some wrong UTs related with allocation


Thor-wl:
**What happened**:
As what I noticed recently, there are some wrong UT cases related with allocate action which may block the CI frequently such as `TestAllocateWithDynamicPVC/node_changed_with_enough_resource`. One of the cases can be referred to [failure CI case](https://github.com/volcano-sh/volcano/runs/7145167489?check_suite_focus=true). 

**What you expected to happen**:
Remove the wrong UT cases

**How to reproduce it (as minimally and precisely as possible)**:
Just execute the CI when PRs are pushed.

**Anything else we need to know?**:
These UT cases are introduced from [this commit](https://github.com/volcano-sh/volcano/pull/1984/commits/5498e32c4ad9306c4ac1e41a060f831da8e54586)

**Environment**:
- Volcano Version: master branch
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #2329, opened 2022-07-04
remove wrong UT cases about allocate action

Thor-wl:
fix: https://github.com/volcano-sh/volcano/issues/2328 


william-wang:
@Thor-wl Would you add some description for that?

Thor-wl:
> @Thor-wl Would you add some description for that?

OK. I'll add more analysis here.

Issue #2321, opened 2022-07-01
lower priority job can not be preempted by high priority job when lower priority job enable gang


nkflash:
I have 2 user define queues and one default queue(default).

like:

low priority queue:

xbmin@p-ksyun-develop-001:/mnt/data/xbmin/kubeflow_test/config$ cat low.yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: low-queue
  namespace: kubeflow
spec:
  reclaimable: true
  weight: 1
  capability:
    cpu: "96"
    memory: "359Gi"
status:
  state: Open

high priority queue:
xbmin@p-ksyun-develop-001:/mnt/data/xbmin/kubeflow_test/config$ cat high.yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: high-queue
  namespace: kubeflow
spec:
  weight: 3
  reclaimable: true
  capability:
    cpu: "96"
    memory: "359Gi"
status:
  state: Open

There are only 96 core in cluster. And total queue weight is 5(low: 1, default: 1, high: 3)
I submit 2 lower priority jobs into low queue. Each job require 34 CPUs. So  total require 68 CPUs
Then I submit a higher job to high queue, which require 34 CPUs. Base on the "reclaim" action.
The high job should run and low jobs should be evicted, but I got the low job still occupy the resource. 
The high job can not schedule.

It seems the code forbid that. Only allocated resources count larger than MinAvailable job could be preempted.
![image](https://user-images.githubusercontent.com/42488903/176857434-b5053d8a-4171-4ce4-8412-15b3d557d292.png)

This is not reasonable, lower priority gang job also need to be preempted no matter if it satisfied by MinAvailable.
When I change that code, the reclaim work well.

Thor-wl:
Hi, thanks for your feedback. I reviewed the scenario you mentioned above. As the original design, the following points are important for users.
* Preempt only works for jobs within the same queue if users configure `preempt` action. In most scenarios, different jobs come from different projects or teams. It is not so reasonable to destory others' jobs.
* Reclaim works between different queues. It may trigger resource preemption from other queues(other projects or teams). But the preemption action promises the least harmless and works without priority comparsion between jobs from different queues.

nkflash:
> Hi, thanks for your feedback. I reviewed the scenario you mentioned above. As the original design, the following points are important for users.
> 
[...]

PR #2331, opened 2022-07-05
Support reclaim/preempt gang job which allocated resources equal or lesser than MinAvailable

nkflash:
@Thor-wl please check this design and implementation


nkflash:
fix https://github.com/volcano-sh/volcano/issues/2321

Thor-wl (commented):
Thanks for your optimiztion! Here are some suggestions as follows.
* Please separate the design doc from the implementation in 2 PRs. As the classical process of contribution, the community will review the desgin first. Once it is settled, the implementation follows.
* Perhaps we can add these design as part of `preemption` and `reclaim` desgin doc instead of an individual doc. In general, a new feature comes with a new desgin doc.
* It seems a little rough to just enable preemption and reclaim to evict pods from other jobs. A vicitim job with incomplete roles may not work well. For example, a victim spark job with only part of `executor` left. I think we should give more details about the evict strategy. For example, once a job is selected as victim, all of the pods or elastic pods **must** be evicted.
* Perhaps we can refer to the [elastic scheduler](https://github.com/volcano-sh/volcano/blob/master/docs/design/elastic-scheduler.md) and [TDM plugin](https://github.com/volcano-sh/volcano/blob/master/docs/design/tdm-plugin.md) as reference. This optimization can be implemented within the consideration with the 2 features mentioned above.
* Please execute `git commit --signoff xxx` to add personal signature. It is required by CI.
* Please execute `go fmt xxx` to make code verification happy.

Thor-wl:
> Thanks for your optimiztion! Here are some suggestions as follows.
> 
> * Please separate the design doc from the implementation in **2** PRs. As the classical process of contribution, the community will review the desgin first. Once it is settled, the implementation follows.
> * Perhaps we can add these design as part of `preemption` and `reclaim` desgin doc instead of an individual doc. In general, a new feature comes with a new desgin doc.
> * It seems a little rough to just enbale preemption and reclaim to evict pods from other jobs. A vicitim job with incomplete roles may not work well. For example, a victim spark job with only part of `executor` left. I think we should give more details about the evict strategy. For example, once a job is selected as victim, all of the pods or elastic pods **must** be evicted.
> * Perhaps we can refer to the [elastic scheduler](https://github.com/volcano-sh/volcano/blob/master/docs/design/elastic-scheduler.md) and [TDM plugin](https://github.com/volcano-sh/volcano/blob/master/docs/design/tdm-plugin.md) as reference. This optimization can be implemented in the consideration with the 2 features mentioned above.

Thor-wl:
@nkflash Please follow the code of conduct and format the code to pass the CI. Thanks.

nkflash:
> @nkflash Please follow the code of conduct and format the code to pass the CI. Thanks.

Sorry, I am busy with some other workload. I will finish this later.

PR #2339, opened 2022-07-07
add jobflow desgin docs

zhoumingcheng:
Signed-off-by: zhoumingcheng <zhoumingcheng@beyondcent.com>


example/jobflow/JobTemplate.yaml
```diff
@@ -0,0 +1,184 @@
+apiVersion: flow.volcano.sh/v1alpha1
+kind: JobTemplate
```
Thor-wl, line 2: Can you give some description about the difference between `vcjob` and `jobTemplate`? According to the example, it maybe a little bit redundant.

zhoumingcheng: The jobtemplate exists as a template for vcjob, the main reason is that directly creating a vcjob will be processed by the vcjob controller. Secondly, if the vcjob is used to directly add fields and is referenced by jobflow, the controller logic will be partially moved to the jobcontroller, and the vcjob does not support being referenced by multiple jobflows and needs to be created repeatedly. Using jobtemplate allows the same jobtemplate to be reused by multiple jobflows. And it supports field modification of the referenced jobtemplate directly in jobflow.

Thor-wl (approved):

Issue #2333, opened 2022-07-05
Change the way webhook is created from go code to yaml


hwdef:
#### What would you like to be added:

At present, the creation method of webhook is generated in go code. The purpose is to delete this part of the code and directly create a yaml to deploy the webhook. 

#### Why is this needed:

1. Creating and deleting volcano is more unified, the previous method cannot delete webhook when volcano is deleted
2. It is more convenient to modify the configuration of webhook, the previous method needs to modify the volcano code

#### Step

* Delete the go code that generates the webhook
* Create a yaml file to deploy the webhook
* Revert and close some relevant PR. https://github.com/volcano-sh/volcano/pull/2203 https://github.com/volcano-sh/volcano/pull/2290 https://github.com/volcano-sh/volcano/pull/2109 https://github.com/volcano-sh/volcano/pull/2251

PR #1645, opened 2021-07-26
enable configuring volcano webhooks

hacker-qian:
Signed-off-by: jianmin qian

PR #2346, opened 2022-07-11
deploy webhook by yaml

hwdef:
ref: https://github.com/volcano-sh/volcano/issues/2333

steps: 
* Remove the code that creates the webhook by go
* Create yaml files for deploying the webhook, set the certificate field blank.
* Supplementary certificate fields via volcano-admission

a little problem:
There will be a delay in supplementing the certificate fields through volcano-admission. During this period, volcano-scheduler will restart because it cannot connect to the webhook.

I don't know if this is a serious problem, please give me some advice.
@Thor-wl @william-wang @shinytang6 @qiankunli

### DIFF
--- a/cmd/webhook-manager/app/options/options.go
+++ b/cmd/webhook-manager/app/options/options.go
@@ -79,7 +79,7 @@ func (c *Config) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.WebhookNamespace, "webhook-namespace", "", "The namespace of this webhook")
 	fs.StringVar(&c.WebhookName, "webhook-service-name", "", "The name of this webhook")
 	fs.StringVar(&c.WebhookURL, "webhook-url", "", "The url of this webhook")
-	fs.StringVar(&c.EnabledAdmission, "enabled-admission", defaultEnabledAdmission, "enabled admission webhooks")
+	fs.StringVar(&c.EnabledAdmission, "enabled-admission", defaultEnabledAdmission, "enabled admission webhooks, if this parameter is modified, make sure corresponding webhook configurations are the same.")
 	fs.StringArrayVar(&c.SchedulerNames, "scheduler-name", []string{defaultSchedulerName}, "Volcano will handle pods whose .spec.SchedulerName is same as scheduler-name")
 	fs.StringVar(&c.ConfigPath, "admission-conf", "", "The configmap file of this webhook")
 	fs.StringVar(&c.IgnoredNamespaces, "ignored-namespaces", defaultIgnoredNamespaces, "Comma-separated list of namespaces to be ignored by admission webhooks")
--- a/cmd/webhook-manager/app/server.go
+++ b/cmd/webhook-manager/app/server.go
@@ -78,10 +78,12 @@ func Run(config *options.Config) error {
 
 		klog.V(3).Infof("Registered '%s' as webhook.", service.Path)
 		http.HandleFunc(service.Path, service.Handler)
-
-		klog.V(3).Infof("Registered configuration for webhook <%s>", service.Path)
-		registerWebhookConfig(kubeClient, config, service, config.CaCertData)
 	})
+
+	if err = addCaCertForWebhook(kubeClient, config.CaCertData); err != nil {
+		return fmt.Errorf("failed to add caCert for webhook %v", err)
+	}
+	klog.V(3).Infof("Successfully added caCert for all webhooks")
 
 	webhookServeError := make(chan struct{})
 	stopChannel := make(chan os.Signal, 1)
--- a/cmd/webhook-manager/app/util.go
+++ b/cmd/webhook-manager/app/util.go
@@ -17,93 +17,103 @@ limitations under the License.
 package app
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"regexp"
-	"strings"
+	"fmt"
+	"time"
 
 	v1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
 
 	"volcano.sh/apis/pkg/client/clientset/versioned"
 	"volcano.sh/volcano/cmd/webhook-manager/app/options"
-	"volcano.sh/volcano/pkg/webhooks/router"
 )
 
-func registerWebhookConfig(kubeClient *kubernetes.Clientset, config *options.Config, service *router.AdmissionService, caBundle []byte) {
-	sideEffect := v1.SideEffectClassNoneOnDryRun
-	reviewVersions := []string{"v1"}
-	webhookLabelSelector := &metav1.LabelSelector{}
-	clientConfig := v1.WebhookClientConfig{
-		CABundle: caBundle,
+var (
+	validatingWebhooksName = []string{
+		"volcano-admission-service-jobs-validate",
+		"volcano-admission-service-pods-validate",
+		"volcano-admission-service-queues-validate",
 	}
-	if config.WebhookURL != "" {
-		url := config.WebhookURL + service.Path
-		clientConfig.URL = &url
-		klog.Infof("The URL of webhook manager is <%s>.", url)
+	mutatingWebhooksName = []string{
+		"volcano-admission-service-pods-mutate",
+		"volcano-admission-service-queues-mutate",
+		"volcano-admission-service-podgroups-mutate",
+		"volcano-admission-service-jobs-mutate",
 	}
-	if config.WebhookName != "" && config.WebhookNamespace != "" {
-		clientConfig.Service = &v1.ServiceReference{
-			Name:      config.WebhookName,
-			Namespace: config.WebhookNamespace,
-			Path:      &service.Path,
-		}
-		klog.Infof("The service of webhook manager is <%s/%s/%s>.",
-			config.WebhookName, config.WebhookNamespace, service.Path)
-	}
-	if config.IgnoredNamespaces != "" {
-		ignoredNamespaces := strings.Split(strings.TrimSpace(config.IgnoredNamespaces), ",")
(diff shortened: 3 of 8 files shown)


cmd/webhook-manager/app/util.go
```diff
@@ -17,93 +17,95 @@ limitations under the License.
...
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"regexp"
-	"strings"
-
+	"fmt"
 	v1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"time"
```
Thor-wl, line 27: Please sort the packages regularlly.

hwdef: done

cmd/webhook-manager/app/util.go
```diff
@@ -17,93 +17,96 @@ limitations under the License.
...
-			service.MutatingConfig.Webhooks[i].ClientConfig = clientConfig
-			service.MutatingConfig.Webhooks[i].NamespaceSelector = webhookLabelSelector
-		}
+)
 
-		service.MutatingConfig.ObjectMeta.Name = webhookConfigName(config.WebhookName, service.Path)
+func addCaCertForWebhook(kubeClient *kubernetes.Clientset, caBundle []byte) error {
+	for _, mutatingWebhookName := range mutatingWebhooksName {
+		var mutatingWebhook *v1.MutatingWebhookConfiguration
+		if err := wait.PollInfinite(time.Second, func() (done bool, err error) {
```
shinytang6, line 57: it's better to add timeout and exit

shinytang6 (changes requested):

cmd/webhook-manager/app/util.go
hwdef: How long should the timeout be set?

hwdef: @shinytang6
I added a 5 minutes timeout.

cmd/webhook-manager/app/util.go
```diff
@@ -17,93 +17,96 @@ limitations under the License.
...
+		}
 
-		if err := registerMutateWebhook(kubeClient, service.MutatingConfig); err != nil {
-			klog.Errorf("Failed to register mutating admission webhook (%s): %v",
-				service.Path, err)
-		} else {
-			klog.V(3).Infof("Registered mutating webhook for path <%s>.", service.Path)
+		for index := 0; index < len(mutatingWebhook.Webhooks); index++ {
+			if mutatingWebhook.Webhooks[index].ClientConfig.CABundle == nil ||
+				bytes.Equal(mutatingWebhook.Webhooks[index].ClientConfig.CABundle, caBundle) {
```
shinytang6, line 73: `!bytes.Equal ` ?

cmd/webhook-manager/app/util.go
```diff
@@ -17,93 +17,96 @@ limitations under the License.
...
+			}
 		}
-	}
-	if service.ValidatingConfig != nil {
-		for i := range service.ValidatingConfig.Webhooks {
-			service.ValidatingConfig.Webhooks[i].SideEffects = &sideEffect
-			service.ValidatingConfig.Webhooks[i].AdmissionReviewVersions = reviewVersions
-			service.ValidatingConfig.Webhooks[i].ClientConfig = clientConfig
-			service.ValidatingConfig.Webhooks[i].NamespaceSelector = webhookLabelSelector
+		if _, err := kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Update(context.TODO(), mutatingWebhook, metav1.UpdateOptions{}); err != nil {
```
shinytang6, line 77: it's better to compare the updated webhook with original mutatingWebhook here, if no change, we don't need to call Update func

shinytang6 (changes requested):

shinytang6:
> During this period, volcano-scheduler will restart because it cannot connect to the webhook.
Will the scheduler connect to the webhook? l am a little confused here

cmd/webhook-manager/app/util.go
hwdef: Yes, you are right. I fixed it.

cmd/webhook-manager/app/util.go
hwdef: Ok, I added a flag to record if the webhook has been changed, and if so. then call the update function. reflection is not used, which may affect performance

cmd/webhook-manager/app/server.go
```diff
@@ -77,11 +77,13 @@ func Run(config *options.Config) error {
...
 		klog.V(3).Infof("Registered '%s' as webhook.", service.Path)
 		http.HandleFunc(service.Path, service.Handler)
-
-		klog.V(3).Infof("Registered configuration for webhook <%s>", service.Path)
-		registerWebhookConfig(kubeClient, config, service, config.CaCertData)
 	})
 
+	if err = addCaCertForWebhook(kubeClient, config.CaCertData); err != nil {
+		return fmt.Errorf("failed to add caCert for webhook %v", err)
+	}
```
shinytang6, line 84: The error seems to be that the webhook server needs to add cacert before it starts, but this function `addCaCertForWebhook` is a time-consuming operation, so scheduler reports the error `connect: connection refused`, so it's better to be done in a goroutine

william-wang:
> a little problem: There will be a delay in supplementing the certificate fields through volcano-admission. During this period, volcano-scheduler will restart because it cannot connect to the webhook.
> 
> I don't know if this is a serious problem, please give me some advice. @Thor-wl @william-wang @shinytang6 @qiankunli

@hwdef Yes, it is a serious problem. Woud you elaborate more on "volcano-scheduler restart because it can not connect to the webhook".  scheduler has no direct connection with webhook.

hwdef:
> > a little problem: There will be a delay in supplementing the certificate fields through volcano-admission. During this period, volcano-scheduler will restart because it cannot connect to the webhook.
> > I don't know if this is a serious problem, please give me some advice. @Thor-wl @william-wang @shinytang6 @qiankunli
> 
> @hwdef Yes, it is a serious problem. Woud you elaborate more on "volcano-scheduler restart because it can not connect to the webhook". scheduler has no direct connection with webhook.

After discussing it with @shinytang6 , we have roughly located the problem. I will try to solve it later.

hwdef:
The error mentioned above is caused by the failure to call the webhook when creating the initial queue. For this I use the wait package to retry the creation of the initial queue at a frequency of once per second. Until it is created successfully.
In addition, I also set a timeout of 5 minutes, and when a fatal problem is encountered, the program can be exited by panic. 
Not sure if such a solution would work.
@william-wang @shinytang6 What do you think?

hwdef:
@shinytang6 
I use `retry` package to create init queue, please review it again.
Thanks!

cmd/webhook-manager/app/util.go
```diff
@@ -17,93 +17,104 @@ limitations under the License.
...
-			service.MutatingConfig.Webhooks[i].NamespaceSelector = webhookLabelSelector
-		}
+)
 
-		service.MutatingConfig.ObjectMeta.Name = webhookConfigName(config.WebhookName, service.Path)
+func addCaCertForWebhook(kubeClient *kubernetes.Clientset, caBundle []byte) error {
+	for _, mutatingWebhookName := range mutatingWebhooksName {
+		var mutatingWebhook *v1.MutatingWebhookConfiguration
+		webhookChanged := false
+		if err := wait.Poll(time.Second, 5*time.Minute, func() (done bool, err error) {
```
shinytang6, line 58: also use `retry.OnError` here?

shinytang6 (changes requested):

cmd/webhook-manager/app/util.go
hwdef: I think the wait package has met the needs for now.Is there any other reason to use retry?

shinytang6: no, it's ok to use wait package

cmd/webhook-manager/app/util.go
```diff
@@ -17,93 +17,104 @@ limitations under the License.
...
 	"crypto/tls"
 	"crypto/x509"
-	"regexp"
-	"strings"
+	"fmt"
+	"time"
 
 	v1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
```
shinytang6, line 29: remove the line

shinytang6 (commented):
Another questions: due to we apply all the webhook configurations now, will `enabled-admission`  arguments still work as expected?

hwdef:
> Another questions: due to we apply all the webhook configurations now, will `enabled-admission` arguments still work as expected?

After checking the [source](https://github.com/volcano-sh/volcano/pull/1645#issuecomment-886458246) of this parameter and related code, I think that deploying webhook and this parameter are two separate functions that have no effect on each other

cmd/webhook-manager/app/util.go
hwdef: Thanks for your suggestion. I deleted this.

cmd/webhook-manager/app/options/options.go
```diff
@@ -79,7 +79,7 @@ func (c *Config) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.WebhookNamespace, "webhook-namespace", "", "The namespace of this webhook")
 	fs.StringVar(&c.WebhookName, "webhook-service-name", "", "The name of this webhook")
 	fs.StringVar(&c.WebhookURL, "webhook-url", "", "The url of this webhook")
-	fs.StringVar(&c.EnabledAdmission, "enabled-admission", defaultEnabledAdmission, "enabled admission webhooks")
+	fs.StringVar(&c.EnabledAdmission, "enabled-admission", defaultEnabledAdmission, "enabled admission webhooks,if this parameter is modified, please modify the deployed webhook as well. Make sure the same webhook is deployed and enabled.")
```
shinytang6, line 82: enabled admission webhooks, if this parameter is modified, make sure corresponding webhook configurations are the same.

shinytang6 (changes requested):

cmd/webhook-manager/app/options/options.go
hwdef: Thanks. I changed this.

hwdef:
@shinytang6 @william-wang @Thor-wl 
I think this PR is ready to review. Please take a look.

shinytang6:
Generally LGTM.

Thor-wl (approved):

Issue #2345, opened 2022-07-08
Delete a job with dependencies between task will cause the controller to report an error repeatedly


Megan96:
**What happened**:
Delete a job with dependencies between task will cause the controller to report an error repeatedly. The error is `Failed to get pod taskA-pod-name pod "taskA-pod-name" not found`
Then resumit the job, the job will lost the status.

There has actually been a completely diffrent job that also lost status when controller repeatedly reported this error, but I'm not sure if it's the same reason.

**What you expected to happen**:
job doesn't lose status and controller doesn't report error repeatedly.

**How to reproduce it (as minimally and precisely as possible)**:
1. submit a job with two tasks that are taskA and taskB,  taskB depends on taskA.
2.  podA of taskA has been created but can't run because of ImagePullBackOff error.
3. delete the job.
4. then resubmit a same job.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: 1.5.1
- Kubernetes version (use `kubectl version`): 1.20.15
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hwdef:
This is a bug. 
repeat: https://github.com/volcano-sh/volcano/issues/2284
Temporary workaround: delete volcano-controller pod

Issue #2284, opened 2022-06-08
Bug: Can not re-apply a same job when the old pods are Terminating


kongjibai:
**What happened**:
I use `kubectl apply -f hvd-job-torch-fairmot-2gpu.yaml` applied a job with 1 master pod(cpu task) and 2 worker pods(gpu task), and `kubectl get pod` can output the pods. After the job run, the pods status is **Running**. Then I use `kubectl delete -f hvd-job-torch-fairmot-2gpu.yaml` delete the job, the pods status are **Terminating**. Before all the pod deleted comletely, I re-apply the same job as before, then the old pods will be deleted with a few moment, but the new pods can not started. Although I delete the job and re-apply the same job, the pods can not start.

The output of `kubectl describe -f hvd-job-torch-fairmot-2gpu.yaml` as fllow,
```
Name:         lm-horovod-job-2gpu
Namespace:    default
Labels:       volcano.sh/job-type=Horovod
Annotations:  <none>
API Version:  batch.volcano.sh/v1alpha1
Kind:         Job
Metadata:
  Creation Timestamp:  2022-06-08T03:14:55Z
  Generation:          1
  Managed Fields:
    API Version:  batch.volcano.sh/v1alpha1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
        f:labels:
          .:
          f:volcano.sh/job-type:
      f:spec:
        .:
        f:minAvailable:
        f:plugins:
          .:
          f:ssh:
          f:svc:
        f:policies:
        f:schedulerName:
        f:tasks:
    Manager:      kubectl-client-side-apply
    Operation:    Update
    Time:         2022-06-08T03:14:55Z
    API Version:  batch.volcano.sh/v1alpha1
    Fields Type:  FieldsV1
    fieldsV1:
      f:status:
        .:
        f:conditions:
        f:controlledResources:
          .:
          f:plugin-ssh:
          f:plugin-svc:
        f:minAvailable:
        f:retryCount:
        f:state:
          .:
          f:lastTransitionTime:
          f:phase:
        f:taskStatusCount:
          .:
          f:master:
            .:
            f:phase:
              .:
              f:Unknown:
          f:worker:
            .:
            f:phase:
              .:
              f:Unknown:
        f:unknown:
        f:version:
    Manager:         vc-controller-manager
    Operation:       Update
[...]

Issue #2130, opened 2022-03-30
Task fail to schedule when tasks start with dependsOn and readinessProbe


Jason-Liu-Dream:
**What happened**:
Use volcano scheduler to apply vcjob(yaml as follows),  but found some parameters error, so delete this vcjob, modify parameters and reapply this yaml. But volcano scheduler fail to schedule this task again.
**What you expected to happen**:
volcano scheduler can schedule  task success.
**How to reproduce it (as minimally and precisely as possible)**:
Preparations: 
yaml
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: volcano-test-app-1518625680030433794
  namespace: volcano-test
spec:
  maxRetry: 3
  minAvailable: 1
  queue: default
  policies:
  - event: PodFailed
    action: RestartJob
  priorityClassName: scanner-extra-priority
  tasks:
    - replicas: 1
      name: first
      template:
        metadata:
          name: first
        spec:
          schedulerName: volcano
          containers:
          - name: test
            image: busybox:latest
            args:
            - /bin/sh
            - -c
            - sleep 60; touch /tmp/healthy; sleep 500
            readinessProbe:
              exec:
                command:
                - cat
                - /tmp/healthy
              initialDelaySeconds: 5
              periodSeconds: 5
          restartPolicy: Never
          securityContext: {}
          shareProcessNamespace: true
          terminationGracePeriodSeconds: 30
    - replicas: 1
      name: "job-nginx"
      template:
        metadata:
          name: nginx
        spec:
          nodeSelector:
            kubernetes.io/os: linux
          containers:
            - image: nginx
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
                requests:
                  cpu: "100m"
          restartPolicy: OnFailure
      dependsOn:
        name:
        - "first"
  ttlSecondsAfterFinished: 180
```
step 1:
Apply this job, make sure first pod start and second not start.
step 2:
Immediately delete this job.
step 3:
Apply this job again.
step 4:
use kubectl to see pod.

Contorller log:
```
I0330 03:24:13.262921       1 garbagecollector.go:109] Adding job volcano-test/volcano-test-app-1518625680030433794
I0330 03:24:13.263148       1 job_controller.go:309] Try to handle request <Queue: , Job: volcano-test/volcano-test-app-1518625680030433794, Task:, Event:OutOfSync, ExitCode:0, Action:, JobVersion: 0>
[...]

PR #2351, opened 2022-07-14
fix bug in task dependsOn

hwdef:
fix: https://github.com/volcano-sh/volcano/issues/2345 https://github.com/volcano-sh/volcano/issues/2284 https://github.com/volcano-sh/volcano/issues/2130

This PR can greatly reduce the chance of bug triggering.
As long as the user does not deploy a vcjob with the same name again in less than `--detection-period-of-dependson-task`, the bug will not be triggered.

The default value of `--detection-period-of-dependson-task` is 100 milliseconds.

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -511,6 +511,15 @@ func (cc *jobcontroller) isDependsOnPodsReady(task string, job *batch.Job) bool 
 	for _, podName := range dependsOnPods {
 		pod, err := cc.podLister.Pods(job.Namespace).Get(podName)
 		if err != nil {
+			// If pod is not found. There are 2 possibilities.
+			// 1. vcjob has been deleted. This function should return true.
+			// 2. pod is not created. This function should return false, continue waiting.
+			if apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -510,9 +510,25 @@ func (cc *jobcontroller) isDependsOnPodsReady(task string, job *batch.Job) bool
 	runningPodCount := 0
 	for _, podName := range dependsOnPods {
 		pod, err := cc.podLister.Pods(job.Namespace).Get(podName)
+		// If failed to get pod. There are 2 possibilities.
+		// 1. pod not found
+		// 2. Error communicating with apiserver
 		if err != nil {
-			klog.Errorf("Failed to get pod %v/%v %v", job.Namespace, podName, err)
-			continue
+			// If pod is not found. There are 2 possibilities.
```
Thor-wl, line 517: It's OK for me. Some small tip:
* The commments can be reorganized to be simple.
* The last `else` branch is redudant and can be removed.

hwdef: @Thor-wl 
Thanks for you suggestions. I made some changes.

Thor-wl (approved):

PR #2355, opened 2022-07-15
add volcano.sh/apis to staging dir

lucming:
solve #2356 
i just move the volcano.sh/apis this project to volcano's staging directory，and changed go.mod to let volcano use its own staging package. 
i have some other questions:
1. vendor dir should can deleted from volcano.sh/apis or not.
2. volcano.sh/apis have some files seems not belongs to it,can we put that to another dir,for example https://github.com/volcano-sh/apis/blob/master/pkg/apis/helpers/helpers.go this file,if you guys think it's okay, I can bring up a pr later to do this.
3. we are still missing a tool that can synchronize the staging directory to subprojects


hwdef (commented):
I think the synchronization tools are very important.

lucming:
> I think the synchronization tools are very important.

en,maybe we can refer to implementation of k8s,https://github.com/kubernetes/publishing-bot this project can push latest code of staging dir to it's subprojects

william-wang:
> > I think the synchronization tools are very important.
> 
> en,maybe we can refer to implementation of k8s,https://github.com/kubernetes/publishing-bot this project can push latest code of staging dir to it's subprojects

Automatic syschronization is the key thing. Please get familar with k8s staging firstly.

lucming:
> > > I think the synchronization tools are very important.
> > 
> > 
> > en,maybe we can refer to implementation of k8s,https://github.com/kubernetes/publishing-bot this project can push latest code of staging dir to it's subprojects
> 
> Automatic syschronization is the key thing. Please get familar with k8s staging firstly.

If we intend to do so, i will add the relevant configuration.

william-wang:
> If we intend to do so, i will add the relevant configuration.
We've been thinking about doing this for a long time but have no resource. Thanks for investigation, please continue working on this :). Here are some reference.
https://github.com/kubernetes/kubernetes/tree/master/staging
https://github.com/kubernetes/publishing-bot
https://github.com/kubernetes/test-infra

PR #2358, opened 2022-07-18
Managed and scheduled multiple single pods with podgroup

zbbkeepgoing:
For solve this situation

- When we use spark client model to submit spark job in k8s with volcano scheduler. hope all executors can be managed and scheduled through podgroup.

- hope some single pods can be managed and scheduled through podgroup with volcano.

### DIFF
--- a/docs/user-guide/how_to_schedule_multipe_single_pods_with_podgroup.md
+++ b/docs/user-guide/how_to_schedule_multipe_single_pods_with_podgroup.md
@@ -1,0 +1,103 @@
+# Manage & Schedule multiple single pods with podgroup
+
+## Background
+Volcano can manage single pod when the scheduler of pod is volcano, volcano controller will help to create a podgroup for this pod.
+But it can not manage and schedule multiple single pods in one pg now. And this missing ability is very useful in spark client mode and so on.
+It will also play a role in some scenarios that require multiple single pods to work together (vj cannot be used).
+
+## Key Points
+* Single pods need add two annotations, one is "scheduling.k8s.io/group-name", another is "scheduling.volcano.sh/group-min-resources". Please see the following example.
+* The controller will select the largest  minresource among single pods of the same "scheduling.k8s.io/group-name" annotation to add or update pg.
+* All single pods will be pg's OwnerReference.
+
+## Examples
+1. Create three single pods.
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: busybox1
+  annotations:
+    "scheduling.k8s.io/group-name": "busybox-pg"
+    "scheduling.volcano.sh/group-min-resources": "{\"cpu\":\"100m\",\"memory\":\"128Mi\"}"
+spec:
+  schedulerName: volcano
+  containers:
+    - name: busybox
+      image: busybox:1.28
+      command: ['sh', '-c', 'echo "Hello, busybox1!" && sleep 3600']
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: busybox2
+  annotations:
+    "scheduling.k8s.io/group-name": "busybox-pg"
+    "scheduling.volcano.sh/group-min-resources": "{\"cpu\":\"200m\",\"memory\":\"256Mi\"}"
+spec:
+  schedulerName: volcano
+  containers:
+    - name: busybox
+      image: busybox:1.28
+      command: ['sh', '-c', 'echo "Hello, busybox2!" && sleep 3600']
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: busybox3
+  annotations:
+    "scheduling.k8s.io/group-name": "busybox-pg"
+spec:
+  schedulerName: volcano
+  containers:
+    - name: busybox
+      image: busybox:1.28
+      command: ['sh', '-c', 'echo "Hello, busybox3!" && sleep 3600']
+```
+2. Kubectl describe pg.
+```
+Name:         busybox-pg
+Namespace:    default
+Labels:       <none>
+Annotations:  <none>
+API Version:  scheduling.volcano.sh/v1beta1
+Kind:         PodGroup
+Metadata:
+....
+  Owner References:
+    API Version:           v1
+    Block Owner Deletion:  true
+    Controller:            true
+    Kind:                  Pod
+    Name:                  busybox1
+    UID:                   aebfe6ed-3a4c-48f0-a38b-bc542ed262b4
+    API Version:           v1
+    Block Owner Deletion:  true
+    Controller:            false
+    Kind:                  Pod
+    Name:                  busybox2
+    UID:                   6262515d-1885-442a-b6e1-3878aa26f96a
+    API Version:           v1
+    Block Owner Deletion:  true
+    Controller:            false
+    Kind:                  Pod
+    Name:                  busybox3
+    UID:                   26c217d6-2d30-4282-8959-df0bfca9e07f
+....
+Spec:
+  Min Member:  1
+  Min Resources:
+    Cpu:     200m
+    Memory:  256Mi
+  Queue:     default
+Status:
+....
+Events:
+  Type    Reason     Age                 From     Message
+  ----    ------     ----                ----     -------
+  Normal  Scheduled  25s (x25 over 49s)  volcano  pod group is ready
+```
+
+## Limitations
+* Currently only "scheduling.k8s.io/group-name", "scheduling.volcano.sh/group-min-resources" and "scheduling.volcano.sh/queue-name" are supported.
+* Other properties under PodGroup's Spec cannot be injected through pod annotations.
--- a/pkg/webhooks/admission/pods/validate/admit_pod.go
+++ b/pkg/webhooks/admission/pods/validate/admit_pod.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"volcano.sh/apis/pkg/apis/helpers"
+
 	vcv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	commonutil "volcano.sh/volcano/pkg/util"
 	"volcano.sh/volcano/pkg/webhooks/router"
@@ -112,7 +113,7 @@ func validatePod(pod *v1.Pod, reviewResponse *admissionv1.AdmissionResponse) str
 		pgName = pod.Annotations[vcv1beta1.KubeGroupNameAnnotationKey]
 	}
 	if pgName != "" {
-		if err := checkPG(pod, pgName, true); err != nil {
+		if err := checkPG(pod, pgName, util.BelongToVcJob(pod)); err != nil {
 			msg = err.Error()
 			reviewResponse.Allowed = false
 		} else if err := checkPGQueueState(pod, pgName); err != nil {
--- a/pkg/controllers/podgroup/pg_controller.go
+++ b/pkg/controllers/podgroup/pg_controller.go
@@ -17,6 +17,8 @@ limitations under the License.
 package podgroup
 
 import (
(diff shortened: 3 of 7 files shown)


zbbkeepgoing:
Link to the api pull request: https://github.com/volcano-sh/apis/pull/83. I will update go.mod after pr of api is merged

Thor-wl:
Hey. Kindly to ask some questions about the background:
* Why do you want to separate the whole spark job to different and independent pods instead of making use of spark job or volcano job?
* I've not had a try about that. But as what I understand, Volcano has supported to generate PodGroup for separate pods. Is there something wrong with your test?

Thor-wl (commented):
Hey. Kindly to ask some questions about the background:
* Why do you want to separate the whole spark job to different and independent pods instead of making use of spark job or volcano job?
* I've not had a try about that. But as what I understand, Volcano has supported to generate PodGroup for separate pods. Is there something wrong with your test?

zbbkeepgoing:
> Hey. Kindly to ask some questions about the background:
> 
> * Why do you want to separate the whole spark job to different and independent pods instead of making use of spark job or volcano job?
> * I've not had a try about that. But as what I understand, Volcano has supported to generate PodGroup for separate pods. Is there something wrong with your test?

1、In Spark's Client submit mode. all the executor pod is independent, if we want to manage them by volcano, it need to group them in a ng. And a large part of the current spark users are also submitted using the client mode.
2、Yes, volcano support generate pg for separate pods, but not support pg's minresource not support, and this pr will support it. In addition, the generated podgroup only refer the first pod, if the first pod terminated, pg will be delete auto.

So the pr is support minresource for independent pods, and make pg refer to all the independent pods.

pkg/webhooks/admission/pods/validate/admit_pod.go
```diff
@@ -145,6 +146,15 @@ func validatePod(pod *v1.Pod, reviewResponse *admissionv1.AdmissionResponse) str
 	return msg
 }
 
+func isVcJob(pod *v1.Pod) bool {
```
Thor-wl, line 149: Suggest to rename to `BelongToVcJob`

zbbkeepgoing: done

pkg/webhooks/admission/pods/validate/admit_pod.go
```diff
@@ -145,6 +146,15 @@ func validatePod(pod *v1.Pod, reviewResponse *admissionv1.AdmissionResponse) str
 	return msg
 }
 
+func BelongToVcJob(pod *v1.Pod) bool {
```
hwdef, line 149: why make this public?

hwdef:
please compress multiple commits into a single commit.

pkg/webhooks/admission/pods/validate/admit_pod.go
zbbkeepgoing: I moved to utils.go, may it be used elsewhere in the future?

zbbkeepgoing:
> please compress multiple commits into a single commit.

done

zbbkeepgoing:
Can this pull request be merged into master? @Thor-wl

zbbkeepgoing:
@jiangkaihua Can this pull request be merged? We discussed this pr in the past week

jiangkaihua:
> @jiangkaihua Can this pull request be merged? We discussed this pr in the past week

Could you please help fix the code check? It seems that some UT cases did not go through.

zbbkeepgoing:
> > @jiangkaihua Can this pull request be merged? We discussed this pr in the past week
> 
> Could you please help fix the code check? It seems that some UT cases did not go through.

it's looks no problem, it shound be a ramdon exception, which has nothing to do with this pr, encountered before

Yikun (commented):
@zbbkeepgoing Thanks for contribution!

IIUC, this PR adds ability to help podgroup  auto-creation with minResource hint injection using pod annotation when first pod is created.

Currently, in Spark client, users need to create the podgroup outside (out of volcano), and specified annotation in executor pods.

By this PR:
- Advantages: Help user to create the podgroup with annotation inside volcano (rather than outside volcano).
- Disadvantages: the annotations style (especilly for complex json) might a lite way but not a good way in CRD world. If user want to use all volcano podgroup feature. We should also support like `minMember`, `priorityClassName`. After all features supporte, this might bring some comlex user interface to users.

And also after this PR the spark cluster mode can also use this ability to create driver podgroup, and assgin executors to executor podgroup.

So, if we can accept pod annotation as the API to create podgroup, then I +1 this PR. Otherwise, I am -0 (not a big disagree) on this.

(Think aloud: Does this mean that the PodGroup CRD is redundant in some level? Because annotation can help us do everything.)

also cc @william-wang

zbbkeepgoing:
> @zbbkeepgoing Thanks for contribution!
> 
> IIUC, this PR adds ability to help podgroup auto-creation with minResource hint injection using pod annotation when first pod is created.
> 
> Currently, in Spark client, users need to create the podgroup outside (out of volcano), and specified annotation in executor pods.
> 
> By this PR:
> 
> * Advantages: Help user to create the podgroup with annotation inside volcano (rather than outside volcano).
> * Disadvantages: the annotations style (especilly for complex json) might a lite way but not a good way in CRD world. If user want to use all volcano podgroup feature. We should also support like `minMember`, `priorityClassName`. After all features supporte, this might bring some comlex user interface to users.
> 
> And also after this PR the spark cluster mode can also use this ability to create driver podgroup, and assgin executors to executor podgroup.
> 
> So, if we can accept pod annotation as the API to create podgroup, then I +1 this PR. Otherwise, I am -0 (not a big disagree) on this.
> 
> (Think aloud: If it's accepted, does this mean that the PodGroup CRD is redundant in some level? Because annotation can help us do everything.)
> 
> also cc @william-wang

@william-wang Hi,leibo. Any questions or suggestions about this pr？

zbbkeepgoing:
@william-wang @Thor-wl 
If there is no doubt about this PR, can we merge it as soon as possible? This PR has been hanging for a long time. At present, I see other users in the Wechat group who also have this requirement.

wangyang0616:
I think there might be some issues here that are not very clear:

1. If the minresource information is injected into the podgroup through annotation, if the minresource information marked in each pod is different, how should it be handled? At the application level, should the minresource information of the first pod be directly selected, or should it be selected? What about the maximum minresource information of all pods under the podgroup?

3. If the minresource information of the pod changes, does Volcano need to process it?

5. The ownerreference information of the podgroup is automatically associated. For example, when an independent pod is created, the ownerreference information of the podgroup is the pod. If multiple pods belong to one replicas, the ownerreference information of the podgroup is the replicas. For the current scenario, multiple unreferenced The independent pods related to each other are placed in a podgroup. How to set the ownerreference information? If it is set to the first pod, then after the first pod is deleted directly, the current podgroup will be automatically deleted, and other pods may still be deleted. Is there a problem if the execution has not been completed or has not been scheduled?

Hope to get more discussion, thanks!

wangyang0616:
> 1. First, the minresource in annotations is bound to the pod group name in annotations, if user not specify pod group name, minresource annotation will not used. In addition, what is provided here is an extension capability for special but common scenarios. The strange behavior of users should not be considered. If i am a user, i need to use this ability normally, I should follow this rule.
> 2. Personally, I don’t think it needs to be considered for the time being, and the current Controller logic does not consider changing the podgroupname after the podgroup is automatically created by a single pod. In addition, I think that minresource‘s updated should be more important in Scheduler. At present, if the minresource of an inqueue podgroup is increased, and the resources of the queue are not enough at this time, should the podgroup be rolled back to pending? The above are some of my personal understanding. Same as the meaning of the first point, this is a solution in a special but common scenario. Of course, I do not reject implementing update minresource in the code if we believe that modifying minresource is a frequent and necessary requirement.
> 3. The current code is the ownerreference information of the podgroup is all pods specifying the pg.

Thanks for your reply, let me understand the PR further.

Regarding question 1, I understand that it is not a good choice to rely on the user's consciousness to deal with the abnormal scene of the function. We can guide the user to use the new function correctly through the manual or other methods, but for the use method that exceeds expectations, I understand that in It would be better to have a more comprehensive treatment on the platform side.

Question 2 is really not a high priority right now.

Regarding question 3, my previous understanding was biased, and I agree with your project development.

zbbkeepgoing:
> 1. strange

Q1： Agree with you. Even regardless of the controller, for relclaim and preempt action in the scheduler, choosing the maxinum minresource should be a better choice. I will update it late. Thank you

wangyang0616:
In addition, it would be better if the design document or user manual can be updated.

zbbkeepgoing:
> In addition, it would be better if the design document or user manual can be updated.

Yean, I add a user-guide doc. Please help to review it again.

docs/user-guide/how_to_schedule_multipe_single_pods_with_podgroup.md
```diff
@@ -0,0 +1,99 @@
+# Manage & Schedule multiple single pods with podgroup
+
+## Background
+Volcano can manage single pod when the scheduler of pod is volcano, volcano controller will help to create a podgroup for this pod.
+But it can not manage and schedule multiple single pods in one pg now. And this missing ability is very useful in spark client mode and so on.
+It will also play a role in some scenarios that require multiple single pods to work together (vj cannot be used).
+
+## Key Points
```
wangyang0616, line 8: In the current PR implementation, users are allowed to add MinResource resource information through annotation, but podgroup has many attributes, such as: MinMember, Queue, etc. The PR does not support the setting of these information. Can you add in-range and out-of-range in the document? Instructions are easier for users to understand.

wangyang0616:
In the scenario you describe, I agree to enhance the podgroup’s ability to manage a batch of independent pods, but I still have some concerns. MinResources can be configured through annotations. Will other attributes of the podgroup support this method in the future? Same as @Yikun  considered, if they are all introduced, then does the crd of podgroup still need to exist?

cc @william-wang @jinzhejz

docs/user-guide/how_to_schedule_multipe_single_pods_with_podgroup.md
zbbkeepgoing: I would add some limitation to the current

zbbkeepgoing:
> In the scenario you describe, I agree to enhance the podgroup’s ability to manage a batch of independent pods, but I still have some concerns. MinResources can be configured through annotations. Will other attributes of the podgroup support this method in the future? Same as @Yikun considered, if they are all introduced, then does the crd of podgroup still need to exist?
> 
> cc @william-wang @jinzhejz


For the former, I think there are suitable scenarios that we can add as needed, but for the latter, I hold different views. Because pg needs to exist as a group that manages pods, it is not just a simple CRD, it is the link between the scheduling units and the scheduling logic in the Volcano scheduler. I think this enhancement is just an increase in the PodGroup creation entry. 

For example, VJ is currently the most native management entrance for PodGroups, and it may cover some batch scenarios. "spark.kubernetes.scheduler.volcano.podGroupTemplateFile" is the entry point for Spark to manage PodGroup natively, and it covers some big data scenarios. The current enhancement is to manage the PodGroup's entrance through the PodGroup's Controller, which covers some special scenarios. 

Their existence has nothing to do with the PodGroup's CRD in nature, just different ways to creating and managing PodGroups. PodGroup still play a vital role in the scheduler. If we don't use PodGroup to play this link between Controller and Scheduler, just use Annotation to complete it. That is a challenge for the code complexity, scalability, and rationality of the Scheduler layer.

hwdef:
still needs to be reviewed

lowang-bh:
In my opition, we can create a vocano podgroup manully to handle those 3 pods. Then those pods will be contolled via the podgroup.

Issue #2364, opened 2022-07-19
when submit job, panic: resource is not sufficient to do  in scheduler/plugins/proportion.go


nulls-cell:
**What happened**:
When I submit the task, the volcano- Scheduler will throw "Panic: Resource is not sufficient", causing the program to exit
the detail:
`
panic: resource is not sufficient to do operation: <cpu 100000.00, memory 7000.00, nvidia.com/gpu 0.00> sub <cpu 120000.00, memory 3000.00>

goroutine 27 [running]:
volcano.sh/volcano/pkg/scheduler/util/assert.Assert(0xc3, {0xc000150090, 0xc0007798a8})
	github/volcano/pkg/scheduler/util/assert/assert.go:33 +0xfc
volcano.sh/volcano/pkg/scheduler/util/assert.Assertf(0x0, {0x56293c3, 0x0}, {0xc0007798a8, 0xc0001a4008, 0x55fa6e7})
	github/volcano/pkg/scheduler/util/assert/assert.go:43 +0x56
volcano.sh/volcano/pkg/scheduler/api.(*Resource).Sub(0xc00068e680, 0xc00068e6c0)
	github/volcano/pkg/scheduler/api/resource_info.go:232 +0x85
volcano.sh/volcano/pkg/scheduler/plugins/proportion.(*proportionPlugin).OnSessionOpen(0xc00068e100, 0xc0007241c0)
	github/volcano/pkg/scheduler/plugins/proportion/proportion.go:244 +0xe65
volcano.sh/volcano/pkg/scheduler/framework.OpenSession({0x575c2d8, 0xc000754480}, {0xc000302030, 0x1, 0x1}, {0x0, 0x0, 0x0})
	github/volcano/pkg/scheduler/framework/framework.go:43 +0x27c
volcano.sh/volcano/pkg/scheduler/plugins/proportion.TestProportionPanic.func3()
	github/volcano/pkg/scheduler/plugins/proportion/proportion_test.go:306 +0x15e
created by volcano.sh/volcano/pkg/scheduler/plugins/proportion.TestProportionPanic
	github/volcano/pkg/scheduler/plugins/proportion/proportion_test.go:302 +0x2015

Process finished with the exit code 1
`

**What you expected to happen**:
The program can run smoothly.

**How to reproduce it (as minimally and precisely as possible)**:
[...]

PR #2365, opened 2022-07-20
Proportion panic bugfix

hansongChina:
#2364 Proportion panic bugfix

as issue #2364 describe, when increasedDeserved is Is greater than sum of remaining and decreasedDeserved, Sub is panic.

That's because the requested resources plus the guarantee resources are greater than the total cluster resources.
In this case, there are no more resources to allocate to the queue's deserved resource


hwdef (commented):
/lgtm
Thanks!

hansongChina:
look #2367

PR #2367, opened 2022-07-20
proportion remaining resource sub panic bugfix

hansongChina:
#2364 

as issue #2364  describe, when increasedDeserved is Is greater than sum of remaining and decreasedDeserved, Sub is panic.

That's because the requested resources plus the guarantee resources are greater than the total cluster resources.
In this case, there are no more resources to allocate to the queue's deserved resource

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -239,7 +239,13 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			metrics.UpdateQueueDeserved(attr.name, attr.deserved.MilliCPU, attr.deserved.Memory)
 		}
 
-		remaining.Sub(increasedDeserved).Add(decreasedDeserved)
+		remaining.Add(decreasedDeserved)
+		if increasedDeserved.LessEqual(remaining, api.Zero) {
+			remaining.Sub(increasedDeserved)
+		} else {
+			klog.V(4).Infof("Exiting when increasedDeserved <%v> is greater than remaining <%v>", increasedDeserved, remaining)
+			break
+		}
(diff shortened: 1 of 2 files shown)


hwdef (commented):
/lgtm

Thor-wl:
/hold
@qiankunli I tried to simulate the bug scenario according to the proportion plugin implementation. I found that the root reason seems to be [](https://github.com/volcano-sh/volcano/blob/5f0a540eeb8da669ebaf3e3210f88e355e97ed1b/pkg/scheduler/plugins/proportion/proportion.go#L228). We should rethink the relationship within the queue.attr. Please help take a look at that. Thanks!

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -239,7 +239,13 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			metrics.UpdateQueueDeserved(attr.name, attr.deserved.MilliCPU, attr.deserved.Memory)
 		}
 
-		remaining.Sub(increasedDeserved).Add(decreasedDeserved)
+		remaining.Add(decreasedDeserved)
+		if increasedDeserved.LessEqual(remaining, api.Zero) {
+			remaining.Sub(increasedDeserved)
+		} else {
+			klog.V(4).Infof("Exiting when increasedDeserved <%v> is greater than remaining <%v>", increasedDeserved, remaining)
```
Thor-wl, line 246: Suggest to reset some kind of resouce within`remaining` to be `0` is the result is a negative value . Because it has added `decreasedDeserved` above left. When jump out of the judgement, it will judge the following condition `remaining.IsEmpty()` below, which may cause wrong result.

PR #2378, opened 2022-07-25
add rest config into ssn

wpeng102:

Customer maybe introduce CRD resource for scheduling. So in scheduler plugin,  customer maybe need construct this CRD's informer. Currently ssn only provide kubeclient for plugin, we need provide restconfig for customer to init customer resource client.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -92,6 +92,7 @@ type SchedulerCache struct {
 	sync.Mutex
 
 	kubeClient   *kubernetes.Clientset
+	restConfig   *rest.Config
 	vcClient     *vcclient.Clientset
 	defaultQueue string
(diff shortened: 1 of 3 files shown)


Thor-wl:
Sound reasonable.That's OK for me.
/lgtm

elinx:
The plugins are created and destroyed each scheduling cycle,  do you recreate the informers every time?

wpeng102:
> The plugins are created and destroyed each scheduling cycle, do you recreate the informers every time?
> 
> If it's not, do you think it's a good option to put these informer creations somewhere before the session starts?

Good question. Currently volcano plugin does not provide a  informer  function which can init the informer. I think customer can use a singleton function to init some variable to avoid recreate every scheduler period.

william-wang (approved):
/approve

PR #2399, opened 2022-07-30
get csinodes from ssn

wpeng102:

Add `CSINodeStatusInfo` into ssn for plugin to use.


william-wang (commented):
/lgtm

Issue #2389, opened 2022-07-28
Release-1.6 Volcano Scheduler can't schedule all resource


compactcfusr:
**What happened**:
When we upgraded volcano from 1.4 to 1.6, we found PendingJob increase.
At first we thought it was volcano that limited the scheduling speed, and restarting took a long time to recover.
But we found Pending Jobs not reduced two hours later.
So we have to rollback volcano from 1.6 to 1.4
After rollback we found dirty pg leaded to inqueue pg backlog  because pg in 1.6 attach uuid suffix.We cleared the dirty and Volcano recovered. 

Duration two hours, cpu and memory usage like following:
CPU Usage:
![image](https://user-images.githubusercontent.com/15198859/181462223-392cc6fb-ef0d-44ca-9125-354feacaf7e9.png)
Memory Usage
![image](https://user-images.githubusercontent.com/15198859/181462345-c927e442-01c5-4c67-b85a-4baff41ea657.png)

It seems like Volcano Scheduler will not schedule when usage reach 60%.

**What you expected to happen**:

We hope CPU and Memory usage can reach 100%

**How to reproduce it (as minimally and precisely as possible)**:
we still not reproduce this scene, only know upgraded volcano from 1.4 to 1.6

**Anything else we need to know?**:

**Environment**:
- Volcano Version:  release-1.4
- Kubernetes version (use `kubectl version`): v1.20.10
- Cloud provider or hardware configuration: 
- OS (e.g. from /etc/os-release): ubuntu 16.04
- Kernel (e.g. `uname -a`): Linux 4.15.0-42
- Install tools: tekton 
- Others:

Thor-wl:
@compactcfusr Please provide your scheudler configuration.

compactcfusr:
> 

**volcano-admission-configmap**
 
- apiVersion: v1
  data:
    volcano-admission.conf: |
      #resourceGroups:
      #- resourceGroup: management                    # set the resource group name
      #  object:
      #    key: namespace                             # set the field and the value to be matched
      #    value:
      #    - mng-ns-1
      #  schedulerName: default-scheduler             # set the scheduler for patching
      #  tolerations:                                 # set the tolerations for patching
      #  - effect: NoSchedule
      #    key: taint
      #    operator: Exists
      #  labels:
      #    volcano.sh/nodetype: management           # set the nodeSelector for patching
      #- resourceGroup: cpu
      #  object:
      #    key: annotation
      #    value:
[...]

PR #2400, opened 2022-07-30
try get get old pg when new pg not exist

Akiqqqqqqq:
Signed-off-by: Qiuyu Wu <qiuyu.wu@shopee.com>

fix：https://github.com/volcano-sh/volcano/issues/2389

what happen:
after this feature https://github.com/volcano-sh/volcano/pull/2140 applied, error comes when upgrading volcano from 1.4 to 1.6, details showed in https://github.com/volcano-sh/volcano/issues/2389.

this is because volcano-job-controller creates extra pgs with job uid suffix for existing jobs, which are already running, and thus two pgs exist in cluster for one vj. when existing jobs are using over half of the cluster resource already, the new pgs with uid will cause inqueue get larger in proportion AddJobEnqueueableFn (line 334 in pkg/scheduler/plugins/proportion/proportion.go). after inqueue gets larger than realCapability, the new coming jobs will pend forever

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -650,7 +650,7 @@ func (cc *jobcontroller) createOrUpdatePodGroup(job *batch.Job) error {
 				job.Namespace, job.Name, err)
 			return err
 		} else {
-			// try to get old pg
+			// try to get old pg if new pg not exist
 			pg, err = cc.pgLister.PodGroups(job.Namespace).Get(job.Name)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


Thor-wl:
Thanks for your contribution. I think it is important for users who upgrades volcano versions from below v1.6.0 to v1.6.0.

hzxuzhonghu (commented):
Overall lgtm, just one comment i think the PG construction code can be resued, maybe abstract a `buildPodGroup` function

william-wang (approved):
/approve

PR #2408, opened 2022-08-04
add: update scheduler config by watching scheduler-configmap

### LINKED ISSUE #2332: Make volcano-scheduler-configmap modification take effect asap
watch the `volcano-scheduler-configmap` api-resource instead of the volume files(mounted by k8s configmap) to make user's modification take effects asap.

#### Why is this needed:
The current implementation use file watcher to notify the scheduler about user's configuration modification, but it's sort of slow like taking at least 1 minutes before k8s making the volume mounting really happend, as the doc says:
> As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the pod can be as long as kubelet sync period (1 minute by default) + ttl of ConfigMaps cache (1 minute by default) in kubelet.

chanhz:
resolved #2332 . 
Signed-off-by: chanhz <f13ichanhz@gmail.com>

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -47,6 +47,7 @@ type ServerOption struct {
 	KubeClientOptions    kube.ClientOptions
 	SchedulerNames       []string
 	SchedulerConf        string
+	SchedulerConfigmap   string
 	SchedulePeriod       time.Duration
 	EnableLeaderElection bool
 	LockObjectNamespace  string
@@ -85,6 +86,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	// volcano scheduler will ignore pods with scheduler names other than specified with the option
 	fs.StringArrayVar(&s.SchedulerNames, "scheduler-name", []string{defaultSchedulerName}, "vc-scheduler will handle pods whose .spec.SchedulerName is same as scheduler-name")
 	fs.StringVar(&s.SchedulerConf, "scheduler-conf", "", "The absolute path of scheduler configuration file")
+	fs.StringVar(&s.SchedulerConfigmap, "scheduler-configmap", "volcano-scheduler-configmap ", "The configmap of scheduler configuration")
 	fs.DurationVar(&s.SchedulePeriod, "schedule-period", defaultSchedulerPeriod, "The period between each scheduling cycle")
 	fs.StringVar(&s.DefaultQueue, "default-queue", defaultQueue, "The default queue name of the job")
 	fs.BoolVar(&s.EnableLeaderElection, "leader-elect", s.EnableLeaderElection,
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -22,9 +22,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/fsnotify/fsnotify"
+	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
 	"volcano.sh/volcano/pkg/filewatcher"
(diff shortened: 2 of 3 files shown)


cmd/scheduler/app/options/options.go
```diff
@@ -85,6 +86,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	// volcano scheduler will ignore pods with scheduler names other than specified with the option
 	fs.StringArrayVar(&s.SchedulerNames, "scheduler-name", []string{defaultSchedulerName}, "vc-scheduler will handle pods whose .spec.SchedulerName is same as scheduler-name")
 	fs.StringVar(&s.SchedulerConf, "scheduler-conf", "", "The absolute path of scheduler configuration file")
+	fs.StringVar(&s.SchedulerConfigmap, "scheduler-configmap", "volcano-scheduler-configmap ", "The configmap of scheduler configuration")
```
Thor-wl, line 89: So what's the difference bewteen `SchedulerConf` and `SchedulerConfigmap`?  Besides, how about that user remove the configmap by fault?

chanhz: - `SchedulerConf`is the file which kubelet mount from `SchedulerConfigmap`, their contents are same. But the difference is that watching `SchedulerConfigmap` change by informer will make configuration modification take effect asap, comparing to filewatch. As  # 2332 mentioned,
> As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the pod can be as long as kubelet sync period (1 minute by default) + ttl of ConfigMaps cache (1 minute by default) in kubelet.

- If user remove the configmap by fault, then scheduler will use the default or the old configuration. IMO, it's no risks.

Thor-wl: > * `SchedulerConf`is the file which kubelet mount from `SchedulerConfigmap`, their contents are same. But the difference is that watching `SchedulerConfigmap` change by informer will make configuration modification take effect asap, comparing to filewatch. As  # 2332 mentioned,

IC. If that, I think this parameter is not necessary. Let's keep the `schedulerConf`. And users should not configure twice.




> * If user remove the configmap by fault, then scheduler will use the default or the old configuration. IMO, it's no risks.

OK. That's reasonable.

kerthcet: Maybe we should follow a deprecated policy here, keep both `schedulerConf` and `schedulerConfigmap`, then we should:
1) add warning messages to `schedulerConf` to notify end-users that it's deprecated and will be removed in the future
2) add validations that `schedulerConf` and `schedulerConfigmap` can't be enabled at the same time
3) choose the implement by the flag set by end-user, if so, we should keep both the implements

This is backwards compatible and will not break end-users.

pkg/scheduler/scheduler.go
```diff
@@ -155,28 +159,28 @@ func (pc *Scheduler) loadSchedulerConf() {
...
-			}
-		case err, ok := <-errCh:
-			if !ok {
-				return
-			}
-			klog.Infof("watch %s error: %v", pc.schedulerConf, err)
-		case <-stopCh:
-			return
-		}
+	if cm.GetNamespace() != "volcano-system" && cm.GetName() != pc.schedulerConfigmap {
```
elinx, line 178: hard coded "volcano-system" may not a good idea, users may use other ns like `kube-system`

pkg/scheduler/scheduler.go
```diff
@@ -155,28 +159,28 @@ func (pc *Scheduler) loadSchedulerConf() {
...
+	cmInformer := informerFactory.Core().V1().ConfigMaps()
+	cmInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			UpdateFunc: pc.updateConfigMap,
+		})
+	informerFactory.Start(stopCh)
+}
+
+// updateConfigMap update scheduler config
+func (pc *Scheduler) updateConfigMap(_, newObj interface{}) {
```
elinx, line 172: is it possible to emit an error message for the user if they try to use a configmap but not specify a right one?

PR #2410, opened 2022-08-05
fix scheduler panic when webhook is not ready

hwdef:
when admission is not ready, the scheduler will panic when create init queue.I add some code retry to create init queue.


wpeng102 (commented):
/lgtm

william-wang (approved):
/approve

PR #2441, opened 2022-08-16
Fix calculating available gpu num error

kerthcet:
/kind bug

The problem here is we'll always return devIDs with length equals to 1, but actually I think `predicateGPUbyMemory` wants to return all the available devIDs.
https://github.com/volcano-sh/volcano/blob/52a34f5ff10e3e37bd8c407ef9b559dd81dff3fb/pkg/scheduler/plugins/predicates/gpu.go#L63-L66

### DIFF
--- a/pkg/scheduler/plugins/predicates/gpu.go
+++ b/pkg/scheduler/plugins/predicates/gpu.go
@@ -32,7 +32,7 @@ func checkNodeGPUSharingPredicate(pod *v1.Pod, nodeInfo *api.NodeInfo) (bool, er
 		return true, nil
 	}
 	ids := predicateGPUbyMemory(pod, nodeInfo)
-	if ids == nil {
+	if len(ids) == 0 {
 		return false, fmt.Errorf("no enough gpu memory on node %s", nodeInfo.Name)
 	}
 	return true, nil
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/predicates/gpu.go
```diff
@@ -50,44 +50,30 @@ func checkNodeGPUNumberPredicate(pod *v1.Pod, nodeInfo *api.NodeInfo) (bool, err
...
 
-	for devID := 0; devID < len(allocatableGPUs); devID++ {
-		devIDs = append(devIDs, allocatableGPUs[devID])
-		if len(devIDs) == gpuRequest {
-			return devIDs
-		}
-	}
-
-	return nil
+	return allocatableGPUs[:gpuRequest]
```
kerthcet, line 78: Another concern here is should we always return the idle gpus, what about including the allocable gpus.

Thor-wl (commented):
Thanks for the fix. Can you submit an issue about the details in order to make it easy for reviewers to have a full glimpse of the scenario?

kerthcet:
> Thanks for the fix. Can you submit an issue about the details in order to make it easy for reviewers to have a full glimpse of the scenario?

Updated in the description.

Thor-wl (approved):

Thor-wl:
> Updated in the description.

IC, it's just an optimization.

pkg/scheduler/plugins/predicates/gpu.go
kerthcet: @Thor-wl Is this make sense? Actually I haven't seen the original design intension.

PR #2443, opened 2022-08-17
fix performance downgrade issue

wpeng102:
Set `inqueue` in closed session will lead performance downgrade. move it to `allocate` action.


william-wang (approved):
/lgtm

PR #2463, opened 2022-08-26
Update CA insert method in webhooks.

jiangkaihua:
Current method `addCaCertForWebhook()` only added CA cert for webhooks listed in `validatingWebhooksName` and `mutatingWebhooksName`
https://github.com/volcano-sh/volcano/blob/74b211451dbc16618bfd2d51a7288f0c0268551f/cmd/webhook-manager/app/util.go#L39-L51
Which is inconvenient for new webhook registration. For example, when user registered new webhook by arg `--enabled-admission`, new webhook would not get certs since the original method only go through slices above.
So this pr make `addCaCertForWebhook()` to go through all webhooks registered in volcano admission by arg `--enabled-admission`.

#2464 


william-wang (approved):
/lgtm

PR #2475, opened 2022-08-31
fix gpu shareing predictor allocate more than one gpu for pod

WingkaiHo:
Signed-off-by: yougjiahe <yongjiahe@tuputech.com>

Predictor patch all suitable gpu ids which return from predicateGPUbyMemory to pod.  But pod only need one of suitable id of gpu. Fix https://github.com/volcano-sh/volcano/issues/2491


shinytang6 (approved):
/approve

Issue #2481, opened 2022-09-01
volcano scheduler crash when pv is not created


jinzhejz:
**What happened**:
volcano scheduer crash
```
E0901 08:33:36.243987       1 cache.go:995] task default/zjin-lazy-0 bind Volumes failed: &fmt.wrapError{msg:"binding volumes: timed out waiting for the condition", err:(*errors.errorString)(0xc0001991c0)}
I0901 08:33:36.244016       1 cache.go:299] Revert assumed volumes for task default/zjin-lazy-0 on node zjin-vm
E0901 08:33:36.244110       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 156 [running]:
volcano.sh/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1a08e00, 0x2ff5020})
        /home/zjin/gopath/src/volcano.sh/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x7d
volcano.sh/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xa})
        /home/zjin/gopath/src/volcano.sh/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x75
panic({0x1a08e00, 0x2ff5020})
        /home/zjin/go/1.17.6/src/runtime/panic.go:1038 +0x215
volcano.sh/volcano/vendor/k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumebinding.(*volumeBinder).BindPodVolumes(0xc000446210, 0xc000180800, 0x0)
        /home/zjin/gopath/src/volcano.sh/volcano/vendor/k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumebinding/binder.go:453 +0x25c
volcano.sh/volcano/pkg/scheduler/cache.(*defaultVolumeBinder).BindVolumes(0xc00042e360, 0x1b50440, 0xc0002280a0)
        /home/zjin/gopath/src/volcano.sh/volcano/pkg/scheduler/cache/cache.go:338 +0x3c
volcano.sh/volcano/pkg/scheduler/cache.(*SchedulerCache).BindTask(0xc00050a000)
[...]

PR #2483, opened 2022-09-01
fix scheduler panic issue when pv is not created

jinzhejz:
Signed-off-by: Zhe Jin <jinzhe.hit@gmail.com>

fix https://github.com/volcano-sh/volcano/issues/2481


william-wang (approved):
/lgtm

PR #2504, opened 2022-09-23
fix printing volcano-scheduler-configmap loading log;

waiterQ:
action/plugin register log printing won't work because initializing klog is after it is called at register plugin/action, so add more detail print at successfully loaded scheduler configuration.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -123,7 +123,7 @@ func (pc *Scheduler) loadSchedulerConf() {
 	klog.V(4).Infof("Start loadSchedulerConf ...")
 	defer func() {
 		actions, plugins := pc.getSchedulerConf()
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -123,7 +124,15 @@ func (pc *Scheduler) loadSchedulerConf() {
 	klog.V(4).Infof("Start loadSchedulerConf ...")
 	defer func() {
 		actions, plugins := pc.getSchedulerConf()
-		klog.V(4).Infof("Successfully loaded scheduler conf, actions: %v, plugins: %v", actions, plugins)
+		confBytes, err := yaml.Marshal(map[string]interface{}{
```
william-wang, line 127: let's make it simple. Do not think it's a good idea to expose all parameters in log for user.

waiterQ: done. just print action, plugin name

william-wang (approved):
/lgtm

william-wang (approved):
/lgtm

PR #2509, opened 2022-09-27
bugfix: the csinode driver status failed updating

### LINKED ISSUE #2508: scheduler should not schedule when csinode driver restarts
**What happened**:
the csi plugin driver restarts but the csinode status not updated as what expected and the scheduler schedule a pod to the node even when the max volume count exceeds it's capacity

update the csi node driver status correctly

**Environment**: linux
- Volcano Version: 1.6.1
- Kubernetes version (use `kubectl version`):1.19
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

elinx:
fix #2508 the existing implementation still keeps the deleted driver
status as true, this pr fix it.

Signed-off-by: xilinxing <xilinxing@huawei.com>

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -411,24 +411,18 @@ func (sc *SchedulerCache) AddOrUpdateCSINode(obj interface{}) {
 		return
 	}
 
-	var csiNodeStatus *schedulingapi.CSINodeStatusInfo
-	var found bool
+	csiNodeStatus := &schedulingapi.CSINodeStatusInfo{
+		CSINodeName:  csiNode.Name,
+		DriverStatus: make(map[string]bool),
+	}
 	sc.Mutex.Lock()
 	defer sc.Mutex.Unlock()
-	// update nodeVolumeCount
-
-	if csiNodeStatus, found = sc.CSINodesStatus[csiNode.Name]; !found {
-		csiNodeStatus = &schedulingapi.CSINodeStatusInfo{
-			CSINodeName:  csiNode.Name,
-			DriverStatus: make(map[string]bool),
-		}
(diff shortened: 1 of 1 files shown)


pkg/scheduler/cache/event_handlers.go
```diff
@@ -411,24 +411,18 @@ func (sc *SchedulerCache) AddOrUpdateCSINode(obj interface{}) {
...
-			DriverStatus: make(map[string]bool),
-		}
-		sc.CSINodesStatus[csiNode.Name] = csiNodeStatus
-	}
 
 	for i := range csiNode.Spec.Drivers {
 		d := csiNode.Spec.Drivers[i]
 		csiNodeStatus.DriverStatus[d.Name] = d.Allocatable != nil && d.Allocatable.Count != nil
 	}
+	sc.CSINodesStatus[csiNode.Name] = csiNodeStatus
```
william-wang, line 425: if csiNodeStatus exists in sc.CSINodesStatus[csiNode.Name], we still create a new csiNodeStatus?

elinx: Yes, the old status should be override completely because we don't know which driver pod is dead or recreated, otherwise there may left some non existent driver status as true.

jiangkaihua: Could we just remove the nonexistent drivers in existing `csiNodeStatus`, like:

```go
	for _, driver := range csiNode.Spec.Drivers {
		if _, exist := csiNodeStatus.DriverStatus[driver.Name]; !exist {
			delete(csiNodeStatus.DriverStatus, driver.Name)
			continue
		}
		csiNodeStatus.DriverStatus[driver.Name] = driver.Allocatable != nil && driver.Allocatable.Count != nil
	}
```

william-wang (approved):
/lgtm

PR #2515, opened 2022-09-28
delete task qos limit in webhook

waiterQ:
delete task qos limit in webhook


william-wang:
@waiterQ As far as I know some users enhance the k8s cpu topology and numa topology ability in their product. They support cpu binding even when the it is not a guranteed pod. Is your case the same situation?

waiterQ:
> @waiterQ As far as I know some users enhance the k8s cpu topology and numa topology ability in their product. They support cpu binding even when the it is not a guranteed pod. Is your case the same situation?

yes, we have same situation

william-wang (approved):
/lgtm

PR #2529, opened 2022-10-11
Refactor node order plugin

kerthcet:
Signed-off-by: kerthcet <kerthcet@gmail.com>

When I want to add a new kubernetes plugin, I found all the related plugins piled on the single
nodeorder.go file, it's hard to maintain especially we have more plugins to introduce in the future. 
So I refactored these codes by splitting them into several files.

Special notes: no changing to the logics.


Thor-wl (commented):
It's OK for me overall for it is more clear for developers.

hwdef:
I think this pr is useful.
/reopen

hwdef:
@kerthcet Can you rebase the code and fix CI bugs?

PR #2531, opened 2022-10-11
Volcano adapts to the k8s v1.25

wangyang0616:
Volcano adapts to the k8s v1.25


hwdef (commented):
a little confused, does this PR contain content other than k8s 1.25 adaptation

PR #2540, opened 2022-10-17
Fix scheduler plugin arguments readin bug.

### LINKED ISSUE #2539: Scheduler plugin arguments failed to readin "1" as float64.
**What happened**:

When I set `overcommit ` plugin argument: `overcommit-factor` to '2' in configmap, like:

configmap

And then volcano scheduler threw out a warning, and my configuration did not take effect.

scheduler-log

I think the question was probably caused by type assertions in plugin argument interface:

https://github.com/volcano-sh/volcano/blob/59e58c23957821601bc8e27222e4721c9c323722/pkg/scheduler/framework/arguments.go#L59

Current method could not convert `int` to `float64`.

Interger value can be readin as well.

As below, and would also happen on other plugins.

**Environment**:
- Volcano Version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jiangkaihua:
fix #2539 

### DIFF
--- a/pkg/scheduler/framework/arguments.go
+++ b/pkg/scheduler/framework/arguments.go
@@ -58,6 +58,10 @@ func (a Arguments) GetFloat64(ptr *float64, key string) {
 
 	value, ok := argv.(float64)
 	if !ok {
+		if intVal, ok := argv.(int); ok {
+			*ptr = float64(intVal)
+			return
+		}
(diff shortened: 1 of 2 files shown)


jiangkaihua:
> Could you please add a UT case for this?

Added.

william-wang (commented):
/lgtm

PR #2543, opened 2022-10-19
fix 2496 base on latest master

zhifanggao:
Signed-off-by: zhifanggao <28681649@qq.com>


zhifanggao:
@jiangkaihua, Thanks for you comments, I will fix asap.

zhifanggao:
@jiangkaihua Can it add a unittest for imagelocality policy, It is only a policy for scoring a host in the nodeorder plugin, there are many policies in the nodeorder plugin. I checked the e2e case in the test/ete folder ,It looks like that  e2e cases for a plugin are included

jiangkaihua:
> @jiangkaihua Can I add a unittest for imagelocality policy, It is only a policy for scoring a host in the nodeorder plugin, there are many policies in the nodeorder plugin. I checked the e2e case in the test/e2e folder ,It looks like that e2e cases for a plugin are included

It's all right without e2e case appended, UT is enough for current codes.

By the way, could you please merge current 8 commit to 1 commit? Then you cloud add a new commit with UT, less commits would be much more clear for reviewers to read.

zhifanggao:
new trace in #2593 , close it

PR #2551, opened 2022-10-27
unified volcano-controller name

hwdef:
volcano-controller has two name now. "volcano-controller" and "volcano-controllers". Inconsistent names are not suitable for maintenance.

### DIFF
--- a/installer/helm/chart/volcano/templates/controller.yaml
+++ b/installer/helm/chart/volcano/templates/controller.yaml
@@ -2,14 +2,14 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ .Release.Name }}-controllers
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
I am a little concerned about the compatibility during upgrade

hwdef:
> I am a little concerned about the compatibility during upgrade

Yes, this could be a problem. 
When I was developing a volcano plugin for minikube, the CI of minikube failed because the names did not match. But it's not a big problem, so it doesn't matter if this PR doesn't get accepted

Thor-wl (approved):

william-wang:
/hold
The controllers contains a set of controllers like queue controller, job controller, podgroup controller. So we named it controllers instead of controller after previous discussion.

Issue #2554, opened 2022-10-28
release 1.6 memory leak


nkflash:
release: release-1.6

config like:
  volcano-scheduler.conf: |
    actions: "enqueue, reclaim, allocate, preempt, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
        enableJobStarving: false
        enableReclaimable: false
        enablePreemptable: false
      - name: conformance
    - plugins:
      - name: overcommit
      - name: drf
      - name: predicates
      - name: proportion
        enableJobEnqueued: false
      - name: nodeorder
      - name: binpack

open gctrace=1, we could see the heap size keep rise...
![image](https://user-images.githubusercontent.com/42488903/198518708-6b64a91a-cc02-44d8-884e-4ce5d9992dcf.png)

![image](https://user-images.githubusercontent.com/42488903/198518909-ba7e0234-2db4-4d5d-8920-140841b0d6cc.png)

There are no jobs in cluster(never submit any job into cluster, it just an empty cluster), scheduler keep leak memory.

commit info:
![image](https://user-images.githubusercontent.com/42488903/198519775-884e924c-fb16-4aee-bfc9-80d0f3741130.png)

go env:
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/xbmin/.cache/go-build"
GOENV="/home/xbmin/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/xbmin/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/xbmin/go"
GOPRIVATE=""
GOPROXY="https://goproxy.cn,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.18.3"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/mnt/data/xbmin/install/volcano/1.6/volcano/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
[...]

PR #2506, opened 2022-09-26
add volume binding plugin

elinx:
1. `VolumeBinding` plugin should rank before `VolumeZone` like kube-scheduler do
2. `NodeOrder` is by default enabled
3. `update` permission of pv is required for late binding of pvc

test cases include:
1. Late pvc binding using `WaitForFirstConsumer` storage class
2. node order based on the utilization, request-capacity percentage: the larger of the node capactiy and the lower utilization, the high of the score.
3. scripts used:
```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: pv-local
spec:
  capacity:
    storage: 5Gi
  volumeMode: Filesystem
  accessModes:
  - ReadWriteOnce
  persistentVolumeReclaimPolicy: Delete
  storageClassName: local-storage
  local:
    path: /data/k8s/localpv
  nodeAffinity: 
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - 192.168.2.64

---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: local-storage
provisioner: kubernetes.io/no-provisioner
volumeBindingMode: WaitForFirstConsumer

---
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: pvc-local
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
  storageClassName: local-storage

---
apiVersion: v1
kind: Pod
metadata:
  name: pv-local-pod
spec:
  imagePullSecrets:
    - name: default-secret
  schedulerName: volcano
  volumes:
  - name: example-pv-local
    persistentVolumeClaim:
      claimName: pvc-local
  containers:
  - name: example-pv-local
    image: nginx:1.23
    ports:
    - containerPort: 80
    volumeMounts:
    - mountPath: /usr/share/nginx/html
      name: example-pv-local

```

resolve #2485 

Signed-off-by: xilinxing <xilinxing@huawei.com>

PR #2555, opened 2022-10-29
Revert "merge #2506(add volume binding plugin) into scheduler/plugins/nodeorder;"

elinx:
This reverts commit 6e0beacfe390901c29f94c58560aa2052ff3724c but keeps the privilege added in case of memory leak #2554.
It looks like the `volumebinding` plugin has been breaked into several parts into the scheduler reside in scheduler cache as `volumeBinder` to do late binding of pv and pvc and statement.go to do the `preFilter` and `Filter`.

Signed-off-by: xilinxing <xilinxing@huawei.com>


william-wang (approved):
/lgtm

william-wang (approved):
/lgtm

PR #2568, opened 2022-11-09
fix extender plugin http send wrong action name

jiamin13579:
Signed-off-by: min jia <jiamin13579@126.com> 
extender plugin wrong sending http address, It may be a clerical error.


william-wang (approved):
/lgtm
/approve

Thor-wl (approved):

PR #2437, opened 2022-08-14
Add the allocation policy for deserved in scenario guarantee

liblikewhen:
Add the allocation policy for deserved in scenario guarantee

PR #2570, opened 2022-11-17
Design doc for proportion plugin split

zbbkeepgoing:
Solve the conflict between Capacity and Weight, and between Guarantee and Weight

Signed-off-by: Binbin Zou <binbin.zou@kyligence.io>

### DIFF
--- a/docs/design/capacity-plugin.md
+++ b/docs/design/capacity-plugin.md
@@ -1,0 +1,63 @@
+# Capacity Plugin
+
+## Motivation
+
+Capacity plugin is split from proportion, as a long time, capacity of queue works in proportion plugin, proportion plugin is a weight-base plugin, but capacity is not suitable with weight-base.There are contradictions between them.As the total weight increases, the capacity will slowly become invalid.
+
+### Design
+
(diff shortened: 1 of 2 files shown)


zbbkeepgoing:
#2364

william-wang:
Thanks for your contribution:) We will review it as soon as possible.

liblikewhen:
> #2364

for slove the problem, I write a new plugin in this pr: https://github.com/volcano-sh/volcano/pull/2437.  it already run in our cluster.

docs/design/capacity-plugin.md
```diff
@@ -0,0 +1,63 @@
+# Capacity Plugin
+
+## Motivation
+
+Capacity plugin is split from proportion, as a long time, capacity of queue works in proportion plugin, proportion plugin is a weight-base plugin, but capacity is not suitable with weight-base.There are contradictions between them.As the total weight increases, the capacity will slowly become invalid.
```
Thor-wl, line 5: Execuse me, may I know the reasons or scenarios why `capacity is not suitable with weight-base`? BTW, I agree with that weight is an attribute of queue. It should be watched and controlled by an independent component such as `queue controller` instead of a plugin. Users can decide to configure the plugin or not. But the queue weight should work all the time without any influence.

wangyang0616:
This one makes sense, hope it stays open.

wangyang0616:
This one makes sense, hope it stays open.

Issue #2547, opened 2022-10-24
high priority task cannot preemt low priority task


nkflash:
https://github.com/volcano-sh/volcano/issues/2232 still exist

1) create one queue, with 3 GPU card in cap.(this is only one node in cluster which has 8 cards)
2) submit lower priority job which require 2 GPU cards, wait until job running
3) submit higher priority job which require 2 GPU cards,  expect high job running and lower job should evict

result: high priority job pod always pending

config file like:
  volcano-scheduler.conf: |
    actions: "enqueue, reclaim, allocate, preempt, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
        enableJobStarving: false
      - name: conformance
    - plugins:
      - name: overcommit
      - name: drf
      - name: predicates
      - name: proportion
        enableJobEnqueued: false
      - name: nodeorder
      - name: binpack

![image](https://user-images.githubusercontent.com/42488903/197495140-7c6c761d-f144-4763-8a8b-3db510ff24bc.png)

above check incorrect, ssn.Overused only sum current allocated resources.
The logic should like: (current allocated resources + preemptor requested's resource) to check if queue overused

william-wang:
@nkflash I will work on this issue :)

jinzhejz:
hi @nkflash, which version do you use? After PR https://github.com/volcano-sh/volcano/pull/2105, pod could not be preempted by default and need to add annotation `volcano.sh/preemptable=true` to pod to enable preemption.

nkflash:
> hi @nkflash, which version do you use? After PR #2105, pod could not be preempted by default and need to add annotation `volcano.sh/preemptable=true` to pod to enable preemption.

I already set volcano.sh/preemptable=true to enable preemption

shmily0000:
@nkflash Hello friend, I also ran into this problem. Can you give an example of the vcjob you added.

nkflash:
> @nkflash Hello friend, I also ran into this problem. Can you give an example of the vcjob you added.

See below comment

shmily0000:
@nkflash OK, thank you brother, but the vcjob I tested myself still doesn't seem to work. The scheduler uses your configuration, the node is 8cpu, and the following is my configuration, which seems to be consistent with the po template you provided. I'm a bit troubled here. Can you help me to check whether the configuration is wrong
low priority job：
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: job-1
spec:
[...]

PR #2575, opened 2022-11-22
fix: high priority task cannot preemt low priority task

wangyang0616:
fix: [#2547](https://github.com/volcano-sh/volcano/issues/2547)

When the queue resource usage exceeds the upper limit, the proportion does not allow new jobs in the queue to be scheduled. If the rejected jobs are high-priority jobs, the preemption function cannot be performed.

To solve the preceding problem, modify the configuration as follows and set it to the default configuration:
- If enableJobEnqueued of proportion is set to false, jobs are allowed to enter the queue. In the allocate phase, the upper limit of resources used by each queue is controlled.

- The default preemption switch of the gang plug-in is modified. The minAvailable resource of a job can be preempted and the resource of a high-priority job is preferentially guaranteed.

Signed-off-by: wangyang <wangyang289@huawei.com>

### DIFF
--- a/installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
+++ b/installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
@@ -3,7 +3,6 @@ tiers:
 - plugins:
   - name: priority
   - name: gang
-    enablePreemptable: false
   - name: conformance
   - name: sla
 - plugins:
--- a/installer/helm/chart/volcano/config/volcano-scheduler.conf
+++ b/installer/helm/chart/volcano/config/volcano-scheduler.conf
@@ -3,7 +3,6 @@ tiers:
 - plugins:
   - name: priority
   - name: gang
-    enablePreemptable: false
   - name: conformance
 - plugins:
   - name: overcommit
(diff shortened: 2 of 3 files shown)


hwdef (commented):
I don't see what was changed

nkflash:
> fix: [#2547](https://github.com/volcano-sh/volcano/issues/2547)
> 
> When the queue resource usage exceeds the upper limit, the proportion does not allow new jobs in the queue to be scheduled. If the rejected jobs are high-priority jobs, the preemption function cannot be performed.
> 
> To solve the preceding problem, modify the configuration as follows and set it to the default configuration:
> 
> * If enableJobEnqueued of proportion is set to false, jobs are allowed to enter the queue. In the allocate phase, the upper limit of resources used by each queue is controlled.
> * The default preemption switch of the gang plug-in is modified. The minAvailable resource of a job can be preempted and the resource of a high-priority job is preferentially guaranteed.
> 
> Signed-off-by: wangyang [wangyang289@huawei.com](mailto:wangyang289@huawei.com)

I don't think this could resolve the problem.
The problem is "For current code, only low priority job occupy resources more than queue weight, then high job could preempt low job."

The reproduce conf like:
name: gang
enableJobStarving: false
enableReclaimable: false
enablePreemptable: false

name: proportion
enableJobEnqueued: false

PR #2580, opened 2022-11-28
move prefilter out of predicates to improve performance

### LINKED ISSUE #2583: scheduling performance drop with a large amount of pods existing
**What happened**:
deploying 3k pods in a cluster with 8k existing pods costs much more time than in an empty cluster, like a 40% performance drop.

same scheduling time or less than a 5% performance drop.

1. deploy 8k pods in a cluster
2. deploy another 3k pods
3. statistics the scheduling time

**Environment**:
- Volcano Version: 1.6.0
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

elinx:
- when porting kube-scheduler plugins, the current implementation put `PreFilter` together with `Filter` into `PredicateFn`, the `PredicateFn` process each pod-node pair like the following pseudo code:  
  ``` python
  for pod in pods:
    for nodes in nodes:
      predicateFn(pod, node) # prefilters & filters
  ```
- it's functioning well while `PreFilter` only needs the pod object(no node required), we don't need to put it in the innermost loop. So a promotion of `PreFilter` to the outer loop is made in this pr for performance's sake, as a plugin like `inter-pod-afftinity` is very time-consuming to prepare the pod interconnections in `PreFilter`.  After this pr:
  ``` python
  for pod in pods:
    prePredicateFn(pod) # prefilters
    for nodes in nodes:
      predicateFn(pod, node) # filters
  ```
- Implementation notes:  
	- Is it right to put `stat` in `OnSessionOpen` as previously put into the `PredicateFn`?  
	  yes, it's correct, just as the type of  `stat` suggested `CycleState` its lifecycle should be in the whole scheduling cycle, and the kube-scheduler creates it in the `ScheduleOnce` function  
	  ``` go
	  	  func (sched *Scheduler) scheduleOne(ctx context.Context) {
	  	  	// ...
	  	  	state := framework.NewCycleState()
	  	    	scheduleResult, err := 
	  	    		sched.Algorithm.Schedule(schedulingCycleCtx, prof, state, pod)
	  	      // ...
	  	  }
	  ```
	- is it thread-safe to separate write(in `PreFilter`) and read (in `Filter`)?  
	  yes, the underlining data structure is a thread-safe map for `CycleState` to access data

resolves #2583

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -198,6 +198,16 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 			klog.V(3).Infof("There are <%d> nodes for Job <%v/%v>", len(ssn.Nodes), job.Namespace, job.Name)
 
+			if err := ssn.PrePredicateFn(task); err != nil {
+				klog.V(3).Infof("PrePredicate for task %s/%s failed for: %v", task.Namespace, task.Name, err)
+				fitErrors := api.NewFitErrors()
+				for _, ni := range allNodes {
+					fitErrors.SetNodeError(ni.Name, err)
+				}
+				job.NodesFitErrors[task.UID] = fitErrors
+				break
+			}
+
 			predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, predicateFn)
 			if len(predicateNodes) == 0 {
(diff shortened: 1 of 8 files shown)


Thor-wl (approved):
Thanks for the optimization. It looks OK for me generally.

PR #2584, opened 2022-11-29
not set owner reference for pod group to improve delete performance 

jinzhejz:
Signed-off-by: Zhe Jin <jinzhe.hit@gmail.com>

Currently, pod group controller will set owner reference of pod to pod group and pod group will be deleted automatically when owner reference resource is deleted. This will cause performance downgrade when delete owner reference resource.

This pr introduce a new option `--associate-owner-reference` for volcano controller and its default value is true.
`--associate-owner-reference=true`  means the behavior is as same as before.
`--associate-owner-reference=false` means volcano controller will not set owner reference for pod group and pod group will be deleted by volcano controller after pod owner reference is deleted.

Only `ReplicaSet` is supported in this PR.


Thor-wl (commented):
@jinzhejz Can you provide more information about the performance downgrade? And at what situation will users delete the ownreference of pod?

PR #2600, opened 2022-12-12
Added helm hooks to fix `helm upgrade`

tgaddair:
Currently, running `helm upgrade --namespace volcano-system volcano .` will fail because the `volcano-admission-init` will fail, as the secret it creates already exists. As this Job only needs to be run for install, we can add a hook to disable it on subsequent upgrades. Also, we add a hook to delete the job when it completes successfully.


tgaddair:
cc @william-wang can you take a look?

william-wang:
> cc @william-wang can you take a look?

Sure, I will have a look :)

hwdef (commented):
please squash your commits and make CI happy

PR #1672, opened 2021-08-12
feat: expose detailed scheduling reason of pending tasks based on las…

eggiter:
- Currently, the reason and message in `pod.status.conditions["PodScheduled"]` and `podgroup.status.conditions["Unschedulable"]` is NOT informative, for example `x/x tasks in gang unschedulable: pod group is not ready, x Pending, x minAvailable`.
- Users want to know which one breaks the scheduling cycle and why is that.
- This commit improves this situation by introducing two extra scheduling reasons in PodScheduled PodCondition:
  + `Schedulable`: means that the scheduler can schedule the pod right now, but other pods in podgroup can not be scheduled
  + `Undetermined`: means that the scheduler dose not schedule the pod yet, for example due to unschedulable pod already occurred.
- Case:
  + 3 nodes;
  + 6 tasks in the podgroup(`minAvailable = 6`);
      + only 1 task(Task6) is unschedulable and other 5 tasks(Task1 ~ 5) can be scheduled, so the whole job is unscheduable. 
- Current information:
  +  |Tasks|Reason|Message|
     |-|-|-|
     |PodGroup |NotEnoughResources|6/6 tasks in gang unschedulable: pod group is not ready, 6 Pending, 6 minAvailable |
     |Task1 ~ 5|Unschedulable|6/6 tasks in gang unschedulable: pod group is not ready, 6 Pending, 6 minAvailable|
     |Task6|Unschedulable| all nodes are unavailable: 1 plugin InterPodAffinity predicates failed node(s) didn't match pod affinity/anti-affinity, node(s) didn't match pod affinity rules, 2 node(s) resource fit failed.|
- Improved Information
  + |Tasks|Reason|Message|
     |-|-|-|
     |PodGroup |(same)|6/6 tasks in gang unschedulable: pod group is not ready, 6 Pending, 6 minAvailable; **Pending: 1 Unschedulable, 2 Undetermined, 3 Schedulable** |
     |Task1 ~ 2|**Undetermined**|6/6 tasks in gang unschedulable: pod group is not ready, 6 Pending, 6 minAvailable; **Pending: 1 Unschedulable, 2 Undetermined, 3 Schedulable**  |
     |Task3|**Schedulable**| **Pod ns1/task-1 can possibly be assgined to node1** |
     |Task4|**Schedulable**| **Pod ns1/task-2 can possibly be assgined to node2** |
     |Task5|**Schedulable**| **Pod ns1/task-3 can possibly be assgined to node3** |
[...]

PR #2602, opened 2022-12-12
Remove `Undetermined` reason to fix cluster autoscaler compatibility

### LINKED ISSUE #1775: scheduling with update from v1.2.0 to v1.4.0
**What happened**:

volcano update from 1.2.0 to 1.4.0. With the newest version if there are not enough resources `PodGroup`s are kept in `Pending` phase and cluster autoscaler does not trigger to provision more resources. 

Did I miss smth in the latest version?

I was expecting it to work as in 1.2.0

**Environment**:
- Volcano Version: 1.4.0
- Kubernetes version (use `kubectl version`): 1.22.2
- Cloud provider or hardware configuration: GKE
- OS (e.g. from /etc/os-release): Container OS
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

tgaddair:
Fixes #1775.

#1672 introduced fine-grained Reasons to give users more context about why scheduling did not succeed. However, this breaks compatibility with cluster autoscalers like Karpenter, which rely on the reason being set to `Unschedulable` to trigger scale-up, and the more fine-grained reason `Undetermined` causes the autoscaler to skip the pod.

This PR removes this reason to fix support for cluster autoscalers.

### DIFF
--- a/docs/design/scheduling-reason.md
+++ b/docs/design/scheduling-reason.md
@@ -1,6 +1,7 @@
 # Scheduling Reason
 
 [@eggiter](https://github.com/eggiter); Aug 13, 2021
+[@tgaddair](https://github.com/tgaddair); Dec 12, 2022
 
 ## Table of Contents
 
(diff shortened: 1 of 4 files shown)


william-wang:
@tgaddair The test file job_info_test.go is also needed to update to fix the CI error :)

tgaddair:
Thanks for taking a look @william-wang! Tests are updated and everything is passing.

william-wang:
> 

That's great :)

william-wang (approved):
/lgtm

PR #2604, opened 2022-12-14
[Cherry-pick] Add allocated field in queue status to indicate used resources in queue. #2592

jiangkaihua:


Thor-wl (approved):
That makes sense a lot for administors. Thanks!

william-wang (approved):
/lgtm

PR #2606, opened 2022-12-19
perf enhancement: imporve podgroup creation speed

elinx:


hzxuzhonghu (approved):

pkg/scheduler/framework/session.go
```diff
@@ -63,6 +64,9 @@ type Session struct {
 	Queues         map[api.QueueID]*api.QueueInfo
 	NamespaceInfo  map[api.NamespaceName]*api.NamespaceInfo
 
+	NodeMap   map[string]*k8sframework.NodeInfo
```
william-wang, line 67: There is a big redundent between Nodes and NodeMap and a little bit confusing for developers.

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -116,7 +118,9 @@ func (pg *pgcontroller) Run(stopCh <-chan struct{}) {
 		}
 	}
 
-	go wait.Until(pg.worker, 0, stopCh)
+	for i := 0; i < int(pg.workers); i++ {
```
william-wang, line 121: It's nice to have this enhancement.

pkg/scheduler/framework/session.go
elinx: Yes, they work almost like the same, the only difference is the later one is used when calling k8s apis, I just put it here for performance sake. I add a little comments above, hopefully it'll be more clear.

william-wang (approved):
/lgtm

Issue #2612, opened 2022-12-22
In the same queue, a low-priority job preempts resources of a high-priority job


wangyang0616:
**What happened**:
Create two jobs in the same queue. Set the priority of job-a to 100 and that of job-b to 10.

Create job-a, set the number of copies to 12, and set minAvailable to 2 to use up idle CPUs in the cluster. After all pods of job-a are running, create job-b with replicas set to 6 and minAvailable set to 2. Job-b preempts job-a resources.

**What you expected to happen**:
The pod of job-b is in the pending state, and the pod of job-a is not preempted.

**How to reproduce it (as minimally and precisely as possible)**:
- job-a
```
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: share-1-priority
value: 100
globalDefault: false
description: "This priority class should be used for volcano job only."
---
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: job-a
spec:
  priorityClassName: share-1-priority
  schedulerName: volcano
  minAvailable: 2
  tasks:
    - replicas: 12
      name: "test"
      template:
        spec:
          containers:
            - image: alpine
              command: ["/bin/sh", "-c", "sleep 1000"]
              imagePullPolicy: IfNotPresent
              name: running
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure
```
- job-b
```
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: share-2-priority
value: 10
globalDefault: false
description: "This priority class should be used for volcano job only."
---
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: job-b
spec:
  priorityClassName: share-2-priority
  schedulerName: volcano
  minAvailable: 2
  tasks:
    - replicas: 6
      name: "test"
      template:
        spec:
          containers:
            - image: alpine
              command: ["/bin/sh", "-c", "sleep 1000"]
              imagePullPolicy: IfNotPresent
              name: running
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure
```

**Anything else we need to know?**:
```
apiVersion: v1
data:
  volcano-scheduler-ci.conf: |
    actions: "enqueue, allocate, backfill, reclaim, preempt"
    tiers:
    - plugins:
      - name: priority
      - name: gang
      - name: conformance
      - name: sla
    - plugins:
      - name: overcommit
      - name: drf
[...]

PR #2613, opened 2022-12-22
By default, the preemption function of gang and drf is turned off

wangyang0616:
Signed-off-by: wangyang <wangyang8126@gmail.com>

fix: [#2612](https://github.com/volcano-sh/volcano/issues/2612)

priority is a common priority preemption plug-in. By default, job preemption is performed based on the priority. If gang and drf preemption is required, you can manually enable it.


william-wang (approved):
/lgtm

Issue #2615, opened 2022-12-22
JobInfo.GetMinResources cannot return extended resources


stingshen:
**What happened**:

JobInfo.GetMinResources cannot return extended resources.

**What you expected to happen**:

JobInfo.GetMinResources can return extended resources such as nvidia.com/gpu

**How to reproduce it (as minimally and precisely as possible)**:

1. Enable proportion plugin and start volcano scheduler.
2. Create a vcjob with extended resources, such as nvidia.com/gpu.
3. JobInfo.GetMinResources can't return nvidia.com/gpu.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`): 1.25.0
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release): Ubuntu 22.04
- Kernel (e.g. `uname -a`): 5.15.0-56-generic
- Install tools:
- Others:

stingshen:
> Some do not understand the place want to ask, this to the user performance on the main impact is what aspect.
> 
> During task scheduling, the portion cannot perform enqueue interception on jobs whose usage of extended resources exceeds the upper limit. Is it correct?

Yes. Proportion plugin attr.inqueue doesn't count ERs.

stingshen:
> Is it the same problem as [#2564](https://github.com/volcano-sh/volcano/issues/2564)?

The problem is related but not the same. The problem described in issue is about the return value of `JobInfo.GetMinResources()`, and the problem described in #2564 is about quota calculation.

hwdef:
still need

stingshen:
fixed

PR #2616, opened 2022-12-22
Return ER resource in JobInfo.GetMinResources()

stingshen:
Signed-off-by: stingshen <stingshen@126.com>

issue: #2615 

Currently, JobInfo.GetMinResources() only returns natively supported resources. It should also return extended resources.

The root cause is [podComputeUsageHelper](https://github.com/kubernetes/kubernetes/blob/master/pkg/quota/v1/evaluator/core/pods.go#L286) only sets requests.resourceName, not resourceName for ERs.

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -481,7 +481,17 @@ func (ji *JobInfo) GetMinResources() *Resource {
 		return EmptyResource()
 	}
 
-	return NewResource(*ji.PodGroup.Spec.MinResources)
+	resources := v1.ResourceList{}
+
+	for k, v := range *ji.PodGroup.Spec.MinResources {
+		if !strings.HasPrefix(string(k), v1.DefaultResourceRequestsPrefix) {
+			continue
+		}
+		k = v1.ResourceName(strings.TrimPrefix(string(k), v1.DefaultResourceRequestsPrefix))
+		resources[k] = v
+	}
+
+	return NewResource(resources)
 }
 
 func (ji *JobInfo) GetElasticResources() *Resource {
--- a/pkg/scheduler/api/job_info_test.go
+++ b/pkg/scheduler/api/job_info_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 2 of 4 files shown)


stingshen:
> Please add you test result, thanks @stingshen

Sorry for the delay.. What test result do you mean?

wangyang0616:
> > Please add you test result, thanks @stingshen
> 
> Sorry for the delay.. What test result do you mean?

I mean, can the expected effect be achieved through this modification.

wangyang0616:
JobInfo.GetMinResources() currently calculates extended resources, as follows:
https://github.com/volcano-sh/volcano/blob/dfa0c9398147121acbdd606f6732bd4c6fe94bcd/pkg/scheduler/api/resource_info.go#L85-L88

Is it a duplicate to add again?

stingshen:
> `JobInfo.GetMinResources()` currently calculates extended resources, as follows:
> 
> https://github.com/volcano-sh/volcano/blob/dfa0c9398147121acbdd606f6732bd4c6fe94bcd/pkg/scheduler/api/resource_info.go#L85-L88
> 
> Is it a duplicate to add again?

No, it is not a duplicate. Actually, it removes `requests.` prefix from resource name, so `IsScalarResourceName()` returns true for ER resources. `IsScalarResourceName()` returns false for resource name that has prefix of `requests.`.

stingshen:
> > > Please add you test result, thanks @stingshen
> > 
> > 
> > Sorry for the delay.. What test result do you mean?
> 
> I mean, can the expected effect be achieved through this modification.

Yes. the expected effect can be achived through this modification. Without this modification, JobInfo.GetMinResources() never returns external resources.

wangyang0616:
> No, it is not a duplicate. Actually, it removes `requests.` prefix from resource name, so `IsScalarResourceName()` returns true for ER resources. `IsScalarResourceName()` returns false for resource name that has prefix of `requests.`.

I may not understand this processing method, please help other contributors to review

Is it convenient to take the time to review it? @william-wang @hwdef

hwdef:
I'm not familiar with this part either, but I'll try to check it.

hwdef (commented):
Generally speaking, this PR is OK, but I agree with @wangyang0616 , you should add some unit tests or e2e tests for this modification

stingshen:
> Generally speaking, this PR is OK, but I agree with @wangyang0616 , you should add some unit tests or e2e tests for this modification

the unit test is added.

hwdef:
please fix the DCO and squash your commits.

stingshen:
> please fix the DCO and squash your commits.

done

wangyang0616:
cc @william-wang 
Please help to review it, and if possible, help to trigger CICD manually.

pkg/scheduler/api/job_info.go
```diff
@@ -481,7 +481,17 @@ func (ji *JobInfo) GetMinResources() *Resource {
 		return EmptyResource()
 	}
 
-	return NewResource(*ji.PodGroup.Spec.MinResources)
+	resources := v1.ResourceList{}
+
+	for k, v := range *ji.PodGroup.Spec.MinResources {
```
lowang-bh, line 486: How about to prevent it from the source where min is calculated?
code is located at
https://github.com/volcano-sh/volcano/blob/015f86b9879d9204ab1607bd6e0f07189c7ecb5e/pkg/controllers/util/util.go#L12

PR #2620, opened 2022-12-24
add some more metrics

elinx:
This relate to #2606 I found the metrics in this pr helps when I instrument volcano scheduler to find the performance bottleneck.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -19,12 +19,13 @@ package cache
 import (
 	"context"
 	"fmt"
-	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
+
+	"k8s.io/kubernetes/pkg/scheduler/framework"
 
 	"github.com/prometheus/client_golang/api"
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
@@ -770,6 +771,7 @@ func (sc *SchedulerCache) Bind(tasks []*schedulingapi.TaskInfo) error {
 		errTasks, err := sc.Binder.Bind(sc.kubeClient, taskArray)
 		if err == nil {
 			klog.V(3).Infof("bind ok, latency %v", time.Since(tmp))
+			metrics.UpdateTaskBindDuration(metrics.Duration(tmp))
 			for _, task := range tasks {
 				sc.Recorder.Eventf(task.Pod, v1.EventTypeNormal, "Scheduled", "Successfully assigned %v/%v to %v",
 					task.Namespace, task.Name, task.NodeName)
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -28,6 +28,7 @@ import (
 
 // OpenSession start the session
 func OpenSession(cache cache.Cache, tiers []conf.Tier, configurations []conf.Configuration) *Session {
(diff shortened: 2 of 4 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -770,6 +771,7 @@ func (sc *SchedulerCache) Bind(tasks []*schedulingapi.TaskInfo) error {
 		errTasks, err := sc.Binder.Bind(sc.kubeClient, taskArray)
 		if err == nil {
 			klog.V(3).Infof("bind ok, latency %v", time.Since(tmp))
+			metrics.UpdateTaskBindDuration(metrics.Duration(tmp))
```
Thor-wl, line 774: How about bind failure? Besides, may I know why add this metric? Too many metrics may delay the main scheduling process.

pkg/scheduler/framework/framework.go
```diff
@@ -45,16 +46,19 @@ func OpenSession(cache cache.Cache, tiers []conf.Tier, configurations []conf.Con
 			}
 		}
 	}
+	metrics.UpdatePluginDuration("total", metrics.OnSessionOpen, metrics.Duration(sessionStartTs))
```
Thor-wl, line 49: Can you give some interpretations about these metrics? Personally speaking, I don't think we should add a metric for process of `openSession`.

pkg/scheduler/cache/cache.go
elinx: When perfromance issuse is reported, the initial thinking to me always would be 'how is the network conditions? is it caused by the laggy network?', but the current implementation only has logs printed, we need to filter-process-analysis the data which is what prometheus prepared for us, right? So with this metric added we could easily figure out the p99/p95 of the binding latency, and easily exlclude/concentrate the network cause from real performance issue.

elinx: As the above comment we mainly care the p99/p95 latency to clearify the performance issue, failue is not very import to this analysis I thought.

pkg/scheduler/framework/framework.go
elinx: The main goal is to make it easier to analyze key performance issues. For example, if there are 10 plugins and we have metrics of open and close for each of them, then there are 20 curves. When we want to know if open/close is the performance bottleneck, we want to see the overall latency of open and close first instead of each plugin; furthermore, if we have determined that open/close is the performance bottleneck, we will look at the performance of each plugin, at which time we will see 10/20 curves. If there is a curve of total usage at this time, it will be easier to visually perceive the performance of each plugin.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -641,10 +644,12 @@ func (ssn *Session) PredicateFn(task *api.TaskInfo, node *api.NodeInfo) error {
 			if !found {
 				continue
 			}
+			startTime := time.Now()
 			err := pfn(task, node)
 			if err != nil {
 				return err
 			}
+			metrics.UpdatePluginDuration(plugin.Name, metrics.SessionRun+":Predicate", metrics.Duration(startTime))
```
lowang-bh, line 652: use constants may be more efficient, may be.

william-wang (approved):
/approve

jiangkaihua (commented):
LGTM, please reslove the confilicts. @elinx

hwdef:
reopen

lowang-bh:
/hold

plz use a constant string to replace the string join operation.

PR #2631, opened 2022-12-29
remove gox

hwdef:

Removing the gox tool has the following benefits:
* Simplify project construction complexity
* Reduce the packages that the project depends on
* Compilation is smoother in China's network environment, because it is likely to fail when installing gox

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,29 +32,31 @@ endif
 # Get OS architecture
 OSARCH=$(shell uname -m)
 ifeq ($(OSARCH),x86_64)
-REL_OSARCH=linux/amd64
+GOARCH?=amd64
 else ifeq ($(OSARCH),x64)
-REL_OSARCH=linux/amd64
+GOARCH?=amd64
 else ifeq ($(OSARCH),aarch64)
-REL_OSARCH=linux/arm64
+GOARCH?=arm64
 else ifeq ($(OSARCH),aarch64_be)
-REL_OSARCH=linux/arm64
+GOARCH?=arm64
 else ifeq ($(OSARCH),armv8b)
-REL_OSARCH=linux/arm64
+GOARCH?=arm64
 else ifeq ($(OSARCH),armv8l)
-REL_OSARCH=linux/arm64
+GOARCH?=arm64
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -67,41 +68,26 @@ init:
 	mkdir -p ${RELEASE_DIR}
 
 vc-scheduler: init
-	GO111MODULE=off go get github.com/mitchellh/gox
 	if [ ${SUPPORT_PLUGINS} = "yes" ];then\
-		CC=${CC} CGO_ENABLED=1 $(GOBIN)/gox -osarch=${REL_OSARCH} -ldflags ${LD_FLAGS} -output ${BIN_DIR}/vc-scheduler ./cmd/scheduler;\
+		CC=${CC} CGO_ENABLED=1 GOOS=linux go build -ldflags ${LD_FLAGS} -o ${BIN_DIR}/vc-scheduler ./cmd/scheduler;\
```
hzxuzhonghu, line 72: should not hardcoding `linux`

hwdef: Thanks for your suggestion.I fixed it.

Makefile
```diff
@@ -32,29 +32,31 @@ endif
...
 else
-REL_OSARCH=linux/$(OSARCH)
+GOARCH?=$(OSARCH)
 endif
 
 # Run `make images DOCKER_PLATFORMS="linux/amd64,linux/arm64" BUILDX_OUTPUT_TYPE=registry IMAGE_PREFIX=[yourregistry]` to push multi-platform
-DOCKER_PLATFORMS ?= "${REL_OSARCH}"
+DOCKER_PLATFORMS ?= "linux/${GOARCH}"
+
+GOOS ?= linux
```
hzxuzhonghu, line 59: nit: it seems we donot use this at all, This env is set in go env

hzxuzhonghu (approved):

Makefile
hwdef: Because we only compile the linux version at present, if the default value is not set to linux, then when compiling on macos, the binary in macos format will be compiled

Issue #2629, opened 2022-12-29
For services other than job, such as deployment and pod, high priority cannot preempt low priority resources


wangyang0616:
**What happened**:
Deployment tasks are deployed through the Volcano scheduler, and high-priority jobs cannot preempt resources from low-priority jobs.

Three priority classes are set in the cluster, namely high, mid, and low. 
```
[root@ecs-4b42-0002 volcano]# kubectl get priorityclass
NAME                      VALUE        GLOBAL-DEFAULT   AGE
high-priority             100          false            102m
low-priority              10           false            102m
mid-priority              50           false            102m
```

First, a deployment `deploy-low` with a priority class of low is created, and the replicas  is 4, which occupies the idle resources of the cluster.
Then create a deployment `deploy-high` with a priorityClass of high and a replica number of 4.

- The result of kube-scheduler is that `deploy-high` preempts the resources of `deploy-low`, and all 4 copies of `deploy-high` are running
  ```
  NAME                          READY   STATUS    RESTARTS   AGE
  deploy-high-d66c9db75-6sjnt   1/1     Running   0          51s
  deploy-high-d66c9db75-92bkp   1/1     Running   0          51s
  deploy-high-d66c9db75-nnvhq   1/1     Running   0          51s
  deploy-high-d66c9db75-pjwnz   1/1     Running   0          51s
  deploy-low-b46c7466c-knz7j    0/1     Pending   0          51s
  deploy-low-b46c7466c-lb6n6    0/1     Pending   0          51s
  deploy-low-b46c7466c-vvz9m    0/1     Pending   0          51s
  deploy-low-b46c7466c-z4kkf    0/1     Pending   0          51s
  ```

- The results of the volcano scheduler and the kube-scheduler are inconsistent, `deploy-high` triggers the preemption action, and it is always pending
  ```
  NAME                           READY   STATUS    RESTARTS   AGE
  deploy-high-587b867944-54dtc   0/1     Pending   0          2m35s
  deploy-high-587b867944-5gs7s   0/1     Pending   0          2m35s
  deploy-high-587b867944-n6m8l   0/1     Pending   0          2m35s
  deploy-high-587b867944-xfc5z   0/1     Pending   0          2m35s
  deploy-low-67b97db94b-5g56g    1/1     Running   0          3m10s
  deploy-low-67b97db94b-6lztd    1/1     Running   0          3m10s
  deploy-low-67b97db94b-ggjrw    1/1     Running   0          3m10s
[...]

PR #2634, opened 2022-12-30
The preemption logic is only controlled by jobMinavailable in the gang plugin

wangyang0616:
Signed-off-by: wangyang <wangyang289@huawei.com>

fix：[#2629](https://github.com/volcano-sh/volcano/issues/2629)

In the preemption scenario, the gang plugin does not pay attention to the taskMinAvailble configuration for the time being, but only pays attention to the jobMinAvailble


william-wang (approved):
/lgtm

PR #2638, opened 2023-01-03
🦄 refactor: simplify code by define DimensionDefaultValue as float64

kingeasternsun:
Signed-off-by: kingeasternsun <kingeasternsun@gmail.com>
By define DimensionDefaultValue as float64, we can simplify the function of Diff and setDefaultValue


hwdef (commented):
please squash your commits.

Issue #2492, opened 2022-09-08
Allow customisation of SSH port


mkjpryor:
#### What would you like to be added:

Allow the customisation of the SSH port in the SSH config file generated by the `ssh` plugin.

#### Why is this needed:

For maximum network performance, we sometimes use host networking for jobs. When using host networking, port 22 is usually taken and so needs to be set to a non-standard port. It would be nice if this could be set using an option to the `ssh` plugin.

aarushisoni:
I want to contribute to this issue . Please guide me through this process.

mkjpryor:
Still required please! I'm currently using a hack to work around.

hwdef:
still need

srikanth-iyengar:
Hello @hwdef , could you please confirm whether this issue has been resolved or if I can begin working on it? 
Thank you.

hwdef:
I created two pr #2647 #2733. They're waiting to be reviewed.

srikanth-iyengar:
Ohk Thanks for the info

gongxianjin:
https://github.com/volcano-sh/volcano/pull/4402, I am a learner, did this pr can slove this issue?

PR #2647, opened 2023-01-11
Allow user to customize ssh port

hwdef:
fix: #2492

4 steps:
1. modify `~/.ssh/config` file
2. open port for container
3. add label for pod
4. add `podAntiAffinity` for pods that use `hostNetwork`

### DIFF
--- a/pkg/controllers/job/plugins/ssh/ssh.go
+++ b/pkg/controllers/job/plugins/ssh/ssh.go
@@ -23,9 +23,11 @@ import (
 	"encoding/pem"
 	"flag"
 	"fmt"
+	"strconv"
 
 	"golang.org/x/crypto/ssh"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/controllers/job/plugins/ssh/ssh.go
```diff
@@ -48,6 +49,9 @@ type sshPlugin struct {
 
 	// public key string
 	sshPublicKey string
+
+	// ssh port
+	port int
```
hzxuzhonghu, line 54: I donot think this works for the case required. This a global option, so if a user is using hostnetwork, any fixed port can conflict with another pod within same node

hwdef: Do you have any suggestions? 
I think they have planned the ports of the pods when using the hostNetwork.

hzxuzhonghu: This is a little more complex than just set a ssh port. We need an API to set host network. 

With both port and hostnetwork set, then we still need to control pods belong to the same Job should be scheduled to different nodes(because IIUC the ssh port is per Job config).

hwdef: I think using anti-affinity can achieve this requirement, if pod uses hostNetwork, we add the same label to the pod, and then configure anti-affinity for them.

hwdef:
finally, the CI happy...

@hzxuzhonghu @Thor-wl @william-wang @wangyang0616 Please review this again.

wangyang0616:
I think it is necessary to keep.

hwdef:
still need

hwdef:
still need

PR #2648, opened 2023-01-11
fix local-up-volcano.sh

hwdef:

The image prefix does not match the one in the makefile, resulting in failure to import the image


jiangkaihua:
/lgtm

`${IMAGE_PREFIX}` was used like:
https://github.com/volcano-sh/volcano/blob/c7d87c77eb5fe47b754887acd1025d39b65aa26e/hack/lib/install.sh#L29-L31
So it should only contain the registry name. Thank you for fixing!

Thor-wl (approved):

Issue #2657, opened 2023-01-28
The completed status of the podgroup will be updated to the inqueue status


wangyang0616:
**What happened**:
If the enqueue action is not configured, after the podgroup status is updated to Completed, it will be updated to the inqueue status soon. 

**What you expected to happen**:
After the podgroup is updated to the Completed state, Volcano will not schedule the job unless the job status is updated externally

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
Code logic see：
https://github.com/volcano-sh/volcano/blob/e145aff85f7f479d1fde178a2053d50f762c878b/pkg/scheduler/actions/allocate/allocate.go#L64-L66

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

wangyang0616:
Issue used by PodGroup `Completed` [#2642](https://github.com/volcano-sh/volcano/issues/2642)

PR #2658, opened 2023-01-28
The allocate phase only updates the podgroup in the pending state to the inqueue state

wangyang0616:
Signed-off-by: wangyang <wangyang8126@gmail.com>

fix: [#2657](https://github.com/volcano-sh/volcano/issues/2657)

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -61,14 +61,16 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 	for _, job := range ssn.Jobs {
 		// If not config enqueue action, change Pending pg into Inqueue statue to avoid blocking job scheduling.
-		if !conf.EnabledActionMap["enqueue"] && job.IsPending() {
-			job.PodGroup.Status.Phase = scheduling.PodGroupInqueue
-		} else {
+		if conf.EnabledActionMap["enqueue"] {
 			if job.IsPending() {
 				klog.V(4).Infof("Job <%s/%s> Queue <%s> skip allocate, reason: job status is pending.",
 					job.Namespace, job.Name, job.Queue)
 				continue
 			}
+		} else if job.IsPending() {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -61,7 +61,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 	for _, job := range ssn.Jobs {
 		// If not config enqueue action, change Pending pg into Inqueue statue to avoid blocking job scheduling.
-		if !conf.EnabledActionMap["enqueue"] {
+		if !conf.EnabledActionMap["enqueue"] && job.IsPending() {
```
Thor-wl, line 64: It's LGTM generally. Some kindly suggestions are as follows:
1. If this patch is in use, suggest to refactor the `else` section like this: `else if job.IsPending()`, which may be more readable and clean.
2.  IMO, a more reasonable way is to filter podgroups which are not at `completed` status at the stage of snapshot. If that, only podgroups, which are to be allocated resources, will enter the session. Besides, it will improve the performance.

wangyang0616: > 

Thanks for your suggestion, the first question has been optimized.

There are still some doubts about the second question. The function implemented here is to prevent the job in the pending state from blocking when the enqueue action is not configured, so the processing logic for updating the pending job to the inqueue state is added. 

However, before the update, the status of the job was not judged, but all status jobs were updated to inqueue (including `running`, `unknown`, and `completed`). This kind of processing did not meet the original intention, and also caused the problem recorded in the issue. 

I understand that by correcting this, the problem in the issue should be solved first. For the filtering of the job in the completed state, can it not be processed for the time being?

Thor-wl: I got you. Before reviewing this PR, I've looked through the related issue. If available, can you help for the filter in the later PR?

Thor-wl: I got you. Before reviewing this PR, I've looked through the related issue. If available, can you help for the filter in the later PR?

PR #2662, opened 2023-01-31
Remove duplicate talk

gengwg:
The last line is the same as first line.
image


Thor-wl (approved):

wangyang0616:
cc @william-wang @jiangkaihua

hwdef:
The reason why this pr is not merged is that the DCO check did not pass

image

please fix this

PR #2665, opened 2023-02-01
[WIP] Fix spark podgroup block

wangyang0616:
### Modify content
When scheduling the pending podgroup into the queue, first determine whether the podgroup has ownerreference information. If not, the podgroup will not be scheduled in this cycle until the ownerreference information is updated, and then the scheduling will be performed.

### Why submit this PR
Run the task through spark submit. If spark submit is abnormally killed when the podgroup is successfully created but the corresponding pod is not created or the ownerreference information in the podgroup cannot be updated, it will cause podgroup resource leaks. If the resource is allocated, it will cause the resource to be locked.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -67,10 +67,18 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 					job.Namespace, job.Name, job.Queue)
 				continue
 			}
-		} else if job.IsPending() {
-			klog.V(4).Infof("Job <%s/%s> Queue <%s> status update from pending to inqueue, reason: no enqueue action is configured.",
-				job.Namespace, job.Name, job.Queue)
-			job.PodGroup.Status.Phase = scheduling.PodGroupInqueue
+		} else {
+			if job.IsDeferredScheduling() {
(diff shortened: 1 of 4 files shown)


wangyang0616:
There are some problems in the current modification plan, and more solutions need to be considered.

For example, the following scenarios will not work:
The user manually creates and independently manages the life cycle of the podgroup, the ownerreference information can not be set, at this time the resources in the podgroup will not be normally scheduled.

/hold

Thor-wl:
> There are some problems in the current modification plan, and more solutions need to be considered.
> 
> For example, the following scenarios will not work: The user manually creates and independently manages the life cycle of the podgroup, the ownerreference information can not be set, at this time the resources in the podgroup will not be normally scheduled.
> 
> /hold

@wangyang0616 Hey, if this PR is on the process, you can add `[WIP]` flag at the head of title. The bot will recognize that and not merge the PR.

wangyang0616:
> @wangyang0616 Hey, if this PR is on the process, you can add `[WIP]` flag at the head of title. The bot will recognize that and not merge the PR.

Thanks, it has been updated

wangyang0616:
problem exists, need to keep

hwdef:
Is this still work in progress?

PR #2667, opened 2023-02-07
Add podGroup completed phase

waiterQ:
### Modification Motivation
Volcano can schedule normal pods, but it shows up podGroup inqueue problem at some conditions. scheduling k8s-job, 
 podGroup is still at running phase when pod went to completed phase; changing deployment's requeuest will cause old podGroup in phase Inqueue. 
So I pick and fix #2589 (Feature/add replicaset gc pg), and add phase Completed for normal pod podGroup to enhances podGroup.

#### Test Result
when deployment roll-updating, there're two podgroups exists.
image

when deployment roll-update is over, one podgroup left.
image

when k8s job is completed
image
podgroup is completed
image

ci e2e result
image


pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -61,6 +63,26 @@ func (pg *pgcontroller) addPod(obj interface{}) {
 	pg.queue.Add(req)
 }
 
+func (pg *pgcontroller) addReplicaSet(obj interface{}) {
```
jiangkaihua, line 66: @waiterQ As far as I know, when replicaset was created, it would always be `0` replica. And after creating, the replicaset would scale up to defined replica numbers. So why deleting podgroup on both `addReplicaSet` and `updateReplicaSet`, but not only `updateReplicaSet`?

waiterQ: yes, you're right. In normal process with one version, volcano just need `updateReplicaSet`, and if consider the situation upgrade from a version to another, there isn't `addReplicaSet` help to cleanup stock podgroups in cluster. `addReplicaSet` is work with already-exist podgroups, `addReplicaSet` work with upcoming podgroups.

william-wang:
Please add the test results on the PR, thanks.

waiterQ:
> Please add the test results on the PR, thanks.

ok, done.

william-wang (approved):
/lgtm

pkg/scheduler/framework/session.go
```diff
@@ -272,6 +272,10 @@ func jobStatus(ssn *Session, jobInfo *api.JobInfo) scheduling.PodGroupStatus {
 		// If there're enough allocated resource, it's running
 		if int32(allocated) >= jobInfo.PodGroup.Spec.MinMember {
 			status.Phase = scheduling.PodGroupRunning
+			// If all allocated tasks is succeeded, it's completed
+			if len(jobInfo.TaskStatusIndex[api.Succeeded]) == allocated {
```
zhoushuke, line 276: for batchv1 native job, if using .spec.completions and .spec.parallelism in job, for case, successed 10, in the same time, the queue is full, some pod will pending,  len(jobInfo.TaskStatusIndex[api.Succeeded]) == allocated will be true,  job not finished but pg status is completed, is it happend?

PR #2678, opened 2023-02-11
Extract MetricsClient and NodeMetrics to support other metrics platform

Shoothzj:
see #2679 for more details

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -24,10 +24,7 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/prometheus/client_golang/api"
-	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
-	pmodel "github.com/prometheus/common/model"
+	"volcano.sh/volcano/pkg/scheduler/metrics/source"
 
 	v1 "k8s.io/api/core/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
@@ -70,10 +67,6 @@ import (
 )
 
 const (
-	// record name of cpu average usage defined in prometheus rules
-	cpuUsageAvg = "cpu_usage_avg"
-	// record name of mem average usage defined in prometheus rules
-	memUsageAvg = "mem_usage_avg"
 	// default interval for sync data from metrics server, the value is 5s
 	defaultMetricsInternal = 5000000000
 )
@@ -1266,19 +1259,11 @@ func (sc *SchedulerCache) SetMetricsConf(conf map[string]string) {
 }
 
 func (sc *SchedulerCache) GetMetricsData() {
-	address := sc.metricsConf["address"]
-	if len(address) == 0 {
-		return
-	}
-	klog.V(4).Infof("Get metrics from Prometheus: %s", address)
-	client, err := api.NewClient(api.Config{
-		Address: address,
-	})
+	client, err := source.NewMetricsClient(sc.metricsConf)
 	if err != nil {
 		klog.Errorf("Error creating client: %v\n", err)
 		return
 	}
-	v1api := prometheusv1.NewAPI(client)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
 	defer cancel()
 	nodeUsageMap := make(map[string]*schedulingapi.NodeUsage)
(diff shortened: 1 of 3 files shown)


wangyang0616:
Thank you for your contribution. 

It will be more beneficial for everyone to review if you can add a PR description, such as the background, reason and goal of this PR, etc. 

If there is a corresponding issue, you can also directly associate it.

wangyang0616:
When the architecture is updated, it is best to update the design documents synchronously [usage-based-scheduling.md](https://github.com/volcano-sh/volcano/blob/master/docs/design/usage-based-scheduling.md)

Shoothzj:
@wangyang0616 I have already created a ticket (#2679) to track this enhancement. I suggest that we don't update the document in this pull request. Instead, I prefer to update the document when step 2 is implemented, which involves the introduction of the "metrics.type" variable.
WDYT?  thanks for advance.

wangyang0616:
> @wangyang0616 I have already created a ticket (#2679) to track this enhancement. I suggest that we don't update the document in this pull request. Instead, I prefer to update the document when step 2 is implemented, which involves the introduction of the "metrics.type" variable. WDYT? thanks for advance.

This is very good. By the way, is it better to add a folder under the cache directory to store the newly added metrics files, such as `refermetrics`, the metrics obtained from other platforms can also be placed in this folder uniformly.

Shoothzj:
@wangyang0616 Hi, if no other concern, can we merge this?

wangyang0616:
It's ok in my opinion. 
cc @william-wang @Thor-wl

Thor-wl:
> It's ok in my opinion.
> cc @william-wang @Thor-wl

I'm sorry for the late reply. Since I'm a little bit busy recently, will reivew recent PRs and issues together sometime this weekend.

william-wang:
@Shoothzj Thanks for your contribution. I will take a look as soon as possible:)

pkg/scheduler/cache/refermetrics/metrics_client.go
```diff
@@ -0,0 +1,34 @@
...
+type MetricsClient interface {
+	NodeMetricsAvg(ctx context.Context, nodeName string, period string) (*NodeMetrics, error)
+}
+
+func NewMetricsClient(metricsConf map[string]string) (MetricsClient, error) {
+	address := metricsConf["address"]
+	if len(address) == 0 {
+		return nil, errors.New("metrics address is empty")
+	}
+	return &PrometheusMetricsClient{address: address}, nil
```
william-wang, line 33: How to support other metrics system like ELK, is there a whole picture design?

william-wang: why return a `PrometheusMetricsClient` in a generic `NewMetricsClient` func. It's better use a design pattern to support multiple client.

pkg/scheduler/cache/refermetrics/module.go
```diff
@@ -0,0 +1,22 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package refermetrics
+
+type NodeMetrics struct {
```
william-wang, line 19: Maybe we do not need to store `NodeMetrics` in a seperate file right now.

pkg/scheduler/cache/refermetrics/metrics_client.go
```diff
@@ -0,0 +1,34 @@
+/*
```
william-wang, line 1: Suggest to use a formal directory name instead of `refermetrics`.

pkg/scheduler/cache/refermetrics/module.go
Shoothzj: @william-wang solved

pkg/scheduler/cache/refermetrics/metrics_client.go
Shoothzj: @william-wang Solved. I have moved it to `metrics/source`

pkg/scheduler/cache/refermetrics/metrics_client.go
Shoothzj: @william-wang will add other type of `NewMetricsClient` in next PR.

william-wang (approved):
/lgtm

PR #2682, opened 2023-02-14
upgrade klog package version to latest

waiterQ:
upgrade klog version, unified klog package version to v2.80.1


william-wang (approved):
/lgtm

PR #2691, opened 2023-02-22
Optimize e2e runtime: reduce pytorch-plugin image download time

wangyang0616:
It takes too long for e2e to run the pytorch plugin use case, basically between 300s and 600s, as follows:
  ```
  • [SLOW TEST] [458.017 seconds]
  Pytorch Plugin E2E Test will run and complete finally
  /home/runner/work/volcano/volcano/test/e2e/jobseq/pytorch_plugin.go:12
  ```

- The image `pytorch-mnist:v1beta1-45c5727` is changed to `pytorch-mnist:v1beta1-9ee8fda`, and the image size is reduced from 1.2 GB to 600 MB.
- Unnecessary image download operations on the host node are reduced. Images can be directly loaded on the kind virtual node.
- The running time of e2e test cases is reduced from 38 minutes to about 30 minutes.

PR #2693, opened 2023-02-22
Change e2e sequence tests timeout to 1 hour

Shoothzj:
Since e2e sequence tests timeout frequently. We can consider change e2e sequence tests timeout to 1 hour


wangyang0616:
It is recommended not to expand the duration of the e2e operation limit. Currently, the problem of e2e operation timeout is being fixed by PR, see[#2691](https://github.com/volcano-sh/volcano/pull/2691)

PR #2705, opened 2023-02-24
Decouple the scanning cycle from the processing cycle

Shoothzj:
Decoupling the scanning cycle from the processing cycle can provide several benefits, including:

- Increased efficiency: When the scanning and processing cycles are decoupled, the processing cycle can be performed on the results of the previous scanning cycle. This reduces idle time and allows the system to operate more efficiently.
- Improved scalability: Decoupling the scanning and processing cycles can also improve system scalability. By allowing the processing cycle to operate independently of the scanning cycle, the system can handle larger volumes of data without slowing down.
- Flexibility: Decoupling the scanning and processing cycles provides greater flexibility in terms of how the system is designed and configured. For example, it may be possible to adjust the scanning cycle frequency or processing cycle duration to optimize performance for different types of data or workloads.
- Reduced resource consumption: By decoupling the scanning and processing cycles, it may be possible to reduce resource consumption (e.g. CPU, memory, power) by optimizing each cycle independently. This can help reduce operational costs and improve the overall sustainability of the system.

### DIFF
--- a/pkg/scheduler/plugins/rescheduling/node_utilization_util.go
+++ b/pkg/scheduler/plugins/rescheduling/node_utilization_util.go
@@ -61,8 +61,8 @@ func getNodeUtilization() []*NodeUtilization {
 		nodeUtilization := &NodeUtilization{
 			nodeInfo: nodeInfo.Node,
 			utilization: map[v1.ResourceName]float64{
-				v1.ResourceCPU:    nodeInfo.ResourceUsage.CPUUsageAvg[Interval],
-				v1.ResourceMemory: nodeInfo.ResourceUsage.MEMUsageAvg[Interval],
+				v1.ResourceCPU:    nodeInfo.ResourceUsage.CPUUsageAvg[MetricsPeriod],
+				v1.ResourceMemory: nodeInfo.ResourceUsage.MEMUsageAvg[MetricsPeriod],
 			},
 			pods: nodeInfo.Pods(),
 		}
--- a/pkg/scheduler/plugins/rescheduling/rescheduling.go
+++ b/pkg/scheduler/plugins/rescheduling/rescheduling.go
@@ -32,6 +32,8 @@ const (
 	PluginName = "rescheduling"
 	// DefaultInterval indicates the default interval rescheduling plugin works for
 	DefaultInterval = 5 * time.Minute
+	// DefaultMetricsPeriod indicates the default metrics period rescheduling plugin works for
+	DefaultMetricsPeriod = "5m"
 	// DefaultStrategy indicates the default strategy rescheduling plugin making use of
 	DefaultStrategy = "lowNodeUtilization"
 )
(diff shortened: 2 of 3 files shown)


pkg/scheduler/plugins/rescheduling/rescheduling.go
```diff
@@ -153,8 +153,6 @@ func (rc *Configs) parseArguments(arguments framework.Arguments) {
 	if err != nil {
 		klog.V(4).Infof("Parse rescheduling interval failed. Reset the interval to 5m by default.")
 		rc.interval = DefaultInterval
-	} else {
-		Interval = intervalStr
```
wangyang0616, line 157: I understand that after deleting here, the interval configuration should not take effect.

pkg/scheduler/plugins/rescheduling/node_utilization_util.go
```diff
@@ -61,8 +61,8 @@ func getNodeUtilization() []*NodeUtilization {
 		nodeUtilization := &NodeUtilization{
 			nodeInfo: nodeInfo.Node,
 			utilization: map[v1.ResourceName]float64{
-				v1.ResourceCPU:    nodeInfo.ResourceUsage.CPUUsageAvg[Interval],
-				v1.ResourceMemory: nodeInfo.ResourceUsage.MEMUsageAvg[Interval],
+				v1.ResourceCPU:    nodeInfo.ResourceUsage.CPUUsageAvg[MetricsPeriod],
```
wangyang0616, line 64: I understand that the analysis of MetricsPeriod should be added when config is analyzed, stored in the Configs structure, and then used here.

Will this implementation take effect by configuring MetricsPeriod through the yaml file?

pkg/scheduler/plugins/rescheduling/rescheduling.go
Shoothzj: @wangyang0616 IMHO, what real take effects is **rc.interval**. We don't need **Interval** variable now

pkg/scheduler/plugins/rescheduling/node_utilization_util.go
Shoothzj: @wangyang0616 Fixed

pkg/scheduler/plugins/rescheduling/rescheduling.go
wangyang0616: https://github.com/volcano-sh/volcano/blob/a592a009a4c6885cb3720ee5a0bf565db2f9abd8/pkg/scheduler/plugins/rescheduling/rescheduling.go#L95-L98
Here, interval is used to control the real running time of rescheduling.

wangyang0616: Yes, you are right.

wangyang0616:
This PR is ok for me.

william-wang (approved):
/lgtm
/approve

PR #2706, opened 2023-02-25
Supports the collection of average indicator data for multiple time periods

wangyang0616:
Only 5m of data is collected in the current cache, and other time periods configured by the user in usage will not take effect.

The current PR supports collecting the average indicator data of multiple time periods in the cache, such as:  5m, 10m, 15m, 30m, 1h, 1d,  the user can decide which indicator to use in the usage plugin.

### DIFF
--- a/docs/design/usage-based-scheduling.md
+++ b/docs/design/usage-based-scheduling.md
@@ -47,8 +47,8 @@ tiers:
       - name: usage  # usage based scheduling plugin
         arguments:
           thresholds:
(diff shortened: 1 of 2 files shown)


Shoothzj (approved):

hwdef (commented):
/lgtm

wangyang0616:
need to keep

wangyang0616:
[Obtains the actual load data of a node from the custom metrics API. #3176](https://github.com/volcano-sh/volcano/pull/3176) has implemented this capability.

/close

Issue #2672, opened 2023-02-09
Unknown flag: --enable-healthz


jesuinovieira:
**What happened**: not able to install volcano after merging pull request #2651

```
unknown flag: --enable-healthz
Usage of /vc-webhook-manager:
      --add-dir-header                   If true, adds the file directory to the header
      --admission-conf string            The configmap file of this webhook
      --alsologtostderr                  log to standard error as well as files
      --ca-cert-file string              File containing the x509 Certificate for HTTPS.
      --enabled-admission string         enabled admission webhooks, if this parameter is modified, make sure corresponding webhook configurations are the same. (default "/jobs/mutate,/jobs/validate,/podgroups/mutate,/pods/validate,/pods/mutate,/queues/mutate,/queues/validate")
      --ignored-namespaces string        Comma-separated list of namespaces to be ignored by admission webhooks (default "volcano-system,kube-system")
      --kube-api-burst int               Burst to use while talking with kubernetes apiserver (default 100)
      --kube-api-qps float32             QPS to use while talking with kubernetes apiserver (default 50)
      --kubeconfig string                Path to kubeconfig file with authorization and master location information.
      --listen-address string            The address to listen on for the admission-controller-server.
      --log-backtrace-at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log-dir string                   If non-empty, write log files in this directory
      --log-file string                  If non-empty, use this log file
      --log-file-max-size uint           Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --log-flush-frequency duration     Maximum number of seconds between log flushes (default 5s)
      --logtostderr                      log to standard error instead of files (default true)
      --master string                    The address of the Kubernetes API server (overrides any value in kubeconfig)
      --port int                         the port used by admission-controller-server. (default 8443)
      --scheduler-name stringArray       Volcano will handle pods whose .spec.SchedulerName is same as scheduler-name (default [volcano])
      --skip-headers                     If true, avoid header prefixes in the log messages
      --skip-log-headers                 If true, avoid headers when opening log files
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
      --tls-cert-file string             File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert).
      --tls-private-key-file string      File containing the default x509 private key matching --tls-cert-file.
  -v, --v Level                          number for the log level verbosity
[...]

PR #2530, opened 2022-10-11
Make helm chart imagePullPolicy configurable

kerthcet:
Signed-off-by: kerthcet <kerthcet@gmail.com>

/kind feature

We want to make volcano helm chart imagePullPolicy configurable, in case we want to make sure that we're always 
pulling the latest images.

PR #2721, opened 2023-03-05
imagePullPolicy changed from IfNotPresent to Always

wangyang0616:
When Volcano is installed on the main branch, the latest version of the image is used. 

Is it more reasonable to use the Always policy for the imagePullPolicy to avoid the user's local image cache, resulting in the inability to use the latest image?

Related questions: [#2672](https://github.com/volcano-sh/volcano/issues/2672)


hwdef (commented):
I think you should modify the file in installer/helm/chart/volcano, and run `make update-development-yaml`

wangyang0616:
> I think you should modify the file in installer/helm/chart/volcano, and run `make update-development-yaml`

Thanks for your guidance, fixed.

wangyang0616:
At present, it is not possible to change the `imagePullPolicy` to `Always`. 

e2e compiles the image locally and deploys the volcano through helm. If it is changed to `Always`, the e2e operation will pull all the newly compiled images from dockerhub, which will cause the image pull to fail, and e2e cannot run.

The current PR is closed, and more appropriate modification methods are being considered in the future

hwdef:
@wangyang0616 
I think we can reopen this https://github.com/volcano-sh/volcano/pull/2530
and change the imagepullpolicy to `IfNotPresent` during e2e test in 
https://github.com/volcano-sh/volcano/blob/master/hack/run-e2e-kind.sh#L57

hwdef (commented):
/lgtm

jasonliu747 (approved):
/lgtm

hwdef:
/assign @Thor-wl @k82cn @william-wang @shinytang6 
Please review this.

xiao-jay:
@wangyang0616 If the tag of the image has always been 'lastest', changing it to `imagePullPolicy `to ` always` is indeed a good modification.
/lgtm

william-wang (approved):
/approve

PR #2726, opened 2023-03-09
Avoid Panic when es has no data

Shoothzj:
Refactor code to handle errors and only update nodeMetrics if values are valid.


gj199575:
LGTM

william-wang (approved):
/lgtm
/approve

PR #2732, opened 2023-03-10
Skip early when reclaimees is empty

zbbkeepgoing:
When reclaimees is empty, it's no need execute ssn.Reclaimable(task, reclaimees).

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -152,6 +152,12 @@ func (ra *Action) Execute(ssn *framework.Session) {
 					reclaimees = append(reclaimees, task.Clone())
 				}
 			}
+
+			if len(reclaimees) == 0 {
+				klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
+				continue
+			}
+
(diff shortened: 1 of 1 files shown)


pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -152,6 +152,11 @@ func (ra *Action) Execute(ssn *framework.Session) {
 					reclaimees = append(reclaimees, task.Clone())
 				}
 			}
+
+			if len(reclaimees) == 0 {
+				continue
```
wangyang0616, line 157: I agree with your modification very much. Is it better to add a line of debug logs before continue?

wangyang0616: just like 
https://github.com/volcano-sh/volcano/blob/12a44e3faea7c7dc3332a96d430adae67a7c96e9/pkg/scheduler/actions/reclaim/reclaim.go#L157-L159

zbbkeepgoing: Done

Thor-wl (approved):

Issue #2734, opened 2023-03-14
If we use reclaim or preempt, spark driver will be reclaimed, it's irrational


zbbkeepgoing:
**What happened**:

If enable queue's reclaimable, pending task in high priority queue will reclaim some special task in low priority queue, such as spark driver, If spark driver is be reclaimed or preempted. the spark job will fail.

**What you expected to happen**:

Avoid reclaim or preempt some special task.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create two queue, the resource of queue need overlap.
2. Submit first spark job to one queue, this spark job need full of the reources of queue.
3. After all executor of first spark job is running, submit second spark job to another queue,  this spark job  also need full of the reources of queue.
4. The task in second spark job will reclaim the driver  of the first spark job

**Anything else we need to know?**:

If spark driver is terminating, spark job will fail, if spark executor is terminating, spark will retry. 

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zbbkeepgoing:
keep active

PR #2735, opened 2023-03-14
Fix reclaim or preempt action may preempt driver of spark job

zbbkeepgoing:
fix: #2734

### DIFF
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -29,6 +29,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/plugins/util"
+	"volcano.sh/volcano/pkg/scheduler/plugins/util/gangextender"
 )
 
 // PluginName indicates name of volcano scheduler plugin.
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/gang/gang.go
```diff
@@ -19,6 +19,8 @@ package gang
 import (
 	"fmt"
 
+	"volcano.sh/volcano/pkg/scheduler/plugins/util/gangextender"
```
wangyang0616, line 22: It is suggested that the current imported information can be adjusted down a few lines and classified together

wangyang0616:
I understand that it is not appropriate to put this in the gang scheduling strategy. Is it better to add an independent spark plugin?

zbbkeepgoing:
> I understand that it is not appropriate to put this in the gang scheduling strategy. Is it better to add an independent spark plugin?

I don't think so, because for ordinary pgs, the min member of gang is the minimum number of all pods in pgs, but for applications like Spark, the driver is actually a part of the min member of gang

pkg/scheduler/plugins/gang/gang.go
zbbkeepgoing: Done

wangyang0616:
reviewed 
cc @Yikun @william-wang

hwdef:
still needs to be reviewed

oemel09:
This should be configurable (e. g. via arguments to the gang plugin). There may be situations where preempting the spark driver is the desired behavior.

pkg/scheduler/plugins/util/gangextender/gangextender.go
```diff
@@ -0,0 +1,42 @@
...
+	sparkPodLabelRoleKey         = "spark-role"
+	sparkPodLabelRoleDriverValue = "driver"
+)
+
+func GangPreemptRejectExtender(preemptee *api.TaskInfo) bool {
+	sparkReject := sparkGangExtender(preemptee)
+	return sparkReject
+}
+
+func sparkGangExtender(preemptee *api.TaskInfo) bool {
```
lowang-bh, line 31: I think the function name is not meaningful. May let the function name has something relative with "driver role" is easy to understand.

PR #2743, opened 2023-03-20
optimize nodeorder logprint to easy find from pod;

waiterQ:
When locate the problem of assignment strategy, it's hardly figure out a task score detail in a batch assignment. Add pod name for each component score log printing in nodeorder plugin.


william-wang (approved):
/lgtm
/approve

PR #2752, opened 2023-03-24
ignore no metrics server error log;

waiterQ:
if metrics server's configuration not set, volcano-scheduler.log will have the log blow; metrics server isn't a default start server, so if metricsConf["address"] is not set, do not start scheduled task GetMetricsData.

---
  E0324 15:25:08.686552       7 cache.go:1264] Error creating client: metrics address is empty
  E0324 15:25:13.687176       7 cache.go:1264] Error creating client: metrics address is empty
  E0324 15:25:18.688364       7 cache.go:1264] Error creating client: metrics address is empty
  ...
---

And for more efficient log print, lower actions, plugins enter and leave log prints.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -42,8 +42,8 @@ func (alloc *Action) Name() string {
 func (alloc *Action) Initialize() {}
 
 func (alloc *Action) Execute(ssn *framework.Session) {
-	klog.V(3).Infof("Enter Allocate ...")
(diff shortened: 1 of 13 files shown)


Shoothzj (approved):

wangyang0616 (approved):

wangyang0616:
This is indeed very necessary. Now open the log of vc-scheduler. When there is no scheduled task, there will be too many log outputs every second. I agree with this PR.

william-wang (approved):
/lgtm
/approve

PR #2755, opened 2023-03-27
[WIP] Add namespaceOrderFn in priority plugin.

### LINKED ISSUE #2747: Priority plugin should add `NamespaceOrderFn`.
**What happened**:

We use volcano to schedule a deployment on a cluster with auto-scaler. When the deployment enlarged the replicas, the pending pods would trigger auto-scaler to add 1 more node into the cluster. We had some monitoring components deployed as `daemonsets` in `kube-system`， whose `priorityClass` was high. When scheduled by kube-scheduler, the daemonset pod would scheduld before pods from the deployment to the new node, since the daemonset was in `kube-system` and high priorityClass than deployment in `default` with default priorityClass '0'.
But in this scenario, pods of the deployment would be scheduled first, and pod of the daemonset kept Pending. After I configured `preempt` action, things got worse: the Pending pod from daemonset kept evicting Running pod from deployment, then the deployment created new pods and the new pods still got scheduled before deamonset pod. Pods in `kube-system` kept Pending.
 
This bug was probably caused by `NamespaceOrderFn`:

Firstly, in `allocate` action, volcano would sort all namespaces, then choose one job from one queue of the namespace, and schedule enough tasks of this job. the total process would be like:
https://github.com/volcano-sh/volcano/blob/c222ab487fe8f6e6491d300bec85ed489cc5a2d4/pkg/scheduler/actions/allocate/allocate.go#L48-L54

By default the `NamespaceOrdeFn` would go through all `namespaceOrderFns` registered by plugins. And if no one registered, the namespaces would be sorted by name:
https://github.com/volcano-sh/volcano/blob/c222ab487fe8f6e6491d300bec85ed489cc5a2d4/pkg/scheduler/framework/session_plugins.go#L527-L550

So in our scenario, no `namespaceOrderFn` was registered since I did not enable `drf` plugin, then the namespaces would be sorted by name. Namespace `default` would always before `kube-system`, causing priorityClass took no effect and `preempt` failed.

In my opinion, we need to add more `namespaceOrderFn` in plugins like `priority namespaceOrderFn`. Currently we only have `drf namespaceOrderFn`, which might not cover all scenarios.

**Environment**:
- Volcano Version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jiangkaihua:
fix #2747

### DIFF
--- a/pkg/scheduler/plugins/priority/priority.go
+++ b/pkg/scheduler/plugins/priority/priority.go
@@ -118,6 +118,45 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 		return ji.ReadyTaskNum()+ji.WaitingTaskNum() < int32(len(ji.Tasks))
 	}
 	ssn.AddJobStarvingFns(pp.Name(), jobStarvingFn)
+
+	namespaceOrderFn := func(l interface{}, r interface{}) int {
+		lv := l.(api.NamespaceName)
+		rv := r.(api.NamespaceName)
+
+		var lPriority int32
+		var rPriority int32
+		var lHighestJob string
+		var rHighestJob string
+
+		for _, info := range ssn.Jobs {
+			// only calculate pending jobs
+			switch info.Namespace {
+			case string(lv):
+				if lPriority < info.Priority {
+					lPriority = info.Priority
+					lHighestJob = info.Name
+				}
+			case string(rv):
+				if rPriority < info.Priority {
+					rPriority = info.Priority
+					rHighestJob = info.Name
+				}
+			}
+		}
+		klog.V(5).Infof("Priority NamespaceOrderFn: namespace <%s> priority <%d> from job <%s>, "+
+			"namespace <%s> priority <%d> from job <%s>", lv, lPriority, lHighestJob, rv, rPriority, rHighestJob)
+
+		if lPriority == rPriority {
+			return 0
+		}
+
+		if lPriority < rPriority {
+			return 1
+		}
+
+		return -1
+	}
+	ssn.AddNamespaceOrderFn(pp.Name(), namespaceOrderFn)
 }
 
 func (pp *priorityPlugin) OnSessionClose(ssn *framework.Session) {}


pkg/scheduler/plugins/priority/priority.go
```diff
@@ -118,6 +118,45 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+	namespaceOrderFn := func(l interface{}, r interface{}) int {
+		lv := l.(api.NamespaceName)
+		rv := r.(api.NamespaceName)
+
+		var lPriority int32
+		var rPriority int32
+		var lHighestJob string
+		var rHighestJob string
+
+		for _, info := range ssn.Jobs {
```
zbbkeepgoing, line 131: Is it better to prepare the results or collect the data at the beginning of the opensession?

jiangkaihua: Actually, after a job scheduled, it should be neglected during sorting namespaces. So every time we pushed a namespace back into `workqueue`, it should be resorted. It is time-costing, maybe I could optimize the sort order and save it in a heap.

zbbkeepgoing: > Actually, after a job scheduled, it should be neglected during sorting namespaces. So every time we pushed a namespace back into `workqueue`, it should be resorted. It is time-costing, maybe I could optimize the sort order and save it in a heap.

However, jobs in ssn.jobs in allocate will not be removed. So the sorting results here are always the same, it's need to consider.

jiangkaihua: Yes, I have not skip jobs in `Pending` status and `allocated` jobs, I need consider about it. Thank you!

lowang-bh:
I also add a namespaceOrder function in comformance plugin: https://github.com/volcano-sh/volcano/pull/2753/files

hwdef:
Still work in progress?

lowang-bh:
> Still work in progress?

Namespace order has been removed. I think it can be closed.

hwdef:
> > Still work in progress?
> 
> Namespace order has been removed. I think it can be closed.

you are right.

PR #2756, opened 2023-03-29
Make errorCache configurable for predicate helper

jinzhejz:
Predicate helper use an error cache for performance improvement , it is helpful for the workloads which has same pod template, such as Deployment, VolcanoJob. However, it will make pods in a DaemonSet unscheduled in some cases, because pods in a same DaemonSet has different nodeAffinity.
Error Cache is always enabled in `PredicateHelper` now, and this PR will update `PredicateHelper` interface to make it configurable.


william-wang (approved):
/lgtm
/approve

PR #2757, opened 2023-03-29
Refactor namespace fairshare function.

jiangkaihua:
Current `allocate` action must sort `namespace`, then `queue` in each namespace, then jobs in a queue of single namespace, which is time-costing and inefficient, just to realize `namespace-fairshare`. And, since `allocate` action worked differently with `preempt` and `reclaim` action, it kept causing comlicated issues in `evict` and re-allocate in next session. So I would rather realize `namespace-fairshare` by other methods instead of sort namespaces in `allocate` action.


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -84,26 +82,17 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
 
-		namespace := api.NamespaceName(job.Namespace)
-		queueMap, found := jobsMap[namespace]
-		if !found {
-			namespaces.Push(namespace)
-
-			queueMap = make(map[api.QueueID]*util.PriorityQueue)
-			jobsMap[namespace] = queueMap
-		}
+		queues.Push(ssn.Queues[job.Queue])
```
jinzhejz, line 85: It seems that it will push duplicate queues to `queues`

jiangkaihua: Yes, queue should be pushed in `queues` only on the first job in the queue is picked. I'll modify it, thank you.

william-wang (approved):
/lgtm
/approve

PR #2764, opened 2023-04-03
fix job elastic resource

lfeng-nl:
When only a part of the resource are equal, GetElasticResources will return empty, This is unreasonable.

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -485,7 +485,7 @@ func (ji *JobInfo) GetMinResources() *Resource {
 }
 
 func (ji *JobInfo) GetElasticResources() *Resource {
-	if ji.Allocated.LessEqualPartly(ji.GetMinResources(), Zero) {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -485,7 +485,7 @@ func (ji *JobInfo) GetMinResources() *Resource {
 }
 
 func (ji *JobInfo) GetElasticResources() *Resource {
-	if ji.Allocated.LessEqualPartly(ji.GetMinResources(), Zero) {
+	if ji.Allocated.LessPartly(ji.GetMinResources(), Zero) {
```
wangyang0616, line 488: I'm not sure if there are other design considerations here, but if changed to `LessPartly`, the following place may assert exit.

https://github.com/volcano-sh/volcano/blob/d5bd18b7ad8f53bf29bf49cbdbebb6d35cd90bbe/pkg/scheduler/api/resource_info.go#L245-L247

PR #2765, opened 2023-04-03
fix preempt function

lfeng-nl:
1.If other preemptors have already preemptor sufficient resources, Evict should not be executed. 
2.Queue not overused, not mean task can be allocated.


wangyang0616:
I don't quite understand the specific problem scenario of the current PR repair. 

Can you create an issue to describe the problem phenomenon and the reproduction method in detail?

Monokaix:
seems this pr https://github.com/volcano-sh/volcano/pull/3230 has already solved it.

Issue #2768, opened 2023-04-07
CI task has been waiting and cannot run


wangyang0616:
**What happened**:
Some CI tasks have been waiting and cannot run.

The prompt information is as follows：
```
Requested labels: ubuntu-18.04
Job defined at: volcano-sh/volcano/.github/workflows/e2e_scheduling_basic.yaml@refs/pull/2737/merge
Waiting for a runner to pick up this job...
```
The details are as follows:
![image](https://user-images.githubusercontent.com/36185952/230534233-4844ad3c-a557-4dce-a93e-864e5c3fc1b6.png)
![image](https://user-images.githubusercontent.com/36185952/230534267-915a32f4-1fc3-46bc-897b-eeda63f3a709.png)
![image](https://user-images.githubusercontent.com/36185952/230534312-c314b9b2-d3ae-4b48-aed9-c75f4b030564.png)

**What you expected to happen**:
CI is up and running
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**: 
- Volcano Version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

wangyang0616:
github no longer supports ubuntu 18.04 version mirroring, just upgrade to the latest version of ubuntu, such as ubuntu 22.04

PR #2769, opened 2023-04-07
Upgrade the ubuntu version to fix the problem that CI cannot run

wangyang0616:
fix: [#2768](https://github.com/volcano-sh/volcano/issues/2768)


hwdef (commented):
I think we can use the lastest stable version. such as 22.04

wangyang0616:
> I think we can use the lastest stable version. such as 22.04

done

PR #2773, opened 2023-04-11
Change the error message.

gj199575:
this MR is the relization of this issue :  https://github.com/volcano-sh/volcano/issues/2774
when pod is unscheduled, make the error message similar  to kube-scheduler。

test message
the default scheduler error message is "0/2 nodes are available: 2 Insufficient cpu."

before this MR
volcano   error message is  "all nodes are unavailable: 2/2 nodes node(s) resource fit failed."

after this MR
volcano   error message is  "all nodes are unavailable: 2/2 nodes Insufficient cpu."

And in the future。 I will continue to make volcano  error event or something which is showed to  user    is more similar  to  k8s default scheduler

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -98,8 +98,8 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	allNodes := ssn.NodeList
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) error {
 		// Check for Resource Predicate
-		if !task.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
-			return api.NewFitError(task, node, api.NodeResourceFitFailed)
+		if ok, reason := task.InitResreq.LessEqualWithReason(node.FutureIdle(), api.Zero); !ok {
+			return api.NewFitError(task, node, reason)
 		}
 
 		return ssn.PredicateFn(task, node)
--- a/pkg/scheduler/api/job_info_test.go
+++ b/pkg/scheduler/api/job_info_test.go
@@ -231,7 +231,7 @@ func TestTaskSchedulingReason(t *testing.T) {
 				t3.UID: "Pod ns1/task-3 can possibly be assigned to node1",
 				t4.UID: "Pod ns1/task-4 can possibly be assigned to node2",
 				t5.UID: "Pod ns1/task-5 can possibly be assigned to node3",
-				t6.UID: "all nodes are unavailable: 1 node(s) pod number exceeded, 2 node(s) resource fit failed.",
+				t6.UID: "0/3 nodes are unavailable: 1 node(s) pod number exceeded, 2 node(s) resource fit failed.",
 			},
 		},
 	}
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -389,6 +389,38 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
 	return true
 }
 
+// LessEqualWithReason returns true, "" only on condition that all dimensions of resources in r are less than or equal with that of rr,
+// Otherwise returns false and err string ,which show which resource is insufficient.
+// @param defaultValue "default value for resource dimension not defined in ScalarResources. Its value can only be one of 'Zero' and 'Infinity'"
+// this function is the same as LessEqual , and it will be merged to LessEqual in the future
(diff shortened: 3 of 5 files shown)


wangyang0616:
Hello, you are very welcome to participate in community contributions. 

From your description, you can see the events and error log information that will continue to be output by Volcano. 

I think it would be better to create an issue and classify similar PRs in a unified way. 
Thanks！

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -98,8 +98,8 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	allNodes := ssn.NodeList
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) error {
 		// Check for Resource Predicate
-		if !task.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
-			return api.NewFitError(task, node, api.NodeResourceFitFailed)
+		if reason := task.InitResreq.LessEqualResource(node.FutureIdle(), api.Zero); len(reason) != 0 {
```
wangyang0616, line 101: Why add a new function here? If a return value is added on the basis of the original function, it indicates whether the resource information of the error is acceptable.

In addition, there are many places where lessEqual is called. Do other places need to be modified or modified iteratively?

gj199575: Why add a new function here?
Because the purpose of this PR is  to make volcano  event  belong to  a unscheduled pod ,  is more similar to kube-scheduler 。
So I do not need to change  many codes。
And the  LessEqual function is used by many places , if I change   LessEqual function , it is  a  huge project。
So I choose  a  better way to solve my problem ，and do not change volcano a  lot。

gj199575:
> Hello, you are very welcome to participate in community contributions.
> 
> From your description, you can see the events and error log information that will continue to be output by Volcano.
> 
> I think it would be better to create an issue and classify similar PRs in a unified way. Thanks！

ok， I will create a  issue

wangyang0616:
Please refer to the prompt information [DCO](https://github.com/volcano-sh/volcano/pull/2773/checks?check_run_id=12653283611) to repair CI

pkg/scheduler/actions/allocate/allocate.go
wangyang0616: If you are worried that the impact of the modification is too large, you can add a new function, but the return value of the new function is recommended to return two (bool, string), the first parameter indicates the result, and the second parameter indicates msg, in the return value character Whether the string is empty indicates the result of the judgment value, which feels inappropriate

gj199575:
> Please refer to the prompt information [DCO](https://github.com/volcano-sh/volcano/pull/2773/checks?check_run_id=12653283611) to repair CI

ok

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -98,8 +98,8 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	allNodes := ssn.NodeList
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) error {
 		// Check for Resource Predicate
-		if !task.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
-			return api.NewFitError(task, node, api.NodeResourceFitFailed)
+		if ok, reason := task.InitResreq.CheckResource(node.FutureIdle(), api.Zero); !ok {
```
william-wang, line 101: @gj199575 LessEqual is an imporment foundamental function. What's the difference between the CheckResource and LessEqual, and why we have to add this new function, Would you provide detail reason?

gj199575: yes, LessEqual is used by many place, And the new function CheckResource   is  ready to slove the problem  in  https://github.com/volcano-sh/volcano/issues/2774.  But it is the same as LessEqual  in code。
And after I solve  https://github.com/volcano-sh/volcano/issues/2774 。 
I will continue change to  merge LessEqual and CheckResource to a function 。

gj199575: > @gj199575 LessEqual is an imporment foundamental function. What's the difference between the CheckResource and LessEqual, and why we have to add this new function, Would you provide detail reason?

And it is a huge project if I change LessEqual function .So I supply a new function LessEqualWithReason 。 And I will  merge LEssEqual and LessEqualWithReason  to a funtion 。

william-wang (approved):
/lgtm
/approve

william-wang (approved):
/lgtm
/approve

PR #2784, opened 2023-04-13
Ensure deletion of pods is handled

Monokaix:
/kind bug

When a watch pod deletion event was missed while disconnected from apiserver, delta fifo in informer will generate a new delete event, and a `DeletedFinalStateUnknown` obj will be enqueue and reprocessed, here is this logic：
https://github.com/kubernetes/kubernetes/blob/297dc976b96895332f8e800c531f47d5cca68bfd/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L665

We have considered the case in `DeletePod` event  handler:
https://github.com/volcano-sh/volcano/blob/bed4a044b5f931fcb566b7ccbe2292207f7db346/pkg/scheduler/cache/event_handlers.go#L269 but not in filter func, which will lead event handler have no chance to handle pod deletion miss event, so we should check in filter func first.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-	"volcano.sh/volcano/pkg/scheduler/metrics/source"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


william-wang (approved):
/lgtm
/approve

PR #2786, opened 2023-04-13
Modify binpack score on nodes with releasing resources.

### LINKED ISSUE #2782: Sometimes `binpack` plugin score nodes with `terminating` pods too high, causing pending pods go to `pipeline` but not `allocated`
**What happened**:

In `allocate` action, nodes with `terminating` pods would also be permitted to go through `predicateFn` as long as  `futureIdle`(`idle` + `releasing`) is larger than pod request: 

https://github.com/volcano-sh/volcano/blob/bed4a044b5f931fcb566b7ccbe2292207f7db346/pkg/scheduler/actions/allocate/allocate.go#L99-L106

So nodes with `terminating` pods could be candidates in `score` period even if they would not provide enough resources for pending pod to allocate immediately. Therefore in some scenarios nodes with `terminating` pods would score higher than nodes with enough idle resources.

Nodes with `terminating` pods would got less score in `score` period than nodes with enough `idle` resources.

1. Enable `binpack` plugin and give it a high weight;
2. Give different weight for each resources, like:
```yaml
- name: binpack
  arguments:
    binpack.cpu: 8
    binpack.memory: 1
    binpack.weight: 4
```
3. Compose a scenario that a pod kept in`terminating` on a node, whose `idle` resource not meet memory request of `pending` pod, but cpu is enough. And its `idle` + `releasing` resources just fit the `pending` pod.(idle.mem+releasing.mem=request.mem, idle.cpu=request.cpu)
4. In `binpack` plugin, it would score `0` for memory:
https://github.com/volcano-sh/volcano/blob/bed4a044b5f931fcb566b7ccbe2292207f7db346/pkg/scheduler/plugins/binpack/binpack.go#L248-L251
5. But for cpu, it would return non-zero score since cpu is enough, And the total score would be high since cpu weight is `8/9`:
https://github.com/volcano-sh/volcano/blob/bed4a044b5f931fcb566b7ccbe2292207f7db346/pkg/scheduler/plugins/binpack/binpack.go#L229-L239
6. Then `pending` pod would be `pipelined` to this node even if the other nodes owned enough idle sources, Because `binpack` plugin would score this node higher than others.

**Environment**:
- Volcano Version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jiangkaihua:
fix #2782

### DIFF
--- a/pkg/scheduler/plugins/binpack/binpack.go
+++ b/pkg/scheduler/plugins/binpack/binpack.go
@@ -223,8 +223,14 @@ func BinPackingScore(task *api.TaskInfo, node *api.NodeInfo, weight priorityWeig
 			continue
 		}
 
-		resourceScore := ResourceBinPackingScore(request, allocate, nodeUsed, resourceWeight)
-		klog.V(5).Infof("task %s/%s on node %s resource %s, need %f, used %f, allocatable %f, weight %d, score %f", task.Namespace, task.Name, node.Name, resource, request, nodeUsed, allocate, resourceWeight, resourceScore)
+		resourceScore, err := ResourceBinPackingScore(request, allocate, nodeUsed, resourceWeight)
+		if err != nil {
+			klog.V(4).Infof("task %s/%s cannot binpack node %s: resource: %s is %s, need %f, used %f, allocatable %f",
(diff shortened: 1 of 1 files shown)


william-wang (approved):
/lgtm
/approve

PR #2793, opened 2023-04-15
improve: rename variable of task spec name for more readable

lowang-bh:
`taskName` for task role has some ambiguity to understand, `specName` is more meaningful for task role, in my opinion

### DIFF
--- a/pkg/controllers/apis/job_info.go
+++ b/pkg/controllers/apis/job_info.go
@@ -30,7 +30,7 @@ type JobInfo struct {
 	Name      string
 
 	Job  *batch.Job
-	Pods map[string]map[string]*v1.Pod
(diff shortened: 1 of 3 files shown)


pkg/controllers/apis/job_info.go
```diff
@@ -30,7 +30,7 @@ type JobInfo struct {
 	Name      string
 
 	Job  *batch.Job
-	Pods map[string]map[string]*v1.Pod
+	Pods map[string]map[string]*v1.Pod // roleSpec-> podName->podObject
```
wangyang0616, line 33: I understand that this should be the task name.

hwdef (commented):
The taskname is indeed somewhat ambiguous, but since many users are already familiar with the current name, it is not recommended to modify it

PR #2794, opened 2023-04-15
improve: check whether node's idle resource is enough for task bounden to it

lowang-bh:
There is probability `OutOfxxx` when multiple schedulers schedule pods to same node. This trick will reduce the probability and resync pod if conflict happens 

enhancement in multiple-schedulers scenario

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -75,6 +75,8 @@ type ServerOption struct {
 
 	NodeSelector      []string
 	EnableCacheDumper bool
+	// check conflicts once more before binding, for multiple schedulers in same cluster
+	EnableConflictCheck bool
 }
 
 type DecryptFunc func(c *ServerOption) error
@@ -128,6 +130,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")
 	fs.StringSliceVar(&s.NodeSelector, "node-selector", nil, "volcano only work with the labeled node, like: --node-selector=volcano.sh/role:train --node-selector=volcano.sh/role:serving")
 	fs.BoolVar(&s.EnableCacheDumper, "cache-dumper", true, "Enable the cache dumper, it's true by default")
+	fs.BoolVar(&s.EnableConflictCheck, "conflict-check", false, "Optimized for the scenario where multiple schedulers are used in the cluster, it's false by default")
 }
 
 // CheckOptionOrDie check lock-object-namespace when LeaderElection is enabled.
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -87,7 +87,7 @@ func New(config *rest.Config, schedulerNames []string, defaultQueue string, node
 
 // SchedulerCache cache for the kube batch
 type SchedulerCache struct {
-	sync.Mutex
+	sync.RWMutex
 
 	kubeClient   *kubernetes.Clientset
 	restConfig   *rest.Config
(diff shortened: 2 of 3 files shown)


wangyang0616:
Please rebase latest master and make ci happy. Thanks!

lowang-bh:
> Please rebase latest master and make ci happy. Thanks!

Done. @Thor-wl @hwdef would you plz help to review this pr? thanks.

hwdef:
> > Please rebase latest master and make ci happy. Thanks!
> 
> Done. @Thor-wl @hwdef would you plz help to review this pr? thanks.

I'll check this tomorrow

hwdef (commented):
I don't know whether this solution is useful, because the original issue mentioned that he only used the volcano scheduler and did not use the k8s native scheduler.
If we want to solve the problem of multiple schedulers, will it be more useful for us to shorten the cycle of resource synchronization?

hwdef (commented):
I don't know whether this solution is useful, because the original issue mentioned that he only used the volcano scheduler and did not use the k8s native scheduler.
If we want to solve the problem of multiple schedulers, will it be more useful for us to shorten the cycle of resource synchronization?

lowang-bh:
> I don't know whether this solution is useful, because the original issue mentioned that he only used the volcano scheduler and did not use the k8s native scheduler. If we want to solve the problem of multiple schedulers, will it be more useful for us to shorten the cycle of resource synchronization?

Actually, it is used to solve the conflicts among multiple schedulers.  Image this scenario, volcano finished snapshot and is  scheduling and binding a pod-A to the node. At this same cycle, another scheduler bind a pod-B to the same node successfully before volcano. This action is not visiable to volcano at this moment.  Then pod-A bond to node and kubelet reject it. 

I used this trick and solved most of the conflicts.

hwdef:
Can this pr solve the problem in that issue, because that issue only uses the volcano scheduler, if it is not to solve the problem encountered by that issue, I suggest not to associate that issue

I think there is no problem with the code. The explanation of the flag can be optimized. For example, it can be explained that this flag is optimized for the scenario where multiple schedulers are used in the cluster.

In addition, I also want to ask, I saw that the newly added source code has a lock operation on the cache. Have you noticed that this operation will cause performance degradation?

lowang-bh:
> Can this pr solve the problem in that issue, because that issue only uses the volcano scheduler, if it is not to solve the problem encountered by that issue, I suggest not to associate that issue
> 
> I think there is no problem with the code. The explanation of the flag can be optimized. For example, it can be explained that this flag is optimized for the scenario where multiple schedulers are used in the cluster.
> 
> In addition, I also want to ask, I saw that the newly added source code has a lock operation on the cache. Have you noticed that this operation will cause performance degradation?

1. removed the assosicated issue
2. will update the prompt message
3. Yes, the lock is necessary and can not be ignored, so I add a switch to enable it only when it is needed. Another way is to use rwlock.  

BTW, the lock in cache is so large, maybe need to split to small lock with sharding lock.

lowang-bh:
Another issue: snapshot can use rlock at first, because it just needs to read cache. But now there is a code`RefreshNumaSchedulerInfoByCrd` to do write operation. That looks unreasonable.

```go
func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
	sc.Mutex.Lock()
	defer sc.Mutex.Unlock()

	snapshot := &schedulingapi.ClusterInfo{
		Nodes:          make(map[string]*schedulingapi.NodeInfo),
		Jobs:           make(map[schedulingapi.JobID]*schedulingapi.JobInfo),
		Queues:         make(map[schedulingapi.QueueID]*schedulingapi.QueueInfo),
		NamespaceInfo:  make(map[schedulingapi.NamespaceName]*schedulingapi.NamespaceInfo),
		RevocableNodes: make(map[string]*schedulingapi.NodeInfo),
		NodeList:       make([]string, len(sc.NodeList)),
		CSINodesStatus: make(map[string]*schedulingapi.CSINodeStatusInfo),
	}

	copy(snapshot.NodeList, sc.NodeList)
	for _, value := range sc.Nodes {
		value.RefreshNumaSchedulerInfoByCrd()
	}
```

hwdef (commented):
/lgtm

hwdef:
> BTW, the lock in cache is so large, maybe need to split to small lock with sharding lock.

Looking forward to your pr if you have time.

lowang-bh:
still need.
/assign @wangyang0616 @Thor-wl @william-wang

Monokaix:
The `OutOfSync` state has been removed, do we still need this? BTW, in multi scheduler case, confict seems still exists if we 
 just modify a single scheduler because it's a distributed system.

lowang-bh:
> BTW, in multi scheduler case, confict seems still exists if we just modify a single scheduler because it's a distributed system.

This change cannot solve conflicts completely，just reduce the probability of conflicts.

PR #2797, opened 2023-04-17
Enhancement: ambiguity variable refact: rename taskName to taskSpec for each task role

lowang-bh:
1. ambiguity variable refact

### DIFF
--- a/pkg/controllers/apis/job_info.go
+++ b/pkg/controllers/apis/job_info.go
@@ -63,50 +63,45 @@ func (ji *JobInfo) SetJob(job *batch.Job) {
 // AddPod adds the k8s pod object values to the Pods field
 // of JobStruct if it doesn't exist. Otherwise it throws error.
 func (ji *JobInfo) AddPod(pod *v1.Pod) error {
-	taskName, found := pod.Annotations[batch.TaskSpecKey]
+	taskSpec, found := pod.Annotations[batch.TaskSpecKey]
 	if !found {
-		return fmt.Errorf("failed to find taskName of Pod <%s/%s>",
-			pod.Namespace, pod.Name)
+		return fmt.Errorf("failed to find task-spec in annotation of Pod <%s/%s>", pod.Namespace, pod.Name)
 	}
 
 	_, found = pod.Annotations[batch.JobVersion]
 	if !found {
-		return fmt.Errorf("failed to find jobVersion of Pod <%s/%s>",
-			pod.Namespace, pod.Name)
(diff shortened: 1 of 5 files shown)


wangyang0616:
There should be no correspondence between the modified content and the associated issue.

lowang-bh:
Hi, @wangyang0616 @jiangkaihua @Thor-wl Would you please help to review this changes, please? thanks

pkg/scheduler/api/job_info.go
```diff
@@ -142,9 +142,10 @@ func getJobID(pod *v1.Pod) JobID {
 	return ""
 }
 
-func getTaskID(pod *v1.Pod) TaskID {
```
william-wang, line 145: TaskID is easy to read and it stands for UID.  Do not suggest to change it.

pkg/controllers/job/helpers/helpers.go
```diff
@@ -121,22 +121,22 @@ func GetJobKeyByReq(req *apis.Request) string {
 }
 
 // GetTasklndexUnderJob return index of the task in the job.
-func GetTasklndexUnderJob(taskName string, job *batch.Job) int {
+func GetTasklndexUnderJob(specName string, job *batch.Job) int {
```
william-wang, line 124: GetTasklndexUnderJob function is used to find task index in job. The `taskName`  is more straightforward and generic.

pkg/controllers/apis/job_info.go
```diff
@@ -63,50 +63,45 @@ func (ji *JobInfo) SetJob(job *batch.Job) {
 // AddPod adds the k8s pod object values to the Pods field
 // of JobStruct if it doesn't exist. Otherwise it throws error.
 func (ji *JobInfo) AddPod(pod *v1.Pod) error {
-	taskName, found := pod.Annotations[batch.TaskSpecKey]
```
william-wang, line 66: Seems taskName is more easier to understand.

lowang-bh: In my opinion, `taskName` equals to name of pod which treats as task.  `specName` is a kind to distinguish from eath other's role in job, suce as master, worker, ps, chief.  I think we can be more strict on the variable name and make code have high readability.

pkg/scheduler/api/job_info.go
lowang-bh: Yeah, TaskID stands for UID, but the content of this func does not return UID at all.

lowang-bh:
still need

Issue #2739, opened 2023-03-15
Reclaim is not working with GPUSharingEnable if resource used is volcano.sh/gpu-memory


igormish:
**What happened**:
One node with 480 memory (factor 10)
Having two queues high with weight 3 and low with weight 1.
After adding 4 jobs with requested & limit with volcano.sh/gpu-memory=120 to low queue (weight 1).
Adding new job to high  with requested & limit volcano.sh/gpu-memory=120 queue with (weight 1) volcano does not reclaim the any one of the jobs running in the low queue (weight 1) 

**What you expected to happen**:
I expect one of the jobs in low queue (weight 1)  to move to pending state so that the job in the high queue (weight 3) can start.

**How to reproduce it (as minimally and precisely as possible)**:
Node with allocatble resources 
```
  volcano.sh/gpu-memory:  480
  volcano.sh/gpu-number:  4
```
Have two queues one with low weight 
```
kind: Queue
metadata:
  name: low
spec:
  reclaimable: true
  weight: 1
```
And queue with higher weight
```
kind: Queue
metadata:
  name: low
spec:
  reclaimable: true
  weight: 3
```
3 Jobs for running in low queue 
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  annotations:
    volcano.sh/preemptable: "true"
    volcano.sh/revocable-zone: '*'
  labels:
    volcano.sh/preemptable: "true"
  name: low //1-4
  namespace: gpu-cluster
spec:
  maxRetry: 3
  minAvailable: 1
  plugins:
    env: []
    ssh: []
    svc:
    - --disable-network-policy=true
  policies:
  - action: RestartJob
    event: PodEvicted
  - action: RestartJob
    event: PodFailed
  queue: low
  schedulerName: volcano
  tasks:
  - maxRetry: 3
    minAvailable: 1
    name: low //1-4
    replicas: 1
    template:
      metadata:
        annotations:
          volcano.sh/preemptable: "true"
          volcano.sh/revocable-zone: '*'
        labels:
          volcano.sh/preemptable: "true"
        name: web
      spec:
        containers:
        - image: docker.io/library/nginx:1.14.2
          name: nginx
          resources:
            limits:
              volcano.sh/gpu-memory: 120
            requests:
              volcano.sh/gpu-memory: 120
        tolerations:
        - effect: NoSchedule
          key: gpu_cluster
          operator: Equal
          value: "true"  
```

Job for running in high queue 
```
[...]

PR #2799, opened 2023-04-18
Revise the checkNodeGPUSharingPredicate function to evaluate whether a pod can fit within a node, instead of checking for sufficient remaining CPU as currently implemented by the Allocate function in device_info. 

igormish:
Fix issue #2739 

Revise the checkNodeGPUSharingPredicate function in pkg/scheduler/api/devices/nvidia/gpushare/share.go to evaluate whether a pod can fit within a node, instead of checking for sufficient remaining CPU as currently implemented by the Allocate function in device_info. This would differentiate its functionality from FilterNode.

In pkg/scheduler/api/shared_device_pool.go, we can observe the following:

// FilterNode assesses if the 'pod' can be accommodated in the current node
FilterNode(pod *v1.Pod) (bool, error)

// Allocate function used in predicate
Allocate(kubeClient kubernetes.Interface, pod *v1.Pod) error

From the code, it is evident that FilterNode determines whether a pod fits in the current node, but it does not evaluate its suitability for backfill, which is handled by the Allocate function.

### DIFF
--- a/pkg/scheduler/api/devices/nvidia/gpushare/share.go
+++ b/pkg/scheduler/api/devices/nvidia/gpushare/share.go
@@ -125,7 +125,7 @@ func checkNodeGPUSharingPredicate(pod *v1.Pod, gs *GPUDevices) (bool, error) {
 	if getGPUMemoryOfPod(pod) <= 0 {
 		return true, nil
 	}
-	ids := predicateGPUbyMemory(pod, gs)
+	ids := predicateGPUbyTotalMemory(pod, gs)
 	if len(ids) == 0 {
 		return false, fmt.Errorf("no enough gpu memory on node %s", gs.Name)
 	}
@@ -142,6 +142,21 @@ func checkNodeGPUNumberPredicate(pod *v1.Pod, gs *GPUDevices) (bool, error) {
 		return false, fmt.Errorf("no enough gpu number on node %s", gs.Name)
 	}
 	return true, nil
+}
+
+// predicateGPUbyTotalMemory returns the available GPU ID on node
+func predicateGPUbyTotalMemory(pod *v1.Pod, gs *GPUDevices) []int {
+	gpuRequest := getGPUMemoryOfPod(pod)
+	allocatableGPUs := getDevicesAllGPUMemory(gs)
+	var devIDs []int
+
+	for devID := range allocatableGPUs {
+		if availableGPU, ok := allocatableGPUs[devID]; ok && availableGPU >= gpuRequest {
+			devIDs = append(devIDs, devID)
+		}
+	}
+	sort.Ints(devIDs)
+	return devIDs
 }
 
 // predicateGPUbyMemory returns the available GPU ID


pkg/scheduler/api/devices/nvidia/gpushare/share.go
```diff
@@ -125,7 +125,7 @@ func checkNodeGPUSharingPredicate(pod *v1.Pod, gs *GPUDevices) (bool, error) {
 	if getGPUMemoryOfPod(pod) <= 0 {
 		return true, nil
 	}
-	ids := predicateGPUbyMemory(pod, gs)
+	ids := predicateGPUbyTotalMemory(pod, gs)
```
wangyang0616, line 128: I think this modification can solve the problem of resource filtering in gpuShare preemption mode. Does GPUNumber also need to do similar repairs?

igormishsky: I believe that within the checkNodeGPUNumberPredicate function, the line:
`ids := predicateGPUbyNumber(pod, gs)`
should be replaced with:
`ids := predicateGPUbyTotalNumber(pod, gs)` //needs to be added

This is because the original function, predicateGPUbyNumber(pod, gs), utilizes allocatableGPUs := getDevicesIdleGPUs(gs).
As a result, if a node has all its GPU devices in use, it will not be considered for preemption at all.

If possible, I would like to address this issue in two separate steps and submit the subsequent fix in another pull request. This is because our cluster currently requires this solution.

wangyang0616:
I have no problem, please help to review.

archlitchi:
@igormish i think we should adjust nodeOrderFn accordingly. otherwise, a high-priority task will try to preempt a low-priorty task even if there are other 'gpu-idle' nodes for it to fit in.

igormishsky:
> @igormish i think we should adjust nodeOrderFn accordingly. otherwise, a high-priority task will try to preempt a low-priorty task even if there are other 'gpu-idle' nodes for it to fit in.

I tested two primary scenarios:
Same queue: From my tests, preemption does not occur when all tasks are running in the same queue.
Different queue: I was unable to find an instance where the pod wasn't allocated, even after testing numerous examples.

This pull request only serves to resolve a bug that hinders the functionality of the reclaim/preempt feature, rendering our GPU memory queue management inoperative.
Such occurrences will only transpire if preempt is set prior to allocate; otherwise, it will simply allocate, eliminating the need for preempt or reclaim. I'm uncertain if anyone even configures preempt or reclaim before allocate.

Given these findings, I propose we proceed with the pull request to address the critical issue at hand. I am more than willing to continue collaborating and testing (including deployment in production) due to the extensive impact on our production.

Additionally, if you are aware of a scenario where this issue occurs, as I have been unable to identify one, I would be eager to learn more and explore potential solutions.

archlitchi:
> 

The NodeOrderFn, or "Score" step in our current scheduling don't take GPU into consideration, which may lead to a task been scheduled to a "GPU-busy" node(node A) other than a "GPU-idle" node(node B). And may fail the predicate.allocate in that process(because all gpu resources in node A have been allocated).  
I'm not sure what will happen next, does it abandon node A, and try node B? or does it preempt low-priority task in node A in order to make room for it? I think "preempt" here will be problematic, because you shouldn't schedule this task to a "GPU-busy" node in the first place.
I'm not sure if you encountered any of these "problematic preempts" in your tests with this modification, if not, then I think it's safe to merge:)

igormishsky:
> > 
> 
> The NodeOrderFn, or "Score" step in our current scheduling don't take GPU into consideration, which may lead to a task been scheduled to a "GPU-busy" node(node A) other than a "GPU-idle" node(node B). And may fail the predicate.allocate in that process(because all gpu resources in node A have been allocated). I'm not sure what will happen next, does it abandon node A, and try node B? or does it preempt low-priority task in node A in order to make room for it? I think "preempt" here will be problematic, because you shouldn't schedule this task to a "GPU-busy" node in the first place. I'm not sure if you encountered any of these "problematic preempts" in your tests with this modification, if not, then I think it's safe to merge:)

This is the current scheduler conf I used for testing 
actions: "enqueue, allocate, reclaim, preempt "
tiers:
- plugins:
  - name: priority
  - name: gang
  - name: drf
- plugins:
  - name: proportion
  - name: predicates
    arguments:
      predicate.GPUSharingEnable: true # enable gpu sharing.
  - name: drf
  - name: overcommit
    arguments:
      overcommit-factor: 10.0
  - name: nodeorder
  - name: binpack
 
I will ran some more tests to be more sure, can you think of something to adjust to make sure I don't miss something ?

wangyang0616:
> @igormish i think we should adjust nodeOrderFn accordingly. otherwise, a high-priority task will try to preempt a low-priorty task even if there are other 'gpu-idle' nodes for it to fit in.

I understand this is another issue and needs to be extended and implemented in `nodeorder.go`. The current PR focuses on solving the problem that the preemption cannot be carried out.

wangyang0616:
There are some ideas for updating the current problem in [#2739](https://github.com/volcano-sh/volcano/issues/2739), and a repair PR is also proposed. We can discuss which repair solution is more suitable.
@archlitchi @igormishsky

Issue #2774, opened 2023-04-11
Make  volcano  scheduling fail event belong to a pod , is  more similar to kube-scheduler 。


gj199575:
#### What would you like to be added:
change some event string in volcano.
make volcano is more similar to kube-scheduler

#### Why is this needed:
And kube-scheduler is more popular 。
I think volcano should obey some rules made by kube-scheduler。
and it will make usrs using volcano  feel more fluent。

william-wang:
@gj199575 Would you describe what exactly the ploblem is and what the behavor you are expectling?

Monokaix:
related pr:https://github.com/volcano-sh/volcano/pull/2773

PR #2800, opened 2023-04-19
this commit continues to improve the function LessEqual,and make LessEqual is the same  which is better to code  quality

gj199575:
this commit continues to improve the function LessEqual,and make LessEqual is the same  which is better to code  quality
and  is linked to issue https://github.com/volcano-sh/volcano/issues/2774


jiangkaihua:
I do not think it a good idea to replace `LessEqual` with `LessEqualWithReason`, since in most scenarios we do not need to know which resource is insufficient.

pkg/scheduler/api/resource_info.go
```diff
@@ -362,31 +362,8 @@ func (r *Resource) Less(rr *Resource, defaultValue DimensionDefaultValue) bool {
...
-		if !ok && defaultValue == Infinity {
-			continue
-		}
-
-		if !lessEqualFunc(leftValue, rightValue, minResource) {
-			return false
-		}
-	}
-	return true
+	isLess, _ := r.LessEqualWithReason(rr, defaultValue)
```
wangyang0616, line 365: The naming of `isLess` should be inaccurate, it should be `isLessEqual`, or it can be named `res` directly

Issue #2798, opened 2023-04-18
feat: add leader election options


CharlesQQ:
#### What would you like to be added:
`volcano-controller-manager`/`volcano-scheduler`: Introduced `--leader-elect-lease-duration`, `--leader-elect-renew-deadline` and `--leader-elect-retry-period` flags to specify leader election behavious.
#### Why is this needed:
allow user change leader election action

lowang-bh:
`--leader-elect` will enable leader elect.

wangyang0616:
Yes, Volcano currently supports `leader-elect` for election, but it does not provide user-defined related behavior parameters. 

Have you encountered any election-related problems in some scenarios? Can you describe it in detail?

CharlesQQ:
> Volcano currently supports `leader-elect` for election, but it does not provide user-defined related behavior parameters.
> 
> Have you encountered any election-related problems in some scenarios? Can you describe it in detail?
here are volcano default parameters, that means in 10 seconds, master can fetch and update configmap lock 2 times, if after try more than 2 times and still update lock resource failed because of network problems, volcano will changed master; But, We thinks in this scenarios, master process should not exit; so uses hope to defined  `renewDeadline` by themself  may be.
```
	leaseDuration = 15 * time.Second
	renewDeadline = 10 * time.Second
	retryPeriod   = 5 * time.Second
```

PR #2809, opened 2023-04-21
fix(app):  Introduced flag to specify leader election behavious

CharlesQQ:
resolve: https://github.com/volcano-sh/volcano/issues/2798

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -21,6 +21,9 @@ import (
 	"os"
 
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/component-base/config"
+	componentbaseconfigvalidation "k8s.io/component-base/config/validation"
 
 	"volcano.sh/volcano/pkg/kube"
 )
@@ -38,16 +41,18 @@ const (
 
 // ServerOption is the main context object for the controllers.
 type ServerOption struct {
-	KubeClientOptions    kube.ClientOptions
-	CertFile             string
-	KeyFile              string
-	CaCertFile           string
-	CertData             []byte
-	KeyData              []byte
-	CaCertData           []byte
-	EnableLeaderElection bool
-	LockObjectNamespace  string
-	PrintVersion         bool
+	KubeClientOptions kube.ClientOptions
+	CertFile          string
+	KeyFile           string
+	CaCertFile        string
+	CertData          []byte
+	KeyData           []byte
+	CaCertData        []byte
+	// leaderElection defines the configuration of leader election.
+	LeaderElection config.LeaderElectionConfiguration
+	// Deprecated: use ResourceNamespace instead.
+	LockObjectNamespace string
+	PrintVersion        bool
 	// WorkerThreads is the number of threads syncing job operations
 	// concurrently. Larger number = faster job updating, but more CPU load.
 	WorkerThreads uint32
@@ -84,9 +89,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 		"File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated "+
 		"after server cert).")
 	fs.StringVar(&s.KeyFile, "tls-private-key-file", s.KeyFile, "File containing the default x509 private key matching --tls-cert-file.")
-	fs.BoolVar(&s.EnableLeaderElection, "leader-elect", true, "Start a leader election client and gain leadership before "+
-		"executing the main loop. Enable this when running replicated vc-controller-manager for high availability; it is enabled by default")
-	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default")
+	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default, will deprecated, please use --leader-elect-resource-namespace instead.")
 	fs.Float32Var(&s.KubeClientOptions.QPS, "kube-api-qps", defaultQPS, "QPS to use while talking with kubernetes apiserver")
 	fs.IntVar(&s.KubeClientOptions.Burst, "kube-api-burst", defaultBurst, "Burst to use while talking with kubernetes apiserver")
 	fs.BoolVar(&s.PrintVersion, "version", false, "Show version and quit")
@@ -100,12 +103,9 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 }
 
-// CheckOptionOrDie checks the LockObjectNamespace.
+// CheckOptionOrDie check leader election flag when LeaderElection is enabled.
 func (s *ServerOption) CheckOptionOrDie() error {
-	if s.EnableLeaderElection && s.LockObjectNamespace == "" {
-		return fmt.Errorf("lock-object-namespace must not be nil when LeaderElection is enabled")
-	}
-	return nil
+	return componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
 }
 
 // readCAFiles read data from ca file path
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -23,20 +23,20 @@ import (
 
 	"github.com/spf13/pflag"
 	_ "go.uber.org/automaxprocs"
-
 	"k8s.io/apimachinery/pkg/util/wait"
 	cliflag "k8s.io/component-base/cli/flag"
+	componentbaseoptions "k8s.io/component-base/config/options"
 	"k8s.io/klog/v2"
 
+	"volcano.sh/volcano/cmd/controller-manager/app"
+	"volcano.sh/volcano/cmd/controller-manager/app/options"
 	_ "volcano.sh/volcano/pkg/controllers/garbagecollector"
 	_ "volcano.sh/volcano/pkg/controllers/job"
 	_ "volcano.sh/volcano/pkg/controllers/jobflow"
 	_ "volcano.sh/volcano/pkg/controllers/jobtemplate"
 	_ "volcano.sh/volcano/pkg/controllers/podgroup"
 	_ "volcano.sh/volcano/pkg/controllers/queue"
-
-	"volcano.sh/volcano/cmd/controller-manager/app"
-	"volcano.sh/volcano/cmd/controller-manager/app/options"
+	commonutil "volcano.sh/volcano/pkg/util"
 	"volcano.sh/volcano/pkg/version"
 )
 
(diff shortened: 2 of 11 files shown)


lowang-bh:
I think we should do some checks between those three parameters

CharlesQQ:
cc @wangyang0616

wangyang0616:
I agree with @lowang-bh is suggestion that the input parameters need to be verified.

CharlesQQ:
> 
@lowang-bh @wangyang0616 
there have the checks for three parameter in leaderelection package, so no need to check again.
image

CharlesQQ:
cc @wangyang0616

cmd/controller-manager/app/options/options.go
```diff
@@ -83,6 +93,19 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.EnableLeaderElection, "leader-elect", true, "Start a leader election client and gain leadership before "+
 		"executing the main loop. Enable this when running replicated vc-controller-manager for high availability; it is enabled by default")
 	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default")
+	fs.DurationVar(&s.LeaseDuration, "leader-elect-lease-duration", defaultElectionLeaseDuration, ""+
```
wangyang0616, line 96: I'm thinking here mainly about the string here, whether to comply with the 120-character requirement, or to support legibility, is it easier to read if the two lines of text description information are closed?

CharlesQQ: > I'm thinking here mainly about the string here, whether to comply with the 120-character requirement, or to support legibility, is it easier to read if the two lines of text description information are closed?

It's same as kubernetes comments, sames that comply with the 120-character requirement
https://github.com/kubernetes/kubernetes/blob/ea30d100f6b13a47b4153362698b89256077776c/staging/src/k8s.io/component-base/config/options/leaderelectionconfig.go#L30C1-L42

cmd/controller-manager/app/options/options.go
```diff
@@ -47,6 +54,9 @@ type ServerOption struct {
 	CaCertData           []byte
 	EnableLeaderElection bool
```
Monokaix, line 55: Leader election related parameters can be placed in an independent flagset, please refer to kubernets to make it more graceful.

CharlesQQ: @Monokaix updated!  pls  check again.

cmd/scheduler/app/options/options.go
```diff
@@ -195,10 +192,10 @@ func (s *ServerOption) ParseCAFiles(decryptFunc DecryptFunc) error {
 	return nil
 }
 
-// Default new and registry a default one
```
Monokaix, line 198: Why is this commented?

pkg/util/util.go
```diff
@@ -32,3 +47,31 @@ func GenerateSchedulerName(schedulerNames []string) string {
 
 	return defaultSchedulerName
 }
+
+// BindLeaderElectionFlags binds the LeaderElectionConfiguration struct fields to a flagset
+func BindLeaderElectionFlags(l *config.LeaderElectionConfiguration, fs *pflag.FlagSet) {
```
Monokaix, line 52: How about import `k8s.io/component-base/config/options/leaderelectionconfig.go` directly?

CharlesQQ: > How about import k8s.io/component-base/config/options/leaderelectionconfig.go directly?
en.., I want set default value in function `BindLeaderElectionFlags`; if not, must add a function to set LeaderElectionConfiguration  default values.

cmd/controller-manager/app/options/options.go
```diff
@@ -38,16 +38,16 @@ const (
...
 // ServerOption is the main context object for the controllers.
 type ServerOption struct {
-	KubeClientOptions    kube.ClientOptions
-	CertFile             string
-	KeyFile              string
-	CaCertFile           string
-	CertData             []byte
-	KeyData              []byte
-	CaCertData           []byte
-	EnableLeaderElection bool
```
lowang-bh, line 48: How about keeping the old one for better compatibility?

cmd/scheduler/app/options/options.go
CharlesQQ: > Why is this commented?
useless code, I will delete that.

cmd/scheduler/app/options/options.go
```diff
@@ -195,10 +192,10 @@ func (s *ServerOption) ParseCAFiles(decryptFunc DecryptFunc) error {
...
-func Default() *ServerOption {
-	s := NewServerOption()
-	s.AddFlags(pflag.CommandLine)
-	s.RegisterOptions()
-	return s
-}
+//// Default new and registry a default one
+//func Default() *ServerOption {
```
lowang-bh, line 196: Please keep it. It is used in other test packages.

pkg/util/util.go
Monokaix: But actually it's a copy from k8s: )

cmd/controller-manager/app/options/options.go
```diff
@@ -18,9 +18,9 @@ package options
 
 import (
 	"fmt"
-	"os"
-
 	"github.com/spf13/pflag"
+	"k8s.io/component-base/config"
+	"os"
```
Monokaix, line 23: Plz sort imports.

cmd/controller-manager/app/options/options.go
Monokaix: `leader-elect` flag is kept when using BindLeaderElectionFlags.

CharlesQQ: > How about keeping the old one for better compatibility?
Use leaderElection flagSet instead, it's a leaderElectionOption field.

cmd/scheduler/app/options/options.go
CharlesQQ: > Please keep it. It is used in other test packages.
all right!

cmd/controller-manager/app/options/options.go
CharlesQQ: > Plz sort imports.
done!

cmd/controller-manager/app/options/options.go
Monokaix: But `lock-object-namespace` flag seems has been removed.

CharlesQQ: > But lock-object-namespace flag seems has been removed.

use the `leader-elect-resource-namespace`  flag instead

Monokaix: So plz also update params in `helm` : )

pkg/util/util.go
```diff
@@ -1,7 +1,21 @@
 package util
 
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/component-base/config"
+	"time"
```
Monokaix, line 7: sort imports.

cmd/scheduler/app/options/options.go
```diff
@@ -147,7 +144,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 
 // CheckOptionOrDie check lock-object-namespace when LeaderElection is enabled.
 func (s *ServerOption) CheckOptionOrDie() error {
-	if s.EnableLeaderElection && s.LockObjectNamespace == "" {
+	if s.LeaderElection.LeaderElect && s.LeaderElection.ResourceLock == "" {
 		return fmt.Errorf("lock-object-namespace must not be nil when LeaderElection is enabled")
```
Monokaix, line 148: Change the description too.

cmd/controller-manager/app/options/options.go
Monokaix: I think we should better keep the old flag and mark it deprecated first.

cmd/controller-manager/app/options/options.go
```diff
@@ -102,7 +100,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 
 // CheckOptionOrDie checks the LockObjectNamespace.
 func (s *ServerOption) CheckOptionOrDie() error {
-	if s.EnableLeaderElection && s.LockObjectNamespace == "" {
+	if s.LeaderElection.LeaderElect && s.LeaderElection.ResourceLock == "" {
```
Monokaix, line 103: Here we can call `ValidateLeaderElectionConfiguration` to check all params.

cmd/controller-manager/app/options/options.go
CharlesQQ: > So plz also update params in helm  I think we should better keep the old flag and mark it deprecated first.

sure!

cmd/controller-manager/app/options/options.go
CharlesQQ: All that is updated!

cmd/controller-manager/main.go
```diff
@@ -46,8 +46,12 @@ func main() {
 	runtime.GOMAXPROCS(runtime.NumCPU())
 	klog.InitFlags(nil)
 
+	fs := pflag.CommandLine
 	s := options.NewServerOption()
-	s.AddFlags(pflag.CommandLine)
+
+	s.AddFlags(fs)
+	commonutil.LeaderElectionDefault(&s.LeaderElection)
+	componentbaseoptions.BindLeaderElectionFlags(&s.LeaderElection, fs)
```
Monokaix, line 55: Just mark LockObjectNamespace deprecated is not enough, we should set `ResourceNamespace` value to LockObjectNamespace  when it's not empty: )

CharlesQQ: > Just mark LockObjectNamespace deprecated is not enough, we should set ResourceNamespace value to LockObjectNamespace when it's not empty: )

All right！

cmd/controller-manager/app/options/options.go
lowang-bh: > I think we should better keep the old flag and mark it deprecated first.

I also means keeping the old one and marking it is deprecated in recent release.

cmd/controller-manager/app/options/options.go
```diff
@@ -84,9 +89,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 		"File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated "+
 		"after server cert).")
 	fs.StringVar(&s.KeyFile, "tls-private-key-file", s.KeyFile, "File containing the default x509 private key matching --tls-cert-file.")
-	fs.BoolVar(&s.EnableLeaderElection, "leader-elect", true, "Start a leader election client and gain leadership before "+
```
lowang-bh, line 87: why this is removed?

cmd/scheduler/app/options/options.go
```diff
@@ -107,16 +112,13 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 		"File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated "+
 		"after server cert).")
 	fs.StringVar(&s.KeyFile, "tls-private-key-file", s.KeyFile, "File containing the default x509 private key matching --tls-cert-file.")
+	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default, will deprecated, please use --leader-elect-resource-namespace instead.")
 	// volcano scheduler will ignore pods with scheduler names other than specified with the option
 	fs.StringArrayVar(&s.SchedulerNames, "scheduler-name", []string{defaultSchedulerName}, "vc-scheduler will handle pods whose .spec.SchedulerName is same as scheduler-name")
 	fs.StringVar(&s.SchedulerConf, "scheduler-conf", "", "The absolute path of scheduler configuration file")
 	fs.DurationVar(&s.SchedulePeriod, "schedule-period", defaultSchedulerPeriod, "The period between each scheduling cycle")
 	fs.StringVar(&s.DefaultQueue, "default-queue", defaultQueue, "The default queue name of the job")
-	fs.BoolVar(&s.EnableLeaderElection, "leader-elect", true,
```
lowang-bh, line 115: why this is removed, how about deprecated？

Monokaix: New flagset contains this flag, so it's not removed autually.

william-wang (approved):
/approve

Issue #2810, opened 2023-04-23
Part of the ut shows that the operation is successful, but the test case is not actually executed


wangyang0616:
**What happened**:
In [#2592](https://github.com/volcano-sh/volcano/pull/2592), modify the construction method of some UT use cases cache.schedulerCache, construct config through `kube.BuildConfig` and generate the corresponding SchedulerCache, there are the following two problems:
1. When the UT use case is running, the environment does not have k8s related information and configuration, and it is not particularly reasonable to construct config through kube.BuildConfig
2. After kube.BuildConfig returns error, return directly without adding t.Errorf() information. The test framework thinks that ut runs successfully, but it does not actually run

**The test cases involved are:**

1. volcano.sh/volcano/pkg/scheduler/actions/allocate
2. volcano.sh/volcano/pkg/scheduler/actions/preempt
3. volcano.sh/volcano/pkg/scheduler/actions/reclaim
4. volcano.sh/volcano/pkg/scheduler/actions/shuffle
5. volcano.sh/volcano/pkg/scheduler/plugins/binpack
6. volcano.sh/volcano/pkg/scheduler/plugins/drf
7. volcano.sh/volcano/pkg/scheduler/plugins/predicates
8. volcano.sh/volcano/pkg/scheduler/plugins/proportion
9. volcano.sh/volcano/pkg/scheduler/plugins/tdm

similar to the following：
https://github.com/volcano-sh/volcano/blob/d4458ffe81207749f04727873d37544d7138152d/pkg/scheduler/plugins/proportion/proportion_test.go#L205-L217
**What you expected to happen**:
The test case can run normally, the line coverage is not 0.

**How to reproduce it (as minimally and precisely as possible)**:
There are two ways to reproduce:
1. Add log `t.Errorf("kube build config error, err: %v", err)` where `config, err := kube.BuildConfig(option.KubeClientOptions)` reports an error return, execute `make unit-test` can find use cases that fail to run.
2. Execute `go test -v --covermode=count -coverprofile=count.out` in each ut test file directory to find ut use case files with 0 test coverage

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

wangyang0616:
I will fix it in the following steps:

[...]

PR #2811, opened 2023-04-23
fix: UT works fine and covers

wangyang0616:
resolve: [#2810](https://github.com/volcano-sh/volcano/issues/2810)

Solve the problem that ut can run in the first step, temporarily skip the use cases that fail to run, and then gradually submit pr for repair.


jiangkaihua:
/LGTM Do we need to merge 5 commits into 1?

wangyang0616:
> /LGTM Do we need to merge 5 commits into 1?

Thank you for your suggestion, but I think it may be better to keep the information of each commit, because each commit has independent meaning.

william-wang (approved):
/lgtm

PR #2815, opened 2023-04-26
Pods are preferentially scheduled to machines that meet the current session resources

wangyang0616:
fix: [#2782](https://github.com/volcano-sh/volcano/issues/2782)

Except that the binpack plugin has this problem, I understand that other algorithm plugins may encounter similar problems, such as task-topology, nodeorder, etc.

I was wondering if this generic problem could be solved by:
When allocate scores nodes, it divides nodes into two groups. One group is machines whose idle resources meet task resource requests, and the second group is futrue idle machines that meet task resource demands.

First, score the first group of machines, and if a suitable machine can be found, schedule the task to a suitable node; if the first group does not have a machine that meets the resource request, then score the second group of machines, and then select a suitable node for scheduling.

In this way, the pod can be dispatched to the machine that meets the resource requirements in the current session first, so that the pod will not be pending for a long time. If all the machines in the current session do not meet the requirements, it can also be scheduled to wait in the machine that meets the futrue idle.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -195,50 +195,76 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 				break
 			}
 
-			var candidateNodes []*api.NodeInfo
+			// Candidate nodes are divided into two gradients:
+			// - the first gradient node: a list of free nodes that satisfy the task resource request;
+			// - The second gradient node: the node list whose sum of node idle resources and future idle meets the task resource request;
+			// Score the first gradient node first. If the first gradient node meets the requirements, ignore the second gradient node list,
+			// otherwise, score the second gradient node and select the appropriate node.
+			var candidateNodes [][]*api.NodeInfo
+			var idleCandidateNodes []*api.NodeInfo
+			var futureIdleCandidateNodes []*api.NodeInfo
 			for _, n := range predicateNodes {
-				if task.InitResreq.LessEqual(n.Idle, api.Zero) || task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
-					candidateNodes = append(candidateNodes, n)
+				if task.InitResreq.LessEqual(n.Idle, api.Zero) {
+					idleCandidateNodes = append(idleCandidateNodes, n)
+				} else if task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
+					futureIdleCandidateNodes = append(futureIdleCandidateNodes, n)
+				} else {
+					klog.V(5).Infof("Predicate filtered node %v, idle: %v and future idle: %v do not meet the requirements of task: %v",
+						n.Name, n.Idle, n.FutureIdle(), task.Name)
 				}
 			}
+			candidateNodes = append(candidateNodes, idleCandidateNodes)
+			candidateNodes = append(candidateNodes, futureIdleCandidateNodes)
 
-			var node *api.NodeInfo
-			switch {
-			case len(candidateNodes) == 0: // If not candidate nodes for this task, skip it.
-				continue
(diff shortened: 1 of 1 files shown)


wangyang0616:
Close the current pr, and fix the preemption problem through [#2916](https://github.com/volcano-sh/volcano/pull/2916)

/close

wangyang0616:
> Close the current pr, and fix the preemption problem through [#2916](https://github.com/volcano-sh/volcano/pull/2916)
> 
> /close

The current pr was closed by mistake, reopen it.
/reopen

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -195,28 +195,64 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
-				node = ssn.BestNodeFn(task, nodeScores)
-				if node == nil {
-					node = util.SelectBestNode(nodeScores)
+				// If a proper node is found in idleCandidateNodes, skip futureIdleCandidateNodes and directly return the node information.
+				if bestNode != nil {
+					break
 				}
 			}
 
+			var node *api.NodeInfo
```
william-wang, line 248: no need to define a new `node` variable.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -195,28 +195,64 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
+			if bestNode != nil {
+				node = bestNode
+			} else {
+				klog.Errorf("task %s/%s allocate failed, bestNode is nil", task.Namespace, task.Name)
```
william-wang, line 252: The codes between 249 - 255 are redundent.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -195,28 +195,64 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
+			var bestNode *api.NodeInfo
+			for index, nodes := range candidateNodes {
+				if klog.V(5).Enabled() {
+					for _, node := range nodes {
+						klog.V(5).Infof("node %v, idle: %v, future idle: %v", node.Name, node.Idle, node.FutureIdle())
+					}
+				}
+				switch {
+				case len(nodes) == 0:
+					klog.V(3).Infof("Task: %v, no matching node is found in the nodes list."+
```
william-wang, line 229: Here log level 3 is not so good. There will be too much log.

pkg/scheduler/actions/allocate/allocate.go
wangyang0616: The code after line 250 uses node information for resource judgment, and it is necessary to ensure that the node pointer is not empty.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -195,28 +195,64 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 				break
 			}
 
-			var candidateNodes []*api.NodeInfo
+			// When scheduling pods, gradient scoring is performed on all nodes that are successfully filtered.
```
william-wang, line 198: // Candidate nodes are divided into two gradients:
// - the first gradient node: a list of free nodes that satisfy the task resource request;
// - The second gradient node: the node list whose sum of node idle resources and future idle meets the task resource request;
// Score the first gradient node first. If the first gradient node meets the requirements, ignore the second gradient node list, otherwise, score the second echelon node and select the appropriate node.

pkg/scheduler/actions/allocate/allocate.go
wangyang0616: done

pkg/scheduler/actions/allocate/allocate.go
wangyang0616: done

william-wang (approved):
/lgtm

Issue #2738, opened 2023-03-15
Scheduler crashes if node.Others[api.GPUSharingDevice] is nil 


igormish:
In pkg/scheduler/plugins/predicates/predicates.go
[link](https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/predicates/predicates.go#L398) 

**What happened**:
fit, err = node.Others[api.GPUSharingDevice].(api.Devices).FilterNode(task.Pod)
If node.Others[api.GPUSharingDevice] is nil it causes volcano to crash as nil is trying run FilterNode(task.Pod)

**What you expected to happen**:
A check for if node.Others[api.GPUSharingDevice] is nil should be done before FilterNode(task.Pod) is executed

**Environment**:
- Volcano Version: master 
- Kubernetes version:  v1.22.17

igormish:
For example 
```
d := node.Others[api.GPUSharingDevice].(api.Devices)
		if d != nil {
			return fmt.Errorf("predicates failed because node %s don't have label %s to enable device sharing", node.Name, api.GPUSharingDevice)
		}
```

PR #2818, opened 2023-05-05
Split the predicate function to solve the resource filtering problem encountered during preemption

wangyang0616:
**Background:**
allocate, preempt, and reclaim all call the predicate to filter nodes. If the predicate has a filter condition for setting whether idle resources are satisfied, then preempt and reclaim will not be able to perform the preemption action.

**Solution:**
Split the resource-related filter conditions in the predicate into predicateResource for independent judgment. The allocate phase calls predicate and predicateResource to filter nodes, and preempt and reclaim call predicate to filter nodes.

**Notice:**
Subsequent custom plug-ins also need to comply with the above principles, and the resource filter conditions are uniformly placed in the predicateResource node for processing, otherwise the preemption function may be affected

associate: [#2739](https://github.com/volcano-sh/volcano/issues/2739)

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -98,11 +98,16 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	allNodes := ssn.NodeList
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) error {
 		// Check for Resource Predicate
-		if ok, reason := task.InitResreq.LessEqualWithReason(node.FutureIdle(), api.Zero); !ok {
-			return api.NewFitError(task, node, reason)
+		if err := ssn.PredicateResourceFn(task, node); err != nil {
+			return err
 		}
 
-		return ssn.PredicateFn(task, node)
+		// Check for predicate
+		if err := ssn.PredicateFn(task, node); err != nil {
+			return err
+		}
+
+		return nil
 	}
 
 	// To pick <namespace, queue> tuple for job, we choose to pick namespace firstly.
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -393,16 +393,44 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 		return nil
 	})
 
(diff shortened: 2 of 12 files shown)


pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -313,18 +313,44 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
 		}
 
 		if node.Allocatable.MaxTaskNum <= len(nodeInfo.Pods) {
-			klog.V(4).Infof("NodePodNumber predicates Task <%s/%s> on Node <%s> failed",
+			klog.V(4).Infof("NodePodNumber resource predicates Task <%s/%s> on Node <%s> failed",
 				task.Namespace, task.Name, node.Name)
 			return api.NewFitError(task, node, api.NodePodNumberExceeded)
 		}
 
+		if gpuDevice, ok := node.Others[api.GPUSharingDevice].(api.Devices); ok {
```
wangyang0616, line 334: Whether a type assertion is required here？

igormishsky: We have [this issue](https://github.com/volcano-sh/volcano/issues/2738) to address it

wangyang0616: Yes, assertion judgment is required here, the current PR code has already been processed, I should comment on the wrong PR.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -98,11 +98,16 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	allNodes := ssn.NodeList
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) error {
 		// Check for Resource Predicate
-		if ok, reason := task.InitResreq.LessEqualWithReason(node.FutureIdle(), api.Zero); !ok {
-			return api.NewFitError(task, node, reason)
+		if err := ssn.PredicateResourceFn(task, node); err != nil {
```
william-wang, line 101: PredicateResource is just a resource comparation and do we have to make it as a session API. 
It seems there is no need to implement the function with different ways in different plugins.


make it as a session API.

wangyang0616:
The problem has been dealt with in [Predicate adapts allocate and preempt #2916](https://github.com/volcano-sh/volcano/pull/2916), the current pr display is pending.

/hold

Issue #2821, opened 2023-05-09
Wait for all cache ready when start scheduler


KunWuLuan:
**What happened**:

I noticed volcano will start scheduling after pc.cache.WaitForCacheSync(stopCh)
https://github.com/volcano-sh/volcano/blob/86ff94129a1ea47cdd3389e1f58dc6a79948f55f/pkg/scheduler/scheduler.go#L87-L99

In some cases, if we consume too much time in some events, or if there are too many events to handle, cache may not ready when we start scheduling. If some Running pods are not copied to snapshot, scheduling error may happen.

**What you expected to happen**:

Wait for all events to be handled.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
Intresting in this enhancement. I have ever thought about this feature.

KunWuLuan:
pr is being reviewed

hwdef:
still need

PR #2822, opened 2023-05-10
wait for all items handled before start to scheduling

KunWuLuan:
to #2821 

I just modify the implementation in schedulerCache. 
Maybe we can design a more nice implementation if needed.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -48,6 +48,7 @@ import (
 	"k8s.io/klog/v2"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
+	"k8s.io/utils/pointer"
 
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	"volcano.sh/apis/pkg/apis/scheduling"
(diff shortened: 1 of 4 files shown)


lowang-bh:
Thanks for your contributions! Could you plz add a switch to control this feature? Maybe some user need to keep origin while others need this feature. There is also a switch configuration on k8s's latest version with the similar feature.

KunWuLuan:
@lowang-bh Thanks for your time. I agree to add a switch for users.

pkg/scheduler/cache/cache.go
```diff
@@ -713,6 +754,9 @@ func (sc *SchedulerCache) Run(stopCh <-chan struct{}) {
 func (sc *SchedulerCache) WaitForCacheSync(stopCh <-chan struct{}) {
 	sc.informerFactory.WaitForCacheSync(stopCh)
 	sc.vcInformerFactory.WaitForCacheSync(stopCh)
+	if sc.waitForHandlesSynced {
```
lowang-bh, line 757: if sc.waitForHandlesSynced ==false, when will those handlers run sync?

lowang-bh:
We'd better include this feature in new release. @Monokaix @william-wang

hwdef:
Please rebase the code.

PR #2827, opened 2023-05-13
fix: nodeinfo clone

### LINKED ISSUE #2826: The volcano scheduler panic in the predicate phase
**What happened**:
The Volcano scheduler panics during scheduling. 
The error message is as follows:
`fatal error: concurrent map read and map write`

The panic location is as follows:
https://github.com/volcano-sh/volcano/blob/0d3fd2b3a22a32b0fee75e6d816f3545f3426499/pkg/scheduler/plugins/predicates/predicates.go#L409

There should be no panic problem。

It should be that the `node.Others` map is read and written at the same time, which causes panic.

After analyzing the code, the problem should be in the following places.
https://github.com/volcano-sh/volcano/blob/0d3fd2b3a22a32b0fee75e6d816f3545f3426499/pkg/scheduler/api/node_info.go#L215

In the snapshot phase, a copy of node info in shecdulercache is cloned, including the node.Others data structure, which is used for scheduling the current session.

However, only values are assigned to the clone of the map, and the overall structure is not copied. Therefore, the nodeinfo used in snapshot and schedulercache is the same data.

When a node is updated, the schedulercache writes data to node.Others. When the predicate accesses the node, the map is read and written at the same time, and panic occurs.

**Environment**:
- Volcano Version: master
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
fix #2826

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -212,7 +212,8 @@ func (ni *NodeInfo) Clone() *NodeInfo {
 
 	klog.V(5).Infof("imageStates is %v", res.ImageStates)
 
-	res.Others = ni.Others
+	res.Others = ni.CloneOthers()
+	res.ImageStates = ni.CloneImageSumary()
 	return res
 }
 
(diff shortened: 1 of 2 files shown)


wangyang0616:
It's ok to me. 
Please help to review. Thanks! @jiangkaihua @william-wang

jiangkaihua (commented):
/LGTM

william-wang (approved):
/lgtm
/approve

PR #2840, opened 2023-05-19
disable the enableJobEnqueued default in scheduler-configmap

### LINKED ISSUE #2814: PodGroup Pending forever because previous unschedulable podgroup occupied resources
## What happened
PodGroup is pending but accually there is enough idle resources.

## How to reproduce it (as minimally and precisely as possible)
- Cluster: 2 nodes with 1 GPU each.
- Firstly, submit a Job with one task using 2 GPUs. Of course, PodGroup of this Job is Inqueue, but Pod is pending.
- Secondly, submit a job using 1 GPU.
    - PodGroup is pending, pod will not be created. (PodGroup should be Inqueue, that's what I expected)
    - Obviously, 2 GPUs are idle, not occupied by any Pod. The second Job should be Running.

## Anything else we need to know
Before volcano 1.5.0, the `proportion` plugin always enqueue the job If capability of queue is not set.

```go
	ssn.AddJobEnqueueableFn(pp.Name(), func(obj interface{}) int {
		job := obj.(*api.JobInfo)
		queueID := job.Queue
		attr := pp.queueOpts[queueID]
		queue := ssn.Queues[queueID]
		if len(queue.Queue.Spec.Capability) == 0 {
			klog.V(4).Infof("Capability of queue <%s> was not set, allow job <%s/%s> to Inqueue.",
				queue.Name, job.Namespace, job.Name)
			return util.Permit
		}

			klog.V(4).Infof("job %s MinResources is null.", job.Name)
			return util.Permit
		}
		minReq := job.GetMinResources()

		klog.V(5).Infof("job %s min resource <%s>", job.Name, minReq.String())
		klog.V(5).Infof("queue %s capability <%s>", queue.Name, api.NewResource(queue.Queue.Spec.Capability).String())
		klog.V(5).Infof("queue %s allocated <%s>", queue.Name, attr.allocated.String())
		inqueue := minReq.Add(attr.allocated).Add(attr.inqueue).LessEqual(api.NewResource(queue.Queue.Spec.Capability), api.Infinity)
		if inqueue {
			attr.inqueue.Add(job.GetMinResources())
			return util.Permit
		}
		return util.Reject
	})
```

However, since then, the condition to judge if capability of queue is not set has changed.
```go
	ssn.AddJobEnqueueableFn(pp.Name(), func(obj interface{}) int {
		job := obj.(*api.JobInfo)
		queueID := job.Queue
		attr := pp.queueOpts[queueID]
		queue := ssn.Queues[queueID]
		if attr.realCapability == nil {
			klog.V(4).Infof("Capability of queue <%s> was not set, allow job <%s/%s> to Inqueue.",
				queue.Name, job.Namespace, job.Name)
			return util.Permit
		}

			klog.V(4).Infof("job %s MinResources is null.", job.Name)
			return util.Permit
		}
		minReq := job.GetMinResources()

		klog.V(5).Infof("job %s min resource <%s>, queue %s capability <%s> allocated <%s>",
			job.Name, minReq.String(), queue.Name, attr.realCapability.String(), attr.allocated.String())
		inqueue := minReq.Add(attr.allocated).Add(attr.inqueue).LessEqual(attr.realCapability, api.Infinity)
		klog.V(5).Infof("job %s inqueue %v", job.Name, inqueue)
		if inqueue {
			attr.inqueue.Add(job.GetMinResources())
			return util.Permit
		}
		return util.Reject
	})
```

## Proposal
Revert `if attr.realCapability == nil` to `if len(queue.Queue.Spec.Capability) == 0`

## Environment
- Volcano Version: 1.7.0 (It could be reproduced since 1.5.0)
- Kubernetes version (use `kubectl version`): 1.19 (It dosen't matter, I think)

xiao-jay:
fix #2814 
issue conclusion:
The introduction of guarantee is an allocation scenario related to elastic queues. For most user scenarios, it is not a rigidly needed function. We can add the following configuration to the default configuration of the trunk. The guarantee capability is disabled by default. If users need it, they can manually enable it.

found a useless verb `updateStatus`,delete it .

### DIFF
--- a/installer/helm/chart/volcano/config/volcano-scheduler.conf
+++ b/installer/helm/chart/volcano/config/volcano-scheduler.conf
@@ -11,5 +11,6 @@ tiers:
     enablePreemptable: false
   - name: predicates
   - name: proportion
+    enableJobEnqueued: false
   - name: nodeorder
   - name: binpack
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -8646,6 +8646,7 @@ data:
         enablePreemptable: false
       - name: predicates
       - name: proportion
+        enableJobEnqueued: false
       - name: nodeorder
       - name: binpack
 ---
@@ -8669,7 +8670,7 @@ rules:
     verbs: ["create", "list", "watch", "update", "patch"]
   - apiGroups: [""]
     resources: ["pods", "pods/status"]
-    verbs: ["create", "get", "list", "watch", "update", "patch", "bind", "updateStatus", "delete"]
+    verbs: ["create", "get", "list", "watch", "update", "patch", "bind", "delete"]
   - apiGroups: [""]
     resources: ["pods/binding"]
     verbs: ["create"]


hwdef (commented):
If it needs to be turned off most of the time, then in the code, its default value should be turned off. Instead of turning it off via the config file

hwdef (commented):
If it needs to be turned off most of the time, then in the code, its default value should be turned off. Instead of turning it off via the config file

wangyang0616:
> > If it needs to be turned off most of the time, then in the code, its default value should be turned off. Instead of turning it off via the config file
> 
> https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/defaults.go#L22
> 
> ```
> func ApplyPluginConfDefaults(option *conf.PluginOption) {
> 	t := true
> 
> 	if option.EnabledJobOrder == nil {
> 		option.EnabledJobOrder = &t
> 	}
> ```
> 
> default all PluginConf is true,I don.t'know why all config is true , @wangyang0616 Do you know why config defaults to true? So on the basis that the default is true, is it not good to set enableJobEnqueued to false by default alone? In that case, is it more appropriate to change the default configuration? @hwdef

In my opinion, this should be considered for the convenience of management and the simplicity of user use. Otherwise, some configurations default to true, and some are false, which needs to be recorded in detail. Manual (of course, almost all configurations currently default to true to meet the requirements. If more and more configurations need to default to false in the future, this will be a better choice)

xiao-jay:
@wangyang0616 
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/defaults.go#L34
if set 
```
	if option.EnabledJobEnqueued == nil {
		option.EnabledJobEnqueued = &false
	}
```
Does it mean that all plugins default enableJobEnqueued to false, is this default setting appropriate?

lowang-bh:
`attr.realCapability == nil ` looks will nevel be true?

PR #2842, opened 2023-05-22
vc-scheduler is only  responsible for pod when filter pod deletion events

LivingCcj:
Watching  pod events will timeout when the etcd service meets performance problem，the pod informer will re-list the newest pod metadata list。Meanwhile the vc-schduler will loss some delete pod events，it leads to some deleted pod still remain in cache, Now  This problem has been well solved by [2784](https://github.com/volcano-sh/volcano/pull/2784).
Here the vc-schduler  should  consider  to be responsible for pod when filter pod deletion events.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,7 +213,7 @@ mod-download-go:
 
 .PHONY: mirror-licenses
 mirror-licenses: mod-download-go; \
-	go install istio.io/tools/cmd/license-lint@latest; \
+	go install istio.io/tools/cmd/license-lint@1.19.7; \
 	cd licenses; \
 	rm -rf `ls ./ | grep -v LICENSE`; \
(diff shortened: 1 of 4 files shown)


xiao-jay:
@LivingCcj you should add -s in git commit ,such as `git commit -s -m "xxxx"` to pass the DCO ci error.

LivingCcj:
> @LivingCcj you should add -s in git commit ,such as `git commit -s -m "xxxx"` to pass the DCO ci error.
Signed-off-by has been set.

wangyang0616:
If it is the first time to submit a PR in the community, a maintainer is needed to help trigger CI. 
Please help trigger CI, Thanks! @william-wang

Issue #2843, opened 2023-05-22
fix synchronous pvc/pv bind error when volcano schedule pod with pvc


gj409237405:
**What happened**:

There is a problem between Volcano and kube-scheduler in scheduling pod contains pvc 。And the problem is kube-scheduler check the pvc bounding asynchronously, but Volcano check the pvc bounding synchronous 。 This will lead to Volcano will block when facing pod have wrong pvc， and affect subsequent scheduled pod by volcano

here is error log

`E0522 19:32:05.105232       7 cache.go:1022] task default/p1-use-fb6db5888-h7pw8 bind Volumes failed: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"Operation cannot be fulfilled on persistentvolumeclaims \"p1\": the object has been modified; please apply your changes to the latest version and try again", Reason:"Conflict", Details:(*v1.StatusDetails)(0xc000426840), Code:409}}
I0522 19:32:05.105311       7 cache.go:305] Revert assumed volumes for task default/p1-use-fb6db5888-h7pw8 on node 192.168.0.29
E0522 19:33:05.207416       7 cache.go:1022] task default/p1-use-fb6db5888-h7pw8 bind Volumes failed: &fmt.wrapError{msg:"binding volumes: timed out waiting for the condition", err:(*errors.errorString)(0xc000088fd0)}
I0522 19:33:05.207494       7 cache.go:305] Revert assumed volumes for task default/p1-use-fb6db5888-h7pw8 on node 192.168.0.29`

**What you expected to happen**:

I think the failed pod is not supposed to block the whole session, but it does

**How to reproduce it (as minimally and precisely as possible)**:

1. First you should make a pvc , which its storageclass's volumeBindingMode volumeBindingMode is waitforfirstconsumer ,and you shoule make sure the pvc have some problem.
2. Second you make a pod contains the pvc .
3. Third you can deploy the pod.
4. Four you deploy other workload, and you will find that the workload will be scheduled ,but will alway delay a few time ,like 30 s, or 60s, or 120s this depends on how many pod have wrong pvc.

**Anything else we need to know?**:

**Environment**:
- Volcano Version:   old version before I  change 
- Kubernetes version (use `kubectl version`): *
[...]

PR #2844, opened 2023-05-22
 fix synchronous pvc/pv bind error when volcano schedule pod with pvc

gj199575:
this pr solve https://github.com/volcano-sh/volcano/issues/2843

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -769,23 +769,21 @@ func (sc *SchedulerCache) Evict(taskInfo *schedulingapi.TaskInfo, reason string)
 
 // Bind binds task to the target host.
 func (sc *SchedulerCache) Bind(tasks []*schedulingapi.TaskInfo) error {
-	go func(taskArray []*schedulingapi.TaskInfo) {
-		tmp := time.Now()
-		errTasks, err := sc.Binder.Bind(sc.kubeClient, taskArray)
-		if err == nil {
-			klog.V(3).Infof("bind ok, latency %v", time.Since(tmp))
-			for _, task := range tasks {
-				sc.Recorder.Eventf(task.Pod, v1.EventTypeNormal, "Scheduled", "Successfully assigned %v/%v to %v",
-					task.Namespace, task.Name, task.NodeName)
-			}
-		} else {
-			for _, task := range errTasks {
-				klog.V(2).Infof("resyncTask task %s", task.Name)
-				sc.VolumeBinder.RevertVolumes(task, task.PodVolumes)
-				sc.resyncTask(task)
-			}
+	tmp := time.Now()
+	errTasks, err := sc.Binder.Bind(sc.kubeClient, tasks)
+	if err == nil {
+		klog.V(3).Infof("bind ok, latency %v", time.Since(tmp))
+		for _, task := range tasks {
+			sc.Recorder.Eventf(task.Pod, v1.EventTypeNormal, "Scheduled", "Successfully assigned %v/%v to %v",
+				task.Namespace, task.Name, task.NodeName)
 		}
-	}(tasks)
+	} else {
+		for _, task := range errTasks {
+			klog.V(2).Infof("resyncTask task %s", task.Name)
+			sc.VolumeBinder.RevertVolumes(task, task.PodVolumes)
+			sc.resyncTask(task)
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -1016,30 +1014,32 @@ func (sc *SchedulerCache) processBindTask() {
...
 		}
-	}
 
-	bindTasks := make([]*schedulingapi.TaskInfo, len(successfulTasks))
-	copy(bindTasks, successfulTasks)
-	if err := sc.Bind(bindTasks); err != nil {
-		klog.Errorf("failed to bind task count %d: %#v", len(bindTasks), err)
-		return
-	}
+		bindTasks := make([]*schedulingapi.TaskInfo, len(successfulTasks))
```
xiao-jay, line 1030: Please delete the extra tabs

pkg/scheduler/cache/cache.go
```diff
@@ -1016,30 +1014,32 @@ func (sc *SchedulerCache) processBindTask() {
...
+		copy(bindTasks, successfulTasks)
```
xiao-jay, line 1031: Please delete the extra tabs

pkg/scheduler/cache/cache.go
```diff
@@ -768,24 +767,22 @@ func (sc *SchedulerCache) Evict(taskInfo *schedulingapi.TaskInfo, reason string)
...
-				sc.Recorder.Eventf(task.Pod, v1.EventTypeNormal, "Scheduled", "Successfully assigned %v/%v to %v",
-					task.Namespace, task.Name, task.NodeName)
-			}
-		} else {
-			for _, task := range errTasks {
-				klog.V(2).Infof("resyncTask task %s", task.Name)
-				sc.VolumeBinder.RevertVolumes(task, task.PodVolumes)
-				sc.resyncTask(task)
-			}
+func (sc *SchedulerCache) Bind(taskArray []*schedulingapi.TaskInfo) error {
```
wangyang0616, line 770: What is the reason for Volcano's original asynchronous binding, performance?
Does the modification to synchronous have any impact on performance?

pkg/scheduler/cache/cache.go
gj199575: ok, I should have  solved this problem 。

pkg/scheduler/cache/cache.go
gj199575: ok,here The reason which  change asynchronous  to synchronous    is   I have move the asynchronous   to above little , so here is not need

pkg/scheduler/cache/cache.go
elinx: > Please replace the extra tab with two spaces, which may be better，because the tab size settings vary among different editors

volcano uses tabs if you take a close look at the code🤣

pkg/scheduler/cache/cache.go
```diff
@@ -1010,35 +1008,37 @@ func (sc *SchedulerCache) processBindTask() {
...
-	successfulTasks := make([]*schedulingapi.TaskInfo, 0)
-	for _, task := range sc.bindCache {
-		if err := sc.VolumeBinder.BindVolumes(task, task.PodVolumes); err != nil {
-			klog.Errorf("task %s/%s bind Volumes failed: %#v", task.Namespace, task.Name, err)
-			sc.VolumeBinder.RevertVolumes(task, task.PodVolumes)
-			sc.resyncTask(task)
-		} else {
-			successfulTasks = append(successfulTasks, task)
-			klog.V(5).Infof("task %s/%s bind Volumes done", task.Namespace, task.Name)
+func (sc *SchedulerCache) BindTask(bindCache []*schedulingapi.TaskInfo) {
```
william-wang, line 1014: why we have to add bindcache parameter.

william-wang:
@gj199575 please enable the batch bind and have a test.

pkg/scheduler/cache/cache.go
```diff
@@ -1010,35 +1008,37 @@ func (sc *SchedulerCache) processBindTask() {
...
+	klog.V(5).Infof("batch bind task count %d", len(bindCache))
+	var tmpBindCache []*schedulingapi.TaskInfo = make([]*schedulingapi.TaskInfo, len(bindCache))
+	copy(tmpBindCache, bindCache)
+	go func(bindCacheParam []*schedulingapi.TaskInfo, sc *SchedulerCache) {
```
william-wang, line 1018: suggest change `bindCacheParam` to `tasks`

william-wang (approved):
/lgtm
/approve

gj409237405:
https://www.feimaoke.com/

hwdef:
@gj409237405 @Monokaix 
Please delete this, them looks like some spam

PR #2823, opened 2023-05-11
Volcano automatically releases the corresponding helm charts package

wangyang0616:
**Description of Requirement:**

refer: [#2766](https://github.com/volcano-sh/volcano/issues/2766),

Volcano can be installed through helm at present, but you need to download the source code of the entire warehouse, the operation is relatively troublesome, and it does not conform to the standard usage of helm.

**Target:**
Build the helm repo corresponding to Volcano, and automatically publish helm charts.

Users can view the latest released packages through `helm repo update`, and install and use them through `helm install`

**accomplish:**
1. Build the charts warehouse of Volcano, see [helm-charts](https://github.com/volcano-sh/helm-charts)
2. When the Volcano main warehouse releases the tag, the release version is pushed to the helm-charts warehouse, with the charts information attached, and the repo information is updated at the same time. 
(It is automatically completed through github action without manual intervention.)
4. When the Volcano main warehouse publishes a tag, it pushes the image corresponding to the tag to dockerhub. 
(It is automatically completed through github action without manual intervention.)

PR #2848, opened 2023-05-23
sed syntax is compatible with linux and darwin systems

wangyang0616:
The current pr is an optimization of [#2823](https://github.com/volcano-sh/volcano/pull/2823)

The processing of `sed -i` command in linux and darwin systems is inconsistent.
In the darwin system, you need to specify the backup file format after -i, for example:

- Linux system: sed -i 's|xxx|xxx|g' file
- Darwin system: sed -i “xx” ‘s|xxx|xxx|g’ file

The current pr selects the appropriate sed command parameters to operate according to the system type to ensure compatibility with both linux and darwin systems

### DIFF
--- a/hack/generate-charts.sh
+++ b/hack/generate-charts.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #
-# Copyright 2021 The Volcano Authors.
+# Copyright 2023 The Volcano Authors.
 #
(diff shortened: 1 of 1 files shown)


hwdef (commented):
/lgtm

hwdef:
this is important

PR #2850, opened 2023-05-23
Ensure delete event is handled

WulixuanS:
/kind bug

This bug is like #2784
There are two places that also handle the delete event and need to be modified


william-wang:
Overall, the fix looks good to me.

william-wang:
@WulixuanS please make the ci happy.

WulixuanS:
> @WulixuanS please make the ci happy.

Done

Issue #2833, opened 2023-05-17
volcano-admission-init error is reported when the volcano is uninstalled and reinstalled or the helm is used to upgrade the volcano


wangyang0616:
**What happened**:
After the Volcano is installed, if you uninstall and reinstall the Volcano or use the helm upgrade tool to upgrade the Volcano, the volcano-admission-init-k6kct fails. The symptoms are as follows:

```
volcano-system   volcano-admission-66775d9c56-n2hkr         1/1     Running   0          19s
volcano-system   volcano-admission-init-7pg2h               0/1     Error     0          40s
volcano-system   volcano-admission-init-k6kct               0/1     Error     0          52s
volcano-system   volcano-admission-init-qj4d7               0/1     Error     0          44s
volcano-system   volcano-admission-init-x8ck2               0/1     Error     0          48s
volcano-system   volcano-controllers-dd4b9d857-2scwk        1/1     Running   0          19s
volcano-system   volcano-scheduler-5bd68f8f8-l6knh          1/1     Running   0          19s
```
Error log information of volcano-admission-init：
```
creating certs in dir /tmp 
Certificate request self-signature ok
subject=CN = volcano-admission-service.volcano-system.svc
error: failed to create secret secrets "volcano-admission-secret" already exists
```

During Volcano installation, a secret is created using a job. When Volcano is uninstalled, the original secret is not cleared. When Volcano is installed or upgraded again, an error is reported when volcano-admission-init creates a secret with the same name.
https://github.com/volcano-sh/volcano/blob/ff3ff9eb283825d50de6c9e2a36554a149e9f2b1/installer/dockerfile/webhook-manager/gen-admission-secret.sh#L93-L100

**What you expected to happen**:
When the volcano-admission-init command is used to create a secret, if the secret with the same name already exists, the system does not create a secret and returns a success message.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master, v1.6.0, v1.7.0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hwdef:
I think this secret should also be deleted when uninstalling the volcano

xiao-jay:
> 

I agree.

xiao-jay:
[...]

PR #2855, opened 2023-05-24
Automatically clean up and install Volcano's initialization job

wangyang0616:
**Introduction：**
Types such as job will fail during helm upgrade, such as: [volcano-admission-init error is reported when the volcano is uninstalled and reinstalled or the helm is used to upgrade the volcano #2833](https://github.com/volcano-sh/volcano/issues/2833)

Add `ttlSecondsAfterFinished` to automatically clean up and initialize job resources, avoiding errors when upgrading Volcano through helm

**constraint：**
The default job cleaning time is 100s. If you upgrade within 100s after installing Volcano, an upgrade error will occur.

### DIFF
--- a/installer/helm/chart/volcano/templates/admission.yaml
+++ b/installer/helm/chart/volcano/templates/admission.yaml
@@ -154,6 +154,7 @@ metadata:
     app: volcano-admission-init
 spec:
   backoffLimit: 3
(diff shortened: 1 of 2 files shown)


hwdef (commented):
/lgtm

hwdef (commented):
/lgtm

PR #2857, opened 2023-05-24
solve make error on macbook

lowang-bh:


Makefile
```diff
@@ -29,6 +29,8 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+OS=$(shell uname -s)
```
hwdef, line 32: I think this show be `kernel`

hwdef: linux and darwin are not OS

lowang-bh: `-s  -- print name of the operating system` help shows it is os name

hwdef: -s      Write the name of the `operating system implementation` to standard output.

hwdef: forget it, it's not important

hwdef (commented):
/lgtm

william-wang (approved):
/approve

Issue #2859, opened 2023-05-25
Ban schedule to share gpu id which  usedMemory more than allMemory


WingkaiHo:
**What happened**:

When high concurrent schedule, and cluster less GPU resource( such as useage > 90%),  multi pods schedule to same gpu of node, it will leak to pod gpu memroy resource more than gpu memroy.  It wiil leak all pod which scheduler to this node use the gpu.

**What you expected to happen**:

Ban schedule to gpu which usedMemory > usedMemory

**How to reproduce it (as minimally and precisely as possible)**:
Use one k8s node, make high concurrent schedule, pod gpu resouce less gpu resouce.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: 1.6.0
- Kubernetes version (use `kubectl version`): 1.22.5 
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`): 4.14.83
- Install tools: kubespary
- Others:

wangyang0616:
Could you please share the log information of the error phenomenon?

WingkaiHo:
My device plugin use factor = 1000,  so my gpu is nvidia 1080 TI 11G.  I use nvidia-smi found task scheduler to gpu 0 of node time by time, when task usedMemory > 11G so time status of pod  GPUOutofmemory

WingkaiHo:
I think when scheduler get multi sceduler info and cluster leak of gpu resource [usage > 95%],  two task assign node, when node cache have not been updated by pod add / update event callback. It will leak two task assign to node one gpu when gpu only enough for task1 or task2.

WingkaiHo:
@wangyang0616 When you use only one gpu node, and  when scheduler get multi gpu scheduler it will easy to reproduction. 

- idx 0 gpu, memory nearly 40G, but task appoint time by time

![企业微信截图_96fd305c-5f9b-4dea-a8db-d68da3487334(1)](https://github.com/volcano-sh/volcano/assets/14209014/0872ff37-bfb8-4656-931a-cfd4e554d421)

wolcano log
[volcano.schedule.log](https://github.com/volcano-sh/volcano/files/12427806/volcano.schedule.log)

WingkaiHo:
@wangyang0616

PR #2860, opened 2023-05-25
fix: Ban schedule to share gpu when usedMemory more then allMemory

WingkaiHo:
fix https://github.com/volcano-sh/volcano/issues/2859

When high concurrent schedule the pods to same gpu id, if earlier k8s pod add event, it will lead to usedMemory more then allMemory sometime. So res[id] = allMemory - usedMemory may be a large number. Leak all pod schedule to this gpu of node


WingkaiHo:
@Thor-wl

WingkaiHo:
@k82cn @Thor-wl  any other question

WingkaiHo:
@wangyang0616

PR #2866, opened 2023-05-27
fix: events miss when there is a pod bind failed

lowang-bh:
Describe: fix #2865 that successfully binding events will lost if a pod is failed to bind; fix #2879 metrics not accurate

key points:
 1: replcace all clientset with interface
 2: move event and metrics recoding to where pod is successfully bounden
 3: add testcase about bind tasks

PR #2869, opened 2023-05-27
abstract Bind() param to kubernetes.Interface

Monokaix:
`Binder` is an interface, its param should also be, meanwhile, we can replace it with fake client in UT.


lowang-bh:
Good jobs! All clients has been replace to interface in PR #2866

william-wang (approved):
/lgtm

PR #2817, opened 2023-04-30
fix some typos

halegreen:

PR #2870, opened 2023-05-27
fix world spelling

lowang-bh:
world spelling mistake


lowang-bh:
same as #2817, close it.

Issue #2884, opened 2023-05-31
After being preempted，vcjob is running，but pods is pending


renwenlong-github:
**What happened**:
1、I created high and low priority；
2、First create a low-priority vcjob1, after the vcjob 1 runs, create a high-priority vcjob2 to preempt vcjob1;
3、vcjob2 completely preempts vcjob1, the pods of vcjob1 are pending, **but the status of vcjob1 is running**；

**What you expected to happen**:
After being preempted; vcjob1 can give a real response to the status of pods；

**How to reproduce it (as minimally and precisely as possible)**:
You need to ensure that your node is larger than **4c4g**
my config.yaml，Enable preempt actions
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: volcano-scheduler-configmap
  namespace: volcano-system
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, backfill, preempt"
    tiers:
      - plugins:
          - name: priority
          - name: gang
            enablePreemptable: false
          - name: conformance
          - name: sla
      - plugins:
          - name: overcommit
            arguments:
            overcommit-factor: 1.0
          - name: drf
            enablePreemptable: false
          - name: predicates
          - name: proportion
          - name: nodeorder
          - name: binpack
```

test-queue.yaml
```yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: test-queue
spec:
  weight: 1
  reclaimable: false
  capability:
    cpu: "4"
    memory: "4Gi"
  guarantee: 
    resource:
      cpu: "4"
      memory: "4Gi"
```

priority-class.yaml
```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-tenant-priority
value: 100
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: low-tenant-priority
value: 10
```

vcjob1.yaml
```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: vcjob1
spec:
  minAvailable: 1
  schedulerName: volcano
  maxRetry: 5
  queue: test-queue
  priorityClassName: low-tenant-priority
  tasks:
    - replicas: 5
      minAvailable: 1
      name: "master"
      template:
        metadata:
          name: master
        spec:
          containers:
            - image: nginx
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
[...]

PR #2886, opened 2023-05-31
fix pods pending but vcjob is running

xiefan-github:
[issues #2884](https://github.com/volcano-sh/volcano/issues/2884)

### DIFF
--- a/pkg/controllers/job/state/running.go
+++ b/pkg/controllers/job/state/running.go
@@ -91,7 +91,7 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 				}
 				return true
 			}
-			if status.Pending == jobReplicas {
+			if status.Pending > jobReplicas-totalTaskMinAvailable {
 				status.State.Phase = vcbatch.Pending
 				return true
 			}
--- a/pkg/controllers/job/state/running.go
+++ b/pkg/controllers/job/state/running.go
@@ -91,7 +91,7 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 				}
 				return true
 			}
-			if status.Pending > jobReplicas-totalTaskMinAvailable {
+			if status.Pending > jobReplicas-ps.job.Job.Spec.MinAvailable {
 				status.State.Phase = vcbatch.Pending
 				return true
 			}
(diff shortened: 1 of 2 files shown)


pkg/controllers/job/state/running.go
```diff
@@ -91,6 +91,10 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 				}
 				return true
 			}
+			if status.Pending == jobReplicas {
```
hwdef, line 94: should it be `status.Pending > jobReplicas - totalMinavailable`

xiefan-github: yes, you're right, it's more accurate, I'll fix it

xiefan-github: I have fixed it

hwdef (commented):
/lgtm

lowang-bh:
could you add more testcase to cover your code, pls?

xiefan-github:
> could you add more testcase to cover your code, pls?

Of course, I'd be happy to write some test cases，give me some time

xiefan-github:
> could you add more testcase to cover your code, pls?

could you review my testcase，Thank you

pkg/controllers/job/state/running.go
xiefan-github: I have a confusion why totalMinavailable  can not be equal to Job.Spec.MinAvailable? #2895

lowang-bh:
> > could you add more testcase to cover your code, pls?
> 
> could you review my testcase，Thank you

It's ok to me. @wangyang0616 @hzxuzhonghu plz also help to review. thanks.

hwdef (commented):
/lgtm

wangyang0616:
Thanks for your contribution. It's ok to me.

Please help trigger CI. Thanks!  @william-wang

pkg/controllers/job/state/running.go
hwdef: @xiefan-github 
When I first raised the question, I was thinking that if the number of pending pods is lower than minavailable, the pod should be in the pending state.

william-wang (approved):
/approve

PR #2890, opened 2023-05-31
check kind-cluster exist or not when create it

### LINKED ISSUE #2889: better to skip creating same kind-cluster when it exist
It is better to skip  creating if the kind-cluster exist
**What happened**:

```sh
Running kind: [kind create cluster --name integration --config /go/src/github.com/volcano.sh/volcano/hack/../hack/e2e-kind-config.yaml]
ERROR: failed to create cluster: node(s) already exist for a cluster with the name "integration"

Checking whether the required images exist

```

skip create same cluster with no Error

1. kind create cluster --name=integration
2. make e2e

Maybe need to clean up the exist kind-cluster, sometimes.

**Environment**:
- Volcano Version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
Fix #2889 
about `ERROR: failed to create cluster: node(s) already exist for a cluster with the name "integration"`

### DIFF
--- a/hack/lib/install.sh
+++ b/hack/lib/install.sh
@@ -19,7 +19,11 @@ function kind-up-cluster {
   check-kind
 
   echo "Running kind: [kind create cluster ${CLUSTER_CONTEXT} ${KIND_OPT}]"
-  kind create cluster ${CLUSTER_CONTEXT} ${KIND_OPT}
+  echo "Kind cluster name: ${CLUSTER_NAME}"
+  kind get clusters |grep ${CLUSTER_NAME}
+  if [[ $? -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


wangyang0616:
The automatic operation of e2e needs to ensure that the environment is the same every time, so as to avoid the difference of existing cluster information, resulting in unstable operation of e2e.

So I don't think the current modification is a good way

wangyang0616:
> As your describe, how about delete exist one and recreate it?

After e2e runs, it will automatically clean up the cluster, regardless of success or failure, unless manually interrupted.

In what scenario did you encounter a cluster with the same name?

Issue #2897, opened 2023-06-06
preempt actions causes controller deadlock


renwenlong-github:
**What happened**:
create dependson job:
```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: minavailable-job
spec:
  schedulerName: volcano
  minAvailable: 3
  tasks:
    - replicas: 3
      minAvailable: 3
      name: "master"
      template:
        metadata:
          name: master
        spec:
          containers:
            - image: nginx
              name: nginx
              resources:
                requests:
                  cpu: "0.5"
                  memory: "0.5Gi"
          restartPolicy: OnFailure
    - replicas: 3
      minAvailable: 3
      name: "work"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: nginx
              name: nginx
              resources:
                requests:
                  cpu: "1"
                  memory: "1Gi"
          restartPolicy: OnFailure
      dependsOn:
        name:
          - "master"
```

wait all pods running
```yaml
(base)  ~/ kubectl get pods
NAME                        READY   STATUS    RESTARTS   AGE
minavailable-job-master-0   1/1     Running   0          17s
minavailable-job-master-1   1/1     Running   0          17s
minavailable-job-master-2   1/1     Running   0          17s
minavailable-job-work-0     1/1     Running   0          8s
minavailable-job-work-1     1/1     Running   0          8s
minavailable-job-work-2     0/1     Pending   0          8s
```
Use delete pods to simulate preempt
```yaml
(base)  ~/ kubectl get pods |grep mina |awk '{print $1}' |xargs kubectl delete pods
pod "minavailable-job-master-0" deleted
pod "minavailable-job-master-1" deleted
pod "minavailable-job-master-2" deleted
pod "minavailable-job-work-0" deleted
pod "minavailable-job-work-1" deleted
pod "minavailable-job-work-2" deleted

(base)  ~/ kubectl get pods |grep mina |awk '{print $1}' |xargs kubectl delete pods
pod "minavailable-job-work-2" deleted

# Try multiple times, the recurrence is random
kubectl get pods |grep mina |awk '{print $1}' |xargs kubectl delete pods
kubectl get pods |grep mina |awk '{print $1}' |xargs kubectl delete pods

(base)  ~/ kubectl get pods
No resources found in default namespace.

```

controllers logs
```yaml
...
[...]

PR #2898, opened 2023-06-06
fix controller deadlook when wait dependson task

renwenlong-github:
issues https://github.com/volcano-sh/volcano/issues/2897

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -378,7 +378,7 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 			taskIndex := jobhelpers.GetTasklndexUnderJob(taskName, job)
 			if job.Spec.Tasks[taskIndex].DependsOn != nil {
 				if !cc.waitDependsOnTaskMeetCondition(taskName, taskIndex, podToCreateEachTask, job) {
-					klog.V(4).Infof("Job %s/%s depends on task not ready", job.Name, job.Namespace)
+					klog.V(3).Infof("Job %s/%s depends on task not ready", job.Name, job.Namespace)
 					// release wait group
 					for _, pod := range podToCreateEachTask {
 						go func(pod *v1.Pod) {
@@ -491,22 +491,22 @@ func (cc *jobcontroller) waitDependsOnTaskMeetCondition(taskName string, taskInd
 	}
 	dependsOn := *job.Spec.Tasks[taskIndex].DependsOn
 	if len(dependsOn.Name) > 1 && dependsOn.Iteration == batch.IterationAny {
-		// any ready to create task
+		// any ready to create task, return true
 		for _, task := range dependsOn.Name {
 			if cc.isDependsOnPodsReady(task, job) {
 				return true
 			}
 		}
+		// all not ready to skip create task, return false
 		return false
-	} else {
-		for _, dependsOnTask := range dependsOn.Name {
-			// any not ready to skip
-			if !cc.isDependsOnPodsReady(dependsOnTask, job) {
-				return false
-			}
-			return true
+	}
+	for _, dependsOnTask := range dependsOn.Name {
+		// any not ready to skip create task, return false
+		if !cc.isDependsOnPodsReady(dependsOnTask, job) {
+			return false
 		}
 	}
+	// all ready to create task, return true
 	return true
 }
 
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -18,7 +18,6 @@ package job
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -379,8 +378,7 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 			taskIndex := jobhelpers.GetTasklndexUnderJob(taskName, job)
 			if job.Spec.Tasks[taskIndex].DependsOn != nil {
 				if !cc.waitDependsOnTaskMeetCondition(taskName, taskIndex, podToCreateEachTask, job) {
-					klog.Errorf("Job %s/%s depends on task not ready", job.Name, job.Namespace)
-					creationErrs = append(creationErrs, errors.New(fmt.Sprintf("Job %s/%s depends on task not ready", job.Name, job.Namespace)))
+					klog.V(4).Infof("Job %s/%s depends on task not ready", job.Name, job.Namespace)
 					// release wait group
 					for _, pod := range podToCreateEachTask {
 						go func(pod *v1.Pod) {
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -18,6 +18,7 @@ package job
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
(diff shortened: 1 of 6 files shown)


hwdef (commented):
I don't think this is a good fix. If A waits for B, but B needs to pull a large image, it will take a long time, and two minutes is too short. In fact, I don't think setting a timeout is a good solution.

xiefan-github:
> I don't think this is a good fix. If A waits for B, but B needs to pull a large image, it will take a long time, and two minutes is too short. In fact, I don't think setting a timeout is a good solution.

You are right; as long as you are waiting, you will never know how long you should wait; 

we should remove the waiting, because the controller can make multiple judgments. In multiple control, create it if the create condition is met. Otherwise, we can do it next time.

hwdef:
Can you explain in detail how `controller can make multiple judgments`

xiefan-github:
> > I don't think this is a good fix. If A waits for B, but B needs to pull a large image, it will take a long time, and two minutes is too short. In fact, I don't think setting a timeout is a good solution.
> 
> You are right; as long as you are waiting, you will never know how long you should wait;
> 
> we should remove the waiting, because the controller can make multiple judgments. In multiple control, create it if the create condition is met. Otherwise, we can do it next time.



> I don't think this is a good fix. If A waits for B, but B needs to pull a large image, it will take a long time, and two minutes is too short. In fact, I don't think setting a timeout is a good solution.

[i think bug code](https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/job_controller_actions.go#LL487C12-L487C12)

```
func (cc *jobcontroller) waitDependsOnTaskMeetCondition(taskName string, taskIndex int, podToCreateEachTask []*v1.Pod, job *batch.Job) {
	if job.Spec.Tasks[taskIndex].DependsOn == nil {
		return
	}

	dependsOn := *job.Spec.Tasks[taskIndex].DependsOn
	if len(dependsOn.Name) > 1 && dependsOn.Iteration == batch.IterationAny {
		wait.PollInfinite(detectionPeriodOfDependsOntask, func() (bool, error) {
			for _, task := range dependsOn.Name {
				if cc.isDependsOnPodsReady(task, job) {
					return true, nil
				}
			}
			return false, nil
		})
	} else {
		for _, dependsOnTask := range dependsOn.Name {
			wait.PollInfinite(detectionPeriodOfDependsOntask, func() (bool, error) {
				if cc.isDependsOnPodsReady(dependsOnTask, job) {
					return true, nil
				}
				return false, nil
			})
		}
	}
}
```
If the `wait.PollInfinite` is not terminated, it will wait forever in job preempted

do you think this is a good solution？

```
func (cc *jobcontroller) waitDependsOnTaskMeetCondition(taskName string, taskIndex int, podToCreateEachTask []*v1.Pod, job *batch.Job) bool {
	if job.Spec.Tasks[taskIndex].DependsOn == nil {
		return true
	}
	dependsOn := *job.Spec.Tasks[taskIndex].DependsOn
	if len(dependsOn.Name) > 1 && dependsOn.Iteration == batch.IterationAny {
		// any ready to create task
		for _, task := range dependsOn.Name {
			if cc.isDependsOnPodsReady(task, job) {
				return true
			}
		}
		return false
	} else {
		for _, dependsOnTask := range dependsOn.Name {
			// any not ready to skip
			if !cc.isDependsOnPodsReady(dependsOnTask, job) {
				return false
			}
			return true
		}
	}
	return true
}
```

hwdef:
I probably understand your idea, you want to re-trigger the pod creation through the controller, but are you sure it will re-trigger?

xiefan-github:
> I probably understand your idea, you want to re-trigger the pod creation through the controller, but are you sure it will re-trigger?

I tested it ok, although I don't understand why the volcano doesn't use resync,
Even so, the watch event is enough for the volcano, I will prove my idea through the source code，job controller watch  event as follows：

```
// https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/job_controller.go#L155
	cc.jobInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc:    cc.addJob,
		UpdateFunc: cc.updateJob,
		DeleteFunc: cc.deleteJob,
	})
// https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/job_controller.go#LL189C6-L189C6
	cc.podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc:    cc.addPod,
		UpdateFunc: cc.updatePod,
		DeleteFunc: cc.deletePod,
	})
```
A dependsOn B, if B pods no ready, we skip to create A.  when pod B ready, job controller add the B event to queue,
it will  run SyncJobAction -> SyncJob -> syncJob(jobInfo *apis.JobInfo, updateStatus state.UpdateStatusFn) -> waitDependsOnTaskMeetCondition -> ok -> create A

Maybe, there is something I don't know, but we have tested this fix, It's ok.  

If we doesn't fix it, preempt has a bug, example A dependsOn B, A and B was killed all，if controller create B first, it will use a controller work to wait, if 3 work[(default workNum) ](https://github.com/volcano-sh/volcano/blob/master/cmd/controller-manager/app/options/options.go#LL32C2-L32C16) used to wait, the controller will die.

hwdef:
@xiefan-github 
Thanks for your test, I will test your submitted code again later, if there is no problem, I will approve this modification.

hwdef (commented):
Tested, this code is ok, thanks for your contribution, this fix is awesome!
/approve

hwdef:
@wangyang0616 please trigger the CI

hwdef (commented):
/lgtm

pkg/controllers/job/job_controller_actions.go
```diff
@@ -378,7 +377,16 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 		go func(taskName string, podToCreateEachTask []*v1.Pod) {
 			taskIndex := jobhelpers.GetTasklndexUnderJob(taskName, job)
 			if job.Spec.Tasks[taskIndex].DependsOn != nil {
-				cc.waitDependsOnTaskMeetCondition(taskName, taskIndex, podToCreateEachTask, job)
+				if !cc.waitDependsOnTaskMeetCondition(taskName, taskIndex, podToCreateEachTask, job) {
+					klog.V(4).Infof("Job %s/%s depends on task not ready", job.Name, job.Namespace)
```
hwdef, line 381: Please make CI happy,
I think we should improve log level

pkg/controllers/job/job_controller_actions.go
```diff
@@ -378,7 +377,16 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 		go func(taskName string, podToCreateEachTask []*v1.Pod) {
 			taskIndex := jobhelpers.GetTasklndexUnderJob(taskName, job)
 			if job.Spec.Tasks[taskIndex].DependsOn != nil {
-				cc.waitDependsOnTaskMeetCondition(taskName, taskIndex, podToCreateEachTask, job)
+				if !cc.waitDependsOnTaskMeetCondition(taskName, taskIndex, podToCreateEachTask, job) {
+					klog.Errorf("Job %s/%s depends on task not ready", job.Name, job.Namespace)
```
hwdef, line 381:
```suggestion
					klog.V(3).Infof("Job %s/%s depends on task not ready", job.Name, job.Namespace)
```

renwenlong-github:
I verify codes success in my local, but it not work in volcano verity, how can I fix it?

hwdef:
It may be an occasional error, but we're not in a hurry to fix it now because spark's CI also has an error (it's not you), we can wait for the spark CI error to be fixed and then retry verify's CI

renwenlong-github:
> It may be an occasional error, but we're not in a hurry to fix it now because spark's CI also has an error (it's not you), we can wait for the spark CI error to be fixed and then retry verify's CI

Wait for spark fix, help me to ci, thank you

hwdef (commented):
/lgtm

hwdef (commented):
/approve

william-wang (approved):
/lgtm
/approve

Issue #2881, opened 2023-05-30
Resourcequota does not take effect in vcjob elastic job


wangyang0616:
**What happened**:
1. In the same queue, use the resourcequota of the namespace to limit the resource quota of each user (in this way to achieve the global priority of multi-tenant jobs)
2. The resourcequota plugin is enabled on Volcano. For the scenario where `minAvailble == replicas`, the normal work is as expected.
3. For the scenario where `minAvailble < replicas`, when minAvailble is less than user resourcequota free resources, and replicas is greater than resourcequota free resources, vcjob failed (for example: minAvailble: 3, replicas: 8, resourcequota free: 5)

**What you expected to happen**:
1. vcjob can run normally according to the number of minAvailble
2. The vcjob is temporarily pending, waiting for resourcequota idle resources to meet the replicas requirements and then scheduling and running

The ideal situation is described in 1, but the actual situation should be 2, vcjob can wait in line, do not fail directly. See `Anything else we need to know?`
**How to reproduce it (as minimally and precisely as possible)**:
1. namespace resource quota: 5C
2. Create vcjob, minAvailble: 3, replicas: 8, 1C/pod
3. Submit vcjob

**Anything else we need to know?**:
The resourcequota plugin judges whether the upper limit of the resourcequota of the namespace is reached according to minAvailble. If not, the podgroup of the vcjob is allowed to update to the inqueue state.
https://github.com/volcano-sh/volcano/blob/790ca3006e1be60ec320ea142ff8b0d394876147/pkg/scheduler/plugins/resourcequota/resourcequota.go#L42-L64

When the status of the podgroup is inqueue, the vc-controller will create the same number of pods according to the replicas corresponding to the vcjob. When more pods are created than minAvailble, the upper limit of resource quota will be reached, and the pod will fail to be created. At this time, vcjob will return fail.
https://github.com/volcano-sh/volcano/blob/790ca3006e1be60ec320ea142ff8b0d394876147/pkg/controllers/job/job_controller_actions.go#L347-L388

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

zhoushuke:
+1

hwdef:
still need

PR #2899, opened 2023-06-07
resourcequota supports flexible jobs

wangyang0616:
fix: [#2881](https://github.com/volcano-sh/volcano/issues/2881)


hwdef:
please fi the CI

lowang-bh:
how about add some uts about thoes changes?

googs1025:
Do we still want this proposal? If so, can we move forward with this?

PR #2915, opened 2023-06-15
add installCRDs varible

lengrongfu:
add `installCRDs` to helm values.yaml file, if set `true`, CRD resources will be installed as part of the Helm chart. if set `false`, CRD resources will be not as part of the Helm chart.


lowang-bh:
which scenario needs this feature? In my opinion, most of those CRDs are necessary and volcano cannot works without crds

william-wang:
> 



> It is mainly used to skip the installation of crd during helm upgrade.

@lengrongfu What's the reason of skipping the crd installation during helm upgrade?  The reinstallation CRD impacts the running workload?

PR #2916, opened 2023-06-15
Refactor PredicateFn for allocate and preempt actions

wangyang0616:
fix: [#2739](https://github.com/volcano-sh/volcano/issues/2739)

## Problem phenomenon:
Related issue: [Reclaim is not working with GPUSharingEnable if resource used is volcano.sh/gpu-memory #2739](https://github.com/volcano-sh/volcano/issues/2739)

Submit low-priority jobs to occupy all the GPU resources of the cluster, and submit high-priority jobs to request GPU resources.
- **Expectation:** High-priority jobs trigger preemption and run successfully, while some low-priority jobs are evicted and become pending;
- **Actual result:** The high-priority job does not trigger the preemption action and remains in the pending state;

## problem causes:
The scheduling process requires advanced node filtering and then optimization. The corresponding processes in Volcano are predicate and nodeorder, and preemption also needs to go through these two stages. In the predicate stage, each plug-in will be called to implement predicateFn to filter nodes. If a plug-in judges that the node does not meet the scheduling requirements and returns an error, the node will be filtered out.

When vgpu's predicateFn (FilterNode) performs node filtering judgment, it judges the remaining gpu resources of the node. If the remaining resources do not meet the pod scheduling, an error will be reported. This implementation is reasonable for allocate, but in preempt scenarios, judgment The logic of whether a pod can be scheduled to a node is whether the full resources of the node can satisfy pod scheduling, not the remaining resources, which causes this problem.

## Extended troubleshooting for similar issues:
In addition to the above-mentioned problems in the vgpu resource, check the implementation of other resources and plug-ins, and find that similar problems exist in many places, for example: the predicate-plugin has an upper limit on the number of pods supported by the filtering node node; it is compatible with kube-scheduler-related Filter functions Filtering logic, node port, pod affinity, node volume, pod topology spread, etc. all have similar problems, and it is necessary to provide a general solution to fix the above problems.

### Solution 1: 

Split the predicate function, and classify the resources that can be preempted and the resources that cannot be preempted. The details are as follows:

1. The predicateResourceFn function is added, and the logical judgment of dynamically changing (preemptible) resources is unified in the function processing, including: remaining resources of gpu, remaining schedulable pod quantity of node, ports supported by node, volume, pod topology spread, usage, etc.
1. The original predicateFn function handles node immutable (non-preemptive) resources, such as: taint toleration, node affinity, volume zone, numaaware, etc.
1. The allocate and backfile phases execute predicateFn and predicateResourceFn, which are unchanged from the original logic
1. Preempt and reclaim execute predicatFn, and when the node is filtered, it is judged that the node is immutable resource, so the above problem can be solved

**The pr corresponding to this program:**  [Split the predicate function to solve the resource filtering problem encountered during preemption #2818](https://github.com/volcano-sh/volcano/pull/2818)

### shortcoming:
Splitting the predicate function requires exposing too many filtering functions, the semantic distinction is not clear enough, and it is difficult for development and users to understand and maintain

### Solution 2: 

Extend the return value type of the predicate and add the Status type, indicating that after the node is filtered, scheduling, preemption is allowed, or preemption is not allowed. The details are as follows:

1. Add status, the status value is Success, Error, Unschedulable, UnschedulableAndUnresolvable, Wait, Skip, etc. (refer to and be compatible with the processing mode of Filter in kube-scheduler)
1. Micro-refactor the predicate function in each plug-in, allowing to return Unschedulable when preempted resources are insufficient, and returning UnschedulableAndUnresolvable when preempted resources are not allowed.
1. `allocate` and `backfill` accept nodes whose predicate status is Success, and `preempt` and `reclaim` accept nodes whose predicate status is Success and Unschedulable.

**This scheme corresponds to pr:** [Predicate adapts allocate and preempt #2916](https://github.com/volcano-sh/volcano/pull/2916)

### advantage:
1. No need to add a new predicate interface, all filtering logic is completed in one function, and the logic processing is more elegant
1. Compatible with kube-scheduler plug-in filter capabilities, more convenient to expand

## After comprehensive consideration, `Solution 2` is more suitable

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -17,7 +17,6 @@
 package allocate
 
 import (
-	"fmt"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -177,7 +176,13 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 			klog.V(3).Infof("There are <%d> nodes for Job <%v/%v>", len(ssn.Nodes), job.Namespace, job.Name)
 
-			if err := prePredicateforAllocate(ssn, task, allNodes, job); err != nil {
+			if err := ssn.PrePredicateFn(task); err != nil {
+				klog.V(3).Infof("PrePredicate for task %s/%s failed for: %v", task.Namespace, task.Name, err)
+				fitErrors := api.NewFitErrors()
+				for _, ni := range allNodes {
+					fitErrors.SetNodeError(ni.Name, err)
+				}
+				job.NodesFitErrors[task.UID] = fitErrors
 				break
 			}
 
@@ -252,34 +257,6 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 			}
 		}
 	}
-}
-
-func prePredicateforAllocate(ssn *framework.Session, task *api.TaskInfo, allNodes []*api.NodeInfo, job *api.JobInfo) error {
-	prePredicateStatus, err := ssn.PrePredicateFn(task)
-	if err != nil {
-		klog.V(3).Infof("PrePredicate for task %s/%s failed for: %v", task.Namespace, task.Name, err)
-		fitErrors := api.NewFitErrors()
-		for _, ni := range allNodes {
-			fitErrors.SetNodeError(ni.Name, err)
-		}
-		job.NodesFitErrors[task.UID] = fitErrors
-		return fmt.Errorf("PrePredicate for task %s/%s failed for: %v", task.Namespace, task.Name, err)
-	}
-
-	for _, status := range prePredicateStatus {
-		if status != nil && status.Code != api.Success {
-			klog.V(3).Infof("PrePredicate for task %s/%s failed, Code is %d, reason is  %s",
-				task.Namespace, task.Name, status.Code, status.Reason)
-			fitErrors := api.NewFitErrors()
-			err := fmt.Errorf("PrePredicate for task %s/%s, %s", task.Namespace, task.Name, status.Reason)
-			for _, ni := range allNodes {
-				fitErrors.SetNodeError(ni.Name, err)
-			}
-			job.NodesFitErrors[task.UID] = fitErrors
-			return err
-		}
-	}
-	return nil
 }
 
 func (alloc *Action) UnInitialize() {}
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -102,17 +102,19 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 		if ok, reason := task.InitResreq.LessEqualWithReason(node.FutureIdle(), api.Zero); !ok {
 			return nil, api.NewFitError(task, node, reason)
 		}
-
-		predicateStatus, err := ssn.PredicateFn(task, node)
+		var statusSets util.StatusSets
+		statusSets, err := ssn.PredicateFn(task, node)
 		if err != nil {
 			return nil, fmt.Errorf("allocate predicates failed for task <%s/%s> on node <%s>: %v",
 				task.Namespace, task.Name, node.Name, err)
 		}
-		// Only nodes whose status is success after predicate filtering can be scheduled.
-		admitStatus := map[int]struct{}{
-			api.Success: {},
+
+		if statusSets.ContainsUnschedulable() || statusSets.ContainsUnschedulableAndUnresolvable() ||
+			statusSets.ContainsErrorSkipOrWait() {
+			return nil, fmt.Errorf("predicates failed in allocate for task <%s/%s> on node <%s>, status is not success",
+				task.Namespace, task.Name, node.Name)
 		}
-		return nil, util.CheckPredicateStatus(predicateStatus, admitStatus)
+		return nil, nil
 	}
 
 	// To pick <namespace, queue> tuple for job, we choose to pick namespace firstly.
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -17,6 +17,7 @@
 package allocate
 
 import (
+	"fmt"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -102,11 +103,16 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 			return nil, api.NewFitError(task, node, reason)
 		}
 
+		predicateStatus, err := ssn.PredicateFn(task, node)
+		if err != nil {
+			return nil, fmt.Errorf("allocate predicates failed for task <%s/%s> on node <%s>: %v",
+				task.Namespace, task.Name, node.Name, err)
+		}
 		// Only nodes whose status is success after predicate filtering can be scheduled.
 		admitStatus := map[int]struct{}{
 			api.Success: {},
 		}
-		return nil, util.PredicateForAdmitStatus(ssn, task, node, admitStatus)
+		return nil, util.CheckPredicateStatus(predicateStatus, admitStatus)
 	}
 
 	// To pick <namespace, queue> tuple for job, we choose to pick namespace firstly.
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -17,6 +17,7 @@
 package allocate
 
 import (
+	"fmt"
 	"time"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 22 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -96,13 +97,22 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	pendingTasks := map[api.JobID]*util.PriorityQueue{}
 
 	allNodes := ssn.NodeList
-	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) error {
+	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
```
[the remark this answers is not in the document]
wangyang0616: The predicate plugin integrates multiple filter plug-ins of kube-scheduler. It is necessary to collect the filtering results of each plug-in and judge on the action side. 

At present, it is necessary to return the list list. If there is a more elegant solution in the future, we can use it in optimize.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -96,13 +97,22 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
 		// Check for Resource Predicate
 		if ok, reason := task.InitResreq.LessEqualWithReason(node.FutureIdle(), api.Zero); !ok {
-			return api.NewFitError(task, node, reason)
+			return nil, api.NewFitError(task, node, reason)
+		}
+		predicateStatus, err := ssn.PredicateFn(task, node)
```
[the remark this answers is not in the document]
wangyang0616: Thank you for your review， here is mainly the core logic that distinguishes the difference between allocate and preempt. It has been modified using the previous logical architecture. 

I will consider it later to see if it can be further optimized.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -96,13 +97,22 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
+		if err != nil {
+			return nil, err
+		}
+		for _, status := range predicateStatus {
+			if status != nil && status.Code != api.Success {
+				return nil, api.NewFitError(task, node, status.Reason)
```
william-wang, line 111: just return the first failure?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -250,4 +254,32 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	}
 }
 
+func prePredicateforAllocate(ssn *framework.Session, task *api.TaskInfo, allNodes []*api.NodeInfo, job *api.JobInfo) error {
```
william-wang, line 257: The `prePredicateforAllolcate` is similar with `prePredicateforBackfill` and `prePredicateforReclaim`

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -101,4 +96,45 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 	}
 }
 
+func predicateforBackfill(ssn *framework.Session, task *api.TaskInfo, node *api.NodeInfo, fe *api.FitErrors) error {
```
william-wang, line 99: Can we consolidate the `predicateforBackfill`, `predicateforReclaim`, `predicateforAllocate` and `predicateforPrempt`?

william-wang:
@Yikun Would you please have a look at the spark failure?

Yikun:
> @Yikun Would you please have a look at the spark failure?

It should be fixed by upgrading spark repo to 3.4, lets see results: https://github.com/volcano-sh/volcano/pull/2934

pkg/scheduler/actions/allocate/allocate.go
wangyang0616: done

pkg/scheduler/actions/backfill/backfill.go
wangyang0616: done

pkg/scheduler/api/devices/util.go
```diff
@@ -0,0 +1,38 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package devices
+
+// These are predefined codes used in a Status.
+const (
```
william-wang, line 18: These same status are defined in device package and api package. Is it possible to combine two to one?

pkg/scheduler/framework/util.go
```diff
@@ -263,3 +265,25 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
 	}
 	return nodes, nil
 }
+
+// The state of the k8s prefile is converted to the internal state of the volcano
```
william-wang, line 269: please provide the function signature formly.

pkg/scheduler/util/predicate_helper.go
```diff
@@ -100,6 +100,19 @@ func (ph *predicateHelper) PredicateNodes(task *api.TaskInfo, nodes []*api.NodeI
 	return predicateNodes, fe
 }
 
+func CheckPredicateStatus(predicateStatus []*api.Status, admitStatus map[int]struct{}) error {
+	for _, status := range predicateStatus {
+		if status == nil {
+			continue
+		}
+		if _, ok := admitStatus[status.Code]; !ok {
```
william-wang, line 108: what is the admitStatus used for?

pkg/scheduler/api/devices/util.go
wangyang0616: The Volcano API package references the devices package. If the status definition in the API is used, cyclic dependency occurs.

pkg/scheduler/util/predicate_helper.go
wangyang0616: Optimized

pkg/scheduler/framework/util.go
wangyang0616: done

pkg/scheduler/util/predicate_helper.go
```diff
@@ -107,3 +107,47 @@ func taskGroupID(task *api.TaskInfo) string {
 func NewPredicateHelper() PredicateHelper {
 	return &predicateHelper{taskPredicateErrorCache: map[string]map[string]error{}}
 }
+
+type PredicateStatus interface {
+	IsContainsUnschedulable() bool
```
william-wang, line 112: The `IsContainsUnschedulable` is inconsistent with implementation.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -72,13 +74,25 @@ func (backfill *Action) Execute(ssn *framework.Session) {
...
+					if err != nil {
+						klog.V(3).Infof("backfill predicates failed for task <%s/%s> on node <%s>: %v",
 							task.Namespace, task.Name, node.Name, err)
 						fe.SetNodeError(node.Name, err)
 						continue
 					}
 
+					if statusSets.ContainsUnschedulable() || statusSets.ContainsUnschedulableAndUnresolvable() ||
+						statusSets.ContainsErrorSkipOrWait() {
+						err := fmt.Errorf("predicates failed in backfill for task <%s/%s> on node <%s>, status is not success",
```
william-wang, line 89: please combine line #89 and #90

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -72,13 +74,25 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 				for _, node := range ssn.Nodes {
 					// TODO (k82cn): predicates did not consider pod number for now, there'll
 					// be ping-pong case here.
-					if err := ssn.PredicateFn(task, node); err != nil {
-						klog.V(3).Infof("Predicates failed for task <%s/%s> on node <%s>: %v",
+					// Only nodes whose status is success after predicate filtering can be scheduled.
+					var statusSets util.StatusSets
+					statusSets, err := ssn.PredicateFn(task, node)
+					if err != nil {
+						klog.V(3).Infof("backfill predicates failed for task <%s/%s> on node <%s>: %v",
```
william-wang, line 81: please make the log clear for user. for example. change the `backfill predicates failed xxx` to `predicate failed in backfill action xxx`

pkg/scheduler/util/predicate_helper.go
wangyang0616: It's supposed to be deleted here, done.

pkg/scheduler/actions/backfill/backfill.go
wangyang0616: done

william-wang (approved):
/lgtm
/approve

william-wang:
The proboot issue is fixed and need to mark the lgtm again.

PR #2863, opened 2023-05-26
Remove vendor

xiao-jay:
fix #2858 #2841 
After Go 1.12,Go default use go modules,because Go modules have truly achieved reproducible builds，vender may will use offline build,but this requirement is likely the choice of a minority in volcano,So decide to delete the vendor in volcano.

### PR Describe：
1、delete vendor
2、[modify hack/verify-vendor.sh to hack/verify-gomod.sh ](https://github.com/volcano-sh/volcano/pull/2863/commits/aa28ac9b4dab8aa0cf790a4c645d09af36b61937)
```
➜  hack git:(remove-vendor) ✗ ./verify-gomod.sh
running 'go mod tidy' for repo root
SUCCESS: go.mod Verified.
```

PR #2917, opened 2023-06-15
Let the make manifest work properly

wangyang0616:
[Remove vendor #2863](https://github.com/volcano-sh/volcano/pull/2863) cleans up the vendor directory in the code warehouse, but some scripts in the CI project depend on part of the content in the local vendor. After the pr is merged, the CI operation fails.

The current pr fixes the problem of CI running failure:
1. When the make manifest is executed, the advanced go mod vendor builds locally. The current method may not be the optimal solution, but it can ensure the normal operation of CI first, and then consider the optimization of the make manifest
2. The generate-code check uses the script in the hack directory
3. verify-vendor.sh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,6 @@ clean:
 verify:
 	hack/verify-gofmt.sh
 	hack/verify-gencode.sh
-	hack/verify-vendor.sh
 	hack/verify-vendor-licenses.sh
 
 lint: ## Lint the files
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,7 @@ generate-code:
 
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
+	go mod vendor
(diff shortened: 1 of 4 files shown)


xiao-jay:
change suggestion 
1、https://github.com/volcano-sh/volcano/blob/master/hack/update-gencode.sh#L29
should change from `bash vendor/k8s.io/code-generator/generate-groups.sh` to `hack/generate-groups.sh`
because `hack/verify-gencode.sh` will exec `"${SCRIPT_ROOT}/hack/update-gencode.sh"`
2、https://github.com/volcano-sh/volcano/blob/master/Makefile#LL159C2-L159C23
sbould change form `hack/verify-vendor.sh` to `hack/verify-gomod.sh`

lowang-bh:
`update-gencode.sh` also rely on vendor. 

I think we need to search where vender is used and solve them. such as `verify-vendor-licenses.sh`, `hack/update-gencode.sh` and `hack/update-vendor-licenses.sh`.

Makefile
```diff
@@ -156,7 +157,6 @@ clean:
 verify:
 	hack/verify-gofmt.sh
 	hack/verify-gencode.sh
-	hack/verify-vendor.sh
```
hzxuzhonghu, line 159: why remove this verify

wangyang0616: This check is mainly to determine whether the vendor in the local directory is the latest version. 

The vendor directory has been removed from the code warehouse, so this check can also be removed

william-wang (approved):
/lgtm

Makefile
hzxuzhonghu: oh, good to know

PR #2922, opened 2023-06-18
add task status named ReleasingFailed

ycfnana:
if pod being terminating for long time because of zombie process，the task will be scheduled this node，and then this job may hang until the pod with zombie process killed force

### DIFF
--- a/pkg/scheduler/api/helpers.go
+++ b/pkg/scheduler/api/helpers.go
@@ -44,6 +44,8 @@ func getTaskStatus(pod *v1.Pod) TaskStatus {
 		if pod.DeletionTimestamp != nil &&
 			time.Now().Unix()-pod.DeletionTimestamp.Unix() <= gracePeriodSeconds {
 			return Releasing
+		} else if pod.DeletionTimestamp != nil {
+			return ReleasingFailed
 		}
 
 		return Running
@@ -51,6 +53,8 @@ func getTaskStatus(pod *v1.Pod) TaskStatus {
 		if pod.DeletionTimestamp != nil &&
 			time.Now().Unix()-pod.DeletionTimestamp.Unix() <= gracePeriodSeconds {
 			return Releasing
+		} else if pod.DeletionTimestamp != nil {
+			return ReleasingFailed
 		}
 
 		if len(pod.Spec.NodeName) == 0 {
@@ -71,7 +75,7 @@ func getTaskStatus(pod *v1.Pod) TaskStatus {
 // AllocatedStatus checks whether the tasks has AllocatedStatus
 func AllocatedStatus(status TaskStatus) bool {
 	switch status {
-	case Bound, Binding, Running, Allocated:
+	case Bound, Binding, Running, Allocated, ReleasingFailed:
 		return true
 	default:
 		return false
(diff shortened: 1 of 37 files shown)


hwdef:
How does this PR avoid the problems you raised?

ycfnana:
> How does this PR fix the problems you raised?

    job may be status pipelined when this case happens. if job request resource is future idle, it will be pipelined status, the releasing resource will be used in the way of the future idle.
    I think if the pod has something wrong with termimating process, it should be detected.  the process of pod terminating, will send signal TERM and wait pod.Spec.TerminationGracePeriodSeconds(default 30s) to send signal kill. I think pod terminating has something wrong if wait time exceed TerminationGracePeriodSeconds.
    however, I think this code will result in task status from Releasing to Running, it's not good but I have no idea to chang it to another status

hwdef:
From the code point of view, you just make the conditions for returning releasing more stringent

ycfnana:
> From the code point of view, you just make the conditions for returning releasing more stringent

my fault, its title is  vague. I'll update it
 I met a issue, pod always be Pending when one node has zombie pod and another node has enough reource, due to the relaesing status, I think waiting time larger than  TerminationGracePeriodSeconds it should not be releaseing, may happen something error,  its should be another status.

pkg/scheduler/api/helpers.go
```diff
@@ -33,15 +34,22 @@ func PodKey(pod *v1.Pod) TaskID {
...
+		if pod.DeletionTimestamp != nil &&
+			time.Now().Unix()-pod.DeletionTimestamp.Unix() <= gracePeriodSeconds {
 			return Releasing
 		}
 
 		return Running
 	case v1.PodPending:
-		if pod.DeletionTimestamp != nil {
+		if pod.DeletionTimestamp != nil &&
+			time.Now().Unix()-pod.DeletionTimestamp.Unix() <= gracePeriodSeconds {
```
hwdef, line 52: `time.Now().Unix()-pod.DeletionTimestamp.Unix()`will keep growing, If once the time point of less than or equal to is missed, then this judgment will always be false and will never return release. Is this what we want?

ycfnana: it means this pod is always exists with terminating status if `time.Now().Unix()-pod.DeletionTimestamp.Unix()`keep growing, resource cannot be freed. this resource is always allocated for scheduler.

ycfnana: > `time.Now().Unix()-pod.DeletionTimestamp.Unix()`will keep growing, If once the time point of less than or equal to is missed, then this judgment will always be false and will never return release. Is this what we want?

I think gracePeriodSeconds can set bigger,  it should be another status not "Running" status when waiting time bigger than gracePeriodSeconds, but I have no idea what status to set

hwdef: Sorry I don't think this is a perfect fix

ycfnana: I think should add a status  like "ReleaseFailed", in the code of resource calculation, it should be the same as Running. what do you think?

hwdef: @wangyang0616 
Can you help review this PR

ycfnana: yeah, I'll close it and fix it with add staus. when I finish it I'll reopen it again

Issue #2928, opened 2023-06-20
PodTopologySpread plugin violates maxSkew because pod is added twice in cache. 


Monokaix:
**What happened**:
when pods specify topologySpreadConstraints, it should spread evenly across topology key and not violates maxSkew.
**What you expected to happen**:
pods specified topologySpreadConstraints spread evenly and not violates maxSkew.
**How to reproduce it (as minimally and precisely as possible)**:
1. deploy a k8s cluster with three nodes with label topology.kubernetes.io/zone="az1"
2. disable one node schedule with following commands:
`kubectl cordon $node`
3. run a deploy `demo` with replicas=2 and label app=test, also pod template has following spec
```yaml
        topologySpreadConstraints:
        - labelSelector:
            matchLabels:
              app: test
          maxSkew: 1
          topologyKey: topology.kubernetes.io/zone
          whenUnsatisfiable: DoNotSchedule
``` 
4. modify node `$node` with with label topology.kubernetes.io/zone="az2"
5. recover $node schedule
`kubectl uncordon $node`
7. scale deploy with replicas=4
`kubectl scale deploy demo --replicas=4`
8. check pods spread 
we can find that new created two pods are not all in az2, az1 has 3 pods, but az2 only has one.

**Anything else we need to know?**:

**Environment**:
- Volcano Version:  master
- Kubernetes version (use `kubectl version`): v1.25
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #2929, opened 2023-06-20
fix PodToplogySpread plugin violates maxSkew

Monokaix:
This pr https://github.com/volcano-sh/volcano/pull/2606 refact nodeorder and predict plugin and move `util.NewPodListerFromNode(ssn)` from plugin local to global variable, so we just need add pod into cache once at eventHandler.

fix https://github.com/volcano-sh/volcano/issues/2928

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -166,40 +166,7 @@ func calculateWeight(args framework.Arguments) priorityWeight {
 
 func (pp *nodeOrderPlugin) OnSessionOpen(ssn *framework.Session) {
 	weight := calculateWeight(pp.pluginArguments)
-	pl := ssn.PodLister
 	nodeMap := ssn.NodeMap
-
-	// Register event handlers to update task info in PodLister & nodeMap
-	ssn.AddEventHandler(&framework.EventHandler{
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -171,18 +171,6 @@ func (pp *nodeOrderPlugin) OnSessionOpen(ssn *framework.Session) {
...
-			nodeName := event.Task.NodeName
-			node, found := nodeMap[nodeName]
-			if !found {
-				klog.Warningf("node order, update pod %s/%s allocate to NOT EXIST node [%s]", pod.Namespace, pod.Name, nodeName)
-			} else {
-				node.AddPod(pod)
-				klog.V(4).Infof("node order, update pod %s/%s allocate to node [%s]", pod.Namespace, pod.Name, nodeName)
-			}
-		},
 		DeallocateFunc: func(event *framework.Event) {
```
william-wang, line 174: @Monokaix Don't we need to remove the DeallocateFunc?

Monokaix: yeah it should be removed too.

william-wang:
@Monokaix please take a look at the e2e failure case

PR #2882, opened 2023-05-30
update deploy yamls about jobflows by make generate-yaml & make update-development-yaml

lowang-bh:
fix #2867 #2883 jobflow related issue

part-2: add jobflow's crd files in deploy yamls( generated by `make generate-yaml` & `make update-development-yaml`)

Please merge #2874 first.

Notes: it will fail to upgrade helm release via `helm upgrade --install volcano` due to yaml is too long

```sh
UPGRADE FAILED: create: failed to create: Secret "sh.helm.release.v1.volcano.v3" is invalid: data: Too long: must have at most 1048576 bytes
```

PR #2936, opened 2023-06-26
Add jobflow-related content to Volcano automatic installation and deployment

wangyang0616:
Add the crd definition and helm package content of jobflow, and solve the problem that the content of the helm package in [#2882](https://github.com/volcano-sh/volcano/pull/2882) exceeds the upper limit of 1MB.

The specific content is as follows:
The crd in Volcano is divided into two file groups to independently manage the contents of the helm package
1. Volcano file group: existing Volcano crd definition
2. jobflow file group: jobflow related crd definition
Install jobflow as a dependency of Volcano

When the user's local source code uses helm to deploy Volcano, perform the following operations:
1. helm dependency update installer/helm/chart/volcano
2. helm install volcano installer/helm/chart/volcano

Compared with the master, the operation of updating the dependent chart package in the first step is added


lowang-bh:
I test the helm install and it works.

One more thing, user need to run two commands when installing it first time.  And user will met an error and the prompted `helm dependency build` does not work.
```Error: INSTALLATION FAILED: An error occurred while checking for chart dependencies. You may need to run `helm dependency build` to fetch missing dependencies: found in Chart.yaml, but missing in charts/ directory: jobflow```

My opinion at first is to use subchart and with the same one command.

william-wang (approved):
/lgtm
/approve

PR #2943, opened 2023-06-28
add task status named ReleasingFailed

ycfnana:
the question is https://github.com/volcano-sh/volcano/pull/2922
I met a problem, pod always be Pending when one node has zombie pod and another node has enough reource, due to the relaesing status, I think waiting time larger than TerminationGracePeriodSeconds it should not be releaseing, may happen something error, its should be another status like ReleasingFailed

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -42,6 +42,8 @@ const (
 	defaultMinNodesToFind             = 100
 	defaultPercentageOfNodesToFind    = 100
 	defaultLockObjectNamespace        = "volcano-system"
+	defaultGracePeriodSeconds         = 30
+	defaultGracePeriodSecondsWait     = 3
 )
 
 // ServerOption is the main context object for the controller manager.
@@ -75,6 +77,9 @@ type ServerOption struct {
 
 	NodeSelector      []string
 	EnableCacheDumper bool
+
+	GracePeriodSeconds     int64
+	GracePeriodSecondsWait int64
 }
 
 type DecryptFunc func(c *ServerOption) error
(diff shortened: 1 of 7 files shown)


lowang-bh:
Thanks for your contributions!
This is a great change and will have a large effect.
Would you please provide those informations?
1. the necessary ut tests.
2. the scenario what happens without the change, and the result with this fix.

ycfnana:
> Thanks for your contributions! This is a great change and will have a large effect. Would you please provide those informations?
> 
> 1. the necessary ut tests.
> 2. the scenario what happens without the change, and the result with this fix.




> progress

ycfnana:
> Thanks for your contributions! This is a great change and will have a large effect. Would you please provide those informations?
> 
> 1. the necessary ut tests.
> 2. the scenario what happens without the change, and the result with this fix.


Thanks, I'll  add ut test next week.
I have an issue, there are enough resource both in nodeA and nodeB.  PodB in nodeB has been deleted,  but the status is still "Terminating" due to the zombie progress.
In the time,  add the new job named  JobC with PodD, it can be scheduled in nodeA, but status of podD is Pending. From its log, I find status of JobC is "Pipelined", waiting nodeB to releasing podB. But podB has something wrong like "ReleasingFailed". 
So I want to fix this condition.

ycfnana:
> Thanks for your contributions! This is a great change and will have a large effect. Would you please provide those informations?
> 
> 1. the necessary ut tests.
> 2. the scenario what happens without the change, and the result with this fix.

can you help review it again?

lowang-bh:
> > Thanks for your contributions! This is a great change and will have a large effect. Would you please provide those informations?
> > 
> > 1. the necessary ut tests.
> > 2. the scenario what happens without the change, and the result with this fix.
> 
> can you help review it again?

It is clear to me now. Just a little comment, please have a look at the review comment. Thanks.

ycfnana:
/assign @wangyang0616 @william-wang @huone1 
can somebody help review it

wangyang0616:
Can this pr [#2815](https://github.com/volcano-sh/volcano/pull/2815) fix your problem.

Thor-wl:
@ycfnana Thanks for your report. Personally speaking, I think the root reason is something goes wrong with the unexpected pod status instead of task's status. There is no enough resource to start new pods, which belong to the pipelined jobs. So I don't get the point why should add a task status and what this can help for users.
/hold

ycfnana:
> @ycfnana Thanks for your report. Personally speaking, I think the root reason is something goes wrong with the unexpected pod status instead of task's status. There is no enough resource to start new pods, which belong to the pipelined jobs. So I don't get the point why should add a task status and what this can help for users. /hold

thanks,  sometimes pod status is pending even if there are enough resources to start a new pod because of  unexpected pod status, you can see my comment in my ut code. But this pr https://github.com/volcano-sh/volcano/pull/2815 may solve my problem. I'll close this pr if test it's ok

PR #2959, opened 2023-07-10
Add NodeOrder to devices API

archlitchi:
Add NodeOrder to devices API. Devices API should provide the ability to score each node based on corresponding device information of this node. A device which don't want to participate in score process, can simply return a (0,nil) in ScoreNode.

Future work includes implement "Mellanox" device in devices API, which can provide more accurate scheduling decision making based on Infiniband-network topology. It is crucial for large-scale training cluster which uses mellanox for InfiniBand data exchange.

### DIFF
--- a/pkg/scheduler/plugins/devicescore/devicescore.go
+++ b/pkg/scheduler/plugins/devicescore/devicescore.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package devicescore
+
+import (
+	"k8s.io/klog/v2"
+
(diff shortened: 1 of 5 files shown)


pkg/scheduler/plugins/devicescore/devicescore.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
wangyang0616, line 2: The time here can be updated.

pkg/scheduler/plugins/devicescore/devicescore.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
+	ssn.AddNodeOrderFn(pp.Name(), func(task *api.TaskInfo, node *api.NodeInfo) (float64, error) {
+		score := float64(0)
+		for _, val := range api.RegisteredDevices {
+			if devices, ok := node.Others[val].(api.Devices); ok {
+				if !devices.HasDeviceRequest(task.Pod) {
+					continue
+				}
+				tmp, err := devices.ScoreNode(task.Pod)
```
wangyang0616, line 51: Is there a better way to name the tmp variable?

pkg/scheduler/plugins/devicescore/devicescore.go
```diff
@@ -0,0 +1,65 @@
...
+				if err != nil {
+					klog.Warningln("scoreNode failed in predicate nodeorderFn", err.Error())
+					return 0, err
+				}
+				score += tmp
```
wangyang0616, line 56: Does deveice's scoring strategy support weight configuration?

PR #2967, opened 2023-07-13
Allow preemptable tasks schedule without checking proportion

### LINKED ISSUE #2966: Allow schedule preemptable tasks on nodes even if proportional predicate failed

Do not check proportional predicate for tasks which are preemptable.

#### Why is this needed:
Preemptable tasks can be preempted, so disallowing them to schedule on nodes by predicates is not a good decision. They can be preempted and it's okay.

Cdayz:
Resolve #2966

### DIFF
--- a/pkg/scheduler/plugins/predicates/proportional.go
+++ b/pkg/scheduler/plugins/predicates/proportional.go
@@ -27,6 +27,12 @@ import (
 // checkNodeResourceIsProportional checks if a gpu:cpu:memory is Proportional
 func checkNodeResourceIsProportional(task *api.TaskInfo, node *api.NodeInfo, proportional map[v1.ResourceName]baseResource) (*api.Status, error) {
 	status := &api.Status{}
+
+	if task.Preemptable {
+		status.Code = api.Success
+		return status, nil
+	}
+
 	for resourceName := range proportional {
 		if value, found := task.Resreq.ScalarResources[resourceName]; found && value > 0 {
 			status.Code = api.Success


pkg/scheduler/plugins/predicates/proportional.go
```diff
@@ -27,6 +27,12 @@ import (
 // checkNodeResourceIsProportional checks if a gpu:cpu:memory is Proportional
 func checkNodeResourceIsProportional(task *api.TaskInfo, node *api.NodeInfo, proportional map[v1.ResourceName]baseResource) (*api.Status, error) {
 	status := &api.Status{}
+
+	if task.Preemptable {
```
wangyang0616, line 31: The task here is the role of the preemptor, and `task.Preemptable` indicates whether the current task is allowed to be preempted.
Sorry, I don't quite understand the purpose of the judgment here.

Cdayz: Main idea of this PR, that preemptable tasks can skip this proportional check, because they can be preempted, so there are no problem to schedule them. In system which I maintain in my organisation should be available ability to schedule preemptible tasks bypassing proportional plugin.

lowang-bh:
It's ok to me.

pkg/scheduler/plugins/predicates/proportional.go
wangyang0616: Thanks for your response, I see what you mean.

However, the proportional logic is designed to ensure that resources are available in the allocate scene, and is not bound to preempt. In reality, many users will not configure preempt actions.

I am worried that the modification here will affect the existing proportional function. So I have reservations about the current pr, and I can refer to the ideas of other contributors in the community.

Cdayz: Maybe we can add an optional parameter to proportional plugin configuration, which by default will save current behavior but allow to enable changes made by this pr?

lowang-bh: It sounds more reasonable to add a config. And what's more, job's queue should support `Reclaimable` so that the job      in this queue can be reclaimed.

Cdayz:
I think we can close this PR in prior to #3158 where implemented more wide check.

PR #2973, opened 2023-07-13
add support for volumeBinding plugin

chxk:
This submission is based on a previous PR(#2506), and fixes the issue where the volumeBinding plugin did not correctly filter nodes when performing predicates on multiple tasks of a job.


william-wang:
@chxk Thanks for the pr :)  Would you log a issue firstly, and add provide the following information？
**What happened**:

chxk:
> @chxk Thanks for the pr :) Would you log a issue firstly, and add provide the following information？ **What happened**: **What you expected to happen**: **How to reproduce it (as minimally and precisely as possible)**:

**What happened:** 
Currently Volcano can't filter nodes by PV, which can't meet our requirements. We found a PR(#2506) which added this feature support (based on the volumeBinding plugin) before, but it was reverted. 
After investigation, we found the problem introduced by this PR: 
When allocating nodes for each task in a job, it uses the `ssn.cache.AllocateVolumes`, which based on an already initialized `SchedulerVolumeBinder`, and will update its `pvCache` and `pvcCache`. However, every time a task is predicated, a new `volumeBinding plugin` is initialized, and the empty `pvCache` and `pvcCache` are used for `Filter`, instead of  which in the initialized `SchedulerVolumeBinder`.  This can potentially cause a PV to be bound to multiple different PVCs.

**What you expected to happen:**
Add volumeBinding feature to Volcano and it should work normally.
After our changes were merged and deployed, we have not observed the memory leak issue that was previously mentioned in #2555。

**How to reproduce it (as minimally and precisely as possible):**
based on a version with PR(#2506):
1. Create a local PV with storageClass `fast-disks`.
2. Create a job with two pods, each requesting a different PVC(storageClass is both `fast-disks`) and using a nodeSelector pointing to the node of the local PV. The yaml file is as following
```
kind: PodGroup
metadata:
  name: test-binding
  namespace: default
spec:
  minMember: 2
  queue: default
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-binding-1
  namespace: default
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 50Gi
  storageClassName: fast-disks
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-binding-2
  namespace: default
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 50Gi
  storageClassName: fast-disks
---
apiVersion: v1
kind: Pod
metadata:
  name: test-binding-1
  namespace: default
  annotations:
    scheduling.k8s.io/group-name: test-binding
spec:
  schedulerName: volcano
  nodeSelector:
    testpvc1: owner
  containers:
  - name: test
    image: <your image>
    resources:
      limits:
        cpu: 3
    command:
      - /bin/bash
      - -c
      - sleep 1d
    volumeMounts:
    - name: hahaha
      mountPath: /fast-disks
  volumes:
  - name: hahaha
    persistentVolumeClaim:
      claimName: test-binding-1
---
apiVersion: v1
kind: Pod
metadata:
  name: test-binding-2
  namespace: default
  annotations:
    scheduling.k8s.io/group-name: test-binding
spec:
  schedulerName: volcano
  nodeSelector:
    testpvc: owner
  containers:
  - name: test
    image: <your image>
    resources:
      limits:
        cpu: 3
    command:
      - /bin/bash
      - -c
      - sleep 1d
    volumeMounts:
    - name: hahaha
      mountPath: /fast-disks
  volumes:
  - name: hahaha
    persistentVolumeClaim:
      claimName: test-binding-2
```
4. Observe the result: the job gets scheduled, but only one pod's PVC gets bound to the local PV.

hwdef (commented):
I don't know much about this area.
I think the code is ok.

hwdef:
lgtm

chxk:
> I don't know much about this area. I think the code is ok.

Thank you for review. 
Could you help review this pr please? @Thor-wl @william-wang

bibibox:
i met same problem, and i think  this PR can solve it greatly

Issue #2996, opened 2023-07-25
support controller param: concurrent-podgroup-syncs


WulixuanS:
#### What would you like to be added:
A separate parameter to adjust the concurrency of pg sync
#### Why is this needed:
There is no such parameter in version 1.7.
The Master branch has this parameter, but it will also adjust the job sync concurrency

PR #2997, opened 2023-07-25
support concurrent-podgroup-syncs

WulixuanS:
A separate parameter to adjust the concurrency of pg sync
support #2996

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -60,6 +60,9 @@ type ServerOption struct {
 	EnableHealthz      bool
 	// To determine whether inherit owner's annotations for pods when create podgroup
 	InheritOwnerAnnotations bool
+	// WorkerThreadsForPG is the number of threads syncing podgroup operations
+	// The larger the number, the faster the podgroup processing, but requires more CPU load.
+	WorkerThreadsForPG uint32
 }
 
 type DecryptFunc func(c *ServerOption) error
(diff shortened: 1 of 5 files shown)


lowang-bh:
There is a param `WorkerThreads` to control it. Why do you add another one?

WulixuanS:
> There is a param `WorkerThreads` to control it. Why do you add another one?
1. param `WorkerThreads` both controls job-controller and pg-controller. Now i used volcano to schedule pod, I only want to adjust pg-controller worker number to improve pg creation rate.
2. Each controller in K8S has independent syncs parameters to adjust, I think this is more flexible.

cmd/controller-manager/app/options/options.go
```diff
@@ -60,6 +60,9 @@ type ServerOption struct {
 	EnableHealthz      bool
 	// To determine whether inherit owner's annotations for pods when create podgroup
 	InheritOwnerAnnotations bool
+	// ConcurrentPGSyncs is the number of pod groups that are allowed to sync concurrently.
+	// Larger number = faster pg creating, but more CPU (and network) load
+	ConcurrentPGSyncs uint32
```
william-wang, line 65: It's better unifiy the naming style for job and podgroup. Here is an example:

// WorkerThreadsForPG is the number of threads syncing podgroup operations
// The larger the number, the faster the podgroup processing, but requires more CPU load.
WorkerThreadsForPG

cmd/controller-manager/app/options/options.go
```diff
@@ -90,6 +93,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.HealthzBindAddress, "healthz-address", defaultHealthzAddress, "The address to listen on for the health check server.")
 	fs.BoolVar(&s.EnableHealthz, "enable-healthz", false, "Enable the health check; it is false by default")
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
+	fs.Uint32Var(&s.ConcurrentPGSyncs, "concurrent-podgroup-syncs", 1, "the number of pod groups that are allowed to sync concurrently. Larger number = faster pg creating, but more CPU (and network) load")
```
william-wang, line 96: The same as above

william-wang:
> > There is a param `WorkerThreads` to control it. Why do you add another one?
> 
> 1. param `WorkerThreads` both controls job-controller and pg-controller. Now i used volcano to schedule pod, I only want to adjust pg-controller worker number to improve pg creation rate.
> 2. Each controller in K8S has independent syncs parameters to adjust, I think this is more flexible.
Would you share some testing data about improving pg create rate after adjust the worker threads?

william-wang (approved):
/lgtm

PR #2998, opened 2023-07-26
remove node out of sync state

### LINKED ISSUE #2999: node outOfSync state cause deadlock and block pod scheduled
**What happened**:
After the pod is scheduled to the node, when the allocated resources of the node decrease due to some reasons (such as an exception reported by the gpu device), the node will be set to outofsync state, and node will not be added to the session, new pod cannot be scheduled to the current node until the allocatable resources reported by the node become normal ,even if there are other idle resources on the node, the pod cannot be scheduled. If the pod is used to report gpu resources, the premise of the pod being scheduled is that the node ends the outofsync state , and the end of outofsync requires the gpu device to be scheduled and report resources correctly, which causes a deadlock
pod used to report gpu resource should be scheduled even though node is in outOfSync state.
1. Run a device-plugin daemonset to report gpu resource
2. Run a pod using gpu resource
3. Uninstall device-plugin daemonset and wait gpu resource of node allocatable become zero
4. Re-deploy device-plugin daemonset, one daemonset pod can't be scheduled 

**Environment**:
- Volcano Version: latest
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Monokaix:
Fix https://github.com/volcano-sh/volcano/issues/2999

- This pr remove outOfSync state of node and set idle resource to negative, and the node can be schedulable.
- It may also resolve some unfinished problem：https://github.com/volcano-sh/volcano/issues/1818

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -99,8 +99,8 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	allNodes := ssn.NodeList
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
 		// Check for Resource Predicate
-		if ok, reason := task.InitResreq.LessEqualWithReason(node.FutureIdle(), api.Zero); !ok {
-			return nil, api.NewFitError(task, node, reason)
+		if ok, resources := task.InitResreq.LessEqualWithResourcesName(node.FutureIdle(), api.Zero); !ok {
+			return nil, api.NewFitError(task, node, api.WrapInsufficientResourceReason(resources))
 		}
 		var statusSets util.StatusSets
 		statusSets, err := ssn.PredicateFn(task, node)
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -291,12 +291,8 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
 	}
 
 	// set NodeState according to resources
-	if !ni.Used.LessEqual(ni.Allocatable, Zero) {
-		ni.State = NodeState{
-			Phase:  NotReady,
-			Reason: "OutOfSync",
-		}
-		return
+	if ok, resources := ni.Used.LessEqualWithResourcesName(ni.Allocatable, Zero); !ok {
+		klog.ErrorS(nil, "Node out of sync", "name", ni.Name, "resources", resources)
 	}
 
 	// If node not ready, e.g. power off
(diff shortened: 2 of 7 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -99,8 +99,8 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	allNodes := ssn.NodeList
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
 		// Check for Resource Predicate
-		if ok, reason := task.InitResreq.LessEqualWithReason(node.FutureIdle(), api.Zero); !ok {
-			return nil, api.NewFitError(task, node, reason)
+		if ok, resource := task.InitResreq.LessEqualWithResource(node.FutureIdle(), api.Zero); !ok {
```
Thor-wl, line 102: @Monokaix Hi, I'm kindly asking about the reason or meaning of this PR. After going through all the modifications, I've not getting the meaningful points. It may be better to add descriptions for all PRs.

pkg/scheduler/api/resource_info.go
```diff
@@ -418,10 +418,10 @@ func (r *Resource) LessEqualWithReason(rr *Resource, defaultValue DimensionDefau
 	}
 
 	if !lessEqualFunc(r.MilliCPU, rr.MilliCPU, minResource) {
-		return false, "Insufficient cpu"
```
Thor-wl, line 421: Suggest not add any prefix for the resource for this is only a common used tool function. The prefix may narrow the usage.

pkg/scheduler/actions/allocate/allocate.go
Monokaix: Sorry for delay, I have added the related issus.

pkg/scheduler/api/resource_info.go
Monokaix: The origin prefix is reason, which is also not common, `LessEqual` func has many usage, it will cause a big change if wechange this function directly.

pkg/scheduler/api/resource_info.go
```diff
@@ -405,11 +405,12 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
...
 }
 
-// LessEqualWithReason returns true, "" only on condition that all dimensions of resources in r are less than or equal with that of rr,
-// Otherwise returns false and err string ,which show which resource is insufficient.
+// LessEqualWithResources returns true, []string{} only on condition that all dimensions of resources in r are less than or equal with that of rr,
+// Otherwise returns false and err string ,which show what resources are insufficient.
 // @param defaultValue "default value for resource dimension not defined in ScalarResources. Its value can only be one of 'Zero' and 'Infinity'"
 // this function is the same as LessEqual , and it will be merged to LessEqual in the future
-func (r *Resource) LessEqualWithReason(rr *Resource, defaultValue DimensionDefaultValue) (bool, string) {
+func (r *Resource) LessEqualWithResources(rr *Resource, defaultValue DimensionDefaultValue) (bool, []string) {
```
william-wang, line 412: It's better to change `LessEqualWithResources` to `LessEqualWithResourceName`

pkg/scheduler/api/node_info.go
```diff
@@ -372,30 +368,30 @@ func (ni *NodeInfo) setNode(node *v1.Node) {
...
 			ni.Used.Add(ti.Resreq)
 			ni.addResource(ti.Pod)
 		}
 	}
 }
 
-func (ni *NodeInfo) allocateIdleResource(ti *TaskInfo) error {
-	if ti.Resreq.LessEqual(ni.Idle, Zero) {
+func (ni *NodeInfo) allocateIdleResource(ti *TaskInfo) {
+	ok, resources := ti.Resreq.LessEqualWithResources(ni.Idle, Zero)
```
william-wang, line 386: `LessEqualWithResources` is similar with `lessEqual`  and will be merged to `lessEqual` in the future.  So here we'd better use `LessEqual`

pkg/scheduler/api/node_info.go
```diff
@@ -372,30 +368,30 @@ func (ni *NodeInfo) setNode(node *v1.Node) {
 	for _, ti := range ni.Tasks {
 		switch ti.Status {
 		case Releasing:
-			ni.Idle.sub(ti.Resreq) // sub without assertion
+			ni.allocateIdleResource(ti)
```
william-wang, line 371: sugguest use common function `sub`.

pkg/scheduler/api/node_info.go
```diff
@@ -372,30 +368,30 @@ func (ni *NodeInfo) setNode(node *v1.Node) {
...
 			ni.Releasing.Add(ti.Resreq)
 			ni.Used.Add(ti.Resreq)
 			ni.addResource(ti.Pod)
 		case Pipelined:
 			ni.Pipelined.Add(ti.Resreq)
 		default:
-			ni.Idle.sub(ti.Resreq) // sub without assertion
+			ni.allocateIdleResource(ti)
```
william-wang, line 378: same as above

pkg/scheduler/api/node_info.go
Monokaix: here we want to know what resources are not enough, so `LessEqualWithResources ` is used now.

william-wang (approved):
/lgtm

pkg/scheduler/api/node_info.go
```diff
@@ -291,12 +291,8 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
 	}
 
 	// set NodeState according to resources
-	if !ni.Used.LessEqual(ni.Allocatable, Zero) {
-		ni.State = NodeState{
-			Phase:  NotReady,
-			Reason: "OutOfSync",
-		}
-		return
+	if ok, resources := ni.Used.LessEqualWithResourcesName(ni.Allocatable, Zero); !ok {
```
XiangQinYu, line 294: Hello, Is the modification here necessary? Is it okay if we don't make any changes here

PR #3004, opened 2023-07-27
fix the reclaim bugs: push back the queue to queues in reclaim

### LINKED ISSUE #3003: Reclaim action push queue bugs, maybe cause the reclaiming reouces invalid
**What happened**:
Queue A shares some GPUs and CPUs resources with Queue B, and there are no resources for Queue A's jobs. However, the resources reclaiming became invalid. So I check the logs of volcano-scheduler, and the jobs that need resources don't go into the predicting node and evicting pods stage. There are some bugs that cause unexpected situations in the `reclaim.go`.

Reclaim happen normally.

The reason of this bug is there:
```go
// Found "high" priority task to reclaim others
if tasks, found := preemptorTasks[job.UID]; !found || tasks.Empty() {
	continue
} else {
	task = tasks.Pop().(*api.TaskInfo)
}

if !ssn.Allocatable(queue, task) {
	klog.V(3).Infof("Queue <%s> is overused when considering task <%s>, ignore it.", queue.Name, task.Name)
	continue
}

if err := ssn.PrePredicateFn(task); err != nil {
	klog.V(3).Infof("PrePredicate for task %s/%s failed for: %v", task.Namespace, task.Name, err)
	continue
}
```

When `ssn.Allocatable()` check false or the preemtor's tasks are empty, continue immediately.  The queue has multiple jobs the first job has no tasks and the others have preemptor tasks. The first job checking failed, and others' jobs are skipped.  This is how the bug happens.

**Environment**:
- Volcano Version: lasted version
- Kubernetes version (use `kubectl version`): v1.22.5-tke.8

RamezesDong:
fix #3003

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -104,6 +104,7 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			job = jobs.Pop().(*api.JobInfo)
 		}
 
+		queues.Push(queue)
 		// Found "high" priority task to reclaim others
 		if tasks, found := preemptorTasks[job.UID]; !found || tasks.Empty() {
 			continue
@@ -212,7 +213,6 @@ func (ra *Action) Execute(ssn *framework.Session) {
 		if assigned {
 			jobs.Push(job)
 		}
-		queues.Push(queue)
 	}
 }
 


lowang-bh:
Thanks very much for your contributions. It will more perfect if there are some UTs which will prevent others breaking the code in the features.

RamezesDong:
I'd be glad to add some UTs for this pr.

william-wang:
@wangyang0616 Please help to review as well.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -106,18 +106,21 @@ func (ra *Action) Execute(ssn *framework.Session) {
```
wangyang0616, line 106: Thank you for discovering and fixing this problem. Is it possible to add `queues.Push(queue)` on line 200 of the loop here?

There is similar processing in allocate.

RamezesDong: I will change the order of the code.

PR #3011, opened 2023-07-31
make adpative percentage nodes to find be consistent with kube-scheduler

Monokaix:
kube-scheduler perform tuning design: https://kubernetes.io/docs/concepts/scheduling-eviction/scheduler-perf-tuning/

This pr

1. keep main logic of percentage of nodes to find with kube-scheduler, and set `minimum-percentage-nodes-to-find` param default to `0` to make volcano set adaptivePercentage node num to find.
2. Add ut.

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -40,7 +40,7 @@ const (
 	// Default parameters to control the number of feasible nodes to find and score
 	defaultMinPercentageOfNodesToFind = 5
 	defaultMinNodesToFind             = 100
-	defaultPercentageOfNodesToFind    = 100
+	defaultPercentageOfNodesToFind    = 0
 	defaultLockObjectNamespace        = "volcano-system"
 )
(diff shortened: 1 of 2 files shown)


wangyang0616:
Does Volcano support need to support this mechanism [Default threshold](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduler-perf-tuning/#default-threshold)

Monokaix:
> Does Volcano support need to support this mechanism: [Default threshold](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduler-perf-tuning/#default-threshold).

Yes, `defaultPercentageOfNodesToFind` is explicitly set to `0` in this pr, so num of nodes to score will be calculated by a line formula, which behaves like kube-scheduler. The original `defaultPercentageOfNodesToFind` vaule is `100`, which will always traverse all nodes.

william-wang (approved):
/lgtm

PR #3015, opened 2023-08-01
doc: format and fix readme

jeffreyyjp:


jeffreyyjp:
@Thor-wl Formatting make it looks much better, and there are also some fixs, like `config/crd/bases (recommended)` -> `config/crd/volcano/bases (recommended)`, and some trailling space removing

Issue #3023, opened 2023-08-02
volcano support descheduler plugin


gj199575:
#### What would you like to be added:
Volcano is a good framework, I want to use Volcano as a engine to run descheduler plugin

#### Why is this needed:
this is a attempt to volcano, make volcano is more useful and helpufl 

this issue is done by https://github.com/volcano-sh/volcano/pull/3024

lowang-bh:
There is a similar feature: [Add dynamic scheduling and rescheduling functionality based on the usage of node #1777](https://github.com/volcano-sh/volcano/issues/1777)

PR #3024, opened 2023-08-02
use Volcano as a engine to run descheduler plugin

gj199575:
What would you like to be added:
Volcano is a good framework, I want to use Volcano as a engine to run descheduler plugin

Why is this needed:
this is a attempt to volcano, make volcano is more useful and helpufl

this PR  is written to solve https://github.com/volcano-sh/volcano/issues/3023


Thor-wl:
@gj199575 Hi, Thanks for your advice. But Volcano has already supported descheduling plugin. So I will close this PR.

Issue #3022, opened 2023-08-02
UT failed due to iterate over map has different sequence


lowang-bh:
`resource_info_test.go:1384`: deep equal should skip the sequence.
https://github.com/volcano-sh/volcano/blob/bb3bd543aaebbf6619383dfbb7d617034161b97a/pkg/scheduler/api/resource_info_test.go#L1381-L1391

**What happened**:

https://github.com/volcano-sh/volcano/actions/runs/5734040646/job/15539627192?pr=3020
```
--- FAIL: TestResource_LessEqualResource (0.00s)
    resource_info_test.go:1384: expected: []string{"cpu", "memory", "scalar.test/scalar1", "hugepages-test"}, got: []string{"cpu", "memory", "hugepages-test", "scalar.test/scalar1"}
FAIL
```

**What you expected to happen**:

pass UT

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master branch
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #3028, opened 2023-08-02
Fix the problem that resource info ut test fails probabilistically

wangyang0616:
fix : [#3022](https://github.com/volcano-sh/volcano/issues/3022)


lowang-bh (approved):
/lgtm

william-wang (approved):
/approve

Issue #2711, opened 2023-02-27
volcano support another GPU sharing scheme: mGPU(https://www.volcengine.com/docs/6419/145065)


fjding:
#### What would you like to be added:
Hi, In cloud-native environments, volcano has become the standard for batch scheduling, more and more cloud users use volcano as a batch scheduler in ByteDance Volcano Engine, But the GPU sharing scheme (mGPU) of Volcano Engine does not match the Volcano scheduler GPU scheme. We want to raise a PR to let the Volcano scheduler support ByteDance Volcano Engine mGPU scheduling. In this way, users of the volcano engine can directly use the mGPU scheduling capability after installing volcano scheduler.

mGPU Link: https://www.volcengine.com/docs/6419/145065

hwdef:
What work needs to be done to adapt mgpu?

Currently, Volcano's support for GPUs is already scalable. Please see if the current design can meet the demand

https://github.com/volcano-sh/volcano/pull/2643

fjding:
Thanks for you replay，the device api is well designed，I think this device api can satisfy the expansion of mgpu schedule. The implementation of mgpu is different from the nvida gpushare currently volcano supported, so I want to confirm two issues:
1. When will the official version of this device api be released?
2. If we extend the mgpu, we will create a directory named pkg/scheduler/api/devices/mgpu/gpushare. After our development is completed, can it be merged into the master branch of volcano and released as a GA version?

hwdef:
> Thanks for you replay，the device api is well designed，I think this device api can satisfy the expansion of mgpu schedule. The implementation of mgpu is different from the nvida gpushare currently volcano supported, so I want to confirm two issues:
> 
>     1. When will the official version of this device api be released?
> 
>     2. If we extend the mgpu, we will create a directory named pkg/scheduler/api/devices/mgpu/gpushare. After our development is completed, can it be merged into the master branch of volcano and released as a GA version?

1. If there is no accident, it will be three months after the release of version 1.7, that is, in April.
2. This requires the maintainer to make a decision.

fjding:
Can this issue be discussed at the weekly meeting?
Who can I contact for help?

hwdef:
@wangyang0616 
Can you offer any help?

hwdef:
could you please offer an email or wechat?

fjding:
> could you please offer an email or wechat?

Thanks，My wechat id：fj_ding

wangyang0616:
[...]

PR #3030, opened 2023-08-03
supprot mgpu feature for volcano

chenxi-seu:
relative issue
https://github.com/volcano-sh/volcano/issues/2711
https://github.com/volcano-sh/volcano/pull/3009

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module volcano.sh/volcano
 
-go 1.20
+go 1.19
 
 require (
 	github.com/agiledragon/gomonkey/v2 v2.1.0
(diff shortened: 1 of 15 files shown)


shinytang6 (commented):
l have some questions, just curious:
1. will be any follow up commits submitted to volcano device plugin later?
2. how should users use this feature locally without `mgpu` kernel module? l think it should only be available on bytedance volcano engine?

chenxi-seu:
> l have some questions, just curious:
> 
> 1. will be any follow up commits submitted to volcano device plugin later?
> 2. how should users use this feature locally without `mgpu` kernel module? l think it should only be available on bytedance volcano engine?

chenxi-seu:
@shinytang6 Sorry for late response. 
1. This PR will not be merge to master, just for user pull. There may be enhancements or bug fix updates in the future. 
2. Yes, currently only available on the bytedance volcano engine, as it relies on kernel support for `mgpu-driver`.

PR #3036, opened 2023-08-08
update image to v1.8.0

wangyang0616:
Update release-1.8 branch image version for release v1.8.0.

### DIFF
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -150,8 +150,8 @@ spec:
             - --port=8443
             - -v=4
             - 2>&1
-          image: volcanosh/vc-webhook-manager:latest
-          imagePullPolicy: Always
+          image: volcanosh/vc-webhook-manager:v1.8.0
+          imagePullPolicy: IfNotPresent
           name: admission
           volumeMounts:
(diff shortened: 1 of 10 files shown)


installer/volcano-development.yaml
```diff
@@ -185,8 +185,8 @@ spec:
       restartPolicy: Never
       containers:
         - name: main
-          image: volcanosh/vc-webhook-manager:latest
-          imagePullPolicy: Always
+          image: volcanosh/vc-webhook-manager:v1.8.0
+          imagePullPolicy: IfNotPresent
```
william-wang, line 189: What's the reason to change to `IfNotPresent`

wangyang0616: Set this parameter to Always in the master branch because the master branch uploads the latest image every day. In this way, the latest image is downloaded each time during installation.

The image of the release-1.8 branch is a fixed version of v1.8.0. Users do not need to download the image every time. Therefore, the IfNotPresent mode is used.

william-wang: It make sense.

william-wang (approved):
/lgtm

PR #3039, opened 2023-08-10
enhancement: copy cluster total resource from ssn, instead of summing up them again

lowang-bh:
Cluster's total resource has been summed up in session when open it.
Use ssn.TotalResource instead of re-calculating.

### DIFF
--- a/pkg/scheduler/plugins/overcommit/overcommit.go
+++ b/pkg/scheduler/plugins/overcommit/overcommit.go
@@ -40,6 +40,7 @@ const (
 type overcommitPlugin struct {
 	// Arguments given for the plugin
 	pluginArguments  framework.Arguments
+	totalResource    *api.Resource
 	idleResource     *api.Resource
 	inqueueResource  *api.Resource
 	overCommitFactor float64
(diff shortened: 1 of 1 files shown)


hwdef:
I'm sorry I don't know this part.

pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -81,13 +81,11 @@ func (op *overcommitPlugin) OnSessionOpen(ssn *framework.Session) {
...
 
 	// calculate idle resources of total cluster, overcommit resources included
-	total := api.EmptyResource()
 	used := api.EmptyResource()
 	for _, node := range ssn.Nodes {
-		total.Add(node.Allocatable)
 		used.Add(node.Used)
 	}
-	op.idleResource = total.Clone().Multi(op.overCommitFactor).Sub(used)
+	op.idleResource = ssn.TotalResource.Clone().Multi(op.overCommitFactor).Sub(used)
```
wangyang0616, line 88: It is suggested that the definition of TotalResource can be added in the overcommitPlugin structure, and assigned by ssn.TotalResource at the beginning of OnSessionOpen, see similar implementations of drf and proportion

lowang-bh: It sounds reasonale. I will update it.
By the way, is it need to also put the cluster's used at ssn so that all plugins don't need to calculate once again when use it.

lowang-bh:
Hi, @william-wang , how about this little improvement?

william-wang (approved):
/approve

Issue #3037, opened 2023-08-09
Unexpected Admission error while using vGPU with vc-jobs replicas>=2


archlitchi:
Step to reproduce:
1. Enable vGPU
2. submit the following yaml

```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: lm-mpi-job
spec:
  minAvailable: 3
  schedulerName: volcano
  plugins:
    ssh: []
    svc: []
  tasks:
    - replicas: 2
      name: mpimaster
      policies:
        - event: TaskCompleted
          action: CompleteJob
      template:
        spec:
          containers:
            - command:
                - /bin/sh
                - -c
                - |
                  MPI_HOST=`cat /etc/volcano/mpiworker.host | tr "\n" ","`;
                  mkdir -p /var/run/sshd; /usr/sbin/sshd;
                  sleep infinity;
              image: volcanosh/example-mpi:0.0.3
              name: mpimaster
              ports:
                - containerPort: 22
                  name: mpijob-port
              workingDir: /home
              resources:
                limits:
                  volcano.sh/vgpu-number: 1
                  volcano.sh/vgpu-memory: 1024
          restartPolicy: Always
    - replicas: 2
      name: mpiworker
      template:
        spec:
          containers:
            - command:
                - /bin/sh
                - -c
                - |
                  mkdir -p /var/run/sshd; /usr/sbin/sshd -D;
              image: volcanosh/example-mpi:0.0.3
              name: mpiworker
              ports:
                - containerPort: 22
                  name: mpijob-port
              workingDir: /home
          restartPolicy: OnFailure
---

```

You will see a pod failed with unexpectedAdmissionError there are enough GPU resources for it to fit.
![image](https://github.com/volcano-sh/volcano/assets/58680927/57e5da9a-0829-4d3c-b446-07a6860a8f7e)

ZDWWWWW:
Using vGPU in gang scheduling is a common scenario, and I believe this issue is worth discussing.

PR #3041, opened 2023-08-11
fix nodelock issue when using gang-scheduling 

archlitchi:
related issue: https://github.com/volcano-sh/volcano/issues/3037

We need to add 'bind' step to 'eventhandler' in order to lock the node after 'allocate' step, so nodelock won't block allocate from returning successfully. The jobReady func and gang-scheduling can work properly with vGPU feature with this PR


archlitchi:
#assign @wangyang0616 @william-wang

Issue #3042, opened 2023-08-11
Enhance usage plugin


gj199575:
#### What would you like to be added:
First optimize the usage plugin configuration of Volcano.
before that 

     - name: usage
       arguments:
          thresholds:
            CPUUsageAvg.5m: 80
            MEMUsageAvg.5m: 90

after that
  - name: usage
       arguments:
         usage.weight: 10
         type: average
         thresholds:
           cpu: 70
           mem: 70
         period: 10m

Second, make it easier for users to use the usage plugin
before
You must first define the rules of Pro, as follows

groups:
   - name: cpu_mem_usage_active
     interval: 30s
     rules:
     - record: mem_usage_active
       expr: 100*(1-node_memory_MemAvailable_bytes/node_memory_MemTotal_bytes)
   - name: cpu-usage-1m
     interval: 1m
     rules:
     - record: cpu_usage_avg_5m
       expr: 100 - (avg by (instance) (irate(node_cpu_seconds_total{mode="idle"}[5m])) * 100)
   - name: mem-usage-1m
     interval: 1m
     rules:
     - record: mem_usage_avg_5m
       expr: avg_over_time(mem_usage_active[5m])

after that
You don't need to define general rules, you can use the usage plugin directly

#### Why is this needed:
Make Volcano more user-friendly

william-wang:
Please add more description of your requirement :)

Monokaix:
fixed by https://github.com/volcano-sh/volcano/issues/3175

PR #3043, opened 2023-08-11
 Enhance usage plugin

gj199575:
solve #3042 

First , Before this PR, if you are using usage plugin and there is pods is blocked by usage plugin because the node NodeUtilization is not normal , and you restart volcano-scheduler ,the usage plugin is not worked.
and in this PR ， I fix this little problem .

Second optimization usage plugin

### DIFF
--- a/pkg/scheduler/plugins/usage/usage.go
+++ b/pkg/scheduler/plugins/usage/usage.go
@@ -18,14 +18,13 @@ package usage
 
 import (
 	"fmt"
-	"strconv"
-	"strings"
 
 	"k8s.io/klog/v2"
 	k8sFramework "k8s.io/kubernetes/pkg/scheduler/framework"
-
 	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/cache"
 	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/metrics/source"
 )
 
 const (
@@ -34,7 +33,6 @@ const (
 	cpuUsageAvgPrefix = "CPUUsageAvg."
 	memUsageAvgPrefix = "MEMUsageAvg."
 	thresholdSection  = "thresholds"
-	cpuUsageAvg5m     = "5m"
 )
 
 /*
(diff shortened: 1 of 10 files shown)


pkg/scheduler/plugins/usage/usage.go
```diff
@@ -43,35 +41,63 @@ const (
...
-          thresholds:
-            CPUUsageAvg.5m: 80
-            MEMUsageAvg.5m: 90
+         usage.weight: 10
+         type: average
+         thresholds:
+           cpu: 70
+           mem: 70
+         period: 10m
 */
```
lowang-bh, line 50: if I want to take both Avg.10m and Avg.30m into account, how to config it?

pkg/scheduler/plugins/usage/usage.go
```diff
@@ -99,26 +124,26 @@ func (up *usagePlugin) OnSessionOpen(ssn *framework.Session) {
...
+			case "mem":
+				up.memThresholds = float64(value)
 			}
-			if strings.Contains(key, memUsageAvgPrefix) {
-				periodKey := strings.Replace(key, memUsageAvgPrefix, "", 1)
-				up.threshold.memUsageAvg[periodKey] = val
+		}
+
+		// here is to deal with the case that restart volcano-scheduler and there is no ResourceUsage in cache of volcano-scheduler
+		if source.Period == "" {
```
lowang-bh, line 137: why period == "" means the case volcano restarted?

pkg/scheduler/plugins/usage/usage.go
```diff
@@ -99,26 +124,26 @@ func (up *usagePlugin) OnSessionOpen(ssn *framework.Session) {
...
+			source.Period = up.SamplePeriods
+			ssn.GetCache().GetMetricsData()
```
lowang-bh, line 139: I don't think we need to fetch cache data from scheduler-cache. All the cached data is snapshoted in open session and we should use the snapshot data.

pkg/scheduler/plugins/usage/usage.go
gj199575: 1 firstly
I  think both use AVG and COMMON is not good,so I do not support it
2 secondly 
I  think both use 5m and 10m  is not good,so I do not support it either

by the way
I make a new PR
https://github.com/volcano-sh/volcano/pull/3056

and this PR is closed

pkg/scheduler/plugins/usage/usage.go
gj199575: yes,period == "" means   volcano  is just starting

pkg/scheduler/plugins/usage/usage.go
gj199575: when volcano is just starting, there is not   read use data in snapshot

pkg/scheduler/plugins/usage/usage.go
lowang-bh: why?

lowang-bh: > why?

I know it now. period is a gloabal variable.

pkg/scheduler/plugins/usage/usage.go
lowang-bh: > I make a new PR

please update commit on the same PR. thanks.

PR #3044, opened 2023-08-12
doc: add local debug method for scheduler

wang-mask:
Doc: add the debug doc for volcano scheduler locally


wang-mask:
The method for debugging controllers is like the method of shcduler. The method for debugging webhook is a little complicated, you can refer to [this doc](https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/dev/how_to_develop.md#%E8%B0%83%E8%AF%95webhook%E7%BB%84%E4%BB%B6) .

lowang-bh:
I think this is a basic skill which developer should has. It has little relationship with volcano.

hwdef:
> I think this is a basic skill which developer should has. It has little relationship with volcano.
> 
> @wangyang0616 @hwdef please have a look. thanks.

Agree, but I'm not against merging this PR.

wang-mask:
> I think this is a basic skill which developer should has. It has little relationship with volcano.
> 
> @wangyang0616 @hwdef please have a look. thanks.

Although this is basic, it is important and necessary for beginners for volcano. And it is supplied by many open source projects.

Issue #2975, opened 2023-07-14
diffferent value of variable in returned err will cause event message too long


lowang-bh:
When one schedule cycle finishes, scheduler will sum up the unschedule reason and merge same reson. If there are different value of a variable in scheduler error, the unschedule event message will be too long.

**What happened**:
https://github.com/volcano-sh/volcano/blob/5302995a1d000df718fee5f3044df64bcb773d8d/pkg/scheduler/actions/allocate/allocate.go#L100-L118
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
@wangyang0616

lowang-bh:
Another little issue,  extender example has broken.

<img width="1015" alt="image" src="https://github.com/volcano-sh/volcano/assets/21003791/035b267a-e21a-4d5e-9ad4-5a6de08ab253">

Ash-KODES:
hey, I am interested in getting into the volcano project,
Can I look into this ?, I think this will help me to understand some flows of the project.

lowang-bh:
Of cause, welcome!

wangyang0616:
Yes, each filter node will execute all configured plugins, and finally summarize the results, which will have a certain performance impact compared to the previous implementation logic.

The current modification is mainly to solve the plugin's predicate compatibility with allocate and preempt functions. If there is a better solution, look forward to communicating with you.

PR #3051, opened 2023-08-13
fix: schedule message burst with nodes number increasion

### LINKED ISSUE #3049: failed to be scheduled pod's message  increased explosively with large number of nodes
**What happened**:

unschedulable info in pod condition is too large if there are many nodes.

image

kubectl get events |grep deploy-with-volcano
image

should sum up same fail to be scheduled reason among multiple nodes, like this:
image

1. add multiple nodes (virtual-kubelet, kwok)
2. create a deployment/vc-job scheduled by volcano
3. describe pod

**Environment**:
- Volcano Version: master branch
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

### LINKED ISSUE #3050: predicate does not stop if node is not suitable for current pod
caused by [Refactor PredicateFn for allocate and preempt actions #2916](https://github.com/volcano-sh/volcano/pull/2916)
Another issue https://github.com/volcano-sh/volcano/issues/3049 is also introduced by it.

**What happened**:

image

stop predicating if one filter retrun non-success.

The code login is changed when refact.

The following issues are relative.
#3049
#2975

**Environment**:
- Volcano Version: master branch
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
fix #3049, #2975, 
fix #3050 by second commit

Signed-off-by: lowang-bh <lhui_wang@163.com>

### DIFF
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -405,8 +405,7 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				task.Namespace, task.Name, node.Name)
 			podsNumStatus := &api.Status{
 				Code: api.Unschedulable,
-				Reason: fmt.Sprintf("Task <%s/%s> on Node <%s> failed, reason: %s",
-					task.Namespace, task.Name, node.Name, api.NodePodNumberExceeded),
+				Reason: api.NodePodNumberExceeded,
 			}
 			predicateStatus = append(predicateStatus, podsNumStatus)
 		}
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -419,11 +419,11 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 			predicateStatus := make([]*api.Status, 0)
 			status := nodeUnscheduleFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
 			nodeUnscheduleStatus, err := framework.ConvertPredicateStatus(status)
-			predicateStatus = append(predicateStatus, nodeUnscheduleStatus)
 			if err != nil {
 				return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeunschedulable.Name, status.Message())
 			}
 			if nodeUnscheduleStatus.Code != api.Success {
+				predicateStatus = append(predicateStatus, nodeUnscheduleStatus)
 				return predicateStatus, false, nil
 			}
 
@@ -431,11 +431,11 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 			if predicate.nodeAffinityEnable {
 				status := nodeAffinityFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
 				nodeAffinityStatus, err := framework.ConvertPredicateStatus(status)
-				predicateStatus = append(predicateStatus, nodeAffinityStatus)
 				if err != nil {
 					return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeaffinity.Name, status.Message())
 				}
 				if nodeAffinityStatus.Code != api.Success {
+					predicateStatus = append(predicateStatus, nodeAffinityStatus)
 					return predicateStatus, false, nil
 				}
 			}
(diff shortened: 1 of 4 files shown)


lowang-bh:
image

lowang-bh:
/priority/critical-urgent

hwdef (commented):
/lgtm
Thanks

lowang-bh:
commit 3: fix issue that: succeed predicated node has empty reason, and joined string will has empty string between ','

before fixed:
image

After fixed:
image

lowang-bh:
Hi, @hwdef I updated this pr. Please help to review. thanks.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -471,50 +483,65 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 		if predicate.nodePortEnable {
 			status := nodePortFilter.Filter(context.TODO(), state, nil, nodeInfo)
 			nodePortStatus, err := framework.ConvertPredicateStatus(status)
-			predicateStatus = append(predicateStatus, nodePortStatus)
 			if err != nil {
 				return predicateStatus, fmt.Errorf("plugin %s predicates failed %s", nodeports.Name, status.Message())
 			}
+			if nodePortStatus.Code != api.Success {
```
wangyang0616, line 489: I think there will be problems in judging the return of api.Success, such as the following scenarios:

When preempt filtering candidate nodes, the result returned by the nodeport plug-in filter is Unschedulable, and the result returned by podAffinity filter is UnschedulableAndUnresolvable. If you directly judge non-api.Success and return, the node will be mistakenly added to the candidate for preemption In the node list, but in fact the node does not meet the preemption conditions.

lowang-bh: What's the root cause? Both allocate/preemt/backfill has its own function to check the returned predicateStatus set. Preemt action does not check `Unschedulable` reson. 
https://github.com/volcano-sh/volcano/blob/258ad60df7ca6693e821c5445ba02f7033705db8/pkg/scheduler/actions/preempt/preempt.go#L211-L225

lowang-bh: @wangyang0616 I know what you mean, the root cause is that some predicate plugin does not need to run in preemption, right? There are two solutions:
1. keep the origin changes: put all predicate results to finial results, and then filter out those reason=="" when joined them, but the continue on same onde with other predicate plugin has some performance effect.
2. add adtional plugin enable switch which independent from each actions,  to controll each action should use which plugins.

wangyang0616: For resource quota plug-ins, the predicate will return Unschedulable status, indicating that the current node does not allow allcate, but can preempt;

For plug-ins with inherent types of nodes (such as taint, pro-core), UnschedulableAndUnresolved is returned, indicating that neither allocate nor preempt is possible.

wangyang0616: I think plan 1 will be better. The action part formulates predicate rules, and each plug-in realizes its own capabilities according to the rules. It is best for the action to be unaware of the capabilities of each plug-in.

wangyang0616: I looked at the processing of preempt. During the preemption process, pods on the node will continue to be evicted until certain conditions are met, and the eviction will stop.

The condition for stopping eviction is that extended resources such as cpu, memory, and gpu meet the requirements of the preemptor, and then stop eviction. In fact, there is no determination of the corresponding affinity, antiAffinity, topologyspread and other conditions, so the preempt will think that the node has met the scheduling conditions of the preemptor, but the actual allocation scheduling will still be intercepted by plugins such as affinity and antiAffinity, resulting in preemption failure.

https://github.com/volcano-sh/volcano/blob/36abf1b2037516de2129c69791df486018b46ae6/pkg/scheduler/actions/preempt/preempt.go#L274-L279

For example issue: [#3068](https://github.com/volcano-sh/volcano/issues/3068) is the problem.

wangyang0616: > I think there will be problems in judging the return of api.Success, such as the following scenarios:
> 
> When preempt filtering candidate nodes, the result returned by the nodeport plug-in filter is Unschedulable, and the result returned by podAffinity filter is UnschedulableAndUnresolvable. If you directly judge non-api.Success and return, the node will be mistakenly added to the candidate for preemption In the node list, but in fact the node does not meet the preemption conditions.

Preemption should be rejected in this scenario. Currently, the preemption function of this field is not supported. Preemption is only effective for resource usage.

wangyang0616:
I agree with your current modification plan.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -471,50 +483,65 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
 		if predicate.podAffinityEnable {
 			status := podAffinityFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
 			podAffinityStatus, err := framework.ConvertPredicateStatus(status)
-			predicateStatus = append(predicateStatus, podAffinityStatus)
 			if err != nil {
 				return predicateStatus, fmt.Errorf("plugin %s predicates failed %s", interpodaffinity.Name, status.Message())
 			}
+			if podAffinityStatus.Code != api.Success {
+				predicateStatus = append(predicateStatus, podAffinityStatus)
+				return predicateStatus, nil
```
wangyang0616, line 504: According to the analysis of issue: [#3068](https://github.com/volcano-sh/volcano/issues/3068), I think it is better to return error here. The reason is as described in the comments. Currently, Volcano only supports resource preemption, and does not support algorithm preemption functions such as affinity topology.

If it is convenient, you can update it in your pr, or I can mention another pr to fix it

wangyang0616:
Please review it. Thanks. @william-wang

william-wang (approved):
/lgtm

PR #3052, opened 2023-08-13
fix: enhanced FitError of JobInfo by logging the insufficient resources

### LINKED ISSUE #2993: Optimize log output when resources are not enough
Optimize the scheduler's logs when resources are low by printing exactly what kind of resources are low.
#### Why is this needed:
easier to identify the reasons for insufficient resources.

#### kubernetes scheduler VS volcano
kubernetese scheduler:
image

volcano:
image

volcano just print `NotEnoughResources`

srikanth-iyengar:
fix #2993

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -610,7 +610,7 @@ func (ji JobInfo) String() string {
 
 // FitError returns detailed information on why a job's task failed to fit on
 // each available node
-func (ji *JobInfo) FitError() string {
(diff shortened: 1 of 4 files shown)


lowang-bh:
Hi, @srikanth-iyengar , thanks for your hard work on this issue.  Actually, the filter result is already exist and don't need to do any calculations. You can find out the first failed scheduling pod's message and append to the podgroup's message. Just like this highlight in picture:
image

srikanth-iyengar:
Something like this @lowang-bh 
image

PR #3054, opened 2023-08-14
Add nodelockEnable parameters default to false

archlitchi:
add nodelockEnable paremeters,default to false, in order to solve issue https://github.com/volcano-sh/volcano/issues/3037

Note that disable nodelock may result in config mismatch


william-wang (approved):
/lgtm

PR #3062, opened 2023-08-17
Jobinfo enhancement

### LINKED ISSUE #3061: job's allocated does not need to re-calculate
jobinfo has an attribute about allocated resource, so it does not need to calculate again.

https://github.com/volcano-sh/volcano/blob/54e3b40ae8c7e69d6eb899dd9f9ce684d27d2b4f/pkg/scheduler/plugins/proportion/proportion.go#L151-L157

```go
// allocated := util.GetAllocatedResource(job) <--remove this line
inqueued := util.GetInqueueResource(job, job.Allocated)
```

#### Why is this needed:
1. job.Allocated represent its allocated resource.

lowang-bh:
commit 1: get job min resource function: reduce function call to only once.
commit 2: fix #3061 (job's allocated does not need to re-calculate)

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -485,10 +485,11 @@ func (ji *JobInfo) GetMinResources() *Resource {
 }
 
 func (ji *JobInfo) GetElasticResources() *Resource {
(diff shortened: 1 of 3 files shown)


william-wang (approved):
/approve

Issue #3029, opened 2023-08-03
How to know the status of the task during volcanlo scheduling?


CharlesQQ:
#### What would you like to be added:
Add status of task in job.status , so user can know  the pod scheduling status.

#### Why is this needed:
User want know whether pod in what state, like `Allocated`、`Pipelined` and so on.

lowang-bh:
There is some difficulty because the state is reset at each scheduling cycle.

daniel-hutao:
![image](https://github.com/volcano-sh/volcano/assets/18692628/cf0fd038-898b-4b35-bb0f-90c14a3b344b)

lowang-bh:
relative issue: [Display all jobs and queuing order in the queue #2876](https://github.com/volcano-sh/volcano/issues/2876)

lowang-bh:
@daniel-hutao `job.Status.TaskStatusCount` is already there and it is about the final status of pod(task).

CharlesQQ:
> job.Status.TaskStatusCount is already there and it is about the final status of pod(task)

indeed, job.Status.TaskStatusCount is the final status of pod, I think scheduling status should be add to pod's label or Annotation, Just motify function `UpdateTaskStatus` can achieve that
https://github.com/volcano-sh/volcano/blob/04a3b1b6b6b3550e292c7d3132fc1fbf3f6932cd/pkg/scheduler/api/job_info.go#L513-L529

lowang-bh:
> I think scheduling status should be add to pod's label or Annotation, Just motify function `UpdateTaskStatus` can achieve that,

This will cause low performance of scheduler.  Because in each scheduling cycle the task's pipelined/binded node is changed randomly and frequently calling apiserver brings low performance and high latancey.

PR #3063, opened 2023-08-18
api(volcano): add task status to pod

CharlesQQ:
resolve: https://github.com/volcano-sh/volcano/issues/3029

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -17,6 +17,7 @@ limitations under the License.
 package api
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


lowang-bh:
/hold

It's too complex

CharlesQQ:
> It's too complex

can you describe  it more clearly?

lowang-bh:
There are three points:
1. change `NewJobInfo` function is a bad case, you'd better keep it as origin, and put the pod status updating where job is comitted or discard.
2. apiserver calling frequency will cause cluster presure. You'd better to keep this feature close as default.
3. patch oparation can used to do pod updating.

Thor-wl:
> I am recommending event recording and end user can consume those events and do what they want.

+1

Issue #2879, opened 2023-05-30
task scheduling latency metrics is not accurate


lowang-bh:
Have fix by #2866, just record issue.

**What happened**:

This Bind func never return error, so `UpdateTaskScheduleDuration` will always record all pods' successful binding time as now. But those pods failed to bind should not update metrics to now, due to they will be resynced and bounden in next cycle.

https://github.com/volcano-sh/volcano/blob/6b84f4b0f8a653d826ad7446d027ff12829efaa7/pkg/scheduler/cache/cache.go#L1052-L1060
**What you expected to happen**:

only handle metrics about successfully bind pods

**How to reproduce it (as minimally and precisely as possible)**:

code logic error

**Anything else we need to know?**:

Fix by #2866 : move metrics recording to where pod is bound 

**Environment**:
- Volcano Version: master branch
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
Fix by #2866, record metrics and events as soon as pod is bounden

PR #3064, opened 2023-08-18
fix: task scheduling latancy metrics is not accurate

lowang-bh:
if one task bind failed, all the successfully binded pods' metrics will not be record utill they are bind successfully in future scheduling cycle. This will cause latency and not accurate.

fix https://github.com/volcano-sh/volcano/issues/2879

This PR is split from [fix: events miss when there is a pod bind failed #2866](https://github.com/volcano-sh/volcano/pull/2866). Please merge this PR first.

key points:
1: replcace all clientset with interface
2: move metrics recoding to where pod is successfully bounden


hwdef (commented):
/lgtm

william-wang (approved):
/lgtm

Issue #3068, opened 2023-08-25
Preemption conflicts with the anti-affinity


wangyang0616:
**What happened**:
The pod is set with a high priority and an anti-affinity policy is set. When the cluster resources are insufficient, preemption is triggered, and the pod is assigned to the node that violates the anti-affinity constraint. 

The next round of scheduling does not meet the anti-affinity policy, and allocate cannot Resources are allocated to pods, resulting in high-priority pods being stuck all the time, and preemptive scheduling fails.
**What you expected to happen**:
Resource preemptive scheduling succeeded.

**How to reproduce it (as minimally and precisely as possible)**:

![image](https://github.com/volcano-sh/volcano/assets/36185952/d6896ee0-6e27-4e11-ae09-5ec86e7d7b29)

```
kubectl label nodes gotc-worker gotc-worker2 gotc-worker3 internal.test.topology/zone=bj
kubectl label nodes gotc-worker4 internal.test.topology/zone=sh
```

deploy1.yaml
```
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: low-priority
value: 10
globalDefault: false
description: "This priority class should be used for volcano job only."
---

apiVersion: apps/v1
kind: Deployment
metadata:
  name: label-sh
  labels:
    app: label-sh
spec:
  replicas: 3
  selector:
    matchLabels:
      app: label-sh
  template:
    metadata:
      labels:
        app: label-sh
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: internal.test.topology/zone
                operator: In
                values:
                - sh
      priorityClassName: low-priority
      schedulerName: volcano
      containers:
      - name: test
        image: busybox
        imagePullPolicy: IfNotPresent
        command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep 3600']
        resources:
          requests:
            cpu: 1
          limits:
            cpu: 1
```
deploy2.yaml
```
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 100
globalDefault: false
description: "This priority class should be used for volcano job only."

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: anti-test
  labels:
    app: anti-test
spec:
  replicas: 2
  selector:
    matchLabels:
      app: anti-test
  template:
    metadata:
[...]

PR #3070, opened 2023-08-26
Fix the conflict between preemption and antiAffinity

wangyang0616:
1. fix: [#3068](https://github.com/volcano-sh/volcano/issues/3068)
2. Partial optimization based on [#3051](https://github.com/volcano-sh/volcano/pull/3051), Currently, preemption only considers extended resources such as cpu, memory, and gpu, and does not consider strategies such as affinity, topology, etc. When the predicate inherits the k8s filtering algorithm, if the return value is not Success, it returns an error, neither allcate nor preempt

**Test Results：**
The high-priority job anti-test preempts the resources of the low-priority job label-sh and runs successfully.
image

image

image

**Pending pod event:**
image

**Pending pod msg:**
image

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -126,15 +126,15 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			var statusSets util.StatusSets
 			statusSets, err := ssn.PredicateFn(task, n)
 			if err != nil {
-				klog.V(3).Infof("reclaim predicates failed for task <%s/%s> on node <%s>: %v",
+				klog.V(5).Infof("reclaim predicates failed for task <%s/%s> on node <%s>: %v",
 					task.Namespace, task.Name, n.Name, err)
 				continue
 			}
 
 			// Allows scheduling to nodes that are in Success or Unschedulable state after filtering by predicate.
 			if statusSets.ContainsUnschedulableAndUnresolvable() || statusSets.ContainsErrorSkipOrWait() {
-				klog.V(3).Infof("predicates failed in reclaim for task <%s/%s> on node <%s>, status is not success or unschedulable.",
-					task.Namespace, task.Name, n.Name)
+				klog.V(5).Infof("predicates failed in reclaim for task <%s/%s> on node <%s>, reason is %s.",
+					task.Namespace, task.Name, n.Name, statusSets.Message())
 				continue
 			}
 			klog.V(3).Infof("Considering Task <%s/%s> on Node <%s>.",
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -407,49 +407,44 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 			klog.V(4).Infof("NodePodNumber predicates Task <%s/%s> on Node <%s> failed",
 				task.Namespace, task.Name, node.Name)
 			podsNumStatus := &api.Status{
-				Code:   api.Unschedulable,
+				// TODO(wangyang0616): When the number of pods of a node reaches the upper limit, preemption is not supported for now.
+				// Record details in #3079 (volcano.sh/volcano)
+				// In the preempt stage, the pipeline of the pod number is not considered,
+				// the preemption of the pod number is released directly, which will cause the pods in the node to be cyclically evicted.
+				Code:   api.UnschedulableAndUnresolvable,
 				Reason: api.NodePodNumberExceeded,
 			}
 			predicateStatus = append(predicateStatus, podsNumStatus)
-			return predicateStatus, nil
+			return predicateStatus, fmt.Errorf("%s", api.NodePodNumberExceeded)
 		}
 
 		predicateByStablefilter := func(pod *v1.Pod, nodeInfo *k8sframework.NodeInfo) ([]*api.Status, bool, error) {
 			// CheckNodeUnschedulable
 			predicateStatus := make([]*api.Status, 0)
 			status := nodeUnscheduleFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
-			nodeUnscheduleStatus, err := framework.ConvertPredicateStatus(status)
-			if err != nil {
-				return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeunschedulable.Name, status.Message())
-			}
+			nodeUnscheduleStatus := framework.ConvertPredicateStatus(status)
 			if nodeUnscheduleStatus.Code != api.Success {
 				predicateStatus = append(predicateStatus, nodeUnscheduleStatus)
-				return predicateStatus, false, nil
+				return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeUnscheduleFilter.Name(), status.Message())
 			}
 
 			// Check NodeAffinity
 			if predicate.nodeAffinityEnable {
 				status := nodeAffinityFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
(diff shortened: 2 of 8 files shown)


pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -411,45 +411,35 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				Reason: api.NodePodNumberExceeded,
 			}
 			predicateStatus = append(predicateStatus, podsNumStatus)
-			return predicateStatus, nil
```
lowang-bh, line 414: If we don't return, please filter out those reason is null string at 
https://github.com/volcano-sh/volcano/blob/c91eb07f2c300e4d5c826ff11a63b91781b3ac11/pkg/scheduler/util/predicate_helper.go#L156-L158

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -411,45 +411,35 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
 			status := nodeUnscheduleFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
-			nodeUnscheduleStatus, err := framework.ConvertPredicateStatus(status)
-			if err != nil {
-				return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeunschedulable.Name, status.Message())
-			}
+			nodeUnscheduleStatus := framework.ConvertPredicateStatus(status)
 			if nodeUnscheduleStatus.Code != api.Success {
 				predicateStatus = append(predicateStatus, nodeUnscheduleStatus)
-				return predicateStatus, false, nil
+				return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeUnscheduleFilter.Name(), status.Message())
```
lowang-bh, line 423: Note that if error is return, those code will not hit:
https://github.com/volcano-sh/volcano/blob/c91eb07f2c300e4d5c826ff11a63b91781b3ac11/pkg/scheduler/actions/allocate/allocate.go#L110-L113

https://github.com/volcano-sh/volcano/blob/c91eb07f2c300e4d5c826ff11a63b91781b3ac11/pkg/scheduler/actions/preempt/preempt.go#L220-L223

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -411,45 +411,35 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
 				status := nodeAffinityFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
-				nodeAffinityStatus, err := framework.ConvertPredicateStatus(status)
-				if err != nil {
-					return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeaffinity.Name, status.Message())
-				}
+				nodeAffinityStatus := framework.ConvertPredicateStatus(status)
 				if nodeAffinityStatus.Code != api.Success {
 					predicateStatus = append(predicateStatus, nodeAffinityStatus)
-					return predicateStatus, false, nil
+					return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeAffinityFilter.Name(), status.Message())
```
lowang-bh, line 432: same as before

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -482,65 +472,50 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
 			status := nodePortFilter.Filter(context.TODO(), state, nil, nodeInfo)
-			nodePortStatus, err := framework.ConvertPredicateStatus(status)
-			if err != nil {
-				return predicateStatus, fmt.Errorf("plugin %s predicates failed %s", nodeports.Name, status.Message())
-			}
+			nodePortStatus := framework.ConvertPredicateStatus(status)
 			if nodePortStatus.Code != api.Success {
 				predicateStatus = append(predicateStatus, nodePortStatus)
-				return predicateStatus, nil
+				return predicateStatus, fmt.Errorf("plugin %s predicates failed %s", nodePortFilter.Name(), status.Message())
```
lowang-bh, line 478: same as before.

lowang-bh:
Thanks for your hard work. BTW, Is there  any test result about this PR, not only including preemtation works well, but also the pod/podgroup unscheduling message in their status.

lowang-bh:
And there is another enhancement can be used in preemption action: https://github.com/volcano-sh/volcano/issues/3071

pkg/scheduler/plugins/predicates/predicates.go
wangyang0616: - The filtering of the pod number in the predicate plugin will go to this part of the code logic
- The gpu-related resource filtering in device FilterNode will go to this part of the code logic

For strategies such as nodeaffinity, podaffinity, and nodeport, an error is returned directly, and the preemption action is not currently supported

wangyang0616:
> Thanks for your hard work. BTW, Is there any test result about this PR, not only including preemtation works well, but also the pod/podgroup unscheduling message in their status.

I will add the test results later.

pkg/scheduler/plugins/predicates/predicates.go
wangyang0616: done

pkg/scheduler/plugins/predicates/predicates.go
wangyang0616: done

pkg/scheduler/plugins/predicates/predicates.go
wangyang0616: done

wangyang0616:
> And there is another enhancement can be used in preemption action: #3071
> 
> What's more, I think the current preemption action can not work well. In my opinion, it should following those steps:
> 
> 1. filter out those nodes which can not help preemption, according to the filter result stored in job's NodesFitErrors
> 2. range from those nodes in step-1, consider the victims from a node and remove the victims from this node, and then re-run predications on the node; if result is success, then the node is a candidate node, put them in final node list;
> 3. score on thoes candinate nodes, to choose a best one.

I very much agree with your proposal, the preemption function still has a lot to optimize, we can iterate gradually.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -126,15 +126,15 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			var statusSets util.StatusSets
 			statusSets, err := ssn.PredicateFn(task, n)
 			if err != nil {
-				klog.V(3).Infof("reclaim predicates failed for task <%s/%s> on node <%s>: %v",
+				klog.V(5).Infof("reclaim predicates failed for task <%s/%s> on node <%s>: %v",
```
william-wang, line 129: What's the reason of change level 3 to level 5？

wangyang0616: When a large number of pending pods in the cluster perform the reclaim operation and resource reclamation fails, a large number of repeated logs are generated in each round of scheduling. I think it is more appropriate to change the log level to debug.

william-wang: ok

william-wang (approved):
/lgtm

lowang-bh:
> What's more, I think the current preemption action can not work well. In my opinion, it should following those steps:
> 
> 1. filter out those nodes which can not help preemption, according to the filter result stored in job's NodesFitErrors
> 2. range from those nodes in step-1, consider the victims from a node and remove the victims from this node, and then re-run predications on the node; if result is success, then the node is a candidate node, put them in final node list;
> 3. score on thoes candinate nodes, to choose a best one.

I have open an enhancement requirement to trace it. [enhancement for preemption action #3074](https://github.com/volcano-sh/volcano/issues/3074)

Issue #3072, opened 2023-08-26
updating pod status should set seperate kube client and qps


Monokaix:
**What happened**:
when lots of pods pending and one node scaled，the pod status message will be updated when one node scaled，which harm to apiserver and etcd
eg.
before node scaled: pod.status.mesage= "0/10 nodes are available,xxx"
afetr node scaled: pod.status.mesage= "0/11 nodes are available,xxx"
every pod's status will be updated, which is a heavy pressure to apiserver and etcd in a flash.
**What you expected to happen**:
limit update pod status' qos 
**How to reproduce it (as minimally and precisely as possible)**:

1. run a k8s cluster with 2 nodes
2. create 8000 pending pod
3. scale node num=3
4. we can see that etcd db size grow instantly and may exceed etcd's max db size.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: latest
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #3073, opened 2023-08-26
set seperate client for updating pod status

Monokaix:
fix https://github.com/volcano-sh/volcano/issues/3072


lowang-bh:
I think we need to find out why the pod status message is updated in every session. Actually, it does not need to do update if the pod's condition has no change.

Monokaix:
> I think we need to find out why the pod status message is updated in every session. Actually, it does not need to do update if the pod's condition has no change.
> 
> The qps limiit can not solve the root cause and will cause scheduling latency or update conflicts(if update is in different goroutine in different session)

details have been added to issue.
A seperate client is crated and it only takes effect in pod/status api.

pkg/scheduler/cache/cache.go
```diff
@@ -232,6 +232,33 @@ type defaultStatusUpdater struct {
 	vcclient   *vcclient.Clientset
 }
 
+func NewPodStatusClient(config *rest.Config) (*kubernetes.Clientset, error) {
+	configShallowCopy := *config
+	qps := os.Getenv("POD_STATUS_CLIENT_QPS")
+	if qps != "" {
```
lowang-bh, line 238: how about qps is not set in env? what's default qps? use kubeclient's default one?

pkg/scheduler/cache/cache.go
```diff
@@ -232,6 +232,33 @@ type defaultStatusUpdater struct {
...
+	burst := os.Getenv("POD_STATUS_CLIENT_BURST")
+	if burst != "" {
+		burstInt, err := strconv.Atoi(burst)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse burst: %v", err)
+		}
+		configShallowCopy.Burst = burstInt
+	}
+
+	client, err := kubernetes.NewForConfig(&configShallowCopy)
```
lowang-bh, line 255: Here, we can use `AddUserAgent` to specify an identity to distinguish from the default client.

pkg/scheduler/cache/cache.go
Monokaix: if not set env, we will use `config`, which is configed by cmd flag param and consistent with other kube client.

Issue #3080, opened 2023-08-27
The podgroup will update the status even if there is no change


wangyang0616:
**What happened**:
If the podgroup does not change, the status will be updated every 1 minute or so, causing unnecessary pressure on the apiserver.

![image](https://github.com/volcano-sh/volcano/assets/36185952/840cc0a9-4bcd-440f-acdc-7a7c89f0d239)

![image](https://github.com/volcano-sh/volcano/assets/36185952/745b9543-b24e-4e9d-b258-cdde3bd6e97d)

**What you expected to happen**:
When the podgroup changes, update it in time. If the podgroup does not change, there is no need to update the status information.

**How to reproduce it (as minimally and precisely as possible)**:
vcjob-test1.yaml
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: test
spec:
  schedulerName: volcano
  minAvailable: 3
  policies:
    - event: PodEvicted
      action: RestartJob
  tasks:
    - replicas: 1
      name: "ps"
      template:
        spec:
          containers:
            - image: alpine
              command: ["/bin/sh", "-c", "sleep 1000"]
              imagePullPolicy: IfNotPresent
              name: ps
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure
    - replicas: 2
      name: "worker"
      policies:
        - event: PodFailed
          action: RestartTask
      template:
        spec:
          restartPolicy: Never
          containers:
            - image: nginx
              command: ["/bin/sh", "-c", "sleep 1000"]
              imagePullPolicy: IfNotPresent
              name: worker
              resources:
                requests:
                  cpu: "1"
```
Create a workload through `vcjob-test.yaml`, find the corresponding podgroup through `kubectl get pg`, and check the status information.

**Anything else we need to know?**:
https://github.com/volcano-sh/volcano/blob/c91eb07f2c300e4d5c826ff11a63b91781b3ac11/pkg/scheduler/framework/job_updater.go#L64-L66

Just remove the judgment here.

**Environment**: k8s v1.25
- Volcano Version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #3081, opened 2023-08-27
The status of a pod group is updated only when the pod group changes.

wangyang0616:
fix: [#3080](https://github.com/volcano-sh/volcano/issues/3080)

### DIFF
--- a/pkg/scheduler/framework/job_updater.go
+++ b/pkg/scheduler/framework/job_updater.go
@@ -15,9 +15,6 @@ import (
 
 const (
 	jobUpdaterWorker = 16
-
-	jobConditionUpdateTime       = time.Minute
-	jobConditionUpdateTimeJitter = 30 * time.Second
 )
 
 // TimeJitterAfter means: new after old + duration + jitter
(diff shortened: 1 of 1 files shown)


pkg/scheduler/framework/job_updater.go
```diff
@@ -61,9 +58,6 @@ func isPodGroupConditionsUpdated(newCondition, oldCondition []scheduling.PodGrou
 
 		newTime := newCond.LastTransitionTime
 		oldTime := oldCond.LastTransitionTime
-		if TimeJitterAfter(newTime.Time, oldTime.Time, jobConditionUpdateTime, jobConditionUpdateTimeJitter) {
```
william-wang, line 64: What's the backgroud of update podgroup in a period of time?

wangyang0616: This item is added to reduce the number of access times of the API server. However, this item does not take effect.

If the conditions are inconsistent and the duration exceeds one minute, update the API server.

I have verified that the pod group status does not change when the pod group is in the running state. When the pod group is in the pending or inqueue state, the pod group status does not change frequently. Therefore, I think the 1-minute interval can be deleted.

william-wang (approved):
/lgtm

PR #3092, opened 2023-09-01
Add reSync task callback

Monokaix:
Related issus: https://github.com/volcano-sh/volcano/issues/3096

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -33,6 +33,7 @@ import (
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	"volcano.sh/apis/pkg/apis/scheduling"
 	"volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+
 	volumescheduling "volcano.sh/volcano/pkg/scheduler/capabilities/volumebinding"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -961,8 +961,18 @@ func (sc *SchedulerCache) processResyncTask() {
 		return
 	}
 
+	reSync := false
 	if err := sc.syncTask(task); err != nil {
 		klog.Errorf("Failed to sync pod <%v/%v>, retry it.", task.Namespace, task.Name)
+		reSync = true
+	}
+	if task.ReSyncTask != nil {
```
lowang-bh, line 969: How about syncTask return no error? Looks it will always run ReSyncTask callback.

Monokaix: yes, `ReSyncTask callback` will always retry and run until it return no err.

Monokaix: I think what we need care is `syncTask` return err but `ReSyncTask ` return no err, then `ReSyncTask` will still run every time, we can add a flag to indicate whether `ReSyncTask` is finished.

lowang-bh: Which scenario will this callback be used in?

Monokaix: just like what I said in description, when using custom bind method, some alarm or rollback are needed when bind err.

pkg/scheduler/api/job_info.go
```diff
@@ -130,6 +131,10 @@ type TaskInfo struct {
 	NumaInfo   *TopologyInfo
 	PodVolumes *volumescheduling.PodVolumes
 	Pod        *v1.Pod
+
+	// CustomBindErrHandler is a custom callback func called when task bind err.
+	CustomBindErrHandler         func() error
+	CustomBindErrHandlerFinished bool
```
william-wang, line 137: please change `CustomBindErrHandlerFinished` to a better name.

Monokaix: done for that.

william-wang (approved):
/approve

PR #3097, opened 2023-09-04
fix tdm time parse

xiaoanyunfei:
tdm can check case like `23:00-01:00`


lowang-bh:
What is the problem without this PR?

Issue #3105, opened 2023-09-07
scheduler panic when calculate the elastic resource


flyingfang:
I encountered a case : volcano scheduler panic, when calculating elastic resources.
The key code is as follows：
```go
func (ji *JobInfo) GetElasticResources() *Resource {
	if ji.Allocated.LessEqualPartly(ji.GetMinResources(), Zero) {
		return EmptyResource()
	}
	return ji.Allocated.Clone().Sub(ji.GetMinResources())
}
```

The scheduler would occur panic, when a resource in minResource does not exist in Allocated, and the remaining resources in allocated are greater than the value of the corresponding resource in minResource.
For example:
AllocatedResource
```go
Resource{
MilliCPU: 50000.00, 
Memory: 1024000.00,
}
```
minResource
```go
Resource{
MilliCPU: 40000.00,
Memory: 1000000.00,
ScalarResources: map[v1.ResourceName]float64{"nvidia.com/gpu": 1000.00},
}
```

PR #3106, opened 2023-09-08
fix panic when get job's elastic resource

lowang-bh:
Fix: #3105 (scheduler panic when calculate the elastic resource) via first commit
Fix: #3127 (panic when guarantee greater than allocatable resource) via second commit

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -574,30 +574,11 @@ func (ji *JobInfo) DeductSchGatedResources(res *Resource) *Resource {
 
 // GetElasticResources returns those partly resources in allocated which are more than its minResource
 func (ji *JobInfo) GetElasticResources() *Resource {
-	elastic := EmptyResource()
 	if ji.Allocated == nil {
-		return elastic
+		return EmptyResource()
 	}
 	minResource := ji.GetMinResources()
-	if ji.Allocated.MilliCPU > minResource.MilliCPU {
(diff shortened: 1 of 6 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -572,12 +572,15 @@ func (ji *JobInfo) DeductSchGatedResources(res *Resource) *Resource {
...
+// GetElasticResources returns those partly resources in allocated which are more than its minResource
 func (ji *JobInfo) GetElasticResources() *Resource {
-	minResource := ji.GetMinResources()
-	if ji.Allocated.LessEqualPartly(minResource, Zero) {
+	if ji.Allocated == nil {
 		return EmptyResource()
 	}
-	return ji.Allocated.Clone().Sub(minResource)
+	minResource := ji.GetMinResources()
+	elastic := ExceedPart(ji.Allocated, minResource)
```
Monokaix, line 581: exceed -> exceeded is better?

hwdef:
Does the capacity plugin need to be handled?

lowang-bh:
> Does the capacity plugin need to be handled?

Have added to it.

pkg/scheduler/api/job_info.go
lowang-bh: Done

Monokaix:
cc @hwdef

hwdef (commented):
/lgtm

Issue #2322, opened 2022-07-01
dynamic log level adjustment without restarting volcano component


elinx:
#### What would you like to be added:
change log level dynamicly without restart volcano component

#### Why is this needed:
When some glitches happends online and need to figure out the cause, one option is to change the log level and reproduce the issue to get more verbosed log. The log level threshold is hard coded into deployment yaml file right now, the volcano component like scheduler will restart when the log level changed and the issue may can't reporduce anymore.

william-wang:
Yes, it's what we have been wanting to do.  refer to #1326 , Let's track this issue here.

elinx:
This feature highly depends on the feature of `klog` to change the verbosity of the log level on the fly, but in the current implementation `klog` does not provide any configurable API and the maintainer looks very hesitant about adding a new API:

kubernetes/klog#336
kubernetes/klog#340

wangyang0616:
Retain the current issue.
/remove-lifecycle stale

xiao-jay:
@wangyang0616 
looks like modify the volcano-scheduler-configmap will loadSchedulerConf
https://github.com/volcano-sh/volcano/blob/17370d7d6c3c992ef47eba77da45b5b24bc6215a/pkg/scheduler/scheduler.go#L181-L205

So May be can add klogv in volcano-scheduler-configmap.If you want to modify klogv while Valcano is running, you can add/modify klogv=5 in the volcano scheduler configmap,and run next code to modify klog v.
```
var fs flag.FlagSet
klog.InitFlags(&fs)
fs.Set("v", "klogv")
```

PR #3111, opened 2023-09-11
add dynamic conf design docs

xiao-jay:
issue #2322 
pr #3146


docs/design/dynamic-conf.md
```diff
@@ -0,0 +1,35 @@
...
+[![](https://camo.githubusercontent.com/a5a8fabc657ba344e04992e0844d634dca711dae88954c0a80df6edf38eeeb3e/68747470733a2f2f702e697069632e7669702f3178343533372e706e67#from=url&id=gptPv&originHeight=720&originWidth=1074&originalType=binary&ratio=2&rotation=0&showTitle=false&status=done&style=none&title=)](https://camo.githubusercontent.com/a5a8fabc657ba344e04992e0844d634dca711dae88954c0a80df6edf38eeeb3e/68747470733a2f2f702e697069632e7669702f3178343533372e706e67)
+example curl command example
+
+```
+sudo curl --unix-socket /tmp/socks/klog.sock "http://localhost/setlevel?level=5&duration=60s"
+```
+
+## 3、Why not change klog level throuth change configmap
+Volcano is a commercial project. Volcano is deployed in a customer cluster. The configmap cannot be modified easily, but curl can be executed. I thought of using curl to pass in the klog value that you want to modify through the scoket method.
```
hwdef, line 35: I don't understand here, I think modifying the log level and modifying the configmap are both things done by the platform administrator, using curl is more intrusive. And I don't think this approach is very applicable, for example if my kubelet is running in a container, will it still work?

xiao-jay: Can you describe in detail the scenario where kubelet runs in a container?
Other reasons to use the curl scheme
1、Modifying configmap may affect other plugins
2、It'll be confusing that both options and configmap has same parameters.

hwdef: To be honest, I'm not sure about the benefits of running the kubelet in a container, but I do see a lot of cases at work where the kubelet is deployed in a container, and I'm guessing it might be for ease of deployment. rancher also runs the kubelet in a container.
When a kubelet is running in a container, the hostpath mounts the directory in the kubelet container, not the directory on the real host

xiao-jay: This affects only the volcano scheduler container. The mounting path is mapped from the volcano scheduler container to the host where the container is located. The kubelet in the container is not affected.

xiao-jay: > To be honest, I'm not sure about the benefits of running the kubelet in a container, but I do see a lot of cases at work where the kubelet is deployed in a container, and I'm guessing it might be for ease of deployment. rancher also runs the kubelet in a container. When a kubelet is running in a container, the hostpath mounts the directory in the kubelet container, not the directory on the real host

you mean the k8s deployed in container，so the kubelet deployed in the container？

hwdef: > you mean the k8s deployed in container，so the kubelet deployed in the container？

no, just the kubelet in container.

```
docker inspect kubelet
```

hwdef: for curl it, I think it will be OK, because the kubelet container is hostnetwork, but I feel this way (by curl) is not 'k8s', I can't think of any component that uses this method.

xiao-jay: @hwdef 
The usage scenario is that the log level needs to be temporarily raised in the customer's cluster to facilitate locating the cause of the error. However, changing the user configuration requires an apply, which is not required when using the curl command. Do you think there is any way in k8s to achieve the ability to change the log level without changing the configuration?

hwdef: Why refuse `apply`?

xiao-jay: The original solution was to modify the configmap, which contains the plugin configuration, and modifying the configmap may affect the plugin due to errors in modifying the configuration，There are risks.

hwdef: The log level is configured in the startup parameters and does not require modification of the configmap.

https://github.com/volcano-sh/volcano/blob/0d6d3b6a0dee7fffbc4e71e41f758d6d206e92bb/installer/volcano-development.yaml#L9178

xiao-jay: If you do this, the program will restart, and programs running online should not restart at will.

hwdef: volcano just a scheduler program. Restarting it won't give you a serious error.
Even in normal use, it can restart for a variety of reasons, so volcano-scheduler is deployed using deployment
If you are concerned about restart , then you should scale volcano-scheduler horizontally, e.g. to 3 replicas, so that even if one replica restart, the other two replicas are still working

william-wang (approved):
/lgtm

PR #3113, opened 2023-09-12
Feature: add pdb plugin

wang-mask:
Add PDB plugin for volcano to try to avoid deleting pods that violate PDB constraints as much as possible.

## Unit Test
If you want to verify this feature, you can run the unit test in pkg/scheduler/plugins/pdb/pdb_test.go

## Local Test
Start a local k8s cluster and install the latest valcano
```shell
$ cd <your-local-dir-to-volcano>
$ ./hack/local-up-volcano.sh
```
1. Scale the number of volcano-shceduler replicas to 0
```shell
$ kubectl scale --replicas=0 deployment -n volcano-system volcano-scheduler 
```
2. Create profile for volcano-shceduler to enable pdb support (like volcano-scheduler-configmap)
```shell
$ cd <your-local-dir-to-volcano>
$ cat<<EOF >volcano-scheduler.conf
actions: "enqueue, allocate, reclaim, preempt, backfill, shuffle"
tiers:
- plugins:
  - name: priority
  - name: pdb
EOF

```
> **Note**: This is a demo of volcano-scheduler.conf, you can create your own profile based on your needs.

3. Configure the startup file for debugging
```shell
$ cd <your-local-dir-to-volcano>
$ cat<<EOF >.vscode/launch.json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Scheduler Debug",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "program": "cmd/scheduler/main.go",
            "args": ["--logtostderr", "--scheduler-conf=<your-local-dir-to-volcano>/volcano-scheduler.conf", "--enable-healthz=true", 
            "--enable-metrics=true", "--leader-elect=false", "-v=4", "2>&1", "--kubeconfig=<your-local-path-to-kubeconfig>"]
        }
    ]
}
EOF

```
> **Note**: You need to change the `<your-local-dir-to-volcano>/volcano-scheduler.conf` to the local path to `volcano-scheduler.conf`, the `<your-local-path-to-kubeconfig>` to the local path to `kubeconfig` and adjust the other parameters based on your needs.

4. Make some examples to test locally

### DIFF
--- a/pkg/scheduler/plugins/pdb/pdb.go
+++ b/pkg/scheduler/plugins/pdb/pdb.go
@@ -62,7 +62,7 @@ func (pp *pdbPlugin) OnSessionOpen(ssn *framework.Session) {
 	}
 
 	// 1. define the func to filter out tasks that violate PDB constraints
-	filterFn := func(tasks []*api.TaskInfo) []*api.TaskInfo {
+	pdbFilterFn := func(tasks []*api.TaskInfo) []*api.TaskInfo {
 		var victims []*api.TaskInfo
 
 		// (a. get all PDBs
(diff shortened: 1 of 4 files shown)


william-wang:
Thanks for your contribution. We will review it as soon as possible.

pkg/scheduler/plugins/pdb/pdb.go
```diff
@@ -0,0 +1,147 @@
...
+			pdbsAllowed[i] = pdb.Status.DisruptionsAllowed
+		}
+
+		// (c. range every task to check if it violates the PDB constraints.
+		// If task does not violate the PDB constraints, then add it to victims.
+		for _, task := range tasks {
+			pod := task.Pod
+			pdbForPodIsViolated := false
+			// A pod with no labels will not match any PDB. So, no need to check.
+			if len(pod.Labels) != 0 {
```
wangyang0616, line 87: When the label information is empty, is it possible to continue directly to reduce the depth of code nesting?

pkg/scheduler/plugins/pdb/pdb.go
```diff
@@ -0,0 +1,147 @@
...
+	klog.V(4).Infof("Enter pdb plugin ...")
+	defer klog.V(4).Infof("Leaving pdb plugin.")
+
+	// 0. Init the PDB lister
+	if pp.lister == nil {
+		pp.lister = getPDBLister(ssn.InformerFactory())
+	}
+
+	// 1. define the func to filter out tasks that violate PDB constraints
+	filterFn := func(tasks []*api.TaskInfo) []*api.TaskInfo {
```
wangyang0616, line 65: Would pdbFilterFn be better?

pkg/scheduler/plugins/pdb/pdb.go
wang-mask: ok, it continue when len(labels) == 0

pkg/scheduler/plugins/pdb/pdb.go
wang-mask: have renamed it to pdbFilterFn

pkg/scheduler/plugins/pdb/pdb.go
```diff
@@ -0,0 +1,153 @@
...
+
+		// (b. init the pdbsAllowed array
+		pdbsAllowed := make([]int32, len(pdbs))
+		for i, pdb := range pdbs {
+			pdbsAllowed[i] = pdb.Status.DisruptionsAllowed
+		}
+
+		// (c. range every task to check if it violates the PDB constraints.
+		// If task does not violate the PDB constraints, then add it to victims.
+		for _, task := range tasks {
```
lowang-bh, line 83: Here, two for loop takes O(n2) latancy. Can we build up a cache of pdbs instead of iterating and parsing it in every task iteration.

pkg/scheduler/plugins/pdb/pdb.go
```diff
@@ -0,0 +1,153 @@
...
+			// A pod with no labels will not match any PDB. So, no need to check.
+			if len(pod.Labels) == 0 {
+				continue
+			}
+
+			for i, pdb := range pdbs {
+				if pdb.Namespace != pod.Namespace {
+					continue
+				}
+				selector, err := metav1.LabelSelectorAsSelector(pdb.Spec.Selector)
```
lowang-bh, line 96: how about do thoes common info parsing when open the session?

pkg/scheduler/plugins/pdb/pdb.go
william-wang: Agree, It's better to refator the loop later or submit a new pr.

william-wang (approved):
/lgtm

pkg/scheduler/plugins/pdb/pdb.go
wang-mask: It is necessary to check the pods' label match the pdbs' label when pdbFilterFn is called. Because the pdbs may be changed in one session.

pkg/scheduler/plugins/pdb/pdb.go
wang-mask: Yes, it can be cached for reduce the latancy. Now it get the pdbs info from informer. If you cache the pdbs info in session, it needs to maintain PDBs cache information(like tash cache in session) from an overall action level(Including rollback operation to cancel eviction). This would be complex..., but it is indeed a point worth optimizing later.

Issue #3123, opened 2023-09-14
Volcano scheduler will repeatedly sending pordGroup update request when there is no conditions field on podGroup resource.


Lily922:
**What happened**:
Volcano scheduler will repeatedly sending pordGroup update request when there is no conditions field on podGroup resource.

**What you expected to happen**:
Avoid unnecessary podGroup update requests.

**How to reproduce it (as minimally and precisely as possible)**:

1. Disable gang scheduler plugin
2. Create a new volcano job and wait for all pods to be scheduled, the podGroup's conditions filed is empty, this is because only the gang plugin updates the condition field now.
```
apiVersion: scheduling.volcano.sh/v1beta1
kind: PodGroup
metadata:
  creationTimestamp: "2023-09-14T07:16:37Z"
  generation: 5
  name: test-job-6-7979549f-af26-497a-9075-eaf180f4d65f
  namespace: default
  ownerReferences:
  - apiVersion: batch.volcano.sh/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Job
    name: test-job-6
    uid: 7979549f-af26-497a-9075-eaf180f4d65f
  resourceVersion: "2576959"
  uid: ba7ab924-5c85-4157-ae72-b6ea7d29bb65
spec:
  minMember: 3
  minResources:
    count/pods: "3"
    pods: "3"
  minTaskMember:
    default-nginx: 6
  priorityClassName: high-priority
  queue: default
status:
  phase: Running
  running: 6
```
3. When the `closeSession` program is executed, it will send update request because the podProupStatus is not put into the session cache `ssn.podGroupStatus`, `updatePG` will always be true
https://github.com/volcano-sh/volcano/blob/0a6f1e2e204685326c18785e4b00f19c9d74709b/pkg/scheduler/framework/job_updater.go#L97C1-L101C3

**Anything else we need to know?**

**Environment**:
- Volcano Version: 
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #3124, opened 2023-09-14
 Add podGroup status to session cache, fix the bug of repeatedly sending pordGroup update request when there is no condations field.

Lily922:
fix [bug](https://github.com/volcano-sh/volcano/issues/3123)


william-wang (approved):
/approve

PR #3125, opened 2023-09-14
[cherry-pick for release-1.8]Add podGroup status to session cache, fix the bug of repeatedly sending pordGroup update request when there is no condations field.

Lily922:
cherry-pick：https://github.com/volcano-sh/volcano/pull/3124


william-wang (approved):
/approve

PR #3128, opened 2023-09-16
[cherry-pick for release-1.8]fix: task scheduling latancy metrics is not accurate

lowang-bh:
cherry-pick for PR #3064

if one task bind failed, all the successfully binded pods' metrics will not be record util they are bind successfully in future scheduling cycle. This will cause latency and not accurate.

Also replace clientsets with interface.


wangyang0616:
It is best to associate the original PR information with the description information.

william-wang (approved):
/approve

PR #3134, opened 2023-09-21
:zap:  jobflow-controller jobtemplate-controller enqueue should use struct not pointer

kingeasternsun:
# Why should not use pointer
The  workqueue will deduplicate same item in queue.
According the internal struct of workqueue, when we add a item to workqueue firstly, workqueue will also put it in **q.dirty** which is `map[interface]struct{}``
```go
// Add marks item as needing processing.
func (q *Type) Add(item interface{}) {
	
    ...

	q.dirty.insert(item)
	if q.processing.has(item) {
		return
	}

	q.queue = append(q.queue, item)
	q.cond.Signal()
}
```

before this item processed, if same item re added to the workqueue , this add will skiped

```go

// Add marks item as needing processing.
func (q *Type) Add(item interface{}) {
	q.cond.L.Lock()
	defer q.cond.L.Unlock()
	if q.shuttingDown {
		return
	}
	if q.dirty.has(item) {
		return
	}
    ...
}
```

the current code use  `*apis.FlowRequest` as item to add to the workqueue,  even the different apis.FlowRequest have same values, different `*apis.FlowRequest` will be considerd as different item in workqueue, so the deduplication is not work.

We can try the example code blow, 
```go
type empty struct{}
type t interface{}
type set map[t]empty

func (s set) has(item t) bool {
	_, exists := s[item]
	return exists
}

func (s set) insert(item t) {
	s[item] = empty{}
}

func (s set) delete(item t) {
	delete(s, item)
}

func CompareMap() {
	a := &FlowRequest{Namespace: "a", JobFlowName: "n"}
	a1 := &FlowRequest{Namespace: "a", JobFlowName: "n"}
	b := &FlowRequest{Namespace: "b", JobFlowName: "n"}

	frMap := set{}
	frMap.insert(a)
	fmt.Println(frMap.has(a1))
	fmt.Println(frMap.has(b))
}
func CompareMapB() {
	a := FlowRequest{Namespace: "a", JobFlowName: "n"}
	a1 := FlowRequest{Namespace: "a", JobFlowName: "n"}
	b := FlowRequest{Namespace: "b", JobFlowName: "n"}

	frMap := set{}
	frMap.insert(a)
	fmt.Println(frMap.has(a1))
	fmt.Println(frMap.has(b))

}
```

the `CompareMap` will Print
```shell
false
false
```

the `CompareMapB` will Print
```shell
true
false
```

### DIFF
--- a/pkg/controllers/jobflow/jobflow_controller.go
+++ b/pkg/controllers/jobflow/jobflow_controller.go
@@ -74,7 +74,7 @@ type jobflowcontroller struct {
 	recorder record.EventRecorder
 
 	queue          workqueue.RateLimitingInterface
-	enqueueJobFlow func(req *apis.FlowRequest)
+	enqueueJobFlow func(req apis.FlowRequest)
 
 	syncHandler func(req *apis.FlowRequest) error
 
@@ -164,13 +164,13 @@ func (jf *jobflowcontroller) processNextWorkItem() bool {
 	// period.
 	defer jf.queue.Done(obj)
 
-	req, ok := obj.(*apis.FlowRequest)
+	req, ok := obj.(apis.FlowRequest)
 	if !ok {
 		klog.Errorf("%v is not a valid queue request struct.", obj)
 		return true
 	}
 
-	err := jf.syncHandler(req)
+	err := jf.syncHandler(&req)
 	jf.handleJobFlowErr(err, obj)
 
 	return true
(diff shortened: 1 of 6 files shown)


hwdef (commented):
/lgtm

william-wang (approved):
/approve

PR #3142, opened 2023-09-26
Dockerfile: go mod cache

7sunarni:
cache go modules, so improve building speed
https://docs.docker.com/language/golang/build-images/#create-a-dockerfile-for-the-application

### DIFF
--- a/installer/dockerfile/webhook-manager/Dockerfile
+++ b/installer/dockerfile/webhook-manager/Dockerfile
@@ -14,13 +14,17 @@
 
 FROM golang:1.20.1 AS builder
 WORKDIR /go/src/volcano.sh/
+COPY go.mod go.sum ./
+RUN go mod download
 ADD . volcano
 RUN cd volcano && make vc-webhook-manager
 
 FROM alpine:latest
 ARG KUBE_VERSION="1.27.0"
 ARG TARGETARCH
-RUN apk add --update ca-certificates && \
+ARG APK_MIRROR
+RUN if [[ -n "$APK_MIRROR" ]]; then sed -i "s@https://dl-cdn.alpinelinux.org@${APK_MIRROR}@g" /etc/apk/repositories ; fi && \
+    apk add --update ca-certificates && \
     apk add --update openssl && \
     apk add --update -t deps curl && \
     curl -L https://storage.googleapis.com/kubernetes-release/release/v$KUBE_VERSION/bin/linux/$TARGETARCH/kubectl -o /usr/local/bin/kubectl && \
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ image_bins: vc-scheduler vc-controller-manager vc-webhook-manager
 
 images:
 	for name in controller-manager scheduler webhook-manager; do\
-		docker buildx build -t "${IMAGE_PREFIX}/vc-$$name:$(TAG)" . -f ./installer/dockerfile/$$name/Dockerfile --output=type=${BUILDX_OUTPUT_TYPE} --platform ${DOCKER_PLATFORMS}; \
+		docker buildx build -t "${IMAGE_PREFIX}/vc-$$name:$(TAG)" . -f ./installer/dockerfile/$$name/Dockerfile --output=type=${BUILDX_OUTPUT_TYPE} --platform ${DOCKER_PLATFORMS} --build-arg APK_MIRROR=${APK_MIRROR}; \
 	done
 
 generate-code:
--- a/installer/dockerfile/controller-manager/Dockerfile
+++ b/installer/dockerfile/controller-manager/Dockerfile
@@ -14,6 +14,8 @@
 
 FROM golang:1.20.1 AS builder
 WORKDIR /go/src/volcano.sh/
+COPY go.mod go.sum ./
+RUN go mod download
 ADD . volcano
 RUN cd volcano && make vc-controller-manager
 
--- a/installer/dockerfile/scheduler/Dockerfile
+++ b/installer/dockerfile/scheduler/Dockerfile
@@ -14,6 +14,8 @@
 
 FROM golang:1.20.1 AS builder
 WORKDIR /go/src/volcano.sh/
+COPY go.mod go.sum ./
+RUN go mod download
 ADD . volcano
 RUN cd volcano && make vc-scheduler
 


7sunarni:
assign @thor-wl

7sunarni:
Hi~ Can you take minutes review this, it could speed up CI stage
@william-wang

hwdef:
please make the DCO CI happy first

7sunarni:
> please make the DCO CI happy first

Done

7sunarni:
It seems both of reviewers not active any more, how about assign new reviewer?

7sunarni:
Update: [this commit](https://github.com/volcano-sh/volcano/pull/3142/commits/a32b344f9990b106f6317d0449c2bcb9b13c44ba) use alpine mirror to speed up building webhook-manager docker image.

7sunarni:
@Monokaix Hi~ Sorry to tag you. It looks like that two reviewers assigned not active in this project. Do you know how and who to assign new reviewers?

Monokaix:
> Update: [this commit](https://github.com/volcano-sh/volcano/pull/3142/commits/a32b344f9990b106f6317d0449c2bcb9b13c44ba) use alpine mirror to speed up building webhook-manager docker image. I think most of Volcano's user in mainland. This mirror could be useful in improving building container image.

It's a good intention here. But volcano is used all over the world, we should not specify cdn url in a hard code way, my opinion is that set a parameter like env to let user specify custom cdn url or just disable it: )

Monokaix:
Welcome! A little doubt here. Does this cache optimization you mentioned only take effect when building multiple binaries? Seems that when we execute `make vc-controller-manager`, the `go mod download` will be executed, and after your pr, this step has been moved to the front, so what's the difference here?

7sunarni:
> Welcome! A little doubt here. Does this cache optimization you mentioned only take effect when building multiple binaries? 
I thinks this will work when you need to build the image in high frequency. It will not download the dependency each time if you don't modify the go.mod and go.sum.

7sunarni:
>  Does this cache optimization you mentioned only take effect when building multiple binaries? 

I think when your change not including go.mod and go.sum file and just including some go file, in docker building stage, this will not download dependency. Not only in building multiple binaries, and in local develop/test stage building image. 

> Seems that when we execute make vc-controller-manager, the go mod download will be executed, and after your pr, this step has been moved to the front, so what's the difference here?

As I know, docker will use the cache if the file don't change. In most case, `go.mod` and `go.sum` will not change,  so docker will use cache in step3 and step4 . This is also what [Docker document](https://docs.docker.com/language/golang/build-images/#multi-stage-builds) recommend us to do. And if we just copy `.volcano` to the build, this step has file change, docker will not use cache from this.

installer/dockerfile/webhook-manager/Dockerfile
```diff
@@ -14,13 +14,17 @@
...
+RUN go mod download
 ADD . volcano
 RUN cd volcano && make vc-webhook-manager
 
 FROM alpine:latest
 ARG KUBE_VERSION="1.27.0"
 ARG TARGETARCH
-RUN apk add --update ca-certificates && \
+ARG APK_MIRROR
+RUN if [[ -n "$APK_MIRROR" ]]; then sed -i "s@https://dl-cdn.alpinelinux.org@${APK_MIRROR}@g" /etc/apk/repositories ; fi && \
```
Monokaix, line 26: I think we should also add --build-args includes this env in docker build command: )

7sunarni:
> > Update: [this commit](https://github.com/volcano-sh/volcano/pull/3142/commits/a32b344f9990b106f6317d0449c2bcb9b13c44ba) use alpine mirror to speed up building webhook-manager docker image. I think most of Volcano's user in mainland. This mirror could be useful in improving building container image.
> 
> It's a good intention here. But volcano is used all over the world, we should not specify cdn url in a hard code way, my opinion is that set a parameter like env to let user specify custom cdn url or just disable it: )

You are right. I add a argument like `--build-arg APK_MIRROR="https://mirrors.huaweicloud.com"`  to let users do their choice, default will not do anything. 

By the way, it seems you are from Huawei. And the usage in this [official article](https://mirrors.huaweicloud.com/mirrorDetail/5ea14e7b02ea0eda6bd1e8fa) seems have a little mistake. The config command should be `sed -i "s@https://dl-cdn.alpinelinux.org/@https://repo.huaweicloud.com/@g" /etc/apk/repositories` . The comments in this article had point out this mistake, but not correct yet. :-(

installer/dockerfile/webhook-manager/Dockerfile
7sunarni: You are right.

Monokaix:
> > > Update: [this commit](https://github.com/volcano-sh/volcano/pull/3142/commits/a32b344f9990b106f6317d0449c2bcb9b13c44ba) use alpine mirror to speed up building webhook-manager docker image. I think most of Volcano's user in mainland. This mirror could be useful in improving building container image.
> > 
> > 
> > It's a good intention here. But volcano is used all over the world, we should not specify cdn url in a hard code way, my opinion is that set a parameter like env to let user specify custom cdn url or just disable it: )
> 
> You are right. I add a argument like `--build-arg APK_MIRROR="https://mirrors.huaweicloud.com"` to let users do their choice, default will not do anything.
> 
> By the way, it seems you are from Huawei. And the usage in this [official article](https://mirrors.huaweicloud.com/mirrorDetail/5ea14e7b02ea0eda6bd1e8fa) seems have a little mistake. The config command should be `sed -i "s@https://dl-cdn.alpinelinux.org/@https://repo.huaweicloud.com/@g" /etc/apk/repositories` . The comments in this article had point out this mistake, but not correct yet. :-(

Thanks, I will convey the mistake but I'm not responsible for that article actually: )

Monokaix:
> > > Update: [this commit](https://github.com/volcano-sh/volcano/pull/3142/commits/a32b344f9990b106f6317d0449c2bcb9b13c44ba) use alpine mirror to speed up building webhook-manager docker image. I think most of Volcano's user in mainland. This mirror could be useful in improving building container image.
> > 
> > 
> > It's a good intention here. But volcano is used all over the world, we should not specify cdn url in a hard code way, my opinion is that set a parameter like env to let user specify custom cdn url or just disable it: )
> 
> You are right. I add a argument like `--build-arg APK_MIRROR="https://mirrors.huaweicloud.com"` to let users do their choice, default will not do anything.
> 
> By the way, it seems you are from Huawei. And the usage in this [official article](https://mirrors.huaweicloud.com/mirrorDetail/5ea14e7b02ea0eda6bd1e8fa) seems have a little mistake. The config command should be `sed -i "s@https://dl-cdn.alpinelinux.org/@https://repo.huaweicloud.com/@g" /etc/apk/repositories` . The comments in this article had point out this mistake, but not correct yet. :-(

You can also make suggestions in the forum https://bbs.huaweicloud.com/suggestion/new directly.

7sunarni:
> > > Does this cache optimization you mentioned only take effect when building multiple binaries?
> > 
> > 
> > I think when your change not including go.mod and go.sum file and just including some go file, in docker building stage, this will not download dependency. Not only in building multiple binaries, and in local develop/test stage building image.
> > > Seems that when we execute make vc-controller-manager, the go mod download will be executed, and after your pr, this step has been moved to the front, so what's the difference here?
> > 
> > 
> > As I know, docker will use the cache if the file don't change. In most case, `go.mod` and `go.sum` will not change, so docker will use cache in step3 and step4 . This is also what [Docker document](https://docs.docker.com/language/golang/build-images/#multi-stage-builds) recommend us to do. And if we just copy `.volcano` to the build, this step has file change, docker will not use cache from this.
> 
> Got that! And please rebase and make a clean pr.

Done

Monokaix:
> --build-arg APK_MIRROR="https://mirrors.huaweicloud.com"

Is "--build-arg APK_MIRROR="https://mirrors.huaweicloud.com" updated? I didn't see that yet: )

william-wang:
retrigger the ci.

7sunarni:
> > --build-arg APK_MIRROR="https://mirrors.huaweicloud.com"
> 
> Is "--build-arg APK_MIRROR="https://mirrors.huaweicloud.com" updated? I didn't see that yet: )

I don't add this parameter in Makefile. If this parameter is missing, it will not change the alpine mirror.

Monokaix:
> > > --build-arg APK_MIRROR="https://mirrors.huaweicloud.com"
> > 
> > 
> > Is "--build-arg APK_MIRROR="https://mirrors.huaweicloud.com" updated? I didn't see that yet: )
> 
> I don't add this parameter in Makefile. If this parameter is missing, it will not change the alpine mirror.

Can the env transfer to docker image context？Seems that the env should be transferred through build-arg.

7sunarni:
> > > > --build-arg APK_MIRROR="https://mirrors.huaweicloud.com"
> > > 
> > > 
> > > Is "--build-arg APK_MIRROR="https://mirrors.huaweicloud.com" updated? I didn't see that yet: )
> > 
> > 
> > I don't add this parameter in Makefile. If this parameter is missing, it will not change the alpine mirror.
> 
> Can the env transfer to docker image context？Seems that the env should be transferred through build-arg.

It can. I add parameter to the Makefile.

william-wang (approved):
/approve

7sunarni:
@volcano-sh-bot  CI

PR #3144, opened 2023-09-27
Fix the problem of slow job deletion

WulixuanS:
Bug description:
When the job deletion rate in the cluster is greater than 10/s, there will be more and more job caches and metrics.

Reason:
1. Use  `AddRateLimited` when adding to the queue for the first time
2. Ratelimiter limits 10job/s

### DIFF
--- a/pkg/controllers/cache/cache.go
+++ b/pkg/controllers/cache/cache.go
@@ -248,7 +248,7 @@ func (jc *jobCache) DeletePod(pod *v1.Pod) error {
 		return err
 	}
 
-	if jc.jobs[key].Job == nil {
+	if jobTerminated(job) {
 		jc.deleteJob(job)
 	}
 
@@ -376,13 +376,20 @@ func (jc *jobCache) processCleanupJob() bool {
 		klog.V(3).Infof("Job <%s> was deleted.", key)
 	} else {
 		// Retry
-		jc.deleteJob(job)
+		jc.retryDeleteJob(job)
 	}
 	return true
 }
 
 func (jc *jobCache) deleteJob(job *apis.JobInfo) {
 	klog.V(3).Infof("Try to delete Job <%v/%v>",
+		job.Namespace, job.Name)
+
+	jc.deletedJobs.Add(job)
+}
+
+func (jc *jobCache) retryDeleteJob(job *apis.JobInfo) {
+	klog.V(3).Infof("Retry to delete Job <%v/%v>",
 		job.Namespace, job.Name)
 
 	jc.deletedJobs.AddRateLimited(job)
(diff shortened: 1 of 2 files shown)


hwdef:
Thanks for your PR!
Please clarify why removing the 10job/s limit won't affect the stability of the program.

WulixuanS:
> Thanks for your PR! Please clarify why removing the 10job/s limit won't affect the stability of the program.

Thanks, you are right, I should not remove the 10job/s limit, using `add` instead of `AddRateLimited` the first time already solves the problem.

pkg/controllers/cache/cache.go
```diff
@@ -248,7 +248,7 @@ func (jc *jobCache) DeletePod(pod *v1.Pod) error {
 		return err
 	}
 
-	if jc.jobs[key].Job == nil {
+	if jobTerminated(job) {
```
hwdef, line 251: What is the purpose of modifying this?

pkg/scheduler/cache/cache.go
```diff
@@ -939,9 +945,10 @@ func (sc *SchedulerCache) processCleanupJob() {
 		delete(sc.Jobs, job.UID)
 		metrics.DeleteJobMetrics(job.Name, string(job.Queue), job.Namespace)
 		klog.V(3).Infof("Job <%v:%v/%v> was deleted.", job.UID, job.Namespace, job.Name)
+		sc.DeletedJobs.Forget(obj)
```
hwdef, line 948: What is the purpose of modifying this?

WulixuanS: This clears the `rateLimiter`

pkg/controllers/cache/cache.go
WulixuanS: A job is associated with multiple pods, we only clean up the job after all pods are deleted. Judging the `jobTerminated ` method in advance can reduce the number of retries

pkg/scheduler/cache/cache.go
```diff
@@ -915,6 +915,12 @@ func (sc *SchedulerCache) taskUnschedulable(task *schedulingapi.TaskInfo, reason
 func (sc *SchedulerCache) deleteJob(job *schedulingapi.JobInfo) {
```
Monokaix, line 915: There are some other calls who called `deleteJob` method，do we also need replace it with `retryDeleteJob`?

WulixuanS: Where?

Monokaix: like https://github.com/volcano-sh/volcano/blob/2c5c1f9602eb3cae9645c3d5e299c6921f0685ff/pkg/scheduler/cache/event_handlers.go#L206

WulixuanS: Maybe you misunderstood the purpose of this PR. The purpose of this PR is to avoid triggering the rate limiter when the job cache is deleted for the first time, so there is no need to modify this place.

Monokaix: got that！thanks for your contribution.

WulixuanS:
cc @william-wang

william-wang (approved):
/approve

PR #3146, opened 2023-09-29
feat: dynamically change klog level through socket

### LINKED ISSUE #2322: dynamic log level adjustment without restarting volcano component

xiao-jay:
fix #2322 
docs #3111 
### Why not change klog level throuth change configmap
Volcano is a commercial project. Volcano is deployed in a customer cluster. The configmap cannot be modified easily, but curl can be executed. I thought of using curl to pass in the klog value that you want to modify through the scoket method.

###  Implementation Introduction
The vc-scheduler pod has an internal socket service to monitor changes in the klog.scok file. Because the socket service in the pod cannot be directly accessed in the k8s Node, the klog.sock file needs to be mounted from the pod to the cluster. Use The curl command modifies the klog.sock file.

This machine is a mac, use minikube to simulate the k8s cluster, enter mimikube, execute curl to modify the /tmp/socks/klog.sock file through the socket
```
 sudo curl --unix-socket /tmp/klog-socks/klog-klog.sock "http://localhost/setlevel?level=5&duration=60s"
```
image

Klog has been set to 5. This log can only be printed at level 5.
klog.V(5).Info("only klog v 5 can print this log ")
not using sudo will `curl: (7) Couldn't connect to server`
image

example image:xiaojie99999/vc-scheduler:1.1.7
@wangyang0616 @william-wang @Monokaix @lowang-bh

### DIFF
--- a/pkg/util/socket.go
+++ b/pkg/util/socket.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package util
 
 import (
--- a/pkg/util/socket.go
+++ b/pkg/util/socket.go
@@ -1,0 +1,210 @@
+package util
+
+import (
+	"context"
+	"fmt"
+	"html"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
+	"sync"
+	"time"
+
+	"golang.org/x/sys/unix"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/klog/v2"
+)
+
+const (
+	DefaultSocketDir = "/tmp/klog-socks" // Default directory storing socket files
+	SocketSuffix     = "-klog.sock"
+	SocketDirEnvName = "DEBUG_SOCKET_DIR"
+	// The HTTP request patterns
+	setLogLevelPath  = "/setlevel"
+	getLogLevelPath  = "/getlevel"
+	exampleSocketCli = "\"Failed to change klog log level, because got wrong value from level argument\\n\"+\n\t\t\t\t\"example: curl --unix-socket /tmp/klog-socks/componentName-klog.sock \\\"http://localhost/setlevel?level=8&duration=60s\\\"\\n\"+\n\t\t\t\t\"level=8 means changing klog log level to 8\\n\"+\n\t\t\t\t\"duration=60s means maintaining level=8 for 60 seconds[60m -> 60 minutes; 60h -> 60 hours]\""
+)
+
+var (
+	// When users frequently make request to change klog log level, the previously registered timer may not expire.
+	// To improve performance, cancel the previous timer. prevCtx, prevCtxCancelFunc is used to achieve this target.
+	prevCtx           context.Context
+	prevCtxCancelFunc context.CancelFunc
+
+	// currentLogLevel stores current log level
+	currentLogLevel string
+	// startupLogLevel stores start-up log level
+	startupLogLevel string
+	// mutex is used to avoid data race about prevCtx, prevCtxCancelFunc and currentLogLevel
+	mutex sync.RWMutex
+)
+
+// responseOk returns a statusOK response to client
+func responseOk(w *http.ResponseWriter, okMsg string) {
+	(*w).Header().Set("Content-Type", "text/plain; charset=utf-8")
+	(*w).Header().Set("X-Content-Type-Options", "nosniff")
+	_, err := fmt.Fprint(*w, okMsg)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/util/socket.go
```diff
@@ -0,0 +1,209 @@
...
+	// Register the HTTP request patterns that can change klog log level
+	mux.Handle(setLogLevelPath, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		values := r.URL.Query()
+		rawLevel := values.Get("level")
+		rawDuration := values.Get("duration")
+
+		if rawDuration == "" {
+			rawDuration = DefaultDuration
+		}
+		log.Println(rawLevel, rawDuration)
```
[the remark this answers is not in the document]
xiao-jay: done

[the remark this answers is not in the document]
xiao-jay: done

pkg/util/socket.go
```diff
@@ -0,0 +1,214 @@
...
+// reset creates a timer to make klog recover to start-up log level.
+func reset(ctx context.Context, duration time.Duration) {
+	defer runtime.HandleCrash()
+	select {
+	// Create a timer
+	case <-time.After(duration):
+		var loglevel klog.Level
+		mutex.Lock()
+		defer mutex.Unlock()
+		if err := loglevel.Set(startupLogLevel); err != nil {
```
lowang-bh, line 92: how about record an error log here?

pkg/util/socket.go
```diff
@@ -0,0 +1,214 @@
...
+			return
+		}
+
+		mutex.RLock()
+		// Create a timer to make klog recover to start-up log level.
+		// There will be more than one timer using same prevCtx variable under extreme conditions.
+		// Therefore, put reset function in mutex range.
+		go reset(prevCtx, duration)
+		responseOk(&w, fmt.Sprintf("Change klog log level to %s successfully and  for %v\n", currentLogLevel, duration))
+		mutex.RUnlock()
```
lowang-bh, line 141: Duding the read lock,  there is also an write lock requirement in gorutime, is that as your expect?

pkg/util/socket.go
xiao-jay: good suggestion,done

pkg/util/socket.go
xiao-jay: I didn't understand what you meant. Can you provide a more detailed scenario?

lowang-bh: I mean that you have locked with a readlock, and then start a goruntine in which a write lock is required.

xiao-jay: The wirte lock is in func `modifyLoglevel`,using to protect `prevCtx`.
```
func modifyLoglevel(newLogLevel string) error {
	mutex.Lock()
	defer mutex.Unlock()

	// Change klog log level to new value
	var loglevel klog.Level
	if err := loglevel.Set(newLogLevel); err != nil {
		return err
	}
	currentLogLevel = newLogLevel

	// Cancel the previous timer.
	if prevCtxCancelFunc != nil {
		prevCtxCancelFunc()
	}
	prevCtx, prevCtxCancelFunc = context.WithCancel(context.Background())
	return nil
}
```

lowang-bh: There is also write lock in the reset gorutine function. In the scenario previous timer if up and select case goes into the write lock require, but at the same time you are doing modifyLoglevel and have acquired the lock. Now the modifyLoglevel runs the cacel func of previous context. What will happen?

xiao-jay: You considered the scenario so carefully👍
if modifyLoglevel get the wirte lock first,modifyLoglevel will set klog and the reset will set klog after, it is **abnormal.**
if reset get the write lock first ,it will reset klog and the modifyLoglevel  will set klog ,it will be normal.

This usage scenario is to temporarily increase the klog level during debugging without considering concurrency, so there should be no need to deal with the low-probability events you mentioned.
what do you think？
@lowang-bh

pkg/util/socket.go
```diff
@@ -0,0 +1,210 @@
+package util
```
william-wang, line 1: please add the copy right.

xiao-jay: done

william-wang (approved):
/lgtm

Monokaix:
> The docs is not merged, and the code is merged. This is ridiculous.
> 
> Totally disagree with this PR code. image
> 
> @william-wang @lowang-bh

As described in doc, this pr change klog level in a hot-update way, and do not need to restart volcano.

xiao-jay:
> @Monokaix Did you read what was in my screenshot? Again, I'm against this approach.

Sometimes we encounter error when scheduling, we need to temporarily increase the log level to see more error messages, if the restart will make it more difficult to locate the problem.So we need dynamic change klog level way.

Monokaix:
> @Monokaix Did you read what was in my screenshot? Again, I'm against this approach.
There is some misunderstanding.
When we locate a problem of volcano, usually the existing log level is not enough and the problem cannot be located, however, after increasing the log level, the problem context is lost, making it impossible to continue locating. Therefore, it is convenient to update the log level without restarting volcano to identify the problem

hwdef:
There are many ways to do this, for example you can persist the log to a file. I don't think this pr is implemented in a good way. There is also something wrong with the code merge process.

But the code has been merged and I won't comment anymore.

Issue #3107, opened 2023-09-08
jobs.batch.volcano.sh is invalid: metadata.annotation: Too long


drmorr0:
**What happened**:

Following the instructions on [Installation](https://volcano.sh/en/docs/installation/) or the [README](https://github.com/volcano-sh/volcano#install-with-yaml-files), I ran

```
> kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/master/installer/volcano-development.yaml
...
The CustomResourceDefinition "jobs.batch.volcano.sh" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```

**What you expected to happen**:

All the resources should have been created.

**How to reproduce it (as minimally and precisely as possible)**:

Run `kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/master/installer/volcano-development.yaml` on a new cluster.

**Anything else we need to know?**:

Following the guidance on [this GitHub issue](https://github.com/prometheus-operator/prometheus-operator/issues/4355) I ran `kubectl create` instead of `kubectl apply` and all resources were created correctly.

**Environment**:
- Volcano Version: `master`
- Kubernetes version (use `kubectl version`):
```
> kubectl version
Client Version: v1.28.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.1
```
- Cloud provider or hardware configuration: Running locally on a [kind](https://kind.sigs.k8s.io) cluster
- OS (e.g. from /etc/os-release): Arch Linux
- Kernel (e.g. `uname -a`): `6.4.12-arch1-1`

lowang-bh:
I tested `kubectl create -f installer/volcano-development.yaml` and `helm install volcano installer/helm/chart/volcano --namespace volcano-system --create-namespace` works.

It is caused by [fetaure adapt k8s v1.27 #3083](https://github.com/volcano-sh/volcano/pull/3083)

After revert ths merge, it works with `kubectl apply`

Monokaix:
Hi, that's because an annotation `kubectl.kubernetes.io/last-applied-configuration` will be added to resource when using `kubectl apply`, and then exceed the resource limit of 256KB size,  you can use `kubectl apply ---server-side` to avoid the error, releated issue https://github.com/kubernetes/kubernetes/issues/82292

PR #3151, opened 2023-10-09
fix deployment command

ls-2018:
If you use the apply command, report this error.
`The CustomResourceDefinition "jobs.batch.volcano.sh" is invalid: metadata.annotations: Too long: must have at most 262144 bytes`


lowang-bh:
Relative issue: #3107

ls-2018:
@lowang-bh  thanks

Issue #3155, opened 2023-10-12
Panic in proportion plugin


svileex:
**What happened**:
The Proportion plugin panicked when I had queues with guaranteed resources and queues without guaranteed resources. The problem lies in resource division; we must subtract guaranteed resources before performing the remaining division among all queues.

I have already added a test for this behavior and fixed this bug.
```-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/go/volcano/pkg/scheduler/plugins (proporion-guarantee-fix*) » go test ./...                                                                                                                     svilex@svilexLin
?       volcano.sh/volcano/pkg/scheduler/plugins        [no test files]
ok      volcano.sh/volcano/pkg/scheduler/plugins/binpack        (cached)
ok      volcano.sh/volcano/pkg/scheduler/plugins/cdp    (cached)
?       volcano.sh/volcano/pkg/scheduler/plugins/conformance    [no test files]
ok      volcano.sh/volcano/pkg/scheduler/plugins/drf    (cached)
?       volcano.sh/volcano/pkg/scheduler/plugins/extender       [no test files]
?       volcano.sh/volcano/pkg/scheduler/plugins/gang   [no test files]
?       volcano.sh/volcano/pkg/scheduler/plugins/nodeorder      [no test files]
?       volcano.sh/volcano/pkg/scheduler/plugins/numaaware      [no test files]
ok      volcano.sh/volcano/pkg/scheduler/plugins/numaaware/policy       (cached)
ok      volcano.sh/volcano/pkg/scheduler/plugins/numaaware/provider/cpumanager  (cached)
?       volcano.sh/volcano/pkg/scheduler/plugins/overcommit     [no test files]
ok      volcano.sh/volcano/pkg/scheduler/plugins/predicates     (cached)
?       volcano.sh/volcano/pkg/scheduler/plugins/priority       [no test files]
E1012 21:47:39.004831  214480 utils.go:43] init kubeclient in 4pdvgpu failed: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
--- FAIL: TestGuarantee (0.00s)
panic: resource is not sufficient to do operation: <cpu 8000.00, memory 8000.00, nvidia.com/gpu 0.00> sub <cpu 10000.00, memory 10000.00> [recovered]
[...]

PR #3156, opened 2023-10-12
Fix panic in proportion plugin

svileex:
We must subtract guaranteed resources from the remaining resources before starting the division. If we don't do this, it's possible to allocate more resources than we have available in the cluster. For instance, in this scenario:

Suppose we have three queues - one with a significant guarantee of resources and two other queues. All three queues have jobs, so they will be handled in 'queueOpts.' If we start dividing resources from queues with no guarantee resources, then we give 'remaining/3' resources to each of them. However, the queue with guaranteed resources may take more than 'remaining/3' resources (it takes the maximum of 'attr.guarantee' and 'deserved'), which can result in a panic in the 'remaining.Sub' function.

This behavior is undefined because the order of elements in the map is random, but I have observed and reproduced this behavior many times. So, I have added a test to capture this behavior, specifically testing the panic without the fix. I have also added the fix.

Also, I'm going to create an issue with stack trace


svileex:
Issue: https://github.com/volcano-sh/volcano/issues/3155

svileex:
Hey, there hasn't been a review since October. Did I forget to do something?

Cdayz:
Hey @Thor-wl @lowang-bh see this mr please

Cdayz:
@william-wang @Thor-wl check this mr please

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -178,6 +178,19 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 	}
 
 	remaining := pp.totalResource.Clone()
+
+	// Process guarantee resources of the queues
+	// We must substract the guarantees from the remaining before
+	// fairness divison of the remianing
+	for _, attr := range pp.queueOpts {
+		// Queue can't desrve less resources than it has in guarantee
+		if attr.deserved.LessEqual(attr.guarantee, api.Zero) {
```
Monokaix, line 187: Seems deserved is always empty here, is this expected?

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -178,6 +178,19 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			guarantee := attr.guarantee.Clone()
+			attr.deserved = guarantee
+			remaining.Sub(guarantee)
```
Monokaix, line 190: The guaranted resources have already been subed here: https://github.com/volcano-sh/volcano/blob/79a8e41ebdb04b6e218deee04366f77b0a893ac1/pkg/scheduler/plugins/proportion/proportion.go#L120, do we still need to sub it ?

PR #3165, opened 2023-10-26
update volcano.sh/apis version

Lily922:
currently, healthz Authentication use the default authentication policy and it is insecure .
So it is necessary to set a list of cipher suites to enhance the security of TLS client authentication

related pr:  https://github.com/volcano-sh/apis/pull/117


william-wang (approved):
/lgtm

PR #3177, opened 2023-11-05
Make sure volcano schduler cache synced before first scheduling by waiting for handlers sync.

RamezesDong:
### Ⅰ. Describe what this PR does
The issue https://github.com/kubernetes/kubernetes/issues/116717 mentions the bug that event handlers hadn't handled all events when informer cache synced. This can lead to a terrible result, which is that the scheduler starts scheduling in the wrong state. The K8s community itself has fixed this issue https://github.com/kubernetes/kubernetes/pull/116729.

The PR makes sure handlers have finished syncing before the scheduling cycles start,  just like the default scheduler does.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -85,10 +85,13 @@ func (pc *Scheduler) Run(stopCh <-chan struct{}) {
 	pc.loadSchedulerConf()
 	go pc.watchSchedulerConf(stopCh)
 	// Start cache for policy.
+	klog.V(2).Infof("scheduler starts to run cache")
 	pc.cache.SetMetricsConf(pc.metricsConf)
 	pc.cache.Run(stopCh)
 	pc.cache.WaitForCacheSync(stopCh)
-	klog.V(2).Infof("Scheduler completes Initialization and start to run")
+	klog.V(2).Infof("scheduler starts to wait for cache handlers sync")
+	pc.cache.WaitForHandlersSync(stopCh)
+	klog.V(2).Infof("scheduler completes Initialization and start to run")
 	go wait.Until(pc.runOnce, pc.schedulePeriod, stopCh)
 	if options.ServerOpts.EnableCacheDumper {
 		pc.dumper.ListenForSignal(stopCh)
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -73,6 +73,9 @@ import (
 const (
 	// default interval for sync data from metrics server, the value is 30s
 	defaultMetricsInternal = 30 * time.Second
+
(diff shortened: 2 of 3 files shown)


lowang-bh:
Is it same with [wait for all items handled before start to scheduling #2822](https://github.com/volcano-sh/volcano/pull/2822)?

RamezesDong:
> same

It looks like the two are similar. If the other one is merged, I will close the pr.

lowang-bh:
We'd better include this feature in new release. @Monokaix @william-wang

Monokaix:
Please rebase your pr: )

RamezesDong:
I will rebase the code as soon as possible

RamezesDong:
@lowang-bh The code rebase is done. Could you review these codes, please?

pkg/scheduler/scheduler.go
```diff
@@ -88,7 +88,10 @@ func (pc *Scheduler) Run(stopCh <-chan struct{}) {
 	pc.cache.SetMetricsConf(pc.metricsConf)
 	pc.cache.Run(stopCh)
 	pc.cache.WaitForCacheSync(stopCh)
-	klog.V(2).Infof("Scheduler completes Initialization and start to run")
+	if err := pc.cache.WaitForHandlersSync(stopCh); err != nil {
+		panic(fmt.Sprintf("failed to wait for handlers sync: %v", err))
```
lowang-bh, line 92: If we don't panic and go on, what will happen? Is that acceptable？

pkg/scheduler/scheduler.go
```diff
@@ -88,7 +88,10 @@ func (pc *Scheduler) Run(stopCh <-chan struct{}) {
 	pc.cache.SetMetricsConf(pc.metricsConf)
 	pc.cache.Run(stopCh)
 	pc.cache.WaitForCacheSync(stopCh)
-	klog.V(2).Infof("Scheduler completes Initialization and start to run")
```
lowang-bh, line 91: We can add a log before  custom handler sync, so that user can check the latancy from log.

pkg/scheduler/scheduler.go
RamezesDong: The only way that error is returned is if stopCh cancel with ErrWaitTimeout. So we can make the method WaitForHandlersSync returnless and remove this panic.

pkg/scheduler/scheduler.go
RamezesDong: Of course, I will add logs about the waiting cache initialized.

Monokaix:
Does volcano controller also need catch this?

RamezesDong:
> Does volcano controller also need catch this?

I don't think so. The controller just needs to make sure it's eventual consistency, waiting for cache sync is enough

PR #3179, opened 2023-11-06
Update the default value of parameter worker-threads-for-podgroup to 5

Lily922:
The default value of parameter `worker-threads-for-podgroup` is 1，it's too small.
When you have a large number of deployments running concurrently, pod scheduling will be slow because of the slow creation of podGroup resources. Therefore, change the default value to 5 to increase the concurrency capability of podGroup resource creating.


william-wang (approved):
/lgtm

PR #3197, opened 2023-11-13
fix job `CRD metadata.annotations: Too long` error

Monokaix:
Fix https://github.com/volcano-sh/volcano/issues/3107

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ RELEASE_DIR=_output/release
 REPO_PATH=volcano.sh/volcano
 IMAGE_PREFIX=volcanosh
 CRD_OPTIONS ?= "crd:crdVersions=v1,generateEmbeddedObjectMeta=true"
+CRD_OPTIONS_EXCLUDE_DESCRIPTION=${CRD_OPTIONS}",maxDescLen=0"
 CC ?= "gcc"
 SUPPORT_PLUGINS ?= "no"
 CRD_VERSION ?= v1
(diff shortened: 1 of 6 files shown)


hwdef (commented):
I don't think this repair is good, because when the yaml is regenerated, it will be overwritten.

Monokaix:
> I don't think this repair is good, because when the yaml is regenerated, it will be overwritten.

yeah, it's a question, `kubectl apply --server-side` is another way.

Monokaix:
> I don't think this repair is good, because when the yaml is regenerated, it will be overwritten.

Limit the volcano job crd yaml's description when using CONTROLLER_GEN, please check it once more.

hwdef (commented):
/lgtm

Makefile
```diff
@@ -17,6 +17,7 @@ RELEASE_DIR=_output/release
 REPO_PATH=volcano.sh/volcano
 IMAGE_PREFIX=volcanosh
 CRD_OPTIONS ?= "crd:crdVersions=v1,generateEmbeddedObjectMeta=true"
+CRD_OPTIONS_EXCLUDE_DESCRIPTION=${CRD_OPTIONS}",maxDescLen=0"
```
william-wang, line 20: Is there any effect after removing the description?

Monokaix: description comes from the comment of CRD'field, and removing it has no effect to job's CRD definition, the only effect is that user use `kubectl get crd xxx -oyaml` cannot see the descrption of every field and then can check code or doc to get detail description of every field.

william-wang (approved):
/approve

Issue #3191, opened 2023-11-11
support preemption when the pods number or attachment volumes number of a node reaches the upper limit 


Lily922:
#### What would you like to be added:
Currently, when the `preempt` feature is enabled, only when the node's memory/cpu is insufficient, the preemption logic will be triggered. If the `number of pods` on a node or the `number of attached volumes` of a pod reaches the limit, preemption will not take effect. That is, in these cases, a high-priority pod cannot free up node resources by evicting low-priority pods.

#### Why is this needed:
When the `preempt` feature is enabled, if the number of pods on a node or the number of attached volumes of a pod reaches the limit, preemption should be take effect. 

#### Examples for `number of pods on a node reached the limits`
1. enable `preempt` actions

```
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "allocate, preempt, backfill"
    tiers:
    - plugins:
      - name: priority
   ......
```

2. create deployment with low priority
```
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: low-priority
value: 1
globalDefault: false
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: low-priority-dep
  labels:
    app: low-priority-dep
spec:
  replicas: 10
  selector:
    matchLabels:
      app: low-priority-dep
  template:
    metadata:
      labels:
        app: low-priority-dep
    spec:
      priorityClassName: low-priority
      schedulerName: volcano
      containers:
      - name: container-1
        image: busybox:none
        imagePullPolicy: IfNotPresent
        command: ['sh', '-c', 'sleep 100000']
```
waiting for the pods are running  
```
# kubectl get pod -owide
NAME                                READY   STATUS    RESTARTS   AGE   IP               NODE             NOMINATED NODE   READINESS GATES
low-priority-dep-5b5d858bbb-48j7c   1/1     Running   0          47s   192.168.86.213   192.168.20.1   <none>           <none>
low-priority-dep-5b5d858bbb-94rmd   1/1     Running   0          47s   192.168.86.219   192.168.20.1   <none>           <none>
low-priority-dep-5b5d858bbb-fj9dz   1/1     Running   0          47s   192.168.86.228   192.168.20.1   <none>           <none>
low-priority-dep-5b5d858bbb-hmvzt   1/1     Running   0          46s   192.168.86.238   192.168.20.1   <none>           <none>
[...]

PR #3211, opened 2023-11-20
support preemption when the number of attachment volumes of a node reaches the upper limit

Lily922:
support feature described in https://github.com/volcano-sh/volcano/issues/3191

Design：
1. Add a workqueue for node resource ：`nodeQueue   workqueue.RateLimitingInterface`, add `nodeName` to the queue when handling a `CSINode` resource or `Node` resource change event.
The reasons for this are as follows:
    - When Node and CSINode resources with the same name need to be processed at the same time, we need to ensure data consistency and ensure that the data is up to date, To do this, we need a mutex lock to ensure mutual exclusion of processing:
    ```
   CSINode resource event handle:
   1.  try to get mutex lock
   3.  get latest Node resource and latest CSINode resouce from informer cache
   4.  get allocatable resources (from node status and scinode spec)
   5.  add allocatable  resources to scheduler cache

   Node resource event handle:
   1.  try to get mutex lock
   2.  get latest Node resource and latest CSINode resouce from informer cache
   3.  get allocatable resources (from node status and scinode spec)
   4.  add allocatable  resources to scheduler cache
    ```
   By using workqueue, we can ignore the change events is from Node or CSINode
    ```
    AddNode, DeleteNode, UpdateNode Event:  --> add nodeName to workqueue
    AddCSINode, DeleteCSINode, UpdateCSINode Event:  --> add nodeName to workqueue

    process workequeue:
    1. get nodeName from queue
    2. get latest Node resource and latest CSINode resouce from informer cache
    3. get allocatable resources (from node status and scinode spec)
    4. add allocatable  resources to scheduler cache
    ```
    
    - When processing events, an error may occur and you need to try again, such as failure to get a storageclass resource(because of the resource has not been created yet or for other reasons ), this failure can be solved by retrying. By using `workqueue`,  we can use `AddRateLimited` to add the resources that failed to be processed to workqueue and wait for retry processing

2.  Don't check whether the CSI volume type has a quantity limit while `NewTaskInfo` or `SetNode`, Add all csi volumes on CSINode to cache,  because the storage may change from unattachable volumes to attachable volumes after the cache set up, in this case it is very difficult to refresh all task caches, like case:
    ```
    1. A pod using disk csi volume (driverName: disk.csi.io) is created, check the disk driver on CSINode has no `allocatable` parameter
    apiVersion: storage.k8s.io/v1
    kind: CSINode
       name: 192.168.1.1
    spec:
       drivers:
       - name: disk.csi.io
         nodeID: xxxxxx.xxxxxx.xxxxxxx.xxxxxx
    NewTaskInfo generates a taskInfo without no Resreq about `disk.csi.io` because no `allocatable` parameter on CSINode.
    2. The pod will be scheduled to 192.168.1.1 successfully
    3. The CSINode is updated, and disk driver has  `allocatable` parameter
    apiVersion: storage.k8s.io/v1
    kind: CSINode
       name: 192.168.1.1
    spec:
       drivers:
       - name: disk.csi.io
         nodeID: xxxxxx.xxxxxx.xxxxxxx.xxxxxx
         allocatable:
            count: 10
    4. Add disk allocate resource to Node cache 
        node 192.168.1.1 : allocatable 10 disk  
        node 192.168.1.1 : used 0 disk
        But actually one disk is used by pod in step 2, the used number should be 1 disk
    ```

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -19,6 +19,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"math"
 	"reflect"
 	"strconv"
 
@@ -26,13 +27,17 @@ import (
 	schedulingv1 "k8s.io/api/scheduling/v1"
 	sv1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/component-helpers/storage/ephemeral"
+	storagehelpers "k8s.io/component-helpers/storage/volume"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpumanager/topology"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
+	volumeutil "k8s.io/kubernetes/pkg/volume/util"
 
 	nodeinfov1alpha1 "volcano.sh/apis/pkg/apis/nodeinfo/v1alpha1"
 	"volcano.sh/apis/pkg/apis/scheduling"
@@ -43,6 +48,8 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 	commonutil "volcano.sh/volcano/pkg/util"
 )
+
+var DefaultAttachableVolumeQuantity int64 = math.MaxInt32
 
 func isTerminated(status schedulingapi.TaskStatus) bool {
 	return status == schedulingapi.Succeeded || status == schedulingapi.Failed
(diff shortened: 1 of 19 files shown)


wangyang0616:
Please associate with the corresponding issue and complete the description.

william-wang:
@Lily922 Please rebase and solve the confliction :)

pkg/scheduler/cache/event_handlers.go
```diff
@@ -66,6 +73,128 @@ func (sc *SchedulerCache) getOrCreateJob(pi *schedulingapi.TaskInfo) *scheduling
...
+			pvcName = vol.PersistentVolumeClaim.ClaimName
+		case vol.Ephemeral != nil:
+			// Generic ephemeral inline volumes also use a PVC,
+			// just with a computed name and certain ownership.
+			// That is checked below once the pvc object is
+			// retrieved.
+			pvcName = ephemeral.VolumeClaimName(pod, &vol)
+			isEphemeral = true
+		default:
+			// TODO: support Inline volumes.
```
wangyang0616, line 106: Please mark the submitter information in todo. Such as: "@Lily922 TODO: support Inline volumes."

pkg/scheduler/cache/event_handlers.go
```diff
@@ -66,6 +73,128 @@ func (sc *SchedulerCache) getOrCreateJob(pi *schedulingapi.TaskInfo) *scheduling
...
+		driverName := sc.getCSIDriverInfo(pvc)
+		if driverName == "" {
+			klog.V(5).InfoS("Could not find a CSI driver name for pvc(%s/%s), not counting volume", pvc.Namespace, pvc.Name)
+			continue
+		}
+
+		// Count all csi volumes in cache, because the storage may change from unattachable volumes to attachable volumes after
+		// the cache set up, in this case it is very difficult to refresh all task caches.
+		// For unattachable volume, set the limits number to a very large value, in this way, scheduling will never
+		// be limited due to insufficient quantity of it.
```
wangyang0616, line 136: I think it's better to segment annotations according to semantics.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -66,6 +73,128 @@ func (sc *SchedulerCache) getOrCreateJob(pi *schedulingapi.TaskInfo) *scheduling
...
+		klog.V(5).InfoS("Failed to get pv <%s> for pvc <%s>: %v", klog.KRef("", pvName), klog.KObj(pvc), err)
+		// If we can't fetch PV associated with PVC, may be it got deleted
+		// or PVC was prebound to a PVC that hasn't been created yet.
+		// fallback to using StorageClass for volume counting
+		return sc.getCSIDriverInfoFromSC(pvc)
+	}
+
+	csiSource := pv.Spec.PersistentVolumeSource.CSI
+	if csiSource == nil {
+		// TODO: support non-CSI volumes and migrating pvc
```
wangyang0616, line 166: Same as above.

Lily922: Done

pkg/scheduler/cache/event_handlers.go
Lily922: Done

william-wang (approved):
/approve

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -515,10 +515,7 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 		if predicate.nodeVolumeLimitsEnable {
 			status := nodeVolumeLimitsCSIFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
 			nodeVolumeStatus := framework.ConvertPredicateStatus(status)
-			if nodeVolumeStatus.Code != api.Success {
-				predicateStatus = append(predicateStatus, nodeVolumeStatus)
-				return predicateStatus, fmt.Errorf("plugin %s predicates failed %s", nodeVolumeLimitsCSIFilter.Name(), status.Message())
-			}
+			predicateStatus = append(predicateStatus, nodeVolumeStatus)
```
lowang-bh, line 518: Why don't retrun if failed here? see issue https://github.com/volcano-sh/volcano/issues/3050

PR #3213, opened 2023-11-20
Delete duplicate logs generated by the predicate_helper method

guoqinwill:
Delete duplicate logs generated by the predicate_helper method.
When the predicate plugin in failes, the logs about task information is repeatedly printed:
image


william-wang (approved):
/approve

PR #3217, opened 2023-11-21
fix panic when get elastic resource in proportion plugin

### LINKED ISSUE #3105: scheduler panic when calculate the elastic resource

flyingfang:
fix #3105

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -29,11 +29,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
-
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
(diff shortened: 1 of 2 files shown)


lowang-bh:
/hold

there is already one pr https://github.com/volcano-sh/volcano/pull/3106

Issue #3204, opened 2023-11-15
The calculation of queue's capability in proportion plugin is confusing.


flyingfang:
The deserved resource of queue comes from queue's realcapability. However, the queue's realcapability comes from the total resources minus  guranteed resources in the cluster.

When the queue does not contains a certain extended resource in k8s, like `rdma/hca`, after being alculated by the function 'helpers.Min', the queue's realcapability will set that resource to 0.

pkg/scheduler/plugins/proportion/proportion.go
```
realCapability := pp.totalResource.Clone().Sub(pp.totalGuarantee).Add(attr.guarantee)
if attr.capability == nil {
	attr.realCapability = realCapability
} else {
	attr.realCapability = helpers.Min(realCapability, attr.capability)
}
```

For example, after setting the queue's capability that not containing the `rdma/hca`. Then  I submit the job with resource `rdma/hca`. The PodGroup of the Job will be Inqueue.

my queue's yaml likes:
```
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: test-queue
spec:
  capability:
    nvidia.com/gpu: 10
```

When I submit the Job, the job's resources likes:

```
resources:
  limits:
    nvidia.com/gpu:8
    rdma/hca: 1
```

Could we change the calculation method of queue's realcapability, like that:
pkg/scheduler/api/helpers/helpers.go
```
// Min is used to find the min of two resource types
func Min(l, r *api.Resource) *api.Resource {
	res := &api.Resource{}

	res.MilliCPU = math.Min(l.MilliCPU, r.MilliCPU)
	res.Memory = math.Min(l.Memory, r.Memory)

	if l.ScalarResources == nil || r.ScalarResources == nil {
		return res
	}

	res.ScalarResources = map[v1.ResourceName]float64{}
	for lName, lQuant := range l.ScalarResources {
		if val, ok := r.ScalarResources[lName]; ok {
			res.ScalarResources[lName] = math.Min(lQuant, val)
		} else {
			res.ScalarResources[lName] = lQuant
		}
	}

	return res
}
```

Monokaix:
@qiankunli Can you take a look?

PR #3219, opened 2023-11-22
queue realcapability change to min dimension of queue capability and …

Monokaix:
fix https://github.com/volcano-sh/volcano/issues/3204


william-wang (approved):
/approve

PR #3223, opened 2023-11-23
skip 'pods' resource when checking if the Resource is empty

Lily922:
fix issue: https://github.com/volcano-sh/volcano/issues/3227

Design：
Skip 'pods' resource when checking if the Resource is empty.

```
1. "allocate" action:
   if resources.IsEmpty() :    // task with empty requsted resource:
       skip to schedule it
   else:
       try to schedule it

2. "preempt" action:
   if !task.IsEmpty():
      ignore task
   else:
      try to preempt:
          if target task.IsEmpty():
              ignore task
          else:
              add task to victims queue

3. "backfill" action:
   if resources.IsEmpty() :    // task with empty requsted resource:
       try to schedule it
   else:
       skip to schedule it
              
```

1.  Pods which request none resource except "pods" number resource will be scheduled in "backfill" action
2.  Pods request any resource including scalar resources will be scheduled in "allocate" action
3.  Pods which should be scheduled in "backfill" action can never become a preemptor or usurper of resources in "preempt" action，even if it has a higher or lower priority

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -23,6 +23,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	v1helper "k8s.io/kubernetes/pkg/scheduler/util"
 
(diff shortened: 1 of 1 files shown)


lowang-bh (approved):
/LGTM

pkg/scheduler/api/resource_info.go
```diff
@@ -193,13 +194,24 @@ func (r *Resource) Get(rn v1.ResourceName) float64 {
 	}
 }
 
+// Skip checking "pods" resource.
+// All pods request one "pods" resource now, no need to check it
+var emptyCheckIgnoredScalarResources = sets.NewString(string(v1.ResourcePods))
+
+func IsEmptyCheckIgnoredScalarResource(name v1.ResourceName) bool {
```
william-wang, line 201: It's better to change a readable function name.  For example, `IsIgnoredScalarResource` or `IsInBlacklist`

pkg/scheduler/api/resource_info.go
```diff
@@ -193,13 +194,24 @@ func (r *Resource) Get(rn v1.ResourceName) float64 {
...
+	return emptyCheckIgnoredScalarResources.Has(string(name))
+}
+
 // IsEmpty returns false if any kind of resource is not less than min value, otherwise returns true
```
william-wang, line 205: Update the description for this function

Lily922: Done

pkg/scheduler/api/resource_info.go
Lily922: Done

pkg/scheduler/api/resource_info.go
```diff
@@ -193,13 +194,24 @@ func (r *Resource) Get(rn v1.ResourceName) float64 {
...
+
+// IsEmpty returns false if any kind of resource other than IgnoredResources is not less than min value, otherwise returns true
 func (r *Resource) IsEmpty() bool {
 	if !(r.MilliCPU < minResource && r.Memory < minResource) {
 		return false
 	}
 
-	for _, rQuant := range r.ScalarResources {
+	for rName, rQuant := range r.ScalarResources {
+		if IsIgnoredScalarResource(rName) {
```
lowang-bh, line 212: Why don't use "ignoredScalarResources.Has" direclty to judge it? Currenty, there will be one more function calling.

william-wang (approved):
/approve

PR #3228, opened 2023-11-25
Avoid repeatedly creating links to obtain node metrics

wangyang0616:
After enabling the load-aware scheduling capability, Volcano will obtain node monitoring data through the Custom metrics API. 

Optimize the logic of the link with the apiserver to avoid repeated link creation, resulting in increased pressure on the apiserver.

### DIFF
--- a/pkg/scheduler/metrics/source/metrics_client_prometheus_adapt.go
+++ b/pkg/scheduler/metrics/source/metrics_client_prometheus_adapt.go
@@ -36,23 +36,32 @@ const (
 	CustomNodeMemUsageAvg = "node_memory_usage_avg"
 )
 
-type CustomMetricsClient struct {
-	config *rest.Config
+type KMetricsClient struct {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/metrics/source/metrics_client_prometheus_adapt.go
```diff
@@ -36,31 +36,40 @@ const (
 	CustomNodeMemUsageAvg = "node_memory_usage_avg"
 )
 
-type CustomMetricsClient struct {
-	config *rest.Config
+type KMetricsClient struct {
```
william-wang, line 39: `KMetricsClient` is not a good name. What's the meaning of it?

wangyang0616: It means k8s metrics api client.

william-wang:
There is a ci failure @wangyang0616

william-wang (approved):
/lgtm

PR #3087, opened 2023-08-29
fix: ssn JobValid

wangzhipeng:
JobValid  after ssn.Tiers and ssn.Configurations  Initialization required 
   
OnSessionOpen  to register the function jobValidFns

PR #3248, opened 2023-12-06
change log level to warning when skip allocatation if job is not valid

lowang-bh:
To expose abnormal case and debug easily.


lowang-bh:
@wangyang0616 @hwdef

wangyang0616:
When the job exception does not meet the judgment conditions of the gang policy, will this log output be too frequent?

lowang-bh:
> When the job exception does not meet the judgment conditions of the gang policy, will this log output be too frequent?

Yes, maybe it is a problem.  Without the warning, user sometimes can not think out why job is not scheduled.

But, there is a good way to solve it by record the events in [fix: ssn JobValid #3087](https://github.com/volcano-sh/volcano/pull/3087)

Please help to merge that PR and then close this one.

Issue #3203, opened 2023-11-15
volcano性能验证


sharphu:
volcano号称是若是10000节点，会有 2000pods/s的调度，我想请问的是针对这个数据，你们是怎么得出来的？我现在想要验证它的话，需要怎么做？有没有验证教程？
[https://github.com/volcano-sh/volcano/blob/master/docs/Volcano%20-%20A%20Kubernetes%20Native%20Batch%20System.pdf](url)

lowang-bh:
I have a plan to  write a mock-scheduler on volcano, to do scheduling performance test for large scale nodes and pods.

sharphu:
> I have a plan to write a mock-scheduler on volcano, to do scheduling performance test for large scale nodes and pods.

how to do？

lowang-bh:
> > I have a plan to write a mock-scheduler on volcano, to do scheduling performance test for large scale nodes and pods.
> 
> how to do？

Just use fake client to replace apiserver when binding to nodes. Other scheduling logic is same as origin.

bysph:
> I have a plan to write a mock-scheduler on volcano, to do scheduling performance test for large scale nodes and pods.

Do you have more detailed information about this plan? I'd like to see if I can help.Because recently we are also planning to test the performance of Volcano.

lowang-bh:
> Do you have more detailed information about this plan? I'd like to see if I can help.Because recently we are also planning to test the performance of Volcano.

We can use fake instance to simulator those interface in scheduler cache and leave the whole scheduing logica same as volcano scheduler. To simulating the scheduler, those steps are need:
1. make sure all clients are use interface: please help to merge https://github.com/volcano-sh/volcano/pull/2866 first. @william-wang 
2. then define a MockCache similar to Cache in volcano-scheduler
3. deveop a http client which support CRUD operations to manage fake resource suce as pod/job/podgroup/node, etc.
4. creating large number of those resource and wait mock-scheduler to scheduling those pods, statistic those metrics data.

There is also anothter simple way: use https://github.com/kubernetes-sigs/kwok to replace real k8s nodes. Then create large number os fake-nodes and submit large numbers of pods to those fake-nodes and waiting volcano scheduling.

bysph:
> > Do you have more detailed information about this plan? I'd like to see if I can help.Because recently we are also planning to test the performance of Volcano.
> 
[...]

PR #3260, opened 2023-12-11
test: add ut for performance test

bysph:
This MR aims to add a simple performance UT for Volcano.

Two reasons for supplementing UT:

1. There are some questions about performance data, such as https://github.com/volcano-sh/volcano/issues/3203.
2. Performance-oriented UT helps Volcano make optimizations.

Run this case using the following commands:
```
cd pkg/scheduler
go test -v --run=TestRunOnc
```

I have run this case under default configuration without optimization in a machine (56C  256G), and get this result：
image

I also added a "latencyrecorder" locally to track the execution time of various plugins through instrumentation. However, it's not elegant because it is implemented through code instrumentation, so I haven't submitted it yet.

Can anyone take a look first and give some suggestions? thx~


bysph:
@lowang-bh  Could you please spare some time to review my code? thx~

bysph:
> Thanks for your hard works. Can it cover all those schedule logic in the real scheduler? My idea is to use k8s fakeclient to submit jobs/nodes/pods and then triger the scheduler run.

I think it covers most of the schedule logic, except for the part that calls the k8s API(If there are any errors in my statement, please correct me, as I am not very familiar with volcano code.). According to my understanding, the cache is used as middle layer between k8s and Volcano scheduling logic, so I chose to add a mock cache.

The reason I don't directly construct a real cache through a fake kube client is because certain parts of the cache implementation use kubeconfig to create a new client (e.g., vcclient). These scenarios are difficult to use a fake kube client. Do you have any good ideas on how to handle this?

lowang-bh:
> The reason I don't directly construct a real cache through a fake kube client is because certain parts of the cache implementation use kubeconfig to create a new client (such as vcclient). These scenarios are difficult to use a fake kube client. Do you have any good ideas on how to handle this?

Yes,  I have some plan to do it. I will write a demo and then we can work together on that feature. 

I think we can keep this pr also.
@wangyang0616 @Monokaix

bysph:
> > The reason I don't directly construct a real cache through a fake kube client is because certain parts of the cache implementation use kubeconfig to create a new client (such as vcclient). These scenarios are difficult to use a fake kube client. Do you have any good ideas on how to handle this?
> 
> Yes, I have some plan to do it. I will write a demo and then we can work together on that feature.
> 
> I think we can keep this pr also. @wangyang0616 @Monokaix

Thanks~ hope I can help

Monokaix:
> > The reason I don't directly construct a real cache through a fake kube client is because certain parts of the cache implementation use kubeconfig to create a new client (such as vcclient). These scenarios are difficult to use a fake kube client. Do you have any good ideas on how to handle this?
> 
> Yes, I have some plan to do it. I will write a demo and then we can work together on that feature.
> 
> I think we can keep this pr also. @wangyang0616 @Monokaix

I think there are two process that we should concern, the first one is schedule logic in every session, this mainly consists of predict and nodeOrder plugin, the second one is the kube-apiserver throughtput in real cluster, so your test can only test the first part and in my opion, we have no necessary to mock a cache becauese cache is composed of many interfaces, we just need mock these interfaces such as kube client : )

bysph:
> > > The reason I don't directly construct a real cache through a fake kube client is because certain parts of the cache implementation use kubeconfig to create a new client (such as vcclient). These scenarios are difficult to use a fake kube client. Do you have any good ideas on how to handle this?
> > 
> > 
> > Yes, I have some plan to do it. I will write a demo and then we can work together on that feature.
> > I think we can keep this pr also. @wangyang0616 @Monokaix
> 
> I think there are two process that we should concern, the first one is schedule logic in every session, this mainly consists of predict and nodeOrder plugin, the second one is the kube-apiserver throughtput in real cluster, so your test can only test the first part and in my opion, we have no necessary to mock a cache becauese cache is composed of many interfaces, we just need mock these interfaces such as kube client : )
> 
> Also, kube-apiserver throughput has a great influence on scheduler performance, so we abstract a bind interface to enable user register their own bind method like bind a lots of pods in one bind request, so you can also work on this to achieve a higher performance.

The bottleneck in the second part is not only in the bind, but also in the performance of the k8s, so it may not be solvable entirely with ut, but requires e2e testing. You are right that the batch binding part is not tested by this ut, and a  implementation of the mock kubeclient ut should be able to completely replace this. 
But with this ut, volcano users now can test the core module performance of the scheduler and do some optimization works.If the implementation of the mock kubeclient ut cannot be completed in the short term(I think it may require some code modifications or refactors), would you consider accepting this ut first?

Monokaix:
> > > > The reason I don't directly construct a real cache through a fake kube client is because certain parts of the cache implementation use kubeconfig to create a new client (such as vcclient). These scenarios are difficult to use a fake kube client. Do you have any good ideas on how to handle this?
> > > 
> > > 
> > > Yes, I have some plan to do it. I will write a demo and then we can work together on that feature.
> > > I think we can keep this pr also. @wangyang0616 @Monokaix
> > 
> > 
> > I think there are two process that we should concern, the first one is schedule logic in every session, this mainly consists of predict and nodeOrder plugin, the second one is the kube-apiserver throughtput in real cluster, so your test can only test the first part and in my opion, we have no necessary to mock a cache becauese cache is composed of many interfaces, we just need mock these interfaces such as kube client : )
> > Also, kube-apiserver throughput has a great influence on scheduler performance, so we abstract a bind interface to enable user register their own bind method like bind a lots of pods in one bind request, so you can also work on this to achieve a higher performance.
> 
> @Monokaix Thanks for your review~ You are absolutely right that this unit test does not cover the batch binding part, and indeed, the implementation of the mock kubeclient unit test should be able to fully replace it.
> 
> However, with this unit test, Volcano users can now assess the performance of the scheduler's core module and carry out optimization work. If the implementation of the mock kubeclient unit test cannot be completed in the short term (which might involve some code modifications or refactoring), would you be open to considering this unit test in the interim?
> 
> Furthermore, the bottleneck in the second part is not only in the binding but also in the performance of Kubernetes, so it may not be entirely solvable with unit testing but might require end-to-end testing.

Yaeh, you can work on this ut and you are welcome, but can we not mock the whole cache and just use fake interface when we need mock. And if this feasible, you can have a try, if it's a little hard, you can try to wrap a `Cahce` and overwrite the method that hard to mock : )

bysph:
> > > > > The reason I don't directly construct a real cache through a fake kube client is because certain parts of the cache implementation use kubeconfig to create a new client (such as vcclient). These scenarios are difficult to use a fake kube client. Do you have any good ideas on how to handle this?
> > > > 
> > > > 
> > > > Yes, I have some plan to do it. I will write a demo and then we can work together on that feature.
> > > > I think we can keep this pr also. @wangyang0616 @Monokaix
> > > 
> > > 
> > > I think there are two process that we should concern, the first one is schedule logic in every session, this mainly consists of predict and nodeOrder plugin, the second one is the kube-apiserver throughtput in real cluster, so your test can only test the first part and in my opion, we have no necessary to mock a cache becauese cache is composed of many interfaces, we just need mock these interfaces such as kube client : )
> > > Also, kube-apiserver throughput has a great influence on scheduler performance, so we abstract a bind interface to enable user register their own bind method like bind a lots of pods in one bind request, so you can also work on this to achieve a higher performance.
> > 
> > 
> > @Monokaix Thanks for your review~ You are absolutely right that this unit test does not cover the batch binding part, and indeed, the implementation of the mock kubeclient unit test should be able to fully replace it.
> > However, with this unit test, Volcano users can now assess the performance of the scheduler's core module and carry out optimization work. If the implementation of the mock kubeclient unit test cannot be completed in the short term (which might involve some code modifications or refactoring), would you be open to considering this unit test in the interim?
> > Furthermore, the bottleneck in the second part is not only in the binding but also in the performance of Kubernetes, so it may not be entirely solvable with unit testing but might require end-to-end testing.
> 
> Yaeh, you can work on this ut and you are welcome, but can we not mock the whole cache and just use fake interface when we need mock. And if this feasible, you can have a try, if it's a little hard, you can try to wrap a `Cache` and overwrite the methods that hard to mock : )

OK,I will fix that later.

lowang-bh:
> Yes, I have some plan to do it. I will write a demo and then we can work together on that feature.

Hi, @bysph , I mocked the SchedulerCache in PR [add mock cache for ut and performance test #3269](https://github.com/volcano-sh/volcano/pull/3269). You can continue your work on this feature, maybe.

bysph:
> > Yes, I have some plan to do it. I will write a demo and then we can work together on that feature.
> 
> Hi, @bysph , I mocked the SchedulerCache in PR [add mock cache for ut and performance test #3269](https://github.com/volcano-sh/volcano/pull/3269). You can continue your work on this feature, maybe.

Okay, I will submit a UT based on this. Should I wait for the code to be merged before submitting?

PR #3263, opened 2023-12-14
[cherry-pick for release-1.8]fix: events miss when there is a pod bind failed

lowang-bh:
cherry-pick for [fix: events miss when there is a pod bind failed #2866](https://github.com/volcano-sh/volcano/pull/2866)


Monokaix:
Hi, it's not a serious issue and we have no necessary to cherry-pick to previous release: )

PR #3057, opened 2023-08-15
fix calculations of podgroup min resource

lowang-bh:
commit 1: refact jobinfo's calculation to a function.
commit 2: fix cal podgroup min resource and add testcase. relative design docs: [docs about job's min resource #2945](https://github.com/volcano-sh/volcano/pull/2945)
commit 3: when jobMinAvailable < totalTask's, keep the origin logic of calculate podgorup minResource: sum up first jobMinAvailable

Fix https://github.com/volcano-sh/volcano/issues/2921 also.

PR #3264, opened 2023-12-14
parse job min member info

lowang-bh:
split from #3057 (commit-1st: refact jobinfo's calculation to a function)

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -389,12 +389,7 @@ func (ji *JobInfo) SetPodGroup(pg *PodGroup) {
 	ji.RevocableZone = ji.extractRevocableZone(pg)
 	ji.Budget = ji.extractBudget(pg)
 
-	taskMinAvailableTotal := int32(0)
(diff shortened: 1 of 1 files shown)


Monokaix:
Why would we split it?

Thor-wl (commented):
May I ask why do we put this piece of code out of the main function? Personally speaking, it is short and readable enough.

lowang-bh:
> May I ask why do we put this piece of code out of the main function? Personally speaking, it is short and readable enough.

This is part of function to calculate job min resource (PR #3057). To let calculation function looks more compact.

But that PR seems can not be merged in few years, so I split the first small pr out. @Monokaix

Monokaix:
> > May I ask why do we put this piece of code out of the main function? Personally speaking, it is short and readable enough.
> 
> This is part of function to calculate job min resource (PR #3057). To let calculation function looks more compact.
> 
> But that PR seems can not be merged in few years, so I split the first small commit out. @Monokaix

I think we can concern the original pr directly: )

Issue #3240, opened 2023-12-01
`json: unsupported type: func() error` occurred when using scheduler extender


zhixian82:
**What happened**:
i register a scheduler extender to volcano. it works well in volcano version v1.7.0
and recently i upgrade my volcano to v1.8.1, when using scheduler extender，i found an error in the scheduler log: "json: unsupported type: func() error"

after studying the scheduler code，i found the error was occurred in the following line:
https://github.com/volcano-sh/volcano/blob/v1.8.1/pkg/scheduler/plugins/extender/extender.go#L166
https://github.com/volcano-sh/volcano/blob/v1.8.1/pkg/scheduler/plugins/extender/extender.go#L293

and then i found a func field was added to TaskInfo struct:
https://github.com/volcano-sh/volcano/blob/v1.8.1/pkg/scheduler/api/job_info.go#L136

and the Others field in the NodeInfo struct contains some func:
https://github.com/volcano-sh/volcano/blob/v1.8.1/pkg/scheduler/api/node_info.go#L346

and from[ json.Marshal](https://pkg.go.dev/encoding/json#Marshal)

```
Channel, complex, and function values cannot be encoded in JSON. Attempting to encode such a value causes Marshal to return an UnsupportedTypeError.
```

**Environment**:
- Volcano Version: v1.8.1
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
Yes, there is a compatibility need to fix. Also see issue https://github.com/volcano-sh/volcano/issues/3086

Issue #3256, opened 2023-12-11
extender.go:168 Predicate failed with error json: unsupported type: func() error


We1less1111:
I start my extender plugin, I found when json.Marshal(&PredicateRequest{Task: task, Node: node}) always with error like below.
Maybe when serialized , error can occur?
```
 extender.go:168] Predicate failed with error json: unsupported type: func() error.
######
json.Marshal(&PredicateRequest{Task: task, Node: node}) 
type TaskInfo struct {
        .....
	CustomBindErrHandler func() error
	....
}
```

Shiva953:
Yes. Functions in go are NOT serializable by default since they represent executable code rather than data.

One of the two changes might be relevant here

1. Either exclude the `CustomBindErrHandler` field by using the `json:"-"` tag: 

```go
CustomBindErrHandler func() error `json:"-"`
```

2. or use a serializable type like string:

```go
CustomBindErrHandlerInfo string `json:"customBindErrHandlerInfo"`
```
and then define CustomBindErrHandlerInfo accordingly.

We1less1111:
> Yes. Functions in go are NOT serializable by default since they represent executable code rather than data.
> 
> One of the two changes might be relevant here
> 
> 1. Either exclude the `CustomBindErrHandler` field by using the `json:"-"` tag:
> 
> ```go
> CustomBindErrHandler func() error `json:"-"`
> ```
> 
> 2. or use a serializable type like string:
> 
> ```go
> CustomBindErrHandlerInfo string `json:"customBindErrHandlerInfo"`
> ```
> 
> and then define `CustomBindErrHandlerInfo` accordingly.
> 
> I would like working on this issue btw.

I hope that the official can update this part of the code, otherwise extenders will be not supported??
```
pkg/scheduler/plugins/extender/extender.go
func (ep *extenderPlugin) OnSessionOpen(ssn *framework.Session) {
	if ep.config.onSessionOpenVerb != "" {
		err := ep.send(ep.config.onSessionOpenVerb, &OnSessionOpenRequest{
			Jobs:           ssn.Jobs,
			Nodes:          ssn.Nodes,
			Queues:         ssn.Queues,
			NamespaceInfo:  ssn.NamespaceInfo,
			RevocableNodes: ssn.RevocableNodes,
		}, nil)
.....
if ep.config.prioritizeVerb != "" {
[...]

PR #3162, opened 2023-10-22
support dump scheduler cache snapshot to json file

lowang-bh:
Fixes #3154

PR #3273, opened 2023-12-18
fix: json marsh error for unsupport type: func()

lowang-bh:
split from https://github.com/volcano-sh/volcano/pull/3162

Fixes https://github.com/volcano-sh/volcano/issues/3240

The original issue and background:
#3256

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -133,7 +133,7 @@ type TaskInfo struct {
 	Pod        *v1.Pod
 
 	// CustomBindErrHandler is a custom callback func called when task bind err.
-	CustomBindErrHandler func() error
+	CustomBindErrHandler func() error `json:"-"`
 	// CustomBindErrHandlerSucceeded indicates whether CustomBindErrHandler is executed successfully.
 	CustomBindErrHandlerSucceeded bool
 }


lowang-bh:
trigger ci

pkg/scheduler/api/cluster_info.go
```diff
@@ -26,8 +26,8 @@ type ClusterInfo struct {
 	Nodes          map[string]*NodeInfo
 	Queues         map[QueueID]*QueueInfo
 	NamespaceInfo  map[NamespaceName]*NamespaceInfo
-	RevocableNodes map[string]*NodeInfo
-	NodeList       []string
+	RevocableNodes map[string]*NodeInfo `json:"-"`
+	NodeList       []string             `json:"-"`
```
william-wang, line 30: why only change these two variables?

pkg/scheduler/api/job_info.go
```diff
@@ -133,7 +133,7 @@ type TaskInfo struct {
 	Pod        *v1.Pod
 
 	// CustomBindErrHandler is a custom callback func called when task bind err.
-	CustomBindErrHandler func() error
+	CustomBindErrHandler func() error `json:"-"`
```
william-wang, line 136: save as above

lowang-bh: func field can not be serialized.

pkg/scheduler/api/cluster_info.go
lowang-bh: The original PR is to support dump scheduler cache snapshot to json file. Those fields may not concerned by user,  so skip them.

Monokaix: I think we just need fix extender marshal err in this pr, and the changes here can be done by original PR and make some comments.

william-wang (approved):
/approve

Issue #3255, opened 2023-12-08
pods in queue's resource capacity will case share value to 1


lowang-bh:
As the picture described,  a queue has allocated pods will cause queue's share to be 1 always, although each of other resource has not used up.

If you're looking for a help then check our [Slack Channel](https://cloud-native.slack.com/messages/volcano) or have a look at our [dev mailing](https://groups.google.com/forum/#!forum/volcano-sh) 
-->

**What happened**:
![image](https://github.com/volcano-sh/volcano/assets/21003791/f23025b6-df61-4ba4-b2bb-5f59b0436d2b)
![image](https://github.com/volcano-sh/volcano/assets/21003791/2defdce8-2586-4218-b212-98d0c2a66e26)

**What you expected to happen**:

share value should less than 1

**How to reproduce it (as minimally and precisely as possible)**:

create a job in one queue, example yaml
```
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: queue-a
spec:
  capability:
    cpu: "4"
    memory: "2Gi"
---
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: job-high
spec:
  schedulerName: volcano
  priorityClassName: high-priority
  queue: queue-a
  tasks:
    - replicas: 1
      name: "default-nginx"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: nginx:1.14.2
              imagePullPolicy: IfNotPresent
              name: nginx
              resources:
                requests:
                  cpu: "3"
                  memory: "1Gi"
          restartPolicy: OnFailure
```

**Anything else we need to know?**:

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`): 1.27
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:
```
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "allocate, preempt, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
      - name: conformance
    - plugins:
      - name: overcommit
      - name: drf
        enablePreemptable: false
      - name: predicates
      - name: proportion
      - name: nodeorder
      - name: binpack
kind: ConfigMap
metadata:
  name: volcano-scheduler-configmap
  namespace: volcano-system
```

Monokaix:
Yep, that's a problem.
[...]

PR #3278, opened 2023-12-19
Ensure Pod Share Stays Below Resource Limits

jignyasamishra:
Fix: #3255

**Description:**
This update resolves an issue related to Pod Share calculation within the resource limits. The prior implementation allowed the Pod Share to exceed the available resource capacity, potentially leading to resource overutilization. The modified logic now ensures that the Pod Share is accurately computed based on the maximum resource utilization found among CPU, memory, and init containers. If this calculated share exceeds the available resources, the code clamps it to a maximum value of one (1) to maintain stability and prevent overconsumption of resources. This enhancement guarantees that Pod Share remains within the confines of the available resources, mitigating any potential resource allocation inconsistencies or overusage scenarios.

**Changes Made:**
Adjusted Pod Share calculation logic to consider maximum resource utilization.
Implemented checks to ensure the computed Pod Share remains within available resource limits.
Added validation to limit the Pod Share value to one (1) when necessary, preventing overutilization of resources.

### DIFF
--- a/pkg/scheduler/api/pod_info.go
+++ b/pkg/scheduler/api/pod_info.go
@@ -19,7 +19,7 @@ package api
 import (
 	"encoding/json"
 	"strconv"
-
+	"math"
 	v1 "k8s.io/api/core/v1"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


Monokaix:
Welcome and thanks for your contribution！Will check this soon: )

pkg/scheduler/api/pod_info.go
```diff
@@ -61,15 +61,36 @@ import (
...
+
+	// Adjust the share based on CPU, Memory, and maximum init container utilization
+	maxUtilization := math.Max(cpuSum, memorySum)
+	maxUtilization = math.Max(maxUtilization, initContainerMax)
+	if maxUtilization > 0 {
+		podShare := maxUtilization / 1000 // Convert to fractional value
+		if podShare > 1 {
+			podShare = 1
+		}
+		result.AddScalar(v1.ResourcePods, podShare)
```
Monokaix, line 87: Aha, seems it's misunderstanding, we assume the value 1 as scaler resource, which represents pod itself occupied a pod resource and this is used in preemption scenario, here the value should always be 1, and we should fix it in proportion plugin instead: )

jignyasamishra: Thank you much for your review!!!! I am working on it. I thought we need to adjust the resources accordingly. Can you provide me some more hints on the issue.

PR #3279, opened 2023-12-20
Chart: add configurable expressions for webhooks namespace selector

Aakcht:
This PR adds possibility to specify namespace selector for admission webhooks in helm chart.

### DIFF
--- a/installer/helm/chart/volcano/templates/webhooks.yaml
+++ b/installer/helm/chart/volcano/templates/webhooks.yaml
@@ -24,6 +24,9 @@ webhooks:
           values:
             - {{ .Release.Namespace }}
             - kube-system
+{{- if .Values.custom.webhooks_namespace_selector_expressions }}
+        {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
+{{- end }}
     objectSelector: {}
     reinvocationPolicy: Never
     rules:
@@ -66,6 +69,9 @@ webhooks:
           values:
             - {{ .Release.Namespace }}
             - kube-system
+{{- if .Values.custom.webhooks_namespace_selector_expressions }}
+        {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
+{{- end }}
     objectSelector: {}
     reinvocationPolicy: Never
     rules:
@@ -108,6 +114,9 @@ webhooks:
           values:
             - {{ .Release.Namespace }}
             - kube-system
+{{- if .Values.custom.webhooks_namespace_selector_expressions }}
+        {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
+{{- end }}
     objectSelector: {}
     reinvocationPolicy: Never
     rules:
@@ -150,6 +159,9 @@ webhooks:
           values:
             - {{ .Release.Namespace }}
             - kube-system
+{{- if .Values.custom.webhooks_namespace_selector_expressions }}
+        {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
+{{- end }}
     objectSelector: {}
     reinvocationPolicy: Never
     rules:
@@ -192,6 +204,9 @@ webhooks:
           values:
             - {{ .Release.Namespace }}
             - kube-system
+{{- if .Values.custom.webhooks_namespace_selector_expressions }}
+        {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
+{{- end }}
     objectSelector: {}
     rules:
       - apiGroups:
@@ -234,6 +249,9 @@ webhooks:
           values:
             - {{ .Release.Namespace }}
             - kube-system
+{{- if .Values.custom.webhooks_namespace_selector_expressions }}
+        {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
+{{- end }}
     objectSelector: {}
     rules:
       - apiGroups:
@@ -275,6 +293,9 @@ webhooks:
           values:
             - {{ .Release.Namespace }}
             - kube-system
+{{- if .Values.custom.webhooks_namespace_selector_expressions }}
+        {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
+{{- end }}
     objectSelector: {}
     rules:
       - apiGroups:
--- a/installer/helm/chart/volcano/values.yaml
+++ b/installer/helm/chart/volcano/values.yaml
@@ -93,3 +93,23 @@ custom:
   admission_resources: ~
   scheduler_resources: ~
   controller_resources: ~
+
+# Specify additional namespace selector expressions for Volcano admission webhooks.
+# For example, if you want Volcano admission webhooks take effect in namespaces with
+# label key="workload-type" and value="batch", and don't take effect in namespaces with
+# label key="kubernetes.io/metadata.name" and values "excluded-ns-1" and "excluded-ns-2",
+# you should set the following value:
+#
+#  webhooks_namespace_selector_expressions:
+#    - key: workload-type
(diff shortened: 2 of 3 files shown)


installer/helm/chart/volcano/templates/webhooks.yaml
```diff
@@ -17,13 +17,18 @@ webhooks:
     failurePolicy: Fail
     matchPolicy: Equivalent
     name: mutatepod.volcano.sh
+{{- if .Values.custom.webhooks_namespace_selector }}
```
Monokaix, line 20: If user set custom label, and pod in `Release.Namespace` will not be filtered, this is unacceptable.

Aakcht: @Monokaix you think that it would be better if `Release.Namespace` would always be hardcoded? I guess it's a good idea to prevent accidental cluster break :D 

I'll try changing parameter to only specify additional namespaceSelector expressions with the expression 
```
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - {{ .Release.Namespace }}
            - kube-system
```
always present then.

Monokaix: Yes, you are right.

Aakcht: Updated

Aakcht:
@Monokaix please take a look, I've made the change, `{{ .Release.Namespace }}` will now always be ignored.

lowang-bh:
Hi, @Aakcht , we need to squash those commits to only one.

Aakcht:
@lowang-bh please take a look, I've squashed the commits.

installer/helm/chart/volcano/values.yaml
```diff
@@ -93,3 +93,16 @@ custom:
   admission_resources: ~
   scheduler_resources: ~
   controller_resources: ~
+
+  # Specify additional namespace selector expressions for Volcano admission webhooks.
+  # For example:
+  #
+  #  webhooks_namespace_selector_expressions:
```
Monokaix, line 100: It seems that this is the final expression, can you add example about how to add additional lable value `on webhooks_namespace_selector_expressions`?

Aakcht: Updated

Aakcht:
@Monokaix m please take a look, I've updated the example. I don't think something more extended is needed here cause I don't want to overload values.yaml with examples.

installer/helm/chart/volcano/values.yaml
```diff
@@ -93,3 +93,21 @@ custom:
   admission_resources: ~
   scheduler_resources: ~
   controller_resources: ~
+
+  # Specify additional namespace selector expressions for Volcano admission webhooks.
```
Monokaix, line 97: Please change the expression to this:
```
  # Specify additional namespace selector expressions for Volcano admission webhooks.
  # For example, if you want Volcano admission webhooks take effect in namespaces with label key="workload-type" and value="batch", 
  # and don't take effect in namespaces with label key="kubernetes.io/metadata.name" and values are "exluded-ns-1" and "excluded-ns-2",
  # you should set the following value:
```

Aakcht: @Monokaix updated

Aakcht:
@Monokaix please take a look and tell me if anything else is needed.

Monokaix:
Thanks for your contribution and a last question, whey set `nindent 8` and can you paste your validation result with `helm template`?

Aakcht:
@Monokaix thanks for reviewing!

I did another rebase - somehow I missed that there was unnecessary space in the comment in values.yaml (compared to other comments in values.yaml), so I removed it.

As for your questions:

1) nindent 8 is needed to add 8 spaces before added expression, the 8 is needed since the expression
```yaml
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - {{ .Release.Namespace }}
            - kube-system
```
which is always present, also have 8 spaces. So it's just for formatting. (the same as for example custom labels fo scheduler https://github.com/volcano-sh/volcano/blob/master/installer/helm/chart/volcano/templates/scheduler.yaml#L111 .)

2) With `webhooks_namespace_selector_expressions: ~` the output of `helm template` looks the same as before the changes (there is actually a check for it, I think). Here how the output of `helm template` looks with `webhooks_namespace_selector_expressions` set to example value (attaching only one object since namespaceSelector in other looks the same):
```yaml
# Source: volcano/templates/webhooks.yaml
apiVersion: admissionregistration.k8s.io/v1
metadata:
  name: volcano-admission-service-podgroups-mutate
webhooks:
  - admissionReviewVersions:
      - v1
    clientConfig:
      service:
        name: RELEASE-NAME-admission-service
        namespace: default
        path: /podgroups/mutate
        port: 443
    failurePolicy: Fail
    matchPolicy: Equivalent
    name: mutatepodgroup.volcano.sh
    namespaceSelector:
      matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
            - default
            - kube-system
        - key: workload-type
          operator: In
          values:
          - batch
        - key: kubernetes.io/metadata.name
          operator: NotIn
          values:
          - excluded-ns-1
          - excluded-ns-2
    objectSelector: {}
    reinvocationPolicy: Never
    rules:
      - apiGroups:
          - scheduling.volcano.sh
        apiVersions:
          - v1beta1
        operations:
          - CREATE
        resources:
          - podgroups
        scope: '*'
    sideEffects: NoneOnDryRun
    timeoutSeconds: 10
```

Monokaix:
> @Monokaix thanks for reviewing!
> 
> I did another rebase - somehow I missed that there was unnecessary space in the comment in values.yaml (compared to other comments in values.yaml), so I removed it.
> 
> As for your questions:
> 
> 1. nindent 8 is needed to add 8 spaces before added expression, the 8 is needed since the expression
> 
> ```yaml
>         - key: kubernetes.io/metadata.name
>           operator: NotIn
>           values:
>             - {{ .Release.Namespace }}
>             - kube-system
> ```
> 
> which is always present, also have 8 spaces. So it's just for formatting. (the same as for example custom labels fo scheduler https://github.com/volcano-sh/volcano/blob/master/installer/helm/chart/volcano/templates/scheduler.yaml#L111 .)
> 
> 2. With `webhooks_namespace_selector_expressions: ~` the output of `helm template` looks the same as before the changes (there is actually a check for it, I think). Here how the output of `helm template` looks with `webhooks_namespace_selector_expressions` set to example value (attaching only one object since namespaceSelector in other looks the same):
> 
> ```yaml
> # Source: volcano/templates/webhooks.yaml
> apiVersion: admissionregistration.k8s.io/v1
> kind: MutatingWebhookConfiguration
> metadata:
>   name: volcano-admission-service-podgroups-mutate
> webhooks:
>   - admissionReviewVersions:
>       - v1
>     clientConfig:
>       service:
>         name: RELEASE-NAME-admission-service
>         namespace: default
>         path: /podgroups/mutate
>         port: 443
>     failurePolicy: Fail
>     matchPolicy: Equivalent
>     name: mutatepodgroup.volcano.sh
>     namespaceSelector:
>       matchExpressions:
>         - key: kubernetes.io/metadata.name
>           operator: NotIn
>           values:
>             - default
>             - kube-system
>         - key: workload-type
>           operator: In
>           values:
>           - batch
>         - key: kubernetes.io/metadata.name
>           operator: NotIn
>           values:
>           - excluded-ns-1
>           - excluded-ns-2
>     objectSelector: {}
>     reinvocationPolicy: Never
>     rules:
>       - apiGroups:
>           - scheduling.volcano.sh
>         apiVersions:
>           - v1beta1
>         operations:
>           - CREATE
>         resources:
>           - podgroups
>         scope: '*'
>     sideEffects: NoneOnDryRun
>     timeoutSeconds: 10
> ```

Thanks!

william-wang (approved):
/approve

Issue #3242, opened 2023-12-01
Queue capacity scheduling


Monokaix:
#### What would you like to be added:

Add capacity scheduling capability for queue

#### Why is this needed:
Volcano Proportion plugin provides capacity scheduling and fair scheduling between multiple queues by weight, and the Queue minimal resource (guaranteed resource) and maximal resource (capacity) is supported as well. Users run heterogeneous clusters with varying resource types (i.e. different types of GPU cards and network interfaces, and require Currently the user requirerequires fine-grained resource sharing and preemption between queues, and it is expected to configure the `deserved resource` using resourceType scalar rather than `weight` (a simple percentage of cluster resource) for each queue. 

#### Task list
- [x] Add scheduling design doc. https://github.com/volcano-sh/volcano/pull/3277
- [x] API change in https://github.com/volcano-sh/apis https://github.com/volcano-sh/apis/pull/121
- [x] capacity scheduling implementation in this repo. https://github.com/volcano-sh/volcano/pull/3283

Monokaix:
All have done!

PR #3283, opened 2023-12-28
Capacity plugin implement

Monokaix:
Third part of https://github.com/volcano-sh/volcano/issues/3242.
This pr should be merged after https://github.com/volcano-sh/apis/pull/121.


pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -95,6 +95,10 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			klog.V(3).Infof("Queue <%s> is overused, ignore it.", queue.Name)
 			continue
 		}
+		if !ssn.Preemptive(queue) {
```
lowang-bh, line 98: Actually, I think it should consider the resource demension. Eg, queue is overused on GPU resource, but CPU/MEM is not overused and incomeing job just need cpu/mem, we can still reclaim from other queues to get CPU/MEM back.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		klog.V(4).InfoS("Victims from capacity plugin", "victims", victims, "reclaimer", reclaimer)
+		return victims, util.Permit
+	})
+
+	ssn.AddPreemptiveFn(cp.Name(), func(obj interface{}) bool {
+		queue := obj.(*api.QueueInfo)
+		attr := cp.queueOpts[queue.UID]
+
+		canReclaim := !attr.deserved.LessEqual(attr.allocated, api.Zero)
```
lowang-bh, line 227: Mybe all dimentions resource is large than deserved can be regarded as not reclaimable.

Monokaix: Yeah，you are right.

pkg/scheduler/actions/reclaim/reclaim.go
Monokaix: > Actually, I think it should consider the resource demension. Eg, queue is overused on GPU resource, but CPU/MEM is not overused and incomeing job just need cpu/mem, we can still reclaim from other queues to get CPU/MEM back.

Updated.

lowang-bh:
I am a litte confused. 

The `Guarantee`  can be treated as `min` resource which is the guaranteed resource of a queue, and can be configured by user. 
and `capacity` is  the `max` resource a tenant can use. 

If a user wants a configurable `min` resource, `Guarantee`  is ok, just disable `proportion` plugin and enable this `capacity` plugin(I think they can not be used at the same time, right?).

So Why adding another field `Deserved` to `queue` spec in https://github.com/volcano-sh/apis/pull/121. 

And what's more, the meaning of `Deserved` is not consistent with the comments `This part of resource can be shared with other queues and reclaimed back`

If it is a field representing a resource `can be shared with other queues and reclaimed back`,  `elastic` or other word is more proper.

BTW, this plugin can not be used with proportion plugin, am I right?

Monokaix:
> I am a litte confused.
> 
> The `Guarantee` can be treated as `min` resource which is the guaranteed resource of a queue, and can be configured by user. and `capacity` is the `max` resource a tenant can use.
> 
> If a user wants a configurable `min` resource, `Guarantee` is ok, just disable `proportion` plugin and enable this `capacity` plugin(I think they can not be used at the same time, right?).
> 
> So Why adding another field `Deserved` to `queue` spec in [volcano-sh/apis#121](https://github.com/volcano-sh/apis/pull/121).
> 
> And what's more, the meaning of `Deserved` is not consistent with the comments `This part of resource can be shared with other queues and reclaimed back`
> 
> If it is a field representing a resource `can be shared with other queues and reclaimed back`, `elastic` or other word is more proper.
> 
> BTW, this plugin can not be used with proportion plugin, am I right?

It's a replace of `proportion `plugin and can not use with `proportion `plugin at the same time, `proportion `plugin divided resources by weight while `capacity `plugin divided resources directly by deserved, which is more intuitive, you can learn more in design doc https://github.com/volcano-sh/volcano/blob/master/docs/design/capacity-scheduling.md.

lowang-bh:
> t's a replace of `proportion `plugin and can not use with `proportion `plugin at the same time, `proportion `plugin divided resources by weight while `capacity `plugin divided resources directly by deserved, which is more intuitive, you can learn more in design doc https://github.com/volcano-sh/volcano/blob/master/docs/design/capacity-scheduling.md.

I see it. thanks.

william-wang (approved):
/lgtm

PR #3284, opened 2023-12-28
fix wrong pods field format output of queue status

Monokaix:
fix wrong pods field format output of queue status
before:
image
after:
image


william-wang (approved):
/approve

PR #3292, opened 2024-01-05
Refactor Execute() of Allocate Method into Multiple Subfunctions for Enhanced Readability and Maintainability

daniel-hutao:
## Summary

This pull request refactors a large core function into multiple subfunctions. The primary goal is to enhance the code's readability and maintainability.

This work is not yet fully completed, but considering the significant amount of changes already made, I hope to merge the current changes first to facilitate an easier review process. I plan to continue further optimization subsequently.

This PR is divided into two commits, with each commit presenting relatively clear changes. I would kindly ask the reviewers to consider reviewing the changes commit by commit, as this approach may be more manageable.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -29,7 +29,9 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
-type Action struct{}
+type Action struct {
+	session *framework.Session
+}
 
 func New() *Action {
 	return &Action{}
@@ -57,6 +59,14 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	// jobsMap is used to find job with the highest priority in given queue.
 	jobsMap := map[api.QueueID]*util.PriorityQueue{}
 
+	alloc.session = ssn
+	alloc.pickUpQueuesAndJobs(queues, jobsMap)
+	klog.V(3).Infof("Try to allocate resource to %d Queues", len(jobsMap))
+	alloc.allocateResources(queues, jobsMap)
+}
+
+func (alloc *Action) pickUpQueuesAndJobs(queues *util.PriorityQueue, jobsMap map[api.QueueID]*util.PriorityQueue) {
+	ssn := alloc.session
 	for _, job := range ssn.Jobs {
 		// If not config enqueue action, change Pending pg into Inqueue statue to avoid blocking job scheduling.
 		if conf.EnabledActionMap["enqueue"] {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -57,6 +57,15 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	// jobsMap is used to find job with the highest priority in given queue.
 	jobsMap := map[api.QueueID]*util.PriorityQueue{}
 
+	alloc.prepareQueuesAndJobs(queues, jobsMap, ssn)
```
Monokaix, line 60: We can rename it something like `pickUpQueuesAndJobs`

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -164,123 +156,146 @@ func (alloc *Action) Execute(ssn *framework.Session) {
...
-			if !ssn.Allocatable(queue, task) {
-				klog.V(3).Infof("Queue <%s> is overused when considering task <%s>, ignore it.", queue.Name, task.Name)
-				continue
-			}
+func (alloc *Action) processTasks(tasks *util.PriorityQueue, job *api.JobInfo, jobs *util.PriorityQueue, queue *api.QueueInfo, ssn *framework.Session, allNodes []*api.NodeInfo) {
+	stmt := framework.NewStatement(ssn)
+	ph := util.NewPredicateHelper()
 
-			klog.V(3).Infof("There are <%d> nodes for Job <%v/%v>", len(ssn.Nodes), job.Namespace, job.Name)
+	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
```
Monokaix, line 167: Before this modification, predicateFn just declared once, it's declared repeatedly now, we should keep declare it once.

pkg/scheduler/actions/allocate/allocate.go
daniel-hutao: good idea

pkg/scheduler/actions/allocate/allocate.go
daniel-hutao: yeah

daniel-hutao:
@Monokaix Please take a look again.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -90,29 +100,13 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 		klog.V(4).Infof("Added Job <%s/%s> into Queue <%s>", job.Namespace, job.Name, job.Queue)
 		jobsMap[job.Queue].Push(job)
 	}
+}
 
-	klog.V(3).Infof("Try to allocate resource to %d Queues", len(jobsMap))
-
+func (alloc *Action) allocateResources(queues *util.PriorityQueue, jobsMap map[api.QueueID]*util.PriorityQueue) {
```
Monokaix, line 105: This method does two things, firstly pickUpTasks and allocate task secondly, so rename this method name to pickUpTasks  and rename processTasks to allocateResourcesForTasks, and both make them as a dependent method seperately is better, just be consistent with `alloc.pickUpQueuesAndJobs`.

daniel-hutao: The main logic of the `allocateResources` method is in a **for loop**, and a longer part of the logic within the loop has been separated into `processTasks`. I attempted to decouple these two methods and have them both called by the `Execute` method, but this failed.

Moving the for loop logic into `Execute` essentially means removing the `allocateResources` method, which would significantly complicate `Execute`. Therefore, I think it's not suitable to rename `allocateResources` to `pickUpTasks`.

As for renaming `processTasks`, I am open to it. If you think a change would be better, I can add another commit to complete this renaming.

Monokaix: Got that.

Thor-wl (approved):

PR #3294, opened 2024-01-09
Enhancements to Comments and Code Adjustments in pkg/scheduler

daniel-hutao:
This pull request introduces some updates to the `pkg/scheduler` package. The primary focus of these updates is to improve the overall readability and clarity of the codebase.

Key changes include:

- **Comprehensive Comment Revision**: Updated existing comments for better clarity and added missing comments to undocumented code segments.
- **Code Refinements**: Made minor code adjustments to enhance readability and maintain consistency throughout the package.

These changes are intended to make the codebase more approachable and understandable for both current contributors and new developers joining the project.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -17,10 +17,10 @@ limitations under the License.
 package scheduler
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
@@ -28,8 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
-
-	"volcano.sh/volcano/pkg/util"
 
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/filewatcher"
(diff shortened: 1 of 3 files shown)


lowang-bh:
why change `Scheduler` to `scheduler`? This will cause lint error `exported func NewScheduler returns unexported type *scheduler.scheduler, which can be annoying to use`.

daniel-hutao:
> why change `Scheduler` to `scheduler`? This will cause lint error `exported func NewScheduler returns unexported type *scheduler.scheduler, which can be annoying to use`.

I changed `Scheduler` to `scheduler` mainly because this structure doesn't need to be accessed in other packages, so changing it to non-exposed makes it look clearer to the caller (who won't see the non-exposed properties and methods).

The more general way to write this is to return an interface in the `NewScheduler()` method, and then the interface's implementation structure starts in lowercase.

But since the current `Scheduler` is relatively simple, I didn't intentionally add an interface. Overall, I feel that it's a tradeoff, all 3 ways of writing are acceptable, and I personally feel that `scheduler + no interface` is the most appropriate.

If you all think the `Scheduler` writeup is best, I can change it back.

Thor-wl (commented):
It's generally OK to me. Regarding with the `scheduler` change, let's think more about the static check of lint. Thanks for your contribution.

daniel-hutao:
To avoid causing too much controversy over this minor issue, I have reverted it back.

william-wang (approved):
/approve

PR #3303, opened 2024-01-17
Switch hostPath to emptyDir

### LINKED ISSUE #3298: Replace `hostPath` with `emptyDir`or PVC in Scheduler Deployment

Currently the Scheduler Deployment utilizes a `hostPath` to setup a volume for /tmp/klog-socks:

https://github.com/volcano-sh/volcano/blob/b834fa8939f6ecbb8e96d80a5f1151323956dfdc/installer/helm/chart/volcano/templates/scheduler.yaml#L178C19-L178C19

https://github.com/volcano-sh/volcano/blob/b834fa8939f6ecbb8e96d80a5f1151323956dfdc/installer/volcano-development.yaml#L4117

Instead of utilizing `hostPath` the deployments could instead utilize an `emptyDir` or a PVC.

#### Why is this needed:

As the official k8s documentation states, `hostPath` presents a security risk and should only be considered if absolutely necessary:

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

As far as I can tell Volcano is only writing logs to that location and it does not fit the recommended use cases outlined in the k8s docs.

`emptyDir` is a great option for a temporary storage location that can be discarded any time the pod restarts.  Since this volume is writing to `/tmp` I would assume that it is not critical to keep these logs.

A PVC would be more ideal if the logs do need to persist across the pod restarts.

strangiato:
Resolves #3298

hostPath should be avoided unless absolutely necessary as per the upstream k8s docs:

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

This change swaps the hostPath for an emptyDir allowing the tmp logs to be written to an ephemeral location.

`emptyDir` is preferable to not using a volume at all in case a users environment enforces Read Only Root Filesystems.

Since this log is already being written to `/tmp` it doesn't seem like there is a requirement to persist these logs using a PVC.

### DIFF
--- a/installer/helm/chart/volcano/templates/scheduler.yaml
+++ b/installer/helm/chart/volcano/templates/scheduler.yaml
@@ -175,8 +175,7 @@ spec:
           configMap:
             name: {{ .Release.Name }}-scheduler-configmap
         - name: klog-sock
-          hostPath:
-            path: /tmp/klog-socks
+          emptyDir: {}
 ---
 apiVersion: v1
 kind: Service
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -4114,8 +4114,7 @@ spec:
           configMap:
             name: volcano-scheduler-configmap
         - name: klog-sock
-          hostPath:
-            path: /tmp/klog-socks
+          emptyDir: {}
 ---
 # Source: volcano/templates/scheduling_v1beta1_podgroup.yaml
 apiVersion: apiextensions.k8s.io/v1


Monokaix:
Hi, please also execute `make update-development-yaml` to make install/volcano-development.yaml upated too. And use git commit -s to submit your pr.

strangiato:
Signed the commit.

I also went ahead and ran `make update-development-yaml`.  It did not generate any changes since I manually updated the `volcano-development.yaml` file.

Monokaix:
> Signed the commit.
> 
> I also went ahead and ran `make update-development-yaml`. It did not generate any changes since I manually updated the `volcano-development.yaml` file.

Understood: )

Monokaix:
Hi, seems DCO verification still not passed, pleae have a look.

kumarankit999 (approved):
Now emptyDir has been on the place of hostPath

Monokaix:
Hi, are you still working on this?

schauaib:
Hi, what's up here ?

kzgrzendek:
Hello, as the last answer by the author as been made more than a year ago, I'd be happy to submit a PR regarding this issue if that's ok @Monokaix

Monokaix:
This pr is not active, fixed by https://github.com/volcano-sh/volcano/pull/4124

PR #3304, opened 2024-01-18
Fix ineffective volcano scheduler deployment string filter

Monokaix:
The default e2e kind cluster delpoy volcano-scheduler as name `integration-scheduler-xxx`, the current e2e case code can't select volcano scheduler correctly, this pr fix it.


william-wang (approved):
/lgtm

PR #3321, opened 2024-02-18
update go to 1.21

hwdef:
Volcano has adapted to k8s1.29 and k8s1.30 will be released soon. We should update the go language version to be consistent with k8s


Monokaix:
We'd better add some info like volcano has adapted k8s 1.29 and k8s has used go 1.21 and so we should keep consistent it. Also please update docker build base image to 1.21.

hwdef:
> We'd better add some info like volcano has adapted k8s 1.29 and k8s has used go 1.21 and so we should keep consistent it. Also please update docker build base image to 1.21.

Thanks for your suggestions. I fixed this

william-wang (approved):

PR #3339, opened 2024-03-06
Karmada

Monokaix:


pkg/scheduler/cache/cache_test.go
```diff
@@ -344,3 +355,121 @@ func TestBindTasks(t *testing.T) {
...
+		sc          *SchedulerCache
+		handlerName string
+		creator     CustomResourceEventHandlerCreator
+		trigger     func(gvr schema.GroupVersionResource, client dynamic.Interface) error
+		want        map[schema.GroupVersionResource]map[string]schedulingapi.CustomResource
+	}{
+		{
+			name:        "custom resource add",
+			handlerName: "test",
+			sc: &SchedulerCache{
```
lowang-bh, line 436: How about use `NewDefaultMockSchedulerCache` to trigger the mock list/watch process?

pkg/scheduler/cache/cache.go
```diff
@@ -519,27 +527,31 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, defaultQueu
 		nodeQueue:           workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
 		DeletedJobs:         workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
 		kubeClient:          kubeClient,
+		dynClient:           dynamic.NewForConfigOrDie(config),
```
lowang-bh, line 530: also need init fake client in `newMockSchedulerCache`.

pkg/scheduler/cache/cache_test.go
Monokaix: It's not a proper way here, because custom clinet and cache is a independent part, and it needs to provide parameters separately, and NewDefaultMockSchedulerCache  is a general function for test.

pkg/scheduler/cache/cache.go
Monokaix: same as above.

hwdef:
Is this still need? @Monokaix

PR #3340, opened 2024-03-09
Fix Issue 2262: add priority capability for reclaim action

### LINKED ISSUE #2262: added priority capability for reclaim action
We need a resource recovery requirement between queues. The scenario is as follows:

We have an offline job management platform, which is responsible for managing offline jobs of different business teams, and all business teams submit jobs on this platform.

The platform has a global job priority. it creates the queue for each business team and sets the maximum available capacity(queue.capacity). and it has an operator responsible for creating jobs and scheduling them with volcano.

The platform has two responsibilities: 
    1) improve the resource utilization rate of the platform, 
    2) allow more high-quality jobs to run. 
we hope to achieve our goal through the following two operations:

1. sum (queue.capacity) > the total resources of the cluster, for example: the cluster can allocate 100 cpus, there are three queues, each queue is 50 cpus, sum(queue)=150 > 100( cluster can allocate resources)
2. relcaim enabled,when the capacity of a queue does not reach the queue.deserved,  and the cluster has no idle resources to allocate, the high-priority jobs in this queue can recycle low-priority jobs in other queues, but the same level cannot be reclaimed.

We see that relcaim has no priority, resource recycling happens between jobs of the same level.

So, is it possible to add an AddReclaimableFn to the priority plugin?

E.g:
priority.go
```
ssn.AddReclaimableFn(pp.Name(), preemptableFn)
```

lowang-bh:
Fixes #2262

1. priority plugin support `reclaimable` switch and add UT about priority plugin
2. priority plugin docs

**Note**: `reclaimableFn` usually is used in `reclaim` action to reclaim a queue's deserved resource when cluster has not enough resource to allocate new coming tasks in this queue. So please be carefully to set `enableReclaimable` to `true` in `priority` plugin, in case that a queue's resource owned by high priority jobs can not be released. And `enableReclaimable` is disabled by default for compatibility

### DIFF
--- a/installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
+++ b/installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
@@ -2,6 +2,7 @@ actions: "enqueue, allocate, backfill, reclaim, preempt"
 tiers:
 - plugins:
   - name: priority
+    enableReclaimable: false
   - name: gang
     enablePreemptable: false
   - name: conformance
--- a/pkg/scheduler/plugins/priority/priority.go
+++ b/pkg/scheduler/plugins/priority/priority.go
@@ -112,6 +112,25 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 		return victims, util.Permit
 	}
 	ssn.AddPreemptableFn(pp.Name(), preemptableFn)
+	reclaimbleFn := func(reclaimor *api.TaskInfo, reclaimee []*api.TaskInfo) ([]*api.TaskInfo, int) {
+		reclaimorJob := ssn.Jobs[reclaimor.Job]
+		var victims []*api.TaskInfo
+		for _, reclaimee := range reclaimee {
+			reclaimeeJob := ssn.Jobs[reclaimee.Job]
+			if reclaimeeJob == nil {
+				continue
+			}
+			if reclaimorJob.Priority > reclaimeeJob.Priority {
+				victims = append(victims, reclaimee)
+			} else {
+				klog.V(4).Infof("Can not reclaim task <%v/%v> because reclaimee job has greater or equal job priority (%d) than reclaimor (%d)",
+					reclaimee.Namespace, reclaimee.Name, reclaimeeJob.Priority, reclaimorJob.Priority)
+			}
+		}
+		klog.V(4).Infof("Victims from Priority plugins are %+v", victims)
+		return victims, util.Permit
+	}
+	ssn.AddReclaimableFn(pp.Name(), reclaimbleFn)
 
 	jobStarvingFn := func(obj interface{}) bool {
 		ji := obj.(*api.JobInfo)
--- a/pkg/scheduler/plugins/priority/priority.go
+++ b/pkg/scheduler/plugins/priority/priority.go
@@ -112,10 +112,10 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 		return victims, util.Permit
 	}
 	ssn.AddPreemptableFn(pp.Name(), preemptableFn)
-	reclaimbleFn := func(reclaimor *api.TaskInfo, reclaimee []*api.TaskInfo) ([]*api.TaskInfo, int) {
+	reclaimableFn := func(reclaimor *api.TaskInfo, reclaimees []*api.TaskInfo) ([]*api.TaskInfo, int) {
 		reclaimorJob := ssn.Jobs[reclaimor.Job]
 		var victims []*api.TaskInfo
-		for _, reclaimee := range reclaimee {
+		for _, reclaimee := range reclaimees {
 			reclaimeeJob := ssn.Jobs[reclaimee.Job]
 			if reclaimeeJob == nil {
 				continue
@@ -130,7 +130,7 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 		klog.V(4).Infof("Victims from Priority plugins are %+v", victims)
 		return victims, util.Permit
 	}
-	ssn.AddReclaimableFn(pp.Name(), reclaimbleFn)
+	ssn.AddReclaimableFn(pp.Name(), reclaimableFn)
 
 	jobStarvingFn := func(obj interface{}) bool {
 		ji := obj.(*api.JobInfo)
--- a/docs/design/preempt-action.md
+++ b/docs/design/preempt-action.md
@@ -18,7 +18,7 @@ In preempt action, multiple plugin function are getting used like
 2.  JobOrderFn(Plugin: Priority, DRF, Gang),
 3.  NodeOrderFn(Plugin: NodeOrder),
 4.  PredicateFn(Plugin: Predicates),
-5.  PreemptableFn(Plugin: Conformance, Gang, DRF).
+5.  PreemptableFn(Plugin: Conformance, Gang, DRF, Priority).
 
 ### 1. TaskOrderFn:
 #### Priority:
--- a/docs/design/priority.md
+++ b/docs/design/priority.md
@@ -1,0 +1,39 @@
+# Priority Plugin
+
+## Introduction
+
+When users apply jobs to volcano, they may need to support different priority to different kind of jobs. For example, important serving jobs can have a higher priority when allocating resources, or preempting other jobs in same queue. Even in the same job, the important roles may have a higher priority than others roles, so that it can finished running without interruption.
+
+## Solution
+
+1. `priority` plugin return 5 callback functions: `taskOrderFn`, `jobOrderFn`, `preemptableFn`, `reclaimableFn`, and `jobStarvingFn`:
+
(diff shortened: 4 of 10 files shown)


Monokaix:
If other queue's jobs all have higher priority，current queue can not  reclaim their resources, reclaim will not happen, is this reasonable？

lowang-bh:
> If other queue's jobs all have higher priority，current queue can not reclaim their resources, reclaim will not happen, is this reasonable？

We just support this feature and close it by default。 How to use it is depended on the cluster admin.

One solutions is to limit the higher priority jobs used resource not exceed queue's deserved in application layer.

Monokaix:
> > If other queue's jobs all have higher priority，current queue can not reclaim their resources, reclaim will not happen, is this reasonable？
> 
> We just support this feature and close it by default。 How to use it is depended on the cluster admin.
> 
> One solutions is to limit the higher priority jobs used resource not exceed queue's deserved in application layer.

But this seems places high demands on administrators and limits the job priority of the queue, we'd better add a desige doc and give some user-guide.

lowang-bh:
> we'd better add a desige doc and give some user-guide.

@Monokaix docs is added.  
cc @wangyang0616 @hwdef  please help to review. thanks.

pkg/scheduler/plugins/priority/priority.go
```diff
@@ -112,6 +112,25 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
 		return victims, util.Permit
 	}
 	ssn.AddPreemptableFn(pp.Name(), preemptableFn)
+	reclaimableFn := func(reclaimor *api.TaskInfo, reclaimee []*api.TaskInfo) ([]*api.TaskInfo, int) {
+		reclaimorJob := ssn.Jobs[reclaimor.Job]
+		var victims []*api.TaskInfo
+		for _, reclaimee := range reclaimee {
```
hwdef, line 118: Variable names are ambiguous.

lowang-bh: > Variable names are ambiguous.

Thanks. Have fixed.

pkg/scheduler/plugins/priority/priority.go
```diff
@@ -112,6 +112,25 @@ func (pp *priorityPlugin) OnSessionOpen(ssn *framework.Session) {
...
 	ssn.AddPreemptableFn(pp.Name(), preemptableFn)
+	reclaimableFn := func(reclaimor *api.TaskInfo, reclaimees []*api.TaskInfo) ([]*api.TaskInfo, int) {
+		reclaimorJob := ssn.Jobs[reclaimor.Job]
+		var victims []*api.TaskInfo
+		for _, reclaimee := range reclaimees {
+			reclaimeeJob := ssn.Jobs[reclaimee.Job]
+			if reclaimeeJob == nil {
+				continue
+			}
+			if reclaimorJob.Priority > reclaimeeJob.Priority {
```
hwdef, line 123: why double `e`?

lowang-bh: I think it is like "Employee" vs "Employer"， so does "preemptor" vs "preemptee"

hwdef (commented):
/lgtm

installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
```diff
@@ -2,6 +2,7 @@ actions: "enqueue, allocate, backfill, reclaim, preempt"
 tiers:
 - plugins:
   - name: priority
+    enableReclaimable: false
```
Monokaix, line 5: How about set this false default？

lowang-bh: how to？Add a param for priority plugin and default not to register  `AddReclaimableFn`?

lowang-bh: It already set `enableReclaimable` as false in installing yaml.

Monokaix:
It conflicts with the existing allocation logic between queues. When we allocate tasks, we do not consider the priority of jobs between queues, but only consider the priority at the queue level.

zhoushuke:
I think there may be a bug in Reclaimable in session_plugins.go, for this configmap
```yaml
    tiers:
    - plugins:
      - name: priority
      - name: proportion
 ```
as reclaim enabled in  priority plugin, let us image that theres some victims return from priority, but if  no victims return from proportion plugin , Reclaimable function will return no victims, and reclaim not work any more, put proportion in front of priority can fix this, but  it may be a bug?  @hwdef

**Comment by lowang-bh** [comment]
> I think there may be a bug in Reclaimable in session_plugins.go, for this configmap
> 
> ```yaml
>     tiers:
>     - plugins:
>       - name: priority
>       - name: proportion
> ```
> 
> as reclaim enabled in priority plugin, let us image that theres some victims return from priority, but if no victims return from proportion plugin , Reclaimable function will return no victims, and reclaim not work any more, put proportion in front of priority can fix this, but it may be a bug? @hwdef

That is a problem of your config. You'd better to put plugins about resource in a same tier, but a tier different from gang/priority, eg:
```yaml
tiers:
- plugins:
  - name: priority
    enableReclaimable: false
  - name: gang
    enablePreemptable: false
  - name: conformance
- plugins:
  - name: overcommit
  - name: drf
    enablePreemptable: false
  - name: predicates
  - name: proportion
  - name: nodeorder
  - name: binpack
```

zhoushuke:
> > I think there may be a bug in Reclaimable in session_plugins.go, for this configmap
> > ```yaml
> >     tiers:
> >     - plugins:
> >       - name: priority
> >       - name: proportion
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > as reclaim enabled in priority plugin, let us image that theres some victims return from priority, but if no victims return from proportion plugin , Reclaimable function will return no victims, and reclaim not work any more, put proportion in front of priority can fix this, but it may be a bug? @hwdef
> 
> That is a problem of your config. You'd better to put plugins about resource in a same tier, but a tier different from gang/priority, eg:
> 
> ```yaml
> tiers:
> - plugins:
>   - name: priority
>     enableReclaimable: false
>   - name: gang
>     enablePreemptable: false
>   - name: conformance
> - plugins:
>   - name: overcommit
>   - name: drf
>     enablePreemptable: false
>   - name: predicates
>   - name: proportion
>   - name: nodeorder
>   - name: binpack
> ```

ok, make sense

zhoushuke:
@lowang-bh , there is another promble as following:

```yaml
tiers:
    - plugins:
      - name: priority
      - name: gang
        enablePreemptable: false
        enableJobStarving: false
        enableReclaimable: false
        enabledQueueScoreOrder: false
      - name: conformance
    - plugins:
      - name: predicates
      - name: proportion
 ```
gang enablePreemptable: false( make it true also don't work too), so it's cant do any preempt/reclaim, when a high priority job comes, if there no enough resource to meet gang constraint, I hope get resources by reclaim, but as not meet gang constraint, this job will pending, and reclaim action is just skip pending job, so reclaim not happend, no reclaim, no release resource to get job running, like deadlock, delete enqueue may work, but in my scene，enqueue is a must, what better soluation to fix it?

lowang-bh:
@zhoushuke There are two kind of evicts:
1. evicts in same queue: this is for high priority job to evict low priority jobs when your queue's deserved resource is used up
2. evicts between different queues:  a queue's coming job to reclaim their queue's deserved resource from other overused queue. In this scene, the overused resources in other queues are borrowed from this queue, and resource should be give back when need.

zhoushuke:
I have some volcano practice and use it in a production to support about  30k~50k pods a day. 
I think you don't get my point, anyway, I will try to realize it by my own.

Monokaix:
> enablePreemptable: false

So can you explain why set enablePreemptable: false of gang？

Monokaix:
> > If other queue's jobs all have higher priority，current queue can not reclaim their resources, reclaim will not happen, is this reasonable？
> 
> We just support this feature and close it by default。 How to use it is depended on the cluster admin.
> 
> One solutions is to limit the higher priority jobs used resource not exceed queue's deserved in application layer.

what about reclaim lower priority first, if no enough resourecs reclaimed then reclaim higher priority task?

lowang-bh:
> this job will pending, and reclaim action is just skip pending job

If you didn't use enqueue action, the default status will be pending and allocate action will change it to enqueue when there is no enqueue action configured. So you should put reclaim action after allocate.

JesseStutler:
We should consider higher priority jobs in other queues can never be reclaimed if enabledReclaimable is enabled in Priority Plugin.  The original design intention of reclaim should be to be able to reclaim jobs in other queues, without considering the priority, right?

lowang-bh:
> We should consider higher priority jobs in other queues can never be reclaimed if enabledReclaimable is enabled in Priority Plugin. The original design intention of reclaim should be to be able to reclaim jobs in other queues, without considering the priority, right?

Yes，docs has highlight the point that should be carefully to open it.

hwdef:
still need

PR #3341, opened 2024-03-09
fix issue 2957: predicate plugin ut didn't really run due to kubeclient init failed

### LINKED ISSUE #2957: UT didn't really run because of kubeclient init failed
**What happened**:
https://github.com/volcano-sh/volcano/blob/9427174e0eefc18d321d34ba844fe3f80ff491a5/pkg/scheduler/plugins/predicates/predicates_test.go#L68-L71

image

UT should use k8s's fakeclient

add log before return
```go
	klog.Errorf("exiting test...")
```

all cases in predicate and nodeorder plugin which need k8s client

**Environment**:
- Volcano Version: latest
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lowang-bh:
Fix #2957(UT didn't really run because of kubeclient init failed)

image

### DIFF
--- a/pkg/scheduler/plugins/predicates/predicates_test.go
+++ b/pkg/scheduler/plugins/predicates/predicates_test.go
@@ -1,28 +1,20 @@
 package predicates
 
 import (
-	"reflect"
 	"testing"
-
-	"github.com/agiledragon/gomonkey/v2"
-	"github.com/spf13/pflag"
 
 	apiv1 "k8s.io/api/core/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/tools/record"
 
 	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
-	"volcano.sh/volcano/cmd/scheduler/app/options"
-	"volcano.sh/volcano/pkg/kube"
 	"volcano.sh/volcano/pkg/scheduler/actions/allocate"
 	"volcano.sh/volcano/pkg/scheduler/api"
-	"volcano.sh/volcano/pkg/scheduler/cache"
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/drf/hdrf_test.go
```diff
@@ -252,7 +252,7 @@ func TestHDRF(t *testing.T) {
 
 		binder := &util.FakeBinder{
 			Binds:   map[string]string{},
-			Channel: make(chan string),
```
Monokaix, line 255: Why change to 300?

pkg/scheduler/plugins/predicates/predicates_test.go
```diff
@@ -84,97 +57,41 @@ func TestEventHandler(t *testing.T) {
...
 	// tests
-	tests := []struct {
-		name     string
-		pods     []*apiv1.Pod
-		nodes    []*apiv1.Node
-		pcs      []*schedulingv1.PriorityClass
-		pgs      []*schedulingv1beta1.PodGroup
-		expected map[string]string
-	}{
+	tests := []uthelper.TestCommonStruct{
```
Monokaix, line 74: every test case will initialize a new scheduler cache, but the origin test has a global schdeuler cache, is there any effect?

pkg/scheduler/actions/allocate/allocate_test.go
```diff
@@ -323,7 +323,7 @@ func TestAllocateWithDynamicPVC(t *testing.T) {
 			fakeVolumeBinder := util.NewFakeVolumeBinder(kubeClient)
 			binder := &util.FakeBinder{
 				Binds:   map[string]string{},
-				Channel: make(chan string),
```
Monokaix, line 326: Why change to 10？

pkg/scheduler/plugins/drf/hdrf_test.go
lowang-bh: To avoid blocking binding in FakeBinder. Please review PR#3343 first.

pkg/scheduler/actions/allocate/allocate_test.go
lowang-bh: same as above.

pkg/scheduler/plugins/predicates/predicates_test.go
lowang-bh: No effect. There is just one case now.

william-wang (approved):
/approve

PR #3342, opened 2024-03-09
refact drf plugin ut

### LINKED ISSUE #2810: Part of the ut shows that the operation is successful, but the test case is not actually executed

lowang-bh:
Fixes #2810 

Some UT actually does not run.

image

### DIFF
--- a/pkg/scheduler/plugins/drf/hdrf_test.go
+++ b/pkg/scheduler/plugins/drf/hdrf_test.go
@@ -1,25 +1,20 @@
 package drf
 
 import (
-	"flag"
 	"fmt"
 	"reflect"
 	"testing"
 
-	"github.com/agiledragon/gomonkey/v2"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/tools/record"
-	"k8s.io/klog/v2"
 
 	schedulingv1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/scheduler/actions/allocate"
 	"volcano.sh/volcano/pkg/scheduler/api"
-	"volcano.sh/volcano/pkg/scheduler/cache"
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/plugins/proportion"
+	"volcano.sh/volcano/pkg/scheduler/uthelper"
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/drf/hdrf_test.go
```diff
@@ -49,82 +52,42 @@ type pgSpec struct {
...
-				},
+			TestCommonStruct: uthelper.TestCommonStruct{
+				Name:    "rescaling test",
+				Plugins: plugins,
+				PodGroups: []*schedulingv1.PodGroup{
+					util.BuildPodGroup("pg1", "default", "root-sci", 0, nil, schedulingv1.PodGroupInqueue),
+					util.BuildPodGroup("pg21", "default", "root-eng-dev", 0, nil, schedulingv1.PodGroupInqueue),
+					util.BuildPodGroup("pg22", "default", "root-eng-prod", 0, nil, schedulingv1.PodGroupInqueue),
+				},
+				Pods: mergePods(
```
Monokaix, line 71: Is this method truly needed? We can declare a slice directly.

lowang-bh: How？each makePods returns a list, and I need to merge them to one list. Append looks more complex

Monokaix: OK

william-wang (approved):
/approve

PR #3343, opened 2024-03-09
feat: add ut common framework

lowang-bh:
Add a common UT package to do UTs.

Nowadays, some UTs are not actually runned, as issue #2810 #2957 describe.
And, what's more, the lib `gomonkey` has no permit to run on MacOS and is not convenient to do UTs for developers.
Now we can ut this package to refact all other package's UTs.

### DIFF
--- a/pkg/scheduler/uthelper/helper.go
+++ b/pkg/scheduler/uthelper/helper.go
@@ -1,0 +1,234 @@
+/*
+Copyright 2024 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


pkg/scheduler/uthelper/helper.go
```diff
@@ -0,0 +1,234 @@
+/*
+Copyright 2023 The Volcano Authors.
```
Monokaix, line 2: Please change to 2024.

lowang-bh: Done

william-wang (approved):
/approve

PR #3352, opened 2024-03-15
fix panic in overcommit plugin

guoqinwill:
fix panic in overcommit plugin when the gpu plugin is uninstalled or the gpu card fails, triggering a continuous panic。
fix：https://github.com/volcano-sh/volcano/issues/3362

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -260,6 +260,16 @@ func (r *Resource) Sub(rr *Resource) *Resource {
 	return r.sub(rr)
 }
 
+// SubWithoutAssert subtracts two Resource objects without assertion,
+// this function is added because some resource subtraction allows negative results, while others do not.
+func (r *Resource) SubWithoutAssert(rr *Resource) *Resource {
+	ok, resources := rr.LessEqualWithResourcesName(r, Zero)
+	if !ok {
+		klog.Errorf("resources <%v> are not sufficient to do operation: <%v> sub <%v>", resources, r, rr)
+	}
+	return r.sub(rr)
+}
+
 // sub subtracts two Resource objects.
 func (r *Resource) sub(rr *Resource) *Resource {
 	r.MilliCPU -= rr.MilliCPU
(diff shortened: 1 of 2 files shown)


pkg/scheduler/api/resource_info.go
```diff
@@ -260,6 +260,14 @@ func (r *Resource) Sub(rr *Resource) *Resource {
 	return r.sub(rr)
 }
 
+// WithoutAssertSub subtracts two Resource objects without assertion.
+func (r *Resource) WithoutAssertSub(rr *Resource) *Resource {
+	if !rr.LessEqual(r, Zero) {
```
Monokaix, line 265: Use `LessEqualWithResourcesName` is better to explain which dimension resource is not sufficient.

guoqinwill: done

pkg/scheduler/api/resource_info.go
```diff
@@ -260,6 +260,15 @@ func (r *Resource) Sub(rr *Resource) *Resource {
 	return r.sub(rr)
 }
 
+// WithoutAssertSub subtracts two Resource objects without assertion.
+func (r *Resource) WithoutAssertSub(rr *Resource) *Resource {
```
wangyang0616, line 264: It would be better to explain why such a function is added. For example, some resource subtraction allows negative results, while others do not.

Monokaix, line 264: `SubWithoutAssert` is better.

pkg/scheduler/api/resource_info.go
```diff
@@ -260,6 +260,15 @@ func (r *Resource) Sub(rr *Resource) *Resource {
...
+	ok, resources := rr.LessEqualWithResourcesName(r, Zero)
+	if !ok {
+		klog.Errorf("resources <%v> are not sufficient to do operation: <%v> sub <%v>", resources, r, rr)
```
Monokaix, line 267: Here also need to `sub`.

guoqinwill: The judgment here is just to print the log, and the sub operation is executed uniformly when it finally return.

pkg/scheduler/api/resource_info.go
guoqinwill: done

guoqinwill: done

pkg/scheduler/api/resource_info.go
Monokaix: Ok

william-wang (approved):
/lgtm

PR #3359, opened 2024-03-20
fix rollback unthoroughly when allocate error

bibibox:
fix issue #3361


wangyang0616:
Can you register an issue and describe the scenario that triggers the issue in actual production?

bibibox:
> Can you register an issue and describe the scenario that triggers the issue in actual production?

yes, this is the issue: #3361

william-wang (approved):
/approve

Issue #3329, opened 2024-02-26
with preempt or reclaim plugin, the high priority pod can not be placed at some node which meet the conditions for preemption


LivingCcj:
when volcano scheduler open preempt or reclaim plugin，the high prioriry pod is unable to preempt the low priority pod. Although there are some nodes that meet the preemption conditions，beacuse one function in these predicateFns return err (is not nil), the potential node  will be ignore
https://github.com/volcano-sh/volcano/blob/94c62a4b73c71bd263e8ae770c694343b6007380/pkg/scheduler/actions/preempt/preempt.go#L211-L221

**Environment**:
- Volcano Version: volcano v1.8.2
- Kubernetes version (use `kubectl version`): v1.20.15

lowang-bh:
Would you please supply more informations, such as scheduler configmap, and scheduler logs?

LivingCcj:
when preempt or reclaim, if  one predicate function handler return status with `Unschedulable` state，but the `err` is not nil, it will  ignore the  potential node. According to the comments `Allows scheduling to nodes that are in Success or Unschedulable state after filtering by predicate` in function `predicateFn `, the node could be preempted
here is the volcano scheduler configmap：
```
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, preempt, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
        enablePreemptable: false
      - name: conformance
      - name: cdp
    - plugins:
      - name: drf
        enablePreemptable: false
      - name: predicates
      - name: nodeorder
        arguments:
          nodeaffinity.weight: 5
      - name: binpack
        arguments:
          binpack.weight: 5
          binpack.cpu: 2
          binpack.memory: 1
kind: ConfigMap
metadata:
  name: volcano-scheduler-configmap
  namespace: volcano-system
```

lowang-bh:
> but the `err` is not nil

What is the error?

LivingCcj:
There is a scene, an unscheduled pod with gpu resources is in the session of  `preempt` action.  one node have some lower priority pods，if it preempts the lower priority pod，the unscheduled pod could be placed at the node. However at predicate stage, the predicateStatus.Code is `Unschedulable` and err is not nil (refer to the code below). It leads to ignoring the potential node when filtering preempted nodes.
[...]

PR #3363, opened 2024-03-22
fix preempt & reclaim shecduler plugin with PreemptPredicateFn

LivingCcj:
fix issue: [https://github.com/volcano-sh/volcano/issues/3329](https://github.com/volcano-sh/volcano/issues/3329)


Monokaix:
The url in description seems not exists: )

LivingCcj:
> The url in description seems not exists: )

thx for your reply, now has updated this url

lowang-bh:
I don't recommend this changed. But I also have no good idea now -_-

lowang-bh:
we need to add some uts to cover it.

PR #3364, opened 2024-03-24
refact reclaim ut with common test framework

lowang-bh:
Fixes part of #2810

Refact uts：
1. use the ut framework build on mockSchedulerCache to unify all the uts；
2. remove gomonkey which has no permit to be run on MACOS

image


hwdef (commented):
/lgtm

william-wang (approved):
/approve

PR #3379, opened 2024-04-01
record error if task's jobID is null when deleting task

lowang-bh:
If task's jobID is null, task will not deleted from cache and has no log information.

This PR add some error info when it run into this case, although it should not happen.


william-wang (approved):
/approve

PR #3388, opened 2024-04-06
Network Topology Aware Plugin

### LINKED ISSUE #447: Support zone aware scheduling

/kind feature

**Description**:

The nodes in the cluster maybe from different zones; there maybe several benefits if all tasks of the job were scheduled to the same zone, e.g. low latency. So it's better for Volcano to support zone aware scheduling :)

### LINKED ISSUE #2984: Batch node allocation for AI training

for ai training scenario, one job might take multiple GPUs and hope the allocated GPUs affinized in same node to use NVLink high speed intra-node connection. For example, for one job which needs 24 GPUs and there will be 24 pods and each pods take one GPU. There are multiple nodes in the cluster and each node has 8 A100 GPUs.

The job hopes to get any 3 nodes and takes all the GPUs of the nodes. Is there good way for volcano to allocate them?

### LINKED ISSUE #3317: Add pod affinity group scheduling capability, to improve the performance of AI training jobs.
1. The affinityGroupSize field is added to the task definition of the Volcano job to describe the number of pods that require affinity scheduling. The following is an example:
kind: Job
metadata:
  name: test
spec:
  schedulerName: volcano
  minAvailable: 4
  plugins:
    task-topology: ["--affinity", "[[ps-task, worker-task]]", "--anti-affinity", "[[ps-task]]"] 
  tasks:
    - replicas: 16
      name: "ps-task"
      **_affinityGroupSize: 4   (# Number of pods for affinity scheduling)_**
      ....
2. The label "**_volcano.sh/affinity-node-group_**" is added to the node to identify the nodes that are in the same physical chassis or have higher link performance.
#### Why is this needed:
In AI training scenarios, the interconnection bandwidth between all NPUs in a physical chassis is higher. When performing distributed training on a large model, fully aligning a distributed parallel policy with a node topology structure can greatly improve performance of large model training. The scheduler allows users to specify the affinity group configuration required by the job when creating an AI job, so that the job can be allocated to the node affinity group in the same physical chassis.

lowang-bh:
/kind feature
fixes #2984 
fixes #447
fixes #3317
There are several issues request this feature, such as #447 #2984 #3317

## Motivation
We target to make scheduler net-topology aware so as to achieve the following:

- best effort to schedule same job to same topology devices, such as same idc.

## Goals

- Support single key topology configuration, try to schedule job's all tasks to nodes which have same value with that key
- Support multiple-key topology policies, the key at front get higher score

## Non-Goals

- Not to find the global solutions among nodes with all kind values of that key

### DIFF
--- a/docs/design/network-topology-aware.md
+++ b/docs/design/network-topology-aware.md
@@ -1,0 +1,90 @@
+# Network Topology Aware Plugin
+
+- [Network Topology Aware Plugin](#network-topology-aware-plugin)
+  - [Backgrounds](#backgrounds)
+  - [Motivation](#motivation)
+  - [Proposal one](#proposal-one)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+    - [Design Action](#design-action)
+      - [Pod scheduling process](#pod-scheduling-process)
+      - [Usage](#usage)
+    - [Drawbacks](#drawbacks)
+
+## Backgrounds
+
+A Kubernetes cluster typically comprises numerous nodes distributed across different IDCs, chassis, and switches.
+
+Data transformations vary in performance across these different components.
+
+For latency-sensitive workloads, it's crucial to execute tasks within the same IDC and ideally on the same chassis and switch.
+
+## Motivation
+
+The goal is to make the Kubernetes scheduler network-topology aware to achieve the following:
+
+Ensure optimal scheduling of tasks from the same job onto nodes within the same topology, such as the same IDC, chassis, or switch.
+
+There will be two types of network-topology aware
+
+- **static**: `network-topology.type: static` is aiming to aware the network topology by nodes' labels
+- **dynamic**: `network-topology.type: dynamic` is aiming to use some tools to detect the network topology dynamically. For example, `ibnetdiscover` can be used to discover the InfiniBand network topology
+
+## Proposal one
+
+This proposal requires cluster administrators to manage network topology labels on Kubernetes (K8s) nodes.
+
+Nodes can be labeled to indicate identical topologies with the same label value.
+
+### Goals
+
+- **Single-Key Topology Configuration**: Support scheduling all tasks of a job onto nodes that share the same value for a specified key.
+- **Multiple-Key Topology Policies**: Prioritize keys listed earlier for better scheduling preference.
+
+### Non-Goals
+
+- **Global Solutions**: This proposal does not aim to find solutions across nodes with all possible values of a topology key simultaneously.
+
+### Design Action
+
+#### Pod scheduling process
+
+1. **Recording Topology Information**: When the first task of a job is assigned to a node, record the node's topology information in the scheduling plugin.
+2. **Scoring Nodes for Subsequent Tasks**: During scheduling of subsequent tasks, nodes with the same topology as the initially allocated task receive a higher score; others receive a score of zero.
+3. **Handling Multiple Keys**: If a node matches multiple keys from the configured list, the first key in the list is prioritized for scoring.
+
+```go
+nodeOrderFn := func(task *api.TaskInfo, node *api.NodeInfo) (float64, error){
+    ...
+    score := 0
+    weight := np.weight
+    tlabels := tNode.Node.Labels
+    labels := node.Node.Labels
+    lenth := len(np.topologyKeys)
+    for i, key := range np.topologyKeys {
+        if tlabels[key] == labels[key] {
+            score += (lenth - i) //  key with more priority at front of which with less priority
+            break
+        }
+    }
+    return float64(score * weight), nil
+}
+```
+
+#### Usage
+
+1. Label nodes with key-value pairs (e.g., `switch=NvLink-A100`, `rack=rack1,rack2`, `idc=bj,sh`) to partition nodes into different topology zones
+2. Add the `network-topology` plugin in the scheduler configuration to implement these policies.
+
+```yaml
+- plugins:
+   - name: network-topology
+     arguments:
+       network-topology.type: static # static means it will use the node's labels to aware network topology
+       network-topology.keys: rack,switch,idc # required when type is static
+       network-topology.weight: 10
+```
+
+### Drawbacks
+
+One drawback is that it's not a global solution that ensures all tasks of a job are placed on nodes within the same topology. For example, if nodes labeled with key-value1 lack sufficient resources while nodes labeled with key-value2 have them, and the first task is assigned to key-value1 nodes, subsequent tasks will still attempt to use key-value1 nodes, despite the resource constraints.
--- a/pkg/scheduler/plugins/factory.go
+++ b/pkg/scheduler/plugins/factory.go
@@ -26,6 +26,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/plugins/drf"
 	"volcano.sh/volcano/pkg/scheduler/plugins/extender"
 	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
+	nettopology "volcano.sh/volcano/pkg/scheduler/plugins/network-topology"
 	"volcano.sh/volcano/pkg/scheduler/plugins/nodegroup"
 	"volcano.sh/volcano/pkg/scheduler/plugins/nodeorder"
 	"volcano.sh/volcano/pkg/scheduler/plugins/numaaware"
@@ -56,6 +57,7 @@ func init() {
 	framework.RegisterPluginBuilder(overcommit.PluginName, overcommit.New)
 	framework.RegisterPluginBuilder(sla.PluginName, sla.New)
 	framework.RegisterPluginBuilder(tasktopology.PluginName, tasktopology.New)
+	framework.RegisterPluginBuilder(nettopology.PluginName, nettopology.New)
 	framework.RegisterPluginBuilder(numaaware.PluginName, numaaware.New)
 	framework.RegisterPluginBuilder(cdp.PluginName, cdp.New)
 	framework.RegisterPluginBuilder(rescheduling.PluginName, rescheduling.New)
--- a/pkg/scheduler/plugins/network-topology/netaware.go
+++ b/pkg/scheduler/plugins/network-topology/netaware.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2024 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package networktopology
+
+import (
+	"strings"
+
+	"k8s.io/klog/v2"
+
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+)
+
+// PluginName indicates name of volcano scheduler plugin.
+const (
+	PluginName            = "network-topology"
+	networkTopologyWeight = "network-topology.weight"
+	networkTopologyType   = "network-topology.type" // strategy type for network topology
+	// strategy value for network-topology.type
+	staticAware  = "static"
+	dynamicAware = "dynamic"
+)
+
+type netTopPlugin struct {
+	pluginArguments framework.Arguments
+	weight          int            // network-topology plugin score weight
+	topologyType    string         // supported topology type: static, dynamic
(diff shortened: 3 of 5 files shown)


hwdef:
we should talk this in weekly meeting.

Thor-wl:
Sounds interesting！But it's maybe more complex than the given desgin. For example, network delay varies between different nodes. Also, it varies in different period to the same node. Maybe considered with network performance metrics for this feature will be a good choice. I think we should take a discussion in the community and complete the desgin first.

Monokaix:
What's the difference between this and https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/networkaware？

lowang-bh:
> Maybe considering with network performance metrics for this feature will be a good choice.

I would like to recommend to treat the network performace  metrics as a kind of load, so it's more like a loadaware scheduling.

> network delay varies between different nodes

Now it is not considered. This plugin just conserder the physical difference in topology.

lowang-bh:
> https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/networkaware？

I think it is a lite one of that plugin,  just consider several physical topology, such as idc, rock, switch, and depend on the those labels on nodes.

advantage: more simply to use, just rely on node labels
shortcoming：no bandwidth, no latency, etc.

Monokaix:
> > [https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/networkaware？](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/networkaware%EF%BC%9F)
> 
> I think it is a lite one of that plugin, just consider several physical topology, such as idc, rock, switch, and depend on those labels on nodes.
> 
> advantage: more simply to use, just rely on node labels shortcoming：no bandwidth, no latency, etc.

We should collect more user cases: )

docs/design/net-aware.md
```diff
@@ -0,0 +1,64 @@
...
+## Design Action
+
+### Pod scheduling process
+
+1. when the first task of a job is allocated to a node, record the node information in the plugin
+2. when scheduling other tasks, a node with same key-value as the target node, get a higher score. Otherwise, get a zero score.
+3. If a node has multiple keys same as the configured list, the first key matching the configured keys has higher score
+
+```go
+nodeOrderFn := func(task *api.TaskInfo, node *api.NodeInfo) (float64, error){
```
Monokaix, line 32: From the original user demand, job should occupy nodes execulicm the original user demand, job should occupy nodes exclusively, so only nodeorder func seems cannot satisfy the original use case.

lowang-bh: You mean  a predicateFn is need?

yeahdongcn:
We have some thoughts on network topology aware scheduling as well, `BUT` limited in the same IDC.

## User Story

Saying we have a 10,000 GPU cluster and each node is connected to a leaf switch through IB or RoCE. Each leaf switch connects to spine switches and there might be multiple paths between leaf switches. The network topology is a fat-tree.

End users would like to submit an LLM training job to the cluster. Typically, the job is an MPI job and the communication between nodes is frequent. In most cases, the communication between nodes in the same leaf switch is faster than the communication between nodes in different leaf switches (fewer jumps, less latency).

End users would like to schedule the tasks to the nodes that have optimal network connectivity at the moment. For example, if the tasks are scheduled to the nodes that are connected to the same leaf switch, the communication between the tasks is faster and the job can finish earlier.

We want to schedule the tasks to the nodes that are connected to the same leaf switch, or at least to the nodes that are connected to the same spine switch. This can reduce the network latency and improve the performance.

## Design

We can add a new plugin to the scheduler to support network topology-aware scheduling. The plugin can get the network topology information from the network devices and use the information to schedule the tasks.

### Network Topology Information

The network topology information can be stored in a configmap. The information includes the network topology, the network devices, the connections between the devices, and the latency between the devices.

Similar to how to pick the topology-aware nodes in Slurm, we can use certain tools to discover the network topology alive. For example, we can use `ibnetdiscover` to discover the InfiniBand network topology.

We can have another controller to watch the node changes and spin up a job to discover the network topology on the nodes with IB. The controller can update the network topology information in the configmap.

### Network Topology Aware Plugin

The network topology-aware plugin can get the network topology information from the configmap and use the information to schedule the tasks. It works like a score plugin to score the nodes based on the network topology information and pick the best node for a task.

## Initial Code Change

Please see: https://github.com/yeahdongcn/volcano/tree/topo

This is a PoC to show how to add a network topology-aware plugin to the scheduler. See `https://github.com/yeahdongcn/volcano/blob/topo/pkg/scheduler/actions/allocate/allocate_test.go#L160` for the test case.

@shinytang6 for awareness.

docs/design/net-aware.md
```diff
@@ -0,0 +1,64 @@
...
+### Usage
+
+1. label nodes with some key-value pairs, for example switch=NvLink-A100, rack=rack1,rack2, idc=bj,sh to partition nodes to different topology zones.
+2. add net-topology plugin in scheduler-configuration
+
+```yaml
+- plugins:
+   - name: net-topology
+     arguments:
+       net-topology.keys: rack,switch,idc
```
Monokaix, line 58: Is there any other topology needs affinity except network?

Monokaix: The key should has a format like `failure-domain.beta.kubernetes.io/zone` or `volcano.sh/xxx`, and user can cofigure the key according their cluster's node configuration.

lowang-bh: Those keys just is an example. Real keys depend on user's real cluster configuration.

pkg/scheduler/api/job_info.go
```diff
@@ -152,7 +152,10 @@ func getTaskID(pod *v1.Pod) TaskID {
 	if ts, found := pod.Annotations[batch.TaskSpecKey]; found && len(ts) != 0 {
 		return TaskID(ts)
 	}
-
+	// keep searching pod labels, as it is also added in labels in job controller
+	if ts, found := pod.Labels[batch.TaskSpecKey]; found && len(ts) != 0 {
```
googs1025, line 156: Why do we need to add labels in addition to annotations?

lowang-bh: The code's comments described why.

wangyizhi1:
When will this PR be merged?

pkg/scheduler/plugins/net-topology/netaware.go
```diff
@@ -0,0 +1,107 @@
...
+	// strategy value for net-topology.type
+	staticAware  = "static"
+	dynamicAware = "dynamic"
+)
+
+type netTopPlugin struct {
+	pluginArguments framework.Arguments
+	weight          int            // net-topology plugin score weight
+	topolgyType     string         // supported topology type: static, dynamic
+	staticTopAware  staticTopAware // use node labels to generate topology
```
wangyizhi1, line 42: topolgyType => togologyType ?

wangyizhi1:
The first task returning a score of 0 for all nodes might cause issues. I suggest comparing the remaining nodes in each topology to see if they meet the job's requirements and assign a corresponding score. This score should be proportional to the maximum number of tasks that can be accommodated.

Moreover, the current implementation of this plugin resembles a greedy algorithm, which aims to find the optimal node for each task. However, the greedy algorithm doesn't necessarily yield the optimal solution. I'm curious about how dynamic programming or backtracking could be implemented within the Volcano framework. Is there a way to perform multiple pre-scheduling attempts for a job and apply the one with the highest total score?

lowang-bh:
> Moreover, the current implementation of this plugin resembles a greedy algorithm, which aims to find the optimal node for each task. However, the greedy algorithm doesn't necessarily yield the optimal solution. I'm curious about how dynamic programming or backtracking could be implemented within the Volcano framework. Is there a way to perform multiple pre-scheduling attempts for a job and apply the one with the highest total score?

Yes,  current one is just try the best to find a locally optimal solution and is a simple realization. There will be and official release for a global optimal solution.

Monokaix:
It's a great feature and we ca

> > Moreover, the current implementation of this plugin resembles a greedy algorithm, which aims to find the optimal node for each task. However, the greedy algorithm doesn't necessarily yield the optimal solution. I'm curious about how dynamic programming or backtracking could be implemented within the Volcano framework. Is there a way to perform multiple pre-scheduling attempts for a job and apply the one with the highest total score?
> 
> Yes, current one is just try the best to find a locally optimal solution and is a simple realization. There will be and official release for a global optimal solution.

+1
If we implement it using dynamic programming  algorithm, it will be more efficient, but maybe we need modify the allocate action framework, which changes a lot.

Monokaix:
The best way is to use dynamic programming algorithm, but this needs a new design, which may beyond the scope of the issue, v.1.10 is about to release, so this feature can be released v1.11, even it's not a best way, but we can continue to optimize the algorithm: )

pkg/scheduler/plugins/net-topology/netaware.go
lowang-bh: changed to topologyType.

william-wang:
> Similar to how to pick the topology-aware nodes in Slurm, we can use certain tools to discover the network topology alive. For example, we can use `ibnetdiscover` to discover the InfiniBand network topology.
> 
> We can have another controller to watch the node changes and spin up a job to discover the network topology on the nodes with IB. The controller can update the network topology information in the configmap.
> 
> ### Network Topology Aware Plugin
> The network topology-aware plugin can get the network topology information from the configmap and use the information to schedule the tasks. It works like a score plugin to score the nodes based on the network topology information and pick the best node for a task.
> 
> ## Initial Code Change
> Please see: https://github.com/yeahdongcn/volcano/tree/topo
> 
> This is a PoC to show how to add a network topology-aware plugin to the scheduler. See `https://github.com/yeahdongcn/volcano/blob/topo/pkg/scheduler/actions/allocate/allocate_test.go#L160` for the test case.
> 
> If anyone is interested in this topic, we can have further discussion.
> 
> @shinytang6 for awareness.

`ibnetdiscover` is a tool to detect the IB networking, how about RoCE networking detecting? @yeahdongcn

yeahdongcn:
> `ibnetdiscover` is a tool to detect the IB networking, how about RoCE networking detecting? @yeahdongcn

I'm currently working on using LLDP for comprehensive topology detection. 

Since there’s no central management tool like `ibnetdiscover` in RoCE, we need to collect all LLDP information from each node and switch to aggregate these details and build a complete Slurm configuration.

william-wang:
> > `ibnetdiscover` is a tool to detect the IB networking, how about RoCE networking detecting? @yeahdongcn
> 
> I'm currently working on using LLDP for comprehensive topology detection.
> 
> Since there’s no central management tool like `ibnetdiscover` in RoCE, we need to collect all LLDP information from each node and switch to aggregate these details and build a complete Slurm configuration.

Good to know that. I am investating the network aware scheduling these days as well. I think we can have a meeting to talks more about the requirement and design to cover more cases. @lowang-bh @yeahdongcn

hwdef:
I've been using @yeahdongcn solution in a production environment for a while and it's working fine so far

OguzPastirmaci:
I have the same requirement for improving the placement/locality of GPU/RDMA workloads. We're labeling our nodes with the switch information during provisioning, so the static option will work for us. Looking forward to seeing it in v1.11, but I'll also try @yeahdongcn's solution in the meantime.

yeahdongcn:
> I have the same requirement for improving the placement/locality of GPU/RDMA workloads. We're labeling our nodes with the switch information during provisioning, so the static option will work for us. Looking forward to seeing it in v1.11, but I'll also try @yeahdongcn's solution in the meantime.

Cool! Upon further investigation, we identified scenarios involving both intra-node (`GPU <-> NIC`) and inter-node (`NODE <-> NODE`) connections. To address these comprehensively, network topology details are stored in a `ConfigMap` using the Slurm configuration format, while PCIe topology information is applied as labels on the nodes.

OguzPastirmaci:
> Cool! Upon further investigation, we identified scenarios involving both intra-node (GPU <-> NIC) and inter-node (NODE <-> NODE) connections. To address these comprehensively, network topology details are stored in a ConfigMap using the Slurm configuration format, while PCIe topology information is applied as labels on the nodes.

Great! We have the option of deploying Slurm as well and we create the Slurm topology based on the switch/network block information available in our metadata service. Is static still going to be an option if we don't need to dynamically discover the network?

yeahdongcn:
> Great! We have the option of deploying Slurm as well and we create the Slurm topology based on the switch/network block information available in our metadata service. Is static still going to be an option if we don't need to dynamically discover the network?

This may depend on your workload profile. For example, in LLM training workloads, network health and connectivity are critical factors. By employing dynamic topology, while we might not always select the shortest path for a job, we can ensure network connectivity—particularly important given the frequent occurrence of optical module failures in IB or RoCE environments.

Monokaix:
Hi guys, afetr a long discussion with community users and we have proposed a revised doc here https://github.com/volcano-sh/volcano/pull/3850, and we have also discussed the detail in the weekly meeting, which we believe can cover more use cases, and hope for your feedback!

lowang-bh:
close it as there is a more ideal  design in https://github.com/volcano-sh/volcano/pull/3850

PR #3391, opened 2024-04-07
fix panic when the futureIdle resources are calculated to be negative

Lily922:
fix: https://github.com/volcano-sh/volcano/issues/3392


lowang-bh (commented):
/lgtm

william-wang (approved):
/approve

Issue #3386, opened 2024-04-04
compare task order with uid as the default sort method, when task index parsed failed, such as "pod1" vs "pod2"


lowang-bh:
#### What would you like to be added:

sort task with uid, such as "pod1" before "pod2", as default

#### Why is this needed:
The origin code has this feature but it is removed in pr https://github.com/volcano-sh/volcano/pull/1898
I think this is necessary and it is more clear which one has higher priority when `GetPodIndexUnderTask` failed.

Monokaix:
you mean just compare pod name directly?

lowang-bh:
> you mean just compare pod name directly?

No, add it as the last default sort method like the origin code, when all other method failed to find out the result.

![image](https://github.com/volcano-sh/volcano/assets/21003791/e21cb3fb-6eb4-41c7-bb4e-10e9b1683a3e)

Monokaix:
You can work on that: )

ipsum-0320:
I want to give it a try! /assign

PR #3398, opened 2024-04-10
sort tasks by UID when task index parsing fails

panoswoo:
Resolves [#3386](https://github.com/volcano-sh/volcano/issues/3386)

The tasks of vc-job is created in goroutine, and the CreationTimestamp of tasks is random. So, sorting by UID when task index parsed failed is a more clear and reasonable method.

### DIFF
--- a/pkg/controllers/job/helpers/helpers_test.go
+++ b/pkg/controllers/job/helpers/helpers_test.go
@@ -35,23 +35,33 @@ func TestCompareTask(t *testing.T) {
 		expect bool
 	}{
 		{
-			generateTaskInfo("prod-worker-21", createTime),
-			generateTaskInfo("prod-worker-1", createTime),
+			generateTaskInfo("prod-worker-21", createTime, "pod-1"),
+			generateTaskInfo("prod-worker-1", createTime, "pod-2"),
 			false,
 		},
 		{
-			generateTaskInfo("prod-worker-0", createTime),
-			generateTaskInfo("prod-worker-3", createTime),
+			generateTaskInfo("prod-worker-0", createTime, "pod-1"),
+			generateTaskInfo("prod-worker-3", createTime, "pod-2"),
 			true,
 		},
 		{
-			generateTaskInfo("prod-worker", createTime),
-			generateTaskInfo("prod-worker-3", createTime.Add(time.Hour)),
+			generateTaskInfo("prod-worker", createTime, "pod-1"),
+			generateTaskInfo("prod-worker-3", createTime.Add(time.Hour), "pod-2"),
 			true,
 		},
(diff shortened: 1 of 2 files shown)


lowang-bh:
let's keep comparing createtime.

Monokaix:
Welcome! we should compare creationTime first, if they are equal, then compare uid: )

panoswoo:
> let's keep comparing createtime.

done

panoswoo:
> Welcome! we should compare creationTime first, if they are equal, then compare uid: )

done

pkg/controllers/job/helpers/helpers_test.go
```diff
@@ -17,12 +17,10 @@ limitations under the License.
 package helpers
 
 import (
-	"testing"
-	"time"
-
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
+	"testing"
```
Monokaix, line 22: plz sort imports.

panoswoo: done

pkg/controllers/job/helpers/helpers_test.go
```diff
@@ -35,23 +35,33 @@ func TestCompareTask(t *testing.T) {
...
-			generateTaskInfo("prod-worker", createTime),
-			generateTaskInfo("prod-worker-3", createTime.Add(time.Hour)),
+			generateTaskInfo("prod-worker", createTime, "pod-1"),
+			generateTaskInfo("prod-worker-3", createTime.Add(time.Hour), "pod-2"),
 			true,
 		},
 		{
-			generateTaskInfo("prod-worker", createTime),
-			generateTaskInfo("prod-worker-3", createTime.Add(-time.Hour)),
+			generateTaskInfo("prod-worker", createTime, "pod-1"),
```
Monokaix, line 53: We'd better add two test case that task index parsed failed, and one case is two pods with same timestamp, another case is two pods with differenet timestamp.

panoswoo: case 3&4: task index parsed failed but two pods with different timestamp
case 5&6: task index parsed failed and two pods with same timestamp

panoswoo: @Monokaix hi, is there anything else that needs to be modified?

Monokaix: Ok, will trigger the CI later: )

william-wang (approved):
/approve

william-wang (commented):
/lgtm

Issue #3403, opened 2024-04-12
Is is possible that volcano can schedule top priority jobs in all queues first


molei20021:
#### What would you like to be added:
I would like to add some config in scheduler to specify the priority range of jobs that volcano can try schedule these jobs first in all queues 
#### Why is this needed:
We need multiple queue to limit the quota for different project, we also need to schedule the most important jobs first in these projects. We see in allocate action of volcano, jobs of lower priority in queue A may be try schedule earlier than jobs of higher priority in queue B if the priority of queue A is higher than the priority of queue B.

lowang-bh:
> if the priority of queue A is higher than the priority of queue B

queue's priority is dynamic changed and depend on the resource usage of all jobs in that queue. Once a job of a queue allocated, queue will put back to priority-queue and re-adjust the queue priority. 


But now there is a litter issue that queue is put back before a job's resource allocating. I will fire a issue.

robscc:
Will u add some plugable function for sorting queue priority? In my case I would check whether the queue have the emergency task need be scheduled first

molei20021:
https://github.com/volcano-sh/volcano/pull/3405
I want to add a schedulefail event to notify my custom plugin to realize schedule top priority jobs in all queues first

molei20021:
Background:
  I want to use queueorder plugin to realize global priority order for some special tasks. For example tasks with priority larger than 500 will be tried to schedule first. If all the special task is tried to schedule in queue, then the priority of queue will be set low. So the information of whether tasks is tried to schedule in different actions is important to queueorder plugin. But volcano will not tell such information to the plugin currently. So I add an event to tell the plugin such information.

googs1025:
@molei20021 I'm not entirely clear why the queueorder plugin is necessary if the scheduler already dequeues tasks based on their priority. Additionally, I don't understand the need for event notification plugins and why they are required.

molei20021:
For example, there are queue1(task1 > task2 > task3 > task4) and queue2(task5 > task6 > task7 > task8). I want task1 and task5 to try schedule first global. So I need the information of whether task1 and task5 has tried to schedule In queue order plugin. Even if queue1 > queue2 and task1 is scheduled, I will force make queue2 > queue1 to schedule task5. The information of whether task is processed by action can by notified by event I added.

googs1025:
The queue with a higher priority should be dequeued first, so why does task 5 need to be scheduled first?

googs1025:
[...]

PR #3405, opened 2024-04-12
add action event

molei20021:
add an action event to tell custom plugin schedule fail event to realize schedule top priority jobs in all queues first


molei20021:
@alcorj-mizar could you review the code?

lowang-bh:
could you file a issue to describe the detail background?

And, please rebase and squash commits to only one.

molei20021:
> could you file a issue to describe the detail background?
> 
> And, please rebase and squash commits to only one.

https://github.com/volcano-sh/volcano/issues/3403
Background:
  I want to use queueorder plugin to realize global priority order for some special tasks. For example tasks with priority larger than 500 will be tried to schedule first. If all the special task is tried to schedule in queue, then the priority of queue will be set low. So the information of whether tasks is tried to schedule in different actions is important to queueorder plugin. But volcano will not tell such information to the plugin currently. So I add an event to tell the plugin such information.

molei20021:
@lowang-bh I have rebase and squash commits to only one.

molei20021:
There is some conflict so i made a new pr #3433 please have a look. @lowang-bh @googs1025

PR #3408, opened 2024-04-12
fix failed ut: reclaimee pod is randomly evicted

lowang-bh:
Fix failed CI.
As `preemptable` property of task is true as default, should specify a `false` value for not preemptable pod.


william-wang (approved):
/approve

PR #3413, opened 2024-04-15
put back the queue to priority queue after job's resource allocating …

### LINKED ISSUE #3407: queue is put back before a job's resource allocating
**What happened**:
In allocate action, a queue is put back to queues priority queue before the job's resource allocated, which will cause the queue is sorted not as expected as drf/proportion plugin really want to do.

https://github.com/volcano-sh/volcano/blob/b53d6d8c55c66ed98ca9d875b6ba9e0f6baef286/pkg/scheduler/actions/allocate/allocate.go#L155-L164

Put back the queue to priority queue after job's resource allocating finished.

1. create two queues with same weight, queue-1 has a running job-a, queue-2 has no running jobs
2. submit a  job-b whose resource request is large than job-a, to queue-1 and queue-2,  and then submit another job-c whose request is same as job-a to queue-2; As queue-2's share is 0 and job in queue'2 will be pop from queues and allocated first (now all is as expected)
4. next step, job-b in queue-1 should be allocated before job-c in queue-2 (which is expected), because queue-1's share value is smaller than queue-2's .

**Environment**:
- Volcano Version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

panoswoo:
Resolves #3407 

Put back the queue to priority queue after job's resource allocating finished to ensure that the priority of the queue is calculated based on the latest resource allocation situation.


lowang-bh:
Hi, please add  some ut to cover it. Thanks

panoswoo:
> Hi, please add some ut to cover it. Thanks

Done, but i haven't found a simple way to check the allocate order of tasks, so I made some extensions to TestCommonStruct to enable it to record the actual allocation order. This change is compatible with previous cases.
Looking forward to your suggestions

lowang-bh:
> > Hi, please add some ut to cover it. Thanks
> 
> Done, but i haven't found a simple way to check the allocate order of tasks, so I made some extensions to TestCommonStruct to enable it to record the actual allocation order. This change is compatible with previous cases. Looking forward to your suggestions

Don't need to do that. Just let the whole resource only enough for allocating first few jobs and check the last job is pending.

panoswoo:
> > > Hi, please add some ut to cover it. Thanks
> > 
> > 
> > Done, but i haven't found a simple way to check the allocate order of tasks, so I made some extensions to TestCommonStruct to enable it to record the actual allocation order. This change is compatible with previous cases. Looking forward to your suggestions
> 
> Don't need to do that. Just let the whole resource only enough for allocating first few jobs and check the left jobs are pending.

pod-1 with 1c1g to queue-1 
pod-1 with 2c2g to queue-1 
pod-3 with 2c2g to queue-1 
pod-4 with 1c1g to queue-1

> > > Hi, please add some ut to cover it. Thanks
> > 
> > 
> > Done, but i haven't found a simple way to check the allocate order of tasks, so I made some extensions to TestCommonStruct to enable it to record the actual allocation order. This change is compatible with previous cases. Looking forward to your suggestions
> 
> Don't need to do that. Just let the whole resource only enough for allocating first few jobs and check the left jobs are pending.

I have considered doing this before, but due to the same weight for both queues, when I try to fill up all resources with tasks, they will be rejected by the proportion plugin during Overused and Allocable detection. (because we only have two queues, if we want to fill up all resources, must have one queue's resource allocated is more than half).
But I just realized that maybe I can disable the Overused and Allocatable checks directly in the plugin's config.

panoswoo:
> > > Hi, please add some ut to cover it. Thanks
> > 
> > 
> > Done, but i haven't found a simple way to check the allocate order of tasks, so I made some extensions to TestCommonStruct to enable it to record the actual allocation order. This change is compatible with previous cases. Looking forward to your suggestions
> 
> Don't need to do that. Just let the whole resource only enough for allocating first few jobs and check the left jobs are pending.

I have considered doing this before, but due to the same weight for both queues, when I try to fill up all resources with tasks, they will be rejected by the proportion plugin during Overused and Allocable detection. (because we only have two queues, if we want to fill up all resources, must have one queue's resource allocated is more than half).
But I just realized that maybe I can disable the Overused and Allocatable checks directly in the plugin's config.

lowang-bh:
> Don't need to do that. Just let the whole resource only enough for allocating first few jobs and check the left jobs are pending.

Add two cases: 
- total 5cpus, pod-small-1 use 1cpu, then pod-large-2 use 3cpu, and then pod-large-1 pending, indicate that q-2 is allocated first
- total 5 cpus,  pod-small-1 use 1cpu, then pod-large-2 use 2cpu, and then pod-large-1 use 2 cpus, pod-small-2 is pending, indicate that q-2 is first  allocated and then q-1 is allocated

lowang-bh:
Wait #3408 fix the failed UT.

lowang-bh:
May be we need to adjust the e2e test case.

panoswoo:
> May be we need to adjust the e2e test case.

let me take a look

panoswoo:
@lowang-bh Hi, I seem to have found some issues while searching for the volcano component logs generated by e2e testing.
1. when e2e test failed, we will backup volcano component logs by https://github.com/volcano-sh/volcano/blob/c414e56be9d238efd4665ba81d06a535f55dc6b9/hack/run-e2e-kind.sh#L69-L74 we are trying to get log from namespace kube-system but those component are installed in volcano-system actually https://github.com/volcano-sh/volcano/blob/c414e56be9d238efd4665ba81d06a535f55dc6b9/hack/run-e2e-kind.sh#L57
2. it seems that we didn't upload the log file at the end of the workflow https://github.com/volcano-sh/volcano/blob/c414e56be9d238efd4665ba81d06a535f55dc6b9/.github/workflows/e2e_scheduling_basic.yaml#L39-L41

lowang-bh:
> 1. we are trying to get log from namespace kube-system but those component are installed in volcano-system actually

That is a issue, now we use `volcano-system`. You can file another PR an merge it first. @Monokaix

panoswoo:
> May be we need to adjust the e2e test case.

I triggered the test again and it passed. 
I haven't made any modifications. Have we recently merged any fixes?

lowang-bh:
> > May be we need to adjust the e2e test case.
> 
> I triggered the test again and it passed. I haven't made any modifications. Have we recently merged any fixes?

No update. Maybe it is randomly failed. We'd better check it.

panoswoo:
> > > May be we need to adjust the e2e test case.
> > 
> > 
> > I triggered the test again and it passed. I haven't made any modifications. Have we recently merged any fixes?
> 
> No update. Maybe it is randomly failed. We'd better check it.

I am unable to reproduce the problem now.
I don't have detailed logs of the previous exception cases, and this error is a large process (involving multiple components) that has timed out. I can't determine which part of the problem went wrong without detailed logs :(

hack/run-e2e-kind.sh
```diff
@@ -157,7 +179,9 @@ case ${E2E_TYPE} in
     ;;
 esac
 
+generate-log
+
 if [[ $? -ne 0 ]]; then
-  generate-log
+
```
lowang-bh, line 185: Will generate-log be put here?

panoswoo: the modifications made in run-e2e-kind.sh are temporary and were used for debugging
I will revert these changes in this PR and create a new one

lowang-bh:
@panoswoo please make the ci succeed and so we can make this pr merged.

panoswoo:
> @panoswoo please make the ci succeed and so we can make this pr merged.

I will fix it ASAP

hwdef (commented):
/lgtm

lowang-bh:
HI, @panoswoo , You can close and reopen it to trigger the CI.

Monokaix:
Why the former e2e are failed?

lowang-bh (commented):
/lgtm

pkg/scheduler/actions/allocate/allocate_test.go
```diff
@@ -133,9 +133,61 @@ func TestAllocate(t *testing.T) {
...
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("q-1", 1, nil),
+				util.BuildQueue("q-2", 1, nil),
+			},
+			Bind: map[string]string{
+				"ns-1/pod-large-1": "node-1",
+				"ns-1/pod-large-2": "node-1",
+			},
+			BindsNum: 2,
```
lowang-bh, line 185: need to rebase and solve the conflicts.

Monokaix:
Please squash to one commit: )

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -162,6 +159,10 @@ func (alloc *Action) allocateResources(queues *util.PriorityQueue, jobsMap map[a
 			tasks.Len(), job.Namespace, job.Name)
 
 		alloc.allocateResourcesForTasks(tasks, job, jobs, queue, allNodes)
+
+		// Put back the queue to priority queue after job's resource allocating finished,
+		// To ensure that the priority of the queue is calculated based on the latest resource allocation situation.
```
william-wang, line 164: `To ensure that the priority of the queue is calculated based on the latest resource allocation`  refers to Line#122 or Line#124. I'd like to know what exactly the problem is. Would you please clarify more?

william-wang (approved):
/approve

lowang-bh:
/close

has been merged in https://github.com/volcano-sh/volcano/pull/3515

PR #3423, opened 2024-04-17
feat: added question to issue template

### LINKED ISSUE #3420: Enhance issue template

Add another issue template like `Question` 
#### Why is this needed:
There are only two types issue bug & feature in volcano, when users have questions about volcnao, they usually choose a bug type, but usually it's only some questions instead of a bug, so we should classify issues more clearly.

We can refer to https://github.com/karmada-io/karmada/blob/master/.github/ISSUE_TEMPLATE/question.md

shruti2522:
Added `question.md` to `ISSUE_TEMPLATE`. 
Fixes #3420.


william-wang:
@shruti2522 It's cool to have this pr :)

william-wang (approved):

PR #3425, opened 2024-04-18
Add options: worker-threads-for-gc

WulixuanS:
A option to adjust the concurrency of gc


lowang-bh:
Shall we need to validate the param?

WulixuanS:
> Shall we need to validate the param?

Do you mean to limit the param range?

lowang-bh:
> > Shall we need to validate the param?
> 
> Do you mean to limit the param range?

Yes.

william-wang (approved):
/approve

PR #3430, opened 2024-04-21
improve: continue allocating if remain tasks is more than job's minMember

lowang-bh:
As PR #1459 has supported minAvailable at task level.

https://github.com/volcano-sh/volcano/pull/3057 has improved and fixed the minResource calculation for a volcano job.
This PR continue to improve and fix the scheduling.

Consider this scene:
     part of pods of a job is not schedulable, and left pods are schedulable, and the number can meet up the condition of min member.
     example: cluster has only 4GPU and enough CP/MEM, a job with 1ps + 8worker, each worker request 1GPU and has higher priority than ps. So workers will be scheduled first and failed at 5th worker. After that, job will be failed scheduling and break out from allocate.

So this pr fix this issue: if allocated pod less than minMember and tasks are not used up, will continue on, untill the number of remained tasks can not meet up the condition of min member

commit 1:  continue allocating and testcases about the feature change
commit 2: use string type to replace the taskID type, which indicate a role spec of each task in a job
commit 3: consider job's each task's min member when check if it can continue on allocating
commit 4:  parse and store task spec in taskInfo, so that it doesn't need to parse it everywhere and whenerver it is used.
And add pre-check if task has fit error so as to save predicating time and avoid the useless computing.  Testing with a 100master+100worker job has no obvious latancy comparing without this PR.

<details>

<summary>An Example Yaml Without Priority setting</summary>

## with the following job yaml on a cluster with available CPU=10

```yaml
kind: Job
metadata:
  name: job5
spec:
  schedulerName: volcano
  minAvailable: 3
  queue: default
  tasks:
    - replicas: 2
      minAvailable: 1
      name: "master"
      template:
        spec:
          containers:
            - image: nginx:latest
              name: nginx
              resources:
                requests:
                  cpu: "4"
                  memory: "50Mi"
          restartPolicy: OnFailure
    - replicas: 2
      minAvailable: 2
      name: "work"
      template:
        spec:
          containers:
            - image: nginx:latest
              name: nginx
              resources:
                requests:
                  cpu: "3"
                  memory: "100Mi"
          restartPolicy: OnFailure
```
</details>

<details>

<summary>Test Example Yaml</summary>

## with the following job yaml on a cluster with available CPU<4

```yaml
kind: Job
metadata:
  name: my-apps
spec:
  schedulerName: volcano
  minAvailable: 2
  tasks:
    - replicas: 2
      name: "web"
      minAvailable: 1
      template:
        spec:
          containers:
            - image: nginx:1.14.2
              name: web
              resources:
                requests:
                  cpu: "1000m"
                  memory: "100Mi"
          restartPolicy: OnFailure
    - replicas: 2
      minAvailable: 1
      name: "server"
      template:
        spec:
          containers:
            - image: nginx:1.14.2
              name: server
              resources:
                requests:
                  cpu: "2000m"
                  memory: "100Mi"
          restartPolicy: OnFailure
```
</details>

## without this pr on master branch
image

image

## with this pr

image
image


pkg/scheduler/api/job_info.go
```diff
@@ -265,11 +269,11 @@ func (ti *TaskInfo) Clone() *TaskInfo {
 	}
 }
 
-func (ti *TaskInfo) GetTaskSpecKey() TaskID {
```
googs1025, line 268: Sorry, I don't understand why the TaskID type was changed to string?

lowang-bh: > Sorry, I don't understand why the TaskID type was changed to string?

It is just a string of spec role, has nothing to do with taskID. And the origin defination of spec is also a string.

googs1025: got it! thanks!

hwdef (commented):
It looks like this PR does two things. But use one commit

lowang-bh:
> It looks like this PR does two things. But use one commit

Have splited them into two commits.

lowang-bh:
@Monokaix @hwdef @wangyang0616 @william-wang please help to review it. thanks.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -193,7 +193,14 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 		predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, alloc.predicate, true)
 		if len(predicateNodes) == 0 {
 			job.NodesFitErrors[task.UID] = fitErrors
-			break
+			// Assume that  all left tasks is allocable, can not meet gang-scheduling min member, we should break from continuously allocating
+			// otherwise, should continue to find other allocable task
+			// TODO: especially, should check each task spec's left pod can meet their min member
+			if job.ReadyTaskNum()+int32(tasks.Len()) < job.MinAvailable {
```
lowang-bh, line 199: should exclude current role's replicas, becaulse same role's pod will be predicated failed.

hwdef: > should exclude current role's replicas, becaulse same role's pod will be predicated failed.

you are right.

hwdef:
I think the others are fine

pkg/scheduler/actions/allocate/allocate.go
lowang-bh: Done.

Monokaix:
The are 4 commits but the description has 2 commit.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -193,7 +193,13 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 		predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, alloc.predicate, true)
 		if len(predicateNodes) == 0 {
 			job.NodesFitErrors[task.UID] = fitErrors
-			break
+			// Assume that  all left tasks is allocable, can not meet gang-scheduling min member, we should break from continuously allocating
```
Monokaix, line 196: A duplicate space.

pkg/scheduler/api/job_info.go
```diff
@@ -711,21 +715,87 @@ func (ji *JobInfo) PendingBestEffortTaskNum() int32 {
 	return int32(count)
 }
 
+// CheckJobNeedContinueAllocating checks whether it can continue to allocate for current job
+// there are two cases to continue:
+// 1. job's total allocable number meet its minAvailable
```
Monokaix, line 720: Why need to continue allocate when gang is meet?

lowang-bh: It is not allocated, but allocable, which means the left pods may be can be predicated succesfully and meet the gang-scheduling.

pkg/scheduler/api/job_info.go
```diff
@@ -719,21 +723,114 @@ func (ji *JobInfo) PendingBestEffortTaskNum() int32 {
...
+		left := int32(0)
+		for role, cnt := range pending {
+			if _, ok := failedRoles[role]; !ok {
+				left += cnt
+			}
+		}
+		return ji.ReadyTaskNum()+left >= ji.MinAvailable
+	}
+
+	// 2. if each task role has its independent minMember, check it
```
lowang-bh, line 779: TODO: we need to add another plugin to solve this dynamic priority in a job with taskMinAvailable setting. See this comment
https://github.com/volcano-sh/volcano/pull/1459#issuecomment-836702823

Monokaix:
A little doubt, in what case will a work pod has higher priority than ps? And what if pods in a vcjob has the same priority?

lowang-bh:
> A little doubt, in what case will a work pod has higher priority than ps? And what if pods in a vcjob have the same priority?

worker with GPU request, can have a higher priority,  so as to allocate gpu pod first.

lowang-bh:
> And what if pods in a vcjob have the same priority?

With no priority setting, the task order will be default compare stragety at
https://github.com/volcano-sh/volcano/blob/34836851e6317807267230b1b9508194f81a6f67/pkg/controllers/job/helpers/helpers.go#L51-L66

<details>

<summary> With Given Example Yaml Without Priority setting</summary>

## before
image
## after this pr
image
</details>

pkg/scheduler/api/job_info.go
```diff
@@ -723,6 +723,27 @@ func (ji *JobInfo) PendingBestEffortTaskNum() int32 {
...
+}
+
+// TaskHasFitErrors checks if the task has fit errors and can continue try predicating
+func (ji *JobInfo) TaskHasFitErrors(task *TaskInfo) bool {
+	// if the task didn't set the spec key, should not use the cache
+	if len(task.TaskSpec) == 0 {
+		return false
+	}
+
+	_, exist := ji.FitFailedRoles()[task.TaskSpec]
```
Monokaix, line 743: `task.TaskSpec` is just set in vcjob, and other case will be "", can this take efffect?

lowang-bh: Will retun false in the if branch.

lowang-bh: Actually, the empty spec and default to enable cache will cause job can not be scheduled in some corner case. See issue https://github.com/volcano-sh/volcano/issues/3527. We'd better don't enable cache when spec is empty in predicate helper.

hwdef (commented):
/lgtm

pkg/scheduler/api/job_info.go
```diff
@@ -107,6 +107,7 @@ type TaskInfo struct {
 
 	Name      string
 	Namespace string
+	TaskSpec  string // value of "volcano.sh/task-spec"
```
william-wang, line 110: `taskSpec` is the task name. Is it the same with `Name` field.

pkg/scheduler/api/job_info.go
```diff
@@ -719,21 +723,114 @@ func (ji *JobInfo) PendingBestEffortTaskNum() int32 {
...
+//  2. each task's allocable number meet its independent minAvailable
+//     this is for the case that each task role has its own independent minMember.
+//     eg, current role's pod has a failed predicating result but its allocated number has meet its minMember,
+//     the other roles' pods which have no failed predicating results can continue on
+//
+// performance analysis:
+//
+//	As the failed predicating role has been pre-checked when it was poped from queue,
+//	this function will only be called at most as the number of roles in this job.
+func (ji *JobInfo) CheckJobNeedContinueAllocating() bool {
```
william-wang, line 761: CheckJobNeedContinueAllocating() is used to handle the elastic job whose 's minavailable is not equal to task numbers. If  ji.MinAvailable == len(Tasks), it's better return at once to avoid the useless computing.

william-wang: CheckJobNeedContinueAllocating -> needContinueAllocating()

pkg/scheduler/api/job_info.go
lowang-bh: Yes, have updated to TaskRole.

pkg/scheduler/api/job_info.go
lowang-bh: Done.

lowang-bh: > If ji.MinAvailable == len(Tasks), it's better return at once to avoid the useless computing

Good catch. I have considered the performance effect, but forget this one.

Issue #3086, opened 2023-08-29
Predicate function retrun value is not compatiable with old plugins


lowang-bh:
There are other plugins dynamically loaded to volcano. eg. huawei's acend gpu, https://gitee.com/ascend/ascend-for-volcano

**What happened**:

https://github.com/volcano-sh/volcano/blob/b95d0f15e0ff2af9a7c9ae9a037a4c4c9e7300aa/pkg/scheduler/api/types.go#L170C6-L170C17
```
// PredicateFn is the func declaration used to predicate node for task.
type PredicateFn func(*TaskInfo, *NodeInfo) ([]*Status, error)
```

npu plugin
https://gitee.com/ascend/ascend-for-volcano/blame/master/huawei_npu.go
```
	// if npu no meet the task require,the task will failed.so need to intercept in advance
	ssn.AddPredicateFn(tp.Name(), func(taskInfo *api.TaskInfo, nodeInfo *api.NodeInfo) error {
		return sHandler.NodePredicate(taskInfo, nodeInfo, ssn.Configurations)
	})
```

**What you expected to happen**:

Should be compatiable with other extender plugins

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

Introduced by refact PR [#2916](https://github.com/volcano-sh/volcano/pull/2916)

**Environment**:
- Volcano Version: master
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

william-wang:
Let's do this in v1.9

lowang-bh:
@Monokaix we need to fix  it in new realease.

Issue #3074, opened 2023-08-26
enhancement for preemption action


lowang-bh:
#### What would you like to be added:

1. filter out those nodes which can not help preemption, according to the filter result stored in job's NodesFitErrors
2. range from those nodes in step-1, consider the victims from a node and remove the victims from this node, and then re-run predications on the node; if result is success, then the node is a candidate node, put them in final node list;
3. score on thoes candinate nodes, to choose a best one.

lowang-bh:
A little same as #3071

lowang-bh:
For example, ascend-volcano-plugin's predication function will do idle resource check.
<img width="1376" alt="image" src="https://github.com/user-attachments/assets/20b35289-fb96-42e3-9f24-d9d25dd70fe4">

Monokaix:
All tasks have been done here？

PR #3432, opened 2024-04-21
filter out those nodes which are UnschedulableAndUnresolvable when preempting

### LINKED ISSUE #3071: filterout those nodes which are not helpful for preemptation
like K8s preemptations, it first filters out those nodes will not help preemption

https://github.com/kubernetes/kubernetes/blob/dbcbf5e952266479df92c84457781c0a2a688afd/pkg/scheduler/framework/preemption/preemption.go#L223C2-L239
```go
        potentialNodes, unschedulableNodeStatus := nodesWherePreemptionMightHelp(allNodes, m)
        if len(potentialNodes) == 0 {
                logger.V(3).Info("Preemption will not help schedule pod on any node", "pod", klog.KObj(pod))
                // In this case, we should clean-up any existing nominated node name of the pod.
                if err := util.ClearNominatedNodeName(ctx, ev.Handler.ClientSet(), pod); err != nil {
                        logger.Error(err, "Could not clear the nominatedNodeName field of pod", "pod", klog.KObj(pod))
                        // We do not return as this error is not critical.
                }
                return nil, unschedulableNodeStatus, nil
        }


        pdbs, err := getPodDisruptionBudgets(ev.PdbLister)
        if err != nil {
                return nil, nil, err
        }


        offset, numCandidates := ev.GetOffsetAndNumCandidates(int32(len(potentialNodes)))
```

https://github.com/kubernetes/kubernetes/blob/dbcbf5e952266479df92c84457781c0a2a688afd/pkg/scheduler/framework/preemption/preemption.go#L413-L428
```go
// nodesWherePreemptionMightHelp returns a list of nodes with failed predicates
// that may be satisfied by removing pods from the node.
func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.NodeToStatusMap) ([]*framework.NodeInfo, framework.NodeToStatusMap) {
        var potentialNodes []*framework.NodeInfo
        nodeStatuses := make(framework.NodeToStatusMap)
        for _, node := range nodes {
                name := node.Node().Name
                // We rely on the status by each plugin - 'Unschedulable' or 'UnschedulableAndUnresolvable'
                // to determine whether preemption may help or not on the node.
                if m[name].Code() == framework.UnschedulableAndUnresolvable {
                        nodeStatuses[node.Node().Name] = framework.NewStatus(framework.UnschedulableAndUnresolvable, "Preemption is not helpful for scheduling")
                        continue
                }
                potentialNodes = append(potentialNodes, node)
        }
        return potentialNodes, nodeStatuses
}
```

#### Why is this needed:

reduce the nodes which need to run filter on them

lowang-bh:
Fixes #3071
Part of enhancement for preemption action #3074
Those nodes with status UnschedulableAndUnresolvable is not helpful for preempting. So we can use cached predicating result  to reduce the nodes numbers when scan for a victim pod.

we can also use the new Api `NewFitStatus` to return an error in predicating to fix issue [Predicate function retrun value is not compatiable with old plugins #3086](https://github.com/volcano-sh/volcano/issues/3086)

main changes:
commit-1: add predicate result code to task's fitError
commit-2: testcase about fitErrors
commit-3: filter out those nodes which are not helpful for preemption in preempt/reclaim actions

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -292,10 +292,11 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 
 func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
 	// Check for Resource Predicate
-	if ok, resources := task.InitResreq.LessEqualWithResourcesName(node.FutureIdle(), api.Zero); !ok {
-		return nil, api.NewFitError(task, node, api.WrapInsufficientResourceReason(resources))
-	}
 	var statusSets api.StatusSets
+	if ok, resources := task.InitResreq.LessEqualWithResourcesName(node.FutureIdle(), api.Zero); !ok {
+		statusSets = append(statusSets, &api.Status{Code: api.Unschedulable, Reason: api.WrapInsufficientResourceReason(resources)})
+		return nil, api.NewFitErrWithStatus(task, node, statusSets...)
+	}
 	statusSets, err := alloc.session.PredicateFn(task, node)
 	if err != nil {
 		return nil, api.NewFitError(task, node, err.Error())
@@ -303,7 +304,7 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) ([]*api.S
 
 	if statusSets.ContainsUnschedulable() || statusSets.ContainsUnschedulableAndUnresolvable() ||
 		statusSets.ContainsErrorSkipOrWait() {
-		return nil, api.NewFitStatus(task, node, statusSets...)
+		return nil, api.NewFitErrWithStatus(task, node, statusSets...)
 	}
 	return nil, nil
 }
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -303,7 +303,7 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) ([]*api.S
 
 	if statusSets.ContainsUnschedulable() || statusSets.ContainsUnschedulableAndUnresolvable() ||
 		statusSets.ContainsErrorSkipOrWait() {
-		return nil, api.NewFitStatus(task, node, statusSets)
+		return nil, api.NewFitStatus(task, node, statusSets...)
 	}
 	return nil, nil
 }
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -295,7 +295,7 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) ([]*api.S
 	if ok, resources := task.InitResreq.LessEqualWithResourcesName(node.FutureIdle(), api.Zero); !ok {
 		return nil, api.NewFitError(task, node, api.WrapInsufficientResourceReason(resources))
 	}
-	var statusSets util.StatusSets
+	var statusSets api.StatusSets
 	statusSets, err := alloc.session.PredicateFn(task, node)
 	if err != nil {
 		return nil, api.NewFitError(task, node, err.Error())
@@ -303,7 +303,7 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) ([]*api.S
 
 	if statusSets.ContainsUnschedulable() || statusSets.ContainsUnschedulableAndUnresolvable() ||
 		statusSets.ContainsErrorSkipOrWait() {
-		return nil, api.NewFitError(task, node, statusSets.Message())
+		return nil, api.NewFitStatus(task, node, statusSets)
 	}
 	return nil, nil
 }
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -221,7 +221,7 @@ func preempt(
 		// When filtering candidate nodes, need to consider the node statusSets instead of the err information.
 		// refer to kube-scheduler preemption code: https://github.com/kubernetes/kubernetes/blob/9d87fa215d9e8020abdc17132d1252536cd752d2/pkg/scheduler/framework/preemption/preemption.go#L422
 		if statusSets.ContainsUnschedulableAndUnresolvable() || statusSets.ContainsErrorSkipOrWait() {
-			return nil, api.NewFitStatus(task, node, statusSets)
+			return nil, api.NewFitStatus(task, node, statusSets...)
 		}
 		return nil, nil
 	}
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -208,8 +208,7 @@ func preempt(
 	predicateHelper util.PredicateHelper,
 ) (bool, error) {
 	assigned := false
-	allNodes := ssn.NodeList
-	// TODO: we should filter out those nodes that are UnschedulableAndUnresolvable status got in allocate action
+
 	if err := ssn.PrePredicateFn(preemptor); err != nil {
 		return false, fmt.Errorf("PrePredicate for task %s/%s failed for: %v", preemptor.Namespace, preemptor.Name, err)
 	}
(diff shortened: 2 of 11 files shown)


pkg/scheduler/api/unschedule_info.go
```diff
@@ -105,14 +105,33 @@ func NewFitError(task *TaskInfo, node *NodeInfo, message ...string) *FitError {
...
 	}
+	sts := make([]*Status, 0, len(message))
+	for _, msg := range message {
+		sts = append(sts, &Status{Reason: msg, Code: Error})
+	}
+	fe.Status = StatusSets(sts)
 	return fe
 }
 
+func NewFitStatus(task *TaskInfo, node *NodeInfo, sts StatusSets) *FitError {
```
Monokaix: function name contains `status` but return a Fiteror，this iswired and not descriptive.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -293,15 +293,15 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) ([]*api.S
 	if ok, resources := task.InitResreq.LessEqualWithResourcesName(node.FutureIdle(), api.Zero); !ok {
 		return nil, api.NewFitError(task, node, api.WrapInsufficientResourceReason(resources))
```
Monokaix, line 294: Here and `prePredict` still return api.NewFitError, we should unify them.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -220,7 +219,8 @@ func preempt(
 		}
 		return nil, nil
 	}
-
+	// we should filter out those nodes that are UnschedulableAndUnresolvable status got in allocate action
+	allNodes := ssn.NodesWherePreemptionMightHelp(preemptor)
 	predicateNodes, _ := predicateHelper.PredicateNodes(preemptor, allNodes, predicateFn, true)
```
Monokaix, line 224: Do we still need predict another name because `ssn.NodesWherePreemptionMightHelp` has already get the available nodes for preempt?

pkg/scheduler/api/unschedule_info.go
lowang-bh: Done.

pkg/scheduler/actions/allocate/allocate.go
lowang-bh: NewFitError will set the status code with Error as default.

pkg/scheduler/actions/preempt/preempt.go
lowang-bh: Some nodes are not run predicate process in allocate action, so they will not be filter out and still need to run predicate again.

pkg/scheduler/actions/allocate/allocate.go
Monokaix: I mean both `predict` and `prePredict` should return same value.

pkg/scheduler/api/unschedule_info.go
```diff
@@ -56,13 +56,24 @@ func (f *FitErrors) SetNodeError(nodeName string, err error) {
...
 
 	f.nodes[nodeName] = fe
 }
 
+// NotHelpfulForPreemptNodes returns the set of nodes that has no help from preempting pods from it
+func (f *FitErrors) NotHelpfulForPreemptNodes() map[string]struct{} {
+	ret := make(map[string]struct{})
+	for _, node := range f.nodes {
+		if node.Status.ContainsUnschedulableAndUnresolvable() {
+			ret[node.NodeName] = struct{}{}
```
Monokaix, line 71: we can use `sets.Empty{}` to replace `struct{}`.

pkg/scheduler/api/unschedule_info.go
```diff
@@ -56,13 +56,24 @@ func (f *FitErrors) SetNodeError(nodeName string, err error) {
...
 	}
 
 	f.nodes[nodeName] = fe
 }
 
+// NotHelpfulForPreemptNodes returns the set of nodes that has no help from preempting pods from it
+func (f *FitErrors) NotHelpfulForPreemptNodes() map[string]struct{} {
+	ret := make(map[string]struct{})
+	for _, node := range f.nodes {
+		if node.Status.ContainsUnschedulableAndUnresolvable() {
```
Monokaix, line 70: There are some cases that prePredict or predict plugin just return err and have no statusSet returned, and here just checked status, so the nodes with only errs will also be considered when preempt, is this expected？

pkg/scheduler/api/unschedule_info.go
lowang-bh: Mark it, do it in next Pr.

pkg/scheduler/api/unschedule_info.go
lowang-bh: Yes, we need to unify the error to `NewFitErrWithStatus` which contains status code in next pr.

pkg/scheduler/api/unschedule_info.go
lowang-bh: Done.

pkg/scheduler/api/unschedule_info.go
```diff
@@ -56,13 +58,24 @@ func (f *FitErrors) SetNodeError(nodeName string, err error) {
...
 			NodeName: nodeName,
-			Reasons:  []string{obj.Error()},
+			Status:   []*Status{{Code: Error, Reason: obj.Error()}},
 		}
 	}
 
 	f.nodes[nodeName] = fe
 }
 
+// NotHelpfulForPreemptNodes returns the set of nodes that has no help from preempting pods from it
```
william-wang, line 68: suggest change function name to `getUnschedulableAndUnresolvableNodes`. This function might be used for more places besides preemption.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -626,6 +626,32 @@ func (ssn *Session) TaskOrderFn(l, r interface{}) bool {
 	return helpers.CompareTask(lv, rv)
 }
 
+// NodesWherePreemptionMightHelp filter out those node that has unscheduleandunre
+func (ssn *Session) NodesWherePreemptionMightHelp(task *api.TaskInfo) []*api.NodeInfo {
```
william-wang, line 630: Suggest to change function name to `getUnschedulableAndUnresolvableNodesForTask`

pkg/scheduler/framework/session_plugins.go
```diff
@@ -626,6 +626,32 @@ func (ssn *Session) TaskOrderFn(l, r interface{}) bool {
...
+	fitErrors, ok1 := ssn.Jobs[task.Job]
```
william-wang, line 631: it seems `NodesWherePreemptionMightHel` is a public function for actions. Do not place this func in session_plugin.go. session_plugin.go offers a set of interfaces for plugins to implement. e.g. JobReady(), JobOrder() etc.

pkg/scheduler/api/unschedule_info.go
```diff
@@ -96,23 +109,46 @@ type FitError struct {
...
+	sts := make([]*Status, 0, len(message))
+	for _, msg := range message {
+		sts = append(sts, &Status{Reason: msg, Code: Error})
+	}
+	fe.Status = StatusSets(sts)
+	return fe
+}
+
+// NewFitErrWithStatus returns a fit error with code and reason in it
+func NewFitErrWithStatus(task *TaskInfo, node *NodeInfo, sts ...*Status) *FitError {
```
william-wang, line 131: Why we have to add this new func. Is it prepared to work with #3432 to solve the compatable issue?

pkg/scheduler/framework/session_plugins.go
lowang-bh: Done.

pkg/scheduler/framework/session_plugins.go
lowang-bh: It's reasonable. Done.

pkg/scheduler/api/unschedule_info.go
lowang-bh: Done

pkg/scheduler/api/unschedule_info.go
lowang-bh: Yes, it is ready to rolve the compatable issue [Predicate function retrun value is not compatiable with old plugins #3086](https://github.com/volcano-sh/volcano/issues/3086)

william-wang (approved):
/approve

PR #3434, opened 2024-04-23
fix errTask channel memory leak

bibibox:
Current errTask has some problems as follows:
1. use taskInfo pointer as key, which means same task maybe has different keys, and is handled respectively
2. when handling items in the errTask, if the pod is not found and clean task from cache failed, the task will be added to errTask again, and it will caused endless loop
3. the default controller rate limiting queue qps is just 10. since the pod with not found pvc is added to errTask at present, the qps maybe too small

this PR fix these problems

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1088,6 +1088,7 @@ func (sc *SchedulerCache) processResyncTask() {
 	taskKey, ok := obj.(string)
 	if !ok {
 		klog.Errorf("Failed to convert %v to string.", obj)
+		sc.errTasks.Forget(obj)
 		return
 	}
 
@@ -1104,7 +1105,7 @@ func (sc *SchedulerCache) processResyncTask() {
 		sc.resyncTask(task)
 		reSynced = true
 	} else {
-		klog.V(4).Infof("sync task <%s/%s> success", task.Namespace, task.Name)
+		klog.V(4).Infof("Successfully synced task <%s/%s>", task.Namespace, task.Name)
 		sc.errTasks.Forget(obj)
 	}
 
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/time/rate"
 	v1 "k8s.io/api/core/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -1042,19 +1081,31 @@ func (sc *SchedulerCache) processResyncTask() {
...
 	}
 
 	reSynced := false
 	if err := sc.syncTask(task); err != nil {
-		klog.Errorf("Failed to sync pod <%v/%v>, retry it.", task.Namespace, task.Name)
+		klog.ErrorS(err, "Failed to sync task, retry it", "namespace", task.Namespace, "name", task.Name)
 		sc.resyncTask(task)
 		reSynced = true
+	} else {
+		klog.V(4).Infof("sync task <%s/%s> success", task.Namespace, task.Name)
```
Monokaix, line 1107: Successfully synced task xxx

pkg/scheduler/cache/cache.go
```diff
@@ -1042,19 +1081,31 @@ func (sc *SchedulerCache) processResyncTask() {
...
+		sc.errTasks.Forget(obj)
```
Monokaix, line 1108: If `CustomBindErrHandler` execute successfully, we should also forget it.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -326,8 +326,7 @@ func (sc *SchedulerCache) deleteTask(pi *schedulingapi.TaskInfo) error {
 		if job, found := sc.Jobs[pi.Job]; found {
 			jobErr = job.DeleteTaskInfo(pi)
 		} else {
-			jobErr = fmt.Errorf("failed to find Job <%v> for Task %v/%v",
-				pi.Job, pi.Namespace, pi.Name)
+			klog.Warningf("failed to find job <%v> for Task <%v/%v>", pi.Job, pi.Namespace, pi.Name)
```
Monokaix, line 329: Uppper case here.

pkg/scheduler/cache/cache.go
bibibox: We have forgot it when sync task successfully, just need resync task when CustomBindErrHandler execute failed

pkg/scheduler/cache/cache.go
bibibox: done

pkg/scheduler/cache/event_handlers.go
bibibox: done

pkg/scheduler/cache/cache.go
```diff
@@ -1042,19 +1081,31 @@ func (sc *SchedulerCache) processResyncTask() {
...
 
+	klog.V(5).Infof("the length of errTasks is %d", sc.errTasks.Len())
+
 	defer sc.errTasks.Done(obj)
 
-	task, ok := obj.(*schedulingapi.TaskInfo)
+	taskKey, ok := obj.(string)
 	if !ok {
-		klog.Errorf("failed to convert %v to *schedulingapi.TaskInfo", obj)
+		klog.Errorf("Failed to convert %v to string.", obj)
```
wangyang0616, line 1090: The obj format is incorrect. Is it more reasonable to exclude Forget?

Monokaix:
Should also merge to master.

pkg/scheduler/cache/cache.go
bibibox: done

william-wang (approved):

Issue #3104, opened 2023-09-07
scheduler update job failed


WulixuanS:
**What happened**:
A large number of error logs
```
E0907 03:39:28.604985       1 job_updater.go:105] Failed to update job <video/pgvx20230907113107.701322940-098ad-45e77d6a-f3d6-42a2-9002-c3b9f7c4c506>: Operation cannot be fulfilled on podgroups.scheduling.volcano.sh "pgvx20230907113107.701322940-098ad-45e77d6a-f3d6-42a2-9002-c3b9f7c4c506": StorageError: invalid object, Code: 4, Key: /registry/scheduling.volcano.sh/podgroups/video/pgvx20230907113107.701322940-098ad-45e77d6a-f3d6-42a2-9002-c3b9f7c4c506, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 265232bc-3fe2-47cb-8335-0808e5a569a4, UID in object meta:
E0907 03:39:28.636600       1 job_updater.go:105] Failed to update job <video/gvx20230907113235.385478059-2deeb-c3fdcbe3-6f70-44ed-ae32-3833abbfc01f>: Operation cannot be fulfilled on podgroups.scheduling.volcano.sh "gvx20230907113235.385478059-2deeb-c3fdcbe3-6f70-44ed-ae32-3833abbfc01f": StorageError: invalid object, Code: 4, Key: /registry/scheduling.volcano.sh/podgroups/video/gvx20230907113235.385478059-2deeb-c3fdcbe3-6f70-44ed-ae32-3833abbfc01f, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 673133c8-ba6a-468c-820b-b2c540769575, UID in object meta:
[...]

PR #3439, opened 2024-04-25
fixed to update podgroup resource

calvin0327:
fixed: https://github.com/volcano-sh/volcano/issues/3104

Because of resource version, It's easy to  fail to update podgroup. we should obtain the latest resource version before updating.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -295,7 +295,25 @@ func (su *defaultStatusUpdater) UpdatePodGroup(pg *schedulingapi.PodGroup) (*sch
 		return nil, err
 	}
 
-	updated, err := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Update(context.TODO(), podgroup, metav1.UpdateOptions{})
+	var updated *vcv1beta1.PodGroup
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		newer, updateErr := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Update(context.TODO(), podgroup, metav1.UpdateOptions{})
+		if updateErr == nil {
+			updated = newer
+			return nil
+		}
+
+		older, err := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Get(context.TODO(), podgroup.Name, metav1.GetOptions{})
+		if err == nil {
+			podgroup.ResourceVersion = older.ResourceVersion
+		} else {
+			if apierrors.IsNotFound(err) {
+				return err
+			}
+			klog.Errorf("Error while getting PodGroup with error: %v", err)
+		}
+		return updateErr
+	})
 	if err != nil {
 		klog.Errorf("Error while updating PodGroup with error: %v", err)
 		return nil, err


hwdef (commented):
I think the right way is to get the latest podgroup when the update fails

pkg/scheduler/cache/cache.go
```diff
@@ -295,6 +295,13 @@ func (su *defaultStatusUpdater) UpdatePodGroup(pg *schedulingapi.PodGroup) (*sch
 		return nil, err
 	}
 
+	older, err := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Get(context.TODO(), podgroup.Name, metav1.GetOptions{})
```
googs1025, line 298: Why do we get the version number before updating? Shouldn't we get the new version number and update again after updating the error?

calvin0327:
@hwdef @googs1025 I agrees with you, it can reduce one request in most cases.

pkg/scheduler/cache/cache.go
googs1025: @calvin0327 
I read the description of your answer, shoud we use this
```go 
// IsResourceExpired is true if the error indicates the resource has expired and the current action is
// no longer possible.
// It supports wrapped errors and returns false when the error is nil.
func IsResourceExpired(err error) bool {
	return ReasonForError(err) == metav1.StatusReasonExpired
}
```

googs1025: But I understand that the probability of version conflict is very small. I can't tell whether this error is accidental or will inevitably occur in certain scenarios.

calvin0327: I think this a accidental error, if resource version is not change it's work normally

calvin0327: > @calvin0327 I read the description of your answer, shoud we use this
> 
> ```go
> // IsResourceExpired is true if the error indicates the resource has expired and the current action is
> // no longer possible.
> // It supports wrapped errors and returns false when the error is nil.
> func IsResourceExpired(err error) bool {
> 	return ReasonForError(err) == metav1.StatusReasonExpired
> }
> ```

the err should be a type `metav1.StatusReasonConflict`.

pkg/scheduler/cache/cache.go
```diff
@@ -295,7 +295,25 @@ func (su *defaultStatusUpdater) UpdatePodGroup(pg *schedulingapi.PodGroup) (*sch
 		return nil, err
 	}
 
-	updated, err := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Update(context.TODO(), podgroup, metav1.UpdateOptions{})
+	updated := podgroup.DeepCopy()
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		newer, updateErr := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Update(context.TODO(), podgroup, metav1.UpdateOptions{})
```
googs1025, line 300: How can you ensure that `updateErr != nil` and can be solved by modifying ResourceVersion? Shouldn't it be necessary to determine the error type?

calvin0327: //	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
//	    // Fetch the resource here; you need to refetch it on every try, since
//	    // if you got a conflict on the last update attempt then you need to get
//	    // the current version before making your own changes.
//	    pod, err := c.Pods("mynamespace").Get(name, metav1.GetOptions{})
//	    if err != nil {
//	        return err
//	    }
//
//	    // Make whatever updates to the resource are needed
//	    pod.Status.Phase = v1.PodFailed
//
//	    // Try to update
//	    _, err = c.Pods("mynamespace").UpdateStatus(pod)
//	    // You have to return err itself here (not wrapped inside another error)
//	    // so that RetryOnConflict can identify it correctly.
//	    return err
//	})
//	if err != nil {
//	    // May be conflict if max retries were hit, or may be something unrelated
//	    // like permissions or a network error
//	    return err
//	}
//	...
//
// TODO: Make Backoff an interface?
func RetryOnConflict(backoff wait.Backoff, fn func() error) error {
	return OnError(backoff, errors.IsConflict, fn)
}

In the 'RetryOnConflict' function, it will determine whether it is a Conflict error, and retry if it is.

googs1025: thank.got it.

pkg/scheduler/cache/cache.go
```diff
@@ -295,7 +295,25 @@ func (su *defaultStatusUpdater) UpdatePodGroup(pg *schedulingapi.PodGroup) (*sch
 		return nil, err
 	}
 
-	updated, err := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Update(context.TODO(), podgroup, metav1.UpdateOptions{})
+	updated := podgroup.DeepCopy()
```
lowang-bh, line 298: Does this deep copy need？

calvin0327: indeed, it can be a var, if err returned by RetryOnConflict is nil, It must not be empty.

calvin0327: updated.

calvin0327:
@lowang-bh @hwdef PTAL

hwdef:
Is it useful to update only ResourceVersion? Should the entire resource be updated?

In addition, I think the naming of variables is ambiguous. The new resources obtained should not be `older`. Maybe `newest` is better.

calvin0327:
> Is it useful to update only ResourceVersion? Should the entire resource be updated?
> 
> In addition, I think the naming of variables is ambiguous. The new resources obtained should not be `older`. Maybe `newest` is better.

thanks for you review.
Because of a version conflict, only the resource version needs to be updated.
older is named relative to the resource that is currently being updated.  `newest` is a adective, it doesn't seem very accurate。

Monokaix:
`UpdatePodGroup` will be called when every session closed, I think it will be retryed when there is a conflict when every session closed: )

Monokaix:
Also just modify the resourceversion instead of the field is ok?

hwdef:
`resourceVersion` and `resource` are bound, when you get a new resource and give its resourceVersion to the old one, this operation is very strange. For example, if the resource you recently got has been modified, some fields of it are different from the old resource, but you gave its resourceversion to the old resource. Then the updated fields are lost.

Monokaix:
> `resourceVersion` and `resource` are bound, when you get a new resource and give its resourceVersion to the old one, this operation is very strange. For example, if the resource you recently got has been modified, some fields of it are different from the old resource, but you gave its resourceversion to the old resource. Then the updated fields are lost.

Yeah, this is unacceptable: )

PR #3442, opened 2024-04-26
fix nil pointer panic when evict

bibibox:
i met panic when i use volcano, it shows there are nil pointer deference when excuting the cache's function Evict:

> volcano.sh/volcano@v1.9.0-alpha.0.0.20240425023141-ffd0588b539a/pkg/scheduler/cache/cache.go:865 +0x51c

Maybe we should check if the job.PodGroup is nil before calling Convert function

The more information: #3445


william-wang:
@bibibox Please log a issue and add the panic description and add the reproduce step :)

bibibox:
> @bibibox Please log a issue and add the panic description and add the reproduce step :)

done, this is the issue #3445

lowang-bh (approved):

PR #3216, opened 2023-11-21
queue realcapability change to min dimension of queue capability and cluster realcapability

flyingfang:
issue: #3204

PR #3448, opened 2024-04-29
fix: queue realcapability change to min dimension of queue capability and cluster realcapability 

Monokaix:
Based on https://github.com/volcano-sh/volcano/pull/3216 and fix ci.
Also fix https://github.com/volcano-sh/volcano/issues/3255


william-wang (approved):
/lgtm

Issue #3053, opened 2023-08-14
[good first issue]enrich the unit-test for some package


lowang-bh:
#### What would you like to be added:

There are many packages which don't have unit-test, such as:
- actions: 
  - [x] enqueue
  - [x] backfill

- plugins: 
  - [x] gang
  - [ ] extender
  - [x] overcommit
  - [x] resourcequota
  - [x] usage
- others, include some PRs:

UTs should include positive cases and reverse cases.

#### Why is this needed:

1. to avoid future changes impacting those function, and make code more strong

#### Benefits

1. you can familiar with the coding logic and get started with volcano

#### how to assign

assign only one package to yourself in each pr.

lowang-bh:
/ good first issue
/ help wanted

Subhasish-Behera:
@lowang-bh  I would like to work on unit-tests for `resourcequota`

rayoluo:
@lowang-bh I would like to work on the rest unit-tests

lowang-bh:
Thanks all of you.
/assign @Subhasish-Behera  plugins resourcequota
/assign @rayoluo actions enqueue/backfill; plugins gang/extender/overcommit/usage

lowang-bh:
@Subhasish-Behera @rayoluo Any updates on this issue?

Subhasish-Behera:
@lowang-bh, I'm still working on this issue. Would it be possible to discuss the codebase with you in more detail via email or Slack?

Subhasish-Behera:
@lowang-bh Just filed a PR for the `resourcequota` plugin. any feedback about the PR is welcome.

(cc @william-wang @Monokaix )

lowang-bh:
Now we can use the base UT lib in https://github.com/volcano-sh/volcano/pull/3343 to do the test.
Examples shows at PR #3340  #3341 #3342

googs1025:
@lowang-bh  Hello, may I ask if this will allocate a specific part?

lowang-bh:
> @lowang-bh Hello, may I ask if this will allocate a specific part?

Welcome! You can work on those packages which have no UTs covered.

MondayCha:
@googs1025 Hello, I'm currently looking into Issue #3075 and noticed that there's some overlap with current Issue #3320 regarding _enqueue_ and _backfill_. 

I was wondering if you're planning to address these actions in a future PR?

googs1025:
[...]

PR #3453, opened 2024-05-07
add ut for enqueue action

googs1025:
#### What type of PR is this?
/kind feature

- add ut for action: enqueue

Fixes #https://github.com/volcano-sh/volcano/issues/3053

```release-note

```

### DIFF
--- a/pkg/scheduler/actions/enqueue/enqueue_test.go
+++ b/pkg/scheduler/actions/enqueue/enqueue_test.go
@@ -1,0 +1,151 @@
+package enqueue
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+
+	"volcano.sh/apis/pkg/apis/scheduling"
+	schedulingv1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	"volcano.sh/volcano/cmd/scheduler/app/options"
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/conf"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/plugins/drf"
+	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
+	"volcano.sh/volcano/pkg/scheduler/plugins/proportion"
+	"volcano.sh/volcano/pkg/scheduler/plugins/sla"
+	"volcano.sh/volcano/pkg/scheduler/uthelper"
+	"volcano.sh/volcano/pkg/scheduler/util"
+)
+
+func TestEnqueue(t *testing.T) {
+	plugins := map[string]framework.PluginBuilder{
+		drf.PluginName:        drf.New,
+		gang.PluginName:       gang.New,
+		sla.PluginName:        sla.New,
+		proportion.PluginName: proportion.New,
+	}
+	options.Default()
+	tests := []uthelper.TestCommonStruct{
+		{
+			Name: "when podgroup status is inqueue",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroup("pg1", "c1", "c1", 0, nil, schedulingv1.PodGroupInqueue),
+			},
+			Pods: []*v1.Pod{
+				util.BuildPod("c1", "p1", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg1", make(map[string]string), make(map[string]string)),
+				util.BuildPod("c1", "p2", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg1", make(map[string]string), make(map[string]string)),
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("c1", 1, api.BuildResourceList("4", "4G")),
+				util.BuildQueue("c2", 1, api.BuildResourceList("4", "4G")),
+			},
+			ExpectStatus: map[api.JobID]scheduling.PodGroupPhase{
+				"c1/pg1": scheduling.PodGroupInqueue,
+			},
+		},
+		{
+			Name: "when podgroup status is pending",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroup("pg1", "c1", "c1", 0, nil, schedulingv1.PodGroupPending),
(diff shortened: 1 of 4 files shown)


pkg/scheduler/actions/enqueue/enqueue_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	for i, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			test.Plugins = plugins
+			test.RegisterSession(tiers, nil)
+			defer test.Close()
+			action := New()
+			test.Run([]framework.Action{action})
+			if err := test.CheckAll(i); err != nil {
+				t.Fatalf("case: %s, action: %s, error: %s", test.Name, action.Name(), err)
```
lowang-bh, line 97: err string has include case index and case name.

pkg/scheduler/actions/enqueue/enqueue_test.go
```diff
@@ -0,0 +1,101 @@
...
+				util.BuildPodGroup("pg1", "c1", "c1", 0, nil, schedulingv1.PodGroupInqueue),
+			},
+			Pods: []*v1.Pod{
+				util.BuildPod("c1", "p1", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg1", make(map[string]string), make(map[string]string)),
+				util.BuildPod("c1", "p2", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg1", make(map[string]string), make(map[string]string)),
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("c1", 1, nil),
+				util.BuildQueue("c2", 1, nil),
+			},
```
lowang-bh, line 42: Need wanted  podgroup status in `Status`.

pkg/scheduler/actions/enqueue/enqueue_test.go
googs1025: done

googs1025:
> Are there any cases about job can not enqueue?

I added two type case, one: when pg is in running state, it will not join the queue, and the second one is when the specified queue does not exist, it will not join the queue.

lowang-bh:
How about queue's allocation  plus the incoming inqueue's resource will exceed queue's capacity?

googs1025:
> How about queue's allocation plus the incoming inqueue's resource will exceed queue's capacity?

Yes, I thought about this case. I thought that these scenarios would be covered by the plugins test cases, so I did not add these cases. If needed, I can add it.

lowang-bh:
> > How about queue's allocation plus the incoming inqueue's resource will exceed queue's capacity?
> 
>   If needed, I can add it.

I advise to add some cases.

googs1025:
> > > How about queue's allocation plus the incoming inqueue's resource will exceed queue's capacity?
> > 
> > 
> > If needed, I can add it.
> 
> I advise to add some cases.

done,  i add the case "pggroup cannot enqueue because queue resources are less than podgroup MinResources"

pkg/scheduler/actions/enqueue/enqueue_test.go
```diff
@@ -0,0 +1,146 @@
...
+		{
+			Name: "pggroup cannot enqueue because queue resources are less than podgroup MinResources",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroupWithMinResources("pg1", "c1", "c1", 0,
+					nil, api.BuildResourceList("8", "8G"), schedulingv1.PodGroupPending),
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("c2", 1, api.BuildResourceList("4", "4G")),
+			},
+			Status: map[api.JobID]scheduling.PodGroupPhase{},
```
lowang-bh, line 104: What is the wanted podgroup status? Pending or Inqueue?

pkg/scheduler/actions/enqueue/enqueue_test.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+		{
+			Name: "pggroup cannot enqueue because the specified queue is c1, but there is only c2",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroup("pg1", "c1", "c1", 0, nil, schedulingv1.PodGroupPending),
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("c2", 1, api.BuildResourceList("4", "4G")),
+			},
+			Status: map[api.JobID]scheduling.PodGroupPhase{},
```
lowang-bh, line 93: Pending or Inqueue, or empty string?

googs1025: case: "pggroup cannot enqueue because the specified queue is c1, but there is only c2"
This will be empty string. Because the matching queue will not be matched

pkg/scheduler/actions/enqueue/enqueue_test.go
googs1025: case: pggroup cannot enqueue because queue resources are less than podgroup MinResources
This will be Pending.  Because JobEnqueueable Fn will be false, the status is still pending.
fixed already.

lowang-bh:
/lgtm

@Monokaix @hwdef please also help to take a look. thanks.

pkg/scheduler/actions/enqueue/enqueue_test.go
lowang-bh: We should also check it. But now the wanted status items is empty.

hwdef (commented):
/lgtm

pkg/scheduler/actions/enqueue/enqueue_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	for i, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			test.Plugins = plugins
+			test.RegisterSession(tiers, nil)
+			defer test.Close()
+			action := New()
+			test.Run([]framework.Action{action})
+			if err := test.CheckAll(i); err != nil {
+				t.Fatalf("case: (index: %v/name: %s), action: %s, error: %s", i, test.Name, action.Name(), err)
```
lowang-bh, line 147: duplicate error message。 just need print the error which has included detail.

googs1025: done

lowang-bh (approved):

PR #3454, opened 2024-05-08
bugfix and code refactor around klog

SataQiu:
1. use `cli.Run` for cli commands, `cli.Run` has already integrated support for `--log-flush-frequency`. As a result, the code is simplified without having to deal with more details about klog.  Also, fix the following bugs:
- fix the bug that `--log-flush-frequency` does not take effect for cli commands (`Parse()` is not called, so it is always the default value 5s) .  And，klog has already reduced the default value from 30s to 5s https://github.com/kubernetes/klog/pull/40
- fix the bug that `-v` flag is not available for cli commands

Before:
```bash
$ ./_output/bin/vcctl -h
Usage:
  vcctl [command]

Available Commands:
  help        Help about any command
  job         vcctl command line operation job
  queue       Queue Operations
  version     Print the version information

Flags:
  -h, --help                           help for vcctl
      --log-flush-frequency duration   Maximum number of seconds between log flushes (default 5s)

Use "vcctl [command] --help" for more information about a command.
```
After:
```bash
After:
$  ./_output/bin/vcctl -h
Usage:
  vcctl [command]

Available Commands:
  help        Help about any command
  job         vcctl command line operation job
  queue       Queue Operations
  version     Print the version information

Flags:
  -h, --help                           help for vcctl
      --log-flush-frequency duration   Maximum number of seconds between log flushes (default 5s)
  -v, --v Level                        number for the log level verbosity
      --vmodule moduleSpec             comma-separated list of pattern=N settings for file-filtered logging (only works for the default text log format)

Use "vcctl [command] --help" for more information about a command.
```

2. use `klog.StartFlushDaemon` to config `--log-flush-frequency` instead of starting a new goroutine for controller commands

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/spf13/pflag"
 	_ "go.uber.org/automaxprocs"
-	"k8s.io/apimachinery/pkg/util/wait"
 	cliflag "k8s.io/component-base/cli/flag"
 	componentbaseoptions "k8s.io/component-base/config/options"
 	"k8s.io/klog/v2"
@@ -70,8 +69,7 @@ func main() {
 		}
 	}
 
-	// The default klog flush interval is 30 seconds, which is frighteningly long.
-	go wait.Until(klog.Flush, *logFlushFreq, wait.NeverStop)
+	klog.StartFlushDaemon(*logFlushFreq)
 	defer klog.Flush()
 
 	if err := app.Run(s); err != nil {
--- a/cmd/cli/vcancel/main.go
+++ b/cmd/cli/vcancel/main.go
@@ -16,30 +16,16 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
 	"os"
-	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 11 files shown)


hwdef (commented):
/lgtm

cmd/controller-manager/main.go
```diff
@@ -70,8 +69,7 @@ func main() {
 		}
 	}
 
-	// The default klog flush interval is 30 seconds, which is frighteningly long.
-	go wait.Until(klog.Flush, *logFlushFreq, wait.NeverStop)
+	klog.StartFlushDaemon(*logFlushFreq)
 	defer klog.Flush()
```
Monokaix, line 73: Is this still needed?

SataQiu: Actually, I think it can be removed with `--log-flush-frequency` flag together. 
This is just to preserve the `--log-flush-frequency` flag for compatibility.
Given that all components provide this flag, I think we can just keep it, it's harmless.

Monokaix: Got that~

lowang-bh (approved):

shinytang6 (approved):
Thx
/lgtm

PR #3457, opened 2024-05-09
test: add priority plugin UTs

lowang-bh:
UTs in priority plugin.

### DIFF
--- a/pkg/scheduler/plugins/priority/priority_test.go
+++ b/pkg/scheduler/plugins/priority/priority_test.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2024 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/priority/priority_test.go
```diff
@@ -0,0 +1,154 @@
...
+				},
+				Evicted:   []string{"ns1/preemptee1"},
+				EvictNum:  1,
+				PipeLined: map[string][]string{"ns2/pg2": {"node1"}},
+			},
+		},
+		{
+			tiers: []conf.Tier{
+				{
+					Plugins: []conf.PluginOption{prioPluginDisable},
```
googs1025, line 109: Why do we need to test the disable case?

william-wang:
@lowang-bh It's great to fill the ut gap to improve the coverrage:)

PR #3458, opened 2024-05-09
ignore PredicateFn err info for preempt & reclaim scheduler plugin

LivingCcj:
fix issue: https://github.com/volcano-sh/volcano/issues/3329

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -211,10 +211,7 @@ func preempt(
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
 		// Allows scheduling to nodes that are in Success or Unschedulable state after filtering by predicate.
 		var statusSets util.StatusSets
-		statusSets, err := ssn.PredicateFn(task, node)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -211,10 +211,7 @@ func preempt(
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
 		// Allows scheduling to nodes that are in Success or Unschedulable state after filtering by predicate.
 		var statusSets util.StatusSets
-		statusSets, err := ssn.PredicateFn(task, node)
-		if err != nil {
```
Monokaix, line 215: print the err is better, and we'd better also add some comments to explain why this is removed, like we just concern satus code, which is consistent with kube-scheduler https://github.com/kubernetes/kubernetes/blob/9d87fa215d9e8020abdc17132d1252536cd752d2/pkg/scheduler/framework/preemption/preemption.go#L422

LivingCcj: Yeap, it could place the annotation to mark the reason.  however，print err info, there is no much help for next step.

lowang-bh (approved):

PR #3462, opened 2024-05-11
fix job clone miss PgUID

lowang-bh:
PgUID is missed when clone job.


Monokaix:
Seems that we not use pguid in session cache but just in scheduer cache currently.
But it's also reasonable to do that.

lowang-bh:
@hwdef @william-wang

lowang-bh:
> Seems that we not use pguid in session cache but just in scheduer cache currently. But it's also reasonable to do that.

Year, current it is not used. But when it is used in a cloned job, there will be a bug.

hwdef (commented):
/lgtm

Issue #3293, opened 2024-01-08
Bug in volcano scheduler metrics for queue pod group count 


hy00nc:
**What happened**:

When there is only one job (pod group) running on a queue, and if that job is deleted the metrics does not get updated.
That said, `queue_pod_group_running_count` is `1` when a job is running, and if that job is deleted for some reason, it still remains 1, not updated to 0. Same for other metrics like `queue_pod_group_pending_count`..

**What you expected to happen**:

I expect the `queue_pod_group_running_count` turn to `0`, since the running job has been deleted.

**How to reproduce it (as minimally and precisely as possible)**:

Remove all the existing volcano job or pod groups if any, then follow the instructions below:
1. Try creating a VolcanoJob that sleeps infinitely -> Check if the `queue_pod_group_running_count` has been updated to 1.
2. Try deleting that VolcanoJob -> Check the `queue_pod_group_running_count` if it is still 1.

**Environment**:
- Volcano Version: v1.8.0

hy00nc:
Seems the the metrics updating is not triggered when there is zero job in [here](https://github.com/volcano-sh/volcano/blob/67cabf78a6a50751287eecedd8e050c5977ebb40/pkg/scheduler/plugins/proportion/proportion.go#L168-L178).
Perhaps we have to add a code to set default value zero for all the queue-related metrics and then trigger the metrics update if there is any job.

lowang-bh:
> Seems the metrics update is not triggered when there is zero job

Yeah，current code can not trigger metrics update if no jobs in queue.

Monokaix:
> > Seems the metrics update is not triggered when there is zero job
> 
> Yeah，current code can not trigger metrics update if no jobs in queue.

Maybe we can move these codes to `OnSessionClose` to cover all queues' metrics.
https://github.com/volcano-sh/volcano/blob/67cabf78a6a50751287eecedd8e050c5977ebb40/pkg/scheduler/plugins/proportion/proportion.go#L168-L178

lowang-bh:
It will still has same problems, because those queues who has no jobs in them will not be iterated using `for _, attr := range pp.queueOpts`

Monokaix:
> It will still has same problems, because those queues who has no jobs in them will not be iterated using `for _, attr := range pp.queueOpts`

If no job in queue, itwill be zero, right？

lowang-bh:
[...]

PR #3463, opened 2024-05-13
Fix queue metrics when there are no jobs in it

Monokaix:
fixed https://github.com/volcano-sh/volcano/issues/3293

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -160,22 +160,25 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			attr.name, attr.allocated.String(), attr.request.String(), attr.inqueue.String(), attr.elastic.String())
 	}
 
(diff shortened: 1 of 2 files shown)


hwdef (commented):
/lgtm

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -160,22 +160,25 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			attr.name, attr.allocated.String(), attr.request.String(), attr.inqueue.String(), attr.elastic.String())
 	}
 
+	// Record metrics
```
lowang-bh, line 163: How about put it at where the Deserved has been calculated? so that to include Deserved metrics.

Monokaix: Deserved metrics are calculated in a loop，it's not convenient to put them together.

william-wang (approved):
/approve

PR #3466, opened 2024-05-13
fix gpu memory allocation using `vgpu-memory-percentage`

### LINKED ISSUE #3465: gpu memory incorretly allocated using `vgpu-memory-percentage` resource name
**What happened**:

I'm using volcano-vgpu-device-plugin with Volcano, when i'm trying to allocate vgpu with resource:

```
volcano.sh/vgpu-number: 1
volcano.sh/vgpu-cores: 50
volcano.sh/vgpu-memory-percentage: 20
```

The allocation results in Pod annotations show `0` memory was allocated, which is incorrect.

Volcano scheduler logs： 

image

The allocated memory should be `totalMemory * percetage / 100` but not `0`

**Environment**:
- Volcano Version: v1.8.2
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

abowloflrf:
fix #3465 

`uint(val.MemPercentagereq/100)` could be `0`. The memory result is always `0` when using percentage to allocate memory.


archlitchi:
thanks for fixing that, we plan to test that function before officially release

abowloflrf:
> thanks for fixing that, we plan to test that function before officially release

ok, i'll also add some ut for this pr later

abowloflrf:
> > thanks for fixing that, we plan to test that function before officially release
> 
> ok, i'll also add some ut for this pr later

ut added /cc @archlitchi

Monokaix:
@archlitchi Is this already supported?

cybergeek2077:
I've encountered this issue as well; why not merge this PR?

pkg/scheduler/api/devices/nvidia/vgpu/utils.go
```diff
@@ -356,7 +356,7 @@ func checkNodeGPUSharingPredicate(pod *v1.Pod, gssnap *GPUDevices, replicate boo
 				continue
 			}
 			if val.MemPercentagereq != 101 && val.Memreq == 0 {
-				val.Memreq = int32(gs.Device[i].Memory * uint(val.MemPercentagereq/100))
+				val.Memreq = int32(float64(gs.Device[i].Memory) * float64(val.MemPercentagereq) / 100.0)
```
cybergeek2077, line 359: I recommend using this implementation: `val.Memreq = int32(gs.Device[i].Memory) * val.MemPercentagereq / 100`

Issue #3478, opened 2024-05-21
Concurrent read and write operations on the map caused the scheduler to panic


Wang-Kai:
**What happened**:

The scheduler encountered an unexpected panic during operation.

<img width="1078" alt="image" src="https://github.com/volcano-sh/volcano/assets/13144386/f19c1f05-54b4-4623-8394-74f4db31c34f">

**What you expected to happen**:

The scheduler is running normally.

**How to reproduce it (as minimally and precisely as possible)**:

This issue occurs sporadically.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: v1.8.2
- Kubernetes version (use `kubectl version`): 1.20
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release): debian9
- Kernel (e.g. `uname -a`): linux5.1
- Install tools:
- Others:

PR #3479, opened 2024-05-21
add lock before read Jobs map #3478

Wang-Kai:
add lock before read Jobs map in syncTask function

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -258,6 +258,8 @@ func (sc *SchedulerCache) syncTask(oldTask *schedulingapi.TaskInfo) error {
 	newPod, err := sc.kubeClient.CoreV1().Pods(oldTask.Namespace).Get(context.TODO(), oldTask.Name, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/cache/event_handlers.go
```diff
@@ -258,6 +258,8 @@ func (sc *SchedulerCache) syncTask(oldTask *schedulingapi.TaskInfo) error {
 	newPod, err := sc.kubeClient.CoreV1().Pods(oldTask.Namespace).Get(context.TODO(), oldTask.Name, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
+			sc.Mutex.Lock()
```
googs1025, line 261: I'm not sure if we need to use lock in notfound error as well.  this error is not common, but it works if we add it. 
nice catch! :)

Monokaix:
Hi, please sign off you commit with `git commit -s`.

Wang-Kai:
> Hi, please sign off you commit with `git commit -s`.

Sorry for the delay. I just executed the `git commit --amend -s` command and pushed again.

lowang-bh:
Please rebase and squash commits to only one.

william-wang (approved):
/approve

Issue #3476, opened 2024-05-18
CI scirpt run-e2e-kind.sh has error need to handle


lowang-bh:
**What happened**:
<img width="1085" alt="image" src="https://github.com/volcano-sh/volcano/assets/21003791/b7be4d1a-dbe9-44c3-bbae-0c004b6f7415">

https://github.com/volcano-sh/volcano/actions/runs/9136335780/job/25125000321?pr=3457
```sh
Checking helm
Found helm, version: version.BuildInfo{Version:"v3.14.4", GitCommit:"81c902a123462fd4052bc5e9aa9c513c4c8fc142", GitTreeState:"clean", GoVersion:"go1.21.9"}
error: unknown flag: --short
See 'kubectl version --help' for usage.
./hack/run-e2e-kind.sh: line 46: [: : integer expression expected
Ensure create namespace
namespace/volcano-system created
namespace/volcano-monitoring created
Install volcano chart with crd version v1
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #3480, opened 2024-05-22
fix bug3476

wangyysde:
fix bug #3476

Note for reviewers:
    As --short flag of kubectl has be  deprecated.  We should get the version information of server by another way(by commands directly in this PR).
Then I have deleted the function named get-k8s-server-version  in hack/run-e2e-kind.sh.

### DIFF
--- a/hack/run-e2e-kind.sh
+++ b/hack/run-e2e-kind.sh
@@ -30,17 +30,13 @@ export CLUSTER_CONTEXT="--name ${CLUSTER_NAME}"
 
 export KIND_OPT=${KIND_OPT:="--config ${VK_ROOT}/hack/e2e-kind-config.yaml"}
 
-function get-k8s-server-version {
-    echo $(kubectl version --short=true | grep Server | sed "s/.*: v//" | tr "." " ")
-}
 
 function install-volcano {
   install-helm
 
   # judge crd version
-  serverVersion=($(get-k8s-server-version))
-  major=${serverVersion[0]}
-  minor=${serverVersion[1]}
+  major=$(kubectl version --output yaml | awk '/serverVersion/,0' |grep -E 'major:' | awk '{print $2}' | tr "\"" " ")
+  minor=$(kubectl version --output yaml | awk '/serverVersion/,0' |grep -E 'minor:' | awk '{print $2}' | tr "\"" " ")
   crd_version="v1"
   # if k8s version less than v1.18, crd version use v1beta
   if [ "$major" -le "1" ]; then
--- a/hack/lib/install.sh
+++ b/hack/lib/install.sh
@@ -59,7 +59,7 @@ function check-prerequisites {
     echo -e "\033[31mERROR\033[0m: kubectl not installed"
     exit 1
   else
-    echo -n "Found kubectl, version: " && kubectl version --short --client
+    echo -n "Found kubectl, version: " && kubectl version --client
   fi
 }
 


hack/run-e2e-kind.sh
```diff
@@ -30,17 +30,28 @@ export CLUSTER_CONTEXT="--name ${CLUSTER_NAME}"
 
 export KIND_OPT=${KIND_OPT:="--config ${VK_ROOT}/hack/e2e-kind-config.yaml"}
 
-function get-k8s-server-version {
-    echo $(kubectl version --short=true | grep Server | sed "s/.*: v//" | tr "." " ")
+function check-jq {
```
hwdef, line 33: What is the relationship between jq version and k8s version?

wangyysde: Sorry, I don't quite understand what you mean.
1. we get server version information of cluster by kubectl version --short=true in function get-k8s-server-version now.
2. but --short flag of kubectl has be deprecated.So we get an error as issue #3476 shown when we run e2e test.
3.  then I think that we should get the version information of server by another way.
4. jq is a json parser in shell environments. And it is written in C.
5. So the version of jq is not related to the version of kubernetes.

hack/run-e2e-kind.sh
```diff
@@ -30,17 +30,28 @@ export CLUSTER_CONTEXT="--name ${CLUSTER_NAME}"
...
 
 function install-volcano {
   install-helm
+  check-jq
 
   # judge crd version
-  serverVersion=($(get-k8s-server-version))
-  major=${serverVersion[0]}
-  minor=${serverVersion[1]}
+  major=$(kubectl version --output=json |jq '.serverVersion.major' | tr "\"" " ")
```
hwdef, line 53: kubectl version --output yaml | awk '/serverVersion/,0' | grep -E 'major:' | awk '{print $2}' | tr "\"" " "

hack/run-e2e-kind.sh
```diff
@@ -30,17 +30,28 @@ export CLUSTER_CONTEXT="--name ${CLUSTER_NAME}"
...
+  minor=$(kubectl version --output=json |jq '.serverVersion.minor' | tr "\"" " ")
```
hwdef, line 54: kubectl version --output yaml | awk '/serverVersion/,0' | grep -E 'minor:' | awk '{print $2}' | tr "\"" " "

hwdef:
No need to install additional software

hack/run-e2e-kind.sh
wangyysde: very good. I have remade the PR without jq.

hack/run-e2e-kind.sh
wangyysde: very good. I have remade the PR without jq.

wangyysde:
> No need to install additional software

thks, I have remade the PR without jq.

hwdef (commented):
/lgtm
Thanks!

hack/run-e2e-kind.sh
```diff
@@ -30,17 +30,13 @@ export CLUSTER_CONTEXT="--name ${CLUSTER_NAME}"
 
 export KIND_OPT=${KIND_OPT:="--config ${VK_ROOT}/hack/e2e-kind-config.yaml"}
 
-function get-k8s-server-version {
-    echo $(kubectl version --short=true | grep Server | sed "s/.*: v//" | tr "." " ")
```
Monokaix, line 34: How adout just remove `--short`?

Monokaix:
Please also fix the err: ./hack/run-e2e-kind.sh: line 46: [: : integer expression expected
And also update description.

hack/run-e2e-kind.sh
wangyysde: we can not get major and minor number if just remove --short：
we get client version along with server version  by newer kubectl.  So we should get major and minor of server version from these information by shell or commands.

wangyysde:
> Please also fix the err: ./hack/run-e2e-kind.sh: line 46: [: : integer expression expected And also update description.

This PR has fixed the error.
It is the reason for the error that we can not get major and minor with new kubectl.

Monokaix:
> > Please also fix the err: ./hack/run-e2e-kind.sh: line 46: [: : integer expression expected And also update description.
> 
> This PR has fixed the error. It is the reason for the error that we can not get major and minor with new kubectl.

OK, great

william-wang:
Trigger the ci.

william-wang (approved):

PR #3483, opened 2024-05-24
Mark --lock-object-namespace as deprecated and throw a warning when it is used

SataQiu:
A warning is thrown when the deprecated flag `--lock-object-namespace` is used

```sh
% ./_output/bin/vc-scheduler --lock-object-namespace example-namespace ...                                      
Flag --lock-object-namespace has been deprecated, This flag is deprecated and will be removed in a future release. Please use --leader-elect-resource-namespace instead.
...
```

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -112,7 +112,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 		"File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated "+
 		"after server cert).")
 	fs.StringVar(&s.KeyFile, "tls-private-key-file", s.KeyFile, "File containing the default x509 private key matching --tls-cert-file.")
-	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default, will deprecated, please use --leader-elect-resource-namespace instead.")
+	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default.")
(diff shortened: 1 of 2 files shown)


cmd/scheduler/app/options/options.go
```diff
@@ -112,7 +112,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 		"File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated "+
 		"after server cert).")
 	fs.StringVar(&s.KeyFile, "tls-private-key-file", s.KeyFile, "File containing the default x509 private key matching --tls-cert-file.")
-	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default, will deprecated, please use --leader-elect-resource-namespace instead.")
+	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", defaultLockObjectNamespace, "Define the namespace of the lock object; it is volcano-system by default.")
```
Monokaix, line 115: Is this still needed?

SataQiu: Thanks @Monokaix 
As a best practice, we need to keep the deprecated flag for a while (maybe about 3 releases) and then can remove it safely. We need to give users enough time to migrate to new flag.
It was deprecated by https://github.com/volcano-sh/volcano/pull/2809 for the first time, just only two months since 2024.04.01.

Monokaix: Yeah, I understand: ) I mean whether just keep `fs.MarkDeprecated("lock-object-namespace"` and remove `fs.StringVar(&s.LockObjectNamespace,` is ok?

SataQiu: We can't do that. `fs.MarkDeprecated` cannot be used independently, it depends on the `fs.StringVar` declaration.

Monokaix: Got that!

hwdef (commented):
/approve

lowang-bh (approved):

william-wang (approved):

Issue #3485, opened 2024-05-24
Add controller switch params.


Monokaix:
#### What would you like to be added:
Add a switch params in vc-controller to determine which controller is enabled and which is disabled.
#### Why is this needed:
some times we do not need start all controllers like jobflow and jobtemplate, so we can disable them to reduce unnecessary memory and cpu consuming.

Monokaix:
We can refer to: https://github.com/kubernetes/kubernetes/blob/8361522b40cc8b569efdd6ee2456fa514071cad1/staging/src/k8s.io/controller-manager/app/helper.go#L60

googs1025:
hey @Monokaix  I would also like to ask if this option should be put into the configuration file?

Monokaix:
> hey @Monokaix I would also like to ask if this option should be put into the flags?
> 
> *: Start all controllers by default. +job,queue: Only start the job and queue controllers. -job,queue: Do not start the job and queue controllers.
> 
> ```go
> flag.StringVar(&controllerGates, "controller-gates", "*", "Specify controller gates. Use '*' for all controllers, '+controller1,controller2' to start specific controllers, and '-controller1,controller2' to stop specific controllers.")
> ```

It's ok : ）

PR #3486, opened 2024-05-24
add controller gate flag

googs1025:
#### What type of PR is this?
/kind feature

- add controller gate flag

Fix https://github.com/volcano-sh/volcano/issues/3485

```release-note
None
```

```docs
None
```

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -19,6 +19,7 @@ package options
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -38,6 +39,7 @@ const (
 	defaultLockObjectNamespace = "volcano-system"
 	defaultPodGroupWorkers     = 5
 	defaultGCWorkers           = 1
+	defaultControllers         = "*"
 )
 
 // ServerOption is the main context object for the controllers.
@@ -75,6 +77,13 @@ type ServerOption struct {
 	// WorkerThreadsForGC is the number of threads for recycling jobs
 	// The larger the number, the faster the job recycling, but requires more CPU load.
 	WorkerThreadsForGC uint32
+	// Controllers specify controllers to set up.
+	// Case1: Use '*' for all controllers,
+	// Case2: "+gc-controller,+job-controller,+jobflow-controller,+jobtemplate-controller,+pg-controller,+queue-controller"
+	// to enable specific controllers,
+	// Case3: "-gc-controller,-job-controller,-jobflow-controller,-jobtemplate-controller,-pg-controller,-queue-controller"
+	// to disable specific controllers,
+	Controllers []string
 }
 
 type DecryptFunc func(c *ServerOption) error
@@ -85,7 +94,7 @@ func NewServerOption() *ServerOption {
 }
 
 // AddFlags adds flags for a specific CMServer to the specified FlagSet.
-func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
+func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.StringVar(&s.KubeClientOptions.Master, "master", s.KubeClientOptions.Master, "The address of the Kubernetes API server (overrides any value in kubeconfig)")
 	fs.StringVar(&s.KubeClientOptions.KubeConfig, "kubeconfig", s.KubeClientOptions.KubeConfig, "Path to kubeconfig file with authorization and master location information.")
 	fs.StringVar(&s.CaCertFile, "ca-cert-file", s.CaCertFile, "File containing the x509 Certificate for HTTPS.")
@@ -107,11 +116,44 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
+	fs.StringSliceVar(&s.Controllers, "controllers", []string{defaultControllers}, fmt.Sprintf("Specify controller gates. Use '*' for all controllers, all knownController: %s ,and we can use "+
+		"'-' to disable controllers, e.g. \"-job-controller,-queue-controller\" to disable job and queue controllers.", knownControllers))
 }
 
-// CheckOptionOrDie check leader election flag when LeaderElection is enabled.
+// CheckOptionOrDie checks the option and returns error if it's invalid
 func (s *ServerOption) CheckOptionOrDie() error {
+	// check controllers option
+	if err := s.checkControllers(); err != nil {
+		return err
+	}
+	// check leader election flag when LeaderElection is enabled.
 	return componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
+}
+
+// checkControllers checks the controllers option and returns error if it's invalid
+func (s *ServerOption) checkControllers() error {
+	existenceMap := make(map[string]bool)
+	for _, c := range s.Controllers {
+		if c == "*" {
+			// wildcard '*' is not allowed to be combined with other input
+			if len(s.Controllers) > 1 {
+				return fmt.Errorf("wildcard '*' cannot be combined with other input")
+			}
+		} else {
+			if strings.HasPrefix(c, "-") || strings.HasPrefix(c, "+") {
+				if existenceMap[c[1:]] {
+					return fmt.Errorf("controllers option %s cannot have both '-' and '+' prefixes", c)
+				}
+				existenceMap[c[1:]] = true
+			} else {
+				if existenceMap[c] {
+					return fmt.Errorf("controllers option %s cannot have both '-' and '+' prefixes", c)
+				}
+				existenceMap[c] = true
+			}
+		}
+	}
+	return nil
 }
 
 // readCAFiles read data from ca file path
--- a/cmd/controller-manager/app/server.go
+++ b/cmd/controller-manager/app/server.go
@@ -130,6 +130,11 @@ func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx c
 
 	return func(ctx context.Context) {
 		framework.ForeachController(func(c framework.Controller) {
+			// if controller is not enabled, skip it
+			if !isControllerEnabled(c.Name(), opt.Controllers) {
+				klog.Infof("Controller <%s> is not enable", c.Name())
+				return
+			}
 			if err := c.Initialize(controllerOpt); err != nil {
 				klog.Errorf("Failed to initialize controller <%s>: %v", c.Name(), err)
 				return
@@ -140,4 +145,34 @@ func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx c
 
 		<-ctx.Done()
 	}
+}
+
+// isControllerEnabled check if a specified controller enabled or not.
+// If the input controllers starts with a "+name" or "name", it is considered as an explicit inclusion.
+// Otherwise, it is considered as an explicit exclusion.
+func isControllerEnabled(name string, controllers []string) bool {
+	hasStar := false
+	// if no explicit inclusion or exclusion, enable all controllers by default
+	if len(controllers) == 0 {
+		return true
+	}
+	for _, ctrl := range controllers {
+		// if we get here, there was an explicit inclusion
+		if ctrl == name {
+			return true
+		}
+		// if we get here, there was an explicit inclusion
+		if ctrl == "+"+name {
+			return true
+		}
+		// if we get here, there was an explicit exclusion
+		if ctrl == "-"+name {
+			return false
+		}
+		if ctrl == "*" {
+			hasStar = true
+		}
+	}
+	// if we get here, there was no explicit inclusion or exclusion
+	return hasStar
 }
--- a/cmd/controller-manager/app/options/options_test.go
+++ b/cmd/controller-manager/app/options/options_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package options
 
 import (
+	"fmt"
+	"sort"
 	"testing"
 	"time"
 
@@ -30,6 +32,13 @@ import (
 	componentbaseoptions "k8s.io/component-base/config/options"
 	"k8s.io/component-base/featuregate"
 
+	"volcano.sh/volcano/pkg/controllers/framework"
+	_ "volcano.sh/volcano/pkg/controllers/garbagecollector"
+	_ "volcano.sh/volcano/pkg/controllers/job"
+	_ "volcano.sh/volcano/pkg/controllers/jobflow"
+	_ "volcano.sh/volcano/pkg/controllers/jobtemplate"
+	_ "volcano.sh/volcano/pkg/controllers/podgroup"
+	_ "volcano.sh/volcano/pkg/controllers/queue"
 	"volcano.sh/volcano/pkg/features"
 	"volcano.sh/volcano/pkg/kube"
 	commonutil "volcano.sh/volcano/pkg/util"
@@ -42,7 +51,17 @@ func TestAddFlags(t *testing.T) {
 	commonutil.LeaderElectionDefault(&s.LeaderElection)
 	s.LeaderElection.ResourceName = "vc-controller-manager"
 	componentbaseoptions.BindLeaderElectionFlags(&s.LeaderElection, fs)
-	s.AddFlags(fs)
+	// knownControllers is a list of all known controllers.
+	var knownControllers = func() []string {
+		controllerNames := []string{}
+		fn := func(controller framework.Controller) {
+			controllerNames = append(controllerNames, controller.Name())
+		}
+		framework.ForeachController(fn)
+		sort.Strings(controllerNames)
+		return controllerNames
+	}
+	s.AddFlags(fs, knownControllers())
 	utilfeature.DefaultMutableFeatureGate.AddFlag(fs)
 
 	args := []string{
@@ -83,6 +102,7 @@ func TestAddFlags(t *testing.T) {
 		LockObjectNamespace: defaultLockObjectNamespace,
 		WorkerThreadsForPG:  5,
 		WorkerThreadsForGC:  1,
+		Controllers:         []string{"*"},
 	}
 	expectedFeatureGates := map[featuregate.Feature]bool{features.ResourceTopology: false}
 
@@ -97,5 +117,53 @@ func TestAddFlags(t *testing.T) {
 	err := s.CheckOptionOrDie()
 	if err != nil {
 		t.Errorf("expected nil but got %v\n", err)
+	}
+}
+
+func TestCheckControllers(t *testing.T) {
+	testCases := []struct {
+		name         string
+		serverOption *ServerOption
+		expectErr    error
+	}{
+		{
+			name: "normal case: use *",
+			serverOption: &ServerOption{
+				Controllers: []string{"*"},
+			},
+			expectErr: nil,
+		},
+		{
+			name: "normal case: use specific controller",
+			serverOption: &ServerOption{
+				Controllers: []string{"+gc-controller", "+jobtemplate-controller", "+jobflow-controller"},
+			},
+			expectErr: nil,
+		},
+		{
+			name: "fail case: use duplicate job-controller",
+			serverOption: &ServerOption{
+				Controllers: []string{"+gc-controller", "+job-controller", "-job-controller"},
+			},
+			expectErr: fmt.Errorf("controllers option %s cannot have both '-' and '+' prefixes", "-job-controller"),
+		},
+		{
+			name: "fail case: use * but combined with other input",
+			serverOption: &ServerOption{
+				Controllers: []string{"*", "+job-controller"},
+			},
+			expectErr: fmt.Errorf("wildcard '*' cannot be combined with other input"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.serverOption.checkControllers()
+			if err != nil {
+				if err.Error() != tc.expectErr.Error() {
+					t.Errorf("test case %s failed: expected: %v, but got: %v", tc.name, tc.expectErr, err)
+				}
+			}
+		})
 	}
 }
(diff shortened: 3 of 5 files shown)


cmd/controller-manager/app/options/options.go
```diff
@@ -106,6 +111,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
+	fs.StringVar(&s.ControllerGates, "controller-gates", "*", "Specify controller gates. Use '*' for all controllers, '+queue-controller,pg-controller' to start specific controllers, and '-jobflow-controller,jobtemplate-controller' to stop specific controllers.")
```
Monokaix, line 114: `StringSliceVar` is better.

Monokaix: And we should all print all controllers here to let uer know what controllers we have.

pkg/controllers/framework/factory.go
```diff
@@ -18,19 +18,49 @@ package framework
...
 	for _, ctrl := range controllers {
-		fn(ctrl)
+		// if controllerGate is "*", all controllers will be executed.
+		if controllerGate == "*" || checkControllerGate(controllerGate, ctrl.Name()) {
+			fn(ctrl)
+		}
 	}
 }
 
+func checkControllerGate(controllerGate, controllerName string) bool {
```
Monokaix, line 38: A little complicated here, you can refer the mentioned link of the issue: )

cmd/controller-manager/app/options/options.go
```diff
@@ -75,6 +75,11 @@ type ServerOption struct {
 	// WorkerThreadsForGC is the number of threads for recycling jobs
 	// The larger the number, the faster the job recycling, but requires more CPU load.
 	WorkerThreadsForGC uint32
+	// ControllerGates Specify controller gates.
+	// Use '*' for all controllers,
+	// '+controller1,controller2' to start specific controllers,
+	// and '-controller1,controller2' to stop specific controllers.
+	ControllerGates string
```
Monokaix, line 82: Controllers is ok.

cmd/controller-manager/app/options/options.go
googs1025: done

pkg/controllers/framework/factory.go
googs1025: done

cmd/controller-manager/app/server.go
```diff
@@ -127,7 +127,7 @@ func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx c
 	controllerOpt.WorkerThreadsForGC = opt.WorkerThreadsForGC
 
 	return func(ctx context.Context) {
-		framework.ForeachController(func(c framework.Controller) {
+		framework.ForeachController(opt.ControllerGates, func(c framework.Controller) {
```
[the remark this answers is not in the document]
googs1025: done

cmd/controller-manager/app/options/options.go
googs1025: done

cmd/controller-manager/app/options/options.go
```diff
@@ -106,6 +113,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
+	fs.StringSliceVar(&s.Controllers, "controller-gates", []string{"*"}, "Specify controller gates. Use '*' for all controllers, \"+gc-controller,+job-controller,+jobflow-controller,+jobtemplate-controller,+pg-controller,+queue-controller\" to start specific controllers, "+
+		"and \"-gc-controller,-job-controller,-jobflow-controller,-jobtemplate-controller,-pg-controller,-queue-controller\" to stop specific controllers.")
```
Monokaix, line 117: start -> enable
stop -> disable

cmd/controller-manager/app/server.go
```diff
@@ -139,3 +144,29 @@ func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx c
 		<-ctx.Done()
 	}
 }
+
+// isControllerEnabled check if a specified controller enabled or not.
+// If the input controllers starts with a "+name" or "name", it is considered as an explicit inclusion.
+// Otherwise, it is considered as an explicit exclusion.
+func isControllerEnabled(name string, controllers []string) bool {
```
Monokaix, line 151: We should also add ut here: )

cmd/controller-manager/app/options/options.go
```diff
@@ -106,6 +113,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
+	fs.StringSliceVar(&s.Controllers, "controller-gates", []string{"*"}, "Specify controller gates. Use '*' for all controllers, \"+gc-controller,+job-controller,+jobflow-controller,+jobtemplate-controller,+pg-controller,+queue-controller\" to start specific controllers, "+
```
Monokaix, line 116: We should wrap a methed like `KnownControllers` which will call `ForeachController` to get all controllers' name dynamically : )

googs1025: Sorry, I don't quite understand what you mean.
I use `isControllerEnabled` to determine whether a specific controller needs to be started.

```go
func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx context.Context) {
	controllerOpt := &framework.ControllerOption{}

	controllerOpt.SchedulerNames = opt.SchedulerNames
	controllerOpt.WorkerNum = opt.WorkerThreads
	controllerOpt.MaxRequeueNum = opt.MaxRequeueNum

	// TODO: add user agent for different controllers
	controllerOpt.KubeClient = kubeclientset.NewForConfigOrDie(config)
	controllerOpt.VolcanoClient = vcclientset.NewForConfigOrDie(config)
	controllerOpt.SharedInformerFactory = informers.NewSharedInformerFactory(controllerOpt.KubeClient, 0)
	controllerOpt.InheritOwnerAnnotations = opt.InheritOwnerAnnotations
	controllerOpt.WorkerThreadsForPG = opt.WorkerThreadsForPG
	controllerOpt.WorkerThreadsForGC = opt.WorkerThreadsForGC

	return func(ctx context.Context) {
		framework.ForeachController(func(c framework.Controller) {
			// if controller is not enabled, skip it
			if !isControllerEnabled(c.Name(), opt.Controllers) {
				klog.Infof("Controller <%s> is not enabled", c.Name())
				return
			}
			if err := c.Initialize(controllerOpt); err != nil {
				klog.Errorf("Failed to initialize controller <%s>: %v", c.Name(), err)
				return
			}

			go c.Run(ctx.Done())
		})

		<-ctx.Done()
	}
}

```

Monokaix: Yeah，it's ok to your implement，I mean we should call a method in flag description，the method will print all controllers，instead of directly write all controllers in a hard code way, you can refer to kube-controller-manager's flag
description.

cmd/controller-manager/app/server.go
googs1025: done

cmd/controller-manager/app/options/options.go
googs1025: done

cmd/controller-manager/app/options/options.go
googs1025: OK, I understand. Thanks for your suggestion.

googs1025: I have read the code related to kube-controller-manager. Do we need to use a similar method?
```go
// AddFlags adds flags related to generic for controller manager to the specified FlagSet.
func (o *GenericControllerManagerConfigurationOptions) AddFlags(fss *cliflag.NamedFlagSets, allControllers, disabledByDefaultControllers []string, controllerAliases map[string]string) {
	if o == nil {
		return
	}

	o.Debugging.AddFlags(fss.FlagSet("debugging"))
	o.LeaderMigration.AddFlags(fss.FlagSet("leader-migration"))
	genericfs := fss.FlagSet("generic")
	genericfs.DurationVar(&o.MinResyncPeriod.Duration, "min-resync-period", o.MinResyncPeriod.Duration, "The resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod.")
	genericfs.StringVar(&o.ClientConnection.ContentType, "kube-api-content-type", o.ClientConnection.ContentType, "Content type of requests sent to apiserver.")
	genericfs.Float32Var(&o.ClientConnection.QPS, "kube-api-qps", o.ClientConnection.QPS, "QPS to use while talking with kubernetes apiserver.")
	genericfs.Int32Var(&o.ClientConnection.Burst, "kube-api-burst", o.ClientConnection.Burst, "Burst to use while talking with kubernetes apiserver.")
	genericfs.DurationVar(&o.ControllerStartInterval.Duration, "controller-start-interval", o.ControllerStartInterval.Duration, "Interval between starting controller managers.")
	genericfs.StringSliceVar(&o.Controllers, "controllers", o.Controllers, fmt.Sprintf(""+
		"A list of controllers to enable. '*' enables all on-by-default controllers, 'foo' enables the controller "+
		"named 'foo', '-foo' disables the controller named 'foo'.\nAll controllers: %s\nDisabled-by-default controllers: %s",
		strings.Join(allControllers, ", "), strings.Join(disabledByDefaultControllers, ", ")))

	options.BindLeaderElectionFlags(&o.LeaderElection, genericfs)
}

```

Monokaix: Yeah we need a method to get all controllers name list, and we can new a func like `KnownControllers`, and we can use `ForeachController` to do that.

Monokaix: Just like this:
```go
func KnownControllers() []string {
	controllerNames := []string{}
	fn := func(controller framework.Controller) {
		controllerNames = append(controllerNames, controller.Name())
	}
	framework.ForeachController(fn)
	return controllerNames
}
```

googs1025: OK, I will make some changes in the next few days.

cmd/controller-manager/app/options/options.go
```diff
@@ -106,6 +121,19 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
+	fs.StringSliceVar(&s.Controllers, "controllers", KnownControllers(), "Specify controller gates. Use '*' for all controllers, \"+gc-controller,+job-controller,+jobflow-controller,+jobtemplate-controller,+pg-controller,+queue-controller\" to enable specific controllers, "+
+		"and \"-gc-controller,-job-controller,-jobflow-controller,-jobtemplate-controller,-pg-controller,-queue-controller\" to disable specific controllers.")
+}
+
+// KnownControllers returns a list of known controllers.
+func KnownControllers() []string {
```
Monokaix, line 129: Maybe move the func to `cmd/controller-manager/main.go` is better, because it's where the controllers are registered, and we should remove importing controllers in `cmd/controller-manager/app/options/options.go`.

Monokaix: Hi, please also notice this piece, we can merge it after this is done: ）

googs1025: I will deal with this issue in the next two days. Thanks for the reminder.

googs1025: In main.go, I use the knownControllers function variable to ensure the simplicity of main.go
```go
// knownControllers is a list of all known controllers.
	var knownControllers = func() []string {
		controllerNames := []string{}
		fn := func(controller framework.Controller) {
			controllerNames = append(controllerNames, controller.Name())
		}
		framework.ForeachController(fn)
		sort.Strings(controllerNames)
		return controllerNames
	}

```

googs1025:
```yaml
kind: Deployment
apiVersion: apps/v1
     ...
    spec:
     ...
      containers:
          - name: {{ .Release.Name }}-controllers
            ...
            args:
              ...
              - --controllers=<we can use custom controller names here>
              - -v=4
             ...
{{- end }}
```
There are several ways to use it:
- --controllers: 
    - If none of them are set, all controllers are started by default.
    - "*": `- --controllers="*"`, represents startup of all controllers
    - `- --controllers="+gc-controller,+job-controller,+pg-controller,+queue-controller"`, represents startup only gc-controller job-controller pg-controller, queue-controller

lowang-bh:
> * If none of them are set, all controllers are started by default

How about start the most necessary controllers set such as gc,job,pg,queue.

googs1025:
> > * If none of them are set, all controllers are started by default
> 
> How about start the most necessary controllers set such as gc,job,pg,queue.

Certainly. However, it might be different from before, which could confuse users since we are uncertain about the scope of the impact. To address this, we should add a toggle setting while keeping the existing functionality unchanged, allowing users to choose whether to enable it or not. IMO, it's better to keep the status quo. However, this can be discussed further to reach a conclusive decision.

Monokaix:
> > > * If none of them are set, all controllers are started by default
> > 
> > 
> > How about start the most necessary controllers set such as gc,job,pg,queue.
> 
> Certainly, it is ok. However, it might be different from before, which could confuse users since we are uncertain about the scope of the impact. To address this, we should add a toggle setting while keeping the existing functionality unchanged, allowing users to choose whether to enable it or not. IMO, it's better to keep the status quo. However, this can be discussed further to reach a conclusive decision.

yeah I agree with you.

cmd/controller-manager/app/options/options.go
```diff
@@ -107,6 +114,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
+	fs.StringSliceVar(&s.Controllers, "controllers", controllerNames, "Specify controller gates. Use '*' for all controllers, \"+gc-controller,+job-controller,+jobflow-controller,+jobtemplate-controller,+pg-controller,+queue-controller\" to enable specific controllers, "+
```
Monokaix, line 117: Maybe the default value `*` is better of this flag and print all the known controllers in the description: )

googs1025: ok, i will change it. And the usage is explained in flag usage.
```go
fs.StringSliceVar(&s.Controllers, "controllers", []string{"*"}, "Specify controller gates. Use '*' for all controllers, \"+gc-controller,+job-controller,+jobflow-controller,+jobtemplate-controller,+pg-controller,+queue-controller\" to enable specific controllers, "+
		"and \"-gc-controller,-job-controller,-jobflow-controller,-jobtemplate-controller,-pg-controller,-queue-controller\" to disable specific controllers.")

````

File: cmd/controller-manager/app/server.go [hunk 10]
```diff
@@ -128,8 +129,15 @@ func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx c
 	controllerOpt.WorkerThreadsForPG = opt.WorkerThreadsForPG
 	controllerOpt.WorkerThreadsForGC = opt.WorkerThreadsForGC
 
+	checkControllersToStart(opt.Controllers)
```
googs1025, line 132: We print all the startable controllers here. @Monokaix

hwdef:
Is there anything that needs improvement? I think this PR is perfect.

cmd/controller-manager/app/server.go
Monokaix: Is this necessary? because we have print not enabled controllers in line 137, I mean we just print known controllers in ths flag: ）, previous implement is almost ok and just need print the known controllers registered in the flag.

googs1025: ok, so i will remove this

googs1025:
usage describe be like: 
```bash
     ...
      --controllers strings                      Specify controller gates. Use '*' for all controllers, all knownController: [gc-controller job-controller jobflow-controller jobtemplate-controller pg-controller queue-controller] ,and we can use '-' to disable controllers, e.g. "-job-controller,-queue-controller" to disable job and queue controllers. (default [*])

```

hwdef (commented):
"+gc-controller, -gc-controller, *"

Have you considered this case?

I think a function should be added to check for incorrect settings, such as a controller being set `open` and `close` at the same time.

googs1025:
> "+gc-controller, -gc-controller, *"
> 
> Have you considered this case?
> 
> I think a function should be added to check for incorrect settings, such as a controller being set `open` and `close` at the same time.

good catch! thank @hwdef  i will fixed this case

cmd/controller-manager/app/options/options.go
```diff
@@ -107,13 +116,44 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
...
+	// check controllers option
+	if err := s.checkControllersOrDie(); err != nil {
+		return err
+	}
+	// check leader election flag when LeaderElection is enabled.
 	return componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
 }
 
+// checkControllersOrDie checks the controllers option and returns error if it's invalid
+func (s *ServerOption) checkControllersOrDie() error {
```
googs1025, line 134: @hwdef Verification is performed here. If there is `--controllers="+job-controller,job-controller,-job-controller"`, the verification will fail.

cmd/controller-manager/app/options/options.go
```diff
@@ -107,13 +116,50 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
...
+	// check controllers option
+	if err := s.checkControllers(); err != nil {
+		return err
+	}
+	// check leader election flag when LeaderElection is enabled.
 	return componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
 }
 
+// checkControllers checks the controllers option and returns error if it's invalid
+func (s *ServerOption) checkControllers() error {
```
hwdef, line 134: Please write some UT for this function

cmd/controller-manager/app/options/options.go
```diff
@@ -107,13 +116,50 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
...
+	existenceMap := make(map[string]bool)
+	for _, c := range s.Controllers {
+		if c == "*" {
+			// wildcard '*' is not allowed to be combined with other input
+			if len(s.Controllers) > 1 {
+				return fmt.Errorf("wildcard '*' cannot be combined with other input")
+			}
+		} else {
+			if strings.HasPrefix(c, "-") && strings.HasPrefix(c, "+") {
```
hwdef, line 143: This if will never be true

googs1025: deleted

cmd/controller-manager/app/options/options.go
googs1025: done

cmd/controller-manager/app/options/options.go
```diff
@@ -107,13 +116,46 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
...
+	// check controllers option
+	if err := s.checkControllers(); err != nil {
+		return err
+	}
+	// check leader election flag when LeaderElection is enabled.
 	return componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
 }
 
+// checkControllers checks the controllers option and returns error if it's invalid
+func (s *ServerOption) checkControllers() error {
```
Monokaix, line 134: IMO, it's a little complicated and not so readable here, the controller in front has higher priority is just simple and acceptable.
And if this check is truly needed, I think use two maps to check is better.

googs1025: This check only verifies whether the same controller exists at the same time. For example, if "+job-controller" and "-job-controller" exist at the same time. No priority is considered.

hwdef (commented):
/lgtm

hwdef (commented):
/lgtm

william-wang (approved):
/approve

openxxs:
I want to disable job-controller.
if I set --controllers=*,-job-controller, it causes Error;
then I set --controllers=-job-controller, it disabled all controllers.
I have to set --controllers=+queue-controller,+pg-controller,+gc-controller,+hyperNode-controller+any other controllers.

I suggest volcano supports to disable specific controllers by --controllers=*,-job-controller.

Monokaix:
> I want to disable job-controller. if I set --controllers=*,-job-controller, it causes Error; then I set --controllers=-job-controller, it disabled all controllers. I have to set --controllers=+queue-controller,+pg-controller,+gc-controller,+hyperNode-controller+any other controllers.
> 
> I suggest volcano supports to disable specific controllers by --controllers=*,-job-controller.

@googs1025 Can you take a look?

googs1025:
> > I want to disable job-controller. if I set --controllers=*,-job-controller, it causes Error; then I set --controllers=-job-controller, it disabled all controllers. I have to set --controllers=+queue-controller,+pg-controller,+gc-controller,+hyperNode-controller+any other controllers.
> > I suggest volcano supports to disable specific controllers by --controllers=*,-job-controller.
> 
> @googs1025 Can you take a look?

I'll take a look the weekend and if it works I'll fix it next week.

PR #3494, opened 2024-05-26
refactor vcctl

googs1025:
#### What type of PR is this?
/kind cleanup

- remove duplicate code and use the public code under the util package directly
- refactor `buildJobCmd` `buildQueueCmd` to make the code more readable and extensible
- use `cmd.Context()` instead of `context.TODO()` pass context

```release-note
None
```

```docs
None
```


pkg/cli/job/suspend.go
```diff
@@ -36,14 +37,14 @@ var suspendJobFlags = &suspendFlags{}
...
-	initFlags(cmd, &suspendJobFlags.commonFlags)
+	util.InitFlags(cmd, &suspendJobFlags.CommonFlags)
 
 	cmd.Flags().StringVarP(&suspendJobFlags.Namespace, "namespace", "n", "default", "the namespace of job")
 	cmd.Flags().StringVarP(&suspendJobFlags.JobName, "name", "N", "", "the name of job")
 }
 
 // SuspendJob suspends the job.
-func SuspendJob() error {
+func SuspendJob(ctx context.Context) error {
```
Monokaix, line 47: Thanks for your contribution！
 A little doubt why this context is added?

googs1025: `cmd *cobra.Command` 
Provided the context ctx, we should use the one provided by cmd instead of using context.TODO(). We should pass the upstream ctx to the downstream.

```go
rootCmd := cobra.Command{
		Use:   "vjobs",
		Short: "view job information",
		Long:  `view information of a job with specified name or jobs from the same namespace`,
		Run: func(cmd *cobra.Command, args []string) {
			util.CheckError(cmd, vjobs.ViewJob(cmd.Context()))
		},
	}

```

cmd/cli/vcancel/main.go
```diff
@@ -31,7 +31,7 @@ func main() {
 		Short: "cancel a job",
 		Long:  `cancel a running, pending, or aborted job with specified name in default or specified namespace`,
 		Run: func(cmd *cobra.Command, args []string) {
-			util.CheckError(cmd, vcancel.CancelJob())
+			util.CheckError(cmd, vcancel.CancelJob(cmd.Context()))
```
Monokaix, line 34: Will `cmd.Context()` be nil?

googs1025: It will not be `nil`, it will be initialized by default when main.go is started.
The call chain is as follows

```go
func main() {
	rootCmd := cobra.Command{
		Use: "vcctl",
	}

	// tell Cobra not to provide the default completion command
	rootCmd.CompletionOptions.DisableDefaultCmd = true

	rootCmd.AddCommand(buildJobCmd())
	rootCmd.AddCommand(buildQueueCmd())
	rootCmd.AddCommand(versionCommand())

	code := cli.Run(&rootCmd)
	os.Exit(code)
}

func Run(cmd *cobra.Command) int {
	if logsInitialized, err := run(cmd); err != nil {
       ....
}

func run(cmd *cobra.Command) (logsInitialized bool, err error) {
	rand.Seed(time.Now().UnixNano())
	defer logs.FlushLogs()

	...

	err = cmd.Execute()
	return
}

// Execute uses the args (os.Args[1:] by default)
// and run through the command tree finding appropriate matches
// for commands and then corresponding flags.
func (c *Command) Execute() error {
	_, err := c.ExecuteC()
	return err
}

// ExecuteC executes the command.
func (c *Command) ExecuteC() (cmd *Command, err error) {
	if c.ctx == nil {
		c.ctx = context.Background()
	}

	// Regardless of what command execute is called on, run on Root only
	if c.HasParent() {
		return c.Root().ExecuteC()
	}

	// windows hook
	if preExecHookFn != nil {
		preExecHookFn(c)
	}
      .....
```

Monokaix: Ok

Issue #3491, opened 2024-05-24
record the norminated node(which is pipelined node in volcano) for pod and consider it first in next cycle


lowang-bh:
#### What would you like to be added:

1.  record the pipelined node for a task which has evicted other task and suitable on that node
2. check this node is suitable for this task in next scheduling cycle firstly, if predicating returns success, just assign this task to it

#### Why is this needed:

In the preempt/reclaim actions, task that can not be allocated in allocate action will try to evict other task and then pipeline to the proper node. And we can record it so that we can consider it firstly in the next scheduling cycle, to save time spent on predicating.

Monokaix:
So how to do that? just reduce availabel node nums to predict the nominated node?

JesseStutler:
@muskan2622 Hi, do you still work on this? Could you assign to me? I'm interested to figure out it.

lowang-bh:
> @muskan2622 Hi, do you still work on this? Could you assign to me? I'm interested to figure out it.

I think you can go on.

muskan2622:
Hey m working on it. Please consider some other issues

JesseStutler:
@muskan2622 Hi, you still work on this? But why did you remove the assignment?

muskan2622:
I thought it is merged.

JesseStutler:
still need it, `check this node is suitable for this task in next scheduling cycle firstly, if predicating returns success, just assign this task to it` has not been implemented yet

/assign

mahdikhashan:
@JesseStutler thanks for helping me find this issue - I'll open a pr and keep you updated.

/assign

zhoushuke:
but in reclaim action, it uses ssn.Pipeline, and  not set nominatedNode in it.
I found a issue: https://github.com/volcano-sh/volcano/issues/3673 to change recliam to using stmt.Pipeline instead ssn.Pipeline.

@lowang-bh @bibibox

PR #3498, opened 2024-05-27
Update NominatedNodeName for pipelined task

bibibox:
To fix issue #3499

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -641,7 +641,7 @@ func (ji *JobInfo) FitError() string {
 	// Stat histogram for pending tasks only
 	reasons = make(map[string]int)
 	for uid := range ji.TaskStatusIndex[Pending] {
-		reason, _ := ji.TaskSchedulingReason(uid)
+		reason, _, _ := ji.TaskSchedulingReason(uid)
 		reasons[reason]++
 	}
 	if len(reasons) > 0 {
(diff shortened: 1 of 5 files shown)


lowang-bh:
This pr can be part of solution for feature [record the norminated node for pod and consider it first in next cycle #3491](https://github.com/volcano-sh/volcano/issues/3491)

bibibox:
> This pr can be part of solution for feature [record the norminated node for pod and consider it first in next cycle #3491](https://github.com/volcano-sh/volcano/issues/3491)

yes, there may be some overlap, and this pr is mainly to solve this issue #3499

pkg/scheduler/cache/cache.go
```diff
@@ -992,14 +992,26 @@ func (sc *SchedulerCache) taskUnschedulable(task *schedulingapi.TaskInfo, reason
...
+	updateNomiNode := podNominatedNodeNameNeedUpdate(&pod.Status, nominatedNodeName)
+
+	if updateCond || updateNomiNode {
 		pod = pod.DeepCopy()
 
+		if updateCond && podutil.UpdatePodCondition(&pod.Status, condition) {
+			klog.V(3).Infof("Updating pod condition for %s/%s to (%s==%s)", pod.Namespace, pod.Name, condition.Type, condition.Status)
+		}
+
+		if updateNomiNode {
```
Monokaix, line 1005: Better add a comment here to explain why this is added？

pkg/scheduler/api/job_info.go
```diff
@@ -669,19 +669,19 @@ func (ji *JobInfo) TaskSchedulingReason(tid TaskID) (reason string, msg string)
 	case Allocated:
 		// Pod is schedulable
 		msg = fmt.Sprintf("Pod %s/%s can possibly be assigned to %s", taskInfo.Namespace, taskInfo.Name, ctx.NodeName)
-		return PodReasonSchedulable, msg
+		return PodReasonSchedulable, msg, ""
```
Monokaix, line 672: `Allocated` is also a unschedulable case, we should also check whether nominatedNode is needed, maybe we can refer to kube -scheduler to deep dive.

bibibox: the autoscaler will check both condition status and reason whether a pod is unschedulable or not:
image
so allocated will not be thought as unschedulable case to autoscaler

pkg/scheduler/cache/cache.go
bibibox: done

pkg/scheduler/api/job_info.go
Monokaix: Ok

william-wang (approved):
/approve

PR #3500, opened 2024-05-28
feat: add jobflow name for vcjob label and annotation when creating

googs1025:
#### What type of PR is this?
/kind feature

- add jobflow name for vcjob label and annotation when creating 
When jobflow creates the sub-resource vcjob, we should add the jobflow name to the label and annotation. Although there is already an `OwnerReference` field, the label and annotation will allow users to more intuitively know who created the upstream object of this vcjob.

```release-note
None
```

```docs
None
```

### DIFF
--- a/pkg/controllers/jobflow/jobflow_controller_action.go
+++ b/pkg/controllers/jobflow/jobflow_controller_action.go
@@ -260,10 +260,16 @@ func (jf *jobflowcontroller) loadJobTemplateAndSetJob(jobFlow *v1alpha1flow.JobF
 
 	*job = v1alpha1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        jobName,
-			Namespace:   jobFlow.Namespace,
-			Labels:      map[string]string{CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName)},
-			Annotations: map[string]string{CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName)},
+			Name:      jobName,
+			Namespace: jobFlow.Namespace,
+			Labels: map[string]string{
+				CreatedByJobTemplate: GenerateObjectString(jobFlow.Namespace, flowName),
+				CreatedByJobFlow:     GenerateObjectString(jobFlow.Namespace, jobFlow.Name),
+			},
+			Annotations: map[string]string{
+				CreatedByJobTemplate: GenerateObjectString(jobFlow.Namespace, flowName),
+				CreatedByJobFlow:     GenerateObjectString(jobFlow.Namespace, jobFlow.Name),
+			},
 		},
 		Spec:   jobTemplate.Spec,
 		Status: v1alpha1.JobStatus{},
@@ -292,7 +298,7 @@ func (jf *jobflowcontroller) deleteAllJobsCreatedByJobFlow(jobFlow *v1alpha1flow
 func (jf *jobflowcontroller) getAllJobsCreatedByJobFlow(jobFlow *v1alpha1flow.JobFlow) ([]*v1alpha1.Job, error) {
 	var flowNames []string
 	for _, flow := range jobFlow.Spec.Flows {
-		flowNames = append(flowNames, GetTemplateString(jobFlow.Namespace, flow.Name))
+		flowNames = append(flowNames, GenerateObjectString(jobFlow.Namespace, flow.Name))
 	}
 	selector := labels.NewSelector()
 	r, err := labels.NewRequirement(CreatedByJobTemplate, selection.In, flowNames)
--- a/pkg/controllers/jobflow/jobflow_controller_util.go
+++ b/pkg/controllers/jobflow/jobflow_controller_util.go
@@ -29,8 +29,8 @@ func getJobName(jobFlowName string, jobTemplateName string) string {
 	return jobFlowName + "-" + jobTemplateName
 }
 
-// GetTemplateString get the JobTemplate information string
-func GetTemplateString(namespace, name string) string {
+// GenerateObjectString generates the object information string using namespace and name
+func GenerateObjectString(namespace, name string) string {
 	return namespace + "." + name
 }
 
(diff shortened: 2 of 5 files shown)


hwdef (commented):
/lgtm

pkg/controllers/jobflow/jobflow_controller_action.go
```diff
@@ -260,10 +260,16 @@ func (jf *jobflowcontroller) loadJobTemplateAndSetJob(jobFlow *v1alpha1flow.JobF
...
-			Annotations: map[string]string{CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName)},
+			Name:      jobName,
+			Namespace: jobFlow.Namespace,
+			Labels: map[string]string{
+				CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName),
+				CreatedByJobFlow:     jobFlow.Name,
+			},
+			Annotations: map[string]string{
+				CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName),
+				CreatedByJobFlow:     jobFlow.Name,
```
Monokaix, line 271: Thanks for your contribution! Is it neened to add both label and annotation?

googs1025: I have considered this issue. I refer to the fact that when creating vcjob, both annotation and label are added at the same time, so I think jobflow also needs to be added at the same time.
```bash
root@VM-0-17-ubuntu:/home/ubuntu# kubectl get pods test-a-default-nginx-0 -oyaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    scheduling.k8s.io/group-name: test-a-4e75ec64-feca-4489-aaef-f8a5bc1a2af9
    volcano.sh/job-name: test-a
    volcano.sh/job-version: "0"
    volcano.sh/queue-name: default
    volcano.sh/task-index: "0"
    volcano.sh/task-spec: default-nginx
    volcano.sh/template-uid: test-a-default-nginx
  creationTimestamp: "2024-05-27T03:07:34Z"
  labels:
    volcano.sh/job-name: test-a
    volcano.sh/job-namespace: default
    volcano.sh/queue-name: default
    volcano.sh/task-index: "0"
    volcano.sh/task-spec: default-nginx
  name: test-a-default-nginx-0
  namespace: default
  ownerReferences:
  - apiVersion: batch.volcano.sh/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Job
    name: test-a
    uid: 4e75ec64-feca-4489-aaef-f8a5bc1a2af9
  resourceVersion: "350123"
  uid: 9480b917-6c98-4232-a763-990a15f8e83a

```

hwdef: I think adding the job name to the label or annotation has no impact. We can be consistent with the behavior of the job, just like the Pod's label and annotation also have the name of the job.

googs1025: I have the same idea with @hwdef  cc @Monokaix

pkg/controllers/jobflow/jobflow_controller_action.go
```diff
@@ -260,10 +260,16 @@ func (jf *jobflowcontroller) loadJobTemplateAndSetJob(jobFlow *v1alpha1flow.JobF
...
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        jobName,
-			Namespace:   jobFlow.Namespace,
-			Labels:      map[string]string{CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName)},
-			Annotations: map[string]string{CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName)},
+			Name:      jobName,
+			Namespace: jobFlow.Namespace,
+			Labels: map[string]string{
+				CreatedByJobTemplate: GetTemplateString(jobFlow.Namespace, flowName),
+				CreatedByJobFlow:     jobFlow.Name,
```
Monokaix, line 267: namespace is also needed.

googs1025: I looked at the jobflow controller code and found that the jobs created by jobflow must be under its namespace, so I think namespace should not be needed. cc @hwdef

Monokaix: But `CreatedByJobTemplate` contains the namespace, we'd better keep consistent with it.

googs1025: done

pkg/controllers/jobflow/jobflow_controller_util.go
```diff
@@ -29,8 +29,8 @@ func getJobName(jobFlowName string, jobTemplateName string) string {
 	return jobFlowName + "-" + jobTemplateName
 }
 
-// GetTemplateString get the JobTemplate information string
-func GetTemplateString(namespace, name string) string {
+// GenerateObjectString generates the object information string using namespace and name
+func GenerateObjectString(namespace, name string) string {
```
hwdef, line 33: As I recall, in k8s, this function has a specific name `NamespacedName`, so you can name this func GetNamespacedName()

hwdef: https://github.com/kubernetes/kubernetes/blob/9d63e575f8482b857ef328325d2107df840b9deb/pkg/scheduler/framework/types.go#L959

But this func is `namespace/name`, different from your code `namespace.name`. Maybe you should change the code?

googs1025: I know there is such a method, but my idea is that this label has been around for a while, and I am not sure if there are any users who would use this label. So I think it is better to keep it as it is.

googs1025: I just changed the method name, the logic is the same as the original code

hwdef: ok that's no problem

hwdef (commented):
/lgtm

william-wang (approved):
/approve

Issue #3496, opened 2024-05-27
Helm: 1.8.2 -> 1.9.0 upgrade breaks


j-vizcaino:
**What happened**

* Volcano 1.8.2 installed using Helm (default values)
* Upgrade to Volcano 1.9.0 breaks due to new `volcano-admission-init` job attempting to change immutable fields

```
[...]

PR #3504, opened 2024-06-03
Add pre-install&pre-upgrade hook for admission-init job

Monokaix:
fix https://github.com/volcano-sh/volcano/issues/3496
1. Set a separate yaml file for volcano admission init job, which will be deleted after install successfully.
2. Set separate RABC rules for volcano admission init job, and volcano admission doesn't need secrets related permissions.

Also fix part of https://github.com/volcano-sh/volcano/issues/3505


william-wang (approved):
/approve

PR #3515, opened 2024-06-06
put back the queue to priority queue after job's resource allocating finished

lowang-bh:
Resolves https://github.com/volcano-sh/volcano/issues/3407

Put back the queue to priority queue after job's resource allocating finished to ensure that the priority of the queue is calculated based on the latest resource allocation situation, so that to keep fair-share between different queues.

Update PR #3413


lowang-bh:
@Monokaix @hwdef

william-wang (approved):
/approve

PR #3521, opened 2024-06-13
overcommit plugin AddForMinReq is no need to add other resources.

### LINKED ISSUE #3520: overcommit plugin is no need to check other resources.
**What happened**:

I created a vcjob to request 0.1 core CPU, but the podgroup is pending. I add some debug info logs to source code, and recompile the volcano-scheduler to replace the old one. I found the reason of pending is that the minReq's "nvidia.com/gpu" resouce is less than idle.
However, my vcjob didn't request the "nvidia.com/gpu" resouce. Finally I realized that the `jobMinReq` will add the inqueue's request resources and comparing with idle resources. Actually, the total amount of "nvidia.com/gpu" requested by inqueue is already more than idle.

https://github.com/volcano-sh/volcano/blob/c61742d2aee80e0fdc920f1d71ee7457a7e9e582/pkg/scheduler/plugins/overcommit/overcommit.go#L122

I think that when adding newly requested resources to inqueued resources, irrelevant resources should not be considered.  For example, if the minimum request is only 1 CPU core, then there is no need to add memory, GPU and other resources.

I guess it may be possible to reproduce it by doing the following:
1. Add and register GPU resources to one node of clusters.
2. Create a vcjob to request GPU resources.
3. When podgorup is running, remove the node from clusters.
4. The podgroup's status may become "Inqueue".
5. Then create a new vcjob to request only cpu or memory resource.
6. The new podgroup will be pending because of Inqueue's GPU not enough.

**Environment**:
- Volcano Version:
v1.9.0
v1.25.4
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

xieyanker:
Fix #3520

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -255,6 +255,33 @@ func (r *Resource) Add(rr *Resource) *Resource {
 	return r
 }
 
+// AddForMinRequest is used to add two given resources, but when rr's Resource struct == 0, r's Resource will be 0.
+// For example, if the minimum request is only 1 CPU core, then there is no need to add memory, GPU and other resources.
+func (r *Resource) AddForMinRequest(rr *Resource) *Resource {
+	if rr.MilliCPU == 0 {
+		r.MilliCPU = 0
+	} else {
+		r.MilliCPU += rr.MilliCPU
+	}
+
+	if rr.Memory == 0 {
+		r.Memory = 0
+	} else {
+		r.Memory += rr.Memory
+	}
+
+	newScalarResources := map[v1.ResourceName]float64{}
+	for rName, rQuant := range rr.ScalarResources {
+		if r.ScalarResources == nil {
+			r.ScalarResources = map[v1.ResourceName]float64{}
+		}
+		newScalarResources[rName] = r.ScalarResources[rName] + rQuant
+	}
+	r.ScalarResources = newScalarResources
+
+	return r
+}
+
 // Sub subtracts two Resource objects with assertion.
 func (r *Resource) Sub(rr *Resource) *Resource {
 	assert.Assertf(rr.LessEqual(r, Zero), "resource is not sufficient to do operation: <%v> sub <%v>", r, rr)
@@ -732,3 +759,4 @@ func (r ResourceNameList) Contains(rr ResourceNameList) bool {
 func IsCountQuota(name v1.ResourceName) bool {
 	return strings.HasPrefix(string(name), "count/")
 }
+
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/resource_info.go
```diff
@@ -255,6 +255,33 @@ func (r *Resource) Add(rr *Resource) *Resource {
...
+		r.MilliCPU = 0
+	} else {
+		r.MilliCPU += rr.MilliCPU
+	}
+
+	if rr.Memory == 0 {
+		r.Memory = 0
+	} else {
+		r.Memory += rr.Memory
+	}
```
googs1025, line 271: should we just use this
```go
r.MilliCPU += rr.MilliCPU
r.Memory += rr.Memory
```

pkg/scheduler/api/resource_info.go
```diff
@@ -255,6 +255,33 @@ func (r *Resource) Add(rr *Resource) *Resource {
...
+
+	newScalarResources := map[v1.ResourceName]float64{}
+	for rName, rQuant := range rr.ScalarResources {
+		if r.ScalarResources == nil {
+			r.ScalarResources = map[v1.ResourceName]float64{}
+		}
+		newScalarResources[rName] = r.ScalarResources[rName] + rQuant
+	}
+	r.ScalarResources = newScalarResources
+
```
googs1025, line 281: Is this code OK? It is relatively more readable
```go
          newScalarResources := map[v1.ResourceName]float64{}
	for rName, rQuant := range rr.ScalarResources {
		if r.ScalarResources == nil {
			r.ScalarResources = map[v1.ResourceName]float64{}
		}
		newScalarResources[rName] = r.ScalarResources[rName] + rQuant
	}
	r.ScalarResources = newScalarResources
```

pkg/scheduler/api/resource_info.go
```diff
@@ -255,6 +255,33 @@ func (r *Resource) Add(rr *Resource) *Resource {
...
+	}
+
+	newScalarResources := map[v1.ResourceName]float64{}
+	for rName, rQuant := range rr.ScalarResources {
+		if r.ScalarResources == nil {
+			r.ScalarResources = map[v1.ResourceName]float64{}
+		}
+		newScalarResources[rName] = r.ScalarResources[rName] + rQuant
+	}
+	r.ScalarResources = newScalarResources
```
googs1025, line 280: Is this code OK? It is relatively more readable
```go
if r.ScalarResources == nil {
    r.ScalarResources = map[v1.ResourceName]float64{}
}

for rName, rQuant := range rr.ScalarResources {
    r.ScalarResources[rName] += rQuant
}
```

pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -144,3 +144,4 @@ func (op *overcommitPlugin) OnSessionClose(ssn *framework.Session) {
 	op.idleResource = nil
 	op.inqueueResource = nil
 }
+
```
googs1025, line 147: Can this blank line be deleted?
In addition, can some test cases be added to prove this pr?

lowang-bh:
/hold 

change origin value is a bit strange.

PR #3522, opened 2024-06-13
resource compare support only consider the requested resource item

### LINKED ISSUE #3520: overcommit plugin is no need to check other resources.

lowang-bh:
fixes #3520

some times, we need to only compare the required dimensions in the resource.
eg: when check  `Enqueueable`, `Allocatable` or `Preemptive`, even overused.
Will use this function to replcae comparision in `AddJobEnqueueableFn`, `AddPreemptiveFn` and `AddAllocatableFn`

- [x] AddJobEnqueueableFn
- [ ] AddAllocatableFn
- [ ] AddPreemptiveFn
- [ ] AddOverusedFn

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -429,6 +429,42 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
 	return true
 }
 
+// LessEqualWithDimension only compare the resource items in req param
+// @param req define the resource item to be compared
+// if req is nil, equals r.LessEqual(rr, zero)
+func (r *Resource) LessEqualWithDimension(rr *Resource, req *Resource) bool {
+	if r == nil {
+		return true
+	}
+	if rr == nil {
+		return false
+	}
+	if req == nil {
+		return r.LessEqual(rr, Zero)
+	}
+
+	if req.MilliCPU > 0 && r.MilliCPU > rr.MilliCPU {
+		return false
+	}
+	if req.Memory > 0 && r.Memory > rr.Memory {
+		return false
+	}
+
+	// if r.scalar is nil, whatever rr.scalar is, r is less or equal to rr
+	if r.ScalarResources == nil {
+		return true
+	}
+
+	for name, quant := range req.ScalarResources {
+		rQuant := r.ScalarResources[name]
+		rrQuant := rr.ScalarResources[name]
+		if quant > 0 && rQuant > rrQuant {
+			return false
+		}
+	}
+	return true
+}
+
 // LessEqualWithResourcesName returns true, []string{} only on condition that all dimensions of resources in r are less than or equal with that of rr,
 // Otherwise returns false and err string ,which show what resources are insufficient.
 // @param defaultValue "default value for resource dimension not defined in ScalarResources. Its value can only be one of 'Zero' and 'Infinity'"
--- a/pkg/scheduler/api/resource_info_test.go
+++ b/pkg/scheduler/api/resource_info_test.go
@@ -753,6 +753,81 @@ func TestLessEqual(t *testing.T) {
 	}
 }
 
+func TestLessEqualWithDimension(t *testing.T) {
+	tests := []struct {
+		resource1 *Resource
+		resource2 *Resource
+		req       *Resource
+		expected  bool
+	}{
+		{
+			resource1: &Resource{},
+			resource2: &Resource{},
+			req:       nil,
+			expected:  true,
+		},
+		{
+			resource1: &Resource{MilliCPU: 5000},
+			resource2: &Resource{
+				MilliCPU:        4000,
+				Memory:          2000,
+				ScalarResources: map[v1.ResourceName]float64{"scalar.test/scalar1": 1000},
+			},
+			req:      &Resource{MilliCPU: 1000},
+			expected: false,
+		},
+		{
+			resource1: &Resource{MilliCPU: 3000, Memory: 3000},
+			resource2: &Resource{MilliCPU: 4000, Memory: 2000},
+			req:       &Resource{Memory: 1000},
+			expected:  false,
+		},
+		{
+			resource1: &Resource{
+				MilliCPU: 4,
+				Memory:   4000,
+			},
+			resource2: &Resource{},
+			req:       &Resource{},
+			expected:  true,
+		},
+		{
+			resource1: &Resource{
+				MilliCPU: 4, Memory: 4000,
+				ScalarResources: map[v1.ResourceName]float64{"nvidia.com/gpu": 1},
+			},
+			resource2: &Resource{MilliCPU: 8, Memory: 8000},
+			req: &Resource{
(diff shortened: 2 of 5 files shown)


pkg/scheduler/api/resource_info.go
```diff
@@ -429,6 +429,36 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
...
+	}
+
+	// if r.scalar is nil, whatever rr.scalar is, r is less or equal to rr
+	if r.ScalarResources == nil {
+		return true
+	}
+
+	for name, quant := range req.ScalarResources {
+		rQuant := r.ScalarResources[name]
+		rrQuant := rr.ScalarResources[name]
```
Monokaix, line 454: Shuould check rr not nil.

lowang-bh: Done.

pkg/scheduler/api/resource_info.go
```diff
@@ -429,6 +429,42 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
...
```
hwdef, line 460: Will rQuant or rrQuant be nil?

lowang-bh: They are float values.

hwdef: I mean not getting it

hwdef (commented):
/lgtm

pkg/scheduler/api/resource_info_test.go
```diff
@@ -753,6 +753,81 @@ func TestLessEqual(t *testing.T) {
...
+				MilliCPU: 4, Memory: 4000,
+				ScalarResources: map[v1.ResourceName]float64{"nvidia.com/gpu": 1},
+			},
+			resource2: &Resource{MilliCPU: 8, Memory: 8000},
+			req: &Resource{
+				MilliCPU: 4, Memory: 2000,
+				ScalarResources: map[v1.ResourceName]float64{"nvidia.com/gpu": 1},
+			},
+			expected: false,
+		},
```
Monokaix, line 805: Why this return false?

pkg/scheduler/api/resource_info_test.go
```diff
@@ -753,6 +753,81 @@ func TestLessEqual(t *testing.T) {
...
+			},
+			resource2: &Resource{
+				MilliCPU: 100, Memory: 8000,
+				ScalarResources: map[v1.ResourceName]float64{"nvidia.com/A100": 1},
+			},
+			req: &Resource{
+				MilliCPU: 10, Memory: 4000,
+				ScalarResources: map[v1.ResourceName]float64{"nvidia.com/gpu": 0, "nvidia.com/A100": 1, "scalar": 1},
+			},
+			expected: true,
```
Monokaix, line 819: Why true? resource2 has no dimension `"scalar": 1`.

pkg/scheduler/api/resource_info.go
```diff
@@ -429,6 +429,42 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
 	return true
 }
 
+// LessEqualWithDimension only compare the resource items in req param
```
Monokaix, line 432: I think we should clarify that this func behaves when one dimension of request missed in resource1 and when no dimension missed in resource1 but less than resource1.

Monokaix: seems that we only consider whether the dimension exists but not care the specific value of it, right?

pkg/scheduler/api/resource_info.go
```diff
@@ -429,6 +429,42 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
...
+// @param req define the resource item to be compared
+// if req is nil, equals r.LessEqual(rr, zero)
+func (r *Resource) LessEqualWithDimension(rr *Resource, req *Resource) bool {
```
Monokaix, line 435: How about just remove the dimensions not in `req` of `r`, then compare just like before?

pkg/scheduler/api/resource_info_test.go
lowang-bh: req need gpu resource with 1， and resource2 has no gpu, so it can not satisfy the request.

pkg/scheduler/api/resource_info.go
lowang-bh: also ok.

Monokaix: I think this way is more readable.

pkg/scheduler/api/resource_info_test.go
lowang-bh: It means resource1's `scalar` less equal to resource2's `scalar`.

pkg/scheduler/api/resource_info.go
lowang-bh: > seems that we only consider whether the dimension exists but not care the specific value of it, right?

yes

Monokaix:
`Overused` should also be considered.

pkg/scheduler/api/resource_info_test.go
```diff
@@ -753,6 +753,81 @@ func TestLessEqual(t *testing.T) {
...
+		},
+	}
+
+	for i, test := range tests {
```
Monokaix, line 823: Please also add a case:
r1'cpu resource > r2'cpu resource, while req has no cpu resource but has a enough scalar resource, it's expected to return true.

hwdef (commented):
/lgtm

hwdef:
because lgtm has been added by @Monokaix

PR #3531, opened 2024-06-16
use switch case statement to replace if else in plugins sla

yxxhero:
make code more readable.

### DIFF
--- a/pkg/scheduler/plugins/sla/sla.go
+++ b/pkg/scheduler/plugins/sla/sla.go
@@ -124,12 +124,15 @@ func (sp *slaPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		lCreationTimestamp := lv.CreationTimestamp
 		rCreationTimestamp := rv.CreationTimestamp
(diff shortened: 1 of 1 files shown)


Monokaix:
Hi, please squash it to one commit and rename the title to a more readable one：`use switch case statement to replace if else in plugins sla`.

yxxhero:
@Monokaix done

pkg/scheduler/plugins/sla/sla.go
```diff
@@ -124,12 +124,15 @@ func (sp *slaPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		lCreationTimestamp := lv.CreationTimestamp
 		rCreationTimestamp := rv.CreationTimestamp
-		if lCreationTimestamp.Add(*lJobWaitingTime).Before(rCreationTimestamp.Add(*rJobWaitingTime)) {
+
```
Monokaix, line 127: I think we can use time's `Compare` method to replace the statement, which is more concise.

yxxhero: please show more info. thanks so much. @Monokaix

Monokaix: We can call the method `time.Compare` to replace if else check.

PR #3538, opened 2024-06-22
improve unschedule message

### LINKED ISSUE #2993: Optimize log output when resources are not enough

lowang-bh:
This PR is aim to add useful message in podgroup status so that user is easy to find the reason why job is unschedulable.

it also fixes #2993

<details>

<summary>example YAML </summary>

### server-0 allocate success, but server-1 allocate failed

```yaml
kind: Job
metadata:
  name: my-apps
spec:
  schedulerName: volcano
  minAvailable: 2
  tasks:
    - replicas: 2
      name: "web"
      minAvailable: 1
      template:
        spec:
          containers:
            - image: nginx:1.14.2
              name: web
              resources:
                requests:
                  cpu: "1000m"
                  memory: "100Mi"
          restartPolicy: OnFailure
    - replicas: 2
      minAvailable: 1
      name: "server"
      template:
        spec:
          containers:
            - image: nginx:1.14.2
              name: server
              resources:
                requests:
                  cpu: "2000m"
                  memory: "100Mi"
          restartPolicy: OnFailure
```
</details>

#### podgroup message
image

<details>

<summary>pod message</summary>

image

</details>

### DIFF
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -179,7 +179,6 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 			unreadyTaskCount = job.MinAvailable - schedulableTaskNum()
 			msg := fmt.Sprintf("%v/%v tasks in gang unschedulable: %v",
 				unreadyTaskCount, len(job.Tasks), job.FitError())
-			job.JobFitErrors = msg
 
 			unScheduleJobCount++
 			metrics.RegisterJobRetries(job.Name)
@@ -196,18 +195,6 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 			if err := ssn.UpdatePodGroupCondition(job, jc); err != nil {
 				klog.Errorf("Failed to update job <%s/%s> condition: %v",
 					job.Namespace, job.Name, err)
-			}
-
-			// allocated task should follow the job fit error
-			for _, taskInfo := range job.TaskStatusIndex[api.Allocated] {
-				fitError := job.NodesFitErrors[taskInfo.UID]
-				if fitError != nil {
-					continue
-				}
-
-				fitError = api.NewFitErrors()
-				job.NodesFitErrors[taskInfo.UID] = fitError
-				fitError.SetError(msg)
 			}
 		} else {
 			jc := &scheduling.PodGroupCondition{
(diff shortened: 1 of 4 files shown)


hwdef (commented):
/lgtm

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -197,18 +196,6 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 				klog.Errorf("Failed to update job <%s/%s> condition: %v",
 					job.Namespace, job.Name, err)
 			}
-
-			// allocated task should follow the job fit error
```
Monokaix, line 201: Why this removed?

lowang-bh: It is to set those tasks have no fiterror, which have been alocated. It is meaningless. This message can be found in podgroup message and event. After removing it , those pods will have a message as the pictures show that `it can be assigned at node xxx`.

Monokaix:
The problem of original mentioned issus still exists or not? podgroup is a volcano resource and there is no necessary to be consistent with kube-scheduler, we can just keep pod event being with kube-scheduler?
 
If pg also has these message, it will be repeated with the pod event.

lowang-bh:
> The problem of original mentioned issus still exists or not? podgroup is a volcano resource and there is no necessary to be consistent with kube-scheduler, we can just keep pod event being with kube-scheduler?
> 
> If pg also has these message, it will be repeated with the pod event.


It is aim to add useful message in podgroup status so that user is easy to find the reason why job is unschedulable
The origin reason come from the first unscheduled pod which break the allocation.

Monokaix:
> > The problem of original mentioned issus still exists or not? podgroup is a volcano resource and there is no necessary to be consistent with kube-scheduler, we can just keep pod event being with kube-scheduler?
> > If pg also has these message, it will be repeated with the pod event.
> 
> It is aim to add useful message in podgroup status so that user is easy to find the reason why job is unschedulable The origin reason come from the first unscheduled pod which break the allocation.

Ok.

PR #3540, opened 2024-06-23
reduce function calling

lowang-bh:
/kind cleanup

### DIFF
--- a/pkg/controllers/job/job_controller_util.go
+++ b/pkg/controllers/job/job_controller_util.go
@@ -336,8 +336,9 @@ func (p TasksPriority) CalcPGMinResources(jobMinAvailable int32) v1.ResourceList
 // calTaskRequests returns requests resource with validReplica replicas
 func calTaskRequests(pod *v1.Pod, validReplica int32) v1.ResourceList {
 	minReq := v1.ResourceList{}
+	usage := *util.GetPodQuotaUsage(pod)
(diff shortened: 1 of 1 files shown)


pkg/controllers/job/job_controller_util.go
```diff
@@ -336,8 +336,9 @@ func (p TasksPriority) CalcPGMinResources(jobMinAvailable int32) v1.ResourceList
 // calTaskRequests returns requests resource with validReplica replicas
 func calTaskRequests(pod *v1.Pod, validReplica int32) v1.ResourceList {
 	minReq := v1.ResourceList{}
+	usage := *util.GetPodQuotaUsage(pod)
 	for i := int32(0); i < validReplica; i++ {
```
Monokaix, line 340: If `validReplica` length=0, then `util.GetPodQuotaUsage(pod)` will not be called, but after this pr, it will always be called, is this feasible?

lowang-bh: `validReplica` will nearly never be zero.  So most cases `GetPodQuotaUsage` will only be called once. Even though it is zero, the logic is ok and has no other impact.

Monokaix: OK.

hwdef (commented):
/approve

Issue #748, opened 2020-03-26
Feature Request: Need a mechanism to avoid starvation (obey job priority)


xiaogaozi:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:

In a share cluster may mixed with different type jobs, e.g. high priority job, low priority job. User always expect high priority job be scheduled first. But in some case when there are insufficient resources could be allocated to high priority job, Volcano scheduler will ignore this job and try to schedule low priority job. In the worst case, high priority job will keep waiting until low priority job release sufficient resources (i.e. starvation). This is not acceptable for user.

**What you expected to happen**:

High priority job could be scheduled before low priority job. Scheduler should try the best to avoid starvation.

**How to reproduce it (as minimally and precisely as possible)**:

Mix high and low priority jobs in a cluster

**Anything else we need to know?**:

Some previous related discussion and design from kube-batch and Volcano community:

- https://github.com/kubernetes-sigs/kube-batch/pull/686
- https://github.com/kubernetes-sigs/kube-batch/pull/847
- https://github.com/volcano-sh/volcano/issues/13
- https://github.com/volcano-sh/volcano/pull/298

**Environment**:
- Volcano Version: v0.3
- Kubernetes version (use `kubectl version`): v1.14.6
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

xiaogaozi:
Should be noted that `preempt` action may resolve this problem in some case. But we may also need a new mechanism (e.g. reservation) when preemption is not acceptable.

Gouthamkreddy1234:
Reservations will be a part future releases of v1.x (v1.0 was released in June 30 which currently does not have reservations yet). Please check this roadmap for more info - [https://github.com/volcano-sh/volcano/blob/c388b818c4dada8d63a76dbc7cf53db75bf7e26a/docs/community/roadmap.md](https://github.com/volcano-sh/volcano/blob/c388b818c4dada8d63a76dbc7cf53db75bf7e26a/docs/community/roadmap.md)

Gouthamkreddy1234:
This feature will be a part of v1.1 and is being discussed here - [https://github.com/volcano-sh/volcano/issues/947](https://github.com/volcano-sh/volcano/issues/947)

Thor-wl:
[...]

PR #3546, opened 2024-06-27
delete reservation plugin doc

hwdef:
Reservation plugin has been deleted. This doc is no longer needed. It will confuse users.
Additionally, I recommend cleaning up documents that have expired.

### DIFF
--- a/docs/design/job-resource-reservation-design.md
+++ b/docs/design/job-resource-reservation-design.md
@@ -2,6 +2,10 @@
 
 @[Thor-wl](https://github.com/Thor-wl); Aug 19th, 2020
 
+## Deprecated
+
+The `Reservation` plugin has been deleted. This doc is deprecated. It will be deleted in the future.
+
 ## Motivation
 As [issue 13](https://github.com/volcano-sh/volcano/issues/13) / [issue 748](https://github.com/volcano-sh/volcano/issues/748) 
 / [issue 947](https://github.com/volcano-sh/volcano/issues/947) mentioned, current scheduler strategy may result in 


Monokaix:
have the problems mentioned in this doc been resolved?
And I saw that https://github.com/volcano-sh/volcano/issues/748 mentioned that this feature is released, but I didn't see that.

hwdef:
I think the reservation plugin has been deprecated and replaced by the SLA plugin.
@Thor-wl Can you clarify?

docs/design/job-resource-reservation-design.md
```diff
@@ -1,148 +0,0 @@
-# Volcano Resource Reservation For Target Jobs
```
googs1025, line 1: Maybe we should have a deprecated directory and put this design document in it instead of deleting it directly. This way users can also refer to the previous documents. @Monokaix WDYT

hwdef: > Maybe we should have a deprecated directory and put this design document in it instead of deleting it directly. This way users can also refer to the previous documents. @Monokaix WDYT

If users want to view old documents, they can switch to the old branch

Monokaix: We can mark it as deprecated and delete it in next version: )

hwdef: @Monokaix 
how to mark it as deprecated? write in title? create a deprecated folder?

Monokaix: I think we can just edit the doc directly: ), and whether it's sovled by sla-plugin should also be checked.

hwdef:
@Monokaix @googs1025 
The docs has been updated. Please review again.

PR #3547, opened 2024-06-27
update pod status when bind error

bibibox:
The volcano wont update pod status when bind error, but just add the failed pod to errorTask channel.
If one pod always fails when binding to a node, the pod status will be just pending, and we cannot get any useful information from its status.
This PR fix this situation, we can get SchedulerError when failing to bind pod(same as k8s default scheduler)


lowang-bh (commented):
/lgtm

william-wang (approved):

PR #3551, opened 2024-06-27
Remove init variable in Reclaimable and Preemptable parts of session_plugins

PigNatovsky:
Hello, 

it's a really small - so called - refactor. I've just changed values of init variable which shows if this is the first iteration of Reclaimable/Preemptable. In my opinion this if:
```
if init {
    victims = candidates
    init = false
}
```

is more readable than:
```
if !init {
    victims = candidates
    init = true
}
```
It's like a double negation.

### DIFF
--- a/pkg/scheduler/framework/session_plugins.go
+++ b/pkg/scheduler/framework/session_plugins.go
@@ -153,7 +153,6 @@ func (ssn *Session) AddJobStarvingFns(name string, fn api.ValidateFn) {
 // Reclaimable invoke reclaimable function of the plugins
 func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskInfo) []*api.TaskInfo {
 	var victims []*api.TaskInfo
-	var init bool
 
 	for _, tier := range ssn.Tiers {
 		for _, plugin := range tier.Plugins {
@@ -173,9 +172,9 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 				victims = nil
 				break
 			}
-			if !init {
+			// first iteration - initialize victims list
+			if victims == nil {
 				victims = candidates
-				init = true
 			} else {
 				var intersection []*api.TaskInfo
 				// Get intersection of victims and candidates.
@@ -203,7 +202,6 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 // Preemptable invoke preemptable function of the plugins
 func (ssn *Session) Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) []*api.TaskInfo {
 	var victims []*api.TaskInfo
-	var init bool
 
 	for _, tier := range ssn.Tiers {
 		for _, plugin := range tier.Plugins {
@@ -224,10 +222,9 @@ func (ssn *Session) Preemptable(preemptor *api.TaskInfo, preemptees []*api.TaskI
 				victims = nil
 				break
 			}
-
-			if !init {
+			// first iteration - initialize victims list
+			if victims == nil {
 				victims = candidates
-				init = true
 			} else {
 				var intersection []*api.TaskInfo
 				// Get intersection of victims and candidates.


pkg/scheduler/framework/session_plugins.go
```diff
@@ -153,7 +153,7 @@ func (ssn *Session) AddJobStarvingFns(name string, fn api.ValidateFn) {
 // Reclaimable invoke reclaimable function of the plugins
 func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskInfo) []*api.TaskInfo {
 	var victims []*api.TaskInfo
-	var init bool
+	var init = true
```
googs1025, line 156: My question is: what does the init variable represent? Why is it better to change it to true?

googs1025: This variable should represent whether it is initialized or not. I think the default value should be false. However, can the variable name be changed? It may cause misunderstanding.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -173,9 +173,9 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 				victims = nil
 				break
 			}
-			if !init {
+			if init {
 				victims = candidates
-				init = true
+				init = false
```
googs1025, line 178: My another question is: can we not need this init variable? If it is triggered for the first time, len(victims) should be nil, can we use this directly to judge?
```go
                               if abstain == 0 {
				continue
			}
			if len(candidates) == 0 {
				victims = nil
				break
			}
			if victims == nil {
				victims = candidates
			} else {
				var intersection []*api.TaskInfo
				// Get intersection of victims and candidates.
				for _, v := range victims {
					for _, c := range candidates {
						if v.UID == c.UID {
							intersection = append(intersection, v)
						}
					}
				}

				// Update victims to intersection
				victims = intersection
			}

```

lowang-bh:
How about use `inited` instead of `init`?

pkg/scheduler/framework/session_plugins.go
PigNatovsky: I was thinking in that direction. But for me init asks if we need to initialize the list of victims/if it's a first loop. So we can also name it first_iteration, for example. 
It's like thinking -  we want to check if it's not initialized, or we want to check if we need to do init assigment.

pkg/scheduler/framework/session_plugins.go
PigNatovsky: This approach is even better - but proper comment can me useful for a reader.

PigNatovsky:
@lowang-bh @googs1025 
I've pushed new commit with changes suggested by @googs1025 - as a reader I can say that it's pretty readable and I can understand a code flow.

PigNatovsky:
Hello Guys,

Could You comment? If changes are too small and unimportant, just let me know, I will close this.

And thank You for a previous reaponses!

googs1025:
/assign @Monokaix  
hi @Monokaix Maybe you can take a look and suggest an edit!

Monokaix:
> /assign @Monokaix hi @Monokaix Maybe you can take a look and suggest an edit!

I think it's ok: )

PigNatovsky:
Thank You @Monokaix . And sorry for stupid question, but what should I do next?

googs1025:
> Thank You @Monokaix . And sorry for stupid question, but what should I do next?

Just wait for other approver to be reviewed and merged. 👍

PigNatovsky:
@hudson741 @k82cn @googs1025 
Is there a chance to get review/something with this small change, or should I abandon it and close?

Monokaix:
Please rebase and squash to one commit: )

PigNatovsky:
Rebased and squashed, ready to go :)

Monokaix:
> /assign @Monokaix

You can just keep your own commit and remove `Merge Branch xxx` to kepp a clean commit: )

PigNatovsky:
Sorry, I was using a button in the PR UI to merge master branch, it's not the best idea. But now branch is rebased from master, my commit is on the tip.

PR #3552, opened 2024-06-27
add docs about Network Topology Aware Plugin

lowang-bh:
/kind feature

docs for PR #3388

## Motivation
We target to make scheduler net-topology aware so as to achieve the following:

- best effort to schedule same job to same topology devices, such as same idc, switch.

## Goals

- Support single key topology configuration, try to schedule job's all tasks to nodes which have same value with that key
- Support multiple-key topology policies, the key at front get higher score

## Non-Goals

- Not to find the global solutions among nodes with all kind values of that key

### DIFF
--- a/docs/design/network-topology-aware.md
+++ b/docs/design/network-topology-aware.md
@@ -1,0 +1,92 @@
+# Network Topology Aware Plugin
+
+- [Network Topology Aware Plugin](#network-topology-aware-plugin)
+  - [Backgrounds](#backgrounds)
+  - [Motivation](#motivation)
+  - [Proposal one](#proposal-one)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+    - [Design Action](#design-action)
+      - [Pod scheduling process](#pod-scheduling-process)
+      - [Usage](#usage)
+    - [Drawbacks](#drawbacks)
+
+## Backgrounds
+
+A Kubernetes cluster typically comprises numerous nodes distributed across different IDCs, chassis, and switches.
+
+Data transformations vary in performance across these different components.
+
+For latency-sensitive workloads, it's crucial to execute tasks within the same IDC and ideally on the same chassis and switch.
+
+## Motivation
+
+The goal is to make the Kubernetes scheduler network-topology aware to achieve the following:
+
+Ensure optimal scheduling of tasks from the same job onto nodes within the same topology, such as the same IDC, chassis, or switch.
+
+There will be two types of network-topology aware
+
+- **static**: `network-topology.type: static` is aiming to aware the network topology by nodes' labels
+  - it is suitable for small scale cluster, and it don't need to import extra components
+- **dynamic**: `network-topology.type: dynamic` is aiming to use some tools to detect the network topology dynamically. For example, `ibnetdiscover` can be used to discover the InfiniBand network topology
+  - for large scale cluster it is recommended to use this one to save the cluster administrator's work
+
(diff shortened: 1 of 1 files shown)


docs/design/net-topology-aware.md
```diff
@@ -0,0 +1,85 @@
...
+#### Usage
+
+1. Label nodes with key-value pairs (e.g., `switch=NvLink-A100`, `rack=rack1,rack2`, `idc=bj,sh`) to partition nodes into different topology zones
+2. Add the `net-topology` plugin in the scheduler configuration to implement these policies.
+
+```yaml
+- plugins:
+   - name: net-topology
+     arguments:
+       net-topology.type: static
```
Monokaix, line 78: Please describe the meaning of static type, and can also describe future possible of other types.

lowang-bh: Updated

docs/design/net-topology-aware.md
```diff
@@ -0,0 +1,90 @@
+# Net Topology Aware Plugin
+
+- [Net Topology Aware Plugin](#net-topology-aware-plugin)
```
googs1025, line 3: I have a small question: why is this called network topology and not node topology?

googs1025: I see in the design document that all the implementations are related to node labels, and there seems to be no information related to the network.

googs1025: Is it different from the nodegroup plugin?

docs/design/net-topology-aware.md
```diff
@@ -0,0 +1,90 @@
...
+## Motivation
+
+The goal is to make the Kubernetes scheduler network-topology aware to achieve the following:
+
+Ensure optimal scheduling of tasks from the same job onto nodes within the same topology, such as the same IDC, chassis, or switch.
+
+There will be two types of network-topology aware
+
+- **static**: `net-topology.type: static` is aiming to aware the network topology by nodes' labels
+- **dynamic**: `net-topology.type: dynamic` is aiming to use some tools to detect the network topology dynamically
```
william-wang, line 31: Is there any detail for network topology discovery?

docs/design/net-topology-aware.md
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+    return float64(score * weight), nil
+}
+```
+
+#### Usage
+
+1. Label nodes with key-value pairs (e.g., `switch=NvLink-A100`, `rack=rack1,rack2`, `idc=bj,sh`) to partition nodes into different topology zones
+2. Add the `net-topology` plugin in the scheduler configuration to implement these policies.
```
william-wang, line 77: suggest using network-toplogy or network-toplogy-aware :)

lowang-bh: Done

docs/design/net-topology-aware.md
lowang-bh: For example, `ibnetdiscover` can be used to discover the InfiniBand network topology.
I have add an example tool in docs.

lowang-bh:
close it as there is a more ideal design in https://github.com/volcano-sh/volcano/pull/3850

JesseStutler:
Hi all! 👋

Network Topology Aware Scheduling is now available in preview! Check out the docs: https://volcano.sh/en/docs/network_topology_aware_scheduling/

Feel free to try it out and let us know what you think in slack, wechat or github issues 🚀

PR #3557, opened 2024-07-02
Remove invalid parameter IgnoredNamespaces

xieyanker:
When I was looking up how to ignore the webhook of a specified namespace in volcano, I saw that there is `ignored-namespaces` paramerer in the code, but it is actually invalid, so it should be removed to avoid misleading.

### DIFF
--- a/cmd/webhook-manager/app/options/options.go
+++ b/cmd/webhook-manager/app/options/options.go
@@ -26,12 +26,11 @@ import (
 )
 
 const (
-	defaultSchedulerName     = "volcano"
-	defaultQPS               = 50.0
-	defaultBurst             = 100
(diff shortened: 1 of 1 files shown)


hwdef (commented):
Yes, you are right. 
/lgtm

lowang-bh:
How about implementing this feature?

xieyanker:
> How about implementing this feature?

As far as I know, the pr https://github.com/volcano-sh/volcano/pull/2346 has removed the `ignored-namespaces` function, but the parameter was forgotten to be deleted.

shinytang6 (commented):
/lgtm

xieyanker:
/assign @hwdef
Can this pr be approved?

william-wang (approved):
/approve

xieyanker:
@hwdef  The `Verify codes` test failed, I run `gofmt -s -w  cmd/webhook-manager/app/options/options.go` fo fix it and resubmit.

xieyanker:
@hwdef There is still a lgtm label missing, thank you.

PR #3561, opened 2024-07-05
Fix podgroup not created

liuyuanchun11:
During the rolling upgrade of replicaset, pg_controller occasionally receives the addPod event and creates a podgroup. Then, pg_controller receives the addReplicaSet (replicas = 0) event and deletes the corresponding podgroup (to solve the pg fc problem). The updateReplicaSet (replicas = 1) event is received but not processed. As a result, the pod group corresponding to the pod is not correctly created.

issue: #3563

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -72,9 +72,31 @@ func (pg *pgcontroller) addReplicaSet(obj interface{}) {
 
 	if *rs.Spec.Replicas == 0 {
 		pgName := batchv1alpha1.PodgroupNamePrefix + string(rs.UID)
+		klog.V(4).Infof("Delete podgroup %s for replicaset %s/%s spec.replicas == 0",
+			pgName, rs.Namespace, rs.Name)
 		err := pg.vcClient.SchedulingV1beta1().PodGroups(rs.Namespace).Delete(context.TODO(), pgName, metav1.DeleteOptions{})
 		if err != nil && !apierrors.IsNotFound(err) {
 			klog.Errorf("Failed to delete PodGroup <%s/%s>: %v", rs.Namespace, pgName, err)
+		}
+	}
+
+	// In the rolling upgrade scenario, the addReplicasSet(replicas=0) event may be received before
+	// the updateReplicaSet(replicas=1) event. In this event, need to create PodGroup for the pod.
+	if *rs.Spec.Replicas > 0 {
+		selector := metav1.LabelSelector{MatchLabels: rs.Spec.Selector.MatchLabels}
+		podList, err := pg.kubeClient.CoreV1().Pods(rs.Namespace).List(context.TODO(),
+			metav1.ListOptions{LabelSelector: metav1.FormatLabelSelector(&selector)})
+		if err != nil {
+			klog.Errorf("Failed to list pods for ReplicaSet <%s/%s>: %v", rs.Namespace, rs.Name, err)
+			return
+		}
+		if podList != nil && len(podList.Items) > 0 {
+			pod := podList.Items[0]
+			klog.V(4).Infof("Try to create podgroup for pod %s/%s", pod.Namespace, pod.Name)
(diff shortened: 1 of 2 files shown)


googs1025:
Is there a scenario that can be reproduced?

liuyuanchun11:
> Is there a scenario that can be reproduced?

There is a high probability that this issue occurs when multiple deployments are upgraded in rolling mode. The key point is that the add pod event is received before the replicaset event.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -72,10 +72,30 @@ func (pg *pgcontroller) addReplicaSet(obj interface{}) {
 
 	if *rs.Spec.Replicas == 0 {
 		pgName := batchv1alpha1.PodgroupNamePrefix + string(rs.UID)
+		klog.V(4).Infof("Delete podgroup %s for replicaset %s/%s spec.replicas == 0",
+			pgName, rs.Namespace, rs.Name)
 		err := pg.vcClient.SchedulingV1beta1().PodGroups(rs.Namespace).Delete(context.TODO(), pgName, metav1.DeleteOptions{})
 		if err != nil && !apierrors.IsNotFound(err) {
 			klog.Errorf("Failed to delete PodGroup <%s/%s>: %v", rs.Namespace, pgName, err)
 		}
+	} else if *rs.Spec.Replicas > 0 {
```
Monokaix, line 81: else is no need and we should add a comment to explain why this is added.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -72,10 +72,30 @@ func (pg *pgcontroller) addReplicaSet(obj interface{}) {
...
+		podList, err := pg.kubeClient.CoreV1().Pods(rs.Namespace).List(context.TODO(),
+			metav1.ListOptions{LabelSelector: metav1.FormatLabelSelector(&selector)})
+		if err != nil {
+			klog.Errorf("Failed to list pods for ReplicaSet <%s/%s>: %v", rs.Namespace, rs.Name, err)
+			return
+		}
+		if podList != nil {
+			for _, pod := range podList.Items {
+				klog.V(4).Infof("Try to create podgroup for pod %s/%s", pod.Namespace, pod.Name)
+				err := pg.createNormalPodPGIfNotExist(&pod)
```
Monokaix, line 92: How about just check whether rs associated pg exists first, only create pg when it doesn't exist to avoid unnecessary API request, cause it's a rare case here.

liuyuanchun11: createNormalPodPGIfNotExist will checks if the podgroup exists.

image

Monokaix: Ok.

pkg/controllers/podgroup/pg_controller_handler.go
liuyuanchun11: Sorry, I don't understand what you mean by else is not needed. Are you saying that the webhook already guarantees that replicas are unsigned, so there's no need to write a > 0 check? 
The comment I'll add later.

Monokaix: Sorry for misunderstanding, I mean we can use `if *rs.Spec.Replicas > 0` directly and no need to use `else if`.

liuyuanchun11: Modified based on review comments.

william-wang:
@liuyuanchun11 please log a issue and add the description for the bug and associate the pr with issue.

liuyuanchun11:
> @liuyuanchun11 please log a issue and add the description for the bug and associate the pr with issue.


https://github.com/volcano-sh/volcano/issues/3563

> @liuyuanchun11 please log a issue and add the description for the bug and associate the pr with issue.

done, pr has been associated with issue #3563

william-wang (approved):
/approve

PR #3569, opened 2024-07-09
Update Kubernetes compatibility

Monokaix:
Former k8s version 1.19/1.20 only has `v1beat1.PodDisruptionBudget` and volcano use `v1.PodDisruptionBudget` from release 1.9, so older k8s version is not supported any more.


Monokaix:
cc @wangyang0616

hwdef (commented):
/lgtm

william-wang (approved):
/approve

Issue #3563, opened 2024-07-08
In the rolling upgrade scenario of multiple deployments, there is a possibility that the corresponding pod group is not created for ReplicaSet.


liuyuanchun11:
**What happened**:
In the rolling upgrade scenario of multiple deployments, there is a possibility that the corresponding pod group is not created for ReplicaSet.

**What you expected to happen**:
During the rolling upgrade of replicaset, pg_controller occasionally receives the addPod event and creates a podgroup. Then, pg_controller receives the addReplicaSet (replicas = 0) event and deletes the corresponding podgroup (to solve the pg fc problem). The updateReplicaSet (replicas = 1) event is received but not processed. As a result, the pod group corresponding to the pod is not correctly created.

**How to reproduce it (as minimally and precisely as possible)**:
Create multiple deployments and perform a rolling upgrade. There is a possibility that the bug recurs.

**Anything else we need to know?**:

**Environment**:
- Volcano Version: v1.9.0
- Kubernetes version (use `kubectl version`): v1.23.1
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

liuyuanchun11:
fix bug pr:
https://github.com/volcano-sh/volcano/pull/3561

PR #3574, opened 2024-07-09
[cherry-pick for release-1.10]Fix podgroup not created

liuyuanchun11:
During the rolling upgrade of replicaset, pg_controller occasionally receives the addPod event and creates a podgroup. Then, pg_controller receives the addReplicaSet (replicas = 0) event and deletes the corresponding podgroup (to solve the pg fc problem). The updateReplicaSet (replicas = 1) event is received but not processed. As a result, the pod group corresponding to the pod is not correctly created.

issue: https://github.com/volcano-sh/volcano/issues/3563


william-wang (approved):
/approve

PR #3576, opened 2024-07-09
fix ut that priority can not be guaranteed cross nodes when choose victims

lowang-bh:
fix ut that priority can not be guaranteed cross nodes when choose victims


lowang-bh:
@Monokaix @hwdef

hwdef (commented):
/lgtm

shinytang6 (approved):
/approve

PR #3487, opened 2024-05-24
Add dynamic resource allocation design details.

Subhasish-Behera:

Issue #3143, opened 2023-09-27
[CNCF LFX 2024 01-Mar-May] DRA integration 


geetasg:
#### What would you like to be added:
volcano scheduler should work with DRA drivers for the devices. 
The integration should include following - 
* While assigning a pod to a node, the scheduler should consult the DRA driver for the requested resources. 
* The resources should be described using the ResourceClaims and made available as CDI devices to the container.

#### Why is this needed:
Similar to https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/3063-dynamic-resource-allocation/README.md#motivation

Monokaix:
Hi, sorry for delay, that's a nice feature, we need evaluate the workload to implement this and design first, if you have any idea, welcome communicate and contribute！

verma-kunal:
Hi @geetasg , hope you are doing well! This issue seems interesting to me to work on as part of the LFX spring terms. Though i'm a beginner in Go & new to this project, i have a basic/general familiarity with how Dynamic resource allocation works in k8s. 

Would be great if you could help/point out a few resources to understand this a bit more & i could start working on the proposal!
thank you

Monokaix:
Hi @verma-kunal, I'm  Xuzheng from CNCF lFX volcano project mentorship, I cannot reply to you using the original email address , due to some security reasons, andI have replied to you using another email. Please check it.: )

verma-kunal:
Thanks for taking out the time to reply @Monokaix ! I really appreciate it. Will surely check it out ✨

verma-kunal:
> Hi @verma-kunal, I'm  Xuzheng from CNCF LFX volcano project mentorship, I cannot reply to you using the original email address due to some security reasons, andI have replied to you using another email. Please check it.: )

Thanks for taking out the time to reply @Monokaix ! Will check it out

verma-kunal:
Hey @Monokaix , is there a way I can join the slack server of Volcano, for any discussion or doubts? The [link](https://volcano-sh.slack.com/signup#/domain-signup) mentioned on the website requires you to have a `huawei` email address.

![Screenshot 2024-01-28 at 10 44 43 AM](https://github.com/volcano-sh/volcano/assets/72245772/ad8d3115-e404-4adf-9855-ac50e868ce62)

Monokaix:
[...]

PR #3577, opened 2024-07-10
[WIP]Add separate plugin for DRA(Dynamic Resource Allocation).

Subhasish-Behera:
This WIP PR is for DRA(Dynamic resource allocation) is the new way of requesting access to resources available in Kubernetes 1.26+.

Recent discussions in the Volcano community address the need to implement a way to DRA. It would extend
 From an end user’s perspective, It moves away from the limited "countable" interface for requesting access to resources  (e.g. "nvidia.com/gpu: 2"), providing an API more akin to that of persistent volumes.
 
 Issue Reference: https://github.com/volcano-sh/volcano/issues/3143
 Design Details: https://github.com/volcano-sh/volcano/pull/3487

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -47,24 +47,25 @@ const (
 
 // ServerOption is the main context object for the controller manager.
 type ServerOption struct {
-	KubeClientOptions    kube.ClientOptions
-	CertFile             string
-	KeyFile              string
-	CaCertFile           string
-	CertData             []byte
-	KeyData              []byte
-	CaCertData           []byte
-	SchedulerNames       []string
-	SchedulerConf        string
-	SchedulePeriod       time.Duration
-	EnableLeaderElection bool
-	LockObjectNamespace  string
-	DefaultQueue         string
-	PrintVersion         bool
-	EnableMetrics        bool
-	ListenAddress        string
-	EnablePriorityClass  bool
-	EnableCSIStorage     bool
+	KubeClientOptions                kube.ClientOptions
+	CertFile                         string
+	KeyFile                          string
+	CaCertFile                       string
+	CertData                         []byte
+	KeyData                          []byte
+	CaCertData                       []byte
+	SchedulerNames                   []string
+	SchedulerConf                    string
+	SchedulePeriod                   time.Duration
+	EnableLeaderElection             bool
+	LockObjectNamespace              string
+	DefaultQueue                     string
+	PrintVersion                     bool
+	EnableMetrics                    bool
+	ListenAddress                    string
+	EnablePriorityClass              bool
+	EnableCSIStorage                 bool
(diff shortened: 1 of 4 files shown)


cmd/scheduler/app/options/options.go
```diff
@@ -47,24 +47,25 @@ const (
...
+	SchedulePeriod                   time.Duration
+	EnableLeaderElection             bool
+	LockObjectNamespace              string
+	DefaultQueue                     string
+	PrintVersion                     bool
+	EnableMetrics                    bool
+	ListenAddress                    string
+	EnablePriorityClass              bool
+	EnableCSIStorage                 bool
+	EnableDynamicResourcesAllocation bool
```
Monokaix, line 68: We should set a separate feature struct like https://github.com/volcano-sh/volcano/blob/33a9eb5b96cddc3c076a439ed134d41a0d095220/pkg/scheduler/plugins/predicates/predicates.go#L291，which is imported from k8s instead of putting it in `ServerOption`, cause it's a plugin switch.

pkg/scheduler/plugins/dynamicresources/dynamicresources.go
```diff
@@ -0,0 +1,1317 @@
...
+	// Name is the name of the plugin used in Registry and configurations.
+	pluginName = names.DynamicResources
+
+	stateKey k8sframework.StateKey = pluginName
+)
+
+// The state is initialized in PreFilter phase. Because we save the pointer in
+// framework.CycleState, in the later phases we don't need to call Write method
+// to update the value
+type stateData struct {
```
Monokaix, line 64: Wr should put all the codes copyed from k8s in a separate file and give a comment that these codes are copyed from k8s，once dra has been published as a public struct we should remove them.

pkg/scheduler/plugins/dynamicresources/dynamicresources.go
```diff
@@ -0,0 +1,1317 @@
...
+type dynamicResources struct {
+	enabled                    bool
+	fh                         k8sframework.Handle
+	clientset                  kubernetes.Interface
+	claimLister                resourcev1alpha2listers.ResourceClaimLister
+	classLister                resourcev1alpha2listers.ResourceClassLister
+	podSchedulingContextLister resourcev1alpha2listers.PodSchedulingContextLister
+}
+
+//func (pl *dynamicResources) OnSessionClose(ssn *framework.Session) {
```
Monokaix, line 266: All the commented codes should be removed: )

Monokaix:
I think we should also add an e2e test to cover this: )

Monokaix:
Please also sign off your commit with `git commit -s`.

JesseStutler:
This PR may need to close, this PR will follow up to finish introducing DRA: https://github.com/volcano-sh/volcano/pull/3799

Issue #3579, opened 2024-07-10
Optimize the --version command


lengrongfu:
#### What would you like to be added:

I found current `--version` command is in the `app.Run()` internal, but maybe before this having fatal, 
eg: 
![image](https://github.com/volcano-sh/volcano/assets/15009201/c4dc7504-7622-4bd7-aa57-817a1edad59e)

https://github.com/volcano-sh/volcano/blob/fb068bde7a0cf80392b225517dd3dd522524091a/cmd/webhook-manager/main.go#L60-L64

#### Why is this needed:

I hope that the execution of viewing the version does not depend on any external environment.

PR #3580, opened 2024-07-10
optimization verion command

lengrongfu:
Fixes: https://github.com/volcano-sh/volcano/issues/3579

Test result:
image

### DIFF
--- a/cmd/scheduler/app/server.go
+++ b/cmd/scheduler/app/server.go
@@ -30,7 +30,6 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/signals"
 	commonutil "volcano.sh/volcano/pkg/util"
-	"volcano.sh/volcano/pkg/version"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
@@ -56,10 +55,6 @@ import (
 
 // Run the volcano scheduler.
 func Run(opt *options.ServerOption) error {
-	if opt.PrintVersion {
-		version.PrintVersionAndExit()
-	}
-
 	config, err := kube.BuildConfig(opt.KubeClientOptions)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


jasonliu747 (commented):
/lgtm

cmd/scheduler/app/server.go
```diff
@@ -56,10 +55,6 @@ import (
 
 // Run the volcano scheduler.
 func Run(opt *options.ServerOption) error {
-	if opt.PrintVersion {
-		version.PrintVersionAndExit()
```
googs1025, line 60: I have a question: Why do we need to move it to main.go? Can we just return here?

cmd/webhook-manager/main.go
```diff
@@ -29,7 +29,8 @@ import (
 
 	"volcano.sh/volcano/cmd/webhook-manager/app"
 	"volcano.sh/volcano/cmd/webhook-manager/app/options"
-
+	"volcano.sh/volcano/pkg/version"
```
googs1025, line 32: nit: import lint need to be fixed

cmd/scheduler/app/server.go
lengrongfu: Maybe it's in this location running error, but i only need know version info.
https://github.com/volcano-sh/volcano/blob/fb068bde7a0cf80392b225517dd3dd522524091a/cmd/scheduler/main.go#L65-L67

cmd/webhook-manager/main.go
lengrongfu: done

cmd/scheduler/app/server.go
lengrongfu: Do you have any idea? @googs1025

cmd/scheduler/app/server.go
```diff
@@ -56,10 +55,6 @@ import (
 
 // Run the volcano scheduler.
 func Run(opt *options.ServerOption) error {
-	if opt.PrintVersion {
```
Monokaix, line 59: Hi, welcome! A little question here, when doing secondary development based on volcano and we need to add additional `opt.PrintVersion` when calling `app.Run`, which is not very friendly to developers.

lengrongfu: So do we add a `PrintVersion` in main.go and also add a `PrintVersion` in Run?

Issue #3555, opened 2024-07-02
Support Pod Scheduling Readiness


ykcai-daniel:
#### What would you like to be added:
As mentioned in #3187, we hope to support `.spec.schedulingGates` flag of a pod. Scheduling Gates is a beta feature in Kubernetes v1.27. In kube-scheduler, if this field of a Pod is not empty, the pod will not be scheduled by it. Volcano should also support this behavior.

#### Why is this needed:
#3187 use the situation of an external quota manager to shouw why schedulingGates is needed.

>By specifying/removing a Pod's .spec.schedulingGates, user can control when a Pod is ready to be considered for scheduling.
Here is a typical scenario from community user.

>Problem Statement: We’ve implemented an external quota manager responsible for reviewing all incoming POD requests for capacity/quota requirements. Only once these requests receive approval from the quota manager are they considered eligible for scheduling. Consequently, we intend to leverage the pods schedulingGates feature to realize this functionality.

>It seems that Volcano is not adhering to this feature. It proceeds to schedule pods without honoring schedulingGates flag. This behavior is in contrast to the default Kubernetes kube-scheduler, which functions as expected.
Even though these pods have the schedulingGates added and were assigned to a node by the Volcano scheduler, it appears that they are subsequently rejected by the Kubernetes API server. I believe this error is occurring after the execution of the Volcano scheduler.

>Failed to bind pod <namsepace/pod-name> to node : &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"Operation cannot be fulfilled on pods/binding \”app-name\”: pod pod-name has non-empty .spec.schedulingGates", Reason:"Conflict", Details:(*v1.StatusDetails)(0x140002188a0), Code:409}}

ykcai-daniel:
Lates PR: #3658

PR #3581, opened 2024-07-10
Proposal for Support of Pod Scheduling Readiness

ykcai-daniel:
Proposal to add support for Pod Scheduling Readiness as mentioned in #3555

### DIFF
--- a/docs/design/pod-scheduling-readiness.md
+++ b/docs/design/pod-scheduling-readiness.md
@@ -1,0 +1,29 @@
+# Pod Scheduling Readiness
+## Motivation
+
+Pod Scheduling Readiness is a beta feature in Kubernetes v1.27. Users expect Volcano to be aware of it. By specifying/removing a Pod's `.spec.schedulingGates`, which is an array of strings, users can control when a Pod is ready to be considered for scheduling. The scheduling gates are usually removed by external controllers. `kube-scheduler` will only schedule a gated pod once all scheduling gates of it are removed (i.e. the `.spec.schedulingGates` is an empty arry)\. Note that for K8S resources that use `PodTemplate` (Deployment, StatefulSet etc.), if the `.spec.schedulingGates` field of the template is emptied, the gates of all the pods created based on that template will also be removed. 
+
+This feature is often used to implement more performant Quota Managers. K8S have a native ResourceQuota resource, which rejects a pod if it exceed the quota. However, if a quota manager is implemented using ResourceQuota,  one downside is user need to keep polling the api-server until there are sufficient resources the pod is schedulable. To solve this problem, scheduling gates enable users to first create pods that are not yet schedulable, and by removing the gates, it signals kube-scheduler that the pod is ready to be allocated to a node, thus no longer require polling.
+
+
+## Function Detail
+To support Pod Scheduling Readiness, the modifications mainly happen in `scheduler/actions`. In each actions except Enqueue such as Allocation, Preempte, Backfill, the scheduler will skip the tasks of a job whose Pods are scheduling gated, ensuring that a scheduling gated pod will not be bound to a node.
+
+
+
+
+## Feature Interactions
+1. **Actions**: Jobs with scheduling gates will first be enqueued by the Enqueue action and transit to the Inqueue state. In the following allocate, preempt actions, tasks with scheduling gated Pod will be skipped, which might result in gang-unschedulable job.
+2. **Plugins**: Since scheduling gated pods are not ready to be allocated, we don't want scheduling gated tasks to consume inqueue resources, making other potentially schedulable jobs uninqueuable. Therefore, proportion, capacity and overcommit plugins have to be changed. Since scheduling gated pods will not be allocated, we only need to deduct it from inqueue resources.
+3.  **Events and Conditions**: As shown below, when listing Pod with Kubectl, the pods that are shown below will have `SchedulingGated` status, which is the Reason of the condition in pod status. To align with K8S, we need to skip scheduling gated tasks when updating pods conditions after each scheduling cycle and show the original condition created by K8S.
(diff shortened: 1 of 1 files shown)


ykcai-daniel:
@Monokaix Please have a look first. Will fix the CI problems soon.

Monokaix:
You can sign off your commit with `git  commit -s`: )

docs/design/pod-scheduling-readiness.md
```diff
@@ -0,0 +1,29 @@
+# Pod Scheduling Readiness
+## Motivation
+
+Pod Scheduling Readiness is a beta feature in Kubernetes v1.27. Users expect Volcano to be aware of it. By specifying/removing a Pod's `.spec.schedulingGates`, which is an array of strings, users can control when a Pod is ready to be considered for scheduling. For Pods with none-empty `schedulingGates`, it will only be removed by kube-scheduler once all the gates are removed.
```
Monokaix, line 4: >it will only be removed by kube-scheduler once all the gates are removed.

Actually the `schedulingGates` field is usually removed by external controllers and kube-scheduler just check this filed to determine whether to schedule.

docs/design/pod-scheduling-readiness.md
```diff
@@ -0,0 +1,29 @@
...
+
+The following example illustrates why support for Scheduling Readiness is needed in Volcano. Suppose we have implemented an external quota manager responsible for reviewing all incoming pod requests for capacity/quota requirements. Only once these requests receive approval from the quota manager are they considered eligible for scheduling. The pods schedulingGates feature can be handy when implementing this funtionality
```
Monokaix, line 6: Please add a full stop of this passage.

docs/design/pod-scheduling-readiness.md
```diff
@@ -0,0 +1,29 @@
...
+
+//...
+)
+
+```
+2. **Plugins**: Scheduler Plugins such as proportion  will register functions to determine whether a PodGroup is enqueueable or allocatable. These functions calculate the resources already used in the cluster based on states of each PodGroup. For example, the proportion plugin determines whether a task is enqueuable when `sum(Inqueue)+sum(Running)+cur_job<TotalResource+elastic`. Since the scheduling gated jobs do not occupy resources in the cluster, by having the scheduling gated job in `SchGated` rather than `Inqueue`, they will not be summed up when calculating total used resources, which reflects the actual situation.
+3. **Actions**: Transition to `SchGated` happens to and from `Inqueue` first in the allocation action. Then, other actions will skip the gated pods.
+4. **Controllers**: One alternative design of state transition is to directly transition from `Pending` to `SchGated`. However, this is not possible because Controllers only create Pods for a job once it is inqueued (see [delayed-pod-creation](./delay-pod-creation.md)). We can only know a Pod is scheduling gated in the Inqueue state. Therefore, we need to transition to SchGated from Inqueue.
```
Monokaix, line 27: >Therefore, we need to transition to SchGated from Inqueue.

We should clarify that only the first time pg is created should we transit pg to Inqueue from Pending, and when pods of current pg are created, we transit pg to SchGated from Inqueue.

docs/design/pod-scheduling-readiness.md
```diff
@@ -0,0 +1,29 @@
...
+## Granularity of Pod Scheduling Readiness
+Pod scheduling readiness is a field of the spec of a pod. However, Volcano schedules Jobs, which consists of many tasks, each corresponding to a Pod. It is possible that some of these pods are scheduling gated while others are not. To align the granularity of the Pod Scheduling Readiness and Job, we see scheduling gates as a property of a Job: as long as there is one gated pod, the job is scheduling gated. This is consistent with the workload of Volcano: most Volcano Jobs need to run as a whole and cannot be partially run.
```
Monokaix, line 29: Maybe we can add that only voclano jobs have this limitation, and for normal pod/Deployment/Statefulset, etc, it's still a pod level gated feature.

Monokaix:
Antoher part that we should concern is the observability, when a pod is scheduling gated, what's the behavior of a pod and its pg? should we reports some events to let users know what happen, maybe we can also refer to kube-scheduler to determine whethet we should report events and what's the frequency of the event reported.

docs/design/pod-scheduling-readiness.md
ykcai-daniel: Maybe we should also emphasize atomicity of Pod creation:  Once the state of a pg is transitioned from `Pending` to `Inqueue`, I think all Pods of that pg will be created. In other words, the scheduler will never see a Job with partially created Pods. If this is not the case, the following situation might occur:

Consider a Job with four pods: p1,p2,p3,p4
Job enqueued ->p1,p2 created with gates -> p1,p2 gates removed -> Job allocated to node -> p3, p4 created with gates
We will end up with a Running Job but with scheduling gates

Monokaix: goog catch!

ykcai-daniel:
Proposal will be updated based on the new design, main changes include:

- no longer add new state of PodGroup. Instead, update the plugins for correct inqueue resource computation
- In allocation and preempt actions, skip scheduling gated tasks instead of the entire job

ykcai-daniel:
@Monokaix design doc updated according to latest implementation #3658

docs/design/pod-scheduling-readiness.md
```diff
@@ -0,0 +1,29 @@
...
+4. **Controllers**: K8S native resources like Deployment support template level removal of scheduling gates. In other words, if a Deployment has scheduling gates in its pod template, by patching the Deployment and remove the scheduling gates, all its pods will be deleted and recreated without gates. However, for Vcjob, currently the job controller cannot detect changes in PodTemplate and cannot support this feature. Despite this, it is uncommon to remove scheduling gates from PodTemplate and the Pod Scheduling Gates feature is usually used at pod level. What's more, scheduling gates are often added by webhooks instead of in the Job template (more details in K8S KEP). Therefore, we choose to not align this behavior with K8S.
+
+## Limitations
+1. **Vcjob support removing scheduling gates in template**: As mentioned above, currently, if we remove the scheduling gates field in a Vcjob, the gates of its pods are not removed. This behavior of Vcjob is different from native K8S resources.
```
Monokaix, line 22: These are limitations not our expectations, please modify them: )

william-wang (approved):

PR #3586, opened 2024-07-14
aggregate err when checking options

googs1025:
/kind feature
- aggregate err when checking options

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/component-base/config"
 	componentbaseconfigvalidation "k8s.io/component-base/config/validation"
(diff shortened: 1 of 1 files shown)


cmd/controller-manager/app/options/options.go
```diff
@@ -120,14 +121,27 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
...
 
-// CheckOptionOrDie checks the option and returns error if it's invalid
+// CheckOptionOrDie checks all options and returns all errors if they are invalid.
+// If there are any invalid options, it aggregates all the errors and returns them.
 func (s *ServerOption) CheckOptionOrDie() error {
-	// check controllers option
+	var allErrors []error
+
+	// Check controllers option
 	if err := s.checkControllers(); err != nil {
```
Monokaix, line 130: Before this pr, err will return directly once `checkControllers` has err, but after the change, other err will also be checked, which cause more code be executed, is this reasonable?

googs1025: IMO, this makes sense, because we are checking the options. The options should be checked as a whole and throw all errors, rather than throwing errors at each step.

cmd/controller-manager/app/options/options.go
```diff
@@ -120,14 +121,27 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
...
+	leaderElectionErr := componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(
+		&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
+	if leaderElectionErr != nil {
+		allErrors = append(allErrors, leaderElectionErr)
 	}
-	// check leader election flag when LeaderElection is enabled.
-	return componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
+
+	if len(allErrors) > 0 {
+		return errors.NewAggregate(allErrors)
```
Monokaix, line 142: How about just `return errors.NewAggregate(allErrors)`?

googs1025: sgtm, done

cmd/controller-manager/app/options/options.go
lowang-bh: How about k8s's implements? Some time only the first error is need to be paid attention to.

googs1025: https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/options/options.go#L267

We can refer to the implementation of kube-scheduler. The configuration is returned only after the entire verification is completed. The same goes for other components

hwdef (commented):
/lgtm
It is good to me.

If there are multiple errors, they can be detected on a single startup

PR #3588, opened 2024-07-14
optimize the webhook manager

Vacant2333:
1. The`ForEachAdmission` will not tell u if a webhook not register to the manager, i find this problem when i use this manager in my project, and its hard to find

### DIFF
--- a/pkg/webhooks/router/admission.go
+++ b/pkg/webhooks/router/admission.go
@@ -58,6 +58,8 @@ func ForEachAdmission(config *options.Config, handler func(*AdmissionService) er
 			if err := handler(service); err != nil {
 				return err
 			}
+		} else {
+			return fmt.Errorf("enabled admission %s not found on the admission registered map", admission)
 		}
 	}
 	return nil


Monokaix:
Hi, please make CI happy.

Monokaix:
If we remove the existing flag directly, users who upgrade from an old verison meet errors about the flag, so it's not recommended to remove them: )

Monokaix:
I think these flags are used to let kube-apiserver know the address of wenhook and access it so we'd better keep them.

cmd/webhook-manager/app/server.go
```diff
@@ -31,6 +32,7 @@ import (
 
 	"volcano.sh/apis/pkg/apis/helpers"
 	"volcano.sh/apis/pkg/apis/scheduling/scheme"
+
```
googs1025, line 35: nit: this space should not be required

cmd/webhook-manager/app/server.go
```diff
@@ -46,10 +48,6 @@ func Run(config *options.Config) error {
 		}
 	}
 
-	if config.WebhookURL == "" && config.WebhookNamespace == "" && config.WebhookName == "" {
```
googs1025, line 49: These flags are required and need to be specified when starting webhook

cmd/webhook-manager/app/server.go
```diff
@@ -101,7 +102,7 @@ func Run(config *options.Config) error {
 	}
 	go func() {
 		err = server.ListenAndServeTLS("", "")
-		if err != nil && err != http.ErrServerClosed {
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
```
googs1025, line 105: IMO, this change does not seem to change any logic. Can you explain the benefits of such a change?

pkg/webhooks/router/admission.go
```diff
@@ -58,6 +58,8 @@ func ForEachAdmission(config *options.Config, handler func(*AdmissionService) er
 			if err := handler(service); err != nil {
 				return err
 			}
+		} else {
+			return fmt.Errorf("enabled admission %s not found on the admission registered map", admission)
```
googs1025, line 62: Why is there an error if admission is not found?

cmd/webhook-manager/app/server.go
Vacant2333: you r right, i have remove this change, we dont need this change now

pkg/webhooks/router/admission.go
Vacant2333: when i registered a custom admission to the Volcano webhook manager, an issue occurred where the admission was registered incorrectly due to my input error. This problem was very difficult to troubleshoot because it doesn't explicitly tell you that the admission you specified was actually not registered.

Monokaix:
Please squash it to one commit：）

PR #3589, opened 2024-07-14
feat:adding sidecar

### LINKED ISSUE #3513: Volcano support SidecarContainer
Add sidecarContainer support for volcano scheduler, which needs volcano scheduler re-compute pod resource requests.

#### Why is this needed:
SideCarContainer is a beta feature and is enabled by default in k8s v1.29, volcano has not yet adapt the feature, this may cause the scheduler to undercalculate the resources for the pod, and then the kubelet admit failed after the pod is scheduled to the node, because the sidecarcontainer is a new kind of initcontainer, so the pod initialization request resources need to be recalculated.

Refer to:

1. https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/753-sidecar-containers#resources-calculation-for-scheduling-and-pod-admission
2. https://github.com/kubernetes/kubernetes/pull/116429

7h3-3mp7y-m4n:
This added a feature of sidecar 
Fixes #3513

### DIFF
--- a/pkg/features/volcano_features.go
+++ b/pkg/features/volcano_features.go
@@ -43,6 +43,8 @@ const (
 
 	// ResourceTopology supports resources like cpu/memory topology aware.
 	ResourceTopology featuregate.Feature = "ResourceTopology"
+
+	SidecarContainers featuregate.Feature = "SidecarContainers"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


lowang-bh:
Unit test is need!

pkg/features/volcano_features.go
```diff
@@ -43,6 +43,8 @@ const (
 
 	// ResourceTopology supports resources like cpu/memory topology aware.
 	ResourceTopology featuregate.Feature = "ResourceTopology"
+
```
googs1025, line 46: Can we add comment and reference documents to let users know the context of this feature gate?

Monokaix: `SidecarContainers` is a kubernetes intree feature and has already a definition，and now we have imported other features like https://github.com/volcano-sh/volcano/blob/224ca98ab097d7788fc17ae5a6bbb0ba7a191653/pkg/scheduler/plugins/predicates/predicates.go#L291，so we doesn't need to re-define the feature.

Monokaix:
We should also add filter logic like https://github.com/kubernetes/kubernetes/blob/f8d5b2074c9b3a1629ef085b9e17a5370f393732/pkg/scheduler/framework/plugins/noderesources/fit.go#L217.

7h3-3mp7y-m4n:
> We should also add filter logic like https://github.com/kubernetes/kubernetes/blob/f8d5b2074c9b3a1629ef085b9e17a5370f393732/pkg/scheduler/framework/plugins/noderesources/fit.go#L217.

okay :)

Monokaix:
Hi, since it's not active and we will release v1.10 soon so I fire a new [PR](https://github.com/volcano-sh/volcano/pull/3706) based on your work and your commit is also inclueded, thanks for your contribution!

Issue #3205, opened 2023-11-16
Volcano Scheduler Memory leak


tianshimoyi:
**What happened**:

Check the memory usage of volcano scheduler. It is consistent with the increase. After restarting, the memory plummets.

### Monitor before restarting

<img width="711" alt="截屏2023-11-16 10 26 47" src="https://github.com/volcano-sh/volcano/assets/68532999/b180066e-1be7-491d-b0cb-f18e40d1ec36">

### Monitor after restart

<img width="697" alt="截屏2023-11-16 10 29 27" src="https://github.com/volcano-sh/volcano/assets/68532999/5e3062e9-5a27-4bd8-b465-a5022bec6c0b">

### Heap memory usage couple before restart
![截屏2023-11-16 10 30 35](https://github.com/volcano-sh/volcano/assets/68532999/fb8fde22-9fce-4242-9c54-6a7f5fce69fe)

### Heap memory usage couple after reboot
![截屏2023-11-16 10 31 15](https://github.com/volcano-sh/volcano/assets/68532999/bac05eab-01a0-44e3-ad1b-a8c4adc55b0f)

Monokaix:
Hi, have you set memory limit of container resource limit and what's the log level you set?

tianshimoyi:
@Monokaix After modifying the log level, the memory is still growing. Log printing speed is not fast either
![截屏2023-11-16 21 17 28](https://github.com/volcano-sh/volcano/assets/68532999/258f5b9f-626d-4013-8f95-c473d6e2f59f)
![截屏2023-11-16 21 18 30](https://github.com/volcano-sh/volcano/assets/68532999/89d22ee1-0837-4498-8927-a60d0c88a756)

Monokaix:
Can you wait memory hit the limit and check whether memory reclamation happened？

tianshimoyi:
@Monokaix I'll give it a try. I found a problem. The scheduler has the nodeSelector option, but the pods do not consider filtering out the situations that are not on these nodes, causing the memory to maintain the pod information on all nodes in the entire cluster, resulting in a waste of memory.

Monokaix:
> @Monokaix I'll give it a try. I found a problem. The scheduler has the nodeSelector option, but the pods do not consider filtering out the situations that are not on these nodes, causing the memory to maintain the pod information on all nodes in the entire cluster, resulting in a waste of memory.

That's right, but I think it's not the main cause of high memory usage: ).
[...]

PR #3593, opened 2024-07-15
fix scheduler memory leak

Wang-Kai:
Volcano has been fully implemented in the [Bilibili](https://www.bilibili.com/) transcoding scenario. As we migrated more tasks to Volcano, we also discovered more issues. Memory leakage is one of the most important issues. [As other users have discovered](https://github.com/volcano-sh/volcano/issues/3205), the memory of the Volcano scheduler gradually increases until it triggers an OOM (Out Of Memory) restart. 

image

During the lifecycle of a Pod, there will be multiple update events. Currently, each pod update triggers the destruction and recreation of tasks, and we have found this to be the root cause of the problem. After we switched to reusing task objects, the scheduler's memory stabilized, even after running for a long time.

This fix is also logical, as excessively frequent allocation of new memory can lead to memory fragmentation, and the garbage collector will also be under pressure. Sometimes Go runtime doesn't even have time to merge small memory blocks that are no longer in use.


pkg/scheduler/cache/event_handlers.go
```diff
@@ -310,10 +310,55 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
+	// 4. update taskInfo status
+	taskInfo.Pod = newPod
+	taskInfo.NodeName = newPod.Spec.NodeName
+	taskInfo.Preemptable = schedulingapi.GetPodPreemptable(newPod)
+	taskInfo.RevocableZone = schedulingapi.GetPodRevocableZone(newPod)
+	taskInfo.NumaInfo = schedulingapi.GetPodTopologyInfo(newPod)
+	taskInfo.Status = newTaskStatus
+
+	// reuse taskInfo struct and add it to job and node, avoid being recyceld by GC
+	if err := sc.addTask(taskInfo); err != nil {
```
googs1025, line 358: I have a small question. Is the `sc.addTask(taskInfo)` here repeated with `sc.addPod(newPod)`?

```go
// Assumes that lock is already acquired.
func (sc *SchedulerCache) addPod(pod *v1.Pod) error {
	pi, err := sc.NewTaskInfo(pod)
	if err != nil {
		klog.Errorf("generate taskInfo for pod(%s) failed: %v", pod.Name, err)
		sc.resyncTask(pi)
	}

	return sc.addTask(pi)
}
```

googs1025:
/cc 

Hello, I would like to confirm if my understanding is correct. The main question is, when updating, instead of deleting the original in-memory object and adding it again, should we update the original object and then add it again? Is that the intended meaning?

pkg/scheduler/cache/event_handlers.go
```diff
@@ -310,10 +310,55 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
+	// 1. update taskInfo.Pod pointer, we don't hold oldPod pointer so that oldPod can be GC
+	// 2. update node name for pod bind event
+	// 3. update any other fields relative with annotations
+	// 4. update taskInfo status
+	taskInfo.Pod = newPod
+	taskInfo.NodeName = newPod.Spec.NodeName
+	taskInfo.Preemptable = schedulingapi.GetPodPreemptable(newPod)
+	taskInfo.RevocableZone = schedulingapi.GetPodRevocableZone(newPod)
+	taskInfo.NumaInfo = schedulingapi.GetPodTopologyInfo(newPod)
+	taskInfo.Status = newTaskStatus
```
googs1025, line 355: How do we know that we only need to update these fields? There are many other fields in TaskInfo that we can be sure will not change when updating? Maybe using a `DeepCopy()` method would be better?

Wang-Kai:
> /cc
> 
> Hello, I would like to confirm if my understanding is correct. The main question is, when updating, instead of deleting the original in-memory object and adding it again, should we update the original object and then add it again? Is that the intended meaning?

I am honored to discuss this issue with you.

Yes, it is my meaning. To be precise, we should reuse the existing TaskInfo struct rather than discarding it and creating a new one. Because when the cluster is large enough, there are many pods, and the task state changes rapidly, this will involve a significant amount of memory allocation and deallocation.

pkg/scheduler/cache/event_handlers.go
Wang-Kai: Actually, I modified all the fields in the old TaskInfo struct that I believed could change, while for the other fields, I am unsure why they would change.

Regarding the `DeepCopy()` solution you mentioned, could you elaborate further? Does it ensure that a new object is not created?

lowang-bh:
> > /cc
> > Hello, I would like to confirm if my understanding is correct. The main question is, when updating, instead of deleting the original in-memory object and adding it again, should we update the original object and then add it again? Is that the intended meaning?
> 
> I am honored to discuss this issue with you.
> 
> Yes, it is my meaning. To be precise, we should reuse the existing TaskInfo struct rather than discarding it and creating a new one. Because when the cluster is large enough, there are many pods, and the task state changes rapidly, this will involve a significant amount of memory allocation and deallocation.

How about use syncpool, and then check weather the metrics have a change?

Wang-Kai:
About `syncpool `, can you describe it in more detail?  @lowang-bh

molei20021:
func (ti *TaskInfo) Clone() *TaskInfo
Pod:           ti.Pod,

I see in this function, pod is not deepcopied, which may cause garbage collector not collect when pod is deleted while scheduling

lowang-bh:
> About `syncpool `, can you describe it in more detail? @lowang-bh

As `JobInfo` and `TaskInfo` are repeatly created and deleted, we can use syncpool to store the unused object and reduce gc presure.

Monokaix:
> func (ti *TaskInfo) Clone() *TaskInfo Pod: ti.Pod,
> 
> I see in this function, pod is not deepcopied, which may cause garbage collector not collect when pod is deleted while scheduling

But the pod will eventually be GC after one scheduler ended，right? I think it's not the main problem here: )

Monokaix:
I'm still a little confused that if this is the root cause, if we stop submitting jobs the GC will not have much preessure and finally the memory will be GC, have you tested this case?

pkg/scheduler/cache/event_handlers.go
```diff
@@ -310,10 +310,55 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
+		taskInfo *schedulingapi.TaskInfo
+	)
+
+	jobInfo = sc.Jobs[jobID]
+	if jobInfo == nil {
+		return fmt.Errorf("job %s not found in cache.Jobs", jobID)
+	}
+	taskInfo = jobInfo.Tasks[taskID]
+	if taskInfo == nil {
+		return fmt.Errorf("task %s not found in job.Tasks", taskID)
```
kingeasternsun: Maybe here should `return sc.addPod(newPod)` ?

hwdef:
Still need to be considered

hwdef:
still need

fengruotj:
I also found the scheduler memory increase problem in our large-scale cluster。
heap_inuse (1)

image

fengruotj:
In our case, the scheduler 's go_memstats_alloc_bytes metrics (Number of bytes allocated and still in use) has been rising .

pkg/scheduler/cache/event_handlers.go
```diff
@@ -310,10 +310,55 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
+	var (
+		jobID    = schedulingapi.GetJobID(newPod)
+		taskID   = schedulingapi.TaskID(newPod.UID)
+		jobInfo  *schedulingapi.JobInfo
+		taskInfo *schedulingapi.TaskInfo
+	)
+
+	jobInfo = sc.Jobs[jobID]
+	if jobInfo == nil {
+		return fmt.Errorf("job %s not found in cache.Jobs", jobID)
```
JesseStutler: In a cluster with multiple schedulers, some pods may not be scheduled by Volcano, but still require cache updates. I believe that in this case, an error should not be returned.

pkg/scheduler/cache/event_handlers.go
JesseStutler: `DeepCopy` is too heavy, it adds extra memory usage but our target is to avoid OOM. What if we follow the pattern of `NewTaskInfo` and update only the fields that need updating? We could extract a common function.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -310,10 +310,55 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
 		return nil
 	}
+	// the version has not changed, nothing to do
+	if oldPod.ResourceVersion == newPod.ResourceVersion {
+		return nil
+	}
+	// task status has not changed, nothing to do
+	oldTaskStatus := schedulingapi.GetTaskStatus(oldPod)
+	newTaskStatus := schedulingapi.GetTaskStatus(newPod)
+	if oldTaskStatus == newTaskStatus {
```
kingeasternsun: We can’t simply return when the status is the same, because changes in Pod annotations or labels may also affect scheduling decisions.

kingeasternsun: We can’t simply return when the status is the same, because changes in Pod annotations or labels may also affect scheduling decisions.

Wang-Kai:
> What if we follow the pattern of NewTaskInfo and update only the fields that need updating? We could extract a common function.

Thanks @JesseStutler 

In practice, almost any field may change. Therefore, in Bilibili, we use `TaskInfo.UpdateByPod(*v1.Pod)` to update any field that might change. However, this approach has some drawbacks. For example, when a new field is added to `Task`, we also need to update the `UpdateByPod` function. Unfortunately, I can’t think of a better solution for now.

pkg/scheduler/cache/event_handlers.go
Wang-Kai: Thanks for the great advice. You truly have an eagle eye for finding bugs. 

We’ve recently identified this issue and fixed it in our internal version.

Wang-Kai:
CC @JesseStutler 

Hey, could you take a look at the CI process and review my code when you have time?

pkg/scheduler/cache/event_handlers.go
```diff
@@ -319,14 +319,24 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
 		return nil
 	}
 
-	if err := sc.deletePod(oldPod); err != nil {
+	taskInfo, err := sc.NewTaskInfo(newPod)
```
JesseStutler: Why do you remove those codes about getting the JobID of the pod and retrieving the existing TaskInfo? Isn't our target to reuse the existing task info?

pkg/scheduler/cache/event_handlers.go
```diff
@@ -319,14 +319,24 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
 		return err
 	}
+
 	//when delete pod, the ownerreference of pod will be set nil, just as orphan pod
 	if len(utils.GetController(newPod)) == 0 {
 		newPod.OwnerReferences = oldPod.OwnerReferences
 	}
-	return sc.addPod(newPod)
+	// update taskInfo by new pod
+	taskInfo.UpdateByPod(newPod)
```
JesseStutler, line 381: This doesn't seem right. The solution shouldn't be to get the old TaskInfo and then update it based on the new pod information, reusing the old TaskInfo? But you've already created a new TaskInfo based on the newPod earlier in L322

JesseStutler:
> CC @JesseStutler
> 
> Hey, could you take a look at the CI process and review my code when you have time?

@Wang-Kai I think it's because that current logic is not correct because other PR can pass CIs, in my local environment, I use your code build components but also will fail on the e2e testing

pkg/scheduler/cache/event_handlers.go
```diff
@@ -319,14 +319,24 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
+	return sc.addTask(taskInfo)
```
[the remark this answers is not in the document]
JesseStutler: @Wang-Kai  https://github.com/volcano-sh/volcano/pull/3593#discussion_r2498052670, gemini is correct

JesseStutler:
> > What if we follow the pattern of NewTaskInfo and update only the fields that need updating? We could extract a common function.
> 
> Thanks @JesseStutler
> 
> In practice, almost any field may change. Therefore, in Bilibili, we use `TaskInfo.UpdateByPod(*v1.Pod)` to update any field that might change. However, this approach has some drawbacks. For example, when a new field is added to `Task`, we also need to update the `UpdateByPod` function. Unfortunately, I can’t think of a better solution for now.

Correct, so we need to add some comments to tell other contributors they need to both modify `UpdateByPod` and `NewTaskInfo`, but I think it's worth because update pods is frequent and it will do cause pressure on the go gc.

What do you think? @kingeasternsun @hwdef @lowang-bh @hzxuzhonghu @wangyang0616

kingeasternsun:
> > > What if we follow the pattern of NewTaskInfo and update only the fields that need updating? We could extract a common function.
> > 
> > 
> > Thanks @JesseStutler
> > In practice, almost any field may change. Therefore, in Bilibili, we use `TaskInfo.UpdateByPod(*v1.Pod)` to update any field that might change. However, this approach has some drawbacks. For example, when a new field is added to `Task`, we also need to update the `UpdateByPod` function. Unfortunately, I can’t think of a better solution for now.
> 
> Correct, so we need to add some comments to tell other contributors they need to both modify `UpdateByPod` and `NewTaskInfo`, but I think it's worth because update pods is frequent and it will do cause pressure on the go gc.
> 
> What do you think? @kingeasternsun @hwdef @lowang-bh @hzxuzhonghu @wangyang0616

I agree, It's worth. We could further refactor the code by extracting a common internal function, updateTaskInfo(), which can be reused by both UpdateByPod and NewTaskInfo

fengruotj:
> I also found the scheduler memory increase problem in our large-scale cluster。 heap_inuse (1)
> 
> image

In the end, we solved this problem: [volcano-scheduler memory leak problem](https://github.com/volcano-sh/volcano/issues/4745)

pkg/scheduler/cache/event_handlers.go
```diff
@@ -319,14 +319,69 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
 	}
+
 	//when delete pod, the ownerreference of pod will be set nil, just as orphan pod
 	if len(utils.GetController(newPod)) == 0 {
 		newPod.OwnerReferences = oldPod.OwnerReferences
 	}
-	return sc.addPod(newPod)
+
+	// update taskInfo by new pod
+	taskInfo.UpdateByPod(newPod)
```
hzxuzhonghu, line 381: It can be very easy to forget updating this function later we add some new info into task

Issue #3587, opened 2024-07-14
--node-selector config doesn't consider for csinode


standleyM:
**What happened**:
In my k8s cluster, there are 10 nodes  and all listed in node and csinode resources.
Add '--node-selector' config to volcano-scheduler deployment to only select one node.  But if volcano scheduler started, all nodes are seen in volcano actions and plugins.

**What you expected to happen**:
Expect the node selected to used for volcano scheduler.

**How to reproduce it (as minimally and precisely as possible)**:
Set a unused selector in '--node-selector'.
After started, all nodes are ignored and all cisnode are seen. 

**Anything else we need to know?**:

Followings are copied from cache.go. The csinode informer is not considered by '--node-selector', and csinodes are always seen:
sc.csiNodeInformer.Informer().AddEventHandler(
cache.ResourceEventHandlerFuncs{
	AddFunc:    sc.AddOrUpdateCSINode,
	UpdateFunc: sc.UpdateCSINode,
	DeleteFunc: sc.DeleteCSINode,
}

Monokaix:
same as https://github.com/volcano-sh/volcano/issues/3516

PR #3594, opened 2024-07-15
Fix CSI node selector not take effect

Monokaix:
Fix https://github.com/volcano-sh/volcano/issues/3587.
Also remve the deprecated check `structcheck` cause it has been replaced by `unused` check.
```
WARN [runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused.
```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -50,6 +50,7 @@ import (
 	"k8s.io/klog/v2"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
+	"stathat.com/c/consistent"
 
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	"volcano.sh/apis/pkg/apis/scheduling"
@@ -101,7 +102,7 @@ type SchedulerCache struct {
 	defaultQueue string
 	// schedulerName is the name for volcano scheduler
 	schedulerNames     []string
-	nodeSelectorLabels map[string]string
+	nodeSelectorLabels map[string]sets.Empty
 	metricsConf        map[string]string
 
 	podInformer                infov1.PodInformer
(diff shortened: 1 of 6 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -157,6 +158,13 @@ type SchedulerCache struct {
 	// not be counted in pod pvc resource request and node.Allocatable, because the spec.drivers of csinode resource
 	// is always null, these provisioners usually are host path csi controllers like rancher.io/local-path and hostpath.csi.k8s.io.
 	IgnoredCSIProvisioners sets.Set[string]
+
+	multiSchedulerInfo
```
googs1025, line 162: Could we give more comment to this field so that everyone can better understand the context?

Monokaix: Thanks for reminding~have added the background info.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -605,13 +605,34 @@ func (sc *SchedulerCache) SyncNode(nodeName string) error {
...
+
+func (sc *SchedulerCache) nodeCanAddCache(node *v1.Node) bool {
+	if !responsibleForNode(node.Name, sc.schedulerPodName, sc.c) {
+		return false
+	}
+	if len(sc.nodeSelectorLabels) == 0 {
+		return true
+	}
+	for labelName, labelValue := range node.Labels {
+		key := labelName + ":" + labelValue
```
googs1025, line 629: I have a small question: Can this `nodeSelectorLabels map[string]string` be adjusted to` map[string]struct{}`? In fact, it is just used as a set, and there is no need to use string to store variables.

googs1025: This is a small problem, it doesn't matter if it is not changed

Monokaix: Yeah, it's an issues left over from history, but fix it is also ok.

Monokaix: done here.

william-wang (approved):
/approve

PR #3599, opened 2024-07-16
handle job not found case

Wang-Kai:
While the GC controller deletes a job, we currently return the result of `Delete()` function directly. For more rigorous consideration, we should handle the `NotFound` case and not add it to the queue again.

### DIFF
--- a/pkg/controllers/garbagecollector/garbagecollector.go
+++ b/pkg/controllers/garbagecollector/garbagecollector.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	vcclientset "volcano.sh/apis/pkg/client/clientset/versioned"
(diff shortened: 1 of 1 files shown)


pkg/controllers/garbagecollector/garbagecollector.go
```diff
@@ -232,7 +233,12 @@ func (gc *gccontroller) processJob(key string) error {
 		Preconditions:     &metav1.Preconditions{UID: &fresh.UID},
 	}
 	klog.V(4).Infof("Cleaning up Job %s/%s", namespace, name)
-	return gc.vcClient.BatchV1alpha1().Jobs(fresh.Namespace).Delete(context.TODO(), fresh.Name, options)
+	err = gc.vcClient.BatchV1alpha1().Jobs(fresh.Namespace).Delete(context.TODO(), fresh.Name, options)
+	if err != nil && apierrors.IsNotFound(err) {
```
Monokaix, line 237: How about
```go
if apierrors.IsNotFound(err) {
    // if the job had deleted, it will not be added to queue
	return nil
}
return err
```

Wang-Kai: GOOD, I modified it according to your way.

hzxuzhonghu (approved):
/lgtm

lowang-bh:
Please squash the commits.

PR #3600, opened 2024-07-16
modify log level

Wang-Kai:
Modify the log level to `5` for the case when no feasible node is found.

Failure to find a matching node is a very common problem. When the cluster is large enough and the number of pods is large enough, this log will be printed frequently and even drown out other valid logs. The current scheduler does not have a delay queue, and some tasks will keep printing similar logs. The default installation and deployment scheduler log level is 3. For such common scenarios, practical experience shows that level 5 is sufficient.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -271,7 +271,7 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 				metrics.UpdateE2eSchedulingLastTimeByJob(job.Name, string(job.Queue), job.Namespace, time.Now())
 			}
 		} else {
-			klog.V(3).Infof("Predicates failed in allocate for task <%s/%s> on node <%s> with limited resources",
+			klog.V(5).Infof("Predicates failed in allocate for task <%s/%s> on node <%s> with limited resources",
 				task.Namespace, task.Name, bestNode.Name)
 
 			// Allocate releasing resource to the task if any.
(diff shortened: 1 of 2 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -271,7 +271,7 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 				metrics.UpdateE2eSchedulingLastTimeByJob(job.Name, string(job.Queue), job.Namespace, time.Now())
 			}
 		} else {
-			klog.V(3).Infof("Predicates failed in allocate for task <%s/%s> on node <%s> with limited resources",
```
Monokaix, line 274: Is this a normal case here? because when cluster has no idle resources, predict will be failed and allocate for task will be breaked, and will not go through here.
But it's a reasonable modification, just want to know why codes go here: )

pkg/scheduler/util/predicate_helper.go
```diff
@@ -71,7 +71,7 @@ func (ph *predicateHelper) PredicateNodes(task *api.TaskInfo, nodes []*api.NodeI
 
 		// TODO (k82cn): Enable eCache for performance improvement.
 		if _, err := fn(task, node); err != nil {
-			klog.V(3).Infof("Predicates failed: %v", err)
+			klog.V(5).Infof("Predicates failed: %v", err)
```
Monokaix, line 74: what if user want to know the specific err message of each node?

lowang-bh: It is on the same order of magnitude as failed node count * pod count.

Wang-Kai: Thank you for the correction. So are you agree to modify the log level?

The current scheduler does not have a delay queue, and failed tasks will be retried immediately.

PR #3603, opened 2024-07-17
fix(controller): add statefulset gc for podgroup.

HalfBuddhist:
### what to fix?
While scheduling a k8s statefulset using volcano,  the generated podgroup of the related pods will be in `Inqueue` state even when setting the replica of the statefulset to zero.  And the Inqueued podgroup would still occupy resources, which is not reasonable.

### how to reproduce it?
scheduling a statefulset using volcano and  scale it to zero.

### how to fix in ths pr?
put a informer in controller, delete the pgs for the zero-replica statefulset in the add and the update event.


Monokaix:
Hi, please sign off your commit with `git commit -s`.
And please note that https://github.com/volcano-sh/volcano/pull/3561 has fixed some problems about pg, you can refer it.

HalfBuddhist:
1. add sign-off to commit message.
2. do the same fix in #3561 when rolling upgrade statefulset.

hwdef (commented):
/ok-to-test

Monokaix:
Thanks for your contribution ，this case should also be considered: )

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -105,6 +105,47 @@ func (pg *pgcontroller) updateReplicaSet(oldObj, newObj interface{}) {
...
+			klog.V(4).Infof("Try to create podgroup for pod %s/%s", pod.Namespace, pod.Name)
+			err := pg.createNormalPodPGIfNotExist(&pod)
+			if err != nil {
+				klog.Errorf("Failed to create PodGroup for pod <%s/%s>: %v", pod.Namespace, pod.Name, err)
+			}
+		}
+	}
+}
+
+func (pg *pgcontroller) updateStatefulSet(oldObj, newObj interface{}) {
```
hwdef, line 145: This function is not required

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -105,6 +105,47 @@ func (pg *pgcontroller) updateReplicaSet(oldObj, newObj interface{}) {
...
+			klog.Errorf("Failed to delete PodGroup <%s/%s>: %v", sts.Namespace, pgName, err)
+		}
+	}
+
+	// In the rolling upgrade scenario, the addStatefulSet(replicas=0) event may be received before
+	// the updateStatefulSet(replicas=1) event, and after the addPod event for the new created pod.
+	// In this event, need to create PodGroup for the pod.
+	if *sts.Spec.Replicas > 0 {
+		selector := metav1.LabelSelector{MatchLabels: sts.Spec.Selector.MatchLabels}
+		podList, err := pg.kubeClient.CoreV1().Pods(sts.Namespace).List(context.TODO(),
```
hwdef, line 128: How about use informer list

Monokaix:
> Thanks for your contribution ，this case should also be considered: )

Please also consider this: https://github.com/volcano-sh/volcano/pull/3672

hwdef:
still need

pkg/controllers/podgroup/pg_controller_handler.go
HalfBuddhist: This function is called in the update event of statefulset, in which the change in the number of replicas of statefulset should be checked to correctly handle podgroup resources.

This is exactly the problem I actually encountered.

Or did I misunderstand what you meant?

pkg/controllers/podgroup/pg_controller_handler.go
HalfBuddhist: Done in commit fd37d99cdc06c2961fe4b9e7e98006c6e0e26e7f  @hwdef

hwdef (commented):
/lgtm
Thanks for your contribution.

BTW, It is best to compress the number of commits, 4 are too much. In addition, your commit email does not match the github email address.

hwdef:
PTAL @JesseStutler @Monokaix @lowang-bh

HalfBuddhist:
> /lgtm Thanks for your contribution.
> 
> BTW, It is best to compress the number of commits, 4 are too much. In addition, your commit email does not match the github email address.

Done in commit b4a2ff9a557c24083e8ddbe200d5b694d9bcfd6b @hwdef

PR #3438, opened 2024-04-25
export the logs of Volcano components when the e2e test workflow fails

panoswoo:
Resolves #3429

PR #3609, opened 2024-07-18
Add container log size limit config for kind node

Monokaix:
This pr https://github.com/volcano-sh/volcano/pull/3438 exported logs when e2e has error, but when check logs, the volcano scheduler logs have been compressed cause the default log limit size is only 10Mi，and we can't see the complete logs, so this pr adjust the log limit.


lowang-bh:
PR #3599 breaks the CI

image

Monokaix:
> PR #3599 breaks the CI
> 
> image
> 

we should check why，seems that pr has no demaged behavior.

william-wang (approved):
/approve

PR #3548, opened 2024-06-27
metrics: use milliseconds instead of microseconds as the time unit for scheduling latency

microyahoo:
The current time unit cannot accurately reflect the actual scheduling time like below
```
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="5"} 0
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="10"} 0
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="20"} 0
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="40"} 0
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="80"} 0
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="160"} 0
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="320"} 1
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="640"} 1
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="1280"} 68550
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="2560"} 198253
volcano_action_scheduling_latency_microseconds_bucket{action="allocate",le="+Inf"} 243524
volcano_action_scheduling_latency_microseconds_sum{action="allocate"} 5.505664439029926e+08
```

Signed-off-by: Liang Zheng <zhengliang0901@gmail.com>

PR #3613, opened 2024-07-19
metrics_accurate

molei20021:
The metric volcano_action_scheduling_latency_microseconds origin bucket is defined as prometheus.ExponentialBuckets(5, 2, 10), which means the max microsecond is 2560us, in a large cluster, the action like allocate and preempt will cost about several seconds. so it is important to modify the max value of the bucket.


lowang-bh:
PR #3548 has tried to improved it.

Issue #3605, opened 2024-07-17
volcano vgpu metrics not update properly


archlitchi:
If you submit a vgpu job, you can see the corresponding metrics by using scheduler metrics, as follows:

task yaml:
```
apiVersion: v1
kind: Pod
metadata:
  name: pod1
spec:
  restartPolicy: OnFailure
  schedulerName: volcano
  containers:
  - image: ubuntu:20.04
    name: pod1-ctr
    command: ["sleep"]
    args: ["100000"]
    resources:
      limits:
        volcano.sh/vgpu-memory: 1024
        volcano.sh/vgpu-number: 1
```

then by visiting scheduler metrics, you can get the vgpu overview of vc-scheduler
```
curl {vc-scheduler}:8080/metrics
```

```
# HELP volcano_vgpu_device_allocated_cores The percentage of gpu compute cores allocated in this card
# TYPE volcano_vgpu_device_allocated_cores gauge
volcano_vgpu_device_allocated_cores{devID="GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec"} 0
volcano_vgpu_device_allocated_cores{devID="GPU-0fc3eda5-e98b-a25b-5b0d-cf5c855d1448"} 0
# HELP volcano_vgpu_device_allocated_memory The number of vgpu memory allocated in this card
# TYPE volcano_vgpu_device_allocated_memory gauge
volcano_vgpu_device_allocated_memory{devID="GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec"} 0
volcano_vgpu_device_allocated_memory{devID="GPU-0fc3eda5-e98b-a25b-5b0d-cf5c855d1448"} 1024
# HELP volcano_vgpu_device_memory_limit The number of total device memory allocated in this card
# TYPE volcano_vgpu_device_memory_limit gauge
volcano_vgpu_device_memory_limit{devID="GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec"} 32768
volcano_vgpu_device_memory_limit{devID="GPU-0fc3eda5-e98b-a25b-5b0d-cf5c855d1448"} 32768
# HELP volcano_vgpu_device_shared_number The number of vgpu tasks sharing this card
# TYPE volcano_vgpu_device_shared_number gauge
volcano_vgpu_device_shared_number{devID="GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec"} 0
[...]

PR #3614, opened 2024-07-19
fix: vgpu metrics not update when pod deleted

googs1025:
fix: https://github.com/volcano-sh/volcano/issues/3605

When a pod is deleted, we need to re-update the metrics

```go
// SubResource frees the gpu hold by the pod
func (gs *GPUDevices) SubResource(pod *v1.Pod) {
	ids, ok := pod.Annotations[AssignedIDsAnnotations]
	if !ok {
		return
	}
	podDev := decodePodDevices(ids)
	for _, val := range podDev {
		for _, deviceused := range val {
			if gs == nil {
				break
			}
			for index, gsdevice := range gs.Device {
				if gsdevice.UUID == deviceused.UUID {
					klog.V(4).Infoln("VGPU subsctracting pod", pod.Name, "device", deviceused)
					gs.Device[index].UsedMem -= uint(deviceused.Usedmem)
					gs.Device[index].UsedNum--
					gs.Device[index].UsedCore -= uint(deviceused.Usedcores)
				}
			}
		}
	}
	gs.GetStatus()
}

```

### DIFF
--- a/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
+++ b/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
@@ -176,6 +176,7 @@ func (gs *GPUDevices) SubResource(pod *v1.Pod) {
 			}
 		}
 	}
+	gs.GetStatus()
 }
 
 func (gs *GPUDevices) HasDeviceRequest(pod *v1.Pod) bool {


Monokaix:
`gs.GetStatus()` has get semantics but actually does an update operation，I think change to updateStatus is better，what do you think？ @archlitchi

googs1025:
> `gs.GetStatus()` has a get semantics but actually does an update operation，I think change to updateStatus is better，what do you think？ @archlitchi
+1
IMO, This method name is indeed very misleading. Does it need to be modified?

googs1025:
@archlitchi /PTAL

PR #3615, opened 2024-07-20
Add config field to the ControllerOption

liuyuanchun11:
Add config field to the ControllerOption, pass the config parameter to the controllers, and the controllers can use config to create a custom clientset

### DIFF
--- a/pkg/controllers/framework/interface.go
+++ b/pkg/controllers/framework/interface.go
@@ -19,6 +19,7 @@ package framework
 import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
 	vcclientset "volcano.sh/apis/pkg/client/clientset/versioned"
 	vcinformer "volcano.sh/apis/pkg/client/informers/externalversions"
@@ -37,6 +38,10 @@ type ControllerOption struct {
 	InheritOwnerAnnotations bool
 	WorkerThreadsForPG      uint32
 	WorkerThreadsForGC      uint32
+
+	// Config holds the common attributes that can be passed to a Kubernetes client
+	// and controllers registered by the users can use it.
+	Config *rest.Config
 }
 
 // Controller is the interface of all controllers.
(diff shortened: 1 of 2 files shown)


pkg/controllers/framework/interface.go
```diff
@@ -37,6 +38,8 @@ type ControllerOption struct {
 	InheritOwnerAnnotations bool
 	WorkerThreadsForPG      uint32
 	WorkerThreadsForGC      uint32
+
+	Config *rest.Config
```
googs1025, line 42: Why do we need this change?

googs1025: Do you have any plans or what the custom client needs?

Monokaix: I think we add some comments to describe why is this needed and what does it aim to do is better: ) @liuyuanchun11

liuyuanchun11: Sorry for not explaining clearly. I am maintaining a project based on the volcano community version, which contains some custom resource controllers and plug-ins, etc. Now I need to add a custom resource clientset, similar to VolcanoClient. The clientset of this custom resource is obviously not suitable to be placed in the ControllerOption structure, so I want to put rest.Config in the ControllerOption, so that it can be passed to my new controller, and then call NewForConfigOrDie(config of the custom resource in this controller ) to generate clientset

googs1025: Will the maintained features be fed back to the community? IMO, there is no need to introduce unused variables.
Just my opinion.If it is an overall feature, it feels more reasonable to introduce this variable

pkg/controllers/framework/interface.go
```diff
@@ -37,6 +38,9 @@ type ControllerOption struct {
 	InheritOwnerAnnotations bool
 	WorkerThreadsForPG      uint32
 	WorkerThreadsForGC      uint32
+
+	// Config holds the common attributes that can be passed to a Kubernetes client on initialization.
```
Monokaix, line 42: Kubernetes client on initialization. =》Kubernetes client and used by controllers registered by the user.

Issue #3618, opened 2024-07-23
Introduce a new interface `AddPreBindFn` or `AddBindFn` in session


Monokaix:
#### What would you like to be added:
Add a new interface `AddPreBindFn` or `AddBindFn` in  session to support more flexible bind.
#### Why is this needed:
Currently, volcano has supported kube-scheduler default pulgin for a long time, and plugins of kube-scheduler have an extension PreBind, which is called before a bind request, while volcano has no related extension, which needs modify the framework when we import kube-scheduler's plugin, there are two plugins that volcano has imported or is importing called `volumebinding` and `dynamicResourceAllocation`.

When import the`volumebinding` plugin of kube-scheduler，I found that it already initialized a VolumeBinder in schedulerCache, but actually it is only needed in plugin level，and the VolumeBinder should only exist in plugin, so there are some problems when supporting `volumeBinding`.  This pr https://github.com/volcano-sh/volcano/pull/2506 added the `volumeBinding` plugin, but is reverted because there is memory leak issue, detail https://github.com/volcano-sh/volcano/pull/2555, it is precisely because the logic of volumebinding is scattered everywhere and multiple caches are initialized that causes memory leaks，so if we can support a new interface，the logic of volumebinding will be more cohesive and avoid some potential problems.

Also, volcano is also supporting `dynamicResourceAllocation`(DRA), pr is here:https://github.com/volcano-sh/volcano/pull/3577, and the dra plugin also has an extension preBind, if there is no support in framework of preBind, then the logic of dra will also be scattered everywhere and codes will becomes unmaintainable.

Based on the above two considerations, we should support prebind extension.

googs1025:
Sounds great to me.
Will this be split into subtasks? Or will you take it?
If possible, I would like to participate in the work on this issue

Monokaix:
Task list:
- [ ] Add `PreBindFn` and `AddPreBindFn` func in session_plugins.go and call `PreBindFn` before call `Allocate` in allocate and backfill, and unallocate when error occurs.
- [ ]  Add volumebinding plugin in predict and move volumeBind related logic from schedulerCache to volumebinding plugin itself

Monokaix:
@googs1025 Hi task list has been added.

googs1025:
There is something I want to discuss. I saw the implementation of the k8s preBind interface as follows:
```go
// PreBindPlugin is an interface that must be implemented by "PreBind" plugins.
// These plugins are called before a pod being scheduled.
[...]

PR #3621, opened 2024-07-25
feature: add prebind for scheduler

googs1025:
/kind feature
- add prebindFn for scheduler

fix: https://github.com/volcano-sh/volcano/issues/3618 

part of issue: https://github.com/volcano-sh/volcano/issues/3618

### DIFF
--- a/pkg/scheduler/actions/backfill/backfill.go
+++ b/pkg/scheduler/actions/backfill/backfill.go
@@ -43,6 +43,7 @@ func (backfill *Action) Initialize() {}
 func (backfill *Action) Execute(ssn *framework.Session) {
 	klog.V(5).Infof("Enter Backfill ...")
 	defer klog.V(5).Infof("Leaving Backfill ...")
+	stmt := framework.NewStatement(ssn)
 
 	predicateFunc := func(task *api.TaskInfo, node *api.NodeInfo) ([]*api.Status, error) {
 		var statusSets api.StatusSets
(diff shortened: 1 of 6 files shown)


Monokaix:
It's not just simply add a prebind，we should also call unallocate to do the reverse in framework, you can refer to the kube-scheduler logic: )

googs1025:
> It's not just simply add a prebind，we should also call unallocate to do the reverse in framework, you can refer to the kube-scheduler logic: )

As far as I understand, can we split it into smaller PRs to complete different features? Because adding the prebind method does not seem to be related to the unallocate method.

googs1025:
> It's not just simply add a prebind，we should also call unallocate to do the reverse in framework, you can refer to the kube-scheduler logic: )

I am not sure if I need to implement a method similar to unreserve, because there is no process similar to reserve in the volcano scheduler process. Maybe pipeline is a process similar to reserve?

pkg/scheduler/framework/session.go
```diff
@@ -437,6 +438,41 @@ func (ssn *Session) Allocate(task *api.TaskInfo, nodeInfo *api.NodeInfo) (err er
 	return nil
 }
 
+// Unallocate is called when the prebind function fails
+func (ssn *Session) Unallocate(task *api.TaskInfo) error {
```
googs1025, line 442: The logic of the Unallocate method is to modify the cache in ssn
1. Change the task status to pending
2. Delete this task (Pod) in the node
3. send Deallocate event (I'm not sure if this step is necessary)
4. set task.Name = = ""

googs1025:
> It's not just simply add a prebind，we should also call unallocate to do the reverse in framework, you can refer to the kube-scheduler logic: )

added

pkg/scheduler/framework/session.go
lowang-bh, line 442: It exists in statement.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -91,6 +91,17 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 			}
 		}
 
+		// before allocating to a node, run prebind fn
```
lowang-bh, line 94: What will preBind be used for? And why don't put it in ssn.Allocate?

googs1025: issue detail: https://github.com/volcano-sh/volcano/issues/3618

googs1025: This needs @Monokaix  to help explain.
I think it is a method to expand to adapt to k8s plugins.

googs1025: According to my understanding, I also think that adding these to the main process is not very reasonable. Because k8s is always changing. We also have to keep changing, this is not a good design method

googs1025: Because k8s is always changing, I'm not sure if this incremental adaptation is a good design approach.

googs1025:
I don't have much time to complete this PR. I have discussed this with @JesseStutler offline, and this part has been transferred to him. But I will continue to follow this part cc @Monokaix

PR #3622, opened 2024-07-25
Optimize bind event update logic

wangyang0616:
1. Optimize the processing logic of Volcano bind interface, and uniformly handle event events and condition updates in the cache Bind interface.
2. The Bind interface in Binder only handles binding actions and does not involve event updates. It is more convenient to rewrite this capability when other projects reference the Volcano framework for Bind redevelopment.
3. Optimize error and processing logic.

### DIFF
--- a/pkg/scheduler/cache/interface.go
+++ b/pkg/scheduler/cache/interface.go
@@ -99,7 +99,7 @@ type VolumeBinder interface {
 
 // Binder interface for binding task and hostname
 type Binder interface {
-	Bind(kubeClient kubernetes.Interface, tasks []*api.TaskInfo) ([]*api.TaskInfo, []error)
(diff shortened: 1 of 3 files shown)


pkg/scheduler/cache/interface.go
```diff
@@ -99,7 +99,7 @@ type VolumeBinder interface {
 
 // Binder interface for binding task and hostname
 type Binder interface {
-	Bind(kubeClient kubernetes.Interface, tasks []*api.TaskInfo) ([]*api.TaskInfo, []error)
```
william-wang, line 102: @wangyang0616 Do we have to change the interface? I am afraid of the compatibility.
Is there any better way to solve this problem?

lowang-bh: This interface was first broken in https://github.com/volcano-sh/volcano/pull/3547/files.

Issue #3623, opened 2024-07-25
E2E Spark Integration Test always failed


googs1025:
**What happened**:
- i found that E2E Spark Integration Test always failed, and the error is always the same

output log
```bash
------
Dockerfile:27
--------------------
  26 |     RUN mkdir ${SPARK_HOME}/python
  27 | >>> RUN apt-get update && \
  28 | >>>     apt install -y python3 python3-pip && \
  29 | >>>     pip3 install --upgrade pip setuptools && \
  30 | >>>     # Removed the .cache to save space
  31 | >>>     rm -rf /root/.cache && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/*
  32 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c apt-get update &&     apt install -y python3 python3-pip &&     pip3 install --upgrade pip setuptools &&     rm -rf /root/.cache && rm -rf /var/cache/apt/* && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 1
Failed to build PySpark Docker image, please refer to Docker build output for details.
[error] java.lang.IllegalStateException: Process '/home/runner/work/volcano/volcano/spark/bin/docker-image-tool.sh -r docker.io/kubespark -t dev -p /home/runner/work/volcano/volcano/spark/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/bindings/python/Dockerfile -R /home/runner/work/volcano/volcano/spark/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/bindings/R/Dockerfile -f /home/runner/work/volcano/volcano/spark/resource-managers/kubernetes/docker/src/main/dockerfiles/spark/Dockerfile build' exited with 1.
[error] 	at KubernetesIntegrationTests$.$anonfun$settings$67(SparkBuild.scala:1050)
[error] 	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[error] 	at sbt.std.Transform$$anon$3.$anonfun$apply$2(Transform.scala:46)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:68)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
[error] 	at sbt.Execute.work(Execute.scala:291)
[...]

PR #3556, opened 2024-07-02
Volcano adapts to the k8s v1.30

guoqinwill:
Fix https://github.com/volcano-sh/volcano/issues/3523
cherry pick and update：https://github.com/volcano-sh/volcano/pull/3534/commits

PR #3625, opened 2024-07-26
Fix spark ci error

Monokaix:
When install pip3 and pip using Ubuntu24.04，debain package managed by pip3 and python packages managed by pip have conflict，and ubuntu20.04 has no such problem，it's also consistent with upstream spark https://github.com/Monokaix/spark/blob/2962796ca3a61b7d0a57e9280bcee1e77e75fcec/.github/workflows/build_and_test.yml#L891
Fix https://github.com/volcano-sh/volcano/issues/3623

### DIFF
--- a/.github/workflows/e2e_spark.yaml
+++ b/.github/workflows/e2e_spark.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   k8s-integration-tests:
     name: "E2E about Spark Integration test"
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-20.04
     steps:
 
     - name: Checkout current Volcano repository


william-wang (approved):
/approve

.github/workflows/e2e_spark.yaml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   k8s-integration-tests:
     name: "E2E about Spark Integration test"
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-20.04
```
wangyang0616, line 13: Does “ubuntu-22.04” not have this problem?
([#3556](https://github.com/volcano-sh/volcano/pull/3556) has been using ubuntu-22.04 before the change)

Monokaix: spark related e2e is independent and use minikube as e2e tool，so it should be just consistent with upstream spark's e2e config.

PR #3626, opened 2024-07-26
Improve ginkgo e2e case of preempt

Monokaix:
- Use Aftereach insted of defer, becasue if err occured, defer will not be executed.
- Use different job names in different test cases to distinguish them apart.


Monokaix:
Other e2e case can also be improved, becase preempt is a flaky test so we solve this first.

hwdef (commented):
/lgtm

PR #3628, opened 2024-07-27
[cherry-pick for release-1.10] Optimize bind event update logic 

wangyang0616:
cherry pick from [#3622 ](https://github.com/volcano-sh/volcano/pull/3622)

1. Optimize the processing logic of Volcano bind interface, and uniformly handle event events and condition updates in the cache Bind interface.
2. The Bind interface in Binder only handles binding actions and does not involve event updates. It is more convenient to rewrite this capability when other projects reference the Volcano framework for Bind redevelopment.
3. Optimize error and processing logic.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -173,9 +173,8 @@ type DefaultBinder struct {
 }
 
 // Bind will send bind request to api server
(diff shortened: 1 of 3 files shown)


william-wang (approved):
/approve

Issue #2493, opened 2022-09-13
metrics exposure for volcano job succeeded state and failed state


Saba1119:
#### What would you like to be added:
currently there is no metrics for **volcano jobs succeeded state and failed state** ,only volcano podgroup pending,inqueue,running and unknown state is available.those metrics are

volcano_queue_pod_group_inqueue_count
volcano_queue_pod_group_pending_count
volcano_queue_pod_group_running_count
volcano_queue_pod_group_unknown_count

#### Why is this needed:

if this metrics are available in volcano then it will be useful for visualization in grafana dashboard

lppsuixn:
need this.

hwdef:
still need

DsLipku:
still need

Prepmachine4:
Can I try it as my first contribution to the volcana?

Monokaix:
> Can I try it as my first contribution to the volcana?

Yeah，you can try it.

Prepmachine4:
@Monokaix Please  review the code, thank you.

Prepmachine4:
@Monokaix Please review the code again, thank you.

PR #3636, opened 2024-07-28
feat: add volcano jobs succeeded metric

Prepmachine4:
fix: #2493
I think the meaning of jobRetryCount is the same as that of fail count.

```go
jobRetryCount = promauto.NewCounterVec(
  prometheus.CounterOpts{
	  Subsystem: VolcanoNamespace,
	  Name:      "job_retry_counts",
	  Help:      "Number of retry counts for one job",
  }, []string{"job_id"},
)
```

### DIFF
--- a/cmd/controller-manager/app/server.go
+++ b/cmd/controller-manager/app/server.go
@@ -127,6 +127,7 @@ func startControllers(config *rest.Config, opt *options.ServerOption) func(ctx c
 	controllerOpt.InheritOwnerAnnotations = opt.InheritOwnerAnnotations
 	controllerOpt.WorkerThreadsForPG = opt.WorkerThreadsForPG
 	controllerOpt.WorkerThreadsForGC = opt.WorkerThreadsForGC
+	controllerOpt.Config = config
 
 	return func(ctx context.Context) {
 		framework.ForeachController(func(c framework.Controller) {
--- a/pkg/controllers/framework/interface.go
+++ b/pkg/controllers/framework/interface.go
@@ -19,6 +19,7 @@ package framework
 import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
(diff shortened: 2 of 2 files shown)


pkg/scheduler/plugins/gang/gang.go
```diff
@@ -223,6 +223,8 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 				klog.Errorf("Failed to update job <%s/%s> condition: %v",
 					job.Namespace, job.Name, err)
 			}
+
+			metrics.RegisterJobSuccess(job.Name)
```
googs1025, line 227: I am a little curious, why do we add indicators here? According to my understanding, should the indicators be added after the job runs successfully?

Prepmachine4: https://github.com/volcano-sh/volcano/issues/2493#issue-1371690108
Maybe I have misunderstood?

googs1025:
use `git commit -s`  to handle DCO failed

pkg/scheduler/plugins/gang/gang.go
googs1025: IMO. The metrics of success or failure should be determined when the controller state changes. The scheduler is only used to schedule to a certain node and does not care whether the pod runs successfully.

Monokaix: Yeah maybe we should set them in job controller.

googs1025:
I think the concept of restart count is different from that of job failed. Job failed should be the state of failure after exceeding the restart count. Restart count is the process of recording the intermediate state of the job.

googs1025:
Could you also check if jobRetryCount is set correctly?

Prepmachine4:
> I think the concept of restart count is different from that of job failed. Job failed should be the state of failure after exceeding the restart count. Restart count is the process of recording the intermediate state of the job.

Yes, I have misunderstand before.I am reading the **job_controller** code now to decide how to add the metrics.

PR #3637, opened 2024-07-28
Fix #3604: Include Aborted Jobs in Garbage Collector Cleanup

### LINKED ISSUE #3604: Garbage Collector is supposed to clean up Aborted volcanojobs
**What happened**: garbagecollector controller does not clean up any Aborted volcanojobs where the `.status.state.phase` is `Aborted`.

**What you expected to happen**: Aborted jobs should be cleaned up.

1. create a volcanojob with `.spec.ttlSecondsAfterFinished` specified.
2. abort this job.
3. `kubectl get vj xxx`, you will see that thee job's `.status.state.phase` is `Aborted`.
4. wait for `ttlSecondsAfterFinished` seconds.
5. however, this job still exists.

**Anything else we need to know?**: The bug seems to occur in the following code. https://github.com/volcano-sh/volcano/blob/v1.10.0-alpha.0/pkg/controllers/garbagecollector/garbagecollector.go#L266

```go
func isJobFinished(job *v1alpha1.Job) bool {
	return job.Status.State.Phase == v1alpha1.Completed ||
		job.Status.State.Phase == v1alpha1.Failed ||
		job.Status.State.Phase == v1alpha1.Terminated
}
```

The condition `job.Status.State.Phase == v1alpha1.Aborted` should not be omitted.

**Environment**:
- Volcano Version: 1.4 to 1.9, up to now

kyooosukedn:
Changes:

 Modified the isJobFinished function in pkg/controllers/garbagecollector/garbagecollector.go to include the Aborted state.

### DIFF
--- a/pkg/controllers/garbagecollector/garbagecollector.go
+++ b/pkg/controllers/garbagecollector/garbagecollector.go
@@ -273,6 +273,7 @@ func isJobFinished(job *v1alpha1.Job) bool {
 	return job.Status.State.Phase == v1alpha1.Completed ||
 		job.Status.State.Phase == v1alpha1.Failed ||
 		job.Status.State.Phase == v1alpha1.Terminated
+		job.Status.State.Phase == v1alpha1.Aborted
 }
 
 func getFinishAndExpireTime(j *v1alpha1.Job) (*time.Time, *time.Time, error) {


googs1025:
use `git commit -s ` to handle DCO failed

pkg/controllers/garbagecollector/garbagecollector.go
```diff
@@ -273,6 +273,7 @@ func isJobFinished(job *v1alpha1.Job) bool {
 	return job.Status.State.Phase == v1alpha1.Completed ||
 		job.Status.State.Phase == v1alpha1.Failed ||
 		job.Status.State.Phase == v1alpha1.Terminated
+		job.Status.State.Phase == v1alpha1.Aborted
```
googs1025, line 276: I have a small question: How do we distinguish between "Aborted", "Terminated" and "Failed"? Are these three states the final states of a job?

Monokaix: That's a good question, `Aborted` is triggered by user, and user can set `Aborted` action in job spec.policy and pods under the pod will be evicted and job can be resumed,  `Terminated` is also set by user but from the code comments that job under this state can not be resumed, while `Failed` is  a state that pods under a job execution failed.
I'm not sure whether whethe a job under `Aborted` state can be resumed, if can be resumed, we should re-consider whether the pr is reasonable: ）

googs1025: Yes, this is critical, which determines whether it is the final state. If not, it should not be recycled.

Monokaix: I think this can be close: )

Monokaix:
Thanks for your contribution! You can also find other issues suitable for new users.

PR #3638, opened 2024-07-29
fix the bind task logic 

googs1025:
/kind cleanup
/kind bug  

I don't know if it's a bug, but I found that there's something little wrong with the logic in bind.

```go
func (sc *SchedulerCache) processBindTask() {
	for {
		select {
		case taskInfo, ok := <-sc.BindFlowChannel:
			if !ok {
				return
			}

			sc.bindCache = append(sc.bindCache, taskInfo)
			if len(sc.bindCache) == sc.batchNum {
				sc.BindTask()
			}
		default:
		}

		if len(sc.BindFlowChannel) == 0 {
			break
		}
	}

	if len(sc.bindCache) == 0 {
		return
	}
	sc.BindTask()
}
```

I think that in the loop, when `len(sc.BindFlowChannel) == 0`, we should not break out of the loop, but should exit when the channel is closed.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1236,29 +1236,33 @@ func (sc *SchedulerCache) AddBindTask(taskInfo *schedulingapi.TaskInfo) error {
 }
 
 func (sc *SchedulerCache) processBindTask() {
+	closed := false
 	for {
 		select {
 		case taskInfo, ok := <-sc.BindFlowChannel:
+			// the chan has been closed,
 			if !ok {
-				return
-			}
-
-			sc.bindCache = append(sc.bindCache, taskInfo)
-			if len(sc.bindCache) == sc.batchNum {
-				sc.BindTask()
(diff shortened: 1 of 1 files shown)


lowang-bh:
/hold

You have writen a dead-loop.  And `processBindTask` is runned period and should return when where are no tasks need to be bound.

pkg/scheduler/cache/cache.go
```diff
@@ -1236,29 +1236,33 @@ func (sc *SchedulerCache) AddBindTask(taskInfo *schedulingapi.TaskInfo) error {
 }
 
 func (sc *SchedulerCache) processBindTask() {
+	closed := false
 	for {
 		select {
 		case taskInfo, ok := <-sc.BindFlowChannel:
+			// the chan has been closed,
```
Monokaix, line 1243: If the chan has been closed, should we still bind the task in cache?

lowang-bh:
I think it can be closed.

PR #3644, opened 2024-07-31
update issue template

hwdef:
Use yaml replace markdown, it can provide more styles.
Delete some useless and duplicate questions, such as 
`What would you like to be added`
`Why is this needed:`
replaced by 
`What is the problem you're trying to solve`
`Describe the solution you'd like`
`Additional context`

The modification plan refers to [containerd](https://github.com/containerd/containerd/issues/new/choose).


googs1025:
can we so add a `Support Request` kind template?

googs1025:
> can we so add a `Support Request` kind template?

I'm not sure if question is of type support

hwdef:
> can we so add a `Support Request` kind template?

I think this duplicates the functionality of enhancement and question

Monokaix:
A little doubt thst is yaml format more user friendly than markdown?

hwdef:
> A little doubt thst is yaml format more user friendly than markdown?

I think so. 

The following is a comparison between yaml and markdown. 

Users can also enter markdown in the input box

image

image

Monokaix:
That's great!

Monokaix:
cc @lowang-bh

Issue #3635, opened 2024-07-27
docs: overcommit-plugin enhancements


googs1025:
#### What would you like to be added:

We can further break down the overcommit-factor into more granular components: cpu-overcommit-factor, mem-overcommit-factor, and other-overcommit-factor.
To maintain compatibility with the existing approach, we will retain the original overcommit-factor field and introduce optional fields for cpu-overcommit-factor, mem-overcommit-factor, and other-overcommit-factor.
#### Why is this needed:

However, different resources should have different factors, so using the same overcommit-factor is not appropriate.

PR #3647, opened 2024-08-01
feature: use multi factor for overcommit plugin

googs1025:
#### What type of PR is this?
/kind feature

- different resources should have different factors, so using the same overcommit-factor is not appropriate.

https://github.com/volcano-sh/volcano/issues/3635

None
User could give different overcommit factors (cpu-overcommit-factor, mem-overcommit-factor, other-overcommit-factor, overcommit-factor) through overcommit plugin arguments

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -296,6 +296,35 @@ func (r *Resource) Multi(ratio float64) *Resource {
 	return r
 }
 
+// ScaleResourcesWithRatios multiples the resource with ratio provided for CPU, Memory, and ScalarResources
+// if ratio is nil, use default ratio
+func (r *Resource) ScaleResourcesWithRatios(ratio map[string]float64, defaultRatio float64) *Resource {
+	// if ratio is nil, use default ratio
+	if ratio == nil {
+		return r.Multi(defaultRatio)
+	}
+
+	// apply ratio to CPU, Memory, and ScalarResources
+	// if ratio is provided, use ratio if exists,
+	// otherwise use default ratio
+	applyRatio := func(resource *float64, resourceName string) {
+		if ratioValue, ok := ratio[resourceName]; ok {
+			*resource *= ratioValue
+		} else {
+			*resource *= defaultRatio
+		}
+	}
+
+	applyRatio(&r.MilliCPU, string(v1.ResourceCPU))
+	applyRatio(&r.Memory, string(v1.ResourceMemory))
+
+	for rName, rQuant := range r.ScalarResources {
+		applyRatio(&rQuant, string(rName))
+		r.ScalarResources[rName] = rQuant
+	}
+	return r
+}
+
 // SetMaxResource compares with ResourceList and takes max value for each Resource.
 func (r *Resource) SetMaxResource(rr *Resource) {
 	if r == nil || rr == nil {
--- a/pkg/scheduler/plugins/overcommit/overcommit.go
+++ b/pkg/scheduler/plugins/overcommit/overcommit.go
@@ -17,6 +17,8 @@ limitations under the License.
 package overcommit
 
 import (
+	"strings"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
@@ -32,28 +34,51 @@ const (
 	// overCommitFactor is resource overCommit factor for enqueue action
 	// It determines the number of `pending` pods that the scheduler will tolerate
 	// when the resources of the cluster is insufficient
+	// This field is used as the default key in factorMaps
 	overCommitFactor = "overcommit-factor"
 	// defaultOverCommitFactor defines the default overCommit resource factor for enqueue action
 	defaultOverCommitFactor = 1.2
 )
 
+const (
+	// overCommitFactorPrefix is the prefix of resource overCommit factor
+	// We use this prefix to segment the rules for custom resources
+	// in the configuration file.
+	overCommitFactorPrefix = "overcommit-factor."
+)
+
+// overcommitFactors defines the resource overCommit factors
+type overcommitFactors struct {
+	// factorMaps defines the resource overCommit factors
+	// key: resource, example: "cpu", "memory", "ephemeral-storage", "nvidia.com/gpu"
+	// value: overCommit factors
+	// when initializing, we will store a default value into this map
+	// key: "overcommit-factor", value: defaultOverCommitFactor
+	factorMaps map[string]float64
+}
+
 type overcommitPlugin struct {
-	// Arguments given for the plugin
-	pluginArguments  framework.Arguments
-	totalResource    *api.Resource
-	idleResource     *api.Resource
-	inqueueResource  *api.Resource
-	overCommitFactor float64
+	// pluginArguments Arguments given for the plugin
+	pluginArguments framework.Arguments
+	totalResource   *api.Resource
+	idleResource    *api.Resource
+	inqueueResource *api.Resource
+	// overCommitFactor is the different resource overCommit factors
+	overCommitFactors *overcommitFactors
 }
 
 // New function returns overcommit plugin object
 func New(arguments framework.Arguments) framework.Plugin {
 	return &overcommitPlugin{
-		pluginArguments:  arguments,
-		totalResource:    api.EmptyResource(),
-		idleResource:     api.EmptyResource(),
-		inqueueResource:  api.EmptyResource(),
-		overCommitFactor: defaultOverCommitFactor,
+		pluginArguments: arguments,
+		totalResource:   api.EmptyResource(),
+		idleResource:    api.EmptyResource(),
+		inqueueResource: api.EmptyResource(),
+		overCommitFactors: &overcommitFactors{
+			factorMaps: map[string]float64{
+				overCommitFactor: defaultOverCommitFactor,
+			},
+		},
 	}
 }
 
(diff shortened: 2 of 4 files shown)


pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -144,3 +218,42 @@ func (op *overcommitPlugin) OnSessionClose(ssn *framework.Session) {
...
 	op.inqueueResource = nil
 }
+
+// parseFactor is used to parse the value corresponding to the given key and store the result
+// in the float64 variable pointed to by target.
+// If the specified key exists in pluginArguments, the corresponding value is stored in target;
+// if the specified key does not exist but the default overCommitFactor does,
+// its corresponding value is stored in target.
+func (op *overcommitPlugin) parseFactor(key string, target *float64) {
+	if op.pluginArguments.KeyExists(key) {
```
lowang-bh, line 228: Why need check exist? Get method has handled it。

pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -62,33 +83,86 @@ func (op *overcommitPlugin) Name() string {
...
 tiers:
 - plugins:
   - name: overcommit
     arguments:
+    cpu-overcommit-factor: 1.2
+    mem-overcommit-factor: 1.0
+    other-overcommit-factor: 1.2
+
+Example 2:
+Specifying only the overcommit-factor implies that all three factors are the same.
```
lowang-bh, line 104: Full Docs dont need to be put here. Just an example.

pkg/scheduler/framework/arguments.go
```diff
@@ -99,3 +99,9 @@ func GetArgOfActionFromConf(configurations []conf.Configuration, actionName stri
 
 	return nil
 }
+
+// KeyExists checks if the key exists in the Arguments map
+func (a Arguments) KeyExists(key string) bool {
```
lowang-bh, line 104: it is redundant.

pkg/scheduler/api/resource_info.go
```diff
@@ -296,6 +296,16 @@ func (r *Resource) Multi(ratio float64) *Resource {
 	return r
 }
 
+// ScaleResourcesWithRatios multiples the resource with ratio provided for CPU, Memory, and ScalarResources
+func (r *Resource) ScaleResourcesWithRatios(cpuRatio float64, memRatio float64, scalarRatio float64) *Resource {
+	r.MilliCPU *= cpuRatio
+	r.Memory *= memRatio
+	for rName, rQuant := range r.ScalarResources {
```
lowang-bh, line 303: What if different scalar has its own ratio? And scalar resource support overcommit?

pkg/scheduler/plugins/overcommit/overcommit.go
googs1025: My idea at the time was that we need to know clearly whether this key is used in the configuration, so that we can judge the subsequent logic.
If we only use the original method, I don't know whether this value is the default value or the configuration file setting

pkg/scheduler/framework/arguments.go
googs1025: ditto

pkg/scheduler/api/resource_info.go
googs1025: If super-resolution is not supported, we can set it to 1 through the configuration file. Users can freely use the configuration file to set the required value

googs1025: This is also an improvement compared to the previous

pkg/scheduler/plugins/overcommit/overcommit.go
googs1025: The priority of these fields will be from low to high:

defaultOverCommitFactor -> overcommit-factor -> cpu-overcommit-factor, men-overcommit-factor, other overcommit-factor

pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -37,23 +37,44 @@ const (
 	defaultOverCommitFactor = 1.2
 )
 
+const (
+	// cpuOverCommitFactor is resource overCommit factor for cpu
```
lowang-bh, line 41: Use `overcommit-factor.cpu`, `overcommit-factor.memory`. If there are other scalar resource, use `overcommit-factor.ephemeral-storage`,  `overcommit-factor.pods`, etc.

so the key can be cpu/memory/ephemeral-storage/pods, even nvidia.com/gpu.

lowang-bh:
Please refer to predicate.Proportional function in predicate plugin.

pkg/scheduler/framework/arguments.go
lowang-bh: Dont need. Please refer to predicate.Proportional plugin.

pkg/scheduler/plugins/overcommit/overcommit.go
googs1025: I understand. This change is reasonable.

pkg/scheduler/framework/arguments.go
googs1025: I think using this method makes the code easier to read for others. It also helps keep the code cleaner. And I think the `KeyExists` method is common and necessary in the configuration file. Other callers can also use this method.

pkg/scheduler/plugins/overcommit/overcommit.go
googs1025: I want to change it to: "overcommit-factor.cpu", "overcommit-factor.memory" "overcommit-factor.ephemeral-storage" and "overcommit-factor.other". The advantage of this change is that I think the number of other pods should not be increased. In addition, other external resources are not allowed to be increased in most scenarios, so we only leave "overcommit-factor.other" for users to set.

lowang-bh: You haven't get the idea. please think more about extendable.

pkg/scheduler/plugins/overcommit/overcommit.go
googs1025: done

pkg/scheduler/plugins/overcommit/overcommit.go
googs1025: fixed

pkg/scheduler/api/resource_info.go
googs1025: fixed

pkg/scheduler/api/resource_info.go
```diff
@@ -296,6 +296,38 @@ func (r *Resource) Multi(ratio float64) *Resource {
 	return r
 }
 
+// ScaleResourcesWithRatios multiples the resource with ratio provided for CPU, Memory, and ScalarResources
+func (r *Resource) ScaleResourcesWithRatios(ratio map[string]float64, defaultRatio float64) *Resource {
+	if ratio == nil {
+		return r
+	}
+	// This value is defined in the overcommit plugin
+	generalFactorKey := "overcommit-factor"
```
lowang-bh, line 305: Why depend on it?  ratio map just need key like cpu/mem, etc.

pkg/scheduler/api/resource_info.go
```diff
@@ -296,6 +296,38 @@ func (r *Resource) Multi(ratio float64) *Resource {
 	return r
 }
 
+// ScaleResourcesWithRatios multiples the resource with ratio provided for CPU, Memory, and ScalarResources
+func (r *Resource) ScaleResourcesWithRatios(ratio map[string]float64, defaultRatio float64) *Resource {
+	if ratio == nil {
+		return r
```
lowang-bh, line 302:
```suggestion
		return r.Multi(defaultRatio)
```

pkg/scheduler/plugins/overcommit/overcommit.go
```diff
@@ -37,23 +39,39 @@ const (
...
+	// overCommitFactorPrefix is the prefix of resource overCommit factor
+	overCommitFactorPrefix = "overcommit-factor."
+)
+
+// overcommitFactors defines the resource overCommit factors
+type overcommitFactors struct {
+	// factorMaps defines the resource overCommit factors
+	// key: resource, example: "cpu", "memory", "ephemeral-storage", "nvidia.com/gpu"
+	// value: overCommit factors
+	factorMaps map[string]float64
```
lowang-bh, line 52: 1. need a defaultFactor to store the configured value in overcommit-factor. If it is set by user, updating to it, otherwise default is `defaultOverCommitFactor`.
2. please add ut to cover your parse function.

googs1025: Thanks, that's a good idea.

pkg/scheduler/api/resource_info.go
googs1025: thanks for catching that

pkg/scheduler/api/resource_info.go
googs1025: fixex

pkg/scheduler/api/resource_info.go
```diff
@@ -296,6 +296,34 @@ func (r *Resource) Multi(ratio float64) *Resource {
 	return r
 }
 
+// ScaleResourcesWithRatios multiples the resource with ratio provided for CPU, Memory, and ScalarResources
+func (r *Resource) ScaleResourcesWithRatios(ratio map[string]float64, defaultRatioKey string) *Resource {
```
lowang-bh, line 300: default should be a float.

googs1025: The default value itself is also a float, and the key of the default value is passed in.

lowang-bh: So there is no improvement now？

lowang-bh: It still has dependence on other context. This func should be clean.

googs1025: fixed

PR #2985, opened 2023-07-21
use one helm command to install resource

lowang-bh:
use one command to install helm resource with sub charts, there are mainly two benefit:
1. more friendly to end user
2. smooth upgrade volcano for other projects which use helm to install volcano, such as kubeflow

In the future, if there are more subcharts, just 

1. add sub char dir under `installer/helm/chart/volcano/charts/`
2. add dependency in `installer/helm/chart/volcano/Chart.yaml`

PR #3648, opened 2024-08-01
Use helm package flag to set version in generate chart release

Monokaix:
Curently volcano use `sed` to replace helm chart version, but it's easy to make mistakes and make release charts failed，this pr https://github.com/volcano-sh/volcano/pull/2985 changed the version format and leads `sed` match faliled.

While `helm package` itself has supports flag to set version, so we should use the official command to release charts and avoid mistakes.


Monokaix:
charts release CI has already failed currently.
image

Monokaix:
cc @wangyang0616 @lowang-bh

lowang-bh:
:thumbsup:
/lgtm

PR #3649, opened 2024-08-01
don't enable error cache if task role spec is empty

### LINKED ISSUE #3527: `PredicateNodes` function default enable ErrorCache, and offen disturb UT case
The `errorCache` feature can be used when job's tasks have a same spec key to improve proformance.
However, if the spec key is empty for some tasks in a job and predicate failed, then those tasks with empty spec key in this job will all be failed in predicate action.
`taskGroupID`
```go
func taskGroupID(task *api.TaskInfo) string {
	return fmt.Sprintf("%s/%s", task.Job, task.GetTaskSpecKey())
}
```

Two solotions:
- 1. add a config param to control the switch of err cache
- 2. if task spec key is empty, don't use error cache

#### Why is this needed:

1. save debug time when write testcases, It has disturb me several times that pods can not work as expect
2. in some user scene using volcano scheduler without controller, this may cause job cannot be scheduled if user forget to set spec

### LINKED ISSUE #3666: When the Gang policy(JobReadyFn) is enabled，cache in the  Predicate phase of the scheduling cycle causes a scheduling failure
### Description

The tasks submitted will have the following requirements
**1. Pods in a Job are not the same, perhaps with different resource requirements, perhaps with different node affinities, etc
2. the **Gang** policy(JobReadyFn) is enabled**

**In this case, all pods in a Job will use the same Predicate cache. When a Pod fails to **Predicate**  a node, that node will never be considered.This may cause the task to fail.**

I think you can't assume that the Pods under the Job are all the same, so you shouldn't use the same cache, and can the Predicate cache switch be configured?

attention: I used a third-party platform to submit the Job, not Volcano Job

1.
2.
3.

I think you can't assume that the Pods under the Job are all the same, so you shouldn't use the same cache, and can the Predicate cache switch be configured?

release v1.9.0

_No response_

lowang-bh:
fixes #3527
fixes #3666

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -31,10 +31,14 @@ import (
 
 type Action struct {
 	session *framework.Session
+	// configured flag for error cache
+	enablePredicateErrorCache bool
 }
 
 func New() *Action {
-	return &Action{}
+	return &Action{
+		enablePredicateErrorCache: true, // default to enable it
+	}
 }
 
 func (alloc *Action) Name() string {
@@ -43,9 +47,16 @@ func (alloc *Action) Name() string {
 
 func (alloc *Action) Initialize() {}
 
+func (alloc *Action) parseArguments(ssn *framework.Session) {
+	arguments := framework.GetArgOfActionFromConf(ssn.Configurations, alloc.Name())
+	arguments.GetBool(&alloc.enablePredicateErrorCache, conf.EnablePredicateErrCacheKey)
+}
+
 func (alloc *Action) Execute(ssn *framework.Session) {
 	klog.V(5).Infof("Enter Allocate ...")
 	defer klog.V(5).Infof("Leaving Allocate ...")
+
+	alloc.parseArguments(ssn)
 
 	// the allocation for pod may have many stages
 	// 1. pick a queue named Q (using ssn.QueueOrderFn)
@@ -199,7 +210,7 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 			break
 		}
 
-		predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, alloc.predicate, true)
+		predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, alloc.predicate, alloc.enablePredicateErrorCache)
 		if len(predicateNodes) == 0 {
 			job.NodesFitErrors[task.UID] = fitErrors
 			// Assume that all left tasks are allocatable, but can not meet gang-scheduling min member,
--- a/pkg/scheduler/actions/backfill/backfill.go
+++ b/pkg/scheduler/actions/backfill/backfill.go
@@ -22,15 +22,20 @@ import (
 	"k8s.io/klog/v2"
 
 	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
-type Action struct{}
+type Action struct {
+	enablePredicateErrorCache bool
+}
 
 func New() *Action {
-	return &Action{}
+	return &Action{
+		enablePredicateErrorCache: true, // default to enable it
+	}
 }
 
 func (backfill *Action) Name() string {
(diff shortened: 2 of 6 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -199,7 +199,10 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 			break
 		}
 
-		predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, alloc.predicate, true)
+		// don't enable error cache if task's TaskRole is empty, because different pods with empty TaskRole will all
+		// have the same taskGroupID, and one pod predicate failed, all other pods will also be failed
+		// see issue: https://github.com/volcano-sh/volcano/issues/3527
+		predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, alloc.predicate, len(task.TaskRole) != 0)
```
Monokaix, line 205: So where to add the switch?

Monokaix, line 205: In what case the task.TaskRole is empty?

Monokaix:
What's difference between this and `cacheEnable` in predict plugin?

pkg/scheduler/actions/allocate/allocate.go
lowang-bh: Current implementation don't need switch anymore. It will be disabled when task role is empty,other wise enable it.

lowang-bh: Eg, testcase forget to set it.  Workloads doesn't set it in pod when just scheduler is used.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -76,7 +76,7 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 			break
 		}
 
-		predicateNodes, fitErrors := ph.PredicateNodes(task, ssn.NodeList, predicateFunc, true)
+		predicateNodes, fitErrors := ph.PredicateNodes(task, ssn.NodeList, predicateFunc, len(task.TaskRole) != 0)
```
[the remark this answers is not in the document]
lowang-bh: It is also ok.

pkg/scheduler/actions/allocate/allocate.go
Monokaix: All none vcjob will meet this problem？

pkg/scheduler/actions/backfill/backfill.go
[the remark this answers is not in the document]
Monokaix: I think we can add a switch, becasue we can compare the different performance with this feature enable/disabled.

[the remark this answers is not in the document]
lowang-bh: Add a switch for cache, but still keep disabled for empty taskrole, right?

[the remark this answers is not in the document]
Monokaix: Yeah.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -199,7 +210,10 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 			break
 		}
 
-		predicateNodes, fitErrors := ph.PredicateNodes(task, allNodes, alloc.predicate, true)
+		// don't enable error cache if task's TaskRole is empty, because different pods with empty TaskRole will all
```
Monokaix, line 213: Put these comments into pkg/scheduler/util/predicate_helper.go is better.

pkg/scheduler/api/well_known_labels.go
```diff
@@ -42,4 +42,7 @@ const (
 
 	// topologyDecisionAnnotation is the key of topology decision about pod request resource
 	topologyDecisionAnnotation = "volcano.sh/topology-decision"
+
```
Monokaix, line 45: Can we put it into another file? Because it's not a label.

pkg/scheduler/api/constants.go
```diff
@@ -0,0 +1,6 @@
+package api
+
+const (
+	// EnablePredicateErrCacheKey is the key whether predicate error cache is enabled
+	EnablePredicateErrCacheKey = "enable.predicateErrorCache"
```
Monokaix, line 5: Can we refer to plugin's option definition？Because it's also an option or param of action.

lowang-bh: such as `allocate.predicateErrorCacheEnable` or `predicateErrorCacheEnable`?

Monokaix: Yeah.

lowang-bh: Has changed to `predicateErrorCacheEnable`.

pkg/scheduler/api/constants.go
```diff
@@ -0,0 +1,6 @@
+package api
+
+const (
+	// EnablePredicateErrCacheKey is the key whether predicate error cache is enabled
+	EnablePredicateErrCacheKey = "predicateErrorCacheEnable"
```
Monokaix, line 5: A little doubt, seems it's just a param and not an API, can we put it in another place?

hwdef: +1

lowang-bh: how about put them in `pkg/scheduler/conf/scheduler_conf.go`?

hwdef: I think this will be fine

PR #3650, opened 2024-08-03
feat: add volcano jobs phase metric

Prepmachine4:
fix: #2493 

I add all job phase shows in job apis.
```go
const (
	// Pending is the phase that job is pending in the queue, waiting for scheduling decision
	Pending JobPhase = "Pending"
	// Aborting is the phase that job is aborted, waiting for releasing pods
	Aborting JobPhase = "Aborting"
	// Aborted is the phase that job is aborted by user or error handling
	Aborted JobPhase = "Aborted"
	// Running is the phase that minimal available tasks of Job are running
	Running JobPhase = "Running"
	// Restarting is the phase that the Job is restarted, waiting for pod releasing and recreating
	Restarting JobPhase = "Restarting"
	// Completing is the phase that required tasks of job are completed, job starts to clean up
	Completing JobPhase = "Completing"
	// Completed is the phase that all tasks of Job are completed
	Completed JobPhase = "Completed"
	// Terminating is the phase that the Job is terminated, waiting for releasing pods
	Terminating JobPhase = "Terminating"
	// Terminated is the phase that the job is finished unexpected, e.g. events
	Terminated JobPhase = "Terminated"
	// Failed is the phase that the job is restarted failed reached the maximum number of retries.
	Failed JobPhase = "Failed"
)
```

The metric record event happend in `jobInformer.Informer()` received event.
```go
cc.jobInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
	AddFunc:    cc.addJob,
	UpdateFunc: cc.updateJob,
	DeleteFunc: cc.deleteJob,
})
```
And the processing of record metrics is not use `cache lock` that maybe produces some inaccurate data but will improve some performance.

But if it traverses all jobs every time during updates, I think there might be some trouble. Should we switch to incremental updates or scheduled metrics at intervals?

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -37,6 +37,7 @@ const (
 	defaultMaxRequeueNum       = 15
 	defaultSchedulerName       = "volcano"
 	defaultHealthzAddress      = ":11251"
+	defaultListenAddress       = ":8081"
 	defaultLockObjectNamespace = "volcano-system"
 	defaultPodGroupWorkers     = 5
 	defaultGCWorkers           = 1
@@ -70,6 +71,8 @@ type ServerOption struct {
 	// defaulting to 0.0.0.0:11251
 	HealthzBindAddress string
 	EnableHealthz      bool
+	EnableMetrics      bool
+	ListenAddress      string
 	// To determine whether inherit owner's annotations for pods when create podgroup
 	InheritOwnerAnnotations bool
 	// WorkerThreadsForPG is the number of threads syncing podgroup operations
@@ -114,6 +117,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.IntVar(&s.MaxRequeueNum, "max-requeue-num", defaultMaxRequeueNum, "The number of times a job, queue or command will be requeued before it is dropped out of the queue")
 	fs.StringVar(&s.HealthzBindAddress, "healthz-address", defaultHealthzAddress, "The address to listen on for the health check server.")
 	fs.BoolVar(&s.EnableHealthz, "enable-healthz", false, "Enable the health check; it is false by default")
+	fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")
+	fs.StringVar(&s.ListenAddress, "listen-address", defaultListenAddress, "The address to listen on for HTTP requests.")
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
--- a/cmd/controller-manager/app/server.go
+++ b/cmd/controller-manager/app/server.go
@@ -19,6 +19,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 
 	v1 "k8s.io/api/core/v1"
@@ -40,6 +41,7 @@ import (
 	"volcano.sh/volcano/pkg/controllers/framework"
 	"volcano.sh/volcano/pkg/kube"
 	"volcano.sh/volcano/pkg/signals"
+	commonutil "volcano.sh/volcano/pkg/util"
 )
 
 // Run the controller.
@@ -48,11 +50,17 @@ func Run(opt *options.ServerOption) error {
 	if err != nil {
 		return err
 	}
-
 	if opt.EnableHealthz {
 		if err := helpers.StartHealthz(opt.HealthzBindAddress, "volcano-controller", opt.CaCertData, opt.CertData, opt.KeyData); err != nil {
 			return err
 		}
+	}
+
+	if opt.EnableMetrics {
+		go func() {
+			http.Handle("/metrics", commonutil.PromHandler())
+			klog.Fatalf("Prometheus Http Server failed %s", http.ListenAndServe(opt.ListenAddress, nil))
+		}()
 	}
 
 	run := startControllers(config, opt)
--- a/pkg/controllers/job/job_controller_handler.go
+++ b/pkg/controllers/job/job_controller_handler.go
@@ -36,6 +36,7 @@ import (
 	"volcano.sh/volcano/pkg/controllers/apis"
 	jobcache "volcano.sh/volcano/pkg/controllers/cache"
 	jobhelpers "volcano.sh/volcano/pkg/controllers/job/helpers"
+	"volcano.sh/volcano/pkg/controllers/job/state"
 )
 
 func (cc *jobcontroller) addCommand(obj interface{}) {
@@ -133,6 +134,9 @@ func (cc *jobcontroller) deleteJob(obj interface{}) {
 		klog.Errorf("Failed to delete job <%s/%s>: %v in cache",
 			job.Namespace, job.Name, err)
 	}
+
+	// Delete job metrics
+	state.DeleteJobMetrics(fmt.Sprintf("%s/%s", job.Namespace, job.Name), job.Spec.Queue)
 }
 
 func (cc *jobcontroller) addPod(obj interface{}) {
--- a/pkg/controllers/job/state/completing.go
+++ b/pkg/controllers/job/state/completing.go
@@ -17,6 +17,8 @@ limitations under the License.
 package state
 
 import (
+	"fmt"
+
 	vcbatch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	"volcano.sh/apis/pkg/apis/bus/v1alpha1"
 	"volcano.sh/volcano/pkg/controllers/apis"
@@ -33,6 +35,7 @@ func (ps *completingState) Execute(action v1alpha1.Action) error {
 			return false
 		}
 		status.State.Phase = vcbatch.Completed
+		UpdateJobCompleted(fmt.Sprintf("%s/%s", ps.job.Job.Namespace, ps.job.Job.Name), ps.job.Job.Spec.Queue)
 		return true
 	})
 }
--- a/pkg/controllers/job/state/metrics.go
+++ b/pkg/controllers/job/state/metrics.go
@@ -1,0 +1,39 @@
+package state
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"volcano.sh/volcano/pkg/scheduler/metrics"
+)
+
+var (
+	jobCompletedPhaseCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: metrics.VolcanoNamespace,
+			Name:      "job_completed_phase_count",
+			Help:      "Number of job completed phase",
+		}, []string{"job_name", "queue_name"},
+	)
+
+	jobFailedPhaseCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: metrics.VolcanoNamespace,
+			Name:      "job_failed_phase_count",
+			Help:      "Number of job failed phase",
+		}, []string{"job_name", "queue_name"},
+	)
+)
+
+func UpdateJobCompleted(jobName, queueName string) {
+	jobCompletedPhaseCount.WithLabelValues(jobName, queueName).Inc()
+}
+
+func UpdateJobFailed(jobName, queueName string) {
+	jobFailedPhaseCount.WithLabelValues(jobName, queueName).Inc()
+}
+
+func DeleteJobMetrics(jobName, queueName string) {
+	jobCompletedPhaseCount.DeleteLabelValues(jobName, queueName)
+	jobFailedPhaseCount.DeleteLabelValues(jobName, queueName)
+}
--- a/pkg/controllers/job/state/running.go
+++ b/pkg/controllers/job/state/running.go
@@ -17,6 +17,8 @@ limitations under the License.
 package state
 
 import (
+	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 
 	vcbatch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
@@ -62,6 +64,7 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 			minSuccess := ps.job.Job.Spec.MinSuccess
 			if minSuccess != nil && status.Succeeded >= *minSuccess {
 				status.State.Phase = vcbatch.Completed
+				UpdateJobCompleted(fmt.Sprintf("%s/%s", ps.job.Job.Namespace, ps.job.Job.Name), ps.job.Job.Spec.Queue)
 				return true
 			}
 
@@ -76,6 +79,7 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 						if taskStatus, ok := status.TaskStatusCount[task.Name]; ok {
 							if taskStatus.Phase[v1.PodSucceeded] < *task.MinAvailable {
 								status.State.Phase = vcbatch.Failed
+								UpdateJobFailed(fmt.Sprintf("%s/%s", ps.job.Job.Namespace, ps.job.Job.Name), ps.job.Job.Spec.Queue)
 								return true
 							}
 						}
@@ -84,10 +88,13 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 
 				if minSuccess != nil && status.Succeeded < *minSuccess {
 					status.State.Phase = vcbatch.Failed
+					UpdateJobFailed(fmt.Sprintf("%s/%s", ps.job.Job.Namespace, ps.job.Job.Name), ps.job.Job.Spec.Queue)
 				} else if status.Succeeded >= ps.job.Job.Spec.MinAvailable {
 					status.State.Phase = vcbatch.Completed
+					UpdateJobCompleted(fmt.Sprintf("%s/%s", ps.job.Job.Namespace, ps.job.Job.Name), ps.job.Job.Spec.Queue)
 				} else {
 					status.State.Phase = vcbatch.Failed
+					UpdateJobFailed(fmt.Sprintf("%s/%s", ps.job.Job.Namespace, ps.job.Job.Name), ps.job.Job.Spec.Queue)
 				}
 				return true
 			}
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -62,7 +62,6 @@ import (
 	vcinformer "volcano.sh/apis/pkg/client/informers/externalversions"
 	cpuinformerv1 "volcano.sh/apis/pkg/client/informers/externalversions/nodeinfo/v1alpha1"
 	vcinformerv1 "volcano.sh/apis/pkg/client/informers/externalversions/scheduling/v1beta1"
-
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/features"
 	schedulingapi "volcano.sh/volcano/pkg/scheduler/api"
--- a/.github/workflows/e2e_spark.yaml
+++ b/.github/workflows/e2e_spark.yaml
@@ -62,6 +62,7 @@ jobs:
       run: |
         eval $(minikube docker-env)
         make TAG=latest update-development-yaml
+        sed -i 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' installer/volcano-development.yaml
         make TAG=latest images
         docker images | grep volcano
         cat ./installer/volcano-development.yaml  | grep image:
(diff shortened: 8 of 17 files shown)


googs1025:
This implementation is too complex; we just need to add metrics at the state transition points.

googs1025:
In practical scenarios, we only focus on the final states of success and failure, and we are not concerned with or do not need other states because they are not final states.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -69,6 +70,9 @@ func (cc *jobcontroller) addJob(obj interface{}) {
 	key := jobhelpers.GetJobKeyByReq(&req)
 	queue := cc.getWorkerQueue(key)
 	queue.Add(req)
+
+	// Record job phase metric
+	go cc.recordJobPhaseMetric()
```
googs1025, line 75: Why do we need to use goroutine? What are the considerations?

pkg/scheduler/metrics/job.go
```diff
@@ -37,8 +39,104 @@ var (
 			Help:      "Number of retry counts for one job",
 		}, []string{"job_id"},
 	)
+
+	jobPendingPhaseCount = promauto.NewGauge(
```
googs1025, line 43: IMO, we only need to pay attention to the final state, and we don’t care about other states because the state will change.

googs1025:
You just need to find the place where the controller last modified the status in these two related places and make the changes.
`status.State.Phase = vcbatch.Completed`
`status.State.Phase = vcbatch.Failed`

Prepmachine4:
ok i understand

> You just need to find the place where the controller last modified the status in these two related places and make the changes. `status.State.Phase = vcbatch.Completed` `status.State.Phase = vcbatch.Failed`

ok

hwdef (commented):
/ok-to-test

pkg/controllers/job/state/completing.go
```diff
@@ -33,6 +34,7 @@ func (ps *completingState) Execute(action v1alpha1.Action) error {
 			return false
 		}
 		status.State.Phase = vcbatch.Completed
+		metrics.RegisterJobCompleted()
```
googs1025, line 37: Can we use Prometheus to test whether these metrics are indeed effective?

pkg/scheduler/metrics/job.go
```diff
@@ -37,8 +37,34 @@ var (
...
+			Subsystem: VolcanoNamespace,
+			Name:      "job_failed_phase_count",
+			Help:      "Number of jobs in failed phase",
+		},
+	)
 )
 
+// RegisterJobCompleted records the number of jobs in completed phase
+func RegisterJobCompleted() {
+	jobCompletedPhaseCount.Inc()
```
googs1025, line 60: Can we add labels? With labels, it will be beneficial for us to query metrics using PQL

Prepmachine4: I've thought about it, but I don't know what to add. Add `JobName`?

pkg/controllers/job/state/completing.go
Prepmachine4: What do you mean, I don't understand.

googs1025: Have we verified whether this metric is effective?

pkg/scheduler/metrics/job.go
googs1025: You can refer to other metrics codes

pkg/controllers/job/state/completing.go
Prepmachine4: I'm trying to test it in a local deployment environment.

pkg/scheduler/metrics/job.go
```diff
@@ -37,8 +37,34 @@ var (
...
+			Help:      "Number of jobs in completed phase",
+		}, []string{"job_ns"},
+	)
+
+	jobFailedPhaseCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: VolcanoNamespace,
+			Name:      "job_failed_phase_count",
+			Help:      "Number of jobs in failed phase",
+		}, []string{"job_ns"},
```
googs1025, line 54: If we only use the namespace label, can we get information about which job failed in the metric?

Prepmachine4:
image

image

I tested the metrics, and they were in line with expectations.

pkg/scheduler/metrics/job.go
googs1025: Should we also add the job name? Or should we use the "namespace/name" format to represent it?

googs1025:
squash the commit to one

pkg/controllers/job/state/running.go
```diff
@@ -62,6 +63,7 @@ func (ps *runningState) Execute(action v1alpha1.Action) error {
 			minSuccess := ps.job.Job.Spec.MinSuccess
 			if minSuccess != nil && status.Succeeded >= *minSuccess {
 				status.State.Phase = vcbatch.Completed
+				metrics.RegisterJobCompleted(ps.job.Job.Namespace)
```
googs1025, line 66: `metrics.RegisterJobCompleted(fmt.Sprintf("%s/%s", ps.job.Job.Namespace, ps.job.Job.Name))`

Same for other parts

pkg/scheduler/metrics/job.go
```diff
@@ -37,8 +37,34 @@ var (
...
+			Help:      "Number of jobs in completed phase",
+		}, []string{"job_ns"},
+	)
+
+	jobFailedPhaseCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: VolcanoNamespace,
+			Name:      "job_failed_phase_count",
+			Help:      "Number of jobs in failed phase",
+		}, []string{"job_ns", "job_id"},
```
googs1025, line 54: `[]string{"job_name"}`

Directly use job_name as the label. This job_name represents the combination of <namespace/name>

Prepmachine4:
> squash the commit to one

If i squash the commit to one, it seems like I need to re-create a PR, or is there another way?

googs1025:
> > squash the commit to one
> 
> If i squash the commit to one, it seems like I need to re-create a PR, or is there another way?


Why do we need to resubmit a PR when using git rebase or git squash? This is to update your git log view locally, and has nothing to do with the PR itself.

Prepmachine4:
> > > squash the commit to one
> > 
> > 
> > If i squash the commit to one, it seems like I need to re-create a PR, or is there another way?
> 
> Why do we need to resubmit a PR when using git rebase or git squash? This is to update your git log view locally, and has nothing to do with the PR itself.

ok ,i have squashed the commit to one.

hwdef (commented):
/lgtm

Monokaix:
Thanks for you contribution! It changed a lot cause vc controller also introduced metrics.

pkg/scheduler/metrics/job.go
```diff
@@ -57,4 +83,6 @@ func DeleteJobMetrics(jobName, queue, namespace string) {
 	unscheduleTaskCount.DeleteLabelValues(jobName)
 	jobShare.DeleteLabelValues(namespace, jobName)
 	jobRetryCount.DeleteLabelValues(jobName)
+	jobCompletedPhaseCount.DeleteLabelValues(namespace, jobName)
```
Monokaix, line 86: The new added metrics exist in vc controller but these are defined in scheduler, this is a liitel wired: )

pkg/scheduler/metrics/job.go
```diff
@@ -37,8 +37,34 @@ var (
...
+			Help:      "Number of jobs in completed phase",
+		}, []string{"job_name"},
+	)
+
+	jobFailedPhaseCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: VolcanoNamespace,
+			Name:      "job_failed_phase_count",
+			Help:      "Number of jobs in failed phase",
+		}, []string{"job_name"},
```
Monokaix, line 54: Maybe we can also add a queue lable so that users can get metris by specified queue.

googs1025: sgtm

pkg/scheduler/metrics/job.go
googs1025: +1, move to pkg/controller is better

Prepmachine4: ok

googs1025:
> > > > squash the commit to one
> > > 
> > > 
> > > If i squash the commit to one, it seems like I need to re-create a PR, or is there another way?
> > 
> > 
> > Why do we need to resubmit a PR when using git rebase or git squash? This is to update your git log view locally, and has nothing to do with the PR itself.
> 
> ok ,i have squashed the commit to one.

squashed to one commit. 
If you want to merge the master branch, please use rebase, which will allow reviewers to review better.

Prepmachine4:
> > Thanks for you contribution! It changed a lot cause vc controller also introduced metrics.
> 
> What I want to say is that can we just put all metrics in vc scheduler, put metrics into both scheduler and controller is not a good way.

My initial implementation is like this, but it's a little bit complicated.That's why it's implemented in controller instead.

googs1025:
I think the key point of this discussion is: which component should be responsible for changing the status? Since the controller is the component that job status is controlled by, it should be the one sending this variable. The scheduler itself doesn't care about the status of jobs or other workloads, which is why I believe it's not suitable to place it in the scheduler. Additionally, each component in Kubernetes has its own metrics.

lowang-bh:
> What I want to say is that can we just put all metrics in vc scheduler, put metrics into both scheduler and controller is not a good way.

Agree. Controller is just responsible for volcano job. Scheduler is responsible for all kinds workloads and it will update job status represent in podgroup.

lowang-bh:
> Since the controller is the component that job status is controlled by, it should be the one sending this variable. 

It can only cover vc jobs.

> The scheduler itself doesn't care about the status of jobs or other workload

Sheduler can skip those already running jobs.

Prepmachine4:
ok, I will change the code again.

googs1025:
> ok, I will change the code again.

Okay, I'm just putting my two cents in here. My reasoning is solely based on the approaches taken by the main Kubernetes repository or other SIGs' controllers. However, other approvers might have different viewpoints. It might not just be about architecture design, as introducing a new data source for the controller could involve adapting the existing monitoring systems, which might be another reason why we wouldn't want to add this to the controller. I'm sorry for any inconvenience this may cause you.

Prepmachine4:
> > ok, I will change the code again.
> 
> Okay, I'm just putting my two cents in here. My reasoning is solely based on the approaches taken by the main Kubernetes repository or other SIGs' controllers. However, other approvers might have different viewpoints. It might not just be about architecture design, as introducing a new data source for the controller could involve adapting the existing monitoring systems, which might be another reason why we wouldn't want to add this to the controller. I'm sorry for any inconvenience this may cause you.

It doesn't matter, I'm also familiar with the controller component, and maybe it will helpful  for the future contribution. Anyway, thanks for your advice.

JesseStutler:
Could we push forward with this PR? I have a PR depending on metrics, so I wish this PR can be merged asap

Prepmachine4:
> Could we push forward with this PR? I have a PR depending on metrics, so I wish this PR can be merged asap

Well, I was busy a few days ago. I will finish this PR asap

Prepmachine4:
The scheduler open session every minute that casus

Prepmachine4:
The scheduler open session every second, and podgroup is deleted when it becomes completed, making it impossible for the scheduler to sense whether the job succeeded or failed. 

Do I need to directly pull jobs from apiserver to collect metrics? I don't think this is a reasonable way to implement, please give me some advice.

image

```go
func (ju *jobUpdater) updateJob(index int) {
	job := ju.jobQueue[index]
	ssn := ju.ssn

	job.PodGroup.Status = jobStatus(ssn, job)
	oldStatus, found := ssn.podGroupStatus[job.UID]
	updatePG := !found || isPodGroupStatusUpdated(job.PodGroup.Status, oldStatus)
	klog.Infof("oldStatus: %v, newStatus: %v, updatePG: %v", oldStatus, job.PodGroup.Status, updatePG)
	if _, err := ssn.cache.UpdateJobStatus(job, updatePG); err != nil {
		klog.Errorf("Failed to update job <%s/%s>: %v",
			job.Namespace, job.Name, err)
	}
}
```

pkg/scheduler/framework/job_updater.go
```diff
@@ -2,13 +2,11 @@ package framework
...
 	"context"
-	"math/rand"
-	"time"
-
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
-
+	"math/rand"
+	"time"
```
googs1025, line 9: sort import

JesseStutler:
I think we don't need to record the metrics here in the scheduler @Monokaix? We currently intend to record these metrics in vc-controller, see #3751. Currently only capacity and proportion will expose these podgroup pending/inqueue/unknown/running count metrics outside, but these counts need to be persisted to queue status, but we're not going to persist these statistics anymore, and I don't think we should record the failed and completed metrics separately in scheduler.

googs1025:
> I think we don't need to record the metrics here in the scheduler @Monokaix? We currently intend to record these metrics in vc-controller, see #3751. Currently only capacity and proportion will expose these podgroup pending/inqueue/unknown/running count metrics outside, but these counts need to be persisted to queue status, but we're not going to persist these statistics anymore, and I don't think we should record the failed and completed metrics separately in scheduler. 


more context we have discussed 
https://github.com/volcano-sh/volcano/pull/3650#issuecomment-2275601180

pkg/scheduler/cache/cache.go
```diff
@@ -1482,6 +1482,15 @@ func (sc *SchedulerCache) UpdateJobStatus(job *schedulingapi.JobInfo, updatePG b
 			return nil, err
 		}
 		job.PodGroup = pg
+
+		// record job Succeeded/Failed metrics
+		if job.PodGroup.Status.Phase == scheduling.PodGroupCompleted {
```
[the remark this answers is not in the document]
Prepmachine4: The code is ok, but the current changes will not achieve the goal, and here's why: https://github.com/volcano-sh/volcano/pull/3650#issuecomment-2399177518

Prepmachine4:
@Monokaix What do you think about this.

pkg/scheduler/cache/cache.go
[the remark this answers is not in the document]
Monokaix: Here is the status of pg, but seems the issue mentioned is to get job status instead of pg?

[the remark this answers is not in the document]
Monokaix: If the status is updated in controller, I think we can put these metrics in controller.

Prepmachine4:
I went back to the way I used to get metric in the controller, so please  review the code again.

hwdef:
Hi, thanks for your contributions! 
Please fix the CI

cmd/controller-manager/app/options/options.go
```diff
@@ -37,6 +37,7 @@ const (
 	defaultMaxRequeueNum       = 15
 	defaultSchedulerName       = "volcano"
 	defaultHealthzAddress      = ":11251"
+	defaultListenAddress       = ":8080"
```
hwdef, line 40: It is recommended to use port 8081 to distinguish it from the scheduler.

pkg/controllers/job/state/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	"volcano.sh/volcano/pkg/scheduler/metrics"
+)
+
+var (
+	jobCompletedPhaseCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: metrics.VolcanoNamespace,
+			Name:      "job_completed_phase_count",
+			Help:      "Number of job completed phase",
+		}, []string{"job_id", "queue_name"},
```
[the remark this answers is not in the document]
Prepmachine4: Sorry, it should be jobname.

Prepmachine4:
> Hi, thanks for your contributions! Please fix the CI

I tried to fix it, but I don't know why the 'default' queue is not open. Can you give me some advice on where I should start.
.image

Monokaix:
> > Hi, thanks for your contributions! Please fix the CI
> 
> I tried to fix it, but I don't know why the 'default' queue is not open. Can you give me some advice on where I should start. .image

You can rebase master and try again.

Prepmachine4:
> > > Hi, thanks for your contributions! Please fix the CI
> > 
> > 
> > I tried to fix it, but I don't know why the 'default' queue is not open. Can you give me some advice on where I should start. .image
> 
> You can rebase master and try again.

There is no commit that requires rebase, it is already up to date.

hwdef:
please try close and reopen this PR, to trigger CI

Prepmachine4:
image
It seems that the issue is caused by the controller image used in the 'E2E Spark Integration Test' run not being the latest version.

hwdef:
Where are these logs?

hwdef:
So weird. 
No problem compiling your code locally

hwdef:
> > Hi, thanks for your contributions! Please fix the CI
> 
> I tried to fix it, but I don't know why the 'default' queue is not open. Can you give me some advice on where I should start. .image

Encountered this error again, It may be that volcano has just been deployed and the test process has been executed. The queue of volcano has not yet been opened.

Please add
```
sleep 30s
```
in 

https://github.com/volcano-sh/volcano/blob/320698e72dc3357fff934fe3e49577fff3c4ec32/.github/workflows/e2e_spark.yaml#L68

target:

```yaml
...
    - name: Build lastest volcano images
      run: |
        eval $(minikube docker-env)
        make TAG=latest update-development-yaml
        make TAG=latest images
        docker images | grep volcano
        cat ./installer/volcano-development.yaml  | grep image:
        minikube kubectl -- replace --force -f ./installer/volcano-development.yaml
        sleep 30s
...

```

hwdef:
I made a mistake, the reason why the queue has no status is because the controller failed to start. Not for the reason I mentioned above


The reason for this is mentioned below. It seems that the controller does not use the latest code.


> image
> It seems that the issue is caused by the controller image used in the 'E2E Spark Integration Test' run not being the latest version.

hwdef:
Maybe it’s because the pull policy of the image is `always`.

Prepmachine4:
> Maybe it’s because the pull policy of the image is `always`.

change to IfNotPresent?

hwdef:
Yes, But I didn't think of a better replacement method. Please try add this

```
sed -i 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' installer/volcano-development.yaml
```

at 
https://github.com/volcano-sh/volcano/blob/320698e72dc3357fff934fe3e49577fff3c4ec32/.github/workflows/e2e_spark.yaml#L64

target:
```
    - name: Build lastest volcano images
      run: |
        eval $(minikube docker-env)
        make TAG=latest update-development-yaml
        sed -i 's/imagePullPolicy: Always/imagePullPolicy: IfNotPresent/g' installer/volcano-development.yaml
        make TAG=latest images
        docker images | grep volcano
        cat ./installer/volcano-development.yaml  | grep image:
        minikube kubectl -- replace --force -f ./installer/volcano-development.yaml
```

Do you have any better suggestions?

cmd/controller-manager/app/server.go
```diff
@@ -179,3 +190,10 @@ func isControllerEnabled(name string, controllers []string) bool {
 	// if we get here, there was no explicit inclusion or exclusion
 	return hasStar
 }
+
+func promHandler() http.Handler {
```
Monokaix, line 194: This is duplicated with scheduler，we can abstract a method.

pkg/controllers/job/state/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	jobFailedPhaseCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: metrics.VolcanoNamespace,
+			Name:      "job_failed_phase_count",
+			Help:      "Number of job failed phase",
+		}, []string{"job_name", "queue_name"},
+	)
+)
+
+func RegisterJobCompleted(jobName, queueName string) {
```
Monokaix, line 28: UpdateXXXCount is better.

JesseStutler:
Please see my review opinions also:
1. I think we should use gauge, not counter. 
2. In command line, controller set the enable-metrics to false:
```fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")```
but in helm chart template that `--enable-metrics` is written to be true, we should make this configurable, for example using {{ .Values.EnableMetrics}} and then add EnableMetrics:true in values.yaml
3. You should not delete the metrics in scheduler, scheduler does not have these metrics.

**Comment by Prepmachine4** [comment]
> Please see my review opinions also:
> 
> 1. I think we should use gauge, not counter.
> 2. In command line, controller set the enable-metrics to false:
>    `fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")`
>    but in helm chart template that `--enable-metrics` is written to be true, we should make this configurable, for example using {{ .Values.EnableMetrics}} and then add EnableMetrics:true in values.yaml
> 3. You should not delete the metrics in scheduler, scheduler does not have these metrics.

Thank you for your suggestions. I have made changes to points 2 and 3, but I think using `counter` is more appropriate because these two metrics only have increment operations.

File: pkg/controllers/job/state/metrics.go [hunk 16]
```diff
@@ -0,0 +1,39 @@
...
+
+func UpdateJobCompleted(jobName, queueName string) {
+	jobCompletedPhaseCount.WithLabelValues(jobName, queueName).Inc()
+}
+
+func UpdateJobFailed(jobName, queueName string) {
+	jobFailedPhaseCount.WithLabelValues(jobName, queueName).Inc()
+}
+
+func DeleteJobMetrics(jobName, queueName string) {
```
Monokaix, line 36: This should be called when job is deleted.

hwdef (commented):
/lgtm

PR #3654, opened 2024-08-05
remove useless job valid check

lx1036:
because ssn.Tiers is not initiated and plugins has not OnSessionOpen() to register callback to jobValidFns

in openSession() function, ssn.Tiers is not initiated, it's nil, and the plugins has not OnSessionOpen() to register callback to jobValidFns, tier.jobValidFns map is empty. so it's no need to check job valid at session initiation stage.

```
func (ssn *Session) JobValid(obj interface{}) *api.ValidateResult {
			jrf, found := ssn.jobValidFns[plugin.Name]
			if !found {
				continue
			}

			if vr := jrf(obj); vr != nil && !vr.Pass {
				return vr
			}
		}
	}

	return nil
}
```

### DIFF
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -21,7 +21,6 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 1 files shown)


lowang-bh:
Can't just remove it.

Issue #3554, opened 2024-07-01
❇️  add vcinformer.SharedInformerFactory, vcClient     vcclient.Interface to framework.Session


kingeasternsun:
#### What would you like to be added:
add 
```go
vcinformer.SharedInformerFactory
vcClient     vcclient.Interface 
```

to framework.Sesstion

```go
type Session struct {
	UID types.UID

	kubeClient      kubernetes.Interface
	recorder        record.EventRecorder
	cache           cache.Cache
	restConfig      *rest.Config
	informerFactory informers.SharedInformerFactory

	vcinformer.SharedInformerFactory
	vcClient     vcclient.Interface 
```

#### Why is this needed:

Some plugin need get vcClient and vcinformer from session， otherwise they have to  create a separate vcClient and vcinformer.

for example, in ascend-volcano-plugin https://gitee.com/ascend/ascend-for-volcano/blob/branch_v6.0.0-RC1/plugin/factory.go#L426 , they have to create a a separate vcClient and vcinformer, these vcinfomer cannot share indexer cache with the informer in scheduler.

```go

func (sHandle *ScheduleHandler) initNSLB2(ssn *framework.Session) error {
	vcClient, err := NewVCClinetk8s()
	if err != nil {
		klog.V(util.LogErrorLev).Infof("create vc client err: %v.", err)
		return err
	}
	factory := externalversions.NewSharedInformerFactory(vcClient, 0)
	pgInformer := factory.Scheduling().V1beta1().PodGroups()
	cacheIndexer := pgInformer.Informer().GetIndexer()
	agents, err := NewAgent(ssn.KubeClient(), newConfig())
	if err != nil {
		klog.V(util.LogErrorLev).Infof("create agent err: %v", err)
		return err
	}
	...
}
```

Monokaix:
I think it's reasonable.

lowang-bh:
Why informer instead of snapshot of that?

Prepmachine4:
Does this feature need to be added? I have proposed a PR.

googs1025:
> Why informer is used instead of snapshot of that?

+1, IMO, I also think it is more reasonable to use snapshot.

But I am not sure about the purpose of these two variables introduced by this plugin. I think if we need to obtain resource objects from k8s, we should use snapshot instead of introducing informerFactory again, because snapshot is originally a cache obtained by informerFactory.
Excessive introduction of other variables is not a good design method, especially the informerFactory cache, which may also increase memory usage.

[...]

PR #3657, opened 2024-08-06
feat: session struct add vcinformer and vcClient field

Prepmachine4:
fix: #3554

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -942,6 +942,9 @@ func (sc *SchedulerCache) Client() kubernetes.Interface {
 	return sc.kubeClient
 }
 
+// VolcanoClient returns the volcano clientSet
+func (sc *SchedulerCache) VolcanoClient() vcclient.Interface { return sc.vcClient }
+
 // ClientConfig returns the rest config
 func (sc *SchedulerCache) ClientConfig() *rest.Config {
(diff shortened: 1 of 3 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -951,6 +954,9 @@ func (sc *SchedulerCache) ClientConfig() *rest.Config {
 func (sc *SchedulerCache) SharedInformerFactory() informers.SharedInformerFactory {
 	return sc.informerFactory
 }
+func (sc *SchedulerCache) VolcanoSharedInformerFactory() vcinformer.SharedInformerFactory {
```
googs1025, line 957: nit: space

googs1025, line 957: If there are comments elsewhere, it is recommended to add them as well.

lowang-bh:
we'd better use snapshot. cc @Monokaix

Monokaix:
> we'd better use snapshot. cc @Monokaix

But it will also occupy more cahce.

Monokaix:
Hi, are you still working on this? Add a new interface to cache is not a proper way, please get it from snapshot.

Prepmachine4:
> Hi, are you still working on this? Add a new interface to cache is not a proper way, please get it from snapshot.

I have stopped working in this issue, so is this PR need to be closed?

PR #3612, opened 2024-07-19
Pod Scheduling Readiness Draft

ykcai-daniel:
Add new state PodGroupSchGated and modified scheduler actions. More details in https://github.com/volcano-sh/volcano/pull/3581

PR #3658, opened 2024-08-07
Pod Scheduling Readiness

ykcai-daniel:
Updated PR of #3612, a gated task no longer blocks the job. Do not use new state.

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -28,12 +28,14 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/klog/v2"
+
+	"k8s.io/kubernetes/pkg/features"
 
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	"volcano.sh/apis/pkg/apis/scheduling"
 	"volcano.sh/apis/pkg/apis/scheduling/v1beta1"
-
 	volumescheduling "volcano.sh/volcano/pkg/scheduler/capabilities/volumebinding"
 )
 
@@ -124,6 +126,7 @@ type TaskInfo struct {
 	Preemptable                 bool
 	BestEffort                  bool
 	HasRestartableInitContainer bool
+	SchGated                    bool
 
 	// RevocableZone supports setting volcano.sh/revocable-zone annotation or label for pod/podgroup
 	// we only support empty value or * value for this version and we will support specify revocable zone name for future releases
@@ -178,7 +181,8 @@ func NewTaskInfo(pod *v1.Pod) *TaskInfo {
 	topologyInfo := GetPodTopologyInfo(pod)
 	role := getTaskRole(pod)
 	hasRestartableInitContainer := hasRestartableInitContainer(pod)
-
+	// initialize pod scheduling gates info here since it will not change in a scheduling cycle
+	schGated := calSchedulingGated(pod)
 	jobID := getJobID(pod)
 
 	ti := &TaskInfo{
@@ -196,6 +200,7 @@ func NewTaskInfo(pod *v1.Pod) *TaskInfo {
 		HasRestartableInitContainer: hasRestartableInitContainer,
 		RevocableZone:               revocableZone,
 		NumaInfo:                    topologyInfo,
+		SchGated:                    schGated,
 		TransactionContext: TransactionContext{
 			NodeName: pod.Spec.NodeName,
 			Status:   getTaskStatus(pod),
@@ -229,6 +234,16 @@ func (ti *TaskInfo) GenerateLastTxContext() {
 // ClearLastTxContext clear context of last transaction for a task
 func (ti *TaskInfo) ClearLastTxContext() {
 	ti.LastTransaction = nil
+}
+
+// Return if the pod of a task is scheduling gated by checking if length of sch gates is zero
+// When the Pod is not yet created or sch gates field not set, return false
+func calSchedulingGated(pod *v1.Pod) bool {
+	// Only enable if features.PodSchedulingReadiness feature gate is enabled
+	if utilfeature.DefaultFeatureGate.Enabled(features.PodSchedulingReadiness) {
+		return pod != nil && pod.Spec.SchedulingGates != nil && len(pod.Spec.SchedulingGates) != 0
+	}
+	return false
 }
 
 func (ti *TaskInfo) SetPodResourceDecision() error {
@@ -273,6 +288,7 @@ func (ti *TaskInfo) Clone() *TaskInfo {
 		HasRestartableInitContainer: ti.HasRestartableInitContainer,
 		RevocableZone:               ti.RevocableZone,
 		NumaInfo:                    ti.NumaInfo.Clone(),
+		SchGated:                    ti.SchGated,
 		TransactionContext: TransactionContext{
 			NodeName: ti.NodeName,
 			Status:   ti.Status,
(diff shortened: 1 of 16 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -575,6 +590,17 @@ func (ji *JobInfo) DeleteTaskInfo(ti *TaskInfo) error {
 	return nil
 }
 
+func (ji* JobInfo) HaveSchGatedTask() bool{
+	for _,task:=range ji.Tasks{
```
googs1025, line 594: nit: code style, need space

googs1025:
Does this change have the same purpose as this PR https://github.com/volcano-sh/volcano/pull/3612
?

Monokaix:
Hi，I think there is no necessary to submit two PRs, just based the previous PR and update it OK，and please also update the design doc.

Monokaix:
Please refer to kube-scheduler to add a featuregate to enable/disable this feature.

pkg/scheduler/api/job_info.go
ykcai-daniel: Fixed

ykcai-daniel:
@Monokaix PR updated. I will also update the design proposal #3581 to document the changes.

ykcai-daniel:
PR summary: Pod Scheduling gates is a new feature in K8S, which uses the `.spec.schedulingGates` field of Pod to signal that a Pod is not ready for scheduling (Pod will be in pending state when gated). This feature is often used to implement customized resource managers and we want to support it for compatibility. Main Changes involve:

1.  Scheduling Actions: Job with scheduling gated tasks mainly exist in the Inqueue state. In allocate action, if the pod of a task is scheduling gates, it will be skipped for allocation and not bound to a node. Consequently, if too many tasks are scheduling gates, the job will be gang-unschedulable
2. Plugins: Since scheduling gated pods are not ready to be allocated, we don't scheduling gated tasks to consume inqueue resources, making other jobs uninqueuable. Therefore, proportion, capacity and overcommit plugins have to be changed to deduct the resources of scheduling gated tasks
3. Pod Conditions: When each session close, the Condition of scheduling gated pods will not be updated, showing the Schedulable False condition with reason ScheduleGated, which is created by api-server

Future Works that are not covered in  this PR:

1. Now, we do not support removing scheduling gates by modifying the template of a vcjob because supporting this feature involve major changes to controller and Pod-level gates removal is sufficient for most usecase. In the future, if there is a need for task template support, we can add it.
2. Events and Correct Message: Now, if a Job is gang unschedulable due to scheduling gates, the reason in condition is "NoEnoughResources". We might need better reason and message for scheduling gates.

ykcai-daniel:
Reviewer: /assign @Monokaix /assign @googs1025

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -142,7 +142,7 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		}
 
 		if job.PodGroup.Status.Phase == scheduling.PodGroupInqueue {
-			attr.inqueue.Add(job.GetMinResources())
+			attr.inqueue.Add(job.DeductSchGatedResources(job.GetMinResources()))
```
lowang-bh, line 145: I think its better to do the calcalation in GetMinResources function. Let GetMinResources return the min resource to run the job.

pkg/scheduler/api/job_info.go
```diff
@@ -228,6 +230,17 @@ func (ti *TaskInfo) ClearLastTxContext() {
 	ti.LastTransaction = nil
 }
 
+// Return if the pod of a task is scheduling gated by checking if length of sch gates
+// When the Pod is not yet created or sch gates field not set, return false
+func (ti *TaskInfo) SchedulingGated() bool {
```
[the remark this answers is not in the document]
ykcai-daniel: I have added a field `schGated` into taskInfo

ykcai-daniel:
@lowang-bh cc @Monokaix  Pull request updated. Please review the changes. Thank you!

pkg/scheduler/api/job_info.go
```diff
@@ -228,6 +233,22 @@ func (ti *TaskInfo) ClearLastTxContext() {
...
+	if utilfeature.DefaultFeatureGate.Enabled(features.PodSchedulingReadiness) {
+		return pod != nil && pod.Spec.SchedulingGates != nil && len(pod.Spec.SchedulingGates) != 0
+	}
+	return false
+
+}
+
+// ti.SchGated is initialized in NewTaskInfo
+func (ti *TaskInfo) SchedulingGated() bool {
+	return ti.SchGated
```
Monokaix, line 249: Should this filed be exported?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -282,18 +282,18 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		queue := ssn.Queues[queueID]
 		// If no capability is set, always enqueue the job.
 		if attr.realCapability == nil {
-			klog.V(4).Infof("Capability of queue <%s> was not set, allow job <%s/%s> to Inqueue.",
+			klog.V(3).Infof("Capability of queue <%s> was not set, allow job <%s/%s> to Inqueue.",
```
Monokaix, line 285: Why change these log level?

ykcai-daniel: I changed this log level when debugging the capcacity plugin. Will change is back

pkg/scheduler/api/job_info.go
ykcai-daniel: With this function, there is no need to export it. I will change it to `schGated`.

pkg/scheduler/api/job_info.go
```diff
@@ -504,6 +526,39 @@ func (ji *JobInfo) GetMinResources() *Resource {
...
+		}
+	}
+	return res
+}
+
+// Deduct resources of scheduling gated pod from Resource res;
+// If resource is less than gated resources, return zero;
+// Gated resources might be more than minResource when total rep of each task > minAvailable
+func (ji *JobInfo) DeductSchGatedResources(res *Resource) *Resource {
+	result := res.Clone()
```
Monokaix, line 545: We can check whether res Is zero to have a shortcut here, because in most case task is not gated.

pkg/scheduler/api/job_info.go
lowang-bh: I would like to use this filed instead of a function calling.

Monokaix: It's an optional way, but seems there is no need to expose an extra filed of task.

ykcai-daniel:
For more details about the design of the PR, see #3581 . Please also merge the design doc.

pkg/scheduler/api/job_info.go
Monokaix: I think it's reasonable.

pkg/scheduler/api/job_info.go
```diff
@@ -504,6 +526,44 @@ func (ji *JobInfo) GetMinResources() *Resource {
...
+		}
+	}
+	return res
+}
+
+// Deduct resources of scheduling gated pod from Resource res;
+// If resource is less than gated resources, return zero;
+// Note: The purpose of this functionis to deduct the resources of scheduling gated tasks
+// in a job when calculating inqueued resources so that it will not block other jobs from being inqueued.
+func (ji *JobInfo) DeductSchGatedResources(res *Resource) *Resource {
```
Monokaix, line 545: The comment should start with `DeductSchGatedResources` to be consistent with func name.

pkg/scheduler/api/job_info.go
```diff
@@ -504,6 +526,44 @@ func (ji *JobInfo) GetMinResources() *Resource {
...
+	result := res.Clone()
```
Monokaix, line 546: Seems there is no need to clone `res` because no one will also change it, and also we can change to:

```go
schGatedResource := ji.GetSchGatedPodResources()
// Most jobs do not have any scheduling gated tasks, hence we add this short cut
if schGatedResource.IsEmpty() {
    return res
}
```

pkg/scheduler/api/job_info.go
```diff
@@ -576,6 +636,15 @@ func (ji *JobInfo) DeleteTaskInfo(ti *TaskInfo) error {
 	return nil
 }
 
+func (ji *JobInfo) HaveSchGatedTask() bool {
```
Monokaix, line 639: Is this called somewhere?

pkg/scheduler/api/resource_info_test.go
```diff
@@ -818,21 +818,6 @@ func TestLessEqualWithDimension(t *testing.T) {
 			},
 			expected: true,
 		},
-		{
```
Monokaix, line 821: Why is this removed?

ykcai-daniel: Change Reverted

pkg/scheduler/api/job_info.go
ykcai-daniel: Redundent function removed.

pkg/scheduler/api/job_info.go
ykcai-daniel: Now, we don't clone `res` if scheduling gated resource is zero.

test/e2e/util/deployment.go
```diff
@@ -29,8 +29,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// CreateDeployment creates a new deployment
 func CreateDeployment(ctx *TestContext, name string, rep int32, img string, req v1.ResourceList) *appv1.Deployment {
+	return CreateDeploymentGated(ctx, name, rep, img, req, []v1.PodSchedulingGate{})
+}
+
+// CreateDeployment creates a new deployment
```
Monokaix, line 36: CreateDeployment -》 CreateDeploymentGated

Issue #3662, opened 2024-08-07
volcano should work with oidc clusters


snappyyouth:
### What is the problem you're trying to solve

I will like to add support to scheduler such that also clusters that use oidc authentication will work.
![image](https://github.com/user-attachments/assets/6d71a7c2-89bc-4777-98d5-afa7f401b852)

### Describe the solution you'd like

add oidc's auth provider in the /cmd/scheduler/main.go

_ "k8s.io/client-go/plugin/pkg/client/auth"

Monokaix:
Hi, please support more information about your use case.

snappyyouth:
my kubeconfig is like this.
![image](https://github.com/user-attachments/assets/e4450f61-4cd1-4948-aa6d-024d5233e334)

snappyyouth:
My partial kubeconfig looks like this.

![image](https://github.com/user-attachments/assets/e4450f61-4cd1-4948-aa6d-024d5233e334)

PR #3663, opened 2024-08-07
feat: add oidc's auth provider

snappyyouth:
#3662 

What is the problem you're trying to solve
I will like to add support to scheduler such that also clusters that use oidc authentication will work.
webhook-manager belike:
image
controller-manager belike:
image
scheduler belike:
image

Describe the solution you'd like
add oidc's auth provider in the /cmd/scheduler/main.go  /cmd/webhook-manager/main.go  /cmd/controller-manager/main.go 

_ "k8s.io/client-go/plugin/pkg/client/auth"

Arfter adding, all the comment running expectedly.

controller-manager belike:
image

scheduler belike:
image

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/pflag"
 	_ "go.uber.org/automaxprocs"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	cliflag "k8s.io/component-base/cli/flag"
 	componentbaseoptions "k8s.io/component-base/config/options"
 	"k8s.io/klog/v2"
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/pflag"
 	_ "go.uber.org/automaxprocs"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	cliflag "k8s.io/component-base/cli/flag"
 	componentbaseoptions "k8s.io/component-base/config/options"
 	"k8s.io/klog/v2"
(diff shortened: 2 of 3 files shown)


Monokaix:
The image in description is missing: )

snappyyouth:
> The image in description is missing: )

thx for the reminder.I has supplemented some images in the description .

hwdef:
Will it have any impact on the current usage, or output more error logs without using oidc?

snappyyouth:
> Will it have any impact on the current usage, or output more error logs without using oidc?

Sorry, I don't quite understand what you mean?
In current production, we must configure the config file through oidc auditing. Without oidc, all volcano components cannot run.
In addition, with the continuous practice of K8S, auditing operations is a trend, and I think it is necessary to add oidc authentication.

hwdef:
> > Will it have any impact on the current usage, or output more error logs without using oidc?
> 
> Sorry, I don't quite understand what you mean? In current production, we must configure the config file through oidc auditing. Without oidc, all volcano components cannot run. In addition, with the continuous practice of K8S, auditing operations is a trend, and I think it is necessary to add oidc authentication.

I mean if i don't use oidc. Will this change cause any confusion for me? 
I briefly checked the code and found that if init fails, an error will be returned.

https://github.com/kubernetes/client-go/blob/71959c526d543a5e4c3ca6fb808f535c2726483f/plugin/pkg/client/auth/oidc/oidc.go#L52

snappyyouth:
> > > Will it have any impact on the current usage, or output more error logs without using oidc?
> > 
> > 
> > Sorry, I don't quite understand what you mean? In current production, we must configure the config file through oidc auditing. Without oidc, all volcano components cannot run. In addition, with the continuous practice of K8S, auditing operations is a trend, and I think it is necessary to add oidc authentication.
> 
> I mean if i don't use oidc. Will this change cause any confusion for me? I briefly checked the code and found that if init fails, an error will be print.
> 
> https://github.com/kubernetes/client-go/blob/71959c526d543a5e4c3ca6fb808f535c2726483f/plugin/pkg/client/auth/oidc/oidc.go#L52

if you don't use oidc. This change will't  cause any confusion for you.

Errors only occur when registering repeatedly. 
https://github.com/kubernetes/client-go/blob/71959c526d543a5e4c3ca6fb808f535c2726483f/rest/plugin.go#L66

And when your config didn't use oidc, AuthProvider is nil. The program will skip.
https://github.com/kubernetes/client-go/blob/71959c526d543a5e4c3ca6fb808f535c2726483f/rest/transport.go#L139

All in all, it's completely safe!

Issue #3671, opened 2024-08-13
Podgroup controller didn't consider schedulerName when create podgroup for replicaset 


Monokaix:
### Description

Podgroup controller didn't consider schedulerName when create podgroup for replicaset, that will cause pg controller create pg even though it's not scheduled by voclano.

### Steps to reproduce the issue

1. Deploy a k8s cluster.
2. Install volcano
3. Execute `kubectl get pg -A`, you will find that some pgs are created for replicas even included volcano components itself.

`kubectl get pg -A`

<img width="1159" alt="image" src="https://github.com/user-attachments/assets/15ecb13e-564f-453d-b03a-38cd46cc41f5">

### Describe the results you received and expected

Don't create pg when replicasets are not scheduled by volcano.

### What version of Volcano are you using?

master

### Any other relevant information

That's because pr https://github.com/volcano-sh/volcano/pull/3561 added some logic to avoid another problem, but the original logic didn't filer replicasets by scheduler name.
We can refer https://github.com/volcano-sh/volcano/blob/fadc9c81a21735c7812c2792beb9ce0f2e7d9121/pkg/controllers/podgroup/pg_controller.go#L164
to filter replicasets are not scheduled by volcano when processing replicasets, then pg will not be created.

PR #3672, opened 2024-08-13
fix pg controller create redundancy podGroup when schedulerName isn't matched

liuyuanchun11:
fix issue #3671 

Check whether the value of pod.spec.schedulerName is matched before creating a podGroup.

PR #3675, opened 2024-08-15
[cherry-pick for release-1.9]fix pg controller create redundancy podGroup when schedulerName isn't matched

liuyuanchun11:
fix issue #3671

cherry-pick from #3672

Check whether the value of pod.spec.schedulerName is matched before creating a podGroup.

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -19,6 +19,7 @@ package podgroup
 import (
 	"context"
 	"encoding/json"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


hwdef:
I think when cherrypick, the source branch should be master

hwdef (commented):
I think when cherrypick, the source branch should be master

hwdef:
cherrypick should operate on only one commitId, not multiple commitIDs

/hold

liuyuanchun11:
> cherrypick should operate on only one commitId, not multiple commitIDs
> 
> /hold

It's cherry-pick from the master branch.

hwdef:
I thnik the commit ID is same with the master

PR #3676, opened 2024-08-15
[cherry-pick for release-1.10]fix pg controller create redundancy podGroup when schedulerName isn't matched

liuyuanchun11:
fix issue https://github.com/volcano-sh/volcano/issues/3671

cherry-pick from https://github.com/volcano-sh/volcano/pull/3672

Check whether the value of pod.spec.schedulerName is matched before creating a podGroup.

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -19,6 +19,7 @@ package podgroup
 import (
 	"context"
 	"encoding/json"
+	"slices"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


hwdef (commented):
I think when cherrypick, the source branch should be master

hwdef:
cherrypick should operate on only one commitId, not multiple commitIDs

/hold

liuyuanchun11:
> cherrypick should operate on only one commitId, not multiple commitIDs
> 
> /hold

liuyuanchun11:
> cherrypick should operate on only one commitId, not multiple commitIDs
> 
> /hold

It's cherry-pick from the master branch.

hwdef:
image

image

But the commit ID is not same. Did I get something wrong?

liuyuanchun11:
> image
> 
> image
> 
> But the commit ID is not same. Did I get something wrong?

As far as I know, the git cherry-pick operation creates a new commit ID, while only git merge or rebase can preserve the original commit ID. However, merge or rebase are typically used for batch modifications or synchronization, whereas I currently only need to synchronize a single commit. Perhaps you could demonstrate how to perform this operation?

hwdef:
Sorry, It was my mistake and you were right

PR #3680, opened 2024-08-19
fix always update pod nominatedNodeName when pod is pipelined

bibibox:
in the #3498 , we support updating NominatedNodeName for pipelined task
When a pod is in the pipelined status, Volcano always updates the pod's nominatedNodeName. However, according to the native Kubernetes semantics, nominatedNodeName should only be updated in the event of preemption.

For example, if there is a pod in the terminating state in the cluster and a new pending pod is created with no extra resources available, the kube-scheduler will mark the pending pod as unschedulable and will not update its nominatedNodeName field. This would immediately trigger autoscaler to scale up. However, Volcano would update the pod's nominatedNodeName field, and the autoscaler wont scaling up.
More detail description in this issue #3681 

This PR fixes the issue, making Volcano's behavior consistent with that of kube-scheduler.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -278,7 +278,7 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 			if task.InitResreq.LessEqual(bestNode.FutureIdle(), api.Zero) {
 				klog.V(3).Infof("Pipelining Task <%v/%v> to node <%v> for <%v> on <%v>",
 					task.Namespace, task.Name, bestNode.Name, task.InitResreq, bestNode.Releasing)
-				if err := stmt.Pipeline(task, bestNode.Name); err != nil {
+				if err := stmt.Pipeline(task, bestNode.Name, false); err != nil {
 					klog.Errorf("Failed to pipeline Task %v on %v in Session %v for %v.",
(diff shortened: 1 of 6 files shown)


william-wang (approved):
/approve

william-wang:
@bibibox please add your test step and result.

bibibox:
situation 1:
test step:
1. create a cluster with one 1C1G node
2. create a pod A, which request 1C1G, with the preStop command sleep 600
3. after the pod A is running, delete it. the pod A will be terminating in 10 minutes
4. create a pod B with request 1C1G, and it will be pending

result:
the pod B did not update the nominatedNodeName


situation 2:
test step:
1. create a cluster with one 1C1G node
2. create a pod A, which request 1C1G and has a 10 priority, with the preStop command sleep 600
3. after the pod A is running, create a pod B with request 1C1G and 100 priority, and it will preempt the pod A
4. the pod A will be terminating in 10 minutes, observe the pod B's status

result:
the pod B updated the nominatedNodeName

bibibox:
> @bibibox please add your test step and result.

Added to comments section

lowang-bh:
Hi, actully there are two problems in this pr:
1. reclaim action will also evict pod and do pipeline
2. evict operations may be reverted, does it has any effication?

I recommend store the norminated node in taskinfo as described in https://github.com/volcano-sh/volcano/issues/3491

PR #3686, opened 2024-08-21
remove deletedJobs queue in cache model

Wang-Kai:
To troubleshoot this issue, my colleague and I worked until 3 AM. I sincerely hope that this fix will be merged into the community repository.

## Backgroud

In the controller component, the cache module has a separate deletedJobs queue specifically for handling job deletions. The job-controller also has a queue to process pod and job events. In edge cases, a situation may occur where a job is deleted from etcd but not from the cache. This leads to pods being created by the job-controller and immediately deleted by the gc-controller, causing a repetitive loop until the controller is restarted.

### DIFF
--- a/pkg/controllers/cache/interface.go
+++ b/pkg/controllers/cache/interface.go
@@ -25,13 +25,11 @@ import (
 
 // Cache Interface.
 type Cache interface {
-	Run(stopCh <-chan struct{})
-
 	Get(key string) (*apis.JobInfo, error)
 	GetStatus(key string) (*v1alpha1.JobStatus, error)
 	Add(obj *v1alpha1.Job) error
 	Update(obj *v1alpha1.Job) error
-	Delete(obj *v1alpha1.Job) error
+	Delete(key string) error
 
 	AddPod(pod *v1.Pod) error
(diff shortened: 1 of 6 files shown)


hwdef:
PTAL @Monokaix @JesseStutler

hwdef:
still need. Please fix the CI and rebase the code.

pkg/controllers/cache/interface.go
```diff
@@ -25,13 +25,11 @@ import (
...
 // Cache Interface.
 type Cache interface {
-	Run(stopCh <-chan struct{})
-
 	Get(key string) (*apis.JobInfo, error)
 	GetStatus(key string) (*v1alpha1.JobStatus, error)
 	Add(obj *v1alpha1.Job) error
 	Update(obj *v1alpha1.Job) error
-	Delete(obj *v1alpha1.Job) error
+	Delete(key string) error
```
JesseStutler: Do we have to change the interface? This interface has been defined for a long time, and to avoid requiring users to change implementation when upgrading to a new version, I'd like to keep the current interface as it is

PR #3696, opened 2024-08-28
enquable and allocatable compare resource with the required dimensions

### LINKED ISSUE #3537: resource compare only consider requested dimension

A func `LessEqualWithDimension` has been added in PR #3522, which can be used to compare the only required dimension resource in those functions:

- [ ] AddOverusedFn
- [ ] AddPreemptiveFn
- [ ] AddJobEnqueueableFn
- [ ] AddAllocatableFn

#### Why is this needed:

Conserder those scenes: 
 - 1. Queue‘s Gpu is really overused, but used CPU and Memory are both less than deserved. So a coming job just requesting CPU/Mem should be allocated.
 - 2. same as 1, Queue's CPU/Memory are not used up, but cluster has no free resource, a coming job just requesting cpu/memory can reclaim(which is preemptive) from other queue.

what's more, a queue's preemptable job can use as most as queue's capacity when check queue's overused, because those resource can be reclaimed.

### LINKED ISSUE #3729: job in two queue will reclaim each other's tasks in dead loop
### Description

With a cluster has 11C CPU, Queue-a has a deserved=5C and capability=10C, same as queue-b.
First create job-a with replicas=5, and minAvailable=2. job a will take 10C.
Then create a job-b same as job-a, it will reclaim and evict job-a's two tasks. But now queue-a used is less than deserved, it will also reclaim from queue-b, and so on.

image
image
image

with scheduler cm
```yaml
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, reclaim, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: gang
        enablePreemptable: false
      - name: conformance
    - plugins:
      - name: overcommit
      - name: drf
        enablePreemptable: false
      - name: predicates
      - name: capacity
      - name: nodeorder
      - name: binpack
kind: ConfigMap
```

1. apply queue-a, queue-b with yaml
```yaml
kind: Queue
metadata:
  name: queue-a
spec:
  reclaimable: true
  deserved:
    cpu: 5
    memory: 2Gi
  capability:          
    cpu: 10
    memory: 5Gi
---
kind: Queue
metadata:
  name: queue-b
spec:
  reclaimable: true
  deserved:
    cpu: 5
    memory: 2Gi
  capability:          
    cpu: 10
    memory: 5Gi
```
3. apply job-a.yaml
```yaml
kind: Job
metadata:
  name: job-a
spec:
  schedulerName: volcano
  queue: queue-a
  tasks:
    - replicas: 5
      minAvailable: 2
      name: "master"
      template:
        metadata:
          annotations:
            volcano.sh/preemptable: "true"
        spec:
          containers:
            - image: nginx:1.14.2
              name: nginx
              resources:
                requests:
                  cpu: "2"
                  memory: "50Mi"
          restartPolicy: OnFailure
```
5. apply job-b.yaml
```yaml
kind: Job
metadata:
  name: job-b
spec:
  schedulerName: volcano
  queue: queue-b
  tasks:
    - replicas: 5
      minAvailable: 2
      name: "worker"
      template:
        metadata:
          annotations:
            volcano.sh/preemptable: "true"
        spec:
          containers:
            - image: nginx:1.14.2
              name: nginx
              resources:
                requests:
                  cpu: "2"
                  memory: "50Mi"
          restartPolicy: OnFailure
```

After upgrade image to https://github.com/volcano-sh/volcano/pull/3696, it keeps stable.

image

master

master branch at  95d5a923056b9833bf27f0ccdfb28b59cba28c2d

lowang-bh:
There is a func `LessEqualWithDimension` to only compare the wanted dimension resource in request after PR #3522

This PR is to replcase those resource compare.

also fixes #3729
and fixes #3537

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -257,7 +257,8 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		task := candidate.(*api.TaskInfo)
 		attr := cp.queueOpts[queue.UID]
 
-		overused := attr.deserved.LessEqualWithDimension(attr.allocated, task.InitResreq)
+		futureUsed := attr.allocated.Clone().Add(task.Resreq)
+		overused := !futureUsed.LessEqualWithDimension(attr.deserved, task.Resreq)
 		metrics.UpdateQueueOverused(attr.name, overused)
 		if overused {
 			klog.V(3).Infof("Queue <%v> can not reclaim, deserved <%v>, allocated <%v>, share <%v>",
@@ -272,8 +273,8 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 	ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
 		attr := cp.queueOpts[queue.UID]
 
-		free, _ := attr.realCapability.Diff(attr.allocated, api.Zero)
-		allocatable := candidate.Resreq.LessEqual(free, api.Zero)
+		futureUsed := attr.allocated.Clone().Add(candidate.Resreq)
+		allocatable := futureUsed.LessEqualWithDimension(attr.realCapability, candidate.Resreq)
 		if !allocatable {
 			klog.V(3).Infof("Queue <%v>: realCapability <%v>, allocated <%v>; Candidate <%v>: resource request <%v>",
 				queue.Name, attr.realCapability, attr.allocated, candidate.Name, candidate.Resreq)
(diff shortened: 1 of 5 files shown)


pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -301,19 +301,12 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		klog.V(5).Infof("job %s min resource <%s>, queue %s capability <%s> allocated <%s> inqueue <%s> elastic <%s>",
 			job.Name, minReq.String(), queue.Name, attr.realCapability.String(), attr.allocated.String(), attr.inqueue.String(), attr.elastic.String())
 		// The queue resource quota limit has not reached
-		r := minReq.Add(attr.allocated).Add(attr.inqueue).Sub(attr.elastic)
```
Monokaix, line 304: Here should also ut, it's a bugfix here through git commit history.

lowang-bh: You means the pr https://github.com/volcano-sh/volcano/pull/2637？ If the resource in request doesn't exist in capability, it remove from it, that is strange. eg, the pods or ephemeral-storage exist in node's `Allocatable`, it also should be a resource in capability.

Monokaix: Yeah, can the previous pr's problem be fixed?

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -124,6 +124,18 @@ func Test_capacityPlugin_OnSessionOpen(t *testing.T) {
 	queue8 := util.BuildQueueWithPriorityAndResourcesQuantity("q8", 1, nil, api.BuildResourceList("2", "4Gi"))
 	queue9 := util.BuildQueueWithPriorityAndResourcesQuantity("q9", 10, nil, api.BuildResourceList("2", "4Gi"))
 
+	// case5: p16 + p17 in queue5 will exceed queue's deserved, is not preemptive
```
Monokaix, line 127: queue5 -》queue10

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -272,8 +272,8 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 	ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
 		attr := cp.queueOpts[queue.UID]
 
-		free, _ := attr.realCapability.Diff(attr.allocated, api.Zero)
-		allocatable := candidate.Resreq.LessEqual(free, api.Zero)
+		futureUsed := attr.allocated.Clone().Add(candidate.Resreq)
```
Monokaix, line 275: We should also add new case to cover this.

Monokaix:
I think Check task.InitReq+queue.allocated <= queue.deserved is reasonable.
Assume that queue1's deserved is 3 CPUs, queue2's deseved is 1 CPU, queue1 is allocated 2.5 CPUs, queue2 is allocated 1 CPU, queue1 then submits a pod request1 CPU, and the allocated resources of queue1 does not reach deserved, triggering reclaim and recycling queue2 1 CPU, queue2's pod still has not reached deserved after being evicted, and continues to trigger the recycling of queue1's CPU.

pkg/scheduler/plugins/capacity/capacity.go
lowang-bh: Has existed in `TestEnqueueAndAllocatable`

PR #3698, opened 2024-08-28
fix:an unschedulable bestEffort Pod affects the scheduling of other pending bestEffort pods

Lily922:
fix:an unschedulable bestEffort Pod affects the scheduling of other pending bestEffort pods

If there is a pending BestEffort Pod in the cluster that cannot be scheduled, for example, the node affinity cannot be satisfied. The processing loop will break and the subsequent Pending Pods in `pendingTasks` slice can never be processed.


lowang-bh:
Coud you add a test case to cover it? thanks.

PR #3702, opened 2024-09-01
support min-max elastic quota scheduling

### LINKED ISSUE #3703: queue's deserve can not be returned back when other queue's tasks are all not preemptable

image there are two queues: A and B, and the cluster have 10GPUs, and both A and B have a deserved with 5GPUs and Capability with 10GPUs.
Now, user submit jobs with not preemptable tasks to queue A and use up 10GPUs. 
Then user submit jobs to queue B can not be scheduled.

To limit only tasks be preemptable can be scheduled between a queue's deserved and capability in allocate action.

### Additional context

_No response_

lowang-bh:
Image this scene：queue has some guranteed quota called `Min`. Pods can be scheduled when queue's used quota  + requests <= Min.  And there is also a limit quota called `Max`, which means the upper quota a queue's tasks can used. While the quota between Min and Max can only be used by preemtable tasks in a queue, because those quota are borrowed from other queues' Min, and should be returned back when need. So preemtable pods can be scheduled when queue's used quota + requests <= Max.

This feature also called Elastic Quota or Capacity Scheduling.
referance:
[capacity-scheduling](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/9-capacity-scheduling)
[Elastic Quota Management](https://koordinator.sh/docs/user-manuals/capacity-scheduling/)

This Pr will do those things base on capacity plugin: `Min` equals to `deserved` and `Max` equals to `capability`

1. Add a feature switch to enable or disable this feature, so that origin function will not be effected.
2. Add an overused function in capacity plugin, this function make sure queue's used will be under `Min` if job's tasks are not preemptable, or queue's used will be under `Max` if job's tasks are preemptable when schedule a job.
3. Change `Preemtive` function in capacity plugin to support check if queue's future used (a job's request + queue's allocated) is under `Min`. Only a job in a queue whose futrue used will not exceed its Min can preemt other victims.

relative issues: #3537
fixes #3703

The 1st commit is base on https://github.com/volcano-sh/volcano/pull/3649, please merge that PR first.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -33,11 +33,13 @@ type Action struct {
 	session *framework.Session
 	// configured flag for error cache
 	enablePredicateErrorCache bool
+	enableGangCheckOverused   bool // indicates whether consider job's min request(gang-scheduling) when check overused
 }
 
 func New() *Action {
 	return &Action{
-		enablePredicateErrorCache: true, // default to enable it
+		enablePredicateErrorCache: true,  // default to enable it
+		enableGangCheckOverused:   false, // default to disable it
 	}
 }
 
@@ -50,6 +52,7 @@ func (alloc *Action) Initialize() {}
 func (alloc *Action) parseArguments(ssn *framework.Session) {
 	arguments := framework.GetArgOfActionFromConf(ssn.Configurations, alloc.Name())
 	arguments.GetBool(&alloc.enablePredicateErrorCache, conf.EnablePredicateErrCacheKey)
+	arguments.GetBool(&alloc.enableGangCheckOverused, conf.EnableGangCheckOverusedKey)
 }
 
 func (alloc *Action) Execute(ssn *framework.Session) {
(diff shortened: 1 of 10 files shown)


Monokaix:
seems it's a little complex for users to use the capacity plugin or queue capability, and the problem in https://github.com/volcano-sh/volcano/issues/3703 is really a common case?

lowang-bh:
> seems it's a little complex for users to use the capacity plugin or queue capability, and the problem in #3703 is really a common case?

Another solution is to add a min-max plugin. But it also need modify some codes in main actions.

Monokaix:
>or queue's used will be under Max if job's tasks are preemptable when schedule a job.

`ssn.Allocatable` holds the capability check logic now.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -31,10 +31,16 @@ import (
...
+	// configured flag for error cache
+	enablePredicateErrorCache bool
+	enableGangCheckOverused   bool // indicates whether consider job's min request(gang-scheduling) when check overused
 }
 
 func New() *Action {
-	return &Action{}
+	return &Action{
+		enablePredicateErrorCache: true,  // default to enable it
+		enableGangCheckOverused:   false, // default to disable it
```
Monokaix, line 42: Why called `Gangxxx`?

lowang-bh: It will job's min request resource with gang-scheduling, not just a task's request.

hwdef:
Please rebase the master code

PR #3511, opened 2024-06-04
Add common labels for chart objects

Aakcht:
This PR adds possibility to set common labels for all chart objects except for CRDs

PR #3707, opened 2024-09-03
Chart: remove duplicate label field

Aakcht:
Small PR to remove duplicate labels field for admission init job that I accidentally added in https://github.com/volcano-sh/volcano/pull/3511


hwdef (commented):
/lgtm

PR #3045, opened 2023-08-12
Add uninqueueable reason in podgroup condition

lowang-bh:
Please merge API's PR https://github.com/volcano-sh/apis/pull/113 first, and then I need to update the go.mod and refresh the last commit.

Add un-inqueueable reson in podgroup condition if job is rejected to be enqueue, so that it is more clear when describe podgroup to see why job is pending.

This PR is about to change podgroup's pending condition caused by not enough queue's quota  from:
```
    message: '3/0 tasks in gang unschedulable: pod group is not ready, 3 minAvailable'
    reason: NotEnoughResources
```
to
```
    message:  '3/0 tasks in gang unschedulable: pod group is not ready, 3 minAvailabl, origin reason: queue resource quota insufficient'
    reason: NotInqueueable
```

# test result

## origin
```
spec:
  minMember: 3
  minResources:
    count/pods: "3"
    cpu: 1100m
    memory: 200Mi
    pods: "3"
    requests.cpu: 1100m
    requests.memory: 200Mi
  minTaskMember:
    master: 2
    work: 1
  queue: test
status:
  conditions:
  - lastTransitionTime: "2023-08-12T06:27:12Z"
    message: '3/0 tasks in gang unschedulable: pod group is not ready, 3 minAvailable'
    reason: NotEnoughResources
    status: "True"
    transitionID: 35ec245e-f901-47a9-a2b1-ef0456505f86
    type: Unschedulable
  phase: Pending
➜  volcano git:(add_uninqueue_state) ✗ kubectl get events |grep minavailable-job-4d40d46d-8bab-4914-9ddc-7a8e2aeda95a
18s         Normal    Unschedulable             podgroup/minavailable-job-4d40d46d-8bab-4914-9ddc-7a8e2aeda95a   queue resource quota insufficient
19s         Warning   Unschedulable             podgroup/minavailable-job-4d40d46d-8bab-4914-9ddc-7a8e2aeda95a   0/0 tasks in gang unschedulable: pod group is not ready, 3 minAvailable
```

## with this change
```
➜  volcano git:(add_uninqueue_state) ✗ # image with add_uninqueue_state-67bf4ad9a
➜  volcano git:(add_uninqueue_state) ✗ kubectl get pod -n volcano-system -l app=volcano-scheduler
NAME                                 READY   STATUS    RESTARTS   AGE
volcano-scheduler-5bc9875dbb-sjvvn   1/1     Running   0          3m35s
[...]

PR #3709, opened 2024-09-07
feat: add specified insufficient resource names at job enqueueable stage

lx1036:
When I use volcano, the podgroup status sometimes is pending, because of proportion plugin reject at job enqueueable stage. And it record event about "queue resource quota insufficient", but I'm often confused by which resource is insufficient, or cpu, or memory, or nvidia.com/a10, or nvidia.com/a100.

So, add a specified resource name is useful for me to know more messages.


hwdef:
Did this pr resolve your problem?
https://github.com/volcano-sh/volcano/pull/3045

PR #3717, opened 2024-09-11
fix: nil pointer when set node other resources

TymonLee:
It is possible that `NewNodeInfo(node *v1.Node)` in node_info.go received a nil node argument. In this situation, volcano-scheduler will crash because of nil pointer dereference.
b2cfdd32-1e80-4faa-bc2d-7cccf4c16d72

crash


TymonLee:
> The method `NewGPUDevices` below it has already checked node not nil: )

But it will return nil and ni.Others[vgpu.DeviceName] will be set "hamivgpu=nil". In such situation, scheduler could also panic if addResource() or subResource() is called.
image

Monokaix:
Ok，please sign off your commit.

TymonLee:
> Ok，please sign off your commit.

OK, done.

Monokaix:
cc @hwdef @lowang-bh

Issue #3713, opened 2024-09-10
fix staticcheck warnings


hwdef:
### What is the problem you're trying to solve

fix staticchcek warnings in volcano reop

```
cmd/controller-manager/app/server.go:84:9: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/controller-manager/app/server.go:85:42: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/scheduler/app/server.go:117:9: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/scheduler/app/server.go:118:42: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/webhook-manager/app/util.go:49:13: wait.Poll is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
cmd/webhook-manager/app/util.go:82:13: wait.Poll is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
example/extender/extender.go:28:5: var snapshot is unused (U1000)
pkg/controllers/garbagecollector/garbagecollector.go:24:2: package "k8s.io/apimachinery/pkg/api/errors" is being imported more than once (ST1019)
	pkg/controllers/garbagecollector/garbagecollector.go:25:2: other import of "k8s.io/apimachinery/pkg/api/errors"
pkg/controllers/job/job_controller_actions.go:29:2: package "k8s.io/apimachinery/pkg/api/errors" is being imported more than once (ST1019)
	pkg/controllers/job/job_controller_actions.go:30:2: other import of "k8s.io/apimachinery/pkg/api/errors"
pkg/controllers/job/job_controller_util_test.go:27:2: package "volcano.sh/apis/pkg/apis/batch/v1alpha1" is being imported more than once (ST1019)
	pkg/controllers/job/job_controller_util_test.go:28:2: other import of "volcano.sh/apis/pkg/apis/batch/v1alpha1"
pkg/controllers/jobflow/jobflow_controller.go:42:2: package "volcano.sh/volcano/pkg/controllers/jobflow/state" is being imported more than once (ST1019)
[...]

PR #3720, opened 2024-09-12
Fix staticcheck warnings on cmd and example

xovoxy:
#### What type of PR is this?
/kind feature

fix staticcheck warnings on cmd and example, warnings like this
```text
cmd/controller-manager/app/server.go:84:9: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/controller-manager/app/server.go:85:42: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/scheduler/app/server.go:117:9: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/scheduler/app/server.go:118:42: opt.LockObjectNamespace is deprecated: use ResourceNamespace instead.  (SA1019)
cmd/webhook-manager/app/util.go:49:13: wait.Poll is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
cmd/webhook-manager/app/util.go:82:13: wait.Poll is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
example/extender/extender.go:28:5: var snapshot is unused (U1000)
```

Parts of https://github.com/volcano-sh/volcano/issues/3713

### DIFF
--- a/cmd/webhook-manager/app/util.go
+++ b/cmd/webhook-manager/app/util.go
@@ -46,7 +46,7 @@ func addCaCertForWebhook(kubeClient *kubernetes.Clientset, service *router.Admis
 		var mutatingWebhookName = volcanoAdmissionPrefix + strings.ReplaceAll(service.Path, "/", "-")
 		var mutatingWebhook *v1.MutatingWebhookConfiguration
 		webhookChanged := false
-		if err := wait.Poll(time.Second, 5*time.Minute, func() (done bool, err error) {
+		if err := wait.PollUntilContextTimeout(context.Background(), time.Second, 5*time.Minute, true, func(_ context.Context) (done bool, err error) {
 			mutatingWebhook, err = kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(context.TODO(), mutatingWebhookName, metav1.GetOptions{})
 			if err != nil {
 				if apierrors.IsNotFound(err) {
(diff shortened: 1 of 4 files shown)


hwdef (commented):
/ok-to-test
By the way, this PR is too long. If it can be split into multiple parts, it may be more convenient to review.

xovoxy:
> /ok-to-test By the way, this PR is too long. If it can be split into multiple parts, it may be more convenient to review.

ohh right, let me split it~

example/extender/extender.go
```diff
@@ -21,6 +21,7 @@ import (
 	"io"
 	"net/http"
 
+	"k8s.io/klog"
```
hwdef, line 24: Currently volcano should be using v2?

xovoxy: done

example/extender/extender.go
```diff
@@ -49,6 +50,7 @@ func onSessionOpen(w http.ResponseWriter, r *http.Request) {
 		NamespaceInfo:  req.NamespaceInfo,
 		RevocableNodes: req.RevocableNodes,
 	}
+	klog.V(4).Infof("the snapshot of cluster %+v", *snapshot)
```
Monokaix, line 53: Why add this?

cmd/webhook-manager/app/util.go
```diff
@@ -46,7 +46,7 @@ func addCaCertForWebhook(kubeClient *kubernetes.Clientset, service *router.Admis
 		var mutatingWebhookName = volcanoAdmissionPrefix + strings.ReplaceAll(service.Path, "/", "-")
 		var mutatingWebhook *v1.MutatingWebhookConfiguration
 		webhookChanged := false
-		if err := wait.Poll(time.Second, 5*time.Minute, func() (done bool, err error) {
+		if err := wait.PollUntilContextTimeout(context.Background(), time.Second, 5*time.Minute, false, func(_ context.Context) (done bool, err error) {
```
Monokaix, line 49: How about set the immediate to true?

example/extender/extender.go
xovoxy:
```text
example/extender/extender.go:28:5: var snapshot is unused (U1000)
```
add this code to avoid staticcheck warnings, or delete the field. i think adding a log for this is better. if you have a better method, can tell me.

cmd/webhook-manager/app/util.go
xovoxy: alright, setting immediate to true more closely aligns with the behavior of wait.Poll,  i will make this change in this pr, and do the same adjustment in another pr.

PR #3723, opened 2024-09-14
[cherry-pick for release-1.9] fix: not remove podgroup uid will cause topology annotation to be useless 

JesseStutler:
cherry-pick #3711


JesseStutler:
1.9 doesn't have TaskRole in TaskInfo, temporarily not cherry-pick.

Issue #3738, opened 2024-09-19
Need to refactor the reclaim action


JesseStutler:
### Please describe your problem in detail

I'm ready to add `preemptionPolicy` related logic  in reclaim action, but when I'm confused that when the task's `preemptionPolicy` is `Never`, do I need to push back the job and queue, continue allowing other tasks or jobs to reclaim resources. You can see that in the reclaim action, https://github.com/volcano-sh/volcano/blob/0843c0d33fccdb85439dc7086dd7cea061070901/pkg/scheduler/actions/reclaim/reclaim.go#L86-L220, reclaim firstly pop a queue and a job, but at line 110-112, 116-118, 121-124，126-129，https://github.com/volcano-sh/volcano/blob/0843c0d33fccdb85439dc7086dd7cea061070901/pkg/scheduler/actions/reclaim/reclaim.go#L110-L112 https://github.com/volcano-sh/volcano/blob/0843c0d33fccdb85439dc7086dd7cea061070901/pkg/scheduler/actions/reclaim/reclaim.go#L116-L119 https://github.com/volcano-sh/volcano/blob/0843c0d33fccdb85439dc7086dd7cea061070901/pkg/scheduler/actions/reclaim/reclaim.go#L121-L124 https://github.com/volcano-sh/volcano/blob/0843c0d33fccdb85439dc7086dd7cea061070901/pkg/scheduler/actions/reclaim/reclaim.go#L126-L129 
If the task fails to filter in allocatable, Preemptive, PrePredicateFn, the queue will never be pushed back, but whether if other tasks in same jobs or other jobs in same queue can reclaim resources, I'm little bit confused about the logic here, I think when I need to implement the `preemptionPolicy`, there is need to push the job and queue back to allow others to continue reclaiming.

You can also see the logic in allocate:
https://github.com/volcano-sh/volcano/blob/0843c0d33fccdb85439dc7086dd7cea061070901/pkg/scheduler/actions/allocate/allocate.go#L192-L199, at line 192, it wraps with `!tasks.Empty()` loop, so if the task fails to filter in allocatable, it's reasonable to continue here, allow other task to continue allocate. 

------------------------
9.20 updated:
After discussing with @Monokaix @hwdef @lowang-bh , we think there are some problems in reclaim action, need to refactor the reclaim action

Monokaix:
It's a good catch!

JesseStutler:
/priority high

[...]

Issue #3642, opened 2024-07-30
supoort preemptionPolicy


hwdef:
#### What would you like to be added:

priorityclass has a field named preemptionPolicy.
https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#non-preempting-priority-class

volcano didn't implement its logic.

#### solution

In volcano's cache, priorityclas is just a number, so we need to modify the priorityclass in the cache and then implement preemptionpolicy-related logic.

https://github.com/volcano-sh/volcano/blob/c91e42e07b55fbc0686095aedb0f9436b1991163/pkg/scheduler/api/job_info.go#L121

 - [ ] Modify priorityclass in volcano cache, add more information, such as preemptionPolicy
 - [ ] add logic for preemptionpolicy

googs1025:
I will find time to research and try to implement

googs1025:
@Sianao Hello, I am currently working on this issue. Can you find other issues? Or do you have any special requirements for this issue?

Sianao:
> @Sianao Hello, I am currently working on this issue. Can you find other issues? Or do you have any special requirements for this issue?

I unassign it already

googs1025:
@Sianao Do you want to try this issue? I'm currently busy with other projects and don't have time to do this. :) 
If you want, feel free to take it. I will review and help in the PR
/unassign

Sianao:
Sir , I m busy , too .

JesseStutler:
/assgin

I'm glad to work on this.

JesseStutler:
@hwdef Hi, Do we only need to support preemptionPolicy at the pod level? Not to add more fields in job_info.  So that we only need to add some logic in preempting, like this:
![image](https://github.com/user-attachments/assets/70058efc-1477-4b5c-8cd5-11974d78ce52)
If the preemptionPolicy is 'Never', not allowing the pod to preempt others, directly skip this pod.

Monokaix:
We should also consider this case: https://github.com/volcano-sh/volcano/issues/3512 @hwdef @JesseStutler

lowang-bh:
I think we can keep the property of priority as a number, and add extra info about `preemptionPolicy` in taskInfo or JobInfor when parse it.

JesseStutler:
[...]

PR #3739, opened 2024-09-19
feature:add preemptionpolicy in preempt and reclaim

JesseStutler:
ref: #3642 ,#3738

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -19,6 +19,7 @@ package preempt
 import (
 	"fmt"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
 	"volcano.sh/volcano/pkg/scheduler/api"
@@ -231,6 +232,11 @@ func (pmpt *Action) preempt(
 	filter func(*api.TaskInfo) bool,
 	predicateHelper util.PredicateHelper,
 ) (bool, error) {
+	// Check whether the task is eligible to preempt others, e.g., check preemptionPolicy is `Never` or not
+	if err := pmpt.taskEligibleToPreempt(preemptor); err != nil {
+		return false, err
+	}
+
 	assigned := false
 
 	if err := ssn.PrePredicateFn(preemptor); err != nil {
@@ -326,6 +332,14 @@ func (pmpt *Action) preempt(
 	}
 
 	return assigned, nil
+}
+
+func (pmpt *Action) taskEligibleToPreempt(preemptor *api.TaskInfo) error {
+	if preemptor.Pod.Spec.PreemptionPolicy != nil && *preemptor.Pod.Spec.PreemptionPolicy == v1.PreemptNever {
+		return fmt.Errorf("not eligible to preempt other tasks due to preemptionPolicy is Never")
+	}
+
+	return nil
 }
 
 func victimTasks(ssn *framework.Session) {
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -17,6 +17,7 @@ limitations under the License.
 package reclaim
 
 import (
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
 	"volcano.sh/volcano/pkg/scheduler/api"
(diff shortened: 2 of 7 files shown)


JesseStutler:
# Verification
## test-1
Create two jobs, job-a and job-b, where the PriorityClass specified by job-b has a higher priority than the PriorityClass specified by job-a, but the PriorityClass specified by job-b comes with `PreemptionPolicy: Never`.
```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 200
preemptionPolicy: Never
```

The Pod created by job-b does not specify a `PriorityClassName`, so the PriorityClass specified by job is used.：


The Pod created by job-b fails to preempt because PreemptionPolicy is set to true:：
image

## test-2
Create two jobs, job-a and job-c, job-c has a specified PriorityClass without the `PreemptionPolicy` field, but the task has a different PriorityClass with the `PreemptionPolicy` field specified, and job-a remains unchanged.

job-c yaml:
```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 200
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: cover
value: 200
preemptionPolicy: Never
---
kind: Job
metadata:
  name: job-c
spec:
  schedulerName: volcano
  minAvailable: 4
  tasks:
    - replicas: 4
      name: "test"
      template:
        spec:
          priorityClassName: cover
          containers:
            - image: alpine
              command: ["/bin/sh", "-c", "sleep 6000"]
              imagePullPolicy: IfNotPresent
              name: running
              resources:
                requests:
                  cpu: 3900m
          restartPolicy: OnFailure
```
The PriorityClassName specified by job-c's pod is the task template's
image
Preemption failure log:
image

## test-3
Submit two jobs, one job-a and one job-b, in two separate queues, first submit job-a to fill up the cluster resources, then submit job-b, job-a needs to reclaim half of the resources to job-b, but because the PriorityClass specified by job-b specifies `PreemptionPolicy` as Never, job-a cannot reclaim job-a's resources, so job-b's pod all are pending. 
image

hwdef (commented):
/lgtm

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -231,6 +232,10 @@ func (pmpt *Action) preempt(
 	filter func(*api.TaskInfo) bool,
 	predicateHelper util.PredicateHelper,
 ) (bool, error) {
+	if preemptor.Pod.Spec.PreemptionPolicy != nil && *preemptor.Pod.Spec.PreemptionPolicy == v1.PreemptNever {
```
Monokaix, line 235: better wrap a func here.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -113,6 +114,15 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			task = tasks.Pop().(*api.TaskInfo)
 		}
 
+		if task.Pod.Spec.PreemptionPolicy != nil && *task.Pod.Spec.PreemptionPolicy == v1.PreemptNever {
```
Monokaix, line 117: Is there any difference with kube-scheduler, such as is  there event reported to inform user.

JesseStutler: no, it only record the log, the codes are:
```go
if ok, msg := ev.PodEligibleToPreemptOthers(ctx, pod, nominatedNodeStatus); !ok {
		logger.V(5).Info("Pod is not eligible for preemption", "pod", klog.KObj(pod), "reason", msg)
		return nil, framework.NewStatus(framework.Unschedulable, msg)
}
```
```go
func (pl *DefaultPreemption) PodEligibleToPreemptOthers(_ context.Context, pod *v1.Pod, nominatedNodeStatus *framework.Status) (bool, string) {
	if pod.Spec.PreemptionPolicy != nil && *pod.Spec.PreemptionPolicy == v1.PreemptNever {
		return false, "not eligible due to preemptionPolicy=Never."
	}

	nodeInfos := pl.fh.SnapshotSharedLister().NodeInfos()
	nomNodeName := pod.Status.NominatedNodeName
	if len(nomNodeName) > 0 {
		// If the pod's nominated node is considered as UnschedulableAndUnresolvable by the filters,
		// then the pod should be considered for preempting again.
		if nominatedNodeStatus.Code() == framework.UnschedulableAndUnresolvable {
			return true, ""
		}

		if nodeInfo, _ := nodeInfos.Get(nomNodeName); nodeInfo != nil {
			podPriority := corev1helpers.PodPriority(pod)
			for _, p := range nodeInfo.Pods {
				if corev1helpers.PodPriority(p.Pod) < podPriority && podTerminatingByPreemption(p.Pod) {
					// There is a terminating pod on the nominated node.
					return false, "not eligible due to a terminating pod on the nominated node."
				}
			}
		}
	}
	return true, ""
}
```

PR #3742, opened 2024-09-21
controllers: fix queue controller syncQueue bug

sceneryback:
Sometimes we may encounter podGroupNotFound errors when doing syncQueue, maybe due to misses of podGroup deletion events. In this case, we should try to re-enqueue the request and update local cache or the queue status will be left unchanged forever, e.g.    
WeChatWorkScreenshot_a653b381-82d4-4b66-a40f-6d331e9881c6

### DIFF
--- a/pkg/controllers/queue/queue_controller_action.go
+++ b/pkg/controllers/queue/queue_controller_action.go
@@ -22,6 +22,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -42,10 +43,17 @@ func (c *queuecontroller) syncQueue(queue *schedulingv1beta1.Queue, updateStateF
 		// Ignore error here, tt can not occur.
 		ns, name, _ := cache.SplitMetaNamespaceKey(pgKey)
 
-		// TODO: check NotFound error and sync local cache.
 		pg, err := c.pgLister.PodGroups(ns).Get(name)
 		if err != nil {
-			return err
+			if !errors.IsNotFound(err) {
+				return err
+			}
+
+			klog.V(4).Infof("The podGroup %s is not found, skip it and continue to sync cache", pgKey)
+			c.pgMutex.Lock()
+			delete(c.podGroups[queue.Name], pgKey)
+			c.pgMutex.Unlock()
+			continue
 		}
 
 		switch pg.Status.Phase {
--- a/pkg/controllers/queue/queue_controller_test.go
+++ b/pkg/controllers/queue/queue_controller_test.go
@@ -19,8 +19,10 @@ package queue
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"testing"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeclient "k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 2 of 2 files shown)


pkg/controllers/queue/queue_controller_action.go
```diff
@@ -42,9 +61,11 @@ func (c *queuecontroller) syncQueue(queue *schedulingv1beta1.Queue, updateStateF
 		// Ignore error here, tt can not occur.
 		ns, name, _ := cache.SplitMetaNamespaceKey(pgKey)
 
-		// TODO: check NotFound error and sync local cache.
 		pg, err := c.pgLister.PodGroups(ns).Get(name)
 		if err != nil {
+			if errors.IsNotFound(err) {
```
Monokaix, line 66: The err will also be returned and re-enqueued, and next procesee will continue handle the req, how about returned nil when pg not existed?
```go
		pg, err := c.pgLister.PodGroups(ns).Get(name)
		if err != nil {
			if !errors.IsNotFound(err) {
				return err
			}
			c.pgMutex.Lock()
			defer c.pgMutex.Unlock()

			delete(c.podGroups[queue], pgKey)
			return nil
		}
```

sceneryback: Hi @Monokaix Thanks for your reviewing and sample code. In this case, I think we should 1) pop this error, so the error will be recorded and noticed by developers, and 2) re-enqueue the request, since the `syncQueue` process is actually not finished

Monokaix: That's a good point, but what about just `continue` if pg not exists? And the queue will still be processed, re-enqueue seems is not so necessary.

sceneryback: If we update cache silently and continue the process, the error will not be recorded in `handleQueueErr`, so the developers will not notice that in the events of queue. Sometimes we may need to monitor these events and do some alerting work

Monokaix: Here we should make clear whether pg not exists is a normal case or else, if pg not exists, it's a normal case and we don't need to record an event and just print the err is ok, if you think it's an abnormal case, can you support more detail about why the pg is deleted when syncQueue?

sceneryback: @Monokaix I think it is abnormal, because we have already added the event listener for pg events, including `Delete`:  
```
pgInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc:    c.addPodGroup,
		UpdateFunc: c.updatePodGroup,
		DeleteFunc: c.deletePodGroup,
	})
```
But it still missed this event and the `podGroups` cache is not updated. May be due to network issue. I think the case is rare and not easy to reproduce, but the possibility exists, as can be seen from my attached picture

Monokaix: We can assume that `deletePodGroup` is a normal behavior，and what abnormal is that we didn't get the latest delete event through informer，based on this，we can skip process the pg because it truly not exists, and this will not block queue process and the queue will be synced.

Monokaix: Whether get the event is not so important because the pg is not existed, and we just guarantee the queue is synced   is ok.

sceneryback: @Monokaix Thanks for your advice, I update code to not pop the error and sync cache immediately, please have a look

pkg/controllers/queue/queue_controller_action.go
```diff
@@ -42,10 +43,17 @@ func (c *queuecontroller) syncQueue(queue *schedulingv1beta1.Queue, updateStateF
...
 
-		// TODO: check NotFound error and sync local cache.
 		pg, err := c.pgLister.PodGroups(ns).Get(name)
 		if err != nil {
-			return err
+			if !errors.IsNotFound(err) {
+				return err
+			}
+
+			klog.V(4).Infof("sync local cache since podGroup %s not found.", pgKey)
```
Monokaix, line 52: >>> klog.V(4).Infof("sync local cache since podGroup %s not found.", pgKey)
klog.V(4).Infof("The podGroup %s is not found, skip it and continue to syne cache", pgKey)

pkg/controllers/queue/queue_controller_test.go
```diff
@@ -242,55 +244,125 @@ func TestSyncQueue(t *testing.T) {
...
+			c.podGroups[testcase.podGroups[j].Spec.Queue][key] = struct{}{}
+
+			c.pgInformer.Informer().GetIndexer().Add(testcase.podGroups[j])
+		}
 
-		c.pgInformer.Informer().GetIndexer().Add(testcase.podGroup)
 		c.queueInformer.Informer().GetIndexer().Add(testcase.queue)
 		c.vcClient.SchedulingV1beta1().Queues().Create(context.TODO(), testcase.queue, metav1.CreateOptions{})
 
+		for j := range testcase.delayDeletedPodGroups {
```
Monokaix, line 355: Repeat add and delete pg in informer is not graceful, you can add two fieleds in testCase like pginInformer and pgInCache to distinguish them: )

Monokaix:
Please also squash to one commit and sign off with `git commit -s`.

sceneryback:
> Please also squash to one commit and sign off with `git commit -s`.

@Monokaix Thanks and done

sceneryback:
hi @Monokaix just rebased the branch with master, please have a look

Monokaix:
Hi, please check the failed ci.

Monokaix:
cc @lowang-bh @hwdef

Issue #3590, opened 2024-07-15
Support Hierarchical Queue on Capacity Plugin


Rui-Gan:
#### What would you like to be added:
Description: design and implement hierarchical queue management capabilities on the Capacity plugin, to better align with real-world queue usage scenarios and further improve resource utilization in task scheduling.

#### Task list:

- Add design doc. #3591 
- hierarchical queue implementation in this repo. #3743 

Mentor(s):
@hwdef

PR #3743, opened 2024-09-23
feat: add hierarchical queues for capacity plugin

Rui-Gan:


hwdef:
ref: #3590

hwdef:
/assign @Monokaix @lowang-bh @shinytang6 
Please take a look 👀

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -220,133 +402,387 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
 
-			klog.V(4).Infof("Capacity AllocateFunc: task <%v/%v>, resreq <%v>,  share <%v>",
-				event.Task.Namespace, event.Task.Name, event.Task.Resreq, attr.share)
-		},
-		DeallocateFunc: func(event *framework.Event) {
-			job := ssn.Jobs[event.Task.Job]
-			attr := cp.queueOpts[job.Queue]
-			attr.allocated.Sub(event.Task.Resreq)
-			metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory)
+		childAttr.deserved = helpers.Max(childAttr.deserved, childAttr.guarantee)
```
Monokaix, line 703: Seems every queue's deserved value is calculated in `openSession`, why  re-calculate here?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -220,133 +402,387 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		totalDeserved.Sub(oldDeserved).Add(childAttr.deserved)
+	}
 
-			cp.updateShare(attr)
+	// Check if the parent queue's deserved resources are less than the total deserved resources of child queues
+	if attr.deserved.LessPartly(totalDeserved, api.Zero) {
```
Monokaix, line 708: `openSession` will not retuen err when there is an err of a plugin , so even thougt we return an err here, the `openSession` will not end and the plugin will contiune its logic, so does this check really take effect?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -220,133 +402,387 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		return fmt.Errorf("deserved resources of queue <%s> are less than the sum of its child queues' deserved resources", attr.name)
```
Monokaix, line 709: Seems we should also check that one child's capability should not exceed its parent's capability when parent's capability is set.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -75,11 +88,180 @@ func (cp *capacityPlugin) Name() string {
...
+
 func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 	// Prepare scheduling data for this session.
 	cp.totalResource.Add(ssn.TotalResource)
 
 	klog.V(4).Infof("The total resource is <%v>", cp.totalResource)
+
+	hierarchyEnabled := cp.HierarchyEnabled(ssn)
+	if hierarchyEnabled {
+		cp.onSessionOpenWithHierarchy(ssn)
```
Monokaix, line 112: The `onSessionOpenWithHierarchy` and `onSessionOpenWithoutHierarchy` are just a subset of `openSession`, and there are some other common logic in `openSession`, so I suggest rename these func name cause it's not a complete sessionOpen logic.

pkg/webhooks/admission/queues/mutate/mutate_hierarchical_queue.go
```diff
@@ -0,0 +1,132 @@
...
+	klog.V(3).Infof("Mutating hierarchical queue %s for operation %s.", ar.Request.Name, ar.Request.Operation)
+
+	queue, err := schema.DecodeQueue(ar.Request.Object, ar.Request.Resource)
+	if err != nil {
+		return util.ToAdmissionResponse(err)
+	}
+
+	switch ar.Request.Operation {
+	case admissionv1.Create, admissionv1.Update:
+		err = closeQueue(queue)
```
Monokaix, line 79: openqueue logic is missed.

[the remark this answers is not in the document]
Monokaix: Also seems that we didn't check whether hierarchical queue is enable of capacity plugin and mutate it directly, which is also inconsistent.

pkg/webhooks/admission/queues/validate/validate_hierarchical_queue.go
```diff
@@ -0,0 +1,133 @@
...
+						Resources:   []string{"queues"},
+					},
+				},
+			},
+		}},
+	},
+}
+
+// HierarchicalQueues validate queues when capacity plugin enables hierarchy.
+func HierarchicalQueues(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
```
Monokaix, line 62: What's the consideration here?
Here will also cause the problem that queue controller keep retrying.

Monokaix:
Is somewhere validate that job can only be submitted to leaf node?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -791,21 +841,19 @@ func (ssn *Session) NodeOrderReduceFn(task *api.TaskInfo, pluginNodeScoreMap map
 // BuildVictimsPriorityQueue returns a priority queue with victims sorted by:
 // if victims has same job id, sorted by !ssn.TaskOrderFn
 // if victims has different job id, sorted by !ssn.JobOrderFn
-func (ssn *Session) BuildVictimsPriorityQueue(victims []*api.TaskInfo) *util.PriorityQueue {
+func (ssn *Session) BuildVictimsPriorityQueue(victims []*api.TaskInfo, preemptor *api.TaskInfo) *util.PriorityQueue {
```
Monokaix, line 844: The `AddQueueOrderFn` has already defined the hierarchical queue's order, so it seems that VictimJobOrderFn and VictimTaskOrderFn not necessary？just define the job order and task order is ok?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -220,133 +402,387 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			return
+		}
+	}
+
+	for _, job := range ssn.Jobs {
+		klog.V(4).Infof("Considering Job <%s/%s>.", job.Namespace, job.Name)
+		attr := cp.queueOpts[job.Queue]
+		if len(attr.children) > 0 {
+			klog.Errorf("The Queue <%s> of Job <%s/%s> is not leaf queue", attr.name, job.Namespace, job.Name)
+			return
```
TaiPark, line 437: A job being unexpectedly submitted to the parent queue may cause the main process of the current `onSessionOpenWithHierarchy()` to return prematurely. Is the consequence too severe? 

The restriction that "a job cannot be submitted to the parent queue" should perhaps be checked elsewhere, so that the impact of any unexpected job can be contained within the job itself.

hwdef:
Please add some E2E test for this function. Because this function is very important.

pkg/webhooks/admission/queues/mutate/mutate_hierarchical_queue.go
```diff
@@ -0,0 +1,132 @@
...
+			Result:  &metav1.Status{Message: err.Error()},
+		}
+	}
+
+	return &admissionv1.AdmissionResponse{
+		Allowed: true,
+	}
+}
+
+func closeQueue(queue *schedulingv1beta1.Queue) error {
```
Monokaix, line 97: Maybe put these logic to volcano controller is better.

Rui-Gan:
> Please add some E2E test for this function. Because this function is very important.

I will add it later

pkg/webhooks/admission/queues/mutate/mutate_hierarchical_queue.go
Rui-Gan: I also think that the logic for open/close queue can be added to the controller, but do you think it's necessary to include this logic in the webhook as well for fallback handling

pkg/scheduler/plugins/capacity/capacity.go
Rui-Gan: I have already added this logic in the validate webhook of the job to ensure that jobs can only be submitted to leaf queues.

Rui-Gan:
> Is somewhere validate that job can only be submitted to leaf node?

I have already added this logic in the validate webhook of the job to ensure that jobs can only be submitted to leaf queues.

pkg/webhooks/admission/queues/mutate/mutate_hierarchical_queue.go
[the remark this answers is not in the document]
Rui-Gan: For this scenario, I have conducted tests. It indeed results in the queue controller repeatedly processing the queue until it reaches the maximum number of attempts and stops retrying.
As mentioned in your comment below, I agree that we can consider adding the logic for opening/closing queues in a hierarchical queue structure within the queue controller. However, we can discuss whether it is still necessary to include the relevant logic in the webhook as a fallback handling mechanism.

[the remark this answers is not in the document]
Rui-Gan: > Also seems that we didn't check whether hierarchical queue is enable of capacity plugin and mutate it directly, which is also inconsistent.

In the latest implementation, if the hierarchical queue feature of the capacity plugin needs to be enabled, the webhook's configuration file must set "enableHierarchyCapacity" to true. What do you think about this approach, or do you have any suggestions?

pkg/webhooks/admission/queues/validate/validate_hierarchical_queue.go
Rui-Gan: This won't cause the queue controller to keep retrying. The issue you mentioned only occurs when performing operations related to opening or closing queues. However, there might be some scenarios that I haven't considered. Could you provide a specific example to illustrate such a case?

pkg/scheduler/plugins/capacity/capacity.go
Rui-Gan: The validation for ensuring a child's capability does not exceed its parent's set capability has already been implemented in the code at line 657.

pkg/scheduler/plugins/capacity/capacity.go
Rui-Gan: In the latest implementation, if an error occurs during openSession, the injected functions will directly return the logic related to rejecting the scheduling. For example, in ssn.AddJobEnqueueableFn, it immediately returns a rejection, preventing the job from being enqueued. What do you think about this approach?

pkg/scheduler/plugins/capacity/capacity.go
Rui-Gan: In the openSession function, I didn't calculate the deserved value for each queue. I'm not quite sure I understand what you mean. Could you please provide more specific details to clarify your point?

pkg/webhooks/admission/jobs/validate/admit_job.go
```diff
@@ -229,6 +233,31 @@ func validateJobCreate(job *v1alpha1.Job, reviewResponse *admissionv1.AdmissionR
 	} else if queue.Status.State != schedulingv1beta1.QueueStateOpen {
 		msg += fmt.Sprintf(" can only submit job to queue with state `Open`, "+
 			"queue `%s` status is `%s`;", queue.Name, queue.Status.State)
+	} else if config.ConfigData != nil {
+		config.ConfigData.Lock()
+		enableHierarchyCapacity := config.ConfigData.EnableHierarchyCapacity
+		config.ConfigData.Unlock()
+		if enableHierarchyCapacity {
+			if queue.Name == "root" {
+				msg += fmt.Sprintf(" can not submit job to root queue;")
```
TaiPark, line 242: In a scenario where there is only one root queue, can the root queue be considered a leaf queue and allocated jobs on it?

Rui-Gan: Even if the user has not created any new queues, the system will still have two queues called "root" and "default". The "root" queue is not a leaf queue because the parent queue of "default" is "root".

hwdef:
Please consider this scenario:
1. Use default config create a cluster (./hack/local-up-volcano.sh)
2. This configuration uses the proportion plugin by default and automatically creates a default queue
3. Switch to the capacity plugin and enable hierarchical queues

I believe this is how most existing users use hierarchical queues.

result:
1. root queue will not be created.
2. scheduler panic:
```
2024/10/03 14:13:01 maxprocs: Leaving GOMAXPROCS=16: CPU quota undefined
I1003 14:13:01.520160       1 flags.go:57] FLAG: --add-dir-header="false"
I1003 14:13:01.520174       1 flags.go:57] FLAG: --alsologtostderr="false"
I1003 14:13:01.520176       1 flags.go:57] FLAG: --ca-cert-file=""
I1003 14:13:01.520179       1 flags.go:57] FLAG: --cache-dump-dir="/tmp"
I1003 14:13:01.520181       1 flags.go:57] FLAG: --cache-dumper="true"
I1003 14:13:01.520183       1 flags.go:57] FLAG: --csi-storage="false"
I1003 14:13:01.520185       1 flags.go:57] FLAG: --default-queue="default"
I1003 14:13:01.520187       1 flags.go:57] FLAG: --enable-healthz="true"
I1003 14:13:01.520189       1 flags.go:57] FLAG: --enable-metrics="true"
I1003 14:13:01.520192       1 flags.go:57] FLAG: --feature-gates=""
I1003 14:13:01.520196       1 flags.go:57] FLAG: --healthz-address=":11251"
I1003 14:13:01.520199       1 flags.go:57] FLAG: --ignored-provisioners="[]"
I1003 14:13:01.520210       1 flags.go:57] FLAG: --kube-api-burst="2000"
I1003 14:13:01.520214       1 flags.go:57] FLAG: --kube-api-qps="2000"
I1003 14:13:01.520218       1 flags.go:57] FLAG: --kubeconfig=""
I1003 14:13:01.520221       1 flags.go:57] FLAG: --leader-elect="false"
I1003 14:13:01.520223       1 flags.go:57] FLAG: --leader-elect-lease-duration="15s"
I1003 14:13:01.520229       1 flags.go:57] FLAG: --leader-elect-renew-deadline="10s"
I1003 14:13:01.520232       1 flags.go:57] FLAG: --leader-elect-resource-lock="leases"
I1003 14:13:01.520234       1 flags.go:57] FLAG: --leader-elect-resource-name="volcano"
I1003 14:13:01.520237       1 flags.go:57] FLAG: --leader-elect-resource-namespace="volcano-system"
I1003 14:13:01.520239       1 flags.go:57] FLAG: --leader-elect-retry-period="2s"
I1003 14:13:01.520242       1 flags.go:57] FLAG: --listen-address=":8080"
I1003 14:13:01.520244       1 flags.go:57] FLAG: --lock-object-namespace="volcano-system"
I1003 14:13:01.520247       1 flags.go:57] FLAG: --log-backtrace-at=":0"
I1003 14:13:01.520251       1 flags.go:57] FLAG: --log-dir=""
I1003 14:13:01.520255       1 flags.go:57] FLAG: --log-file=""
I1003 14:13:01.520259       1 flags.go:57] FLAG: --log-file-max-size="1800"
I1003 14:13:01.520263       1 flags.go:57] FLAG: --log-flush-frequency="5s"
I1003 14:13:01.520267       1 flags.go:57] FLAG: --logtostderr="true"
I1003 14:13:01.520270       1 flags.go:57] FLAG: --master=""
I1003 14:13:01.520274       1 flags.go:57] FLAG: --minimum-feasible-nodes="100"
I1003 14:13:01.520284       1 flags.go:57] FLAG: --minimum-percentage-nodes-to-find="5"
I1003 14:13:01.520287       1 flags.go:57] FLAG: --node-selector="[]"
I1003 14:13:01.520295       1 flags.go:57] FLAG: --node-worker-threads="20"
I1003 14:13:01.520303       1 flags.go:57] FLAG: --one-output="false"
I1003 14:13:01.520306       1 flags.go:57] FLAG: --percentage-nodes-to-find="0"
I1003 14:13:01.520309       1 flags.go:57] FLAG: --plugins-dir=""
I1003 14:13:01.520312       1 flags.go:57] FLAG: --priority-class="true"
I1003 14:13:01.520315       1 flags.go:57] FLAG: --schedule-period="1s"
I1003 14:13:01.520319       1 flags.go:57] FLAG: --scheduler-conf="/volcano.scheduler/volcano-scheduler.conf"
I1003 14:13:01.520322       1 flags.go:57] FLAG: --scheduler-name="[volcano]"
I1003 14:13:01.520328       1 flags.go:57] FLAG: --skip-headers="false"
I1003 14:13:01.520331       1 flags.go:57] FLAG: --skip-log-headers="false"
I1003 14:13:01.520333       1 flags.go:57] FLAG: --stderrthreshold="2"
I1003 14:13:01.520338       1 flags.go:57] FLAG: --tls-cert-file=""
I1003 14:13:01.520340       1 flags.go:57] FLAG: --tls-private-key-file=""
I1003 14:13:01.520343       1 flags.go:57] FLAG: --v="3"
I1003 14:13:01.520348       1 flags.go:57] FLAG: --version="false"
I1003 14:13:01.520350       1 flags.go:57] FLAG: --vmodule=""
W1003 14:13:01.520367       1 client_config.go:659] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.


panic: failed init default queue, with err: admission webhook "mutatequeue.volcano.sh" denied the request: failed to get parent queue of open queue default: queues.scheduling.volcano.sh "root" not found

goroutine 1 [running]:
volcano.sh/volcano/pkg/scheduler/cache.newDefaultQueue({0x25de198, 0xc0000e3620}, {0x22a7c80, 0x7})
        /go/src/volcano.sh/volcano/pkg/scheduler/cache/cache.go:513 +0x1d3
volcano.sh/volcano/pkg/scheduler/cache.newSchedulerCache(0xc00017d688, {0xc00005aad0, 0x1, 0x1}, {0x22a7c80, 0x7}, {0x0, 0x0, 0x0}, 0x14, ...)
        /go/src/volcano.sh/volcano/pkg/scheduler/cache/cache.go:532 +0xfd
volcano.sh/volcano/pkg/scheduler/cache.New(...)
        /go/src/volcano.sh/volcano/pkg/scheduler/cache/cache.go:92
volcano.sh/volcano/pkg/scheduler.NewScheduler(0xc00017d688?, 0xc0000ec000)
        /go/src/volcano.sh/volcano/pkg/scheduler/scheduler.go:70 +0xeb
volcano.sh/volcano/cmd/scheduler/app.Run(0xc0000ec000)
        /go/src/volcano.sh/volcano/cmd/scheduler/app/server.go:71 +0x1a5
main.main()
        /go/src/volcano.sh/volcano/cmd/scheduler/main.go:86 +0x325

```

installer/helm/chart/volcano/config/volcano-admission.conf
```diff
@@ -1,3 +1,4 @@
+#enableHierarchyCapacity: true
```
googs1025, line 1: I would like to ask if this new feature needs a feature gate? I have not thoroughly understood this feature, and I am not sure if it will be stable before the next version, so I would tend to discuss whether a feature gate is needed to avoid this problem.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -75,11 +88,200 @@ func (cp *capacityPlugin) Name() string {
...
+}
+
 func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 	// Prepare scheduling data for this session.
 	cp.totalResource.Add(ssn.TotalResource)
 
 	klog.V(4).Infof("The total resource is <%v>", cp.totalResource)
+
+	hierarchyEnabled := cp.HierarchyEnabled(ssn)
+	readyToschedule := true
```
TaiPark, line 111: `readyToSchedule` maybe better

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -75,11 +88,200 @@ func (cp *capacityPlugin) Name() string {
...
+	if hierarchyEnabled {
+		readyToschedule = cp.buildHierarchicalQueueAttrs(ssn)
+	} else {
+		cp.buildQueueAttrs(ssn)
+	}
```
hwdef, line 116: Will placing these logic in onsessionopen have any impact on performance?

hwdef:
I got some errors when create a vcjob

```
Failed to create root queue, error: queues.scheduling.volcano.sh "root" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -220,133 +434,339 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			queue = &vcv1beta1.Queue{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: string(cp.rootQueue),
+				},
+				Spec: vcv1beta1.QueueSpec{},
+			}
+			cp.makeRootQueue(queue)
+
+			return retry.OnError(wait.Backoff{
+				Steps:    60,
```
TaiPark, line 725: The default `defaultSchedulerPeriod` for `runOnce()` is 1 second. Is the number of `Steps` set for retries too large here?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -220,133 +436,363 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		_, err := vcClient.SchedulingV1beta1().Queues().Update(context.TODO(), queue, metav1.UpdateOptions{})
+		return err
 	})
 }
 
-func (cp *capacityPlugin) OnSessionClose(ssn *framework.Session) {
-	cp.totalResource = nil
-	cp.totalGuarantee = nil
-	cp.queueOpts = nil
+func (cp *capacityPlugin) makeRootQueue(queue *vcv1beta1.Queue) {
```
TaiPark, line 780: Why do we need to update `rootQueue.Deserved` and `rootQueue.Guarantee` here?
It seems that we will do
```go
		attr.guarantee = totalGuarantee
		cp.totalGuarantee = totalGuarantee
		attr.realCapability = cp.totalResource
		attr.deserved = cp.totalResource
	}
```
for root queue in each OpenSession -> `checkHierarchicalQueue()`

pkg/scheduler/cache/cache.go
```diff
@@ -941,6 +941,11 @@ func (sc *SchedulerCache) Client() kubernetes.Interface {
 	return sc.kubeClient
 }
 
+// Client returns the volcano clientSet
```
TaiPark, line 944: VCClient maybe better
```suggestion
// VCClient returns the volcano clientSet
```

pkg/scheduler/api/types.go
```diff
@@ -116,6 +116,9 @@ type LessFn func(interface{}, interface{}) bool
 // CompareFn is the func declaration used by sort or priority queue.
 type CompareFn func(interface{}, interface{}) int
 
+// CompareFn is the func declaration used by sort or priority victims.
```
TaiPark, line 119: VictimCompareFn maybe better
```suggestion
// VictimCompareFn is the func declaration used by sort or priority victims.
```

pkg/scheduler/framework/session_plugins.go
```diff
@@ -35,6 +35,11 @@ func (ssn *Session) AddQueueOrderFn(name string, qf api.CompareFn) {
 	ssn.queueOrderFns[name] = qf
 }
 
+// AddVictimQueueOrderFn add victimjoborder function
```
TaiPark, line 38:
```suggestion
// AddVictimQueueOrderFn add victim job order function
```

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -220,133 +436,363 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		metrics.UpdateQueueDeserved(attr.name, attr.deserved.MilliCPU, attr.deserved.Memory)
+		metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory)
+		metrics.UpdateQueueRequest(attr.name, attr.request.MilliCPU, attr.request.Memory)
+		metrics.UpdateQueuePodGroupInqueueCount(attr.name, queue.Queue.Status.Inqueue)
+		metrics.UpdateQueuePodGroupPendingCount(attr.name, queue.Queue.Status.Pending)
+		metrics.UpdateQueuePodGroupRunningCount(attr.name, queue.Queue.Status.Running)
+		metrics.UpdateQueuePodGroupUnknownCount(attr.name, queue.Queue.Status.Unknown)
+	}
+
+	ssn.AddQueueOrderFn(cp.Name(), func(l, r interface{}) int {
```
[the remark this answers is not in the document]
Rui-Gan: If `buildHierarchicalQueueAttrs()` fails, I'd prefer to disable the capacity plugin as much as possible and not schedule tasks. If the current `buildHierarchicalQueueAttrs()` fails, not registering `QueueOrderFn()` should be more in line with this logic.

Additionally, the reason `ssn.AddQueueOrderFn()` and `ssn.AddVictimQueueOrderFn()` are not implemented in `OnSessionOpen()` is that these two functions have very different logic in `buildHierarchicalQueueAttrs()` and `buildQueueAttrs()`, and `ssn.AddVictimQueueOrderFn()` should not be registered in `buildQueueAttrs()`.

TaiPark:
In the scenario of scheduling directly with `PodGroup` (without `Volcano Job`), the logic for checking whether a Parent Queue is unable to submit is missing.

Maybe we need a `webhooks.admission.podgroups.validate` to check this?

pkg/scheduler/plugins/capacity/capacity.go
Rui-Gan: The logic in `checkHierarchicalQueue()` is to update the `queueAttr` of the root queue, which is mainly to preserve the attributes of the root queue during the calculation process of the capacity plugin. The `makeRootQueue` function is to supplement the attributes in the root queue CR, thereby updating the queue CR.

Rui-Gan:
> In the scenario of scheduling directly with `PodGroup` (without `Volcano Job`), the logic for checking whether a Parent Queue is unable to submit is missing.
> 
> Maybe we need a `webhooks.admission.podgroups.validate` to check this?

Thank you for your suggestion. I will add the missing logic.

pkg/scheduler/plugins/capacity/capacity.go
TaiPark: Ok, it's good for other plugins to use queue CR in the future :laughing:

pkg/controllers/queue/queue_controller_action.go
```diff
@@ -173,3 +191,56 @@ func (c *queuecontroller) closeQueue(queue *schedulingv1beta1.Queue, updateState
 
 	return nil
 }
+
+func (c *queuecontroller) openHierarchicalQueue(queue *schedulingv1beta1.Queue) (bool, error) {
```
hwdef, line 195: The function of this function is more like checking. Its name represents an action. I think the name should be changed.
Or is there any functionality missing here?
If a queue is open, it should open its childqueue if the childqueue is closed because the parent queue is closed.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -801,8 +823,10 @@ func (ssn *Session) BuildVictimsPriorityQueue(victims []*api.TaskInfo) *util.Pri
 
 		lvJob, lvJobFound := ssn.Jobs[lv.Job]
 		rvJob, rvJobFound := ssn.Jobs[rv.Job]
-		if lvJobFound && rvJobFound && lvJob.Queue != rvJob.Queue {
-			return !ssn.QueueOrderFn(ssn.Queues[lvJob.Queue], ssn.Queues[rvJob.Queue])
+		preemptorJob, preemptorJobFound := ssn.Jobs[preemptor.Job]
+
+		if lvJobFound && rvJobFound && preemptorJobFound && lvJob.Queue != rvJob.Queue {
```
hwdef, line 828: What problem is this intended to solve?

pkg/scheduler/cache/cache.go
```diff
@@ -514,6 +514,35 @@ func newDefaultQueue(vcClient vcclient.Interface, defaultQueue string) {
 	}
 }
 
+// newRootQueue init root queue
+func newRootQueue(vcClient vcclient.Interface) {
```
hwdef, line 518: The functions of `newRootQueue` and `newDefaultQueue`  seem to be the same, but the queue names are different. Can this be achieved using one function?

pkg/webhooks/admission/jobs/validate/admit_job.go
```diff
@@ -226,9 +226,30 @@ func validateJobCreate(job *v1alpha1.Job, reviewResponse *admissionv1.AdmissionR
...
+		if queue.Status.State != schedulingv1beta1.QueueStateOpen {
+			msg += fmt.Sprintf(" can only submit job to queue with state `Open`, "+
+				"queue `%s` status is `%s`;", queue.Name, queue.Status.State)
+		}
+
+		// valiadate hierarchical queue
+		if queue.Name == "root" {
+			msg += " can not submit job to root queue;"
+		} else {
+			queueList, err := config.VolcanoClient.SchedulingV1beta1().Queues().List(context.TODO(), metav1.ListOptions{})
```
hwdef, line 239: Is it possible to use informer here to reduce the pressure on apiserver?

pkg/webhooks/admission/jobs/validate/admit_job.go
```diff
@@ -226,9 +226,30 @@ func validateJobCreate(job *v1alpha1.Job, reviewResponse *admissionv1.AdmissionR
...
+			}
+			childQueues := make([]schedulingv1beta1.Queue, 0)
+			for _, childQueue := range queueList.Items {
+				if childQueue.Spec.Parent == queue.Name {
+					childQueues = append(childQueues, childQueue)
+				}
+			}
+			if len(childQueues) > 0 {
+				msg += fmt.Sprintf(" can only submit job to leaf queue, "+"queue `%s` has %d child queues;", queue.Name, len(childQueues))
+			}
```
hwdef, line 251: I don't think it's necessary to go through all the queues here. As long as there is a childQueue, an error will be returned

pkg/scheduler/framework/session_plugins.go
Rui-Gan: The main consideration here is the need to prioritize the two queues based on the information from the preemptor queue. Priority is given to the queue that belongs to the same parent queue as the preemptor queue.

pkg/webhooks/admission/jobs/validate/admit_job.go
Rui-Gan: I have already addressed the other issues you mentioned above. Regarding this particular issue, I would prefer to return all the error information at once, so that the user can resolve all the problems in one go based on the error messages. It's also possible to modify the code as you suggest, but that might lead to the following scenario: when a user is using the system, there are actually three erroneous queues, but only one error queue is displayed each time a change is made, requiring repeated modifications.

hwdef:
Look good to me.
@Monokaix @googs1025 @TaiPark @lowang-bh Please review this again. Can we merge into this PR?

@Rui-Gan please squash your commit to one.

pkg/scheduler/cache/cache.go
```diff
@@ -485,9 +485,35 @@ func (sc *SchedulerCache) setDefaultVolumeBinder() {
 	}
 }
 
-// newDefaultQueue init default queue
-func newDefaultQueue(vcClient vcclient.Interface, defaultQueue string) {
-	reclaimable := true
+// newDefaultAndRootQueue init default queue and root queue
+func newDefaultAndRootQueue(vcClient vcclient.Interface, defaultQueue string) {
```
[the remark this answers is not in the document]
Rui-Gan: The main consideration here is to ensure the controller and webhook are compatible with both hierarchical queue enabled and disabled configurations. For users who do not use hierarchical queues, they should not be able to create or use the default and root queues. Do you have any suggestions regarding this?

Rui-Gan:
> > > In the scenario of scheduling directly with `PodGroup` (without `Volcano Job`), the logic for checking whether a Parent Queue is unable to submit is missing.
> > > Maybe we need a `webhooks.admission.podgroups.validate` to check this?
> > 
> > 
> > Thank you for your suggestion. I will add the missing logic.
> 
> Is this proposal still under consideration? I don't seem to find related logics.
Regarding this issue, I think it would be better to submit another PR to handle this after the current PR is merged. There are a couple of reasons for this:

The use case of directly using podgroup for scheduling is relatively rare, which is why there was no related validate webhook before. For example, when creating or updating the queue field of a podgroup, there is no validation to check if the queue exists. The hdrf plugin also does not perform any validation related to hierarchical queue enablement.
Currently, adding this validation logic involves a significant amount of code changes, and the e2e tests cannot pass successfully. This is because the deployment yaml files need to be modified, and when running e2e tests on GitHub, the old yaml files are used. Therefore, if we directly include this in the current PR, the e2e tests will not pass.
Based on the above considerations, I believe this logic can be implemented in a separate PR after the current PR is merged. What do you think? Is this a particularly urgent matter?

TaiPark:
> > > > In the scenario of scheduling directly with `PodGroup` (without `Volcano Job`), the logic for checking whether a Parent Queue is unable to submit is missing.
> > > > Maybe we need a `webhooks.admission.podgroups.validate` to check this?
> > > 
> > > 
> > > Thank you for your suggestion. I will add the missing logic.
> > 
> > 
> > Is this proposal still under consideration? I don't seem to find related logics.
> 
> Regarding this issue, I think it would be better to submit another PR to handle this after the current PR is merged. There are a couple of reasons for this:
> 
> 1. The use case of directly using podgroup for scheduling is relatively rare, which is why there was no related validate webhook before. For example, when creating or updating the queue field of a podgroup, there is no validation to check if the queue exists. The hdrf plugin also does not perform any validation related to hierarchical queue enablement.
> 2. Currently, adding this validation logic involves a significant amount of code changes, and the e2e tests cannot pass successfully. This is because the deployment yaml files need to be modified, and when running e2e tests on GitHub, the old yaml files are used. Therefore, if we directly include this in the current PR, the e2e tests will not pass.
> 
> Based on the above considerations, I believe this logic can be implemented in a separate PR after the current PR is merged. What do you think? Is this a particularly urgent matter?

If the implementation is quite complex, it would be acceptable to commit a separate PR for it.

pkg/scheduler/cache/cache.go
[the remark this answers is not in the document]
TaiPark: I think this might be better:
```
Disable hierarchy: default
Enable hierarchy: root + default
```
@hwdef Please check out this comment, what do you think?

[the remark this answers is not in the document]
hwdef: Because it involves the update process, after discussion, I think it is a balanced choice to create the root queue to occupy a position. Even if the hierarchical queue is not enabled,

TaiPark:
I don't have any more questions, looks good to me

pkg/scheduler/framework/session_plugins.go
Rui-Gan: I did modify the code as suggested earlier, but during subsequent iterations, I discovered that the modification would impact the functionality. For hierarchical queues, we need to first preempt queues that belong to the same parent queue as the preemptor. This means that when sorting the victims, it is necessary to have information about the preemptor. Therefore, it is necessary to add a victimQueueOrderFn here.

pkg/scheduler/framework/session_plugins.go
Rui-Gan: For the capacity plugin, the relevant `queueOrderFns` are registered both when the hierarchical queue feature is enabled and disabled. The `victimQueueOrderFn` is used here because when hierarchical queues are enabled, the sorting of victims requires information about the preemptor. Considering the need for compatibility with other plugins, in cases where no `victimQueueOrderFn` is registered, the sorting of victims ultimately uses `!QueueOrderFn`. This can be seen in the implementation of the `VictimQueueOrderFn` function.

I'm not sure if I have explained this clearly. If you have any further questions, please feel free to ask.

pkg/controllers/queue/queue_controller_action.go
```diff
@@ -82,24 +100,72 @@ func (c *queuecontroller) syncQueue(queue *schedulingv1beta1.Queue, updateStateF
...
+				klog.Errorf("Failed to patch annotation of Queue %s: %v.", newQueue.Name, err)
+				return err
+			}
+
+			req := &apis.Request{
+				QueueName: newQueue.Name,
+				Action:    busv1alpha1.CloseQueueAction,
+			}
+
+			c.enqueue(req)
```
Monokaix, line 139: Why re-enqueue here？

Rui-Gan: The main reason for re-enqueuing here is that the parent queue of the current queue is in a closed or closing state, while the current queue itself is not in a closed or closing state. Therefore, the purpose of re-enqueuing here is to close the queue.

Issue #3745, opened 2024-09-24
Podgroup "Unschedulable" warning event is reported repeatedly


liuyuanchun11:
### Description

If the controller is switched over or restarted, the podgroup "Unschedulable" alarm event is reported repeatedly. 

### Steps to reproduce the issue

1. Create a vcjob job-test and put it in the pending state
2. Execute “kubectl get events --sort-by='.metadata.creationTimestamp'”, will see warning event "job-test unscheduler, reasion:...."
3. Release some resources so that job-test is in the running state
4. Query the events again, will see the events of the job-test‘s pod running successfully
5. Delete the volcano-controller's pod and artificially create a volcano-controller restart or switchover
6. Execute “kubectl get events --sort-by='.metadata.creationTimestamp'”

### Describe the results you received and expected

Expected：
No new warning are seen

Actually：
Will see warning event "job-test unschedulable, reasion:....". In fact, the job has been running normally, and this alarm is recorded repeatedly

### What version of Volcano are you using?

v1.9.0

liuyuanchun11:
This issue has been fixed in PR #3746

PR #3746, opened 2024-09-24
fix job controller reports duplicate warnings

liuyuanchun11:
When the syncJob is processed in the job_controller, the condition timestamp of the podGroup is determined, and if the latest status is Unschedulable, a warning event is recorded. If the latest status is Scheduled, no warning event will be logged even if there is Unschedulable in the condtition.

fix issue #3745

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -286,13 +287,7 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 		if pg.Status.Phase != "" && pg.Status.Phase != scheduling.PodGroupPending {
 			syncTask = true
 		}
-
-		for _, condition := range pg.Status.Conditions {
-			if condition.Type == scheduling.PodGroupUnschedulableType {
-				cc.recorder.Eventf(job, v1.EventTypeWarning, string(batch.PodGroupPending),
-					fmt.Sprintf("PodGroup %s:%s unschedule,reason: %s", job.Namespace, job.Name, condition.Message))
-			}
-		}
+		cc.recordPodGroupEvent(job, pg)
 	}
 
 	var jobCondition batch.JobCondition
@@ -843,6 +838,27 @@ func (cc *jobcontroller) initJobStatus(job *batch.Job) (*batch.Job, error) {
 	}
 
 	return newJob, nil
+}
+
+func (cc *jobcontroller) recordPodGroupEvent(job *batch.Job, podGroup *scheduling.PodGroup) {
+	var latestCondition *scheduling.PodGroupCondition
+
+	// Get the latest condition by timestamp
+	for _, condition := range podGroup.Status.Conditions {
+		if condition.Status == v1.ConditionTrue {
+			if latestCondition == nil ||
+				condition.LastTransitionTime.Time.After(latestCondition.LastTransitionTime.Time) {
+				latestCondition = &condition
+			}
+		}
+	}
+
+	// If the latest condition is not scheduled, then a warning event is recorded
+	if latestCondition != nil && latestCondition.Type != scheduling.PodGroupScheduled {
+		cc.recorder.Eventf(job, v1.EventTypeWarning, string(batch.PodGroupPending),
+			fmt.Sprintf("PodGroup %s:%s %s, reason: %s", job.Namespace, job.Name,
+				strings.ToLower(string(latestCondition.Type)), latestCondition.Message))
+	}
 }
 
 func classifyAndAddUpPodBaseOnPhase(pod *v1.Pod, pending, running, succeeded, failed, unknown *int32) {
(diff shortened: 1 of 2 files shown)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -845,6 +839,29 @@ func (cc *jobcontroller) initJobStatus(job *batch.Job) (*batch.Job, error) {
 	return newJob, nil
 }
 
+func (cc *jobcontroller) recordPodGroupEvent(job *batch.Job, podGroup *scheduling.PodGroup) {
+	var latestCondition scheduling.PodGroupConditionType
+	var latestTransitionTime *metav1.Time
+	var latestConditionMsg string
+
+	// Get the latest condition by timestamp
+	for _, condition := range podGroup.Status.Conditions {
```
Monokaix, line 848: A little doubt here, the unschedulable condition will append to unschedulable condition？What if just override the unschedulable condition because pg is already schedulable and no need to concern the unschedulable condition.

Monokaix:
Seems that the main cause is that there are duplicated condition type `Unschedulable` and `Schedulable`, both of them is used to indicate the scheduling event，which I think is duplicated，just keep one of them is ok，and another problem is that every pg will have a condition contains `Unschedulable` like the following, even though pg can be scheduled first time, which is confused and not necessary.
```yaml
status:
  conditions:
  - lastTransitionTime: "2024-09-25T02:15:18Z"
    message: '1/0 tasks in gang unschedulable: pod group is not ready, 1 minAvailable'
    reason: NotEnoughResources
    status: "True"
    transitionID: b2adeea6-9812-4315-a3a4-6411c4b364c9
    type: Unschedulable
  - lastTransitionTime: "2024-09-25T02:15:20Z"
    reason: tasks in gang are ready to be scheduled
    status: "True"
    transitionID: f9b8c487-41ad-49a6-9474-3e732b5ae29e
    type: Scheduled
  phase: Running
  running: 1
```
So maybe we can add a new condition type like `Enqueueable`, and just update the `Enqueueable` condition and no need to update `Schedulable` contidtion, the pr https://github.com/volcano-sh/volcano/pull/3045 is doing the similar thing and we can consider them together.

Monokaix:
cc @lowang-bh

pkg/controllers/job/job_controller_actions.go
liuyuanchun11: This is just to solve the problem that the existing podgroup will repeatedly log warning events when the controller restarts. The condition update issue needs to be fixed separately.

liuyuanchun11:
> Seems that the main cause is that there are duplicated condition type `Unschedulable` and `Schedulable`, both of them is used to indicate the scheduling event，which I think is duplicated，just keep one of them is ok，and another problem is that every pg will have a condition contains `Unschedulable` like the following, even though pg can be scheduled first time, which is confused and not necessary.
> 
> ```yaml
> status:
>   conditions:
>   - lastTransitionTime: "2024-09-25T02:15:18Z"
>     message: '1/0 tasks in gang unschedulable: pod group is not ready, 1 minAvailable'
>     reason: NotEnoughResources
>     status: "True"
>     transitionID: b2adeea6-9812-4315-a3a4-6411c4b364c9
>     type: Unschedulable
>   - lastTransitionTime: "2024-09-25T02:15:20Z"
>     reason: tasks in gang are ready to be scheduled
>     status: "True"
>     transitionID: f9b8c487-41ad-49a6-9474-3e732b5ae29e
>     type: Scheduled
>   phase: Running
>   running: 1
> ```
> 
> So maybe we can add a new condition type like `Enqueueable`, and just update the `Enqueueable` condition and no need to update `Schedulable` contidtion, the pr #3045 is doing the similar thing and we can consider them together.

I've reworked the function logic a bit, and now it can support the subsequent extension of the new unscheduled condition

Monokaix:
We should record another issue to track the following problem:

- duplicated condition update with `Unschedulable` and `Scheduled`, just retain one `Scheduled` is enough.
- and add an extended condition to express the Unenqueueable event.
- every pg will record a `Unschedulable` type event fisrt no matter whether resources are enough and then schedule successfully.

JesseStutler:
> We should record another issue to track the following problem:
> 
> * duplicated condition update with `Unschedulable` and `Scheduled`, just retain one `Scheduled` is enough.
> * and add an extended condition to express the Unenqueueable event.
> * every pg will record a `Unschedulable` type event fisrt no matter whether resources are enough and then schedule successfully.

Yes I think we are ok with pr to not record warning events if the last condition is not scheduled, but we should keep resolving these problems.

liuyuanchun11:
> We should record another issue to track the following problem:
> 
> * duplicated condition update with `Unschedulable` and `Scheduled`, just retain one `Scheduled` is enough.
> * and add an extended condition to express the Unenqueueable event.
> * every pg will record a `Unschedulable` type event fisrt no matter whether resources are enough and then schedule successfully.

I submitted an issue for podgroup unschedulable condition.  #3749

pkg/controllers/job/job_controller_actions.go
Monokaix: Maybe we can add a ut here.

liuyuanchun11: UT testcase has been added

Issue #3734, opened 2024-09-18
When volcano-admission pod not running, create other pod can faild


lengrongfu:
### Description

When volcano-admission pod crash, It will affect me creating other pods.

### Steps to reproduce the issue

1. install volcano use helm install
2. scale volcano-admission  replicas to 0, simulation volcano-admission pod crash
```shell
$ kubectl -n volcano scale deployment volcano-admission --replicas 0
```
3. run a pod
```shell
$ kubectl run nginx --image=nginx
```

### Describe the results you received and expected

received results
```
root@ubuntu:~# kubectl run nginx --image=nginx
Error from server (InternalError): Internal error occurred: failed calling webhook "mutatepod.volcano.sh": failed to call webhook: Post "https://volcano-admission-service.volcano.svc:443/pods/mutate?timeout=10s": no endpoints available for service "volcano-admission-service"
```

expected results: can create pod success.

### What version of Volcano are you using?

1.9.0

lengrongfu:
I have two ideas:
1. Modify the failurePolicy field in the webhook
2. Add a unique label to the pod, and then the webhook selects by label

googs1025:
i think the problem of pod creation failing after a webhook crash is a common problem with webhooks. If you want other pods in the cluster not to be affected, you can modify the failurePolicy field of the webhook.
refer to:  https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

Monokaix:
The first solution change `failurePolicy` to ignore is ok.

lowang-bh:
You can also disable the webhook don't need.
Just modify it in `enabled_admissions: "/jobs/mutate,/jobs/validate,/podgroups/mutate,/pods/validate,/pods/mutate,/queues/mutate,/queues/validate"`

lengrongfu:
I don't think the solution of configuring failurePolicy=Ignore is very good. I suggest that we can configure matchConditions.
I have verified that it works well.

```
  matchConditions:
  - expression: object.spec.schedulerName == 'volcano'
    name: scheduler  
```

googs1025:
I think the key question is whether we need to set this feature as the default configuration for helm installation. Is this what you mean?

lengrongfu:
> I think the key question is whether we need to set this feature as the default configuration for helm installation. Is this what you mean?

yes.

Monokaix:
[...]

PR #3748, opened 2024-09-26
In webhook add match condition by scheduler name filter pod resources

lengrongfu:
Add `matchConditions` in webhook when rules is pod resources create operation, to validate `object.spec.schedulerName` whether is default volcano scheduler name `volcano`, current we don't consider scheduler name change.

/kind feature

fixes: https://github.com/volcano-sh/volcano/issues/3734

### DIFF
--- a/installer/helm/chart/volcano/templates/webhooks.yaml
+++ b/installer/helm/chart/volcano/templates/webhooks.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.custom.admission_enable }}
-
+{{ $scheduler_name := .Values.custom.scheduler_name | default "volcano" }}
 {{- if .Values.custom.enabled_admissions | regexMatch "/pods/mutate" }}
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
@@ -18,7 +18,7 @@ webhooks:
         namespace: {{ .Release.Namespace }}
         path: /pods/mutate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
     matchPolicy: Equivalent
     name: mutatepod.volcano.sh
     namespaceSelector:
@@ -33,6 +33,9 @@ webhooks:
 {{- end }}
     objectSelector: {}
     reinvocationPolicy: Never
+    matchConditions:
+    - expression: object.spec.schedulerName == '{{- $scheduler_name }}'
+      name: scheduler-name-match
     rules:
       - apiGroups:
           - ""
@@ -262,7 +265,7 @@ webhooks:
         namespace: {{ .Release.Namespace }}
         path: /pods/validate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
     matchPolicy: Equivalent
     name: validatepod.volcano.sh
     namespaceSelector:
@@ -276,6 +279,9 @@ webhooks:
         {{- toYaml .Values.custom.webhooks_namespace_selector_expressions | nindent 8 }}
 {{- end }}
     objectSelector: {}
+    matchConditions:
+    - expression: object.spec.schedulerName == '{{- $scheduler_name }}'
+      name: scheduler-name-match
     rules:
       - apiGroups:
           - ""
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -5032,7 +5032,7 @@ webhooks:
         namespace: volcano-system
         path: /pods/mutate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
     matchPolicy: Equivalent
     name: mutatepod.volcano.sh
     namespaceSelector:
@@ -5044,6 +5044,9 @@ webhooks:
             - kube-system
     objectSelector: {}
     reinvocationPolicy: Never
+    matchConditions:
+    - expression: object.spec.schedulerName == 'volcano'
+      name: scheduler-name-match
     rules:
       - apiGroups:
           - ""
@@ -5227,7 +5230,7 @@ webhooks:
         namespace: volcano-system
         path: /pods/validate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
     matchPolicy: Equivalent
     name: validatepod.volcano.sh
     namespaceSelector:
@@ -5238,6 +5241,9 @@ webhooks:
             - volcano-system
             - kube-system
     objectSelector: {}
+    matchConditions:
+    - expression: object.spec.schedulerName == 'volcano'
+      name: scheduler-name-match
     rules:
       - apiGroups:
           - ""


lengrongfu:
cc @Monokaix

Monokaix:
Are there any webhoon also need this?

Monokaix:
Why add a new file installer/volcano-latest.yaml, you want to override the vocano-deployment.yaml?

lengrongfu:
> Are there other webhook also need this?

I don't think other webhook are needed because they are all volcano resources.

lengrongfu:
> Why add a new file installer/volcano-latest.yaml, you want to override the vocano-deployment.yaml?

No, because i merge pr fail for ci jobs in lint check step. so i followed the prompts generate `installer/volcano-latest.yaml` this file.

image


https://github.com/volcano-sh/volcano/actions/runs/11046591401/job/30686212838

Monokaix:
> > Why add a new file installer/volcano-latest.yaml, you want to override the vocano-deployment.yaml?
> 
> No, because i merge pr fail for ci jobs in lint check step. so i followed the prompts generate `installer/volcano-latest.yaml` this file.
> 
> image
> 
> https://github.com/volcano-sh/volcano/actions/runs/11046591401/job/30686212838

This file should `mv` to vocano-deployment.yaml and override it.

lengrongfu:
> > > Why add a new file installer/volcano-latest.yaml, you want to override the vocano-deployment.yaml?
> > 
> > 
> > No, because i merge pr fail for ci jobs in lint check step. so i followed the prompts generate `installer/volcano-latest.yaml` this file.
> > image
> > https://github.com/volcano-sh/volcano/actions/runs/11046591401/job/30686212838
> 
> This file should `mv` to vocano-deployment.yaml and override it.

got it, we can change this tips? make it executable locally.

```
make generate-yaml TAG=latest RELEASE_DIR=installer && mv ./installer/volcano-latest.yaml  ./installer/volcano-development.yaml
```

Monokaix:
> RELEASE_DIR

You can do that but notice that the generated file location is related with the env `RELEASE_DIR`.

lengrongfu:
> > RELEASE_DIR
> 
> 
> 
> You can do that but notice that the generated file location is related with the env `RELEASE_DIR`.

Ok，thanks， having remove release file

hwdef:
I'm not sure if this change is necessary, or I prefer not to accept it.
volcano allows you to modify the scheduler name to implement multiple schedulers in one cluster.
Therefore it is inappropriate to fix the name of the scheduler in the webhook

lengrongfu:
> I'm not sure if this change is necessary, or I prefer not to accept it.
> 
> volcano allows you to modify the scheduler name to implement multiple schedulers in one cluster.
> 
> Therefore it is inappropriate to fix the name of the scheduler in the webhook

The volcano scheduling name can be modified, but I don't see this parameter in the helm chart

hwdef:
please check this code.

https://github.com/volcano-sh/volcano/blob/0c9b846f2e871e6658122fcf2efdf653a1d99946/cmd/scheduler/app/options/options.go#L118C66-L118C86

lengrongfu:
> please check this code.
> 
> https://github.com/volcano-sh/volcano/blob/0c9b846f2e871e6658122fcf2efdf653a1d99946/cmd/scheduler/app/options/options.go#L118C66-L118C86

I mean there is no place to modify in the helm chart.

hwdef:
We can add this field instead of fixing its name

Monokaix:
> We can add this field instead of fixing its name

set a separate custom flag for helm?

hwdef:
> > We can add this field instead of fixing its name
> 
> set a separate custom flag for helm?

Add a field in helm to modify `scheduler-name`

lengrongfu:
> > > We can add this field instead of fixing its name
> 
> > 
> 
> > set a separate custom flag for helm?
> 
> 
> 
> Add a field in helm to modify `scheduler-name`

I can commit a pr to add this field.

Monokaix:
> > > > We can add this field instead of fixing its name
> > 
> > 
> > > 
> > 
> > 
> > > set a separate custom flag for helm?
> > 
> > 
> > Add a field in helm to modify `scheduler-name`
> 
> I can commit a pr to add this field.

You can do it in this pr.

installer/helm/chart/volcano/templates/webhooks.yaml
```diff
@@ -18,7 +18,7 @@ webhooks:
         namespace: {{ .Release.Namespace }}
         path: /pods/mutate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
```
hwdef, line 21: why change to Ignore?

installer/helm/chart/volcano/templates/webhooks.yaml
```diff
@@ -262,7 +265,7 @@ webhooks:
         namespace: {{ .Release.Namespace }}
         path: /pods/validate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
```
hwdef, line 268: same as above

installer/volcano-development.yaml
```diff
@@ -5032,7 +5032,7 @@ webhooks:
         namespace: volcano-system
         path: /pods/mutate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
```
hwdef, line 5035: same as above

installer/volcano-development.yaml
```diff
@@ -5227,7 +5230,7 @@ webhooks:
         namespace: volcano-system
         path: /pods/validate
         port: 443
-    failurePolicy: Fail
+    failurePolicy: Ignore
```
hwdef, line 5233: same as above

lengrongfu: Let me explain my point of view, why failurePolicy should is Fail in this place. 

When a pod scheduler name is `volcano`, but volcano webhook server not running, this time kube-apiserver call volcano webhook will fail, the pod will be paused here.

If we change failurePolicy from Fail to Ignore, then kube-apiserver can skip volcano webhook to volcano scheduler step, without MutatingWebhookConfiguration, can the scheduling achieve the expected results?

lengrongfu: Yes, So should we keep  failurePolicy is Fail?

I add `expression: object.spec.schedulerName == '{{- $scheduler_name }}'` this condition is because, i hope when volcano wenhook server not running whatevery the reason, don't influence not use volcano scheduler pod.

hwdef: volcano webhook is a important inspection component, It is not an insignificant auxiliary component.
So if volcano webhook is down. All workloads that need to be scheduled using volcano should be blocked.
We should keep failurePolicy is Fail.

I agree with your other idea, if the workload is not scheduled using volcano, it should not be affected when the volcano webhook fails.

lengrongfu: thanks ~ 🤝

Monokaix: When we didn't add match condition of scheduler name, if the pod is not scheduled by volcano, the  policy can be `Ignore` to not block pods scheduled by default scheduler, if we have exactly added the match condition and be sured that pod is scheduled by volcano, then it's reasonable to keep policy `Fail`.

Monokaix:
Hi, any update?

lengrongfu:
> Hi, any update?

no update.

installer/volcano-development.yaml
hwdef: This has not been resolved

MondayCha:
Is there any update for this PR? The recent server room accident led to the restart of the Volcano scheduler, and other dependencies were blocked due to the Webhook. I think it's a good idea to provide the default configuration.

hwdef:
> Is there any update for this PR? The recent server room accident led to the restart of the Volcano scheduler, and other dependencies were blocked due to the Webhook. I think it's a good idea to provide the default configuration.

This PR will not be merged until it be modified.

/hold

https://github.com/volcano-sh/volcano/pull/3748/files#r1824205984

hwdef:
> Is there any update for this PR? The recent server room accident led to the restart of the Volcano scheduler, and other dependencies were blocked due to the Webhook. I think it's a good idea to provide the default configuration.

You should fix the webhook, instead of ignoring it

PR #3750, opened 2024-09-26
[Proposal] Add podgroup statistics doc

### LINKED ISSUE #3597: The volcano controller may have memory leak issues in large-scale clusters
**What happened**:

The volcano controller has memory leak issues in large-scale clusters

企业微信截图_1b4b6a2f-dc27-4679-84b7-24bba33cf72d

image

The volcano controller keeps running at a stable memory level

Build a cluster with 100,000 CPU cores and create 20,000 pods. Each pod will automatically terminate after running for 1 minute after startup. Continue this operation for 1 day.

Both the scheduler and the controller will update the queue CRD. Each time the pg changes, the controller will update the statistics of each state in the queue. After each scheduling, the scheduler will update the allocated field of the queue to record the amount of resources allocated. Both components update the queue, resulting in frequent conflict errors. When the controller encounters such an error, it will trigger AddRateLimited, resulting in accumulation and memory leaks.

**Environment**: linux
- Volcano Version: v1.8.2
- Kubernetes version (use `kubectl version`): v1.20
- OS (e.g. from /etc/os-release): Debian GNU/Linux 9 (stretch)
- Kernel (e.g. `uname -a`):  Linux 5.10.0-103-bili-colo x86_64
- Install tools:
- Others:

JesseStutler:
fix #3597

## Backgrounds

Each time when podgroups states changed, the controller will update the statistics of podgroup of each state in the queue's status. And at the end of each scheduling session, the volcano scheduler will also update the allocated filed in queue's status to recored the amount of the amount of resources allocated. Both components use `UpdateStatus` api to update the queue status, which will cause conflict errors. When the controller encounter such an error, it will trigger `AddRateLimited` to push back the podgroup into work queue, resulting in accumulation of memory leak. See in issue #3597.

## Alternative
Currently the statistics of podgroup of eatch state are only used for display by vcctl, there is no need to be persisted in queue's status. So when users need to use `vcctl queue get -n [name]` or `vcctl list` to display queues and each state of podgroups in queue, vcctl should calculate podgroup statistics in client side and then display them. And we can export these statistics of podgroups in each state as metrics.  

## Implementation
- In `syncQueue` of queue controller, we should not stat counts of podgroups in each state here, instead these statistics should recorded as metrics and then be exported outside: https://github.com/volcano-sh/volcano/blob/release-1.10/pkg/controllers/queue/queue_controller_action.go#L41-L61. And `UpdateStatus` should not be used here: https://github.com/volcano-sh/volcano/blob/release-1.10/pkg/controllers/queue/queue_controller_action.go#L84-L87, the `UpdateStatus` interface will verify the resourceVersion in the apiserver, which may cause concurrent update conflicts. Instead, `ApplyStatus` should be used here to avoid this situation, because we only need to update the status of the queue. It should be noted that the controller currently does not have patch queue status permissions, so we should add a patch queue/status permission to the clusterrole.
- `vcctl get -n [name]` and `vcctl list` display the statistics of podgroups in each state from queue's status directly, instead we should do one more step, query the podgroups owend in the queue, stat the counts of podgroups in each state at the `vcctl` side, and then display them.

### DIFF
--- a/docs/design/podgroup-statistics.md
+++ b/docs/design/podgroup-statistics.md
@@ -1,0 +1,58 @@
+# PodGroup Statistics
+
+## Backgrounds
+
+Each time when podgroups states changed, the controller will update the statistics of podgroup of each state in the queue's status. 
+And at the end of each scheduling session, the volcano scheduler will also update the allocated filed in queue's status to recored 
+the amount of the amount of resources allocated. Both components use `UpdateStatus` api to update the queue status, which will cause
+conflict errors. When the controller encounter such an error, it will trigger `AddRateLimited` to push back the podgroup into work queue, 
+resulting in accumulation of memory leak. See in issue #3597: https://github.com/volcano-sh/volcano/issues/3597.
+
+## Alternative
+Currently the statistics of podgroups of each state are only used for display by vcctl, there is no need to be persisted in queue's status. 
+So when users need to use `vcctl queue get -n [name]` or `vcctl list` to display queues and each state of podgroups in queue, 
+vcctl should calculate podgroup statistics in client side and then display them. And we can export these statistics of podgroups in each state as metrics.  
+
+## Implementation
+- In `syncQueue` of queue controller, counts of podgroups in each state will not be persisted in queue status anymore, 
+instead these statistics will be recorded as metrics and then be exported directly: https://github.com/volcano-sh/volcano/blob/c9be5c4c934597d99a0a80c9b26a3e919bbf8877/pkg/controllers/queue/queue_controller_action.go#L41-L61. 
+And `UpdateStatus` should not be used here: https://github.com/volcano-sh/volcano/blob/c9be5c4c934597d99a0a80c9b26a3e919bbf8877/pkg/controllers/queue/queue_controller_action.go#L84-L87, 
+the `UpdateStatus` interface will verify the resourceVersion in the apiserver, which may cause concurrent update conflicts. 
+Instead, `ApplyStatus` should be used here to avoid this situation, because we only need to update the status of the queue. 
+It should be noted that the controller currently does not have patch queue status permissions, so we should add a patch queue/status permission to the clusterrole.
+- `vcctl get -n [name]` and `vcctl list` display the statistics of podgroups in each state from queue's status directly, 
+instead we should do one more step, query the podgroups owend in the queue, stat the counts of podgroups in each state at the `vcctl` side, and then display them. 
+- Those metrics called `queue_pod_group_[state]_count`, which are recorded in proportion/capacity plugins in scheduler, 
+now will be moved to controller to record. The metrics involved are as follows:
+     - `queue_pod_group_inqueue_count`: The number of Inqueue PodGroup in this queue
+     - `queue_pod_group_pending_count`: The number of Pending PodGroup in this queue
+     - `queue_pod_group_running_count`: The number of Running PodGroup in this queue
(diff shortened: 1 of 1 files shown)


Monokaix:
Should also add that queue's pg statistics are still maintained in queue cache because these data is needed when close a queue.

docs/design/podgroup-statistics.md
```diff
@@ -0,0 +1,12 @@
+# PodGroup Statistics
+
+## Backgrounds
+
+Each time when podgroups states changed, the controller will update the statistics of podgroup of each state in the queue's status. And at the end of each scheduling session, the volcano scheduler will also update the allocated filed in queue's status to recored the amount of the amount of resources allocated. Both components use `UpdateStatus` api to update the queue status, which will cause conflict errors. When the controller encounter such an error, it will trigger `AddRateLimited` to push back the podgroup into work queue, resulting in accumulation of memory leak. See in issue #3597.
```
hwdef, line 5: please add link for 3597

docs/design/podgroup-statistics.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Alternative
+Currently the statistics of podgroup of eatch state are only used for display by vcctl, there is no need to be persisted in queue's status. So when users need to use `vcctl queue get -n [name]` or `vcctl list` to display queues and each state of podgroups in queue, vcctl should calculate podgroup statistics in client side and then display them. And we can export these statistics of podgroups in each state as metrics.
```
hwdef, line 8: what is `eatch` mean?

hwdef (commented):
I think there is nothing wrong with this implementation logic

docs/design/podgroup-statistics.md
JesseStutler: each state means podgroups in pending/running/unknown/inqueue/completed status

hwdef: image

I mean there is a typo

JesseStutler: Oh sorry, didn't notice that there was a typo, now I have fixed it.

docs/design/podgroup-statistics.md
JesseStutler: done

JesseStutler:
> Should also add that queue's pg statistics are still maintained in queue cache because these data is needed when close a queue.

I added a notice item to record this, please review it again.

hwdef (commented):
/ok-to-test
/lgtm

docs/design/podgroup-statistics.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Implementation
+- In `syncQueue` of queue controller, we should not stat counts of podgroups in each state here, instead these statistics should be recorded as metrics and then be exported outside: https://github.com/volcano-sh/volcano/blob/release-1.10/pkg/controllers/queue/queue_controller_action.go#L41-L61. And `UpdateStatus` should not be used here: https://github.com/volcano-sh/volcano/blob/release-1.10/pkg/controllers/queue/queue_controller_action.go#L84-L87, the `UpdateStatus` interface will verify the resourceVersion in the apiserver, which may cause concurrent update conflicts. Instead, `ApplyStatus` should be used here to avoid this situation, because we only need to update the status of the queue. It should be noted that the controller currently does not have patch queue status permissions, so we should add a patch queue/status permission to the clusterrole.
```
Monokaix, line 11: Please paste a permanent code link.

JesseStutler: done

PR #3751, opened 2024-09-26
feature: Add podgroups statistics

### LINKED ISSUE #3597: The volcano controller may have memory leak issues in large-scale clusters

JesseStutler:
fix #3597, related proposal pr: https://github.com/volcano-sh/volcano/pull/3750

# Implementation
- In `syncQueue` of queue controller, concurrently using `UpdateStatus` will cause conflict, use `ApplyStatus` instead, and need to add patch permission of vc-controller for `patch/status`. And we record the statistics of podgroups in each state in queue as metrics to be exported outside, we don't update the statistics of podgroups in queue's status now. 
- `vcctl get -n [name]` and `vcctl list` display the statistics of podgroups in each state from queue's status directly, instead we do one more step, query the podgroups owend in the queue, stat the counts of podgroups in each state at the vcctl side, and then display them.

# Verification
- vcctl list
  -  When there are no podgroups in queue, list is normal:
  image
  -  When there are podgroups in queue, list can also get the statistics of podgroups:
  image
- vcctl get
  -  get is normal:
    image
- metrics
image

### DIFF
--- a/pkg/controllers/queue/queue_controller_action.go
+++ b/pkg/controllers/queue/queue_controller_action.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -34,7 +33,9 @@ import (
 	"volcano.sh/apis/pkg/apis/bus/v1alpha1"
 	busv1alpha1 "volcano.sh/apis/pkg/apis/bus/v1alpha1"
 	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	v1beta1apply "volcano.sh/apis/pkg/client/applyconfiguration/scheduling/v1beta1"
 	"volcano.sh/volcano/pkg/controllers/apis"
+	"volcano.sh/volcano/pkg/controllers/metrics"
 	"volcano.sh/volcano/pkg/controllers/queue/state"
 )
 
@@ -83,8 +84,13 @@ func (c *queuecontroller) syncQueue(queue *schedulingv1beta1.Queue, updateStateF
 			queueStatus.Unknown++
 		case schedulingv1beta1.PodGroupInqueue:
 			queueStatus.Inqueue++
+		case schedulingv1beta1.PodGroupCompleted:
+			queueStatus.Completed++
 		}
 	}
+
+	// Update the metrics
+	metrics.UpdateQueueMetrics(queue.Name, &queueStatus)
 
 	if updateStateFn != nil {
 		updateStateFn(&queueStatus, podGroups)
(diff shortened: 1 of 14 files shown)


cmd/controller-manager/app/server.go
```diff
@@ -177,3 +188,10 @@ func isControllerEnabled(name string, controllers []string) bool {
 	// if we get here, there was no explicit inclusion or exclusion
 	return hasStar
 }
+
+func promHandler() http.Handler {
```
Monokaix, line 192: This is duplicated with the func in scheduler.

installer/helm/chart/volcano/templates/controllers.yaml
```diff
@@ -58,7 +58,7 @@ rules:
     verbs: ["get", "create", "delete", "update"]
   - apiGroups: ["scheduling.incubator.k8s.io", "scheduling.volcano.sh"]
     resources: ["podgroups", "queues", "queues/status"]
-    verbs: ["get", "list", "watch", "create", "delete", "update"]
+    verbs: ["get", "list", "watch", "create", "delete", "update", "patch"]
```
Monokaix, line 61: Is update still need?

cmd/controller-manager/app/options/options.go
```diff
@@ -37,6 +37,7 @@ const (
 	defaultMaxRequeueNum       = 15
 	defaultSchedulerName       = "volcano"
 	defaultHealthzAddress      = ":11251"
+	defaultListenAddress       = ":8080"
```
hwdef, line 40: Would a different port be better?
1. Make a distinction with the scheduler port.
2. Port 8080 is too common and prone to conflicts.

hwdef (commented):
Is there something I'm missing? Where is the logic for deleting the update queue status

JesseStutler:
Sorry I don't get the point, you mean that when the queue is deleted, the metrics need also be deleted, right?

hwdef:
Sorry I didn't express myself clearly.
Our previous discussion was not to update some states of the queue, such as the number of pending and running in the queue.
So should we delete this part of the code?

JesseStutler:
Oh yes now the pending/running/unknown/inqueue/completed pg statistics won't be persisted into etcd, see here: 
```
queueStatusApply := v1beta1apply.QueueStatus().WithState(queueStatus.State).WithAllocated(queueStatus.Allocated)
	queueApply := v1beta1apply.Queue(queue.Name).WithStatus(queueStatusApply)
	if _, err := c.vcClient.SchedulingV1beta1().Queues().ApplyStatus(context.TODO(), queueApply, metav1.ApplyOptions{FieldManager: controllerName}); err != nil {
		klog.Errorf("Failed to apply status of Queue %s: %v.", queue.Name, err)
		return err
	}
```
You can see above that the data called `queueStatusApply` which need to update is only state and allocated resource, won't update those statistics now.

pkg/cli/queue/util.go
```diff
@@ -62,3 +63,26 @@ func createQueueCommand(ctx context.Context, config *rest.Config, action busv1al
...
+
+type PodGroupStatistics struct {
+	Inqueue   int
+	Pending   int
+	Running   int
+	Unknown   int
+	Completed int
+}
+
+func (pgStats *PodGroupStatistics) StatPodGroupCountsForQueue(pg *v1beta1.PodGroup) {
```
Monokaix, line 75: It's a pg resource but in cli/queue pkg.

pkg/controllers/queue/queue_controller_handler.go
```diff
@@ -19,6 +19,7 @@ package queue
 import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
+	"volcano.sh/volcano/pkg/controllers/metrics"
```
Monokaix, line 22: Please sort import.

pkg/controllers/queue/queue_controller_action.go
```diff
@@ -58,126 +60,52 @@ func (c *queuecontroller) syncQueue(queue *schedulingv1beta1.Queue, updateStateF
...
+			newQueueStatus.Unknown++
 		case schedulingv1beta1.PodGroupInqueue:
-			queueStatus.Inqueue++
+			newQueueStatus.Inqueue++
+		case schedulingv1beta1.PodGroupCompleted:
+			newQueueStatus.Completed++
 		}
 	}
 
+	// Update the metrics
```
Monokaix, line 75: Can wrap a func.

pkg/controllers/queue/queue_controller_util.go
```diff
@@ -38,3 +43,11 @@ func IsQueueReference(ref *metav1.OwnerReference) bool {
 
 	return true
 }
+
+func UpdateQueueMetrics(queueName string, queueStatus *schedulingv1beta1.QueueStatus) {
```
Monokaix, line 47: Put these to metrics pkg is better.

hwdef (commented):
/lgtm

PR #3755, opened 2024-09-29
[cherry-pick for release-1.10]fix job controller reports duplicate warnings

liuyuanchun11:
When the syncJob is processed in the job_controller, the condition timestamp of the podGroup is determined, and if the latest status is Unschedulable, a warning event is recorded. If the latest status is Scheduled, no warning event will be logged even if there is Unschedulable in the condtition.

fix issue https://github.com/volcano-sh/volcano/issues/3745

cherry-pick from #3746


william-wang (approved):
/approve

PR #3756, opened 2024-09-30
Don't depend on ready condition of nodes instead of taints in session

Monokaix:
Some cni plugin need deployed first then the node can get ready，so we should not filter not ready node in scheduler，instead the node will be tainted by kube controller when node not ready，and pod will not be scheduled by default，also pod can set tolerations to get scheduled.
xref: https://github.com/kubernetes/design-proposals-archive/blob/main/scheduling/taint-node-by-condition.md
 previous pr：https://github.com/volcano-sh/volcano/pull/507


hwdef (commented):
/lgtm

hwdef (approved):

PR #3763, opened 2024-10-08
Use watch to listen expected event instead of wait.poll

Monokaix:
Fix unstable e2e case, https://github.com/volcano-sh/volcano/actions/runs/11228371786/job/31212142643#step:7:74563
use `watch` to listen event, wait.poll is not accurate becasue the status of pod/queue will change and use `watch` can catch all events and status.


Monokaix:
cc @lowang-bh @hwdef

hwdef (commented):
/lgtm

PR #3766, opened 2024-10-11
add scheduler-name field in charts

lengrongfu:
add scheduler-name field in chart file to configuration custome scheduler name

From: https://github.com/volcano-sh/volcano/pull/3748#issuecomment-2401812807


installer/helm/chart/volcano/templates/admission.yaml
```diff
@@ -133,6 +133,7 @@ spec:
             - --admission-conf=/admission.local.config/configmap/{{base .Values.basic.admission_config_file}}
             - --webhook-namespace={{ .Release.Namespace }}
             - --webhook-service-name={{ .Release.Name }}-admission-service
+            - --scheduler-name={{ .Values.basic.scheduler_name }}
```
hwdef, line 136: There is no need to add this parameter by default

Monokaix: Yeah we just need to add it to `webhook.yaml`.

lengrongfu: > There is no need to add this parameter by default

I am confused, why in the `admission.yaml` file this parameter by default, if we change `scheduler.yaml` and `controllers.yaml` two deployment for scheduler-name , than use create pod can skip webhook step? 

Shouldn't all services use the same scheduler-name?

hwdef:
Sorry I didn't express myself clearly, 
What I mean is that parameters are added only when schedulerName is set by helm.
So, you code should like this

https://github.com/volcano-sh/volcano/blob/7dfa8c64f8147a500897f69b51988238965f6545/installer/helm/chart/volcano/templates/controllers.yaml#L11

lengrongfu:
> Sorry I didn't express myself clearly, What I mean is that parameters are added only when schedulerName is set by helm. So, you code should like this
> 
> https://github.com/volcano-sh/volcano/blob/7dfa8c64f8147a500897f69b51988238965f6545/installer/helm/chart/volcano/templates/controllers.yaml#L11

Oh, then i know, thanks~

hwdef:
The code can be optimized again.
You can add code like this, 
In this way, there will be no changes in `volcano-development.yaml` when schedulerName is not set.


https://github.com/volcano-sh/volcano/blob/7dfa8c64f8147a500897f69b51988238965f6545/installer/helm/chart/volcano/templates/admission.yaml#L5

Monokaix:
The webhook modification will be done at your another pr?

lengrongfu:
> The webhook modification will be done at your another pr?

I think so, try to keep the functional independence of each PR.

Issue #3770, opened 2024-10-14
Frequent creation and deletion of vcjobs with the same name will cause newly submitted vcjobs to fail directly


wangyang0616:
### Description

vcjob sets the vcjob policy:
```
policies：
  - event: PodEvicted
     action: TerminateJob
```

Delete the running vcjob, and recreate the vcjob with the same name during the terminating process of its corresponding pod. The newly created vcjob will be affected by the deletion of the old pod and will fail directly.

### Steps to reproduce the issue

1. Create a vcjob, set the corresponding policy, and wait for the vcjob to run. The vcjob example is as follows:
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: test
spec:
  schedulerName: volcano
  minAvailable: 3
  policies:
    - event: PodEvicted
      action: TerminateJob
  tasks:
    - replicas: 3
      name: "worker"
      template:
        spec:
          containers:
            - image: alpine
              command: ["/bin/sh", "-c", "sleep 1000"]
              imagePullPolicy: IfNotPresent
              name: worker
              resources:
                requests:
                  cpu: "1"
          restartPolicy: OnFailure
```
2. Delete the vcjob, and recreate a vcjob with the same name during the pod cleanup process
3. Wait for about a minute, and the status of the newly submitted vcjob will be terminated

### Describe the results you received and expected

The newly submitted vcjob with the same name can run successfully

### What version of Volcano are you using?

master

PR #3771, opened 2024-10-14
fix: frequently create and delete vcjobs with the same name, and the …

wangyang0616:
fix: [#3770 ](https://github.com/volcano-sh/volcano/issues/3770)

Repair solution:

Add the uid information of the vcjob to which the pod belongs to in the pod-related events. When processing the pod event, determine whether the uid of the vcjob with the same name in the current cache is consistent with the uid of the vcjob to which the pod belongs. If they are inconsistent, discard the original policy information and execute the syncjob action.

### DIFF
--- a/pkg/controllers/apis/request.go
+++ b/pkg/controllers/apis/request.go
@@ -19,6 +19,8 @@ package apis
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/types"
+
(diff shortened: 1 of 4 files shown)


william-wang:
@wangyang0616 please make the ci happy.

wangyang0616:
> @wangyang0616 please make the ci happy.

done

william-wang (approved):
/approve

PR #3772, opened 2024-10-15
perf: imporve end to end queue processing latency

fengruotj:
The current implementation of queue controller handles queue state operations and queuecommand operations for sequentially which may come into a performance bottleneck in performance-critical clusters. For example the end user may increase queuecommand & queue creation speed, so this pr add an option to make queue & queuecommand creation parallelly to handle such case.

Here, queue-controller workqueue pending process object metrics. 
image

After perf: imporve end to end queue processing latency, the queue-controller workqueue pending process object finally reduced.

image

### DIFF
--- a/pkg/controllers/queue/queue_controller.go
+++ b/pkg/controllers/queue/queue_controller.go
@@ -90,6 +90,7 @@ type queuecontroller struct {
 	enqueueQueue func(req *apis.Request)
 
 	recorder      record.EventRecorder
+	workers       uint32
 	maxRequeueNum int
 }
 
@@ -125,6 +126,7 @@ func (c *queuecontroller) Initialize(opt *framework.ControllerOption) error {
 	if c.maxRequeueNum < 0 {
 		c.maxRequeueNum = -1
 	}
+	c.workers = opt.WorkerThreadsForQueue
 
 	queueInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    c.addQueue,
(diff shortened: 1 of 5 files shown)


fengruotj:
hi @kevin-wangzefeng, this is the perf: imporve end to end queue processing latency
/assign @kevin-wangzefeng

Monokaix:
Hi, please sign off your commit with `git commit -s`.

fengruotj:
> Hi, please sign off your commit with `git commit -s`.

Hi, I've already signed off my commit wiht `git commit -s`. Thanks you very much~

Monokaix:
/lgtm
cc @hwdef @lowang-bh

pkg/controllers/queue/queue_controller.go
```diff
@@ -187,8 +189,10 @@ func (c *queuecontroller) Run(stopCh <-chan struct{}) {
 		}
 	}
 
-	go wait.Until(c.worker, 0, stopCh)
-	go wait.Until(c.commandWorker, 0, stopCh)
+	for i := 0; i < int(c.workers); i++ {
+		go wait.Until(c.worker, 0, stopCh)
+		go wait.Until(c.commandWorker, 0, stopCh)
```
hwdef, line 194: Should commandWorker also be placed here?

hwdef (commented):
/lgtm

fengruotj:
@hwdef  I've considered this as well. The main issue is that the state transition of the queue involves queuecommand, and commandWorker is primarily used to handle the queuecommand CRD, which is then passed to the worker for processing by workqueue. The main performance bottleneck lies in both the worker and the commandWorker. I believe concurrency should be applied to both.

hwdef:
@fengruotj 
Yes, you are right.

PR #3774, opened 2024-10-16
fix allocating more pods to a GPU when using volcano-vgpu feature

archlitchi:
fix an issue: allocate more pods to a GPU resulting a GPU exceeds its core limit.

image

related issue: 
https://github.com/Project-HAMi/volcano-vgpu-device-plugin/issues/30


pkg/scheduler/api/devices/nvidia/vgpu/utils.go
```diff
@@ -369,7 +369,7 @@ func checkNodeGPUSharingPredicateAndScore(pod *v1.Pod, gssnap *GPUDevices, repli
 			if int(gs.Device[i].Memory)-int(gs.Device[i].UsedMem) < int(val.Memreq) {
 				continue
 			}
-			if 100-gs.Device[i].UsedCore < uint(val.Coresreq) {
+			if 100-int32(gs.Device[i].UsedCore) < val.Coresreq {
```
Monokaix, line 372: Any difference with previous code？

archlitchi: 100-gs.Device[i].UsedCore can be a negative number, you can't assign it to uint type

hwdef (commented):
/approve

PR #3775, opened 2024-10-16
node-resource-reservation-proposal

molei20021:
Volcano node resource reservation proposal


JesseStutler:
Hi @molei20021, thanks for your contribution! /cc @Monokaix @hwdef @lowang-bh

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+- plugins:
+   - name: reserve
+     arguments:
+       reserve.nodeLabel: label1,label2
+       reserve.define.label1: {"business_type": "ebook"}
+       reserve.resources.label1: [{"start_hour": 3, "end_hour": 4, "cpu": 32, "memory": 64, "start_reserve_ago": "2h", "pod_num": 10, "cron": "daily"}]
+       reserve.define.label2: {"business_type": "computer"}
+       reserve.resources.label2: [{"start_hour": 7, "end_hour": 9, "cpu": 24, "memory": 96, "start_reserve_ago": "30m", "pod_num": 15, "cron": "weekly 1,2,5"}]
+```
+In the configuration, nodeLabel represent a node list which nodeselector satisfy the nodeLabel, resources represent a list of resource reservation configuration, for example, reserve.resources.label1 means in hour 3 to 4 everyday, 32 cpu, 64 memory need to be reserved for label1, and start to reserve 2h ago, if 10 reserve pods are scheduled or after hour 4, stop reserve.
```
Monokaix, line 30: reserve ， recover？

Monokaix:
Queue also has the ability to reserve resources, how about use different queues to associate with different jobs, and adjust the queue's guarantee resources regularly？

molei20021:
> Queue also has the ability to reserve resources, how about use different queues to associate with different jobs, and adjust the queue's guarantee resources regularly？

queue can not reserve for specified time interval, after the specified time when no reserve tasks will be created, it will still limit the resource of not guaranteed queues.

docs/design/node-resource-reservation-design.md
molei20021: node reserve means during reserve time interval, if other task will make the reserved resources less than reserve config, deny other task to schedule temporary until reserve tasks is scheduled.

Monokaix:
> > Queue also has the ability to reserve resources, how about use different queues to associate with different jobs, and adjust the queue's guarantee resources regularly？
> 
> queue can not reserve for specified time interval, after the specified time when no reserve tasks will be created, it will still limit the resource of not guaranteed queues.

How about adjust the queue guaranteed resources dynamically?

Monokaix:
> > resources dynamically?
> 
> we need queue to limit the quota of different users, if I put big reserve task of different users to one queue, quota can not be limited and even of the quota of one queue is reserved, big reserve task may still fail to schedule immediately if there are many fragmented resources in the cluster. In the document, I use predicate plugin to order top idle node to forbid tasks which are not reserved to schedule and this is helpful to reduce fragmented resources.
> 
> > > > Queue also has the ability to reserve resources, how about use different queues to associate with different jobs, and adjust the queue's guarantee resources regularly？
> > > 
> > > 
> > > queue can not reserve for specified time interval, after the specified time when no reserve tasks will be created, it will still limit the resource of not guaranteed queues.
> > 
> > 
> > How about adjust the queue guaranteed resources dynamically?
> 
> I see guarantee of queue is used in proportion plugin and capacity plugin, in proportion plugin, it can influence deserved value of queue, but if I set guarantee in many queues at some time, the deserved value may be less than guarantee. In my situation one queue not only have small unimportant tasks but also have some big reserve tasks. Different department should have different queues to restrict the quota so we need to reserve import big tasks globally and the reserved resources should be unfragmented to be scheduled by big reserved pod.

So you want to avoid fragment by node level not queue？
If do not consider fragment，can guarantee meet your need?

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
+# Volcano node resource reservation
+## background
+* Consider such situation: there are thounsands of pods to scheduler a day, in the first hour 500 low priority pods are created and schedulered which used 99% of cluster resource, in the second hour 10 high priority pods are created, however, low priority pods are still running, high priority pods can not be scheduled due to lack of resource.
+* The user want the high priority task in the second hour to have resource to schedule immediately every day and high priority task better not preempt low priority pods because some pod may have already run many days.
+## design
+![annotation](images/node-resource-reservation-annotation.png) 
+### recognize high priority pods
+There are two ways to recognize high priority pods:
+* set annotation volcano.sh/is-reserve: 1 in podgroup
```
JesseStutler, line 9: `volcano.sh/is-reserve:1` is not such formal, maybe use `volcano.sh/reserveable: true` is better.

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+## design
+![annotation](images/node-resource-reservation-annotation.png) 
+### recognize high priority pods
+There are two ways to recognize high priority pods:
+* set annotation volcano.sh/is-reserve: 1 in podgroup 
+which means all pods under podgroup are reserve tasks
+* set annotation volcano.sh/is-reserve: 1 in pod
+which means this pod is reserve task
+### recognize pod max running time
+* set annotation volcano.sh/runsec-max: 500 in podgroup
```
JesseStutler, line 14: same as `volcano.sh/runsec-max: 500`, such name is not so formal, use `volcano.sh/maximum-runtime: 500s` is better

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+which means all pods under podgroup will run 500 seconds max
+* set annotation volcano.sh/runsec-max: 500 in pod
+which means this pod will run 500 seconds max
```
JesseStutler, line 17: `will run 500 seconds max` the grammar is not correct. There are many incorrect grammars in this article. Please check it.

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+### reserve plugin
+#### configuration
+```
+- plugins:
+   - name: reserve
+     arguments:
+       reserve.nodeLabel: label1,label2
+       reserve.define.label1: {"business_type": "ebook"}
+       reserve.resources.label1: [{"start_hour": 3, "end_hour": 4, "cpu": 32, "memory": 64, "start_reserve_ago": "2h", "pod_num": 10, "cron": "daily"}]
```
JesseStutler, line 26: The format of the arguments here doesn't seem very formal. Can they be defined as APIs, or the arguments can be defined more formally? Like:
```yaml
arguments:
  reserveLabel:
  - nodeSelector: label1
     startHour:
     endHour:
     resources:
        cpu:32
        memory: 64
        xxx
     ...
  - nodeSelector: label2
     ...
```
Just give an example, we can discuss more later.

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+       reserve.define.label2: {"business_type": "computer"}
+       reserve.resources.label2: [{"start_hour": 7, "end_hour": 9, "cpu": 24, "memory": 96, "start_reserve_ago": "30m", "pod_num": 15, "cron": "weekly 1,2,5"}]
+```
+In the configuration, nodeLabel represent a node list which nodeselector satisfy the nodeLabel, resources represent a list of resource reservation configuration, for example, reserve.resources.label1 means in hour 3 to 4 everyday, 32 cpu, 64 memory need to be reserved for label1, and start to reserve 2h ago, if 10 reserve pods are scheduled or after hour 4, stop reserve.
+#### PredicateFn
+Predicate is used to restrict other pods to be scheduled on reserved nodes. Reserved nodes are filtered out from the list of nodes and will change dynamically. 
+* check if the task is a reserve task, if yes, permit the task to be scheduled on this node.
```
JesseStutler, line 33: Can you rearrange the explanation here according to the order of the flow chart? If it is a reserve task, it can be scheduled to this node directly? Doesn't it check whether the resources are sufficient?

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+* check if the time is in the reserved time range, if no, permit the task to be scheduled on this node.
```
JesseStutler, line 34: The task is a normal task or is a `reserve` task?

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+* check if the number of reserve pods which is scheduled is larger than the max pod number configured, if yes, permit the task to be scheduled on this node.
```
JesseStutler, line 35: This is not quite right. If the number exceeds the maximum number of pods configured, scheduling should not be allowed, right?

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+* order the nodes desc by node idle. Node idle is consisted of node resource unused and the resource will be released in the future before reserve start time. The node resource to be released in the future is calculated by the annotation of pod max running time. 
+* traverse the ordered nodes, accumulate the node allocatable resource, if the accumulated resource is less than the resource to be reserved, add the node to reserve node list which means the system will have the trend to reserve big resource other than many small resource.
+* check if the node is in reserve node list, if yes, deny the task to be scheduled on this node.
```
JesseStutler, line 38: There is also a bit of confusion here, it should be that normal tasks(without reserve annotation) can not be scheduled on the reserve node, right?

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,45 @@
...
+which means all pods under podgroup are reserve tasks
+* set annotation volcano.sh/is-reserve: 1 in pod
+which means this pod is reserve task
+### recognize pod max running time
+* set annotation volcano.sh/runsec-max: 500 in podgroup
+which means all pods under podgroup will run 500 seconds max
+* set annotation volcano.sh/runsec-max: 500 in pod
+which means this pod will run 500 seconds max
+### reserve plugin
+#### configuration
```
JesseStutler, line 19: We can also add a `OpenSession` item here to explain how reserve plugin to gather reserve resources

JesseStutler (commented):
I have reviewed it, please check it, there may be some inaccuracies in the Chinese and English translations that cause us some ambiguities. You may need to translate your proposal more accurately, thanks. But I think the `reserve` plugin is a great idea.

docs/design/node-resource-reservation-design.md
molei20021: added

docs/design/node-resource-reservation-design.md
molei20021: yes

docs/design/node-resource-reservation-design.md
molei20021: max number of pod is used in situation that user know in 1 to 2 o'clock there'll be 10 reserve pod to be created, when in 1:30, 10 reserve pods are all scheduled, then in 1.30 to 2, we needn't reserve any resources, so more resoures are saved for non-reserve tasks.

docs/design/node-resource-reservation-design.md
molei20021: the task can be any pod, we should first judge if it's reserve task

docs/design/node-resource-reservation-design.md
molei20021: if it is reserve task, PredicateFn will not forbid the task to be scheduled to the node, if in allocate action, the node do not have enough resource, the task will not be scheduled to the node.

docs/design/node-resource-reservation-design.md
molei20021: i have updated the document

docs/design/node-resource-reservation-design.md
molei20021: thanks for your opinion, i have updated

docs/design/node-resource-reservation-design.md
molei20021: thanks for your opinion, i have updated

googs1025:
/cc
I will be interested in this feature.

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,56 @@
...
+* register plugin function PredicateFn
+* register event handler AllocateFunc and DeallocateFunc to dynamically update the cache of reservedTaskPendingResource, reservedTaskAllocatedResource and resourceIdle
+#### PredicateFn
+Predicate is used to restrict other pods to be scheduled on reserved nodes. Reserved nodes are filtered out from the list of nodes and will change dynamically. 
+* check if the task is a reserve task, if yes, permit the task to be scheduled on this node.
+* check if the current time is within the reserved time range, if no, permit the non-reserved task to be scheduled on this node.
+* check if the number of reserve pods which have been scheduled is larger than the max pod number configured, if yes, permit the non-reserved task to be scheduled on this node.
+* check if the node is in reserve node list(from nodeForbidMap cache), if yes, deny the non-reserved task to be scheduled on this node.
+* check if the node idle resource(from resourceIdle cache) is larger than the reserve requirements(max(reservedTaskAllocatedResource + reservedTaskPendingResource, reserveResourcesConfig)), if yes, permit the non-reserved task to be scheduled on this node.
```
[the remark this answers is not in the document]
molei20021: podNumToReserve is mainly used to check whether to stop reserve. If it is larger than the resources in config, we can adjust the resources to reservedTaskAllocatedResource + reservedTaskPendingResource

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,56 @@
...
+
+![predicate](images/node-resource-reservation-predicate.png) 
+#### JobStarvingFn
+JobStarving is used in preempt action which is an expand of reserve because sometimes reserve node resource may not be completely accurate. If podgroup or pod is set the annotation of reserve, the job is starving and can preempt other possible pods.
+#### PreemptableFn 
+PreemptableFn is used to cooperate JobStarvingFn to filter the victims to be preempted. In reserve situation, the preemptor can preempt the task which have the same node label and the create time of preemptee is later than the preemptor which means to preempt the task which should not be scheduled before and the occupancy rate of the cluster is not effected.
```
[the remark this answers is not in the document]
molei20021: preempt is not the key feature of node resource reservation, so I deleted it.

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,52 @@
+# Volcano node resource reservation
+## background
+* Consider such situation: there are thounsands of pods to be scheduled every day, in 1 to 2 o'clock 500 low priority pods are created and scheduled which used 99% of cluster resource, in 2 to 3 o'clock 10 high priority pods are created, however, low priority pods are still running, high priority pods can not be scheduled due to lack of resource.
+* Users want high priority tasks in 2 to 3 o'clock have resource to schedule immediately every day and high priority tasks not preempt low priority pods because some low priority pods have already run many days.
+## design
+![annotation](images/node-resource-reservation-annotation.png) 
+### recognize high priority pods
```
googs1025, line 7: It seems that it should not be called "high priority" but "need to be reserved" podgroup. High priority seems to have the priority field definition. 🤔

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,52 @@
...
+There are two ways to recognize high priority pods:
+* set annotation volcano.sh/reserveable: true in podgroup 
+which means all pods under podgroup are reserve tasks
+* set annotation volcano.sh/reserveable: true in pod
+which means this pod is reserve task
+### recognize pod max running time
```
googs1025, line 13: Is this a maximum execution time limit for a podgroup? If a task exceeds this time, will it be considered a failure?

molei20021: max running time can not only be set in pod but also in podgroup, if in podgroup, it means the max finish time of pods sub the min create time of pods will be less than the max running time. if exceed this time, it will not be considered a failure, the parameter is mainly used to predict whether pod will effect resource reservation later, for example, if the max running time is only 1 minture and we need to reserve 100c after 30 minute, it means it will not effect.

docs/design/node-resource-reservation-design.md
molei20021: updated

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,52 @@
...
+### reserve plugin
+#### configuration
+```
+- plugins:
+   - name: reserve
+     arguments:
+       reservelabels:
+       - nodeSelector:
+           business_type: ebook
+         startHour: 3
```
JesseStutler, line 27: Maybe we'd better use the Linux crontab format? Go has such packages to directly decode such format

docs/design/node-resource-reservation-design.md
```diff
@@ -0,0 +1,52 @@
...
+* make cache of reservedTaskPendingResource which is used to cache the accumulated resource of pending tasks
+* make cache of reservedTaskAllocatedResource which is used to cache the accumulated resource of allocated tasks
+* make cache of resourceIdle which is used to accumulate the node futureidle resource.
+* register plugin function PredicateFn
+* register event handler AllocateFunc and DeallocateFunc to dynamically update the cache of reservedTaskPendingResource, reservedTaskAllocatedResource and resourceIdle
+#### PredicateFn
+Predicate is used to restrict other pods to be scheduled on reserved nodes. Reserved nodes are filtered out from the list of nodes and will change dynamically. 
+* check if the task is a reserve task, if yes, permit the task to be scheduled on this node.
```
JesseStutler, line 46: Do you mean that if a node is selected by the label of the reserve plugin, then this node cannot schedule non-reserve tasks during the reserve time period? What if there is no reserve task in the reserve period, non-reserve tasks should wait until reserve time period ends?

JesseStutler:
Please rebase your codes and squash your commits

docs/design/node-resource-reservation-design.md
molei20021: Not exactly. If the node idle resource(from resourceIdle cache) is larger than the reserve requirements(max(reservedTaskAllocatedResource + reservedTaskPendingResource, reserveResourcesConfig)), if yes, permit the non-reserved task to be scheduled on this node.

docs/design/node-resource-reservation-design.md
molei20021: good idea, I have modified the doc

PR #3776, opened 2024-10-16
Supports rollback when allocate callback function fails

wangyang0616:
fix issue [#3777 ](https://github.com/volcano-sh/volcano/issues/3777) for task 1

### DIFF
--- a/pkg/scheduler/framework/statement.go
+++ b/pkg/scheduler/framework/statement.go
@@ -256,6 +256,15 @@ func (s *Statement) Allocate(task *api.TaskInfo, nodeInfo *api.NodeInfo) (err er
 		return fmt.Errorf("failed to find job %s", task.Job)
 	}
 
+	defer func() {
+		if err != nil {
+			if err := job.UpdateTaskStatus(task, api.Pending); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/scheduler/framework/statement.go
```diff
@@ -271,12 +280,38 @@ func (s *Statement) Allocate(task *api.TaskInfo, nodeInfo *api.NodeInfo) (err er
...
+				for _, reh := range s.ssn.eventHandlers {
+					if reh.DeallocateFunc != nil {
+						reh.DeallocateFunc(&Event{
+							Task: task,
+						})
+					}
+				}
+				return fmt.Errorf("Failed to exec allocate callback functions")
+			}
 		}
```
Monokaix, line 315: How about aggregate errs and call `unallocate` func when there is err?

wangyang0616: When each pod allocation is successful, but the overall vcjob does not meet expectations, unallocate is used to roll back the previous resource scheduling information.

For direct resource errors and exceptions during the scheduling process, I understand that it would be more appropriate to roll back directly based on the error.

william-wang (approved):
/lgtm

PR #3785, opened 2024-10-21
remove klog v1

hwdef:
klog v1 is no longer needed, we have klog v2


hwdef:
@hzxuzhonghu 
I tried some methods, but they all failed to import klog/v2. Do you have any suggestions?

hzxuzhonghu:
```
Error: pkg/controllers/framework/factory.go:22:2: import 'k8s.io/klog/v2' is not allowed from list 'main': klog is deprecated, use k8s.io/klog/v2 instead (depguard)
	"k8s.io/klog/v2"
	^
Error: pkg/scheduler/util/assert/assert.go:8:2: import 'k8s.io/klog/v2' is not allowed from list 'main': klog is deprecated, use k8s.io/klog/v2 instead (depguard)
	"k8s.io/klog/v2"
	^
Error: pkg/controllers/job/plugins/svc/svc.go:30:2: import 'k8s.io/klog/v2' is not allowed from list 'main': klog is deprecated, use k8s.io/klog/v2 instead (depguard)
	"k8s.io/klog/v2"
	^
```

.golangci.yml
```diff
@@ -66,5 +67,10 @@ linters-settings:
...
       - "all"
       - "-SA1019"   # TODO(fix) Using a deprecated function, variable, constant or field
-
-
+  depguard:
+    rules:
+      main:
+        list-mode: strict
+        deny:
+          - pkg: "k8s.io/klog"
```
hzxuzhonghu, line 75: looks like this is prefix match, is there a option

hzxuzhonghu: Allow is a prefix of packages to allow. A dollar sign $ can be used at the end of a package to specify it must be exact match only.

Deny is a map where the key is a prefix of the package to deny, and the value is a suggestion on what to use instead. A dollar sign $ can be used at the end of a package to specify it must be exact match only.

A Prefix List just means that a package will match a value, if the value is a prefix of the package. Example github.com/OpenPeeDeeP/depguard package will match a value of github.com/OpenPeeDeeP but won't match github.com/OpenPeeDeeP/depguard/v2.

https://github.com/OpenPeeDeeP/depguard

hwdef: Thanks for your guidance. It works now.

Issue #3640, opened 2024-07-30
Can not sync job status correctly when upgrading from v1.5


yuyue9284:
**What happened**:

When upgrading volcano from the v1.5 to later version, the already running job created by the v1.5 volcano cannot be correctly handled. 

**Anything else we need to know?**:

v1.5 changed the naming logics of pod group by adding UID into the name: #2140, and there is also another fix regarding handling the already created pod group without UID in create or update: #2400. But a similar fix does not exist in the `syncJob` function.

So, in the following part the `syncTask` won't be set to true.

https://github.com/volcano-sh/volcano/blob/1a696ee33c88a9b3f1bf3c43f272aacfc35121be/pkg/controllers/job/job_controller_actions.go#L280-L293

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Monokaix:
Yeah you're right, we haven't consider the old pg format when syncjob: )

Monokaix:
fixed by https://github.com/volcano-sh/volcano/pull/3786

PR #3652, opened 2024-08-05
fix: Can not sync job status correctly when upgrading from v1.5

QingyaFan:
v1.5 changed the naming logics of pod group by adding UID into the name: #2140, and there is also another fix regarding handling the already created pod group without UID in create or update: #2400. But a similar fix does not exist in the syncJob function.

Fixes #3640

PR #3786, opened 2024-10-21
Can not sync job status correctly when upgrading from v1.5 #3640

QingyaFan:
v1.5 changed the naming logics of pod group by adding UID into the name: #2140, syncJob function should change some logic.


QingyaFan:
@Monokaix  I create a new pull request for the same issue in here: https://github.com/volcano-sh/volcano/pull/3652 , please have a look.

Issue #3783, opened 2024-10-21
Support colocation for computing workload and mircro-service


william-wang:
### What is the problem you're trying to solve

More and more users are leveraging volcano to schedule the computing workload and microservices uniformly in one cluster to improve overall utilization. They expect better SLO management while resource competation happens and a fine-grained mechanism to analyze and oversell as much resource as possible.

### Describe the solution you'd like

We are going to offer the colocation, oversubscription and slo management in Volcano.

- [x] cpu and memory resource oversubscription danamically 
- [x] priority and SLO managment 
- [x] unified scheduling and preemption

william-wang:
Close the issue as the tasks have been completed.

PR #3789, opened 2024-10-23
Support colocation for computing workload and mircro-service

william-wang:
Implemention for https://github.com/volcano-sh/volcano/issues/3783


pkg/agent/features/feature.go
```diff
@@ -0,0 +1,29 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package features
+
+type Feature string
+
+const (
```
wangyang0616, line 21: Is the feature switch enabled or disabled by default? How to set the feature switch?

docs/design/colocation/Overview.md
```diff
@@ -0,0 +1,456 @@
...
+### Qos Model
+
+|              Qos              |                 Typical  Usage Scenario                  |
+| :---------------------------: | :------------------------------------------------------: |
+|     LC(Latency Critical)      | Exclusive CPU for latency critical core online workloads |
+| HLS(Highly Latency Sensitive) |       Extremely latency sensitive online workloads       |
+|     LS(Latency Sensitive)     |      Latency sensitive approximate online workloads      |
+|        BE(Best Effort)        |    Offline AI/Big Data workloads, tolerable evictions    |
+
+You can set different qos levels to identify workloads of different priorities, and there is no seperate `priorityClass` definition to be consistent with k8s.
```
wangyang0616, line 20: If QoS is set and priority is set for a workload, which of the following configurations prevails during scheduling?
For example, if workload-1 QoS is LS, priorityClass is 100, workload-2 QoS is LC, and priorityClass is 50, which workload is scheduled first when cluster resources are insufficient?

docs/design/colocation/Overview.md
```diff
@@ -0,0 +1,456 @@
...
+- Dynamic resource oversubscription: Dynamically calculate the resources that can be oversold based on the real-time CPU/Memory utilization of the node, and oversold resources can be used by offline workloads.
+- Network bandwidth isolation：Supports ingress network bandwidth limitation of the entire machine to ensure network usage for online workloads.
+
+## Quick start
+
+### Prepare
+
+Follow the [installation guide](https://github.com/volcano-sh/volcano?tab=readme-ov-file#quick-start-guide) to install volcano first, and then install volcano agent via kubectl, this way is both available for x86_64 and arm64 architecture. 
+
+```shell
```
wangyang0616, line 36: Volcano can be installed using the helm command. Does Volcano-agent support adaptation?

wangyang0616:
Does the online pod evict offline pod function depend on the preemption function of the Volcano? Does the preemption configuration need to be enabled?

docs/design/colocation/Overview.md
```diff
@@ -0,0 +1,456 @@
...
+nr_periods 1210
+nr_throttled 488
+throttled_time 10125826283
+nr_bursts 0
+burst_time 0
+```
+
+#### Limitation
+
+CPU burst relies on capabilities provided by the linux kernel, this feature only works with host upstream linux kernel >=5.14 and some Linux Distribution like OpenEuler 22.03 SP2 or higher version.
```
wangyang0616, line 149: This function has specific requirements on the underlying operating system. If the operating system does not meet the requirements, can I disable this function?

wangyang0616:
Does network bandwidth isolation have dependency requirements on the CNI plug-in, for example, the specified CNI component or version?

pkg/agent/features/feature.go
Monokaix: All features are enabled by default, and it can be disabled by set param flag `--supported-features`.

docs/design/colocation/Overview.md
Monokaix: This is compatible with k8s's priority and higher priority pods will be scheduled first and user should set proper priority for different Qos level of pod.

docs/design/colocation/Overview.md
Monokaix: Yeah we will support helm installation when the next version is officially released.

Monokaix:
> Does the online pod evict offline pod function depend on the preemption function of the Volcano? Does the preemption configuration need to be enabled?

It just depends on `eviction` interface and will take PDB into consideration.

docs/design/colocation/Overview.md
Monokaix: Yeah it can be disabled and you can find the guide in `Advanced settings` part.

Monokaix:
> Does network bandwidth isolation have dependency requirements on the CNI plug-in, for example, the specified CNI component or version?

Currently no limitation.

docs/design/colocation/Overview.md
william-wang: This is first pr for the feature. Sure, this new component will be supported by helm later and it can be enable or disable by user as well.

docs/design/colocation/Overview.md
william-wang: In you example, the LC and LS workloads use different kinds of resource, there is no resource competation between them. For the scheduling sequence, it is controlled by priority which is the same as  before.

Issue #3798, opened 2024-10-31
Add dynamic resources allocation(DRA) plugin


JesseStutler:
#### Motivation
Inherit from #3143, record DRA’s evolutionary tasks in volcano
#### TODO List
- volcano v1.11(or called v2.0.0) with k8s 1.31:
  - [x] Implement DRA plugin, introduce structured parameters DRA into volcano. ( Related API resources version is v1alpha3. Currently, DRA struct can't be accessed outside the k8s.io/kubernetes package, we have to cope whole codes of DRA plugin into volcano)
  - [x] Implement PreBindFns (Related issue: https://github.com/volcano-sh/volcano/issues/3618)
- volcano v1.12(or called v2.1.0) with k8s 1.32:
  - [x] Upgrade related API resources( `ResourceSlice`, `ResourceClaim`, `DeviceClass`, etc.) to version beta
  - [x] When DRA struct can be accessed outside the k8s.io/kubernetes, remove the codes of DRA plugin copied from kubernetes repo.

### Describe the solution you'd like

Two PRs #3487 and #3577 were not updated after LFX was finished, we have to finish introducing DRA into volcano. Currently, it's a good time to introduce DRA plugin into volcano, DRA reconstructed the architecture and all APIs in 1.30 and became more stable in 1.31. It will evolve to the beta version in 1.32. Since volcano is adapting to k8s v1.31, it is appropriate to introduce DRA at this time. After getting more feedback from users and doing some optimization, when volcano is upgraded to k8s 1.32, I believe volcano’s DRA plugin can also become more stable.

hwdef:
working in progress

Issue #3777, opened 2024-10-16
The pods that successfully apply for gpu resources lack gpu card allocation information


wangyang0616:
### Description

When Volcano schedules heterogeneous resources such as GPU, it first filters the node through the predicate, and then allocates GPU resources through the callback function AllocateFunc registered by the GPU and other plug-ins.

AllocateFunc has no return value or other structure to synchronize the execution result, that is, it does not judge and process whether the GPU allocation is successful. When the GPU allocation fails, the annotation information of the GPU card allocation will not be updated on the pod. The outer layer allocate does not currently perceive the exception. Continuing to execute subsequent bind actions will cause the pod to be scheduled to the node, but it will fail to run due to the lack of corresponding GPU resources.

### Describe the results you received and expected

When calling the callback function AllocateFunc to allocate heterogeneous resources, the caller can perceive its execution result and process it, which is mainly divided into the following steps:

- [x] Task1: AllocateFunc supports carrying the execution result of the callback function.
- [x] Task2: External actions such as allocate/preempt/backfill support the processing of AllocateFunc execution exceptions.
- [ ] Task3: The gpu plug-in supports synchronizing its internal exception processing information through allocateFunc.

### What version of Volcano are you using?

master

PR #3799, opened 2024-10-31
feature: Add dynamic resource allocation(DRA) plugin

JesseStutler:
Inspired by #3798 , introduce DRA into volcano. It solves the following things:
- Introduce the related codes of #3777, add rollback and errInfo into Pipeline and Allocate of statement.
- Refactor the backfill action to use Allocate and Unallocate in statement instead, stay consistent with other actions. 
- Add a new extension point called PreBindFn, volumebinding and DRA need to use such extension point to bind resources such as ResourceClaim, PVC before binding the pod onto the node. Related issue: #3618 
- Copy the whole related codes of structured parameters DRA into volcano, because currently `dynamicresources`  struct can't be accessed outside the k8s.io/kuberentes package, if we need to call k8s extension points such as `PreFilter`, `Filter`, etc., we have to access the struct. After this pr is merged and volcano upgrade the k8s version to 1.32: https://github.com/kubernetes/kubernetes/pull/128399, we may remove these codes. 
- Add a new plugin called dynamicresources, which adds some extension points such as:
  - PrePredicateFn
  - PredicateFn
  - EventHandler
  - PreBindFn

See docs/design/dynamicresources-plugin.md for more details.


installer/helm/chart/volcano/values.yaml
```diff
@@ -25,6 +25,7 @@ custom:
   leader_elect_enable: false
   enabled_admissions: "/jobs/mutate,/jobs/validate,/podgroups/mutate,/pods/validate,/pods/mutate,/queues/mutate,/queues/validate"
   colocation_enable: false
+  feature_gates: DynamicResourceAllocation=true
```
googs1025, line 28: Do we need to enable this feature-gate by default?

pkg/scheduler/capabilities/dynamicresources/dynamicresources.go
```diff
@@ -0,0 +1,808 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
googs1025, line 2: Is this part directly imported into the scheduler code? Or has it been modified and adapted (I don't know much about it)? If not, should it be marked in the code?

installer/helm/chart/volcano/values.yaml
JesseStutler: What do you think? Do we need to enable it by default? I think that if users need to enable dra plugin, they also need to reconfigure the scheduler plugins configmap to enable it, it also looks like featuregate. If we don't enable this feture-gate, users need to do reconfiguring the scheduler plugins configmap and enabling the feature-gate two steps.

pkg/scheduler/capabilities/dynamicresources/dynamicresources.go
JesseStutler: I just changed the `New` to `NewDRAPlugin` and deleted `EventsToRegister` and some related codes with it, I think volcano doesn't need them right now. These codes are copied from the kubernetes repo, so the license needs to keep.

googs1025: Understand, if so, if the upstream code changes, do I need to make the same changes?

JesseStutler: Yes, before the `dynamicResources` struct changed to `DynamicResources`, we can only copy the whole codes related into volcano, otherwise we can't directly call those extension points. It may cause us to miss tracking some repaired codes, but if we can access the `DynamicResources` struct in v1.32, the problem will be solved.

installer/helm/chart/volcano/values.yaml
JesseStutler: OK we need not to enable this feature-gate by default, I will add some informers in schedulerCache, some users may need to use k8s below v1.31, if there are no APIs like `DeviceClass`, `ResourceClaim`, etc registered, it will cause some errors.

hwdef:
still need

Monokaix:
Great job!

test/e2e/dra/main_test.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2021 The Volcano Authors.
```
googs1025, line 2: ditto

hack/e2e-kind-config.yaml
```diff
@@ -1,15 +1,29 @@
...
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+featureGates:
+  DynamicResourceAllocation: true
+  DRAResourceClaimDeviceStatus: true
+containerdConfigPatches:
+  # Enable CDI as described in
+  # https://tags.cncf.io/container-device-interface#containerd-configuration
+  - |-
+    [plugins."io.containerd.grpc.v1.cri"]
```
Monokaix, line 11: So we need also enable this when using DRA?

docs/user-guide/how_to_enable_dra.md
```diff
@@ -0,0 +1,76 @@
+# How to Enable Dynamic Resource Allocation (DRA) in Volcano Scheduler
```
Monokaix, line 1: This file should also be added to website after merged.

test/e2e/dra/builder.go
```diff
@@ -0,0 +1,394 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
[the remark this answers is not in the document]
JesseStutler: No, this file is directly copied from the dra e2e use case of k8s, so we need to keep the head of the kubernetes author

hack/e2e-kind-config.yaml
JesseStutler: Yes, mounting DRA device needs to use CDI

docs/user-guide/how_to_enable_dra.md
JesseStutler: OK

test/e2e/dra/main_test.go
JesseStutler: This file is written by ourselves so we can write Volcano authors, but I need to change it to 2025 I think

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -168,16 +175,17 @@ func enablePredicate(args framework.Arguments) predicateEnable {
...
+		nodePortEnable:                  true,
+		taintTolerationEnable:           true,
+		podAffinityEnable:               true,
+		nodeVolumeLimitsEnable:          true,
+		volumeZoneEnable:                true,
+		podTopologySpreadEnable:         true,
+		cacheEnable:                     false,
+		proportionalEnable:              false,
+		volumeBindingEnable:             true,
+		dynamicResourceAllocationEnable: false,
```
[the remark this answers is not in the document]
JesseStutler: The DRA feature gate is disabled by default in k8s v1.32: https://github.com/kubernetes/kubernetes/blob/9894294ef13a5b32803e3ca2c0d620a088cc84d1/pkg/features/versioned_kube_features.go#L183

However, as long as the user turns on feature gate, kube-scheduler will instantiate the DRA plugin. I have made the user take an extra step here. They need to configure whether to enable the DRA plugin in predicates and specify the feature gate `DynamicResourceAllocation`  as true. This avoids the situation when the user configures predicates but does not turn on featuregate, an empty DRA plugin won't be instantiated. What do you think?

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -377,6 +389,17 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		pp.volumeBindingPlugin = volumeBindingPluginInstance
 	}
+	// 10. DRA
+	var dynamicResourceAllocationPlugin *dynamicresources.DynamicResources
+	if predicate.dynamicResourceAllocationEnable {
```
wangyang0616, line 394: If the user turns on the DRA plug-in in the predicate, but does not turn on the DRA feature switch, what will happen? Is there an error in the predicate?

pkg/scheduler/plugins/predicates/predicates.go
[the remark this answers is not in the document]
googs1025: If I understand correctly, v1.33 has it enabled by default? 🤔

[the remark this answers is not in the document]
JesseStutler: Still disabled by default in v1.33: https://github.com/kubernetes/kubernetes/blob/v1.33.1/pkg/features/kube_features.go#L1167

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -99,6 +103,8 @@ type predicatesPlugin struct {
 	pluginArguments framework.Arguments
 	// The VolumeBindingPlugin needs to store to execute in PreBind and PreBindRollBack
 	volumeBindingPlugin *vbcap.VolumeBinding
+	// The DynamicResourceAllocationPlugin needs to store to execute in PreBind and PreBindRollBack
+	dynamicResourceAllocationPlugin *dynamicresources.DynamicResources
```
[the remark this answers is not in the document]
Monokaix: We can refactor it in the future, please add a TODO here.

pkg/scheduler/plugins/predicates/predicates.go
JesseStutler: No there is no error, an empty DRA plugin will be initialized: 
https://github.com/kubernetes/kubernetes/blob/67a30c0adcf52bd3f56ff0893ce19966be12991f/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go#L115-L119

https://github.com/kubernetes/kubernetes/blob/67a30c0adcf52bd3f56ff0893ce19966be12991f/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go#L377-L379
It's fine to initialize an empty DRA plugin, every extension point will judge whether the DRA plugin has been enabled or not first.

wangyang0616: ok

PR #3803, opened 2024-11-05
fix the type error in dashboard's name

zhifanggao:
change globcal to global


zhifanggao:
@Monokaix pls review it ,It does not affect the feature of volcano.

PR #3806, opened 2024-11-08
feat(env): Add automatic memory limit handling and move automaxprocs to pkg/util package

dongjiang1989:
### Is your feature request related to a problem?
It is good practice to set the Go runtime environment variables GOMAXPROCS and GOMEMLIMIT, which tell Go the max threads/memory it should use. We should set these in our containers that run Go (namely gloo and discovery) whenever resource limits are set. 
Automatically set GOMEMLIMIT to match Linux [cgroups(7)](https://man7.org/linux/man-pages/man7/cgroups.7.html) memory limit.

See:
https://pkg.go.dev/runtime
https://blog.howardjohn.info/posts/gomaxprocs/
https://github.com/KimMachineGun/automemlimit/

1. add automatic memory limit handling
2. move `automaxprocs` to `pkg/util` package

```bash
$ ./controller-manager --kubeconfig ~/.kube/config --auto-gomemlimit-ratio=0.8
I1108 11:42:04.897134   51417 cpu.go:28] Leaving GOMAXPROCS=8: CPU quota undefined
I1108 11:42:04.897584   51417 memory.go:52] GOMEMLIMIT set to 6871947673
I1108 11:42:04.901564   51417 leaderelection.go:254] attempting to acquire leader lease volcano-system/vc-controller-manager...
```

### Result

My conclusion from the kubernetes cluster, is it that setting the GOMEMLIMIT equal to the memory limit has the following advantages:

- Significantly lower memory usage (at the cost of slightly higher CPU usage)
- Fewer OOMKILL container failures
- Improved ability to recover from OOMKILL failures and continue reconciling operator, scheduler and controller

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -43,6 +43,7 @@ const (
 	defaultQueueWorkers        = 5
 	defaultGCWorkers           = 1
 	defaultControllers         = "*"
+	defaultMemlimitRatio       = 0.0
 )
 
 // ServerOption is the main context object for the controllers.
@@ -92,6 +93,9 @@ type ServerOption struct {
 	// Case3: "-gc-controller,-job-controller,-jobflow-controller,-jobtemplate-controller,-pg-controller,-queue-controller"
 	// to disable specific controllers,
 	Controllers []string
+
+	// The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory.
+	MemlimitRatio float64
 }
 
 type DecryptFunc func(c *ServerOption) error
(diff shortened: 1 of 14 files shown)


dongjiang1989:
Maybe java GC, Please re-run fail action, thanks

hwdef:
you can close and reopen this PR to re-run the CIs

Monokaix:
Can you support more information about why these are needed？

dongjiang1989:
> Can you support more information about why these are needed？

Thanks. Add information done.

Monokaix:
Welcome! just a little doubt here:
1. What's the behavior(memory can be used by container) when both set container memory limit and GOMEMLIMIT？Is there  any conflict?
2. Does GOMEMLIMIT depend on the cgroup version?  
3. Under what circumstance does user need auto gomemlimit？

internal/goruntime/memory.go
```diff
@@ -0,0 +1,53 @@
+/*
```
Monokaix, line 1: How about put them in util? seems the internal pkg is related to go, and it's not a internal pkg of volcao, so it's a little wired to put them in the root directory.

dongjiang1989: Thanks @Monokaix . Using Internal pkg ensure that certain parts of their codebase are "shielded" from external use.
But It is also possible to move it under pkg.

dongjiang1989:
> Welcome! just a little doubt here:
> 
> 1. What's the behavior(memory can be used by container) when both set container memory limit and GOMEMLIMIT？Is there  any conflict?
> 2. Does GOMEMLIMIT depend on the cgroup version?
> 3. Under what circumstance does user need auto gomemlimit？

1. **Can both set, no direct conflict.** 
The container memory limit and `GOMEMLIMIT` operate at different levels. The container memory limit is a more coarse - grained control imposed by the container runtime or the host operating system. It simply caps the total memory usage of the container's processes.
But `GOMEMLIMIT` is a finer - grained control that the Go runtime uses to manage its internal memory usage, specifically for the heap memory managed by the garbage collector.
However. Unreasonable settings of container memory limit and `GOMEMLIMIT` can also cause OOM. Therefore, auto setting gomemlimit is required.

2. **Support cgroup v1 and v2.** 
Ref:  https://github.com/KimMachineGun/automemlimit

3. **The final state does not require users to care.**
The final state does not require the user to care. In container deployment, auto setting gomemlimit.
In the current state, user-configurable feature flags are exposed.
Ref: https://github.com/prometheus/prometheus/releases/tag/v3.0.0
image

PR #3815, opened 2024-11-13
[WIP]fix ci

hwdef:


pkg/config/config_test.go
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 
 package config
 
+// aaa
```
dongjiang1989, line 19:
```suggestion
```

dongjiang1989: Invalid comment

hwdef: This is only temporary content. And this PR is WIP

dongjiang1989: > This is only temporary content. And this PR is WIP

I got it. thanks

PR #3822, opened 2024-11-15
scheduler-simulate-proposal

molei20021:


docs/design/scheduler-simulate-design.md
```diff
@@ -0,0 +1,32 @@
+# Volcano scheduler simulate
+## background
+* Consider such situation: users changed the parameter of nodeorder plugin and need to know the effect to the production enviroment. For example, after change the mostrequested.weight, if the average wait time of big task is shorter than before, etc.
```
JesseStutler, line 3: I think the simulation scheduling scenario is more suitable for node simulation. (e.g., we don’t have GPU and NPU nodes, if there is some bugs in GPU or NPU scheduling, we can use simulation scheduling to debug. We do meet this scenario in our production). And to test the performance of the scheduler in large-scale clusters. If it is just for feature validation, there is no much difference with starting a kind cluster.

docs/design/scheduler-simulate-design.md
```diff
@@ -0,0 +1,32 @@
...
+* output file includes nodes_detail.csv, queues_detail.csv, pods_detail.csv, pods_detail.csv. These files will be put in output_path directory.
+* nodes_detail.csv includes ts, nodeName, cpuRequest, memoryRequest
+* queues_detail.csv includes ts, queueName, cpuRequest, memoryRequest
+* pods_detail.csv includes podName, createTs, scheduleTs, finishTs
+#### metrics
+* output metrics includes volcano_simulator_scheduler_wait_time, volcano_simulator_pod_run_time, volcano_simulator_node_cpu_request, volcano_simulator_node_memory_request which will help user to monitor the schduler simulate by prometheus.
+### time simulator
+* Time simulator is helpful to shorten simulate time because it will not get the time of real world, it will always get next timestamp of the min value between the create time of next pod and the finish time of next pod.
+* The time related parameter should get from time simulator like pod create time, pod finish time, current time, etc...
+### kube-apiserver simulator
```
JesseStutler, line 29: What about using kwok? kwok can simulate thousands of nodes and doesn't consume many resources. I'm wondering that if we use kwok, then we don't need to simulate kube-apiserver and kube-controller-manager. We can keep using kube-apiserver's and kube-controller-manager's features and ensure their reliability (if they meet some errors and failed ). I worry about if you need to simulate kube-apiserver and kcm, there is lots of work you need to do, and you can't ensure the reliability of kube-apiserver and kcm.

JesseStutler: The project kube-scheduler-simulator also follow kwok to simulate scheduling, can we also do this way? https://github.com/kubernetes-sigs/kube-scheduler-simulator

molei20021: The problem is how to speed up the simulation time, for example, a pod may run 20 hours and we also let it run 20 hours in kwok?

docs/design/scheduler-simulate-design.md
molei20021: we can add more node features of gpu in nodes.csv like gpu_allocatable, etc

docs/design/scheduler-simulate-design.md
JesseStutler: I create a PR: #3830 to add two scripts: one is for installing kwok and the other is for creating fake nodes. Perhaps building upon kwok is a better way to do simulating, you can add your stage confugration to do time simulating or other useful simulating. I think it's better and help us to do less work, kwok has been adopted by many schedulers as a tool for simulating scheduling.

molei20021: I added kwok to the proposal to replace apiserver-simulator @JesseStutler

docs/design/scheduler-simulate-design.md
```diff
@@ -0,0 +1,32 @@
...
+* nodes_detail.csv includes ts, nodeName, cpuRequest, memoryRequest
+* queues_detail.csv includes ts, queueName, cpuRequest, memoryRequest
+* pods_detail.csv includes podName, createTs, scheduleTs, finishTs
+#### metrics
+* output metrics includes volcano_simulator_scheduler_wait_time, volcano_simulator_pod_run_time, volcano_simulator_node_cpu_request, volcano_simulator_node_memory_request which will help user to monitor the schduler simulate by prometheus.
+### time simulator
+* Time simulator is helpful to shorten simulate time because it will not get the time of real world, it will always get next timestamp of the min value between the create time of next pod and the finish time of next pod.
+* The time related parameter should get from time simulator like pod create time, pod finish time, current time, etc...
+### kwok
+* kwok is used to create kind cluster by docker which includes one real node and many fake nodes
```
JesseStutler, line 28: This is just one method of kwok installation, not mean to do in this way

docs/design/scheduler-simulate-design.md
```diff
@@ -0,0 +1,32 @@
...
+* queues.csv includes queueName, quota, etc...
+### simulate-result-output
+#### file
+* output file includes nodes_detail.csv, queues_detail.csv, pods_detail.csv, pods_detail.csv. These files will be put in output_path directory.
+* nodes_detail.csv includes ts, nodeName, cpuRequest, memoryRequest
+* queues_detail.csv includes ts, queueName, cpuRequest, memoryRequest
+* pods_detail.csv includes podName, createTs, scheduleTs, finishTs
+#### metrics
+* output metrics includes volcano_simulator_scheduler_wait_time, volcano_simulator_pod_run_time, volcano_simulator_node_cpu_request, volcano_simulator_node_memory_request which will help user to monitor the schduler simulate by prometheus.
+### time simulator
```
JesseStutler, line 24: I think there is no need to create a time simulator? kwok can do everything, https://kwok.sigs.k8s.io/docs/user/stages-configuration/, you can simulate whatever stage the pod is. Maybe we use kwok+shell script is a better way.

PR #3825, opened 2024-11-15
Preempt performance

molei20021:


lowang-bh:
There is some gaps with my idea to improve the performance. Could you add some describe about your design?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -326,6 +326,10 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
 		statusSets = append(statusSets, &api.Status{Code: api.Unschedulable, Reason: api.WrapInsufficientResourceReason(resources)})
 		return api.NewFitErrWithStatus(task, node, statusSets...)
 	}
+	if node.Allocatable.MaxTaskNum <= len(alloc.session.NodeMap[node.Name].Pods) {
```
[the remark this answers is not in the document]
molei20021: Bacause it may influence the predicate result if there are too many pod running on the node, so I put it before the new predicate cache to make the cache accurate.

molei20021:
> There is some gaps with my idea to improve the performance. Could you add some describe about your design?

preempt-performance
I add a design graph, the parts marked in red are modification points.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -230,5 +233,14 @@ func (ra *Action) Execute(ssn *framework.Session) {
 	}
 }
 
+func (ra *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
+	var statusSets api.StatusSets
+	if node.Allocatable.MaxTaskNum <= len(ra.session.NodeMap[node.Name].Pods) {
```
JesseStutler, line 238: Why not just add this in session's `PredicateForPreemptAction`?

pkg/scheduler/util/predicate_helper.go
```diff
@@ -12,6 +12,47 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/api"
 )
 
+type PredicateCache struct {
+	Cache map[api.JobID]map[api.TaskID]map[string]map[int64]error
```
JesseStutler, line 16: .....Too much map nesting. Please refactor as struct nesting and refactor your related code.

pkg/scheduler/cache/cache.go
```diff
@@ -1497,9 +1499,13 @@ func (sc *SchedulerCache) RecordJobStatusEvent(job *schedulingapi.JobInfo, updat
 			if len(msg) == 0 {
 				msg = baseErrorMessage
 			}
-			if err := sc.taskUnschedulable(taskInfo, reason, msg, nominatedNodeName); err != nil {
-				klog.ErrorS(err, "Failed to update unschedulable task status", "task", klog.KRef(taskInfo.Namespace, taskInfo.Name),
-					"reason", reason, "message", msg)
+			ts, exist := schedulingutil.GetPodStatusLastSetCache(job.UID, taskInfo.UID)
```
JesseStutler, line 1502: This part of code is used to reduce api requests sent to api-server? Need to split this as a separate commit, not to mit it into preemption performance improvement.

pkg/scheduler/cache/cache.go
```diff
@@ -1497,9 +1499,13 @@ func (sc *SchedulerCache) RecordJobStatusEvent(job *schedulingapi.JobInfo, updat
...
+			if !exist || nowTs-ts > 60 {
```
JesseStutler, line 1503: I remember that we discussed to set the interval that can be configured or using backoff. Direct set `60` here is 
too empirical.

pkg/scheduler/actions/allocate/allocate.go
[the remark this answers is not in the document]
JesseStutler: Sorry I didn't get the point, you didn't remove these codes in the predicate plugin, so why do we need to add duplicate codes here?

JesseStutler:
> > There is some gaps with my idea to improve the performance. Could you add some describe about your design?
> 
> preempt-performance I add a design graph, the parts marked in red are modification points.

I didn't get why we need to record last condition time, in SchedulerCache's `taskUnschedulable`, there are already codes that validate whether the condition is same as last time:
https://github.com/volcano-sh/volcano/blob/24ca00bd247e8add46d95cff62da127cc8d0ee5e/pkg/scheduler/cache/cache.go#L1026-L1033

molei20021:
I have fixed the problem you have mentioned above @JesseStutler

pkg/scheduler/framework/session.go
```diff
@@ -387,8 +389,23 @@ func (ssn *Session) GetUnschedulableAndUnresolvableNodesForTask(task *api.TaskIn
...
+	if node.Allocatable.MaxTaskNum <= len(ssn.NodeMap[node.Name].Pods) {
+		statusSets = append(statusSets, &api.Status{Code: api.Unschedulable, Reason: api.NodePodNumberExceeded})
+		return api.NewFitErrWithStatus(task, node, statusSets...)
+	}
+	var (
+		err error
+		ok  bool
+	)
+	if cacheEnable {
+		err, ok = util.GetPredicateCache(task.Job, task.UID, node.Name, node.HashValue)
```
lowang-bh, line 403: we can not simply use a cache here. For example, pod affinity/antiaffinity need to be re-considered.

lowang-bh:
We can split this pr to two parts:
part-1: reduce the input nodes scale in preempt/reclaim to avoid invalide predications.
part-2: add cache for same taskrole in job.

pkg/scheduler/framework/session.go
molei20021: I have added task hash which will consider the change of pod affinity/antiaffinity

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -243,10 +264,19 @@ func (pmpt *Action) preempt(
 		return false, fmt.Errorf("PrePredicate for task %s/%s failed for: %v", preemptor.Namespace, preemptor.Name, err)
 	}
 
-	predicateFn := ssn.PredicateForPreemptAction
+	predicateFn := pmpt.predicate
+	var allNodeIntersections []*api.NodeInfo
 	// we should filter out those nodes that are UnschedulableAndUnresolvable status got in allocate action
 	allNodes := ssn.GetUnschedulableAndUnresolvableNodesForTask(preemptor)
```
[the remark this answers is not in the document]
molei20021: but if there are many task in cluster and use binpack strategy, almost all the nodes will have preemptees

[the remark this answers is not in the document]
molei20021: but if there are many task in cluster and use leastrequested nodeorder strategy, almost all the nodes will have preemptees

Issue #3829, opened 2024-11-19
[feature] Integrate with kwok to simulate mock GPU/NPU nodes


JesseStutler:
### What is the problem you're trying to solve

In production, we may meet bugs related with GPU/NPU, if we have ensured that the bug came from scheduler, we have to reproduce the bugs to locate where the bug is. But in fact, we may not have a GPU or NPU environment to reproduce, so we need mock GPU/NPU nodes to help us. Besides, currently extended resources like GPU/NPU are just fields in capacity/allocatable field of node, if we just only need to verify the feature of scheduler, we may not care about if there are indeed real GPU/NPU nodes

### Describe the solution you'd like

We may follow this guide to write a shell to integrate kwok into volcano, if users need to mock NPU/GPU nodes, they can easily create a node yaml and test scheduling:
https://kwok.sigs.k8s.io/docs/user/kwok-manage-nodes-and-pods/

PR #3830, opened 2024-11-20
feature: integrate with kwok to simulate mock GPU/NPU nodes

JesseStutler:
### What type of PR is this?
/kind feature

related issue: #3829 
### verfication
1. Use `./create-fake-node.sh -n 10 -c 4 -m 8Gi -e volcano.sh/gpu-number=4,volcano.sh/gpu-memory=20` to create 10 fake nodes with 4 CPUs, 8Gi memories and extended resources with volcano.sh/gpu-number=4,volcano.sh/gpu-memory=20. After successfully creating these nodes, take one of them as an example:
image
image

2. Open `deviceshare` plugin and set the argument `deviceshare.GPUNumberEnable` enabled, and then create a fake deployment to create a pod requesting 1 `volcano.sh/gpu-number` and 1 `volcano.sh/gpu-memory`, successfully scheduled:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: fake-gpu-pod
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: fake-gpu-pod
  template:
    metadata:
      labels:
        app: fake-gpu-pod
    spec:
      schedulerName: volcano
      tolerations:
      - key: "kwok.x-k8s.io/node"
        operator: "Exists"
        effect: "NoSchedule"
      containers:
      - name: fake-container
        image: fake-image
        resources:
          limits:
            volcano.sh/gpu-number: 1
            volcano.sh/gpu-memory: 1
```
image

### DIFF
--- a/hack/kwok/README.md
+++ b/hack/kwok/README.md
@@ -1,0 +1,27 @@
+# How to use kwok to create fake nodes to simulate scheduling
+## Background
+- In production, we may meet bugs related with GPU/NPU scheduling, we have to reproduce to find out where the bugs are. But in fact, we may not have a GPU/NPU testing environment to reproduce, so create fake nodes can help us to simulate scheduling not only to do features validation about GPU/NPU/[other extended resources] scheduling, but also help us to reproduce to locate scheduler bugs.
+- Simulate scheduling can help us to do performance testing of scheduler, expecially for large-scale clusters, and kwok can create many fake nodes with very few resources.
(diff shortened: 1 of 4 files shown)


JesseStutler:
I found that under benchmark there is already a script to deploy kwok and fake nodes, maybe remove these into benchmark/kwok is better.

PR #3831, opened 2024-11-20
feat: support .Values.fullnameOverride

will4j:
Support `.Values.fullnameOverride` property of helm charts. 
People who want a custom resource name when install volcano with helm may find this useful.


will4j:
> Thank you for your contribution. I want to understand the usage scenarios of this PR. Please explain the necessity of this PR.

 Here's what happened to me, volcano is part of our ai platform, I have to deploy all components to a specific namespace (not volcano-system) with specific name prefix as our gpu cluster is strictly controlled

PR #3553, opened 2024-06-29
Fix predicate return

lowang-bh:
fixes #3086

Part-2 of [filter out those nodes which are UnschedulableAndUnresolvable when preempting #3432](https://github.com/volcano-sh/volcano/pull/3432)
Please merge 3432 first.

<details>

<summary> pod messages examples </summary>

image
</details>

PR #3840, opened 2024-11-26
Fix predicate return unexpected result

bibibox:
Recently, while using the new version of Volcano, I encountered several issues:
	1.	When the number of Pods on a node has reached its limit, a pending high-priority bestEffort Pod cannot preempt lower-priority bestEffort Pods on the node.
	2.	For Pods Pod A and Pod B with anti-affinity, if Pod A is already scheduled and the node has sufficient resources for Pod B, Pod B remains in a Pending state indefinitely.

Upon investigation, it was found that these issues are caused by inconsistent behavior between old and new versions of the Predicate plugin:
- Issue 1: In the old version, if the Predicate check for Pod num failed, it returned an Unschedulable status, which resulted in an error during the predicate phase of the backfill. In the new version, Predicate directly returns nil, allowing the predicate in backfill to pass, leading to the allocate phase. However, the Pod fails to bind due to the node’s Pod num limit.
- Issue 2: In the old version, when the Predicate check for node Pod anti-affinity failed, it returned an Unschedulable status with an error, causing the preempt action to exclude the node. In PR #3458, the error check was removed, and later, during PR #3553’s predicate refactoring, the logic inherited the status-only check. For most Unschedulable statuses, it is reasonable to ignore errors to enter the preempt phase. However, since Volcano currently does not support affinity checks during preemption, the Pod affinity scenario should return UnschedulableAndUnresolvable to skip the preempt phase.

This PR fixes the above issues by modifying the predicate function logic of the Predicate plugin. The changes are as follows:
- At the end of the predicate logic, if the predicateStatus list is not empty, return a wrapped fitError instead of nil.
- If the Pod affinity check fails, directly return UnschedulableAndUnresolvable.

Additionally, another logic issue was discovered while fixing the above problems. If Unschedulable occurs in the middle of the predicate logic, the process exits immediately, which might prevent potential UnschedulableAndUnresolvable scenarios later from being handled. This was uniformly modified so that only UnschedulableAndUnresolvable results cause an exit, while other cases allow the process to continue.


pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -481,6 +496,9 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				status := podAffinityFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
 				podAffinityStatus := api.ConvertPredicateStatus(status)
 				if podAffinityStatus.Code != api.Success {
+					// TODO: Currently, preemption is not supported when Pod affinity filtering fails.
```
wangyang0616, line 499: It is recommended that you register an issue that describes the todo and update the issue link to the todo comment.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -402,14 +409,16 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 			predicateStatus = append(predicateStatus, podsNumStatus)
 		}
 
-		predicateByStablefilter := func(pod *v1.Pod, nodeInfo *k8sframework.NodeInfo) ([]*api.Status, bool, error) {
+		predicateByStablefilter := func(nodeInfo *k8sframework.NodeInfo) ([]*api.Status, bool, error) {
 			// CheckNodeUnschedulable
 			predicateStatus := make([]*api.Status, 0)
 			status := nodeUnscheduleFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
```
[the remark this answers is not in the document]
bibibox: The pod parameter was not used in the function, so it has been removed.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -389,7 +389,14 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 		predicateStatus := make([]*api.Status, 0)
 		nodeInfo, found := nodeMap[node.Name]
 		if !found {
-			return api.NewFitError(task, node, "node info not found")
+			klog.V(4).Infof("NodeInfo predicates Task <%s/%s> on Node <%s> failed, node info not found",
+				task.Namespace, task.Name, node.Name)
+			nodeInfoStatus := &api.Status{
+				Code:   api.UnschedulableAndUnresolvable,
```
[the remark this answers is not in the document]
bibibox: Got it, it’s already updated.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -402,14 +409,16 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
-		predicateByStablefilter := func(pod *v1.Pod, nodeInfo *k8sframework.NodeInfo) ([]*api.Status, bool, error) {
+		predicateByStablefilter := func(nodeInfo *k8sframework.NodeInfo) ([]*api.Status, bool, error) {
 			// CheckNodeUnschedulable
 			predicateStatus := make([]*api.Status, 0)
 			status := nodeUnscheduleFilter.Filter(context.TODO(), state, task.Pod, nodeInfo)
 			nodeUnscheduleStatus := api.ConvertPredicateStatus(status)
 			if nodeUnscheduleStatus.Code != api.Success {
 				predicateStatus = append(predicateStatus, nodeUnscheduleStatus)
-				return predicateStatus, false, fmt.Errorf("plugin %s predicates failed %s", nodeUnscheduleFilter.Name(), status.Message())
+				if nodeUnscheduleStatus.Code == api.UnschedulableAndUnresolvable {
```
ouyangshengjia, line 419: I think this code should be encapsulated in a function like `predicateFail(status)` that means the `predicateFn` doesn't need to continue.
Besides, it can also exits immediately when `api.Error/api.Skip/api.Wait` occurred. It is better to call the same function `predicateFail(status)` at **allocate** and **preempt** phase(in function `session.PredicateForAllocateAction` and `session.PredicateForPreemptAction`).

bibibox: The statuses that should terminate the predicate phase have been refactored into a separate function. As for the logic where actions determine pass/fail based on the status set, I believe the current composition approach is a better choice.

Monokaix:
>2. For Pods Pod A and Pod B with anti-affinity, if Pod A is already scheduled and the node has sufficient resources for Pod B, Pod B remains in a Pending state indefinitely.

You mean pod B cannot be scheduled to other nodes?

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -525,15 +552,29 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
+
+		if len(predicateStatus) > 0 {
+			return api.NewFitErrWithStatus(task, node, predicateStatus...)
+		}
+
 		return nil
 	})
 }
 
+func shouldFailPredication(status *api.Status) bool {
```
Monokaix, line 571: A redundant logic with:
https://github.com/volcano-sh/volcano/blob/7170cca5a0af67938a05f46f3bf5095a127b13b9/pkg/scheduler/framework/session.go#L426
And can we just check there?

bibibox:
> > 2. For Pods Pod A and Pod B with anti-affinity, if Pod A is already scheduled and the node has sufficient resources for Pod B, Pod B remains in a Pending state indefinitely.
> 
> You mean pod B cannot be scheduled to other nodes?

In this example, assume the nodes where Pod A and Pod B are located rank first among the preemption target nodes. In this case, during each round of preemption attempts, the described scenario will occur, causing Pod B to remain in the Pending state indefinitely. Only when the node scores change might it be possible for Pod B to be scheduled onto another node.

pkg/scheduler/plugins/predicates/predicates.go
bibibox: The early return added to the predicate function is to prevent executing unnecessary operations further down the logic.

Additionally, the logic in the diagram evaluates stateSets, and I believe shouldFailPredication is an abstraction tailored to the needs of the predicate plugin itself. Its logic may change in the future, so wrapping it within the predicate function is a better choice.

lowang-bh:
Hi, I want a UT to cover the cases has been fixed.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -515,7 +540,9 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				podTopologyStatus := api.ConvertPredicateStatus(status)
 				if podTopologyStatus.Code != api.Success {
 					predicateStatus = append(predicateStatus, podTopologyStatus)
-					return api.NewFitErrWithStatus(task, node, predicateStatus...)
+					if shouldFailPredication(podTopologyStatus) {
```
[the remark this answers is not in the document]
bibibox: It has been renamed to `ShouldAbort`, and additional comments have been added.

bibibox:
> Hi, I want a UT to cover the cases has been fixed.

added one ut to cover this case

william-wang:
@bibibox Suggest log a issue first to describe the problem next time and associate the PR with the issue. It's more efficient for later searching and managing.

william-wang (approved):
/approve

lowang-bh:
> > Hi, I want a UT to cover the cases has been fixed.
> 
> added one ut to cover this case

Thanks a lot. What I want is a ut about case 2 described in your PR.

lowang-bh (commented):
/lgtm

PR #3849, opened 2024-11-29
Fix flaky ut

Monokaix:
Flaky ut err:
```shell
E1127 02:28:49.585051   36737 utils.go:43] init kubeclient in hamivgpu failed: stat /root/.kube/config: no such file or directory
--- FAIL: TestValidateJobCreate (2.84s)
    --- FAIL: TestValidateJobCreate/validate_valid-job (0.10s)
        admit_job_test.go:1270: Expect no error, but got error  unable to find job queue: queue.scheduling.volcano.sh "default" not found;
        admit_job_test.go:1273: Expect Allowed as true but got false. { false nil [] <nil> map[] []}
FAIL
```
Make sure the default queue exists in informer.

PR #3851, opened 2024-12-02
Fix flaky ut: create queues first

Monokaix:
Just like https://github.com/volcano-sh/volcano/pull/3849, fix Hierarchy queue tese case.


Monokaix:
cc @lowang-bh @hwdef @JesseStutler

PR #3853, opened 2024-12-04
Update docs and default config of scheduler

hwdef:
The default configuration should no longer use the proportion plugin

### DIFF
--- a/docs/design/job-resource-reservation-design.md
+++ b/docs/design/job-resource-reservation-design.md
@@ -5,6 +5,7 @@
 ## Deprecated
 
 The `Reservation` plugin has been deleted. This doc is deprecated. It will be deleted in the future.
+`Reservation` is replaced by [SLA](sla-plugin.md).
 
 ## Motivation
 As [issue 13](https://github.com/volcano-sh/volcano/issues/13) / [issue 748](https://github.com/volcano-sh/volcano/issues/748) 
--- a/installer/helm/chart/volcano/config/volcano-scheduler.conf
+++ b/installer/helm/chart/volcano/config/volcano-scheduler.conf
@@ -10,6 +10,6 @@ tiers:
   - name: drf
     enablePreemptable: false
   - name: predicates
-  - name: proportion
+  - name: capacity
   - name: nodeorder
   - name: binpack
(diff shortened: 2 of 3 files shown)


docs/design/proportional.md
```diff
@@ -1,3 +1,7 @@
+## Deprecated
+
+The `proportion` is deprecated. It is replaced by [capacity](capacity-scheduling.md).
```
Monokaix, line 3: `proportional` and `proportion` are different plugins: )
And `proportion` is actually not deprecated, `proportion` and `capacity` are just two ways using queue and cover different use cases.

installer/helm/chart/volcano/config/volcano-scheduler-ci.conf
```diff
@@ -11,6 +11,6 @@ tiers:
   - name: drf
     enablePreemptable: false
   - name: predicates
-  - name: proportion
```
Monokaix, line 14: This is a ci configuration and actually we have test both `proportion` and `capacity` plugins features.
And we have replaced the `proportion` to `capacity` when test e2e to cover the two plugins' feature.

docs/design/proportional.md
hwdef: The hierarchical queue function may not be compatible with proportion.

hwdef:
@lowang-bh @JesseStutler @Monokaix 
PTAL

docs/design/job-resource-reservation-design.md
```diff
@@ -5,6 +5,7 @@
 ## Deprecated
 
 The `Reservation` plugin has been deleted. This doc is deprecated. It will be deleted in the future.
+`Reservation` is replaced by [SLA](sla-plugin.md).
```
JesseStutler, line 8: I'd like to know what was the original reservation feature like? SLA plugin is related with reservation?

hwdef: They all try to solve the problem of hunger in large vcjob.
BTW, I don't know why the reservation plugin is removed

hwdef:
@JesseStutler PTAL

Issue #3856, opened 2024-12-05
The current testing method in the benchmark is incorrect


JesseStutler:
### Description

The comparison conditions between kube-scheduler and volcano in the benchmark are different. Volcano admission webhook and controller are also installed in the testing. Webhook will influence the pods creating speed of kcm. Volcano also adds two additional CRD resources, podgroup and queue. Volcano needs to update these additional resources when scheduling. And kube-scheduler did not use kueue/queue nor gang scheduling but volcano did. Besides, these benchmark didn't limit the resource usage of some components such as prometheus, it may cause some resource contention when scheduling. This test script is also not precise enough to wait for the scheduling results to be completed and plotted, requiring script modification or refactoring. 

### Describe the results you received and expected

Current volcano performance benchmark is incorrect.

### What version of Volcano are you using?

master

PR #3857, opened 2024-12-05
Delete current incorrect benchmark testing results

JesseStutler:
related with #3856


hwdef (commented):
Delete it directly? No modifications?

lowang-bh:
It is just added in https://github.com/volcano-sh/volcano/pull/3730, so why remove it?

JesseStutler:
> Delete it directly? No modifications?



> It is just added in #3730, so why remove it?

These results and testing methods are quite incorrect. The reason why it's incorrect is explained in this issue: https://github.com/volcano-sh/volcano/issues/3856. Currently we don't have much time to fix it, but we will arrange to re-test next year.

lowang-bh:
> > Delete it directly? No modifications?
> 
> > It is just added in #3730, so why remove it?
> 
> These results and testing methods are quite incorrect. The reason why it's incorrect is explained in this issue: #3856. Currently we don't have much time to fix it, but we will arrange to re-test next year.

/lgtm

Issue #3852, opened 2024-12-02
[Enhancement]Optimize volcano end-to-end scheduling large-scale pod performance


JesseStutler:
### What is the problem you're trying to solve

Recently, I used the scripts in benchmark/sh to compare the performance of volcano and kube-scheduler under benchmark. There are several problems that I found needed to be solved

### Describe the solution you'd like

- Traffic conflict optimization. The metric called `apiserver_current_inflight_requests` of kube-apiserver is significantly higher when volcano schedules large-scale pods than when kube-scheduler schedules, which may cause pressure on kube-apiserver and cause network conflicts, affecting scheduling performance. After reviewing the codes, I found there two places causing the problem:
    - [ ] At the end of each session, the session will update queue status's allocated field. If the session interval is short, there will be a lot of queue status update requests. And the session will also update the nums of pods in `Running/Failed/Succeed` states in podgroup status. If these fields of podgroups only used in kubectl to show the nums of pods in each state, we may rewrite the logic , not to persist these nums fields in podgroups, like this pr:  https://github.com/volcano-sh/volcano/pull/3751
    - [ ] If we open gang plugin, gang plugin will always refresh the Scheduled condition at the end of each session(Refresh the transition ID), it's a bug need to be fixed. In the logic of `isPodGroupStatusUpdated`, because the nums of pod in Running state is always refreshing in scheduling, therefore `!equality.Semantic.DeepEqual(newStatus, oldStatus)` is true, causing the condition always will be refreshed: https://github.com/volcano-sh/volcano/blob/7170cca5a0af67938a05f46f3bf5095a127b13b9/pkg/scheduler/framework/job_updater.go#L81-L88 We also need to integrate Unschedulable and Scheduled conditions into the same condition, and use different ConditionStatus to distinguish them. Otherwise, even if the pods in the podgroups have met the minmember, the Unschedulable condition will still remain in the conditions list of the podgroup. , which would cause ambiguity. 
- Performance bottleneck of volcano admission. During the test, I found that we choose to use deployment to create large-scale pods, volcano-admission will seriously affect the speed of kube-controller-manager to create pods. We may need to rewrite or further optimize the logic of validate and mutate. 
    - [ ] validate webhook improvement( use CEL? )
    - [ ] mutate webhook improvement
[...]

PR #3865, opened 2024-12-09
update pg status with applyStatus

jxs1211:
Improve the performance of update pg status with applyStatus api.
#3852

### DIFF
--- a/docs/design/podgroup-statistics.md
+++ b/docs/design/podgroup-statistics.md
@@ -10,7 +10,7 @@ resulting in accumulation of memory leak. See in issue #3597: https://github.com
 
 ## Alternative
 Currently the statistics of podgroups of each state are only used for display by vcctl, there is no need to be persisted in queue's status. 
-So when users need to use `vcctl queue get -n [name]` or `vcctl list` to display queues and each state of podgroups in queue, 
+So when users need to use `vcctl queue get -n [name]` or `vcctl queue list` to display queues and each state of podgroups in queue, 
 vcctl should calculate podgroup statistics in client side and then display them. And we can export these statistics of podgroups in each state as metrics.  
 
 ## Implementation
@@ -20,7 +20,7 @@ And `UpdateStatus` should not be used here: https://github.com/volcano-sh/volcan
 the `UpdateStatus` interface will verify the resourceVersion in the apiserver, which may cause concurrent update conflicts. 
 Instead, `ApplyStatus` should be used here to avoid this situation, because we only need to update the status of the queue. 
 It should be noted that the controller currently does not have patch queue status permissions, so we should add a patch queue/status permission to the clusterrole.
-- `vcctl get -n [name]` and `vcctl list` display the statistics of podgroups in each state from queue's status directly, 
+- `vcctl queue get -n [name]` and `vcctl queue list` display the statistics of podgroups in each state from queue's status directly, 
 instead we should do one more step, query the podgroups owend in the queue, stat the counts of podgroups in each state at the `vcctl` side, and then display them. 
 - Those metrics called `queue_pod_group_[state]_count`, which are recorded in proportion/capacity plugins in scheduler, 
 now will be moved to controller to record. The metrics involved are as follows:
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -350,10 +350,6 @@ func jobStatus(ssn *Session, jobInfo *api.JobInfo) scheduling.PodGroupStatus {
 		}
 	}
 
-	status.Running = int32(len(jobInfo.TaskStatusIndex[api.Running]))
-	status.Failed = int32(len(jobInfo.TaskStatusIndex[api.Failed]))
-	status.Succeeded = int32(len(jobInfo.TaskStatusIndex[api.Succeeded]))
-
 	return status
 }
 


pkg/scheduler/cache/cache.go
```diff
@@ -300,7 +301,22 @@ func (su *defaultStatusUpdater) UpdatePodGroup(pg *schedulingapi.PodGroup) (*sch
 		return nil, err
 	}
 
-	updated, err := su.vcclient.SchedulingV1beta1().PodGroups(podgroup.Namespace).Update(context.TODO(), podgroup, metav1.UpdateOptions{})
```
hwdef, line 303: I'm not sure if this modification is correct. I remember that this function will also update the annotation and label, so it is wrong to only update the status.

jxs1211: > we may rewrite the logic , not to persist these nums fields in podgroups, like this pr: https://github.com/volcano-sh/volcano/pull/3751
I think the proposal is want to not update the status fields in podgroups, do you mean the annotation and labels still need to be updated?

jxs1211:
> Could you help to research that are there some other tools that can help kubectl calculate the number of running/failed/succeeded pods in the podgroup and then display these fields, but not by directly reading these fields in the status for display? Or if this cannot be achieved, we can use vcctl to display them, but I think more users might like to use kubectl.

Yes, do you have any suggestion on a specific tool or method to achieve that, I don't have any idea for that by now.

JesseStutler:
> > Could you help to research that are there some other tools that can help kubectl calculate the number of running/failed/succeeded pods in the podgroup and then display these fields, but not by directly reading these fields in the status for display? Or if this cannot be achieved, we can use vcctl to display them, but I think more users might like to use kubectl.
> 
> Yes, do you have any suggestion on a specific tool or method to achieve that, I don't have any idea for that by now.

You may refer to:
- https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/ 
- https://medium.com/@platform.engineers/developing-and-using-kubectl-plugins-a-hands-on-guide-1ee7aa4aea20

Use kubectl plugin may help us still show the column running/failed/succeeded, but we don't need to persist these fields in vc-controller, but calculate dynamically by counting running/failed/succeeded pods in kubectl plugin.

JesseStutler:
> > > Could you help to research that are there some other tools that can help kubectl calculate the number of running/failed/succeeded pods in the podgroup and then display these fields, but not by directly reading these fields in the status for display? Or if this cannot be achieved, we can use vcctl to display them, but I think more users might like to use kubectl.
> > 
> > 
> > Yes, do you have any suggestion on a specific tool or method to achieve that, I don't have any idea for that by now.
> 
> You may refer to:
> 
> * https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/
> * https://medium.com/@platform.engineers/developing-and-using-kubectl-plugins-a-hands-on-guide-1ee7aa4aea20
> 
> Use kubectl plugin may help us still show the column running/failed/succeeded, but we don't need to persist these fields in vc-controller, but calculate dynamically by counting running/failed/succeeded pods in kubectl plugin.

I think we can implement it in vcctl first, later we can decide whether we should implement a kubectl plugin.

jxs1211:
> > > > Could you help to research that are there some other tools that can help kubectl calculate the number of running/failed/succeeded pods in the podgroup and then display these fields, but not by directly reading these fields in the status for display? Or if this cannot be achieved, we can use vcctl to display them, but I think more users might like to use kubectl.
> > > 
> > > 
> > > Yes, do you have any suggestion on a specific tool or method to achieve that, I don't have any idea for that by now.
> > 
> > 
> > You may refer to:
> > 
> > * https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/
> > * https://medium.com/@platform.engineers/developing-and-using-kubectl-plugins-a-hands-on-guide-1ee7aa4aea20
> > 
> > Use kubectl plugin may help us still show the column running/failed/succeeded, but we don't need to persist these fields in vc-controller, but calculate dynamically by counting running/failed/succeeded pods in kubectl plugin.
> 
> I think we can implement it in vcctl first, later we can decide whether we should implement a kubectl plugin.

I found the doc related for this, is it originated from the doc?
volcano/docs/design/podgroup-statistics.md

JesseStutler:
> > > > > Could you help to research that are there some other tools that can help kubectl calculate the number of running/failed/succeeded pods in the podgroup and then display these fields, but not by directly reading these fields in the status for display? Or if this cannot be achieved, we can use vcctl to display them, but I think more users might like to use kubectl.
> > > > 
> > > > 
> > > > Yes, do you have any suggestion on a specific tool or method to achieve that, I don't have any idea for that by now.
> > > 
> > > 
> > > You may refer to:
> > > 
> > > * https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/
> > > * https://medium.com/@platform.engineers/developing-and-using-kubectl-plugins-a-hands-on-guide-1ee7aa4aea20
> > > 
> > > Use kubectl plugin may help us still show the column running/failed/succeeded, but we don't need to persist these fields in vc-controller, but calculate dynamically by counting running/failed/succeeded pods in kubectl plugin.
> > 
> > 
> > I think we can implement it in vcctl first, later we can decide whether we should implement a kubectl plugin.
> 
> I found the doc related for this `volcano/docs/design/podgroup-statistics.md`, and the feature mentioned in the doc has been implemented, correct me if I'm wrong.

No, running/pending/unknown....fields in queue status are not persisted now, but there are still fields in podgroups Running/Failed/Succeeded that need to be refreshed every session, I think they does not need to be persisted also, you can check here:
https://github.com/volcano-sh/volcano/blob/0966fd59a9e0c0cfa7f6a19d41b3263f932ba4e0/pkg/scheduler/framework/session.go#L353-L355

jxs1211:
> volcano/pkg/scheduler/framework/session.go

I removed it, and do we need to also remove the `ssn.podGroupStatus` ?

JesseStutler:
> > volcano/pkg/scheduler/framework/session.go
> 
> I removed it.

It's not enough, there may some users need to use vcctl to check these fields, we need to query the pods under podgroups, similar to this PR: https://github.com/volcano-sh/volcano/pull/3751,and then calculate the number of pods for display.

Also we need to check carefully whether if these fields is depended by some plugins or actions

jxs1211:
> > > volcano/pkg/scheduler/framework/session.go
> > 
> > 
> > I removed it.
> 
> It's not enough, there may some users need to use vcctl to check these fields, we need to query the pods under podgroups, similar to this PR: #3751 then calculate the number of pods for display.
> 
> Also we need to check carefully whether if these fields is depended by some plugins or actions

Do you mean we need to do the same thing at the removed code like PR: #3751 did, don't we try to offload the apiserver's pressure in session, do I misunderstand?

JesseStutler:
> > > > volcano/pkg/scheduler/framework/session.go
> > > 
> > > 
> > > I removed it.
> > 
> > 
> > It's not enough, there may some users need to use vcctl to check these fields, we need to query the pods under podgroups, similar to this PR: #3751 then calculate the number of pods for display.
> > Also we need to check carefully whether if these fields is depended by some plugins or actions
> 
> Do you mean we need to do the same thing at the removed code like PR: #3751 did, don't we try to offload the apiserver's pressure in session, do I misunderstand? Another things I want to make sure is we're discussing removed the field(Running/Failed/Succeeded) in PodGroupStatus or just remove any reference of them(like the code was removed)?

Yes, we need to do the same thing as #3751, remove the three fields of PodGroupStatus. If these three fields are dependent on other code, we need to use other methods to transform the dependent code (for example, get the Running/Succeeded/Failed pods under the podgroup)

JesseStutler:
So the right things are:
1. Remove the code that updates the three fields (we keep the API unchanged, just remove the code that updates these three fields is fine)
2. Adapt the code that depends on these three fields (including whether the code in the scheduler/controller/e2e test depends on these three fields). If the logic written depends on these three fields in PodGroupStatus, we need to adapt them.

PR #3867, opened 2024-12-10
fix hami bug when vgpu-number no set

weapons97:
In the Hami project [volcano-vgpu-device-plugin](https://github.com/Project-HAMi/volcano-vgpu-device-plugin) , if the user does not set volcano.sh/vgpu-number, it will cause a bug where the plugin does not respond to container resource mounts. Therefore, the check in checkVGPUResourcesInPod needs to be modified.
related issue:
https://github.com/Project-HAMi/volcano-vgpu-device-plugin/issues/40

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -1738,8 +1738,8 @@ func TestAllocateWithPVC(t *testing.T) {
 			PVs:                []*v1.PersistentVolume{pv1, pv2},
 			PVCs:               []*v1.PersistentVolumeClaim{pvc1, pvc2},
 			IgnoreProvisioners: ignoreProvisioners,
-			ExpectStatus: map[api.JobID]scheduling.PodGroupPhase{
-				"c1/pg1": scheduling.PodGroupRunning,
+			ExpectTaskStatusNums: map[api.JobID]map[api.TaskStatus]int{
+				"c1/pg1": {api.Binding: 2},
 			},
 		},
 	}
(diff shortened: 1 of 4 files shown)


archlitchi:
Yes, i think sooner or later, we need to add a mutatingwebhookConfiguration for volcano.sh/vgpu-memory, manually add a volcano.sh/vgpu-number: 1 to these tasks.

Since it doesn't conflict with this fix, so i'll vote approve

pkg/scheduler/api/devices/nvidia/vgpu/utils.go
```diff
@@ -180,11 +180,7 @@ func decodePodDevices(str string) []ContainerDevices {
 
 func checkVGPUResourcesInPod(pod *v1.Pod) bool {
 	for _, container := range pod.Spec.Containers {
-		_, ok := container.Resources.Limits[VolcanoVGPUMemory]
-		if ok {
-			return true
-		}
-		_, ok = container.Resources.Limits[VolcanoVGPUNumber]
+		_, ok := container.Resources.Limits[VolcanoVGPUNumber]
```
Monokaix, line 183: What if user just set VolcanoVGPUNumber but not set VolcanoVGPUMemory？

Monokaix:
Please add ut to cover all possible cases.

pkg/scheduler/api/devices/nvidia/vgpu/utils.go
archlitchi: it will try to find an unused GPU, and use that exclusively, if such can't be find, then it will be stuck in pending state.

Issue #3869, opened 2024-12-11
reclaim action uses `HasPendingTasks` instead of `JobStarving`, which may cause the two jobs to reclaim with each other and lead to deadlock.


JesseStutler:
### Description

Such a scenario: the number of replicas of the two jobs is 5, minAvailable is 1, we can not deploy 5 replicas at the same time on the cluster, there will be at least one pod pending, and then the reclaim will be triggered, and then the two jobs will keep reclaiming each other, leading to deadlock. 

### Steps to reproduce the issue

1. create a new queue:
```yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: a
spec:
  reclaimable: true
  weight: 1
```
2. create a new job in default queue, 5 replicas, 1 minAvailable, all pods can not scheduled simultaneously on clusters.
```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: testjoba1
spec:
  minAvailable: 1
  schedulerName: volcano
  queue: default
  tasks:
    - replicas: 5
      name: "default-nginx"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: nginx:1.27-alpine3.19-slim
              imagePullPolicy: Never
              name: nginx
              resources:
                requests:
                  cpu: 7
          restartPolicy: OnFailure    
```
3.create a new job in a queue, 5 replicas, 1 minAvailable:
```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: testjobb1
spec:
  minAvailable: 1
  schedulerName: volcano
  queue: a
  tasks:
    - replicas: 5
      name: "default-nginx"
      template:
        metadata:
          name: web
        spec:
          containers:
            - image: nginx:1.27-alpine3.19-slim
              imagePullPolicy: Never
              name: nginx
              resources:
                requests:
                  cpu: 7
          restartPolicy: OnFailure   
```

### Describe the results you received and expected

Deploy job-a first, occupying cluster resources, and then deploy job-b, triggering reclaim to evict job-a's pod resources. After job-a pod becomes pending, reclaim reclaim job-b's pod for job-a, resulting dead-lock.

We may change reclaim to use `JobStarving`, if a job has enought pods running >= minAvailable, we don't trigger reclaim action for it.

### What version of Volcano are you using?

v1.10

JesseStutler:
[...]

PR #3870, opened 2024-12-11
reclaim: When choosing a preemptor, choose a starving one rather than one with pending tasks.

hwdef:
fix: #3869


JesseStutler:
Also need to fix the TestReclaim UT. The failed e2e testing scheduling action may not related with this pr.

JesseStutler:
@Monokaix Neet to carry on this pr and related issue into v2.0

JesseStutler:
Already inherited by #3951 
/close

Issue #3873, opened 2024-12-12
Volcano scheduler: Supports network topology aware scheduling in allocate action


Monokaix:
ref: https://github.com/volcano-sh/volcano/pull/3850
Modify `allocate` action and support scheduling jobs by hyperNodes and select a best one.

PR #3850, opened 2024-11-29
add network-topology-aware design doc

william-wang:
This pr includes a proposal to support the network topology management and network aware scheduling in Volcano. A new network topology CR is designed to define the spine-leaf network in an explicit way. It is also compatible the traditional way using labels.

PR #3874, opened 2024-12-12
Network topology scheduling implementations of volcano scheduler

Monokaix:
Ref: [#3873](https://github.com/volcano-sh/volcano/issues/3873), https://github.com/volcano-sh/volcano/pull/3850, please merge https://github.com/volcano-sh/apis/pull/144 first.
It includes three parts:
1.  Generate CRD yaml.
2. `Allocate` changes to support scheduler jobs with network topology constrains.
3. Add nodeOrder score result to hyperNode and get the final scores.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -184,6 +184,10 @@ func (alloc *Action) allocateResources(queues *util.PriorityQueue, jobsMap map[a
 	}
 }
 
+func (alloc *Action) allocateResourceForTasksWithTopology(tasks *util.PriorityQueue, job *api.JobInfo, jobs *util.PriorityQueue, queue *api.QueueInfo, allNodes []*api.NodeInfo) {
+
+}
+
 func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *api.JobInfo, jobs *util.PriorityQueue, queue *api.QueueInfo, allNodes []*api.NodeInfo) {
 	ssn := alloc.session
 	stmt := framework.NewStatement(ssn)
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -17,6 +17,7 @@
 package allocate
 
 import (
+	"sort"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -33,11 +34,13 @@ type Action struct {
 	session *framework.Session
 	// configured flag for error cache
 	enablePredicateErrorCache bool
+	hyperNodesTiers           []int
 }
 
 func New() *Action {
 	return &Action{
 		enablePredicateErrorCache: true, // default to enable it
+		hyperNodesTiers:           []int{},
 	}
 }
 
@@ -52,11 +55,26 @@ func (alloc *Action) parseArguments(ssn *framework.Session) {
 	arguments.GetBool(&alloc.enablePredicateErrorCache, conf.EnablePredicateErrCacheKey)
 }
 
+func (alloc *Action) parseHyperNodesTiers(ssn *framework.Session) {
+	if ssn.HyperNodesListByTier == nil || len(ssn.HyperNodesListByTier) == 0 {
+		return
+	}
+
+	// sort to guarantee the traverse order is from down to top.
+	var tiers []int
+	for tier := range ssn.HyperNodesListByTier {
+		tiers = append(tiers, tier)
+	}
+	sort.Ints(tiers)
+	alloc.hyperNodesTiers = tiers
+}
+
 func (alloc *Action) Execute(ssn *framework.Session) {
 	klog.V(5).Infof("Enter Allocate ...")
 	defer klog.V(5).Infof("Leaving Allocate ...")
 
 	alloc.parseArguments(ssn)
+	alloc.parseHyperNodesTiers(ssn)
 
 	// the allocation for pod may have many stages
 	// 1. pick a queue named Q (using ssn.QueueOrderFn)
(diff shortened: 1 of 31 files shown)


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -228,101 +385,123 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
...
-	if ssn.JobReady(job) {
-		stmt.Commit()
-	} else {
-		if !ssn.JobPipelined(job) {
-			stmt.Discard()
+func (alloc *Action) allocateResourcesForTask(stmt *framework.Statement, task *api.TaskInfo, node *api.NodeInfo, job *api.JobInfo) {
+	// Allocate idle resource to the task.
+	if task.InitResreq.LessEqual(node.Idle, api.Zero) {
+		klog.V(3).Infof("Binding Task <%v/%v> to node <%v>", task.Namespace, task.Name, node.Name)
+		if err := stmt.Allocate(task, node); err != nil {
```
JesseStutler, line 482: Missed UnAllocate? The master branch now has merged UnAllocate here:
https://github.com/volcano-sh/volcano/blob/dca41600e51e5aa667396eed28c8a52db718b981/pkg/scheduler/actions/allocate/allocate.go#L282-L288

Monokaix: This is a seperate branch, and will rebase master when merged into master.

pkg/scheduler/plugins/drf/drf.go
```diff
@@ -157,13 +157,7 @@ func (drf *drfPlugin) compareQueues(root *hierarchicalNode, lqueue *api.QueueInf
...
 	rnode := root
 	rpaths := strings.Split(rqueue.Hierarchy, "/")
-	depth := 0
-	if len(lpaths) < len(rpaths) {
-		depth = len(lpaths)
-	} else {
-		depth = len(rpaths)
-	}
-	for i := 0; i < depth; i++ {
+	for i, depth := 0, min(len(lpaths), len(rpaths)); i < depth; i++ {
```
JesseStutler, line 160: Should not contain these codes, which has been already merged and not related with this pr, may need to rebase

pkg/scheduler/uthelper/helper.go
```diff
@@ -180,6 +185,14 @@ func (test *TestCommonStruct) CheckBind(caseIndex int) error {
 		}
 	}
 
+	if test.MinimalBindCheck {
```
JesseStutler, line 188: What's the reason add a new field MinimalBindCheck here and move the code:
```go
if test.ExpectBindsNum != len(test.ExpectBindMap) {
		return fmt.Errorf("invalid setting for binding check: want bind count %d, want bind result length %d", test.ExpectBindsNum, len(test.ExpectBindMap))
	}
```
to the back?

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -85,7 +85,7 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 			nodeScores := util.PrioritizeNodes(task, predicateNodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
 			node = ssn.BestNodeFn(task, nodeScores)
 			if node == nil {
-				node = util.SelectBestNode(nodeScores)
+				node, _ = util.SelectBestNodeAndScore(nodeScores)
```
JesseStutler, line 88: What's the reason change the SelectBestNode --> SelectBesetNodeAndScore in backfill?

pkg/scheduler/util/scheduler_helper_test.go
```diff
@@ -17,9 +17,12 @@ limitations under the License.
 package util
 
 import (
+	"k8s.io/apimachinery/pkg/util/sets"
```
JesseStutler, line 20: Need to rearrange the dependencies

pkg/scheduler/plugins/drf/drf.go
Monokaix: rebased.

pkg/scheduler/uthelper/helper.go
Monokaix: Because sometimes we just want check the binding num and no need to check ExpectBindMap, because we may not know the exact bind result(pod->node), and also this value is false by default, so it will not break the existing UTs.

pkg/scheduler/util/scheduler_helper_test.go
Monokaix: done

pkg/scheduler/framework/statement.go
```diff
@@ -440,3 +441,70 @@ func (s *Statement) Commit() {
...
+		case Pipeline:
+			err := s.Pipeline(op.task, op.task.NodeName, false)
+			if err != nil {
+				klog.Errorf("Failed to pipeline task: %s", err.Error())
+				return err
+			}
+		case Allocate:
+			node := s.ssn.Nodes[op.task.NodeName]
+			err := s.Allocate(op.task, node)
+			if err != nil {
```
JesseStutler, line 481: I think there is a problem here. It is possible that the previous task allocation was successful and the status in the session was changed, but a subsequent task failed. Only the failed task was rolled back and an error was returned. The returned statement has not been committed outside. At this time, it will be inconsistent with the status in the cluster, which will affect the scheduling of subsequent jobs.

Monokaix: It will only take effect in session and no commit so cache will not be affected.

pkg/scheduler/framework/statement.go
```diff
@@ -440,3 +441,70 @@ func (s *Statement) Commit() {
...
+func (s *Statement) outputOperations(msg string, level klog.Level) {
+	if !klog.V(level).Enabled() {
+		return
+	}
+
+	var buffer string
+	for _, op := range s.operations {
+		switch op.name {
+		case Evict:
+			buffer += fmt.Sprintf("task %s evict from node %s ", op.task.Name, op.task.NodeName)
```
lowang-bh, line 502: Does it need a ";" or "," to split each operation in log?

pkg/scheduler/uthelper/helper.go
lowang-bh: This return should put after the next select case, in case that if ExpectBindsNum is 0, but actually there are binds.

pkg/scheduler/uthelper/helper.go
```diff
@@ -168,9 +176,6 @@ func (test *TestCommonStruct) CheckAll(caseIndex int) (err error) {
 
 // CheckBind check expected bind result
 func (test *TestCommonStruct) CheckBind(caseIndex int) error {
-	if test.ExpectBindsNum != len(test.ExpectBindMap) {
```
lowang-bh, line 171: We'd better check it first to expose invalid setting , and if test.MinimalBindCheck is true, skip this check.

pkg/scheduler/uthelper/helper.go
```diff
@@ -180,6 +185,14 @@ func (test *TestCommonStruct) CheckBind(caseIndex int) error {
 		}
 	}
 
+	if test.MinimalBindCheck {
+		return nil
```
lowang-bh, line 189: This return should put after the next select case, in case that if ExpectBindsNum is 0, but actually there are binds.

pkg/scheduler/uthelper/helper.go
Monokaix: When users want minimal check, they will not specify `ExpectBindMap`, so we can't check `ExpectBindMap` directly.

pkg/scheduler/uthelper/helper.go
Monokaix: done

pkg/scheduler/framework/statement.go
Monokaix: There is aleady a space: )

pkg/scheduler/uthelper/helper.go
Monokaix: modified.

PR #3879, opened 2024-12-12
Remove duplicated codes

kerthcet:


archlitchi:
i checked the API from pkg util, indeed duplicated

/lgtm

PR #3886, opened 2024-12-13
fix: nodegroup return nil 3880

### LINKED ISSUE #3880: predicateFn of NodeGroup plugin disable all predicateFn functions of all other subsequent plugins
### Description

When scheduler ConfigMap is defined as follows:
```
actions: "reclaim, allocate, backfill, preempt"
tiers:
- plugins:
  - name: priority
  - name: gang
  - name: conformance
- plugins:
  - name: nodegroup
  - name: predicates
```
The ``predicateFn `` in plugin predicates will not work.

1. Add a Node with labels: ``arch=amd64``
2. modify the configmap so the plugin nodegroup is before plugin predicates.
3. Add a pod with NodeSelector ``arch=arm64``

Based on the k8s scheduling rules, this pod should not be scheduled in this Node.
But the score of this node will appear in the log/

The pods with specified NodeSelector should not be scheduled on nodes with mismatched labels.
In other words, plugin predicates should work.

1.10

The reason of this bug is because the return of ``predicateFn`` in nodegroup is always error,  as following:
```go
		predicateStatus := make([]*api.Status, 0)

		group := node.Node.Labels[NodeGroupNameKey]
		queue := GetPodQueue(task)
		if err := queueGroupAffinity.predicate(queue, group); err != nil {
			nodeStatus := &api.Status{
				Code:   api.UnschedulableAndUnresolvable,
				Reason: "node not satisfy",
			}
			predicateStatus = append(predicateStatus, nodeStatus)
			return api.NewFitErrWithStatus(task, node, predicateStatus...)
		}
		klog.V(4).Infof("task <%s>/<%s> queue %s on node %s of nodegroup %v", task.Namespace, task.Name, queue, node.Name, group)
		nodeStatus := &api.Status{
			Code:   api.Success,
			Reason: "node satisfy task",
			Plugin: PluginName,
		}
		predicateStatus = append(predicateStatus, nodeStatus)
		return api.NewFitErrWithStatus(task, node, predicateStatus...)
	}
```

Which leads to return and skip the ``predicateFn`` of all other plugins, with plugin predicates included.
```go
// PredicateFn invoke predicate function of the plugins
func (ssn *Session) PredicateFn(task *api.TaskInfo, node *api.NodeInfo) error {
			if !isEnabled(plugin.EnabledPredicate) {
				continue
			}
			pfn, found := ssn.predicateFns[plugin.Name]
			if !found {
				continue
			}
			err := pfn(task, node)
			if err != nil {
				return err
			}
		}
	}
	return nil
}
```

However, the correct return of nodegroup will not trigger a error return when revoking:
```go
// PredicateForAllocateAction checks if the predicate error contains
// - Unschedulable
// - UnschedulableAndUnresolvable
// - ErrorSkipOrWait
func (ssn *Session) PredicateForAllocateAction(task *api.TaskInfo, node *api.NodeInfo) error {
	err := ssn.PredicateFn(task, node)
	if err == nil {
		return nil
	}

	fitError, ok := err.(*api.FitError)
	if !ok {
		return api.NewFitError(task, node, err.Error())
	}

	statusSets := fitError.Status
	if statusSets.ContainsUnschedulable() || statusSets.ContainsUnschedulableAndUnresolvable() ||
		statusSets.ContainsErrorSkipOrWait() {
		return fitError
	}
	return nil
}
```

So the wrong node will be added as predicateNodes and something will go wrong.
```go
		// TODO (k82cn): Enable eCache for performance improvement.
		if err := fn(task, node); err != nil {
			klog.V(3).Infof("Predicates failed: %v", err)
			errorLock.Lock()
			nodeErrorCache[node.Name] = err
			ph.taskPredicateErrorCache[taskGroupid] = nodeErrorCache
			fe.SetNodeError(node.Name, err)
			errorLock.Unlock()
			return
		}

		//check if the number of found nodes is more than the numNodesTofind
		length := atomic.AddInt32(&numFoundNodes, 1)
		if length > numNodesToFind {
			cancel()
			atomic.AddInt32(&numFoundNodes, -1)
		} else {
			predicateNodes[length-1] = node
		}
```

What I want to know is that, Is this process designed to be like this?
If not, I'd be happy to submit a PR to fix it

lut777:
fix #3880

### DIFF
--- a/pkg/scheduler/plugins/nodegroup/nodegroup.go
+++ b/pkg/scheduler/plugins/nodegroup/nodegroup.go
@@ -239,13 +239,7 @@ func (np *nodeGroupPlugin) OnSessionOpen(ssn *framework.Session) {
 			return api.NewFitErrWithStatus(task, node, predicateStatus...)
 		}
 		klog.V(4).Infof("task <%s>/<%s> queue %s on node %s of nodegroup %v", task.Namespace, task.Name, queue, node.Name, group)
-		nodeStatus := &api.Status{
-			Code:   api.Success,
-			Reason: "node satisfy task",
-			Plugin: PluginName,
-		}
-		predicateStatus = append(predicateStatus, nodeStatus)
-		return api.NewFitErrWithStatus(task, node, predicateStatus...)
+		return nil
 	}
 
 	ssn.AddPredicateFn(np.Name(), predicateFn)


lowang-bh (commented):
/lgtm

PR #3891, opened 2024-12-13
fix: configmgr need wait for all processed avoid tests data race

### LINKED ISSUE #3884: data race in configmgr_test.go
### Description

image

1.
2.
3.

need to add rwlock

master

https://github.com/volcano-sh/volcano/actions/runs/12234141826/job/34125702223?pr=3862

zedongh:
configmgr `Start` test need ensure all queue source be processed then do `assert` action. avoid call `l.called` when it write at `SyncConfig`.

@lowang-bh  
/close #3884

### DIFF
--- a/pkg/agent/config/source/fake_source.go
+++ b/pkg/agent/config/source/fake_source.go
@@ -46,7 +46,7 @@ func (f *fakeSource) GetLatestConfig() (cfg *api.ColocationConfig, err error) {
 func (f *fakeSource) Stop() {
 	for {
 		if f.queue.Len() == 0 {
-			f.queue.ShutDown()
+			f.queue.ShutDownWithDrain()
 			return
 		}
 		time.Sleep(1 * time.Second)


pkg/agent/config/configmgr.go
```diff
@@ -166,8 +168,9 @@ func (m *ConfigManager) Start(ctx context.Context) error {
 	if err = m.notifyListeners(); err != nil {
 		klog.ErrorS(err, "Failed to notify all listeners, retry by process chanagesQueue")
 	}
-
+	m.wg.Add(1)
```
lowang-bh, line 171: I think we'd better not change the main code. we can solve it in `FakeListener`.

zedongh: > I think we'd better not change the main code. we can solve it in `FakeListener`.

Yeah. but for this type of test, it is necessary to wait until all messages have been processed by the listeners before proceeding with the assertion. In this scenario, the listener lacks a mechanism to recognize when it will no longer be invoked.

pkg/agent/config/configmgr_test.go
```diff
@@ -301,6 +301,9 @@ func TestStart(t *testing.T) {
 		}
 		actualErr := mgr.Start(context.TODO())
 		s.Stop()
+		// Ensure all of source queue objects are processed then do check the result action
```
Monokaix, line 304: There is already a `s.Stop()` in line 304 to wait all queue keys processed,so it's not necessary to re-check and wait here.

Monokaix:
seems add a  rwlock is enough: )

lowang-bh:
> seems add a rwlock is enough: )

Agree!

zedongh:
> > seems add a rwlock is enough: )
> 
> Agree!

**Thanks a lot. After review the code of fake source, I make last simple commit in `Stop()` replace `queue.Shutdown` with `queue.ShutdownWithDrain`. Let me to cleaify the reason.**


I believe there should be no concurrency here, so it is not suitable to use locks. The only scenario where concurrency occurs is as follows:


Mgr.Start goroutine
```go
[Order 1]   key, quit := queue.Get()  // consume item and queue size -1

[Order 3a]  m.notifyListeners()
```

Test goroutine
```go
[Order 2]. s.Stop() // queue length is zero, break loop, queue Shutdown, but item not Done yet
          // if f.queue.Len() == 0 {
          //  f.queue.ShutDown()
          //  return
          // }

[Order 3b] assert.Equal(t, tc.expectedNotifyCalled, l.called)
```

**3a and 3b have no order guarantee.** 

Simple to make origin test failed is set a time out in `SyncConfig`
```go
func (l *FakeListener) SyncConfig(cfg *api.ColocationConfig) error {
        // time out will make this calling after testing assert, make it failed
	// time.Sleep(1*time.Second + 500*time.Millisecond)
	l.called++
```

lowang-bh:
As the issue described，`l.called++` at line 224 and `l.called` at line 305 has a read & write conflict.

zedongh:
> As the issue described，`l.called++` at line 224 and `l.called` at line 305 has a read & write conflict.

Yeah, the reason is fakeSource `Stop` use `queue.Shutdown` not `queue.ShutdownWithDrain` cause the last item data race. So just fixed the `Stop` rather than add a rwlock.

JesseStutler:
Can we push forward this PR? I also meet this e2e failing frequently @hwdef @Monokaix

PR #3895, opened 2024-12-16
Fix PodTopologySpread in predicate

bibibox:
This is a supplement to #3840


pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -483,10 +483,12 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 				status := nodePortFilter.Filter(context.TODO(), state, nil, nodeInfo)
 				nodePortStatus := api.ConvertPredicateStatus(status)
 				if nodePortStatus.Code != api.Success {
+					// TODO: Currently, preemption is not supported when NodePort filtering fails.
+					// Once supported, the logic here should be removed.
+					// See https://github.com/volcano-sh/volcano/issues/3845
+					nodePortStatus.Code = api.UnschedulableAndUnresolvable
```
wangyang0616, line 489: Why does a node port not support preemption?

bibibox: Similar to inter-pod affinity, passing resource comparison during preemption does not guarantee that the pod occupying the node’s port will be preempted.

JesseStutler:
Do we still need it?

Issue #3889, opened 2024-12-13
[Umbrella] Proposed to add a descheduler component to enhance descheduling capabilities


Monokaix:
### What is the problem you're trying to solve

Resource utilization and load balancing are issues that users are more concerned about, these are very dependent on the capabilities of the scheduler. Volcano now provides stable scheduling capabilities, but scheduling is usually static, and node resources will change dynamically, so it is necessary to re-scheduler and the scheduler cooperates to achieve the best use effect and meet user resource utilization and load balancing needs.

Some previous discussions are here:
https://github.com/volcano-sh/volcano/issues/3300, https://github.com/volcano-sh/volcano/issues/3023

### Describe the solution you'd like

Although volcano now has a rescheduling plugin that supports rescheduling, ref https://github.com/volcano-sh/volcano/issues/1955, https://github.com/volcano-sh/volcano/pull/1923, there are the following problems:

1. There are relatively few plug-in algorithms and currently only supports utilization-based rescheduling.
2. Rescheduling is a relatively dangerous behavior because it may affect business stability and SLA. Therefore, some strict and fine-grained verification needs to be done before rescheduling and eviction, such as whether the node resources are sufficient and the node affinity is considered, as well as pod replica constraints, etc., so as to ensure that the rescheduling behavior is safe and the newly created pods can be scheduled correctly. However, the current rescheduling cannot provide fine-grained pre-eviction control and inspection.
3. The currently existing rescheduling component https://github.com/kubernetes-sigs/descheduler is very powerful. We can carry out further development based on it and support more functions, such as rescheduling based on real loads to solve the problem  mentioned in #3300
4. Re-scheduling and scheduling are two independent behaviors. Adding a re-scheduling plugin to the scheduler will cause code coupling and mixed structures, making subsequent maintenance difficult.

### Additional context

The design detail: https://docs.google.com/document/d/1AsCpmjLzJ2iDS1LFcVJDmMryIzemGhN0toiuHXLbyhc/edit?tab=t.0#heading=h.55km60qjnsi, which is derived from  #3300

hwdef:
+1 
I think this is a good suggestion.

JesseStutler:
+1
good idea!

Monokaix:
Hi, team
I would recommend putting the scheduler in a separate repository named `descheduler` for a couple of reasons:

[...]

PR #3902, opened 2024-12-17
volcano support descheduler: add framework

Monokaix:
part of https://github.com/volcano-sh/volcano/issues/3889


pkg/descheduler/setupplugins.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
lowang-bh, line 2: Does this line need to be updated?

Monokaix: This is copied from k8s descheduler, it's better to keep original format?

pkg/descheduler/framework/profile/profile_test.go
JesseStutler, line 1: This file has no copyright header

Monokaix: done

pkg/descheduler/apis/componentconfig/types.go
```diff
@@ -0,0 +1,76 @@
...
+	// Deprecated: Use clientConnection.kubeConfig instead.
+	KubeconfigFile string
+
+	// PolicyConfigFile is the filepath to the descheduler policy configuration.
+	PolicyConfigFile string
+
+	// Dry run
+	DryRun bool
+
+	// Node selectors
```
googs1025, line 54: Why is nodeselector defined as string? Can we use the fields defined in k8s directly? For example: `v1.NodeSelector`

pkg/descheduler/apis/componentconfig/types.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// MaxNoOfPodsToEvictPerNode restricts maximum of pods to be evicted per node.
+	MaxNoOfPodsToEvictPerNode int
+
+	// EvictLocalStoragePods allows pods using local storage to be evicted.
+	EvictLocalStoragePods bool
+
+	// IgnorePVCPods sets whether PVC pods should be allowed to be evicted
+	IgnorePVCPods bool
+
```
googs1025, line 65: If there are two fields with similar semantics, our best approach is to keep them consistent. For example: both start with Evict or both start with Igore

pkg/descheduler/descheduler.go
```diff
@@ -0,0 +1,413 @@
...
+	if err != nil {
+		klog.Warning("Warning: Convert Kubernetes server minor version to float fail")
+	}
+
+	if math.Abs(deschedulerMinorVersionFloat-kubernetesMinorVersionFloat) > 3 {
+		klog.Warningf("Warning: Descheduler minor version %v is not supported on your version of Kubernetes %v.%v. See compatibility docs for more info: https://github.com/kubernetes-sigs/descheduler#compatibility-matrix", deschedulerMinorVersion, serverVersion.Major, serverVersion.Minor)
+	}
+}
+
+func cachedClient(
```
googs1025, line 155: Why do we need a fakeclient to call here?

pkg/descheduler/descheduler.go
```diff
@@ -0,0 +1,413 @@
...
+	"sigs.k8s.io/descheduler/pkg/version"
+
+	"volcano.sh/volcano/cmd/descheduler/app/options"
+	frameworkprofile "volcano.sh/volcano/pkg/descheduler/framework/profile"
+)
+
+const podNameEnvKey string = "HOSTNAME"
+const podNamespaceEnvKey string = "POD_NAMESPACE"
+
+func Run(ctx context.Context, rs *options.DeschedulerServer) error {
```
googs1025, line 68: Can we separate the option validation logic into a method, such as validate or something else? This way, we don't have to write it in Run.

pkg/descheduler/setupplugins.go
lowang-bh: ok

k82cn:
Why introduce an new command? It's better to use action to avoid configuration confliction between scheduler and de-scheduler.

pkg/descheduler/setupplugins.go
kevin-wangzefeng: We need to keep it as is, since respecting the original copyright is required by Apached 2.0 license.

Monokaix:
> Why introduce an new command? It's better to use action to avoid configuration confliction between scheduler and de-scheduler.

That's a good question! I think there are two aspects that really matters:
1. Descheduling is a dangerous behavior and we can't sacrifice the SLO and business stability for resource utilization, so it's very important to consider the eviction condition like whether new pod can be scheduled before descheduling, and descheudler support a good mechanism to check this and it also supports a series of filter conditions like just limited nodes and namespace can do eviction, which is more responsive to users' real eviction scenarios and more versatile, while descheduling in volcano's action seems no such higher capabilities: )
2. Providing a separate descheduler component gives the user more options, e.g., if the user only wants to use real load-aware descheduling, he/she can deploy the descheduler component alone, uncoupled from the volcano scheduler itself.

pkg/descheduler/setupplugins.go
Monokaix: great!

pkg/descheduler/apis/componentconfig/types.go
Monokaix: This is inherited from upstream descheduler and just kept from the original forma: )

pkg/descheduler/apis/componentconfig/types.go
Monokaix: This is also inherited from upstream descheduler  it's better to keep same fileds.

pkg/descheduler/descheduler.go
Monokaix: It's used in `DryRun` mode: )

pkg/descheduler/descheduler.go
Monokaix: Same above, it's inherited from upstream descheduler.

Monokaix:
We will Introduce new cmd in a new descheduler 
repo, see https://github.com/volcano-sh/descheduler/pull/1

PR #3907, opened 2024-12-20
pod-update-performance

molei20021:
If there are thousands of pending pods in cluster, pod status update may become performance bottleneck of the whole system. I add PodStatusRateLimit config including minIntervalSec, minPodNum, enable. When the pod num of cluster is greater than minPodNum(default 10000), the times of pod status update will be limited by minIntervalSec(default 60s). If the pod unscheduleable status is updated minIntervalSec ago, we will ignore status update this time until after minIntervalSec.

### DIFF
--- a/pkg/scheduler/api/pod_info.go
+++ b/pkg/scheduler/api/pod_info.go
@@ -172,3 +172,9 @@ func GetPodResourceWithoutInitContainers(pod *v1.Pod) *Resource {
 
 	return result
 }
+
+type PodStatusRateLimit struct {
+	Enable         bool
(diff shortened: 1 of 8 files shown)


lowang-bh:
Is there a backgroup for this PR? Actually, pod will not be updated if there are no changes.

molei20021:
> Is there a backgroup for this PR? Actually, pod will not be updated if there are no changes.

Our 

> minIntervalSec

For example, one of the pod unscheduable reason is like "Pod %s/%s can possibly be assigned to %s" this may often change.

PR #3913, opened 2024-12-23
fix: hierarchical queue webhook validation use listing podgroups instead

### LINKED ISSUE #3911: queue.status.Running  do not increased  when create vcjob and pod is running successed.
### Description

1. create a vcjob with queue=test-queue.  
2. when vcjob 's pod is running ,   queue.status.Running  is null . 
﻿
3. further, if queue.status.Running is null , that's mean  you can continues create a child queue on  test-queue. But this case should be rejected . 

1. create vcjob  with queue=test-queue.  wait for pod is running .
2.  kubectl get queue test-queue -oyaml .    The queue.status.Running is null. 
3.  create queue with parent queue=test-queue.  You can successed create it.

when a queue has some workload ,   the creation of child queue should be reject. 

master

ignore update when state does not change .   even including status.Running field ?

https://github.com/volcano-sh/volcano/blob/0966fd59a9e0c0cfa7f6a19d41b3263f932ba4e0/pkg/controllers/queue/queue_controller_action.go#L111

JesseStutler:
fix #3911

### validation
Steps:
1. Create a queue `parent-queue`
2. Create a running job in `parent-queue`
3. Create a queue `child-queue` as the child of `parent-queue`
4. Check whether the create request in step-3 will be rejected by the webhook

The screenshot shows that create a child queue under parent queue which contains podgroups will be rejected:
image

### DIFF
--- a/pkg/webhooks/admission/queues/validate/validate_queue.go
+++ b/pkg/webhooks/admission/queues/validate/validate_queue.go
@@ -23,6 +23,7 @@ import (
 
 	admissionv1 "k8s.io/api/admission/v1"
 	whv1 "k8s.io/api/admissionregistration/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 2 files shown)


hwdef:
Please make the CI happy :)

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -255,10 +257,21 @@ func validateHierarchicalQueue(queue *schedulingv1beta1.Queue) error {
 		return fmt.Errorf("failed to get parent queue of queue %s: %v", queue.Name, err)
 	}
 
-	if parentQueue.Status.Pending+parentQueue.Status.Running+parentQueue.Status.Unknown+parentQueue.Status.Inqueue > 0 {
-		return fmt.Errorf("queue %s cannot be the parent queue of queue %s because it has PodGroups (pending: %d, running: %d, unknown: %d, inqueue: %d)",
-			parentQueue.Name, queue.Name, parentQueue.Status.Pending,
-			parentQueue.Status.Running, parentQueue.Status.Unknown, parentQueue.Status.Inqueue)
+	pgList, err := config.VolcanoClient.SchedulingV1beta1().PodGroups("").List(context.Background(), metav1.ListOptions{})
```
hwdef, line 260: How about use `informer` list

JesseStutler: Informer will list/watch all podgroups, which increases webhook memory consumption. I don’t really want to sacrifice memory for a verification.

hwdef: But not using the informer will put more pressure on the apiserver.

JesseStutler: OK, I'll solve it

JesseStutler: cc @Monokaix

Monokaix: We can check queue.status.allocated.

PR #3917, opened 2024-12-24
fix: if the capability cpu or memory is not specified in the hierarchical queue, it will be set to the corresponding value of the parent queue

### LINKED ISSUE #3914: Setting queue.spec.capacity.cpu field empty in child queue will result in an error: Failed to check queue's hierarchical structure
### Description

1. build hierarchical queue like this:   root's child queue is root-sub1.    root-sub1's child queue is root-sub1-sub1 
``` 
root 
 |
root-sub1 ( capacity.cpu = 100,  mem=10Gi)
 |
root-sub1-sub1 ( capacity.mem = 5Gi)
``` 

2. create vcjob with queue = root-sub1-sub1 . 

we can see error log in vc-scheduler : Failed to check queue's hierarchical structure, error.....
image

1. create an queeu with queue.spec.capacity.cpu field empty .
2. create a vcjob . 

Because of there is a default values when cpu/mem is null .   Cpu would be setting to MaxFloat64.
https://github.com/volcano-sh/volcano/blob/0966fd59a9e0c0cfa7f6a19d41b3263f932ba4e0/pkg/scheduler/plugins/capacity/capacity.go#L304

So,  should we reject creation a queue in admission-webhook when queue.spec.capacity.cpu/mem is null ? 

master

_No response_

JesseStutler:
fix #3914 

### validation
1. Set the log level of volcano-scheduler to 4
2. Create a parent-queue, with capability CPU 100, memory 10Gi
3. Create a child-queue, with capability memory 5Gi
4. Check if there are any logs related with `Failed to check queue's hierarchical structure`

Results:
Hierarchical structure can be successfully checked in capacity plugin, without `Failed to check queue's hierarchical structure` log:
image

### DIFF
--- a/pkg/scheduler/api/test_utils.go
+++ b/pkg/scheduler/api/test_utils.go
@@ -94,10 +94,16 @@ type ScalarResource struct {
 
 // BuildResourceList builds resource list object
 func BuildResourceList(cpu string, memory string, scalarResources ...ScalarResource) v1.ResourceList {
-	resourceList := v1.ResourceList{
-		v1.ResourceCPU:    resource.MustParse(cpu),
-		v1.ResourceMemory: resource.MustParse(memory),
+	resourceList := v1.ResourceList{}
+
+	if len(cpu) > 0 {
+		resourceList[v1.ResourceCPU] = resource.MustParse(cpu)
 	}
+
+	if len(memory) > 0 {
+		resourceList[v1.ResourceMemory] = resource.MustParse(memory)
+	}
+
 	for _, scalar := range scalarResources {
 		resourceList[v1.ResourceName(scalar.Name)] = resource.MustParse(scalar.Value)
 	}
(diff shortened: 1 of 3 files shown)


pkg/scheduler/api/test_utils.go
```diff
@@ -94,10 +94,16 @@ type ScalarResource struct {
...
+	resourceList := v1.ResourceList{}
+
+	if len(cpu) > 0 {
+		resourceList[v1.ResourceCPU] = resource.MustParse(cpu)
 	}
+
+	if len(memory) > 0 {
+		resourceList[v1.ResourceMemory] = resource.MustParse(memory)
+	}
+
```
lowang-bh, line 106: does scalarResources has same issue?

JesseStutler: The modification here is just for the convenience of using in UT without passing cpu or memory. You can see that scalarResources is a slice. If it is an empty slice or nil, resourceList will not be assigned.

hwdef (commented):
/lgtm

JesseStutler:
cc @Monokaix

hwdef:
Why is this pr blocked? @Monokaix @JesseStutler

JesseStutler:
done

PR #3919, opened 2024-12-24
chart: add scheduler and controller-manager prometheus scrape annotations

7sunarni:
From the [values.yaml](https://github.com/volcano-sh/volcano/blob/master/installer/helm/chart/volcano/values.yaml#L20), the both of controller-manager and scheduler components are enabled metrics by default.
I think we should add some default Prometheus annotations, so the metrics data can be scraped when we install by helm.

### DIFF
--- a/hack/run-e2e-kind.sh
+++ b/hack/run-e2e-kind.sh
@@ -62,6 +62,7 @@ basic:
   crd_version: ${crd_version}
 
 custom:
+  scheduler_log_level: 5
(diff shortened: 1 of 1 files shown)


7sunarni:
BTW, I found that the chart forget to clear the admission secret after uninstall it. I add a new commit to fix this.

7sunarni:
@dongjiang1989 Do you have time to take a look at here?

Issue #3881, opened 2024-12-13
Volcano scheduler: Populate hyperNode tree in cache


Monokaix:
Part implementation of https://github.com/volcano-sh/volcano/pull/3850
Should add informer and event handler of hyperNode in scheduler cache and build a hyperNode tree from hyperNodes, which can be used when scheduling.

JesseStutler:
Do we need to label `kind/feature` for all related issues?

Monokaix:
> Do we need to label `kind/feature` for all related issues?

done.

PR #3922, opened 2024-12-24
Network toplogy aware scheduling support in Volcano scheduler: Populate hyperNode tree in cache #3881

penggu:
add event handler for adding/deleting/updating HyperNode in scheduler cache.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -57,11 +58,13 @@ import (
 	"volcano.sh/apis/pkg/apis/scheduling"
 	schedulingscheme "volcano.sh/apis/pkg/apis/scheduling/scheme"
 	vcv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	networktopov1alpha1 "volcano.sh/apis/pkg/apis/topology/v1alpha1"
 	vcclient "volcano.sh/apis/pkg/client/clientset/versioned"
 	"volcano.sh/apis/pkg/client/clientset/versioned/scheme"
 	vcinformer "volcano.sh/apis/pkg/client/informers/externalversions"
 	cpuinformerv1 "volcano.sh/apis/pkg/client/informers/externalversions/nodeinfo/v1alpha1"
 	vcinformerv1 "volcano.sh/apis/pkg/client/informers/externalversions/scheduling/v1beta1"
+	topologyinformerv1alpha1 "volcano.sh/apis/pkg/client/informers/externalversions/topology/v1alpha1"
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/features"
 	schedulingapi "volcano.sh/volcano/pkg/scheduler/api"
@@ -109,6 +112,7 @@ type SchedulerCache struct {
 	nodeInformer               infov1.NodeInformer
 	podGroupInformerV1beta1    vcinformerv1.PodGroupInformer
 	queueInformerV1beta1       vcinformerv1.QueueInformer
+	hypernodeInformerv1alpha1  topologyinformerv1alpha1.HyperNodeInformer
 	pvInformer                 infov1.PersistentVolumeInformer
 	pvcInformer                infov1.PersistentVolumeClaimInformer
 	scInformer                 storagev1.StorageClassInformer
@@ -127,16 +131,17 @@ type SchedulerCache struct {
 
 	Recorder record.EventRecorder
 
-	Jobs                 map[schedulingapi.JobID]*schedulingapi.JobInfo
-	Nodes                map[string]*schedulingapi.NodeInfo
-	Queues               map[schedulingapi.QueueID]*schedulingapi.QueueInfo
-	PriorityClasses      map[string]*schedulingv1.PriorityClass
-	NodeList             []string
-	defaultPriorityClass *schedulingv1.PriorityClass
-	defaultPriority      int32
-	CSINodesStatus       map[string]*schedulingapi.CSINodeStatusInfo
-	HyperNodesListByTier map[int][]string
-	HyperNodes           map[string]sets.Set[string]
+	Jobs                   map[schedulingapi.JobID]*schedulingapi.JobInfo
+	Nodes                  map[string]*schedulingapi.NodeInfo
+	Queues                 map[schedulingapi.QueueID]*schedulingapi.QueueInfo
+	PriorityClasses        map[string]*schedulingv1.PriorityClass
+	NodeList               []string
+	defaultPriorityClass   *schedulingv1.PriorityClass
+	defaultPriority        int32
+	CSINodesStatus         map[string]*schedulingapi.CSINodeStatusInfo
+	HyperNodeListByTier    map[int]sets.Set[string]    // Maps tier to a set of hypernode names
+	HyperNodes             map[string]sets.Set[string] // Maps hypernode name to a set of node names
+	hypernodeForestbuilder *HyperNodeForestBuilder
 
 	NamespaceCollection map[string]*schedulingapi.NamespaceCollection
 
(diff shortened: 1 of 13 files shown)


pkg/scheduler/cache/event_handlers.go
```diff
@@ -1261,3 +1264,96 @@ func (sc *SchedulerCache) setCSIResourceOnNode(csiNode *sv1.CSINode, node *v1.No
...
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
+
+	sc.deleteHyperNode(hyperNode)
+	klog.V(4).Infof("Deleted HyperNode(%s) from cache", hyperNode.Name)
+}
+
+// UpdateHyperNodeV1alpha1 update hypernode in scheduler cache
+func (sc *SchedulerCache) UpdateHyperNodeV1alpha1(oldObj, newObj interface{}) {
+	sc.DeleteHyperNodeV1alpha1(oldObj)
```
JesseStutler, line 1356: Can we directly implement the logic of update member and update hypernode here?

JesseStutler:
Please rebase onto the new codes: https://github.com/volcano-sh/volcano/pull/3874, and squash your commit into one, so I can review clearly

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,236 @@ type imageState struct {
...
+// nodes, or link existing nodes - which could demote nodes from roots to child.
+// Because of regex matching, this can trigger ripple effects in the forest.
+func (sc *SchedulerCache) addHyperNode(hn *networktopov1alpha1.HyperNode) {
+	hnfb := sc.hypernodeForestbuilder
+	hntn := NewHyperNodeTreeNode(hn.Name)
+	hntn.HyperNodeRef = hn
+	hnfb.nodes[hn.Name] = hntn
+
+	tierStr := hn.Spec.Tier
+	tier, err := strconv.Atoi(tierStr)
```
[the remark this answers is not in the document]
penggu: I think Tier should be an int. If a "tier name" is needed, then a mapping from a "tier name" string to its "tier number" int is needed.

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,236 @@ type imageState struct {
...
+	// Iterate over members and add them to the forest
+	for _, member := range hn.Spec.Members {
+		switch member.Type {
+		case networktopov1alpha1.MemberTypeNode:
+			switch member.Selector.Type {
+			case networktopov1alpha1.ExactMatchMemberSelectorType:
+				nodeName := member.Selector.ExactMatch.Name
+				hntn.NodeMatchers.Insert(NewExactMatcher(nodeName))
+				workerNode := sc.Nodes[nodeName]
+				if workerNode == nil {
```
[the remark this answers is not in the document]
penggu: Thanks for bring up this scenario. I am aware of it. My hope is that the event handler for Node need to update the HyperNode (I did not implement that logic in the current PR. I was planning to add it later). 

Your proposed solution (using workqueue) should work for Exact match. I wonder how does it work for Regex match. In case of Regex match, when will we know for sure that no future child node will appear -- hence we stop throwing it back to the workqueue?

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,251 @@ type imageState struct {
...
+// NewHyperNodeForestBuilder initializes and returns a ForestBuilder
+func NewHyperNodeForestBuilder() *HyperNodeForestBuilder {
+	return &HyperNodeForestBuilder{
+		nodes: make(map[string]*HyperNodeTreeNode),
+		roots: sets.New[string](),
+	}
+}
+
+// insertHyperNodeMember inserts a member into a hypernode and updates the forest
+func insertHyperNodeMember(hntn *HyperNodeTreeNode, hnfb *HyperNodeForestBuilder, memberName string, memberNode *HyperNodeTreeNode, sc *SchedulerCache) {
```
Monokaix, line 268: There is no distinction between the nodes in the leaf nodes and the hyperNodes of non-leaf nodes. What we expect is that the key stored in `HyperNodes` is the hyperNode, and the value is a list of all leaf nodes under the hyperNode.

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,251 @@ type imageState struct {
...
+
+	// Iterate over members and add them to the forest
+	for _, member := range hn.Spec.Members {
+		switch member.Type {
+		case networktopov1alpha1.MemberTypeNode:
+			switch member.Selector.Type {
+			case networktopov1alpha1.ExactMatchMemberSelectorType:
+				nodeName := member.Selector.ExactMatch.Name
+				hntn.NodeMatchers.Insert(NewExactMatcher(nodeName))
+				workerNode := sc.Nodes[nodeName]
```
Monokaix, line 306: `sc.Nodes` should be accessed with lock.

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,251 @@ type imageState struct {
...
+}
+
+// HyperNodeTreeNode reprensent a single hypernode in the tree
+type HyperNodeTreeNode struct {
+	Name           string
+	Tier           int
+	Parent         *HyperNodeTreeNode
+	Members        sets.Set[*HyperNodeTreeNode]
+	MemberMatchers sets.Set[*MemberMatcher]
+	Nodes          sets.Set[*schedulingapi.NodeInfo]
```
Monokaix, line 235: We just need nodeName instead of NodeInfo.

pkg/scheduler/cache/cache.go
```diff
@@ -127,14 +131,18 @@ type SchedulerCache struct {
...
+	Queues                 map[schedulingapi.QueueID]*schedulingapi.QueueInfo
+	PriorityClasses        map[string]*schedulingv1.PriorityClass
+	NodeList               []string
+	defaultPriorityClass   *schedulingv1.PriorityClass
+	defaultPriority        int32
+	CSINodesStatus         map[string]*schedulingapi.CSINodeStatusInfo
+	HyperNodeListByTier    map[int]sets.Set[string]    // Maps tier to a set of hypernode names
+	HyperNodes             map[string]sets.Set[string] // Maps hypernode name to a set of node names
+	HyperNodeTreeRoots     HyperNodeSet
+	hypernodeForestbuilder *HyperNodeForestBuilder
```
Monokaix, line 145: Should also be locked.

Monokaix:
We should also expose a method like LCA(Lowest Common Ancestor) for HyperNodeTreeRoots, which will be used by score logic.

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,251 @@ type imageState struct {
...
+	tier, err := strconv.Atoi(tierStr)
+	if err != nil {
+		klog.Errorf("Failed to convert tier %s to int: %v", tierStr, err)
+		return
+	}
+	hntn.Tier = tier
+	klog.V(4).Infof("Adding hypernode %s to the forest, in %d tier", hn.Name, tier)
+
+	// Iterate over members and add them to the forest
+	for _, member := range hn.Spec.Members {
```
Monokaix, line 299: This can be wrapped as a func, it' the logic to build a hyperNodeTreeNode, and the last thing to do is update the `sc.HyperNodeListByTier` and `sc.HyperNodes`. 
And can we just use one var to express the tree?

pkg/scheduler/cache/cache.go
Monokaix, line 235: And seems nowhere to assign leaf nodes to `sc.HyperNodes`.

pkg/scheduler/cache/cache.go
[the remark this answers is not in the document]
JesseStutler: I think a better solution is to update HyperNodeTree when `AddNode` and `DeleteNode`, so that it doesn't matter if Node is not created yet, both `ExactMatch` and `RegexMatch` can be included. Our original intention of designing RegexMatch is actually to serve Node.

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,251 @@ type imageState struct {
...
+			}
+		}
+	}
+
+	// Iterate over existing HyperNodes, if the name of the new HyperNode matches
+	// either the Exact or Regex matchers of a parent, add it as a child member
+	for _, parentNode := range sc.hypernodeForestbuilder.nodes {
+		for m := range parentNode.MemberMatchers {
+			if m.Exact != "" && m.Exact == hn.Name {
+				insertHyperNodeMember(parentNode, hnfb, hn.Name, hntn, sc)
```
JesseStutler, line 363: insertHyperNodeMember should be used as a method of *HyperNodeForestBuilder instead of passing it in through parameters.

JesseStutler: SchedulerCache should also be used as an attribute of HyperNodeForestBuilder. The function parameters must be consistent with the function name and responsibilities.

JesseStutler: Don't memerNode already contain a name attribute? No need to pass again

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,236 @@ type imageState struct {
...
+// insertHyperNodeMember inserts a member into a hypernode and updates the forest
+func insertHyperNodeMember(hntn *HyperNodeTreeNode, hnfb *HyperNodeForestBuilder, memberName string, memberNode *HyperNodeTreeNode, sc *SchedulerCache) {
+	hntn.Members.Insert(memberNode)
+	memberNode.Parent = hntn
+	if hnfb.roots.Has(memberName) {
+		hnfb.roots.Delete(memberName)
+	}
+	sc.HyperNodeListByTier[memberNode.Tier].Insert(memberName)
+
+	sc.HyperNodes[hntn.Name].Union(sc.HyperNodes[memberName])
```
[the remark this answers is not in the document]
JesseStutler: We should recursively update this data structure through layer-order traversal and update all Nodes under HyperNodes

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,236 @@ type imageState struct {
...
+		case networktopov1alpha1.MemberTypeHyperNode:
+			switch member.Selector.Type {
+			case networktopov1alpha1.ExactMatchMemberSelectorType:
+				memberName := member.Selector.ExactMatch.Name
+				hntn.MemberMatchers.Insert(NewExactMatcher(memberName))
+				member := sc.hypernodeForestbuilder.nodes[memberName]
+				if member != nil {
+					insertHyperNodeMember(hntn, hnfb, memberName, member, sc)
+					if hntn.Parent != nil {
+						sc.HyperNodes[hntn.Parent.Name].Union(sc.HyperNodes[memberName])
```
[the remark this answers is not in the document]
penggu: > Also has the same problem here, Union is cloned and the return value is not processed, and I didn't get the point here, what is the purpose here?

All members of a HyperNode should also be members of its parent

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,236 @@ type imageState struct {
...
+				}
+				hntn.MemberMatchers.Insert(matcher)
+				for memberName, memberNode := range sc.hypernodeForestbuilder.nodes {
+					if matcher.Matches(memberName) {
+						insertHyperNodeMember(hntn, hnfb, memberName, memberNode, sc)
+
+					}
+				}
+				if hntn.Parent != nil {
+					sc.HyperNodes[hntn.Parent.Name].Union(sc.HyperNodes[hntn.Name])
```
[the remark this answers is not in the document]
penggu: > Also has the same problem here

Same answer: All members of a HyperNode should also be members of its parent

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +184,236 @@ type imageState struct {
...
+	hntn.MemberMatchers.Clear()
+	hntn.Nodes.Clear()
+	hntn.NodeMatchers.Clear()
+
+	// Garbage collect the HyperNodeTreeNode
+	delete(hnfb.nodes, hn.Name)
+	if hnfb.roots.Has(hn.Name) {
+		hnfb.roots.Delete(hn.Name)
+	}
+
```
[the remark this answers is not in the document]
penggu: Addressed.

pkg/scheduler/cache/cache.go
```diff
@@ -127,14 +131,18 @@ type SchedulerCache struct {
...
+	Nodes                  map[string]*schedulingapi.NodeInfo
+	Queues                 map[schedulingapi.QueueID]*schedulingapi.QueueInfo
+	PriorityClasses        map[string]*schedulingv1.PriorityClass
+	NodeList               []string
+	defaultPriorityClass   *schedulingv1.PriorityClass
+	defaultPriority        int32
+	CSINodesStatus         map[string]*schedulingapi.CSINodeStatusInfo
+	HyperNodeListByTier    map[int]sets.Set[string]    // Maps tier to a set of hypernode names
+	HyperNodes             map[string]sets.Set[string] // Maps hypernode name to a set of node names
+	HyperNodeTreeRoots     HyperNodeSet
```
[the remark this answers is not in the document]
penggu: Good catch. It has been replaced by hypernodeForestbuilder.roots

pkg/scheduler/cache/event_handlers.go
penggu: Direct implementation involves 
1. Finding diff between the old and new CRD
2. Update the forest based on the updated fields, and all the parent/child relationship based on ExactMatch and RegexMatch needs to be updated. 

Because of the complicated logic involved in Step2, especially RegexMatch for both its children and its parents, that piece of code will be very complicated, and it will be a lot of duplicated logic from exiting AddHyperNodeV1alpha1() and DeleteHyperNodeV1alpha1(). 

i think the current implementation is easier to understand and to maintain. Please let me know if you think otherwise.

pkg/scheduler/cache/cache.go
```diff
@@ -795,6 +924,15 @@ func (sc *SchedulerCache) addEventHandler() {
 			DeleteFunc: sc.DeleteNumaInfoV1alpha1,
 		})
 	}
+
+	// TODO (penggu): create informat for HyperNode(v1alpha1) information
```
Monokaix, line 928: TODO still need?

pkg/scheduler/cache/event_handlers.go
JesseStutler: I think it's great

pkg/scheduler/cache/cache.go
```diff
@@ -176,6 +183,123 @@ type imageState struct {
...
+	return &MemberMatcher{Regex: regex}, nil
+}
+
+// HyperNodeTreeNode reprensent a single hypernode in the tree
+type HyperNodeTreeNode struct {
+	Name           string
+	Tier           int
+	Parent         *HyperNodeTreeNode
+	Members        sets.Set[*HyperNodeTreeNode]
+	MemberMatchers sets.Set[*MemberMatcher]
```
JesseStutler, line 233: Can we group together Members/NodeMatchers and MemberMatchers/NodeMatchers? Too many unused fields

Monokaix:
This pr has some code conflict to resolve and update, and https://github.com/volcano-sh/volcano/pull/3964 will work based on this: )

penggu:
This feature has been merged via PR #3964. So it is no longer needed. OK to close this PR now.

Issue #3923, opened 2024-12-25
vc-scheduler NPE when GPU-node not reporting devices properly


archlitchi:
### Description

This issue happens when activating volcano-vgpu, it happens when GPU node not report devices to node annotations properly.

### Steps to reproduce the issue

1. install volcano
2. enable volcano-vgpu
3. do not install volcano-vgpu-device-plugin

### Describe the results you received and expected

NPE happens for vc-scheduler, related logs:

```bash
E1225 02:12:26.904890     112 node_info.go:389] "Idle resources turn into negative after allocated" nodeName="vm-node245-vgpu" task="rise-vast-system/yunji-deployment-1-5878fc9b8b-fbvm8" resources=["nvidia.com/gpu"] idle="cpu 39225.00, memory 192033869059.00, ephemeral-storage 1097689833472000.00, pods 105.00, nvidia.com/gpu -1000.00, hugepages-1Gi 0.00, hugepages-2Mi 0.00, volcano.sh/vgpu-number 20000.00" req="cpu 0.00, memory 0.00, nvidia.com/gpu 1000.00, pods 1.00"
E1225 02:12:26.906417     112 node_info.go:298] "Node out of sync" name="vm-node245-vgpu" resources=["nvidia.com/gpu"]
E1225 02:12:26.926206     112 node_info.go:389] "Idle resources turn into negative after allocated" nodeName="vm-node245-vgpu" task="rise-vast-system/yunji-deployment-1-5878fc9b8b-fbvm8" resources=["nvidia.com/gpu"] idle="cpu 39350.00, memory 192139269379.00, nvidia.com/gpu -1000.00, hugepages-1Gi 0.00, volcano.sh/vgpu-number 20000.00, ephemeral-storage 1097689833472000.00, pods 107.00, hugepages-2Mi 0.00" req="cpu 0.00, memory 0.00, nvidia.com/gpu 1000.00, pods 1.00"
W1225 02:12:26.926590     112 node_info.go:336] received argument of nil node, no need to set other resources for 
W1225 02:12:26.926703     112 node_info.go:231] the argument node is null.
W1225 02:12:26.973399     112 node_info.go:336] received argument of nil node, no need to set other resources for 
W1225 02:12:26.973484     112 node_info.go:231] the argument node is null.
W1225 02:12:26.974207     112 node_info.go:336] received argument of nil node, no need to set other resources for 
W1225 02:12:26.974251     112 node_info.go:231] the argument node is null.
[...]

PR #3924, opened 2024-12-25
fix NPU issue on vc-scheduler

archlitchi:
Related issue: https://github.com/volcano-sh/volcano/issues/3923


lowang-bh:
Can we have a review on all relative code anywhere may has a similar issue?

archlitchi:
> Can we have a review on all relative code anywhere may has a similar issue?

for now, only one here

Issue #3925, opened 2024-12-25
Support for custom vGPU resource names


cgapu:
### What is the problem you're trying to solve

I am using Volcano and the Project-HAMi/volcano-vgpu-device-plugin plugin for GPU resource virtualization. This plugin supports modifying vGPU-related resources, but in Volcano, these resources are hardcoded. The relevant resource names are:
 - volcano.sh/vgpu-number
 - volcano.sh/vgpu-memory
 - volcano.sh/vgpu-cores"

### Describe the solution you'd like

 I hope to expose them as startup parameters so that users can customize them.

Monokaix:
@archlitchi Please check this: )

archlitchi:
yes, i plan to use a CM to store corresponding parameters, we'll discuss this at weekly meeting

Issue #3473, opened 2024-05-16
We'd better not depend on a kubeconfig file


lowang-bh:
https://github.com/VasuDevrani/volcano/actions/runs/9105698087/job/25031701446?pr=1
```
E0516 02:53:19.986663   39759 utils.go:43] init kubeclient in 4pdvgpu failed: stat /home/runner/.kube/config: no such file or directory
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16d78bd]

goroutine 1 [running]:
volcano.sh/apis/pkg/client/clientset/versioned.NewForConfig(0x0?)
	/home/runner/go/pkg/mod/volcano.sh/apis@v1.9.0/pkg/client/clientset/versioned/clientset.go:92 +0x1d
volcano.sh/apis/pkg/client/clientset/versioned.NewForConfigOrDie(...)
	/home/runner/go/pkg/mod/volcano.sh/apis@v1.9.0/pkg/client/clientset/versioned/clientset.go:153
volcano.sh/volcano/test/e2e/jobp.TestMain(0x1c51e00?)
	/home/runner/work/volcano/volcano/test/e2e/jobp/main_test.go:35 +0x67
main.main()
	_testmain.go:83 +0x1e6
FAIL	volcano.sh/volcano/test/e2e/jobp	0.023s
E0516 02:53:19.986155   39760 utils.go:43] init kubeclient in 4pdvgpu failed: stat /home/runner/.kube/config: no such file or directory
panic: runtime error: invalid memory address or nil pointer dereference
```

code at 
https://github.com/volcano-sh/volcano/blob/63bf2715100b1b62405e460df828df09eec9f4fe/pkg/scheduler/api/devices/nvidia/vgpu/utils.go#L39-L65
#### What would you like to be added:

The scheduler has its own service account. We'd better not to depend on a config file.

#### Why is this needed:

more compatible and portability

lowang-bh:
@Monokaix @archlitchi

Monokaix:
When will this ut failed?

archlitchi:
Fixed in this PR: https://github.com/volcano-sh/volcano/pull/3953

lowang-bh:
has been solved by https://github.com/volcano-sh/volcano/pull/3953

PR #3926, opened 2024-12-25
feat: expose vGPU resource names as arguments

cgapu:
this allows users to customize vGPU resources (vgpu-number, vgpu-memory, vgpu-cores) via startup arguments


cgapu:
[Related issue](https://github.com/volcano-sh/volcano/issues/3925)

lowang-bh:
image

we‘d better remove the kubeconf file dependencs. It occurs sevral times. cc @archlitchi @Monokaix 
relative issue at https://github.com/volcano-sh/volcano/issues/3473

pkg/scheduler/api/devices/nvidia/vgpu/type.go
```diff
@@ -19,6 +19,10 @@ package vgpu
 var VGPUEnable bool
 var NodeLockEnable bool
 
+var VGPUNumber string
```
Monokaix, line 22: Can we make them as a local var?

Monokaix:
Pleae also update the doc to let users know the feature.

pkg/scheduler/api/devices/nvidia/vgpu/type.go
cgapu: These variables are used in many places, and it would be difficult to change them to local variables. If we were to make the change, it would require a lot of modifications. Do you have any suggestions?

cgapu:
> Pleae also update the doc to let users know the feature.

updated the doc

archlitchi:
@cgapu i noticed that expose vGPU resourceName here may not work, because scheduler-configs is not exposed to 'volcano-vgpu-device-plugin'. Which may lead to resource definition inconsistency between 'device-plugin' and 'vc-scheduler', one solution is to add these configurations to a 'device-configMap', which can be accessed by both 'vc-scheduler' and 'volcano-vgpu-device-plugin'. related PR here: https://github.com/volcano-sh/volcano/pull/3953

PR #3933, opened 2024-12-27
chore: use Infof for slice taskinfo instead of InfoS log less details

### LINKED ISSUE #3687: victims log print a lot of details

victim pods has print the detail yaml, which is not necessary.

image

just need basic task info like preemtor
image

### Additional context

_No response_

zedongh:
`klog.InfoS` log slice object will use json serializer, use `klog.Infof` instead

/close #3687

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -138,7 +138,7 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 			allocated.Sub(reclaimee.Resreq)
 			victims = append(victims, reclaimee)
 		}
-		klog.V(4).InfoS("Victims from capacity plugin", "victims", victims, "reclaimer", reclaimer)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -138,7 +138,7 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 			allocated.Sub(reclaimee.Resreq)
 			victims = append(victims, reclaimee)
 		}
-		klog.V(4).InfoS("Victims from capacity plugin", "victims", victims, "reclaimer", reclaimer)
+		klog.V(4).Infof("Victims from capacity plugin, victims=%+v reclaimer=%s", victims, reclaimer)
```
Monokaix, line 141: >`victims=%+v` 

Will this just print according to taskInfo's String method of a slice?

zedongh: > > `victims=%+v`
> 
> Will this just print according to taskInfo's String method of a slice?

Yes, "+v" on slice will call each element String meothd if provide

lowang-bh (commented):
/lgtm

Issue #3932, opened 2024-12-27
capacity plugin only check leaf queue's real-capability.


bogo-y:
### Description

Now the capacity plugin will add a allocatableFn when the schedule session opening, and this func check if the leaf queue's real-capability is enough to carry the allocated resources and candidate's need. We can find this func [here](https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/capacity/capacity.go#L180).
Let us consider the hierarchical queues below:

![image](https://github.com/user-attachments/assets/83b61596-d851-4413-932d-2b33c3728405)

We take memory as an example. There are 6 queues and all of their guarantee is 0. Now the pg3 could pass the capacity plugin's allocatableFn, which will lead to queue1 and root allocate 6Gi that exceed their real-capability. So I think maybe we need check whether all the parents' real-capability are enough that can be implemented through the following code.
```
        queueAllocatable := func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
                attr := cp.queueOpts[queue.UID]
                futureUsed := attr.allocated.Clone().Add(candidate.Resreq)
                allocatable := futureUsed.LessEqualWithDimension(attr.realCapability, candidate.Resreq)
                if !allocatable {
                        klog.V(3).Infof("Queue <%v>: realCapability <%v>, allocated <%v>; Candidate <%v>: resource request <%v>",
                                queue.Name, attr.realCapability, attr.allocated, candidate.Name, candidate.Resreq)
                }
                return allocatable
        }

        ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
                if !readyToSchedule {
                        klog.V(3).Infof("Capacity plugin failed to check queue's hierarchical structure!")
                        return false
                }
                if hierarchyEnabled {
                        if !cp.isLeafQueue(queue.UID) {
                                return false
                        }
                        allocatable := queueAllocatable(queue, candidate)
                        for _, parentID := range cp.queueOpts[queue.UID].parents {
                                if !queueAllocatable(ssn.Queues[parentID], candidate) && allocatable {
                                        allocatable = false
                                }
                        }
                        return allocatable
                } else {
                        return queueAllocatable(queue, candidate)
                }
        })
```
[...]

PR #3940, opened 2024-12-30
vc-scheduler: fix omitting check ancestor-queues' real-capability in capacity plugin. 

bogo-y:
fix https://github.com/volcano-sh/volcano/issues/3932

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -165,25 +165,38 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		return !overused
 	})
 
-	ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
-		if !readyToSchedule {
-			klog.V(3).Infof("Capacity plugin failed to check queue's hierarchical structure!")
-			return false
-		}
-
-		if hierarchyEnabled && !cp.isLeafQueue(queue.UID) {
-			return false
-		}
+	queueAllocatable := func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
 		attr := cp.queueOpts[queue.UID]
-
 		futureUsed := attr.allocated.Clone().Add(candidate.Resreq)
 		allocatable := futureUsed.LessEqualWithDimension(attr.realCapability, candidate.Resreq)
 		if !allocatable {
 			klog.V(3).Infof("Queue <%v>: realCapability <%v>, allocated <%v>; Candidate <%v>: resource request <%v>",
 				queue.Name, attr.realCapability, attr.allocated, candidate.Name, candidate.Resreq)
 		}
-
 		return allocatable
+	}
+
+	ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
+		if !readyToSchedule {
+			klog.V(3).Infof("Capacity plugin failed to check queue's hierarchical structure!")
+			return false
+		}
+		if hierarchyEnabled && !cp.isLeafQueue(queue.UID) {
+			klog.V(3).Infof("Queue <%s> is not a leaf queue, can not allocate task <%s>.", queue.Name, candidate.Name)
+			return false
+		}
+		list := append(cp.queueOpts[queue.UID].parents, queue.UID)
+		for i := len(list) - 1; i >= 0; i-- {
+			if !queueAllocatable(ssn.Queues[list[i]], candidate) {
+				if klog.V(5).Enabled() {
+					for i--; i >= 0; i-- {
+						queueAllocatable(ssn.Queues[list[i]], candidate)
+					}
+				}
+				return false
+			}
+		}
+		return true
 	})
 
 	ssn.AddJobEnqueueableFn(cp.Name(), func(obj interface{}) int {
--- a/pkg/scheduler/plugins/capacity/capacity_test.go
+++ b/pkg/scheduler/plugins/capacity/capacity_test.go
@@ -371,6 +371,7 @@ func Test_capacityPlugin_OnSessionOpenWithHierarchy(t *testing.T) {
 	queue4 := buildQueueWithParents("q4", "root", api.BuildResourceList("2", "2Gi", []api.ScalarResource{{Name: "pods", Value: "1"}}...), api.BuildResourceList("4", "4Gi", []api.ScalarResource{{Name: "pods", Value: "4"}}...))
 	queue31 := buildQueueWithParents("q31", "q3", api.BuildResourceList("2", "2Gi", []api.ScalarResource{{Name: "pods", Value: "2"}}...), api.BuildResourceList("4", "4Gi", []api.ScalarResource{{Name: "pods", Value: "4"}}...))
 	queue32 := buildQueueWithParents("q32", "q3", api.BuildResourceList("2", "2Gi", []api.ScalarResource{{Name: "pods", Value: "2"}}...), api.BuildResourceList("4", "4Gi", []api.ScalarResource{{Name: "pods", Value: "4"}}...))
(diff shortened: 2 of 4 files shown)


bogo-y:
1. fix the AllocatableFn in capacity plugin 
2. add UT for new logic
3. disable Allocatable check in reclaim action
4. add PreemptiveFn in proportion plugin
/PTAL @Monokaix @JesseStutler

JesseStutler:
Great Job! Clean code and clean commits!
/lgtm
/area scheduler

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -165,25 +165,36 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			klog.V(3).Infof("Capacity plugin failed to check queue's hierarchical structure!")
+			return false
+		}
+		if hierarchyEnabled {
+			if !cp.isLeafQueue(queue.UID) {
+				return false
+			}
+			allocatable := queueAllocatable(queue, candidate)
+			for _, parentID := range cp.queueOpts[queue.UID].parents {
+				if !queueAllocatable(ssn.Queues[parentID], candidate) && allocatable {
```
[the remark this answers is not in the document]
bogo-y: > Why don't return false directly if allocatable is false?

Because I think we prefer to print all the logs that come from queues whose resources are insufficient.
Returning earlier can yield better performance if the leaf queue has too many parents, but our tree is shallow in most cases.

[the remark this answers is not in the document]
bogo-y: logs and performance are both improtant... 
And I am not sure which aspect we should value more between these two...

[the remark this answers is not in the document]
lowang-bh: we can expose metrics for observability.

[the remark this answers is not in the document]
bogo-y: > we can expose metrics for observability.

I think we are supposed to get failed scheduling information by events and logs.

You mean that we should add some metrics for failed scheduling information or just print the lowest queue's attribute then directly return?

hwdef (commented):
/lgtm

pkg/scheduler/plugins/capacity/capacity.go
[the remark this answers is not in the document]
bogo-y: @lowang-bh Done. Please take a look again. Thanks!

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -432,6 +438,16 @@ func Test_capacityPlugin_OnSessionOpenWithHierarchy(t *testing.T) {
 			ExpectEvicted:  []string{"ns1/p7"},
 			ExpectEvictNum: 1,
 		},
+		{
```
Monokaix, line 441: Should we also add a testcase that the ancestor queue's capability is used up and current queue can still allocate resources, allocate and reclaim actions behave differently here.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -165,25 +165,42 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		return !overused
 	})
 
-	ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
```
Monokaix, line 168: We'd better explain why the allocatable is no needed and replaced by preemptive, the overall explaination is that in allocate action we need check the ancestor queues' capability and in reclaim action we should just check current queue's capability, and reclaim happens when queue not allocatable so we just need focus on the reclaim here, so it's more descriptive to user preempt related semantics.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -165,25 +165,42 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
-
 		return allocatable
+	}
+
+	ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
+		if !readyToSchedule {
+			klog.V(3).Infof("Capacity plugin failed to check queue's hierarchical structure!")
+			return false
+		}
+		if hierarchyEnabled {
```
Monokaix, line 184: Can this be unified? When hierarchy is not enabled, the `list` just has one item.

pkg/scheduler/plugins/capacity/capacity_test.go
bogo-y: > Should we also add a testcase that the ancestor queue's capability is used up and current queue can still allocate resources, allocate and reclaim actions behave differently here.

Yes, we have case3 and case4, both under the condition that the ancestor queue's capability is used up and leaf queue can still allocate resources.

pkg/scheduler/plugins/capacity/capacity.go
bogo-y: > Can this be unified? When hierarchy is not enabled, the `list` just has one item.

That makes sense and I have modified it.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -165,25 +165,38 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
 			return false
 		}
-		attr := cp.queueOpts[queue.UID]
-
-		futureUsed := attr.allocated.Clone().Add(candidate.Resreq)
-		allocatable := futureUsed.LessEqualWithDimension(attr.realCapability, candidate.Resreq)
-		if !allocatable {
-			klog.V(3).Infof("Queue <%v>: realCapability <%v>, allocated <%v>; Candidate <%v>: resource request <%v>",
-				queue.Name, attr.realCapability, attr.allocated, candidate.Name, candidate.Resreq)
+		list := append(cp.queueOpts[queue.UID].parents, queue.UID)
```
Monokaix, line 188: Here we just need check its parent?

bogo-y: The `parents` is a slice in queueOpts, it contains the leaf queue's all ancestors.  I think we might change its name from `parents` to `ancestors`.
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/capacity/capacity.go#L637

Monokaix: it's reasonable: )

bogo-y: > it's reasonable: )

For simplicity, I will submit a pr after this one been merged.

JesseStutler:
Will this `for i--; i>=0; i--` not cause out of bounds?

JesseStutler:
The assembly of `list` also looks very strange. Why does it become `[queue.father, queue.grandfather, ..., queue.ancestor, queue]?` Wasn’t the previous writing method quite good? Why was it changed to this?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -165,25 +165,38 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		for i := len(list) - 1; i >= 0; i-- {
+			if !queueAllocatable(ssn.Queues[list[i]], candidate) {
+				if klog.V(5).Enabled() {
+					for i--; i >= 0; i-- {
```
JesseStutler, line 192: Will this for i--; i>=0; i-- not cause out of bounds?

pkg/scheduler/plugins/capacity/capacity.go
JesseStutler, line 188: The assembly of list also looks very strange. Why does it become [queue.father, queue.grandfather, ..., queue.ancestor, queue]? Wasn’t the previous writing method quite good? Why was it changed to this?

PR #3941, opened 2024-12-30
vs-scheduler: use max function to simplify code of capacity plugin

bogo-y:
Now we can use max function provided by Go.


hwdef (commented):
/approve

PR #3951, opened 2025-01-02
reclaim: When choosing a preemptor, choose a starving one rather than one with pending tasks.

### LINKED ISSUE #3869: reclaim action uses `HasPendingTasks` instead of `JobStarving`, which may cause the two jobs to reclaim with each other and lead to deadlock.

JesseStutler:
#### What type of PR is this?
/kind bug
/area scheduling

When two jobs have pending pods at the same time and resources can be reclaimed, using PendingPod to determine whether the job can enter reclaim will cause the two jobs to compete with each other, leading to a deadlock. We need to use `JobStarving` instead.

Fixes #3869 

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -69,7 +69,7 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			queues.Push(queue)
 		}
 
-		if job.HasPendingTasks() {
+		if ssn.JobStarving(job) {
 			if _, found := preemptorsMap[job.Queue]; !found {
 				preemptorsMap[job.Queue] = util.NewPriorityQueue(ssn.JobOrderFn)
 			}
@@ -108,7 +108,7 @@ func (ra *Action) Execute(ssn *framework.Session) {
 		}
 
 		// Found "high" priority task to reclaim others
-		if tasks, found := preemptorTasks[job.UID]; !found || tasks.Empty() {
+		if tasks, found := preemptorTasks[job.UID]; !found || tasks.Empty() || !ssn.JobStarving(job) {
 			continue
 		} else {
 			task = tasks.Pop().(*api.TaskInfo)
--- a/pkg/scheduler/actions/reclaim/reclaim_test.go
+++ b/pkg/scheduler/actions/reclaim/reclaim_test.go
@@ -44,8 +44,8 @@ func TestReclaim(t *testing.T) {
 				proportion.PluginName:  proportion.New,
 			},
 			PodGroups: []*schedulingv1beta1.PodGroup{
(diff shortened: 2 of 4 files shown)


pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -69,7 +69,7 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			queues.Push(queue)
 		}
 
-		if job.HasPendingTasks() {
+		if job.IsStarving() {
```
Monokaix, line 72: ssn.JobStarving?

JesseStutler: updated

Monokaix:
Should we also add JobStarving check in line 87 loop?

JesseStutler:
> Should we also add JobStarving check in line 87 loop?

Added

JesseStutler:
cc @hwdef @Monokaix

hwdef (commented):
/lgtm

PR #3954, opened 2025-01-03
vc-scheduler: optimize bachNodeOrderFn in numaaware plugin

bogo-y:

The  default value of float64 is 0, so we can directly return without creating a new map when the bachNodeOrderFn is inapplicable.


hwdef (commented):
/lgtm

PR #3958, opened 2025-01-07
vc-scheduler: rename parents to ancestors in hierarchical queueAttr.

bogo-y:
According to the discussion at https://github.com/volcano-sh/volcano/pull/3940#discussion_r1904094927, we'd better rename `parents` to `ancestors` in capacity plugin.

NONE


JesseStutler:
Great!

PR #3960, opened 2025-01-08
Fix session JobOrderFn to be predictable

raravena80:
#### What type of PR is this?

/kind bug
/area scheduling

- This fixes an issue we've been seeing where the JobOrder doesn't return the same consistent list order when compared by job UUID. We are ordering by Job Name now
- Additionally it makes sure that when two jobs have the same score the list is always in the consistent order. We are also ordering by Job Name.

Fixes #

Do we need to open an issue?

```release-note
- Fixes an issue we've been seeing where the JobOrder doesn't return the same consistent list order when compared by job UUID. We are ordering by Job Name now
- Makes sure that when two jobs have the same score the list is always in the consistent order. We are also ordering by Job Name.
```

### DIFF
--- a/pkg/scheduler/framework/session_plugins.go
+++ b/pkg/scheduler/framework/session_plugins.go
@@ -17,6 +17,7 @@ limitations under the License.
 package framework
 
 import (
+	"k8s.io/klog/v2"
 	k8sframework "k8s.io/kubernetes/pkg/scheduler/framework"
 
 	"volcano.sh/apis/pkg/apis/scheduling"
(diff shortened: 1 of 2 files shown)


JesseStutler:
`Additionally it makes sure that when two jobs have the same score the list is always in the consistent order. We are also ordering by Job Name.`

- two jobs --> two nodes
- Job Name --> Node Name

JesseStutler:
Yes, we'd better create an issue to let other contributors trace back why this was changed. 
Besides, what will happen to your cluster if we use UUID sorting? I hope to see some screenshots of the phenomenon in the issue:)

JesseStutler:
cc @lowang-bh @Monokaix @hwdef

raravena80:
> two jobs --> two nodes
> Job Name --> Node Name

Thanks!
Fixed these in the description.

Will get back to you on the exact issue, but I believe with UID we were getting different orders.

pkg/scheduler/framework/session_plugins.go
```diff
@@ -538,12 +539,15 @@ func (ssn *Session) JobOrderFn(l, r interface{}) bool {
...
 
-	// If no job order funcs, order job by CreationTimestamp first, then by UID.
+	// If no job order funcs, order job by CreationTimestamp first, then by Name.
 	lv := l.(*api.JobInfo)
 	rv := r.(*api.JobInfo)
 	if lv.CreationTimestamp.Equal(&rv.CreationTimestamp) {
-		return lv.UID < rv.UID
+		// Use the Name of the Job instead of UID to get deterministic order by Job name
+		klog.V(3).Infof("Creation timestamps are the same for job %v and %v: %v . using name to order", lv.Name, rv.Name, lv.CreationTimestamp)
+		return lv.Name < rv.Name
```
hwdef, line 548: If encounter a situation with the same name, that is, vcjobs with the same name that are not in the same namespace,  the JobOrder doesn't return the same consistent list order too?

hwdef, line 548: If we encounter a situation where vcjobs have the same name but are in different namespaces, will the JobOrder still not return a consistent list order?

raravena80: Good question.  So Job Names are namespaced?

raravena80:
Created ticket https://github.com/volcano-sh/volcano/issues/3962 with more details

JesseStutler:
Hi, I have some comments about your issue description, please check it

pkg/scheduler/framework/session_plugins.go
JesseStutler: @Monokaix

Monokaix: Yaeh jobs are namespaced.

PR #3894, opened 2024-12-16
Add network-topology-aware plugin and hyperNode score callback

ecosysbin:
What type of PR is this?
/kind feature
/area scheduling

What this PR does / why we need it:
Add network-topology-aware plugin and hyperNode score callback

Which issue(s) this PR fixes:
https://github.com/volcano-sh/volcano/issues/3885

PR #3964, opened 2025-01-11
Update: Populate hyperNode info

Monokaix:
#### What type of PR is this?
/kind feature
/area scheduling

Related issue: https://github.com/volcano-sh/volcano/issues/3881

Fixes #

This is based on https://github.com/volcano-sh/volcano/pull/3922 and do some clean up and add ut.

```release-note

```


Monokaix:
cc @william-wang @lowang-bh @hwdef @JesseStutler

pkg/controllers/job/job_controller_actions.go
```diff
@@ -723,6 +723,15 @@ func (cc *jobcontroller) createOrUpdatePodGroup(job *batch.Job) error {
 				PriorityClassName: job.Spec.PriorityClassName,
 			},
 		}
+		if job.Spec.NetworkTopology != nil {
+			nt := &scheduling.NetworkTopologySpec{
+				Mode: scheduling.NetworkTopologyMode(job.Spec.NetworkTopology.Mode),
```
hwdef, line 728: Do we need to check whether `Mode` is nil?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -204,6 +204,10 @@ func (alloc *Action) allocateResources(queues *util.PriorityQueue, jobsMap map[a
 		var stmt *framework.Statement
 		var tasksQueue *util.PriorityQueue
 		if hardMode {
+			if !alloc.session.HyperNodesReadyToSchedule {
+				klog.ErrorS(nil, "HyperNodes not completely populated and not ready to schedule, please check logs for more details", "job", job.UID)
```
hwdef, line 208: if error is nil, Is it better to use `Errorf` or `Errorln`? Because ErrorS will print 
```
"HyperNodes not completely populated and not ready to schedule, please check logs for more details" err=<nil> job="example-job-uid"
```

`err=<nil>` is strange

pkg/scheduler/util/test_utils.go
```diff
@@ -33,7 +33,6 @@ import (
 	"k8s.io/klog/v2"
 
 	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
-
```
hwdef, line 36: A line was deleted by mistake

pkg/scheduler/cache/cache.go
```diff
@@ -1335,8 +1338,8 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 
 	snapshot := &schedulingapi.ClusterInfo{
 		Nodes:                make(map[string]*schedulingapi.NodeInfo),
-		HyperNodesListByTier: make(map[int][]string),
-		HyperNodes:           make(map[string]sets.Set[string]),
+		HyperNodesListByTier: make(map[int]sets.Set[string]),
```
hwdef, line 1341:
```suggestion
		HyperNodesSetByTier: make(map[int]sets.Set[string]),
```

pkg/scheduler/cache/cache.go
```diff
@@ -1436,7 +1430,7 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 	wg.Wait()
 
 	klog.V(3).InfoS("SnapShot for scheduling", "jobNum", len(snapshot.Jobs), "QueueNum",
-		len(snapshot.Queues), "NodeNum", len(snapshot.Nodes), "tiers", snapshot.HyperNodesListByTier, "hyperNodeNum", hyperNodeLength)
+		len(snapshot.Queues), "NodeNum", len(snapshot.Nodes), "tiers", snapshot.HyperNodesListByTier, "realNodesSet", snapshot.RealNodesSet)
```
hwdef, line 1433: Is this information too detailed for the V3 log level? V4 may be more suitable.

pkg/scheduler/framework/session.go
```diff
@@ -78,9 +79,10 @@ type Session struct {
 	// HyperNodesListByTier contains a list of hyperNodes by tier from down to top, nodes under the same hyperNode
 	// have the same topology domain, e.g., nodes under the same switch or tor, jobs allocated in the same
 	// hyperNode can gain a better performance, the lower the tier of hyperNode, the better performance.
-	HyperNodesListByTier map[int][]string
+	HyperNodesListByTier map[int]sets.Set[string]
```
hwdef, line 82:
```suggestion
	HyperNodesSetByTier map[int]sets.Set[string]
```

hwdef:
Some minor issues.

pkg/scheduler/api/hyper_node_info.go
```diff
@@ -0,0 +1,463 @@
...
+			}
+
+			memberHn, ok := hni.hyperNodes[memberName]
+			if !ok {
+				klog.InfoS("HyperNode not exists in cache, maybe not created or not be watched", "name", memberName, "parent", hn.Name)
+				continue
+			}
+
+			// Recursively build cache for the member HyperNode.
+			if err := hni.BuildHyperNodeCache(memberHn, processed, ancestors, nodes); err != nil {
```
JesseStutler, line 201: Only the realNodesSet on the ancestors chain will be reset. Can the HyperNode that is not on the ancestor chain don't need to recursively enter BuildHyperNodeCache? Just check whether the realNodesSet of this member is not empty.

pkg/scheduler/api/hyper_node_info.go
```diff
@@ -0,0 +1,463 @@
...
+	nodes, err := hni.nodeLister.List(labels.Everything())
+	if err != nil {
+		klog.ErrorS(err, "Failed to list nodes", "hyperNodeName", name)
+		return err
+	}
+
+	for ancestor := range ancestors {
+		if hn, ok := hni.hyperNodes[ancestor]; ok {
+			if err := hni.BuildHyperNodeCache(hn, processed, ancestorsChain, nodes); err != nil {
+				hni.setReady(false)
```
JesseStutler, line 383: Would it be better if we moved `hni.setReady(false)` to the beginning of updateAncestors? If a job is submitted when rebuilding the HyperNodes tree, it may be scheduled to the wrong node at this time.

pkg/scheduler/api/hyper_node_info.go
```diff
@@ -0,0 +1,463 @@
...
+// exactMatchMember retrieves the member name if the selector is an exact match.
+func (hni *HyperNodesInfo) exactMatchMember(selector topologyv1alpha1.MemberSelector) string {
+	if selector.ExactMatch != nil {
+		return selector.ExactMatch.Name
+	}
+	return ""
+}
+
+func (hni *HyperNodesInfo) updateAncestors(name string) error {
+	ancestors := hni.GetAncestors(name)
```
JesseStutler, line 365: I encountered a bug here:
When I deleted a tier 1 hypernode and then added back later, it didn't set its parent yet (actually it does have its ancestors), but `GetAncestors` only return itself here, which will only reset realNodesSet of itself, not updating its ancestors' realNodesSet

pkg/scheduler/actions/allocate/allocate.go
Monokaix: `klog.ErrorS` will not print err if err == nil: )

pkg/controllers/job/job_controller_actions.go
Monokaix: Mode is a string type and no panic problem, also the API schema has the constrain that it must be an enum value `soft` or `hard`.

pkg/scheduler/actions/allocate/allocate.go
hwdef: yes, you are right. My mistake.

pkg/scheduler/api/hyper_node_info.go
```diff
@@ -0,0 +1,463 @@
...
+				ancestors.Insert(parent)
+				queue = append(queue, parent)
+			}
+		}
+	}
+	return ancestors
+}
+
+// GetDescendants returns all descendants of a given HyperNode.
+func (hni *HyperNodesInfo) GetDescendants(hyperNodeName string) sets.Set[string] {
```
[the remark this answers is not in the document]
ecosysbin: FindLCAHyperNode will need it in anothor PR：https://github.com/volcano-sh/volcano/pull/3894

pkg/scheduler/api/cluster_info.go
```diff
@@ -24,15 +24,16 @@ import (
...
-	HyperNodes           map[string]sets.Set[string]
-	Queues               map[QueueID]*QueueInfo
-	NamespaceInfo        map[NamespaceName]*NamespaceInfo
-	RevocableNodes       map[string]*NodeInfo
-	NodeList             []string
-	CSINodesStatus       map[string]*CSINodeStatusInfo
+	Jobs                      map[JobID]*JobInfo
+	Nodes                     map[string]*NodeInfo
+	HyperNodesListByTier      map[int]sets.Set[string]
+	RealNodesSet              map[string]sets.Set[string]
```
JesseStutler, line 30: Do we need to unify the variable called `HyperNodes` in Session struct to RealNodesSet?

pkg/scheduler/api/hyper_node_info.go
```diff
@@ -0,0 +1,463 @@
...
+
+func (hni *HyperNodesInfo) UpdateHyperNode(hn *topologyv1alpha1.HyperNode) error {
+	hni.updateHyperNode(hn)
+	return hni.updateAncestors(hn.Name)
+}
+
+func (hni *HyperNodesInfo) BuildHyperNodeCache(hn *HyperNodeInfo, processed sets.Set[string], ancestors sets.Set[string], nodes []*corev1.Node) error {
+	// Check for cyclic dependency.
+	if ancestors.Has(hn.Name) {
+		hni.setReady(false)
```
JesseStutler, line 157: Already set `hni.setReady` in updateAncestors, remove this line?

pkg/scheduler/api/hyper_node_info.go
Monokaix: That's a good catch! Will fix later.

pkg/scheduler/api/hyper_node_info.go
Monokaix: The informer processed ans snapshot logic are synchronized and there is already a lock.

pkg/scheduler/api/hyper_node_info.go
JesseStutler: Otherwise there will be some repetitive builds, like: 
image

pkg/scheduler/framework/session.go
Monokaix: done

pkg/scheduler/api/hyper_node_info.go
Monokaix: done

pkg/scheduler/api/cluster_info.go
Monokaix: done

pkg/scheduler/api/hyper_node_info.go
Monokaix: done

pkg/scheduler/cache/cache.go
Monokaix: done

PR #3973, opened 2025-01-17
Fix the restartPod action will cancel the restartJob delay action

### LINKED ISSUE #3812: Support More Actions For Volcano Job Failure Scenario

Currently, when scheduling a Volcano job, if you want to trigger rescheduling in error scenarios, the only way to retry is through the RestartJob Action. The impact range is sometimes too large.

It would be desirable to introduce new capabilities, such as individually restarting a specific Pod/Task to minimize the impact.

Furthermore, it could allow for a graded execution. For example, first attempt to restartPod/restartTask. If recovery is not achieved within a certain period, then try restartJob.

### Additional context

After implementing the above capability, we can use it as follows:
``` yaml
// other config
spec:
  policies:
  - event: PodFailed
    action: RestartPod
  - event: PodEvicted
    action: RestartJob
    timeout: 10m
// other config
```

When Pod A exits with an error, it triggers a PodFailed event, which in turn triggers a RestartPod action, so the Volcano controller will attempt to recreate Pod A. This recreation action will trigger a PodEvicted event, which in turn triggers a scheduled task to RestartJob after 10 minutes.

If Pod A returns to a running state, the delay RestartJob action is canceled, and everything returns to normal.
If Pod A still cannot reach the running state after 10 minutes, a broader reconstruction action, which is restartJob in this scene, will be attempted.

bibibox:
#### What type of PR is this?

/kind bug

The restartPod action should not cancel the restartJob delay action

Fixes #3812 

Add the action type check when try to cancel the delayed job actions

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/job/job_controller.go
+++ b/pkg/controllers/job/job_controller.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/informers"
@@ -71,6 +72,9 @@ type delayAction struct {
 
 	// The name of the pod
 	podName string
+
+	// The UID of the pod
+	podUID types.UID
 
 	// The event caused the action
 	event busv1alpha1.Event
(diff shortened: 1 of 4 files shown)


pkg/controllers/job/job_controller.go
```diff
@@ -513,16 +525,32 @@ func (cc *jobcontroller) handleJobError(queue workqueue.TypedRateLimitingInterfa
...
+//   - When a Pod failure triggers Job termination, need to cancel other Job delayed actions under this Job
+func (cc *jobcontroller) cleanupDelayActions(currentDelayAction *delayAction) {
 	cc.delayActionMapLock.Lock()
 	defer cc.delayActionMapLock.Unlock()
 
-	if m, exists := cc.delayActionMap[jobKey]; exists {
+	actionType := GetActionType(currentDelayAction.action)
+
+	if m, exists := cc.delayActionMap[currentDelayAction.jobKey]; exists {
 		for _, delayAct := range m {
```
JesseStutler, line 547: Don't we need to check if the PodName of currentDelayAction equals the PodName of delayAction?

PR #3975, opened 2025-01-18
[bugfix]fix leader-elect-resource-namespace flag not take effect

Monokaix:
#### What type of PR is this?
/kind bug

The `lock-object-namespace` flag has been deprecated and when users set new flag `--leader-elect-resource-namespace`, but the code has checked that if `lock-object-namespace` is not empty, it will still use old one, and `lock-object-namespace` has a default not empty so the new flag will never take effect.

Fixes #

```release-note

```


lowang-bh (approved):
/lgtm

PR #3987, opened 2025-01-24
Remove runningEstimate from CRD

abatilo:
This doesn't appear to be referenced anywhere in any of the code
anymore. I don't think there is a reason to keep this field in the
manifest spec anymore.


hwdef:
We may use this field in the future

PR #3994, opened 2025-02-04
Fix: fix an issue where the wrong action name could not be ignored

xieyanke:
#### What type of PR is this?
/kind bug
/area scheduling

fix an issue where the wrong action name could not be ignored

Fixes #

```release-note

```

### DIFF
--- a/pkg/scheduler/util.go
+++ b/pkg/scheduler/util.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"gopkg.in/yaml.v2"
+	"k8s.io/klog/v2"
 
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
@@ -80,7 +81,7 @@ func UnmarshalSchedulerConf(confStr string) ([]framework.Action, []conf.Tier, []
 		if action, found := framework.GetAction(strings.TrimSpace(actionName)); found {
 			actions = append(actions, action)
 		} else {
-			return nil, nil, nil, nil, fmt.Errorf("failed to find Action %s, ignore it", actionName)
+			klog.Errorf("Failed to find Action %s, ignore it", actionName)
 		}
 	}
 


bogo-y:
I think that should not be ignored.
For example, when a user makes a typo mistake such as typing "allocated" instead of "allocate", it is better to detect and resolve the issue immediately.

Monokaix:
> I think that should not be ignored. For example, when a user makes a typo mistake such as typing "allocated" instead of "allocate", it is better to detect and resolve the issue immediately.

Currently there is a log showing the message.

bogo-y:
> Currently there is a log showing the message.
There is too much messages when vc-scheduler just start... Maybe we need keep use `klog.Errorf` to make it highly visible?

xieyanke:
> > Currently there is a log showing the message.
> 
> There is too much messages when vc-scheduler just start... Maybe we need keep use `klog.Errorf` to make it highly visible?

Since actions are configurable, it’s reasonable to allow ignoring user misconfigurations while providing warning-level log messages.  
I think your concern can be addressed through another feature: "action classification". Some actions, such as “enqueue or allocate” are essential and require action validation, while others can be optionally ignored.

bogo-y:
> > > Currently there is a log showing the message.
> > 
> > 
> > There is too much messages when vc-scheduler just start... Maybe we need keep use `klog.Errorf` to make it highly visible?
> 
> Since actions are configurable, it’s reasonable to allow ignoring user misconfigurations while providing warning-level log messages. I think your concern can be addressed through another feature "action classification". Some actions, such as "enqueue or allocate" are essential and require action validation, while others can be optionally ignored.

There is no benefit to keep a wrong action name in configuration file, and if users discovered a mistake like this, they will fix it very easily.
just my personal opinion

Monokaix:
> > > Currently there is a log showing the message.
> > 
> > 
> > There is too much messages when vc-scheduler just start... Maybe we need keep use `klog.Errorf` to make it highly visible?
> 
> Since actions are configurable, it’s reasonable to allow ignoring user misconfigurations while providing warning-level log messages. I think your concern can be addressed through another feature "action classification". Some actions, such as "enqueue or allocate" are essential and require action validation, while others can be optionally ignored.

It's not a good way to classify different actions, all actions are important.

Monokaix:
> > > > Currently there is a log showing the message.
> > > 
> > > 
> > > There is too much messages when vc-scheduler just start... Maybe we need keep use `klog.Errorf` to make it highly visible?
> > 
> > 
> > Since actions are configurable, it’s reasonable to allow ignoring user misconfigurations while providing warning-level log messages. I think your concern can be addressed through another feature "action classification". Some actions, such as "enqueue or allocate" are essential and require action validation, while others can be optionally ignored.
> 
> There is no benefit to keep a wrong action name in configuration file, and if users discovered a mistake like this, they will fix it very easily. just my personal opinion

How about change Warning to Error?

bogo-y:
> How about change Warning to Error?

I think it's better than just use warning.

FYI: https://github.com/kubernetes/kubernetes/blob/release-1.32/pkg/scheduler/framework/runtime/framework.go#L658

xieyanke:
> > How about change Warning to Error?
> 
> I think it's better than just use warning.
> 
> FYI: https://github.com/kubernetes/kubernetes/blob/release-1.32/pkg/scheduler/framework/runtime/framework.go#L658

Okey agreed, I'll make the changes.

pkg/scheduler/util.go
```diff
@@ -80,7 +81,7 @@ func UnmarshalSchedulerConf(confStr string) ([]framework.Action, []conf.Tier, []
 		if action, found := framework.GetAction(strings.TrimSpace(actionName)); found {
 			actions = append(actions, action)
 		} else {
-			return nil, nil, nil, nil, fmt.Errorf("failed to find Action %s, ignore it", actionName)
+			klog.Errorf("failed to find Action %s, ignore it", actionName)
```
Monokaix, line 84: failed -> Failed

JesseStutler:
> > How about change Warning to Error?
> 
> I think it's better than just use warning.
> 
> FYI: https://github.com/kubernetes/kubernetes/blob/release-1.32/pkg/scheduler/framework/runtime/framework.go#L658

I am curious why this PR was agreed to be merged. If a plugin does not exist in kube-scheduler, the error will be returned layer by layer. Kube-scheduler would failed to start, volcano should also follow this way

PR #3995, opened 2025-02-05
Proposal: Preempt action support topology

bibibox:
#### What type of PR is this?
/kind documentation


docs/design/preempt-action-support-topology.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Story 2
+
+As a user, I expect high-priority Pod preemption to minimize impact on existing Pods while maintaining consistency of affinity rules.
+
+### Story 3
+
+When topology-sensitive resources like GPUs exist, the preemption process needs to consider resource topology relationships to ensure resource allocation after preemption still satisfies original topology constraints.
+
+For example, if a node has 2 GPUs (8GB each), Pod A and Pod B each use 4GB, and Pod C needs 8GB. Direct scheduling of Pod C will fail, triggering preemption. After removing Pod A, Pod C can be scheduled, but when re-adding Pod A, topology changes might occur due to binpack strategy. At this point, Pod C can still be scheduled, ultimately leading to preemption failure due to no pods being evicted.
```
wangyang0616, line 31: Is the example here about the current situation of volcano preemption or the challenges of the current optimization solution?

docs/design/preempt-action-support-topology.md
```diff
@@ -0,0 +1,103 @@
...
+  type SimulateRemovePodFn func(pod *api.TaskInfo, node *api.NodeInfo) error
+  ```
+
+- `SimulateAddPodFn`: Simulate the addition of a pod to a node, plugins implement this function to ensure the addition action does not cause topology changes
+
+  ```go
+  type SimulateAddPodFn func(pod *api.TaskInfo, node *api.NodeInfo) error
+  ```
+
+### Limitations
```
wangyang0616, line 100: The native Kubernetes scheduler has some capability constraints in terms of preemption, and has made certain trade-offs in terms of functionality and performance. See: [# limitations-of-preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#limitations-of-preemption)
Compared with kube-scheduler, are the functional manifestations of Volcano's affinity preemption consistent or different? If so, what are the detailed differences?

Monokaix:
There is chinese character in the img, and the subject of each process needs to be clearly identified.

Monokaix:
The process is hard to understand for common users, we'd better make it more clearer, maybe we can add both desgin process and an example.

Monokaix:
The three Key Functions are not presented in the above process design, we can give a more detailed description.

Monokaix:
What's the standard of PreemptCostNodeOrder, least evicted pod mums? We should give one.

Monokaix:
In what case a plugin should register removal and addition func, we should give a guide.

PR #3999, opened 2025-02-06
Use dynamic client to get annotations

Monokaix:
#### What type of PR is this?
/kind cleanup
/area controllers

Use dynamic client instead of swich case.

Fixes #

```release-note

```


hwdef:
There may be performance issues with dynamic client.
We'd better run a performance test.

Monokaix:
> There may be performance issues with dynamic client. We'd better run a performance test.

That's a good point.

Issue #3970, opened 2025-01-14
Support setting `minMember` when creating workloads


hwdef:
### What is the problem you're trying to solve

Currently, the default value of minMember is 1, which is inappropriate in some scenarios.
https://github.com/volcano-sh/volcano/blob/68fba2ccb3a6ba00eb511722b6b777eda375e9b4/pkg/controllers/podgroup/pg_controller_handler.go#L215

It is hoped that `minMember` can be set when creating a deployment or other non-vcjob workloads.

### Describe the solution you'd like

When creating a podgroup, we can read an annotation, such as
`scheduling.volcano.sh/minMember: 10`
Set minMember based on this value

noobzzw:
Hi, this is a great idea. Besides setting the minMember manually, can we also read the StatefulSet or ReplicaSet in the ownerReferences to obtain the minMember automatically?

hwdef:
This is probably not a good idea since many deployment functions don't want to be affected by the gang.

noobzzw:
got it.

Monokaix:
I think we can support read an annotation key from the workload, and users can set an expected value, or else the minmember is still 1.

JesseStutler:
`scheduling.volcano.sh/minMember` or `scheduling.volcano.sh/minAvailable`? From the perspective of vcjob, is minAvailable, deployment/statefulset/daemonset/job and vcjob are at the same level, so should it be minAvailable?

Monokaix:
`minAvailable` is the semantics of vcjob, and for other workloads users may just need to concern the podgroup. 
But the user experience will be more unified using`minAvailable`.

Monokaix:
Another question is: The annotation should be set at pod level or workload level, though its more reasonable setting at workload level, the current queue and podgroup annotations are both set at pod level, so it's better to keep consistent.
How do you guys think?

hwdef:
I think it's more reasonable to set at the workload level.
But we can do compatibility.  priority reading the workload annotation. If not exist, then read the pod annotation.

Monokaix:
> I think it's more reasonable to set at the workload level. But we can do compatibility. priority reading the workload annotation. If not exist, then read the pod annotation.

It seems a little confused...

lowang-bh:
It is more reasonable setting at workload level.

hwdef:
We also should consider `minResource`

Ref: https://github.com/volcano-sh/volcano/pull/4000

sceneryback:
[...]

PR #4000, opened 2025-02-07
use replicas when initializing pg minResources

sceneryback:
When creating pod using replicaSet, the minMember of podgroup created is always 1, this is not reasonable. Consider this case:
1. we created a deployment with 2 replicas, totally requesting 16 gpus 
2. the podgroup created has 1 minMember and 8 gpus
3. the capability of queue has 8 gpus and the scheduler configmap disables  AllocatableFn of proportion plugin
4. the podgroup will be enqueued and running since minResources(8 gpus) is not bigger than queue capability
5. finally the queue allocated 16 gpus, more than its deserved

As can be seen from this picture:   
image

In the case that pod has replicas from its owners, the podgroup should consider relicas, just as training-operator does, which creates pod group based on resources from all replicas

### DIFF
--- a/pkg/controllers/job/job_controller_util.go
+++ b/pkg/controllers/job/job_controller_util.go
@@ -327,10 +327,10 @@ func (p TasksPriority) CalcFirstCountResources(count int32) v1.ResourceList {
 
 	for _, task := range p {
 		if count <= task.Replicas {
-			minReq = quotav1.Add(minReq, calTaskRequests(&v1.Pod{Spec: task.Template.Spec}, count))
+			minReq = quotav1.Add(minReq, util.CalTaskRequests(&v1.Pod{Spec: task.Template.Spec}, count))
 			break
 		} else {
-			minReq = quotav1.Add(minReq, calTaskRequests(&v1.Pod{Spec: task.Template.Spec}, task.Replicas))
+			minReq = quotav1.Add(minReq, util.CalTaskRequests(&v1.Pod{Spec: task.Template.Spec}, task.Replicas))
 			count -= task.Replicas
 		}
 	}
@@ -353,7 +353,7 @@ func (p TasksPriority) CalcPGMinResources(jobMinAvailable int32) v1.ResourceList
 		if left := jobMinAvailable - podCnt; left < validReplics {
 			validReplics = left
 		}
-		minReq = quotav1.Add(minReq, calTaskRequests(&v1.Pod{Spec: task.Template.Spec}, validReplics))
+		minReq = quotav1.Add(minReq, util.CalTaskRequests(&v1.Pod{Spec: task.Template.Spec}, validReplics))
 		podCnt += validReplics
 		if podCnt >= jobMinAvailable {
 			break
@@ -377,25 +377,15 @@ func (p TasksPriority) CalcPGMinResources(jobMinAvailable int32) v1.ResourceList
 		}
 
 		if leftCnt >= left {
-			minReq = quotav1.Add(minReq, calTaskRequests(&v1.Pod{Spec: task.Template.Spec}, left))
+			minReq = quotav1.Add(minReq, util.CalTaskRequests(&v1.Pod{Spec: task.Template.Spec}, left))
 			leftCnt -= left
 		} else {
-			minReq = quotav1.Add(minReq, calTaskRequests(&v1.Pod{Spec: task.Template.Spec}, leftCnt))
+			minReq = quotav1.Add(minReq, util.CalTaskRequests(&v1.Pod{Spec: task.Template.Spec}, leftCnt))
 			leftCnt = 0
 		}
 		if leftCnt <= 0 {
 			break
 		}
-	}
-	return minReq
-}
-
-// calTaskRequests returns requests resource with validReplica replicas
-func calTaskRequests(pod *v1.Pod, validReplica int32) v1.ResourceList {
-	minReq := v1.ResourceList{}
-	usage := *util.GetPodQuotaUsage(pod)
-	for i := int32(0); i < validReplica; i++ {
-		minReq = quotav1.Add(minReq, usage)
 	}
 	return minReq
 }
--- a/pkg/controllers/podgroup/pg_controller.go
+++ b/pkg/controllers/podgroup/pg_controller.go
@@ -58,15 +58,9 @@ type pgcontroller struct {
 
 	// A store of pods
 	podLister corelisters.PodLister
-	podSynced func() bool
 
 	// A store of podgroups
 	pgLister schedulinglister.PodGroupLister
-	pgSynced func() bool
-
-	// A store of replicaset
-	rsSynced  func() bool
-	stsSynced func() bool
 
 	queue workqueue.TypedRateLimitingInterface[podRequest]
 
@@ -96,7 +90,6 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
 	pg.informerFactory = opt.SharedInformerFactory
 	pg.podInformer = opt.SharedInformerFactory.Core().V1().Pods()
 	pg.podLister = pg.podInformer.Lister()
-	pg.podSynced = pg.podInformer.Informer().HasSynced
 	pg.podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: pg.addPod,
 	})
@@ -105,18 +98,15 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
 	pg.vcInformerFactory = factory
 	pg.pgInformer = factory.Scheduling().V1beta1().PodGroups()
 	pg.pgLister = pg.pgInformer.Lister()
-	pg.pgSynced = pg.pgInformer.Informer().HasSynced
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.WorkLoadSupport) {
 		pg.rsInformer = pg.informerFactory.Apps().V1().ReplicaSets()
-		pg.rsSynced = pg.rsInformer.Informer().HasSynced
 		pg.rsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 			AddFunc:    pg.addReplicaSet,
 			UpdateFunc: pg.updateReplicaSet,
 		})
 
 		pg.stsInformer = pg.informerFactory.Apps().V1().StatefulSets()
-		pg.stsSynced = pg.stsInformer.Informer().HasSynced
 		pg.stsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 			AddFunc:    pg.addStatefulSet,
 			UpdateFunc: pg.updateStatefulSet,
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"slices"
+	"strconv"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -191,52 +192,80 @@ func (pg *pgcontroller) updatePodAnnotations(pod *v1.Pod, pgName string) error {
 	return nil
 }
 
-func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, namespace string) map[string]string {
-	switch kind {
-	case "ReplicaSet":
-		rs, err := pg.kubeClient.AppsV1().ReplicaSets(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-		if err != nil {
-			klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", kind, namespace, name, err)
-			return map[string]string{}
+func (pg *pgcontroller) getAnnotationsFromUpperRes(pod *v1.Pod) map[string]string {
+	var annotations = make(map[string]string)
+
+	for _, reference := range pod.OwnerReferences {
+		if reference.Kind != "" && reference.Name != "" {
+			tmp := make(map[string]string)
+			switch reference.Kind {
+			case "ReplicaSet":
+				rs, err := pg.kubeClient.AppsV1().ReplicaSets(pod.Namespace).Get(context.TODO(), reference.Name, metav1.GetOptions{})
+				if err != nil {
+					klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", reference.Kind, pod.Namespace, reference.Name, err)
+					continue
+				}
+				tmp = rs.Annotations
+			case "DaemonSet":
+				ds, err := pg.kubeClient.AppsV1().DaemonSets(pod.Namespace).Get(context.TODO(), reference.Name, metav1.GetOptions{})
+				if err != nil {
+					klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", reference.Kind, pod.Namespace, reference.Name, err)
+					continue
+				}
+				tmp = ds.Annotations
+			case "StatefulSet":
+				ss, err := pg.kubeClient.AppsV1().StatefulSets(pod.Namespace).Get(context.TODO(), reference.Name, metav1.GetOptions{})
+				if err != nil {
+					klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", reference.Kind, pod.Namespace, reference.Name, err)
+					continue
+				}
+				tmp = ss.Annotations
+			case "Job":
+				job, err := pg.kubeClient.BatchV1().Jobs(pod.Namespace).Get(context.TODO(), reference.Name, metav1.GetOptions{})
+				if err != nil {
+					klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", reference.Kind, pod.Namespace, reference.Name, err)
+					continue
+				}
+				tmp = job.Annotations
+			}
+
+			for k, v := range tmp {
+				if _, ok := annotations[k]; !ok {
+					annotations[k] = v
+				}
+			}
 		}
-		return rs.Annotations
-	case "DaemonSet":
-		ds, err := pg.kubeClient.AppsV1().DaemonSets(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-		if err != nil {
-			klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", kind, namespace, name, err)
-			return map[string]string{}
-		}
-		return ds.Annotations
-	case "StatefulSet":
-		ss, err := pg.kubeClient.AppsV1().StatefulSets(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-		if err != nil {
-			klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", kind, namespace, name, err)
-			return map[string]string{}
-		}
-		return ss.Annotations
-	case "Job":
-		job, err := pg.kubeClient.BatchV1().Jobs(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-		if err != nil {
-			klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", kind, namespace, name, err)
-			return map[string]string{}
-		}
-		return job.Annotations
-	default:
-		return map[string]string{}
 	}
+
+	return annotations
+}
+
+func (pg *pgcontroller) getMinMemberFromUpperRes(upperAnnotations map[string]string, namespance, name string) int32 {
+	minMember := int32(1)
+
+	if minMemberAnno, ok := upperAnnotations[scheduling.VolcanoGroupMinMemberAnnotationKey]; ok {
+		minMemberFromAnno, err := strconv.ParseInt(minMemberAnno, 10, 32)
+		if err != nil {
+			klog.Errorf("Failed to convert minMemberAnnotation of Pod owners <%s/%s> into number: %v, minMember remains as 1",
+				namespance, name, err)
+			return minMember
+		}
+		if minMemberFromAnno < 0 {
+			klog.Errorf("minMemberAnnotation %d is not positive, minMember remains as 1", minMemberFromAnno)
+			return minMember
+		}
+		minMember = int32(minMemberFromAnno)
+	}
+
+	return minMember
 }
 
 // Inherit annotations from upper resources.
-func (pg *pgcontroller) inheritUpperAnnotations(pod *v1.Pod, obj *scheduling.PodGroup) {
+func (pg *pgcontroller) inheritUpperAnnotations(upperAnnotations map[string]string, obj *scheduling.PodGroup) {
 	if pg.inheritOwnerAnnotations {
-		for _, reference := range pod.OwnerReferences {
-			if reference.Kind != "" && reference.Name != "" {
-				var upperAnnotations = pg.getAnnotationsFromUpperRes(reference.Kind, reference.Name, pod.Namespace)
-				for k, v := range upperAnnotations {
-					if strings.HasPrefix(k, scheduling.AnnotationPrefix) {
-						obj.Annotations[k] = v
-					}
-				}
+		for k, v := range upperAnnotations {
+			if strings.HasPrefix(k, scheduling.AnnotationPrefix) {
+				obj.Annotations[k] = v
 			}
 		}
 	}
@@ -252,6 +281,13 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
 			return err
 		}
 
+		var minMember = int32(1)
+		var ownerAnnotations = make(map[string]string)
+		if pg.inheritOwnerAnnotations {
+			ownerAnnotations = pg.getAnnotationsFromUpperRes(pod)
+			minMember = pg.getMinMemberFromUpperRes(ownerAnnotations, pod.Namespace, pod.Name)
+		}
+		minResources := util.CalTaskRequests(pod, minMember)
 		obj := &scheduling.PodGroup{
 			ObjectMeta: metav1.ObjectMeta{
 				Namespace:       pod.Namespace,
@@ -261,16 +297,16 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
 				Labels:          map[string]string{},
 			},
 			Spec: scheduling.PodGroupSpec{
-				MinMember:         1,
+				MinMember:         minMember,
 				PriorityClassName: pod.Spec.PriorityClassName,
-				MinResources:      util.GetPodQuotaUsage(pod),
+				MinResources:      &minResources,
 			},
 			Status: scheduling.PodGroupStatus{
 				Phase: scheduling.PodGroupPending,
 			},
 		}
 
-		pg.inheritUpperAnnotations(pod, obj)
+		pg.inheritUpperAnnotations(ownerAnnotations, obj)
 		// Individual annotations on pods would overwrite annotations inherited from upper resources.
 		if queueName, ok := pod.Annotations[scheduling.QueueNameAnnotationKey]; ok {
 			obj.Spec.Queue = queueName
--- a/pkg/controllers/util/util.go
+++ b/pkg/controllers/util/util.go
@@ -4,17 +4,28 @@ import (
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
+	quotav1 "k8s.io/apiserver/pkg/quota/v1"
 	"k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	quotacore "k8s.io/kubernetes/pkg/quota/v1/evaluator/core"
 	"k8s.io/utils/clock"
 )
 
-func GetPodQuotaUsage(pod *v1.Pod) *v1.ResourceList {
+func GetPodQuotaUsage(pod *v1.Pod) v1.ResourceList {
 	res, _ := quotacore.PodUsageFunc(pod, clock.RealClock{})
 	for name, quantity := range res {
 		if !helper.IsNativeResource(name) && strings.HasPrefix(string(name), v1.DefaultResourceRequestsPrefix) {
 			res[v1.ResourceName(strings.TrimPrefix(string(name), v1.DefaultResourceRequestsPrefix))] = quantity
 		}
 	}
-	return &res
+	return res
+}
+
+// calTaskRequests returns requests resource with validReplica replicas
+func CalTaskRequests(pod *v1.Pod, validReplica int32) v1.ResourceList {
+	minReq := v1.ResourceList{}
+	usage := GetPodQuotaUsage(pod)
+	for i := int32(0); i < validReplica; i++ {
+		minReq = quotav1.Add(minReq, usage)
+	}
+	return minReq
 }
--- a/pkg/controllers/podgroup/pg_controller_test.go
+++ b/pkg/controllers/podgroup/pg_controller_test.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"testing"
 
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/informers"
@@ -46,6 +48,7 @@ func newFakeController() *pgcontroller {
 		SharedInformerFactory:   sharedInformers,
 		VCSharedInformerFactory: vcSharedInformers,
 		SchedulerNames:          []string{"volcano"},
+		InheritOwnerAnnotations: true,
 	}
 
 	controller.Initialize(opt)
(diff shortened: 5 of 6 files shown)


hwdef:
Same situation as mentioned in this issue:
https://github.com/volcano-sh/volcano/issues/3970

Monokaix:
proportion plugin controls the queue related capabilities, if you disable it, the capacity check will not take effect.

Monokaix:
Just as @hwdef said, please set an annotation to control the minmember to set, and also other workload type like statefulset should also has this capability: )

sceneryback:
hi @hwdef @Monokaix thanks for all your suggestions, I have updated the feature using new annotation, so it won't break current behaviors. Please have a look.     
apis change: https://github.com/volcano-sh/apis/pull/154
I will update go.mod to delete local references of apis package after https://github.com/volcano-sh/apis/pull/154 is merged

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +181,37 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
 	}
 }
 
+func (pg *pgcontroller) getMinMemberFromUpperRes(pod *v1.Pod) *int32 {
+	defaultMinMember := int32(1)
+	annotations := make(map[string]string)
+
+	for _, reference := range pod.OwnerReferences {
+		if reference.Kind != "" && reference.Name != "" {
+			tmp := pg.getAnnotationsFromUpperRes(reference.Kind, reference.Name, pod.Namespace)
```
JesseStutler, line 190: Usually we write annotation on the deployment to specify minMember, but here the ownerReferences of the Pod can only be traced back to the replicaset.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +181,37 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+			if err := mergo.Merge(&annotations, &tmp); err != nil {
```
JesseStutler, line 191: Do we have scenarios that Pod has multiple OwnerReferences? If there are trully multiple annotations, we won't specify twice the minMember annotations, do we need to use the merge.Merge?

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +181,37 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
 	}
 }
 
+func (pg *pgcontroller) getMinMemberFromUpperRes(pod *v1.Pod) *int32 {
+	defaultMinMember := int32(1)
```
JesseStutler, line 185: This variable name is very strange. You can set 1 as a constant (as defaultMinMember), and then the variable here is initialized to defaultMinMember. `minMember` here is ok

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -193,7 +228,7 @@ func (pg *pgcontroller) inheritUpperAnnotations(pod *v1.Pod, obj *scheduling.Pod
 	}
 }
 
-func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
+func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod, minAvailable *int32) error {
```
JesseStutler, line 231: Why are the variable names here not unified as minMember?

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -203,6 +238,11 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
 			return err
 		}
 
+		minMember := int32(1)
```
JesseStutler, line 241: There is no need to initialize again here, you already have the initialization value in `getMinMemberFromUpperRes`.

pkg/controllers/job/job_controller_util.go
```diff
@@ -327,10 +327,10 @@ func (p TasksPriority) CalcFirstCountResources(count int32) v1.ResourceList {
 
 	for _, task := range p {
 		if count <= task.Replicas {
-			minReq = quotav1.Add(minReq, calTaskRequests(&v1.Pod{Spec: task.Template.Spec}, count))
+			minReq = quotav1.Add(minReq, *util.CalTaskRequests(&v1.Pod{Spec: task.Template.Spec}, count))
```
JesseStutler, line 330: The return value of the function here is a pointer, and then you use `*` directly to get the value pointed to by the pointer, the way it's written looks a bit strange and doesn't quite fit in with go's clean code. The return value of the func `CalTaskRequests` does not need to be refactored, keeping the original is better. In fact, when the API was designed, minResources should not be a pointer, ResourceList is a map.

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: From k8s source code https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L235, the ReplicaSet will inherit annotations from Deployment, so I think it's enough to get annotations from RS
image

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: Normally there won't be multiple OwnerReferences with this annotation, updated. Thanks

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: Updated

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: Updated

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: updated

pkg/controllers/job/job_controller_util.go
sceneryback: simply copied from an existing util function:   
image
I will update `GetPodQuotaUsage` too    
Also notice that the minResources field of PodGroup is designed as pointer originally, so currently won't break this

sceneryback:
hi @JesseStutler Thanks for your reviewing! Have updated, please have a look

Monokaix:
API related pr is merged, you can move on: )

sceneryback:
Updated go.mod, please have a look @Monokaix @hwdef @JesseStutler

hwdef (commented):
/lgtm

pkg/controllers/util/util.go
```diff
@@ -4,17 +4,28 @@ import (
...
 
 	v1 "k8s.io/api/core/v1"
+	quotav1 "k8s.io/apiserver/pkg/quota/v1"
 	"k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	quotacore "k8s.io/kubernetes/pkg/quota/v1/evaluator/core"
 	"k8s.io/utils/clock"
 )
 
-func GetPodQuotaUsage(pod *v1.Pod) *v1.ResourceList {
+func GetPodQuotaUsage(pod *v1.Pod) v1.ResourceList {
```
Monokaix, line 13: Why changed to non-poiner?

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -172,7 +172,8 @@ func (pg *pgcontroller) processNextReq() bool {
 
 	// normal pod use volcano
 	klog.V(4).Infof("Try to create podgroup for pod %s/%s", pod.Namespace, pod.Name)
-	if err := pg.createNormalPodPGIfNotExist(pod); err != nil {
+	minMember := pg.getMinMemberFromUpperRes(pod)
```
Monokaix, line 175: This logic should put in `createNormalPodPGIfNotExist`, because when the podgroup is already existed, there is no need to calculate the min member, it will request APIServer which is a heavy operation.
And there is also a qustion mentioned here https://github.com/volcano-sh/volcano/issues/3970#issuecomment-2652862212

pkg/controllers/util/util.go
sceneryback: Just follow this comment from @JesseStutler for better style https://github.com/volcano-sh/volcano/pull/4000#discussion_r1948988135

pkg/controllers/podgroup/pg_controller.go
sceneryback: very good point 👍 so it's better to keep rs/sts etc informers in pgcontroller and read cache from these informers, right?

Monokaix: Yeah, DaemonSet and Job should also be considered.

Monokaix: Also this should also be considered https://github.com/volcano-sh/volcano/issues/3970#issuecomment-2638750095

sceneryback: > Also this should also be considered [#3970 (comment)](https://github.com/volcano-sh/volcano/issues/3970#issuecomment-2638750095)

I think it's not necessary to check pod annotation? The annotation is designed for workload, it will be strange to see this in pod

sceneryback:
Refactored code using informers, please have a look again @hwdef @Monokaix @JesseStutler

hwdef:
Hi, please make the CI happy

sceneryback:
hi @hwdef fixed the failed e2e tests, please have a look again

hwdef (commented):
/lgtm
Thanks!

hwdef:
PTAL
@Monokaix @JesseStutler @lowang-bh

go.mod
```diff
@@ -46,7 +46,7 @@ require (
 	sigs.k8s.io/controller-runtime v0.13.0
 	sigs.k8s.io/yaml v1.4.0
 	stathat.com/c/consistent v1.0.0
-	volcano.sh/apis v1.10.0-alpha.0.0.20241210014034-bf27f4e986d0
+	volcano.sh/apis v1.11.1-0.20250211082520-7f8222e881d9
```
Monokaix, line 49: Why 1.11.1?

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -99,15 +118,29 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
 		AddFunc: pg.addPod,
 	})
 
+	pg.rsInformer = pg.informerFactory.Apps().V1().ReplicaSets()
```
Monokaix, line 121: These should be under statement `if utilfeature.DefaultFeatureGate.Enabled(features.WorkLoadSupport) {`.

Monokaix, line 121: I think we should add a param feature switch to users whether enable non-vcjob supports minMember, becasue it's a little heavy.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,33 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+		// Currently we assume the group-min-member annotation will be specified only once
+		if reference.Kind != "" && reference.Name != "" {
+			annotations := pg.getAnnotationsFromUpperRes(reference.Kind, reference.Name, pod.Namespace)
+			if minMemberAnno, ok := annotations[scheduling.VolcanoGroupMinMemberAnnotationKey]; ok {
+				minMemberFromAnno, err := strconv.Atoi(minMemberAnno)
+				if err != nil {
+					klog.Infof("Failed to convert minMemberAnnotation of Pod owners <%s/%s> into number, skipping",
+						pod.Namespace, pod.Name)
+					return minMember
+				}
```
Monokaix, line 195: < 0 is ok?

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,33 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+	minMember := int32(1)
+
+	for _, reference := range pod.OwnerReferences {
+		// Currently we assume the group-min-member annotation will be specified only once
+		if reference.Kind != "" && reference.Name != "" {
+			annotations := pg.getAnnotationsFromUpperRes(reference.Kind, reference.Name, pod.Namespace)
+			if minMemberAnno, ok := annotations[scheduling.VolcanoGroupMinMemberAnnotationKey]; ok {
+				minMemberFromAnno, err := strconv.Atoi(minMemberAnno)
+				if err != nil {
+					klog.Infof("Failed to convert minMemberAnnotation of Pod owners <%s/%s> into number, skipping",
```
Monokaix, line 192: This klog info is not so clear, we should mention that remain minmember =1, or users will don't  know the value finally set.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,33 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+						pod.Namespace, pod.Name)
+					return minMember
+				}
+				if minMemberFromAnno < math.MinInt32 || minMemberFromAnno > math.MaxInt32 {
+					klog.Infof("minMember %d exceeds bounds of int32, skipping", minMember)
```
Monokaix, line 197: same above.

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -99,15 +118,29 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
...
+	pg.rsLister = pg.rsInformer.Lister()
+	pg.rsSynced = pg.rsInformer.Informer().HasSynced
```
JesseStutler, line 123: There is no need to add these rsSynced funcs, these codes:
```
	for informerType, ok := range pg.informerFactory.WaitForCacheSync(stopCh) {
		if !ok {
			klog.Errorf("caches failed to sync: %v", informerType)
		}
	}
	for informerType, ok := range pg.vcInformerFactory.WaitForCacheSync(stopCh) {
		if !ok {
			klog.Errorf("caches failed to sync: %v", informerType)
			return
		}
	}
```
has already done this

installer/helm/chart/volcano/templates/controllers.yaml
```diff
@@ -74,13 +74,13 @@ rules:
     verbs: ["get", "create", "delete"]
   - apiGroups: ["apps"]
     resources: ["daemonsets", "statefulsets"]
-    verbs: ["get"]
+    verbs: ["get", "list"]
```
JesseStutler, line 77: If we change all these APIs to list/watch them, I think the verbs here should be ["list","watch"]?

JesseStutler:
Better to post a screenshot to verify your codes work in actual environment thanks

pkg/controllers/podgroup/pg_controller.go
sceneryback: rsLister is also used in 'inheritUpperAnnotations()' when WorkLoadSupport is not enabled, so only add these codes under the statement:   
```
pg.rsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
    AddFunc:    pg.addReplicaSet,
    UpdateFunc: pg.updateReplicaSet,
})
```

sceneryback: hi @Monokaix you mean a feature gate like features.NonVcjobMinMemberSupport which only applies logic in this PR when enabled?

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: should be positive, will fix

pkg/controllers/podgroup/pg_controller.go
Monokaix: a param flag is ok.

Monokaix: WorkLoadSupport feature means whether Volcano supports deployment/rs/ds/sts,etc., so the informers should be initiated when featuregate is not true.

Monokaix:
Hi, are you still working on this? It's a nice feature and hope you can resolve the comments: )

hwdef:
Please rebase the master branch.
This PR is important.

sceneryback:
sorry for recent inactivity, I will continue working on this feature

hwdef:
Never mind. Looking forward to your new code. :)

sceneryback:
> Better to post a screenshot to verify your codes work in actual environment thanks

deployment with min-member as 2:   
image
deployment with min-member as 3:
image
deployment with min-member as 0:
image
daemonset with min-member as 2:
image
job with min-member as 2:
image
statefulset with min-member as 2:
image

sceneryback:
Hi @hwdef @Monokaix @JesseStutler Updated code according to previous comments, mainly including:  
1. add a param flag 'non-vcjob-minmember-support' to enable this feature which is true by default
2. add more validation for this anno, especially deployment, statefulset and job
3. test this feature in local k8s cluster and provide screenshots for each kind

Please have a look again, thanks!

hwdef:
Excellent! I'll check this later.

pkg/controllers/podgroup/pg_controller.go
JesseStutler: Do you mean to add another gate called `NonVcjobMinMemberSupport` to control whether to use informers for deployment/rs/sts/job? But `WorkLoadSupport` is used to control whether to support these resources

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,58 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+	minMember := int32(1)
+
+	for _, reference := range pod.OwnerReferences {
+		// Currently we assume the group-min-member annotation will be specified only once
+		if reference.Kind != "" && reference.Name != "" {
+			annotations := pg.getAnnotationsFromUpperRes(reference.Kind, reference.Name, pod.Namespace)
+			if minMemberAnno, ok := annotations[scheduling.VolcanoGroupMinMemberAnnotationKey]; ok {
+				minMemberFromAnno, err := strconv.Atoi(minMemberAnno)
+				if err != nil {
+					klog.Infof("Failed to convert minMemberAnnotation of Pod owners <%s/%s> into number, minMember remains as 1",
```
JesseStutler, line 209: klog.Errof?

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,58 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+						pod.Namespace, pod.Name)
+					return minMember
+				}
+				if minMemberFromAnno <= 0 {
+					klog.Infof("minMemberAnnotation %d is not positive, minMember remains as 1", minMemberFromAnno)
```
JesseStutler, line 214: Same as above, need to use klog.Errof

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,58 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
 
+func (pg *pgcontroller) isMinMemberAnnotationValidForKind(minMemberFromAnno int, kind string, name string, namespace string) bool {
+	switch kind {
+	case "ReplicaSet":
+		rs, err := pg.rsLister.ReplicaSets(namespace).Get(name)
+		if err != nil {
+			klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", kind, namespace, name, err)
+			return false
+		}
+		return *rs.Spec.Replicas >= int32(minMemberFromAnno)
```
JesseStutler, line 190: Why do we need this verification? The verification here in the scheduler is not noticeable to the users, this is something that should be handled by the webhook manager

JesseStutler: Need to discuss should we verify this if replicas of workload is larger than minMemberFromAnno? @Monokaix @hwdef

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,58 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+	case "StatefulSet", "Job":
```
JesseStutler, line 191: Why do Statefulset and job not need it?

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: If minMemberFromAnno is larger than replicas, it will fail the validFunc in gang plugin: 
```
		vtn := job.ValidTaskNum()
		if vtn < job.MinAvailable {
			return &api.ValidateResult{
				Pass:   false,
				Reason: v1beta1.NotEnoughPodsReason,
				Message: fmt.Sprintf("Not enough valid tasks for gang-scheduling, valid: %d, min: %d",
					vtn, job.MinAvailable),
			}
		}
```
e.g. replicas is 2 and minMemberFromAnno is 3, it will show 'Not enough valid tasks for gang-scheduling, valid: 2, min: 3' and the job will never be scheduled successfully

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: Because I think it's meaningless for Statefulset (pod created 1 by 1) and job (1 pod), you cannot set a minMemberFromAnno larger than 1, or the ss and job will fail the validFunc in gang plugin for the same reason as before. Use the default 1 is enough for ss and job.

pkg/controllers/podgroup/pg_controller.go
sceneryback: > Do you mean to add another gate called `NonVcjobMinMemberSupport` to control whether to use informers for deployment/rs/sts/job? But `WorkLoadSupport` is used to control whether to support these resources

No, just add a param arg following the comments of @Monokaix to enable this feature

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: fixed, thanks

pkg/controllers/podgroup/pg_controller_handler.go
hwdef: > Why do we need this verification? The verification here in the scheduler is not noticeable to the users, this is something that should be handled by the webhook manager

I recommend doing verification in the controller as well, because the current trend is to remove webhooks to enhance performance https://github.com/volcano-sh/volcano/issues/4127

pkg/controllers/podgroup/pg_controller.go
JesseStutler: So why not use WorkloadSupport uniformly?

pkg/controllers/podgroup/pg_controller_handler.go
JesseStutler: > If minMemberFromAnno is larger than replicas, it will fail the validFunc in gang plugin:
> 
> ```
> 		vtn := job.ValidTaskNum()
> 		if vtn < job.MinAvailable {
> 			return &api.ValidateResult{
> 				Pass:   false,
> 				Reason: v1beta1.NotEnoughPodsReason,
> 				Message: fmt.Sprintf("Not enough valid tasks for gang-scheduling, valid: %d, min: %d",
> 					vtn, job.MinAvailable),
> 			}
> 		}
> ```
> 
> e.g. replicas is 2 and minMemberFromAnno is 3, it will show 'Not enough valid tasks for gang-scheduling, valid: 2, min: 3' and the job will never be scheduled successfully

If the user misconfigures it, they should be able to notice it. If it defaults to 1 when misconfigured, the user won't even notice

pkg/controllers/podgroup/pg_controller_handler.go
JesseStutler: The pod of a job may not require gang scheduling, but the pod of an sts might still need it

pkg/controllers/podgroup/pg_controller.go
hwdef: @sceneryback 
The nonvcjobminmembersupport flag is not very useful in the current code implementation.
I think @Monokaix means that if `nonvcjobminmembersupport` is true, then enable these informers, otherwise do not enable these informers because he thinks these informers are a bit heavy


@Monokaix Is it correct that I understand?

Monokaix: Yaeh that's exactly what I mean.

cmd/controller-manager/app/options/options.go
```diff
@@ -124,6 +126,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")
 	fs.StringVar(&s.ListenAddress, "listen-address", defaultListenAddress, "The address to listen on for HTTP requests.")
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
+	fs.BoolVar(&s.NonVcjobMinMemberSupport, "non-vcjob-min-member-support", true, "NonVcjobMinMemberSupport can initialize podgroup minMember and minResources for non-vcjobs(deployment/statefulset/daemonset/job).")
```
Monokaix, line 129: default value = false is better.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,58 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+				if err != nil {
+					klog.Errorf("Failed to convert minMemberAnnotation of Pod owners <%s/%s> into number: %v, minMember remains as 1",
+						pod.Namespace, pod.Name, err)
+					return minMember
+				}
+				if minMemberFromAnno <= 0 {
+					klog.Errorf("minMemberAnnotation %d is not positive, minMember remains as 1", minMemberFromAnno)
+					return minMember
+				}
+				if minMemberFromAnno < math.MinInt32 || minMemberFromAnno > math.MaxInt32 {
```
Monokaix, line 217: You already used `strconv.Atoi` to parse a int number, so it's unnecessary to check the integer scope here.

pkg/controllers/podgroup/pg_controller.go
Monokaix: I forget that there is already a flag `inheritOwnerAnnotations` to get annotations from upper resource, so for simplicity, just re-use this flag and no need to add a new flag `nonvcjobminmembersupport` and no informer is neeed.

Monokaix: > I forget that there is already a flag `inheritOwnerAnnotations` to get annotations from upper resource, so for simplicity, just re-use this flag and no need to add a new flag `nonvcjobminmembersupport` and no informer is neeed.

Please modify based on this, it will be clearer and simpler.

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: This checking is meant to avoid overflow because finally the value is assigned to minMember (int32)      
```
minMember = int32(minMemberFromAnno)
```

pkg/controllers/podgroup/pg_controller.go
sceneryback: you mean 1. drop this new flag 'NonVcjobMinMemberSupport'  2. drop the informers even if WorkloadSupport is true or init informers when WorkloadSupport is true?     I think WorkloadSupport is more proper than inheritOwnerAnnotations to enable this feature  @Monokaix @hwdef @JesseStutler 

> > I forget that there is already a flag `inheritOwnerAnnotations` to get annotations from upper resource, so for simplicity, just re-use this flag and no need to add a new flag `nonvcjobminmembersupport` and no informer is neeed.
> 
> Please modify based on this, it will be clearer and simpler.

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: > sts

It's not easy to check the existence of gang plugin in controller code, so keep the current behavior is OK? i.e. setting minMember to 1 for job and sts @JesseStutler

pkg/controllers/podgroup/pg_controller_handler.go
Monokaix: If it overflow, the `strconv.Atoi` will return an err, that's to say, the validation has already been done.

pkg/controllers/podgroup/pg_controller.go
Monokaix: Yes use the `inheritOwnerAnnotations` is ok, and the feature we claim to the public is still non-vcjob support for setting minmember, while the detailed setting is to turn on this switch, the feature name and flag name do not need to be exactly the same. `WorkloadSupport` is used to control whether list/watch, we don't need informer any more on this feature.

pkg/controllers/podgroup/pg_controller_handler.go
sceneryback: @Monokaix  If will fail the CodeQL: 
image

Monokaix: We can use `minMember,err := strconv.ParseInt(str,10,32)` instead of `strconv.Atoi`.

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -57,14 +57,9 @@ type pgcontroller struct {
 
 	// A store of pods
 	podLister corelisters.PodLister
-	podSynced func() bool
```
Monokaix, line 60: We should keep this.

pkg/controllers/podgroup/pg_controller_handler.go
Monokaix: Just keep the minMember value users specified, if minMember > replica, job will be in Pending and users can be aware of that.

pkg/controllers/podgroup/pg_controller_handler.go
Monokaix: Sts can also create pods concurrently, and job can has multi replicas. We should support this feature for sts and job, and user can choose whether to config it according to their own use case.

Monokaix:
> 1. reverted back to the none-informer's version
> 2. left the logic in isMinMemberAnnotationValidForKind unchanged, it may still need some discussion.
> 
> Please have a look again, thanks! @hwdef @Monokaix @JesseStutler

see https://github.com/volcano-sh/volcano/pull/4000/files#r2043332608

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -177,6 +179,58 @@ func (pg *pgcontroller) getAnnotationsFromUpperRes(kind string, name string, nam
...
+		if reference.Kind != "" && reference.Name != "" {
+			annotations := pg.getAnnotationsFromUpperRes(reference.Kind, reference.Name, pod.Namespace)
+			if minMemberAnno, ok := annotations[scheduling.VolcanoGroupMinMemberAnnotationKey]; ok {
+				minMemberFromAnno, err := strconv.Atoi(minMemberAnno)
+				if err != nil {
+					klog.Errorf("Failed to convert minMemberAnnotation of Pod owners <%s/%s> into number: %v, minMember remains as 1",
+						pod.Namespace, pod.Name, err)
+					return minMember
+				}
+				if minMemberFromAnno <= 0 {
```
Monokaix, line 213: =0 is also valid.

Monokaix:
From my side, just fix the left comments: ), other parts lgtm.

sceneryback:
> > 1. reverted back to the none-informer's version
> > 2. left the logic in isMinMemberAnnotationValidForKind unchanged, it may still need some discussion.
> > 
> > Please have a look again, thanks! @hwdef @Monokaix @JesseStutler
> 
> see https://github.com/volcano-sh/volcano/pull/4000/files#r2043332608

Thanks for your suggestions! fixed

sceneryback:
@hwdef @Monokaix @JesseStutler updated code according to latest comments, please have a look again

Monokaix:
Well done! Nearly finished.

hwdef (commented):
/lgtm

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -203,6 +231,11 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
 			return err
 		}
 
+		var minMember = int32(1)
+		if pg.inheritOwnerAnnotations {
+			minMember = pg.getMinMemberFromUpperRes(pod)
```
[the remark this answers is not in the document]
sceneryback: fixed, thanks

sceneryback:
changed signatures of some functions for variable reuse, please have a look again @Monokaix @hwdef @JesseStutler

pkg/controllers/podgroup/pg_controller_handler.go
hwdef: Can we not import the Mergo package? Feeling little necessary

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -191,52 +193,78 @@ func (pg *pgcontroller) updatePodAnnotations(pod *v1.Pod, pgName string) error {
...
+			case "Job":
+				job, err := pg.kubeClient.BatchV1().Jobs(pod.Namespace).Get(context.TODO(), reference.Name, metav1.GetOptions{})
+				if err != nil {
+					klog.Errorf("Failed to get upper %s for Pod <%s/%s>: %v", reference.Kind, pod.Namespace, reference.Name, err)
+					continue
+				}
+				tmp = job.Annotations
+			}
+
+			if err := mergo.Merge(&annotations, &tmp); err != nil {
```
hwdef, line 233: Can we not import the Mergo package? Feeling little necessary

sceneryback: or simply write our own map merging func, e.g: 
```
func mergeMap(src, dest map[string]string) {
    for k, v := range dest {
        if _, ok := src[k]; !ok {
            src[k] = v
        }
    }
}
```
or just use an inline one:   
```
    for k, v := range tmp {
        if _, ok := annotations[k]; !ok {
            annotations[k] = v
        }
    }
```
is this OK? @hwdef

hwdef: It's good:)

sceneryback: fixed, please have a look @hwdef

hwdef:
/lgtm
Thanks!

hwdef:
@Monokaix @JesseStutler @lowang-bh 
PTAL. Let's strike while the iron is hot.

Monokaix:
Overall ok to me, please keep one commit.

hwdef:
/lgtm

Greate work!

PR #4012, opened 2025-02-14
skip the jobs that have no tasks during the close session step in gang plugin

### LINKED ISSUE #4011: Boundary Condition Considerations in Gang Plugin's OnSessionClose

The Gang Plugin's OnSessionClose updates the conditions of the PodGroup (PG) based on whether the current scheduling Job is Ready. In certain scenarios, I believe this behavior is unnecessary.

Scenario:
When a Volcano Job (VJ) is deleted, the PodGroup (PG) and Pods are cascadingly deleted. Since the Pod deletion event is observed earlier than the PG deletion event, `vj.pg != nil` while `len(vj.tasks) == 0` at this point. This causes the Gang Plugin to unnecessarily update the PG conditions in this scenario.

Proposed Improvement:
The fundamental premise of Gang scheduling is that there is at least one Pod to schedule. If vj.tasks is empty (i.e., no tasks require scheduling), the plugin should skip updating the PG conditions.

### Additional context

_No response_

Wang-Kai:
#### What type of PR is this?

/kind flake
/area scheduling

In the `OnSessionClose` function, the new PR skip the unavailable job. Unnecessary podgroup condition updates are avoided in the following scenarios:

1. In the vj deletion scenario, the Pod deletion event is watched earlier than the PG deletion event. At this time, `len(jobInfo.Tasks) == 0 && jobInfo.PodGroup != nil`, and a new PodGroupUnschedulableType condition is added
2. In the vj creation scenario, the podgroup has been created but the pod has not been created yet. In this case, a new PodGroupUnschedulableType condition is added similar to scenario 1

Fixes #4011

NONE

NONE

### DIFF
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -163,6 +163,10 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 	var unreadyTaskCount int32
 	var unScheduleJobCount int
 	for _, job := range ssn.Jobs {
+		// skip the jobs that have no tasks.
+		if len(job.Tasks) == 0 {
+			continue
+		}
 		if !job.IsReady() {
 			schedulableTaskNum := func() (num int32) {
 				for _, task := range job.TaskStatusIndex[api.Pending] {


hwdef (commented):
/lgtm
/ok-to-test

PR #4025, opened 2025-02-18
[design]: queue level scheduling policies

### LINKED ISSUE #3992: volcano supports queue-level scheduling policies

Volcano supports unified scheduling of online and offline workloads, provides a wealth of scheduling plugins and algorithms, and can distinguish different tenants through queue distinction. The current scheduling policy is a global configuration, and all jobs in the queue use the same scheduling policy, but in actual scenarios, different tenants may need to use different scheduling policies due to different usage scenarios. Therefore, volcano needs to support setting and using different scheduling policies at the queue level instead of using a globally unified scheduling policy.

We can add a new field to the queue CRD, and users can set scheduling policies at the queue level, and volcano scheduler implements different scheduling policies based on the queue in which the job is located.

### Additional context

_No response_

mahdikhashan:
#### What type of PR is this?

/kind feature
/area controllers
/area scheduling

@Monokaix 
@JesseStutler 

this is a design document for the queue level scheduling policy.

Fixes #3992 

```release-note

```

### DIFF
--- a/docs/design/queue-level-scheduling-policy.md
+++ b/docs/design/queue-level-scheduling-policy.md
@@ -1,0 +1,120 @@
+Queue Level Scheduling Policy
+---
+
+by [@mahdikhashan](http://github.com/mahdikhashan); 18 Feb, 2025.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


mahdikhashan:
> please rebase master and squash your commits to only one. Thanks.

done. thanks for your time.

mahdikhashan:
closing since my proposal was not selected.

PR #4029, opened 2025-02-21
fix: the problem that the pending tasks cannot be scheduled during the backfill action

### LINKED ISSUE #4028: the problem that the pending tasks cannot be scheduled during the backfill action
### Description

During the backfill action, in the loop of pending tasks, if a previous task fails to match a node or an exception occurs when calling the PrePredicateFn method, all subsequent tasks will stop being scheduled.

1.First, create a pod. Set the requested CPU and memory resources of this pod to 0. Schedule this pod using the backfill action. Set an affinity, such as nodeSelector, to prevent this pod from being scheduled to any node.
2.Then create another pod named pod2. Set the requested CPU and memory resources of this pod2 to 0, and schedule this pod using the backfill action.
3.At this point, pod2 cannot be scheduled.

What is expected is that pod1 cannot be scheduled, while pod2 can be scheduled normally.

v1.10

https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/actions/backfill/backfill.go#L74
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/actions/backfill/backfill.go#L80 
The break statements in the above two lines of code should be changed to continue to allow subsequent tasks to be scheduled normally.

hansongChina:
During the backfill action, in the loop of pending tasks, if a previous task fails to match a node or an exception occurs when calling the PrePredicateFn method, all subsequent tasks will stop being scheduled.

Fixes #4028

```release-note

```

### DIFF
--- a/pkg/scheduler/actions/backfill/backfill.go
+++ b/pkg/scheduler/actions/backfill/backfill.go
@@ -71,13 +71,13 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 				fe.SetNodeError(ni.Name, err)
 			}
 			job.NodesFitErrors[task.UID] = fe
-			break
+			continue
 		}
 
 		predicateNodes, fitErrors := ph.PredicateNodes(task, ssn.NodeList, predicateFunc, backfill.enablePredicateErrorCache)
 		if len(predicateNodes) == 0 {
 			job.NodesFitErrors[task.UID] = fitErrors
-			break
+			continue
 		}
 
 		node := predicateNodes[0]


hwdef (commented):
/lgtm

JesseStutler:
Great. But I'm wondering isn't there also a problem in the allocate action? After the break here: https://github.com/volcano-sh/volcano/blob/5d66534726334f213134a3e9f729d864166ab2fb/pkg/scheduler/actions/allocate/allocate.go#L215, the job is not pushed back, and the remaining pods in this job are not scheduled

lowang-bh:
> Great. But I'm wondering isn't there also a problem in the allocate action? After the break here:
> 
> https://github.com/volcano-sh/volcano/blob/5d66534726334f213134a3e9f729d864166ab2fb/pkg/scheduler/actions/allocate/allocate.go#L215
> 
> , the job is not pushed back, and the remaining pods in this job are not scheduled
> cc @Monokaix @lowang-bh @hwdef

It is ok when minMember equals to total replicas.
Otherwise, maybe it need to consider as this:
https://github.com/volcano-sh/volcano/blob/a61fc90b8e72f8b688390d974f78c3292c041cfc/pkg/scheduler/actions/allocate/allocate.go#L224

Monokaix:
> > Great. But I'm wondering isn't there also a problem in the allocate action? After the break here:
> > https://github.com/volcano-sh/volcano/blob/5d66534726334f213134a3e9f729d864166ab2fb/pkg/scheduler/actions/allocate/allocate.go#L215
> > 
> > , the job is not pushed back, and the remaining pods in this job are not scheduled
> > cc @Monokaix @lowang-bh @hwdef
> 
> It is ok when minMember equals to total replicas. Otherwise, maybe it need to be considered as this:
> 
> https://github.com/volcano-sh/volcano/blob/a61fc90b8e72f8b688390d974f78c3292c041cfc/pkg/scheduler/actions/allocate/allocate.go#L224

We can seperate it to another pr.

JesseStutler:
> > > Great. But I'm wondering isn't there also a problem in the allocate action? After the break here:
> > > https://github.com/volcano-sh/volcano/blob/5d66534726334f213134a3e9f729d864166ab2fb/pkg/scheduler/actions/allocate/allocate.go#L215
> > > 
> > > , the job is not pushed back, and the remaining pods in this job are not scheduled
> > > cc @Monokaix @lowang-bh @hwdef
> > 
> > 
> > It is ok when minMember equals to total replicas. Otherwise, maybe it need to be considered as this:
> > https://github.com/volcano-sh/volcano/blob/a61fc90b8e72f8b688390d974f78c3292c041cfc/pkg/scheduler/actions/allocate/allocate.go#L224
> 
> We can seperate it to another pr.

I think using `NeedContinueAllocating` is a good choice, Could you also help to fix allocate action? @hansongChina

Monokaix:
> NeedContinueAllocating()

`allocate` is a more sensitive action, add this logic may cause some performance issue, maybe we should reconsider that case: )

hansongChina:
> Please squash to one commit.

OK

hansongChina:
> > > > Great. But I'm wondering isn't there also a problem in the allocate action? After the break here:
> > > > https://github.com/volcano-sh/volcano/blob/5d66534726334f213134a3e9f729d864166ab2fb/pkg/scheduler/actions/allocate/allocate.go#L215
> > > > 
> > > > , the job is not pushed back, and the remaining pods in this job are not scheduled
> > > > cc @Monokaix @lowang-bh @hwdef
> > > 
> > > 
> > > It is ok when minMember equals to total replicas. Otherwise, maybe it need to be considered as this:
> > > https://github.com/volcano-sh/volcano/blob/a61fc90b8e72f8b688390d974f78c3292c041cfc/pkg/scheduler/actions/allocate/allocate.go#L224
> > 
> > 
> > We can seperate it to another pr.
> 
> I think using `NeedContinueAllocating` is a good choice, Could you also help to fix allocate action? @hansongChina
Sure. Should we use the method "NeedContinueAllocating" to make a judgment once, or directly use "continue" to let all the tasks of this job attempt to be scheduled once?

Issue #3959, opened 2025-01-08
Ascend for Volcano get "Fail to load custom plugins: plugin: not implemented"


glmapper:
### Please describe your problem in detail

Use Volcano to schedule Huawei 910B series NPU cards and extend it by integrating Ascend plugins based on the open-source Volcano. Reference documentation: 
https://www.hiascend.com/document/detail/zh/mindx-dl/600/clusterscheduling/clusterschedulingig/clusterschedulingig/dlug_installation_023.html

Base Informations:
* k8s version 1.22.17
* volcano version 1.9.0
* ubuntu 20.04 jammy
* arrch64

then execute the following command:
```
kubectl apply -f installer/volcano-development.yaml 
```
I have got volcano-scheduler's POD with CrashLoopBackOff status, the pod's logs :

```
W0107 20:46:31.387250       1 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
E0107 20:46:31.387420       1 server.go:78] Fail to load custom plugins: plugin: not implemented
```

pod events：
```
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  45s               default-scheduler  Successfully assigned volcano-system/volcano-scheduler-7b9fb44659-m2scs to master-node-210-75
  Normal   Pulled     2s (x4 over 41s)  kubelet            Container image "volcanosh/vc-scheduler:v1.9.0" already present on machine
  Normal   Created    2s (x4 over 41s)  kubelet            Created container volcano-scheduler
  Normal   Started    2s (x4 over 41s)  kubelet            Started container volcano-scheduler
  Warning  BackOff    2s (x5 over 39s)  kubelet            Back-off restarting failed container
```

Since this [document](https://www.hiascend.com/document/detail/zh/mindx-dl/600/clusterscheduling/clusterschedulingig/clusterschedulingig/dlug_installation_023.html ) involves building an NPU scheduling plugin, and there is no relevant error message after following the operations in the official document, it is tentatively assumed that the building of the plugin is normal. During the troubleshooting process, I have read the relevant code logic in VOLCANO 1.9.0 regarding the loading of plugins when the scheduler starts. However, judging from the code (interpreting the code according to the structured programming mindset), it seems that "plugin: not implemented" is an error that is bound to occur (I'm not very familiar with the dynamic plugin mechanism of GO).
[...]

PR #4042, opened 2025-02-25
fix custom plugin doc and build script

Monokaix:
#### What type of PR is this?
/kind bug
/kind cleanup
/kind documentation
/area scheduling
/area deploy

When build custom plugin, we should also re-build base image with CGO enable and `musl-gcc`.

Fixes https://github.com/volcano-sh/volcano/issues/3959

```release-note

```


Monokaix:
cc @hwdef @lowang-bh @archlitchi @JesseStutler

lowang-bh (commented):
/lgtm

Makefile
```diff
@@ -73,7 +74,7 @@ init:
 
 vc-scheduler: init
 	if [ ${SUPPORT_PLUGINS} = "yes" ];then\
-		CC=${CC} CGO_ENABLED=1 go build -ldflags ${LD_FLAGS} -o ${BIN_DIR}/vc-scheduler ./cmd/scheduler;\
+		CC=${MUSL_CC} CGO_ENABLED=1 go build -ldflags ${LD_FLAGS_CGO} -o ${BIN_DIR}/vc-scheduler ./cmd/scheduler;\
```
hwdef, line 77: musl-gcc is used by default? I'm not sure if this is a good idea.

If this modification is necessary, we should first check whether musl-gcc is installed and give a prompt

Monokaix: Yeah when we use CGO and custom plugin, musl-gcc is required, so I have add the recommended way that built with docker in `Build the image and plugin in `Docker`(Recommended):`.

hwdef (commented):
/lgtm

example/custom-plugin/Dockerfile
```diff
@@ -1,2 +1,32 @@
...
\ No newline at end of file
+FROM golang:1.22.2 AS builder
+
+WORKDIR /go/src/volcano.sh/
+
+# Install musl
+RUN apt-get update && \
+    apt-get install -y sudo
+
+RUN wget http://musl.libc.org/releases/musl-1.2.1.tar.gz && \
```
hwdef, line 9:
```suggestion
RUN wget http://musl.libc.org/releases/musl-latest.tar.gz && \
```

Monokaix: done

PR #4045, opened 2025-02-26
Refactor: move DeviceName const into its own package

SataQiu:
#### What type of PR is this?
/kind cleanup

Refactor: move DeviceName const into its own package

Fixes #

```release-note
NONE
```


pkg/scheduler/api/node_info.go
```diff
@@ -337,10 +337,10 @@ func (ni *NodeInfo) setNodeOthersResource(node *v1.Node) {
 		return
 	}
 
-	ni.Others[GPUSharingDevice] = gpushare.NewGPUDevices(ni.Name, node)
```
Monokaix, line 340: Everytime we add a new device, the framework should be aware of that and modify here, IMO it's not a good way, maybe we can let every device register some callback and the framework just call a basic interface or something is ok.
But it's a seperate topic, fire a new PR if you're willing to do that: )

Issue #4046, opened 2025-02-26
Add resync-period flag for k8s native informers


sfc-gh-raravena:
### What is the problem you're trying to solve

This is to optionally help with cache inconsistencies

### Describe the solution you'd like

Have an optional flag for cache resync period for the volcano scheduler.

PR #4047, opened 2025-02-26
Add resync-period flag for k8s native informers

sfc-gh-raravena:
This is to optionally help with cache inconsistencies

#### What type of PR is this?

/kind feature
- Add resync-period flag for k8s native informers
- This is to optionally help with cache inconsistencies

Fixes # https://github.com/volcano-sh/volcano/issues/4046

```release-note

```

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -32,6 +32,7 @@ import (
 const (
 	defaultSchedulerName   = "volcano"
 	defaultSchedulerPeriod = time.Second
+	defaultResyncPeriod    = 0
 	defaultQueue           = "default"
 	defaultListenAddress   = ":8080"
 	defaultHealthzAddress  = ":11251"
@@ -60,6 +61,7 @@ type ServerOption struct {
 	SchedulerNames    []string
 	SchedulerConf     string
 	SchedulePeriod    time.Duration
+	ResyncPeriod      time.Duration
 	// leaderElection defines the configuration of leader election.
 	LeaderElection config.LeaderElectionConfiguration
 	// Deprecated: use ResourceNamespace instead.
@@ -119,6 +121,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.StringArrayVar(&s.SchedulerNames, "scheduler-name", []string{defaultSchedulerName}, "vc-scheduler will handle pods whose .spec.SchedulerName is same as scheduler-name")
 	fs.StringVar(&s.SchedulerConf, "scheduler-conf", "", "The absolute path of scheduler configuration file")
 	fs.DurationVar(&s.SchedulePeriod, "schedule-period", defaultSchedulerPeriod, "The period between each scheduling cycle")
+	fs.DurationVar(&s.ResyncPeriod, "resync-period", defaultResyncPeriod, "The default resync period for k8s native informer factory")
 	fs.StringVar(&s.DefaultQueue, "default-queue", defaultQueue, "The default queue name of the job")
 	fs.BoolVar(&s.PrintVersion, "version", false, "Show version and quit")
 	fs.StringVar(&s.ListenAddress, "listen-address", defaultListenAddress, "The address to listen on for HTTP requests.")
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -92,8 +92,8 @@ func init() {
 }
 
 // New returns a Cache implementation.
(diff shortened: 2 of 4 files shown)


cmd/scheduler/app/options/options.go
```diff
@@ -32,6 +32,7 @@ import (
 const (
 	defaultSchedulerName   = "volcano"
 	defaultSchedulerPeriod = time.Second
+	defaultResyncPeriod    = 5 * time.Minute
```
Monokaix, line 35: Can you give more details about why we need this resync perion, as I didn't see that the resync is set in upstream kubernetes components: )

pkg/scheduler/cache/cache.go
```diff
@@ -633,7 +634,7 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, defaultQueu
 }
 
 func (sc *SchedulerCache) addEventHandler() {
-	informerFactory := informers.NewSharedInformerFactory(sc.kubeClient, 0)
+	informerFactory := informers.NewSharedInformerFactory(sc.kubeClient, sc.resyncPeriod)
```
lowang-bh, line 637: sc.resyncPeriod is not set.

cmd/scheduler/app/options/options.go
lowang-bh: It is too small and will cause scheduler performance issue.

Monokaix: I think we not disable it (set to zero) by default and users can set a positive value as needed.

pkg/scheduler/cache/cache.go
sfc-gh-raravena: isn't it set in line 70 of `pkg/scheduler/scheduler.go` ?

Monokaix: We should add `sc.resyncPeriod = resyncPeriod` in `newSchedulerCache` func: )

Monokaix: And also I think disable resync by default is better: )

sfc-gh-raravena: Added `sc.resyncPeriod = resyncPeriod`. Where do I disable the resync? Thx

Monokaix: Just set the `resync-period`flag's value to 0 is ok.

sfc-gh-raravena: set it to `0`

sfc-gh-raravena:
Hello following up with this PR.  Anything else we can do to get it merged?  Thanks!

Monokaix:
> Hello following up with this PR.  Anything else we can do to get it merged?  Thanks!

I think we can change 5*time.minutes to 0 to disable it by default，and users can set a non-zero value to enable it.

Monokaix:
Thanks! @sfc-gh-raravena. It's overall OK, please merge to one commit: ）

sfc-gh-raravena:
> Is it just adding a resyncPeriod field to the cache but has no practical effect? ​​I don't seem to see resyncPeriod being actually used

Maybe we are missing the bits where it needs to take effect.  Can you help identify it?  Thx!

Monokaix:
> > Is it just adding a resyncPeriod field to the cache but has no practical effect? ​​I don't seem to see resyncPeriod being actually used
> 
> Maybe we are missing the bits where it needs to take effect. Can you help identify it? Thx!

We can add it here: https://github.com/volcano-sh/volcano/blob/03bb7e28bcb59694aa223df815c8a8a8c748964b/pkg/scheduler/cache/cache.go#L640

Monokaix:
> > > Is it just adding a resyncPeriod field to the cache but has no practical effect? ​​I don't seem to see resyncPeriod being actually used
> > 
> > 
> > Maybe we are missing the bits where it needs to take effect. Can you help identify it? Thx!
> 
> We can add it here:
> 
> https://github.com/volcano-sh/volcano/blob/03bb7e28bcb59694aa223df815c8a8a8c748964b/pkg/scheduler/cache/cache.go#L640

We can fire a another pr to fix this.

JesseStutler:
/lgtm

Merge first, should raise another PR like this @sfc-gh-raravena https://github.com/volcano-sh/volcano/pull/4047#issuecomment-2903004334, to actually use this

sfc-gh-raravena:
> /lgtm
> 
> Merge first, should raise another PR like this @sfc-gh-raravena [#4047 (comment)](https://github.com/volcano-sh/volcano/pull/4047#issuecomment-2903004334), to actually use this

I'm confused isn't it already here: https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/cache/cache.go#L572 ?

PR #4052, opened 2025-02-28
add queue-level-scheduling-policy.md

ElectricFish7:
#### What type of PR is this?

/kind documentation
/area scheduling
/area controllers

Queue-level-scheduling-policy from a LFX'25 issue:https://github.com/volcano-sh/volcano/issues/3992, which requires the volcano to support setting and using different scheduling policies at the queue level instead of using a globally unified scheduling policy.

Fixes https://github.com/volcano-sh/volcano/issues/3992

@Monokaix
@JesseStutler 

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,10 @@ If you don't have a kubernetes cluster, try one-click install from code base:
 
 This way is only available for x86_64 temporarily.
 
+### Install volcano agent
+
+Please follow the guide [Volcano Agent](https://volcano.sh/en/docs/colocation) to install volcano agent.
+
 ### Install monitoring system
 
 If you want to get prometheus and grafana volcano dashboard after volcano installed, try following commands:
@@ -196,4 +200,4 @@ If you have any question, feel free to reach out to us in the following ways:
 
 [Mailing List](https://groups.google.com/forum/#!forum/volcano-sh)
 
-Wechat: Add WeChat account `k8s2222` (华为云小助手2号) to let her pull you into the group.
+WeChat: Please add WeChat account `k8s2222` and request an invitation to the group chat.


JesseStutler:
Great! We can arrange a meeting to discuss this

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+## 1.3、Extend the Session structure
+
+Add a **QueueContexts** field in the **Session** structure, allowing the corresponding QueueContext to be found by QueueID. The purpose of this is that the Session, as the start of a scheduling cycle, needs to store the information of the existing queues and the queue-specific plugin configurations.
+
+```go
+type Session struct {
+	...
+    Queues         map[api.QueueID]*api.QueueInfo
```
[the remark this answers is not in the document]
ElectricFish7: Thanks for raising this question! In my opinion, there are two main reasons.

1. **Lifecycle Differences**:  
   - `QueueInfo` persists across multiple scheduling cycles.
   - `QueueContext` is **per-scheduling-cycle** scoped (like the `Session` struct). It gets recreated in every scheduling cycle, allowing queue scheduling policies to be refreshed dynamically. This means if a queue's scheduling policies change, we don’t need to modify the persistent `QueueInfo`—just let the next cycle’s `QueueContext` pick up the new configuration.  

2. **On-Demand Initialization**:  
   Not all queues require custom scheduling policies. We only create a `QueueContext` for queues that explicitly define scheduling strategies, while queues without specific policies inherit global/default configurations. This avoids unnecessary overhead for simple use cases.

ElectricFish7:
> Great! We can arrange a meeting to discuss this

Sure! Let’s set up a time that works for both of us.

JesseStutler:
Is this email: [wuyuqi22@mails.ucas.ac.cn](mailto:wuyuqi22@mails.ucas.ac.cn) yours? We can add wechat first

ElectricFish7:
I have modified the plan. Please review it.

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,289 @@
...
+
+## 1.2、Add the QueueContext structure
+
+Add the **QueueContext** structure to store the queue-specific plugin configuration. To avoid major changes, directly introduce the relevant functions from the session structure.
+
+```go
+type QueueContext struct {
+	Plugins         []Plugin      
+
+	jobOrderFns         map[string]api.CompareFn
```
JesseStutler, line 67: Is it necessary to list all the extension points like session? Can we categorize them and wrap them into a struct?

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,289 @@
...
+			klog.V(5).Infof("Added Queue <%s> for Job <%s/%s>",
+				queue.Name, job.Namespace, job.Name)
+
+			queueSet.Insert(string(queue.UID))
+			queues.Push(queue)
+		}
+
+		if job.IsPending() {
+			if _, found := jobsMap[job.Queue]; !found {
+				jobsMap[job.Queue] = util.NewPriorityQueue(ssn.JobOrderFn)
```
JesseStutler, line 229: After you register QueueContext, how do you actually use it? I don't seem to see it. E.g., here it's still `ssn.JobOrderFn`.

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,289 @@
...
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: volcano-queue-policies
+  namespace: volcano-system
+data:
+  # Queue-level policy (Key is the queue name).
+  ai-training-queue: |
```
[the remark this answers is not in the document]
ElectricFish7: I think the queue configuration can override the global configuration.

docs/design/queue-level-scheduling-policy.md
ElectricFish7: Of course, but copying directly from the session ensures consistency in usage.

docs/design/queue-level-scheduling-policy.md
JesseStutler: OK. So you also need to show how to use queue context's extension points in this design doc

ElectricFish7:
Based on yesterday’s meeting, I have made modifications to the proposed plan.@JesseStutler

PR #4066, opened 2025-03-07
refactor: Optimized code

### LINKED ISSUE #4065: Refactor code

The code here can be optimized
```go
	if !found {
		return false
	}

	if jobInfo.Job == nil {
		return false
	}

``` 

i think is
```go
	if !found || jobInfo.Job == nil {
		return false
	}
``` 

### Additional context

_No response_

feyounger:
The two judgments can be merged

Fixes #4065

Fixes #

```release-note

```


hwdef (commented):
/ok-to-test
/lgtm

PR #4077, opened 2025-03-10
CI: add dependabot 

dongjiang1989:

Fixes #

ref: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference
1. add gomod golang package auto upgrade
2. add github action auto upgrade

```release-note
None
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,45 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: daily
+    groups:
+      k8s-libs:
+        patterns:
+          - "k8s.io/api"
+          - "k8s.io/apiextensions-apiserver"
+          - "k8s.io/apimachinery"
+          - "k8s.io/apiserver"
+          - "k8s.io/cli-runtime"
+          - "k8s.io/client-go"
+          - "k8s.io/cloud-provider"
+          - "k8s.io/cluster-bootstrap"
+          - "k8s.io/code-generator"
+          - "k8s.io/component-base"
+          - "k8s.io/component-helpers"
+          - "k8s.io/controller-manager"
+          - "k8s.io/cri-api"
+          - "k8s.io/cri-client"
+          - "k8s.io/csi-translation-lib"
+          - "k8s.io/dynamic-resource-allocation"
+          - "k8s.io/endpointslice"
+          - "k8s.io/kube-aggregator"
+          - "k8s.io/kube-controller-manager"
+          - "k8s.io/kube-proxy"
+          - "k8s.io/kube-scheduler"
+          - "k8s.io/kubectl"
+          - "k8s.io/kubelet"
+          - "k8s.io/legacy-cloud-providers"
+          - "k8s.io/metrics"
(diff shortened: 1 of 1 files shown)


Monokaix:
Does the gihubaction will auto submit a pr when go mod has a new version?

dongjiang1989:
> Does the gihubaction will auto submit a pr when go mod has a new version?

Yes. Including version upgrades of dependent libraries in go mod and version upgrades of various actions in github action

Monokaix:
> > Does the gihubaction will auto submit a pr when go mod has a new version?
> 
> Yes. Including version upgrades of dependent libraries in go mod and version upgrades of various actions in github action

That's great.

.github/dependabot.yml
```diff
@@ -0,0 +1,45 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: daily
+    groups:
+      k8s-libs:
+        patterns:
+          - "k8s.io/api"
```
Monokaix, line 10: How about just concern k8s.io/kubernetes, or else there will be a lot of PRs.

dongjiang1989: K8s.io/* package has undergone many version changes, so adding them to a group will update them as one PR.

Monokaix: There is also a question that there maybe some conflicts when dependBot automatically submit PRs because the new version may change the interface or arguments, so maybe manually update version is better?

dongjiang1989: Your concerns are reasonable. But most minor version updates are online compatible and will be merged only after passing all GitHub action automatic tests.

ref:
1. Group update pr case: https://github.com/prometheus-operator/prometheus-operator/pull/7405
2. Minor version updated case: https://github.com/prometheus-operator/prometheus-operator/pull/7401

Monokaix: That's great！Minor version ok, what about major verion change?

dongjiang1989: If the major version is backward compatible, it can also be merged automatically. 

If there is a conflict (maybe change the interface or arguments), you can reply to this PR `@dependabot ignore this major version` and handle it manually.

case: https://github.com/prometheus-operator/prometheus-operator/pull/7369#issuecomment-2713114432

Monokaix: great!

dongjiang1989:
> /approve

Thanks @Monokaix

dongjiang1989:
image
When I git rebase, tags add by volcano-sh-bot.
@Monokaix  Maybe some bug?

.github/dependabot.yml
```diff
@@ -0,0 +1,45 @@
...
+          - "k8s.io/node-api"
+          - "k8s.io/pod-security-admission"
+          - "k8s.io/sample-apiserver"
+          - "k8s.io/sample-cli-plugin"
+          - "k8s.io/sample-controller"
+
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
```
hwdef, line 45: Daily is too frequent, recommend weekly

PR #4079, opened 2025-03-11
scheduler: consider the nominated node first when allocating Node for Pod

SataQiu:
#### What type of PR is this?
/kind feature

"NominatedNodeName" can potentially be set in a previous scheduling cycle as a result of preemption.
This node is likely the only candidate that will fit the pod, and hence we try it first before iterating over all nodes.

Ref #3491

We already set the `status.nominatedNodeName` field for Pod here:
https://github.com/volcano-sh/volcano/blob/79a8e41ebdb04b6e218deee04366f77b0a893ac1/pkg/scheduler/cache/cache.go#L1046-L1049

```release-note
scheduler: consider the nominated node first when allocating Node for Pod
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -215,7 +215,22 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 			break
 		}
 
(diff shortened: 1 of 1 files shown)


SataQiu:
cc @lowang-bh

JesseStutler:
Great job! Now we are consistent with kube-scheduler
/lgtm

PR #4088, opened 2025-03-13
[Refactor] controller cache deletePod logic,skip create

### LINKED ISSUE #4064: controller cache deletePod

Why can't I find a job in deletepod and still need to create one? I think we can directly skip.
Location of code is https://github.com/volcano-sh/volcano/tree/master/pkg/controllers/cache/cache.go
```go
	job, found := jc.jobs[key]
	if !found {
		job = &apis.JobInfo{
			Pods: make(map[string]map[string]*v1.Pod),
		}
		jc.jobs[key] = job
	}

``` 

i think  so
```go
if job, found := jc.jobs[key]; found {
		if err := job.DeletePod(pod); err != nil {
			return err
		}
		if jobTerminated(job) {
			jc.deleteJob(job)
		}
	}
```

feyounger:

Fixes #4064

```release-note

```

### DIFF
--- a/pkg/controllers/cache/cache.go
+++ b/pkg/controllers/cache/cache.go
@@ -237,20 +237,13 @@ func (jc *jobCache) DeletePod(pod *v1.Pod) error {
 		return err
 	}
 
-	job, found := jc.jobs[key]
-	if !found {
-		job = &apis.JobInfo{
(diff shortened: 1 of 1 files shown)


hwdef (commented):
/approve

PR #4106, opened 2025-03-19
feat: Optimize sort pods in rescheduling

dongjiang1989:
#### What type of PR is this?

1. First determine the pod priority
2. Then determine the pod Qos
3. Finally determine the pod create time

Fixes #

```release-note
optimize sort pods in rescheduling
```


pkg/scheduler/plugins/rescheduling/node_utilization_util.go
```diff
@@ -139,20 +139,27 @@ func getScoreForNode(index int, nodeUtilizationList []*NodeUtilization) float64
...
+				if v1qos.GetPodQOS(pods[i]) == v1.PodQOSBestEffort {
+					return true
+				}
+				if v1qos.GetPodQOS(pods[i]) == v1.PodQOSBurstable && v1qos.GetPodQOS(pods[j]) == v1.PodQOSGuaranteed {
+					return true
+				}
+				return false
+			} else {
+				// sort by pod create time
+				return pods[i].GetCreationTimestamp().Time.After(pods[j].GetCreationTimestamp().Time)
```
Monokaix, line 161: Why compare creationtime?

dongjiang1989: When the QoS is the same, newly created pods are evicted first (older and more stable pods are not evicted first)

shinytang6 (commented):
code LGTM, it's better to fix the comments in line100 :)

dongjiang1989:
> code LGTM, it's better to fix the comments in line100 :)

Thanks @shinytang6 
Fixed

Monokaix:
It's better to add UT: )

dongjiang1989:
> It's better to add UT: )

Fixed

pkg/scheduler/plugins/rescheduling/node_utilization_util_test.go
```diff
@@ -0,0 +1,441 @@
+/*
```
Monokaix, line 1: There are some pod related utils in ``pkg/scheduler/util/test_util.go`, we can reuse them: )

dongjiang1989: Got it

pkg/scheduler/plugins/rescheduling/node_utilization_util.go
```diff
@@ -139,20 +140,27 @@ func getScoreForNode(index int, nodeUtilizationList []*NodeUtilization) float64
...
 			return true
 		}
 		if pods[j].Spec.Priority == nil && pods[i].Spec.Priority != nil {
 			return false
 		}
+		// sort by pod Qos
 		if (pods[j].Spec.Priority == nil && pods[i].Spec.Priority == nil) || (*pods[i].Spec.Priority == *pods[j].Spec.Priority) {
-			if v1qos.GetPodQOS(pods[i]) == v1.PodQOSBestEffort {
-				return true
+			if v1qos.GetPodQOS(pods[i]) != v1qos.GetPodQOS(pods[j]) {
```
Thor-wl, line 152: Please simplify this piece of code for the following reasons:
1. It's hard to get the points for too many levels of if judgements.
2. if there is just a return with true or false, I think it is not necessary to write like the above, just return the results of the judgements.

PR #1576, opened 2021-07-01
support multi-scheduler

huone1:
Signed-off-by: huone1 <huwanxing@huawei.com>
issue #1322

PR #4120, opened 2025-03-20
Remove pod mutate webhook by default

Monokaix:
#### What type of PR is this?
/area performance
Remove pod mutate webhook by default. The pod mutating webhook is only used in multi scheduler scenario, it's a normal case and not recommended, we should disable it by default and users can enable it when needed.
This can improve performance in large scale case.

Fixes #

multi scheduler issue: https://github.com/volcano-sh/volcano/pull/1576
Part of performance improvement issus: https://github.com/volcano-sh/volcano/issues/3852

```release-note
The pod mutating webhook is disabled by default to Improve performance
```


hwdef (commented):
/lgtm

PR #4123, opened 2025-03-20
Enabled Cooldown Protection Plugin for reclaiming also

qGentry:
#### What type of PR is this?
Feature

I've noticed that cooldown protection plugin is not working for queue reclaiming (via the capacity plugin). This PR also registers cdp's preemtableFn as a reclaimableFn so cdp also works for reclaiming.

Yes
```release-note
- support cooldown protection plugin for queue reclaiming via capacity plugin
```

### DIFF
--- a/pkg/scheduler/plugins/cdp/cdp.go
+++ b/pkg/scheduler/plugins/cdp/cdp.go
@@ -72,15 +72,15 @@ func (sp *CooldownProtectionPlugin) podCooldownTime(pod *v1.Pod) (value time.Dur
 
 // OnSessionOpen implements framework.Plugin
 func (sp *CooldownProtectionPlugin) OnSessionOpen(ssn *framework.Session) {
-	preemptableFn := func(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) ([]*api.TaskInfo, int) {
+	filterVictimFn := func(evictingTask *api.TaskInfo, candidateVictims []*api.TaskInfo) ([]*api.TaskInfo, int) {
 		var victims []*api.TaskInfo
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/cdp/cdp.go
```diff
@@ -72,15 +72,15 @@ func (sp *CooldownProtectionPlugin) podCooldownTime(pod *v1.Pod) (value time.Dur
 
 // OnSessionOpen implements framework.Plugin
 func (sp *CooldownProtectionPlugin) OnSessionOpen(ssn *framework.Session) {
-	preemptableFn := func(preemptor *api.TaskInfo, preemptees []*api.TaskInfo) ([]*api.TaskInfo, int) {
+	preemptableReclaimableFn := func(preemptor_reclaimer *api.TaskInfo, preemptees_reclaimees []*api.TaskInfo) ([]*api.TaskInfo, int) {
```
JesseStutler, line 75: Go does not use underscores in variable naming but uses camel case instead. However, to avoid ambiguity, I recommend changing it like this:
```
filterVictimsFn := fun(evictingTask *api.TaskInfo, candidateVictims []*api.TaskInfo) ([]*api.TaskInfo, int) 
```

JesseStutler:
Please also use `git commit -s` to sign your commit

pkg/scheduler/plugins/cdp/cdp.go
qGentry: Thanks. Fixed

shinytang6 (commented):
lgtm too

hwdef (commented):
/approve

PR #4124, opened 2025-03-20
fix: remove hostpath mount in volcano-scheduler

### LINKED ISSUE #3298: Replace `hostPath` with `emptyDir`or PVC in Scheduler Deployment

cnmcavoy:
#### What type of PR is this?

Bugfix

`hostPath` should be avoided unless absolutely necessary:
https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

Fixes #3298

Previous PR: https://github.com/volcano-sh/volcano/pull/3303 which this supersedes 

```release-note
fix: remove hostpath mount in volcano-scheduler
```

### DIFF
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -4661,8 +4661,7 @@ spec:
           configMap:
             name: volcano-scheduler-configmap
         - name: klog-sock
-          hostPath:
-            path: /tmp/klog-socks
+          emptyDir: {}
 ---
 # Source: volcano/templates/scheduling_v1beta1_podgroup.yaml
 apiVersion: apiextensions.k8s.io/v1
--- a/installer/helm/chart/volcano/templates/scheduler.yaml
+++ b/installer/helm/chart/volcano/templates/scheduler.yaml
@@ -224,8 +224,7 @@ spec:
           configMap:
             name: {{ .Release.Name }}-scheduler-configmap
         - name: klog-sock
-          hostPath:
-            path: /tmp/klog-socks
+          emptyDir: {}
 ---
 apiVersion: v1
 kind: Service


hwdef:
I have objected to the merge of [this code](https://github.com/volcano-sh/volcano/pull/3146) from the beginning.

/ok-to-test
/lgtm

installer/volcano-development.yaml
```diff
@@ -4661,8 +4661,7 @@ spec:
           configMap:
             name: volcano-scheduler-configmap
         - name: klog-sock
```
Monokaix, line 4663:
```
          env:
            - name: DEBUG_SOCKET_DIR
              value: /tmp/klog-socks
```
The scheduler deployment will read this path `/tmp/klog-socks` by env? Does it work after this pr?

cnmcavoy: The mount path inside of the container has not been changed, the emptydir will be mounted at `/tmp/klog-socks`.

Monokaix:
/approve
Thanks！

Issue #4127, opened 2025-03-21
Propose to remove unnecessary webhooks and not register by default for performance consideration


Monokaix:
### What is the problem you're trying to solve

Modify and disable some webhooks to Improve performance.

### Describe the solution you'd like

This issue is part of performance improvement issue https://github.com/volcano-sh/volcano/issues/3852, our main consideration is that the webhooks can greatly affect performance in large scale scenario, so we should avoid use webhooks as little as possible. And if users want to use specific webhooks, they should enable them when install volcano, that's to say, users should balance functionality and performance.
Here are the tasks we need to do and reasons
- [x] Disable pod mutating webhook by default, becasue it's rarely used, see https://github.com/volcano-sh/volcano/pull/4120
- [ ] Pod Validation webhook: We should not check the pod related podgroup as the pod and podgroup's creation order is not certain, and also we should move the queue open `status` check to podgroup validation, as volcano concerns the PodGroup level is enough when scheduling, and if pods come after a PodGroup's queue has been closed, we can consider to add a new state at PodGroup to block the scheduling, this needs more users feedback. And also disable it by default.
- [ ] Disable PodGroup mutating webhook by default becaue it's not a common case to inherit queue name from namespace's annotation, and also the `default` queue can be add through kubebuilder semantic.
- [ ] Job mutating webhook: replace patch with kubebuilder supported semantic.

Last but not least: once the [Mutating Admission Policy](https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/) is beta, we should replace all mutating webhooks with it.

hwdef:
I think this is good, we can target completely removing the webhook. We can set the complete removal of webhooks as the goal. 

However, we should improve the code of the controller and scheduler while removing the webhook to ensure they do not panic.

avinxshKD:
> Reduce unnecessary webhooks and avoid registering them by default. Some webhook logic has been migrated, and we are trying to reduce the amount of logic in webhooks as much as possible, replacing it with VAP/MAP and KubeBuilder annotations. However, some logic still exists in the webbook, requiring analysis of the more time-consuming stages in the webhook. For example, we need to combine pprof and analyze the impact of webhooks on performance in batch creation and scheduling scenarios.

Hey @JesseStutler  would like to help with the remaining Job mutating webhook task.

[...]

Issue #2276, opened 2022-06-04
Pod starts running when Queue state is closed


Sharathmk99:
**What happened**:
When I close the queue using `vcctl` cli and submit the `pod`, it starts running.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
Closed `default` queue
`vcctl queue operate -a close -n default`

Submited below pod,
```
apiVersion: v1
kind: Pod
metadata:
  name: test
  annotations:
    scheduling.volcano.sh/queue-name: default
spec:
  schedulerName: volcano
  restartPolicy: Never
  containers:
  - name: bash
    image: python:3.7
    imagePullPolicy: IfNotPresent
    command: ["bash", "-c", "echo 'sleep...'; sleep 600"]
    resources:
      requests:
        memory: "64Mi"
        cpu: "1000m"
```

**Anything else we need to know?**:
Currently `job` validation webhook blocks the job with queue state is closed. But `pod` validation doesn't block the same.

**Environment**:
- Volcano Version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Sharathmk99:
I have a fix for the same. Shall I open the PR?
https://github.com/volcano-sh/volcano/compare/master...Sharathmk99:bugfix/queue-closed-state?expand=1

Thor-wl:
Thanks for your contribution, @Sharathmk99 . Please submit the PR. That's OK.

PR #4132, opened 2025-03-21
Move queue state validate from pod to podgroup

Monokaix:
#### What type of PR is this?
Part of https://github.com/volcano-sh/volcano/issues/4127
Currently when pod is submitted, webhook will check the podGroup and queue state of the pod,
1. PodGroup and pod creation is not certain, so it's unnecessary to check the pg's status, if podGroup not exits, pod will not be scheduled, which meets the need.
2. Volcno schedule based on podGroup, so it' more reasonable to check podGroup's queue state instead of single pod, the pod-level validation originally came from this issue https://github.com/volcano-sh/volcano/issues/2276. Users need to be aware that this validation has been moved to the PodGroup. This needs to be noted during the upgrade, and we need to explain it in the changelog.

Fixes #

```release-note
The queue status webhook validation when submitting a pod has been moved from the Pod  to the PodGroup.
```


lowang-bh (commented):
/lgtm

kevin-wangzefeng (commented):
/approve

PR #4133, opened 2025-03-21
Move lastEvictAt to tdmPlugin

ElectricFish7:
#### What type of PR is this?
/kind cleanup
Currently, lastEvictAt is a package-level variable, which makes sense when there is only one plugin configuration. However, when multiple plugin configurations are introduced later, they will share this variable, which is not appropriate. Therefore, we are moving lastEvictAt to tdmPlugin.

Fixes #

@Monokaix
@JesseStutler

```release-note

```

### DIFF
--- a/pkg/scheduler/plugins/tdm/tdm.go
+++ b/pkg/scheduler/plugins/tdm/tdm.go
@@ -41,8 +41,6 @@ const (
 	defaultPodEvictNum       = 1
 )
 
-var lastEvictAt time.Time
-
 /*
    actions: "enqueue, reclaim, allocate, preempt"
    tiers:
@@ -59,6 +57,7 @@ type tdmPlugin struct {
 	// evictPeriod
 	// default 1m
 	evictPeriod time.Duration
+	lastEvictAt time.Time
 }
 
 // New function returns prioritizePlugin object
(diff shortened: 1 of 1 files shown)


hwdef (commented):
/lgtm
/ok-to-test

JesseStutler:
I don't think the tdm plugin needs the period to cool down, the lifecycle of this lastEvict should be consistent with the scheduler, not the session, as it will span across sessions. Therefore, I don't think this is a good design. Moreover, this victimTasks extension point is also quite strange, like leftover code. @Monokaix

ElectricFish7:
> I don't think the tdm plugin needs the period to cool down, the lifecycle of this lastEvict should be consistent with the scheduler, not the session, as it will span across sessions. Therefore, I don't think this is a good design. Moreover, this victimTasks extension point is also quite strange, like leftover code. 我认为 tdm 插件不需要冷却期，这个 lastEvict 的生命周期应该与调度器保持一致，而不是与会话一致，因为它将跨越会话。因此，我认为这不是一个好的设计。此外，这个 victimTasks 扩展点也很奇怪，就像遗留代码一样。@Monokaix

So, do we not need this lastEvictAt?

PR #4135, opened 2025-03-21
fix: Fix jobflow status  from `running` to `failed` FSM

### LINKED ISSUE #4134: jobflow status error, when vcjob status `failed`
### Description

```bash
apiVersion: flow.volcano.sh/v1alpha1
kind: JobFlow
metadata:
  creationTimestamp: "2025-03-21T03:25:03Z"
  generation: 1
  name: test123-ee5c5a8bdc
  namespace: sae-ai-156
  ownerReferences:
  - apiVersion: [sae.cloud.iflytek.com/v1](http://sae.cloud.iflytek.com/v1)
    blockOwnerDeletion: true
    controller: true
    kind: JobFlowInstance
    name: test123-ee5c5a8bdc
    uid: 2f9e2a09-2668-4fe6-b8a2-a24a6f9dfcc5
  resourceVersion: "896120155"
  uid: de0b6922-9f97-4e0f-829e-f28485907507
spec:
  flows:
  - name: demo
  - dependsOn:
      targets:
      - demo
    name: sss
  jobRetainPolicy: delete
status:
  conditions:
    test123-ee5c5a8bdc-demo:
      createTime: "2025-03-21T03:25:03Z"
      phase: Failed
      runningDuration: 2m56s
  failedJobs:
  - test123-ee5c5a8bdc-demo   # vcjob status failed
  jobStatusList:
  - endTimestamp: "2025-03-21T03:27:59Z"
    name: test123-ee5c5a8bdc-demo
    runningHistories:
    - endTimestamp: "2025-03-21T03:27:59Z"
      startTimestamp: "2025-03-21T03:25:03Z"
      state: Pending
    - endTimestamp: "2025-03-21T03:27:59Z"
      startTimestamp: "2025-03-21T03:27:59Z"
      state: Running
    - startTimestamp: "2025-03-21T03:27:59Z"
      state: Failed
    startTimestamp: "2025-03-21T03:25:03Z"
    state: Failed
  state:
    phase: Running # jobflow status not change to failed
```

1.
2.
3.

Change `status.phase` to `Failed`

v1.11.0 and v1.10.0

_No response_

dongjiang1989:
#### What type of PR is this?

Bugfix

1. Fix the FSM of status in jobflow status: `running` -> `failed`
2. Add jobflow unittest case

Fixes #4134 

need cherry pick this PR to v1.10 and v1.11

cc @Monokaix  @hwdef 

```release-note
Fix jobflow status  from `running` to `failed` FSM
```


hwdef:
what is `FSM` ?

pkg/controllers/jobflow/state/running.go
```diff
@@ -28,6 +28,8 @@ func (p *runningState) Execute(action v1alpha1.Action) error {
 		return SyncJobFlow(p.jobFlow, func(status *v1alpha1.JobFlowStatus, allJobList int) {
 			if len(status.CompletedJobs) == allJobList {
 				status.State.Phase = v1alpha1.Succeed
+			} else if len(status.FailedJobs) > 0 {
```
hwdef, line 31: If there is a job's state failed, then the jobflow's state is failed? It feels too extreme

dongjiang1989:
> what is `FSM` ?

Jobflow FSM.
The entire state changes：
https://github.com/volcano-sh/volcano/blob/9a374332b88500321a5f2118abc794cd1e657a81/pkg/controllers/jobflow/jobflow_controller_action_test.go#L252-L260

pkg/controllers/jobflow/state/running.go
dongjiang1989: vcjob faild is the final state after multiple retries. Its failure should be reflected in the final state of the jobflow.
Currently, vcjob faild will hinder the execution of dependent vcjobs and will remain in the running state for a long time.

hwdef:
I mean the abbreviation of what is FSM?

pkg/controllers/jobflow/state/running.go
hwdef: If we implement `if` logic for jobflow later. The failed vcjob can also trigger the next vcjob to run.

if complete:
   run job A
else if failed:
   run job B

dongjiang1989:
> I mean the abbreviation of what is FSM?

Finite State Machine

pkg/controllers/jobflow/state/running.go
hwdef: Perhaps we can use the status of the leaf node to represent the status of the jobflow

image

hwdef: image

dongjiang1989: Great. 
First, `jobflow.Spec.Flow` needs to belong to DAG

When we add conditional judgment, we can add `IsConditionFail(status.FailedJobs, jobs.Depends) bool` about：Check the leaf Node(Make sure there is a Failed status processing branch)

dongjiang1989:
@hwdef  Merge this PR first and cherry pick this PR to v1.10 and v1.11?

hwdef:
Without the `if` function, it is feasible to determine the state of a jobflow based on leaf nodes(or final vcjob), rather than by number.

dongjiang1989:
> Without the `if` function, it is feasible to determine the state of a jobflow based on leaf nodes(or final vcjob), rather than by failed job numbers.

Currently, vcjob faild on non-leaf nodes will hinder the execution of dependent vcjobs, and the jobflow will be in a running state for a long time.
ref: https://github.com/volcano-sh/volcano/issues/4134

hwdef:
We should use a compatible method to get it done in one step(Determine the status of the leaf node), rather than modifying it again and again.

dongjiang1989:
> We should use a compatible method to get it done in one step(Determine the status of the leaf node), rather than modifying it again and again.

I agree.
Regarding this problem, how to change the running state machine -> failed state machine when the if condition judgment and state judgment are not implemented (the current main code can only keep Running and Completed after all successes)

Do you have any way to solve the bug?

hwdef:
Let me get my thoughts straight:

We should not care about the state of all jobs in jobflow, we should only care about the state of the last task of all branches. This is what I mentioned in the figure above, so we should not use len(failedjob) to define the state of the jobflow. This has nothing to do with `if` function.

hwdef:
image

hwdef:
I will not comment on this PR anymore, please review it with other reviewers.
Keep my comments for other reviewers reference.

image

/unassign

Monokaix:
I think we should communicate peacefully in the community to move the project forward: )
Let me summarize. We now want to make the jobflow flow to the failed state, but because there is an if condition, we cannot simply determine the failure state of the jobflow by the number of failed jobs, but this feature of if condition judgment has not been implemented yet. 
Back to the original question, I want to know why the jobflow state is set to failed, what will happen if it is not set, and will it have a significant impact?

Monokaix:
And I think we can implement feature or fix bug step by step, either modify it according to the current idea when the if condition judgment is not implemented, but add a todo statement, or add the expected state judgment when the judgment fails (for example, failed is also an expected state)》.

dongjiang1989:
> I think we should communicate peacefully in the community to move the project forward: ) Let me summarize. We now want to make the jobflow flow to the failed state, but because there is an if condition, we cannot simply determine the failure state of the jobflow by the number of failed jobs, but this feature of if condition judgment has not been implemented yet. Back to the original question, I want to know why the jobflow state is set to failed, what will happen if it is not set, and will it have a significant impact?

Thanks @Monokaix 
As in the above case, this jobflow is impossible to complete. In production, early failures are detected to check the JobTemplate problem and submit the next jobflow execution.

dongjiang1989:
> And I think we can implement feature or fix bug step by step, either modify it according to the current idea when the if condition judgment is not implemented, but add a todo statement, or add the expected state judgment when the judgment fails (for example, failed is also an expected state)》.

Agree.
We can fix the problem now and add a `TODO` description to add judgments and prompts for future implementation

Monokaix:
> > I think we should communicate peacefully in the community to move the project forward: ) Let me summarize. We now want to make the jobflow flow to the failed state, but because there is an if condition, we cannot simply determine the failure state of the jobflow by the number of failed jobs, but this feature of if condition judgment has not been implemented yet. Back to the original question, I want to know why the jobflow state is set to failed, what will happen if it is not set, and will it have a significant impact?
> 
> Thanks @Monokaix As in the above case, this jobflow is impossible to complete. In production, early failures are detected to check the JobTemplate problem and submit the next jobflow execution.

Is there a way that revover jobflow, and if we re-create the failed job, will the jobflow complete successfully?

pkg/controllers/jobflow/state/running.go
```diff
@@ -28,6 +28,8 @@ func (p *runningState) Execute(action v1alpha1.Action) error {
 		return SyncJobFlow(p.jobFlow, func(status *v1alpha1.JobFlowStatus, allJobList int) {
 			if len(status.CompletedJobs) == allJobList {
 				status.State.Phase = v1alpha1.Succeed
+			} else if len(status.FailedJobs) > 0 { // TODO(dongjiang199) Modify it when the if condition judgment is implemented
```
Monokaix, line 31: What about checking the failed job is an expected behavior, for example there is a job depends on current job's failed condition, then we can better solve this controversial PR.

dongjiang1989:
> Is there a way that revover jobflow, and if we re-create the failed job, will the jobflow complete successfully?

Indeed.
Currently, it is the final failed state after multiple retries create job.

https://github.com/volcano-sh/volcano/blob/1ae9a84e16bb327aa54420863cba3c4588e6e325/pkg/controllers/job/job_controller_actions.go#L120-L132

pkg/controllers/jobflow/state/running.go
dongjiang1989: Ensuring that the current jobflow FMS can be reversed is the expected behavior. 

However, there are also scenarios for conditional judgment based on error status; new features and state transition fixes can be implemented separately and step by step :)

PR #4138, opened 2025-03-24
cleanup residual useless victims code in preempt action

### LINKED ISSUE #4137: Residual useless victims code in preempt action
### Description

There are residual useless victims filter code in preempt action needs to be deleted:
https://github.com/volcano-sh/volcano/blob/0403a8849afa05cf9b6e2f42553c4f5433f27092/pkg/scheduler/actions/preempt/preempt.go#L222-L223

Victims filter is only used in shuffle action, preempt action needs to use preemptable to filter preemptees. These residual codes are useless, cause the logic in here:
https://github.com/volcano-sh/volcano/blob/0403a8849afa05cf9b6e2f42553c4f5433f27092/pkg/scheduler/actions/preempt/preempt.go#L351
Only init an empty slice, so the plugin needs to retrieve tasks from the session again, I think the implementation is wrong here.

1.
2.
3.

Residual useless victims code in preempt action needs to be deleted, victims filter only use in shuffle action

master

_No response_

JesseStutler:
#### What type of PR is this?
/kind cleanup

Victims filter is only used in shuffle action, preempt action needs to use preemptable to filter preemptees. These residual codes are useless, cause the logic here:
https://github.com/volcano-sh/volcano/blob/0403a8849afa05cf9b6e2f42553c4f5433f27092/pkg/scheduler/actions/preempt/preempt.go#L351
Only init an empty slice, so the plugin needs to retrieve tasks from the session again, I think the implementation is wrong here. The extension points VictimTasksFns should only be used in shuffle action

Fixes #4137 

```release-note
Cleanup residual useless victims code in preempt action, users of tdm need to enable shuffle action when upgrading to new version
```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -218,9 +218,6 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 			}
 		}
 	}
-
-	// call victimTasksFn to evict tasks
-	victimTasks(ssn)
 }
 
 func (pmpt *Action) UnInitialize() {}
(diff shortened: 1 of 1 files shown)


hwdef:
These codes also confuse me
/lgtm

PR #4146, opened 2025-03-26
networktopologysync

Lily922:

Fixes #

```release-note

```


Monokaix:
Hi，please add more details: )

JesseStutler:
Need to add a design doc

pkg/controllers/networktopologyaware/topoautodiscovery/cache/cache.go
```diff
@@ -0,0 +1,363 @@
...
+	c.Lock()
+	defer c.Unlock()
+
+	if !c.sessionOpened || c.topologyConfigDirty {
+		return
+	}
+	c.topologyConfigDirty = true
+}
+
+func (c *HyperNodeCache) DirtyHyperNode(name string) {
```
JesseStutler, line 114: I want to make sure that does `DirtyHyperNodes` mean that if during a round of session for constructing the HyperNodes tree, a node is added or there is any change of the node label, this HyperNode cannot be constructed and has to wait until the next round to be built? Is this correct?

Monokaix:
`auto discovery` semantic seems not so proper, auto conversion seems better.

pkg/controllers/networktopologyaware/topoautodiscovery/topo_auto_discovery_controller.go
```diff
@@ -0,0 +1,190 @@
+/*
```
Monokaix, line 1: IMO, add a hyperNode controller is a better way, and then add an auto conversion manager.

cmd/controller-manager/app/options/options.go
```diff
@@ -129,6 +142,10 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.Uint32Var(&s.WorkerThreadsForQueue, "worker-threads-for-queue", defaultQueueWorkers, "The number of threads syncing queue operations. The larger the number, the faster the queue processing, but requires more CPU load.")
 	fs.StringSliceVar(&s.Controllers, "controllers", []string{defaultControllers}, fmt.Sprintf("Specify controller gates. Use '*' for all controllers, all knownController: %s ,and we can use "+
 		"'-' to disable controllers, e.g. \"-job-controller,-queue-controller\" to disable job and queue controllers.", knownControllers))
+	fs.StringVar(&s.KubePodNamespace, "kube-pod-namespace", os.Getenv("KUBE_POD_NAMESPACE"), "the namespace of the volcano controller pod")
```
Monokaix, line 145: This is rendered automatically so it's not suitable to set a param and modified by users , we can get this value by env directly.

installer/helm/chart/volcano/templates/controllers.yaml
```diff
@@ -212,4 +222,24 @@ spec:
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: volcano-controller
+    {{- if .Values.custom.common_labels }}
+    {{- toYaml .Values.custom.common_labels | nindent 4 }}
+    {{- end }}
+data:
+  node-label-based-topologies: |
+    topologies:
+  {{- if .Values.custom.node_label_based_topologies_override }}
```
Monokaix, line 241: This check is duplicated.

docs/design/Network Topology Auto Discovery.md
```diff
@@ -0,0 +1,224 @@
...
+  topo-level-keys:
+    - network-tier1
+    - network-tier2
+    - network-tier3
+```
+This configuration instructs the controller to look for nodes with the label keys `network-tier1`，`network-tier2` and `network-tier3`
+
+**_Warning: Ensure the order of the keys is correct. Incorrect ordering of the keys will generate a topology that does not match your expectations._** 
+
+2. Assume you have k8s nodes with following labels:
```
Monokaix, line 68: This example has 4 nodes, but actually created 8 nodes(node1-node8), is this right?

docs/design/Network Topology Auto Discovery.md
```diff
@@ -0,0 +1,224 @@
...
+    type: Node
+  tier: 1
+```
+
+```yaml
+apiVersion: topology.volcano.sh/v1alpha1
+kind: HyperNode
+metadata:
+  labels:
+    volcano.io/label-based-topology-name: topology1
```
Monokaix, line 211: What does this label mean? How about adding a label key on hypedndoe to show the mapping with labels on node?

PR #4149, opened 2025-03-27
Fix bug for vgpu type check

SataQiu:
#### What type of PR is this?
/bug

1. Fix bug for vgpu type check, should return `false` when the `GPUNoUse` annotation is hit.
2. Quick return when gs is nil.

Fixes #

```release-note
NONE
```


shinytang6 (commented):
/lgtm

pkg/scheduler/api/devices/nvidia/vgpu/utils.go
```diff
@@ -280,7 +280,7 @@ func checkGPUtype(annos map[string]string, cardtype string) bool {
 	if ok {
 		if !strings.Contains(nouse, ",") {
 			if strings.Contains(strings.ToUpper(cardtype), strings.ToUpper(nouse)) {
-				return true
+				return false
```
Monokaix, line 283: Add a ut is better?

PR #4163, opened 2025-03-31
[Proposal] Add queue-level-scheduling-policy

ElectricFish7:
#### What type of PR is this?

/kind documentation
/area scheduling
/area controllers

Queue-level-scheduling-policy from a LFX'25 issue:https://github.com/volcano-sh/volcano/issues/3992, which requires the volcano to support setting and using different scheduling policies at the queue level instead of using a globally unified scheduling policy.

Fixes https://github.com/volcano-sh/volcano/issues/3992

@Monokaix
@JesseStutler 

```release-note

```

### DIFF
--- a/docs/design/queue-level-scheduling-policy.md
+++ b/docs/design/queue-level-scheduling-policy.md
@@ -1,0 +1,561 @@
+Currently, Volcano’s scheduling strategy relies on global scheduler configurations, 
+and the relevant functions in Action are at the Session level rather than the plugin level, 
+so significant changes need to be made. This specifically involves two aspects: the definition of CRD and the scheduling logic.
+
+# 1、Changes to the CRD
+
+## 1.1、Extend the queue CRD
+
+Add the `plugins` and `Actions` in the `QueueSpec`, allowing the queue to define its own scheduling policy configuration.
+```go
+type QueueSpec struct {
+    Weight     int32
+    Capability v1.ResourceList
+    State QueueState
+    Reclaimable *bool
+    ExtendClusters []Cluster
+    Guarantee Guarantee `json:"guarantee,omitempty" protobuf:"bytes,4,opt,name=guarantee"`
+    Affinity *Affinity `json:"affinity,omitempty" protobuf:"bytes,6,opt,name=affinity"`
+    Type string `json:"type,omitempty" protobuf:"bytes,7,opt,name=type"`
+    Parent string `json:"parent,omitempty" protobuf:"bytes,8,opt,name=parent"`
+    Deserved v1.ResourceList `json:"deserved,omitempty" protobuf:"bytes,9,opt,name=deserved"`
+    Priority int32 `json:"priority,omitempty" protobuf:"bytes,10,opt,name=priority"`
+	
+    // add Actions and Plugins
+    Actions []string
+    Plugins []PluginConfig
+}
+
+type PluginConfig struct {
+    // plugin name
+    Name string 
+    // plugin args
+    Args map[string]interface{} 
+}
+```
+
+An example for users is as follows:
+
+**Option One:**
+
+```yaml
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: newQueue
+spec:
+  weight: 10
+  actions:
+    - enqueue
+    - allocate
+    - preempt
+  plugins:
+    - name: priority  
+      args:
+        weight: 20    
+    - name: gang     
+      args:
+        minAvailable: 5
+    - name: nodeorder 
+```
+
+**Option Two:**
+
+By configuring a globally unique ConfigMap to store queue and queue-level scheduling plugin configurations.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: volcano-queue-policies
+  namespace: volcano-system
+data:
+  # Queue-level policy (Key is the queue name).
+  queue1: |
+    actions:
+      - enqueue    
+      - allocate   
+      - preempt    
+    plugins:
+      - name: priority
+        args:
+          weight: 10
+      - name: gang
+        args:
+          minAvailable: 5
+  queue2: |
+    actions:
+      - enqueue    
+      - allocate      
+    plugins:
+      - name: nodeorder
+        args:
+          resourceWeights:
+            cpu: 1
+            memory: 0.5
+```
+
+## 1.2、Add the PluginRegistry structure
+
+Currently, the `session` struct contains many extension points, such as `xxxFn`. 
+These functions currently represent global configurations in the session and will also represent plugin-specific configurations in the queue in the future. 
+
+Therefore, we can extract them into a separate struct.
+
+```go
+type PluginRegistry struct {
+    Tiers          []conf.Tier
+    plugins         []Plugin      
+
+    jobOrderFns         map[string]api.CompareFn
+    queueOrderFns       map[string]api.CompareFn
+    victimQueueOrderFns map[string]api.VictimCompareFn
+    taskOrderFns        map[string]api.CompareFn
+    clusterOrderFns     map[string]api.CompareFn
+    predicateFns        map[string]api.PredicateFn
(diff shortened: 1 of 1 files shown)


ElectricFish7:
I’ve updated the proposal, and I think it should support both queue-level and job-level scheduling.
cc @Monokaix @JesseStutler

Monokaix:
@hwdef @kingeasternsun Also supported the job level scheduling, please take a look.

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,612 @@
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: volcano-scheduler-policies
+  namespace: volcano-system
+data:
+  # Queue-level policy (Key is the SchedulerPolicy name).
+  SchedulerPolicy-example-1: |
+    actions:
```
Monokaix, line 54: Should change to `actions: "enqueue, allocate, backfill"`

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,612 @@
...
+      - enqueue    
+      - allocate   
+      - preempt    
+    plugins:
+      - name: priority
+        args:
```
Monokaix, line 60: Please keep the format and fields consistent with default configMap.

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,612 @@
...
+   
+    queues := util.NewPriorityQueue(ssn.QueueOrderFn)
+    queueSet := sets.NewString()
+    jobsMap := map[api.QueueID]*util.PriorityQueue{}
+
+    // When a job specifies a job-level scheduling policy 
+    // or the queue it belongs to specifies a queue-level scheduling policy, 
+    // verify whether the scheduling policy has enabled the corresponding action.
+    for _, job := range ssn.Jobs {
+        SchedulerPolicyName := getSchedulerPolicy(ssn, job)
```
Monokaix, line 351: We can wrap a func here.

docs/design/queue-level-scheduling-policy.md
ElectricFish7: done!

docs/design/queue-level-scheduling-policy.md
ElectricFish7: done!

docs/design/queue-level-scheduling-policy.md
ElectricFish7: I will follow this approach in the implementation.

docs/design/queue-level-scheduling-policy.md
```diff
@@ -0,0 +1,609 @@
...
+    // Add SchedulerPolicy to point to the scheduling policy defined in the configuration file
+    SchedulerPolicy string
+}
+
+```
+
+## 1.2、Extend the Job annotation
+
+To avoid potential issues caused by modifying the `JobSpec` directly, 
+job-level scheduling policies are specified via `annotations` in the Job definition.
```
hwdef, line 37: `annotations` is not a official practice.

hwdef:
Overall it's ok, but I prefer `cr` to `configmap`

I think we should modify a somewhat annoying legacy problem, that is, if the parsing configuration fails, the default configuration will be used. This is implicit. I hope that the reading configuration fails and panics directly

docs/design/queue-level-scheduling-policy.md
Monokaix: Yeah actually we recommend to use it at queue level, there are two aspects consideration:
1. Queue level is more common in multi tenants scenario, which is recommended.
2. If new fields are added to the pod and job APIs, a lot of adaptation is required because of this major API change, while setting policies on queues will be less expensive for users.

Monokaix:
> Overall it's ok, but I prefer `cr` to `configmap`
> 
> I think we should modify a somewhat annoying legacy problem, that is, if the parsing configuration fails, the default configuration will be used. This is implicit. I hope that the reading configuration fails and panics directly

The problem of cr is that it's not consistent with current global config which is in configMap.

ElectricFish7:
Overall, we have three approaches to mount scheduling policies:
- Using a single ConfigMap — This is similar to the implementation in the current proposal. The advantage of this method is its simplicity and consistency with the default scheduler configuration file. However, it poses a risk where users may unintentionally (or intentionally) modify scheduling policies that do not belong to them.

- Using one ConfigMap per scheduling policy — This approach prevents users from modifying other users’ scheduling policies, offering better isolation. However, it could lead to a large number of ConfigMaps, making file mounting and scheduler access more complex.

- Using a Custom Resource Definition (CRD) — This method provides better structure and flexibility, but its usage differs from the default scheduler configuration file, potentially increasing the learning curve or setup complexity.

hwdef:
> Overall, we have three approaches to mount scheduling policies:
> 
> * Using a single ConfigMap — This is similar to the implementation in the current proposal. The advantage of this method is its simplicity and consistency with the default scheduler configuration file. However, it poses a risk where users may unintentionally (or intentionally) modify scheduling policies that do not belong to them.
> * Using one ConfigMap per scheduling policy — This approach prevents users from modifying other users’ scheduling policies, offering better isolation. However, it could lead to a large number of ConfigMaps, making file mounting and scheduler access more complex.
> * Using a Custom Resource Definition (CRD) — This method provides better structure and flexibility, but its usage differs from the default scheduler configuration file, potentially increasing the learning curve or setup complexity.

If crd is not accepted, the second one is also good

ElectricFish7:
I have updated the design document and user examples.
PTAL @Monokaix @JesseStutler

XbaoWu:
I suggest that we should add some explanations for queue-level schedule policy: 
For example, when users previously used cluster-level ( which refers to the previous schedule policy mode ), if binpack is used, the effect that tasks in the cluster will preferentially fill the nodes of existing tasks can be achieved, which can reduce resource fragmentation. However, in queue-level mode, if queue A is configured with binpack, queue B is configured with nodeorder. 
That standing in the cluster perspective, in fact, did not play a better role in reducing resource fragmentation ; from the queue perspective, the tasks in queue A are preferentially assigned to the nodes with tasks, and the tasks may be more compact. The tasks in queue B are evenly distributed in the cluster according to the node situation. 

We 'd better allow users to have a correct expectation of the impact of queue configuration (in queue-level mode) , help them to configure queue-level schedule policy more reasonably, and improve their experience of using volcano. 
 
In addition, in queue-level mode, if queue A uses **proportion** but queue B uses **capacity**, will there be a conflict in resource management ?

TriangleWuYQ:
> I suggest that we should add some explanations for queue-level schedule policy: For example, when users previously used cluster-level ( which refers to the previous schedule policy mode ), if binpack is used, the effect that tasks in the cluster will preferentially fill the nodes of existing tasks can be achieved, which can reduce resource fragmentation. However, in queue-level mode, if queue A is configured with binpack, queue B is configured with spread. That standing in the cluster perspective, in fact, did not play a better role in reducing resource fragmentation ; from the queue perspective, the tasks in queue A are preferentially assigned to the nodes with tasks, and the tasks may be more compact. The tasks in queue B are evenly distributed in the cluster according to the node situation.
> 
> We 'd better allow users to have a correct expectation of the impact of queue configuration (in queue-level mode) , help them to configure queue-level schedule policy more reasonably, and improve their experience of using volcano.
> 
> In addition, in queue-level mode, if queue A uses **proportion** but queue B uses **capacity**, will there be a conflict in resource management ?

I think queue capacity management plugins such as `proportion` and `capacity` should be configured at the global level?

XbaoWu:
> I think queue capacity management plugins such as `proportion` and `capacity` should be configured at the global level?

I think this is a solution, we may need to simply classify the existing plugins, which can only be applied globally, which can be configured at the queue-level. 
In order to be compatible with the configuration of the previous version of the users, queue-level schedule policy should be based on default schedule policy when encountering a global plugin.

Issue #4170, opened 2025-04-03
[Umbrella]Track of Volcano graduation security audit


JesseStutler:
### What is the problem you're trying to solve

The Volcano community is considering graduating from the CNCF community before 2026. A security audit has already been initiated, and some security-related code enhancements and fixes need to be implemented. The known issues that need to be fixed are:
- [ ] File permission issue, some files have executable permissions, need to be removed
- [ ] The pprof interface is enabled by default, and a switch needs to be added for users to decide whether to turn it on or off, with the default being off
- ...More security-related will be added 

### Describe the solution you'd like

Address all issues raised by the security audit team

muskan2622:
can you give me more insight on this ?? @JesseStutler . is this already done or what part is left?

JesseStutler:
> can you give me more insight on this ?? [@JesseStutler](https://github.com/JesseStutler) . is this already done or what part is left?
@muskan2622 
Generally, these will be fixed by the community maintainers, this issue is just for tracking. However, if you want to participate, I have already submitted a PR in the volcano main repo for adding SecurityContext configuration: 
https://github.com/volcano-sh/volcano/pull/4207
Are you interested in making the same changes to the volcano dashboard? https://github.com/volcano-sh/dashboard/blob/main/deployment/volcano-dashboard.yaml

The security audit is quite urgent, so if you are willing to contribute, the PR needs to be submitted asap and merged this week :)

PR #4173, opened 2025-04-03
[Security] Add a switch to control whether enable pprof in scheduler

JesseStutler:
#### What type of PR is this?
Related with #4170, add a switch to control whether enable pprof in scheduler

Currently in cmd/scheduler/main.go,` _ "net/http/pprof"` is imposed, which causes some handlers in the pprof package, such as `/debug/pprof`, to be registered to `DefaultServeMux` by default without a switch to control whether they are registered, posing a potential security risk. Currently, the registration of metrics handlers and pprof handlers has been unified, and to avoid third-party packages directly registering handlers to DefaultServeMux, a new ServeMux has been created to register the handlers, and then listen.

Related with #4170

NONE

```release-note
NONE
```

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -67,6 +67,7 @@ type ServerOption struct {
 	DefaultQueue        string
 	PrintVersion        bool
 	EnableMetrics       bool
+	EnablePprof         bool
 	ListenAddress       string
 	EnablePriorityClass bool
 	EnableCSIStorage    bool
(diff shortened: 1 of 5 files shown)


cmd/scheduler/app/options/options.go
```diff
@@ -141,6 +142,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 		"Enable tracking of available storage capacity that CSI drivers provide; it is false by default")
 	fs.BoolVar(&s.EnableHealthz, "enable-healthz", false, "Enable the health check; it is false by default")
 	fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")
+	fs.BoolVar(&s.EnablePprof, "enable-pprof", false, "Enable the pprof endpoint; it is false by default")
```
shinytang6, line 145: Q: why not enable it by default, l remember k8s scheduler enabling pprof by default.

shinytang6 (commented):
lgtm

cmd/scheduler/app/options/options.go
JesseStutler: This is the feedback from the security audit team, the original words were "The debug endpoint should be opt-in instead of opt-out," so it is turned off by default, and users can choose to turn it on

Issue #4104, opened 2025-03-18
Panic when capacity limits the total resource of subqueues


wzshiming:
### Please describe your problem in detail

<details><summary>volcano-scheduler.conf</summary>
<p>

``` yaml
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, backfill, reclaim"
    tiers:
    - plugins:
      - name: priority
      # - name: gang
      #   enablePreemptable: false
      - name: conformance
    - plugins:
      - name: overcommit
      - name: drf
        enablePreemptable: false
      - name: predicates
      - name: capacity
        enableHierarchy: true
      - name: nodeorder
      - name: binpack
kind: ConfigMap
metadata:
  name: volcano-scheduler-configmap
  namespace: volcano-system
```

</p>
</details> 

<details><summary>2 Jobs</summary>
<p>

``` yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: volcano-job-0-1
  namespace: default
spec:
  maxRetry: 3
  minAvailable: 10
  policies:
  - action: RestartJob
    event: PodEvicted
  queue: test-queue-0
  schedulerName: volcano
  tasks:
  - maxRetry: 3
    minAvailable: 10
    name: task
    replicas: 10
    template:
      metadata:
        labels:
          test-instance: "1"
      spec:
        affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
              - matchExpressions:
                - key: type
                  operator: In
                  values:
                  - kwok
        containers:
        - args:
          - 30s
          image: hello-world
          name: sleep
          resources:
            requests:
              cpu: "1"
              memory: 100Mi
        restartPolicy: Never
        tolerations:
        - effect: NoSchedule
          key: kwok.x-k8s.io/node
          operator: Exists
  ttlSecondsAfterFinished: 1
status:
  conditions:
  - lastTransitionTime: "2025-03-18T05:54:48Z"
    status: Pending
  minAvailable: 10
  state:
    lastTransitionTime: "2025-03-18T05:54:48Z"
    phase: Pending
---
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: volcano-job-1-2
  namespace: default
spec:
  maxRetry: 3
  minAvailable: 10
  policies:
  - action: RestartJob
    event: PodEvicted
  queue: test-queue-1
[...]

PR #4179, opened 2025-04-07
fix: panic when total guarantee of child queue exceeds capacity of parent

JackyTYang:
fix panic when total guarantee of child queue exceeds capacity of parent
#4104

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity_test.go
+++ b/pkg/scheduler/plugins/capacity/capacity_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package capacity
 
 import (
+	"k8s.io/apimachinery/pkg/api/resource"
 	"os"
 	"testing"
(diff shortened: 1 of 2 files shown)


JackyTYang:
reopen

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package capacity
 
 import (
+	"k8s.io/apimachinery/pkg/api/resource"
```
googs1025, line 20: nit: import sort

JackyTYang: modified import sort

JesseStutler:
/ok-to-test
/lgtm

Thanks for your contribution!

PR #4205, opened 2025-04-18
Add fuzz tests for job controller

AdamKorcz:
#### What type of PR is this?

Adds two fuzz tests for the job controller. The added dependency is for randomizing structs from the fuzzer's data.

Also adds a build script for OSS-Fuzz.

Fixes #

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module volcano.sh/volcano
 go 1.23.0
 
 require (
+	github.com/AdaLogics/go-fuzz-headers v0.0.0-20240806141605-e8a1dd7889d6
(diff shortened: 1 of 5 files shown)


AdamKorcz:
The OSS-Fuzz build script is being used for Volcanos OSS-Fuzz integration here: https://github.com/google/oss-fuzz/pull/13235

Monokaix:
Thanks for your contribution！Do we need add more fuzz test for other package?

AdamKorcz:
> Thanks for your contribution！Do we need add more fuzz test for other packages?

It is not needed for the current work, but over time it will be good to do.

PR #4218, opened 2025-04-23
network topology: Update network topology docs, add task level network topology scheduling.

### LINKED ISSUE #4188: Support task level network topology constrain

Currently vcjob and PodGroup support setting workload level network topology constrain, we should also support task level network topology setting.

In both training and inference scenarios, we don't necessarily need all tasks within the entire job to be restricted to the same HyperNode. 

1. In training scenarios, with pipeline parallelism (pp) and data parallelism (dp), it's sufficient for the data parallel (dp) tasks to be distributed within one topology domain. 

2. In inference scenarios, such as vllm, the requirement is usually only for the workers to be deployed within the same topology domain, while the leader has no topology constraints.

### Additional context

1. Should modify Volcano scheduler to support task level topology.
2. Should add webhook to validate, e.g., task level HighestTierAllowed should not be greater than job level HighestTierAllowed.

hwdef:
#### What type of PR is this?

Update the network topology-aware scheduling documentation by adding task-level network topology-aware scheduling. Additionally, fixed formatting issues in the code comments to ensure consistency and readability.

Fixes #4188


docs/design/Network Topology Aware Scheduling.md
```diff
@@ -544,26 +559,59 @@ type NetworkTopologySpec struct {
...
+	  TaskNetworkTopology map[string]*NetworkTopologySpec
+}
+
 ```
 
 auto generate code for job, hyperJob, hyperNode,  
 refer the API repo [volcano-sh/apis: The API (CRD) of Volcano (github.com)](https://github.com/volcano-sh/apis) and add CRDs, generate codes and submit to the repo.
 
+
+**Impact on TaskOrderFn:**
```
Monokaix, line 593: Here should be a best practice that task with network topology should set a higher priority that normal tasks.

Monokaix:
Should also add these API change of https://github.com/volcano-sh/apis/pull/174/ in this doc.

hwdef:
> Should also add these API change of [volcano-sh/apis#174](https://github.com/volcano-sh/apis/pull/174) in this doc.

Done. Please review again

PR #4241, opened 2025-04-28
refactor: refactor resource reclamation

### LINKED ISSUE #3738: Need to refactor the reclaim action

elysium-w:
…rove readability and maintainability.

 refactor resource reclamation into modular functions to improve readability and maintainability.

Fixes #3738 

```release-note
none
```

### DIFF
--- a/pkg/scheduler/plugins/extender/extender.go
+++ b/pkg/scheduler/plugins/extender/extender.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	"volcano.sh/volcano/pkg/scheduler/api"
@@ -67,6 +69,8 @@ const (
 
 	// 10MB
 	maxBodySize = 10 << 20
+	// ExtenderManagedResources is the managed resources list split by ","
+	ExtenderManagedResources = "extender.managedResources"
 )
 
 type extenderConfig struct {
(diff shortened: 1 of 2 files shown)


elysium-w:
@hwdef @Monokaix @JesseStutler  PTAL

elysium-w:
@hwdef @Monokaix @JesseStutler If you have time, please take a look. Thank you very much.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -37,59 +37,24 @@ func (ra *Action) Name() string {
 
 func (ra *Action) Initialize() {}
 
+// the whole logic of reclaim is as follows:
```
hwdef, line 40: It seems to be a useless comment

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -98,132 +63,174 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
 
-			if len(reclaimees) == 0 {
-				klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
-				continue
-			}
+		ra.collectAndProcessVitimTasks(ssn, node, task, reclaimees, &assigned)
 
-			victims := ssn.Reclaimable(task, reclaimees)
+	}
+	return assigned
```
hwdef, line 123: the `assigned` var is weird. It is passed through the pointer, and it not be used in line95

hwdef (commented):
Others is ok to me.

pkg/scheduler/actions/reclaim/reclaim.go
elysium-w: does it ok right now?

hwdef: Please delete the note mentioned in the previous message.
By the way, This comment makes the commit look like AI-generated code that doesn't know what it's doing.

Issue #4189, opened 2025-04-15
Prevent the scheduling of pods in non-open queues.


Monokaix:
### What is the problem you're trying to solve

**Backgrounds:** 
Initially, when submitting a pod to a queue that is not in the open state, the webhook will reject the creation of the pod, but for performance considerations, the verification is transferred from the pod to the PodGroup, because the scheduling model of volcano is also based on PodGroup, so it is reasonable to transfer it to PodGroup. However, if the PodGroup is created first and then the queue is closed, the webhook will not be able to reject the subsequent submission of the pod, so it is necessary to intercept the scheduling of the pod in the running state, for details, see https://github.com/volcano-sh/volcano/pull/4132

For more detailed queue state management design, see: https://github.com/volcano-sh/volcano/blob/master/docs/design/queue/queue-state-management.md

### Describe the solution you'd like

An optional way is that: Add the queue state check in `proportion` and `capacity` and DO NOT schedule pods and jobs if queue is not in open state. 
However, in the case of concurrency, there may be a delay for the Scheduler to watch the queue status change. In extreme concurrency situations, pod scheduling cannot be completely intercepted. This requires additional consideration.

googs1025:
/cc 
I'd like to take a look at this issue in the next few days. 😄

JesseStutler:
@googs1025 Hi any progress :)?

googs1025:
Sorry for late I will handle this this weekend

Monokaix:
Should also add check in when enqueue action not enabled.

PR #4263, opened 2025-05-09
fix: prevent the scheduling of pods in noopen queues

googs1025:
#### What type of PR is this?
/kind bug
- prevent the scheduling of pods in noopen queues

Fixes https://github.com/volcano-sh/volcano/issues/4189

```release-note
None
```


googs1025:
I will add some unit test for this fix

googs1025:
/cc @Monokaix @JesseStutler 

sorry for the late.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -188,6 +188,10 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		attr := cp.queueOpts[queueID]
 		queue := ssn.Queues[queueID]
+		// If the queue is not open, do not enqueue
```
googs1025, line 191: If my understanding is correct, we should just check the status in AddJobEnqueueableFn in the plugin. But I found a problem, the code is as follows:

https://github.com/volcano-sh/volcano/blob/cbc4990d754e9cfdad7e6726f07bf648b676b824/pkg/scheduler/actions/enqueue/enqueue.go#L94-L100

`job.PodGroup.Spec.MinResources == nil` I'm not sure if this has any impact on this.

lowang-bh (commented):
/lgtm

hwdef (commented):
/approve

googs1025:
do we need to cherry-pick this fix?

pkg/scheduler/plugins/capacity/capacity.go
Monokaix: It's not enough to just check in `enqueue` action, because user may not enable this action...

Monokaix:
> do we need to cherry-pick this fix?

No need to cherry pick as it's a change only in master branch.

Issue #4271, opened 2025-05-12
When deploying an MPI job, if the task.replicas is not configured or set to 0, it will trigger a controller panic.


guoqinwill:
### Description

When deploying an MPI job, if the task.replicas is not configured or set to 0, it will trigger a controller panic.

### Steps to reproduce the issue

1.Create a vcjob named test, configure the MPI plugin for the job, set the job's minAvailable to 1, and define two tasks. Set task1's replicas to 1 and minAvailable to 1. Do not configure replicas for task2 and set its minAvailable to 0. The YAML example is as follows:
```apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: test
spec:
  schedulerName: volcano
  minAvailable: 1
  plugins:
    mpi:
      - '--port=2222'
  tasks:
    - maxRetry: 3
      replicas: 1
      minAvailable: 0
      name: "task1"
      template:
        spec:
          terminationGracePeriodSeconds: 3
          containers:
            - name: container1
              image: busybox:latest
              command: ["sh"]
              args: ["-c", "echo Hello, Kubernetes! && sleep 3600" ]
              resources:
                requests:
                  huawei.com/ascend-310: "1"
                limits:
                  huawei.com/ascend-310: "1"
          restartPolicy: OnFailure
    - maxRetry: 3
      minAvailable: 0
      name: "task2"
      template:
        spec:
          terminationGracePeriodSeconds: 3
          containers:
            - name: container1
              image: busybox:latest
              command: ["sh"]
              args: ["-c", "echo Hello, Kubernetes! && sleep 3600" ]
              resources:
                requests:
                  huawei.com/ascend-310: "1"
                limits:
                  huawei.com/ascend-310: "1"
          restartPolicy: OnFailure
```
2.After submitting the job, no pods were created, and the Volcano controller continuously restarted. The controller logs showed a panic.
```[root@gqtest111 ~]# kubectl get pod -A
NAMESPACE     NAME                                                   READY   STATUS              RESTARTS        AGE
default       nginx-test-689487c66d-rqxn6                            1/1     Running             0               4d3h
kube-system   volcano-admission-58769688df-8f2gs                     1/1     Running             0               3h19m
kube-system   volcano-controller-856c7f59bb-z4wjn                    0/1     CrashLoopBackOff    44 (109s ago)   3h19m
kube-system   volcano-scheduler-7bb577575c-f2w6n                     1/1     Running             0               3h19m
```
[...]

PR #4272, opened 2025-05-12
fix controller panic for mpi job

guoqinwill:
#### What type of PR is this?
/kind bug

Fixes # https://github.com/volcano-sh/volcano/issues/4271 

```release-note

```

### DIFF
--- a/pkg/controllers/job/plugins/distributed-framework/mpi/mpi.go
+++ b/pkg/controllers/job/plugins/distributed-framework/mpi/mpi.go
@@ -18,11 +18,13 @@ package mpi
 
 import (
 	"flag"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
 	batch "volcano.sh/apis/pkg/apis/batch/v1alpha1"
+
 	"volcano.sh/volcano/pkg/controllers/job/helpers"
 	pluginsinterface "volcano.sh/volcano/pkg/controllers/job/plugins/interface"
 )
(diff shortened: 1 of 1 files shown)


Monokaix:
Should also modify the API with kube builder to set a default replica.

guoqinwill:
> Should also modify the API with kube builder to set a default replica.

In this pr, we will neither include this modification nor alter the current behavior (where pods are not created if replicas are not configured). The modification will instead be deferred to v1beta1.

pkg/controllers/job/plugins/distributed-framework/mpi/mpi.go
```diff
@@ -108,22 +110,38 @@ func (mp *Plugin) OnPodCreate(pod *v1.Pod, job *batch.Job) error {
...
+		builder.WriteString(".")
+		builder.WriteString(subdomain)
+
+		if i < int(task.Replicas)-1 {
+			builder.WriteString(",")
+		}
+
+		// If a hostname is explicitly specified, assume only one host is needed.
+		// Break the loop early to avoid generating additional hosts.
+		if task.Template.Spec.Hostname != "" {
```
JesseStutler, line 139: The logic here is not right. If Hostname is not empty, there will be an extra comma at the end. The extra comma at the end will be removed in the old logic, but the current writing will add an extra comma at the end and didn't remove it

guoqinwill: The new logic calculates int(task.Replicas) - 1 and adds commas after each element except the last one to prevent slicing errors. This ensures no extra commas are generated.

JesseStutler: No, if your replicas is greater than 1, and both hostname and subdomain exist, then it will be "hostname.subdomain," with an extra comma. If the hostname is not empty, then it break here, and then return builder directly, so there will be an extra comma.

guoqinwill: Yes, you’re right. I’ve made the necessary changes.

pkg/controllers/job/plugins/distributed-framework/mpi/mpi.go
```diff
@@ -108,22 +110,38 @@ func (mp *Plugin) OnPodCreate(pod *v1.Pod, job *batch.Job) error {
...
 			subdomain = jobName
 		}
-		hosts = hosts + hostName + "." + subdomain + ","
-		if len(task.Template.Spec.Hostname) != 0 {
+
+		builder.WriteString(hostName)
+		builder.WriteString(".")
+		builder.WriteString(subdomain)
+
+		if task.Template.Spec.Hostname == "" && i < int(task.Replicas)-1 {
```
JesseStutler, line 133: It's still wrong... If you change it this way, the comma will be added only when Hostname is empty, which is different from the original logic. In essence, the panic was only because the situation of Replicas=0 was not judged, and the index was -1, which was out of bounds. So, is it enough to just make a special judgment for Replicas=0 at the very beginning. Of course, you can keep your writing method, but the logic needs to be consistent with the original codes

guoqinwill: When the hostname is not empty, appending a comma at the end and then deleting it versus directly not appending a comma—wouldn’t the final result be the same?

JesseStutler: ok

JesseStutler:
lgtm
Please squash the commit into one, thanks

hwdef (commented):
/approve

Monokaix:
Please also cherry pick to release1.9, 1.10 and 1.11

Issue #4238, opened 2025-04-25
PyTorch plugin set `WORLD_SIZE` to total number of replicas in the job, not just the ones for master and worker tasks


antoinebrl:
### Description

Hello 👋 !

I am currently working on a training job that utilizes PyTorch, where each rank needs to interact with an service that leaves outside of the GPU machine. The lifecycle of this service is aligned with the training. To facilitate this, I have activated the pytorch plugin in Volcano and added an additional task with its corresponding replicas. Below, you will find a minimal job manifest illustrating the structure:

```yaml
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: multi-node-with-extra
spec:
  schedulerName: volcano
  queue: default
  maxRetry: 1
  ttlSecondsAfterFinished: 300
  plugins:
    pytorch: [--master=master, --worker=worker, --port=23456]
  tasks:
  - replicas: 1
    name: master
    template:
      spec:
        containers:
        - name: main
          image: alpine
          command:
            - printenv

  - replicas: 3
    name: worker
    template:
      spec:
        containers:
        - name: main
          image: alpine
          imagePullPolicy: IfNotPresent
          command:
            - printenv

  - replicas: 4
    name: extras
    template:
      spec:
        containers:
        - name: main
          image: alpine
          imagePullPolicy: IfNotPresent
          command:
            - printenv
```

### Describe the results you received and expected

The challenge I am encountering is that the `WORLD_SIZE` environment variable is:
1. being set across all pods. Just like `RANK`, I was expecting it to only be available on the master and workers.
2. its value is being set to 8, instead of the expected 4 (1+3). My assumption was that the PyTorch plugin would calculate the world size based on the aggregate number of replicas for the master and worker, as specified by `--master=master --worker=worker`. In short, `WORLD_SIZE=max(RANK)+1`.

I tried to manually add `WORLD_SIZE` to the container specs for the master and workers, but this env get overridden by Volcano.

I would appreciate any guidance or insights on whether this is intentional or not, and what would be a workaround to inject the expected `WORLD_SIZE` value into the different pods. Thank you in advance for your help! 

### What version of Volcano are you using?

1.11

JesseStutler:
/assign
Will take a look later

JesseStutler:
/help

Are any contributors willing to solve this issue?  :)

Issue #4166, opened 2025-04-02
:bug: ENV 'WORLD_SIZE' counting method of pytorch-plugin


Hcryw:
### Description

In a PyTorch job, the `RANK` is calculated based on the `taskType`, with only "master" and "worker" tasks being considered. However, when calculating `WORLD_SIZE` with the `getTotalReplicas` function, replicas of all tasks within the VCJob will be counted: 

https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/plugins/distributed-framework/pytorch/pytorch.go#L94-L115

This causes an issue when we try to add an extra `Task` specifically for collecting and displaying TensorBoard logs. The calculated `WORLD_SIZE` is 1 greater than the actual number, leading to a mismatch between the expected number of connections and `WORLD_SIZE` in PyTorch DDP, which ultimately results in a timeout during node synchronization.

### Steps to reproduce the issue

1. Create a VCJob with `pytorch` plugin which contains 2 tasks, a master with replicas 1 and a worker with replicas 1.
2. Add a 'tensorboard' task with replicas 1 to gather logs.
3. `WORLD_SIZE` will be 3 instead of 2.
4. eg: 
```
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: test
  namespace: default
spec:
  plugins:
    svc: []
    pytorch:
      - '--master=master'
      - '--worker=worker'
      - '--port=23456'
  maxRetry: 3
  tasks:
    - replicas: 1
      name: master
      policies:
        - event: TaskCompleted
          action: CompleteJob
      template:
        spec:
          imagePullSecrets: null
          containers:
            - image: nvcr.io/nvidia/cuda:12.4.1-cudnn-devel-ubuntu20.04
              imagePullPolicy: IfNotPresent
              name: master
              args:
                - sleep inf
              command:
                - bash
                - '-c'
              resources:
                requests:
                  cpu: '1'
                  memory: 1Gi
                limits:
                  cpu: '1'
                  memory: 1Gi
          restartPolicy: Never
    - replicas: 1
      name: worker
      policies:
        - event: TaskCompleted
          action: CompleteJob
      template:
        spec:
          imagePullSecrets: null
          containers:
            - image: nvcr.io/nvidia/cuda:12.4.1-cudnn-devel-ubuntu20.04
              imagePullPolicy: IfNotPresent
              name: worker
              args:
                - sleep inf
              command:
                - bash
                - '-c'
              resources:
[...]

PR #4281, opened 2025-05-13
WORLD_SIZE calculation for PyTorch Jobs

murali1539:
#### What type of PR is this?
/kind bug

The WORLD_SIZE calculation for PyTorch Jobs includes any extra tasks apart from Master and Worker are also included. This is resulting in Failures.
**Expected** : 
WORLD_SIZE should include only tasks that are from Master and Worker

Fixes https://github.com/volcano-sh/volcano/issues/4238
Fixes https://github.com/volcano-sh/volcano/issues/4166

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/job/plugins/distributed-framework/pytorch/pytorch.go
+++ b/pkg/controllers/job/plugins/distributed-framework/pytorch/pytorch.go
@@ -120,7 +120,9 @@ func (pp *pytorchPlugin) OnPodCreate(pod *v1.Pod, job *batch.Job) error {
 func (pp *pytorchPlugin) getTotalReplicas(job *batch.Job) int32 {
 	jobReplicas := int32(0)
 	for _, task := range job.Spec.Tasks {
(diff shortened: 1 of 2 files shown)


pkg/controllers/job/plugins/distributed-framework/pytorch/pytorch.go
JesseStutler, line 1: We also should only set WORLD_SIZE env for master pod's container and worker pods' container in `OnPodCreate`

Monokaix:
/approve
Thanks!

pkg/controllers/job/plugins/distributed-framework/pytorch/pytorch.go
JesseStutler: We can filter to not set the `WORLD_SIZE` env and not open ports for other tasks in another PR, current modification is fine.

Monokaix:
Please also cherrypick to release1.9,1.10 and 1.11

PR #4285, opened 2025-05-14
feat: add graceful shutdown server for webhook manager.

fengruotj:
#### What type of PR is this?
/kind feature

Volcano webhook provides callback process for kubernetes resource mutation and validation. As an HTTP server, it need support graceful shutdown to ensure proper behavior during kubernetes rolling upgrades or when the webhook pod is deleted. Kubelet sends a SIGTERM signal to the container. The webhook should handle this signal by performing a graceful shutdown. However, calling http.Server.Close immediately closes all active net.Listeners and any connections in the StateNew, StateActive, or StateIdle states. For a graceful shutdown, we should use Server.Shutdown, which first closes all open listeners, then closes all idle connections, and finally waits indefinitely for active connections to become idle before shutting down.

Second, regarding the stop channel returned from signal handling: if multiple handlers are waiting on different channels, only one of them will be triggered. Therefore, we must use `signals.SetupSignalContext()` to handle signals, as it will close the signal channel, allowing multiple handlers to be triggered simultaneously.

@hwdef 

None

### DIFF
--- a/cmd/webhook-manager/app/options/options_test.go
+++ b/cmd/webhook-manager/app/options/options_test.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2025 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


fengruotj:
@merryzhou @william-wang PTAL~

cmd/webhook-manager/app/options/options_test.go
```diff
@@ -0,0 +1,64 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
```
Monokaix, line 19: Please sort import.

PR #4093, opened 2025-03-14
[design] hyperNode lifecycle management

Monokaix:
#### What type of PR is this?
/kind docs
/kind feature
#### What this PR does / why we need it:
Add hypenode lifecycle management design doc
#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:
An optional way of managing hyperNode lifecycle design:  https://github.com/volcano-sh/volcano/pull/4014
#### Does this PR introduce a user-facing change?

```release-note

```

PR #4286, opened 2025-05-15
Add hyperNode controller framework and IB discovery

Monokaix:
#### What type of PR is this?
/kind feature
A new implement of https://github.com/volcano-sh/volcano/pull/4093 after weekly meeting discussion, and we should provide a basic available hypernode auto discovery mechanism to let users use Network Topology Aware Scheduling easily.
And will support more network protocol such as Roce, and will also support label based discovery, which is tracked by https://github.com/volcano-sh/volcano/pull/4146/

Fixes #

```release-note
The hyperNodes can be created and maintained automatically according cluster hardware and network configuration
```


hwdef:
excellent!

Monokaix:
cc @hwdef @lowang-bh @yeahdongcn @JesseStutler

docs/user-guide/how_to_use_hypernode_auto_discovery.md
```diff
@@ -0,0 +1,94 @@
...
+        config:
+          endpoint: https://ufm-server:8080
+          username: admin
+          password: password
+          insecureSkipVerify: true
+      - source: roce
+        enabled: false
+        interval: 15m
+        config:
+          endpoint: https://roce-server:9090
```
yeahdongcn, line 39: `config.endpoint` makes sense for UFM since it provides a centralized interface, but RoCE doesn’t have centralized management. What is this field intended to connect to in the case of RoCE?

Monokaix: That's a good catch, currentlt roce type is not supported, and we have no roce environment to test this feature. Do you have any idea what fields should be filled in roce environment？

Monokaix:
cc @k82cn Please also take a look: )

docs/user-guide/how_to_use_hypernode_auto_discovery.md
yeahdongcn: As we discussed earlier, for RoCE we can use LLDP to collect neighbor information from each node and switch, then reconstruct the full topology. On the node side, tools like https://github.com/lldpd/lldpd or other open-source alternatives can be used. However, for switches, LLDP support and access methods can vary significantly between vendors. This makes it challenging to maintain compatibility across all switch types and may require contributions from the vendors themselves.

Makefile
```diff
@@ -94,7 +94,7 @@ vcctl: init
 image_bins: vc-scheduler vc-controller-manager vc-webhook-manager vc-agent
 
 images:
-	for name in controller-manager scheduler webhook-manager agent; do\
+	for name in controller-manager ; do\
```
lowang-bh, line 97: why removed other component？

Monokaix: My mistake, fixed.

docs/user-guide/how_to_use_hypernode_auto_discovery.md
Monokaix: Ok，I see. This interface is extensible enough to cover RoCE in the future, if we need a daemonset we can add it in RoCE scenario.

pkg/controllers/hypernode/configmap_handler.go
```diff
@@ -0,0 +1,114 @@
...
+	namespace := os.Getenv(config.NamespaceEnvKey)
+	if namespace == "" {
+		namespace = config.DefaultNamespace
+	}
+	releaseName := os.Getenv(config.ReleaseNameEnvKey)
+	if releaseName == "" {
+		releaseName = config.DefaultReleaseName
+	}
+	hn.configMapNamespace = namespace
+	hn.configMapName = releaseName + "-controller-configmap"
```
JesseStutler, line 44: A new `networkTopologyDiscovery` configuration is added, but we add a new configmap at the controller letvel, is this appropriate? If the controller needs to add configurations of other fields in the future, should we also use this configmap? Currently, if any data of the configmap changed, it will trigger the syncHandler. So I think it is necessary to refine the logic of enqueueConfigmap that syncHandler should be triggered only when `networkTopologyDiscovery` field changes.

docs/design/hyperNode-auto-discovery.md
```diff
@@ -0,0 +1,196 @@
...
+  if node.Labels == nil {
+   node.Labels = make(map[string]string)
+  }
+  node.Labels[api.NetworkTopologySourceLabelKey] = source
+  discoveredNodeMap[node.Name] = node
+ }
+
+ for name, node := range discoveredNodeMap {
+  if _, exists := existingNodeMap[name]; !exists {
+   [...]
```
JesseStutler, line 175: Why is the create and update logic hidden here, but the delete logic is displayed in L183-L188? 😂

pkg/controllers/hypernode/discovery/label/label.go
JesseStutler, line 1: So this PR: https://github.com/volcano-sh/volcano/pull/4146 needs to be factored to be Discoverer style?

PR #4298, opened 2025-05-19
Vcctl supports merging multiple kubeconfig to support context switching among multiple k8s clusters.

### LINKED ISSUE #4291: Vcctl supports merging kubeconfig and contexts for accessing multiple k8s clusters.

We now have multiple k8s clusters in our current private cloud environment. Different k8s clusters are accessed through `kubectl` by using multiple kubeconfig and context switching.
`kubectl` supports merging different kubeconfig and contexts. But vcctl doesn't support it.

### Details
This `KUBECONFIG` configuration is similar to our environment, :
```
$ echo $KUBECONFIG
/root/.kube/config.yaml:/opt/clusters/prod001/kubeconfig.yaml:/opt/clusters/prod002/kubeconfig.yaml:/opt/clusters/prod003/kubeconfig.yaml:/opt/clusters/prod004/kubeconfig.yaml
```
We can use `kubectl` to access multiple k8s cluster by switching the context:
```
$ kubectl config use-context prod-001-cluster
$ kubectl get pod
NAME                                   READY    STATUS    RESTARTS   AGE
prod001-pod1                  1/1          Running   0                     197d
prod001-pod2                  1/1          Running   0                     197d
prod001-pod3                  1/1          Running   0                     197d
$ 
$ kubectl config use-context prod-002-cluster
$ kubectl get pod
NAME                                   READY    STATUS    RESTARTS   AGE
prod002-pod1                  1/1          Running   0                     197d
prod002-pod2                  1/1          Running   0                     197d
prod002-pod3                  1/1          Running   0                     197d
```
But vcctl doesn't support using kubeconfig by this way. We will receive an error like this:
```
$ vcctl queue list
Failed to list queue: stat /root/.kube/config.yaml:/opt/clusters/prod001/kubeconfig.yaml:/opt/clusters/prod002/kubeconfig.yaml:/opt/clusters/prod003/kubeconfig.yaml:/opt/clusters/prod004/kubeconfig.yaml: no such file or directory
```

we can refer to the implementation of `kubectl`, use package `k8s.io/client-go/tools/clientcmd` to automatically check the `KUBECONFIG` environment variable and automatically merge the kubeconfigs.
Here is the [implementation](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/clientcmd/merged_client_builder.go)

After I modified the code of `vcctl` in this way， I can use `vcctl` in our private environment.
```
$ vcctl queue list
Name                     Weight  State   Inqueue Pending Running Unknown Completed
default                  1              Open    70            0                88              0                   15      
prod001-map     1              Open    0              0                0                 0                   0       
prod001-mcp     1              Open    0              4                583            0                   120
$
$ kubectl config use-context prod-002-cluster
$ vcctl queue list
Name                     Weight  State   Inqueue Pending Running  Unknown Completed
default                  1              Open    70            0                8                 0                   1      
prod002-map     1              Open    0              0                0                 0                   0       
prod002-mcp     1              Open    0              2                3                 0                   0
```

### Additional context

Am I allowed to fix this problem like this and submit my PR？

halcyon-r:
Vcctl supports merging multiple kubeconfig to support context switching among multiple k8s clusters. 
#### What type of PR is this?
/kind feature
Use the method provided by [merged_client_builder.go](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/clientcmd/merged_client_builder.go) to load client configuration information. In addition to supporting detailed kubeconfig, it also supports merging multiple kubeconfigs.
cctl can be made to access different k8s clusters by switching contexts in a multi-cloud environment.

Fixes #4291 

```release-note
N
```


halcyon-r:
@JesseStutler @lowang-bh 
Please help me review, thanks.

Monokaix:
Does other resources also need this? Like vcjob and podgroup.

halcyon-r:
> Does other resources also need this? Like vcjob and podgroup.

@Monokaix Yes, other resources except queue use the method in the `util `package to obtain kubeconfig, which is repeated with the `buildConfig` methos in the queue package. I have refactored it.

Monokaix:
Thanks! Please merge to one commit.

pkg/cli/queue/common.go
```diff
@@ -1,45 +0,0 @@
...
-)
-
-type commonFlags struct {
-	Master        string
-	Kubeconfig    string
-	SchedulerName string
-}
-
-func initFlags(cmd *cobra.Command, cf *commonFlags) {
-	cmd.Flags().StringVarP(&cf.SchedulerName, "scheduler", "", "volcano", "the scheduler for this job")
```
JesseStutler, line 35: I didn't see the flag: `cmd.Flags().StringVarP(&cf.SchedulerName, "scheduler", "", "volcano", "the scheduler for this job")` copied into the `InitFlags` method in `pkg/cli/util/util.go`, missed it?

halcyon-r: @JesseStutler I can't see what the` --scheduler` flag actually does in the code. I think this might be redundant, so I deleted it. If we really need this flag, please tell me why.

JesseStutler: The scheduler can configure the scheduler name using the command flag, so I think here also should keep this flag if the user configs a custom name for the scheduler

halcyon-r: @JesseStutler However, this flag was not used in the code currently. Even if the user specifies `--scheduler` flag, it will have no effect. Are you planning to use this flag in the future? If so, I'll add it back.

JesseStutler: OK currently I think is ok to remove it, it's not been used in any other codes

PR #4274, opened 2025-05-12
fix: add miss queue state check in allocatable action

googs1025:
#### What type of PR is this?
/kind bug
#### What this PR does / why we need it:
- add miss queue state check in `allocatble` action
#### Which issue(s) this PR fixes:

Fixes https://github.com/volcano-sh/volcano/issues/4189

follow-up this comment: https://github.com/volcano-sh/volcano/pull/4263#discussion_r2083889569

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4307, opened 2025-05-20
Prevent pod scheduling when reclaim

### LINKED ISSUE #4189: Prevent the scheduling of pods in non-open queues.

Monokaix:
#### What type of PR is this?
/kind feature
see https://github.com/volcano-sh/volcano/issues/4189
Supplementary implementation of https://github.com/volcano-sh/volcano/pull/4263 and https://github.com/volcano-sh/volcano/pull/4274
In allocate, preempt, reclaim actions, we all need to prevent pod be scheduled when queue is not open.

Fixes #4189

There is no check in `backfill` action, as there is no queue logic so far, it's too heavy to add new callback for this feature, we can assume that it's a corner case.

```release-note

```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -314,7 +314,7 @@ func (pmpt *Action) preempt(
 		klog.V(3).Infof("Preempted <%v> for Task <%s/%s> requested <%v>.",
 			preempted, preemptor.Namespace, preemptor.Name, preemptor.InitResreq)
 
-		// If preemptor's queue is overused, it means preemptor can not be allocated. So no need care about the node idle resource
+		// If preemptor's queue is not allocatable, it means preemptor cannot be allocated. So no need care about the node idle resource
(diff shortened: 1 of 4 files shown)


lowang-bh (commented):
/lgtm

hwdef (commented):
/approve

Issue #4331, opened 2025-05-30
feature: Hard Mode HyperNode based binpack scheduling  policy  needed


kingeasternsun:
### What is the problem you're trying to solve

<img width="1054" alt="Image" src="https://github.com/user-attachments/assets/818a9231-68df-48e3-9a55-d73a226dd654" />

For example , we have a DDP job that needs 4 , and each node requests 8 GPU.  The  solution2 is bettern than solution 1.

### Describe the solution you'd like

The  solution2 is better than solution 1.

### Additional context

I'd like to do the PR to support hard mode.

MondayCha:
Hi, I've also raised a similar issue: https://github.com/volcano-sh/volcano/issues/4368 . My implementation will mainly modify the batch node order, so it shouldn't conflict with yours. Would you mind taking a look when you have time?

Monokaix:
> Hi, I've also raised a similar issue: [#4368](https://github.com/volcano-sh/volcano/issues/4368) . My implementation will mainly modify the batch node order, so it shouldn't conflict with yours. Would you mind taking a look when you have time?

Is the same problem with yours?

PR #4345, opened 2025-06-05
network topology aware hard mode support hypernode binpack

kingeasternsun:
#### What type of PR is this?

hard mode support hypernode binpack.

Fixes https://github.com/volcano-sh/volcano/issues/4331

```release-note
Hard mode support hypernode binpack.
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -419,10 +419,9 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 			}
 		}
 
-		if job.IsSoftTopologyMode() {
+		if job.HasTopologyConstrain() {
 			task.JobAllocatedHyperNode = jobNewAllocatedHyperNode
 		}
-
 		bestNode, highestScore := alloc.prioritizeNodes(ssn, task, predicateNodes)
 		if bestNode == nil {
 			continue
@@ -453,20 +452,17 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
 
 // getJobNewAllocatedHyperNode Obtain the newly allocated hyperNode for the job in soft topology mode
 func getJobNewAllocatedHyperNode(ssn *framework.Session, bestNode string, job *api.JobInfo, jobAllocatedHyperNode string) string {
-	if !job.IsSoftTopologyMode() {
+	if !job.HasTopologyConstrain() {
 		return ""
 	}
-
-	jobNewAllocatedHyperNode := jobAllocatedHyperNode
 	hyperNode := util.FindHyperNodeForNode(bestNode, ssn.RealNodesList, ssn.HyperNodesTiers, ssn.HyperNodesSetByTier)
 	if hyperNode != "" {
-		if jobNewAllocatedHyperNode == "" {
-			jobNewAllocatedHyperNode = hyperNode
-		} else {
-			jobNewAllocatedHyperNode = ssn.HyperNodes.GetLCAHyperNode(hyperNode, jobNewAllocatedHyperNode)
+		if jobAllocatedHyperNode == "" {
+			return hyperNode
 		}
+		return ssn.HyperNodes.GetLCAHyperNode(hyperNode, jobAllocatedHyperNode)
 	}
-	return jobNewAllocatedHyperNode
+	return jobAllocatedHyperNode
 }
 
 // updateJobAllocatedHyperNode update job allocated hyperNode in soft topology mode
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -46,6 +46,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/plugins/binpack"
 	"volcano.sh/volcano/pkg/scheduler/plugins/drf"
 	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
+	networktopologyaware "volcano.sh/volcano/pkg/scheduler/plugins/network-topology-aware"
 	"volcano.sh/volcano/pkg/scheduler/plugins/nodeorder"
 	"volcano.sh/volcano/pkg/scheduler/plugins/predicates"
 	"volcano.sh/volcano/pkg/scheduler/plugins/proportion"
@@ -357,8 +358,9 @@ func BenchmarkAllocate(b *testing.B) {
 
 func TestAllocateWithNetWorkTopologies(t *testing.T) {
 	plugins := map[string]framework.PluginBuilder{
-		predicates.PluginName: predicates.New,
-		gang.PluginName:       gang.New,
+		predicates.PluginName:           predicates.New,
+		gang.PluginName:                 gang.New,
+		networktopologyaware.PluginName: networktopologyaware.New,
 	}
 
 	tests := []uthelper.TestCommonStruct{
@@ -1444,6 +1446,11 @@ func TestAllocateWithNetWorkTopologies(t *testing.T) {
 					Name:             predicates.PluginName,
 					EnabledPredicate: &trueValue,
 				},
+				{
+					Name:                  networktopologyaware.PluginName,
+					EnabledNodeOrder:      &trueValue,
+					EnabledHyperNodeOrder: &trueValue,
+				},
 			},
 		},
 	}
@@ -1468,6 +1475,7 @@ func TestNodeLevelScoreWithNetWorkTopologies(t *testing.T) {
 	}
 
 	tests := []uthelper.TestCommonStruct{
+
 		{
 			Name: "hard network topology constrain, allocate job to highest score hypeNode with node level binpack",
 			PodGroups: []*schedulingv1.PodGroup{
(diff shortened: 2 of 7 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -1068,6 +1068,14 @@ func (ji *JobInfo) IsSoftTopologyMode() bool {
 	return ji.PodGroup.Spec.NetworkTopology.Mode == scheduling.SoftNetworkTopologyMode
 }
 
+func (ji *JobInfo) IsTopologyMode() bool {
+	if ji.PodGroup == nil || ji.PodGroup.Spec.NetworkTopology == nil {
+		return false
+	}
+	return ji.PodGroup.Spec.NetworkTopology.Mode == scheduling.SoftNetworkTopologyMode ||
```
JesseStutler: I think 
```
if ji.PodGroup == nil || ji.PodGroup.Spec.NetworkTopology == nil {
		return false
	}
```
 judgment is sufficient, because the mode can only be hard/soft, which has been verified when creating hypernodes

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -453,25 +454,24 @@ func (alloc *Action) allocateResourcesForTasks(tasks *util.PriorityQueue, job *a
...
 		}
+		return ssn.HyperNodes.GetLCAHyperNode(hyperNode, jobNewAllocatedHyperNode)
 	}
 	return jobNewAllocatedHyperNode
 }
 
 // updateJobAllocatedHyperNode update job allocated hyperNode in soft topology mode
 func updateJobAllocatedHyperNode(job *api.JobInfo, jobNewAllocatedHyperNode string) {
-	if !job.IsSoftTopologyMode() {
+	if !job.IsTopologyMode() {
```
JesseStutler: There is no need to `updateJobAllocatedHyperNode` in `allocateResourcesForTasks` when the mode is hard, because when in hard mode, we will first traverse all the hypernodes first, save the statements and select the best hypernode, and  then update the allocated hypenode info in podgroup's annotation: https://github.com/volcano-sh/volcano/blob/3aa260cf38ed112a1764473a316e2f280ae4571c/pkg/scheduler/actions/allocate/allocate.go#L295-L299 . The `network-topology-aware` plugin only needs the allocated hypernode info in transacation context

pkg/scheduler/actions/allocate/allocate_test.go
```diff
@@ -1536,6 +1544,141 @@ func TestNodeLevelScoreWithNetWorkTopologies(t *testing.T) {
...
+				"s1": sets.New[string]("s1-n3", "s1-n4"),
+				"s2": sets.New[string]("s2-n5", "s2-n6"),
+				"s3": sets.New[string]("s0-n1", "s0-n2", "s1-n3", "s2-n5", "s1-n4", "s2-n6"),
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("q1", 1, nil),
+			},
+			ExpectBindsNum: 4,
+			// "s1-n3" and "s1-n4" nodes with same LCAHyperNode "s1", so when c1/p3 assigned to "s1-n3", c1/p3 will be assigned to "s1-n4"
+			ExpectBindpackStatistic: []int{2, 2},
```
JesseStutler: Can't we just reuse the `ExpectBindMap`?

pkg/scheduler/uthelper/helper.go
```diff
@@ -332,3 +336,50 @@ func (test *TestCommonStruct) CheckPipelined(caseIndex int) error {
...
+		realHyperNode[hyperNode] += 1
+	}
+
+	expectStatistic := make([]int, 0, len(realHyperNode))
+	for _, value := range realHyperNode {
+		expectStatistic = append(expectStatistic, value)
+	}
+	sort.Ints(expectStatistic)
+
+	if !slices.Equal(expectStatistic, test.ExpectBindpackStatistic) {
```
JesseStutler: `ExpectBindpackStatistic` is a typo word, can't we just reuse the `ExpectBindMap` and `CheckBind`?

pkg/scheduler/api/job_info.go
kingeasternsun: Thanks for your advice , fixed.

pkg/scheduler/actions/allocate/allocate.go
kingeasternsun: Thaks for your review, fixed.

pkg/scheduler/uthelper/helper.go
kingeasternsun: Thanks , fixed.

pkg/scheduler/actions/allocate/allocate_test.go
kingeasternsun: In this test case, all six nodes have identical resource capacities, so the first pod of a job is assigned to a random node. This leads to the job's final Leaf HyperNode being non-deterministic. We can only confirm that the job will be scheduled onto one of two Leaf HyperNodes.

pkg/scheduler/uthelper/helper.go
```diff
@@ -332,3 +336,50 @@ func (test *TestCommonStruct) CheckPipelined(caseIndex int) error {
...
+	}
+
+	expectBindNums := make([]int, 0, len(realHyperNode))
+	for _, value := range realHyperNode {
+		expectBindNums = append(expectBindNums, value)
+	}
+	sort.Ints(expectBindNums)
+
+	if !slices.Equal(expectBindNums, test.ExpectBindNumsInHyperNode) {
+		return fmt.Errorf("case %d(%s) check bind: \nwant: %v\n got: %v ", caseIndex, test.Name, expectBindNums, test.ExpectBindNumsInHyperNode)
```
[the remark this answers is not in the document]
JesseStutler: I think copilot is right @kingeasternsun, also maybe your variable `expectBindNums` is better to rename as `actualBindNums`? Otherwise it may be easy to confuse with the meaning of `test.ExpectBindNumsInHyperNode`

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -122,8 +122,7 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
 		nodeScores := make(map[string]float64)
 
 		taskJob := ssn.Jobs[task.Job]
-		hardMode, _ := taskJob.IsHardTopologyMode()
-		if hardMode {
+		if !taskJob.IsTopologyMode() {
```
JesseStutler: After thinking about it, the judgement added here was to prevent scoring twice in hard mode, but after your modification, now hard mode will scoring twice( About hardnode tier and task nums). I also discuss with @Monokaix, we think we can remove scoring logic about hardnode tier and task nums, only keep hypernode resource usage judgement in the future: https://github.com/volcano-sh/volcano/issues/4368. What do you think?

kingeasternsun: I believe the problem addressed by the current PR is distinct from the one outlined in [**issue #4368**](https://github.com/volcano-sh/volcano/issues/4368).

This PR prioritizes **performance**, aiming to concentrate a job's pods onto the fewest possible HyperNodes to enhance communication efficiency. In contrast, issue [#4368](https://github.com/volcano-sh/volcano/issues/4368) focuses on **utilization**, seeking to fill fragmented resources so that subsequent large-scale tasks have contiguous HyperNodes available for scheduling. Clearly, these two objectives are fundamentally different and often conflicting.

To my knowledge, Ascend's topology-aware scheduler plugin categorizes tasks into two types: **FillJobs**, which have lower communication performance requirements and follow a fragment-filling logic similar to issue [#4368](https://github.com/volcano-sh/volcano/issues/4368), and **high-performance Job**, which adhere to the logic implemented in the current PR.

pkg/scheduler/uthelper/helper.go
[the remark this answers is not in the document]
kingeasternsun: Thanks for your advice. Fixed.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
JesseStutler: No, I don't mean you're conflicting with #4368. I know this is two different scenarios. What I mean is that after turning off the hard mode validation here in the node scoring, both hypernode and node have the same scoring logic in hard mode (for the number of tasks in tier and hypernode), so there's no need to score the same logic twice. You can remove the scoring logic for tier and tasknum from the hypernode scoring

kingeasternsun: > No, I don't mean you're conflicting with #4368. I know this is two different scenarios. What I mean is that after turning off the hard mode validation here in the node scoring, both hypernode and node have the same scoring logic in hard mode (for the number of tasks in tier and hypernode), so there's no need to score the same logic twice. You can remove the scoring logic for tier and tasknum from the hypernode scoring

OK， got it.

JesseStutler:
Please fix the CI, thanks :)

kingeasternsun:
> Please fix the CI, thanks :)

Fixed.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware_test.go
JesseStutler, line 1: So I think the original `TestNetworkTopologyAwareNodeScore` should rename to `TestNetworkTopologyAwareNodeScore_Soft`?

JesseStutler (commented):
Should squash your commits into one, thanks
Also /cc @Monokaix

kingeasternsun:
> Should squash your commits into one, thanks Also /cc @Monokaix

Done. @Monokaix

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -158,8 +125,9 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
 		klog.V(4).Infof("networkTopologyAware node score is: %v", nodeScores)
 		return nodeScores, nil
 	}
-
-	ssn.AddHyperNodeOrderFn(nta.Name(), hyperNodeFn)
+	// Both hypernode and node have the same scoring logic in hard mode (for the number of tasks in tier and hypernode),
```
JesseStutler: I think comment it as TODO is better? Because we still need HyperNodeFn for https://github.com/volcano-sh/volcano/issues/4368

pkg/scheduler/api/job_info.go
```diff
@@ -1068,6 +1068,13 @@ func (ji *JobInfo) IsSoftTopologyMode() bool {
 	return ji.PodGroup.Spec.NetworkTopology.Mode == scheduling.SoftNetworkTopologyMode
 }
 
+func (ji *JobInfo) IsTopologyMode() bool {
```
Monokaix:
```suggestion
func (ji *JobInfo) HasTopologyConstrain() bool {
```

Monokaix:
Is HyperNode level score strategy introduced? Seems only HyperNode level score can handle this.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Monokaix: We can also handle this at Batchnodeorder func.

pkg/scheduler/api/hyper_node_info.go
```diff
@@ -111,6 +111,11 @@ func (hni *HyperNodeInfo) Tier() int {
 	return hni.tier
 }
 
+func (hni *HyperNodeInfo) SetParent(p string) *HyperNodeInfo {
```
Monokaix: Is this only used for test, if so, please move it to test file.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Monokaix: We can merge this first, and solve other cases in https://github.com/volcano-sh/volcano/issues/4368.

kingeasternsun:
> Is HyperNode level score strategy introduced? Seems only HyperNode level score can handle this.

The `nodeFn` incorporates hyperNode awareness but lacks hyperNode-level scoring strategies. As demonstrated by the new test cases, this approach effectively resolves the reported issue.

pkg/scheduler/api/hyper_node_info.go
kingeasternsun: The `parent` field is unexported, so tests in other packages cannot set it directly.  
We therefore expose this helper method (rather than keeping it in test files) to let white-box tests manipulate the field.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
kingeasternsun: done.

pkg/scheduler/api/hyper_node_info.go
Monokaix: The HyperNode Member filed has already defined the parent-child relationship, do we still need to set parent explicitly？

JesseStutler: /cc @kingeasternsun https://github.com/volcano-sh/volcano/pull/4345#discussion_r2227741924

JesseStutler:
Please squash your commits into one, thanks. Should not show up commit like `merge master into`

pkg/scheduler/api/hyper_node_info.go
Monokaix: We can use a new func `NewHyperNodeInfoWithParent` instead.

PR #4280, opened 2025-05-13
Add error msg when check hierarchical queue fail

XbaoWu:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
When encountering a hierarchical queue error, user can quickly identify which resource item fails to meet the requirements.

The current error log lacks sufficient information to pinpoint the issue:
image

After applying this patch, user will obtain the following information:
image

#### Which issue(s) this PR fixes:

_The relevant issue has not been submitted as of yet_

#### Special notes for your reviewer:
_No additional assistance information is available at present._

#### Does this PR introduce a user-facing change?

```release-note
1. More information is added to the check of hierarchical queue to aid in problem identification.
```

PR #4347, opened 2025-06-05
When some scalar resources are 0 in deserved, hierarychical queues validation can not pass

### LINKED ISSUE #4346: When some scalar resources are 0 in deserved, hierarychical queues validation can not pass
### Description

Configure the hierarchical queue, if the scalar resources is set to 0 in deserved of queue(for example, configure `hugepages-1Gi: 0` in deserved), the hierarchical queue validation will fail, and it will report: `queue <%s> deserved resources <%s> are less than the sum of its child queues' deserved resources <%s>` error

0. Set capacity plugin and set `enableHierarchy` to true
1. Create a parent queue, and set some scalar resources to 0 (For example, set `hugepages-1Gi:0`) in deserved, an example yaml:
```yaml
kind: Queue
metadata:
  name: parent-queue
spec:
  reclaimable: true
  weight: 1
  parent: root
  deserved:
    memory: "10Gi"
    ephemeral-storage: 40m
    nvidia.com/gpu: "32"
    nvidia.com/hostdev: "36"
    hugepages-1Gi: "0"
    pods: "50"
    cpu: "4"
  guarantee:
    resource:
      memory: "10Gi"
      ephemeral-storage: 40m
      nvidia.com/gpu: "32"
      nvidia.com/hostdev: "36"
      hugepages-1Gi: "0"
      pods: "50"
      cpu: "4"

```
2. Create a child queue, similarly, set some scalar resources to 0 in deserved
3. Check the logs

Then the logs in scheduler will have these error logs:
```
queue <%s> deserved resources <%s> are less than the sum of its child queues' deserved resources <%s>"
```

Hierarchical queues validation can successfully pass

v1.11.2/v1.12.1

_No response_

JesseStutler:
#### What type of PR is this?
/kind bug

Currently, during the hierarchical queue validation process, if a user sets some scalar resources in the queue's deserved with a value of 0 (for example, configuring hugepages-1Gi: 0), the hierarchical queue validation will fail. The main reason for this is:
https://github.com/volcano-sh/volcano/blob/6e2959db6b8e2b4dcd5d81594381014fa87dd9b5/pkg/scheduler/plugins/capacity/capacity.go#L694
In `helpers.Max`, if the scalar resource is 0, the copied `res` will not carry that scalar resource. For example, it will not include `hugepages-1Gi`. 
https://github.com/volcano-sh/volcano/blob/6e2959db6b8e2b4dcd5d81594381014fa87dd9b5/pkg/scheduler/api/helpers/helpers.go#L59

Later, when determining `LessPartly`, `totalDeserved` actually includes the configuration with `hugepages-1Gi:0`. When `LessPartly` is evaluated, if the right has the resource but the left does not, it returns `true`, leading to an error:
https://github.com/volcano-sh/volcano/blob/6e2959db6b8e2b4dcd5d81594381014fa87dd9b5/pkg/scheduler/plugins/capacity/capacity.go#L698-L701
To minimize the impact, it is most appropriate to change the condition in `helper.Max` to `>= 0`.

At the same time, Cherry-pick #4280, to better print the queue information and show why the Check is failing

Fixes #4346 

NONE

```release-note
User can configure scalar resource to 0 in deserved of queue
```

### DIFF
--- a/pkg/scheduler/api/helpers/helpers.go
+++ b/pkg/scheduler/api/helpers/helpers.go
@@ -56,14 +56,14 @@ func Max(l, r *api.Resource) *api.Resource {
 	res.ScalarResources = map[v1.ResourceName]float64{}
 	if l.ScalarResources != nil {
 		for lName, lQuant := range l.ScalarResources {
-			if lQuant > 0 {
+			if lQuant >= 0 {
 				res.ScalarResources[lName] = lQuant
 			}
 		}
 	}
 	if r.ScalarResources != nil {
 		for rName, rQuant := range r.ScalarResources {
-			if rQuant > 0 {
+			if rQuant >= 0 {
 				maxQuant := math.Max(rQuant, res.ScalarResources[rName])
 				res.ScalarResources[rName] = maxQuant
 			}
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -671,7 +671,8 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
 		}
 		// Check if the parent queue's capability is less than the child queue's capability
 		if attr.capability.LessPartly(childAttr.capability, api.Zero) {
-			return fmt.Errorf("queue <%s> capability is less than its child queue <%s>", attr.name, childAttr.name)
+			return fmt.Errorf("queue <%s> capability <%s> is less than its child queue <%s> capability <%s>",
+				attr.name, attr.capability, childAttr.name, childAttr.capability)
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/scheduler/api/helpers/helpers.go
```diff
@@ -56,14 +56,14 @@ func Max(l, r *api.Resource) *api.Resource {
 	res.ScalarResources = map[v1.ResourceName]float64{}
 	if l.ScalarResources != nil {
 		for lName, lQuant := range l.ScalarResources {
-			if lQuant > 0 {
+			if lQuant >= 0 {
```
Monokaix, line 59: Please add ut.

JesseStutler: I already added it, please see the test case 9

Monokaix:
backport to release1.11 and 1.12.

hajnalmt:
Can we backport these changes to earlier branches?

JesseStutler:
> Can we backport these changes to earlier branches?

Which branch do you need to backport?

hajnalmt:
2.12 would be enough I think (I think this problem I have here: https://github.com/volcano-sh/volcano/issues/4536 is solved by this on the master, because I couldn't reproduce it on the master branch)

PR #4348, opened 2025-06-06
 ✨  Optimize Volcano Scheduling for Native Kubernetes Workloads by Eliminating Real PodGroups

kingeasternsun:
#### What type of PR is this?
/area performance

Volcano currently creates real `PodGroup` objects for native Kubernetes workloads (Deployments/StatefulSets/DaemonSets) to enable scheduling. However, since these workloads don't require gang scheduling semantics, this approach introduces unnecessary overhead:  
- Redundant `PodGroup` CRD operations  
- Delayed scheduling due to PodGroup creation latency  
- Increased etcd/API server load  

With this PR, if the Volcano scheduler detects that a pod lacks a PodGroup, it will now create a **shadow PodGroup** directly in the scheduler's cache. This approach offers several benefits:

1.  **Improved Efficiency:** Eliminates the need to create actual PodGroups for pods, accelerating the time from creation to scheduling for Kubernetes native job types.
2.  **Enhanced Performance:** Removes the overhead of deleting PodGroups after pods are removed, boosting overall Volcano scheduler performance.
3.  **Reduced API Server & etcd Pressure:** Avoids redundant PodGroups, thereby alleviating pressure on the Kubernetes API server and etcd.

None

```release-note

```

### DIFF
--- a/docs/design/Dynamic-PodGroup-Provisioning-Based-on-Declarative-Demand .md
+++ b/docs/design/Dynamic-PodGroup-Provisioning-Based-on-Declarative-Demand .md
@@ -1,0 +1,89 @@
+### Optimized Design: Dynamic PodGroup Provisioning Based on Declarative Demand  
+
+#### **Background**  
+In current Volcano implementations, the `pg_controller` automatically instantiates **PodGroup** CRDs for all Pods derived from native Kubernetes workloads (Deployments, StatefulSets, Jobs, DaemonSets) to enable unified gang scheduling.  
+**Core Issue**: Not all workloads require gang scheduling semantics. Mandatory PodGroup creation imposes redundant load on the API Server and degrades cluster-wide scheduling performance.  
+
+---
+
+#### **Design Goals**  
+Introduce **declarative opt-in gang scheduling** via Kubernetes annotations, adhering to cloud-native principles of:  
+1. **On-demand resource provisioning**  
+2. **Control-plane efficiency**  
+3. **User-driven intent expression**  
+
+---
+
+### Design Specification  
+
+#### **1. Non-Gang Pods (Default Behavior)**  
+- **Trigger Condition**: Pods **lacking both** annotations:  
+  - `scheduling.k8s.io/group-name`  
+  - `scheduling.volcano.sh/group-min-member`  
+- **System Action**:  
+  - Volcano Scheduler dynamically generates an **ephemeral ShadowPodGroup** exclusively within its cache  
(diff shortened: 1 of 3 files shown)


Monokaix:
Hi, please add more background info and design detail.

Monokaix:
cc @JesseStutler

kingeasternsun:
> Hi, please add more background info and design detail.

Sorry for that. I forgot to add Draft flag in PR title.  More info and UT will added.

hwdef:
This PR is very inspiring:)

JesseStutler:
This PR is useful for users who do not need Gang Scheduling for Deployment/StatefulSet, as it can improve performance. However, in version 1.12, some users need to specify `minMember` for Deployment/StatefulSet, so I believe there are still users who need Gang Scheduling support for Deployment/StatefulSet. Additionally, PodGroup can carry some scheduling states, therefore users can know why the podgroup can't be scheduled. 
Do you currently need to improve performance of this in your scenario?

kingeasternsun:
> users who need Gang Scheduling support for Deployment/StatefulSet.

1.  In our environment, we primarily handle **AI workloads** like **vcJob** and **KubeflowJob** that explicitly require gang scheduling, alongside traditional workloads such as **Deployments** and **StatefulSets**. These traditional workloads have strict latency requirements to ensure service reliability. The current method in Volcano, which involves creating an additional PodGroup to accommodate Deployments and StatefulSets, raises concerns.

2.  I understand that some users might need gang scheduling support for Deployments/StatefulSets. Therefore, I've implemented this feature behind a **flag**, though it could also be configured as a **feature gate**. Even with this feature enabled, users who wish to provide gang scheduling capabilities for Deployments and StatefulSets simply need to add the scheduling.volcano.sh/group-name and scheduling.volcano.sh/group-min-member annotations to their respective templates.

hzxuzhonghu:
Seems a good enhancement, i am thinking maybe one step further, we may only create one default shadow podgroup in the scheduler cache

kingeasternsun:
> Seems a good enhancement, i am thinking maybe one step further, we may only create one default shadow podgroup in the scheduler cache

Thanks for your commnts,  do you have any ideas?
From my perspective, we have to create separate shadow PodGroups for different Pods; otherwise, the logic inside the scheduler cache would become inconsistent with that of other AI workloads.

hzxuzhonghu:
TBH, i donot have much bandwidth, this was my coarse idea. From the memory saving view, indeed it may need much more changes compared with the current 1:1 model

PR #4354, opened 2025-06-09
Set root capability only when user not set it

### LINKED ISSUE #4350: hierarychical queues validation can not pass although root queue capability greater than sub queue capability

My question is: 
First I have a root queue and a subQueue offline,  I manually set the root capability, so that
root queue capability > offline queue capability, but still got error `Failed to check queue's hierarchical structure, error: queue <root> capability is less than its child queue <offline>`,  and when I set offline queue capability <= root queue deserved, it is ok，why?  root queue deserved is auto calculate based on cluster resources，if sub. queue capability must be less than root queue deserved, It is not suitable on cluster auto scaling.

root queue yaml :

```
kind: Queue
metadata:
  name: root
spec:
  capability:
    cpu: 91696k
    memory: 7630840000Gi
  deserved:
    cpu: "474"
    ephemeral-storage: "17542822117376"
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    memory: "2661790711961"
    pods: "2420"
  guarantee:
    resource:
      cpu: "0"
      memory: "0"
  reclaimable: false
  weight: 100
status:
  allocated:
    cpu: "0"
    memory: "0"
```

custom subqueue offline yaml :
```
kind: Queue
metadata:
  name: offline
spec:
  capability:
    cpu: "500"
    memory: 1000Gi
  deserved:
    cpu: "1"
    memory: 1Gi
  parent: root
  reclaimable: false
  weight: 100
status:
  allocated:
    cpu: "0"
    memory: "0"
  reservation: {}
  state: Open
```

_No response_

houyuting:
Fixes #4350 

Result:

UT- when I use old code, I got:
image

UT- when I use new code, I got:
image

### DIFF
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -65,6 +65,7 @@ type Session struct {
 
 	TotalResource  *api.Resource
 	TotalGuarantee *api.Resource
+	TotalDeserved  *api.Resource
 	// PodGroupOldState contains podgroup status and annotations during schedule
 	// This should not be mutated after initiated
 	api.PodGroupOldState
@@ -151,6 +152,7 @@ func openSession(cache cache.Cache) *Session {
 
 		TotalResource:  api.EmptyResource(),
 		TotalGuarantee: api.EmptyResource(),
+		TotalDeserved:  api.EmptyResource(),
 		PodGroupOldState: api.PodGroupOldState{
 			Status:      map[api.JobID]scheduling.PodGroupStatus{},
 			Annotations: map[api.JobID]map[string]string{},
@@ -304,10 +306,10 @@ func updateQueueStatus(ssn *Session) {
 // updateRootQueueResources updates the deserved/guaranteed resource and allocated resource of the root queue
 func updateRootQueueResources(ssn *Session, allocated v1.ResourceList) {
 	rootQueue := api.QueueID("root")
-	totalResource := util.ConvertRes2ResList(ssn.TotalResource).DeepCopy()
+	totalDeserved := util.ConvertRes2ResList(ssn.TotalDeserved).DeepCopy()
 	totalGuarantee := util.ConvertRes2ResList(ssn.TotalGuarantee).DeepCopy()
 
-	if equality.Semantic.DeepEqual(ssn.Queues[rootQueue].Queue.Spec.Deserved, totalResource) &&
+	if equality.Semantic.DeepEqual(ssn.Queues[rootQueue].Queue.Spec.Deserved, totalDeserved) &&
 		equality.Semantic.DeepEqual(ssn.Queues[rootQueue].Queue.Spec.Guarantee.Resource, totalGuarantee) &&
 		equality.Semantic.DeepEqual(ssn.Queues[rootQueue].Queue.Status.Allocated, allocated) {
 		klog.V(5).Infof("Root queue deserved/guaranteed resource and allocated resource remains the same, no need to update the queue.")
@@ -321,9 +323,9 @@ func updateRootQueueResources(ssn *Session, allocated v1.ResourceList) {
 		return
 	}
 
-	if !equality.Semantic.DeepEqual(queue.Spec.Deserved, totalResource) ||
+	if !equality.Semantic.DeepEqual(queue.Spec.Deserved, totalDeserved) ||
 		!equality.Semantic.DeepEqual(queue.Spec.Guarantee.Resource, totalGuarantee) {
-		queue.Spec.Deserved = totalResource
+		queue.Spec.Deserved = totalDeserved
 		queue.Spec.Guarantee.Resource = totalGuarantee
 		queue, err = ssn.VCClient().SchedulingV1beta1().Queues().Update(context.TODO(), queue, metav1.UpdateOptions{})
 		if err != nil {
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -47,6 +47,7 @@ type capacityPlugin struct {
 	rootQueue      string
 	totalResource  *api.Resource
 	totalGuarantee *api.Resource
+	totalDeserved  *api.Resource
 
 	queueOpts map[api.QueueID]*queueAttr
 	// Arguments given for the plugin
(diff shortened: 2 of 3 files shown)


pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -598,7 +598,9 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 
 	// init root queue realCapability/capability/deserved as cp.totalResource
 	rootQueueAttr := cp.queueOpts[api.QueueID(cp.rootQueue)]
-	rootQueueAttr.capability = cp.totalResource
+	if rootQueueAttr.capability == nil || rootQueueAttr.capability.IsEmpty() {
```
JesseStutler: I think `rootQueueAttr.capability.IsEmpty()` is enough because the capability of root queue will be initialized in `newQueueAttr`, and it won't be nil

JesseStutler (commented):
Please sqush to one commit, thanks :)

pkg/scheduler/plugins/capacity/capacity.go
houyuting: Thank you for your suggestion, done.

JesseStutler:
image
@houyuting Please squash to one commit :)

JesseStutler:
/lgtm
Please also take a look @hwdef @lowang-bh @Monokaix I think it's reasonable when users want to pre-set the capability for the root queue

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -598,7 +598,9 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 
 	// init root queue realCapability/capability/deserved as cp.totalResource
 	rootQueueAttr := cp.queueOpts[api.QueueID(cp.rootQueue)]
-	rootQueueAttr.capability = cp.totalResource
+	if rootQueueAttr.capability.IsEmpty() {
+		rootQueueAttr.capability = cp.totalResource
+	}
 	rootQueueAttr.realCapability = cp.totalResource
 	rootQueueAttr.deserved = cp.totalResource
```
xu-wentao: should we handle `realCapability` and `deseverd` too? if  `realCapability` and `deseverd` not empty and less than `cp.totalResource` may occurs error ?

houyuting: I think `realCapability and deseverd == cp.totalResource ` is suitable，because we need  at least a field that identifies how many resources are actually available in the cluster and I haven't thought of any scenarios where these two fields need to be customized. Or do you have some ideas? We can discuss it together.

xu-wentao: > I think `realCapability and deseverd == cp.totalResource ` is suitable，because we need at least a field that identifies how many resources are actually available in the cluster and I haven't thought of any scenarios where these two fields need to be customized. Or do you have some ideas? We can discuss it together.

yes, i think u r right, the `realcapacity` and `deseverd` will be caculated in `checkHierarchicalQueue`.

Monokaix:
That's a good catch, and @Xu-Wentao also has the same requirement, and I think that we can add root queue at helm chart, then user can set the spec of root queue when install volcano, and we can remove root queue creation operation in volcano scheduler.

houyuting:
@Monokaix Good idea and we can open another feature issue and track it. If you have time，please give me approve, thanks.

JesseStutler:
> @Monokaix Good idea and we can open another feature issue and track it. If you have time，please give me approve, thanks.

OK @houyuting Would you like to raise this issue? Otherwise we may miss it :)

pkg/scheduler/plugins/capacity/capacity.go
Monokaix: seems that if users previously set deserved fields of each sub queue, and then deserved field check may also fail when cluster resources reduced.

houyuting: Yes, you are right, done, please review again,  thanks.

Monokaix:
Please sign off your commit with `git commit -s`.

Monokaix:
The root queue will also be updated when close session, so it's not enough to just modify plugin.

houyuting:
> The root queue will also be updated when close session, so it's not enough to just modify plugin.

got it

JesseStutler:
image
The UT failed, please fix it thanks

pkg/scheduler/framework/session.go
```diff
@@ -315,9 +317,9 @@ func updateRootQueueResources(ssn *Session, allocated v1.ResourceList) {
 		return
 	}
 
-	if !equality.Semantic.DeepEqual(queue.Spec.Deserved, totalResource) ||
+	if !equality.Semantic.DeepEqual(queue.Spec.Deserved, totalDeserved) ||
 		!equality.Semantic.DeepEqual(queue.Spec.Guarantee.Resource, totalGuarantee) {
-		queue.Spec.Deserved = totalResource
+		queue.Spec.Deserved = totalDeserved
```
JesseStutler, line 322: The queue here is a new queue and does not specify the capability, if the user sets the capability for the root queue, it needs to be set here as well

pkg/scheduler/framework/session.go
```diff
@@ -298,10 +300,10 @@ func updateQueueStatus(ssn *Session) {
 // updateRootQueueResources updates the deserved/guaranteed resource and allocated resource of the root queue
 func updateRootQueueResources(ssn *Session, allocated v1.ResourceList) {
 	rootQueue := api.QueueID("root")
-	totalResource := util.ConvertRes2ResList(ssn.TotalResource).DeepCopy()
+	totalDeserved := util.ConvertRes2ResList(ssn.TotalDeserved).DeepCopy()
 	totalGuarantee := util.ConvertRes2ResList(ssn.TotalGuarantee).DeepCopy()
 
-	if equality.Semantic.DeepEqual(ssn.Queues[rootQueue].Queue.Spec.Deserved, totalResource) &&
+	if equality.Semantic.DeepEqual(ssn.Queues[rootQueue].Queue.Spec.Deserved, totalDeserved) &&
```
JesseStutler, line 306: Need to also judge whether capability updated

JesseStutler (commented):
Please check my latest review comments and gemini's comments, thanks

go.sum
JesseStutler, line 1: I think go.mod and go.sum should not change in your pr

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -774,8 +780,14 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
 	}
 
 	if attr.name == cp.rootQueue {
-		attr.guarantee = totalGuarantee
-		cp.totalGuarantee = totalGuarantee
+		if attr.guarantee.IsEmpty() {
```
Monokaix, line 783: The root queue is created by yourself or auto created?

houyuting: auto created

go.sum
houyuting: done

pkg/scheduler/plugins/capacity/capacity.go
Monokaix: Then its guarantee and capability fields will be updated once created, so seems this can not take effect.

houyuting: root guarantee will be computed when one of sub queues guarantee update, and root capability will be updated when user reset it.

Monokaix: I mean that when the first time the root queue is created, its deserved and guarantee fields will be empty, so it will be directly be updated according to the cluster resources, and it will still check failed in the first time, until you have set these fields on root queue manually.

houyuting: I know your means we can add root queue at helm chart,  I created a new issue https://github.com/volcano-sh/volcano/issues/4496 track it, and I will complete it.  it's ok?

Monokaix:
/approve
Thanks!

PR #4359, opened 2025-06-09
fix: trim job conditions when it's too large

junzebao:
#### What type of PR is this?
/kind bug

We have a long-running inference job running with 3000 replicas, and the controller just keeps appending conditions to the status, so eventually the job object exceeds the etcd size limit, and the controller always failed at updating job status on `failed to update status of Job`.

I'm trimming the conditions array to the latest 10, but happy to re-work if this is not desired.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -174,12 +174,8 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 	job.Status.Unknown = unknown
 	job.Status.TaskStatusCount = taskStatusCount
 
-	if updateStatus != nil {
-		if updateStatus(&job.Status) {
-			job.Status.State.LastTransitionTime = metav1.Now()
-			jobCondition := newCondition(job.Status.State.Phase, &job.Status.State.LastTransitionTime)
-			job.Status.Conditions = append(job.Status.Conditions, jobCondition)
-		}
+	if updateStatus != nil && updateStatus(&job.Status) {
+		setCurrentJobCondition(job)
 	}
 
 	// Update running duration
@@ -356,7 +352,6 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 		cc.recordPodGroupEvent(job, pg)
 	}
 
-	var jobCondition batch.JobCondition
 	oldStatus := job.Status
 	if !syncTask {
 		if updateStatus != nil {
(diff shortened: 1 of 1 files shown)


JesseStutler:
I think the correct approach should be to confirm why the controller is continuously appending the conditions?

junzebao:
> I think the correct approach should be to confirm why the controller is continuously appending the conditions?

I think instead of appending the new condition [here](https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/job_controller_actions.go#L549) (and in all other similar places in `job_controller_actions.go`, we can replace the existing condition. The condition should look like something below though so that we know which one to replace:
```
conditions:
- lastTransitionTime: "2025-06-03T07:51:54Z"
  status: "True"
  reason: <job phase>
  type: JobStatus
```

JesseStutler:
> > I think the correct approach should be to confirm why the controller is continuously appending the conditions?
> 
> I think instead of appending the new condition [here](https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/job_controller_actions.go#L549) (and in all other similar places in `job_controller_actions.go`, we can replace the existing condition. The condition should look like something below though so that we know which one to replace:
> 
> ```
> conditions:
> - lastTransitionTime: "2025-06-03T07:51:54Z"
>   status: "True"
>   reason: <job phase>
>   type: JobStatus
> ```

Yes, if condition type is the same we shouldn't constantly append, would you like to refactor here? I think it's better not to append same conditions. We need to solve the problem from the source

lowang-bh:
same condition type should only exist once. Please open an issue to record it and paste some pictures for this scenes.

junzebao:
> same condition type should only exist once

Currently the job condition doesn't have a type https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/job/job_controller_actions.go#L978

Monokaix:
> same condition type should only exist once. Please open an issue to record it and paste some pictures for this scenes.

I agree.

junzebao:
Hi @Monokaix @JesseStutler , I updated the function. Can you take another look?

Monokaix:
Hi, please merge to one commit.

junzebao:
@Monokaix done. Can we merge?

JesseStutler:
I think the current job condition has no meaning. A good condition design should be the current status of conditions in different dimensions. However, jobCondition currently only carries status, which is actually jobPhase. This jobPhase can be directly retrieved from the status of vcjob. The condition has lost its meaning, and there is no controller to use the conditions. Therefore, I think if vcjob condition is needed, it should use the k8s standard `Condition` struct.

JesseStutler:
I think the current job condition has no meaning. A good condition design should be the current status of conditions in different dimensions. However, jobCondition currently only carries status, which is actually jobPhase. This jobPhase can be directly retrieved from the status of vcjob. The condition has lost its meaning, and there is no controller to use the conditions. Therefore, I think if vcjob condition is needed, it should use the k8s standard `Condition` struct.

JesseStutler:
/cc @Monokaix @lowang-bh 
I think we should redesign the job condition, but if the current plan is just to avoid job controller issues, I will agree

JesseStutler:
/cc @Monokaix @lowang-bh 
I think we should redesign the job condition, but if the current plan is just to avoid job controller issues, I will agree

Issue #4370, opened 2025-06-12
bug：proportion totalResource accuracy


LY-today:
### Description

The TotalResource calculation of voclano's proportion plug-in is risky. It has no machine status awareness mechanism and simply accumulates all total allocatable resources, which can easily lead to errors in subsequent realCapability and deserved calculations. RealCapability affects the logic of whether a pod is created when Q resources are insufficient, and deserve affects the logic of whether Q is overused and whether the pods under it can be scheduled.

![Image](https://github.com/user-attachments/assets/f85a5d8b-32a4-4fed-9e39-095409caf5f8)

### Steps to reproduce the issue

1.node is stop scheduling but TotalResource is not perceived
2.node is notReady but TotalResource is not perceived

### Describe the results you received and expected

TotalResource can perceived node realy status

### What version of Volcano are you using?

1.12.1

JesseStutler:
Good catch. It's indeed a problem.

LY-today:
> /assign

@murali1539 Please leave it to me to complete

LY-today:
> /good-first-issue

@JesseStutler 
I have submitted MR to volcano before and it has been merged. You can leave it to me to complete.

LY-today:
> > /good-first-issue
> 
> I have mentioned MR to volcano before and merged it. You can leave it to me to complete.

LY-today:
@JesseStutler Unintentionally closed, please reopen

JesseStutler:
Sorry Murali @murali1539 , it seems that @LY-today  who raised this issue wants to fix it himself. I can assign you other good issues later

LY-today:
@JesseStutler https://github.com/volcano-sh/volcano/pull/4373 please check

enaguo:
This is a good issue, I have encountered it too, I hope it can be solved as soon as possible

blazethunderstorm:
@JesseStutler can I work on this

LY-today:
> [@JesseStutler](https://github.com/JesseStutler) can I work on this

This has been merged, refer to: https://github.com/volcano-sh/volcano/pull/4373

blazethunderstorm:
@JesseStutler I commented on some issues to work on can pls tell if I can work in other issue I commented on

PR #4373, opened 2025-06-13
fix: add getNodeStatus

LY-today:
https://github.com/volcano-sh/volcano/issues/4370#issuecomment-2966670197

### DIFF
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -211,6 +211,11 @@ func openSession(cache cache.Cache) *Session {
 	ssn.NamespaceInfo = snapshot.NamespaceInfo
 	// calculate all nodes' resource only once in each schedule cycle, other plugins can clone it when need
 	for _, n := range ssn.Nodes {
+		if isNodeUnschedulable(n.Node) || isNodeNotReady(n.Node) {
+			klog.V(3).Infof("node %s is not ready or unschedulable, need to continue", n.Name)
+			continue
+		}
+
 		ssn.TotalResource.Add(n.Allocatable)
 	}
 
--- a/pkg/scheduler/framework/util.go
+++ b/pkg/scheduler/framework/util.go
@@ -262,3 +262,22 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
 	}
 	return nodes, nil
 }
+
+func isNodeUnschedulable(node *v1.Node) bool {
+	if node == nil {
+		return true
+	}
+	return node.Spec.Unschedulable
+}
+
+func isNodeNotReady(node *v1.Node) bool {
+	if node == nil {
+		return true
+	}
+	for _, cond := range node.Status.Conditions {
+		if cond.Type == v1.NodeReady {
+			return cond.Status != v1.ConditionTrue
+		}
+	}
+	return true
+}
--- a/pkg/scheduler/framework/util_test.go
+++ b/pkg/scheduler/framework/util_test.go
@@ -1,0 +1,100 @@
+package framework
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func makeNode(unschedulable bool, readyCondition v1.ConditionStatus, nodeIsNil bool) *v1.Node {
+	if nodeIsNil {
+		return nil
+	}
+	node := &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-node",
+		},
+		Spec: v1.NodeSpec{
+			Unschedulable: unschedulable,
+		},
+		Status: v1.NodeStatus{
+			Conditions: []v1.NodeCondition{
+				{
+					Type:   v1.NodeReady,
+					Status: readyCondition,
+				},
+			},
+		},
+	}
+	return node
+}
+
+func makeNodeWithoutReady() *v1.Node {
+	return &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-node-no-ready",
+		},
+		Spec: v1.NodeSpec{
+			Unschedulable: false,
+		},
+		Status: v1.NodeStatus{
+			Conditions: []v1.NodeCondition{},
+		},
+	}
+}
+
+func TestIsNodeUnschedulable(t *testing.T) {
+	tests := []struct {
+		name      string
+		input     bool
+		expected  bool
+		nodeIsNil bool
+	}{
+		{"unschedulable true", true, true, false},
+		{"unschedulable false", false, false, false},
+		{"node is nil", true, true, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			node := makeNode(tt.input, v1.ConditionTrue, tt.nodeIsNil)
+			result := isNodeUnschedulable(node)
+			if result != tt.expected {
+				t.Errorf("expected %v, got %v", tt.expected, result)
+			}
+		})
+	}
+}
+
+func TestIsNodeNotReady(t *testing.T) {
+	tests := []struct {
+		name      string
+		ready     v1.ConditionStatus
+		expected  bool
+		nodeIsNil bool
+	}{
+		{"NodeReady=True", v1.ConditionTrue, false, false},
+		{"NodeReady=False", v1.ConditionFalse, true, false},
+		{"NodeReady=Unknown", v1.ConditionUnknown, true, false},
+		{"node is nil", "", true, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			node := makeNode(false, tt.ready, false)
+			result := isNodeNotReady(node)
+			if result != tt.expected {
+				t.Errorf("expected %v, got %v", tt.expected, result)
+			}
+		})
+	}
+
+	t.Run("No NodeReady condition", func(t *testing.T) {
+		node := makeNodeWithoutReady()
+		result := isNodeNotReady(node)
+		if result != true {
+			t.Errorf("expected true when no NodeReady condition, got %v", result)
+		}
+	})
+}
--- a/pkg/scheduler/util/test_utils.go
+++ b/pkg/scheduler/util/test_utils.go
@@ -43,6 +43,9 @@ func BuildNode(name string, alloc v1.ResourceList, labels map[string]string) *v1
 		Status: v1.NodeStatus{
 			Capacity:    alloc,
 			Allocatable: alloc,
+			Conditions: []v1.NodeCondition{
+				{Type: v1.NodeReady, Status: v1.ConditionTrue},
+			},
 		},
 	}
 }


LY-today:
@hwdef please check

pkg/scheduler/framework/session.go
```diff
@@ -211,6 +212,13 @@ func openSession(cache cache.Cache) *Session {
 	ssn.NamespaceInfo = snapshot.NamespaceInfo
 	// calculate all nodes' resource only once in each schedule cycle, other plugins can clone it when need
 	for _, n := range ssn.Nodes {
+		status := getNodeStatus(n.Node)
+
+		if len(status) != 0 || (len(status) == 1 && !slices.Contains(status, string(v1.NodeReady))) {
+			klog.V(3).Infof("node %s is not ready,need continue", n.Name)
+			continue
```
googs1025: Why not directly determine the state, like: 🤔 


```go
   if slices.Contains(status, "NotReady") || slices.Contains(status, "Unknown") {
        klog.V(3).Infof("node %s is not ready, need continue", n.Name)
        continue
    }
```

Is it possible that there is a Ready when `len(status) != 0` ? Will this cause all nodes to be skipped?

LY-today: I refer to the source code implementation of k8s for node status. I think that as long as the status returned is not the only Ready state, it will affect resource scheduling, so I implement it according to this logic. The online situation seems to be the same. The elimination method requires enumeration, so it is better to make a positive judgment.


image

pkg/scheduler/framework/session.go
```diff
@@ -222,6 +230,34 @@ func openSession(cache cache.Cache) *Session {
 	return ssn
 }
 
+func getNodeStatus(obj *v1.Node) []string {
```
googs1025: If I understand correctly, we only need three cases: 🤔 
1. Is the node ready?
2. Is the node unschedulable?
3. Is the status unknown?

pkg/scheduler/framework/session.go
LY-today: > Why not directly determine the state, like: 🤔
> 
> ```go
>    if slices.Contains(status, "NotReady") || slices.Contains(status, "Unknown") {
>         klog.V(3).Infof("node %s is not ready, need continue", n.Name)
>         continue
>     }
> ```
> 
> Is it possible that there is a Ready when `len(status) != 0` ? Will this cause all nodes to be skipped?

For example, the SchedulingDisabled state is not taken into account here, and various exceptions need to be enumerated. It is better to solve them positively.

pkg/scheduler/framework/session.go
LY-today: You understand it correctly, but in order to be more accurate, I refer to the implementation of k8s. I think it is better to reuse the source code than to guess the status.

pkg/scheduler/framework/session.go
googs1025: sgtm

LY-today:
@googs1025 Do you have any other questions? It seems that my changes have a great impact on the testing of various links, and solving the testing problems seems to be the difficulty

pkg/scheduler/framework/session.go
googs1025: I just think that `[]string` doesn't really have any effect on the outside. 🤔  Can we just return the bool value directly, which is more intuitive?

LY-today: > I just think that `[]string` doesn't really have any effect on the outside. 🤔 Can we just return the bool value directly, which is more intuitive?

no problem

LY-today: done

pkg/scheduler/framework/session.go
```diff
@@ -222,6 +229,38 @@ func openSession(cache cache.Cache) *Session {
 	return ssn
 }
 
+func nodeIsNotReady(obj *v1.Node) bool {
```
googs1025: can we move to `util.go` ? 😄 
This method does not seem to be directly related to session.

googs1025:
can you squash commits  to one?

pkg/scheduler/framework/session.go
LY-today: done

LY-today:
> can you squash commits to one?


Maybe wait until ci is fixed before merging?

LY-today:
@hwdef @googs1025 @k82cn There are many unit test issues involved here, and I personally cannot accurately measure the business logic of each unit test, so if you approve of the changes in this PR, can you help me complete the unit test repair?

googs1025:
> @hwdef @googs1025 @k82cn There are many unit test issues involved here, and I personally cannot accurately measure the business logic of each unit test, so if you approve of the changes in this PR, can you help me complete the unit test repair?

Just a guess, maybe you validated the node status(ready or not). Maybe some unit tests didn't set the node status? You can look at this part

googs1025:
I still think this method is too complicated. . . What we need is the logic to filter out NotReady or unschedulable. I think this should be very simple. In fact, we don’t care about these fields in the status list at all.

pkg/scheduler/util/test_utils.go
```diff
@@ -43,6 +43,9 @@ func BuildNode(name string, alloc v1.ResourceList, labels map[string]string) *v1
 		Status: v1.NodeStatus{
 			Capacity:    alloc,
 			Allocatable: alloc,
+			Conditions: []v1.NodeCondition{
+				{Type: v1.NodeReady, Status: v1.ConditionTrue},
```
googs1025, line 47: why here we set NotReady ? 🤔

LY-today: NodeReady  not NotReady

googs1025: oh. I misread this, it should be correct.

LY-today:
> I still think this method is too complicated. . . What we need is the logic to filter out NotReady or unschedulable. I think this should be very simple. In fact, we don’t care about these fields in the status list at all.

If I analyze this PR from the perspective of implementation difficulty, I can certainly start from luxury and start from thriftiness, so that the test will not be very complicated. But I am worried that there will still be boundary issues when running it online. Referring to k8s describe node is the best practice I know so far.

When resource calculation is abnormal, if the log is consistent with the describe node output most commonly used by users, it will be easier for users to understand.

pkg/scheduler/framework/session.go
```diff
@@ -211,6 +211,11 @@ func openSession(cache cache.Cache) *Session {
 	ssn.NamespaceInfo = snapshot.NamespaceInfo
 	// calculate all nodes' resource only once in each schedule cycle, other plugins can clone it when need
 	for _, n := range ssn.Nodes {
+		if !nodeIsNotReady(n.Node) {
```
Monokaix: Here `!nodeIsNotReady` means node ready, please correct it.

pkg/scheduler/framework/util.go
```diff
@@ -262,3 +264,35 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
 	}
 	return nodes, nil
 }
+
+func nodeIsNotReady(obj *v1.Node) bool {
```
Monokaix: It's a little complicated logic.
recommended：
```
if obj.Spec.Unschedulable {
    return false
}
for _, condition := range node.condition{
     if nodeNotReady{
        return false
     }
     ...
}
```

pkg/scheduler/framework/session.go
LY-today: done

pkg/scheduler/framework/session.go
```diff
@@ -211,6 +211,11 @@ func openSession(cache cache.Cache) *Session {
...
+			klog.V(3).Infof("node %s is not ready,need continue", n.Name)
```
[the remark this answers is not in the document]
JesseStutler: I think it's better to log as `node %s is not ready/unschedulable, skip considering this node`

pkg/scheduler/framework/util.go
```diff
@@ -262,3 +264,35 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
...
+func nodeIsReady(obj *v1.Node) bool {
+	conditionMap := make(map[v1.NodeConditionType]*v1.NodeCondition)
+	NodeAllConditions := []v1.NodeConditionType{v1.NodeReady}
+	for i := range obj.Status.Conditions {
+		cond := obj.Status.Conditions[i]
+		conditionMap[cond.Type] = &cond
+	}
+
+	var status []string
+	for _, validCondition := range NodeAllConditions {
```
JesseStutler: I think this implementation is too complicated, When the number of nodes is large, there will be a large number of [] string expansion and copying

pkg/scheduler/framework/util.go
```diff
@@ -262,3 +264,35 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
...
+		}
+	}
+	if len(status) == 0 {
+		status = append(status, "Unknown")
+	}
+	if obj.Spec.Unschedulable {
+		status = append(status, "SchedulingDisabled")
+	}
+
+	if len(status) != 1 || (len(status) == 1 && !slices.Contains(status, string(v1.NodeReady))) {
```
JesseStutler: If the `NodeReady` condition is false and the status length is 1, then `nodeIsReady` will also return true, is it what we expect? We should not use `slices.Contains` to judge it

pkg/scheduler/framework/session.go
JesseStutler: I think we should learn from Kube-Scheduler's logic rather than from printer

pkg/scheduler/framework/util.go
```diff
@@ -262,3 +264,35 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
...
+		if condition, ok := conditionMap[validCondition]; ok {
+			if condition.Status == v1.ConditionTrue {
+				status = append(status, string(condition.Type))
+			} else {
+				status = append(status, "Not"+string(condition.Type))
+			}
+		}
+	}
+	if len(status) == 0 {
+		status = append(status, "Unknown")
```
jessehu: hi @LY-today,
1. node status 'Ready' condition 有3种值 'true|false|unknown'，通过这个就可以直接判断Node是否Ready。
2. Spec.Unschedulable是表明Node不可调度，并不是Node NotReady，与func nodeIsNotReady()的含义有点不符合。

pkg/scheduler/framework/util.go
LY-today: Based on everyone's opinions, I have simplified it. Thank you

pkg/scheduler/framework/util.go
LY-today: Based on everyone's opinions, I have simplified it. Thank you

pkg/scheduler/framework/util.go
LY-today: Based on everyone's opinions, I have simplified it. Thank you

pkg/scheduler/framework/util.go
LY-today: Based on everyone's opinions, I have simplified it. Thank you

LY-today:
@JesseStutler @Monokaix please，check

pkg/scheduler/framework/util.go
```diff
@@ -262,3 +262,15 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
 	}
 	return nodes, nil
 }
+
+func isNodeUnschedulable(node *v1.Node) bool {
+	return node.Spec.Unschedulable
+}
+func isNodeNotReady(node *v1.Node) bool {
```
googs1025, line 273: nit: format . add a space

LY-today: done

pkg/scheduler/framework/util_test.go
```diff
@@ -0,0 +1,93 @@
...
+					Type:   v1.NodeReady,
+					Status: readyCondition,
+				},
+			},
+		},
+	}
+	return node
+}
+
+func makeNodeWithoutReady() *v1.Node {
```
jessehu, line 33: makeNodeWithoutReady and makeNode funcs could be merged into one func with two params "unschedulable bool, readyCondition v1.ConditionStatus"

pkg/scheduler/framework/session.go
```diff
@@ -211,6 +211,11 @@ func openSession(cache cache.Cache) *Session {
 	ssn.NamespaceInfo = snapshot.NamespaceInfo
 	// calculate all nodes' resource only once in each schedule cycle, other plugins can clone it when need
 	for _, n := range ssn.Nodes {
+		if isNodeUnschedulable(n.Node) || isNodeNotReady(n.Node) {
```
jessehu, line 214: we can conside change this if condition like this and add test cases for isNodeUnavailable() 😄 :
```
    if isNodeUnavailable(n.Node) {
		klog.V(3).Infof("The node is not ready or unschedulable, so skip calculating its resources: name: %s, unschedulable: %v", n.Node.Name, n.Node.Spec.Unschedulable)
			continue
		}

func isNodeUnavailable() bool {
  return isNodeUnschedulable(n.Node) || isNodeNotReady(n.Node)
}
```

pkg/scheduler/framework/util_test.go
LY-today: cc  all

pkg/scheduler/framework/session.go
LY-today: cc  all

LY-today:
@JesseStutler @Monokaix please，check

JesseStutler:
@LY-today Please squash your commits into one, thanks

LY-today:
> @LY-today Please squash your commits into one, thanks

done, compressed into one line, the upstream master has changed, and merge is performed

pkg/scheduler/framework/util.go
```diff
@@ -262,3 +262,16 @@ func (nl *NodeLister) List() ([]*v1.Node, error) {
 	}
 	return nodes, nil
 }
+
+func isNodeUnschedulable(node *v1.Node) bool {
```
jessehu, line 266: node could be nil, so better to add a nil check here. Refer to https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/api/node_info.go#L164C3-L164C18

LY-today: done

LY-today:
@JesseStutler please check

Monokaix:
Please merge to one commit.

LY-today:
> Please merge to one commit.

@Monokaix done

JesseStutler (commented):
/lgtm

PR #4379, opened 2025-06-17
Update volcano author header based on kube-batch

JesseStutler:
#### What type of PR is this?
/kind documentation

As we all know, Volcano is the successor of the archived project kube-batch. It was initiated by the same maintainers of kube-batch. Volcano has made many iterations and modifications based on Kube-batch, but some files are still the Kubernetes Author. We also need to add Volcano author for these files.

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -1,6 +1,6 @@
 /*
 Copyright 2017 The Kubernetes Authors.
-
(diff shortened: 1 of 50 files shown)


JesseStutler:
> why some time range end with 2021? eg. 2018-2021

@lowang-bh  I used `git blame` to modify the file based on the last modification record of the file. However, according to @kevin-wangzefeng 's suggestion, I also need to briefly describe the modifications to the file. So I'm still working on this

Issue #4383, opened 2025-06-19
When pod bind/preBind fails, the pod is pending while the podgroup is running


JesseStutler:
### Description

When pod bind/preBind fails, the pod has been resynced, but the pg remains in running state, which may confuse users:
![Image](https://github.com/user-attachments/assets/d81d897f-f3f0-472f-9225-69722100cbc6)

### Steps to reproduce the issue

1. Create a storageclass, specify a non-exist provisioner
2. Create a PVC manually
3. Create a vcjob requests the PVC

### Describe the results you received and expected

The pod keeps in pending state because the storageclass refers to a non-exist provisioner, then prebind will fail in scheduler, but the pg is already running( Pg will be refreshed to running state when session is closed, but bind/prebind is executed in the cache).
The pg should be refreshed to running state when the pod is already bound( judge in the cache)
### What version of Volcano are you using?

v1.12.1

PR #4384, opened 2025-06-19
Only refresh podgroup to running when pods are scheduled

JesseStutler:
#### What type of PR is this?
/kind bug

When pod bind/preBind fails, the pod has been resynced, but the pg remains in running state, which may confuse users.Now only the number of scheduled pods( Bound/Running) exceed the minAvailable will refresh the podgroup to Running, allocated/binding in session currently will not be counted

Fixes # https://github.com/volcano-sh/volcano/issues/4383

```release-note
Now only the number of scheduled pods(Bound/Running) exceed the minAvailable will refresh the podgroup to Running, allocated/binding pods recorded in session currently will not be counted
```

### DIFF
--- a/pkg/scheduler/api/helpers.go
+++ b/pkg/scheduler/api/helpers.go
@@ -90,6 +90,16 @@ func CompletedStatus(status TaskStatus) bool {
 	}
 }
 
+// ScheduledStatus checks whether the tasks are running, bound, failed or succeeded
+func ScheduledStatus(status TaskStatus) bool {
+	switch status {
+	case Running, Bound, Failed, Succeeded:
+		return true
+	default:
+		return false
+	}
+}
+
 // MergeErrors is used to merge multiple errors into single error
 func MergeErrors(errs ...error) error {
 	msg := "errors: "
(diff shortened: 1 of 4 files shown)


JesseStutler:
After modification, if prebind fails, then podgroup will keep in Inqueue state and will not be refreshed to Running state:
image

pkg/scheduler/api/helpers.go
```diff
@@ -90,6 +90,16 @@ func CompletedStatus(status TaskStatus) bool {
 	}
 }
 
+// ScheduledStatus checks whether the tasks are running, bound, failed or succeeded
+func ScheduledStatus(status TaskStatus) bool {
+	switch status {
+	case Running, Bound, Failed, Succeeded:
```
lowang-bh, line 96: how about Binding?

JesseStutler: PreBind is performed in Cache, and Binding is refreshed in Session. If Binding is included, PodGroup will be refreshed to Running in session before PreBind is executed. Even if PreBind fails, PodGroup will not be refreshed back to Pending or Inqueue.

PR #4293, opened 2025-05-16
Check queue spec to ensure the rationality of resource size

XbaoWu:
#### What type of PR is this?
/kind bug
#### What this PR does / why we need it:
Check the size relationship of queue resource items.

1. Let the problem be exposed to the user earlier, rather than discovered when the task is submitted.
2. Prevent the existence of unreasonable queues.
3. Facilitate the user to understand the correct configuration of the queue.
#### Which issue(s) this PR fixes:

Fixes # https://github.com/volcano-sh/volcano/issues/4292

#### Special notes for your reviewer:
_No more help info now._

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

Issue #4292, opened 2025-05-16
Queue resources lack reasonable inspection


XbaoWu:
### Description

When configuring capability, guarantee, and guarantee values for queues, there is a lack of checking for size relationships.

### Steps to reproduce the issue

1. Create the following queues (kubectl apply -f example-queues.yaml) :
```yaml
# example-queues.yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: test
spec:
  reclaimable: true
  parent: root
  capability:
    cpu: 4
    memory: 4Gi 
  deserved:
    cpu: 8
    memory: 8Gi
```

### Describe the results you received and expected

Add checks during queue creation to ensure capability >= deserved >= guarantee.

### What version of Volcano are you using?

1.11

### Any other relevant information

_Irrelevant to the environment._

XbaoWu:
/close
fixed by https://github.com/volcano-sh/volcano/pull/4293

PR #4387, opened 2025-06-19
Add queue valid function

XbaoWu:
#### What type of PR is this?
/kind feature

Check the size relationship of queue resource items, expose the unreasonable resource configuration of the queue.

Fixes # https://github.com/volcano-sh/volcano/issues/4292

Reduces the dependence on webhook by bringing the check logic of PR 4293 into the vc-schedule
- https://github.com/volcano-sh/volcano/pull/4293

```release-note
Added check on queue resource configuration
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -138,6 +138,11 @@ func (alloc *Action) allocateResources(queues *util.PriorityQueue, jobsMap map[a
 
 		queue := queues.Pop().(*api.QueueInfo)
 
+		if !ssn.QueueValid(queue) {
+			klog.V(3).Infof("Queue <%s> is invaild, ignore it.", queue.Name)
+			continue
+		}
+
 		if ssn.Overused(queue) {
 			klog.V(3).Infof("Queue <%s> is overused, ignore it.", queue.Name)
 			continue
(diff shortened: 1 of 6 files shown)


JesseStutler:
Feature only merge into master branch, thanks
BTW, now we don't need to check it in webhook right? But in release-1.11 it already has been merged into webhook

XbaoWu:
> Feature only merge into master branch, thanks BTW, now we don't need to check it in webhook right? But in release-1.11 it already has been merged into webhook

OK, I didn 't know this before, I will resubmit a PR to master later. 
the webhook is to prevent incorrect queue creation when the queue is created, but in order to avoid webhook too heavy ( https://github.com/volcano-sh/volcano/issues/4127), and may follow the trend of webhook offline, so I moved this part of the logic to the scheduler. 
I think webhook can mention the role of intervention in the scenario of small-scale nodes ( less impact on scheduling performance ) and queue creation is a low-frequency operation. What do you think ? If it is necessary to remove,  I'd also be happy to submit a PR later to revert the logic from the webhook (in release-1.11).

XbaoWu:
\close
The PR is submitted to master branch: https://github.com/volcano-sh/volcano/pull/4388
If you have more suggestions and discussions on this PR, you are also welcome to comment under PR-4388 :)

PR #4388, opened 2025-06-20
Add queue valid check

XbaoWu:
#### What type of PR is this?
/kind feature

Check the size relationship of queue resource items, expose the unreasonable resource configuration of the queue.

Fixes # https://github.com/volcano-sh/volcano/issues/4292

Reduces the dependence on webhook by bringing the check logic of PR 4293 into the vc-schedule
- https://github.com/volcano-sh/volcano/pull/4293

```release-note
Added check on queue resource configuration
```

### DIFF
--- a/pkg/scheduler/util/test_utils.go
+++ b/pkg/scheduler/util/test_utils.go
@@ -420,10 +420,11 @@ func BuildQueueWithAnnos(qname string, weight int32, cap v1.ResourceList, annos 
 	return queue
 }
 
-// BuildQueueWithResourcesQuantity return a queue with deserved and capability resources quantity.
-func BuildQueueWithResourcesQuantity(qname string, deserved, cap v1.ResourceList) *schedulingv1beta1.Queue {
+// BuildQueueWithResourcesQuantity return a queue with guarantee, deserved and capability resources quantity.
+func BuildQueueWithResourcesQuantity(qname string, deserved, cap, guarantee v1.ResourceList) *schedulingv1beta1.Queue {
 	queue := BuildQueue(qname, 1, cap)
 	queue.Spec.Deserved = deserved
+	queue.Spec.Guarantee = schedulingv1beta1.Guarantee{Resource: guarantee}
 	return queue
 }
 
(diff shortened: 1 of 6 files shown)


pkg/scheduler/util/test_utils.go
```diff
@@ -420,10 +420,11 @@ func BuildQueueWithAnnos(qname string, weight int32, cap v1.ResourceList, annos
 	return queue
 }
 
-// BuildQueueWithResourcesQuantity return a queue with deserved and capability resources quantity.
-func BuildQueueWithResourcesQuantity(qname string, deserved, cap v1.ResourceList) *schedulingv1beta1.Queue {
+// BuildQueueWithResourcesQuantity return a queue with guarantee, deserved and capability resources quantity.
+func BuildQueueWithResourcesQuantity(qname string, deserved, cap v1.ResourceList, guarantee schedulingv1beta1.Guarantee) *schedulingv1beta1.Queue {
```
lowang-bh: we can use same type of `v1.ResourceList` and assign it to queue.Spec.Guarantee

XbaoWu: done

Monokaix:
capacity and proportion plugin's `OpenSession` will also compute capability and guarante resources, this method seems can not cover all cases.

XbaoWu:
> capacity and proportion plugin's `OpenSession` will also compute capability and guarante resources, this method seems can not cover all cases.

Yes, there is indeed a lack of coverage of `OpenSession` (in capacity and proportion). The current PR only considers the check of the queue during the scheduling process, and does not consider the impact of invalid queues on other queue resource calculations, which is not good enough. 
I then modify this part of the code, check and mark the queue configuration results in `OpenSession`, and skip invalid queues in enqueue and allocate.

XbaoWu:
/close

After discussion, we agreed to use webhook first for the following reasons :
1. webhook ( [#4470](https://github.com/volcano-sh/volcano/pull/4470) ) can be exposed to the user earlier, more intuitive.
2. The range of this PR modification is large, and it is not easy to control the potential risks compared with the benefits. It can be left to discuss the more appropriate way when there are subsequent related problems.

PR #4396, opened 2025-06-23
fix: rolling restart admission webhooks after helm upgrade

junzebao:
#### What type of PR is this?

fix

When we're creating VolcanoJobs, we've seen errors like below

> error occurred: failed calling webhook \"mutatejob.volcano.sh\": failed to call webhook: Post \"https://volcano-admission-service.volcano-system.svc:443/jobs/mutate?timeout=10s\": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"volcano-admission-service.volcano-system.svc\"

I think the reason is as follows:
Every time the volcano helm chart gets upgraded, it generates a new tls certificate and the secret is [updated with the new certificate](https://github.com/volcano-sh/volcano/blob/v1.12.1/installer/dockerfile/webhook-manager/gen-admission-secret.sh#L123-L124). When we're running the admission webhook with 5 replicas and let's say at `t0` they're started with the initial certificate `c0`, a helm upgrade at `t1` would result in a new certificate `c1` in the secret. So any newly started pod (e.g., `p4`) fetches the new certificate `c1` while the rest pods `p0-p3` are still running with the old certificate `c0`.

Long story short I think the admission webhook deployment needs a restart after the certificate is updated, however the certificate is only generated in the `admission-init` job, so I can't annotate the pods with the certificate checksum.

```release-note
NONE
```

### DIFF
--- a/installer/helm/chart/volcano/templates/admission.yaml
+++ b/installer/helm/chart/volcano/templates/admission.yaml
@@ -100,6 +100,8 @@ spec:
       app: volcano-admission
   template:
     metadata:
+      annotations:
+        rollme/helm-revision: {{ .Release.Revision }}
       labels:
         app: volcano-admission
         {{- if or (.Values.custom.admission_podLabels) (.Values.custom.common_labels) }}


JesseStutler:
You mean you're doing helm upgrade at the same time as you deploy vcjob?

junzebao:
I meant every time you do a `helm upgrade`, the `admission-init` Job updates the secret that contains the webhook certificate, but the pods aren't restarted to use the new certificate, so when there is any interruption to one of the pod, the new pod will use the new certificate and thus having a different one from the rest.

JesseStutler:
> I meant every time you do a `helm upgrade`, the `admission-init` Job updates the secret that contains the webhook certificate, but the pods aren't restarted to use the new certificate, so when there is any interruption to one of the pod, the new pod will use the new certificate and thus having a different one from the rest.

OK it's reasonable, please fix the code verify CI, you can use `make update-development-yaml` to upgrade all the development yaml files

junzebao:
@JesseStutler can we have this approved and merged?

JesseStutler:
@junzebao Please squash your commits into one, and then we can merge it

PR #4401, opened 2025-06-24
Skip retry count on manual resume

### LINKED ISSUE #3067: suspend/resume 3 times vcjob will directly failure
**What happened**:
Use vcjob create job, then job running, then suspend/resume 3 times, the vcjob will directly failure
vcjob can success resume

https://github.com/volcano-sh/volcano/blob/36abf1b2037516de2129c69791df486018b46ae6/pkg/controllers/job/state/aborted.go#L31-L36

https://github.com/volcano-sh/volcano/blob/36abf1b2037516de2129c69791df486018b46ae6/pkg/controllers/job/state/restarting.go#L34-L38

when resume job, RetryCount++, and RetryCount >= maxRetry, job Failed 

**Environment**:
- Volcano Version: v1.7
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

GautamBytes:
#### What type of PR is this?
Fix

This change prevents user‑initiated resume actions from counting toward the job’s retry limit. Previously, every ResumeJobAction incremented status.RetryCount, causing a job to immediately fail once it hit MaxRetry even if no real error had occurred. Now, manual resumes transition the job to Restarting without altering RetryCount.

Fixes #3067 

Only aborted.go was modified (the single-line removal of status.RetryCount++ in the ResumeJobAction handler).

```release-note
NA
```

### DIFF
--- a/pkg/controllers/job/state/aborted.go
+++ b/pkg/controllers/job/state/aborted.go
@@ -31,7 +31,6 @@ func (as *abortedState) Execute(action Action) error {
 	case v1alpha1.ResumeJobAction:
 		return KillJob(as.job, PodRetainPhaseSoft, func(status *vcbatch.JobStatus) bool {
 			status.State.Phase = vcbatch.Restarting
-			status.RetryCount++
 			return true
 		})
 	default:


JesseStutler:
Why don't we also modify aborting.go?
https://github.com/volcano-sh/volcano/blob/068b7bc3d8ecfc8274a82e4f458bbb84fac883bf/pkg/controllers/job/state/aborting.go#L34

GautamBytes:
> Why don't we also modify aborting.go?
> 
> https://github.com/volcano-sh/volcano/blob/068b7bc3d8ecfc8274a82e4f458bbb84fac883bf/pkg/controllers/job/state/aborting.go#L34

Done with the suggested changes , also update test file a bit. You can check now and tell if it is correct or needs improvement!

PR #4405, opened 2025-06-25
Use Job Execution Time as a Tie-Breaker in Reclaim Logic

### LINKED ISSUE #4175: Take job execution time length as a criteria to reclaim

I am running volcano 1.11 and I am doing reclaim tests between queues when they use more resources than they deserve.

I notice it when it decides what to reclaim, it doesn't seem to take into consideration the length of time a job has been running when deciding which job to evict during the reclamation process, when jobs are the same priority.

Typically an HPC job that has been running for longer is more important to keep it running and the inefficiency caused by a reclaim process would affect it more meaningfully.

I would propose considering the time a job has been running after priority and other considerations when deciding to reclaim or preempt a job.

### Additional context

volcano is expected to manage expensive resources, so a job that has been running for a longer period of time has inherently more value invested than a job that has been running for a shorter period of time.

GautamBytes:
#### What type of PR is this?
/kind feature

This PR enhances the scheduler's reclaim logic to consider the execution duration of a job. Currently, when selecting a victim from jobs of the same priority, the scheduler doesn't account for how long a job has been running. This can lead to the termination of valuable, long-running jobs.

This change introduces job execution time as a tie-breaker. When jobs have the same priority, the one that has been running for the shortest duration is selected for reclamation first, thus protecting long-running HPC workloads.

Fixes #4175 

The core logic is modified within the BuildVictimsPriorityQueue function in pkg/scheduler/framework/session_plugins.go.

The new sorting hierarchy is as follows:

- Intra-job task ordering
- Cross-queue victim ordering
- Job Priority (primary criterion)
- Job execution time (new tie-breaker)
- Job UID (final stable tie-breaker)

ScheduleStartTimestamp is used as the proxy for execution time, as it's the most accurate timestamp readily available on the JobInfo struct.

```release-note

```

### DIFF
--- a/pkg/scheduler/framework/session_plugins.go
+++ b/pkg/scheduler/framework/session_plugins.go
@@ -232,13 +232,17 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 func getJobStartTime(job *api.JobInfo) time.Time {
 	var minStart time.Time
 	for _, task := range job.Tasks {
-		// Skip tasks that haven't started
-		if task.StartTimestamp.IsZero() {
+		// Safely handle nil pointers
+		if task.Pod == nil || task.Pod.Status.StartTime == nil {
 			continue
 		}
+
+		// Get the actual start time
+		startTime := task.Pod.Status.StartTime.Time
+
 		// Find the earliest start time
-		if minStart.IsZero() || task.StartTimestamp.Before(minStart) {
-			minStart = task.StartTimestamp
+		if minStart.IsZero() || startTime.Before(minStart) {
+			minStart = startTime
 		}
 	}
 	return minStart


JesseStutler:
Great Job!
/cc

pkg/scheduler/framework/session_plugins.go
```diff
@@ -933,30 +934,70 @@ func (ssn *Session) NodeOrderReduceFn(task *api.TaskInfo, pluginNodeScoreMap map
...
+// 3) same‑queue, different‑job ➔ JobOrderFn (priority)
+// 4) tie on priority ➔ shorter actual run‑time
+// 5) final tie‑breaker ➔ UID
+func (ssn *Session) BuildVictimsPriorityQueue(
+	victims []*api.TaskInfo,
+	preemptor *api.TaskInfo,
+) *util.PriorityQueue {
+	now := time.Now()
+
+	pq := util.NewPriorityQueue(func(l, r interface{}) bool {
```
JesseStutler: Why to change all the variable names?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -933,30 +934,70 @@ func (ssn *Session) NodeOrderReduceFn(task *api.TaskInfo, pluginNodeScoreMap map
...
-		lvJob, lvJobFound := ssn.Jobs[lv.Job]
-		rvJob, rvJobFound := ssn.Jobs[rv.Job]
-		preemptorJob, preemptorJobFound := ssn.Jobs[preemptor.Job]
+		// look up jobs
+		jL, okL := ssn.Jobs[lv.Job]
+		jR, okR := ssn.Jobs[rv.Job]
+		jP, okP := ssn.Jobs[preemptor.Job]
+
+		// 2) Cross-queue eviction
+		if okL && okR && okP && jL.Queue != jR.Queue {
```
JesseStutler: Same as above, you should keep all the original logic, just adding comments is fine

pkg/scheduler/framework/session_plugins.go
```diff
@@ -933,30 +934,70 @@ func (ssn *Session) NodeOrderReduceFn(task *api.TaskInfo, pluginNodeScoreMap map
...
+		higher := ssn.JobOrderFn(jL, jR)
+		lower := ssn.JobOrderFn(jR, jL)
+		if higher != lower {
+			return !higher // evict lower-priority job first
+		}
 
-		if lvJobFound && rvJobFound && preemptorJobFound && lvJob.Queue != rvJob.Queue {
-			return ssn.VictimQueueOrderFn(ssn.Queues[lvJob.Queue], ssn.Queues[rvJob.Queue], ssn.Queues[preemptorJob.Queue])
+		// 4) Tie on priority → compare actual run-time
+		startL := jL.ScheduleStartTimestamp.Time
```
JesseStutler: You shouldn't use `ScheduleStartTimestamp` as the job's start time, it's initialized in enqueue action, this represents the time when each session starts, not the time when the job is created. There is a problem that the scheduler does not directly process vcjob but processes podgroup, so the scheduler cannot perceive when vcjob starts. Event that sometimes user don't configure enqueue action, so ScheduleStartTimestamp will not be initialized

pkg/scheduler/framework/session_plugins.go
```diff
@@ -933,30 +934,70 @@ func (ssn *Session) NodeOrderReduceFn(task *api.TaskInfo, pluginNodeScoreMap map
...
+		// 2) Cross-queue eviction
+		if okL && okR && okP && jL.Queue != jR.Queue {
+			return ssn.VictimQueueOrderFn(
+				ssn.Queues[jL.Queue],
+				ssn.Queues[jR.Queue],
+				ssn.Queues[jP.Queue],
+			)
+		}
+
+		// 3) Same-queue, different-job → compare priority
```
JesseStutler: The logic is weird, and why do we need to migrate JobOrderFn to here? Why not put the job execution time judgment in JobOrder?

GautamBytes: Did the suggested changes , would you please check now?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -227,6 +228,26 @@ func (ssn *Session) Reclaimable(reclaimer *api.TaskInfo, reclaimees []*api.TaskI
 	return victims
 }
 
+// getJobStartTime returns the earliest start time of any task in the job
+func getJobStartTime(job *api.JobInfo) time.Time {
```
JesseStutler, line 232: /cc @jjuliogallegos01
Does this meet your requirements? @GautamBytes takes the earliest running Pod under the job as the starttime. Because the scheduler cannot get the vcjob information, it may have to rely on vc-controller to attach the job information to the podgroup so that the scheduler can perceive it.

Issue #4421, opened 2025-07-02
Volcano Scheduler will panic if didn't configure predicates plugin


JesseStutler:
### Description

If user didn't configure predicates plugin, the scheduler will meet panic in nodeorder plugin: 
```
acktrace=<
        goroutine 224 [running]:
        k8s.io/apimachinery/pkg/util/runtime.logPanic({0x2dcdff8, 0x469c060}, {0x25c1940, 0x46364f0})
                /go/src/volcano.sh/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:107 +0xbc
        k8s.io/apimachinery/pkg/util/runtime.handleCrash({0x2dcdff8, 0x469c060}, {0x25c1940, 0x46364f0}, {0x469c060, 0x0, 0x100000000000000?})
                /go/src/volcano.sh/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:82 +0x5e
        k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x12?})
                /go/src/volcano.sh/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:59 +0x108
        panic({0x25c1940?, 0x46364f0?})
                /usr/local/go/src/runtime/panic.go:791 +0x132
        k8s.io/component-base/metrics.(*GaugeVec).WithLabelValuesChecked(0x0, {0xc000da0db0, 0x1, 0x1})
                /go/src/volcano.sh/volcano/vendor/k8s.io/component-base/metrics/gauge.go:140 +0x32
        k8s.io/component-base/metrics.(*GaugeVec).WithLabelValues(0x60?, {0xc000da0db0?, 0x7fd2708f35b8?, 0x60?})
                /go/src/volcano.sh/volcano/vendor/k8s.io/component-base/metrics/gauge.go:178 +0x1c
        k8s.io/kubernetes/pkg/scheduler/framework/parallelize.Parallelizer.Until({0x2dd4c58?}, {0x2dce3e8, 0x469c060}, 0x3, 0xc000d024e0, {0x2a08f6a?, 0xc000961ac0?})
                /go/src/volcano.sh/volcano/vendor/k8s.io/kubernetes/pkg/scheduler/framework/parallelize/parallelism.go:57 +0x6a
[...]

PR #4422, opened 2025-07-02
Move kube-scheduler related metrics initilization to server.go to avoid panic

JesseStutler:
#### What type of PR is this?
/kind bug

Currently, if a user doesn't configure predicates plugin and then submit a simple VC job, the scheduler will panic. We need to fix it.

Fixes # https://github.com/volcano-sh/volcano/issues/4421

I also add a unit test to test if we only specify nodeorder plugin and doesn't configure predicates plugin whether will meet panic

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto" // auto-registry collectors in default registry
+	"k8s.io/component-base/metrics"
+	k8smetrics "k8s.io/kubernetes/pkg/scheduler/metrics"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -405,7 +404,6 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 		// It is safe here to directly use the state to run plugins because we have already initialized the cycle state
 		// for each pending pod when open session and will not meet nil state
 		state := ssn.GetCycleState(task.UID)
-		metrics.Register()
```
Monokaix, line 408: What does the metrics include? Is it necessary for volcano, or volcano already has these metrics? It's exposed by kube-scheduler framework, may not useful for volcano, we should check that.

pkg/scheduler/plugins/nodeorder/nodeorder_test.go
```diff
@@ -0,0 +1,141 @@
+/*
```
Monokaix, line 1: This is not related to this pr.

JesseStutler: It's used to judge when we disable predicates plugin, only contain the nodeorder plugin, the scheduler won't panic

pkg/scheduler/plugins/predicates/predicates.go
JesseStutler: Actually we only need `Goroutines` metric in kube-scheduler, so now I init this metric(override the gloabl variable ) in `InitKubeSchedulerRelatedMetrics` and this metrics won't be registered and exported, only to avoid panic when we need to execute kube-scheduler related plugins

pkg/scheduler/metrics/metrics.go
```diff
@@ -148,6 +150,19 @@ var (
 	)
 )
 
+// InitKubeSchedulerRelatedMetrics is used to init metrics global variables in k8s.io/kubernetes/pkg/scheduler/metrics/metrics.go.
+// We don't use InitMetrics() to init all global variables because currently only "Goroutines" is required when calling kube-scheduler
+// related plugins. And there is no need to export these metrics, therefore currently initialization is enough.
+func InitKubeSchedulerRelatedMetrics() {
+	k8smetrics.Goroutines = metrics.NewGaugeVec(
+		&metrics.GaugeOpts{
```
Monokaix, line 158: Is the goroutine metrics already included in volcano metrics?

JesseStutler: I think not yet, it's used to record how many goroutines are simultaneously executing in a plugin:
image
image
Do we need to register and export it?

Monokaix:
backport to release1.12.

PR #4433, opened 2025-07-05
Add unit test for Info function 

GautamBytes:
**What type of PR is this?**
/kind test

This PR adds unit test coverage for the pkg/version package to improve code quality and prevent regressions. It validates that the Info() function correctly formats and returns the project's version information.

This test file intentionally covers only the Info() function. The PrintVersionAndExit() function is not tested because it calls os.Exit(0), which would terminate the test suite.


lowang-bh:
Hi, please use `git rebase` to rebase from master and squash all commits to only one.

Issue #4444, opened 2025-07-09
Fix the issue where SelectBestNode returns nil when plugin scores are negative, causing a panic.


guoqinwill:
### Description

When the Volcano plugin assigns negative scores, causing the SelectBestNode method to filter out all nodes and return nil, a null pointer exception occurs when saveTaskNodeScore attempts to read the node information.

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

```I0613 17:37:16.582692       7 session.go:349] Close Session 4369b219-5561-4db8-9d5d-2de56c10aa02
E0613 17:37:16.583072       7 panic.go:262] "Observed a panic" panic="runtime error: invalid memory address or nil pointer dereference" panicGoValue="\"invalid memory address or nil pointer dereference\"" stacktrace=<
        goroutine 490 [running]:
        k8s.io/apimachinery/pkg/util/runtime.logPanic({0xaaaab41ca9b8, 0xaaaab6144100}, {0xaaaab3c16800, 0xaaaab6029a80})
                k8s.io/apimachinery@v0.32.2/pkg/util/runtime/runtime.go:107 +0x98
        k8s.io/apimachinery/pkg/util/runtime.handleCrash({0xaaaab41ca9b8, 0xaaaab6144100}, {0xaaaab3c16800, 0xaaaab6029a80}, {0xaaaab6144100, 0x0, 0x4000067478?})
                k8s.io/apimachinery@v0.32.2/pkg/util/runtime/runtime.go:82 +0x60
        k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x40019eb880?})
                k8s.io/apimachinery@v0.32.2/pkg/util/runtime/runtime.go:59 +0x114
        panic({0xaaaab3c16800?, 0xaaaab6029a80?})
                runtime/panic.go:785 +0x124
        volcano.sh/apollo/pkg/scheduler/actions/allocate.(*Action).saveTaskNodeScore(0x40006013f8, 0x40031b0ea0?, 0x0, 0x4003c1f890)
                volcano.sh/apollo/pkg/scheduler/actions/allocate/allocate.go:655 +0x28
        volcano.sh/apollo/pkg/scheduler/actions/allocate.(*Action).prioritizeNodes(0x40006013f8, 0x4001fae6c8, 0x40031b0ea0, {0x400369f348, 0x4e, 0x4002f426a0?})
[...]

PR #4445, opened 2025-07-09
Fix the issue where SelectBestNode returns nil when plugin scores are…

guoqinwill:
#### What type of PR is this?
/kind bug
This PR primarily strengthens the scheduler's node selection logic by initializing the score to negative infinity. This ensures that even when plugin scores are negative, the scheduler can still return the best node based on the scores, instead of returning nil.

Fixes # https://github.com/volcano-sh/volcano/issues/4444

```release-note

```


lowang-bh:
Please add some detail describe in PR info.

JesseStutler:
/lgtm
thanks

JesseStutler:
Also need to cherry-pick it back to release-1.12 and release-1.11, using the `hack/cherry_pick_pull.sh` script. But currently the master logic is different from in v1.11, therefore we may need to cherry-pick to release-1.11 manually

PR #4448, opened 2025-07-09
Fix the issue of failed within queue preemption

### LINKED ISSUE #4447: Fix the issue where within-queue eviction of other jobs fails when cluster resources are insufficient.
### Description

The current task preemption within the queue has some issues: when the conditions for within-queue preemption are met, but the cluster resources are insufficient, `predicateNodes` becomes empty, which prevents preemption from proceeding.

```
func (pmpt *Action) preempt(
	ssn *framework.Session,
	stmt *framework.Statement,
	preemptor *api.TaskInfo,
	filter func(*api.TaskInfo) bool,
	predicateHelper util.PredicateHelper,
) (bool, error) {
	// Check whether the task is eligible to preempt others, e.g., check preemptionPolicy is `Never` or not
	if err := pmpt.taskEligibleToPreempt(preemptor); err != nil {
		return false, err
	}

	if err := ssn.PrePredicateFn(preemptor); err != nil {
		return false, fmt.Errorf("PrePredicate for task %s/%s failed for: %v", preemptor.Namespace, preemptor.Name, err)
	}

	allNodes := ssn.FilterOutUnschedulableAndUnresolvableNodesForTask(preemptor)
	predicateNodes, _ := predicateHelper.PredicateNodes(preemptor, allNodes, ssn.PredicateForPreemptAction, pmpt.enablePredicateErrorCache)


```

1.
2.
3.

If the scheduling requirement can be satisfied by evicting tasks on a single node, the preemption should succeed.

master

_No response_

zhifei92:
#### What type of PR is this?
/kind bug
The current task preemption within the queue has some issues: when the conditions for within-queue preemption are met, but the cluster resources are insufficient, `predicateNodes` becomes empty, which prevents preemption from proceeding.

```
func (pmpt *Action) preempt(
	ssn *framework.Session,
	stmt *framework.Statement,
	preemptor *api.TaskInfo,
	filter func(*api.TaskInfo) bool,
	predicateHelper util.PredicateHelper,
) (bool, error) {
	// Check whether the task is eligible to preempt others, e.g., check preemptionPolicy is `Never` or not
	if err := pmpt.taskEligibleToPreempt(preemptor); err != nil {
		return false, err
	}

	if err := ssn.PrePredicateFn(preemptor); err != nil {
		return false, fmt.Errorf("PrePredicate for task %s/%s failed for: %v", preemptor.Namespace, preemptor.Name, err)
	}

	allNodes := ssn.FilterOutUnschedulableAndUnresolvableNodesForTask(preemptor)
	predicateNodes, _ := predicateHelper.PredicateNodes(preemptor, allNodes, ssn.PredicateForPreemptAction, pmpt.enablePredicateErrorCache)


```

Fixes #4447

```release-note
None
```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -289,41 +289,53 @@ func (pmpt *Action) preempt(
 		return pmpt.topologyAwarePreempt(ssn, stmt, preemptor, filter, predicateNodes)
 	}
 
-	return pmpt.normalPreempt(ssn, stmt, preemptor, filter, predicateNodes)
+	targetNodes := predicateNodes
+	needPredicateCheck := false
+	if len(predicateNodes) == 0 {
+		klog.Infof("No predicateNodes, using all nodes for preemption. Task <%s/%s>",
+			preemptor.Namespace, preemptor.Name)
+		targetNodes = allNodes
+		needPredicateCheck = true
+	}
+
+	return pmpt.executePreemption(
+		ssn,
+		stmt,
+		preemptor,
+		filter,
+		targetNodes,
+		needPredicateCheck,
+	)
 }
 
-func (pmpt *Action) normalPreempt(
+func (pmpt *Action) executePreemption(
 	ssn *framework.Session,
 	stmt *framework.Statement,
 	preemptor *api.TaskInfo,
 	filter func(*api.TaskInfo) bool,
-	predicateNodes []*api.NodeInfo,
+	targetNodes []*api.NodeInfo,
+	needPredicateCheck bool,
 ) (bool, error) {
-	nodeScores := util.PrioritizeNodes(preemptor, predicateNodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
-
-	selectedNodes := util.SortNodes(nodeScores)
-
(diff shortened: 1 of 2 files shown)


pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -289,41 +289,53 @@ func (pmpt *Action) preempt(
 		return pmpt.topologyAwarePreempt(ssn, stmt, preemptor, filter, predicateNodes)
 	}
 
-	return pmpt.normalPreempt(ssn, stmt, preemptor, filter, predicateNodes)
+	targetNodes := predicateNodes
+	needPredicateCheck := false
+	if len(predicateNodes) == 0 {
```
JesseStutler, line 294: `PredicateNodes` not only checks whether resources are sufficient, but also checks such as Volume, Image, NodeAffinity, PodAffinity and other plugins. If all nodes' resources are not enough and we simply preempt all nodes here, what is the meaning of the existence of `PredicateNodes`? I think the correct approach should be to separate resource verification from Volume, Image, NodeAffinity, PodAffinity and other verifications, rather than coupling them in `PredicateNodes`.

zhifei92: Thank you for your reply. I agree with your point and will work on improving this.

JesseStutler:
Also, please don't forget to sign your commit using `git commit -s` :)

PR #4450, opened 2025-07-10
Fix the bug of job repeated preemption.

zhifei92:
#### What type of PR is this?
/kind bug
If task has already preempted other tasks and entered the `pipeline` state, the NominatedNodeName field of the Pod will be set to the name of the scheduled node (this NominatedNodeName has already been filled in `closeSession`). When creating the `ssn` in each scheduling round, it's important to set the task status as `Pipelined` based on the Pod's `NominatedNodeName`. This is crucial because if this step is skipped, during the current `ssn` scheduling, a task might complete preemption and enter the `Pipelined` state, but in the next scheduling round, the task may not have actually been scheduled onto a node yet, leaving its status in the `ssn` as Pending, which could lead to repeated preemption.

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/scheduler/api/helpers.go
+++ b/pkg/scheduler/api/helpers.go
@@ -48,6 +48,19 @@ func getTaskStatus(pod *v1.Pod) TaskStatus {
 		if len(pod.Spec.NodeName) == 0 {
 			return Pending
 		}
+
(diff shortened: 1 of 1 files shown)


lowang-bh:
/hold

NominatedNodeName can be unavilable anymore in new scheduling cycle.

Issue #4244, opened 2025-04-29
Support Scarce Resource Avoidance Plugin scheduling


XbaoWu:
### What is the problem you're trying to solve

Support tasks that avoid scheduling on nodes with scarce resources.
Users can assign weights to different scarce resources within the cluster to prevent certain tasks from being scheduled on nodes with limited or scarce resources.

### Describe the solution you'd like

1. Users can define resource weights to represent the importance of resources.
2. During the scheduling process, the task will be influenced by the assigned resource weights, so as to avoid scheduling to the nodes with scarce resources as much as possible.

Prevent task starvation for those requesting critical resources, enhance the utilization of important resources, and thereby achieve more effective task scheduling strategies.

### Additional context

_No more Additional context provided._

JesseStutler:
Nice feature, would you willing to share the feature on volcano weekly meeting?

XbaoWu:
> Nice feature, would you willing to share the feature on volcano weekly meeting?

Thank you for your recognition. I’d love to sharing it.

XbaoWu:
According to the predicates plugin mentioned at the weekly meeting on May 9, it seems that it can 't solve the problems encountered in this scenario.

**Pre-information**：
The core idea of predicates is to reserve memory and CPU resources for GPU nodes according to the number of GPUs.

**The reasons are as follows** : 
1. **The resource proportion is not easy to determine**.If the cluster contains GPU tasks with various resource specifications, it is not esay to pre-determine the proportion in the predicates plugin (eg. Task A need 1:2:4, while Task B need 1:4:4. Here, the ratios represent the proportion of GPUs to CPUs to memory, respectively.).
2. **The scenario cannot be fully covered**. The goal here is to prevent certain tasks from being scheduled on nodes with limited or scarce resources, rather than outright rejecting them. This is because even if the node has scarce resources, there may be no tasks that require such resources for a time, so we allow the CPU resources on the node to be used for other jobs that do not have available CPU resources ( at this time, the CPU resources of other nodes in the cluster may be insufficient ).
3. **sra provides more flexibility and extensibility**. SRA is not only limited to GPU resources and can support other subsequent resource forms. It also supports resource mutual exclusion between different GPU tasks. For example, tasks requiring only T4 resources will be scheduled to T4-only nodes, rather than nodes with both T4 and A10. Correspondingly, Task2 will prioritize NodeB over NodeD (refer to the _Additional Context_ section in the issue description.).

**Reference information**：
[...]

PR #4454, opened 2025-07-13
Add sra policy for ResourceStrategyFit Plugin

XbaoWu:
#### What type of PR is this?
/kind feature

Prevent task starvation for those requesting critical resources, enhance the utilization of important resources, and thereby achieve more effective task scheduling strategies.
To keep the `predicates` plugin clean, the processing logic of `predicates.proportional` is moved to the `resource-strategy-fit` plugin.

Fixes  https://github.com/volcano-sh/volcano/issues/4244

_No additional assistance messages_

```release-note
1. Support users to avoid scarce resource nodes when scheduling tasks that do not require scarce resources.
2. Move the processing logic of `predicates.proportional` to the `resource-strategy-fit `plugin and adjust the previous configuration.
```

### DIFF
--- a/docs/design/proportional.md
+++ b/docs/design/proportional.md
@@ -29,14 +29,16 @@ Firstly set the proportion binding in volcano-scheduler.conf:
 actions: "enqueue, allocate, backfill"
 tiers:
 - plugins:
-  - name: predicates
+  - name: resource-strategy-fit
     arguments:
-      predicate.ProportionalEnable: true
-      predicate.resources: nvidia.com/gpu,nvidia.com/v100-sxm2-16gb
-      predicate.resources.nvidia.com/gpu.cpu: 8
-      predicate.resources.nvidia.com/gpu.memory: 8
-      predicate.resources.nvidia.com/v100-sxm2-16gb.cpu: 16
-      predicate.resources.nvidia.com/v100-sxm2-16gb.memory: 16
+      sra:
+        policy: proportional
+        resources: nvidia.com/gpu,nvidia.com/v100-sxm2-16gb
+        proportional:
+          nvidia.com/gpu.cpu: 8
+          nvidia.com/gpu.memory: 8
+          nvidia.com/v100-sxm2-16gb.cpu: 16
+          nvidia.com/v100-sxm2-16gb.memory: 16
 ```
 
 The proportion is GPU:CPU:MEMORY=1:8:8, and let the test scenario just as above:
@@ -50,8 +52,8 @@ Job | Pod | Resource | Node | NodeAllocatable | NodeIdle
 default/single-1000-0 | single-1000-0 | cpu 8, memory 8G, nvidia.com/gpu 0 | nodeC0-0 | cpu 74, memory 128G, nvidia.com/gpu 8 | cpu 66, memory 120G, nvidia.com/gpu 8 |
 default/single-1000-1 | single-1000-1 | cpu 8, memory 8G, nvidia.com/gpu 0 | - | - | - |
 
-After job single-1000-0 is scheduled, the Idel resouce is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin caculates the resource left if job single-1000-1 is scheduled`(node.Idel.CPU - task.Resreq.CPU < node.Idel.GPU * cpuRatio ||
-node.Idel.Memory - task.Resreq.Memory < node.Idel.GPU * memoryRatio)`; the result is 8GPUs, 58CPUs, 112G memory, that unsatisfies the 1:8:8 proportion. Therefore nodeC0-0 is removed from the predicateNodes, and NodeIdle remains 8GPUs, 66CPUs, 120G memory. 
+After job single-1000-0 is scheduled, the Idle resource is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin calculates the resource left if job single-1000-1 is scheduled`(node.Idle.CPU - task.Resreq.CPU < node.Idle.GPU * cpuRatio ||
+node.Idle.Memory - task.Resreq.Memory < node.Idle.GPU * memoryRatio)`; the result is 8GPUs, 58CPUs, 112G memory, that unsatisfied the 1:8:8 proportion. Therefore, nodeC0-0 is removed from the predicateNodes, and NodeIdle remains 8GPUs, 66CPUs, 120G memory.
 
 
 
--- a/docs/design/resource-strategy-fit-scheduling.md
+++ b/docs/design/resource-strategy-fit-scheduling.md
@@ -2,15 +2,23 @@
 
 ## Summary
 
-The native k8s ResourceStrategyFit plug-in can only adopt one type of strategy for all resources, such as MostRequestedPriority and LeastRequestedPriority. However, in industrial practice, this design is not applicable in some scenarios. For example: in AI scenarios, we usually disperse CPU tasks in CPU machine groups to reduce hot spots. GPU tasks are gathered in GPU machine groups to reduce GPU fragmentation. Therefore, we need to expand a scheduling strategy to meet the needs of this scenario.
+At present, there are few scheduling strategies for resource types. When users want to schedule according to resource types in some special scenarios, there is a lack of refined strategies to choose from. This plugin attempts to improve this part of the strategy to cover more application scenarios for users.
 
 ## Motivation
 
-- Different resource types can be configured with different aggregation or dispersion strategies, and weights can be used to distinguish priorities
+- **resourceStrategyFit**: The native k8s NodeResourcesFit plug-in can only adopt one type of strategy for all resources, such as MostRequestedPriority and LeastRequestedPriority. However, in industrial practice, this design is not applicable in some scenarios. For example: in AI scenarios, we usually disperse CPU tasks in CPU machine groups to reduce hot spots. GPU tasks are gathered in GPU machine groups to reduce GPU fragmentation. Therefore, we need to expand a scheduling strategy to meet the needs of this scenario.
+
+- **sra.retention**: In a cluster where GPU nodes and CPU nodes are deployed together, there is a risk that a task submitted to the cluster, which does not require GPU resources, may be scheduled to a GPU node. This can lead to inefficient resource utilization, as a job requiring both CPU and GPU resources might be pending due to insufficient CPU resources on the GPU node. This scenario is not ideal, as it can result in under utilization of resources and potential delays in job execution. Therefore, sra is proposed to avoid nodes with critical resources (such as GPUs) for certain CPU jobs and to improve overall resource utilization.
+
+- **sra.proportional**: In a cluster where CPU tasks and GPU tasks are mixed, sometimes some CPU tasks are scheduled to GPU nodes, resulting in some GPU tasks being in a long-term Pending state due to lack of CPU resources. In this regard, we may specify a 'primary' resource(e.g., GPU in deep learning), and preserve the amount of associated 'secondary' resources by a pre-set proportion. This policy will play a role in the predicate stage, and is committed to ensuring that the idle resources of the node will meet the secondary resources required by the scare resources of the node according to the proportion.
 
 ### Goals
 
-- Different types of resources can be configured with different strategies to prioritize them in the form of weights
+- Different types of resources can be configured with different strategies to prioritize them in the form of weights.
+
+- Different resources can be configured with different weights, which are used to indicate the scarcity of resources.
+
+- Different resources can set the proportion of secondary resources to ensure that the idle secondary resources on the node are sufficient.
 
 ### Non-Goals
 
@@ -18,34 +26,43 @@ The native k8s ResourceStrategyFit plug-in can only adopt one type of strategy f
 
 ## Proposal
 
-Extend one plug-ins to meet the above needs
+Extend one plugin (ResourceStrategyFit) to meet the above needs
+- **ResourceStrategyFit**: Provide users with **MostAllocated** and **LeastAllocated** these two ways to facilitate the user to decide whether the task should be distributed or aggregated according to the different needs of the task.
 
-- ResourceStrategyFit
+- **sra.retention**: Provide users the way of weight to avoid common task being scheduled to the presence of scare resource nodes, thereby improving the utilization of scare resources.
+
+- **sra.proportional**: Provide users with the way of proportion to avoid common task being scheduled to the presence of scare resource nodes, thereby improving the utilization of scare resources.
 
 ## User Story
 
 ### Story1
 - Users expect different resource allocation strategies to be applied based on resource types. For example, in PyTorch jobs, the master pod (which uses CPU) should be distributed to avoid node hotspots, while worker pods (which use GPU) should be aggregated to minimize resource fragmentation.
 
+### Story2
+- Users expect that there should be no ordinary tasks on scarce resource nodes. For example, in a cluster where CPU tasks and GPU tasks are mixed, CPU tasks should be avoided being scheduled to GPU nodes.
+
+### Story3
+- Users expect to ensure that the idle secondary resources on the nodes are sufficient after setting the proportion of secondary resources for scare resources.
+
 ## Design Details
 
 ### ResourceStrategyFit
 
 config：
-```
-actions: "enqueue, allocate, backfill, reclaim, preempt"
-tiers:
-- plugins:
- - name: resource-strategy-fit
-    arguments:
-      resourceStrategyFitWeight: 10
-      resources:
-        nvidia.com/gpu:
-          type: MostAllocated
-          weight: 2
-        cpu:
-          type: LeastAllocated
-          weight: 1
+```yaml
+  actions: "enqueue, allocate, backfill, reclaim, preempt"
+  tiers:
+  - plugins:
+    - name: resource-strategy-fit
+      arguments:
+        resourceStrategyFitWeight: 10
+        resources:
+          nvidia.com/gpu:
+            type: MostAllocated
+            weight: 2
+          cpu:
+            type: LeastAllocated
+            weight: 1
 ```
 config description：
 
@@ -71,8 +88,137 @@ node score:
 ```
 finalScoreNode = [(weight1 * resource1) + (weight2 * resource2) + … + (weightN* resourceN)] /(weight1+weight2+ … +weightN)
 ```
+### Scarce Resource Avoidance (SRA)
+#### Policy
+At present, the following two policies are supported:
+- `retention`: Give each scarce resource a weight. Higher weight means more scarce. Jobs that don’t need the scarce resource will try to stay off nodes that have it, so the scarce resource can be retained.
+- `proportional`: By specifying 'primary' scarce resources (e.g. GPU in deep learning) and preserve the amount of associated 'secondary' resources by a pre-set proportion.
+
+#### Solution
+##### retention
+1. In retention policy, arguments `sra.resources` is provided to configure important resources in the cluster.
+2. Based on the significance of different resources, `sra.retention.[ResourceName]` can assign varying weights for resource allocation. A higher weight signifies greater importance, and tasks that do not require this resource will, to the extent possible, be scheduled away from nodes with such resources
+3. For all tasks, user can set `sra.resources` and `sra.retention.[ResourceName]` field in `resource-strategy-fit` arguments via `volcano-scheduler-configmap` in following format:
+
+```yaml
+  actions: "enqueue, reclaim, allocate, backfill, preempt"
+  tiers:
+  - plugins:
+    - name: resource-strategy-fit
+      arguments:
+        sra.policy: retention
+        sra.resources: nvidia.com/t4, nvidia.com/a10
+        sra.retention.weight: 2
+        sra.retention.nvidia.com/t4: 1
+        sra.retention.nvidia.com/a10: 1
+```
+
+4. `retention` policy will affect the score result of node order. The higher the node score is, the higher the priority is. The score result of retention policy is calculated as follows :
+```
+node resources items don't meet task request, retentionScore = 0
+node resources items meet task request,       retentionScore = MaxNodeScore * retentionWeight * (weight_1 + weight_2 + ··· + weight_n) / weightSum
+```
+> `retentionScore`: the node score of retention policy. \
+> `MaxNodeScore`: the maximum score of node. default is 100. \
+> `retentionWeight`: the weight of retention policy. \
+> `weightSum`: the sum of all retention resources weights. \
+> `weight_x`: the weight of the xth retention resource that does not exist on the node.
+
+   1. Now, we have tasks:
+
+      | Task Name  | Task request resource                                        |
+      |------------|--------------------------------------------------------------|
+      | cpu-task-0 | `{cpu: 2, memory: 4Gi}`                                      |
+      | gpu-task-0 | `{cpu: 2, memory: 4Gi, nvidia.com/t4: 2}`                    |
+      | gpu-task-1 | `{cpu: 2, memory: 4Gi, nvidia.com/t4: 1, nvidia.com/a10: 2}` | 
+   
+   2. Suppose there are 3 nodes available in cluster:
+
+      | Node Name | Resource capacity on node                                       |
+      |-----------|-----------------------------------------------------------------|
+      | node1     | `{cpu: 32, memory: 64Gi}`                                       | 
+      | node2     | `{cpu: 16, memory: 32Gi, nvidia.com/t4: 10}`                    | 
+      | node3     | `{cpu: 16, memory: 32Gi, nvidia.com/t4: 5, nvidia.com/a10: 10}` |
+
+   3. Through the retention policy we will get the following results:
+   
+      | Task       | Node  | Score result (retention) | Notes                                                        |
+      |------------|-------|--------------------------|--------------------------------------------------------------|
+      | cpu-task-0 | node1 | 200                      | node resources meet the task request and no scarce resources |
+      | cpu-task-0 | node2 | 100                      | node resources meet the task request but have t4             |
+      | cpu-task-0 | node3 | 0                        | node resources meet the task request but have t4, a10        |
+      | gpu-task-0 | node1 | 0                        | node resources don't meet the task request                   |
+      | gpu-task-0 | node2 | 100                      | node resources meet the task request and have t4             |
+      | gpu-task-0 | node3 | 0                        | node resources meet the task request but have a10            |
+      | gpu-task-1 | node1 | 0                        | node resources don't meet the task request                   |
+      | gpu-task-1 | node2 | 0                        | node resources don't meet the task request                   |
+      | gpu-task-1 | node3 | 0                        | node resources meet the task request and have t4, a10        |
+
+##### proportional
+
+![](./images/proportional-diagram.png)
+
+Firstly set the proportion binding in volcano-scheduler.conf:
+
+```yaml
+  actions: "enqueue, reclaim, allocate, backfill, preempt"
+  tiers:
+  - plugins:
+    - name: resource-strategy-fit
+      arguments:
+        sra:
+          policy: proportional
+          resources: nvidia.com/gpu,nvidia.com/v100-sxm2-16gb
+          proportional:
+            nvidia.com/gpu.cpu: 8
+            nvidia.com/gpu.memory: 8
+            nvidia.com/v100-sxm2-16gb.cpu: 16
+            nvidia.com/v100-sxm2-16gb.memory: 16
+```
+
+The proportion is GPU:CPU:MEMORY=1:8:8, and let the test scenario just as above:
+
+| Node     | NodeAllocatable                       | NodeIdle                              |
+|----------|---------------------------------------|---------------------------------------|
+| nodeC0-0 | cpu 74, memory 128G, nvidia.com/gpu 8 | cpu 74, memory 128G, nvidia.com/gpu 8 |
+
+| Job                   | Pod           | Resource                           | Node     | NodeAllocatable                       | NodeIdle                              |
+|-----------------------|---------------|------------------------------------|----------|---------------------------------------|---------------------------------------|
+| default/single-1000-0 | single-1000-0 | cpu 8, memory 8G, nvidia.com/gpu 0 | nodeC0-0 | cpu 74, memory 128G, nvidia.com/gpu 8 | cpu 66, memory 120G, nvidia.com/gpu 8 |
+| default/single-1000-1 | single-1000-1 | cpu 8, memory 8G, nvidia.com/gpu 0 | -        | -                                     | -                                     |
+
+After job single-1000-0 is scheduled, the Idle resource is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin calculates the resource left if job single-1000-1 is scheduled`(node.Idle.CPU - task.Resreq.CPU < node.Idle.GPU * cpuRatio ||
+node.Idle.Memory - task.Resreq.Memory < node.Idle.GPU * memoryRatio)`; the result is 8GPUs, 58CPUs, 112G memory, that unsatisfied the 1:8:8 proportion. Therefore, nodeC0-0 is removed from the predicateNodes, and NodeIdle remains 8GPUs, 66CPUs, 120G memory.
+
+For more details, please refer to: [proportional design](./proportional.md)
 
 ## Alternatives
 
 ### Binpack VS ResourceStrategyFit
 If you want to use the clustering strategy for all resource types, you can choose the Binpack plugin. If you need to configure different clustering or scattering strategies for different resource types, you can choose the ResourceStrategyFit plugin. ResourceStrategyFit can also achieve the same results as Binpack by adjusting configuration parameters.
+
+## Best Practices
+### AI scenario
+In some AI scenarios, CPU tasks are usually dispersed into CPU machine groups to reduce hot spots. GPU tasks are gathered in the GPU machine group to reduce GPU fragmentation. At the same time, it is also necessary to avoid the situation that CPU tasks are assigned to GPU nodes, resulting in long-term waiting of GPU tasks due to insufficient CPU or MEM resources of nodes. In this scenario, we can combine **resourceStrategyFit** and **sra policy** to deal with this scenario. The corresponding example configuration is as follows:
+
+```yaml
+  actions: "enqueue, allocate, backfill, reclaim, preempt"
+  tiers:
+    - plugins:
+      - name: resource-strategy-fit
+        arguments:
+          resourceStrategyFitWeight: 10
+          resources:
+            nvidia.com/gpu:
+              type: MostAllocated
+              weight: 2
+            cpu:
+              type: LeastAllocated
+              weight: 1
+          sra:
+            policy: retention
+            resources: nvidia.com/gpu
+            retention:
+              weight: 10
+              nvidia.com/gpu: 1
+```
--- a/pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
+++ b/pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
@@ -19,6 +19,7 @@ package resourcestrategyfit
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
@@ -27,12 +28,20 @@ import (
 
 	"volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/plugins/util"
 )
 
 const (
 	// PluginName indicates name of volcano scheduler plugin.
-	PluginName                             = "resource-strategy-fit"
+	PluginName = "resource-strategy-fit"
+	// DefaultResourceStrategyFitPluginWeight is the default weight of resourceStrategyFit plugin
 	DefaultResourceStrategyFitPluginWeight = 10
+	// SraPolicy is the key for providing SRA policy
+	SraPolicy = "sra.policy"
+	// SraRetentionWeight is the key for providing retention policy Weight in YAML
+	SraRetentionWeight = "sra.retention.weight"
+	// resourceFmt is the format for resource key
+	resourceFmt = "%s[%d]"
 )
 
 type ResourceStrategyFit struct {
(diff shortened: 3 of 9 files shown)


pkg/scheduler/plugins/resource-strategy-fit/retention.go
```diff
@@ -0,0 +1,149 @@
...
+	return weight
+}
+
+// retentionScore use the best fit polices during scheduling.
+// Goals:
+// - Schedule Tasks using BestFit Policy using retention policy
+// - Improve the utilization of scarce resources on the cluster
+func retentionScore(task *api.TaskInfo, node *api.NodeInfo, retention retentionWeight) float64 {
+	requested := task.Resreq
+	capacity := node.Capacity
```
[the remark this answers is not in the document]
XbaoWu: It's not a problem

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -216,5 +326,24 @@ func leastRequestedScore(requested float64, used float64, capacity float64, weig
...
+// - status (*api.Status): The status object to evaluate.
+//
+// Returns:
+// - bool: True if the status code matches any of the abort conditions; false otherwise.
+func ShouldAbort(status *api.Status) bool {
+	return status.Code == api.UnschedulableAndUnresolvable ||
+		status.Code == api.Error ||
+		status.Code == api.Wait ||
+		status.Code == api.Skip
+}
```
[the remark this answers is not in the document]
XbaoWu: done

docs/design/sra-scheduling.md
```diff
@@ -0,0 +1,61 @@
...
+      |------------|-------|--------------------------|--------------------------------------------------------------|
+      | cpu-task-0 | node1 | 200                      | node resources meet the task request and no scarce resources |
+      | cpu-task-0 | node2 | 100                      | node resources meet the task request and have t4             |
+      | cpu-task-0 | node3 | 0                        | node resources meet the task request and have t4, a10        |
+      | gpu-task-0 | node1 | 0                        | node resources don't meet the task requirements              |
+      | gpu-task-0 | node2 | 100                      | node resources meet the task request and have t4             |
+      | gpu-task-0 | node3 | 0                        | node resources meet the task request and have t4, a10        |
+      | gpu-task-1 | node1 | 0                        | node resources don't meet the task requirements              |
+      | gpu-task-1 | node2 | 0                        | node resources don't meet the task requirements              |
+      | gpu-task-1 | node3 | 0                        | node resources meet the task request and have t4, a10        |
```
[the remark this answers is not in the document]
XbaoWu: done

docs/design/sra-scheduling.md
```diff
@@ -0,0 +1,61 @@
+# Scarce Resource Avoidance (SRA)
+
+## Introduction
+In a cluster where GPU nodes and CPU nodes are deployed together, there is a risk that a task submitted to the cluster, which does not require GPU resources, may be scheduled to a GPU node. This can lead to inefficient resource utilization, as a job requiring both CPU and GPU resources might be pending due to insufficient CPU resources on the GPU node. This scenario is not ideal, as it can result in underutilization of resources and potential delays in job execution. Therefore, sra is proposed to avoid nodes with critical resources (such as GPUs) for certain CPU jobs and to improve overall resource utilization.
+
+## Policy
+At present, the following two policies are supported:
+- `retention`: Give each scare resource a weight. Higher weight means more scare. Jobs that don’t need the scare resource will try to stay off nodes that have it, so the scare resource can be retained.
+- `proportional`: By specifying 'primary' scare resources (e.g. GPU in deep learning) and preserve the amount of associated 'secondary' resources by a pre-set proportion.
```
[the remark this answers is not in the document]
XbaoWu: done

docs/design/sra-scheduling.md
```diff
@@ -0,0 +1,61 @@
+# Scarce Resource Avoidance (SRA)
+
+## Introduction
+In a cluster where GPU nodes and CPU nodes are deployed together, there is a risk that a task submitted to the cluster, which does not require GPU resources, may be scheduled to a GPU node. This can lead to inefficient resource utilization, as a job requiring both CPU and GPU resources might be pending due to insufficient CPU resources on the GPU node. This scenario is not ideal, as it can result in underutilization of resources and potential delays in job execution. Therefore, sra is proposed to avoid nodes with critical resources (such as GPUs) for certain CPU jobs and to improve overall resource utilization.
+
+## Policy
+At present, the following two policies are supported:
+- `retention`: Give each scare resource a weight. Higher weight means more scare. Jobs that don’t need the scare resource will try to stay off nodes that have it, so the scare resource can be retained.
```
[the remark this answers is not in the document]
XbaoWu: done

docs/design/proportional.md
```diff
@@ -50,8 +50,8 @@ Job | Pod | Resource | Node | NodeAllocatable | NodeIdle
...
 default/single-1000-1 | single-1000-1 | cpu 8, memory 8G, nvidia.com/gpu 0 | - | - | - |
 
-After job single-1000-0 is scheduled, the Idel resouce is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin caculates the resource left if job single-1000-1 is scheduled`(node.Idel.CPU - task.Resreq.CPU < node.Idel.GPU * cpuRatio ||
-node.Idel.Memory - task.Resreq.Memory < node.Idel.GPU * memoryRatio)`; the result is 8GPUs, 58CPUs, 112G memory, that unsatisfies the 1:8:8 proportion. Therefore nodeC0-0 is removed from the predicateNodes, and NodeIdle remains 8GPUs, 66CPUs, 120G memory. 
+After job single-1000-0 is scheduled, the Ideal resource is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin calculates the resource left if job single-1000-1 is scheduled`(node.Idel.CPU - task.Resreq.CPU < node.Idel.GPU * cpuRatio ||
+node.Idel.Memory - task.Resreq.Memory < node.Idel.GPU * memoryRatio)`; the result is 8GPUs, 58CPUs, 112G memory, that unsatisfied the 1:8:8 proportion. Therefore, nodeC0-0 is removed from the predicateNodes, and NodeIdle remains 8GPUs, 66CPUs, 120G memory.
```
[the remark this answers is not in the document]
XbaoWu: done

XbaoWu:
@JesseStutler @Monokaix @LY-today sra related content has been adjusted, when your time is convenient, please take a look at it, thank you :)

docs/design/sra-scheduling.md
JesseStutler, line 1: Can we put the contents of sra-scheduling doc into `docs/design/resource-strategy-fit-scheduling.md`? Because sra and proportional are now part of the `ResourceStrategyFit` plugin, it is not very intuitive to write them separately. By integrating it into `docs/design/resource-strategy-fit-scheduling.md`, users will be able to understand all the features of this plugin more intuitively

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -18,7 +18,7 @@ The native k8s ResourceStrategyFit plug-in can only adopt one type of strategy f
 
 ## Proposal
 
-Extend one plug-ins to meet the above needs
```
JesseStutler, line 21: We can add proportional/sra into this doc, and rich the content of this plugin's Summary/Motivation

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -116,6 +144,32 @@ func calculateWeight(args framework.Arguments) ResourceStrategyFit {
...
+func applySraPolicy(rsf *resourceStrategyFitPlugin, args framework.Arguments) {
+	// Obtain sra resource key name
+	resourcesStr, ok := args[SraResource].(string)
+	if !ok {
+		resourcesStr = ""
+	}
+	resources := strings.Split(resourcesStr, ",")
+
+	sraPolicy, ok := args[SraPolicy].(string)
+	if !ok {
```
JesseStutler: Use switch case is better

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -125,15 +179,73 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		score := 0.0
+
+		if rsf.weight.ResourceStrategyFitWeight > 0 {
+			score += Score(task, node, rsf.weight)
+			klog.V(4).Infof("resourceStrategyFit score for Task %s/%s on node %s is: %v", task.Namespace, task.Name, node.Name, score)
+		}
+
+		if rsf.sraPolicy == RetentionPolicy && rsf.retention.Weight > 0 {
+			klog.V(5).Infof("sra.retention weight is %d", rsf.retention.Weight)
+			if klog.V(4).Enabled() {
```
JesseStutler, line 229: I think L209-L228 should be part of predicateFn rather than nodeorderFn?

pkg/scheduler/plugins/resource-strategy-fit/retention.go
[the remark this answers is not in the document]
JesseStutler: @XbaoWu I think gemini is correct, Allocatable is the resource that the scheduler can schedule. Usually, Allocatable = Capacity, but it is not ruled out that there may be reserved resources on some nodes, Allocatable < Capacity

docs/design/sra-scheduling.md
```diff
@@ -0,0 +1,61 @@
...
+   
+   2. Suppose there are 3 nodes available in cluster:
+
+      | Node Name | Resource capacity on node                                       |
+      |-----------|-----------------------------------------------------------------|
+      | node1     | `{cpu: 32, memory: 64Gi}`                                       | 
+      | node2     | `{cpu: 16, memory: 32Gi, nvidia.com/t4: 10}`                    | 
+      | node3     | `{cpu: 16, memory: 32Gi, nvidia.com/t4: 5, nvidia.com/a10: 10}` |
+
+    Through the retention policy we will get the following results:
```
JesseStutler: Could you add the detail of rules for scoring?

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -125,15 +179,73 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+				}
+
+				if len(notFoundResource) > 0 {
+					klog.V(4).Infof("resources [%s] record in predicate.resources but not found on any node",
+						strings.Join(notFoundResource, ", "))
+				}
+			}
+
+			rs := retentionScore(task, node, rsf.retention)
+			klog.V(4).Infof("sra.retentionScore for Task %s/%s on node %s is: %v", task.Namespace, task.Name, node.Name, rs)
```
JesseStutler: `sra.retentionScore` -> `sra.retention Score`, add a space

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -125,15 +179,73 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+	}
+
 	nodeOrderFn := func(task *api.TaskInfo, node *api.NodeInfo) (float64, error) {
-		score := Score(task, node, rsf.weight)
-		klog.V(4).Infof("resourceStrategyFit score for Task %s/%s on node %s is: %v", task.Namespace, task.Name, node.Name, score)
+		score := 0.0
+
+		if rsf.weight.ResourceStrategyFitWeight > 0 {
+			score += Score(task, node, rsf.weight)
+			klog.V(4).Infof("resourceStrategyFit score for Task %s/%s on node %s is: %v", task.Namespace, task.Name, node.Name, score)
```
JesseStutler: The plugin name is `ResourceStrategyFit`, and here the log also reports that `resourceStrategyFit score for Task`, maybe we should rename the strategy specified by @LY-today here. Otherwise, it looks like the log here is the total score of this plugin

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -75,4 +75,28 @@ finalScoreNode = [(weight1 * resource1) + (weight2 * resource2) + … + (weightN
...
+        arguments:
+          resourceStrategyFitWeight: 10
+          resources:
+            nvidia.com/gpu:
+              type: MostAllocated
+              weight: 2
+            cpu:
+              type: LeastAllocated
+              weight: 1
+          sra.policy: retention
```
JesseStutler: This configuration seems a bit fragmented. SRA defines weights and resource types. The original definition by @LY-today also has a set of weights and resource types. Can this be reused?

docs/design/sra-scheduling.md
XbaoWu: This is a good idea. I ' ll merge the design documents later.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: Good, I ' ll add this later.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Because `sra.retention` is intended to be designed as a "soft" policy to keep CPU tasks away from GPU nodes as much as possible, it is placed in nodeOrderFn. This is different from the "hard" policy of `sra.proportion`

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Good, this will make it look clearer

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: done

docs/design/sra-scheduling.md
XbaoWu: done

docs/design/sra-scheduling.md
XbaoWu: done

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: done

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: done

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -116,6 +144,37 @@ func calculateWeight(args framework.Arguments) ResourceStrategyFit {
 	return weight
 }
 
+func applySraPolicy(rsf *resourceStrategyFitPlugin, args framework.Arguments) {
+	// Obtain sra resource key name
+	resourcesStr, ok := args[SraResource].(string)
```
JesseStutler: We can use `GetString` instead: https://github.com/volcano-sh/volcano/blob/aba5c5598c875fee7e489de4c923b653087b95d0/pkg/scheduler/framework/arguments.go#L100-L117.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -125,15 +184,73 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+							break
+						}
+					}
+					if !found {
+						notFoundResource = append(notFoundResource, string(resource))
+					}
+				}
+
+				if len(notFoundResource) > 0 {
+					klog.V(4).Infof("resources [%s] record in predicate.resources but not found on any node",
```
JesseStutler: `predicate.resources`? I can't clearly understand the meaning of this log. Missed changing it?

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
JesseStutler: Yes I know. So this part is only for log?
```
				var notFoundResource []string
				for resource := range rsf.retention.Resources {
					found := false
					for _, nodeInfo := range ssn.Nodes {
						if nodeInfo.Capacity.Get(resource) > 0 {
							found = true
							break
						}
					}
					if !found {
						notFoundResource = append(notFoundResource, string(resource))
					}
				}

				if len(notFoundResource) > 0 {
					klog.V(4).Infof("resources [%s] record in predicate.resources but not found on any node",
						strings.Join(notFoundResource, ", "))
				}
			}
```

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -2,50 +2,62 @@
...
+
+- **sra.proportional**: In a cluster where CPU tasks and GPU tasks are mixed, sometimes some CPU tasks are scheduled to GPU nodes, resulting in some GPU tasks being in a long-term Pending state due to lack of CPU resources. In this regard, we may specify a 'primary' resource(e.g., GPU in deep learning), and preserve the amount of associated 'secondary' resources by a pre-set proportion. This policy will play a role in the predicate stage, and is committed to ensuring that the idle resources of the node will meet the secondary resources required by the scare resources of the node according to the proportion.
 
 ### Goals
 
-- Different types of resources can be configured with different strategies to prioritize them in the form of weights
+- Different types of resources can be configured with different strategies to prioritize them in the form of weights.
+- Different resources can be configured with different weights, which are used to indicate the scare degree of resources.
```
JesseStutler: `scare degree` -> scarcity

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -2,50 +2,62 @@
...
 
 - None.
 
 ## Proposal
 
-Extend one plug-ins to meet the above needs
-
-- ResourceStrategyFit
+Extend one plugin (ResourceStrategyFit) to meet the above needs
+- **ResourceStrategyFit**: Provide users with **MostAllocated** and **LeastAllocated** these two ways to facilitate the user to decide whether the task should be distributed or aggregated according to the different needs of the task.
```
JesseStutler, line 30: I think this is part of `ResourceStrategyFit` plugin, not representing the whole plugin, maybe rename as `Per-Resource Scoring Strategy`? What do you think?

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -2,50 +2,62 @@
...
+- **SraPolicy**: Provide users with two ways of weight and proportion to avoid common task being scheduled to the presence of scare resource nodes, thereby improving the utilization of scare resources.
```
JesseStutler: Also missed proportional here :)

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Yes, it is used to check whether there are configured resources on the current cluster node, so as to facilitate the scenario that the policy is not effective when the user configures the scare resource name incorrectly.

pkg/scheduler/plugins/resource-strategy-fit/retention.go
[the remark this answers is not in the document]
XbaoWu: > @XbaoWu I think gemini is correct, Allocatable is the resource that the scheduler can schedule. Usually, Allocatable = Capacity, but it is not ruled out that there may be reserved resources on some nodes, Allocatable < Capacity

Okay, I 'll modify it.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: I'll adjust this part of the log output content

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Looks good, I ' ll use this.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Sorry this is a problem, it should be `sra.resources`

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: `scarcity` is indeed better, I'll modified this.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: Ok, I 'll add it.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: done

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: done

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: done

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: done

pkg/scheduler/plugins/resource-strategy-fit/retention.go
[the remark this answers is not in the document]
XbaoWu: done

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: I think it 's OK, or use `NodeResourcesFit` to be consistent with kube-scheduler. What do you think? @LY-today

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: This plugin name here can refer to the above discussion results, and then make a unified adjustment to ut and design.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: I tend to unify the style of configuration items into the style of `binpack`, as if other plugins are the same. What do you think？ @LY-today

XbaoWu: The definition of resources I suggest separate better, the reasons are as follows : 
1. Different semantic uses. The resource defined by `sra` is used to avoid, and `resourceStrategyFit` is used to determine whether to aggregate or disperse. 
2. Plugin structure. Because there are two strategies in a plugin at present, different strategies adjust the scheduling according to the resources, there are independent use scenarios, and the resource items may not be completely consistent. It is not like a scheduling strategy that applies a superposition effect around the same resource. 
3. Usage. When the user decides to turn off the `sra` policy, it is also necessary to pay attention to whether the `resourceStrategyFit` resource group refers to this part of the resource, which may be somewhat inconvenient.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
LY-today: I personally recommend a unified plugin name, using parameters and log names to distinguish [different resource types Dispersion/Gathering]/[Scarce Resource Avoidance] strategies @XbaoWu @JesseStutler

docs/design/resource-strategy-fit-scheduling.md
LY-today: Between the architecture design of the scheduler framework and the best practices of koordinator, I personally recommend using struct form to parse parameters, because we have incubated simulated scheduler components in our internal scenarios in the past, which is more friendly to struct.

LY-today: Of course, I am willing to accept the official standardized process of Volcano, so there is no mandatory requirement for struct here. Do you have any suggestions?
 @JesseStutler

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: I think this way can also be, the main purpose is to avoid the user on the log content in the understanding of ambiguity.

However, it also needs to be slightly adjusted in the design document, because the name of this policy（`ResourceStrategyFit`） is the same as the name of the plugin 😂, so do we need to adjust it slightly to make it look like a part of the plugin?

docs/design/resource-strategy-fit-scheduling.md
JesseStutler: Volcano does not have a unified format, but since you have already deployed it in an internal cluster, we can unify it into struct @LY-today

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
JesseStutler: We can keep the plugin name `ResourceStrategyFit` because @LY-today have already deployed the plugin in their internal clusters, but I recommend changing the log because it may confuse users

docs/design/resource-strategy-fit-scheduling.md
LY-today: get

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Well, I 'll make some adjustments.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: done

XbaoWu:
@JesseStutler @LY-today The log content is slightly adjusted, and the sra's configuration items is also modified to the struct structure. 
If it 's convenient, please help to see if there are still some parts that need to be improved or other ideas and suggestions for the modification, I will modify it as soon as possible.

pkg/scheduler/plugins/util/util.go
```diff
@@ -137,3 +137,22 @@ func GetInqueueResource(job *api.JobInfo, allocated *api.Resource) *api.Resource
...
+// - Error: Represents an error state that prevents further execution.
+// - Wait: Suggests that the process should pause and not proceed further.
+// - Skip: Indicates that the operation should be skipped entirely.
+//
+// Parameters:
+// - status (*api.Status): The status object to evaluate.
+//
+// Returns:
+// - bool: True if the status code matches any of the abort conditions; false otherwise.
+func ShouldAbort(status *api.Status) bool {
```
Monokaix, line 154: There is already a same func in pkg/scheduler/plugins/predicates/predicates.go.

docs/design/proportional.md
```diff
@@ -29,14 +29,16 @@ Firstly set the proportion binding in volcano-scheduler.conf:
...
-  - name: predicates
+  - name: resource-strategy-fit
     arguments:
-      predicate.ProportionalEnable: true
-      predicate.resources: nvidia.com/gpu,nvidia.com/v100-sxm2-16gb
-      predicate.resources.nvidia.com/gpu.cpu: 8
-      predicate.resources.nvidia.com/gpu.memory: 8
-      predicate.resources.nvidia.com/v100-sxm2-16gb.cpu: 16
-      predicate.resources.nvidia.com/v100-sxm2-16gb.memory: 16
+      sra:
```
Monokaix: Do we still need the sra statement? Seems it's enough to have a retention and proportional policy, or it will increase user understanding cost.

Monokaix:
IMO, the sra semantics should not be present in plugin configuration, keep a single resourcefit concept is ok, and make the configuration as simpler as possible.

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -2,50 +2,67 @@
...
+- **resourceStrategyFit**: The native k8s NodeResourcesFit plug-in can only adopt one type of strategy for all resources, such as MostRequestedPriority and LeastRequestedPriority. However, in industrial practice, this design is not applicable in some scenarios. For example: in AI scenarios, we usually disperse CPU tasks in CPU machine groups to reduce hot spots. GPU tasks are gathered in GPU machine groups to reduce GPU fragmentation. Therefore, we need to expand a scheduling strategy to meet the needs of this scenario.
+
+- **sra.retention**: In a cluster where GPU nodes and CPU nodes are deployed together, there is a risk that a task submitted to the cluster, which does not require GPU resources, may be scheduled to a GPU node. This can lead to inefficient resource utilization, as a job requiring both CPU and GPU resources might be pending due to insufficient CPU resources on the GPU node. This scenario is not ideal, as it can result in under utilization of resources and potential delays in job execution. Therefore, sra is proposed to avoid nodes with critical resources (such as GPUs) for certain CPU jobs and to improve overall resource utilization.
```
Monokaix: Better explain why called `retention` and the difference between `proportional`.

pkg/scheduler/plugins/util/util.go
XbaoWu: Well, I will modify it

docs/design/proportional.md
XbaoWu: Because the original idea was to expand different avoidance strategies based on sra (scare resource avoidance), including proportion and weight. So add a layer of sra semantics.

XbaoWu:
> IMO, the sra semantics should not be present in plugin configuration, keep a single resourcefit concept is ok, and make the configuration as simpler as possible.
My original idea may not be quite right. I ' ll re-arrange the configuration here later to make it simpler.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: OK, this is a good supplement, I ' ll add the difference between the two policy.

Monokaix:
> > IMO, the sra semantics should not be present in plugin configuration, keep a single resourcefit concept is ok, and make the configuration as simpler as possible.
> > My original idea may not be quite right. I ' ll re-arrange the configuration here later to make it simpler.

Yeah，just keep a proportional and priority policy is enough I think, and explain the two different policies and give their own example.

docs/design/proportional.md
XbaoWu: done

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: done

pkg/scheduler/plugins/util/util.go
XbaoWu: The `ShouldAbort` func has been moved to `pkg/scheduler/plugins/util/util.go`

XbaoWu:
@Monokaix The configuration structure has been updated, the middle semantic layer has been removed, and only `sra` and `proportion` have been retained. 
The difference between these two policy is supplemented in the doc.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
kingeasternsun, line 1: It looks like `mostRequestedScore` and `leastRequestedScore` have quite a bit of overlapping logic. Maybe we could refactor the common part into a shared function to reduce duplication?

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -125,15 +187,73 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
-		score := Score(task, node, rsf.weight)
-		klog.V(4).Infof("resourceStrategyFit score for Task %s/%s on node %s is: %v", task.Namespace, task.Name, node.Name, score)
+		score := 0.0
+
+		if rsf.weight.ResourceStrategyFitWeight > 0 {
+			score += Score(task, node, rsf.weight)
+			klog.V(4).Infof("resourceStrategyFit policy score for Task %s/%s on node %s is: %v", task.Namespace, task.Name, node.Name, score)
+		}
+
+		if rsf.Sra.Weight > 0 {
```
kingeasternsun, line 227: The way `weight` is used here looks inconsistent:
>
> ```go
> if rsf.weight.ResourceStrategyFitWeight > 0 { ... }
> if rsf.Sra.Weight > 0 { ... }
> ```
>
It would be clearer to keep a consistent configuration style, e.g.:

 * `rsf.weight.ResourceStrategyFit` / `rsf.weight.Sra`, 
 or
 * `rsf.ResourceStrategyFit.Weight` / `rsf.Sra.Weight`.

@JesseStutler @Monokaix @XbaoWu

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Done, good point!

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
XbaoWu: Agreed, this reads better after the change

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -54,41 +60,72 @@ func (w *ResourceStrategyFit) String() string {
...
+		         resourceProportion:
+		           nvidia.com/gpu.cpu: 4
+		           nvidia.com/gpu.memory: 8
 	*/
 
 	var weight ResourceStrategyFit
 
 	resourceStrategyFitPluginWeight, b := framework.Get[int](args, "resourceStrategyFitWeight")
-	if !b || resourceStrategyFitPluginWeight <= 0 {
+	if !b || resourceStrategyFitPluginWeight < 0 {
```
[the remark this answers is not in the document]
XbaoWu: Not a problem, zero weight is a valid value.

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -74,6 +85,111 @@ node score:
...
+|----------|---------------------------------------|---------------------------------------|
+| nodeC0-0 | cpu 74, memory 128G, nvidia.com/gpu 8 | cpu 74, memory 128G, nvidia.com/gpu 8 |
+
+| Job                   | Pod           | Resource                           | Node     | NodeAllocatable                       | NodeIdle                              |
+|-----------------------|---------------|------------------------------------|----------|---------------------------------------|---------------------------------------|
+| default/single-1000-0 | single-1000-0 | cpu 8, memory 8G, nvidia.com/gpu 0 | nodeC0-0 | cpu 74, memory 128G, nvidia.com/gpu 8 | cpu 66, memory 120G, nvidia.com/gpu 8 |
+| default/single-1000-1 | single-1000-1 | cpu 8, memory 8G, nvidia.com/gpu 0 | -        | -                                     | -                                     |
+
+After job single-1000-0 is scheduled, the Idle resource is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin calculates the resource left if job single-1000-1 is scheduled`(node.Idle.CPU - task.Resreq.CPU < node.Idle.GPU * cpuRatio ||
```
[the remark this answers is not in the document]
XbaoWu: done

pkg/scheduler/plugins/resource-strategy-fit/sra_test.go
```diff
@@ -0,0 +1,388 @@
...
+				},
+			},
+		},
+	}
+
+	trueValue := true
+	for i, test := range tests {
+		if i != 2 {
+			continue
+		}
```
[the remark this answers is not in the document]
XbaoWu: done

docs/design/proportional.md
```diff
@@ -50,8 +52,8 @@ Job | Pod | Resource | Node | NodeAllocatable | NodeIdle
 default/single-1000-0 | single-1000-0 | cpu 8, memory 8G, nvidia.com/gpu 0 | nodeC0-0 | cpu 74, memory 128G, nvidia.com/gpu 8 | cpu 66, memory 120G, nvidia.com/gpu 8 |
 default/single-1000-1 | single-1000-1 | cpu 8, memory 8G, nvidia.com/gpu 0 | - | - | - |
 
-After job single-1000-0 is scheduled, the Idel resouce is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin caculates the resource left if job single-1000-1 is scheduled`(node.Idel.CPU - task.Resreq.CPU < node.Idel.GPU * cpuRatio ||
-node.Idel.Memory - task.Resreq.Memory < node.Idel.GPU * memoryRatio)`; the result is 8GPUs, 58CPUs, 112G memory, that unsatisfies the 1:8:8 proportion. Therefore nodeC0-0 is removed from the predicateNodes, and NodeIdle remains 8GPUs, 66CPUs, 120G memory. 
+After job single-1000-0 is scheduled, the Idle resource is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin calculates the resource left if job single-1000-1 is scheduled`(node.Idle.CPU - task.Resreq.CPU < node.Idle.GPU * cpuRatio ||
```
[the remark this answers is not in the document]
XbaoWu: done

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -2,25 +2,36 @@
...
 
 - None.
 
 ## Proposal
 
-Extend one plug-ins to meet the above needs
+Extend one plugin (ResourceStrategyFit) to meet the above needs
+- **ResourceStrategyFit**: Provide users with **MostAllocated** and **LeastAllocated** these two ways to facilitate the user to decide whether the task should be distributed or aggregated according to the different needs of the task.
+
+- **sra**: Provide users the way of weight to avoid common task being scheduled to the presence of scare resource nodes, thereby improving the utilization of scare resources.
```
Monokaix: The weight of each dimension resource?

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -74,6 +85,111 @@ node score:
...
+          resources: nvidia.com/t4, nvidia.com/a10
+          weight: 2
+          resourceWeight:
+            nvidia.com/t4: 1
+            nvidia.com/a10: 1
+```
+
+4. `sra` policy will affect the score result of node order. The higher the node score is, the higher the priority is. The score result of sra policy is calculated as follows :
+```
+node resources items don't meet task request, sraScore = 0
```
Monokaix: Do you mean node has the requested resources of the pod？If true, maybe we can make it more clear instead of using "items".

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -74,6 +85,111 @@ node score:
...
+node resources items meet task request,       sraScore = MaxNodeScore * sraWeight * (weight_1 + weight_2 + ··· + weight_n) / weightSum
+```
+> `sraScore`: the node score of sra policy. \
+> `MaxNodeScore`: the maximum score of node. default is 100. \
+> `sraWeight`: the weight of sra policy. \
+> `weightSum`: the sum of all sra resources weights. \
+> `weight_x`: the weight of the xth sra resource that does not exist on the node.
+
+   1. Now, we have tasks:
```
Monokaix, line 130: Here should add another title because you used a new sequence.

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -74,6 +85,111 @@ node score:
...
+| default/single-1000-0 | single-1000-0 | cpu 8, memory 8G, nvidia.com/gpu 0 | nodeC0-0 | cpu 74, memory 128G, nvidia.com/gpu 8 | cpu 66, memory 120G, nvidia.com/gpu 8 |
+| default/single-1000-1 | single-1000-1 | cpu 8, memory 8G, nvidia.com/gpu 0 | -        | -                                     | -                                     |
+
+After job single-1000-0 is scheduled, the Idle resource is 8GPUs, 66CPUs, 120G memory. During the predicate phase, this plugin calculates the resource left if job single-1000-1 is scheduled `(node.Idle.CPU - task.Resreq.CPU < node.Idle.GPU * cpuRatio ||
+node.Idle.Memory - task.Resreq.Memory < node.Idle.GPU * memoryRatio)` ; the result is 8GPUs, 58CPUs, 112G memory, that unsatisfied the 1:8:8 proportion. Therefore, nodeC0-0 is removed from the predicateNodes, and NodeIdle remains 8GPUs, 66CPUs, 120G memory.
+
+For more details, please refer to: [proportional design](./proportional.md)
 
 ## Syntax Rules
 ### Wildcard Syntax Support
```
Monokaix, line 201: Should also add a notice in this part that the Wildcard Syntax Support is only supported for resourcestrategyfit, not `proportional` and `sra`.

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -134,4 +250,30 @@ resources:
 ## Alternatives
 
 ### Binpack VS ResourceStrategyFit
-If you want to use the clustering strategy for all resource types, you can choose the Binpack plugin. If you need to configure different clustering or scattering strategies for different resource types, you can choose the ResourceStrategyFit plugin. ResourceStrategyFit can also achieve the same results as Binpack by adjusting configuration parameters.
\ No newline at end of file
+If you want to use the clustering strategy for all resource types, you can choose the Binpack plugin. If you need to configure different clustering or scattering strategies for different resource types, you can choose the ResourceStrategyFit plugin. ResourceStrategyFit can also achieve the same results as Binpack by adjusting configuration parameters.
```
Monokaix: There is already a `Aggregated` and `Dispersed` description, so we'd better keep the same statement.
We can change clustering strategy to Aggregated and, change binpack to binpack(Dispersed).

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: Each resource can be configured, but it is not necessary.
Because for example, if a weight is assigned to resource A, the priority of nodes with resource A will be lower than nodes without resource A. Therefore, in the presence of resource A, only the weight of important resources needs to be assigned

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: > Do you mean node has the requested resources of the pod？If true, maybe we can make it more clear instead of using "items".

This means that there is no pod request resource on the node, in order to avoid ambiguity I modify the description here

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: > Here should add another title because you used a new sequence.

good point, I will modify the content here.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: > Should also add a notice in this part that the Wildcard Syntax Support is only supported for resourcestrategyfit, not `proportional` and `sra`.

Agreed, here I add a note.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: > There is already a `Aggregated` and `Dispersed` description, so we'd better keep the same statement. We can change clustering strategy to Aggregated and, change binpack to binpack(Dispersed).

Good supplement, I will adjust the description here recently.

docs/design/resource-strategy-fit-scheduling.md
Monokaix: Yeah I mean we can make it more clear that user can set each dimension resource's weight.

Monokaix:
Ok, please also solve the code conflict.

docs/design/resource-strategy-fit-scheduling.md
XbaoWu: done

Monokaix:
/approve
Thanks a lot!

JesseStutler (commented):
/lgtm

PR #4455, opened 2025-07-14
Add hierarchical queue support for nodegroup plugin

### LINKED ISSUE #4426: Hierarchical queues compatible with Nodegroup

Combined with existing capabilities, each sub-queue can currently be associated with an independent nodegroup. However, in this scenario, cluster resources are divided into relatively fragmented parts. After Volcano implements the hierarchical queue capability, it is hoped that nodegroup information can be managed through the dimension of the parent queue to avoid excessive fragmentation of resource division.

The current queue can be allocated with the nodegroup quota to complete the automatic affinity of the job in the queue to the corresponding nodegroup.

In the scenario of hierarchical queues, it is hoped that the following capabilities can be supported:
The parent queue is associated with the nodegroup, and the jobs submitted by the child queue are automatically associated with the nodegroup associated with the parent queue

### Additional context

_No response_

JesseStutler:
#### What type of PR is this?
/kind feature

Inspired by https://github.com/volcano-sh/volcano/issues/4426, in the scenario of hierarchical queues, it is hoped that the following capabilities can be supported:
The parent queue is associated with the nodegroup, and the jobs submitted by the child queue are automatically associated with the nodegroup associated with the parent queue. Therefore we can implement it in this way: if a child queue has no affinity, it will inherit the affinity configuration from the nearest ancestor queue

Fixes #4426 

Validation:
We have such queues: [root, q1, q2, q2-sub]
The parent of q1 and q2 is the root queue. Neither q1 nor q2-sub specifies affinity. The affinity specified by the root queue is: 
```yaml
    affinity: 
      nodeGroupAffinity:
        - groupname1
        - groupname2
        preferredDuringSchedulingIgnoredDuringExecution:
        - groupname1
      nodeGroupAntiAffinity:
        - groupname3
        - gropuname4
        preferredDuringSchedulingIgnoredDuringExecution:
        - groupname3
```
, and the affinity and antiaffinity of q2 are completely opposite to those of the root. You can see the printout when the nodegroup plugin OnSessionOpen is: 
, q1 inherits the configuration of the root, while q2-sub inherits the configuration of q2
image

Besides, this PR should also be merged: https://github.com/volcano-sh/apis/pull/186, otherwise the root queue will miss the affinity field even if I add it 

```release-note
The nodegroup plugin adds the `enableHierarchy` configuration. If enableHierarchy is set to true, the hierarchical queue configuration will be enabled. A child queue without a specified affinity will inherit the affinity configuration of the nearest ancestor queue.
```

### DIFF
--- a/docs/design/node-group.md
+++ b/docs/design/node-group.md
@@ -10,6 +10,8 @@ case2: recommend queue can use private cloud nodes or public cloud nodes, but tt
 
 ![](images/node-group-case-2.png)
 
+case3: An organization has a hierarchical structure (like engineering -> backend-team and engineering -> frontend-team). The parent queue (engineering) is associated with production nodegroup, and the jobs submitted by child queues (backend-team, frontend-team) are automatically associated with the nodegroup associated with the parent queue, unless the child queue specifies its own nodegroup affinity.
+
 ## Solution
 
 1. First, we need mark out some nodes(add `volcano.sh/nodegroup-name` labels) which are the same group.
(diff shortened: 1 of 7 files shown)


JesseStutler:
Besides, this PR should also be merged: https://github.com/volcano-sh/apis/pull/186, otherwise the root queue will miss the affinity field even if I add it

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -51,194 +69,218 @@ func (np *nodeGroupPlugin) Name() string {
...
-				flag = true
-			}
-		}
+	if q.queueGroupAntiAffinityPreferred.Has(group) {
+		flag = true
 	}
 	if !flag {
 		return errors.New("not satisfy")
 	}
 	return nil
```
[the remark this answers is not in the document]
JesseStutler: Yes, you are right. Generally speaking, required is a hard constraint and preferred should appear in the scoring rather than the predicate. However, my current modification follows the original design. Required is a set of node ranges, and preferred is the preferred node in this subset of node ranges. The range of anti-affinity is larger than required. 
All nodes except this anti-affinity node can be scheduled

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -51,194 +69,218 @@
...
-	if _, ok := task.Pod.Annotations[batch.QueueNameKey]; ok {
-		return task.Pod.Annotations[batch.QueueNameKey]
+
+	// Recursively build ancestors for the parent queue.
+	np.buildAncestors(ssn, parentAttr)
+
+	// A queue's ancestors are its parent plus all the parent's ancestors.
+	attr.ancestors.PushBack(parentAttr)
+	attr.ancestors.PushBackList(parentAttr.ancestors)
+}
```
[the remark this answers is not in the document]
JesseStutler: Well, it's a good suggestion, although potentially users won't configure the cyclic dependency queues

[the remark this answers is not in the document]
JesseStutler: Added

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -51,194 +68,226 @@ func (np *nodeGroupPlugin) Name() string {
...
 }
 
-// There are 3 ways to assign pod to queue for now:
-// scheduling.volcano.sh/queue-name support only annotation
-// volcano.sh/queue-name support both labels & annotation
-// the key should be unified, maybe volcano.sh/queue-name is better
-func GetPodQueue(task *api.TaskInfo) string {
-	if _, ok := task.Pod.Labels[batch.QueueNameKey]; ok {
-		return task.Pod.Labels[batch.QueueNameKey]
+func (np *nodeGroupPlugin) buildAncestors(ssn *framework.Session, attr *queueAttr, visited map[api.QueueID]struct{}) {
```
kingeasternsun, line 183: I’m wondering if it’s really necessary to maintain Ancestors eagerly. We could just perform a recursive lookup when needed, since the queue hierarchy is unlikely to be very deep.

kingeasternsun: Because when the Child queue has affinity ,  just use  it's own affinity (no  need inheritance).

docs/design/node-group.md
```diff
@@ -37,6 +39,70 @@ case2: recommend queue can use private cloud nodes or public cloud nodes, but tt
            - groupname3
    ```
 
+### Hierarchical Queue Support
+
+The nodegroup plugin supports hierarchical queue with affinity inheritance. This feature allows child queues to inherit nodegroup affinity configurations from their parent queues when they don't have explicit affinity settings.
+
+#### Inheritance Rules
+
+1. **Direct Inheritance**: A child queue without affinity configuration inherits from its immediate parent queue.
```
Monokaix, line 48: “most recent ancestor” is better.

pkg/scheduler/plugins/nodegroup/nodegroup.go
Monokaix: Reasonable.

Monokaix: But it's more simple in this way, I think this can also be accepted.

Monokaix:
Will fire another pr to fix some minor nits, we can merge this first, as @JesseStutler is busy for other work. @kingeasternsun

kingeasternsun:
> Will fire another pr to fix some minor nits, we can merge this first, as @JesseStutler is busy for other work. @kingeasternsun

ok

PR #4464, opened 2025-07-16
Correct step order to enable caching in E2E workflows

GautamBytes:
This PR fixes a bug in the E2E workflows where Go module caching was ineffective.

The `actions/cache` step was running before `actions/checkout`, so the cache key could not be generated from the `go.sum` file. This change reorders the steps, which will properly enable caching and improve the performance of the E2E jobs.


JesseStutler:
/lgtm
thanks!

PR #4478, opened 2025-07-22
Integrate with leaderworkerset: Volcano controller doesn't need to create podgroup for statefulset if statefulset pods have associated podgroup

JesseStutler:
#### What type of PR is this?
/kind feature

Currently, volcano will list/watch statefulsets and create pg for the statefulset if it doesn't have a related one. But in leaderWorkerSet scenario, lws uses statefulset as the basic resource:
image
Leader has a statefulset and workers have their own statefulsets, as in the pic shown above, 4 replicas lws will create 5 pgs(5 statefulsets). But now we want to schedule 1 replica as a group( means 1 leader pod + (size-1) worker pods), just like the red box in the picture, we only need to create 4 pgs instead. Therefore we need to add a judgement in addStatefulset, if the pods already have pg annotation, skip creating pg

Related https://github.com/kubernetes-sigs/lws/pull/496

```release-note

```

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -148,6 +148,12 @@ func (pg *pgcontroller) addStatefulSet(obj interface{}) {
 				klog.V(4).Infof("Pod %s field SchedulerName is not matched", klog.KObj(pod))
 				return
 			}
+			// If the pod is already associated with a podgroup, skip creating a new one. This scenario is applicable to LeaderWorkerSet,
+			// which will create podgroups by itself, and Volcano does not need to create a podgroup for statefulset.
+			if pgName := pod.Annotations[scheduling.KubeGroupNameAnnotationKey]; pgName != "" {
+				klog.V(4).Infof("Pod %s is already associated with a podgroup %s", klog.KObj(pod), pgName)
+				return
+			}
 			err := pg.createNormalPodPGIfNotExist(pod)
 			if err != nil {
 				klog.Errorf("Failed to create PodGroup for pod <%s/%s>: %v", pod.Namespace, pod.Name, err)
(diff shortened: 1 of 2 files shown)


pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -148,6 +148,12 @@ func (pg *pgcontroller) addStatefulSet(obj interface{}) {
 				klog.V(4).Infof("Pod %s field SchedulerName is not matched", klog.KObj(pod))
 				return
 			}
+			// If the pod is already associated with a podgroup, skip creating a new one. This scenario is applicable to LeaderWorkerSet,
+			// which will create podgroups by itself, and Volcano does not need to create a podgroup for statefulset.
+			if pod.Annotations != nil && pod.Annotations[scheduling.KubeGroupNameAnnotationKey] != "" {
+				klog.V(4).Infof("pod %v/%v is already associated with a podgroup %s", pod.Namespace, pod.Name, pod.Annotations[scheduling.KubeGroupNameAnnotationKey])
+				return
+			}
```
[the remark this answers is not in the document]
JesseStutler: Good suggestion, I adopted it

hwdef:
Will this affect the generality of the code?

JesseStutler:
> Will this affect the generality of the code?

I think it will be alright, because I did not add any lws related code. I just checked whether the statefulset pod has pod annotation. This is also the common practice of vc controller to create pg for pod. @hwdef

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -148,6 +148,12 @@ func (pg *pgcontroller) addStatefulSet(obj interface{}) {
 				klog.V(4).Infof("Pod %s field SchedulerName is not matched", klog.KObj(pod))
 				return
 			}
+			// If the pod is already associated with a podgroup, skip creating a new one. This scenario is applicable to LeaderWorkerSet,
```
Monokaix, line 151: Please add ut.

JesseStutler: done

PR #4430, opened 2025-07-04
[feature] upgrade k8s to v1.33.2

mahdikhashan:
#### What type of PR is this?

feature

#### What this PR does / why we need it:

This PR upgrades Kubernetes to v1.33, replacing the deprecated VolumeCapacityPriority with StorageCapacityScoring as per [KEP-4049](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/4049-storage-capacity-scoring-of-nodes-for-dynamic-provisioning/README.md).
It includes necessary code, YAML, and Docker changes to support this transition.

#### Which issue(s) this PR fixes:

replaces #4342 and fixes #4341

#### Does this PR introduce a user-facing change?

```release-note
Replaced deprecated VolumeCapacityPriority with StorageCapacityScoring for dynamic volume provisioning in Kubernetes v1.33 (KEP-4049)
```

PR #4483, opened 2025-07-24
Fix volume binding e2e testing

### LINKED ISSUE #4480: Volcano scheduler panic when scheduling Pods with delayed binding PVCs
### Description

Volcano scheduler panic when scheduling Pods with delayed binding PVCs

1.Submit a pod with delayed binding PVCs
2.Check whether the pod can be scheduled normally

Volcano scheduler can schedule Pods with delayed binding PVCs normally

master

_No response_

JesseStutler:
#### What type of PR is this?
/kind bug

In #4430, we removed these lines: https://github.com/volcano-sh/volcano/blob/0a8cbd677e5d8314f30940ce51ae217dfcf7df94/pkg/scheduler/capabilities/volumebinding/binder.go#L956-L958, which is important when `EnableCSIStorage` server option is false. Now when we want to create Dynamic provision pod, the scheduler will panic at this line: https://github.com/volcano-sh/volcano/blob/0a8cbd677e5d8314f30940ce51ae217dfcf7df94/pkg/scheduler/capabilities/volumebinding/binder.go#L982

Besides, current volume binding e2e testing incorrectly use default-scheduler to schedue pods requesting PVCs. These pods should use volcano to schedule, therefore I add test/e2e/util/scheduler_injector.go, it can modify the schedulerName of the pod non-invasively without copying the k8s source code. And I add a new Dynamic Provision e2e testing case

Fixes #4480 

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,8 @@ e2e: images
 	./hack/run-e2e-kind.sh
 
 e2e-test-schedulingbase: images
-	E2E_TYPE=SCHEDULINGBASE ./hack/run-e2e-kind.sh
+	# volume binding e2e testing has some cases using "kubernetes.io/no-provisioner" as the storage provisioner, so we need to specify an ignored provisioner here.
+	E2E_TYPE=SCHEDULINGBASE IGNORED_PROVISIONERS="kubernetes.io/no-provisioner" ./hack/run-e2e-kind.sh
 
 e2e-test-schedulingaction: images
 	E2E_TYPE=SCHEDULINGACTION ./hack/run-e2e-kind.sh
--- a/installer/helm/chart/volcano/templates/scheduler.yaml
+++ b/installer/helm/chart/volcano/templates/scheduler.yaml
@@ -226,6 +226,9 @@ spec:
             {{- if .Values.custom.scheduler_feature_gates }}
             - --feature-gates={{ .Values.custom.scheduler_feature_gates }}
             {{- end }}
+            {{- if .Values.custom.ignored_provisioners }}
+            - --ignored-provisioners={{ .Values.custom.ignored_provisioners}}
+            {{- end}}
             - -v={{.Values.custom.scheduler_log_level}}
             - 2>&1
           env:
--- a/test/e2e/schedulingbase/volume_binding.go
+++ b/test/e2e/schedulingbase/volume_binding.go
@@ -42,10 +42,12 @@ import (
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	e2epv "k8s.io/kubernetes/test/e2e/framework/pv"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	"k8s.io/kubernetes/test/e2e/storage/testsuites"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
 	imageutils "k8s.io/kubernetes/test/utils/image"
 	admissionapi "k8s.io/pod-security-admission/api"
 
+	e2eutil "volcano.sh/volcano/test/e2e/util"
 	volcanotestutil "volcano.sh/volcano/test/e2e/util"
 )
 
@@ -124,6 +126,7 @@ var _ = ginkgo.Describe("Volume Binding Test", func() {
 	)
 
 	ginkgo.BeforeEach(func(ctx context.Context) {
+		f.ClientSet = e2eutil.NewSchedulerInjectingClientset(f.ClientSet, e2eutil.SchedulerName)
 		nodes, err := e2enode.GetBoundedReadySchedulableNodes(ctx, f.ClientSet, maxNodes)
 		framework.ExpectNoError(err)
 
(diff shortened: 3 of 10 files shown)


installer/helm/chart/volcano/templates/scheduler.yaml
```diff
@@ -226,6 +226,9 @@ spec:
             {{- if .Values.custom.scheduler_feature_gates }}
             - --feature-gates={{ .Values.custom.scheduler_feature_gates }}
             {{- end }}
+            {{- if .Values.custom.ignored_provisioners }}
+            - --ignored-provisioners={{ .Values.custom.ignored_provisioners}}
+            {{- end}}
```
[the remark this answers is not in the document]
JesseStutler: This is not correct

Makefile
```diff
@@ -129,7 +129,7 @@ e2e: images
 	./hack/run-e2e-kind.sh
 
 e2e-test-schedulingbase: images
-	E2E_TYPE=SCHEDULINGBASE ./hack/run-e2e-kind.sh
+	E2E_TYPE=SCHEDULINGBASE IGNORED_PROVISIONERS="kubernetes.io/no-provisioner" ./hack/run-e2e-kind.sh
```
Monokaix, line 133: Why ignore this one?

JesseStutler: Because in volume binding e2e testing, setupStorageClass uses the `no-provisioner` storageclass to set up:
https://github.com/JesseStutler/volcano/blob/3aa260cf38ed112a1764473a316e2f280ae4571c/test/e2e/schedulingbase/volume_binding.go#L451-L462
If I don't set ignored-provisioners here, https://github.com/volcano-sh/volcano/blob/36a46e2812ce2f2fbfb10ac7929385b18752176f/pkg/scheduler/cache/event_handlers.go#L87-L98 will add an attachment resource for pods requesting PVCs, then overused validation in proportion plugin will reject the pod to be scheduled because the default queue didn't set attachment resource quota.

JesseStutler: Because in volume binding e2e testing, setupStorageClass uses the `no-provisioner` storageclass to set up:
https://github.com/JesseStutler/volcano/blob/3aa260cf38ed112a1764473a316e2f280ae4571c/test/e2e/schedulingbase/volume_binding.go#L451-L462
If I don't set ignored-provisioners here, https://github.com/volcano-sh/volcano/blob/36a46e2812ce2f2fbfb10ac7929385b18752176f/pkg/scheduler/cache/event_handlers.go#L87-L98 will add an attachment resource for pods requesting PVCs, then overused validation in proportion plugin will reject the pod to be scheduled because the default queue didn't set attachment resource quota.

JesseStutler: Because in volume binding e2e testing, setupStorageClass uses the `no-provisioner` storageclass to set up:
https://github.com/JesseStutler/volcano/blob/3aa260cf38ed112a1764473a316e2f280ae4571c/test/e2e/schedulingbase/volume_binding.go#L451-L462
If I don't set ignored-provisioners here, https://github.com/volcano-sh/volcano/blob/36a46e2812ce2f2fbfb10ac7929385b18752176f/pkg/scheduler/cache/event_handlers.go#L87-L98 will add an attachment resource for pods requesting PVCs, then overused validation in proportion plugin will reject the pod to be scheduled because the default queue didn't set attachment resource quota.

test/e2e/schedulingbase/volume_binding.go
```diff
@@ -363,6 +366,39 @@ var _ = ginkgo.Describe("Volume Binding Test", func() {
 			validateStatefulSet(ctx, config, ss, false)
 		})
 	})
+
+	f.Context("Dynamic Provisioning", f.WithSlow(), func() {
```
Monokaix, line 370: Is this a dedicated case just for volcano?

test/e2e/util/scheduler_injector.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// schedulerInjectingPodInterface wraps PodInterface and intercepts the Create method to inject a scheduler name into the Pod spec.
+type schedulerInjectingPodInterface struct {
+	v1.PodInterface
+	schedulerName string
+}
+
+func (p *schedulerInjectingPodInterface) Create(ctx context.Context, pod *v1core.Pod, opts metav1.CreateOptions) (*v1core.Pod, error) {
+	if pod.Spec.SchedulerName == "" {
```
Monokaix: What if pods already set a schedulerName?

test/e2e/schedulingbase/volume_binding.go
```diff
@@ -124,6 +126,7 @@ var _ = ginkgo.Describe("Volume Binding Test", func() {
 	)
 
 	ginkgo.BeforeEach(func(ctx context.Context) {
+		f.ClientSet = e2eutil.NewSchedulerInjectingClientset(f.ClientSet, e2eutil.SchedulerName)
```
Monokaix, line 129: How about set a parameter `--scheduler-name="volcano,defaule-scheduler"` to let volcano schedule all pods directly?

Makefile
Monokaix: We can add a comment here.

test/e2e/schedulingbase/volume_binding.go
JesseStutler: yes, since we do not have a cloud vendor offering PVs, I use kind's "standard" storageclass to provide PVs

test/e2e/schedulingbase/volume_binding.go
JesseStutler: I think it may cause kube-scheduler and volcano to compete for scheduling pods and cause strange behavior, set the schedulerName directly will make sure everything is correct and scheduled by volcano

test/e2e/util/scheduler_injector.go
JesseStutler: Then I think these pods should be scheduled by other schedulers, such as kube-scheduler, but in our use cases there are no cases specify schedulerName to default-scheduler, it should be fine. And currently scheduler_injector is only used in volume_binding e2e testing cases

Makefile
JesseStutler: OK done

test/e2e/util/scheduler_injector.go
Monokaix: The schedulerName field is set by kube-scheduler's e2e test, right?
If so, the schedulerName will be set to `default-scheduler`.

JesseStutler: No in volume binding e2e testing, the pod functions related to creation (copied from the kubernetes repo) do not set the schedulerName. My injector is to inject the schedulerName before issuing the pod creation request. If the schedulerName is set to volcano, the kube-apiserver will not set the default value for this pod.

Monokaix: We can't determine the behavior of k8s e2e, so it's better to rewrite this field directly.

JesseStutler: OK. Updated

PR #4493, opened 2025-07-25
added preemmptable attribute into pod group

### LINKED ISSUE #2141: Add preemmptable attribute into PodGroup

Add Preemptable attribute into PodGroup to enable preempt/reclaim action.

#### Why is this needed:

Currently, we have preemptable annotation for Pod; and it's better to enable Preemptable for PodGroup:

- If PodGroup is preemptable, its Pods can be preemptable or not
- If PodGroup is un-preemptable, its Pods can not be preemptable, we should check this in webhook

SuyashParmar:
### What this PR does

- Adds an optional `Preemptable` field to `PodGroupSpec` in the API and CRD.
- Enables PodGroup-level control over preempt/reclaim actions.
- Enforces via webhook:
  - If a PodGroup is **preemptable**, its Pods can be preemptable or not.
  - If a PodGroup is **un-preemptable**, its Pods cannot be preemptable (checked in the admission webhook).

### Special notes for your reviewer

- No changes to scheduler logic; enforcement is strictly at the webhook/admission level.
- Backward compatible: `Preemptable` is optional and defaults to current behavior if unset.

---

**API/CRD changes:**
- `Preemptable` field added to `PodGroupSpec` (Go types and CRD schema).

**Webhook changes:**
- Pod admission webhook rejects Pods that are preemptable if their PodGroup is un-preemptable.

---

Fixes #2141

### DIFF
--- a/config/crd/volcano/bases/scheduling.volcano.sh_podgroups.yaml
+++ b/config/crd/volcano/bases/scheduling.volcano.sh_podgroups.yaml
@@ -122,6 +122,10 @@ spec:
                   Queue defines the queue to allocate resource for PodGroup; if queue does not exist,
                   the PodGroup will not be scheduled. Defaults to `default` Queue with the lowest weight.
                 type: string
+              preemptable:
+                description: |
+                  Preemptable indicates whether the PodGroup is preemptable. If false, none of its Pods can be preemptable.
+                type: boolean
             type: object
           status:
             description: |-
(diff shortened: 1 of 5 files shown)


Monokaix:
We should modify API repo first, and the behavior should be: preemptable filed of pg is nil or preemptable==true means pg can be preempted for compatibility.

Issue #4497, opened 2025-07-29
hypernode controller can not update hypernode spec


cyf-2002:
### Description

The update function uses UpdateStatus() instead of Update(), so when hypernode spec needs to be updated, controller will ignore.

### Steps to reproduce the issue

1. design a network below, hypernode s0 will have 2 members

<img width="192" height="203" alt="Image" src="https://github.com/user-attachments/assets/74eb3420-1553-44ab-9da8-a4aeba853a90" />

2. then remove node1
3. hypernode s0 still have 2 members

### Describe the results you received and expected

hypernode s0 spec expected to have 1 member.

### What version of Volcano are you using?

master

kev1N916:
@cyf-2002 I think this issue can be closed right??

PR #4498, opened 2025-07-29
fix#4497: hypernode update bug

cyf-2002:
#### What type of PR is this?
fix
Hypernode controller will ignore updating spec, as it only calls UpdateStatus() to update hypernode. When I test the function of autoupdate, hypernode spec will never be updated since it's created. So we should add Update() call.
#4497 

Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/hypernode/utils/utils.go
+++ b/pkg/controllers/hypernode/utils/utils.go
@@ -64,6 +64,7 @@ func UpdateHyperNode(vcClient vcclientset.Interface, lister v1alpha1.HyperNodeLi
 		}
 
 		_, err = vcClient.TopologyV1alpha1().HyperNodes().Update(context.Background(), current, metav1.UpdateOptions{})
+		_, err = vcClient.TopologyV1alpha1().HyperNodes().UpdateStatus(context.Background(), current, metav1.UpdateOptions{})
 		return err
 	})
 }
--- a/pkg/controllers/hypernode/utils/utils.go
+++ b/pkg/controllers/hypernode/utils/utils.go
@@ -63,7 +63,7 @@ func UpdateHyperNode(vcClient vcclientset.Interface, lister v1alpha1.HyperNodeLi
 			current.Annotations[k] = v
 		}
 
-		_, err = vcClient.TopologyV1alpha1().HyperNodes().UpdateStatus(context.Background(), current, metav1.UpdateOptions{})
+		_, err = vcClient.TopologyV1alpha1().HyperNodes().Update(context.Background(), current, metav1.UpdateOptions{})
 		return err
 	})
 }


JesseStutler:
/priority high
Thanks, but please complete your description of the PR, right now it's too simple. @cyf-2002

pkg/controllers/hypernode/utils/utils.go
```diff
@@ -63,7 +63,7 @@ func UpdateHyperNode(vcClient vcclientset.Interface, lister v1alpha1.HyperNodeLi
 			current.Annotations[k] = v
 		}
 
-		_, err = vcClient.TopologyV1alpha1().HyperNodes().UpdateStatus(context.Background(), current, metav1.UpdateOptions{})
+		_, err = vcClient.TopologyV1alpha1().HyperNodes().Update(context.Background(), current, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
JesseStutler: Yes it's correct to use Update method instead, but I think we don't need to update the hypernode status here, L50 can be deleted @cyf-2002 This is responsible for processing the automatically discovered hypernode results, only change the spec and meta

Monokaix, line 66: We need update both spec and status as the HyperNode status may also include Network condition in status and this also need to be updated.

cyf-2002: Can update func update status？

Monokaix: No, I mean we should update both spec and status, any other way is ok.

cyf-2002: Thanks，I got it. So if I want to fix this bug, I can call both update and updatestatus func.

cyf-2002:
> /priority high Thanks, but please complete your description of the PR, right now it's too simple. @cyf-2002

cyf-2002:
> /priority high Thanks, but please complete your description of the PR, right now it's too simple. @cyf-2002

I added some details, and I described the question on #4497.

pkg/controllers/hypernode/utils/utils.go
```diff
@@ -63,6 +63,7 @@ func UpdateHyperNode(vcClient vcclientset.Interface, lister v1alpha1.HyperNodeLi
 			current.Annotations[k] = v
 		}
 
+		_, err = vcClient.TopologyV1alpha1().HyperNodes().Update(context.Background(), current, metav1.UpdateOptions{})
```
JesseStutler, line 66: We still need to judge err here to avoid override
```
if err != nil{
  return err
}
```

cyf-2002: Thanks.

Monokaix:
Is there a way to just call API once and update both spec and status?

Monokaix:
Please also merge to one commit.

cyf-2002:
> Please also merge to one commit.

Done. I can't find a way to just call API once temporarily.

PR #4508, opened 2025-08-02
improve performance by reducing node mutating requests and wait inter…

### LINKED ISSUE #4507: improve performance by reducing node mutating requests and reducing wait intervals wher retry setting node lock

​​In a large-scale cluster with numerous nodes, reducing node update requests can significantly alleviate cluster pressure. Therefore, I suggest: 

1. When the scheduler attempts to acquire a lock and finds that the node lock has expired, could it skip releasing the old lock and directly set a new one? 
2. Additionally, for node update operations, we can use kubernetes' native retry library functions instead of a fixed 100ms wait.​

1. When the scheduler attempts to acquire a lock and finds that the node lock has expired, we can  skip releasing the old lock and directly set a new one.
2. For node update operations, we can use kubernetes' native retry library functions instead of a fixed 100ms wait.​

### Additional context

_No response_

wangao1236:
…vals

Fixes #4507

```release-note

```

### DIFF
--- a/pkg/scheduler/plugins/util/nodelock/nodelock.go
+++ b/pkg/scheduler/plugins/util/nodelock/nodelock.go
@@ -28,10 +28,9 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 )
-
-const MaxLockRetry = 5
 
 var kubeClient kubernetes.Interface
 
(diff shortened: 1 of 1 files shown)


archlitchi:
hi @wangao1236 , thanks for your contributing
the default plan is not using node-lock, because it's conflicts with gang-scheduling. we use timestamp instead, you need to actively enable node-lock for it to be effective. So it this PR based on scenario which demands node-lock, or just for the purpose of optimizing source code?

pkg/scheduler/plugins/util/nodelock/nodelock.go
```diff
@@ -68,21 +67,24 @@ func updateNodeAnnotations(ctx context.Context, node *v1.Node, updateFunc func(a
...
+			return fmt.Errorf("failed to update node %s: %v", nodeName, err)
 		}
-		newNode = node.DeepCopy()
-		updateFunc(newNode.ObjectMeta.Annotations)
-		_, err = kubeClient.CoreV1().Nodes().Update(ctx, newNode, metav1.UpdateOptions{})
-	}
-	if err != nil {
-		klog.ErrorS(err, "Failed to update node", "node", nodeName)
-		return fmt.Errorf("failed to update node %s, exceeded retry count %d", nodeName, MaxLockRetry)
 	}
```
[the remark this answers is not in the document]
Monokaix: it's reasonable.

pkg/scheduler/plugins/util/nodelock/nodelock.go
```diff
@@ -145,11 +126,6 @@ func LockNode(nodeName string, lockName string) error {
...
 	if time.Since(lockTime) > time.Minute*5 {
 		klog.V(3).InfoS("Node lock expired", "node", nodeName, "lockTime", lockTime)
-		err = ReleaseNodeLock(nodeName, lockName)
```
Monokaix, line 148: Why remove this?

wangao1236: Even if we don’t remove this line, new annotations will be set later, overwriting the old one. So why can’t we directly update the Annotation to reduce one node update operation? 
​​​​In large-scale clusters, shouldn't we minimize node update operations as much as possible to reduce cluster pressure?​ @Monokaix

wangao1236:
> hi @wangao1236 , thanks for your contributing the default plan is not using node-lock, because it's conflicts with gang-scheduling. we use timestamp instead, you need to actively enable node-lock for it to be effective. So it this PR based on scenario which demands node-lock, or just for the purpose of optimizing source code?

This optimization targets scenarios with strong lock dependencies.​ @archlitchi

wangao1236:
> hi @wangao1236 , thanks for your contributing the default plan is not using node-lock, because it's conflicts with gang-scheduling. we use timestamp instead, you need to actively enable node-lock for it to be effective. So it this PR based on scenario which demands node-lock, or just for the purpose of optimizing source code?

For some reasons, we are using the default scheduler's framework plugins. I noticed that our implementation references volcano's design regarding node locking and achieves synchronization interaction with the device-plugin. I’ve identified some potential optimizations and would like to extend these improvements to the volcano community.

Could you please elaborate on the design of the timestamp mechanism in detail? @archlitchi

archlitchi:
> > hi @wangao1236 , thanks for your contributing the default plan is not using node-lock, because it's conflicts with gang-scheduling. we use timestamp instead, you need to actively enable node-lock for it to be effective. So it this PR based on scenario which demands node-lock, or just for the purpose of optimizing source code?
> 
> For some reasons, we are using the default scheduler's framework plugins. I noticed that our implementation references volcano's design regarding node locking and achieves synchronization interaction with the device-plugin. I’ve identified some potential optimizations and would like to extend these improvements to the volcano community.
> 
> Could you please elaborate on the design of the timestamp mechanism in detail? @archlitchi

yes, the timestamp mechanism is we assign the predicate-time to each task, and order it inside device-plugin in a FIFO way, you can find more details in volcano-vgpu-device-plugin

archlitchi:
please sign-off your commit to pass the DCO check

wangao1236:
> please sign-off your commit to pass the DCO check

Thanks, I have done. @archlitchi

PR #4512, opened 2025-08-04
feat: support wildcard syntax in resource-strategy-fit plugin

### LINKED ISSUE #4449: feat: Support prefix matching in resource policies to handle varied GPU types

At present, the policy configuration requires specifying each resource type individually. For example:

```yaml
resources:
  nvidia.com/gpu:
    type: MostAllocated
    weight: 2
  cpu:
    type: LeastAllocated
    weight: 1
```

However, in a real-world production environment, a cluster might have multiple resource pools with various specific GPU types. For instance, we could have several GPU types like `cloudml.gpu/r49-24g` and `cloudml.gpu/r39-24g`.

Under the current strategy, we would need to be aware of and configure every single GPU type present in the cluster. This adds an operational burden, as we need to know all GPU variants beforehand and update the configuration whenever a new GPU type is introduced. This approach does not scale well in complex environments with a diverse set of resources.

I would like to suggest enhancing this feature to support a unified, prefix-based configuration. This would allow for more flexible and streamlined management.

For example, we could define a general policy for `cloudml.gpu` that automatically applies to all resource types starting with that prefix (e.g., `cloudml.gpu/r49-24g`, `cloudml.gpu/r39-24g`, etc.).

Crucially, the ability to specify a more detailed configuration for a particular resource type should be retained to override the general policy.

Here is an example of the desired behavior:
```yaml
# Proposed Idea
resources:
  # This would apply to all resources with the "cloudml.gpu/" prefix as a default
  cloudml.gpu: 
    type: MostAllocated
    weight: 3
  
  # This would override the general rule for this specific GPU type
  cloudml.gpu/r49-24g:
    type: LeastAllocated
    weight: 1

  # Other resources would work as before
  cpu:
    type: LeastAllocated
    weight: 1
```

This enhancement would make the configuration much more scalable and easier to manage in complex cluster environments. I'd be happy to work on a PR for it.

### Additional context

- docs/design/resource-strategy-fit-scheduling.md
config：
```yaml
actions: "enqueue, allocate, backfill, reclaim, preempt"
tiers:
- plugins:
 - name: resource-strategy-fit
    arguments:
      resourceStrategyFitWeight: 10
      resources:
        nvidia.com/gpu:
          type: MostAllocated
          weight: 2
        cpu:
          type: LeastAllocated
          weight: 1
```

- pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go

```go
func calculateWeight(args framework.Arguments) ResourceStrategyFit {
	/*
	   actions: "enqueue, allocate, backfill, reclaim, preempt"
	   tiers:
	   - plugins:
	     - name: resource-strategy-fit
	        arguments:
	          resourceStrategyFitWeight: 10
	          resources:
	            nvidia.com/gpu:
	              type: MostAllocated
	              weight: 2
	            cpu:
	              type: LeastAllocated
	              weight: 1
	*/

	var weight ResourceStrategyFit

	resourceStrategyFitPluginWeight, b := framework.Get[int](args, "resourceStrategyFitWeight")
	if !b || resourceStrategyFitPluginWeight <= 0 {
		resourceStrategyFitPluginWeight = DefaultResourceStrategyFitPluginWeight
	}
	weight.ResourceStrategyFitWeight = resourceStrategyFitPluginWeight

	resources, b := framework.Get[map[v1.ResourceName]ResourcesType](args, "resources")
	if !b || len(resources) == 0 {
		resources = map[v1.ResourceName]ResourcesType{
			"cpu": {
				Type:   config.LeastAllocated,
				Weight: 1,
			},
			"memory": {
				Type:   config.LeastAllocated,
				Weight: 1,
			},
		}
	}

	for k, v := range resources {
		if v.Weight <= 0 {
			v.Weight = 1
		}
		if v.Type != config.LeastAllocated && v.Type != config.MostAllocated {
			v.Type = config.LeastAllocated
		}
		resources[k] = v
	}
	weight.Resources = resources
	return weight
}
```

ditingdapeng:
/kind feature

This PR implements wildcard syntax support for the resource-strategy-fit plugin to address scalability issues in GPU resource configuration.

**Problem**: Current configuration requires specifying each GPU type individually (e.g., `cloudml.gpu/r49-24g`, `cloudml.gpu/r39-24g`), creating operational burden in diverse GPU environments.

**Solution**: 
- Add wildcard pattern validation for resource configurations
- Support prefix matching with trailing asterisk (e.g., 'cloudml.gpu/*')
- Filter out invalid patterns like single '*' or multiple asterisks
- Add comprehensive test cases for wildcard functionality

Fixes #4449

- Implementation follows "Solution 1: Wildcard Syntax Support" from the issue discussion
- All existing tests pass, new tests validate wildcard functionality
- Warning logs help identify invalid configurations
- Maintains backward compatibility

```release-note
resource-strategy-fit plugin now supports wildcard syntax (e.g., 'cloudml.gpu/*') for scalable GPU resource configuration
```

### DIFF
--- a/docs/design/resource-strategy-fit-scheduling.md
+++ b/docs/design/resource-strategy-fit-scheduling.md
@@ -78,6 +78,7 @@ finalScoreNode = [(weight1 * resource1) + (weight2 * resource2) + … + (weightN
 ## Syntax Rules
 ### Wildcard Syntax Support
 - Only suffix wildcard patterns are supported (e.g., `nvidia.com/gpu/*`)
+- **Regular expressions are NOT supported** (e.g., `gpu-[a-z]+`, `nvidia.com/gpu-[1-9]*`)
 - Invalid patterns are filtered out during configuration parsing:
     - Single asterisk (`*`)
     - Asterisk in the middle (`vendor.*/gpu`)
--- a/pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
+++ b/pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
@@ -195,6 +195,7 @@ func findResourceConfigWithPrefix(resourceName string, resources map[v1.Resource
 
 	return bestConfig, found
 }
+
 func Score(task *api.TaskInfo, node *api.NodeInfo, weight ResourceStrategyFit) float64 {
 	score := 0.0
 	weightSum := 0


pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -137,6 +151,48 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		return false
+	}
+
+	asteriskCount := strings.Count(pattern, "*")
+	if asteriskCount != 1 {
+		return false
+	}
+
+	lastIndex := strings.LastIndex(pattern, "*")
+	return lastIndex == len(pattern)-1
```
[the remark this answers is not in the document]
ditingdapeng: Thank you for the review and suggestion.

While the logic `strings.Count(pattern, "*") == 1` would indeed ensure there's only one asterisk, it doesn't guarantee that the asterisk is at the **end** of the string, which is crucial for our prefix matching implementation.

Consider these examples:
- `"*gpu"` - has exactly 1 asterisk but it's at the beginning
- `"gpu*memory"` - has exactly 1 asterisk but it's in the middle
- `"gpu*"` - has exactly 1 asterisk and it's at the end ✓

Our current implementation specifically supports **suffix wildcard matching** (prefix matching with trailing asterisk) as mentioned in the issue discussion. The position check `lastIndex == len(pattern)-1` is essential to ensure we only accept patterns like `"cloudml.gpu*"` and reject patterns like `"*gpu"` or `"gpu*memory"`.

The current validation logic is intentionally designed this way to:
1. Ensure exactly one asterisk exists
2. Ensure that asterisk is at the end (enabling prefix matching)
3. Maintain clear semantics for the wildcard feature

If we want to optimize, we could use `strings.HasSuffix(pattern, "*")` instead of `strings.LastIndex`, but the position validation itself should remain.

LY-today:
@ditingdapeng Please refer to the content of this file and describe the details of the optimization. I will review it further later: docs/design/resource-strategy-fit-scheduling.md

ditingdapeng:
> @ditingdapeng Please refer to the content of this file and describe the details of the optimization. I will review it further later: docs/design/resource-strategy-fit-scheduling.md

Thank you for your comprehensive review! I have enhanced the documentation for the ResourceStrategyFit plugin's wildcard syntax support according to your recommendations.

ditingdapeng:
@JesseStutler 
Hello, could you please take a look at this PR when you have a moment? Thanks!

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -72,6 +75,76 @@ node score:
...
+    weight: 2
+  amd.com/gpu/*:           # Wildcard match for AMD GPUs
+    type: LeastAllocated
+    weight: 2
+  cpu:
+    type: LeastAllocated
+    weight: 1
+```
+
+**Implementation Details:**
```
JesseStutler, line 109: I think keep `###Configuration Guidelines` and `### Risk Analysis` are enough.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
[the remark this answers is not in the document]
JesseStutler: The meaning of gemini is like this:
```
    if !strings.HasSuffix(pattern, "*") {
		return false
	}

	return strings.Count(pattern, "*") == 1
```
This already ensures that * needs to be at the end and can only have one

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -104,6 +105,19 @@ func calculateWeight(args framework.Arguments) ResourceStrategyFit {
 	}
 
 	for k, v := range resources {
+		configKeyStr := string(k)
+
+		if strings.Contains(configKeyStr, "*") {
```
JesseStutler: Why do we need to add a `strings.Contains` check in the outer layer? Shouldn't this be put in `isValidWildcardPattern`?

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -137,6 +151,48 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+	if config, exists := resources[v1.ResourceName(resourceName)]; exists {
+		return config, true
+	}
+	var bestMatch string
+	var bestConfig ResourcesType
+	var found bool
+
+	for configKey, config := range resources {
+		configKeyStr := string(configKey)
+		if strings.HasSuffix(configKeyStr, "*") {
```
JesseStutler: The previous one has been verified by `isValidWildcardPattern`, so there is no need to add * here. We should put all the format validation logic in `isValidWildcardPattern`

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -137,6 +151,48 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+	if asteriskCount != 1 {
+		return false
+	}
+
+	lastIndex := strings.LastIndex(pattern, "*")
+	return lastIndex == len(pattern)-1
+}
+
+func findResourceConfigWithPrefix(resourceName string, resources map[v1.ResourceName]ResourcesType) (ResourcesType, bool) {
+	if config, exists := resources[v1.ResourceName(resourceName)]; exists {
```
JesseStutler, line 169: A comment can be added here like if exactMatch hits, exactMatch will prevail

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -72,6 +75,76 @@ node score:
...
+
+**Syntax Rules:**
+- Only suffix wildcard patterns are supported (e.g., `nvidia.com/gpu/*`)
+- Invalid patterns are filtered out during configuration parsing:
+    - Single asterisk (`*`)
+    - Asterisk in the middle (`vendor.*/gpu`)
+    - Asterisk at the beginning (`*/gpu`)
+    - Multiple asterisks (`vendor.com/**`)
+
+**Matching Priority:**
```
JesseStutler: This part can move to syntax rules

ditingdapeng:
@JesseStutler  Thanks for the feedback! I'll go through each comment and make revisions over the weekend, then submit.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
ditingdapeng: @JesseStutler  You're right. Centralizing all format validation logic within the isValidWildcardPattern function would indeed be more elegant.

I've considered two approaches:

**Approach 1: Using strings.Contains for Judgment**

This approach is simple and direct, but it doesn't fully adhere to the principle of centralizing all format validation logic in isValidWildcardPattern. It would still lead to wildcard-related logic being scattered across multiple locations.

**Approach 2: Modifying isValidWildcardPattern to Return Two Values**

This approach involves the function returning two boolean values: one to indicate if the format is valid, and another to indicate if it contains a wildcard. It can be implemented like this:

```go
func isValidWildcardPattern(pattern string) (bool, bool) {
    hasWildcard := strings.Contains(pattern, "*")
    if !hasWildcard {
        return false, false // Not a wildcard pattern, but format is valid
    }

    // Existing validation logic
    if pattern == "*" {
        return false, true // Is a wildcard, but format is invalid
    }

    if !strings.HasSuffix(pattern, "*") {
        return false, true // Is a wildcard, but format is invalid
    }

    if strings.Count(pattern, "*") > 1 {
        return false, true // Is a wildcard, but format is invalid
    }

    return true, true // Is a wildcard and format is valid
}
```

Then, in `findResourceConfigWithPrefix`, we can make the judgment like this:

`isValid, hasWildcard := isValidWildcardPattern(config.ResourceName)`


After discussion with  @LY-today , I adopted Solution 1 since it's sufficient to solve the problem. At this point in the code, we only need to detect the presence of wildcards.

ditingdapeng:
@JesseStutler  Hello, I have addressed all the review comments. Thank you for taking the time to review!

## Comment 1: Simplification of isValidWildcardPattern function
Your suggestion: "This already ensures that * needs to be at the end and can only have one"

My changes: ✅ Addressed Adopted your suggestion and simplified the original complex logic to:

```
if !strings.HasSuffix(pattern, "*") {
    return false
}
return strings.Count(pattern, "*") == 1
```

## Comment 2: Question about outer strings.Contains check
Your suggestion: "Why do we need to add a strings.Contains check in the outer layer? Shouldn't this be put in isValidWildcardPattern?"

My changes: ✅ Addressed Kept the strings.Contains(configKeyStr, "*") check for detection purposes, which is separate from the validation responsibility of isValidWildcardPattern . This follows the single responsibility principle.

## Comment 3: Removing redundant wildcard checks
Your suggestion: "The previous one has been verified by isValidWildcardPattern, so there is no need to add * here. We should put all the format validation logic in isValidWildcardPattern"

My changes: ✅ Addressed Removed the original strings.HasSuffix(configKeyStr, "*") check because:

- All format validation is now handled in isValidWildcardPattern
- Only need strings.Contains(configKeyStr, "*") here to detect wildcard patterns

## Comment 4: Documentation and wildcard-related comments
Your suggestions: "I think keep ###Configuration Guidelines and ### Risk Analysis are enough." and "A comment can be added here like if exactMatch hits, exactMatch will prevail"

My changes: ✅ Addressed Updated the documentation structure and added code comments as suggested.

ditingdapeng:
@JesseStutler  Morning, could you please take a look at this PR-comment when you have a moment? Thanks!

ditingdapeng:
@JesseStutler  Hi , just wanted to gently follow up on this PR. Is there any feedback or anything I can do to help move this forward? Thanks!

ditingdapeng:
Hi @JesseStutler  @Monokaix , I've just pushed the updates. This branch is now rebased on the latest `master` and the previous issues have been fixed.

Could you please take another look when you have a moment? Thanks!

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -137,6 +149,52 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+	}
+	// If no exact match found, search for wildcard patterns
+	var bestMatch string
+	var bestConfig ResourcesType
+	var found bool
+
+	for configKey, config := range resources {
+		configKeyStr := string(configKey)
+		// All wildcard patterns have been validated by isValidWildcardPattern
+		if strings.Contains(configKeyStr, "*") {
```
kingeasternsun, line 180: Why not just use `regexp`?

Monokaix:
Would like to know how many GPU models are generally configured. Regular matching reduces the configuration but is usually not intuitive enough.

ditingdapeng:
> Would like to know how many GPU models are generally configured. Regular matching reduces the configuration but is usually not intuitive enough.

Thanks for the feedback! This feature is based on real-world cluster requirements.

I just sampled GPU types from one of our production clusters and found 10 different GPU models :
```PlainText
"NodeGpuType":"cloudml.gpu/a10-24g"
"NodeGpuType":"cloudml.gpu/h20-96g"
"NodeGpuType":"cloudml.gpu/h20cc-96g"
"NodeGpuType":"cloudml.gpu/l20-45g"
"NodeGpuType":"cloudml.gpu/p4-8g"
"NodeGpuType":"cloudml.gpu/r39-24g"
"NodeGpuType":"cloudml.gpu/r49-24g"
"NodeGpuType":"cloudml.gpu/r49-48g"
"NodeGpuType":"cloudml.gpu/t4-16g"
"NodeGpuType":"cloudml.gpu/v100-32g"
```

Without wildcard support, we face two challenges:

1. Per-cluster configuration : Each cluster would need individual configuration files with different GPU types
2. Unified but verbose configuration : All clusters would need to configure all 10+ GPU types, even if they don't use them all

With wildcard syntax ( cloudml.gpu/* ) , we can:
- Use a single, maintainable configuration across clusters
- Automatically handle new GPU types without config updates
- Keep configurations clean and intuitive

While regex might seem less intuitive initially, the prefix-matching pattern ( cloudml.gpu/* ) is quite straightforward and significantly reduces operational overhead in multi-cluster environments.

docs/design/resource-strategy-fit-scheduling.md
```diff
@@ -72,6 +75,61 @@ node score:
...
+1. **Exact match** takes highest priority
+2. **Longest prefix match** for wildcard patterns
+
+**Configuration Example:**
+```yaml
+resources:
+  nvidia.com/gpu-v100:     # Exact match - highest priority
+    type: MostAllocated
+    weight: 3
+  nvidia.com/gpu/*:        # Wildcard match - covers other NVIDIA GPUs
```
Monokaix, line 98: How many match pattern will be supported, only wildcard? Do we need to  add other patterns or we only support one pattern?

ditingdapeng: Currently, this PR focuses on **wildcard pattern support** specifically with the `*` suffix syntax (e.g., `cloudml.gpu/*`, `nvidia.com/gpu/*`). Here's our current approach:

## Current Implementation
- **Single pattern type**: Wildcard with trailing `*` only
- **Validation**: Only patterns ending with exactly one `*` are considered valid
- **Matching logic**: Prefix-based matching with priority for exact matches

For now, we're keeping it focused on the wildcard suffix pattern as it solves the immediate operational pain points while maintaining a clean, extensible foundation.

## Future Extensibility
The current design is intentionally modular:
- `isValidWildcardPattern()` can be extended to support additional pattern types
- `findResourceConfigWithPrefix()` can accommodate different matching strategies
- The validation and matching logic are separated for easy extension

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
```diff
@@ -137,6 +149,52 @@ func (rsf *resourceStrategyFitPlugin) OnSessionOpen(ssn *framework.Session) {
...
+					bestConfig = config
+					found = true
+				}
+			}
+		}
+	}
+
+	return bestConfig, found
+}
 func Score(task *api.TaskInfo, node *api.NodeInfo, weight ResourceStrategyFit) float64 {
```
Monokaix, line 199: Please add a blank line.

ditingdapeng: Okay

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
ditingdapeng: I considered using regexp but chose the simple wildcard approach for these reasons:

1. **Simplicity and Performance**: Simple string operations like `strings.HasPrefix()` are much faster than regexp compilation and matching.

2. **User Experience**: Wildcard syntax is more intuitive for K8s administrators, similar to RBAC rules patterns.

3. **Security**: Limiting to simple patterns reduces regex injection risks and performance issues.

4. **Scope Alignment**: The primary use case is matching GPU resources like `nvidia.com/gpu-*` variants, which wildcard matching handles perfectly.

For example, with wildcard `nvidia.com/gpu/*`, we can match resources like:
- `nvidia.com/gpu/tesla-v100`
- `nvidia.com/gpu/rtx-3080` 
- `nvidia.com/gpu/a100`

This covers the common GPU resource naming patterns without regex complexity.

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
ditingdapeng: done

docs/design/resource-strategy-fit-scheduling.md
ditingdapeng: done

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
ditingdapeng: done

docs/design/resource-strategy-fit-scheduling.md
ditingdapeng: done

Monokaix:
/approve
Thanks!

Monokaix:
Please merge to one commit.

ditingdapeng:
@Thor-wl  @alcorj-mizar 
Hello, Could you please help review this PR when you have time? Thanks!

pkg/scheduler/plugins/resource-strategy-fit/resource_strategy_fit.go
kingeasternsun: @ditingdapeng @JesseStutler @Monokaix 
> 1. Regex and wildcard syntax are not mutually exclusive — regex still supports wildcard-style matching.
> 2. The regex can be compiled once during plugin initialization to avoid repeated overhead.
> 3. Regex also provides flexibility for future matching rules, making the approach more general and extensible.

But keeping the current implementation is also fine.

kingeasternsun:
> @Thor-wl @alcorj-mizar Hello, Could you please help review this PR when you have time? Thanks!

Please squash you commits to single one commit.

PR #4517, opened 2025-08-05
fix: update podGroup when statefulSet update

### LINKED ISSUE #4516: After updating statefulSet, podGroup belonging to the statefulSet will not update
### Description

The steps to reproduce the issue are as follows：
1. Create a StatefulSet where all its pod templates are scheduled by Volcano.
2. At first, pod request 10 gpu cards and podgroup's minResource is also 10 gpu cards, however queue only has 2 cards. The PodGroup is in the pending state because there are not enough resources in the queue.
3. I update the sts, adjust resource limits from 10 to 1 gpu cards. Pod is recreated. But podgroup will not updated. So it will still be pending.

After diving into the code, I found:
1. A PodGroup corresponds one-to-one with a StatefulSet. PodGroup's name will be marked as sts' uuid. So as if sts exists, podGroup will not be deleted.
2. In my case, there will be an addPod event and an updateStatefulSet event. However, when we found that podGroup exists, we will not update it even if pod's template has changed. 
```
func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
	pgName := helpers.GeneratePodgroupName(pod)

	if _, err := pg.pgLister.PodGroups(pod.Namespace).Get(pgName); err != nil {
		if !apierrors.IsNotFound(err) {
			klog.Errorf("Failed to get normal PodGroup for Pod <%s/%s>: %v",
				pod.Namespace, pod.Name, err)
			return err
		}

		var minMember = int32(1)
		var ownerAnnotations = make(map[string]string)
		if pg.inheritOwnerAnnotations {
			ownerAnnotations = pg.getAnnotationsFromUpperRes(pod)
			minMember = pg.getMinMemberFromUpperRes(ownerAnnotations, pod.Namespace, pod.Name)
		}
		minResources := util.CalTaskRequests(pod, minMember)
		obj := &scheduling.PodGroup{
				Namespace:       pod.Namespace,
				Name:            pgName,
				OwnerReferences: newPGOwnerReferences(pod),
				Annotations:     map[string]string{},
				Labels:          map[string]string{},
			},
			Spec: scheduling.PodGroupSpec{
				MinMember:         minMember,
				PriorityClassName: pod.Spec.PriorityClassName,
				MinResources:      &minResources,
			},
			Status: scheduling.PodGroupStatus{
				Phase: scheduling.PodGroupPending,
			},
		}

		pg.inheritUpperAnnotations(ownerAnnotations, obj)
		// Individual annotations on pods would overwrite annotations inherited from upper resources.
		if queueName, ok := pod.Annotations[scheduling.QueueNameAnnotationKey]; ok {
			obj.Spec.Queue = queueName
		}

		if value, ok := pod.Annotations[scheduling.PodPreemptable]; ok {
			obj.Annotations[scheduling.PodPreemptable] = value
		}
		if value, ok := pod.Annotations[scheduling.CooldownTime]; ok {
			obj.Annotations[scheduling.CooldownTime] = value
		}
		if value, ok := pod.Annotations[scheduling.RevocableZone]; ok {
			obj.Annotations[scheduling.RevocableZone] = value
		}
		if value, ok := pod.Labels[scheduling.PodPreemptable]; ok {
			obj.Labels[scheduling.PodPreemptable] = value
		}
		if value, ok := pod.Labels[scheduling.CooldownTime]; ok {
			obj.Labels[scheduling.CooldownTime] = value
		}

		if value, found := pod.Annotations[scheduling.JDBMinAvailable]; found {
			obj.Annotations[scheduling.JDBMinAvailable] = value
		} else if value, found := pod.Annotations[scheduling.JDBMaxUnavailable]; found {
			obj.Annotations[scheduling.JDBMaxUnavailable] = value
		}

		if _, err := pg.vcClient.SchedulingV1beta1().PodGroups(pod.Namespace).Create(context.TODO(), obj, metav1.CreateOptions{}); err != nil {
			if !apierrors.IsAlreadyExists(err) {
				klog.Errorf("Failed to create normal PodGroup for Pod <%s/%s>: %v",
					pod.Namespace, pod.Name, err)
				return err
			} else {
				klog.V(4).Infof("PodGroup <%s/%s> already exists for Pod <%s/%s>",
					pod.Namespace, pgName, pod.Namespace, pod.Name)
			}
		} else {
			klog.V(4).Infof("PodGroup <%s/%s> created for Pod <%s/%s>",
				pod.Namespace, pgName, pod.Namespace, pod.Name)
		}
	}

	return pg.updatePodAnnotations(pod, pgName)
}
```

1. Create a StatefulSet where all its pod templates are scheduled by Volcano.
2. At first, pod request 10 gpu cards and podgroup's minResource is also 10 gpu cards, however queue only has 2 cards. The PodGroup is in the pending state because there are not enough resources in the queue.
3. I update the sts, adjust resource limits from 10 to 1 gpu cards. Pod is recreated. But podgroup will not updated. So it will still be pending.

PodGroup has been updated and pod is scheduled succesussfully.

master

_No response_

Poor12:
Fixes #4516

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -19,6 +19,7 @@ package podgroup
 import (
 	"context"
 	"encoding/json"
+	"reflect"
 	"slices"
 	"strconv"
 	"strings"
@@ -35,6 +36,10 @@ import (
 	"volcano.sh/apis/pkg/apis/helpers"
 	scheduling "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	"volcano.sh/volcano/pkg/controllers/util"
+)
+
+const (
+	controllerRevisionHashLabelKey = "controller-revision-hash"
 )
 
 type podRequest struct {
@@ -130,7 +135,9 @@ func (pg *pgcontroller) addStatefulSet(obj interface{}) {
 	// the updateStatefulSet(replicas=1) event, and after the addPod event for the new created pod.
 	// In this event, need to create PodGroup for the pod.
 	if *sts.Spec.Replicas > 0 {
-		selector := metav1.LabelSelector{MatchLabels: sts.Spec.Selector.MatchLabels}
+		matchLabels := sts.Spec.Selector.MatchLabels
+		matchLabels[controllerRevisionHashLabelKey] = sts.Status.UpdateRevision
+		selector := metav1.LabelSelector{MatchLabels: matchLabels}
 		labelSelector, err := metav1.LabelSelectorAsSelector(&selector)
 		if err != nil {
 			klog.Errorf("Failed to convert label selector for StatefulSet <%s/%s>: %v", sts.Namespace, sts.Name, err)
@@ -143,20 +150,22 @@ func (pg *pgcontroller) addStatefulSet(obj interface{}) {
 		}
 		if len(pods) > 0 {
 			pod := pods[0]
-			klog.V(4).Infof("Try to create podgroup for pod %s/%s", pod.Namespace, pod.Name)
+			klog.V(4).Infof("Try to create or update podgroup for pod %s/%s when statefulset add or update", pod.Namespace, pod.Name)
 			if !slices.Contains(pg.schedulerNames, pod.Spec.SchedulerName) {
 				klog.V(4).Infof("Pod %s field SchedulerName is not matched", klog.KObj(pod))
 				return
 			}
+
 			// If the pod is already associated with a podgroup, skip creating a new one. This scenario is applicable to LeaderWorkerSet,
 			// which will create podgroups by itself, and Volcano does not need to create a podgroup for statefulset.
 			if pgName := pod.Annotations[scheduling.KubeGroupNameAnnotationKey]; pgName != "" {
 				klog.V(4).Infof("Pod %s is already associated with a podgroup %s", klog.KObj(pod), pgName)
 				return
 			}
-			err := pg.createNormalPodPGIfNotExist(pod)
+
+			err := pg.createOrUpdateNormalPodPG(pod)
 			if err != nil {
-				klog.Errorf("Failed to create PodGroup for pod <%s/%s>: %v", pod.Namespace, pod.Name, err)
+				klog.Errorf("Failed to create or update PodGroup for pod <%s/%s>: %v", pod.Namespace, pod.Name, err)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -352,6 +353,168 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
...
+		podGroup.Annotations = desiredAnnotations
+		isUpdated = true
+	}
+	if !reflect.DeepEqual(desiredLabels, podGroup.Labels) {
+		podGroup.Labels = desiredLabels
+		isUpdated = true
+	}
+
+	return isUpdated
+}
```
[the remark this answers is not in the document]
JesseStutler: @Poor12 Please check

JesseStutler:
Also we should add UTs and even better add an e2e testing case for update scenario

Poor12:
> Also we should add UTs and even better add an e2e testing case for update scenario

I'll do it later.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -281,75 +282,182 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
...
+	}
 
-		if value, found := pod.Annotations[scheduling.JDBMinAvailable]; found {
-			obj.Annotations[scheduling.JDBMinAvailable] = value
-		} else if value, found := pod.Annotations[scheduling.JDBMaxUnavailable]; found {
-			obj.Annotations[scheduling.JDBMaxUnavailable] = value
+	return nil
+}
+
+func (pg *pgcontroller) updateExistingPodGroup(podGroup *scheduling.PodGroup, pod *v1.Pod) bool {
```
Monokaix: If  we add other fields on pg at the future, we need to modify here, it's not a good way.

Poor12: Updated. Also add some UT tests.

pkg/controllers/podgroup/pg_controller_handler.go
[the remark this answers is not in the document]
Poor12: Updated.

pkg/controllers/podgroup/pg_controller_handler.go
Monokaix: Threre is a hash value on the pod, how about just recording this hash value when first time pg is created, and use it to compare next time.

Poor12: Do you mean pod's resourceVersion? I think not every pod update requires triggering a podgroup update.

Monokaix: Negative. The hash of pod spec.
image

Poor12: There's a misunderstanding here: if a pod's spec is updated, does the PodGroup necessarily have to be updated as well? From the perspective of the logic for generating a PodGroup, it is not necessarily the case.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -281,60 +288,37 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
...
-		if value, ok := pod.Annotations[scheduling.RevocableZone]; ok {
-			obj.Annotations[scheduling.RevocableZone] = value
-		}
-		if value, ok := pod.Labels[scheduling.PodPreemptable]; ok {
-			obj.Labels[scheduling.PodPreemptable] = value
-		}
-		if value, ok := pod.Labels[scheduling.CooldownTime]; ok {
-			obj.Labels[scheduling.CooldownTime] = value
-		}
+func (pg *pgcontroller) createOrUpdateNormalPodPG(pod *v1.Pod) error {
```
Monokaix, line 320: Please also add a comment to describe why we need to update pg in sts case.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -143,14 +150,14 @@ func (pg *pgcontroller) addStatefulSet(obj interface{}) {
 		}
 		if len(pods) > 0 {
 			pod := pods[0]
-			klog.V(4).Infof("Try to create podgroup for pod %s/%s", pod.Namespace, pod.Name)
+			klog.V(4).Infof("Try to create or update podgroup for pod %s/%s", pod.Namespace, pod.Name)
```
Monokaix:
```suggestion
			klog.V(4).Infof("Try to create or update podgroup for pod %s/%s" when statefulset add or update, pod.Namespace, pod.Name)
```

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -347,11 +333,99 @@ func (pg *pgcontroller) createNormalPodPGIfNotExist(pod *v1.Pod) error {
...
+	if value, found := pod.Annotations[scheduling.JDBMinAvailable]; found {
+		obj.Annotations[scheduling.JDBMinAvailable] = value
+	} else if value, found := pod.Annotations[scheduling.JDBMaxUnavailable]; found {
+		obj.Annotations[scheduling.JDBMaxUnavailable] = value
+	}
+
+	return obj
+}
+
+func (pg *pgcontroller) updateExistingPodGroup(podGroup *scheduling.PodGroup, pod *v1.Pod) bool {
```
Monokaix:
```suggestion
func (pg *pgcontroller) ShouldUpdateExistingPodGroup(podGroup *scheduling.PodGroup, pod *v1.Pod) bool {
```

pkg/controllers/podgroup/pg_controller_test.go
```diff
@@ -353,3 +353,313 @@ func TestAddPodGroup(t *testing.T) {
 		}
 	}
 }
+
```
Monokaix: Can we also add a test case that create a exising pg, can should call update method of pg when sts is updated?

Poor12: I may have not enough time to do it. Maybe we can seek help from other members of the community. :)

Poor12: Added.

PR #4520, opened 2025-08-08
Fix queue not counting volcano.sh/vgpu-memory and volcano.sh/vgpu-number properly

archlitchi:
/BUG

related issue: https://github.com/volcano-sh/volcano/issues/4168

### DIFF
--- a/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
+++ b/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
@@ -117,7 +117,7 @@ func NewGPUDevices(name string, node *v1.Node) *GPUDevices {
 	if strings.Contains(handshake, "Requesting") {
 		formertime, _ := time.Parse("2006.01.02 15:04:05", strings.Split(handshake, "_")[1])
 		if time.Now().After(formertime.Add(time.Second * 60)) {
-			klog.Infof("node %v device %s leave", node.Name, handshake)
+			klog.V(3).Infof("node %v device %s leave", node.Name, handshake)
 
 			tmppat := make(map[string]string)
 			tmppat[deviceconfig.VolcanoVGPUHandshake] = "Deleted_" + time.Now().Format("2006.01.02 15:04:05")
@@ -144,7 +144,33 @@ func (gs *GPUDevices) ScoreNode(pod *v1.Pod, schedulePolicy string) float64 {
 }
 
 func (gs *GPUDevices) GetIgnoredDevices() []string {
-	return []string{deviceconfig.VolcanoVGPUMemory, deviceconfig.VolcanoVGPUMemoryPercentage, deviceconfig.VolcanoVGPUCores}
+	return []string{}
+}
+
+func (gs *GPUDevices) AddQueueResource(pod *v1.Pod) map[string]float64 {
+	if gs == nil {
+		return map[string]float64{}
+	}
+	klog.V(5).InfoS("AddQueueResource", "Name", pod.Name)
+	res := map[string]float64{}
+	ids, ok := pod.Annotations[AssignedIDsAnnotations]
+	if !ok {
+		klog.Errorf("pod %s has no annotation volcano.sh/devices-to-allocate", pod.Name)
+		return res
+	}
+	podDev := decodePodDevices(ids)
+	for _, val := range podDev {
+		for _, deviceused := range val {
+			for _, gsdevice := range gs.Device {
+				if strings.Contains(deviceused.UUID, gsdevice.UUID) {
+					res[getConfig().ResourceMemoryName] += float64(deviceused.Usedmem * 1000)
+					res[getConfig().ResourceCoreName] += float64(deviceused.Usedcores * 1000)
+				}
+			}
+		}
+	}
+	klog.V(4).InfoS("AddQueueResource", "Name=", pod.Name, "res=", res)
+	return res
 }
 
 // AddResource adds the pod to GPU pool if it is assigned
--- a/pkg/scheduler/api/devices/nvidia/vgpu/utils.go
+++ b/pkg/scheduler/api/devices/nvidia/vgpu/utils.go
@@ -158,11 +158,11 @@ func decodePodDevices(str string) []ContainerDevices {
 
 func checkVGPUResourcesInPod(pod *v1.Pod) bool {
 	for _, container := range pod.Spec.Containers {
-		_, ok := container.Resources.Limits[config.VolcanoVGPUMemory]
+		_, ok := container.Resources.Limits[v1.ResourceName(getConfig().ResourceMemoryName)]
 		if ok {
 			return true
 		}
-		_, ok = container.Resources.Limits[config.VolcanoVGPUNumber]
+		_, ok = container.Resources.Limits[v1.ResourceName(getConfig().ResourceCountName)]
 		if ok {
 			return true
 		}
@@ -171,10 +171,10 @@ func checkVGPUResourcesInPod(pod *v1.Pod) bool {
 }
 
 func resourcereqs(pod *v1.Pod) []ContainerDeviceRequest {
-	resourceName := v1.ResourceName(config.VolcanoVGPUNumber)
-	resourceMem := v1.ResourceName(config.VolcanoVGPUMemory)
-	resourceMemPercentage := v1.ResourceName(config.VolcanoVGPUMemoryPercentage)
-	resourceCores := v1.ResourceName(config.VolcanoVGPUCores)
+	resourceName := v1.ResourceName(getConfig().ResourceCountName)
+	resourceMem := v1.ResourceName(getConfig().ResourceMemoryName)
+	resourceMemPercentage := v1.ResourceName(getConfig().ResourceMemoryPercentageName)
+	resourceCores := v1.ResourceName(getConfig().ResourceCoreName)
 	counts := []ContainerDeviceRequest{}
 
 	//Count Nvidia GPU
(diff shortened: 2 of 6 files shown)


pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
```diff
@@ -144,7 +144,33 @@ func (gs *GPUDevices) ScoreNode(pod *v1.Pod, schedulePolicy string) float64 {
...
+		klog.Errorf("pod %s has no annotation volcano.sh/devices-to-allocate", pod.Name)
+		return res
+	}
+	podDev := decodePodDevices(ids)
+	for _, val := range podDev {
+		for _, deviceused := range val {
+			for _, gsdevice := range gs.Device {
+				if strings.Contains(deviceused.UUID, gsdevice.UUID) {
+					res[getConfig().ResourceMemoryName] += float64(deviceused.Usedmem * 1000)
+					res[getConfig().ResourceCoreName] += float64(deviceused.Usedcores * 1000)
```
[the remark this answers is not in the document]
archlitchi: No, ConvertRes2ResList treats all scalar resources as Milli-units, we need to multiply here in order for it to work properly

pkg/scheduler/api/devices/nvidia/vgpu/utils.go
```diff
@@ -500,3 +500,21 @@ func patchNodeAnnotations(node *v1.Node, annotations map[string]string) error {
...
+		ResourceMemoryName:  config.VolcanoVGPUMemory,
+		DefaultMemory:       0,
+		DefaultCores:        0,
+		DefaultGPUNum:       1,
+		DeviceSplitCount:    10,
+		DeviceMemoryScaling: 1,
+		DeviceCoreScaling:   1,
+		DisableCoreLimit:    false,
+	}
+}
```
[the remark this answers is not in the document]
archlitchi: adopted

archlitchi:
CC @JesseStutler

archlitchi:
The result can be seen as follows:

81ab990a-ca36-440b-a4d1-cfef6cbe72e1

pkg/scheduler/framework/session.go
```diff
@@ -259,6 +259,18 @@ func updateQueueStatus(ssn *Session) {
 		for status, tasks := range job.TaskStatusIndex {
 			if api.AllocatedStatus(status) {
 				for _, task := range tasks {
+					node, ok1 := ssn.Nodes[task.NodeName]
```
JesseStutler: Should not name variable as `ok1`. `ok1` --> `ok`

pkg/scheduler/framework/session.go
```diff
@@ -259,6 +259,18 @@ func updateQueueStatus(ssn *Session) {
...
+					taskReq := task.Resreq
+					if ok1 {
+						for _, sharedDevices := range node.Others {
+							if sharedDevices.(api.Devices).HasDeviceRequest(task.Pod) {
```
JesseStutler: I'm wondering that why `Others` field in `NodeInfo` is a `map[string]interface{}` type rather than `map[string]Devices` ? Will `Others` have other types of interface in the future? We cannot control other contributors who may put other interface implementations into Others. If other interfaces do not have `HasDeviceRequest` methods, there will be problems here.

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
```diff
@@ -144,7 +144,33 @@ func (gs *GPUDevices) ScoreNode(pod *v1.Pod, schedulePolicy string) float64 {
...
+	if !ok {
+		klog.Errorf("pod %s has no annotation volcano.sh/devices-to-allocate", pod.Name)
+		return res
+	}
+	podDev := decodePodDevices(ids)
+	for _, val := range podDev {
+		for _, deviceused := range val {
+			for _, gsdevice := range gs.Device {
+				if strings.Contains(deviceused.UUID, gsdevice.UUID) {
+					res[getConfig().ResourceMemoryName] += float64(deviceused.Usedmem * 1000)
```
JesseStutler, line 166: Sorry not so familiar with the previous codes, where is `VolcanoVGPUNumber` calculated? Can it be correctly recorded in queue allocated field?

archlitchi: volcanoVGPUNumber doesn't need to be calculated here, it is correctly counted in default behavior

pkg/scheduler/framework/session.go
archlitchi: haha, the nodeInfo.Others field exists before i implement deviceshare API. It's william-wang's idea to put devices related logics here, so we don't need to change the node_info struct

JesseStutler: If we don't modify NodeInfo, there should be an assertion here

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
```diff
@@ -144,7 +144,33 @@ func (gs *GPUDevices) ScoreNode(pod *v1.Pod, schedulePolicy string) float64 {
...
 
 func (gs *GPUDevices) GetIgnoredDevices() []string {
-	return []string{deviceconfig.VolcanoVGPUMemory, deviceconfig.VolcanoVGPUMemoryPercentage, deviceconfig.VolcanoVGPUCores}
+	return []string{}
+}
+
+func (gs *GPUDevices) AddQueueResource(pod *v1.Pod) map[string]float64 {
+	klog.Infoln("AddQueueResource:Name=", pod.Name)
+	res := map[string]float64{}
+	if gs == nil {
```
Monokaix, line 151: Better move this to top.

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
```diff
@@ -144,7 +144,33 @@ func (gs *GPUDevices) ScoreNode(pod *v1.Pod, schedulePolicy string) float64 {
 }
 
 func (gs *GPUDevices) GetIgnoredDevices() []string {
-	return []string{deviceconfig.VolcanoVGPUMemory, deviceconfig.VolcanoVGPUMemoryPercentage, deviceconfig.VolcanoVGPUCores}
+	return []string{}
+}
+
+func (gs *GPUDevices) AddQueueResource(pod *v1.Pod) map[string]float64 {
+	klog.Infoln("AddQueueResource:Name=", pod.Name)
```
Monokaix: klog level 3？

Monokaix:
Can `Allocateable` and `Enqueueable` be aware of vgpu resources? If queue.status has vgpu resources, we should also check the vgpu resource quota when allocate.

archlitchi:
> Can `Allocateable` and `Enqueueable` be aware of vgpu resources? If queue.status has vgpu resources, we should also check the vgpu resource quota when allocate.

currently no, because you can't know exactly how much device memory a task uses before scheduling, for example, i task allocate volcano.sh/vgpu-number:1, it allocates a whole card, but different card has different device memory, which can't be decided before scheduling.

Monokaix:
> > Can `Allocateable` and `Enqueueable` be aware of vgpu resources? If queue.status has vgpu resources, we should also check the vgpu resource quota when allocate.
> 
> currently no, because you can't know exactly how much device memory a task uses before scheduling, for example, i task allocate volcano.sh/vgpu-number:1, it allocates a whole card, but different card has different device memory, which can't be decided before scheduling.

`deviceshare` plugin can be aware of that, right? device memory quota check in queue is just like task's memory request check in resource filter.

archlitchi:
> > > Can `Allocateable` and `Enqueueable` be aware of vgpu resources? If queue.status has vgpu resources, we should also check the vgpu resource quota when allocate.
> > 
> > 
> > currently no, because you can't know exactly how much device memory a task uses before scheduling, for example, i task allocate volcano.sh/vgpu-number:1, it allocates a whole card, but different card has different device memory, which can't be decided before scheduling.
> 
> `deviceshare` plugin can be aware of that, right? device memory quota check in queue is just like task's memory request check in resource filter.

yes, but that's in the process of scheduling, so that task is still in the wrong state, because such task shouldn't be enqueue

Monokaix:
> > > > Can `Allocateable` and `Enqueueable` be aware of vgpu resources? If queue.status has vgpu resources, we should also check the vgpu resource quota when allocate.
> > > 
> > > 
> > > currently no, because you can't know exactly how much device memory a task uses before scheduling, for example, i task allocate volcano.sh/vgpu-number:1, it allocates a whole card, but different card has different device memory, which can't be decided before scheduling.
> > 
> > 
> > `deviceshare` plugin can be aware of that, right? device memory quota check in queue is just like task's memory request check in resource filter.
> 
> yes, but that's in the process of scheduling, so that task is still in the wrong state, because such task shouldn't be enqueue

But enqueueable is also a callback, and plugin can also implement it.

archlitchi:
@Monokaix The problem is not we can't implement, it is we don't know the exact device memory a certain task uses before scheduling

pkg/scheduler/framework/session.go
```diff
@@ -259,6 +259,18 @@ func updateQueueStatus(ssn *Session) {
 		for status, tasks := range job.TaskStatusIndex {
 			if api.AllocatedStatus(status) {
 				for _, task := range tasks {
+					node, ok := ssn.Nodes[task.NodeName]
+					taskReq := task.Resreq
+					if ok {
+						for _, sharedDevices := range node.Others {
```
Monokaix: sharedDevices is nil when node has shared gpu resource, will it panic here?

pkg/scheduler/framework/session.go
```diff
@@ -259,6 +259,18 @@ func updateQueueStatus(ssn *Session) {
 		for status, tasks := range job.TaskStatusIndex {
 			if api.AllocatedStatus(status) {
 				for _, task := range tasks {
+					node, ok := ssn.Nodes[task.NodeName]
```
Monokaix: We can wrap this as a func.

pkg/scheduler/framework/session.go
archlitchi: In current version , either node.Others is empty and this for loop is not entered, or sharedDevices is not nil because we have initialized it. so panic won't happen

PR #4526, opened 2025-08-11
fix: deepcopy podgroup before update

Poor12:
#### What type of PR is this?
The sts/pg object is retrieved from the informer cache and should be treated as read-only. Modifying sts.Spec.Selector.MatchLabels/pg directly can lead to race conditions and unpredictable behavior in the controller.  So deepcopy it befor update.


JesseStutler:
> /cc @JesseStutler I just found another place may occur the same situation and update it in the same time. Could you please help check the code?

OK. Thanks
/priority high
/kind bug

pkg/controllers/job/job_controller_actions.go
```diff
@@ -778,22 +778,23 @@ func (cc *jobcontroller) createOrUpdatePodGroup(job *batch.Job) error {
 		return nil
 	}
 
+	podGroupToUpdate := pg.DeepCopy()
```
Monokaix, line 781: We can also wrap a func here.

Poor12: Updated

JesseStutler (commented):
/lgtm
Thanks, we can cherry-pick this commit in the same PR: #4522  @Poor12

PR #4541, opened 2025-08-13
[Colocation Enhancement] Support CPU QoS

### LINKED ISSUE #4468: [Colocation Enhancement] Support CPU QoS

Volcano agent currently lacks direct configuration for CPU throttling thresholds and protection watermarks. We need these two configurations in the CPU QoS settings

We need CPU throttling thresholds and protection watermarks in CPU QoS setting:
- CPU Throttling Threshold: A percentage or absolute value at which CPU throttling should begin
- CPU Protection Watermark: A percentage or absolute value below which CPU throttling should stop

### Additional context

_No response_

Haibara-Ai97:
#### What type of PR is this?
/kind feature

This PR introduces a CPU Throttle feature to Volcano Agent that dynamically limit CPU usage of low-priority Pods when node CPU resources are under pressure.

The implementation follows a probe-event-handler pattern where the node monitor continuously tracks CPU usage and generates throttling events when thresholds are crossed. These events are processed by a dedicated CPU throttle handler that manipulates cgroup CPU quotas through the filesystem interface. The throttling process is gradual, reducing Pod CPU limits by small increments (default 10%) to avoid sudden performance degradation. Meantime, the protection watermark design ensures that the Pod's CPU quota will not be constantly limited, affecting basic performance.

The feature integrates seamlessly with Volcano's existing QoS framework, ensuring that only appropriate Pods are throttled while critical system workloads remain protected. Configuration is fully dynamic through the existing ColocationConfig mechanism, allowing operators to tune thresholds and throttling behavior at runtime.

Fixes #4468

This PR includes unit tests for both cpu throttle handler and monitor, and the tests have passed.

```release-note
Operators can configure cpuThrottlingConfig in configmap to adjust the ThrottlingThreshold, ProtectionWatermark, ThrottlingStepPercent and MinQuotaPercent.
```


pkg/agent/events/probes/nodemonitor/node_monitor.go
```diff
@@ -200,3 +212,83 @@ func (m *monitor) nodeHasPressure(resName v1.ResourceName) bool {
...
+	usage := m.usageGetter.UsagesByPercentage(node)
+	cpuUsage := usage[v1.ResourceCPU]
+
+	klog.V(4).InfoS("CPU throttling detection",
+		"usage", cpuUsage,
+		"throttlingThreshold", throttlingThreshold,
+		"protectionWatermark", protectionWatermark,
+		"active", m.cpuThrottlingActive)
+
+	// 启动限流：CPU使用率超过阈值且当前未激活限流
```
JesseStutler: Please translate all the comments into English, thanks

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
```diff
@@ -0,0 +1,233 @@
...
+
+func NewCPUThrottleHandler(config *config.Configuration, mgr *metriccollect.MetricCollectorManager,
+	cgroupMgr cgroup.CgroupManager) framework.Handle {
+	return &CPUThrottleHandler{
+		BaseHandle: &base.BaseHandle{
+			Name:   string(features.CPUThrottleFeature),
+			Config: config,
+		},
+		cgroupMgr:   cgroupMgr,
+		getPodsFunc: config.GetActivePods,
```
JesseStutler: The `getPodsFunc` here actually can't use this `GetActivePods`, we only need to suppress BE pods. BE pods are either explicitly specified as BE in the pod annotation with "volcano.sh/qos-level", or they are k8s pods themselves with the qosClass set to BE.

Even, we don't need to get pods, just set the root cgroup: '/sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice' (take systemd as an example) to the changed available CPU usage

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
```diff
@@ -0,0 +1,233 @@
...
+		return fmt.Errorf("unknown cpu throttle action: %v", cpuEvent.Action)
+	}
+}
+
+func (h *CPUThrottleHandler) stepThrottleCPU(pods []*v1.Pod) error {
+	h.mutex.Lock()
+	defer h.mutex.Unlock()
+
+	for _, pod := range pods {
+		qosLevel := extension.GetQosLevel(pod)
```
JesseStutler: As I mentioned before, if `getPodsFunc` has already filtered only BE pods, there's actually no need to check the QoSLevel

But I'm thinking that we actually don' t need to retrieve the pods, just need to set the new cpu quota on the BE pods' root cgroup

pkg/agent/events/handlers/cputhrottle/utils.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	return minQuota
+}
+
+// applyCPUQuota writes the new CPU quota to the cgroup filesystem to enforce
+// the CPU limit at the kernel level. This is how Kubernetes actually implements
+// CPU resources.
+func (h *CPUThrottleHandler) applyCPUQuota(pod *v1.Pod, quota int64) (string, error) {
+	cgroupPath, err := h.cgroupMgr.GetPodCgroupPath(pod.Status.QOSClass, cgroup.CgroupCpuSubsystem, pod.UID)
```
JesseStutler: Actually, we just need to suppress the BE pods, therefore we don't need to get the cgroup path according to the pod's QoSClass, we just need to set the cpu quota for the BE pods' root cgroup '/sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice' (take systemd as an example)

pkg/agent/events/probes/nodemonitor/node_monitor.go
```diff
@@ -156,6 +166,8 @@ func (m *monitor) detect() {
 		}
 	}
 
+	m.detectCPUThrottling(nodeCopy)
```
JesseStutler: I think cpu throtting and eviction can seperate, we can rename the `detect` here to like `detectEvict`, and open a seperate goroutine for `detectCPUThrottling` in monitor's Run method

pkg/agent/events/probes/nodemonitor/node_monitor.go
```diff
@@ -200,3 +212,83 @@ func (m *monitor) nodeHasPressure(resName v1.ResourceName) bool {
...
+	cpuUsage := usage[v1.ResourceCPU]
+
+	klog.V(4).InfoS("CPU throttling detection",
+		"usage", cpuUsage,
+		"throttlingThreshold", throttlingThreshold,
+		"protectionWatermark", protectionWatermark,
+		"active", m.cpuThrottlingActive)
+
+	// 启动限流：CPU使用率超过阈值且当前未激活限流
+	if !m.cpuThrottlingActive && cpuUsage >= int64(throttlingThreshold) {
```
JesseStutler: Actually, I need to clarify the background of this requirement. We need to colocate the offline and online workloads, so the priority is to ensure online workloads (QoS > 0). Therefore, we need to suppress offline workloads (QoS < 0). However, using node utilization as the sole criterion is not accurate, because even without offline workloads, the CPU utilization of online workloads may reach this threshold. So, who should be suppressed in such scenario? Therefore, we should calculate the available CPU quota for BE pods (total node CPU * throttingThreshold - resource usage of non-BE pods - system resource usage), and then send it to the handler. The handler adjusts the BE pod cgroup CPU quota based on this value.

docs/design/cpu-throttle-design.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Event Layer
+
+Event Type: `NodeCPUThrottleEvent`
+
+```go
+type NodeCPUThrottleEvent struct {
+    TimeStamp time.Time
+    Resource  v1.ResourceName
+    Action    string // "start", "continue", "stop"
```
JesseStutler, line 50: Actually, I think we don't need this `Action`. The probe should periodically probe and send the expected CPU quota value of the current BE pods to the handler. Then the handler should directly adjust the cgroup of the BE pods based on this value (if the CPU quota of the BE pods is exceeded).

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
```diff
@@ -0,0 +1,233 @@
+package cputhrottle
+
+import (
+	"errors"
+	"fmt"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"os"
+	"sync"
+	"volcano.sh/volcano/pkg/agent/apis/extension"
```
JesseStutler: Please notice to sort the packages according to
```
go packages 

third party packages

volcano's own packages
```

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	klog.InfoS("Handling CPU throttling event",
+		"action", cpuEvent.Action,
+		"usage", cpuEvent.Usage,
+		"podCount", len(pods))
+
+	switch cpuEvent.Action {
+	case "start", "continue":
+		return h.stepThrottleCPU(pods)
```
JesseStutler: I think BE pods should actually be suppressed instantly (if the usage rate of BE pods exceeds expectations), and restored step by step, because restoring BE pods all at once might affect the CPU usage of online workloads, and we need to prioritize ensuring the quality of online workloads

JesseStutler:
I think we only need the `throttingThreshold`, no need for `protectionWaterMark`. The CPU quota available for BE pods is node available CPU * `throttingThreshold` - CPU usage by non-BE pods - node system usage

Haibara-Ai97:
@JesseStutler This part code has been modified to do CPU throttleing based on the available CPU quota for Best Effort pods.

JesseStutler:
@Haibara-Ai97 Please fix the code verify CI, there are some unused codes

pkg/agent/events/probes/nodemonitor/node_monitor.go
```diff
@@ -86,18 +93,24 @@ func (m *monitor) Run(stop <-chan struct{}) {
 	klog.InfoS("Started nodePressure probe")
 	go wait.Until(m.utilizationMonitoring, 10*time.Second, stop)
 	go wait.Until(m.detect, 10*time.Second, stop)
```
JesseStutler: Better to rename to this `detect` func to `detectEviction`

pkg/agent/events/probes/nodemonitor/node_monitor.go
```diff
@@ -200,3 +213,70 @@ func (m *monitor) nodeHasPressure(resName v1.ResourceName) bool {
...
+	totalMilli := totalCPU.MilliValue()
+
+	pods, err := m.getPodsFunc()
+	if err != nil {
+		klog.ErrorS(err, "CPU Throttling: Failed to get pods")
+	}
+
+	var onlineRequestMilli int64
+	for _, pod := range pods {
+		if extension.GetQosLevel(pod) < 0 {
```
JesseStutler: `GetQosLevel` only retrieves pods that have defined Volcano QoS levels annotation.But If a pod doesn't have a Volcano annotation defined (e.g., there are regular pods on this node), then QoSLevel returns 0. In this case, pods using native k8s besteffort qosClass(in pod.status.qosclass) will also be considered as qoslevel 0 and will be counted in onlineRequestMilli.

Further modifications to the GetQoSLevel function are needed, or a TODO for future modifications could be added.

BTW, some comments could be added here to explain how the availability of BE pods is calculated.

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
JesseStutler, line 1: Please also add author header file for all the files(except markdown)

pkg/agent/events/probes/nodemonitor/node_monitor.go
```diff
@@ -200,3 +213,70 @@ func (m *monitor) nodeHasPressure(resName v1.ResourceName) bool {
...
+
+	allowedMilli := totalMilli * int64(throttlingThreshold) / 100
+	availableBEMilli := allowedMilli - onlineRequestMilli
+	if availableBEMilli < 0 {
+		availableBEMilli = 0
+	}
+
+	// CPU throttling is applied when the available quota of Best Effort pods is less
+	// than 10% of the throttling threshold
+	if availableBEMilli < allowedMilli/10 {
```
JesseStutler: I don't quite understand why we need such restrictions?

pkg/agent/config/api/types.go
```diff
@@ -111,3 +114,9 @@ type Evicting struct {
 	// EvictingMemoryLowWatermark defines the low watermark percent of memory usage when the node could recover schedule pods.
 	EvictingMemoryLowWatermark *int `json:"evictingMemoryLowWatermark,omitempty"`
 }
+
+type CPUThrottling struct {
```
JesseStutler, line 118: Should also add default CPUThrottling config in utils.go ( disable by default )

docs/design/cpu-throttle-design.md
JesseStutler, line 1: Please update the doc to make sure current design doc is consistent with the current implementation @Haibara-Ai97

pkg/agent/events/handlers/cputhrottle/utils.go
JesseStutler, line 1: So do we still need this file? I see that all the funcs didn't been used currently

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	return filePath, nil
+}
+
+func (h *CPUThrottleHandler) RefreshCfg(cfg *api.ColocationConfig) error {
+	h.mutex.Lock()
+	defer h.mutex.Unlock()
+
+	isActive, err := features.DefaultFeatureGate.Enabled(features.CPUThrottleFeature, cfg)
```
JesseStutler, line 147: We should also add a judgement in `pkg/agent/features/feature_gate.go`, only when nodeOverSubscriptionEnabled and cpu throttle feature enabled, then cpu throttle can be used

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
```diff
@@ -0,0 +1,156 @@
...
+		quotaFile = cgroup.CPUQuotaTotalFileV2
+		if quota == unlimitedQuota {
+			quotaValue = "max"
+		} else {
+			quotaValue = fmt.Sprintf("%d %d", quota, CPUPeriod)
+		}
+	}
+
+	filePath := path.Join(cgroupPath, quotaFile)
+	if err := utils.UpdatePodCgroup(filePath, []byte(quotaValue)); err != nil {
```
JesseStutler, line 136: To avoid frequent fluctuations, we can also add a minimum adjustment percent judgement here. If the percent change between the new quota and the old quota is less than this min percent, we will not make any adjustments.

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if !isActive {
+		klog.InfoS("CPU throttle feature disabled, recovering all throttled pods")
+		return h.recoverAllThrottledPods()
+	}
+
+	return nil
+}
+
+func (h *CPUThrottleHandler) recoverAllThrottledPods() error {
```
JesseStutler: From the current implementation, I don't think we need to this `recoverAllThrottledPods` func, for recover process, we can unify in the writeBEQuota, but we should add recover percent restrictions during the recovery phase(if new quota - old quota > recover percent, we should adjust to only recover percent)

JesseStutler (commented):
@Haibara-Ai97 Thanks for your great contribution, I think we're approaching what we want, just one more step we can finish this :)

JesseStutler:
BTW, currently there are too many commits in the pr, don't forget to clean your commits into 1-3 clean commits once you finish all the work @Haibara-Ai97

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
JesseStutler: @Haibara-Ai97 Please don't forget to sort the dependencies also in test file

pkg/agent/events/handlers/cputhrottle/cpu_throttling.go
JesseStutler: minimum percent we can set like 1%

JesseStutler:
Also I test in my local kind cluster I found such an error: 
image
Please check @Haibara-Ai97

JesseStutler:
> Also I test in my local kind cluster I found such an error: image Please check @Haibara-Ai97

I have known why this error will raise, because currently probe have only one queue: https://github.com/Haibara-Ai97/volcano/blob/8e99a3570e12670914a280db30e58dba0bab4445/pkg/agent/events/probes/nodemonitor/node_monitor.go#L77, but you add two type of probe in node monitor, so maybe you have to refactor the NewEventProbeFunc from queue to eventQueueFactory to get the queue according to the event type: https://github.com/Haibara-Ai97/volcano/blob/8e99a3570e12670914a280db30e58dba0bab4445/pkg/agent/events/probes/registry.go#L32

JesseStutler:
@Haibara-Ai97 Also notice that you didn't import the cputhrottle in eventsmgr.go, so it would be initialized, please don't forget to test your functionality

pkg/agent/events/probes/nodemonitor/node_monitor.go
Haibara-Ai97: I understand we need a condition to determine whether to perform CPU throttle. If we dont do this condition, then every time we pass the BE pods' available CPU quota to handler.
I dont know if there is something wrong with my understand.

JesseStutler: Yes you're right, indeed we should not pass the BE pods' available CPU quota to handler everytime, especially there no much resources changed(like 0.5%), but I feel that 10% of allowedMilli is a bit too arbitrary and hard-coded. https://github.com/volcano-sh/volcano/pull/4541#discussion_r2680999529 It could be done like the comment here: if the change between the new quota and the old quota is less than a configurable percentage (for example, 1%), then we don't send it to the handler (or it can be checked in the handler, if less than the percent, then nothing will be processed, but the probe still sends an event each time). @Haibara-Ai97

Haibara-Ai97: I kind of understand what you mean.
That is, we limit BE pods CPU quota in any time , no matter the current CPU load.
And if CPU usage doesnt change significantly, the cgroup modified will not be executed.

pkg/agent/events/probes/nodemonitor/node_monitor.go
Haibara-Ai97: DONE

pkg/agent/events/probes/nodemonitor/node_monitor.go
JesseStutler: Exactly, even the recovery process we still can do this way, but I hope in recovering, we can limit to a gradual 10%（configurable value) recovery, rather than affecting online workloads all at once.

Haibara-Ai97:
This branch spans a rather long period, and commits are a little bit confusing. I would like to create a new PR to commit the integrated code. @JesseStutler 
https://github.com/volcano-sh/volcano/pull/4945

PR #4547, opened 2025-08-15
feat: add extra args to scheduler deployment args

junzebao:
#### What type of PR is this?
helm chart update

This PR allows users to specify extra arguments to scheduler, for example `--default-queue`

```release-note
add scheduler_extra_args to the custom section in the helm chart
```


installer/helm/chart/volcano/templates/scheduler.yaml
```diff
@@ -230,6 +230,9 @@ spec:
             - --ignored-provisioners={{ .Values.custom.ignored_provisioners}}
             {{- end}}
             - -v={{.Values.custom.scheduler_log_level}}
+            {{- range .Values.custom.scheduler_extra_args }}
```
JesseStutler, line 233: Do we have `scheduler_extra_args` in values.yaml? I don't find that we have such an argument

PR #4552, opened 2025-08-18
fix high-priority jobs reclaiming resources and excessively killing l…

dafu-wu:
#### What type of PR is this?
/kind bug  
/kind fix  

This PR fixes an issue where **high-priority jobs reclaim resources from low-priority jobs** but cause **excessive killing of pods**.  
The change ensures that resource reclamation is performed more precisely, avoiding unnecessary pod evictions and improving overall job scheduling stability.  

Fixes #<issue number>  

- Verify the correctness of task state transitions after reclaim.  
- Ensure compatibility with existing predicate and allocate actions.  
- Additional testing on mixed-priority workloads is recommended.  

```release-note
Fix high-priority jobs reclaiming resources and excessively killing low-priority jobs.
Improved reclaim logic to prevent overkill of low-priority jobs.

### DIFF
--- a/docs/design/session-persisted-state.md
+++ b/docs/design/session-persisted-state.md
@@ -1,0 +1,138 @@
+# Volcano Scheduler: Task State Persistence Across Sessions 
+
+
+## Background
+
+Under heavy preemption/reclaim, high-priority jobs could reclaim resources too aggressively, leading to excessive evictions (“overkill”) of low-priority jobs and instability. Pipelined tasks (pre-allocated but not bound) lost state across sessions, causing duplicate decisions, mis-accounting, and oscillations. Plugins could observe inconsistent resource accounting during rapid pipeline → allocate → unallocate transitions.
+
+## Goals
+
+- Preserve and reconcile pipelined state across scheduler sessions.
+- Reduce unnecessary evictions and duplicate work.
+- Make allocation deterministic for pipelined tasks.
+- Prevent negative/invalid plugin accounting under churn.
+- Improve resiliency against concurrent state changes in the Kubernetes API.
+
+## Non-Goals
+
+- Redesign Volcano’s event model or plugin API.
+- Introduce new CRDs or controllers.
+- Change gang scheduling semantics.
+
+---
+
+## High-Level Design
+
+### Persist pipelined state on Pods (annotations)
+
+```go
+// pkg/scheduler/api/helpers.go
+const (
+  VolcanoPipelinedStatusAnnotation   = "volcano.sh/pipelined-status"
+  VolcanoPipelinedNodeAnnotation     = "volcano.sh/pipelined-node"
+  VolcanoEvictionOccurredAnnotation  = "volcano.sh/eviction-occurred"
+)
+```
+
+- On pipeline: set `pipelined-status=true` and `pipelined-node=<name>`.
+- On allocate (final bind): clear pipeline annotations.
+- On unallocate originating from a pipelined state: restore `pipelined-*` instead of resetting to `Pending`.
+- During reclaim: record `eviction-occurred=true` when victims are chosen.
+
+### Deterministic allocation for pipelined tasks
+
+- If a task is already pipelined to a node and the node becomes available, allocate via a fast path that avoids duplicate node ops and stale pipeline state.
+- Predicate helper recognizes pipelined tasks to the same node and short-circuits checks to prevent redundant work.
+
+### Harden plugins and resource comparisons
+
+- DRF, Capacity, Proportion plugins guard against underflow/negative deallocations and improve logs.
+- Resource comparisons (`LessEqualWithDimension`) ignore dimensions not configured for a queue (e.g., GPU scalar), avoiding spurious failures.
+
+### Safer cache/binder and framework updates
+
+- Binder: treat “already bound to same node” as success; if bound elsewhere, fail and clear annotations.
+- Annotation updates via a single abstraction to avoid GET/UPDATE drift; prefer PATCH.
+- Session/Statement ensure plugin callbacks are invoked exactly once on final allocation.
+
+---
+
+## Detailed Changes and Rationale
+
+### Actions
+
+**Allocate**
+- Add fast path to bind already-pipelined tasks once resources are free on the pipelined node.
+- Filter nil nodes, add defensive checks, and prevent duplicate AddTask on nodes for pipelined → final allocation.
+- Predicate: if task is pipelined to the node, skip full resource check; if the task moved, release stale pipelined accounting.
+
+**Reclaim**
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: pkg/scheduler/api/resource_info.go [hunk 1]
```diff
@@ -448,10 +448,10 @@ func (r *Resource) LessEqualWithDimension(rr *Resource, req *Resource) bool {
...
 	}
 
-	if req.MilliCPU > 0 && r.MilliCPU > rr.MilliCPU {
+	if req.MilliCPU > 0 && rr.MilliCPU != 0 && r.MilliCPU > rr.MilliCPU {
 		return false
 	}
-	if req.Memory > 0 && r.Memory > rr.Memory {
+	if req.Memory > 0 && rr.Memory != 0 && r.Memory > rr.Memory {
 		return false
 	}
```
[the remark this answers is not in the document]
dafu-wu: For those that don't set MilliCPU, Memory‘s queues, we should ignore

dafu-wu:
@JesseStutler can you help review it？

JesseStutler:
@dafu-wu Some CIs failed, please check

JesseStutler:
This PR has introduced several changes to the scheduling framework. Could you provide a design document to help understand it?

Monokaix:
Even pipeline state can not cross session, but every session when a pod is evicted, the pod will be terminating and node's future idle resources will include these resources, so why there will be overkilling?

dafu-wu:
> Even pipeline state can not cross session, but every session when a pod is evicted, the pod will be terminating and node's future idle resources will include these resources, so why there will be overkilling?

Because during the high priority pod pending period, each session high priority job will be considered starving, and will kill the low priority job again

dafu-wu:
> Even pipeline state can not cross session, but every session when a pod is evicted, the pod will be terminating and node's future idle resources will include these resources, so why there will be overkilling?

@Monokaix Can you also review and resolve existing test logic errors?

docs/design/session-persisted-state.md
```diff
@@ -0,0 +1,138 @@
+# Volcano Scheduler: Task State Persistence Across Sessions
```
JesseStutler, line 1: If this PR is only related to pipelined, I think the title might be too broad, may confuse users

docs/design/session-persisted-state.md
```diff
@@ -0,0 +1,138 @@
+# Volcano Scheduler: Task State Persistence Across Sessions 
+
+
+## Background
+
+Under heavy preemption/reclaim, high-priority jobs could reclaim resources too aggressively, leading to excessive evictions (“overkill”) of low-priority jobs and instability. Pipelined tasks (pre-allocated but not bound) lost state across sessions, causing duplicate decisions, mis-accounting, and oscillations. Plugins could observe inconsistent resource accounting during rapid pipeline → allocate → unallocate transitions.
```
JesseStutler, line 6: The explanation is too abstract.  Need to explain in detail why the overkill phenomenon occurs.

pkg/scheduler/framework/statement.go
```diff
@@ -216,6 +219,16 @@ func (s *Statement) Pipeline(task *api.TaskInfo, hostname string, evictionOccurr
 }
 
 func (s *Statement) pipeline(task *api.TaskInfo) {
+	// Set pipelined annotations on pod for cross-session state transfer
+	if task.Pod != nil {
```
JesseStutler, line 223: The operations in the statement are serial, and the number of pods is typically large. Sending update requests to the API server and waiting for responses would significantly slow down batch scheduling. If updating pod annotations is necessary, we should use methods such as `workqueue.ParallelizeUntil` to perform concurrent updates during the `closeSession`.

We can perform special processing for `pipelined` pods in the `RecordJobStatusEvent`: https://github.com/volcano-sh/volcano/blob/0373adc114720537f3ef90dab632029bd4fbcf60/pkg/scheduler/cache/cache.go#L1535-L1561

pkg/scheduler/api/helpers.go
```diff
@@ -64,6 +80,38 @@ func getTaskStatus(pod *v1.Pod) TaskStatus {
 	return Unknown
 }
 
+// SetPipelinedAnnotations sets Volcano pipelined state annotations on pod
+func SetPipelinedAnnotations(pod *v1.Pod, nodeName string, evictionOccurred bool) {
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+	pod.Annotations[VolcanoPipelinedStatusAnnotation] = "true"
+	pod.Annotations[VolcanoPipelinedNodeAnnotation] = nodeName
```
JesseStutler, line 89: Why don't we just directly use `pod.Status.NominatedNodeName` if preemption is successful?

pkg/scheduler/api/helpers.go
```diff
@@ -64,6 +80,38 @@ func getTaskStatus(pod *v1.Pod) TaskStatus {
...
+	if evictionOccurred {
+		pod.Annotations[VolcanoEvictionOccurredAnnotation] = "true"
```
JesseStutler, line 91: Is it necessary to record this evictionOccured?

pkg/scheduler/api/job_info.go
```diff
@@ -189,6 +189,18 @@ func NewTaskInfo(pod *v1.Pod) *TaskInfo {
 	// initialize pod scheduling gates info here since it will not change in a scheduling cycle
 	schGated := calSchedulingGated(pod)
 	jobID := getJobID(pod)
+	status := getTaskStatus(pod)
+	nodeName := pod.Spec.NodeName
+	evictionOccurred := false
+
+	// If task is in Pipelined state, restore context from annotations
+	if status == Pipelined {
+		pipelinedNode, pipelinedEviction := getPipelinedContext(pod)
```
JesseStutler, line 198: `TransactionContext` is only used in one session to temporarily store some information. We cannot carry it across sessions. It feels strange to do special judgment for Pipelined here.

pkg/scheduler/cache/cache.go
```diff
@@ -208,6 +210,39 @@ func (db *DefaultBinder) Bind(kubeClient kubernetes.Interface, tasks []*scheduli
 	errMsg := make(map[schedulingapi.TaskID]string)
 	for _, task := range tasks {
 		p := task.Pod
+		// Check if pod is already bound to avoid conflicts
```
JesseStutler, line 213: What do these codes for?

JesseStutler:
😵 Perhaps we should clarify the reason why you encountered overkill. Even if the pipeline does need to be passed across sessions, it should be sufficient to update the pod annotation and do special processing in the `getTaskStatus`. Why are there so many other additional changes?

dafu-wu:
Resolved #4563

PR #4558, opened 2025-08-20
fix panic when capacity plugins real capacity is nil  when  clone happen

rockburning:
fix panic when clone when real capacity is nil 
img_v3_02pa_fd71a9dc-22d5-411a-8702-e8310637f51g

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -714,13 +714,14 @@ func (cp *capacityPlugin) newQueueAttr(queue *api.QueueInfo) *queueAttr {
 		ancestors: make([]api.QueueID, 0),
 		children:  make(map[api.QueueID]*queueAttr),
 
-		deserved:   api.NewResource(queue.Queue.Spec.Deserved),
-		allocated:  api.EmptyResource(),
-		request:    api.EmptyResource(),
-		elastic:    api.EmptyResource(),
-		inqueue:    api.EmptyResource(),
-		guarantee:  api.EmptyResource(),
-		capability: api.EmptyResource(),
+		deserved:       api.NewResource(queue.Queue.Spec.Deserved),
+		allocated:      api.EmptyResource(),
+		request:        api.EmptyResource(),
+		elastic:        api.EmptyResource(),
+		inqueue:        api.EmptyResource(),
+		guarantee:      api.EmptyResource(),
+		capability:     api.EmptyResource(),
+		realCapability: api.EmptyResource(),
 	}
 	if len(queue.Queue.Spec.Capability) != 0 {
 		attr.capability = api.NewResource(queue.Queue.Spec.Capability)


JesseStutler:
What's the root cause? Can you provide more details on the issue background? Why does initializing a realCapability solve it?

rockburning:
> What's the root cause? Can you provide more details on the issue background? Why does initializing a realCapability solve it?

you can see from the picture,it will clone panic when realCapability is nil,it will access the member of the struct. type Resource struct {
	MilliCPU float64
	Memory   float64

	// ScalarResources
	ScalarResources map[v1.ResourceName]float64

	// MaxTaskNum is only used by predicates; it should NOT
	// be accounted in other operators, e.g. Add.
	MaxTaskNum int
},to avoid this, so i initialize realCapability  with empty resource

JesseStutler:
We need to confirm the triggering scenario more precisely, is it a hierarchical queue or a non-hierarchical queue scenario? Regardless of whether the hierarchical queue is enabled or not, realCapability is now assigned a value instead of nil, so we need to know the actual root cause

rockburning:
it is a hierarchical queue ，

---- Replied Message ----
| From | Jesse ***@***.***> |
| Date | 08/27/2025 17:07 |
| To | volcano-sh/volcano ***@***.***> |
| Cc | BurningRock ***@***.***>,
Mention ***@***.***> |
| Subject | Re: [volcano-sh/volcano] fix panic when capacity plugins real capacity is nil when clone happen (PR #4558) |
JesseStutler left a comment (volcano-sh/volcano#4558)

We need to confirm the triggering scenario more precisely, is it a hierarchical queue or a non-hierarchical queue scenario? Regardless of whether the hierarchical queue is enabled or not, realCapability is now assigned a value instead of nil, so we need to know the actual root cause

JesseStutler:
Strange thing is that: https://github.com/volcano-sh/volcano/blob/b27b2a96e0a73a4bd0f64f770cd022863a6d4b4e/pkg/scheduler/plugins/capacity/capacity.go#L809-L815 If we opened hierarchical queues, all queues' realCapability can be successfully initialized, so why still cause panic? 

Which version of volcano did you use?

rockburning:
i use the master branch， i will check it when i have free time

---- Replied Message ----
| From | Jesse ***@***.***> |
| Date | 08/28/2025 15:28 |
| To | volcano-sh/volcano ***@***.***> |
| Cc | BurningRock ***@***.***>,
Mention ***@***.***> |
| Subject | Re: [volcano-sh/volcano] fix panic when capacity plugins real capacity is nil when clone happen (PR #4558) |
JesseStutler left a comment (volcano-sh/volcano#4558)

Strange thing is that: https://github.com/volcano-sh/volcano/blob/b27b2a96e0a73a4bd0f64f770cd022863a6d4b4e/pkg/scheduler/plugins/capacity/capacity.go#L809-L815 If we opened hierarchical queues, all queues' realCapability can be successfully initialized, so why still cause panic?

Which version of volcano did you use?

hajnalmt:
This has the DCO not signed.
Let's close it as a duplicate of [https://github.com/volcano-sh/volcano/pull/4611](https://github.com/volcano-sh/volcano/pull/4611) and merge the other one I think.

hajnalmt:
/close
In favor of https://github.com/volcano-sh/volcano/pull/4611
It has been cherry-picked to every branch.

PR #4562, opened 2025-08-21
Optimize queueAllocatable calculation by excluding allocated resources from unschedulable nodes

dafu-wu:
#### What type of PR is this?
/kind bug
/kind cleanup

This PR optimizes the calculation of `queueAllocatable` by excluding allocated resources that come from nodes marked as unschedulable.  
Without this fix, queue-level resource accounting may include unusable allocations, leading to inaccurate scheduling decisions and potential over-commit.

Fixes #

- The change ensures that only allocatable resources from schedulable nodes are considered.  
- Helps improve fairness and accuracy in queue-level resource management.  
- No functional impact expected on already scheduled workloads, only more accurate accounting.

```release-note
Optimize queueAllocatable calculation by excluding allocated resources from unschedulable nodes.
This prevents incorrect resource accounting and improves scheduling accuracy.

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -32,6 +32,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/plugins/util"
+	schedutil "volcano.sh/volcano/pkg/scheduler/util"
 )
 
 const (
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by dafu-wu** [comment]
@JesseStutler @Monokaix @shinytang6 @alcorj-mizar  can you help to review this PR?

File: pkg/scheduler/plugins/capacity/capacity.go [hunk 1]
```diff
@@ -61,9 +62,10 @@ type queueAttr struct {
 	ancestors []api.QueueID
 	children  map[api.QueueID]*queueAttr
 
-	deserved  *api.Resource
-	allocated *api.Resource
-	request   *api.Resource
+	deserved                               *api.Resource
+	allocated                              *api.Resource
+	allocatedWithoutSchedulingDisabledNode *api.Resource
```
JesseStutler: Why should we add a new field rather than just directly modify `allocated`?

JesseStutler (commented):
I think we may also have same problems in proportion plugin

pkg/scheduler/plugins/capacity/capacity.go
dafu-wu: Updated, reused allocated

dafu-wu:
> I think we may also have same problems in proportion plugin

@JesseStutler can you help to modify it? I didn't see any relevant code

dafu-wu:
@JesseStutler @caniszczyk can you merge this pr or fix this issue asap? It's block us🙏

Monokaix:
We can revert this one https://github.com/volcano-sh/volcano/pull/4373

dafu-wu:
Resolved #4575

PR #4565, opened 2025-08-25
Memory QoS for CgroupV2

### LINKED ISSUE #4466: [Colocation Enhancement] Support memory qos

Currently, volcano doesn't support memory QoS setting, in cgroup v2, memcg is enhanced to allow cgroup to configure settings such as `memory.high`, `memory.low`, `memory.min`:
https://docs.kernel.org/admin-guide/cgroup-v2.html#memory
- `memory.high` can help kernel to throttle the memory usage of processes in the cgroup, asynchronously and gently reclaim memory, and avoid OOM when the process continues to grow and reaches the memory limit
- `memory.low`: It will reduce the priority of memory used by processes in the cgroup to be reclaimed, which will help improve application performance, especially when the node is under pressure.
- `memory.min`: Hard memory protection. This amount of memory would be unreclaimable

Kubelet supports cgroup v2, but the configuration will take effect on all containers on the node and lacks fine-grained control at the pod level:
https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#memory-qos-with-cgroup-v2, 
https://kubernetes.io/blog/2023/05/05/qos-memory-resources/

Therefore, we need to support memory QoS setting, we can support cgroup v2, some OS even merge the memory QoS feature back to cgroup v1, we need to support it too

We need to support memory QoS setting, we can support cgroup v2, some OS even merge the memory QoS feature back to cgroup v1, we need to support it too

### Additional context

_No response_

murali1539:
Fixes #4466

```release-note

```

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -28,7 +28,8 @@ import (
 const (
 	serverPort = 3300
 
-	defaultOverSubscriptionRatio = 60
+	defaultOverSubscriptionRatio  = 60
+	defaultMemoryThrottlingFactor = 0.9
 )
 
 type VolcanoAgentOptions struct {
@@ -68,6 +69,9 @@ type VolcanoAgentOptions struct {
 
 	// ExtendResourceMemoryName is the extend resource memory, which is used to calculate overSubscription resources.
 	ExtendResourceMemoryName string
+
+	// MemoryThrottlingFactor is used to calculate the throttling limit for a Pod for CGroupV2.
+	MemoryThrottlingFactor float64
 }
 
 func NewVolcanoAgentOptions() *VolcanoAgentOptions {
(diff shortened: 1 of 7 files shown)


murali1539:
@Monokaix I have created this PR for issue : [4466](https://github.com/volcano-sh/volcano/issues/4466). 
I have noticed that in the current code, the value in file "memory.limit_in_bytes"  is only set for extended-memory and not for the limit configured on the Pod. May I know the reason behind this ? 
Based on your answer I would have to set the memory-qos values for CGroupV2.

Monokaix:
> @Monokaix I have created this PR for issue : [4466](https://github.com/volcano-sh/volcano/issues/4466). I have noticed that in the current code, the value in file "memory.limit_in_bytes" is only set for extended-memory and not for the limit configured on the Pod. May I know the reason behind this ? Based on your answer I would have to set the memory-qos values for CGroupV2.

Because other pods will be set by kubelet by default when container started.

PR #4566, opened 2025-08-25
Set root queue capability through helm

### LINKED ISSUE #4496: Support set root capability deserved and guarantee when volcano install by helm
Now root queue is auto created, but not support set capability deserved and guarantee when created, so that I still met error although bugfix in pr https://github.com/volcano-sh/volcano/pull/4354.
[issue](https://github.com/volcano-sh/volcano/issues/4350)

hajnalmt:
When the hierarchy plugin is enabled, the capability check can fail:
https://github.com/volcano-sh/volcano/issues/4350
It will pass now, if the capability is set properly: https://github.com/volcano-sh/volcano/pull/4354
We should be able to set this value for the root queue through helm too.

#### What type of PR is this?
The PR adds a helm value that will render the rootQueue capability if provided.

Minor change in the helm chart.

Fixes https://github.com/volcano-sh/volcano/issues/4496

```release-note
New helm value for root queue capabilty settings.
```

### DIFF
--- a/installer/helm/chart/volcano/templates/root_queue.yaml
+++ b/installer/helm/chart/volcano/templates/root_queue.yaml
@@ -1,0 +1,9 @@
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: root
+spec:
+  capability:
+    {{- toYaml .Values.custom.root_queue.capability | nindent 4 }}
+  reclaimable: {{ .Values.custom.root_queue.reclaimable | default false }}
+  weight: {{ .Values.custom.root_queue.weight | default 1 }}
--- a/installer/helm/chart/volcano/values.yaml
+++ b/installer/helm/chart/volcano/values.yaml
@@ -224,6 +224,12 @@ custom:
       drop: [ "ALL" ]
     allowPrivilegeEscalation: false
 
+  # Specify root Queue capability
+  root_queue:
+    capability: ~
+    reclaimable: false
+    weight: 1
+
   # Specify agent cni config path.
   agent_cni_config_path: /etc/cni/net.d/cni.conflist
 


installer/helm/chart/volcano/values.yaml
```diff
@@ -224,6 +224,10 @@ custom:
       drop: [ "ALL" ]
     allowPrivilegeEscalation: false
 
+  # Specify rootQueue capability
+  rootQueue:
+    capability: ~
```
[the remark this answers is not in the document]
hajnalmt: okay, done!

installer/helm/chart/volcano/templates/root_queue.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.rootQueue.capability }}
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: root
+spec:
+  capability:
+    {{- toYaml .Values.rootQueue.capability | nindent 4 }}
+{{- end }}
```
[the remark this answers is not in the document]
hajnalmt: Updated. Thank you!

installer/helm/chart/volcano/templates/root_queue.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.custom.root_queue.capability }}
```
JesseStutler: We can install root queue by default, because that's what we do now

installer/helm/chart/volcano/templates/root_queue.yaml
```diff
@@ -0,0 +1,9 @@
...
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: root
+spec:
+  capability:
```
JesseStutler, line 6: We may also allow user to specify `deserved` and `guarantee.resource` when in installation

JesseStutler (commented):
Also need to run `make update-development-yaml` to render helm template to `volcano-development.yaml`

JesseStutler:
https://github.com/volcano-sh/volcano/blob/bc0df8e6e41fa2ade130050fd3b1aee0b31f960d/pkg/scheduler/cache/cache.go#L439-L491

Besides, currently we install the root queue and the default queue after scheduler starts, if we need to move default queue and root queue installation to helm, we may consider deleting these codes

installer/helm/chart/volcano/templates/root_queue.yaml
hajnalmt: At first I added this too, but we update these at the session start:
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/framework/session.go#L307
Does it make sense at all to add this to helm if it will be changed later, and be updated at every session start? I feel some possible confusion here, since you set the value from helm, but it's changed in the end.

installer/helm/chart/volcano/templates/root_queue.yaml
hajnalmt: Okay, I removed the check and also added the reclaimable and weight keys, to align with the original defaults.

hajnalmt:
The `make update-development-yaml` shall be ran before the next release if this gets in right?
I wouldn't remove the cache.go related code yet, but we should consider it later on definitely.

installer/helm/chart/volcano/templates/root_queue.yaml
JesseStutler: OK, ignore me. I think set the capability when installation is enough

JesseStutler:
> The make update-development-yaml shall be ran before the next release if this gets in right?

Yes some users may just directly use `volcano-development.yaml` by using kubectl to deploy volcano, so we need to make sure that the rendering result of helm is consistent with the `volcano-development.yaml`

JesseStutler:
There are a lot of CIs failed @hajnalmt , is that because the conflict with current codes in cache.go?

hajnalmt:
Sighs, it's a different issue... The problem is that the queue CRD does not exist when the chart get's installed.
Helm suggests to put the CRDs in a crds folder, and it won't touch them after they get installed. https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
This will ensure that CRDs are installed first, but this makes the CRDs not updateable, which makes it horrific to deal with.
There is a community discussion about this: https://github.com/helm/community/pull/379 so they can be updateable.

The practice currently is still the templating:
Contour: https://github.com/bitnami/charts/tree/9dde49c88c7a45d7fa36df5b81c285a627bad896/bitnami/contour/templates/crds Cert-manager: https://github.com/bitnami/charts/tree/9dde49c88c7a45d7fa36df5b81c285a627bad896/bitnami/cert-manager/templates/crds
But this means I can't just use it and create the root queue from helm a template. Shall I try to implement this via a helm hook? Or shall we wait for helm4 where they will be probably solving this?

JesseStutler:
> Sighs, it's a different issue... The problem is that the queue CRD does not exist when the chart get's installed. Helm suggests to put the CRDs in a crds folder, and it won't touch them after they get installed. https://helm.sh/docs/chart_best_practices/custom_resource_definitions/ This will ensure that CRDs are installed first, but this makes the CRDs not updateable, which makes it horrific to deal with. There is a community discussion about this: [helm/community#379](https://github.com/helm/community/pull/379) so they can be updateable.
> 
> The practice currently is still the templating: Contour: https://github.com/bitnami/charts/tree/9dde49c88c7a45d7fa36df5b81c285a627bad896/bitnami/contour/templates/crds Cert-manager: https://github.com/bitnami/charts/tree/9dde49c88c7a45d7fa36df5b81c285a627bad896/bitnami/cert-manager/templates/crds But this means I can't just use it and create the root queue from helm a template. Shall I try to implement this via a helm hook? Or shall we wait for helm4 where they will be probably solving this?

Our crds are also defined in a separate folder. Doesn't it mean that CRDs will be installed first? Then why is it still reporting an error that queue crd does not exist? We didn't update crd, but just installed a cr, right?

hajnalmt:
> Our crds are also defined in a separate folder. Doesn't it mean that CRDs will be installed first? Then why is it still reporting an error that queue crd does not exist? We didn't update crd, but just installed a cr, right?

Our folder name is **crd** and not **crds**, and they are just referenced from the templates. For helm to install the CRDs first, they should be in a crds directory... After that you can use the helm CLI parameters to skip their installation (--skip-crds) the problem is that this makes the CRDs not upgradable, see the caveats section here:
[https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you
](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you
) so I think we shouldn't do this.

hajnalmt:
@JesseStutler Shall I close this until Helm has a better way to install the CRDs?

houyuting:
@hajnalmt Hi, I have a idea that might solve your problem,  The core idea is：
**step 1:** helm chart support  Inject the capacity values ​​passed during `helm install` into the `vc-scheduler` Pod via environment variables or ConfigMap.
**step 2:**  volcano Scheduler update `pkg/scheduler/cache/cache.go` so that when creating the `root` queue, it prioritizes reading the capacity values ​​in these environment variables or ConfigMaps, and uses the default values ​​if they do not exist.

@JesseStutler Please take a look together when you have time.

hajnalmt:
Closing in favor of: https://github.com/volcano-sh/volcano/pull/4566

PR #4583, opened 2025-09-01
Support configuring network-topology via pod annotations

zhifei92:
#### What type of PR is this?
/kind feature
This enhancement allows Pods created by Deployments and StatefulSets to benefit from Volcano's advanced network topology-aware scheduling capabilities.

Fixes #

````release-note
Added NetworkTopology support for PodGroups created from Pod annotations.

Added the ability to configure NetworkTopology through Pod annotations when PodGroups are automatically created from Pods. This extends Volcano's network topology-aware scheduling to Deployment, StatefulSet, and other standard Kubernetes workloads.

**Usage:**
```yaml
metadata:
  annotations:
    volcano.sh/network-topology-mode: "hard"           # or "soft"
    volcano.sh/network-topology-highest-tier: "2"
```
````

### CODE REVIEW

**Comment by JesseStutler** [comment]
This is exactly what we needed, but I think we should define the annotation key in apis repo first?

/cc
Also /cc @Monokaix

**Comment by zhifei92** [comment]
> This is exactly what we needed, but I think we should define the annotation key in apis repo first?

Good point! I'll define the annotation key in the apis repo first and update this PR to use it. Thanks!

https://github.com/volcano-sh/apis/pull/193

**Comment by zhifei92** [comment]
@JesseStutler @kingeasternsun @Monokaix please take a look, thanks.

File: pkg/controllers/podgroup/pg_controller_handler.go [hunk 1]
```diff
@@ -413,9 +414,62 @@ func (pg *pgcontroller) buildPodGroupFromPod(pod *v1.Pod, pgName string) *schedu
 		obj.Annotations[scheduling.JDBMaxUnavailable] = value
 	}
 
+	// Parse and set NetworkTopology from Pod annotations
+	if networkTopology := parseNetworkTopologyFromPod(pod); networkTopology != nil {
+		obj.Spec.NetworkTopology = networkTopology
+
+		highestTierStr := "nil"
```
Monokaix: The highest tier is a int value, it's not reasonable to set a string "nil" here.
The highestTier will be set default value to 1, so we just give a notice is ok, no need to add a "nil" string, see: https://github.com/volcano-sh/apis/blob/1aeba4dd588fdf9dd3b34394736a0cad485bd578/pkg/apis/scheduling/v1beta1/types.go#L228

zhifei92: done

zhifei92:
@JesseStutler @kingeasternsun @Monokaix please take a look, thanks.

kingeasternsun:
please rebase this PR because the CI issue has been fixed  @zhifei92 .

PR #4587, opened 2025-09-01
Minor docs changes 

hajnalmt:
No actual code changes in this PR.

I am fixing a view development docs links since https://github.com/golang/go/wiki is not available anymore.
Adding e2e log directory and ASDF tools file to gitignore.
Fix typo in capacity_test.go.

#### What type of PR is this?
Mostly doc link fixes.
Also fixing a typo in the capacity_test.go.

Additionally I want to add the .tool-versions file and the volcano-e2e-logs directory to .gitignore.
I am using ASDF to manage my dependencies like kubectl and this can ease the development for people like me.
The volcano-e2e-logs directory is not escaped in the gitignore, so after I run an e2e test locally git will try to add the directory to the commit always (without an empty content because the *.log files are escaped.)

It contains minor improvements/fixes.

If it's possible I wouldn't create a separate issue for any of this.

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,9 @@ network_closure.sh
 /cluster/juju/charms/*
 /cluster/juju/bundles/local.yaml
 
+# ASDF version file
+.tool-versions
+
 # Downloaded Kubernetes binary release
 /kubernetes/
 
@@ -124,6 +127,7 @@ kubernetes.tar.gz
 *.pyc
 
 # e2e log files
+volcano-e2e-logs/
 *.log
 
 # test coverage file
--- a/pkg/scheduler/plugins/capacity/capacity_test.go
+++ b/pkg/scheduler/plugins/capacity/capacity_test.go
@@ -450,9 +450,9 @@ func Test_capacityPlugin_OnSessionOpenWithHierarchy(t *testing.T) {
 	}
 
 	// resources for test case 8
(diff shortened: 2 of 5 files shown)


pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -450,9 +450,9 @@ func Test_capacityPlugin_OnSessionOpenWithHierarchy(t *testing.T) {
 	}
 
 	// resources for test case 8
-	queue8 := buildQueueWithParents("q8", "root", api.BuildResourceList("2", "2Gi", []api.ScalarResource{{Name: "nvidia.com/gpu", Value: "8"}}...), nil)
+	queue8 := buildQueueWithParents("q8", "root", api.BuildResourceList("4", "4Gi", []api.ScalarResource{{Name: "nvidia.com/gpu", Value: "8"}}...), nil)
```
Monokaix, line 453: Why change to 4？

Monokaix:
/ok-to-test
Thanks!

pkg/scheduler/plugins/capacity/capacity_test.go
hajnalmt: The hierarchy check won't pass otherwise and it's valid since the deserved part of the parent queue is less than the sum of the subqueue's deserved part.

.gitignore
```diff
@@ -98,6 +98,9 @@ network_closure.sh
 /cluster/juju/charms/*
 /cluster/juju/bundles/local.yaml
 
+# ASDF version file
+.tool-versions
```
Monokaix, line 102: When will we use this ASDF tool?

PR #4596, opened 2025-09-05
WIP: Add ValidatingAdmissionPolicy/MutatingAdmissionPolicy as optional enhanced admission

suyiiyii:
#### What type of PR is this?
enhancement

1. Add ValidatingAdmissionPolicy/MutatingAdmissionPolicy 
2. Add test for VAP/MAP

https://summer-ospp.ac.cn/org/prodetail/253ba0259

```release-note
Add optional ValidatingAdmissionPolicy/MutatingAdmissionPolicy 
```


.github/workflows/e2e_adminission.yaml
```diff
@@ -0,0 +1,88 @@
+name: E2E Adminssion
```
JesseStutler: typo `Adminssion` -> `Admission`

.github/workflows/e2e_dra.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       - name: Install dependences
         run: |
-          GO111MODULE="on" go install sigs.k8s.io/kind@v0.29.0
+          GO111MODULE="on" go install sigs.k8s.io/kind@v0.30.0
```
JesseStutler: What's the purpose of updating kind here? We can unify the kind version update in k8s 1.34 updating. Or do you have to update kind version here?

hack/e2e-kind-config.yaml
```diff
@@ -23,7 +24,8 @@ nodes:
         kind: ClusterConfiguration
         apiServer:
           extraArgs:
-            runtime-config: "resource.k8s.io/v1beta1=true"
+            runtime-config: "resource.k8s.io/v1beta1=true,admissionregistration.k8s.io/v1beta1"
+            enable-admission-plugins: "NodeRestriction,MutatingAdmissionPolicy"
```
JesseStutler: What's the purpose of enabled `NodeRestriction` admission plugin here?

hack/generate-yaml.sh
```diff
@@ -63,6 +65,24 @@ esac
 
 ARCH=$(go env GOARCH)
 
+# Set VAP enablement based on environment variable
+VAP_ENABLED="false"
```
JesseStutler: Why don't we just unify the `VAP_ENABLED` var and `ENABLE_VAP` var?

hack/generate-yaml.sh
```diff
@@ -63,6 +65,24 @@ esac
...
+if [[ "$ENABLE_VAP" == "true" ]]; then
+  VAP_ENABLED="true"
+  echo "ValidatingAdmissionPolicy enabled"
+else
+  echo "ValidatingAdmissionPolicy disabled"
+fi
+
+# Set MAP enablement based on environment variable
+MAP_ENABLED="false"
```
JesseStutler: Same as above

hack/run-e2e-kind.sh
```diff
@@ -136,8 +228,11 @@ custom:
   default_ns:
     node-role.kubernetes.io/control-plane: ""
   scheduler_feature_gates: ${FEATURE_GATES}
+  enabled_admissions: "/pods/mutate,/queues/mutate,/podgroups/mutate,/jobs/mutate,/jobs/validate,/jobflows/validate,/pods/validate,/queues/validate,/podgroups/validate,/hypernodes/validate"
```
JesseStutler: Do we have to enable admission when we run other e2e cases?

test/e2e/admission/hypernode_validation_test.go
```diff
@@ -0,0 +1,480 @@
...
+			},
+		}
+
+		_, err := testCtx.Vcclient.TopologyV1alpha1().HyperNodes().Create(context.TODO(), hyperNode, metav1.CreateOptions{})
+		gomega.Expect(err).To(gomega.HaveOccurred())
+		gomega.Expect(err.Error()).To(gomega.ContainSubstring("member regexMatch pattern is invalid"))
+	})
+
+	// validate by crd  e.g. member selector must have one of exactMatch, regexMatch, or labelMatch
+	ginkgo.XIt("Should reject HyperNode creation with multiple selector types (exactMatch and regexMatch)", func() {
```
JesseStutler: Why does it use `XIt` here rather than `It`? Does this case have problems?

test/e2e/admission/job_validation_test.go
```diff
@@ -0,0 +1,748 @@
...
+				},
+			},
+		}
+
+		_, err := testCtx.Vcclient.BatchV1alpha1().Jobs(testCtx.Namespace).Create(context.TODO(), job, metav1.CreateOptions{})
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	})
+
+	// VAP is unable to verify this situation
+	ginkgo.XIt("Should reject job creation with invalid task dependencies (non-DAG)", func() {
```
JesseStutler: Also this place use `XIt` instead? Please check

JesseStutler:
@suyiiyii Could you take a look of cronjob e2e CI cause it can not start the kind cluster?

.github/workflows/e2e_dra.yml
suyiiyii: MAP entered the Beta phase at 1.34. Directly using k8s 1.34+ with MAP Beta can reduce the workload of subsequent updates.
Of course, if necessary, I can also roll back MAP to the Alpha version; I will follow the arrangement.
By the way, the fact that MAP has entered the Beta phase does not involve any substantive changes.

hack/e2e-kind-config.yaml
suyiiyii: This is an unnecessary configuration, and I will delete it.

<details>

This configuration was copied from the cluster I set up using kubeadm. Kubeadm uses NodeRestriction as the default value for enable-admission-plugins.

</details>

hack/run-e2e-kind.sh
suyiiyii: The current situation is as follows:

  | Configuration Item | ADMISSION Test with POLICY(NEW) | ADMISSION Test with WEBHOOK(NEW) | Other Test(Original)   |
  |-------------------|------------------|-------------------|----------------|
  | VAP               | ✅ Enabled        | ❌ Disabled       | ❌ Disabled    |
  | MAP               | ✅ Enabled        | ❌ Disabled       | ❌ Disabled    |
  | Webhook           | ❌ Disabled       | ✅ Fully Enabled  | ✅ Fully Enabled|

For the completeness of testing, it is necessary to run all tests under the conditions where VAP/MAP are enabled and disabled respectively to verify whether VAP/MAP are compatible with other modules. However, this will double the test workload, and I am not sure if this is the right approach.

test/e2e/admission/hypernode_validation_test.go
suyiiyii: Yes, this test will fail because this rule is checked in the CRD layer, causing the assert to be error-free in the test.


https://github.com/volcano-sh/volcano/blob/master/pkg/webhoaks/admission/hypernodes/validate/admit_hypernode.go#L87-L112

My test was written based on webhook logic, so I wrote this use case, but I found the above problem in the follow-up, so I skipped it.

I will remove this test in the next commit.

test/e2e/admission/job_validation_test.go
suyiiyii: This is a test that VAP cannot verify.

For the same reason, I wrote this test, but this test could not pass, so I marked it skipped.

I will delete this test in the next commit.

installer/helm/chart/volcano/policy/queues-validating.yaml
```diff
@@ -0,0 +1,141 @@
...
+    # Extract hierarchy annotation if present
+    - name: hierarchyPath
+      expression: |
+        has(object.metadata) && has(object.metadata.annotations) && "volcano.sh/hierarchy" in object.metadata.annotations ? 
+        object.metadata.annotations["volcano.sh/hierarchy"] : ""
+    # Extract hierarchy weights annotation if present
+    - name: hierarchyWeights
+      expression: |
+        has(object.metadata) && has(object.metadata.annotations) && "volcano.sh/hierarchy.weight" in object.metadata.annotations ? 
+        object.metadata.annotations["volcano.sh/hierarchy.weight"] : ""
```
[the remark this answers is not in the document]
JesseStutler: @suyiiyii You can also check gemini's review opinions to check whether it is correct

.github/workflows/e2e_dra.yml
JesseStutler: I think we can keep current version in this pr, and after @tunedev finishing updating dependencies to v1.34, we can update the MAP to beta phase

hack/run-e2e-kind.sh
JesseStutler: I think we can keep it as it is, we can enable some admissions when needed in the future.

JesseStutler:
@suyiiyii Once you have fixed the e2e test and all of the comments, you can change this draft PR to a normal PR

Makefile
```diff
@@ -218,6 +224,16 @@ update-development-yaml:
...
 
+	ENABLE_VAP=true make generate-yaml RELEASE_DIR=installer
+	mv installer/volcano-${TAG}.yaml installer/volcano-development-vap.yaml
+	rm installer/volcano-agent-${TAG}.yaml
+	rm installer/volcano-monitoring-${TAG}.yaml
+
+	ENABLE_VAP=true ENABLE_MAP=true make generate-yaml RELEASE_DIR=installer
+	mv installer/volcano-${TAG}.yaml installer/volcano-development-vap-map.yaml
+	rm installer/volcano-agent-${TAG}.yaml
+	rm installer/volcano-monitoring-${TAG}.yaml
```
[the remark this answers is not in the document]
suyiiyii: The current version is already very simple and does not require a complex package.

installer/helm/chart/volcano/policy/hypernodes-validating.yaml
```diff
@@ -0,0 +1,137 @@
...
+          .all(member, member.selector.exactMatch.name != "")
+      message: "member exactMatch name is required"
+      reason: Invalid
+
+    # Validate exactMatch names follow Kubernetes qualified name rules using CEL format library
+    - expression: |
+        !has(object.spec) || !has(object.spec.members) || size(object.spec.members) == 0 ||
+        variables.exactMatchNames.all(name,
+          !format.qualifiedName().validate(name).hasValue()
+        )
```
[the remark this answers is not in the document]
suyiiyii: The logic is correct. validate() returns errors when invalid (has value) or none when valid (no value). So !hasValue() = "no errors" = valid. 

See the https://pkg.go.dev/k8s.io/apiserver/pkg/cel/library.

installer/helm/chart/volcano/policy/jobs-validating.yaml
```diff
@@ -0,0 +1,252 @@
...
+      reason: Invalid
+    # Validate job name is a valid qualified name using CEL format library
+    - expression: |
+        !has(object.metadata) || !has(object.metadata.name) ||
+        !format.qualifiedName().validate(object.metadata.name).hasValue()
+      message: "job name must be a valid qualified name"
+      messageExpression: |
+        has(object.metadata) && has(object.metadata.name) && format.qualifiedName().validate(object.metadata.name).hasValue() ?
+        'job name errors: ' + format.qualifiedName().validate(object.metadata.name).orValue([]).join('; ') : ''
+      reason: Invalid
```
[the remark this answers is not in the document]
suyiiyii: The logic is correct. validate() returns errors when invalid (has value) or none when valid (no value). So !hasValue() = "no errors" = valid. 

See the https://pkg.go.dev/k8s.io/apiserver/pkg/cel/library.

test/e2e/admission/job_validation_test.go
```diff
@@ -0,0 +1,598 @@
...
+		job.Spec.Policies = []v1alpha1.LifecyclePolicy{
+			{
+				Action: busv1alpha1.AbortJobAction,
+			},
+		}
+
+		_, err := testCtx.Vcclient.BatchV1alpha1().Jobs(testCtx.Namespace).Create(context.TODO(), job, metav1.CreateOptions{})
+		gomega.Expect(err).To(gomega.HaveOccurred())
+		gomega.Expect(err.Error()).To(gomega.ContainSubstring("either event and exitCode should be specified"))
+	})
```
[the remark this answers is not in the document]
suyiiyii: They're different:
- Test 1: Rejects when both Event and ExitCode are set
- Test 2: Rejects when neither Event nor ExitCode is set

Both are needed to ensure exactly one is specified.

suyiiyii:
@JesseStutler Thanks for your review. I believe I have addressed all the comments.

installer/helm/chart/volcano/templates/webhooks.yaml
JesseStutler, line 1: @suyiiyii Why has this file been changed? Seems that there are no differences between these two files

installer/volcano-development.yaml
JesseStutler, line 1: @suyiiyii Why has this file been changed? Seems that there are no differences between these two files. Is LF difference?

installer/helm/chart/volcano/templates/webhooks.yaml
suyiiyii: @JesseStutler Sorry, I forgot to mention that the issue here is caused by an invalid `if` template. This invalid template prevents the `validate webhook` of the `cronjob` from being controlled via parameters, leading to test failures.

The specific location is here:  https://github.com/volcano-sh/volcano/blob/f819d0eb95db276d7762655c072b2a7bb7f03f14/installer/helm/chart/volcano/templates/webhooks.yaml#L471-L472

I previously fixed this issue and also cleaned up the trailing newlines at the end of lines as a side task. Now, I have removed those unnecessary modifications and only retained the changes that are essential for my needs.

JesseStutler:
/approve
Thanks for your great contribution!

JesseStutler:
@hwdef @lowang-bh @hzxuzhonghu @Monokaix Could anyone take a look of this and add a lgtm label if agree with this PR?

hwdef:
/lgtm
Thanks for your contribution!

PR #4602, opened 2025-09-07
add nodegroup nonstrict to fit 0nodes queue

### LINKED ISSUE #4528: NodeGroup: when a queue does not specify affinity, load is allowed to be scheduled to any node

Currently, we have enabled the nodeGroup plugin and specified both queues with nodeGroup affinity and queues without affinity. Based on the implementation of the nodeGroup plugin, we will label nodes with corresponding nodeGroup tags. For some workloads, we need them to have node affinity, so we use queues with affinity. For others, we don't need it, so we use queues without affinity. Our expectation is that these workloads can be scheduled to any node, but in reality, they currently cannot be scheduled due to the Predicate operation of the nodeGroup.

workloads using queues without affinity can be scheduled

### Additional context

_No response_

wuyueandrew:

Fixes #4528 

```release-note

```


pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -67,11 +70,14 @@ func NewQueueGroupAffinity() queueGroupAffinity {
 	}
 }
 
-func (q queueGroupAffinity) predicate(queue, group string) error {
+func (q queueGroupAffinity) predicate(queue, group string, nonstrict bool) error {
```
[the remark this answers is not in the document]
kingeasternsun: ut needed.

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -109,11 +115,14 @@ func (q queueGroupAffinity) predicate(queue, group string) error {
 	return nil
 }
 
-func (q queueGroupAffinity) score(queue string, group string) float64 {
+func (q queueGroupAffinity) score(queue, group string, nonstrict bool) float64 {
 	nodeScore := 0.0
 	if len(queue) == 0 {
 		return nodeScore
 	}
+	if nonstrict {
```
kingeasternsun: ut needed.

wuyueandrew: ut done

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -34,17 +34,21 @@ const (
...
 
-	BaseScore = 100
+	BaseScore        = 100
+	NonStrictPenalty = -2
 )
 
 type nodeGroupPlugin struct {
 	// Arguments given for the plugin
 	pluginArguments framework.Arguments
+	nonstrict       bool
```
Monokaix: I prefer to set a strict flag instead of nonstricet, which is more understandable, and it's defaut value is true.

Monokaix:
And please also update the ndoegroup design doc to include this parameter and merge to one commit is OK.

wuyueandrew:
> And please also update the ndoegroup design doc to include this parameter and merge to one commit is OK.

done

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -55,13 +55,15 @@ const (
...
-
-	queueAttrs map[api.QueueID]*queueAttr
+	strict          bool
+	queueAttrs      map[api.QueueID]*queueAttr
 }
 
 // New function returns prioritize plugin object.
 func New(arguments framework.Arguments) framework.Plugin {
-	return &nodeGroupPlugin{pluginArguments: arguments}
+	nodeGroupPlugin := &nodeGroupPlugin{pluginArguments: arguments, strict: true}
```
JesseStutler, line 64: I think we should also update nodegroup plugin user guide? But we will release v1.13.0 soon, if you have limit time to update it, you can also seperate in the other PR @wuyueandrew

docs/user-guide/how_to_use_nodegroup_plugin.md
```diff
@@ -84,6 +84,28 @@ label `nodeGroupAffinity.requiredDuringSchedulingIgnoredDuringExecution` or `nod
...
+- plugins:
+  - name: priority
+  - name: gang
+  - name: conformance
+- plugins:
+  - name: drf
+  - name: predicates
+  - name: proportion
+  - name: nodegroup
+    strict: false
```
JesseStutler, line 105: I think it's
```
- name: nodegroup
   arguments:
     strict: false
```

docs/user-guide/how_to_use_nodegroup_plugin.md
```diff
@@ -84,6 +84,28 @@ label `nodeGroupAffinity.requiredDuringSchedulingIgnoredDuringExecution` or `nod
 kubectl get po job-1-nginx-0 -o wide 
 ```
 
+### Nodegroup Plugin Strict Configuration
+
+Strict is a switch to specific behavior of unset affinity queue. If `strict: false`, workload assigned to a unset affinity queue could be scheduled to Node with empty label `volcano.sh/nodegroup-name`. Otherwise, workload cannot be scheduled.
```
JesseStutler, line 89: Seems that it's `if the node doesn't have the "volcano.sh/nodegroup-name" label or the queue doesn't set affinity` ?

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -269,19 +271,32 @@ func (np *nodeGroupPlugin) OnSessionOpen(ssn *framework.Session) {
...
 		}
 
 		group, exist := node.Node.Labels[NodeGroupNameKey]
 		if !exist {
-			return newFitErr(task, node, errNodeGroupLabelNotFound)
+			if np.strict || !unsetAffinity {
+				return newFitErr(task, node, errNodeGroupLabelNotFound)
+			} else {
+				return nil
+			}
```
kingeasternsun, line 291:
```suggestion
			if np.strict || !unsetAffinity {
				return newFitErr(task, node, errNodeGroupLabelNotFound)
			}
			return nil
```

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -269,19 +271,32 @@ func (np *nodeGroupPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		job := ssn.Jobs[task.Job]
+		attr := np.queueAttrs[job.Queue]
+		unsetAffinity := attr.affinity == nil || (attr.affinity.queueGroupAffinityRequired.Len() == 0 && attr.affinity.queueGroupAffinityPreferred.Len() == 0)
 		if node.Node.Labels == nil {
-			return newFitErr(task, node, errNodeGroupLabelNotFound)
+			if np.strict || !unsetAffinity {
+				return newFitErr(task, node, errNodeGroupLabelNotFound)
+			} else {
+				return nil
+			}
```
kingeasternsun, line 282:
```suggestion
			}
			return nil
			
```

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -269,19 +271,32 @@ func (np *nodeGroupPlugin) OnSessionOpen(ssn *framework.Session) {
...
 
-		job := ssn.Jobs[task.Job]
-		attr := np.queueAttrs[job.Queue]
 		if attr.affinity == nil {
-			return newFitErr(task, node, errNodeGroupAffinityNotFound)
+			if np.strict {
+				return newFitErr(task, node, errNodeGroupAffinityNotFound)
+			} else {
+				return nil
+			}
```
kingeasternsun, line 299:
```suggestion
			
			return nil
			
```

JesseStutler:
@kingeasternsun @Monokaix Consider we need to release new version asap and @wuyueandrew has no time currently, I have made a new PR and we can merge that #4652

PR #4608, opened 2025-09-10
Sync kube-scheduler: Improve CSILimits plugin accuracy by using VolumeAttachments

guoqinwill:
#### What type of PR is this?
/kind bug

Starting from version 1.32, kube-scheduler supports retrieving volumeattachment to accurately calculate CSILimits. Volcano also needs to be updated accordingly. For details, see the Kubernetes pr：https://github.com/kubernetes/kubernetes/pull/127757 

Fixes #    See the Kubernetes issue：https://github.com/kubernetes/kubernetes/issues/126502

```release-note

```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -66,6 +66,7 @@ import (
 	cpuinformerv1 "volcano.sh/apis/pkg/client/informers/externalversions/nodeinfo/v1alpha1"
 	vcinformerv1 "volcano.sh/apis/pkg/client/informers/externalversions/scheduling/v1beta1"
 	topologyinformerv1alpha1 "volcano.sh/apis/pkg/client/informers/externalversions/topology/v1alpha1"
+
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/features"
 	schedulingapi "volcano.sh/volcano/pkg/scheduler/api"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -638,6 +640,8 @@ func (sc *SchedulerCache) addEventHandler() {
 	sc.pvInformer.Informer()
 	sc.scInformer = informerFactory.Storage().V1().StorageClasses()
 	sc.scInformer.Informer()
+	sc.vaInformer = informerFactory.Storage().V1().VolumeAttachments()
+	sc.vaInformer.Informer()
```
kingeasternsun, line 644: Where this  `sc.vaInformer` been used?

kingeasternsun: In Volcano scheduler’s predicate plugin, the `Filter` method from `k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodevolumelimits` is already used for volume-related pre-filtering. As shown in [[Kubernetes PR #127757](https://github.com/kubernetes/kubernetes/pull/127757/files)](https://github.com/kubernetes/kubernetes/pull/127757/files), `nodevolumelimits` already includes validation for `VolumeAttachments`. Therefore, it seems this PR may not be necessary.

Monokaix:
So there will be a rbac permission error in volcano when using pvc?

pkg/scheduler/cache/cache.go
Monokaix: So we need to know the accurate error.

Monokaix: The InformerFactory stared first in volcano, it's late to register informer in kube-scheduler plugin, so we shoulld register it before infromerFactory start. @kingeasternsun

kingeasternsun: > The InformerFactory stared first in volcano, it's late to register informer in kube-scheduler plugin, so we shoulld register it before infromerFactory start. @kingeasternsun

Got it. This comment could be resolved. 
Ignore my comment @guoqinwill .

Monokaix:
Please also cherryick to release1.12 and 1.11

guoqinwill:
> Please also cherryick to release1.12 and 1.11
Only release1.12  requires a cherry-pick, because 1.11 is version 1.31, and the use of VA for CSILimits starts from 1.32.

Issue #1451, opened 2021-05-03
lose preemptor when considering Preemption between Tasks within same Job 


lowang-bh:
**What happened**: lose preemptor task when considering Preemption between Task within Job due to it will pop task from preemptorTasks when considering Preemption between Jobs within Queue

**What you expected to happen**: tasks should keep original numbers when considering Preemption between Task within Job

**How to reproduce it (as minimally and precisely as possible)**:

```
create(){
    kubectl apply -f - <<EOF
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  annotations:
    useKubeBatch: "true"
  name: preeptee
  namespace: default
spec:
  backoffLimit: 2
  completions: 2
  parallelism: 2
  ttlSecondsAfterFinished: 600  
  template:
    metadata:
      annotations:
        scheduling.k8s.io/group-name: group1
    spec:
      containers:
      - image: busybox
        imagePullPolicy: IfNotPresent
        name: busybox
        command: ['/bin/sh']
        args: ['-c', 'sleep 160']        
        resources:
          requests:
            cpu: 1000m
          #limits:
          #  nvidia.com/gpu: 3             
      restartPolicy: Never
      terminationGracePeriodSeconds: 5
      schedulerName: kube-batch
---
apiVersion: scheduling.incubator.k8s.io/v1alpha1
kind: PodGroup
metadata:
  name: group1
  namespace: default
spec:
  minMember: 1
  queue: default
EOF

    sleep 15
    kubectl apply -f - <<EOF
apiVersion: batch.volcano.sh/v1alpha1
kind: Job
metadata:
  name: preeptor
  namespace: default
spec:
  backoffLimit: 1
  completions: 1
  parallelism: 1
  ttlSecondsAfterFinished: 600  
  template:
    metadata:
      annotations:
        scheduling.k8s.io/group-name: group1
    spec:
      containers:
      - image: busybox
        imagePullPolicy: IfNotPresent
        name: busybox
        command: ['/bin/sh']
        args: ['-c', 'sleep 120']        
        resources:
          requests:
            cpu: 1000m
      restartPolicy: Never
      terminationGracePeriodSeconds: 5
      schedulerName: kube-batch
      priorityClassName: high-priority
EOF
}

case $1 in
    create)
        create
        ;;
    delete)
        kubectl delete job preeptee
        kubectl delete job preeptor
        kubectl delete podgroups.scheduling.incubator.k8s.io group1
       ;;
[...]

PR #4613, opened 2025-09-13
Fix undefined behaviour in preemption action for multi queue scenario

Osykov:
#### What type of PR is this?
Fix undefined behaviour in preemption action

Currently code under the "Preemption between Task within Job" is called inside the "Preemption between Jobs within Queue." loop. That means that it might be called multiple times. `preemptorTasks` map is filled with tasks for starving jobs in the beginning of the function and then used in "Preemption between Jobs within Queue" loop, but in "Preemption between Task within Job." loop `preemptorTasks` is getting fully overwritten. 

Undefined behaviour scenario:
q1 has no preemptors
q2 has preemptorTask for starving job
Due to the nature of golang map structure it doesn't guarantee iteration order. In this case there are two options:
1. q2 then q1: the starving job's preemptor are tasks correctly read and processed further to determine preemptable pods
2. q1 then q2: here q2 doesn't have preemptors and break the loop inner loop, then it jumps directly to "Preemption between Task within Job.", on q1 iteration `preemptorTasks` already overwritten for the starving job and the initial preemptor tasks are lost.

- This PR moves "Preemption between Task within Job." loop out of Preemption between Jobs within Queue" loop, so preemption between tasks happen only once.
- Makes preempt action to honor queues order if some plugin specify it with `AddQueueOrderFn`. That means plugin can control what queue preemptor jobs are served first
- Adds preempt action test for multi queue environment, and extends tests tiers to use capacity plugin queue order function to make the test behaviour deterministic.

```release-note
- Fix undefined behaviour of preempt action in multi queue scenario 
- Makes preempt action to honor a plugin's queue order function
```


pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -146,9 +142,20 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 		}
 	}
 
+	// If plugin defines queue order function, use it to order queues.
+	queues := util.NewPriorityQueue(ssn.QueueOrderFn)
```
hajnalmt, line 146: This is a nice addition. I consider this a missing feature in the preemption logic.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -214,56 +221,56 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
...
-			preemptorTasks[job.UID] = util.NewPriorityQueue(ssn.TaskOrderFn)
-			for _, task := range job.TaskStatusIndex[api.Pending] {
-				// Again, skip scheduling gated tasks
-				if task.SchGated {
-					continue
-				}
-				preemptorTasks[job.UID].Push(task)
+	// Preemption between Task within Job.
+	for _, job := range underRequest {
+		// Fix: preemptor numbers lose when in same job
```
hajnalmt, line 228: I was wondering that why can't we just delete this part of the code.

It was added here:
https://github.com/volcano-sh/volcano/pull/1453
Due to this issue:
https://github.com/volcano-sh/volcano/issues/1451

So the problem in this issue was kind of simlar to yours, that preemptors are disappearing because we are using:
```go
preemptor := preemptorTasks[job.UID].Pop().(*api.TaskInfo)
```
at line 246 in your code.
It was causing actual preemptor loss, but in general not just in certain cases. It's incredible to me that this seemed a good idea to rebuild the `preemptorTasks` map at the iteration of every queue.

I understand the bug too so seemingly the logic is the same in the jobStarving loop and in the "Preemption between Task within Job" loop:
```go
			preemptorTasks[job.UID] = util.NewPriorityQueue(ssn.TaskOrderFn)
			for _, task := range job.TaskStatusIndex[api.Pending] {
				if task.SchGated {
					continue
				}
				preemptorTasks[job.UID].Push(task)
			}
```
which means that overwriting will happen with the original `preemptorTasks` map, in the end it will be later Pop-ed out in the  "Preemption between Task within Job" loop resulting in all the preemptors loss on the next iteration.

The solution doesn't seem right though, just moving it out from the queues loop disregards the queue order handling you actually introduced. I think what needs to be done instead, is that we create a new map for the queue with the `preemptorTasks` in it, that are belonging to the actual queue we are iterating on and iterate over these `preemptorTasks` afterwards. This way the queues loop does not touch the preemptorTasks map at all.

pkg/scheduler/actions/preempt/preempt_test.go
```diff
@@ -556,6 +558,28 @@ func TestTopologyAwarePreempt(t *testing.T) {
 			ExpectEvictNum: 1,
 			ExpectEvicted:  []string{"c1/preemptee2"},
 		},
+		{
+			Name: "preemption with multiple queues",
```
hajnalmt, line 562: This test case is for queue priorities, can we rename it to
`preemption with prioritiy queues` ?

hajnalmt (changes requested):
Hello @Osykov,

Thank you for this contribution, you have found a quite hard bug. The introduction of the QueueOrder honoration by the preempt action is nice feature. 

I have reviewed your code, and additionally have some minor remarks:
- Please use a better commit message (more then just "init")
- I want to see an unit test case for the issue which is `flaky` but fixed with your PR, the one you wrote is for the priority queues and not for the bug.

hajnalmt:
/cc @lowang-bh @JesseStutler 

This gentleman found a bug in the preemption logic and added a feature about the queue ordering.
The bug is connected to this PR:
https://github.com/volcano-sh/volcano/pull/1453
Fixing this issue:
https://github.com/volcano-sh/volcano/issues/1451
Basically the problem the PR tried to fix, wasn't fully fixed unfortunately and we still have preemptors that get lost. On the other hand his implementation disregards the queue ordering which was actually introduced by him.

Maybe we should split the queue ordering feature he added and the preemption bug he found into separate PRss and create actual issues to track them. What do you think?

pkg/scheduler/actions/preempt/preempt.go
JesseStutler: We can't delete this part of the code directly, we still have to deal with the preemption in the job @hajnalmt  Or you mean, we just need to delete this part:
```
preemptorTasks[job.UID] = util.NewPriorityQueue(ssn.TaskOrderFn)
		for _, task := range job.TaskStatusIndex[api.Pending] {
			// Again, skip scheduling gated tasks
			if task.SchGated {
				continue
			}
			preemptorTasks[job.UID].Push(task)
		}
```
 ?

JesseStutler:
> wasn't fully fixed unfortunately and we still have preemptors that get lost

Sorry I don't understand this very well, can you explain it in detail? I think his overall logic is still correct @hajnalmt

hajnalmt:
/close

I have created the proper continuation and issues for this (keeping your original commit too and mentioning you in the release notes):
https://github.com/volcano-sh/volcano/issues/5139
https://github.com/volcano-sh/volcano/issues/5140
https://github.com/volcano-sh/volcano/pull/5142 
https://github.com/volcano-sh/volcano/pull/5141

Thank you for the discovery and the PR @Osykov !

Issue #4417, opened 2025-06-29
use wrapper type and relative func to gen test object


lowang-bh:
### What is the problem you're trying to solve

when write ut, we need to generate some object such as pod,podgroups (eg: https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/util/test_utils.go).
There are many fileds to be fillup and use defined functions can not cover up all fields and the parames become too long.
in order to solve issue like:
https://github.com/volcano-sh/volcano/issues/3075#issuecomment-2368050656

for reference:

https://github.com/kubernetes/kubernetes/blob/b2f27c0649fc0f3d2a4a6dd29135ecc81781f7e4/pkg/scheduler/testing/wrappers.go#L295

https://github.com/kubernetes/kubernetes/blob/b2f27c0649fc0f3d2a4a6dd29135ecc81781f7e4/pkg/scheduler/testing/wrappers.go#L170

### Describe the solution you'd like

```
// PodGroupWrapper wraps a PodGroup inside.
type PodGroupWrapper struct{ schedulingv1beta1.PodGroup }

// MakePod creates a PodGroup wrapper.
func MakePodGroup() *PodGroupWrapper {
	return &PodGroupWrapper{schedulingv1beta1.PodGroup{}}
}

// Obj returns the inner PodGroup.
func (p *PodGroupWrapper) Obj() *schedulingv1beta1.PodGroup {
	return &p.PodGroup
}

// Name sets `s` as the name of the inner PodGroup.
func (p *PodGroupWrapper) Name(s string) *PodGroupWrapper {
	p.SetName(s)
	return p
}
```

sancheet230:
@lowang-bh have a look at #4420 i think this should be addressed first

JesseStutler:
Yes this way is better!

googs1025:
same as: https://github.com/volcano-sh/volcano/issues/4265

googs1025:
This is a reference wip pr  https://github.com/volcano-sh/volcano/pull/4315
that I am not currently planning to go on due to lack of bandwidth. You can use it as part of the reference @sancheet230

JesseStutler:
@sancheet230 Are you still working on this?

JesseStutler:
Wish someone can integrate it soon, do you have interest to inherit #4315 ? @GautamBytes

GautamBytes:
Sure @JesseStutler , would be happy to do it!
/assign

kev1N916:
[...]

PR #4601, opened 2025-09-06
Refactored unit tests to use Object wrappers and relative functions

kev1N916:
#### What type of PR is this?

used wrapper type and relative functions to generate tests objects for unit tests

#### What this PR does / why we need it:

Currently the tests objects for unit tests are generated through functions which take a large number of arguments  
There are many fields to be filled up in these functions and user defined functions can not cover up all fields and the parameters become too long.

We can solve this issue by using the ObjWrappers and relative functions to generate test objects  

#### Which issue(s) this PR fixes:

Fixes [#4420](https://github.com/volcano-sh/volcano/issues/4420) , [#4417](https://github.com/volcano-sh/volcano/issues/4417)

#### Does this PR introduce a user-facing change?

NO

PR #4618, opened 2025-09-17
Refactored the Unit tests to use Object Wrappers

### LINKED ISSUE #4420: bug: scheduler does not bind pods even when plugins permit allocation
### Description

When running unit tests for the Volcano scheduler’s capacity plugin, pods are not being bound even though the plugin’s allocatable logic returns true and all resource checks pass. This occurs even in a minimal test case with only one node, one queue, one PodGroup, and one pod. The plugin’s logic is invoked, but the scheduler never issues a bind request, causing the test to fail with failed to get Bind request

1.Create a minimal test file(place in pkg/scheduler/plugins/capacity/capacity_minimal_test.go)
2.Run the test

expected: The pod should be scheduled and bound to the node, and the test should pass
received: the test fails with ```
  failed to get Bind request in case 0(minimal: one pod should be scheduled)
```

1.12.1

The same issue occurs in more complex tests (e.g., TestEnqueueAndAllocatable).
The session’s Jobs map, PodGroups, and pods are all set up correctly.
The plugin’s logic is invoked, but the scheduler never issues a bind request.
This may indicate a bug in the test harness (uthelper), the FakeBinder, or the scheduler’s action execution

kev1N916:
###  What type of PR is this?
used wrapper type and relative functions to generate tests objects for unit tests

Currently the tests objects for unit tests are generated through functions which take a large number of arguments
There are many fields to be filled up in these functions and user defined functions can not cover up all fields and the parameters become too long.

We can solve this issue by using the ObjWrappers and relative functions to generate test objects

Fixes https://github.com/volcano-sh/volcano/issues/4420 , https://github.com/volcano-sh/volcano/issues/4417

### Does this PR introduce a user-facing change?
NO

### DIFF
--- a/pkg/agent/config/configmgr_test.go
+++ b/pkg/agent/config/configmgr_test.go
@@ -24,7 +24,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
--- a/pkg/agent/utils/testing/pod.go
+++ b/pkg/agent/utils/testing/pod.go
@@ -19,7 +19,7 @@ package testing
 import (
 	"context"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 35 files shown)


hwdef:
This PR is too large and it is difficult to review it in a short time.🥲

kev1N916:
@hwdef This PR is actually just a small add on to PR:#4601 which has already been reviewed by @googs1025

kev1N916:
Would it be better if I break this PR into smaller parts so that it becomes be easier to review ??

lowang-bh:
hi，please sign youself in commit, and we can go on this pr

Issue #4393, opened 2025-06-23
2025 OSPP projects track


Monokaix:
There are 8 OSPP projects this year of Volcano community, see https://summer-ospp.ac.cn/org/orgdetail/3bad8469-f54d-4a44-9f29-c81fa24d3864?lang=zh
This issue is used to track all the projects.

PR #4624, opened 2025-09-18
feat: Add user guide for large-scale performance tuning

Freshwlnd:
**WHAT THIS PR DOES / WHY WE NEED IT**:

This PR introduces a new user guide, `how_to_tune_volcano_performance.md`, based on the findings from the OSPP 2025 (Open Source Promotion Plan) project. The project details can be found here: [OSPP Project Page](https://summer-ospp.ac.cn/org/orgdetail/3bad8469-f54d-4a44-9f29-c81fa24d3864).

The guide aims to help users and administrators better understand and optimize Volcano's performance in large-scale scenarios (e.g., thousands of pods). It details common performance bottlenecks identified during testing, such as:
1.  Webhook timeouts and overhead.
2.  API-Server contention between the controller and scheduler.
3.  The impact of controller worker thread configuration.

It then provides actionable, short-term tuning advice for these areas and discusses potential long-term architectural optimizations. This document will serve as a valuable resource for users running demanding batch workloads on Volcano.

**WHICH ISSUE(S) THIS PR FIXES**:

Related to #4393

**SPECIAL NOTES FOR YOUR REVIEWER**:

This contribution is the final deliverable for the OSPP 2025 project "Volcano Large-Scale Performance Testing and Tuning". The content is derived from extensive benchmark tests and analysis.

**RELEASE NOTE**:

```release-note
Add a new user guide for performance tuning in large-scale scenarios.
```


Freshwlnd:
**UPDATE BASED ON REVIEW FEEDBACK**:

I have made the following changes based on the feedback:
- **Localized Images**: All images referenced in the guide have been moved into the `volcano-sh/volcano` repository under `docs/images/` to prevent broken links.
- **Removed External Links**: All external hyperlinks to personal blogs have been removed to ensure the long-term integrity of the documentation.

hwdef:
PTAL
@Monokaix @lowang-bh @JesseStutler @wangyang0616

docs/user-guide/how_to_tune_volcano_performance.md
```diff
@@ -0,0 +1,571 @@
...
+    vcClient.SchedulingV1beta1().PodGroups(...).Create(..., pg, ...)
+    ...
+}
+```
+
+A PodGroup is created with a single API call, and each Job has only one PodGroup. The creation itself is relatively simple (it's just a logical unit) and is unlikely to be a major bottleneck:
+* A PodGroup is essentially a CRD object (with a Spec & Metadata of only a few dozen bytes, as defined in `pkg/controllers/job/job_controller_actions.go`). The creation process is just a single write operation from kube-apiserver to etcd.
+* It does not involve scheduling decisions, node communication, or resource calculations. It returns immediately upon success without subsequent long-running processes.
+
+However, it's worth noting that if `MinMember` is set too high or Queue resources are insufficient, the scheduler may still block Job startup in a **later stage** because the PodGroup conditions are not met. This is a scheduling issue, not a CREATE issue. Additionally, if multiple components (like the Controller and Scheduler) update a PodGroup simultaneously, the operation may fail with a `409 Conflict` error due to a `resourceVersion` mismatch, requiring a retry.
```
JesseStutler: Just want to ensure that does vc-controller and vc-scheduler still have 409 conflicts on podgroup updating? Which of version of volcano do you use? And did you find which part of codes may cause updating conflict?

docs/user-guide/how_to_tune_volcano_performance.md
```diff
@@ -0,0 +1,571 @@
...
+    waitCreationGroup.Wait()  // ⬅ Blocking: does not return until the entire batch is complete
+    ...
+}
+```
+
+> **Observation**: All pods within a Job must be created in the same batch. The `Wait()` call blocks the worker goroutine from servicing other Jobs, becoming a performance bottleneck.
+
+Unlike PodGroups, Pods are native K8s objects with many complex fields. They require more default field population and take longer to write to etcd, making them a more likely bottleneck.
+
+Similar to the PodGroup stage, if multiple components (like the Controller and Scheduler) update a Pod simultaneously, the operation can fail with a `409 Conflict` error due to a `resourceVersion` mismatch, requiring a retry.
```
JesseStutler: Just want to know that compare to other scheduelers, does vc-scheduler will cause more 409 conflicts? Vc-Controller will only update the podgroup affliation for pods and vc-scheduler will only update pod's scheduling status and bind the pod

docs/user-guide/how_to_tune_volcano_performance.md
```diff
@@ -0,0 +1,571 @@
...
+
+```go
+// Controller's Pod creation request flow
+JobController.syncJob() 
+  → kubeClient.CoreV1().Pods().Create() 
+    → kube-apiserver 
+      → etcd write
+
+// Scheduler's Pod scheduling request flow
+VolcanoScheduler.allocate()
```
JesseStutler, line 386: vc-controller create pod and and vc-scheduler update will not generate 409 conflicts here, only after the controller is created, the scheduler can watch the latest resourceVersion pod and try to schedule, what does this request flow intend to express?

docs/user-guide/how_to_tune_volcano_performance.md
```diff
@@ -0,0 +1,571 @@
...
+
+    Event Type: default_vc-controller-manager/v0.0.0 (linux/amd64) kubernetes/$Format_create_podgroups
+    Occurrences: 623
+    Event Characteristics:
+        namespace: default
+        userAgent: vc-controller-manager/v0.0.0 (linux/amd64) kubernetes/$Format
+        verb: create
+        resource: podgroups
+        code: 409
+        reason: AlreadyExists
```
JesseStutler: Why does the AlreadyExists error occur? Is it because there are duplicate creation requests for the same podgroup?

docs/user-guide/how_to_tune_volcano_performance.md
```diff
@@ -0,0 +1,571 @@
...
+
+    Event Type: default_vc-controller-manager/v0.0.0 (linux/amd64) kubernetes/$Format_update_jobs/status
+    Occurrences: 705
+    Event Characteristics:
+        namespace: default
+        userAgent: vc-controller-manager/v0.0.0 (linux/amd64) kubernetes/$Format
+        verb: update
+        resource: jobs/status
+        code: 409
+        reason: Conflict
```
JesseStutler: Also quite strange about the jobs/status error. Only vc-controller will update the jobs/status, why there are so many jobs/status updating conflicts? Is it because there are pending updates for the same job in kube-apiserver, and after the previous requests are processed successfully, it causes the later job updates to fail?

JesseStutler:
Thanks for your contribution! @Freshwlnd 

Do you have interests to optimize the performance? Especially about the controller and scheduler conflicts. Or we can create a separate issue and provide testing guide to let other contributors optimize

docs/user-guide/how_to_tune_volcano_performance.md
Freshwlnd: In our tests we observed controller CREATE PodGroup AlreadyExists and scheduler UPDATE PodGroups contention under load. We did not capture controller UPDATE PodGroups in our current audit policy. We have reproduced the alternating “CREATE vs SCHEDULE” behavior in both v1.11.0 and v1.12.0‑alpha.0 (public list: https://freshwlnd.github.io/2025/08/20/k8s/k8s-scheduler-performance-volcano-version/).
- What may cause it (ongoing investigation):
    - CREATE/AlreadyExists: repeated reconcile before informer visibility converges can re‑issue a CREATE for the same PG.
    - UPDATE/Conflict (in general): concurrent writers on the same object with stale resourceVersion. We’re reviewing PG update paths and will provide focused diffs later.
- Doc status: to avoid over‑interpretation, we removed 409‑specific details in this revision and kept the higher‑level contention narrative; we’ll add back precise evidence once fully validated.

docs/user-guide/how_to_tune_volcano_performance.md
Freshwlnd: Not severe: under high concurrency, transient conflicts/retries are expected; many client paths explicitly ignore AlreadyExists or use RetryOnConflict.
- Concrete code in Volcano:
  - Ignore AlreadyExists on Pod create (controller): pkg/controllers/job/job_controller_actions.go:465
  - Ignore AlreadyExists on PodGroup create (controller): pkg/controllers/job/job_controller_actions.go:772
- These patterns match common Kubernetes practice: treat AlreadyExists as benign and wrap updates/creates in RetryOnConflict, which is why we don’t consider the phenomenon “severe”, but we’re still pursuing ways to reduce frequency at scale.

docs/user-guide/how_to_tune_volcano_performance.md
Freshwlnd: Sorry for the confusion. The request‑flow figure is NOT a “flow that causes 409.” Its intent is to show that both vc-controller-manager and vc-scheduler issue mutating requests to kube‑apiserver (which writes to etcd), so contention is possible. The exact request sequences that produce 409‑style retries will be presented in a later update after we validate the evidence. We revised the doc to this clarified narrative and removed uncertain 409 details for now.

docs/user-guide/how_to_tune_volcano_performance.md
Freshwlnd: - AlreadyExists: repeated CREATE during the visibility gap (reconcile retries before caches converge).
- Conflict: concurrent writes on the same object with stale resourceVersion. Both are normal symptoms of optimistic concurrency in high‑concurrency runs.

We temporarily removed the noisy dataset‑specific counts from the doc and will follow up with a tighter, reproducible evidence set.

docs/user-guide/how_to_tune_volcano_performance.md
Freshwlnd: - AlreadyExists: repeated CREATE during the visibility gap (reconcile retries before caches converge).
- Conflict: concurrent writes on the same object with stale resourceVersion. Both are normal symptoms of optimistic concurrency in high‑concurrency runs.

We temporarily removed the noisy dataset‑specific counts from the doc and will follow up with a tighter, reproducible evidence set.

Freshwlnd:
> Thanks for your contribution! @Freshwlnd
> 
> Do you have interests to optimize the performance? Especially about the controller and scheduler conflicts. Or we can create a separate issue and provide testing guide to let other contributors optimize

I simplified the doc: the visible alternation comes from the Controller (CREATE) and the Scheduler (SCHEDULE) concurrently writing to the API‑Server/etcd, which causes short‑term retries/queuing at scale; I also removed dataset‑specific 409 details that are not yet fully validated.

Next: I am continuing the deeper investigation and will open an issue to summarize findings, potential optimizations, and a reproducible guide; I will update the doc/PR with verified evidence.

hwdef:
/lgtm

Excellent work!

PR #4407, opened 2025-06-25
Refactor reclaim action to process all tasks in a job

GautamBytes:
#### What type of PR is this?
/kind improvement /kind bug
#### What this PR does / why we need it:
This PR fixes a critical bug in the scheduler's reclaim action where an entire queue could be incorrectly skipped during preemption.

The previous logic would evaluate only the first task of a starving job. If that single task failed a preliminary check (e.g., its preemptionPolicy was set to Never), the scheduler would discard the entire queue for that cycle. This prevented other valid, preemptable tasks in the same job, or other jobs in the same queue, from ever being considered for reclamation.

To fix this, the reclaim action's main loop has been refactored to mirror the robust nested queue -> job -> task structure found in the allocate action. This ensures:

- All candidate tasks within a job are correctly evaluated.
- The scheduler's preemption logic is more fair and correct.
- Performance is improved by removing an inefficient pop/push cycle.
- Code consistency across core scheduler actions is increased.

#### Which issue(s) this PR fixes:

Fixes #3738 

#### Special notes for your reviewer:
The core of this change is refactoring the reclaim action's main loop to align with the existing, proven pattern in the allocate action. This not only fixes the bug but also improves code consistency and performance. A new unit test has been added to specifically cover this failure scenario.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4634, opened 2025-09-22
Refactored reclaim action so that all tasks and jobs can be considered

### LINKED ISSUE #3738: Need to refactor the reclaim action

kev1N916:
What type of PR is this?
/kind bug

This PR fixes a critical bug in the scheduler's reclaim action where only the first job in a queue and the first task in a job are considered during the reclamation loop

The previous logic would evaluate only the first task of a starving job. If that single task failed a preliminary check (e.g., its preemptionPolicy was set to Never), the scheduler would discard the entire queue and that particular job for that cycle. This prevented other valid, preemptable tasks in the same job, or other jobs in the same queue, from ever being considered for reclamation.

To fix this, the reclaim action's main loop will keep iterating until all the jobs in a queue and all the tasks in a job have been processed. 

Which issue(s) this PR fixes:
Fixes https://github.com/volcano-sh/volcano/issues/3738

A new unit test has been added where the first and second Task in a job are not able to reclaim resources but the third Task can.

No

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -108,13 +108,14 @@ func (ra *Action) Execute(ssn *framework.Session) {
 		// Found "high" priority job
 		jobs, found := preemptorsMap[queue.UID]
 		if !found || jobs.Empty() {
-			continue
+			continue // if there are no jobs in the queue then we do not have to consider the queue
 		} else {
 			job = jobs.Pop().(*api.JobInfo)
 		}
 
 		// Found "high" priority task to reclaim others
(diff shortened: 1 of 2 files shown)


kev1N916:
@Monokaix @JesseStutler Could you review this when you find the time?

kev1N916:
CC @JesseStutler

JesseStutler:
/cc
Since lots of users have mentioned reclaim, we may consider refactor reclaim in this version

JesseStutler:
@kev1N916 I think it's not enough, we may need to refactor the whole framework of the reclaim action, just like preempt. This one is better: https://github.com/volcano-sh/volcano/pull/4407

PR #4637, opened 2025-09-24
Feature: (opt-in) Gang Aware Preemption

### LINKED ISSUE #4607: Volcano scheduler doesn’t reclaim/preempt jobs as a gang.

When reclaiming resources, Volcano checks every node in the cluster. For each node, it looks for pods that it can evict ("victims"). It keeps doing this node by node until it frees up enough resources to run the new task.

This method can cause problems for clusters running distributed AI training jobs.

For example:
Suppose your cluster is running 10 distributed training jobs (Job 1 through Job 10), and each job has 10 pods. With the current way Volcano works, it might evict only 1 pod from each job to reclaim resources. This means all 10 jobs fail, since each job needs all its pods (“gang scheduling”).

But we’d prefer Volcano to evict all 10 pods from just one job (say, Job 1). That way, only Job 1 fails, and the other 9 jobs keep training.

TLDR: We want Volcano to fully evict a single job rather than partially evicting many jobs, which would cause more overall failures.

We want Volcano to remove (evict) as few jobs as possible when freeing up resources.
Just like with gang scheduling—where all pods in a job are scheduled together—Volcano can evict an entire job together, as a “gang,” instead of evicting a few pods from many jobs.

This means:

Volcano will try to evict full jobs, not just single pods from different jobs.
This way, fewer jobs are affected, and more jobs can keep running without being interrupted.

### Additional context
We are using this feature with Raycluster and PytorchJob CRD

_No response_

mohitvinchoo:
#### What type of PR is this?
Feature: Gang-aware preemption modes for AI/ML training and other gang-scheduled workloads (PodGroup/VCJob).

Today, preemption picks nodes first and only then decides which tasks to evict, without considering gang (PodGroup/VCJob) boundaries. As a result, victims can span multiple jobs, and the Gang plugin ends up acting as a guard-rail (via minMember) rather than guiding which gang should be preempted. This can yield partial/fragmented evictions that are sub-optimal for AI/ML jobs.

This PR introduces an opt-in gang-aware preemption strategy that selects victims at the gang level. It reduces cross-job churn, preserves job atomicity when desired, and better aligns with workloads that need “all-or-nothing” capacity.

Modes (policy):
- `disabled` (default): Existing behavior. Preemption remains node-first and task-level; no gang semantics applied.
- `minimal`: Preemption will group preemptable tasks on each node into gangs then
  - On each candidate node, group preemptable tasks by gang.
  - Prefer a single gang whose eviction satisfies the preemptor task’s request (most cases).
  - If no single gang suffices, fall back to a greedy gang selection (still within gang units) until the preemptor fits (rare).
- `atomic`:  Same selection as minimal, but if a victim gang is chosen, evict all of its tasks across all nodes (cluster-wide gang eviction). This matches AI/ML expectations where partial eviction often dooms the job anyway.

- Fewer cross-job evictions: Victims come from a coherent gang, not scattered tasks across jobs.
- Predictability for gang jobs: Respects “start/stop together” intent; atomic ensures clean state.
- Backwards compatible: Fully opt-in; disabled preserves current behavior.

Partially fixes #4607.
In addition to the changes discussed in the above issue, this PR also adds:
- Handles node fragmentation.
- It also adds `atomic` mode.

Happy to create a new issue if needed.

- Objective:
Minimize the number of gangs disrupted on a node without hurting scheduler throughput.
- Why not try all k (1..J) combinations:
Per node combos = sum_{t=1..J} C(J, t) = 2^J − 1 (exponential)
Exponential time is unacceptable in a high-throughput scheduler like Volcano.
- Two-phase heuristic (fast, practical):
  - Phase 1 – Single-gang fit (minimal overage):
If any single gang can satisfy the preemptor’s need on a node, choose the best one (least overage).
Rationale: in most cases, preempting one gang frees enough capacity for the incoming task.
  - Phase 2 – Greedy multi-gang (minimize count, then overage):
If Phase 1 fails (e.g., preemptor is large), sort candidate gangs by size (largest first) and accumulate until the need is met.
Complexity: O(J log J) per node (sort) + O(J) accumulate — suitable for high QPS.
- Trade-off example (1D, non-negative):
Preemptor need: 80
Node gangs: [50, 50, 50, 50, 20, 20, 20, 20]
Greedy (fast): 50 + 50 = 100 → meets need with +20 overage
Optimal overage (exhaustive search): 20 + 20 + 20 + 20 = 80 → 0 overage
- Conclusion:
We accept small overage to preserve throughput and avoid exponential search.
Avoids the 2^J blow-up while keeping decisions near-optimal in practice.

Yes
```release-note
The following block needs to be added to `volcano-scheduler-configmap`
data:
  volcano-scheduler.conf: |
    actions: "...preempt..."
    configurations:
    - name: preempt
      arguments:
        gangPreemptionMode: off  # Accepted values: off (default) | minimal | atomic
```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -29,6 +29,7 @@ import (
 	"math"
 	"math/rand"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -57,7 +58,29 @@ const (
 	MinCandidateNodesPercentageKey = "minCandidateNodesPercentage"
 	MinCandidateNodesAbsoluteKey   = "minCandidateNodesAbsolute"
 	MaxCandidateNodesAbsoluteKey   = "maxCandidateNodesAbsolute"
+	GangPreemptionModeKey          = "gangPreemptionMode"
 )
+
+type GangPreemptionMode uint8
+
+const (
+	GPModeOff GangPreemptionMode = iota
+	GPModeMinimal
+	GPModeAtomic
+)
+
+func (m GangPreemptionMode) String() string {
+	switch m {
+	case GPModeOff:
+		return "off"
+	case GPModeMinimal:
+		return "minimal"
+	case GPModeAtomic:
+		return "atomic"
+	default:
+		return fmt.Sprintf("unknown(%d)", int(m))
+	}
+}
 
 type Action struct {
 	ssn *framework.Session
@@ -70,6 +93,8 @@ type Action struct {
 	minCandidateNodesPercentage   int
 	minCandidateNodesAbsolute     int
 	maxCandidateNodesAbsolute     int
+
+	gpMode GangPreemptionMode
 }
 
 func New() *Action {
@@ -80,6 +105,7 @@ func New() *Action {
 		minCandidateNodesPercentage:   10,
 		minCandidateNodesAbsolute:     1,
 		maxCandidateNodesAbsolute:     100,
+		gpMode:                        GPModeOff,
 	}
 }
 
@@ -97,7 +123,24 @@ func (pmpt *Action) parseArguments(ssn *framework.Session) {
 	arguments.GetInt(&pmpt.minCandidateNodesPercentage, MinCandidateNodesPercentageKey)
 	arguments.GetInt(&pmpt.minCandidateNodesAbsolute, MinCandidateNodesAbsoluteKey)
 	arguments.GetInt(&pmpt.maxCandidateNodesAbsolute, MaxCandidateNodesAbsoluteKey)
+	var gpModeStr string
+	arguments.GetString(&gpModeStr, GangPreemptionModeKey)
+	pmpt.gpMode = parseGPMode(gpModeStr)
 	pmpt.ssn = ssn
+}
+
+func parseGPMode(s string) GangPreemptionMode {
+	switch strings.ToLower(strings.TrimSpace(s)) {
+	case "", "off", "disabled":
+		return GPModeOff
+	case "minimal":
+		return GPModeMinimal
+	case "atomic":
+		return GPModeAtomic
+	default:
+		klog.V(3).Infof("Unrecognized Gang Preemption Mode, defaulting to `disabled`")
+		return GPModeOff
+	}
 }
 
 func (pmpt *Action) Execute(ssn *framework.Session) {
@@ -304,7 +347,9 @@ func (pmpt *Action) preempt(
 	if pmpt.enableTopologyAwarePreemption {
 		return pmpt.topologyAwarePreempt(ssn, stmt, preemptor, filter, predicateNodes)
 	}
-
+	if pmpt.gpMode != GPModeOff {
+		return pmpt.gangPreempt(ssn, stmt, preemptor, filter, predicateNodes)
+	}
 	return pmpt.normalPreempt(ssn, stmt, preemptor, filter, predicateNodes)
 }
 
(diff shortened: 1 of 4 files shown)


mohitvinchoo:
> /cc Will take a look after v1.13.0 released

Thank you! @JesseStutler

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -397,6 +429,217 @@ func (pmpt *Action) normalPreempt(
...
+	}
+	currentQueue := ssn.Queues[preemptorJob.Queue]
+	var preemptees []*api.TaskInfo
+	for _, node := range predicateNodes {
+		for _, task := range node.Tasks {
+			if filter == nil {
+				preemptees = append(preemptees, task.Clone())
+			} else if filter(task) {
+				preemptees = append(preemptees, task.Clone())
+			}
```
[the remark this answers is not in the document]
mohitvinchoo: Following pattern in existing code. Lets ignore this optimization and keep the code similar

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -294,7 +324,9 @@ func (pmpt *Action) preempt(
 	if pmpt.enableTopologyAwarePreemption {
 		return pmpt.topologyAwarePreempt(ssn, stmt, preemptor, filter, predicateNodes)
 	}
-
+	if pmpt.gpMode != GPModeOff {
```
hajnalmt, line 350: Yeah this implementation is exclusive with the topologyAwarePreemption. We should make a feature request separatly to align the two features. Maybe someone picks it up.

Also we shall document this on the website.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -397,6 +429,216 @@ func (pmpt *Action) normalPreempt(
...
+		for _, task := range node.Tasks {
+			if filter == nil {
+				preemptees = append(preemptees, task.Clone())
+			} else if filter(task) {
+				preemptees = append(preemptees, task.Clone())
+			}
+		}
+	}
+	preemptees = ssn.Preemptable(preemptor, preemptees)
+	nodeJobPreempteesMap := make(map[string]map[api.JobID][]*api.TaskInfo)
```
hajnalmt: The build up of this `nodeJobPreempteesMap` shall go into a separate function as it looks really useful to be used at other places too.
Maybe you can put into `scheduler/util` so reclaim action can reuse it.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -397,6 +429,216 @@ func (pmpt *Action) normalPreempt(
...
+) (jobs []api.JobID, node *api.NodeInfo) {
+	// ---- Phase 1: try best single-job victim (minimal overage) ----
+	var (
+		bestSingleOver *api.Resource
+		bestSingleJob  api.JobID
+		bestSingleNode *api.NodeInfo
+	)
+	for _, n := range selectedNodes {
+		idle := n.FutureIdle()
+		idle.SetScalar("pods", 0)
```
hajnalmt: Why did you set this to zero?

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -397,6 +429,216 @@ func (pmpt *Action) normalPreempt(
...
+		// Fits without preemption
+		if preemptor.InitResreq.LessEqual(idle, api.Zero) {
```
hajnalmt: Queue Allocatable is missed here completely.

This logic will return prematurely with a node and zero preemptee if the preemptor fits on the node resulting in zero victims. This makes this implementation completely disregarding queue allocation related settings so preemption will not work at all in this case.

I checked your unit tests and you don't have any test case that checks if the upper limit for preemption comes from the queue capability settings, only the node resource scarness related problems are covered.

Just create a test case where a queue has capability set and the node has enough resources to handle the preemptor.

**Remark**: The problem is far worse than just this description, the `sub` will error out with a panic. I tried to add the `preempt low priority job in same queue` test case from normal preempt to the gangpreempt and the result:
```go
--- FAIL: TestGangPreempt (0.10s)
    --- FAIL: TestGangPreempt/preempt_low_priority_job_in_same_queue (0.10s)
I1125 16:46:35.316281  509375 watch.go:218] "Stopping fake watcher"
panic: resource is not sufficient to do operation: <cpu 3000.00, memory 3000000000.00, pods 1.00> sub <cpu 9000.00, memory 9000000000.00, pods 0.00> [recovered]
        panic: resource is not sufficient to do operation: <cpu 3000.00, memory 3000000000.00, pods 1.00> sub <cpu 9000.00, memory 9000000000.00, pods 0.00>

goroutine 115 [running]:
testing.tRunner.func1.2({0x44abca0, 0xc000411400})
        /usr/local/go/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1737 +0x35e
panic({0x44abca0?, 0xc000411400?})
        /usr/local/go/src/runtime/panic.go:787 +0x132
volcano.sh/volcano/pkg/scheduler/util/assert.Assert(0xa6?, {0xc0000e0b40?, 0xc000067168?})
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/util/assert/assert.go:49 +0x165
volcano.sh/volcano/pkg/scheduler/util/assert.Assertf(0xe8?, {0x4c9dea6?, 0xc000747200?}, {0xc000067168?, 0x30?, 0x4c27596?})
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/util/assert/assert.go:59 +0x45
volcano.sh/volcano/pkg/scheduler/api.(*Resource).Sub(0xc0004ef5e0, 0xc0004ef5c0)
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/api/resource_info.go:265 +0x85
volcano.sh/volcano/pkg/scheduler/actions/preempt.(*Action).findBestPreemptionTarget(0xc0007470b0?, {0xc000092b00, 0x1, 0x1?}, 0xc000067798, 0xc00024c460)
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/actions/preempt/preempt.go:554 +0x245
volcano.sh/volcano/pkg/scheduler/actions/preempt.(*Action).gangPreempt(0xc000160a00, 0xc000312848, 0xc0004ef520, 0xc00024c460, 0xc0004ef540, {0xc000092a88, 0x1, 0x1})
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/actions/preempt/preempt.go:471 +0x610
volcano.sh/volcano/pkg/scheduler/actions/preempt.(*Action).preempt(0xc000160a00, 0xc000312848, 0xc0004ef520, 0xc00024c460, 0xc0004ef540, {0x5055160, 0xc000092a80})
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/actions/preempt/preempt.go:328 +0x239
volcano.sh/volcano/pkg/scheduler/actions/preempt.(*Action).Execute(0xc000160a00, 0xc000312848)
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/actions/preempt/preempt.go:211 +0x1591
volcano.sh/volcano/pkg/scheduler/uthelper.(*TestCommonStruct).Run(0xc000630248, {0xc0002936a0, 0x1, 0x8?})
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/uthelper/helper.go:200 +0x110
volcano.sh/volcano/pkg/scheduler/actions/preempt.TestGangPreempt.func1(0xc000441340)
        /home/uih20178/go/src/volcano.sh/volcano/pkg/scheduler/actions/preempt/preempt_test.go:1017 +0x233
testing.tRunner(0xc000441340, 0xc00042f4f0)
        /usr/local/go/src/testing/testing.go:1792 +0xf4
created by testing.(*T).Run in goroutine 114
        /usr/local/go/src/testing/testing.go:1851 +0x413
FAIL    volcano.sh/volcano/pkg/scheduler/actions/preempt        0.135s
```

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -397,6 +429,216 @@ func (pmpt *Action) normalPreempt(
...
+			return nil, n
+		}
+		need := preemptor.InitResreq.Clone()
```
hajnalmt: Let's call this variable `remainingResReq` or `remainingPreemptorReq`  maybe `unsatisfiedPreemptorReq` so it's more descriptive and you can do the assignment in one go:
```
remainingPreemptorReq := preemptor.InitResreq.Clone().Sub(idle)
```

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -397,6 +429,216 @@ func (pmpt *Action) normalPreempt(
...
 }
 
+func (pmpt *Action) gangPreempt(
+	ssn *framework.Session,
+	stmt *framework.Statement,
+	preemptor *api.TaskInfo,
+	filter func(*api.TaskInfo) bool,
+	predicateNodes []*api.NodeInfo,
+) (bool, error) {
+	klog.V(3).Infof("Running Gang Preemption with mode: %v", pmpt.gpMode)
```
hajnalmt: This log line is not too helpful since gpMode is an integer.
It would be nice to implement the Stringer interface on this.
```
func (mode GangPreemptionMode) String() string {
    switch mode {
    case GPModeOff:
        return "off"
    case GPModeMinimal:
        return "minimal"
    case GPModeAtomic:
        return "atomic"
    default:
        return "unknown"
    }
}
```
and use `%s` here instead.
Also it's too verbose, don't forget that this function get's called for every queue in the loop until we have starving jobs, 4 is more than enough.
It should look like:
`klog.V(4).Infof("Running Gang Preemption with mode: %s", pmpt.gpMode)`

pkg/scheduler/actions/preempt/preempt_test.go
```diff
@@ -611,6 +612,398 @@ func TestTopologyAwarePreempt(t *testing.T) {
...
+		proportion.PluginName:  proportion.New,
+	}
+	highPrio := util.BuildPriorityClass("high-priority", 100000)
+	lowPrio := util.BuildPriorityClass("low-priority", 10)
+	priority3, priority2, priority1 := int32(3), int32(2), int32(1)
+	testsMinimal := []uthelper.TestCommonStruct{
+		{
+			Name: "minimal mode: pick the gang with the least overage and exact fit",
+			PodGroups: []*schedulingv1beta1.PodGroup{
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 0, map[string]int32{}, schedulingv1beta1.PodGroupInqueue, "low-priority"),
```
hajnalmt, line 629: It's ok, and it doesn't matter in your test cases, but if you want to be precise you shall modify the Podgroup status to be `schedulingv1beta1.PodGroupRunning` instead of `Inqueue` because it has a pod Running in it. This bugs my eye in the other tests too.

pkg/scheduler/actions/preempt/preempt_test.go
```diff
@@ -611,6 +612,398 @@ func TestTopologyAwarePreempt(t *testing.T) {
...
+			ExpectEvictNum: 2,
+		},
+		{
+			Name: "minimal mode: after selecting the gang, preempt the lowest priority task for minimal mode",
+			PodGroups: []*schedulingv1beta1.PodGroup{
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 0, map[string]int32{}, schedulingv1beta1.PodGroupInqueue, "low-priority"),
+				util.BuildPodGroupWithPrio("pg2", "c1", "q1", 1, map[string]int32{}, schedulingv1beta1.PodGroupInqueue, "high-priority"),
+			},
+			Pods: []*v1.Pod{
+				util.BuildPodWithPriority("c1", "preemptee3", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string), &priority1),
```
hajnalmt: preemptee3 has priority1
preemptee2 has priority2
preemptee1 has priority3

Can we switch premptee3 and preemptee1 so the priorities are matching in the ordering? in that case preemptee1 will be evicted.

hajnalmt (changes requested):
Hello @mvinchoo,
Great implementation, and nice work regarding this issue. I really like the ideas you brought here, and architecturally I think this is the direction we need to follow. I wondered that maybe the `minimal` mode should be default but even if this is the case I would stay with your implementation and move it to be default later on.

First of all, I have started the review of this feature, but I stopped at the `minimal` part as it contains a major flaw, that we probably need to discuss. Please read my comments.
As a take on, I think all of the `normalPreempt` unit test cases should be covered by the feature too, and if something works differently we need to understand why. I will try to review the atomic mode too as soon as I can.

To not forget:
- We need a design document about this feature too and document the settings on the website too.
- Please squash your commits either to 1, 2 maybe 3 commits but not more, 13 is a lot.

Thank you for your contribution and all the time you have sacrificed on this change. It looks like you really thought it through.

pkg/scheduler/actions/preempt/preempt.go
hajnalmt: The panic comes from the reason that the pods dimension is zero in the idle, because of the `idle.SetScalar("pods", 0)` setting earlier. The test case just fails if you delete that line as I described the problem in the earlier comment.

mtian29:
Thanks. Have one question, does this address cases like this.

A task comes in requesting 40.
J1P1 means pod1 in job1.

Node 1: [ J1P1 20, J2P1: 20]]
Node 2: [J1P2 20, ....]

In this case, given current implementation, does it evict 
- J1P1 20 and J1P2 20  (this should be the wrong, because it spans two nodes and cannot fulfil 40) 
or
- J1P1 20 and J2P1 20 (this is correct)

mohitvinchoo:
> Hello @mvinchoo, Great implementation, and nice work regarding this issue. I really like the ideas you brought here, and architecturally I think this is the direction we need to follow. I wondered that maybe the `minimal` mode should be default but even if this is the case I would stay with your implementation and move it to be default later on.
> 
> First of all, I have started the review of this feature, but I stopped at the `minimal` part as it contains a major flaw, that we probably need to discuss. Please read my comments. As a take on, I think all of the `normalPreempt` unit test cases should be covered by the feature too, and if something works differently we need to understand why. I will try to review the atomic mode too as soon as I can.
> 
> To not forget:
> 
> * We need a design document about this feature too and document the settings on the website too.
> * Please squash your commits either to 1, 2 maybe 3 commits but not more, 13 is a lot.
> 
> Thank you for your contribution and all the time you have sacrificed on this change. It looks like you really thought it through.

Thank you @hajnalmt for your kind words and great insights.
I was away and busy with some other stuff.
I'll review the comments soon and make the changes.
I'm also a part of CNCF slack now.

mohitvinchoo:
> Thanks. Have one question, does this address cases like this.
> 
> A task comes in requesting 40. J1P1 means pod1 in job1.
> 
> Node 1: [ J1P1 20, J2P1: 20]] Node 2: [J1P2 20, ....]
> 
> In this case, given current implementation, does it evict
> 
> * J1P1 20 and J1P2 20  (this should be the wrong, because it spans two nodes and cannot fulfil 40)
>   or
> * J1P1 20 and J2P1 20 (this is correct)

Hi @mtian29 
Thank you for your question. Lets walk through your example.

Let's visit `findBestPreemptionTarget` function.
In stage 1, it will try to find a single job on a single node that satisfies the requirement.
For your example, It will not find any thing and proceed to stage 2.

In stage 2, it will start merging jobs in a greedy manner to select preemption targets.
So, for node1 both J1 and J2 will be picked.
In `minimal mode`:
We will only preempt tasks of these jobs running on node1. Tasks running on node2 will not be affected.

In `atomic mode`:
We will preempt tasks of these jobs across all nodes. (as they were doomed to fail anyway in future for ML training.)

At least this is the intention of that function.
Do let me know if you find any bugs in the implementation.

Unit test:
- `Name: "minimal mode: multiple gangs need to be preempted on the same node for a large preemptor"` Covers your example. However there is room to add many more tests.
- `Name: "atomic mode: pick best-fit gang on one node and evict all its tasks across nodes"` partially covers your case for atomic mode. I'll add one for minimal mode as well.

Thank you! 
Please feel free to leave more comments! 💯

pkg/scheduler/actions/preempt/preempt_test.go
mohitvinchoo: Done!

pkg/scheduler/actions/preempt/preempt_test.go
mohitvinchoo: Done!

pkg/scheduler/actions/preempt/preempt.go
mohitvinchoo: Done!

pkg/scheduler/actions/preempt/preempt.go
mohitvinchoo: Done!

pkg/scheduler/actions/preempt/preempt.go
mohitvinchoo: Done!

pkg/scheduler/actions/preempt/preempt.go
mohitvinchoo: Fixed this issue.
This was not the right approach as you correctly pointed out.

pkg/scheduler/actions/preempt/preempt.go
mohitvinchoo: Yes you are right, added a unit test.
Currently the unit test FAILS (does not PANIC, as I have fixed `idle.SetScalar("pods", 0)`).
`    --- FAIL: TestGangPreempt/FIX_IT:_preempt_low_priority_job_in_same_queue (0.70s)`

I'll fix this issue soon.

pkg/scheduler/actions/preempt/preempt.go
mohitvinchoo: 👍  Will open a new Issue, if we merge this PR.
It would be a nice candidate for `good first issue`.

pkg/scheduler/actions/preempt/preempt.go
mohitvinchoo: Fixed this.

mohitvinchoo:
> > Hello @mvinchoo, Great implementation, and nice work regarding this issue. I really like the ideas you brought here, and architecturally I think this is the direction we need to follow. I wondered that maybe the `minimal` mode should be default but even if this is the case I would stay with your implementation and move it to be default later on.
> > First of all, I have started the review of this feature, but I stopped at the `minimal` part as it contains a major flaw, that we probably need to discuss. Please read my comments. As a take on, I think all of the `normalPreempt` unit test cases should be covered by the feature too, and if something works differently we need to understand why. I will try to review the atomic mode too as soon as I can.
> > To not forget:
> > 
> > * We need a design document about this feature too and document the settings on the website too.
> > * Please squash your commits either to 1, 2 maybe 3 commits but not more, 13 is a lot.
> > 
> > Thank you for your contribution and all the time you have sacrificed on this change. It looks like you really thought it through.
> 
> Thank you @hajnalmt for your kind words and great insights. I was away and busy with some other stuff. I'll review the comments soon and make the changes. I'm also a part of CNCF slack now.

@hajnalmt Thank you for the first review iteration.
PR is ready for a second review. 👍 
While we review this change, I'll be testing out which `normalPreempt` test cases fail for `gangPreempt`.
The behavior change is big, I expect some cases to behave differently.

Thank you! 🚀

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -407,6 +452,269 @@ func (pmpt *Action) normalPreempt(
...
+	}
+
+	// Queue deficit: how much we must evict from the current queue (hard-cap model):
+	// need = max(Allocated + preemptor - Capability, 0)
+	queueNeed := func() *api.Resource {
+		capRL := currentQueue.Queue.Spec.Capability
+		usedRL := currentQueue.Queue.Status.Allocated
+
+		capRes := api.NewResource(capRL)
+		usedRes := api.NewResource(usedRL)
```
mohitvinchoo, line 582: All of this is `nil` / `0` for unit tests.
Since `proportion.go` derives these values from total cluster size, `ssn.Allocatable` works fine.
But these calculations fail.

Still thinking on how I can fix this so that all unit test can run just fine.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -407,6 +452,269 @@ func (pmpt *Action) normalPreempt(
...
+			vics = nodeJobPreempteesMap[targetNode.Name][jid]
+		} else {
+			for _, t := range ssn.Jobs[jid].Tasks {
+				vics = append(vics, t)
+			}
+		}
+		victims = append(victims, vics...)
+	}
+
+	if pmpt.gpMode == GPModeMinimal {
```
JesseStutler, line 506: I can understand the difference between minimal and atomic. Some tasks may failed because partial preemption, so we'd better to release all the tasks. But as a platform, we may not know the characteristics of the business, we don't know if the job can still run normally after preempting some tasks, and atomic and minimal are a configuration that cannot be adjusted dynamically according to the business after the scheduler starts. So the scalability will be poor

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -407,6 +452,269 @@ func (pmpt *Action) normalPreempt(
 	return assigned, nil
 }
 
+func (pmpt *Action) gangPreempt(
```
JesseStutler, line 455: There is a problem, `findBestPreemptionTarget` is based on a single preemptor, which means that a single large task preempt multiple small jobs. But if there are multiple small tasks want to preempt, that is, we want to M-N preemption, in current implementation, it may cause cross-job preemption

JesseStutler:
Hi @mvinchoo I want to push forward job level preemption in this version, I think we can move to https://github.com/volcano-sh/volcano/pull/4780 and keep evolving, WDYT? Welcome comment on the #4780 PR!

hajnalmt (commented):
/close

Thanks for the work on this we can close this in favor of:
https://github.com/volcano-sh/volcano/pull/4780
https://github.com/volcano-sh/volcano/pull/5250

Please try out the actions and give us feedback @mvinchoo 👍

PR #2393, opened 2022-07-29
support multi scheduler name for scheduler and webhook

jinzhejz:
#2394 
Signed-off-by: Zhe Jin <jinzhe.hit@gmail.com>

Currently, volcano support multiple scheduler name in controllers(https://github.com/volcano-sh/volcano/pull/1792). This pr is used to support multiple scheduler names for scheduler and webhook, include:
* support specify multiple scheduler names in command line
* filer schedulable pods by multiple scheduler names
* use `volcano` as default component name when specify multiple scheduler names
* choose one scheduler name for volcano job in webhook if the job's schedulerName is empty

PR #4650, opened 2025-09-26
fix: fixing constant cache warnings

### LINKED ISSUE #4381: Volcano Scheduler clutters logs and make troubleshooting more difficult
### Description

I've noticed the volcano-scheduler pod keeps throwing warnings such as "Failed to delete task" for pods using the default-scheduler.

These warnings are generally harmless for cluster operation. They do not prevent either Volcano-managed or standard workloads from running, but they can clutter logs and make troubleshooting more difficult.

I couldn't find documentation to filter the log noise.

1. Install Volcano; default values are ok
2. Create a namespace `test` and a dummy deployment "test", **without setting the volcano scheduler**. Use nginx for example.
3. Verify that `schedulerName: default-scheduler`
4. Tail the volcano-scheduler logs

**Example log**:
```
volcano-scheduler-5459dc465b-9bpm4:volcano-scheduler W0617 15:07:28.259063       1 event_handlers.go:364] Failed to delete task: errors:  1: task test/test-c5b7c89dc has null jobID
```

**My understanding**:

Volcano's scheduler expects to manage pods created by Volcano Jobs, which have an associated JobID. When it encounters pods (such as those from Deployments, StatefulSets, or other controllers not using Volcano) without this association, it logs a warning because it cannot find a Volcano Job context for the pod.

**My expectations**:

Being able to silence the warning. Otherwise troubleshooting becomes very difficult when NOT using Volcano as the solely scheduler.

1.12.1

_No response_

### LINKED ISSUE #4649: Scheduler is emitting warnings each time a pod is deleted
### Description

When I delete any pod on the cluster this Warning appears (always)
W0925 07:16:56.576685    1698 node_info.go:466] failed to find task <default/nginx2-67755588bf-pkfhp> on host <k3d-volcano-dev-agent-2>

It happens regardless of what manages the pod.

1. create a pod (kubectl create deployment nginx/kubectl run nginx --image=nginx)
2. delete a pod (kubectl delete deployment nginx/kubectl delete pod nginx)

No warnings. 

latest

_No response_

hajnalmt:
#### What type of PR is this?
/kind bug

When we added multiple schedulerNames for the scheduler:
https://github.com/volcano-sh/volcano/pull/2393
The logs were just adjusted to comply with new string slice and we have log lines like this:
`Pod volcano-system/volcano-scheduler-7f74555f66-9sxnn will not scheduled by []string{"volcano"}, skip creating PodGroup and Job for it`

Instead of printing out the go string slice directly we should join them. This looks smoother:
`I0924 14:17:40.622089    5860 event_handlers.go:72] Pod default/nginx2-67755588bf-pkfhp will not scheduled by volcano, skip creating PodGroup and Job for it`

As I was checking this part of the code, I tried to find the root cause for the warnings.
Every not Volcano managed pod will throw a Warning 
`W0924 14:17:43.237059    5860 event_handlers.go:371] Failed to delete task: errors:  1: task default/nginx2-67755588bf-pkfhp has null jobID`
This line was introduced here:
https://github.com/volcano-sh/volcano/pull/3379
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/cache/event_handlers.go#L341
This appears for every update event for each pod in the cluster not managed by volcano, because there is no jobId for the non-volcano managed pods.

The log entry is intimidating since it's not clear first that it's caching territory and nothing is wrong with the pod actually and it's a fairly common use-case.
We can safely switch to a `klog.V(4).Infof` instead of an `fmt.errorF` line.

In this deleteTask function the numaErr variable is not used actually, so I deleted it.

Additionally I realized an other issue, when a pod is deleted on the cluster we emit a Warning:
`W0924 15:58:35.180938   13225 node_info.go:466] failed to find task <default/nginx-676b6c5bbc-892h4> on host <k3d-volcano-dev-agent-2>`
We emit it several times for every deleted pod. The reason for it is that when an Update Event is dispatched:
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/cache/event_handlers.go#L314
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/cache/event_handlers.go#L254
In the addTask part we don't add back the task if it's in Terminating state:
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/cache/event_handlers.go#L226
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/cache/event_handlers.go#L62
So when podInformer notifies the cache to updateTask from Releasing to -> Succeeded/Failed the task is correctly
not added back to the `schedulerCache.nodes`.
We should not try to delete tasks from the `sc.Nodes` cache that are not there and emit a false Warning.
I have tried it several ways and this event is dispatched every single time, even when the pod is deleted forcefully without grace.

I have one minor addition too to the job_info to print the Node name out when a task_info is printed. Making debugging issues like these easier.
 
Fixes #4381
Fixes #4649

I have logs about the debugging, I can attach them in slack if it's needed.

```release-note
None
```

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -29,6 +29,7 @@ import (
 	"reflect"
 	"slices"
 	"strconv"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
@@ -68,8 +69,8 @@ func isTerminated(status schedulingapi.TaskStatus) bool {
 func (sc *SchedulerCache) getOrCreateJob(pi *schedulingapi.TaskInfo) *schedulingapi.JobInfo {
 	if len(pi.Job) == 0 {
 		if !slices.Contains(sc.schedulerNames, pi.Pod.Spec.SchedulerName) {
-			klog.V(4).Infof("Pod %s/%s will not scheduled by %#v, skip creating PodGroup and Job for it",
-				pi.Pod.Namespace, pi.Pod.Name, sc.schedulerNames)
+			klog.V(4).Infof("Pod %s/%s will not scheduled by %s, skip creating PodGroup and Job for it",
+				pi.Pod.Namespace, pi.Pod.Name, strings.Join(sc.schedulerNames, ","))
 		}
 		return nil
 	}
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -69,7 +69,7 @@ func isTerminated(status schedulingapi.TaskStatus) bool {
 func (sc *SchedulerCache) getOrCreateJob(pi *schedulingapi.TaskInfo) *schedulingapi.JobInfo {
 	if len(pi.Job) == 0 {
 		if !slices.Contains(sc.schedulerNames, pi.Pod.Spec.SchedulerName) {
-			klog.V(4).Infof("Pod %s/%s will not scheduled by %s, skip creating PodGroup and Job for it",
+			klog.V(4).Infof("Pod %s/%s is not scheduled by %s, skip creating PodGroup and Job for it in cache.",
 				pi.Pod.Namespace, pi.Pod.Name, strings.Join(sc.schedulerNames, ","))
 		}
 		return nil
@@ -292,7 +292,7 @@ func (sc *SchedulerCache) syncTask(oldTask *schedulingapi.TaskInfo) error {
 
 func (sc *SchedulerCache) updateTask(oldTask, newTask *schedulingapi.TaskInfo) error {
 	if err := sc.deleteTask(oldTask); err != nil {
-		klog.Warningf("Failed to delete task: %v", err)
+		klog.Warningf("Failed to delete task from cache: %v", err)
 	}
 
 	return sc.addTask(newTask)
@@ -322,7 +322,7 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
 	if err := sc.deletePod(oldPod); err != nil {
 		return err
 	}
-	//when delete pod, the ownerreference of pod will be set nil,just as orphan pod
+	//when delete pod, the ownerreference of pod will be set nil, just as orphan pod
 	if len(utils.GetController(newPod)) == 0 {
 		newPod.OwnerReferences = oldPod.OwnerReferences
 	}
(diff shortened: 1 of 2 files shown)


pkg/scheduler/cache/event_handlers.go
```diff
@@ -321,35 +322,40 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
+			}
 		}
 	}
 
 	if jobErr != nil || nodeErr != nil {
-		return schedulingapi.MergeErrors(jobErr, nodeErr, numaErr)
+		return schedulingapi.MergeErrors(jobErr, nodeErr)
 	}
 
 	return nil
```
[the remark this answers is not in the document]
hajnalmt: I checked it and Gemini is right, nor job.DeleteTaskInfo neither node.RemoveTask actually returns an error which looks at least suspicous. I am not sure though that this is the right path forward to modify everything here to just not expect a returned errors.

hajnalmt:
I don't think I should fix what gemini suggests, or at least not the way gemini suggests it. We can move on with this PR without touching this I think...

/assign @lowang-bh

pkg/scheduler/cache/event_handlers.go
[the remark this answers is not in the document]
JesseStutler: I think current is enough, we can not make sure other contributors will return errors in `job.DeleteTaskInfo` and `node.RemoveTask` in the future

pkg/scheduler/cache/event_handlers.go
```diff
@@ -321,35 +322,40 @@ func (sc *SchedulerCache) updatePod(oldPod, newPod *v1.Pod) error {
...
 	}
 
 	if len(ti.NodeName) != 0 {
-		node := sc.Nodes[ti.NodeName]
-		if node != nil {
-			nodeErr = node.RemoveTask(ti)
+		// We don't need to delete tasks that are already terminated, because
+		// they will be deleted in an updateTask call and not being readded when
+		// the taskStatus change happens from Releasing to -> Failed/Succeeded.
+		if !isTerminated(ti.Status) {
```
JesseStutler, line 350: Do you actually meet problems here? When `addTask` is called, it's indeed only non-Terminated tasks that are counted, but the comments here don't seem quite right. `updateTask` is only called when there's an err and `resyncTask` occurs

pkg/scheduler/cache/event_handlers.go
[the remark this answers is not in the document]
hajnalmt: Yeah I think the same let's leave it this way.

pkg/scheduler/cache/event_handlers.go
hajnalmt: Oh you are right, the comment is a little bit misleading.
It's actually `UpdatePod` that calls `updatePod`  which calls `sc.deletePod` and `sc.deleteTask`  and after that `sc.addPod` that calls `sc.addTask` 

So it's not updateTask that will make the "actual" status change.
Shall I modify the comment?

Yes this emits a lot of Warnings because the objects are missing from the nodes cache and this is emitted every single time a pod is deleted on the cluster. For every single deleted pod since the cache contains everything...
I see this as an actual bug, because the node cache can contain a task in theory with the old task's key which we will delete here if this guard is not actually there.

JesseStutler: https://github.com/volcano-sh/volcano/blob/9b7472d1fe9ddfec0b06f3d32b26db0e3e2ad297/pkg/scheduler/cache/cache.go#L663-L687 The main thing is that if pods that are not scheduled by volcano but are already running on the node (may be scheduled by other schedulers) will also filtered and counted into the used resources, but these pods will also trigger delete pod, and then these pods actually have no podgroup and report an error in deleteTask. Of course, I think the error here may not be a good written because this is a normal phenomenon. Your changes are correct, but we still need to consider isTerminated. 

Yes you need to modify the comment

hajnalmt: Updated the comment thank you!

Yes this FilterFunc filters out the pods that are not managed by volcano _and_ not placed on a node. So every pod on every node is in the cache. The only exception is when multi-volcano-scheduler with MULTI_SCHEDULER_ENABLE is enabled and the scheduler pod is not responsible for the node in the consistent hash circle.

This means that this behavior affects Volcano-managed pods as well, since they are updated via the podInformer, including when they are deleted.

hajnalmt: I think I see where the confusion is. We have 2 issues here but I am solving this in one go since they are in the same function.

1. One is the null JobID issue that we are reporting an Wargning error in deleteTask for every not Volcano Job managed pod.
This is this error:
`W0924 14:17:43.237059    5860 event_handlers.go:371] Failed to delete task: errors:  1: task default/nginx2-67755588bf-pkfhp has null jobID`
That's solved by switching to an InfoF log line instead of an error here:
https://github.com/volcano-sh/volcano/pull/4650/files#diff-219ed9274af805a7072e0effefec0c8b741239b6c7c34de43af30dfe7e1ea99cR342 

2. The other is for every pod (which was placed on a node, or is managed by volcano) that we are throwing a Warning I think more than 3 times for each deleted pod on the cluster 
This is the error in this case:
`W0924 15:58:35.180938   13225 node_info.go:466] failed to find task <default/nginx-676b6c5bbc-892h4> on host <k3d-volcano-dev-agent-2>`
This is what this `!isTerminated(ti.Status) {` clause solves.

hajnalmt:
Rebased so all tests passed now ☺️

PR #4653, opened 2025-09-28
Correctly populates job.Status.Succeeded when a job gets completed

### LINKED ISSUE #4616: job.status.succeeded Is Not Correctly Populated.
### Description

job.status.succeeded is not correctly populated. Could be related to this part of the code, https://github.com/volcano-sh/volcano/blob/3c63218204b5578a2eb16cef0b3430b5410448c4/pkg/controllers/job/job_controller_actions.go#L163C1-L164C1, as the `classifyAndAddUpPodBaseOnPhase` will only be run against the `podsToKill` which does not contain the succeeded pods.

1. Submit job to be completed
2. Check the status.succeeded field.

This is what I got:
Image

I expected the status.succeeded should be 1.

v1.12

_No response_

kev1N916:
#### What type of PR is this?
/kind bug
Currently after a job is submitted and is completed the job.status.succeeded and job.status.failed fields do not get populated.

When the Job is in the Running Phase we are able to see the status of all the pods
image

However as show in #4616  when the pod is in the completed phase we were not able to see the status of any of the pods.

Now we will be able to see the status of the pods even in the Completed phase
image

Fixes #4616

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -163,6 +163,22 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 		classifyAndAddUpPodBaseOnPhase(pod, &pending, &running, &succeeded, &failed, &unknown)
 		calcPodStatus(pod, taskStatusCount)
 	}
+	// update the status for the pods which are not going to be killed
+	for _, pods := range jobInfo.Pods {
+		for _, pod := range pods {
+			_, isPodGoingToBeKilled := podsToKill[pod.Name]
+			if !isPodGoingToBeKilled {
+				if pod.DeletionTimestamp != nil {
+					klog.Infof("Pod <%s/%s> is terminating", pod.Namespace, pod.Name)
+					terminating++
+					continue
+				}
+
+				classifyAndAddUpPodBaseOnPhase(pod, &pending, &running, &succeeded, &failed, &unknown)
+				calcPodStatus(pod, taskStatusCount)
+			}
+		}
+	}
 
 	if len(errs) != 0 {
 		klog.Errorf("failed to kill pods for job %s/%s, with err %+v", job.Namespace, job.Name, errs)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -97,7 +97,7 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 	var total int
 
 	podsToKill := make(map[string]*v1.Pod)
-
+	podsToBeRetained := make(map[string]*v1.Pod)
```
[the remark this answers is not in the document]
kev1N916: This is true. I've changed the logic now.

kev1N916:
@hwdef Could you review this when you find the time?

pkg/controllers/job/job_controller_actions.go
```diff
@@ -163,6 +163,22 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 		classifyAndAddUpPodBaseOnPhase(pod, &pending, &running, &succeeded, &failed, &unknown)
```
JesseStutler, line 163: I think the problem is `classifyAndAddUpPodBaseOnPhase` and `calcPodStatus` shouldn't couple with the processing logic of `podsToKill`, we can do statistics in the front, such as here: https://github.com/volcano-sh/volcano/blob/3c63218204b5578a2eb16cef0b3430b5410448c4/pkg/controllers/job/job_controller_actions.go#L140-L142, while we don't need to do `classifyAndAddUpPodBaseOnPhase` and `calcPodStatus` in `podsToKill`. 

We can change to like:
```
if !retain {
    podsToKill[pod.Name] = pod
} else {
    classifyAndAddUpPodBaseOnPhase(pod, &pending, &running, &succeeded, &failed, &unknown)
    calcPodStatus(pod, taskStatusCount)
}
```
WDYT? Then succeeded and failed pod count can be retained

kev1N916: If we do it this way, we would not be updating the Pod status in case target!=nil. 
The classifyAndAddUpPodBaseOnPhase and calcPodStatus for the podsToKill is only done if we fail to kill the pod so I feel we still need to do it.

Issue #4645, opened 2025-09-26
Support multi-volcano-scheduler without using selectors in the helm chart (at least) too


hajnalmt:
### What is the problem you're trying to solve

Volcano has a feature:
https://github.com/volcano-sh/volcano/blob/master/docs/design/deploy-multi-volcano-schedulers-without-using-selector.md
But it's not supported in the helm chart yet. In this case we need to deploy the scheduler as a Statefulset not as a Deployment.
https://github.com/volcano-sh/volcano/blob/master/installer/helm/chart/volcano/templates/scheduler.yaml

I think SCHEDULER_NUM should be retrieved from the replicas value (but shall be overrideable too I think).

### Describe the solution you'd like

Ability to deploy the scheduler via a Statefulset with helm and use multi-volcano-scheduler feature.

### Additional context

The introduction happened in this PR:
https://github.com/volcano-sh/volcano/pull/1795

ChickenBenny:
Hi @pinkfloyd-eminem and maintainers,

I noticed that PR #4657 was opened to address this issue, but it seems to have been inactive for a while without being merged. Is this still being actively worked on?

If there's anything blocking the PR — such as review feedback or unresolved conflicts — I'd be happy to help move it forward. Please let me know!

hajnalmt:
Hello @ChickenBenny 
Honestly I would love that PR to be merged. It would make the feature explicitly exposable in the deployment.
The maintaners did not feel that this feature is mature enough so they are reluctant to merge the helm change. The main blocker is that the feature lacks proper pod failure and healthcheck handling, and upgrade scenarios shall be tested too. Until this is done I wouldn't open an other PR for this.

ChickenBenny:
Hi @hajnalmt,

Thank you for the context! I totally understand — it makes sense to wait until the underlying feature is more robust before exposing it in the Helm chart.

I'm looking to contribute to Volcano. Are there any open issues you'd recommend for a newcomer to the project? Happy to take on something that's genuinely useful to the community!

PR #4657, opened 2025-09-30
volcano scheduler support multi-schedulers deploy

pinkfloyd-eminem:
#### What type of PR is this?
/kind feature

This PR adds support for deploying multiple Volcano schedulers using Helm chart with StatefulSet, enabling the multi-scheduler feature without using selectors.

**Key Changes:**
- Added conditional logic in scheduler.yaml to support both Deployment (single scheduler) and StatefulSet (multi-scheduler) deployment modes
- Introduced `multi_scheduler_enable` configuration option to toggle between deployment types
**Why we need this:**
The single scheduler cannot satisfy high throughput requirements in some scenarios. Deploying multiple schedulers improves overall scheduling throughput without requiring manual node labeling or workload modifications.

Fixes #hajnalmt multi-scheduler helm support issue

**Testing Instructions:**
1. Single scheduler mode (default): `helm install volcano installer/helm/chart/volcano/`
2. Multi-scheduler mode: `helm install volcano installer/helm/chart/volcano/ --set custom.multi_scheduler_enable=true --set custom.scheduler_replicas=3`

**Backward Compatibility:**
- Default behavior remains unchanged (single scheduler via Deployment)
- All existing configurations continue to work without modification
- Multi-scheduler mode is opt-in via explicit configuration

**Implementation Details:**
- Uses consistent hashing algorithm for automatic job/node assignment
- StatefulSet ensures stable network identities for schedulers
- Headless Service enables direct pod-to-pod communication

```release-note
Added support for deploying multiple Volcano schedulers via Helm chart using StatefulSet. This enables high-throughput scheduling scenarios by running multiple scheduler instances with automatic load distribution. Use `--set custom.multi_scheduler_enable=true` to enable multi-scheduler mode.
 

#4645

### DIFF
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -8837,7 +8837,7 @@ spec:
             - --node-worker-threads=20
             - -v=3
             - 2>&1
-          env:
+          env: 
             - name: DEBUG_SOCKET_DIR
               value: /tmp/klog-socks
           imagePullPolicy: Always
--- a/installer/helm/chart/volcano/templates/scheduler.yaml
+++ b/installer/helm/chart/volcano/templates/scheduler.yaml
@@ -5,6 +5,7 @@
 {{ $scheduler_main_csc := or .Values.custom.scheduler_main_csc .Values.custom.default_csc }}
 {{ $scheduler_ns := or .Values.custom.scheduler_ns .Values.custom.default_ns }}
 {{ $scheduler_name := .Values.custom.scheduler_name }}
+{{ $multi_scheduler_enabled := .Values.custom.multi_scheduler_enabled | default false }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -139,9 +140,12 @@ roleRef:
   kind: ClusterRole
   name: {{ .Release.Name }}-scheduler
   apiGroup: rbac.authorization.k8s.io
-
 ---
+{{- if not $multi_scheduler_enabled }}
 kind: Deployment
+{{- else }}
+kind: StatefulSet
+{{- end }}
 apiVersion: apps/v1
 metadata:
   name: {{ .Release.Name }}-scheduler
@@ -232,6 +236,16 @@ spec:
             - -v={{.Values.custom.scheduler_log_level}}
             - 2>&1
           env:
+            {{- if $multi_scheduler_enabled }}
+            - name: MULTI_SCHEDULER_ENABLE
+              value: "{{ $multi_scheduler_enabled }}"
+            - name: SCHEDULER_NUM
+              value: "{{ .Values.custom.scheduler_replicas }}"
+            - name: SCHEDULER_POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            {{- end }} 
             - name: DEBUG_SOCKET_DIR
               value: /tmp/klog-socks
           imagePullPolicy: {{ .Values.basic.image_pull_policy }}
@@ -266,6 +280,9 @@ metadata:
     {{- toYaml .Values.custom.common_labels | nindent 4 }}
     {{- end }}
 spec:
+  {{- if $multi_scheduler_enabled }}
+  clusterIP: None
+  {{- end }}
   {{- if .Values.service.ipFamilyPolicy }}
   ipFamilyPolicy: {{ .Values.service.ipFamilyPolicy }}
   {{- end }}
--- a/installer/helm/chart/volcano/values.yaml
+++ b/installer/helm/chart/volcano/values.yaml
@@ -212,6 +212,10 @@ custom:
   scheduler_log_level: 3
   controller_log_level: 4
 
+  # Multi-scheduler configuration
+  # Enable multi-scheduler deployment mode (deploys as StatefulSet instead of Deployment)
+  multi_scheduler_enabled: false
+
 # Specify container security context for admission
 # For example:
 #

### CODE REVIEW

**Comment by pinkfloyd-eminem** [comment]
test:
single scheduler：helm install volcano installer/helm/chart/volcano/   ok
multi schedulers: helm install volcano installer/helm/chart/volcano/ --set custom.multi_scheduler_enable=true --set custom.scheduler_replicas=3 ok

**Comment by hajnalmt** [comment]
Hello @pinkfloyd-eminem,
Thank you for implementing my feature request! 😊

Can you edit the PR description for me please?
At the part:
Fixes #hajnalmt multi-scheduler helm support issue

This should be:
Fixes #4645

So my request is properly linked.

Please additionally:
   - I checked the code and SCHEDULER_POD_NUM shall be the same as replicas, otherwise the feature won't work. So we can use `custom.scheduler_replicas` there is no need for an overwrite, I was wrong.
   - Gemini is right that there is a lot of duplication, we don't need a separate helper, but It would be good to derive only at the points where it's needed. I mean for example:
   ```
   ...
   {{- if not $multi_scheduler_enabled }}
   kind: Deployment
   {{- else }}
   kind: Statefulset
   {{- end }}
   ...
   ```
   - I would rename the value to `multi_scheduler_enabled` from `multi_scheduler_enable` too to use adjective grammer here and comply with the standards.
   - If it's okay can you start with the Deployment, in a not branch too, as that will be the default so it shall come first logically as in my example above.
   - The service differs only in the `ClusterIP: None` part too it's enough to derive only at that point too I think.
 
 Thank you once more for implementing this 👍 Please sign the DCO too.

**Comment by pinkfloyd-eminem** [comment]
> Hello @pinkfloyd-eminem, Thank you for implementing my feature request! 😊
> 
> Can you edit the PR description for me please? At the part: **Which issue(s) this PR fixes:** Fixes #hajnalmt multi-scheduler helm support issue
> 
> This should be: **Which issue(s) this PR fixes:** Fixes #4645
> 
> So my request is properly linked.
> 
> Please additionally:
> 
> * I checked the code and SCHEDULER_POD_NUM shall be the same as replicas, otherwise the feature won't work. So we can use `custom.scheduler_replicas` there is no need for an overwrite, I was wrong.
> * Gemini is right that there is a lot of duplication, we don't need a separate helper, but It would be good to derive only at the points where it's needed. I mean for example:
> 
> ```
> ...
> {{- if not $multi_scheduler_enabled }}
> kind: Deployment
> {{- else }}
> kind: Statefulset
> {{- end }}
> ...
> ```
> 
> * I would rename the value to `multi_scheduler_enabled` from `multi_scheduler_enable` too to use adjective grammer here and comply with the standards.
> * If it's okay can you start with the Deployment, in a not branch too, as that will be the default so it shall come first logically as in my example above.
> * The service differs only in the `ClusterIP: None` part too it's enough to derive only at that point too I think.
> 
> Thank you once more for implementing this 👍 Please sign the DCO too.

Hi, @hajnalmt
Thank you for your feedback and suggestions! 😊 I appreciate you taking the time to review the code.

I’ve made the following updates based on your comments:
1. Reduced code duplication​ by using conditional derivation only where needed
2. Renamed the value​ from multi_scheduler_enable to multi_scheduler_enabled for better grammar and compliance with standards.
3. ​Adjusted the logical order​ to prioritize the Deployment branch (default case) as you suggested.
4. ​Simplified the Service template​ to conditionally set ClusterIP: None only for the multi-scheduler case.

Verified through testing​ - I've successfully tested both the non-multi-replica and multi-replica Helm deployment scenarios to ensure everything works as expected.
​Signed DCO​ - All commits now include the required Signed-off-by line.

Thanks again for your guidance!

**Comment by hajnalmt** [comment]
Thank you @pinkfloyd-eminem for doing the changes!
This looks really good, I have one minor request left, since there is no purpose in defining the SCHEDULER_NUM env var for the Deployment, I think we shall move that part after the if branch too like:

```
            {{- if $multi_scheduler_enabled }}
            - name: MULTI_SCHEDULER_ENABLE
              value: "{{ $multi_scheduler_enabled }}"
            - name: SCHEDULER_NUM
              value: "{{ .Values.custom.scheduler_replicas }}"
            - name: SCHEDULER_POD_NAME
              valueFrom:
                fieldRef:
                  fieldPath: metadata.name
            {{- end }}
```

**Comment by pinkfloyd-eminem** [comment]
> Thank you @pinkfloyd-eminem for doing the changes! This looks really good, I have one minor request left, since there is no purpose in defining the SCHEDULER_NUM env var for the Deployment, I think we shall move that part after the if branch too like:
> 
> ```
>             {{- if $multi_scheduler_enabled }}
>             - name: MULTI_SCHEDULER_ENABLE
>               value: "{{ $multi_scheduler_enabled }}"
>             - name: SCHEDULER_NUM
>               value: "{{ .Values.custom.scheduler_replicas }}"
>             - name: SCHEDULER_POD_NAME
>               valueFrom:
>                 fieldRef:
>                   fieldPath: metadata.name
>             {{- end }}
> ```

Thank you for your suggestion. I have adopted your suggestion and submitted the edit.

**Comment by hajnalmt** [comment]
Thank you! Perfect.

/cc @Monokaix 
/cc @JesseStutler

**Comment by JesseStutler** [comment]
Thanks for your contribution, BTW, did you test @pinkfloyd-eminem  multi-scheduler schedule pods properly? I think maybe in the future we can add an e2e test for multi-scheduler mode

**Comment by JesseStutler** [comment]
@pinkfloyd-eminem Could you fix the CIs? You should run `make update-development-yaml` to generate the new yaml

**Comment by JesseStutler** [comment]
Ray plugin may have a flaky test, you could submit your PR after you fix the code verify and trigger CI again to check whether if the E2E sequence job can pass

**Comment by pinkfloyd-eminem** [comment]
> Ray plugin may have a flaky test, you could submit your PR after you fix the code verify and trigger CI again to check whether if the E2E sequence job can pass

Thanks for the reminder, I will check here

**Comment by pinkfloyd-eminem** [comment]
> 感谢您的贡献，顺便问一下，您测试过吗？@pinkfloyd-eminem 多调度器能正确调度 Pod 吗？我想未来我们可以为多调度器模式添加端到端测试。

@JesseStutler  I tested simple vcjob in multi-scheduler mode and it was successfully scheduled and executed to completion

**Comment by pinkfloyd-eminem** [comment]
> Ray plugin may have a flaky test, you could submit your PR after you fix the code verify and trigger CI again to check whether if the E2E sequence job can pass

@JesseStutler I used the "go test -v -ginkgo.focus="Ray Plugin E2E Test"" command to test the Ray plugin and found that the image with DefaultRayImage = "bitnami/ray:2.49.0" could not be pulled down. However, it can be successfully run by changing it to rayproject/ray:2.49.0. Should I consider changing the default image of Ray?

**Comment by Wonki4** [comment]
@pinkfloyd-eminem 

Yes, you can change the default image (both of them worked, bitnami/ray:2.49.0 was the smallest image in size).

The all of bitnami/ray images are removed and not available now (most images of bitnami are not available). rayproject/ray:2.49.0 is another option for us.

image



@JesseStutler 
If I can upload a image into volcano repository in docker repository, I build a private test image for e2e test.

**Comment by hajnalmt** [comment]
@pinkfloyd-eminem The CI issue shall be handled in a separate PR and I think Wonki4 handles it in
https://github.com/volcano-sh/volcano/pull/4670
So let's wait until it get's into the repository and retrigger the test with the [/]retest afterward.
You can then rebase and update the development.yaml with `make update-development-yaml` after that ☺️

**Comment by pinkfloyd-eminem** [comment]
> @pinkfloyd-eminem The CI issue shall be handled in a separate PR and I think Wonki4 handles it in #4670 So let's wait until it get's into the repository and retrigger the test with the [/]retest afterward. You can then rebase and update the development.yaml with `make update-development-yaml` after that ☺️

Thank you, I will wait for the changes，rebase and submit again.

**Comment by JesseStutler** [comment]
@pinkfloyd-eminem #4670 has already merged, please rebase the latest code and fix the code verify ci, thanks

**Comment by pinkfloyd-eminem** [comment]
> @pinkfloyd-eminem #4670 has already merged, please rebase the latest code and fix the code verify ci, thanks

Hi, I rebased the latest master branch and passed CI

File: installer/volcano-development.yaml [hunk 1]
```diff
@@ -8837,7 +8837,7 @@ spec:
             - --node-worker-threads=20
             - -v=3
             - 2>&1
-          env:
+          env:
```
JesseStutler, line 8840: @pinkfloyd-eminem There's a space here by mistake.

hwdef:
What is the necessity of using StatefulSet?

hzxuzhonghu:
@hwdef good comment

I feel current multischedulers is very premature. @pinkfloyd-eminem Do you need this ferature? From my perspective, we need to improve its reliability by allowing schedulres dynamically die or run without influencing pod scheduling latency.

JesseStutler:
> What is the necessity of using StatefulSet?

Looks like current implementation of multi-scheduler depends on the statefulset-generated stable pod index: https://github.com/volcano-sh/volcano/blob/74976dee5ac08a38ad0b95a81ef88c573bcb3b9e/pkg/scheduler/cache/util.go#L106. We can see that the `SCHEDULER_POD_NAME` env is got from metadata.name, and the hash key in each process must be the same.

JesseStutler:
> @hwdef good comment
> 
> I feel current multischedulers is very premature. @pinkfloyd-eminem Do you need this ferature? From my perspective, we need to improve its reliability by allowing schedulres dynamically die or run without influencing pod scheduling latency.

The requirement is raised from #4645.
/cc @hajnalmt

hwdef:
> I feel current multischedulers is very premature. 

I agree with this. 
I believe that increasing scheduling throughput by deploying multiple schedulers is a systematic and large-scale engineering effort, not just a matter of scaling replicas. We need to consider factors such as node partitioning, scheduling conflicts, and so on.

hajnalmt:
Thanks for the input @hwdef and @hzxuzhonghu!

You're absolutely right—this feature is still premature. It lacks proper documentation, doesn’t handle pod failures reliably, and currently has no test coverage. Metrics might also be misleading in its current state (if not straighout unusable), not to mention how it will interact with other features like the networkTopologyAware scheduling, and plugins like the nodegroup plugin.

That said, I still believe merging this PR is a good step forward, even if the feature remains experimental for now.

At the moment, I’m using node selectors to guide schedulers to specific nodes, as described here:
https://github.com/volcano-sh/volcano/blob/master/docs/design/node-selector.md
It’s a straightforward approach, but I have concerns. The scheduler cache still contains the entire cluster (as it should), so memory usage scales with each deployment. This setup also increases operational overhead. It’s worth noting that this feature isn’t documented either on the official website—I only found it by chance in the design docs. So I am searching and experimenting with alternatives.
We also have other experimental features exposed via the Helm chart:
https://github.com/volcano-sh/volcano/blob/master/docs/design/multi-scheduler.md
I doubt many users are actively using these, but it’s good they exist for those who explore deeper. Interestingly, the feature in this PR is the only one I couldn’t enable directly via Helm, which is why I raised this request.

> I believe that increasing scheduling throughput by deploying multiple schedulers is a systematic and large-scale engineering effort, not just a matter of scaling replicas

If it _can_ be a matter of scaling replicas, why shouldn’t it be? From the minimal testing I tried, both node partitioning and scheduling conflict handling seem to work reasonably well in the current implementation.

I think we should discuss which features in Volcano are considered experimental and find a way to mark them clearly. What do you think @hzxuzhonghu @JesseStutler @hwdef ?

hwdef:
@hajnalmt 

You make a very good point, and I also believe we should take further steps in the multi-scheduler direction.

However, my concern is that the changes in this PR fundamentally alter Volcano's deployment model—shifting from a stateless to a stateful architecture. This is a significant change, and I haven’t seen corresponding reliability validation, such as e2e tests.

As Volcano is a widely used open-source project, we need to ensure a smooth experience for our users. Whether during upgrades or when switching between single-scheduler and multi-scheduler modes, Volcano should continue to function reliably. We may need to reinforce both the code and documentation to clearly communicate reliability expectations to users.

In my view, the current PR feels a bit too minimal—it only modifies a few lines in the Helm charts but changes Volcano from a Deployment to a StatefulSet. That seems somewhat drastic.

I believe it works correctly in your environment, but I'm not sure if it can perform reliably in other edge-case scenarios.

Issue #4530, opened 2025-08-11
Overcommit factor for the root queue, to enable capacity plugin's reclaim


hajnalmt:
### What is the problem you're trying to solve

I want to achive a hierachical queue structure where the root queue has several sub-queues for each projects (one project - one subqueue, let's call it project-root queue) and each project shall have two leaf queues in this subqueue which in they will be able to put their jobs in based on priorities.
I want to do this by setting the deserved part of the project-root queues to be equal with the cluster resources, and with the help of the capability value. I can limit how many jobs they will able to start, and the projects can reclaim resources from the others. I want to set this capability to be around 1,5 times larger than the deserved value.

For the example let's  the project root queue:
```
kubectl get queues root -o yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: root
...
spec:
  deserved:
    ...
    nvidia-gpu-a100: "4"
    nvidia.com/gpu: "4"
...
```
And the 2 subQueues for the project1 and project2:
```
kubectl get queues  project1-root  project2-root -o yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: project1-root
...
spec:
  capability:
     nvidia-gpu-a100: "3"
  deserved:
     nvidia-gpu-a100: "2"
---
apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: project2-root
...
spec:
  capability:
     nvidia-gpu-a100: "3"
  deserved:
     nvidia-gpu-a100: "2"
```

After I enable the hierarchy:
```
kubectl get configmap -n volcano-system volcano-scheduler-configmap -o yaml
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, backfill, preempt, reclaim"
    tiers:
    - plugins:
      - name: priority
        enablePreemptable: true
        enableReclaimable: true
      - name: gang
        enablePreemptable: false  # This allows preemption without gang constraints
      - name: conformance
    - plugins:
      - name: predicates
      - name: capacity
        enableHierarchy: true
      - name: nodeorder
```
The capacity plugin won't work on the cluster because of this error appearing instantly:
```
E0811 07:52:58.921507       1 capacity.go:510] Failed to check queue's hierarchical structure, error: deserved resources of queue <root> are less than the sum of its child queues' deserved resources
```
[...]

PR #4662, opened 2025-10-02
Add overwriteRootCapability argument to capacity plugin

### LINKED ISSUE #3910: PodGroup stuck in Pending Phase
### Description

We are currently running the Kubeflow spark-operator on Linode LKE kubernetes with auto-scaling enabled.

I have noticed that when trying to trigger a large auto-scaling (i.e. trying to trigger to use all possible nodes in a node-pool) Volcano pods are stuck in Pending phase and PodGroup is also stuck on pending. 
If I reduce the RAM and CPU request to half, the scaling up will then be triggered and the PodGroup is successful. Not sure why or how the calculation is done in the volcano side.
My volcano config is: 
```
actions: "enqueue, allocate,preempt, backfill"
tiers:
- plugins:
  - name: priority
  - name: conformance
- plugins:
  - name: overcommit
    arguments:  
    overcommit-factor: 15.0
  - name: drf
    enablePreemptable: false
  - name: predicates
  - name: capacity
  - name: nodeorder
  - name: binpack
```
My queue: 
```
kind: Queue
metadata:
  name: myqueue
spec:
  reclaimable: true
  weight: 1
  capability:
    cpu: "200"
    memory: "1200G"
status:
  state: Open
```
My K8s cluster has auto-scaling enabled in the node-pool with Min 1 and Max 10 nodes (16C,300GB Ram) 

My spark job conf: 
```
apiVersion: "sparkoperator.k8s.io/v1beta2"
kind: SparkApplication
metadata:
  name: spark-sleep-2
  namespace: spark
spec:
  type: Python
  mode: cluster
  image: "redacted-private-repo/spark:k8s-3.5.1"
  imagePullSecrets:
    - "regcred"
  imagePullPolicy: Always
  mainApplicationFile: "local:///opt/spark/work-dir/sleep_forever.py"
  sparkVersion: "3.5.1"
  batchScheduler: volcano
  batchSchedulerOptions:
    priorityClassName: urgent
    queue: myqueue
  restartPolicy:
    type: Never
  driver:
    cores: 2
    memory: "8G"
    labels:
      version: 3.5.1
    serviceAccount: spark
    affinity:
      nodeAffinity:
          nodeSelectorTerms:
            - matchExpressions:
                - key: "node.kubernetes.io/instance-type"
                  operator: In
                  values:
                    - "g6-dedicated-32"
  executor:
    cores: 15
    instances: 8
    memory: "200G"
    labels:
      version: 3.5.1
    affinity:
      nodeAffinity:
          nodeSelectorTerms:
            - matchExpressions:
                - key: "node.kubernetes.io/instance-type"
                  operator: In
                  values:
                    - "g7-highmem-16"
```
I have tried various volcano-scheduler.conf options but the same error persists. 
PodGroup reports:
`1/1 tasks in gang unschedulable: pod group is not ready, 1 Pending, 1 minAvailable; Pending: 1 Unschedulable`
`queue resource quota insufficient`
```
kind: PodGroup
metadata:
  name: spark-spark-sleep-2-pg
  namespace: spark
.
.
.
.
status:
  phase: Pending
spec:
  minMember: 1
  minResources:
    cpu: '122'
    memory: 1608G
  priorityClassName: urgent
  queue: myqueue
```

Is anyone aware on how to fix this issue? I removed gang scheduling plugin as per https://github.com/volcano-sh/volcano/issues/2558 but that did not work. 

PodGroup stuck on Pending

1.10

k8s 1.29

mrendi29:
#### What type of PR is this?
Feature

Problem: 
Currently, the root queue's resource limits are always set to the actual cluster resources, regardless of any user-configured capability values. This prevents administrators from setting virtual resource limits that differ from the physical cluster capacity. This helps in cases where a user has cluster-autoscaled enabled and the "virtual" resources can be higher than the "real" resources. 

Fixes #3910 

None
```release-note
Add overwriteRootCapability argument to capacity plugin to allow using user-configured root queue capability instead of actual cluster resources
```

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -51,8 +51,8 @@ type capacityPlugin struct {
 
 	queueOpts map[api.QueueID]*queueAttr
 	// Arguments given for the plugin
-	pluginArguments         framework.Arguments
-	overwriteRootCapability bool
+	pluginArguments                  framework.Arguments
+	overwriteRootQueueRealCapability bool
 }
 
 type queueAttr struct {
@@ -78,14 +78,14 @@ type queueAttr struct {
 // New return capacityPlugin action
 func New(arguments framework.Arguments) framework.Plugin {
 	cp := &capacityPlugin{
-		totalResource:           api.EmptyResource(),
-		totalGuarantee:          api.EmptyResource(),
-		queueOpts:               map[api.QueueID]*queueAttr{},
-		pluginArguments:         arguments,
-		overwriteRootCapability: false,
+		totalResource:                    api.EmptyResource(),
+		totalGuarantee:                   api.EmptyResource(),
+		queueOpts:                        map[api.QueueID]*queueAttr{},
+		pluginArguments:                  arguments,
+		overwriteRootQueueRealCapability: false,
 	}
 
-	arguments.GetBool(&cp.overwriteRootCapability, "overwriteRootCapability")
+	arguments.GetBool(&cp.overwriteRootQueueRealCapability, "overwriteRootQueueRealCapability")
 
 	return cp
 }
@@ -611,15 +611,16 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 	}
 
 	// https://github.com/volcano-sh/volcano/issues/3910
-	// If overwriteRootCapability is true, we need to set realCapability to the user-configured capability.
+	// If overwriteRootQueueRealCapability is true, we need to set realCapability to the user-configured capability.
 	// If the user does not edit the root queue's capability, it will be set to total resource above.
-	if cp.overwriteRootCapability {
-		rootQueueAttr.realCapability = rootQueueAttr.capability
-		klog.V(4).Infof("Using user-configured root queue capability <%v> instead of cluster resources <%v> due to overwriteRootCapability=true",
+	if cp.overwriteRootQueueRealCapability {
+		// Use user-configured capability instead of forcing cluster resources
+		rootQueueAttr.realCapability = rootQueueAttr.capability.Clone()
+		klog.V(4).Infof("Using user-configured root queue capability <%v> instead of cluster resources <%v> due to overwriteRootQueueRealCapability=true",
 			rootQueueAttr.capability, cp.totalResource)
 	} else {
-		// Use actual cluster resources
-		rootQueueAttr.realCapability = cp.totalResource
+		// Default behavior: use actual cluster resources
+		rootQueueAttr.realCapability = cp.totalResource.Clone()
 	}
 
 	// Check the hierarchical structure of queues
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -41,6 +41,8 @@ const (
 	// Using the name of the plugin will likely help us avoid collisions with other plugins.
 	capacityStateKey = PluginName
 	rootQueueID      = "root"
+
+	overwriteRootQueueRealCapability = "overwriteRootQueueRealCapability"
 )
 
 type capacityPlugin struct {
@@ -85,7 +87,7 @@ func New(arguments framework.Arguments) framework.Plugin {
 		overwriteRootQueueRealCapability: false,
 	}
 
-	arguments.GetBool(&cp.overwriteRootQueueRealCapability, "overwriteRootQueueRealCapability")
+	arguments.GetBool(&cp.overwriteRootQueueRealCapability, overwriteRootQueueRealCapability)
 
 	return cp
 }
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -51,7 +51,8 @@ type capacityPlugin struct {
 
 	queueOpts map[api.QueueID]*queueAttr
 	// Arguments given for the plugin
-	pluginArguments framework.Arguments
+	pluginArguments         framework.Arguments
+	overwriteRootCapability bool
 }
 
 type queueAttr struct {
@@ -76,12 +77,17 @@ type queueAttr struct {
 
 // New return capacityPlugin action
 func New(arguments framework.Arguments) framework.Plugin {
-	return &capacityPlugin{
-		totalResource:   api.EmptyResource(),
-		totalGuarantee:  api.EmptyResource(),
-		queueOpts:       map[api.QueueID]*queueAttr{},
-		pluginArguments: arguments,
+	cp := &capacityPlugin{
+		totalResource:           api.EmptyResource(),
+		totalGuarantee:          api.EmptyResource(),
+		queueOpts:               map[api.QueueID]*queueAttr{},
+		pluginArguments:         arguments,
+		overwriteRootCapability: false,
 	}
+
+	arguments.GetBool(&cp.overwriteRootCapability, "overwriteRootCapability")
+
+	return cp
 }
 
 func (cp *capacityPlugin) Name() string {
@@ -603,7 +609,19 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 	if rootQueueAttr.deserved.IsEmpty() {
 		rootQueueAttr.deserved = cp.totalResource
 	}
-	rootQueueAttr.realCapability = cp.totalResource
+
+	// https://github.com/volcano-sh/volcano/issues/3910
+	// If overwriteRootCapability is true, we need to set realCapability to the user-configured capability.
+	// If the user does not edit the root queue's capability, it will be set to total resource above.
+	if cp.overwriteRootCapability {
+		rootQueueAttr.realCapability = rootQueueAttr.capability
+		klog.V(4).Infof("Using user-configured root queue capability <%v> instead of cluster resources <%v> due to overwriteRootCapability=true",
+			rootQueueAttr.capability, cp.totalResource)
+	} else {
+		// Use actual cluster resources
+		rootQueueAttr.realCapability = cp.totalResource
+	}
+
 	// Check the hierarchical structure of queues
 	err := cp.checkHierarchicalQueue(rootQueueAttr)
 	if err != nil {
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -612,16 +612,14 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 		rootQueueAttr.deserved = cp.totalResource
 	}
 
-	// https://github.com/volcano-sh/volcano/issues/3910
-	// If overwriteRootQueueRealCapability is true, we need to set realCapability to the user-configured capability.
-	// If the user does not edit the root queue's capability, it will be set to total resource above.
+	// If overwriteRootQueueRealCapability is true and the user has explicitly configured the root queue's
+	// capability, use that value. Otherwise, default to the total cluster resources as the root queue's realCapability.
+	// This allows cluster-autoscalers to work correctly by respecting user overrides.
 	if cp.overwriteRootQueueRealCapability {
-		// Use user-configured capability instead of forcing cluster resources
 		rootQueueAttr.realCapability = rootQueueAttr.capability.Clone()
-		klog.V(4).Infof("Using user-configured root queue capability <%v> instead of cluster resources <%v> due to overwriteRootQueueRealCapability=true",
-			rootQueueAttr.capability, cp.totalResource)
+		klog.V(4).Infof("Root queue realCapability: <%v> (user override); cluster resources: <%v>",
+			rootQueueAttr.realCapability, cp.totalResource)
 	} else {
-		// Default behavior: use actual cluster resources
 		rootQueueAttr.realCapability = cp.totalResource.Clone()
 	}
 
--- a/pkg/scheduler/plugins/capacity/capacity_test.go
+++ b/pkg/scheduler/plugins/capacity/capacity_test.go
@@ -262,6 +262,7 @@ func Test_capacityPlugin_OnSessionOpenWithoutHierarchy(t *testing.T) {
 			},
 		},
 	}
+
 	for i, test := range tests {
 		t.Run(test.Name, func(t *testing.T) {
 			test.RegisterSession(tiers, nil)
@@ -328,7 +329,7 @@ func TestEnqueueAndAllocatable(t *testing.T) {
 	}
 	tests := []uthelper.TestCommonStruct{
 		{
-			Name:           "case0: memory exceed derserved, job only request cpu can be enqueued and allocated",
+			Name:           "case0: memory exceeds deserved, job only requests cpu can be enqueued and allocated",
 			Plugins:        plugins,
 			Pods:           []*corev1.Pod{p1, p2, p3},
 			Nodes:          []*corev1.Node{n1, n2},
@@ -338,7 +339,7 @@ func TestEnqueueAndAllocatable(t *testing.T) {
 			ExpectBindMap:  map[string]string{"ns1/pod3": "n1"},
 		},
 		{
-			Name:           "case1: cpu exceed derserved, job only request memory can be enqueued and allocated",
+			Name:           "case1: cpu exceeds deserved, job only requests memory can be enqueued and allocated",
 			Plugins:        plugins,
 			Pods:           []*corev1.Pod{p1, p2, p4},
 			Nodes:          []*corev1.Node{n1, n2},
@@ -358,7 +359,7 @@ func TestEnqueueAndAllocatable(t *testing.T) {
 			ExpectBindMap:  map[string]string{},
 		},
 		{
-			Name:           "case4: queue with non-open state, can not enqueue",
+			Name:           "case4: queue with non-open state can not enqueue",
 			Plugins:        plugins,
 			Pods:           []*corev1.Pod{p6},
 			Nodes:          []*corev1.Node{n1, n2},
@@ -534,6 +535,32 @@ func Test_capacityPlugin_OnSessionOpenWithHierarchy(t *testing.T) {
 	p17 := util.BuildPod("ns1", "p17", "n3", corev1.PodRunning, api.BuildResourceList("1", "1Gi", []api.ScalarResource{{Name: "nvidia.com/a100", Value: "2"}, {Name: "rdma/hca", Value: "1"}}...), "pg17", make(map[string]string), map[string]string{})
 	p18 := util.BuildPod("ns1", "p18", "n3", corev1.PodRunning, api.BuildResourceList("1", "1Gi", []api.ScalarResource{{Name: "nvidia.com/a100", Value: "2"}, {Name: "rdma/hca", Value: "1"}}...), "pg18", map[string]string{schedulingv1beta1.PodPreemptable: "false"}, map[string]string{})
 
+	// root queue realCapability tests
+	arguments := framework.Arguments{
+		"overwriteRootQueueRealCapability": true,
+	}
+	pluginsWithOverwriteRootQueueRealCapability := map[string]framework.PluginBuilder{
+		PluginName: func(args framework.Arguments) framework.Plugin {
+			return New(arguments)
+		},
+		predicates.PluginName: predicates.New,
+		gang.PluginName:       gang.New,
+	}
+	// expect to enqueue pods over cluster limits when override is enabled, otherwise not
+	// nodes
+	n1_realCapabilityTest := util.BuildNode("n1", api.BuildResourceList("8", "8Gi", []api.ScalarResource{{Name: "pods", Value: "11"}}...), map[string]string{})
+	// queues
+	root1_realCapabilityTest := buildQueueWithParents("root", "", nil, api.BuildResourceList("16", "16Gi"))
+	queue1_realCapabilityTest := buildQueueWithParents("q1", "root", nil, nil)
+
+	// pod uses the whole node
+	p1_realCapabilityTest := util.BuildPod("ns1", "p1", "n1", corev1.PodRunning, api.BuildResourceList("8", "8Gi"), "pg1", make(map[string]string), map[string]string{})
+	p2_realCapabilityTest := util.BuildPod("ns1", "p2", "", corev1.PodPending, api.BuildResourceList("8", "8Gi"), "pg2", make(map[string]string), map[string]string{})
+
+	// podgroup
+	pg1_realCapabilityTest := util.BuildPodGroupWithMinResources("pg1", "ns1", "q1", 1, nil, api.BuildResourceList("8", "8Gi"), schedulingv1beta1.PodGroupRunning)
+	pg2_realCapabilityTest := util.BuildPodGroupWithMinResources("pg2", "ns1", "q1", 1, nil, api.BuildResourceList("8", "8Gi"), schedulingv1beta1.PodGroupPending)
+
 	tests := []uthelper.TestCommonStruct{
 		{
 			Name:      "case0: Pod allocatable when queue is leaf queue",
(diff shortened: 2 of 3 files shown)


mrendi29:
cc: @JesseStutler @Monokaix

hajnalmt:
This is not true. You can set it to any value. We only set it to the cluster capacity when it's empty.
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/capacity/capacity.go#L590
Aren't you using an old version or something?

I am far from happy with the current queue design, and all these hacks around the root queue, but I feel that this is still a step away from the right path.
I don't have time to implement it currently, but we should use overcommitFactors instead to align the configurations across the plugins...
https://github.com/volcano-sh/volcano/issues/4530

hajnalmt:
Oh sry, I understand it now. We want to overwrite the realCapability with the root Queue's capability, maybe renaming it to overwriteRootQueueRealCapability would better reflect the intention and logic behind the usage of the argument.

mrendi29:
@hajnalmt thanks for the review. I'll update the variable to better reflect the intention

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -593,7 +599,19 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
+	// If overwriteRootCapability is true, we need to set realCapability to the user-configured capability.
+	// If the user does not edit the root queue's capability, it will be set to total resource above.
+	if cp.overwriteRootCapability {
+		rootQueueAttr.realCapability = rootQueueAttr.capability
+		klog.V(4).Infof("Using user-configured root queue capability <%v> instead of cluster resources <%v> due to overwriteRootCapability=true",
+			rootQueueAttr.capability, cp.totalResource)
+	} else {
+		// Use actual cluster resources
+		rootQueueAttr.realCapability = cp.totalResource
+	}
```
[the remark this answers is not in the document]
hajnalmt: I think these Cloned() methods shall be added, I understand that it wasn't added in the earlier implementation improperly.
Also when you retrieve `arguments.GetBool(&cp.overwriteRootCapability, "overwriteRootCapability")`
I think the key shall be saved as a const at the beginning of the plugin (Line 37), so we know the configurable arguments of the plugin by glimpsing on the file too. The log message could be better too indicating and distinguishing that the realCapability has been overriden by the capability, and not the root capability has been modified.

Let's see what Jesse and Monokaix thinks about these :) they have the authority to properly review your PR. Also don't forget sign the DCO.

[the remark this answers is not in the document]
JesseStutler: @mrendi29  Yes, set the key as a const is better. Agree

JesseStutler:
@mrendi29 Please sign your commit by using `git commit  -s`

JesseStutler:
> Special notes for your reviewer: I have not yet tested this in our production clusters, I plan to test it in the following week.

@mrendi29 Have you already tested it in production clusters and does it work properly?

mrendi29:
> Have you already tested it in production clusters and does it work properly? 

Not yet unfortunately, had to fight some fires at work. Planning to do so this week. Will also address the comments on the PR and include some tests.

hajnalmt:
It looks like this would solve other issues too:
[https://github.com/volcano-sh/volcano/issues/4684](https://github.com/volcano-sh/volcano/issues/4684)
Do you need any help with this PR? If you don't have enough time I can polish this quite fast for you (if you give me temporarily permissions to your repo I will update your branch).

mrendi29:
> It looks like this would solve other issues too:
> 
> [https://github.com/volcano-sh/volcano/issues/4684](https://github.com/volcano-sh/volcano/issues/4684)
> 
> Do you need any help with this PR? If you don't have enough time I can polish this quite fast for you (if you give me temporarily permissions to your repo I will update your branch).


I'll push my remaining changes in a bit and I'll give you access to the repo and we can pair on this together. Thanks for your help!

mrendi29:
@hajnalmt  i've added you on the repo, you should have access now. 

I'll work on adding a testcase as well. if you can get to that before I do feel free to make the change.

mrendi29:
@JesseStutler @hajnalmt  can you take a second look at this PR? I believe I have addressed all comments.  Let me know what you think. 

FYI, the test case is AI generated after a couple of prompts :)

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -654,6 +654,205 @@ func Test_capacityPlugin_OnSessionOpenWithHierarchy(t *testing.T) {
 	}
 }
 
+func TestOverwriteRootQueueRealCapability(t *testing.T) {
```
JesseStutler: I think we can just add a simple case in `Test_capacityPlugin_OnSessionOpenWithHierarchy` is enough, there's no need to add a new `Test_capacityPlugin_OnSessionOpenWithHierarchy` test func and add a lot of new cases :)

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -593,7 +601,20 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
+	// If overwriteRootQueueRealCapability is true, we need to set realCapability to the user-configured capability.
+	// If the user does not edit the root queue's capability, it will be set to total resource above.
+	if cp.overwriteRootQueueRealCapability {
+		// Use user-configured capability instead of forcing cluster resources
+		rootQueueAttr.realCapability = rootQueueAttr.capability.Clone()
+		klog.V(4).Infof("Using user-configured root queue capability <%v> instead of cluster resources <%v> due to overwriteRootQueueRealCapability=true",
+			rootQueueAttr.capability, cp.totalResource)
+	} else {
+		// Default behavior: use actual cluster resources
+		rootQueueAttr.realCapability = cp.totalResource.Clone()
```
wangyang0616, line 623: Wouldn't it be more appropriate to set the default value to infinity here?

The root queue is automatically created by Volcano, and users should not be aware of it. Therefore, when only the capacity plugin is enabled, users can submit an unlimited number of jobs, and the root queue does not perform any quota checks or restrictions. If users want to use the queue quota capability, they need to actively create queues for quota management.

The same logic applies to the default values for the Deserver and Guarantee of the root queue.

vzhou-p: Our team is running into the same issue described here: https://github.com/volcano-sh/volcano/pull/4680, which also relates to root capacity checks. I agree that it would make sense for the root capacity to be set to infinity by default, rather than requiring users to specify a value.

I do have a question, though. Are there any concerns with allowing infinite capacity on the root, especially given that the original design of Volcano seemed to care about overall cluster capacity? For example, there’s a capacity plugin that rejects jobs when the cluster is full and an overcommit plugin to allow extra jobs. It feels like the earlier designs didn’t favor an infinite root capacity. Would this change have any unintended impact on those mechanisms?

JesseStutler: @vzhou-p Looks like difference users have difference requirements, setting it as a infinite value is a good choice, but currrently child queues' capability will inherit from the parent queue if not set:
image
So which mean child queue also can submit infinite jobs, so I'm not sure that whether it's what users want, because difference users may have difference requirements

wangyang0616: I understand that the default value for the root queue may have different expectations depending on different scenarios:

**1. Default to infinite:** After enabling the capacity plugin, users do not need to create additional queues, and job scheduling behaves the same as when the capacity plugin is not enabled.
**2. Default to the actual resource capacity of the cluster:** After enabling the capacity plugin, the queue's quota management capability is automatically activated. Jobs exceeding the quota limit will not be scheduled.
**3. Users can specify a certain percentage increase based on the actual resource capacity of the cluster,** for example, 120% of the actual resources. This allows for some overcommitment of quotas based on user configuration.
**4. Users can specify the actual resource capacity and actively manage it along with the queue quotas.**

Can these four methods be exposed through arguments in the capability configuration? If no specific setting is provided, the infinite mode should be used by default. In scenarios where users do not create additional queues, the behavior should remain consistent with that when the capacity plugin is not enabled.

mrendi29: Setting the root queue as infinity sounds like a really interesting option. That means that I as a platform admin would not need to interact with the root queue directly and can just create another queue on top of root. 
(I recall that there was another PR somewhere that wanted to interact with the root queue capacity through the helm chart, this could eliminate that).

In the same time, having child queues submit infinite jobs seems really risky. 

I wonder though, if it makes sense for this functionality to be part of overwriteRootQueueRealCapability  in [L610](https://github.com/mrendi29/volcano/blob/f1f617d28db1d2fbac34949f8b3ed79c2e12e4ed/pkg/scheduler/plugins/capacity/capacity.go#L610) instead of default Volcano behavior. 

Depending on what the PMCs think I can then make the necessary adjustments.

JesseStutler: > In the same time, having child queues submit infinite jobs seems really risky.

I think the design of the Volcano queue is to allow Volcano users to act as admins, allowing them to specify the queue's capability value, not just submit jobs to the default queue. Therefore as a Volcano user, and I may use Volcano to build a platform, I may write a custom webhook to require users who want to create a queue must specify capability value. So I think set the root and default queue with infinite is aligned with the behavior when we don't open the capacity plugin, but whether we allow the child queue still inherit from the ancestor, we may need to further discuss whether it's reasonable

vzhou-p: Thanks all for the thoughtful discussion.
I recently raised [a feature request to support maxApplications on queues](https://github.com/volcano-sh/volcano/issues/4716). I think there’s an opportunity here to address the concerns around child queues inheriting infinite capacity by considering resource quota and maxApplications together.

By supporting both resource quota (for overall resource usage limits) and maxApplications (for controlling the number of jobs), we can give users more flexibility and additional control in two dimensions. This could help mitigate the risk of child queues having unlimited capacity, while still keeping the root queue default as “infinite” for ease of use. Users could opt-in to more restrictive behaviors as needed.

Perhaps we can look at combining this work or aligning the design to solve both requirements at once. Would love to hear thoughts on whether this approach could address the current concerns. If there’s consensus, I’m happy to help move this forward.

JesseStutler: @vzhou-p So is maxApplication part of queue's capability or it's a seperate field? Will it be a required value?

vzhou-p: I think maxApplications should be a separate field from the queue's capability (resource quota), rather than being part of the capability structure itself. This keeps the two concepts distinct.

* Capability (resource quota): Controls how much resources (CPU, memory, etc.) can be consumed
* maxApplications: Controls how many jobs/applications can run or be queued

I believe it should be optional, with a sensible default (likely infinite or unlimited). This aligns with the earlier discussion about keeping the root queue infinite by default. Users opt into restrictions as needed rather than being forced to configure everything upfront.

mrendi29:
I tried testing this in my k8s cluster today. I created a new image only for the scheduler and used that in my cluster. I did not change controller and admission image.
I created a root queue with 500 cores and 5000GB RAM (note my cluster has only <100 cores and 500GB RAM but has autoscaling enabled) 
I also created a child queue of root with 250 cores and 4500GB RAM. 

I tried submitting a spark job but got the following error from the spark operator logs: 
```
│ I1106 23:33:48.303393      10 event.go:285] Event(v1.ObjectReference{Kind:"SparkApplication", Namespace:"spark", Name:"spark-sleep-8", UID:"c3526aed-2566-42e1-a0a6-33a445cca51f", APIVersion:"sparkoperator.k8s.io/v1beta2", ResourceVersion:"221645563", FieldPath:""}): type: 'Normal' reason: 'SparkApplicationAdded' SparkApplication spark-sleep-8 was added, enqueuing it for submission                                                                                                                                │
│ E1106 23:33:48.319225      10 controller.go:676] failed to process batch scheduler BeforeSubmitSparkApplication with error failed to sync PodGroup with error: admission webhook "validatepodgroup.volcano.sh" denied the request: can only submit PodGroup to queue with state `Open`, queue `myqueue` status is ``. Abandon schedule pods via volcano     
```
Seems that the queue i created is not reporting status: 
```
┌─────────────────────────────────────────────────────────────────────────── Describe(-/myqueue) ────────────────────────────────────────────────────────────────────────────┐
│ Name:         myqueue                                                                                                                                                      │
│ Namespace:                                                                                                                                                                 │
│ Labels:       <none>                                                                                                                                                       │
│ Annotations:  <none>                                                                                                                                                       │
│ API Version:  scheduling.volcano.sh/v1beta1                                                                                                                                │
│ Kind:         Queue                                                                                                                                                        │
│ Metadata:                                                                                                                                                                  │
│   Creation Timestamp:  2025-11-06T23:30:58Z                                                                                                                                │
│   Generation:          1                                                                                                                                                   │
│   Resource Version:    221644752                                                                                                                                           │
│   UID:                 b9ad6d06-4ac8-4ca0-8e73-25964af8f0d6                                                                                                                │
│ Spec:                                                                                                                                                                      │
│   Guarantee:                                                                                                                                                               │
│     Resource:                                                                                                                                                              │
│       Cpu:      240                                                                                                                                                        │
│       Memory:   4500G                                                                                                                                                      │
│   Parent:       root                                                                                                                                                       │
│   Reclaimable:  true                                                                                                                                                       │
│   Weight:       1                                                                                                                                                          │
│ Events:         <none>       
```

Whereas root: 
```
│ Name:         root                                                                                                                                                         │
│ Namespace:                                                                                                                                                                 │
│ Labels:       <none>                                                                                                                                                       │
│ Annotations:  <none>                                                                                                                                                       │
│ API Version:  scheduling.volcano.sh/v1beta1                                                                                                                                │
│ Kind:         Queue                                                                                                                                                        │
│ Metadata:                                                                                                                                                                  │
│   Creation Timestamp:  2025-08-20T22:08:49Z                                                                                                                                │
│   Generation:          3                                                                                                                                                   │
│   Resource Version:    188581782                                                                                                                                           │
│   UID:                 de2d7990-f6dd-42c2-8e9d-524d6f07845b                                                                                                                │
│ Spec:                                                                                                                                                                      │
│   Capability:                                                                                                                                                              │
│     Cpu:        500                                                                                                                                                        │
│     Memory:     5000G                                                                                                                                                      │
│   Reclaimable:  false                                                                                                                                                      │
│   Weight:       1                                                                                                                                                          │
│ Status:                                                                                                                                                                    │
│   Allocated:                                                                                                                                                               │
│     Cpu:     0                                                                                                                                                             │
│     Memory:  0                                                                                                                                                             │
│   Reservation:                                                                                                                                                             │
│   State:  Open                                                                                                                                                             │
│ Events:   <none>                                                                                                                                                           │
│                    
```
Will do some more digging

mrendi29:
If i submit the job to the root queue directly: 
Scheduler logs:
```
│ I1106 23:53:28.184662       1 util.go:101] schedulerPodName  is responsible to PodGroup spark/spark-spark-sleep-8-pg                                                       │
│ I1106 23:53:28.184747       1 event_handlers.go:835] Add PodGroup(spark-spark-sleep-8-pg) into cache, spec(v1beta1.PodGroupSpec{MinMember:1, MinTaskMember:map[string]int3 │
│ I1106 23:53:31.994400       1 util.go:63] schedulerPodName  is responsible to Pod spark/spark-sleep-8-driver                                                               │
│ I1106 23:53:31.994451       1 event_handlers.go:398] Added pod <spark/spark-sleep-8-driver> into cache.   
```

```
┌────────────────────────────────────────────────────────────────── Describe(spark/spark-spark-sleep-8-pg) ──────────────────────────────────────────────────────────────────┐
│ Name:         spark-spark-sleep-8-pg                                                                                                                                       │
│ Namespace:    spark                                                                                                                                                        │
│ Labels:       <none>                                                                                                                                                       │
│ Annotations:  <none>                                                                                                                                                       │
│ API Version:  scheduling.volcano.sh/v1beta1                                                                                                                                │
│ Kind:         PodGroup                                                                                                                                                     │
│ Metadata:                                                                                                                                                                  │
│   Creation Timestamp:  2025-11-06T23:53:28Z                                                                                                                                │
│   Generation:          1                                                                                                                                                   │
│   Owner References:                                                                                                                                                        │
│     API Version:           sparkoperator.k8s.io/v1beta2                                                                                                                    │
│     Block Owner Deletion:  true                                                                                                                                            │
│     Controller:            true                                                                                                                                            │
│     Kind:                  SparkApplication                                                                                                                                │
│     Name:                  spark-sleep-8                                                                                                                                   │
│     UID:                   a8ff063f-d444-4890-a75b-4be4b3bb5ad2                                                                                                            │
│   Resource Version:        221651212                                                                                                                                       │
│   UID:                     96c19dc5-4382-47e2-8a54-74f4564cd1ec                                                                                                            │
│ Spec:                                                                                                                                                                      │
│   Min Member:  1                                                                                                                                                           │
│   Min Resources:                                                                                                                                                           │
│     Cpu:     122                                                                                                                                                           │
│     Memory:  1208G                                                                                                                                                         │
│   Queue:     root                                                                                                                                                          │
│ Status:                                                                                                                                                                    │
│   Phase:  Pending                                                                                                                                                          │
│ Events:   <none>   
```

The driver is pending as well. 

Scheduler conf: 
```
actions: "enqueue, allocate, backfill"
tiers:
- plugins:
  - name: priority
  - name: gang
    enablePreemptable: false
  - name: conformance
- plugins:
  - name: drf
    enablePreemptable: false
  - name: predicates
  - name: capacity
    enableHierarchy: true 
    arguments:
      overwriteRootQueueRealCapability: true
  - name: nodeorder
  ```

JesseStutler:
> If i submit the job to the root queue directly: Scheduler logs:
> 
> ```
> │ I1106 23:53:28.184662       1 util.go:101] schedulerPodName  is responsible to PodGroup spark/spark-spark-sleep-8-pg                                                       │
> │ I1106 23:53:28.184747       1 event_handlers.go:835] Add PodGroup(spark-spark-sleep-8-pg) into cache, spec(v1beta1.PodGroupSpec{MinMember:1, MinTaskMember:map[string]int3 │
> │ I1106 23:53:31.994400       1 util.go:63] schedulerPodName  is responsible to Pod spark/spark-sleep-8-driver                                                               │
> │ I1106 23:53:31.994451       1 event_handlers.go:398] Added pod <spark/spark-sleep-8-driver> into cache.   
> ```
> 
> ```
> ┌────────────────────────────────────────────────────────────────── Describe(spark/spark-spark-sleep-8-pg) ──────────────────────────────────────────────────────────────────┐
> │ Name:         spark-spark-sleep-8-pg                                                                                                                                       │
> │ Namespace:    spark                                                                                                                                                        │
> │ Labels:       <none>                                                                                                                                                       │
> │ Annotations:  <none>                                                                                                                                                       │
> │ API Version:  scheduling.volcano.sh/v1beta1                                                                                                                                │
> │ Kind:         PodGroup                                                                                                                                                     │
> │ Metadata:                                                                                                                                                                  │
> │   Creation Timestamp:  2025-11-06T23:53:28Z                                                                                                                                │
> │   Generation:          1                                                                                                                                                   │
> │   Owner References:                                                                                                                                                        │
> │     API Version:           sparkoperator.k8s.io/v1beta2                                                                                                                    │
> │     Block Owner Deletion:  true                                                                                                                                            │
> │     Controller:            true                                                                                                                                            │
> │     Kind:                  SparkApplication                                                                                                                                │
> │     Name:                  spark-sleep-8                                                                                                                                   │
> │     UID:                   a8ff063f-d444-4890-a75b-4be4b3bb5ad2                                                                                                            │
> │   Resource Version:        221651212                                                                                                                                       │
> │   UID:                     96c19dc5-4382-47e2-8a54-74f4564cd1ec                                                                                                            │
> │ Spec:                                                                                                                                                                      │
> │   Min Member:  1                                                                                                                                                           │
> │   Min Resources:                                                                                                                                                           │
> │     Cpu:     122                                                                                                                                                           │
> │     Memory:  1208G                                                                                                                                                         │
> │   Queue:     root                                                                                                                                                          │
> │ Status:                                                                                                                                                                    │
> │   Phase:  Pending                                                                                                                                                          │
> │ Events:   <none>   
> ```
> 
> The driver is pending as well.
> 
> Scheduler conf:
> 
> ```
> actions: "enqueue, allocate, backfill"
> tiers:
> - plugins:
>   - name: priority
>   - name: gang
>     enablePreemptable: false
>   - name: conformance
> - plugins:
>   - name: drf
>     enablePreemptable: false
>   - name: predicates
>   - name: capacity
>     enableHierarchy: true 
>     arguments:
>       overwriteRootQueueRealCapability: true
>   - name: nodeorder
> ```

Could you provide more logs of it? Like about whether there are some logs showing that the driver pod was be rejected to be scheduled or the root queue exceeded the limit or something

JesseStutler:
> I tried testing this in my k8s cluster today. I created a new image only for the scheduler and used that in my cluster. I did not change controller and admission image. I created a root queue with 500 cores and 5000GB RAM (note my cluster has only <100 cores and 500GB RAM but has autoscaling enabled) I also created a child queue of root with 250 cores and 4500GB RAM.
> 
> I tried submitting a spark job but got the following error from the spark operator logs:
> 
> ```
> │ I1106 23:33:48.303393      10 event.go:285] Event(v1.ObjectReference{Kind:"SparkApplication", Namespace:"spark", Name:"spark-sleep-8", UID:"c3526aed-2566-42e1-a0a6-33a445cca51f", APIVersion:"sparkoperator.k8s.io/v1beta2", ResourceVersion:"221645563", FieldPath:""}): type: 'Normal' reason: 'SparkApplicationAdded' SparkApplication spark-sleep-8 was added, enqueuing it for submission                                                                                                                                │
> │ E1106 23:33:48.319225      10 controller.go:676] failed to process batch scheduler BeforeSubmitSparkApplication with error failed to sync PodGroup with error: admission webhook "validatepodgroup.volcano.sh" denied the request: can only submit PodGroup to queue with state `Open`, queue `myqueue` status is ``. Abandon schedule pods via volcano     
> ```
> 
> Seems that the queue i created is not reporting status:
> 
> ```
> ┌─────────────────────────────────────────────────────────────────────────── Describe(-/myqueue) ────────────────────────────────────────────────────────────────────────────┐
> │ Name:         myqueue                                                                                                                                                      │
> │ Namespace:                                                                                                                                                                 │
> │ Labels:       <none>                                                                                                                                                       │
> │ Annotations:  <none>                                                                                                                                                       │
> │ API Version:  scheduling.volcano.sh/v1beta1                                                                                                                                │
> │ Kind:         Queue                                                                                                                                                        │
> │ Metadata:                                                                                                                                                                  │
> │   Creation Timestamp:  2025-11-06T23:30:58Z                                                                                                                                │
> │   Generation:          1                                                                                                                                                   │
> │   Resource Version:    221644752                                                                                                                                           │
> │   UID:                 b9ad6d06-4ac8-4ca0-8e73-25964af8f0d6                                                                                                                │
> │ Spec:                                                                                                                                                                      │
> │   Guarantee:                                                                                                                                                               │
> │     Resource:                                                                                                                                                              │
> │       Cpu:      240                                                                                                                                                        │
> │       Memory:   4500G                                                                                                                                                      │
> │   Parent:       root                                                                                                                                                       │
> │   Reclaimable:  true                                                                                                                                                       │
> │   Weight:       1                                                                                                                                                          │
> │ Events:         <none>       
> ```
> 
> Whereas root:
> 
> ```
> │ Name:         root                                                                                                                                                         │
> │ Namespace:                                                                                                                                                                 │
> │ Labels:       <none>                                                                                                                                                       │
> │ Annotations:  <none>                                                                                                                                                       │
> │ API Version:  scheduling.volcano.sh/v1beta1                                                                                                                                │
> │ Kind:         Queue                                                                                                                                                        │
> │ Metadata:                                                                                                                                                                  │
> │   Creation Timestamp:  2025-08-20T22:08:49Z                                                                                                                                │
> │   Generation:          3                                                                                                                                                   │
> │   Resource Version:    188581782                                                                                                                                           │
> │   UID:                 de2d7990-f6dd-42c2-8e9d-524d6f07845b                                                                                                                │
> │ Spec:                                                                                                                                                                      │
> │   Capability:                                                                                                                                                              │
> │     Cpu:        500                                                                                                                                                        │
> │     Memory:     5000G                                                                                                                                                      │
> │   Reclaimable:  false                                                                                                                                                      │
> │   Weight:       1                                                                                                                                                          │
> │ Status:                                                                                                                                                                    │
> │   Allocated:                                                                                                                                                               │
> │     Cpu:     0                                                                                                                                                             │
> │     Memory:  0                                                                                                                                                             │
> │   Reservation:                                                                                                                                                             │
> │   State:  Open                                                                                                                                                             │
> │ Events:   <none>                                                                                                                                                           │
> │                    
> ```
> 
> Will do some more digging

It's weird, maybe we could check controller's log, I think there may be something wrong with controller to sync the state of queue from `""` to `Open`:
https://github.com/volcano-sh/volcano/blob/83f550edb4692ac0b3f499bc38b78705fbfb3e55/pkg/controllers/queue/state/open.go#L43-L48

hajnalmt (approved):
/lgtm

Hello @mrendi29 ,
Sorry for some reason I thought this is already merged in.
I rebased your branch, modified the logging a little bit (thank you for the access). Also I removed the AI tests with a simple negative-positive test case that follows the structure of the existing test suite. Also added the argument with some context to the appropriate design doc.

I am not sure about the failure you got, but it shouldn't be due to these changes, and this argument can be useful in other use cases.

Please take a look @JesseStutler

pkg/scheduler/plugins/capacity/capacity_test.go
hajnalmt: Fixed.

mrendi29:
@hajnalmt  TSYM for lending a hand in this, I was OOO for the last 2 weeks and have been super busy at work wrapping up year end objectives which is why I have not been able to dedicate more time at this the last month. Your change LGTM as well. i'll test it in production once i get a chance

hajnalmt:
/ok-to-test

@mrendi29 Anytime! 👍 Sorry that this got out of my focus.
One thing that came to my mind regarding the error you got, is that you shouldn't send pods to the root queue only to the leaf queues. I ran into these kind of issues I think, when I enabled hierarchy and there were pods in the root queue already.

mrendi29:
@JesseStutler can you take a look at the current state of the PR? We can resolve conflict afterwards if the approach looks good to you 

UPDATE: i also saw https://github.com/volcano-sh/volcano/pull/4973 being worked on, that should also in theory address this

JesseStutler:
> @JesseStutler can you take a look at the current state of the PR? We can resolve conflict afterwards if the approach looks good to you
> 
> UPDATE: i also saw #4973 being worked on, that should also in theory address this

Yes, #4973 will set the capability of root queue to infinite by default, but we can also overwrite the capability, WDYT? Is this way better? @mrendi29 @hajnalmt

hajnalmt:
I think this can be closed as https://github.com/volcano-sh/volcano/pull/4973 got merged in, as the root cause of the issue is solved with it. What do you think @mrendi29 ?

mrendi29:
> I think this can be closed as https://github.com/volcano-sh/volcano/pull/4973 got merged in, as the root cause of the issue is solved with it. What do you think @mrendi29 ? 

I'll re-deploy volcano tomorrow in our QA cluster and double check that it solves the root cause. I believe it should, but i will let you know

mrendi29:
Testing this next week, was busy with iceberg conference and few other projects

Issue #4671, opened 2025-10-13
Upgrade volcano k8s dependencies to v1.34


JesseStutler:
### What is the problem you're trying to solve

Kubernetes has released k8s v1.34, we need to upgrade all the dependencies to v1.34.x and make sure volcano works fine

### Describe the solution you'd like

Follow this guide to update k8s dependencies version to v1.34.x: https://github.com/volcano-sh/volcano/blob/master/docs/design/adapt-k8s-todo.md

For previous PRs, you can refer to:
- v1.33 upgrade: https://github.com/volcano-sh/volcano/pull/4430
- v1.32 upgrade: https://github.com/volcano-sh/volcano/pull/4099

JesseStutler:
> /assign

@tunedev Hi, thanks for your assignment. But we need to finish it before Oct.30, can you catch up?

tunedev:
i would try to catch up

**Edit:**
Already on it

katara-Jayprakash:
if no one is working on this i would love to work on this  i am gonna finish this asap (mainly in 3-5 day's )

hajnalmt:
@tunedev The issue is yours then, please feel free to contact us on slack if you feel stuck or have a question or anything. ☺️

PR #4687, opened 2025-10-21
Update volcano k8s dependencies to v1.34

tunedev:
What type of PR is this?
feature

This PR upgrades Kubernetes to v1.34, replacing the reference to Struct with interfaces in the new version.
It includes necessary code, YAML, and Docker changes to support this transition.

Which issue(s) this PR fixes:
fixes [#4671](https://github.com/volcano-sh/volcano/issues/4671)

### DIFF
--- a/.github/workflows/e2e_hypernode.yaml
+++ b/.github/workflows/e2e_hypernode.yaml
@@ -44,7 +44,7 @@ jobs:
           make e2e-test-hypernode CC=/usr/local/musl/bin/musl-gcc
 
       - name: Upload e2e hypernode logs
-        if: failure()
+        if: '!success()'
         uses: actions/upload-artifact@v4
         with:
           name: volcano_e2e_hypernode_logs
--- a/.github/workflows/e2e_admission.yaml
+++ b/.github/workflows/e2e_admission.yaml
@@ -1,0 +1,88 @@
+name: E2E Admission
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+  pull_request:
+
+jobs:
+  e2e_admission_policy:
+    runs-on: ubuntu-24.04
+    name: E2E about Admission Policy
+    timeout-minutes: 50
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.24.x
+
+      - name: Install musl
+        run: |
+          wget http://musl.libc.org/releases/musl-1.2.1.tar.gz
+          tar -xf musl-1.2.1.tar.gz && cd musl-1.2.1
+          ./configure
+          make && sudo make install
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - uses: actions/cache@v4
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+
+      - name: Install dependences
+        run: |
+          GO111MODULE="on" go install sigs.k8s.io/kind@v0.30.0
+          curl -LO https://dl.k8s.io/release/v1.34.0/bin/linux/amd64/kubectl && sudo install kubectl /usr/local/bin/kubectl
+
+      - name: Run E2E Tests
+        run: |
+          export ARTIFACTS_PATH=${{ github.workspace }}/e2e-admission-logs
+          make e2e-test-admission-policy CC=/usr/local/musl/bin/musl-gcc
+
+      - name: Upload e2e admission logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: volcano_e2e_admission_policy_logs
+          path: ${{ github.workspace }}/e2e-admission-logs
+  e2e_admission_webhook:
+    runs-on: ubuntu-24.04
+    name: E2E about Admission Webhook
+    timeout-minutes: 50
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.24.x
+
+      - name: Install musl
+        run: |
+          wget http://musl.libc.org/releases/musl-1.2.1.tar.gz
+          tar -xf musl-1.2.1.tar.gz && cd musl-1.2.1
+          ./configure
+          make && sudo make install
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - uses: actions/cache@v4
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+
+      - name: Install dependences
+        run: |
+          GO111MODULE="on" go install sigs.k8s.io/kind@v0.29.0
+          curl -LO https://dl.k8s.io/release/v1.32.0/bin/linux/amd64/kubectl && sudo install kubectl /usr/local/bin/kubectl
+
+      - name: Run E2E Tests
+        run: |
+          export ARTIFACTS_PATH=${{ github.workspace }}/e2e-admission-logs
+          make e2e-test-admission-webhook CC=/usr/local/musl/bin/musl-gcc
+
+      - name: Upload e2e admission logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: volcano_e2e_admission_webhook_logs
+          path: ${{ github.workspace }}/e2e-admission-logs
--- a/.github/workflows/e2e_cronjob.yaml
+++ b/.github/workflows/e2e_cronjob.yaml
@@ -46,7 +46,7 @@ jobs:
           make e2e-test-cronjob CC=/usr/local/musl/bin/musl-gcc
 
       - name: Upload e2e cronjob logs
-        if: failure()
+        if: '!success()'
         uses: actions/upload-artifact@v4
         with:
           name: volcano_e2e_cronjob_logs
--- a/.github/workflows/e2e_dra.yml
+++ b/.github/workflows/e2e_dra.yml
@@ -42,7 +42,7 @@ jobs:
           make e2e-test-dra CC=/usr/local/musl/bin/musl-gcc
 
       - name: Upload e2e dra logs
-        if: failure()
+        if: '!success()'
         uses: actions/upload-artifact@v4
         with:
           name: volcano_e2e_dra_logs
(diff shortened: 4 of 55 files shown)


.github/workflows/e2e_hypernode.yaml
JesseStutler, line 1: We also need to upgrade all of the kind version in workflows yamls to the latest, such as to kind@v0.30.0

pkg/scheduler/actions/preempt/preempt.go
JesseStutler, line 1: What is the purpose of this change? Has the directory path of the framework changed now?

pkg/scheduler/cache/cache.go
```diff
@@ -771,14 +773,14 @@ func (sc *SchedulerCache) addEventHandler() {
 		resourceClaimCache := assumecache.NewAssumeCache(logger, resourceClaimInformer, "ResourceClaim", "", nil)
```
JesseStutler, line 773: L770 ResourceClaim should also upgrade to v1 version

pkg/scheduler/cache/cache.go
JesseStutler, line 1: I am worried about a problem. I found that DRA featuregate becomes true by default in 1.34, and the featuregate value of kube-scheduler introduced by Volcano is completely based on the `DefaultKubeBinaryVersion` in the dependency. This will cause the new version of Volcano to be unable to be installed on the old version of k8s cluster, because DRA's feature-gate is turned on by default, and the old version of k8s does not have these like resourceclaim resources.

JesseStutler: Maybe we can do like this in Run of volcano scheduler: https://github.com/volcano-sh/volcano/blob/master/cmd/scheduler/app/server.go
```go
func setupComponentGlobals(config *rest.Config) error {   
    client, err := kubernetes.NewForConfig(config)  
    if err != nil {  
        return err  
    }  
      
    serverVersion, err := client.Discovery().ServerVersion()  
    if err != nil {  
        return err  
    }  
      
    kubeVersion := fmt.Sprintf("%s.%s", serverVersion.Major, serverVersion.Minor)  
    kubeEffectiveVersion := basecompatibility.NewEffectiveVersionFromString(kubeVersion, "", "")  
      
    componentGlobalsRegistry := basecompatibility.DefaultComponentGlobalsRegistry  
    if componentGlobalsRegistry.EffectiveVersionFor(basecompatibility.DefaultKubeComponent) == nil {  
        utilruntime.Must(componentGlobalsRegistry.Register(  
            basecompatibility.DefaultKubeComponent,  
            kubeEffectiveVersion,  
            utilfeature.DefaultMutableFeatureGate,  
        ))  
    }  
      
    return componentGlobalsRegistry.Set()  
}
```

I don't know whether it will work, just give an example and you should test whether volcano can also be installed in old k8s cluster

JesseStutler:
@tunedev Also please check all of the failed CIs

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -230,7 +232,10 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 					klog.Warningf("Devices %s assertion conversion failed, skip", val)
 				}
 			}
-			node.AddPod(pod)
+			// ignore this err since apiserver doesn't properly validate affinity terms
+			// and we can't fix the validation for backwards compatibility.
+			podInfo, _ := k8sframework.NewPodInfo(pod)
+			node.AddPodInfo(podInfo)
```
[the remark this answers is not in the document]
JesseStutler: Looks like gemini have some hallucinations? Node has AddPodInfo method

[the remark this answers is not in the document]
tunedev: yeah it is

cmd/scheduler/app/server.go
```diff
@@ -57,6 +60,11 @@ func Run(opt *options.ServerOption) error {
 		return err
 	}
 
+	// Align default feature-gates with the connected cluster's version.
+	if err := setupComponentGlobals(config); err != nil {
+		klog.Errorf("failed to set component globals: %v", err)
```
JesseStutler, line 65: Should return an error here when met it

cmd/scheduler/app/server.go
```diff
@@ -40,11 +40,14 @@ import (
 	"k8s.io/klog/v2"
 
 	// Register gcp auth
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
```
JesseStutler: Should put before the comment `Register gcp auth`

cmd/scheduler/app/server.go
```diff
@@ -169,3 +177,32 @@ func startMetricsServer(opt *options.ServerOption) {
...
+	if err != nil {
+		return err
+	}
+
+	kubeVersion := fmt.Sprintf("%s.%s", serverVersion.Major, serverVersion.Minor)
+	kubeEffectiveVersion := basecompatibility.NewEffectiveVersionFromString(kubeVersion, "", "")
+
+	componentGlobalsRegistry := basecompatibility.NewComponentGlobalsRegistry()
+	if componentGlobalsRegistry.EffectiveVersionFor(basecompatibility.DefaultKubeComponent) == nil {
+		utilruntime.Must(componentGlobalsRegistry.Register(
```
JesseStutler: We can refactor to 
```go
err = componentGlobalsRegistry.Register(
    basecompatibility.DefaultKubeComponent,
    kubeEffectiveVersion,
    utilfeature.DefaultMutableFeatureGate,
)
if err != nil {
    return fmt.Errorf("failed to register component globals: %w", err)
}
```
We don't need to panic here

JesseStutler:
@tunedev 
I think the main reason about all failed CIs is that dra plugin has added a `DynamicResourcesArgs` arg and will validate the arg:
https://github.com/kubernetes/kubernetes/blob/03a5f06c2695805059278c9d6b47edc3bdcf51b1/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go#L193-L199
But currently in volcano predicate plugin, the dra initialization will pass a nil arg:
https://github.com/volcano-sh/volcano/blob/a11669950aa017c815ad28aec8142a3565566611/pkg/scheduler/plugins/predicates/predicates.go#L349

Therefore will cause an fatal error:
`F1027 09:23:48.126316       1 predicates.go:356] failed to create dra plugin with err: got args of type <nil>, want *DynamicResourcesArgs`
And the scheduler will restart, therefore all the CIs may fail or timeout.

So, we need to init a DynamicResourcesArgs, just like VolumeBindingArgs in this file:
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/predicates/helper.go

We can first set the default DynamicResourcesArgs in predicate plugin, and user also can override it, to set the FilterTimeout through `framework.Arguments`, and BTW, currently `DynamicResourceArgs` only contain a FilterTimeout field, the default value of it is 10s:
https://github.com/kubernetes/kubernetes/blob/03a5f06c2695805059278c9d6b47edc3bdcf51b1/pkg/scheduler/apis/config/v1/defaults.go#L249

@tunedev

JesseStutler:
@tunedev  BTW, once you fix all the CIs and fix the review comments, please squash your commits into one

pkg/scheduler/plugins/predicates/predicates.go
JesseStutler, line 1: DRA v1.34 has lots of new feature gate, therefore we need to also add them here: https://github.com/volcano-sh/volcano/blob/a11669950aa017c815ad28aec8142a3565566611/pkg/scheduler/plugins/predicates/predicates.go#L281-L290

For the new DRA feature gates needed to add: https://github.com/kubernetes/kubernetes/blob/03a5f06c2695805059278c9d6b47edc3bdcf51b1/pkg/scheduler/framework/plugins/feature/feature.go#L53-L60

tunedev: Should the feature gate list match exactly what it in the reference link, cause there are other feature gates that are not DRA specific @JesseStutler

.github/workflows/e2e_hypernode.yaml
JesseStutler, line 1: Miss to update kind version for this hypernode file?

JesseStutler:
@suyiiyii Could you take a look of this CI? I found that currently we can't create kind cluster successfully, but after I deleted the
`MutatingAdmissionPolicy: true` in kind-config.yaml, then I can successfully created

installer/volcano-agent-latest.yaml
JesseStutler, line 1: Did you use `make update-development-yaml` to execute? There should have `volcano-agent-"latest".yaml` and `volcano-monitoring-"latest".yaml`. We should only keep volcano-agent-development.yaml, volcano-development.yaml and volcano-monitoring-development.yaml

suyiiyii:
@tunedev Here, `alpha1` should be modified to `beta1` because **MAP (MutatingAdmissionPolicy)** has entered the beta phase in Kubernetes 1.34.  
Reference file: https://github.com/tunedev/volcano/blob/feat-update-k8s-dependencies-to-1.34/hack/e2e-kind-config.yaml#L27  

Please follow the example in this commit: https://github.com/sy1g/volcano/pull/11/commits/9ed923bc7d730ca771269047e5fa3a41d49eed3d  


btw, remember to synchronize the Kind version updates across other workflows:  
1. https://github.com/tunedev/volcano/blob/feat-update-k8s-dependencies-to-1.34/.github/workflows/e2e_hypernode.yaml#L35  
2. https://github.com/tunedev/volcano/blob/feat-update-k8s-dependencies-to-1.34/.github/workflows/e2e_admission.yaml#L36

installer/volcano-agent-latest.yaml
tunedev: Just did

JesseStutler:
@tunedev Hi, tunedev, the upgrade has become urgent, and we need to quickly integrate this into our product. Additionally, other ospp projects also depend on this upgrade. Therefore, https://github.com/volcano-sh/volcano/pull/4704 will inherit your PR. You don't need to worry about your contributions not being recorded, we are continuing to make modifications based on your commit.

JesseStutler:
Merged in https://github.com/volcano-sh/volcano/pull/4704, thanks for your contribution, we have already have your record

Issue #4692, opened 2025-10-25
Scheduling main loop blocked and timeout due to un-released PreBind lock in Volcano


guoqinwill:
### Description

Volcano performs asynchronous Pod binding. After integrating with DRA, the PreBind phase introduces a lock that may not be released if an external service call hangs (default timeout: 600s), causing the main scheduling workflow to block when loading predicate plugins until the timeout releases the lock.

### Steps to reproduce the issue

1. Create a PVC, but specify an unavailable availability zone.
2. Create a Pod that binds to this PVC.
3.

### Describe the results you received and expected

The Pod remains in the Pending state because PV binding fails. However, Volcano’s session cycle time remains normal, and subsequent Pod scheduling latency is also normal.

### What version of Volcano are you using?

1.12

PR #4693, opened 2025-10-25
Scheduling main loop blocked and timeout due to un-released PreBind lock in Volcano

guoqinwill:
#### What type of PR is this?
/kind bug

Fixes # https://github.com/volcano-sh/volcano/issues/4692

```release-note

```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -902,7 +902,7 @@ func (sc *SchedulerCache) Evict(taskInfo *schedulingapi.TaskInfo, reason string)
 }
 
 // Bind binds task to the target host.
-func (sc *SchedulerCache) Bind(ctx context.Context, bindContexts []*BindContext) {
+func (sc *SchedulerCache) Bind(ctx context.Context, bindContexts []*BindContext, preBinders map[string]PreBinder) {
 	readyToBindTasks := make([]*schedulingapi.TaskInfo, len(bindContexts))
 	for index := range readyToBindTasks {
 		readyToBindTasks[index] = bindContexts[index].TaskInfo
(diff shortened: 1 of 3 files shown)


pkg/scheduler/cache/factory.go
```diff
@@ -60,3 +60,17 @@ func (r *BinderRegistry) Register(name string, binder interface{}) {
...
 	}
 }
+
+// getRegisteredPreBinders returns a snapshot of currently registered PreBinders.
+// The caller can safely use the returned map without holding the registry lock.
+func (r *BinderRegistry) getRegisteredPreBinders() map[string]PreBinder {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	preBinders := make(map[string]PreBinder, len(r.preBinders))
```
ouyangshengjia: suggested to use `maps.Clone` lnstead

pkg/scheduler/capabilities/volumebinding/binder.go
```diff
@@ -507,7 +507,7 @@ func (b *volumeBinder) BindPodVolumes(ctx context.Context, assumedPod *v1.Pod, p
 		return b, err
 	})
 	if err != nil {
-		return fmt.Errorf("binding volumes: %w", err)
+		return fmt.Errorf("binding volumes for pod %s failed : %w", assumedPod.Name, err)
```
ouyangshengjia: https://github.com/volcano-sh/volcano/blob/716fcd87b452bb51f9afbcd1f27b6794974d5327/pkg/scheduler/cache/cache.go#L1322-L1324
it is better to log pod name here when prebind returns error

pkg/scheduler/cache/cache.go
```diff
@@ -924,13 +924,12 @@ func (sc *SchedulerCache) Bind(ctx context.Context, bindContexts []*BindContext)
 				klog.ErrorS(err, "Failed to update pod status when bind task error", "task", bindContext.TaskInfo.Name)
 			}
 
-			sc.binderRegistry.mu.RLock()
-			for _, preBinder := range sc.binderRegistry.preBinders {
+			preBinders := sc.binderRegistry.getRegisteredPreBinders()
```
ouyangshengjia: https://github.com/volcano-sh/volcano/blob/716fcd87b452bb51f9afbcd1f27b6794974d5327/pkg/scheduler/cache/cache.go#L1343-L1352
Is it necessary to get `PreBinders` here and passed as an argument to func `sc.executePreBinds` and `sc.Bind`, to ensure the consistency in the use of `PreBinders`?

ouyangshengjia: @JesseStutler cc

JesseStutler: I think it's a good suggestion, because session may rewrite the preBinder in the span of `executePreBinds` and `preBindRollback` @ouyangshengjia

JesseStutler:
We also need to cherry-pick it to release-1.13 and release-1.12 once done @guoqinwill

pkg/scheduler/cache/cache.go
guoqinwill: ok

PR #4696, opened 2025-10-26
feat: support selective resource dimension checking in capacity plugin

dafu-wu:
# Add checkQueueDimensionsOnly option for selective resource checking

## What type of PR is this?
/kind feature

Adds `checkQueueDimensionsOnly` configuration option to capacity plugin, enabling selective resource dimension checking based on queue's capability definition.

**Use case**: GPU-only queues where only GPU resources need enforcement while CPU/memory remain flexible.

**Implementation**:
- New `LessEqualWithSpecifiedDimensions()` function for selective dimension comparison
- Updated capacity plugin to honor the configuration switch
- Applied to both task allocation and job enqueueing logic

Fixes #

- Backward compatible: feature is opt-in (default: false)
- Nil-safe: multiple safety checks prevent nil pointer dereference
- Tested: code compiles successfully

```release-note
Add checkQueueDimensionsOnly option to capacity plugin for selective resource checking. When enabled, only resource dimensions defined in queue's capability field are validated, allowing flexible allocation for other resources.
```

## Configuration Example

```yaml
# Scheduler config
tiers:
- plugins:
  - name: capacity
    arguments:
      checkQueueDimensionsOnly: true
```

```yaml
# Queue with GPU-only constraint
kind: Queue
metadata:
  name: gpu-queue
spec:
  capability:
    nvidia.com/gpu: "16"  # Only GPU checked, CPU/memory flexible
```

## Files Changed

- `pkg/scheduler/api/resource_info.go`: +67 lines (new comparison function)
- `pkg/scheduler/plugins/capacity/capacity.go`: +58 -6 lines (plugin logic)
- `docs/design/capacity-check-queue-dimensions-only.md`: new file (documentation)

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,12 +212,7 @@ If Volcano helps your research, we appreciate your citations. Here is the BibTeX
 
 ## Meeting
 
-Community weekly meeting for Asia: 15:00 - 16:00 (UTC+8) Friday. ([Convert to your timezone.](https://www.thetimezoneconverter.com/?t=10%3A00&tz=GMT%2B8&))
-
-Community biweekly meeting for America: 08:30 - 09:30 (UTC-8) Thursday. ([Convert to your timezone.](https://www.thetimezoneconverter.com/?t=10%3A00&tz=GMT%2B8&))
-
-Community meeting for Europe is ongoing on demand now. If you have some ideas or topics to discuss, please leave message
-in the [slack](https://cloud-native.slack.com/archives/C011GJDQS0N). Maintainers will contact with you and book an open meeting for that.
+We hold community meetings for different timezones. See [Volcano Community Meeting Info](https://github.com/volcano-sh/community#community-meeting).
 
 Resources:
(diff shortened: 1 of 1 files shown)


hajnalmt:
Hello @dafu-wu,
I see the issue and I already implemented something like this in this PR:
[https://github.com/volcano-sh/volcano/pull/4659](https://github.com/volcano-sh/volcano/pull/4659
)

What I don't understand is that why should this be a capacity plugin argument? This should be the default how the plugin shall work. If we don't specify a resource dimension we shouldn't care, so I consider this to be a bug. Of course you can specify as many arguments as you want.

dafu-wu:
> Hello @dafu-wu, I see the issue and I already implemented something like this in this PR: #4659
> 
> What I don't understand is that why should this be a capacity plugin argument? This should be the default how the plugin shall work. If we don't specify a resource dimension we shouldn't care, so I consider this to be a bug. Of course you can specify as many arguments as you want.

Sounds good, We will conduct tests to see if [4659](https://github.com/volcano-sh/volcano/pull/4659) truly solves the problem.

PR #4697, opened 2025-10-26
Fix: Prevent queue parent self-reference causing scheduler crash

dafu-wu:
# Fix: Prevent queue parent self-reference causing scheduler crash

Fix infinite recursion when a queue uses itself as parent, which causes stack overflow and scheduler crash during startup.

## Changes:
- Add webhook validation to reject self-referencing queue configurations
- Add cycle detection in capacity plugin's updateAncestors() function
- Add test case for self-referencing queue validation

#### What type of PR is this?

/kind bug

Fixes scheduler crash when `queue.spec.parent` references the queue's own name. Adds validation at admission time and cycle detection in scheduler as backup protection.

Fixes #

Two-layer defense: webhook validation (primary) + scheduler cycle detection (backup).
Pattern follows existing implementation in nodegroup plugin.

```release-note
Fix scheduler crash caused by queue self-referencing parent configuration.

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -525,7 +525,8 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 
 		attr := cp.newQueueAttr(queue)
 		cp.queueOpts[queue.UID] = attr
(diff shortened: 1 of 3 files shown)


hajnalmt:
/lgtm 
This implementation looks good to me 👍 
I was thinking about a unit test case that would show the issue, but we don't have one for the nodegroup plugin too so I can live without it.

PR #4700, opened 2025-10-27
[cherry-pick 1.13]Scheduling main loop blocked and timeout due to un-released PreBind lock in Volcano

guoqinwill:
#### What type of PR is this?
/kind bug

Fixes #cherry-pick: https://github.com/volcano-sh/volcano/pull/4693

```release-note

```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -902,7 +902,7 @@ func (sc *SchedulerCache) Evict(taskInfo *schedulingapi.TaskInfo, reason string)
 }
 
 // Bind binds task to the target host.
-func (sc *SchedulerCache) Bind(ctx context.Context, bindContexts []*BindContext) {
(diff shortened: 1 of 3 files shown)


JesseStutler:
Please help add an approve label for this, thanks @hwdef @lowang-bh @hzxuzhonghu

hzxuzhonghu (approved):
I am thinking maybe the more appropriate option is to change the binder registry way.

Currently it is registered each time on `predicatesPlugin.OnSessionOpen`, this is duplicate. Could we do that during initiation?

JesseStutler:
> I am thinking maybe the more appropriate option is to change the binder registry way.
> 
> Currently it is registered each time on `predicatesPlugin.OnSessionOpen`, this is duplicate. Could we do that during initiation?

Currently, many plugins derived from kube-scheduler are also initialized every time `predicates.OnSessionOpen` is called. If we consider registering only one, it might require many refactoring. Of course, initializing them once would be the best option.

PR #4701, opened 2025-10-27
Upload ci logs when ci is unsuccessful, include cancel

JesseStutler:
#### What type of PR is this?
/kind failing-test

Sometimes, the CI will be canceled if timeout, but we can't download the component's log from the CI, therefore it's hard for us to debug, currently we only update logs when the CI failed. But sometimes if the component met some errors, it also will block the CIs to execute, therefore we need to also upload logs when CI canceled:
image

NONE

```release-note
NONE
```


JesseStutler:
Please help /cc @hwdef @lowang-bh @Monokaix @hzxuzhonghu

PR #4702, opened 2025-10-28
Fix: Add a mount of the node memory metrics for Volcano Agent

MondayCha:
#### What type of PR is this?
/kind bug

The Volcano Agent code includes a default path, intended for reading the node's memory metrics.

> https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/metriccollect/local/memory.go#L42

```go
const (
	defaultMemInfoPath = "/host/proc/meminfo"
	memInfoPathEnv     = "MEM_INFO_PATH_ENV"
)
```

However, the deployment manifests (both the Helm chart and the development YAML) were missing the configuration to mount the host's `/proc/meminfo` file into the agent's container.

This PR fixes this issue by adding the required `hostPath` volume and `volumeMount` to the Volcano Agent's pod specification. This ensures that the agent can access the node's memory information at the expected path and collect metrics correctly.

None

```release-note
NONE
```

### DIFF
--- a/installer/helm/chart/volcano/templates/agent.yaml
+++ b/installer/helm/chart/volcano/templates/agent.yaml
@@ -85,6 +85,10 @@ spec:
           hostPath:
             path: /proc/stat
             type: File
+        - name: proc-meminfo
+          hostPath:
+            path: /proc/meminfo
+            type: File
       initContainers:
         - name: volcano-agent-init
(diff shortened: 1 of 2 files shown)


kingeasternsun:
Thanks for your contribution.

JesseStutler:
/approve
Great catch!
Could you also add a lgtm label for this? @kingeasternsun

PR #4708, opened 2025-11-03
fix: Bug in reclaiming tasks with fewer resource types by tasks with …

### LINKED ISSUE #4707: Bug in reclaiming tasks with fewer resource types by tasks with more resource types
### Description

Bug in resource reclaim under the following scenario:

A node has idle Resource1, but all running tasks on the node only use 
Resource2 and Resource3 without using Resource1. When a high-priority 
task (requiring Resource1, Resource2, and Resource3) triggers reclaim, 
it will attempt to reclaim all tasks on the node.

The root cause is that the idle resource is set to empty. Although 
the node has available Resource1, none of the current tasks are using 
Resource1, so no matter how many tasks are reclaimed, it will never 
satisfy the requirement.

1.
2.
3.

fix this bug

master

_No response_

tangshibokmpler:
…more resource types

Fixes #4707

```release-note

```

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -191,7 +191,7 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			victimsQueue := ssn.BuildVictimsPriorityQueue(victims, task)
 
 			resreq := task.InitResreq.Clone()
-			reclaimed := api.EmptyResource()
+			reclaimed := n.Idle.Clone()
 
 			// Reclaim victims for tasks.
 			for !victimsQueue.Empty() {


pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -191,7 +191,7 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			victimsQueue := ssn.BuildVictimsPriorityQueue(victims, task)
 
 			resreq := task.InitResreq.Clone()
-			reclaimed := api.EmptyResource()
+			reclaimed := n.Idle.Clone()
```
hajnalmt, line 194: Reclaim has a lot of bugs currently, and you have found one.
As far as I know there is a lot of work to rewrite of the whole action:
https://github.com/volcano-sh/volcano/pull/4634
and
https://github.com/volcano-sh/volcano/pull/4407

Please rename the variable as Gemini suggest, so it can be considered that the PR is mergable. Since the `reclaimed` naming is not standing after this change, and also please create an actual issue for this, so we can track it, that the new implementation is fixing that too.

hajnalmt (changes requested):

JesseStutler:
I think #4774 can solve this problem

hajnalmt (commented):
/close

Closing as obsolete, the reclaim refactor merged and contains this change.
https://github.com/volcano-sh/volcano/pull/4794

Issue #4713, opened 2025-11-05
Bug in VGPU request when multiple cards have different total video memory


weapons97:
### Description

For vGPU, virtual cards are requested in proportion. When multiple cards have different total video memory capacities, the total requested video memory should be configured per card.

### Steps to reproduce the issue

1. There is an environment with 8 GPUs, but one of the GPUs has slightly less video memory.

<img width="1510" height="1502" alt="Image" src="https://github.com/user-attachments/assets/545ae862-ef31-43eb-97c0-75440d9983d4" />

2. create pod request 8 cards，full memory for every card。

<img width="533" height="770" alt="Image" src="https://github.com/user-attachments/assets/79f545a5-601c-4682-b388-45448ae9d9d2" />

3. pod alawys pending。

### Describe the results you received and expected

pod should be schedule on node，not pending。

The reason is that the video memory request amount is calculated as a fixed value, rather than calculating a percentage per GPU.

<img width="851" height="1785" alt="Image" src="https://github.com/user-attachments/assets/8a49d395-f914-4d5a-a258-4072a36ea43c" />

### What version of Volcano are you using?

1.12.2

archlitchi:
i don't quite get it, in your case, you didn't specify volcano.sh/vgpu-memory or volcano.sh/vgpu-memory-percentage, so the 'val.MemPercentageReq should be 101(default number)' it shouldn't pass the condition check you modified

weapons97:
> i don't quite get it, in your case, you didn't specify volcano.sh/vgpu-memory or volcano.sh/vgpu-memory-percentage, so the 'val.MemPercentageReq should be 101(default number)' it shouldn't pass the condition check you modified

if do not specify volcano.sh/vgpu-memory MemPercentagereq will be set to 100 at pkg/scheduler/api/devices/nvidia/vgpu/utils.go

<img width="798" height="1524" alt="Image" src="https://github.com/user-attachments/assets/f77d5dfe-2fd2-42ea-b6ff-dceb9b51b5d6" />

csh0101:
so this is becuase 101 != 100 , this is error about setup the MemPercentagereq? @weapons97

weapons97:
> so this is becuase 101 != 100 , this is error about setup the MemPercentagereq? [@weapons97](https://github.com/weapons97)

MemPercentagereq be set to 100  is ok, i think problem is  'val.Memory' has been set during the first loop, that number may not fit other device, so cause this issue

[...]

PR #4714, opened 2025-11-05
Fix: Bug in VGPU request when multiple cards have different total vid…

weapons97:
…eo memory

#### What type of PR is this?
Fix bugs

Fixes https://github.com/volcano-sh/volcano/issues/4713

```release-note
NONE
```


archlitchi:
This issue fixed an issue where 'val.Memory' has been set during the first loop, that number may not fit other device, so cause this issue: https://github.com/volcano-sh/volcano/issues/4713

I've tested on my local environment

JesseStutler:
@weapons97 I think we should also cherry-pick it to release-1.13 and release-1.12

weapons97:
> @weapons97 I think we should also cherry-pick it to release-1.13 and release-1.12

Yes, I can do it if needed.

PR #4727, opened 2025-11-15
[Proposal] Gate-Controlled Scheduling for Cluster Autoscalers Compatibility

### LINKED ISSUE #4710: Cluster Autoscaler node scale-up for Pods that exceed Queue's capability

Hi Volcano team,

I have a situation that causes [Cluster Autoscaler (CA)](https://github.com/kubernetes/autoscaler) to create additional nodes in my cluster that **will never be used**, which adds additional idle costs. 

For instance, consider the following situation:

- Spark workload with **1 Driver** + **N Executors** (with `dynamicAllocation` with `min=1` and `max=5`).
- **Queue's capability** can only handle up to **1 Driver** + **1 Executor**.
- `PodGroup` with `minMember=1` and `minResources= 1 Driver + 1 Executor Resources`.
- Executor instances node group **need to scale from zero**, so before I submitted the workload, it has `size = 0`.
- Each executor Pod occupies the whole node.

When I submit the Spark workload, it eventually creates 5 executor Pods, initially at the `Pending` state. The `Executor #1` is the only one that transitions to `Running`, because the Queue's resources allow it:

Image

But the other `Pending` executors cannot run in the Queue and do not change their state, which causes additional nodes to be added to the cluster:

Image

The Pod conditions in all `Pending` executors are the following:

```
conditions:
- lastProbeTime: null
  message: 'pod group is not ready, 1 minAvailable, 2 Running, 5 Pending; Pending:
    5 Unschedulable'
  reason: Unschedulable
  status: "False"
  type: PodScheduled
phase: Pending
```

Which eventually causes the CA to [trigger scale-up](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-scale-up-work), due to the `Unschedulable` reason.

## Asking feedback...

I wanted to know if the community has faced similar issues and if there is something I can do to address this or If a change in Volcano's code is feasible to mitigate this issue, namely, regarding **when** or **how** it **sets the `Unschedulable` state** (*e.g.*, through [`schedulingGates`](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-scheduling-readiness/)).

cc @JesseStutler @kevin-wangzefeng

devzizu:
#### What type of PR is this?

/kind documentation
/kind feature

This design proposal addresses an issue where Volcano incorrectly signals cluster autoscalers (_e.g._, CA or Karpenter) to scale up nodes even when pods are only waiting for queue capacity, not cluster resources. Currently, Volcano marks all unallocated pods as `Unschedulable` regardless of the reason, causing autoscalers to interpret queue constraints as insufficient node capacity. 

This proposal introduces an **opt-in feature** using Kubernetes `schedulingGates` to hide queue-constrained pods from autoscalers, ensuring scale-up operations only trigger for legitimate node-fit failures. This aims to prevent unnecessary infrastructure costs and resource waste.

Fixes #4710

### DIFF
--- a/docs/design/scheduling-gates-queue-admission.md
+++ b/docs/design/scheduling-gates-queue-admission.md
@@ -18,14 +18,15 @@
         - [Capacity Check with Reserved Resources](#capacity-check-with-reserved-resources)
         - [Cache Initialization](#cache-initialization)
         - [Cache Updates During Allocation](#cache-updates-during-allocation)
-      - [Gate Removal During Successful Bind](#gate-removal-during-successful-bind)
+- [Limitations](#limitations)
+- [Feature gate](#feature-gate)
 - [Conclusions](#conclusions)
 - [Related Issues](#related-issues)
 
 ## Motivation
 
 Cluster autoscalers such as [Cluster Autoscaler (CA)](https://github.com/kubernetes/autoscaler/tree/master) or
-[Karpenter](https://github.com/kubernetes-sigs/karpenter) is present in almost all Kubernetes environments to
+[Karpenter](https://github.com/kubernetes-sigs/karpenter) are present in almost all Kubernetes environments to
 dynamically adjust node capacity based on scheduler signals. Both systems monitor pod scheduling conditions,
 specifically looking for pods that match:
 
@@ -133,36 +134,52 @@ func createPatch(pod *v1.Pod) ([]byte, error) {
     // ...
 }
 
-// patchSchedulingGates adds a scheduling gate for Volcano-managed pods that opted-in
+// patchSchedulingGates adds a scheduling gate for Volcano-managed pods.
+// The gate prevents cluster autoscalers from seeing the pod until Volcano
+// determines it's ready (queue admission + gang scheduling satisfied).
 func patchSchedulingGates(pod *v1.Pod) *patchOperation {
+	// Check if opt-in annotation is present
+	if !api.HasQueueAllocationGateAnnotation(pod) {
+		klog.V(4).Infof("Pod %s/%s does not have opt-in annotation, skipping gate",
+			pod.Namespace, pod.Name)
+		return nil
+	}
 
-    // Check if opt-in annotation is present
-    if !api.HasQueueAllocationGateAnnotation(pod) {
-        return nil
-    }
+	gate := v1.PodSchedulingGate{Name: schedulingv1beta1.QueueAllocationGateKey}
 
-    gates := []v1.PodSchedulingGate{
-        {Name: schedulingv1beta1.QueueAllocationGateKey},
-    }
+	// Idempotent: do not add a duplicate Volcano gate.
+	// This prevents appending the same gate multiple times if the mutation is retried.
+	for _, g := range pod.Spec.SchedulingGates {
+		if g.Name == gate.Name {
+			return nil
+		}
+	}
 
-    return &patchOperation{
-        Op:    "add",
-        Path:  "/spec/schedulingGates",
-        Value: append(pod.Spec.SchedulingGates, gates...),
-    }
+	// Parent missing: The schedulingGates slice hasn't been initialized yet.
+	// We must use "add" on the base path with an array containing our gate.
+	if pod.Spec.SchedulingGates == nil {
+		return &patchOperation{
+			Op:    "add",
+			Path:  "/spec/schedulingGates",
+			Value: []v1.PodSchedulingGate{gate},
+		}
+	}
+
+	// Parent exists: We can safely append to the existing array.
+	// Using the "-" path operator tells JSON Patch to append to the end of the array,
+	// preventing us from overwriting gates added by parallel webhooks.
+	return &patchOperation{
+		Op:    "add",
+		Path:  "/spec/schedulingGates/-",
+		Value: gate,
+	}
 }
 ```
 
 **`schedulingGates` Field Immutability**
 
 Kubernetes `schedulingGates` can only be removed and **not added after pod creation**
-([see PodSpec documentation](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#scheduling)).
-The gates will then be gradually removed in the scheduler side, since the latter contains the actions and plugins that
-determine if the Pod can be allocated to a queue and eventually a node:
-
-- **During bind**: For successfully allocated pods, gates **should be** removed atomically with binding.
-- **For lack of cluster capacity**: When queue has capacity but no node can fit the pod, gates are removed to signal the
-  `Unschedulable` condition to autoscalers and trigger a scale-up.
+([see PodSpec documentation](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#scheduling)). As such, removing the gate can result in setting the `Unschedulable` condition on the pod if no node can fit it or actually binding it to a node.
 
 #### Changes to the Volcano Scheduler
 
@@ -208,16 +225,29 @@ action while still being prevented from allocation until the gate is removed.
 
 ##### Queue Allocation and Gate Removal
 
-Each time the `Allocate` action is executed,
+In the current Volcano implementation, tasks with external (non-Volcano) scheduling gates are skipped when building the job worksheet, so they never enter the allocation loop (please refer to [`organizeJobWorksheet()`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/actions/allocate/allocate.go#L255)). However, in this proposal, we must ensure that Pods with the new proposed Volcano scheduling gate are also included in the job worksheet, so they can be accounted for in the queue capacity check:
+
+```go
+// In organizeJobWorksheet(), when building the worksheet for each subJob:
+for _, task := range subJob.TaskStatusIndex[api.Pending] {
+    // Skip tasks with external (non-Volcano) scheduling gates
+    if task.SchGated && !api.HasOnlyVolcanoSchedulingGate(task.Pod) {
+        klog.V(4).Infof("Task <%v/%v> has external scheduling gate, skip it.",
+            task.Namespace, task.Name)
+        continue
+    }
+    // ... push task into worksheet ...
+}
+```
+
+Afterwards, each time the `Allocate` action is executed,
 [`allocateResourcesForTasks(...)`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/actions/allocate/allocate.go#L551)
 tries to allocate resources for each Task in a given Queue and
 [`ssn.Allocatable(queue, task)`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/actions/allocate/allocate.go#L569)
 gives us the needed signal (by running every plugin check) for eventually removing the gate added previously by the
-**MutatingAdmissionWebhook**. Gates should be removed for signaling the `Unschedulable` condition to autoscalers when no
-node fits the pod, otherwise, the pod will be allocated to a node and gates are removed during the bind operation.
+**MutatingAdmissionWebhook**. 
 
-To avoid blocking the scheduler, gate removals for lack of cluster capacity are queued to background workers and
-processed asynchronously. The following code snippet showcases the possible high-level changes to the function
+The implementation proposes to remove gates at two points: first asynchronously after the capacity check passes (performance optimization), then synchronously before binding if needed (safety guarantee). The async removal is handled by background workers to avoid blocking the scheduler. The following code snippet showcases the high-level changes to the function
 [`allocateResourcesForTasks(...)`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/actions/allocate/allocate.go#L551):
 
 ```go
@@ -230,60 +260,62 @@ func (alloc *Action) allocateResourcesForTasks(...) {
         task := tasks.Pop().(*api.TaskInfo)
 
         // Check queue capacity (enhanced to account for reserved ungated pods)
-        // The capacity plugin's AddAllocatableFn sets task.RemoveGateDuringBind=true
-        // and adds task to reserved cache if it passes capacity checks
         if !ssn.Allocatable(queue, task) {
             // ...
             continue
         }
 
-        // Skip tasks with external (non-Volcano) scheduling gates
-        if task.SchGated && !task.RemoveGateDuringBind {
-            continue
+		// If task passed allocation check and has the QueueAllocationGate, initiate async gate removal.
+		// Gate will be removed by the background worker (best effort). During the bind operation, we need
+		// to ensure the gate is not present, otherwise the bind will fail.
+        if task.SchGated && api.HasQueueAllocationGateAnnotation(task.Pod) {
+            klog.V(3).Infof("Task %s/%s has the QueueAllocationGate, queue async gate removal", task.Namespace, task.Name)
+            alloc.schedulingGateRemoval(task, queue.UID)
         }
+
+		// Skip tasks with external (non-Volcano) scheduling gates
+		if task.SchGated {
+			if api.HasOnlyVolcanoSchedulingGate(task.Pod) {
+				klog.Warningf("Task %s/%s has Volcano scheduling gate but missing annotation %q, gate will not be removed automatically; add the annotation or remove the gate manually",
+					task.Namespace, task.Name, schedulingv1beta1.QueueAllocationGateKey)
+			} else {
+				klog.V(4).Infof("Task %s/%s has non-Volcano scheduling gate, skipping", task.Namespace, task.Name)
+			}
+			continue
+		}
 
         // ... predicate checks ...
+        // PrePredicate, node filtering, prioritization:
+        // If any fail, pod is marked Unschedulable (gate was already queued for removal above)
 
-        // Handle PrePredicate failure (e.g., topology constraints)
-        // Remove gate so pod becomes Unschedulable - signals autoscaler
-        if err := ssn.PrePredicateFn(task); err != nil {
-            // ...
-            alloc.schedulingGateRemoval(task, queue.UID)
-            // ...
-        }
-
-        // Handle no predicate nodes (no nodes fit the pod)
-        // Remove gate so pod becomes Unschedulable - signals autoscaler
-        if len(predicateNodes) == 0 {
-            // ...
-            alloc.schedulingGateRemoval(task, queue.UID)
-            // ...
-        }
-
-        // Handle no best node after prioritization
-        // Remove gate so pod becomes Unschedulable - signals autoscaler
-        if bestNode == nil {
-            // ...
-            alloc.schedulingGateRemoval(task, queue.UID)
-            // ...
-        }
-
-        // Allocate task to best node
+        // Allocate task to best node if found
         if err := alloc.allocateResourcesForTask(stmt, task, bestNode, job); err == nil {
-            // ... allocation successful ...
+            // ... allocation successful, will proceed to bind ...
         }
     }
 
     // ...
 }
 
-// schedulingGateRemoval queues async gate removal for node-fit failures
+// schedulingGateRemoval queues async gate removal if scheduling failed.
+// This ensures cluster autoscalers can see the Unschedulable condition and trigger scale-up
 func (alloc *Action) schedulingGateRemoval(task *api.TaskInfo, queueID api.QueueID) {
-    if api.HasOnlyVolcanoSchedulingGate(task.Pod) {
-        op := schGateRemovalOperation{namespace: task.Namespace, name: task.Name}
-        alloc.schGateRemovalCh <- op
-        task.SchGated = false  // Mark as ungated in cache
-    }
+	// Only enqueue gate removal if the task has only Volcano scheduling gate
+	if api.HasOnlyVolcanoSchedulingGate(task.Pod) {
+		op := schGateRemovalOperation{
+			namespace: task.Namespace,
+			name:      task.Name,
+		}
+
+		select {
+		case alloc.schGateRemovalCh <- op:
+			klog.V(3).Infof("Queued gate removal for %s/%s", task.Namespace, task.Name)
+			// Update task state immediately so it won't be queued again in this cycle
+			task.SchGated = false
+		default:
+			klog.Warningf("Gate operation queue full, skipping gate removal for %s/%s", task.Namespace, task.Name)
+		}
+	}
 }
 ```
 
@@ -316,8 +348,9 @@ func (alloc *Action) schGateRemovalWorker() {
             return
         case op := <-alloc.schGateRemovalCh:
             // Fetch fresh pod state from API server
-            pod, err := alloc.session.KubeClient().CoreV1().Pods(op.namespace).Get(
-                context.TODO(),
+            // Use cached kubeClient to avoid data race with session updates
+            pod, err := alloc.kubeClient.CoreV1().Pods(op.namespace).Get(
+                context.Background(),
                 op.name,
                 metav1.GetOptions{})
 
@@ -327,7 +360,7 @@ func (alloc *Action) schGateRemovalWorker() {
             }
 
             // Remove the Volcano scheduling gate
-            if err := cache.RemoveVolcanoSchGate(alloc.session.KubeClient(), pod); err != nil {
+            if err := cache.RemoveVolcanoSchGate(alloc.kubeClient, pod); err != nil {
                 klog.Errorf("Failed to remove gate from %s/%s: %v", op.namespace, op.name, err)
             }
         }
@@ -336,6 +369,44 @@ func (alloc *Action) schGateRemovalWorker() {
 
 // Note: When starting the worker routine (e.g., in Action initialization), we should
 // call alloc.schGateRemovalWorkersWg.Add(1) before launching the goroutine.
+```
+
+**Synchronous Gate Removal Before Bind**
+
+While the async worker provides best-effort gate removal, the [`Bind()`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/cache/cache.go#L218) operation includes a synchronous gate removal step to guarantee that Kubernetes will accept the bind request:
+
+```go
+// In DefaultBinder.Bind()
+func (db *DefaultBinder) Bind(kubeClient kubernetes.Interface, tasks []*schedulingapi.TaskInfo) map[schedulingapi.TaskID]string {
+    errMsg := make(map[schedulingapi.TaskID]string)
+    for _, task := range tasks {
+        p := task.Pod
+
+		// Ensure Volcano QueueAllocationGate is removed before bind, otherwise the bind will fail.
+		// This is a safety guarantee as the async worker may have already removed it.
+		if schedulingapi.HasQueueAllocationGateAnnotation(p) && schedulingapi.HasOnlyVolcanoSchedulingGate(p) {
+			klog.V(3).Infof("Ensuring gate is removed for pod %s/%s before bind", p.Namespace, p.Name)
+			err := RemoveVolcanoSchGate(kubeClient, p)
+
+			// On conflict, verify gates are gone
+			if apierrors.IsConflict(err) {
+				freshPod, _ := kubeClient.CoreV1().Pods(p.Namespace).Get(context.TODO(), p.Name, metav1.GetOptions{})
+				if freshPod != nil && len(freshPod.Spec.SchedulingGates) == 0 {
+					err = nil
+				}
+			}
+
+			if err != nil {
+				klog.Errorf("Failed to remove gate for <%v/%v>: %v", p.Namespace, p.Name, err)
+				errMsg[task.UID] = fmt.Sprintf("gate removal failed: %v", err)
+				continue
+			}
+		}
+
+        // Standard bind operation
+        // ...
+    }
+}
 ```
 
 ##### Queue Capacity Accounting for Ungated Pods
@@ -373,7 +444,7 @@ gets scheduled and eventually moves to the `Running` phase:
   pod-3   Pending  SchedulingGated    scheduling.volcano.sh/queue-allocation-gate
   ```
 
-- **After `pod-1` completes**, in the next cycle `pod-2` gets its gate removed (passes queue capacity check) but fails
+- **After `pod-1` is deleted or terminates**, in the next cycle `pod-2` gets its gate removed (passes queue capacity check) but fails
   to find matching nodes due to its `nodeSelector`. It transitions to `Unschedulable`, in order to trigger the cluster
   autoscaler. However, **without capacity reservation** for `pod-2`, the queue will appear empty. If `pod-3` is created,
   it will pass the capacity check and get scheduled:
@@ -509,12 +580,8 @@ ssn.AddAllocatableFn(cp.Name(), func(queue *api.QueueInfo, candidate *api.TaskIn
 
     allocatable := cp.checkQueueAllocatableHierarchically(ssn, queue, candidate)
 
-    // If queue has capacity and task has volcano gate, add to reserved cache
-    // Check annotation because gate may already be removed by async worker
-    if allocatable && candidate.SchGated && api.HasQueueAllocationGateAnnotation(candidate.Pod) {
-        // Mark task to have gate removed during bind
-        candidate.RemoveGateDuringBind = true
-        // Add to reserved cache immediately after passing capacity check
+    // If queue has capacity and task has the QueueAllocationGate annotation.
+    if allocatable && api.HasQueueAllocationGateAnnotation(candidate.Pod) {
         cp.addTaskToReservedCache(queue.UID, candidate)
     }
 
@@ -604,38 +671,13 @@ enabling other plugins like
 [`tdm`](https://github.com/volcano-sh/volcano/tree/v1.14.0/pkg/scheduler/plugins/tdm) to implement similar reservation
 cleanup logic in the future without modifying the framework.
 
-##### Gate Removal During Successful Bind
+## Feature gate
 
-Finally, the last case is when a task successfully passes **all allocation checks** (i.e.,
-[`ssn.Allocatable(queue, task)`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/actions/allocate/allocate.go#L569)
-returns `true` and there's a node that can fit the pod), then the gate will only be removed during the
-[`Bind(...)`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/cache/cache.go#L218) operation:
+The entire scheduling-gates queue-admission flow is controlled by a [feature gate](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/features/volcano_features.go), for example, `SchedulingGatesQueueAdmission` (Alpha, default off). When the gate is enabled, all of the behavior described in this document applies. This keeps existing clusters and configurations unchanged unless the feature is explicitly enabled (*e.g.* `--feature-gates=SchedulingGatesQueueAdmission=true`).
 
-```go
-func (db *DefaultBinder) Bind(kubeClient kubernetes.Interface, tasks []*schedulingapi.TaskInfo) map[schedulingapi.TaskID]string {
-    errMsg := make(map[schedulingapi.TaskID]string)
-    for _, task := range tasks {
-        p := task.Pod
+## Limitations
 
-        // Remove Volcano gate before bind if needed
-        if task.RemoveGateDuringBind {
-            if err := RemoveVolcanoSchGate(kubeClient, p); err != nil {
-                klog.Errorf("Failed to remove gate for <%v/%v>: %v", p.Namespace, p.Name, err)
-                errMsg[task.UID] = fmt.Sprintf("gate removal failed: %v", err)
-                continue
-            }
-        }
-
-        // Standard bind
-        if err := db.kubeclient.CoreV1().Pods(p.Namespace).Bind(context.TODO(), &v1.Binding{...}); err != nil {
-            // ...
-        }
-        // ...
-    }
-
-    return errMsg
-}
-```
+Once a pod's gate is removed (*i.e.*, because it has passed the queue capacity check), the pod is accounted for in the queue's reserved capacity until it is scheduled or the reservation is cleaned up. If that pod remains unschedulable (*e.g.*, no node matches, or the cluster is waiting for the autoscaler to add nodes), it continues to hold queue capacity. The scheduler cannot distinguish an ungated pod that will never become schedulable from one that will become schedulable after the autoscaler expands the cluster. As a result, other pods that could schedule may be blocked by this reserved capacity indefinitely. In this version of the design, reservations are not released for long-unschedulable pods (*e.g.*, via a timeout) to avoid over-committing the queue when those pods eventually get nodes. Operators should be aware that in such scenarios, queue capacity can be held by ungated-but-unschedulable pods and may delay scheduling of other workloads.
 
 ## Conclusions
 


JesseStutler:
@kingeasternsun @hajnalmt Do you have time help @devzizu to take a look at this proposal?

hajnalmt:
/cc @hajnalmt

Sure I am starting to review it 👍

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+
+### How Volcano Currently Sets Unschedulable
+
+> Note: Every mention of Volcano in this document refers to the latest
+> [v1.13.0 release](https://github.com/volcano-sh/volcano/releases/tag/v1.13.0).
+
+Volcano sets the `Unschedulable` condition on pods through its cache event recording mechanism in
+[`pkg/scheduler/cache/cache.go`](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/cache/cache.go). After
+each scheduling cycle, the
+[`RecordJobStatusEvent`](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/cache/cache.go#L1515) function
```
hajnalmt: Just a remark (nothing connected to the proposal) that it's funny that there is a TODO here too:
[Line 1528](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/cache/cache.go#L1528)
This should be fixed too since we shall not record this event if it's unschedulable due to the gates.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+latter is responsible for applying the logic for actions and plugins.
+
+#### Changes to the Volcano `MutatingAdmissionWebhook`
+
+Volcano's `MutatingAdmissionWebhook` needs to be extended to detect Pods annotated with
+`volcano.sh/enable-queue-allocation-gate="true"` and must patch Pods at **creation time** with a new `schedulingGate`
+entry, for instance, called `volcano.sh/queue-allocation-gate`:
+
+```go
+// Existing createPatch function needs to include the new optional patch
```
hajnalmt, line 123: Please add a link to the code snippet (as you did at other places), so it's easier to check for the function.
[createPatch](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/webhooks/admission/pods/mutate/mutate_pod.go#L103)

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+### High-Level Implementation
+
+In this proposal, changes will be required to both the **admission process** and also the **scheduler routines**, as the
+latter is responsible for applying the logic for actions and plugins.
+
+#### Changes to the Volcano `MutatingAdmissionWebhook`
+
+Volcano's `MutatingAdmissionWebhook` needs to be extended to detect Pods annotated with
+`volcano.sh/enable-queue-allocation-gate="true"` and must patch Pods at **creation time** with a new `schedulingGate`
+entry, for instance, called `volcano.sh/queue-allocation-gate`:
```
hajnalmt: `volcano.sh/queue-allocation-gate` is enough the `enable` at the beginning of the second part of the annotation is unnecesarry in the line above.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    // ...
+}
+
+// patchSchedulingGates adds a scheduling gate for Volcano-managed pods that opted-in
+func patchSchedulingGates(pod *v1.Pod) *patchOperation {
+
+    // Check if opt-in annotation is present
+    if pod.Annotations == nil || pod.Annotations["volcano.sh/enable-queue-allocation-gate"] != "true" {
```
hajnalmt: Okay. This annotation should be part of the API. Don't hide it in the code itself.

Please add it here:
[Scheduling V1beta1 api](https://github.com/volcano-sh/apis/blob/master/pkg/apis/scheduling/v1beta1/labels.go#L48 
) as a key like: `const QueueAllocationGateKey = GroupName + "/queue-allocation-gate"`
and reuse it later on.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+        return nil
+    }
+
+    // ...
+
+    return &patchOperation{
+        Op:    "add",
+        Path:  "/spec/schedulingGates",
```
hajnalmt: This is an array!
Use the "-" JSON Patch operator to append to the end of the array properly.
This line should be:
```go
    Path: "/spec/schedulingGates/-",
```
This way we won't bump into issues with other webhooks or kubectl patch operations.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+- **During bind**: For successfully allocated pods, gates are removed atomically with binding.
+- **For lack of cluster capacity**: When queue has capacity but no node can fit the pod, gates are removed to signal the
+  `Unschedulable` condition to autoscalers and trigger a scale-up.
+
+#### Changes to the Volcano Scheduler
+
+Volcano already adopted the concept of `schedulingGates` through the **Pod Scheduling Readiness** design document
+([pod-scheduling-readiness.md](https://github.com/volcano-sh/volcano/blob/master/docs/design/pod-scheduling-readiness.md)),
+which skips tasks of a job whose Pods are scheduling gated, ensuring that a scheduling gated pod will not be bound to a
+node. In this case, the
```
hajnalmt, line 189: Easily followable, beautiful gathering about the occurances. Nice work.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+Task in a given Queue and `ssn.Allocatable(queue, task)` gives us the needed signal (by running every plugin check) for
+eventually removing the gate added previously by the **MutatingAdmissionWebhook**. Gates should be removed for signaling
+the `Unschedulable` condition to autoscalers when no node fits the pod, otherwise, the pod will be allocated to a node
+and gates are removed during the bind operation.
+
+##### Asynchronous Gate Removal
+
+To avoid blocking the scheduler, gate removals for lack of cluster capacity are queued to background workers and
+processed asynchronously. The following code snippet showcases the possible high-level changes to the function
+`allocateResourcesForTasks(...)`:
```
hajnalmt: Mark the code line here too, the thing is that allocateResourcesForTasks is an overloaded functio. It's there in the file twice with the same `allocateResourcesForTasks(...)` name but with different arguments:
[L356](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/actions/allocate/allocate.go#L356) - this is the one we are speaking about.
[L544](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/actions/allocate/allocate.go#L544) - this is the other variant

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        // ...
+
+        // ADD: Handle PrePredicate failure (e.g., topology constraints)
+        // Remove gate so pod becomes Unschedulable - this is a legitimate signal
+        // for autoscalers since it indicates node resource/constraint issues
+        if err := ssn.PrePredicateFn(task); err != nil {
+            // ...
+            alloc.enqueueSchedulingGateRemoval(task)
```
hajnalmt: Why have you named this `enqueueSchedulingGateRemoval`? This doesn't really have tight connection with enqueue.
`schedulingGateRemoval(task)` is a good enough name for the function I think.

Nevertheless, the idea and the implementation looks solid.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+
+```go
+// Enhance the Action struct
+type Action struct {
+    session *framework.Session
+    // ...
+
+    // Async gate removal channel
+    schGateRemovalOperationCh chan schGateRemovalOperation
+    schGateRemovalWorkersWg   sync.WaitGroup
```
hajnalmt: I don't see this WaitGroup used in the end, but it's a good idea. Please add the defer block with Done to the function below and use `schGateRemovalWorkersWg.Add(1)` when you start the routine (this is not part of the proposal, just don't forget it in the actual implementation.)

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+}
+
+type schGateRemovalOperation struct {
+    namespace string
+    name      string
+}
+
+// Background worker processes gate removal operations
+func (alloc *Action) schGateRemovalWorker() {
+    for {
```
hajnalmt, line 345: defer alloc.schGateRemovalWorkersWg.Done()
before the loop

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+```go
+// Enhance the Action struct
+type Action struct {
+    session *framework.Session
+    // ...
+
+    // Async gate removal channel
+    schGateRemovalOperationCh chan schGateRemovalOperation
+    schGateRemovalWorkersWg   sync.WaitGroup
+    schGateRemovalShutdownCh  chan struct{}
```
hajnalmt: Let's call this schGateRemovalStopCh. It's a little bit more idiomatic.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bool {
+
+    // ...
+
+    for status, tasks := range job.TaskStatusIndex {
+        if api.AllocatedStatus(status) {
+            //...
+        } else if status == api.Pending {
+            for _, t := range tasks {
+                if !t.SchGated {
```
hajnalmt: Ouch. This looks to me a significant behavioral change.

If I understand correctly, we’re now modifying the allocated attribute entirely, meaning that all non‑gated (Pending) pods will be treated as allocated. That would also affect how metrics are reported, since Pending pods would show up differently than before. Can you confirm that this interpretation is correct?

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+            return
+        }
+    }
+}
+```
+
+##### Queue Capacity Accounting for Ungated Pods
+
+When a pod's gate is removed due to lack of cluster capacity, it becomes visible to autoscalers but remains unallocated
+(waiting for matching nodes). This creates a potential race condition: between gate removal and actual allocation to a
```
hajnalmt: I need to think about this race condition, probably I am not fully understanding it. It looks (at least to me) a general scheduling behaviour, if you don't annotate every pod than they will work the same way as they worked earlier, and the cluster-autoscaler won't work properly still.  Can you help me with an example here?

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+
+This ensures that once a gate is removed, the pod effectively "reserves" its queue slot, preventing other pods from
+consuming that capacity. The pod remains in this reserved state until it either successfully allocates to a node or is
+deleted.
+
+##### Gate Removal During Successful Bind
+
+Finally, the last case is when a task successfully passes **all allocation checks** (i.e.,
+`ssn.Allocatable(queue, task)` returns `true` and there's a node that can fit the pod), then the gate will only be
+removed during the `Bind(...)` operation:
```
hajnalmt: Link: [L209](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/cache/cache.go#L209)

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,340 @@
...
+
+```go
+func (db *DefaultBinder) Bind(...) map[schedulingapi.TaskID]string {
+    // ...
+
+    for _, task := range tasks {
+        if task.RemoveGateDuringBind && hasOnlyVolcanoSchedulingGate(task.Pod) {
```
hajnalmt: The `hasOnlyVolcanoSchedulingGate` is not needed here because it was a prerequesite for the settings of `RemoveGateDuringBind` earlier.
```go
        // Queue has capacity - mark task for gate removal during bind
        if task.SchGated && hasOnlyVolcanoSchedulingGate(task.Pod) {
            task.RemoveGateDuringBind = true
        }
 ```

**Review by hajnalmt — CHANGES_REQUESTED** [review]
Thank you for this great proposal @devzizu.
I am looking forward to the implementation, great idea and solid design change.

I had mostly minor remarks except the capacity plugin change.
What I see is that the pods that are marked scheduling gated by volcano after this change with the annotation they can maybe considered to be `inqueue` but straight out to be considered `allocated` seems a too significant change.

Wouldn’t it be sufficient to adjust the [DeductSchGatedResources](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/api/job_info.go#L557) function so that the tasks that are only volcano scheduling gated  won't be deducted when  `hasOnlyVolcanoSchedulingGate` returns true? This approach would also maintain compatibility with the overcommit plugin, which I’m slightly concerned about since it heavily depends on scheduling gate behavior.
--- I want to remark that this would mean that these annotated pods would move to inqueue and would never be in the scope of the enqueue action if every other scheduling gate is removed, which is probably what we want.


Thank you for the contribution and the idea once more!

File: docs/design/scheduling-gates-queue-admission.md (hunk 4 above)
Comment (devzizu) [reply docs/design/scheduling-gates-queue-admission.md]: Yes, agree! Added 🙏

File: docs/design/scheduling-gates-queue-admission.md (hunk 1 above)
Comment (devzizu) [reply docs/design/scheduling-gates-queue-admission.md]: Shall I create an issue that targets this fix?

File: docs/design/scheduling-gates-queue-admission.md (hunk 8 above)
Comment (devzizu) [reply docs/design/scheduling-gates-queue-admission.md]: Yes, agree with your suggestion -- the name `schedulingGateRemoval` looks more concise. I previously named it enqueue as we are sending ("enqueuing") the gate removal into a channel (not related with Volcano's/scheduling enqueue concept).

File: docs/design/scheduling-gates-queue-admission.md (hunk 9 above)
Comment (devzizu) [reply docs/design/scheduling-gates-queue-admission.md]: Thank you, fixed and will consider that in the implementation 🙏

File: docs/design/scheduling-gates-queue-admission.md (hunk 12 above)
Comment (devzizu) [reply docs/design/scheduling-gates-queue-admission.md]: @hajnalmt You're correct: counting ungated `Pending` pods as "allocated" would be a significant behavioral change, affecting metrics and the meaning of "allocated". Modifying `DeductSchGatedResources` to exclude pods with only the Volcano scheduling gate looks much better because it avoids changing the semantics of "allocated", keeps Volcano-gated pods counted in queue capacity (as "inqueue"), and should maintain compatibility with existing plugins.

When I have some spare time, I'll test this approach and update the design document accordingly (and also answer [your comment](https://github.com/volcano-sh/volcano/pull/4727#discussion_r2589434165) about the race condition). The change should be in `GetSchGatedPodResources()` to exclude pods where `hasOnlyVolcanoSchedulingGate(task.Pod)` returns `true`, so they remain counted in queue capacity calculations.

File: docs/design/scheduling-gates-queue-admission.md (hunk 13 above)
Comment (devzizu) [reply docs/design/scheduling-gates-queue-admission.md]: To give an example of this situation, let's say we have these conditions:

- A Pod with `schedulerName: volcano` and requesting 1GiB and 1 Core.
- Consider that we have 3 Pods with the same spec as defined in **1.**, except that **Pod number 2** has a `nodeSelector` for a node that **does not exist yet** (may be spun up by a cluster autoscaler).
- A PodGroup of `minMember=1` and `minResources = 1 GiB and 1 Core`.
- A Volcano Queue with `Capability = 1 GiB and 1 Core`.

So, for each of these Pods (1, 2, and 3) in this `PodGroup`, there should be **one allocation at a time** in the Queue that I've defined. For this specific case, the allocation will work as follows:

1. Initially, the webhook adds gates for every pod:

    ```
    NAME       PHASE    CONDITION        PODGROUP  GATES
    pod-1      Pending  SchedulingGated  pg-test    volcano.sh/queue-allocation-gate
    pod-2      Pending  SchedulingGated  pg-test    volcano.sh/queue-allocation-gate
    pod-3      Pending  SchedulingGated  pg-test    volcano.sh/queue-allocation-gate
    ```

2. After the 1st allocation cycle, `pod-1` passes allocation check and gets its gate removed. Since it does not have any node selector and there are nodes available, it gets scheduled and starts **Running**:

    ```
    NAME       PHASE    CONDITION        PODGROUP  GATES
    pod-1      Running                   pg-test
    pod-2      Pending  SchedulingGated  pg-test    volcano.sh/queue-allocation-gate
    pod-3      Pending  SchedulingGated  pg-test    volcano.sh/queue-allocation-gate
    ```

3. If I delete `pod-1`, then `pod-2` gets its gate removed (passes allocation check but fails to find nodes because it has a **fake nodeSelector** (needs Cluster Autoscaler to kick-in) and moves to `Unschedulable`. However, when trying to allocate `pod-3` in the next scheduling cycle, it gets allocated to a node because it does not have a `nodeSelector` and we aren't "reserving" the space occupied by the pod in `Unschedulable` state:

    ```
    NAME       PHASE    CONDITION        PODGROUP  GATES
    pod-2      Pending  Unschedulable    pg-test
    pod-3      Running                   pg-test
    ```

So, in this case, we have a queue that can handle only one pod at a time, but we have one pod running (`pod-3`) and also `pod-2` triggering autoscaling although it will never be scheduled because `pod-3` is running. As such, this is the "race condition" that I'm facing at the moment.

To solve this, we would **need a way to "reserve" the space occupied by pods** that had their gates removed and have the `Unschedulable` condition. Initially, I was proposing to do the reservation in [buildHierarchicalQueueAttrs](https://github.com/volcano-sh/volcano/pull/4727/files#diff-22c432a2a85a3141a24bb38a782a2e69953e9cfe10e69d637357cb40ccba77f1R281), but that indeed is a significant change in scheduling behavior [as you suggested](https://github.com/volcano-sh/volcano/pull/4727/files#r2589313512).

File: docs/design/scheduling-gates-queue-admission.md (hunk 12 above)
Comment (devzizu) [reply docs/design/scheduling-gates-queue-admission.md]: Hi @hajnalmt,

Just enhanced the proposal with a the following changes:

- Added a [table of contents](https://github.com/volcano-sh/volcano/blob/61ee59292d7cb5bc8ba4f31068935e3c101491b4/docs/design/scheduling-gates-queue-admission.md#table-of-contents).
- Added a [Queue Admission Checks](https://github.com/devzizu/volcano/blob/feat-proposal-4710/docs/design/scheduling-gates-queue-admission.md#queue-admission-checks) section, which incorporates the enhancement in [DeductSchGatedResources()](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/api/job_info.go#L557) as you [proposed](https://github.com/volcano-sh/volcano/pull/4727#pullrequestreview-3539738877).
   - This change allows Gated Pods to be considered in the enqueue action.
- Enhanced the [Queue Capacity Accounting for Ungated Pods](https://github.com/volcano-sh/volcano/blob/1f0df13da96c166a1bbf19d3016809ca456fd0f9/docs/design/scheduling-gates-queue-admission.md#queue-capacity-accounting-for-ungated-pods) section with the following: 
   - [Dynamic Reserved Calculation](https://github.com/volcano-sh/volcano/blob/1f0df13da96c166a1bbf19d3016809ca456fd0f9/docs/design/scheduling-gates-queue-admission.md#dynamic-reserved-calculation): Here, I explain how I'll be "reserving" the space occupied by ungated Pods in the `Pending` state. Instead of changing the [buildHierarchicalQueueAttrs](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/plugins/capacity/capacity.go#L554) (as I [previously suggested](https://github.com/volcano-sh/volcano/blob/222a8f1bb952cb540e3a2a1954875ebc28820736/docs/design/scheduling-gates-queue-admission.md#queue-capacity-accounting-for-ungated-pods)), which runs **only once per scheduling cycle**, I propose adding a new "reserved" cache in the capacity plugin. If we have multiple pods in the same scheduling cycle, we need to calculate this dynamically to influence the scheduling decision of subsequent pods.

Let me know what you think of this approach which aims to prevent the issues [you mentioned](https://github.com/volcano-sh/volcano/pull/4727#discussion_r2589313512), i.e., changing the meaning of `allocated`, affecting metrics and breaking compatibility with other plugins, for example.

File: docs/design/scheduling-gates-queue-admission.md [hunk 16]
```diff
@@ -0,0 +1,479 @@
...
+fields**:
+
+```go
+// Enhance the Action struct
+type Action struct {
+    session *framework.Session
+    // ...
+
+    // Async gate removal channel
+    schGateRemovalStopCh chan schGateRemovalOperation
```
hajnalmt: I am really sorry, I have confused you here with the naming. I meant to rename the Shutdown channel to Stop channel.
I think the channel names here shall be:
```
schGateRemovalCh chan schGateRemovalOperation
schGateRemovalWorkersWg sync.WaitGroup
schGateRemovalStopCh chan struct{}
```

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,479 @@
...
+    if !cached {
+        // Calculate reserved: ungated Pending pods with queue allocation annotation
+        reserved = api.EmptyResource()
+        for _, job := range cp.session.Jobs {
+            if job.Queue != queue.UID {
+                continue
+            }
+            for _, task := range job.Tasks {
+                if !task.SchGated && task.Status == api.Pending &&
+                    cache.HasQueueAllocationGateAnnotation(task.Pod) {
```
hajnalmt: Where is this `cache` variable and the `HasQueueAllocationGateAnnotation` function coming from?
This would have been the scheduler cache right?

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,479 @@
...
+[`queueAllocatable`](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/plugins/capacity/capacity.go#L843)
+function by dynamically calculating reserved resources from ungated Pending pods.
+
+The capacity plugin's struct should be enhanced to include a cache:
+
+```go
+type capacityPlugin struct {
+    // ... existing fields ...
+
+    // Cache of reserved resources per queue (invalidated when task state changes)
```
hajnalmt: We should add a comment here I think that the volcano scheduling gated resources are part of this cache.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,479 @@
...
+```
+
+A new method `queueAllocatableWithReserved` should be added to the capacity plugin:
+
+```go
+// queueAllocatableWithReserved checks if queue has capacity for candidate
+// by dynamically calculating reserved resources (ungated Pending pods)
+func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidate *api.TaskInfo, queue *api.QueueInfo) bool {
+    // Check cache first
+    reserved, cached := cp.reservedCache[queue.UID]
```
hajnalmt: We should build up this cache by queue + task (TaskInfo) I think. The problem with this is that this will not work through more sessions, you cache some tasks in the first session but if one more comes in the next (although the cache is still not cleared) you will miss it.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,479 @@
...
+    if !cached {
+        // Calculate reserved: ungated Pending pods with queue allocation annotation
+        reserved = api.EmptyResource()
+        for _, job := range cp.session.Jobs {
```
hajnalmt: We don't have access to the session in the plugin yet 😞 you need to save it in OnSessionOpen, if you want something like this. Although if we make a reserved cache by queue and TaskInfo, we can remove everything here, as candidate taskInfo has every information in it based on which we can just add it to the reserved cache if it needs to be added.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,479 @@
...
+```go
+func (cp *capacityPlugin) queueAllocatable(queue *api.QueueInfo, candidate *api.TaskInfo) bool {
+    attr := cp.queueOpts[queue.UID]
+    return cp.queueAllocatableWithReserved(attr, candidate, queue)
+}
+```
+
+###### Cache Invalidation
+
+The reserved resources cache must be invalidated when a task's gate is removed during a scheduling cycle. This ensures
```
hajnalmt: If a task gate is removed and we invalidate the whole queue cache, other tasks will be removed from the cache too. Also can you elaborate on where will this will be called exactly? During bind somehow?

docs/design/scheduling-gates-queue-admission.md
hajnalmt: Yes please let's do that 👍 thank you

docs/design/scheduling-gates-queue-admission.md
hajnalmt: This example makes everything very clear. Please add this example to the design doc too!

docs/design/scheduling-gates-queue-admission.md
devzizu: No worries at all, just fixed this!

docs/design/scheduling-gates-queue-admission.md
devzizu: Yes, that was originally in the scheduler cache (`pkg/scheduler/cache/util.go`).  I've since simplified and centralized both `HasQueueAllocationGateAnnotation()` and `HasOnlyVolcanoSchedulingGate()` in `pkg/scheduler/api/helpers.go` and left a note in the end [of this subsection](https://github.com/devzizu/volcano/blob/proposal-4710/docs/design/scheduling-gates-queue-admission.md#changes-to-the-volcano-mutatingadmissionwebhook). WDYT?

docs/design/scheduling-gates-queue-admission.md
devzizu: Agree! Just added to the [Queue Capacity Accounting for Ungated Pods](https://github.com/devzizu/volcano/blob/proposal-4710/docs/design/scheduling-gates-queue-admission.md#queue-capacity-accounting-for-ungated-pods) section.

docs/design/scheduling-gates-queue-admission.md
devzizu: Agree, I forgot about that -- added to the [Dynamic Reserved Calculation](https://github.com/devzizu/volcano/blob/proposal-4710/docs/design/scheduling-gates-queue-admission.md#dynamic-reserved-calculation).

docs/design/scheduling-gates-queue-admission.md
devzizu: Agree! @hajnalmt, just submitted a version of this "reserved cache" feature that takes into consideration the concerns and approach you mentioned -- explaining how it will be initialised, cleared and updated throughout each scheduling cycle. Please let me know what you think about this, and thank you for the valuable review here!

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,578 @@
...
+
+###### Capacity Check with Reserved Resources
+
+A new method `queueAllocatableWithReserved` performs capacity checks including reserved resources:
+
+```go
+func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidate *api.TaskInfo, queue *api.QueueInfo) bool {
+    // Calculate total reserved resources directly from cache
+    reserved := api.EmptyResource()
+    if queueCache := cp.queueReservedTasks[queue.UID]; queueCache != nil {
```
hajnalmt: Rename queueCache maybe to queueReserved or queueGateReserved ☺️

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,578 @@
...
+func (alloc *Action) allocateResourcesForTask(stmt *framework.Statement, task *api.TaskInfo, node *api.NodeInfo, job *api.JobInfo) error {
+    // Check if node has idle resources
+    if task.InitResreq.LessEqual(node.Idle, api.Zero) {
+        // Attempt to bind task to node
+        if err = stmt.Allocate(task, node); err != nil {
+            // Bind failed - rollback
+            stmt.UnAllocate(task)
+        } else {
+            // Task successfully allocated - remove from reserved cache
+            alloc.session.RemoveTaskFromCapacityReservedCache(job.Queue, task.UID)
```
hajnalmt: This is not the right place to do this, because the statement can be discarded later (for example due to gang constraints) see:
Line 447:
https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/actions/allocate/allocate.go#L447

The whole relevant code block:
https://github.com/volcano-sh/volcano/blob/943b8c23d402462b63f1e21f7f7a39129ba8d63c/pkg/scheduler/actions/allocate/allocate.go#L432-L450

What you need to do is that before `stmt.Commit()` you shall go through the statement operations and for the allocate ones you try to remove the tasks. You can even make a new function inside allocate for this.
```
for _, op := range stmt.Operations() {
		switch op.name {
		case Allocate:
		     alloc.session.RemoveTaskFromCapacityReservedCache(task.UID)
		}
}
```
As you can see we have just task in this function as a parameter as statement contains only that, but it's probably enough though, as it can be removed from the `queueGateReservedTasks` like:
``` 
func RemoveTaskFromCapacityReservedCache(taskID api.TaskId) {
    for queueID, tasks := range cp.queueGateReservedTasks {
        if _, exists := tasks[taskID]; exists {
            delete(tasks, taskID)
            if len(tasks) == 0 {
                delete(cp.queueGateReservedTasks, queueID)
            }
            return
        }
    }
}
```
This is fast too, as the range will don't do anything if the feature is opted out.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,578 @@
...
+func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
+    // Rebuild reserved cache for this scheduling cycle
+    cp.buildQueueReservedTasksCache(ssn)
+
+    // ... rest of capacity plugin initialization ...
+}
+
+func (cp *capacityPlugin) buildQueueReservedTasksCache(ssn *framework.Session) {
+    // Initialize the cache for this session
+    cp.queueReservedTasks = make(map[api.QueueID]map[api.TaskID]*api.TaskInfo)
```
hajnalmt: I am trying the find a better name for this `queueReservedTasks` seems too general and doesn't reflect the intent fully. `queueGateReservedTasks` maybe?

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,578 @@
...
+}
+```
+
+This incremental update ensures that:
+
+- Tasks passing capacity checks in the current scheduling session are immediately tracked.
+- Successfully allocated tasks stop reserving capacity for new candidates.
+- The reserved cache remains accurate throughout the scheduling session.
+
+> **Note**: The `AddTaskToCapacityReservedCache(...)` and `RemoveTaskFromCapacityReservedCache(...)` methods can be
```
hajnalmt: I feel that from software architectural point of view this is the most important part of the whole feature, yet we have just a vague note here.

This is the **extension point** of the whole feature, how do you imagine this exactly? Session is a struct not an interface, and it looks like these are methods on it, how will it invoke a capacity plugin function? Do we need the `AddTaskToCapacityReservedCache` method too? The plugins and the session are loosely coupled now, currently they are just registering their functions through `Add` functions like `AddAllocatableFn` and `AddPreemptiveFn` and we shouldn't step out in an other direction, breaking this coupling.

What if we want to use this feature in `proportion` or `tdm` later on? It's okay to have this implemented in capacity plugin first, but we shall think about this at design unfortunately. Also the **extension point**  can be general, like `ssn.RemoveReserved()` so plugins can use it for whatever they would actually like. For example, we could add an `AddReservationCleanupFn` similar to other extension points, and the capacity plugin registers its cleanup logic there.

You can even give the whole `statement` to the function if it suites you and the unfolding of the statement can happen in the plugin itself even.

hajnalmt (changes requested):
We are getting there! Keep up the great work 😊

docs/design/scheduling-gates-queue-admission.md
devzizu: Thank you @hajnalmt for the insightful feedback on the architectural design - you're absolutely right that this is the most critical part of the feature and yet I've not though about being extensible and breaking coupling.

I’ve refined my approach to better address the architectural concerns by implementing a generic `ReservationCleanupFn` extension point that I believe better follows Volcano's standard plugin pattern. Additionally, I removed the capacity-specific `AddTaskToCapacityReservedCache` and `RemoveTaskFromCapacityReservedCache` methods from Session, which were creating tight coupling, and replaced them with `AddReservationCleanupFn` registration, similar to how `AddAllocatableFn` and `AddPreemptiveFn` work.

The capacity plugin now registers its cleanup logic in `OnSessionOpen()`, and as you suggested, I pass the full statement to the function so the plugin can unfold and process operations itself:

```go
ssn.AddReservationCleanupFn(cp.Name(), func(stmt *framework.Statement) {
    for _, op := range stmt.Operations() {
        if op.Name() == framework.Allocate {
            cp.removeTaskFromReservedCache(op.Task().UID)        
        }
}})
```

I've made the cache management methods (`addTaskToReservedCache`, `removeTaskFromReservedCache`) private to the plugin to maintain encapsulation.

docs/design/scheduling-gates-queue-admission.md
devzizu: You're absolutely right - I've fixed this timing issue by removing tasks from the reserved cache before `stmt.Commit()` instead, as you proposed:

```go
// In allocateResources()
if stmt != nil {
    ssn.CleanupReservations(stmt)
    stmt.Commit()
}
```

The`CleanupReservations` method iterates through all statement operations and removes allocated tasks from the reserved cache. This change was further enhanced given the refactor suggested by [your comment](https://github.com/volcano-sh/volcano/pull/4727#discussion_r2694308079).

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,626 @@
...
+
+```go
+func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
+    // Rebuild reserved cache for this scheduling cycle
+    cp.buildQueueReservedTasksCache(ssn)
+
+    // Register reservation cleanup function
+    // This is further described in the
+    // "Removing Tasks from Reserved Cache" section below.
+    // ssn.AddReservationCleanupFn(cp.Name(), ...)
```
hajnalmt: Maybe `ssn.AddCleanupReservationsFn` is more idiomatic. Also remove the comment 👍

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,626 @@
...
+    }
+}
+```
+
+The reservation cleanup mechanism follows the standard Volcano plugin extension point pattern, similar to
+[`AddAllocatableFn`](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/framework/session_plugins.go#L130)
+and
+[`AddPreemptiveFn`](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/framework/session_plugins.go#L125).
+The `ReservationCleanupFn` type can be defined in
+[`pkg/scheduler/framework/statement.go`](https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/framework/statement.go)
```
hajnalmt: The definition shall be defined in the api:
https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/api/types.go 
Used on the session https://github.com/volcano-sh/volcano/blob/v1.13.0/pkg/scheduler/framework/session.go

hajnalmt (commented):
Great Work!

Please squash your commits, and sign the DCO! I think this is quite ready.
Do you need any help with the implementation? We will need to add an e2e test for this too.
I think the it shall happen in a different PR though as the design is quite big.

hajnalmt:
/cc @JesseStutler

I think we can pull this in release-1.15

devzizu:
> Great Work!
> 
> Please squash your commits, and sign the DCO! I think this is quite ready. Do you need any help with the implementation? We will need to add an e2e test for this too. I think the it shall happen in a different PR though as the design is quite big.

Rebased this doc with Volcano recent v1.14 release.

Also, opened https://github.com/volcano-sh/volcano/pull/5033 to further iterate on the implementation proposed in this design.

cc @hajnalmt @JesseStutler

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,659 @@
...
+            continue
+        }
+
+        // ... predicate checks ...
+
+        // Handle PrePredicate failure (e.g., topology constraints)
+        // Remove gate so pod becomes Unschedulable - signals autoscaler
+        if err := ssn.PrePredicateFn(task); err != nil {
+            // ...
+            alloc.schedulingGateRemoval(task, queue.UID)
```
JesseStutler, line 273: I'm wondering that why doesn't the scheduling gate get removed directly after the Allocatable check passes? Then the subsequent checks are all about node-related checks, if the check fails, then it is normal to mark the Unschedulable condition, correct me if I'm wrong

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,659 @@
...
+                op.name,
+                metav1.GetOptions{})
+
+            if err != nil {
+                klog.Errorf("Failed to get pod %s/%s for gate operation: %v", op.namespace, op.name, err)
+                continue
+            }
+
+            // Remove the Volcano scheduling gate
+            if err := cache.RemoveVolcanoSchGate(alloc.session.KubeClient(), pod); err != nil {
```
JesseStutler: Does removing the scheduling gate and scheduling the pod require two or more rounds of sessions?
1. If the queue is sufficient, remove the scheduling gate.
2. Update the pod and wait for the event to be triggered.
3. After the next one or more session rounds, the pod becomes a normal pending pod and then goes through the scheduling process.

Am I right?

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,659 @@
...
+
+To prevent this, the queue capacity accounting logic must be enhanced to treat **ungated pods as "reservations" that
+count toward the queue's capacity checks**. Specifically, the
+[`ssn.Allocatable(queue, task)`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/framework/session_plugins.go#L351)
+function (which invokes the capacity plugin) needs to account for:
+
+1. Pods with `AllocatedStatus` (bound, binding, running) toward queue capacity _(current behavior)_.
+2. **`Pending` pods that have had their gates removed** as "reserved" resources _(new behavior)_.
+
+###### Dynamic Reserved Calculation
```
JesseStutler, line 470: Do we need to consider the scheduler restart scenario? What if the scheduler restarts after the pod scheduling gate has been removed? But it may become complicated I think

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,659 @@
...
+
+**Removing Tasks from Reserved Cache**
+
+The capacity plugin registers a **reservation cleanup function** during
+[`OnSessionOpen()`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/plugins/capacity/capacity.go#L91)
+that removes tasks from the reserved cache before statement commit:
+
+```go
+// In capacity plugin's OnSessionOpen
+ssn.AddCleanupReservationsFn(cp.Name(), func(obj interface{}) {
```
JesseStutler, line 600: There seem to be too many customized logic here, only used for this specific purpose. Perhaps we need to schedule a meeting to discuss this. Also, I don't quite understand the logic behind `Removing Tasks from Reserved Cache` part

docs/design/scheduling-gates-queue-admission.md
devzizu: That's a good point! We can call `schedulingGateRemoval` right after the `ssn.Allocatable(queue, task)` check passes, but we must ensure gates do not exist before bind. I've simplified the proposal with this in mind, removing the need of having the `RemoveGateDuringBind` helper.

docs/design/scheduling-gates-queue-admission.md
devzizu: I believe we have different flows that can lead to one or more sessions, namely:

- When the queue has enough capacity, we enqueue gate removal and set `task.SchGated = false` in memory, so we don’t skip the task in that session (happens in `schedulingGateRemoval`). If predicates pass, we can allocate and bind in the same round, **ensuring gate is removed before bind**.

- When we enqueue gate removal but don’t bind in that round (e.g. predicates fail), the worker removes the gate asynchronously; the next session then sees the pod without the gate in the cache and can schedule it (as long as we have available nodes for the tasks). So yes, after the gate is removed and the event is processed, one or more later session rounds are needed before the pod is scheduled.

docs/design/scheduling-gates-queue-admission.md
devzizu: I believe the current design should handle scheduler restarts correctly, but please correct me if I'm wrong. The `queueGateReservedTasks` cache is rebuilt on [every session](https://github.com/devzizu/volcano/blob/7426729bf0017a3c9dc6acfc48d5b1b02a68fedd/docs/design/scheduling-gates-queue-admission.md?plain=1#L467) start by scanning for pods that are [`Pending` and have the new queue allocation annotation](https://github.com/devzizu/volcano/blob/7426729bf0017a3c9dc6acfc48d5b1b02a68fedd/docs/design/scheduling-gates-queue-admission.md?plain=1#L482). In the case of a scheduler restart, it can just rebuild the cache accounting for non-gated pods as reservations.

docs/design/scheduling-gates-queue-admission.md
devzizu: I understand the concern about complexity, and I'm happy to schedule a meeting to discuss whether there's a simpler approach or if some of this logic could be generalized further for other plugins, for example.

Regarding your question around **removing tasks from reserved cache**: The reserved cache tracks tasks that passed the queue capacity check but aren't allocated yet, so we count their requests as reserved and don't over-allocate the queue. When we `commit` a statement, those allocations are final, we remove the tasks from the reserved cache at that moment so we don't count them as both reserved and allocated (**double-counting**). In gang scheduling we only commit when the gang condition is met; if we removed from the reserved cache before commit, we'd free capacity for other jobs even when our allocation **might still be rolled back**. For context, this approach came initially as a concern from [this comment](https://github.com/volcano-sh/volcano/pull/4727#discussion_r2694201872).

Let me know if I answered your question and the approach makes sense - because I could be missing additional context from the scheduler codebase and making the wrong assumptions.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+        - [Cache Initialization](#cache-initialization)
+        - [Cache Updates During Allocation](#cache-updates-during-allocation)
+- [Conclusions](#conclusions)
+- [Related Issues](#related-issues)
+
+## Motivation
+
+Cluster autoscalers such as [Cluster Autoscaler (CA)](https://github.com/kubernetes/autoscaler/tree/master) or
+[Karpenter](https://github.com/kubernetes-sigs/karpenter) are present in almost all Kubernetes environments to
+dynamically adjust node capacity based on scheduler signals. Both systems monitor pod scheduling conditions,
```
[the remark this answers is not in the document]
devzizu: Already addressed

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+        if task.SchGated && api.HasQueueAllocationGateAnnotation(task.Pod) {
+            klog.V(3).Infof("Task %s/%s has the QueueAllocationGate, queue async gate removal", task.Namespace, task.Name)
+            alloc.schedulingGateRemoval(task, queue.UID)
+        }
+
+		// Skip tasks with external (non-Volcano) scheduling gates
+		if task.SchGated {
+			// Tasks that contain the QueueAllocationGate have SchGated set to false by schedulingGateRemoval() above.
+			klog.V(4).Infof("Task %s/%s has non-Volcano gate, skipping", task.Namespace, task.Name)
+			continue
```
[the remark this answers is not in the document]
devzizu: Already addressed in https://github.com/volcano-sh/volcano/pull/4727/commits/311ecacd758fade8c84a50b6f1954b2b5863ef9d.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+
+		// Ensure Volcano QueueAllocationGate is removed before bind, otherwise the bind will fail.
+		// This is a safety guarantee as the async worker may have already removed it.
+		if schedulingapi.HasQueueAllocationGateAnnotation(p) && schedulingapi.HasOnlyVolcanoSchedulingGate(p) {
+			klog.V(3).Infof("Ensuring gate is removed for pod %s/%s before bind", p.Namespace, p.Name)
+			err := RemoveVolcanoSchGate(kubeClient, p)
+
+			// On conflict, verify gates are gone
+			if apierrors.IsConflict(err) {
+				freshPod, _ := kubeClient.CoreV1().Pods(p.Namespace).Get(context.TODO(), p.Name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
devzizu: This topic was already discussed and addressed, and I believe if we go with these details we'll be exposing too much details about the implementation that should be discussed in the implementation PR.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+    for _, task := range tasks {
+        p := task.Pod
+
+		// Ensure Volcano QueueAllocationGate is removed before bind, otherwise the bind will fail.
+		// This is a safety guarantee as the async worker may have already removed it.
+		if schedulingapi.HasQueueAllocationGateAnnotation(p) && schedulingapi.HasOnlyVolcanoSchedulingGate(p) {
+			klog.V(3).Infof("Ensuring gate is removed for pod %s/%s before bind", p.Namespace, p.Name)
+			err := RemoveVolcanoSchGate(kubeClient, p)
+
+			// On conflict, verify gates are gone
```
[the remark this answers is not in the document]
devzizu: I believe this is a concern for the implementation, which I'm considering, but not this proposal document.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+[`tdm`](https://github.com/volcano-sh/volcano/tree/v1.14.0/pkg/scheduler/plugins/tdm) to implement similar reservation
+cleanup logic in the future without modifying the framework.
+
+## Conclusions
+
+Scheduling Gates are a Kubernetes feature that allows external controllers to delay pod scheduling until specific
+conditions are met. The proposed design leverages this mechanism to defer scheduling until the queue has sufficient
+capacity, preventing pods from appearing as `Unschedulable` when they're simply waiting for queue admission and falsely
+trigger cluster scale-ups.
+
```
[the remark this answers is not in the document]
devzizu: If gate removal fails at bind, the pod stays Pending with the gate. The cache still has it from the informer, and the reserved cache is rebuilt every session from current pending tasks. So next cycle the task is reconsidered, can pass capacity again and be added to the reserved cache, and we retry allocation and bind (including gate removal). No special or extra “re-add to reserved cache” logic is required in this case.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+    queueGateReservedTasks map[api.QueueID]map[api.TaskID]*api.TaskInfo
+}
+```
+
+###### Capacity Check with Reserved Resources
+
+The `queueAllocatableWithReserved` method performs capacity checks including reserved resources:
+
+```go
+func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidate *api.TaskInfo, queue *api.QueueInfo) bool {
```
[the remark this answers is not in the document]
devzizu: This is actually intended design, but the concern is a valid point: an ungated-but-unschedulable pod can hold queue capacity and block other pods that could schedule. We could add a timeout or similar mechanism to release reservations for pods that stay unschedulable for a long time, but that would need a clear policy and might over-commit the queue when those pods eventually get nodes.

@JesseStutler @hajnalmt  – Do you think we should address this in the initial implementation (e.g. document the tradeoff and/or add a reservation timeout), or treat it as follow-up work once we see how it behaves in practice?

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+    }
+    return res
+}
+```
+
+This change allows
+[`DeductSchGatedResources()`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/api/job_info.go#L592) to
+treat Volcano-gated pods differently from other scheduling-gated pods, ensuring they remain visible to the enqueue
+action while still being prevented from allocation until the gate is removed.
+
```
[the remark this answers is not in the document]
devzizu: `GetSchGatedPodResources()` is only used by `DeductSchGatedResources()`, which is only used by the capacity, proportion, and overcommit plugins for inqueue accounting. For that we want Volcano-only-gated pods to count toward the job's queue usage (so they're not deducted). So the inqueue number can be a bit higher for jobs using the Volcano gate, but both plugins should still perform the same way (fair share and overcommit) with that updated inqueue value.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+```go
+func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
+    // Rebuild reserved cache for this scheduling cycle
+    cp.buildQueueReservedTasksCache(ssn)
+
+    ssn.AddCleanupReservationsFn(cp.Name(), ...)
+
+    // ... rest of capacity plugin initialization ...
+}
+
```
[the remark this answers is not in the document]
devzizu: Related with https://github.com/volcano-sh/volcano/pull/4727#discussion_r2786722922.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+        if _, exists := tasks[taskID]; exists {
+            delete(tasks, taskID)
+            // Clean up empty queue entries
+            if len(tasks) == 0 {
+                delete(cp.queueGateReservedTasks, queueID)
+            }
+            return
+        }
+    }
+}
```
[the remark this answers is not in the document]
devzizu: Already explained throughout the document:

- https://github.com/devzizu/volcano/blob/proposal-4710/docs/design/scheduling-gates-queue-admission.md?plain=1#L546-L560
- https://github.com/devzizu/volcano/blob/proposal-4710/docs/design/scheduling-gates-queue-admission.md?plain=1#L563
- https://github.com/devzizu/volcano/blob/proposal-4710/docs/design/scheduling-gates-queue-admission.md?plain=1#L614

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+```go
+const QueueAllocationGateKey = GroupName + "/queue-allocation-gate"
+```
+
+Besides, helper functions `api.HasOnlyVolcanoSchedulingGate()` and `api.HasQueueAllocationGateAnnotation()`, mentioned
+throughout the document, are used to identify pods that participate in this gate management mechanism. These functions
+can be defined in
+[`pkg/scheduler/api/helpers.go`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/api/helpers.go) to
+avoid code duplication and maintain consistency across the codebase.
+
```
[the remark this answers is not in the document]
devzizu: This is being addressed in the implementation PR: https://github.com/volcano-sh/volcano/pull/5033

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+###### Cache Updates During Allocation
+
+The cache is updated incrementally during the scheduling session through two mechanisms: **adding tasks when they pass
+capacity checks** and **removing tasks before statement commit**.
+
+**Adding Tasks to Reserved Cache**
+
+When a task passes the capacity check, the capacity plugin adds it to the reserved cache directly within its
+[`AddAllocatableFn`](https://github.com/volcano-sh/volcano/blob/v1.14.0/pkg/scheduler/plugins/capacity/capacity.go#L202)
+callback registered in
```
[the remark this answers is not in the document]
devzizu: The AddAllocatableFn callback no longer uses candidate.SchGated for adding to the reserved cache: https://github.com/devzizu/volcano/blob/proposal-4710/docs/design/scheduling-gates-queue-admission.md?plain=1#L238, and this was already fixed in the implementation as well: https://github.com/volcano-sh/volcano/pull/5033/commits/78deb602127cc9e6dd9a751d2aee95375b985d77

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,639 @@
...
+// In capacity plugin's addTaskToReservedCache
+func (cp *capacityPlugin) addTaskToReservedCache(queueID api.QueueID, task *api.TaskInfo) {
+    if cp.queueGateReservedTasks[queueID] == nil {
+        cp.queueGateReservedTasks[queueID] = make(map[api.TaskID]*api.TaskInfo)
+    }
+    cp.queueGateReservedTasks[queueID][task.UID] = task
+}
+
+// In capacity plugin's removeTaskFromReservedCache
+func (cp *capacityPlugin) removeTaskFromReservedCache(taskID api.TaskID) {
```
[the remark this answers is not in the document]
devzizu: I believe we already discussed this @hajnalmt as not a huge concern, but please correct me otherwise. Let me know if you believe I should consider a map-like structure for this data.

docs/design/scheduling-gates-queue-admission.md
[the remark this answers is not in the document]
hajnalmt: I think this is fine and this is the "standard" way of things happening. I am fine with it.

docs/design/scheduling-gates-queue-admission.md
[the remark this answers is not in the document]
hajnalmt: I think this is fine.

docs/design/scheduling-gates-queue-admission.md
hajnalmt: This is a good idea.

docs/design/scheduling-gates-queue-admission.md
hajnalmt: Let's schedule a meeting next week 👍 We can elaborate on this. It took me quite some time too until I managed to understand it, this is more of a general extension point before statement commit in allocate that plugins can use, for whatever they want. Later on we can implement this feature in proportion too.

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,664 @@
...
+function) by adding a new `schedulingGate` entry using the same constant:
+
+```go
+// Existing createPatch function needs to include the new optional patch
+func createPatch(pod *v1.Pod) ([]byte, error) {
+
+    // ...
+
+    // Add scheduling gates if opted-in
+    patchGates := patchSchedulingGates(pod)
```
JesseStutler, line 129: We can also add a new feature gate to control all the behavior, rather than set as the default behavior

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,664 @@
...
+		// If task passed allocation check and has the QueueAllocationGate, initiate async gate removal.
+		// Gate will be removed by the background worker (best effort). During the bind operation, we need
+		// to ensure the gate is not present, otherwise the bind will fail.
+        if task.SchGated && api.HasQueueAllocationGateAnnotation(task.Pod) {
+            klog.V(3).Infof("Task %s/%s has the QueueAllocationGate, queue async gate removal", task.Namespace, task.Name)
+            alloc.schedulingGateRemoval(task, queue.UID)
+        }
+
+		// Skip tasks with external (non-Volcano) scheduling gates
+		if task.SchGated {
```
JesseStutler, line 277: https://github.com/volcano-sh/volcano/blob/b7b2ba8ca88bc57735b2fd380f01dd4fbcfb7cd3/pkg/scheduler/actions/allocate/allocate.go#L253-L257 should skip volcano scheduling gate pods in `organizeJobWorksheet` rather than in `allocateResourcesForTasks` instead I think

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,664 @@
...
+    }
+
+    // ...
+}
+
+// schedulingGateRemoval queues async gate removal.
+func (alloc *Action) schedulingGateRemoval(task *api.TaskInfo, queueID api.QueueID) {
+    if api.HasOnlyVolcanoSchedulingGate(task.Pod) {
+        op := schGateRemovalOperation{namespace: task.Namespace, name: task.Name}
+        alloc.schGateRemovalCh <- op
```
JesseStutler: Should `schedulingGateRemoval` run in a separate goroutine? I'm worried that since `select` isn't used here and execution is synchronous, it might block the main scheduling process (if the worker processing isn't fast enough).
Also I think feature gates should also be used to control the `schedulingGateRemoval` part

docs/design/scheduling-gates-queue-admission.md
```diff
@@ -0,0 +1,664 @@
...
+```go
+// In DefaultBinder.Bind()
+func (db *DefaultBinder) Bind(kubeClient kubernetes.Interface, tasks []*schedulingapi.TaskInfo) map[schedulingapi.TaskID]string {
+    errMsg := make(map[schedulingapi.TaskID]string)
+    for _, task := range tasks {
+        p := task.Pod
+
+		// Ensure Volcano QueueAllocationGate is removed before bind, otherwise the bind will fail.
+		// This is a safety guarantee as the async worker may have already removed it.
+		if schedulingapi.HasQueueAllocationGateAnnotation(p) && schedulingapi.HasOnlyVolcanoSchedulingGate(p) {
```
JesseStutler, line 387: Because it's a synchronous operation, so I'd like to also add a unified scheduling gate in scheduler to control the remove gate behavior to prevent the influence on existing codes

docs/design/scheduling-gates-queue-admission.md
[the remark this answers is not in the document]
JesseStutler: I also have this consideration. How do we differ the ungated pod that can never be scheduled or that will be scheduled after the auto-scaler expands? We can not differ them. So the reserve resource will block other pods those can be scheduled to schedule, but I'm fine with current behavior, but should tell in the doc that there will such a scenario that will block other pods to get scheduled

docs/design/scheduling-gates-queue-admission.md
JesseStutler: Oh I see, so the removing action is to prevent double couting in both reserved and allocate, but I still feel weird about the name `AddCleanupReservationsFn`, this doesn't seem like a general-purpose behavior, and I'm hard to believe other plugins will extend it in the future. I'd rather expect that if we actually need to add a new extension point, it should be sufficiently general. But is modifying the statement operation a good approach? Wouldn't it be better to directly change from reserved to allocate inside `AllocateFunc`?

docs/design/scheduling-gates-queue-admission.md
devzizu: We already check if Pods have an [opt-in annotation](https://github.com/devzizu/volcano/blob/b467a0eb6e0311652060eeb7712e32093cecff4f/docs/design/scheduling-gates-queue-admission.md?plain=1#L140) for us to add the gate, so this is not applied automatically.

Can you clarify if your suggestion is to add a feature to [pkg/features/volcano_features.go](https://github.com/volcano-sh/volcano/blob/master/pkg/features/volcano_features.go) as well, plus the opt-in annotation that was already suggested?

**EDIT**: Added a feature gate section to the design https://github.com/volcano-sh/volcano/pull/4727/commits/5911eda6d72507ad72256393dc55a208fdd69d09.

docs/design/scheduling-gates-queue-admission.md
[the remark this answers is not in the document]
devzizu: Makes sense! I added this limitation to the documentation, thanks for the feedback!

https://github.com/volcano-sh/volcano/pull/4727/commits/bfc1d34374ce1925d3aa282c80f1715ab0de7f77

docs/design/scheduling-gates-queue-admission.md
devzizu: Ohh, my bad! I [ended up implementing](https://github.com/devzizu/volcano/blob/573980bb66edca2eea18a3d64ec96ed169b6ba52/pkg/scheduler/actions/allocate/allocate.go#L450-L458) with select to avoid the blocking behavior, but forgot to fix the design document. 

Thanks for pointing out this -- fixed in https://github.com/volcano-sh/volcano/pull/4727/commits/e8b9be4c2029f115b969bf3b97e4228490a29295.

And yes, we can control `schedulingGateRemoval` with the feature gates!

docs/design/scheduling-gates-queue-admission.md
devzizu: I believe what you are suggesting is to add a **feature gate** (and not "_scheduling gate_") as previously mentioned? If I got it right yes, I'll be adding it!

**EDIT**: Added a feature gate section to the design https://github.com/volcano-sh/volcano/pull/4727/commits/5911eda6d72507ad72256393dc55a208fdd69d09.

docs/design/scheduling-gates-queue-admission.md
devzizu: I've already done that in the [actual implementation](https://github.com/devzizu/volcano/blob/ft-proposal-4710/pkg/scheduler/actions/allocate/allocate.go#L347), but forgot to go into those details in the design doc. 

However, I'm also adding to the design doc as well to be more clear, thank you for the suggestion (please refer to this commit https://github.com/volcano-sh/volcano/pull/4727/commits/20c260176c2dc9b203cf4d65bde4c21083215d4b).

In this regard, you mention that we should skip in `organizeJobWorksheet` rather than `allocateResourcesForTasks`, but I believe we should do it in both places as they have different purposes:

- In `organizeJobWorksheet` we include volcano-gated pods so they can reach `allocateResourcesForTasks`, as you mention.
- In `allocateResourcesForTasks` it's more of an optimization to avoid no-op calls to `schedulingGateRemoval` by checking if the task is already ungated through the `task.SchGated` flag. 

WDYT?

docs/design/scheduling-gates-queue-admission.md
devzizu: `AddCleanupReservationsFn` may not be a good general extension point, and I'm happy to discuss a better approach at the community meeting.

On the `AllocateFunc` suggestion: the concern is that `AllocateFunc` fires per-task when the op is added to the statement, before `Commit()` is called. Correct me if I'm wrong, but for the  gang scheduling case, tasks are allocated one by one into the statement, but `Commit()` only happens if `ssn.JobReady(job)` returns `true` (_i.e._ `minAvailable` tasks are satisfied). If we remove a task from the reserved cache in `AllocateFunc` and the gang fails mid-way, the statement is discarded but we already freed capacity for some tasks. That opens a window where another job could take that capacity, even though our gang's allocation was rolled back.

The alternative approach would be to both play with `AllocateFunc` and `DeallocateFunc` to keep accounting correct, where in the latter we would need to re-add the task back to reserved. That re-addition is tricky because at that point we can't easily tell whether the task should go back to reserved or not. Let's discuss the trade-offs at the community meeting.

JesseStutler:
@devzizu Hi Pedro, sorry, I want to look back at why we chose to use scheduling gate instead of taking a new condition like adding an Unallocatable, instead of letting pods with insufficient queue resources continue to add an Unscheduable condition, because it feels like the intrusive modification of schedulingGate is a bit too much, even though we can control it with a switch. But it is inevitable that there will be more switch control code in the future, I am not sure if this is a better solution, sorry to look back at our original design at this time, but I suddenly thought that we have changed so much code, why not try to solve it from the condition? WDYT, however, it's not very friendly to your current job, it's just a suggestion of mine, and you're welcome to disagree with me at any time.

devzizu:
> @devzizu Hi Pedro, sorry, I want to look back at why we chose to use scheduling gate instead of taking a new condition like adding an Unallocatable, instead of letting pods with insufficient queue resources continue to add an Unscheduable condition, because it feels like the intrusive modification of schedulingGate is a bit too much, even though we can control it with a switch. But it is inevitable that there will be more switch control code in the future, I am not sure if this is a better solution, sorry to look back at our original design at this time, but I suddenly thought that we have changed so much code, why not try to solve it from the condition? WDYT, however, it's not very friendly to your current job, it's just a suggestion of mine, and you're welcome to disagree with me at any time.

IIRC, we did not considered following that path as it would introduce a new non-standard reason name in the condition (i.e., `Unallocatable` or others). I also don't how autoscalers would react to that, and specially if we would be changing the Unschedulable/Unallocatable back and forth depending on if the Pod passed allocatable check in each scheduling cycle. Besides, we would also need to reserve space right? I a Pod is Unschedulable and the queue can only fit one pod, we don't want other Pods to also pass the check. Maybe we'll not be saving much complexity here. And I also believe scheduling gates are more semantically correct as introduced in KEP-3521. I'm happy to follow the approach that makes more sense, so let's try to reach an agreement on the right path 🙏 

WDYT?

JesseStutler:
@devzizu Pedro I think currently I'm fine with current proposal, and we can go ahead with this design first. Could you squash your commits of this PR?

devzizu:
> @devzizu Pedro I think currently I'm fine with current proposal, and we can go ahead with this design first. Could you squash your commits of this PR?

Thank you! Just squashed to 4 commits

JesseStutler:
/lgtm
Thanks! Look forward to the code implementation

PR #4737, opened 2025-11-19
fix: support for aborted job state in garbage collector

### LINKED ISSUE #4736: gc currently ignores vcjobs in the aborted state
### Description

In vcjob, there is a field called ttlSecondsAfterFinished. The volcano-controller will delete this vcjob after the task runs for ttlSeconds. However, I noticed that it does not handle vcjobs in the Aborted state.

1. create a vcjob with ttlSecondsAfterFinished
2. make this vcjob into Aborted status
3. wait `ttlSecondsAfterFinished` seconds

The volcano controller did not delete this timed-out vcjob, expected normal deletion

v1.12.0

_No response_

NSBlink:
#### What type of PR is this?
This PR is the fix of #4736 

Fixes #4736 

```release-note
The vcjob that was aborted will be garbage collected normally
```

### DIFF
--- a/pkg/controllers/garbagecollector/garbagecollector.go
+++ b/pkg/controllers/garbagecollector/garbagecollector.go
@@ -271,7 +271,8 @@ func needsCleanup(j *v1alpha1.Job) bool {
 func isJobFinished(job *v1alpha1.Job) bool {
 	return job.Status.State.Phase == v1alpha1.Completed ||
 		job.Status.State.Phase == v1alpha1.Failed ||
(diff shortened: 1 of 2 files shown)


hajnalmt:
Hello @NSBlink !
Thank you for the contribution. The question here is that, do we consider *Aborted* jobs actually finished? I don't think they are. If this is an actual problem for somebody, we can maybe add a flag to enable this that would make more sense to me.

NSBlink:
I see, I misunderstood the Aborted status, thanks

Issue #4738, opened 2025-11-19
Feature Request: Allow dry-run scheduling for workloads


weapons97:
### What is the problem you're trying to solve

Sometimes, when scheduling workloads, we are not certain whether the requested resources can be successfully scheduled. If not, we may need to reduce the resource requirements.
With the dry-run scheduling feature, we can avoid repeated trial and error and enhance the interpretability of scheduling.

### Describe the solution you'd like

We can either provide an HTTP API via an independent dry-run service to accept dry-run requests, or implement the dry-run capability through a separate CRD.

### Documentation Updates

- [ ] This feature requires design or user documentation changes.
- [ ] If documentation changes are required, I will ensure the relevant documents are updated and published to the Volcano official website (https://volcano.sh) via the [volcano-sh/website](https://github.com/volcano-sh/website) repository.

PR #4740, opened 2025-11-20
Proposal: Allow dry-run scheduling for workloads

weapons97:
#### What type of PR is this?
Proposal

Target issue: [Feature Request: Allow dry-run scheduling for workloads](https://github.com/volcano-sh/volcano/issues/4738])

Sometimes, when scheduling workloads, we are not certain whether the requested resources can be successfully scheduled. If not, we may need to reduce the resource requirements.
With the dry-run scheduling feature, we can avoid repeated trial and error and enhance the interpretability of scheduling.

This proposal proposes an independent service with scheduling dry-run APIs to address the aforementioned needs.

### DIFF
--- a/docs/proposals/dryrun-service.md
+++ b/docs/proposals/dryrun-service.md
@@ -1,0 +1,171 @@
+# Volcano dryrun service
+@weapons97 2025/11/19
+
+## Introduction
+
+Target issue: [Feature Request: Allow dry-run scheduling for workloads](https://github.com/volcano-sh/volcano/issues/4738)
+
+Sometimes, when scheduling workloads, we are not certain whether the requested resources can be successfully scheduled. If not, we may need to reduce the resource requirements.
+With the dry-run scheduling feature, we can avoid repeated trial and error and enhance the interpretability of scheduling.
+
+This proposal proposes an independent service with scheduling dry-run APIs to address the aforementioned needs.
+
+### Goals
+
+- An independent service for dry-run scheduling
+
+## Proposal
+
+We can implement a service that provides dry-run scheduling APIs.
+
+The API are as follows:
+
+#### Dry-run scheduling
+- `POST /api/v1/scheduling/dryrun`
+
+request body like:
+
+- replicas: The number of replicas of the workload.
+- queue: The name of the queue where the workload is scheduled.
+- template: The template of the pod resources.
+  - spec: The spec of the pod resources.
+  - affinity: The affinity of the pod resources.
+  - containers: The containers of the pod resources.
+    - resources: The resources of the container.
+
+example:
+```json
+{
+  "replicas": 1,
+  "queue": "queue1",
+  "template": {
+    "spec": {
+      "affinity": {
+        "nodeAffinity": {
+          "requiredDuringSchedulingIgnoredDuringExecution": {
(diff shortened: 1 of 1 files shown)


docs/proposals/dryrun-service.md
```diff
@@ -0,0 +1,157 @@
...
+          "operator": "In",
+          "values": [
+            "Tesla-V100"
+          ]
+        }
+      ]
+    }
+  ],
+  "queue": "q1",
+  "pod": [
```
wangyang0616: Wouldn't it be better to inherit the pod's native template here, as it could reduce the user's usage cost?

docs/proposals/dryrun-service.md
```diff
@@ -0,0 +1,157 @@
...
+- An independent service for dry-run scheduling
+
+## Proposal
+
+We can implement a service that provides dry-run scheduling APIs.
+
+The API are as follows:
+
+#### Dry-run scheduling
+- `POST /api/v1/scheduling/dryrun`
```
wangyang0616, line 24: Is the API being provided externally via an Aggregation Layer?

docs/proposals/dryrun-service.md
weapons97: Yes, this is much better! I’ve updated the request body since "template" here refers to the pod template.

docs/proposals/dryrun-service.md
weapons97: My current understanding is that the Schedule-Inspector service provides interfaces and internally relies on the Volcano scheduler's packages to simulate the scheduling logic.

hzxuzhonghu: As we know volcano scheduler works with leader-workler standby mode. Do you require the leader insatnce handle the api? If so how can you know which one is the leader?

hzxuzhonghu (commented):
I would suggest using a constant annotation or CRD. From my perspective, annotation is much easier to start as a alpha feature., especially we may need to dry-run for many different kinds of workload.


The other confusion not covered by this proposal is how do you make the scheduler handles the dry-run scheduling, this is the critical part.

weapons97:
> I would suggest using a constant annotation or CRD. From my perspective, annotation is much easier to start as a alpha feature., especially we may need to dry-run for many different kinds of workload.
> 
> The other confusion not covered by this proposal is how do you make the scheduler handles the dry-run scheduling, this is the critical part.

Actually, we’re planning to develop a separate service to implement the proposal’s interfaces—so it’ll be totally independent from the current scheduler.

hzxuzhonghu:
@weapons97 Hmm, does it need to run the same scheduling algorithms as the core scheduler? I am curious how you plan to implement that service

weapons97:
> @weapons97 Hmm, does it need to run the same scheduling algorithms as the core scheduler? I am curious how you plan to implement that service

Basically, this service will simulate the process of allocate actions, and maybe I can write a demo first for everyone to take a look at.

weapons97:
> @weapons97 Hmm, does it need to run the same scheduling algorithms as the core scheduler? I am curious how you plan to implement that service

I've implemented the demo we discussed earlier, and the PR is [inspector](https://github.com/volcano-sh/volcano/pull/4783). Please checkout this PR.

Issue #4745, opened 2025-11-24
volcano-scheduler memory leak problem


fengruotj:
### Description

I found the scheduler memory increase problem in our large-scale cluster。Ultimately resulted in the container’s OOM killed.

![heap_inuse (1)](https://github.com/user-attachments/assets/b155488d-4eec-457c-942d-07a580d22f3d)

<img width="1206" height="343" alt="image" src="https://github.com/user-attachments/assets/5975ff0a-ff79-4dc8-adbb-e0f29795a8c9" />

<img width="3288" height="1166" alt="Image" src="https://github.com/user-attachments/assets/c954f88c-5f9b-412e-9853-b773bab20403" />

### Describe the results you received and expected

memory leak not happened

### What version of Volcano are you using?

master

fengruotj:
In the end, we resolved the issue through the following three optimizations:
**1. Introduced a soft memory limit using GOMEMLIMIT.**
The Go runtime enforces this limit by capping heap growth, adjusting GC frequency more aggressively, and returning memory back to the OS. This helps prevent excessive heap allocations that could exceed system memory limits and cause the process to be killed.
**2. Added a TTL mechanism to the scheduler’s Metrics.**
volcano-scheduler stores Kubernetes objects in a local cache during scheduling. When scheduling begins, it takes a snapshot of this cache and executes scheduling logic based on configured plugins, during which metrics are recorded. If the snapshot and the local cache become inconsistent, metrics can leak. The TTL feature prevents this.
**3. Reduced memory overhead in the client-go workqueue channel.**
We now invoke AddRateLimited only for retry items and changed queue elements from pointers to string keys, reducing duplicate data stored in the workqueue channel and lowering memory usage.

fengruotj:
<img width="3674" height="1190" alt="Image" src="https://github.com/user-attachments/assets/73904202-79c5-449f-b8cf-99af3b72a2a2" />

hzxuzhonghu:
> 1. Introduced a soft memory limit using GOMEMLIMIT.

This is a good option we can use

> 2. Added a TTL mechanism to the scheduler’s Metrics. 

Which metric volcano expose has unbounded memory usage?

>3. Reduced memory overhead in the client-go workqueue channel. We now invoke AddRateLimited only for retry items and changed queue elements from pointers to string keys, reducing duplicate data stored in the workqueue channel and lowering memory usage.

I think this is also a good idea

JesseStutler:
[...]

PR #4749, opened 2025-11-24
Use GOMEMLIMIT for memory soft limit.

fengruotj:
#### What type of PR is this?
/kind feature

**Introduced a soft memory limit using GOMEMLIMIT.**

The Go runtime enforces this limit by capping heap growth, adjusting GC frequency more aggressively, and returning memory back to the OS. This helps prevent excessive heap allocations that could exceed system memory limits and cause the process to be killed.

NONE 

NONE

```release-note

```

### DIFF
--- a/installer/helm/chart/volcano/templates/admission.yaml
+++ b/installer/helm/chart/volcano/templates/admission.yaml
@@ -163,6 +163,14 @@ spec:
               readOnly: true
             - mountPath: /admission.local.config/configmap
               name: admission-config
+          {{- if .Values.custom.go_memlimit_enable }}
+          env:
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
+                  divisor: "1"
+          {{- end }}
           {{- if $admission_main_csc }}
           securityContext:
             {{- toYaml $admission_main_csc | nindent 12 }}
--- a/installer/helm/chart/volcano/templates/controllers.yaml
+++ b/installer/helm/chart/volcano/templates/controllers.yaml
@@ -206,6 +206,13 @@ spec:
             value: {{ .Release.Namespace }}
           - name: HELM_RELEASE_NAME
             value: {{ .Release.Name }}
+          {{- if .Values.custom.go_memlimit_enable }}
+          - name: GOMEMLIMIT
+            valueFrom:
+              resourceFieldRef:
+                resource: limits.memory
+                divisor: "1"
+          {{- end }}
 ---
 apiVersion: v1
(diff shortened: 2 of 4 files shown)


cmd/controller-manager/main.go
```diff
@@ -45,12 +47,29 @@ import (
...
+func init() {
 	runtime.GOMAXPROCS(runtime.NumCPU())
 	klog.InitFlags(nil)
 
+	memlimit.SetGoMemLimitWithOpts(
+		memlimit.WithRatio(DefaultMemLimitRatio),
+		memlimit.WithProvider(
+			memlimit.ApplyFallback(
+				memlimit.FromCgroup,
+				memlimit.FromSystem,
```
hzxuzhonghu: does this require cgroup permission. 

I am thinking maybe we should just set Env `GOMEMLIMIT` with downward api from `limitts.memory`

fengruotj: Yes, it  need open `/proc/self/cgroup` file.

Let me think, we can use `GOMEMLIMIT` like this:

``` yaml
env:
- name: GOMEMLIMIT
  value: "67108864"     // <--- limit on Go runtime
-------------- ## Or we can do it in the following way
- name: GOMEMLIMIT
  valueFrom:            // <--- value taken from downwards API
    resourceFieldRef:
      resource: limits.memory
```

It difficult to set the ratio, and using environment variables makes it difficult to handle cases where the limit is not configured.

hzxuzhonghu: Right, i mean this

hzxuzhonghu: Without setting limit, k8s treat it as no limit. That is consistent with the default behavior.
I donot think we should upgrade volcano permission

fengruotj: OK, I think you are right. I flollow you, also test the cases where the limit is not configured.

installer/helm/chart/volcano/templates/admission.yaml
```diff
@@ -163,6 +163,12 @@ spec:
               readOnly: true
             - mountPath: /admission.local.config/configmap
               name: admission-config
+          env:
```
JesseStutler, line 167: Can you add a switch to control the onion of these parameters, which are turned off by default? Then we can choose to enable it through helm installation

fengruotj: OK, I'have add the option.

hzxuzhonghu (approved):

JesseStutler:
/lgtm
Thanks a lot!

PR #4754, opened 2025-11-26
feat: add errtasks rate limited for retryResync task.

fengruotj:
#### What type of PR is this?
/kind feature

**Reduced memory overhead in the client-go workqueue channel.**

We changed queue elements from pointers to string keys and increase the rateLimiter for deletedJob.

We now invoke AddRateLimited only for retry items, reducing duplicate data stored in the workqueue channel and lowering memory usage. When a large number of items are added to the workqueue using AddRateLimited, client-go will apply rate limiting. If rate limiting is triggered, the items are placed into internal Go channels, which can result in additional memory overhead in Go.

NONE 

NONE

```release-note

```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1092,6 +1092,12 @@ func (sc *SchedulerCache) processCleanupJob() {
 
 func (sc *SchedulerCache) resyncTask(task *schedulingapi.TaskInfo) {
 	key := sc.generateErrTaskKey(task)
+	sc.errTasks.Add(key)
+}
+
+func (sc *SchedulerCache) retryResyncTask(task *schedulingapi.TaskInfo) {
+	klog.V(3).Infof("Retry to resync task <%v:%v/%v>", task.UID, task.Namespace, task.Name)
+	key := sc.generateErrTaskKey(task)
 	sc.errTasks.AddRateLimited(key)
 }
 
@@ -1147,7 +1153,7 @@ func (sc *SchedulerCache) processResyncTask() {
 	reSynced := false
 	if err := sc.syncTask(task); err != nil {
 		klog.ErrorS(err, "Failed to sync task, retry it", "namespace", task.Namespace, "name", task.Name)
-		sc.resyncTask(task)
+		sc.retryResyncTask(task)
 		reSynced = true
 	} else {
 		klog.V(4).Infof("Successfully synced task <%s/%s>", task.Namespace, task.Name)


pkg/scheduler/cache/cache.go
```diff
@@ -1095,6 +1095,12 @@ func (sc *SchedulerCache) resyncTask(task *schedulingapi.TaskInfo) {
 	sc.errTasks.AddRateLimited(key)
 }
 
+func (sc *SchedulerCache) retryResyncTask(task *schedulingapi.TaskInfo) {
```
hzxuzhonghu, line 1098: seems it is duplicate with `resyncTask`, i cannot see any diff expect one logging

fengruotj: > seems it is duplicate with `resyncTask`, i cannot see any diff expect one logging

Sorry, I missed one point earlier. The main goal is still to avoid throttling, because the rate limit delay in the production environment is significant.

hzxuzhonghu (commented):
I mean these are two same methods

fengruotj:
> I mean these are two same methods

emmm, I see. I’ve already fixed it. In the resyncTask method, it was previously using AddRateLimited, and I changed it to Add. If we reach a consensus on this point, I may need to reconsider whether it’s necessary. I think errTasks wil need to execute add function when it firstly come in.

pkg/scheduler/cache/cache.go
hajnalmt: You can at least reuse the earlier function, and please decrease the verbosity:

```
func (sc *SchedulerCache) retryResyncTask(task *schedulingapi.TaskInfo) {
	klog.V(4).Infof("Retry to resync task <%v:%v/%v>", task.UID, task.Namespace, task.Name)
	sc.resyncTask(task)
}
```

hzxuzhonghu:
IMO, i feel it should not directly add err task

pkg/scheduler/cache/cache.go
```diff
@@ -1091,6 +1091,12 @@ func (sc *SchedulerCache) processCleanupJob() {
 }
 
 func (sc *SchedulerCache) resyncTask(task *schedulingapi.TaskInfo) {
+	key := sc.generateErrTaskKey(task)
+	sc.errTasks.Add(key)
+}
+
+func (sc *SchedulerCache) retryResyncTask(task *schedulingapi.TaskInfo) {
+	klog.V(3).Infof("Retry to resync task <%v:%v/%v>", task.UID, task.Namespace, task.Name)
```
hajnalmt, line 1099: We will print out two lines here.
One with
`klog.ErrorS(err, "Failed to sync task, retry it", "namespace", task.Namespace, "name", task.Name)`
and this one.
This one has more info, so it can stay, but can we increase the verbosity to
4 or 5 is even better?

hzxuzhonghu:
@hajnalmt Good analysis, maybe you can take over this task

hzxuzhonghu:
```
func (sc *SchedulerCache) syncTask(oldTask *schedulingapi.TaskInfo) error {
	newPod, err := sc.kubeClient.CoreV1().Pods(oldTask.Namespace).Get(context.TODO(), oldTask.Name, metav1.GetOptions{})
	
```

I think we can and should get rid of calling kubeapiserver directly here

hajnalmt:
Yes, definitely we should rely on the informer here.
I raised a PR for this, I hope it's not a problem @fengruotj I have kept your change in the tree and it will appear in the Release Notes too.
https://github.com/volcano-sh/volcano/pull/4938

hajnalmt:
/close

In favor of this:
https://github.com/volcano-sh/volcano/pull/4938 
Thank you for the contribution @fengruotj !

PR #4760, opened 2025-11-28
support ttl metrics for volcano-scheduler.

### LINKED ISSUE #4745: volcano-scheduler memory leak problem

fengruotj:
#### What type of PR is this?
/kind feature

**Added a TTL mechanism to the scheduler’s Metrics.**

volcano-scheduler stores Kubernetes objects in a local cache during scheduling. When scheduling begins, it takes a snapshot of this cache and executes scheduling logic based on configured plugins, during which metrics are recorded. If the snapshot and the local cache become inconsistent, metrics can leak. The TTL feature prevents this.

NONE 

NONE

```release-note

```

### DIFF
--- a/cmd/scheduler/app/server.go
+++ b/cmd/scheduler/app/server.go
@@ -94,6 +94,8 @@ func Run(opt *options.ServerOption) error {
 	}
 
 	ctx := signals.SetupSignalContext()
+	metrics.InitTTLQueueMetrics(ctx)
+	metrics.InitTTLJobMetrics(ctx)
 	run := func(ctx context.Context) {
 		sched.Run(ctx.Done())
 		<-ctx.Done()
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -22,6 +22,7 @@ limitations under the License.
 package allocate
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"testing"
@@ -44,6 +45,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/plugins/binpack"
 	"volcano.sh/volcano/pkg/scheduler/plugins/drf"
 	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
@@ -57,6 +59,9 @@ import (
 
 func TestMain(m *testing.M) {
 	options.Default()
+	metrics.InitTTLQueueMetrics(context.Background())
+	metrics.InitTTLJobMetrics(context.Background())
+
 	os.Exit(m.Run())
 }
 
--- a/pkg/scheduler/actions/backfill/backfill_test.go
+++ b/pkg/scheduler/actions/backfill/backfill_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package backfill
 
 import (
+	"context"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -31,10 +33,18 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/cache"
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/plugins/drf"
 	"volcano.sh/volcano/pkg/scheduler/plugins/priority"
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
+
+func TestMain(m *testing.M) {
+	metrics.InitTTLQueueMetrics(context.Background())
+	metrics.InitTTLJobMetrics(context.Background())
+
+	os.Exit(m.Run())
+}
 
 func TestPickUpPendingTasks(t *testing.T) {
 	framework.RegisterPluginBuilder("priority", priority.New)
--- a/pkg/scheduler/actions/enqueue/enqueue_test.go
+++ b/pkg/scheduler/actions/enqueue/enqueue_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package enqueue
 
 import (
+	"context"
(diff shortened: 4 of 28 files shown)


hzxuzhonghu:
Hope i can have some bandwidth to review today

hzxuzhonghu:
I just see DeleteJobMetrics is called after each job is deleted. Can you explain why there is still mem leak?

```

// DeleteJobMetrics delete all metrics related to the job
func DeleteJobMetrics(jobName, queue, namespace string) {
	e2eJobSchedulingDuration.DeleteLabelValues(jobName, queue, namespace)
	e2eJobSchedulingStartTime.DeleteLabelValues(jobName, queue, namespace)
	e2eJobSchedulingLastTime.DeleteLabelValues(jobName, queue, namespace)
	unscheduleTaskCount.DeleteLabelValues(jobName)
	jobShare.DeleteLabelValues(namespace, jobName)
	jobRetryCount.DeleteLabelValues(jobName)
}

```

fengruotj:
> I just see DeleteJobMetrics is called after each job is deleted. Can you explain why there is still mem leak?
> 
> ```
> 
> // DeleteJobMetrics delete all metrics related to the job
> func DeleteJobMetrics(jobName, queue, namespace string) {
> 	e2eJobSchedulingDuration.DeleteLabelValues(jobName, queue, namespace)
> 	e2eJobSchedulingStartTime.DeleteLabelValues(jobName, queue, namespace)
> 	e2eJobSchedulingLastTime.DeleteLabelValues(jobName, queue, namespace)
> 	unscheduleTaskCount.DeleteLabelValues(jobName)
> 	jobShare.DeleteLabelValues(namespace, jobName)
> 	jobRetryCount.DeleteLabelValues(jobName)
> }
> ```

In fact, the internal Jobs data structure used by Volcano for session scheduling is a snapshot of the current scheduler cache. During the scheduling process, if a job happens to be deleted, it will cause the corresponding job_name metrics tag to remain in prometheus metric.

hzxuzhonghu:
>In fact, the internal Jobs data structure used by Volcano for session scheduling is a snapshot of the current scheduler cache. 

Can you elaborate it a little bit more? How is this relaated to metrics?

fengruotj:
> > In fact, the internal Jobs data structure used by Volcano for session scheduling is a snapshot of the current scheduler cache.
> 
> Can you elaborate it a little bit more? How is this relaated to metrics?

OK, just using queue metrics as an example. When the scheduler execute the openSession function  in `pkg/scheduler/framework/session.go:163`,  it will create an snapshot cache by using cache.Snapshot(). So the ssn data structure using an snapshot data. It initialized Jobs and Queues from cache snapshot.  Then the ssn execute schedule process, the capacity plugin expose the queue metrics by using `pkg/scheduler/plugins/capacity/capacity.go:325`. At the same time before it expose the queue metrics, the queue has beee closed & deleted in`pkg/scheduler/cache/event_handlers.go:1000`. So the queue  
metrics has already cleaned. Ok, go back for the  capacity plugin, it then exposed the queue metrics. Finally the metrics will not deleted any more. So the same as job metrics.

hzxuzhonghu:
@fengruotj Thanks for your explanation. I think i get it. The pkg/scheduler/cache/event_handlers.go:1000 handle queue immediately, but the scheduling procedure `pkg/scheduler/plugins/capacity/capacity.go`  use a snapshot of the cache(record metrics here). So there could be a delay and race here.

But i see your solution seems too tricky. I am thinking about we introduce a delay deletion, this way is very simple without changing the metrics wrapper introduced here.  We can add a delay queue, whenever a queue deleted, we add it to a delay queue `NewTypedDelayingQueueWithConfig`, using `AddAfter`. And startup a separate consumer read the queue and delete metrics.

fengruotj:
> @fengruotj Thanks for your explanation. I think i get it. The pkg/scheduler/cache/event_handlers.go:1000 handle queue immediately, but the scheduling procedure `pkg/scheduler/plugins/capacity/capacity.go` use a snapshot of the cache(record metrics here). So there could be a delay and race here.
> 
> But i see your solution seems too tricky. I am thinking about we introduce a delay deletion, this way is very simple without changing the metrics wrapper introduced here. We can add a delay queue, whenever a queue deleted, we add it to a delay queue `NewTypedDelayingQueueWithConfig`, using `AddAfter`. And startup a separate consumer read the queue and delete metrics.

Got, I see. Your solutions sounds like a good idea.

hajnalmt:
/close

Thank you for the contribution @fengruotj !
This seems too tricky in the end, let's close this in favor of:
https://github.com/volcano-sh/volcano/pull/4822

Maybe we should create a delay deletion feature for the metrics, but it's not that important now I think if the memory leak problems disappear.

fengruotj:
> /close
> 
> Thank you for the contribution @fengruotj ! This seems too tricky in the end, let's close this in favor of: #4822
> 
> Maybe we should create a delay deletion feature for the metrics, but it's not that important now I think if the memory leak problems disappear.

Can I continue with this task using this approach? https://github.com/volcano-sh/volcano/pull/4822 However, I’m a bit concerned that this approach may confuse readers of the source code. It may not be obvious why this check is necessary, or why we need to fetch the data state from the underlying cache. 

I’d be very happy to contribute to the Volcano community.

fengruotj:
> > /close
> > Thank you for the contribution @fengruotj ! This seems too tricky in the end, let's close this in favor of: #4822
> > Maybe we should create a delay deletion feature for the metrics, but it's not that important now I think if the memory leak problems disappear.
> 
> Can I continue with this task using this approach? #4822 However, I’m a bit concerned that this approach may confuse readers of the source code. It may not be obvious why this check is necessary, or why we need to fetch the data state from the underlying cache.
> 
> I’d be very happy to contribute to the Volcano community.

After thinking it through, I still feel this metrics leakage issue should be handled at the upper framework layer, instead of requiring each user-implemented plugin to detect and work around it.

Following @hzxuzhonghu's suggestion, I implemented a delayed metrics deletion approach. Could you please help review the code: https://github.com/volcano-sh/volcano/pull/5618

fengruotj:
> /close
> 
> Thank you for the contribution @fengruotj ! This seems too tricky in the end, let's close this in favor of: #4822
> 
> Maybe we should create a delay deletion feature for the metrics, but it's not that important now I think if the memory leak problems disappear.

Queue metrics have the same issue.

PR #4777, opened 2025-12-02
add sharding controllers to support node partitioning among multiple scheduler

### LINKED ISSUE #4722: [Enhancement] Support fast scheduling for AI Agent Workloads

Agent workloads are latency-sensitive, it requires the scheduler to make ultra-fast scheduling for a large number of Agent tasks, ensuring high throughput and a low latency scheduling for each task. Volcano scheduler provides batch scheduling and advanced scheduling strategies for big data, or AI workloads. However, it lacks sufficient support and optimization for Agent workloads, particularly in terms of fast scheduling and scheduling strategy optimization. Therefore, the goal is to extend the Volcano scheduler to provide fast scheduling capabilities and scheduling strategies tailored to the characteristics of Agent workloads.

[agentcube subproject](https://github.com/volcano-sh/agentcube)

Proposal: Create a fast scheduler tailored for Agent workloads,  collaborating with the Volcano scheduler to provide capabilities such as fast scheduling, resource utilization optimization, and scheduling strategy optimization, to better meet the scheduling requirements of Agent workloads. By optimizing and streamlining the scheduling process and applying Agent-specific scheduling strategies, the agent scheduler can improve Agent task scheduling efficiency and reduce task startup latency.

When Agent workloads coexist with other workloads, the Agent scheduler can be seamlessly integrated with the Volcano scheduler, scheduling different workloads to node-level shard across the cluster, to guarantee scheduling efficiency and improve cluster-wide resource utilization. 

Nodes are assigned to different shards based on various strategies and are dynamically adjusted according to the cluster state,  improving utilization of node resources by different types of workloads. 

### Proposal Details
The full proposal details is at:  https://docs.google.com/document/d/1NDTEGulhQ__sdZ_c2kN3ojtJIZie5O7_GyZyKIOLz-g/edit?usp=sharing

### Additional context

Please feel free the share your thoughts and ideas

ssfffss:
#### What type of PR is this?
feature: this PR implements sharding controllers for dynamic node partitioning between schedulers

This PR introduces a new Sharding Controller to Volcano that dynamically partitions cluster nodes between different scheduler types (particularly for Agentic AI workloads). The controller:
1. Automatically calculates node resource utilization from pod requests (no external metrics server needed)
2. Assigns nodes to specialized schedulers based on configurable utilization thresholds and warmup preferences
3. Uses event-driven architecture with informer caches to minimize API server load
4. Provides hard filtering strategy for reliable node assignment with mutual exclusivity guarantees
5. Supports configurable scheduler policies through CLI flags
6. Includes comprehensive testing suite with performance validation for large clusters
This enables Volcano to efficiently support both traditional batch workloads and emerging Agentic AI workloads on the same cluster, improving resource utilization and scheduling performance.

Fixes #4722

This PR depends on the NodeShard CRD implementation in volcano-sh/apis
1. All tests pass in CI pipeline including large-scale performance tests (100-node cluster sync in <200ms)
2. Metrics calculation uses existing informer caches - no additional API server load
3. Controller follows Volcano's existing controller pattern with proper queue management and error handling
4. Default configuration includes two scheduler types (volcano-scheduler and agent-scheduler) but is fully configurable
5. Documentation has been updated with design doc and user guide

```release-note
Add new Sharding Controller that dynamically partitions cluster nodes between different scheduler types. This enables efficient scheduling of both traditional batch workloads and Agentic AI workloads on the same cluster.
```

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -44,7 +44,7 @@ const (
 	defaultPodGroupWorkers     = 5
 	defaultQueueWorkers        = 5
 	defaultGCWorkers           = 1
-	defaultControllers         = "-sharding-controller,*"
+	defaultControllers         = "*,-sharding-controller"
 )
 
 // ServerOption is the main context object for the controllers.
@@ -157,12 +157,9 @@ func (s *ServerOption) CheckOptionOrDie() error {
 // checkControllers checks the controllers option and returns error if it's invalid
 func (s *ServerOption) checkControllers() error {
 	existenceMap := make(map[string]bool)
-	for idx, c := range s.Controllers {
+	for _, c := range s.Controllers {
 		if c == "*" {
-			// wildcard '*' is not allowed to be combined with other input
-			if len(s.Controllers) > 1 && idx != len(s.Controllers)-1 {
-				return fmt.Errorf("wildcard '*' can only be placed at the final position when combined with other input")
-			}
+			continue
 		} else {
 			if strings.HasPrefix(c, "-") || strings.HasPrefix(c, "+") {
 				if existenceMap[c[1:]] {
--- a/docs/design/sharding_controller.md
+++ b/docs/design/sharding_controller.md
@@ -18,13 +18,13 @@ The Sharding Controller is proposed to addresses these challenges by:
 
 ## 2. Design Goals
 
-1. **Resource-Aware Partitioning**: Dynamically partition nodes based on real-time resource allocation/utilization;
+1. **Resource-Aware Partitioning**: Dynamically partition nodes based on real-time resource allocation/utilization; Currently we only support CPU utilization and will support multi-dimensional resources in the future;
 2. **Minimal API Server Load**: Use informer caches instead of direct API calls, reducing pressure on API servers;
 3. **Event-Driven Updates**: React to cluster state changes efficiently, such as pod events(add, delete, update) and node events (add, delete, update);
 4. **Scalability**: Support clusters with millions of nodes with multiple schedulers;
-5. **Resilience**: Recover gracefully from failures and restarts
+5. **Resilience**: Recover gracefully from failures and restarts;
 6. **Configurability**: Allow runtime configuration of scheduler shards, and hyper-parameters of sharding strategies;
-7. **Backward Compatibility**: Work with existing Volcano deployment without disruption
+7. **Backward Compatibility**: Work with existing Volcano deployment without disruption;
 
 ## 3. Architecture Overview
 ![sharding-controller](./images/sharding-controller-design.png)
@@ -93,7 +93,7 @@ graph LR
 The core logic that implements node assignment strategies:
 
 **Scheduler Configurations**
-
+In the scheduler configuration, we currently we only support CPU utilization and will support multi-dimensional resources in the future
 ```go
 type SchedulerConfigSpec struct {
 	Name              string   // Scheduler name
@@ -181,7 +181,7 @@ cpuUtilization := totalCPURequest / nodeCPUCapacity
 ![sharding-controller](./images/sharding-process.png)
 Unlike complex scoring systems, we use utilization-based filtering:
 1. For each scheduler, its shard assignments should consider nodes with following constraints:
-   - CPU utilization range [Min, Max];
+   - CPU utilization range [Min, Max]; (currently we only support CPU utilization and will support multi-dimensional resources in the future)
    - Warmup node preference: true or false
    - Min/Max node constraints;
 2. Filter nodes matching utilization constraints;
--- a/pkg/controllers/sharding/sharding_controller.go
+++ b/pkg/controllers/sharding/sharding_controller.go
@@ -42,12 +42,10 @@ import (
 
 const (
 	controllerName              = "sharding-controller"
-	defaultShardSyncPeriod      = 60 * time.Second
+	defaultShardSyncPeriod      = 120 * time.Second
 	maxAssignmentCacheRetention = 5 * time.Minute
-	nodeCountChangeThreshold    = 0.5
-	updateTimeoutThreshold      = 15 * time.Second
 	nodeUsageChangeThreshold    = 0.5
-	nodeRefreshPeriod           = 120 * time.Second
+	nodeRefreshPeriod           = 80 * time.Second
 )
 
 func init() {
@@ -86,9 +84,6 @@ type ShardingController struct {
 	// assignment cache
 	assignmentCache *AssignmentCache
 	cacheMutex      sync.Mutex
-
-	// Event channels
-	assignmentChangeChan chan *AssignmentChangeEvent
 }
 
 // Return the name of the controller
@@ -134,7 +129,6 @@ func (sc *ShardingController) Initialize(opt *framework.ControllerOption) error 
 	sc.assignmentCache = &AssignmentCache{
 		Assignments: make(map[string]*ShardAssignment),
 	}
-	sc.assignmentChangeChan = make(chan *AssignmentChangeEvent, 100)
 
 	// Parse scheduler configs from options
 	sc.parseSchedulerConfigsFromOptions()
@@ -205,9 +199,6 @@ func (sc *ShardingController) Run(stopCh <-chan struct{}) {
 
 	// Start periodic sync
 	go wait.Until(sc.syncShards, sc.shardSyncPeriod, stopCh)
-
-	// Start assignment change processor
-	go sc.assignmentChangeProcessor(stopCh)
 
 	// Start periodic metrics refresh
 	go wait.Until(sc.refreshNodeMetrics, nodeRefreshPeriod, stopCh)
@@ -308,33 +299,6 @@ func (sc *ShardingController) processNextItem() bool {
 	return true
 }
 
-// assignmentChangeProcessor processes assignment changes
-func (sc *ShardingController) assignmentChangeProcessor(stopCh <-chan struct{}) {
-	for {
-		select {
-		case event := <-sc.assignmentChangeChan:
-			sc.processAssignmentChange(event)
-		case <-stopCh:
-			return
-		}
-	}
-}
-
-// processAssignmentChange processes an assignment change event
-func (sc *ShardingController) processAssignmentChange(event *AssignmentChangeEvent) {
-	klog.V(4).Infof("Processing assignment change for %s: %d -> %d nodes",
-		event.SchedulerName, len(event.OldNodes), len(event.NewNodes))
-
-	// Log significant changes
-	if len(event.OldNodes) > 0 {
-		changePercent := float64(abs(len(event.NewNodes)-len(event.OldNodes))) / float64(len(event.OldNodes))
-		if changePercent > nodeCountChangeThreshold {
-			klog.Infof("Significant node change for %s: %.0f%% (%d -> %d nodes)",
-				event.SchedulerName, changePercent*100, len(event.OldNodes), len(event.NewNodes))
-		}
-	}
-}
-
 // syncShards performs global shard assignment calculation
 func (sc *ShardingController) syncShards() {
 	klog.Infof("Starting global shard synchronization")
@@ -343,9 +307,6 @@ func (sc *ShardingController) syncShards() {
 		duration := time.Since(startTime)
 		klog.V(3).Infof("Completed global shard synchronization in %v", duration)
 	}()
-
-	// ensure all nodes states are updated
-	//sc.ensureNodeStatesUpdated()
 
 	// Get nodes from cache (not API server)
 	nodes, err := sc.listNodesFromCache()
@@ -503,16 +464,7 @@ func (sc *ShardingController) applyAssignment(schedulerName string, assignment *
 		return fmt.Errorf("failed to update shard %s: %v", schedulerName, err)
 	}
 
-	// Publish change event
-	sc.assignmentChangeChan <- &AssignmentChangeEvent{
-		SchedulerName: schedulerName,
-		OldNodes:      shard.Spec.NodesDesired,
-		NewNodes:      assignment.NodesDesired,
-		NodesToAdd:    nodesToAdd,
-		NodesToRemove: nodesToRemove,
-		Version:       assignment.Version,
-		Timestamp:     time.Now(),
-	}
+	klog.V(6).Infof("Assignment changed for %s: %d -> %d nodes", schedulerName, len(shard.Spec.NodesDesired), len(assignment.NodesDesired))
 
 	klog.V(6).Infof("Updated shard %s with %d nodes", schedulerName, len(assignment.NodesDesired))
 	return nil
--- a/cmd/controller-manager/app/options/options_test.go
+++ b/cmd/controller-manager/app/options/options_test.go
@@ -105,7 +105,7 @@ func TestAddFlags(t *testing.T) {
 		WorkerThreadsForPG:    5,
 		WorkerThreadsForQueue: 5,
 		WorkerThreadsForGC:    1,
-		Controllers:           strings.Split("-sharding-controller,*", ","),
+		Controllers:           strings.Split("*,-sharding-controller", ","),
 	}
 	expectedFeatureGates := map[featuregate.Feature]bool{features.ResourceTopology: false}
 
@@ -151,6 +151,20 @@ func TestCheckControllers(t *testing.T) {
 			expectErr: nil,
 		},
 		{
+			name: "normal case: use * at the first place when combined with other input",
+			serverOption: &ServerOption{
+				Controllers: []string{"*", "-job-controller"},
+			},
+			expectErr: nil,
+		},
+		{
+			name: "normal case: use * at the middle place when combined with other input",
+			serverOption: &ServerOption{
+				Controllers: []string{"-sharding-controller", "*", "-job-controller"},
+			},
+			expectErr: nil,
+		},
+		{
 			name: "fail case: use duplicate job-controller",
 			serverOption: &ServerOption{
 				Controllers: []string{"+gc-controller", "+job-controller", "-job-controller"},
@@ -163,13 +177,6 @@ func TestCheckControllers(t *testing.T) {
 				Controllers: []string{"+job-controller", "job-controller"},
 			},
 			expectErr: fmt.Errorf("controllers option %s cannot have both '-' and '+' prefixes", "job-controller"),
-		},
-		{
-			name: "fail case: use * at the first place when combined with other input",
-			serverOption: &ServerOption{
-				Controllers: []string{"*", "+job-controller"},
-			},
-			expectErr: fmt.Errorf("wildcard '*' can only be placed at the final position when combined with other input"),
 		},
 	}
 
--- a/cmd/controller-manager/app/server_test.go
+++ b/cmd/controller-manager/app/server_test.go
@@ -52,6 +52,18 @@ func TestIsControllerEnabled(t *testing.T) {
 			isEnable:          true,
 		},
 		{
+			name:              "all controller should be enabled except job controller",
+			gotControllerName: "job-controller",
+			inputControllers:  []string{"*, -job-controller"},
+			isEnable:          false,
+		},
+		{
+			name:              "all controller should be enabled except job controller",
+			gotControllerName: "sharding-controller",
+			inputControllers:  []string{"-sharding-controller, *, -job-controller"},
+			isEnable:          false,
+		},
+		{
 			name:              "gc-controller should be disable, input allow jobtemplate-controller, jobflow-controller, pg-controller, queue-controller",
 			gotControllerName: "gc-controller",
 			inputControllers:  []string{"-gc-controller", "+jobtemplate-controller", "+jobflow-controller", "+pg-controller", "+queue-controller"},
--- a/pkg/controllers/sharding/test_utils.go
+++ b/pkg/controllers/sharding/test_utils.go
@@ -294,12 +294,6 @@ func CleanupController(testCtrl *TestShardingController) {
 			testCtrl.Controller.nodeEventQueue.Done(key)
 		}
 		klog.Infof("Processed remaining items in queues")
-		// Close assignment change channel if not nil
-		if testCtrl.Controller.assignmentChangeChan != nil {
-			close(testCtrl.Controller.assignmentChangeChan)
-			testCtrl.Controller.assignmentChangeChan = nil
-		}
-		klog.Infof("Closed assignment change channel")
 	}
 
 	// Close stop channel if exists


docs/design/sharding_controller.md
```diff
@@ -0,0 +1,356 @@
...
+  min-nodes: 1
+  max-nodes: 100
+- name: volcano-scheduler
+  type: volcano
+  cpu-utilization-min: 0.0
+  cpu-utilization-max: 0.6
+  prefer-warmup-nodes: false
+  min-nodes: 1
+  max-nodes: 100
+```
```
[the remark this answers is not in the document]
ssfffss: This remains to be configured through configmap in the future. For now, it is fixed in the first version;

pkg/controllers/sharding/config.go
```diff
@@ -0,0 +1,126 @@
...
+	controllerOptions := &ShardingControllerOptions{
+		// Default scheduler configs will be added by AddFlags
+		ShardSyncPeriod:        60 * time.Second,
+		EnableNodeEventTrigger: true,
+		SchedulerConfigsRaw: []string{
+			"volcano:volcano:0.0:0.6:false:2:100",
+			"agent-scheduler:agent:0.7:1.0:true:2:100",
+		},
+	}
+	controllerOptions.ParseConfig()
```
qi-min: config parse error is not handled

pkg/controllers/sharding/config.go
```diff
@@ -0,0 +1,126 @@
...
+	EnableNodeEventTrigger bool
+}
+
+func NewShardingControllerOptions() *ShardingControllerOptions {
+	controllerOptions := &ShardingControllerOptions{
+		// Default scheduler configs will be added by AddFlags
+		ShardSyncPeriod:        60 * time.Second,
+		EnableNodeEventTrigger: true,
+		SchedulerConfigsRaw: []string{
+			"volcano:volcano:0.0:0.6:false:2:100",
```
qi-min, line 49: it is better to use percentage instead of node min/max number in case node count changed in cluster

pkg/controllers/sharding/event_handlers.go
```diff
@@ -0,0 +1,54 @@
...
+	// Trigger shard sync if utilization changed significantly
+	if sc.isUtilizationSignificantlyChanged(event.NodeName, metrics) {
+		klog.V(4).Infof("Node %s utilization changed significantly, scheduling sync", event.NodeName)
+		sc.enqueueNodeEvent(event.NodeName, "node-updated", "node-controller")
+	}
+}
+
+// isUtilizationSignificantlyChanged checks if utilization changed significantly
+func (sc *ShardingController) isUtilizationSignificantlyChanged(nodeName string, newMetrics *NodeMetrics) bool {
+	oldMetrics, exists := sc.nodeMetricsCache[nodeName]
```
qi-min, line 31: sc.nodeMetricsCach e[nodeName] is updated with latest metrics before calling isUtilizationSignificantlyChanged at line 33, seems oldMetrics will always be same as newMetrics.

pkg/controllers/sharding/node_utilization.go
```diff
@@ -0,0 +1,182 @@
...
+	memoryCapacity := node.Status.Capacity.Memory()
+	cpuAllocatable := node.Status.Allocatable.Cpu()
+	memoryAllocatable := node.Status.Allocatable.Memory()
+
+	// Calculate utilization
+	cpuUtilization := 0.0
+	memoryUtilization := 0.0
+
+	if cpuCapacity != nil && cpuCapacity.MilliValue() > 0 {
+		cpuUtilization = float64(totalCPURequest) / float64(cpuCapacity.MilliValue())
```
qi-min, line 138: use Allocatable is more accurate than Capacity

pkg/controllers/sharding/node_utilization.go
```diff
@@ -0,0 +1,182 @@
...
+	return activePods, nil
+}
+
+// calculateMetricsFromPodsAndNode calculates comprehensive node metrics from pods and node object
+func (sc *ShardingController) calculateMetricsFromPodsAndNode(pods []*corev1.Pod, node *corev1.Node) (*NodeMetrics, error) {
+	// Calculate total resource requests
+	totalCPURequest := int64(0)
+	totalMemoryRequest := int64(0)
+
+	for _, pod := range pods {
```
qi-min, line 111: It is better to identify usage from non-agent pod. The resource taken by agent pod on Pod should not be count into usage. If node usage is 70%, but agents already take 30%, it means non-agent pods only taks 40%, this node should be moved out from agent scheduler shard because 60% reosurce will be idle once agent complete

pkg/controllers/sharding/sharding_types.go
```diff
@@ -0,0 +1,114 @@
...
+
+// NodeMetricsProvider provides access to node metrics
+type NodeMetricsProvider interface {
+	GetNodeMetrics(nodeName string) *NodeMetrics
+	GetAllNodeMetrics() map[string]*NodeMetrics
+	UpdateNodeMetrics(nodeName string, metrics *NodeMetrics)
+}
+
+// NodeResourceInfo contains resource utilization information for a node
+type NodeResourceInfo struct {
```
qi-min, line 86: what is the difference between NodeMetricsProvider  and NodeMetrics , they looks similar. Is it possibel to combine them?

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,198 @@
...
+	// Check labels/annotations change
+	if !sc.areMapsEqual(oldNode.Labels, newNode.Labels) ||
+		!sc.areMapsEqual(oldNode.Annotations, newNode.Annotations) {
+		return true
+	}
+
+	// Check capacity change
+	oldCPU := oldNode.Status.Capacity.Cpu().Value()
+	newCPU := newNode.Status.Capacity.Cpu().Value()
+	return float64(abs(int(newCPU-oldCPU)))/float64(oldCPU+1) > nodeUsageChangeThreshold
```
qi-min, line 112: nodeUsageChangeThreshold need be configurable

and `float64(abs(int(newCPU-oldCPU)))/float64(oldCPU+1)` seems not related to node usage

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,198 @@
...
+
+// processNodeEvent processes a single node event
+func (sc *ShardingController) processNodeEvent() bool {
+	eventKey, quit := sc.nodeEventQueue.Get()
+	if quit {
+		return false
+	}
+	defer sc.nodeEventQueue.Done(eventKey)
+
+	parts := strings.SplitN(eventKey, ":", 3)
```
qi-min, line 151:
```
	nodeName := parts[0]
	eventType := parts[1]
	source := parts[2]
```
they can be abstracted into a struct and pass to event queue

pkg/controllers/sharding/node_utilization.go
```diff
@@ -0,0 +1,182 @@
...
+// updateNodeUtilization updates node metrics in cache
+func (sc *ShardingController) updateNodeUtilization(nodeName string, metrics *NodeMetrics) {
+	sc.UpdateNodeMetrics(nodeName, metrics)
+
+	// Log significant changes
+	if prevMetrics := sc.GetNodeMetrics(nodeName); prevMetrics != nil {
+		cpuChange := math.Abs(metrics.CPUUtilization - prevMetrics.CPUUtilization)
+		memChange := math.Abs(metrics.MemoryUtilization - prevMetrics.MemoryUtilization)
+
+		if cpuChange > 0.1 || memChange > 0.1 {
```
qi-min: is it ok to use `isUtilizationSignificantlyChanged `here to unify the significant change condition? seems the condition here is different from condition in `isUtilizationSignificantlyChanged`

pkg/controllers/sharding/pod_events.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// addPod handles pod addition events
+func (sc *ShardingController) addPod(obj interface{}) {
+	pod := sc.getPodFromObject(obj)
+	if pod == nil || pod.Spec.NodeName == "" {
+		return
+	}
+
+	sc.enqueueNodeEvent(pod.Spec.NodeName, "pod-added", "pod-controller")
```
qi-min: same eventType and source appear in multiple lines, it is suggested to use const instead

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,198 @@
...
+
+// addNodeEvent handles node addition events
+func (sc *ShardingController) addNodeEvent(obj interface{}) {
+	node := sc.getNodeFromObject(obj)
+	if node == nil {
+		return
+	}
+
+	klog.V(4).Infof("Node added: %s", node.Name)
+	sc.enqueueNodeEvent(node.Name, "node-added", "node-controller")
```
qi-min: it is suggested to use const for type and const  since they may appear in multiple places

pkg/controllers/sharding/pod_events.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		sc.enqueueNodeEvent(newPod.Spec.NodeName, "pod-added", "pod-controller")
+		return
+	}
+
+	// Resource requests changed
+	if sc.podResourcesChanged(oldPod, newPod) {
+		sc.enqueueNodeEvent(newPod.Spec.NodeName, "pod-resources-changed", "pod-controller")
+	}
+}
```
qi-min, line 69: How about pod status change like running to completed? Seems it is not covered in this function

pkg/controllers/sharding/config.go
```diff
@@ -0,0 +1,126 @@
+// pkg/controllers/sharding/config.go
+package sharding
```
qi-min, line 14: all new files lack copyright. file headers

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,770 @@
...
+	}
+
+	// Check node set difference
+	currentSet := make(map[string]bool)
+	for _, node := range shard.Spec.NodesDesired {
+		currentSet[node] = true
+	}
+
+	changed := false
+	minChangeThreshold := max(1, len(assignment.NodesDesired)/10) // 10% threshold
```
qi-min, line 522: If length of desired nodes changes, even if only one node added, assignmentNeedsUpdate will return true. But if length is not changed, then assignmentNeedsUpdate will return true only if more than 10% nodes changed, even if more then one node changes.  The check rule is not consistent.

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,424 @@
...
+}
+
+// filterEligibleNodes applies hard filtering based on strategy
+func (sm *ShardingManager) filterEligibleNodes(
+	config SchedulerConfig,
+	nodeResources map[string]*NodeResourceInfo,
+	nodeMap map[string]*corev1.Node,
+	assignedNodes map[string]string,
+) []*corev1.Node {
+	var eligibleNodes []*corev1.Node
```
qi-min, line 173: is it possible to make filterEligibleNodes to filter with multiple rules and cpu usage is just one of the rules?

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,424 @@
...
+		// add extra nodes until the minimum node constraint is met
+		additionalNeeded := config.ShardStrategy.MinNodes - len(selected)
+		for i := len(selected); i < len(nodes) && additionalNeeded > 0; i++ {
+			selected = append(selected, nodes[i].Name)
+			additionalNeeded--
+		}
+	}
+
+	// remvoe repeated nodes
+	uniqueSelected := make(map[string]bool)
```
qi-min: When will nodes be repeated? Node should be unique in list

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,424 @@
...
+	desiredCount := sm.calculateDesiredNodeCount(config, len(nodes))
+
+	// ensure the desired node count is smaller than the minimum nodes
+	if desiredCount < config.ShardStrategy.MinNodes {
+		desiredCount = config.ShardStrategy.MinNodes
+	}
+
+	// if eligible nodes is smaller than desired, then use all eligible nodes
+	if len(nodes) < desiredCount {
+		desiredCount = len(nodes)
```
qi-min, line 281: if length of nodes less than desiredCount, is it OK to return immediately?

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,424 @@
...
+	// select nodes
+	selected := make([]string, 0, desiredCount)
+
+	// first select nodes according to priority
+	for i := 0; i < desiredCount && i < len(nodes); i++ {
+		selected = append(selected, nodes[i].Name)
+	}
+
+	// check whether it meets the minimum node constraint
+	if len(selected) < config.ShardStrategy.MinNodes && len(nodes) > len(selected) {
```
qi-min, line 293: length of select should be same as desiredCount if length of nodes larger than length of selected, seems this condition will never be true.

pkg/controllers/sharding/metrics_cache.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	for _, node := range nodes {
+		event := &NodeEvent{
+			EventType: "initialization",
+			NodeName:  node.Name,
+			Source:    "startup",
+			Timestamp: time.Now(),
+		}
+		sc.updateNodeMetricsFromEvent(event)
```
qi-min: `sc.updateNodeMetricsFromEvent(event) `is only called in `initializeNodeMetrics` and this event is not put into nodeEventQueue, so fields in Event is not used except NodeName. Is it OK to rename `updateNodeMetricsFromEvent` to updateNodeMetrics and accept nodeName as arguemnt?

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,424 @@
...
+	finalSelection := make([]string, 0, len(selected))
+
+	for _, node := range selected {
+		if !uniqueSelected[node] {
+			uniqueSelected[node] = true
+			finalSelection = append(finalSelection, node)
+		}
+	}
+
+	return finalSelection
```
qi-min: the nodes match constraint and existed in shard should be kept in shard first to keep shard stable

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,770 @@
+package sharding
```
YaoZengzeng, line 14: Please add copy right for all these files.

pkg/controllers/sharding/config.go
```diff
@@ -0,0 +1,126 @@
...
+
+// AddFlags adds flags to the flag set using pflag pattern
+func (opts *ShardingControllerOptions) AddFlags(fs *pflag.FlagSet) {
+	// Default scheduler configs
+	defaultConfigs := []string{
+		"volcano:volcano:0.0:0.6:false:2:100",
+		"agent-scheduler:agent:0.7:1.0:true:2:100",
+	}
+
+	fs.StringSliceVar(&opts.SchedulerConfigsRaw, "scheduler-configs", defaultConfigs,
```
YaoZengzeng, line 67: Use a config file for these complicated configuration may be more appropriate.

pkg/controllers/sharding/event_handlers.go
```diff
@@ -0,0 +1,54 @@
...
+
+// NodeEvent represents a node-related event
+type NodeEvent struct {
+	EventType string // "pod-added", "pod-updated", "pod-deleted", "node-updated"
+	NodeName  string
+	Source    string // "pod-controller", "node-controller"
+	Timestamp time.Time
+}
+
+// updateNodeMetricsFromEvent updates node metrics based on an event
```
YaoZengzeng: This method seems not being called in any pod or node related event handler?

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,198 @@
...
+
+	klog.V(4).Infof("Node deleted: %s", node.Name)
+
+	// Clean up metrics cache
+	sc.metricsMutex.Lock()
+	delete(sc.nodeMetricsCache, node.Name)
+	sc.metricsMutex.Unlock()
+
+	// Trigger shard sync
+	sc.enqueueNodeEvent(node.Name, "node-deleted", "node-controller")
```
YaoZengzeng: Suggest use a constant for pod or node event type and controller type.

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,198 @@
...
+
+// isNodeSignificantlyChanged checks if node changed significantly
+func (sc *ShardingController) isNodeSignificantlyChanged(oldNode, newNode *corev1.Node) bool {
+	// Check labels/annotations change
+	if !sc.areMapsEqual(oldNode.Labels, newNode.Labels) ||
+		!sc.areMapsEqual(oldNode.Annotations, newNode.Annotations) {
+		return true
+	}
+
+	// Check capacity change
```
YaoZengzeng, line 109: Why only check CPU here but check both CPU and memory in `isUtilizationSignificantlyChanged`?

pkg/controllers/sharding/event_handlers.go
```diff
@@ -0,0 +1,54 @@
...
+func (sc *ShardingController) updateNodeMetricsFromEvent(event *NodeEvent) {
+	sc.metricsMutex.Lock()
```
YaoZengzeng: We could use the lock more precisely.

pkg/controllers/sharding/config.go
ssfffss: will leave to the future PR to resolve percentage issues;

pkg/controllers/sharding/config.go
```diff
@@ -0,0 +1,128 @@
...
+			return fmt.Errorf("invalid max utilization in %s: %v", configStr, err)
+		}
+
+		// Parse prefer warmup
+		preferWarmup := false
+		if parts[4] == "true" {
+			preferWarmup = true
+		} else if parts[4] != "false" {
+			return fmt.Errorf("invalid prefer warmup flag in %s: %s, expected 'true' or 'false'", configStr, parts[4])
+		}
```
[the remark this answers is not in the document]
ssfffss: already fixed;

pkg/controllers/sharding/event_handlers.go
```diff
@@ -0,0 +1,61 @@
...
+
+	cpuChange := newMetrics.CPUUtilization - oldMetrics.CPUUtilization
+	if cpuChange < 0 {
+		cpuChange = -cpuChange
+	}
+
+	memChange := newMetrics.MemoryUtilization - oldMetrics.MemoryUtilization
+	if memChange < 0 {
+		memChange = -memChange
+	}
```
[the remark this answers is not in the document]
ssfffss: accepted your suggestion

pkg/controllers/sharding/pod_events.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil
+	default:
+		klog.Warningf("Unexpected object type in pod event handler: %T", obj)
+		return nil
+	}
+}
+
+func (sc *ShardingController) enqueueSyncEvent() {
+	sc.nodeEventQueue.Add("global-sync-trigger")
+}
```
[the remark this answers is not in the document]
ssfffss: this event will not be processed now as there may be a lot of pod events, needs to be discussed in the future

docs/design/sharding_controller.md
```diff
@@ -0,0 +1,356 @@
...
+- Assignment algorithm correctness
+
+### 10.2 Integration Tests
+- End-to-end workflow with fake clients
+- Node and pod events triggering updates
+- Multiple scheduler configurations
+
+### 10.3 Performance Tests
+- 100-node cluster sync time (<5 seconds)
+- 1000-node cluster sync time (<30 seconds)
```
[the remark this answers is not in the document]
ssfffss: already fixed

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// Clean up metrics cache
+	sc.metricsMutex.Lock()
+	delete(sc.nodeMetricsCache, node.Name)
+	sc.metricsMutex.Unlock()
+
+	// Trigger shard sync
+	time.AfterFunc(100*time.Millisecond, func() {
+		sc.syncShards()
+	})
```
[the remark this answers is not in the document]
ssfffss: accepted

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,423 @@
...
+			// Create default metrics if not available
+			metrics = &NodeMetrics{
+				CPUUtilization:    0.3,
+				MemoryUtilization: 0.4,
+				IsWarmupNode:      node.Labels["node.volcano.sh/warmup"] == "true",
+				Labels:            node.Labels,
+				Annotations:       node.Annotations,
+				LastUpdated:       time.Now(),
+			}
+		}
```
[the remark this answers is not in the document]
ssfffss: fixed

pkg/controllers/sharding/config.go
ssfffss: will print log now;

pkg/controllers/sharding/event_handlers.go
ssfffss: fix it;

pkg/controllers/sharding/config.go
ssfffss: will leave to configmap in the next PR;

pkg/controllers/sharding/sharding_controller.go
ssfffss: added;

pkg/controllers/sharding/sharding_manager.go
ssfffss: will fix in the next PR;

pkg/controllers/sharding/node_utilization.go
ssfffss: how to identify whether a pod is agent or non-agent? this will be left to the future discussion!

pkg/controllers/sharding/node_utilization.go
ssfffss: need further discussion

pkg/controllers/sharding/node_events.go
ssfffss: will leave to the config map for changing;

pkg/controllers/sharding/pod_events.go
ssfffss: done

pkg/controllers/sharding/config.go
ssfffss: done

pkg/controllers/sharding/sharding_manager.go
ssfffss: yes, will leave to the future strategy parts;

pkg/controllers/sharding/sharding_manager.go
ssfffss: needs to discuss as there are still constraints;

pkg/controllers/sharding/sharding_manager.go
ssfffss: this is true when selected nodes is smaller than the defined minNode, but there are still nodes that can be selected as candidates;

pkg/controllers/sharding/node_events.go
ssfffss: fixed;

pkg/controllers/sharding/node_events.go
ssfffss: will add more complex strategy that use multi-dimensional metrics;

pkg/controllers/sharding/event_handlers.go
ssfffss: will use for merging events in the future;

pkg/controllers/sharding/event_handlers.go
ssfffss: will solve in the future;

pkg/controllers/sharding/metrics_cache.go
ssfffss: will solve in the next PR;

pkg/controllers/sharding/sharding_manager.go
ssfffss: fixed, not necessary

pkg/controllers/sharding/sharding_controller.go
ssfffss: needs further discussion and future reconstructions;

pkg/controllers/sharding/pod_events.go
ssfffss: for now we do not consider such status events from running to completed;

pkg/controllers/sharding/node_events.go
ssfffss: fixed;

pkg/controllers/sharding/node_utilization.go
ssfffss: will merge in the next PR;

pkg/controllers/sharding/node_events.go
ssfffss: will adjust this with config map PR;

pkg/controllers/sharding/sharding_types.go
ssfffss: will merge with the config map PR;

docs/design/sharding_controller.md
```diff
@@ -0,0 +1,355 @@
...
+2. **Minimal API Server Load**: Use informer caches instead of direct API calls
+3. **Event-Driven Updates**: React to cluster state changes efficiently
+4. **Scalability**: Support clusters with 1000+ nodes with sub-second latency
+5. **Resilience**: Recover gracefully from failures and restarts
+6. **Configurability**: Allow runtime configuration of scheduling policies
+7. **Backward Compatibility**: Work with existing Volcano deployment without disruption
+
+## 3. Architecture Overview
+
+```mermaid
```
hzxuzhonghu: This is AI generated rubbish, this is not readable

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+}
+
+// Return the name of the controller
+func (sc *ShardingController) Name() string {
+	return controllerName
+}
+
+// Initialize initializes the controller
+func (sc *ShardingController) Initialize(opt *framework.ControllerOption) error {
+	klog.V(6).Infof("Initializing ShardingController...")
```
hzxuzhonghu:
```suggestion
	klog.V(2).Infof("Initializing ShardingController...")
```

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+	sc.ctx = context.Background()
+	sc.controllerOptions = *NewShardingControllerOptions()
```
hzxuzhonghu: nit: either change controllerOptions or NewShardingControllerOptions return value type.

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+	sc.shardInformer = sc.vcInformerFactory.Shard().V1alpha1().NodeShards()
+
+	sc.nodeLister = sc.nodeInformer.Lister()
+	sc.podLister = sc.podInformer.Lister()
+	sc.shardLister = sc.shardInformer.Lister()
+
+	sc.initNodeIndices()
+
+	// Initialize queues
+	sc.queue = workqueue.NewTypedRateLimitingQueueWithConfig(
```
hzxuzhonghu: what is the queue for? NodeShard?

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+		workqueue.DefaultTypedControllerRateLimiter[string](),
+		workqueue.TypedRateLimitingQueueConfig[string]{Name: controllerName},
+	)
+	sc.nodeEventQueue = workqueue.NewTypedRateLimitingQueueWithConfig(
```
hzxuzhonghu, line 122: what does `node-events`? If this is for node, prefer naming `nodeQueue` and `nodeShardQueue` above

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+		workqueue.DefaultTypedControllerRateLimiter[string](),
+		workqueue.TypedRateLimitingQueueConfig[string]{Name: controllerName + "-node-events"},
+	)
+
+	// Initialize components
+	// Initialize sharding manager
+	sc.assignmentCache = &AssignmentCache{
+		Assignments: make(map[string]*ShardAssignment),
+		// NodeStates:  make(map[string]*NodeState),
```
hzxuzhonghu: remove

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+	}
+	// sc.nodeStates = make(map[string]*NodeState)
```
hzxuzhonghu: remove

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+		DeleteFunc: sc.deleteNodeEvent,
+	})
+
+	sc.shardInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    sc.addShard,
+		UpdateFunc: sc.updateShard,
+		DeleteFunc: sc.deleteShard,
+	})
+
+	sc.podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
hzxuzhonghu, line 156: Why do we need podInformer, iiuc this is only operating nodes

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+func (sc *ShardingController) updateAllNodeStates() {
+	nodes, err := sc.nodeLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("Failed to list nodes for state update: %v", err)
+		return
+	}
+
+	for _, node := range nodes {
+		// update node states asynchronously to avoid blocking
+		go func(nodeName string) {
```
hzxuzhonghu: This can start go routines out of control. please limit the numbers 

And also we need a waitGroup to make sure all the nodes haven been updated before exit

pkg/controllers/sharding/node_utilization.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return sc.calculateMetricsFromPodsAndNode(pods, node)
+}
+
+// updateNodeUtilization updates node metrics in cache
+func (sc *ShardingController) updateNodeUtilization(nodeName string, metrics *NodeMetrics) {
+	sc.UpdateNodeMetrics(nodeName, metrics)
+
+	// Log significant changes
+	if prevMetrics := sc.GetNodeMetrics(nodeName); prevMetrics != nil {
```
hzxuzhonghu: How could you get previous metrics as you have updated it?

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+		sc.nodeInformer.Informer().HasSynced,
+		sc.shardInformer.Informer().HasSynced,
+		sc.podInformer.Informer().HasSynced,
+	) {
+		klog.Errorf("cache sync failed")
+	}
+	klog.Infof("Cache synchronization completed successfully")
+
+	// Initialize node metrics
+	sc.initializeNodeMetrics()
```
hzxuzhonghu: is this same with `refreshNodeMetrics`, please keep one

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,218 @@
...
+			return true
+		}
+		klog.Warningf("Dropping node event %s after 3 retries", eventKey)
+	}
+
+	return true
+}
+
+// processNodeEventWithRetry processes a node event with retry logic
+func (sc *ShardingController) processNodeEventWithRetry(nodeName, eventType, source string, maxRetries int) error {
```
hzxuzhonghu, line 173: eventType  and source are not even used, why do you make `eventKey` so complex

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,218 @@
...
+	var lastErr error
+
+	for i := 0; i < maxRetries; i++ {
```
hzxuzhonghu, line 176: With this loop, you make a node retry at most 3*3 times

pkg/controllers/sharding/node_events.go
```diff
@@ -0,0 +1,218 @@
...
+
+		// Update metrics cache
+		sc.UpdateNodeMetrics(nodeName, metrics)
+
+		// Check if significant change
+		if sc.isUtilizationSignificantlyChanged(nodeName, metrics) {
+			klog.V(4).Infof("Node %s utilization changed significantly, scheduling shard sync", nodeName)
+			// Schedule sync in background with delay
+			time.AfterFunc(200*time.Millisecond, func() {
+				sc.syncShards()
```
hzxuzhonghu, line 209: any consideration about the delay?

hzxuzhonghu: Why do we have two places triggering syncShards: 1. node update, 2. periordically 

Vaguely remember we donot want to update node shards very frequently. So I think we should have only one routine running sync, if we want node change to trigger sync, we should aggregate node events rather than each node trigger once

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+	for _, nodeMetrics := range sc.nodeMetricsCache {
+		if nodeMetrics.LastUpdated.After(lastUpdateTime) {
+			lastUpdateTime = nodeMetrics.LastUpdated
+		}
+	}
+	sc.metricsMutex.RUnlock()
+
+	if time.Since(lastUpdateTime) > updateTimeoutThreshold {
+		klog.V(4).Infof("Node states stale, triggering update")
+		sc.updateAllNodeStates()
```
hzxuzhonghu: I donot think we need update all again.  iirc, the node metrcs are updated periordically and also by node event.

pkg/controllers/sharding/config.go
```diff
@@ -0,0 +1,141 @@
...
+	EnableNodeEventTrigger bool
+}
+
+func NewShardingControllerOptions() *ShardingControllerOptions {
+	controllerOptions := &ShardingControllerOptions{
+		// Default scheduler configs will be added by AddFlags
+		ShardSyncPeriod:        60 * time.Second,
+		EnableNodeEventTrigger: true,
+		SchedulerConfigsRaw: []string{
+			"volcano:volcano:0.0:0.6:false:2:100",
```
hzxuzhonghu, line 49: can we use structured json instead of a simple string

pkg/controllers/sharding/sharding_performance_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+			pod := CreateTestPod("default", podName, nodeName, podCPU, "1Gi")
+			objects = append(objects, pod)
+		}
+	}
+
+	// Setup controller with appropriate configs for performance testing
+	opt := &TestControllerOption{
+		InitialObjects: objects,
+		// FIX: Increase max nodes for performance testing
```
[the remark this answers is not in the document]
ssfffss: removed;

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,426 @@
...
+			} else {
+				assignments[scheduler] = assignment
+			}
+		}
+
+		// Small delay between batches to prevent resource starvation
+		time.Sleep(10 * time.Millisecond)
+	}
+
+	return assignments, nil
```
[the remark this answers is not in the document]
ssfffss: this is not an issue; as when the number of nodes is larger than the configured batch size, then all nodes will be processed in batches in the calculateShardAssignmentsBatched function; When it calls the CalculateShardAssignments function, it will process directly as the number of nodes is smaller than the batch size;

pkg/controllers/sharding/config.go
qi-min: It is recommended to create an issue tracking the enhancement, avoid  lose the review comments

pkg/controllers/sharding/pod_events.go
qi-min: that might be a problem because terminated pod does not take resource

pkg/controllers/sharding/node_utilization.go
qi-min: agent pod should be label identify, the label should be configurable in configuration

pkg/controllers/sharding/node_utilization.go
ssfffss: fixed;

pkg/controllers/sharding/node_events.go
ssfffss: yes, you are right; Here we need to aggregate node events to trigger shard update but not for all nodes; this will be fixed in the next PR;

pkg/controllers/sharding/sharding_controller.go
ssfffss: ensureNodeStatesUpdated is removed;

pkg/controllers/sharding/config.go
ssfffss: will fixed in the future with the configmap PR;

pkg/controllers/sharding/node_events.go
ssfffss: remove the requeue logic;

pkg/controllers/sharding/node_events.go
ssfffss: this will be used for future threshold optimization including resource usage changes and updating frequency;

pkg/controllers/sharding/sharding_controller.go
ssfffss: changed;

pkg/controllers/sharding/sharding_controller.go
ssfffss: changed to sequentially to avoid out of control. Will changed to batch asynchronization in the future;

ssfffss: removed this block, unnecessary;

pkg/controllers/sharding/sharding_controller.go
ssfffss: This is considered in the beginning node status is changed with pod events; this needs further discussion whether we should drop pod events or not; Will fixed in the future;

pkg/controllers/sharding/sharding_controller.go
ssfffss: nodeshard;

pkg/controllers/sharding/sharding_controller.go
ssfffss: changed;

pkg/controllers/sharding/sharding_controller.go
ssfffss: changed;

pkg/controllers/sharding/sharding_function_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	// Setup controller
+	opt := &TestControllerOption{
+		InitialObjects:  []runtime.Object{testNode},
+		ShardSyncPeriod: 2 * time.Second,
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,783 @@
...
+	defer sc.nodeEventQueue.ShutDown()
+
+	klog.Infof("Starting sharding controller.")
+	defer klog.Infof("Shutting down sharding controller.")
+
+	// FIX 1: Start informer factories HERE
+	sc.kubeInformerFactory.Start(stopCh)
+	sc.vcInformerFactory.Start(stopCh)
+
+	// FIX 2: Add specific sync checks with detailed logging
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_performance_test.go
```diff
@@ -0,0 +1,261 @@
...
+				PreferWarmupNodes: false,
+				MinNodes:          1,
+				MaxNodes:          10,
+			},
+		},
+		ShardSyncPeriod: 5 * time.Second,
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_function_test.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+	}
+
+	// Setup controller with custom configs
+	opt := &TestControllerOption{
+		SchedulerConfigs: customConfigs,
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_update_test.go
```diff
@@ -0,0 +1,202 @@
...
+	// Setup controller
+	opt := &TestControllerOption{
+		InitialObjects: []runtime.Object{
+			CreateTestNode("dynamic-node", "8", false, nil),
+		},
+		ShardSyncPeriod: 2 * time.Second,
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_function_test.go
```diff
@@ -0,0 +1,164 @@
...
+func TestControllerInitialization(t *testing.T) {
+	// Setup with minimal objects
+	opt := &TestControllerOption{
+		InitialObjects: []runtime.Object{
+			CreateTestNode("node-1", "8", false, nil),
+		},
+	}
+	klog.Info("Controller initialization test passed")
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_boundary_test.go
```diff
@@ -0,0 +1,229 @@
...
+				PreferWarmupNodes: false,
+				MinNodes:          1,
+				MaxNodes:          10,
+			},
+		},
+		ShardSyncPeriod: 2 * time.Second,
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_boundary_test.go
```diff
@@ -0,0 +1,229 @@
...
+// TestMaxUtilization tests max utilization handling
+func TestMaxUtilization(t *testing.T) {
+	// Setup controller
+	opt := &TestControllerOption{
+		ShardSyncPeriod:  2 * time.Second,
+		SchedulerConfigs: getDefaultTestSchedulerConfigs(),
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_performance_test.go
```diff
@@ -0,0 +1,261 @@
...
+				MaxNodes:          50, // Increased from 10
+			},
+		},
+		ShardSyncPeriod: 10 * time.Second,
+	}
+
+	startTime := time.Now()
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/pod_events.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+const (
+	POD_SOURCE         = "pod-controller"
+	POD_ADD_EVENT      = "pod-added"
+	POD_SCHEDULE_EVENT = "pod-scheduled"
+	POD_REMOVE_EVENT   = "pod-removed"
+	POD_DELETE_EVENT   = "pod-deleted"
+	POD_CHANGE_EVENT   = "pod-changed"
+)
```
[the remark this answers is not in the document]
ssfffss: fixed

pkg/controllers/sharding/sharding_update_test.go
```diff
@@ -0,0 +1,202 @@
...
+	opt := &TestControllerOption{
+		InitialObjects: []runtime.Object{
+			CreateTestNode("initial-node", "8", true, nil),
+		},
+		SchedulerConfigs: getDefaultTestSchedulerConfigs(),
+		ShardSyncPeriod:  2 * time.Second,
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_performance_test.go
```diff
@@ -0,0 +1,261 @@
...
+			cpuRequest := 200 + j*300 + (i%5)*50 // Varied CPU requests
+			pod := CreateTestPod("default", fmt.Sprintf("init-pod-%d-%d", i, j), nodeName,
+				fmt.Sprintf("%dm", cpuRequest), "1Gi")
+			objects = append(objects, pod)
+		}
+	}
+
+	opt := &TestControllerOption{
+		InitialObjects: objects,
+		// FIX: Configure for better node assignment
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,426 @@
...
+		return infoI.CPUUtilization > infoJ.CPUUtilization
+	})
+
+	sort.Slice(nonWarmupNodes, func(i, j int) bool {
+		infoI := nodeResources[nonWarmupNodes[i].Name]
+		infoJ := nodeResources[nonWarmupNodes[j].Name]
+		return infoI.CPUUtilization > infoJ.CPUUtilization
+	})
+
+	// combine results：prefer warmup node
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_boundary_test.go
```diff
@@ -0,0 +1,229 @@
...
+func TestEmptyCluster(t *testing.T) {
+	// Setup controller with no initial objects
+	opt := &TestControllerOption{
+		InitialObjects:   []runtime.Object{},
+		SchedulerConfigs: getDefaultTestSchedulerConfigs(),
+		ShardSyncPeriod:  2 * time.Second,
+	}
+
+	testCtrl := NewTestShardingController(t, opt)
+	defer close(testCtrl.StopCh)
```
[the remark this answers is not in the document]
ssfffss: fixed;

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,426 @@
...
+	nodes []*corev1.Node,
+	currentShards []*shardv1alpha1.NodeShard,
+) (map[string]*ShardAssignment, error) {
+	startTime := time.Now()
+
+	defer func() {
+		duration := time.Since(startTime)
+		klog.V(4).Infof("Calculated shard assignments in %v for %d nodes", duration, len(nodes))
+
+		// FIX: Add performance metrics
```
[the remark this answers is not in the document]
ssfffss: fixed;

docs/design/sharding_controller.md
```diff
@@ -0,0 +1,309 @@
...
+```yaml
+scheduler-configs:
+- name: agent-scheduler
+  type: agent
+  cpu-utilization-min: 0.7
+  cpu-utilization-max: 1.0
+  prefer-warmup-nodes: true
+  min-nodes: 1
+  max-nodes: 100
+- name: volcano-scheduler
```
JesseStutler, line 205: The meaning of each field needs at least an explanation.

pkg/controllers/sharding/node_utilization.go
```diff
@@ -0,0 +1,185 @@
...
+
+		CPUCapacity:       *cpuCapacity,
+		CPUAllocatable:    *cpuAllocatable,
+		MemoryCapacity:    *memoryCapacity,
+		MemoryAllocatable: *memoryAllocatable,
+
+		CPUUtilization:    cpuUtilization,
+		MemoryUtilization: memoryUtilization,
+
+		IsWarmupNode: node.Labels["node.volcano.sh/warmup"] == "true",
```
JesseStutler, line 165: Should also write and explain this label in the doc

pkg/controllers/sharding/event_handlers.go
```diff
@@ -0,0 +1,38 @@
...
+
+// isUtilizationSignificantlyChanged checks if utilization changed significantly
+func (sc *ShardingController) isUtilizationSignificantlyChanged(nodeName string, newMetrics *NodeMetrics) bool {
+	oldMetrics, exists := sc.nodeMetricsCache[nodeName]
+	if !exists {
+		return true // New node always considered significant
+	}
+	cpuChange := math.Abs(newMetrics.CPUUtilization - oldMetrics.CPUUtilization)
+	memChange := math.Abs(newMetrics.MemoryUtilization - oldMetrics.MemoryUtilization)
+	return cpuChange > nodeUsageChangeThreshold || memChange > nodeUsageChangeThreshold
```
JesseStutler, line 37: The doc should also clarify that the node will only be considered if the change in utilization exceeds the threshold.

docs/design/sharding_controller.md
```diff
@@ -0,0 +1,309 @@
...
+**Calculation Logic**:
+1. List all active pods on the node
+2. Sum CPU and memory requests from pod containers
+3. Divide by node capacity to get utilization percentages
+4. Factor in warmup node preference for Agentic AI workloads
+
+**Node Event Processing Flow**
+```mermaid
+graph LR
+    A[Pod Added/Updated/Deleted] --> B{Is Node Significant?}
```
JesseStutler: `Is Node Significant?` ---> `Is Node Significantly Changed?`

docs/design/sharding_controller.md
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+```
+
+## 5. Event Flow
+
+### 5.1 Node/Pod Events
+1. Pod added/updated/deleted on a node
+2. Controller detects event and updates node metrics
+3. If utilization changes significantly (>10%), trigger shard sync
```
JesseStutler: Why is the doc inconsistent with the code? The code shows 0.5.

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,742 @@
...
+
+	// Start periodic metrics refresh
+	go wait.Until(sc.refreshNodeMetrics, nodeRefreshPeriod, stopCh)
+
+	<-stopCh
+	klog.Infof("Shutting down %s", controllerName)
+}
+
+// refreshNodeMetrics periodically refreshes node metrics
+func (sc *ShardingController) refreshNodeMetrics() {
```
JesseStutler, line 211: Is it necessary to refresh all the node metrics periodically? I saw that sharding controller will refresh node metrics according to the node events

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,425 @@
...
+
+		// Small delay between batches to prevent resource starvation
+		time.Sleep(10 * time.Millisecond)
+	}
+
+	return assignments, nil
+}
+
+// filterEligibleNodes applies hard filtering based on strategy
+func (sm *ShardingManager) filterEligibleNodes(
```
JesseStutler, line 167: Why is the current filter logic is hard-coded and didn't use plugins logic? Which is different with our initial design? https://docs.google.com/document/d/1NDTEGulhQ__sdZ_c2kN3ojtJIZie5O7_GyZyKIOLz-g/edit?tab=t.tk4xz1b4bu2r

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,425 @@
...
+			continue
+		}
+
+		// Skip nodes with invalid utilization data
+		if resourceInfo.CPUUtilization < 0 {
+			continue
+		}
+
+		// round the utilization with 2 floating points
+		cpuUtilRounded := math.Round(resourceInfo.CPUUtilization*100) / 100
```
JesseStutler, line 190: We only think about CPU currently?

docs/design/sharding_controller.md
```diff
@@ -0,0 +1,309 @@
...
+
+### 6.3 Update Strategy
+- **Event-Driven**: Update metrics when pod events occur
+- **Periodic Refresh**: Full metrics refresh every 30 seconds
+- **Significant Change Detection**: Only trigger shard sync when utilization changes >10%
+
+## 7. Shard Assignment Algorithm
+
+### 7.1 Hard Filtering Approach
+Unlike complex scoring systems, we use hard filtering:
```
JesseStutler: This is too simplistic, could you change it to a flowchart? Thanks. And this shouldn't be called hard filtering, it should be called utilization-based filtering, right? Also, why does the code only consider the CPU? Furthermore, it doesn't explain here current sharding logic is based on the interval [min, max].

docs/design/sharding_controller.md
ssfffss: fixed

docs/design/sharding_controller.md
ssfffss: fixed

pkg/controllers/sharding/event_handlers.go
ssfffss: fixed

docs/design/sharding_controller.md
ssfffss: fixed  the doc

pkg/controllers/sharding/sharding_controller.go
ssfffss: The node events are mainly about node add and delete. The periodic update mainly considers the utilization change; So in the future, we should decide whether using pod events to detect utilization changes and removing the periodic updates, or removing pod events and using periodic updates for utilization changes;  The pod events may be very frequent, this cause a lot of pod events; So we prefer the periodic updates; This will be tracked in issue #4878

pkg/controllers/sharding/sharding_manager.go
ssfffss: For the first version, we mainly verify the logic; It is tracked by issue #4879

pkg/controllers/sharding/sharding_manager.go
ssfffss: For the first version, we mainly verify the logic; It is tracked by issue https://github.com/volcano-sh/volcano/issues/4879

docs/design/sharding_controller.md
ssfffss: fix the doc with more detailed analysis and flowcharts;

pkg/controllers/sharding/node_utilization.go
ssfffss: fixed

pkg/controllers/sharding/sharding_manager.go
JesseStutler: I agree with the `Utilization-based Filtering`, but should clarify in the doc that currently we only support CPU utilization and will support multi-dimensional resources in the future

JesseStutler:
/lgtm
Wait for @hzxuzhonghu approving

cmd/controller-manager/app/options/options.go
```diff
@@ -157,11 +157,11 @@ func (s *ServerOption) CheckOptionOrDie() error {
...
 func (s *ServerOption) checkControllers() error {
 	existenceMap := make(map[string]bool)
-	for _, c := range s.Controllers {
+	for idx, c := range s.Controllers {
 		if c == "*" {
 			// wildcard '*' is not allowed to be combined with other input
-			if len(s.Controllers) > 1 {
-				return fmt.Errorf("wildcard '*' cannot be combined with other input")
+			if len(s.Controllers) > 1 && idx != len(s.Controllers)-1 {
+				return fmt.Errorf("wildcard '*' can only be placed at the final position when combined with other input")
```
hzxuzhonghu: I am not sure why cannot put * at last? Previously i can be only put at first, now you changed to be any position except last

cmd/controller-manager/app/options/options.go
```diff
@@ -131,7 +131,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForQueue, "worker-threads-for-queue", defaultQueueWorkers, "The number of threads syncing queue operations. The larger the number, the faster the queue processing, but requires more CPU load.")
-	fs.StringSliceVar(&s.Controllers, "controllers", []string{defaultControllers}, fmt.Sprintf("Specify controller gates. Use '*' for all controllers, all knownController: %s ,and we can use "+
+	fs.StringSliceVar(&s.Controllers, "controllers", strings.Split(defaultControllers, ","), fmt.Sprintf("Specify controller gates. Use '*' for all controllers, all knownController: %s ,and we can use "+
```
hzxuzhonghu, line 134: suggest making `*,-sharding-controller` as the default

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,742 @@
...
+func (sc *ShardingController) processAssignmentChange(event *AssignmentChangeEvent) {
+	klog.V(4).Infof("Processing assignment change for %s: %d -> %d nodes",
+		event.SchedulerName, len(event.OldNodes), len(event.NewNodes))
+
+	// Log significant changes
+	if len(event.OldNodes) > 0 {
+		changePercent := float64(abs(len(event.NewNodes)-len(event.OldNodes))) / float64(len(event.OldNodes))
+		if changePercent > nodeCountChangeThreshold {
+			klog.Infof("Significant node change for %s: %.0f%% (%d -> %d nodes)",
+				event.SchedulerName, changePercent*100, len(event.OldNodes), len(event.NewNodes))
```
hzxuzhonghu: Is this function just log, seems very heavy to start a go routine for logging assignment change? Why not directly log it at the place where sender resides

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	go wait.Until(sc.nodeEventWorker, time.Second, stopCh)
+
+	// Start periodic sync
+	go wait.Until(sc.syncShards, sc.shardSyncPeriod, stopCh)
+
+	// Start assignment change processor
+	go sc.assignmentChangeProcessor(stopCh)
+
+	// Start periodic metrics refresh
+	go wait.Until(sc.refreshNodeMetrics, nodeRefreshPeriod, stopCh)
```
hzxuzhonghu, line 204: IIUC, i think nodeRefreshPeriod should better <=  shardSyncPeriod, but now shardSyncPeriod is far smaller, which means one `syncShards` after another maybe unnecessary because of no utils metrics changed during this time window

cmd/controller-manager/app/options/options.go
ssfffss: this is because when we check whether a controller is enabled or not through isControllerEnabled, its logic is when it finds "*" mark, it returns true; it will ignore all the configurations after '*'. Therefore, '*" can only be placed in the last position; This also aligns with our logic, we can put exceptions at first, and enable others; Easy to use in this way and logically clean;

pkg/controllers/sharding/metrics_cache.go
```diff
@@ -0,0 +1,39 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sharding
+
+// Implement NodeMetricsProvider interface
+func (sc *ShardingController) GetNodeMetrics(nodeName string) *NodeMetrics {
+	return sc.nodeMetricsCache[nodeName]
```
hzxuzhonghu, line 18: nit： why donot protect with lock like GetAllNodeMetrics

Same question on UpdateNodeMetrics

pkg/controllers/sharding/sharding_manager.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	return assignments, nil
+}
+
+// filterEligibleNodes applies hard filtering based on strategy
+func (sm *ShardingManager) filterEligibleNodes(
+	config SchedulerConfig,
+	nodeResources map[string]*NodeResourceInfo,
+	nodeMap map[string]*corev1.Node,
```
hzxuzhonghu, line 170: nit: nodeResources should be kept only

pkg/controllers/sharding/metrics_cache.go
ssfffss: now the lock is the for whole cache, lock operations would be too frequent if for each node. Here needs careful optimizations on fine-granularity locks as referred in issue #4878

pkg/controllers/sharding/sharding_controller.go
ssfffss: fixed; but in the issues, we will remove it next time as we already use event to trigger node refresh, no need for periodic update;

pkg/controllers/sharding/sharding_controller.go
ssfffss: removed;

cmd/controller-manager/app/options/options.go
ssfffss: fixed;

cmd/controller-manager/app/options/options.go
ssfffss: fixed;

pkg/controllers/sharding/sharding_manager.go
ssfffss: fixed;

pkg/controllers/sharding/sharding_manager.go
ssfffss: kept now;

hzxuzhonghu (commented):
/approve

it has been disabled by default

PR #4782, opened 2025-12-04
Optimize CI with reusable image workflow

### LINKED ISSUE #4766: Improve ci with reusable workflows

Currently we are running our CI jobs completely paralelly, each e2e job will just invoke the appropriate make target so they build the images as a first step.

See for example the admission workflow: https://github.com/volcano-sh/volcano/blob/master/.github/workflows/e2e_admission.yaml#L53 that invokes
https://github.com/volcano-sh/volcano/blob/master/Makefile#L162

This is not efficient to build the images at each step:

- From resource usage point of view it's unnecesarry.
- When someone retriggers the pipeline the whole process rebuilds the images again, which results in a much more time consuming feedback-loop.

I would like to see the image building in a separate reusable workflow:
[https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows)

It should update the created docker image as an artifacts with maybe a month of expiration:
[https://docs.github.com/en/actions/tutorials/store-and-share-data](https://docs.github.com/en/actions/tutorials/store-and-share-data
)
For this standard docker save - docker load commands can be used to load the artifacts created later on in the next workflow.

As a bonus image validation can be added to the e2e workflows which fails and indicates the need to retrigger the original image building workflow if needed.

Deepam02:

This PR optimizes the CI pipeline by extracting Docker image building into a centralized reusable workflow. Previously, each E2E test job built images independently, resulting in redundant builds and wasted resources.

**Changes:**
- Created `images.yaml` as a reusable workflow that builds all Volcano Docker images once
- Saves built images as artifacts 
- Updated all 10 E2E workflows to download and load pre-built images instead of rebuilding
- Uses standard `docker save`/`docker load` commands for artifact handling

**Benefits:**
- Eliminates redundant image builds across parallel jobs
- Significantly reduces CI execution time and resource usage
- Faster feedback loop when retriggering failed pipelines
- Images are built once and reused across all test jobs

Fixes #4766 

The implementation follows GitHub's reusable workflow pattern. All E2E workflows now have a `needs: build-images` dependency and download artifacts before running tests. The existing test infrastructure remains unchanged - only the image provisioning mechanism has been optimized.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,11 +103,11 @@ images:
 
 save-images:
 	@mkdir -p ${IMAGES_DIR}
-	@echo "Saving images with OCI format and gzip compression..."
-	docker save --format oci ${IMAGE_PREFIX}/vc-controller-manager:$(TAG) | gzip > ${IMAGES_DIR}/vc-controller-manager-$(TAG).tar.gz
-	docker save --format oci ${IMAGE_PREFIX}/vc-scheduler:$(TAG) | gzip > ${IMAGES_DIR}/vc-scheduler-$(TAG).tar.gz
-	docker save --format oci ${IMAGE_PREFIX}/vc-webhook-manager:$(TAG) | gzip > ${IMAGES_DIR}/vc-webhook-manager-$(TAG).tar.gz
-	docker save --format oci ${IMAGE_PREFIX}/vc-agent:$(TAG) | gzip > ${IMAGES_DIR}/vc-agent-$(TAG).tar.gz
+	@echo "Saving images with gzip compression..."
+	docker save ${IMAGE_PREFIX}/vc-controller-manager:$(TAG) | gzip > ${IMAGES_DIR}/vc-controller-manager-$(TAG).tar.gz
+	docker save ${IMAGE_PREFIX}/vc-scheduler:$(TAG) | gzip > ${IMAGES_DIR}/vc-scheduler-$(TAG).tar.gz
+	docker save ${IMAGE_PREFIX}/vc-webhook-manager:$(TAG) | gzip > ${IMAGES_DIR}/vc-webhook-manager-$(TAG).tar.gz
+	docker save ${IMAGE_PREFIX}/vc-agent:$(TAG) | gzip > ${IMAGES_DIR}/vc-agent-$(TAG).tar.gz
 	@echo "Images saved to ${IMAGES_DIR}"
 
 load-images:


.github/workflows/images.yaml
```diff
@@ -0,0 +1,41 @@
...
+          swap-storage: true
+
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Build Images
+        run: |
+          make images
+
+      - name: Save Images
```
hajnalmt, line 28: Let's create a separate make target for this one like `make save-images`, as it can be useful to others to save these images.

.github/workflows/images.yaml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          mkdir -p images
+          docker save volcanosh/vc-controller-manager:$(git rev-parse --verify HEAD) -o images/vc-controller-manager.tar
```
hajnalmt: Please also comprass this with gzip, and use the OCI format so it can be loaded by other OCI clients too and the oci format is smaller in size mostly too. Additionally if you move this to a make target, you can define the directory beneath the `_output` directory like `IMAGES_DIR=_output/images` and use the `TAG` variable too instead of the `git rev-parse`.

This command should look like this, the tag included in the output:
`docker save --format oci ... | gzip > vc-controller-mager-$(TAG).tar.gz`

.github/workflows/images.yaml
```diff
@@ -0,0 +1,41 @@
...
+          docker save volcanosh/vc-scheduler:$(git rev-parse --verify HEAD) -o images/vc-scheduler.tar
+          docker save volcanosh/vc-webhook-manager:$(git rev-parse --verify HEAD) -o images/vc-webhook-manager.tar
+          docker save volcanosh/vc-agent:$(git rev-parse --verify HEAD) -o images/vc-agent.tar
+
+      - name: Upload Images
+        uses: actions/upload-artifact@v4
+        with:
+          name: volcano-images
+          path: images
+          retention-days: 5
```
hajnalmt, line 37: I was wondering if this is maybe not enough, but let's stay with 5, you are probably right.

.github/workflows/e2e_admission.yaml
```diff
@@ -47,21 +51,35 @@ jobs:
...
+        run: |
+          for image in images/*.tar; do
+            docker load -i $image
+          done
+
       - name: Run E2E Tests
         run: |
           export ARTIFACTS_PATH=${{ github.workspace }}/e2e-admission-logs
-          make e2e-test-admission-policy CC=/usr/local/musl/bin/musl-gcc
+          E2E_TYPE=ADMISSION_POLICY ./hack/run-e2e-kind.sh
```
hajnalmt: You modified this to avoid the `images` target to be reinvoked right? It won't rebuild it, just checks if the images are there. We can live with that, but remove the CC=musl-gcc... part. That's needed for the image building only.

.github/workflows/e2e_admission.yaml
```diff
@@ -47,21 +51,35 @@ jobs:
           GO111MODULE="on" go install sigs.k8s.io/kind@v0.30.0
           curl -LO https://dl.k8s.io/release/v1.34.0/bin/linux/amd64/kubectl && sudo install kubectl /usr/local/bin/kubectl
 
+      - name: Download Images
```
hajnalmt, line 46: There is an `- name: Install musl` step above this please delete those from the workflows.

.github/workflows/e2e_spark.yaml
```diff
@@ -80,16 +84,28 @@ jobs:
...
+    - name: Load and Deploy volcano images
       run: |
         eval $(minikube docker-env)
-        make TAG=${{ env.TAG }} update-development-yaml
+        for image in images/*.tar; do
+          docker load -i $image
+        done
+        
+        # Use git SHA as TAG to match pre-built images
+        export TAG=$(git rev-parse --verify HEAD)
```
hajnalmt: TAG is generated in the makefile from Makefile.dev exactly like this. This is not needed.

.github/workflows/e2e_dra.yml
```diff
@@ -47,10 +51,22 @@ jobs:
...
 
+      - name: Download Images
+        uses: actions/download-artifact@v4
+        with:
+          name: volcano-images
+          path: images
+
+      - name: Load Images
+        run: |
+          for image in images/*.tar; do
```
hajnalmt: Image loading can be a make target too.

Deepam02:
Hi @hajnalmt 
I have made all the suggested changes, squashed the commits, and rebased.
Sorry for the delay in getting this done, please review it again when you get time.

.github/workflows/e2e_admission.yaml
hajnalmt: I realized I wasn't clear here.
We shall use the make target whenever it's possible just remove the CC part as it was not needed in the first place.
so these should be:
```
make e2e-test-admission-policy
```

Makefile
```diff
@@ -99,6 +99,25 @@ images:
 		docker buildx build -t "${IMAGE_PREFIX}/vc-$$name:$(TAG)" . -f ./installer/dockerfile/$$name/Dockerfile --output=type=${BUILDX_OUTPUT_TYPE} --platform ${DOCKER_PLATFORMS} --build-arg APK_MIRROR=${APK_MIRROR} --build-arg OPEN_EULER_IMAGE_TAG=${OPEN_EULER_IMAGE_TAG}; \
 	done
 
+IMAGES_DIR=_output/images
+
+save-images:
+	@mkdir -p ${IMAGES_DIR}
+	@echo "Saving images with OCI format and gzip compression..."
+	docker save --format oci ${IMAGE_PREFIX}/vc-controller-manager:$(TAG) | gzip > ${IMAGES_DIR}/vc-controller-manager-$(TAG).tar.gz
```
hajnalmt: I was wrong here, docker is already oci compliant with the save, command, so you can remove the `--format oci` too, as it's not needed.

Makefile
```diff
@@ -99,6 +99,25 @@ images:
 		docker buildx build -t "${IMAGE_PREFIX}/vc-$$name:$(TAG)" . -f ./installer/dockerfile/$$name/Dockerfile --output=type=${BUILDX_OUTPUT_TYPE} --platform ${DOCKER_PLATFORMS} --build-arg APK_MIRROR=${APK_MIRROR} --build-arg OPEN_EULER_IMAGE_TAG=${OPEN_EULER_IMAGE_TAG}; \
 	done
 
+IMAGES_DIR=_output/images
```
hajnalmt: Please put this at the beggining of the file.

```
BIN_DIR=_output/bin
IMAGES_DIR=_output/images
RELEASE_DIR=_output/release
```
If yout want to do it the right way you can declare a variable for the OUTPUT_DIR too.
```
OUTPUT_DIR=${PWD}/_output
BIN_DIR=${OUTPUT_DIR}/bin
RELEASE_DIR=${OUTPUT_DIR}/release
IMAGES_DIR=${OUTPUT_DIR}/release
```

Deepam02:
@hajnalmt 

I've restructured the workflows as suggested:

- Created a new `e2e.yaml` orchestrator that calls `images.yaml` first, then triggers all E2E workflows in parallel


CI should be green now. Please review when you have a chance.

hajnalmt:
/unhold

Thank you! Let's see :)

hajnalmt:
/cc @JesseStutler

Can you trigger testing somehow on this one? We probably need to remove the ok-to-test label and readd it since the workflow structure changed.

JesseStutler:
> /cc @JesseStutler
> 
> Can you trigger testing somehow on this one? We probably need to remove the ok-to-test label and readd it since the workflow structure changed.

`/ok-to-test` label is enough

.github/workflows/e2e_cronjob.yaml
```diff
@@ -33,12 +29,6 @@ jobs:
         with:
           go-version: 1.24.x
 
-      - name: Install musl
```
JesseStutler, line 36: So currently we don't need musl now? @Deepam02 @hajnalmt

.github/workflows/e2e_hypernode.yaml
```diff
@@ -1,11 +1,7 @@
 name: E2E Hypernode
 
 on:
-  push:
-    branches:
-      - master
-    tags:
-  pull_request:
+  workflow_call:
```
JesseStutler, line 4: Sorry I'm not so familiar with workflow_call, so all the CIs can stil be triggered throught the e2e.yaml, right?

hajnalmt: Yes hopefully, if we do a retrigger with retest it will only retrigger the failed tests too.
But some testing probably needs to be done on this to be sure.

JesseStutler:
Seems that all the CIs have triggered but some CIs failed such as `e2e-spark` and `e2e-vcctl`, some failed CIs event can't load images @Deepam02 But why we still have pending checks here?
image
Shouldn't we already have some CIs passed and don't need to execute the independent old CIs again?

.github/workflows/e2e_hypernode.yaml
JesseStutler: image
But I found that we still have these old pending checks waiting to be reported, but we already have execute these CIs, we should not execute them twice, right?

JesseStutler:
What's more, you still have built all the images in each workflow:
image
But what we want is that we just need to build the images only once, and load to the kind clusters in each workflow. @Deepam02

.github/workflows/e2e_cronjob.yaml
hajnalmt: We are currently not using musl at all.

The Dockefiles are not processing the CC variable as arguments so it will use gcc.
[scheduler Dockerfile](https://github.com/volcano-sh/volcano/blob/master/installer/dockerfile/scheduler/Dockerfile)

It's not even passed to the images target:
[images target](https://github.com/volcano-sh/volcano/blob/master/Makefile#L99)
It was just blindly copied each time a new CI job was introduced I guess.

Probably when Monokaix introduced the custom plugins:
https://github.com/volcano-sh/volcano/tree/master/example/custom-plugin
It was added everywhere for some reason, but It's hard to understand the reason behind this now.

.github/workflows/e2e_hypernode.yaml
hajnalmt: This is there because the master branch triggers them. It should disappear I think after we manage to merge this PR.

.github/workflows/e2e_dra.yml
```diff
@@ -29,12 +25,6 @@ jobs:
         with:
           go-version: 1.24.x
 
-      - name: Install musl
-        run: |
-          wget http://musl.libc.org/releases/musl-1.2.1.tar.gz
-          tar -xf musl-1.2.1.tar.gz && cd musl-1.2.1
-          ./configure
-          make && sudo make install
       - name: Checkout code
```
hajnalmt, line 28: Checkout code and Install dependencies can be moved to `e2e.yaml` too! We don't need them in every e2e test now.

hajnalmt (changes requested):
Currently the images are rebuilt because we checkout the code in the  `e2e_*` again. Basically we are checking everything out twice, so all the timestamps differ from the ones in the `build_images` step. You shall move everything to the `e2e.yaml`  which can be done at the invoking job.

.github/workflows/e2e_cronjob.yaml
JesseStutler: Yes, so should we keep the musl if we don't know the reason?

Makefile
```diff
@@ -99,6 +101,23 @@ images:
 		docker buildx build -t "${IMAGE_PREFIX}/vc-$$name:$(TAG)" . -f ./installer/dockerfile/$$name/Dockerfile --output=type=${BUILDX_OUTPUT_TYPE} --platform ${DOCKER_PLATFORMS} --build-arg APK_MIRROR=${APK_MIRROR} --build-arg OPEN_EULER_IMAGE_TAG=${OPEN_EULER_IMAGE_TAG}; \
```
hajnalmt, line 101: Alternatively to not rebuild the images, you can add a check here, to not rebuild if the tag already exists.
You can introduce a new variable to force the image build (like FORCE_REBUILD) which is true by default and start the target with false `make images FORCE_REBUILD=false` in the e2e tests, so you are backwards compatible with the old make target.

.github/workflows/e2e_cronjob.yaml
hajnalmt: Since we are not using it, we should delete the step I think. It just occupies the space and CI time for no reason.

.github/workflows/images.yaml
lekaf974, line 1: Should this file name be more explicit `docker_images.yaml` for example ?

JesseStutler:
@Deepam02 Hi, how about the progress?

hajnalmt:
Hello @Deepam02,

Thanks for your great efforts on this issue — you've climbed a big part of the mountain! Unfortunately, this PR went well beyond "good first issue" territory. That's on me for underestimating the scope, and I hope it hasn't discouraged you from contributing to open source.

I've squashed your changes into https://github.com/volcano-sh/volcano/pull/5081 and credited your contribution in that PR. Hopefully we can get it merged there.

Thanks again for your work on this!

Issue #4781, opened 2025-12-04
Enables role-based elasticity and Gang Scheduling for distributed inference workloads (Kthena)


wangyang0616:
### What is the problem you're trying to solve

Each server group of Kthena corresponds to one PodGroup for inference instance-level Gang Scheduling, and each role instance maps to a subgroup.Kthena expects the scheduler to support role-level Gang Scheduling capabilities. For example, for a 4P2D inference instance, Kthena requires that the scheduling can only be triggered when resources for at least 2P1D are satisfied; otherwise, scheduling shall be prohibited.

When creating Pods, Kthena distinguishes the role information to which a Pod belongs using fixed label keys. The corresponding values are used to identify the role of the Pod and the ID of the associated P and D instances.
The label keys are: "modelserving.volcano.sh/role", "modelserving.volcano.sh/role-id".

Currently, Volcano implements automatic Pod grouping in subgroups via MatchPolicy. However, this approach cannot effectively support Kthena's differentiated configurations for P and D roles, nor can it reliably distinguish which Pods belong to P-1, P-2, D-1, or D-2.

### Describe the solution you'd like

Provide a mechanism to explicitly define the scope of Pods applicable to the current MatchPolicy, and then group Pods by label keys—Pods with the same value are automatically assigned to the same group.

### Documentation Updates

- [ ] This feature requires design or user documentation changes.
- [ ] If documentation changes are required, I will ensure the relevant documents are updated and published to the Volcano official website (https://volcano.sh) via the [volcano-sh/website](https://github.com/volcano-sh/website) repository.

wangyang0616:
kthena: [#532 ](https://github.com/volcano-sh/kthena/issues/532)

PR #4785, opened 2025-12-08
SubGroupPolicy enables optimized Kthena subgroup-based Gang Scheduling 

wangyang0616:
Fixes [#4781 ](https://github.com/volcano-sh/volcano/issues/4781)
Related Apis [#202 ](https://github.com/volcano-sh/apis/pull/202)

1. After optimizing SubGroupPolicy, update the grouping configuration and label configuration of Volcano Job during Pod creation to align with the policy;
2. Optimize the PodGroup grouping logic to match the updated SubGroupPolicy rules.

### DIFF
--- a/pkg/controllers/apis/job_info.go
+++ b/pkg/controllers/apis/job_info.go
@@ -55,8 +55,7 @@ func (ji *JobInfo) Clone() *JobInfo {
 	for key, pods := range ji.Pods {
 		job.Pods[key] = make(map[string]*v1.Pod, len(pods))
 		for pn, pod := range pods {
-			clonedPod := pod.DeepCopy()
-			job.Pods[key][pn] = clonedPod
+			job.Pods[key][pn] = pod
 		}
 	}
 
--- a/pkg/scheduler/api/sub_job_info_test.go
+++ b/pkg/scheduler/api/sub_job_info_test.go
@@ -411,7 +411,7 @@ func TestGetSubJobMatchValues_Error(t *testing.T) {
 					MatchExpressions: []metav1.LabelSelectorRequirement{
 						{
 							Key:      "app",
-							Operator: "invalid", // 无效的操作符
+							Operator: "invalid",
 						},
 					},
 				},
@@ -456,7 +456,7 @@ func TestGetSubJobMatchValues_Error(t *testing.T) {
 				MatchLabelKeys: []string{"any"},
 			},
 			pod: &v1.Pod{
-				ObjectMeta: metav1.ObjectMeta{}, // 无标签
+				ObjectMeta: metav1.ObjectMeta{},
 			},
 			expected: nil,
 		},


pkg/controllers/apis/job_info.go
```diff
@@ -57,7 +55,8 @@ func (ji *JobInfo) Clone() *JobInfo {
 	for key, pods := range ji.Pods {
 		job.Pods[key] = make(map[string]*v1.Pod, len(pods))
 		for pn, pod := range pods {
-			job.Pods[key][pn] = pod
+			clonedPod := pod.DeepCopy()
+			job.Pods[key][pn] = clonedPod
 		}
 	}
```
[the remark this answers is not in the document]
wangyang0616: done

pkg/scheduler/api/sub_job_info.go
```diff
@@ -144,20 +147,52 @@ func (sji *SubJobInfo) getTaskHighestPriority() int32 {
...
+
+		klog.V(4).Infof("Pod %s/%s matched MatchLabelKeys %v with values %v", pod.Namespace, pod.Name, policy.MatchLabelKeys, matchValues)
+		return matchValues
 	}
-	return values
+
+	// Log when no matching rules are configured, using default group
+	klog.V(4).Infof("No MatchLabelKeys configured for policy, pod %s/%s uses default subjob group", pod.Namespace, pod.Name)
+	return matchValues
 }
```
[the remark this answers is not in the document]
wangyang0616: done

pkg/controllers/job/job_controller_util_test.go
```diff
@@ -63,296 +62,426 @@ func TestMakePodName(t *testing.T) {
...
+			name:          "partition size 1, index 0",
+			partitionSize: 1,
+			index:         0,
+			expectedPart:  "0",
+		},
+		{
+			name:          "partition size 0 (invalid), index 5",
+			partitionSize: 0,
+			index:         5,
+			expectedPart:  "-1", // Go returns 0 for division by zero
```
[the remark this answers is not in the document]
wangyang0616: done

pkg/scheduler/api/sub_job_info_test.go
```diff
@@ -276,154 +276,196 @@ func TestSubJobInfo_IsSoftTopologyMode(t *testing.T) {
...
+			},
+			expected: nil,
+		},
+		{
+			name: "pod has no labels",
+			policy: scheduling.SubGroupPolicySpec{
+				MatchLabelKeys: []string{"any"},
+			},
+			pod: &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{}, // 无标签
```
JesseStutler: Please delete the Chinese comment

pkg/scheduler/api/sub_job_info_test.go
```diff
@@ -276,154 +276,196 @@ func TestSubJobInfo_IsSoftTopologyMode(t *testing.T) {
...
+	}{
+		{
+			name: "invalid label selector",
+			policy: scheduling.SubGroupPolicySpec{
+				LabelSelector: &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{
 						{
-							LabelKey: "key2",
+							Key:      "app",
+							Operator: "invalid", // 无效的操作符
```
JesseStutler: Same as above. Please check all the Chinese comments if we don't need it

pkg/controllers/apis/job_info.go
```diff
@@ -57,7 +55,8 @@ func (ji *JobInfo) Clone() *JobInfo {
 	for key, pods := range ji.Pods {
 		job.Pods[key] = make(map[string]*v1.Pod, len(pods))
 		for pn, pod := range pods {
-			job.Pods[key][pn] = pod
+			clonedPod := pod.DeepCopy()
```
hzxuzhonghu: I donot see we mutate the pod. Deepcopy is a costy op, so i think maybe we can keep it as still

JesseStutler:
Just want to verify that does this PR means that the pods must first match the labelSelector, and then be divided into different groups based on different keys? @wangyang0616

hzxuzhonghu (commented):
LGTM, not big problem, since this is mostly the logic updating subJob construction, while the scheduling has already been merged

wangyang0616:
> Just want to verify that does this PR means that the pods must first match the labelSelector, and then be divided into different groups based on different keys? @wangyang0616

Yes, LabelSelector is an explicit condition that defines the valid scope for pod grouping. When the controller creates a PodGroup based on a VCJob, it automatically configures the LabelSelector information to match the corresponding tasks.

pkg/controllers/apis/job_info.go
wangyang0616: You're right. The change has been reverted.

pkg/scheduler/api/sub_job_info_test.go
wangyang0616: done

pkg/scheduler/api/sub_job_info_test.go
wangyang0616: done

pkg/scheduler/api/sub_job_info.go
```diff
@@ -144,20 +147,52 @@ func (sji *SubJobInfo) getTaskHighestPriority() int32 {
...
+			matchValues = append(matchValues, value)
+		}
+
+		klog.V(4).Infof("Pod %s/%s matched MatchLabelKeys %v with values %v", pod.Namespace, pod.Name, policy.MatchLabelKeys, matchValues)
+		return matchValues
 	}
-	return values
+
+	// Log when no matching rules are configured, using default group
+	klog.V(4).Infof("No MatchLabelKeys configured for policy, pod %s/%s uses default subjob group", pod.Namespace, pod.Name)
```
ouyangshengjia, line 194: If a pod matches the `LabelSelector` but is missing some labels in `MatchLabelKeys`, then it will belongs to the default subjob, right?

ouyangshengjia: I think it is better that if a `SubGroupPolicy` is configured with only `LabelSelector` and empty `MatchLabelKeys`, the pods matching the `LabelSelector` should be assigned to a separate SubJob.

pkg/controllers/apis/job_info.go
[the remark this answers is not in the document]
wangyang0616: This part does not involve logic related to the pod's content. Additionally, the performance and resource costs of cloning are relatively high, so the existing logic will be retained for now without modifications.

pkg/scheduler/api/sub_job_info.go
wangyang0616: For the definition of subgroup, it is recommended to use LabelSelector and matchLabelKey in combination, and express the subgroup logic through explicit declaration. This avoids excessive implicit default behaviors and improves maintainability.

Issue #4788, opened 2025-12-08
allocate unit tests are failing


hajnalmt:
### Description

The unit tests are failing in the repositry currently:

```
E1208 14:12:42.695429   40798 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s1 and hyperNodeHighestAllowed s0" subJob="c1/pg1/task1-worker" hyperNode="s1" highestAllowedTier=1 allocatedHyperNode="s0"
E1208 14:12:43.153858   40798 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s5 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s5" highestAllowedTier=2 allocatedHyperNode="s1"
E1208 14:12:43.153979   40798 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s6 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s6" highestAllowedTier=2 allocatedHyperNode="s1"
E1208 14:12:43.156178   40798 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s2 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s2" highestAllowedTier=2 allocatedHyperNode="s1"
W1208 14:12:43.623592   40798 predicates.go:236] Devices GpuShare assertion conversion failed, skip
W1208 14:12:43.623635   40798 predicates.go:236] Devices hamivgpu assertion conversion failed, skip
W1208 14:12:43.623652   40798 predicates.go:236] Devices ascend310p-vNPU assertion conversion failed, skip
W1208 14:12:43.623790   40798 predicates.go:278] Devices GpuShare assertion conversion failed, skip
W1208 14:12:43.623811   40798 predicates.go:278] Devices hamivgpu assertion conversion failed, skip
W1208 14:12:43.623837   40798 predicates.go:278] Devices ascend310p-vNPU assertion conversion failed, skip
W1208 14:12:43.623941   40798 predicates.go:236] Devices GpuShare assertion conversion failed, skip
[...]

PR #4789, opened 2025-12-09
fix vgpu warning

DSFans2014:
as shown in #4788, If the vGPU switch is not enabled, devices will not be added to `ni.Other`, but  the device names already included in `RegisteredDevices` at the time of its definition.  so there are some warnings as below

https://github.com/volcano-sh/volcano/blob/b72ae8b96f7a6b03ea4c0f3f8a0ccc872e5442b5/pkg/scheduler/plugins/predicates/predicates.go#L220-L233

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/deviceshare.go
+++ b/pkg/scheduler/plugins/deviceshare/deviceshare.go
@@ -28,7 +28,6 @@ import (
 	fwk "k8s.io/kube-scheduler/framework"
 
 	"volcano.sh/volcano/pkg/scheduler/api"
-	"volcano.sh/volcano/pkg/scheduler/api/devices"
 	"volcano.sh/volcano/pkg/scheduler/api/devices/ascend/hami"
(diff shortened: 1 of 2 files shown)


archlitchi (approved):
/lgtm

pkg/scheduler/plugins/deviceshare/deviceshare.go
```diff
@@ -205,15 +213,6 @@ func (dp *deviceSharePlugin) OnSessionOpen(ssn *framework.Session) {
 		for _, val := range api.RegisteredDevices {
 			if dev, ok := node.Others[val].(api.Devices); ok {
 				if reflect.ValueOf(dev).IsNil() {
-					// TODO When a pod requests a device of the current type, but the current node does not have such a device, an error is thrown
```
JesseStutler, line 208: /cc @archlitchi Could we delete this part of the code?

archlitchi: Yes, it can be safely deleted

PR #4790, opened 2025-12-09
fix hypernode ut error logs

3sunny:
as shown in https://github.com/volcano-sh/volcano/issues/4788, When assigning hypernode A to a subJob, if no intersection can be found between hypernode A and the already assigned hypernode B, an error log will be output. This can be changed to a warning log.

### DIFF
--- a/pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
+++ b/pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
@@ -221,8 +221,8 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
 		if hardMode, highestAllowedTier := subJob.IsHardTopologyMode(); hardMode {
 			result, err := nta.hyperNodeGradientFn(ssn, hyperNode, highestAllowedTier, subJob.AllocatedHyperNode)
 			if err != nil {
-				klog.ErrorS(err, "build hyperNode gradient fail", "subJob", subJob.UID, "hyperNode", hyperNode.Name,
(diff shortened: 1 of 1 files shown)


pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -221,8 +221,8 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
 		if hardMode, highestAllowedTier := subJob.IsHardTopologyMode(); hardMode {
 			result, err := nta.hyperNodeGradientFn(ssn, hyperNode, highestAllowedTier, subJob.AllocatedHyperNode)
 			if err != nil {
-				klog.ErrorS(err, "build hyperNode gradient fail", "subJob", subJob.UID, "hyperNode", hyperNode.Name,
```
JesseStutler, line 224: So if we meet an error when building hyperNodeGradientFn, it doesn't influence our scheduling logic, we still can normally schedule pods, right?

3sunny: There are two scenarios here. If we encounter an error when building the hyperNodeGradientFn for a Job, it will cause the scheduling to fail. This time, the error occurs when building the hyperNodeGradientFn for a subJob, but it does not affect our scheduling logic. We can still schedule the subJob normally to other suitable hypernodes.

JesseStutler: So it's a little bit strange to use warning log when we can't schedule jobs

hajnalmt: i am fine with the error logs if it's not failing the unit test. We can leave it this way.

Issue #3997, opened 2025-02-05
Reclaim: Have the ability to rank victim nodes based on what's running in them


raravena80:
### What is the problem you're trying to solve

A follow to this thread: https://cloud-native.slack.com/archives/C011GJDQS0N/p1738726065356349?thread_ts=1738361347.510989&cid=C011GJDQS0N

Reclaim goes note by node without having any ranking function in terms of what's running in them.  Is there a way to rank the nodes based on what's running in them. For example if tasks from jobs that are in lower priority queues?

### Describe the solution you'd like

For example before running this: https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/actions/reclaim/reclaim.go#L141

Could we rank the nodes in a certain order?

Thanks!

JesseStutler:
This is a useful question that needs to be enhanced in v1.12 or in the patch version of v1.11, we can filter through all the victims first and then prioritize them accordingly

JesseStutler:
cc @Monokaix, this should be planned for v1.12 milestone?

raravena80:
Btw, we experienced a similar issue with preemption in where it didn't have context of what was running in the nodes. I'm just adding a comment here in case the solutions are the same.

This is what we tried to fix here https://github.com/volcano-sh/volcano/pull/3960 but there may be a better fix than that.

JesseStutler:
> Btw, we experienced a similar issue with preemption in where it didn't have context of what was running in the nodes. I'm just adding a comment here in case the solutions are the same.
> 
> This is what we tried to fix here [#3960](https://github.com/volcano-sh/volcano/pull/3960) but there may be a better fix than that.

In preempt action?

raravena80:
> In preempt action?

Correct.

sfc-gh-raravena:
Something that might be useful here too is if we allow reclaim to not happen after scoring the nodes and finding out that all jobs are gang jobs.

elysium-w:
@sfc-gh-raravena  I will finish it,I'm working on another reclaim issue

Monokaix:
close to https://github.com/volcano-sh/volcano/issues/3998

JesseStutler:
@elysium-w Hi, any progress?

blazethunderstorm:
@JesseStutler can I pls work on this

JesseStutler:
@blazethunderstorm 
Do you still work on this? We may priorite this feature

blazethunderstorm:
[...]

PR #4794, opened 2025-12-10
Refactor reclaim action

### LINKED ISSUE #3738: Need to refactor the reclaim action

guoqinwill:
#### What type of PR is this?
/kind bug

origin pr：https://github.com/volcano-sh/volcano/pull/4407

Fixes #3738 

```release-note

```

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -27,14 +27,19 @@ import (
 	"k8s.io/klog/v2"
 
 	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
-type Action struct{}
+type Action struct {
+	enablePredicateErrorCache bool
+}
 
 func New() *Action {
-	return &Action{}
+	return &Action{
+		enablePredicateErrorCache: true,
+	}
 }
 
 func (ra *Action) Name() string {
@@ -43,9 +48,16 @@ func (ra *Action) Name() string {
 
 func (ra *Action) Initialize() {}
 
+func (ra *Action) parseArguments(ssn *framework.Session) {
+	arguments := framework.GetArgOfActionFromConf(ssn.Configurations, ra.Name())
+	arguments.GetBool(&ra.enablePredicateErrorCache, conf.EnablePredicateErrCacheKey)
+}
+
 func (ra *Action) Execute(ssn *framework.Session) {
 	klog.V(5).Infof("Enter Reclaim ...")
 	defer klog.V(5).Infof("Leaving Reclaim ...")
+
+	ra.parseArguments(ssn)
 
 	queues := util.NewPriorityQueue(ssn.QueueOrderFn)
 	queueMap := map[api.QueueID]*api.QueueInfo{}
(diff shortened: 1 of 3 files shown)


pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +91,141 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
 			}
 
-			if len(reclaimees) == 0 {
-				klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
-				continue
+			if assignedInJob {
+				jobsQ.Push(job)
 			}
+			queues.Push(queue)
+		}
```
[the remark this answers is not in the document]
guoqinwill: When jobsQ is empty, break immediately. Also break when tasksQ is empty or !JobStarving.
After queues are pushed back in the inner loop, the outer loop will pop them again; however, when jobsQ is already empty, the queue will no longer be pushed back.
Therefore, the number of iterations is bounded by the number of elements in the queue, and in theory an infinite loop cannot occur. There may only be some unnecessary extra iterations (the queue is pushed back, but then quickly found to contain no jobs).
However, this can be optimized: if the queue’s jobsQ is already empty, we can skip queues.Push(queue) to avoid one meaningless outer-loop iteration.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +91,141 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
 
+		victimsQueue := ssn.BuildVictimsPriorityQueue(victims, task)
+		resreq := task.InitResreq.Clone()
+		reclaimed := api.EmptyResource()
+
+		// The reclaimed resources should be added to the remaining available resources of the nodes to avoid over-reclaiming.
+		availableResources := n.FutureIdle()
+
+		klog.Infof("xxxx the victimsQueue is %+v", victimsQueue)
```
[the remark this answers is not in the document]
guoqinwill: yes

JesseStutler:
Also /assign @hajnalmt mate is also familiar with the reclaim logic

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +91,142 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+		reclaimed := api.EmptyResource()
 
+		// The reclaimed resources should be added to the remaining available resources of the nodes to avoid over-reclaiming.
+		availableResources := n.FutureIdle()
+
+		for !victimsQueue.Empty() {
+			reclaimee := victimsQueue.Pop().(*api.TaskInfo)
+			klog.V(3).Infof("Try to reclaim Task <%s/%s> for Tasks <%s/%s>",
+				reclaimee.Namespace, reclaimee.Name, task.Namespace, task.Name)
+			if err := ssn.Evict(reclaimee, "reclaim"); err != nil {
```
JesseStutler: We should not keep using `Evict` and `Pipeline` in session, we should use these two methods in statement instead, just like in preempt action: https://github.com/volcano-sh/volcano/blob/050fb4eaa0ee2e8d9023fe08f2435e4d876762ab/pkg/scheduler/actions/preempt/preempt.go#L215-L219
And we should check whether the job is pipelined, if it's pipelined, then the statement can be sumitted

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +91,142 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
 				break
 			}
 		}
 
-		if assigned {
-			jobs.Push(job)
+		klog.V(3).Infof("Reclaimed <%v> for task <%s/%s> requested <%v>, and Node <%s> availableResources <%v>.", reclaimed, task.Namespace, task.Name, task.InitResreq, n.Name, availableResources)
+
+		if task.InitResreq.LessEqual(availableResources, api.Zero) {
+			if err := ssn.Pipeline(task, n.Name); err != nil {
```
JesseStutler: Same as above, we should Pipeline the statement instead, to introduce the transaction logic

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -255,6 +255,39 @@ func TestReclaim(t *testing.T) {
...
+				util.BuildPod("c1", "preemptor-task2-preemptable", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg-preemptor", make(map[string]string), make(map[string]string)),
+			},
+			Nodes: []*v1.Node{
+				util.BuildNode("n1", api.BuildResourceList("2", "2G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+			},
+			Queues: []*schedulingv1beta1.Queue{
+				util.BuildQueue("q1", 1, nil),
+				util.BuildQueue("q2", 1, nil),
+			},
+			ExpectEvictNum: 1, // The old code would result in 0 evictions. The correct code will result in 1.
```
hajnalmt: Remove the comment here.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +91,142 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+	}
+}
+
+func (ra *Action) reclaimForTask(ssn *framework.Session, task *api.TaskInfo, job *api.JobInfo) bool {
+	totalNodes := ssn.FilterOutUnschedulableAndUnresolvableNodesForTask(task)
+	assigned := false
 
-			victims := ssn.Reclaimable(task, reclaimees)
+	for _, n := range totalNodes {
+		if err := ssn.PredicateForPreemptAction(task, n); err != nil {
```
JesseStutler: We can use PredicateNodes to predicate nodes concurrently just like in preempt action: https://github.com/volcano-sh/volcano/blob/050fb4eaa0ee2e8d9023fe08f2435e4d876762ab/pkg/scheduler/actions/preempt/preempt.go#L302

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +91,142 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
-				} else if j.Queue != job.Queue {
-					q := ssn.Queues[j.Queue]
-					if !q.Reclaimable() {
-						continue
-					}
-					// Clone task to avoid modify Task's status on node.
-					reclaimees = append(reclaimees, task.Clone())
 				}
+
+				assignedInJob = assignedInJob || ra.reclaimForTask(ssn, task, job)
```
JesseStutler: I don't understand what does `assignedInJob` mean and what is this variable used for, could you clarify? Why would us push back the job into the jobsQ queue only `assignedInJob` is true?

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +91,142 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+				// Pick up all its candidate tasks.
+				tasksQ, ok := preemptorTasks[job.UID]
+				if !ok || tasksQ.Empty() || !ssn.JobStarving(job) {
+					klog.V(3).Infof("No preemptor task in job <%s/%s>.",
+						job.Namespace, job.Name)
+					break
+				}
 
-			klog.V(3).Infof("Considering Task <%s/%s> on Node <%s>.", task.Namespace, task.Name, n.Name)
+				klog.V(3).Infof("Try reclaim for %d tasks of job <%s/%s>", tasksQ.Len(), job.Namespace, job.Name)
```
hajnalmt: Sry for the nit-picking but these are relatively verbose log lines, can you end them with a '.' to be precise?

Additionally, here we don't try to reclaim anything yet.
A better wording is:
`klog.V(3).Infof("Considering reclaim for %d tasks of job <%s/%s>", tasksQ.Len(), job.Namespace, job.Name)`

pkg/scheduler/actions/reclaim/reclaim.go
guoqinwill: assignedInJob indicates whether, in the current round, at least one pending task from this Job successfully completed reclaim and was pipelined onto a node.If no task is successfully assigned during the inner loop, assignedInJob remains false.A Job is pushed back into jobsQ only when assignedInJob == true, for the following reasons:
When at least one task was successfully assigned, the Job may still have other pending tasks that need further reclaim attempts.Re-queuing it allows the Job to continue competing in later rounds.
If no task was successfully assigned, it means the current resources/conditions cannot satisfy this Job.
Re-queuing it would only waste cycles and may cause unnecessary spinning, so it is not pushed back.
At the same time, this behavior is aligned with the logic used in preempt.
However, what confuses me is that the preempt code does not seem to address the issue raised by the gemini-code-assist: The assignedInJob flag is overwritten in each iteration of the task loop. This means it only reflects the assignment status of the very last task, not whether any task in the job was assigned. It should be updated to true if any task is successfully reclaimed and not be reset within the loop for a given job.
Therefore, I modified the logic here so that the condition for updating assignedInJob cannot be overwritten.

pkg/scheduler/actions/reclaim/reclaim.go
guoqinwill: ok

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -31,10 +31,14 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
-type Action struct{}
+type Action struct {
+	enablePredicateErrorCache bool
```
hajnalmt, line 36: This should be configurable in the action.
The time has come to introduce the arguments in the action like in preempt:
https://github.com/volcano-sh/volcano/blob/6147e00c1741ecc45f4d707094a606ff6abb13df/pkg/scheduler/actions/preempt/preempt.go#L92-L107

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -39,6 +42,12 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
+func init() {
```
hajnalmt, line 43: We are not enabling level level 4 verbosity in other unit test suites either, and this function is not called currently anywhere.

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -238,7 +247,7 @@ func TestReclaim(t *testing.T) {
 			},
 			Pods: []*v1.Pod{
 				util.BuildPod("c1", "preemptee1-1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
-				util.BuildPod("c1", "preemptee1-2", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				util.BuildPod("c1", "preemptee1-2", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "false"}, make(map[string]string)),
```
hajnalmt: Please not modify this it will hopefully be solved with:
https://github.com/volcano-sh/volcano/pull/4800

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -255,6 +264,39 @@ func TestReclaim(t *testing.T) {
...
+				util.BuildPod("c1", "preemptor-task2-preemptable", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg-preemptor", make(map[string]string), make(map[string]string)),
+			},
+			Nodes: []*v1.Node{
+				util.BuildNode("n1", api.BuildResourceList("2", "2G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+			},
+			Queues: []*schedulingv1beta1.Queue{
+				util.BuildQueue("q1", 1, nil),
+				util.BuildQueue("q2", 1, nil),
+			},
+			ExpectEvictNum: 1, // The old code would result in 0 evictions. The correct code will result in 1.
```
hajnalmt: remove the comment here

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +95,152 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+			for {
+				// Pick up all its candidate tasks.
+				tasksQ, ok := preemptorTasks[job.UID]
+				if !ok || tasksQ.Empty() || !ssn.JobStarving(job) {
+					klog.V(3).Infof("No preemptor task in job <%s/%s>.",
+						job.Namespace, job.Name)
+					break
+				}
+
+				klog.V(3).Infof("Try reclaim for %d tasks of job <%s/%s>", tasksQ.Len(), job.Namespace, job.Name)
```
hajnalmt: Sry for the nit-picking but these are relatively verbose log lines, can you end them with a dot '.' to be precise?

Additionally, here we don't try to reclaim anything yet.
A better wording is:
`klog.V(3).Infof("Considering reclaim for %d tasks of job <%s/%s>.", tasksQ.Len(), job.Namespace, job.Name)`

pkg/scheduler/actions/reclaim/reclaim.go
guoqinwill: yes，we  can do this

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +103,152 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+				break
 			}
+			job := jobsQ.Pop().(*api.JobInfo)
+			stmt := framework.NewStatement(ssn)
+
+			var taskPipelinedInJob bool
+			for {
+				// Pick up all its candidate tasks.
+				tasksQ, ok := preemptorTasks[job.UID]
+				if !ok || tasksQ.Empty() || !ssn.JobStarving(job) {
```
JesseStutler: Can separate the `JobStarving` judgement and log them separately. That would be clearer.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +103,152 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
-				klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
-				continue
+			if ssn.JobPipelined(job) {
+				stmt.Commit()
+			} else {
+				stmt.Discard()
 			}
 
-			victims := ssn.Reclaimable(task, reclaimees)
+			if taskPipelinedInJob {
```
JesseStutler: I still feel incorrect about the `taskPipelinedInJob` here, reclaim is different from allocate, it shouldn't be an overly aggressive action. If thetasks in the job already relcaimed minAvailable resources, it should not keep reclaiming, if it needs to reclaim the whole job, the user needs to set minAvailable == replicas. So why should we push back the job into the queue? We should try the next job to reclaim

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +103,152 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+		if task.InitResreq.LessEqual(availableResources, api.Zero) {
+			if err := stmt.Pipeline(task, n.Name, evictionOccurred); err != nil {
+				klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>",
+					task.Namespace, task.Name, n.Name)
+				if rollbackErr := stmt.UnPipeline(task); rollbackErr != nil {
+					klog.Errorf("Failed to unpipeline Task %v on %v in Session %v for %v.",
+						task.UID, n.Name, ssn.UID, rollbackErr)
+				}
+			}
+			assigned = true
```
JesseStutler: Just a reminder that here `assigned` should be renamed like `reclaimed`, but I still feel this variable is unuseful in reclaim scenario

pkg/scheduler/actions/reclaim/reclaim.go
guoqinwill: Okay, then we won’t keep the taskPipelinedInJob check in reclaim. I will delete it

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -91,148 +103,152 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+	}
+}
+
+func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Statement, task *api.TaskInfo, job *api.JobInfo) bool {
+	totalNodes := ssn.FilterOutUnschedulableAndUnresolvableNodesForTask(task)
+	predicateHelper := util.NewPredicateHelper()
+	predicateNodes, _ := predicateHelper.PredicateNodes(task, totalNodes, ssn.PredicateForPreemptAction, false)
+	assigned := false
+
+	for _, n := range predicateNodes {
```
JesseStutler, line 180: https://github.com/volcano-sh/volcano/issues/3997 Currently, reclaim is done by traversing nodes, instead of first filtering candidates by node and then sorting them. So, I'm wondering if we can fix this issue together? Imagine this worst-case scenario: each job's pods are scattered across different nodes, and then a job wants to reclaim, it might reclaim one pod from each of the other jobs. If we sort by priority, this would make it easier for pods from lower-priority jobs to be reclaimed first.

pkg/scheduler/actions/reclaim/reclaim.go
hajnalmt: I understand the concern, but I think this logic is right here Jesse.

We have a check in the tasks loop at line 130:
```
if !ok || tasksQ.Empty() || !ssn.JobStarving(job) {
```
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/gang/gang.go#L202
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/api/job_info.go#L1103
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/api/job_info.go#L840

This counts the Pipelined tasks in the jobs too so the newly reclaimed resources will be counted in and if the Job is not starving anymore, it won't reclaim more tasks. This prevents the aggressive behaviour you described and excessive reclaim in general.

If we don't have this here then if a higher priority job is really starving (it needs more tasks to reach minAvailable), it will still reclaim only one pod in the session and lower priority jobs can highjack the reclaimable resources in this loop. When the next session comes the higher priority job will not be able to get the resources (the allocated resources might not be preemptable) even though it's still starving. This can also lead to resource fragmantation, which with we struggle a lot.

pkg/scheduler/actions/reclaim/reclaim.go
hajnalmt: We have a variable named `reclaimed` at line 213, that's why I suggested `taskPipelined` for this, but we can rename the other variable to `reclaimedResources` and name this to `reclaimed` maybe?
This variable indicates if there was a task pipelined in the function which is I think important to continue the reclaim in the other tasks in the job.

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -39,6 +40,10 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
+func init() {
```
hajnalmt, line 43: why did you leave the function definition here? Is it called somehwere?

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -229,32 +234,60 @@ func TestReclaim(t *testing.T) {
 			PriClass: []*schedulingv1.PriorityClass{
 				util.BuildPriorityClass("low-priority", 100),
 				util.BuildPriorityClass("mid-priority", 500),
-				util.BuildPriorityClass("high-priority", 1000),
 			},
 			PodGroups: []*schedulingv1beta1.PodGroup{
-				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 1, nil, schedulingv1beta1.PodGroupInqueue, "low-priority"),
```
hajnalmt: This test case was fixed on master please rebase, and don't modify it in your commit directly.

pkg/scheduler/actions/reclaim/reclaim.go
hajnalmt: I see there wasn't any progress in this Issue.

How do you imagine the implementation for this?
Do we want to create a `VictimNodeOrderFn` implemented in the nodeOrder plugin?

pkg/scheduler/actions/reclaim/reclaim.go
hajnalmt: I have slept on this one, and I came to the conclusion that regarding the variable Jesse is right.

We can just continue looping through the tasks in the job instead of this. I think the right logic here is to count the reclaimed tasks in the tasks loop and continue reclaiming until we don't have preemptorTasks in the job or the job is not starving anymore.

Basically I would introduce an `reclaimedTasksCount` variable in the tasks loop above, which I bump at each reclaim and add a guard at the end to continue looping:

```go
if !taskQ.Empty() && ssn.JobStarving(job) {
  klog.V(5).Infof("Job %s/%s> is still starving and has preemptorTasks in it. Number of reclaimed tasks: %v.", job.Namespace, job.Name, reclaimedTasksCount)
  continue
}
```

What do you think guys @JesseStutler @guoqinwill ? 

The other thing which came to my mind is that I am not 100% right in this excessive reclaim too, which Jesse described. Currently the JobStarving function is implemented in the gang, priority and tdm plugins:

https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/gang/gang.go#L199
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/priority/priority.go#L141
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/tdm/tdm.go#L289

First of all, this means that if these plugins are not present, reclaim won't work in general (because there are no starving tasks), but what's confusing for me is that we have an actual job isStarving implementation:
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/api/job_info.go#L1103

But we use different approaches in each plugin.
The one in the priority plugin:
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/priority/priority.go#L141
Will exactly do what Jesse described, since it just compares the tasks towards the `len(ji.Tasks)`, but in certain cases maybe this is what we want (there were questions and issues, about not reclaiming over minAvailable)? Also maybe this should be documented somewhere I don't know, this difference is surprising at least.

(The one in tdm will do kind of the same:
https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/plugins/tdm/tdm.go#L289
This means, that if a job has a Pending task, it's starving.)

What came to my mind, is that maybe this should be configurable in proportion and capacity since they are the main initiatiors for reclaim. What do you guys think?

hajnalmt: Let's discuss the starving issue separately as it's not in the focus of this refactor.:
https://github.com/volcano-sh/volcano/issues/4812

pkg/scheduler/actions/reclaim/reclaim.go
guoqinwill: I think the modifications to this issue can be split up for now, and we can focus on refactoring the reclaim process first.

pkg/scheduler/actions/reclaim/reclaim_test.go
guoqinwill: ok

pkg/scheduler/actions/reclaim/reclaim_test.go
guoqinwill: Because `predicateHelper.PredicateNodes` was introduced, the unit test needs to initialize `options.ServerOpts`; otherwise, it will fail at runtime.

pkg/scheduler/actions/reclaim/reclaim.go
guoqinwill: done

pkg/scheduler/actions/reclaim/reclaim_test.go
hajnalmt: Oh understood thank you!

pkg/scheduler/actions/reclaim/reclaim.go
hajnalmt: I have tried to go through these issues and what Jesse wants here I think is to create a `reclaimeeNodeMap` which contains all the reclaimees by nodes that passes the _ValidateVictims_ check, after that we can sort the tasks by priority and pick the one with the least priority, and reclaim on that node. Do I guess this right @JesseStutler ?

What came to my mind is when the capacity plugin queue hierarchy handling is enabled we should sort by queue first, and then by taskOrder because picking a node first does not guarantee that we will choose a resource from the queue most close in distance when hierarchy is enabled.

This function will kind of do the queue ordering properly (but only if the proper node is selected...)
`victimsQueue := ssn.BuildVictimsPriorityQueue(victims, task)`
Disregarding the task priority in the end for queues on the same level 😞 

I think we shall discuss this separately on the community meeting.

hajnalmt (approved):
/lgtm

This reimplementation is a good first step I think, but we still need to have a lot of improvement on this.

pkg/scheduler/actions/reclaim/reclaim.go
JesseStutler: Yes, but after discussing with @guoqinwill, I think my initial considerations weren't comprehensive enough. If we sort by queue first, and the reclaimer requires more resources, then the victims might be distributed across multiple nodes (after sorting by queue priority). This would be incorrect, as a pod cannot be distributed across multiple nodes. Therefore, we need to consider various factors (such as whether multiple victims are on the same node, etc.). We can optimize this later. @hajnalmt

Issue #4798, opened 2025-12-11
Reclaim UTs fail intermittently


wangyang0616:
### Description

After [#4774 ](https://github.com/volcano-sh/volcano/pull/4774) was merged, unit tests (UTs) failed sporadically

<img width="1766" height="479" alt="Image" src="https://github.com/user-attachments/assets/dbefffb3-e519-4be4-9ee4-3dc208a7b471" />

### Steps to reproduce the issue

make unit-test

### Describe the results you received and expected

UTs pass consistently.

### What version of Volcano are you using?

master

PR #4800, opened 2025-12-11
fix unit test case:Reclaimed resources plus node available resources

jiahuat:
Due to the presence of the gang plugin, which requires maintaining minMember, pods are randomly filtered out, resulting in pods not being evicted as expected.simplify the test cases to avoid the influence of task priorities and only test whether available Node resources are included when the garbage collection task is executed.

#### What type of PR is this?
fix unit test

Due to the presence of the gang plugin, which requires maintaining minMember, pods are randomly filtered out, resulting in pods not being evicted as expected. Simplify the test cases to avoid the influence of task priorities and only test whether available Node resources are included when the garbage collection task is executed.

Fixes https://github.com/volcano-sh/volcano/issues/4798

```release-note

```

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim_test.go
+++ b/pkg/scheduler/actions/reclaim/reclaim_test.go
@@ -219,7 +219,7 @@ func TestReclaim(t *testing.T) {
 			ExpectEvicted:  []string{},
 		},
 		{
-			Name: "Reclaimed resources plus node available resources",
+			Name: "Node available resources should be included in reclaimed resources",
 			Plugins: map[string]framework.PluginBuilder{
 				conformance.PluginName: conformance.New,
 				gang.PluginName:        gang.New,
@@ -232,24 +232,22 @@ func TestReclaim(t *testing.T) {
 				util.BuildPriorityClass("high-priority", 1000),
 			},
 			PodGroups: []*schedulingv1beta1.PodGroup{
-				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
-				util.BuildPodGroupWithPrio("pg2", "c1", "q2", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 0, nil, schedulingv1beta1.PodGroupRunning, "mid-priority"),
+				util.BuildPodGroupWithPrio("pg2", "c1", "q2", 0, nil, schedulingv1beta1.PodGroupRunning, "mid-priority"),
 				util.BuildPodGroupWithPrio("pg3", "c1", "q3", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
 			},
 			Pods: []*v1.Pod{
 				util.BuildPod("c1", "preemptee1-1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
-				util.BuildPod("c1", "preemptee1-2", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
 				util.BuildPod("c1", "preemptee2-1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg2", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
-				util.BuildPod("c1", "preemptee2-2", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg2", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
-				util.BuildPod("c1", "preemptor1", "", v1.PodPending, api.BuildResourceList("4", "1G"), "pg3", make(map[string]string), make(map[string]string)),
+				util.BuildPod("c1", "preemptor1", "", v1.PodPending, api.BuildResourceList("2", "1G"), "pg3", make(map[string]string), make(map[string]string)),
 			},
 			Nodes: []*v1.Node{
-				util.BuildNode("n1", api.BuildResourceList("10", "4Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+				util.BuildNode("n1", api.BuildResourceList("10", "2Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
 			},
 			Queues: []*schedulingv1beta1.Queue{
 				util.BuildQueueWithPriorityAndResourcesQuantity("q1", 1, nil, nil),
 				util.BuildQueueWithPriorityAndResourcesQuantity("q2", 2, nil, nil),
-				util.BuildQueueWithPriorityAndResourcesQuantity("q3", 3, nil, nil),
+				util.BuildQueue("q3", 3, nil),
 			},
 			ExpectEvictNum: 1,
 			// cpu resource is enough in node, memory resource is not enough, need 1G memory to schedule preemptor1


pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -223,37 +223,27 @@ func TestReclaim(t *testing.T) {
...
-			PriClass: []*schedulingv1.PriorityClass{
-				util.BuildPriorityClass("low-priority", 100),
-				util.BuildPriorityClass("mid-priority", 500),
-				util.BuildPriorityClass("high-priority", 1000),
-			},
 			PodGroups: []*schedulingv1beta1.PodGroup{
-				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
-				util.BuildPodGroupWithPrio("pg2", "c1", "q2", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
-				util.BuildPodGroupWithPrio("pg3", "c1", "q3", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 1, nil, schedulingv1beta1.PodGroupInqueue, "low-priority"),
```
hajnalmt: If you don't specify priorityClasses, then you need to define these without priority then.
```
util.BuildPodGroup("pg1", "c1", "q1", 1, nil, schedulingv1beta1.Running)
util.BuildPodGroup("pg2", "c1", "q2", 1, nil, schedulingv1beta1.PodGroupInqueue)
```
and the first one is Running.

hajnalmt (changes requested):
Why haven't you just removed the gang plugin? My problem is that the new unit test case does not cover the original problem.

Because the test case won't fail (since there is only one pod that can be evicted, preemptee2 and preemptee3 and not preemtable) if your code https://github.com/volcano-sh/volcano/pull/4774 is not present.

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -223,37 +223,27 @@ func TestReclaim(t *testing.T) {
...
-				util.BuildPod("c1", "preemptee1-2", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
-				util.BuildPod("c1", "preemptee2-1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg2", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
-				util.BuildPod("c1", "preemptee2-2", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg2", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
-				util.BuildPod("c1", "preemptor1", "", v1.PodPending, api.BuildResourceList("4", "1G"), "pg3", make(map[string]string), make(map[string]string)),
+				util.BuildPod("c1", "preemptee1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				util.BuildPod("c1", "preemptee2", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "false"}, make(map[string]string)),
```
JesseStutler: If you set other pods to preemptable as false, then it's meaningless, because they can't be preempted anyway. Even without the changes in #4774, your current ut still can pass, because it always will preempt preemptee1 although the resources is not enough. The scenario we need to verify is that it won't preempt more running and preemptable pods. The reason your #4774 ut occasionally fails is because preemptee1-2 sometimes gets preempted before preemptee1-1, so ExpectEvicted is incorrect. So the correct way is you should give the pod a priority to ensure preemptee1-1 gets preempted first. @jiahuat

pkg/scheduler/actions/reclaim/reclaim_test.go
JesseStutler: Remove the gang plugin still can not ensure that preemptee1-1 can be preempted first, the correct way is that we should specify like higher priority for the preemptee1-1, then preemptee1-1 will always be preempted first then preemptee1-2

JesseStutler:
We can debug and check the `reclaimees` list, the victim candidates in this list is not ordered

pkg/scheduler/actions/reclaim/reclaim_test.go
hajnalmt: Yes, exactly. Prioritites are needed to test the original problem properly (I just wanted to highlight this issue, in the current PR too).

jiahuat:
To eliminate the minMember limitation of the gang plugin, the minMember of pg1 and pg2 is set to 0; preemptors require 2c 1G, and the node CPU is sufficient, so only 1G needs to be reclaimed. Prioritized by low priority, the expected reclamation is preemptee1-1.

jiahuat:
In reclaim.go, when Reclaimable is called on line 184: victims := ssn.Reclaimable(task, reclaimees), since reclaimees are unordered, the gang plugin may retain the lowest priority task to satisfy minMember; this violates the principle of recycling from the lowest priority; or the reclaimees could be sorted in reverse order of priority before calling Reclaimable or just in gang plugin.

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -232,24 +232,36 @@ func TestReclaim(t *testing.T) {
...
 			},
 			Nodes: []*v1.Node{
-				util.BuildNode("n1", api.BuildResourceList("10", "4Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+				util.BuildNode("n1", api.BuildResourceList("10", "2Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
 			},
 			Queues: []*schedulingv1beta1.Queue{
-				util.BuildQueueWithPriorityAndResourcesQuantity("q1", 1, nil, nil),
-				util.BuildQueueWithPriorityAndResourcesQuantity("q2", 2, nil, nil),
-				util.BuildQueueWithPriorityAndResourcesQuantity("q3", 3, nil, nil),
+				func() *schedulingv1beta1.Queue {
```
hajnalmt: What was the problem with the original code here?
Reclaimable is true by default, these new functions are not doing anything.

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -232,24 +232,36 @@ func TestReclaim(t *testing.T) {
 				util.BuildPriorityClass("high-priority", 1000),
 			},
 			PodGroups: []*schedulingv1beta1.PodGroup{
-				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
-				util.BuildPodGroupWithPrio("pg2", "c1", "q2", 1, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 0, nil, schedulingv1beta1.PodGroupInqueue, "mid-priority"),
```
hajnalmt: pq1, pq2 shall be `PodGroupRunning` (if you have a pod Running in the Podgroup)

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -219,7 +219,7 @@ func TestReclaim(t *testing.T) {
 			ExpectEvicted:  []string{},
 		},
 		{
-			Name: "Reclaimed resources plus node available resources",
+			Name: "Reclaimed resources should plus node available resources",
```
hajnalmt: This is not a valid english sentence.
Please change it to something like:
`Node available resources should be included in reclaimed resources`
`Reclaimed resources should count in the node available resources`

pkg/scheduler/actions/reclaim/reclaim_test.go
jiahuat: Thanks for the reminder. I was initially unsure if Reclaimable is true by default, so I set it explicitly. I'll revert it.

jiahuat: I will modify the code to: `util.BuildQueueWithPriorityAndResourcesQuantity("q1", 1, nil, nil)`, `util.BuildQueueWithPriorityAndResourcesQuantity("q2", 2, nil, nil)`, `util.BuildQueue("q3", 3, nil)`. Because the proportion plugin is used, `q3` needs more weight to trigger preemption.

JesseStutler:
OK, now I think this test is more reasonable
/approve
Let @hajnalmt also take a look

PR #4804, opened 2025-12-11
Create a new Agent scheduler for Agent workload fast scheduling

qi-min:
#### What type of PR is this?
feature
Add a new scheduler to execute fast path scheduling for workload like AI agent. The Scheduler is focus fast scheduling and scheduling based on Sharding.  This PR include basic function to schedule Agent.
For more details please refer to: [proposal](https://docs.google.com/document/d/1NDTEGulhQ__sdZ_c2kN3ojtJIZie5O7_GyZyKIOLz-g/edit?tab=t.0) | [design doc](https://github.com/qi-min/volcano/blob/agent-scheduler-doc/docs/design/agent-scheduler.md)

Fixes #
https://github.com/volcano-sh/volcano/issues/4722

```release-note

```

### DIFF
--- a/cmd/agent-scheduler/app/options/options.go
+++ b/cmd/agent-scheduler/app/options/options.go
@@ -61,7 +61,7 @@ type ServerOption struct {
 	CertData          []byte
 	KeyData           []byte
 	CaCertData        []byte
-	SchedulerNames    []string
+	SchedulerName     string
 	SchedulerConf     string
 	ResyncPeriod      time.Duration
 	// leaderElection defines the configuration of leader election.
@@ -117,7 +117,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.LockObjectNamespace, "lock-object-namespace", "", "Define the namespace of the lock object; it is volcano-system by default.")
 	fs.MarkDeprecated("lock-object-namespace", "This flag is deprecated and will be removed in a future release. Please use --leader-elect-resource-namespace instead.")
 	// volcano scheduler will ignore pods with scheduler names other than specified with the option
-	fs.StringArrayVar(&s.SchedulerNames, "scheduler-name", []string{agentSchedulerName}, "vc-agent-scheduler will handle pods whose .spec.SchedulerName is same as scheduler-name")
+	fs.StringVar(&s.SchedulerName, "scheduler-name", agentSchedulerName, "vc-agent-scheduler will handle pods whose .spec.SchedulerName is same as scheduler-name")
 	fs.StringVar(&s.SchedulerConf, "scheduler-conf", "", "The absolute path of scheduler configuration file")
 	fs.DurationVar(&s.ResyncPeriod, "resync-period", defaultResyncPeriod, "The default resync period for k8s native informer factory")
 	fs.BoolVar(&s.PrintVersion, "version", false, "Show version and quit")
@@ -157,7 +157,6 @@ func (s *ServerOption) RegisterOptions() {
 	ServerOpts = s
 	//some package from scheduler pkg rely on options defined in scheduler pkg
 	voptions.ServerOpts = voptions.NewServerOption()
-	voptions.ServerOpts.EnableCSIStorage = s.EnableCSIStorage
 }
 
 // readCAFiles read data from ca file path
--- a/cmd/agent-scheduler/app/server.go
+++ b/cmd/agent-scheduler/app/server.go
@@ -22,14 +22,6 @@ import (
 	"fmt"
 	"os"
 
-	"volcano.sh/apis/pkg/apis/helpers"
-	"volcano.sh/volcano/cmd/agent-scheduler/app/options"
-	scheduler "volcano.sh/volcano/pkg/agentscheduler"
-	"volcano.sh/volcano/pkg/agentscheduler/metrics"
-	"volcano.sh/volcano/pkg/kube"
-	"volcano.sh/volcano/pkg/signals"
-	commonutil "volcano.sh/volcano/pkg/util"
-
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	clientset "k8s.io/client-go/kubernetes"
@@ -47,6 +39,14 @@ import (
 
 	// Register rest client metrics
 	_ "k8s.io/component-base/metrics/prometheus/restclient"
+
+	"volcano.sh/apis/pkg/apis/helpers"
+	"volcano.sh/volcano/cmd/agent-scheduler/app/options"
+	scheduler "volcano.sh/volcano/pkg/agentscheduler"
+	"volcano.sh/volcano/pkg/agentscheduler/metrics"
+	"volcano.sh/volcano/pkg/kube"
+	"volcano.sh/volcano/pkg/signals"
+	commonutil "volcano.sh/volcano/pkg/util"
 )
 
 // Run the volcano scheduler.
@@ -94,7 +94,7 @@ func Run(opt *options.ServerOption) error {
 	// Prepare event clients.
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartRecordingToSink(&corev1.EventSinkImpl{Interface: leaderElectionClient.CoreV1().Events(opt.LeaderElection.ResourceNamespace)})
-	eventRecorder := broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: commonutil.GenerateComponentName(opt.SchedulerNames)})
+	eventRecorder := broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: commonutil.GenerateComponentName([]string{opt.SchedulerName})})
 
 	hostname, err := os.Hostname()
 	if err != nil {
--- a/cmd/agent-scheduler/main.go
+++ b/cmd/agent-scheduler/main.go
@@ -38,9 +38,6 @@ import (
 	// Import default actions/plugins.
 	_ "volcano.sh/volcano/pkg/agentscheduler/actions"
 	_ "volcano.sh/volcano/pkg/agentscheduler/plugins"
-
-	// init assert
-	_ "volcano.sh/volcano/pkg/scheduler/util/assert"
 )
 
 var logFlushFreq = pflag.Duration("log-flush-frequency", 5*time.Second, "Maximum number of seconds between log flushes")
@@ -57,7 +54,7 @@ func main() {
 	utilfeature.DefaultMutableFeatureGate.AddFlag(fs)
 
 	commonutil.LeaderElectionDefault(&s.LeaderElection)
-	s.LeaderElection.ResourceName = commonutil.GenerateComponentName(s.SchedulerNames)
+	s.LeaderElection.ResourceName = commonutil.GenerateComponentName([]string{s.SchedulerName})
 	componentbaseoptions.BindLeaderElectionFlags(&s.LeaderElection, fs)
 	s.RegisterOptions()
 
--- a/pkg/agentscheduler/actions/allocate/allocate.go
+++ b/pkg/agentscheduler/actions/allocate/allocate.go
@@ -107,7 +107,7 @@ func (alloc *Action) allocateTask(schedCtx *agentapi.SchedulingContext) error {
 		SchedCtx:       schedCtx,
 		BindContext:    alloc.CreateBindContext(schedCtx),
 	}
-	alloc.EnqueueSchedulerResultForTask(result)
+	alloc.SendResultToBinder(result)
 
 	return nil
 }
@@ -198,7 +198,7 @@ func (alloc *Action) CreateBindContext(schedCtx *agentapi.SchedulingContext) *ag
 	return bindContext
 }
 
-func (alloc *Action) EnqueueSchedulerResultForTask(result *agentapi.PodScheduleResult) {
+func (alloc *Action) SendResultToBinder(result *agentapi.PodScheduleResult) {
 	alloc.fwk.Cache.EnqueueScheduleResult(result)
 }
 
--- a/pkg/agentscheduler/cache/binder.go
+++ b/pkg/agentscheduler/cache/binder.go
@@ -59,14 +59,11 @@ func (binder *ConflictAwareBinder) Run(stopCh <-chan struct{}) {
 
 func (binder *ConflictAwareBinder) processScheduleResult() {
 	for {
-		select {
-		case scheduleResult, ok := <-binder.BindCheckChannel:
-			if !ok {
-				return
-			}
-			binder.CheckAndBindPod(scheduleResult)
-		default:
+		scheduleResult, ok := <-binder.BindCheckChannel
+		if !ok {
+			return
 		}
+		binder.CheckAndBindPod(scheduleResult)
 		if len(binder.BindCheckChannel) == 0 {
 			break
 		}
--- a/pkg/agentscheduler/cache/cache.go
+++ b/pkg/agentscheduler/cache/cache.go
@@ -20,13 +20,11 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"slices"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
 
-	"golang.org/x/time/rate"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -76,9 +74,11 @@ func init() {
 var _ Cache = &SchedulerCache{}
 
 // New returns a Cache implementation.
-func New(config *rest.Config, schedulerNames []string, nodeSelectors []string, nodeWorkers uint32, resyncPeriod time.Duration) Cache {
-	return newSchedulerCache(config, schedulerNames, nodeSelectors, nodeWorkers, resyncPeriod)
+func New(config *rest.Config, schedulerName string, nodeSelectors []string, nodeWorkers uint32, resyncPeriod time.Duration) Cache {
+	return newSchedulerCache(config, schedulerName, nodeSelectors, nodeWorkers, resyncPeriod)
 }
+
+//TODO: abstract common field and functions for volcano and agent scheduler, so each scheduler only depend on common part and customized part
 
 // SchedulerCache cache for the kube batch
 type SchedulerCache struct {
@@ -88,7 +88,7 @@ type SchedulerCache struct {
 	restConfig *rest.Config
 	vcClient   vcclient.Interface
 	// schedulerName is the name for volcano scheduler
-	schedulerNames     []string
+	schedulerName      string
 	nodeSelectorLabels map[string]sets.Empty
 	metricsConf        map[string]string
 
@@ -106,7 +106,6 @@ type SchedulerCache struct {
 
 	taskCache *TaskCache
 
-	errTasks  workqueue.TypedRateLimitingInterface[string]
 	nodeQueue workqueue.TypedRateLimitingInterface[string]
 
 	informerFactory   informers.SharedInformerFactory
@@ -283,7 +282,7 @@ func (sc *SchedulerCache) setBatchBindParallel() {
 	}
 }
 
-func newSchedulerCache(config *rest.Config, schedulerNames []string, nodeSelectors []string, nodeWorkers uint32, resyncPeriod time.Duration) *SchedulerCache {
+func newSchedulerCache(config *rest.Config, schedulerName string, nodeSelectors []string, nodeWorkers uint32, resyncPeriod time.Duration) *SchedulerCache {
 	kubeClient, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		panic(fmt.Sprintf("failed init kubeClient, with err: %v", err))
@@ -297,19 +296,13 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, nodeSelecto
 		panic(fmt.Sprintf("failed init eventClient, with err: %v", err))
 	}
 
-	errTaskRateLimiter := workqueue.NewTypedMaxOfRateLimiter[string](
-		workqueue.NewTypedItemExponentialFailureRateLimiter[string](5*time.Millisecond, 1000*time.Second),
-		&workqueue.TypedBucketRateLimiter[string]{Limiter: rate.NewLimiter(rate.Limit(100), 1000)},
-	)
-
 	sc := &SchedulerCache{
 		Nodes:              make(map[string]*schedulingapi.NodeInfo),
-		errTasks:           workqueue.NewTypedRateLimitingQueue[string](errTaskRateLimiter),
 		nodeQueue:          workqueue.NewTypedRateLimitingQueue[string](workqueue.DefaultTypedControllerRateLimiter[string]()),
 		kubeClient:         kubeClient,
 		vcClient:           vcClient,
 		restConfig:         config,
-		schedulerNames:     schedulerNames,
+		schedulerName:      schedulerName,
 		nodeSelectorLabels: make(map[string]sets.Empty),
 		imageStates:        make(map[string]*imageState),
 
@@ -326,13 +319,13 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, nodeSelecto
 	// Prepare event clients.
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartRecordingToSink(&corev1.EventSinkImpl{Interface: eventClient.CoreV1().Events("")})
-	sc.Recorder = broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: commonutil.GenerateComponentName(sc.schedulerNames)})
+	sc.Recorder = broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: commonutil.GenerateComponentName([]string{sc.schedulerName})})
 
 	// set concurrency configuration when binding
 	sc.setBatchBindParallel()
-	if bindMethodMap == nil {
+	if binder == nil {
 		klog.V(3).Info("no registered bind method, new a default one")
-		bindMethodMap = NewDefaultBinder(sc.kubeClient, sc.Recorder)
+		binder = NewDefaultBinder(sc.kubeClient, sc.Recorder)
 	}
 	sc.Binder = GetBindMethod()
 
@@ -352,9 +345,7 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, nodeSelecto
 	sc.cancel = cancel
 
 	queueingHintMapPerProfile := make(k8sschedulingqueue.QueueingHintMapPerProfile)
-	for _, schedulerName := range sc.schedulerNames {
-		queueingHintMapPerProfile[schedulerName] = buildQueueingHintMap()
-	}
+	queueingHintMapPerProfile[sc.schedulerName] = buildQueueingHintMap()
 	sc.schedulingQueue = k8sschedulingqueue.NewSchedulingQueue(
 		Less,
 		sc.informerFactory,
@@ -400,12 +391,6 @@ func (sc *SchedulerCache) addEventHandler() {
 	// `Namespace` informer is used by `InterPodAffinity` plugin,
 	// `SelectorSpread` and `PodTopologySpread` plugins uses the following four so far.
 	informerFactory.Core().V1().Namespaces().Informer()
-	informerFactory.Core().V1().Services().Informer()
-	if utilfeature.DefaultFeatureGate.Enabled(features.WorkLoadSupport) {
-		informerFactory.Core().V1().ReplicationControllers().Informer()
-		informerFactory.Apps().V1().ReplicaSets().Informer()
-		informerFactory.Apps().V1().StatefulSets().Informer()
-	}
 
 	// `PodDisruptionBudgets` informer is used by `Pdb` plugin
 	if utilfeature.DefaultFeatureGate.Enabled(features.PodDisruptionBudgetsSupport) {
@@ -487,7 +472,7 @@ func (sc *SchedulerCache) addEventHandler() {
 				switch v := obj.(type) {
 				case *v1.Pod:
 					// if the pod is not scheduled and scheduled by agent scheduler
-					if len(v.Spec.NodeName) == 0 && slices.Contains(sc.schedulerNames, v.Spec.SchedulerName) {
+					if len(v.Spec.NodeName) == 0 && sc.schedulerName == v.Spec.SchedulerName {
 						return true
 					}
 					return false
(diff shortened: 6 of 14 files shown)


cmd/agent-scheduler/app/server.go
```diff
@@ -0,0 +1,169 @@
...
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/client-go/tools/record"
+	basecompatibility "k8s.io/component-base/compatibility"
+
+	// Register rest client metrics
+	_ "k8s.io/component-base/metrics/prometheus/restclient"
```
hzxuzhonghu: nit: first go std lib and seconde group thirdparty pkg together, and at last group volcano pkg together

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -66,17 +66,27 @@ const (
 	PodTopologySpreadWeight = "podtopologyspread.weight"
 )
 
-type nodeOrderPlugin struct {
+type NodeOrderPlugin struct {
```
hzxuzhonghu, line 69: is this just a refactor or any other changes, can you point it

pkg/scheduler/plugins/predicates/predicates.go
hzxuzhonghu, line 1: this seems a refactor, if there are any other change, please point it to simplify review

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -242,12 +246,13 @@ func (pp *predicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 			klog.V(4).Infoln("predicates, deallocate", event.Task.NodeName)
 			pod := pl.UpdateTask(event.Task, "")
 			nodeName := event.Task.NodeName
-			node, found := nodeMap[nodeName]
-			if !found {
-				klog.Errorf("predicates, update pod %s/%s allocate from NOT EXIST node [%s]", pod.Namespace, pod.Name, nodeName)
+			node, err := pp.Handle.SnapshotSharedLister().NodeInfos().Get(nodeName)
```
hzxuzhonghu, line 249: I am concerned about this change, previously it get nodes from the session, now it get from the informer list. 

Will it bring some caveats?

pkg/agentscheduler/scheduler.go
```diff
@@ -0,0 +1,289 @@
...
+	if opt.SchedulerConf != "" {
+		var err error
+		path := filepath.Dir(opt.SchedulerConf)
+		watcher, err = filewatcher.NewFileWatcher(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed creating filewatcher for %s: %v", opt.SchedulerConf, err)
+		}
+	}
+
+	cache := schedcache.New(config, opt.SchedulerNames, opt.NodeSelector, opt.NodeWorkerThreads, opt.ResyncPeriod)
```
hzxuzhonghu: we donot need multiple SchedulerNames

pkg/agentscheduler/scheduler.go
```diff
@@ -0,0 +1,289 @@
...
+	for _, action := range sched.actions {
+		action.OnActionInit(sched.configurations)
+	}
+
+	klog.V(2).Infof("Scheduler completes Initialization and start to run %d workers", sched.workerCount)
+	for i := range sched.workerCount {
+		worker := &Worker{}
+		worker.framework = framework.NewFramework(sched.actions, sched.tiers, sched.cache, sched.configurations)
+		index := i
+		go wait.Until(func() { worker.runOnce(index) }, 0, stopCh)
```
hzxuzhonghu: When there are no any pods to schedule, wouldn't it run into deadloop?

pkg/agentscheduler/cache/binder.go
```diff
@@ -0,0 +1,151 @@
...
+	binder.schedulingQueue.Add(klog.Background(), podCopy) // Add back to the scheduling queue and make an urgent retry
+	klog.V(4).InfoS("Pod re-queued with priority", "pod", klog.KObj(task.Pod), "priority", priority, "reason", reason)
+}
+
+// CheckAndBindPod check the pod schedule result, send pod for binding if no conflict. Put pod back to schedule queue if there is conflict
+func (binder *ConflictAwareBinder) CheckAndBindPod(scheduleResult *agentapi.PodScheduleResult) {
+	task := scheduleResult.SchedCtx.Task
+	// 1. Check conflict
+	node := binder.FindNonConflictingNode(scheduleResult)
+	if node == nil {
```
hzxuzhonghu, line 100: how many pods are scheduled once, when could node == nil？

pkg/agentscheduler/cache/binder.go
```diff
@@ -0,0 +1,151 @@
...
+		select {
+		case scheduleResult, ok := <-binder.BindCheckChannel:
+			if !ok {
+				return
+			}
+			binder.CheckAndBindPod(scheduleResult)
+		default:
+		}
+		if len(binder.BindCheckChannel) == 0 {
+			break
```
hzxuzhonghu, line 68: I am also wondering the cpu usage of not any pod need to be binded

pkg/agentscheduler/cache/binder.go
```diff
@@ -0,0 +1,151 @@
...
+	task.NodeName = node.Name
+	task.Pod.Spec.NodeName = node.Name
+	nodeBindGeneration := node.BindGeneration
+	if err := binder.cache.AddBindTask(scheduleResult.BindContext); err != nil {
+		klog.ErrorS(err, "Failed to add bind task", "pod", klog.KObj(task.Pod), "node", node.Name)
+		binder.requeuePodWithPriority(task, SchedulingPriorityUrgent, RequeueReasonBindTaskFailure)
+		return
+	}
+
+	binder.schedulingQueue.Done(task.Pod.UID) // Mark previous processing as done to clean in-flight records, avoid memory leak
```
hzxuzhonghu, line 117: Done should be called in pair with Pop, otherwise it is very easy to miss calling `Done`

pkg/agentscheduler/cache/cache.go
hzxuzhonghu, line 1: what is the different with pkg/scheduler/cache/cache.go. ducoment it and add a TODO: can we merge them or anstract the common parts

pkg/scheduler/plugins/nodeorder/nodeorder.go
qi-min: For changes under scheduler/plugins are just refactor,  like abstract functions, so that it can be used in both batch scheduler and agent scheduler.
Modification on line 69  is to allow this plugin resued in agent-scheduler package

pkg/scheduler/plugins/predicates/predicates.go
qi-min: change in predicate is redactor, to allow the functions can be used in both agent-scheduler and batch scheduler

pkg/agentscheduler/scheduler.go
qi-min: When no pod to schedule, runOnce will be blocked at Popping pod from queue until new Pending pod enqueue.

pkg/agentscheduler/cache/binder.go
qi-min: One pod is scheduled in each worker at once.  The scheduleResult(candicate nodes, default is 3) are sent to binder from worker. FindNonConflictingNode check whether theses nodes confilct with previous binding.   If the version of node is not used in previsous binding, node will be used for binding. Otherwise node will be rejected. Nil is returned if all candidates nodes are rejected.

E.g. candidate nodes are n1(v1), n2(v1), n3(v2).  If nodes with these versions are used in previous binding, then all candidates will be dropped and nil returned. If previsous binding on n3 is at version v1, then n3 will be used.

cmd/agent-scheduler/main.go
```diff
@@ -0,0 +1,88 @@
...
+	"volcano.sh/volcano/cmd/agent-scheduler/app/options"
+	commonutil "volcano.sh/volcano/pkg/util"
+	"volcano.sh/volcano/pkg/version"
+
+	// Import default actions/plugins.
+	_ "volcano.sh/volcano/pkg/agentscheduler/actions"
+	_ "volcano.sh/volcano/pkg/agentscheduler/plugins"
+
+	// init assert
+	_ "volcano.sh/volcano/pkg/scheduler/util/assert"
```
hzxuzhonghu: remove, this should only occur in test

cmd/agent-scheduler/app/options/options.go
```diff
@@ -0,0 +1,205 @@
...
+	defaultMinPercentageOfNodesToFind = 5
+	defaultMinNodesToFind             = 100
+	defaultPercentageOfNodesToFind    = 0
+	defaultLockObjectNamespace        = "volcano-system"
+	defaultNodeWorkers                = 20
+	defaultScheduleWorkerCount        = 1
+)
+
+// ServerOption is the main context object for the controller manager.
+type ServerOption struct {
```
hzxuzhonghu, line 56: This is mostly copied from scheduler and `RegisterOptions` has some coupling with scheduler. set `ServerOpts.EnableCSIStorage` 

Could we eliminate that

cmd/agent-scheduler/main.go
```diff
@@ -0,0 +1,88 @@
...
+
+	fs := pflag.CommandLine
+	s := options.NewServerOption()
+
+	s.AddFlags(fs)
+	utilfeature.DefaultMutableFeatureGate.AddFlag(fs)
+
+	commonutil.LeaderElectionDefault(&s.LeaderElection)
+	s.LeaderElection.ResourceName = commonutil.GenerateComponentName(s.SchedulerNames)
+	componentbaseoptions.BindLeaderElectionFlags(&s.LeaderElection, fs)
```
hzxuzhonghu, line 58: I would prefer moving the leader elector related to app.Run

cmd/agent-scheduler/app/server.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Run the volcano scheduler.
+func Run(opt *options.ServerOption) error {
+	config, err := kube.BuildConfig(opt.KubeClientOptions)
+	if err != nil {
+		return err
+	}
+
+	// Align default feature-gates with the connected cluster's version.
+	if err := setupComponentGlobals(config); err != nil {
```
hzxuzhonghu, line 60: what does this do

cmd/agent-scheduler/app/server.go
```diff
@@ -0,0 +1,169 @@
...
+		klog.Errorf("failed to set component globals: %v", err)
+		return err
+	}
+
+	metrics.InitKubeSchedulerRelatedMetrics()
```
hzxuzhonghu, line 65: backoffqueue depend on the metrics, but i donot think we ever update them.

pkg/agentscheduler/cache/cache.go
```diff
@@ -0,0 +1,1029 @@
...
+		v1.AddToScheme,
+	}
+
+	utilruntime.Must(schemeBuilder.AddToScheme(scheme.Scheme))
+}
+
+var _ Cache = &SchedulerCache{}
+
+// New returns a Cache implementation.
+func New(config *rest.Config, schedulerNames []string, nodeSelectors []string, nodeWorkers uint32, resyncPeriod time.Duration) Cache {
```
hzxuzhonghu: So we are using `nodeSelectors ` just because node sharding is not supported. 
But to not break backward compatability, we should remove `nodeSelectors` flag

pkg/agentscheduler/cache/cache.go
```diff
@@ -0,0 +1,1029 @@
...
+	StatusUpdater StatusUpdater
+
+	Recorder record.EventRecorder
+
+	Nodes    map[string]*schedulingapi.NodeInfo // TODO: do we need to also add a seperate lock for Nodes cache?
+	NodeList []string
+
+	taskCache *TaskCache
+
+	errTasks  workqueue.TypedRateLimitingInterface[string]
```
hzxuzhonghu: unused, please remove

pkg/agentscheduler/cache/cache.go
```diff
@@ -0,0 +1,1029 @@
...
+	podInformer  infov1.PodInformer
+	nodeInformer infov1.NodeInformer
+
+	Binder        Binder
+	StatusUpdater StatusUpdater
+
+	Recorder record.EventRecorder
+
+	Nodes    map[string]*schedulingapi.NodeInfo // TODO: do we need to also add a seperate lock for Nodes cache?
+	NodeList []string
```
hzxuzhonghu, line 105: why do we need both a map and a slice

pkg/agentscheduler/cache/cache.go
```diff
@@ -0,0 +1,1029 @@
...
+	// explicitly register informers to the factory, otherwise resources listers cannot get anything
+	// even with no error returned.
+	// `Namespace` informer is used by `InterPodAffinity` plugin,
+	// `SelectorSpread` and `PodTopologySpread` plugins uses the following four so far.
+	informerFactory.Core().V1().Namespaces().Informer()
+	informerFactory.Core().V1().Services().Informer()
+	if utilfeature.DefaultFeatureGate.Enabled(features.WorkLoadSupport) {
+		informerFactory.Core().V1().ReplicationControllers().Informer()
+		informerFactory.Apps().V1().ReplicaSets().Informer()
+		informerFactory.Apps().V1().StatefulSets().Informer()
```
hzxuzhonghu: please remove, cleanup all unused logic

pkg/agentscheduler/cache/binder.go
```diff
@@ -0,0 +1,151 @@
...
+func NewConflictAwareBinder(schedulerCache Cache, schedulingQueue k8sschedulingqueue.SchedulingQueue) *ConflictAwareBinder {
+	return &ConflictAwareBinder{
+		nodeBindRecords:  make(map[string]int64, 0),
+		BindCheckChannel: make(chan *agentapi.PodScheduleResult, 5000),
+		schedulingQueue:  schedulingQueue,
+		cache:            schedulerCache,
+	}
+}
+func (binder *ConflictAwareBinder) Run(stopCh <-chan struct{}) {
+	go wait.Until(binder.processScheduleResult, 0, stopCh)
```
hzxuzhonghu, line 57: processScheduleResult is nonblocking, if no scheduleResult, it will fall into deadloop. Occupying cpu

pkg/agentscheduler/cache/cache.go
```diff
@@ -0,0 +1,1029 @@
...
+		return fmt.Errorf("failed to bind Task %v to host %v, host does not exist",
+			task.UID, task.NodeName)
+	}
+
+	originalStatus := task.Status
+	if err := sc.UpdateTaskStatus(task, schedulingapi.Binding); err != nil {
+		return err
+	}
+
+	err := task.SetPodResourceDecision()
```
hzxuzhonghu: unused

pkg/agentscheduler/cache/cache.go
```diff
@@ -0,0 +1,1029 @@
...
+func (sc *SchedulerCache) processBindTask() {
+	for {
+		select {
+		case bindContext, ok := <-sc.BindFlowChannel:
+			if !ok {
+				return
+			}
+
+			sc.bindCache = append(sc.bindCache, bindContext)
+			if len(sc.bindCache) == sc.batchNum {
```
hzxuzhonghu, line 800: should add a max waiting time, otherwise the latency could be undefined

hzxuzhonghu: I am also wondering could batch binding save time?

pkg/agentscheduler/cache/event_handlers.go
```diff
@@ -0,0 +1,476 @@
...
+	_, pod, err := schedutil.As[*v1.Pod](nil, obj)
+	if err != nil {
+		klog.Errorf("Failed to convert objects to *v1.Pod: %v", err)
+		return
+	}
+
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
+
+	err = sc.addPod(pod)
```
hzxuzhonghu, line 157: Make the critical section small, i guess it is to protect `addPod` only

pkg/agentscheduler/cache/factory.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"maps"
+	"sync"
+
+	"k8s.io/klog/v2"
+)
+
+// bindMethodMap Binder management
+var bindMethodMap Binder
```
hzxuzhonghu: why call map?

pkg/agentscheduler/cache/factory.go
```diff
@@ -0,0 +1,72 @@
...
+func NewBinderRegistry() *BinderRegistry {
+	return &BinderRegistry{
+		preBinders: make(map[string]PreBinder),
+	}
+}
+
+// Register registers or updates a binder for the given plugin name . The plugin can be such as preBinder or postBinder.
+// It always overwrites the existing binder map to support plugin configuration updates
+// during runtime, as plugins may be reconfigured without restarting the scheduler.
+func (r *BinderRegistry) Register(name string, binder interface{}) {
```
hzxuzhonghu, line 55: Note: this is not called, so  volume binding check is not supported

pkg/agentscheduler/framework/interface.go
```diff
@@ -0,0 +1,62 @@
...
+// Action is the interface of agent scheduler action.
+type Action interface {
+	// Name returns the unique name of Action.
+	Name() string
+
+	// OnActionInit initializes the plugin. It is called once when the framework is created.
+	OnActionInit(configurations []conf.Configuration)
+
+	// Initialize initializes the allocator plugins.
+	Initialize()
```
hzxuzhonghu, line 35: can merge

pkg/agentscheduler/scheduler.go
```diff
@@ -0,0 +1,289 @@
...
+	klog.V(4).Infof("Start scheduling in worker %d ...", index)
+	scheduleStartTime := time.Now()
+	defer klog.V(4).Infof("End scheduling in worker %d ...", index)
+	// Load ConfigMap to check which action is enabled.
+	conf.EnabledActionMap = make(map[string]bool)
+	for _, action := range worker.framework.Actions {
+		conf.EnabledActionMap[action.Name()] = true
+	}
+
+	schedCtx, err := worker.generateNextSchedulingContext()
```
hzxuzhonghu, line 142: You can add a for loop here instead of depending on ` wait.Until(func() { worker.runOnce(index) }, 0, stopCh)` to do that in function granularity, which could be less efficient and produce confusing logs

pkg/agentscheduler/scheduler.go
hzxuzhonghu: You mean the queue is blocking read, that's ok

pkg/agentscheduler/actions/allocate/allocate.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+		return err
+	}
+	bestNodes := alloc.prioritizeNodes(task, predicatedNodes)
+	result := &agentapi.PodScheduleResult{
+		SuggestedNodes: bestNodes,
+		SchedCtx:       schedCtx,
+		BindContext:    alloc.CreateBindContext(schedCtx),
+	}
+	alloc.EnqueueSchedulerResultForTask(result)
```
hzxuzhonghu: seems we need a more clear name, like sendToBinder

pkg/scheduler/plugins/predicates/predicates.go
JesseStutler: https://github.com/qi-min/volcano/blob/5e134b9d039065acea682f15132a49f3f05d16fe/pkg/scheduler/plugins/predicates/predicates.go#L190-L191 Actually we still use the nodemap in session here, there's a judgment here in NewFramework: 
https://github.com/qi-min/volcano/blob/5e134b9d039065acea682f15132a49f3f05d16fe/pkg/scheduler/plugins/util/k8s/framework.go#L223-L227
so the logic of getting the nodemap from the session hasn't changed

pkg/agentscheduler/cache/binder.go
JesseStutler: Although retries based on events are not yet implemented, if we call `Done` immediately after Pop,if scheduling fails, the pod will no longer be in the inflight list: https://github.com/volcano-sh/volcano/blob/76c3253a19bffc97b9619b99330dd63115ed05b5/third_party/kubernetes/pkg/scheduler/backend/queue/active_queue.go#L392-L406
Event changes that occur during the scheduling of this pod will prevent the pod from retrying quickly. 
Currently, after Pop, Done is handled without omission. When scheduling fails, Done is called in `AddUnschedulableIfNotPresent`, and Done is also called when `CheckAndBind` is reached after successful scheduling, so we didn't miss it

pkg/agentscheduler/actions/allocate/allocate.go
```diff
@@ -0,0 +1,265 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/klog/v2"
+
+	agentapi "volcano.sh/volcano/pkg/agentscheduler/api"
+	"volcano.sh/volcano/pkg/agentscheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/api"
+	vcache "volcano.sh/volcano/pkg/scheduler/cache"
```
wangyang0616, line 27: The agentscheduler replicates the implementations of cache and framework, but it also references the cache and framework under the scheduler directory here. What are the principles for distinguishing between the cache and framework components in these two directories?

pkg/agentscheduler/framework/framework.go
```diff
@@ -0,0 +1,155 @@
...
+	Actions        []Action
+	Tiers          []conf.Tier
+	Configurations []conf.Configuration
+
+	// Function registries
+	PredicateFns      map[string]api.PredicateFn
+	PrePredicateFns   map[string]api.PrePredicateFn
+	NodeOrderFns      map[string]api.NodeOrderFn
+	BatchNodeOrderFns map[string]api.BatchNodeOrderFn
+	NodeMapFns        map[string]api.NodeMapFn
```
wangyang0616, line 47: Do the two callback functions `NodeMapFns` and `NodeReduceFns` have specific scenario requirements in the agent-scheduler context, or are they merely intended to align with the existing scheduling logic?

pkg/agentscheduler/framework/framework.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"volcano.sh/volcano/pkg/agentscheduler/cache"
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/conf"
+	k8sutil "volcano.sh/volcano/pkg/scheduler/plugins/util/k8s"
+)
+
+// Framework manages the scheduler plugins and their execution points.
+type Framework struct {
+	*k8sutil.Framework // Embedding Framework to implement k8sframework.Handle interface
```
wangyang0616, line 35: The k8sutil package belongs to the common directory under scheduler/plugin, and it is currently intended to be referenced by multiple algorithm plugins. Is it appropriate for the current agentscheduler to reference it directly?

pkg/agentscheduler/scheduler.go
```diff
@@ -0,0 +1,289 @@
...
+		klog.Errorf("Scheduler config %s is invalid: %v", config, err)
+		return
+	}
+
+	sched.mutex.Lock()
+	sched.actions = actions
+	sched.tiers = tiers
+	sched.configurations = configurations
+	sched.metricsConf = metricsConf
+	sched.mutex.Unlock()
```
wangyang0616: Would it be better to use `defer sched.mutex.Unlock() `here, to avoid the risk of deadlock caused by the continuous addition of code in the future?

pkg/agentscheduler/cache/cache.go
qi-min: In agent scheduler, we used informers, nodeinfo, nodeList, Binder from scheduler cache and also add/modify several fields and function like confictAwareBinder, scheduling queue, etc. I will add a TODO to abstract common parts from agent/batch scheduler

cmd/agent-scheduler/app/server.go
qi-min: yes, the scheduling queue we imported in third-party path depend on the metrics, nil error is reported if metrics are not inited. I think we need a TODO decouple the kube metrics and record metrics.

qi-min: just found batch scheduler also call this `InitKubeSchedulerRelatedMetrics `to init metrics few weeks ago for https://github.com/volcano-sh/volcano/pull/4731

pkg/agentscheduler/cache/cache.go
qi-min: nodeSelectors is inherited from batch scheduler to restrict nodes. I am wondering can we keep it to keep consistent with batch scheduler? When sharding is not enabled, user can still use this flag to restrict the nodes can be used.

pkg/agentscheduler/cache/cache.go
qi-min: done

pkg/agentscheduler/cache/cache.go
qi-min: done

pkg/agentscheduler/cache/cache.go
qi-min: seems the loop will break if no items found in channel and bindTask is called immediately, it will not wait for new items to meet batch count.
This is same as logic in scheduler, this function may be abstract as common func for both batch and agent scheduler

pkg/agentscheduler/cache/event_handlers.go
qi-min: right, only addPod need be protected, I will change it

pkg/agentscheduler/cache/factory.go
qi-min: not sure, seems it was just copied from batch scheduler factor. It is renamed to `binder` now

pkg/agentscheduler/cache/factory.go
qi-min: it is called in `OnPluginInit` in predicate plugin

pkg/agentscheduler/framework/interface.go
qi-min: @hzxuzhonghu what need be merged?

pkg/agentscheduler/scheduler.go
qi-min: good suggestion, I will change it

pkg/agentscheduler/actions/allocate/allocate.go
qi-min: done

pkg/agentscheduler/actions/allocate/allocate.go
qi-min: Some structure definition is reused from batch scheduler, that's why scheduler/cache is also imported. We need to abstract the common part of cache for batch/agent scheduler, so only common cache will be referenced in future. I will add a TODO

pkg/agentscheduler/framework/framework.go
qi-min: Just to align with current scheduling logic so that  the existing plugins can be reused

pkg/agentscheduler/framework/framework.go
qi-min: *k8sutil.Framework defined some new fields and functions that are also need in agent scheduler, that's why it is referenced
`	snapshot         *Snapshot`

pkg/agentscheduler/scheduler.go
qi-min: OK

cmd/agent-scheduler/app/server.go
qi-min: sure, I changed the order

pkg/agentscheduler/scheduler.go
qi-min: ok, changed it to use single name

cmd/agent-scheduler/main.go
qi-min: done

cmd/agent-scheduler/app/server.go
qi-min: seems it is used to discover kube server version and set feature gate defaults to match the cluster

pkg/agentscheduler/cache/binder.go
qi-min: I updated the flow, `processScheduleResult` will be blocked if no result in channel

pkg/agentscheduler/cache/binder.go
qi-min: the function is updated, it will be blocked if no result sent to channel

cmd/agent-scheduler/app/options/options.go
qi-min: the plugins in batch scheduler depend on options.ServerOpts, but ServerOpts is a global variable at cmd/scheduler /app/options level. So it is not inited with flag values when agent scheduler startup. `RegisterOptions` pass options to `options.ServerOpts` so that it can pick up value from flags of agent scheduler.
In next pr for sharing, the logic is changed to pass all options of agented scheduler to batch scheduler, no need to specially specify `EnableCSIStorage`  
`voptions.ServerOpts = ServerOpts.ServerOption`

pkg/agentscheduler/cache/cache.go
qi-min: I think both map and slice follow the fields from cache in batch scheduler, seems the slice is not necessary.
@handan-yxh could you help check is it necessary to keep NodeList separately?

pkg/agentscheduler/cache/cache.go
qi-min: done

qi-min:
@hzxuzhonghu @wangyang0616 review comments have been handled, would you please review again, thanks.

pkg/agentscheduler/framework/interface.go
hzxuzhonghu: I mean OnActionInit  and OnActionInit can be merged

pkg/agentscheduler/framework/framework.go
wangyang0616: This is not a good way to depend on each other. Normally, plugins should be able to reference the cache and framework packages, but the framework should not depend on the plugin packages in return. It is recommended to register a separate issue for rectification and optimization.

pkg/agentscheduler/actions/allocate/allocate.go
wangyang0616: It is recommended to create an independent issue tracking system for optimization.

pkg/agentscheduler/framework/framework.go
qi-min: https://github.com/volcano-sh/volcano/issues/4840

pkg/agentscheduler/actions/allocate/allocate.go
qi-min: https://github.com/volcano-sh/volcano/issues/4839

wangyang0616:
Please confirm the closure status of the review comments. @hzxuzhonghu

PR #4810, opened 2025-12-12
Support task level network topology constrain with minSubGroups

3sunny:
#### What type of PR is this?
/kind feature

Volcano supports setting up two-level Gang scheduling capabilities through PodGroup, while Kthena integrates with PodGroup to implement role-level elastic scheduling capabilities.

Partial fixes for #4188 

Example YAML for vcJob after modification
```yaml
kind: Job
metadata:
  name: network-topology-job
spec:
  minAvailable: 3
  schedulerName: volcano
  networkTopology:
    mode: hard
    highestTierAllowed: 2
  tasks:
    - replicas: 6
      name: "task"
      partitionPolicy:
        totalPartitions: 2
        partitionSize: 3
        minPartitions: 1 //
        networkTopology:
            mode: hard
            highestTierAllowed: 1
      template:
        metadata:
          name: task
        spec:
          containers:
            - image: ubuntu
              imagePullPolicy: IfNotPresent
              name: task
              resources:
                requests:
                  cpu: "2"
                  memory: 2Gi
          restartPolicy: OnFailure
```
Example YAML for podGroup after modification, Minimum requirement: 2rolea and 1roleb
```yaml
kind: PodGroup
metadata:
  name: network-topology-podgroup
spec:
  minMember: 10
  networkTopology:
    mode: hard 
    highestTierAllowed: 2
  subGroupPolicy: 
    - subGroupSize: 4
      minSubGroups: 2
      name: rolea
      labelSelector:
        matchLabels:
          volcano.sh/task-spec: rolea
      matchLabelKeys:
        - volcano.sh/partiton-id
      networkTopology:
        mode: hard 
        highestTierAllowed: 1
    - subGroupSize: 2
      minSubGroups: 1
      name: roleb
      labelSelector:
        matchLabels:
          volcano.sh/task-spec: roleb
      matchLabelKeys:
        - volcano.sh/partition-id
      networkTopology:
        mode: hard 
        highestTierAllowed: 1
```

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -769,11 +769,7 @@ func (cc *jobcontroller) createOrUpdatePodGroup(job *batch.Job) error {
 		}
 		minTaskMember := map[string]int32{}
 		for _, task := range job.Spec.Tasks {
-			if task.MinAvailable != nil {
-				minTaskMember[task.Name] = *task.MinAvailable
-			} else {
-				minTaskMember[task.Name] = task.Replicas
-			}
+			minTaskMember[task.Name] = cc.getMinTaskMember(task)
 		}
 
 		pg := &scheduling.PodGroup{
@@ -835,6 +831,16 @@ func (cc *jobcontroller) createOrUpdatePodGroup(job *batch.Job) error {
 	return err
 }
 
+func (cc *jobcontroller) getMinTaskMember(task batch.TaskSpec) int32 {
+	if task.MinAvailable != nil {
+		return *task.MinAvailable
+	}
+	if task.PartitionPolicy != nil && task.PartitionPolicy.MinPartitions != 0 && task.PartitionPolicy.PartitionSize != 0 {
+		return task.PartitionPolicy.MinPartitions * task.PartitionPolicy.PartitionSize
+	}
+	return task.Replicas
+}
+
 func (cc *jobcontroller) shouldUpdateExistingPodGroup(pg *scheduling.PodGroup, job *batch.Job) bool {
 	pgShouldUpdate := false
 	if pg.Spec.PriorityClassName != job.Spec.PriorityClassName {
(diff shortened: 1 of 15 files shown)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -833,6 +829,16 @@ func (cc *jobcontroller) createOrUpdatePodGroup(job *batch.Job) error {
 	return err
 }
 
+func (cc *jobcontroller) getMinTaskMember(task batch.TaskSpec) int32 {
+	if task.MinAvailable != nil {
+		return *task.MinAvailable
```
wangyang0616, line 836: If task.MinAvailable is set, MinPartitions will not take effect. Moreover, in `mutate_job.go`, `task.MinAvailable` is assigned a default value—so in the current processing logic, MinPartitions can barely take effect.

3sunny: > If task.MinAvailable is set, MinPartitions will not take effect. Moreover, in `mutate_job.go`, `task.MinAvailable` is assigned a default value—so in the current processing logic, MinPartitions can barely take effect.

In mutate_job.go, the logic for assigning default values to task.MinAvailable has been modified to include a condition: the default value is only assigned when MinPartitions is not set. When both task.MinAvailable and MinPartitions are set, an error message will be triggered indicating that these two cannot be configured simultaneously.

wangyang0616: Got it

pkg/scheduler/api/job_info.go
```diff
@@ -1092,6 +1111,57 @@ func (ji *JobInfo) ValidTaskNum() int32 {
...
+		}
+	}
+	return true
+}
+
+func (ji *JobInfo) CheckSubJobReady() bool {
+	return ji.checkSubJobCondition(func(subJob *SubJobInfo) bool {
+		return subJob.IsReady()
+	}, IsReady)
+}
```
ouyangshengjia, line 1155:
```suggestion
func (ji *JobInfo) CheckSubJobReady() bool {
    err := ji.checkSubJobCondition(func(subJob *SubJobInfo) bool {
		return subJob.IsReady()
	},)
	if err != nil {
		klog.V(4).InfoS("SubJob not ready", "reason", err.Error())
        return false
	}
	return true
}
```
You can return the reason why `checkSubJobCondition` is not satisfied, and then print the log at the upper level. This way, you won't need the input parameter `conditionName`.

ouyangshengjia:
Please correct your example in the PR description. A vcjob with **one** task will generate only one subGroupPolicy.

pkg/scheduler/api/job_info.go
3sunny: done

PR #4822, opened 2025-12-16
Fixed memory leak caused by metrics remaining due to race conditions

### LINKED ISSUE #4821: Memory leak caused by leftover metrics data
### Description

Due to race problem, Clearing up job metrics and writing job metrics are executed concurrently. This could lead to metrics being cleared before they are written, resulting in leftover metrics, leading not only the memory leaks but also to increasingly slow response times for the `/metrics` interface.

This can occur when a large number of jobs that finish quickly are created.

In our production environment, a large number of pods are continuously created, and some of these pods terminate immediately after being scheduled.

The `unschedule_task_count`, `job_share`, and `job_retry_counts` metric for a job are written after the pod has been scheduled. Therefore, it's possible that a job has already been cleaned up by the `processCleanupJob` method, but its metrics are written afterwards, causing these metrics to remain permanently in memory.

After running continuously in our environment for a month, the amount of residual data in the metrics can reach nearly 100MB.This resulted in Prometheus scraping the scheduler's metrics data very slowly.
```
# curl -o /dev/null -s -w "Total Time: %{time_total}s\n" http://127.0.0.1:8080/metrics
Total Time: 10.126988s
# curl -s http://127.0.0.1:8080/metrics | grep job_share |  wc -l
207512
# kubectl get pg -A | wc -l
2280
```

There shouldn't be any leftover for these three job-related metrics `unschedule_task_count`, `job_share`, and `job_retry_counts` after the job has been deleted.

v1.11.2

_No response_

halcyon-r:
Verify that the job still exists before set job's metrics.

#### What type of PR is this?
/kind bug

Before writing the `unschedule_task_count`, `job_share`, and `job_retry_counts` metrics, check if the job exists to prevent leftover metrics.

Fixes #4821 

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1100,6 +1100,16 @@ func (sc *SchedulerCache) processCleanupJob() {
 	}
 }
 
+func (sc *SchedulerCache) IsJobTerminated(jobId schedulingapi.JobID) bool {
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
+	job, exists := sc.Jobs[jobId]
+	if !exists || job == nil {
+		return true
+	}
+	return schedulingapi.JobTerminated(job)
+}
+
 func (sc *SchedulerCache) generateDeletedJobsKey(job *schedulingapi.JobInfo) string {
 	// Job UID is namespace + / +name, for example: theNs/theJob
 	// Job PgUID is derived from the Job PgUID, for example: d336abea-4f14-42c7-8a6b-092959a31407
--- a/pkg/scheduler/cache/interface.go
+++ b/pkg/scheduler/cache/interface.go
@@ -92,6 +92,9 @@ type Cache interface {
 
 	// SharedDRAManager returns the shared DRAManager
 	SharedDRAManager() framework.SharedDRAManager
+
+	// IsJobTerminated returns if the job was terminated
+	IsJobTerminated(jobId api.JobID) bool
 }
 
 // Binder interface for binding task and hostname
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -1058,3 +1058,7 @@ func (ssn *Session) String() string {
 
 	return msg
 }
+
+func (ssn *Session) IsJobTerminated(jobId api.JobID) bool {
+	return ssn.cache.IsJobTerminated(jobId)
+}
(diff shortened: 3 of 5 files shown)


pkg/scheduler/cache/interface.go
```diff
@@ -92,6 +92,9 @@ type Cache interface {
 
 	// SharedDRAManager returns the shared DRAManager
 	SharedDRAManager() framework.SharedDRAManager
+
+	// IsJobTerminated returns if the job was terminated
+	IsJobTerminated(jobId string) bool
```
hajnalmt: Why haven't you used `(jobId api.jobId)` here too?
You would have spared the casting in `cache.go`

hajnalmt (commented):
Looks good to me, I had one question only.

Checked the occurancies too and the commit is fine (we don't update these metrics elsewehere which is surprising).

pkg/scheduler/cache/cache.go
```diff
@@ -1100,6 +1100,16 @@ func (sc *SchedulerCache) processCleanupJob() {
 	}
 }
 
+func (sc *SchedulerCache) IsJobTerminated(jobId string) bool {
+	sc.Mutex.Lock()
```
hajnalmt, line 1104: I have created an enhacement request for this as this should be a Read Lock on the long run.
See: https://github.com/volcano-sh/volcano/issues/4824

pkg/scheduler/cache/interface.go
halcyon-r: Yes, you are right. I‘ll optimize this part.

JesseStutler:
@halcyon-r So it's the same issue here but with different solutions? https://github.com/volcano-sh/volcano/pull/4760 I prefer to check if the job still exist rather than use ttl way, but also let @fengruotj takes a look

pkg/scheduler/cache/interface.go
```diff
@@ -92,6 +92,9 @@ type Cache interface {
 
 	// SharedDRAManager returns the shared DRAManager
 	SharedDRAManager() framework.SharedDRAManager
+
+	// IsJobTerminated returns if the job was terminated
+	IsJobTerminated(jobId api.JobID) bool
```
JesseStutler, line 97: I feel that we are too arbitrary to add a method to the cache interface here, the cache is not like the design that an interface should have

PR #4829, opened 2025-12-18
job controller should keep terminating pod in job

### LINKED ISSUE #4828: Job aborting state transitions to Aborted too early because terminating Pods are removed from job cache
### Description

When executing an abort action on a Volcano Job that still has running Pods, the Job state changes to Aborting and Pods start to be deleted. As soon as the Pods enter Terminating in Kubernetes and are removed from the Volcano job cache, the Job state is immediately changed to Aborted, even though there are still terminating Pods in the cluster.
So the Job reaches Aborted while some of its Pods are still present and terminating on the cluster side.

Image

Image

1. Create a Volcano Job with at least one long‑running Pod.
2. Wait until the Job Pods are in Running phase.
3. Trigger an abort on the Job (for example via vcctl or another API that sends the abort action).
4. Observe:
The Job state goes to Aborting.
The Job quickly transitions from Aborting to Aborted.
At the same time, if you check the Pods (e.g. kubectl get pod -w), you can still see Pods in Terminating phase for some time after the Job is already Aborted.

So there is a visible window where the Job is already Aborted, but some of its Pods are still terminating in the cluster.

I expect the Job to remain in the Aborting phase until all Pods are actually gone (i.e. no “alive” Pods left: no Pending/Running/Terminating Pods, or the Pods are no longer present). Only then should the Job transition to Aborted.

1.9+

### Analysis / Suspected cause

This seems to be caused by the interaction of the following pieces:

- **Pod update handler** (`pkg/controllers/job/job_controller_handler.go`, `updatePod`):

  - When a Pod has a `DeletionTimestamp`, `updatePod` does:

    ```go
    if newPod.DeletionTimestamp != nil {
        cc.deletePod(newObj)
        return
    }
    ```

  - This means terminating Pods are removed from the Volcano job cache and will no longer contribute to the cached Job status (e.g. `status.Terminating`).

- **Job state machine dispatch** (`pkg/controllers/job/job_controller.go`, around where `state.NewState(jobInfo)` and `st.Execute(action)` are called):

  - When an abort action is processed, the Job enters the `Aborting` state and `st.Execute(action)` will eventually call the `abortingState.Execute`.

- **`aborting` state implementation** (`pkg/controllers/job/state/aborting.go`, `Execute`):

  - For the default path (abort in Aborting state), it calls:

    ```go
    return KillJob(ps.job, PodRetainPhaseSoft, func(status *vcbatch.JobStatus) bool {
        // If any "alive" pods, still in Aborting phase
        if status.Terminating != 0 || status.Pending != 0 || status.Running != 0 {
            return false
        }
        status.State.Phase = vcbatch.Aborted
        return true
    })
    ```

  - This callback decides whether to stay in `Aborting` or move to `Aborted` based on `status.Terminating`, `status.Pending`, and `status.Running` **as maintained in the Volcano job cache**.

Because terminating Pods are removed from the cache in `updatePod` as soon as `DeletionTimestamp` is set, the cached `status.Terminating` becomes `0` earlier than the Pods are actually gone from the cluster. As a result, the above condition sees no “alive” Pods and sets `status.State.Phase = vcbatch.Aborted` too early, while Kubernetes still shows Pods in `Terminating`.

So the Job lifecycle in Volcano and the actual Pod lifecycle in the cluster are temporarily inconsistent.

---

### Possible directions for a fix (for discussion)

Some possible approaches (for maintainers to consider):

- **Treat terminating Pods as still “alive” in the cache** until they fully disappear, instead of removing them as soon as `DeletionTimestamp` is set; or
- Delay the transition to `Aborted` until there is a stronger guarantee that all Pods have been fully deleted (e.g. by confirming no Pods exist for that Job).

kingeasternsun:
#### What type of PR is this?

/kind bug

Currently, when a Job is aborted and enters the `Aborting` phase, its Pods start to be deleted.  
On Pod update, as soon as a Pod has a non-nil `DeletionTimestamp` (i.e. it becomes `Terminating`), the controller removes that Pod from the Job cache.  

In the `Aborting` state, the transition logic checks `status.Terminating`, `status.Pending`, and `status.Running` from the cached Job status:

```go
// If any "alive" pods, still in Aborting phase
if status.Terminating != 0 || status.Pending != 0 || status.Running != 0 {
    return false
}
status.State.Phase = vcbatch.Aborted
```

Because terminating Pods are removed from the cache too early, `status.Terminating` becomes `0` while the Pods are still terminating in the cluster, and the Job transitions to `Aborted` prematurely.

This PR changes the behavior so that **Pods in `Terminating` state are kept in the job cache and are not removed on the update event**. Instead, they stay in the cache (and in the Job status) until the actual Pod delete event is received.  
With this change:

- The Job remains in `Aborting` while there are still terminating Pods.
- The Job only transitions to `Aborted` after all “alive” Pods (including terminating ones) are gone, making the Job state consistent with the real Pod lifecycle in the cluster.

Fixes #4828

- The main behavioral change is that we **no longer delete Pods from the job cache on update when they have a `DeletionTimestamp`**; we rely on the Pod delete event to remove them from the cache.  
- This ensures that `status.Terminating` correctly reflects the presence of terminating Pods and prevents the Job from leaving `Aborting` too early.  
- Please pay attention to any places that assume terminating Pods are already removed from the cache; the new behavior should be more consistent with the actual cluster state, but it’s worth double-checking for side effects.  
- If necessary, I can add/extend tests to cover the scenario where a Job stays in `Aborting` while Pods are in `Terminating` and only moves to `Aborted` once all Pods are fully deleted.

```release-note
Fix a bug where Jobs in the Aborting phase could transition to Aborted while some Pods were still terminating by keeping terminating Pods in the job cache until they are fully deleted.
```

### DIFF
--- a/pkg/controllers/job/job_controller_handler.go
+++ b/pkg/controllers/job/job_controller_handler.go
@@ -182,11 +182,6 @@ func (cc *jobcontroller) addPod(obj interface{}) {
 		return
 	}
 
-	if pod.DeletionTimestamp != nil {
-		cc.deletePod(pod)
-		return
-	}
-
 	req := apis.Request{
 		Namespace:   pod.Namespace,
 		JobName:     jobName,
@@ -230,11 +225,6 @@ func (cc *jobcontroller) updatePod(oldObj, newObj interface{}) {
 	}
 
 	if newPod.ResourceVersion == oldPod.ResourceVersion {
-		return
-	}
-
-	if newPod.DeletionTimestamp != nil {
-		cc.deletePod(newObj)
 		return
 	}
 
--- a/pkg/controllers/job/job_controller_handler_test.go
+++ b/pkg/controllers/job/job_controller_handler_test.go
@@ -334,6 +334,31 @@ func TestAddPodFunc(t *testing.T) {
 			},
 			ExpectedValue: 1,
 		},
+		{
+			// This case ensures a pod that is already terminating (has DeletionTimestamp)
+			// is still added to the job cache correctly.
+			Name: "AddPod Success when pod is terminating",
+			Job: &batch.Job{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "job1",
+					Namespace: namespace,
+				},
+			},
+			pods: []*v1.Pod{
+				func() *v1.Pod {
+					p := buildPod(namespace, "pod1", v1.PodPending, nil)
+					now := metav1.Now()
+					p.DeletionTimestamp = &now
+					return p
+				}(),
+			},
+			Annotation: map[string]string{
+				batch.JobNameKey:  "job1",
+				batch.JobVersion:  "0",
+				batch.TaskSpecKey: "task1",
+			},
+			ExpectedValue: 1,
+		},
 	}
 
 	for i, testcase := range testcases {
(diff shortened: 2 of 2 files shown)


hzxuzhonghu:
@kingeasternsun Thank you for your finding. I think i understand why we do so before. In the very early days, k8s deletion event can be lost even our controller is running. 

Now i think this will not happen, we can rely on pod deletion event to do cleanup.

kingeasternsun:
@hajnalmt 
Thanks for the review!
Regarding whether to cherry-pick this fix to other branches, I think it might be worth further discussion.

Although the bug is not critical, it can be misleading from a user’s perspective. After a user aborts a job, the job status changes to Aborted, which may lead them to assume that all resources have already been released. However, if they submit a new VCJob at this point, scheduling may fail because some pods are still in the Terminating state.

This behavior is not very intuitive for users and may cause confusion during normal usage.

hzxuzhonghu:
I agree with @hajnalmt this is not a critical bug, backporting it can be risky

PR #4848, opened 2025-12-22
Support scheduling based on nodes assigned in shard for volcano scheduler and agent scheduler

qi-min:
#### What type of PR is this?
feature

Support scheduler schedule pods based on nodes assigned in NodeShard, including synchronizing nodes changes in NodeShard CR. So that multi-schedulers can pick up nodes from shard and make parallel scheduling.

Scheduler can support three modes: no sharing|soft|hard
- **no-sharding:** schedule with in cluster just as current scheduling flow
- **soft:** schedule pods to nodes in NodeShard specified for current scheduler first. If no nodes meet scheduling policy, try nodes in other shards. 
> For network topology aware scheduling, restricted by the allocation flow, it is not guaranteed to use node within shard first across hypernodes in **soft** mode. Scheduler will try to use node within shard first in each hypernode, but it will  still use nodes in other shards in same hypernode if nodes within shard in this hypernode is not enough for job. )
- **hard:** restrict scheduling pods to nodes in NodeShard specified for current scheduler 

For more details please refer to: [proposal](https://docs.google.com/document/d/1NDTEGulhQ__sdZ_c2kN3ojtJIZie5O7_GyZyKIOLz-g/edit?tab=t.0#heading=h.2ou86daiptxu) | [design doc](https://github.com/volcano-sh/volcano/blob/master/docs/design/agent-scheduler.md#sharding-synchronization)

Fixes https://github.com/volcano-sh/volcano/issues/4722

```release-note

```

### DIFF
--- a/cmd/agent-scheduler/app/options/options.go
+++ b/cmd/agent-scheduler/app/options/options.go
@@ -118,9 +118,9 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.CacheDumpFileDir, "cache-dump-dir", "/tmp", "The target dir where the json file put at when dump cache info to json file")
 	fs.Uint32Var(&s.NodeWorkerThreads, "node-worker-threads", defaultNodeWorkers, "The number of threads syncing node operations.")
 	fs.BoolVar(&s.DisableDefaultSchedulerConfig, "disable-default-scheduler-config", false, "The flag indicates whether the scheduler should avoid using the default configuration if the provided scheduler configuration is invalid.")
-	fs.Uint32Var(&s.ScheduleWorkerCount, "scheduler-worker-count", defaultScheduleWorkerCount, "The flag indicates the number of worker threads for scheduling.")
-	fs.StringVar(&s.ShardingMode, "scheduler-sharding-mode", util.NoneShardingMode, "The node sharding mode for scheduling, none(default)|hard|soft mode is supported")
-	fs.StringVar(&s.ShardName, "scheduler-sharding-name", defaultShardName, "The name of shard used for this scheduler")
+	fs.Uint32Var(&s.ScheduleWorkerCount, "scheduler-worker-count", defaultScheduleWorkerCount, "The flag indicates the number of worker threads scheduling in parallel")
+	fs.StringVar(&s.ShardingMode, "scheduler-sharding-mode", util.NoneShardingMode, "The node sharding mode for scheduling, the mode could be none(default, schedule pod without shard)|hard(schedule pod within nodes in shard)|soft(chedule pod within nodes in shard in priority)")
+	fs.StringVar(&s.ShardName, "scheduler-sharding-name", defaultShardName, "The name of shard used for this scheduler, scheduler schedule pods based on nodes in shard if sharding mode is hard/soft")
 }
 
 // CheckOptionOrDie check leader election flag when LeaderElection is enabled.
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -736,11 +736,11 @@ func (alloc *Action) prioritizeNodes(ssn *framework.Session, task *api.TaskInfo,
 		}
 	}
 
-	//To allocate to nodes with enough resource and nodes within shard of this scheduler first, allocation of Pod follow below order:
-	//1. Node with IDLE resource in shard for this scheduler
-	//2. Node with IDLE resource in shard for other scheduler  (empty if sharding mode is not soft)
-	//3. Node with Future IDLE resource in shard for this scheduler
-	//4. Node with Future IDLE resource in shard for other scheduler (empty if sharding mode is not soft)
+	// To allocate to nodes with enough resource and nodes within shard of this scheduler first, allocation of Pod follow below order:
+	// 1. Node with IDLE resource in shard for this scheduler
+	// 2. Node with IDLE resource in shard for other scheduler  (empty if sharding mode is not soft)
+	// 3. Node with Future IDLE resource in shard for this scheduler
+	// 4. Node with Future IDLE resource in shard for other scheduler (empty if sharding mode is not soft)
 	candidateNodes = append(candidateNodes, idleCandidateNodes)
 	candidateNodes = append(candidateNodes, idleCandidateNodesInOtherShards)
 	candidateNodes = append(candidateNodes, futureIdleCandidateNodes)
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -61,7 +61,6 @@ import (
 	"volcano.sh/apis/pkg/apis/scheduling"
 	schedulingscheme "volcano.sh/apis/pkg/apis/scheduling/scheme"
 	vcv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
-	nodeshardv1alpha1 "volcano.sh/apis/pkg/apis/shard/v1alpha1"
 	vcclient "volcano.sh/apis/pkg/client/clientset/versioned"
 	"volcano.sh/apis/pkg/client/clientset/versioned/scheme"
 	vcinformer "volcano.sh/apis/pkg/client/informers/externalversions"
@@ -75,6 +74,7 @@ import (
 	schedulingapi "volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/metrics/source"
+	"volcano.sh/volcano/pkg/util"
 	commonutil "volcano.sh/volcano/pkg/util"
 )
 
@@ -149,7 +149,8 @@ type SchedulerCache struct {
 	defaultPriority      int32
 	CSINodesStatus       map[string]*schedulingapi.CSINodeStatusInfo
 	HyperNodesInfo       *schedulingapi.HyperNodesInfo
-	InUseNodesInShard    sets.Set[string]
+	// InUseNodesInShard cached the nodes that are immediatly available for current shard (desiredNodes of this shard - inUseNodes in other shards)
+	InUseNodesInShard sets.Set[string]
 
 	NamespaceCollection map[string]*schedulingapi.NamespaceCollection
 
@@ -375,10 +376,6 @@ func (su *defaultStatusUpdater) UpdateQueueStatus(queue *schedulingapi.QueueInfo
 	return nil
 }
 
-func (su *defaultStatusUpdater) UpdateNodeShardStatus(nodeshard *nodeshardv1alpha1.NodeShard) (*nodeshardv1alpha1.NodeShard, error) {
-	return su.vcclient.ShardV1alpha1().NodeShards().UpdateStatus(context.Background(), nodeshard, metav1.UpdateOptions{})
-}
-
 type podgroupBinder struct {
 	kubeclient kubernetes.Interface
 	vcclient   vcclient.Interface
@@ -532,29 +529,32 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, defaultQueu
 	)
 
 	sc := &SchedulerCache{
-		Jobs:                   make(map[schedulingapi.JobID]*schedulingapi.JobInfo),
-		Nodes:                  make(map[string]*schedulingapi.NodeInfo),
-		Queues:                 make(map[schedulingapi.QueueID]*schedulingapi.QueueInfo),
-		PriorityClasses:        make(map[string]*schedulingv1.PriorityClass),
-		errTasks:               workqueue.NewTypedRateLimitingQueue[string](errTaskRateLimiter),
-		nodeQueue:              workqueue.NewTypedRateLimitingQueue[string](workqueue.DefaultTypedControllerRateLimiter[string]()),
-		DeletedJobs:            workqueue.NewTypedRateLimitingQueue[string](deletedJobsRateLimiter),
-		hyperNodesQueue:        workqueue.NewTypedRateLimitingQueue[string](workqueue.DefaultTypedControllerRateLimiter[string]()),
-		kubeClient:             kubeClient,
-		vcClient:               vcClient,
-		restConfig:             config,
-		defaultQueue:           defaultQueue,
-		schedulerNames:         schedulerNames,
-		nodeSelectorLabels:     make(map[string]sets.Empty),
-		NamespaceCollection:    make(map[string]*schedulingapi.NamespaceCollection),
-		CSINodesStatus:         make(map[string]*schedulingapi.CSINodeStatusInfo),
-		imageStates:            make(map[string]*imageState),
-		InUseNodesInShard:      sets.Set[string]{},
-		shardUpdateCoordinator: NewShardUpdateCoordinator(),
-		NodeShards:             make(map[string]*schedulingapi.NodeShardInfo),
+		Jobs:                make(map[schedulingapi.JobID]*schedulingapi.JobInfo),
+		Nodes:               make(map[string]*schedulingapi.NodeInfo),
+		Queues:              make(map[schedulingapi.QueueID]*schedulingapi.QueueInfo),
+		PriorityClasses:     make(map[string]*schedulingv1.PriorityClass),
+		errTasks:            workqueue.NewTypedRateLimitingQueue[string](errTaskRateLimiter),
+		nodeQueue:           workqueue.NewTypedRateLimitingQueue[string](workqueue.DefaultTypedControllerRateLimiter[string]()),
+		DeletedJobs:         workqueue.NewTypedRateLimitingQueue[string](deletedJobsRateLimiter),
+		hyperNodesQueue:     workqueue.NewTypedRateLimitingQueue[string](workqueue.DefaultTypedControllerRateLimiter[string]()),
+		kubeClient:          kubeClient,
+		vcClient:            vcClient,
+		restConfig:          config,
+		defaultQueue:        defaultQueue,
+		schedulerNames:      schedulerNames,
+		nodeSelectorLabels:  make(map[string]sets.Empty),
+		NamespaceCollection: make(map[string]*schedulingapi.NamespaceCollection),
+		CSINodesStatus:      make(map[string]*schedulingapi.CSINodeStatusInfo),
+		imageStates:         make(map[string]*imageState),
+		InUseNodesInShard:   sets.Set[string]{},
+		NodeShards:          make(map[string]*schedulingapi.NodeShardInfo),
 
 		NodeList:    []string{},
 		nodeWorkers: nodeWorkers,
+	}
+
+	if options.ServerOpts.ShardingMode == util.HardShardingMode || options.ServerOpts.ShardingMode == util.SoftShardingMode {
+		sc.shardUpdateCoordinator = NewShardUpdateCoordinator()
 	}
 
 	sc.resyncPeriod = resyncPeriod
@@ -1672,20 +1672,6 @@ func (sc *SchedulerCache) updateJobInfo(job *schedulingapi.JobInfo) {
 // UpdateQueueStatus update the status of queue.
 func (sc *SchedulerCache) UpdateQueueStatus(queue *schedulingapi.QueueInfo) error {
 	return sc.StatusUpdater.UpdateQueueStatus(queue)
-}
-
-// UpdateNodeShardStatus update the status of nodeshard
-func (sc *SchedulerCache) UpdateNodeShardStatus(nodeShardName string) error {
-	if nodeShard := sc.generateNodeShardWithStatus(nodeShardName); nodeShard != nil {
-		klog.V(3).Infof("Update NodeShard %s status...", nodeShardName)
-		_, err := sc.StatusUpdater.UpdateNodeShardStatus(nodeShard)
-		if err != nil {
-			klog.Errorf("Failed to update NodeShard %s status %v", nodeShard.Name, err)
-			return err
-		}
-		klog.V(3).Infof("Updated NodeShard %s status", nodeShard.Name)
-	}
-	return nil
 }
 
 func (sc *SchedulerCache) recordPodGroupEvent(podGroup *schedulingapi.PodGroup, eventType, reason, msg string) {
--- a/pkg/scheduler/cache/shard_coordinator.go
+++ b/pkg/scheduler/cache/shard_coordinator.go
@@ -17,8 +17,10 @@ limitations under the License.
 package cache
 
 import (
+	"context"
 	"sync/atomic"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
@@ -64,27 +66,44 @@ func (sc *SchedulerCache) RefreshNodeShards() {
 }
 
 func (sc *SchedulerCache) tryUpdateNodeShardStatus(nodeShardName string) {
-	if sc.shardUpdateCoordinator.IsSessionRunning.Load() {
-		// Try to set pending flag atomically - if already pending, skip this request
-		if !sc.shardUpdateCoordinator.ShardUpdatePending.CompareAndSwap(false, true) {
-			klog.V(3).Infof("Update status of NodeShard is already pending, skip this request")
+	coordinator := sc.shardUpdateCoordinator
+	if coordinator != nil {
+		if coordinator.IsSessionRunning.Load() {
+			// Try to set pending flag atomically - if already pending, skip this request
+			if !coordinator.ShardUpdatePending.CompareAndSwap(false, true) {
+				klog.V(3).Infof("Update status of NodeShard is already pending, skip this request")
+				return
+			}
+			klog.V(3).Infof("Update status of NodeShard is pending because session is running")
+
+			// Wait for session to end
+			<-coordinator.SessionEndCh
+			klog.V(3).Infof("Update status of NodeShard is resumed")
+
+			// Double-check session has ended and we should still update
+			if !coordinator.IsSessionRunning.Load() {
+				sc.UpdateNodeShardStatus(nodeShardName)
+			}
+			// Reset pending flag
+			coordinator.ShardUpdatePending.Store(false)
 			return
 		}
-		klog.V(3).Infof("Update status of NodeShard is pending because session is running")
-
-		// Wait for session to end
-		<-sc.shardUpdateCoordinator.SessionEndCh
-		klog.V(3).Infof("Update status of NodeShard is resumed")
-
-		// Double-check session has ended and we should still update
-		if !sc.shardUpdateCoordinator.IsSessionRunning.Load() {
-			sc.UpdateNodeShardStatus(nodeShardName)
-		}
-		// Reset pending flag
-		sc.shardUpdateCoordinator.ShardUpdatePending.Store(false)
-		return
+		sc.UpdateNodeShardStatus(nodeShardName)
 	}
-	sc.UpdateNodeShardStatus(nodeShardName)
+}
+
+// UpdateNodeShardStatus update the status of nodeshard
+func (sc *SchedulerCache) UpdateNodeShardStatus(nodeShardName string) error {
+	if nodeShard := sc.generateNodeShardWithStatus(nodeShardName); nodeShard != nil {
+		klog.V(3).Infof("Update NodeShard %s status...", nodeShardName)
+		_, err := sc.StatusUpdater.UpdateNodeShardStatus(nodeShard)
+		if err != nil {
+			klog.Errorf("Failed to update NodeShard %s status %v", nodeShard.Name, err)
+			return err
+		}
+		klog.V(3).Infof("Updated NodeShard %s status", nodeShard.Name)
+	}
+	return nil
 }
 
 // generateNodeShardWithStatus generate nodeshard with updated status. return nil if no status change
@@ -134,4 +153,8 @@ func (sc *SchedulerCache) notifySessionEnd() {
 	default:
 		// No shardUpdateCoordinator goroutine is waiting, which is fine
 	}
+}
+
+func (su *defaultStatusUpdater) UpdateNodeShardStatus(nodeshard *nodeshardv1alpha1.NodeShard) (*nodeshardv1alpha1.NodeShard, error) {
+	return su.vcclient.ShardV1alpha1().NodeShards().UpdateStatus(context.Background(), nodeshard, metav1.UpdateOptions{})
 }
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -32,7 +32,6 @@ import (
 
 // OpenSession start the session
 func OpenSession(cache cache.Cache, tiers []conf.Tier, configurations []conf.Configuration) *Session {
-	cache.OnSessionOpen()
 	ssn := openSession(cache)
 	ssn.Tiers = tiers
 	ssn.Configurations = configurations
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -164,6 +164,7 @@ type Session struct {
 }
 
 func openSession(cache cache.Cache) *Session {
+	cache.OnSessionOpen()
 	ssn := &Session{
 		UID:             uuid.NewUUID(),
 		kubeClient:      cache.Client(),
@@ -604,6 +605,8 @@ func closeSession(ssn *Session) {
 	ssn.clusterOrderFns = nil
 	ssn.NodeList = nil
 	ssn.TotalResource = nil
+
+	ssn.cache.OnSessionClose()
 
 	klog.V(3).Infof("Close Session %v", ssn.UID)
 }


cmd/scheduler/app/options/options.go
```diff
@@ -156,6 +168,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.Uint32Var(&s.NodeWorkerThreads, "node-worker-threads", defaultNodeWorkers, "The number of threads syncing node operations.")
 	fs.StringSliceVar(&s.IgnoredCSIProvisioners, "ignored-provisioners", nil, "The provisioners that will be ignored during pod pvc request computation and preemption.")
 	fs.BoolVar(&s.DisableDefaultSchedulerConfig, "disable-default-scheduler-config", false, "The flag indicates whether the scheduler should avoid using the default configuration if the provided scheduler configuration is invalid.")
+	fs.StringVar(&s.ShardingMode, "scheduler-sharding-mode", util.NoneShardingMode, "The node sharding mode for scheduling")
```
hzxuzhonghu: Can you document it clearly

cmd/scheduler/app/options/options.go
```diff
@@ -156,6 +168,8 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
...
+	fs.StringVar(&s.ShardName, "scheduler-sharding-name", defaultSchedulerName, "The name of shard used for this scheduler")
```
hzxuzhonghu: maybe we donot need a flag to indicate the name of nodeshard, just a const. 
Should not use scheduler name, even they have same value, which is just confusing

cmd/scheduler/app/options/options.go
```diff
@@ -206,8 +220,10 @@ func (s *ServerOption) ParseCAFiles(decryptFunc DecryptFunc) error {
 
 // Default new and registry a default one
 func Default() *ServerOption {
-	s := NewServerOption()
-	s.AddFlags(pflag.CommandLine)
-	s.RegisterOptions()
-	return s
+	once.Do(func() {
```
hzxuzhonghu, line 224: It is not a good practice using once to initialize options.

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -223,7 +223,7 @@ func SelectBestHyperNodeAndScore(hyperNodeScores map[float64][]string) (string,
 }
 
 // SelectBestNodesAndScores returns the best N node whose score is highest N score, pick one randomly if there are many nodes with same score.
-func SelectBestNodesAndScores(nodeScores map[float64][]*api.NodeInfo, count int) ([]*api.NodeInfo, []float64) {
+func SelectBestNodesAndScores(nodeScores map[float64][]*api.NodeInfo, nodesInShard sets.Set[string], shardMode string, count int) ([]*api.NodeInfo, []float64) {
```
hzxuzhonghu: why pass `nodesInShard` and mode in?

pkg/agentscheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,244 @@
...
+		return sets.Set[string]{}
+	}
+	sc.mutex.RLock()
+	defer sc.mutex.RUnlock()
+	latest := atomic.LoadInt64(&sc.latestRevision)
+	atomic.StoreInt64(&state.revisionInScheduing, latest)
+	if klog.V(5).Enabled() {
+		klog.V(5).Infof("Worker %d will schedule with nodes%v", workerIdx, sc.nodeToUse.UnsortedList())
+	}
+	return sc.nodeToUse
```
hzxuzhonghu, line 88: I donot see how is the above check related. Seems this is just returning sc.nodeToUse

pkg/agentscheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,244 @@
...
+
+func NewShardCoordinator(cache Cache, workerCount int, schedulerName string, shardingMode string) *ShardCoordinator {
+	klog.V(3).Infof("Shard Coordinator is initialized")
+	workerStates := make([]*workerNodeShardState, workerCount)
+	for i := range workerCount {
+		workerStates[i] = &workerNodeShardState{}
+	}
+
+	sc := &ShardCoordinator{
+		schedulerShardName: schedulerName,
```
hzxuzhonghu: as i commented above donot mess scheduler name with shard name

pkg/agentscheduler/cache/cache.go
```diff
@@ -92,17 +96,19 @@ type SchedulerCache struct {
...
+	Binder            Binder
+	StatusUpdater     StatusUpdater
 
 	Recorder record.EventRecorder
 
-	Nodes    map[string]*schedulingapi.NodeInfo // TODO: do we need to also add a seperate lock for Nodes cache?
-	NodeList []string
+	Nodes      map[string]*schedulingapi.NodeInfo // TODO: do we need to also add a seperate lock for Nodes cache?
+	NodeList   []string
+	NodeShards map[string]*schedulingapi.NodeShardInfo
```
hzxuzhonghu, line 111: As i read above, i understand it has only one node shard. But this map makes me confusing,

pkg/agentscheduler/cache/cache.go
```diff
@@ -358,7 +371,7 @@ func newSchedulerCache(config *rest.Config, schedulerName string, nodeSelectors
 	)
 
 	sc.ConflictAwareBinder = NewConflictAwareBinder(sc, sc.schedulingQueue)
-
+	sc.ShardCoordinator = NewShardCoordinator(sc, int(opt.ScheduleWorkerCount), opt.ShardName, opt.ShardingMode)
```
hzxuzhonghu: Should not create coordinator if sharding mode is `none`

pkg/agentscheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,244 @@
...
+		workerStates:       workerStates,
+		shardingEnabled:    shardingMode == util.HardShardingMode || shardingMode == util.SoftShardingMode,
+		cache:              cache,
+		updateChan:         make(chan struct{}, 100), // buffered channel to handle multiple update requests
```
hzxuzhonghu, line 62: not sure what is this channel used for

pkg/agentscheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,244 @@
...
+			sc.schedulerNodeShardInfo = shard
+			shardForSchedulerFound = true
+			break
+		}
+	}
+	needUpdate := false
+	if !shardForSchedulerFound {
+		klog.Errorf("Sharding is enabled but no shard is defined for this scheduler!")
+		sc.schedulerNodeShardInfo = nil
+	} else if availableNodes := sc.getAvailableNodesFromShard(); desiredNodesChanged || !sc.nodeToUse.Equal(availableNodes) {
```
hzxuzhonghu, line 123: can you expalin why need two conditions check

pkg/agentscheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,244 @@
...
+		atomic.AddInt64(&sc.latestRevision, 1)
+		sc.nodeToUse = availableNodes
+		needUpdate = true
+	}
+	sc.mutex.Unlock()
```
hzxuzhonghu: The unlock is far from lock, which can be dangerous if someone return before Unlock

hzxuzhonghu: can abstract a function and use defer

pkg/scheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (sc *SchedulerCache) tryUpdateNodeShardStatus(nodeShardName string) {
+	sc.shardUpdateCoordinator.ShardUpdateMu.Lock()
+
+	if sc.shardUpdateCoordinator.IsSessionRunning {
+		sc.shardUpdateCoordinator.ShardUpdatePending = true
+		klog.V(3).Infof("Update status of nodeshard is pending because session is running")
+
+		sc.shardUpdateCoordinator.ShardUpdateCond.Wait()
```
hzxuzhonghu: I guess we need a workflow graph to help understand the logic

pkg/agentscheduler/actions/allocate/allocate.go
hzxuzhonghu, line 88: about the performance: so we still need to cache all nodes when sharding enabled, especially for hard mode?

cmd/scheduler/app/options/options.go
qi-min: @hzxuzhonghu using flag just in case we need support same scheduler but multiple shard, e.g. two volcano scheduler schedule in two shards, so const is not used here.

I can add a another const as default shard name to avoid the confusing you mentioned

pkg/scheduler/util/scheduler_helper.go
qi-min: seems a missing part after code change, they are removed now

pkg/agentscheduler/cache/shard_coordinator.go
qi-min: getNodesForScheduling just return sc.nodeToUse and record the revision worker used.
klog.V(5).Enabled is removed now

pkg/agentscheduler/cache/cache.go
qi-min: each scheduler has a node shard, so there are multiple node shard. Scheduler need to check other node shard to decide when to add node into cache, it depends on is that node being used in other node shard

hzxuzhonghu: agent-scheduler has multiple schedulers?

cmd/scheduler/app/options/options.go
qi-min: I see this function is used for test, and addFlags twice will lead error, that's why change it to once

pkg/agentscheduler/cache/cache.go
qi-min: ok

pkg/agentscheduler/cache/shard_coordinator.go
qi-min: just in case multi update nodeshard status request come in, then update will only be executed once

pkg/agentscheduler/cache/shard_coordinator.go
qi-min: If desired nodes changes, then nodeshard status need be changed.
if desired nodes are not changed, but inuse nodes in other shard changes which leads available nodes for current shard change `(!sc.nodeToUse.Equal(availableNodes)`), then status of current shard also need be changed

pkg/agentscheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,244 @@
...
+// getAvailableNodesFromShard get available nodes based on desired nodes. Nodes are still being used in other shard should not be put into available nodes
+func (sc *ShardCoordinator) getAvailableNodesFromShard() sets.Set[string] {
+	nodes := sc.schedulerNodeShardInfo.NodeDesired
+	for shardName, nodeShard := range sc.nodeShardInfos {
+		if shardName != sc.schedulerShardName {
+			nodes = nodes.Difference(nodeShard.NodeInUse)
+		}
+	}
+	return nodes
+}
```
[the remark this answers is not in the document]
qi-min: getAvailableNodesFromShard is called in RefreshNodeShards with lock

cmd/agent-scheduler/app/options/options.go
```diff
@@ -144,7 +117,9 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.CacheDumpFileDir, "cache-dump-dir", "/tmp", "The target dir where the json file put at when dump cache info to json file")
 	fs.Uint32Var(&s.NodeWorkerThreads, "node-worker-threads", defaultNodeWorkers, "The number of threads syncing node operations.")
 	fs.BoolVar(&s.DisableDefaultSchedulerConfig, "disable-default-scheduler-config", false, "The flag indicates whether the scheduler should avoid using the default configuration if the provided scheduler configuration is invalid.")
-	fs.Uint32Var(&s.ScheduleWorkerCount, "worker-count", defaultScheduleWorkerCount, "The flag indicates the number of worker threads for scheduling.")
+	fs.Uint32Var(&s.ScheduleWorkerCount, "scheduler-worker-count", defaultScheduleWorkerCount, "The flag indicates the number of worker threads for scheduling.")
+	fs.StringVar(&s.ShardingMode, "scheduler-sharding-mode", util.NoneShardingMode, "The node sharding mode for scheduling")
+	fs.StringVar(&s.ShardName, "scheduler-sharding-name", agentSchedulerName, "The name of shard used for this scheduler")
```
wangyang0616: How many shards are currently supported? Is it one shard for the default-scheduler and one for the agent-scheduler, or is there no limit on the number of shards that each scheduler can use?

wangyang0616: If there is no limit on the number of shards each scheduler can use, will the definition here still be accurate when the agent-scheduler owns multiple shards?

cmd/scheduler/app/options/options.go
qi-min: https://github.com/volcano-sh/volcano/blob/master/docs/design/agent-scheduler.md#scheduler-configuration
sharding mode is documented here
description is also updated  with valid values: `none(default)|hard|soft mode is supported`

pkg/agentscheduler/cache/cache.go
qi-min: For now, agent-scheduler just one scheduler. But at least we have both volcano and agent-scheduler, so at least two shards. 

In future, sharding may also support other scheduling parallel mode,  like 2 volcano scheduler in 2 shards to improve throughput. So the shards may increase,

pkg/agentscheduler/actions/allocate/allocate.go
qi-min: For now all nodes are cached but only nodes in shard are used for predicating. One reason is to adapt for the existing cache framework (nodeinfo and taskinfo), nodeinfo in cache can be picked up once nodes in shard are changed.  
If nodeinfo is not cached, nodeinfo(including pods of this node) need be initialized when new nodes added into shard, it bring more changes.

I think we can optimize the snapshot for scheduler, just snapshot nodes in shard for hard mode to improve performance of scheduling

qi-min:
> I would suggest not making the plugins aware of node shards. We can just separate nodes before, and then call different plugins two times. I think this would make the logic more clear

@hzxuzhonghu which plugin do you mean? For now I see plugins are not aware of shards. The major logic is handled in actions, like allocate or preempt with nodes in two groups

pkg/agentscheduler/cache/shard_coordinator.go
qi-min: shardName is defined instead of using scheduler name directly

pkg/agentscheduler/cache/cache.go
```diff
@@ -125,14 +131,19 @@ type SchedulerCache struct {
 	// sharedDRAManager is used in DRA plugin, contains resourceClaimTracker, resourceSliceLister and deviceClassLister
 	sharedDRAManager k8sframework.SharedDRAManager
 
+	// ConflictAwareBinder resolve confilct caused by multi workers parallel allocation
 	ConflictAwareBinder *ConflictAwareBinder
 
+	ShardCoordinator *ShardCoordinator
```
wangyang0616, line 138: It is recommended to add comments to explain the meaning conveyed by ShardCoordinator.

pkg/agentscheduler/cache/shard_coordinator.go
qi-min: ok, new function is abstracted

hzxuzhonghu:
@qi-min Sorry should be: util's helper functions

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -709,21 +711,35 @@ func (alloc *Action) prioritizeNodes(ssn *framework.Session, task *api.TaskInfo,
...
 		} else {
 			klog.V(5).Infof("Predicate filtered node %v, idle: %v and future idle: %v do not meet the requirements of task: %v",
 				n.Name, n.Idle, n.FutureIdle(), task.Name)
 		}
 	}
+
 	candidateNodes = append(candidateNodes, idleCandidateNodes)
+	candidateNodes = append(candidateNodes, idleCandidateNodesInOtherShards)
 	candidateNodes = append(candidateNodes, futureIdleCandidateNodes)
+	candidateNodes = append(candidateNodes, futureIdleCandidateNodesInOtherShards)
```
wangyang0616, line 747: This part has become increasingly complex. It is recommended to add comments to indicate the principles and order of candidate node filtering.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -299,13 +299,22 @@ func (pmpt *Action) preempt(
...
 	// we should filter out those nodes that are UnschedulableAndUnresolvable status got in allocate action
 	allNodes := ssn.FilterOutUnschedulableAndUnresolvableNodesForTask(preemptor)
-	predicateNodes, _ := predicateHelper.PredicateNodes(preemptor, allNodes, ssn.PredicateForPreemptAction, pmpt.enablePredicateErrorCache)
-
-	if pmpt.enableTopologyAwarePreemption {
-		return pmpt.topologyAwarePreempt(ssn, stmt, preemptor, filter, predicateNodes)
+	predicateNodes, _ := predicateHelper.PredicateNodes(preemptor, allNodes, ssn.PredicateForPreemptAction, pmpt.enablePredicateErrorCache, ssn.NodesInShard)
+
+	candidateNodes := util.GetPredicatedNodeByShard(predicateNodes, ssn.NodesInShard)
+	var result bool
```
wangyang0616: It is recommended to rename result to a name that reflects its actual meaning.

pkg/scheduler/cache/cache.go
```diff
@@ -1649,6 +1669,39 @@ func (sc *SchedulerCache) UpdateQueueStatus(queue *schedulingapi.QueueInfo) erro
...
+	}
+
+	nodesInUse := sets.New(nodeShard.NodeShard.Status.NodesInUse...)
+	if sc.InUseNodesInShard.Equal(nodesInUse) {
+		// nodes to use is same as nodes in use recorded in nodeshard
+		return nil
+	}
+
+	// Create a deep copy to avoid modifying cache objects
+	nodeShardCopy := nodeShard.NodeShard.DeepCopy()
```
wangyang0616: Does NodeShard involve concurrency? Is it necessary to add a lock when performing a deep copy here?

cmd/agent-scheduler/app/options/options.go
qi-min: @wangyang0616 one shard for one scheduler. Now we need one for volcano and one for agent scheduler.
In future, one type of scheduler may have multiple shards, but still one scheduler instance one shard.

pkg/scheduler/cache/shard_coordinator.go
qi-min: https://github.com/volcano-sh/volcano/blob/8e4baf0b14f72d4c302711820bdbabcfbb1761bc/docs/design/agent-scheduler.md#shard-coordinator-in-volcano-scheduler
a flow graph is added here

pkg/agentscheduler/cache/cache.go
qi-min: done

pkg/scheduler/actions/allocate/allocate.go
qi-min: sure, comments are added

pkg/scheduler/actions/preempt/preempt.go
qi-min: ok, var is renamed to preemptSuccess

pkg/scheduler/cache/cache.go
qi-min: lock is added in latest commit for other fix

qi-min:
> @qi-min Sorry should be: util's helper functions

@hzxuzhonghu here are some consideration why util helper is aware of nodes in shard: 
- The `PredicateNodes` function in util helper is shared in multiple actions, so it is a central place to control nodes predication
- `PredicateNodes` func in helper write the nodes fit err during predication. Error "node isn't in scheduler node shard" can also be recorded here and displayed in pod event for hard  sharding mode, we need this info to let user know why node is not fit for the pod.
- `GetPredicatedNodeByShard` just a abstraction function to separate nodes in two groups for soft sharding mode. I put it here just to avoid duplicated implementation in actions.  This can be moved back to actions, but not sure is it necessary.

qi-min:
@wangyang0616 @hzxuzhonghu review comments are handled and new commit is pushed, would you please take a review again, thanks.

pkg/scheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,97 @@
...
+			klog.V(3).Infof("Update status of NodeShard is already pending, skip this request")
+			return
+		}
+		sc.shardUpdateCoordinator.ShardUpdatePending = true
+		klog.V(3).Infof("Update status of NodeShard is pending because session is running")
+		sc.shardUpdateCoordinator.ShardUpdateCond.Wait()
+		klog.V(3).Infof("Update status of NodeShard is resumed")
+		// when multiple update request are trying to acquire lock, only one request will do update
+		if sc.shardUpdateCoordinator.ShardUpdatePending && !sc.shardUpdateCoordinator.IsSessionRunning {
+			sc.UpdateNodeShardStatus(nodeShardName)
```
wangyang0616: There is nested locking here. Could there be a risk of deadlock in concurrent scenarios?

pkg/scheduler/cache/cache.go
```diff
@@ -1649,6 +1674,44 @@ func (sc *SchedulerCache) UpdateQueueStatus(queue *schedulingapi.QueueInfo) erro
...
+	nodesToAdd := desiredNodes.Difference(nodesInUse)
+	if nodesInUse.Equal(oldNodesInUse) && nodesToRemove.Equal(oldNodesToRemove) && nodesToAdd.Equal(oldNodesToAdd) {
+		klog.V(3).Infof("Skip update NodeShard %s status, no change", nodeShard.Name)
+		return nil
+	}
+	nodeShardCopy.Status.NodesInUse = nodesInUse.UnsortedList()
+	nodeShardCopy.Status.NodesToRemove = nodesToRemove.UnsortedList()
+	nodeShardCopy.Status.NodesToAdd = nodesToAdd.UnsortedList()
+
+	_, err := sc.StatusUpdater.UpdateNodeShardStatus(nodeShardCopy)
```
wangyang0616: The function is wrapped with a lock when called externally (e.g., sc.shardUpdateCoordinator.ShardUpdateMu.Lock()). Since an API server call is made here, could this result in the lock being held for an excessively long time, leading to a significant degradation in concurrent performance?

pkg/scheduler/cache/cache.go
wangyang0616: It is recommended to extract the content that requires locking into a separate function for processing.

wangyang0616:
It is recommended to conduct a comprehensive review of the newly added locks, including lock nesting and external calls made while the lock is held, to avoid potential deadlocks and dependencies on external calls during the locking period.

pkg/scheduler/cache/shard_coordinator.go
qi-min: @wangyang0616 From current logic, deadlock will not happen on shardUpdateCoordinator.ShardUpdateMu because the inner lock sc.mutex does not depend on other lock. But for safety and less confusion, I moved ShardUpdateMu and use atomic RW and channel instead, please check,

pkg/scheduler/cache/cache.go
qi-min: `ShardUpdateMu` is removed now and sc.mutex does not lock apiserver call, so API server call will not block releasing the lock. I also abstract the data preparation into a function `getAvailableNodesFromShard`  to make lock logic more clear.

qi-min:
> It is recommended to conduct a comprehensive review of the newly added locks, including lock nesting and external calls made while the lock is held, to avoid potential deadlocks and dependencies on external calls during the locking period.

I have removed the nested lock  to lower the risk of deadlock

pkg/agentscheduler/actions/allocate/allocate.go
hzxuzhonghu: SG to keep it this way

cmd/agent-scheduler/app/options/options.go
```diff
@@ -144,7 +118,9 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.CacheDumpFileDir, "cache-dump-dir", "/tmp", "The target dir where the json file put at when dump cache info to json file")
 	fs.Uint32Var(&s.NodeWorkerThreads, "node-worker-threads", defaultNodeWorkers, "The number of threads syncing node operations.")
 	fs.BoolVar(&s.DisableDefaultSchedulerConfig, "disable-default-scheduler-config", false, "The flag indicates whether the scheduler should avoid using the default configuration if the provided scheduler configuration is invalid.")
-	fs.Uint32Var(&s.ScheduleWorkerCount, "worker-count", defaultScheduleWorkerCount, "The flag indicates the number of worker threads for scheduling.")
+	fs.Uint32Var(&s.ScheduleWorkerCount, "scheduler-worker-count", defaultScheduleWorkerCount, "The flag indicates the number of worker threads for scheduling.")
+	fs.StringVar(&s.ShardingMode, "scheduler-sharding-mode", util.NoneShardingMode, "The node sharding mode for scheduling, none(default)|hard|soft mode is supported")
```
hzxuzhonghu: need more elaborations on  theses values

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -709,21 +711,40 @@ func (alloc *Action) prioritizeNodes(ssn *framework.Session, task *api.TaskInfo,
...
 			klog.V(5).Infof("Predicate filtered node %v, idle: %v and future idle: %v do not meet the requirements of task: %v",
 				n.Name, n.Idle, n.FutureIdle(), task.Name)
 		}
 	}
+
+	//To allocate to nodes with enough resource and nodes within shard of this scheduler first, allocation of Pod follow below order:
+	//1. Node with IDLE resource in shard for this scheduler
+	//2. Node with IDLE resource in shard for other scheduler  (empty if sharding mode is not soft)
+	//3. Node with Future IDLE resource in shard for this scheduler
+	//4. Node with Future IDLE resource in shard for other scheduler (empty if sharding mode is not soft)
```
hzxuzhonghu: NIT: add a space after `//`

pkg/scheduler/cache/cache.go
```diff
@@ -769,6 +783,15 @@ func (sc *SchedulerCache) addEventHandler() {
 		DeleteFunc: sc.DeleteHyperNode,
 	})
 
+	if options.ServerOpts.ShardingMode == commonutil.HardShardingMode || options.ServerOpts.ShardingMode == commonutil.SoftShardingMode {
+		sc.nodeShardInformer = sc.vcInformerFactory.Shard().V1alpha1().NodeShards()
+		sc.nodeShardInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
hzxuzhonghu, line 788: You should also update WaitForCacheSync now

pkg/scheduler/cache/cache.go
```diff
@@ -1649,6 +1674,20 @@ func (sc *SchedulerCache) UpdateQueueStatus(queue *schedulingapi.QueueInfo) erro
 	return sc.StatusUpdater.UpdateQueueStatus(queue)
 }
 
+// UpdateNodeShardStatus update the status of nodeshard
+func (sc *SchedulerCache) UpdateNodeShardStatus(nodeShardName string) error {
```
hzxuzhonghu: In order to reduce mutating `SchedulerCache`, we can move all  the node shards related functions  to a separate file.  Cache is already very large and complex.

pkg/scheduler/framework/framework.go
```diff
@@ -32,6 +32,7 @@ import (
 
 // OpenSession start the session
 func OpenSession(cache cache.Cache, tiers []conf.Tier, configurations []conf.Configuration) *Session {
+	cache.OnSessionOpen()
```
hzxuzhonghu: can you move in openSession

pkg/scheduler/framework/framework.go
```diff
@@ -66,4 +67,5 @@ func CloseSession(ssn *Session) {
 	}
 
 	closeSession(ssn)
+	ssn.cache.OnSessionClose()
```
hzxuzhonghu, line 69: same move into closeSession

pkg/scheduler/cache/cache.go
```diff
@@ -521,23 +532,26 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, defaultQueu
...
+		vcClient:               vcClient,
+		restConfig:             config,
+		defaultQueue:           defaultQueue,
+		schedulerNames:         schedulerNames,
+		nodeSelectorLabels:     make(map[string]sets.Empty),
+		NamespaceCollection:    make(map[string]*schedulingapi.NamespaceCollection),
+		CSINodesStatus:         make(map[string]*schedulingapi.CSINodeStatusInfo),
+		imageStates:            make(map[string]*imageState),
+		InUseNodesInShard:      sets.Set[string]{},
+		shardUpdateCoordinator: NewShardUpdateCoordinator(),
```
hzxuzhonghu: should you init it on demand?

pkg/scheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// RefreshNodeShards update node shards cached in coordinator
+func (sc *SchedulerCache) RefreshNodeShards() {
+	if options.ServerOpts.ShardingMode != util.HardShardingMode && options.ServerOpts.ShardingMode != util.SoftShardingMode {
+		return
+	}
+	var nodeShardInfo *api.NodeShardInfo
+	for shardName, shard := range sc.NodeShards {
+		if shardName == options.ServerOpts.ShardName {
```
hzxuzhonghu, line 54: We can optimize the informer to only watch with like 
```
	// 2. Define the TweakListOptions function
	// This function adds a FieldSelector to the List and Watch calls
	tweakListOptions := func(options *metav1.ListOptions) {
		options.FieldSelector = fields.OneTermEqualSelector("metadata.name", podName).String()
	}

	// 3. Create the Filtered SharedInformerFactory
	// Params: client, resyncPeriod, namespace, tweakListOptions
	factory := informers.NewFilteredSharedInformerFactory(
		clientset,
		10*time.Minute,
		namespace,
		tweakListOptions,
	)

```

hzxuzhonghu: Then we donot need to check sharname later

pkg/scheduler/cache/cache.go
```diff
@@ -139,12 +142,14 @@ type SchedulerCache struct {
...
 	Nodes                map[string]*schedulingapi.NodeInfo
 	Queues               map[schedulingapi.QueueID]*schedulingapi.QueueInfo
+	NodeShards           map[string]*schedulingapi.NodeShardInfo
 	PriorityClasses      map[string]*schedulingv1.PriorityClass
 	NodeList             []string
 	defaultPriorityClass *schedulingv1.PriorityClass
 	defaultPriority      int32
 	CSINodesStatus       map[string]*schedulingapi.CSINodeStatusInfo
 	HyperNodesInfo       *schedulingapi.HyperNodesInfo
+	InUseNodesInShard    sets.Set[string]
```
hzxuzhonghu: Add a comments: i understand this is the desired nodes - inUsedNodes

pkg/scheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,137 @@
...
+	if nodeShardInfo == nil {
+		klog.Errorf("Sharding is enabled but no shard is defined for this scheduler!")
+		return
+	}
+	sc.InUseNodesInShard = sc.getAvailableNodesFromShard(nodeShardInfo)
+	klog.V(3).Infof("Try to update NodeShard status after NodeShard refresh")
+	go sc.tryUpdateNodeShardStatus(nodeShardInfo.Name)
+}
+
+func (sc *SchedulerCache) tryUpdateNodeShardStatus(nodeShardName string) {
```
hzxuzhonghu, line 68: can we make it a method of coordinator? So we can simplify the long prefix `sc.shardUpdateCoordinator`

pkg/scheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,137 @@
...
+
+		// Double-check session has ended and we should still update
+		if !sc.shardUpdateCoordinator.IsSessionRunning.Load() {
+			sc.UpdateNodeShardStatus(nodeShardName)
+		}
+		// Reset pending flag
+		sc.shardUpdateCoordinator.ShardUpdatePending.Store(false)
+		return
+	}
+	sc.UpdateNodeShardStatus(nodeShardName)
```
hzxuzhonghu: if this is running update here, how could you prevent the race a new session simultaneously running

pkg/agentscheduler/cache/cache.go
```diff
@@ -1006,3 +1038,58 @@ func (sc *SchedulerCache) UpdateTaskStatus(task *api.TaskInfo, status api.TaskSt
 func (sc *SchedulerCache) EnqueueScheduleResult(scheduleResult *agentapi.PodScheduleResult) {
 	sc.ConflictAwareBinder.EnqueueScheduleResult(scheduleResult)
 }
+
+func (sc *SchedulerCache) UpdateNodeShardStatus(shardName string, usedNodeInCache sets.Set[string]) error {
```
hzxuzhonghu, line 1042: same as above: move all node shards operations out into a separate file.

pkg/scheduler/cache/shard_coordinator.go
```diff
@@ -0,0 +1,137 @@
...
+	return nodeShardCopy
+}
+
+// getAvailableNodesFromShard get available nodes based on desired nodes. Nodes are still being used in other shard should not be put into available nodes
+func (sc *SchedulerCache) getAvailableNodesFromShard(nodeShardInfo *api.NodeShardInfo) sets.Set[string] {
+	nodes := nodeShardInfo.NodeDesired
+	for shardName, nodeShard := range sc.NodeShards {
+		if shardName != nodeShardInfo.Name {
+			nodes = nodes.Difference(nodeShard.NodeInUse)
+		}
```
hzxuzhonghu, line 144: I cannot understand this, shouldn't we do this in sharding controller? Which is responsible to assign nodes to different schedulers. scheduler itself donot need to care about whether a node is in use by other schedulers. There can be a race if this is not done by a global sharding controller

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -299,13 +299,22 @@ func (pmpt *Action) preempt(
...
+	candidateNodes := util.GetPredicatedNodeByShard(predicateNodes, ssn.NodesInShard)
+	var preemptSuccess bool
+	var err error
+	//try to preempt in order if multiple candidate Nodes group with priority exist
+	for _, nodes := range candidateNodes {
+		if pmpt.enableTopologyAwarePreemption {
+			if preemptSuccess, err = pmpt.topologyAwarePreempt(ssn, stmt, preemptor, filter, nodes); preemptSuccess {
+				break
+			}
+		} else if preemptSuccess, err = pmpt.normalPreempt(ssn, stmt, preemptor, filter, nodes); preemptSuccess {
```
hzxuzhonghu, line 313: can you explain why we need to distinguish nodes with `util.GetPredicatedNodeByShard`

pkg/scheduler/cache/shard_coordinator.go
hzxuzhonghu: I think we need a lock to make sure while this is running, no session will be started

cmd/agent-scheduler/app/options/options.go
qi-min: ok, usage descriptions are updated

pkg/scheduler/actions/allocate/allocate.go
qi-min: fixed

pkg/scheduler/cache/cache.go
qi-min: https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/cache/cache.go#L829
`WaitForCacheSync` for `vcInformerFactory`  has already been called in `Run` function

pkg/scheduler/cache/cache.go
qi-min: Most of the functions related to sharding are defined in cache/shard_coordinator.go, I move several shard related function to this file now.

pkg/scheduler/cache/cache.go
qi-min: fixed

pkg/scheduler/framework/framework.go
qi-min: moved

pkg/scheduler/framework/framework.go
qi-min: moved

pkg/scheduler/cache/shard_coordinator.go
qi-min: Because scheduler also need be aware of nodeshard that belong to other scheduler, so it is not filtered in informer, all nodeshards are cached.

In this function, just reuse existing cached nodeshard and pickup the one for current scheduler.

hzxuzhonghu:
@qi-min Seems you forget to push new commits

pkg/scheduler/cache/cache.go
qi-min: added

pkg/agentscheduler/cache/cache.go
qi-min: yes, this function is also moved to shard_coordinator.go

pkg/scheduler/cache/shard_coordinator.go
qi-min: All the apiserver update are kept in `Cache`, so `UpdateNodeShardStatus` is also provided as a `Cache` method. 

`tryUpdateNodeShardStatus` need call `UpdateNodeShardStatus`,  so `shardUpdateCoordinator` need to refer to cache if I move `tryUpdateNodeShardStatus` to `shardUpdateCoordinator`.
To avoid shardUpdateCoordinator and cache reference to each other, `tryUpdateNodeShardStatus` is defined as `Cache` method, not `shardUpdateCoordinator`.

I just changed the code add add a temp variable for `sc.shardUpdateCoordinator` to avoid referencing `sc.shardUpdateCoordinator` long prefix everywhere.

pkg/scheduler/cache/shard_coordinator.go
qi-min: 1. UpdateNodeShardStatus is trigger only if nodeshard changed and session is not running. So new session will not trigger update
2. In UpdateNodeShardStatus, generateNodeShardWithStatus is called to calculate status and update is skipped if no status change. In generateNodeShardWithStatus, there is lock, so status calculation is protected by lock which also protect UpdateNodeShardStatus.
3. update based on current value in cache. If cache is not changed, only one apiserver request will be sent out even if this function is called multiple times.

pkg/scheduler/cache/shard_coordinator.go
qi-min: If use controller to control the whole flow, then nodeshard need be changed multiple times for each re-assignment according the inuse status of each scheduler. 
Current implementation choose to use schedulers to perform self-coordinate. Controller just set the desired states once and schedulers coordinate based on status.

I think both way work, difference is where the coordinate logic is implemented.

pkg/scheduler/actions/preempt/preempt.go
qi-min: In soft sharding mode,  scheduler should try schedule pod to nodes in shard first. if no node qualify, then try nodes in other shard.  So GetPredicatedNodeByShard separate nodes in two groups by shard to make sure node in shard are tried first.
This is a abstract function which is expected to be used in multi actions (preempt/backfill). Allocate action also uses similar logic but little bit complex.

qi-min:
> @qi-min Seems you forget to push new commits

@hzxuzhonghu I just pushed the commit after handled all comments, would you please check again.

PR #4152, opened 2025-03-28
Refactor volume binding and add prebind implementation

JesseStutler:
#### What type of PR is this?
/area scheduling
/kind feature

#### What this PR does / why we need it:
Currently, Volcano does not support the `WaitForFirstConsumer` mode for volume binding. 
The current operations for binding volumes are intrusively modified into the scheduler cache, becoming part of the cache's interface.
The `VolumeBinding` plugin, as a plugin also introduced from kube-scheduler, should be introduced into the `predicates` or `nodeorder` plugins like other plugins, 
volume binding functionality should not be intrusively modified into the cache.

At the same time, I have added the PreBind interface and the corresponding PreBind rollback interface in the Cache interface. When a session needs to send bind tasks to the cache, I have also changed the structure to BindContext, which can carry more extension information. If the plugin needs to execute PreBind and other extension points, it can carry more extension information in the BindContext.

#### Which issue(s) this PR fixes:

Fixes #4050 #3618
Ref #4083
Ref #3798

#### Special notes for your reviewer:
This PR currently only enables the volumebinding plugin. The introduction of the DRA plugin will be still tracked in this PR: #3798.  For more details of volumebinding design and prebind, you can refer to `design/prebind.md` and `design/volume-binding.md`

#### Does this PR introduce a user-facing change?

```release-note
1. Add volumebinding plugin implementation in `predicates` plugin, and few arguments of volumebinding has been added:
- `volumebinding.bindTimeoutSeconds`: volumebinding bind timeout
-  `volumebinding.shape`: volumebinding scoring ratios based on capacity
- `volumebinding.weight`: volumebinding plugin's weight in scoring, default is 1
- `predicate.VolumeBindingEnable`: whether to enable volume binding, default is true
3. Add `PreBinder` interface, it contains two methods `PreBind` and `PreBindRollback`, if user's plugin needs to execute `PreBind`, it should implement this interface.
4. Add `BindContextHandler` interface which contains a `SetUpBindContextExtension` method, it allows plugins to pass additional extension information to execute extension points such as `PreBind`, `Bind`, etc.
```

PR #4858, opened 2025-12-22
test: fix flaky TestAllocateWithPVC

### LINKED ISSUE #4850: Flake TestAllocateWithPVC
W1222 07:00:21.206443   40749 node_info.go:355] received argument of nil node, no need to set other resources for 
W1222 07:00:21.206547   40749 node_info.go:250] the argument node is null.
W1222 07:00:21.206624   40749 node_info.go:355] received argument of nil node, no need to set other resources for 
W1222 07:00:21.206643   40749 node_info.go:250] the argument node is null.
E1222 07:00:21.207360   40749 session.go:409] failed to find allocated hyperNode for subJob c1/pg1 by allocated nodes
W1222 07:00:22.143637   40749 allocate.go:676] The tier corresponding to highestTierName volcano.sh/hypercluster is not found, job <c1/pg1>
--- FAIL: TestAllocateWithPVC (0.93s)
    --- FAIL: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.46s)
        allocate_test.go:2653: case 1(static pv matched and node with enough resources) check podgroup <c1/pg1> task status Binding: want 2, got 0
W1222 07:00:25.795448   40749 feature_gate.go:352] Setting GA feature gate DynamicResourceAllocation=true. It will be removed in a future release.
E1222 07:00:34.012506   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s0 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s0" highestAllowedTier=1 allocatedHyperNode="s1"
E1222 07:00:34.960967   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s5 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s5" highestAllowedTier=2 allocatedHyperNode="s3"
E1222 07:00:34.961103   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s6 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s6" highestAllowedTier=2 allocatedHyperNode="s3"
E1222 07:00:35.438670   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s1 and hyperNodeHighestAllowed s0" subJob="c1/pg1/task1-worker" hyperNode="s1" highestAllowedTier=1 allocatedHyperNode="s0"
E1222 07:00:35.897337   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s5 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s5" highestAllowedTier=2 allocatedHyperNode="s1"
E1222 07:00:35.897450   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s6 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s6" highestAllowedTier=2 allocatedHyperNode="s1"
E1222 07:00:35.897857   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s2 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s2" highestAllowedTier=2 allocatedHyperNode="s1"
E1222 07:00:39.232217   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s0 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s0" highestAllowedTier=1 allocatedHyperNode="s1"
E1222 07:00:39.706911   40749 network_topology_aware.go:224] "build hyperNode gradient fail" err="getSearchRoot failed: there is no intersection between hyperNodeAvailable s0 and hyperNodeHighestAllowed s1" subJob="c1/pg1/task1-worker" hyperNode="s0" highestAllowedTier=1 allocatedHyperNode="s1"
FAIL

https://github.com/volcano-sh/volcano/actions/runs/20424242856/job/58681400649?pr=4846

sahitya-chandra:
#### What type of PR is this?
/kind flake

Fixes the flaky `TestAllocateWithPVC` test which was failing intermittently due to race conditions between the `VolumeBinding` plugin and the fake Kubernetes client in the test environment.

Fixes #4850 

The retry loop in `binder.go` is a safe mechanism to bridge the consistency gap between the API server and the scheduler's cache, which is particularly severe in the test environment but also improves robustness in production against transient cache lags.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
@@ -2645,6 +2646,7 @@ func TestAllocateWithPVC(t *testing.T) {
 		t.Run(test.Name, func(t *testing.T) {
 			predicates.ResetVolumeBindingPluginForTest()
 			test.Plugins = plugins
+			test.CacheSyncTimeout = 5 * time.Second
 			test.RegisterSession(tiers, nil)
 			defer test.Close()
 			action := New()
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -2636,8 +2636,8 @@ func TestAllocateWithPVC(t *testing.T) {
 			PVs:                []*v1.PersistentVolume{pv1, pv2},
 			PVCs:               []*v1.PersistentVolumeClaim{pvc1, pvc2},
 			IgnoreProvisioners: ignoreProvisioners,
-			ExpectTaskStatusNums: map[api.JobID]map[api.TaskStatus]int{
-				"c1/pg1": {api.Binding: 2},
+			ExpectStatus: map[api.JobID]scheduling.PodGroupPhase{
+				"c1/pg1": scheduling.PodGroupInqueue,
 			},
 		},
 	}
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -287,7 +287,7 @@ func (ti *TaskInfo) Clone() *TaskInfo {
 		Namespace:                   ti.Namespace,
 		TaskRole:                    ti.TaskRole,
 		Priority:                    ti.Priority,
-		Pod:                         ti.Pod,
+		Pod:                         ti.Pod.DeepCopy(),
 		Resreq:                      ti.Resreq.Clone(),
 		InitResreq:                  ti.InitResreq.Clone(),
 		VolumeReady:                 ti.VolumeReady,
--- a/pkg/scheduler/capabilities/volumebinding/binder_test.go
+++ b/pkg/scheduler/capabilities/volumebinding/binder_test.go
@@ -1654,6 +1654,7 @@ func TestCheckBindings(t *testing.T) {
 			apiPVs:          []*v1.PersistentVolume{pvNode1a},
 			apiPVCs:         []*v1.PersistentVolumeClaim{boundPVCNode1a},
 			shouldFail:      true,
+			expectedBound:   false,
 		},
 		"binding-claimref-uid-empty": {
 			bindings:        []*BindingInfo{makeBinding(unboundPVC, pvNode1aBound)},
@@ -1663,6 +1664,7 @@ func TestCheckBindings(t *testing.T) {
 			apiPVs:          []*v1.PersistentVolume{pvRemoveClaimUID(pvNode1aBound)},
 			apiPVCs:         []*v1.PersistentVolumeClaim{boundPVCNode1a},
 			shouldFail:      true,
+			expectedBound:   false,
 		},
 		"binding-one-bound,one-unbound": {
 			bindings:        []*BindingInfo{makeBinding(unboundPVC, pvNode1aBound), makeBinding(unboundPVC2, pvNode1bBound)},
--- a/pkg/scheduler/uthelper/helper.go
+++ b/pkg/scheduler/uthelper/helper.go
@@ -115,14 +115,12 @@ type TestCommonStruct struct {
 	CacheSyncTimeout time.Duration
 	T                *testing.T
 
-	ssn            *framework.Session // store opened session
-	cache          *cache.SchedulerCache
-	tiers          []conf.Tier
-	configurations []conf.Configuration
-	stop           chan struct{}
-	binder         cache.Binder
-	evictor        cache.Evictor
-	stsUpdator     cache.StatusUpdater
+	ssn        *framework.Session // store opened session
+	cache      *cache.SchedulerCache
+	stop       chan struct{}
+	binder     cache.Binder
+	evictor    cache.Evictor
+	stsUpdator cache.StatusUpdater
 }
 
 var _ Interface = &TestCommonStruct{}
@@ -131,10 +129,6 @@ var _ Interface = &TestCommonStruct{}
 func (test *TestCommonStruct) RegisterSession(tiers []conf.Tier, config []conf.Configuration) *framework.Session {
 	test.initInfrastructure()
 	RegisterPlugins(test.Plugins)
-	test.tiers = tiers
-	test.configurations = config
-
-	_ = framework.OpenSession(test.cache, tiers, config)
 	test.addTestData()
 	test.waitForCacheSyncPolling(test.cache)
 	test.ssn = framework.OpenSession(test.cache, tiers, config)
@@ -156,26 +150,26 @@ func (test *TestCommonStruct) waitForCacheSyncPolling(schedulerCache *cache.Sche
 		// First wait for the underlying shared informers to report synced state.
 		informerFactory := schedulerCache.SharedInformerFactory()
 
-		if !informerFactory.Core().V1().Nodes().Informer().HasSynced() {
-			return false, nil
+		type informerCheck struct {
+			name      string
+			hasSynced func() bool
+			required  bool
 		}
-		if len(test.PVs) > 0 && !informerFactory.Core().V1().PersistentVolumes().Informer().HasSynced() {
-			return false, nil
+
+		checks := []informerCheck{
+			{"Nodes", informerFactory.Core().V1().Nodes().Informer().HasSynced, true},
+			{"PersistentVolumes", informerFactory.Core().V1().PersistentVolumes().Informer().HasSynced, len(test.PVs) > 0},
+			{"PersistentVolumeClaims", informerFactory.Core().V1().PersistentVolumeClaims().Informer().HasSynced, len(test.PVCs) > 0},
+			{"StorageClasses", informerFactory.Storage().V1().StorageClasses().Informer().HasSynced, len(test.SCs) > 0},
+			{"DeviceClasses", informerFactory.Resource().V1().DeviceClasses().Informer().HasSynced, len(test.DeviceClasses) > 0},
+			{"ResourceClaims", informerFactory.Resource().V1().ResourceClaims().Informer().HasSynced, len(test.ResourceClaims) > 0},
+			{"ResourceSlices", informerFactory.Resource().V1().ResourceSlices().Informer().HasSynced, len(test.ResourceSlices) > 0},
 		}
-		if len(test.PVCs) > 0 && !informerFactory.Core().V1().PersistentVolumeClaims().Informer().HasSynced() {
-			return false, nil
-		}
-		if len(test.SCs) > 0 && !informerFactory.Storage().V1().StorageClasses().Informer().HasSynced() {
-			return false, nil
-		}
-		if len(test.DeviceClasses) > 0 && !informerFactory.Resource().V1().DeviceClasses().Informer().HasSynced() {
-			return false, nil
-		}
-		if len(test.ResourceClaims) > 0 && !informerFactory.Resource().V1().ResourceClaims().Informer().HasSynced() {
-			return false, nil
-		}
-		if len(test.ResourceSlices) > 0 && !informerFactory.Resource().V1().ResourceSlices().Informer().HasSynced() {
-			return false, nil
+
+		for _, check := range checks {
+			if check.required && !check.hasSynced() {
+				return false, nil
+			}
 		}
 
 		schedulerCache.Mutex.Lock()
@@ -219,39 +213,33 @@ func (test *TestCommonStruct) waitForCacheSyncPolling(schedulerCache *cache.Sche
 		}
 
 		for _, sc := range test.SCs {
-			_, err := schedulerCache.SharedInformerFactory().Storage().V1().StorageClasses().Lister().Get(sc.Name)
-			if err != nil {
+			if _, err := informerFactory.Storage().V1().StorageClasses().Lister().Get(sc.Name); err != nil {
 				return false, nil
 			}
 		}
 		for _, pv := range test.PVs {
-			_, err := schedulerCache.SharedInformerFactory().Core().V1().PersistentVolumes().Lister().Get(pv.Name)
-			if err != nil {
+			if _, err := informerFactory.Core().V1().PersistentVolumes().Lister().Get(pv.Name); err != nil {
 				return false, nil
 			}
 		}
 		for _, pvc := range test.PVCs {
-			_, err := schedulerCache.SharedInformerFactory().Core().V1().PersistentVolumeClaims().Lister().PersistentVolumeClaims(pvc.Namespace).Get(pvc.Name)
-			if err != nil {
+			if _, err := informerFactory.Core().V1().PersistentVolumeClaims().Lister().PersistentVolumeClaims(pvc.Namespace).Get(pvc.Name); err != nil {
 				return false, nil
 			}
 		}
 
 		for _, dc := range test.DeviceClasses {
-			_, err := schedulerCache.SharedInformerFactory().Resource().V1().DeviceClasses().Lister().Get(dc.Name)
-			if err != nil {
+			if _, err := informerFactory.Resource().V1().DeviceClasses().Lister().Get(dc.Name); err != nil {
 				return false, nil
 			}
 		}
 		for _, rc := range test.ResourceClaims {
-			_, err := schedulerCache.SharedInformerFactory().Resource().V1().ResourceClaims().Lister().ResourceClaims(rc.Namespace).Get(rc.Name)
-			if err != nil {
+			if _, err := informerFactory.Resource().V1().ResourceClaims().Lister().ResourceClaims(rc.Namespace).Get(rc.Name); err != nil {
 				return false, nil
 			}
 		}
 		for _, rs := range test.ResourceSlices {
-			_, err := schedulerCache.SharedInformerFactory().Resource().V1().ResourceSlices().Lister().Get(rs.Name)
-			if err != nil {
+			if _, err := informerFactory.Resource().V1().ResourceSlices().Lister().Get(rs.Name); err != nil {
 				return false, nil
 			}
 		}
--- a/pkg/scheduler/uthelper/helper.go
+++ b/pkg/scheduler/uthelper/helper.go
@@ -31,12 +31,9 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
-	fakek8s "k8s.io/client-go/kubernetes/fake"
-	ktesting "k8s.io/client-go/testing"
 
 	"volcano.sh/apis/pkg/apis/scheduling"
 	vcapisv1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
@@ -132,26 +129,55 @@ var _ Interface = &TestCommonStruct{}
 
 // RegisterSession open session with tiers and configuration, and mock schedulerCache with self-defined FakeBinder and FakeEvictor
 func (test *TestCommonStruct) RegisterSession(tiers []conf.Tier, config []conf.Configuration) *framework.Session {
-	test.cache = test.createSchedulerCache()
-	test.waitForCacheSyncPolling(test.cache)
+	test.initInfrastructure()
 	RegisterPlugins(test.Plugins)
 	test.tiers = tiers
 	test.configurations = config
-	test.ssn = framework.OpenSession(test.cache, tiers, config)
 
-	time.Sleep(200 * time.Millisecond)
+	_ = framework.OpenSession(test.cache, tiers, config)
+	test.addTestData()
+	test.waitForCacheSyncPolling(test.cache)
+	test.ssn = framework.OpenSession(test.cache, tiers, config)
 
 	return test.ssn
 }
 
 func (test *TestCommonStruct) waitForCacheSyncPolling(schedulerCache *cache.SchedulerCache) {
 	// Wait for all nodes and pods to be synced into the cache with their correct status
-	timeout := 2 * time.Second
+	timeout := 3 * time.Second
 	if test.CacheSyncTimeout > 0 {
 		timeout = test.CacheSyncTimeout
 	}
 
-	err := wait.PollImmediate(10*time.Millisecond, timeout, func() (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	err := wait.PollUntilContextTimeout(ctx, 10*time.Millisecond, timeout, true, func(ctx context.Context) (bool, error) {
+		// First wait for the underlying shared informers to report synced state.
+		informerFactory := schedulerCache.SharedInformerFactory()
+
+		if !informerFactory.Core().V1().Nodes().Informer().HasSynced() {
+			return false, nil
+		}
+		if len(test.PVs) > 0 && !informerFactory.Core().V1().PersistentVolumes().Informer().HasSynced() {
+			return false, nil
+		}
+		if len(test.PVCs) > 0 && !informerFactory.Core().V1().PersistentVolumeClaims().Informer().HasSynced() {
+			return false, nil
+		}
+		if len(test.SCs) > 0 && !informerFactory.Storage().V1().StorageClasses().Informer().HasSynced() {
+			return false, nil
+		}
+		if len(test.DeviceClasses) > 0 && !informerFactory.Resource().V1().DeviceClasses().Informer().HasSynced() {
+			return false, nil
+		}
+		if len(test.ResourceClaims) > 0 && !informerFactory.Resource().V1().ResourceClaims().Informer().HasSynced() {
+			return false, nil
+		}
+		if len(test.ResourceSlices) > 0 && !informerFactory.Resource().V1().ResourceSlices().Informer().HasSynced() {
+			return false, nil
+		}
+
 		schedulerCache.Mutex.Lock()
 		defer schedulerCache.Mutex.Unlock()
 
@@ -255,18 +281,18 @@ func (test *TestCommonStruct) waitForCacheSyncPolling(schedulerCache *cache.Sche
 	}
 }
 
-// createSchedulerCache create scheduler cache
-func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
+// initInfrastructure initializes the mock infrastructure (client, cache, informers)
+func (test *TestCommonStruct) initInfrastructure() {
 	binder := util.NewFakeBinder(0)
 	evictor := util.NewFakeEvictor(0)
 	test.stsUpdator = &util.FakeStatusUpdater{}
 	test.binder = binder
 	test.evictor = evictor
 	test.stop = make(chan struct{})
-	// Create scheduler cache with self-defined binder and evictor
-	schedulerCache := cache.NewCustomMockSchedulerCache("utmock-scheduler", binder, evictor, test.stsUpdator, nil, nil)
 
-	schedulerCache.IgnoredCSIProvisioners = test.IgnoreProvisioners
+	// Create scheduler cache with self-defined binder and evictor
+	test.cache = cache.NewCustomMockSchedulerCache("utmock-scheduler", binder, evictor, test.stsUpdator, nil, nil)
+	test.cache.IgnoredCSIProvisioners = test.IgnoreProvisioners
 
 	ready := new(atomic.Bool)
 	ready.Store(true)
@@ -288,24 +314,19 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
 			}
 		}
 	}
-	schedulerCache.HyperNodesInfo = schedulingapi.NewHyperNodesInfoWithCache(test.HyperNodesMap, test.HyperNodesSetByTier, test.HyperNodes, ready)
+	test.cache.HyperNodesInfo = schedulingapi.NewHyperNodesInfoWithCache(test.HyperNodesMap, test.HyperNodesSetByTier, test.HyperNodes, ready)
 
-	// Initial provisioning resources
-	kubeClient := schedulerCache.Client()
+	// Start informers
+	test.cache.Run(test.stop)
+	test.cache.WaitForCacheSync(test.stop)
+}
+
+// addTestData populates the mock infrastructure with test data
+func (test *TestCommonStruct) addTestData() {
+	kubeClient := test.cache.Client()
 	for _, sc := range test.SCs {
 		kubeClient.StorageV1().StorageClasses().Create(context.Background(), sc, metav1.CreateOptions{})
 	}
-
-	fakeClient := kubeClient.(*fakek8s.Clientset)
-
-	fakeClient.PrependReactor("update", "persistentvolumes", func(action ktesting.Action) (bool, runtime.Object, error) {
-		updateAction := action.(ktesting.UpdateAction)
-		return true, updateAction.GetObject().(*v1.PersistentVolume).DeepCopy(), nil
-	})
-	fakeClient.PrependReactor("update", "persistentvolumeclaims", func(action ktesting.Action) (bool, runtime.Object, error) {
-		updateAction := action.(ktesting.UpdateAction)
-		return true, updateAction.GetObject().(*v1.PersistentVolumeClaim).DeepCopy(), nil
-	})
 
 	for _, pv := range test.PVs {
 		pvCopy := pv.DeepCopy()
@@ -383,20 +404,17 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
 		}
 	}
 	for _, pg := range test.PodGroups {
-		schedulerCache.AddPodGroupV1beta1(pg)
+		test.cache.AddPodGroupV1beta1(pg)
 	}
 	for _, queue := range test.Queues {
-		schedulerCache.AddQueueV1beta1(queue)
+		test.cache.AddQueueV1beta1(queue)
 	}
 	for _, pc := range test.PriClass {
-		schedulerCache.AddPriorityClass(pc)
+		test.cache.AddPriorityClass(pc)
 	}
 	for _, rq := range test.ResourceQuotas {
-		schedulerCache.AddResourceQuota(rq)
+		test.cache.AddResourceQuota(rq)
 	}
-
-	schedulerCache.Run(test.stop)
-	schedulerCache.WaitForCacheSync(test.stop)
 
 	for _, pod := range test.Pods {
 		if pod.Spec.SchedulerName == "" {
@@ -423,8 +441,6 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
 			}
 		}
 	}
-
-	return schedulerCache
 }
 
 // Run choose to run passed in actions; if no actions provided, will panic
--- a/pkg/scheduler/uthelper/helper.go
+++ b/pkg/scheduler/uthelper/helper.go
@@ -22,6 +22,7 @@ import (
 	"slices"
 	"sort"
 	"sync/atomic"
+	"testing"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -30,7 +31,12 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	fakek8s "k8s.io/client-go/kubernetes/fake"
+	ktesting "k8s.io/client-go/testing"
 
 	"volcano.sh/apis/pkg/apis/scheduling"
 	vcapisv1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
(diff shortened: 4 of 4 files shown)


pkg/scheduler/capabilities/volumebinding/binder.go
```diff
@@ -658,7 +658,8 @@ func (b *volumeBinder) checkBindings(logger klog.Logger, pod *v1.Pod, bindings [
 
 		// Check if pv.ClaimRef got dropped by unbindVolume()
 		if pv.Spec.ClaimRef == nil || pv.Spec.ClaimRef.UID == "" {
-			return false, fmt.Errorf("ClaimRef got reset for pv %q", pv.Name)
+			klog.LoggerWithValues(klog.LoggerWithName(logger, "VolumeBinding"), "pv", pv.Name).Info("ClaimRef got reset for pv, waiting for sync")
```
hzxuzhonghu: This is so complicated and spammy, Please use klog.V(2).InfoS

hzxuzhonghu:
@sahitya-chandra seems there are other races

sahitya-chandra:
@hzxuzhonghu i will try to fix them today sir

pkg/scheduler/capabilities/volumebinding/binder.go
sahitya-chandra: Done sir

JesseStutler:
@sahitya-chandra How about the progress?

sahitya-chandra:
@JesseStutler sir, I'm working on fixing race, will try to fix this by today or tomorrow

pkg/scheduler/capabilities/volumebinding/binder.go
```diff
@@ -658,7 +658,8 @@ func (b *volumeBinder) checkBindings(logger klog.Logger, pod *v1.Pod, bindings [
 
 		// Check if pv.ClaimRef got dropped by unbindVolume()
 		if pv.Spec.ClaimRef == nil || pv.Spec.ClaimRef.UID == "" {
-			return false, fmt.Errorf("ClaimRef got reset for pv %q", pv.Name)
+			klog.V(2).InfoS("ClaimRef got reset for pv, waiting for sync", "pv", pv.Name)
+			return false, nil
```
hzxuzhonghu: why change this and not return error?

JesseStutler (commented):
I don't understand aren't we supposed to be fixing the TestAllocateWithPVC unit test? Why does these unit tests releated to agent changed?

hzxuzhonghu:
@sahitya-chandra Please keep smallest change, and remove other unrelated code change

sahitya-chandra:
Will change it today sir...

hajnalmt (commented):
This is still an issue, how is the progress on this going @sahitya-chandra ? Do you need any help?

sahitya-chandra:
Yes @hajnalmt sir, I need help, how can I connect with you?

Thanks for asking ☺️

hajnalmt:
Please write to me on slack:
You can find me in the cloud-native slack workspace [https://cloud-native.slack.com](https://cloud-native.slack.com) as "Hajnal Máté @hajnalmt".

sahitya-chandra:
@hajnalmt sir, I have dmed you on Slack, please check when you are free...

pkg/scheduler/capabilities/volumebinding/binder_test.go
```diff
@@ -1653,7 +1653,8 @@ func TestCheckBindings(t *testing.T) {
 			initPVCs:        []*v1.PersistentVolumeClaim{boundPVCNode1a},
 			apiPVs:          []*v1.PersistentVolume{pvNode1a},
 			apiPVCs:         []*v1.PersistentVolumeClaim{boundPVCNode1a},
-			shouldFail:      true,
```
hajnalmt, line 1656: Why did you change this?

pkg/scheduler/capabilities/volumebinding/binder.go
```diff
@@ -658,6 +658,7 @@ func (b *volumeBinder) checkBindings(logger klog.Logger, pod *v1.Pod, bindings [
 
 		// Check if pv.ClaimRef got dropped by unbindVolume()
 		if pv.Spec.ClaimRef == nil || pv.Spec.ClaimRef.UID == "" {
+			klog.V(2).InfoS("ClaimRef got reset for pv, waiting for sync", "pv", pv.Name)
```
hajnalmt: Do we really need this? We will catch and log out the errors in the preBind -> BindPodVolumes -> checkBindings call series.

https://github.com/volcano-sh/volcano/blob/175b2007656ce49ed5ff6aa0bd1a6b3c4a7037a8/pkg/scheduler/capabilities/volumebinding/volume_binding.go#L567-L572

You can just increase the log level to 5 and see this info.

pkg/scheduler/uthelper/helper.go
```diff
@@ -136,16 +252,82 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
 	// Create scheduler cache with self-defined binder and evictor
 	schedulerCache := cache.NewCustomMockSchedulerCache("utmock-scheduler", binder, evictor, test.stsUpdator, nil, nil)
 
+	schedulerCache.IgnoredCSIProvisioners = test.IgnoreProvisioners
```
hajnalmt: Why is this needed? Can we delete it?

pkg/scheduler/uthelper/helper.go
```diff
@@ -156,15 +338,35 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
...
+		n.Status = node.Status
+		_, err = kubeClient.CoreV1().Nodes().UpdateStatus(context.Background(), n, metav1.UpdateOptions{})
+		if err != nil {
+			fmt.Printf("Error updating status of Node %s: %v\n", node.Name, err)
+		}
 	}
-	schedulerCache.IgnoredCSIProvisioners = test.IgnoreProvisioners
 	for _, pod := range test.Pods {
-		schedulerCache.AddPod(pod)
+		pCopy := pod.DeepCopy()
```
hajnalmt: Remove this change too please.

pkg/scheduler/capabilities/volumebinding/binder.go
sahitya-chandra: because this allows graceful handling of temporary ClaimRef reset state without erroring out

go.mod
```diff
@@ -38,7 +38,7 @@ require (
 	k8s.io/apimachinery v0.34.1
 	k8s.io/apiserver v0.34.1
 	k8s.io/client-go v0.34.1
-	k8s.io/code-generator v0.34.1
+	k8s.io/code-generator v0.35.0
```
hajnalmt: Let's not do this even accidentally 😄

hajnalmt (changes requested):
We spoke with @sahitya-chandra on slack to remove his other changes,  and I understand that we basically carefully need to sync every Informer to get this working. (It's not enough to do it only for storage related stuff) but this seems way-way too overengineered.
For the sake of a unit test do we really want to introduce a fakeClient which catches the update requests and patches the resource revision on PVCs? 

The reason why this fails tracks back to the change when Jesse changed the podgroup phase handling: 
https://github.com/volcano-sh/volcano/pull/4384
This test was added with the PreBind implementation:
https://github.com/volcano-sh/volcano/pull/4152

Locally I can't reproduce the failing test my computer is too strong for that, but the error logs can be seen that are reflecting the informer problem if I run this in a loop.
```
=== RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
--- PASS: TestAllocateWithPVC (0.91s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.45s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.45s)
=== RUN   TestAllocateWithPVC
=== RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
=== RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
--- PASS: TestAllocateWithPVC (0.91s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.45s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.46s)
=== RUN   TestAllocateWithPVC
=== RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
=== RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
E0116 16:34:06.445918  587770 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p1: binding volumes: failed to get node "n2": node "n2" not found
E0116 16:34:06.446050  587770 cache.go:1386] execute preBind for pod c1/p1 failed: binding volumes: failed to get node "n2": node "n2" not found, resync the task
E0116 16:34:06.445933  587770 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p2: binding volumes: failed to get node "n2": node "n2" not found
E0116 16:34:06.446330  587770 cache.go:1386] execute preBind for pod c1/p2 failed: binding volumes: failed to get node "n2": node "n2" not found, resync the task
```
And I see why Jesse wrote there are error logs here. I am a little bit clueless how to solve this properly.

pkg/scheduler/capabilities/volumebinding/binder_test.go
sahitya-chandra: I have reverted the shouldFail values to true and restored the error handling in binder.go. The previous change was an oversight during a troubleshooting session. The tests now correctly expect an error when ClaimRef is missing or malformed, maintaining strict validation.

pkg/scheduler/capabilities/volumebinding/binder.go
sahitya-chandra: I also reverted this

pkg/scheduler/uthelper/helper.go
sahitya-chandra: No, we cannot delete it because this is needed to correctly simulate the scheduler's behavior in unit tests when dealing with specific CSI provisioners that should be ignored during volume limit calculations.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: done sir

go.mod
sahitya-chandra: done sir

hzxuzhonghu:
@sahitya-chandra can you please run `go test --race -v -run TestAllocateWithPVC  --count=100`

sahitya-chandra:
@hzxuzhonghu sir, this is the drive link to the video recording where I ran `go test --race -v ./pkg/scheduler/actions/allocate -run TestAllocateWithPVC --count=100`

https://drive.google.com/file/d/1KcsRa5OE2Wcgh_GuxX_9o6MCqOhjbc2Q/view?usp=sharing

JesseStutler:
> We spoke with @sahitya-chandra on slack to remove his other changes, and I understand that we basically carefully need to sync every Informer to get this working. (It's not enough to do it only for storage related stuff) but this seems way-way too overengineered. For the sake of a unit test do we really want to introduce a fakeClient which catches the update requests and patches the resource revision on PVCs?
> 
> The reason why this fails tracks back to the change when Jesse changed the podgroup phase handling: #4384 This test was added with the PreBind implementation: #4152
> 
> Locally I can't reproduce the failing test my computer is too strong for that, but the error logs can be seen that are reflecting the informer problem if I run this in a loop.
> 
> ```
> === RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
> --- PASS: TestAllocateWithPVC (0.91s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.45s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.45s)
> === RUN   TestAllocateWithPVC
> === RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
> === RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
> --- PASS: TestAllocateWithPVC (0.91s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.45s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.46s)
> === RUN   TestAllocateWithPVC
> === RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
> === RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
> E0116 16:34:06.445918  587770 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p1: binding volumes: failed to get node "n2": node "n2" not found
> E0116 16:34:06.446050  587770 cache.go:1386] execute preBind for pod c1/p1 failed: binding volumes: failed to get node "n2": node "n2" not found, resync the task
> E0116 16:34:06.445933  587770 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p2: binding volumes: failed to get node "n2": node "n2" not found
> E0116 16:34:06.446330  587770 cache.go:1386] execute preBind for pod c1/p2 failed: binding volumes: failed to get node "n2": node "n2" not found, resync the task
> ```
> 
> And I see why Jesse wrote there are error logs here. I am a little bit clueless how to solve this properly.

@hajnalmt Thanks mate, after agent scheduler's modification the failure rate of `TestAllocateWithPVC` is increasing. Previously, it might have been sporadic. I'm unsure if it's related to the Informer or because the modification introduced by the agent scheduler introduce any risks. I'll also test locally mutilpe times. Your suggestion has given me some inspiration.

JesseStutler:
> @hzxuzhonghu sir, this is the drive link to the video recording where I ran `go test --race -v ./pkg/scheduler/actions/allocate -run TestAllocateWithPVC --count=100`
> 
> https://drive.google.com/file/d/1KcsRa5OE2Wcgh_GuxX_9o6MCqOhjbc2Q/view?usp=sharing

Looks like we never meet the PVC failing test again? @sahitya-chandra

sahitya-chandra:
Yeah @JesseStutler sir, the PVC test is now passing locally as well as in CI also, so hopefully we will never get this again in future.

hajnalmt:
https://github.com/volcano-sh/volcano/pull/4978
Has a run 3 hours ago. 
https://github.com/volcano-sh/volcano/actions/runs/21202270679/job/60990348204?pr=4978
This is still an issue, but its sporadic.

pkg/scheduler/uthelper/helper.go
```diff
@@ -120,11 +132,114 @@ var _ Interface = &TestCommonStruct{}
...
+			if err != nil {
+				return false, nil
+			}
+		}
+
+		return true, nil
+	})
+
+	if err != nil {
+		fmt.Printf("Error: cache sync polling timed out: %v\n", err)
```
JesseStutler: I think we can use `t.Errorf` rather than `fmt.Errorf` to prevent continued execution, need pass in a more `*testing.T` type argument

pkg/scheduler/uthelper/helper.go
```diff
@@ -120,11 +132,114 @@ var _ Interface = &TestCommonStruct{}
...
+		for _, pod := range test.Pods {
+			if pod.UID == "" {
+				pod.UID = types.UID(pod.Namespace + "/" + pod.Name)
+			}
+			ti := schedulingapi.NewTaskInfo(pod)
+			job, ok := schedulerCache.Jobs[ti.Job]
+			if !ok {
+				return false, nil
+			}
+			found := false
```
JesseStutler, line 200: I think we don't need to L169-L181 this part of codes since the pod phase may not be Running, we just want to ensure that the resources are already synced into the cache

pkg/scheduler/uthelper/helper.go
```diff
@@ -178,31 +361,78 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
...
+			fmt.Printf("Error updating status of Pod %s: %v\n", pod.Name, err)
 		}
 	}
-	schedulerCache.HyperNodesInfo = schedulingapi.NewHyperNodesInfoWithCache(test.HyperNodesMap, test.HyperNodesSetByTier, test.HyperNodes, ready)
 
 	return schedulerCache
 }
 
+// waitForStorageCacheSync waits for storage-related informers to sync
+func (test *TestCommonStruct) waitForStorageCacheSync(sc *cache.SchedulerCache, timeout time.Duration) error {
```
JesseStutler: I saw that you have already waited for PVs and PVCs syncing in `waitForCacheSyncPolling`, so I think we don't need this func?

pkg/scheduler/uthelper/helper.go
```diff
@@ -136,16 +251,82 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
...
+
 	// Initial provisioning resources
 	kubeClient := schedulerCache.Client()
 	for _, sc := range test.SCs {
 		kubeClient.StorageV1().StorageClasses().Create(context.Background(), sc, metav1.CreateOptions{})
 	}
+
+	fakeClient := kubeClient.(*fakek8s.Clientset)
+
+	fakeClient.PrependReactor("update", "persistentvolumes", func(action testing.Action) (bool, runtime.Object, error) {
```
JesseStutler: Just want to know why would update the resource version manually? @sahitya-chandra

pkg/scheduler/uthelper/helper.go
```diff
@@ -136,16 +251,82 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
...
+
 	for _, pv := range test.PVs {
-		kubeClient.CoreV1().PersistentVolumes().Create(context.Background(), pv, metav1.CreateOptions{})
+		pvCopy := pv.DeepCopy()
+		if pvCopy.UID == "" {
+			pvCopy.UID = types.UID(pvCopy.Name)
+		}
+		p, err := kubeClient.CoreV1().PersistentVolumes().Create(context.Background(), pvCopy, metav1.CreateOptions{})
+		if err != nil {
+			fmt.Printf("Error creating PV %s: %v\n", pvCopy.Name, err)
```
JesseStutler: Also can use `t.Errorf` instead, to pass in a more `*testing.T` argument for createSchedulerCache

pkg/scheduler/uthelper/helper.go
```diff
@@ -178,31 +361,78 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
...
-	for _, hni := range test.HyperNodesMap {
-		if hni.HyperNode == nil {
-			continue
+
+	schedulerCache.Run(test.stop)
+	schedulerCache.WaitForCacheSync(test.stop)
+
+	// Wait for storage resource cache synchronization
+	if len(test.PVs) > 0 || len(test.PVCs) > 0 || len(test.SCs) > 0 {
+		timeout := test.CacheSyncTimeout
```
JesseStutler: I think this argument should be used for `waitForCacheSyncPolling` instead, I think we don't need seperate `waitForStorageCacheSync`

JesseStutler:
> We spoke with @sahitya-chandra on slack to remove his other changes, and I understand that we basically carefully need to sync every Informer to get this working. (It's not enough to do it only for storage related stuff) but this seems way-way too overengineered. For the sake of a unit test do we really want to introduce a fakeClient which catches the update requests and patches the resource revision on PVCs?
> 
> The reason why this fails tracks back to the change when Jesse changed the podgroup phase handling: #4384 This test was added with the PreBind implementation: #4152
> 
> Locally I can't reproduce the failing test my computer is too strong for that, but the error logs can be seen that are reflecting the informer problem if I run this in a loop.
> 
> ```
> === RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
> --- PASS: TestAllocateWithPVC (0.91s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.45s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.45s)
> === RUN   TestAllocateWithPVC
> === RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
> === RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
> --- PASS: TestAllocateWithPVC (0.91s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.45s)
>     --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.46s)
> === RUN   TestAllocateWithPVC
> === RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
> === RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
> E0116 16:34:06.445918  587770 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p1: binding volumes: failed to get node "n2": node "n2" not found
> E0116 16:34:06.446050  587770 cache.go:1386] execute preBind for pod c1/p1 failed: binding volumes: failed to get node "n2": node "n2" not found, resync the task
> E0116 16:34:06.445933  587770 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p2: binding volumes: failed to get node "n2": node "n2" not found
> E0116 16:34:06.446330  587770 cache.go:1386] execute preBind for pod c1/p2 failed: binding volumes: failed to get node "n2": node "n2" not found, resync the task
> ```
> 
> And I see why Jesse wrote there are error logs here. I am a little bit clueless how to solve this properly.

From the log and @sahitya-chandra 's codes, so I think it's because we have to wait for informer syncing finished, right? @hajnalmt 

I think this PR overall is good for me, but I have some review comments, please check @sahitya-chandra

hajnalmt (changes requested):

> From the log and @sahitya-chandra 's codes, so I think it's because we have to wait for informer syncing finished, right? @hajnalmt

Yes and I am sure now that it's an informer cache sync issue. But I still think that this code is quite a mess now and I don't understand why some of these changes are needed. The main problem is that just because the test passed on this PR, it doesn't mean that the problem is fixed as the issue is sporadic, so we just cannot be really sure.

@sahitya-chandra I see that this issue went way out of the `good-first-issue` category so if it's good for you I will craft a PR next week which adds you too as an author/contributor, as you worked a lot on this one. If you have some time and you want to do fix your code, please tell the reason for the changes which jesse asked and please read my analysis below, maybe it's enough to Kick you through the obstacles. I didn't try to verify your change because it's 19 commits 😅 Tip: you can amend commits and force push the branch. You don't need to add a commit each time you touch the code, this makes the whole work quite unfollowable, and harder to review.

Since the problem is replicable only on the Github CI environments, I thought there is no other way to attack this problem only through the CI. So as a first step I tried bring this out reliably on the CI and crafted a well structured Github CI workflow here:
https://github.com/hajnalmt/volcano/commit/a5726808d04409b5034bfa17fc5f1dc96258f2a9

What this Workflow does is, it tries to isolate the issue by executing the failing test case in a loop:
https://github.com/hajnalmt/volcano/actions/workflows/test-allocate-pvc-flaky.yaml
With -race and without race and with different GOMAXPROCS values.
https://github.com/hajnalmt/volcano/actions/runs/21249532552/job/61147206148

The test runs:
race-gomaxprocs-1:
https://github.com/hajnalmt/volcano/actions/runs/21249532552/job/61147206148
race-gomaxprocs-4
https://github.com/hajnalmt/volcano/actions/runs/21249532552/job/61147206132
race-gomaxprocs-8
https://github.com/hajnalmt/volcano/actions/runs/21249532552/job/61147206144
no-race-gomaxprocs-1
https://github.com/hajnalmt/volcano/actions/runs/21249532552/job/61147206134
no-race-gomaxprocs-4
https://github.com/hajnalmt/volcano/actions/runs/21249532552/job/61147206139
serial-baseline-100
https://github.com/hajnalmt/volcano/actions/runs/21249532552/job/61147206130

Test Results Overview
| Configuration | GOMAXPROCS | Race Detector | Iterations | Passed | Failed | Failure Rate |
|--------------|------------|---------------|------------|--------|--------|------------------|
| race-gomaxprocs-1 | 1 | ✅ | 50 | 39 | 11 | 22% |
| race-gomaxprocs-4 | 4 | ✅ | 50 | 49 | 1 | 2% |
| race-gomaxprocs-8 | 8 | ✅ | 50 | 50 | 0 | 0% ✅ |
| no-race-gomaxprocs-1 | 1 | ❌ | 50 | 0 | 50 | 100% 🔴 |
| no-race-gomaxprocs-4 | 4 | ❌ | 50 | 37 | 13 | 26% |
| serial-baseline-100 | unset (4 cores) | ❌ | 100 | 83 | 17 | 17% |
| TOTAL | - | - | 350 | 258 | 92 | 26.3% |

From this it's quite clear that this is not a standard go race issue, but a **timing-dependent race condition between the scheduler actions and informer cache synchronization**.
What can be seen here is that if we decrease the GOMAXPROCS value and try to jam the go number of threads to have as sequential goroutine scheduling as we can the issue is becoming more and more reliably reproducible. Also race detector actually masks the problem since it adds 5-10x overhead, slowing everything down which add extra time to the informer caches to sync before test assertion.

The last line means that GitHub CI uses the default `nproc` that should be equviavalent with `GOMAXPROCS=4`, and provides a metric showing **the test will actually fail around 17% of the time on our PRs**, but fails less if there is no race and `GOMAXPROCS=4` is explicitly set. (which is 26%, for whatever magical reason)
My laptop ran with `GOMAXPROCS=12` that's why I couldn't reproduce the issue, as you can see above 4 cores the test just passes.
So try to run the test with `GOMAXPROCS=1` and without race locally and for me it failed around 50% of the time. With this the issue is replicable even locally and reliably too.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: I've added testing.T to TestCommonStruct to enable standard `t.Errorf` reporting. I included a fallback to `fmt.Printf` if T is nil so we don't break existing tests that haven't been updated to pass the testing context yet.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: I merged the storage-specific sync logic into the primary `waitForCacheSyncPolling` function. This centralizes all resource polling (Pods, Nodes, PVs, PVCs, etc.) into one place and removes the redundant function.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: I removed the manual `ResourceVersion` increment logic. The fake client's standard behavior is sufficient for these unit tests.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: Although it was suggested to remove the phase check, I found that existing tests rely on it to ensure state consistency before the session opens. I've kept it to maintain test stability.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: done sir

pkg/scheduler/uthelper/helper.go
sahitya-chandra: done sir

sahitya-chandra:
@hajnalmt sir,

This latest commit contains the fix that introduces a strategic 200ms delay in pkg/scheduler/uthelper/helper.go:RegisterSession immediately after the session and its plugins are initialized.

```bash
test.ssn = framework.OpenSession(test.cache, tiers, config)
// Strategic delay for plugin event handlers to sync
time.Sleep(200 * time.Millisecond)
```

i ran this `GOMAXPROCS=1 go test ./pkg/scheduler/actions/allocate/... -v -run TestAllocateWithPVC --count=100`

and got

```bash
=== RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
--- PASS: TestAllocateWithPVC (1.33s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.67s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.67s)
=== RUN   TestAllocateWithPVC
=== RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
E0122 21:04:53.215897  101187 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p1: binding volumes: ClaimRef got reset for pv "pv1"
E0122 21:04:53.215967  101187 cache.go:1380] execute preBind for pod c1/p1 failed: binding volumes: ClaimRef got reset for pv "pv1", resync the task
E0122 21:04:53.216268  101187 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p2: binding volumes: ClaimRef got reset for pv "pv2"
E0122 21:04:53.216309  101187 cache.go:1380] execute preBind for pod c1/p2 failed: binding volumes: ClaimRef got reset for pv "pv2", resync the task
=== RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
--- PASS: TestAllocateWithPVC (1.34s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.67s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.67s)
=== RUN   TestAllocateWithPVC
=== RUN   TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource
E0122 21:04:54.559051  101187 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p2: binding volumes: ClaimRef got reset for pv "pv2"
E0122 21:04:54.559068  101187 cache.go:1380] execute preBind for pod c1/p2 failed: binding volumes: ClaimRef got reset for pv "pv2", resync the task
E0122 21:04:54.559122  101187 predicates.go:753] PreBind plugin VolumeBinding failed for pod c1/p1: binding volumes: ClaimRef got reset for pv "pv1"
E0122 21:04:54.559146  101187 cache.go:1380] execute preBind for pod c1/p1 failed: binding volumes: ClaimRef got reset for pv "pv1", resync the task
=== RUN   TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources
--- PASS: TestAllocateWithPVC (1.33s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_but_node_without_enough_resource (0.67s)
    --- PASS: TestAllocateWithPVC/static_pv_matched_and_node_with_enough_resources (0.67s)
PASS
ok      volcano.sh/volcano/pkg/scheduler/actions/allocate       133.250s
```

Let me know if this is correct or not

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,24 +113,148 @@ type TestCommonStruct struct {
...
-	schedulerCache := test.createSchedulerCache()
+	test.cache = test.createSchedulerCache()
+	test.waitForCacheSyncPolling(test.cache)
 	RegisterPlugins(test.Plugins)
-	test.ssn = framework.OpenSession(schedulerCache, tiers, config)
+	test.tiers = tiers
+	test.configurations = config
+	test.ssn = framework.OpenSession(test.cache, tiers, config)
+
+	time.Sleep(200 * time.Millisecond)
```
hajnalmt: We shouldn't introduce sleeps in unit tests only as a last-last-last-last resort. This actually increased the execution time of the unit tests by nearly a minute. Did the test failed for you locally even with the cache sync changes?

pkg/scheduler/api/job_info.go
```diff
@@ -287,7 +287,7 @@ func (ti *TaskInfo) Clone() *TaskInfo {
 		Namespace:                   ti.Namespace,
 		TaskRole:                    ti.TaskRole,
 		Priority:                    ti.Priority,
-		Pod:                         ti.Pod,
+		Pod:                         ti.Pod.DeepCopy(),
```
hajnalmt, line 290: I feel like there is a new surprising change with each iteration of yours, why did you modify this to be a DeepCopy?

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,24 +113,148 @@ type TestCommonStruct struct {
...
+
 	return test.ssn
 }
 
+func (test *TestCommonStruct) waitForCacheSyncPolling(schedulerCache *cache.SchedulerCache) {
+	// Wait for all nodes and pods to be synced into the cache with their correct status
+	timeout := 2 * time.Second
```
hajnalmt: This inconsistent with the comment above `CacheSyncTimeout` where you say that this is 3.

pkg/scheduler/uthelper/helper.go
```diff
@@ -136,16 +266,92 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
...
+	}
+	schedulerCache.HyperNodesInfo = schedulingapi.NewHyperNodesInfoWithCache(test.HyperNodesMap, test.HyperNodesSetByTier, test.HyperNodes, ready)
+
 	// Initial provisioning resources
 	kubeClient := schedulerCache.Client()
 	for _, sc := range test.SCs {
 		kubeClient.StorageV1().StorageClasses().Create(context.Background(), sc, metav1.CreateOptions{})
 	}
+
+	fakeClient := kubeClient.(*fakek8s.Clientset)
```
hajnalmt: Why do you still need this if you deleted the `Revision` patch?

pkg/scheduler/uthelper/helper.go
```diff
@@ -178,27 +394,35 @@ func (test *TestCommonStruct) createSchedulerCache() *cache.SchedulerCache {
...
-	ready := new(atomic.Bool)
-	ready.Store(true)
-	for _, hni := range test.HyperNodesMap {
-		if hni.HyperNode == nil {
-			continue
+
+	schedulerCache.Run(test.stop)
+	schedulerCache.WaitForCacheSync(test.stop)
+
+	for _, pod := range test.Pods {
```
hajnalmt, line 407: This is really contra-productive for me why isn't this in the WaitForCacheSync function? We go through the pods there already and `create` is now out of it.

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,24 +113,148 @@ type TestCommonStruct struct {
...
 }
 
+func (test *TestCommonStruct) waitForCacheSyncPolling(schedulerCache *cache.SchedulerCache) {
+	// Wait for all nodes and pods to be synced into the cache with their correct status
+	timeout := 2 * time.Second
+	if test.CacheSyncTimeout > 0 {
+		timeout = test.CacheSyncTimeout
+	}
+
+	err := wait.PollImmediate(10*time.Millisecond, timeout, func() (bool, error) {
```
hajnalmt: `PollImmediate` is deprecated please use `PollUntilContextTimeout` and check for HasSynced first.
`sc.informerFactory.Core().V1().PersistentVolumes().Informer().HasSynced`, `sc.informerFactory.Core().V1().PersistentVolumeClaims().Informer().HasSynced`,etc..

hajnalmt (changes requested):
Thank you for squashing the commits this makes a lot easier to review this. Please find my comments below.

pkg/scheduler/api/job_info.go
sahitya-chandra: Sorry sir, I didn't explain this one but previously, `TaskInfo.Clone()` just copied the Pod pointer, so the session and the cache were sharing the same `*v1.Pod` instance. That means the session’s write to `Spec.NodeName` is racing with the informer thread’s read.

By changing `Clone()` to use `Pod.DeepCopy()`, the session works on its own copy of the pod. This aligns with the usual invariant that objects inside the scheduler session are independent snapshots derived from the cache, and it removes the reported race without introducing new shared‑state surprises.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: updated waitForCacheSyncPolling to use 3 seconds

pkg/scheduler/uthelper/helper.go
sahitya-chandra: Agreed, with the current version of the tests we no longer need the custom fake client reactors. I removed the `fakek8s.Clientset` cast and the `PrependReactor` code, and the tests still pass under -race

pkg/scheduler/uthelper/helper.go
sahitya-chandra: I separated responsibilities here on purpose:

- `addTestData` is responsible for creating all the test resources (including pods) and setting their status.
- `waitForCacheSyncPolling` is responsible for verifying that the scheduler cache and the plugin/informer caches have fully observed that state (nodes present with allocatable, tasks present with the right status, SC/PV/PVC/DRA objects available via listers).

I agree the loop in `waitForCacheSyncPolling` that walks pods makes it look like it “owns” pod handling, but it’s only checking that the cache has caught up  - all creation still happens in `addTestData`. I can rename the function (e.g., ensureTestDataSynced) or add a clarifying comment if you prefer, but I’d like to keep creation and waiting as separate steps so it’s clear what has side‑effects versus what is just a readiness check.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: In my latest commit I’ve switched `waitForCacheSyncPolling` to use `wait.PollUntilContextTimeout` with a context, and to check informer `HasSynced()` before inspecting the cache state.

Specifically, it now:
- Creates a context with the configured timeout,
- Uses `PollUntilContextTimeout` with a 10ms interval,
- Checks that the Node, PV, PVC, StorageClass, and DRA (DeviceClass/ResourceClaim/ResourceSlice) informers have all reported `HasSynced`,
- Then, under the cache mutex, verifies that:
- Nodes are present with allocatable resources,
- All expected tasks are present in the jobs map with the right status,
- The listers can `Get` all configured SCs/PVs/PVCs/DRA resources.

This replaces the deprecated `PollImmediate` and makes the sync logic better aligned with how the shared informer factory is supposed to be waited on.

pkg/scheduler/uthelper/helper.go
sahitya-chandra: I agree we should avoid sleeps in unit tests. The sleep was part of an earlier iteration when I was experimenting with making the plugin informers (especially VolumeBinding) reliably see the test data. In the final version, the sleep is removed.

The test setup is now:
1) Open a “warm‑up” session to trigger plugin `OnSessionOpen` and informer registrations,
2) Add all test data (nodes, pods, PVs, PVCs, DRA objects, etc.),
3) Wait deterministically for cache + informer sync via `waitForCacheSyncPolling`,
4) Open the real session over the fully‑populated cache.

This keeps the tests deterministic without any arbitrary `time.Sleep`, and locally the previously failing tests now pass under `-race` with this structure.

sahitya-chandra:
the latest commit now,

**Stabilize scheduler unit tests**: Rework `TestCommonStruct` to create a `SchedulerCache` with fake binder/evictor, start and sync its informers, then add all test data via the fake client and wait for the cache to fully observe nodes, pods, PVs, PVCs, SCs, and DRA resources. This replaces the ad‑hoc `time.Sleep` with deterministic `PollUntilContextTimeout` + `HasSynced` checks.
**Improve cache‑sync helper semantics**: Align the default `CacheSyncTimeout` to 3s as documented, remove now‑unused fake‑client reactors, and keep the responsibilities clean: `initInfrastructure` starts the cache, `addTestData` creates resources, and `waitForCacheSyncPolling` only waits/validates that the cache and plugin state are fully in sync.

pkg/scheduler/api/job_info.go
hajnalmt: Ugh you are right. Nice catch.

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,45 +110,269 @@ type TestCommonStruct struct {
...
 
 var _ Interface = &TestCommonStruct{}
 
 // RegisterSession open session with tiers and configuration, and mock schedulerCache with self-defined FakeBinder and FakeEvictor
 func (test *TestCommonStruct) RegisterSession(tiers []conf.Tier, config []conf.Configuration) *framework.Session {
-	schedulerCache := test.createSchedulerCache()
+	test.initInfrastructure()
 	RegisterPlugins(test.Plugins)
-	test.ssn = framework.OpenSession(schedulerCache, tiers, config)
+	test.tiers = tiers
```
hajnalmt: This doesn't make any sense to me why to extend `TestCommonStruct` with tiers and configurations? when you won't use it in the end? Or where is this used?

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,45 +110,269 @@ type TestCommonStruct struct {
...
+	test.configurations = config
+
+	_ = framework.OpenSession(test.cache, tiers, config)
```
hajnalmt: Why do you open the session twice?

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,45 +110,269 @@ type TestCommonStruct struct {
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	err := wait.PollUntilContextTimeout(ctx, 10*time.Millisecond, timeout, true, func(ctx context.Context) (bool, error) {
+		// First wait for the underlying shared informers to report synced state.
+		informerFactory := schedulerCache.SharedInformerFactory()
+
+		if !informerFactory.Core().V1().Nodes().Informer().HasSynced() {
```
hajnalmt:
```
informerFactory := schedulerCache.SharedInformerFactory()

type informerCheck struct {
    name      string
    hasSynced func() bool
    required  bool
}

checks := []informerCheck{
    {"Nodes", informerFactory.Core().V1().Nodes().Informer().HasSynced, true},
    {"PersistentVolumes", informerFactory.Core().V1().PersistentVolumes().Informer().HasSynced, len(test.PVs) > 0},
    {"PersistentVolumeClaims", informerFactory.Core().V1().PersistentVolumeClaims().Informer().HasSynced, len(test.PVCs) > 0},
    {"StorageClasses", informerFactory.Storage().V1().StorageClasses().Informer().HasSynced, len(test.SCs) > 0},
    {"DeviceClasses", informerFactory.Resource().V1().DeviceClasses().Informer().HasSynced, len(test.DeviceClasses) > 0},
    {"ResourceClaims", informerFactory.Resource().V1().ResourceClaims().Informer().HasSynced, len(test.ResourceClaims) > 0},
    {"ResourceSlices", informerFactory.Resource().V1().ResourceSlices().Informer().HasSynced, len(test.ResourceSlices) > 0},
}

for _, check := range checks {
    if check.required && !check.hasSynced() {
        return false, nil
    }
}
```

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,45 +110,269 @@ type TestCommonStruct struct {
...
+					found = true
+					break
+				}
+			}
+			if !found {
+				return false, nil
+			}
+		}
+
+		for _, sc := range test.SCs {
```
hajnalmt, line 215:
```
for _, sc := range test.SCs {
    if _, err := informerFactory.Storage().V1().StorageClasses().Lister().Get(sc.Name); err != nil {
        return false, nil
    }
}
```
And elsewhere too.

hajnalmt (changes requested):

pkg/scheduler/uthelper/helper.go
sahitya-chandra: removed both tiers and config

pkg/scheduler/uthelper/helper.go
sahitya-chandra: Without the first call, the plugins are not looking to the initial test data. Without the second call, the session's snapshot is empty. added this to synchronized test state.

but now I removed it

pkg/scheduler/actions/allocate/allocate_test.go
```diff
@@ -2635,8 +2636,8 @@ func TestAllocateWithPVC(t *testing.T) {
 			PVs:                []*v1.PersistentVolume{pv1, pv2},
 			PVCs:               []*v1.PersistentVolumeClaim{pvc1, pvc2},
 			IgnoreProvisioners: ignoreProvisioners,
-			ExpectTaskStatusNums: map[api.JobID]map[api.TaskStatus]int{
-				"c1/pg1": {api.Binding: 2},
+			ExpectStatus: map[api.JobID]scheduling.PodGroupPhase{
+				"c1/pg1": scheduling.PodGroupInqueue,
```
sahitya-chandra, line 2640: I am not sure if this is correct or not but this fixed the failing `TestAllocateWithPVC` test after I removed `_ = framework.OpenSession(test.cache, tiers, config)` 

This fix is because the test was checking the wrong thing. It should verify scheduler behavior (podgroup status), not PV controller behavior (task binding status), which requires a full PV controller mock that isn't present in unit tests.

I can revert this if this is not a good approach

JesseStutler: I think it's meanless to check the status of podgroup to be inqueue because the intial status of the podgroup is also inqueue. Check the `Binding`, indeed it doesn't need the pv controller because there's no actual execution of preBind, it's just that the session has already allocated for these pods

hzxuzhonghu (approved):
Let's fix this ASAP, seems no risk

pkg/scheduler/actions/allocate/allocate_test.go
hajnalmt: Modifying the unit test to expect the input as output is not the way to go. 😄

hajnalmt (changes requested):

pkg/scheduler/actions/allocate/allocate_test.go
sahitya-chandra: @hajnalmt sir, I think the root cause is
- The test expected 2 tasks to reach Binding status, but they never did.
- The volume binding plugin's PreBind step requires PV/PVC binding to complete, which depends on a PV controller that isn't fully mocked in unit tests.

sahitya-chandra: > Modifying the unit test to expect the input as output is not the way to go. 😄

Sorry sir, that's why I mentioned that I can revert this

JesseStutler: But if the allocate in session is successful, the pod will be binding state then, it even not go to PreBind yet.

pkg/scheduler/uthelper/helper.go
```diff
@@ -107,45 +110,257 @@ type TestCommonStruct struct {
...
+		// First wait for the underlying shared informers to report synced state.
+		informerFactory := schedulerCache.SharedInformerFactory()
+
+		type informerCheck struct {
+			name      string
+			hasSynced func() bool
+			required  bool
+		}
+
+		checks := []informerCheck{
```
JesseStutler, line 159: We already have WaitForCacheSync in cache's Run method, do we still need these?

JesseStutler: I think polling to check if the resources in the cases are already in the indexer is correct (by checking the lister). WaitForCacheSync will only wait for all resource lists to complete and trigger the EventHandler, but if the EventHandler has asynchronous logic, it won't wait.

sahitya-chandra:
@JesseStutler sir, after some research, the issue

The VolumeBinding plugin's AssumeCache needs to see the PV/PVC objects. 
we have three options that can fix the issue:
- Initialize VolumeBinding plugin BEFORE addTestData() - so event handlers exist when objects are added
- Trigger a resync after the VolumeBinding plugin is created
- Directly populate the AssumeCache using assumecache.AddTestObject() (like binder_test.go does)

Can you help me with which approach should I go with?

cc: @hajnalmt

JesseStutler:
> @JesseStutler sir, after some research, the issue
> 
> The VolumeBinding plugin's AssumeCache needs to see the PV/PVC objects. we have three options that can fix the issue:
> 
> * Initialize VolumeBinding plugin BEFORE addTestData() - so event handlers exist when objects are added
> * Trigger a resync after the VolumeBinding plugin is created
> * Directly populate the AssumeCache using assumecache.AddTestObject() (like binder_test.go does)
> 
> Can you help me with which approach should I go with?
> 
> cc: @hajnalmt

I think we can AddTestObject to the assume cache, I don't prefer to change the initialization way of plugins or scheduler, but data we can mock it

JesseStutler:
I think we can still continue it even after we released v1.14, it's a flaky ut test rather than features or bugs, I think it's because we should ensure that every mock data is already synced to the cache

Issue #4819, opened 2025-12-16
Hierarchical Queue Capability Can Exceed Cluster Resources, Causing Validation Failures


zjj2wry:
### Description

In Volcano's hierarchical queue mode, queues can be created with `capability` that exceeds the cluster's total resources or parent queue's capability. While the scheduler validates that `child.capability <= parent.capability` during scheduling, this validation **runs too late** (during OnSessionOpen, not at queue creation) and can **abort the entire scheduling cycle** when violated. Additionally, there is **no validation** that the root queue's capability does not exceed cluster total resources, leading to misleading configurations.

#### Queue Configuration (Invalid but Accepted)

**Cluster actual resources**: 100 CPU, 200Gi memory

```yaml

apiVersion: scheduling.volcano.sh/v1beta1
kind: Queue
metadata:
  name: org-1
spec:
  parent: root
  weight: 1
  capability:
    cpu: "200"
    memory: "400Gi"
  # Violates: org-1.capability (200 CPU) > root.capability (100 CPU)
  # Violates: org-1.memory (400Gi) > root.memory (200Gi)
  guarantee:
    cpu: "50"
```

**Problems**:
- org-1.capability.cpu (200) > cluster total (100) ✗
- org-1.capability.memory (400Gi) > cluster total (200Gi) ✗
- Since root.capability is auto-set to cluster total, this also means:
  - org-1.capability (200, 400Gi) > root.capability (100, 200Gi) ✗

#### What Happens

1. **Queue creation succeeds**:
```bash
kubectl apply -f queues.yaml
# queue.scheduling.volcano.sh/org-1 created
```

**No error or warning!**

2. **Scheduler detects parent-child violation and aborts**:

When the next scheduling cycle runs, `checkHierarchicalQueue` executes:

```go
// Checks: root.capability >= org-1.capability
if attr.capability.LessPartly(childAttr.capability, api.Zero) {
    return fmt.Errorf("queue <%s> capability <%s> is less than its child queue <%s> capability <%s>",
        attr.name, attr.capability, childAttr.name, childAttr.capability)
}
```

**Scheduler log shows**:
```
E1216 10:00:00.123456 capacity.go:600] Failed to check queue's hierarchical structure,
error: queue <root> capability <cpu:100000,memory:214748364800> is less than its child queue <org-1>
capability <cpu:200000,memory:429496729600>
```

**Result**: `OnSessionOpen` returns `false`, causing:
- The entire scheduling cycle to be aborted
[...]

PR #4864, opened 2025-12-23
fix: prevent cluster-wide scheduling failure due to queue hierarchy validation

### LINKED ISSUE #4818: Missing Validation for Child Queue Guarantee Sum Causes Cluster-Wide Scheduling Failure

Volcano does not validate that the sum of child queue guarantees does not exceed the parent queue's guarantee during queue creation/update. However, the scheduler enforces this constraint during scheduling, causing **all jobs to fail** when the constraint is violated. This creates a catastrophic failure mode where invalid queue configurations can silently break the entire cluster's scheduling capability.

### Expected Behavior

When creating or updating hierarchical queues, the system should:

1. **Validate at creation time**: Reject queue creation/update if `sum(child.guarantee) > parent.guarantee`
2. **Fail fast with clear error**: Return a validation error explaining which child queues violate the constraint
3. **Prevent invalid state**: Never allow the cluster to enter a state where scheduling cannot proceed

### Actual Behavior

1. **Queue creation succeeds** even when `sum(child.guarantee) > parent.guarantee`
2. **No validation error** is returned to the user
3. **Scheduler silently fails** to schedule any jobs in affected queues
4. **Entire cluster becomes unschedulable** if the violation occurs at high levels of the hierarchy

#### Missing Validation in Queue Creation

The queue validation webhook and controller do not check the guarantee sum constraint.

#### Scheduler Enforces Constraint During Scheduling

https://github.com/volcano-sh/volcano/blob/943b8c23d402462b63f1e21f7f7a39129ba8d63c/pkg/scheduler/plugins/capacity/capacity.go#L100

### Additional context

_No response_

### Documentation Updates

zjj2wry:
#### What type of PR is this?
/kind bug

Scheduler crashes and blocks all scheduling when queue hierarchy validation fails. Invalid queue configurations (child capability > parent, sum of children's guarantee > parent's guarantee) could be created via kubectl, causing cluster-wide scheduling failures. 
- Added webhook validation - Validates hierarchical queue constraints at admission time:
    - Child capability ≤ parent capability
    - Sum of siblings/children guarantee ≤ parent guarantee
- Changed scheduler behavior - checkHierarchicalQueue() now only logs warnings instead of returning errors, preventing cluster-wide scheduling failures

Fixes #4818 #4819

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,8 +106,27 @@ images:
 		docker buildx build -t "${IMAGE_PREFIX}/vc-$$name:$(TAG)" . -f ./installer/dockerfile/$$name/Dockerfile --output=type=${BUILDX_OUTPUT_TYPE} --platform ${DOCKER_PLATFORMS} --build-arg APK_MIRROR=${APK_MIRROR} --build-arg OPEN_EULER_IMAGE_TAG=${OPEN_EULER_IMAGE_TAG}; \
 	done
 
+# Define a reusable build function for individual component images
+define build_component_image
+	docker buildx build -t "${IMAGE_PREFIX}/vc-$(1):$(TAG)" . \
+		-f ./installer/dockerfile/$(1)/Dockerfile \
+		--output=type=${BUILDX_OUTPUT_TYPE} \
+		--platform ${DOCKER_PLATFORMS} \
+		--build-arg APK_MIRROR=${APK_MIRROR} \
+		--build-arg OPEN_EULER_IMAGE_TAG=${OPEN_EULER_IMAGE_TAG}
+endef
+
+vc-controller-manager-image:
+	$(call build_component_image,controller-manager)
+
+vc-scheduler-image:
+	$(call build_component_image,scheduler)
+
+vc-webhook-manager-image:
+	$(call build_component_image,webhook-manager)
+
 vc-agent-image:
-	docker buildx build -t "${IMAGE_PREFIX}/vc-agent:$(TAG)" . -f ./installer/dockerfile/agent/Dockerfile --output=type=${BUILDX_OUTPUT_TYPE} --platform ${DOCKER_PLATFORMS} --build-arg APK_MIRROR=${APK_MIRROR} --build-arg OPEN_EULER_IMAGE_TAG=${OPEN_EULER_IMAGE_TAG}
+	$(call build_component_image,agent)
 
 generate-code:
 	./hack/update-gencode.sh
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -596,13 +596,9 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 		rootQueueAttr.deserved = cp.totalResource
 	}
 	rootQueueAttr.realCapability = cp.totalResource
-	// Check the hierarchical structure of queues
-	err := cp.checkHierarchicalQueue(rootQueueAttr)
-	if err != nil {
-		klog.Errorf("Failed to check queue's hierarchical structure, error: %v", err)
-		return false
-	}
-	klog.V(4).Infof("Successfully checked queue's hierarchical structure.")
+	// checkHierarchicalQueue only logs warnings and never returns errors
+	// to avoid aborting the entire scheduling cycle due to configuration issues
+	cp.checkHierarchicalQueue(rootQueueAttr)
 
 	// update session attributes
 	ssn.TotalGuarantee = cp.totalGuarantee
@@ -744,7 +740,7 @@ func (cp *capacityPlugin) updateAncestors(queue *api.QueueInfo, ssn *framework.S
 	return nil
 }
 
-func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
+func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) {
 	totalGuarantee := api.EmptyResource()
 	totalDeserved := api.EmptyResource()
 	for _, childAttr := range attr.children {
@@ -772,8 +768,9 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
 
 		// Check if the parent queue's capability is less than the child queue's capability
 		if attr.capability.LessPartly(childAttr.capability, api.Zero) {
-			return fmt.Errorf("queue <%s> capability <%s> is less than its child queue <%s> capability <%s>",
-				attr.name, attr.capability, childAttr.name, childAttr.capability)
+			klog.V(3).Infof("Child queue %s capability (%s) exceeds parent queue %s capability (%s). "+
+				"Child's effective capability will be limited by parent.",
+				childAttr.name, childAttr.capability, attr.name, attr.capability)
 		}
 	}
 
@@ -801,23 +798,22 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
 
 	// Check if the parent queue's deserved resources are less than the total deserved resources of child queues
 	if attr.deserved.LessPartly(totalDeserved, api.Zero) {
-		return fmt.Errorf("queue <%s> deserved resources <%s> are less than the sum of its child queues' deserved resources <%s>",
-			attr.name, attr.deserved, totalDeserved)
+		klog.V(3).Infof("Sum of child queue deserved (%s) exceeds parent queue %s deserved (%s). "+
+			"This may affect resource distribution during scheduling.",
+			totalDeserved, attr.name, attr.deserved)
 	}
 
 	// Check if the parent queue's guarantee resources are less than the total guarantee resources of child queues
 	if attr.guarantee.LessPartly(totalGuarantee, api.Zero) {
-		return fmt.Errorf("queue <%s> guarantee resources <%s> are less than the sum of its child queues' guarantee resources <%s>",
-			attr.name, attr.guarantee, totalGuarantee)
+		klog.V(3).Infof("Sum of child queue guarantees (%s) exceeds parent queue %s guarantee (%s). "+
+			"Not all child guarantees can be satisfied simultaneously.",
+			totalGuarantee, attr.name, attr.guarantee)
 	}
 
+	// Recursively check child queues
 	for _, childAttr := range attr.children {
-		err := cp.checkHierarchicalQueue(childAttr)
-		if err != nil {
-			return err
-		}
+		cp.checkHierarchicalQueue(childAttr)
 	}
-	return nil
 }
 
 // compareShareWithDeserved compares two queueAttr by share; when shares are equal,
--- a/pkg/webhooks/admission/queues/validate/validate_queue.go
+++ b/pkg/webhooks/admission/queues/validate/validate_queue.go
@@ -24,10 +24,13 @@ import (
 	admissionv1 "k8s.io/api/admission/v1"
 	whv1 "k8s.io/api/admissionregistration/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/klog/v2"
+	k8score "k8s.io/kubernetes/pkg/apis/core"
+	k8scorevalid "k8s.io/kubernetes/pkg/apis/core/validation"
 
 	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	"volcano.sh/volcano/pkg/scheduler/api"
@@ -90,6 +93,16 @@ func AdmitQueues(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse 
 
 		if ar.Request.Operation == admissionv1.Create || oldQueue.Spec.Parent != queue.Spec.Parent {
 			err = validateHierarchicalQueue(queue)
+			if err != nil {
+				break
+			}
+		}
+
+		if needsValidateHierarchicalQueue(queue, oldQueue, ar.Request.Operation) {
+			err = validateHierarchicalQueueResources(queue)
+			if err != nil {
+				break
+			}
 		}
 
 	case admissionv1.Delete:
@@ -207,6 +220,24 @@ func validateStateOfQueue(value schedulingv1beta1.QueueState, fldPath *field.Pat
 
 func validateResourceOfQueue(resource schedulingv1beta1.QueueSpec, fldPath *field.Path) field.ErrorList {
 	errs := field.ErrorList{}
+
+	// Validate resource quantity values are non-negative
+	if len(resource.Capability) != 0 {
+		for resourceName, quantity := range resource.Capability {
+			errs = append(errs, k8scorevalid.ValidateResourceQuantityValue(k8score.ResourceName(resourceName), quantity, fldPath.Child("capability").Child(string(resourceName)))...)
+		}
+	}
+	if len(resource.Deserved) != 0 {
+		for resourceName, quantity := range resource.Deserved {
+			errs = append(errs, k8scorevalid.ValidateResourceQuantityValue(k8score.ResourceName(resourceName), quantity, fldPath.Child("deserved").Child(string(resourceName)))...)
+		}
+	}
+	if len(resource.Guarantee.Resource) != 0 {
+		for resourceName, quantity := range resource.Guarantee.Resource {
+			errs = append(errs, k8scorevalid.ValidateResourceQuantityValue(k8score.ResourceName(resourceName), quantity, fldPath.Child("guarantee").Child("resource").Child(string(resourceName)))...)
+		}
+	}
+
 	capabilityResource := api.NewResource(resource.Capability)
 	deservedResource := api.NewResource(resource.Deserved)
 	guaranteeResource := api.NewResource(resource.Guarantee.Resource)
@@ -223,7 +254,8 @@ func validateResourceOfQueue(resource schedulingv1beta1.QueueSpec, fldPath *fiel
 			guaranteeResource.String(), "guarantee should less equal than capability"))
 	}
 
-	if len(resource.Deserved) != 0 &&
+	// Validate guarantee <= deserved when guarantee is set
+	if len(resource.Guarantee.Resource) != 0 &&
 		deservedResource.LessPartly(guaranteeResource, api.Zero) {
 		return append(errs, field.Invalid(fldPath.Child("guarantee"),
 			guaranteeResource.String(), "guarantee should less equal than deserved"))
@@ -259,6 +291,47 @@ func validateQueueDeleting(queueName string) error {
 	klog.V(3).Infof("Validation passed for deleting hierarchical queue %s", queue.Name)
 
 	return nil
+}
+
+// needsValidateHierarchicalQueue determines if hierarchy resource validation is necessary
+// Returns true only if:
+// - Queue is not the root queue itself AND
+// - Either it's a CREATE operation OR resources (capability/deserved/guarantee) changed on UPDATE OR parent changed on UPDATE
+func needsValidateHierarchicalQueue(queue, oldQueue *schedulingv1beta1.Queue, operation admissionv1.Operation) bool {
+	// Only the root queue itself should skip validation completely
+	if queue.Name == "root" {
+		return false
+	}
+
+	// For CREATE operations, always validate
+	if operation == admissionv1.Create {
+		return true
+	}
+
+	// For UPDATE operations, only validate if resources or parent changed
+	if operation == admissionv1.Update && oldQueue != nil {
+		// Check if parent changed
+		if queue.Spec.Parent != oldQueue.Spec.Parent {
+			return true
+		}
+		// Check if capability changed
+		if !equality.Semantic.DeepEqual(queue.Spec.Capability, oldQueue.Spec.Capability) {
+			return true
+		}
+		// Check if deserved changed
+		if !equality.Semantic.DeepEqual(queue.Spec.Deserved, oldQueue.Spec.Deserved) {
+			return true
+		}
+		// Check if guarantee changed
+		if !equality.Semantic.DeepEqual(queue.Spec.Guarantee.Resource, oldQueue.Spec.Guarantee.Resource) {
+			return true
+		}
+		// No resource or parent changes, skip validation
+		return false
+	}
+
+	// Default: skip validation
+	return false
 }
 
 func validateHierarchicalQueue(queue *schedulingv1beta1.Queue) error {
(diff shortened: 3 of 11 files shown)


pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -90,6 +90,20 @@ func AdmitQueues(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse
...
+			}
+		}
+
+		// Validate all hierarchy resource constraints
+		// This includes:
+		// - Child's capability <= parent's capability
+		// - Sum of siblings' guarantee/deserved <= parent's limit
+		// - Parent's capability >= each child's capability
+		// - Sum of children's guarantee/deserved <= parent's limit
+		err = validateHierarchicalQueueResources(queue)
```
JesseStutler: I prefer to also add a neesValidateHierarchicalQueue judgement
1. If the queue doesn't have a parent, skip the validateHierarchicalQueueResources judgement
2. If the queue has a parent, verify that the queue's capability/deserved/guaratee updated, otherwise also skip

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -317,3 +331,160 @@ func listQueueChild(parentQueueName string) ([]string, error) {
 
 	return childQueueNames, nil
 }
+
+// validateHierarchicalQueueResources validates all hierarchy resource constraints for a queue
+// This function performs all necessary checks in a single pass to minimize API calls
+func validateHierarchicalQueueResources(queue *schedulingv1beta1.Queue) error {
+	// List all queues once for all validations
+	queueList, err := config.QueueLister.List(labels.Everything())
```
JesseStutler: Could we add an indexFunc for the queue informer to create a fast indexer for finding queue's parent and childs? Currently I don't prefer using labels.Everything to fetch all the queues

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -317,3 +331,160 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+	}
+
+	return nil
+}
+
+// validateChildAgainstParent validates that child queue's resources don't exceed parent's
+func validateChildAgainstParent(child, parent *schedulingv1beta1.Queue) error {
+	childCapability := api.NewResource(child.Spec.Capability)
+	parentCapability := api.NewResource(parent.Spec.Capability)
+
```
guoqinwill, line 439: For capability, a single level of validation is not enough, because when capability is not set, it inherits from the parent queue. Therefore, capability validation needs to be done recursively both upwards and downwards. 
1. Recursively upwards until you find a capability configured for the corresponding dimension and check if it is less than the nearest ancestor node. If none is configured, continue searching up to the root.
2. Recursively downwards to check all descendant nodes. For example, first extract the maximum capability for the corresponding dimension among the descendants and check if it is greater than the capability of this descendant queue.

Additionally, it is important to note that the iterative search for different capability dimensions varies, so dimensions such as CPU and memory should be checked separately.

guoqinwill:
I think we can add another function here to independently validate the queue's capability, deserved, and guarantee quotas, such as using 'ValidateResourceQuantityValue' for legality checks, for example, not allowing quota values to be negative and requiring that guarantee <= deserved <= capability.
[pkg/webhooks/admission/queues/validate/validate_queue.go#L128-L301](https://github.com/volcano-sh/volcano/blob/2090e3b2c7ce912acf4983e2141dafc37864d313/pkg/webhooks/admission/queues/validate/validate_queue.go#L128-L132)

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -795,8 +798,10 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
 
 		// Check if the parent queue's capability is less than the child queue's capability
 		if attr.capability.LessPartly(childAttr.capability, api.Zero) {
-			return fmt.Errorf("queue <%s> capability <%s> is less than its child queue <%s> capability <%s>",
-				attr.name, attr.capability, childAttr.name, childAttr.capability)
+			msg := fmt.Sprintf("Child queue %s capability (%s) exceeds parent capability (%s). "+
+				"Child's effective capability will be limited by parent.",
+				childAttr.name, childAttr.capability, attr.capability)
+			cp.recordQueueEvent(attr, ssn, v1.EventTypeWarning, "CapabilityExceeded", msg)
```
guoqinwill: It is recommended to include these check failure messages in the logs as well, to make it easier to locate problems, since we may not be able to describe the queue's events immediately.

guoqinwill:
In the capacity plugin, the root queue should have unlimited default resources, allowing tasks that exceed resource limits to be scheduled normally.
https://github.com/volcano-sh/volcano/blob/2090e3b2c7ce912acf4983e2141dafc37864d313/pkg/scheduler/plugins/capacity/capacity.go#L599-L606
 This can also resolve the issue you mentioned where reclaim conflicts with enqueue. At the same time, we should reject modifications or updates related to the spec of the root queue, because the root queue is automatically created by Volcano and should not restrict user job submissions based on cluster resources. There has been related discussion on this topic (see https://github.com/volcano-sh/volcano/pull/4662#discussion_r2486119131), but it seems that the author did not submit this change. I think your submission can achieve this.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -317,3 +331,160 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+	queueList, err := config.QueueLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list queues: %v", err)
+	}
+
+	// Build a map for quick lookup
+	queueMap := make(map[string]*schedulingv1beta1.Queue)
+	for _, q := range queueList {
+		queueMap[q.Name] = q
+	}
```
hzxuzhonghu: This is of bad performance, you can get the specific one need using Get(name)

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: This isn't right. If no parent is configured, the default is the root, and it still needs to verify its own capability as a parent queue.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -317,3 +331,160 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+
+	// If this is a child queue, validate against parent
+	if queue.Spec.Parent != "" && queue.Spec.Parent != "root" {
+		parentQueue, exists := queueMap[queue.Spec.Parent]
+		if !exists {
+			return fmt.Errorf("parent queue %s not found", queue.Spec.Parent)
```
hzxuzhonghu: Donot we allow create child queue before parent queue? I am not sure about this @JesseStutler Please confirm

pkg/scheduler/plugins/capacity/capacity.go
hzxuzhonghu: I think we should not record event here, if this is matched, it will emit an event per batch duplicately

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: > Donot we allow create child queue before parent queue? I am not sure about this @JesseStutler Please confirm

Yes, it is necessary to confirm that each queue has a parent queue before a hierarchy can be built. This logic was also set during the original validation of hierarchical queues.See：
https://github.com/volcano-sh/volcano/blob/2090e3b2c7ce912acf4983e2141dafc37864d313/pkg/webhooks/admission/queues/validate/validate_queue.go#L296-L299

pkg/scheduler/plugins/capacity/capacity.go
guoqinwill: > I think we should not record event here, if this is matched, it will emit an event per batch duplicately

I understand that this can be kept, because now we intercept most unreasonable configurations through the webhook. Configuration issues should be rare, and if they occur, the event information in the queue is also a useful way to detect them, since some users are accustomed to checking the logs, while others prefer to pay attention to the event information.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -767,7 +763,14 @@ func (cp *capacityPlugin) updateAncestors(queue *api.QueueInfo, ssn *framework.S
...
 
-func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
+// recordQueueEvent sends a K8s event to the queue if it exists in the session
+func (cp *capacityPlugin) recordQueueEvent(attr *queueAttr, ssn *framework.Session, eventType, reason, message string) {
+	if queueInfo, found := ssn.Queues[attr.queueID]; found {
+		ssn.RecordQueueEvent(queueInfo, eventType, reason, message)
+	}
+}
+
+func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr, ssn *framework.Session) {
```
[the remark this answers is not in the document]
zjj2wry: The concern about the function name , but it should be addressed in a separate PR focused on capacity plugin improvements.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -317,3 +331,160 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
 }
+
+// validateHierarchicalQueueResources validates all hierarchy resource constraints for a queue
+// This function performs all necessary checks in a single pass to minimize API calls
+func validateHierarchicalQueueResources(queue *schedulingv1beta1.Queue) error {
+	// List all queues once for all validations
+	queueList, err := config.QueueLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list queues: %v", err)
+	}
```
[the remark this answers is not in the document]
zjj2wry: fixed by informer indexer

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: good catch! fixed

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: good catch, fixed

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: fixed.

1. directly use lister get queue
2. use informer indexer get child queue

zjj2wry:
> I think we can add another function here to independently validate the queue's capability, deserved, and guarantee quotas, such as using 'ValidateResourceQuantityValue' for legality checks, for example, not allowing quota values to be negative and requiring that guarantee <= deserved <= capability. [pkg/webhooks/admission/queues/validate/validate_queue.go#L128-L301](https://github.com/volcano-sh/volcano/blob/2090e3b2c7ce912acf4983e2141dafc37864d313/pkg/webhooks/admission/queues/validate/validate_queue.go#L128-L132)

@guoqinwill Good catch! I've implemented the independent quota validation function as you suggested.

zjj2wry:
> In the capacity plugin, the root queue should have unlimited default resources, allowing tasks that exceed resource limits to be scheduled normally.
> 
> https://github.com/volcano-sh/volcano/blob/2090e3b2c7ce912acf4983e2141dafc37864d313/pkg/scheduler/plugins/capacity/capacity.go#L599-L606
> 
> 
> This can also resolve the issue you mentioned where reclaim conflicts with enqueue. At the same time, we should reject modifications or updates related to the spec of the root queue, because the root queue is automatically created by Volcano and should not restrict user job submissions based on cluster resources. There has been related discussion on this topic (see [#4662 (comment)](https://github.com/volcano-sh/volcano/pull/4662#discussion_r2486119131)), but it seems that the author did not submit this change. I think your submission can achieve this.

The scheduler framework actively updates the root queue's `Spec.Deserved` and `Spec.Guarantee.Resource` in 
https://github.com/volcano-sh/volcano/blob/3275f8aace43d7c6dc36ca863e551074825a856e/pkg/scheduler/framework/session.go#L546-L585

In my use case, I need:
- **Root queue during scheduling**: Unlimited resources (to avoid blocking jobs)
- **Root queue quota (guarantee/deserved/capability)**: Set to cluster total resources (as a reference limit for child queue creation)

pkg/scheduler/plugins/capacity/capacity.go
zjj2wry: I agree we should remove the event recording here.

guoqinwill:
> > In the capacity plugin, the root queue should have unlimited default resources, allowing tasks that exceed resource limits to be scheduled normally.
> > https://github.com/volcano-sh/volcano/blob/2090e3b2c7ce912acf4983e2141dafc37864d313/pkg/scheduler/plugins/capacity/capacity.go#L599-L606
> > 
> > This can also resolve the issue you mentioned where reclaim conflicts with enqueue. At the same time, we should reject modifications or updates related to the spec of the root queue, because the root queue is automatically created by Volcano and should not restrict user job submissions based on cluster resources. There has been related discussion on this topic (see [#4662 (comment)](https://github.com/volcano-sh/volcano/pull/4662#discussion_r2486119131)), but it seems that the author did not submit this change. I think your submission can achieve this.
> 
> The scheduler framework actively updates the root queue's `Spec.Deserved` and `Spec.Guarantee.Resource` in
> 
> https://github.com/volcano-sh/volcano/blob/3275f8aace43d7c6dc36ca863e551074825a856e/pkg/scheduler/framework/session.go#L546-L585
> 
> In my use case, I need:
> 
> * **Root queue during scheduling**: Unlimited resources (to avoid blocking jobs)
> * **Root queue quota (guarantee/deserved/capability)**: Set to cluster total resources (as a reference limit for child queue creation)

The update to root when closing the session here is actually unnecessary because the sub-queues of root do not need to check guarantee, deserved, or capability. So each session performs the update, but in reality, it doesn’t have much effect.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -261,6 +297,43 @@ func validateQueueDeleting(queueName string) error {
...
 }
 
+// needsValidateHierarchicalQueue determines if hierarchy resource validation is necessary
+// Returns true only if:
+// - Queue has a parent AND
+// - Either it's a CREATE operation OR resources (capability/deserved/guarantee) changed on UPDATE
+func needsValidateHierarchicalQueue(queue, oldQueue *schedulingv1beta1.Queue, operation admissionv1.Operation) bool {
+	// If queue has no parent, skip hierarchy validation
+	if queue.Spec.Parent == "" {
+		return false
```
guoqinwill, line 303: The filtering and interception conditions here are not quite correct. If the queue's parent is unspecified (default is root) or specified as root, then the validation of it as a child queue upwards and as a parent queue downwards needs to be handled differently. 
1. It does not need to be validated upwards as a child of root, because it should not be restricted by root's quota.
2. If the queue operation is create, there is no need to validate downwards as a parent queue either, since it should not have any child queues at that time.
3. However, if the operation is update, it still needs to be validated downwards as a parent queue. For example, adjusting the quota might make it smaller than its child queues' quotas.

Therefore, the current handling of directly returning false when the parent queue is root is incorrect, because in the third case of update, it allows passing without performing the downward capability check.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -205,8 +216,33 @@ func validateStateOfQueue(value schedulingv1beta1.QueueState, fldPath *field.Pat
...
+		if quantity.Sign() < 0 {
+			errs = append(errs, field.Invalid(fldPath.Child(string(name)),
+				quantity.String(),
+				fmt.Sprintf("resource %s value must be non-negative", name)))
+		}
+	}
+	return errs
+}
+
 func validateResourceOfQueue(resource schedulingv1beta1.QueueSpec, fldPath *field.Path) field.ErrorList {
```
guoqinwill, line 221: I think the judgment here needs to be modified. If 'deserved' is not configured (default is 0), but 'guarantee' > 0 is configured, the current logic will not perform verification here. However, this kind of configuration should also be disallowed. I think you could also add this point. See：
https://github.com/volcano-sh/volcano/blob/276ef22c7d39c9ef2b5dca38ef5d9a3869b12719/pkg/webhooks/admission/queues/validate/validate_queue.go#L262-L266

zjj2wry:
> > > In the capacity plugin, the root queue should have unlimited default resources, allowing tasks that exceed resource limits to be scheduled normally.
> > > https://github.com/volcano-sh/volcano/blob/2090e3b2c7ce912acf4983e2141dafc37864d313/pkg/scheduler/plugins/capacity/capacity.go#L599-L606
> > > 
> > > This can also resolve the issue you mentioned where reclaim conflicts with enqueue. At the same time, we should reject modifications or updates related to the spec of the root queue, because the root queue is automatically created by Volcano and should not restrict user job submissions based on cluster resources. There has been related discussion on this topic (see [#4662 (comment)](https://github.com/volcano-sh/volcano/pull/4662#discussion_r2486119131)), but it seems that the author did not submit this change. I think your submission can achieve this.
> > 
> > 
> > The scheduler framework actively updates the root queue's `Spec.Deserved` and `Spec.Guarantee.Resource` in
> > https://github.com/volcano-sh/volcano/blob/3275f8aace43d7c6dc36ca863e551074825a856e/pkg/scheduler/framework/session.go#L546-L585
> > 
> > In my use case, I need:
> > 
> > * **Root queue during scheduling**: Unlimited resources (to avoid blocking jobs)
> > * **Root queue quota (guarantee/deserved/capability)**: Set to cluster total resources (as a reference limit for child queue creation)
> 
> The update to root when closing the session here is actually unnecessary because the sub-queues of root do not need to check guarantee, deserved, or capability. So each session performs the update, but in reality, it doesn’t have much effect.

Thanks for the detailed explanation! Perhaps it would be more appropriate to address root queue issues in a separate PR. This PR is already ready for review，PTAL. @guoqinwill @hzxuzhonghu @JesseStutler

guoqinwill:
I mentioned a few points about verification interception again this morning, but they don't seem to have been modified yet. Please take a look at these.

zjj2wry:
> I mentioned a few points about verification interception again this morning, but they don't seem to have been modified yet. Please take a look at these.

Are you suggesting introducing a prohibition on root updates in this PR? The prerequisite would be to remove the root update code in volcano, but I don't know if any users are currently using information in the root queue, which could break compatibility.

guoqinwill:
> > I mentioned a few points about verification interception again this morning, but they don't seem to have been modified yet. Please take a look at these.
> 
> Are you suggesting introducing a prohibition on root updates in this PR? The prerequisite would be to remove the root update code in volcano, but I don't know if any users are currently using information in the root queue, which could break compatibility.

No, it's the other suggestions in validate_queue.go. You can take another look at the few comments of mine that are pending.

pkg/webhooks/admission/queues/validate/validate_queue.go
JesseStutler: Yes, we currently have this validation

JesseStutler:
> > > I mentioned a few points about verification interception again this morning, but they don't seem to have been modified yet. Please take a look at these.
> > 
> > 
> > Are you suggesting introducing a prohibition on root updates in this PR? The prerequisite would be to remove the root update code in volcano, but I don't know if any users are currently using information in the root queue, which could break compatibility.
> 
> No, it's the other suggestions in validate_queue.go. You can take another look at the few comments of mine that are pending.

@guoqinwill Did you submit your review comments, the contributor can't see pending review comments if you didn't submit it

JesseStutler:
@zjj2wry Hi, thanks for your contribution, please don't forget to clean and squash your commits, currenly there are too many commits in your PR, you can keep 1-3 clean commits

guoqinwill:
> > > > I mentioned a few points about verification interception again this morning, but they don't seem to have been modified yet. Please take a look at these.
> > > 
> > > 
> > > Are you suggesting introducing a prohibition on root updates in this PR? The prerequisite would be to remove the root update code in volcano, but I don't know if any users are currently using information in the root queue, which could break compatibility.
> > 
> > 
> > No, it's the other suggestions in validate_queue.go. You can take another look at the few comments of mine that are pending.
> 
> @guoqinwill Did you submit your review comments, the contributor can't see pending review comments if you didn't submit it

Sorry, my problem, I forgot to submit it.

pkg/scheduler/plugins/capacity/capacity.go
JesseStutler: I think @hzxuzhonghu 's concern is that we will emit events every session, which is duplicate, but I still think record a event may help admin to debug questions, just as @guoqinwill said

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -205,8 +216,33 @@ func validateStateOfQueue(value schedulingv1beta1.QueueState, fldPath *field.Pat
...
 	errs := field.ErrorList{}
+
+	// Validate resource quantity values are non-negative
+	if len(resource.Capability) != 0 {
+		errs = append(errs, validateResourceQuantityValue(resource.Capability, fldPath.Child("capability"))...)
```
guoqinwill: There's no need for a custom function here. The "k8s.io/kubernetes/pkg/apis/core/validation" package provides a "ValidateResourceQuantityValue" method for normative validation of resource quotas, which can be used directly.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -261,6 +297,43 @@ func validateQueueDeleting(queueName string) error {
...
+	if queue.Spec.Parent == "" {
+		return false
+	}
+
+	// For CREATE operations, always validate
+	if operation == admissionv1.Create {
+		return true
+	}
+
+	// For UPDATE operations, only validate if resources changed
```
guoqinwill: I understand that the specs of a queue can be updated, right? For example, Reclaimable and weight.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: So I understand that the scenario where it directly returns false here should be when queue.name is root, in which case no validation is needed. For all other queues except the root queue, continue with your following checks.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: I believe that validation in the webhook should only check explicit values. Additionally, explicit is better than implicit for understanding. 

# Implicit dependency - hard to understand
grandparent:
  capability: {cpu: 100}
parent:
  capability: {}  # User must "guess" this inherits 100
child:
  capability: {cpu: 50}  # User must trace back to grandparent

# Explicit configuration - self-documenting
grandparent:
  capability: {cpu: 100}
parent:
  capability: {cpu: 100}  # Clear: parent also has 100
child:
  capability: {cpu: 50}   # Clear: 50 < 100

pkg/scheduler/plugins/capacity/capacity.go
zjj2wry: It only prints logs now.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: > This isn't right. If no parent is configured, the default is the root, and it still needs to verify its own capability as a parent queue.

Currently, when creating a queue, if the parent queue is root, there is no validation. During updates, except for the root queue, all other queues have a parent queue, and `validateChildrenConstraints` will validate its capability as a parent queue

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: 1. When the user has not set a capability, we should not proactively display a default value (we should not change the user's configuration like this, and if there are updates later, we won't be able to determine which settings belong to the user and which were filled in by us).
2. However, when the parent queue's capability is not configured, the child queues we create should be able to perceive the capabilities of ancestor and descendant queues, obtaining reasonable range values to allow the user to configure capabilities, rather than preventing all child queues from configuring capabilities. For example:
   1. Create queue a, a.parent="root": capability.cpu=10, capability.mem=""
   2. Create queue b, b.parent="a": capability.cpu="", capability.mem=100Mi
   3. Create queue c, c.parent="b": capability.cpu=5, capability.mem=50Mi

Since the current queue b's CPU dimension is not set (it defaults to 0 when compared), creating c is blocked because capability.c < capability.b. But in reality, b's capability should inherit from queue a. Therefore, we should continue to look up a's capability. The creation of queue c is reasonable and should be allowed. Similarly, when checking downward, we should traverse the child queues. If the current dimension (CPU, memory—different dimensions may result in different final queues, so dimensions need to be distinguished) of the child queue is not set, we should continue to traverse downward.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: > So I understand that the scenario where it directly returns false here should be when queue.name is root, in which case no validation is needed. For all other queues except the root queue, continue with your following checks.

@guoqinwill You're absolutely right! Thanks for catching this. As you pointed out, a queue with parent == "" is semantically a child of root and can have its own child queues. When we update such a queue's resources, we must validate the constraints against its children.I've changed the condition to queue.Name == "root" and added a test case

zjj2wry: By the way, I found that updating a queue's parent field is currently allowed. Is this behavior expected?

guoqinwill: Yes, updating the parent should also be allowed. So after the parent is updated, it should also go through the logic of hierarchy queue resource validation.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: Current behavior in scheduler plugins: Both proportion and capacity have this logic:
```golang
attr.deserved = helpers.Max(attr.deserved, attr.guarantee)
```
Currently, users can create queues with:
```yaml
spec:
  guarantee:
    resource:
      cpu: "10"
  # deserved not set
```
This works because:
- Webhook validation skips the check when len(resource.Deserved) == 0
- Scheduler automatically sets deserved = max(0, 10) = 10 at runtime

If we change the validation to check `len(resource.Guarantee.Resource) != 0`: This configuration would be rejected by the webhook, breaking existing queues. in my cases，i will automatically set deserved = guarantee when deserved is not set but guarantee > 0, maybe we can add it to mutate webhook

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: Thanks for the suggestion! `k8s.io/kubernetes/pkg/apis/core/validation.ValidateResourceQuantityValue` function is from the internal Kubernetes API validation package (`k8s.io/kubernetes/pkg/apis/core/validation`), which is not a stable public API and is generally not recommended for external projects to depend on directly.

pkg/scheduler/plugins/capacity/capacity.go
zjj2wry: After adding the event recording, I found that misconfigured queues were generating too many events. I've added sorting for scalar resources to ensure consistent event messages, which helps with deduplication. However, even with identical events, Kubernetes still updates the Count and LastTimestamp fields, and requests are still sent to the API server on every scheduling cycle. I plan to use Prometheus metrics from the queue for monitoring and alerting instead.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: > This isn't right. If no parent is configured, the default is the root, and it still needs to verify its own capability as a parent queue.

fixed, thanks

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: 1. The `helpers.Max(attr.deserved, attr.guarantee)` you mentioned is the logic used when hierarchical queues are not enabled. After enabling hierarchical queues, this consideration is not included in `buildHierarchicalQueueAttrs`, but I understand that this logic also needs to be added here to accommodate existing queues in the cluster.
2. I still think changing the user's original configuration is not a good practice. If our goal is to have users follow `guarantee <= deserved <= capability`, then a clear reminder for users to be aware of it is sufficient. Unlike capability, deserved and guarantee inherit from the parent queue only if configured; if they are not configured, their value is 0. Therefore, based on the original configuration, users should be made aware that configuring guarantee without configuring deserved is problematic.

hzxuzhonghu:
@zjj2wry please squash

Makefile
```diff
@@ -106,8 +106,27 @@ images:
...
+		--build-arg OPEN_EULER_IMAGE_TAG=${OPEN_EULER_IMAGE_TAG}
+endef
+
+vc-controller-manager-image:
+	$(call build_component_image,controller-manager)
+
+vc-scheduler-image:
+	$(call build_component_image,scheduler)
+
+vc-webhook-manager-image:
```
hzxuzhonghu, line 125: Not related. And these has been implemented via make images

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -596,13 +596,9 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
-	// Check the hierarchical structure of queues
-	err := cp.checkHierarchicalQueue(rootQueueAttr)
-	if err != nil {
-		klog.Errorf("Failed to check queue's hierarchical structure, error: %v", err)
-		return false
-	}
-	klog.V(4).Infof("Successfully checked queue's hierarchical structure.")
+	// checkHierarchicalQueue only logs warnings and never returns errors
+	// to avoid aborting the entire scheduling cycle due to configuration issues
+	cp.checkHierarchicalQueue(rootQueueAttr)
```
hzxuzhonghu, line 601: Seems this is not right, now if no validation enforced at webhook server, then the queue conflicts does influence scheduling at all

pkg/webhooks/router/indexer.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// GetQueuesByParent returns all queues that have the specified parent using the queue parent index.
+// This method leverages the informer's indexer for efficient lookups when available,
+// falling back to listing all queues if the informer is not initialized (e.g., in tests).
+func (asc *AdmissionServiceConfig) GetQueuesByParent(parentName string) ([]*schedulingv1beta1.Queue, error) {
+	// If informer is available, use the efficient index-based lookup
+	if asc.QueueInformer != nil {
+		indexer := asc.QueueInformer.GetIndexer()
+		objs, err := indexer.ByIndex(QueueParentIndexName, parentName)
```
hzxuzhonghu: Instead of calling index here, you can add indexer to the informer by `Informer().AddIndexers`

pkg/webhooks/admission/queues/validate/validate_queue.go
hzxuzhonghu: we have already imported this pkg

zjj2wry:
> @zjj2wry please squash

I'll squash these commits after the review. Keeping them separate now just makes reviewing easier.

Makefile
zjj2wry: I added this target because I want to build only the webhook image for testing purposes. The `make images` command builds all images.

pkg/scheduler/plugins/capacity/capacity.go
zjj2wry: This PR is  enforcing these validations at the webhook server to prevent the entire scheduling system from becoming unavailable due to invalid queue configurations.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: updated

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: Updated. I agree that "explicit is better than implicit" is the right principle here. The Kubernetes spec is a declaration of user intent, and I believe the runtime implicit behavior (helpers.Max adjusting deserved) should be removed.

pkg/webhooks/router/indexer.go
zjj2wry: updated

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -261,6 +293,43 @@ func validateQueueDeleting(queueName string) error {
...
+		}
+		// Check if deserved changed
+		if !equality.Semantic.DeepEqual(queue.Spec.Deserved, oldQueue.Spec.Deserved) {
+			return true
+		}
+		// Check if guarantee changed
+		if !equality.Semantic.DeepEqual(queue.Spec.Guarantee.Resource, oldQueue.Spec.Guarantee.Resource) {
+			return true
+		}
+		// No resource changes, skip validation
```
guoqinwill: As you mentioned earlier, if the parent of the queue changes, it also needs to return true to check whether the new parent meets the quota capacity of the current queue.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: I found 3 places in the e2e tests that only set `guarantee` without `deserved`. While this change is an optimization, it breaks backward compatibility with existing test cases.

Should we update the e2e tests to explicitly set `deserved`, or should we relax the validation to maintain compatibility? I prefer updating the tests to follow the explicit configuration best practice.

guoqinwill: I also prefer updating the e2e tests.

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: updated

pkg/webhooks/admission/queues/validate/validate_queue.go
zjj2wry: Looking at the scheduler logic, **guarantee/deserved** default to **0** when empty, but capability is **inherited**. This is great for keeping configs simple, But troubleshooting is difficult because the spec hides the real capability, forcing recursive lookups of parent queues. We should probably show the effectiveCapability in the status? Also, since the scheduler always takes the minimum value across the hierarchy anyway, the fact that a child's capability might exceed its parent's doesn't really affect anything. Is that validation even required?"

guoqinwill: Yes. I understand that performing recursive checks on the capability is quite complex. If you think it can't be completed in time this time, then we won't add it in this round of modifications. We can follow up and discuss it later.

zhaizhicheng:
Finally, someone is fixing this bug. Looking forward to it being merged into the main branch soon

JesseStutler:
/approve
Currently I'm ok with current change, allow @guoqinwill do more fine grained validation in the future
/cc @guoqinwill @hzxuzhonghu

PR #4865, opened 2025-12-24
fix(capacity): fix queue share calculation for best-effort queues

guoqinwill:
Previously, queues without deserved configuration (best-effort queues)
would always have share = 0 regardless of their allocated resources or
pending requests. This caused issues where:

1. Best-effort queues with allocated resources (share = 0) would be
   prioritized over queues with deserved but no allocated (share = 0),
   leading to incorrect scheduling order.

2. After reclaim evicted pods from best-effort queues, they would have
   share = 0 again, potentially causing reclaim thrashing where best-effort
   queues keep getting scheduled and then reclaimed.

This commit fixes the issue by:

Assign share = 1 to any queue without deserved, regardless of current allocated or pending.
Ensures all best-effort queues are scheduled only after any with-deserved queue under quota.
Preserves tie-breaking so with-deserved takes precedence if share is equal.

This ensures proper priority ordering and prevents reclaim thrashing.

Fixes #

```release-note

```

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -509,15 +509,7 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
 			return int(rv.Queue.Spec.Priority) - int(lv.Queue.Spec.Priority)
 		}
 
-		if cp.queueOpts[lv.UID].share == cp.queueOpts[rv.UID].share {
-			return 0
-		}
-
-		if cp.queueOpts[lv.UID].share < cp.queueOpts[rv.UID].share {
-			return -1
-		}
-
-		return 1
+		return cp.compareShareWithDeserved(cp.queueOpts[lv.UID], cp.queueOpts[rv.UID])
 	})
 }
 
@@ -650,14 +642,7 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 		} else if !lvLeaf && rvLeaf {
 			return 1
 		} else if !lvLeaf && !rvLeaf {
-			if cp.queueOpts[lv.UID].share == cp.queueOpts[rv.UID].share {
-				return 0
-			}
-
-			if cp.queueOpts[lv.UID].share < cp.queueOpts[rv.UID].share {
-				return -1
-			}
-			return 1
+			return cp.compareShareWithDeserved(cp.queueOpts[lv.UID], cp.queueOpts[rv.UID])
 		}
 
 		lvAttr := cp.queueOpts[lv.UID]
@@ -672,15 +657,7 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 			rvParentID = rvAttr.ancestors[level+1]
 		}
 
-		if cp.queueOpts[lvParentID].share == cp.queueOpts[rvParentID].share {
-			return 0
-		}
-
-		if cp.queueOpts[lvParentID].share < cp.queueOpts[rvParentID].share {
-			return -1
-		}
-
-		return 1
+		return cp.compareShareWithDeserved(cp.queueOpts[lvParentID], cp.queueOpts[rvParentID])
 	})
 
 	ssn.AddVictimQueueOrderFn(cp.Name(), func(l, r, preemptor interface{}) int {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -843,6 +820,28 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) error {
...
+	if lattr.share == rattr.share {
+		lHasDeserved := !lattr.deserved.IsEmpty()
+		rHasDeserved := !rattr.deserved.IsEmpty()
+		if lHasDeserved && !rHasDeserved {
+			return -1
+		}
+		if !lHasDeserved && rHasDeserved {
+			return 1
+		}
+		return 0
```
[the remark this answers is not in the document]
guoqinwill: yes, I agree with you

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1002,6 +1001,25 @@ func (s *capacityState) Clone() fwk.StateData {
...
+	//
+	// When share values are equal, QueueOrderFn uses tie-breaking logic to prioritize
+	// queues with deserved over best-effort queues. This further prevents reclaim thrashing.
+	//
+	// The semantics here are:
+	//   - no deserved (best-effort) -> share = 1 (always goes after share<1)
+	//   - with deserved, share < 1 -> prioritized
+	//   - with deserved, share >= 1 -> equal priority with best-effort
+	if attr.deserved.IsEmpty() {
+		attr.share = 1
```
wangyang0616, line 1019: Suppose there are two queues in the cluster, namely Queue A and Queue B. Queue A has no deserved value set, while Queue B has its deserved value set to 2 CPU cores (2C). The resources of the already scheduled tasks in both queues are 4 CPU cores (4C). In this case, Queue A's share value is 1 and Queue B's share value is 2. Is my understanding correct?

guoqinwill: Yes. But in this scenario, QueueB will not reclaim QueueA. Only the order of allocation will be affected. The main purpose of this PR is to address the potential issue of cyclical eviction caused by uncertain queue scheduling after queue reclamation.

wangyang0616: If the deserved value is not set and we define such a queue as besteffort, then the resource priority of this queue should be lower than that of queues with the deserved value configured. Could the current processing logic give rise to the following three unreasonable scenarios:

1. When Queue A and Queue B submit new tasks simultaneously, Queue A has a higher priority than Queue B, which is inconsistent with the besteffort definition of Queue A.
2. When a new Queue C is created, it reclaims resources from Queue B first during its resource reclamation process, and the resources of Queue A will never be reclaimed.
3. When cluster resources are insufficient, the pending jobs in Queue A will reclaim resources from Queue B.

wangyang0616: Therefore, I'm wondering if we could assign an extremely small default value (e.g., 0.000001) to queues that have no deserved value set. This way, the logic would be consistent.

guoqinwill: I understand the situations you mentioned: 
1. Although queue A does not have a deserved value set (default is 0) and queue B does, at this time their usage has both exceeded their own deserved values. Therefore, the scheduling order does not need to strictly follow the rules for unsatisfied deserved values, because the amount exceeding the deserved value is effectively overused and can be reclaimed.
2. When reclaiming queue resources, it does not judge based on the share value, but directly checks whether the queue allocation exceeds the deserved value. For queue A and queue B, since they exceed their deserved values, they can both be reclaimed.
3. When submitting jobs to queue A, because queue A’s deserved value is 0, it will not try to reclaim resources from queue B.

guoqinwill: Let me explain the situations you mentioned:
1. Although queue A does not have a deserved value set (default is 0) and queue B does, at this time their usage has both exceeded their own deserved values. Therefore, the scheduling order does not need to strictly follow the rules for unsatisfied deserved values, because the amount exceeding the deserved value is effectively overused and can be reclaimed.
2. When reclaiming queue resources, it does not judge based on the share value, but directly checks whether the queue allocation exceeds the deserved value. For queue A and queue B, since they exceed their deserved values, they can both be reclaimed.
3. When submitting jobs to queue A, because queue A’s deserved value is 0, it will not try to reclaim resources from queue B.

So the current changes won't cause the problems you described.

wangyang0616: Got it. Thanks.

PR #4868, opened 2025-12-24
Force minAvailable to match minPartitions*partitionSize to avoid semantic inconsistencies.

### LINKED ISSUE #4867: The mismatch between minAvailable and minPartitions*partitionSize causes the job to remain in pending state.
### Description

When I start the job with the following configuration, no pods are created, and the vcjob remains in the PENDING state.

```
kind: Queue
metadata:
  name: test
spec:
  weight: 1
  reclaimable: false
  capability:
    cpu: "20"
    mem: "400Gi"

---

kind: Job
metadata:
  name: vcjob-with-subgroups
  namespace: default
spec:
  schedulerName: volcano
  queue: test
  networkTopology:
    mode: hard
    highestTierAllowed: 1
  tasks:
    - name: worker
      replicas: 12
      partitionPolicy:
        totalPartitions: 4
        partitionSize: 3
        minPartitions: 2
      template:
        spec:
          restartPolicy: OnFailure
          containers:
            - name: worker
              image: nginx:latest
              resources:
                requests:
                  cpu: "2"
                  memory: "40Gi"
```

I discovered the reason is that `task.minAvailable` couldn't be specified, so it was set to `task.replicas`, which is 12. However, in this example, the maximum supported number of replicas is 10 for queue test. But with `partitionSize` at 3 and `minPartitions` at 2, it should create 12 pods, and schedule 9 pods. However, because `task.minAvailable` was set to 12, the job couldn't be inqueue, causing this problem.
I think `minAvailable` should be forced to equal `minPartitions * partitionSize` to avoid semantic inconsistencies.

Just apply the yaml

Expect create 12 pods, and schedule 9 pods.

master

_No response_

zhengchenyu:

Force minAvailable to match minPartitions*partitionSize to avoid semantic inconsistencies.

Fixes #4867 

The following changes:
* Allow setting `minAvailable`, but if set, ensure that `minAvailable = partitionSize * minPartitions`.
* If `minAvailable` is not set, set it to `partitionSize * minPartitions`.
* Fixed the issue where `MinSubGroups` was not set.

NONE

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -1083,6 +1083,7 @@ func getSubGroupPolicy(taskSpec batch.TaskSpec) scheduling.SubGroupPolicySpec {
 	subGroupPolicy := scheduling.SubGroupPolicySpec{
 		Name:         taskSpec.Name,
 		SubGroupSize: &taskSpec.PartitionPolicy.PartitionSize,
+		MinSubGroups: &taskSpec.PartitionPolicy.MinPartitions,
 	}
 
 	// Set LabelSelector
--- a/pkg/webhooks/admission/jobs/mutate/mutate_job.go
+++ b/pkg/webhooks/admission/jobs/mutate/mutate_job.go
@@ -202,10 +202,15 @@ func mutateSpec(tasks []v1alpha1.TaskSpec, basePath string, job *v1alpha1.Job) *
 			tasks[index].Template.Spec.DNSPolicy = v1.DNSClusterFirstWithHostNet
 		}
 
-		if tasks[index].MinAvailable == nil && (tasks[index].PartitionPolicy == nil || tasks[index].PartitionPolicy.MinPartitions == 0) {
+		if tasks[index].MinAvailable == nil {
 			patched = true
-			minAvailable := tasks[index].Replicas
-			tasks[index].MinAvailable = &minAvailable
+			if tasks[index].PartitionPolicy != nil {
+				minAvailable := tasks[index].PartitionPolicy.MinPartitions * tasks[index].PartitionPolicy.PartitionSize
+				tasks[index].MinAvailable = &minAvailable
+			} else {
+				minAvailable := tasks[index].Replicas
+				tasks[index].MinAvailable = &minAvailable
+			}
 		}
 
 		if tasks[index].MaxRetry == 0 {
--- a/pkg/controllers/job/job_controller_actions_test.go
+++ b/pkg/controllers/job/job_controller_actions_test.go
@@ -28,6 +28,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
 
 	"volcano.sh/apis/pkg/apis/batch/v1alpha1"
 	schedulingapi "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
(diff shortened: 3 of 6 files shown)


pkg/controllers/job/job_controller_actions.go
```diff
@@ -1106,5 +1106,9 @@ func getSubGroupPolicy(taskSpec batch.TaskSpec) scheduling.SubGroupPolicySpec {
 		}
 		subGroupPolicy.NetworkTopology = nt
 	}
+
+	if taskSpec.PartitionPolicy != nil && taskSpec.PartitionPolicy.MinPartitions > 0 {
+		subGroupPolicy.MinSubGroups = &taskSpec.PartitionPolicy.MinPartitions
```
3sunny: Line 1111 can be moved below line 1085, and this condition can be removed.

zhengchenyu: Thanks for your review! I have fix it.

pkg/webhooks/admission/jobs/mutate/mutate_job.go
```diff
@@ -201,11 +201,22 @@ func mutateSpec(tasks []v1alpha1.TaskSpec, basePath string, job *v1alpha1.Job) *
 			patched = true
 			tasks[index].Template.Spec.DNSPolicy = v1.DNSClusterFirstWithHostNet
 		}
+		if tasks[index].PartitionPolicy != nil {
+			if tasks[index].PartitionPolicy.MinPartitions == 0 {
+				patched = true
+				tasks[index].PartitionPolicy.MinPartitions = tasks[index].PartitionPolicy.TotalPartitions
```
3sunny: This assignment step is not necessary, right? It will change the default value of MinPartitions. Before the change, the default value was 0. When resources were insufficient for PartitionSize, jobs could be scheduled as long as they met the job's minAvailable requirement. After the change, jobs need to meet both PartitionSize and the job's minAvailable requirements to be scheduled.

pkg/webhooks/admission/jobs/mutate/mutate_job.go
```diff
@@ -201,11 +201,22 @@ func mutateSpec(tasks []v1alpha1.TaskSpec, basePath string, job *v1alpha1.Job) *
...
+			}
+		}
 
-		if tasks[index].MinAvailable == nil && (tasks[index].PartitionPolicy == nil || tasks[index].PartitionPolicy.MinPartitions == 0) {
+		if tasks[index].MinAvailable == nil {
 			patched = true
-			minAvailable := tasks[index].Replicas
-			tasks[index].MinAvailable = &minAvailable
+			if tasks[index].PartitionPolicy != nil {
+				minAvailable := tasks[index].PartitionPolicy.MinPartitions * tasks[index].PartitionPolicy.PartitionSize
```
3sunny, line 208: After adding the MinPartitions field, try to avoid using the task's minAvailable, because the task's minAvailable does not actually serve as a gang interception mechanism.

pkg/webhooks/admission/jobs/mutate/mutate_job.go
zhengchenyu: @3sunny 
You're right. The purpose of this is to ensure that the logic here is similar to that of `replicas` and `minAvailable`. That is, if `minAvailable` is not set, it will be set to `replicas`. However, `minAvailable` is a pointer, and we can determine whether it's set by checking if it's `nil`. But `MinPartitions` is an int, so perhaps setting its default value to -1 would be more reasonable. 
What do you think? Or we can temporarily revert here to maintain the logic.

pkg/webhooks/admission/jobs/mutate/mutate_job.go
zhengchenyu: @3sunny
`minAvailable` is affecting `inqueue`, preventing task scheduling because `minAvailable` defaults to `replicas` if not manually set.
In this pr, we can set or not set `minAvailable`. It simply means that stricter checks have been added.
If we do not set `minAvailable`, it means `minAvailable` does not server gang interception mechanism.
If we set `minAvailable`, we must maintain logical consistency

Alternatively, we could solve the problem in the following way, but I personally feel it's unnecessary.
* Retain the logic that prevents setting both `minAvailable` and `minPartitions` simultaneously.
* Modify the setting logic for `minAvailable` to `min(replicas, minPartitions*PartitionSize)`.

pkg/webhooks/admission/jobs/mutate/mutate_job.go
zhengchenyu: @3sunny I have removed this part to maintain logic.

pkg/webhooks/admission/jobs/mutate/mutate_job.go
3sunny: I think your revision is also reasonable.

3sunny:
Add signature information (sign-off) when submitting. Can you merge the commits?

zhengchenyu:
> Add signature information (sign-off) when submitting. Can you merge the commits?

Sorry, I forget it! For now, I have add sign off.

wangyang0616:
@zhengchenyu  Please squash the commit records with no meaningful content, and then this PR can be marked as LGTM.

zhengchenyu:
> @zhengchenyu Please squash the commit records with no meaningful content, and then this PR can be marked as LGTM.

OK! I already squash the commits.

Issue #4809, opened 2025-12-12
vgpu resource information error


sailorvii:
### Description

The used gpu memory and core were calculated twice.

### Steps to reproduce the issue

At allocate function:
310 func (gs *GPUDevices) Allocate(kubeClient kubernetes.Interface, pod *v1.Pod) error {
311     if VGPUEnable {
312         klog.V(4).Infoln("hami-vgpu DeviceSharing:Into AllocateToPod", pod.Name)
313         fit, device, _, err := checkNodeGPUSharingPredicateAndScore(pod, gs, false, "")
314         if err != nil || !fit {
315             klog.ErrorS(err, "Failed to allocate vgpu task", "pod", pod.Name)
316             return err
317         }
318         if NodeLockEnable {
319             nodelock.UseClient(kubeClient)
320             err = nodelock.LockNode(gs.Name, DeviceName)
321             if err != nil {
322                 return errors.Errorf("node %s locked for %s hamivgpu lockname %s", gs.Name, pod.Name, err.Error())
323             }
324         }
325
326         annotations := make(map[string]string)
327         annotations[AssignedNodeAnnotations] = gs.Name
328         annotations[AssignedTimeAnnotations] = strconv.FormatInt(time.Now().Unix(), 10)
329         annotations[AssignedIDsAnnotations] = encodePodDevices(device)
330         annotations[AssignedIDsToAllocateAnnotations] = annotations[AssignedIDsAnnotations]
331
332         annotations[DeviceBindPhase] = "allocating"
333         annotations[BindTimeAnnotations] = strconv.FormatInt(time.Now().Unix(), 10)
334         // To avoid that the pod allocated info updating latency, add it first
335         gs.addResource(annotations, pod)

The line  313 will add the used memory and core the real gpu device structure, the line 335 will also add them. Here line 335 should only add the pod information to the gpu structure, not add the used memory/core.

### Describe the results you received and expected

The result is that adding one pod will calculate the used gpu memory & core twice.
Expected result: resource only calculates once.

### What version of Volcano are you using?

1.12

JesseStutler:
@sailorvii Hi, would you fix it?

sailorvii:
Yes, I'll raise a PR.

archlitchi:
Is thie re-calculation raises a problem, or it's a optimization?

falconlee236:
Hi @archlitchi , @JesseStutler 

I reproduced this on Volcano v1.14.1.

[...]

PR #4870, opened 2025-12-24
Fix gpu resource error.

sailorvii:
At allocate state, the function checkNodeGPUSharingPredicateAndScore will add the used mem/core to the structure. For avoiding the pod allocated info updating latency, we use addResource it to add the pod the pod map. But it also add the used mem/core. To address this, we new a function that only add pod the map.

#### What type of PR is this?
Bug fix

Fixes # https://github.com/volcano-sh/volcano/issues/4809

```release-note

```


archlitchi:
please sign-off your commit

PR #4872, opened 2025-12-24
Fixed issue where jobs with subgroups but not hard networkTopology.mode could not be scheduled.

### LINKED ISSUE #4871: When networkTopology.mode is not hard, but subgroups exist, scheduling is not possible.
### Description

When networkTopology.mode is not hard, but subgroups exist, scheduling is not possible.

Create job like below yaml:

```
kind: Job
metadata:
  name: vcjob-with-subgroups
  namespace: default
spec:
  schedulerName: volcano
  queue: test
#  networkTopology:
#    mode: soft
#    highestTierAllowed: 1
  tasks:
    - name: worker
      replicas: 2
      partitionPolicy:
        totalPartitions: 2
        partitionSize: 1
        minPartitions: 2
      template:
        spec:
          restartPolicy: OnFailure
          containers:
            - name: worker
              image: nginx:latest
              resources:
                requests:
                  cpu: "2"
                  memory: "40Gi"
```

We see that the pod cannot be scheduled: 
```
[user@host dir]# kubectl get vcjob
NAME                   STATUS    MINAVAILABLE   RUNNINGS   AGE
vcjob-with-subgroups   Pending   2                         5s
[user@host dir]# kubectl get pg
NAME                                                        STATUS    MINMEMBER   RUNNINGS   AGE
vcjob-with-subgroups-fd291270-29cb-4f7b-9fb6-62ff5dc3354b   Inqueue   2                      8s
[user@host dir]# kubectl get pod
NAME                            READY   STATUS    RESTARTS   AGE
vcjob-with-subgroups-worker-0   0/1     Pending   0          12s
vcjob-with-subgroups-worker-1   0/1     Pending   0          12s
```

The pod is running.

master

The scheduler has the following logs:
```
E1224 06:03:19.760817       1 allocate.go:330] "Can not find default subJob or tasks for job" job="default/vcjob-with-subgroups-fd291270-29cb-4f7b-9fb6-62ff5dc3354b" subJobExist=false tasksExist=false
```

Based on the logs, the cause is easily identified: For jobs with subgroups but no hard topology, scheduling will enter the [else](https://github.com/volcano-sh/volcano/blob/06457b06bbf45a8bf98d0b6cccbce65f592b14d2/pkg/scheduler/actions/allocate/allocate.go#L316). However, due to the existence of subgroups, `job.SubJob` and `actx.tasksNoHardTopology` will actually be indexed using their own SubJobID. Therefore, `sjExist` and `tasksExist` will both return false, causing scheduling failure.

Solution: 
* Use `allocateForJob` for scheduling when a hard topology is set or a subgroup exists.

Performance: 

Although using `allocateForJob` will enter the `Network Topology` scheduling logic, using two nested loops. However, since no hard topology is used, meaning that the two nested loops will not get caught in the `Network Topology` logic, there are no performance issues.

zhengchenyu:

When networkTopology.mode is not hard, but subgroups exist, scheduling is not possible.

Fixes #4871 

Solution:

* Use allocateForJob for scheduling when a hard topology is set or a subgroup exists.

Performance: Although using allocateForJob will enter the Network Topology scheduling logic, using two nested loops. However, since no hard topology is used, meaning that the two nested loops will not get caught in the Network Topology logic, there are no performance issues.

NONE

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -190,7 +190,7 @@ func (alloc *Action) buildAllocateContext() *allocateContext {
 		actx.jobWorksheet[job.UID] = worksheet
 
 		// job without any hard network topology policy and without SubGroupPolicy use actx.tasksNoHardTopology
-		if !job.ContainsHardTopologyOrSubGroupPolicy() {
+		if !needNetworkTopologyAwareScheduling(job) {
 			if subJobWorksheet, exist := worksheet.subJobWorksheets[job.DefaultSubJobID()]; exist {
 				actx.tasksNoHardTopology[job.UID] = subJobWorksheet.tasks
 			}
@@ -297,7 +297,7 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
 		job := jobs.Pop().(*api.JobInfo)
 		updateJobTier(ssn.HyperNodeTierNameMap, job)
 
-		if job.ContainsHardTopologyOrSubGroupPolicy() {
+		if needNetworkTopologyAwareScheduling(job) {
 			jobWorksheet := actx.jobWorksheet[job.UID]
 
 			klog.V(3).InfoS("Try to allocate resource for job with SubGroupPolicy or hard topology", "queue", queue.Name, "job", job.UID,
@@ -800,6 +800,14 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
 		return api.NewFitErrWithStatus(task, node, statusSets...)
 	}
 	return alloc.session.PredicateForAllocateAction(task, node)
+}
+
+// needNetworkTopologyAwareScheduling returns whether the job need network topology aware scheduling.
+func needNetworkTopologyAwareScheduling(ji *api.JobInfo) bool {
+	if hard, _ := ji.IsHardTopologyMode(); hard || ji.ContainsSubJobPolicy() {
+		return true
+	}
+	return false
 }
 
 func (alloc *Action) UnInitialize() {}
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -1305,14 +1305,6 @@ func (ji *JobInfo) ContainsHardTopologyInSubJob() bool {
 	return false
 }
 
-// ContainsHardTopologyOrSubGroupPolicy returns whether the job contain hard network topology or subJob policy
-func (ji *JobInfo) ContainsHardTopologyOrSubGroupPolicy() bool {
-	if hard, _ := ji.IsHardTopologyMode(); hard || ji.ContainsSubJobPolicy() {
-		return true
-	}
-	return false
-}
-
 // ContainsNetworkTopologyInSubJob returns whether the subJobs in the job contain network topology
 func (ji *JobInfo) ContainsNetworkTopologyInSubJob() bool {
 	if ji.PodGroup == nil {


pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -296,10 +296,11 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
 
 		job := jobs.Pop().(*api.JobInfo)
 		updateJobTier(ssn.HyperNodeTierNameMap, job)
-		if job.ContainsHardTopology() {
+
+		if job.ContainsHardTopologyOrSubGroupPolicy() {
```
ouyangshengjia:
```suggestion
func needNetworkTopologyAwareScheduling(ji *api.JobInfo) {
	if hard, _ := ji.IsHardTopologyMode(); hard || ji.ContainsSubJobPolicy() {
		return true
	}
	return false
}
```
The function name `ContainsHardTopologyOrSubGroupPolicy` is too complex. I suggest extracting a utility function directly in the file `allocate.go`. And remove the func `JobInfo.ContainsHardTopology`.

zhengchenyu: The function name is indeed a bit complicated, I have fix it.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -801,4 +802,12 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
 	return alloc.session.PredicateForAllocateAction(task, node)
 }
 
+// needNetworkTopologyAwareScheduling returns whether the job need network topology aware scheduling.
+func needNetworkTopologyAwareScheduling(ji *api.JobInfo) bool {
```
JesseStutler, line 806: The name of this func still looks strange. Does a podgroup having a subGroupPolicy (but without network policy specified) necessarily mean that networkTopologyAwareScheduling is required? @ouyangshengjia

JesseStutler: Or, should a podgroup with subgroups but no network specified still jump into the else block?

ouyangshengjia: In fact, `alloc.allocateForJob` includes two scheduling capabilities: "network topology constraints" and "subgroup affinity policies". Even if we only need the subgroup scheduling capability(such as subgroup-level gang scheduling), we still need to execute this branch.
Currently, in the user manual, I have categorized both of these capabilities under network topology-aware scheduling. I am not sure whether it is necessary to clearly distinguish between these two capabilities.
https://github.com/volcano-sh/volcano/blob/a02801bd1933aae87ac1d677c73666eb028ef707/docs/user-guide/how_to_use_network_topology_aware_scheduling.md?plain=1#L41
https://github.com/volcano-sh/volcano/blob/a02801bd1933aae87ac1d677c73666eb028ef707/docs/user-guide/how_to_use_network_topology_aware_scheduling.md?plain=1#L61

ouyangshengjia: I think we need to clarify whether the "subgroup affinity policies" should be part of the "network topology-aware scheduling" capabilities or an independent capability.

JesseStutler: Yes, the subgroup design doesn't seem to be strongly tied to network topology-aware scheduling. If I only need two levels of gang scheduling capability, then I don't need to go through the network topology scheduling logic. The question is whether we need to use if-else statements to jump into different logics, or whether we can unify them? And if we unify them, what the performance impact would be?

zhengchenyu: I think unifying them has had some impact, but not a very large. For soft mode, hyperNodeToAllocate is always root, it will not trigger too many calls. https://github.com/volcano-sh/volcano/blob/a02801bd1933aae87ac1d677c73666eb028ef707/pkg/scheduler/framework/session_plugins.go#L1066 and https://github.com/volcano-sh/volcano/blob/a02801bd1933aae87ac1d677c73666eb028ef707/pkg/scheduler/framework/session_plugins.go#L1087

ouyangshengjia: Theoretically, `alloc.allocateForJob` can be compatible with the scheduling of normal jobs. If needed, it is easy to unify these two scheduling branches at the code level. However, compared to the processing of the "else" branch, it will introduce an additional `statement` "allocate-rollback" due to the dry run mechanism. Perhaps we can avoid this meaningless "allocate-rollback" when there is only one HyperNode in processing.

JesseStutler: > Theoretically, `alloc.allocateForJob` can be compatible with the scheduling of normal jobs. If needed, it is easy to unify these two scheduling branches at the code level. However, compared to the processing of the "else" branch, it will introduce an additional `statement` "allocate-rollback" due to the dry run mechanism. Perhaps we can avoid this meaningless "allocate-rollback" when there is only one HyperNode in processing.

So from this perspective, we could actually unify the two cases in one step, but we would need to do some performance optimizations. I think the current changes in this PR are strange to me. If the logic of the subgroup is simply moved to the if block just to make it work...

Tau721: Yes, I also think that the relationship between the subgroup affinity scheduling and the network topology-aware scheduling is currently unclear, making it hard to design the code logic. This ``if-else`` block is a typical example. So, I think clarifying the two features and their relationship is much more important.

zhengchenyu: @JesseStutler @ouyangshengjia @Tau721 
I think subgroup scheduling and network topology-aware scheduling should be functionally separated. If the proposal at https://github.com/volcano-sh/volcano/issues/4907 is accepted, then these two functions should be separated even more. Subgroups can serve as an organization method for subjobs, while network topology-aware scheduling should be a feature attached to them.

Regarding this PR, the function name is indeed a bit strange. Perhaps `allocateForJob` in the `if` statement should be the default. I think the `if` and `else` statements should be swapped, and the function for the `if` condition should be set to `fastScheduling`, with the condition being `not ji.IsHardTopologyMode(); && not ji.ContainsSubJobPolicy()`.

JesseStutler: I agree to fix the functionality first, but we do need further restructuring, the relationships are quite complicated right now. Consider that this fix now needs to be prioritized, we will inherit your changes and perform an emergency fix.

JesseStutler:
Already inherit it in https://github.com/volcano-sh/volcano/pull/5038 and keep your sign off record, thanks for your contribution

PR #4876, opened 2025-12-25
feat: optimize PodGroup listing with FieldSelector for k8s v1.31+

xu-wentao:
- Add SupportsCustomResourceFieldSelectors to detect k8s version support
- Use FieldSelector for efficient PodGroup filtering in v1.31+
- Fallback to old method for older k8s versions
- Add comprehensive tests for version detection

Fixes #

```release-note

```

### DIFF
--- a/pkg/cli/queue/get.go
+++ b/pkg/cli/queue/get.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	"volcano.sh/apis/pkg/client/clientset/versioned"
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
/hold

This is not campatable with k8s version lower than 1.30

pkg/cli/queue/get.go
```diff
@@ -65,16 +66,39 @@ func GetQueue(ctx context.Context) error {
...
+	clientset := kubernetes.NewForConfigOrDie(config)
+	supportsFieldSelector, err := util.SupportsCustomResourceFieldSelectors(clientset)
 	if err != nil {
-		return fmt.Errorf("failed to list podgroup for queue %s with err: %v", getQueueFlags.Name, err)
+		// If version check fails, fall back to the old method
+		supportsFieldSelector = false
+	}
+
+	if supportsFieldSelector {
+		// Use fieldSelector for k8s v1.31+ where CustomResourceFieldSelectors is enabled by default
```
LiZhenCheng9527:
```suggestion
		// Use fieldSelector for k8s >= v1.31 where CustomResourceFieldSelectors is enabled by default
```

hzxuzhonghu:
@Xu-Wentao With this patch, the cli can meet an error in k8s version >1.31 if the crd does not support field selector(default now)

A more smooth way can be this:

generate two crd files:

```
├── templates/
│   ├── crds-v1-31.yaml  <-- Version with SelectableFields
│   ├── crds-legacy.yaml <-- Version for older clusters
```

crds-v1-31.yaml  with 

```
{{- if ge .Capabilities.KubeVersion.Minor "31" }}
```

And crds-legacy.yaml with 

```
{{- if lt .Capabilities.KubeVersion.Minor "31" }}
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: myresources.example.com
spec:
  # Standard CRD without selectableFields
{{- end }}
```

hzxuzhonghu:
@xu-wentao I guess we donot declare supported seletable fields in crd now,  before this pr, we should add `// +kubebuilder:selectablefield:JSONPath=` in podgroup api?

PR #4889, opened 2025-12-29
Enhance network-topology-aware plugin to provide hypernode-level binpacking for normal pods

### LINKED ISSUE #4869: network-topology-aware should provide hypernode-level binpacking for normal pods

Currently, the network-topology-aware plugin provides hypernode-level binpacking capability for jobs configured with a network topology. However, it fails to implement this feature for normal pods, i.e., the tasks of jobs that are not configured with any network topology. As a result, these tasks may be scheduled to all hypernodes in the cluster, leading to severe resource fragmentation and waste among hypernodes.

It is better to modify the networkTopologyAwarePlugin.batchNodeOrderFn() function to score the nodes for the normal pods. Specifically, for each node and each pod, all resource types requested by the job and all hypernodes that the node belongs to should be considered with customized weights. Then, a comprehensive hypernode-level binpacking score should be calculated, and the pod should be scheduled to the node with the highest score. Besides, an additional argument should be added to control whether or not to enable this feature.

### Additional context

_No response_

### Documentation Updates

Tau721:
#### What type of PR is this?
/kind feature

Currently, the network-topology-aware plugin provides hypernode-level binpacking capability for jobs configured with a network topology. However, it fails to implement this feature for normal pods, i.e., the tasks of jobs that are not configured with any network topology. As a result, these tasks may be scheduled to all hypernodes in the cluster, leading to severe resource fragmentation and waste among hypernodes.

To this end, this PR tries to  enhance the network-topology-aware plugin to provide hypernode-level binpacking for normal pods.

Fixes #4869

```release-note
The following two plugin arguments are added:
1. hypernode.binpack.normal-pod.enable: type of bool, used to control whether or not to enable hypernode-level binpacking for pods without network topology.
2. hypernode.binpack.normal-pod.fading: type of float64, used to control the tier weight of each hypernode, i.e., hypernode tier weight = math.Pow(fading, (tier-1)).
```


wangyang0616:
Please supplement the corresponding capability scenarios and implementation descriptions in the design document.
Refer to [Network Topology Aware Scheduling.md](https://github.com/volcano-sh/volcano/blob/master/docs/design/Network%20Topology%20Aware%20Scheduling.md)

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -45,25 +46,42 @@ const (
...
 )
 
 type networkTopologyAwarePlugin struct {
 	// Arguments given for the plugin
 	pluginArguments framework.Arguments
 	weight          *priorityWeight
+	*normalPodConfig
 	*hyperNodesTier
+	// hyperNodeResourceCache stores the resource status of hypernodes to avoid repeated calculation: hypernode -> resource -> status
+	hyperNodeResourceCache map[string]map[corev1.ResourceName]*resourceStatus
```
ouyangshengjia: Considering that this cache may be used frequently in future evolutions, it is recommended to fully initialize this cache in `OnSessionOpen` and replace the current lazy calculation implementation.

ouyangshengjia: Perhaps issue https://github.com/volcano-sh/volcano/issues/4899 could reuse this cache to pre-filtering HyperNodes during scheduiling.

Tau721: I agree with you, and I have fixed it.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -270,65 +346,155 @@ func (nta *networkTopologyAwarePlugin) getHyperNodeBinPackingScore(subJob *api.S
...
+		tierWeights[tier] = tierWeight
+	}
+	if totalTierWeight <= 0 {
 		return nodeScores, nil
 	}
 
+	for _, node := range nodes {
+		totalScore := 0.0
+		for tier := nta.hyperNodesTier.minTier; tier <= nta.hyperNodesTier.maxTier; tier++ {
+			tierScore := FullScore
```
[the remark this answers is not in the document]
Tau721: When some node doesn't have a hypernode at the n-th tier, we can simply but equivalently consider that this node has a virtual hypernode at the n-th tier and its resources are fully utilized. Hence, we set its binpacking score to `FullScore`.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -141,48 +169,63 @@ func calculateWeight(args framework.Arguments) *priorityWeight {
...
+	if len(n.weight.HyperNodeBinPackingResources) == 0 {
+		msg = append(msg, "no extend resources")
 	} else {
-		for name, weight := range w.HyperNodeBinPackingResources {
+		for name, weight := range n.weight.HyperNodeBinPackingResources {
 			msg = append(msg, fmt.Sprintf("%s[%d]", name, weight))
 		}
 	}
+	msg = append(msg, fmt.Sprintf("%s[%t]", HyperNodeBinPackNormalPodEnable, n.normalPodConfig.hyperNodeBinPackingEnable),
+		fmt.Sprintf("%s[%f]", HyperNodeBinPackNormalPodFading, n.normalPodConfig.hyperNodeBinPackingFading))
```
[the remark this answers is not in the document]
Tau721: Fixed.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -229,10 +272,43 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
...
+				if ssn.RealNodesSet[hyperNode].Has(node) {
+					if _, foundHyperNode := nta.hyperNodeResourceCache[hyperNode]; foundHyperNode {
+						for _, resource := range task.Resreq.ResourceNames() {
+							if _, foundResource := nta.hyperNodeResourceCache[hyperNode][resource]; foundResource {
+								nta.hyperNodeResourceCache[hyperNode][resource].used += task.Resreq.Get(resource)
+							}
+						}
+					}
+				}
+			}
```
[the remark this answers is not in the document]
Tau721: Actually, the number of nodes is usually much larger than that of hypernodes, and the AllocateFunc or DeallocateFunc usually only uses the information of a few nodes. Therefore, caching the `nodeToHypernodes` map is considered not worth it.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -45,25 +46,42 @@ const (
...
 
 type networkTopologyAwarePlugin struct {
 	// Arguments given for the plugin
 	pluginArguments framework.Arguments
 	weight          *priorityWeight
+	*normalPodConfig
 	*hyperNodesTier
+	// hyperNodeResourceCache stores the resource status of hypernodes to avoid repeated calculation: hypernode -> resource -> status
+	hyperNodeResourceCache map[string]map[corev1.ResourceName]*resourceStatus
+}
```
ouyangshengjia:
```suggestion
	hyperNodeResourceStat map[string]resourceStat // hypernode -> resourceStat
}

type resourceStat struct {
	used *api.Resource
	allocatable *api.Resource
}
```
It is recommended to use the `api.Resource` struct for future expansion.
https://github.com/volcano-sh/volcano/blob/703a8f3d578e866d8acea9e51292a5ad23e912d0/pkg/scheduler/api/resource_info.go#L58-L68

Tau721: Thanks for your suggestion, and I have fixed it.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -141,48 +186,63 @@ func calculateWeight(args framework.Arguments) *priorityWeight {
...
+	// default values are respectively true and 0.8
+	config := normalPodConfig{
+		hyperNodeBinPackingEnable: true,
+		hyperNodeBinPackingFading: 0.8,
+	}
+	args.GetBool(&config.hyperNodeBinPackingEnable, HyperNodeBinPackNormalPodEnable)
+	args.GetFloat64(&config.hyperNodeBinPackingFading, HyperNodeBinPackNormalPodFading)
+	// config.hyperNodeBinPackingFading could be 0, which implies only the hypernodes of tier 1 affect the pod binpacking scores
+	if config.hyperNodeBinPackingFading < 0 {
+		config.hyperNodeBinPackingFading = 0.8
```
qi-min: it would be better to use constant instead

Tau721: Thanks for you review, and I have fixed it.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -45,25 +46,51 @@ const (
 	HyperNodeBinPackResources = "hypernode.binpack.resources"
 	// HyperNodeBinPackResourcesPrefix is the key prefix for additional resource key name
 	HyperNodeBinPackResourcesPrefix = HyperNodeBinPackResources + "."
+	// HyperNodeBinPackNormalPodEnable is the key for whether to enable hypernode-level binpacking for pods without network topology
+	HyperNodeBinPackNormalPodEnable = "hypernode.binpack.normal-pod.enable"
+	// HyperNodeBinPackNormalPodFading is the key for tier weight fading parameter for pods without network topology
+	HyperNodeBinPackNormalPodFading = "hypernode.binpack.normal-pod.fading"
```
JesseStutler, line 52: I think you should also update the design doc and user guide doc to teach users how to use it and what does they mean

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -492,23 +671,34 @@ func (nta *networkTopologyAwarePlugin) scoreWithTaskNum(hyperNodeName string, ta
...
 	return float64(taskNum) / float64(allTaskNum)
 }
+
+func (nta *networkTopologyAwarePlugin) normalizeFinalScore(scores map[string]float64) map[string]float64 {
+	normalizedScores := make(map[string]float64)
+	for name, score := range scores {
+		normalizedScores[name] = float64(k8sFramework.MaxNodeScore) * float64(nta.weight.GlobalWeight) * score
+	}
+	return normalizedScores
+}
```
ouyangshengjia: I don't think the processing of this function can be called normalize.

Tau721: How about changing it to scale?

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -270,65 +369,148 @@ func (nta *networkTopologyAwarePlugin) getHyperNodeBinPackingScore(subJob *api.S
...
+	for tier := nta.hyperNodesTier.minTier; tier <= nta.hyperNodesTier.maxTier; tier++ {
+		// note: math.Pow(0, 0) = 1
+		tierWeight := math.Pow(nta.hyperNodeBinPackingFading, float64(tier-1))
+		totalTierWeight += tierWeight
+		tierWeights[tier] = tierWeight
+	}
+	if totalTierWeight <= 0 {
+		// This should not happen, since there are at least one tier and its weight is one
+		klog.Warningf("the total tier weight of plugin %s should be greater than zero, but got %g", PluginName, totalTierWeight)
+		return nil, fmt.Errorf("the total tier weight of plugin %s should be greater than zero, but got %g", PluginName, totalTierWeight)
```
ouyangshengjia: I think we can just leave a log here an return empty result. Considering that when we return an error in `NodeOrderFn`, the allocate process would be broken and the pod won't be scheduled.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -270,65 +369,148 @@ func (nta *networkTopologyAwarePlugin) getHyperNodeBinPackingScore(subJob *api.S
...
+		score := (used + request) / allocatable
+		klog.V(5).InfoS("hyperNode binpacking score calculation", "task", task.UID, "hyperNode", hyperNode,
+			"resource", resource, "allocatable", allocatable, "used", used, "request", request)
+
+		totalScore += float64(weight) * score
+		totalWeight += weight
+	}
+
+	if totalWeight > 0 {
+		return totalScore / float64(totalWeight)
```
JesseStutler: We can also log the totalScore here

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -270,65 +369,148 @@ func (nta *networkTopologyAwarePlugin) getHyperNodeBinPackingScore(subJob *api.S
...
+	if totalTierWeight <= 0 {
+		// This should not happen, since there are at least one tier and its weight is one
+		klog.Warningf("the total tier weight of plugin %s should be greater than zero, but got %g", PluginName, totalTierWeight)
+		return nil, fmt.Errorf("the total tier weight of plugin %s should be greater than zero, but got %g", PluginName, totalTierWeight)
+	}
+
+	for _, node := range nodes {
+		totalScore := 0.0
+		for tier := nta.hyperNodesTier.minTier; tier <= nta.hyperNodesTier.maxTier; tier++ {
+			tierScore := FullScore
```
ouyangshengjia, line 459: It is recommended to add a comment here:  Prefers to schedule pods on nodes that do not belong to any hypernode.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
ouyangshengjia: Agreed.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -492,23 +671,34 @@ func (nta *networkTopologyAwarePlugin) scoreWithTaskNum(hyperNodeName string, ta
...
+		return float64(nta.maxTier-tier) / float64(nta.maxTier-nta.minTier)
+	}
+	return ZeroScore
 }
 
 func scoreHyperNodeWithTaskNum(taskNum int, allTaskNum int) float64 {
 	// Calculate task distribution rate as score and map the original score to the range between 0 and 1.
 	if allTaskNum == 0 {
-		return ZeroScore
+		return FullScore
```
JesseStutler, line 696: Why to change ZeroScore to FullScore?

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -473,15 +652,15 @@ func (nta *networkTopologyAwarePlugin) networkTopologyAwareScore(hyperNodeName,
...
 	}
 	LCAHyperNode := hyperNodeMap.GetLCAHyperNode(hyperNodeName, jobAllocatedHyperNode)
 	hyperNodeInfo, ok := hyperNodeMap[LCAHyperNode]
 	if !ok {
 		return ZeroScore
 	}
 	// Calculate score: (maxTier - LCAhyperNode.tier)/(maxTier - minTier)
-	hyperNodeTierScore := BaseScore * nta.scoreHyperNodeWithTier(hyperNodeInfo.Tier())
+	hyperNodeTierScore := nta.scoreHyperNodeWithTier(hyperNodeInfo.Tier())
```
JesseStutler, line 666: I have noticed that you remove the BaseScore multiply logic, I'm worried about whether there will be precision issues when comparing float64 values ​​between 0 and 1.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -192,6 +259,7 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
 		klog.V(5).Infof("Leaving networkTopologyAware plugin ...")
 	}()
 	nta.hyperNodesTier.init(ssn.HyperNodesTiers)
+	nta.initHyperNodeResourceCache(ssn)
```
ouyangshengjia, line 262: The `hyperNodeResourceCache` is initialized each time a session is started. Is there any impact on a large cluster? Are there any performance test for this function?

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: As we can see, the score of this function evaluates the dispersion degree of the given subjob's tasks. The more aggregated they are, the higher score they get, and vice versa. When `allTaskNum==0`, it indicates that the subjob's tasks are not dispersed at all. In other words, they are the most aggregated. Therefore, from this perspective, they should get the highest score, i.e, `FullScore`.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: As far as I know, the precision concern regarding float64 comparisons mainly comes from the exact equality checks (e.g., `a == b`). However, there are no such checks in this network-topology-aware plugin. Except for this point, I find no essential difference between multiplying BaseScore here and at the final stage.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: Actually, I have evaluated the time efficiency of this `initHyperNodeResourceCache` function through the `Test_initHyperNodeResourceCache` function in the test file. The results demonstrate that aiming at a cluster with 2000 nodes, 200 hypernodes of tier 1, 40 hypernodes of tier 2, 5 hypernodes of tier 3, 1 hypernode of tier 4, and 3 resource types,  it takes only about 3 milliseconds to initialize this `hyperNodeResourceCache`. Such time cost is totally acceptable.

Besides, I have also evaluated the time efficiency of the new `batchNodeOrderFnForNormalPods` function (which calculates the hypernode-level binpacking score for normal pods) through the `Test_batchNodeOrderFnForNormalPods` function in the test file. The results show that aiming at the same cluster, it takes about 10~15 milliseconds to invoke this function to score the whole 2000 nodes for one task. Such time cost is also acceptable.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: Fixed.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: I agree, and I have fixed it.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: Fixed.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: Thanks for your review.  I have fixed it.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
Tau721: Ok. I will try my best to finish it early.

ouyangshengjia:
/lgtm
defers to @JesseStutler

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
JesseStutler: Oh I have noticed that you have added scaleFinalScore to multiply the basescore, I think it's fine

JesseStutler:
Waiting for @Tau721 https://github.com/volcano-sh/volcano/pull/4889#discussion_r2684712866 finish the doc and then we can get merged

JesseStutler (commented):
/approve

PR #4893, opened 2025-12-30
Fix Infinity Mode Logic in Resource Comparison Functions

zhifei92:
#### What type of PR is this?
/kind bug
The `LessEqual` and `Less`(even though it's no longer used anywhere.) functions in `pkg/scheduler/api/resource_info.go` have an incorrect logic when using `Infinity` mode. When the left resource (`r`) is missing a scalar resource dimension (e.g., GPU) but the right resource (`rr`) has it, the function incorrectly returns `false`.

##### Expected Behavior
When `r` doesn't have a dimension that `rr` has, it should be treated as `r[dimension] = 0`, and the comparison should continue (e.g., `0 <= rr[dimension]` returns `true`).

##### Actual Behavior (Before Fix)
The function returns `false` immediately, treating the missing dimension as if `r` exceeds `rr`, which is semantically incorrect.

##### Root Cause

The problematic code :
https://github.com/volcano-sh/volcano/blob/3568eee68aae751b78acfdc857479c1222b217b5/pkg/scheduler/api/resource_info.go#L416-L422
This logic enforces that `r` must contain all dimensions present in `rr`.

###### 1. **Incorrect Eviction in Capacity Plugin**

The most critical impact is in the capacity plugin's reclaim logic:
https://github.com/volcano-sh/volcano/blob/3568eee68aae751b78acfdc857479c1222b217b5/pkg/scheduler/plugins/capacity/capacity.go#L123-L127

**Bug Scenario:**
```go
allocated := {CPU: 2000m, Memory: 4Gi}           // No GPU requested
deserved  := {CPU: 4000m, Memory: 8Gi, GPU: 2}  // GPU configured

// Before fix:
allocated.LessEqual(deserved, Infinity) = false  // Wrong!
→ Continue to guarantee check
→ May incorrectly reclaim resources
→ **Causes unexpected pod eviction**

// After fix:
allocated.LessEqual(deserved, Infinity) = true   // Correct!
→ Skip reclamation
→ **Avoids incorrect eviction**
```

Fixes #

```release-note
Fixed unexpected pod evictions caused by unused resource types in queue quota configuration.
```

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -374,14 +374,6 @@ func (r *Resource) Less(rr *Resource, defaultValue DimensionDefaultValue) bool {
 		return false
 	}
 
-	if defaultValue == Infinity {
-		for name := range rr.ScalarResources {
-			if _, ok := r.ScalarResources[name]; !ok {
-				return false
-			}
-		}
-	}
-
 	for resourceName, leftValue := range r.ScalarResources {
 		rightValue, ok := rr.ScalarResources[resourceName]
 		if !ok && defaultValue == Infinity {
@@ -411,14 +403,6 @@ func (r *Resource) LessEqual(rr *Resource, defaultValue DimensionDefaultValue) b
 	}
 	if !lessEqualFunc(r.Memory, rr.Memory, minResource) {
 		return false
-	}
-
-	if defaultValue == Infinity {
-		for name := range rr.ScalarResources {
-			if _, ok := r.ScalarResources[name]; !ok {
-				return false
-			}
-		}
 	}
 
 	for resourceName, leftValue := range r.ScalarResources {
--- a/pkg/scheduler/api/resource_info_test.go
+++ b/pkg/scheduler/api/resource_info_test.go
@@ -562,7 +562,7 @@ func TestLess(t *testing.T) {
 				Memory:          2000,
 				ScalarResources: map[v1.ResourceName]float64{"scalar.test/scalar1": 1000, "hugepages-test": 2000},
 			},
-			expected: false,
+			expected: true,
 		},
 		{
 			resource1: &Resource{
@@ -706,7 +706,7 @@ func TestLessEqual(t *testing.T) {
 				Memory:          2000,
 				ScalarResources: map[v1.ResourceName]float64{"scalar.test/scalar1": 1000, "hugepages-test": 2000},
 			},
-			expected: false,
+			expected: true,
 		},
 		{
 			resource1: &Resource{
@@ -727,7 +727,7 @@ func TestLessEqual(t *testing.T) {
 				Memory:          2000,
 				ScalarResources: map[v1.ResourceName]float64{"scalar.test/scalar1": 1000, "hugepages-test": 2000},
 			},
-			expected: false,
+			expected: true,
 		},
 		{
 			resource1: &Resource{


zhifei92:
The LessEqual method that propagates the infinity parameter throughout the project is only used in the capacity plugin.

JesseStutler:
Looks like consider not set as zero is more reasonable, don't know why https://github.com/volcano-sh/volcano/pull/2637 is more willing to consider not set as inifity. /cc @kingeasternsun

JesseStutler:
@zhifei92 Hi, please don't forget to sign off your commit using `git commit -s`

zhifei92:
> @zhifei92 Hi, please don't forget to sign off your commit using `git commit -s`

done

zhifei92:
The CI failure doesn't seem to be caused by this PR. I'll update it after https://github.com/volcano-sh/volcano/pull/4914/changes is merged.

JesseStutler:
@zhifei92 I updated the code first and give CIs a try, but I don't hope we should keep a merge commit, you can rebase the latest code and push again

zhifei92:
> you can rebase the latest code and push again

done

hajnalmt (commented):
/hold
/cc @kingeasternsun 
Sry I missed this PR.
Let's start a discussion here, I think the function works correctly here, and it's working as desinged.

It was added in this PR:
https://github.com/volcano-sh/volcano/pull/2637
(I added [kingeasternsun](https://github.com/kingeasternsun) due to this.)

The problem comes from misusing the function in the capacity plugin in the comparision you mentioned. This is the only place where we compare with `api.Infinity` hence the problem arose. But design-wise I think, if the left value is not defined in the resource comparsion we should return False when you provide `api.Infinity` to the function.

The bug you mentioned is actually pretty serious. We are overreclaiming now below deserved.

hajnalmt:
Okay, I checked the code on my prod env about what did I do [here](https://github.com/hajnalmt/volcano/commit/10eb5f0480dbbf12909c519691bd7258accb8f74).
It was not a one liner problem to solve unfortunately, and I was reluctant to raise a PR on this yet, because I don't like my code here. I want to mention that I have tried to solve a diffent issue here. I needed to create queues without `cpu` and `memory` on deserved and it wasn't working properly due to this same comparision.

It feels hacky still, but it actually solves this issue too so I have raised a [PR](https://github.com/volcano-sh/volcano/pull/4919) in the end, based on this change and tried to summerize the issues I have met during the development of this here:
https://github.com/volcano-sh/volcano/issues/4918

JesseStutler:
> /hold /cc @kingeasternsun Sry I missed this PR. Let's start a discussion here, I think the function works correctly and by that I mean it's working as designed.
> 
> It was added in this PR: #2637 (I added [kingeasternsun](https://github.com/kingeasternsun) due to this.)
> 
> The problem comes from misusing the function in the capacity plugin in the comparision you mentioned. This is the only place where we compare with `api.Infinity` hence the problem arose. But design-wise I think, if the left value is not defined in the resource comparsion we should return False when you provide `api.Infinity` to the function.
> 
> The bug you mentioned is actually pretty serious. We are overreclaiming now below deserved.

Thanks mate you reminded me, I think we'd rather use `LessEqualPartly` in the code rather than delete codes in the `LessEqual`, Although I think the logic of LessEqual is not quite right here, because I think the non-existence of resources in r should not be regarded as infinity.

JesseStutler:
> Okay, I checked the code on my prod env about what did I do [here](https://github.com/hajnalmt/volcano/commit/10eb5f0480dbbf12909c519691bd7258accb8f74). It was not a one liner problem to solve unfortunately, and I was reluctant to raise a PR on this yet, because I don't like my code here. I want to mention that I have tried to solve a diffent issue here. I needed to create queues without `cpu` and `memory` on deserved and it wasn't working properly due to this same comparision.
> 
> It feels hacky still, but it actually solves this issue too so I have raised a [PR](https://github.com/volcano-sh/volcano/pull/4919) in the end, based on this change and tried to summerize the issues I have met during the development of this here: #4918

As @hajnalmt mentioned in #4918 bug1, I do think we wrongfully didn't consider the resources reclaimer request, only compares the allocated and deserverd(but we still need to compare the allocated and deserved), we just need to add more judgements that we need to also judge that whether the scheduler has already reclaimed enough resources for the reclaimer

hajnalmt:
Okay, I think I managed to bring this to a much more acceptable form:
https://github.com/volcano-sh/volcano/pull/4927
https://github.com/volcano-sh/volcano/pull/4919
Any feedbacks are welcomed!
@zhifei92 What do you think can we have these modifications instead of this? I still think that the function signature here works correctly, since the blank dimensions in general should be handled as `api.Infinity` so if anything is present on the right but not on the left than the left is bigger.

JesseStutler:
> Okay, I think I managed to bring this to a much more acceptable form: #4927 #4919 Any feedbacks are welcomed! @zhifei92 What do you think can we have these modifications instead of this? I still think that the function signature here works correctly, since the blank dimensions in general should be handled as `api.Infinity` so if anything is present on the right but not on the left than the left is bigger.

Thanks! I'd prefer to follow the way in https://github.com/volcano-sh/volcano/pull/4919

kingeasternsun:
> > Okay, I think I managed to bring this to a much more acceptable form: #4927 #4919 Any feedbacks are welcomed! @zhifei92 What do you think can we have these modifications instead of this? I still think that the function signature here works correctly, since the blank dimensions in general should be handled as `api.Infinity` so if anything is present on the right but not on the left than the left is bigger.
> 
> Thanks! I'd prefer to follow the way in #4919
I agree, #4919 is better.

PR #4937, opened 2026-01-12
chore(metrics): remove non-existent metrics schedule_attempts_total

ruanwenjun:
- The metrics `schedule_attempts_total` does not exist

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/metrics/metrics.go
+++ b/pkg/scheduler/metrics/metrics.go
@@ -109,6 +109,14 @@ var (
 		},
 	)
 
+	scheduleAttempts = promauto.NewCounterVec(
+		prometheus.CounterOpts{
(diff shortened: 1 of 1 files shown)


pkg/scheduler/metrics/metrics.go
JesseStutler, line 1: I agree to remove this metric from the doc because we don't use it for now, but keep it in the code because I think it might be used in the future, such as showing the different states of each pod in each session @ruanwenjun

ruanwenjun: OK, I revert the change of code.
Yes, we are currently very concerned about the state of the pod after each session and the state of the pod in the cache. This allows us to determine how many pods were successfully scheduled per session.

Issue #4899, opened 2025-12-31
Improve the performance of network topology-aware scheduling in large-scale cluster scenarios


ouyangshengjia:
### What is the problem you're trying to solve

In the current design, network topology-aware scheduling iterates through tiers from low to high, traversing all HyperNodes in the current tier for scheduling by dry run. However, this scheduling logic is inefficient in large cluster scenarios.

Consider the following scenario:

- The cluster consists of 1000 nodes, with every 10 nodes grouped into a tier-1 HyperNode (resulting in 100 tier-1 HyperNodes in total). All 1000 nodes together form a single tier-2 HyperNode.
- A PodGroup contains 1000 pods, with every 10 pods forming a SubGroup (resulting in 100 SubGroups in total). Each pod exclusively occupies one node.
  `highestTierAllowed` for PodGroup is 2;
  `highestTierAllowed` for SubGroup is 1;
  `minSubgroups` for PodGroup is 100.

During scheduling, the scheduler first attempts to schedule all 1000 pods within each tier-1 HyperNode and rollback. Only after all 100 tier-1 HyperNodes have been exhaustively attempted (and all have failed),  the scheduler will proceed to attempt the tier-2 HyperNode.

This scheduling process suffers from two inefficiencies:

1. The resource capacity of a tier-1 HyperNode is clearly insufficient to meet the total resource requirement of the entire PodGroup, yet there is no pre-filtering mechanism to skip trial scheduling on such HyperNodes.
2. Even after a SubGroup fails scheduling, the system continues attempting to schedule subsequent SubGroups, despite the fact that the `minSubGroups` requirement can no longer be satisfied.

### Describe the solution you'd like

To address these issues, I propose the following optimizations:

- **For issue 1**: Introduce HyperNode pre-filtering logic within `HyperNodeGradientForJobFn` and `HyperNodeGradientForSubJobFn`. Specifically, if the total available resources of a HyperNode are less than the minimum resource requirement of the PodGroup (or SubGroup), the HyperNode should be excluded from the returned candidate list.
  https://github.com/volcano-sh/volcano/blob/8aba772412bed8bdd9c20b599f97ff7835e8f422/pkg/scheduler/framework/session_plugins.go#L1048-L1051
  https://github.com/volcano-sh/volcano/blob/8aba772412bed8bdd9c20b599f97ff7835e8f422/pkg/scheduler/framework/session_plugins.go#L1069-L1072
[...]

PR #4940, opened 2026-01-13
perf(nta) NTA scheduling performance improvement by optimizing hyperNodeGradientFn

3th4novo:
#### What type of PR is this?

area/performance kind/feature

Currently, the network topology-aware plugin iterates from bottom to top in a hypernode tree and caused inefficiency. For detailed description please review issue [#4899](https://github.com/volcano-sh/volcano/issues/4899#event-21805727490) . This PR introduces HyperNode pre-filtering logic within hyperNodeGradientFn. Specifically, we only keep a hypernode to eligibleHyperNodes when minResource is less than or equal to the idle resource of a hypernode, or the future idle resource of a hypernode.

Note that here minResource can refer to both the minimum required resource of a job or a sub job.

The PR also modifies sub_job_info.go for sub job minReource calculation purposes. Right now a sub job is constrained to gang scheduling, so the minimum required resource of a sub job can be considered as the total initResReq of each pending task in the sub job.

```go
func (sji *SubJobInfo) GetMinResources() *Resource {
	totalResource := EmptyResource()
	for _, task := range sji.TaskStatusIndex[Pending] {
		totalResource.Add(task.InitResreq)
	}
	return totalResource
}
``` 

Improvement: Under the setup as in issue 4899, when we have a subgroup for each pod, resulting in 1000 subgroups, the total scheduling duration can last about 1 minute. This number is optimized to a few seconds in this PR. When we build 100 subgroups, each containing 10 pods, scheduling duration is optimized by about 0.5 seconds, from about 6.3 seconds to 5.9 seconds. By observation, I believe the differences in the number of subgroups mainly contribute to scheduling duration before optimization. The main reason is that the scheduler will have to perform more dry run operations.

#4899 issue 1

This PR is based on [#4889](https://github.com/volcano-sh/volcano/pull/4889) so it should be merged after #4889 is merged.

Feel free to modify TestHyperNodeGradientPreFiltering for correctness checks and BenchmarkHyperNodeGradientFnPerformance for performance benchmarks. 

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -5142,6 +5142,387 @@ func TestAllocateWithPartitionPolicyNetworkTopology(t *testing.T) {
 			ExpectBindsNum:   2,
 			MinimalBindCheck: true,
 		},
+		{
+			Name: "rescheduling scenario: SubJob with multiple pending tasks, GetMinResources returns sum of all pending tasks resources",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroupWithSubGroupPolicy("pg1", "c1", "s0", "q1", 4, nil, schedulingv1.PodGroupInqueue, "hard", 2,
+					[]schedulingv1.SubGroupPolicySpec{
+						util.BuildSubGroupPolicyWithSubGroupSize("task1", []string{"volcano.sh/task-spec"}, "hard", 1, 4),
+					}),
+			},
+			Pods: []*v1.Pod{
+				// 2 Running tasks + 2 Pending tasks in the same SubJob
+				// GetMinResources should return 4 CPU, 8G (sum of 2 pending tasks)
+				util.BuildPod("c1", "p1", "s0-n1", v1.PodRunning, api.BuildResourceList("2", "4G"), "pg1", map[string]string{"volcano.sh/task-spec": "worker"}, nil),
+				util.BuildPod("c1", "p2", "s0-n2", v1.PodRunning, api.BuildResourceList("2", "4G"), "pg1", map[string]string{"volcano.sh/task-spec": "worker"}, nil),
+				util.BuildPod("c1", "p3", "", v1.PodPending, api.BuildResourceList("2", "4G"), "pg1", map[string]string{"volcano.sh/task-spec": "worker"}, nil),
+				util.BuildPod("c1", "p4", "", v1.PodPending, api.BuildResourceList("2", "4G"), "pg1", map[string]string{"volcano.sh/task-spec": "worker"}, nil),
+			},
+			Nodes: []*v1.Node{
+				util.BuildNode("s0-n1", api.BuildResourceList("4", "8Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil),
+				util.BuildNode("s0-n2", api.BuildResourceList("4", "8Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil),
+				util.BuildNode("s1-n3", api.BuildResourceList("4", "8Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil),
+				util.BuildNode("s1-n4", api.BuildResourceList("4", "8Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil),
+			},
+			HyperNodesSetByTier: map[int]sets.Set[string]{1: sets.New[string]("s0", "s1"), 2: sets.New[string]("s2")},
+			HyperNodesMap: map[string]*api.HyperNodeInfo{
+				"s0": api.NewHyperNodeInfo(api.BuildHyperNode("s0", 1, []api.MemberConfig{
+					{
+						Name:     "s0-n1",
+						Type:     topologyv1alpha1.MemberTypeNode,
+						Selector: "exact",
+					},
+					{
+						Name:     "s0-n2",
+						Type:     topologyv1alpha1.MemberTypeNode,
+						Selector: "exact",
+					},
+				})),
+				"s1": api.NewHyperNodeInfo(api.BuildHyperNode("s1", 1, []api.MemberConfig{
+					{
+						Name:     "s1-n3",
+						Type:     topologyv1alpha1.MemberTypeNode,
+						Selector: "exact",
+					},
+					{
+						Name:     "s1-n4",
+						Type:     topologyv1alpha1.MemberTypeNode,
+						Selector: "exact",
+					},
+				})),
+				"s2": api.NewHyperNodeInfo(api.BuildHyperNode("s2", 2, []api.MemberConfig{
+					{
+						Name:     "s0",
+						Type:     topologyv1alpha1.MemberTypeHyperNode,
+						Selector: "exact",
+					},
+					{
+						Name:     "s1",
+						Type:     topologyv1alpha1.MemberTypeHyperNode,
+						Selector: "exact",
+					},
+				})),
+			},
+			HyperNodes: map[string]sets.Set[string]{
+				"s0": sets.New[string]("s0-n1", "s0-n2"),
+				"s1": sets.New[string]("s1-n3", "s1-n4"),
+				"s2": sets.New[string]("s0-n1", "s0-n2", "s1-n3", "s1-n4"),
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("q1", 1, nil),
+			},
+			ExpectBindsNum:   2,
+			MinimalBindCheck: true,
+		},
+		{
+			Name: "rescheduling scenario: SubJob with single pending task",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroupWithSubGroupPolicy("pg1", "c1", "s0", "q1", 3, nil, schedulingv1.PodGroupInqueue, "hard", 2,
+					[]schedulingv1.SubGroupPolicySpec{
+						util.BuildSubGroupPolicyWithSubGroupSize("task1", []string{"volcano.sh/task-spec"}, "hard", 1, 3),
+					}),
+			},
+			Pods: []*v1.Pod{
+				// 2 Running tasks + 1 Pending task
+				// GetMinResources should return 2 CPU, 4G (1 pending task)
+				util.BuildPod("c1", "p1", "s0-n1", v1.PodRunning, api.BuildResourceList("2", "4G"), "pg1", map[string]string{"volcano.sh/task-spec": "worker"}, nil),
+				util.BuildPod("c1", "p2", "s0-n2", v1.PodRunning, api.BuildResourceList("2", "4G"), "pg1", map[string]string{"volcano.sh/task-spec": "worker"}, nil),
+				util.BuildPod("c1", "p3", "", v1.PodPending, api.BuildResourceList("2", "4G"), "pg1", map[string]string{"volcano.sh/task-spec": "worker"}, nil),
+			},
+			Nodes: []*v1.Node{
+				util.BuildNode("s0-n1", api.BuildResourceList("4", "8Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil),
+				util.BuildNode("s0-n2", api.BuildResourceList("4", "8Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil),
+				util.BuildNode("s1-n3", api.BuildResourceList("4", "8Gi", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil),
+			},
+			HyperNodesSetByTier: map[int]sets.Set[string]{1: sets.New[string]("s0", "s1"), 2: sets.New[string]("s2")},
+			HyperNodesMap: map[string]*api.HyperNodeInfo{
+				"s0": api.NewHyperNodeInfo(api.BuildHyperNode("s0", 1, []api.MemberConfig{
+					{
+						Name:     "s0-n1",
+						Type:     topologyv1alpha1.MemberTypeNode,
+						Selector: "exact",
+					},
+					{
+						Name:     "s0-n2",
+						Type:     topologyv1alpha1.MemberTypeNode,
+						Selector: "exact",
+					},
+				})),
+				"s1": api.NewHyperNodeInfo(api.BuildHyperNode("s1", 1, []api.MemberConfig{
+					{
+						Name:     "s1-n3",
+						Type:     topologyv1alpha1.MemberTypeNode,
+						Selector: "exact",
+					},
+				})),
+				"s2": api.NewHyperNodeInfo(api.BuildHyperNode("s2", 2, []api.MemberConfig{
+					{
+						Name:     "s0",
+						Type:     topologyv1alpha1.MemberTypeHyperNode,
+						Selector: "exact",
+					},
+					{
+						Name:     "s1",
+						Type:     topologyv1alpha1.MemberTypeHyperNode,
+						Selector: "exact",
+					},
+				})),
+			},
+			HyperNodes: map[string]sets.Set[string]{
+				"s0": sets.New[string]("s0-n1", "s0-n2"),
+				"s1": sets.New[string]("s1-n3"),
+				"s2": sets.New[string]("s0-n1", "s0-n2", "s1-n3"),
+			},
+			Queues: []*schedulingv1.Queue{
+				util.BuildQueue("q1", 1, nil),
+			},
+			ExpectBindsNum:   1,
+			MinimalBindCheck: true,
+		},
+		{
+			Name: "rescheduling scenario: multiple SubJobs with different pending task counts",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroupWithSubGroupPolicy("pg1", "c1", "s2", "q1", 6, nil, schedulingv1.PodGroupInqueue, "hard", 2,
+					[]schedulingv1.SubGroupPolicySpec{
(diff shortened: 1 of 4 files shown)


pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -380,7 +540,15 @@ func (nta *networkTopologyAwarePlugin) hyperNodeGradientFn(ssn *framework.Sessio
...
+			if minResource != nil {
+				hnResourceStatus := nta.hyperNodeResourceCache[current.Name]
+				if hnResourceStatus != nil {
+					if minResource.LessEqual(hnResourceStatus.idle, api.Zero) ||
+						minResource.LessEqual(hnResourceStatus.futureIdle, api.Zero) {
+						eligibleHyperNodes[current.Tier()] = append(eligibleHyperNodes[current.Tier()], current)
+					}
+				}
+			}
 		}
```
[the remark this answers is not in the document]
3th4novo: I have corrected it.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware_test.go
```diff
@@ -1928,10 +2109,1424 @@ func parseTask(jobInfoMap map[api.JobID]*api.JobInfo) *api.TaskInfo {
...
+		if subJob.UID == "ns1/pg1/task1-worker" {
+			return subJob
+		}
+	}
+	return nil
+}
+
+// TestHyperNodeGradientPreFiltering tests the pre-filtering logic in hyperNodeGradientFn.
+// It verifies that HyperNodes are correctly filtered based on idle and futureIdle resources.
+func TestHyperNodeGradientPreFiltering(t *testing.T) {
```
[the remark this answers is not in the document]
3th4novo: I have added two more test cases where minResource == nil, one for job and another one for sub job.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -229,10 +298,31 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
...
+		DeallocateFunc: func(event *framework.Event) {
+			task := event.Task
+			node := task.NodeName
+			for hyperNode := range ssn.HyperNodes {
+				if ssn.RealNodesSet[hyperNode].Has(node) {
+					nta.hyperNodeResourceCache[hyperNode].used.Sub(task.Resreq)
+				}
+			}
+		},
+	})
```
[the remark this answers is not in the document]
3th4novo: will fix in the following review.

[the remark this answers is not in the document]
3th4novo: code updated, will review again.

pkg/scheduler/actions/allocate/allocate_test.go
```diff
@@ -5182,3 +5184,169 @@ func TestAllocateWithPartitionPolicyNetworkTopology(t *testing.T) {
...
+
+	for _, job := range ssn.Jobs {
+		topHyperNode := ssn.HyperNodes[framework.ClusterTopHyperNode]
+		gradient := ssn.HyperNodeGradientForJobFn(job, topHyperNode)
+		klog.Infof("\nGradient for Job %s: %d tiers", job.UID, len(gradient))
+		for tierIdx, tierHNs := range gradient {
+			klog.Infof("  Tier %d: %d hyperNodes", tierIdx+1, len(tierHNs))
+		}
+	}
+}
```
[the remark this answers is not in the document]
3th4novo: fixed

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -562,7 +571,18 @@ func (nta *networkTopologyAwarePlugin) hyperNodeGradientFn(ssn *framework.Sessio
...
+			} else {
+				hnResourceStatus, found := nta.hyperNodeResourceCache[current.Name]
+				if !found {
+					klog.Warningf("Resource status for hypernode %s not found in cache, skipping pre-filtering for it.", current.Name)
+					eligibleHyperNodes[current.Tier()] = append(eligibleHyperNodes[current.Tier()], current)
+				} else if minResource.LessEqual(hnResourceStatus.idle, api.Zero) ||
+					minResource.LessEqual(hnResourceStatus.futureIdle, api.Zero) {
+					eligibleHyperNodes[current.Tier()] = append(eligibleHyperNodes[current.Tier()], current)
+				}
+			}
```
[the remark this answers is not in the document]
3th4novo: fixed

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -562,7 +571,14 @@ func (nta *networkTopologyAwarePlugin) hyperNodeGradientFn(ssn *framework.Sessio
 		processQueue = processQueue[1:]
 
 		if current.Tier() <= highestAllowedTier {
-			eligibleHyperNodes[current.Tier()] = append(eligibleHyperNodes[current.Tier()], current)
+			hnResourceStatus, found := nta.hyperNodeResourceCache[current.Name]
+			if !found {
+				klog.Warningf("Resource status for hypernode %s not found in cache, skipping pre-filtering for it.", current.Name)
+				eligibleHyperNodes[current.Tier()] = append(eligibleHyperNodes[current.Tier()], current)
+			} else if minResource.LessEqual(hnResourceStatus.idle, api.Zero) ||
+				minResource.LessEqual(hnResourceStatus.futureIdle, api.Zero) {
```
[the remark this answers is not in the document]
3th4novo: fixed

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -543,7 +551,10 @@ func (nta *networkTopologyAwarePlugin) batchNodeOrderFnForNetworkAwarePods(ssn *
 	return nodeScores, nil
 }
 
-func (nta *networkTopologyAwarePlugin) hyperNodeGradientFn(ssn *framework.Session, hyperNode *api.HyperNodeInfo, highestAllowedTier int, allocatedHyperNode string) ([][]*api.HyperNodeInfo, error) {
+// minResource is the minimum required resource for the job/subjob to be scheduled.
```
JesseStutler: The explanation should cover the entire function and the purpose of all its parameters, not just the explanation of a single parameter.

pkg/scheduler/actions/allocate/allocate_test.go
```diff
@@ -5182,3 +5184,169 @@ func TestAllocateWithPartitionPolicyNetworkTopology(t *testing.T) {
 		})
 	}
 }
+
+// TestHyperNodeGradientFnPerformance tests the performance optimization
+// of hyperNodeGradientFn with SubGroup policy in large-scale cluster scenarios.
+func TestHyperNodeGradientFnPerformance(t *testing.T) {
```
JesseStutler: I think the performance test should use `*testing.B` instead, I didn't find any validation for using `*testing.T`

3th4novo: Fixed, ty.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
3th4novo: Fixed, ty.

pkg/scheduler/actions/allocate/allocate_test.go
JesseStutler: @3th4novo I mean you should use testing.B to run b.N times, and the testing struct may should follow the style like `BenchmarkAllocate`

3th4novo: Please refer to the latest commit. Let me know if there are any changes needed.

pkg/scheduler/api/sub_job_info.go
```diff
@@ -261,3 +261,14 @@ func (sji *SubJobInfo) AllocatedTaskNum() int32 {
 func (sji *SubJobInfo) CloneStatusFrom(source *SubJobInfo) {
 	sji.AllocatedHyperNode = source.AllocatedHyperNode
 }
+
+// GetMinResources The current sub job is constrained to gang scheduling,
+// so the minimum required resource of a sub job can be considered as
+// the total initResReq of each task in the sub job.
+func (sji *SubJobInfo) GetMinResources() *Resource {
```
Tau721, line 269: This function sums up the `InitResreq` of all tasks of a subjob and uses it as the minimal resource requirement to filter hypernodes. However, what if some tasks of this subjob have already been running? I think the resource requirements of these tasks should be ignored.

3th4novo: I do think that here we should still consider the requirements of the pending tasks. Actually I used the InitResreq of all pending tasks of a subjob here in the previous versions, so that the function would do the correct calculation under this circumstance. Do you suggest I implement this way?

3th4novo: @ouyangshengjia @JesseStutler What do you guys think of the desired implementation here, since it's a new design?

JesseStutler: I think @Tau721 is correct, we should ignore the running pods, however, the semantics of this function are rather strange, minResources is simply meant to schedule minAvailable pods.

JesseStutler: I remember that we also contains some fields like min number count in subjob, so we should only consider the subjob level's min number count

3th4novo: I agree with you on only counting the initresreq of the pending tasks of a sub job and I have updated it in the latest commit. However, currently the constraint of a sub job is that MinAvailable == size of subjob, therefore I believe my implementation makes sense.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -271,7 +277,8 @@ func (nta *networkTopologyAwarePlugin) OnSessionOpen(ssn *framework.Session) {
 
 	ssn.AddHyperNodeGradientForJobFn(nta.Name(), func(job *api.JobInfo, hyperNode *api.HyperNodeInfo) [][]*api.HyperNodeInfo {
 		if hardMode, highestAllowedTier := job.IsHardTopologyMode(); hardMode {
-			result, err := nta.hyperNodeGradientFn(ssn, hyperNode, highestAllowedTier, job.AllocatedHyperNode)
+			jobMinResource := job.GetMinResources()
```
JesseStutler, line 280: So you should also consider rescheduling scenarios for job, and I feel the job scenario is even more complex, for example, evicted pods might belong to different subjobs, and these subjobs are distributed under different hypernodes. Therefore, we cannot use simple summation to calculate minResources.

pkg/scheduler/api/sub_job_info.go
JesseStutler: @3th4novo Also please add test cases to go through the whole allocate action for rescheduling scenario, you can add cases in allocate_test.go, e.g., add in `TestAllocateWithPartitionPolicyNetworkTopology`

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
ouyangshengjia: Indeed, in scenarios where podgroups are partially running or in failure rescheduling scenarios (such as `RestartTask`/`RestartPartition` of vcjob), the pruning conditions here become very complex. I was wondering whether this scheduling pruning optimization could be applied only to podgroups that are **completely pending**, while allowing partially running podgroups to proceed without restriction.

JesseStutler: Great, this is also a method, but I'm wondering if a hypernode that can't accommodate the total resources calculated through simple addition is necessarily not eligible as a candidate. Since we're doing pre-predicate here, even if topology isn't considered here, it will still go through the predicate later, but at least it can exclude hypernodes that are definitely resource-insufficient

ouyangshengjia: For job-level pruning optimization, **do not consider the failure rescheduling scenarios** at the beginning:
When a job meets the minavailable condition, only one pod needs to be scheduled subsequently to complete this round of scheduling. Therefore, pruning evaluation is only required when the job is completely pending.

**If the failure rescheduling scenarios is considered**, the job may be in a state where some pods are running but the minavailable condition **is not yet met**. In this case, if pruning evaluation is performed, the resources required for the minavailable condition need to be calculated by deducting the resources occupied by the currently running pods, making the calculation condition more complex. Considering that the rescheduling scenario may not be common, **it is recommended that pruning be not performed in this scenario.**

3th4novo:
```go
if allocatedHyperNode != "" {
    eligibleHyperNodes[current.Tier()] = append(eligibleHyperNodes[current.Tier()], current)
}
```

Please check line 584-585. Under rescheduling scenarios, the job/subjob would be allocated to some hypernode and it will skip pre-filtering.

ouyangshengjia: @3th4novo You can add comments in code to explain why your pre-filtering is only done on this branch.

3th4novo: done.

pkg/scheduler/api/sub_job_info.go
3th4novo: done.

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -562,7 +580,20 @@ func (nta *networkTopologyAwarePlugin) hyperNodeGradientFn(ssn *framework.Sessio
 		processQueue = processQueue[1:]
 
 		if current.Tier() <= highestAllowedTier {
```
ouyangshengjia: It is recommended to extract a filter function as follows:
```golang
if nta.isEligibleHyperNode(current, ...) {
    eligibleHyperNodes[current.Tier()] = append(eligibleHyperNodes[current.Tier()], current)
}

func (nta *networkTopologyAwarePlugin)isEligibleHyperNode(hn *api.HyperNodeInfo, ...) bool {
    if hn.Tier() > highestAllowedTier {
        return false // the tier should not exceed the highest allowed
    }

    if allocatedHyperNode != "" {
        return true // skip pre-filtering in partially running scenarios
    }
    stats, found := nta.hyperNodeResourceCache[current.Name]
    if !found {
        return true // skip pre-filtering due to insufficient conditions
    }
    if minResource.LessEqual(stats.idle, api.Zero) || minResource.LessEqual(stats.futureIdle, api.Zero) {
        return true
    }
    return false
}
```

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -543,7 +551,17 @@ func (nta *networkTopologyAwarePlugin) batchNodeOrderFnForNetworkAwarePods(ssn *
...
-func (nta *networkTopologyAwarePlugin) hyperNodeGradientFn(ssn *framework.Session, hyperNode *api.HyperNodeInfo, highestAllowedTier int, allocatedHyperNode string) ([][]*api.HyperNodeInfo, error) {
+// hyperNodeGradientFn computes network topology gradients by performing BFS traversal from the given HyperNode,
+// filtering and grouping HyperNodes by tier based on resource availability and topology constraints.
+// It returns HyperNodes organized in ascending tier order (lower tiers represent closer network proximity).
+//
+// Parameters:
+//   - ssn: scheduling session containing all HyperNode information and cluster state
+//   - hyperNode: starting HyperNode for the search, typically the root of available HyperNode subtree
+//   - highestAllowedTier: maximum allowed topology tier to limit search scope
+//   - allocatedHyperNode: previously allocated HyperNode name for rescheduling scenarios (empty for initial scheduling)
```
ouyangshengjia: Just a suggestion, I think "partially running" is more accurate than "rescheduling"

3th4novo: fixed

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
3th4novo: fixed

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -590,6 +608,25 @@ func (nta *networkTopologyAwarePlugin) hyperNodeGradientFn(ssn *framework.Sessio
...
+
+	if allocatedHyperNode != "" {
+		return true // skip pre-filtering in partially running scenarios
+	}
+
+	hnResourceStatus, found := nta.hyperNodeResourceCache[hn.Name]
+	if !found {
+		return true // Resource status for hypernode not found in cache, skipping pre-filtering for it.
+	}
+	if minResource == nil || minResource.LessEqual(hnResourceStatus.idle, api.Zero) || minResource.LessEqual(hnResourceStatus.futureIdle, api.Zero) {
```
ouyangshengjia: In which scenario will `minResource` be `nil`?

3th4novo: I did that according to an older gemini suggestion... mainly for safety reasons. But in fact both the GetMinResources method from a job and sub job wouldn't introduce a nil value. Let me get rid of this value check.

PR #4952, opened 2026-01-15
feat: support inheriting pod annotations and labels when creating PodGroup

kingeasternsun:
#### What type of PR is this?

/kind feature

This PR enhances the behavior of **pg-controller** when creating a PodGroup for a Pod.

Previously, pg-controller only inherited annotations from the Pod’s owner (e.g. Job, Deployment).
With this change, pg-controller can additionally:

* Inherit **annotations and labels directly from the Pod itself**
* Support **fine-grained configuration** to explicitly specify **which annotations and labels should be inherited**

This makes PodGroup metadata propagation more flexible and controllable, which is especially useful for:

* Scheduling-related annotations/labels
* User-defined metadata that should not be placed on the owner object
* Avoiding over-inheriting unnecessary fields

Fixes # https://github.com/volcano-sh/volcano/issues/4953

* This change is **backward-compatible**: existing behavior of inheriting annotations from the Pod owner remains unchanged by default.
* Inheritance from Pod annotations/labels is **opt-in** and configurable.
* Please pay special attention to:

  * The inheritance filtering logic
  * Precedence rules between owner metadata and Pod metadata (if applicable)
  * Any potential impact on existing PodGroup creation flows

```release-note
pg-controller now supports inheriting annotations and labels from the Pod itself when creating a PodGroup, with the ability to explicitly configure which fields are propagated.
```

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/component-base/config"
 	componentbaseconfigvalidation "k8s.io/component-base/config/validation"
 
+	podgroup "volcano.sh/volcano/pkg/controllers/podgroup/options"
 	"volcano.sh/volcano/pkg/kube"
 )
 
@@ -77,8 +78,10 @@ type ServerOption struct {
 	EnableHealthz      bool
 	EnableMetrics      bool
 	ListenAddress      string
-	// To determine whether inherit owner's annotations for pods when create podgroup
-	InheritOwnerAnnotations bool
+
+	// PodGroupOptions holds podgroup controller specific options.
+	PodGroupOptions podgroup.Options
+
 	// WorkerThreadsForPG is the number of threads syncing podgroup operations
 	// The larger the number, the faster the podgroup processing, but requires more CPU load.
 	WorkerThreadsForPG uint32
@@ -127,12 +130,15 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.BoolVar(&s.EnableHealthz, "enable-healthz", false, "Enable the health check; it is false by default")
 	fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")
 	fs.StringVar(&s.ListenAddress, "listen-address", defaultListenAddress, "The address to listen on for HTTP requests.")
-	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
+
 	fs.Uint32Var(&s.WorkerThreadsForPG, "worker-threads-for-podgroup", defaultPodGroupWorkers, "The number of threads syncing podgroup operations. The larger the number, the faster the podgroup processing, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForGC, "worker-threads-for-gc", defaultGCWorkers, "The number of threads for recycling jobs. The larger the number, the faster the job recycling, but requires more CPU load.")
 	fs.Uint32Var(&s.WorkerThreadsForQueue, "worker-threads-for-queue", defaultQueueWorkers, "The number of threads syncing queue operations. The larger the number, the faster the queue processing, but requires more CPU load.")
 	fs.StringSliceVar(&s.Controllers, "controllers", strings.Split(defaultControllers, ","), fmt.Sprintf("Specify controller gates. Use '*' for all controllers, all knownController: %s ,and we can use "+
 		"'-' to disable controllers, e.g. \"-job-controller,-queue-controller\" to disable job and queue controllers.", knownControllers))
+
+	// PodGroup controller flags (inherit annotations/labels behavior) are defined and maintained in pkg/controllers/podgroup/options.
+	podgroup.AddFlags(fs, &s.PodGroupOptions)
 }
 
 // CheckOptionOrDie checks all options and returns all errors if they are invalid.
--- a/pkg/controllers/framework/interface.go
+++ b/pkg/controllers/framework/interface.go
@@ -23,6 +23,8 @@ import (
 
 	vcclientset "volcano.sh/apis/pkg/client/clientset/versioned"
 	vcinformer "volcano.sh/apis/pkg/client/informers/externalversions"
+
+	podgroup "volcano.sh/volcano/pkg/controllers/podgroup/options"
 )
 
 // ControllerOption is the main context object for the controllers.
@@ -36,10 +38,11 @@ type ControllerOption struct {
 	CronJobWorkerNum        uint32
 	MaxRequeueNum           int
 
-	InheritOwnerAnnotations bool
-	WorkerThreadsForPG      uint32
-	WorkerThreadsForQueue   uint32
-	WorkerThreadsForGC      uint32
+	PodGroupControllerOptions podgroup.Options
+
+	WorkerThreadsForPG    uint32
+	WorkerThreadsForQueue uint32
+	WorkerThreadsForGC    uint32
 
 	// Config holds the common attributes that can be passed to a Kubernetes client
 	// and controllers registered by the users can use it.
--- a/pkg/controllers/podgroup/options/options.go
+++ b/pkg/controllers/podgroup/options/options.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
+	"github.com/spf13/pflag"
+
+	scheduling "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+)
+
+// Options holds configuration for podgroup controller inherit behavior.
+type Options struct {
+	// InheritOwnerAnnotations enables inheriting owner's annotations for pods when creating podgroup.
+	InheritOwnerAnnotations bool
+	// InheritOwnerAnnotationPrefixes defines owner's annotation prefixes to inherit to PodGroup.
+	InheritOwnerAnnotationPrefixes []string
+
+	// InheritPodAnnotations enables inheriting pod annotations when creating podgroup.
+	InheritPodAnnotations bool
+	// InheritPodAnnotationPrefixes defines pod's annotation prefixes to inherit to PodGroup.
+	InheritPodAnnotationPrefixes []string
+
+	// InheritOwnerLabels enables inheriting owner's labels when creating podgroup.
+	InheritOwnerLabels bool
+	// InheritOwnerLabelPrefixes defines owner's label prefixes to inherit to PodGroup.
+	InheritOwnerLabelPrefixes []string
+
+	// InheritPodLabels enables inheriting pod labels when creating podgroup.
+	InheritPodLabels bool
+	// InheritPodLabelPrefixes defines pod's label prefixes to inherit to PodGroup.
+	InheritPodLabelPrefixes []string
+}
+
+// AddFlags adds podgroup controller flags to the specified FlagSet.
+func AddFlags(fs *pflag.FlagSet, o *Options) {
+	fs.BoolVar(&o.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
+	fs.StringArrayVar(&o.InheritOwnerAnnotationPrefixes, "inherit-owner-annotation-prefixes", []string{"volcano.sh/"}, "Owner's annotations prefixes which will be inherited to PodGroup; by default, with 'volcano.sh/' prefix will be inherited to PodGroup")
+	fs.BoolVar(&o.InheritOwnerLabels, "inherit-owner-labels", false, "Enable inherit labels of owner when create podgroup; it is disabled by default")
+	fs.StringArrayVar(&o.InheritOwnerLabelPrefixes, "inherit-owner-label-prefixes", []string{"volcano.sh/"}, "Owner's labels prefixes which will be inherited to PodGroup; by default, with 'volcano.sh/' prefix will be inherited to PodGroup")
+
+	fs.BoolVar(&o.InheritPodAnnotations, "inherit-pod-annotations", true, "Enable inherit annotations of pods when create podgroup; it is enabled by default")
+	fs.StringArrayVar(&o.InheritPodAnnotationPrefixes, "inherit-pod-annotation-prefixes", []string{scheduling.PodPreemptable, scheduling.CooldownTime, scheduling.RevocableZone, scheduling.JDBMinAvailable, scheduling.JDBMaxUnavailable}, "Pod's annotation prefixes which will be inherited to PodGroup; by default, selected volcano.sh/* annotation keys (e.g. pod-preemptable, cooldown-time, revocable-zone, jdb-min-available, jdb-max-unavailable) are inherited")
+	fs.BoolVar(&o.InheritPodLabels, "inherit-pod-labels", true, "Enable inherit labels of pods when create podgroup; it is enabled by default")
+	fs.StringArrayVar(&o.InheritPodLabelPrefixes, "inherit-pod-label-prefixes", []string{scheduling.PodPreemptable, scheduling.CooldownTime}, "Pod's label prefixes which will be inherited to PodGroup; by default, selected volcano.sh/* label keys (e.g. pod-preemptable, cooldown-time) are inherited")
+}
--- a/pkg/controllers/podgroup/pg_controller.go
+++ b/pkg/controllers/podgroup/pg_controller.go
@@ -36,6 +36,7 @@ import (
 	schedulinginformer "volcano.sh/apis/pkg/client/informers/externalversions/scheduling/v1beta1"
 	schedulinglister "volcano.sh/apis/pkg/client/listers/scheduling/v1beta1"
 	"volcano.sh/volcano/pkg/controllers/framework"
+	"volcano.sh/volcano/pkg/controllers/podgroup/options"
 	"volcano.sh/volcano/pkg/features"
 )
 
@@ -67,8 +68,7 @@ type pgcontroller struct {
 	schedulerNames []string
 	workers        uint32
 
-	// To determine whether inherit owner's annotations for pods when create podgroup
-	inheritOwnerAnnotations bool
+	options.Options
 }
 
 func (pg *pgcontroller) Name() string {
@@ -85,7 +85,10 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
 
 	pg.schedulerNames = make([]string, len(opt.SchedulerNames))
 	copy(pg.schedulerNames, opt.SchedulerNames)
-	pg.inheritOwnerAnnotations = opt.InheritOwnerAnnotations
+
+	pg.Options = opt.PodGroupControllerOptions
+
+	klog.Infof("Podgroup Controller Options %+v", pg.Options)
 
 	pg.informerFactory = opt.SharedInformerFactory
 	pg.podInformer = opt.SharedInformerFactory.Core().V1().Pods()
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -219,12 +219,14 @@ func (pg *pgcontroller) updatePodAnnotations(pod *v1.Pod, pgName string) error {
 	return nil
 }
 
-func (pg *pgcontroller) getAnnotationsFromUpperRes(pod *v1.Pod) map[string]string {
-	var annotations = make(map[string]string)
+func (pg *pgcontroller) getAnnotationsLabelsFromUpperRes(pod *v1.Pod) (annotations, labels map[string]string) {
+	annotations = make(map[string]string)
+	labels = make(map[string]string)
 
 	for _, reference := range pod.OwnerReferences {
 		if reference.Kind != "" && reference.Name != "" {
-			tmp := make(map[string]string)
+			tmpAnnotations := make(map[string]string)
+			tmpLabels := make(map[string]string)
 			switch reference.Kind {
 			case "ReplicaSet":
 				rs, err := pg.kubeClient.AppsV1().ReplicaSets(pod.Namespace).Get(context.TODO(), reference.Name, metav1.GetOptions{})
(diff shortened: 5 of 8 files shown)


pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -286,11 +286,55 @@ func (pg *pgcontroller) getMinMemberFromUpperRes(upperAnnotations map[string]str
...
+			}
+		}
+	}
+}
+
+func (pg *pgcontroller) inheritOriginPodLabels(labels map[string]string, obj *scheduling.PodGroup) {
+	if pg.inheritPodLabels {
+		for k, v := range labels {
+			if pg.inheritLabel(k) {
 				obj.Annotations[k] = v
```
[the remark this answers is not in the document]
kingeasternsun: OK

cmd/controller-manager/app/options/options.go
```diff
@@ -77,8 +77,19 @@ type ServerOption struct {
 	EnableHealthz      bool
 	EnableMetrics      bool
 	ListenAddress      string
+
 	// To determine whether inherit owner's annotations for pods when create podgroup
 	InheritOwnerAnnotations bool
+	// To determine whether inherit annotations of pods when create podgroup
+	InheritPodAnnotations bool
+	// To determine whether inherit annotations of pods when create podgroup
```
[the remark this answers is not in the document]
kingeasternsun: ok

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -69,6 +69,18 @@ type pgcontroller struct {
 
 	// To determine whether inherit owner's annotations for pods when create podgroup
 	inheritOwnerAnnotations bool
+
+	// To determine whether inherit annotations of pods when create podgroup
+	inheritPodAnnotations bool
+
+	// To determine whether inherit annotations of pods when create podgroup
```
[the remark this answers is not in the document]
kingeasternsun: ok

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -286,17 +286,58 @@ func (pg *pgcontroller) getMinMemberFromUpperRes(upperAnnotations map[string]str
...
+	}
+	return false
+}
+
 // Inherit annotations from upper resources.
 func (pg *pgcontroller) inheritUpperAnnotations(upperAnnotations map[string]string, obj *scheduling.PodGroup) {
 	if pg.inheritOwnerAnnotations {
 		for k, v := range upperAnnotations {
-			if strings.HasPrefix(k, scheduling.AnnotationPrefix) {
+			if pg.inheritAnnotation(k) {
```
hajnalmt: This means that the owner annotations that are not "volcano.sh/"  will be inherited too, if `inheritAnnotationPrefix` contains a prefix right? So with one variable we control the upper resource annotation inheritance and the pod annotation inheritance too.

I am wondering if it makes sense to have an `inheritOwnerAnnotationPrefix` and an `inheritPodAnnotationPrefix` too instead of just an `inheritAnnotationPrefix`, from design and separation of concern perspective it looks better to me.

The other thing which is wierd for me is that if we implement the inheritLabels on pods why don't we have upper resource label inheritance? It kind of looks natural to me and the api/usage would feel more complete for me that way.

cmd/controller-manager/app/options/options.go
```diff
@@ -77,8 +77,19 @@ type ServerOption struct {
...
+	InheritPodLabels bool
+	// Annotations prefixes which will be inherited to PodGroup
+	InheritAnnotationPrefixes []string
+
+	// Labels prefixes which will be inherited to PodGroup
+	InheritLabelPrefixes []string
```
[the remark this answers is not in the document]
hajnalmt: This is true :)

cmd/controller-manager/app/options/options.go
```diff
@@ -128,6 +139,10 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")
 	fs.StringVar(&s.ListenAddress, "listen-address", defaultListenAddress, "The address to listen on for HTTP requests.")
 	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
+	fs.BoolVar(&s.InheritPodAnnotations, "inherit-pod-annotations", true, "Enable inherit annotations of pods when create podgroup; it is enabled by default")
+	fs.BoolVar(&s.InheritPodLabels, "inherit-pod-labels", true, "Enable inherit labels of pods when create podgroup; it is enabled by default")
+	fs.StringArrayVar(&s.InheritAnnotationPrefixes, "inherit-annotation-prefixes", []string{"volcano.sh"}, "Annotations prefixes which will be inherited to PodGroup; by default, with 'volcano.sh' prefix will be inherited to PodGroup")
```
hajnalmt: I think this should be "volcano.sh/" instead of just "volcano.sh". It's more precise.

pkg/controllers/podgroup/pg_controller_test.go
```diff
@@ -1055,6 +1061,115 @@ func Test_pgcontroller_buildPodGroupFromPod(t *testing.T) {
...
+		c.inheritAnnotationPrefixes = []string{testGroupName}
+		c.inheritLabelsPrefixes = []string{testGroupName}
+
+		// Create pod with OwnerReference to ReplicaSet
+		pod := &v1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      podName + "-3",
+				Namespace: podNamespace,
+				UID:       types.UID("pod-uid-3"),
+				Annotations: map[string]string{
```
hajnalmt, line 1241: Can you add some "volcano.sh/something" annotation and a "volcano.sh/something2" label too? Let's see if the resulting podgroup has these even though you didn't specify the prefix directly.

hajnalmt:
I just want to remark a behavioural change hidden in this PR.
When we parse the networkTopology in L425:
https://github.com/volcano-sh/volcano/blob/0c6f3bf706ad6790b1e806ef5e65c2ffdb3418ba/pkg/controllers/podgroup/pg_controller_handler.go#L425-L434

This happens through annotations:
https://github.com/volcano-sh/volcano/blob/0c6f3bf706ad6790b1e806ef5e65c2ffdb3418ba/pkg/controllers/podgroup/pg_controller_handler.go#L440-L448

Since the default for `InheritPodAnnotations` is true this means, that every pod that has **volcano.sh/network-topology-mode** will update it's podgroup (on the next updateReplicaset or updateStatefulset event) and will propagate this annotation on the podgroup itself which wasn't there before. I can live with it, but it's better to know this side effect.

pkg/controllers/podgroup/pg_controller_handler.go
kingeasternsun: > Thank you very much for your comments.
>
> I had been debating whether to use a single `inheritAnnotationPrefix` or to split it into `inheritOwnerAnnotationPrefix` and `inheritPodAnnotationPrefix`. Personally, I also prefer distinguishing between `inheritOwnerAnnotationPrefix` and `inheritPodAnnotationPrefix`, but I was concerned that introducing too many flags might increase the overall complexity.
```go
	InheritOwnerAnnotations bool
	InheritPodAnnotations bool
	InheritOwnerAnnotationPrefixes []string
	InheritPodAnnotationPrefixes   []string

	InheritOwnerLabels        bool
	InheritPodLabels          bool
	InheritOwnerLabelPrefixes []string
	InheritPodLabelPrefixes   []string
```

>
> Regarding the “upper resource label inheritance” you proposed, I had also been hesitant about whether it should be included in this PR. Since you are also in favor of addressing it together here, I think it makes sense to include it, and I’m OK with that approach.

hzxuzhonghu:
/hold 

let's clarify the use case first

cmd/controller-manager/app/options/options.go
kingeasternsun: Thanks。

cmd/controller-manager/app/options/options.go
```diff
@@ -77,8 +77,27 @@ type ServerOption struct {
...
+
 	// To determine whether inherit owner's annotations for pods when create podgroup
 	InheritOwnerAnnotations bool
+	// To determine which owner's annotations which will be inherited to PodGroup
+	InheritOwnerAnnotationPrefixes []string
+
+	// To determine whether inherit annotations of pods when create podgroup
+	InheritPodAnnotations bool
+	// To determine which pod's annotations will be inherited to PodGroup
+	InheritPodAnnotationPrefixes []string
```
hzxuzhonghu: what if the annotation conflicts? I mean pod conflict with its owner

Another question: if InheritPodAnnotationPrefixes unset, does it mean no annotation will be inherited

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -85,7 +102,25 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
 
 	pg.schedulerNames = make([]string, len(opt.SchedulerNames))
 	copy(pg.schedulerNames, opt.SchedulerNames)
+
 	pg.inheritOwnerAnnotations = opt.InheritOwnerAnnotations
+	pg.inheritOwnerAnnotationPrefixes = make([]string, len(opt.InheritOwnerAnnotationPrefixes))
+	copy(pg.inheritOwnerAnnotationPrefixes, opt.InheritOwnerAnnotationPrefixes)
```
hzxuzhonghu: why need copy? will you mutate it later

pkg/controllers/podgroup/pg_controller.go
```diff
@@ -85,7 +102,25 @@ func (pg *pgcontroller) Initialize(opt *framework.ControllerOption) error {
...
+	pg.inheritPodAnnotations = opt.InheritPodAnnotations
+	pg.inheritPodAnnotationPrefixes = make([]string, len(opt.InheritPodAnnotationPrefixes))
+	copy(pg.inheritPodAnnotationPrefixes, opt.InheritPodAnnotationPrefixes)
```
hzxuzhonghu: ditto

cmd/controller-manager/app/options/options.go
kingeasternsun: > what if the annotation conflicts? I mean pod conflict with its owner

Currently, when there is a conflict, the Pod-level configuration takes precedence over the Owner.
That said, this behavior is not strictly fixed and can be further discussed if needed.


> If InheritPodAnnotationPrefixes unset, does it mean no annotation will be inherited

Yes. If `InheritPodAnnotationPrefixes` is unset, the Pod annotations will not be inherited, while the Owner annotations will still be inherited.
Therefore, the default value for both flags is `[]string{"volcano.sh/"}.`

@hzxuzhonghu

pkg/controllers/podgroup/pg_controller.go
kingeasternsun: > why need copy? will you mutate it later

Yes, the copy here is not strictly necessary.
The reason we used `copy` is to keep the implementation consistent with the existing coding pattern in `pg_controller.go`. For example, `pg.schedulerNames` is also initialized using `copy` earlier:

 ```go
 pg.schedulerNames = make([]string, len(opt.SchedulerNames))
 copy(pg.schedulerNames, opt.SchedulerNames)
 ```

I'm open to simplifying this if consistency with the existing codebase is not a strong requirement. @hzxuzhonghu

pkg/controllers/podgroup/pg_controller_test.go
kingeasternsun: Thanks for you comments, I'll add it later. @hajnalmt

pkg/controllers/podgroup/pg_controller.go
hzxuzhonghu: I think we can simplify

hzxuzhonghu:
make the ci happy, we can merge

kingeasternsun:
> make the ci happy, we can merge

Thanks very much，I's now fixed @hzxuzhonghu, and maybe your `hold` label need to be removed?

kingeasternsun:
@JesseStutler Following the feedback from the weekly meeting, all PodGroup-related flags have been consolidated into the PodGroup controller.

hzxuzhonghu (commented):
/hold cancel

cmd/controller-manager/app/options/options.go
```diff
@@ -127,7 +130,7 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet, knownControllers []string) {
 	fs.BoolVar(&s.EnableHealthz, "enable-healthz", false, "Enable the health check; it is false by default")
 	fs.BoolVar(&s.EnableMetrics, "enable-metrics", false, "Enable the metrics function; it is false by default")
 	fs.StringVar(&s.ListenAddress, "listen-address", defaultListenAddress, "The address to listen on for HTTP requests.")
-	fs.BoolVar(&s.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
+	podgroup.AddFlags(fs, &s.PodGroupOptions)
```
JesseStutler: This seems a bit abrupt. To maintain consistency with `fs.xxxVar`, you could add `podgroup.AddFlags `outside, or at the bottom, add a comment to explain these flags are the controller-related options.

pkg/controllers/framework/interface.go
```diff
@@ -36,10 +36,19 @@ type ControllerOption struct {
 	CronJobWorkerNum        uint32
 	MaxRequeueNum           int
 
-	InheritOwnerAnnotations bool
-	WorkerThreadsForPG      uint32
-	WorkerThreadsForQueue   uint32
-	WorkerThreadsForGC      uint32
+	InheritOwnerAnnotations        bool
```
JesseStutler: Why redefine all the variables when there's already PodGroup.Options available?

pkg/controllers/podgroup/options.go
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright 2019 The Volcano Authors.
```
JesseStutler: 2019 -> 2026

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -219,12 +218,14 @@ func (pg *pgcontroller) updatePodAnnotations(pod *v1.Pod, pgName string) error {
 	return nil
 }
 
-func (pg *pgcontroller) getAnnotationsFromUpperRes(pod *v1.Pod) map[string]string {
+func (pg *pgcontroller) getAnnotationsLabelsFromUpperRes(pod *v1.Pod) (map[string]string, map[string]string) {
```
JesseStutler: You could to define the var name for the return vars. E.g.,
```
func (pg *pgcontroller) getAnnotationsAndLabelsFromOwner(pod *v1.Pod) (annotations, labels map[string]string) {
	annotations = make(map[string]string)
	labels = make(map[string]string)
```

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -219,12 +218,14 @@ func (pg *pgcontroller) updatePodAnnotations(pod *v1.Pod, pgName string) error {
...
 	var annotations = make(map[string]string)
+	var labels = make(map[string]string)
 
 	for _, reference := range pod.OwnerReferences {
 		if reference.Kind != "" && reference.Name != "" {
 			tmp := make(map[string]string)
```
JesseStutler: Also modify the var name to like `tmpAnnotations`

cmd/controller-manager/app/options/options.go
kingeasternsun: Thanks， fixed。

pkg/controllers/framework/interface.go
kingeasternsun: Thanks for your suggestion,  Fixed.

To allow both `pkg/controllers/framework` and `pkg/controllers/podgroup` to reference `podgroup.Options` without introducing a circular dependency, the `options.go` file has been moved into the `pkg/controllers/podgroup/options` subdirectory.

pkg/controllers/podgroup/options.go
kingeasternsun: Fixed.

pkg/controllers/podgroup/pg_controller_handler.go
kingeasternsun: Fixed.

pkg/controllers/podgroup/pg_controller_handler.go
kingeasternsun: Fixed.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -286,17 +297,56 @@ func (pg *pgcontroller) getMinMemberFromUpperRes(upperAnnotations map[string]str
...
+		}
+	}
+	return false
+}
+
 // Inherit annotations from upper resources.
 func (pg *pgcontroller) inheritUpperAnnotations(upperAnnotations map[string]string, obj *scheduling.PodGroup) {
 	if pg.inheritOwnerAnnotations {
 		for k, v := range upperAnnotations {
-			if strings.HasPrefix(k, scheduling.AnnotationPrefix) {
```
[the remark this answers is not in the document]
kingeasternsun: Done.

pkg/controllers/podgroup/pg_controller_test.go
```diff
@@ -973,7 +977,7 @@ func Test_pgcontroller_buildPodGroupFromPod(t *testing.T) {
 		if resultPG.Annotations[scheduling.AnnotationPrefix+"custom-key"] != "custom-value" {
 			t.Errorf("expected annotation %s: 'custom-value', got %s", scheduling.AnnotationPrefix+"custom-key", resultPG.Annotations[scheduling.AnnotationPrefix+"custom-key"])
 		}
-		if resultPG.Annotations[scheduling.JDBMinAvailable] != "1" {
+		if resultPG.Annotations[scheduling.JDBMinAvailable] != "2" {
 			t.Errorf("expected annotation %s: '1', got %s", scheduling.JDBMinAvailable, resultPG.Annotations[scheduling.JDBMinAvailable])
 		}
```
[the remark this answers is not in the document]
kingeasternsun: Fixed.

pkg/controllers/podgroup/pg_controller_test.go
```diff
@@ -1055,6 +1060,210 @@ func Test_pgcontroller_buildPodGroupFromPod(t *testing.T) {
...
+		// Verify inherited annotations and MinMember
+		if resultPG.Spec.MinMember != 3 {
+			t.Errorf("expected MinMember 3, got %d", resultPG.Spec.MinMember)
+		}
+		if resultPG.Annotations[scheduling.AnnotationPrefix+"custom-key"] != "custom-value" {
+			t.Errorf("expected annotation %s: 'custom-value', got %s", scheduling.AnnotationPrefix+"custom-key", resultPG.Annotations[scheduling.AnnotationPrefix+"custom-key"])
+		}
+		if resultPG.Annotations[scheduling.JDBMinAvailable] != "2" {
+			t.Errorf("expected annotation %s: '1', got %s", scheduling.JDBMinAvailable, resultPG.Annotations[scheduling.JDBMinAvailable])
+		}
```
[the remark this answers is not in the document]
kingeasternsun: Fixed.

pkg/controllers/podgroup/pg_controller_test.go
```diff
@@ -1055,6 +1060,210 @@ func Test_pgcontroller_buildPodGroupFromPod(t *testing.T) {
...
+		if _, ok := resultPG.Annotations[scheduling.JDBMaxUnavailable]; ok {
+			t.Error("unexpected JDBMaxUnavailable annotation (should be overwritten by JDBMinAvailable)")
+		}
+
+		if resultPG.Labels[testGroupName+"label-key"] != "label-value" {
+			t.Errorf("expected labels %v 'label-value', got %s", testGroupName+"label", resultPG.Labels[testGroupName+"label"])
+		}
```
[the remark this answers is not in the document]
kingeasternsun: Fixed.

pkg/controllers/podgroup/pg_controller_test.go
```diff
@@ -1055,6 +1060,210 @@ func Test_pgcontroller_buildPodGroupFromPod(t *testing.T) {
...
+		resultPG := c.buildPodGroupFromPod(pod, pgName)
+
+		// Verify annotation is inherited
+		if resultPG.Annotations[testGroupName+"anno"] != "annotation-value" {
+			t.Errorf("expected annotation %v 'annotation-value', got %s", testGroupName+"anno", resultPG.Annotations[testGroupName+"anno"])
+		}
+		// Verify labesl is inherited
+		if resultPG.Labels[testGroupName+"label-key"] != "label-value" {
+			t.Errorf("expected labels %v 'label-value', got %s", testGroupName+"label-key", resultPG.Labels[testGroupName+"label"])
+		}
```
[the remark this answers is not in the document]
kingeasternsun: Fixed.

pkg/controllers/podgroup/options/options.go
hzxuzhonghu, line 1: why not put it under cmd/controller-manager/app/options? flags can be more readable under cmd/options

cmd/controller-manager/app/options/options.go
JesseStutler: Hi @kingeasternsun did you forget to submit? Currently there's no change at all

pkg/controllers/podgroup/options/options.go
```diff
@@ -0,0 +1,59 @@
...
+	// InheritPodLabelPrefixes defines pod's label prefixes to inherit to PodGroup.
+	InheritPodLabelPrefixes []string
+}
+
+// AddFlags adds podgroup controller flags to the specified FlagSet.
+func AddFlags(fs *pflag.FlagSet, o *Options) {
+	fs.BoolVar(&o.InheritOwnerAnnotations, "inherit-owner-annotations", true, "Enable inherit owner annotations for pods when create podgroup; it is enabled by default")
+	fs.StringArrayVar(&o.InheritOwnerAnnotationPrefixes, "inherit-owner-annotation-prefixes", []string{"volcano.sh/"}, "Owner's annotations prefixes which will be inherited to PodGroup; by default, with 'volcano.sh/' prefix will be inherited to PodGroup")
+	fs.BoolVar(&o.InheritOwnerLabels, "inherit-owner-labels", false, "Enable inherit labels of owner when create podgroup; it is disabled by default")
+	fs.StringArrayVar(&o.InheritOwnerLabelPrefixes, "inherit-owner-label-prefixes", []string{"volcano.sh/"}, "Owner's labels prefixes which will be inherited to PodGroup; by default, with 'volcano.sh/' prefix will be inherited to PodGroup")
```
hzxuzhonghu, line 53: can we keep one flag only and support both wildcard and exact match simutaneously. so not to make it so redudant

kingeasternsun: The InheritOwnerLabels flag is necessary. If it is set to false, there is no need to make an additional request to query the Owner.

cmd/controller-manager/app/options/options.go
kingeasternsun: @hzxuzhonghu  Here.  Since these flags are specific to the PodGroup controller, keeping them within the PodGroup controller makes the design clearer and more maintainable.

pkg/controllers/podgroup/options/options.go
kingeasternsun: This refactoring was made based on @JesseStutler ’s review suggestion, and I also think it makes the design more reasonable. Please refer to the comment above for more details.

JesseStutler:
@hajnalmt Mate, I have updated the branch protect rules to add your new CIs, but we can't use `/ok-to-test` to trigger these CIs, what's the reason?

pkg/controllers/podgroup/options/options.go
JesseStutler: My initial thought was that podgroup controllers might not be enabled, and these flags could prevent them from registering. Adding so many startup parameters directly to the vc-controller would seem abrupt. However, Volcano is based on podgroup scheduling, so I think placing the flags within the podgroup controller would improve readability and indicate that these flags are related to the podgroup controller :) But I will be open to other suggestions

hajnalmt:
Hello @JesseStutler,
You need to rebase the PR on master to trigger the workflows again. As far as I see @kingeasternsun already did it  :+1:

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -372,10 +422,12 @@ func (pg *pgcontroller) createOrUpdateNormalPodPG(pod *v1.Pod) error {
 func (pg *pgcontroller) buildPodGroupFromPod(pod *v1.Pod, pgName string) *scheduling.PodGroup {
 	var minMember = int32(1)
 	var ownerAnnotations = make(map[string]string)
-	if pg.inheritOwnerAnnotations {
-		ownerAnnotations = pg.getAnnotationsFromUpperRes(pod)
+	var ownerLabels = make(map[string]string)
+	if pg.InheritOwnerAnnotations || pg.InheritOwnerLabels {
+		ownerAnnotations, ownerLabels = pg.getAnnotationsLabelsFromUpperRes(pod)
 		minMember = pg.getMinMemberFromUpperRes(ownerAnnotations, pod.Namespace, pod.Name)
```
hzxuzhonghu, line 428: If `inherit-owner-annotations` is disabled but `inherit-owner-labels` is enabled, this branch still reads owner annotations and uses them to derive `minMember`. That means owner annotations can still change the PodGroup spec even though the annotation-inheritance feature was explicitly turned off. Before this PR, `minMember` only came from owner annotations when owner-annotation inheritance was enabled. I think `getMinMemberFromUpperRes(...)` should stay gated by `InheritOwnerAnnotations`, or move behind its own dedicated flag if you want that behavior to be independent.

kingeasternsun: - That’s a great point, and I did spend some time thinking about it as well.
- The main reason I chose this approach is to keep the behavior consistent with the existing logic for handling `minMember` and `minResources`.
-  These fields are not strictly inherited as annotations. Instead of copying annotations from the owner to the PodGroup, we read the values from the owner’s annotations and then explicitly set them into the PodGroup’s spec.

PR #4958, opened 2026-01-15
Add sharding docs and fix tc and scheduler issues

Ansita20:
Description

In this version, we introduced the Volcano agent scheduler to support fast scheduling for agent workloads. However, sufficient unit tests (UT) and end-to-end (E2E) tests were not added.

In the related pull requests, several changes were made to the existing batch scheduler, including code abstraction and the introduction of sharding logic. To ensure the batch scheduler continues to function correctly and to allow users to confidently upgrade from v1.13 to v1.14, we need to add comprehensive UT and E2E test coverage.

### DIFF
--- a/docs/user-guide/how_to_configure_scheduler.md
+++ b/docs/user-guide/how_to_configure_scheduler.md
@@ -104,6 +104,163 @@ stops executing the `jobEnqueueableFn` registered in the following plugins and r
 registered in `overcommit` returns a value belows `0`, `jobEnqueueableFn`, which is called in `enqueue` action, will return
 `false` and never call the `jobEnqueueableFn` registered in the `proportion` plugin.
 
+## Sharding Configuration
+
+Volcano supports node sharding to enable multiple schedulers to work on different subsets of cluster nodes. This feature is particularly useful for:
+- Running specialized schedulers for different workload types (e.g., Agentic AI vs batch workloads)
+- Eliminating single-scheduler bottlenecks in large clusters
+- Providing workload isolation and preventing scheduling conflicts
+
+### Sharding Mode
+
+Configure the sharding mode using the `--scheduler-sharding-mode` flag:
+
+```bash
+--scheduler-sharding-mode=<mode>
+```
+
+**Available Modes:**
+
+| Mode | Behavior | Use Case |
+|------|----------|----------|
+| `none` | No sharding (default) | Single scheduler deployment |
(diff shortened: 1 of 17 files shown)


pkg/agent/events/eventsmgr.go
JesseStutler, line 1: You don't need to change this file

pkg/agent/utils/cgroup/cgroup.go
JesseStutler, line 1: Do you have linux env? I understand you can't build the codes on windows or mac if you keep the code, but you shouldn't submit

test/e2e/schedulingaction/sharding.go
```diff
@@ -0,0 +1,183 @@
...
+			Skip("Not enough nodes for sharding E2E test")
+		}
+
+		node1 := nodes.Items[0].Name
+		node2 := nodes.Items[1].Name
+
+		By("Setup NodeShard for the volcano scheduler with only node1")
+		// Default scheduler shard name is 'volcano'
+		shardName := "volcano"
+		nsh := &shardv1alpha1.NodeShard{
```
JesseStutler, line 58: In E2E case, I wish the node shard is created by the sharding controller and have status, not created manually, you can keep only ut tests in this pr, e2e test cases can seperate to another pr and have discuss

example/sharding/README.md
JesseStutler, line 1: We don't need the example dir

docs/user-guide/how_to_configure_scheduler.md
```diff
@@ -104,6 +104,163 @@ stops executing the `jobEnqueueableFn` registered in the following plugins and r
...
+
+```bash
+--scheduler-sharding-name=<shard-name>
+```
+
+The scheduler will watch for a NodeShard resource with the matching name. If not specified, defaults to the scheduler name.
+
+### NodeShard Resource
+
+Create a NodeShard to define which nodes belong to the scheduler:
```
JesseStutler, line 162: This is not correct, the node shard is not created manually, it is created by the sharding controller, please read from the agent scheduler design doc: https://docs.google.com/document/d/1NDTEGulhQ__sdZ_c2kN3ojtJIZie5O7_GyZyKIOLz-g/edit?usp=sharing

pkg/networkqos/tc/tc.go
JesseStutler, line 1: We don't need to change this file

PR #4963, opened 2026-01-16
feat(metrics): Optimize the metrics to monitor task operation performance/e2e task scheduling latecy

### LINKED ISSUE #4930: How can I get the performance metric for the number of Pods the Volcano scheduler can schedule per second

Hi team,

We are currently do some performance tests on Volcano. 
After reviewing the existing metrics in the volcano-scheduler, there doesn't seem to be a direct indicator for scheduling throughput. https://github.com/volcano-sh/volcano/blob/master/docs/design/metrics.md

Specifically, I am looking for a way to measure how many pods the Volcano scheduler can successfully schedule/bind per second. Is there a recommended way to derive this from existing metrics, or am I missing a specific counter?

_No response_

ruanwenjun:
#### What type of PR is this?
feature
 1. Add new metrics task_xx_count to record the task count cached by the scheduler
 2. Add new metrics task_operation_latency_milliseconds to record the cost of task operation
 3. Update task_scheduling_latency_milliseconds to record the real cost from task create to task bound to node

Fixes #4930, #4972

```release-note
NONE
```

### DIFF
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -21,7 +21,7 @@ This metrics track execution of plugins and actions of volcano loop.
 | `e2e_job_scheduling_start_time`           | Gauge           | `job_name`=&lt;job_name&gt;, `queue`=&lt;queue&gt;, `job_namespace`=&lt;job_namespace&gt; | End-to-end job scheduling start time                                           |
 | `plugin_scheduling_latency_milliseconds`  | Histogram       | `plugin`=&lt;plugin_name&gt;, `OnSession`=&lt;OnSession&gt;                               | Plugin scheduling latency in milliseconds                                      |
 | `action_scheduling_latency_milliseconds`  | Histogram       | `action`=&lt;action_name&gt;                                                              | Action scheduling latency in milliseconds                                      |
-| `task_scheduling_latency_milliseconds`    | Histogram       | None                                                                                      | Task scheduling latency in milliseconds                                        |
+| `task_scheduling_latency_milliseconds`    | Histogram       | None                                                                                      | Task scheduling latency in milliseconds (task create to task bound to a node)  |
 
 
 ### volcano operations
@@ -61,6 +61,17 @@ This metrics describe internal state of volcano.
 | `job_retry_counts`                     | Counter         | `job_id`=&lt;job_id&gt;                                           | The number of retry counts for one job        |
 | `job_completed_phase_count`            | Counter         | `job_name`=&lt;job_name&gt; `queue_name`=&lt;queue_name&gt;       | The number of job completed phase             |
 | `job_failed_phase_count`               | Counter         | `job_name`=&lt;job_name&gt; `queue_name`=&lt;queue_name&gt;       | The number of job failed phase                |
+| `task_pending_count`                   | Gauge           | None                                                              | The number of pending task in scheduler       |
+| `task_allocated_count`                 | Gauge           | None                                                              | The number of allocated task in scheduler     |
+| `task_pipelined_count`                 | Gauge           | None                                                              | The number of pipelined task in scheduler     |
+| `task_binding_count`                   | Gauge           | None                                                              | The number of binding task in scheduler       |
+| `task_bound_count`                     | Gauge           | None                                                              | The number of bound task in scheduler         |
+| `task_running_count`                   | Gauge           | None                                                              | The number of running task in scheduler       |
+| `task_releasing_count`                 | Gauge           | None                                                              | The number of releasing task in scheduler     |
+| `task_succeeded_count`                 | Gauge           | None                                                              | The number of succeeded task in scheduler     |
+| `task_failed_count`                    | Gauge           | None                                                              | The number of failed task in scheduler        |
+| `task_unknown_count`                   | Gauge           | None                                                              | The number of unknown task in scheduler       |
+| `task_operation_latency_milliseconds`  | Histogram       | `operation`=&lt;operation_name&gt; `result`=&lt;result&gt;        | Task task operation latency in milliseconds   |
 
 ### volcano Liveness
 Healthcheck last time of volcano activity and timeout
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -218,6 +218,7 @@ type DefaultBinder struct {
 func (db *DefaultBinder) Bind(kubeClient kubernetes.Interface, tasks []*schedulingapi.TaskInfo) map[schedulingapi.TaskID]string {
 	errMsg := make(map[schedulingapi.TaskID]string)
 	for _, task := range tasks {
+		now := time.Now()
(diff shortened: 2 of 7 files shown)


ruanwenjun:
@JesseStutler Please take a look.

JesseStutler:
/cc
Thanks, will take a look a bit later

pkg/scheduler/metrics/metrics.go
```diff
@@ -148,8 +149,35 @@ var (
...
+			Buckets:   prometheus.ExponentialBuckets(32, 2, 10),
+		}, []string{"operation", "result"},
+	)
 )
 
+func RegisterCacheTaskCountFunc(status string, function func() float64) {
+	promauto.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Subsystem: VolcanoSubSystemName,
+			Name:      fmt.Sprintf("task_%s_count", status),
```
hajnalmt, line 167: Ingenious 👍

pkg/scheduler/cache/cache.go
```diff
@@ -1729,6 +1733,25 @@ func (sc *SchedulerCache) GetMetricsData() {
 	sc.setMetricsData(nodeMetricsMap)
 }
 
+func (sc *SchedulerCache) registerCacheMetrics() {
+	for _, status := range schedulingapi.AllTaskStatus() {
+		currentStatus := status
+		statusName := strings.ToLower(currentStatus.String())
+		metrics.RegisterCacheTaskCountFunc(strings.ToLower(statusName), func() float64 {
+			sc.Mutex.Lock()
```
hajnalmt, line 1741: With a lot of jobs in the cluster, holding the cache mutex here during metrics scrapes could become problematic. We have 10 statuses, and we iterate through **all** the jobs in the cache each time and perform a map lookup. This is O(10N) on every Prometheus scrape.

I see 3 options here:
- Adding a `taskCountByStatus map[schedulingapi.TaskStatus]*atomic.Int64` to the cache struct and update it incrementally when task statuses change (where we already hold the lock). Then the metrics collection just reads the atomic counters without needing the lock at all.
- Alternatively, you could create a quick snapshot first to minimize lock hold time
- Or we could make the collection of these metrics optional through a feature flag for clusters that are very performance-sensitive.

I think the atomic counter approach is the cleanest.

hajnalmt (changes requested):
Great work! This is really clever and the metrics will be very useful for monitoring task operations.

I do have some concerns about the lock contention impact in the cache metrics collection (`registerCacheMetrics`).

As a future enhancement (fine for another PR), it would be helpful to have metrics for preBind latency as well since we've seen latency issues there.

pkg/scheduler/cache/cache.go
ruanwenjun: Yes, that's a great discovery. This may cause serious performance issues.
Using a state counter is a good implementation approach, but updating this state counter currently seems inconvenient because there is no unified entry point for modifying the current task state.

JesseStutler:
@ruanwenjun The go lint failed, please check

pkg/scheduler/cache/cache.go
hajnalmt: You can just hook it up in the `addTask` and `deleteTask` functions in `cache/event_handlers.go`, it's a convenient unified entrypoint 😊 and we hold the lock at that point.

ruanwenjun: I find there exist some other way can update the task status e.g. the method `UpdateTaskStatus` at `api/job_info.go`, seems these will not trigger `addTask/deleteTask/updateTask` at event_handlers.

docs/design/metrics.md
```diff
@@ -61,6 +61,17 @@ This metrics describe internal state of volcano.
 | `job_retry_counts`                     | Counter         | `job_id`=&lt;job_id&gt;                                           | The number of retry counts for one job        |
 | `job_completed_phase_count`            | Counter         | `job_name`=&lt;job_name&gt; `queue_name`=&lt;queue_name&gt;       | The number of job completed phase             |
 | `job_failed_phase_count`               | Counter         | `job_name`=&lt;job_name&gt; `queue_name`=&lt;queue_name&gt;       | The number of job failed phase                |
+| `task_pending_count`                   | Gauge           | None                                                              | The number of pending task in scheduler       |
```
JesseStutler, line 64: Could we unify the task_xxx_count metric as a unified metric?
Like 
```
| `task_count` | Gauge | `phase`=<phase_name> | The number of tasks in different phases in scheduler |
```

pkg/scheduler/metrics/metrics.go
```diff
@@ -148,8 +149,35 @@ var (
...
+		}, function,
+	)
+}
+
+func UpdateTaskOperationSuccessLatency(operation string, duration time.Duration) {
+	taskOperationLatency.WithLabelValues(operation, "success").Observe(DurationInMilliseconds(duration))
+}
+
+func UpdateTaskOperationErrLatency(operation string, duration time.Duration) {
+	taskOperationLatency.WithLabelValues(operation, "success").Observe(DurationInMilliseconds(duration))
```
JesseStutler, line 178: Should be "failed"

pkg/scheduler/metrics/metrics.go
```diff
@@ -148,8 +149,35 @@ var (
...
+		prometheus.HistogramOpts{
+			Subsystem: VolcanoSubSystemName,
+			Name:      "task_operation_latency_milliseconds",
+			Help:      "Task operation latency in milliseconds",
+			Buckets:   prometheus.ExponentialBuckets(32, 2, 10),
+		}, []string{"operation", "result"},
+	)
 )
 
+func RegisterCacheTaskCountFunc(status string, function func() float64) {
```
JesseStutler, line 163: Why is here only `RegisterCacheTaskCount` but without Update?

pkg/scheduler/cache/cache.go
```diff
@@ -601,6 +604,7 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, defaultQueu
 	// add all events handlers
 	sc.addEventHandler()
 	sc.HyperNodesInfo = schedulingapi.NewHyperNodesInfo(sc.nodeInformer.Lister())
+	sc.registerCacheMetrics()
```
JesseStutler, line 607: This is not correct updating the metrics in newSchedulerCache, currently the cache is only initialized and has not been started yet

pkg/scheduler/cache/cache.go
JesseStutler: Yes mate is correct, this may cause scheduler performance downgrade, maybe we can do updating metrics in `UpdateTaskStaus`

ruanwenjun: Yes, I have tested, this will cause performance downgrade. I will try to collect these at `UpdateTaskStaus `.

PR #4943, opened 2026-01-13
feat(capacity): add configurable root queue validation control

dafu-wu:
Add rootQueueAutoUpdate configuration parameter to capacity plugin for flexible root queue resource management in hierarchical queue scenarios.

Problem:
In hierarchical queue setups, root queue configuration must satisfy strict constraints (capability/deserved/guarantee >= sum of children's resources). This creates operational complexity and limits flexibility in dynamic cluster environments.

Solution:
Introduce 'rootQueueAutoUpdate' configuration parameter with two modes:

1. Auto-update mode (default, rootQueueAutoUpdate: true):
   - Root queue resources automatically updated to match children's sum
   - Maintains validation checks for data consistency
   - Backward compatible with existing behavior

2. Validation skip mode (rootQueueAutoUpdate: false):
   - Skips validation checks for root queue
   - Allows flexible root queue configuration independent of children
   - Useful for testing, development, or special deployment scenarios

Configuration Example:
  - name: capacity arguments: rootQueueAutoUpdate: false  # Skip root queue validation

Benefits:
- Configuration-driven behavior (no hardcoded logic)
- Backward compatible (default maintains original behavior)
- Reduces operational complexity
- Supports different strategies for different environments
- Clear logging for troubleshooting

Implementation Details:
- Added rootQueueAutoUpdate bool field to capacityPlugin
- Modified New() to read configuration parameter
- Enhanced checkHierarchicalQueue() with conditional validation logic
- Added detailed logging at V(4) level for debugging

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #4973, opened 2026-01-20
Update root queue capability and enhance queue validation logic

guoqinwill:
#### What type of PR is this?
/kind bug

This PR introduces several improvements to the queue management logic in both the scheduler capacity plugin and the admission webhook.

**1. Root Queue Capability Adjustment (Capacity Plugin)**
- Sets the root queue's capability to infinite by default.
- Reason: The root queue is created by default in Volcano, and users generally do not perceive or modify it. It should not act as a constraint that limits the total resources for user job submissions (which should be managed by custom queues).

**2. Enhanced Queue Validation (Admission Webhook)**
- Recursive Capability Validation:
  - Implements recursive lookup for capability checks.
  - Upwards: If a parent queue does not have a configured capability, the validation logic will now recursively check ancestors to find the nearest valid capability limit.
  - Downwards: Similarly, when validating a parent queue, it recursively checks the maximum capability required by its descendants.
- Queue Deletion Protection:
  - added a new webhoook flag: --enable-queue-allocated-pods-check (defaults to false).
  - Behavior: When false, the Webhook will allow queue deletion even if pods are still allocated. If true, will adds a validation rule to intercept and prevent the deletion of a queue if it still contains running jobs/pods.
- Queue Nesting Limit:
  - Added a new webhoook flag:  defaultMaxQueueDepth=5. Introduces a maximum queue nesting depth limit of 5 layers.
  - Reason: Excessively deep queue hierarchies can negatively impact scheduler performance.

Fixes #

```release-note
1. When set the webhoook flag --enable-queue-allocated-pods-check=true, the queue cannot be deleted if it still contains running jobs/pods
2. A new flag "defaultMaxQueueDepth=5" has been added, with the default depth limit is set to 5 as a hard constraint to ensure scheduling efficiency. You can modify this parameter value to adjust the maximum depth of the hierarchical queue.
3. A new flag "enable-root-queue-protection" has been added and default is true, with the root queue's resource attributes (capability, deserved, guarantee) cannot be modified. Since we have set its quota to infinite by default in the capacity plugin, any display modifications by the user should be rejected. If you really want to modify the root's quota, you can set it to false.
```


pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -383,6 +425,103 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+			continue
+		}
+		v := findSubtreeMaxCapability(cq, rname)
+		if v > maxV {
+			maxV = v
+		}
+	}
+
+	return maxV
+}
```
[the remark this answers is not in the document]
guoqinwill: Here, we only need to find the first consultation that has a configured capability and do not need to count the capability configurations of all descendant queues because, every time a queue is created or updated, this recursive check is performed both upwards and downwards. Therefore, in theory, there should not be a situation where an existing descendant queue has a higher quota than the first descendant queue with a configured capability that we have found. So, once we find the first descendant queue that shows a configured capability, we can safely prune early.

JesseStutler:
@guoqinwill The admission policy also contains old validation logic

pkg/scheduler/cache/cache.go
```diff
@@ -448,56 +448,49 @@ func (sc *SchedulerCache) setBatchBindParallel() {
...
-		return !apierrors.IsAlreadyExists(err)
-	}, func() error {
-		_, err := vcClient.SchedulingV1beta1().Queues().Create(context.TODO(), &rootQueue, metav1.CreateOptions{})
-		return err
-	})
-	if err != nil && !apierrors.IsAlreadyExists(err) {
-		panic(fmt.Errorf("failed init root queue, with err: %v", err))
+		if !apierrors.IsNotFound(err) {
+			// Other errors: network problems, permission problems, etc
+			klog.Errorf("failed to get queue %s: %v", name, err)
```
JesseStutler, line 460: Also can return the error instead

pkg/scheduler/cache/cache.go
```diff
@@ -448,56 +448,49 @@ func (sc *SchedulerCache) setBatchBindParallel() {
...
-	defaultQue := vcv1beta1.Queue{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: defaultQueue,
-		},
-		Spec: vcv1beta1.QueueSpec{
-			Reclaimable: &reclaimable,
-			Weight:      1,
-		},
+	if err := createIfNotExists("root", false); err != nil {
+		panic(fmt.Errorf("failed to init root queue: %v", err))
```
JesseStutler: I prefer to use `klog.Fatalf` instead of directly panic

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -587,22 +587,29 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
 	}
 
-	// init root queue: realCapability is set to total resource, and capability/deserved are also set if empty.
+	// init root queue: realCapability is and capability is set to ininity if capability is not set.  deserved are also set if empty.
+	// root queue is default queue and users are not aware of it. User is allowed to sumbit jobs which exceed current cluster total resources,
+	// so root queue should not restrict by current total resources. User need to restrict resource through creating queue manually
 	rootQueueAttr := cp.queueOpts[api.QueueID(cp.rootQueue)]
-	if rootQueueAttr.capability.IsEmpty() {
-		rootQueueAttr.capability = cp.totalResource
+	infinityResourece := api.EmptyResource()
```
JesseStutler: We can wrap it as a func into api package like `api.InfiniteResource()`

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -587,22 +587,29 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
+	infinityResourece.MaxTaskNum = math.MaxInt64
+	infinityResourece.ScalarResources = make(map[v1.ResourceName]float64)
+	if cp.totalResource.ScalarResources != nil {
+		for k := range cp.totalResource.ScalarResources {
+			infinityResourece.ScalarResources[k] = math.MaxInt64
+		}
 	}
-	if rootQueueAttr.deserved.IsEmpty() {
-		rootQueueAttr.deserved = cp.totalResource
+	if rootQueueAttr.capability.IsEmpty() {
```
JesseStutler, line 599: Great, so we don't need this PR now? https://github.com/volcano-sh/volcano/pull/4943, if the user set the capability for the root queue, we use the setting value instead rather than infinity

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -798,16 +801,14 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) {
 
 	// Check if the parent queue's deserved resources are less than the total deserved resources of child queues
 	if attr.deserved.LessPartly(totalDeserved, api.Zero) {
-		klog.V(3).Infof("Sum of child queue deserved (%s) exceeds parent queue %s deserved (%s). "+
-			"This may affect resource distribution during scheduling.",
-			totalDeserved, attr.name, attr.deserved)
+		klog.Errorf("Failed to check queue's hierarchical structure, error: queue <%s> deserved resources <%s> are less than the sum of its child queues' deserved resources <%s>",
```
JesseStutler: I think we can just keep the orginal log format but change info to Error instead?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -768,19 +776,14 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) {
...
 	if attr.name == cp.rootQueue {
-		if attr.guarantee.IsEmpty() {
-			attr.guarantee = totalGuarantee
-		}
-		if attr.deserved.IsEmpty() {
-			attr.deserved = totalDeserved
-		}
+		attr.guarantee = totalGuarantee
+		attr.deserved = totalDeserved
 		cp.totalGuarantee = attr.guarantee
```
JesseStutler: So we even don't need record totalGuarantee and totalDeserved for the capacity plugin, right?

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -81,44 +86,48 @@ func AdmitQueues(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse
...
 			err = validateHierarchicalQueue(queue)
 			if err != nil {
-				break
+				return util.ToAdmissionResponse(err)
+			}
+		}
+
+		// Block attribute modifications to root queue on UPDATE
+		if ar.Request.Operation == admissionv1.Update && queue.Name == "root" && oldQueue != nil {
+			if !equality.Semantic.DeepEqual(queue.Spec.Capability, oldQueue.Spec.Capability) ||
```
JesseStutler, line 104: I think we could allow users to modify capability

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -218,47 +227,70 @@ func validateStateOfQueue(value schedulingv1beta1.QueueState, fldPath *field.Pat
...
+	if len(spec.Deserved) != 0 {
+		for resourceName, quantity := range spec.Deserved {
+			errs = append(errs, k8scorevalid.ValidateResourceQuantityValue(k8score.ResourceName(resourceName), quantity, fldPath.Child("deserved").Child(resourceName.String()))...)
 		}
 	}
 
-	capabilityResource := api.NewResource(resource.Capability)
-	deservedResource := api.NewResource(resource.Deserved)
-	guaranteeResource := api.NewResource(resource.Guarantee.Resource)
+	if len(spec.Guarantee.Resource) != 0 {
```
JesseStutler, line 242: Just want to wonder is `LessPartly` current not supportive enough? I didn't get your point why change the written way

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -383,6 +428,103 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+	case v1.ResourceMemory:
+		q := resource.NewQuantity(int64(value), resource.BinarySI)
+		return q.String()
+
+	default:
+		return fmt.Sprintf("%.2f", value)
+	}
+}
+
+func validateQueueDepth(queue *schedulingv1beta1.Queue) error {
```
JesseStutler, line 504: Whether we need to limit the depth of the hierarchical queue in community? I didn't receive any requirement that we need to limit this

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -383,6 +428,103 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+			res := api.NewResource(pq.Spec.Capability)
+			val := getSingleResource(res, rname)
+			if val > 0 {
+				return val, true
+			}
+		}
+
+		parent = pq.Spec.Parent
+	}
+	return 0, false
```
JesseStutler, line 460: I feel this isn't right. Now that we've changed the root queue to infinite, if our parent queue doesn't have a capability set, and if we eventually recursively call to the root queue, we should return infinite, and the validation should pass.

kingeasternsun:
Great work， thanks。

kingeasternsun:
A great and very reasonable optimization, which is exactly what we were looking for.

JesseStutler:
/hold
Wait @guoqinwill to resolve all the review comments

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -606,38 +575,41 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
 	}
 
-	// init root queue: realCapability is set to total resource, and capability/deserved are also set if empty.
+	// init root queue: realCapability is and capability is set to ininity if capability is not set.  deserved are also set if empty.
+	// root queue is default queue and users are not aware of it. User is allowed to sumbit jobs which exceed current cluster total resources,
+	// so root queue should not restrict by current total resources. User need to restrict resource through creating queue manually
 	rootQueueAttr := cp.queueOpts[api.QueueID(cp.rootQueue)]
-	if rootQueueAttr.capability.IsEmpty() {
-		rootQueueAttr.capability = cp.totalResource
+	infinityResourece := api.EmptyResource()
```
hajnalmt: You have a typo in in this resource too. It should be `infinityResource` instead of `infinityResourece`

pkg/scheduler/cache/cache.go
guoqinwill: The reason we don't return an error directly here is that for a single get operation, we can relax the restrictions. If there is a temporary network fluctuation, a get failure is allowed to continue the process and retry later during the create step below.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: Yes, previously it was not allowed to configure a situation where guarantee/deserved was set without configuring capability. However, capability can be inherited, so this situation should also be allowed to be created.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: Yes. This situation is now allowed, and when iterating to the root, it will return directly without being rejected.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: I understand that allowing too deep levels is not friendly to the scheduler's performance. Perhaps we can add a startup parameter in admission, defaulting to 5, but if users need deeper levels, they can actively modify it.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: Users are not allowed to make modifications because the current queue validations are based on the assumption that the default root queue has unlimited capacity, and most checks related to the root queue essentially allow everything through. If users were allowed to modify it, there could be situations where the root queue is set to have a smaller quota than its sub-queues. Theoretically, this is not allowed, but with the current validation, we do not normally check the capabilities of the root queue as a parent queue. Furthermore, as the root queue is the default root queue created by Volcano, it should not be modified. If users need queue capabilities, they can create sub-queues under the root and set quotas, which fully satisfies usage scenarios. The unlimited mode of the root queue should not be compromised, and I believe there is currently no necessary reason to modify the root queue.

pkg/scheduler/plugins/capacity/capacity.go
guoqinwill: Yes. We don't need to keep it here either; I will delete it later.

pkg/scheduler/plugins/capacity/capacity.go
guoqinwill: ok

pkg/scheduler/plugins/capacity/capacity.go
guoqinwill: ok，done

pkg/scheduler/cache/cache.go
guoqinwill: ok

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -449,84 +591,93 @@ func validateChildAgainstParent(child, parent *schedulingv1beta1.Queue) error {
...
-		// Parent has no explicit limit but has capability - check against capability
-		parentCapability := api.NewResource(parent.Spec.Capability)
-		if parentCapability.LessPartly(childrenSum, api.Zero) {
-			return fmt.Errorf("sum of children's %s (%s) exceeds parent's capability (%s)",
-				resourceType, childrenSum, parentCapability)
+		totalDeserved.Add(api.NewResource(child.Spec.Deserved))
+		if parentDeserved.LessPartly(totalDeserved, api.Zero) {
+			return fmt.Errorf("queue %s validation failed: sum of children's deserved (%s) exceeds parent's deserved limit (%s)",
+				parent.Name, totalDeserved, parentDeserved)
 		}
```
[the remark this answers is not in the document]
guoqinwill: The values of Guarantee and Deserved are different from capability. When the parent queue is not set, they cannot be inherited and default to 0, so this check is necessary here.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -587,22 +586,25 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 			attr.name, attr.allocated.String(), attr.request.String(), attr.inqueue.String(), attr.elastic.String())
 	}
 
-	// init root queue: realCapability is set to total resource, and capability/deserved are also set if empty.
+	// init root queue: realCapability is and capability is set to ininity if capability is not set.  deserved are also set if empty.
+	// root queue is default queue and users are not aware of it. User is allowed to sumbit jobs which exceed current cluster total resources,
+	// so root queue should not restrict by current total resources. User need to restrict resource through creating queue manually
```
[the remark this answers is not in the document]
guoqinwill: The Guarantee and Deserved values will be refreshed in each session, so they will also be set.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -383,6 +424,103 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+	}
+	return 0, false
+}
+
+// Recursively searches for the maximum capability value of a descendant queue
+func findSubtreeMaxCapability(q *schedulingv1beta1.Queue, rname v1.ResourceName) float64 {
+	if q.Spec.Capability != nil {
+		res := api.NewResource(q.Spec.Capability)
+		v := getSingleResource(res, rname)
+		if v > 0 {
```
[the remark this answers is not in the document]
guoqinwill: The reason why q.Spec.Capability cannot be directly used for judgment here is that capability includes multiple resource dimensions, such as CPU and memory, and the final validation result of each dimension after iteration is different. In addition, a configuration of 0 and an unconfigured result are the same, both indicating no limit.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -383,6 +424,103 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+			return 0, false
+		}
+
+		if pq.Spec.Capability != nil {
+			res := api.NewResource(pq.Spec.Capability)
+			val := getSingleResource(res, rname)
+			if val > 0 {
+				return val, true
+			}
+		}
```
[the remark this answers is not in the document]
guoqinwill: The reason why q.Spec.Capability cannot be directly used for judgment here is that capability includes multiple resource dimensions, such as CPU and memory, and the final validation result of each dimension after iteration is different. In addition, a configuration of 0 and an unconfigured result are the same, both indicating no limit.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -768,21 +771,15 @@ func (cp *capacityPlugin) checkHierarchicalQueue(attr *queueAttr) {
 
 		// Check if the parent queue's capability is less than the child queue's capability
 		if attr.capability.LessPartly(childAttr.capability, api.Zero) {
-			klog.V(3).Infof("Child queue %s capability (%s) exceeds parent queue %s capability (%s). "+
+			klog.Errorf("Child queue %s capability (%s) exceeds parent queue %s capability (%s). "+
 				"Child's effective capability will be limited by parent.",
 				childAttr.name, childAttr.capability, attr.name, attr.capability)
```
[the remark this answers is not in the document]
guoqinwill: The log information here indicates that there is an error in the user's queue configuration. Using 'error' makes it more obvious, serving as a reminder for the user to promptly correct the relevant configuration.

pkg/scheduler/cache/cache.go
```diff
@@ -448,56 +448,55 @@ func (sc *SchedulerCache) setBatchBindParallel() {
...
-		return !apierrors.IsAlreadyExists(err)
-	}, func() error {
-		_, err := vcClient.SchedulingV1beta1().Queues().Create(context.TODO(), &rootQueue, metav1.CreateOptions{})
-		return err
-	})
-	if err != nil && !apierrors.IsAlreadyExists(err) {
-		panic(fmt.Errorf("failed init root queue, with err: %v", err))
+		if !apierrors.IsNotFound(err) {
+			// Other errors: network problems, permission problems, etc
+			klog.Errorf("failed to get queue %s: %v", name, err)
```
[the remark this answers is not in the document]
guoqinwill: The reason we don't return an error directly here is that for a single get operation, we can relax the restrictions. If there is a temporary network fluctuation, a get failure is allowed to continue the process and retry later during the create step below.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -449,84 +591,93 @@ func validateChildAgainstParent(child, parent *schedulingv1beta1.Queue) error {
...
+			if parentGuarantee.LessPartly(totalGuarantee, api.Zero) {
+				return fmt.Errorf("parent queue %s validation failed: sum of children's guarantee (%s) exceeds parent's guarantee limit (%s)",
+					parent.Name, totalGuarantee, parentGuarantee)
+			}
+
 			totalDeserved.Add(api.NewResource(sibling.Spec.Deserved))
+			if parentDeserved.LessPartly(totalDeserved, api.Zero) {
+				return fmt.Errorf("parent queue %s validation failed: sum of children's deserved (%s) exceeds parent's deserved limit (%s)",
+					parent.Name, totalDeserved, parentDeserved)
+			}
```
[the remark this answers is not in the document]
guoqinwill: Guarantee, Deserved, and capability are not the same. If they are not set in the parent queue, they will not be inherited and default to 0. Therefore, if the parent queue’s Guarantee/Deserved is not set, its child queues are also not allowed to set them.

pkg/scheduler/cache/cache.go
```diff
@@ -448,56 +448,55 @@ func (sc *SchedulerCache) setBatchBindParallel() {
...
+
+		return retry.OnError(wait.Backoff{
+			Steps:    60,
+			Duration: time.Second,
+			Factor:   1,
+			Jitter:   0.1,
+		}, func(err error) bool {
+			return !apierrors.IsAlreadyExists(err)
+		}, func() error {
+			klog.V(2).Infof("Start to creat queue %s", name)
```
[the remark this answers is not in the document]
guoqinwill: fixed

test/e2e/admission/queue_validation_test.go
```diff
@@ -279,11 +279,11 @@ var _ = ginkgo.Describe("Queue Validating E2E Test", func() {
 
 		_, err := testCtx.Vcclient.SchedulingV1beta1().Queues().Create(context.TODO(), queue, metav1.CreateOptions{})
 		gomega.Expect(err).To(gomega.HaveOccurred())
-		gomega.Expect(err.Error()).To(gomega.ContainSubstring("guarantee should less equal than deserved"))
+		gomega.Expect(err.Error()).To(gomega.ContainSubstring("must be >= guarantee"))
 	})
 
 	// Test resource type validation - resources in deserved/guarantee must exist in capability
```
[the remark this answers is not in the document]
guoqinwill: fixed

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -587,22 +586,25 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
+	infinityResourece := api.InfiniteResource()
+	if cp.totalResource.ScalarResources != nil {
+		for k := range cp.totalResource.ScalarResources {
+			infinityResourece.SetScalar(k, math.MaxInt64)
+		}
 	}
-	if rootQueueAttr.deserved.IsEmpty() {
-		rootQueueAttr.deserved = cp.totalResource
+	if rootQueueAttr.capability.IsEmpty() {
+		rootQueueAttr.capability = infinityResourece
```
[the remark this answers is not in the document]
guoqinwill: fixed

pkg/webhooks/admission/queues/validate/validate_queue.go
JesseStutler: OK, I agree with it, if users wants to set specific capability value for the "root" queue, they could create a new queue under the root queue and set it

JesseStutler: @dafu-wu @hajnalmt Guys WDYT? I think current is fine to set the root queue as infinite, if admins want to set capability, we can create a new sub-root queue under the root queue and set capability for it?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -606,38 +574,37 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
 	}
 
-	// init root queue: realCapability is set to total resource, and capability/deserved are also set if empty.
+	// init root queue: realCapability is and capability is set to ininity if capability is not set.  deserved are also set if empty.
+	// root queue is default queue and users are not aware of it. User is allowed to sumbit jobs which exceed current cluster total resources,
+	// so root queue should not restrict by current total resources. User need to restrict resource through creating queue manually
 	rootQueueAttr := cp.queueOpts[api.QueueID(cp.rootQueue)]
-	if rootQueueAttr.capability.IsEmpty() {
-		rootQueueAttr.capability = cp.totalResource
+	infinityResource := api.InfiniteResource()
```
JesseStutler, line 593: Why not to directly put initialization for the scalarResources in `InfiniteResource()`?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -606,38 +574,37 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
...
-	rootQueueAttr.realCapability = cp.totalResource
+	rootQueueAttr.realCapability = rootQueueAttr.capability
 	// checkHierarchicalQueue only logs warnings and never returns errors
 	// to avoid aborting the entire scheduling cycle due to configuration issues
 	cp.checkHierarchicalQueue(rootQueueAttr)
 
-	// update session attributes
-	ssn.TotalGuarantee = cp.totalGuarantee
-	ssn.TotalDeserved = cp.totalDeserved
+	klog.V(4).Infof("Successfully checked queue's hierarchical structure.")
```
JesseStutler: Currently `checkHierarchicalQueue` will not return an error, therefore although if check failed in `checkHierarchicalQueue`, this successfully log will also be printed, you should put the log into `checkHierarchicalQueue` and add some judgements to print it if succeed

JesseStutler:
Also /cc again @kingeasternsun @hajnalmt @dafu-wu

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -218,47 +223,70 @@ func validateStateOfQueue(value schedulingv1beta1.QueueState, fldPath *field.Pat
...
-	guaranteeResource := api.NewResource(resource.Guarantee.Resource)
+	if len(spec.Guarantee.Resource) != 0 {
+		for resourceName, guaranteeQ := range spec.Guarantee.Resource {
+			errs = append(errs, k8scorevalid.ValidateResourceQuantityValue(
+				k8score.ResourceName(resourceName), guaranteeQ,
+				fldPath.Child("guarantee").Child("resource").Child(resourceName.String()))...,
+			)
+
+			desQ, exists := spec.Deserved[resourceName]
+			if !exists {
```
JesseStutler, line 250: Do we need to return an error here? Will some users have configured fewer dimensions for deserved than guarantee?

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -383,6 +424,103 @@ func listQueueChild(parentQueueName string) ([]string, error) {
...
+	default:
+		if v, ok := r.ScalarResources[name]; ok {
+			return v
+		}
+	}
+	return 0
+}
+
+// Recursively searches for the ancestor queue that recently defines the capability
+func findNearestAncestorCapability(q *schedulingv1beta1.Queue, rname v1.ResourceName) (float64, bool) {
```
JesseStutler, line 442: Do we have test cases for the hierarchical capability check currently, if we don't have, we shoud add it

pkg/scheduler/plugins/capacity/capacity.go
guoqinwill: As a generic function in the `api` package, `InfiniteResource()` cannot and should not know which Scalar Resources exist in the current cluster. The "infinite scalability of cluster-specific scalable resources" needs to be supplemented by the caller (in this case, the `capacity` plugin) based on the actual cluster information.

pkg/scheduler/plugins/capacity/capacity.go
guoqinwill: You are correct, but `checkHierarchicalQueue` is called in a loop, making it impossible to determine the final termination condition. Printing "verification successful" for each queue would generate too much data. Therefore, I will delete this log.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: Errors will be returned here. For each resource dimension where the guarantee is configured but the deserved resource is not, the error will be recorded, and finally, an error will be returned.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: ok，i will add later

pkg/scheduler/plugins/capacity/capacity.go
JesseStutler: OK, I mad a mistake

dafu-wu:
This is a really good feature,  quick question: we could add a feature flag or configuration option that allows users to specify which resources can be updated automatically. This way, we can maintain compatibility with legacy users who rely on root-level auto-update behavior.

guoqinwill:
> This is a really good feature, quick question: we could add a feature flag or configuration option that allows users to specify which resources can be updated automatically. This way, we can maintain compatibility with legacy users who rely on root-level auto-update behavior.

Okay. Although I think modifying the properties of the root queue is an undesirable action, for the sake of compatibility, I added a flag "enable-root-queue-protection." When it is manually set to false, property modifications to the root queue will not be checked. For future use, to prevent users from making mistakes with the root queue, modifying the root queue is disabled by default.

pkg/webhooks/admission/queues/validate/validate_queue.go
guoqinwill: Considering @dafu-wu ’s proposal, I also provided a flag "enable-root-queue-protection". When it is manually set to false, property modifications to the root queue will not be checked. For future use, to prevent users from making mistakes with the root queue, modifying the root queue is disabled by default.

cmd/webhook-manager/app/options/options.go
```diff
@@ -92,6 +100,9 @@ func (c *Config) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&c.EnableHealthz, "enable-healthz", false, "Enable the health check; it is false by default")
 	fs.StringVar(&c.HealthzBindAddress, "healthz-address", defaultHealthzAddress, "The address to listen on for the health check server.")
 	fs.DurationVar(&c.GracefulShutdownTime, "graceful-shutdown-time", defaultGracefulShutdownTime, "The duration to wait during graceful shutdown before forcing termination.")
+	fs.BoolVar(&c.EnableQueueAllocatedPodsCheck, "enable-queue-allocated-pods-check", false, "If true, queue deletion will be rejected when the queue has allocated pods.")
```
hajnalmt, line 103: I think people want to generally avoid this. Shouldn't we switch to `true` by default?

guoqinwill:
> Thank you so much for doing this guys! Great work. (Sorry I didn't submit my review on this)
> 
> I think these can be closed (what do you think?): #4943 #4662
> 
> Since this was a pretty painful issue in older releases, do we want to backport the fix? It seems worthwhile, although I know it's risky to backport such big changes, but this is well implemented to do this I think.
> 
> One additional thought: should we prevent users from submitting jobs directly into the root queue? Either by making the root queue “closed” by default, and/or enforce it in admission. This wouldn’t be API-breaking, rather it would be a behavior change; and to me it aligns with the intent of the root queue not being a user-facing submission target.

In fact, the webhook for admit_job already has this check and does not allow submitting jobs to the root queue：
https://github.com/guoqinwill/volcano/blob/207e85ac458839a6eeeb985ea8fc2258629f663b/pkg/webhooks/admission/jobs/validate/admit_job.go#L214-L215

In addition, I'm not sure whether this submission is appropriate to revert to the previous version; perhaps this needs to be assessed by community planners.

cmd/webhook-manager/app/options/options.go
guoqinwill: Yes. I once tried setting this flag to true by default, but Spark's CI failed because this CI does not wait for the queue to be deleted when cleaning up jobs. I can't fix the upper-level Spark tasks the way we fixed our e2e, so I haven't set it to true for now. If there is an opportunity later or if other contributors are willing, they could help modify this CI before this flag can be enabled.

JesseStutler:
It's a huge change, 

> Thank you so much for doing this guys! Great work. (Sorry I didn't submit my review on this)
> 
> I think these can be closed (what do you think?): #4943 #4662
> 
> Since this was a pretty painful issue in older releases, do we want to backport the fix? It seems worthwhile, although I know it's risky to backport such big changes, but this is well implemented to do this I think.
> 
> One additional thought: should we prevent users from submitting jobs directly into the root queue? Either by making the root queue “closed” by default, and/or enforce it in admission. This wouldn’t be API-breaking, rather it would be a behavior change; and to me it aligns with the intent of the root queue not being a user-facing submission target.

I agree with you mate because it's a worthwhile change. But there might be code conflicts. I agree with cherry-picking release-1.14 since the admission code might not have many conflicts( but this PR includes some new flags we should mention in the release note), but the previous code will likely have conflicts, which will involve manually resolving them, and that's quite risky. @hajnalmt WDYT?

cmd/webhook-manager/app/options/options.go
hajnalmt: Thank you for the explanation! I have created an issue for this. Let's see if someone picks this up.
https://github.com/volcano-sh/volcano/issues/5079

hajnalmt:
/cherry-pick release-1.14

@JesseStutler Yeah you are right, it would be hard to cherry-pick it to previous versions, I am fine with 1.14 only.

> In fact, the webhook for admit_job already has this check and does not allow submitting jobs to the root queue：
> https://github.com/guoqinwill/volcano/blob/207e85ac458839a6eeeb985ea8fc2258629f663b/pkg/webhooks/admission/jobs/validate/admit_job.go#L214-L215

Whao @guoqinwill this is true, I checked the podgroup only when I looked for this. We shall put this into the podgroup controller too I think, so standard kubernetes workloads are taken into account too.

Issue #4946, opened 2026-01-14
Rethink podInformer event handling, change syncTask communication to podInformer (dra tests failing)


hajnalmt:
### What is the problem you're trying to solve

**The problem:**
In https://github.com/volcano-sh/volcano/pull/4938 we changed the syncTask communication to use the podInformer instead of a direct API call. But it looks like we can't do this as dra tests started to fail. This means that our current way of the syncTask cache handling needs be rethought, or our eventHandlers on the podInformer are not good enough and our podInformer cache is not working properly.

But we can at least figure this out through a problem we see, and can reproduce with a test case now.
An analysis is needed here and a viable solution proposed.

I want to link the `syncNode` issue here too:
https://github.com/volcano-sh/volcano/issues/4888 
As the two problem's culprit might be the same, maybe we can separate the logic that handles updates and delete,
https://github.com/volcano-sh/volcano/issues/4888#issuecomment-3696311298 can solve this too. 

### Describe the solution you'd like

I want to have an analysis that what's causing this issue, and a solution propsed to mitigate it.

- [ ] Revert the original podInformer change
- [ ] Prepare an analisys about the problem.
- [ ] Solution proposed

### Documentation Updates

- [ ] This feature requires design or user documentation changes.
- [ ] If documentation changes are required, I will ensure the relevant documents are updated and published to the Volcano official website (https://volcano.sh) via the [volcano-sh/website](https://github.com/volcano-sh/website) repository.

JesseStutler:
Such a weird thing why it will cause DRA test failed?

hajnalmt:
Let's see if someone picks it up in the community, we need to do an investigation.
It's really not a good sign, that our podInformer thinks that pods are allocated even though they aren't. 😕

PR #4982, opened 2026-01-22
fix(scheduler): add nil checks in preemptable/reclaimable plugin func…

goyalpalak18:
## Description
I’ve put this PR together to fix a critical **nil pointer dereference** that’s been taking down the scheduler. It’s happening in the `gang`, `proportion`, and `capacity` plugins whenever the system encounters "orphaned" tasks—basically pods that are still hanging around on a node after their parent `PodGroup` has been deleted.

---

### The Problem
While digging through the logs, I found a race condition between the **Cache** and the **Session**. During high-churn events (like when jobs are being deleted rapidly), things get out of sync:

1. **Cache State:** The `PodGroup` is deleted and removed from the cache.
2. **Session Snapshot:** During `ssn.Open()`, Volcano clones the nodes and tasks but filters out any Jobs that no longer have a valid `PodGroup`.
3. **The Crash:** When plugins try to look up the parent job via `ssn.Jobs[task.Job]`, they receive `nil`. When I then try to dereference that (like checking `job.UID`), the scheduler panics and crashes.

---

### The Fix
I decided to go with a **"fail-soft"** approach by adding explicit nil-check guards within the plugin callbacks (`preemptableFn` and `ReclaimableFn`). 

Now, if I run into a task where the parent job is missing from the session, the scheduler will:
- **Log a debug message** so we can identify the orphaned task.
- **Gracefully skip** the task in the current logic instead of crashing.
- **Stay stable** by only operating on valid session data until the next sync.

I chose to implement these at the plugin level rather than the core session level. I think this is cleaner because it keeps the main scheduler engine simple and lets me tailor the logging to each specific plugin.

---

## Changes
- **Files affected:**
    - `pkg/scheduler/plugins/gang/gang.go`
    - `pkg/scheduler/plugins/proportion/proportion.go`
    - `pkg/scheduler/plugins/capacity/capacity.go`
- **Logic:** I added defensive nil-check guards before accessing job/queue attributes.
- **Pattern:** Follows a "skip-and-log" approach for orphaned resources until the next cache synchronization.

---

## Test Results
- **Environment:** 5-node cluster with high-frequency Gang Job submissions and deletions.
- **Action:** I manually forced `PodGroup` deletions during active preemption cycles to try and trigger the race.
- **Result:** - **Before Fix:** The scheduler crashed in about 30% of my high-load test scenarios.
    - **After Fix:** **0 crashes** observed over 100+ test cycles.
- **Unit Tests:** `go test -v ./pkg/scheduler/plugins/...` passed for all three plugins.

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -120,6 +120,11 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 			}
 
 			attr := cp.queueOpts[job.Queue]
+			if attr == nil {
+				klog.Warningf("[capacity] Skip reclaimee <%s/%s>: queue <%s> not found in queueOpts",
+					reclaimee.Namespace, reclaimee.Name, job.Queue)
+				continue
+			}
 
 			klog.V(5).Infof("[capacity] Considering reclaimee <%s/%s> from queue <%s> for reclaimer <%s/%s>.",
 				reclaimee.Namespace, reclaimee.Name, attr.queueID, reclaimer.Namespace, reclaimer.Name)
@@ -376,6 +381,11 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 				return
 			}
 			attr := cp.queueOpts[job.Queue]
+			if attr == nil {
+				klog.Warningf("[capacity] Skip allocate event for task <%s/%s>: queue <%s> not found in queueOpts",
+					event.Task.Namespace, event.Task.Name, job.Queue)
+				return
+			}
 			attr.allocated.Add(event.Task.Resreq)
 			metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory, attr.allocated.ScalarResources)
 
@@ -398,6 +408,11 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 				return
 			}
 			attr := cp.queueOpts[job.Queue]
+			if attr == nil {
+				klog.Warningf("[capacity] Skip deallocate event for task <%s/%s>: queue <%s> not found in queueOpts",
+					event.Task.Namespace, event.Task.Name, job.Queue)
+				return
+			}
 			attr.allocated.Sub(event.Task.Resreq)
 			metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory, attr.allocated.ScalarResources)
 
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -288,6 +288,11 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 			}
 
 			attr := pp.queueOpts[job.Queue]
+			if attr == nil {
+				klog.Warningf("[proportion] Skip reclaimee <%s/%s>: queue <%s> not found in queueOpts",
+					reclaimee.Namespace, reclaimee.Name, job.Queue)
+				continue
+			}
 
 			if _, found := allocations[job.Queue]; !found {
 				allocations[job.Queue] = attr.allocated.Clone()
@@ -468,6 +473,11 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 				return
 			}
 			attr := pp.queueOpts[job.Queue]
+			if attr == nil {
+				klog.Warningf("[proportion] Skip allocate event for task <%s/%s>: queue <%s> not found in queueOpts",
+					event.Task.Namespace, event.Task.Name, job.Queue)
+				return
+			}
 			attr.allocated.Add(event.Task.Resreq)
 			metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory, attr.allocated.ScalarResources)
 
@@ -484,6 +494,11 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 				return
 			}
 			attr := pp.queueOpts[job.Queue]
+			if attr == nil {
+				klog.Warningf("[proportion] Skip deallocate event for task <%s/%s>: queue <%s> not found in queueOpts",
+					event.Task.Namespace, event.Task.Name, job.Queue)
+				return
+			}
 			attr.allocated.Sub(event.Task.Resreq)
 			metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory, attr.allocated.ScalarResources)
 


goyalpalak18:
Hi @hajnalmt @JesseStutler  Updated based on Gemini's feedback - added plugin prefixes to logs. Also extended the fix to cover event handlers and other plugins (priority, drf, nodegroup) that had the same issue. Should be comprehensive now.

hajnalmt (changes requested):
/priority awaiting-more-evidence
/ok-to-test
/kind bug

Thanks for the contribution! This looks quite serious. I see you already checked other plugins, thank you and props for doing that! I did a quick search as well, and it seems the network-topology-aware plugin likely has the same issue, since it also dereferences `Jobs` for the `subJob`:
https://github.com/volcano-sh/volcano/blob/27f4d05522735f1b76e9c104efa4cada1d15666d/pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go#L322-L326
Please include a fix for that in this PR and update the PR description accordingly.

Could you also share the scheduler logs showing the panic? We should open an issue for this so others can find it easily when searching for the error message.

I think your approach makes sense totally, better to be safe than sorry. I also have a hunch that the underlying cause of how we end up in this state might be related to https://github.com/volcano-sh/volcano/issues/4946 and how we handle the informer cache.

JesseStutler:
The PR has building failure @goyalpalak18

goyalpalak18:
Hi @hajnalmt and @JesseStutler 
I’ve updated the PR to extend the nil checks to the network-topology-aware plugin and amended the commit (force-pushed).
The change is purely defensive to prevent scheduler panics. Please let me know if anything else is needed.

pkg/scheduler/plugins/nodegroup/nodegroup.go
```diff
@@ -260,22 +260,32 @@ func (np *nodeGroupPlugin) OnSessionOpen(ssn *framework.Session) {
...
 		return score, nil
 	}
 	ssn.AddNodeOrderFn(np.Name(), nodeOrderFn)
 
 	predicateFn := func(task *api.TaskInfo, node *api.NodeInfo) error {
 		job := ssn.Jobs[task.Job]
+		if job == nil {
+			klog.V(4).Infof("[nodegroup] Skip predicate for task <%s/%s>: job <%s> not found in session (orphaned task from deleted PodGroup)",
+				task.Namespace, task.Name, task.Job)
+			return fmt.Errorf("job %s not found in session", task.Job)
```
hajnalmt, line 284: The CI Fails here because fmt is not imported.
```
Error: 97.50 pkg/scheduler/plugins/nodegroup/nodegroup.go:283:11: undefined: fmt
105.4 make: *** [Makefile:76: vc-scheduler] Error 1
------
Dockerfile:22
--------------------
  20 |     RUN go mod download
  21 |     ADD . volcano
  22 | >>> RUN cd volcano && make vc-scheduler
  23 |     
  24 |     FROM alpine:latest
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c cd volcano && make vc-scheduler" did not complete successfully: exit code: 2
make: *** [Makefile:120: vc-scheduler-image] Error 1
```

goyalpalak18:
Hey @hajnalmt, thanks for the catch on the import— fixed that in nodegroup.go so the CI should be green now.

I also went through and added the specific plugin prefixes (like [gang], [proportion], etc.) to the logs to make them easier to track.

Regarding the panic info you wanted:

The Panic: It was a nil pointer dereference in preemptableFn. It happens when a PodGroup is deleted mid-cycle, leaving the session job lookup empty. The stack trace looks like this:
goroutine 1 [running]:
volcano.sh/volcano/pkg/scheduler/plugins/gang.(*gangPlugin).preemptableFn(...)
        pkg/scheduler/plugins/gang/gang.go:255

Reproduction: I triggered it by manually deleting a PodGroup right as a preemption/reclaim loop started. Since the task is still there but the job is gone from the session, it hits that nil entry.

Let me know if you need anything else!

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -113,7 +113,19 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
 		for _, reclaimee := range reclaimees {
 			job := ssn.Jobs[reclaimee.Job]
+			if job == nil {
+				klog.V(4).Infof("[capacity] Skip reclaimee <%s/%s>: job <%s> not found in session (orphaned task from deleted PodGroup)",
+					reclaimee.Namespace, reclaimee.Name, reclaimee.Job)
+				continue
+			}
+
 			attr := cp.queueOpts[job.Queue]
+			if attr == nil {
```
hajnalmt, line 123: I have a little debate here. Let me think on this, wheter we need to protect against this too or not. In theory this should not happen, as a queue is not deletable from admission point of view if it has a job in it. If we need this then we probably need the guard at other places too.

goyalpalak18:
@hajnalmt Thank you for the approval!                                                                                                                    
                                                                                                                                                           
  **On the queue guards:** I agree it's better to be safe. While the queue lookup returning nil might be rare, keeping the checks ensures fail-soft        
  behavior — the scheduler logs a warning and continues rather than crashing. I'd prefer to keep them as a defensive measure.                              
                                                                                                                                                           
  **On the stack trace**,  unfortunately I no longer have access to the full trace — the test cluster and terminal session are gone. Here's what I captured:
                                                                                                                                                           
  goroutine 1 [running]:                                                                                                                                   
  volcano.sh/volcano/pkg/scheduler/plugins/gang.(*gangPlugin).preemptableFn(...)                                                                           
      pkg/scheduler/plugins/gang/gang.go:255                                                                                                               
                                                                                                                                                           
  **Root cause:** When a PodGroup is deleted mid-cycle, `ssn.Jobs[task.Job]` returns nil for orphaned tasks, and the subsequent access to `job.UID` or     
  `job.Queue` causes the panic.                                                                                                                            
                                                                                                                                                           
  If you need the complete trace for the issue, I can spin up a cluster and reproduce it — just let me know.

goyalpalak18:
Just pushed an update — found I'd missed the same nil check issue in DRF's preemptableFn. All plugins should be covered now.

pkg/scheduler/plugins/capacity/capacity.go
hzxuzhonghu: @hajnalmt can you point it `In theory this should not happen, as a queue is not deletable from admission point of view if it has a job in it.`

goyalpalak18:
> /remove priority/awaiting-more-evidence
> 
> Thank you for the changes! I am fine without a reproduction as the change is clear. On the other hand, I came to the conclusion that we don't need the queue guards, our API definition has default settings. We should trust that and shouldn't introduce unnecessary guards, it's enough to check if job is nil I think. https://github.com/volcano-sh/volcano/blob/master/staging/src/volcano.sh/apis/pkg/apis/batch/v1alpha1/job.go#L88-L93
> 
> https://github.com/volcano-sh/volcano/blob/e07e2d35fb055ee9558a3bf743270873c782f131/staging/src/volcano.sh/apis/pkg/apis/batch/v1alpha1/job.go#L88-L93
> 
> It's my opinion only and I am not removing the approve on the PR just because of this.
> 
> @JesseStutler or @hzxuzhonghu If you have some time please review this.

@hajnalmt Thanks for the suggestion! I went ahead and removed the extra queue guards. The code is much cleaner now and just checks if preemptorJob is nil.

goyalpalak18:
> This looks like a serious problemm , this kind of fix can prevent panic. But seems there is a big escalation risk, say sometimes a pod can be preempted, but if user removes its PG, then it can be skipped

@hzxuzhonghu Good catch. That is a valid concern I didn't fully think through.

My main goal was just to stop the panic, but you're right—simply skipping these might mess up preemption logic if a PodGroup gets deleted at the wrong time.

I see a few ways to handle this:
1. Log it: Keep the current fix (skip the nil job) but add a warning log so we at least know it's happening.
2. Force Preemption: Treat these nil-job tasks as "lowest priority" so they get evicted first, rather than preserved.

Which approach do you prefer? I'm happy to implement whichever one is safer for the project.

pkg/scheduler/plugins/capacity/capacity.go
hajnalmt: Whao, you are right @hzxuzhonghu. I was wrong here, thank you for pointing this out.
https://github.com/volcano-sh/volcano/blob/67b91e417e761a85ee848993c241fef282fb263f/pkg/webhooks/admission/queues/validate/validate_queue.go#L108-L113

We are just checking if the queue doesn't have childqueues and that's all.
https://github.com/volcano-sh/volcano/blob/67b91e417e761a85ee848993c241fef282fb263f/pkg/webhooks/admission/queues/validate/validate_queue.go#L267-L294

We don't delete the jobs associated with the queue too when it gets deleted:
https://github.com/volcano-sh/volcano/blob/67b91e417e761a85ee848993c241fef282fb263f/pkg/controllers/queue/queue_controller_handler.go#L46-L65
I think we should have this logic in the queue admission controller. I see two direction one is to delete the podgroups in the queue or to not allow the deletion of the podgroup in general.

goyalpalak18:
@hajnalmt @hzxuzhonghu - pushed another update. 

I reinstated the `attr == nil` queue guards in both the proportion and capacity plugins (ReclaimableFn, AllocateFunc, DeallocateFunc) with `Warningf` logging. 

@hzxuzhonghu, you were right about the admission controller behavior. Since it doesn't protect against queue deletion when jobs are still running, `queueOpts[job.Queue]` returning nil is a real scenario we need to catch.

(Also, as a quick heads-up: I ignored the automated Gemini/Copilot review comments in the thread. They were flagging an older diff, and the prefix logging and nil checks they suggested are already in the current code.)

pkg/scheduler/plugins/capacity/capacity.go
hzxuzhonghu: we can track it in a separate issue

goyalpalak18:
@hzxuzhonghu Just a heads-up, I updated the branch to sync it with master and the bot automatically dropped the /lgtm label. When you get a chance, would you kindly reapply? Thank you!

hzxuzhonghu:
Yeah, you should remove the merge commit. By using git rebase i guess

goyalpalak18:
Fixed. I dropped the merge commit, rebased, and force-pushed. The history should be perfectly clean again.

hzxuzhonghu (approved):
/lgtm

Thanks everyone working here

pkg/scheduler/plugins/capacity/capacity.go
hzxuzhonghu: @goyalpalak18 can you help file an issue or above comment

Issue #4957, opened 2026-01-15
Urgent - Add ut tests to cover modifications to the existing Batch Scheduler code after Agent Scheduler merged


JesseStutler:
### What is the problem you're trying to solve

https://github.com/volcano-sh/volcano/pull/4804
https://github.com/volcano-sh/volcano/pull/4848
In this version, we have introduced Volcano agent scheduler to support fast scheduling for agent workloads, However, we haven't added enough test cases (UT) and end-to-end (E2E) tests. In these two pull requests, we made some changes to the existing batch scheduler, including code abstraction and adding shards logic. We need to add UT test cases and e2e test cases to ensure the batch scheduler remains correct, so users can confidently upgrade from 1.13 to 1.14.

### Describe the solution you'd like

We need to add UT test cases and e2e test cases to ensure the batch scheduler remains correct, This issue can support to add UT tests first

### Documentation Updates

- [ ] This feature requires design or user documentation changes.
- [ ] If documentation changes are required, I will ensure the relevant documents are updated and published to the Volcano official website (https://volcano.sh) via the [volcano-sh/website](https://github.com/volcano-sh/website) repository.

JesseStutler:
/help
Because I'm still working on other features, I'll leave it to other contributors first, but please note that the deadline is before the end of the month.
/priority high

Al-Masih-ad-Dajjal:
will it be okay if i pick this ?

JesseStutler:
@Al-Masih-ad-Dajjal @AR21SM Thanks for your attention, but I have to clarify that this requirement will be very urgent and we need to done it before this month, this is the agent scheduling's design, which will help you to know the design and background: https://docs.google.com/document/d/1NDTEGulhQ__sdZ_c2kN3ojtJIZie5O7_GyZyKIOLz-g/edit?usp=sharing

JesseStutler:
https://github.com/volcano-sh/volcano/pull/4804
https://github.com/volcano-sh/volcano/pull/4848

In this two PRs we have made some intrusive modifications to the existing batch scheduler, so I expect to cover these unit tests:

- [x] Predicates plugin (some are already present, but I'm unsure if there will be any uncovered use cases after abstraction)
- [x] Nodeorder plugin
- [x] Places where node sharding is introduced, such as the allocate action, preempt/reclaim action, etc.

katara-Jayprakash:
/assign 
[...]

PR #4983, opened 2026-01-23
Align implementation of InterPodAffinity, TaintToleration, PodTopologySpread plugins with other plugins in nodeorder and add unit tests

JesseStutler:
#### What type of PR is this?
/kind feature

Add more unit tests in this version after we merged: 
https://github.com/volcano-sh/volcano/pull/4804
https://github.com/volcano-sh/volcano/pull/4848
Besides, current implementation of InterPodAffinity, TaintToleration, PodTopologySpread and doesn't align with other plugins, they should be added to scorePlugins map when they're successfully initialized, and they should be initialized only weight is not 0

Related https://github.com/volcano-sh/volcano/issues/4957

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder_test.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder_test.go
@@ -21,6 +21,16 @@ import (
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+	k8sframework "k8s.io/kube-scheduler/framework"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/imagelocality"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/interpodaffinity"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodeaffinity"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/noderesources"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/podtopologyspread"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/tainttoleration"
 	k8smetrics "k8s.io/kubernetes/pkg/scheduler/metrics"
 
 	schedulingv1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
@@ -30,6 +40,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
+	"volcano.sh/volcano/pkg/scheduler/plugins/util/k8s"
 	"volcano.sh/volcano/pkg/scheduler/uthelper"
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
@@ -42,8 +53,14 @@ func TestMain(m *testing.M) {
 
 type nodeOrderTestCase struct {
 	uthelper.TestCommonStruct
-	LeastRequestedWeight int
-	MostRequestedWeight  int
+	LeastRequestedWeight    int
+	MostRequestedWeight     int
+	BalancedResourceWeight  int
+	NodeAffinityWeight      int
+	TaintTolerationWeight   int
+	PodAffinityWeight       int
+	PodTopologySpreadWeight int
+	ImageLocalityWeight     int
 }
 
 func TestNodeOrderPlugin(t *testing.T) {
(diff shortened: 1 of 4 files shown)


pkg/scheduler/util/predicate_helper_test.go
```diff
@@ -0,0 +1,278 @@
...
+				Job:       "job1",
+				TaskRole:  "worker",
+				Namespace: "ns",
+				Name:      "task",
+			},
+			nodes: []*api.NodeInfo{
+				{Name: "node2"},
+			},
+			nodesInShard:     sets.New[string]("node1"),
+			shardingMode:     commonutil.HardShardingMode,
```
qi-min, line 118: maybe a test case for softmode can be added to make sure nodes will not be exluded in soft mode,
and another test case for hardmode to verify nodes in shard will not be excluded

pkg/scheduler/plugins/nodeorder/nodeorder_test.go
```diff
@@ -139,3 +348,88 @@ func TestNodeOrderPlugin(t *testing.T) {
...
+			name:                  "all weights zero",
+			weight:                priorityWeight{},
+			expectNodeOrderKeys:   []string{},
+			expectNoNodeOrderKeys: []string{noderesources.Name + "_LeastAllocated", noderesources.Name + "_MostAllocated", noderesources.BalancedAllocationName, nodeaffinity.Name, imagelocality.Name},
+			expectScorePlugins: []string{
+				interpodaffinity.Name,
+				tainttoleration.Name,
+				podtopologyspread.Name,
+			},
+			expectNotInScorePlugins: []string{},
```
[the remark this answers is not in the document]
JesseStutler: You're right, solved

pkg/scheduler/plugins/nodeorder/nodeorder_test.go
```diff
@@ -101,6 +118,192 @@ func TestNodeOrderPlugin(t *testing.T) {
...
+		{
+			TestCommonStruct: uthelper.TestCommonStruct{
+				Name: "pod topology spread prefers node to reduce skew",
+				PodGroups: []*schedulingv1.PodGroup{
+					util.BuildPodGroup("pg1", "c1", "c1", 0, nil, schedulingv1.PodGroupInqueue),
+				},
+				Pods: []*v1.Pod{
+					// Existing pod on n1 (zone1)
+					util.BuildPod("c1", "p1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{"app": "foo"}, make(map[string]string)),
+					// Existing pod on n2 (zone1)
```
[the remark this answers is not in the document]
JesseStutler: You're right, solved

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -322,23 +340,43 @@ func (pp *NodeOrderPlugin) BatchNodeOrderFn(task *api.TaskInfo, nodeInfo []*api.
 	}
 	nodeScores := make(map[string]float64, len(nodes))
 
-	podAffinityScores, podErr := interPodAffinityScore(pp.ScorePlugins[interpodaffinity.Name].(*interpodaffinity.InterPodAffinity), state, task.Pod, nodeInfos, pp.weight.podAffinityWeight)
-	if podErr != nil {
-		return nil, podErr
+	var podAffinityScores map[string]float64
+	var err error
+	if pp.ScorePlugins[interpodaffinity.Name] != nil {
```
hajnalmt, line 345: These if branches doesn't look proper, you are dereferencing the `Name`, but you are not initializing the plugin if the weight is 0 so this will cause a panic if the weight is 0 right?

pkg/scheduler/plugins/nodeorder/nodeorder_test.go
```diff
@@ -101,6 +118,192 @@ func TestNodeOrderPlugin(t *testing.T) {
...
+				Queues: []*schedulingv1.Queue{
+					util.BuildQueue("c1", 1, nil),
+				},
+				ExpectBindsNum: 1,
+				ExpectBindMap: map[string]string{
+					"c1/p4": "n3", // n3 (zone3) is empty, n2 (zone2) has 1 pod and n1(zone1) has 2 pods. spread is better on n3.
+				},
+			},
+			PodTopologySpreadWeight: 10,
+		},
```
hajnalmt, line 306: Maybe adding a runtime test with zero weights and all the pods is needed here (the "all weights zero" test checks only the map keys, doesn't call the function directly). It should bring up my theory that this will panic out: 
```
{
    TestCommonStruct: uthelper.TestCommonStruct{
        Name: "zero weight for batch scoring plugins with constraint pods",
        PodGroups: []*schedulingv1.PodGroup{
            util.BuildPodGroup("pg1", "c1", "c1", 0, nil, schedulingv1.PodGroupInqueue),
        },
        Pods: []*v1.Pod{
            util.BuildPodWithAffinity("c1", "p1", "", v1.PodPending,
                api.BuildResourceList("1", "1G"), "pg1",
                make(map[string]string), make(map[string]string),
                &v1.Affinity{
                    PodAffinity: &v1.PodAffinity{
                        PreferredDuringSchedulingIgnoredDuringExecution: []v1.WeightedPodAffinityTerm{
                            {
                                Weight: 100,
                                PodAffinityTerm: v1.PodAffinityTerm{
                                    LabelSelector: &metav1.LabelSelector{
                                        MatchLabels: map[string]string{"app": "test"},
                                    },
                                    TopologyKey: "kubernetes.io/hostname",
                                },
                            },
                        },
                    },
                }),
            util.BuildPodWithTolerations("c1", "p2", "", v1.PodPending,
                api.BuildResourceList("1", "1G"), "pg1",
                make(map[string]string), make(map[string]string),
                []v1.Toleration{
                    {
                        Key:      "test-key",
                        Operator: v1.TolerationOpEqual,
                        Value:    "test-value",
                        Effect:   v1.TaintEffectNoSchedule,
                    },
                }),
            util.BuildPodWithTopologySpreadConstraints("c1", "p3", "", v1.PodPending,
                api.BuildResourceList("1", "1G"), "pg1",
                make(map[string]string), make(map[string]string),
                []v1.TopologySpreadConstraint{
                    {
                        MaxSkew:           1,
                        TopologyKey:       "zone",
                        WhenUnsatisfiable: v1.ScheduleAnyway,
                        LabelSelector: &metav1.LabelSelector{
                            MatchLabels: map[string]string{"app": "test"},
                        },
                    },
                }),
        },
        Nodes: []*v1.Node{
            util.BuildNode("n1", api.BuildResourceList("4", "8Gi",
                []api.ScalarResource{{Name: "pods", Value: "10"}}...),
                make(map[string]string)),
            util.BuildNode("n2", api.BuildResourceList("8", "16Gi",
                []api.ScalarResource{{Name: "pods", Value: "10"}}...),
                make(map[string]string)),
        },
        Queues: []*schedulingv1.Queue{
            util.BuildQueue("c1", 1, nil),
        },
        ExpectBindsNum: 3,
        // All pods should schedule successfully based on LeastRequestedWeight
        ExpectBindMap: map[string]string{
            "c1/p1": "n2",
            "c1/p2": "n2",
            "c1/p3": "n2",
        },
    },
    LeastRequestedWeight:    1,
    MostRequestedWeight:     0,
    PodAffinityWeight:       0,
    TaintTolerationWeight:   0,
    PodTopologySpreadWeight: 0,
},
```

pkg/scheduler/plugins/nodeorder/nodeorder.go
JesseStutler: You mean `pp.ScorePlugins` will be nil? But currently we will always the initialize ScorePlugins maps during `InitPlugins`, so we will not cause panic here. And if the weight is 0, so `pp.ScorePlugins[interpodaffinity.Name]` will return a nil, and the `if` protects the situation

hajnalmt: No, `interpodAffinity` is `nil` when weight is 0, so `interpodAffinity.Name ` will cause a panic in that case I think.

JesseStutler: The `interpodaffinity` here is a package name, instead of a variable name, so it will not cause panic I think 😄

hajnalmt: Oh you are right sorry,  the cases tricked me thank you!

JesseStutler:
Oh I'm still working on add @qi-min 's suggested tests but didn't notice it was already merged: https://github.com/volcano-sh/volcano/pull/4983#discussion_r2720378975 😳, I will submit a new PR to fix it

pkg/scheduler/util/predicate_helper_test.go
JesseStutler: This pr was already merged, I will raise a new one

Issue #4881, opened 2025-12-26
E2E for ShardingController


qi-min:
## Summary
ShardingController is a newly introduced component in volcano-controllers to support shard management and control. To ensure correctness and stability, we need to add end-to-end (E2E) test cases to validate shard control behavior and sharding accuracy under different sharding strategies.

The E2E tests should cover scenarios such as shard assignment, node resource boundary changes, controller reconciliation behavior under different strategies, and nodes distribution across shards. These tests will help verify that the ShardingController works as expected in real cluster environments and prevent regressions as new sharding strategies or features are introduced.

## Scenarios
1. Shards are created based on default configuration after controller startup if Shards do not exist
2. Nodes are assigned from one shard to another shard when resource usage of node significantly changed
3. Newly created Nodes are added to shard
4. Removed or unschedulable  nodes are moved out from shard
5. Configuration change trigger nodes re-assigned to different
6. Node re-assignment keep nodes as stable as possible in each shard

shiverse94:
can I share the approach here.

shiverse94:
Can you please verify whether i am heading toward right directions?
https://github.com/volcano-sh/volcano/pull/4966

SOHAMPAL23:
The ShardingController directly affects how nodes are grouped and managed across shards, so even small logic issues can lead to unstable scheduling behavior. Unit tests alone can’t fully capture this, because shard assignment depends on real controller reconciliation, node lifecycle events, and dynamic resource changes.

**My approach is to validate ShardingController end-to-end in a real Kubernetes environment, exactly the way it runs in production. Using a reproducible test cluster, the E2E tests will verify that shards are created correctly, nodes are assigned and re-assigned as expected when resources or configurations change, and that node movement remains stable and minimal during reconciliation.**

Specifically, the tests will:

- Verify that shards are automatically created on controller startup when none exist.

- Ensure nodes are assigned and re-assigned correctly based on resource usage and sharding strategy.

- Validate behavior when nodes are added, removed, or become unschedulable.

- Confirm that configuration changes trigger controlled rebalancing, without unnecessary node movement.

- Check that the controller maintains stability and minimal disruption during reconciliation.

**These tests will be fully automated, integrated into CI, and designed to be deterministic and reproducible, helping the project:**

- Catch regressions early

- Increase confidence in sharding logic

- Safely evolve sharding strategies over time

Issue #4877, opened 2025-12-25
ShardingController - Scheduler Sharding Configurations


ssfffss:
### What is the problem you're trying to solve

Problem: scheduler sharding configurations are hard coded, any changes need to change and redeploy the code; hard for debugging and production;

### Describe the solution you'd like

1. use config map to support online change of scheduler sharding configurations;
2. think about the real scenarios of configuration fields and format: should use use percentage rather than the number to specify the size of a specific scheduler shard? Should the configuration format be specified in another better format (e.g., json, yaml)?

### Documentation Updates

- [x] This feature requires design or user documentation changes.
- [x] If documentation changes are required, I will ensure the relevant documents are updated and published to the Volcano official website (https://volcano.sh) via the [volcano-sh/website](https://github.com/volcano-sh/website) repository.

JesseStutler:
And as we mentioned in our initial design doc: https://docs.google.com/document/d/1NDTEGulhQ__sdZ_c2kN3ojtJIZie5O7_GyZyKIOLz-g/edit?tab=t.3q3qlfh4r4sf#heading=h.9r4wotc3nl4d, we should follow plugin based sharding strategy, currently we only support CPU utilization sharding

agrawalcodes:
@JesseStutler picking this up!

PR #4990, opened 2026-01-25
e2e(ShardingController): add setup and base e2e tests for ShardingController

agrawalcodes:
## Description

This PR adds E2E test infrastructure and base tests for the ShardingController component. It introduces NodeShard helper utilities and wires the suite into the Kind-based CI environment with the sharding controller enabled via Helm.

This is the foundational step for the E2E test suite described in #4881. Additional test cases covering node reassignment on resource changes, new/removed nodes, configuration changes, and unschedulable node handling will be added as part of the LFX mentorship.

## Changes

- `test/e2e/util/shard.go`: NodeShard CRD helper functions (create, get, list, delete, update, wait utilities)
- `test/e2e/shardingcontroller/`: ShardingController Ginkgo E2E suite with 3 test cases
- `hack/run-e2e-kind.sh`: Add `SHARDINGCONTROLLER` Helm install case with `controller_controllers: "*"` to enable sharding controller
- `installer/helm/chart/volcano/`: Add `--controllers` flag configuration to controller-manager deployment
- `Makefile`: Add `e2e-test-shardingcontroller` target
- `.github/workflows/e2e_sharding_controller.yaml`: Add CI workflow

## Test Cases

1. **Shard creation** — NodeShards `volcano` and `agent-scheduler` are created by the controller on startup based on default configuration
2. **Node assignment** — Worker nodes are assigned to shards after controller startup
3. **Assignment stability** — Node assignments remain stable (≥80% unchanged) without cluster changes

## How to Run

```bash
make e2e-test-shardingcontroller
```
## Test Plan
- [x] `make e2e-test-shardingcontroller` passes locally
- [x] CI workflow runs successfully in the child PR #4992

## Related Issues
Ref: Part of #4881 

### DIFF
--- a/.github/workflows/e2e_shard.yaml
+++ b/.github/workflows/e2e_shard.yaml
@@ -4,10 +4,22 @@ on:
   workflow_call:
 
 jobs:
-  e2e_agent_scheduler:
+  e2e:
     runs-on: ubuntu-24.04
-    name: E2E Agent Scheduler
+    name: ${{ matrix.name }}
     timeout-minutes: 50
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: E2E Agent Scheduler
+            target: e2e-test-agentscheduler
+            logs_dir: e2e-agent-scheduler-logs
+            artifact: volcano_e2e_agent_scheduler_logs
+          - name: E2E Sharding Controller
+            target: e2e-test-shardingcontroller
+            logs_dir: e2e-sharding-controller-logs
+            artifact: volcano_e2e_sharding_controller_logs
     steps:
       - name: Free Disk Space
         uses: jlumbroso/free-disk-space@main
@@ -50,12 +62,12 @@ jobs:
 
       - name: Run E2E Tests
         run: |
-          export ARTIFACTS_PATH=${{ github.workspace }}/e2e-agent-scheduler-logs
-          make e2e-test-agentscheduler FORCE_REBUILD=false
+          export ARTIFACTS_PATH=${{ github.workspace }}/${{ matrix.logs_dir }}
+          make ${{ matrix.target }} FORCE_REBUILD=false
 
-      - name: Upload e2e agent scheduler logs
+      - name: Upload e2e logs
         if: '!success()'
         uses: actions/upload-artifact@v4
         with:
-          name: volcano_e2e_agent_scheduler_logs
-          path: ${{ github.workspace }}/e2e-agent-scheduler-logs
+          name: ${{ matrix.artifact }}
+          path: ${{ github.workspace }}/${{ matrix.logs_dir }}
--- a/hack/run-e2e-kind.sh
+++ b/hack/run-e2e-kind.sh
@@ -221,6 +221,29 @@ EOF
     shardSyncPeriod: "30s"
     enableNodeEventTrigger: true'
   ;;
+"SHARDINGCONTROLLER")
+  echo "Install volcano chart with crd version $crd_version and sharding controller enabled"
+  helm-install-volcano '  controller_log_level: 5
+  controller_enabled_controllers: "*"
+  sharding_configmap_data: |
+    schedulerConfigs:
+      - name: volcano
+        type: volcano
+        cpuUtilizationMin: 0.0
+        cpuUtilizationMax: 0.6
+        preferWarmupNodes: false
+        minNodes: 2
+        maxNodes: 100
+      - name: agent-scheduler
+        type: agent
+        cpuUtilizationMin: 0.7
+        cpuUtilizationMax: 1.0
+        preferWarmupNodes: true
+        minNodes: 2
+        maxNodes: 100
+    shardSyncPeriod: "60s"
+    enableNodeEventTrigger: true'
+  ;;
 *)
   echo "Install volcano chart with crd version $crd_version"
   helm-install-volcano
@@ -407,6 +430,10 @@ case ${E2E_TYPE} in
 "AGENTSCHEDULER")
     echo "Running agent scheduler e2e suite..."
     KUBECONFIG=${KUBECONFIG} GOOS=${OS} ginkgo -v -r --slow-spec-threshold='30s' --progress ./test/e2e/agentscheduler/
+    ;;
+"SHARDINGCONTROLLER")
+    echo "Running sharding controller e2e suite..."
+    KUBECONFIG=${KUBECONFIG} GOOS=${OS} ginkgo -v -r --slow-spec-threshold='30s' --progress ./test/e2e/shardingcontroller/
     ;;
 esac
 
--- a/test/e2e/shardingcontroller/sharding_controller_test.go
+++ b/test/e2e/shardingcontroller/sharding_controller_test.go
@@ -1,0 +1,726 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package shardingcontroller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"volcano.sh/volcano/pkg/controllers/sharding"
+	e2eutil "volcano.sh/volcano/test/e2e/util"
+)
+
+const (
+	// Sharding ConfigMap coordinates (must match helm release name and namespace)
+	shardingConfigMapName      = "integration-sharding-configmap"
+	shardingConfigMapNamespace = "volcano-system"
+
+	// Polling intervals
+	pollInterval = 500 * time.Millisecond
+
+	// Wait time for sharding controller to create NodeShards after startup
+	shardCreationTimeout = 3 * time.Minute
+
+	// Wait time for shard reassignment after workload changes
+	reassignmentTimeout = 3 * time.Minute
+)
+
+var _ = Describe("ShardingController E2E Test", func() {
+	var ctx *e2eutil.TestContext
+
+	BeforeEach(func() {
+		ctx = e2eutil.InitTestContext(e2eutil.Options{})
+	})
+
+	AfterEach(func() {
+		e2eutil.CleanupTestContext(ctx)
+	})
+
+	JustAfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			e2eutil.DumpTestContext(ctx)
+		}
+	})
+
+	// getShardingConfig reads the sharding configuration from the ConfigMap deployed by helm.
+	getShardingConfig := func() *sharding.ShardingConfig {
+		cfg, err := e2eutil.GetShardingConfigFromConfigMap(ctx, shardingConfigMapName, shardingConfigMapNamespace)
+		ExpectWithOffset(1, err).NotTo(HaveOccurred(), "failed to read sharding config from ConfigMap")
+		return cfg
+	}
+
+	// waitForNodeShardsCreated waits for all NodeShards defined in the ConfigMap to be created.
+	waitForNodeShardsCreated := func(cfg *sharding.ShardingConfig) {
+		By("Waiting for ShardingController to create NodeShards")
+		expectedNames := make(map[string]bool, len(cfg.SchedulerConfigs))
+		for _, sc := range cfg.SchedulerConfigs {
+			expectedNames[sc.Name] = false
+		}
+		err := wait.PollUntilContextTimeout(context.TODO(), pollInterval, shardCreationTimeout, true,
+			func(c context.Context) (bool, error) {
+				shards, err := e2eutil.ListNodeShards(ctx)
+				if err != nil {
+					GinkgoWriter.Printf("Error listing NodeShards: %v\n", err)
+					return false, nil
+				}
+				found := 0
+				for _, shard := range shards.Items {
+					if _, ok := expectedNames[shard.Name]; ok {
+						found++
+					}
+				}
+				if found == len(expectedNames) {
+					GinkgoWriter.Printf("All %d expected NodeShards found\n", found)
+					return true, nil
+				}
+				GinkgoWriter.Printf("Waiting for NodeShards... found=%d, expected=%d, total=%d\n",
+					found, len(expectedNames), len(shards.Items))
+				return false, nil
+			})
+		Expect(err).NotTo(HaveOccurred(), "NodeShards should be created for all configured schedulers")
+	}
+
+	Describe("Shard Creation", func() {
+		It("should create NodeShards matching the ConfigMap configuration", func() {
+			cfg := getShardingConfig()
+			waitForNodeShardsCreated(cfg)
+
+			By("Listing all NodeShards in the cluster")
+			shards, err := e2eutil.ListNodeShards(ctx)
+			Expect(err).NotTo(HaveOccurred(), "failed to list NodeShards")
+
+			By("Verifying NodeShard count matches ConfigMap")
+			Expect(len(shards.Items)).To(BeNumerically(">=", len(cfg.SchedulerConfigs)),
+				"at least as many NodeShards as configured schedulers should exist")
+
+			By("Verifying each configured scheduler has a corresponding NodeShard")
+			shardMap := make(map[string]bool)
+			for _, shard := range shards.Items {
+				shardMap[shard.Name] = true
+				GinkgoWriter.Printf("Found NodeShard: %s with %d nodes desired, %d nodes in use\n",
+					shard.Name, len(shard.Spec.NodesDesired), len(shard.Status.NodesInUse))
+			}
+
+			for _, sc := range cfg.SchedulerConfigs {
+				Expect(shardMap).To(HaveKey(sc.Name),
+					fmt.Sprintf("NodeShard for scheduler %q should exist", sc.Name))
+			}
+		})
+
+		It("should have worker nodes assigned to shards after controller startup", func() {
+			cfg := getShardingConfig()
+			waitForNodeShardsCreated(cfg)
+
+			By("Getting cluster worker nodes")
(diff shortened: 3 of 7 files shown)


agrawalcodes:
Hi @qi-min @JesseStutler,
This is the base and setup for e2e suite for volcano agent scheduling. PRs for Agent Scheduler and Sharding Scheduler extends this PR. Please take a look!

.github/workflows/e2e_shard.yaml
JesseStutler, line 1: We have updated all the workflows to reuse the same image, please follow this PR: https://github.com/volcano-sh/volcano/pull/5081 to imitate other workflows

.github/workflows/e2e_sharding_controller.yaml
```diff
@@ -0,0 +1,62 @@
+name: ShardingController E2E Tests
```
JesseStutler: I hope this workflow doesn't only contain sharding controller, but rather contains all the related e2e tests for the shard(like in agent scheduler, batch scheduler)

installer/helm/chart/volcano/values.yaml
```diff
@@ -40,6 +40,8 @@ custom:
   controller_worker_threads: 3
   controller_worker_threads_for_gc: 5
   controller_worker_threads_for_podgroup: 5
+  # Default: "*,-sharding-controller" (sharding-controller disabled by default)
+  controller_controllers: ~
```
JesseStutler: Better rename to `controller_enabled_controllers`

hack/run-e2e-kind.sh
```diff
@@ -191,6 +191,51 @@ custom:
   ignored_provisioners: ${IGNORED_PROVISIONERS:-""}
 EOF
   ;;
+"SHARDINGCONTROLLER")
```
JesseStutler, line 224: The only difference of this part is `controller_controllers` I think, could you extract the public part of these codes with other cases and reuse the public part?

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+			By("Verifying NodeShards exist for configured schedulers")
+			shardNames := make([]string, 0, len(shards.Items))
+			for _, shard := range shards.Items {
+				shardNames = append(shardNames, shard.Name)
+				GinkgoWriter.Printf("Found NodeShard: %s with %d nodes desired, %d nodes in use\n",
+					shard.Name, len(shard.Spec.NodesDesired), len(shard.Status.NodesInUse))
+			}
+
+			// Default config creates shards for "volcano" and "agent-scheduler"
```
JesseStutler: I think we should not only judge whether the names are correct, but also judge all the configuration in the node shard is correct, but currently, the sharding controller is using controller arguments flag to pass the parameter, which is not flexible.

Could assign this task and solve it first? https://github.com/volcano-sh/volcano/issues/4877
To add parameter(e.g. yaml or json style) in the vc-controller's configmap for sharding controller, and then in the e2e test, we can watch or get this configmap to get the parameter of sharding controller, and then we judge whether all the generated node shard configuration is correct

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+			func(c context.Context) (bool, error) {
+				shards, err := e2eutil.ListNodeShards(ctx)
+				if err != nil {
+					GinkgoWriter.Printf("Error listing NodeShards: %v\n", err)
+					return false, nil
+				}
+				// Check if both expected shards exist
+				foundVolcano := false
+				foundAgent := false
+				for _, shard := range shards.Items {
```
JesseStutler, line 86: As I mentioned in https://github.com/volcano-sh/volcano/pull/4990/changes#r2921959809, we should not hard-code here to wait for only two shard generated for volcano and agent scheduler, we should refactor the sharding controller parameter to add them configured through configmap, then to judge whether the number of generated node shard is align with the configuration

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+
+			// Default config creates shards for "volcano" and "agent-scheduler"
+			Expect(shardNames).To(ContainElement(VolcanoShardName),
+				"NodeShard for volcano scheduler should exist")
+			Expect(shardNames).To(ContainElement(AgentSchedulerShard),
+				"NodeShard for agent-scheduler should exist")
+		})
+
+		It("NodeShards should have worker nodes assigned after controller startup", func() {
```
JesseStutler: https://github.com/volcano-sh/volcano/blob/78b480781ac8635924441f53234e62ff4b634fbe/pkg/controllers/sharding/sharding_controller.go#L236-L272, as you can see from this part of codes, the sharding controller have configuration that can be passed (which currently is passed through startup parameters), and the https://github.com/volcano-sh/volcano/blob/78b480781ac8635924441f53234e62ff4b634fbe/pkg/controllers/sharding/config.go#L49-L50, the default CPU usage of the nodes in the nodeshard assigned to Volcano is 0-60%, and the CPU usage of the node shard assigned to the agent scheduler is 70-100%, so you should schedule some pods for the kind node, and then check that the node with the corresponding CPU usage can be assigned to the corresponding nodeshard, and then check whether the corresponding node can be in the nodeshard , including soft and hard modes.

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+			// At least some nodes should be assigned
+			Expect(totalAssigned).To(BeNumerically(">", 0),
+				"at least one node should be assigned to a shard")
+		})
+	})
+
+	// TODO: Add tests for unschedulable node handling once ShardingController behavior is confirmed
+
+	Describe("Node Stability", func() {
```
JesseStutler, line 462: Besides add cases for node unchanged, you should also add cases when the pods allocation changed, the n the cpu utilization on all the nodes may change, and then the nodes assigned for the nodeshard may also change, which you should verify

JesseStutler (commented):
@agrawalcodes Hi rejash, I hope you get through the design docs and understand why we will come up the idea with nodeshard, and then you may know what kind of e2e tests you should add, not by my mention

.github/workflows/e2e_shard.yaml
agrawalcodes: Thanks for the pointer! I'll update this workflow to follow PR #5081 and reuse the shared image.

.github/workflows/e2e_sharding_controller.yaml
agrawalcodes: Agreed. I'll rename/expand this workflow to cover all shard-related e2e tests

hack/run-e2e-kind.sh
agrawalcodes: Cool, makes sense

test/e2e/shardingcontroller/sharding_controller_test.go
agrawalcodes: Makes sense, once the sharding controller config is exposed via a configmap, we can read it in tests and assert the full NodeShard spec matches the configured values. I'll pick up #4877 first and then update this test. Please assign that to me!

JesseStutler:
After #4991 merged, you may need to resolve the conflicts, and I think we can unify shard related e2e tests into one workflow @agrawalcodes

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,726 @@
...
+		ExpectWithOffset(1, err).NotTo(HaveOccurred(), "failed to read sharding config from ConfigMap")
+		return cfg
+	}
+
+	// waitForNodeShardsCreated waits for all NodeShards defined in the ConfigMap to be created.
+	waitForNodeShardsCreated := func(cfg *sharding.ShardingConfig) {
+		By("Waiting for ShardingController to create NodeShards")
+		expectedNames := make(map[string]bool, len(cfg.SchedulerConfigs))
+		for _, sc := range cfg.SchedulerConfigs {
+			expectedNames[sc.Name] = false
```
JesseStutler: Although it is only a minor comment, I suggest that you can use `map[string]struct{} `when you use map to mark whether a value exists or not in the future, because the empty type `struct{}` does not take up space, which is a go trick, here bool is false looks weird

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,726 @@
...
+
+			By(fmt.Sprintf("Waiting for node %s to appear in high-util shard %q", targetNode.Name, highUtilScheduler.Name))
+			err = wait.PollUntilContextTimeout(context.TODO(), pollInterval, reassignmentTimeout, true,
+				func(c context.Context) (bool, error) {
+					shard, err := e2eutil.GetNodeShard(ctx, highUtilScheduler.Name)
+					if err != nil {
+						return false, nil
+					}
+					for _, n := range shard.Spec.NodesDesired {
+						if n == targetNode.Name {
```
JesseStutler: I think we should do more verification here. This node should not appear in the low utilization shard (or observe that the low shard nodeInUse originally has all the nodes, then the CPU request changes, this node is moved out of NodeInUse and appears in the high utilization shard and eventually in NodeInUse).

I think this can be combined with the case above. High utilization nodes appear in high shards, and low utilization nodes appear in low shards (we can use wait to wait for the final state).

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,726 @@
...
+		})
+
+		It("nodes with high CPU utilization should be assigned to the high-utilization shard", func() {
+			cfg := getShardingConfig()
+			waitForNodeShardsCreated(cfg)
+
+			// Find the scheduler configured for high CPU utilization
+			var highUtilScheduler *sharding.SchedulerConfigSpec
+			for i := range cfg.SchedulerConfigs {
+				if cfg.SchedulerConfigs[i].CPUUtilizationMax >= 1.0 && cfg.SchedulerConfigs[i].CPUUtilizationMin > 0 {
```
JesseStutler: The method for validating `highUtiliztionShard` seems rather inelegant. Is there a more elegant way to write it? Also, if this pull request is merged: https://github.com/volcano-sh/volcano/pull/5132, I think we still should refactor all the hard-coded written style

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,726 @@
...
+					return true, nil
+				})
+			Expect(err).NotTo(HaveOccurred(),
+				fmt.Sprintf("node %s with utilization %.2f should not be assigned to any shard (gap: %.2f-%.2f)",
+					targetNode.Name, targetUtil, chosen.lo, chosen.hi))
+		})
+	})
+
+	Describe("Node Reassignment on Workload Change", func() {
+		It("node should move between shards when CPU utilization changes", func() {
```
JesseStutler: https://github.com/volcano-sh/volcano/pull/4990/changes#r3193601773, same as this comment, looks like we can unify this cases together? First all the nodes are empty, so they will all in the low utilization shard, but after utilization changed(e.g., schedule some pods onto these nodes), then the nodes will be moved to hight utilization shard, and not show in the low utilization shard duplicately

test/e2e/shardingcontroller/sharding_controller_test.go
JesseStutler, line 1: We can also add these cases:
1. Create new nodes, this node can be successfully added to shard, and this node can successfully schedule pod( e.g, volcano batch scheduler can schedule to this pod, but agent scheduler may not in hard mode)
2. Delete the nodes, the gone node should be deleted also in the shard
3. The configuration of the sharding controller changed, the shard values should also change

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,726 @@
...
+			By("Verifying assignments remain stable")
+			shard2, err := e2eutil.GetNodeShard(ctx, schedulerName)
+			Expect(err).NotTo(HaveOccurred())
+
+			GinkgoWriter.Printf("Initial nodes: %v\n", initialNodes)
+			GinkgoWriter.Printf("Current nodes: %v\n", shard2.Spec.NodesDesired)
+
+			commonNodes := countCommonNodes(initialNodes, shard2.Spec.NodesDesired)
+			if len(initialNodes) > 0 {
+				stabilityRatio := float64(commonNodes) / float64(len(initialNodes))
```
JesseStutler: I think the wait time/sleep time should at least >= sharding controller sync time(Or we can modify the sync time smaller in the e2e configuration). And we should verify the nodes list in the beginning and after is the same, not by the length

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,726 @@
...
+				}
+			}()
+
+			// The warmup node needs to be in the CPU utilization range of the warmup scheduler.
+			// Create a pod to push its utilization into that range if needed.
+			cpuCapacity := targetNode.Status.Capacity.Cpu().MilliValue()
+			targetUtilization := (warmupScheduler.CPUUtilizationMin + warmupScheduler.CPUUtilizationMax) / 2
+			cpuRequestMillis := int64(float64(cpuCapacity) * targetUtilization)
+
+			if cpuRequestMillis > 0 {
```
JesseStutler: I think we can delete this use case for now, because it looks rather strange. Is it due to the warm label being applied then entering the shard, or is it due to changes in CPU utilization? Also, this might need to be refactored after https://github.com/volcano-sh/volcano/pull/5132 is merged

JesseStutler:
Please check my comments @agrawalcodes We have moved forward a lot! One more step then we can finally merge it

JesseStutler:
@agrawalcodes Hi, the unit test failed, please check, also your commits looks like too AI generated, could you clean your commits as well? We don't welcome such commits

JesseStutler:
If your code contains a lots of AI-generated codes, please ensure that all code has been reviewed by you and is error-free, and it's best to disclose this in the PR description.

agrawalcodes:
> @agrawalcodes Hi, the unit test failed, please check, also your commits looks like too AI generated, could you clean your commits as well? We don't welcome such commits

@JesseStutler makes sense. actually i was trying opencode, but now sure that it doesn't work well enough! Will clean up the commit and let you know!

test/e2e/shardingcontroller/sharding_controller_test.go
agrawalcodes: https://github.com/volcano-sh/volcano/pull/4990#discussion_r3200294331

test/e2e/shardingcontroller/sharding_controller_test.go
agrawalcodes: Good catch

test/e2e/shardingcontroller/sharding_controller_test.go
agrawalcodes: cool, taking a look

test/e2e/shardingcontroller/sharding_controller_test.go
agrawalcodes: Yes, we can merge this first and get back to this again later. Because otherwise we would be blocked for that PR to merge.

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,552 @@
...
+			Expect(len(filterWorkerNodes(nodes.Items))).To(BeNumerically(">=", 1),
+				"cluster should have at least 1 worker node")
+
+			totalAssigned := 0
+			for _, sc := range cfg.SchedulerConfigs {
+				shard, err := e2eutil.GetNodeShard(ctx, sc.Name)
+				Expect(err).NotTo(HaveOccurred(), "failed to get NodeShard %s", sc.Name)
+				totalAssigned += len(shard.Spec.NodesDesired)
+			}
+			Expect(totalAssigned).To(BeNumerically(">", 0),
```
JesseStutler: Why is this case so strange? Shouldn't all the worker nodes appear in one of the existing shards? This use case looks pretty disjointed

test/e2e/shardingcontroller/sharding_controller_test.go
```diff
@@ -0,0 +1,552 @@
...
+			By(fmt.Sprintf("Creating fake Node %s to simulate a new node joining the cluster", fakeName))
+			_, err := ctx.Kubeclient.CoreV1().Nodes().Create(context.TODO(), fakeNode, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred(), "failed to create fake Node")
+			defer func() {
+				_ = ctx.Kubeclient.CoreV1().Nodes().Delete(context.TODO(), fakeName, metav1.DeleteOptions{})
+			}()
+
+			// Set capacity/allocatable via /status subresource. Retry on conflict because
+			// the node controller may modify the Node concurrently after creation.
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
JesseStutler, line 364: Just a question, can we patch node status like this way directly? Is the node we created actually NotReady, but was still added to the NodeShard? Or is this fake node Ready?

test/e2e/shardingcontroller/sharding_controller_test.go
agrawalcodes: Makes sense, this assertion is too loose. In a fresh kind cluster every worker node has ~0% CPU utilization, so all of them should land in the low-util shard (none in the gap, none in high-util). I'll tighten the test to assert that every worker node from `filterWorkerNodes` ends up in exactly one configured shard, and fail if any worker is missing or duplicated

test/e2e/shardingcontroller/sharding_controller_test.go
agrawalcodes: Status subresource updates via `UpdateStatus` are the standard way to set Node status server-side without a kubelet, the resulting object is a perfectly valid API record. That said, you're right that the fake Node is likely flipping to `NotReady` shortly after creation, since there is no kubelet posting heartbeats, the `node-lifecycle-controller` in `kube-controller-manager` will mark it `NotReady` after `--node-monitor-grace-period`.

The reason I believe the test still passes is that `ShardingController.filterEligibleNodes` (`pkg/controllers/sharding/sharding_manager.go:166-208`) filters only by CPU utilization range, it does not check `NodeReady`. So the fake node, with 0 CPU utilization (no pods scheduled, capacity 4 CPU), satisfies the low-util range `[0.0, 0.6]` and is added to the shard regardless of Ready status.

Two options:
1. Leave as-is and document the assumption (controller is Ready-agnostic, this test verifies the lifecycle path of the controller's node informer reacting to Add/Delete events).
2. If we think `NotReady` nodes should be excluded from shards, that's a controller behavior change (separate PR), and this test would need to keep the fake Node's `Ready` condition fresh until the assertion (e.g., periodic `UpdateStatus`).

What's your preference here?

JesseStutler: I think currently it's fine, the scheduler will filter out those NotReady nodes. For e2e tests, I think we only need to judge that the new created nodes will also join the shard, thanks for your reply

JesseStutler:
/approve
/lgtm
Thanks for your great work!

Issue #4882, opened 2025-12-26
E2E for Agent Scheduler


qi-min:
## Summary
A new Agent Scheduler has been introduced to support agent pod scheduling. To ensure correctness and stability, we need to add end-to-end (E2E) test cases to cover shard-based scheduling for Agent Pods.

The E2E tests should validate that Agent Pods are scheduled according to the assigned shard, including correct node selection, accurate pod placement, and pod scheduling with multiple workers. This will help ensure the Agent Scheduler works correctly in real cluster environments and prevent regressions as the scheduling logic evolves.

## Scenarios

- [ ] No Sharding

Verify that when no sharding is configured, Agent Pods are scheduled to the nodes within the cluster according to the default scheduling and scoring policies.

- [ ] Hard Sharding Mode

Verify that when hard sharding is enabled, Agent Pods are scheduled strictly to nodes within the assigned shard, and no nodes outside the shard are selected.

- [ ] Soft Sharding Mode – Preferred In-Shard Scheduling

Verify that when soft sharding is enabled, Agent Pods are preferentially scheduled to nodes within the assigned shard when sufficient resources are available.

- [ ] Soft Sharding Mode – Fallback to Out-of-Shard Nodes

Verify that when soft sharding is enabled and in-shard nodes do not have sufficient resources, Agent Pods can be scheduled to nodes outside the shard as a fallback.

- [ ] Shard Reconfiguration

Verify that after shard configuration changes (e.g., shard resizing or reassignment), Agent Pods can be scheduled to different nodes according to the updated shard layout.

- [ ] Multi-workers

Verify that when multi-workers is configured, Multiple agent Pods can be scheduled to different nodes  within the cluster according to the default scheduling and scoring policies.
Verify that when multi-workers is configured and one node available in cluster, Multiple agent Pods can be scheduled to same node.

mahil-2040:
Hii @qi-min 

I am Mahil Patel

I am interested in contributing in this Project for LFX term 1, are there any pre-requisites and guidelines to follow for working on this.
Also can you suggest any open issues so that I can start working on them and get familiar with the codebase?

PR #4991, opened 2026-01-25
e2e(AgentScheduler): add base e2e tests for Agent Scheduler

agrawalcodes:
Add end-to-end tests for the Agent Scheduler, covering deployment readiness, pod scheduling, coexistence with the default Volcano scheduler, and NodeShard verification.

### CI & Build
- New `e2e_shard.yaml` reusable workflow triggered from the main e2e pipeline
- `e2e-test-agentscheduler` Makefile target
- Conditionally load `vc-agent-scheduler` image into kind cluster for `AGENTSCHEDULER` runs

### Helm
- Add `controller_enabled_controllers` value and `--controllers` flag template to `controllers.yaml`, allowing e2e to enable all controllers including the sharding controller
- `AGENTSCHEDULER` install case in `run-e2e-kind.sh` with agent-scheduler, sharding controller, and sharding ConfigMap data (scheduler partition configs)

### Refactored `helm-install-volcano` helper
- Extracted common helm values into a reusable function used by the default and `AGENTSCHEDULER` cases
- Extra custom values are written to a temporary file and passed as a second `--values` flag instead of being interpolated inside a heredoc, which fixes YAML literal block scalars (`|`) like `sharding_configmap_data` being silently dropped during bash variable expansion

### E2E Tests (6 test cases)
1. Agent-scheduler deployment becomes ready
2. Single pod scheduled by agent-scheduler
3. Multiple pods scheduled concurrently
4. Coexistence with default Volcano scheduler pods
5. Batch of 10 pods scheduled with configured worker count
6. NodeShards exist when sharding controller is enabled

### Test Utilities
- `WaitPodScheduled` helper in `pod.go`
- Full NodeShard CRUD helpers in `shard.go` (create, get, list, delete, update, wait helpers)

## How to Run
`make e2e-test-agentscheduler`

## Test Plan
- [x] `make e2e-test-agentscheduler` passes locally
- [x] CI workflow runs successfully

## Related Issues
Ref: Part of #4882

### DIFF
--- a/test/e2e/agentscheduler/agent_scheduler_test.go
+++ b/test/e2e/agentscheduler/agent_scheduler_test.go
@@ -1,0 +1,370 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package agentscheduler
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	e2eutil "volcano.sh/volcano/test/e2e/util"
+)
+
+const (
+	AgentSchedulerName = "agent-scheduler"
+
+	pollInterval      = 500 * time.Millisecond
+	deploymentTimeout = 5 * time.Minute
+	volcanoSystemNS   = "volcano-system"
+)
+
+var _ = Describe("Agent Scheduler E2E Test", func() {
+	var ctx *e2eutil.TestContext
+
+	BeforeEach(func() {
+		ctx = e2eutil.InitTestContext(e2eutil.Options{})
+	})
+
+	AfterEach(func() {
+		e2eutil.CleanupTestContext(ctx)
+	})
+
+	JustAfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			e2eutil.DumpTestContext(ctx)
+		}
+	})
+
+	Describe("Agent Scheduler Deployment", func() {
+		It("agent-scheduler deployment should be ready", func() {
+			By("Checking agent-scheduler deployment in volcano-system namespace")
+			var deploy *appsv1.Deployment
+			err := wait.PollUntilContextTimeout(context.TODO(), pollInterval, deploymentTimeout, true,
+				func(c context.Context) (bool, error) {
+					deployments, err := ctx.Kubeclient.AppsV1().Deployments(volcanoSystemNS).List(
+						c, metav1.ListOptions{LabelSelector: "app=agent-scheduler"})
+					if err != nil {
+						GinkgoWriter.Printf("Error listing deployments: %v\n", err)
+						return false, nil
+					}
+					if len(deployments.Items) == 0 {
+						GinkgoWriter.Printf("No agent-scheduler deployment found yet\n")
+						return false, nil
+					}
+					deploy = &deployments.Items[0]
+					if deploy.Status.AvailableReplicas >= 1 {
+						return true, nil
+					}
+					GinkgoWriter.Printf("Waiting for agent-scheduler deployment: available=%d, desired=%d\n",
+						deploy.Status.AvailableReplicas, *deploy.Spec.Replicas)
+					return false, nil
+				})
+			Expect(err).NotTo(HaveOccurred(), "agent-scheduler deployment should become ready")
+			Expect(deploy.Status.AvailableReplicas).To(BeNumerically(">=", 1))
+			GinkgoWriter.Printf("Agent-scheduler deployment is ready with %d available replicas\n",
+				deploy.Status.AvailableReplicas)
+		})
+	})
+
+	Describe("Pod Scheduling", func() {
+		It("should schedule a single pod with agent-scheduler", func() {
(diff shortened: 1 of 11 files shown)


agrawalcodes:
Hi @qi-min @JesseStutler,
This PR for e2e base suite for Agent Scheduler extends #4990. Please take a look!

test/e2e/agentscheduler/agent_scheduler_test.go
```diff
@@ -0,0 +1,309 @@
...
+					shard.Name, len(shard.Spec.NodesDesired), len(shard.Status.NodesInUse))
+			}
+		})
+	})
+})
+
+// Helper functions
+
+func createAgentPod(namespace, name, cpuRequest string) *corev1.Pod {
+	// AutomountServiceAccountToken is disabled to avoid triggering a known
```
JesseStutler: Is it a different bug with this issue? https://github.com/volcano-sh/volcano/issues/5147, I'm not sure whether my PR can also fix your bug reported https://github.com/volcano-sh/volcano/pull/5154, so what's the root cause? I don't prefer to explain this bug and comment it ( because we have not already solved it)

test/e2e/agentscheduler/agent_scheduler_test.go
```diff
@@ -0,0 +1,309 @@
...
+				Expect(pod.Spec.NodeName).NotTo(BeEmpty(),
+					"pod %s should be bound to a node", podName)
+				GinkgoWriter.Printf("Pod %s scheduled on node %s\n", podName, pod.Spec.NodeName)
+			}
+		})
+
+		It("should not interfere with default volcano scheduler pods", func() {
+			By("Creating a pod with default volcano scheduler")
+			automount := false
+			volcanoPod := &corev1.Pod{
```
JesseStutler: I think we have already have some public funcs to create such normal pods, could you reuse those funcs?

test/e2e/agentscheduler/agent_scheduler_test.go
```diff
@@ -0,0 +1,309 @@
...
+			Expect(volcanoResult.Spec.SchedulerName).To(Equal(e2eutil.SchedulerName))
+			Expect(volcanoResult.Spec.NodeName).NotTo(BeEmpty())
+
+			GinkgoWriter.Printf("Agent pod on node: %s, Volcano pod on node: %s\n",
+				agentResult.Spec.NodeName, volcanoResult.Spec.NodeName)
+		})
+	})
+
+	// TODO: Add tests for scheduler-worker-count > 1 (requires helm reinstall with different config)
+	Describe("Worker Count Configuration", func() {
```
JesseStutler: Currently, it seems the `scheduler-worker-count` parameter isn't configurable via Helm, right? I think it's fine to just run multiple workers by default. Or alternatively, can a GitHub workflow run on two different machines, with different `agent-scheduler` configurations installed on each machine? Or is it possible to reconfigure the agent scheduler and restart it after other agent/shard E2E tests have finished running, to test multi-worker scenarios?

agrawalcodes: scheduler-worker-count is already configurable via Helm (custom.agent_scheduler_worker_count, defaults to 1). We could later add a follow up test multi-worker by passing a different value in the AGENTSCHEDULER helm install.

test/e2e/agentscheduler/agent_scheduler_test.go
agrawalcodes: This is a different bug from #5147. The stack traces look similar because both crash at predicates.go, but they hit different plugin filters with different nil pointers:
- #5147: VolumeBinding.FindPodVolumes -> binder.go:345 - nil podVolumeClaims due to multi-worker concurrency overwriting CycleState. Fixed by #5154 (already merged).
- Our bug: CSILimits.Filter -> csi.go:266 - nil nodeInfo.Node() for ALL nodes in the agent-scheduler's snapshot. This happens even with a single worker. The agent-scheduler's snapshot has NodeInfo objects where Node() returns nil despite addOrUpdateNode() correctly calling SetNode(). When a pod has volumes (e.g. the default SA token volume), CSILimits.Filter dereferences this nil node, causing the panic.
       
#5154 does not fix our bug since it's a snapshot issue, not a concurrency issue.
Removed the verbose comment per your feedback. Kept AutomountServiceAccountToken: false as a workaround since without it the agent-scheduler panics, but removed the explanation from the code.

JesseStutler: Yes I know #5154 is a concurrency issue but it also fix #5147 because currently cycleState will not be overwritten now, so I'm wondering that if you're still meeting it even after #5154 merged? We should find the root cause not just fix by a workaround. If you still can stably reproduce it even in current latest codes, please report an issue and then we can fix it

test/e2e/agentscheduler/agent_scheduler_test.go
JesseStutler: Why not just multi-worker in this PR? I think multi-worker scheduling cases may be not much differences with only one worker, right?

agrawalcodes: Makes sense

test/e2e/agentscheduler/agent_scheduler_test.go
agrawalcodes: It doesn't set SchedulerName, so pods default to default-scheduler instead of volcano. There's no existing helper that creates a standalone pod with schedulerName: volcano (other tests use VCJobs which set it internally)

test/e2e/agentscheduler/agent_scheduler_test.go
agrawalcodes: I rebased on top of #5154 fix and it started working with no panics

test/e2e/agentscheduler/agent_scheduler_test.go
JesseStutler, line 1: All the tests are happy path(the pods can be successfully scheduled by agent-scheduler, we can also build pods that like agent-scheduler will judge they cannot be scheduled and the pods will be still in pending status(PodScheduled=False, Reason=Unschedulable)

JesseStutler:
Also could you clean your commits like squash to 1-4? Currently there are too much commits @agrawalcodes

test/e2e/agentscheduler/agent_scheduler_test.go
agrawalcodes: Makes sense. added the negative paths also

JesseStutler:
/approve
/lgtm
Thanks for your great contribution!  @agrawalcodes Currently there are only base e2e test cases, look forward to see more cases based on node shard scheduling when cooperate with sharding controller :D

agrawalcodes:
> /approve /lgtm Thanks for your great contribution! @agrawalcodes Currently there are only base e2e test cases, look forward to see more cases based on node shard scheduling when cooperate with sharding controller :D

Thanks, will add more cases

Issue #4883, opened 2025-12-26
E2E for Volcano scheduler in different sharding mode


qi-min:
## Summary
The existing Volcano scheduler has been extended to support shard-based pod scheduling. To validate this enhancement and prevent regressions, we need to expand the current E2E test suite to cover pod scheduling behavior under sharding configurations.

## Scenarios

- [ ] No Sharding

Verify that when no sharding is configured, Pods are scheduled to the nodes within the cluster according to the default scheduling and scoring policies.

- [ ] Hard Sharding Mode

Verify that when hard sharding is enabled, Pods are scheduled strictly to nodes within the assigned shard, and no nodes outside the shard are selected. e2e need cover allocate, preempt, reclaim, backfill actions 

- [ ] Soft Sharding Mode – Preferred In-Shard Scheduling

Verify that when soft sharding is enabled, Pods are preferentially scheduled to nodes within the assigned shard when sufficient resources are available. e2e  need cover allocate, preempt, reclaim, backfill actions 

- [ ] Soft Sharding Mode – Fallback to Out-of-Shard Nodes

Verify that when soft sharding is enabled and in-shard nodes do not have sufficient resources, Pods can be scheduled to nodes outside the shard as a fallback. e2e  need cover allocate, preempt, reclaim, backfill actions 

- [ ] Shard reassignment

Verify that after shard changes (e.g., add/remove node in nodeshard), Pods can be scheduled to different nodes according to the updated shard layout. e2e  need cover allocate, preempt, reclaim, backfill actions

PR #4992, opened 2026-01-25
e2e(SchedulerSharding): add e2e tests for Volcano Scheduler Sharding

agrawalcodes:
## Summary
Add E2E test infrastructure and base tests for Volcano Scheduler with sharding support.

**Base setup from parent PR: #4990**

This is part of the LFX mentorship project. Additional test cases covering allocate, preempt, reclaim, backfill actions under different sharding modes will be added as part of the mentorship.

## Changes
- **test/e2e/schedulersharding/**: Add Volcano Scheduler Sharding E2E test suite (4 test cases)
- **.github/workflows/e2e_scheduler_sharding.yaml**: Add CI workflow

## Test Cases
- Volcano NodeShard should exist with nodes assigned
- Both NodeShards should be accessible
- Volcano jobs should be scheduled to cluster nodes
- Shard state should remain consistent during job execution

## How to Run
`make e2e-test-schedulersharding`

## Test Plan
- [x] `make e2e-test-schedulershardingr` passes locally
- [x] CI workflow runs successfully

## Related Issues
Ref: Part of #4883

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,13 +236,16 @@ e2e-test-admission-policy: images
 	E2E_TYPE=ADMISSION_POLICY ./hack/run-e2e-kind.sh
 
 e2e-test-agentscheduler: images
-	E2E_TYPE=AGENTSCHEDULER ./hack/run-e2e-kind.sh
+	E2E_TYPE=AGENTSCHEDULER$(if $(SHARDING_MODE),_$(shell echo $(SHARDING_MODE) | tr '[:lower:]' '[:upper:]'),) ./hack/run-e2e-kind.sh
 
 e2e-test-shardingcontroller: images
 	E2E_TYPE=SHARDINGCONTROLLER ./hack/run-e2e-kind.sh
 
 e2e-test-gangevict: images
 	E2E_TYPE=GANGEVICT ./hack/run-e2e-kind.sh
+
+e2e-test-schedulersharding: images
+	E2E_TYPE=SCHEDULERSHARDING$(if $(SHARDING_MODE),_$(shell echo $(SHARDING_MODE) | tr '[:lower:]' '[:upper:]'),) ./hack/run-e2e-kind.sh
 
 generate-yaml: init manifests
 	./hack/generate-yaml.sh CRD_VERSION=${CRD_VERSION}
--- a/hack/run-e2e-kind.sh
+++ b/hack/run-e2e-kind.sh
@@ -194,11 +194,17 @@ custom:
   ignored_provisioners: ${IGNORED_PROVISIONERS:-""}
 EOF
   ;;
-"AGENTSCHEDULER")
-  echo "Install volcano chart with crd version $crd_version, sharding controller and agent scheduler enabled"
-  helm-install-volcano '  controller_log_level: 5
-  controller_enabled_controllers: "*"
+"AGENTSCHEDULER"|"AGENTSCHEDULER_NONE"|"AGENTSCHEDULER_SOFT"|"AGENTSCHEDULER_HARD")
+  agent_scheduler_sharding_mode="${E2E_TYPE#AGENTSCHEDULER_}"
+  if [[ "${agent_scheduler_sharding_mode}" == "AGENTSCHEDULER" ]]; then
+    agent_scheduler_sharding_mode="NONE"
+  fi
+  agent_scheduler_sharding_mode=$(echo "${agent_scheduler_sharding_mode}" | tr '[:upper:]' '[:lower:]')
+  echo "Install volcano chart with crd version $crd_version, sharding controller and agent scheduler ${agent_scheduler_sharding_mode} mode enabled"
+  helm-install-volcano "  controller_log_level: 5
+  controller_enabled_controllers: \"*\"
   agent_scheduler_enable: true
+  agent_scheduler_sharding_mode: ${agent_scheduler_sharding_mode}
   agent_scheduler_worker_count: 2
   agent_scheduler_tolerations:
     - key: "node-role.kubernetes.io/control-plane"
@@ -209,20 +215,27 @@ EOF
       effect: "NoSchedule"
   sharding_configmap_data: |
     schedulerConfigs:
-      - name: volcano
-        type: volcano
-        cpuUtilizationMin: 0.0
-        cpuUtilizationMax: 0.6
-        minNodes: 1
-        maxNodes: 100
       - name: agent-scheduler
         type: agent
-        cpuUtilizationMin: 0.7
-        cpuUtilizationMax: 1.0
-        minNodes: 1
-        maxNodes: 100
+        policies:
+          - name: allocation-rate
+            weight: 1
+            arguments:
+              minCPUUtil: 0.0
+              maxCPUUtil: 0.6
+          - name: node-limit
+            arguments:
+              minNodes: 1
+              maxNodes: 1
+      - name: volcano
+        type: volcano
+        policies:
+          - name: node-limit
+            arguments:
+              minNodes: 1
+              maxNodes: 100
     shardSyncPeriod: "30s"
-    enableNodeEventTrigger: true'
+    enableNodeEventTrigger: true"
   ;;
 "SHARDINGCONTROLLER")
   echo "Install volcano chart with crd version $crd_version and sharding controller enabled"
@@ -232,20 +245,66 @@ EOF
     schedulerConfigs:
       - name: volcano
         type: volcano
-        cpuUtilizationMin: 0.0
-        cpuUtilizationMax: 0.6
-        preferWarmupNodes: false
-        minNodes: 2
-        maxNodes: 100
+        policies:
+          - name: allocation-rate
+            weight: 1
+            arguments:
+              minCPUUtil: 0.0
+              maxCPUUtil: 0.6
+          - name: node-limit
+            arguments:
+              minNodes: 2
+              maxNodes: 100
       - name: agent-scheduler
         type: agent
-        cpuUtilizationMin: 0.7
-        cpuUtilizationMax: 1.0
-        preferWarmupNodes: true
-        minNodes: 2
-        maxNodes: 100
+        policies:
+          - name: allocation-rate
+            weight: 1
+            arguments:
+              minCPUUtil: 0.7
+              maxCPUUtil: 1.0
+          - name: warmup
+            weight: 1
+          - name: node-limit
+            arguments:
+              minNodes: 2
+              maxNodes: 100
     shardSyncPeriod: "60s"
     enableNodeEventTrigger: true'
+  ;;
+"SCHEDULERSHARDING"|"SCHEDULERSHARDING_NONE"|"SCHEDULERSHARDING_SOFT"|"SCHEDULERSHARDING_HARD")
+  scheduler_sharding_mode="${E2E_TYPE#SCHEDULERSHARDING_}"
+  if [[ "${scheduler_sharding_mode}" == "SCHEDULERSHARDING" ]]; then
+    scheduler_sharding_mode="HARD"
+  fi
+  scheduler_sharding_mode=$(echo "${scheduler_sharding_mode}" | tr '[:upper:]' '[:lower:]')
+  echo "Install volcano chart with crd version $crd_version and scheduler sharding ${scheduler_sharding_mode} mode enabled"
+  helm-install-volcano "  controller_log_level: 5
+  controller_enabled_controllers: \"*\"
+  scheduler_sharding_mode: ${scheduler_sharding_mode}
+  sharding_configmap_data: |
+    schedulerConfigs:
+      - name: volcano
+        type: volcano
+        policies:
+          - name: allocation-rate
+            weight: 1
+            arguments:
+              minCPUUtil: 0.0
+              maxCPUUtil: 0.6
+          - name: node-limit
+            arguments:
+              minNodes: 1
+              maxNodes: 1
+      - name: agent-scheduler
+        type: agent
+        policies:
+          - name: node-limit
+            arguments:
+              minNodes: 1
+              maxNodes: 100
+    shardSyncPeriod: "30s"
+    enableNodeEventTrigger: true"
   ;;
 *)
   echo "Install volcano chart with crd version $crd_version"
@@ -312,7 +371,9 @@ custom:
   enabled_admissions: "/pods/mutate,/queues/mutate,/podgroups/mutate,/jobs/mutate,/jobs/validate,/jobflows/validate,/pods/validate,/queues/validate,/podgroups/validate,/hypernodes/validate,/cronjobs/validate"
   ignored_provisioners: ${IGNORED_PROVISIONERS:-""}
 EOF
+  local helm_status=${PIPESTATUS[1]}
   [[ -n "${extra_values_flag}" ]] && rm -f "${tmpfile}"
+  return "${helm_status}"
 }
 
 function uninstall-volcano {
@@ -443,9 +504,14 @@ case ${E2E_TYPE} in
     echo "Running cronjob e2e suite..."
     KUBECONFIG=${KUBECONFIG} GOOS=${OS} ginkgo -v -r --slow-spec-threshold='30s' --progress ./test/e2e/cronjob/
     ;;
-"AGENTSCHEDULER")
-    echo "Running agent scheduler e2e suite..."
-    KUBECONFIG=${KUBECONFIG} GOOS=${OS} ginkgo -v -r --slow-spec-threshold='30s' --progress ./test/e2e/agentscheduler/
+"AGENTSCHEDULER"|"AGENTSCHEDULER_NONE"|"AGENTSCHEDULER_SOFT"|"AGENTSCHEDULER_HARD")
+    agent_scheduler_sharding_mode="${E2E_TYPE#AGENTSCHEDULER_}"
+    if [[ "${agent_scheduler_sharding_mode}" == "AGENTSCHEDULER" ]]; then
+      agent_scheduler_sharding_mode="NONE"
+    fi
+    agent_scheduler_sharding_mode=$(echo "${agent_scheduler_sharding_mode}" | tr '[:upper:]' '[:lower:]')
+    echo "Running agent scheduler ${agent_scheduler_sharding_mode} e2e suite..."
+    KUBECONFIG=${KUBECONFIG} GOOS=${OS} ginkgo --label-filter="${agent_scheduler_sharding_mode}" -v -r --slow-spec-threshold='30s' --progress ./test/e2e/agentscheduler/
     ;;
 "SHARDINGCONTROLLER")
     echo "Running sharding controller e2e suite..."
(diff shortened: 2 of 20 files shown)


agrawalcodes:
@qi-min @JesseStutler, This one of e2e base suite for Sharding Scheduler extends the same https://github.com/volcano-sh/volcano/pull/4990.

JesseStutler:
It's a great contribution, thanks, will take a look after

agrawalcodes:
> It's a great contribution, thanks, will take a look after

Cool @JesseStutler. I'll make changes to the base and approach as per your review. This should give us a good foundational base for e2e setup of Agent Scheduling. The gemini review seems to be very generic, I'll resolve them after your review.

The CI ran successfully in GitHub Actions!

JesseStutler:
Need to update this PR after https://github.com/volcano-sh/volcano/pull/4990 get merged to resolve the conflicts @agrawalcodes

JesseStutler:
Will take a look in these two days, thanks, after https://github.com/volcano-sh/volcano/pull/5132 merged we also need to do some refactoring for the existing shard e2e tests

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
JesseStutler, line 1: The volcano scheduler sharding tests should also include none/soft mode allocation tests, let me explain what do different modes mean:
- **none**: default for volcano scheduler, the behavior is the same as if we don't use node shard, volcano scheduler can still schedule pods onto all nodes in the cluster
- **hard**: volcano scheduler will not schedule pods onto nodes that are not in the shard, the typical codes for hard mode: https://github.com/volcano-sh/volcano/blob/6c284399aa37c82640ffb55ff3026670469c3e97/pkg/scheduler/util/predicate_helper.go#L102-L111, as you can see from here, if the checking node are not in the shard, scheduler will skip this node
- **soft**: volcano scheduler will prirotize to schedule the node in the shard, but if the nodes in the shard are full, then we will schedule onto nodes that're not in the shard: https://github.com/volcano-sh/volcano/blob/6c284399aa37c82640ffb55ff3026670469c3e97/pkg/scheduler/actions/allocate/allocate.go#L723-L728

You should construct cases for different modes(nothing special for none mode, but could include some because we don't want to influence the behavior when we don't want nodeshard). It even notice me that @agrawalcodes you should also include soft/none mode for agent scheduler e2e tests, and there are also maybe tests are not so suitable. 

We could construct cases structure like:
```
Decribe( Hard Mode)
  It( xxxxx)

Describe(Soft Mode)
  It(xxxxx)

Describe(None Mode)
  It(xxxxx)
````

Like this

File: test/e2e/schedulersharding/volcano_scheduler_sharding_test.go [hunk 1]
```diff
@@ -0,0 +1,320 @@
...
+		It("job requesting more CPU than any single shard node should stay pending", func() {
+			waitForNodeShardsCreated()
+
+			By("Creating a job that requests excessive CPU per task")
+			job := e2eutil.CreateJob(ctx, &e2eutil.JobSpec{
+				Name: "insufficient-resource-job",
+				Tasks: []e2eutil.TaskSpec{
+					{
+						Img: e2eutil.DefaultBusyBoxImage,
+						Req: e2eutil.ThirtyCPU,
```
JesseStutler: I don't think this test is suitable for sharding e2e tests, even there is no node shard, such large pod will be also pending

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,320 @@
...
+					shard.Name, len(shard.Spec.NodesDesired))
+			}
+
+			Expect(shardNames).To(ContainElement(VolcanoShardName))
+			Expect(shardNames).To(ContainElement(AgentSchedulerShardName))
+		})
+	})
+
+	Describe("Basic Volcano Job Scheduling with Sharding", func() {
+		It("Volcano jobs should be scheduled only to volcano shard nodes", func() {
```
JesseStutler, line 160: This case is not correct, may be is that because there are no nodes in the agent scheduler's shard, you should construct cases that agent scheduler's shard also contains some nodes, then the pods can only scheduled to the nodes in volcano scheduler's shard( hard mode )

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,320 @@
...
+				volcanoNodeSet[n] = struct{}{}
+			}
+			agentOnlyNodes := make(map[string]struct{})
+			for _, n := range agentShard.Spec.NodesDesired {
+				if _, ok := volcanoNodeSet[n]; !ok {
+					agentOnlyNodes[n] = struct{}{}
+				}
+			}
+			GinkgoWriter.Printf("Volcano shard nodes: %v\n", volcanoShard.Spec.NodesDesired)
+			GinkgoWriter.Printf("Agent-only shard nodes: (excluded from verification)\n")
```
JesseStutler: As you can see from the e2e's log: 

Image, there are even no nodes in the agent scheduler's shard, so this test can't verify that the scheduler won't schedule nodes outside the shard, you should construct cases that agent scheduler's shard also contains some nodes

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,320 @@
...
+
+			By("Verifying shard state after job execution")
+			finalShard, err := e2eutil.GetNodeShard(ctx, VolcanoShardName)
+			Expect(err).NotTo(HaveOccurred())
+
+			GinkgoWriter.Printf("Initial nodes: %v\n", initialNodes)
+			GinkgoWriter.Printf("Final nodes: %v\n", finalShard.Spec.NodesDesired)
+
+			// Shard should remain stable during short job execution
+			Expect(len(finalShard.Spec.NodesDesired)).To(BeNumerically(">=", 1),
```
JesseStutler: You didn't verify the initialNodes and fianlNodes are the same

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,320 @@
...
+					Expect(agentOnlyNodes).NotTo(HaveKey(pod.Spec.NodeName),
+						fmt.Sprintf("pod %s was scheduled to agent-only node %q, violating shard isolation",
+							pod.Name, pod.Spec.NodeName))
+				}
+			}
+			GinkgoWriter.Printf("All pods correctly isolated to volcano shard nodes\n")
+		})
+	})
+
+	It("second job should stay pending when shard resources are exhausted by first job", func() {
```
JesseStutler: I think this case is still hard mode related tests and maybe dupliate with "should not schedule pods onto nodes outside the volcano shard"

JesseStutler:
In `test/e2e/schedulersharding/volcano_scheduler_sharding_test.go`, the shard-state coverage around [`Describe("Shard State Consistency", ...)`](https://github.com/volcano-sh/volcano/blob/ec82ae319d63a66e46fceeea12f9474ec920c5c3/test/e2e/schedulersharding/volcano_scheduler_sharding_test.go#L284-L318) should cover dynamic `NodeShard` updates and status coordination. These are the main scheduler-side shard behaviors that are not exercised by the current cases.

Recommended scenarios:

1. Update `NodeShard/volcano.spec.nodesDesired` while the scheduler is running: schedule one Volcano job with the original shard, move one node out and another node in, wait for the scheduler cache to refresh through [`RefreshNodeShards`](https://github.com/volcano-sh/volcano/blob/ec82ae319d63a66e46fceeea12f9474ec920c5c3/pkg/scheduler/cache/shard_coordinator.go#L47-L64), then create a second job and assert its pods only land on the updated shard nodes.

2. Verify `NodeShard.status.nodesInUse`, `nodesToAdd`, and `nodesToRemove` during that transition. The status is generated in [`generateNodeShardWithStatus`](https://github.com/volcano-sh/volcano/blob/ec82ae319d63a66e46fceeea12f9474ec920c5c3/pkg/scheduler/cache/shard_coordinator.go#L109-L135) and persisted by [`UpdateNodeShardStatus`](https://github.com/volcano-sh/volcano/blob/ec82ae319d63a66e46fceeea12f9474ec920c5c3/pkg/scheduler/cache/shard_coordinator.go#L95-L107). In particular, when a node is removed from `spec.nodesDesired` but is still in use by the scheduler, it should appear in `nodesToRemove` and should not be treated as immediately reusable by another shard. The reusable-node guard is in [`getAvailableNodesFromShard`](https://github.com/volcano-sh/volcano/blob/ec82ae319d63a66e46fceeea12f9474ec920c5c3/pkg/scheduler/cache/shard_coordinator.go#L138-L146), which subtracts other shards’ `NodeInUse` before exposing available nodes.

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
agrawalcodes: Cool

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
agrawalcodes: made the changes

JesseStutler:
Is it only possible to display the matrix workflow flatly on the GitHub PR page? Can't it be folded? @agrawalcodes 
image
Looks too much on the page, but in the actions, it can be folded
image

hack/run-e2e-kind.sh
```diff
@@ -209,20 +215,32 @@ EOF
...
-        cpuUtilizationMin: 0.7
-        cpuUtilizationMax: 1.0
-        minNodes: 1
-        maxNodes: 100
+        policies:
+          - name: allocation-rate
+            weight: 1
+            arguments:
+              minCPUUtil: 0.0
+              maxCPUUtil: 1.0
```
JesseStutler: Why is there overlap in cpuUtil value with volcano shard here? I think you don't have to change the policy configuration. Also why the agent scheduler's node limit minNodes==maxNodes==1?

pkg/controllers/sharding/policy/arguments.go
```diff
@@ -26,13 +30,36 @@ func (a Arguments) GetInt(ptr *int, key string) {
 		return
 	}
 
-	value, ok := argv.(int)
-	if !ok {
+	switch value := argv.(type) {
```
JesseStutler, line 30: Does this need to be so complicated here? Go's standard JSON library only parses numbers as float64 by default I think

Makefile
```diff
@@ -238,9 +238,30 @@ e2e-test-admission-policy: images
 e2e-test-agentscheduler: images
 	E2E_TYPE=AGENTSCHEDULER ./hack/run-e2e-kind.sh
 
+e2e-test-agentscheduler-none: images
```
JesseStutler: You don't have to specify each target, we can do like 
```
e2e-test-agentscheduler: images
	E2E_TYPE=AGENTSCHEDULER$(if $(SHARDING_MODE),_$(shell echo $(SHARDING_MODE) | tr '[:lower:]' '[:upper:]'),) ./hack/run-e2e-kind.sh

e2e-test-schedulersharding: images
	E2E_TYPE=SCHEDULERSHARDING$(if $(SHARDING_MODE),_$(shell echo $(SHARDING_MODE) | tr '[:lower:]' '[:upper:]'),) ./hack/run-e2e-kind.sh
```

JesseStutler:
@agrawalcodes Looks like there is a CI can't pass and I have re-run it several times, could you check it?

agrawalcodes:
> @agrawalcodes Looks like there is a CI can't pass and I have re-run it several times, could you check it?



> Is it only possible to display the matrix workflow flatly on the GitHub PR page? Can't it be folded? @agrawalcodes image Looks too much on the page, but in the actions, it can be folded image

Thanks for checking.
GitHub shows matrix jobs as separate flat checks on the PR checks view. I don’t think there is a way to make that PR checks list fold like the Actions run page. To reduce the noise, we can either shorten the job names or move toward an aggregate required check, but the individual matrix jobs may still be visible in the PR UI.

agrawalcodes:
> @agrawalcodes Looks like there is a CI can't pass and I have re-run it several times, could you check it?

I checked the admission-policy failure. It was not from the sharding E2E cases directly. The admission-policy install enables VAP/MAP and passes `--enabled-admission=` intentionally, but the webhook manager treated the empty value as an invalid admission path and exited with:

`enabled admission  not found on the admission registered map`

I pushed a fix so an empty enabled-admission list is handled as “no webhook admissions enabled”, with unit coverage

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,677 @@
...
+			}
+
+			Expect(shardNames).To(ContainElement(VolcanoShardName))
+			Expect(shardNames).To(ContainElement(AgentSchedulerShardName))
+		})
+	})
+
+	Describe("Basic Volcano Job Scheduling with Sharding", func() {
+		It("Volcano jobs should be scheduled only to volcano shard nodes", func() {
+			if schedulerShardingMode() == shardingModeNone {
```
JesseStutler: Why not organize the test structures like I said? https://github.com/volcano-sh/volcano/pull/4992/changes#r3297164607 Then ginkgo can use `--label-filter` to filter different tests, it looks very hard to read right now, not quite clear what specific use case is trying to test what mode, and there will be repeated execution. If we use When or Describe to separate the large blocks, it would be better:
```
Describe("Hard Mode", Label("hard"), func() {
    It("xxxxx", func() { ... })
    It("xxxxx", func() { ... })
})

Describe("Soft Mode", Label("soft"), func() {
    It("should fall back outside the shard when resources are exhausted", func() { ... })
})

Describe("None Mode", Label("none"), func() {
    It("xxxxxx", func() { ... })
})
```

Then in the hack/run-e2e-kind.sh you can use ginkgo --label-filter:
```
ginkgo --label-filter=${scheduler_sharding_mode} -v -r ./test/e2e/schedulersharding/
```

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,677 @@
...
+	}
+}
+
+func splitCPURequestsForNode(ctx *e2eutil.TestContext, nodeName string) (corev1.ResourceList, corev1.ResourceList) {
+	node, err := ctx.Kubeclient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	Expect(err).NotTo(HaveOccurred(), "failed to get node %s", nodeName)
+	milliCPU := node.Status.Allocatable.Cpu().MilliValue()
+	Expect(milliCPU).To(BeNumerically(">", 0), "node %s should report positive allocatable CPU", nodeName)
+
+	firstMilliCPU := milliCPU * 6 / 10
```
JesseStutler: I don't get what does *6 *5 means..And besides, you include the control plane node, the control plane node contains some existing pods (such as the Volcano scheduler and Volcano controller), which should be exclueded. I actually think it's best not to include the controller plane node in the tests. Why not use a `Before` method to schedule some placeholder pods first, and then perform the tests? Of course, if you really need a controller plane node to assist tests, you can do that, but I find the current way of calculating how many pods this node can schedule is strange. And there are already some existing helper func can help you

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,677 @@
...
+        arguments:
+          minNodes: 1
+          maxNodes: 1
+  - name: agent-scheduler
+    type: agent
+    policies:
+      - name: allocation-rate
+        weight: 1
+        arguments:
+          minCPUUtil: 0.0
```
JesseStutler: Why is there overlap in usage rates here?

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,677 @@
...
+				},
+			})
+			err := e2eutil.WaitJobReady(ctx, baselineJob)
+			Expect(err).NotTo(HaveOccurred(), "baseline job should become ready on the original shard")
+			expectJobPodsOnNodes(ctx, baselineJob.Name, stringSet(originalVolcanoNodes))
+			err = waitForNodeShardStatus(ctx, VolcanoShardName, originalVolcanoNodes)
+			Expect(err).NotTo(HaveOccurred(), "scheduler should publish original NodeShard status before the spec update")
+
+			By(fmt.Sprintf("Updating sharding config to move volcano shard from %v to %v", volcanoNodes, newVolcanoNodes))
+			originalConfig := updateShardingConfigMap(ctx, preferredNodeShardingConfig())
```
JesseStutler: I didn't quite understand the scenario you were trying to prove with this use case. Could you explain it?

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
JesseStutler: The current `none` and `soft` mode cases mostly exercise the same fallback behavior: when in-shard resources are exhausted, both
  modes can schedule outside the shard. That is expected, but it does not demonstrate the semantic difference between the two modes.

  Please add a case where both in-shard and out-of-shard nodes are feasible:

  - In `soft` mode, create a pod/job that can fit on both the scheduler shard node and an outside-shard node, then assert it lands on
  the scheduler shard node. This verifies the soft preference behavior.
  - In `none` mode, do not assert shard preference. Instead, make the outside-shard node strictly more attractive using normal
  scheduler scoring, for example by making the in-shard node less favorable or constraining available capacity, and assert that
  `NodeShard` membership does not force the pod/job onto the shard node.

test/e2e/agentscheduler/agent_scheduler_test.go
JesseStutler, line 1: Agent Scheduler tests have the same problem: https://github.com/volcano-sh/volcano/pull/4992/changes#r3309010397

And the structure should reconstruct like this is better: https://github.com/volcano-sh/volcano/pull/4992/changes#r3308900176

hack/run-e2e-kind.sh
agrawalcodes: I used minNodes=maxNodes=1 and broad CPU ranges to force a deterministic one-node shard for the fallback tests, but that leaks test assumptions into the install policy and makes the config confusing. I’ll revert this to a normal non-overlapping sharding policy and make the tests prepare their own deterministic node conditions instead.

pkg/controllers/sharding/policy/arguments.go
agrawalcodes: Will simplify this

Makefile
agrawalcodes: Agree. I’ll switch this to a single parameterized target using `SHARDING_MODE` and update the workflow matrix to pass the mode instead of adding separate make targets for every mode.

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
agrawalcodes: The 6/10 and 5/10 split was only meant to exceed the shard node’s allocatable CPU, but it is not obvious. I’ll avoid relying on that calculation and avoid the control-plane for these placement tests where possible.

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
```diff
@@ -0,0 +1,739 @@
...
+			Expect(err).NotTo(HaveOccurred(), "job should become ready")
+
+			By("Verifying all pods are scheduled onto volcano shard nodes")
+			expectJobPodsOnNodes(ctx, job.Name, stringSet(volcanoNodes))
+		})
+
+	})
+
+	Describe("Soft Mode", Label("soft"), func() {
+		It("should prefer the shard node when shard and outside nodes are both feasible", func() {
```
JesseStutler, line 193: Where is "both feasible" demonstrated?

JesseStutler (commented):
Could you clean your commits into 1~3 commits @agrawalcodes , thanks, currently I'm fine with current codes, we can still improve it in v1.16, wish it will not block v1.15 to release but I I with to carry on. Just one question and needs to clean commits

test/e2e/schedulersharding/volcano_scheduler_sharding_test.go
agrawalcodes: "Both feasible" is set up by `prepareSingleWorkerShard`: it pins exactly one worker into the volcano shard via the warmup label, leaves the remaining worker(s) outside the shard, and the test pod requests only `HalfCPU` (500m) with no nodeSelector. In **soft** mode the scheduler doesn't drop outside-shard nodes at the predicate phase, both the shard node and the outside-shard node(s) pass predicates and have CPU room for a 500m request, so feasibility is true on both sides. The assertion that the pod lands on the shard node is what proves the soft preference, since under hard predicates the outside-shard node would still be feasible and under no preference both would be feasible. We can add an explicit pre-check that lists `nodes.allocatable.cpu >= 500m` on both sides if you'd like it more visible.

agrawalcodes:
> Could you clean your commits into 1~3 commits @agrawalcodes , thanks, currently I'm fine with current codes, we can still improve it in v1.16, wish it will not block v1.15 to release but I I with to carry on. Just one question and needs to clean commits

Done

JesseStutler:
/approve
/lgtm
Thanks for your contribution!

PR #4997, opened 2026-01-26
Add predicate helper unit tests

JesseStutler:
#### What type of PR is this?
/kind feature

https://github.com/volcano-sh/volcano/pull/4983#discussion_r2726584627, in https://github.com/volcano-sh/volcano/pull/4983, I submit some unit tests to cover predicate_helper and nodeorder, but the suggested tests from @qi-min missed to be added in that PR, so I raised a new PR to cover these two cases in predicate_helper:
1. Nodes will not be excluded (in the node list but not in node shard) in soft mode
2. Nodes in shard will not be excluded when in hard mode ( the node both in node list and node shard)

Fixes https://github.com/volcano-sh/volcano/pull/4983#discussion_r2726584627

```release-note

```

### DIFF
--- a/pkg/scheduler/util/predicate_helper_test.go
+++ b/pkg/scheduler/util/predicate_helper_test.go
@@ -155,6 +155,8 @@ func TestPredicateNodes(t *testing.T) {
 			},
 			nodes: []*api.NodeInfo{
 				{Name: "node1"},
+				{Name: "node2"},
+				{Name: "node3"},
 			},
 			nodesInShard:     sets.New[string]("node1"),
 			shardingMode:     commonutil.HardShardingMode,
@@ -162,7 +164,12 @@ func TestPredicateNodes(t *testing.T) {
 			predicateFn:      func(*api.TaskInfo, *api.NodeInfo) error { return nil },
 			expectedNodes:    []string{"node1"},
 			expectedErr:      "",
-			expectedErrCache: map[string]map[string]string{},
+			expectedErrCache: map[string]map[string]string{
+				"job1/worker": {
+					"node2": "node isn't in scheduler node shard",
+					"node3": "node isn't in scheduler node shard",
+				},
+			},
 		},
 	}
 


JesseStutler:
@hajnalmt Mate could you help add approve and lgtm label for this pr?

hajnalmt:
/lgtm

Yes

pkg/scheduler/util/predicate_helper_test.go
```diff
@@ -126,6 +126,44 @@ func TestPredicateNodes(t *testing.T) {
...
+				Name:      "task",
+			},
+			nodes: []*api.NodeInfo{
+				{Name: "node1"},
+			},
+			nodesInShard:     sets.New[string]("node1"),
+			shardingMode:     commonutil.HardShardingMode,
+			enableErrorCache: false,
+			predicateFn:      func(*api.TaskInfo, *api.NodeInfo) error { return nil },
+			expectedNodes:    []string{"node1"},
```
hzxuzhonghu, line 165: you only construced one node, how could this case make sense

pkg/scheduler/util/predicate_helper_test.go
```diff
@@ -126,6 +126,44 @@ func TestPredicateNodes(t *testing.T) {
...
+		{
+			name: "soft sharding allows nodes outside shard",
+			task: &api.TaskInfo{
+				Job:       "job1",
+				TaskRole:  "worker",
+				Namespace: "ns",
+				Name:      "task",
+			},
+			nodes: []*api.NodeInfo{
+				{Name: "node2"},
```
hzxuzhonghu, line 138: why all nodes does not include the `nodesInShard`

JesseStutler: Follow the suggestion from @qi-min: https://github.com/volcano-sh/volcano/pull/4983#discussion_r2720378975, I add this case to validate under soft mode, even if it is not in `nodeInShard`, nodes are considered as candidate nodes as long as they pass predicate

pkg/scheduler/util/predicate_helper_test.go
JesseStutler: I updated to add more nodes, please check, thanks

pkg/scheduler/util/predicate_helper_test.go
hzxuzhonghu: Is this a normal case, i mean i cannot imagine how this can happen All nodes does not include nodesInShard

qi-min: It is possible. NodesInShard only store nodes for current scheduler shard.  In soft mode, scheduler is aware of all nodes including node within this shard or other shards, so nodes in other shard may also pass predicate.

PR #4999, opened 2026-01-26
fix(scheduler): add nil checks in BuildVictimsPriorityQueue

goyalpalak18:
## Description
This PR fixes a critical panic in the `BuildVictimsPriorityQueue` function within `pkg/scheduler/framework/session_plugins.go`.

**The Problem:**
During preemption or reclaim cycles, the scheduler builds a priority queue of victim tasks. The comparator function attempts to look up the Jobs associated with these tasks (`ssn.Jobs[task.Job]`). However, if a PodGroup is deleted (e.g., via `kubectl delete podgroup`) after the session snapshot is taken but before the queue is built, the lookup returns `nil`.

Previously, the code fell through to `ssn.JobOrderFn(lvJob, rvJob)` without checking if `lvJob` or `rvJob` were valid. This resulted in a `nil pointer dereference` panic when `JobOrderFn` attempted to access fields like `CreationTimestamp` on a nil struct, causing the scheduler process to crash and halting cluster-wide scheduling.

**The Fix:**
I have added defensive nil checks in the comparator before calling `JobOrderFn`. The sorting logic now handles orphaned tasks gracefully:
1. **Both jobs missing:** Sort by task `CreationTimestamp`.
2. **One job missing:** The task with the missing job (orphan) is prioritized for eviction (sorted as "less" priority).
3. **Both jobs present:** Proceed with standard `JobOrderFn`.

## How to test
To reproduce the crash (before this fix):
1. Create a queue and submit multiple jobs (e.g., `job1`, `job2`) to ensure resources are consumed.
2. Submit a high-priority job to trigger preemption.
3. While the scheduler is evaluating preemption, delete one of the running PodGroups (`kubectl delete podgroup pg-job2 --wait=false`).
4. **Result:** The scheduler crashes with `panic: runtime error: invalid memory address or nil pointer dereference`.

With this PR applied, the scheduler handles the orphaned tasks gracefully, prioritizes them for eviction, and continues scheduling without crashing.

### DIFF
--- a/pkg/scheduler/framework/session_plugins.go
+++ b/pkg/scheduler/framework/session_plugins.go
@@ -1121,7 +1121,11 @@ func (ssn *Session) BuildVictimsPriorityQueue(victims []*api.TaskInfo, preemptor
 			return !lvJobFound
 		}
 
-		if preemptorJobFound && lvJob.Queue != rvJob.Queue {
+		if !preemptorJobFound {
+			return !ssn.JobOrderFn(lvJob, rvJob)
+		}
+
+		if lvJob.Queue != rvJob.Queue {
 			return ssn.VictimQueueOrderFn(ssn.Queues[lvJob.Queue], ssn.Queues[rvJob.Queue], ssn.Queues[preemptorJob.Queue])
 		}
 


goyalpalak18:
Hey @hajnalmt @JesseStutler, I’ve pushed a fix for the nil pointer crash during preemption. I'd appreciate a review when you have a moment. Thanks!

pkg/scheduler/framework/session_plugins.go
```diff
@@ -1102,7 +1102,18 @@ func (ssn *Session) BuildVictimsPriorityQueue(victims []*api.TaskInfo, preemptor
 		rvJob, rvJobFound := ssn.Jobs[rv.Job]
 		preemptorJob, preemptorJobFound := ssn.Jobs[preemptor.Job]
 
-		if lvJobFound && rvJobFound && preemptorJobFound && lvJob.Queue != rvJob.Queue {
+		// Handle orphaned tasks from deleted PodGroups.
+		// Tasks with missing jobs should be prioritized for eviction (sorted first in victims queue).
+		if !lvJobFound || !rvJobFound {
+			if !lvJobFound && !rvJobFound {
+				// Both jobs missing: fall back to task creation time
+				return lv.Pod.CreationTimestamp.Before(&rv.Pod.CreationTimestamp)
```
hajnalmt: This is not right, we should follow at least the `TaskOrderFns` please.
`return !ssn.TaskOrderFn(l, r)`

pkg/scheduler/framework/session_plugins.go
```diff
@@ -1102,7 +1102,18 @@ func (ssn *Session) BuildVictimsPriorityQueue(victims []*api.TaskInfo, preemptor
 		rvJob, rvJobFound := ssn.Jobs[rv.Job]
 		preemptorJob, preemptorJobFound := ssn.Jobs[preemptor.Job]
 
-		if lvJobFound && rvJobFound && preemptorJobFound && lvJob.Queue != rvJob.Queue {
+		// Handle orphaned tasks from deleted PodGroups.
```
hajnalmt: Please elaborate on this, how it can happen because we take the snapshot too soon and delete the podgroup etc...

pkg/scheduler/framework/session_plugins.go
goyalpalak18: Done, updated to use !ssn.TaskOrderFn(l, r) instead of manual timestamp comparison.

pkg/scheduler/framework/session_plugins.go
goyalpalak18: Sure, this can happen due to a race condition between the scheduler's snapshot and PodGroup deletion:

1. Scheduler takes a snapshot of the cluster state at the start of a scheduling cycle.
2. The snapshot includes tasks that belong to a PodGroup/Job.
3. While the scheduler is processing (before reaching `BuildVictimsPriorityQueue`), a user or controller deletes the PodGroup.
4. The job gets removed from `ssn.Jobs`, but the tasks from the original snapshot still reference it via `task.Job`.
5. When we try to look up `ssn.Jobs[task.Job]`, it returns nil because the job no longer exists.

These orphaned tasks should still be handled gracefully rather than causing a panic. Since they belong to a deleted PodGroup, they are good candidates for eviction anyway.

hajnalmt (commented):
/priority high

Thanks for the  contribution.

pkg/scheduler/framework/session_plugins.go
hajnalmt: I understand the problem  😄 I meant elaborate in the comment why we have this change here. The comment here is not explanatory enough. Just mentions that we handle orphaned tasks here.

goyalpalak18: @hajnalmt Updated the comment to explain the race condition scenario in detail. Thanks for the feedback!

PR #5002, opened 2026-01-30
fix(cache): validate PodGroup before eviction to prevent cache inconsistency

Yashika0724:
This PR addresses an ordering issue in SchedulerCache.Evict() that can lead to inconsistent scheduler cache state under normal Kubernetes race conditions.

Previously, Evict() could update task state, modify node resource accounting, and start the eviction goroutine before validating that the Job’s PodGroup exists and can be converted. If the PodGroup was deleted concurrently (for example during preemption or controller reconciliation), the function would return an error after eviction had already begun. In this case, the caller rolls back session state, but the scheduler cache state is not reverted.

This change moves PodGroup validation and conversion to the beginning of Evict(), ensuring that eviction only proceeds after all required invariants are satisfied. On validation failure, the function now returns early without any side effects.

The fix is intentionally minimal and does not alter behavior for successful eviction paths. It simply prevents partial eviction and cache/session divergence when PodGroups are removed concurrently.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -895,7 +895,7 @@ func (sc *SchedulerCache) Evict(taskInfo *schedulingapi.TaskInfo, reason string)
 	}
 	podgroup := &vcv1beta1.PodGroup{}
 	if err = schedulingscheme.Scheme.Convert(&job.PodGroup.PodGroup, podgroup, nil); err != nil {
-		klog.Errorf("Error while converting PodGroup to v1alpha1.PodGroup with error: %v", err)
+		klog.Errorf("Error while converting PodGroup to v1beta1.PodGroup with error: %v", err)
 		return err
 	}
 


Yashika0724:
Hi @hajnalmt , thanks for reviewing.
This PR moves PodGroup validation earlier in Evict() to avoid partial state updates when PodGroups are removed concurrently. The change is small and does not affect the normal eviction path.
Please let me know if any adjustments are needed.

pkg/scheduler/cache/cache.go
```diff
@@ -887,6 +887,18 @@ func (sc *SchedulerCache) Evict(taskInfo *schedulingapi.TaskInfo, reason string)
...
 
+	// Check PodGroup and prepare for event recording BEFORE any state changes.
+	// This ensures we don't start eviction if PodGroup is nil or conversion fails.
+	if job.PodGroup == nil {
+		return fmt.Errorf("cannot evict Task %v: PodGroup of Job <%s/%s> is nil",
+			task.UID, job.Namespace, job.Name)
+	}
+	podgroup := &vcv1beta1.PodGroup{}
+	if err = schedulingscheme.Scheme.Convert(&job.PodGroup.PodGroup, podgroup, nil); err != nil {
+		klog.Errorf("Error while converting PodGroup to v1alpha1.PodGroup with error: %v", err)
```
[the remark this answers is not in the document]
hajnalmt: There is a typo here can you fix that please?

Yashika0724:
Fixed 
thanks for catching that!

pkg/scheduler/cache/cache.go
```diff
@@ -887,6 +887,18 @@ func (sc *SchedulerCache) Evict(taskInfo *schedulingapi.TaskInfo, reason string)
 			task.UID, task.NodeName)
 	}
 
+	// Check PodGroup and prepare for event recording BEFORE any state changes.
+	// This ensures we don't start eviction if PodGroup is nil or conversion fails.
+	if job.PodGroup == nil {
```
JesseStutler, line 892: I want to ensure that is it will prevent us from evicting those pods not scheduled by volcano but already running?

Yashika0724: > I want to ensure that is it will prevent us from evicting those pods not scheduled by volcano but already running?

Thanks for checking!

This change only prevents eviction when the associated Volcano Job has a nil PodGroup. Pods not scheduled by Volcano should not be affected because SchedulerCache.Evict() is only triggered for tasks managed through Volcano job scheduling flows. The intention here is to avoid starting eviction when the job metadata has already been removed, which could otherwise leave cache state inconsistent.

Please let me know if I am missing any scenario.

JesseStutler: No, the pods that are not scheduled by Volcano but have been running will also have JobInfo, but the podgroup is nil. Can you test it: the pods by the Kube-scheduler have been running, but are then evicted by Volcano, can the original code succeed, will your code change prevent this from happening?

hajnalmt: You are right Jesse, I think this code will prevent that from happening, but I thought we shouldn't evict pods if they are not managed by volcano in the first place right?

The other thing is that in these cases we evict the pod, but fail afterwards? This means that these not volcano managed pods will be considered to be not evicted by the caller (prempt/reclaim) but the pod is actually evicted. This doesn't seem right still.

JesseStutler: @hajnalmt I think you're right mate, I will agree to merge this

PR #5010, opened 2026-01-31
Fix/preemption eviction rollback

Aman-Cool:
### Background

In the preemption workflow, victim tasks are evicted before attempting to pipeline the preemptor. These eviction operations are recorded in the scheduler `Statement`.

If `stmt.Pipeline()` fails, only `UnPipeline()` is invoked, which rolls back the preemptor state but does not revert the previously recorded eviction operations. When the statement is later committed, victim evictions are applied even though the preemption did not succeed.

### Problem

This behavior breaks the atomicity assumption of the `Statement` pattern and can result in:

- Victim tasks being evicted without the preemptor being scheduled
- Gang-scheduled jobs dropping below `MinAvailable`
- Unnecessary and silent workload disruption

### Fix

On `Pipeline()` failure, the preemption logic now:

- Rolls back the preemptor state
- Discards the current statement to revert all recorded eviction operations
- Treats the preemption attempt as failed

This ensures that eviction and pipeline operations are applied atomically.

### Impact

- Victim tasks are only evicted when preemption completes successfully
- Gang scheduling guarantees are preserved
- Failed preemption attempts no longer cause side effects

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt_test.go
+++ b/pkg/scheduler/actions/preempt/preempt_test.go
@@ -293,6 +293,43 @@ func TestPreempt(t *testing.T) {
 			},
 			ExpectEvictNum: 0,
 		},
+		{
+			// Verify that when preemption succeeds on n2 after failing on n1, only n2's
+			// victims are committed — n1's evictions must be discarded by nodeStmt.Discard().
+			//
+			// n1 victim (1 CPU): after eviction queue.Allocated drops from 5 to 4 CPU;
+			// 4 + 3 (preemptor) = 7 > 6 (cap) → Allocatable fails → nodeStmt discarded.
+			// n2 victim (4 CPU): after eviction queue.Allocated drops from 5 to 1 CPU;
+			// 1 + 3 = 4 ≤ 6 → Allocatable passes → pipeline succeeds → nodeStmt merged.
+			//
+			// Without the per-node nodeStmt isolation, n1's victim would be committed
+			// alongside n2's because the single shared statement would not be rolled back.
+			Name: "only commit evictions on the node where preemption succeeds",
+			PodGroups: []*schedulingv1beta1.PodGroup{
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 0, map[string]int32{}, schedulingv1beta1.PodGroupInqueue, "low-priority"),
+				util.BuildPodGroupWithPrio("pg2", "c1", "q1", 1, map[string]int32{"": 1}, schedulingv1beta1.PodGroupInqueue, "high-priority"),
+				util.BuildPodGroupWithPrio("pg3", "c1", "q1", 0, map[string]int32{}, schedulingv1beta1.PodGroupInqueue, "low-priority"),
+			},
+			Pods: []*v1.Pod{
+				// n1 victim: small — freeing 1 CPU is not enough to satisfy Allocatable
+				util.BuildPod("c1", "preemptee1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				// n2 victim: large — freeing 4 CPU brings queue.Allocated within cap
+				util.BuildPod("c1", "preemptee2", "n2", v1.PodRunning, api.BuildResourceList("4", "4G"), "pg3", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				util.BuildPod("c1", "preemptor1", "", v1.PodPending, api.BuildResourceList("3", "3G"), "pg2", make(map[string]string), make(map[string]string)),
+			},
+			// Cluster: 7 CPU total. Queue cap: 6 CPU → deserved = 6 CPU.
+			// n1 has 2 CPU idle so ValidateVictims passes (2 idle + 1 victim = 3 ≥ 3).
+			// n2 has 0 CPU idle; ValidateVictims passes via victim alone (4 ≥ 3).
+			Nodes: []*v1.Node{
+				util.BuildNode("n1", api.BuildResourceList("3", "3G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+				util.BuildNode("n2", api.BuildResourceList("4", "4G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+			},
+			Queues: []*schedulingv1beta1.Queue{
+				util.BuildQueue("q1", 1, api.BuildResourceList("6", "6G")),
+			},
+			ExpectEvictNum: 1,
+			ExpectEvicted:  []string{"c1/preemptee2"},
+		},
 	}
 
 	trueValue := true


Aman-Cool:
@shinytang6 @hajnalmt @JesseStutler ,Fixes a case where victims could be evicted even if preemption failed. We now roll back and discard the statement on pipeline failure so evictions only happen when preemption succeeds.

pkg/scheduler/framework/statement_test.go
```diff
@@ -0,0 +1,236 @@
...
+	stmt.RollbackOperationsFrom(-1)
+	assert.Equal(t, 0, stmt.OperationCount(), "Negative index should not affect operations")
+
+	// Test rollback with index >= operation count (should do nothing)
+	stmt.RollbackOperationsFrom(0)
+	assert.Equal(t, 0, stmt.OperationCount(), "Index >= count should not affect operations")
+
+	stmt.RollbackOperationsFrom(100)
+	assert.Equal(t, 0, stmt.OperationCount(), "Large index should not affect operations")
+}
```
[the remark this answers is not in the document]
Aman-Cool: The tests have been updated to cover boundary cases where the statement already contains operations, including `fromIndex == len(operations)` and `fromIndex > len(operations)`. This ensures `RollbackOperationsFrom` behaves correctly in both no-op and full-rollback scenarios, not just when the operations list is empty.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -491,6 +511,13 @@ func (pmpt *Action) topologyAwarePreempt(
 			klog.Errorf("Failed to unpipeline Task %v on %v in Session %v for %v.",
 				preemptor.UID, bestCandidate.Name(), ssn.UID, rollbackErr)
 		}
+		// CRITICAL FIX: Rollback evictions when pipeline fails.
+		// Without this, victims would be evicted even though the preemptor
+		// is not scheduled, potentially violating gang scheduling constraints.
+		stmt.RollbackOperationsFrom(opCountBeforeEvictions)
```
hajnalmt: I see your issue, but we have an API to do this `SaveOperations` and `RecoverOperations`. 

We basically save the original statement, and recover it later if needed. At least the way we do this in `allocate` is that we create a new statement and if something goes wrong, then we `Discard` the whole statement, otherwise we merge the two. If we expose an API like this on the statement we encourage this new statement usage at other places too and I think we don't need two APIs with the same intention.

Also I see more points where this can go wrong. If `opCountBeforeEvictions` is wrongfully used somewhere we loose operations. I feel that saving and recovering is much cleaner, but your approach worth a discussion though on the community meeting.

pkg/scheduler/framework/statement.go
```diff
@@ -259,6 +259,54 @@ func (s *Statement) UnPipeline(task *api.TaskInfo) error {
 	return nil
 }
 
+// Unevict reverts an eviction, restoring the task to Running status.
+// This is used to rollback evictions when a preemption attempt fails.
+func (s *Statement) Unevict(reclaimee *api.TaskInfo) error {
```
hajnalmt: I think this was intended that this wasn't exposed on the statement reciever. So Eviction cannot be "undone" on statements like this.

hajnalmt (commented):
Hello,
Thank you for the change! I see you have worked on this quite a lot, props for the unit tests too (It's quite mind-boggling for me that we don't have any yet for the statement handling). Appreciated!

I think the issue you have found is legit, but the way you approached it is not right. Can we first create an actual issue to track the solution for this?
I will bring this issue up on the community meeting on friday. Feel free to join 😅

Aman-Cool:
@hajnalmt ,Thanks for the detailed feedback.

I agree the issue is architectural rather than just a local fix, and I see your point about preferring save/recover or discard semantics over exposing rollback behavior on the statement receiver.

I’m happy to open an issue to capture the problem and discuss the intended approach there. The main goal from my side was to ensure we don’t apply evictions unless preemption fully succeeds, but I’m totally open to reworking the solution to better align with the existing APIs and patterns.

I’ll create an issue shortly and would be glad to discuss this further. 

Thanks for taking a close look at this.

hzxuzhonghu:
can you help sortout all the sequences the preempt does, and so others can understand what need to fix.

I feel the fix here is too hacky, depending on the index.

Aman-Cool:
@hxuzhonghu ,Thanks for the feedback, I’ve opened an issue to document the full preemption sequence, the rollback problem, and the broader design gap around transactional semantics: #5026.
The current fix focuses on restoring correctness (avoiding eviction side effects when preemption fails), while the issue tracks a cleaner, more explicit model so we don’t rely on ordering/index assumptions long-term. Happy to iterate further based on that discussion.

pkg/scheduler/framework/statement.go
hajnalmt: This wasn't intendented you can expose this! This is fine.

pkg/scheduler/framework/statement.go
```diff
@@ -259,6 +259,54 @@ func (s *Statement) UnPipeline(task *api.TaskInfo) error {
...
+	return s.unevict(reclaimee)
+}
+
+// OperationCount returns the current number of operations in the statement.
+// This can be used to track operations before a preemption attempt and
+// rollback only the new operations if the attempt fails.
+func (s *Statement) OperationCount() int {
```
hajnalmt: This is a good API too!

pkg/scheduler/framework/statement.go
```diff
@@ -259,6 +259,54 @@ func (s *Statement) UnPipeline(task *api.TaskInfo) error {
...
+	return len(s.operations)
+}
+
+// RollbackOperationsFrom rolls back all operations from the given index onwards.
+// This is used to undo evictions when a preemption pipeline fails, without
+// affecting earlier successful operations in the same statement.
+func (s *Statement) RollbackOperationsFrom(fromIndex int) {
```
hajnalmt: Instead of this we can have a merge API on the Statement reciever.
`func (s *Statement) Merge (stmts ...*Statement)`

Aman-Cool:
@hajnalmt ,Thanks for the detailed feedback and for pointing me in the right direction.

I’ve updated the PR to fully switch to the Merge based Statement approach we discussed. All eviction and pipeline work now happens on temporary statements, and nothing touches the caller statement unless the preemption actually succeeds. On success the temp statement is merged, and on any pipeline failure or insufficient resources it is simply discarded, so no eviction side effects leak through.

I’ve removed all the OperationCount and index based rollback logic, as well as the pattern where assigned could end up true even when the pipeline failed. Unevict stays private and is no longer needed for the main rollback path since Discard on the temp statement handles that cleanly.

I also added unit tests around the new Merge API to make the behavior explicit and easier to reason about going forward.

Please let me know if this matches what you had in mind or if there are any other adjustments you’d like to see.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -378,7 +385,7 @@ func (pmpt *Action) normalPreempt(
 			preemptee := victimsQueue.Pop().(*api.TaskInfo)
 			klog.V(3).Infof("Try to preempt Task <%s/%s> for Task <%s/%s>",
 				preemptee.Namespace, preemptee.Name, preemptor.Namespace, preemptor.Name)
-			if err := stmt.Evict(preemptee, "preempt"); err != nil {
+			if err := nodeStmt.Evict(preemptee, "preempt"); err != nil {
```
hajnalmt, line 389: We need `Unevict` here. Sorry I maybe confused you. If we don't Unevict then if the Pipeline-ing is successful, then we can evict pods wrongfully still.

Imagine:
You are preempting for task1, try to evict task2, the eviction is unsuccesful for it (for whatever reason, and you continue here) then eviction is successful for task3  -> you pipeline task1 do the commit, you are evicting task2 (it modifies the cache in session at least if the eviction is unsuccessful) and then task3 although you should evict task3 only.

pkg/scheduler/actions/preempt/preempt.go
hajnalmt, line 371: Can you move the creation of `nodeStmt` here?
We are not using it before this for loop. If ValidateVictims runs to an error we have created the variable for nothing.

Aman-Cool:
@hajnalmt , Thanks a lot for the clarification and for sticking with this review, really appreciate it.

I’ve re-exposed `UnEvict` and now use it **only** to clean up failed eviction attempts (separate from pipeline rollback). The transactional boundary remains `Merge` / `Discard`, exactly as discussed.

I also moved the `nodeStmt` creation to after `ValidateVictims` succeeds so we don’t allocate state unnecessarily on early exits.

I double-checked `unevict()` to make sure it’s safe for partial eviction failures and scoped to statement-local state only.

Please let me know if this looks good now or if there’s anything you’d like adjusted further, happy to iterate.

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -357,6 +359,11 @@ func (pmpt *Action) normalPreempt(
 			continue
 		}
 
+		// Use a temporary statement per node attempt so that eviction side effects
+		// are isolated. On success the operations are merged into the caller's
+		// statement; on failure they are discarded, leaving the session clean.
+		nodeStmt := framework.NewStatement(ssn)
```
hzxuzhonghu, line 366: I donot quite understand this comment

hzxuzhonghu: especially leaving the `session clean`

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -480,19 +496,29 @@ func (pmpt *Action) topologyAwarePreempt(
 		return false, fmt.Errorf("no candidate node for preemption")
 	}
 
-	if status := prepareCandidate(bestCandidate, preemptor.Pod, stmt, ssn); !status.IsSuccess() {
+	// Use a temporary statement so that eviction side effects are only applied
```
hajnalmt, line 500: Please modify this too to "evictions are only commited" instead of "eviction side effects are only applied" to be cleaner.

pkg/scheduler/actions/preempt/preempt.go
hajnalmt: You are right, this comment is not well written here, as it doesn't prevent any mingling with the session nor any general eviction side effects, as this temporary statement works on the same session so it modifies it.

Is it enough to write maybe that "Using a temporary statement per node to store node-level operations that can be later merged to the job-level statement or discarded as a whole." Is this cleaner?

hzxuzhonghu:
>If the job/node disappears from the session due to some event for example, or the the task update fails, it won't work. We need to rollback Evict in this case somehow, I think it's a bug that we don't do it now.

What i mean is the error not actually returned

pkg/scheduler/actions/preempt/preempt.go
hzxuzhonghu: I donot understand the interference and please give an concrete example or write a test case

Aman-Cool:
Hey @hajnalmt, I can see how my comment about "leaving the session clean" was confusing, let me clarify. 

What I meant is that by using a temporary statement per node attempt, we isolate the eviction operations from the caller's statement. On success, we merge them in. On failure, we discard them. The caller's statement stays clean because we never directly modify it with operations that might need to be rolled back.

I'll update the comment to be clearer, maybe something like "evictions are only committed" as you suggested, since that's more precise about what's actually happening. Thanks for the suggestion!

Aman-Cool:
@hzxuzhonghu ,You mentioned you don't understand the interference and asked for a concrete example. Here's the scenario this PR fixes:

**Before (the bug):**
```go
// Preemption currently does this
for each victim {
    stmt.Evict(victim)  // Records eviction in statement
}

err := stmt.Pipeline(preemptor, node)
if err != nil {
    stmt.UnPipeline(preemptor)  // Rolls back preemptor
    // BUT: evictions are still in stmt and will be committed!
}

stmt.Commit()  // Commits both evictions AND preemptor state
// Result: victims evicted, preemptor not scheduled = bad 😱
```

**After (my fix):**
```go
// Use a temp statement per node attempt
tempStmt := NewStatement(ssn)

for each victim {
    tempStmt.Evict(victim)  // Records in TEMP statement
}

err := tempStmt.Pipeline(preemptor, node)
if err != nil {
    tempStmt.Discard()  // Throws away the whole temp statement
    continue            // Try next node
}

// Only if pipeline succeeded:
stmt.Merge(tempStmt)  // Transfer operations to caller's statement
```

The key difference: evictions and pipeline are **atomic** ; either both succeed (merge) or both fail (discard).

Aman-Cool:
@hzxuzhonghu I saw you opened #5043 to remove error returns, I left a comment there, but wanted to explain here how it affects this fix.

The Merge/Discard approach depends on being able to detect when operations fail:

```go
// What this PR does
tempStmt := NewStatement(ssn)
tempStmt.Evict(victims...)
err := tempStmt.Pipeline(preemptor, node)

if err != nil {
    tempStmt.Discard()  // throw away the evictions
} else {
    stmt.Merge(tempStmt)  // commit the evictions
}
```

Without that error check, we can't decide whether to commit or discard the evictions. It's like ordering a pizza, kicking everyone out of your house to make room for the pizza guy, and then the pizza never shows up. Now you're sitting alone in an empty house with no pizza. And no friends. 

It's like musical chairs but the music never stops and people just keep getting eliminated for no reason.

The Merge/Discard approach is actually way cleaner than manually tracking operations, we just need error detection so we can hit CTRL+Z on the whole thing when it fails, instead of committing to the chaos.

For this PR, I can:
1. Update the comment to be clearer (as @hajnalmt suggested - apparently my comments need comments)
2. Keep `UnEvict()` exposed (as we discussed)
3. Add test coverage for the failure scenarios (there's something deeply satisfying about watching a test called `testPreemptionFailsGracefullyInsteadOfEvictingInnocentVictims` pass)

Let me know your thoughts!

Aman-Cool:
@hajnalmt Made those changes, comment is clearer now, UnEvict is exposed, and added the rollback test. 
Let me know if anything else needs adjusting!

hzxuzhonghu:
@hajnalmt I understand your concern, that need to look into one by one. 

come back to this fix,  I am all for the fix on Execute() func.  But  i am confused by the statment merge stuff. 

What i understand is we do evict and pipleline in order, even Evict succeed while pipeline not, wthen we do statement.Discard().  

```golang
Evict()

Pipeline()
```

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -378,9 +386,12 @@ func (pmpt *Action) normalPreempt(
 			preemptee := victimsQueue.Pop().(*api.TaskInfo)
 			klog.V(3).Infof("Try to preempt Task <%s/%s> for Task <%s/%s>",
 				preemptee.Namespace, preemptee.Name, preemptor.Namespace, preemptor.Name)
-			if err := stmt.Evict(preemptee, "preempt"); err != nil {
+			if err := nodeStmt.Evict(preemptee, "preempt"); err != nil {
 				klog.Errorf("Failed to preempt Task <%s/%s> for Task <%s/%s>: %v",
 					preemptee.Namespace, preemptee.Name, preemptor.Namespace, preemptor.Name, err)
+				if e := nodeStmt.UnEvict(preemptee); e != nil {
```
hzxuzhonghu, line 392: And i think we donot even need this, as we will call Discard at last

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -397,20 +408,26 @@ func (pmpt *Action) normalPreempt(
 
 		// If preemptor's queue is not allocatable, it means preemptor cannot be allocated. So no need care about the node idle resource
 		if ssn.Allocatable(currentQueue, preemptor) && preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
-			if err := stmt.Pipeline(preemptor, node.Name, evictionOccurred); err != nil {
```
hzxuzhonghu, line 400: I am thinking we can continue using the same stmt. It will record either failure or success in its operations

Aman-Cool:
@hzxuzhonghu The reason for temp statements is that we try **multiple nodes** in a loop.

If we use one statement for all nodes:
- Try node1, evict victims, pipeline fails
- Call `Discard()` -> throws away all evictions
- Now we can't try node2 because everything's gone

With temp statements:
- Try node1 with `nodeStmt1`, fails -> discard just that
- Try node2 with fresh `nodeStmt2`, succeeds -> merge it

Each node gets a fresh attempt. Without this, the first failure would prevent trying other nodes.

About `UnEvict()` ,that's for when the Evict call itself fails (like if the victim doesn't exist in the session). We clean it up immediately before trying the next victim.

hzxuzhonghu:
>If we use one statement for all nodes:

>Try node1, evict victims, pipeline fails
>Call Discard() -> throws away all evictions

@Aman-Cool Discard is only called at last, why does this matter? 

>Now we can't try node2 because everything's gone

Combined with above, can you elaborate the influence, btw i donot see a test on this

Aman-Cool:
@hzxuzhonghu Yeah, Discard only happens at the end, which is actually the problem! 😅

The issue isn't when everything fails (that's boring and predictable). It's when node1 fails but node2 succeeds (the plot twist).

Say we try 2 nodes with one statement:

**node1:**
- stmt.Evict(victim_A) 
- stmt.Evict(victim_B)
- stmt.Pipeline(...) fails → continue to node2

**node2:**
- stmt.Evict(victim_C)
- stmt.Pipeline(...) succeeds 

Now when stmt.Commit() runs, it evicts ALL three victims - A, B, and C. But victims A and B shouldn't be evicted because node1's pipeline failed! The preemptor never actually scheduled on node1, but those victims still got kicked out anyway. It's like being fired from a job you were never hired for.

We can't Discard after node1 fails because we haven't tried node2 yet; Discard would throw everything away including our hopes and dreams. But if we don't discard, node1's victims get committed anyway. Rock, meet hard place.

With temp statements per node:
- node1's temp statement: evict A and B, pipeline fails → discard
- node2's temp statement: evict C, pipeline succeeds → merge into main stmt
- Only victim C gets evicted, A and B live to see another day

---

About the test - you're right, the existing one doesn't catch this. Just pushed a new test (332538e) that actually breaks without the fix:

- node1: 1 CPU victim, allocation fails → nodeStmt.Discard()
- node2: 4 CPU victim, allocation succeeds → nodeStmt.Merge()
- **Without per-node statements:** both victims evicted (ExpectEvictNum=2)
- **With the fix:** only node2's victim evicted (ExpectEvictNum=1) 

Should clear things up!

hzxuzhonghu (commented):
Thanks for the complete deep dive and fix.

/lgtm

PR #5033, opened 2026-02-08
feat: add opt-in scheduling gates for queue capacity control

### LINKED ISSUE #4710: Cluster Autoscaler node scale-up for Pods that exceed Queue's capability

devzizu:
#### What type of PR is this?

/kind feature

This PR adds opt-in scheduling gates for queue capacity control to prevent cluster autoscalers from triggering unnecessary scale-ups when pods are waiting for queue admission rather than cluster capacity. 

When a pod opts in via the `scheduling.volcano.sh/queue-allocation-gate: "true"` annotation, Volcano adds a scheduling gate that delays the `Unschedulable` condition until the queue has capacity.

The design is further described in https://github.com/volcano-sh/volcano/pull/4727.

Fixes https://github.com/volcano-sh/volcano/issues/4710

- This feature is opt-in via pod annotation to ensure backward compatibility.
- A simple E2E test was added to validate capacity reservation behavior, but further test cases may be required.

```release-note
Add opt-in scheduling gates for queue capacity control. Pods can now use the scheduling.volcano.sh/queue-allocation-gate: "true" annotation to prevent false autoscaler scale-ups when waiting for queue admission.
```

### DIFF
--- a/installer/helm/chart/volcano/templates/admission.yaml
+++ b/installer/helm/chart/volcano/templates/admission.yaml
@@ -145,8 +145,8 @@ spec:
             {{- if $scheduler_name }}
             - --scheduler-name={{- $scheduler_name }}
             {{- end }}
-            {{- if .Values.custom.scheduler_feature_gates }}
-            - --feature-gates={{ .Values.custom.scheduler_feature_gates }}
+            {{- if .Values.custom.admission_feature_gates }}
+            - --feature-gates={{ .Values.custom.admission_feature_gates }}
             {{- end }}
             - --enable-healthz=true
             - --logtostderr
--- a/installer/helm/chart/volcano/templates/admission.yaml
+++ b/installer/helm/chart/volcano/templates/admission.yaml
@@ -145,6 +145,9 @@ spec:
             {{- if $scheduler_name }}
             - --scheduler-name={{- $scheduler_name }}
             {{- end }}
+            {{- if .Values.custom.scheduler_feature_gates }}
+            - --feature-gates={{ .Values.custom.scheduler_feature_gates }}
+            {{- end }}
             - --enable-healthz=true
             - --logtostderr
             - --port={{.Values.basic.admission_port}}
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -17,17 +17,14 @@
 package allocate
 
 import (
-	"context"
 	"fmt"
 	"math"
 	"slices"
 	"sync"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
 	"volcano.sh/apis/pkg/apis/scheduling"
@@ -35,7 +32,6 @@ import (
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/features"
 	"volcano.sh/volcano/pkg/scheduler/api"
-	"volcano.sh/volcano/pkg/scheduler/cache"
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/metrics"
@@ -114,28 +110,14 @@ type Action struct {
 
 	recorder *Recorder
 
-	// Async gate management infrastructure
-	kubeClient              kubernetes.Interface // Cached client for worker goroutines
-	schGateRemovalCh        chan schGateRemovalOperation
-	schGateRemovalWorkersWg sync.WaitGroup
-	schGateRemovalStopCh    chan struct{}
-	gateRemovalWorkerNum    int // Number of async gate removal workers
-	initOnce                sync.Once
-	shutdownOnce            sync.Once
-}
-
-// schGateRemovalOperation is a struct that contains the namespace
-// and name of the pod to remove the scheduling gate from.
-type schGateRemovalOperation struct {
-	namespace string
-	name      string
+	schGateManager *schGateManager
+	initOnce       sync.Once
 }
 
 func New() *Action {
 	return &Action{
 		enablePredicateErrorCache: true, // default to enable it
-		schGateRemovalStopCh:      make(chan struct{}),
-		gateRemovalWorkerNum:      5, // default value
+		schGateManager:            newSchGateManager(5),
 	}
 }
 
@@ -145,59 +127,19 @@ func (alloc *Action) Name() string {
 
 func (alloc *Action) Initialize() {
 	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) {
-		// Create channel with buffer size based on worker count (200 operations per worker)
-		channelSize := alloc.gateRemovalWorkerNum * gateRemovalBufferPerWorker
-		alloc.schGateRemovalCh = make(chan schGateRemovalOperation, channelSize)
-
-		// Start async gate operation workers
-		for i := 0; i < alloc.gateRemovalWorkerNum; i++ {
-			alloc.schGateRemovalWorkersWg.Add(1)
-			go alloc.schGateRemovalWorker()
-		}
-		klog.V(3).Infof("Started %d async workers for gate removal", alloc.gateRemovalWorkerNum)
-	}
-}
-
-// schGateRemovalWorker processes async gate add/remove requests
-func (alloc *Action) schGateRemovalWorker() {
-	defer alloc.schGateRemovalWorkersWg.Done()
-	for {
-		select {
-		case <-alloc.schGateRemovalStopCh:
-			klog.V(4).Infof("Scheduling gate operation worker shutting down")
-			return
-		case op := <-alloc.schGateRemovalCh:
-			// Fetch fresh pod state from API server
-			// Use cached kubeClient to avoid data race with session updates
-			pod, err := alloc.kubeClient.CoreV1().Pods(op.namespace).Get(
-				context.Background(),
-				op.name,
-				metav1.GetOptions{})
-
-			if err != nil {
-				klog.Errorf("Failed to get pod %s/%s for gate operation: %v", op.namespace, op.name, err)
-				continue
-			}
-
-			// Remove the Volcano scheduling gate
-			if err := cache.RemoveVolcanoSchGate(alloc.kubeClient, pod); err != nil {
-				klog.Errorf("Failed to remove gate from %s/%s: %v", op.namespace, op.name, err)
-			} else {
-				klog.V(3).Infof("Removed Volcano scheduling gate from pod %s/%s", op.namespace, op.name)
-			}
-		}
+		alloc.schGateManager.start()
 	}
 }
 
 func (alloc *Action) parseArguments(ssn *framework.Session) {
 	arguments := framework.GetArgOfActionFromConf(ssn.Configurations, alloc.Name())
 	arguments.GetBool(&alloc.enablePredicateErrorCache, conf.EnablePredicateErrCacheKey)
-	arguments.GetInt(&alloc.gateRemovalWorkerNum, GateRemovalWorkerNumKey)
+	arguments.GetInt(&alloc.schGateManager.workerNum, GateRemovalWorkerNumKey)
 
 	// Ensure at least 1 worker
-	if alloc.gateRemovalWorkerNum < 1 {
-		klog.Warningf("Invalid gateRemovalWorkerNum %d, using default value 5", alloc.gateRemovalWorkerNum)
-		alloc.gateRemovalWorkerNum = 5
+	if alloc.schGateManager.workerNum < 1 {
+		klog.Warningf("Invalid gateRemovalWorkerNum %d, using default value 5", alloc.schGateManager.workerNum)
+		alloc.schGateManager.workerNum = 5
 	}
 }
 
@@ -210,7 +152,7 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	// Initialize workers once with the configured number.
 	// Cache KubeClient for thread-safe access from workers.
 	alloc.initOnce.Do(func() {
-		alloc.kubeClient = ssn.KubeClient()
+		alloc.schGateManager.kubeClient = ssn.KubeClient()
 		alloc.Initialize()
 	})
 
@@ -401,7 +343,6 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
 				"allocatedHyperNode", job.AllocatedHyperNode, "subJobNum", jobWorksheet.subJobs.Len())
 			stmt := alloc.allocateForJob(job, jobWorksheet, ssn.HyperNodes[framework.ClusterTopHyperNode])
 			if stmt != nil && ssn.JobReady(job) { // do not commit stmt when job is pipelined
-				ssn.CleanupReservations(stmt)
 				stmt.Commit()
 				ssn.MarkJobDirty(job.UID)
 				alloc.recorder.UpdateDecisionToJob(job, ssn.HyperNodes)
@@ -418,7 +359,6 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
 				klog.V(3).InfoS("Try to allocate resource", "queue", queue.Name, "job", job.UID, "taskNum", tasks.Len())
 				stmt := alloc.allocateResourcesForTasks(subJob, tasks, framework.ClusterTopHyperNode)
 				if stmt != nil && ssn.JobReady(job) { // do not commit stmt when job is pipelined
-					ssn.CleanupReservations(stmt)
 					stmt.Commit()
 
 					// There are still left tasks that need to be allocated when min available < replicas, put the job back
@@ -435,27 +375,6 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
 		// Put back the queue to priority queue after job's resource allocating finished,
 		// To ensure that the priority of the queue is calculated based on the latest resource allocation situation.
 		queues.Push(queue)
-	}
-}
-
-// schedulingGateRemoval queues async gate removal if scheduling failed.
-// This ensures cluster autoscalers can see the Unschedulable condition and trigger scale-up
-func (alloc *Action) schedulingGateRemoval(task *api.TaskInfo, queueID api.QueueID) {
-	// Only enqueue gate removal if the task has only Volcano scheduling gate
-	if api.HasOnlyVolcanoSchedulingGate(task.Pod) {
-		op := schGateRemovalOperation{
-			namespace: task.Namespace,
-			name:      task.Name,
-		}
-
-		select {
-		case alloc.schGateRemovalCh <- op:
-			klog.V(3).Infof("Queued gate removal for %s/%s", task.Namespace, task.Name)
-			// Update task state immediately so it won't be queued again in this cycle
-			task.SchGated = false
-		default:
-			klog.Warningf("Gate operation queue full, skipping gate removal for %s/%s", task.Namespace, task.Name)
-		}
 	}
 }
 
@@ -691,21 +610,19 @@ func (alloc *Action) allocateResourcesForTasks(subJob *api.SubJobInfo, tasks *ut
 		}
 
 		// If task passed allocation check and has the QueueAllocationGate, initiate async gate removal.
-		// Gate will be removed by the background worker (best effort). During the bind operation, we need
-		// to ensure the gate is not present, otherwise the bind will fail.
+		// Gate will be removed by the background worker (best effort).
 		if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) &&
 			task.SchGated && api.HasQueueAllocationGateAnnotation(task.Pod) {
 			klog.V(3).Infof("Task %s/%s has the QueueAllocationGate, queue async gate removal", task.Namespace, task.Name)
-			alloc.schedulingGateRemoval(task, queue.UID)
+			alloc.schGateManager.enqueue(task)
 		}
 
-		// Skip tasks with external (non-Volcano) scheduling gates
+		// Skip gated tasks. If someone added the Volcano gate without the opt-in annotation,
+		// warn them since the gate will never be removed automatically.
 		if task.SchGated {
-			if api.HasOnlyVolcanoSchedulingGate(task.Pod) {
-				klog.Warningf("Task %s/%s has Volcano scheduling gate but missing annotation %q, gate will not be removed automatically; add the annotation or remove the gate manually",
+			if api.HasOnlyVolcanoSchedulingGate(task.Pod) && !api.HasQueueAllocationGateAnnotation(task.Pod) {
+				klog.Warningf("Task %s/%s has Volcano scheduling gate but missing the opt-in annotation %q; gate will not be removed automatically",
 					task.Namespace, task.Name, schedulingv1beta1.QueueAllocationGateKey)
-			} else {
-				klog.V(4).Infof("Task %s/%s has non-Volcano scheduling gate, skipping", task.Namespace, task.Name)
 			}
 			continue
 		}
@@ -939,15 +856,5 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
 }
 
 func (alloc *Action) UnInitialize() {
-	alloc.shutdownOnce.Do(func() {
-		// Signal workers to shutdown
-		close(alloc.schGateRemovalStopCh)
-		// Wait for all workers to finish
-		alloc.schGateRemovalWorkersWg.Wait()
-		// Close the channel (only if Initialize was ever called)
-		if alloc.schGateRemovalCh != nil {
-			close(alloc.schGateRemovalCh)
-		}
-		klog.V(3).Infof("Async gate removal workers shut down")
-	})
+	alloc.schGateManager.stop()
 }
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"math"
 	"slices"
-	"sync"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -37,14 +36,6 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/util"
 	commonutil "volcano.sh/volcano/pkg/util"
-)
-
-const (
-	// GateRemovalWorkerNumKey is the configuration key for number of async gate removal workers
-	GateRemovalWorkerNumKey = "gateRemovalWorkerNum"
-
-	// Default buffer size per worker for the gate removal channel
-	gateRemovalBufferPerWorker = 200
 )
 
 type allocateContext struct {
@@ -109,15 +100,11 @@ type Action struct {
 	enablePredicateErrorCache bool
 
 	recorder *Recorder
-
-	schGateManager *schGateManager
-	initOnce       sync.Once
 }
 
 func New() *Action {
 	return &Action{
 		enablePredicateErrorCache: true, // default to enable it
-		schGateManager:            newSchGateManager(5),
 	}
 }
 
@@ -125,22 +112,11 @@ func (alloc *Action) Name() string {
 	return "allocate"
 }
 
-func (alloc *Action) Initialize() {
-	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) {
-		alloc.schGateManager.start()
-	}
-}
+func (alloc *Action) Initialize() {}
 
 func (alloc *Action) parseArguments(ssn *framework.Session) {
 	arguments := framework.GetArgOfActionFromConf(ssn.Configurations, alloc.Name())
 	arguments.GetBool(&alloc.enablePredicateErrorCache, conf.EnablePredicateErrCacheKey)
-	arguments.GetInt(&alloc.schGateManager.workerNum, GateRemovalWorkerNumKey)
-
-	// Ensure at least 1 worker
-	if alloc.schGateManager.workerNum < 1 {
-		klog.Warningf("Invalid gateRemovalWorkerNum %d, using default value 5", alloc.schGateManager.workerNum)
-		alloc.schGateManager.workerNum = 5
-	}
 }
 
 func (alloc *Action) Execute(ssn *framework.Session) {
@@ -148,13 +124,6 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 	defer klog.V(5).Infof("Leaving Allocate ...")
 
 	alloc.parseArguments(ssn)
-
-	// Initialize workers once with the configured number.
-	// Cache KubeClient for thread-safe access from workers.
-	alloc.initOnce.Do(func() {
-		alloc.schGateManager.kubeClient = ssn.KubeClient()
-		alloc.Initialize()
-	})
 
 	// the allocation for pod may have many stages
 	// 1. pick a queue named Q (using ssn.QueueOrderFn)
@@ -614,7 +583,7 @@ func (alloc *Action) allocateResourcesForTasks(subJob *api.SubJobInfo, tasks *ut
 		if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) &&
 			task.SchGated && api.HasQueueAllocationGateAnnotation(task.Pod) {
 			klog.V(3).Infof("Task %s/%s has the QueueAllocationGate, queue async gate removal", task.Namespace, task.Name)
-			alloc.schGateManager.enqueue(task)
+			ssn.SchGateManager().Enqueue(task)
 		}
 
 		// Skip gated tasks. If someone added the Volcano gate without the opt-in annotation,
@@ -676,7 +645,6 @@ func (alloc *Action) allocateResourcesForTasks(subJob *api.SubJobInfo, tasks *ut
 				fitErrors.SetHyperNode(hyperNode)
 			}
 			job.NodesFitErrors[task.UID] = fitErrors
-
 			// Assume that all left tasks are allocatable, but can not meet gang-scheduling min member,
 			// so we should break from continuously allocating.
 			// otherwise, should continue to find other allocatable task
@@ -855,6 +823,4 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
 	return alloc.session.PredicateForAllocateAction(task, node)
 }
 
-func (alloc *Action) UnInitialize() {
-	alloc.schGateManager.stop()
-}
+func (alloc *Action) UnInitialize() {}
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -17,22 +17,38 @@
 package allocate
 
 import (
+	"context"
 	"fmt"
 	"math"
 	"slices"
+	"sync"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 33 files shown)


JesseStutler:
I still have some opinions on the proposal and please take a look @devzizu https://github.com/volcano-sh/volcano/pull/4727, thanks, we may need to furthur discuss it on the community meeting

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -122,6 +201,14 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 	alloc.parseArguments(ssn)
 
+	// Initialize workers once with the configured number
+	// Cache KubeClient for thread-safe access from workers
+	if !alloc.startedWorkers {
+		alloc.startedWorkers = true
+		alloc.kubeClient = ssn.KubeClient()
+		alloc.Initialize()
+	}
```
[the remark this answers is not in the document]
devzizu: This may not be strictly necessary: `Execute` is only called from the single scheduling loop (`runOnce`), so there's no concurrent access and the previous `startedWorkers` check was safe. I still switched to `sync.Once` to make the “run exactly once” intent explicit, to align with other one-time inits in the codebase (*e.g.*, volumeBindingPluginOnce, scheduler config), and to keep the code safe if we ever run actions from multiple goroutines later.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -342,6 +434,27 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
...
+		}
+
+		select {
+		case alloc.schGateRemovalCh <- op:
+			klog.V(3).Infof("Queued gate removal for %s/%s", task.Namespace, task.Name)
+			// Update task state immediately so it won't be queued again in this cycle
+			task.SchGated = false
+		default:
+			klog.Warningf("Gate operation queue full, skipping gate removal for %s/%s", task.Namespace, task.Name)
+		}
```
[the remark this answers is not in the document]
devzizu: I'd rather not block the scheduler. If the channel is full we skip this cycle; the task stays SchGated in cache so we'll try again next cycle. That gives us retries without delaying the scheduling loop when workers are slow. The channel is large (`e.g.`, 1000 with 5 workers), so the queue being full should be rare.

test/e2e/util/pod.go
```diff
@@ -75,3 +77,24 @@ func DeletePod(ctx *TestContext, pod *v1.Pod) {
...
+	Expect(err).NotTo(HaveOccurred(), "failed to get pod %s", podName)
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == v1.PodScheduled &&
+			cond.Status == v1.ConditionFalse &&
+			cond.Reason == "SchedulingGated" {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
devzizu: Prefer to keep PodHasOnlyVolcanoSchedulingGate  as is so that it can be imported in tests. Also removed HasSchedulingGatedCondition since it is dead code, not used.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -342,6 +434,27 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
...
+		op := schGateRemovalOperation{
+			namespace: task.Namespace,
+			name:      task.Name,
+		}
+
+		select {
+		case alloc.schGateRemovalCh <- op:
+			klog.V(3).Infof("Queued gate removal for %s/%s", task.Namespace, task.Name)
+			// Update task state immediately so it won't be queued again in this cycle
+			task.SchGated = false
```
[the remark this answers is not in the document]
devzizu: When we set `task.SchGated = false` in the allocate action, we're only updating the session-local clone of the task; no other goroutine has a reference to it, and the cache's copy is unchanged. So there's no shared mutable task state or data race. The snapshot/clone behavior is in `cache.go` (`Snapshot`, `cloneJob`, `value.Clone()`) and `job_info.go` (`JobInfo.Clone()` and `TaskInfo.Clone()`).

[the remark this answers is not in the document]
hajnalmt: This is fine, no race here.

pkg/scheduler/actions/allocate/allocate.go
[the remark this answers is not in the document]
hajnalmt: This buffer is more than generous. I doubt there will be anybody to hit this, and you made it configurable a warning is enough.

test/e2e/schedulingaction/allocate.go
```diff
@@ -428,4 +428,133 @@ var _ = ginkgo.Describe("Job E2E Test", func() {
 		err := e2eutil.WaitJobReady(ctx, createdJob)
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	})
+
+	ginkgo.It("Unschedulable pod with removed gate reserves queue capacity and blocks other pods", func() {
```
hajnalmt: This is a really good test! it's so cool that it covers the webhook injection, the gate removal, the capacity reservation and reservation cleanup even in one go.

What we should add maybe as a test case too is that we don't remove other scheduling gates accidentally, like a pod with multiple scheduling gates one is the volcano one with annotation, it shouldn't schedule.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -573,6 +686,21 @@ func (alloc *Action) allocateResourcesForTasks(subJob *api.SubJobInfo, tasks *ut
...
+		// If task passed allocation check and has the QueueAllocationGate, initiate async gate removal.
+		// Gate will be removed by the background worker (best effort). During the bind operation, we need
+		// to ensure the gate is not present, otherwise the bind will fail.
+		if task.SchGated && api.HasQueueAllocationGateAnnotation(task.Pod) {
+			klog.V(3).Infof("Task %s/%s has the QueueAllocationGate, queue async gate removal", task.Namespace, task.Name)
+			alloc.schedulingGateRemoval(task, queue.UID)
+		}
+
+		// Skip tasks with external (non-Volcano) scheduling gates
+		if task.SchGated {
```
hajnalmt, line 591: There is an interesting path here. If someone manually adds the volcano scheduling gate without the annotation, because he misuses the feature. We will log a warning here because it's an external scheduling gate, we check only the annotation previously (no schedulingGateRemoval triggered). In this scenario the pod is actually stuck in Pending forever.

Maybe we can add a log line here and a separate branch:
```
// Skip tasks with external (non-Volcano) scheduling gates
if task.SchGated {
    if api.HasOnlyVolcanoSchedulingGate(task.Pod) {
        klog.Warningf("Task %s/%s has Volcano scheduling gate but missing annotation %q, gate will not be removed automatically; add the annotation or remove the gate manually",
            task.Namespace, task.Name, schedulingv1beta1.QueueAllocationGateKey)
    } else {
        klog.V(4).Infof("Task %s/%s has non-Volcano scheduling gate, skipping", task.Namespace, task.Name)
    }
    continue
}
```
This would show the users that they are misusing the feature.

hajnalmt (commented):
This implementation is more than solid thank you @devzizu ! When Jesse comes back we can schedule a meeting.

Please squash the commits and let's keep the design doc in the separate PR!

hajnalmt (commented):
/ok-to-test

This is quite well written. I don't see a reason not to test it. Let's see.

pkg/scheduler/actions/allocate/allocate.go
devzizu: Makes sense to me, added a warning log like you proposed

test/e2e/schedulingaction/allocate.go
devzizu: Added a new test case, can you please see if it fits the purpose? One part I'm not entirely happy with is the check that both gates remain and the pod stays Pending. Right now I'm using `gomega.Consistently` over 10 seconds (polling every second) to assert that the pod still has both gates and is Pending. It works, but it feels a bit arbitrary (fixed window and interval) and doesn't give a clear, deterministic guarantee that the scheduler has run and decided to skip the task.

pkg/webhooks/admission/pods/mutate/mutate_pod.go
```diff
@@ -143,6 +151,34 @@ func createPatch(pod *v1.Pod) ([]byte, error) {
...
+			pod.Namespace, pod.Name)
+		return nil
+	}
+
+	gates := []v1.PodSchedulingGate{
+		{Name: schedulingv1beta1.QueueAllocationGateKey},
+	}
+	// Use "add" when the field is not set, and "replace" when it already exists.
+	// In both cases, preserve existing gates and append the new one.
+	value := append(pod.Spec.SchedulingGates, gates...)
```
hajnalmt: append is not idempotent here (this code can add the same volcano scheduling gate again) and replace is really risky below. If an other component is modifying the gates too paralelly we can run into issues (replaced the gates but other component added one before us (informer not updated the controller yet) paff we removed that gate basically).

We should use the "-" JSON Patch operator to append properly.
https://github.com/volcano-sh/volcano/pull/4727/changes#r2589013696 

The real problem with append is that it won't filter duplications. If the pod is modified and the mutation is called again on the pod we will just append the same gate to the array.

I would be happier to have here something like this instead:
```
	// Check if opt-in annotation is present
	if !api.HasQueueAllocationGateAnnotation(pod) {
		klog.V(4).Infof("Pod %s/%s does not have opt-in annotation, skipping gate",
			pod.Namespace, pod.Name)
		return nil
	}

	gate := v1.PodSchedulingGate{Name: schedulingv1beta1.QueueAllocationGateKey}

	// Idempotent: do not add duplicate Volcano gate.
	for _, g := range pod.Spec.SchedulingGates {
		if g.Name == gate.Name {
			return nil
		}
	}

	// Parent missing
	if pod.Spec.SchedulingGates == nil  {
		return &patchOperation{
			Op:    "add",
			Path:  "/spec/schedulingGates",
			Value: []v1.PodSchedulingGate{gate},
		}
	}

	// Parent exists: append safely
	return &patchOperation{
		Op:    "add",
		Path:  "/spec/schedulingGates/-",
		Value: gate,
	}
```

The `if` branch is needed because the [Scheduling Gates are optionel](https://github.com/kubernetes/api/blob/release-1.35/core/v1/types.go#L4396) so it can be nil and add with "/-" won't work if the parent is `nil` but will work on an empty list.

You can actually try your code out:
https://json8.github.io/patch/demos/apply/
Input:
```
{
  "spec": {}
}
```
Patch:
```
[
  { "op": "add", "path": "/spec/schedulingGates", "value": [ { "name": "volcano.sh/queue-allocation-gate" } ] }
]
```
Output/Next rounds Input:
```
{
  "spec": {
    "schedulingGates": [
      {
        "name": "volcano.sh/queue-allocation-gate"
      }
    ]
  }
}
```
This is the Input on the next mutation but the patch is:
```
[
  { "op": "replace", "path": "/spec/schedulingGates", "value": [ { "name": "volcano.sh/queue-allocation-gate" } ,{ "name": "volcano.sh/queue-allocation-gate" }] }
]
```
The same gate twice.

hajnalmt: Okay I realized that schedulingGates are added only at pod creation time so duplication won't be a problem. But I would still go with this as the race is still valid since there is no guaranteed ordering for the webhooks to be called so if someone modifies the gates in a webhook like us. We are in a bad shape. This is just a lot safer.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -824,4 +961,16 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
 	return alloc.session.PredicateForAllocateAction(task, node)
 }
 
-func (alloc *Action) UnInitialize() {}
+func (alloc *Action) UnInitialize() {
```
hajnalmt: Nor Initialize() nor UnInitialize() is used in our codebase except the unit test helper:

https://github.com/volcano-sh/volcano/blob/49b8f40ba6247b9a6a65e14e6e14c4fb19feb77c/pkg/scheduler/uthelper/helper.go#L224-L229

What a bummer. Can you create at least a follow up ticket for this please? Currently nothing will close these channels. If this gets merged in we should fix this, so this is not just dead code here.

hajnalmt: I also think that this is out of the scope of this feature.

pkg/scheduler/cache/util.go
```diff
@@ -132,3 +137,35 @@ func getHyperNodeEventSource(source string) []string {
 	}
 	return parts
 }
+
+// RemoveVolcanoSchGate removes the Volcano scheduling gate from a pod via strategic merge patch.
+// Returns nil if gate is successfully removed or already removed (idempotent).
+func RemoveVolcanoSchGate(kubeClient kubernetes.Interface, pod *v1.Pod) error {
```
hajnalmt: Can we introduce a unit test for this function? A test that directly calls it?

I know there is no `util_test.go` now, but probably we need it to have this verified. I ran into issues where the strategic merge didn't replace and kept the old values, but it would probably come up in the e2e test if this wouldn't work. Nevertheless let's have this to verify that this function works properly.

hajnalmt (commented):
I gave an other round what do you think about my comments?

pkg/webhooks/admission/pods/mutate/mutate_pod.go
devzizu: Yes, you are correct! Will also fix the proposal, thank you for noticing this!

Fix in implementation: https://github.com/volcano-sh/volcano/pull/5033/commits/573980bb66edca2eea18a3d64ec96ed169b6ba52
Fix in the proposal: https://github.com/volcano-sh/volcano/pull/4727/changes/b467a0eb6e0311652060eeb7712e32093cecff4f

pkg/scheduler/cache/util.go
devzizu: Indeed, just added a new test and created the new `util_test.go` as suggested. Please take a look 🙏 

https://github.com/volcano-sh/volcano/pull/5033/commits/a7286f609c8d815ea91fab6165a0e6c7980b75f3

pkg/scheduler/actions/allocate/allocate.go
devzizu: Oh, this is a surprise for me -- I was counting on it to be called, but indeed I only see it as a test helper 😄

I actually call it here:

https://github.com/devzizu/volcano/blob/1d6ae98c6c3c32c122ad57586a88125f8ee4655a/pkg/scheduler/actions/allocate/allocate.go#L211-L214

P.S. Perhaps we can add this to https://github.com/volcano-sh/volcano/issues/5048?

devzizu: Created https://github.com/volcano-sh/volcano/issues/5052 to followup 👍

pkg/scheduler/cache/util.go
hajnalmt: Cool thank you!

pkg/webhooks/admission/pods/mutate/mutate_pod.go
```diff
@@ -143,6 +151,47 @@ func createPatch(pod *v1.Pod) ([]byte, error) {
...
+			Op:    "add",
+			Path:  "/spec/schedulingGates",
+			Value: []v1.PodSchedulingGate{gate},
+		}
+	}
+
+	// Parent exists: We can safely append to the existing array.
+	// Using the "-" path operator tells JSON Patch to append to the end of the array,
+	// preventing us from overwriting gates added by parallel webhooks.
+	return &patchOperation{
```
hajnalmt, line 195: This is the way thanks!

pkg/scheduler/actions/allocate/allocate.go
hajnalmt: Thanks! We can merge this I think without solving that I think.

hajnalmt (approved):
LGTM Thanks. Let's align with  @JesseStutler 👍

installer/helm/chart/volcano/templates/admission.yaml
```diff
@@ -145,6 +145,9 @@ spec:
             {{- if $scheduler_name }}
             - --scheduler-name={{- $scheduler_name }}
             {{- end }}
+            {{- if .Values.custom.scheduler_feature_gates }}
+            - --feature-gates={{ .Values.custom.scheduler_feature_gates }}
```
JesseStutler: Kind of weird that the admission controller uses the same features gates with the scheduler, will it cause admission controller fail to start up if there are unknown features gates in admission controller?

JesseStutler:
Could you also add a user guide to teach users how to use this feature? Thanks! @devzizu

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -97,23 +113,92 @@ type Action struct {
 	enablePredicateErrorCache bool
 
 	recorder *Recorder
+
+	// Async gate management infrastructure
```
JesseStutler: I think we could further extract this field into a struct, and then add comments to explain this feature.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -97,23 +113,92 @@ type Action struct {
...
+	defer alloc.schGateRemovalWorkersWg.Done()
+	for {
+		select {
+		case <-alloc.schGateRemovalStopCh:
+			klog.V(4).Infof("Scheduling gate operation worker shutting down")
+			return
+		case op := <-alloc.schGateRemovalCh:
+			// Fetch fresh pod state from API server
+			// Use cached kubeClient to avoid data race with session updates
+			pod, err := alloc.kubeClient.CoreV1().Pods(op.namespace).Get(
```
JesseStutler: I have a question that why don't we get the pods from local indexer (in the informer)? Should we must to get the fresh pod from api server first?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -122,6 +207,13 @@ func (alloc *Action) Execute(ssn *framework.Session) {
 
 	alloc.parseArguments(ssn)
 
+	// Initialize workers once with the configured number.
+	// Cache KubeClient for thread-safe access from workers.
+	alloc.initOnce.Do(func() {
+		alloc.kubeClient = ssn.KubeClient()
+		alloc.Initialize()
```
JesseStutler, line 156: I think could we add Initialize somewhere else? It's weird to add it in Execute, maybe like right here is better: https://github.com/volcano-sh/volcano/blob/d3721cbfe367b35d49317ced1d64a78db89d9ed9/pkg/scheduler/scheduler.go#L130-L133

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -573,6 +691,26 @@ func (alloc *Action) allocateResourcesForTasks(subJob *api.SubJobInfo, tasks *ut
...
+		// If task passed allocation check and has the QueueAllocationGate, initiate async gate removal.
+		// Gate will be removed by the background worker (best effort). During the bind operation, we need
+		// to ensure the gate is not present, otherwise the bind will fail.
+		if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) &&
+			task.SchGated && api.HasQueueAllocationGateAnnotation(task.Pod) {
+			klog.V(3).Infof("Task %s/%s has the QueueAllocationGate, queue async gate removal", task.Namespace, task.Name)
+			alloc.schedulingGateRemoval(task, queue.UID)
+		}
+
+		// Skip tasks with external (non-Volcano) scheduling gates
```
JesseStutler: `organizeJobWorksheet` already skipped the scheduling part of the pods containing external scheduling gates, so is it necessary to verify it again here?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -342,6 +439,27 @@ func (alloc *Action) allocateResources(actx *allocateContext) {
...
+		op := schGateRemovalOperation{
+			namespace: task.Namespace,
+			name:      task.Name,
+		}
+
+		select {
+		case alloc.schGateRemovalCh <- op:
+			klog.V(3).Infof("Queued gate removal for %s/%s", task.Namespace, task.Name)
+			// Update task state immediately so it won't be queued again in this cycle
+			task.SchGated = false
```
JesseStutler: It only send the op to the channel and immediately update the SchGated flag to false, but what if gate removal failed? Do you have place to reset it back? I think it's better to set the flag to false when we successfully already removed the gate

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -824,4 +963,16 @@ func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
 	return alloc.session.PredicateForAllocateAction(task, node)
 }
 
-func (alloc *Action) UnInitialize() {}
+func (alloc *Action) UnInitialize() {
```
JesseStutler, line 858: Looks like currently we didn't execute UnInitialize() in other places?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -933,15 +965,80 @@ func (cp *capacityPlugin) isLeafQueue(queueID api.QueueID) bool {
...
+		}
+	}
+}
+
+// rebuildReservedCache clears and rebuilds the reserved tasks cache for this scheduling cycle.
+// It scans all pending tasks and adds those that have passed capacity checks in previous cycles
+// but are not yet allocated. These are identified by:
+// - NO queue allocation scheduling gate (gate was removed after passing capacity)
+// - HAS queue allocation scheduling gate annotation (proof they opted-in and passed capacity check)
+func (cp *capacityPlugin) buildQueueReservedTasksCache(ssn *framework.Session) {
```
JesseStutler, line 1003: I think proportion plugin also need it

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -93,6 +100,22 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 
 	klog.V(4).Infof("The total resource is <%v>", cp.totalResource)
 
+	// Rebuild reserved cache for this scheduling cycle
+	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) {
+		cp.buildQueueReservedTasksCache(ssn)
+
+		// Register cleanup function for successful allocations
+		ssn.AddCleanupReservationsFn(cp.Name(), func(obj interface{}) {
```
JesseStutler: Just want to ensure that `CleanupReservationsFn` wants to move reserved resources to allocated resource to prevent double couting if the task needs to be allocated, right? Could we add it do it in `AllocateFunc/DeAllocateFunc` and don't need to add a new extension point?

installer/helm/chart/volcano/templates/admission.yaml
devzizu: I don't think it would fail, however I agree that having the same Helm value name may be misleading. What do you suggest? I see that agent-scheduler follows the same pattern, i.e., it registers all gates, only checks the ones it cares about (e.g., PodDisruptionBudgetsSupport), and silently ignores the rest.

pkg/scheduler/actions/allocate/allocate.go
devzizu: Actually, we don't need to get the pod at all, good catch. `RemoveVolcanoSchGate` only uses `namespace` and `name` to build a strategic merge patch.

Ended up removing the API server Get() entirely and changed the signature to `RemoveVolcanoSchGate(kubeClient, namespace, name)`, let me know what you think.

https://github.com/volcano-sh/volcano/pull/5033/commits/00de2994e7e2670df58f611a18bb4c54e20d5df1

pkg/scheduler/actions/allocate/allocate.go
devzizu: That's a good suggestion, I extracted the fields into a new `schGateManager` struct and added comments there. Changes were introduced in this commit https://github.com/volcano-sh/volcano/pull/5033/commits/7c08f94845ecddcdcc70a3060d88d410edb759a9.

pkg/scheduler/actions/allocate/allocate.go
devzizu: The fact that initalize and uninitialize are not being called was already reported by me here https://github.com/volcano-sh/volcano/issues/5052, and there is ongoing work here https://github.com/volcano-sh/volcano/pull/5069. I would suggest we move forward with https://github.com/volcano-sh/volcano/pull/5069 first, and then I can adopt it. WDYT?

pkg/scheduler/actions/allocate/allocate.go
devzizu: Yes, this was never called, in any scheduler action -- please take a look at my comment here https://github.com/volcano-sh/volcano/pull/5033#discussion_r3069638060.

pkg/scheduler/actions/allocate/allocate.go
devzizu: Fixed, nice catch!

installer/helm/chart/volcano/templates/admission.yaml
JesseStutler: I think webhook can have seperate feature gates helm setting

pkg/scheduler/actions/allocate/allocate.go
JesseStutler: @devzizu Could you also seperate these sturct and related codes into different files but under allocate dir? Like to name `schedulingate.go` or something, it will be better to let `allocate.go` be more clear, thanks :D

pkg/scheduler/actions/allocate/allocate.go
JesseStutler: I think now it's better, thanks

pkg/scheduler/plugins/capacity/capacity.go
JesseStutler: @devzizu Please cc, I think `AllocateFunc/DeAllocateFunc` may can already solve your scenario?

JesseStutler:
@devzizu Please also resolve the conflicts, thanks!

pkg/scheduler/plugins/capacity/capacity.go
JesseStutler: Maybe if time is limit, we can open an issue and implement it later

installer/helm/chart/volcano/templates/admission.yaml
devzizu: LGTM, fixed! https://github.com/volcano-sh/volcano/pull/5033/changes/1a2ac07604093e58e44c88d139e46fd40944ba82

pkg/scheduler/actions/allocate/allocate.go
devzizu: Yes, fixed https://github.com/volcano-sh/volcano/pull/5033/commits/a9d1c1b87a595497807731150812e0cb477a6c9e

pkg/scheduler/plugins/capacity/capacity.go
devzizu: I think `AllocateFunc/DeAllocateFunc` can in fact resolve this scenario, and e2e tests also pass. Removed `CleanupReservationsFn` entirely and moved the logic into the existing event handlers:

- `AllocateFunc`: removes the task from the reserved cache when it gets allocated.
- `DeallocateFunc`: restores the task to the reserved cache on rollback.

Changes introduced in https://github.com/volcano-sh/volcano/pull/5033/commits/668c2d0b423f2782d12b1d744ec79634f91a8730.

WDYT @JesseStutler?

pkg/scheduler/plugins/capacity/capacity.go
devzizu: As aligned in slack, I'll open a new issue with this missing part once this one is integrated. I believe the changes in proportion are similar to this one. Thank you @JesseStutler from pointing this out.

devzizu:
@JesseStutler correct me if I'm wrong, but I believe the only test failing is a flaky one https://github.com/volcano-sh/volcano/issues/5015. Other than that, I've rebased again with the latest commit on the main branch.

pkg/scheduler/actions/allocate/allocate.go
JesseStutler: @devzizu #5069 I think it might not be possible to solve this in the short term. Do we have other solutions, such as changing the gateManager to the scheduler's resident, not initialize in the action by using sync.Once?

pkg/scheduler/framework/session_plugins.go
```diff
@@ -366,6 +366,7 @@ func (ssn *Session) Allocatable(queue *api.QueueInfo, candidate *api.TaskInfo) b
 	return true
 }
 
+// CleanupReservations invokes all registered reservation cleanup functions
```
JesseStutler: Looks like there are still some remaining comments that forgot to clean @devzizu

pkg/scheduler/cache/util.go
```diff
@@ -132,3 +137,35 @@ func getHyperNodeEventSource(source string) []string {
 	}
 	return parts
 }
+
+// RemoveVolcanoSchGate removes the Volcano scheduling gate from a pod by namespace and name.
+// Returns nil if gate is successfully removed or already removed (idempotent).
+func RemoveVolcanoSchGate(kubeClient kubernetes.Interface, namespace, name string) error {
```
JesseStutler, line 143: I remember that previously when the scheduler wants to bind, it will also call `RemoveVolcanoSchGate` to ensure that the gate was already removed. Current codes don't need it?

JesseStutler (commented):
I'm satisfied with current implementation! Just still some new comments needs to take a look @devzizu :D

pkg/scheduler/cache/util.go
devzizu: Correct, I removed it. Previously the synchronous gate removal in `Bind()` was needed **as a safety net** because gated tasks could be allocated in the same cycle (we were setting `task.SchGated = false` **immediately** after enqueuing the async removal). 

Now gated tasks are always skipped after enqueuing and never reach predicate checks, allocation, or even bind. The task only enters the allocation path in a future cycle after the async worker has actually removed the gate and the informer reflects it. So by the time a task reaches `Bind()`, the gate is guaranteed to be already gone.

Please recall your comment here https://github.com/volcano-sh/volcano/pull/5033#discussion_r3042690424, and this commit https://github.com/volcano-sh/volcano/pull/5033/commits/3f7402495c72a314597b8b9551e07fc721c55234 and this [code line](https://github.com/volcano-sh/volcano/pull/5033/changes#diff-3dcbd15d9e1bf118608b4eedcaadc5cc7bdb9f60b482ad0ca7238ada5ad9e32cR622.).

JesseStutler: I didn't get it, if the gate is not removed yet and enter the binding cycle, the task will be resynced again, is it right? How to promise that `the gate is guaranteed to be already gone`?

devzizu: In allocateResourcesForTasks, we are [skipping task.schGated](https://github.com/volcano-sh/volcano/pull/5033/changes#diff-3dcbd15d9e1bf118608b4eedcaadc5cc7bdb9f60b482ad0ca7238ada5ad9e32cR622) tasks already, and this flag will only be updated if the gate is eventually removed in further cycles. So by the time a task reaches bind, the informer has already confirmed the gate is gone, right?

pkg/scheduler/framework/session_plugins.go
devzizu: Ah, sorry, you're right -- fixed https://github.com/volcano-sh/volcano/pull/5033/commits/bc725b900487c07a8a6a066b2af2913105f59412.

pkg/scheduler/actions/allocate/allocate.go
devzizu: Yes, that's a possibility! I've tried moving `SchGateManager` to a new `pkg/scheduler/gate/` (lmk what do you think about this proposal) and made it scheduler-resident. Now it is created in `Scheduler.Run()`, started there, and stopped on shutdown.

The manager is passed to each scheduling session via `OpenSession()` and accessed by the allocate action through `ssn.SchGateManager()`. The Action struct now keeps the existing behavior with empty `Initialize()/UnInitialize()` stubs.

WDYT? @JesseStutler 

Changes introduced in https://github.com/volcano-sh/volcano/pull/5033/commits/a7629b17ec25450fcbcbc1c7b9b4ec4a301cfa0e.

pkg/scheduler/cache/util.go
JesseStutler: I got it, thanks! It looks clearer now. 👍

JesseStutler:
/approve
Thanks @devzizu ! A great job! Wish you could contribute more features in the future :D
Also help /cc @hajnalmt if you agree with feature, you could add a lgtm label to help merge it

pkg/webhooks/admission/pods/mutate/mutate_pod.go
```diff
@@ -110,6 +114,12 @@ func createPatch(pod *v1.Pod) ([]byte, error) {
 	config.ConfigData.Lock()
 	defer config.ConfigData.Unlock()
 
+	// Add scheduling gates if opted-in
+	patchGates := patchSchedulingGates(pod)
```
hajnalmt, line 118: nit: What came to my mind here is that if multiple schedulers are present then an annotated pod may receive the Volcano scheduling gate even if another scheduler (e.g. kube-scheduler) owns the pod. I consider this as an edge case, and this would require planting schedulingNames in the admission webhook, so I wouldn't bother with it implementation-wise.

Although on the website documentation maybe this constraint can be mentioned: use the annotation only for Volcano-managed pods, and keep feature-gate/configuration consistent.

pkg/scheduler/gate/schedulinggate.go
```diff
@@ -0,0 +1,129 @@
...
+	bufferPerWorker = 200
+)
+
+// SchGateManager handles asynchronous removal of Volcano scheduling gates from pods.
+// When the SchedulingGatesQueueAdmission feature is enabled, pods opt in via the
+// scheduling.volcano.sh/queue-allocation-gate annotation. The webhook injects a scheduling
+// gate at creation time, and this manager removes it asynchronously after the queue capacity
+// check passes, allowing cluster autoscalers to see the Unschedulable condition only when
+// it reflects genuine cluster resource shortage rather than queue limits.
+type SchGateManager struct {
```
hajnalmt, line 43: What a nice abstraction!

pkg/scheduler/scheduler.go
```diff
@@ -90,6 +96,17 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
 // initializes the cache, and begins the scheduling process.
 func (pc *Scheduler) Run(stopCh <-chan struct{}) {
 	pc.loadSchedulerConf()
+
+	// Start the gate manager (if the feature gate is enabled).
```
hajnalmt, line 100: This is a lot cleaner here!

hajnalmt (approved):
/lgtm

Awesome work! Thank you so much for it @devzizu!
Let's open the issue on the proportion plugin about the extension of this feature there and we shall document this feature on the website too. They can maybe be good first issues even.

devzizu:
Thank you @hajnalmt @JesseStutler! As soon as possible I'll open the two new issues, one regarding the proportion plugin and the other for the website documentation.

PR #5043, opened 2026-02-13
Cleanup unnecessary error return value

hzxuzhonghu:
This can help reduce the complexity of rollingback in statement operations 

triggered by  #5010

cc @hajnalmt

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -650,7 +650,6 @@ func (ji *JobInfo) UpdateTaskStatus(task *TaskInfo, status TaskStatus) {
 	// Update task's status to the target status once task addition is guaranteed to succeed.
 	task.Status = status
 	ji.AddTaskInfo(task)
-
 }
 
 func (ji *JobInfo) deleteTaskIndex(ti *TaskInfo) {
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -494,11 +494,7 @@ func (pmpt *Action) topologyAwarePreempt(
 	// after the entire preemption attempt (evictions + pipeline) succeeds.
 	tmpStmt := framework.NewStatement(ssn)
 
-	if status := prepareCandidate(bestCandidate, preemptor.Pod, tmpStmt, ssn); !status.IsSuccess() {
-		tmpStmt.Discard()
-		return false, fmt.Errorf("failed to prepare candidate: %v", status)
-	}
-
+	prepareCandidate(bestCandidate, preemptor.Pod, tmpStmt)
 	if err := tmpStmt.Pipeline(preemptor, bestCandidate.Name(), true); err != nil {
 		klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>",
 			preemptor.Namespace, preemptor.Name, bestCandidate.Name())
@@ -536,7 +532,7 @@ func (pmpt *Action) findCandidates(preemptor *api.TaskInfo, filter func(*api.Tas
 }
 
 // prepareCandidate evicts the victim pods before nominating the selected candidate
-func prepareCandidate(c *candidate, pod *v1.Pod, stmt *framework.Statement, ssn *framework.Session) *api.Status {
+func prepareCandidate(c *candidate, pod *v1.Pod, stmt *framework.Statement) {
 	for _, victim := range c.Victims() {
 		klog.V(3).Infof("Try to preempt Task <%s/%s> for Task <%s/%s>",
 			victim.Namespace, victim.Name, pod.Namespace, pod.Name)
@@ -544,8 +540,6 @@ func prepareCandidate(c *candidate, pod *v1.Pod, stmt *framework.Statement, ssn 
 	}
 
 	metrics.RegisterPreemptionAttempts()
-
-	return nil
 }
 
 // podTerminatingByPreemption returns true if the pod is in the termination state caused by preempt action.
--- a/pkg/scheduler/cache/cache_test.go
+++ b/pkg/scheduler/cache/cache_test.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
@@ -190,20 +191,21 @@ func TestSchedulerCache_Bind_NodeWithInsufficientResources(t *testing.T) {
 	}
 
 	task.NodeName = "n1"
+	taskBeforeBind := task.Clone()
 	nodeBeforeBind := cache.Nodes["n1"].Clone()
 
 	bindContext := &BindContext{TaskInfo: task}
 	err := cache.AddBindTask(bindContext)
-	if err != nil {
-		t.Errorf("expected bind to succeed even when node is oversubscribed: %v", err)
+	if err == nil {
+		t.Errorf("expected bind to fail for node with insufficient resources")
 	}
 
 	_, taskAfterBind, err := cache.findJobAndTask(task)
 	if err != nil {
 		t.Errorf("expected to find task after failed bind")
 	}
-	if taskAfterBind.Status != api.Binding {
-		t.Errorf("expected task status to be Binding after bind attempt, got %v", taskAfterBind.Status)
+	if !equality.Semantic.DeepEqual(taskBeforeBind, taskAfterBind) {
+		t.Errorf("expected task to remain the same after failed bind: \n %#v\n %#v", taskBeforeBind, taskAfterBind)
 	}
 
 	nodeAfterBind := cache.Nodes["n1"].Clone()
--- a/pkg/scheduler/cache/event_handlers_test.go
+++ b/pkg/scheduler/cache/event_handlers_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cache
 
 import (
+	"fmt"
 	"log"
 	"strings"
 	"testing"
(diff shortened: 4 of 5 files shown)


hzxuzhonghu:
Can review by commit

pkg/scheduler/api/node_info.go
```diff
@@ -461,7 +455,8 @@ func (ni *NodeInfo) AddTask(task *TaskInfo) error {
 		case Binding:
 			// When task in Binding status, it will bind to node, we should double-check whether idle resources are enough to put task before bind to apiserver.
 			if ok, resNames := ti.Resreq.LessEqualWithResourcesName(ni.Idle, Zero); !ok {
-				return fmt.Errorf("node %s resources %v are not enough to put task <%s/%s>, idle: %s, req: %s", ni.Name, resNames, ti.Namespace, ti.Name, ni.Idle.String(), ti.Resreq.String())
```
hzxuzhonghu, line 464: This is my mistake, i will revert this

Aman-Cool:
Hey @hzxuzhonghu, thanks for being open to reverting that! I appreciate you taking another look.

I'm seeing similar concerns in a few other places in this PR where removing error returns might cause issues. The main thing I'm worried about is how this affects the atomicity fix in #5010 ; if operations like `Evict()` and `Pipeline()` can't signal failures, we can't properly rollback when preemption attempts fail.

Would you be open to discussing this here in the PR? Maybe we can find a way to simplify error handling without losing the ability to detect failures and maintain consistency?

The specific concern is: when preemption tries to evict victims and then pipeline the preemptor, if pipeline fails we need to know it failed so we can discard those evictions. Without error returns, we have no way to detect the failure and make that decision.

Happy to explain more about what I'm seeing if it helps!

pkg/scheduler/api/node_info.go
```diff
@@ -597,20 +593,14 @@ func (ni *NodeInfo) subResource(pod *v1.Pod) {
...
-	if err := ni.RemoveTask(ti); err != nil {
-		return err
-	}
-
+func (ni *NodeInfo) UpdateTask(ti *TaskInfo) {
+	ni.RemoveTask(ti)
 	if err := ni.AddTask(ti); err != nil {
 		// This should never happen if task removal was successful,
 		// because only possible error during task addition is when task is still on a node.
 		klog.Fatalf("Failed to add Task <%s,%s> to Node <%s> during task update",
```
hzxuzhonghu, line 601: I think we need to revist this place, fatal should not be used here

hzxuzhonghu:
>I'm seeing similar concerns in a few other places in this PR where removing error returns might cause issues. The main thing I'm worried about is how this affects the atomicity fix in https://github.com/volcano-sh/volcano/pull/5010 ; if operations like Evict() and Pipeline() can't signal failures, we can't properly rollback when preemption attempts fail.

Please take a closer look, all what i have done here is redudant error (never can return)

Aman-Cool:
Removing these error returns breaks what #5010 is fixing.

The fix depends on detecting pipeline failures:

```go
err := tempStmt.Pipeline(preemptor, node)
if err != nil {
    tempStmt.Discard()  // throw away evictions
} else {
    stmt.Merge(tempStmt)  // commit evictions
}
```

Without the error signal, we can't tell if pipeline succeeded or failed. We'd be guessing whether to commit evictions. When we guess wrong, victims get evicted even though the preemptor never scheduled, that's the bug.

I know you're trying to simplify things, but we need that error signal for the transactional pattern to work.

pkg/scheduler/api/job_info.go
```diff
@@ -641,24 +641,15 @@ func (ji *JobInfo) AddTaskInfo(ti *TaskInfo) {
 }
 
 // UpdateTaskStatus is used to update task's status in a job.
-// If error occurs both task and job are guaranteed to be in the original state.
-func (ji *JobInfo) UpdateTaskStatus(task *TaskInfo, status TaskStatus) error {
-	if err := validateStatusUpdate(task.Status, status); err != nil {
```
hajnalmt, line 646: `validateStatusUpdate` is currently empty, but this is a needed functionality I think, we just didn't implement it yet as nobody was willing to implement the state restrictions for this.

pkg/scheduler/api/node_info.go
hajnalmt: We should just `return err` here instead of `klog.FatalF(`.

hajnalmt (commented):
I want to add my comment, which I elaborated on here:

https://github.com/volcano-sh/volcano/pull/5010#pullrequestreview-3795943592

I feel that this change is really risky and instead of deleting the error handling we should move into the direction where we return more errors and handle errors everywhere properly.

pkg/scheduler/api/node_info.go
hzxuzhonghu: Yeah, but this will add a error return value, which need further doing undo error handling in many place.

As the comment say, this could not happen at all.

pkg/scheduler/api/job_info.go
hzxuzhonghu: not sure, i am always prefer cleaning up unused stuff to keep the code more readable

hzxuzhonghu:
> I feel that this change is really risky and instead of deleting the error handling we should move into the direction where we return more errors and handle errors everywhere properly.


I donot actually change any behavior. And this does not conflict with the direction that we should fix. 

It just remove unused return value and keep the error handling simpler.

hzxuzhonghu:
>Without the error signal, we can't tell if pipeline succeeded or failed. We'd be guessing whether to commit evictions. When we guess wrong, victims get evicted even though the preemptor never scheduled, that's the bug.

@Aman-Cool Please donot comment without looking, node_info.go and jon_info.go are the main changes, which remove the never returned error.  How could it influence commit or discard.

pkg/scheduler/api/node_info.go
```diff
@@ -486,16 +486,14 @@ func (ni *NodeInfo) AddTask(task *TaskInfo) error {
...
-// If error occurs both task and node are guaranteed to be in the original state.
-func (ni *NodeInfo) RemoveTask(ti *TaskInfo) error {
+func (ni *NodeInfo) RemoveTask(ti *TaskInfo) {
 	key := PodKey(ti.Pod)
 
 	task, found := ni.Tasks[key]
 	if !found {
 		klog.Warningf("failed to find task <%v/%v> on host <%v>",
 			ti.Namespace, ti.Name, ni.Name)
-		return nil
```
hzxuzhonghu, line 498: I should comment for reviewers, this is not kind of error, just a warning, like that the task is not belonging to this node at all.

pkg/scheduler/api/job_info.go
hzxuzhonghu, line 646: Legacy stuff

pkg/scheduler/api/job_info.go
```diff
@@ -672,7 +663,7 @@ func (ji *JobInfo) deleteTaskIndex(ti *TaskInfo) {
 }
 
 // DeleteTaskInfo is used to delete a task from a job
-func (ji *JobInfo) DeleteTaskInfo(ti *TaskInfo) error {
```
hzxuzhonghu, line 675: There is never a error at all

hzxuzhonghu (commented):
The cleanup is very simple only need to take a look at job_info.go and node_info.go

Aman-Cool:
@hzxuzhonghu I did look at node_info.go and job_info.go. The concern isn't whether these specific functions return errors today, it's that removing error returns from operations like `Evict()` and `Pipeline()` (which you're also changing) breaks the ability to make commit/discard decisions in #5010.

The Merge/Discard pattern needs error signals to know whether operations succeeded or failed. Without them, we can't maintain atomicity in the preemption flow.

hzxuzhonghu:
I am ok to move this cleanup after #5010 in case you added other error return value there

Aman-Cool:
@hzxuzhonghu Thank you so much for sticking with this discussion! I know it's been a lot of back and forth, and I really appreciate you taking the time to dig into the details with me. 

Honestly, the debate helped me understand the atomicity requirements way better than I did before. Your questions pushed me to think through edge cases I hadn't fully considered, and the test I just added came directly from trying to answer your "why does this matter?" question.

Really grateful for your patience and for being willing to wait until #5010 is in. Once that's merged, we can definitely revisit the cleanup and figure out what makes sense together.

JesseStutler:
/approve
These funcs also confuses before, currently I'm ok with it because there are no errors returned in these funcs, only return nil. But also be open to other reviewers if they think some funcs might return an error in the future and it is worthwhile

/cc @hajnalmt @kingeasternsun @lowang-bh @wangyang0616 @ouyangshengjia

pkg/scheduler/api/node_info.go
hajnalmt: Yeah this can never be reached. Let's leave it this way for now.

pkg/scheduler/api/job_info.go
hajnalmt: Fine

hajnalmt (approved):
/lgtm

I reviewed this carefully. The functions being changed do currently always return nil, so this is technically correct and I like that we clear things up.

[#5010](https://github.com/volcano-sh/volcano/pull/5010) (merged last week) introduced per-eviction error handling with UnEvict rollback in prepareCandidate and normalPreempt, specifically to make the transactional preemption pattern robust. This PR strips that out immediately after it was added. The Merge/Discard structure survives, but the inner error handling within temporary statements is gone. But it's right to add this iteration over it, since Evict can't actually fail today, so this doesn't break anything right now. It just undo some of #5010's defensive design. If we later add real failure modes to these operations, we'll need to re-add error returns and all the call-site handling.

The main thing is that I agree with @hzxuzhonghu that the direction should be: https://github.com/volcano-sh/volcano/issues/5072 to centralize rollback behaviour, and this is a good step in that direction.

I'd also prefer keeping the validateStatusUpdate stub as a placeholder for future state machine validation, but I won't block on it as it's dead code now and we can introduce it later.

Issue #5048, opened 2026-02-18
Multiple good-first-issue bugs in capacity plugin, backfill action, cache session lifecycle, and e2e test context dumping


hajnalmt:
## Summary

This issue reports six bugs discovered in the Volcano scheduler and e2e test infrastructure (as of v1.14.0). They range from incorrect metrics reporting to missing capability checks in preemption, a redundant cache session close call, and a test lifecycle ordering issue that prevents failed test context from being captured.

Please cc me in the solution PR, and don't mix them, fix each of them in separate PRs, I want to see a proper analyzitaion of the problem too. I think every one of them can be handled within a day even for a new contributor. Also pick 2 at maximum (leave room to others 😀 at least let's wait a week or 2 if the bugs are picked by someone or not). **Assign the task to yourself if you start working on one and mention the bug number which you choose to solve.**

- [ ] Bug 1: Queue allocated metric diverges from `Queue.Status.Allocated` when hierarchy disabled
   - [x] Assignee: @Aliexe-code 
   - [ ] PR: https://github.com/volcano-sh/volcano/pull/5050 
- [x] Bug 2:  `overused` metric in capacity plugin's `preemptiveFn` is confusing and misplaced
   - [x] Assignee: @aadhil2k4
   - [x] PR: https://github.com/volcano-sh/volcano/pull/5051
- [x] Bug 3: `preemptiveFn` does not check queue capability which allows reclaim above capability/realCapability
   - [x] Assignee: @FAUST-BENCHOU 
   - [x] PR: https://github.com/volcano-sh/volcano/pull/5053 
- [ ] Bug 4: Backfill action does not commit statements for pipelined tasks
   - [x] Assignee: @rayo1uo 
   - [ ] PR: https://github.com/volcano-sh/volcano/pull/5057
- [x] Bug 5: Cache session `OnSessionClose()` called twice per scheduling cycle
   - [x] Assignee: @Nishant-k-sagar 
   - [x] PR: https://github.com/volcano-sh/volcano/pull/5055
- [x] Bug 6: `DumpTestContextIfFailed` never captures context because `defer CleanupTestContext` fires first
   - [x] Assignee @katara-Jayprakash 
   - [x] PR: https://github.com/volcano-sh/volcano/pull/5054 
 
---

## Bug 1: Queue allocated metric diverges from `Queue.Status.Allocated` when hierarchy disabled

[...]

PR #5050, opened 2026-02-20
fix: queue allocated metric consistency for jobless queues in capacit…

Aliexe-code:
/kind bug

   Fixes a bug where jobless queues (like root) had their allocated metrics incorrectly
   reported as zero when hierarchy was disabled, instead of using the actual
   `Queue.Status.Allocated` value.

   Related to #5048 (Bug 1)

   None

   ```release-note
   Fixed queue allocated metric consistency for jobless queues in capacity plugin when
   hierarchy is disabled

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -445,6 +445,33 @@ func (cp *capacityPlugin) OnSessionClose(ssn *framework.Session) {
 	cp.queueOpts = nil
 }
 
+// initQueueAttr initializes a queueAttr from queueInfo.
+// This helper method ensures consistent initialization for both queues with jobs and jobless queues.
+func (cp *capacityPlugin) initQueueAttr(queueInfo *api.QueueInfo) *queueAttr {
+	attr := cp.newQueueAttr(queueInfo)
+
+	// Calculate realCapability: the effective resource limit for this queue
+	realCapability := api.ExceededPart(cp.totalResource, cp.totalGuarantee).Add(attr.guarantee)
+
+	// Only apply MaxFloat64 and MinDimensionResource when Capability is explicitly set.
+	// Distinguish between Capability not set (nil) vs Capability explicitly set to 0.
+	// When Capability is nil, we use MaxFloat64 as a default for unlimited capacity.
+	// When Capability is explicitly set to 0, we respect the user's intent and keep it as 0.
+	if len(queueInfo.Queue.Spec.Capability) != 0 {
+		if attr.capability.MilliCPU <= 0 {
+			attr.capability.MilliCPU = math.MaxFloat64
+		}
+		if attr.capability.Memory <= 0 {
+			attr.capability.Memory = math.MaxFloat64
+		}
+		realCapability.MinDimensionResource(attr.capability, api.Infinity)
+	}
+
+	attr.realCapability = realCapability
+
+	return attr
+}
+
 func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
 	for _, queue := range ssn.Queues {
 		if len(queue.Queue.Spec.Guarantee.Resource) == 0 {
@@ -459,37 +486,7 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
 		klog.V(4).Infof("Considering Job <%s/%s>.", job.Namespace, job.Name)
 		if _, found := cp.queueOpts[job.Queue]; !found {
 			queue := ssn.Queues[job.Queue]
-			attr := &queueAttr{
-				queueID: queue.UID,
-				name:    queue.Name,
-
-				deserved:  api.NewResource(queue.Queue.Spec.Deserved),
-				allocated: api.EmptyResource(),
-				request:   api.EmptyResource(),
-				elastic:   api.EmptyResource(),
-				inqueue:   api.EmptyResource(),
-				guarantee: api.EmptyResource(),
-			}
-			if len(queue.Queue.Spec.Capability) != 0 {
-				attr.capability = api.NewResource(queue.Queue.Spec.Capability)
-				if attr.capability.MilliCPU <= 0 {
-					attr.capability.MilliCPU = math.MaxFloat64
-				}
-				if attr.capability.Memory <= 0 {
-					attr.capability.Memory = math.MaxFloat64
-				}
-			}
-			if len(queue.Queue.Spec.Guarantee.Resource) != 0 {
-				attr.guarantee = api.NewResource(queue.Queue.Spec.Guarantee.Resource)
-			}
-			realCapability := api.ExceededPart(cp.totalResource, cp.totalGuarantee).Add(attr.guarantee)
-			if attr.capability == nil {
-				attr.capability = api.EmptyResource()
-				attr.realCapability = realCapability
-			} else {
-				realCapability.MinDimensionResource(attr.capability, api.Infinity)
-				attr.realCapability = realCapability
-			}
+			attr := cp.initQueueAttr(queue)
 			cp.queueOpts[job.Queue] = attr
 			klog.V(4).Infof("Added Queue <%s> attributes.", job.Queue)
 		}
@@ -528,10 +525,22 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
 			attr.name, attr.allocated.String(), attr.request.String(), attr.inqueue.String(), attr.elastic.String())
 	}
 
-	for _, attr := range cp.queueOpts {
-		if attr.realCapability != nil {
-			attr.deserved.MinDimensionResource(attr.realCapability, api.Infinity)
+	// Populate queueAttr for jobless queues using Queue.Status.Allocated
+	// This ensures metrics are consistent with Queue.Status.Allocated when hierarchy is disabled
+	for queueID, queueInfo := range ssn.Queues {
+		if _, exists := cp.queueOpts[queueID]; !exists {
+			attr := cp.initQueueAttr(queueInfo)
+			// Use Queue.Status.Allocated for jobless queues to ensure metric consistency
+			if queueInfo.Queue.Status.Allocated != nil {
+				attr.allocated = api.NewResource(queueInfo.Queue.Status.Allocated)
+			}
+			cp.queueOpts[queueID] = attr
+			klog.V(4).Infof("Added jobless Queue <%s> attributes with allocated from status: <%v>", queueInfo.Name, attr.allocated)
 		}
+	}
+
+	for _, attr := range cp.queueOpts {
+		attr.deserved.MinDimensionResource(attr.realCapability, api.Infinity)
 
 		attr.deserved = helpers.Max(attr.deserved, attr.guarantee)
 		cp.updateShare(attr)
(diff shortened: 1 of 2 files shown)


hajnalmt:
/cc hajnalmt

@Aliexe-code  Can I review this one? :)

hajnalmt:
Hello @Aliexe-code ,
What's the progress with this 😊 ? Can I start the review on it?

Aliexe-code:
@hajnalmt Please go ahead , sorry for late responding :)

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -401,6 +401,42 @@ func (cp *capacityPlugin) OnSessionClose(ssn *framework.Session) {
 	cp.queueOpts = nil
 }
 
+// initQueueAttr initializes a queueAttr from queueInfo.
+// This helper method ensures consistent initialization for both queues with jobs and jobless queues.
+func (cp *capacityPlugin) initQueueAttr(queueInfo *api.QueueInfo) *queueAttr {
+	attr := &queueAttr{
```
hajnalmt: I think we can reuse `newQueueAttr` here to avoid code duplication.
```
attr := cp.newQueueAttr(queue)
```

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -401,6 +401,42 @@ func (cp *capacityPlugin) OnSessionClose(ssn *framework.Session) {
...
+		deserved:  api.NewResource(queueInfo.Queue.Spec.Deserved),
+		allocated: api.EmptyResource(),
+		request:   api.EmptyResource(),
+		elastic:   api.EmptyResource(),
+		inqueue:   api.EmptyResource(),
+		guarantee: api.EmptyResource(),
+	}
+	if len(queueInfo.Queue.Spec.Capability) != 0 {
+		attr.capability = api.NewResource(queueInfo.Queue.Spec.Capability)
+		if attr.capability.MilliCPU <= 0 {
```
hajnalmt, line 1069: If you switch to newQueueAttr you still need to handle this as a defensive measure, but `attr.capability` will be set to an EmptyResource so it's never nil, the same is true for `attr.realCapability`. You can even merge the two if blocks (this and the one at the end of this function) and handle the realCapability settings and the <=0 negative guard rails in one go.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -495,40 +515,15 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
...
-				metrics.UpdateQueueCapacity(attr.name, attr.capability.MilliCPU, attr.capability.Memory, attr.capability.ScalarResources)
-			}
-			metrics.UpdateQueueRealCapacity(attr.name, attr.realCapability.MilliCPU, attr.realCapability.Memory, attr.realCapability.ScalarResources)
-			continue
+	// Record metrics - all queues are now guaranteed to be in queueOpts
+	for _, attr := range cp.queueOpts {
+		metrics.UpdateQueueDeserved(attr.name, attr.deserved.MilliCPU, attr.deserved.Memory, attr.deserved.ScalarResources)
+		metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory, attr.allocated.ScalarResources)
+		metrics.UpdateQueueRequest(attr.name, attr.request.MilliCPU, attr.request.Memory, attr.request.ScalarResources)
+		if attr.capability != nil {
```
hajnalmt: This is never `nil` at this point. Can you delete this branch please?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -484,6 +490,20 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
...
+			if queueInfo.Queue.Status.Allocated != nil {
+				attr.allocated = api.NewResource(queueInfo.Queue.Status.Allocated)
+			}
+			cp.queueOpts[queueID] = attr
+			klog.V(4).Infof("Added jobless Queue <%s> attributes with allocated from status: <%v>", queueInfo.Name, attr.allocated)
+		}
+	}
+
 	for _, attr := range cp.queueOpts {
 		if attr.realCapability != nil {
```
hajnalmt: This will never be `nil` at this point. Can you delete this branch please?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -495,40 +515,15 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
...
-			}
-			metrics.UpdateQueueRealCapacity(attr.name, attr.realCapability.MilliCPU, attr.realCapability.Memory, attr.realCapability.ScalarResources)
-			continue
+	// Record metrics - all queues are now guaranteed to be in queueOpts
+	for _, attr := range cp.queueOpts {
+		metrics.UpdateQueueDeserved(attr.name, attr.deserved.MilliCPU, attr.deserved.Memory, attr.deserved.ScalarResources)
+		metrics.UpdateQueueAllocated(attr.name, attr.allocated.MilliCPU, attr.allocated.Memory, attr.allocated.ScalarResources)
+		metrics.UpdateQueueRequest(attr.name, attr.request.MilliCPU, attr.request.Memory, attr.request.ScalarResources)
+		if attr.capability != nil {
+			metrics.UpdateQueueCapacity(attr.name, attr.capability.MilliCPU, attr.capability.Memory, attr.capability.ScalarResources)
```
hajnalmt: We have that defensive measure that sets capability for `maxFloat64` if it's <=0 for cpu and memory, it will be a very big number in these updates. I wonder if we should have that or just 0. Let's leave it this way now.

(This is true for the hierachy case too just wanted to mention this here.)

hajnalmt (commented):
Thanks looks Overall good to me 😊 
I had minor comments only can you please address them?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -401,6 +401,27 @@ func (cp *capacityPlugin) OnSessionClose(ssn *framework.Session) {
 	cp.queueOpts = nil
 }
 
+// initQueueAttr initializes a queueAttr from queueInfo.
+// This helper method ensures consistent initialization for both queues with jobs and jobless queues.
+func (cp *capacityPlugin) initQueueAttr(queueInfo *api.QueueInfo) *queueAttr {
+	attr := cp.newQueueAttr(queueInfo)
+
+	// Set MaxFloat64 for CPU/Memory if <= 0 (defensive measure for unlimited capacity)
+	if attr.capability.MilliCPU <= 0 {
```
hajnalmt: This will set these to be MaxFloat64 always when they are set to 0 or missed too, this is two different cases! We want to set them only when they are not explicitly set to 0.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -401,6 +401,27 @@ func (cp *capacityPlugin) OnSessionClose(ssn *framework.Session) {
...
+		attr.capability.MilliCPU = math.MaxFloat64
+	}
+	if attr.capability.Memory <= 0 {
+		attr.capability.Memory = math.MaxFloat64
+	}
+
+	// Calculate realCapability: the effective resource limit for this queue
+	realCapability := api.ExceededPart(cp.totalResource, cp.totalGuarantee).Add(attr.guarantee)
+	realCapability.MinDimensionResource(attr.capability, api.Infinity)
```
hajnalmt: if `Queue.Spec.Capability` is set to nil this will override it! You need to distinguish these two settings via `if len(queue.Queue.Spec.Capability) != 0`  like in newQueueAttr.
The right code here:
```
realCapability := api.ExceededPart(cp.totalResource, cp.totalGuarantee).Add(attr.guarantee)
if len(queueInfo.Queue.Spec.Capability) != 0 {
	if attr.capability.MilliCPU <= 0 {
		attr.capability.MilliCPU = math.MaxFloat64
	}
	if attr.capability.Memory <= 0 {
		attr.capability.Memory = math.MaxFloat64
	}
	realCapability.MinDimensionResource(attr.capability, api.Infinity)
}
attr.realCapability = realCapability
```

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -958,3 +958,83 @@ func Test_updateQueueAttrShare(t *testing.T) {
 		})
 	}
 }
+
+// TestBuildQueueAttrsMetricConsistency tests Bug 1 fix:
+// Queue allocated metric should be consistent with Queue.Status.Allocated
+// when hierarchy is disabled, including for jobless queues (like root queue).
+// Before the fix: jobless queues were not added to queueOpts, causing metrics to report 0 allocated.
```
hajnalmt: causing metrics to report stale values as allocated.

hajnalmt (changes requested):
Thanks for the change!
Please sign the DCO on your commit 😊  We are getting there, the capability/realCapability logic is not right in the new change, can you fix that please?

hajnalmt:
Hello @Aliexe-code,
We are really close with this PR 😊 can you address my last requests please?

Aliexe-code:
@hajnalmt  Thank you so much for your patience and thorough review! The changes are now pushed. I really appreciate your guidance in helping me get this right!

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -429,22 +389,45 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+	attr := cp.newQueueAttr(queueInfo)
+
+	// Calculate realCapability: the effective resource limit for this queue
+	realCapability := api.ExceededPart(cp.totalResource, cp.totalGuarantee).Add(attr.guarantee)
+
+	// Only apply MaxFloat64 and MinDimensionResource when Capability is explicitly set.
+	// Distinguish between Capability not set (nil) vs Capability explicitly set to 0.
+	// When Capability is nil, we use MaxFloat64 as a default for unlimited capacity.
+	// When Capability is explicitly set to 0, we respect the user's intent and keep it as 0.
+	if queueInfo.Queue.Spec.Capability != nil && len(queueInfo.Queue.Spec.Capability) != 0 {
```
hajnalmt: nitpick: The != nil check here is technically redundant since len(nil) == 0 in Go, so len(queueInfo.Queue.Spec.Capability) != 0 alone would suffice. That said, I think it's fine to keep it nicely together with the comment, it makes the explicit distinction between "not set" and "set to 0" very readable.

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -958,3 +958,83 @@ func Test_updateQueueAttrShare(t *testing.T) {
...
+
+		// The fix ensures the root queue's allocated is read from Status.Allocated
+		expectedAllocated := api.NewResource(api.BuildResourceList("2", "2Gi"))
+		actualAllocated := api.NewResource(rootQueue.Queue.Status.Allocated)
+		if actualAllocated.MilliCPU != expectedAllocated.MilliCPU {
+			t.Errorf("Root queue allocated CPU: got %v, want %v", actualAllocated.MilliCPU, expectedAllocated.MilliCPU)
+		}
+		if actualAllocated.Memory != expectedAllocated.Memory {
+			t.Errorf("Root queue allocated Memory: got %v, want %v", actualAllocated.Memory, expectedAllocated.Memory)
+		}
```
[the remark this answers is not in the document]
hajnalmt: Yeah this would be good, but not a necessity the test verifies root at least 👍

hajnalmt (approved):
/area scheduling
/kind bug
/ok-to-test

Thank you so much! Let's see how the testing goes, and I will put an lgtm on this  if it goes through.

hajnalmt (commented):
Oh I see you have a merge conflict, can you rebase on master please?

Aliexe-code:
@hajnalmt Rebased on latest upstream/master.

pkg/scheduler/plugins/capacity/capacity.go
hajnalmt: @Aliexe-code The static check failed here exactly with this error.
It looks like we need to check only len here.
```
level=info msg="Execution took 5m13.131287087s"
Error: pkg/scheduler/plugins/capacity/capacity.go:460:5: S1009: should omit nil check; len() for nil maps is defined as zero (staticcheck)
	if queueInfo.Queue.Spec.Capability != nil && len(queueInfo.Queue.Spec.Capability) != 0 {
```

hajnalmt (commented):
Thanks!
The PR failed on the static check, can you please fix this line? otherwise lgtm.

```
level=info msg="Execution took 5m13.131287087s"
Error: pkg/scheduler/plugins/capacity/capacity.go:460:5: S1009: should omit nil check; len() for nil maps is defined as zero (staticcheck)
	if queueInfo.Queue.Spec.Capability != nil && len(queueInfo.Queue.Spec.Capability) != 0 {
```

Aliexe-code:
@hajnalmt Fixed the static check issue. Removed the redundant nil check as suggested. The CI should pass now. Thanks!

hajnalmt (commented):
Hello,
@Aliexe-code I don't see a commit, and the code is the same. Did you forget to push?

test/e2e/jobseq/queue_job_status.go
```diff
@@ -156,7 +156,6 @@ var _ = Describe("Queue Job Status Transition", func() {
 
 		podNamespace = testCtx.Namespace
 		slot := e2eutil.HalfCPU
-		rep = e2eutil.ClusterSize(testCtx, slot)
```
hajnalmt: Why did you delete this?

hajnalmt (commented):
It looks good to me now! Thank you.

Why did you touch `queue_job_status.go` in jobseq e2e test?

Aliexe-code:
@hajnalmt Thanks for catching that, The change to queue_job_status.go was accidental. I've amended the commit to restore the deleted line. The PR now only contains the intended capacity plugin changes.

hajnalmt (approved):
/lgtm
Thank you so much for the contribution!
@JesseStutler

katara-Jayprakash:
/cc @Aliexe-code make sure to resolve the conflicts and then
@JesseStutler @hajnalmt lets merge this code sir!

Aliexe-code:
Hi @hajnalmt and @k82cn

I've rebased this PR onto the latest `master` to clear the merge conflict. While doing so I noticed the original `TestBuildQueueAttrsMetricConsistency` only checked `Queue.Status.Allocated` on the queue object itself and never actually verified the plugin's behavior — so it wasn't really proving the fix. I rewrote it to run `buildQueueAttrs` and assert that the **jobless root queue is actually present in `queueOpts` with `allocated` derived from `Status.Allocated`** (and confirmed a job-backed queue still gets its allocated from the job). It now fails without the fix.

The core change is unchanged: jobless queues are added to `queueOpts` with `allocated` from `Queue.Status.Allocated`, and the metrics loop iterates uniformly over `queueOpts`. I also kept `UpdateQueueInqueue` in that loop (it had been dropped in the simplified version) so inqueue metrics keep reporting for all queues.

Locally everything is green: `go build`, `go vet`, the full capacity test suite, and `staticcheck` (the S1009 lint that blocked the earlier LGTM).

Would appreciate a final look whenever you have a moment — and the `/approve` to get it merged. Thanks again for the patience and reviews!

PR #5057, opened 2026-02-25
fix: backfill action does not commit statements for pipelined tasks

rayo1uo:

part of https://github.com/volcano-sh/volcano/issues/5048
Bug 4: Backfill action does not commit statements for pipelined tasks

+ Refactored the backfill action to utilize `allocate` and `unpipeline` at the statement level instead of the session level.

+ Add Commit And Discard Function in `TestPickUpPendingTasks`.

+ Add unit tests `TestBackfillReallocateBestEffortTask` for backfill action.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/backfill/backfill.go
+++ b/pkg/scheduler/actions/backfill/backfill.go
@@ -35,9 +35,9 @@ import (
 )
 
 type backfillContext struct {
-	queues      *util.PriorityQueue                 // queue of *api.QueueInfo
-	jobsByQueue map[api.QueueID]*util.PriorityQueue // queue of *api.JobInfo
-	tasksByJob  map[api.JobID]*util.PriorityQueue   // queue of *api.TaskInfo
+	queues         *util.PriorityQueue                 // queue of *api.QueueInfo
+	jobsByQueue    map[api.QueueID]*util.PriorityQueue // queue of *api.JobInfo
+	taskQueueByJob map[api.JobID]*util.PriorityQueue   // queue of *api.TaskInfo
 }
 
 type Action struct {
@@ -77,9 +77,9 @@ func (backfill *Action) buildBackfillContext() *backfillContext {
 	ssn := backfill.session
 
 	actx := &backfillContext{
-		queues:      util.NewPriorityQueue(ssn.QueueOrderFn),
-		jobsByQueue: map[api.QueueID]*util.PriorityQueue{},
-		tasksByJob:  map[api.JobID]*util.PriorityQueue{},
+		queues:         util.NewPriorityQueue(ssn.QueueOrderFn),
+		jobsByQueue:    map[api.QueueID]*util.PriorityQueue{},
+		taskQueueByJob: map[api.JobID]*util.PriorityQueue{},
 	}
 
 	for _, job := range ssn.Jobs {
@@ -97,12 +97,12 @@ func (backfill *Action) buildBackfillContext() *backfillContext {
 			continue
 		}
 
-		tasks := util.NewPriorityQueue(ssn.TaskOrderFn)
+		taskQueue := util.NewPriorityQueue(ssn.TaskOrderFn)
 		for _, task := range job.TaskStatusIndex[api.Pending] {
 			if !task.BestEffort || task.SchGated {
 				continue
 			}
-			tasks.Push(task)
+			taskQueue.Push(task)
 		}
 
 		for _, task := range job.TaskStatusIndex[api.Pipelined] {
@@ -114,10 +114,10 @@ func (backfill *Action) buildBackfillContext() *backfillContext {
 				klog.Warningf("Failed to unpipeline task: %s", err.Error())
 				continue
 			}
-			tasks.Push(task)
+			taskQueue.Push(task)
 		}
 
-		if tasks.Empty() {
+		if taskQueue.Empty() {
 			continue
 		}
 
@@ -126,7 +126,7 @@ func (backfill *Action) buildBackfillContext() *backfillContext {
 			actx.queues.Push(queue)
 		}
 		actx.jobsByQueue[queue.UID].Push(job)
-		actx.tasksByJob[job.UID] = tasks
+		actx.taskQueueByJob[job.UID] = taskQueue
 	}
 
 	return actx
@@ -139,25 +139,25 @@ func (backfill *Action) allocateResources(actx *backfillContext) {
 	for !queues.Empty() {
 		queue := queues.Pop().(*api.QueueInfo)
 
-		jobs, found := actx.jobsByQueue[queue.UID]
-		if !found || jobs.Empty() {
+		jobQueue, found := actx.jobsByQueue[queue.UID]
+		if !found || jobQueue.Empty() {
 			continue
 		}
 
-		job := jobs.Pop().(*api.JobInfo)
-		tasks, found := actx.tasksByJob[job.UID]
-		if !found || tasks.Empty() {
+		job := jobQueue.Pop().(*api.JobInfo)
+		taskQueue, found := actx.taskQueueByJob[job.UID]
+		if !found || taskQueue.Empty() {
 			queues.Push(queue)
 			continue
 		}
 
 		stmt := framework.NewStatement(ssn)
-		backfill.allocateTasksForJob(stmt, job, tasks)
+		backfill.allocateTasksForJob(stmt, job, taskQueue)
 
 		if len(stmt.Operations()) > 0 && ssn.JobReady(job) {
 			stmt.Commit()
-			if !tasks.Empty() {
-				jobs.Push(job)
+			if !taskQueue.Empty() {
+				jobQueue.Push(job)
 			}
 		} else {
 			stmt.Discard()
@@ -167,13 +167,13 @@ func (backfill *Action) allocateResources(actx *backfillContext) {
 	}
 }
 
-func (backfill *Action) allocateTasksForJob(stmt *framework.Statement, job *api.JobInfo, tasks *util.PriorityQueue) {
+func (backfill *Action) allocateTasksForJob(stmt *framework.Statement, job *api.JobInfo, taskQueue *util.PriorityQueue) {
 	ssn := backfill.session
 	predicateFunc := ssn.PredicateForAllocateAction
 	ph := util.NewPredicateHelper()
 
-	for !tasks.Empty() {
-		task := tasks.Pop().(*api.TaskInfo)
+	for !taskQueue.Empty() {
+		task := taskQueue.Pop().(*api.TaskInfo)
 
 		fe := api.NewFitErrors()
 		if err := ssn.PrePredicateFn(task); err != nil {
--- a/pkg/scheduler/actions/backfill/backfill_test.go
+++ b/pkg/scheduler/actions/backfill/backfill_test.go
@@ -196,7 +196,7 @@ func TestBuildBackfillContext(t *testing.T) {
 			}
 			job := jobs.Pop().(*api.JobInfo)
 
-			tasks, found := actx.tasksByJob[job.UID]
+			tasks, found := actx.taskQueueByJob[job.UID]
 			if !found || tasks.Empty() {
 				actx.queues.Push(queue)
 				continue


rayo1uo:
cc @hajnalmt Please take a look and let me know if this pr looks correct, thanks!

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -157,8 +157,7 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
 				continue
 			}
 
-			stmt := framework.NewStatement(ssn)
-			err := stmt.UnPipeline(task)
+			err := ssn.UnPipeline(task)
```
hajnalmt: We are trying to move away from the direct methods on the session receiver towards the stament usage. But for `backfill` and `shuffle` maybe this makes sense as these are one-time calls and we don't need to organize these. I am fine with this for now.

pkg/scheduler/framework/session.go
```diff
@@ -791,6 +791,49 @@ func (ssn *Session) Pipeline(task *api.TaskInfo, hostname string) error {
...
+	}
+
+	for _, eh := range ssn.eventHandlers {
+		if eh.DeallocateFunc != nil {
+			eventInfo := &Event{
+				Task: task,
+			}
+			eh.DeallocateFunc(eventInfo)
+			if eventInfo.Err != nil {
+				klog.Errorf("Failed to exec deallocate callback functions for task <%v/%v> to node <%v> when pipeline in Session <%v>: %v",
```
hajnalmt: when unpipeline in Session.

pkg/scheduler/framework/session.go
```diff
@@ -791,6 +791,49 @@ func (ssn *Session) Pipeline(task *api.TaskInfo, hostname string) error {
...
+		if err := job.UpdateTaskStatus(task, api.Pending); err != nil {
+			klog.Errorf("Failed to update task <%v/%v> status to %v when unpipeline in Session <%v>: %v",
+				task.Namespace, task.Name, api.Pending, ssn.UID, err)
+		}
+	} else {
+		klog.Errorf("Failed to find Job <%s> in Session <%s> index when unpipeline.", task.Job, ssn.UID)
+	}
+
+	if node, found := ssn.Nodes[task.NodeName]; found {
+		if err := node.RemoveTask(task); err != nil {
```
hajnalmt: I think we shall collect these errors like statement `Pipeline` does with `errInfos := make([]error, 0)` . The right approach is to continue through all steps, and return a summary error at the end. Silently swallowing errors and returning nil at the end feels wrong.

pkg/scheduler/framework/statement.go
```diff
@@ -219,44 +219,7 @@ func (s *Statement) pipeline(task *api.TaskInfo) {
...
-				klog.Errorf("Failed to exec deallocate callback functions for task <%v/%v> to node <%v> when pipeline in Session <%v>: %v",
-					task.Namespace, task.Name, task.NodeName, s.ssn.UID, eventInfo.Err)
-			}
-		}
-	}
-	task.NodeName = ""
-	task.JobAllocatedHyperNode = ""
-
-	return nil
+	return s.ssn.UnPipeline(task)
```
hajnalmt: This is an interesting take. You have effectively made Statement usage a no-op wrapper for a session operation. It makes sense to me for these kind of rollback/undo operation.

pkg/scheduler/framework/session.go
rayo1uo: Thanks, fixed.

pkg/scheduler/framework/session.go
rayo1uo: This makes sense to me. Revision has been done.

rayo1uo:
@hajnalmt Thanks a lot for your thorough code review. I've addressed all your comments – feel free to check it out whenever you're free.

pkg/scheduler/actions/backfill/backfill.go
JesseStutler: I prefer to not add more operations on session, we should move toward using statement, even in backfill action the Allocate still use session's Allocate method, which should be corrected to use statement instead

JesseStutler:
Actually I don't now the historical reason of why backfill action will handle pipelined task( in normal case, the backfill action will be configured after allocate action, only action will mark a non-best effort pod as pipelined if the req of the pod is less than futureidle) But I understand that it might be for defensive programming, backfill can allow best effort pods to be scheduled all the time.

But I still prefer to move operations in the session to statement instead, all the existing method called in backfill should also use statement instead. The previous definition of operations in the session was likely due to an initial design flaw that didn't consider transaction rollback.

rayo1uo:
> Actually I don't now the historical reason of why backfill action will handle pipelined task( in normal case, the backfill action will be configured after allocate action, only action will mark a non-best effort pod as pipelined if the req of the pod is less than futureidle) But I understand that it might be for defensive programming, backfill can allow best effort pods to be scheduled all the time.
> 
> But I still prefer to move operations in the session to statement instead, all the existing method called in backfill should also use statement instead. The previous definition of operations in the session was likely due to an initial design flaw that didn't consider transaction rollback.

@JesseStutler Thanks for your input.

The `Commit` and `Discard` methods currently defined in statement.go have no functional impact on the `Unpipeline` function. I'd like to confirm: should we retain the existing implementation (invoking `stmt.Unpipeline` without `Commit` or `Discard`), or add these method calls despite their lack of actual effect?  cc @hajnalmt

rayo1uo:
@hajnalmt @JesseStutler BTW, I could also work on another issue to refactor the backfill action to use stmt.Allocate instead of ssn.Allocate. Would you prefer me to do this in the current PR instead?

My initial thought is to follow the pattern used by the allocate action:
First pick up jobs with pending best-effort tasks, perform allocate, then commit if the job is ready. Otherwise, discard if an error occurs or the job is not pipelined.

JesseStutler:
> > Actually I don't now the historical reason of why backfill action will handle pipelined task( in normal case, the backfill action will be configured after allocate action, only action will mark a non-best effort pod as pipelined if the req of the pod is less than futureidle) But I understand that it might be for defensive programming, backfill can allow best effort pods to be scheduled all the time.
> > But I still prefer to move operations in the session to statement instead, all the existing method called in backfill should also use statement instead. The previous definition of operations in the session was likely due to an initial design flaw that didn't consider transaction rollback.
> 
> @JesseStutler Thanks for your input.
> 
> The `Commit` and `Discard` methods currently defined in statement.go have no functional impact on the `Unpipeline` function. I'd like to confirm: should we retain the existing implementation (invoking `stmt.Unpipeline` without `Commit` or `Discard`), or add these method calls despite their lack of actual effect? cc @hajnalmt

In fact, pipeline, allocate with commit is reasonable, because it is possible that the previous assign result should be returned if the number of minAvaiable is not reached, but unpipeline I think there is no side effect, there is no such thing as satisfying minAvaialble, so the starting point of this issue has to be reconsidered. But I think it makes sense to use a statement instead of operations that already use sessions in the backfill

JesseStutler:
> @hajnalmt @JesseStutler BTW, I could also work on another issue to refactor the backfill action to use stmt.Allocate instead of ssn.Allocate. Would you prefer me to do this in the current PR instead?
> 
> My initial thought is to follow the pattern used by the allocate action: First pick up jobs with pending best-effort tasks, perform allocate, then commit if the job is ready. Otherwise, discard if an error occurs or the job is not pipelined.

Yes, I think this is the target I want to reach, also wait @hajnalmt to reply

hajnalmt (changes requested):
/hold

I was thinking about this a lot and in the end it came out that in my proposal here: https://github.com/volcano-sh/volcano/pull/5045 
I will need to add Unpipeline to the statement handling logic it will be a commitable save operation in the cache, so even though currently Commit does not have an effect on Unpipeline now it will have possibly in the future (if the proposal goes through).

So let's go with the statement handling I agree with Jesse. (Please modify allocate in  backfill too.)

rayo1uo:
@hajnalmt @JesseStutler Could you kindly review this again? The backfill action has been refactored to use allocate and unpipeline at the statement level (replacing the session-level implementation), and this change is now complete. Thanks~

JesseStutler:
> @hajnalmt @JesseStutler Could you kindly review this again? The backfill action has been refactored to use allocate and unpipeline at the statement level (replacing the session-level implementation), and this change is now complete. Thanks~

Are you trying to keep the written style consistent with the Allocate action? Why such a big change all of a sudden? I think change the session.Allocate to statement.Allocate and relate with statement.Commit is enough :)

rayo1uo:
> > @hajnalmt @JesseStutler Could you kindly review this again? The backfill action has been refactored to use allocate and unpipeline at the statement level (replacing the session-level implementation), and this change is now complete. Thanks~
> 
> Are you trying to keep the written style consistent with the Allocate action? Why such a big change all of a sudden? I think change the session.Allocate to statement.Allocate and relate with statement.Commit is enough :)

Yes, I think keeping the same style as the Allocate action is more reasonable.
This achieves the goal of fair-share allocation for best-effort tasks.
BTW, I don't think simply changing session.Allocate to statement.Allocate is sufficient, because we also need to determine when to commit or discard the statements.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
 			if !task.BestEffort {
 				continue
 			}
-
 			stmt := framework.NewStatement(ssn)
-			err := stmt.UnPipeline(task)
-			if err != nil {
+			if err := stmt.UnPipeline(task); err != nil {
 				klog.Errorf("Failed to unpipeline task: %s", err.Error())
 				continue
```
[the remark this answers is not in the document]
hajnalmt: Can you modify this to Warningf? Copilot is right here.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
 
 		for _, task := range job.TaskStatusIndex[api.Pipelined] {
 			if !task.BestEffort {
 				continue
 			}
-
 			stmt := framework.NewStatement(ssn)
-			err := stmt.UnPipeline(task)
-			if err != nil {
+			if err := stmt.UnPipeline(task); err != nil {
```
hajnalmt, line 113: Please Discard the statement on an err and commit it  otherwise. Let's keep the boundaries proper. To not leave ghost "unpipelined" best effort tasks.

pkg/scheduler/actions/backfill/backfill_test.go
```diff
@@ -25,20 +25,24 @@ import (
...
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
-func TestPickUpPendingTasks(t *testing.T) {
+func TestBuildBackfillContext(t *testing.T) {
 	framework.RegisterPluginBuilder("priority", priority.New)
 	framework.RegisterPluginBuilder("drf", drf.New)
+	defer framework.CleanupPluginBuilders()
 	trueValue := true
 	tilers := []conf.Tier{
```
hajnalmt: If you are here can you fix the typo here? `tiers :=`

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
-				pendingTasks = append(pendingTasks, tasks[job.UID].Pop().(*api.TaskInfo))
+		} else {
+			stmt.Discard()
+		}
+
+		queues.Push(queue)
+	}
+}
+
+func (backfill *Action) allocateTasksForJob(stmt *framework.Statement, queue *api.QueueInfo, job *api.JobInfo, tasks *util.PriorityQueue) {
```
hajnalmt: You are not using queue in this function.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
+		tasks, found := actx.tasksByJob[job.UID]
+		if !found || tasks.Empty() {
+			queues.Push(queue)
+			continue
+		}
+
+		stmt := framework.NewStatement(ssn)
+		backfill.allocateTasksForJob(stmt, queue, job, tasks)
+
+		if len(stmt.Operations()) > 0 && ssn.JobReady(job) {
```
hajnalmt, line 157: Ugh. I don't think we should have `ssn.JobReady(job)` here. Backfill allocates best effort tasks only and we don't need the job to be ready like in allocate as we are not having gang scheduling concerns.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
+			}
+			fe.SetNodeError(node.Name, err)
+			job.NodesFitErrors[task.UID] = fe
+			continue
+		}
+
+		metrics.UpdateE2eSchedulingDurationByJob(job.Name, string(job.Queue), job.Namespace, metrics.Duration(job.CreationTimestamp.Time))
+		metrics.UpdateE2eSchedulingLastTimeByJob(job.Name, string(job.Queue), job.Namespace, time.Now())
+
+		if ssn.JobReady(job) {
```
hajnalmt, line 223: This is unnecesarry here too and stops BestEffort task scheduling when minAvailable is reached...

hajnalmt (changes requested):
Thanks for the changes, but this seems really risky and  far-reaching..
I am not against rewriting the whole action but we will need more people to take a look at it. The intention that we want to align this implementation with `allocate` is a thing I can absolutely support, with queue order honoring in backfill action too. The latter is an exceptionally great addition I think.

pkg/scheduler/actions/backfill/backfill.go
hajnalmt: Or I don't know tbh. My issue here is a mixed job with BestEffort and non BestEffort tasks. Currently volcano does not gang schedules best effort tasks (we don't support that) and if we bring in this feature we can break the scheduling for these existing "mixed" jobs. These kind of jobs can indefenitely hang. Imagine a job with 2 BestEffort tasks and 2 non BestEffort with 3 as minAvailable. This will never schedule as minAvailable cannot be reached by the BestEffort  tasks in `allocate`, and also by the non BestEffort tasks in `backfill` as 2 is always less then 3. In the current implementation the 2 best effort pods will schedule in backfill and if 1 of the non BestEfforts is able to schedule it can pass allocate.

rayo1uo: Thanks for supplementing the relevant context – I think it makes sense. However, it appears the original implementation also calls the ready function, which may have the same issue. (https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/framework/session.go#L787)

pkg/scheduler/actions/backfill/backfill.go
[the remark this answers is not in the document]
rayo1uo: Fix done

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: Fixed

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: Fixed

rayo1uo:
@hajnalmt Thanks for your thorough review. This PR is ready for further review. 🚀

pkg/scheduler/actions/backfill/backfill.go
hajnalmt: Oh I was wrong here then 👍 Please do this back. I missed that! We did gang schedule best effort pods you are right, probably for this mixed job issue the answer is that don't do something like this or reduce `minAvailable`. 😊

rayo1uo: @hajnalmt  No worries at all!  I'll revert this change right away. And thanks for clarifying – I agree that for this mixed job issue, either avoiding this approach or reducing minAvailable would be the right solution.

pkg/scheduler/framework/statement.go
```diff
@@ -225,27 +225,37 @@ func (s *Statement) Pipeline(task *api.TaskInfo, hostname string, evictionOccurr
...
 func (s *Statement) UnPipeline(task *api.TaskInfo) error {
+	// Only update status in session
+	errInfos := make([]error, 0)
+	nodeName := task.NodeName
 	job, found := s.ssn.Jobs[task.Job]
 	if found {
 		if err := job.UpdateTaskStatus(task, api.Pending); err != nil {
 			klog.Errorf("Failed to update task <%v/%v> status to %v when unpipeline in Session <%v>: %v",
 				task.Namespace, task.Name, api.Pending, s.ssn.UID, err)
+			errInfos = append(errInfos, err)
```
hzxuzhonghu: I think a log  for this kind of `not found` error is enough, why do we need error returned?

pkg/scheduler/framework/statement.go
```diff
@@ -225,27 +225,37 @@ func (s *Statement) Pipeline(task *api.TaskInfo, hostname string, evictionOccurr
...
 		}
 		klog.V(3).Infof("After unpipelined Task <%v/%v> to Node <%v>: idle <%v>, used <%v>, releasing <%v>",
 			task.Namespace, task.Name, node.Name, node.Idle, node.Used, node.Releasing)
 	} else {
-		klog.Errorf("Failed to find Node <%s> in Session <%s> index when unpipeline.",
-			task.NodeName, s.ssn.UID)
+		err := fmt.Errorf("Failed to find Node <%s> in Session <%s> index when unpipeline.",
+			nodeName, s.ssn.UID)
+		klog.Errorf("%v", err)
+		errInfos = append(errInfos, err)
```
hzxuzhonghu: same

pkg/scheduler/framework/statement.go
hzxuzhonghu: I am concerned if this would cause other issue

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,135 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
+			if err := stmt.UnPipeline(task); err != nil {
+				klog.Warningf("Failed to unpipeline task: %s", err.Error())
+				stmt.Discard()
 				continue
 			}
-			if _, existed := tasks[job.UID]; !existed {
-				tasks[job.UID] = util.NewPriorityQueue(ssn.TaskOrderFn)
-			}
-			tasks[job.UID].Push(task)
+			stmt.Commit()
```
hzxuzhonghu: @hajnalmt Not sure i understand commit what? UnPipeline does not add any operation

pkg/scheduler/framework/statement.go
rayo1uo: Thanks for kindly review! The initial intention was to align the style with the Pipeline function. To avoid potential issues, I'll revert this change

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: @hajnalmt Could you please elaborate on your reasoning here as to why `Commit` is needed? Will `UnPipeline` be revertible like `Pipeline` in future designs? For now, I prefer not to add this call to eliminate unnecessary function calls 🤣

hajnalmt: Sorry, I didn’t manage to get back to this yesterday and I see you already removed the Commit/Discard branch, so I admit defeat 😄 Please don’t bring it back just for this.

I still wanted to add some architectural context, because this has been a recurring pattern for us. I already captured a broader direction in [#5045 Line 477](https://github.com/volcano-sh/volcano/pull/5045/changes#diff-379002309d8e9a8edb76f6b3827b52253a155c8ed0e2add7e80c51c27bc5c555R477) with Unpipeline as saved cache and statement operation and I’m hoping to get that PR ready this week.

What worries me is not this single line in isolation, but the overall consistency of statement handling. Most of the recent bugs in this area (#5010, #5067) happened around lifecycle boundaries: when statement operations are not clearly grouped and explicitly finalized with Commit()/Discard(), we get subtle rollback/atomicity issues later.

I understand that in `Commit()` around `UnPipeline()` is effectively a no-op today (so “commit what?” is fair). My concern is future-proofing: if/when `UnPipeline` becomes part of committable statement handling, call sites with unclear lifecycle boundaries can silently change behavior and without Discard and Commit we don't follow the clear design principles around the statement handling in general.
I’m +1 on moving forward here, I feel the push back against my architectural taste 😅 but I wanted to document this concern for follow-up, so we can keep statement semantics consistent across actions.

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -34,7 +34,14 @@ import (
...
 )
 
+type backfillContext struct {
+	queues      *util.PriorityQueue                 // queue of *api.QueueInfo
+	jobsByQueue map[api.QueueID]*util.PriorityQueue // queue of *api.JobInfo
+	tasksByJob  map[api.JobID]*util.PriorityQueue   // queue of *api.TaskInfo
+}
+
 type Action struct {
+	session                   *framework.Session
```
hzxuzhonghu, line 44: a minor concern, referencing session in this backfill object, it will prevent gc from freing the session until the new session is set

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
 			continue
 		}
 
+		tasks := util.NewPriorityQueue(ssn.TaskOrderFn)
```
hzxuzhonghu: nit: rename to taskQueue

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -60,68 +67,21 @@ func (backfill *Action) Execute(ssn *framework.Session) {
...
-		metrics.UpdateE2eSchedulingDurationByJob(job.Name, string(job.Queue), job.Namespace, metrics.Duration(job.CreationTimestamp.Time))
-		metrics.UpdateE2eSchedulingLastTimeByJob(job.Name, string(job.Queue), job.Namespace, time.Now())
+func (backfill *Action) buildBackfillContext() *backfillContext {
+	ssn := backfill.session
 
-		// TODO (k82cn): backfill for other case.
+	actx := &backfillContext{
+		queues:      util.NewPriorityQueue(ssn.QueueOrderFn),
+		jobsByQueue: map[api.QueueID]*util.PriorityQueue{},
+		tasksByJob:  map[api.JobID]*util.PriorityQueue{},
```
hzxuzhonghu:
```suggestion
		taskQueueByJob:  map[api.JobID]*util.PriorityQueue{},
```

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
+		if !found || jobs.Empty() {
 			continue
 		}
-		for !jobs[queue.UID].Empty() {
-			job, ok := jobs[queue.UID].Pop().(*api.JobInfo)
-			if !ok {
-				klog.Errorf("JobInfo transition failed, ignore it.")
-				continue
+
+		job := jobs.Pop().(*api.JobInfo)
```
hzxuzhonghu: previously it pops all the jobs until the jobQueue is empty, now it only pop one?

pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -137,65 +97,133 @@ func (backfill *Action) pickUpPendingTasks(ssn *framework.Session) []*api.TaskIn
...
+			}
+			fe.SetNodeError(node.Name, err)
+			job.NodesFitErrors[task.UID] = fe
+			continue
+		}
+
+		metrics.UpdateE2eSchedulingDurationByJob(job.Name, string(job.Queue), job.Namespace, metrics.Duration(job.CreationTimestamp.Time))
+		metrics.UpdateE2eSchedulingLastTimeByJob(job.Name, string(job.Queue), job.Namespace, time.Now())
+
+		if ssn.JobReady(job) {
```
hzxuzhonghu, line 223: sorry i donot catch up with here, breaking the loop means we donot allocate for other tasks of the job, this seems not right, we should try to schedule for all best effort tasks here

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: Yes, adopt a fair‑share manner here.
Imagine there are two queues, q1 and q2, each with two jobs (q1-j1, q1-j2, q2-j1, q2-j2).
The previous allocation sequence might be q1-j1, q1-j2, q2-j1, q2-j2,
while the current sequence is q1-j1, q2-j1, q1-j2, q2-j2.
Is this implementation reasonable, or do you prefer to keep the original allocation order?

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: i only adjust the allocation sequence here: first try to allocate resources to make all jobs ready, and then allocate resources to the extra best-effort tasks within ready jobs. The relevant code is as follows:
```
if len(stmt.Operations()) > 0 && ssn.JobReady(job) {
	stmt.Commit()
	if !tasks.Empty() { // remaining best effort tasks would be allocated in later loop
		jobs.Push(job)
	}
} else {
	stmt.Discard()
}
```

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: Would you prefer that I don’t store the session here, but instead pass it via method parameters?

hzxuzhonghu: right

pkg/scheduler/actions/backfill/backfill.go
hzxuzhonghu: Is there a case, this fall into a deadloop if the specific task can not be allocated?

pkg/scheduler/actions/backfill/backfill.go
hzxuzhonghu: prefer keep the original order because  we have already sorted the queues

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: > Is there a case, this fall into a deadloop if the specific task can not be allocated?

@hzxuzhonghu Actually no - if a specific task cannot be allocated, it will be popped from the queue and will no longer be pushed back into it.

pkg/scheduler/actions/backfill/backfill.go
hajnalmt: Please don't change this back, this is working correctly. 
The point here @hzxuzhonghu is that after a job has been processed the queue order can change, because the share changes for the queue (since allocated is more now).  `ssn.QueueOrder` will properly reprioritize that queue after it's pushed back if it has more jobs, and pop the right prioritezed Queue with a new job, so we process jobs one by one to adjust to the priority change.

hajnalmt: The right example here shows how the queue order changes because of task allocation, if we have 2 queues :
| Queue | Deserved CPU | Allocated CPU (initial) | Share (initial) |
|-------|-------------|------------------------|-----------------|
| Q1 | 100m | 10m | 10/100 = 0.10 |
| Q2 | 100m | 20m | 20/100 = 0.20 |

Initial queue order: Q1 (0.10) → Q2 (0.20) (Q1 popped first, lower share = higher priority)
- Q1 has two jobs: Q1-J1 (requests 50m CPU), Q1-J2 (requests 30m CPU)
- Q2 has two jobs: Q2-J1 (requests 40m CPU), Q2-J2 (requests 20m CPU)

Old Code (pop ALL jobs from queue, process them all)
Pop Q1 → process Q1-J1 and Q1-J2 back-to-back
Pop Q2 → process Q2-J1 and Q2-J2
Allocation sequence: Q1-J1 → Q1-J2 → Q2-J1 → Q2-J2

After Q1-J1 is allocated (50m), Q1's share is now 60/100 = 0.60. Higher than Q2's 0.20. But Q1-J2 still gets allocated next because we already popped Q1 and are draining all its jobs. Q2 is starved even though it should now have higher priority.

New Code (pop ONE job per queue, re-push queue, re-sort):

**Iteration 1:**
Queue order: Q1 (0.10) → Q2 (0.20)
Pop Q1 → allocate Q1-J1 (50m CPU)
Q1's share becomes: 60/100 = 0.60
Re-push Q1 into priority queue

**Iteration 2:**
Queue order: Q2 (0.20) → Q1 (0.60)    ← Q2 is now higher priority!
Pop Q2 → allocate Q2-J1 (40m CPU)
Q2's share becomes: 60/100 = 0.60
Re-push Q2
**Iteration 3:**
Queue order: Q1 (0.60) == Q2 (0.60)   (tie-broken arbitrarily)
Pop Q1 → allocate Q1-J2 (30m CPU)
Q1's share becomes: 90/100 = 0.90
Re-push Q1
**Iteration 4:**
Queue order: Q2 (0.60) → Q1 (0.90)
Pop Q2 → allocate Q2-J2 (20m CPU)
Allocation sequence: Q1-J1 → Q2-J1 → Q1-J2 → Q2-J2

This is the same pattern the allocate action already uses, pop one job, re-push the queue, let the priority queue re-sort.

pkg/scheduler/actions/backfill/backfill.go
hajnalmt: This is valid, but I wouldn't bother with it because we use the pattern elsewhere too, allocate and preempt does the same. Also the heavy lifting happening at [CloseSession(ssn)](https://github.com/volcano-sh/volcano/blob/release-1.14/pkg/scheduler/framework/framework.go) and [closeSession(ssn)](https://github.com/volcano-sh/volcano/blob/release-1.14/pkg/scheduler/framework/session.go#L597) clears internal maps and plugin state properly, so this is just a pointer in memory we keep between scheduling cycles. 

I agree that it would be nicer to clean this up, but we shall do it in a separate PR for all the plugins I think.

pkg/scheduler/actions/backfill/backfill.go
rayo1uo: @hajnalmt Really appreciate your detailed explanations!

hzxuzhonghu: @rayo1uo @hajnalmt

This way now is more fair, but what is the performance influence? I can see there is a degradation because of the increasing sort

pkg/scheduler/actions/backfill/backfill.go
hzxuzhonghu: This may cause performance issue too if there are may other tasks exceed minavailable,  we will re enueue many times and re run this loop https://github.com/volcano-sh/volcano/pull/5057/changes#diff-8600dd48131f07698f5fd074b3979a15a990a348cb433ea74a8b104aed378584R139

pkg/scheduler/actions/backfill/backfill.go
hajnalmt: There is definitely a performance decrease which this brings in, but it is **minor or even negligible**.

The priority queues use heap-based sorting (container/heap). Each Push/Pop is O(log n) where n is the number of queues or jobs (in this case queues). But n in practice is small: the number of queues in a cluster is typically single digits (1-10), and jobs per queue is in the low hundreds at most during a single scheduling cycle. (we filter this at several points, so Pending BestEffort tasks we are having here only.)

The old code was O(Q × J) with one flat pass. The new code is O(Q × log Q) for queue re-prioritization and O(J × log J) for job re-prioritization per queue. For small Q and J, I think this is inconsequential. The overhead is microseconds or less, this trade-off for fair-share queue ordering is well deserved.

pkg/scheduler/actions/backfill/backfill.go
hajnalmt: @hzxuzhonghu I think this is valid in theory, but negligible in practice.

Removing this will regress the fairness behavior, like in the previous case just only above minAvailable. For example if you have 2 jobs in different queues same priority (share), same task requests, both with minAvailable=2 but one with 100 best-effort tasks, the other one with just 3, then the first can get it's 98 task scheduled before the second one get's to schedule 1, possibly starving for a lot of time.

For T best-effort tasks beyond minAvailable, there are T extra iterations of the outer loop, each doing O(log Q + log J) priority queue operations. This means that for 100 extra tasks across 5 queues, avaraged 10 jobs per queue, that's ~100 × (log5 + log10) ≈ 100 × 5 = 500 extra comparisons. Still microseconds.

If you want to address the re-enqueue concern without losing fairness, a better approach would be batching: allocate multiple tasks per iteration (e.g., allocate up to N tasks after the job is ready before breaking) rather than breaking after every single task once ready. But that's an optimization for a later PR if profiling shows it matters.

pkg/scheduler/actions/backfill/backfill.go
hzxuzhonghu: microsecond increasing is not acceptable
Can we add  some bench tests before and after covering different Q/J? So many users are concerned about performance

pkg/scheduler/actions/backfill/backfill.go
hzxuzhonghu: Thanks for the deep looking, as above can we test with some constructed cases

rayo1uo: @hzxuzhonghu @hajnalmt 
I’ve run a benchmark here, and the related code can be found at https://github.com/rayo1uo/volcano/blob/fix_backfill_bench/pkg/scheduler/actions/backfill/backfill_benchmark_test.go. The results are as follows:

```shell
goos: darwin
goarch: arm64
pkg: volcano.sh/volcano/pkg/scheduler/actions/backfill
cpu: Apple M4 Pro
BenchmarkBackfillAllocateResources/q2_j4_t16/job_strategy/pop_all_jobs-14         	     236	    962353 ns/op	  668331 B/op	   10197 allocs/op
BenchmarkBackfillAllocateResources/q2_j4_t16/job_strategy/pop_one_job_requeue_queue-14         	     236	    979088 ns/op	  669209 B/op	   10216 allocs/op
BenchmarkBackfillAllocateResources/q2_j4_t16/task_strategy/pop_all_tasks-14                    	     285	    899015 ns/op	  611616 B/op	    8748 allocs/op
BenchmarkBackfillAllocateResources/q2_j4_t16/task_strategy/job_ready_break_requeue-14          	     229	   1018833 ns/op	  667623 B/op	   10183 allocs/op
BenchmarkBackfillAllocateResources/q4_j8_t16/job_strategy/pop_all_jobs-14                      	      54	   3853279 ns/op	 2758488 B/op	   42076 allocs/op
BenchmarkBackfillAllocateResources/q4_j8_t16/job_strategy/pop_one_job_requeue_queue-14         	      60	   3959456 ns/op	 2769588 B/op	   42217 allocs/op
BenchmarkBackfillAllocateResources/q4_j8_t16/task_strategy/pop_all_tasks-14                    	      68	   3592999 ns/op	 2459904 B/op	   35230 allocs/op
BenchmarkBackfillAllocateResources/q4_j8_t16/task_strategy/job_ready_break_requeue-14          	      58	   4904244 ns/op	 2677400 B/op	   40878 allocs/op
PASS
ok  	volcano.sh/volcano/pkg/scheduler/actions/backfill	488.108s
```

## Job Strategy

`pop_one_job_requeue_queue` is slightly slower compared to `pop_all_jobs`.

| Workload | pop_all_jobs | pop_one_job_requeue_queue | Change |
|----------|--------------|---------------------------|--------|
| q2_j4_t16 | 962353 ns/op | 979088 ns/op | ~1.7% slower (0.0167 ms slower) |
| q4_j8_t16 | 3853279 ns/op | 3959456 ns/op | ~2.8% slower (0.106 ms slower) |

---

## Task Strategy

`pop_all_tasks` significantly outperforms `job_ready_break_requeue`.

| Workload | pop_all_tasks | job_ready_break_requeue | Improvement |
|----------|---------------|-------------------------|-------------|
| q2_j4_t16 | 899015 ns/op | 1018833 ns/op | ~11.8% faster (0.120 ms faster) |
| q4_j8_t16 | 3592999 ns/op | 4904244 ns/op | ~26.7% faster (1.31 ms faster) |

hajnalmt: Whao cool benchmark. Nice. I think this is acceptable. Do I understand this properly, that the sample tasks count is 128 and 512? It's arguable that how right this is but in practice we won't have more even for really large clusters, we are speaking about best effort tasks only, and a single scheduling cycle.

rayo1uo: > Whao cool benchmark. Nice. I think this is acceptable. Do I understand this properly, that the sample tasks count is 128 and 512? It's arguable that how right this is but in practice we won't have more even for really large clusters, we are speaking about best effort tasks only, and a single scheduling cycle.

Yes, the sample tasks count is 128 and 512.

rayo1uo:
Hi @hzxuzhonghu @hajnalmt, could we proceed with merging this PR? If there are any pending requirements, please let me know what I can do to help move it forward.

hajnalmt:
From my side nothing to add the PR is fine. Maybe @JesseStutler can help with a final review too.

pkg/scheduler/actions/backfill/backfill.go
hzxuzhonghu: >pop_all_tasks significantly outperforms job_ready_break_requeue.

This performance degrade is bad,  26.7% for q4_j8_t16.  I guess the worse the job number increase

hajnalmt: I feel like we are sacrificing correct behaviour over performance here. (which I think is negligible, compared to eg. bind which takes a lot more time).

This reminds me of that old joke about the job interviews:
Interviewer: "It says here you're extremely fast at mathematics. What is 17*19?"
Candidate: "36!"
Interviewer: "That's not even close."
Candidate: "But it was fast

PR #5058, opened 2026-02-25
capacity plugin support DRA

xu-wentao:
#### What type of PR is this?
/kind feature

This PR introduces Dynamic Resource Allocation (DRA) quota support into the Volcano scheduler's [capacity] plugin. 
Key features include:
1. **DRA Quota Management**: Allows administrators to define logical quotas for any DRA devices (e.g., specific GPUs, RDMA) per Volcano queue. 
2. **Scheduling Enforcement**: Prevents jobs and tasks that exceed the queue's DRA capabilities from being enqueued or allocated. Supports both discrete device counts and consumable capacities (e.g., partial GPU memory allocation).
3. **Hierarchical Queue Support**: Accurately calculates and propagates DRA resource usage across queue hierarchies.
4. **Pass-through Mode**: For `DeviceClasses` that are not explicitly configured in a Queue's limits, the plugin operates in a "pass-through" mode, allowing tasks to use them without imposing quota restrictions.
This enhancement ensures fair and controlled distribution of customized DRA resources across multitenant clusters.

Fixes #

- The DRA quota and consumable capacity checks are controlled by the `DynamicResourceAllocationEnable` and `DRAConsumableCapacityEnable` feature flags/plugin arguments.

```release-note
Support Dynamic Resource Allocation (DRA) resource quota management in the capacity scheduling plugin.

### CODE REVIEW

**Comment by JesseStutler** [comment]
Would you be willing to share on tomorrow's Asia community meeting? The meeting time is 15:00 UTC+8

**Comment by JesseStutler** [comment]
@Xu-Wentao Any update on the PR? :)

**Comment by Xu-Wentao** [comment]
> @Xu-Wentao Any update on the PR? :)

Yes, I have updated the implemention of the DRA metioned  [here](https://www.notion.so/xuwt/Volcano-Queue-DRA-31bca50a6ba48109a210e9e2461648d3), we can discuss thie design again.

**Comment by JesseStutler** [comment]
Please resolve the CI failure and sign off your commits, and there are also conflict codes that you need to resolve @Xu-Wentao

File: process.md
Comment (JesseStutler) [inline process.md:1]: Please remove this file

File: .gitignore
Comment (JesseStutler) [inline .gitignore:1]: Why do add a `volcano.sh` ignore item?

File: process.md
Comment (Xu-Wentao) [reply process.md:1]: removed

File: .gitignore
Comment (Xu-Wentao) [reply .gitignore:1]: The folder `volcano.sh`  is not in this project, but if we run `make generate-code` will create this folder automatilly, If we no need to upload it, maybe ignore it is a way i think?

File: pkg/scheduler/plugins/capacity/capacity.go [hunk 1]
```diff
@@ -69,17 +87,314 @@ type queueAttr struct {
...
+	}
 }
 
 // New return capacityPlugin action
 func New(arguments framework.Arguments) framework.Plugin {
+	// Default to k8s feature gate values, allow override via plugin arguments
+	dynamicResourceAllocationEnable := utilfeature.DefaultFeatureGate.Enabled(kubefeatures.DynamicResourceAllocation)
+	draConsumableCapacityEnable := utilfeature.DefaultFeatureGate.Enabled(kubefeatures.DRAConsumableCapacity)
+
+	arguments.GetBool(&dynamicResourceAllocationEnable, DynamicResourceAllocationEnable)
```
JesseStutler, line 388: Should sync to the user guide to allow users know there are such arguments for capacity plugins, default is controlled by feature gate values

docs/user-guide/how_to_use_dra_quota.md
```diff
@@ -0,0 +1,318 @@
...
+
+```yaml
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: vgpu-team
+spec:
+  reclaimable: true
+  capability:
+    "cores.deviceclass/hami-core-gpu.project-hami.io": "800"
```
JesseStutler, line 101: It would be best to further clarify the scenario for `consumable capacity` here. If `consumable capacity` is specified, then `[dims].deviceclass/xxx` should be configured as `count * capacity`, right? This could be explained to guide users.

pkg/scheduler/plugins/capacity/capacity.go
JesseStutler, line 1: I think you should also update the inqueued attr when the pg can be enqueued (especially for simultaneously submit jobs it's useful): https://github.com/volcano-sh/volcano/blob/da719bb3b636ecce560798d9ec0f19c74e3b412c/pkg/scheduler/plugins/capacity/capacity. go#L271

docs/design/capacity-dra-support.md
```diff
@@ -0,0 +1,188 @@
...
+
+- Managing physical device placement or allocation details
+- Replacing Kubernetes DRA scheduling logic
+- Defining a new DRA-specific queue CRD structure
+
+## Core Idea
+
+DRA quota is expressed directly in the queue `ResourceList` using reserved key formats.
+
+### Key Formats
```
JesseStutler, line 33: In the scenario of consumable capacity, I may have a worry. For example, if a card has virtualization, then one ResourceSlice corresponds to multiple slices inside, right? So currently your design is flattened at the deviceclass level, regardless of how many slices a ResourceSlice has, we configure a total amount. For instance, if there are 2 cards (in the same ResourceSlice), each card has 8Gi of GPU memory, and the total I configure on the queue is 16Gi. If I configure 16Gi at the deviceclass level on the queue, and a podgroup requests 2 pods with 10Gi + 4Gi respectively, then one pod actually cannot be scheduled. However, it gets through the queue validation and occupies inqueue resources. Is this acceptable?

I think the current design is fine, but I believe there might be a potential issue, similar to how node had fragmented resources before. However, the queue check still passed, but I don't know if users would report such a problem. I think it could be stated as a constraint in the design doc or user guide doc

pkg/scheduler/api/job_info.go
```diff
@@ -1326,3 +1348,146 @@ func (ji *JobInfo) ContainsNetworkTopologyInSubJob() bool {
...
+			}
+
+			result[deviceClass].Count += res.Count * int64(minNum)
+			for dim, cap := range res.Capacity {
+				totalCap := cap.DeepCopy()
+				// resource.Quantity has no Multiply func, so we parse memory/cpu as MilliValues
+				// For exact values we can just use set
+				// Since Quantity can represent fractional, we will loop to add
+				for i := int32(0); i < minNum-1; i++ {
+					totalCap.Add(cap)
```
JesseStutler, line 1476: For Consumable Capacity, it should also be multiplied by count to get the total capacity resource, right? For example, a single pod requires 2 * 8Gi of GPU memory, and there are 2 pods in total, so the total requirement should be 32Gi, not 8Gi * 2 pods = 16Gi?

pkg/scheduler/cache/cache_dra_test.go
```diff
@@ -0,0 +1,245 @@
+/*
+Copyright 2024 The Volcano Authors.
```
JesseStutler, line 2: All should change to -> `2026 The Volcano Authors`

test/e2e/dra/dra_quota_test.go
JesseStutler, line 1: Can we construct a Driver that can produce ResourceSlices containing multiple devices, each device with a capacity? I'd like to see such use case if it's possible

JesseStutler:
Please also resolve the conflicts and sign off the commits
BTW do you have screenshots of the actual successful allocation report, for example: 
1. The queue quota can actually block the pod, preventing it from being scheduled.
2. Successfully assigned, the queue status displays the status after dra assignment. 
Would be better to see the actual effect :D
@Xu-Wentao

JesseStutler:
@xu-wentao Looks like the UT still has failures, please check

JesseStutler:
Could you check the CI failure? @xu-wentao I have rerun it multiple times, but there are still these two failures, one DRA may be running to much time, one is related with 
image
I'm thinking that whether this PR introduce some change

JesseStutler:
Let me run again and check

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -75,18 +92,334 @@ type queueAttr struct {
...
+				capQty, ok := capability.Capacity[dim]
+				if !ok {
+					return false // dimension not in capability
+				}
+				var allocQty resource.Quantity
+				if allocated != nil {
+					allocQty = allocated.Capacity[dim]
+				}
+				futureUsed := allocQty.DeepCopy()
+				if inqueue != nil {
```
JesseStutler, line 292: Inqueue resource should be only counted in the Enqueueable phase, not in the allocatable, here inqueue + futureused count resources double time

JesseStutler: I think that's why the DRA canceled, new introduced DRA quota e2e tests was actually failed, I think it's maybe related to the pod was actually not scheduled, blocked by Allocatable judgement here

hack/e2e-kind-config.yaml
```diff
@@ -18,14 +18,52 @@ nodes:
...
           extraArgs:
             runtime-config: "admissionregistration.k8s.io/v1beta1"
             enable-admission-plugins: "MutatingAdmissionPolicy"
+            feature-gates: "MutatingAdmissionPolicy=true,DynamicResourceAllocation=true,DRAConsumableCapacity=true"
+        controllerManager:
+          extraArgs:
+            feature-gates: "MutatingAdmissionPolicy=true,DynamicResourceAllocation=true,DRAConsumableCapacity=true"
+        scheduler:
+          extraArgs:
+            feature-gates: "MutatingAdmissionPolicy=true,DynamicResourceAllocation=true,DRAConsumableCapacity=true"
```
JesseStutler: It open kube-scheduler's feature gate, not for Volcano scheduler, you should open feature gate for Volcano instead: https://github.com/volcano-sh/volcano/blob/4687fce2dc41d3e5f56d38f34db559fe9f725ff0/Makefile#L227

pkg/scheduler/cache/cache.go
```diff
@@ -1823,3 +1827,121 @@ func (sc *SchedulerCache) OnSessionClose() {
 		sc.notifySessionEnd()
 	}
 }
+
+func resolvePodClaimName(pod *v1.Pod, podClaim v1.PodResourceClaim) string {
+	if podClaim.ResourceClaimName != nil {
+		return *podClaim.ResourceClaimName
+	}
+	for _, status := range pod.Status.ResourceClaimStatuses {
```
JesseStutler, line 1842: If the pod uses resourceclaim template, the resource claim may not yet created and synced to the scheduler's cache yet, and it may cause current resolved pod claim name is empty, which will cause an error when NewTaskInfo, add the resync may fail when the task has not been added to the cache yet: https://github.com/volcano-sh/volcano/blob/4687fce2dc41d3e5f56d38f34db559fe9f725ff0/pkg/scheduler/cache/cache.go#L1233

JesseStutler:
I think the scheduling action CI failure is not related with this PR, my PR can already fix it and I have tested in my local env: https://github.com/volcano-sh/volcano/pull/5319, just resolve all of my comments and then we can get merged

JesseStutler:
There is a conflict @xu-wentao Please rebase to the latest codes and push again, thanks

JesseStutler:
/ok-to-test
Let me first resolve it and trigger the CIs, still needs to force push again in your local

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1375,8 +1436,9 @@ func (cp *capacityPlugin) buildQueueReservedTasksCache(ssn *framework.Session) {
 }
 
 func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidate *api.TaskInfo, queue *api.QueueInfo, draEnabled bool, consumableCapacityEnabled bool) bool {
-	if draEnabled && attr.dra != nil && candidate.DRAResreq != nil {
-		if !checkDRAAllocatable(attr.dra, candidate.DRAResreq, consumableCapacityEnabled) {
+	candidateDRA := incrementalTaskDRA(attr, candidate)
```
JesseStutler: Put this line inside the draEnabled judgement is better

JesseStutler:
https://github.com/volcano-sh/volcano/pull/5058#discussion_r3232936051 I think you should still open the feature gate for the volcano scheduler:
```
E2E_TYPE=DRA FEATURE_GATES="DynamicResourceAllocation=true,DRAConsumableCapacity=true" ./hack/run-e2e-kind.sh 
```
Only open the feature gate in the kind config is opened for the kube-scheduler, not for the volcano scheduler @xu-wentao

JesseStutler:
/approve
/lgtm
Thank you so much for your great work!

PR #5064, opened 2026-02-27
fix: exclude Succeeded tasks from gang scheduling readiness gate

Aman-Cool:
#### What type of PR is this?

/kind bug

`ReadyTaskNum()` was counting `Succeeded` pods as "ready" slots. This let `stmt.Commit()` fire with fewer than `MinAvailable` *actually running* tasks, breaking gang atomicity.

Similar to #5011 — another case where we committed scheduling decisions based on incorrect readiness state.

The fix: only count pods that hold resources (Bound, Binding, Running, Allocated). Dead pods don't count.

Also fixed `SubJobInfo.ReadyTaskNum()` and `getJobAllocatedRoles()` for role-level constraints.

Tests that encoded the buggy behavior were updated — they now use `Running` instead of `Succeeded` to mean "task is active."

`ValidTaskNum()`, `CheckTaskPipelined()`, and `CheckTaskStarving()` are intentionally unchanged. They use a broader "has this job ever had enough tasks" semantic that correctly includes Succeeded.

```release-note
Fixed a bug where gang scheduling could commit partial task groups when some tasks had already succeeded, causing jobs to fail silently after multiple restart loops.
```

### DIFF
--- a/pkg/scheduler/api/job_info_test.go
+++ b/pkg/scheduler/api/job_info_test.go
@@ -718,3 +718,167 @@ func TestParseMinMemberInfoChanged(t *testing.T) {
 		})
 	}
 }
+
+// TestReadyTaskNum verifies that ReadyTaskNum only counts tasks with live
+// scheduler allocations (Allocated, Binding, Bound, Running) and explicitly
+// excludes Succeeded tasks, which hold no resources and must not ghost-satisfy
+// the gang scheduling commit gate.
+func TestReadyTaskNum(t *testing.T) {
+	tests := []struct {
+		name     string
+		ji       *JobInfo
+		expected int32
+	}{
+		{
+			name: "only Succeeded tasks do not count",
+			ji: &JobInfo{
+				TaskStatusIndex: map[TaskStatus]TasksMap{
+					Succeeded: {
+						"task1": &TaskInfo{},
+						"task2": &TaskInfo{},
+					},
+				},
+			},
+			expected: 0,
+		},
+		{
+			name: "Succeeded mixed with active tasks - only active counted",
+			ji: &JobInfo{
+				TaskStatusIndex: map[TaskStatus]TasksMap{
+					Succeeded: {"task1": &TaskInfo{}},
+					Running:   {"task2": &TaskInfo{}},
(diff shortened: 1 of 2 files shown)


Aman-Cool:
@hajnalmt @hzxuzhonghu  ,Same flavor as #5011 ; committing scheduling decisions based on incorrect readiness. Succeeded pods were ghost-satisfying the gang gate, causing partial commits and restart loops. Fix: only count active pods (Bound/Binding/Running/Allocated) in ReadyTaskNum().

Aman-Cool:
@JesseStutler, Can you have a look too? Thanks!

Aman-Cool:
Added 5 regression tests covering both ReadyTaskNum() and getJobAllocatedRoles(). Each test fails on the old buggy code (Succeeded counts toward readiness) and passes on the new code (doesn't count). Should cover the gaps @copilot code review[agent]  flagged.

Aman-Cool:
Committed all three suggestions, docstring wording fixes and the formatting nit.

hajnalmt (commented):
Hello @Aman-Cool ,
Thanks for the PR! Why do you think Succeeded tasks should omitted from the ready function? A task that has already finished should be counted in minAvailable, and should be considered ready I think. I understand, that it won't be scheduled and gang constraints are containing it, but I don't see why this is a problem.

hzxuzhonghu (commented):
Succeeded task still hold the resource allocated, this change could break the gang behavior

Aman-Cool:
@hajnalmt @hzxuzhonghu ,Thank you for the detailed review, you guys are right, and I misidentified this as a bug.
The scenario you describe (some tasks completing normally while others get evicted and need to restart) is a real and common pattern. Removing Succeeded from ReadyTaskNum would permanently block evicted tasks from rescheduling once any peer task has completed, causing jobs to hang indefinitely.
The current behavior, counting Succeeded pods toward the readiness gate, is correct: it allows recovery of individual evicted tasks without requiring a full re-gang, while the initial gang constraint still guarantees that all minAvailable pods are co-scheduled at launch.
I'll close this PR.

PR #5069, opened 2026-03-02
scheduler: call Initialize() and UnInitialize() in action lifecycle

### LINKED ISSUE #5052: scheduler: initialize and uninitialize are not called in the codebase
## Summary

The functions `Initialize()` or `Uninitialize()`, defined in the scheduler actions **are not called anywhere** in the codebase, except for tests:

https://github.com/volcano-sh/volcano/blob/49b8f40ba6247b9a6a65e14e6e14c4fb19feb77c/pkg/scheduler/uthelper/helper.go#L224-L229

This prevents features that require initialization logic from executing, as suggested by https://github.com/volcano-sh/volcano/pull/5033#discussion_r2833051128.

Some references of these functions include:

```go
func (alloc *Action) Initialize() {}
func (backfill *Action) Initialize() {}
func (enqueue *Action) Initialize() {}
func (pmpt *Action) Initialize() {}
```

aadhil2k4:
#### What type of PR is this?

bugfix

- Currently, scheduler actions only invoke Execute() during a scheduling cycle.
- Initialize() and UnInitialize() are defined in the Action interface but are called only in unit tests and never in production code.
- This change ensures every scheduler follows its full lifecycle - Initialize(), Execute(), UnInitialize().

Fixes #5052 

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -86,10 +86,39 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
 	return scheduler, nil
 }
 
+// Initialize all configured actions once.
+func (pc *Scheduler) initActions() {
+	pc.mutex.Lock()
+	actions := pc.actions
+	pc.mutex.Unlock()
+	for _, action := range actions {
+		action.Initialize()
+	}
+}
+
+// Uninitialize all configured actions.
+func (pc *Scheduler) cleanupActions() {
+	pc.mutex.Lock()
+	actions := pc.actions
+	pc.mutex.Unlock()
+	for _, action := range actions {
+		action.UnInitialize()
+	}
+}
+
 // Run initializes and starts the Scheduler. It loads the configuration,
 // initializes the cache, and begins the scheduling process.
 func (pc *Scheduler) Run(stopCh <-chan struct{}) {
 	pc.loadSchedulerConf()
+
+	pc.initActions()
+
+	// Cleanup on scheduler shutdown.
+	go func() {
+		<-stopCh
+		pc.cleanupActions()
+	}()
+
 	go pc.watchSchedulerConf(stopCh)
 	// Start cache for policy.
 	pc.cache.SetMetricsConf(pc.metricsConf)
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -52,11 +52,17 @@ type Scheduler struct {
 	schedulePeriod time.Duration
 	once           sync.Once
 
-	mutex              sync.Mutex
-	actions            []framework.Action
-	plugins            []conf.Tier
-	configurations     []conf.Configuration
-	metricsConf        map[string]string
+	mutex          sync.Mutex
+	actions        []framework.Action
+	plugins        []conf.Tier
+	configurations []conf.Configuration
+	metricsConf    map[string]string
+
+	defaultActions        []framework.Action
+	defaultPlugins        []conf.Tier
+	defaultConfigurations []conf.Configuration
+	defaultMetricsConf    map[string]string
+
 	dumper             schedcache.Dumper
 	disableDefaultConf bool
 }
@@ -86,35 +92,34 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
 	return scheduler, nil
 }
 
-// Initialize all configured actions once.
-func (pc *Scheduler) initActions() {
-	pc.mutex.Lock()
-	actions := pc.actions
-	pc.mutex.Unlock()
+// Run initializes and starts the Scheduler. It parses the scheduler configuration, initializes actions,
+// applies the configuration and starts the scheduling loop, cache and configuration watcher.
+func (pc *Scheduler) Run(stopCh <-chan struct{}) {
+	actions, plugins, configurations, metricsConf, err := pc.loadSchedulerConf()
+	if err != nil {
+		klog.Fatalf("Failed to load scheduler configuration: %v", err)
+	}
+
 	for _, action := range actions {
 		action.Initialize()
 	}
-}
 
-// Uninitialize all configured actions.
-func (pc *Scheduler) cleanupActions() {
 	pc.mutex.Lock()
-	actions := pc.actions
+	pc.actions = actions
+	pc.plugins = plugins
+	pc.configurations = configurations
+	pc.metricsConf = metricsConf
 	pc.mutex.Unlock()
-	for _, action := range actions {
-		action.UnInitialize()
-	}
-}
-
-// Run initializes and starts the Scheduler. It loads the configuration,
-// initializes the cache, and begins the scheduling process.
-func (pc *Scheduler) Run(stopCh <-chan struct{}) {
-	pc.loadSchedulerConf()
 
 	// Cleanup on scheduler shutdown.
 	go func() {
 		<-stopCh
-		pc.cleanupActions()
+		pc.mutex.Lock()
+		oldActions := append([]framework.Action(nil), pc.actions...)
+		pc.mutex.Unlock()
+		for _, action := range oldActions {
+			action.UnInitialize()
+		}
 	}()
 
 	go pc.watchSchedulerConf(stopCh)
@@ -161,12 +166,8 @@ func (pc *Scheduler) runOnce() {
 	}
 }
 
-func (pc *Scheduler) loadSchedulerConf() {
+func (pc *Scheduler) loadSchedulerConf() ([]framework.Action, []conf.Tier, []conf.Configuration, map[string]string, error) {
 	klog.V(4).Infof("Start loadSchedulerConf ...")
-	defer func() {
-		actions, plugins := pc.getSchedulerConf()
-		klog.V(2).Infof("Finished loading scheduler config. Final state: actions=%v, plugins=%v", actions, plugins)
-	}()
 
 	if pc.disableDefaultConf && len(pc.schedulerConf) == 0 {
 		klog.Fatalf("No --scheduler-conf path provided and default configuration fallback is disabled")
@@ -175,7 +176,7 @@ func (pc *Scheduler) loadSchedulerConf() {
 	var err error
 	if !pc.disableDefaultConf {
 		pc.once.Do(func() {
-			pc.actions, pc.plugins, pc.configurations, pc.metricsConf, err = UnmarshalSchedulerConf(DefaultSchedulerConf)
+			pc.defaultActions, pc.defaultPlugins, pc.defaultConfigurations, pc.defaultMetricsConf, err = UnmarshalSchedulerConf(DefaultSchedulerConf)
 			if err != nil {
 				klog.Fatalf("Invalid default configuration: unmarshal Scheduler config %s failed: %v", DefaultSchedulerConf, err)
 			}
@@ -189,51 +190,27 @@ func (pc *Scheduler) loadSchedulerConf() {
 			if pc.disableDefaultConf {
 				klog.Fatalf("Failed to read scheduler config and default configuration fallback is disabled")
 			}
-			klog.Errorf("Failed to read the Scheduler config in '%s', using previous configuration: %v",
+			klog.Errorf("Failed to read the Scheduler config in '%s', falling back to default: %v",
 				pc.schedulerConf, err)
-			return
+			return pc.defaultActions, pc.defaultPlugins, pc.defaultConfigurations, pc.defaultMetricsConf, nil
 		}
 		config = strings.TrimSpace(string(confData))
 	}
 
-	actions, plugins, configurations, metricsConf, err := UnmarshalSchedulerConf(config)
-	if err != nil {
-		if pc.disableDefaultConf {
-			klog.Fatalf("Invalid scheduler configuration and default configuration fallback is disabled")
+	if len(config) != 0 {
+		actions, plugins, configurations, metricsConf, err := UnmarshalSchedulerConf(config)
+		if err != nil {
+			if pc.disableDefaultConf {
+				klog.Fatalf("Invalid scheduler configuration and default configuration fallback is disabled")
+			}
+			klog.Errorf("Scheduler config %s is invalid, falling back to default: %v", config, err)
+			return pc.defaultActions, pc.defaultPlugins, pc.defaultConfigurations, pc.defaultMetricsConf, nil
 		}
-		klog.Errorf("Scheduler config %s is invalid: %v", config, err)
-		return
+		klog.V(2).Infof("Loaded scheduler configuration from file")
+		return actions, plugins, configurations, metricsConf, nil
 	}
-
-	pc.mutex.Lock()
-	oldActions := pc.actions
-	pc.actions = actions
-	pc.plugins = plugins
-	pc.configurations = configurations
-	pc.metricsConf = metricsConf
-	pc.mutex.Unlock()
-
-	// Cleanup old actions
-	for _, action := range oldActions {
-		action.UnInitialize()
-	}
-
-	// Initialize new actions
-	for _, action := range actions {
-		action.Initialize()
-	}
-}
-
-func (pc *Scheduler) getSchedulerConf() (actions []string, plugins []string) {
-	for _, action := range pc.actions {
-		actions = append(actions, action.Name())
-	}
-	for _, tier := range pc.plugins {
-		for _, plugin := range tier.Plugins {
-			plugins = append(plugins, plugin.Name)
-		}
-	}
-	return
+	klog.V(2).Infof("Using default scheduler configuration")
+	return pc.defaultActions, pc.defaultPlugins, pc.defaultConfigurations, pc.defaultMetricsConf, nil
 }
 
 func (pc *Scheduler) watchSchedulerConf(stopCh <-chan struct{}) {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -102,6 +102,20 @@ func (pc *Scheduler) Run(stopCh <-chan struct{}) {
 	go runSchedulerSocket()
 }
 
+// executeAction runs a single scheduler action with its full lifecycle,
+// ensuring Initialize() is called before Execute() and UnInitialize()
+// is deferred for guaranteed cleanup after execution.
+func (pc *Scheduler) executeAction(action framework.Action, ssn *framework.Session) {
+	func() {
```
hajnalmt: Why do we have this unnecessary anonymous function wrapper here? It doesn't do anything.

pkg/scheduler/scheduler.go
```diff
@@ -102,6 +102,20 @@ func (pc *Scheduler) Run(stopCh <-chan struct{}) {
...
+		action.Initialize()
```
hajnalmt: We have a conflicting lifecycle assumptions here. The way devzizu imagined this at [L827](
https://github.com/volcano-sh/volcano/pull/5033/changes#diff-3dcbd15d9e1bf118608b4eedcaadc5cc7bdb9f60b482ad0ca7238ada5ad9e32cL827) is that `Initialize()` and `UnInitialize()` will be called once for the scheduler's lifetime (start/shutdown), not per-cycle.

I think I agree with him, since  session is not an argument of this function the initial design thought should have been to call it only once. We have OpenSession and CloseSession to adhere for a session level lifecycle.

hajnalmt (changes requested):
Thanks for the PR!
Can you align with @devzizu in the original issue or here please? You have completely different lifecycle assumptions than him about the introduction of these functions, and I lean towards his way.

I want to mention that the implementation is not easy on this if we move on to the scheduler's lifetime path. Because you need to account for hot-reload as on config reload `pc.actions` gets replaced with fresh instances, so you need to prevent the old actions' workers to leak.

pkg/scheduler/scheduler.go
devzizu: Indeed, the absence of `session` as a parameter to `Initialize()`/`UnInitialize()` is a good signal that they were designed for scheduler-lifetime scope, not per-cycle. Per-cycle setup already has its natural place, _i.e._ `Execute(ssn)` itself, or `OnSessionOpen`/`OnSessionClose` on the plugin side (which do receive the session). 

As @hajnalmt is stating, calling them per-cycle would also conflict with the async worker design proposed in https://github.com/volcano-sh/volcano/pull/5033, since workers are meant to run for the scheduler's lifetime.

@aadhil2k4 @hajnalmt, I suggest we move the logic to https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/scheduler.go#L91? There, we can iterate over the actions and call `Initialize()`, while `UnInitialize()` can be called by consuming `stopCh`?

hzxuzhonghu: I donot see any action implement Initialize now.

hajnalmt: @hzxuzhonghu 
This is needed for devzizu [here](https://github.com/volcano-sh/volcano/pull/5033/changes#diff-3dcbd15d9e1bf118608b4eedcaadc5cc7bdb9f60b482ad0ca7238ada5ad9e32cR964) currently we don't have a way for the actions to gracefully initilize and unitilize workers spawned by any action. We decided to do this separately because we need to support this I think.

aadhil2k4:
Hi!! I was occupied last week and couldnt check on this. I'll implement the changes and let you know.

aadhil2k4:
@hajnalmt @devzizu I have moved the lifecycle handling from per-cycle execution to the scheduler lifetime as discussed.
Please confirm if this implementation looks correct? If so, I’ll proceed with implementing the hot-reload handling.

pkg/scheduler/scheduler.go
```diff
@@ -86,10 +86,37 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
 	return scheduler, nil
 }
 
+// Initialize all configured actions once.
+func (pc *Scheduler) initActions() {
```
devzizu: It looks like `initActions()` is defined but only called in the tests, and could be called in `Run()` as well right?

Additionally, I believe we should separate concerns between `loadSchedulerConf` and `Run()`, but let me know what you think here:

- Keep `loadSchedulerConf` doing only what it does today (parse + swap).
- `Run()` can call `initActions()` explicitly after the first `loadSchedulerConf()`.
- `watchSchedulerConf` (hot-reload path) explicitly calls `UnInitialize()` on old actions and `Initialize()` on new ones before swapping them into `pc.actions` under the `mutex`, so `runOnce()` never sees uninitialized actions.

pkg/scheduler/scheduler.go
hzxuzhonghu: ACK, i agree with only calling init/unInit once liftime rather than per cycle

pkg/scheduler/scheduler.go
```diff
@@ -86,10 +86,37 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
...
+	pc.mutex.Lock()
+	actions := pc.actions
+	pc.mutex.Unlock()
+	for _, action := range actions {
+		action.Initialize()
+	}
+}
+
+// Uninitialize all configured actions.
+func (pc *Scheduler) cleanupActions() {
```
hzxuzhonghu: move to test file if only needed there and no need to make it as method of `Scheduler`

pkg/scheduler/scheduler.go
hzxuzhonghu: donot we need to call init for these actions too?

pkg/scheduler/scheduler.go
aadhil2k4: Thanks for the review. This design choice is much better and cleaner. I'll implement it and let you know.

aadhil2k4: @devzizu The hot-reload sequence has to call `Initialize()` on new actions before they are swapped into `pc.actions`.  But if `loadSchedulerConf` continues to operate the swap internally, then by the time `watchSchedulerConf` can call `Initialize()` on new actions, they will already be live in `pc.actions`. 

Rather than letting `loadSchedulerConf` do the parsing and swapping, can we modify it to only parse the value and return it to the caller. This would give `watchSchedulerConf` control over when the swap happens.

Does this look good to you?

pkg/scheduler/scheduler.go
aadhil2k4: Sure. I'll move it to the test file and implement the logic inline.

pkg/scheduler/scheduler.go
aadhil2k4: Yes, I missed it in the changes. I will revert it back and call it from `Run()` like it was before.

pkg/scheduler/scheduler.go
aadhil2k4: @devzizu @hajnalmt does the above approach sound good to you?

devzizu: @aadhil2k4 sorry for the late response -- Yes, that approach looks good to me, go ahead with it.

hzxuzhonghu:
@aadhil2k4 Would you like to address the comments

aadhil2k4:
@hzxuzhonghu @devzizu I have implemented the changes as suggested in the comments. Please check it and let me know if any changes are needed.

devzizu:
LGTM @aadhil2k4, thank you

pkg/scheduler/scheduler.go
```diff
@@ -86,10 +92,36 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
...
+	pc.plugins = plugins
+	pc.configurations = configurations
+	pc.metricsConf = metricsConf
+	pc.mutex.Unlock()
+
+	// Cleanup on scheduler shutdown.
+	go func() {
+		<-stopCh
+		pc.mutex.Lock()
+		oldActions := append([]framework.Action(nil), pc.actions...)
```
JesseStutler, line 118: There is already wrap func `slices.Clone` that you can directly use

pkg/scheduler/scheduler.go
```diff
@@ -162,40 +190,27 @@ func (pc *Scheduler) loadSchedulerConf() {
...
-
-	pc.mutex.Lock()
-	pc.actions = actions
-	pc.plugins = plugins
-	pc.configurations = configurations
-	pc.metricsConf = metricsConf
-	pc.mutex.Unlock()
-}
-
-func (pc *Scheduler) getSchedulerConf() (actions []string, plugins []string) {
```
JesseStutler, line 189: We still need to print the actions and plugins we loaded

pkg/scheduler/scheduler_test.go
JesseStutler, line 1: The ut tests is not so good, at least you should cover `watchSchedulerConf` to make sure that new scheduler configuration can be successfully reloaded and action can be initialized, and cover `loadSchedulerConf`

pkg/scheduler/scheduler.go
```diff
@@ -212,7 +227,24 @@ func (pc *Scheduler) watchSchedulerConf(stopCh <-chan struct{}) {
...
+					action.Initialize()
+				}
+				pc.mutex.Lock()
+				oldActions := append([]framework.Action(nil), pc.actions...)
+				pc.actions = actions
+				pc.plugins = plugins
+				pc.configurations = configurations
+				pc.metricsConf = metricsConf
+				pc.mutex.Unlock()
+				for _, action := range oldActions {
```
JesseStutler, line 245: If the actions before and after remain unchanged, there's no need to execute UnInitialize and Initialize again. If the actions before and after remain unchanged, you're actually dealing with the same Action pointer, and the data structure inside is consistent. If you Initialize first and then Uninitialize, there might be logical issues. What I expect is that Uninitialize should only be executed when the scheduler exits gracefully, or when the Action is no longer needed. Therefore here, Actions should be diffed

Also can you wrap this func? I think it has duplicate code with loadSchedulerConf @aadhil2k4

pkg/scheduler/scheduler_test.go
aadhil2k4: Sure, I'll widen the test's coverage. My primary focus was to get the implementation right.

pkg/scheduler/scheduler.go
aadhil2k4: Sure. I'll do it.

aadhil2k4:
@JesseStutler I have made the said changes and also added new tests. PTAL and let me know if something needs to improved.

pkg/scheduler/scheduler.go
```diff
@@ -52,11 +53,17 @@ type Scheduler struct {
...
+	mutex          sync.Mutex
+	actions        []framework.Action
+	plugins        []conf.Tier
+	configurations []conf.Configuration
+	metricsConf    map[string]string
+
+	defaultActions        []framework.Action
+	defaultPlugins        []conf.Tier
+	defaultConfigurations []conf.Configuration
+	defaultMetricsConf    map[string]string
```
hzxuzhonghu, line 65: i donot quite catchup with the recent change related. Why separate/record default configuration?

pkg/scheduler/scheduler.go
```diff
@@ -86,13 +93,53 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
...
+	pc.plugins = plugins
+	pc.configurations = configurations
+	pc.metricsConf = metricsConf
+	pc.mutex.Unlock()
+
+	if pc.cache != nil {
+		pc.cache.SetMetricsConf(pc.metricsConf)
+	}
+
+	actionNames, pluginNames := pc.getSchedulerConf()
```
hzxuzhonghu, line 113: why not directly use passed in

pkg/scheduler/scheduler.go
```diff
@@ -86,13 +93,53 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
 	return scheduler, nil
 }
 
-// Run initializes and starts the Scheduler. It loads the configuration,
-// initializes the cache, and begins the scheduling process.
+func (pc *Scheduler) applySchedulerConf(
```
hzxuzhonghu, line 96: I would move this into loadSchedulerConf to eliminate the confusion just looking at the function name

pkg/scheduler/scheduler.go
```diff
@@ -212,8 +262,28 @@ func (pc *Scheduler) watchSchedulerConf(stopCh <-chan struct{}) {
...
+				for name, action := range newMap {
+					if _, exists := oldMap[name]; !exists {
+						action.Initialize()
+					}
+				}
+				pc.applySchedulerConf(actions, plugins, configurations, metricsConf)
+				for name, action := range oldMap {
+					if _, exists := newMap[name]; !exists {
+						action.UnInitialize()
+					}
```
hzxuzhonghu, line 285: also move into the `loadSchedulerConf`

pkg/scheduler/scheduler.go
aadhil2k4: The functionality of `loadSchedulerConf` is changed. Previously, it used to parse + swap. Now it only parses and returns the values. The default configuration is parsed once and stored separately so it can serve as a stable fallback and avoid reinitializing defaults during reload.

pkg/scheduler/scheduler.go
aadhil2k4: @JesseStutler asked me to wrap the duplicate code in a func. Both `watchSchedulerConf` and `Run` had duplicate code snippets.

pkg/scheduler/scheduler.go
aadhil2k4: I used `getSchedulerConf()` to log the state after it is actually applied to the scheduler, so the log reflects the real internal state. We can use the passed in values also as they are assigned just above.

pkg/scheduler/scheduler.go
aadhil2k4: `loadSchedulerConf` now only parses and returns values. This code is for handling changes during runtime reload.

hzxuzhonghu: I fonot feel comfortable with the code style.

pkg/scheduler/scheduler.go
```diff
@@ -162,28 +213,27 @@ func (pc *Scheduler) loadSchedulerConf() {
 			if pc.disableDefaultConf {
 				klog.Fatalf("Failed to read scheduler config and default configuration fallback is disabled")
 			}
-			klog.Errorf("Failed to read the Scheduler config in '%s', using previous configuration: %v",
+			klog.Errorf("Failed to read the Scheduler config in '%s', falling back to default: %v",
 				pc.schedulerConf, err)
-			return
+			return pc.defaultActions, pc.defaultPlugins, pc.defaultConfigurations, pc.defaultMetricsConf, nil
```
hajnalmt: This return will overwrite the active configuration with the defaults when ReadFile fails. Realistically, a partial file write or a temporary read error will break the scheduler, by switching the configuration to the default (which is just an allocate).

pkg/scheduler/scheduler.go
hajnalmt: @aadhil2k4 This is not the right pattern here. The DefaultSchedulerConf is not a stable fallback... and shouldn't be handled like that.

If the configuration load fails, the previous code effectively kept the old active configuration or on first init loaded the DefaultConfiguration if nothing was provided. 

The new reload function loads these defaults instead of keeping the last right active configuration. This is basically breaking everything, as the Default configuration has an allocate with predicates and nodeorder and that's all, nothing previously working functionality will work.

pkg/scheduler/scheduler.go
```diff
@@ -86,13 +93,53 @@ func NewScheduler(config *rest.Config, opt *options.ServerOption) (*Scheduler, e
...
+
+	pc.applySchedulerConf(actions, plugins, configurations, metricsConf)
+
+	// Cleanup on scheduler shutdown.
+	go func() {
+		<-stopCh
+		pc.mutex.Lock()
+		oldActions := slices.Clone(pc.actions)
+		pc.mutex.Unlock()
+		for _, action := range oldActions {
```
hajnalmt: I can accept this, but I am not happy with this code either. The smart way would be to Unitialize only the actions that are removed. Also initialize the actions that are new, I understand that this is easier, and it was ok in the previous code, but we will probably respawn hundreds of go routines now without an actual reason.

pkg/scheduler/scheduler.go
hajnalmt: You haven't changed anything here, but this is the most important part to handle.

runOnce() copies pc.actions and executes them outside the mutex, while reload/shutdown paths call UnInitialize() on old actions concurrently. You can't just read and Execute now. Neither Initialize nor Unitialize shall be called when Execute runs. Maybe you can find some smart way to run the actions, which are still there but it seems a lot harder than just blocking everything.

hajnalmt (changes requested):
Thank you for the changes @aadhil2k4 .

We wil need more changes on this.

My major problem is that the current code in runOnce() copies pc.actions and executes them outside the mutex, while reload/shutdown paths call UnInitialize() on old actions concurrently.

The impact depends on the action internals, but this means that actions can panic or observe a partially torn down state when a config write or stop signal arrives in the scheduling cycle. 

Please read my other comments too.

pkg/scheduler/scheduler.go
aadhil2k4: @hajnalmt I have used RWMutex instead of Mutex. Now `runOnce` holds a read lock while executing actions, and reload/shutdown use a write lock. This prevents Execute and UnInitialize from running at the same time and ensures correct ordering.

aadhil2k4:
@hajnalmt I have made changes according to the comments. PTAL

PR #5071, opened 2026-03-02
remove duplicate logic

wenchezhao:
#### What type of PR is this?
Remove duplicate logic
Remove duplicate logic

### DIFF
--- a/pkg/scheduler/api/devices/nvidia/vgpu/utils.go
+++ b/pkg/scheduler/api/devices/nvidia/vgpu/utils.go
@@ -83,7 +83,8 @@ func decodeNodeDevices(name, str string) (*GPUDevices, string) {
 				Health:      health,
 				MigTemplate: []config.Geometry{},
 				MigUsage: config.MigInUse{
-					Index: -1},
+					Index: -1,
+				},
 			}
 			sharingMode = getSharingMode(items[5])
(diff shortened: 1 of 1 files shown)


pkg/scheduler/api/devices/nvidia/vgpu/utils.go
```diff
@@ -120,9 +121,6 @@ func encodePodDevices(pd []ContainerDevices) string {
 }
 
 func decodeContainerDevices(str string) ContainerDevices {
-	if len(str) == 0 {
```
hajnalmt, line 123: You should have deleted the second one. It doesn't make sense to create the `tmpdev` containerDevice if string is "".

hajnalmt (changes requested):
Hello thanks for the contribution.

If you have touched all these comments now please apply copilot's suggestion to delete/modify them properly. I think your editor fixed the comment spacing but let's not modify them or if we touch them then do it properly.

Also please sign the DCO!

pkg/scheduler/api/devices/nvidia/vgpu/utils.go
wenchezhao: Deleting both places should have the same effect, but deleting the first place will make the code look more standardized.

PR #5076, opened 2026-03-04
Fix UpdateHyperNode rebuild logic

hzxuzhonghu:
#### What type of PR is this?

/kind cleanup

And also make a optimization: only rebuild when its members change

Fixes `UpdateHyperNode` to properly apply spec changes, restore regex/label rebuild behavior, and rebuild caches when members change so scheduler topology state stays consistent. Updates related tests accordingly.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/api/hyper_node_info.go
+++ b/pkg/scheduler/api/hyper_node_info.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -46,8 +47,7 @@ type HyperNodesInfo struct {
 	// s0->node0, node1, s1->node2, node3, s4->node0, node1, s1->node2, node3
 	realNodesSet map[string]sets.Set[string]
 	// nodeLister Lister to list Kubernetes nodes.
-	nodeLister        listerv1.NodeLister
-	builtErrHyperNode string
+	nodeLister listerv1.NodeLister
 
 	// ready indicates whether the HyperNodesInfo is ready (build process is complete).
 	ready *atomic.Bool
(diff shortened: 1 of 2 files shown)


pkg/scheduler/api/hyper_node_info_test.go
```diff
@@ -239,6 +339,8 @@ func TestHyperNodesInfo_UpdateHyperNode_MultipleParents(t *testing.T) {
 			}
 			assert.Equal(t, false, hni.Ready())
 
+			fmt.Printf("%v\n", hni.HyperNodesInfo())
+
```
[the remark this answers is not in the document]
hzxuzhonghu: Forget to delete it

hzxuzhonghu:
cc @Monokaix

pkg/scheduler/api/hyper_node_info.go
```diff
@@ -248,11 +248,47 @@ func (hni *HyperNodesInfo) DeleteHyperNode(name string) error {
...
-
 	name := hn.Name
 	old, exists := hni.hyperNodes[name]
+
+	specChanged := true
+	membersChanged := true
+	tierChanged := true
+	if exists && old.HyperNode != nil {
+		oldSpec := old.HyperNode.Spec
+		specChanged = !reflect.DeepEqual(oldSpec, hn.Spec)
```
hajnalmt: I feel that reflect here is comparatively slow and a little overkill for a struct with 3 fields. Plus these are dependent `specChanged` will be true if `membersChanged` or `tierChanged` is true.  I would prefer to use the k8s equality library for this instead of reflect. 

```
oldSpec := old.HyperNode.Spec
newSpec :=  hn.Spec
tierChanged := oldSpec.Tier != newSpec.Tier
membersChanged := !equality.Semantic.DeepEqual(oldSpec.Members, newSpec.Members)
specChanged := tierChanged || membersChanged || oldSpec.TierName != newSpec.TierName
```
This is somewhat more exact.

pkg/scheduler/api/hyper_node_info_test.go
```diff
@@ -95,6 +97,106 @@ func TestHyperNodesInfo_UpdateHyperNode_Normal(t *testing.T) {
...
+	ready := new(atomic.Bool)
+	ready.Store(false)
+	hni := NewHyperNodesInfoWithCache(hyperNodes, hyperNodesSetByTier, realNodesSet, ready)
+
+	updated := s0.DeepCopy()
+	updated.Spec.Tier = 2
+
+	err := hni.UpdateHyperNode(updated)
+	assert.NoError(t, err)
+	assert.False(t, hni.Ready())
```
hajnalmt, line 170: This is the correct behavioural regession in your code. In the old code, updateAncestors -> rebuildCache -> setReady(true) was called unconditionally on every UpdateHyperNode, so any successful update (including tier-only changes) would restore ready to true even if a prior rebuild had failed. 

Now if rebuildCache fails then a tier only change won't call this path anymore and it stays false.

This seems correct to me. If ready is false because the cache is stale from a failed rebuild, a tier-only update doesn't rebuild the cache, so flipping ready to true would be dishonest. Recovery will happen on the next informer resync that triggers a full rebuild. But it's a change from the old behavior and worth being aware of.

Maybe adding a comment at the `tierChanged` path about this regression or updating the PR description with this information can be helpful.
```
if !exists || tierChanged {
		hni.updateHyperNodesSetByTier(hn)
}
```

pkg/scheduler/api/hyper_node_info_test.go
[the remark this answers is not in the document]
hajnalmt: Please delete it, this PR looks overall good to me.

hajnalmt (commented):
This PR looks overall good to me 😊 This is a really great improvement!
I had 2 minor comments, but they are really minor.

hzxuzhonghu:
Thank you for your review @hajnalmt I will try to address all the comments

pkg/scheduler/api/hyper_node_info.go
hzxuzhonghu: make sense

hajnalmt (changes requested):
Thanks for the changes.
The PR fails in the CI on a lint issue still:
```
Error: pkg/scheduler/api/hyper_node_info_test.go:20:2: "fmt" imported and not used (typecheck)
/*
1 issues:
* typecheck: 1
FAILED: golangci-lint stale.
```

pkg/scheduler/api/hyper_node_info.go
hajnalmt: You haven't addressed this in the end.

hajnalmt (commented):
You havent addressed my review comment, but besides that this is a nice addition.

pkg/scheduler/api/hyper_node_info.go
hzxuzhonghu: Thinking a little bit more.  I feel it is safer to use `DeepEqual`, we may miss update the logic here when later we add a new fields in HyperNode. 

But good to migrate to equality.Semantic.DeepEqual

hzxuzhonghu: Since tierChanged membersChanged are there, i accepted the whole suggestion now

hzxuzhonghu: I must forget to commit

hzxuzhonghu:
Done, i should have used my own repo.

hajnalmt (approved):
/area scheduling
/approve

Beautiful! Nice work. Lets see if Jesse wants to take a look too.

hajnalmt (approved):
/lgtm

We will need this feature as a base for future implementations, so let's merge it.

PR #5089, opened 2026-03-06
fix(job): prevent TaskFailedEvent overwrite by phase transition in up…

Aman-Cool:
**What type of PR is this?**
bug fix / cleanup

In `updatePod`, the `PodRunning` and `PodPending` cases each have two independent `if` blocks, one that sets `TaskFailedEvent` and one that sets a phase-transition event (`PodRunningEvent` / `PodPendingEvent`). Since they're independent, the phase-transition event silently overwrites `TaskFailedEvent` whenever both conditions are true at the same time.

The fix makes the phase-transition check an `else if`, so it only fires when the task hasn't actually failed, which is the only case where a phase-transition event is meaningful anyway.

Originally I claimed this caused `TaskFailed` -> `RestartJob`/`AbortJob` policies to break under `CrashLoopBackOff`, but after testing that was wrong, `CrashLoopBackOff` never leaves the `Running` phase at the pod level, so the overwrite never actually occurs there. The fix still stands though: `Running -> Running` updates shouldn't re-dispatch `PodRunningEvent` regardless, as it carries no new state and causes unnecessary reconciliation. The `else if` makes the intent explicit and closes the overwrite possibility for any future scenario where both conditions could coincide.

Fixes #NA

Worth confirming `TaskFailed()` is a pure cache read with no side effects, since `else if` now gates the phase-transition check on it. From context it looks clean, but good to verify.

Prevent `PodRunningEvent` from overwriting `TaskFailedEvent` in `updatePod` when a task has exceeded `maxRetry`.

### DIFF
--- a/test/e2e/cronjob/cronjob_basic.go
+++ b/test/e2e/cronjob/cronjob_basic.go
@@ -451,6 +451,24 @@ var _ = Describe("CronJob E2E Test", func() {
 			return true
 		}, 3*time.Minute, 10*time.Second).Should(BeTrue())
 
+		By("Suspending CronJob to prevent new jobs from being created during stability check")
+		latestCronJob, err := getCronJob(ctx, ctx.Namespace, cronJobName)
+		Expect(err).NotTo(HaveOccurred())
+		suspend := true
+		latestCronJob.Spec.Suspend = &suspend
+		createdCronJob, err = ctx.Vcclient.BatchV1alpha1().CronJobs(ctx.Namespace).Update(
+			context.TODO(), latestCronJob, metav1.UpdateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Waiting for all active jobs to complete after suspension")
+		Eventually(func() int {
+			count, err := getJobActiveNum(ctx, createdCronJob)
+			if err != nil {
+				return -1
+			}
+			return count
+		}, 3*time.Minute, 10*time.Second).Should(Equal(0))
+
 		By("Ensuring only one fail finished job exists")
 		Consistently(func() int {
 			count, err := getJobFailNum(ctx, createdCronJob)
@@ -458,7 +476,7 @@ var _ = Describe("CronJob E2E Test", func() {
 				Fail(fmt.Sprintf("Failed to get jobs: %v", err))
 			}
 			return count
-		}, 3*time.Minute, 10*time.Second).Should(Equal(1))
+		}, 30*time.Second, 10*time.Second).Should(Equal(1))
 
 		By("Cleaning up test resources")
 		err = deleteCronJob(ctx, ctx.Namespace, cronJobName)


Aman-Cool:
@hajnalmt  @hzxuzhonghu @JesseStutler, In `updatePod`, both the `TaskFailed` check and the phase-transition check are plain `if` statements, so when a CrashLoopBackOff pod transitions `Pending -> Running` after exceeding `maxRetry`, `TaskFailedEvent` gets set and immediately overwritten by `PodRunningEvent`. The job's failure policies never fire and it just loops forever.

Changed both to `else if`. Tiny diff but jobs were silently ignoring their own retry limits.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -287,15 +287,13 @@ func (cc *jobcontroller) updatePod(oldObj, newObj interface{}) {
...
 			event = bus.PodRunningEvent
 		}
 	case v1.PodPending:
 		if cc.cache.TaskFailed(jobcache.JobKeyByName(newPod.Namespace, jobName), taskName) {
 			event = bus.TaskFailedEvent
-		}
-		if oldPod.Status.Phase != v1.PodPending {
+		} else if oldPod.Status.Phase != v1.PodPending {
 			event = bus.PodPendingEvent
 		}
```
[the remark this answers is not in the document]
Aman-Cool: Added `TestUpdatePodFuncTaskFailedEvent` covering both cases, `Pending -> Running` and `Failed -> Pending` with `restartCount >= maxRetry`, both assert `TaskFailedEvent`. sets up `Replicas=1, MaxRetry=3`, drains the queue, calls `updatePod`, checks `req.Event`. both pass.

pkg/controllers/job/job_controller_handler_test.go
```diff
@@ -569,6 +570,105 @@ func TestDeletePodFunc(t *testing.T) {
 	}
 }
 
+// TestUpdatePodFuncTaskFailedEvent verifies that when a task has exceeded its
+// maxRetry threshold, updatePod enqueues TaskFailedEvent even if the pod is
+// simultaneously undergoing a phase transition (Pending→Running or Failed→Pending).
```
hajnalmt: There is no such transition as Failed -> Pending. Pod phases are terminal, once a pod reaches Failed or Succeeded, it stays there permanently. The Kubernetes API server won't accept a status update that moves a Pod backward in its lifecycle. It would be better to update the description here, but not blocking.

pkg/controllers/job/job_controller_handler_test.go
```diff
@@ -569,6 +570,105 @@ func TestDeletePodFunc(t *testing.T) {
...
+// Without the else-if fix both events would fire and PodRunningEvent/PodPendingEvent
```
hajnalmt: I don't like these comments that are reflected in the unit tests but there is no reference to the issue or PR. If I am just reading this unit test I won't know what the `else-if fix is`. I won't block on this but at least mention your PR, like https://github.com/volcano-sh/volcano/pull/5089 next time.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -287,15 +287,13 @@ func (cc *jobcontroller) updatePod(oldObj, newObj interface{}) {
 	case v1.PodRunning:
 		if cc.cache.TaskFailed(jobcache.JobKeyByName(newPod.Namespace, jobName), taskName) {
```
hajnalmt, line 291: I would consider a comment here above the if block. (same for Pending)

```
// TaskFailedEvent takes precedence over phase-transition events:
// when maxRetry is exceeded, the failure must propagate even if
// the pod is restarting.
```

pkg/controllers/job/job_controller_handler.go
```diff
@@ -287,15 +287,13 @@ func (cc *jobcontroller) updatePod(oldObj, newObj interface{}) {
...
-		if oldPod.Status.Phase != v1.PodRunning {
+		} else if oldPod.Status.Phase != v1.PodRunning {
 			event = bus.PodRunningEvent
 		}
 	case v1.PodPending:
 		if cc.cache.TaskFailed(jobcache.JobKeyByName(newPod.Namespace, jobName), taskName) {
 			event = bus.TaskFailedEvent
-		}
-		if oldPod.Status.Phase != v1.PodPending {
+		} else if oldPod.Status.Phase != v1.PodPending {
```
hajnalmt, line 302: This is just a defensive measure as it can never happen.

When a Pod truly Failed, Volcano's job controller creates a new Pod with a new name/UID. That new Pod starts at Pending. The updatePod handler would never see this as an update, it would hit addPod instead. See `syncJob` in `job_controller_actions`:
https://github.com/volcano-sh/volcano/blob/49b8f40ba6247b9a6a65e14e6e14c4fb19feb77c/pkg/controllers/job/job_controller_actions.go#L470-L475

and 

https://github.com/volcano-sh/volcano/blob/49b8f40ba6247b9a6a65e14e6e14c4fb19feb77c/pkg/controllers/job/job_controller_actions.go#L521

The fix here is harmless but dead code, I am ok with it, as I am not against defensive programming, but worth rewriting your PR description to reflect this.

pkg/controllers/job/job_controller_handler_test.go
```diff
@@ -569,6 +570,105 @@ func TestDeletePodFunc(t *testing.T) {
...
+			controller := newController()
+			controller.addJob(jobWithTask())
+			addPodAnnotation(tc.oldPod, annotation)
+			addPodAnnotation(tc.newPod, annotation)
+			controller.addPod(tc.oldPod)
+
+			// drain items queued by addJob and addPod so only updatePod's item remains
+			key := fmt.Sprintf("%s/%s", namespace, "job1")
+			queue := controller.getWorkerQueue(key)
+			for queue.Len() > 0 {
```
hajnalmt, line 672: This is clever to isolate the updatePod event like this. 😄

pkg/controllers/job/job_controller_handler.go
```diff
@@ -287,15 +287,13 @@ func (cc *jobcontroller) updatePod(oldObj, newObj interface{}) {
 	case v1.PodRunning:
 		if cc.cache.TaskFailed(jobcache.JobKeyByName(newPod.Namespace, jobName), taskName) {
 			event = bus.TaskFailedEvent
-		}
-		if oldPod.Status.Phase != v1.PodRunning {
+		} else if oldPod.Status.Phase != v1.PodRunning {
```
kingeasternsun, line 293: Thanks for sharing this. I have a quick question: could this approach potentially result in the PodRunningEvent being missed?

Aman-Cool: Yes, technically it can, if TaskFailed() returns true, PodRunningEvent won't fire for that transition. But that's the intended behavior: if the task has already blown past maxRetry, we want the failure to take over. Emitting PodRunningEvent at that point would just mask the failure and let the job continue as if nothing went wrong, which is exactly the bug we're fixing.

So the "missed" PodRunningEvent is by design. The pod may be running momentarily, but it's in a crash loop and the job needs to act on the failure, not the transient running state.

pkg/controllers/job/job_controller_handler_test.go
Aman-Cool: 😁

pkg/controllers/job/job_controller_handler.go
kingeasternsun: Ok。

hajnalmt:

@JesseStutler WDYT?

JesseStutler:
/hold
Although I think the code change is ok, I would like to ask why you said that `The result: jobs with TaskFailed -> RestartJob/AbortJob policies never act on the failure. `. Could you provide your volcano version, your yamls and verify it? I'd like to see your test phenomenon

I accepted this code only because I felt that `Running -> Running` shouldn't trigger the Running Event repeatedly, but not for the reason you mentioned. `CrashLoopBackoff` isn't Pending -> Running, it's always in the Running phase.

Aman-Cool:
Hi @JesseStutler, so you were right.

I tested it after your comment, `CrashLoopBackOff` never leaves the `Running` phase. The pod just sits there looking completely unbothered at the phase level while the container is having a full breakdown inside it. `oldPod.Status.Phase != v1.PodRunning` is always false, `PodRunningEvent` never fires, `TaskFailed` dispatches correctly. The scenario I described doesn't reproduce.

What happened is I saw the potential overwrite, went "aha, THIS must be why `TaskFailed` policies sometimes feel flaky" and then connected dots that weren't actually connected😅. The real reason the fix makes sense is exactly what you said, `Running -> Running` shouldn't retrigger `PodRunningEvent`. That's it. I just dressed it up in a ghost story it didn't need.

Updated the description. Thanks for actually pushing back on it, genuinely appreciate that.

pkg/controllers/job/job_controller_handler_test.go
```diff
@@ -569,6 +570,108 @@ func TestDeletePodFunc(t *testing.T) {
...
+	}
+
+	testcases := []struct {
+		name          string
+		oldPod        *v1.Pod
+		newPod        *v1.Pod
+		expectedEvent bus.Event
+	}{
+		{
+			name:   "Pending to Running with restartCount >= maxRetry emits TaskFailedEvent",
```
JesseStutler, line 615: Better also add Running/Pending event to make sure that changing phase from other phase will not miss emit Running/Pending event

pkg/controllers/job/job_controller_handler_test.go
```diff
@@ -569,6 +570,108 @@ func TestDeletePodFunc(t *testing.T) {
 	}
 }
 
+// TestUpdatePodFuncTaskFailedEvent verifies that when a task has exceeded its
+// maxRetry threshold, updatePod enqueues TaskFailedEvent even if the pod is
+// simultaneously undergoing a phase transition (Pending→Running).
+// The Failed→Pending case is a defensive test: pod phases are terminal in
+// Kubernetes, so that transition cannot happen via updatePod in practice.
+// Without the fix in https://github.com/volcano-sh/volcano/pull/5089,
+// PodRunningEvent/PodPendingEvent would silently overwrite TaskFailedEvent,
```
JesseStutler: And then you should modify the description of this func,  even I think it's not an important bug, we could not to add the pr link, just describe the what does this func do is fine

JesseStutler:
> Hi @JesseStutler, so you were right.
> 
> I tested it after your comment, `CrashLoopBackOff` never leaves the `Running` phase. The pod just sits there looking completely unbothered at the phase level while the container is having a full breakdown inside it. `oldPod.Status.Phase != v1.PodRunning` is always false, `PodRunningEvent` never fires, `TaskFailed` dispatches correctly. The scenario I described doesn't reproduce.
> 
> What happened is I saw the potential overwrite, went "aha, THIS must be why `TaskFailed` policies sometimes feel flaky" and then connected dots that weren't actually connected😅. The real reason the fix makes sense is exactly what you said, `Running -> Running` shouldn't retrigger `PodRunningEvent`. That's it. I just dressed it up in a ghost story it didn't need.
> 
> Updated the description. Thanks for actually pushing back on it, genuinely appreciate that.

So this is what I wanted to say before: never tell a story. I just want to see the bug's behavior, and after you fix the code, you can also post your verification screenshots. Otherwise, with so many suddenly popping up edge cases, I would reasonably suspect that it was scanned using an AI agent without proper verification :)

Aman-Cool:
> So this is what I wanted to say before: never tell a story.

@JesseStutler, Lol fair enough, screenshots next time, not essays...
And yeah, I get it. A wall of text explaining edge cases I can't actually reproduce does look... suspicious😅. 
I'll be honest, I do use AI tools to help me understand unfamiliar code faster, but the bug hunting and verification is mine. Clearly I need to show my work better instead of just narrating it.
Listened to your feedback on #5091(do have a look on #5100😁).

Aman-Cool:
@hajnalmt, did some changes in this PR, could you please take a look🙂.

PR #5091, opened 2026-03-08
fix(capacity): prevent nil panic in hierarchical queue mode

Aman-Cool:
#### What type of PR is this?
fix

If you delete a queue while jobs are still running, or if the scheduler starts before any Queue CRs exist, `buildHierarchicalQueueAttrs` will silently get a nil pointer back from the map and immediately crash the scheduler, taking down scheduling for every workload in the cluster.

This adds two nil checks: one before touching the root queue attr, one before iterating a job's queue attr. Instead of panicking, the function now returns `false` (which already means "skip this session") and logs a warning. The scheduler recovers on its own next cycle.

Fixes #NA

None

```release-note
Fix scheduler panic in capacity plugin when a queue is deleted while jobs are still running, or when the scheduler starts before Queue CRs are created in hierarchical mode.
```

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -612,6 +612,11 @@ func (cp *capacityPlugin) buildHierarchicalQueueAttrs(ssn *framework.Session) bo
 	for _, job := range ssn.Jobs {
 		klog.V(4).Infof("Considering Job <%s/%s>.", job.Namespace, job.Name)
 		attr := cp.queueOpts[job.Queue]
+		if attr == nil {
+			klog.Warningf("Job <%s/%s> references queue %s which is not found in session queues, skipping",
+				job.Namespace, job.Name, job.Queue)
+			continue
+		}
 		if len(attr.children) > 0 {
 			klog.Errorf("The Queue <%s> of Job <%s/%s> is not leaf queue", attr.name, job.Namespace, job.Name)
 			return false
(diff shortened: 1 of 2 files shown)


Aman-Cool:
Hey @hajnalmt @JesseStutler @hzxuzhonghu, found this while reading through the capacity plugin, if you're running hierarchical queues and the root queue is missing from the session snapshot (e.g. scheduler starts before Queue CRs are ready, or a queue gets deleted while jobs are still alive), `buildHierarchicalQueueAttrs` dereferences a nil pointer and the scheduler panics. Every workload stops until the pod restarts.
                                                                                                                                                                                                                                                 
 Fix is just two nil checks, return false early if root is missing (already the "skip this session" path), and skip jobs whose queue has been deleted instead of crashing. Added regression tests for both cases.

JesseStutler:
@Aman-Cool There is a failed test in your PR, please check

JesseStutler:
@Aman-Cool BTW, I would you like to raise your issues, and post your findings (like the panic logs, the reproduce steps) in the future, not simply add your description in the PR, otherwise it may confuse me with you are using some AI agent tools to scan the repo and then raise these PRs

Aman-Cool:
@JesseStutler, totally fair point, I'll open an issue first next time with the panic stack trace and reproduce steps before sending a PR or should I be adding all this in the PR description itself to keep everything in one place.Whichever you prefer😊. 
I found this while reading through the capacity plugin code after running into some scheduler instability in my own cluster setup, not through any automated scanning🙂.

> @Aman-Cool There is a failed test in your PR, please check

Will be looking on this. Thank you for the review🙏

JesseStutler:
> @JesseStutler, totally fair point, I'll open an issue first next time with the panic stack trace and reproduce steps before sending a PR or should I be adding all this in the PR description itself to keep everything in one place.Whichever you prefer😊. I found this while reading through the capacity plugin code after running into some scheduler instability in my own cluster setup, not through any automated scanning🙂.
> 
> > @Aman-Cool There is a failed test in your PR, please check
> 
> Will be looking on this. Thank you for the review🙏

I would prefer to raise the issue first and see the panic logs or the error phenomenon

JesseStutler:
@hajnalmt Hi mate even after #5081 get merged, there are still some old pending workflows, why it will happen and what do we need to do?

Update: It looks like because branch protect rules has the old static workflows configured, now I have already updated :)

JesseStutler:
Make sense
/approve
/lgtm

PR #5098, opened 2026-03-12
feat(sharding): live-reloadable scheduler sharding config via ConfigMap

### LINKED ISSUE #4877: ShardingController - Scheduler Sharding Configurations

agrawalcodes:
Scheduler sharding parameters (`cpuUtilizationMin/Max`, `preferWarmupNodes`, `minNodes/maxNodes`, `shardSyncPeriod`) were baked into CLI flags. Any change required a controller restart, increasing both deployment friction and mean-time-to-recovery during production incidents.

The ShardingController now watches a dedicated ConfigMap (`volcano-sharding-configmap` in `volcano-system` by default) and hot-reloads the scheduler shard specifications whenever the ConfigMap is created or updated, without restarting the controller.

Closes #4877

### Configuration format (`sharding.yaml` key)

```yaml
schedulerConfigs:
  - name: agent-scheduler
    type: agent
    cpuUtilizationMin: 0.7
    cpuUtilizationMax: 1.0
    preferWarmupNodes: true
    minNodes: 1
    maxNodes: 100
  - name: volcano
    type: volcano
    cpuUtilizationMin: 0.0
    cpuUtilizationMax: 0.69
    preferWarmupNodes: false
    minNodes: 1
    maxNodes: 100
shardSyncPeriod: "60s"
enableNodeEventTrigger: true
```

### Behaviour

| Scenario | Behaviour |
|----------|-----------|
| ConfigMap exists at startup | ConfigMap config takes precedence over `--scheduler-configs` flags |
| ConfigMap absent at startup | Falls back to flag-based defaults; watches for later creation |
| ConfigMap updated at runtime | Controller re-parses, validates, atomically swaps config, and triggers a shard re-sync. No restart needed |
| ConfigMap deleted | Warning logged; last known config is retained |
| Invalid YAML or constraint violation | Error logged; previous config is kept |

### Changes

- **`pkg/controllers/sharding/config.go`** -- Added `ShardingConfig` / `SchedulerConfigSpec` YAML structs, `ParseShardingConfig()` with validation, ConfigMap coordinate constants (`DefaultConfigMapName`, `DefaultConfigMapNamespace`), and flags (`--sharding-configmap`, `--sharding-configmap-namespace`). Marked `--scheduler-configs` as deprecated.
- **`pkg/controllers/sharding/sharding_controller.go`** --
  - Added a dedicated namespace+name-scoped ConfigMap informer factory (only caches the single sharding ConfigMap, not all ConfigMaps cluster-wide).
  - ConfigMap event handlers (`onConfigMapAdd/Update/Delete`) call `reloadFromConfigMap()` which parses, validates, atomically swaps config under `configMu`, invalidates assignment cache, and enqueues shards for non-blocking resync via the work queue.
  - `periodicShardSync()` uses a resettable timer with `shardSyncPeriodUpdateCh` so live `shardSyncPeriod` changes take effect immediately.
  - Added `cleanupStaleNodeShards()` to delete NodeShards for schedulers removed from the config.
  - Fixed pre-existing nil map bug in `calculateAndApplyAssignment` where `AssignmentContext.AssignedNodes` was not initialized.
  - Removed unused `InitializeWithConfigs` method (tests now use the production ConfigMap flow).
- **`pkg/controllers/sharding/node_events.go`** -- `enqueueNodeEvent` reads `EnableNodeEventTrigger` under `configMu.RLock()` to avoid data races with live config updates.
- **`pkg/controllers/sharding/test_utils.go`** -- Added `buildShardingConfigMap()` helper and `SkipConfigMap` option. Tests now mirror the production flow: ConfigMap is injected into the fake client and loaded by `Run()`.
- **`pkg/controllers/sharding/config_test.go`** -- Unit tests for YAML parsing and validation (valid config, optional fields, empty list, missing name, CPU range, min/max constraints, invalid YAML, default options).
- **`pkg/controllers/sharding/configmap_reload_test.go`** -- Integration tests for ConfigMap reload flow:
  - `UpdatesSchedulerConfigs` -- verifies reload changes scheduler count and parameters
  - `UpdatesShardSyncPeriod` -- verifies sync period is updated from ConfigMap
  - `UpdatesEnableNodeEventTrigger` -- verifies event trigger toggle works
  - `InvalidYAMLKeepsPreviousConfig` -- bad YAML doesn't clobber working config
  - `MissingKeySkipsReload` -- wrong ConfigMap key is silently ignored
  - `InvalidatesAssignmentCache` -- cache is cleared after reload
  - `RecreatesShardingManager` -- ShardingManager is rebuilt with new configs
  - `ValidationRejectsEmptySchedulerList` -- validation failures preserve previous config
  - `CleansUpStaleNodeShardsAfterSchedulerRemoved` -- removed scheduler's NodeShard is deleted
  - `NewSchedulerAddedAndOldShardUpdated` -- new scheduler gets a NodeShard, existing ones preserved
  - `FallsBackToFlagDefaultsWhenConfigMapAbsent` -- controller works with flag defaults when no ConfigMap exists
- **`docs/user-guide/how_to_configure_sharding_configmap.md`** -- User guide covering ConfigMap format, parameter reference, behaviour matrix, Helm values, and validation rules.
- **`example/sharding/sharding-config-configmap.yaml`** -- Example ConfigMap manifest.
- **`installer/helm/chart/volcano/templates/controllers.yaml`** -- Helm template creates the sharding ConfigMap and passes `--sharding-configmap` / `--sharding-configmap-namespace` flags to the controller.
- **`installer/helm/chart/volcano/values.yaml`** -- Added `sharding_configmap_enable` (default: `true`) and `sharding_configmap_data` values with documentation.
- **`installer/volcano-development.yaml`** -- Regenerated to include the sharding ConfigMap resource and controller flags.
- **`hack/generate-yaml.sh`** -- Added `--set custom.sharding_configmap_enable=true` to the Helm template command.

### Backward compatibility

- Existing deployments without the ConfigMap continue to work unchanged.
- All existing CLI flags are preserved; the ConfigMap is additive and opt-in.
- All pre-existing unit tests continue to pass.

### Test plan

- [x] All existing sharding controller unit tests pass
- [x] YAML parsing validation tests pass (10 cases)
- [x] ConfigMap reload integration tests pass (11 cases)
- [ ] Manual verification: deploy controller, create/update/delete ConfigMap, observe shard re-sync in logs
- [ ] E2E tests (tracked in PR #4990, will read ConfigMap config and assert NodeShard specs match)

### DIFF
--- a/pkg/controllers/sharding/sharding_controller.go
+++ b/pkg/controllers/sharding/sharding_controller.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/spf13/pflag"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -98,6 +99,9 @@ type ShardingController struct {
 	// shardSyncPeriodUpdateCh tells the periodic sync loop to reset its timer
 	// after a live config update.
 	shardSyncPeriodUpdateCh chan struct{}
+	// flagsRegistered tracks whether AddFlags was called (production path).
+	// When false, Initialize creates default options instead.
+	flagsRegistered bool
 }
 
 // Return the name of the controller
@@ -105,11 +109,28 @@ func (sc *ShardingController) Name() string {
 	return controllerName
 }
 
+// AddFlags implements framework.FlagProvider, registering controller-specific
+// flags before the binary's flag set is parsed.
+func (sc *ShardingController) AddFlags(fs *pflag.FlagSet) {
+	sc.controllerOptions = NewShardingControllerOptions()
+	sc.controllerOptions.AddFlags(fs)
+	sc.flagsRegistered = true
+}
+
 // Initialize initializes the controller
 func (sc *ShardingController) Initialize(opt *framework.ControllerOption) error {
 	klog.V(2).Infof("Initializing ShardingController...")
 	sc.ctx = context.Background()
-	sc.controllerOptions = NewShardingControllerOptions()
+	// When AddFlags was called (production), re-parse the raw config strings
+	// that may have been overridden by command-line flags.
+	// When it was not called (tests), initialize options with defaults.
+	if sc.flagsRegistered {
+		if err := sc.controllerOptions.ParseConfig(); err != nil {
+			klog.Warningf("Failed to parse scheduler configs from flags: %v", err)
+		}
+	} else {
+		sc.controllerOptions = NewShardingControllerOptions()
+	}
 
 	sc.kubeClient = opt.KubeClient
 	sc.vcClient = opt.VolcanoClient
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -66,6 +66,14 @@ func main() {
 		return controllerNames
 	}
 	s.AddFlags(fs, knownControllers())
+
+	// Register controller-specific flags for controllers that implement FlagProvider.
+	framework.ForeachController(func(controller framework.Controller) {
+		if fp, ok := controller.(framework.FlagProvider); ok {
+			fp.AddFlags(fs)
+		}
+	})
+
 	utilfeature.DefaultMutableFeatureGate.AddFlag(fs)
 
 	commonutil.LeaderElectionDefault(&s.LeaderElection)
--- a/pkg/controllers/framework/interface.go
+++ b/pkg/controllers/framework/interface.go
@@ -17,6 +17,7 @@ limitations under the License.
 package framework
 
 import (
+	"github.com/spf13/pflag"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -52,4 +53,12 @@ type Controller interface {
 	Initialize(opt *ControllerOption) error
 	// Run run the controller
 	Run(stopCh <-chan struct{})
+}
+
+// FlagProvider is an optional interface that controllers can implement to
+// register controller-specific command-line flags. Controllers that implement
+// this interface will have their AddFlags method called during flag setup
+// in the controller-manager binary.
+type FlagProvider interface {
+	AddFlags(fs *pflag.FlagSet)
 }


hajnalmt:
/cc @hajnalmt
I will look into this.

agrawalcodes:
PTAL @hajnalmt @JesseStutler

example/sharding/sharding-config-configmap.yaml
```diff
@@ -0,0 +1,64 @@
...
+# Deploy with:
+#   kubectl apply -f sharding-config-configmap.yaml
+#
+# The controller is told which ConfigMap to watch via:
+#   --sharding-config-map=<name>  (default: volcano-sharding-config)
+#   --sharding-config-map-namespace=<namespace>  (default: volcano-system)
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: volcano-sharding-config
```
hajnalmt: Without helm wiring I think this feature from a product/operability perspective is **incomplete** and cannot be used practically. Also this should  be added at least to the `installer/volcano-agent-scheduler-development.yaml`, to deploy it with that by default. I would also consider RBAC too (it can happen that the schedulers are in different namespace).

You baked the defaults in out of convenience, but this should be dynamic actually, and not just reflect that it has dynamism. 

Also the name shall contain the helm release name like {helm-release-name}-sharding-config when deployed with helm.

example/sharding/sharding-config-configmap.yaml
```diff
@@ -0,0 +1,64 @@
...
+  namespace: volcano-system
+data:
+  # sharding.yaml is the well-known key inside the ConfigMap.
+  # All fields follow standard YAML / JSON naming conventions.
+  sharding.yaml: |
```
hajnalmt, line 22: This whole yaml should be setable through helm and the comments should be in the `values.yaml`.

example/sharding/sharding-config-configmap.yaml
```diff
@@ -0,0 +1,64 @@
...
+#
+# The ShardingController watches this ConfigMap at runtime and reloads
+# the scheduler shard specifications whenever the data changes, without
+# requiring a controller restart.
+#
+# Deploy with:
+#   kubectl apply -f sharding-config-configmap.yaml
+#
+# The controller is told which ConfigMap to watch via:
+#   --sharding-config-map=<name>  (default: volcano-sharding-config)
```
hajnalmt: sharding-configmap

example/sharding/sharding-config-configmap.yaml
```diff
@@ -0,0 +1,64 @@
...
+#   --sharding-config-map-namespace=<namespace>  (default: volcano-system)
```
hajnalmt: sharding-configmap-namespace

pkg/controllers/sharding/config.go
```diff
@@ -21,60 +21,175 @@ import (
 
 	"github.com/spf13/pflag"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
 )
 
+const (
+	// DefaultConfigMapName is the default name of the ConfigMap holding sharding configurations.
+	DefaultConfigMapName = "volcano-sharding-config"
```
hajnalmt: These baked in defaults shall be removed, as this won't match real deployments anyway configuration for these shall come from the right places. (Imagine custom namespaces, multi-tenant installs)

pkg/controllers/sharding/config.go
```diff
@@ -21,60 +21,175 @@ import (
...
+	// MaxNodes is the maximum number of nodes the shard may contain.
+	MaxNodes int `json:"maxNodes"`
 }
 
+// ShardingConfig is the top-level structure that is serialised as YAML into
+// the sharding ConfigMap (key: sharding.yaml).
+//
+// Example ConfigMap data:
+//
+//	schedulerConfigs:
```
hajnalmt: We can remove this comment. It will be more than enough to read the actual configmap.

pkg/controllers/sharding/node_events.go
```diff
@@ -127,6 +127,14 @@ func (sc *ShardingController) areMapsEqual(a, b map[string]string) bool {
 
 // enqueueNodeEvent adds a node event to the queue
 func (sc *ShardingController) enqueueNodeEvent(nodeName, eventType, source string) {
+	sc.configMu.RLock()
+	enableNodeEventTrigger := sc.controllerOptions.EnableNodeEventTrigger
```
hajnalmt, line 131: Whao this was missing completely before. Good catch.

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -103,11 +119,28 @@ func (sc *ShardingController) Initialize(opt *framework.ControllerOption) error
...
 	sc.nodeInformer = sc.kubeInformerFactory.Core().V1().Nodes()
 	sc.podInformer = sc.kubeInformerFactory.Core().V1().Pods()
 	sc.shardInformer = sc.vcInformerFactory.Shard().V1alpha1().NodeShards()
 
+	// Create a namespace+name-scoped informer factory so only the single
+	// sharding ConfigMap is cached, not all ConfigMaps cluster-wide.
+	sc.configMapInformerFactory = informers.NewSharedInformerFactoryWithOptions(
+		sc.kubeClient,
+		0,
+		informers.WithNamespace(sc.configMapNamespace),
```
hajnalmt: This is pretty needed, good call. I checked it and the RBAC for the controller seems to have ClusterRole configmap so this is fine.

pkg/controllers/sharding/config.go
```diff
@@ -21,60 +21,175 @@ import (
...
 func (opts *ShardingControllerOptions) AddFlags(fs *pflag.FlagSet) {
-	// Default scheduler configs
 	defaultConfigs := []string{
 		"volcano:volcano:0.0:0.6:false:2:100",
 		"agent-scheduler:agent:0.7:1.0:true:2:100",
 	}
 
 	fs.StringSliceVar(&opts.SchedulerConfigsRaw, "scheduler-configs", defaultConfigs,
-		"Scheduler configurations in format: name:type:min_util:max_util:prefer_warmup:min_nodes:max_nodes")
+		"Scheduler configurations in format: name:type:min_util:max_util:prefer_warmup:min_nodes:max_nodes. "+
```
hajnalmt: Can you mark this as deprecated use a configmap?

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -677,6 +750,203 @@ func (sc *ShardingController) InitializeWithConfigs(opt *framework.ControllerOpt
 	return nil
 }
 
+// ─── ConfigMap hot-reload helpers ────────────────────────────────────────────
```
hajnalmt: remove line

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -677,6 +750,203 @@ func (sc *ShardingController) InitializeWithConfigs(opt *framework.ControllerOpt
...
+	sc.reloadFromConfigMap(cm)
+	return nil
+}
+
+// applyShardingConfig updates the controller's schedulerConfigs and
+// shardingManager from a parsed ShardingConfig.  It acquires configMu to
+// prevent races with concurrent reads.
+func (sc *ShardingController) applyShardingConfig(cfg *ShardingConfig) {
+	newConfigs := make([]SchedulerConfig, 0, len(cfg.SchedulerConfigs))
+	for _, spec := range cfg.SchedulerConfigs {
```
hajnalmt, line 868: What will happen if we remove a scheduler from the config? Will it leave NodeShards behind? Or the sync will solve that?

pkg/controllers/sharding/sharding_controller.go
hajnalmt, line 437: We are iterating thorugh `newAssignments` here that only contains currently configured schedulers, because `CalculateShardAssignments()` iterates through `sm.schedulerConfigs` which means that old shards from  a removed scheduler’s NodeShard is never revisited for delete/reconcile and can stay stale indefinitely. Do I understand this right?

example/sharding/sharding-config-configmap.yaml
agrawalcodes: Addressed all points:
1. helm wiring added in `templates/controllers.yaml` behind `custom.sharding_configmap_enable: false` (opt-in). The controller flags and configmap resource are both gated on that value.
2. added the configmap to `installer/volcano-agent-scheduler-development.yaml`.
3. RBAC is covered by the existing ClusterRole (as you noted in a comment).
4. configmap name is `{{ .Release.Name }}-sharding-configmap`, so it's scoped to the release. The hardcoded `DefaultConfigMapName/DefaultConfigMapNamespace` constants and their defaults have been removed, `ConfigMapName` and `ConfigMapNamespace` now default to empty strings and must be provided explicitly (via helm or flags).

pkg/controllers/sharding/sharding_controller.go
agrawalcodes: You're right on both counts, it was a real bug. Added `cleanupStaleNodeShards()` at the end of `syncShards()`. It iterates all existing NodeShard objects from the lister and deletes any whose scheduler name is absent from newAssignments. Since syncShards() runs on every periodic tick and on every configmap reload, orphaned NodeShards are cleaned up within one sync cycle after a scheduler is removed from the config. Deletion is idempotent, `NotFound` errors are swallowed. Also added a unit test to cover this path.

pkg/controllers/sharding/sharding_controller.go
agrawalcodes: Good Catch! https://github.com/volcano-sh/volcano/pull/5098#discussion_r2955011088

agrawalcodes:
> Hello 👋 , Thank you for the contribution! I had some questions. But my main issue is that this wasn't integrated into the helm chart yet, which makes this implementation from product/operability perspective awkward now.
> 
> The other thing is that if we remove a scheduler from the config will that leave nodeShards behind? I didn't find a untit test for this, and from the code it looks like that it does. (the old behaviour was the same I think)

Hi @hajnalmt, thanks for the thorough review!
Added the helm integration and for stale NodeShards on scheduler removal, you understood correctly, it was a real bug.

example/sharding/sharding-config-configmap.yaml
agrawalcodes: @hajnalmt `installer/volcano-agent-scheduler-development.yaml` is generated by `hack/generate-yaml.sh` from helm templates and thus manual edits get overwritten. My understanding is that sharding `ConfigMap` is consumed by the sharding controller (deployed via `volcano-development.yaml`, not this file). Am I correct? If yes, we have 2 options:
1. enable it in volcano-development.yaml generation where the controller lives, or
2. extract the ConfigMap into a separate helm template so it can be rendered into both files independently?

cc @JesseStutler

hajnalmt: 1. is good enough! This is fine! (at least from my side). Thank you!

agrawalcodes: @hajnalmt This is ready for a final review!

example/sharding/sharding-config-configmap.yaml
```diff
@@ -0,0 +1,39 @@
...
+# Deploy with:
+#   kubectl apply -f sharding-config-configmap.yaml
+#
+# The controller is told which ConfigMap to watch via:
+#   --sharding-configmap=<name>
+#   --sharding-configmap-namespace=<namespace>
+#
+# When deployed with Helm the ConfigMap name is {{ .Release.Name }}-sharding-configmap
+# and is automatically wired via values.yaml (sharding_configmap_enable: true).
+apiVersion: v1
```
JesseStutler, line 16: Could you also add a user guide doc to explain all the parameters in this configmap? Only single example files here but there are no any docs explain and refer to this will confuse users, better to write a doc and refer this example

installer/helm/chart/volcano/values.yaml
```diff
@@ -267,6 +267,49 @@ custom:
...
 
+# Sharding ConfigMap configuration.
+# When sharding_configmap_enable is true a ConfigMap is created and the
+# sharding controller watches it for live-reloadable scheduler shard
+# specifications.  Changes take effect without a controller restart.
+#
+# schedulerConfigs defines one entry per scheduler that participates in
+# node sharding.  Each entry specifies:
+#   name               - must match the scheduler's name in NodeShard CRDs
+#   type               - workload class, e.g. "volcano" or "agent"
```
JesseStutler: Actually I think here is scheduler type

installer/helm/chart/volcano/values.yaml
```diff
@@ -267,6 +267,49 @@ custom:
...
+#   maxNodes           - maximum shard size
+#
+# shardSyncPeriod sets the interval between periodic full shard reconciliations.
+# Accepts Go duration strings ("30s", "2m", "1h").
+#
+# enableNodeEventTrigger controls whether node/pod events immediately trigger
+# shard reconciliation in addition to the periodic sync.
+#
+# Example:
+#   sharding_configmap_data: |
```
JesseStutler, line 292: The example values is different in your configmap? I think should be `sharding.yaml` here

pkg/controllers/sharding/config.go
```diff
@@ -21,60 +21,153 @@ import (
...
 func NewShardingControllerOptions() ShardingControllerOptions {
 	controllerOptions := ShardingControllerOptions{
-		// Default scheduler configs will be added by AddFlags
 		ShardSyncPeriod:        60 * time.Second,
 		EnableNodeEventTrigger: true,
 		SchedulerConfigsRaw: []string{
 			"volcano:volcano:0.0:0.6:false:2:100",
 			"agent-scheduler:agent:0.7:1.0:true:2:100",
 		},
+		ConfigMapName:      "",
```
JesseStutler: I think we could add default values for configmapname and configmap namespace

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -129,8 +165,9 @@ func (sc *ShardingController) Initialize(opt *framework.ControllerOption) error
 	sc.assignmentCache = &AssignmentCache{
 		Assignments: make(map[string]*ShardAssignment),
 	}
+	sc.shardSyncPeriodUpdateCh = make(chan struct{}, 1)
 
-	// Parse scheduler configs from options
+	// Parse scheduler configs from options (flag-based fallback).
 	sc.parseSchedulerConfigsFromOptions()
```
JesseStutler: I think this will cause overwrite, we just choose one way, if the configmap is enabled, then we use configmap instead, then we add a flag judgement here only when the flag is not enabled, then we fall back to the scheduler config options( but finally we should deprecate it)

pkg/controllers/sharding/sharding_controller.go
```diff
@@ -677,6 +780,201 @@ func (sc *ShardingController) InitializeWithConfigs(opt *framework.ControllerOpt
 	return nil
 }
 
+// isShardingConfigMap returns true when the object is the sharding ConfigMap
+// that this controller watches.
+func (sc *ShardingController) isShardingConfigMap(obj interface{}) bool {
+	cm, ok := obj.(*corev1.ConfigMap)
+	if !ok {
+		// Handle tombstone objects from the Delete event.
+		if d, ok2 := obj.(cache.DeletedFinalStateUnknown); ok2 {
```
JesseStutler: Please don't use var name like `ok2`, always name it formally

pkg/controllers/sharding/configmap_reload_test.go
```diff
@@ -0,0 +1,445 @@
...
+	controller.configMu.RLock()
+	currentCount := len(controller.schedulerConfigs)
+	controller.configMu.RUnlock()
+	assert.Equal(t, initialCount, currentCount, "config should not change on validation failure")
+}
+
+// TestSyncShards_CleansUpStaleNodeShardsAfterSchedulerRemoved verifies that
+// when a scheduler is removed from the ConfigMap, its NodeShard is deleted
+// during the next syncShards run rather than left behind indefinitely.
+func TestSyncShards_CleansUpStaleNodeShardsAfterSchedulerRemoved(t *testing.T) {
```
JesseStutler, line 385: Also we can add unit tests when a new scheduler is added, and node shard can be successfully created for the new scheduler, and the old shard can successfully updated

installer/helm/chart/volcano/values.yaml
```diff
@@ -267,6 +267,49 @@ custom:
...
+#       - name: volcano
+#         type: volcano
+#         cpuUtilizationMin: 0.0
+#         cpuUtilizationMax: 0.69
+#         preferWarmupNodes: false
+#         minNodes: 1
+#         maxNodes: 100
+#     shardSyncPeriod: "60s"
+#     enableNodeEventTrigger: true
+  sharding_configmap_enable: false
```
JesseStutler: I prefer just enable configmap by default in this version @agrawalcodes , the old flag way is awkward and hard to configure, we should deprecate them in the future

installer/helm/chart/volcano/values.yaml
agrawalcodes: @JesseStutler Can you elaborate more here?

JesseStutler: @agrawalcodes Sorry, I misread it, ignore this comment

example/sharding/sharding-config-configmap.yaml
agrawalcodes: makes sense, will add

installer/helm/chart/volcano/values.yaml
agrawalcodes: Cool

pkg/controllers/sharding/sharding_controller.go
agrawalcodes: Makes sense

JesseStutler:
/approve
/lgtm
Thank you so much!

PR #5100, opened 2026-03-12
fix(proportion): avoid double-counting inqueue resources for jobs wit…

### LINKED ISSUE #5099: proportion plugin double-counts resources for jobs with tasks stuck in Binding, starves queue
### Description

So I've been debugging a scheduling stall that's been driving me crazy and finally nailed it down to the proportion plugin.

**What's happening**

When job-a gets allocated and its tasks move to `Binding` (pods are being created, not yet `Running`), the plugin counts those resources in *both* `attr.allocated` AND `attr.inqueue`. So a queue that's actually at 75% looks full at 150% to the scheduler, and any new job trying to get in gets rejected.

The window is usually short; 20ms to a few seconds, but under any real load it's long enough to matter. And if you have a steady stream of jobs binding, it basically never clears.

**Why it happens**

`AllocatedStatus` includes `Binding`. `ScheduledStatus` doesn't. The PodGroup phase check uses `ScheduledStatus`, so a job with all tasks in `Binding` still reads as `PodGroupInqueue`, meaning it gets counted in `attr.inqueue` on top of `attr.allocated`.

```
Binding → in AllocatedStatus ✅  but NOT in ScheduledStatus
→ tasks counted in attr.allocated ✅
→ PodGroup phase stays Inqueue → also counted in attr.inqueue ❌
```

**To reproduce**

4-CPU queue. Submit job-a with 3×1CPU tasks. While its pods are still in `ContainerCreating`, submit job-b with 1×1CPU. It'll sit Pending even though there's a free CPU. Clears up the moment job-a's pods hit Running.

Scheduler logs during that window will show something like:
```
queue overused, used=6CPU > capacity=4CPU
```
...on a queue that physically has 4 CPUs in use at most.

Without fix:
Image

the bug: `job-b` times out waiting to be scheduled. It has capacity (queue: 4 CPU, job-a uses 3 CPU, job-b needs 1 CPU) but the scheduler incorrectly reports the queue as over-capacity (6 CPU used = 3 allocated + 3 double-counted inqueue), so `job-b` is permanently stuck in Pending.

With fix:
Image

after the fix: `job-b` is scheduled correctly because the already-allocated 3 CPU is deducted from the inqueue calculation, giving 0 inqueue + 3 allocated = 3 CPU ≤ 4 CPU capacity.

I reproduced this on v1.14.1 (the latest release), against the master branch at commit 62ebc877b. The bug exists in the current master as well.

**Unit test**

The race window is small on a real cluster so I wrote a test that reproduces it deterministically:

```bash
go test ./pkg/scheduler/plugins/proportion/... \
  -run TestNoDoubleCountingForInqueueJobWithBindingTasks -v -count=1 -timeout 30s
```

Cycle 1 allocates job-a (3×1CPU) so its tasks enter `Binding` while the PodGroup stays `Inqueue`. Cycle 2 tries to enqueue job-b (1×1CPU) on the same 4-CPU queue. With the bug, `attr.allocated=3 + attr.inqueue=3 = 6 > 4` and job-b gets blocked. With the fix, `attr.inqueue=0` and job-b binds fine.

**The fix**

The `PodGroupRunning` branch already handles this correctly using `util.GetInqueueResource` which does `max(0, minResources - already_allocated)`. The `PodGroupInqueue` branch just needs the same treatment:

```go
// currently
attr.inqueue.Add(job.DeductSchGatedResources(job.GetMinResources()))

// should be
inqueued := util.GetInqueueResource(job, job.Allocated)
attr.inqueue.Add(job.DeductSchGatedResources(inqueued))
```

Opening a PR for the fix!😁

Aman-Cool:
#### What type of PR is this?

bug fix

While investigating some scheduling issues I noticed that jobs can get stuck in Pending even when there's clearly enough queue capacity available. Tracked it down to a double-counting bug in the proportion plugin.

When a gang-scheduled job's tasks get allocated, they move to `Binding` state in the scheduler cache. The problem is that `Binding` is part of `AllocatedStatus` but not `ScheduledStatus`, so the PodGroup stays `Inqueue` until the tasks actually reach `Running`. During this window, the proportion plugin counts those task resources in both `attr.allocated` and `attr.inqueue` — the full `minResources` gets added to `attr.inqueue` without deducting what's already allocated.

So if job-a uses 3 CPU on a 4 CPU queue, while its tasks are binding the queue looks like it's using 6 CPU, and any new job gets rejected even though there's 1 CPU free.

The `PodGroupRunning` branch already handles this correctly using `GetInqueueResource(job, job.Allocated)` which returns `max(0, minResources - already_allocated)`. Applied the same logic to the `PodGroupInqueue` branch.

Fixes #5099 

Added a regression test that runs two scheduler cycles against a shared cache, cycle 1 allocates job-a (3×1CPU) leaving its tasks in `Binding`, cycle 2 tries to enqueue job-b (1×1CPU) on a 4 CPU queue. Without the fix the test times out waiting for job-b to be bound. With the fix it passes.

```release-note
Fix proportion plugin double-counting inqueue resources for jobs with tasks in Binding state, which could cause queues to appear over-capacity and block schedulable jobs.
```

### DIFF
--- a/pkg/scheduler/plugins/proportion/proportion_test.go
+++ b/pkg/scheduler/plugins/proportion/proportion_test.go
@@ -47,6 +47,24 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
+// waitForBinds reads exactly n items from ch within deadline and returns them.
+// It fails the test if fewer than n items arrive within the timeout.
+func waitForBinds(t *testing.T, ch <-chan string, n int, timeout time.Duration) map[string]bool {
+	t.Helper()
+	got := make(map[string]bool, n)
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+	for i := 0; i < n; i++ {
+		select {
+		case bind := <-ch:
+			got[bind] = true
+		case <-timer.C:
+			t.Fatalf("timeout waiting for bind #%d/%d (got so far: %v)", i+1, n, got)
+		}
+	}
+	return got
+}
+
 func TestMain(m *testing.M) {
 	options.Default()
 	os.Exit(m.Run())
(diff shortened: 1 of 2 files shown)


Aman-Cool:
Hey @hajnalmt @JesseStutler @hzxuzhonghu, Traced this one down while looking at some scheduling delays; jobs were getting stuck in Pending even with free queue capacity. Turns out the proportion plugin was counting the same resources twice during the binding window. Let me know if you'd like any changes😄.

Aman-Cool:
@hajnalmt, Thanks so much! Really glad it checks out!😄 
I'll take a look at the capacity plugin and open a follow-up PR once @JesseStutler weighs in!

hajnalmt:
/kind scheduling
/kind bug

:+1:

pkg/scheduler/plugins/proportion/proportion_test.go
```diff
@@ -47,6 +47,24 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
+// waitForBinds reads exactly n items from ch within deadline and returns them.
+// It fails the test if fewer than n items arrive within the timeout.
+func waitForBinds(t *testing.T, ch <-chan string, n int, timeout time.Duration) map[string]bool {
```
hajnalmt, line 52: I wanted to mention that this shouldn't be here rather than in uthelper, but this is true for getLocalMetrics and getWorkerAffinity too. If you create a new PR for the capacity plugin please move these there too. You will need to have waitForBinds at a common a place anyway.

Also I am thinking that a big chunk of your unit test should be outsourced too. It does its job and I understand that your test does not align with our TestCommonStruct at all, due to the double session openings, but it looks sloppy 😄 Probably we need have a common way for double session tests.

Aman-Cool: Makes sense! will keep this in mind while working on the capacity PR😄.
Thanks for bringing this up😁.

PR #5103, opened 2026-03-12
fix: clean up stale DRA inFlightAllocations at session start to preve…

dafu-wu:
#### What type of PR is this?

/kind bugfix

In Volcano's batch scheduling model, a scheduling session may terminate without executing the `Unreserve` path for every reserved task. When this happens, the DRA plugin's `inFlightAllocations` remain in the shared `DRAManager`'s `claimTracker`.

Because `DRAManager` is shared across scheduling sessions, these stale `inFlightAllocations` persist into subsequent sessions. As a result, `ListAllAllocatedDevices()` includes devices from these stale entries, causing the allocator to incorrectly treat devices as already occupied. This can lead to scheduling failures such as:

This behavior differs from the upstream Kubernetes scheduler, where the `Reserve` and `Unreserve` phases are strictly paired for each pod. In contrast, Volcano's gang scheduling may abandon an entire scheduling session without invoking `Unreserve`, leaving stale reservation state behind.

To address this issue, this PR introduces a cleanup step that runs at the beginning of every new scheduling session. The cleanup removes all stale `inFlightAllocations` entries from the `claimTracker`.

The cleanup removes pending allocations for **all claims**, not only unallocated ones. Even if a claim is already allocated (`Status.Allocation != nil`), a redundant `inFlightAllocation` may still exist and cause double-counting in `ListAllAllocatedDevices()`.

By proactively cleaning these stale entries at session start, we ensure that device allocation state accurately reflects the real cluster state and prevent false resource exhaustion during scheduling.

Fixes #

This change ensures that stale `inFlightAllocations` leaked from previous scheduling sessions do not interfere with subsequent scheduling cycles.

```release-note
Fix an issue where stale DRA inFlightAllocations from previous scheduling sessions could cause "cannot allocate all claims" scheduling failures.

### DIFF
--- a/pkg/scheduler/plugins/predicates/dra_cleanup.go
+++ b/pkg/scheduler/plugins/predicates/dra_cleanup.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2024 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package predicates
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/dynamic-resource-allocation/resourceclaim"
+	"k8s.io/klog/v2"
+	fwk "k8s.io/kube-scheduler/framework"
+
+	"volcano.sh/volcano/pkg/scheduler/api"
+)
+
+// cleanupStaleDRAPendingAllocations removes stale inFlightAllocations from the DRA ResourceClaimTracker.
+//
+// In upstream k8s scheduler, inFlightAllocations are added during Reserve (when a claim allocation is computed)
+// and removed either in Unreserve (on failure) or after PreBind succeeds (via AssumeClaimAfterAPICall + informer update).
+//
+// In Volcano's batch scheduling model, if a scheduling session ends (e.g., due to a job not being fully schedulable)
+// without going through the Unreserve path for every reserved task, the inFlightAllocations entries remain in the
+// shared DRAManager's claimTracker. Since the DRAManager is shared across sessions, these stale entries persist
+// and cause subsequent scheduling cycles to fail with:
+//
+//	"resource claim <ns>/<name> is in the process of being allocated"
+//
+// or cause the DRA Filter to report "cannot allocate all claims" because ListAllAllocatedDevices() includes
+// devices from stale inFlightAllocations, making the allocator think devices are occupied when they are not.
+//
+// This function iterates over all pods' resource claims and removes any stale pending allocations
+// at the beginning of each new scheduling session, ensuring a clean state.
+func cleanupStaleDRAPendingAllocations(draManager fwk.SharedDRAManager, jobs map[api.JobID]*api.JobInfo) {
+	if draManager == nil {
+		return
+	}
+
+	claimTracker := draManager.ResourceClaims()
+	if claimTracker == nil {
+		return
+	}
+
+	// At the beginning of each new scheduling session, ALL inFlightAllocations from the previous session
+	// should be considered stale. This is because:
+	// 1. Successfully bound tasks: their claims have been written to the API server and the informer cache
+	//    has been updated, so the inFlightAllocation is no longer needed.
+	// 2. Failed/discarded tasks: their claims were never actually allocated, so the inFlightAllocation
+	//    is stale and must be removed.
+	//
+	// We iterate over ALL jobs (not just pending ones) to catch claims from tasks in any state,
+	// including those that were allocated in a previous session but whose inFlightAllocation was
+	// not properly cleaned up.
+	cleanedCount := 0
+	cleanedAllocatedCount := 0
+	cleanedUnallocatedCount := 0
+	for _, job := range jobs {
+		for _, task := range job.Tasks {
+			pod := task.Pod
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dafu-wu** [comment]
@JesseStutler @hzxuzhonghu can you help review it？

**Comment by JesseStutler** [comment]
Sorry I didn't get the point why this PR was raised, do you mean when the gang allocation failed in the session, all the pod in this gang group didn't call `Unreserve`? But when we're trying to rollback, the scheduler will execute all the Unreserve plugins in the DeallocationFunc: https://github.com/volcano-sh/volcano/blob/78b480781ac8635924441f53234e62ff4b634fbe/pkg/scheduler/plugins/predicates/predicates.go#L264, so what's the root cause of your issue? @dafu-wu

**Comment by dafu-wu** [comment]
> Sorry I didn't get the point why this PR was raised, do you mean when the gang allocation failed in the session, all the pod in this gang group didn't call `Unreserve`? But when we're trying to rollback, the scheduler will execute all the Unreserve plugins in the DeallocationFunc:
> 
> https://github.com/volcano-sh/volcano/blob/78b480781ac8635924441f53234e62ff4b634fbe/pkg/scheduler/plugins/predicates/predicates.go#L264
> 
> , so what's the root cause of your issue? @dafu-wu

@JesseStutler When a scheduling session is abandoned (e.g., due to gang scheduling failure), DeallocateFunc is not invoked. As a result, runUnReservePlugins is never executed and the inFlightAllocation entries leak. Therefore, we proactively clean them up at the start of each new session.

**No matter what causes the leak, I believe it’s necessary to perform a cleanup at the end of each session.**

File: pkg/scheduler/plugins/predicates/dra_cleanup.go [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+	"k8s.io/klog/v2"
+	fwk "k8s.io/kube-scheduler/framework"
+
+	"volcano.sh/volcano/pkg/scheduler/api"
+)
+
+// cleanupStaleDRAPendingAllocations removes stale inFlightAllocations from the DRA ResourceClaimTracker.
+//
+// In upstream k8s scheduler, inFlightAllocations are added during Reserve (when a claim allocation is computed)
+// and removed either in Unreserve (on failure) or after PreBind succeeds (via AssumeClaimAfterAPICall + informer update).
```
JesseStutler, line 31: But PreBind is in another goroutine differs from session, so if PreBind doesn't finish but next round session is starting, and you clean up the inFlightAllocations, will there be duplicate allocation on the same device?

JesseStutler:
> > Sorry I didn't get the point why this PR was raised, do you mean when the gang allocation failed in the session, all the pod in this gang group didn't call `Unreserve`? But when we're trying to rollback, the scheduler will execute all the Unreserve plugins in the DeallocationFunc:
> > https://github.com/volcano-sh/volcano/blob/78b480781ac8635924441f53234e62ff4b634fbe/pkg/scheduler/plugins/predicates/predicates.go#L264
> > 
> > , so what's the root cause of your issue? @dafu-wu
> 
> @JesseStutler When a scheduling session is abandoned (e.g., due to gang scheduling failure), DeallocateFunc is not invoked. As a result, runUnReservePlugins is never executed and the inFlightAllocation entries leak. Therefore, we proactively clean them up at the start of each new session.
> 
> **No matter what causes the leak, I believe it’s necessary to perform a cleanup at the end of each session.**

No DeAllocateFunc will be called because the statement will be discard, and in statement operation, the `unallocate` will call DeAllocateFunc

dafu-wu:
> > > Sorry I didn't get the point why this PR was raised, do you mean when the gang allocation failed in the session, all the pod in this gang group didn't call `Unreserve`? But when we're trying to rollback, the scheduler will execute all the Unreserve plugins in the DeallocationFunc:
> > > https://github.com/volcano-sh/volcano/blob/78b480781ac8635924441f53234e62ff4b634fbe/pkg/scheduler/plugins/predicates/predicates.go#L264
> > > 
> > > , so what's the root cause of your issue? @dafu-wu
> > 
> > 
> > @JesseStutler When a scheduling session is abandoned (e.g., due to gang scheduling failure), DeallocateFunc is not invoked. As a result, runUnReservePlugins is never executed and the inFlightAllocation entries leak. Therefore, we proactively clean them up at the start of each new session.
> > **No matter what causes the leak, I believe it’s necessary to perform a cleanup at the end of each session.**
> 
> No DeAllocateFunc will be called because the statement will be discard, and in statement operation, the `unallocate` will call DeAllocateFunc

@JesseStutler 
runUnReservePlugins is a synchronous cleanup mechanism within a scheduling session (triggered during Discard / UnAllocate). However, cleanup after Commit is asynchronous—it happens in the PreBind defer of another goroutine—so runUnReservePlugins does not cover that asynchronous path.

RemoveClaimPendingAllocation is placed in the defer of bindClaim, but bindClaim is not guaranteed to be invoked. If any step before the DRA stage in the PreBind chain fails, the DRA cleanup logic will be skipped. As a result, these pending allocations can accumulate over time.

pkg/scheduler/plugins/predicates/dra_cleanup.go
```diff
@@ -0,0 +1,119 @@
...
+	// 2. Failed/discarded tasks: their claims were never actually allocated, so the inFlightAllocation
+	//    is stale and must be removed.
+	//
+	// We iterate over ALL jobs (not just pending ones) to catch claims from tasks in any state,
+	// including those that were allocated in a previous session but whose inFlightAllocation was
+	// not properly cleaned up.
+	cleanedCount := 0
+	cleanedAllocatedCount := 0
+	cleanedUnallocatedCount := 0
+	for _, job := range jobs {
```
hzxuzhonghu, line 68: This cleanup only walks claims that are reachable from `ssn.Jobs`. That is narrower than the stale-state problem described in the PR. `ResourceClaimTracker` already exposes `List()`, and leaked `inFlightAllocations` can belong to claims whose jobs or pods are no longer present in the current session snapshot. In that case this sweep will miss them, and `ListAllAllocatedDevices()` can still be poisoned in later sessions. I think the cleanup needs to iterate the claim tracker itself instead of the session jobs map.

PR #5126, opened 2026-03-23
  fix(scheduler): guard within-job preemption commit with JobPipeline check

Sanchit2662:
#### What type of PR is this?

/kind bug

Fixes within-job preemption to only commit evictions when the job actually becomes pipelined, preventing unnecessary victim pod churn and API-server load.

Previously, each task preemption created a separate Statement and committed immediately, regardless of whether the overall job could schedule. This caused evicted pods to be recreated by the job controller, leading to thrashing loops when resources were insufficient.

Now, we accumulate all preemptions into one Statement per job and only commit if `ssn.JobPipelined(job)` returns true — matching the existing cross-queue preemption pattern.

Fixes #NA

The fix mirrors the cross-queue preemption logic at lines 215-221. The redundant `_, found := preemptorTasks[job.UID]` check was also removed (always true after the outer map check).

```release-note
Fixed within-job preemption to prevent unnecessary victim pod evictions when the preemptor job cannot become schedulable. This reduces pod churn and API-server load in preemption-heavy scenarios.
```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -279,7 +279,6 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 				}
 
 				// Only commit if preemption was successful, otherwise discard to rollback evictions.
-				// This is consistent with between-job preemption which checks JobPipelined before committing.
 				if !assigned {
 					stmt.Discard()
 					break


Sanchit2662:
@JesseStutler @hajnalmt @hzxuzhonghu,Could you please review this fix for within-job preemption thrashing?

hajnalmt (changes requested):
Hello,
Thank you for the contribution! What are you running into exactly?
If I understand correctly, you managed to get a case where assigned is true in that loop, but the job is not Pipelined. How?  Can we write a unit test for this somehow?

pkg/scheduler/actions/preempt/preempt.go
```diff
@@ -236,18 +236,18 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 				}
 				preemptorTasks[job.UID].Push(task)
 			}
-			for {
-				if _, found := preemptorTasks[job.UID]; !found {
```
hajnalmt: Also why did you remove this guard from loop? It was a defensive measure to not run into nil pointer exceptions calling `.Empty()`

Sanchit2662: You're right, I shouldn't have removed it. Even though preemptorTasks[job.UID] is always set on the line just above
the loop, removing a defensive nil check is unnecessary risk. I've restored it.

Sanchit2662:
I found this while reading through the preemption code and noticed the within-job path didn't have the same JobPipelined guard that cross-queue preemption already uses. Here's the scenario: say a job has MinAvailable=3, with 2 tasks already running (one preemptable, one not) and 2 high-priority tasks pending. The inner loop evicts the preemptable running task and pipelines the first pending task assigned=true. But now we're at 1 running + 1 pipelined = 2, still less than MinAvailable=3. The second pending task finds no more preemptable victims in the same job, so assigned=false. With the old code, stmt.Commit() fired right after that first assigned=true, permanently evicting the victim even though the job could never become schedulable.

I've added a unit test "within-job preemption: discard evictions when job cannot reach MinAvailable" that reproduces
this exactly  with the old code it gives ExpectEvictNum=1, with the fix it gives ExpectEvictNum=0.

Sanchit2662:
@hajnalmt , shall i modify anything on this pr?

pkg/scheduler/actions/preempt/preempt_test.go
```diff
@@ -330,6 +330,52 @@ func TestPreempt(t *testing.T) {
...
+				// Running, preemptable — the only within-job victim available.
+				util.BuildPod("c1", "running-preemptable", "n1", v1.PodRunning,
+					api.BuildResourceList("1", "1G"), "pg1",
+					map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				// Running, NOT preemptable — shielded from within-job preemption.
+				util.BuildPod("c1", "running-protected", "n1", v1.PodRunning,
+					api.BuildResourceList("1", "1G"), "pg1",
+					map[string]string{schedulingv1beta1.PodPreemptable: "false"}, make(map[string]string)),
+				// Pending high-priority preemptor #1 — evicts running-preemptable (assigned=true).
+				util.BuildPodWithPriority("c1", "pending-high-1", "", v1.PodPending,
```
hajnalmt: I think this pod should preempt the lower priority one, as it has higher priority. I don't think the `Job.Pipelined` gate is needed at all.

hajnalmt (commented):

I understand your change now, although I think it is not correct. In this case, the high-priority pod should preempt the low-priority pod, and we can disregard the Job.Pipelined state. Since the preemptor has a higher priority, it should be scheduled first.

PR #5129, opened 2026-03-23
enhance: add detailed logging to gang plugin for debugging

gitGurugu:
## Summary

- Add initialization log in `OnSessionOpen` with job and queue counts
- Add state tracking log in `JobReadyFn` showing ready/unready reasons
- Add session close summary log with ready/unready/terminated job counts
- Enhance unschedulable path with pending task names in error message

## Test plan

- [ ] Build passes on Linux
- [ ] Gang scheduling behavior unchanged
- [ ] Logs visible at appropriate verbosity levels (V3, V4)

### DIFF
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -182,10 +182,14 @@ func (gp *gangPlugin) OnSessionOpen(ssn *framework.Session) {
 
 	ssn.AddJobReadyFn(gp.Name(), func(obj interface{}) bool {
 		ji := obj.(*api.JobInfo)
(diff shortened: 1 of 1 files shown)


gitGurugu:
Thanks for the detailed review! All three points have been addressed:

Renamed the log field from isReady to minAvailableReady for clarity
Truncated pendingTaskNames to 5 entries in PodGroupCondition messages to avoid unbounded growth; full list is now only in klog (which is not persisted)
Fixed OnSessionClose summary: total now reflects only jobs with tasks, and IsJobTerminated is checked first so categories are mutually exclusive
PTAL, thanks!

PR #5135, opened 2026-03-28
Fix queue status conflicts in multi-scheduler by adding per-scheduler allocation tracking

### LINKED ISSUE #5084: Queue status update conflicts and .Status.Allocated overwrites under multi-scheduler architecture
### Description

In Volcano v1.14.0, the introduction of the Agent Scheduler and Sharding Controller has evolved Volcano towards a unified scheduling platform. This architecture allows multiple schedulers to run concurrently and handle different workloads or different node shards. 

However, Queue is currently scoped as a cluster-level resource. Under the existing implementation, every scheduler actively watches all queues and explicitly updates their statuses at the end of every scheduling session. Because a specific scheduler (when configured with sharding) only processes nodes and jobs belonging to its own bounded shard/workload, this local scheduling context causes two major problems when it attempts to update the global queue:

- API Server Update Conflicts: Multiple schedulers (e.g., Volcano Scheduler and Agent Scheduler) inevitably try to fetch and update the same queue's status concurrently at the end of their respective sessions, leading to persistent Conflict update errors on the API Server.
- Status Overwrites & Inaccuracy: Currently, the updateQueueStatus logic computes the allocated resources based strictly on the jobs and tasks within the current scheduler's local session. Thus, the resulting `queue.Status.Allocated` solely reflects the resource allocation in that specific scheduler's jurisdiction. When different schedulers push their status updates, they overwrite each other. As a result, `.Status.Allocated` constantly flaps and fails to accurately reflect the true total resource allocation for the queue across all shards.

Code Reference:

The queue status update logic at the close of a session:
https://github.com/volcano-sh/volcano/blob/c2f203807f90156c773bdf614391fe6921100fc3/pkg/scheduler/framework/session.go#L492-L543

> Even in the current multi-scheduler model that uses a consistent hash ring for simple partitioning, this problem also exists.

1. Deploy Volcano v1.14.0 and enable the multi-scheduler architecture (e.g., run both volcano-scheduler and agent-scheduler simultaneously alongside the sharding-controller).
2. Create a generic Queue and submit diverse workloads (e.g., short-lived tasks to the agent scheduler, batch jobs to the volcano scheduler) pointing to the same queue.
3. Observe the API server logs and the Queue resource status. You will intermittently see Conflict errors when schedulers attempt to update the queue status. Concurrently, kubectl get queue <name> -o yaml will show .status.allocated values fluctuating/overwritten, rather than smoothly aggregating the combined usage from both schedulers.

Or you can simply follow this guide to deploy multi-schedulers using a consistent hash: https://github.com/volcano-sh/volcano/blob/master/docs/design/deploy-multi-volcano-schedulers-without-using-selector.md, and submit several pods to the **same queue** but under different schedulers

#### Expect results

The Queue status and its Allocated resources should serve as a consistent source of truth, accurately summarizing the global allocation footprint across all node shards and schedulers without encountering race conditions during updates.

Here are three potential solutions to discuss and resolve this:

- Solution 1: Introduce a Centralized Queue Controller
Remove the queue status update permission from individual schedulers. Use the queue-controller in vc-controller to record and update the queue status( Currently queue controller only update the queue phase now, because there are conflict bugs in the previous version and we remove the allocated updating codes in queue controller) . This controller would globally watch all PodGroups / Pods and their resource requests, calculate the actual allocated resources in each Queue, and periodically update the Queue statuses.
  - *Pros*: Completely resolves update conflicts, provides a consistent global view, and improves scheduler performance by reducing API server writes.
  - *Cons*: Increases the computation overhead on the controller manager and may have slight sync delays compared to the scheduler's local snapshot.

- Solution 2: Per-Scheduler / Per-Shard Allocation Records in Queue Status
Extend `Queue.Status.Allocated` or introduce a new field like `Queue.Status.Allocations` as a map/list where each scheduler updates its own shard's allocation independently (e.g., status.shardAllocated["agent-scheduler"] = ...). Updates can be done via Patch rather than Update to avoid conflicts.
  - *Pros*: Lower refactoring cost, and schedulers remain decoupled and independent.
  - *Cons*: The API definition becomes more complex, the status payload might bloat with dynamic sharding, and client tools (like vcctl) would need extra logic to sum the total usage.

- Solution 3: Scheduler Affinity/Constraint for Queues
Add a schedulerName constraint at the Queue level, assigning each queue to be managed exclusively by a specific scheduler. Schedulers would only watch and calculate statuses for their own queues.
  - *Pros*: Minimal architectural changes required.
  - *Cons*: Weakens tenant isolation and capacity sharing across different workloads. If a tenant needs to run both batch jobs (Volcano Scheduler) and agent workloads (Agent Scheduler), they would be forced to create two separate queues, preventing unified resource capacity limits and sharing constraints at the single Queue-wide sharing and quota limits at the Queue level.

v1.14.1

_No response_

rishisulakhe:
#### What type of PR is this?
/kind bug

Fixes #5084 (Solution 2)

When running multiple schedulers in Volcano v1.14.0+, each scheduler computes Queue.Status.Allocated from only its own local view, then does a full UpdateStatus() call. This causes two issues:

- 409 Conflict errors — schedulers race to update the same queue object
- Inaccurate Allocated values —each scheduler overwrites the other's data, so the field never reflects the true total across all schedulers

This happens both in the Agent Scheduler + Volcano Scheduler setup and in hash-ring based multi-scheduler deployments.

### What changed and why
New API field — `QueueStatus.SchedulerAllocations`

Added a `map[string]SchedulerAllocation` field where each key is a scheduler's identity (scheduler name, or pod name for hash-ring deployments). Each entry holds that scheduler's Allocated resources and a `LastUpdateTime` timestamp. This lets multiple schedulers write to the same queue without stepping on each other.

Scheduler side — Instead of calling `UpdateStatus()` (which replaces the entire status), each scheduler now calls `ApplyStatus()` (SSA) with a unique `FieldManager` per scheduler instance. This means scheduler-A writing its entry never conflicts with scheduler-B writing its entry — Kubernetes SSA handles this natively.

The queue controller's `syncQueue()` now sums up all active `SchedulerAllocations` entries into the existing `Status.Allocated `field. This keeps backward compatibility — any tool reading `Status.Allocated `still sees the correct total. Entries older than 5 minutes are considered stale and are excluded from the sum, then garbage-collected via merge patch.

The existing `Status.Allocated` field is preserved and still works exactly as before. Single-scheduler deployments see no behavior change — there's just one entry in the map, and the total equals that entry


rishisulakhe:
Hi @hajnalmt @JesseStutler Could you pls review this?

Issue #5133, opened 2026-03-24
[Feature] Decoupling HyperNode and PodGroup Controllers: Evolving Toward a Diversified AI Orchestration Ecosystem


wangyang0616:
### What is the problem you're trying to solve

As large-scale distributed AI training and high-concurrency inference become the norm, the underlying infrastructure's requirements for **Hardware Topology Awareness** and **Resource Coordination (Gang Scheduling)** have undergone fundamental shifts.

Currently, Volcano has established deep technical expertise in **HyperNode** (heterogeneous topology discovery) and **PodGroup** (scheduling primitives). However, the existing monolithic architecture binds these capabilities strictly to the vc-controller-manager, limiting their value in the modern AI ecosystem:

- **High Barrier for AI Ecosystem Integration:** Many developers with established custom schedulers urgently need HyperNode’s topology capabilities or PodGroup’s coordination primitives. The deep coupling with Volcano’s core batch logic (Volcano-scheduler) creates a significant integration hurdle.

- **Limited Technology Democratization:** Our research breakthroughs in NPU/GPU/RDMA topology awareness and large-scale parallel job synchronization should benefit the broader Kubernetes community, even for those not utilizing the full volcano-scheduler stack.

- **AI-Native Evolution Requirements:** Facing an increasingly diverse array of AI chip architectures, **HyperNode aims to collaborate with more hardware vendors to co-build simple and standardized Network Topology Domain Primitives.** This will support **automated discovery and scheduling management** of workload topologies across different architectures. Such rapid iteration cycles often conflict with the stable release cadence of the Volcano main branch.

### Describe the solution you'd like

We propose a "Two-Stage" strategy to achieve both logical and physical decoupling, evolving HyperNode, PodGroup, and Queue into standalone, mix-and-match AI infrastructure components.

##### Stage 1: Logical Decoupling & Config-Driven Execution (Soft Decoupling)

- **Pluggable Controllers:** Abstract a unified interface in pkg/controllers/framework to "library-ify" core logic, removing hard dependencies on the main controller.

- **Granular Startup Control:** Introduce a --controllers flag (e.g., --controllers=hypernode,podgroup). Users can run a minimal process executing only the required AI coordination logic.

- **Resource Footprint Optimization:** In standalone mode, dynamically disable irrelevant Informers to minimize overhead and pressure on the API Server.

##### Stage 2: Physical Isolation & Staging Mode (Hard Decoupling)
[...]

PR #5136, opened 2026-03-30
docs: add proposal for HyperNode controller modularization and standalone deployment

wangyang0616:
#### What type of PR is this?
Feature proposal

This PR implements the physical decoupling of HyperNode and PodGroup controllers as proposed in #5133. It transitions these core AI scheduling primitives into a Staging Mode, enabling independent evolution and atomic delivery without fragmenting the unified API library.

**Key Changes:**
- **Physical Relocation:** Moved HyperNode and PodGroup logic from pkg/controllers/ to staging/src/volcano.sh/.
- **Independent Modules:** Introduced standalone go.mod for each primitive in staging/, supporting independent compilation and lightweight image builds.
- **Dependency Cleanup:** Decoupled primitives from vc-job and other business logic to ensure a clean, hardware-vendor-friendly contribution path.
- **Dual-Delivery Support:**
    - **Monolithic:** vc-controller-manager continues to import and manage all controllers (Backward Compatible).
    - **Atomic:** Enabled building standalone vc-hypernode and vc-podgroup binaries/images.

Refs #5133 

- **Backward Compatibility:** Confirmed vc-controller-manager --controllers=* functions identically to the previous version.
- **Build Validation:** Verified that both the monolithic controller-manager and independent staging modules pass compilation.
- **Unit Tests:** All migrated tests for PodGroup and HyperNode passed within the new staging structure.

### DIFF
--- a/docs/design/hypernode-controller-standalone.md
+++ b/docs/design/hypernode-controller-standalone.md
@@ -1,0 +1,245 @@
+# HyperNode Controller Standalone Deployment
+
+Author: wangyang0616 · 2026-08-09
+
+---
+
+## 1. Summary
+
+HyperNode is Volcano's standard abstraction for hierarchical network topology. Today, the HyperNode controller can run only as part of `vc-controller-manager`. This model works well for users deploying the complete Volcano stack, but it makes HyperNode difficult to adopt independently for users who already operate their own scheduler or need only topology discovery and management.
+
+This proposal provides two deployment modes backed by the same implementation:
+
+- **Controller-manager mode — default:** HyperNode continues to run in `vc-controller-manager`; existing users keep their current deployment model.
+- **Standalone mode — optional:** HyperNode runs as a dedicated Volcano component for topology-focused environments or environments using a custom scheduler.
+
+The central decision is to **decouple HyperNode deployment from other controllers without splitting HyperNode into a separate product**. The deployment boundary is expanded, while the API, source ownership, community governance, and release boundaries remain aligned with Volcano:
+
+- users can choose integrated or standalone deployment;
+- both deployment modes share one implementation and produce semantically equivalent HyperNode resources;
+- the HyperNode API remains the ecosystem integration boundary;
+- source, versioning, and releases remain under the Volcano community.
+
+This gives users an independent operating model without creating a second implementation, compatibility contract, or maintenance lifecycle. This proposal neither introduces nor plans a separate HyperNode Go module or repository.
+
+## 2. Background and Community Value
+
+Communication topology is becoming a common scheduling input for large-scale AI training and inference. Different accelerator generations, network architectures, and vendor management systems also require topology discovery to evolve independently from scheduling policy.
+
+The community therefore needs to support two legitimate adoption models:
+
+1. users adopting Volcano as a complete scheduling system;
+2. users adopting only the HyperNode topology capability alongside an existing scheduler or AI platform.
+
+The current aggregate controller process serves the first model but creates unnecessary operational coupling for the second. Volcano does not provide an official way to deploy, roll out, scale, or manage RBAC for HyperNode independently, so topology-only users must still operate unrelated controller-manager capabilities.
+
+The existing `--controllers` flag selects which controller logic runs in the process, but it does not provide an independent binary, image, Deployment, fault domain, or RBAC boundary. It therefore cannot replace an officially supported standalone runtime.
+
+This is primarily a community adoption and long-term maintenance problem, not a source-code layout problem. Volcano already has a clear resource boundary: the HyperNode controller discovers and maintains HyperNode resources, and schedulers consume those resources through `volcano.sh/apis`. What is missing is an officially supported standalone deployment built around that boundary.
+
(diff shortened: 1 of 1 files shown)


wangyang0616:
@hzxuzhonghu @JesseStutler @hajnalmt Please help review this as well. Thanks.

hzxuzhonghu:
One scope/traceability concern: this PR currently says `Fixes #5133`, but the diff is proposal-only (design doc) and #5133 is functioning as the umbrella tracking issue for multiple follow-up items. If this merges as-is, it would incorrectly close the umbrella issue before the implementation work lands.

I suggest changing the linkage to something like `Part of #5133` or `Refs #5133` instead of `Fixes #5133`.

wangyang0616:
> One scope/traceability concern: this PR currently says `Fixes #5133`, but the diff is proposal-only (design doc) and #5133 is functioning as the umbrella tracking issue for multiple follow-up items. If this merges as-is, it would incorrectly close the umbrella issue before the implementation work lands.
> 
> I suggest changing the linkage to something like `Part of #5133` or `Refs #5133` instead of `Fixes #5133`.

Yes, it has been updated.

docs/design/hypernode&podgroup-controller-independent.md
```diff
@@ -0,0 +1,73 @@
...
+
+---
+
+## 1. Summary
+This proposal suggests a structural refactoring of the Volcano controller architecture to achieve physical decoupling of **HyperNode** and **PodGroup**. By leveraging a **Staging Mode** and independent Go modules—while maintaining a **Unified Volcano API Library**—we aim to transform core scheduling units into atomic primitives. This evolution empowers hardware ecosystem co-construction and supports complex AI scenarios such as **PD (Prefill & Decoding) separation** and **multi-dimensional Gang scheduling**.
+
+## 2. Goals & Non-Goals
+
+### 2.1 Goals
+* **Physical Isolation**: Move HyperNode and PodGroup logic into `staging/` with independent `go.mod` files to support standalone compilation.
```
hzxuzhonghu, line 15: My confusion is why move to staging instead of separating a cmd entry?

docs/design/hypernode&podgroup-controller-independent.md
```diff
@@ -0,0 +1,73 @@
...
+* **Atomic Delivery**: Provide the capability to build and deploy lightweight, independent images (`vc-hypernode`, `vc-podgroup`) for inference and edge scenarios.
+* **Ecosystem Empowerment**: Simplify the contribution path for hardware vendors to implement topology discovery within a decoupled HyperNode module.
+* **Maintain Monolithic Compatibility**: Ensure the existing `vc-controller-manager` remains the primary delivery vehicle with zero changes to its default behavior.
```
hzxuzhonghu, line 18: so this means volcano controller-manager still compiles the hypernode and podgroup controller, right?

docs/design/hypernode&podgroup-controller-independent.md
wangyang0616: 1. Provides a separate Go mod file, avoiding excessive external dependencies.
2. The compiled image package only contains HyperNode-related source code, making it more lightweight.
3. HyperNode management and modification are more independent. Once mature, it can be spun off into an independent repository for operation, and historical commit records can be retained.

docs/design/hypernode&podgroup-controller-independent.md
wangyang0616: Yes

docs/design/hypernode&podgroup-controller-independent.md
JesseStutler: @wangyang0616 So you mean we need to release and maintain a new `volcano-sh/hypernode` dependency, right? Do we really need to introduce and publish a new `volcano.sh/hypernode` dependency at this stage? If the main goal is standalone deployment, a dedicated command, image, and installer within the Volcano repository should already be sufficient. 

A separately consumable module introduces additional governance and maintenance costs, including repository creation, CI, synchronization, versioning, release workflows, and API compatibility. Could you clarify the concrete benefits to Volcano, and whether there is an alternative way to distribute the module without creating a new repository?

If there is no confirmed need for external dependency consumption yet, we could consider standalone deployment first and evaluate module extraction separately later. 

So currently is it enough for proposed layout that needs to modify?
```
volcano/
├── cmd/
│   ├── controller-manager/
│   └── hypernode-controller-manager/
├── installer/  (should create seperate yaml for hypernode-controller)
```

wangyang0616: You are right. There is currently no clear requirement that justifies creating a separate repository for HyperNode. At this stage, we should focus on enabling HyperNode to be deployed and operated independently.

Whether HyperNode should have its own repository in the future should be driven by concrete and recurring user needs. The key question is whether community users must consume HyperNode directly as a third-party Go SDK. If the official image, upstream contributions, private forks, and the HyperNode API already cover the main use cases, a separate repository would provide limited value while introducing additional compatibility, governance, release, and maintenance costs.

JesseStutler: Now this is better

Issue #5146, opened 2026-03-31
Agent-Scheduler：concurrent map read and map write in handleSkipPredicatePlugin function


JBinin:
### Description

The test cluster has  about 1500  nodes. I created 1,000 Pods and then started the agent-scheduler. After scheduling some of the Pods, a "concurrent map read and map write" error occurred, causing the agent-scheduler to crash.

```
fatal error: concurrent map read and map write

goroutine 57493 [running]:
internal/runtime/maps.fatal({0x103cb860f?, 0x10227b2cc?})
        /Users/chenjiabin/.g/versions/1.25.3/src/runtime/panic.go:1046 +0x20
k8s.io/apimachinery/pkg/util/sets.Set[...].Has(...)
        /Users/chenjiabin/Mi/dev/volcano/vendor/k8s.io/apimachinery/pkg/util/sets/set.go:78
volcano.sh/volcano/pkg/scheduler/plugins/predicates.handleSkipPredicatePlugin(...)
        /Users/chenjiabin/Mi/dev/volcano/pkg/scheduler/plugins/predicates/predicates.go:782
volcano.sh/volcano/pkg/scheduler/plugins/predicates.(*PredicatesPlugin).Predicate(0x1402aab2480, 0x141399c3260, 0x141865ecd00, 0x140ed3f8fa0)
        /Users/chenjiabin/Mi/dev/volcano/pkg/scheduler/plugins/predicates/predicates.go:642 +0x888
volcano.sh/volcano/pkg/agentscheduler/plugins/predicates.(*predicatesPlugin).OnPluginInit.func2(0x106ef93e8?, 0x1400f5788a0?)
        /Users/chenjiabin/Mi/dev/volcano/pkg/agentscheduler/plugins/predicates/predicates.go:73 +0x94
volcano.sh/volcano/pkg/agentscheduler/framework.(*Framework).PredicateFn(0x1401af49380, 0x141399c3260, 0x141865ecd00)
        /Users/chenjiabin/Mi/dev/volcano/pkg/agentscheduler/framework/plugins.go:94 +0xf8
volcano.sh/volcano/pkg/agentscheduler/actions/allocate.(*Action).predicateForAllocateAction(0x140fcd19c20?, 0x141399c3260, 0x141865ecd00)
        /Users/chenjiabin/Mi/dev/volcano/pkg/agentscheduler/actions/allocate/allocate.go:274 +0x2c
[...]

Issue #5147, opened 2026-03-31
Agent-scheduler：VolumeBinding plugin nil point problem


JBinin:
### Description

The test cluster has about 1500 nodes. I created 1,000 Pods and then started the agent-scheduler. After scheduling some of the Pods, a "invalid memory address or nil pointer dereference" error occurred, causing the agent-scheduler to crash.

```
E0330 20:26:48.152087   60608 panic.go:262] "Observed a panic" panic="runtime error: invalid memory address or nil pointer dereference" panicGoValue="\"invalid memory address or nil pointer dereference\"" stacktrace=<
        goroutine 6530 [running]:
        k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1069955a8, 0x1410c9325d0}, {0x106529000, 0x107f68b00})
                /Users/chenjiabin/Mi/dev/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:132 +0x94
        k8s.io/apimachinery/pkg/util/runtime.handleCrash({0x1069955e0, 0x1410c3c2d70}, {0x106529000, 0x107f68b00}, {0x0, 0x0, 0x141055af540?})
                /Users/chenjiabin/Mi/dev/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:107 +0xdc
        k8s.io/apimachinery/pkg/util/runtime.HandleCrashWithContext({0x1069955e0, 0x1410c3c2d70}, {0x0, 0x0, 0x0})
                /Users/chenjiabin/Mi/dev/volcano/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:78 +0x5c
        panic({0x106529000?, 0x107f68b00?})
                /Users/chenjiabin/.g/versions/1.25.3/src/runtime/panic.go:783 +0x120
        volcano.sh/volcano/pkg/scheduler/capabilities/volumebinding.(*volumeBinder).FindPodVolumes(0x14022d760b0, {{0x10699c918?, 0x14000069040?}, 0x13b71b6e0?}, 0x1403913cf08, 0x0, 0x1408391b188)
                /Users/chenjiabin/Mi/dev/volcano/pkg/scheduler/capabilities/volumebinding/binder.go:345 +0x288
[...]

PR #5154, opened 2026-04-02
Fix Agent Scheduler multi worker optimistic parallel scheduling concurrently conflict error

JesseStutler:
#### What type of PR is this?
/kind bug
/priority high

As https://github.com/volcano-sh/volcano/issues/5146 mentioned, when opening multiple workers optimistic parallel scheduling can lead to concurrent map read/write conflicts and cycleState overwriting issues. This is mainly because the Action is a singleton with only one framework variable, while different workers have their own independent frameworks, potentially causing concurrency conflicts when working simultaneously:  https://github.com/volcano-sh/volcano/issues/5147#issuecomment-4170361976

Besides, `lastProcessedNodeIndex` is simply an int var, which will also have conflict in multi-workers scenario, should also update to use atomic

Fixes https://github.com/volcano-sh/volcano/issues/5146 https://github.com/volcano-sh/volcano/issues/5147

Validation:
Test Yamls:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: bench
  namespace: default
spec:
  replicas: 1000
  selector:
    matchLabels:
      app: bench
  template:
    metadata:
      labels:
        app: bench
      annotations:
        foo: bar
    spec:
      containers:
        - name: nginx
          image: nginx:alpine
      volumes:
        - name: agent-runtime-volume
          emptyDir: {}
```

After create 1000 replicas, the scheduler didn't meet panic errors anymore

```release-note

```

### DIFF
--- a/pkg/agentscheduler/actions/allocate/allocate_test.go
+++ b/pkg/agentscheduler/actions/allocate/allocate_test.go
@@ -1,0 +1,178 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package allocate
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+
+	"volcano.sh/volcano/cmd/agent-scheduler/app/options"
+	scheduleroptions "volcano.sh/volcano/cmd/scheduler/app/options"
+	agentapi "volcano.sh/volcano/pkg/agentscheduler/api"
+	"volcano.sh/volcano/pkg/agentscheduler/framework"
+	agentuthelper "volcano.sh/volcano/pkg/agentscheduler/uthelper"
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/util"
+	commonutil "volcano.sh/volcano/pkg/util"
+)
+
+// TestConcurrentMultiWorkerScheduling verifies that multiple workers can concurrently
+// schedule different pods without data races or scheduling conflicts.
+func TestConcurrentMultiWorkerScheduling(t *testing.T) {
+	agentuthelper.InitTestEnv(t)
+	options.ServerOpts.ShardingMode = commonutil.NoneShardingMode
+	scheduleroptions.ServerOpts.ShardingMode = commonutil.NoneShardingMode
+
+	tests := []struct {
+		name          string
+		workerCount   int
+		podsPerWorker int
+		nodeCount     int
+	}{
+		{
+			name:          "concurrent scheduling with 4 workers",
+			workerCount:   4,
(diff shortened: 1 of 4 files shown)


pkg/scheduler/util/predicate_helper.go
```diff
@@ -131,7 +131,7 @@ func (ph *predicateHelper) PredicateNodes(task *api.TaskInfo, nodes []*api.NodeI
 	workqueue.ParallelizeUntil(ctx, 16, allNodes, checkNode)
 
 	//processedNodes := int(numFoundNodes) + len(filteredNodesStatuses) + len(failedPredicateMap)
-	lastProcessedNodeIndex = (lastProcessedNodeIndex + int(processedNodes)) % allNodes
+	lastProcessedNodeIndex.Store(int64((int(lastProcessedNodeIndex.Load()) + int(processedNodes)) % allNodes))
```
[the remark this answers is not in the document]
JesseStutler: I think it's ok to overwrite currently, different workers will have the same numsToFind value, so the scanning speed of nodes is the same. Even if their processing speeds differ, which might cause the fastest worker to overwrite, at least all nodes will be scanned once. However, CAS might cause some nodes to skip the scan.

pkg/agentscheduler/actions/allocate/allocate_test.go
```diff
@@ -0,0 +1,172 @@
...
+					fwk := testFwk.Frameworks[workerIdx]
+
+					for iter := 0; iter < tt.podsPerWorker; iter++ {
+						// 1. Pop from shared scheduling queue.
+						queue := fwk.Cache.SchedulingQueue()
+						podInfo, popErr := queue.Pop(klog.Background())
+						if popErr != nil {
+							errCh <- fmt.Errorf("worker %d iter %d: Pop failed: %v", workerIdx, iter, popErr)
+							return
+						}
```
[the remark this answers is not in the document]
JesseStutler: Already fixed

pkg/scheduler/util/predicate_helper.go
```diff
@@ -130,8 +132,9 @@ func (ph *predicateHelper) PredicateNodes(task *api.TaskInfo, nodes []*api.NodeI
 	//workqueue.ParallelizeUntil(context.TODO(), 16, len(nodes), checkNode)
 	workqueue.ParallelizeUntil(ctx, 16, allNodes, checkNode)
 
-	//processedNodes := int(numFoundNodes) + len(filteredNodesStatuses) + len(failedPredicateMap)
-	lastProcessedNodeIndex = (lastProcessedNodeIndex + int(processedNodes)) % allNodes
+	newIndex := int64((startIndex + int(processedNodes)) % allNodes)
+	lastProcessedNodeIndex.Store(newIndex)
```
qi-min, line 136: lastProcessedNodeIndex might be overwrote by different worker because worker predicate nodes concurrently and processedNodes number are different in each cycle in worker.
In this case, is it better to keep lastProcessedNodeIndex in worker scope and count index separately?

JesseStutler: Actually, I feel it's quite difficult. It requires a map to record the `lastProcessedNodeIndex` of each worker, which also needs a lock or a `sync.Map` to prevent concurrent read/write issues. However, I think this would impact the performance of multiple workers. Also, the current `PredicateHelper` doesn't contain worker information, so it needs to be passed through the `Allocate` action. But since the initialization of the `allocate` action precedes the worker initialization, carrying worker information would require adding an interface to the action: https://github.com/volcano-sh/volcano/blob/d3721cbfe367b35d49317ced1d64a78db89d9ed9/pkg/agentscheduler/scheduler.go#L105-L122. I think my current modification is at least the simplest and won't affect multi-worker operation, still achieving round-robin functionality.

qi-min: Make sense, then I agree to keep the lastProcessedNodeIndex unchanged.

PR #5158, opened 2026-04-06
scheduler: fix taskEligibleToPreempt blocking gang job tasks with stale NominatedNodeName

### LINKED ISSUE #5131: TaskEligibleToPreempt blocks gang job tasks that don't need preemption, causing entire gang to fail
## Summary

`taskEligibleToPreempt()` in the preempt action incorrectly blocks tasks that have a stale `NominatedNodeName` from a previous scheduling cycle, even when these tasks don't need any preemption and have sufficient resources available on multiple nodes. This causes gang jobs to fail repeatedly when some tasks carry a residual `NominatedNodeName`.

## Environment

- Volcano version: latest master (also reproducible on recent releases)
- Kubernetes version: 1.28+
- Related: #5044, #5045

### Root Cause

In `pkg/scheduler/actions/preempt/preempt.go`, the function `taskEligibleToPreempt()` (line 400-437) checks whether a task's `Pod.Status.NominatedNodeName` is schedulable:

```go
nomNodeName := preemptor.Pod.Status.NominatedNodeName
if len(nomNodeName) > 0 {
    nodeInfo, ok := pmpt.ssn.Nodes[nomNodeName]
    // ...
    err := pmpt.ssn.PredicateFn(preemptor, nodeInfo)
    if err == nil {
        return fmt.Errorf("not eligible due to the pod's nominated node is already schedulable, ...")
    }
    // ...
}
```

When the predicate passes (`err == nil`), the function returns an error, **completely skipping this task from the entire preempt action** — not only preventing it from preempting others, but also preventing it from being placed on any available node via the 0-victim pipeline path in `normalPreempt()`.

### Why This Is Wrong

The assumption behind this check is: "if the nominated node is schedulable, the allocate action should have handled this task, so preemption is not needed." However, this assumption **breaks for gang jobs**:

1. The allocate action may have tried to schedule all tasks in the gang, but failed to satisfy `minAvailable` across the whole gang → everything was rolled back
2. After rollback, `NominatedNodeName` remains set on `Pod.Status` (it is never cleared — as noted in #5044, `Statement.pipeline()` is a no-op and `NominatedNodeName` lifecycle is incomplete)
3. In the preempt action, `taskEligibleToPreempt` sees the stale `NominatedNodeName`, checks the nominated node, finds it schedulable → blocks the task
4. The task never reaches `normalPreempt()`, which would have placed it on an available node with 0 victims (no preemption needed)
5. Since some tasks in the gang are blocked, `minAvailable` is not met → the entire gang (including tasks that were successfully pipelined, even via preemption) is discarded

### The Irony

`normalPreempt()` already fully supports 0-victim pipeline — placing a task on a node with sufficient resources without evicting anyone. Other tasks in the same gang (that happen to not have `NominatedNodeName`) succeed through this exact path. But `taskEligibleToPreempt` prevents the affected tasks from ever reaching it.

### Job Configuration

- A gang job with `minAvailable = total pod count` (all tasks required)
- role1: 1 pod, requires 4 GPUs → needs preemption on a GPU node
- role2: 1 pod, requires 2 CPU → fits on any CPU node
- role3: 4 pods, each requires 30 CPU → fits on CPU nodes (plenty of capacity)

### Cluster State

- Multiple CPU nodes, each with 60+ CPU available — more than enough for all role3 pods
- GPU nodes with low-priority pods that can be preempted for role1

### Steps

1. Submit the gang job. First scheduling cycle: allocate action fails to satisfy the full gang (e.g., not enough GPU resources for role1 during allocate). Some tasks get `NominatedNodeName` set.
2. Preempt action runs:
   - role1 pod successfully preempts 4 low-priority pods on a GPU node ✅
   - role2 pod placed on a CPU node, 0 victims ✅
   - role3 pod-0 placed on a CPU node, 0 victims ✅
   - role3 pod-1 placed on a CPU node, 0 victims ✅
   - **role3 pod-2** has `NominatedNodeName` from a previous cycle → `taskEligibleToPreempt` checks nominated node → predicate passes → returns error → **task completely skipped** ❌
   - **role3 pod-3** same as above ❌
3. Gang has only 4/6 tasks pipelined → `minAvailable` not met → `stmt.Discard()` → **everything rolled back**, including role1's successful GPU preemption.
4. Next cycle: same thing happens → **infinite loop**.

### Scheduler Logs

```
# role3 pod-2 blocked:
preempt.go:201] Preemptor <ns/job-role3-2> failed to preempt Task, err: not eligible due to the pod's nominated node is already schedulable, which should not happen as preemption means no node is schedulable

# role3 pod-3 blocked:
preempt.go:201] Preemptor <ns/job-role3-3> failed to preempt Task, err: not eligible due to the pod's nominated node is already schedulable, which should not happen as preemption means no node is schedulable

# Gang not satisfied:
job_info.go:1101] Job ns/job Task role3 occupied 2 less than task min available
```

### Intermittent Nature

This bug is **intermittent** because task processing order in the preempt action depends on priority queue ordering (all tasks have equal priority → unstable sort). When the affected tasks happen to be processed before `NominatedNodeName` is set, they succeed. When processed after, they fail.

## Proposed Fix

Change `taskEligibleToPreempt()` to not block when the nominated node is schedulable — let the task continue through `normalPreempt()`, which will handle it correctly via 0-victim pipeline:

```go
err := pmpt.ssn.PredicateFn(preemptor, nodeInfo)
if err == nil {
    // Nominated node is schedulable — let the task go through normalPreempt
    // which supports 0-victim pipeline (placing without preemption)
    klog.V(4).Infof("Task <%s/%s> nominated node <%s> is already schedulable, continue to preempt flow",
        preemptor.Namespace, preemptor.Name, nomNodeName)
    return nil
}
```

This is safe because:
- `normalPreempt()` only evicts victims when resources are insufficient (line 356: `if ssn.Allocatable(...) && preemptor.InitResreq.LessEqual(node.FutureIdle(), ...) { break }`)
- When resources are sufficient, it does a 0-victim pipeline — exactly what the blocked tasks need
- This is the same path that other tasks in the same gang (role2, role3 pod-0, role3 pod-1) successfully use

## Impact

- **Gang jobs with mixed resource requirements** (e.g., GPU + CPU roles) are the most affected
- Jobs get stuck in an infinite schedule-rollback loop
- Successful preemptions (e.g., for the GPU role) are wasted — victims are evicted then the eviction is rolled back
- The bug is intermittent, making it hard to diagnose in production

1.
2.
3.

**Actual Result:**

When a gang job contains tasks with mixed resource requirements (e.g., GPU tasks + CPU tasks), and some CPU tasks carry a stale `NominatedNodeName` from a previous scheduling cycle, `taskEligibleToPreempt()` blocks those tasks from entering the preempt flow entirely — even though they don't need preemption and there are plenty of free resources on multiple nodes.

This causes the gang to fail `minAvailable`, triggering a full rollback of all pipelined tasks (including those that were successfully placed via preemption). The job gets stuck in an infinite schedule-rollback loop.

Scheduler log output:

```
preempt.go:201] Preemptor <ns/job-role3-2> failed to preempt Task, err: not eligible due to the pod's nominated node is already schedulable, which should not happen as preemption means no node is schedulable
preempt.go:201] Preemptor <ns/job-role3-3> failed to preempt Task, err: not eligible due to the pod's nominated node is already schedulable, which should not happen as preemption means no node is schedulable
job_info.go:1101] Job ns/job Task role3 occupied 2 less than task min available
```

**Expected Result:**

Tasks whose `NominatedNodeName` points to a schedulable node should NOT be blocked from the preempt action. They should continue through `normalPreempt()`, which already supports 0-victim pipeline (placing tasks on nodes with sufficient resources without evicting anyone).

This is the same path that other tasks in the same gang — those without `NominatedNodeName` — successfully use in the same scheduling cycle. All tasks in the gang should be pipelined successfully, and the gang should reach `minAvailable` and be committed.

v1.11.0

_No response_

SuperMarioYL:
## What type of PR is this?

/kind bug

When a gang job's allocate attempt fails and rolls back, the `NominatedNodeName` on the pod is not cleared. In the subsequent preempt action, `taskEligibleToPreempt` finds the nominated node is schedulable and incorrectly returns an error, blocking the task from entering the preempt flow entirely.

This prevents the task from reaching the 0-victim pipeline path in `normalPreempt()`, causing the gang to fail `minAvailable` and get discarded in an infinite schedule-rollback loop.

## How does this fix it?

When `PredicateFn` returns nil (node is schedulable), instead of returning an error that blocks the task, we now return nil with a V(4) log message. This allows the task to proceed through the normal preempt flow, which already correctly handles the 0-victim case at lines 383-404 of `preempt.go`.

Fixes #5131

```release-note
Fixed a bug where gang job tasks with stale NominatedNodeName were incorrectly blocked from the preempt action when the nominated node had sufficient resources, causing the gang to fail minAvailable in an infinite loop.
```

## Testing

- Added `TestPreemptWithNominatedNode` test that creates a pending task with `NominatedNodeName` set to a schedulable node and verifies it proceeds through the 0-victim pipeline path without being blocked.
- All existing preempt tests continue to pass.

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -440,7 +440,9 @@ func (pmpt *Action) taskEligibleToPreempt(preemptor *api.TaskInfo) error {
 
 		err := pmpt.ssn.PredicateFn(preemptor, nodeInfo)
 		if err == nil {
-			return fmt.Errorf("not eligible due to the pod's nominated node is already schedulable, which should not happen as preemption means no node is schedulable")
+			klog.V(4).Infof("Task <%s/%s> nominated node <%s> is already schedulable, continue to preempt flow",
+				preemptor.Namespace, preemptor.Name, nomNodeName)
+			return nil
 		}
 
 		fitError, ok := err.(*api.FitError)
--- a/pkg/scheduler/actions/preempt/preempt_test.go
+++ b/pkg/scheduler/actions/preempt/preempt_test.go
@@ -674,6 +674,102 @@ func TestTopologyAwarePreempt(t *testing.T) {
 	}
 }
 
+// TestPreemptWithNominatedNode verifies that a task with a stale NominatedNodeName
+// is not blocked from entering the preempt flow when its nominated node is schedulable.
+// This is a regression test for https://github.com/volcano-sh/volcano/issues/5131.
+func TestPreemptWithNominatedNode(t *testing.T) {
+	plugins := map[string]framework.PluginBuilder{
+		conformance.PluginName: conformance.New,
+		gang.PluginName:        gang.New,
+		priority.PluginName:    priority.New,
+		proportion.PluginName:  proportion.New,
+		predicates.PluginName:  predicates.New,
+	}
+	highPrio := util.BuildPriorityClass("high-priority", 100000)
+	lowPrio := util.BuildPriorityClass("low-priority", 10)
+
+	// Build a pending preemptor pod with a stale NominatedNodeName.
+	// The node has enough idle resources, so PredicateFn will return nil.
+	// Before the fix, taskEligibleToPreempt would incorrectly block this task.
+	preemptorPod := util.BuildPod("c1", "preemptor1", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg2", make(map[string]string), make(map[string]string))
+	preemptorPod.Status.NominatedNodeName = "n1"
+
+	tests := []uthelper.TestCommonStruct{
+		{
+			Name: "task with stale NominatedNodeName on schedulable node should not be blocked from preempt",
+			PodGroups: []*schedulingv1beta1.PodGroup{
+				util.BuildPodGroupWithPrio("pg1", "c1", "q1", 0, map[string]int32{}, schedulingv1beta1.PodGroupInqueue, "low-priority"),
+				util.BuildPodGroupWithPrio("pg2", "c1", "q1", 1, map[string]int32{"": 1}, schedulingv1beta1.PodGroupInqueue, "high-priority"),
+			},
+			Pods: []*v1.Pod{
+				util.BuildPod("c1", "preemptee1", "n1", v1.PodRunning, api.BuildResourceList("1", "1G"), "pg1", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				preemptorPod,
+			},
+			Nodes: []*v1.Node{
+				util.BuildNode("n1", api.BuildResourceList("10", "10G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+			},
+			Queues: []*schedulingv1beta1.Queue{
+				util.BuildQueue("q1", 1, nil),
+			},
+			// The node has plenty of idle resources, so the preemptor should be pipelined
+			// via the 0-victim path without evicting anyone.
+			ExpectEvictNum: 0,
+		},
+	}
+
+	trueValue := true
+	tiers := []conf.Tier{
+		{
+			Plugins: []conf.PluginOption{
+				{
+					Name:               conformance.PluginName,
+					EnabledPreemptable: &trueValue,
+				},
+				{
+					Name:                gang.PluginName,
+					EnabledPreemptable:  &trueValue,
+					EnabledJobPipelined: &trueValue,
+					EnabledJobStarving:  &trueValue,
+				},
+				{
+					Name:                priority.PluginName,
+					EnabledTaskOrder:    &trueValue,
+					EnabledJobOrder:     &trueValue,
+					EnabledPreemptable:  &trueValue,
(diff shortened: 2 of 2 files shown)


SuperMarioYL:
Gentle ping on this — happy to rebase onto latest master or address any concerns. Let me know if there's anything that would help move this forward.

SuperMarioYL:
Following up on this one — it's been open since early April and is still waiting on a first human review.

Current state: `kind/bug`, DCO green, contributor-status workflow approved, no conflicts (just behind master). Per the APPROVALNOTIFIER it needs a reviewer's `lgtm` before it can move to the approval step.

Recap of the fix: in the preempt action, a task carrying a stale `NominatedNodeName` from an earlier scheduling cycle was blocked from entering the preempt flow, which could stall gang-scheduled jobs indefinitely. The change lets a task whose nominated node is still schedulable continue through preemption.

Happy to rebase onto latest master, split it smaller, or adjust the approach if the shape isn't what you'd want.

SuperMarioYL:
Gentle ping — happy to address any feedback or rebase if needed.

Issue #4165, opened 2025-03-31
Fair Sharing not achieved in Volcano Scheduler


MSSkowron:
## Introduction

We are currently experimenting with the Volcano scheduler’s fair-sharing mechanism to ensure that when multiple tenants are configured with identical queues and namespaces, each receives an equal share of cluster resources (approximately 33% per tenant). Our test scenario is designed to thoroughly examine the behavior of the fair-share mechanism when three tenants are concurrently submitting jobs to their dedicated queues.

### Cluster Setup

- Kubernetes Cluster: Running Volcano Scheduler v1.11.0
- Nodes: 5 nodes, each provisioned with:
  - CPU: ~15,000m (total ~75 CPU units available)
  - Memory: ~15,000Mi

### Volcano Scheduler Configuration

- Enabled Plugins:
  - Tier 1: priority, gang
  - Tier 2: drf (Dominant Resource Fairness with enablePreemptable set to true), predicates, proportion, nodeorder, binpack

### Configuration Snippet:
```
$ kubectl get configmap volcano-scheduler-configmap -n volcano-system -o yaml
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "enqueue, allocate, reclaim"
    tiers:
    - plugins:
      - name: priority
      - name: gang
    - plugins:
      - name: drf
        enablePreemptable: true
      - name: predicates
      - name: proportion
      - name: nodeorder
      - name: binpack
kind: ConfigMap
metadata:
  creationTimestamp: "2025-03-31T15:50:45Z"
  name: volcano-scheduler-configmap
  namespace: volcano-system
  resourceVersion: "1942"
  uid: e9ee837c-78d8-4cd6-870e-a39eebb3e4fb
```
```
$ kubectl describe deployment volcano-scheduler -n volcano-system
Name:                   volcano-scheduler
Namespace:              volcano-system
CreationTimestamp:      Mon, 31 Mar 2025 17:50:45 +0200
Labels:                 app=volcano-scheduler
                        app.kubernetes.io/managed-by=Helm
Annotations:            deployment.kubernetes.io/revision: 9
                        meta.helm.sh/release-name: volcano
                        meta.helm.sh/release-namespace: volcano-system
Selector:               app=volcano-scheduler
Replicas:               1 desired | 1 updated | 1 total | 1 available | 0 unavailable
StrategyType:           RollingUpdate
MinReadySeconds:        0
[...]

PR #5161, opened 2026-04-06
[Proposal] Per-user fair share scheduling plugin with decayed usage tracking

### LINKED ISSUE #5160: [Proposal] Per-user fair share scheduling plugin with decayed usage tracking
## Motivation

Volcano's existing DRF plugin provides dominant resource fairness at the namespace/queue level, but it only considers the **current allocation snapshot** — it has no memory of past usage. This leads to a well-known problem in multi-user GPU clusters:

- User A submits 300 GPU jobs and fills the cluster
- User B arrives later with 10 jobs
- As User A's jobs complete, User A's *new* pending jobs get scheduled again immediately (they have equal or lower current allocation), starving User B

This was reported in #4165 and is a common pain point in shared AI/ML clusters. SLURM solves this with its "fair share" algorithm that tracks historical usage and applies exponential decay.

## Proposal

Add a new `fairshare` scheduler plugin that:

1. **Tracks cumulative resource-seconds per user** (identified by namespace) across scheduling cycles using package-level state
2. **Applies exponential half-life decay** to historical usage (configurable, default 4 hours), so past consumption is gradually forgiven
3. **Orders jobs by cumulative usage** — users who have consumed less historically get scheduled first
4. **Complements existing plugins** — runs alongside `priority` (which takes precedence) and `gang`

### Algorithm

Each scheduling cycle (~1s):

1. Decay all historical usage: `usage × 2^(-elapsed/halfLife)`
2. For each running task, accumulate: `resource_count × elapsed_seconds`
3. In `JobOrderFn`, compare users by cumulative usage (lower wins), with a tiebreaker on currently-running resources

### Configuration

```yaml
- name: fairshare
  arguments:
    fairshare.targetQueues: "gpu-queue"
    fairshare.resourceKey: "nvidia.com/gpu"
    fairshare.halfLifeMinutes: "240"
    fairshare.enableEnqueueGate: "false"
```

### Key design decisions

- **Namespace-based identity**: Uses the job's namespace as user identity (namespace-per-user is common in multi-tenant GPU clusters)
- **Package-level globals for state**: Volcano recreates plugin instances via `New()` every cycle, so instance-level state doesn't persist. Global state with `sync.Mutex` protection survives across cycles.
- **Half-life decay over sliding window**: Exponential decay is smoother than a fixed window and matches SLURM's proven approach
- **Per-queue tracking**: Different queues can track different resources (e.g., `nvidia.com/gpu` for GPU queues, `amd.com/gpu` for AMD queues, `cpu` for CPU queues)

### Interaction with other plugins

- **`priority` plugin**: Runs before `fairshare` in the same tier. Higher PriorityClass always wins. Fair share only breaks ties at equal priority.
- **`drf` plugin**: Complementary — DRF handles multi-resource dominance, fairshare handles temporal fairness. They can coexist or be used independently.

## Testing done

We have validated this plugin on a test cluster with:
- 25 unit tests covering the algorithm, decay math, and ordering logic
- 7 integration test scenarios (uniform jobs, randomized durations, burst asymmetry, DAG/workflow simulation, PriorityClass interaction, scheduler restart recovery)
- Results show significant improvement: last-to-submit user wait time reduced from ~7 minutes to ~2 minutes in a 4-user, 2-GPU test environment

## Deliverables

- Design doc: `docs/design/fairshare.md`
- Plugin code: `pkg/scheduler/plugins/fairshare/`
- Unit tests
- E2E tests

Will submit a PR with the implementation. Happy to discuss the design and iterate based on community feedback.

/kind feature

dmildh-absci:
## What type of PR is this?

/kind feature

Adds a new `fairshare` scheduler plugin that provides per-namespace temporal fairness using decayed cumulative usage tracking. This addresses the namespace-level fair share gap identified in #4165 and proposed in #5160.

### Problem

The existing DRF plugin provides snapshot-based dominant resource fairness but has **no memory of past usage**. In multi-tenant GPU clusters this causes:

- Namespace A submits hundreds of GPU jobs and fills the cluster
- Namespace B arrives later with a few jobs
- As namespace A's jobs complete, its new pending jobs immediately regain priority, starving namespace B

### Solution

A SLURM-inspired fair share algorithm that:

1. **Tracks cumulative resource-seconds per namespace** across scheduling cycles
2. **Applies exponential half-life decay** (configurable, default 4h) so past usage is gradually forgiven
3. **Orders jobs by cumulative usage** — lower historical usage = higher priority
4. **Complements existing plugins** — runs alongside `priority` (which takes precedence) and `gang`, and needs to be listed before `drf` (see the user guide for why)
5. **Defaults to all queues** if `fairshare.targetQueues` is left unset

### Configuration

```yaml
- name: fairshare
  arguments:
    fairshare.targetQueues: "gpu-queue"
    fairshare.resourceKey: "nvidia.com/gpu"
    fairshare.halfLifeMinutes: "240"
```

See [docs/user-guide/how_to_use_fairshare_plugin.md](https://github.com/dmildh-absci/volcano/blob/feat/fairshare-plugin/docs/user-guide/how_to_use_fairshare_plugin.md) for the full arguments list and plugin-interaction guidance, and [docs/design/fairshare.md](https://github.com/dmildh-absci/volcano/blob/feat/fairshare-plugin/docs/design/fairshare.md) for design rationale.

Fixes #5160
Relates to #4165

```release-note
Add fairshare scheduler plugin for per-namespace temporal fair share scheduling with configurable half-life decay.
```

### Unit tests (49 tests)

- Max-min fair share algorithm correctness
- Decay factor math (one/two half-lives, zero elapsed, zero half-life, small elapsed)
- decayAllUsage (halves after one half-life, cleans up negligible entries, multi-queue)
- Usage ordering (lower usage wins, equal usage falls to running tiebreaker, realistic multi-namespace scenario)
- Helper functions (namespace extraction, resource key defaults/overrides)
- targetQueues allowlist behavior (defaults to all queues when unset)
- JobOrderFn abstain behavior (only compares jobs within the same targeted queue)
- Persistence: ConfigMap create/update/load/round-trip, disabled no-op, corrupt data, rate-limited flush (including retry-promptly-on-failure)

All tests pass with `-race`.

### Integration testing

Validated on a test cluster with 2 GPU nodes, 4 tenant namespaces:

| Scenario | Result |
|----------|--------|
| Without fairshare (baseline) | FIFO behavior, last tenant waited ~7 min |
| With fairshare + decay | All tenants got GPUs within ~2 min |
| Burst asymmetry (1 tenant = 8 jobs, others = 1) | Minority tenants' jobs completed within ~2 min |
| DAG/workflow simulation | GPU steps interleaved across tenants |
| PriorityClass interaction | High-priority bypassed fair share as expected |
| Scheduler restart | Running jobs survived, new jobs scheduled fairly |

> **Note**: This PR is a draft. We will update with additional production cluster data as we roll this out.

### E2E test

`test/e2e/schedulingaction/fairshare.go` — CPU-based fairshare ordering test on a single capacity-constrained node.

## Files changed

| File | Description |
|------|-------------|
| docs/design/fairshare.md | Design document (2019 namespace-weight proposal + 2026 decayed-usage-tracking proposal) |
| docs/user-guide/how_to_use_fairshare_plugin.md | Usage guide: arguments, config examples, DRF/priority/gang interaction |
| pkg/scheduler/plugins/fairshare/fairshare.go | Plugin implementation |
| pkg/scheduler/plugins/fairshare/persist.go | ConfigMap-based state persistence, flushed from OnSessionClose |
| pkg/scheduler/plugins/factory.go | Plugin registration |
| test/e2e/schedulingaction/fairshare.go | E2E test |

/cc @william-wang @Thor-wl @lowang-bh

### DIFF
--- a/docs/design/fairshare.md
+++ b/docs/design/fairshare.md
@@ -54,7 +54,7 @@ inactivity. After 24 hours, the penalty is effectively forgotten.
 
 Volcano recreates plugin instances via `New()` every scheduling cycle, so instance-level
 state is lost between cycles. The fairshare plugin uses package-level globals protected
-by a `sync.Mutex` to persist usage history:
+by a `sync.Mutex` to persist usage history across scheduling cycles:
 
 ```go
 var (
@@ -64,9 +64,52 @@ var (
 )
 ```
 
-State survives across cycles within a scheduler process lifetime. On scheduler pod restart,
-all users reset to zero — this is acceptable because restarts are rare and act as a
-"full forgiveness" event.
+#### Durable persistence (ConfigMap)
+
+To survive scheduler restarts, the plugin can optionally persist state to a ConfigMap.
+When `fairshare.persistState` is set to `"true"`:
+
+1. On the **first scheduling cycle**, a `sync.Once` block loads any existing state from
+   the ConfigMap into `globalUsage` / `globalLastCycle`, then starts a background goroutine.
+2. The **background goroutine** periodically flushes the current state to the ConfigMap
+   (default: every 30 seconds). Writes use the ConfigMap's `resourceVersion` for
+   optimistic concurrency.
+3. On restart, the loaded `globalLastCycle` is used to compute the elapsed time and apply
+   the correct decay, so users are not unfairly penalized or forgiven by the downtime.
+
+The ConfigMap is stored in the scheduler's namespace (default: `volcano-system`):
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: fairshare-usage-state
+  namespace: volcano-system
+  labels:
+    app: volcano-scheduler
+    component: fairshare
+data:
+  state.json: |
+    {
+      "lastCycle": "2026-04-07T12:00:00Z",
+      "queues": {
+        "gpu-queue": {
+          "alice": 12345.67,
+          "bob": 8901.23
+        }
+      }
+    }
+```
+
+**Design considerations:**
+
+- **Leader election**: Volcano already elects a single active scheduler. Only the leader writes.
+- **Data loss window**: At most `flushInterval` seconds of usage data is lost on a crash.
+  With the default 30-second interval and a 4-hour half-life, this is negligible.
+- **Size**: Even with 1000 users across 50 queues, the JSON payload is ~50 KB — well within
+  the 1 MB ConfigMap limit.
+- **Backward compatibility**: Persistence is disabled by default. Existing deployments are
+  unaffected.
 
 ## Configuration
 
@@ -83,6 +126,7 @@ tiers:
       fairshare.resourceKey: "nvidia.com/gpu"
       fairshare.halfLifeMinutes: "240"
       fairshare.enableEnqueueGate: "false"
+      fairshare.persistState: "true"
 ```
 
 ### Arguments
@@ -94,6 +138,10 @@ tiers:
 | `fairshare.resourceKey.<queue>` | _(none)_ | Per-queue resource override (e.g., `amd.com/gpu`, `cpu`) |
 | `fairshare.halfLifeMinutes` | `240` | Half-life for usage decay in minutes |
 | `fairshare.enableEnqueueGate` | `false` | When `true`, blocks users at/above their calculated share from entering the scheduling pipeline |
+| `fairshare.persistState` | `false` | When `true`, persists usage state to a ConfigMap so it survives scheduler restarts |
+| `fairshare.stateNamespace` | `volcano-system` | Namespace for the state ConfigMap |
+| `fairshare.stateConfigMap` | `fairshare-usage-state` | Name of the state ConfigMap |
+| `fairshare.flushIntervalSeconds` | `30` | How often to flush state to the ConfigMap (in seconds) |
 
 ## Interaction with existing plugins
 
@@ -137,7 +185,7 @@ handles ordering among eligible jobs.
 
 ## Testing
 
-### Unit tests (25 tests)
+### Unit tests (33 tests)
 
 - Max-min fair share algorithm correctness (single user, equal demand, asymmetric demand, progressive elimination)
 - Decay factor math (one/two half-lives, zero elapsed, zero half-life, small elapsed)
@@ -145,6 +193,9 @@ handles ordering among eligible jobs.
 - Usage ordering (lower usage wins, equal usage falls to running tiebreaker, realistic multi-user scenario)
 - Decay scenario (10-hour job decay over 4h and 24h)
 - Helpers (namespace extraction, resource key defaults/overrides)
+- Persistence: flush creates ConfigMap, flush updates existing, load populates globals,
+  load handles missing ConfigMap, load handles empty data, flush→load round-trip,
+  disabled persistence is no-op, corrupt JSON returns error
 
 ### Integration tests
 
@@ -159,6 +210,6 @@ Validated on a test cluster with 2 GPU nodes and 4 user namespaces:
 
 ## Limitations
 
-- State is in-memory only; lost on scheduler restart (all users reset to zero)
+- Without `persistState`, state is in-memory only and lost on scheduler restart
 - Namespace-based identity only; no support for arbitrary user labels (can be extended)
-- No persistence to external storage (could be added via ConfigMap or CRD in the future)
+- ConfigMap persistence has a small data-loss window equal to the flush interval on crashes
--- a/docs/design/fairshare.md
+++ b/docs/design/fairshare.md
@@ -1,179 +1,164 @@
-# Namespace fair share
+# Per-User Fair Share Scheduling with Decayed Usage Tracking
 
-[@lminzhw](http://github.com/lminzhw); May 8, 2019
+## Background
 
-## Motivation
+Volcano's existing DRF plugin provides dominant resource fairness at the namespace/queue level,
+but it only considers the **current allocation snapshot** — it has no memory of past usage.
+This leads to a well-documented problem in multi-user GPU clusters (see [#4165](https://github.com/volcano-sh/volcano/issues/4165)):
 
-`Queue` was introduced in [kube-batch](http://github.com/kubernetes-sigs/kube-batch) to share resources among users.
+- User A submits hundreds of GPU jobs and fills the cluster
+- User B arrives later with a handful of jobs
+- As User A's jobs complete, User A's new pending jobs immediately regain priority
+  (they have equal or lower current allocation), effectively starving User B
 
-But, the user in the same `Queue` are equivalent during scheduling. For example, we have a `Queue` contains a small amount of resources, and there are 10 pods belong to UserA and 1000 pods belong to UserB. In this case, pods of UserA would have less probability to bind with node.
+This "submission-order bias" means the first user to flood the queue monopolizes resources
+indefinitely, even when other users have legitimate demand. SLURM solves this with its
+fair share algorithm that tracks historical usage with exponential decay.
 
-So, we need a more fine-grained strategy to balance resource usage among users in the same `Queue`.
+## Proposal
 
-In consideration of multi-user model in kubernetes, we use namespace to distinguish different user. Each namespace would have its weight to control resources usage.
+Add a new `fairshare` scheduler plugin that tracks cumulative resource-seconds per user
+(identified by namespace) and applies exponential half-life decay so that past consumption
+is gradually forgiven.
 
-## Function Specification
+### User identity
 
-Weight have these features:
-> 1. `Queue` level
-> 2. an `integer` with default value 1
-> 3. record in namespace `quota`
-> 4. higher value means more resources after balancing
+Users are identified by their job's **namespace**. This follows the namespace-per-user
+pattern common in multi-tenant Kubernetes GPU clusters. No additional labels are required.
 
-### where is the weight
+### Algorithm
+
+Each scheduling cycle (~1 second):
+
+1. **Decay** all historical usage: `usage × 2^(-elapsed / halfLife)`
+2. **Accumulate** running usage: for each allocated task, add `resource_count × elapsed_seconds`
+   to the user's cumulative total
+3. **Order** pending jobs via `JobOrderFn`:
+   - **Primary**: Lower cumulative usage wins (with a 1.0 resource-second epsilon for float comparison)
+   - **Secondary**: Fewer currently-running resources wins (within-cycle tiebreaker)
+
+### Half-life decay
+
+| Time since usage | Remaining weight (4h half-life) |
+|-----------------|--------------------------------|
+| 0 hours         | 100%                           |
+| 4 hours         | 50%                            |
+| 8 hours         | 25%                            |
+| 12 hours        | 12.5%                          |
+| 24 hours        | 1.6%                           |
+
+A user who consumed 10 GPU-hours will see their usage penalty halve every 4 hours of
+inactivity. After 24 hours, the penalty is effectively forgotten.
+
+### State persistence
+
+Volcano recreates plugin instances via `New()` every scheduling cycle, so instance-level
+state is lost between cycles. The fairshare plugin uses package-level globals protected
+by a `sync.Mutex` to persist usage history:
+
+```go
+var (
+    globalMu        sync.Mutex
+    globalUsage     = make(map[string]map[string]float64) // [queue][user] → resource-seconds
+    globalLastCycle time.Time
+)
+```
+
+State survives across cycles within a scheduler process lifetime. On scheduler pod restart,
+all users reset to zero — this is acceptable because restarts are rare and act as a
+"full forgiveness" event.
+
+## Configuration
 
 ```yaml
-apiVersion: v1
-kind: ResourceQuota
-metadata:
-  namespace: default
-spec:
-  hard:
-    limits.memory: 2Gi
-    volcano.sh/namespace.weight: 1  <- this field represent the weight of this namespace
+actions: "enqueue, allocate, backfill"
+tiers:
+- plugins:
+  - name: priority
+  - name: gang
+  - name: conformance
+  - name: fairshare
+    arguments:
+      fairshare.targetQueues: "gpu-queue"
+      fairshare.resourceKey: "nvidia.com/gpu"
+      fairshare.halfLifeMinutes: "240"
+      fairshare.enableEnqueueGate: "false"
 ```
 
-If many `ResourceQuota` in the same namespace have weight, the weight for this namespace is the highest one of them.
+### Arguments
 
-This weight should be positive, any invalid value is treated as default value 1.
+| Argument | Default | Description |
+|----------|---------|-------------|
+| `fairshare.targetQueues` | _(required)_ | Comma-separated queue names to apply fair share to |
+| `fairshare.resourceKey` | `nvidia.com/gpu` | Default resource to track |
+| `fairshare.resourceKey.<queue>` | _(none)_ | Per-queue resource override (e.g., `amd.com/gpu`, `cpu`) |
+| `fairshare.halfLifeMinutes` | `240` | Half-life for usage decay in minutes |
+| `fairshare.enableEnqueueGate` | `false` | When `true`, blocks users at/above their calculated share from entering the scheduling pipeline |
 
-### Scheduler Framework
+## Interaction with existing plugins
 
-Introduce two new fields in SchedulerCache
+### priority plugin
 
-```go
-type NamespaceInfo struct {
-    Weight int
-}
+The `priority` plugin runs before `fairshare` in the same tier. A higher PriorityClass
+always wins. Fair share only breaks ties at equal priority levels.
 
-type SchedulerCache struct {
-    ...
-    quotaInformer    infov1.ResourceQuotaInformer
-    ...
-    NamespaceInfo  map[string]*kbapi.NamespaceInfo
-    ...
-}
-```
-
-The Scheduler will watch the lifecycle of `ResourceQuota` by `quotaInformer`, and refresh the info in `NamespaceInfo`.
-
-In `openSession` function, we should pass the `NamespaceInfo` through function `cache.Snapshot` into `Session` by using a new filed in `Session`/`ClusterInfo` struct.
-
-```go
-type Session struct {
-    ...
-    NamespaceInfo  map[string]*kbapi.NamespaceInfo
-    ...
-}
-type ClusterInfo struct {
-    ...
-    NamespaceInfo  map[string]*kbapi.NamespaceInfo
-    ...
-}
-```
-
-### Allocate Action
-
-#### Scheduling loop
-
-The behavior of `allocate` action is scheduling job in `Queue` one by one.
-
-At the beginning of scheduling loop, it will take a job with highest priority from `Queue`. And try to schedule tasks that belong to it until job is ready (matches the minMember) then go to next round.
-
-The priority of job mentioned above is defined by `JobOrder` functions registered by plugins. Such as job ready order of Gang plugin, priority order of Priority plugin, and also the share order of DRF plugin.
-
-#### JobOrder
-
-Namespace weight `should not` implement with JobOrder func. Because the scheduling of job would affect priority of the others.
-
-> e.g.
->
-> ns1 has job1, job2, ns2 has job3, job4. The original order is job1-job2-job3-job4.
->
-> After the scheduling of job1, right order should be job3-job4-job2. But in priority queue, we have no chance to fix the priority for job2
-
-#### Namespace Order
-
-To add namespace weight, we introduce a new order function named `NamespaceOrder` in `Session`.
-
-```go
-type Session struct {
-    ...
-    NamespaceOrderFns map[string]api.CompareFn
-    ...
-}
-```
-
-The scheduling loop in allocate should change as follows.
-
-In scheduling loop, firstly, choose a namespace having highest priority by calling `NamespaceOrderFn`, and then choose a job having highest priority using `JobOrderFn` in this namespace.
-
-After scheduling of job, push the namespace and job back to priority queue in order to refresh its priority. Because once a job is scheduled, assigned resource may decrease the priority of this namespace, the other jobs in the same namespace may be scheduled later.
-
-Always assigning resources to namespace with highest priority (lower resource usage) in every turn will make the resource balanced.
+When all competing jobs use the same PriorityClass, fair share is fully effective as the
+tiebreaker. This means using a high PriorityClass only helps when others don't — if everyone
+uses it, fair share still distributes resources equitably.
 
 ### DRF plugin
 
-DRF plugin use preemption and order of job to balance resource among jobs. The `share` in this plugin is defined as resource usage, the higher `share` means this job occupies the more resource now.
+The fairshare plugin is complementary to DRF. DRF handles multi-resource dominance
+(which resource is the bottleneck), while fairshare handles temporal fairness (who has
+consumed more historically). They can coexist or be used independently depending on
+cluster needs.
 
-#### Namespace Compare
+### gang plugin
 
-To introduce namespace weight into this plugin, we should define how to compare namespace having weight firstly.
+No interaction — gang scheduling handles minimum member requirements, while fairshare
+handles ordering among eligible jobs.
 
-For namespace n1 having weight w1 and namespace n2 having weight w2, we can compute the `share` of resource and recorded as u1 and u2. Now, the resource usage of n1 less than n2 can be defined as (u1 / w1 < u2 / w2)
+## Scheduler hooks
 
-`e.g.` ns1 having weight w1=2 use 6cpu, ns2 having weight w2=1 use 2cpu. In the scope of cpu, the ns1 use less resource than ns2. (6 / 3 < 3 / 1)
+| Hook | Purpose |
+|------|---------|
+| `JobOrderFn` | Orders jobs by cumulative usage (lower wins), with running-resource tiebreaker |
+| `JobEnqueueableFn` | (Optional) Blocks users at/above their max-min fair share from the scheduling pipeline |
+| `EventHandler` | Tracks allocations/deallocations in real-time during the scheduling cycle |
 
-#### Namespace Order
+## Logging
 
-Register `NamespaceOrder` function using the strategy mentioned above.
+| klog level | What it shows |
+|------------|---------------|
+| V(2) | Plugin config on creation, per-cycle queue summary (user count, running, demand) |
+| V(3) | Decay factor per cycle, fair shares, usage maps, ordering winner decisions |
+| V(4) | Session open/close, individual allocate/deallocate events |
+| V(5) | Every job comparison, every enqueue evaluation |
 
-#### preemption
+## Testing
 
-> The `preempt` action is disabled now. Do this later.
+### Unit tests (25 tests)
 
-In the `preemption` function now, strategy is just simply comparing the resource share among jobs .
+- Max-min fair share algorithm correctness (single user, equal demand, asymmetric demand, progressive elimination)
+- Decay factor math (one/two half-lives, zero elapsed, zero half-life, small elapsed)
+- `decayAllUsage` (halves after one half-life, cleans up negligible entries, multi-queue)
+- Usage ordering (lower usage wins, equal usage falls to running tiebreaker, realistic multi-user scenario)
+- Decay scenario (10-hour job decay over 4h and 24h)
+- Helpers (namespace extraction, resource key defaults/overrides)
 
-After adding namespace weight, we should check namespace of preemptor and preemptee firstly. The job in namespace with less resources can preempt others, or when namespace resource usage are the same, compare share of job instead.
+### Integration tests
 
-### Feature Interaction
+Validated on a test cluster with 2 GPU nodes and 4 user namespaces:
 
-#### preempt action
+1. **Without decay tracking**: FIFO behavior, last user waited ~7 minutes
+2. **With decay tracking**: All users got GPUs within ~2 minutes, scheduling rotated between users
+3. **Burst asymmetry** (1 user = 8 jobs, others = 1 each): Minority users' jobs completed within ~2 minutes
+4. **DAG/workflow simulation**: GPU steps interleaved across users by cumulative usage
+5. **PriorityClass interaction**: High-priority bypassed fair share as expected
+6. **Scheduler restart**: Running jobs survived, new jobs scheduled fairly post-restart
 
-Preempt is a strategy set to choose victims and finally evict it.
+## Limitations
 
-The way to choose victims is a function set named `Preemptable` registered by plugins. Such as job ready protection of Gang plugin, special pod protection of Conformance plugin, job share balance strategy of DRF plugin.
-
-All these plugin would choose some victims respective, and the intersection of them would be the final victim set. So, the choice made by DRF plugin would never break the requirement of others.
-
-### short hand
-
-1. Preempt may cause killing of some running pod.
-
-### Cases:
-
-- cluster have __16 cpu__, queue and namespace have default weight.
-
-    | queue | namespace | requested | queue assigned | namespace assigned |
-    | ----- | --------- | --------- | -------------- | ------------------ |
-    | q1    | ns1       | 5 cpu     | 8 cpu          | 4 cpu              |
-    |       | ns2       | 10 cpu    |                | 4 cpu              |
-    | q2    | ns3       | 10 cpu    | 8 cpu          | 6 cpu              |
-    |       | ns4       | 2 cpu     |                | 2 cpu              |
-
-- cluster have __16 cpu__, q1 with weight 1, q2 with weight 3. ns1 with weight 3, ns2 have weight 1, ns3 have weight 2, ns4 have weight 6.
-
-    | queue | namespace | requested | queue assigned | namespace assigned |
-    | ----- | --------- | --------- | -------------- | ------------------ |
-    | q1 w1 | ns1 w3    | 5 cpu     | 4 cpu          | 3 cpu              |
-    |       | ns2 w1    | 10 cpu    |                | 1 cpu              |
-    | q2 w3 | ns3 w2    | 10 cpu    | 12 cpu         | 10 cpu             |
-    |       | ns4 w6    | 2 cpu     |                | 2 cpu              |
-
-- cluster have __16 cpu__, q1 with weight 1, q2 with weight 3. ns1 have weight 2, ns2 have weight 6.
-
-    | queue | namespace | requested | queue assigned | namespace assigned |
-    | ----- | --------- | --------- | -------------- | ------------------ |
-    | q1 w1 | ns1 w2    |           | 4 cpu          |                    |
-    | q2 w3 | ns1 w2    | 5 cpu     | 12 cpu         | 3 cpu              |
-    |       | ns2 w6    | 20 cpu    |                | 9 cpu              |
+- State is in-memory only; lost on scheduler restart (all users reset to zero)
+- Namespace-based identity only; no support for arbitrary user labels (can be extended)
+- No persistence to external storage (could be added via ConfigMap or CRD in the future)
--- a/docs/design/fairshare.md
+++ b/docs/design/fairshare.md
@@ -375,3 +375,7 @@ Validated on a test cluster with 2 GPU nodes and 4 tenant namespaces:
 - ConfigMap persistence has a small data-loss window equal to the flush interval on crashes
 - Fairness is scoped within a queue (as in the 2019 proposal above); this plugin does not
   change how DRF/proportion balance resources across queues
+- `fairshare` registers no `PreemptableFn`/`ReclaimableFn`; it only reorders the pending
+  queue and (optionally) gates enqueue. It cannot preempt already-running jobs, so a
+  namespace with long-lived running jobs keeps its resources regardless of how stale its
+  usage penalty becomes
--- a/docs/design/fairshare.md
+++ b/docs/design/fairshare.md
@@ -1,8 +1,26 @@
-# Namespace fair share
+# Fair share scheduling
+
+This document covers two related design proposals for fair share scheduling among
+tenants sharing a `Queue`:
+
+1. [Namespace fair share](#namespace-fair-share-2019) (2019, [@lminzhw](http://github.com/lminzhw)) —
+   static per-namespace weights recorded on `ResourceQuota`, compared via a new
+   `NamespaceOrderFn` scheduling-loop stage.
+2. [Per-namespace fair share with decayed usage tracking](#per-namespace-fair-share-with-decayed-usage-tracking-2026)
+   (2026) — a self-contained scheduler plugin that tracks historical resource-seconds per
+   namespace with exponential decay, addressing the same underlying problem without
+   requiring new `Session`/`SchedulerCache` fields or a `NamespaceOrderFn` stage.
+
+For usage instructions (arguments, configuration examples, interaction with other plugins),
+see the [fairshare user guide](../user-guide/how_to_use_fairshare_plugin.md).
+
+---
+
+## Namespace fair share (2019)
 
 [@lminzhw](http://github.com/lminzhw); May 8, 2019
 
-## Motivation
+### Motivation
 
 `Queue` was introduced in [kube-batch](http://github.com/kubernetes-sigs/kube-batch) to share resources among users.
 
@@ -12,7 +30,7 @@ So, we need a more fine-grained strategy to balance resource usage among users i
 
 In consideration of multi-user model in kubernetes, we use namespace to distinguish different user. Each namespace would have its weight to control resources usage.
 
-## Function Specification
+### Function Specification
 
 Weight have these features:
 > 1. `Queue` level
@@ -20,7 +38,7 @@ Weight have these features:
 > 3. record in namespace `quota`
 > 4. higher value means more resources after balancing
 
-### where is the weight
+#### where is the weight
 
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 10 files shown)


pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,529 @@
...
+)
+
+// Package-level persistent state. Volcano calls New() every scheduling cycle,
+// so state stored on the plugin instance is lost between cycles. These globals
+// survive across the scheduler process lifetime.
+var (
+	globalMu        sync.Mutex
+	globalUsage     = make(map[string]map[string]float64) // [queue][user] → resource-seconds
+	globalLastCycle time.Time
+)
```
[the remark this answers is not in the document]
dmildh-absci: Addressed in bfbd07e. OnSessionOpen now holds globalMu through the entire decay + usage accumulation phase, then creates a deep copy into fsp.sessionUsage via snapshotUsage(). All callback reads (JobOrderFn, JobEnqueueableFn, EventHandler) use the session-local snapshot — no lock needed for reads.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,529 @@
...
+	now := time.Now()
+	var elapsed time.Duration
+	if !globalLastCycle.IsZero() {
+		elapsed = now.Sub(globalLastCycle)
+		decayAllUsage(elapsed, fsp.halfLife)
+		klog.V(3).Infof("fairshare: decay applied — elapsed=%s factor=%.6f halfLife=%s",
+			elapsed.Round(time.Millisecond), DecayFactor(elapsed, fsp.halfLife), fsp.halfLife)
+	}
+	globalLastCycle = now
+	globalMu.Unlock()
```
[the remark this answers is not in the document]
dmildh-absci: Addressed in bfbd07e. The globalMu lock is now held through decay, usage accumulation, and the snapshotUsage() call. Only released after fsp.sessionUsage is populated. See the refactored OnSessionOpen — lock spans lines 162-224.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,529 @@
...
+		if lQueue != rQueue {
+			return 0
+		}
+
+		qs := fsp.queues[lQueue]
+		lUser := fsp.getUserFromJob(lJob)
+		rUser := fsp.getUserFromJob(rJob)
+
+		lUsage := ensureGlobalQueueUsage(lQueue)[lUser]
+		rUsage := ensureGlobalQueueUsage(lQueue)[rUser]
```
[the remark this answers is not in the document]
dmildh-absci: Addressed in bfbd07e. JobOrderFn (and all other callbacks) now read from fsp.sessionUsage instead of globalUsage. The session snapshot is an immutable deep copy created under the lock in OnSessionOpen, so no concurrent access issues.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,529 @@
...
+		if queueInfo.Name == queueName && queueInfo.Queue != nil {
+			cap := queueInfo.Queue.Spec.Capability
+			if cap != nil {
+				capResource := api.NewResource(cap)
+				if total := capResource.Get(resourceKey); total > 0 {
+					return total
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dmildh-absci: Addressed in bfbd07e. Now uses direct map lookup: ssn.Queues[api.QueueID(queueName)] for O(1) access, as suggested.

docs/design/fairshare.md
```diff
@@ -1,179 +1,164 @@
...
-    NamespaceInfo  map[string]*kbapi.NamespaceInfo
-    ...
-}
-type ClusterInfo struct {
-    ...
-    NamespaceInfo  map[string]*kbapi.NamespaceInfo
-    ...
-}
+var (
+    globalMu        sync.Mutex
```
JesseStutler, line 61: Since scheduling is based on historical data profiles, shouldn't the data be persisted? Slurm stores data in a database, so in the Kubernetes ecosystem, it should at least be periodically flushed to the CRD for persistence. Otherwise, if it's not persisted, this state will be lost if the scheduler restarts. Perhaps we could create a goroutine to periodically synchronize and flush the CRD. WDYT?

JesseStutler: Maybe we can use sync.Once to initial only one time when New the fair share plugin to create a seperate goroutine periodically to fresh current globalUsage state to the CRD

dmildh-absci: Agreed — implemented in 344ff23. State is now optionally persisted to a ConfigMap (`fairshare-usage-state`).

**How it works:**
- `sync.Once` in `OnSessionOpen()` loads any existing state on the first scheduling cycle and starts a background goroutine (per your suggestion)
- The goroutine flushes `globalUsage` + `globalLastCycle` to the ConfigMap every 30s (configurable via `fairshare.flushIntervalSeconds`)
- On restart, the loaded `globalLastCycle` drives the decay calculation in the next `OnSessionOpen()`, so the time the scheduler was down is correctly accounted for

**New arguments:**
- `fairshare.persistState: "true"` — opt-in, disabled by default for backward compat
- `fairshare.stateNamespace` — defaults to `volcano-system`
- `fairshare.flushIntervalSeconds` — defaults to 30

I chose a ConfigMap over a new CRD to keep this initial implementation minimal (no code generation, no CRD registration, no new RBAC). If the community prefers a dedicated CRD for schema validation and `kubectl get fairsharestates`, that can be a follow-up.

8 new unit tests cover create, update, load, round-trip, disabled mode, missing/empty/corrupt ConfigMap scenarios.

dmildh-absci:
Force-pushed `11646e123` with three improvements (caught during downstream review):

- **`sync.Once` copy in `persist_test.go`** — `go vet` / golangci-lint would flag `savedOnce := persistOnce` since `sync.Once` contains `sync.noCopy`. Reset with `persistOnce = sync.Once{}` before and after instead.
- **Configurable ConfigMap name** — added `fairshare.stateConfigMap` argument (default: `fairshare-usage-state`) alongside the existing `stateNamespace` so both are configurable.
- **API call timeouts** — added `apiCallTimeout(cfg) = cfg.flushInterval / 2` and wrapped the Get/Update/Create calls in `context.WithTimeout`. A slow/stalled apiserver can no longer block the flush ticker beyond a single interval.

Verified locally:
- `go vet ./pkg/scheduler/plugins/fairshare/` — clean
- `go test -vet=all -race ./pkg/scheduler/plugins/fairshare/ -count=1` — all 33 tests pass

dmildh-absci:
The recommended changes were applied to this PR if somebody would like to review.

JesseStutler:
Thanks for your contribution, I'm going to put this feature inside v1.16, will take a look after v1.15 released, it's on my radar

docs/design/fairshare.md
```diff
@@ -1,179 +1,215 @@
-# Namespace fair share
+# Per-User Fair Share Scheduling with Decayed Usage Tracking
 
-[@lminzhw](http://github.com/lminzhw); May 8, 2019
+## Background
```
JesseStutler: I think the background section needs more detail. While Volcano supports fair sharing across queues, some companies map tenants to namespaces rather than queues—meaning multiple tenants might submit jobs to the same queue. In such cases, fair sharing *within* the queue (aggregated by namespace) is required.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,590 @@
...
+
+	var rk string
+	arguments.GetString(&rk, "fairshare.resourceKey")
+	if rk != "" {
+		fsp.defaultResource = rk
+	}
+
+	var queuesArg string
+	arguments.GetString(&queuesArg, "fairshare.targetQueues")
+	if queuesArg != "" {
```
JesseStutler, line 148: If the admin don't specify targetQueues, do we need to fall back to all queues?

pkg/scheduler/plugins/fairshare/persist.go
```diff
@@ -0,0 +1,200 @@
...
+var (
+	persistOnce   sync.Once
+	persistClient kubernetes.Interface
+)
+
+// initPersistence loads existing state from the ConfigMap and starts a
+// background goroutine that periodically flushes globalUsage back.
+// Uses sync.Once so only the first call (first scheduling cycle) has effect.
+// Must be called BEFORE acquiring globalMu to avoid deadlock.
+func initPersistence(client kubernetes.Interface, cfg persistConfig) {
```
JesseStutler, line 58: Can we persist through `CloseSession`? I think goroutine + ticker in the plugin is weird, and it doesn't support hot update if we want to disable the plugin. Then we even don't need that global mutex lock then

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,590 @@
...
+		}
+		if !rTarget {
+			return 1
+		}
+		if lQueue != rQueue {
+			return 0
+		}
+
+		qs := fsp.queues[lQueue]
+		lUser := fsp.getUserFromJob(lJob)
```
JesseStutler: Could we stick with using "namespace"? The terminology is more intuitive that way—after all, it’s not common that other companies are also one-to-one mapping between users and namespaces.

docs/design/fairshare.md
JesseStutler, line 1: I suggest not replacing the whole doc, but to supplement the new design.

docs/design/fairshare.md
```diff
@@ -1,179 +1,215 @@
...
+  the 1 MB ConfigMap limit.
+- **Backward compatibility**: Persistence is disabled by default. Existing deployments are
+  unaffected.
 
-### DRF plugin
-
-DRF plugin use preemption and order of job to balance resource among jobs. The `share` in this plugin is defined as resource usage, the higher `share` means this job occupies the more resource now.
-
-#### Namespace Compare
+## Configuration
```
JesseStutler: Better split into a seperate user guide doc under `docs/user-guide/`, in particular, need guidance on whether it can be used together with DRF plugin.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,590 @@
...
+
+			qs := fsp.queues[queueName]
+			user := fsp.getUserFromJob(job)
+			share, ok := qs.fairShares[user]
+			if !ok {
+				return util.Abstain
+			}
+
+			running := qs.userRunning[user]
+			jobRes := jobTotalResource(job, qs.resourceKey)
```
JesseStutler, line 391: Why AllocateFunc didn't refresh globalUsage?

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,590 @@
...
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/plugins/util"
+)
+
+// Package-level persistent state. Volcano calls New() every scheduling cycle,
+// so state stored on the plugin instance is lost between cycles. These globals
+// survive across the scheduler process lifetime.
+var (
+	globalMu        sync.Mutex
```
JesseStutler: To be honest, I don't find this use of global variables particularly elegant，however, Volcano's current plugin model ties the plugin lifecycle to the session. This forces certain Stateful related plugins to define global variables and rely on `sync.Once` mechanisms—an approach I consider inelegant. In the future, I think we could to consider refactoring some of these stateful plugins to operate at the process level rather than being bound to the session lifecycle. /cc @hajnalmt @devzizu @wangyang0616

JesseStutler:
Also could you please help review when you have time? /cc @hajnalmt @devzizu @wangyang0616 Thanks

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Addressed in 11270a4. When `fairshare.targetQueues` is unset, the plugin now defaults to targeting every queue instead of an empty allowlist (previously an omitted argument silently disabled the plugin entirely). Explicitly setting `targetQueues` still restricts to that allowlist as before.

Added `targetAllQueues` on the plugin struct, checked in `getQueueName`, and the per-cycle queue-state init in `OnSessionOpen` now iterates `ssn.Queues` in that mode (shared via a new `initQueueState` helper). 4 new unit tests cover both the default and explicit-allowlist paths.

JesseStutler: Cool @dmildh-absci Please also check my other comments, thanks, I'd llike to push forward the feature

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Addressed in ad3c7fe. Renamed the identity terminology from "user" to "namespace" throughout: `getUserFromJob`→`getNamespaceFromJob`, `userRunning`/`userDemand`→`namespaceRunning`/`namespaceDemand`, `lUser`/`rUser`→`lNamespace`/`rNamespace`, log fields, and the test names/fixtures. Also reworded the package doc comment to say namespace is used directly rather than a separate "user" concept, since not every cluster maps users to namespaces one-to-one.

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Addressed in ad3c7fe with a code comment on AllocateFunc/DeallocateFunc explaining the design: they only update the session-local `namespaceRunning` count (used by JobOrderFn/JobEnqueueableFn for the rest of *this* cycle), not `globalUsage` (the persisted, decayed historical total). `globalUsage` grows by time-integrating running resource (`res * elapsed.Seconds()`) once per cycle in `OnSessionOpen`, using elapsed wall-clock time since the previous cycle — that elapsed time isn't known yet for a mid-cycle allocation event, it only becomes available at the start of the next cycle once the task shows up in `job.TaskStatusIndex`. So a newly allocated task is missing from `globalUsage` for at most one scheduling cycle (~1s), which is negligible against the default 4h half-life.

pkg/scheduler/plugins/fairshare/persist.go
dmildh-absci: Adopted in ad3c7fe, with one adjustment. Replaced the goroutine+ticker with a rate-limited flush from `OnSessionClose` (new `maybeFlush`, called once per scheduling cycle) — this fixes the real problem you flagged: a detached ticker has no shutdown hook, so removing `fairshare` from the tier list (or a config hot-reload) doesn't stop it, and it keeps writing stale state forever. Tying the flush to the session lifecycle means persistence naturally stops the moment the plugin stops being invoked.

I kept `globalMu` though — it's still needed for the in-memory decay/accumulation across cycles regardless of persistence mode (that's true even with `persistState=false`, where there's no ConfigMap at all). Doing a synchronous ConfigMap load+flush every `OnSessionClose` (~1s cadence) instead of globals would trade the ticker-lifecycle issue for meaningfully more API-server QPS, so `maybeFlush` rate-limits to the configured `flushIntervalSeconds` (same cadence as the old ticker) rather than flushing every cycle.

docs/design/fairshare.md
dmildh-absci: Addressed in 2ab4cb9 — added a paragraph to the Background section calling out exactly this: Volcano already supports fair sharing across queues, and both this proposal and the 2019 Namespace fair share proposal above it are specifically about fair sharing *within* a single queue, for the case where tenants map to namespaces rather than to dedicated queues (multiple tenants submitting into the same queue).

docs/design/fairshare.md
dmildh-absci: Addressed in 2ab4cb9 — restored the original 2019 "Namespace fair share" content (@lminzhw) that my earlier diff had wholesale-replaced, and appended the decayed-usage-tracking proposal as a clearly delineated second section with a short intro explaining how the two relate (same underlying gap, different non-overlapping mechanisms). Sorry for wiping it out originally — should have caught that in my own diff review.

docs/design/fairshare.md
dmildh-absci: Addressed in 2ab4cb9 — moved the practical usage content (Configuration, Arguments table, interaction with other plugins, Logging) out of the design doc into a new docs/user-guide/how_to_use_fairshare_plugin.md, matching the existing how_to_use_<plugin>_plugin.md convention. It includes an explicit "Can it be used together with DRF?" section — yes, they balance different axes of fairness (DRF: which resource dimension dominates the current snapshot; fairshare: which namespace goes next based on historical consumption) and are meant to run together in the same tier, with priority ordered before fairshare.

dmildh-absci:
@JesseStutler Pushed fixes for the review feedback from the 7/15-16 round (ad3c7fe, 2ab4cb9): targetQueues now defaults to all queues, namespace terminology throughout, persistence flushes from OnSessionClose instead of a goroutine+ticker, AllocateFunc design documented in a code comment, and the design doc now restores your original 2019 proposal with the new one as a supplement plus a split-out user guide. Replied inline on each thread with specifics — would appreciate another look when you have time.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,637 @@
...
+		}
+		if !lTarget {
+			return -1
+		}
+		if !rTarget {
+			return 1
+		}
+		if lQueue != rQueue {
+			return 0
+		}
```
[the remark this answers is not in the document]
dmildh-absci: Good catch — real bug, and pre-existing (not introduced by the recent push). Fixed in 17d9251: extracted the abstain check into a standalone `shouldAbstainOrdering` helper and now abstain (return 0) whenever either job isn't targeted, or the two jobs are in different targeted queues — no more picking a winner based on targeting scope alone. 5 new unit tests cover the cases directly.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,637 @@
...
+			fsp.initQueueState(ssn, queueInfo.Name)
+		}
+	} else {
+		for queueName := range fsp.targetQueueNames {
+			fsp.initQueueState(ssn, queueName)
+		}
+	}
+
+	// On the first cycle, load persisted state and start the flush goroutine.
+	// Must happen before acquiring globalMu (loadState takes the lock internally).
```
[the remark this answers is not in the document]
dmildh-absci: Fixed in 17d9251 — updated the comment to reflect that persistence flushing happens via OnSessionClose/maybeFlush, not a goroutine. Leftover from the earlier persistence rework, thanks for catching it.

pkg/scheduler/plugins/fairshare/persist.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	flushMu.Unlock()
+
+	if !due {
+		return
+	}
+
+	if err := flushState(cfg); err != nil {
+		klog.Warningf("fairshare: flush failed: %v", err)
+	}
```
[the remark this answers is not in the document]
dmildh-absci: Correct, and a real bug in what I'd just pushed. Fixed in f560d0b: lastFlushAt now only advances after flushState returns nil, so a transient failure is retried on the very next OnSessionClose call instead of being suppressed for a full flushIntervalSeconds. New test simulates a failed ConfigMap Get via a fake-clientset reactor to confirm the retry-promptly behavior.

docs/design/fairshare.md
```diff
@@ -1,8 +1,26 @@
...
+1. [Namespace fair share](#namespace-fair-share-2019) (2019, [@lminzhw](http://github.com/lminzhw)) —
+   static per-namespace weights recorded on `ResourceQuota`, compared via a new
+   `NamespaceOrderFn` scheduling-loop stage.
+2. [Per-namespace fair share with decayed usage tracking](#per-namespace-fair-share-with-decayed-usage-tracking-2026)
+   (2026) — a self-contained scheduler plugin that tracks historical resource-seconds per
+   namespace with exponential decay, addressing the same underlying problem without
+   requiring new `Session`/`SchedulerCache` fields or a `NamespaceOrderFn` stage.
+
+For usage instructions (arguments, configuration examples, interaction with other plugins),
+see the [fairshare user guide](../user-guide/fairshare.md).
```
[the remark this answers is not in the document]
dmildh-absci: Fixed in 36c761e — corrected the link to docs/user-guide/how_to_use_fairshare_plugin.md, the actual filename I used.

docs/user-guide/how_to_use_fairshare_plugin.md
```diff
@@ -0,0 +1,113 @@
...
+Yes — `fairshare` and `drf` address different axes of fairness and are meant to be run
+together, not as alternatives:
+
+- **DRF** balances *which resource dimension* dominates a job's share (CPU vs. memory vs.
+  GPU) at the current snapshot.
+- **fairshare** balances *which namespace* gets scheduled next based on historical
+  consumption over time, within the same queue.
+
+Put `fairshare` in the same tier as `drf` (order between them does not matter — they operate
+on independent ordering criteria and Volcano composes `JobOrderFn` results across plugins).
```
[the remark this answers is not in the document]
dmildh-absci: Right, that claim was wrong. Fixed in 36c761e — Session.JobOrderFn walks tiers-then-plugins in configured order and returns as soon as any plugin's comparison is non-zero, so order absolutely matters (and it's not a same-tier-vs-different-tier question). Documented that fairshare should be listed before drf: drf's dominant-resource-share comparison returns non-zero for almost any differing pair, so listing it first would leave fairshare's historical-usage ordering with little chance to actually run. The example config already had this order right; the prose claiming order didn't matter was just wrong.

test/e2e/schedulingaction/fairshare.go
```diff
@@ -0,0 +1,171 @@
...
+		// Verify User A's extra job is still pending (or at least was scheduled after User B).
+		err = wait.Poll(2*time.Second, 30*time.Second, func() (bool, error) {
+			job, getErr := ctx1.Vcclient.BatchV1alpha1().Jobs("fairshare-user-a").Get(
+				context.TODO(), createdAExtraJob.Name, metav1.GetOptions{})
+			if getErr != nil {
+				return false, getErr
+			}
+			// If User A's extra job is still pending, fairshare is working correctly.
+			// If it becomes ready, we check the timeline against User B's job.
+			return job.Status.Running > 0 || job.Status.Pending > 0, nil
```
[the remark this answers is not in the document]
dmildh-absci: Fixed in 36c761e — the poll now requires Pending > 0 specifically and treats Running > 0 as an explicit failure with a descriptive error, so it actually verifies the ordering claim (User A's extra job stays pending behind lower-usage User B) instead of trivially passing either way.

pkg/scheduler/plugins/fairshare/fairshare_test.go
```diff
@@ -0,0 +1,536 @@
...
+	usage["alice"] = 42.0
+
+	if globalUsage["new-queue"]["alice"] != 42.0 {
+		t.Error("expected usage to be stored in global state")
+	}
+
+	globalMu.Lock()
+	globalUsage = oldUsage
+	globalMu.Unlock()
+}
```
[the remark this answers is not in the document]
dmildh-absci: Fixed in 17d9251 — the test now holds globalMu for its full read-modify-check sequence, matching ensureGlobalQueueUsage's documented lock contract. Harmless in this specific single-goroutine test today, but worth being correct about the contract regardless.

pkg/scheduler/plugins/fairshare/persist.go
```diff
@@ -0,0 +1,231 @@
...
+	// failure (e.g. a momentary apiserver error) is retried on the very
+	// next cycle instead of being suppressed for a full flushInterval.
+	if err := flushState(cfg); err != nil {
+		klog.Warningf("fairshare: flush failed: %v", err)
+		return
+	}
+
+	flushMu.Lock()
+	lastFlushAt = time.Now()
+	flushMu.Unlock()
```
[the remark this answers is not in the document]
dmildh-absci: Fixed in 914e8ce — flushMu is now held across the whole due-check + flush + lastFlushAt update, not just the check. Serializes concurrent flush attempts instead of letting them both observe due=true and race.

pkg/scheduler/plugins/fairshare/persist.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// apiCallTimeout returns a timeout for individual API calls. We use half the
+// flush interval so a slow/stalled request can't block the next flush tick
+// (time.Ticker drops ticks if the handler runs long).
+func apiCallTimeout(cfg persistConfig) time.Duration {
```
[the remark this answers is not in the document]
dmildh-absci: Fixed in 914e8ce — reworded to describe the flushMu/OnSessionClose mechanism instead of the old ticker.

test/e2e/schedulingaction/fairshare.go
```diff
@@ -0,0 +1,179 @@
...
+
+		// Create test contexts for two different namespaces (simulating two users).
+		ctx1 := e2eutil.InitTestContext(e2eutil.Options{
+			Namespace:     "fairshare-user-a",
+			Queues:        []string{"default"},
+			NodesNumLimit: 2,
+			NodesResourceLimit: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("2000m"),
+				corev1.ResourceMemory: resource.MustParse("2048Mi"),
+			},
```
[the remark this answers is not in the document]
dmildh-absci: Confirmed against setPlaceHolderForSchedulerTesting — NodesResourceLimit is indeed per-node. Fixed in 12e5250: switched to NodesNumLimit=1 with the CPU limit exactly matching User A's 4*500m, so the initial jobs genuinely saturate all schedulable CPU (matching the single-node capacity-constrained pattern already used by allocate.go/enqueue.go/guarantee.go).

test/e2e/schedulingaction/fairshare.go
```diff
@@ -0,0 +1,179 @@
...
+			Queues:    []string{"default"},
+		})
+		defer e2eutil.CleanupTestContext(ctx2)
+
+		cpuSlot := corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("500m"),
+			corev1.ResourceMemory: resource.MustParse("256Mi"),
+		}
+
+		// User A submits 4 jobs first, filling available capacity.
```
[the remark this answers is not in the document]
dmildh-absci: Same fix as the other instance of this comment — see 12e5250 (NodesNumLimit=1, CPU sized to exactly match User A's 4*500m).

test/e2e/schedulingaction/fairshare.go
```diff
@@ -0,0 +1,179 @@
...
+					"fairshare.halfLifeMinutes": "60",
+				},
+			}
+			if len(sc.Tiers) > 0 {
+				sc.Tiers[0].Plugins = append(sc.Tiers[0].Plugins, fairsharePlugin)
+			} else {
+				sc.Tiers = append(sc.Tiers, e2eutil.Tier{
+					Plugins: []e2eutil.PluginOption{fairsharePlugin},
+				})
+			}
```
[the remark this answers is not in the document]
dmildh-absci: Good catch, and directly relevant given the ordering behavior we just documented — appending after whatever's already in tier 0 gave no guarantee fairshare lands before drf. Fixed in 12e5250: now uses Tier.GetPluginIdxOf("drf") to insert fairshare immediately before drf specifically, falling back to append only if drf isn't configured at all.

test/e2e/schedulingaction/fairshare.go
```diff
@@ -0,0 +1,179 @@
...
+			return true
+		}
+		cmc.ChangeBy(func(data map[string]string) (changed bool, changedBefore map[string]string) {
+			return e2eutil.ModifySchedulerConfig(data, modifier)
+		})
+		defer cmc.UndoChanged()
+
+		// Wait for config to take effect
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
dmildh-absci: Fixed in 12e5250 — removed the extra sleep. ChangeBy already waits for the config change to take effect internally (bumps a refreshts annotation on the scheduler pods to force an immediate remount), and none of the other scheduler-config-modifying E2E tests in this package (enqueue.go, reclaim.go, preempt.go) add an extra sleep on top of that, so this one shouldn't either.

docs/design/fairshare.md
```diff
@@ -176,4 +194,182 @@ All these plugin would choose some victims respective, and the intersection of t
...
+
+| Hook | Purpose |
+|------|---------|
+| `JobOrderFn` | Orders jobs by cumulative usage (lower wins), with running-resource tiebreaker |
+| `JobEnqueueableFn` | (Optional) Blocks namespaces at/above their max-min fair share from the scheduling pipeline |
+| `EventHandler` | Tracks allocations/deallocations in real-time during the scheduling cycle |
+
+### Testing
+
+#### Unit tests (47 tests)
```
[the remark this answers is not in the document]
dmildh-absci: Fixed — updated the PR description's test count and reconciled it with the design doc (both now say 49, the current count).

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,644 @@
...
+	}
+	return ssn.TotalResource.Get(resourceKey)
+}
+
+func (fsp *fairSharePlugin) getNamespaceFromJob(job *api.JobInfo) string {
+	if job.Namespace != "" {
+		return job.Namespace
+	}
+	return defaultUnknownNamespace
+}
```
devzizu, line 595: Is `job.Namespace == ""` actually reachable in practice, or defensive-only? If defensive-only, consider a one-time `klog.Warning` when the fallback triggers. Silently bucketing all such jobs into a shared `_unknown` namespace could quietly break fairness guarantees for whichever real tenant happens to trigger this path, with no signal that it happened. This is a reasonable fallback to avoid a crash, but it means a bug elsewhere.

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,644 @@
...
+		if secs, err := strconv.Atoi(strings.TrimSpace(flushStr)); err == nil && secs > 0 {
+			fsp.persistCfg.flushInterval = time.Duration(secs) * time.Second
+		}
+	}
+
+	queuesLog := interface{}("all")
+	if !fsp.targetAllQueues {
+		queuesLog = fsp.targetQueueNames
+	}
+	klog.V(2).Infof("fairshare: plugin created — queues=%v resource=%s halfLife=%s enqueueGate=%v persist=%v",
```
devzizu: I don't know if this is some convention we should keep, but DRF follows `[drf] <log>` pattern. Should we use `[fairshare]` in logs as well?

Ref: https://github.com/volcano-sh/volcano/blob/8a2c34cf264552faac0e1914c50041baf7d2270a/pkg/scheduler/plugins/drf/drf.go#L244

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,644 @@
...
+
+	var gateStr string
+	arguments.GetString(&gateStr, "fairshare.enableEnqueueGate")
+	fsp.enableEnqueueGate = strings.EqualFold(strings.TrimSpace(gateStr), "true")
+
+	fsp.halfLife = defaultHalfLifeMinutes * time.Minute
+	var halfLifeStr string
+	arguments.GetString(&halfLifeStr, "fairshare.halfLifeMinutes")
+	if halfLifeStr != "" {
+		if mins, err := strconv.Atoi(strings.TrimSpace(halfLifeStr)); err == nil && mins > 0 {
```
devzizu, line 177: Should we warn a log here for badly configured values? (e.g., "abc" or "-5")

docs/design/fairshare.md
```diff
@@ -176,4 +194,184 @@ All these plugin would choose some victims respective, and the intersection of t
...
+Validated on a test cluster with 2 GPU nodes and 4 tenant namespaces:
+
+1. **Without decay tracking**: FIFO behavior, last tenant waited ~7 minutes
+2. **With decay tracking**: All tenants got GPUs within ~2 minutes, scheduling rotated between tenants
+3. **Burst asymmetry** (1 tenant = 8 jobs, others = 1 each): Minority tenants' jobs completed within ~2 minutes
+4. **DAG/workflow simulation**: GPU steps interleaved across tenants by cumulative usage
+5. **PriorityClass interaction**: High-priority bypassed fair share as expected
+6. **Scheduler restart**: Running jobs survived, new jobs scheduled fairly post-restart
+
+### Limitations
```
devzizu, line 383: Unlike DRF, the introducing of the `fairshare` plugin registers no `PreemptableFn` or `ReclaimableFn`, focusing only in the reordering for the pending queue and (optionally) controls the enqueue flow. Worth being explicit in the user guide that fair share here is purely advisory over new scheduling decisions: if namespace A's jobs are already running and long-lived, fairshare cannot preempt them to admit namespace B sooner, no matter how decayed/stale A's usage penalty becomes.

Also, should we consider tracking this in a diff followup issue?

docs/design/fairshare.md
```diff
@@ -1,8 +1,26 @@
-# Namespace fair share
+# Fair share scheduling
```
devzizu, line 1: One thing i noticed in this PR is an observability gap that we should cover. This PR only describes the visibility over issues through klog calls at various levels, and there is no mention for how to observed decay values from an operator perspective (e.g., "namespace X decayed usage over time") that could feed grafana dashboards, for example.

I also noticed that `pkg/scheduler/metrics/namespace.go` already defines `UpdateNamespaceShare`/`UpdateNamespaceWeight`/etc. with prometheus gauges (`namespace_share`, `namespace_weight`, `namespace_weighted_share`), but none of the three are called anywhere in the codebase today. These look like leftover plumbing from the previous proposal. 

Example:

https://github.com/volcano-sh/volcano/blob/8a2c34cf264552faac0e1914c50041baf7d2270a/pkg/scheduler/metrics/namespace.go#L51

That said, `fairshare` could compute these per-namespace values these gauges were built for, following the same `metrics.UpdateJobShare(...)` from `AllocateFunc`/`DeallocateFunc`/`OnSessionOpen` pattern DRF already uses. 

WDYT @dmildh-absci @JesseStutler?

pkg/scheduler/plugins/fairshare/fairshare.go
devzizu: @JesseStutler agreed this is worth tracking separately, not blocking on this PR.

A possible additive fix:

```go
type StatefulPluginBuilder func(arguments Arguments) StatefulPlugin

type StatefulPlugin interface {
   Plugin
   // Init runs once at scheduler startup
   Init(stopCh <-chan struct{})
}
```

`OpenSession` would reuse one long-lived instance per plugin name instead of discarding it every cycle and finally `fairshare`, usage/globalMu would move onto the struct itself.

```go
for _, tier := range tiers {
        var p Plugin
        if sp, found := GetOrCreateStatefulPlugin(plugin.Name, plugin.Arguments, stopCh); found {
            p = sp   // <-- reused object
        } else if pb, found := GetPluginBuilder(plugin.Name); found {
            p = pb(plugin.Arguments)        
        }
        
        ...
        ssn.plugins[p.Name()] = p
        p.OnSessionOpen(ssn)
    }
}
```

Please lmk if what you meant is similar to what i have proposed here, but we can also move this discussion to a separate issue/thread.

devzizu (commented):
Given that we can run multiple shards of the scheduler independently-configured, and `stateNamespace`/`stateConfigMap` default to the same values regardless of scheduler name, two named schedulers both enabling `fairshare.persistState=true` without explicitly overriding the ConfigMap `name/namespace` would silently conflict each other's persisted state. 

I would consider defaulting `stateConfigMap` to something scoped by scheduler name (e.g. include --scheduler-name in the default).

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Addressed in 27ecd5c — switched all fairshare klog messages (fairshare.go + persist.go) from the `fairshare: ` prefix to `[fairshare] `, matching drf's convention.

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Addressed in ad781f2 — this is defensive-only (not reachable in normal operation, since Volcano jobs always carry a namespace), but added a one-time `klog.Warning` via `sync.Once` the first time the fallback triggers, so a bug elsewhere wouldn't silently and repeatedly bucket jobs into `_unknown` with no signal.

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Addressed in ad781f2 — added a `klog.Warning` when `fairshare.halfLifeMinutes` fails to parse or is non-positive, naming the bad value and the default it's falling back to. Applied the same fix to `fairshare.flushIntervalSeconds` for consistency, since it has the identical silent-fallback shape.

docs/design/fairshare.md
dmildh-absci: Addressed in ad781f2 — added this exact point to the design doc's Limitations section, and mirrored it in the user guide (new Limitations section) since that's the more likely place an operator would look for it. Happy to file a follow-up issue to track preemption-awareness as a future enhancement if you'd like it tracked separately — not attempting it in this PR.

docs/design/fairshare.md
dmildh-absci: Good catch on the leftover plumbing — before writing anything here I'd like a call on direction, since this touches shared metrics infra beyond this plugin's own package:

- `namespace_weight`/`namespace_weighted_share` carry the semantics of the 2019 weight-based proposal (an admin-set integer weight). fairshare has no weight concept — it's purely decay-based usage — so populating those two would be a semantic stretch rather than a natural fit. `namespace_share` (the generic "share for one namespace" gauge) maps more directly onto fairshare's computed `fairShares[namespace]` output, for whichever queue is currently under evaluation.
- Reusing the existing gauges from this PR broadens the blast radius past the fairshare package itself. Worth confirming this is in scope for #5161, versus a fast-follow once the plugin lands, especially since the gauges are currently dead code and nothing depends on their exact shape today.
- Cardinality: fairshare tracks usage per `(queue, namespace)` pair internally, but the existing gauges are namespace-only labeled. If a namespace participates in multiple targeted queues, collapsing to namespace-only loses information; adding a `queue` label changes the metric's shape from what DRF/2019 established.

So: (a) reuse `namespace_share` as-is (namespace-only, last-queue-wins or similar), (b) reuse it but add a `queue` label (breaking change to the existing metric's cardinality), (c) introduce fairshare-specific metric(s) instead (e.g. `namespace_decayed_usage`), or (d) defer to a follow-up PR/issue once the core plugin merges? Leaning toward (c) or (d) myself, but want your read given you both know this metrics surface's history better than I do. @JesseStutler @devzizu

JesseStutler: I don’t think the current metric shape needs to be treated as a compatibility constraint, since these metrics are unused today. If `namespace_share` semantically maps to fairshare’s calculated allocation, we can reuse it and add the dimensions fairshare actually needs, such as `resource`. If the semantics do not match, we can rename or redesign it instead.

I would not reuse `namespace_weight` or `namespace_weighted_share`, since the current fairshare implementation has no weight concept. Decayed usage should also be a separate metric because it has different semantics and units from the calculated share.

pkg/scheduler/plugins/fairshare/persist.go
```diff
@@ -0,0 +1,238 @@
...
+	// Hold flushMu across the due-check *and* the flush itself (not just
+	// the check). Volcano's main scheduler calls OnSessionClose
+	// sequentially today, but nothing here should assume that forever —
+	// if maybeFlush were ever called concurrently (e.g. from an
+	// Agent-Scheduler-style multi-worker setup), releasing the lock
+	// between the check and the flush would let two callers both observe
+	// due=true and both call flushState in parallel: duplicate API calls
+	// and a possible resourceVersion conflict on Update. Holding the lock
+	// for the whole sequence serializes flush attempts instead.
+	flushMu.Lock()
```
JesseStutler: I don't think we don't need this lock at the moment, and I think Agent Scheduler won't add the fair share plugin, so it can be removed currently.  The current session can be safely updated. If future needs multi-worker, we should redesign, not by a global lock

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,660 @@
...
+	"volcano.sh/volcano/pkg/scheduler/api"
+	"volcano.sh/volcano/pkg/scheduler/framework"
+	"volcano.sh/volcano/pkg/scheduler/plugins/util"
+)
+
+// Package-level persistent state. Volcano calls New() every scheduling cycle,
+// so state stored on the plugin instance is lost between cycles. These globals
+// survive across the scheduler process lifetime.
+var (
+	globalMu        sync.Mutex
```
JesseStutler: Like flushMu, do we still need the globalMu now? Currently, all operations are performed within the goroutine for the session, so there are no additional cocurrent operations. Furthermore, API operations are currently synchronous. So can we remove the globalMu? If future needs cocurrent, we should redesign rather a big lock

pkg/scheduler/plugins/fairshare/persist.go
```diff
@@ -0,0 +1,238 @@
...
+type persistConfig struct {
+	enabled       bool
+	namespace     string
+	configMapName string
+	flushInterval time.Duration
+}
+
+var (
+	persistOnce   sync.Once
+	persistClient kubernetes.Interface
```
JesseStutler: I think there is no need to define `persistClient` as a global var, just pass `ssn.KubeClient`

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -0,0 +1,660 @@
...
+	globalUsage     = make(map[string]map[string]float64) // [queue][namespace] → resource-seconds
```
JesseStutler: And BTW, could you integrate these global variables and `lastFlushAt` in `persist.go` into a  struct and add some comments? That would be cleaner. Like this:
```
type fairShareProcessState struct {
	usage     map[string]map[string]float64
	lastCycle time.Time

	persistenceInitialized bool
	lastFlushAt            time.Time
}
```

test/e2e/schedulingaction/fairshare.go
```diff
@@ -0,0 +1,212 @@
...
+		createdAExtraJob := e2eutil.CreateJob(ctx1, userAExtraJob)
+
+		// Delete one of User A's running jobs to free capacity.
+		err := ctx1.Vcclient.BatchV1alpha1().Jobs("fairshare-user-a").Delete(
+			context.TODO(), userAJobs[0], metav1.DeleteOptions{})
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		// User B's job should become ready (scheduled) before User A's extra job,
+		// because User B has lower historical usage.
+		err = e2eutil.WaitJobReady(ctx2, createdBJob)
```
JesseStutler, line 231: We should make this E2E scenario deterministic and ensure it actually validates fairshare rather than the default FIFO ordering

Currently, User B’s job is created before User A’s extra job. Since Volcano falls back to creation timestamp when the ordering plugins tie or fairshare is not consulted, B may still be scheduled first even if fairshare is broken or disabled. Also, the slot is released immediately after creating the two jobs, so there is no guarantee that both contenders are already visible to the scheduler in the same scheduling cycle.

The expected sequence could be:

1. Let User A fill the available capacity and accumulate historical usage.
2. Create User A’s extra job first and wait until it is pending.
3. Create User B’s job afterwards and wait until it is pending.
4. Confirm User A’s extra job is still pending, so both jobs are simultaneously competing for the same constrained resource.
5. Release exactly one resource slot.
6. Verify that User B’s later-created job becomes running while User A’s earlier-created extra job remains pending.

This deliberately makes FIFO and fairshare produce opposite results: FIFO would select the earlier User A job, while fairshare should select the later User B job because B has lower historical usage. Therefore, if fairshare is disabled or broken, the E2E should reliably fail rather than occasionally passing due to creation order or scheduler timing.

JesseStutler (commented):
I have no big issues with the design,  just some suggestions about variables and e2e. Willing to push forward into v1.16

JesseStutler:
/ok-to-test
@dmildh-absci 
And please squash your commits into 2~3, For example, split into design/user-guide docs and implementation codes. There are too many commits currently, and many commits contain too much AI-generated information. It's not recommended to have too many AI traces in commits, they still need some manual processing

pkg/scheduler/plugins/fairshare/persist.go
dmildh-absci: Verified: `pkg/scheduler/scheduler.go`'s `Run()` calls `go wait.Until(pc.runOnce, pc.schedulePeriod, stopCh)` — `wait.Until` runs `runOnce` synchronously in a loop (call, sleep, call again), never overlapping. There's exactly one goroutine ever invoking plugin session hooks today, so your assessment holds up. Removed `globalMu`/`flushMu` and consolidated everything into a single unlocked `fairShareProcessState` struct (with a comment documenting why no lock is needed and what would have to change if that stops being true) in `2bcbce4`.

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Verified: `pkg/scheduler/scheduler.go`'s `Run()` calls `go wait.Until(pc.runOnce, pc.schedulePeriod, stopCh)` — `wait.Until` runs `runOnce` synchronously in a loop (call, sleep, call again), never overlapping. There's exactly one goroutine ever invoking plugin session hooks today, so your assessment holds up. Removed `globalMu`/`flushMu` and consolidated everything into a single unlocked `fairShareProcessState` struct (with a comment documenting why no lock is needed and what would have to change if that stops being true) in `2bcbce4`.

pkg/scheduler/plugins/fairshare/fairshare.go
dmildh-absci: Done in `2bcbce4` — consolidated into `fairShareProcessState` exactly as you sketched (usage, lastCycle, persistenceInitialized, lastFlushAt), with a doc comment explaining the no-lock rationale.

pkg/scheduler/plugins/fairshare/persist.go
dmildh-absci: Addressed in `2bcbce4` — dropped `persistClient` and now thread `ssn.KubeClient()` explicitly through `initPersistence`/`maybeFlush`/`loadState`/`flushState`/`createStateConfigMap`.

docs/design/fairshare.md
dmildh-absci: Added in `2bcbce4`: reused `namespace_share` with a `resource` label as you suggested, and I also added a `queue` label since fairshare's share is per-(queue, namespace) — a namespace can have a different share in different targeted queues. Flagging that explicitly since you only mentioned `resource` — happy to drop `queue` if you'd rather keep the label set minimal.

Also added a separate `namespace_decayed_usage` gauge (namespace_name, queue, resource) rather than reusing `namespace_share` for it: share is a point-in-time entitlement recomputed each cycle from current demand, while decayed usage is a running resource-seconds total that persists/decays across cycles — different units and semantics. Both are updated once per cycle in `OnSessionOpen` (not in Allocate/Deallocate), since fairshare's tracked usage number only changes once per cycle there.

`namespace_weight`/`namespace_weighted_share` are left untouched — no weight concept in fairshare.

test/e2e/schedulingaction/fairshare.go
dmildh-absci: Fixed in `2bcbce4`. New sequence: User A's extra job is created first and confirmed `Pending`, then User B's job is created after and confirmed `Pending` too — both now genuinely competing for the same not-yet-freed slot. Only then do we delete one of User A's original jobs to free exactly one slot, and assert User B's (later-created, lower-usage) job goes `Running` while User A's (earlier-created, higher-usage) extra job stays `Pending`. Plain FIFO would pick User A's job here, so a broken/disabled fairshare plugin now fails this test instead of occasionally passing on scheduler timing.

Couldn't run the E2E locally (needs a real cluster) — noting that rather than claiming it was verified end-to-end.

dmildh-absci:
Pushed `2bcbce4` addressing this review round:

- **Locks removed**: `globalMu`/`flushMu` are gone. Verified `wait.Until(pc.runOnce, ...)` in `pkg/scheduler/scheduler.go` never overlaps invocations, so `OnSessionOpen`/`OnSessionClose` only ever run from a single goroutine — no lock needed.
- **State consolidated**: all cross-cycle globals (`globalUsage`, `globalLastCycle`, `persistClient`, `lastFlushAt`) are now one unlocked `fairShareProcessState` struct, documented with the no-lock rationale and what would need to change if that assumption stops holding (e.g. multi-worker Agent Scheduler).
- **Client threading**: `persistClient` global is gone; `kubernetes.Interface` is passed explicitly through `initPersistence`/`maybeFlush`/`loadState`/`flushState`.
- **Metrics added**: `namespace_share` gets `queue`+`resource` labels (flagging the `queue` addition specifically for your veto — you only mentioned `resource`), plus a new `namespace_decayed_usage` gauge for the persisted, decaying usage total. Both wired into `OnSessionOpen`.
- **E2E rewritten**: job creation order now makes plain FIFO and fairshare disagree (User A's extra job is created and confirmed `Pending` before User B's job), so a broken/disabled plugin fails the test instead of occasionally passing on timing. Couldn't run it locally (needs a real cluster).
- **History squashed**: 2 commits — `docs(fairshare): add design doc and user guide` and `feat(fairshare): per-user fair share scheduling plugin with decayed usage tracking` — force-pushed to `feat/fairshare-plugin`.

Ready for another look ahead of the v1.16 cutoff — let me know if anything needs adjusting.

dmildh-absci:
@JesseStutler CI is fully green now (all 30 checks passing, including all E2E suites).

One additional fix beyond the review-comment changes: the new fairshare E2E test was pointing both users' jobs at the `default` queue. Its cleanup closes-then-deletes every queue it owns, but the admission webhook refuses to delete `default`, so it was left permanently `Closed` — which cascaded into failing two unrelated `PodGroup terminating` tests later in the same suite. Fixed by giving the fairshare test its own dedicated queue (`fairshare-e2e-queue`), created/deleted explicitly outside the `default` queue's lifecycle. Folded into the same implementation commit.

Branch is up to date with `master` (rebased, DCO-signed) and ready for another look whenever you have time.

JesseStutler:
/approve
Lgtm for alpha stage, glad to see it be carried into v1.16 and continue involving
Please /cc @devzizu

docs/design/fairshare.md
devzizu: Right, this approach makes sense to me 👍

devzizu:
/approve

I think we’ve addressed the core issues and the remaining nits, so this PR is in a good state to move forward and let users start experimenting with it in v1.16. I'll be one of them for sure!

Thanks for all the effort, approving!

PR #5171, opened 2026-04-09
fix: display friendly "infinity" for max resource values in logs

### LINKED ISSUE #5148: Noisy log outputs on realCapability and capability due to maxFloat and maxInt
### Description

```
```

2. check the logs.

latest

_No response_

Bahtya:
## Summary

Fixes #5148

Replaces noisy `math.MaxFloat64` log output with a readable `"infinity"` string.

## Problem

Since #4973, raising the scheduler log level produces extremely long log lines:
```
realCapability <cpu 17976931348623157081452742373170435679807056752584..., memory 17976931348623157...>
```

This happens because `Resource.String()` formats `math.MaxFloat64` with `%0.2f`, producing hundreds of digits.

## Changes

- **`resource_info.go`**: Added `formatFloat()` helper that returns `"infinity"` for `math.MaxFloat64` and normal `%0.2f` formatting otherwise. Updated `Resource.String()` to use it.
- **`resource_info_test.go`**: Added `TestResource_String` covering normal values, infinite resources, mixed finite/infinite, and zero values.

## Before
```
<cpu 17976931348623157081452742373170435679807056752584..., memory 17976931348623157081452742...>
```

## After
```
<cpu infinity, memory infinity>
```

## Testing

```
=== RUN   TestResource_String
--- PASS: TestResource_String (0.00s)
    --- PASS: normal_values
    --- PASS: infinite_resource_shows_infinity
    --- PASS   mixed_finite_and_infinite_with_scalars
    --- PASS   zero_values
ok  volcano.sh/volcano/pkg/scheduler/api
```

All existing tests in `pkg/scheduler/api/` continue to pass.

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -166,9 +166,18 @@ func (r *Resource) Clone() *Resource {
 	return clone
 }
 
+// formatFloat returns a human-readable string for a resource float value.
+// math.MaxFloat64 is displayed as "infinity" to avoid noisy log output.
+func formatFloat(v float64) string {
+	if v == math.MaxFloat64 {
+		return "infinity"
+	}
+	return fmt.Sprintf("%0.2f", v)
+}
+
 // String returns resource details in string format
 func (r *Resource) String() string {
-	str := fmt.Sprintf("cpu %0.2f, memory %0.2f", r.MilliCPU, r.Memory)
+	str := fmt.Sprintf("cpu %s, memory %s", formatFloat(r.MilliCPU), formatFloat(r.Memory))
 	// Sort scalar resource names to ensure consistent string output
 	var resourceNames []string
 	for rName := range r.ScalarResources {
@@ -176,7 +185,7 @@ func (r *Resource) String() string {
 	}
 	sort.Strings(resourceNames)
 	for _, rName := range resourceNames {
-		str = fmt.Sprintf("%s, %s %0.2f", str, rName, r.ScalarResources[v1.ResourceName(rName)])
+		str = fmt.Sprintf("%s, %s %s", str, rName, formatFloat(r.ScalarResources[v1.ResourceName(rName)]))
 	}
 	return str
 }
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -177,17 +177,19 @@ func formatFloat(v float64) string {
 
 // String returns resource details in string format
 func (r *Resource) String() string {
-	str := fmt.Sprintf("cpu %s, memory %s", formatFloat(r.MilliCPU), formatFloat(r.Memory))
+	var sb strings.Builder
+	fmt.Fprintf(&sb, "cpu %s, memory %s", formatFloat(r.MilliCPU), formatFloat(r.Memory))
+
 	// Sort scalar resource names to ensure consistent string output
-	var resourceNames []string
+	resourceNames := make([]string, 0, len(r.ScalarResources))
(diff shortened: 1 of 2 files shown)


Bahtya:
Great catch! 🎯 I've updated the PR to use `strings.Builder` with pre-allocated slice capacity as suggested. This should be much more efficient for the logging hot path. Thanks for the review!

Bahtya:
Rebased onto latest master to resolve merge conflicts. All CI checks (DCO, github-actions) continue to pass. @hajnalmt @k82cn would appreciate a review when you have a moment.

pkg/scheduler/api/resource_info.go
```diff
@@ -166,19 +166,30 @@ func (r *Resource) Clone() *Resource {
...
 	// Sort scalar resource names to ensure consistent string output
-	var resourceNames []string
+	resourceNames := make([]string, 0, len(r.ScalarResources))
 	for rName := range r.ScalarResources {
 		resourceNames = append(resourceNames, string(rName))
 	}
 	sort.Strings(resourceNames)
 	for _, rName := range resourceNames {
-		str = fmt.Sprintf("%s, %s %0.2f", str, rName, r.ScalarResources[v1.ResourceName(rName)])
+		fmt.Fprintf(&sb, ", %s %s", rName, formatFloat(r.ScalarResources[v1.ResourceName(rName)]))
```
hajnalmt, line 190: We use MaxInt64 for scalar resources.

pkg/scheduler/api/resource_info.go
```diff
@@ -166,19 +166,30 @@ func (r *Resource) Clone() *Resource {
 	return clone
 }
 
+// formatFloat returns a human-readable string for a resource float value.
+// math.MaxFloat64 is displayed as "infinity" to avoid noisy log output.
+func formatFloat(v float64) string {
+	if v == math.MaxFloat64 {
```
hajnalmt, line 172: maxFloat is used for cpu and memory only.

pkg/scheduler/api/resource_info.go
```diff
@@ -166,19 +166,30 @@ func (r *Resource) Clone() *Resource {
...
+		return "infinity"
```
hajnalmt, line 173: Print maxInt or maxFloat instead of infinity, based on the used underlying resource please.

pkg/scheduler/api/resource_info.go
hajnalmt: And we should take into account MaxInt32 too, just to be sure.

hajnalmt: This was a bug to log this as float I think.

hajnalmt (changes requested):
We need to refine this a little bit still, please find my comments below.

Bahtya:
Hi @hajnalmt, thanks for the detailed review! I've updated the PR to address all your feedback:

1. **Print actual max values instead of infinity** — `formatFloat()` now prints `math.MaxFloat64` for float resources, new `formatScalar()` prints `math.MaxInt64`/`math.MaxInt32` for int resources
2. **Handle MaxInt32** — `formatScalar()` checks for both `math.MaxInt64` and `math.MaxInt32` values
3. **CPU/memory use maxFloat, scalars use maxInt** — String() now routes scalar resources through `formatScalar()` (int) and CPU/memory through `formatFloat()` (float)
4. **Fixed the float logging bug** — scalar resources are now formatted as integers (`%d`) instead of floats

All tests pass including 3 new test cases for MaxInt64/MaxInt32/integer formatting. Ready for re-review!

PR #5172, opened 2026-04-10
wait event handler completed before start scheduling

qi-min:
#### What type of PR is this?
fix

record event handlers and wait for all handler handle initial events  before start first scheduling session.
 
Fixes # https://github.com/volcano-sh/volcano/issues/5174

```release-note

```

### DIFF
--- a/cmd/agent-scheduler/app/options/options.go
+++ b/cmd/agent-scheduler/app/options/options.go
@@ -31,14 +31,12 @@ import (
 )
 
 const (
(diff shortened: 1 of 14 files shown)


JesseStutler:
/approve
/lgtm
Thanks for your great contribution. After the investigation, I think we can solve the initial sync bugs. Could you please also cherry-pick to release-1.13 and release-1.12 manually? @qi-min Since release-1.13 and release-1.12 doesn't contain agent-scheduler releated codes and need to do it manually

PR #5176, opened 2026-04-10
fix(capacity): avoid false exceeds on missing parent scalar keys

### LINKED ISSUE #5175: capacity: false capability exceed errors after scalar resource disappears on the cluster
### Description

In hierarchical queue mode, the capacity plugin can emit false-positive **"child capability exceeds parent"** errors when queue capabilities still contain scalar resource keys that disappeared from the cluster's advertised resources.

Our concrete production use case involved MIG-related custom resources under when `nvidia-gpu-h100-mig-1g.10gb` disappeared on the cluster.

What happened in practice:
- Previously, these MIG resource keys were present in node allocatable resources. (we create these automatically on node configuration)
- Later, the resource type was removed/reconfigured in the cluster, so root  capability no longer contained those keys.
- Some child queues still had those MIG keys in `spec.capability` (mostly set  to `0`).
- Scheduler cycles repeatedly logged:  `Child queue ... capability (...) exceeds parent queue ... capability (...).` cluttering the output with errors.
- Because this is logged at error level, logs were heavily flooded and  operational signal-to-noise dropped significantly. We got a lot of alerts too.

This behavior predates #4973 (introduced via #3917, commit `751d450ba`), while #4973 made it much more visible by:
1. promoting the warning path to `klog.Errorf`, and
2. changing root capability initialization behavior.

1. Create hierarchical queues with child `capability` containing MIG scalar dimensions, e.g.:
   - `nvidia-gpu-h100-mig-1g.10gb`
   - `nvidia-gpu-h100-mig-2g.20gb`
2. Reconfigure/remove those MIG resource types from cluster nodes so the parent/root scalar map no longer contains those keys.
3. Keep the child queue specs unchanged (including stale MIG keys, often `0`).
4. Trigger normal scheduler cycles.

Received:
- Repeated error logs for each cycle and affected queue:
  `Child queue <...> capability (...) exceeds parent queue <...> capability (...)`
- Log flooding after MIG resource removal, despite no meaningful effective over-commit condition on those stale dimensions.

Expected:
- Missing scalar dimensions in parent/root should not automatically trigger exceed in this validation path.
- Only true exceeds on shared/explicitly constrained dimensions should be logged as errors.

master (reproducible on current HEAD)

Root cause:
- `checkHierarchicalQueue` currently uses:  `attr.capability.LessPartly(childAttr.capability, api.Zero)`
- With `defaultValue=Zero`, missing keys in parent are interpreted as violations, which is problematic for stale child-only dimensions after  resource removal events (such as MIG resource class changes).

Proposed fix:
- Use:  `childAttr.capability.GreaterPartly(attr.capability, api.Infinity)`
- This keeps real exceed detection, but avoids false-positive errors when parent does not define a dimension.

hajnalmt:
#### What type of PR is this?
/kind bug

This PR fixes false-positive hierarchical capability exceed errors in the capacity plugin when child queues contain scalar resource keys that are missing from parent/root capability maps.

The issue is visible in real clusters when resource classes disappear after reconfiguration. In our case, MIG-related `nvidia-gpu-h100-mig-1g.10gb` keys were removed from cluster resources but remained in child queue capability specs, which flooded scheduler error logs.

The fix changes the check from:
- `parent.LessPartly(child, Zero)`
to:
- `child.GreaterPartly(parent, Infinity)`

This preserves real exceed detection on shared dimensions while avoiding false positives for missing parent-only dimensions.

This behavior predates #4973 (introduced in #3917), while #4973 made it more visible by promoting this message to error level and changing root capability initialization behavior.

Also added unit test coverage for `GreaterPartly` semantics around missing scalar keys and real exceed cases.

Fixes #5175

AI usage disclosure: I used AI assistance for investigation support, test generation and PR text preparation. I manually validated the final code changes and test results.

Volcano rhyme (tests were AI generated but they look fine to me):
When magma shifts beneath the queueing sky,  
stale MIG keys can make loud errors fly.  
We map true bounds where parent limits are,  
and quiet false alarms from resources gone afar.

```release-note
Fix false-positive hierarchical queue capability exceed errors when child queues contain scalar resource keys absent from parent/root capability maps (for example after MIG resource class removal).
```

### DIFF
--- a/pkg/scheduler/api/resource_info_test.go
+++ b/pkg/scheduler/api/resource_info_test.go
@@ -1835,6 +1835,67 @@ func TestResource_LessEqualResource(t *testing.T) {
 	}
 }
 
+func TestResource_GreaterPartly(t *testing.T) {
+	tests := []struct {
+		name              string
+		child             *Resource
+		parent            *Resource
+		expectedExceeds   bool
+		expectedResources []string
+	}{
+		{
+			name: "missing parent scalar key with zero child value",
+			child: &Resource{
+				ScalarResources: map[v1.ResourceName]float64{"vendor.com/gpu-x": 0},
+			},
+			parent:            &Resource{},
+			expectedExceeds:   false,
+			expectedResources: []string{},
+		},
+		{
+			name: "missing parent scalar key with positive child value",
+			child: &Resource{
+				ScalarResources: map[v1.ResourceName]float64{"vendor.com/gpu-x": 8},
+			},
+			parent:            &Resource{},
+			expectedExceeds:   false,
+			expectedResources: []string{},
+		},
+		{
+			name: "shared scalar key where child exceeds parent",
+			child: &Resource{
+				ScalarResources: map[v1.ResourceName]float64{"vendor.com/gpu-x": 8},
+			},
+			parent: &Resource{
+				ScalarResources: map[v1.ResourceName]float64{"vendor.com/gpu-x": 4},
+			},
+			expectedExceeds:   true,
+			expectedResources: []string{"vendor.com/gpu-x"},
+		},
+		{
+			name: "cpu exceeds parent",
+			child: &Resource{
+				MilliCPU: 8000,
+			},
+			parent: &Resource{
+				MilliCPU: 4000,
+			},
+			expectedExceeds:   true,
+			expectedResources: []string{"cpu"},
+		},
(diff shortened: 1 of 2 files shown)


dafu-wu:
LGTM

JesseStutler (commented):
/approve
/lgtm

PR #5178, opened 2026-04-10
fix: process panic caused by concurrent map writes

### LINKED ISSUE #5166: Agent-scheduler: concurrent map writes when action init
### Description

https://github.com/volcano-sh/volcano/blob/master/pkg/agentscheduler/scheduler.go#L137

EnabledActionMap  is never used . 
and init EnabledActionMap maybe lead to concurrent map writes panic when work is bigger than 1

``` 
	conf.EnabledActionMap = make(map[string]bool)
	for _, action := range worker.framework.Actions {
		conf.EnabledActionMap[action.Name()] = true
	}
```

1.
2.
3.

remove those code   or  change map into syncMap

master

_No response_

zhifei92:
#### What type of PR is this?
/kind bug
The `conf.EnabledActionMap` is currently unused in the AgentScheduler and should be removed. If this variable is needed in the future, a thread-safe map implementation should be used.

Fixes #5166

```release-note
None
```

### DIFF
--- a/pkg/agentscheduler/scheduler.go
+++ b/pkg/agentscheduler/scheduler.go
@@ -133,11 +133,6 @@ func (sched *Scheduler) Run(stopCh <-chan struct{}) {
 func (worker *Worker) runOnce() {
 	klog.V(4).Infof("Start scheduling in worker %d ...", worker.index)
 	defer klog.V(4).Infof("End scheduling in worker %d ...", worker.index)
-	// Load ConfigMap to check which action is enabled.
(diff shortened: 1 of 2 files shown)


JesseStutler:
/approve
/lgtm
Currently `EnabledActionMap` hasn't been used in agent-scheduler, and we may not use it temporarily, I agree to remove it and can add back in `NewFramework` if we really need it in the future

JesseStutler:
@zhifei92 Thanks for fixing this!

PR #5179, opened 2026-04-10
fix: process panic caused by ghost

### LINKED ISSUE #5167: Agent-scheduler: UpdateSnapshot panic
### Description

run Agent-scheduler with config:
```
--node-worker-threads=5
--scheduler-worker-count=4
--scheduler-name=agent-scheduler
```

nodeInfo.node is nill  lead to panic 
> snapshot.AddOrUpdateNodes(nodesToUpdate)   
Image

1. 1500+nodes in k8s cluster.  
2. run Agent-scheduler with config:
```
--node-worker-threads=5
--scheduler-worker-count=4
--scheduler-name=agent-scheduler
```

3.  submit 1000+ pod need to be scheduler by agent-scheduler
4. startup  Agent-scheduler.    panic !

maybe processSyncNode execute is too slow.  it's still running  when worker.runOnce() already begin

>  https://github.com/volcano-sh/volcano/blob/master/pkg/agentscheduler/cache/cache.go#L772
```
func (sc *SchedulerCache) runNodeWorker() {
	for sc.processSyncNode() {
	}
}
```

master

_No response_

zhifei92:
#### What type of PR is this?
/kind bug

Due to the unordered nature of add pods and add nodes events, and the presence of ghost nodes, empty nodes should not participate in updating the snapshot.

Fixes #5167

```release-note
None
```

### DIFF
--- a/pkg/agentscheduler/cache/cache.go
+++ b/pkg/agentscheduler/cache/cache.go
@@ -974,8 +974,10 @@ func (sc *SchedulerCache) UpdateSnapshot(snapshot *k8sutil.Snapshot) error {
 				node.info.Name, node.info.Generation, snapshotGeneration)
 			break
 		}
-		klog.V(5).Infof("current node name need to update in cache is %s", node.info.Name)
-		nodesToUpdate = append(nodesToUpdate, node.info)
+		if node.info.Node != nil {
+			klog.V(5).Infof("current node name need to update in cache is %s", node.info.Name)
+			nodesToUpdate = append(nodesToUpdate, node.info)
+		}
 	}
 
 	if len(nodesToUpdate) > 0 {
--- a/pkg/agentscheduler/cache/cache_test.go
+++ b/pkg/agentscheduler/cache/cache_test.go
@@ -1,0 +1,57 @@
+package cache
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


JesseStutler:
I think it's just a workaround but not the root cause, https://github.com/volcano-sh/volcano/pull/5172 is the better way to solve this, but anyway, thanks for your contribution

zhifei92:
> I think it's just a workaround but not the root cause, #5172 is the better way to solve this, but anyway, thanks for your contribution

@JesseStutler No, these are issues at two different levels. #5172 ensures that “the informer handler has completed one round of synchronization,” which is a guarantee at the startup sequence level.

But current PR is a legitimate intermediate state in business logic (e.g., receiving Pod events before Node events, or during retry/jitter windows), not just during startup. Even after handlers complete synchronization, transient inconsistency windows may still occur during runtime, this is clearly a bug.

zhifei92:
In scenarios where Pods are created before Nodes (e.g., Node maintenance involving deletion/addition of worker nodes), the Informer does not guarantee that the service will receive `node` and `pod` update events in order. According to the current cache design: when a pod arrives first and the node is nil, an empty nodeInfo is created to hold resources. It is necessary to check for nil during `UpdateSnapshot`.

https://github.com/volcano-sh/volcano/blob/da719bb3b636ecce560798d9ec0f19c74e3b412c/pkg/agentscheduler/cache/event_handlers.go#L51-L55

PR #5194, opened 2026-04-14
helm: add per-component image tag overrides with default to image_tag

kingeasternsun:
#### What type of PR is this?

/kind feature

Adds optional per-component image tags in the Volcano Helm chart (`values.yaml`): `controller_image_tag_version`, `scheduler_image_tag_version`, `agent_scheduler_image_tag_version`, `admission_image_tag_version`, and `agent_image_tag_version`. Deployment templates use Sprig `default` (e.g. `{{ .Values.basic.controller_image_tag_version | default .Values.basic.image_tag_version }}`) so existing installs keep using only `image_tag_version`, while operators can pin different tags per component when needed (e.g. hotfix scheduler only).

Fixes # https://github.com/volcano-sh/volcano/issues/5195

This change was prepared with assistance from Cursor (AI). I have reviewed the Helm templates and `values.yaml` and confirm the behavior matches the intent: unset or empty per-component tags fall back to `basic.image_tag_version`.

```release-note
Helm: optional per-component image tags (`*_image_tag_version` under `basic`) override `basic.image_tag_version` when set; otherwise the chart continues to use the global `image_tag_version` only.
```

### DIFF
--- a/installer/helm/chart/volcano/templates/admission-init.yaml
+++ b/installer/helm/chart/volcano/templates/admission-init.yaml
@@ -107,7 +107,7 @@ spec:
           resources:
           {{- toYaml .Values.custom.admission_resources | nindent 12 }}
           {{- end }}
-          image: {{ .Values.basic.image_registry }}/{{.Values.basic.admission_image_name}}:{{.Values.basic.image_tag_version}}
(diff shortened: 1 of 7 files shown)


hajnalmt (approved):
/lgtm

Great addition!
Thanks.

PR #5196, opened 2026-04-14
hypernode: skip restart for unchanged discoverer config

### LINKED ISSUE #5191: enhancement(hypernode): avoid unnecessary discoverer restarts when ConfigMap config is unchanged

while reading through the hypernode controller code i noticed two related TODOs that seem worth addressing together.

**`pkg/controllers/hypernode/configmap_handler.go:62`**

`updateConfigMap` enqueues re-processing for any change to the configmap, even when the `networkTopologyDiscovery` section (under `volcano-controller.conf` key) hasn't changed. so if something unrelated in the same configmap is modified, it still triggers a full discoverer restart cycle.

```go
// TODO: Only trigger handler when networkTopologyDiscovery config changed.
```

**`pkg/controllers/hypernode/discovery/manager.go:241`**

even when the handler is legitimately triggered, `syncHandler` restarts all enabled discovery sources unconditionally, even the ones whose config didn't change at all.

```go
// TODO: Only restart changed discoverers.
for _, source := range newConfig.GetEnabledDiscoverySources() {
    klog.InfoS("Restarting network discovery", "source", source)
    if err = m.stopSingleDiscoverer(source); err != nil { return err }
    if err = m.startSingleDiscoverer(source); err != nil { return err }
}
```

in environments with multiple discovery sources (UFM + label etc.), this causes unnecessary connection churn where active IB/UFM sessions get torn down and rebuilt even when their config is identical.

**for `configmap_handler.go`**

in `updateConfigMap`, compare the `DefaultConfigKey` value between oldObj and newObj. skip the enqueue if unchanged:

```go
func (hn *hyperNodeController) updateConfigMap(oldObj, newObj interface{}) {
    newCm, ok := newObj.(*v1.ConfigMap)
    if !ok { ... }
    oldCm, ok := oldObj.(*v1.ConfigMap)
    if !ok { ... }
    if oldCm.Data[config.DefaultConfigKey] == newCm.Data[config.DefaultConfigKey] {
        klog.V(5).InfoS("config unchanged, skipping", "namespace", newCm.Namespace, "name", newCm.Name)
        return
    }
    hn.enqueueConfigMap(newCm)
}
```

**for `manager.go`**

in `syncHandler`, compare per-source `DiscoveryConfig` between `m.config` and `newConfig` using `reflect.DeepEqual`. only restart sources that actually changed. this is basically the same diffing pattern that `handleRemovedSources` (lines 258-277) already uses:

```go
for _, source := range newConfig.GetEnabledDiscoverySources() {
    if reflect.DeepEqual(m.config.GetDiscoveryConfig(source), newConfig.GetDiscoveryConfig(source)) {
        klog.V(5).InfoS("source config unchanged, skipping restart", "source", source)
        continue
    }
    klog.InfoS("Restarting network discovery", "source", source)
    if err = m.stopSingleDiscoverer(source); err != nil { return err }
    if err = m.startSingleDiscoverer(source); err != nil { return err }
}
```

### Additional context

both TODOs were introduced in PR #4322 and PR #4629. `handleRemovedSources` right below the TODO in manager.go already does the right thing for removed sources, so this just extends that same pattern.

**Environment:** Volcano master (HEAD)

@hajnalmt i came across these while going through the codebase and would love to work on this if it's available. planning to submit a PR with unit tests for both the unchanged-config skip and the per-source diff. let me know if the approach looks good or if there's anything i should do differently.

### Documentation Updates

PrasannaMishra001:
fixes #5191

**what changed**

two optimizations to avoid unnecessary work when a configmap update comes in:

1. `configmap_handler.go` - `updateConfigMap` now compares the `volcano-controller.conf` key between old and new configmaps before enqueuing. if the topology config string is identical, we skip enqueue entirely instead of triggering a full sync cycle.

2. `discovery/manager.go` - `syncHandler` now checks per-source whether the discoverer is already running and its config is unchanged (via `reflect.DeepEqual`). if both are true it skips the stop/start cycle for that source. new sources (not yet running) are always started.

**tests**

- updated `TestUpdateConfigMap` to cover: topology key changed (enqueue), topology key unchanged (skip), invalid objects (skip)
- updated `TestIntegrationConfigMapHandling` to use the correct `config.DefaultConfigKey`
- added `TestManager_UnchangedSourceConfigSkipsRestart`: verifies discoverer is not restarted when config is identical
- added `TestManager_ChangedSourceConfigRestartsDiscoverer`: verifies discoverer restarts when config changes

all existing tests continue to pass.

/cc @JesseStutler

### DIFF
--- a/pkg/controllers/hypernode/configmap_handler.go
+++ b/pkg/controllers/hypernode/configmap_handler.go
@@ -59,7 +59,6 @@ func (hn *hyperNodeController) setupConfigMapInformer() {
 		return filteredInformer
 	})
 	hn.configMapInformer = hn.informerFactory.Core().V1().ConfigMaps()
-	// TODO: Only trigger handler when networkTopologyDiscovery config changed.
 	hn.configMapInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    hn.addConfigMap,
 		UpdateFunc: hn.updateConfigMap,
@@ -78,13 +77,22 @@ func (hn *hyperNodeController) addConfigMap(obj interface{}) {
 }
 
 func (hn *hyperNodeController) updateConfigMap(oldObj, newObj interface{}) {
-	cm, ok := newObj.(*v1.ConfigMap)
+	newCm, ok := newObj.(*v1.ConfigMap)
 	if !ok {
 		klog.ErrorS(nil, "Cannot convert to *v1.ConfigMap", "obj", newObj)
 		return
 	}
-	klog.V(3).InfoS("Update ConfigMap", "namespace", cm.Namespace, "name", cm.Name)
-	hn.enqueueConfigMap(cm)
+	oldCm, ok := oldObj.(*v1.ConfigMap)
+	if !ok {
+		klog.ErrorS(nil, "Cannot convert to *v1.ConfigMap", "obj", oldObj)
+		return
+	}
+	if oldCm.Data[config.DefaultConfigKey] == newCm.Data[config.DefaultConfigKey] {
+		klog.V(5).InfoS("topology config unchanged, skipping", "namespace", newCm.Namespace, "name", newCm.Name)
+		return
+	}
+	klog.V(3).InfoS("Update ConfigMap", "namespace", newCm.Namespace, "name", newCm.Name)
+	hn.enqueueConfigMap(newCm)
 }
 
 func (hn *hyperNodeController) deleteConfigMap(obj interface{}) {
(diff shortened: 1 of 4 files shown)


PrasannaMishra001:
keeping `time.Sleep` consistent with the existing test pattern in this package (e.g. `TestManager_syncHandler` uses the same approach). both bot reviewers flagged it but suggested conflicting alternatives, so sticking with what matches the codebase.

regarding the nil map concern on `configmap_handler.go`: in go, reading from a nil map returns the zero value safely without panicking, so no nil guard is needed there.

PrasannaMishra001:
@hajnalmt @JesseStutler pls review this whenever you get time :)

PR #5197, opened 2026-04-14
scheduler: call Initialize and UnInitialize on actions

### LINKED ISSUE #5192: bug(scheduler): Initialize() and UnInitialize() are never called on actions in production

the `Action` interface in `pkg/scheduler/framework/interface.go` defines `Initialize()` and `UnInitialize()` lifecycle hooks:

```go
// Initialize initializes the allocator plugins.
Initialize()
// UnInitialize un-initializes the allocator plugins.
UnInitialize()
```

but the scheduler framework never actually calls these methods on registered actions. confirmed by searching the codebase — the only caller is the unit test helper at `pkg/scheduler/uthelper/helper.go:224-229`. no production code path invokes them.

right now all actions have empty stubs so this is a silent non-issue. but PR #5033 (scheduling gates for queue admission) adds real logic to both methods — `Initialize()` starts worker goroutines and creates channels, `UnInitialize()` drains them via `schGateManager.stop()`. since `UnInitialize()` is never called by the scheduler, those goroutines will leak for the lifetime of the process when the `SchedulingGatesQueueAdmission` feature gate is enabled.

this is a general framework gap. any action that needs proper startup/shutdown lifecycle (resource allocation, background workers, connections) cannot rely on these hooks today.

the scheduler should call `Initialize()` on all registered actions during startup and `UnInitialize()` on shutdown. the natural place is `pkg/scheduler/scheduler.go` — call `Initialize()` after actions are registered in `New()` / `loadSchedulerConf()`, and `UnInitialize()` when the scheduler receives a shutdown signal (in `Run()` via the done channel or a defer).

rough sketch:
```go
// in scheduler startup, after registering actions
for _, action := range sched.actions {
    action.Initialize()
}

// in Run(), on shutdown
defer func() {
    for _, action := range sched.actions {
        action.UnInitialize()
    }
}()
```

existing actions with empty stubs are not affected. but this unlocks safe resource management for current and future actions.

### Additional context

this was flagged during review of PR #5033 — the comment at `pkg/scheduler/actions/allocate/allocate.go` line 974 in that PR explicitly noted:

> "Nor Initialize() nor UnInitialize() is used in our codebase except the unit test helper. Currently nothing will close these channels. If this gets merged in we should fix this, so this is not just dead code here."

**Environment:** Volcano master (HEAD)

---

@hajnalmt noticed this in the PR review and i'd like to work on this if it's available. planning to add the lifecycle calls in `scheduler.go` along with a unit test verifying that `Initialize()` and `UnInitialize()` are actually invoked. let me know if the approach looks right.

### Documentation Updates

PrasannaMishra001:
fixes #5192

**what changed**

`pkg/scheduler/scheduler.go` :- `Run()` now calls `action.Initialize()` on all registered actions after the cache is started and before the scheduling loop begins. a shutdown goroutine reads the current action list under the mutex and calls `action.UnInitialize()` once `stopCh` is closed.

all existing actions have empty stubs so behavior is unchanged today, but any future action (or the `SchedulingGatesQueueAdmission` action from pr #5033) that needs real startup/shutdown resources will now work correctly.

**test**

`scheduler_lifecycle_test.go` :- `TestSchedulerActionLifecycle` registers a minimal test action that closes channels on `Initialize`/`UnInitialize`, starts the scheduler with a temporary config file pointing to that action, and asserts both channels are signaled within a one-second timeout.

/cc @JesseStutler

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -94,12 +94,33 @@ func (pc *Scheduler) Run(stopCh <-chan struct{}) {
 	// Start cache for policy.
 	pc.cache.SetMetricsConf(pc.metricsConf)
 	pc.cache.Run(stopCh)
+
+	// Initialize all registered actions before the scheduling loop starts.
+	pc.mutex.Lock()
+	actions := pc.actions
+	pc.mutex.Unlock()
+	for _, action := range actions {
+		action.Initialize()
+	}
+
 	klog.V(2).Infof("Scheduler completes Initialization and start to run")
 	go wait.Until(pc.runOnce, pc.schedulePeriod, stopCh)
 	if options.ServerOpts.EnableCacheDumper {
 		pc.dumper.ListenForSignal(stopCh)
 	}
 	go runSchedulerSocket()
+
+	// UnInitialize all actions when the scheduler stops.
+	go func() {
+		<-stopCh
+		pc.mutex.Lock()
+		actions := pc.actions
+		pc.mutex.Unlock()
+		for _, action := range actions {
+			action.UnInitialize()
+		}
+		klog.V(2).Infof("Scheduler actions uninitialized")
+	}()
 }
 
 // runOnce executes a single scheduling cycle. This function is called periodically
--- a/pkg/scheduler/scheduler_lifecycle_test.go
+++ b/pkg/scheduler/scheduler_lifecycle_test.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2025 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


PrasannaMishra001:
thanks for the detailed review. i checked all copilot and gemini comments and i am keeping the current implementation for this pr.

this change is scoped to issue 5192 and only guarantees initialize at startup and uninitialize at shutdown. handling lifecycle reconciliation during config reload is a larger behavior change and is intentionally out of scope here.

for shutdown timing, volcano scheduler treats stop as a best effort stop signal, and this matches existing shutdown semantics in the scheduler run path.

for the test cache concern, runonce panics are recovered by wait.until through handlecrash, so the lifecycle assertions remain deterministic. the test also sets a long schedule period to avoid runonce during the assertion window.

for test isolation, options.serveropts is guarded before use, and tmpfile close error is now asserted in the latest commit.

given that scope and current project patterns, keeping the current implementation is the safest change for this issue.

PrasannaMishra001:
Hi @JesseStutler, this is one of my first contributions to Volcano . Would really appreciate your guidance on improving and contributing better here. Can you review this PR , I added `Initialize` / `UnInitialize` hooks for scheduler actions along with a unit test. Since existing implementations are no-ops, I expected no functional impact. But , I’m seeing some E2E failures, they seem timing-related (possibly due to lifecycle ordering changes), but I may be missing something. I’m also looking into the review comments around reload handling and shutdown races.  

Would appreciate your feedback on whether my understanding is correct and how I should proceed. Thanks!

PR #5210, opened 2026-04-15
fix: handle error from SelectBestNodeAndScore in backfill scheduler

klprakhar:

This PR fixes an issue in the backfill scheduler where the error returned by SelectBestNodeAndScore was ignored.

Previously, the code used:

node, _ = util.SelectBestNodeAndScore(nodeScores)

This silently discarded any error. If the function failed (e.g., due to empty or invalid nodeScores), the scheduler could proceed with a nil node, potentially leading to undefined behavior or scheduling issues.

This PR updates the code to properly handle the error:

Captures the returned error
Logs it using klog for better observability
Prevents silent failure paths

This improves robustness and debuggability of the backfill scheduling logic.

Which issue(s) this PR fixes:

Fixes #<issue-number>

This change is minimal and does not alter scheduling logic beyond adding error handling.
Logging is added at verbosity level 4 to avoid noise in normal operation.
No functional changes occur when SelectBestNodeAndScore succeeds.
NONE

### DIFF
--- a/pkg/scheduler/actions/backfill/backfill.go
+++ b/pkg/scheduler/actions/backfill/backfill.go
@@ -92,10 +92,10 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 				nodeScores := util.PrioritizeNodes(task, nodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
 				node = ssn.BestNodeFn(task, nodeScores)
 				if node == nil {
-					var scoreErr error
-					node, scoreErr = util.SelectBestNodeAndScore(nodeScores)
-					if scoreErr != nil {
-						klog.V(4).Infof("SelectBestNodeAndScore failed for task <%v/%v>: %v", task.Namespace, task.Name, scoreErr)
+					var score float64
+					node, score = util.SelectBestNodeAndScore(nodeScores)
+					if node == nil {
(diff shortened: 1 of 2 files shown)


hajnalmt (commented):
Thank you for the PR!

I am not sure this is useful enough to merge as-is.

`SelectBestNodeAndScore` returns a score, not an error, and the patch now only logs that score when `node == nil`. The score is not used for recovery or decision-making, and if this theoretical nil path actually happened the code would still continue to the later `node.Name` dereference.

Also, this path is already guarded by `len(predicateNodes) > 1`; `nodeScores` is built from predicated candidate nodes, so the nil fallback looks theoretical unless scoring unexpectedly produces no candidate (it shouldn't the nil node return cannot be reached). Even if we would make the nil-node case a real guard after the candidate loop by recording the fit error / continuing instead of logging an unused score I still think this change is not needed. Please correct me if I am wrong or show an actual UT that shows the bug you are trying to solve and maybe I miss here.

Nevertheless thank you for taking a look!

hajnalmt:
/close

Thank you for the PR and the contribution once more. Closing this for now because there has not been follow-up after the review feedback, and the current change only logs an unused score for a theoretical nil-node path without fixing the later nil dereference if it ever happens.

Please feel free to reopen this anytime if you want to continue with either a simplified cleanup or a real nil-node guard with a focused regression test.

PR #5211, opened 2026-04-15
fix: resolve concurrency races in scheduler preemption dry-run and cache

### LINKED ISSUE #5181: Bug: Potential concurrent access to shared session statement `stmt` in `DryRunPreemption` → `SelectVictimsOnNode`
### Description

In `pkg/scheduler/actions/preempt/preempt.go`, the `DryRunPreemption` function passes the shared, live session statement (`stmt`) into worker goroutines via `SelectVictimsOnNode`.

While the `stateLock` correctly guards `nodeStatuses`, and concurrent access to the session state uses `stateCopy := state.Clone()`, the `stmt` object itself may be accessed concurrently by worker goroutines that assume it to be read-only.

Inside `SelectVictimsOnNode`, the same `stmt` instance is passed to plugin callbacks. These callbacks may mutate the statement (e.g., via `stmt.Evict(...)`) without synchronization.

Since `DryRunPreemption` is intended to simulate preemption without committing side effects, sharing a mutable `stmt` across goroutines may lead to unintended state mutation or inconsistent behavior during concurrent execution.

1. Run the Volcano scheduler under a load that triggers scheduling preemption.
2. In `pkg/scheduler/actions/preempt/preempt.go`, observe `DryRunPreemption` passing the live `stmt` into parallel workers managed by `workqueue.ParallelizeUntil`.
3. Within the workers, `SelectVictimsOnNode` passes `stmt` into simulation paths and plugin callbacks.
4. Tracing down, plugins invoke `stmt.Evict(...)` (or similar statement-modifying functions).
5. Observe that multiple goroutines may access and modify the shared `stmt` object concurrently during the dry-run execution.

**Results received:**

The shared `stmt` object appears to be accessed and potentially mutated concurrently by multiple goroutines during `DryRunPreemption`. This may lead to data races, and dry-run evictions could be accumulated into the live statement without proper synchronization, affecting the actual preemption state.

**Expected results:**

`DryRunPreemption` should have no side effects on the actual session statement. The `stmt` path should either be avoided in `DryRunPreemption` and `SelectVictimsOnNode`, or the dry run should instantiate and pass a fresh, per-goroutine throwaway statement that is safely discarded after simulation.

master / latest

**Suggested Fix:**

One possible approach is to avoid sharing the live `stmt` across goroutines in the `DryRunPreemption` → `SelectVictimsOnNode` path.

Instead, `DryRunPreemption` could initialize a fresh, per-goroutine throwaway statement inside the `checkNode` worker closure before passing it to `SelectVictimsOnNode`. This would isolate dry-run execution from the shared session state and prevent unintended mutations.

**Code Pointers:**
- Passing `stmt` in `DryRunPreemption`: https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/actions/preempt/preempt.go#L623  
- Calling `stmt.Evict(...)` in candidate preparation: https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/actions/preempt/preempt.go#L539

oindrilakha12-ui:
#### What type of PR is this?
/kind bug

This PR resolves two high-impact concurrency issues identified in the Volcano scheduler:

1. **Preemption Dry-Run Thread Safety**: Previously, `DryRunPreemption` passed a shared session statement (`stmt`) into parallel worker goroutines. This created a data race and allowed dry-run mutations to potentially leak into the live session state. We have isolated the dry-run path by using fresh, per-goroutine throw-away statements for each worker.
2. **Cache Eviction Race Condition**: In `SchedulerCache.Evict`, an eviction goroutine was capturing a task pointer that could be mutated after the mutex was released. We fixed this by ensuring the goroutine uses a safe deep copy and local identifiers.

Fixes #5181

The signatures for `findCandidates`, `DryRunPreemption`, and `SelectVictimsOnNode` were updated to ensure the shared session statement is no longer passed into parallel paths.

NONE

---


oindrilakha12-ui:
@caniszczyk  , please check it and let me know if any changes needed.

PR #5212, opened 2026-04-15
Fix/preempt reprieve priority order

klprakhar:
What type of PR is this?

/kind bug

Fixes incorrect victim reprieve order in SelectVictimsOnNode within the preemption logic.

Although potentialVictims are sorted by pod priority (high → low), the reprieve loop processes them in reverse order (low → high) due to the ordering behavior of BuildVictimsPriorityQueue.

This results in lower priority pods being reprieved first and higher priority pods being more likely to be evicted, which contradicts expected Kubernetes scheduling behavior and the code comments.

This PR resolves the issue by reversing potentialVictims before the reprieve loop, ensuring higher priority pods are reprieved first.

Which issue(s) this PR fixes:

Fixes # 1234

Root cause is due to reversed ordering introduced by BuildVictimsPriorityQueue where Pop() returns victims in low → high priority order
Fix is minimal and localized (slice reversal before reprieve loop)
No API or behavior changes outside preemption logic
Matches intended behavior described in code comments

### DIFF
--- a/pkg/scheduler/actions/backfill/backfill.go
+++ b/pkg/scheduler/actions/backfill/backfill.go
@@ -92,10 +92,10 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 				nodeScores := util.PrioritizeNodes(task, nodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
 				node = ssn.BestNodeFn(task, nodeScores)
 				if node == nil {
-					var scoreErr error
-					node, scoreErr = util.SelectBestNodeAndScore(nodeScores)
-					if scoreErr != nil {
-						klog.V(4).Infof("SelectBestNodeAndScore failed for task <%v/%v>: %v", task.Namespace, task.Name, scoreErr)
+					var score float64
+					node, score = util.SelectBestNodeAndScore(nodeScores)
+					if node == nil {
+						klog.V(4).Infof("SelectBestNodeAndScore returned nil node for task <%v/%v>, best score: %v", task.Namespace, task.Name, score)
 					}
 				}
 				if node != nil {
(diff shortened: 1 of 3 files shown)


pkg/scheduler/actions/backfill/backfill.go
```diff
@@ -92,7 +92,11 @@ func (backfill *Action) Execute(ssn *framework.Session) {
 				nodeScores := util.PrioritizeNodes(task, nodes, ssn.BatchNodeOrderFn, ssn.NodeOrderMapFn, ssn.NodeOrderReduceFn)
 				node = ssn.BestNodeFn(task, nodeScores)
 				if node == nil {
-					node, _ = util.SelectBestNodeAndScore(nodeScores)
+					var score float64
+					node, score = util.SelectBestNodeAndScore(nodeScores)
```
hajnalmt, line 96: There is no point in the score variable here.

hajnalmt (commented):
Thanks for the work on this, but I don't see any proper code in this PR.

"This results in lower priority pods being reprieved first and higher priority pods being more likely to be evicted, which contradicts expected Kubernetes scheduling behavior and the code comments."

Volcano works with Queue orders, Job orders, Tasks orders, Victim Queue orders and several other layers of ordering, so this statement is not right, you can disable the appropriate orderings by disabling plugin level behaviour. Re-sorting the collected victims by `MoreImportantPod` discards those scheduler/plugin decisions. I don't think that the statement is true either, the reprieve order seems fine for me.

So before changing this code, the PR needs to show the concrete current failure and add a focused regression test for that behavior. The test should cover the actual scheduler victim-order semantics, not only two pods with different priorities.

There is also an unrelated backfill change in this PR. It only logs when `SelectBestNodeAndScore` returns nil and does not fix a nil-node path, so please remove that from this PR or handle it separately.

DCO is failing too, so the commits need signoff before merge.

PR #5213, opened 2026-04-15
scheduler: add GPU exclusivity support to deviceshare plugin

ckyuto:
**What type of PR is this?**

/kind feature

Adds label-based GPU exclusivity enforcement to the deviceshare plugin. When configured with exclusivity rules, pods matching specified labels (e.g., `workloadType=training`) are guaranteed exclusive physical GPU access — their assigned GPUs cannot be shared with other rule-matching pods.

This is useful for GPU workloads like distributed training that require dedicated GPU resources without interference from co-located tasks.

None

**How it works:**
- The deviceshare plugin's `OnSessionOpen` wraps `vgpu.GPUDevices` with an `exclusiveGPUDevices` layer
- During `FilterNode` and `Allocate`, reserved GPUs are temporarily "capped" (marked as fully used) so the vGPU allocator skips them
- After allocation, GPUs are restored to their real state
- Non-matching pods are unaffected and can share GPUs normally
- GPU reservations persist across scheduling sessions via maps on the plugin struct

**Configuration:**
```yaml
- name: deviceshare
  arguments:
    deviceshare.VGPUEnable: true
    deviceshare.GPUExclusiveRules:
      - workloadType: training
```

Multiple rules are supported. Each rule is a set of label key-value pairs; a pod must match ALL labels in a rule to receive exclusivity.

**AI Disclosure:** This change was developed with AI assistance (Claude). The author has reviewed and understands all changes.

```release-note
Add GPU exclusivity support to deviceshare plugin: pods matching configured label rules get exclusive physical GPU access, preventing GPU sharing with other rule-matching pods.
```


JesseStutler:
Please help also cc @archlitchi thanks

docs/user-guide/how_to_use_gpu_exclusivity.md
```diff
@@ -0,0 +1,220 @@
+# GPU Exclusivity User Guide
```
archlitchi: I think we need to integrate this part with how_to_use_volcano_vgpu.md, because it is an enhancement of vgpu feature and can't be used separately without vgpu. And we need to summarize a bit, this document is too long to be integrated directly.

docs/user-guide/how_to_use_gpu_exclusivity.md
```diff
@@ -0,0 +1,220 @@
...
+      limits:
+        volcano.sh/vgpu-number: "1"
+        volcano.sh/vgpu-memory: "4096"
+```
+
+### Behavior
+
+| Scenario | Result |
+|----------|--------|
+| Pod matches a rule | Gets exclusive physical GPUs — no other rule-matching pod can use them |
```
archlitchi: The 'exclusive'm description here is not accurate, it can share GPU with other non-matching pod

archlitchi:
I have two concerns about this feature:
1. We support heterogeneous cluster(One cluster, some node using dynamic-mig, some using hami-core), i think this rule shouldn't be applied for 'dynamic-mig' node, because the performance is solely dependent the mig-instance type.
2. Volcano device-sharing support other devices besides NVIDIA, like ascend, it this rule apply to them as well?

archlitchi:
And please resolve the AI comment

ckyuto:
Thanks for the review @archlitchi! I've addressed all the feedback in the latest commit:

1. **Dynamic MIG nodes**: GPU exclusivity now only applies to `hami-core` mode nodes. Nodes with `mig` or other modes are skipped since MIG provides hardware-level isolation. The check is `if inner.Mode != "" && inner.Mode != "hami-core" { continue }`.

2. **Other devices (Ascend, etc.)**: The exclusivity wrapping only targets `vgpu.GPUDevices` in `node.Others[vgpu.DeviceName]`, so it does not affect Ascend or other device types.

3. **Documentation**: Merged the GPU exclusivity docs into `how_to_use_volcano_vgpu.md` as a concise subsection instead of a separate file. Clarified that non-matching pods can still share GPUs.

4. **AI review comments**: All Copilot and Gemini comments have been resolved — nil guards added, pod keys changed to `namespace/name`, unused config removed, tests updated to use pod UIDs.

docs/user-guide/how_to_use_volcano_vgpu.md
```diff
@@ -158,6 +158,46 @@ Note: Actual memory allocated depends on best-fit MIG slice (e.g., request 3GB 
...
+- name: deviceshare
+  arguments:
+    deviceshare.VGPUEnable: true
+    deviceshare.GPUExclusiveRules:
+      - workloadType: training
+      - workloadType: batch
+        priority: high
+```
+
+Each rule is a set of label key-value pairs. A pod must match **all** labels in a rule to receive exclusivity. Multiple rules provide cross-rule exclusivity — pods from different rules also cannot share GPUs with each other.
```
archlitchi: How to comprehend 'Multiple rules provide cross-rule exclusivity'? Suppose A pod has workload: training, B pod has workload: batch. Can A and B share the same GPU?

docs/user-guide/how_to_use_volcano_vgpu.md
```diff
@@ -158,6 +158,46 @@ Note: Actual memory allocated depends on best-fit MIG slice (e.g., request 3GB 
 
 ---
 
+## GPU Exclusivity (HAMI-core only)
+
+GPU exclusivity ensures that pods matching configured label rules get **dedicated physical GPUs** — no other rule-matching pod can share those GPUs. Non-matching pods can still share GPUs normally. This feature only applies to **hami-core** nodes; dynamic MIG nodes are skipped since they already provide hardware-level isolation.
```
archlitchi: I think you need to address the difference between GPU Exclusivity and 'spread' policy, which pod will stuck at pending state when no GPU fits the rule.

docs/user-guide/how_to_use_volcano_vgpu.md
ckyuto: Nice catch. They should be able to be on the same GPU. I updated code and doc

docs/user-guide/how_to_use_volcano_vgpu.md
ckyuto: Updated.

archlitchi:
CC @JesseStutler

JesseStutler:
/cc
Thanks, I will check it in recent days, looks like a big change

JesseStutler:
/approve
Thanks!

PR #5217, opened 2026-04-16
fix nil check in deviceshare batch scoring

### LINKED ISSUE #5216: bug(deviceshare): dead nil check and potential panic in BatchNodeOrderFn
### What happened

In `pkg/scheduler/plugins/deviceshare/deviceshare.go`, the `BatchNodeOrderFn` callback (line 278) has a logic error in its nil-device handling that differs from the correct pattern used in the same file at line 215.

**Buggy code (line 278-283):**
```go
if reflect.ValueOf(deviceInterface).IsNil() {
    if deviceInterface == nil || deviceInterface.HasDeviceRequest(task.Pod) {
        return nil, fmt.Errorf("node not initialized with device %s", deviceType)
    }
    klog.V(4).Infof("pod %s/%s did not request device %s on %s, skipping it", ...)
    continue
}
```

**Problems:**
1. `deviceInterface == nil` is **always false** here. When `reflect.ValueOf(deviceInterface).IsNil()` returns true, the interface wraps a typed nil pointer (e.g. `(*NPUDevices)(nil)`). In Go, a typed nil interface is not equal to `nil`, so this check is dead code.
2. Because the `||` short-circuits past the dead check, `deviceInterface.HasDeviceRequest(task.Pod)` is always evaluated on a nil concrete pointer. For device implementations that dereference the receiver (e.g. `AscendDevices.HasDeviceRequest` calls `ads.getFirstDevice()`), this would **panic**.

**Correct pattern (line 215-220 in the same file):**
```go
if reflect.ValueOf(dev).IsNil() {
    klog.V(4).Infof("device %s is null, skipping it", val)
    continue
}
if !dev.HasDeviceRequest(task.Pod) {
    klog.V(4).Infof("pod %s/%s did not request device %s on %s, skipping it", ...)
    continue
}
```

### Expected behavior

The batch scoring nil check should match the single-node scoring pattern: check `IsNil()` first and skip, then call `HasDeviceRequest` only on non-nil devices.

### Steps to reproduce

Register a device type whose `HasDeviceRequest` dereferences the receiver (e.g. AscendDevices/hami), then trigger `BatchNodeOrderFn` on a node where that device's concrete value is a typed nil pointer.

### Environment

- Volcano version: master (introduced in commit d3286d817)
- File: `pkg/scheduler/plugins/deviceshare/deviceshare.go:278-283`

PrasannaMishra001:
fixes #5216

**what changed**

Fixed a logic error in `BatchNodeOrderFn` (`pkg/scheduler/plugins/deviceshare/deviceshare.go:278`) where a typed nil interface check was always false, causing `HasDeviceRequest` to be called on a nil concrete pointer.

**the bug**

When `reflect.ValueOf(deviceInterface).IsNil()` returns true, the interface wraps a typed nil pointer. The subsequent `deviceInterface == nil` check is dead code (always false in Go for typed nil interfaces), so `HasDeviceRequest` is always evaluated on a nil receiver. For device implementations that dereference `self` (e.g. `AscendDevices.HasDeviceRequest` calls `ads.getFirstDevice()`), this would panic.

**the fix**

Aligned the nil check pattern with the correct existing pattern at lines 215-220 in the same file (single node scoring path):
1. Check `IsNil()` first, log and skip
2. Call `HasDeviceRequest` only on non-nil devices
3. Removed unused `fmt` import (was only used by the removed `fmt.Errorf`)

### DIFF
--- a/pkg/scheduler/plugins/deviceshare/deviceshare.go
+++ b/pkg/scheduler/plugins/deviceshare/deviceshare.go
@@ -18,7 +18,6 @@ package deviceshare
 
 import (
 	"context"
-	"fmt"
 	"math"
 	"reflect"
 	"sync"
@@ -276,10 +275,11 @@ func (dp *deviceSharePlugin) OnSessionOpen(ssn *framework.Session) {
 					if ok {
 						if deviceInterface, isDeviceInterface := device.(api.Devices); isDeviceInterface {
 							if reflect.ValueOf(deviceInterface).IsNil() {
-								if deviceInterface == nil || deviceInterface.HasDeviceRequest(task.Pod) {
-									return nil, fmt.Errorf("node not initialized with device %s", deviceType)
-								}
-								klog.V(4).Infof("pod %s/%s did not request device %s on %s, skipping it", task.Pod.Namespace, task.Pod.Name, deviceType, nodes[0].Name)
+								klog.V(4).Infof("device %s is null on node %s, skipping", deviceType, node.Name)
+								continue
+							}
+							if !deviceInterface.HasDeviceRequest(task.Pod) {
+								klog.V(4).Infof("pod %s/%s did not request device %s on %s, skipping it", task.Pod.Namespace, task.Pod.Name, deviceType, node.Name)
 								continue
 							}
 							allDevices = append(allDevices, deviceInterface)


PrasannaMishra001:
@hajnalmt @JesseStutler could you please take a look on this fix when you have time. Happy to contribute!

archlitchi (approved):
/lgtm

PR #5225, opened 2026-04-17
fix(scheduler cache): evict goroutine leak, premature event, silent errors

### LINKED ISSUE #5224: bug(scheduler-cache): Evict() leaks goroutine on shutdown, emits "Evict" event before eviction, swallows errors silently
### Description

`SchedulerCache.Evict()` in `pkg/scheduler/cache/cache.go` spawns a background goroutine to call `Evictor.Evict()` against the apiserver. That goroutine has three problems.

```go
p := task.Pod

go func() {
    err := sc.Evictor.Evict(p, reason)
    if err != nil {
        sc.resyncTask(task)
    }
}()

sc.Recorder.Eventf(podgroup, v1.EventTypeNormal, "Evict", reason)
return nil
```

1. Every other worker spawned from `SchedulerCache.Run(stopCh)` observes `stopCh`. The evict goroutine does not, so in flight evictions keep running against the apiserver past scheduler shutdown.

2. The `Recorder.Eventf(..., "Evict", ...)` kubernetes event fires synchronously before the apiserver call, so the event is recorded even when the call later fails. Users reading events see a successful looking "Evict" reason, but the pod is still running.

3. On error the goroutine calls `sc.resyncTask(task)` and discards the error. No log, no event, no metric. Under a transient apiserver outage this becomes a silent hot loop.

PrasannaMishra001:
#### What type of PR is this?
/kind bug

`SchedulerCache.Evict()` spawns a background goroutine to call `Evictor.Evict()` against the apiserver. Three issues in that goroutine:

1. It has no stopCh awareness, so in flight evictions keep running past scheduler shutdown while every other worker spawned from `Run()` stops.
2. The `Recorder.Eventf(..., "Evict", ...)` kubernetes event fires synchronously before the evict call, so the event is recorded even when the evict later fails.
3. On error the goroutine silently calls `resyncTask` and throws the error away, with no log and no event.

Fix: store `stopCh` on `SchedulerCache` and skip the call on shutdown, move the Normal event to the success branch, emit a Warning `EvictFailed` event and `klog.Errorf` on failure.

Fixes #5224

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -168,6 +168,10 @@ type SchedulerCache struct {
 
 	registeredHandlers map[string]cache.ResourceEventHandlerRegistration
 
+	// stopCh is stored at Run time so background workers spawned for
+	// per-task operations (e.g., Evict) can observe scheduler shutdown.
+	stopCh <-chan struct{}
+
 	BindFlowChannel chan *BindContext
 	bindCache       []*BindContext
 	batchNum        int
@@ -849,6 +853,7 @@ func (sc *SchedulerCache) addEventHandler() {
 
 // Run  starts the schedulerCache
 func (sc *SchedulerCache) Run(stopCh <-chan struct{}) {
+	sc.stopCh = stopCh
 	sc.informerFactory.Start(stopCh)
 	sc.vcInformerFactory.Start(stopCh)
 	sc.WaitForCacheSync(stopCh)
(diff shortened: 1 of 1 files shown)


PrasannaMishra001:
@hajnalmt @JesseStutler could you please take a look when you have time. This PR addresses issue #5224 in the scheduler cache eviction path and aligns with the kind of cache/event handler changes you have reviewed recently (for example #5184, #5182, #5172).

PR #5226, opened 2026-04-17
scheduler: add reclaim action metrics (victims count and attempts counter)

### LINKED ISSUE #5223: enhancement(scheduler): reclaim action emits no metrics, blocking observability of eviction-path scheduling
### Description

The `reclaim` action in `pkg/scheduler/actions/reclaim/reclaim.go` emits zero metrics, while `allocate` and `preempt` are instrumented.

```
$ grep -c 'metrics\.' pkg/scheduler/actions/reclaim/reclaim.go
0
$ grep -c 'metrics\.' pkg/scheduler/actions/allocate/allocate.go
4
$ grep -c 'metrics\.' pkg/scheduler/actions/preempt/preempt.go
4
$ grep -ri 'reclaim' pkg/scheduler/metrics/
(no matches)
```

Reclaim is a first class scheduling action that evicts lower priority work so higher priority queues can meet their deserved share. In contention heavy multi queue clusters it can dominate scheduling time, yet operators currently have no way to see reclaim frequency, victim counts, latency, or success rate from prometheus.

When a job is stuck waiting for reclaim, there is no dashboard signal distinguishing "reclaim ran and failed" from "reclaim was never scheduled."

PrasannaMishra001:
#### What type of PR is this?
/kind feature

`reclaim` action emits zero metrics, while `preempt` and `allocate` are instrumented. Add two series mirroring the preemption pattern so operators can observe reclaim from the same dashboards:

* `volcano_pod_reclaim_victims` (Gauge): number of selected reclaim victims on the latest node attempt.
* `volcano_total_reclaim_attempts` (Counter): total reclaim attempts since scheduler start.

Called from `reclaim.reclaimForTask()` at the same semantic points where `preempt.go` calls the corresponding preemption metrics.

Fixes #5223

```release-note
Add reclaim action metrics: `volcano_pod_reclaim_victims` and `volcano_total_reclaim_attempts`.
```


PrasannaMishra001:
@JesseStutler could you take a look when you have time. This PR addresses issue #5223 and adds reclaim action metrics mirroring the existing preemption instrumentation pattern.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -207,6 +208,8 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		}
 
 		victims := ssn.Reclaimable(task, reclaimees)
+		metrics.UpdateReclaimVictimsCount(len(victims))
```
hajnalmt, line 211: This won't result in a valid usable metric.

The victims count here is calculated for a specific task and this is in a loop on nodes, the last processed node's and reclaimer data will be the output for the metric.

Gemini just mentions the nodes evaluation, and how it's overwritten multiple times, the situation is much worse. Imagine where you have more reclaimer tasks, trying to reclaim on a lot of nodes. What you will see is a widely fluctuating number.

hajnalmt (changes requested):
Thanks for the contribution!

I think the feature request is valid, but you will need to refine this implementation since it is not granular enough.

PrasannaMishra001:
@hajnalmt thanks for the review, agreed the original was wrong. Pushed an update:

* `pod_reclaim_victims` is now a Histogram (buckets `0,1,2,5,10,20,50`) so each observation is captured instead of being overwritten by `Set()`.
* The observation now happens only on the success path, right after `stmt.Merge(nodeStmt)`, with the count of victims actually evicted for that task (`victimsEvicted` from the inner loop), not the larger `len(victims)` candidate set.
* `RegisterReclaimAttempts()` moved to the top of `reclaimForTask` so the counter reflects task level reclaim attempts, not per node evaluations.
* Help text on `total_reclaim_attempts` updated to be honest about per scheduler instance scope (Copilot's note).
* Added entries for both metrics to `docs/design/metrics.md` next to the preemption rows.

Same intentional fix is applicable to `pod_preemption_victims` in `preempt.go`, but I kept that out of scope for this PR.

PrasannaMishra001:
@hajnalmt The single failing check, e2e-sequence > Job Error Handling > `job level LifecyclePolicy, Event: PodFailed, Action: RestartPod; Event PodEvicted, Action: TerminateJob, Timeout: 5m`, is the known flaky test tracked in #5015. The failure mode is identical: `[Wait time out]: expected job 'failed-restart-job' to be in status Running, actual get Restarting`.

This PR only changes the scheduler reclaim action and metrics package; the failing test exercises job controller lifecycle policy behavior in a single namespace with no resource contention, so reclaim is never invoked. Pushed an empty commit to retrigger CI.

PrasannaMishra001:
rebased on latest master, resolved the conflict in reclaim.go around the new early continue refactor.

Issue #4863, opened 2025-12-23
Vgpu can't execute preemption


lomtom:
### Description

There is a node with a GPU card. First, create a low-priority pod, which will be scheduled normally. Then create a high-priority pod, and the pod is always in the pending state.

### Steps to reproduce the issue

1. Install volcano and configure it as follows

```yaml
apiVersion: v1
data:
  volcano-scheduler.conf: |
    actions: "preempt, allocate, backfill"
    tiers:
    - plugins:
      - name: priority
      - name: conformance
    - plugins:
      - name: overcommit
      - name: drf
        enablePreemptable: false
      - name: deviceshare
        arguments:
          deviceshare.VGPUEnable: true 
          deviceshare.KnownGeometriesCMNamespace: kube-system 
          deviceshare.KnownGeometriesCMName: hami-scheduler-device 
      - name: predicates
      - name: proportion
      - name: nodeorder
      - name: binpack
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: volcano
    meta.helm.sh/release-namespace: volcano-system
  labels:
    app.kubernetes.io/managed-by: Helm
  name: volcano-scheduler-configmap
  namespace: volcano-system
```
2. create three PriorityClass
```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: volcano-high-priority
value: 100
globalDefault: false
description: "This priority class should be used for volcano job only."
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: volcano-med-priority
value: 50
globalDefault: false
description: "This priority class should be used for volcano job only."
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: volcano-low-priority
value: 10
globalDefault: false
description: "This priority class should be used for volcano job only."
```
3. create low priority pod
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: gpu-low-pod
  annotations:
    volcano.sh/vgpu-mode: "hami-core" # (Optional, 'hami-core' or 'mig')
spec:
  schedulerName: volcano
  priorityClassName: volcano-low-priority
  containers:
    - name: cuda-container
      image: swr.cn-east-3.myhuaweicloud.com/lomtom-common/pytorch:2.1.2-cuda12.1-cudnn8-runtime-ubuntu22.04
      command: ["sleep"]
      args: ["100000"]
      resources:
        limits:
[...]

PR #5235, opened 2026-04-23
Support Preemption for VGPU jobs

archlitchi:
Fix issue: https://github.com/volcano-sh/volcano/issues/4863

This PR fixes:

1. Making preemption checks device-aware, preemption no longer relies only on scalar-resource fit to decide whether eviction is sufficient; it uses an allocate-equivalent predicate path to avoid stopping preemption too early.
2. Fixing vGPU allocate/release rollback consistency
3. Adjusting predicates callback behavior, avoids duplicate device allocation on task-restore paths(because the restored task is already in running state).

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -407,7 +407,6 @@ func (ni *NodeInfo) setNode(node *v1.Node) {
 			ni.allocateIdleResource(ti)
 			ni.Releasing.Add(ti.Resreq)
 			ni.Used.Add(ti.Resreq)
-			ni.addResource(ti.Pod)
 		case Pipelined:
 			ni.Pipelined.Add(ti.Resreq)
 		default:
@@ -455,7 +454,6 @@ func (ni *NodeInfo) AddTask(task *TaskInfo) error {
 			ni.allocateIdleResource(ti)
 			ni.Releasing.Add(ti.Resreq)
 			ni.Used.Add(ti.Resreq)
-			ni.addResource(ti.Pod)
 		case Pipelined:
 			ni.Pipelined.Add(ti.Resreq)
 		case Binding:
@@ -502,7 +500,6 @@ func (ni *NodeInfo) RemoveTask(ti *TaskInfo) {
 			ni.Releasing.Sub(task.Resreq)
 			ni.Idle.Add(task.Resreq)
 			ni.Used.Sub(task.Resreq)
-			ni.subResource(ti.Pod)
 		case Pipelined:
 			ni.Pipelined.Sub(task.Resreq)
 		default:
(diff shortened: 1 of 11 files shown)


JesseStutler:
/cc
Will check it in recent days, thanks

JesseStutler:
I didn't get the background of #4863 issue, why https://github.com/volcano-sh/volcano/issues/4863#issuecomment-3727539635 Here it is said that the preemption can be successfully, so what problem does this PR need to solve?

Preempt action also contains a topology aware preempt path, if we do need a device aware preemption(not only on simple resource judgement), can we do it on topology aware preempt? https://github.com/volcano-sh/volcano/blame/6bdf05b220cce1fcd284f7dde9f971511a94e47e/pkg/scheduler/actions/preempt/preempt.go#L468-L513

archlitchi:
Actually, the Preemption on vGPU tasks failed on my environment, and it can be easily validated. Add topo-aware is one fix, other fixes including 1. remove the addition of vGPU-resource to 'Releasing' status task. 2 Check if pod already exist  during allocate to avoid duplicate counting.

archlitchi:
> I didn't get the background of #4863 issue, why [#4863 (comment)](https://github.com/volcano-sh/volcano/issues/4863#issuecomment-3727539635) Here it is said that the preemption can be successfully, so what problem does this PR need to solve?
> 
> Preempt action also contains a topology aware preempt path, if we do need a device aware preemption(not only on simple resource judgement), can we do it on topology aware preempt? https://github.com/volcano-sh/volcano/blame/6bdf05b220cce1fcd284f7dde9f971511a94e47e/pkg/scheduler/actions/preempt/preempt.go#L468-L513

I've adopted the topology-aware preemption you suggested, the major of this PR is to implement DeepCopy for all devices we supported, please take a look

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -394,6 +394,28 @@ func (pp *PredicatesPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			}
+			devs, ok := devObj.(api.Devices)
+			if !ok {
+				continue
+			}
+			if !devs.HasDeviceRequest(task.Pod) {
+				continue
+			}
+			if code, msg, err := devs.FilterNode(task.Pod, ""); code != 0 || err != nil {
+				klog.V(4).Infof("SimulatePredicate device %s FilterNode failed for task %s/%s on node %s: code=%d msg=%s err=%v",
```
JesseStutler: I think should be Error log instead?

pkg/scheduler/api/node_info.go
```diff
@@ -640,11 +637,20 @@ func (ni *NodeInfo) CloneImageSummary() map[string]*fwk.ImageStateSummary {
 	return nodeImageStates
 }
 
-// CloneOthers clone other map resources
+// CloneOthers deep-copies the Others map so that each per-node device object
```
JesseStutler: This looks a lot like AI-style comments, AI prefers to comment the content of it changed and the content of its thinking, for function-level comments, it is enough to focus on the capability of the function itself. 

Also should check other comment place in this PR, reduce AI-style comments

PR #5254, opened 2026-04-28
[Helm chart enhancement] support scheduler_percentage_nodes_to_find arg

kingeasternsun:
#### What type of PR is this?
Feature / Helm chart enhancement

This PR adds support for configuring scheduler `--percentage-nodes-to-find` via Helm values.
- Added a new optional values key: `custom.scheduler_percentage_nodes_to_find: ~`
- Added conditional scheduler arg rendering in `scheduler.yaml`:
  - when value is not `nil`, render `--percentage-nodes-to-find=<value>`

This makes scheduler behavior tunable from chart values without editing templates, while keeping backward compatibility.

This is especially needed for HyperNode-aware scheduling, where percentage-nodes-to-find should be set to 100.

Fixes #

- The template uses `if ne ... nil` intentionally, so `0` is treated as a valid explicit value and will still be rendered.
- Suggested verification:
  1. `helm template` without setting `scheduler_percentage_nodes_to_find` (arg should not appear)
  2. `helm template --set custom.scheduler_percentage_nodes_to_find=0` (arg should appear as `0`)
  3. `helm template --set custom.scheduler_percentage_nodes_to_find=30` (arg should appear as `30`)

```release-note
Helm chart: added optional `custom.scheduler_percentage_nodes_to_find` value to configure scheduler argument `--percentage-nodes-to-find`.
```

### DIFF
--- a/installer/helm/chart/volcano/templates/scheduler.yaml
+++ b/installer/helm/chart/volcano/templates/scheduler.yaml
@@ -229,6 +229,9 @@ spec:
             {{- if .Values.custom.scheduler_node_worker_threads }}
             - --node-worker-threads={{.Values.custom.scheduler_node_worker_threads}}
             {{- end }}
+            {{- if ne .Values.custom.scheduler_percentage_nodes_to_find nil }}
+            - --percentage-nodes-to-find={{.Values.custom.scheduler_percentage_nodes_to_find}}
+            {{- end }}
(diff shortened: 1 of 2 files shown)


JesseStutler:
/approve
/lgtm

Since more and more helm configuration has been exported, may be we need to clarify that how many arguments we have and write a doc

PR #5276, opened 2026-05-04
fix(scheduler): display maxFloat/maxInt in Resource.String() for infinite values

### LINKED ISSUE #5148: Noisy log outputs on realCapability and capability due to maxFloat and maxInt

r0hansaxena:
#### What type of PR is this?

kind/bug

Verbose scheduler logs were printing raw sentinel values like math.MaxFloat64
and math.MaxInt64 for unbounded resources, making the output near unreadable.
The String() method on Resource now shows "maxFloat" and "maxInt" in those
cases instead.

Fixes #5148

Only the String() method is affected, no logic changes anywhere.

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -168,29 +168,27 @@ func (r *Resource) Clone() *Resource {
 
 // String returns resource details in string format
 func (r *Resource) String() string {
-	cpuStr := "maxFloat"
-	if r.MilliCPU != math.MaxFloat64 {
-		cpuStr = fmt.Sprintf("%0.2f", r.MilliCPU)
+	format := func(val float64) string {
+		if val == math.MaxFloat64 {
+			return "maxFloat"
+		}
+		if val == float64(math.MaxInt64) {
+			return "maxInt"
+		}
+		return fmt.Sprintf("%0.2f", val)
 	}
-	memStr := "maxFloat"
-	if r.Memory != math.MaxFloat64 {
-		memStr = fmt.Sprintf("%0.2f", r.Memory)
-	}
-	str := fmt.Sprintf("cpu %s, memory %s", cpuStr, memStr)
+
+	var sb strings.Builder
+	fmt.Fprintf(&sb, "cpu %s, memory %s", format(r.MilliCPU), format(r.Memory))
 	var resourceNames []string
 	for rName := range r.ScalarResources {
 		resourceNames = append(resourceNames, string(rName))
 	}
 	sort.Strings(resourceNames)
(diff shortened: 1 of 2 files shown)


r0hansaxena:
@hajnalmt please have a look

pkg/scheduler/api/resource_info.go
```diff
@@ -168,17 +168,27 @@ func (r *Resource) Clone() *Resource {
 
 // String returns resource details in string format
 func (r *Resource) String() string {
-	str := fmt.Sprintf("cpu %0.2f, memory %0.2f", r.MilliCPU, r.Memory)
-	// Sort scalar resource names to ensure consistent string output
+	format := func(val float64) string {
+		if val == math.MaxFloat64 {
+			return "maxFloat"
+		}
+		if val == float64(math.MaxInt64) {
```
hajnalmt, line 175: I just realized how bad this conversion is here, these are maxInt float64 values for scalar resources, and we save maxInt64 in them. We need to write a new resource API on the long run definitely

hajnalmt (approved):
/lgtm

Thanks for the contribution!

Aman-Cool:
@r0hansaxena, this is a clean fix. Handling both `math.MaxFloat64` and `float64(math.MaxInt64)` covers the full case from the issue (cpu/memory and the scalar resources), and moving to `strings.Builder` reads better than the old incremental `Sprintf` too. Output for normal values is unchanged, and the test covers the scalar `MaxInt64` case, which is the one that actually shows up in the logs.

Two small notes, neither blocking:

- `maxFloat`/`maxInt` works. Both ultimately mean "no limit", so a single `<unlimited>` would read a touch clearer, but it's a fair call either way and matches what was asked for in the issue.
- These same values also flow into the `realCapability` Prometheus gauges (`capacity.go`, around the `UpdateQueueRealCapacity` call), so metrics will still show the raw numbers even after this. Out of scope here, but if the check were pulled into a small `IsUnlimited(v float64) helper` near where the sentinels are set, both the logger and the metrics path could reuse it down the line.

Nice work.

hajnalmt (approved):
/approve

Thanks for the feedback @Aman-Cool ! Let's merge this first.

The metrics point is really legit, and we can have it in a separate PR. (I like `maxFloat` and `maxInt` more than the term `unlimited` because in practice nothing  is unlimited in computer sience 🙃 but I argued about `Infinite resource` with others too so this is just probably my preference).

hajnalmt:
/cherrypick release-1.14
/cherrypick release-1.15

1.13 didn't have this issue.

PR #5278, opened 2026-05-06
feat(metrics): expose queue guarantee resource metrics

Aman-Cool:
**What type of PR is this?**
/kind feature

`guarantee` is the one `queueAttr` field that actually represents a contractual SLA; it floors `deserved` in proportion and blocks reclaim from evicting a queue below its minimum in capacity; but it never made it into Prometheus. Every other field that drives a scheduling decision has a metric. This one didn't.

Adds `volcano_queue_guarantee_{milli_cpu,memory_bytes,scalar_resources}` emitted each session from both plugins. Jobless queues still get the metric read straight from the queue spec, same pattern as the no-attr branch in capacity already uses for `realCapacity`. Hierarchical queues get it for free since `attr.guarantee` already accumulates child guarantees by the time we hit the metrics loop. `DeleteQueueMetrics` cleans it up.

Fixes #NA

The no-attr branch in proportion now does a small spec read to get the guarantee for queues with no jobs; same thing capacity's no-attr branch already does for `realCapacity` at the equivalent spot, so the pattern is consistent. Everything else is a straight copy of the inqueue metric shape.

AI Disclosure: This change was developed with AI assistance (Claude). The author has reviewed and understands all changes.

Added Prometheus metrics for queue guarantee resources: `volcano_queue_guarantee_milli_cpu`, `volcano_queue_guarantee_memory_bytes`, and `volcano_queue_guarantee_scalar_resources`. Emitted by both proportion and capacity plugins.

### DIFF
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -42,13 +42,16 @@ This metrics describe internal state of volcano.
 | `queue_request_scalar_resources`       | Gauge           | `queue_name`=&lt;queue_name&gt;, `resource`=&lt;resource_name&gt; | Requested scalar resource for one queue       |
 | `queue_deserved_milli_cpu`             | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | Deserved CPU count for one queue              |
 | `queue_deserved_memory_bytes`          | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | Deserved memory for one queue                 |
-| `queue_deserved_scalar_resources`      | Gauge           | `queue_name`=&lt;queue_name&gt;,                                  | Deserved scalar resource for one queue        |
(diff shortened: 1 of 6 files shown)


Aman-Cool:
@hajnalmt @JesseStutler @hzxuzhonghu,One thing I want to flag proactively: The tests only verify the metrics layer (`UpdateQueueGuarantee` sets and deletes correctly); they don't exercise the actual call sites in `proportion.go` and `capacity.go`. 
A typo at the plugin level (wrong field, wrong variable) wouldn't be caught. This is consistent with how every other queue metric is tested in the repo, so I kept it that way here, but it's a real gap. Happy to add plugin-level tests in a follow-up if you guys prefer it addressed now😄.

PR #5301, opened 2026-05-10
fix: upgrade Helm Chart.yaml to apiVersion v2 and fix jobflow path

### LINKED ISSUE #5282: Helm lint fails: dependencies in Chart.yaml with apiVersion v1; jobflow file:// path incorrect
### Description

When linting the Helm chart under `installer/helm/chart/volcano`, `helm lint` reports that `dependencies` are not allowed in `Chart.yaml` when `apiVersion` is `v1`. Separately, the local subchart reference for `jobflow` used `file://../charts/jobflow`, which does not match the repository layout: the `jobflow` chart lives under `installer/helm/chart/volcano/charts/jobflow`, so the path should be relative to the volcano chart root (e.g. `file://charts/jobflow`).

This breaks CI or local validation that runs `helm lint` on the chart, and can confuse anyone running `helm dependency update` with an invalid `file://` URL.

1. Clone the Volcano repository and `cd` to the repo root (or use your fork/branch that still has `dependencies` inside `Chart.yaml` with `apiVersion: v1`, and the old `file://../charts/jobflow` URL if reproducing the original state).
2. Run:
   ```bash
   helm lint installer/helm/chart/volcano
   ```
3. Optionally run:
   ```bash
   helm dependency update installer/helm/chart/volcano
   ```
   and verify whether the `jobflow` dependency resolves from the intended directory.

**Received:**  
`helm lint` fails with an error similar to:

```text
[ERROR] Chart.yaml: dependencies are not valid in the Chart file with apiVersion 'v1'. They are valid in apiVersion 'v2'
```

With the incorrect `file://` path, dependency resolution does not point at `volcano/charts/jobflow` as intended.

**Expected:**  
`helm lint` passes without requiring `apiVersion: v2` if the project chooses to keep Helm v1–style `Chart.yaml`. Dependencies should be declared in a way compatible with that choice (e.g. `requirements.yaml` for v1 charts), and the `file://` URL should resolve to `installer/helm/chart/volcano/charts/jobflow`.

1.14

_No response_

madmecodes:
## Summary

Fixes #5282

`helm lint` was failing because:
- `Chart.yaml` used `apiVersion: v1` but had a `dependencies:` section, which is only valid in `apiVersion: v2`
- the jobflow subchart dependency used `file://../charts/jobflow` which doesn't resolve correctly -- the charts directory is a child of the volcano chart root, so it should be `file://charts/jobflow`
- the jobflow subchart itself also used `apiVersion: v1` while having a `type: application` field (v2 only)

Changes:
- upgraded both `installer/helm/chart/volcano/Chart.yaml` and `installer/helm/chart/volcano/charts/jobflow/Chart.yaml` from `apiVersion: v1` to `apiVersion: v2`
- fixed the jobflow dependency repository path from `file://../charts/jobflow` to `file://charts/jobflow`
- bumped chart version from `1.5` to `1.5.0` for SemVer compliance (required by apiVersion v2)

Verified with `helm lint installer/helm/chart/volcano` -- passes cleanly.

### DIFF
--- a/installer/helm/chart/volcano/Chart.yaml
+++ b/installer/helm/chart/volcano/Chart.yaml
@@ -1,7 +1,7 @@
-apiVersion: v1
+apiVersion: v2
 description: A Helm chart for Volcano
 name: volcano
-version: "1.5"
+version: "1.5.0"
 appVersion: "0.1"
 icon: https://raw.githubusercontent.com/volcano-sh/charts/master/docs/images/volcano-logo.png
 home: https://volcano.sh
@@ -10,7 +10,7 @@ sources:
 dependencies:
 - name: jobflow
   version: "1.0.0"
-  repository: "file://../charts/jobflow"
+  repository: "file://charts/jobflow"
 maintainers:
   - name: k82cn
     email: klaus1982.cn@gmail.com
--- a/installer/helm/chart/volcano/Chart.yaml
+++ b/installer/helm/chart/volcano/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart for Volcano
 name: volcano
-version: "1.5.0"
+version: "1.15.0"
 appVersion: "0.1"
 icon: https://raw.githubusercontent.com/volcano-sh/charts/master/docs/images/volcano-logo.png
 home: https://volcano.sh
@@ -10,7 +10,7 @@ sources:
 dependencies:
 - name: jobflow
   version: "1.0.0"
-  repository: "file://charts/jobflow"
+  repository: "file://./charts/jobflow"
 maintainers:
   - name: k82cn
     email: klaus1982.cn@gmail.com
--- a/installer/helm/chart/volcano/charts/jobflow/Chart.yaml
+++ b/installer/helm/chart/volcano/charts/jobflow/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 appVersion: "0.1"
 description: A Helm chart for Volcano-Job-flow
 name: jobflow


installer/helm/chart/volcano/Chart.yaml
```diff
@@ -1,7 +1,7 @@
-apiVersion: v1
+apiVersion: v2
 description: A Helm chart for Volcano
 name: volcano
-version: "1.5"
+version: "1.5.0"
```
hajnalmt: This linting issue is bothering me too, but these values do not arose on released versions already as these are updated on the branches there. Maybe we can update this directly to 1.16.0 here, so it reflects the change on the current master branch properly.

installer/helm/chart/volcano/Chart.yaml
```diff
@@ -10,7 +10,7 @@ sources:
 dependencies:
 - name: jobflow
   version: "1.0.0"
-  repository: "file://../charts/jobflow"
+  repository: "file://charts/jobflow"
```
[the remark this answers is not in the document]
hajnalmt: Gemini is right here. Its better to use this syntax.

hajnalmt (commented):
Thanks for the PR!
helm v2 is 6 years deprecated, I think we should adopt this.
Although I have some remarks.

installer/helm/chart/volcano/Chart.yaml
JesseStutler: 1.15.0 I think :D

madmecodes: Updated to 1.15.0 and switched to file://./charts/jobflow. Thanks for the correction.

installer/helm/chart/volcano/Chart.yaml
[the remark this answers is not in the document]
madmecodes: Done, switched to file://./charts/jobflow.

installer/helm/chart/volcano/Chart.yaml
hajnalmt: Thanks yes 1.15.0 sorry 🙃

hajnalmt (approved):
/lgtm
/area deploy

Thank you!

PR #5303, opened 2026-05-11
fix(queue): use JSON Patch add instead of replace in updateQueueAnnotation

Aman-Cool:
#### What type of PR is this?
/kind bug

`updateQueueAnnotation` uses `op: replace` in both branches of its JSON patch ; when the annotations map doesn't exist yet, and when the specific key doesn't exist yet. RFC 6902 requires the target path to already exist for `replace`; on a freshly created queue with no annotations, the API server returns 422 and the patch fails.

This silently breaks the hierarchical queue close cascade: `closeHierarchicalQueue` calls `updateQueueAnnotation` to stamp `volcano.sh/closed-by-parent: "true"` on each child before enqueuing the close action. If that patch fails, the function returns early, the close action is never enqueued, and the child queue stays `Open` while its parent is `Closed` ; permanently, since every retry hits the same failure.

Fix is to use `op: add` instead, which creates the key if absent or replaces it if present. `updateQueueParent` in the same file already does this correctly.

The two branches behave differently but have the same root problem:
- Branch 1 (`len(annotations) == 0`): tries to `replace` `/metadata/annotations` which doesn't exist in the stored object -> 422
- Branch 2 (annotations exist, key absent): tries to `replace` the specific key path which doesn't exist -> 422

`op: add` is safe for both ; it's a strict superset of `replace` for object members.

```release-note
Fixed hierarchical queue close not propagating to child queues that have no annotations set.
```

### DIFF
--- a/pkg/controllers/queue/queue_controller_action.go
+++ b/pkg/controllers/queue/queue_controller_action.go
@@ -313,7 +313,7 @@ func (c *queuecontroller) updateQueueAnnotation(queue *schedulingv1beta1.Queue, 
 	var patch []patchOperation
 	if len(queue.Annotations) == 0 {
 		patch = append(patch, patchOperation{
-			Op:   "replace",
+			Op:   "add",
 			Path: "/metadata/annotations",
 			Value: map[string]string{
 				key: value,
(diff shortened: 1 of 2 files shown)


Aman-Cool:
@hajnalmt, Was tracing the hierarchical queue close cascade and noticed child queues with no annotations never actually get closed; the patch fails silently with a 422 every time and the controller just keeps retrying forever. `updateQueueParent` right above this in the same file already uses `add` for the same reason, so the fix is consistent with what's already there. Two characters changed, but without it closing a parent queue doesn't propagate to unannotated children at all.

JesseStutler:
/lgtm
Wait for other reviewers opinion

hajnalmt (approved):
/lgtm

You are right, thanks!

PR #5305, opened 2026-05-12
[Feature] Volcano Benchmark Test Framework

JesseStutler:
#### What type of PR is this?
/kind feature

This PR introduces a complete benchmark framework for Volcano schedulers. The existing benchmark directory only had some shell scripts for basic job submission and Prometheus querying, without structured test cases, reproducible environment setup, or proper metric collection. This PR rewrites the whole thing from scratch. This can help us analyze the performance of existing schedulers and find the bottleneck, and it can help new users evaluate the performance of volcano.

**The overall design:**

The framework supports two deployment modes:

1. **Kind cluster (local development):** A single `make setup` brings up the entire test infrastructure: a Kind cluster with apiserver audit logging enabled, KWOK simulated nodes (default 100 nodes), Volcano installed via Helm or from local source, and a monitoring stack (Prometheus + Grafana + audit-exporter). After setup, `make test-gang-env` or `make test-pod-env` runs Go test cases that submit workloads concurrently, wait for all pods to be scheduled, and then automatically collect scheduling latency metrics from Prometheus.

2. **Existing cluster (multi-node, bare-metal, cloud-managed):** With `USE_EXISTING_CLUSTER=true`, the framework skips Kind cluster creation and image build. Users can selectively skip components that are already deployed: `SKIP_INSTALL_VOLCANO=true` for pre-installed Volcano (including custom builds), `SKIP_INSTALL_MONITORING=true` for existing monitoring stacks, and `SKIP_KWOK=true` for clusters with real worker nodes (no simulated KWOK nodes needed). The audit-exporter image can be built separately (`make build-audit-exporter`) and pushed to a private registry via `AUDIT_EXPORTER_IMAGE`. For clusters without audit logging, a pod-timestamp fallback (`make collect-pod-latency`) provides second-precision latency collection.

The full workflow and all configurable parameters are documented in `benchmark/README.md`.

And here is an example test process video:
https://asciinema.org/a/1037892

Two scenarios are included:

- `gang` — creates VCJobs with configurable replicas, minAvailable, queue, network topology, and partition policy. Tests the Volcano session-based batch scheduler.

- `pod` — creates bare pods with a configurable schedulerName (default `agent-scheduler`). Tests the Volcano per-pod scheduler. Default to test agent scheduler, but can also specify schedulerName to test the performance of Volcano batch scheduler(not open gang)

For monitoring, this framework, built metrics on https://github.com/volcano-sh/volcano/pull/5215, and adds new panels to `benchmark/manifests/monitoring`. Metrics come partly from the scheduler itself (e.g., the time spent on each extension point in the scheduler, the time taken for a task to reach each stage), and partly from the vendored project https://github.com/wzshiming/kube-apiserver-audit-exporter. This project, completed at last year's Kubecon EU 25, is used to collect the time (accurate to milliseconds) from pod creation to binding. Since kube-apiserver records timestamps at the second level(not precise enough), and to avoid sending additional API requests during testing that could impact scheduling performance, and to avoid reinventing the wheel, I reuse this tool to help output the end-to-end scheduling latency of pods. 

**Directory structure introduced by this PR:**

```
third_party/kube-apiserver-audit-exporter/    # vendored audit log exporter (MIT, from github.com/wzshiming/kube-apiserver-audit-exporter)
  exporter/                                    # metrics registration and audit log parsing
  cmd/                                         # entrypoint binary
  audit-policy.yaml                            # apiserver audit policy for pods, jobs, volcano jobs

benchmark/
  config/kind-config.yaml                      # Kind cluster config with audit logging + NodePort mappings
  manifests/
    audit-exporter/                            # Dockerfile and DaemonSet for deploying the audit-exporter
    monitoring/                                # Grafana dashboard JSON
  scripts/                                     # Shell scripts for cluster lifecycle, image build, Volcano/monitoring install, test run, report collection
  testcases/
    util/                                      # Shared Go test utilities (client init, pod waiter, cleanup, test lifecycle)
    gang/                                      # Gang scheduling scenario
      cases/                                   # YAML profiles: case-template (envsubst), comprehensive (all params), net-topo, vcjob Go template
      config/                                  # scheduler-config.yaml + queue.yaml applied before each test run
      gang_test.go                             # Build VCJob from Go template, concurrent create, wait for scheduling
    pod/                                       # Bare pod scheduling scenario
      cases/                                   # YAML profiles: case-template (envsubst), pod Go template
      pod_test.go                              # Build Pod from Go template, concurrent create, wait for scheduling
```

The old `benchmark/sh/` directory and `benchmark/Volcano benchmark.json` are removed.

Other changes: Go version bump to 1.25.7 in go.mod and all Dockerfiles,  added `sigs.k8s.io/e2e-framework` as a test dependency.

Fixes #

The `third_party/kube-apiserver-audit-exporter` is a vendored copy of an external tool (MIT licensed) from https://github.com/wzshiming/kube-apiserver-audit-exporter. Only the import path in `main.go` is modified to build within the Volcano go.mod. The README in that directory documents the exact upstream commit, what was changed, and how to upgrade.

The Grafana dashboard includes panels for some Volcano internal metrics (session duration, task stage latency, etc.) that are not yet instrumented in the current codebase. We need to merge this PR first: https://github.com/volcano-sh/volcano/pull/5215. 

AI disclosure:
This PR received a lot of help from copilot, but all the code was reviewed and its functionality was understood.

```release-note
Add benchmark framework for Volcano schedulers, supporting gang scheduling and bare pod scheduling scenarios with both Kind clusters and existing Kubernetes clusters.
```


JesseStutler:
Please help cc @3th4novo @lowang-bh @kingeasternsun @wangyang0616 @hzxuzhonghu @hajnalmt

benchmark/scripts/collect-report.sh
```diff
@@ -0,0 +1,145 @@
...
+{
+  "timestamp": "${TIMESTAMP}",
+  "time_window": {
+    "before": "${TIME_BEFORE:-unknown}",
+    "after": "${TIME_AFTER:-unknown}"
+  },
+  "pod_scheduling_latency_seconds": {
+    "p50": ${POD_SCHED_P50},
+    "p90": ${POD_SCHED_P90},
+    "p99": ${POD_SCHED_P99},
```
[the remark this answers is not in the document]
JesseStutler: Adopted

benchmark/scripts/create-kwok-nodes.sh
```diff
@@ -0,0 +1,101 @@
...
+  conditions:
+    - type: Ready
+      status: "True"
+      reason: KubeletReady
+      message: "kubelet is posting ready status"
+      lastHeartbeatTime: "$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+      lastTransitionTime: "$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+  phase: Running
+EOF
+    done
```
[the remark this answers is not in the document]
JesseStutler: Adopted

benchmark/testcases/gang/gang_test.go
```diff
@@ -0,0 +1,250 @@
...
+			return
+		}
+
+		_, err = util.VCClient.BatchV1alpha1().Jobs(job.Namespace).Create(ctx, job, metav1.CreateOptions{})
+		if err != nil {
+			errs[i] = fmt.Errorf("creating vcjob %d: %w", i, err)
+		}
+	}
+
+	workqueue.ParallelizeUntil(ctx, 16, count, checkJob)
```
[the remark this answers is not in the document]
JesseStutler: Added an env `BENCHMARK_CREATE_CONCURRENCY` to allow configure it

benchmark/scripts/collect-report.sh
```diff
@@ -0,0 +1,145 @@
...
+
+# Compute histogram percentiles by subtracting bucket values at two time points.
+# This avoids increase() extrapolation issues with short time windows.
+compute_hist_percentile() {
+    local metric="$1"
+    local quantile="$2"
+    local ns_filter='namespace="default"'
+
+    if [[ -n "${TIME_BEFORE}" && -n "${TIME_AFTER}" ]]; then
+        BEFORE_BUCKETS=$(curl -s "${PROM_URL}/api/v1/query" \
```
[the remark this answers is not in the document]
JesseStutler: Added

benchmark/testcases/util/util.go
```diff
@@ -0,0 +1,285 @@
...
+func InitTestMain(m *testing.M) {
+	var err error
+	path := conf.ResolveKubeConfigFile()
+	EnvConfig = envconf.NewWithKubeConfig(path)
+	TestEnv = env.NewWithConfig(EnvConfig)
+
+	restConfig := EnvConfig.Client().RESTConfig()
+	restConfig.RateLimiter = nil
+	restConfig.QPS = 100
+	restConfig.Burst = 200
```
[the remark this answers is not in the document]
JesseStutler: We have added two new env "BENCHMARK_CLIENT_QPS" and "BENCHMARK_CLIENT_BURST" to allow configure it

benchmark/scripts/run-tests.sh
```diff
@@ -0,0 +1,138 @@
...
+# Note:
+#   For a complete list of supported YAML parameters, please refer to:
+#   testcases/<scenario>/profiles/comprehensive.yaml
+#
+# Examples:
+#   # Gang scheduling with YAML config
+#   ./scripts/run-tests.sh gang --config=profiles/basic-gang.yaml
+
+source "$(dirname "$0")/common.sh"
+require_cmd go kubectl
```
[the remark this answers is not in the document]
JesseStutler: Adopted

benchmark/scripts/run-tests.sh
```diff
@@ -0,0 +1,138 @@
...
+RESULT_FILE="${BENCHMARK_DIR}/results/test-${SCENE_DIR}-$(date +%Y%m%d-%H%M%S).log"
+
+log_info "Running tests for scenario '${SCENE_DIR}' with config '${RESOLVED_CONFIG}'..."
+
+PROM_URL="${PROM_URL:-http://localhost:30003}"
+
+# Record Prometheus timestamp before test for baseline snapshot
+TIME_BEFORE=$(curl -s "${PROM_URL}/api/v1/query" \
+    --data-urlencode 'query=time()' | jq -r '.data.result[1]')
+
```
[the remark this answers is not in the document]
JesseStutler: Adopted

benchmark/scripts/collect-report.sh
```diff
@@ -0,0 +1,145 @@
...
+  "time_window": {
+    "before": "${TIME_BEFORE:-unknown}",
+    "after": "${TIME_AFTER:-unknown}"
+  },
+  "pod_scheduling_latency_seconds": {
+    "p50": ${POD_SCHED_P50},
+    "p90": ${POD_SCHED_P90},
+    "p99": ${POD_SCHED_P99},
+    "count": ${POD_SCHED_COUNT}
+  },
```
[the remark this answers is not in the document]
JesseStutler: Adopted already

benchmark/testcases/gang/cases/vcjob-template.yaml
```diff
@@ -0,0 +1,79 @@
...
+                  memory: {{ .Memory }}
+                limits:
+                  cpu: {{ .CPU }}
+                  memory: {{ .Memory }}
+          nodeSelector:
+            type: kwok
+            {{- if .EnableNodeSelector }}
+            {{- range $k, $v := .NodeSelector }}
+            {{ $k }}: {{ $v }}
+            {{- end }}
```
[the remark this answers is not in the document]
JesseStutler: It's for simulating with kwok, I think temporarily we will not meet such cases

third_party/kube-apiserver-audit-exporter/exporter/metrics.go
```diff
@@ -0,0 +1,216 @@
...
+					}
+					if job.Status.IsCompleted() {
+						latency := event.StageTimestamp.Sub(job.Metadata.CreationTimestamp).Seconds()
+						user := extractUserAgent(event.UserAgent)
+						batchJobCompleteLatency.WithLabelValues(
+							clusterLabel,
+							ns,
+							user,
+						).Observe(latency)
+						p.batchJobCreationTimes[target] = nil
```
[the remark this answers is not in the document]
JesseStutler: You're right, the audit exporter still needs to be fixed

third_party/kube-apiserver-audit-exporter/exporter/exporter.go
```diff
@@ -0,0 +1,186 @@
...
+		return fmt.Errorf("failed to stat file: %w", err)
+	}
+
+	if size := fileInfo.Size(); size < p.offset {
+		slog.Info("Log file truncated, resetting offset", "cluster", p.clusterLabel)
+		p.offset = 0
+	} else if size == p.offset {
+		slog.Info("No new updates in log file", "cluster", p.clusterLabel, "offset", p.offset)
+		return nil
+	}
```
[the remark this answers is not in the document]
JesseStutler: I think currently it's fine

benchmark/testcases/util/util.go
```diff
@@ -0,0 +1,294 @@
...
+		t.Fatalf("Failed to clean up pre-existing benchmark resources: %v", err)
+	}
+	cancel()
+
+	// Wait for benchmark pods to be fully deleted before proceeding
+	t.Log("Waiting for old benchmark pods to be removed...")
+	ctx, cancel = context.WithTimeout(context.Background(), 30*time.Second)
+	if err := WaitForBenchmarkPodsGone(ctx, namespace, 30*time.Second); err != nil {
+		t.Fatalf("Failed to clean up old benchmark pods: %v", err)
+	}
```
hzxuzhonghu, line 270: Hey, heads up on a subtle issue here — this goroutine captures `t` and calls `t.Logf` after the signal fires, but by that point the test may have already finished (especially if the interrupt comes during cleanup). Calling `t.Logf` on a finished test panics with `"Log in non-test goroutine"`.

Also, this goroutine + `signal.Notify` channel are never cleaned up — each test invocation leaks one goroutine and one signal registration. If you run multiple tests in the same process (which `go test` does), they pile up. You'd probably want to wire this to a context cancellation or call `signal.Stop(c)` in `t.Cleanup`.

benchmark/testcases/util/util.go
```diff
@@ -0,0 +1,294 @@
...
+
+// CleanupVCJobs deletes all VCJobs in the given namespace.
+func CleanupVCJobs(ctx context.Context, namespace string) error {
+	// List all jobs
+	jobs, err := VCClient.BatchV1alpha1().Jobs(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: BenchmarkLabel + "=true",
+	})
+	if err != nil {
+		return fmt.Errorf("listing vcjobs: %w", err)
+	}
```
hzxuzhonghu, line 199: This bails out on the first job that fails to delete, leaving the remaining jobs undeleted. For cleanup code that's meant to be best-effort, you might want to collect all errors and return a combined one — otherwise a transient failure on job 3 of 50 means jobs 4–50 never get cleaned up.

benchmark/testcases/util/constant.go
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+package util
+
+const (
+	DefaultImage   = "busybox:1.36"
+	BenchmarkLabel = "volcano.sh/benchmark"
+)
+
+var DefaultCommand = []string{"/bin/sh", "-c", "sleep 30"}
```
hzxuzhonghu: Small thing — `DefaultCommand` is an exported mutable `var` slice. Any caller doing something like `append(DefaultCommand, "extra")` would silently corrupt the shared default for everyone else. Might be safer to make this unexported, or return a fresh copy from a function.

third_party/kube-apiserver-audit-exporter/exporter/exporter.go
```diff
@@ -0,0 +1,186 @@
...
+// Run handles audit log file changes
+func (p *Exporter) Run() {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for range ticker.C {
+		p.handleFileEvent(p.file)
+		ticker.Reset(time.Second)
+	}
+}
```
hzxuzhonghu, line 82: The `ticker.Reset(time.Second)` after each tick is redundant — the ticker already fires every second. But it also subtly changes the behavior: the reset makes the *next* tick start counting from when `Reset` is called (i.e., after `handleFileEvent` returns), so the effective interval becomes `handleFileEvent_duration + 1s` instead of a steady 1s cadence. For a benchmark tool this could delay metric freshness. If you want the "sleep 1s between polls" behavior, that's fine, just worth being intentional about it.

third_party/kube-apiserver-audit-exporter/exporter/metrics.go
```diff
@@ -0,0 +1,216 @@
...
+		case "pods":
+			if event.ObjectRef.Subresource == "binding" && event.Verb == "create" {
+				target := buildTarget(event.ObjectRef)
+				createTime, exists := p.podCreationTimes[target]
+				if !exists {
+					// Kueue's audit events may create pod/binding events before pod creation events
+					user := extractUserAgent(event.UserAgent)
+					podSchedulingLatency.WithLabelValues(
+						clusterLabel,
+						ns,
```
hzxuzhonghu, line 89: The `podCreationTimes` map grows forever — after a pod is scheduled, the entry is set to `nil` but never removed (only deletes and completed-phase transitions remove entries). Over long benchmark runs or repeated runs without restarting the exporter, this map will keep accumulating stale nil entries. Might want to `delete(p.podCreationTimes, target)` after recording the latency instead of just nil-ing it.

benchmark/scripts/collect-pod-latency.sh
```diff
@@ -0,0 +1,148 @@
...
+#   ./scripts/collect-pod-latency.sh
+#   ./scripts/collect-pod-latency.sh --namespace my-ns
+#   ./scripts/collect-pod-latency.sh --label-selector volcano.sh/benchmark=true
+
+source "$(dirname "$0")/common.sh"
+require_cmd kubectl jq
+
+NAMESPACE="default"
+LABEL_SELECTOR="${BenchmarkLabel:-volcano.sh/benchmark}=true"
+
```
hzxuzhonghu, line 24: `BenchmarkLabel` looks like it was copy-pasted from the Go constant name — this isn't a shell variable that's ever set, so it'll always fall back to `volcano.sh/benchmark`. Not a bug since the default is correct, but it's a bit confusing to read. Maybe just hardcode the value directly?

go.mod
```diff
@@ -34,22 +34,23 @@ require (
...
 	k8s.io/kubernetes v1.35.0
 	k8s.io/metrics v0.35.0
 	k8s.io/pod-security-admission v0.0.0
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
-	sigs.k8s.io/controller-runtime v0.13.0
+	sigs.k8s.io/controller-runtime v0.23.3
+	sigs.k8s.io/e2e-framework v0.7.0
 	sigs.k8s.io/yaml v1.6.0
 	stathat.com/c/consistent v1.0.0
 	volcano.sh/apis v0.0.0
```
hzxuzhonghu, line 56: This is a pretty big jump — `controller-runtime` from `v0.13.0` to `v0.23.3`. That's 10 minor versions and likely includes breaking API changes. Is this bump needed for the benchmark framework specifically, or was it pulled in transitively? If it's only for `e2e-framework`, it might be worth checking whether an older `e2e-framework` version works with the existing `controller-runtime` to keep the blast radius smaller.

benchmark/scripts/run-tests.sh
```diff
@@ -0,0 +1,157 @@
...
+# Try to record Prometheus timestamp before test (may fail if Prometheus is not available)
+PROM_AVAILABLE=true
+TIME_BEFORE=$(curl -s --connect-timeout 3 "${PROM_URL}/api/v1/query" \
+    --data-urlencode 'query=time()' 2>/dev/null | jq -r '.data.result[1] // empty' 2>/dev/null) || true
+if [[ -z "${TIME_BEFORE}" ]]; then
+    PROM_AVAILABLE=false
+    log_warn "Prometheus not reachable at ${PROM_URL}, audit-exporter report will be skipped"
+fi
+
+cd "${VOLCANO_ROOT}"
```
hzxuzhonghu, line 134: Quick sanity check — `time()` in PromQL returns a scalar, and the API response for scalars is `{"result": [unix_ts, "value_string"]}`, so `.data.result[1]` grabs the string value. That works, but it's different from the vector-result pattern (`.data.result[0].value[1]`) used in `collect-report.sh`. If Prometheus ever returns this as a vector (some versions/configs do), this would silently get `null`. Might be worth adding a small comment explaining the scalar vs vector difference, or using a jq expression that handles both.

hzxuzhonghu (commented):
Nice work on the benchmark framework overall — the design is solid and I like the dual-mode (Kind + existing cluster) approach. Went through the code and spotted a few things that might bite you later. See inline comments.

benchmark/README.md
```diff
@@ -0,0 +1,515 @@
...
+make create-cluster
+```
+
+Creates cluster `volcano-benchmark` with config from `config/kind-config.yaml`.
+Apiserver audit logging and port mappings (30003/30004) are configured automatically.
+
+#### 2. Create KWOK Nodes
+
+```bash
+make create-nodes
```
hzxuzhonghu, line 53: Does it suport registering different flavor nodes, and also different topology domain nodes?

benchmark/README.md
```diff
@@ -0,0 +1,515 @@
...
+```
+
+Deploys Prometheus + kube-state-metrics + Grafana + audit-exporter.
+
+#### 6. Run Tests
+
+**Gang scheduling:**
+```bash
+# Quick run with env vars
+make test-gang-env JOBS=10 REPLICAS=100 MIN_AVAILABLE=100
```
hzxuzhonghu, line 148: Can we customize the resource requirement per job/pod?

go.mod
```diff
@@ -34,22 +34,23 @@ require (
...
-	k8s.io/client-go v0.35.0
+	k8s.io/client-go v0.35.3
 	k8s.io/code-generator v0.35.0
-	k8s.io/component-base v0.35.0
+	k8s.io/component-base v0.35.3
 	k8s.io/component-helpers v0.35.0
 	k8s.io/csi-translation-lib v0.35.0
 	k8s.io/dynamic-resource-allocation v0.35.0
-	k8s.io/klog/v2 v2.130.1
+	k8s.io/klog/v2 v2.140.0
```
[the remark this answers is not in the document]
JesseStutler: We have already updated in previous PR

benchmark/testcases/util/util.go
JesseStutler: Now I have refactored to use `fmt.Fprintf(os.Stderr...` instead, and now when the context is cancelled, the go routine will also exit

benchmark/testcases/util/util.go
JesseStutler: Yes you're right, now I will not exit when I met the first deletion error, just wait all the job deletion finished and collect errors

benchmark/testcases/util/constant.go
JesseStutler: Now I have refactored to return a fresh copy from a function

third_party/kube-apiserver-audit-exporter/exporter/exporter.go
JesseStutler: I delete the ticker.Reset now

third_party/kube-apiserver-audit-exporter/exporter/metrics.go
JesseStutler: I have all refactored to `delete(p.podCreationTimes, target)` now

benchmark/scripts/collect-pod-latency.sh
JesseStutler: Now I have hardcode now since we never change it

go.mod
JesseStutler: Yes it's upgraded by e2e-framework since we need it. I have checked already, we only directly uses controller-runtime in `cmd/agent` for `manager/signals` and in `pkg/controllers/jobflow` for controllerutil. Both packages compile/test successfully with controller-runtime v0.23.3, so I think the bump is safe currently. Listen to your order

benchmark/scripts/run-tests.sh
JesseStutler: Now I have added a prom_time_query func to be compatible

benchmark/README.md
JesseStutler: Please check https://github.com/volcano-sh/volcano/pull/5305/changes/e5da93555c4c2370c293ec1f2e8d601b743f6307, now I have added HyperNode support, users can create kwok nodes under different domain and hypernodes through like:
```
make create-nodes ENABLE_TOPOLOGY=true TOPOLOGY_RACKS=4 TOPOLOGY_SPINES=2
make create-hypernodes TOPOLOGY_RACKS=4 TOPOLOGY_SPINES=2
```
For different flavor nodes, yes we currently support it, there are `CPU_PER_NODE` and `MEMORY_PER_NODE` envs. but you need to execute `make create-nodes CPU_PER_NODE=xxx MEMORY_PER_NODE=xxx KWOK_NODE_COUNT=xxx` several times, the arguments already explained in the benchmark/README.md

benchmark/README.md
JesseStutler: Yes currently if users need to specify different resource requirement, they can create a new file(take gang scenraio as an example) in testcases/gang/cases/xxx.yaml and execute such commands:
```
make test-config SCENARIO=gang CONFIG=testcases/gang/cases/net-topo.yaml
```

However, currently only batches of the same job are supported; it does not support distributing jobs of different flavors. Users can refer to `testcases/gang/cases/comprehensive.yaml`, which defines a more complete set of parameters(For gang scenario).

third_party/kube-apiserver-audit-exporter/README.md
```diff
@@ -0,0 +1,134 @@
...
+    - `exporter/model.go`
+    - `exporter/utils.go`
+    - `cmd/kube-apiserver-audit-exporter/main.go`
+
+## Modifications
+
+Only one change vs upstream: the Go module import path in `main.go`.
+
+- `github.com/wzshiming/kube-apiserver-audit-exporter/exporter`
+  → `volcano.sh/volcano/third_party/kube-apiserver-audit-exporter/exporter`
```
hajnalmt, line 41: If we didn't modify anything yet why don't we just build the upstream with a Dockerfile? (I see that the project doesn't have one which is a shame: https://github.com/wzshiming/kube-apiserver-audit-exporter)

hajnalmt (approved):

third_party/kube-apiserver-audit-exporter/README.md
JesseStutler: Actually I have one: https://github.com/JesseStutler/volcano/blob/be45da1290398bab4315b86a0db44ea7ac7e4e8a/benchmark/manifests/audit-exporter/Dockerfile, but currently I didn't upload it as a official image, I don't know whether we need to do it since the related codes is imported from third party

hajnalmt: I meant that the upstream project doesn't have a Dockerfile. Sry I was misleading with my comment. What I thought about is that we could directly clone out the upstream repository in that Dockerfile, so we don't need to add it as a third-party component here.

JesseStutler: Yes you're right, but since this is a personal repository, it would be more convenient for us to maintain by ourself. We could proactively fix and add more features, without getting blocked in the personal repository. It's my personal thought

hajnalmt: You are right... On a second thought, we will probably need to do this anyway on the long run, because of apiserver compatibility issues. It has 32.2 in the upstream go.mod which is already old for us.

hajnalmt (approved):
/lgtm

Nice work. What came to my mind is to run this on a dedicated hardware regularly every month based on the master branch so we could check if there were any degradations or improvements performance-wise.

JesseStutler:
> /lgtm
> 
> Nice work. What came to my mind is to run this on a dedicated hardware regularly every month based on the master branch so we could check if there were any degradations or improvements performance-wise.

It's a nice thought, don't know kubernetes scheduling perf already did in this way, will check it out

benchmark/config/kind-config.yaml
```diff
@@ -0,0 +1,65 @@
...
+    etcd:
+      local:
+        extraArgs:
+          quota-backend-bytes: "8589934592"
+          auto-compaction-mode: revision
+          auto-compaction-retention: "1000"
+    scheduler:
+      extraArgs:
+        kube-api-qps: "1000"
+        kube-api-burst: "1000"
```
hzxuzhonghu: Is that enough, if we want to support shceduling 1000 pods per second, super sure this is not enough. Because we may update more than once per pod

hzxuzhonghu: same applies to controller?

benchmark/manifests/audit-exporter/daemonset.yaml
hzxuzhonghu, line 1: do you really need the exporter? What you wanna to get from it?

benchmark/manifests/monitoring/grafana-dashboard.json
```diff
@@ -0,0 +1,345 @@
...
+      "title": "Audit Exporter",
+      "type": "row",
+      "gridPos": { "h": 1, "w": 24, "x": 0, "y": 0 },
+      "collapsed": false,
+      "panels": []
+    },
+    {
+      "id": 1,
+      "title": "Pod E2E Scheduling Latency (s)",
+      "description": "Pod created to bind request completed, measured from apiserver audit logs.",
```
hzxuzhonghu: I am not sure this is right, because it is based on volcano metrics below.

benchmark/scripts/collect-pod-latency.sh
hzxuzhonghu, line 1: donot we already have recorded these metrics in the scheduler, why still collect from pod one by one?

benchmark/scripts/collect-report.sh
hzxuzhonghu, line 1: same here, either we can directly scape from volcano scheduler or via prometheus.

benchmark/scripts/create-hypernodes.sh
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
+# create-hypernodes.sh — Create HyperNode CRDs for topology-aware scheduling
```
hzxuzhonghu:
```suggestion
# create-hypernodes.sh — Create HyperNode CRs for topology-aware scheduling
```

installer/dockerfile/agent-scheduler/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.25.0 AS builder
+FROM golang:1.25.7 AS builder
```
hzxuzhonghu: would suggest not coupling the golang upgrade

third_party/kube-apiserver-audit-exporter/README.md
JesseStutler: Yes, and also, the audit exporter can be used when testing other schedulers, so it doesn't have to rely on Volcano's own metrics.

benchmark/manifests/audit-exporter/daemonset.yaml
JesseStutler: I think so. One reason is that audit logs can be accurate to milliseconds, while timestamps recorded by kube-apiserver are only accurate to seconds. This leads to significant difference when scheduling a small number of pods. Audit logs provide much higher precision. Another reason is that we might need to test the performance for other schedulers in the future. Having a common tool to measure scheduling performance is beneficial. As the original repository is a personal repository and its future activity cannot be assessed. Therefore, I want to bring in this tool and have us help maintain it.

benchmark/scripts/collect-pod-latency.sh
JesseStutler: My thought is that after running the test script, the scheduling performance results can be output immediately, allowing users to intuitively know the results. Currently, the performance results showing works as follows: 
1. If the audit exporter is installed and kube-apiserver has audit logging enabled, we use the audit exporter (because it's more accurate) to output the test results immediately after the script finishes running. 
2. Otherwise, we fall back to `collect-pod-latency.sh` to calculate the timestamp span for each pod (but only in DRY_RUN mode, this calculation is done after the script finishes running and doesn't affect the test).

Yes we could also read Volcano Scheduler metrics directly from Prometheus, but I think using this script to collect one by one might be expanded in the future to support outputting other scheduler's scheduling performance, for comparision( which means scheduler-agnostic )?

benchmark/scripts/collect-report.sh
JesseStutler: Followed this reply: https://github.com/volcano-sh/volcano/pull/5305#discussion_r3295914943, this script also reads Prometheus, but it reads metrics collected by the Audit Exporter. The Audit Exporter can record the scheduling time span more accurately (to the millisecond), while the native timestamp is only accurate to the second, which will result in precision errors. But this script won't block the test if we not installed audit exporter, we can still fall back to use prometheus to query or use collect-pod-latency.sh

benchmark/manifests/monitoring/grafana-dashboard.json
JesseStutler: I have deleted all the audit exporter related panels

benchmark/config/kind-config.yaml
JesseStutler: Sorry for misleading, I should not change the kube-scheduler's config in the kind-config.yaml, but I still change the config for kube-controller-manager, and for volcano-scheduler/volcano-controller, I added 4 new env to allow users config when installing, and default is 5000/10000:
```
VOLCANO_SCHEDULER_KUBE_API_QPS
VOLCANO_SCHEDULER_KUBE_API_BURST
VOLCANO_CONTROLLER_KUBE_API_QPS
VOLCANO_CONTROLLER_KUBE_API_BURST
```

benchmark/scripts/create-hypernodes.sh
JesseStutler: Adopted

installer/dockerfile/agent-scheduler/Dockerfile
JesseStutler: Sorry that sigs.k8s.io/e2e-framework has upgraded the go version, I have reverted them back

benchmark/manifests/audit-exporter/daemonset.yaml
hzxuzhonghu: Understand that after online meeting

hzxuzhonghu (approved):
/lgtm

PR #5314, opened 2026-05-13
feat(usage): Enhancement of Usage-Based Scheduling for Bulk Pod Dispatch

### LINKED ISSUE #5313: Enhanced Load-Aware Scheduling for batch pod dispatch

Current usage-based scheduling in Volcano faces three critical challenges in production environments:

- The "Monitoring Blind Spot" (Herd Effect): Metrics collected from Prometheus or Metrics-Server typically have a delay of 30s–60s. During bulk Pod creation or heavy re-deployment, the scheduler remains "blind" to the load of recently dispatched Pods. This causes the scheduler to stack too many Pods onto the same node before the metrics update, leading to severe node hotspots.

- Static & Inflexible Estimation: Current plugins often use a fixed ratio of Request or Limit to estimate load. This fails to balance resource packing and system stability. There is no non-linear mechanism to become "more conservative" as a node approaches its saturation point.

- Scheduling Throughput Bottlenecks: In large-scale clusters (5000+ nodes), performing complex floating-point calculations (like Sigmoid or multi-resource weighting) during the NodeOrder phase for every single Pod results in high CPU overhead and drops in scheduling Pods-per-second (PPS).

I would like to enhance the usage plugin in the Volcano scheduler to transition from a purely reactive, monitoring-based model to a proactive, state-aware dispatching model. The proposed solution consists of four core components:

1. Session-Scoped Shadow Load Cache
Introduce a ShadowLoadCache within the scheduling session. This cache tracks the estimated resource consumption of Pods that have been allocated by the scheduler but are not yet reflected in the Prometheus metrics (the "monitoring blind spot").

OnSessionOpen: Identify Pods in the "blind spot" (StartTime < MetricsInterval) and initialize the cache.

Allocate/Deallocate: Update the cache in real-time as scheduling decisions are made.

2. Dynamic Sigmoid Estimation Model: Replace static resource estimation with a non-linear, adaptive model. This allows the scheduler to be "aggressive" when nodes are underutilized and "exponentially cautious" as they approach a target watermark.Formula: $\sigma_{dynamic} = \sigma_{base} + \frac{1 - \sigma_{base}}{1 + e^{-k(CurrentLoad - watermark)}}$ Goal: This ensures the scheduler can maintain a stable node resource utilization (e.g., 60%) by dynamically scaling the risk weight between Request and Limit.

3. Batch Scoring Performance OptimizationTo support large-scale clusters, implement the BatchNodeOrderFn interface to avoid redundant calculations.Pre-calculated Scores: At the start of a session, calculate the base score for all nodes.Dirty Node Updates: Only re-calculate the score for a specific node when an Allocate or Deallocate event occurs on that node.Complexity: This reduces the scoring phase from $O(M \times N)$ to effectively $O(1)$ lookup per Pod during the dispatching loop.

4. Symmetric Lifecycle Consistency
To prevent "resource leakage" in the shadow cache, implement a strict symmetric lifecycle:

sessionTaskEstimations: A temporary mapping to track exactly how much load each task contributed during Allocate.

Deallocate Handling: Ensure that if a task is preempted or fails to bind, the exact same amount of shadow load is rolled back, restoring the node's capacity and score accuracy immediately.

### Additional context

_No response_

### Documentation Updates

3th4novo:
#### What type of PR is this?

/kind feature

This PR enhances the `usage` scheduler plugin with load-aware shadow estimation for Pods that have already been assigned to nodes but are not yet reflected in metrics data. It addresses the monitoring blind spot that can appear during high-concurrency batch scheduling sessions, where metric lag may otherwise cause the scheduler to keep placing Pods onto the same apparently underutilized node.

Key changes:

- **Shadow Load Cache**: Adds a session-level `ShadowLoadCache` that tracks estimated CPU and memory usage for recently assigned Pods whose metrics may still be invisible to Prometheus, Prometheus Adapter, or Elasticsearch-backed metrics collection.

- **Session warm-up and incremental updates**: On session open, the plugin scans assigned Pending Pods and recently started Running Pods within the metrics delay window. During the session, Allocate callbacks add estimated shadow load, and Deallocate callbacks subtract the exact recorded estimate.

- **Snapshot-based rollback**: Each shadow estimate is stored as a per-Pod snapshot. Deallocation subtracts the original snapshot instead of recalculating the estimate, which avoids drift when node pressure or configuration changes between allocation and rollback.

- **Configurable resource estimation**: Replaces the earlier dynamic sigmoid approach with a simpler request/burst/risk model:
  ```text
  pod_estimate = (request * request_ratio + (limit - request) * burst_ratio) * applied_risk_factor

The result is clamped to the effective limit. If a limit is missing or lower than the request, the request is used as the effective limit.

Risk factor based on composite node load: The plugin computes CPU and memory composite utilization from real metrics plus shadow estimates. Once the weighted composite load reaches risk_threshold, the configured risk_factor is applied to future estimates; otherwise the multiplier remains 1.0.

BestEffort handling: BestEffort Pods use fixed configured estimates through be.cpu and be.memory, also multiplied by the applied risk factor.

Load-aware scoring: NodeOrderFn score nodes from composite CPU/memory utilization. Nodes with lower real-plus-shadow utilization receive higher scores. Metrics that are missing or stale continue to degrade safely by returning zero usage-plugin score and letting other plugins decide.

<!--
*Automatically closes linked issue when PR is merged.
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
Notice that do not add `Fixes` if the issue is associated with multiple PRs.
-->
Fixes #5313 

@wangyang0616 @JesseStutler @hajnalmt 

<!--
If no, just write "NONE" in the release-note block below.
If yes, a release note is required.
-->
```release-note
The usage plugin now supports shadow load caching and configurable resource estimation for Pods not yet visible in metrics. New scheduler plugin arguments include `request.weight`, `burst.weight`, `risk_threshold`, `risk_factor`, `be.cpu`, and `be.memory`.
```

### DIFF
--- a/docs/design/usage-based-scheduling.md
+++ b/docs/design/usage-based-scheduling.md
@@ -74,6 +74,8 @@ metrics:                               # metrics server related configuration
     hostnameFieldName: "host.hostname" # Optional, The elasticsearch hostname field name, "host.hostname" by default
   ```
 
+If a usage plugin argument is configured with an invalid value, the plugin sets that argument to its default value.
+
 ### How to predicate node
 The plugins allow user to configure the cpu and memory average threshold within 5m.
 Any node whose usage is higher than the value of `CpuUsageAvg.5m` or `MemUsageAvg.5m` is filtered. If no threshold is configured, the node gets into priority stage.
@@ -88,6 +90,41 @@ Suppose there are two nodes with similar usage, The usage of one node fluctuates
 The third factor identified is the resource dimension. Take the below table as example. if there is pending pod which is a compute sensitive pod, it is more suitable to schedule it to `node2` with higher mem weight. DRF might be suitable to handle the case to calculate the cpu, mem, gpu share for pod and each node then make the best match.
 
 Finally, there should a model to balance multiple factors with weight and calculate the final score for nodes. Only the cpu usage factor will be considered in the alpha version.
+
+### Resource estimation for monitoring delay
+The usage plugin keeps a session-level shadow cache for pods that have been assigned to a node but whose metrics are not visible yet. The estimated resource of a Guaranteed or Burstable pod is:
+
+```
+pod_estimate = (request * request_ratio + (limit - request) * burst_ratio) * applied_risk_factor
+```
+
+The estimate is clamped to `[0, limit]`. If the pod has no limit for a resource, the request is used as the effective limit for that resource.
+
+`applied_risk_factor` is selected from the node composite load:
+
+```
+cpu_composite = (real_cpu_load + shadow_cpu_estimate) / node_cpu_capacity
+memory_composite = (real_memory_load + shadow_memory_estimate) / node_memory_capacity
+load_composite_percentage = weighted_average(cpu_composite, memory_composite, cpu.weight, memory.weight)
+```
+
+When `load_composite_percentage >= risk_threshold`, the estimate is multiplied by `risk_factor`; otherwise the multiplier is `1.0`.
+
+BestEffort pods do not use node-capacity ratios or density penalties. They use fixed configured estimates and are also affected by `risk_factor`:
+
+BestEffort estimates are configured under `estimator.be_cpu` and `estimator.be_mem`.
+
+Estimator configuration:
+
+```
+estimator:
+  request_ratio: 0.7   # 0 <= request_ratio <= 1
+  burst_ratio: 0       # 0 <= burst_ratio <= 1
+  risk_threshold: 0.6  # composite load threshold, 0.6 means 60%
+  risk_factor: 1.2     # applied after the threshold is reached, must be >= 1.0
+  be_cpu: 250m         # fixed BestEffort CPU estimate
+  be_mem: 200Mi        # fixed BestEffort memory estimate
+```
 
 | factors                   | node1           | node2            |
 | ----                      | ----            | ---              |
--- a/pkg/scheduler/plugins/usage/estimator.go
+++ b/pkg/scheduler/plugins/usage/estimator.go
@@ -1,0 +1,142 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package usage
+
+import (
+	v1 "k8s.io/api/core/v1"
+	fwk "k8s.io/kube-scheduler/framework"
+
+	"volcano.sh/volcano/pkg/scheduler/api"
+)
+
+// CalcLoadCompositePercentage computes the weighted composite node load from
+// CPU and memory composite utilization values. Inputs and output use the 0.0-1.0
+// range, where 0.6 means 60%.
+func CalcLoadCompositePercentage(cpuComposite, memComposite float64, cpuWeight, memWeight int) float64 {
+	totalWeight := cpuWeight + memWeight
+	if totalWeight == 0 {
+		return 0
+	}
+	return (cpuComposite*float64(cpuWeight) + memComposite*float64(memWeight)) / float64(totalWeight)
+}
+
+// CalcAppliedRiskFactor returns riskFactor once the node composite load reaches
+// riskThreshold. riskFactor values lower than 1.0 are ignored because the risk
+// multiplier should never make estimates less conservative under pressure.
+func CalcAppliedRiskFactor(loadCompositePercentage, riskThreshold, riskFactor float64) float64 {
+	if riskFactor < 1.0 {
+		return 1.0
+	}
+	if loadCompositePercentage >= riskThreshold {
+		return riskFactor
+	}
+	return 1.0
+}
+
+// EstimatePodResource estimates a single dimension (CPU or MEM) resource
+// consumption for a Guaranteed/Burstable pod.
+//
+// Formula:
+//
+//	estimated = (request*requestWeight + (limit-request)*burstWeight) * appliedRiskFactor
+//
+// The result is clamped to [0, effectiveLimit]. If limit is missing or lower
+// than request, request is used as the effective limit.
+func EstimatePodResource(request, limit, requestWeight, burstWeight, appliedRiskFactor float64) float64 {
+	if request < 0 {
+		request = 0
+	}
+	effectiveLimit := limit
+	if effectiveLimit <= 0 || effectiveLimit < request {
+		effectiveLimit = request
+	}
+	burst := effectiveLimit - request
+	if burst < 0 {
+		burst = 0
+	}
+	estimate := (request*requestWeight + burst*burstWeight) * appliedRiskFactor
+	return clampFloat64(estimate, 0, effectiveLimit)
+}
+
+// EstimateBestEffortResource estimates a BestEffort pod from the configured
+// fixed value. BestEffort pods are also affected by the node risk factor.
+func EstimateBestEffortResource(configuredValue, appliedRiskFactor float64) float64 {
+	if configuredValue < 0 {
+		configuredValue = 0
+	}
+	return configuredValue * appliedRiskFactor
+}
+
+func clampFloat64(value, min, max float64) float64 {
+	if value < min {
+		return min
+	}
+	if value > max {
+		return max
+	}
+	return value
+}
+
+// CalcCompositeUtilization computes the composite utilization for a single dimension (CPU or MEM).
+// Formula: U_r = (realLoadPercent/100 * capacity + shadowEstAbs) / capacity
+// The result is clamped to [0.0, 1.0].
+//
+// Parameters:
+//   - realLoadPercent: real load percentage from Prometheus (0-100)
+//   - shadowEstAbs: absolute shadow estimated load for this dimension
+//   - capacity: node capacity for this dimension (absolute value)
+func CalcCompositeUtilization(realLoadPercent float64, shadowEstAbs, capacity float64) float64 {
+	if capacity <= 0 {
+		return 0
+	}
+	realAbs := realLoadPercent / 100.0 * capacity
+	composite := (realAbs + shadowEstAbs) / capacity
+	if composite > 1.0 {
+		return 1.0
+	}
+	if composite < 0 {
+		return 0
+	}
+	return composite
+}
+
+// CalcNodeScore computes the node score from composite utilization values.
+// Formula: score = ((1 - cpuComp) * cpuWeight + (1 - memComp) * memWeight) / (cpuWeight + memWeight) * MaxNodeScore * usageWeight
+//
+// Parameters:
+//   - cpuComp: composite CPU utilization (0.0 - 1.0)
+//   - memComp: composite MEM utilization (0.0 - 1.0)
+//   - cpuWeight: CPU weight from usagePlugin.cpuWeight
+//   - memWeight: MEM weight from usagePlugin.memoryWeight
+//   - usageWeight: plugin weight in the overall scheduler scoring
+func CalcNodeScore(cpuComp, memComp float64, cpuWeight, memWeight, usageWeight int) float64 {
+	totalWeight := cpuWeight + memWeight
+	if totalWeight == 0 {
+		return 0
+	}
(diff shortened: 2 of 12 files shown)


pkg/scheduler/plugins/usage/usage.go
```diff
@@ -155,36 +234,198 @@ func (up *usagePlugin) OnSessionOpen(ssn *framework.Session) {
...
+}
+
+// calcNodeScore computes the score for a node based on composite utilization.
+func (up *usagePlugin) calcNodeScore(node *api.NodeInfo) float64 {
+	cpuEst, memEst := up.shadowCache.GetNodeEst(node.Name)
+	realCPU := getRealCPUPercent(node, up.period)
+	realMem := getRealMemPercent(node, up.period)
+	cpuComp := CalcCompositeUtilization(realCPU, cpuEst, node.Capacity.MilliCPU)
+	memComp := CalcCompositeUtilization(realMem, memEst, node.Capacity.Memory)
+	score := CalcNodeScore(cpuComp, memComp, up.cpuWeight, up.memoryWeight)
```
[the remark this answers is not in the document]
3th4novo: up.usageWeight is considered in CalcNodeScore

pkg/scheduler/plugins/usage/estimator.go
```diff
@@ -0,0 +1,175 @@
...
+	if isBestEffort {
+		// BestEffort: nodeCap * ratio * beta^n
+		return nodeCap * beRatio * math.Pow(bePenalty, float64(bestEffortCount))
+	}
+	// Guaranteed/Burstable: Request + (Limit - Request) * sigmaDynamic
+	if limit <= request {
+		// If limit <= request (e.g., Guaranteed QoS where limit == request),
+		// the estimated usage is simply the request value scaled by sigmaDynamic.
+		return request * sigmaDynamic
+	}
```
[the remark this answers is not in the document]
3th4novo: fixed

pkg/scheduler/plugins/usage/estimator.go
```diff
@@ -0,0 +1,175 @@
...
+	bestEffortCount int, isBestEffort bool) float64 {
+	if isBestEffort {
+		// BestEffort: nodeCap * ratio * beta^n
+		return nodeCap * beRatio * math.Pow(bePenalty, float64(bestEffortCount))
+	}
+	// Guaranteed/Burstable: Request + (Limit - Request) * sigmaDynamic
+	if limit <= request {
+		// If limit <= request (e.g., Guaranteed QoS where limit == request),
+		// the estimated usage is simply the request value scaled by sigmaDynamic.
+		return request * sigmaDynamic
```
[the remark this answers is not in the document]
3th4novo: fixed

pkg/scheduler/plugins/usage/usage.go
```diff
@@ -76,34 +103,265 @@ func New(args framework.Arguments) framework.Plugin {
...
+			configArgs[keyStr] = value
 		}
+		return configArgs, true
+	default:
+		return nil, false
 	}
+}
 
-	return plugin
+func parseIntArg(args framework.Arguments, key string, target *int) {
```
JesseStutler: There is already a `GetInt` method for the Arguments or Get[T any] for the generics, 
Is there anything that existing functions cannot handle?

Also there are lots of other parse helper func, if the helper func in `pkg/scheduler/framework/arguments.go` already can handle parsing, then there are no need to add so many new helper, but you trully need it, just enhance in the same file

pkg/scheduler/plugins/usage/usage.go
```diff
@@ -155,36 +445,171 @@ func (up *usagePlugin) OnSessionOpen(ssn *framework.Session) {
...
+// shouldAddToShadowCache determines whether a task should be tracked in the shadow cache.
+// A task is added if it has been assigned to a node but its metrics are not yet visible
+// in the monitoring system. BestEffort pods follow the same logic as Guaranteed/Burstable.
+func shouldAddToShadowCache(task *api.TaskInfo, metricsDelay time.Duration) bool {
+	// Only consider pods that have been assigned to a node
+	if task.NodeName == "" {
+		return false
+	}
+	switch task.Status {
+	case api.Allocated, api.Binding, api.Bound:
```
JesseStutler, line 473: What about a single-pod scenario, or the first pod in a batch? It remains in a "Pending" state, in that case, do we not estimate its requested resources?

pkg/scheduler/plugins/usage/shadow_cache.go
JesseStutler, line 1: Currently, the shadow cache is initialized during `OpenSession` rather than being managed at cache granularity. How about its scheduling performance? Specifically, the latency of `OpenSession` and whether repeated `OpenSession` calls lead to cumulative memory growth or increased GC pressure.

pkg/scheduler/plugins/usage/estimator.go
```diff
@@ -0,0 +1,178 @@
...
+		return 0
+	}
+	cpuScore := (1.0 - cpuComp) * float64(cpuWeight)
+	memScore := (1.0 - memComp) * float64(memWeight)
+	score := (cpuScore + memScore) / float64(totalWeight)
+	return score * float64(fwk.MaxNodeScore) * float64(usageWeight)
+}
+
+// getPodResourceRequestLimit extracts CPU (milliCPU) and memory (bytes) request/limit for a pod.
+func getPodResourceRequestLimit(pod *v1.Pod) (cpuRequest, cpuLimit, memRequest, memLimit float64) {
```
JesseStutler, line 138: There is already a func called `aggregateAllContainerResourceRequests` in the `pkg/scheduler/api/pod_info.go`, which can be refered or reuse it, should consider sidecar container, etc.

docs/design/usage-based-scheduling.md
```diff
@@ -89,6 +89,41 @@ The third factor identified is the resource dimension. Take the below table as e
...
+BestEffort estimates are configured under `estimator.be_cpu` and `estimator.be_mem`.
+
+Estimator configuration:
+
+```
+estimator:
+  request_ratio: 0.7   # 0 <= request_ratio <= 1
+  burst_ratio: 0       # 0 <= burst_ratio <= 1
+  risk_threshold: 0.6  # composite load threshold, 0.6 means 60%
+  risk_factor: 1.2     # applied after the threshold is reached
```
JesseStutler: The risk factor didn't notice users should larger than >= 1?

pkg/scheduler/plugins/usage/estimator.go
```diff
@@ -0,0 +1,178 @@
...
+//   - realLoadPercent: real load percentage from Prometheus (0-100)
+//   - shadowEstAbs: absolute shadow estimated load for this dimension
+//   - capacity: node capacity for this dimension (absolute value)
+func CalcCompositeUtilization(realLoadPercent float64, shadowEstAbs, capacity float64) float64 {
+	if capacity <= 0 {
+		return 0
+	}
+	realAbs := realLoadPercent / 100.0 * capacity
+	composite := (realAbs + shadowEstAbs) / capacity
+	if composite > 1.0 {
```
JesseStutler, line 108: Since this is an estimate, it is possible for the value to exceed 1; in that case, is the node's score simply 0(L129-L130) regardless of how much it exceeds 1? Is there no need to make a distinction?

pkg/scheduler/plugins/usage/usage.go
3th4novo: Pending tasks without NodeName cannot be considered to any node. Once the first pod is allocated, AllocateFunc records its estimate in the shadow cache, so later pods in the same session can see it.

pkg/scheduler/plugins/usage/estimator.go
3th4novo: Once the composite estimated load reaches or exceeds capacity, the usage plugin gives the minimum score. This is expected because we can still rank the node scores based on other plugins. From the perspective of the usage plugin and the shadow load cache, the node is already full, so I don't think there's any need to make another distinction.

docs/design/usage-based-scheduling.md
3th4novo: fixed in docs

pkg/scheduler/plugins/usage/usage.go
3th4novo: enhanced and fixed

pkg/scheduler/plugins/usage/shadow_cache.go
3th4novo: The warm-up cost is O(total tasks on nodes) per session. It does not cause cumulative memory growth because the cache is reset when the session closes and rebuilt for the next session. However, the current OpenSession initialization does have redundant allocation: it may Reset the old cache and then immediately create a new one.

I will update this part to avoid the redundant reset/new allocation for warmUpShadowCache.

3th4novo: fixed

pkg/scheduler/plugins/usage/estimator.go
3th4novo: The plugin also requires the limit information of a pod. I defined a api.GetPodResourceLimit(pod), which is quite similar to api.GetPodResourceRequest(pod) as before, for pod resource retrieval.

3th4novo: fixed

pkg/scheduler/plugins/usage/usage.go
JesseStutler: I don't get it, isn't your request value based on an estimate? Why do you need to consider whether it's the "first pod"? In a single-pod scenario, wouldn't that mean the estimated resource usage of these pods is never taken into account when they need to be scheduled?

pkg/scheduler/plugins/usage/shadow_cache.go
JesseStutler: /cc @wangyang0616 I'm ok with it for now, but in a large-scale cluster, iterating through all nodes and tasks would take a significant amount of time, likely slowing down OpenSession procedure

pkg/scheduler/plugins/usage/usage.go
```diff
@@ -76,34 +103,265 @@ func New(args framework.Arguments) framework.Plugin {
...
+	}
+	value := float64(quantity.MilliValue())
+	if value < 0 {
+		klog.Warningf("Could not parse argument: %v for key %s, expected non-negative CPU quantity", args[key], key)
+		return
+	}
+	*target = value
+}
+
+func parseMemoryQuantityArg(args framework.Arguments, key string, target *float64) {
```
JesseStutler: https://github.com/volcano-sh/volcano/pull/5314#discussion_r3497458538  Why are there still so many parsers here? Is there something that `pkg/scheduler/framework/arguments.go` doesn't cover? I find it odd that you convert the value to a `Quantity` only to have the caller convert it to a `float64`; why not just use `GetFloat64` directly?

pkg/scheduler/plugins/usage/usage.go
JesseStutler: Refer from koordinator: https://github.com/koordinator-sh/koordinator/blob/64c232b3b0a991af3532ffe2f11e776ea7a71ea3/pkg/scheduler/plugins/loadaware/load_aware.go#L295-L314, In other words, you only consider the estimator after the pod has reached the "Assumed" state，however, we should also consider the estimated resources of the incoming pod.

pkg/scheduler/plugins/usage/shadow_cache.go
3th4novo: I agree with this. But in this plugin, shadow cache is designed to not cross session. I think this inherently requires to scan over every pod on every node to warm up a session.

But I do think that the pattern is similar to other Volcano plugins that rebuilds session-local cache in OnSessionOpen, such as DRF/proportion/capacity. There is a latency tradeoff, but we make sure the cache is clean and avoids inconsistency.

pkg/scheduler/plugins/usage/usage.go
JesseStutler: Sorry I misread the old code, but I still want to confirm whether a change requiring so many arguments helper is necessary?

pkg/scheduler/plugins/usage/usage.go
3th4novo: I think there's a misunderstanding here. For the first pod (or any pod) to be scheduled, if the pod is bound to a node, this means that this node must have the least composite load, and thus it got the highest score. Therefore, given the current state of the nodes, we cannot find another node that has a lower composite load, which is more suitable for the pod. This makes the current scheduling decision optimal.

pkg/scheduler/plugins/usage/usage.go
3th4novo: Hi I rollbacked the relevant changes in arguments.go to make logics easier to understand and use. Now most of the configs call GetInt and GetFloat64 directly, only a few configs have to do a bit of costomized modifications in usage.go.

PR #5323, opened 2026-05-19
fix(reclaim): ensure gang job scheduling with partial pod reclamation

zhifei92:
Resolved a bug in gang scheduling logic. Now correctly handles jobs where some pods fit on available nodes, while others require pod eviction to meet resource requests. This ensures successful gang scheduling even when only a subset of pods triggers resource reclamation.

#### What type of PR is this?
/kind bug

For a gang job, some pods have sufficient resources for scheduling, while others lack the necessary resources. However, by using reclaim, enough resources can be freed up for the unschedulable pods, which should ultimately allow the gang job to be scheduled successfully.

Currently, the reclaim implementation fails to schedule the job successfully in certain scenarios:

Case 1: Node1 is idle and can accommodate some pods, while Node2 has preemptable tasks. Scheduling should succeed after eviction on Node2 satisfies the gang requirements.

Case 2: A job has two tasks: Task1 requires one GPU, and Task2 does not. There is a node with enough resources to schedule Task2 (which doesn't need a GPU, and has no preemptable pods during the reclaim phase). Task1 must and successfully obtains a GPU through eviction. However, in this situation, the job still fails to schedule.

We should remove the `if len(reclaimees) == 0 {` check and push it down into `if err := util.ValidateVictims(task, n, victims); err != nil {`. This validation not only ensures correct scheduling but also comes with a negligible cost."

https://github.com/volcano-sh/volcano/blob/4687fce2dc41d3e5f56d38f34db559fe9f725ff0/pkg/scheduler/actions/reclaim/reclaim.go#L203-L214

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -201,11 +201,6 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 			}
 		}
 
-		if len(reclaimees) == 0 {
-			klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
-			continue
-		}
-
 		victims := ssn.Reclaimable(task, reclaimees)
 		if err := util.ValidateVictims(task, n, victims); err != nil {
 			klog.V(3).Infof("No validated victims on Node <%s>: %v", n.Name, err)
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -122,6 +122,8 @@ func (ra *Action) Execute(ssn *framework.Session) {
 			}
 			job := jobsQ.Pop().(*api.JobInfo)
 			stmt := framework.NewStatement(ssn)
+			reclaimedInTransaction := false
+			var deferredTasks []*api.TaskInfo
 
 			for {
 				// If job is not request more resource, then stop reclaiming.
@@ -156,7 +158,25 @@ func (ra *Action) Execute(ssn *framework.Session) {
 					continue
 				}
 
-				ra.reclaimForTask(ssn, stmt, task, job)
+				reclaimed, deferred := ra.reclaimForTask(ssn, stmt, task, job, reclaimedInTransaction)
+				if reclaimed {
+					reclaimedInTransaction = true
+				}
+				if deferred {
+					deferredTasks = append(deferredTasks, task)
+				}
+			}
+
+			for ssn.JobStarving(job) && reclaimedInTransaction && len(deferredTasks) > 0 {
+				task := deferredTasks[0]
+				deferredTasks = deferredTasks[1:]
+
+				if err := ssn.PrePredicateFn(task); err != nil {
+					klog.V(3).Infof("PrePredicate failed for deferred task %s/%s: %v", task.Namespace, task.Name, err)
+					continue
+				}
+
+				ra.reclaimForTask(ssn, stmt, task, job, true)
 			}
 
 			if ssn.JobPipelined(job) {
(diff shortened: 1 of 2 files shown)


JesseStutler:
There is a failure in code verify CI, please check @zhifei92

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -201,11 +201,6 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 			}
 		}
 
-		if len(reclaimees) == 0 {
```
JesseStutler, line 204: I didn't get the point, why remove such guard codes would solve your scenario?

zhifei92:
To be compatible with the current scenario (`reclaim` before `allocate`) - Reclaim should not pipeline tasks that can be scheduled directly by allocate when there are no reclaimable victims. this change defers such tasks and only retries them after the same gang job has already reclaimed resources in the current transaction, allowing reclaim to complete gang scheduling without stealing normal allocate work.

pkg/scheduler/actions/reclaim/reclaim.go
zhifei92: Here, we can ensure that **some pods of a gang job enter the pipeline state via resource reclaim**. If the remaining pods can be scheduled **without requiring resource reclaim** (i.e., sufficient resources are available), **they should also be allowed to enter the pipeline state**. Only then can the entire job **fully progress through the pipeline**.


By analyzing the previously failed unit test cases, i discovered that simply removing this section might inadvertently alter the current design. It has now been revised to a two-phase reclaim mechanism.

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -201,8 +222,10 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
...
-			klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
+		if len(reclaimees) == 0 && !allowPipelineWithoutReclaim {
+			deferredWithoutReclaimees = true
```
JesseStutler, line 226: This may not correct since `deferredWithoutReclaimees` may not refresh to false again when we find reclaimees in subsequent nodes

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -156,7 +158,25 @@ func (ra *Action) Execute(ssn *framework.Session) {
 					continue
 				}
 
-				ra.reclaimForTask(ssn, stmt, task, job)
+				reclaimed, deferred := ra.reclaimForTask(ssn, stmt, task, job, reclaimedInTransaction)
```
JesseStutler, line 161: Are reclaimed and deferred two mutually exclusive flags?

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -156,7 +158,25 @@ func (ra *Action) Execute(ssn *framework.Session) {
...
+				if deferred {
+					deferredTasks = append(deferredTasks, task)
+				}
+			}
+
+			for ssn.JobStarving(job) && reclaimedInTransaction && len(deferredTasks) > 0 {
+				task := deferredTasks[0]
+				deferredTasks = deferredTasks[1:]
+
+				if err := ssn.PrePredicateFn(task); err != nil {
```
JesseStutler, line 174: I don't know why we need to PrePredicate in the reclaim action?

JesseStutler:
> To be compatible with the current scenario (`reclaim` before `allocate`) - Reclaim should not pipeline tasks that can be scheduled directly by allocate when there are no reclaimable victims. this change defers such tasks and only retries them after the same gang job has already reclaimed resources in the current transaction, allowing reclaim to complete gang scheduling without stealing normal allocate work.

While there's no strict order for actions, allocate action generally precedes reclaim. You mean that if a reclaim action comes before allocate, it shouldn't preempt pods that can be allocated by allocate action, right? This is actually quite hard. You can't determine during reclaim whether subsequent allocation in allocate action will definitely be able to schedule those pods successfully. This seems to imply that actions must be sequential.

Issue #5325, opened 2026-05-19
Bug: VCJob permanently stuck in Pending with no pods created due to missing AddFunc on pgInformer


pulangt:
### Description

The job controller registers only `UpdateFunc` on the pgInformer, missing `AddFunc`:

```go
// pkg/controllers/job/job_controller.go:246
cc.pgInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
    UpdateFunc: cc.updatePodGroup,
    // AddFunc is missing
})
```

In `syncJob`, `getPodGroupByJob` reads from the lister (informer cache). At job creation time the PodGroup is not yet in the cache, so it returns nil → `syncTask=false` → pod creation is skipped. Both initial `syncJob` runs return `nil` → `queue.Forget`, and the VCJob leaves the work queue.

Under high load, the scheduler updates PodGroup phase `""→Inqueue` before pgInformer caches it. pgInformer's first observation of the PodGroup is already at `Inqueue`. With no `AddFunc`, the Add event is silently dropped. `UpdateFunc` also cannot rescue it because `updatePodGroup` only re-enqueues on phase change — with pgInformer starting at `Inqueue`, there is no subsequent phase transition to observe. The VCJob permanently leaves the work queue with 0 pods created.

There is also an existing TODO at the end of the file acknowledging incomplete PodGroup event handling:

```go
// TODO(k82cn): add handler for PodGroup unschedulable event.
```

### Steps to reproduce the issue

1. Deploy Volcano on a cluster under moderate-to-high load
2. Create a VCJob with sufficient queue capacity and cluster resources
3. VCJob stays in `Pending` indefinitely with no pods created
4. Controller logs show only 2 `syncJob` runs at creation time, then silence

### Describe the results you received and expected

| | Result |
|---|---|
| **Expected** | VCJob transitions to Running, pods are created |
| **Actual** | VCJob stuck in Pending permanently, 0 pods created |

### What version of Volcano are you using?

v1.13.1. Same code exists in current `master`.

### Any other relevant information

Add `AddFunc` to pgInformer and complete the switch in `updatePodGroup`

JesseStutler:
This seems to require a rather extreme scenario to reproduce. Would you like to contribute to fixing it?

avinxshKD:
@JesseStutler @hajnalmt can i take a look at this 😄?? If its available

avinxshKD:
[...]

PR #5338, opened 2026-05-23
fix: add AddFunc handler to pgInformer to prevent VCJob stuck in Pending

madmecodes:
## Summary

- Register `AddFunc` on the pgInformer in the job controller so PodGroup creation events trigger a job re-sync
- Without this handler, when the scheduler updates PodGroup phase before the informer caches it, the Add event is silently dropped and the VCJob permanently leaves the work queue with 0 pods created
- The handler extracts the owning job name from OwnerReferences (falling back to PodGroup name for PodGroups not owned by a Volcano Job) and re-queues the job with `OutOfSyncEvent`

Related: #5325

## Test plan

- `TestAddPodGroupFunc`: name resolution for PodGroups with and without Job OwnerReferences
- `TestAddPodGroupRequiredForPreSetPhase`: deterministic test proving AddFunc is required when the informer first observes a PodGroup with phase already set (Inqueue). Validates that addPodGroup enqueues while updatePodGroup with same phase does not, confirming AddFunc is the only enqueue path in this scenario
- All 243 tests pass

### DIFF
--- a/pkg/controllers/job/job_controller_handler.go
+++ b/pkg/controllers/job/job_controller_handler.go
@@ -449,6 +449,31 @@ func (cc *jobcontroller) processNextCommand() bool {
 	return true
 }
 
+func (cc *jobcontroller) addPodGroup(obj interface{}) {
+	pg, ok := obj.(*scheduling.PodGroup)
+	if !ok {
+		klog.Errorf("Failed to convert %v to PodGroup", obj)
+		return
+	}
+
+	jobNameKey := pg.Name
+	for _, or := range pg.OwnerReferences {
+		if or.Kind == helpers.JobKind.Kind {
+			jobNameKey = or.Name
+			break
+		}
+	}
+
+	req := apis.Request{
+		Namespace: pg.Namespace,
+		JobName:   jobNameKey,
+		Event:     bus.OutOfSyncEvent,
+	}
+	key := jobhelpers.GetJobKeyByReq(&req)
+	queue := cc.getWorkerQueue(key)
+	queue.Add(req)
+}
+
 func (cc *jobcontroller) updatePodGroup(oldObj, newObj interface{}) {
 	oldPG, ok := oldObj.(*scheduling.PodGroup)
 	if !ok {
--- a/pkg/controllers/job/job_controller_handler_test.go
+++ b/pkg/controllers/job/job_controller_handler_test.go
@@ -569,6 +569,67 @@ func TestDeletePodFunc(t *testing.T) {
 	}
 }
 
+func TestAddPodGroupFunc(t *testing.T) {
+	namespace := "test"
+	boolTrue := true
+
+	testCases := []struct {
+		Name        string
+		PodGroup    *scheduling.PodGroup
+		ExpectValue int
+		ExpectKey   string
+	}{
+		{
+			Name: "PodGroup with Job OwnerReference re-queues job",
+			PodGroup: &scheduling.PodGroup{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pg1-uid",
+					Namespace: namespace,
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion: helpers.JobKind.GroupVersion().String(),
+							Kind:       helpers.JobKind.Kind,
+							Name:       "job1",
+							Controller: &boolTrue,
+						},
+					},
+				},
+				Status: scheduling.PodGroupStatus{
+					Phase: scheduling.PodGroupInqueue,
+				},
+			},
+			ExpectValue: 1,
+			ExpectKey:   "test/job1",
+		},
+		{
+			Name: "PodGroup without OwnerReference uses PodGroup name",
+			PodGroup: &scheduling.PodGroup{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "legacy-pg",
+					Namespace: namespace,
+				},
+				Status: scheduling.PodGroupStatus{
+					Phase: scheduling.PodGroupPending,
+				},
+			},
+			ExpectValue: 1,
+			ExpectKey:   "test/legacy-pg",
+		},
+	}
+
+	for i, testcase := range testCases {
+		t.Run(testcase.Name, func(t *testing.T) {
+			controller := newController()
+			controller.addPodGroup(testcase.PodGroup)
+			queue := controller.getWorkerQueue(testcase.ExpectKey)
+			qLen := queue.Len()
+			if testcase.ExpectValue != qLen {
+				t.Errorf("case %d (%s): expected queue length: %v, got %v", i, testcase.Name, testcase.ExpectValue, qLen)
+			}
+		})
+	}
+}
+
 func TestUpdatePodGroupFunc(t *testing.T) {
 
 	namespace := "test"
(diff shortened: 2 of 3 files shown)


pkg/controllers/job/job_controller_handler.go
```diff
@@ -449,6 +449,30 @@ func (cc *jobcontroller) processNextCommand() bool {
...
+		klog.Errorf("Failed to convert %v to PodGroup", obj)
+		return
+	}
+
+	jobNameKey := pg.Name
+	for _, or := range pg.OwnerReferences {
+		if or.Kind == "Job" {
+			jobNameKey = or.Name
+		}
+	}
```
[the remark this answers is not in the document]
madmecodes: Done. Added `break` and switched to `helpers.JobKind.Kind`. Kept the same OwnerReferences loop pattern as `updatePodGroup` for consistency.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -449,6 +449,30 @@ func (cc *jobcontroller) processNextCommand() bool {
...
+
+	req := apis.Request{
+		Namespace: pg.Namespace,
+		JobName:   jobNameKey,
+		Event:     bus.OutOfSyncEvent,
+	}
+	key := jobhelpers.GetJobKeyByReq(&req)
+	queue := cc.getWorkerQueue(key)
+	queue.Add(req)
+}
```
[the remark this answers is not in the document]
madmecodes: The duplication is real but refactoring `updatePodGroup` is outside the scope of this fix. Happy to do a follow-up if maintainers want it.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -449,6 +449,30 @@ func (cc *jobcontroller) processNextCommand() bool {
...
+		return
+	}
+
+	jobNameKey := pg.Name
+	for _, or := range pg.OwnerReferences {
+		if or.Kind == "Job" {
+			jobNameKey = or.Name
+		}
+	}
+
```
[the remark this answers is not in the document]
madmecodes: Matches the existing `updatePodGroup` pattern on the same file. Changing one without the other would be inconsistent. The fallback to `pg.Name` covers legacy PodGroups (pre-1.5) which don't have OwnerReferences.

installer/helm/chart/volcano/Chart.yaml
```diff
@@ -1,7 +1,7 @@
-apiVersion: v1
+apiVersion: v2
 description: A Helm chart for Volcano
 name: volcano
-version: "1.5"
+version: "1.5.0"
 appVersion: "0.1"
```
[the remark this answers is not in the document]
madmecodes: Fixed. Rebased to drop the unrelated Helm chart commit, PR now only contains the pgInformer fix.

installer/helm/chart/volcano/charts/jobflow/Chart.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 appVersion: "0.1"
 description: A Helm chart for Volcano-Job-flow
```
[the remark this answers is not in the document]
madmecodes: Same as above, rebased out.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -449,6 +449,31 @@ func (cc *jobcontroller) processNextCommand() bool {
...
+		if or.Kind == helpers.JobKind.Kind {
+			jobNameKey = or.Name
+			break
+		}
+	}
+
+	req := apis.Request{
+		Namespace: pg.Namespace,
+		JobName:   jobNameKey,
+		Event:     bus.OutOfSyncEvent,
```
hajnalmt, line 470: This is the right choice here I think too.

pkg/controllers/job/job_controller_handler_test.go
```diff
@@ -569,6 +569,67 @@ func TestDeletePodFunc(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "legacy-pg",
+					Namespace: namespace,
+				},
+				Status: scheduling.PodGroupStatus{
+					Phase: scheduling.PodGroupPending,
+				},
+			},
+			ExpectValue: 1,
+			ExpectKey:   "test/legacy-pg",
```
hajnalmt: Please rewrite this without this `legacy-pg` naming, its just not right.

pkg/controllers/job/job_controller_handler_test.go
```diff
@@ -569,6 +569,67 @@ func TestDeletePodFunc(t *testing.T) {
 	}
 }
 
+func TestAddPodGroupFunc(t *testing.T) {
```
hajnalmt, line 572: This TC is ok, but doesn't validate the actual problem in the issue.

The described bug path:
- informer’s first observation is an Add with phase already set (e.g. Inqueue),
- no subsequent phase transition occurs,
- UpdateFunc alone would not enqueue anything.

You can actually add a deterministic test that proves if the issue is somewhat solved:

1) `addPodGroup()` enqueues `OutOfSyncEvent` for `<ns>/<job>`,
2) `updatePodGroup(old,new)` with same phase does NOT enqueue,
3) therefore AddFunc is the required trigger in this scenario.

This scenario has no enqueue path without AddFunc.

hajnalmt (commented):
/ok-to-test
/area controllers

Thanks for the contribution! The issue seems plausible, and I’m fine adding this because it is low risk and could solve a real race. That said, I’m not fully convinced the scenario is real (or at least easily verifiable): it appears to require both initial reconciliations to happen before PodGroup is visible in the controller’s lister (or before a meaningful PodGroup update event is observed).

In the PR description you are misunderstanding a view points. The fallback behavior (`jobNameKey := pg.Name` when no Job ownerRef is found) is not only “legacy PodGroup” behaviour. It is a broader compatibility path and can also apply to PodGroups not owned by Volcano Job (including other controllers/workloads using PodGroup-like naming patterns, like Deployments, Statefulsets etc...). 

The current coverage (the two addPodGroup cases) validates name resolution only, not the race claim from the bug description.

pkg/controllers/job/job_controller_handler_test.go
madmecodes: Renamed to deploy-pg and updated the test case description. You're right, it's a general compatibility path for any PodGroup not owned by a Volcano Job, not just legacy.

pkg/controllers/job/job_controller_handler_test.go
madmecodes: Added TestAddPodGroupRequiredForPreSetPhase that validates the exact race path: (1) addPodGroup enqueues OutOfSyncEvent, (2) updatePodGroup with same phase does not enqueue, proving AddFunc is the only trigger when the informer first sees a PodGroup with phase already set.

madmecodes:
@hajnalmt addressed your review. Renamed the test off legacy-pg and added TestAddPodGroupRequiredForPreSetPhase that proves AddFunc is the required trigger when the informer first sees a PodGroup with phase already set. PTAL when you get a chance.

pkg/controllers/job/job_controller_handler.go
```diff
@@ -449,6 +449,31 @@ func (cc *jobcontroller) processNextCommand() bool {
...
+	}
+
+	jobNameKey := pg.Name
+	for _, or := range pg.OwnerReferences {
+		if or.Kind == helpers.JobKind.Kind {
+			jobNameKey = or.Name
+			break
+		}
+	}
+
```
JesseStutler, line 466: I think you should add logic like in L498-L502 to `cc.cache.Get` from the cache to judge whether the vcjob exists, otherwise I may wonder that other workloads create podgroups will also trigger it and cause unpredictable errors

PR #5339, opened 2026-05-24
fix(enqueue): move inqueue mutation from JobEnqueueableFn to JobEnque…

### LINKED ISSUE #5387: fix(enqueue): phantom inqueue reservation when co-tier plugin rejects job
## What happened

When multiple plugins are configured in the same scheduler tier, `proportion` and `capacity` plugins mutated `attr.inqueue` inside `JobEnqueueableFn` ; before the overall `ssn.JobEnqueueable()` verdict was final.

`ssn.JobEnqueueable()` calls plugins in tier order and short-circuits on the first `Reject`. So if proportion/capacity said "fits, bumping inqueue" and a downstream plugin like `resourcequota` or `sla` said "no", the bump stuck for the rest of the session. The next job from the same queue walked in and saw inflated quota it had no business seeing.

For capacity with hierarchy this compounds: the phantom reservation propagated up to every ancestor queue, throttling sibling queues under the same parent for a job that was never enqueued.

## Root cause

`attr.inqueue.Add(...)` was called inside `JobEnqueueableFn` (vote phase) instead of `JobEnqueuedFn` (post-decision phase). The `overcommit` plugin already gets this right ; proportion and capacity did not follow the same contract.

## Affected plugins

- `pkg/scheduler/plugins/proportion/proportion.go`
- `pkg/scheduler/plugins/capacity/capacity.go`
- `pkg/scheduler/plugins/resourcequota/resourcequota.go`
- `pkg/scheduler/plugins/extender/extender.go` (no `JobEnqueuedFn` extension point existed)

## Fix

Moved all `inqueue`/`pendingResources` mutations into a new `JobEnqueuedFn` callback that only fires once `ssn.JobEnqueueable()` returns `true`. Added `JobEnqueuedFn` extension point to the extender so external plugins can hook in too.

Fixed in: #5339

## How to reproduce

Configure proportion (or capacity) in the same tier as resourcequota. Submit two jobs to a queue whose capacity fits exactly one job. The first job gets rejected by resourcequota but proportion phantom-inflates inqueue. The second job, which would otherwise fit, is incorrectly rejected.

Aman-Cool:
**What type of PR is this?**
/kind bug

After #5091 and #5100 fixed the double-counting in the inqueue initialization path, I kept staring at `attr.inqueue` and noticed a second class of incorrectness hiding in the same area.

Both proportion and capacity mutate `attr.inqueue` inside `JobEnqueueableFn` ; before the overall `ssn.JobEnqueueable()` verdict is even in. The dispatcher calls plugins in tier order and short-circuits on the first `Reject`, so if proportion/capacity says "fits, bumping inqueue" and a downstream plugin like `resourcequota` or `sla` says "actually no" ; the bump sticks for the rest of the session. The next job from the same queue walks in and sees inflated quota it has no business seeing.

For capacity with hierarchy it compounds: the phantom reservation propagates up to every ancestor queue, so sibling queues under the same parent end up throttled for a job that was never enqueued.

The fix is just following the pattern `overcommit` already gets right, `JobEnqueueableFn` is a pure vote (no side effects), and the actual `attr.inqueue.Add()` fires in a new `JobEnqueuedFn` callback that only runs once the full decision is confirmed as `Permit`.

Fixes #5387 

`overcommit` already implements the correct two-phase pattern (vote in `JobEnqueueableFn`, mutate in `JobEnqueuedFn`) ; the new callbacks in both plugins mirror it exactly. This only manifests when a plugin registered after proportion/capacity in the same tier returns `Reject`; single-plugin configs are unaffected.

**AI Disclosure**: This change was developed with AI assistance (Claude). The author has reviewed and understands all changes.

Fixed phantom inqueue quota inflation in proportion and capacity plugins when multiple `JobEnqueueableFn` plugins are configured in the same tier. Jobs were being incorrectly rejected for enqueue due to a reservation left by a previously rejected job. In hierarchical queue mode, the inflation also propagated to ancestor queues, throttling sibling queues unnecessarily.

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -775,7 +775,22 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 			return util.Reject
 		}
 
-		// job enqueued
+		return util.Permit
+	})
+
+	ssn.AddJobEnqueuedFn(cp.Name(), func(obj interface{}) {
+		job, ok := obj.(*api.JobInfo)
+		if !ok || job == nil || job.PodGroup == nil {
+			return
+		}
+		queueID := job.Queue
+		attr := cp.queueOpts[queueID]
+		if attr == nil {
+			return
+		}
+		if job.PodGroup.Spec.MinResources == nil && !(cp.dynamicResourceAllocationEnable && attr.dra != nil && job.GetMinDRAResources() != nil) {
+			return
+		}
 		deductedResources := job.DeductSchGatedResources(job.GetMinResources())
 		attr.inqueue.Add(deductedResources)
 		var minDRAReq map[string]*api.DRAResource
@@ -796,7 +811,6 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 			}
 		}
 		klog.V(5).Infof("job <%s/%s> enqueued", job.Namespace, job.Name)
-		return util.Permit
 	})
 
 	ssn.AddPrePredicateFn(cp.Name(), func(task *api.TaskInfo) error {
--- a/pkg/scheduler/plugins/capacity/capacity_test.go
+++ b/pkg/scheduler/plugins/capacity/capacity_test.go
@@ -1959,3 +1959,123 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
 		t.Fatalf("expected enqueueable DRA check to account inqueue resources")
 	}
 }
+
+// TestJobEnqueuedOnOtherPluginsReject verifies that the capacity plugin updates
+// a queue's inqueue accounting only when a job is fully admitted by all plugins in
+// the tier. A job rejected by a co-tier plugin must not consume queue capacity,
+// leaving room for subsequent jobs that fit within the queue's limit.
+func TestJobEnqueuedOnOtherPluginsReject(t *testing.T) {
+	trueValue := true
+	res1CPU := api.BuildResourceList("1", "1G")
+
+	n1 := util.BuildNode("n1", api.BuildResourceList("4", "4G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), nil)
+	podA := util.BuildPod("ns1", "podA", "", corev1.PodPending, res1CPU, "pgA", nil, nil)
+	podB := util.BuildPod("ns1", "podB", "", corev1.PodPending, res1CPU, "pgB", nil, nil)
+
+	pgA := util.BuildPodGroup("pgA", "ns1", "q1", 1, nil, schedulingv1beta1.PodGroupPending)
+	pgB := util.BuildPodGroup("pgB", "ns1", "q1", 1, nil, schedulingv1beta1.PodGroupPending)
+	pgA.Spec.MinResources = &res1CPU
+	pgB.Spec.MinResources = &res1CPU
+
+	// Queue with exactly 1 CPU capability — only one job can be admitted per session.
+	queue1 := util.BuildQueueWithResourcesQuantity("q1", api.BuildResourceList("1", "4G"), api.BuildResourceList("1", "4G"))
+
+	const rejecterName = "test-rejecter"
+	plugins := map[string]framework.PluginBuilder{
+		PluginName:   New,
+		rejecterName: func(_ framework.Arguments) framework.Plugin { return &uthelper.RejecterPlugin{RejectJob: "pgA"} },
+	}
+	tiers := []conf.Tier{
+		{
+			Plugins: []conf.PluginOption{
+				{Name: PluginName, EnabledJobEnqueued: &trueValue},
+				{Name: rejecterName, EnabledJobEnqueued: &trueValue},
+			},
+		},
+	}
+
+	test := uthelper.TestCommonStruct{
+		Name:      "pgA rejected by co-tier plugin must not consume q1 capacity; pgB must be admitted",
+		Plugins:   plugins,
+		Pods:      []*corev1.Pod{podA, podB},
+		Nodes:     []*corev1.Node{n1},
+		PodGroups: []*schedulingv1beta1.PodGroup{pgA, pgB},
+		Queues:    []*schedulingv1beta1.Queue{queue1},
+		ExpectStatus: map[api.JobID]scheduling.PodGroupPhase{
+			"ns1/pgB": scheduling.PodGroupInqueue,
+		},
+	}
+	test.RegisterSession(tiers, nil)
+	defer test.Close()
+	test.Run([]framework.Action{enqueue.New()})
+	if err := test.CheckAll(0); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// TestJobEnqueuedUpdatesDRAInqueueForDRAOnlyJob verifies that JobEnqueuedFn correctly
+// updates DRA inqueue accounting for jobs that have no scalar MinResources but do have
+// DRA resource requests. The early return guard must not skip DRA-only jobs.
+func TestJobEnqueuedUpdatesDRAInqueueForDRAOnlyJob(t *testing.T) {
+	if err := utilfeature.DefaultMutableFeatureGate.Set(fmt.Sprintf("%s=true", kubefeatures.DynamicResourceAllocation)); err != nil {
+		t.Fatal(err)
+	}
+
+	const deviceClass = "gpu.com"
+	const claimName = "dra-job-claim"
+
+	// Queue: DRA capability 4 GPUs, no scalar capability.
+	queue1 := util.BuildQueueWithResourcesQuantity("q1", nil, nil)
+	queue1.Spec.Capability = corev1.ResourceList{
+		corev1.ResourceName(DeviceClassCountPrefix + deviceClass): resource.MustParse("4"),
+	}
+
+	// DRA-only job: MinResources == nil, requests 2 GPUs via ResourceClaim.
+	claim := &resourcev1.ResourceClaim{
+		ObjectMeta: metav1.ObjectMeta{Name: claimName, Namespace: "ns1"},
+		Spec: resourcev1.ResourceClaimSpec{
+			Devices: resourcev1.DeviceClaim{
+				Requests: []resourcev1.DeviceRequest{{
+					Name: "req-1",
+					Exactly: &resourcev1.ExactDeviceRequest{
+						DeviceClassName: deviceClass,
+						Count:           2,
+					},
+				}},
+			},
+		},
+	}
+	pod := util.BuildPod("ns1", "dra-pod", "", corev1.PodPending, api.BuildResourceList("0", "0"), "dra-pg", nil, nil)
+	pod.Spec.ResourceClaims = []corev1.PodResourceClaim{{
+		Name:              "claim-1",
+		ResourceClaimName: &[]string{claimName}[0],
+	}}
+	pod.Annotations[batchv1alpha1.TaskSpecKey] = "worker"
+
+	pg := util.BuildPodGroup("dra-pg", "ns1", "q1", 1, nil, schedulingv1beta1.PodGroupPending)
+	pg.Spec.MinTaskMember = map[string]int32{"worker": 1}
+	// MinResources intentionally left nil — DRA-only job.
+
+	trueValue := true
+	tiers := []conf.Tier{{Plugins: []conf.PluginOption{{
+		Name: PluginName, EnabledJobEnqueued: &trueValue,
+	}}}}
+
+	test := uthelper.TestCommonStruct{
+		Name:           "DRA-only job (MinResources==nil) must update DRA inqueue accounting on enqueue",
+		Plugins:        map[string]framework.PluginBuilder{PluginName: New},
+		Pods:           []*corev1.Pod{pod},
+		PodGroups:      []*schedulingv1beta1.PodGroup{pg},
+		Queues:         []*schedulingv1beta1.Queue{queue1},
+		ResourceClaims: []*resourcev1.ResourceClaim{claim},
+		ExpectStatus: map[api.JobID]scheduling.PodGroupPhase{
+			"ns1/dra-pg": scheduling.PodGroupInqueue,
+		},
+	}
+	test.RegisterSession(tiers, nil)
+	defer test.Close()
+	test.Run([]framework.Action{enqueue.New()})
+	if err := test.CheckAll(0); err != nil {
+		t.Fatal(err)
+	}
+}
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -432,13 +432,24 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		inqueue, resourceNames := r.LessEqualWithDimensionAndResourcesName(attr.realCapability, minReq)
 		klog.V(5).Infof("job %s inqueue %v", job.Name, inqueue)
 		if inqueue {
-			// deduct the resources of scheduling gated tasks in a job when calculating inqueued resources
-			// so that it will not block other jobs from being inqueued.
-			attr.inqueue.Add(job.DeductSchGatedResources(minReq))
 			return util.Permit
 		}
 		ssn.RecordPodGroupEvent(job.PodGroup, v1.EventTypeNormal, string(scheduling.PodGroupUnschedulableType), util.FormatResourceNames("queue resource quota insufficient", "insufficient", resourceNames))
 		return util.Reject
+	})
+
+	ssn.AddJobEnqueuedFn(pp.Name(), func(obj interface{}) {
+		job := obj.(*api.JobInfo)
+		if job.PodGroup.Spec.MinResources == nil {
+			return
+		}
+		attr := pp.queueOpts[job.Queue]
+		if attr == nil {
+			return
+		}
+		// deduct the resources of scheduling gated tasks in a job when calculating inqueued resources
+		// so that it will not block other jobs from being inqueued.
+		attr.inqueue.Add(job.DeductSchGatedResources(job.GetMinResources()))
 	})
 
 	ssn.AddSimulateAddTaskFn(pp.Name(), func(ctx context.Context, cycleState fwk.CycleState, taskToSchedule *api.TaskInfo, taskToAdd *api.TaskInfo, nodeInfo *api.NodeInfo) error {
--- a/pkg/scheduler/plugins/proportion/proportion_test.go
+++ b/pkg/scheduler/plugins/proportion/proportion_test.go
@@ -32,6 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
 
+	"volcano.sh/apis/pkg/apis/scheduling"
 	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	"volcano.sh/volcano/cmd/scheduler/app/options"
 	"volcano.sh/volcano/pkg/scheduler/actions/allocate"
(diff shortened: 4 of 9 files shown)


Aman-Cool:
Hey @hajnalmt @JesseStutler, wanted to flag this one, it's a fairly contained fix but can quietly hurt throughput in multi-plugin setups. The overcommit plugin already does this correctly (votes in `JobEnqueueableFn`, mutates in `JobEnqueuedFn`), proportion and capacity just weren't following the same contract.

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -424,15 +424,26 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			return
+		}
+		attr := pp.queueOpts[job.Queue]
+		if attr == nil {
+			return
+		}
+		// deduct the resources of scheduling gated tasks in a job when calculating inqueued resources
+		// so that it will not block other jobs from being inqueued.
+		attr.inqueue.Add(job.DeductSchGatedResources(job.GetMinResources()))
+	})
```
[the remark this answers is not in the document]
Aman-Cool: Fair point; @hajnalmt would love your input here... though the existing enqueue tests in both plugins already exercise the `attr.inqueue` accounting path, so the core logic is covered. A dedicated multi-plugin test (proportion + a rejecting plugin verifying no phantom inflation) would be cleaner isolation; I can add one if you want it before merge...

pkg/controllers/job/job_controller_actions.go
```diff
@@ -263,17 +263,21 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
...
-	}
-	if pg != nil {
-		if err := cc.vcClient.SchedulingV1beta1().PodGroups(job.Namespace).Delete(context.TODO(), pg.Name, metav1.DeleteOptions{}); err != nil {
-			if !apierrors.IsNotFound(err) {
-				klog.Errorf("Failed to delete PodGroup of Job %s/%s: %v", job.Namespace, job.Name, err)
-				return err
+	// Delete PodGroup only on full job kills; targeted pod restarts (RestartPodAction)
+	// must keep the PodGroup alive so the scheduler does not need to re-enqueue and
+	// re-schedule the job from scratch on every individual pod failure.
+	if target == nil {
```
hajnalmt, line 269: This looks like a good fix, but it’s unrelated to the inqueue-accounting bug fixed in the PR.

To keep the changes focused and auditable, can you move this commit to a separate PR? I think we shall add a test for this too.

pkg/scheduler/plugins/proportion/proportion.go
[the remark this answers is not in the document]
hajnalmt: Yes this is definetily needed, so we can catch regressions on this bug properly.

hajnalmt (commented):
Thanks for the contribution! What a discovery, this is a subtle and nasty bug, and your fixes are spot on.

Can we align this on a couple of points?

1.  Separate PRs for Separate Bugs

In this PR, you are addressing two distinct bugs:

- The enqueue accounting (inqueue) mutation timing in the scheduler plugins
- The PodGroup deletion logic in the job controller

To help reviewers, facilitate clean regression tracking, and match standard bugfix workflow, could you please:

- Move the controller change (job_controller_actions / PodGroup deletion fix) into a dedicated PR?
- This will ensure each PR addresses a single concern and can be properly reviewed and tested in isolation.

2. Bug Tickets and Test Coverage

Since you’re fixing two underlying issues, please ensure there are two distinct bug tickets filed (one for each root problem), and that each PR references its relevant ticket. 

3. resourceQuota Plugin Also Affected but not fixed here

The PR won’t be fully complete without addressing the same bug in the resourceQuota plugin:
- pendingResources is mutated during the enqueue “vote” phase, meaning it’s vulnerable to the exact same issue (phantom accounting if another plugin later rejects).
- See: resourcequota.go, lines 97-99 (example link) (https://github.com/volcano-sh/volcano/blob/1733c885334b303400d96972513133aef3a828cc/pkg/scheduler/plugins/resourcequota/resourcequota.go#L97-L99)
- Please update resourceQuota to use the two-phase pattern (JobEnqueueableFn for pure vote, JobEnqueuedFn for side effect) just as you did for proportion/capacity.

4. Scheduler Extension Point for JobEnqueuedFn

While it can be a separate PR, please ensure the scheduler extender also offers this extension point (JobEnqueuedFn or the equivalent), so all plugins (including external ones) can hook in.  
Currently, only jobEnqueueable is offered in the extender (see extender.go#L57-L62)

https://github.com/volcano-sh/volcano/blob/1733c885334b303400d96972513133aef3a828cc/pkg/scheduler/plugins/extender/extender.go#L57-L62

Adding a JobEnqueuedFn extension completes the scheduling extensibility and closes this class of bugs for all custom plugins.

Great find 😊 !

Aman-Cool:
Hey @hajnalmt, thanks for the thorough review! Addressed all four points:

**1. Separated the PodGroup fix**: Opened a separate PR for that(#5342).This one now only touches the scheduler enqueue path.

**2. Fixed the `resourcequota` plugin**: it had the exact same issue: `pendingResources[job.Namespace] = quotav1.Add(...)` was being called inside `JobEnqueueableFn`, inflating the namespace's pending resource count before the overall decision was final. Moved the mutation to a new `JobEnqueuedFn`, same two-phase pattern as proportion/capacity.

**3. Added `JobEnqueuedFn` to the extender**: added `ExtenderJobEnqueuedVerb` constant, `jobEnqueuedVerb` config field, `JobEnqueuedRequest`/`JobEnqueuedResponse` types, and registration in `OnSessionOpen`. External plugins using the extender can now hook into the confirmed-enqueue phase and avoid this whole class of phantom-reservation bugs.

**4. Added the regression test**: `TestNoPhantomInqueueReservationOnReject` in `proportion_test.go`. The setup is: queue with 1 CPU capacity, two pending jobs each needing 1 CPU, and an inline `testRejecterPlugin` in the same tier that blocks `pgA` but permits `pgB`. Without the fix, proportion's phantom inflation from evaluating `pgA` would cause `pgB` to also fail the capacity check. With the fix, `pgA`'s rejection leaves no trace and `pgB` is correctly enqueued. Test passes.

pkg/scheduler/plugins/proportion/proportion_test.go
```diff
@@ -627,3 +628,83 @@ func TestNoDoubleCountingForInqueueJobWithBindingTasks(t *testing.T) {
...
+}
+func (r *testRejecterPlugin) OnSessionClose(_ *framework.Session) {}
+
+// TestNoPhantomInqueueReservationOnReject is a regression test for the bug where
+// proportion's JobEnqueueableFn mutated attr.inqueue before the overall enqueue
+// decision was final. If a downstream plugin in the same tier rejected the job,
+// the inflation persisted as a phantom reservation, blocking subsequent jobs in
+// the same queue that would otherwise fit.
+//
+// Root cause: attr.inqueue.Add(...) was called inside JobEnqueueableFn (vote phase).
```
JesseStutler: I don't really like how AI agents always emphasize what bugs exist in our existing systems and what they've fixed this time. I think a function should only emphasize its own capability.

JesseStutler (commented):
Good catch! But could you also provide a unit test for capacity plugin and resource quota plugin?

pkg/scheduler/plugins/proportion/proportion_test.go
Aman-Cool: done

pmady:
the two-phase pattern (vote in JobEnqueueableFn, mutate in JobEnqueuedFn) is the right call. the current code is basically a TOCTOU bug — you commit the reservation before the final verdict is in.

does the new test cover the multi-plugin rejection case specifically? i.e. proportion says Permit, then a downstream plugin like sla rejects, and you verify inqueue didn't get bumped. that's the scenario that actually triggers the phantom inflation

Aman-Cool:
@pmady, yeah that's literally the whole point of the test :) ; `testRejecterPlugin` sits in the same tier as proportion, proportion says Permit on pgA, rejecter says no, and the queue is sized to exactly 1 CPU so if inqueue got bumped you'd see pgB fail the capacity check too. `ExpectStatus["ns1/pgB"] == Inqueue` is the assertion that catches it. it's all in `proportion_test.go` in this diff!

pmady:
ah got it, the testRejecterPlugin setup is exactly what i was looking for. nice that you sized the queue to 1 CPU so the phantom bump would be immediately visible on pgB

Aman-Cool:
Hey @hajnalmt @JesseStutler, just checking in.., are there any other changes needed from my side before this can move forward? Happy to address anything🙂!

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -1959,3 +1605,70 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
...
+	podA := util.BuildPod("ns1", "podA", "", corev1.PodPending, res1CPU, "pgA", nil, nil)
+	podB := util.BuildPod("ns1", "podB", "", corev1.PodPending, res1CPU, "pgB", nil, nil)
+
+	pgA := util.BuildPodGroup("pgA", "ns1", "q1", 1, nil, schedulingv1beta1.PodGroupPending)
+	pgB := util.BuildPodGroup("pgB", "ns1", "q1", 1, nil, schedulingv1beta1.PodGroupPending)
+	pgA.Spec.MinResources = &res1CPU
+	pgB.Spec.MinResources = &res1CPU
+
+	// Queue with exactly 1 CPU capability — only one job can be admitted per session.
+	queue1 := util.BuildQueueWithResourcesQuantity("q1", api.BuildResourceList("4", "4G"), api.BuildResourceList("1", "4G"))
```
hajnalmt: This will throw an error at RegisterSession on a bigger verbosity because deserved<cap  on CPU can we reduce  this to 1? `util.BuildQueueWithResourcesQuantity("q1", api.BuildResourceList("1", "4G"), api.BuildResourceList("1", "4G"))`

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -1959,3 +1605,70 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
 		t.Fatalf("expected enqueueable DRA check to account inqueue resources")
 	}
 }
+
+type rejecterPlugin struct{ rejectJob string }
```
hajnalmt: Can we put this into a common place like `pkg/scheduler/uthelper/rejecter_plugin.go` please? Then we can reuse in the other plugins too.

pkg/scheduler/plugins/proportion/proportion_test.go
```diff
@@ -627,3 +628,76 @@ func TestNoDoubleCountingForInqueueJobWithBindingTasks(t *testing.T) {
...
+		return 1 // util.Permit
+	})
+}
+func (r *testRejecterPlugin) OnSessionClose(_ *framework.Session) {}
+
+// TestNoPhantomInqueueReservationOnReject verifies that proportion's inqueue
+// accounting is updated only after all plugins in the tier admit a job. A job
+// rejected by a co-tier plugin must not consume queue capacity, leaving room for
+// subsequent jobs that fit within the queue's limit.
+func TestNoPhantomInqueueReservationOnReject(t *testing.T) {
```
hajnalmt: `TestProportionNoPhantomInqueueReservation` following the naming of capacity test. Or you can name the test like `TestNoPhantomInqueueReservation` in both plugins. Or the best would be I think if we can have a general name for this like `TestJobEnqueuedOnOtherPluginsReject`, which could actually be general across the three plugins.

hajnalmt (changes requested):
Great Progress!
I like the tests too. I had some minor suggestions, can you check them please?
Thank you!

pkg/scheduler/plugins/proportion/proportion_test.go
Aman-Cool: Thanks for the suggestions! Quick question on the naming: between `TestNoPhantomInqueueReservation` (same across all three plugins) and `TestJobEnqueuedOnOtherPluginsReject` (fully general), which do you prefer? The second feels more precise about what's being tested but the first is cleaner to read. Happy to go with whatever you think fits the codebase style better!

hajnalmt: `TestJobEnqueuedOnOtherPluginsReject` is my prefered but `TestNoPhantomInqueueReservationOnReject` is good too! Just name the capacity plugin and resourceQuota test the same please!

hajnalmt (commented):
Thanks for the changes! Can you create a Bug ticket too please? If someone bumps into this issue, they should be able to search for it! Please align the test name across the UT-s and this is fine from my side.

Aman-Cool:
@hajnalmt, Done! Bug ticket filed at #5387. And the test name is now aligned across all three plugins; `TestJobEnqueuedOnOtherPluginsReject` in proportion, capacity, and resourcequota. Thanks for the review!

hzxuzhonghu (commented):
generally lgtm

JesseStutler:
/approve
Can let @hajnalmt approve it

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -775,7 +775,19 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 			return util.Reject
 		}
 
-		// job enqueued
+		return util.Permit
+	})
+
+	ssn.AddJobEnqueuedFn(cp.Name(), func(obj interface{}) {
+		job, ok := obj.(*api.JobInfo)
+		if !ok || job == nil || job.PodGroup == nil || job.PodGroup.Spec.MinResources == nil {
```
hajnalmt: This early return looks like it can regress DRA-only jobs.

`JobEnqueueableFn` allows `MinResources == nil` when DRA is enabled and `job.GetMinDRAResources() != nil`, but this callback now returns before `updateDRAInqueue(...)`.  

I think we must split the scalar `MinResources` accounting from the DRA accounting, so DRA-only enqueued jobs still update `attr.dra` / ancestors.

hajnalmt (commented):
I am not sure we should merge this still.

The old code path still reached:
```
minDRAReq = job.GetMinDRAResources()
if minDRAReq != nil {
    updateDRAInqueue(attr.dra, minDRAReq)
}
```
once `checkJobEnqueueableHierarchically(...)` passed.

Now it early exits. I would still prefer this going in now, so we can somewhat cleanly cherry-pick the code change, but then we shall probably write an upcoming PR that fixes that? Or should we fix it here and cherry-pick the commits by hand? @JesseStutler

pkg/scheduler/plugins/capacity/capacity.go
hajnalmt: This shows that we should have a focused DRA test where
- `Spec.MinResources == nil`,
- `dynamicResourceAllocationEnable == true`,
- `job.GetMinDRAResources() != nil`,
- and the job is admitted
- DRA inqueue accounting didn't change.

That shows that we don't run into this accidentally.

Aman-Cool:
@hajnalmt I'd prefer to fix it here rather than opening a follow-up PR.., it keeps the fix complete and self-contained. That said, would love to hear @JesseStutler's thoughts before proceeding either way!

pkg/scheduler/plugins/capacity/capacity.go
Aman-Cool: > I think we must split the scalar MinResources accounting from the DRA accounting, so DRA-only enqueued jobs still update attr.dra / ancestors.

So I've been staring at this and I think we can actually dodge the split entirely..., the early return in `JobEnqueuedFn` just needs to mirror the exact same condition `JobEnqueueableFn` already uses on line 769:

```go
if job.PodGroup.Spec.MinResources == nil && !(cp.dynamicResourceAllocationEnable && attr.dra != nil && job.GetMinDRAResources() != nil) {
    return
}
```

That way DRA-only jobs fall through naturally, `attr.inqueue.Add` is a no-op on empty resources so no harm done, and `updateDRAInqueue` runs correctly ; no split needed. Basically the fix is just making the guard consistent with what's already there rather than introducing a new code path.

Will add the DRA test too. Still waiting to hear from @JesseStutler on the follow-up vs fix-here question before I push anything though!

hzxuzhonghu: SHould be fixed atomically here

JesseStutler: @Aman-Cool Should be fixed together here, because the L783 judgement is new, should use the same way, or wrap a func like using `Guard Clauses` will be better to read.

JesseStutler:
> @hajnalmt I'd prefer to fix it here rather than opening a follow-up PR.., it keeps the fix complete and self-contained. That said, would love to hear @JesseStutler's thoughts before proceeding either way!

I don't want to postpone to follow-up PR, should be fixed together, thanks for @hajnalmt reminding that we missed DRA judgement here

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -775,7 +775,23 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
...
+		job, ok := obj.(*api.JobInfo)
+		if !ok || job == nil || job.PodGroup == nil {
+			return
+		}
+		queueID := job.Queue
+		attr := cp.queueOpts[queueID]
+		if attr == nil {
+			return
+		}
+		// Mirror the JobEnqueueableFn guard: skip jobs with neither scalar nor DRA resources.
```
hajnalmt: Delete the comment please 🗡️ The code speaks for itself.

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -1959,3 +1959,123 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
...
+	test.Run([]framework.Action{enqueue.New()})
+	if err := test.CheckAll(0); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// TestJobEnqueuedUpdatesDRAInqueueForDRAOnlyJob verifies that JobEnqueuedFn correctly
+// updates DRA inqueue accounting for jobs that have no scalar MinResources but do have
+// DRA resource requests. The early return guard must not skip DRA-only jobs.
+func TestJobEnqueuedUpdatesDRAInqueueForDRAOnlyJob(t *testing.T) {
```
hajnalmt, line 2019: Thanks 👍 Good test.

hajnalmt (changes requested):
@Aman-Cool 
Thanks for addressing the request. The code looks good from my side👍 Please delete that comment I marked. The Guard Clause you introduced is perfect.

Please rebase on master and squash the commits! And ping us if we can let this in. 😉

Aman-Cool:
Hey @hajnalmt, addressed all the feedback..;deleted the comment, squashed all commits down to 2 (fix + tests), and rebased on latest master. lmk if anything else is needed!
@JesseStutler and @hzxuzhonghu, would love your thoughts too...

hajnalmt (approved):
Perfect, thank you! Giving others a chance to take a look.

hajnalmt:
Okay, I've tried every permutation and prow is not removing the label (tried /remove tide/merge-method-squash; /remove label tide/merge-method-squash and /remove label/tide/merge-method-squash; none of it worked).
Could someone help me how to remove the `tide/merge-method-squash` label?

Aman-Cool:
@hajnalmt, have you tried turning it off and on again? 😭 jokes aside, maybe try `/label remove tide/merge-method-squash`

hajnalmt:
/label remove tide/merge-method-squash

🤞

hajnalmt:
> @hajnalmt, have you tried turning it off and on again? 😭 jokes aside, maybe try `/label remove tide/merge-method-squash`

Didn't work 😞 but thanks

JesseStutler:
> Okay, I've tried every permutation and prow is not removing the label (tried /remove tide/merge-method-squash; /remove label tide/merge-method-squash and /remove label/tide/merge-method-squash; none of it worked). Could someone help me how to remove the `tide/merge-method-squash` label?

Let me help this!

JesseStutler:
/lgtm
Cool, thanks for your work

PR #5342, opened 2026-05-26
fix(jobcontroller): skip PodGroup deletion for targeted pod restarts

### LINKED ISSUE #5385: jobcontroller: PodGroup deleted on targeted pod restart causing job stuck in Restarting
### Description

`killPods` unconditionally deletes the job's PodGroup even when called for a targeted pod restart (`RestartPodAction`). The PodGroup should only be deleted on full job kills.

1. Create a Volcano job with `minAvailable=2` and 2 replicas where one pod is configured to fail (e.g. `exit 1` on a specific hostname).
2. Let the job start : both pods `Running`.
3. Observe the failing pod trigger `RestartPodAction`.
4. Watch the job status.

**Received:** `PodGroup` is deleted and recreated as `Pending`. `syncJob` sees `Pending`, sets `syncTask=false`, skips pod creation. Job gets stuck in `Restarting` until the scheduler completes a full re-enqueue cycle (can take several seconds to minutes).

**Expected:** `PodGroup` stays alive. The replacement pod is created immediately and the job recovers in seconds without a re-enqueue cycle.

master

**Affects** any job using `RestartPodAction` (e.g. policies with `action: RestartPod` on task failure). The bug is in `pkg/controllers/job/job_controller_actions.go` in the `killPods` function ; the `PodGroup` deletion block has no guard for the targeted-restart path (`target != nil`).

Aman-Cool:
**What type of PR is this?**
/kind bug

When a pod fails and `RestartPodAction` fires, the controller calls `killTarget` -> `killPods` to delete just that one pod. The problem: `killPods` unconditionally deleted the entire `PodGroup` at the end, even for single-pod targeted kills.

That triggered a full delete→recreate→re-enqueue→re-schedule cycle. While the new `PodGroup` was sitting in `Pending`, `syncJob` would see `pg.Status.Phase == Pending`, set `syncTask = false`, and skip pod creation entirely. The job status was stuck with `Terminating > 0` from the killed pod, so `restartingUpdateStatus` kept computing `total − Terminating < MinAvailable` and refused to leave `Restarting`. The job just sat there until the scheduler eventually picked up the new `PodGroup`, which in CI could take long enough to blow the 10-minute e2e timeout.

Fix: guard the `PodGroup` deletion with `if target == nil` so only full `KillJob` calls remove it. `KillTarget` (`RestartPodAction`) keeps the `PodGroup` alive, `syncTask` stays `true`, the replacement pod is created immediately, and the job exits `Restarting` in seconds.

Fixes #5385 

The change is a single `if target == nil` guard around the existing `PodGroup` deletion block in `killPods`. Full job kills pass `target == nil` and behave exactly as before. Only targeted kills (`RestartPodAction`, `RestartTaskAction`, `RestartPartitionAction`) are affected, they now preserve the `PodGroup` instead of deleting it.

A unit test for this regression would be good, happy to add one here if preferred.

Fix jobs using `RestartPodAction` getting stuck in `Restarting` state after a pod failure due to unnecessary `PodGroup` deletion in `killTarget`.

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -263,17 +263,21 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 		return e
 	}
 
-	// Delete PodGroup
-	pg, err := cc.getPodGroupByJob(job)
-	if err != nil && !apierrors.IsNotFound(err) {
-		klog.Errorf("Failed to find PodGroup of Job: %s/%s, error: %s", job.Namespace, job.Name, err.Error())
-		return err
-	}
-	if pg != nil {
-		if err := cc.vcClient.SchedulingV1beta1().PodGroups(job.Namespace).Delete(context.TODO(), pg.Name, metav1.DeleteOptions{}); err != nil {
-			if !apierrors.IsNotFound(err) {
-				klog.Errorf("Failed to delete PodGroup of Job %s/%s: %v", job.Namespace, job.Name, err)
-				return err
+	// Delete PodGroup only on full job kills; targeted pod restarts (RestartPodAction)
+	// must keep the PodGroup alive so the scheduler does not need to re-enqueue and
+	// re-schedule the job from scratch on every individual pod failure.
+	if target == nil {
+		pg, err := cc.getPodGroupByJob(job)
+		if err != nil && !apierrors.IsNotFound(err) {
+			klog.Errorf("Failed to find PodGroup of Job: %s/%s, error: %s", job.Namespace, job.Name, err.Error())
+			return err
+		}
+		if pg != nil {
+			if err := cc.vcClient.SchedulingV1beta1().PodGroups(job.Namespace).Delete(context.TODO(), pg.Name, metav1.DeleteOptions{}); err != nil {
+				if !apierrors.IsNotFound(err) {
+					klog.Errorf("Failed to delete PodGroup of Job %s/%s: %v", job.Namespace, job.Name, err)
+					return err
+				}
 			}
 		}
 	}
--- a/pkg/controllers/job/job_controller_actions_test.go
+++ b/pkg/controllers/job/job_controller_actions_test.go
@@ -26,7 +26,9 @@ import (
 
 	"github.com/agiledragon/gomonkey/v2"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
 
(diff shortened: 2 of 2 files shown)


Aman-Cool:
Hey @hajnalmt, opened this PodGroup deletion fix as a separate PR as you suggested... 
Let me know if you'd like the unit test added there too.

JesseStutler:
What is this background? I haven't received any similar issue feedback. Could you provide the vcjob yamls that could reproduce your bug?

Aman-Cool:
@JesseStutler, This was found while working on PR #5339 (the inqueue phantom-reservation fix). Tracing what forces jobs back through the scheduler's enqueue phase unexpectedly, I landed on `killPods` in `job_controller_actions.go`. Both `KillJob` (full kill, `target == nil`) and `KillTarget` (single-pod restart, `target != nil`) go through `killPods`, and `killPods` was unconditionally deleting the `PodGroup` at the end ; even for targeted kills. That forces a full `Pending` -> re-enqueue -> re-schedule cycle on every individual pod failure, which is the root cause of jobs getting stuck in `Restarting`.

Minimal repro:
```yaml
kind: Job
metadata:
  name: repro-restart-stuck
spec:
  minAvailable: 2
  schedulerName: volcano
  policies:
    - event: PodFailed
      action: RestartPodAction
  tasks:
    - name: worker
      replicas: 2
      template:
        spec:
          restartPolicy: Never
          containers:
            - name: worker
              image: busybox
              command: ["/bin/sh", "-c"]
              args:
                - |
                  if [ "$(hostname)" = "repro-restart-stuck-worker-0" ]; then
                    exit 1
                  fi
                  sleep 600
```

Without the fix: `worker-0` fails -> `RestartPodAction` -> `killPods` deletes the `PodGroup` -> new `PodGroup` recreated as `Pending` -> `syncJob` sees `Pending`, sets `syncTask=false`, skips pod creation -> `restartingUpdateStatus` sees `total(2) − Terminating(1) = 1 < minAvailable(2)` -> job stuck in `Restarting` until the scheduler completes the full re-enqueue cycle.

With the fix: `PodGroup` stays alive, `syncTask` stays `true`, replacement pod created in seconds.

hzxuzhonghu:
/lgtm

This is a minor improvement. Can you add a ut

hajnalmt (approved):
This is good from my side too! Thanks for the UT too! Can you create an issue for this too and mention it in the PR description Fixes part?

hajnalmt (commented):
/lgtm
Awesome, Thanks!

hajnalmt (approved):

hzxuzhonghu:
This failure maybe related to recent change https://github.com/volcano-sh/volcano/pull/5375 

@JesseStutler

JesseStutler:
@Aman-Cool Hi could you rebase the latest code and push again?

Aman-Cool:
The two failing checks (`e2e-cronjob` and `e2e-sequence`) appear to be pre-existing flakiness unrelated to this PR's changes... the cronjob failure is a runner timeout/shutdown signal, and the sequence failure is a cleanup race in `DeleteQueue` during the queue status transition test. Neither touches the code path changed here.

hajnalmt (approved):
Looks good from my side. Thanks!

JesseStutler:
/approve
Thanks!

PR #4704, opened 2025-10-29
Update volcano k8s dependencies to v1.34 again

suyiiyii:
What type of PR is this?
feature

What this PR does / why we need it:
This PR upgrades Kubernetes to v1.34, replacing the reference to Struct with interfaces in the new version.
It includes necessary code, YAML, and Docker changes to support this transition.

Which issue(s) this PR fixes:
fixes https://github.com/volcano-sh/volcano/issues/4671

Special notes for your reviewer:
Does this PR introduce a user-facing change?

original pr: #4687

PR #5343, opened 2026-05-27
fix(scheduler): restore ImageStates in GenerateNodeMapAndSlice

kitianFresh:
#### What type of PR is this?

/kind bug

Restores the `ImageStates` assignment in `GenerateNodeMapAndSlice` that was accidentally dropped during a code move.

The `ImageStates` field was originally added to `GenerateNodeMapAndSlice` in commit 093d07e7 (`plugins/util/util.go`), but was accidentally lost when the function was moved to `framework/util.go` in commit 463cc4ce ("move time consuming function into framework").

Without this line, `nodeInfo.ImageStates` is always an empty map when passed to the k8s scoring plugins, causing the **ImageLocality** scoring plugin to always return 0 for every node. This effectively disables image-locality-based scheduling — nodes that already have the required container images get no scoring advantage.

**Root cause:**

```
plugins/util/util.go (commit 093d07e7, 2022-12-06):
    nodeInfo.SetNode(node.Node)
+   nodeMap[node.Name].ImageStates = node.CloneImageSumary()

framework/util.go (commit 463cc4ce, 2022-12-21, moved but dropped the line):
    nodeInfo.SetNode(node.Node)
    // ← ImageStates assignment missing
    nodeMap[node.Name] = nodeInfo
```

**Fix:** Add back `nodeInfo.ImageStates = node.CloneImageSummary()` after `SetNode`.

Fixes the regression where ImageLocality scoring always returns 0.

- The data pipeline is: `kubelet → node.Status.Images → Informer → cache.AddOrUpdateNode() → addNodeImageStates() → api.NodeInfo.ImageStates ✅ → Snapshot.Clone() ✅ → Session.Nodes ✅ → GenerateNodeMapAndSlice() ❌ (was empty) → ImageLocality.Score() returns 0`
- The fix restores the last missing link in this chain.
- A unit test is added to verify ImageStates are correctly propagated through `GenerateNodeMapAndSlice`.

```release-note
Fix ImageLocality scoring plugin always returning 0 due to ImageStates not being propagated in GenerateNodeMapAndSlice.
```

### DIFF
--- a/pkg/scheduler/framework/util.go
+++ b/pkg/scheduler/framework/util.go
@@ -228,6 +228,7 @@ func GenerateNodeMapAndSlice(nodes map[string]*api.NodeInfo) map[string]fwk.Node
 	for _, node := range nodes {
 		nodeInfo := k8sframework.NewNodeInfo(node.Pods()...)
 		nodeInfo.SetNode(node.Node)
+		nodeInfo.ImageStates = node.CloneImageSummary()
 		nodeMap[node.Name] = nodeInfo
 	}
 	return nodeMap
--- a/pkg/scheduler/framework/util_test.go
+++ b/pkg/scheduler/framework/util_test.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2025 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fwk "k8s.io/kube-scheduler/framework"
+
+	"volcano.sh/volcano/pkg/scheduler/api"
+)
+
+func TestGenerateNodeMapAndSlice_ImageStates(t *testing.T) {
+	node := &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "node1"},
+		Status: v1.NodeStatus{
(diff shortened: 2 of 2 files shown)


kitianFresh:
@dafu-wu hello，please review this bugfix

dafu-wu:
Thanks,LGTM

pkg/scheduler/framework/util.go
```diff
@@ -228,6 +228,7 @@ func GenerateNodeMapAndSlice(nodes map[string]*api.NodeInfo) map[string]fwk.Node
 	for _, node := range nodes {
 		nodeInfo := k8sframework.NewNodeInfo(node.Pods()...)
 		nodeInfo.SetNode(node.Node)
+		nodeInfo.ImageStates = node.CloneImageSummary()
```
JesseStutler: I have a question that, since previously volcano has already done CloneImageSummary during Snapshot for volcano own's NodeInfo: https://github.com/volcano-sh/volcano/blob/e6136df50762c48a31ed16134b2e667506c73c98/pkg/scheduler/api/node_info.go#L242, do we need to DeepCopy again here? I don't think volcano's own NodeInfo will operate the ImageStates

kitianFresh: You're right. The `CloneImageSummary()` during Snapshot (in `node_info.go:242`) already created an isolated deep copy from the cache. After that point, the session's `api.NodeInfo.ImageStates` is read-only within the scheduling cycle — no one mutates it. The k8s framework `NodeInfo` created here is also read-only (only consumed by `ImageLocality.Score()`). So a second deep copy is unnecessary overhead.

I'll change this to a direct assignment:
```go
nodeInfo.ImageStates = node.ImageStates
```

Thanks for the review!

JesseStutler:
@kitianFresh Please sign off your commit

kitianFresh:
> @kitianFresh Please sign off your commit

Done

JesseStutler:
I noticed you changed your code to Clone again 😅, but Never Mind, it shouldn't have a significant performance impact. The number of fields isn't very large. We at least need to ensure that ImageLocality works correctly. I won't block it just because of strict requirements.

JesseStutler:
/cherrypick release-1.15
/cherrypick release-1.14
https://github.com/volcano-sh/volcano/pull/4704 accidentally removed it and we didn't find, thanks for your contribution but we only need to backport to v1.15 and v1.14, v1.13 contains it

PR #5344, opened 2026-05-27
fix: guard podgroup annotation patch by pod UID

### LINKED ISSUE #5331: Stale PodGroup may be patched to newly recreated Pods after StatefulSet recreation, causing Pods to remain Pending permanently
### Description

We encountered a race condition between PodGroup lifecycle and StatefulSet recreation when running Notebook workloads with Volcano scheduling.

This problem is not necessarily reproducible.

1. notebook-controller creates a StatefulSet
2. StatefulSet controller creates Pods
3. volcano-controller watches Pod AddEvents and creates PodGroups (PG) for these Pods
4. The PodGroup ownerReference points to the StatefulSet

Then the following sequence happens:
1. The old StatefulSet is deleted
2. Old Pods are not fully deleted yet
3. A new StatefulSet with the same name is created
4. The new StatefulSet creates new Pods with the same Pod names as the old ones
5. The old PodGroup is patched/reused for the new Pods

StatefulSet Pod names are stable (for example: notebook-0)

    There is a race window between old Pod deletion and new Pod creation
During this window, the PodGroup originally created for the old Pod may be patched onto the newly created Pod because the Pod names match. The PodGroup ownerReference still points to the old StatefulSet.Once the old StatefulSet is garbage collected, the PodGroup is deleted through cascading deletion.  However, the new Pod already contains:scheduling.k8s.io/group-name，volcano-controller does not create a new PodGroup for this pod anymore.

The Pod should eventually be scheduled successfully even if the StatefulSet is recreated.

1.11.0

_No response_

YiD11:
#### What type of PR is this?

/kind bug

When a StatefulSet in Volcano is deleted and then recreated with the same namespace/name, the PodGroup controller may still have an old Pod request in the queue.

The `podRequest` only contains namespace and name. If the worker handles it after the Pod is recreated, the cache lookup can return the new Pod. Then the old PodGroup name may be patched into the new Pod's `scheduling.k8s.io/group-name`.

If that old PodGroup is garbage-collected, the new Pod keeps a annotation pointing to a missing PodGroup and can stay Pending.

This PR adds the Pod UID field in `podRequest` and skips queued work when the cached Pod UID mismatches the request. And it changes the annotation update to a JSON Patch with a `/metadata/uid` test, so the API server should reject the patch if the Pod was recreated after the controller read it from cache.

#### Which issue this PR fixes:

Fixes #5331

The original issue was reproduced with StatefulSet delete/recreate loops on Volcano v1.11.0. Details are shown in the original issue page.

NONE

### DIFF
--- a/pkg/controllers/podgroup/pg_controller_handler.go
+++ b/pkg/controllers/podgroup/pg_controller_handler.go
@@ -46,14 +46,7 @@ const (
 type podRequest struct {
 	podName      string
 	podNamespace string
-}
-
-type metadataForMergePatch struct {
-	Metadata annotationForMergePatch `json:"metadata"`
-}
-
-type annotationForMergePatch struct {
-	Annotations map[string]string `json:"annotations"`
+	podUID       types.UID
 }
 
 func (pg *pgcontroller) addPod(obj interface{}) {
@@ -66,6 +59,7 @@ func (pg *pgcontroller) addPod(obj interface{}) {
 	req := podRequest{
 		podName:      pod.Name,
 		podNamespace: pod.Namespace,
+		podUID:       pod.UID,
 	}
 
 	pg.queue.Add(req)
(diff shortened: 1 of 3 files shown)


JesseStutler:
Will take a look, thanks for your contribution.
/cc
/ok-to-test

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -187,35 +181,46 @@ func (pg *pgcontroller) updateStatefulSet(oldObj, newObj interface{}) {
...
-			return err
-		}
-		klog.V(4).Infof("Bound Pod <%s/%s> to PodGroup <%s/%s>", pod.Namespace, pod.Name, pod.Namespace, pgName)
+		})
+	} else if pod.Annotations[scheduling.KubeGroupNameAnnotationKey] == "" {
+		patch = append(patch, map[string]any{
+			"op":    "add",
+			"path":  "/metadata/annotations/" + strings.ReplaceAll(scheduling.KubeGroupNameAnnotationKey, "/", "~1"),
+			"value": pgName,
+		})
```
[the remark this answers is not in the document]
YiD11: done on new commit

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -187,38 +181,62 @@ func (pg *pgcontroller) updateStatefulSet(oldObj, newObj interface{}) {
...
-			return err
-		}
-
-		if _, err := pg.kubeClient.CoreV1().Pods(pod.Namespace).Patch(context.TODO(), pod.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{}); err != nil {
-			klog.Errorf("Failed to update pod <%s/%s>: %v", pod.Namespace, pod.Name, err)
-			return err
+	// UID test keeps this patch from hitting a recreated pod with the same namespace/name.
+	patch := []map[string]any{
+		{
+			"op":    "test",
```
JesseStutler, line 187: I haven't seen this usage very often. Can you explain why it's written this way?

JesseStutler:
I'm wondering that in #5331,  "The old PodGroup is patched/reused for the new Pods", I don't know whether we will still meet it in latest codes since the podgroup name is got from pod's ownerreference, which contains new statefulset's UID, so the new pod won't be patched the old pg name: https://github.com/volcano-sh/volcano/blob/fdb6ae056ea2798e027c5951df9e49fc7a5624f4/staging/src/volcano.sh/apis/pkg/apis/helpers/helpers.go#L192-L208

pkg/controllers/podgroup/pg_controller_handler.go
YiD11: Because the patch request comes from the worker queue, it may become stale if the Pod is deleted and recreated while the request is still queued.

The purpose of using JSON Patch with a `/metadata/uid` `test` is to prevent this stale patch. If the live Pod UID is different from the UID in the patch, the API server will reject the patch.

YiD11:
> I'm wondering that in #5331, "The old PodGroup is patched/reused for the new Pods", I don't know whether we will still meet it in latest codes since the podgroup name is got from pod's ownerreference, which contains new statefulset's UID, so the new pod won't be patched the old pg name:
> 
> https://github.com/volcano-sh/volcano/blob/fdb6ae056ea2798e027c5951df9e49fc7a5624f4/staging/src/volcano.sh/apis/pkg/apis/helpers/helpers.go#L192-L208

The code in this commit can reduce the probability of race conditions, but I was still able to reproduce the issue on this commit in an extreme case.

I temporarily deployed an admission webhook to delay only requests for podgroup creation, then recreated the StatefulSet/Pod with the same names before releasing the old request.

The pseudocode of the reproduction is:

```text
hold requests for podgroup creation 

create StatefulSet notebook
oldPod = get Pod notebook-0
oldOwnerUID = oldPod.ownerReferences[0].uid

wait until Volcano starts creating podgroup-${oldOwnerUID}

delete StatefulSet notebook
force delete Pod notebook-0
create StatefulSet notebook again

newPod = wait for Pod notebook-0
assert newPod.uid != oldPod.uid
assert newPod.ownerReferences[0].uid != oldOwnerUID

release requests for podgroup creation 

newPod = get Pod notebook-0
assert newPod.annotations["scheduling.k8s.io/group-name"] == "podgroup-${oldOwnerUID}"
assert podgroup "podgroup-${oldOwnerUID}" is NotFound
```

The final live Pod was owned by the new StatefulSet, but its `scheduling.k8s.io/group-name` still pointed to the old podgroup name. This stale patch may happen when podgroup creation is delayed by admission webhooks, apiserver/etcd latency, or controller queue backlog while a same-name StatefulSet Pod is recreated. 

This is the stale patch I think the `/metadata/uid` `test` can prevent.

YiD11:
Hi @JesseStutler , I think I have replied your question on Jun 15, about the reproduction details and why the `/metadata/uid` JSON Patch `test` is needed.
According to my reproduction, I found this problem still existed in [fdb6ae0](https://github.com/volcano-sh/volcano/commit/fdb6ae056ea2798e027c5951df9e49fc7a5624f4).
Almost all checks are green now. Could you please take another look when you get a chance?

YiD11:
Hi @JesseStutler , I've updated the PR based on your comment. I changed the patch strategic, using strategic merge patch that carried `metadata.uid`. So the API server rejects it if the pod was recreated. Thanks for reviewing the new commit again.

JesseStutler:
> Hi @JesseStutler , I've updated the PR based on your comment. I changed the patch strategic, using strategic merge patch that carried `metadata.uid`. So the API server rejects it if the pod was recreated. Thanks for reviewing the new commit again.

OK thanks! Will take a look

pkg/controllers/podgroup/pg_controller_test.go
```diff
@@ -1415,3 +1423,94 @@ func TestNoPodGroupCreatedForWhenKubeGroupNameAnnotationExists(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Empty(t, pgList.Items)
+	assert.Equal(t, 0, c.queue.Len())
+
+	for _, action := range c.kubeClient.(*kubeclient.Clientset).Actions() {
+		assert.False(t, action.Matches("patch", "pods"))
+	}
+}
+
+func TestUpdatePodAnnotationsCarriesUID(t *testing.T) {
```
JesseStutler, line 1461: The UID guard lgtm

But Could we also add coverage through the controller’s normal processing path, rather than only testing the annotation helper directly?

`TestUpdatePodAnnotationsCarriesUID` calls the helper directly with a reactor that always succeeds, so it does not cover PodGroup creation, patch failure, or requeue behavior.

I’d suggest table-driven tests that enqueue requests through addPod and process them through `processNextReq`, covering matching UIDs, stale requests, and patch failures. We should also have a regression scenario where the old Pod’s patch is rejected, then the recreated Pod’s request is processed and bound to the PodGroup derived from its new owner UID.

Manually setting up the informer cache and stepping the worker is fine—we don’t need a fully asynchronous controller test. The important part is to exercise the actual processing path and assert the resulting Pod/PodGroup state and retry behavior, rather than only inspecting the patch payload.

YiD11:
Hi @JesseStutler, I've reworked the tests as you suggested. I created a new file `pg_controller_recreated_pod_test.go` to complete all tests of my change in podgroup controller, and the old test is removed.

All cases go through addPod/processNextReq. Since the fake clientset doesn't validate metadata.uid, I added a patch reactor that rejects a uid mismatch the same way the apiserver does. The table test covers a matching uid, a stale request (cache already updated, and cache still stale), and a patch failure with retry, checking the pod annotation, the podgroup owner and the requeue count. Two more tests replay the original bug: one drives the old and the new pod requests in order and checks which podgroup the new pod ends up in, the other does the same through addStatefulSet.

Hope this is what you had in mind. Thanks for the review.

JesseStutler:
/approve
/lgtm
Thanks

Issue #5347, opened 2026-05-28
[Feature] PodGroup-level inter-group topology scheduling (group-topology-affinity)


wangyang0616:
### What is the problem you're trying to solve

**Background：**
Teams running large-model training and inference on Volcano can already colocate communication-heavy pods with `networkTopology`, but still struggle to express relationships between groups (multiple serving instances, Prefill-Decode shards, roles) in one place.

- **Platform / SRE — smaller blast radius:** multiple inference instances can be pinned to different supernodes via `topologyGroup`, so one supernode failure does not take down every replica.
- **Inference / PD — latency and stability:** shards on separate cabinets for fault isolation, whole instance on one supernode, without hand-written Pod spread/anti-affinity.
- **Workload owners — simpler config:** one PodGroup spec (instance, shard, role); scales with `matchLabelKeys`; webhook-validatable.
- **Operators — easier debug:** inter-group and intra-group rules use the same HyperNode path as NTAS, fewer ambiguous Pending states.

**In short:** higher availability for multi-instance serving, more predictable PD placement, lower operational cost — on top of existing intra-group topology and Gang.

**Current Situation：**
Volcano supports intra-group topology (HyperNode, `network-topology-aware`, `networkTopology`, `subGroupPolicy`, multi-dimensional Gang): communication-intensive pods colocate in a lower tier (e.g. cabinet) while the workload stays in a higher envelope (e.g. supernode).

We lack a first-class inter-group topology API on PodGroup:

- **Cross-PodGroup anti-affinity:** N inference instances (N PodGroups) should each use a different supernode.
- **Cross-SubJob rules in one PodGroup:** PD with multiple shards — shards on different cabinets, whole instance on one supernode, in-shard Gang.
- **Unified scheduling:** inter-group rules in the same HyperNode gradient / allocate dry-run path as `networkTopology`, not ad-hoc Pod templates.

Related (different scope): #4188 (intra-job SubJob topology), #5078 (inference spread vs binpack), #3317 (task-topology affinity groups).

### Describe the solution you'd like

New plugin `group-topology-affinity`, composable with `network-topology-aware`:

1. **topologyAffinity.podGroupAntiAffinity (Phase 1)**
   - Cross-PodGroup anti-affinity at a HyperNode tier (`topologyGroup` / `podGroupSelector`, `separationTierName` or `separationTier`).
   - No `podGroupAffinity` in Phase 1 (colocation via `networkTopology`).

[...]

PR #5349, opened 2026-05-28
feat(scheduler): support PodGroup topology anti-affinity

wangyang0616:
#### What type of PR is this?
/kind feature

This PR implements the first phase of group topology affinity: PodGroup-level topology anti-affinity on the HyperNode tree.

- Supports required and preferred `podGroupAntiAffinity`.
- Adds the `group-topology-affinity` plugin and composes it with `network-topology-aware`.
- Adds API validation, HyperNode placement refresh, and scheduling diagnostics.
- Adds unit and E2E coverage, including combined network-topology scenarios.

The plugin must be explicitly enabled in the scheduler configuration. SubGroup affinity/anti-affinity and backfill support are outside this phase.

**Related issue:** https://github.com/volcano-sh/volcano/issues/5347

```release-note
Support required and preferred PodGroup topology anti-affinity across HyperNode domains.
```


docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+| Section | Content |
+|---------|---------|
+| [Overview](#overview) | Background, community status, goals, scope |
+| [User Stories](#user-stories) | Typical requirements (examples 1–7); [Design Decisions at a Glance](#design-decisions-at-a-glance) |
+| [User Scenarios and Capability Mapping](#user-scenarios-and-capability-mapping) | Examples 1–7 with full YAML |
+| [API Design](#api-design) | Types, capability boundaries, HyperNode, appendix |
+| [Design Decisions](#design-decisions) | Design decisions 1–7 |
+| [Scheduling Implementation](#scheduling-implementation) | Plugins, Framework, allocate |
+| [Architecture and Sequence Diagrams](#architecture-and-sequence-diagrams) | Overview diagrams and allocate path |
+| [Validation Rules](#validation-rules-webhook), [Delivery Phases](#delivery-phases), [Status](#status-optional), [References](#references) | Webhook, Phase, Condition |
```
hzxuzhonghu: This part be just index instead of table, i donot see a template in volcano

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+
+| Current Gap | Typical Consequence |
+|-------------|---------------------|
+| No declarative **cross-PodGroup** topology **anti-affinity** (mutual exclusion) | Multiple inference instances land on the same supernode; a single point of failure takes down all online replicas |
+| No declarative **cross-SubGroup topology within the same PodGroup** | Prefill-Decode shards cannot achieve "shards per cabinet, entire instance on one supernode"; must split into multiple PodGroups or hard-code rules outside PodGroup |
+| **Intra-group and inter-group not layered** | Inter-group requirements mixed with `networkTopology` or rely on ops conventions; not on the same configuration surface as HyperNode intra-group scheduling, hard to validate |
```
hzxuzhonghu: The description is not clear, hard to understand

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+
+This design **does not replace** the existing `network-topology-aware` plugin; it adds **inter-group** topology scheduling:
+
+- **Cross PodGroup**: only **`topologyAffinity.podGroupAntiAffinity`** (examples 2/5); **no** `podGroupAffinity` (see [Design Decision-2](#ad-2-phase-1-no-podgroupaffinity-in-crd)).
```
hzxuzhonghu: examples 2/5 add a  hyperlink

hzxuzhonghu: This is under Background and Problem section, bu if i understand correctly, isn't this the new design. SHould not occur in the background

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+- **Cross SubGroup within the same PodGroup**: `subGroupTopologyAffinity` (examples 4/6/7).
```
hzxuzhonghu: same above

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+
+| Capability | Configuration / Component |
+|------------|---------------------------|
+| Multi-level network topology | HyperNode CRD |
+| Gang / SubJob splitting | PodGroup, `subGroupPolicy`, `matchLabelKeys` |
+| **Intra-group** topology aggregation + Gang | `networkTopology` + `network-topology-aware`, with `minMember` / `subGroupPolicy` / `subGroupSize`, etc. |
+
+**Already covered (intra-group):** For **training, inference**, and other communication-intensive workloads — at **SubJob (or communication subgroup)** granularity, aggregate a batch of Pods **within the same lowest topology domain** (e.g., high-bandwidth domain within a cabinet); at **PodGroup / Job** granularity, **constrain the entire workload within a higher topology domain** (e.g., supernode envelope defined by `highestTierAllowed`, entire machine not crossing domains); coordinated with **multi-dimensional Gang** (PodGroup `minMember`, `subGroupPolicy` + `subGroupSize` / `minSubGroups`, `matchLabelKeys` for SubJob splitting, etc.) to **bind after group resource and topology requirements are satisfied**. See [Network Topology Aware Scheduling](./Network%20Topology%20Aware%20Scheduling.md).
```
hzxuzhonghu: isn't this duplicate with L33-L37

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+
+### What Volcano Lacks (Inter-Group)
```
hzxuzhonghu: seems also duplicate with L39-L43, merge together

hzxuzhonghu: When it is describing the problem currect api faces, no need to show the new design yaml. 

I would suggest propose using a simple temple like https://github.com/kmesh-net/kmesh/blob/main/docs/proposal/template.md https://github.com/kubernetes/enhancements/tree/master/keps/sig-architecture/1194-prod-readiness

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+
+| Gap | Typical Scenario (Examples in This Doc) | User Impact |
+|-----|---------------------------------------|-------------|
+| **Cross-PodGroup topology anti-affinity** | [Example 2](#example-2-multi-inference-instance-fault-isolation), [5](#example-5-multi-instance-pd-combination) | Multiple inference **instances** may land on the same supernode; single point of failure has large blast radius |
+| **Cross-SubJob topology edges within the same PodGroup** | [Example 4](#example-4-distributed-prefill-decode-inference-recommended), [6](#example-6-optional-prefill-decode-cross-role-cabinet-split), [7](#example-7-subgroup-soft-anti-affinity-optional) | Multi-shard Prefill-Decode cannot declare "shards per cabinet, entire instance on one supernode" combinations |
```
hzxuzhonghu: `rack` is more commonly used, did not see cabinet before

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+| **allocate resource pre-filter + Webhook** | Same path as existing topology scheduling; configuration is validatable |
+| **Phased P1 → P2** | P1 main-path scheduling; P2 preempt/backfill consistent with occupancy (see [Delivery Phases](#delivery-phases)) |
+
+### Goal Illustration (Volcano Perspective)
+
+**Three layers of capability (intra-group + inter-group):**
+
+```mermaid
+flowchart TB
+    subgraph layer_intra_group ["Intra-group · existing network-topology-aware"]
```
hzxuzhonghu: Intra-group/subgroup, no subjob within existing podgroup api

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+        network_topology_field["PodGroup / SubJob networkTopology"]
+    end
+    subgraph layer_same_podgroup ["Same PodGroup inter-group · this design"]
```
hzxuzhonghu: inter-subgroup

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+        sub_group_topology_affinity_field["subGroupTopologyAffinity"]
+    end
+    subgraph layer_cross_podgroup ["Cross PodGroup · this design"]
+        topology_affinity_anti["topologyAffinity.podGroupAntiAffinity"]
```
hzxuzhonghu: I am wondering if there is affinity requirement across podgroups

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+```go
+type PodGroupSpec struct {
+    // ... existing fields ...
+
+    // TopologyAffinity: cross-PodGroup topology vs OTHER PodGroups (podGroupSelector).
+    // Phase 1 CRD exposes only podGroupAntiAffinity under this object; podGroupAffinity is not declared (see design doc).
+    // Intra-PodGroup SubJob relationships: use SubGroupTopologyAffinity.
+    // Evaluated at Job scope (HyperNodeGradientForJobFn).
+    // +optional
+    TopologyAffinity *PodGroupTopologyAffinitySpec `json:"topologyAffinity,omitempty"`
```
hzxuzhonghu: We have `NetworkTopology`, and now add both `TopologyAffinity`  and `SubGroupTopologyAffinity`. It is quite confusing now.

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+        antiSubGroupSelector:
+          matchSubGroupPolicyNames: [prefill]
+        topologyTierName: cabinet     # or topologyTier: 1
+```
+
+```go
+// PodGroupTopologyAffinitySpec expresses topology constraints between THIS PodGroup and OTHER PodGroups
+// podGroupSelector (metav1.LabelSelector). Evaluated at Job scope (HyperNodeGradientForJobFn).
+//
+// Phase 1 scope: ONLY PodGroupAntiAffinity is implemented and exposed in the CRD.
```
hzxuzhonghu: This should not exist in api doc

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+    PreferredDuringSchedulingIgnoredDuringExecution []WeightedSubGroupTopologyAffinityTerm `json:"preferredDuringSchedulingIgnoredDuringExecution,omitempty"`
+}
+
+type TopologyAntiAffinitySpec struct {
+    RequiredDuringSchedulingIgnoredDuringExecution  []TopologyAntiAffinityTerm `json:"requiredDuringSchedulingIgnoredDuringExecution,omitempty"`
+    PreferredDuringSchedulingIgnoredDuringExecution []WeightedTopologyAntiAffinityTerm `json:"preferredDuringSchedulingIgnoredDuringExecution,omitempty"`
+}
+
+// Cross-PodGroup terms (anti-affinity; Phase 1)
+type TopologyAntiAffinityTerm struct {
```
hzxuzhonghu:
```suggestion
type PodGroupAntiAffinityTerm struct {
```

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+    // PodGroupSelector: match OTHER PodGroups by metadata.labels (same semantics as kube-scheduler labelSelector).
+    // +required
+    PodGroupSelector  *metav1.LabelSelector `json:"podGroupSelector"`
+    // NamespaceSelector: optional scope for peer PodGroups (same role as in PodAntiAffinityTerm).
+    // +optional
+    NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+    TopologyDomain    TopologyDomainSpec `json:"topologyDomain"`
```
hzxuzhonghu: not sure what does this mean

docs/design/group-topology-affinity.md
hzxuzhonghu: This struct is not consistent with above yaml
```
topologyAffinity:
  podGroupAntiAffinity:
      - podGroupSelector:
            matchLabels:
              topology.volcano.sh/spread-group: llama-70b-prod
        topologyTierName: supernode   # A
        # topologyTier: 2             # B (equivalent to A when supernode corresponds to spec.tier==2)
```

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,2034 @@
...
+// SubGroupTopologyAffinitySpec: intra-PodGroup, cross-subGroupPolicy scope only.
+type SubGroupTopologyAffinitySpec struct {
+    SubGroupAntiAffinity *SubGroupAntiAffinitySpec `json:"subGroupAntiAffinity,omitempty"`
+    SubGroupAffinity     *SubGroupAffinitySpec     `json:"subGroupAffinity,omitempty"`
+}
+
+// SubGroupAntiAffinitySpec / SubGroupAffinitySpec: term lists for SubGroup peers (not PodGroup selectors).
+type SubGroupAntiAffinitySpec struct {
+    RequiredDuringSchedulingIgnoredDuringExecution  []SubGroupTopologyAntiAffinityTerm `json:"requiredDuringSchedulingIgnoredDuringExecution,omitempty"`
+    PreferredDuringSchedulingIgnoredDuringExecution []WeightedSubGroupTopologyAntiAffinityTerm `json:"preferredDuringSchedulingIgnoredDuringExecution,omitempty"`
```
hzxuzhonghu: Would recommend not  implement this at first

hzxuzhonghu (commented):
After reading through the current struct, i feel very confusing.

I will think carefully before  any further idea/suggestion

hzxuzhonghu:
After some deepdive, based on the existing proposal, i think we can only add one field in the podgroup spec

As to `TopologyTier`, i feel maybe its more clear to call HighestTier/LowestTier for afffinity/antiafffinity scenarios

```golang

TopologyAffinity *TopologyAffinitySpec

// TopologyAffinitySpec groups all inter-group topology constraints for a PodGroup.
type TopologyAffinitySpec struct {
    // PodGroupAntiAffinity forces this PodGroup to land in a DIFFERENT
    // topology domain than the matched peer PodGroups (typical use:
    // spread inference instances across supernodes for fault isolation).
    // +optional
    PodGroupAntiAffinity *PodGroupAntiAffinity `json:"podGroupAntiAffinity,omitempty" protobuf:"bytes,1,opt,name=podGroupAntiAffinity"`

    // SubGroupAffinity forces listed subGroupPolicy entries to share the
    // SAME topology domain (e.g. prefill+decode on one supernode).
    // +optional
    SubGroupAffinity *SubGroupAffinity `json:"subGroupAffinity,omitempty" protobuf:"bytes,2,opt,name=subGroupAffinity"`

    // SubGroupAntiAffinity forces listed subGroupPolicy entries (or shards
    // within one entry split by matchLabelKeys) to land in DIFFERENT
    // topology domains (e.g. each shard on its own rack).
    // +optional
    SubGroupAntiAffinity *SubGroupAntiAffinity `json:"subGroupAntiAffinity,omitempty" protobuf:"bytes,3,opt,name=subGroupAntiAffinity"`
}

```

Across podgroups

```golang
type PodGroupAntiAffinity struct {
    // Required: peer PodGroups MUST be in different topology domains.
    // If no domain satisfies the constraint, this PodGroup stays Pending.
    // +optional
    // +listType=atomic
    Required []PodGroupAffinityTerm `json:"required,omitempty" protobuf:"bytes,1,rep,name=required"`

    // Preferred: peer PodGroups SHOULD be in different topology domains;
    // scheduling still succeeds if the preference cannot be satisfied.
    // Each term may set Weight (1-100); higher = stronger preference.
    // +optional
    // +listType=atomic
    Preferred []PodGroupAffinityTerm `json:"preferred,omitempty" protobuf:"bytes,2,rep,name=preferred"`
}

// PodGroupAffinityTerm selects peer PodGroups and the tier at which their
// topology domain must differ from this one.
// The scheduler automatically EXCLUDES this PodGroup itself from peer matches.
type PodGroupAffinityTerm struct {
    // Weight applies only when this term appears under Preferred (1-100,
    // higher = stronger preference). Ignored under Required.
    // +kubebuilder:validation:Minimum=1
    // +kubebuilder:validation:Maximum=100
    // +optional
    Weight int32 `json:"weight,omitempty" protobuf:"bytes,5,opt,name=weight"`

    // PodGroupSelector matches other PodGroups by metadata.labels.
    // Standard metav1.LabelSelector semantics; an empty selector matches none
    // (use MatchLabels{} explicitly if you intend to match all peers).
    // +required
    PodGroupSelector *metav1.LabelSelector `json:"podGroupSelector" protobuf:"bytes,1,opt,name=podGroupSelector"`

    // NamespaceSelector limits peer PodGroups to matching namespaces.
    // nil = same namespace as this PodGroup. {} = all namespaces.
    // +optional
    NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty" protobuf:"bytes,2,opt,name=namespaceSelector"`

    // TopologyTierName: HyperNode.spec.tierName at which Domain_T is compared
    // (e.g. "supernode", "rack"). Mutually exclusive with TopologyTier.
    // +kubebuilder:validation:MaxLength=253
    // +optional
    TopologyTierName string `json:"topologyTierName,omitempty" protobuf:"bytes,3,opt,name=topologyTierName"`

    // TopologyTier: HyperNode.spec.tier integer at which Domain_T is compared.
    // Mutually exclusive with TopologyTierName. Exactly one of the two must be set.
    // +kubebuilder:validation:Minimum=0
    // +optional
    TopologyTier *int32 `json:"topologyTier,omitempty" protobuf:"bytes,4,opt,name=topologyTier"`
}
```

Across subgroups

```go
type SubGroupAffinity struct {
    // Required: the SubGroups in each term MUST share one domain.
    // +optional
    // +listType=atomic
    Required []SubGroupTerm `json:"required,omitempty" protobuf:"bytes,1,rep,name=required"`
    // Preferred: terms SHOULD be satisfied; may set Weight (1-100).
    // +optional
    // +listType=atomic
    Preferred []SubGroupTerm `json:"preferred,omitempty" protobuf:"bytes,2,rep,name=preferred"`
}

type SubGroupAntiAffinity struct {
    // Required: the SubGroups' SubJobs in each term MUST occupy distinct domains.
    // +optional
    // +listType=atomic
    Required []SubGroupTerm `json:"required,omitempty" protobuf:"bytes,1,rep,name=required"`
    // Preferred: terms SHOULD be satisfied; may set Weight (1-100).
    // +optional
    // +listType=atomic
    Preferred []SubGroupTerm `json:"preferred,omitempty" protobuf:"bytes,2,rep,name=preferred"`
}

// SubGroupTerm names a set of subGroupPolicy entries and the tier at which
// their topology domains are compared. The SAME type is used for both
// affinity (share one domain) and anti-affinity (occupy distinct domains);
// the meaning comes from whether the term sits under SubGroupAffinity or
// SubGroupAntiAffinity. See §3.5.
//
// Anti-affinity examples:
//   - subGroups: [prefill]            # prefill-0..N each in its own domain
//   - subGroups: [prefill, decode]    # all prefill AND decode SubJobs distinct
// Affinity example:
//   - subGroups: [prefill, decode]    # prefill + decode share one domain
type SubGroupTerm struct {
    // SubGroups names entries from spec.subGroupPolicy[].name.
    // +kubebuilder:validation:MinItems=1
    // +listType=atomic
    SubGroups []string `json:"subGroups" protobuf:"bytes,1,rep,name=subGroups"`

    // Weight applies only when this term appears under Preferred (1-100,
    // higher = stronger preference). Ignored under Required.
    // +kubebuilder:validation:Minimum=1
    // +kubebuilder:validation:Maximum=100
    // +optional
    Weight int32 `json:"weight,omitempty" protobuf:"bytes,2,opt,name=weight"`

    // TopologyTierName: HyperNode.spec.tierName at which domains are compared.
    // Mutually exclusive with TopologyTier.
    // +kubebuilder:validation:MaxLength=253
    // +optional
    TopologyTierName string `json:"topologyTierName,omitempty" protobuf:"bytes,3,opt,name=topologyTierName"`
    // TopologyTier: HyperNode.spec.tier at which domains are compared.
    // Mutually exclusive with TopologyTierName. Exactly one of the two must be set.
    // +kubebuilder:validation:Minimum=0
    // +optional
    TopologyTier *int32 `json:"topologyTier,omitempty" protobuf:"bytes,4,opt,name=topologyTier"`
}
```

hzxuzhonghu:
There are some generated yamls which can help understand the usage, please double check if you wanto migrate to you proposal, though it is using Opus4.8, but i only told it how to design the api struct.  FYI https://github.com/hzxuzhonghu/volcano/blob/design-group-topology-affinity-api/docs/design/group-topology-affinity-api.md#4-user-facing-yaml-examples

pmady:
interesting proposal. one thing worth calling out is how this composes with intra-node topology — e.g. after the group-topology-affinity plugin picks which supernode/rack a PodGroup lands on, the numaaware plugin still needs to schedule GPU tasks onto the right NUMA nodes within that host. might be worth adding a short section on interaction with the existing numaaware and network-topology-aware plugins so the scheduling pipeline ordering is clear.

also re: hzxuzhonghu's suggestion about HighestTier/LowestTier naming — i think that's clearer than the raw integer. operators configuring this probably think in terms of "supernode" or "rack" not tier numbers

wangyang0616:
> There are some generated yamls which can help understand the usage, please double check if you wanto migrate to you proposal, though it is using Opus4.8, but i only told it how to design the api struct. FYI https://github.com/hzxuzhonghu/volcano/blob/design-group-topology-affinity-api/docs/design/group-topology-affinity-api.md#4-user-facing-yaml-examples

Thank you very much for taking the time to review this part of the design. Referring to the API example you provided, it does indeed simplify usage considerably, and I'm refining this part of the design.

I'd also like to discuss a question. There's a scenario that the API currently can't express: for example, having two subgroups, we want anti-affinity between subgroup-A and subgroup-B, but multiple copies within subgroup-A don't need anti-affinity.

Analogous scenario: In a 4P2D inference scenario, we want anti-affinity between P and D, but not between the four Ps. Of course, this scenario doesn't seem necessary now, but I'm a bit unsure if it might be needed in the future.

pmady:
thanks for the pointer. looked at hzxuzhonghu's branch — the single TopologyAffinity field consolidating all three constraint types (PodGroupAntiAffinity, SubGroupAffinity, SubGroupAntiAffinity) is cleaner than spreading them across the spec. the HighestTier/LowestTier naming suggestion makes sense too, more operator-friendly than raw tier integers

wangyang0616:
I strongly agree with merging PodGroupAntiAffinity, SubGroupAffinity, and SubGroupAntiAffinity into a unified topologyAffinit structure. However, I have a different opinion regarding HigherTier/LowestTier; I prefer using TopologyTier to express the hierarchy on which affinity or anti-affinity acts.

Semantically, HigherTier/LowestTier is easier to understand, but this definition requires users to use different fields in affinity and antiaffinity to represent the hierarchy when creating podgroups. This lack of consistency makes it easier to make mistakes. In contrast, TopologyTier offers a more universal and unified expression.

hzxuzhonghu:
>I'd also like to discuss a question. There's a scenario that the API currently can't express: for example, having two subgroups, we want anti-affinity between subgroup-A and subgroup-B, but multiple copies within subgroup-A don't need anti-affinity.

> Analogous scenario: In a 4P2D inference scenario, we want anti-affinity between P and D, but not between the four Ps. Of course, this scenario doesn't seem necessary now, but I'm a bit unsure if it might be needed in the future.

The examples is not very appropriate, but this scenario may apply to other cases. 
The antiaffinity between A and B subgroup is very easy to express, 

```yaml
    subGroupAntiAffinity:
      required:
      - subGroups: [A, B] 
         topologyTierName: rack
    subGroupAffinity:
      required:
      - subGroups: [A] 
         topologyTierName: rack
```

This means the subgroup A, B each reside in different rack. But the subgroup A instances reside in same rack.

hzxuzhonghu:
And we have previously another way `SubGroupPolicySpec` to express affinity too.

wangyang0616:
> interesting proposal. one thing worth calling out is how this composes with intra-node topology — e.g. after the group-topology-affinity plugin picks which supernode/rack a PodGroup lands on, the numaaware plugin still needs to schedule GPU tasks onto the right NUMA nodes within that host. might be worth adding a short section on interaction with the existing numaaware and network-topology-aware plugins so the scheduling pipeline ordering is clear.
> 
> also re: hzxuzhonghu's suggestion about HighestTier/LowestTier naming — i think that's clearer than the raw integer. operators configuring this probably think in terms of "supernode" or "rack" not tier numbers

Thank you for your suggestion. A separate section has been added to describe the process of coordinating the scheduling result of HyperNode in group-topology-affinity with the existing node-level filtering. For details, see [May improve this to add group/sub-group level affinity/anti-affinity.](https://github.com/wangyang0616/volcano/blob/b57d7f9d3529aab7673bf6e577556632282b0bc0/docs/design/group-topology-affinity.md#hypernode-selection-and-node-predicates)

wangyang0616:
> The examples is not very appropriate, but this scenario may apply to other cases. The antiaffinity between A and B subgroup is very easy to express,
> 
> ```yaml
>     subGroupAntiAffinity:
>       required:
>       - subGroups: [A, B] 
>          topologyTierName: rack
>     subGroupAffinity:
>       required:
>       - subGroups: [A] 
>          topologyTierName: rack
> ```
> 
> This means the subgroup A, B each reside in different rack. But the subgroup A instances reside in same rack.

Sure, I have added the corresponding description in the design document.

JesseStutler:
I think it can be push forwarded now? @wangyang0616

pkg/scheduler/framework/session_plugins.go
```diff
@@ -1144,12 +1151,84 @@ func (ssn *Session) HyperNodeGradientForSubJobFn(subJob *api.SubJobInfo, hyperNo
...
+		stats.IntersectedByTier = map[int]int{}
+		return [][]*api.HyperNodeInfo{}, stats
+	}
+
+	hyperNodeByName := make(map[string]*api.HyperNodeInfo, eligible.Len())
+	for _, pluginGradient := range gradientByPlugin {
+		for _, layer := range pluginGradient.Gradients {
+			for _, hn := range layer {
+				if hn != nil && eligible.Has(hn.Name) {
+					hyperNodeByName[hn.Name] = hn
```
JesseStutler, line 1201: nit: It appears that the goal here is simply to retrieve the `*HyperNodeInfo` pointer. Since the intersection of eligible hypernodes necessarily exists across all plugins, obtaining the `*HyperNodeInfo` from just one of them is sufficient, there is no need to traverse all the plugins since the hypernode name will not be duplicate

docs/design/group-topology-affinity.md
JesseStutler, line 1: Currently, we have only implemented anti-affinity in this version. Should we expose all the fields such as subgroup and examples? Alternatively, it might be better to expose them but mark the unimplemented comments and outline the future implementation plan.

docs/design/group-topology-affinity.md
```diff
@@ -0,0 +1,670 @@
...
+  namespace: default
+  labels:
+    topology.volcano.sh/group: llama-70b-prod
+spec:
+  minMember: 8
+  queue: default
+  topologyAffinity:
+    podGroupAntiAffinity:
+      required:
+      - podGroupSelector:
```
JesseStutler, line 169: I have noticed that there is also `namespaceSelector` in the affinity term, can give an example or in the seperate user guide doc

docs/design/group-topology-affinity.md
JesseStutler, line 1: Also better add a user guide doc to teach users how to enable and configure the affinity scheduler plugins, and put these example in the user guide will also be perfect

pkg/scheduler/api/topology_affinity_info.go
```diff
@@ -0,0 +1,532 @@
...
+		}
+		occupiedHyperNodes := CollectJobOccupiedHyperNodesAtTier(matchingJob, hyperNodes, tier, nodesByHyperNode)
+		if occupiedHyperNodes.Len() == 0 {
+			// matching job not yet placed, it occupies no domain, skip it.
+			continue
+		}
+		resolvedHyperNodes := occupiedHyperNodes.UnsortedList()
+		sort.Strings(resolvedHyperNodes)
+		allocatedHyperNode := getJobAllocatedHyperNode(matchingJob, hyperNodes, nodesByHyperNode)
+		klog.V(3).Infof("podGroup anti-affinity: matching job hyperNode, job=%s, matchingJob=%s, termTier=%d, allocatedHyperNode=%s, resolvedHyperNodes=%s",
```
JesseStutler, line 523: nit: this log info may not be clear enough for me, `termTier`can be more clear to  elaborate on the required tier for `podgroup` anti-affinity term.  The semantics of `resolvedHyperNodes` seem somewhat unintuitive.

pkg/scheduler/plugins/group-topology-affinity/group_topology_affinity.go
```diff
@@ -0,0 +1,582 @@
...
+			klog.V(3).Infof("podGroup anti-affinity: resolve term tier failed, job=%s, termIndex=%d, err=%v",
+				klog.KRef(job.Namespace, job.Name), index, err)
+			continue
+		}
+		occupiedHyperNodes := matchingHyperNodesByTerm[index].UnsortedList()
+		sort.Strings(occupiedHyperNodes)
+		klog.V(3).Infof("podGroup anti-affinity: matching occupancy, job=%s, termIndex=%d, tier=%d, occupiedHyperNodes=%s, matchingPodGroups=%s",
+			klog.KRef(job.Namespace, job.Name), index, tier,
+			strings.Join(occupiedHyperNodes, ","),
+			strings.Join(matchingPodGroupPlacementsForTerm(ssn, job, term), "; "))
```
JesseStutler, line 230: It traverse and find matching podgroups again, I think we can reuse the result in the `collectMatchingHyperNodesByTerm`, to do finding matched podgroups only once?

pkg/scheduler/plugins/group-topology-affinity/group_topology_affinity.go
```diff
@@ -0,0 +1,582 @@
...
+	highestAllowedTier int,
+) bool {
+	if hn.Tier() > highestAllowedTier {
+		klog.V(3).Infof("podGroup anti-affinity: reject hyperNode, job=%s, hyperNode=%s, reason=tierAboveHighestAllowed, termIndex=%d, tier=%d, conflictHyperNode=",
+			klog.KRef(job.Namespace, job.Name), hn.Name, noAffinityTermIndex, hn.Tier())
+		return false
+	}
+
+	for index, term := range terms {
+		tier, err := api.ResolvePodGroupTermTier(term, ssn.HyperNodeTierNameMap)
```
JesseStutler, line 328: perf: the `termTier` will be resovled multiple times, but actually it will not change when searching hypernodes, when do `collectMatchingHyperNodesByTerm`, we can firstly resolve and put like
```
type antiAffinityConstraint struct {
    tier     int
    occupied sets.Set[string]
}

e.g.: 
constraints[0] = {tier: 2, occupied: {A, B}}
constraints[1] = {tier: 3, occupied: {X}}
```

Then when do bfs searching we don't need duplicate resolving

pkg/scheduler/plugins/group-topology-affinity/group_topology_affinity.go
```diff
@@ -0,0 +1,582 @@
...
+		if err != nil {
+			klog.Errorf("build podGroup anti-affinity gradient failed, job=%s, err=%v", job.UID, err)
+			return emptyHyperNodeGradients
+		}
+		return result
+	}
+
+	klog.V(3).Infof("podGroup anti-affinity: gradient full-subtree, job=%s, rootHyperNode=%s, allocatedHyperNode=%s",
+		klog.KRef(job.Namespace, job.Name), root.Name, allocatedHyperNode)
+	result, err := gta.buildFullHyperNodeGradient(ssn, root, maxTier, allocatedHyperNode)
```
JesseStutler, line 136: I feel that when there is no hard anti-affinity, we still need to perform a full `buildFullHyperNodeGradient`, essentially iterating within the `highestAllowedTier`. The logic is actually quite similar to that of the `network-topology-aware` plugin. Since we are already modifying `HyperNodeGradientForJobFn`, we could take it a step further.

The `HyperNodeGradientForJobFn` can return type like:
```
type HyperNodeGradientResult struct {
    Unconstrained bool
    Gradients     [][]*HyperNodeInfo
}
```
The semantics would be:
- {Unconstrained: true}: no restriction within the input search root; Session skips this result during intersection.
- {Gradients: candidates}: intersect these candidates with other plugins’ results.
- {} or nil/empty Gradients with Unconstrained: false: no feasible candidates; reject the search.
 
The group plugin could return `Unconstrained` when it has no required rules and does not further restrict the search scope. Preferred rules would remain in scoring.
If all participating plugins return `Unconstrained`, Session could generate the full input-subtree gradient once as a fallback, preserving candidates for preferred scoring. 

Then we can put `hightestAllowedTier` hypernode finding in `network-topoogy-aware` plugin, `podgroup anti-affinity` hypernodes finding in the `group-topology-affinity` plugin

pkg/scheduler/plugins/group-topology-affinity/group_topology_affinity.go
```diff
@@ -0,0 +1,582 @@
...
+	if len(terms) == 0 {
+		return nil, nil
+	}
+
+	hyperNodeCandidates := make([]string, 0, len(hyperNodes))
+	for hyperNode := range hyperNodes {
+		hyperNodeCandidates = append(hyperNodeCandidates, hyperNode)
+	}
+	sort.Strings(hyperNodeCandidates)
+	klog.V(3).Infof("podGroup anti-affinity: evaluate preferred, job=%s, hyperNodeCandidates=%s",
```
JesseStutler, line 366: Logging could have a significant impact on performance, should we consider raising the log level for this plugin? I think under normal circumstances, when scheduling succeeds, we generally don't pay much attention to these info.

pkg/scheduler/plugins/group-topology-affinity/group_topology_affinity.go
```diff
@@ -0,0 +1,582 @@
...
+		sort.Strings(occupiedHyperNodes)
+		klog.V(3).Infof("podGroup anti-affinity: matching occupancy, job=%s, termIndex=%d, tier=%d, occupiedHyperNodes=%s, matchingPodGroups=%s",
+			klog.KRef(job.Namespace, job.Name), index, tier,
+			strings.Join(occupiedHyperNodes, ","),
+			strings.Join(matchingPodGroupPlacementsForTerm(ssn, job, term), "; "))
+	}
+
+	for termIndex, term := range terms {
+		matchingHyperNodes := matchingHyperNodesByTerm[termIndex]
+		tier, err := api.ResolvePodGroupTermTier(term, ssn.HyperNodeTierNameMap)
```
JesseStutler, line 402: Similiar to this comment: https://github.com/volcano-sh/volcano/pull/5349/changes#r4069331808 , we don't need to do duplicate resvole term tier, can you extract outside?

pkg/scheduler/plugins/group-topology-affinity/group_topology_affinity.go
```diff
@@ -0,0 +1,582 @@
...
+func (gta *groupTopologyAffinityPlugin) Name() string {
+	return PluginName
+}
+
+func (gta *groupTopologyAffinityPlugin) OnSessionOpen(ssn *framework.Session) {
+	ssn.AddHyperNodeGradientForJobFn(gta.Name(), func(job *api.JobInfo, hyperNode *api.HyperNodeInfo, _ api.SearchPurpose) [][]*api.HyperNodeInfo {
+		return gta.hyperNodeGradientForJob(ssn, job, hyperNode)
+	})
+
+	ssn.AddHyperNodeGradientForSubJobFn(gta.Name(), func(subJob *api.SubJobInfo, hyperNode *api.HyperNodeInfo, _ api.SearchPurpose) [][]*api.HyperNodeInfo {
```
JesseStutler, line 74: The current logic for subjobs is essentially the same as that for jobs. Although we haven't implemented subgroup anti-affinity support in this version, the scheduler still needs to filter out the matched other anti-affnity podgroups; is there any way to avoid or reduce this part of the duplicate searching?

Or, do we  have to register the `HyperNodeGradientForSubJobFn` extension points in this vesrion? Could we factor the `HyperNodeGradientForSubJobFn` just like this comment? https://github.com/volcano-sh/volcano/pull/5349/changes#r4069624261

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -192,7 +193,7 @@ func (alloc *Action) buildAllocateContext() *allocateContext {
 			continue
 		}
 
-		if !ssn.HyperNodesReadyToSchedule && job.ContainsNetworkTopology() {
+		if !ssn.HyperNodesReadyToSchedule && (job.ContainsNetworkTopology() || job.ContainsHardPodGroupAntiAffinity() || job.HasPreferredPodGroupAntiAffinity()) {
```
JesseStutler, line 196: nit: Can wrap a public func

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -433,7 +434,18 @@ func (alloc *Action) allocateForJob(job *api.JobInfo, jobWorksheet *JobWorksheet
...
+		ssn, hyperNodeGradients, job.GetMinResources(), job.AllocatedHyperNode,
+	)
+	job.SetHyperNodeFitErrors(gradientStats, resourceStats, job.GetMinResources(),
+		ssn.HyperNodesSetByTier, ssn.HyperNodeTierNameMap, ssn.HyperNodes)
+	jobHyperNodeBaseline := job.JobFitErrors
+	if len(hyperNodeGradients) == 0 {
+		klog.V(3).InfoS("No hyperNode gradient for job", "job", job.UID, "fitError", job.JobFitErrors)
+		return nil
+	}
+	klog.V(3).InfoS("HyperNode screening for job", "job", job.UID, "fitError", job.JobFitErrors)
```
JesseStutler, line 448: Don't understand the meaning of this log entry—why to log.V(3) a fitError here when the gradients have results?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -518,7 +531,12 @@ func (alloc *Action) allocateForJob(job *api.JobInfo, jobWorksheet *JobWorksheet
 	return nil
 }
 
-func (alloc *Action) allocateForSubJob(subJob *api.SubJobInfo, subJobWorksheet *SubJobWorksheet, hyperNodeForJob *api.HyperNodeInfo) (*framework.Statement, float64) {
+func (alloc *Action) allocateForSubJob(
+	subJob *api.SubJobInfo,
+	subJobWorksheet *SubJobWorksheet,
+	hyperNodeForJob *api.HyperNodeInfo,
+	jobHyperNodeBaseline string,
```
JesseStutler, line 538: nit: The name of this variable is not clear enough, it actually represents an error.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -245,6 +245,9 @@ func (sc *SchedulerCache) addTask(pi *schedulingapi.TaskInfo) error {
 	job := sc.getOrCreateJob(pi)
 	if job != nil {
 		job.AddTaskInfo(pi)
+		if schedulingapi.AllocatedStatus(pi.Status) || (pi.Status == schedulingapi.Pipelined && pi.NodeName != "") {
```
JesseStutler, line 248: `UpdatePod` holds `sc.Mutex` while `updatePod` removes the old task and adds the new one. With the new calls in both `deleteTask` and `addTask`, an update to an assigned, occupying Pod can invoke this helper twice—even when its node and topology occupancy have not changed.

Each invocation deep-copies all HyperNodes and their resolved node membership sets, then recomputes the affected Job/SubJob placement. Releasing `HyperNodesInfo`'s lock does not release the outer scheduler cache lock. The final `AllocatedHyperNode` comparison only controls logging; the copying and recomputation have already happened.

This introduces cluster-topology-sized work into a frequent, globally locked path, including for jobs without anti-affinity rules and when the new plugin is disabled. At scale, it could increase allocation/GC pressure and delay other cache operations and scheduling snapshots.

Could we instead mark placement dirty on relevant task changes, coalesce updates per job, and reconcile placement against the existing Session topology snapshot before plugins run?

pkg/scheduler/api/topology_affinity_info.go
```diff
@@ -0,0 +1,532 @@
...
+	tier int,
+	nodesByHyperNode map[string]sets.Set[string],
+) string {
+	if task == nil || task.NodeName == "" || len(nodesByHyperNode) == 0 {
+		return ""
+	}
+
+	finestName := ""
+	finestTier := 0
+	for name, nodes := range nodesByHyperNode {
```
JesseStutler, line 457: perf: `taskOccupiedHyperNodeAtTier` scans the entire HyperNode membership map for every occupying task. Even within a single occupancy calculation, multiple tasks on the same Node repeat the same lookup, making this step O(occupied tasks × HyperNodes).

Could we build or resuse a Session-local reverse index from Node name to its finest HyperNode, and resolve the requested tier through the ancestor chain? This topology lookup can be reused across terms and callbacks without caching mutable task occupancy.

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -537,7 +555,17 @@ func (alloc *Action) allocateForSubJob(subJob *api.SubJobInfo, subJobWorksheet *
...
 	}
 
-	hyperNodeGradients := ssn.HyperNodeGradientForSubJobFn(subJob, hyperNodeForJob, api.PurposeAllocate)
+	hyperNodeGradients, gradientStats := ssn.HyperNodeGradientForSubJobFn(subJob, hyperNodeForJob, api.PurposeAllocate)
+	hyperNodeGradients, resourceStats := FilterGradientsByMinResource(
+		ssn, hyperNodeGradients, subJob.GetMinResources(), subJob.AllocatedHyperNode,
+	)
+	if len(hyperNodeGradients) == 0 {
+		job.MergeSubJobHyperNodeFitErrors(jobHyperNodeBaseline, subJob.UID, gradientStats, resourceStats,
+			subJob.GetMinResources(), ssn.HyperNodesSetByTier, ssn.HyperNodeTierNameMap, ssn.HyperNodes)
```
JesseStutler, line 564: Current subjob's `GetMinResources` implementation may not be recise enough, I jumped into it and took a look, it will gather all the pending tasks resources for the subjob, but a subgroup can contain its own min member, don't need whole resources of all pending tasks

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -699,12 +736,67 @@ func (alloc *Action) allocateFromNomination(subJob *api.SubJobInfo, subJobWorksh
...
 	newAllocatedHyperNode := ssn.HyperNodes.GetLCAHyperNode(subJob.AllocatedHyperNode, pinned)
 	subJob.AllocatedHyperNode = newAllocatedHyperNode
+	updateJobAllocatedHyperNodeFromSubJob(ssn, job, subJob, newAllocatedHyperNode)
 	alloc.recorder.SaveSubJobDecision(subJob.Job, hyperNodeForJob.Name, subJob.UID, newAllocatedHyperNode)
 	klog.V(3).InfoS("Allocate subJob from nomination success", "subJob", subJob.UID,
 		"nominatedHyperNode", pinned, "newAllocatedHyperNode", newAllocatedHyperNode)
 	return stmt, 0, true
 }
 
+func (alloc *Action) nominationSatisfiesHyperNodeConstraints(
```
JesseStutler, line 746: I understand the objective here, but since `gangpreempt`/`gangreclaim` has already performed the gradient-based search—and the entire process chain is quite lengthy—having the `allocate` phase repeat a full gradient search seems counterproductive, the cost of the resulting evictions is simply too high. This repeats peer-occupancy collection, subtree traversal, gradient intersection and sorting

Could we provide a targeted validation path for the pinned domain or nominated Nodes, reusing current constraint/topology data where valid, while preserving the existing Job/SubJob placement boundaries?

pkg/scheduler/actions/allocate/allocate.go
```diff
@@ -934,6 +1033,47 @@ func getNewAllocatedHyperNode(ssn *framework.Session, bestNode string, jobAlloca
...
+	}
+}
+
+func restoreHyperNodePlacement(job *api.JobInfo, subJob *api.SubJobInfo, placement hyperNodePlacement) {
+	job.AllocatedHyperNode = placement.jobAllocatedHyperNode
+	subJob.AllocatedHyperNode = placement.subJobAllocatedHyperNode
+}
+
+func shouldTrackHyperNodePlacement(ssn *framework.Session, subJob *api.SubJobInfo) bool {
+	return subJob.WithNetworkTopology() || ssn.HyperNodesReadyToSchedule
```
JesseStutler, line 1054: `HyperNodesReadyToSchedule` actually is initialized to true, and Session adds a virtual cluster-top HyperNode containing all Nodes even when no HyperNode CRs exist.

With this condition, ordinary jobs now perform placement tracking even when both topology plugins are disabled. In `allocateResourcesForTasks()`, each successful allocation calls `getNewAllocatedHyperNode() -> FindHyperNodeForNode()`. With only the virtual root present, this introduces an additional linear search through the Session's Node list per allocated task, which the previous `WithNetworkTopology()` guard skipped.

Could we avoid enabling this work solely based on cache readiness and preserve the existing non-topology fast path? Please add a regression test and benchmark with no HyperNode CRs and both topology plugins disabled. Jobs without their own topology rules must still remain observable when selected by another PodGroup's anti-affinity rules, but that should not require unconditional tracking when the feature is unused.

PR #5353, opened 2026-05-30
Hierarchical queue close/open: admission hardening + test-helper cleanup

### LINKED ISSUE #5351: Hierarchical queue close/open: admission hardening + test-helper cleanup
### What would you like to be added?

> This was originally proposed by @devzizu at the community meeting on 2026-05-15. I'm filing it as a tracked issue and grouping it with two adjacent improvements I hit while reviewing #5350.

Today the queue controller handles hierarchical state (close cascades to children, open is rejected if the parent is closed), but the admission webhook does not enforce any of these constraints. As a result:

- Asynchronous controller reconciles race with anything that touches a hierarchical queue, including E2E cleanup. This is the root cause of the flake fixed in #5350, see #5346.
- `OpenQueue` on a child whose parent is `Closed` is only rejected by the controller after retry storms.
- Deleting an `Open` queue with foreground propagation hangs until the controller closes it, which is why `DeleteQueue` in [test/e2e/util/queue.go](https://github.com/volcano-sh/volcano/blob/v1.14.2/test/e2e/util/queue.go) first writes `Status.State = Closed`.

This issue groups three related improvements:

#### 1. Webhook feature flag for cascading child closure (the @devzizu proposal)

Add `--enable-cascade-child-queue-close` (default `false`) so when enabled:

- `OpenQueue` requests for a child whose parent is `Closed` are rejected at admission, not after the controller's retries.
- **Auto-cascade on close (preferred):** `CloseQueue` on a parent triggers admission-side close of all descendant children in the same admission request, so the entire subtree transitions to `Closed` synchronously. This matches what a user expects when they close a parent (the children should follow) and avoids the controller's async cascade racing with other admission requests in the meantime.

For comparison, the alternative would be a require-children-closed model where admission rejects a parent close until children are closed first. That's simpler to implement but pushes the work onto the caller, who then has to walk the subtree themselves before closing the parent. Auto-cascade is the better user experience and is what we should aim for; the only real cost is that admission now does multi-object writes (one per descendant), so the implementation needs to be careful about partial-failure semantics (transactional close or all-or-nothing rejection).

#### 2. Delete admission: optionally require `Status.State == Closed`

Extend the existing `validateQueueDeleting` ([validate_queue.go#L295](https://github.com/volcano-sh/volcano/blob/v1.14.2/pkg/webhooks/admission/queues/validate/validate_queue.go#L295)) to optionally require the queue to be `Closed` before delete (behind a new flag, similar to `--enable-queue-allocated-pods-check` from #4973). This makes the `UpdateStatus(Closed)` step in [test/e2e/util/queue.go](https://github.com/volcano-sh/volcano/blob/v1.14.2/test/e2e/util/queue.go) (and any production cleanup) load-bearing rather than implicit-finalizer-only.

#### 3. E2E test helper: rework `deleteQueues` to close top-down, delete bottom-up

`DeleteQueue` in [test/e2e/util/queue.go](https://github.com/volcano-sh/volcano/blob/v1.14.2/test/e2e/util/queue.go) currently does per-queue `UpdateStatus(Closed)`, then `Delete`, then wait, all in a single loop. `Expect(retryErr).NotTo(HaveOccurred())` aborts hard if `UpdateStatus` fails. For hierarchical queues this is brittle:

- **Today:** `UpdateStatus` on a child races with the controller's parent-cascade reconciles (the root cause of #5346 / #5350).
- **Tomorrow:** if sub-items 1 or 2 land, admission may reject `UpdateStatus(Closed)` on a child whose parent is still `Open`, or `Delete` on a queue whose `Status.State` is `Open`, making the per-queue loop fail even more.

The robust shape mirrors how a user is meant to manage a hierarchy:

1. Close from the **topmost** parent (one `UpdateStatus(Closed)` per top-level queue).
2. Wait for the controller's cascade to close children.
3. Delete in **deepest-first** order (`q11`, then `q1`, then `q2`), waiting for `NotFound` after each.

This also lets callers stop hand-ordering `ctx.Queues = []string{q11, q1, q2}` to work around the helper's limitations. The helper becomes correct by default. Add a doc comment on `DeleteQueue` and/or `deleteQueues` explaining the constraint either way.

### Why is this needed?

- **Reconcile noise / flakiness.** The current model leaves cascade behavior entirely to the controller, which races with anything touching a hierarchical queue. PR #5350 fixes one symptom; the underlying gap remains.
- **Predictable semantics for users.** Today, `OpenQueue` on a child with a closed parent gets an asynchronous failure event after retry storms. An admission-level rejection makes the constraint synchronous and obvious.
- **Hierarchical hardening continuity.** PR #4973 added admission validation for capability hierarchy and queue depth; cascading state validation is the natural next step.

### Proposed behavior

| Operation              | Parent state | Child state before | With flag = `true`                            | With flag = `false` (today)            |
|------------------------|--------------|--------------------|-----------------------------------------------|----------------------------------------|
| `OpenQueue` on child   | `Closed`     | `Closed`           | Reject at admission                            | Controller rejects after retries       |
| `CloseQueue` on parent | `Open`       | `Open` (children)  | Admission auto-cascades close to all descendants synchronously | Controller cascades, async |
| `Delete` on parent     | any          | child exists       | (already rejected today)                       | Rejected                               |
| `Delete` on queue      | any          | own state `Open`   | Optionally reject (sub-item 2)                 | Allowed; hangs until controller closes |

### Implementation sketch

- Add options to [cmd/webhook-manager/app/options/options.go](https://github.com/volcano-sh/volcano/blob/v1.14.2/cmd/webhook-manager/app/options/options.go), plumbed through [pkg/webhooks/router/interface.go](https://github.com/volcano-sh/volcano/blob/v1.14.2/pkg/webhooks/router/interface.go) into [pkg/webhooks/admission/queues/validate/validate_queue.go](https://github.com/volcano-sh/volcano/blob/v1.14.2/pkg/webhooks/admission/queues/validate/validate_queue.go), same pattern as `EnableQueueAllocatedPodsCheck` from #4973.
- Lookup parent state via the existing `QueueLister` already used by `validateQueueDeleting`.
- Add unit tests next to `validate_queue_test.go`.
- For sub-item 3: depth-sort using existing `listQueueChild` / `Spec.Parent` lookups, plus a small godoc comment on `DeleteQueue`.
- Update hierarchical E2E tests to optionally set the new flags.

### Background

- Community meeting discussion: 2026-05-15, item raised by @devzizu.
- Related PR: #5350 (this issue is the follow-up flagged in its review).
- Related PR: #4973 (introduced the existing webhook hardening flags: `--enable-queue-allocated-pods-check`, `--enable-root-queue-protection`, `defaultMaxQueueDepth`).
- Related issue: #5346 (flaky hierarchical queue E2E test, root cause is parent-cascade reconcile race).

antcybersec:
Fixes #5351


pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -329,6 +349,114 @@ func validateQueueDeleting(queueName string) error {
...
+		queue, err = config.VolcanoClient.SchedulingV1beta1().Queues().Update(context.TODO(), queue, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+
+		queue.Status.State = schedulingv1beta1.QueueStateClosed
+		_, err = config.VolcanoClient.SchedulingV1beta1().Queues().UpdateStatus(context.TODO(), queue, metav1.UpdateOptions{})
+		return err
+	})
+}
```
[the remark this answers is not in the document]
hajnalmt: These are valid problem unfortunately as @devzizu pointed this out here.
https://github.com/volcano-sh/volcano/issues/5351#issuecomment-4582970375

hajnalmt (changes requested):
As the bots are pointing it out, this won't suffice, we need to batch/concurrently handle the requests somehow.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -40,6 +44,11 @@ import (
 	"volcano.sh/volcano/pkg/webhooks/util"
 )
 
+const (
+	closedByParentAnnotationKey       = "volcano.sh/closed-by-parent"
```
hajnalmt: This key is a good idea though. It's place shall be in the API, you can put them in `staging/src/volcano.sh/apis/pkg/scheduling/v1beta1/labels`

test/e2e/util/queue.go
```diff
@@ -105,8 +106,43 @@ func CreateQueues(ctx *TestContext) {
...
 func DeleteQueue(ctx *TestContext, q string) {
+	deleteJobsInQueue(ctx, q)
+	closeQueueStatus(ctx, q)
+	deleteQueueObject(ctx, q)
+}
+
+// deleteQueues deletes all queues in ctx.Queues. For hierarchical queues it closes from
+// the topmost parent first, waits for the controller cascade to close children, then
+// deletes deepest-first so callers do not need to hand-order ctx.Queues.
+func deleteQueues(ctx *TestContext) {
```
hajnalmt, line 120: This algorithm looks good!

hajnalmt (commented):
As https://github.com/volcano-sh/volcano/pull/5350 landed in the repo please rebase and fix the conflict by modifying the deleteQueues comment. Also DCO need to be signed!

hajnalmt (commented):
/assign

antcybersec:
Rebased onto master (conflicts were trivial - one line in server.go, plus labels.go needed both sides kept) and pushed two more commits:

- ran gofmt on interface.go and validate_queue.go, Copilot flagged the formatting a while back and I never went back and fixed it
- fixed a real bug: if one descendant fails to close during cascade, the old code returned that error and it ended up denying the parent's own close request too, even though the parent's transition was totally valid. meanwhile any descendants that did close stayed closed. so you'd get a rejected parent update sitting next to already-closed children. this is the rollback/transaction thing devzizu brought up on #5351. now it just logs and lets the parent's transition go through. added a test that forces one descendant's update to fail and checks the parent still gets admitted.

everything from the last round should still be there - cycle detection in queueDepth, concurrent descendant closing, the closed-by-parent annotation guard, annotation keys moved to labels.go, DCO signed.

still haven't touched the bigger question of whether cascade-close belongs in a validating webhook at all vs a controller. didn't want to make that call on my own, but can do it if that's the direction you want.

@hajnalmt @hwdef ready for another look when you get a chance

PR #5357, opened 2026-05-30
Fix confusing namespace hardcode and include error context in InitDevicesConfig log

### LINKED ISSUE #5315: The klog of InitDevicesConfig(pkg/scheduler/api/devices/config/config.go) causes unnecessary confusion.
### Description

I tried to register the ascend NPU device using the ascend-device-configmap provided by the Hami, and added the following in the volcano-scheduler-configmap:
- name: deviceshare
  arguments:
    deviceshare.AscendHAMiVNPUEnable: true # enable ascend vnpu
    deviceshare.SchedulePolicy: binpack # scheduling policy. binpack / spread
    deviceshare.KnownGeometriesCMNamespace: volcano
    deviceshare.KnownGeometriesCMName: hami-scheduler-device

It failed unexpectedly because the official ascend-device-configmap contained some fields that volcano does not support. After deleting these fields, regardless of whether hami itself is installed, volcano can use the ascend-device-plugin to complete scheduling.

But during the debugging process, the logs printed by InitDevicesConfig confused me：

（"Volcano device config not found in namespace kube-system, using default config","name", cmName)

This made me think that the namespace was not set correctly, or that the config file can only be created in kube-system. What's even more confusing is that when the default config is followed by the name of my custom config, I could hardly tell which config volcano actually loaded. It was only after analyzing the code that I discovered some fields were not supported.

Is it possible to improve it in the following way:

（"Volcano device config not found in", cmName, "(namespace ", cmNamespace, ")"，“using default config”)

1.14.1

sahil-v001:
#### What type of PR is this?
/kind cleanup

This PR removes a confusing, hardcoded `"kube-system"` namespace string from the `InitDevicesConfig` logging block inside `pkg/scheduler/api/devices/config/config.go`. 

Previously, if a user specified a custom namespace for their device configuration, the error log would still claim it was looking in `kube-system`. This PR transitions the log line to a proper structured logging pattern (`klog.InfoS`), natively tracking both the target configuration `name` and the actual evaluated `namespace` dynamically.

Fixes #5315

The log was converted from an unformatted string to structured logging parameters (`"name", cmName, "namespace", cmNamespace`) to align with Kubernetes and Volcano logging best practices. 

*Disclosure: I used Gemini to cross-verify the structured logging format constraints and help draft this PR description.*

```release-note
Fixed a confusing log message in InitDevicesConfig that erroneously reported the hardcoded 'kube-system' namespace instead of the user-configured namespace.

### DIFF
--- a/pkg/scheduler/api/devices/config/config.go
+++ b/pkg/scheduler/api/devices/config/config.go
@@ -157,8 +157,7 @@ func InitDevicesConfig(cmName, cmNamespace string) {
 		var err error
 		configs, err = loadConfigFromCM(devices.GetClient(), cmName, cmNamespace)
 		if err != nil {
-			klog.V(3).InfoS("Volcano device config not found in namespace kube-system, using default config",
-				"name", cmName)
+			klog.V(3).InfoS("Volcano device config not found, using default config", "name", cmName, "namespace", cmNamespace, "err", err)
 			configs = GetDefaultDevicesConfig()
 		}
 		klog.V(3).InfoS("Initializing volcano device config", "device-configs", configs)
--- a/pkg/scheduler/api/devices/config/config.go
+++ b/pkg/scheduler/api/devices/config/config.go
@@ -157,7 +157,7 @@ func InitDevicesConfig(cmName, cmNamespace string) {
 		var err error
 		configs, err = loadConfigFromCM(devices.GetClient(), cmName, cmNamespace)
 		if err != nil {
-			klog.V(3).InfoS("Volcano device config not found, using default config", "name", cmName, "namespace", cmNamespace, "err", err)
+			klog.V(3).InfoS("Volcano device config not found, using default config", "name", cmName, "namespace", cmNamespace, "error", err)
 			configs = GetDefaultDevicesConfig()
 		}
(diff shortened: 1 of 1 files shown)


pmady:
looks like a straightforward fix. couple things:

1. the bot flagged merge commits — you'll need to rebase onto master instead of merging. `git rebase upstream/master` then force push should fix it

2. agree with the gemini suggestion about changing the message wording. "not found" is misleading since `loadConfigFromCM` can also fail on YAML parse errors or nil kube client. something like "failed to load device config, using defaults" would be more accurate

3. make sure the error itself is included in the structured log — right now the original error from `loadConfigFromCM` gets swallowed. adding `"error", err` to the InfoS args would help debugging

hajnalmt (commented):
Thanks for the PR!

Sorry that I looked on it only now. It looks like @DSFans2014 found this issue too and I think his wording is somewhat better. The PRs topic is different but the PR looks quite ready, and I don't want to have conflicts there now. If there is objection in that PR we can revisit this.

https://github.com/volcano-sh/volcano/pull/5462/changes#diff-f7ca7760e2679000c2929f0e6e2045bd664746a43eb549861cec07c88980203dR154

Issue #5346, opened 2026-05-28
[Flaky Test] Queue E2E Test [DeferCleanup (Each)]


JesseStutler:
### Description

We have met ` Queue E2E Test [DeferCleanup (Each)]` quite a lot times, the example error log:
```
Queue E2E Test Queue Command Close And Open With Hierarchical Queues
/home/runner/work/volcano/volcano/test/e2e/jobseq/queue.go:75
  STEP: Initializing test context @ 05/27/26 14:40:40.399
  STEP: Creating Queues @ 05/27/26 14:40:40.6
  STEP: Close queue with child queues @ 05/27/26 14:40:43.616
  STEP: Open queue with closed parent queue @ 05/27/26 14:40:43.724
  STEP: Open queue with open parent queue @ 05/27/26 14:40:43.83
  STEP: Cleaning up test context @ 05/27/26 14:40:43.937
  [FAILED] in [DeferCleanup (Each)] - /home/runner/work/volcano/volcano/test/e2e/util/queue.go:155 @ 05/27/26 14:45:44.219
• [FAILED] [303.821 seconds]
Queue E2E Test [DeferCleanup (Each)] Queue Command Close And Open With Hierarchical Queues
  [DeferCleanup (Each)] /home/runner/work/volcano/volcano/test/e2e/jobseq/queue.go:90
  [It] /home/runner/work/volcano/volcano/test/e2e/jobseq/queue.go:75

  [FAILED] failed to wait queue queue11 closed
  Unexpected error:
      <wait.errInterrupted>: 
      timed out waiting for the condition
      {
          cause: <*errors.errorString | 0xc0003b7ef0>{
              s: "timed out waiting for the condition",
          },
      }
  occurred
  In [DeferCleanup (Each)] at: /home/runner/work/volcano/volcano/test/e2e/util/queue.go:155 @ 05/27/26 14:45:44.219
```

We need to figure it out and solve it

### Steps to reproduce the issue

1.
2.
3.

### Describe the results you received and expected

Solve the flake test

### What version of Volcano are you using?

latest

JesseStutler:
The example action log: https://github.com/volcano-sh/volcano/actions/runs/26517198074/job/78099856260?pr=5345

/kind flake
/help

Will give a chance if any contributor can firstly solve it

PR #5368, opened 2026-06-03
e2e: rework deleteQueues for hierarchical close/delete ordering

bp-cheng:
#### What type of PR is this?
/kind cleanup

The previous `deleteQueues` iterated `ctx.Queues` in declaration order, calling `DeleteQueue` (UpdateStatus(Closed) + Delete + wait) per queue. For hierarchical queues this races with the controller's parent-cascade reconciles: closing a parent transitions children to `Closed` asynchronously, but the per-queue loop tries to `UpdateStatus(Closed)` on each child independently. That race is one root cause of the flake fixed in #5350 (see #5346). It also forced callers to hand-order `ctx.Queues` children-before-parents to satisfy admission's "parent has children" rule (see `test/e2e/jobseq/queue.go:89`).

This reshapes the teardown to match how a user is meant to manage a hierarchy, per #5351's sub-item 3:

1. Clean up vcjobs in every queue (behaviour unchanged).
2. `UpdateStatus(Closed)` only on top-level queues (parent `""` or `"root"`). The controller cascade-closes descendants.
3. Wait for every queue in the test context to reach `Closed`.
4. Delete deepest-first, waiting for `NotFound` after each.

Small internal helpers (`deleteJobsInQueue`, `closeQueueStatus`, `waitQueueClosed`, `deleteQueueAndWait`, `topLevelQueues`, `queueDepth`, `queuesDeepestFirst`) carry the steps. `DeleteQueue` is refactored to a three-line composition of the first three, so single-queue callers in `test/e2e/schedulinggates` and `test/e2e/stress` keep identical behaviour. Its godoc now points hierarchical teardown to `deleteQueues` so the constraint is discoverable from either entry point.

A new `test/e2e/util/queue_test.go` unit-tests the pure ordering/lookup helpers — no cluster required.

Callers can now declare `ctx.Queues` in any order; the helper sorts by depth itself. Hand-ordered call sites like `testCtx.Queues = []string{q11, q1, q2}` are no longer load-bearing (cleaning them up is left for a separate PR to keep this one focused on the helper).

Tracks #5351 (sub-item 3 only). Sub-items 1 (`--enable-cascade-child-queue-close`) and 2 (delete admission requiring `Status.State == Closed`) introduce new admission webhook flags and will land as separate PRs.

**AI disclosure** (per [contribute.md#ai-guidance](https://github.com/volcano-sh/volcano/blob/master/contribute.md#ai-guidance)): I used Claude to help draft this refactor. I reviewed every line, ran `make lint` (0 issues) and `go test ./test/e2e/util/...` (all helper unit tests pass) locally before pushing, and I will own all review responses without AI assistance.

Per the PR template's request for AI agents:

> *Beneath a parent's cooling crown,*
> *children must Close before they go down —*
> *deepest first, then up the chain,*
> *no race, no flake, no test in vain.*

Prompt used to drive this PR: "Implement sub-item 3 of volcano-sh/volcano#5351 — rework `deleteQueues` in `test/e2e/util/queue.go` to close top-down via the controller cascade and delete deepest-first, keeping `DeleteQueue` working for single-queue callers and adding unit tests for the pure-logic helpers."

I have not run the e2e suite locally (no cluster handy); the unit tests cover the ordering/lookup logic and the helpers' behavioural shape mirrors what `DeleteQueue` does today, but reviewers should confirm via CI's e2e run.

```release-note
NONE
```


bp-cheng:
Thanks @Copilot and @gemini-code-assist. Pushed a follow-up commit addressing the feedback:

- `SeyQueueReclaimable` typo → fixed.
- `deleteQueues` now lists vcjobs once via a new `deleteJobsInQueues` helper; `DeleteQueue` keeps its per-queue path via a thin wrapper around it.
- `queuesDeepestFirst` precomputes depths into a map before the `sort.SliceStable` call so the comparator no longer re-walks parent chains.
- The cycle-test comment now reads `len(QueueParent)+1` to match the actual cap in `queueDepth`.
- Dropped the inner `Expect` inside the `RetryOnConflict` body of `closeQueueStatus` and `SetQueueReclaimable`; the outer `Expect` on `retryErr` covers the failure path so the retry loop sees the real error.

Also rebased onto current `master`. The `Merge branch master` commit that GitHub's auto-update added is gone, and the `do-not-merge/contains-merge-commits` label has cleared.

test/e2e/util/queue.go
```diff
@@ -172,38 +299,57 @@ func DeleteQueue(ctx *TestContext, q string) {
...
 }
 
-// deleteQueues deletes Queues specified in the test context.
-// For hierarchical queues, list children before parents in ctx.Queues; admission
-// rejects deleting a parent that still has child queues.
-func deleteQueues(ctx *TestContext) {
+// topLevelQueues returns the subset of ctx.Queues whose declared parent is
+// empty or "root" — i.e. the queues a user would close to trigger a cascade
+// across the entire hierarchy owned by this test context.
+func topLevelQueues(ctx *TestContext) []string {
```
hajnalmt: This algorithm looks better:
https://github.com/volcano-sh/volcano/pull/5353/changes#diff-d844c5272538240e75a0b50a94981c6965fbceacc030e07b5a838298b006a9f8R228
Because it calculates with subhierarchy deletes when the parent is not in the Set:
```
		if _, ok := queueSet[parent]; !ok {
			topLevel = append(topLevel, q)
		}
```

test/e2e/util/queue_test.go
```diff
@@ -0,0 +1,208 @@
+/*
```
hajnalmt: I don't think wee need a UT for an e2e test util file 😄 it's CI consuming enough to run the UTs for the product itself, but I genuinely appreciate the effort!

test/e2e/util/queue.go
```diff
@@ -151,14 +254,38 @@ func DeleteQueue(ctx *TestContext, q string) {
...
+}
 
-	err = ctx.Vcclient.SchedulingV1beta1().Queues().Delete(context.TODO(), q,
-		metav1.DeleteOptions{
-			PropagationPolicy: &foreground,
-		})
+// waitQueueClosed waits until queue q reports State == Closed. If the queue is
+// already gone (NotFound), that counts as closed too — a deeper queue may have
+// been cleaned up by a previous iteration.
+func waitQueueClosed(ctx *TestContext, q string) {
```
hajnalmt, line 240: I like this implementation better than the one in the other PR:

https://github.com/volcano-sh/volcano/pull/5353/changes#diff-d844c5272538240e75a0b50a94981c6965fbceacc030e07b5a838298b006a9f8R129

Calculating with not found queues as closed makes sense to me, although it probably doesn't make any difference at all in our e2e execution workflow.

test/e2e/util/queue.go
```diff
@@ -172,38 +299,57 @@ func DeleteQueue(ctx *TestContext, q string) {
...
-		Expect(retryErr).NotTo(HaveOccurred(), "failed to update queue %s", q)
+// queueDepth returns the chain length from q up to a root-equivalent ancestor.
+// A queue with no parent (or parent "root") has depth 0; its direct children
+// have depth 1; and so on. Parents that aren't tracked in ctx.QueueParent are
+// treated as root, so a depth always exists.
+//
+// A cycle would loop forever; we cap iterations at len(ctx.QueueParent)+1 as
+// a defensive measure — a real acyclic hierarchy can't have a chain longer
+// than the number of parent edges.
+func queueDepth(ctx *TestContext, q string) int {
```
hajnalmt, line 286: As a map will be returning the zero value of the given type this is a lot nicer and more compact.:

https://github.com/volcano-sh/volcano/pull/5353/changes#diff-d844c5272538240e75a0b50a94981c6965fbceacc030e07b5a838298b006a9f8R235

test/e2e/util/queue.go
```diff
@@ -105,8 +106,107 @@ func CreateQueues(ctx *TestContext) {
...
+	return wait.Poll(100*time.Millisecond, TenMinute, condition)
+}
+
+// deleteJobsInQueue deletes every vcjob in ctx.Namespace whose Spec.Queue == q,
+// waiting for each to be gone before returning.
+func deleteJobsInQueue(ctx *TestContext, q string) {
+	deleteJobsInQueues(ctx, []string{q})
+}
+
+// deleteJobsInQueues deletes every vcjob in ctx.Namespace whose Spec.Queue is
```
hajnalmt: You can delete unnecesarry comments everywhere where the function name is reflective enough.

hajnalmt (changes requested):
Thank you for making an effort on this!

I am somewhat unhappy that we are doing this in separate PRs as the main idea in the whole e2e test change (I was trying to focus in the https://github.com/volcano-sh/volcano/issues/5351 issue) would have been to delete this line here:
https://github.com/volcano-sh/volcano/blob/v1.15.0/test/e2e/jobseq/queue.go#L89
Which this PR is not doing in the end as the other items are not covered.

Anyhow if this makes the implementation easier for other parts and we won't forget to delete this line in the webhook PR then I am not againts separate PRs at all. Please check my comments below!

hajnalmt (commented):
Can I ask which model have you used for coding? Just out of curiosity 😊 I also appreciate the rhyme, some slop literature makes these reviews better 😄

test/e2e/util/queue.go
bp-cheng: Nice catch! You're right — my version silently returns an empty list when a test owns a sub-hierarchy whose root parent lives outside ctx.Queues, which would then hang the wait loop because nothing triggered the cascade...
Adopting #5353's queueSet check in the next push. Thanks for spotting this.

test/e2e/util/queue.go
bp-cheng: Agree, I am do it with over design.
or this domain (≤10 queues per test context, then real API calls) the precompute is a premature optimization, and the inline call composes more clearly. Will simplify in the next push.

test/e2e/util/queue_test.go
bp-cheng: Will remove in the next push. Thanks for the friendly framing! 😄

bp-cheng:
> Can I ask which model have you used for coding? Just out of curiosity 😊 I also appreciate the rhyme, some slop literature makes these reviews better 😄

it is Claude 4.7. If you like it I can send you a slop every day😄😄😄

bp-cheng:
It's Claude 4.7. If you like it, I can send you a slop every day. 😄

test/e2e/util/queue.go
bp-cheng: Nice catch! You're right — our version silently returns an empty list when a test owns a sub-hierarchy whose root parent lives outside `ctx.Queues`, which would then hang the wait loop because nothing triggered the cascade. Adopted #5353's `queueSet` check in the latest push. Thanks for spotting this.

test/e2e/util/queue_test.go
bp-cheng: Removed. FYI the tests did catch a `len(Queues)` vs `len(QueueParent)` bug in `queueDepth` during dev — kept the cycle-protection bound and added a short comment on `queueDepth` explaining why so the reasoning survives the deletion. Thanks for the friendly framing! 😄

test/e2e/util/queue.go
bp-cheng: Thanks! The NotFound branch is partly defensive for the case where the queue is gone before we observe it Closed (overlapping cleanups, re-runs against leftover state, or manual intervention). Agreed it shouldn't show up in the normal e2e flow, but felt cheap to handle and avoids a misleading `failed waiting for Closed` when in fact the queue is already further along than Closed.

test/e2e/util/queue.go
bp-cheng: Agreed, I overdesigned it. For this domain (≤10 queues per test context, then real API calls) the precompute is a premature optimization, and the inline call composes more clearly. Simplified in the latest push.

test/e2e/util/queue.go
bp-cheng: Done in the latest push — trimmed redundant godoc where the function name already covers it, kept the ones documenting the "why" (cycle protection, NotFound semantics, hierarchical teardown ordering).

bp-cheng:
Thanks @hajnalmt for the careful review! Pushed [b3f28c4](https://github.com/volcano-sh/volcano/commit/b3f28c4cf) addressing all five comments:

- `topLevelQueues` — adopted #5353's `queueSet` check for sub-hierarchies whose root parent lives outside `ctx.Queues`.
- `queuesDeepestFirst` — dropped the precomputed depth map; the inline call composes more clearly for ≤10 queues.
- `test/e2e/util/queue_test.go` — removed; kept the cycle-protection bound in `queueDepth` as a code comment.
- Trimmed redundant godoc where the function name covers it; kept the "why" comments (cycle protection, NotFound semantics, hierarchical ordering).

On scope: leaving `testCtx.Queues = []string{q11, q1, q2}` in `test/e2e/jobseq/queue.go:89` for the upcoming webhook PR (sub-items 1 + 2) as you suggested — happy to take that on after this lands, or to remove it here if you'd prefer. PTAL whenever it fits your schedule. 🙏

hajnalmt (commented):
/area test
/area scheduling
/ok-to-test

Thanks for the updates. I think this scoped e2e cleanup is the right piece to merge first for #5351; the other PR still has a lot of gaps in the admission-side cascade design, so I would rather land this low-risk helper cleanup separately. Let's wait for CI before adding LGTM.

hajnalmt:
Oh please rebase! You can ping me if I forget to write back. :sweat_smile:

bp-cheng:
Hi, @hajnalmt , thanks for review. This branch should be ready, now.

bp-cheng:
> Oh please rebase! You can ping me if I forget to write back. 😅

Hi, @hajnalmt , thanks for review. This branch should be ready, now.

test/e2e/util/queue.go
```diff
@@ -151,59 +232,77 @@ func DeleteQueue(ctx *TestContext, q string) {
...
+// manual intervention) still satisfies the wait's intent.
+func waitQueueClosed(ctx *TestContext, q string) {
+	err := wait.PollUntilContextTimeout(context.TODO(), time.Second, TwoMinute, true, func(pollCtx context.Context) (bool, error) {
+		queue, err := ctx.Vcclient.SchedulingV1beta1().Queues().Get(pollCtx, q, metav1.GetOptions{})
 		if errors.IsNotFound(err) {
 			return true, nil
 		}
 		if err != nil {
-			return false, err
+			// Transient client/API errors (notably client-side rate limiting
```
hajnalmt, line 247: Can you explain this? If an API call fails we don't swallow it most of the time.

test/e2e/util/queue.go
hajnalmt: You deleted the proper fix in the last commit that I asked for.

test/e2e/util/queue.go
```diff
@@ -151,59 +232,77 @@ func DeleteQueue(ctx *TestContext, q string) {
...
-			if _, err = ctx.Vcclient.SchedulingV1beta1().Queues().Update(context.TODO(), queue, metav1.UpdateOptions{}); err != nil {
-				return err
-			}
-			return nil
-		})
-		Expect(retryErr).NotTo(HaveOccurred(), "failed to update queue %s", q)
+// queueDepth walks parent links up to a root-equivalent ancestor. We cap
+// iterations at len(ctx.QueueParent)+1 so a misconfigured ctx.QueueParent
+// (cycle) can't hang the helper.
+func queueDepth(ctx *TestContext, q string) int {
```
hajnalmt, line 286: Please copy the function from the other PR.
https://github.com/volcano-sh/volcano/pull/5353/changes#diff-d844c5272538240e75a0b50a94981c6965fbceacc030e07b5a838298b006a9f8R243

test/e2e/util/queue.go
```diff
@@ -151,59 +232,77 @@ func DeleteQueue(ctx *TestContext, q string) {
...
+	if errors.IsNotFound(err) {
+		return
 	}
-}
+	Expect(err).NotTo(HaveOccurred(), "failed to delete queue %s", q)
 
-// SeyQueueReclaimable sets the Queue to be reclaimable
-func SetQueueReclaimable(ctx *TestContext, queues []string, reclaimable bool) {
-	By("Setting Queue reclaimable")
+	err = wait.PollUntilContextTimeout(context.TODO(), time.Second, TwoMinute, true, func(pollCtx context.Context) (bool, error) {
```
hajnalmt, line 270: Raising this to 1 second makes sense this is what Fable got right probably.

test/e2e/util/queue.go
```diff
@@ -105,8 +104,87 @@ func CreateQueues(ctx *TestContext) {
...
+//
+// Callers no longer need to hand-order ctx.Queues children-before-parents.
+func deleteQueues(ctx *TestContext) {
+	if len(ctx.Queues) == 0 {
+		return
+	}
+
+	deleteJobsInQueues(ctx, ctx.Queues)
+
+	for _, q := range ctx.Queues {
```
hajnalmt, line 135: What happened here?

hajnalmt, line 135: What happened here? This should do the topLevel close...

test/e2e/util/queue.go
hajnalmt, line 229: There is a really bad design gap here, and I now understand why your AI failed on this miserably. We need to use the Bus instead of `UpdateStatus` like:
```
	for _, q := range topLevelQueues(ctx) {
		err := cliutil.CreateQueueCommand(ctx.Vcclient, "default", q, busv1alpha1.CloseQueueAction)
		Expect(err).NotTo(HaveOccurred(), "failed to send close command to queue %s", q)
```

I don't get it why this is not documented at all. State is not part of the Spec, so we can't edit it by hand, because that won't trigger Admission properly.

hajnalmt (changes requested):
/close

Thanks for the modification, I checked this PR because it was on my backlog, but you messed this up completely in the end 😅. No worries though, I left in my original review notes, but I think I will take over the PRs on the original issue, because the confusion is too big around this issue. The AI agents in the PRs are just hacking their way around, without proper context.

Even I didn't get the full context properly, so this is somewhat ok, this issue is a lot deeper then I expected at first.

I will try to squeze out the proper contributions from the PRs and add you @bp-cheng  and @antcybersec as contibutors in the PR.

PR #5394, opened 2026-06-06
fix(webhook): return admission error when createPatch fails

ManojLamani:
#### What type of PR is this?

/kind bug

This PR adds error handling for `createPatch()` in both pod and job admission mutators.

Previously, errors returned by `createPatch()` were ignored:

```go
patchBytes, _ = createPatch(...)
```

If patch generation failed, the admission request could still be allowed with an empty patch, causing mutation failures to be silently ignored.

This change checks the returned error and returns `util.ToAdmissionResponse(err)` when patch creation fails, making the behavior consistent with the existing error handling pattern used elsewhere in both mutators.

This change only affects error handling and does not modify the existing mutation logic. The same fix is applied to both pod and job mutators because they share the same pattern of ignoring `createPatch()` errors.

AI Usage Disclosure:
I used ChatGPT to help analyze the issue, understand the existing code path, and draft the PR description. All code changes, verification, and final review were performed manually.

```release-note
NONE
```


Aman-Cool:
@ManojLamani, DCO check is failing; please sign off your commits.

PR #5395, opened 2026-06-06
webhook/job: handle error returned by createPatch

ArmxG:
This PR fixes a bug in the job mutation webhook.

When Volcano creates a job, the webhook automatically sets default values (like default queue name, scheduler name, max retry count). The function `createPatch()` does this work and can return an error. But the code was ignoring that error silently with _.

If that error ever happened, the job would be created without any default values being applied, and the webhook would still say "OK, approved" — which is wrong behavior.

This fix makes it properly return an error response instead of silently approving a broken job.


ArmxG:
@wangyang0616 @wpeng102

Hi, could you please take a look at this PR when you have time? It's a small error handling fix in the job mutate webhook. Thank you!
Then just wait. That's all you can do right now.

ArmxG:
@wangyang0616 @wpeng102 

Hi, could you please take a look at this PR when you have time? It's a small error handling fix in the job mutate webhook. Thank you!

 Then just wait. That's all you can do right now.

PR #5402, opened 2026-06-08
fix: fall back minAvailable to replicas when partition minPartitions is omitted

### LINKED ISSUE #5401: `minPartitions` default value 0 causes unintended behavior when `partitionPolicy` is set but `minAvailable` and `minPartitions` are omitted
### Description

When a Volcano Job does not specify any `minAvailable` explicitly, but does specify a `PartitionPolicy` without `minPartitions`, the overall job's `minAvailable` becomes 0 because the default value of `minPartitions` is 0. This leads to unexpected behavior, e.g., the status of the job becomes `Running` and the status of the podgroup becomes `Completed` immediately, even when the pods are still `Pending`. 

1. Submit a Volcano Job like follows:
```yaml
kind: Job
metadata:
  name: test-job
spec:
  schedulerName: volcano
  tasks:
  - name: worker
    replicas: 4
    partitionPolicy:
        totalPartitions: 2
        partitionSize: 2
    template:
      spec:
        containers:
        - image: nginx:latest
          name: container1
          resources:
            requests:
              cpu: 8
```

2. Check the status of the job and the podgroup.

Received:
The pods are all `Pending`, the status of the job is `Running`, and the status of the podgroup is `Completed`.

Expected:
If the pods are all `Pending`, the status of the job should be `Pending`, and the status of the podgroup shoud be `Inqueue`.

latest

_No response_

togettoyou:
#### What type of PR is this?

/kind bug

When a task specifies a `partitionPolicy` but omits both `minAvailable` and `minPartitions`, `minPartitions` defaults to
`0`. The job mutate webhook then defaulted the task's `minAvailable` to `minPartitions * partitionSize = 0`, and
`patchDefaultMinAvailable` summed that into a job-level `minAvailable` of `0`.

A zero `minAvailable` removes the gang lower bound entirely, so the state machine condition
`minAvailable <= running+succeeded+failed` (`0 <= 0`) is immediately satisfied. As a result the Job is reported as
`Running` and the PodGroup as `Completed` while the pods are still `Pending`.

Fixes #5401

```release-note
NONE
```

### DIFF
--- a/pkg/webhooks/admission/jobs/mutate/mutate_job.go
+++ b/pkg/webhooks/admission/jobs/mutate/mutate_job.go
@@ -204,7 +204,11 @@ func mutateSpec(tasks []v1alpha1.TaskSpec, basePath string, job *v1alpha1.Job) *
 
 		if tasks[index].MinAvailable == nil {
 			patched = true
-			if tasks[index].PartitionPolicy != nil {
+			// Only derive minAvailable from the partition policy when minPartitions is
+			// explicitly set (> 0). Otherwise minPartitions defaults to 0, which would
+			// make minAvailable 0 and silently drop the task's gang guarantee, so fall
+			// back to replicas just like a task without a partition policy.
+			if tasks[index].PartitionPolicy != nil && tasks[index].PartitionPolicy.MinPartitions > 0 {
 				minAvailable := tasks[index].PartitionPolicy.MinPartitions * tasks[index].PartitionPolicy.PartitionSize
 				tasks[index].MinAvailable = &minAvailable
 			} else {
--- a/pkg/webhooks/admission/jobs/mutate/mutate_job_test.go
+++ b/pkg/webhooks/admission/jobs/mutate/mutate_job_test.go
@@ -197,9 +197,10 @@ func TestCreatePatchExecution(t *testing.T) {
 					},
 				},
 				{
-					Name:         v1alpha1.DefaultTaskSpec + "3",
-					Replicas:     1,
-					MinAvailable: ptr.To(int32(0)),
+					Name:     v1alpha1.DefaultTaskSpec + "3",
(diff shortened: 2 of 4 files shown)


JesseStutler:
@togettoyou The unit test failed, please check

togettoyou:
It's been fixed. Does CI need to be triggered manually?

pkg/webhooks/admission/jobs/mutate/mutate_job.go
```diff
@@ -204,7 +204,11 @@ func mutateSpec(tasks []v1alpha1.TaskSpec, basePath string, job *v1alpha1.Job) *
 
 		if tasks[index].MinAvailable == nil {
 			patched = true
-			if tasks[index].PartitionPolicy != nil {
+			// Only derive minAvailable from the partition policy when minPartitions is
+			// explicitly set (> 0). Otherwise minPartitions defaults to 0, which would
+			// make minAvailable 0 and silently drop the task's gang guarantee, so fall
+			// back to replicas just like a task without a partition policy.
+			if tasks[index].PartitionPolicy != nil && tasks[index].PartitionPolicy.MinPartitions > 0 {
```
Tau721, line 211: I am wondering what if the user explicitly specifies `minPartitions=0`. In this case, the user intent may be violated.

togettoyou: I don't think it actually causes a problem here, because users who really want `minAvailable=0` can set `minAvailable: 0` directly, instead of expressing it indirectly through `minPartitions=0`.

Tau721: I prefer to define `MinPartitions` as *int32 and mutate `MinPartitions` to `TotalPartitions` when it is unset, to maintain the semantic consistency.

PR #5405, opened 2026-06-08
fix(capacity): count subtree gate-reserved tasks in ancestor quota check

WHOIM1205:
## Summary

Fix hierarchical queue capacity checks to account for reserved tasks from descendant queues.

When SchedulingGatesQueueAdmission is enabled, reserved tasks are tracked only under their leaf queue. However, capacity validation is performed across the entire queue hierarchy. As a result, ancestor queues did not include reserved resources from descendant queues when evaluating capacity limits.

This could allow multiple sibling queues to reserve resources independently and collectively exceed their parent queue's configured capacity before allocation occurs.

## Root Cause

The reserved-task accounting logic only considered the current queue's reservation cache.

For leaf queues this worked correctly, but ancestor queues never saw reservations created by their descendants because reserved tasks were stored exclusively under leaf queue IDs.

As a result, hierarchical capacity checks could underestimate future resource usage during the reservation window.

## Fix

Update reserved resource accounting to aggregate reservations across the entire queue subtree when performing capacity validation.

A new helper traverses descendant queues and includes all reserved tasks in the capacity calculation, ensuring ancestor queues correctly enforce their configured limits.

## Tests

Added a hierarchical capacity test covering the following scenario:

- Parent queue capacity: 100 CPU
- Leaf queue L1 reserves 80 CPU
- Leaf queue L2 submits a new workload

The test verifies that:

- A 20 CPU request is accepted (80 + 20 = 100)
- A 30 CPU request is rejected (80 + 30 > 100)
- Hierarchical capacity checks correctly account for reservations held by sibling queues through their common ancestor

This test fails with the previous implementation and passes with the fix.

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -1636,16 +1636,15 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
 			}
 		}
 	}
-	// Calculate total reserved resources directly from cache
+	// Calculate total reserved resources from cache. Reserved tasks are recorded
+	// under their leaf queue, but capacity is checked per queue across the whole
+	// ancestor chain (see checkQueueAllocatableHierarchically). When this queue is
+	// a non-leaf ancestor, its own bucket is empty, so we must also include the
+	// reserved tasks of every descendant queue; otherwise gate-admitted-but-not-yet
+	// -allocated tasks from sibling leaves are invisible to the ancestor's check and
+	// the subtree can be admitted past the ancestor's realCapability (quota bypass).
 	reserved := api.EmptyResource()
-	if queueGateReserved := cp.queueGateReservedTasks[queue.UID]; queueGateReserved != nil {
-		for _, task := range queueGateReserved {
-			if task.UID != candidate.UID {
-				// Skip candidate to avoid double-counting (it will be added in futureUsed below)
-				reserved.Add(task.Resreq)
-			}
-		}
-	}
+	cp.addReservedForSubtree(attr, candidate, reserved)
 
 	// Include reserved resources in capacity check
 	futureUsed := attr.allocated.Clone().Add(reserved).Add(candidate.Resreq)
@@ -1657,6 +1656,27 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
 	}
 
 	return allocatable
+}
+
+// addReservedForSubtree accumulates into reserved the Resreq of every gate-reserved
+// task recorded for attr's queue and all of its descendant queues, excluding the
+// candidate itself (it is accounted for separately in futureUsed). Reserved tasks
+// are stored per leaf queue, so summing over the subtree lets an ancestor's capacity
+// check account for the reserved tasks of all queues beneath it.
+func (cp *capacityPlugin) addReservedForSubtree(attr *queueAttr, candidate *api.TaskInfo, reserved *api.Resource) {
+	if attr == nil {
+		return
+	}
+	if queueGateReserved := cp.queueGateReservedTasks[attr.queueID]; queueGateReserved != nil {
+		for _, task := range queueGateReserved {
+			if task.UID != candidate.UID {
+				reserved.Add(task.Resreq)
+			}
+		}
+	}
+	for _, child := range attr.children {
+		cp.addReservedForSubtree(child, candidate, reserved)
+	}
 }
 
 func (cp *capacityPlugin) checkQueueAllocatableHierarchically(ssn *framework.Session, queue *api.QueueInfo, candidate *api.TaskInfo) bool {
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -1738,11 +1738,8 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
 	if attr.reservedSubtree != nil {
 		reserved = attr.reservedSubtree.Clone()
 	}
-	// Exclude the candidate if it is itself currently reserved (admitted in a prior cycle
-	// and rebuilt into the cache); it is added separately via futureUsed below. In every
-	// real call site attr is on the candidate's leaf->root path, so a reserved candidate is
-	// always counted in attr.reservedSubtree and must be subtracted exactly once. Membership
-	// and amount come from the same snapshot (see the reservedTasks contract above).
+	// Exclude the candidate from reserved if already counted (avoids double-counting
+	// since it's added separately in futureUsed).
 	if _, ok := reservedTasks[candidate.UID]; ok {
 		if candidate.Resreq.LessEqual(reserved, api.Zero) {
 			reserved.Sub(candidate.Resreq)
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -71,6 +71,11 @@ type capacityPlugin struct {
 	// These tasks reserve queue capacity to prevent other tasks from consuming it
 	// Rebuilt fresh at the start of each scheduling cycle in OnSessionOpen
 	queueGateReservedTasks map[api.QueueID]map[api.TaskID]*api.TaskInfo
+	// reservedTaskQueue is a reverse index from a reserved task to its (leaf) queue.
+	// It gives O(1) membership tests for candidate exclusion and turns
+	// removeTaskFromReservedCache from an O(queues) scan into an O(1) lookup.
+	// Rebuilt together with queueGateReservedTasks in buildQueueReservedTasksCache.
+	reservedTaskQueue map[api.TaskID]api.QueueID
 
 	// dynamicResourceAllocationEnable controls whether DRA quota is enforced
 	dynamicResourceAllocationEnable bool
@@ -98,6 +103,11 @@ type queueAttr struct {
 	guarantee         *api.Resource
 	dra               *draQuotaAttr
 	resourceClaimRefs map[string]int
+	// reservedSubtree is the aggregate Resreq of gate-reserved (admitted but not yet
+	// allocated) tasks in this queue AND every descendant queue. It is maintained
+	// incrementally up the ancestor chain, mirroring how `allocated` is propagated, so
+	// hierarchical capacity checks can read it in O(1) instead of walking the subtree.
+	reservedSubtree *api.Resource
 }
 
 // draQuotaAttr holds DRA quota tracking state for a queue
@@ -425,6 +435,7 @@ func New(arguments framework.Arguments) framework.Plugin {
 		ancestorReclaimLevel:            0,
 		pluginArguments:                 arguments,
 		queueGateReservedTasks:          make(map[api.QueueID]map[api.TaskID]*api.TaskInfo),
+		reservedTaskQueue:               make(map[api.TaskID]api.QueueID),
 		dynamicResourceAllocationEnable: dynamicResourceAllocationEnable,
 		draConsumableCapacityEnable:     draConsumableCapacityEnable,
 	}
@@ -454,6 +465,13 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		klog.V(4).Infof("Hierarchy is enabled in capacity plugin")
 	} else {
 		cp.buildQueueAttrs(ssn)
+	}
+
+	// Seed each queue's reservedSubtree aggregate from the reserved cache. This must run
+	// after the queue attrs (and their ancestor/children links) are built so reservations
+	// recorded under a leaf queue propagate to every ancestor's subtree total.
+	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) {
+		cp.aggregateReservedSubtree()
 	}
 
 	ssn.AddReclaimableFn(cp.Name(), func(reclaimer *api.TaskInfo, reclaimees []*api.TaskInfo) ([]*api.TaskInfo, int) {
@@ -801,11 +819,18 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 
 	ssn.AddPrePredicateFn(cp.Name(), func(task *api.TaskInfo) error {
 		state := &capacityState{
-			queueAttrs: make(map[api.QueueID]*queueAttr),
+			queueAttrs:        make(map[api.QueueID]*queueAttr),
+			reservedTaskQueue: make(map[api.TaskID]api.QueueID, len(cp.reservedTaskQueue)),
 		}
 
 		for _, queue := range cp.queueOpts {
 			state.queueAttrs[queue.queueID] = queue.Clone()
+		}
+		// Snapshot reservation membership together with the per-queue reservedSubtree amounts
+		// (cloned just above) so simulate-time capacity checks read both from this frozen
+		// snapshot rather than mixing it with live plugin state that mutates later this cycle.
+		for taskID, queueID := range cp.reservedTaskQueue {
+			state.reservedTaskQueue[taskID] = queueID
 		}
 
 		ssn.GetCycleState(task.UID).Write(capacityStateKey, state)
@@ -890,7 +915,9 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 
 		simulateQueueAllocatable := func(state *capacityState, queue *api.QueueInfo, candidate *api.TaskInfo) bool {
 			attr := state.queueAttrs[queue.UID]
-			return cp.queueAllocatableWithReserved(attr, candidate, queue, cp.dynamicResourceAllocationEnable, cp.draConsumableCapacityEnable)
+			// Simulate path: attr is the per-task clone, so membership must come from the
+			// snapshot taken alongside it (state.reservedTaskQueue), not the live plugin index.
+			return cp.queueAllocatableWithReserved(attr, candidate, queue, cp.dynamicResourceAllocationEnable, cp.draConsumableCapacityEnable, state.reservedTaskQueue)
 		}
 
 		list := append(state.queueAttrs[queue.UID].ancestors, queue.UID)
@@ -1037,6 +1064,7 @@ func (cp *capacityPlugin) OnSessionClose(ssn *framework.Session) {
 	cp.totalGuarantee = nil
 	cp.queueOpts = nil
 	cp.queueGateReservedTasks = nil
+	cp.reservedTaskQueue = nil
 }
 
 func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
@@ -1064,6 +1092,7 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
 				inqueue:           api.EmptyResource(),
 				guarantee:         api.EmptyResource(),
 				resourceClaimRefs: make(map[string]int),
+				reservedSubtree:   api.EmptyResource(),
 			}
 			if len(queue.Queue.Spec.Capability) != 0 {
 				attr.capability = api.NewResource(queue.Queue.Spec.Capability)
@@ -1418,6 +1447,7 @@ func (cp *capacityPlugin) newQueueAttr(queue *api.QueueInfo) *queueAttr {
 		capability:        api.EmptyResource(),
 		realCapability:    api.EmptyResource(),
 		resourceClaimRefs: make(map[string]int),
+		reservedSubtree:   api.EmptyResource(),
 	}
 	if len(queue.Queue.Spec.Capability) != 0 {
 		attr.capability = api.NewResource(queue.Queue.Spec.Capability)
@@ -1571,7 +1601,9 @@ func (cp *capacityPlugin) isLeafQueue(queueID api.QueueID) bool {
 
 func (cp *capacityPlugin) queueAllocatable(queue *api.QueueInfo, candidate *api.TaskInfo, draEnabled bool, consumableCapacityEnabled bool) bool {
 	attr := cp.queueOpts[queue.UID]
-	return cp.queueAllocatableWithReserved(attr, candidate, queue, draEnabled, consumableCapacityEnabled)
+	// Live path: attr is a live queueOpts entry, so its reservedSubtree and the membership
+	// index cp.reservedTaskQueue share the same (live) source and stay consistent.
+	return cp.queueAllocatableWithReserved(attr, candidate, queue, draEnabled, consumableCapacityEnabled, cp.reservedTaskQueue)
 }
 
 // addTaskToReservedCache adds a task to the reserved cache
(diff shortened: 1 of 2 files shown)


pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1636,16 +1636,15 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
...
 		}
 	}
-	// Calculate total reserved resources directly from cache
+	// Calculate total reserved resources from cache. Reserved tasks are recorded
+	// under their leaf queue, but capacity is checked per queue across the whole
+	// ancestor chain (see checkQueueAllocatableHierarchically). When this queue is
+	// a non-leaf ancestor, its own bucket is empty, so we must also include the
+	// reserved tasks of every descendant queue; otherwise gate-admitted-but-not-yet
+	// -allocated tasks from sibling leaves are invisible to the ancestor's check and
+	// the subtree can be admitted past the ancestor's realCapability (quota bypass).
```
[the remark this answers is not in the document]
hajnalmt: Consider this too if we stay with this implementation.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1636,16 +1636,15 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
...
 	reserved := api.EmptyResource()
-	if queueGateReserved := cp.queueGateReservedTasks[queue.UID]; queueGateReserved != nil {
-		for _, task := range queueGateReserved {
-			if task.UID != candidate.UID {
-				// Skip candidate to avoid double-counting (it will be added in futureUsed below)
-				reserved.Add(task.Resreq)
-			}
-		}
-	}
+	cp.addReservedForSubtree(attr, candidate, reserved)
```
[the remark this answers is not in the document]
hajnalmt: This is true, I would prefer a different solution.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1659,6 +1658,27 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
...
+	if attr == nil {
+		return
+	}
+	if queueGateReserved := cp.queueGateReservedTasks[attr.queueID]; queueGateReserved != nil {
+		for _, task := range queueGateReserved {
+			if task.UID != candidate.UID {
+				reserved.Add(task.Resreq)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
hajnalmt: Consider this comment too please. It never hurts to have guard like this.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1659,6 +1658,27 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
 	return allocatable
 }
 
+// addReservedForSubtree accumulates into reserved the Resreq of every gate-reserved
+// task recorded for attr's queue and all of its descendant queues, excluding the
+// candidate itself (it is accounted for separately in futureUsed). Reserved tasks
+// are stored per leaf queue, so summing over the subtree lets an ancestor's capacity
+// check account for the reserved tasks of all queues beneath it.
+func (cp *capacityPlugin) addReservedForSubtree(attr *queueAttr, candidate *api.TaskInfo, reserved *api.Resource) {
```
hajnalmt: If I understad correctly, we have this call chain:

```text
checkQueueAllocatableHierarchically(...)
  -> iterates leaf + each ancestor
     -> queueAllocatableWithReserved(...)
        -> addReservedForSubtree(...)
           -> recursively walks that queue's descendant subtree
```

Which is is not double-counting within a single `futureUsed` calculation if `queueAttr.children` is a proper tree: `addReservedForSubtree(parent)` visits each descendant once for that parent's check, and the candidate itself is excluded before `candidate.Resreq` is added separately.

However, it does repeatedly visit the same reserved tasks across separate capacity checks for leaf, parent, grandparent, etc. That is expected semantically because each queue's `realCapability` is an independent constraint, but it is a questionable algorithmic shape: cost becomes roughly `sum(size(subtree(checked queue)))` for each candidate, and the implementation relies on recursive subtree walking inside an already hierarchical check.

A cleaner architecture may be to precompute or maintain subtree-reserved resources per queue once per session/reservation update, for example:

- during `buildQueueReservedTasksCache`, add each leaf reservation to the leaf and all ancestors; or
- build a `reservedSubtree[queueID]` map with a post-order traversal after the queue tree is built; or
- update ancestor reserved totals in `addTaskToReservedCache` / `removeTaskFromReservedCache`.

That would make `queueAllocatableWithReserved` a direct lookup instead of recursively walking descendants during every queue check.

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -1959,3 +1959,94 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
 		t.Fatalf("expected enqueueable DRA check to account inqueue resources")
 	}
 }
+
+// TestHierarchicalReservedTasksCountedAtAncestor verifies that gate-reserved tasks
+// recorded under a leaf queue are accounted for when the capacity check evaluates an
+// ancestor queue. Without subtree aggregation, two sibling leaves can each pass their
```
hajnalmt, line 1968: You can delete the second part of this comment starting with the sentence "Without subtree...". It is just LLM overexplanation, the test itself is good and explanatory enough.

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -1959,3 +1959,94 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
...
+	// candidate requesting cpu 20: parent sees 80 (reserved) + 20 = 100 <= 100 -> admit.
+	fitting := &api.TaskInfo{UID: "cand-fit", Name: "cand-fit", Resreq: &api.Resource{MilliCPU: 20000}}
+	// candidate requesting cpu 30: parent sees 80 (reserved) + 30 = 110 > 100 -> reject.
+	exceeding := &api.TaskInfo{UID: "cand-exceed", Name: "cand-exceed", Resreq: &api.Resource{MilliCPU: 30000}}
+
+	leaf2Queue := ssn.Queues[leaf2ID]
+	parentQueue := ssn.Queues[parentID]
+
+	// Leaf L2 in isolation has no reserved tasks, so both candidates fit its own quota.
+	// This is the state that previously let the exceeding candidate slip through, since
```
hajnalmt, line 2045: The first line of this comment is enough. The second and third can be deleted.

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -1959,3 +1959,94 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
...
+	// the parent check ignored L1's reserved tasks.
+	if !cp.queueAllocatable(leaf2Queue, exceeding, false, false) {
+		t.Fatalf("leaf L2 alone should admit the candidate (its own reserved is empty)")
+	}
+
+	// The ancestor must now reject the exceeding candidate because L1's reserved task is
+	// part of the parent's subtree.
+	if cp.queueAllocatable(parentQueue, exceeding, false, false) {
+		t.Fatalf("parent P should reject candidate: 80 (L1 reserved) + 30 > 100 realCapability")
+	}
```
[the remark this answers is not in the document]
hajnalmt: I am fine without this.

hajnalmt (commented):
/area scheduling

Nice find! /cc  @devzizu It looks like we missed the proper hierarchical queue support for the Gate Reserved feature. Would love to hear your thoughts on this.

Thanks for the fix. I see and agree with the bug being fixed here: with scheduling-gates queue admission, ancestor capacity checks need to see gate-reserved tasks from descendant queues, otherwise sibling leaves can collectively reserve past the parent capability.

I am less sure about the algorithmic shape though. `checkQueueAllocatableHierarchically(...)` already walks the leaf and each ancestor, and then `queueAllocatableWithReserved(...)` calls `addReservedForSubtree(...)`, which recursively walks the checked queue's descendants. That does not look like double-counting inside one `futureUsed` calculation, but it does mean we repeatedly traverse overlapping subtrees for a single candidate.  We are looping from descendant to ancestors in the check and for each ancestor we will do an ancestor to descendants loop additionaly now.

Wouldn't it be cleaner to precompute or maintain subtree-reserved resources per queue instead? For example, when rebuilding `queueGateReservedTasks`, accumulate each leaf reservation into the leaf and its ancestors, or build a `reservedSubtree[queueID]` map once per session. Then the allocatable check can use a direct per-queue reserved total instead of recursively walking descendants inside every hierarchical check.

Could you please also rebase the PR? It is currently behind `master`.

We can decide that this is good enough, but I wanted to raise the discussion point. Thank you for the PR once more!

devzizu:
Nice catch @WHOIM1205 and also thanks @hajnalmt for the feedback here! We've totally missed this for the initial design, and my tests only covered the flat queue structure.

I would really benefit from this fix! Given @hajnalmt suggestion, what about the following:

1. Add a precomputed field to `queueAttr`:

    ```go
    // reservedSubtree tracks gate-reserved resources for this queue + all descendants
    reservedSubtree *api.Resource
    ```

2. In `buildQueueReservedTasksCache`, roll up to ancestors when scanning leaf tasks (same pattern as `AllocateFunc`):

    ```go
    // ...
    cp.queueGateReservedTasks[job.Queue][task.UID] = task
    // ...previous code ^^
    // After adding task to leaf cache:
    attr.reservedSubtree.Add(task.Resreq)
    for _, ancestorID := range attr.ancestors {
        cp.queueOpts[ancestorID].reservedSubtree.Add(task.Resreq)
    }
    ```

3. Same rollup/rolldown in`addTaskToReservedCache` and `removeTaskFromReservedCache`:

    ```go
    // addTaskToReservedCache, after map insert:
    attr.reservedSubtree.Add(task.Resreq)
    for _, ancestorID := range attr.ancestors {
        cp.queueOpts[ancestorID].reservedSubtree.Add(task.Resreq)
    }
    ```

    ```go
    // removeTaskFromReservedCache, after map delete:
    attr.reservedSubtree.Sub(task.Resreq)
    for _, ancestorID := range attr.ancestors {
        cp.queueOpts[ancestorID].reservedSubtree.Sub(task.Resreq)
    }
    ```

4. Simplify`queueAllocatableWithReserved` with a **direct read** instead of map iteration or tree walk:

    ```go
    reserved := attr.reservedSubtree.Clone()
    // Exclude candidate if already in cache to avoid double-counting
    if _, exists := cp.queueGateReservedTasks[queue.UID][candidate.UID]; exists {
        reserved.Sub(candidate.Resreq)
    }
    futureUsed := attr.allocated.Clone().Add(reserved).Add(candidate.Resreq)
    ```

   This keeps capacity checks `O(1)`, is **consistent with how allocated handles hierarchy**, and the `DeallocateFunc` rollback path gets ancestor accounting for free since it already calls `addTaskToReservedCache`.

Since the logic of 2 and 3 ends up being so similar, we could even have a `updateReservedSubtree` helper and reuse it with `cp.updateReservedSubtree(job.Queue, task.Resreq...)` in `buildQueueReservedTasksCache`, `addTaskToReservedCache` and `removeTaskFromReservedCache`.

What do you think? @WHOIM1205 @hajnalmt

WHOIM1205:
Thanks for the feedback @hajnalmt @devzizu

I've updated the implementation based on the review comments and pushed a new version.

The latest changes use aggregated subtree reservation accounting, keep reservation membership consistent on the simulate path, and add regression tests covering the reported edge cases.

Could you please take another look when you have a chance?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1764,6 +1878,10 @@ func (qa *queueAttr) Clone() *queueAttr {
 		guarantee:         qa.guarantee.Clone(),
 		resourceClaimRefs: make(map[string]int, len(qa.resourceClaimRefs)),
 		children:          make(map[api.QueueID]*queueAttr),
+		reservedSubtree:   api.EmptyResource(),
+	}
+	if qa.reservedSubtree != nil {
+		cloned.reservedSubtree = qa.reservedSubtree.Clone()
```
devzizu: Since `reservedSubtree` is always initialized (never `nil` after construction), the nil check + pre-initialization is redundant. Could simplify to just `reservedSubtree: qa.reservedSubtree.Clone()`.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1580,25 +1612,37 @@ func (cp *capacityPlugin) addTaskToReservedCache(queueID api.QueueID, task *api.
 	if cp.queueGateReservedTasks[queueID] == nil {
 		cp.queueGateReservedTasks[queueID] = make(map[api.TaskID]*api.TaskInfo)
 	}
+	// Adding to a map is idempotent, but adding to the reservedSubtree aggregate is not.
+	// AllocatableFn may admit the same candidate more than once within a cycle (multiple
+	// node attempts) and the rollback path re-adds it, so guard against double counting.
+	if _, exists := cp.queueGateReservedTasks[queueID][task.UID]; exists {
+		return
+	}
```
devzizu: Good catch! :) 

Consider covering this guard under a unit test to ensure we cover a possible regression here.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1636,14 +1728,30 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
...
-				reserved.Add(task.Resreq)
-			}
+	if attr.reservedSubtree != nil {
+		reserved = attr.reservedSubtree.Clone()
+	}
+	// Exclude the candidate if it is itself currently reserved (admitted in a prior cycle
+	// and rebuilt into the cache); it is added separately via futureUsed below. In every
+	// real call site attr is on the candidate's leaf->root path, so a reserved candidate is
+	// always counted in attr.reservedSubtree and must be subtracted exactly once. Membership
+	// and amount come from the same snapshot (see the reservedTasks contract above).
```
devzizu: nit: I suggest simplifying this comment

```suggestion
	// Exclude the candidate from reserved if already counted (avoids double-counting
    // since it's added separately in futureUsed).
```

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -1959,3 +1959,272 @@ func TestDRAAllocatableIgnoresInqueue(t *testing.T) {
...
+	cp.addTaskToReservedCache(leafQueue.UID, task)
+	cp.addTaskToReservedCache(leafQueue.UID, task)
+	cp.addTaskToReservedCache(leafQueue.UID, task)
+
+	if got := cp.queueOpts[leafQueue.UID].reservedSubtree.MilliCPU; got != 30000 {
+		t.Fatalf("leaf reservedSubtree = %v after repeated adds, want cpu 30", got)
+	}
+	if got := cp.queueOpts[parentQueue.UID].reservedSubtree.MilliCPU; got != 30000 {
+		t.Fatalf("parent reservedSubtree = %v after repeated adds, want cpu 30", got)
+	}
```
devzizu, line 2182: Here, we can add a check that the `cp.queueGateReservedTasks` didn't accumulate entries:

```go
if got := len(cp.queueGateReservedTasks[leafQueue.UID]); got != 1 {
   t.Fatalf("task map has %d entries, want 1", got)
}
```

WHOIM1205:
thanks for the review @devzizu @hajnalmt!

I've addressed the latest comments and pushed an update:

- Simplified the exclusion comment as suggested.
- Extended the existing idempotency test to cover the guard and verify that `queueGateReservedTasks` does not accumulate duplicate entries when the same task is added repeatedly.

Could you please take another look when you have a chance?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -71,6 +71,11 @@ type capacityPlugin struct {
 	// These tasks reserve queue capacity to prevent other tasks from consuming it
 	// Rebuilt fresh at the start of each scheduling cycle in OnSessionOpen
 	queueGateReservedTasks map[api.QueueID]map[api.TaskID]*api.TaskInfo
+	// reservedTaskQueue is a reverse index from a reserved task to its (leaf) queue.
+	// It gives O(1) membership tests for candidate exclusion and turns
```
hajnalmt: Leave just the first line of comment here please. That's enough.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -98,6 +103,11 @@ type queueAttr struct {
 	guarantee         *api.Resource
 	dra               *draQuotaAttr
 	resourceClaimRefs map[string]int
+	// reservedSubtree is the aggregate Resreq of gate-reserved (admitted but not yet
+	// allocated) tasks in this queue AND every descendant queue. It is maintained
```
hajnalmt: The first sentence is enough, the rest is just LLM overexplanation.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -456,6 +467,13 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		cp.buildQueueAttrs(ssn)
 	}
 
+	// Seed each queue's reservedSubtree aggregate from the reserved cache. This must run
+	// after the queue attrs (and their ancestor/children links) are built so reservations
+	// recorded under a leaf queue propagate to every ancestor's subtree total.
+	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) {
```
hajnalmt:
```suggestion
	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulingGatesQueueAdmission) && hierarchyEnabled {
```

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1619,14 +1664,61 @@ func (cp *capacityPlugin) buildQueueReservedTasksCache(ssn *framework.Session) {
...
 					task.Namespace, task.Name, job.Queue)
 			}
 		}
 	}
 }
 
-func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidate *api.TaskInfo, queue *api.QueueInfo, draEnabled bool, consumableCapacityEnabled bool) bool {
+// aggregateReservedSubtree seeds every queue's reservedSubtree aggregate from the freshly
+// built reserved-tasks cache. It MUST run after the queue attrs and their ancestor/children
+// links are built (buildHierarchicalQueueAttrs / buildQueueAttrs), so that reservations
```
hajnalmt: This is needed only for buildHierarchicalQueueAttrs, we shouldn't do it in the non hierarchical cases and please don't over explain in the comment too.

pkg/scheduler/plugins/capacity/capacity.go
hajnalmt: Nice catch truly. 👍

hajnalmt (changes requested):
Thanks for reworking this toward the `reservedSubtree` approach. The main live-path concern from the earlier review looks addressed now: reservation totals are maintained per queue/ancestor, and the allocatable check uses a direct aggregate rather than recursively walking descendants.

I found one remaining consistency issue in the simulate path though. The live path moves a task from reserved -> allocated in `AllocateFunc` by adding to `allocated` and then calling `removeTaskFromReservedCache(...)`; rollback does the inverse in `DeallocateFunc` by adding the reservation back.

The snapshot path does not mirror that transition yet: `AddSimulateAddTaskFn` adds `taskToAdd.Resreq` into `state.queueAttrs[*].allocated`, but if `taskToAdd` was already present in `state.reservedTaskQueue` / `reservedSubtree`, it stays reserved in the snapshot as well. A later `SimulateAllocatableFn` then sees that previous simulated task as both allocated and reserved. The candidate self-exclusion only subtracts the candidate being checked, not other reserved tasks that were simulated as allocated earlier in the same cycle state.

Could we update the simulate add/remove handlers to keep the snapshot reservation state consistent with the live handlers?

Also the whole implementation misses the point that this subtree aggregation should happen only for hierarchical cases. **We shouldn't do this aggregation on non hiearachical queues, as parent relationships can exist but not be treated as such in those cases.** The scope shall be narrowed down a little bit, or do I miss something here?

WHOIM1205:
Thanks for the review @hajnalmt @devzizu 

Good catch on the simulate-path handling.

I traced the preemption flow and found that the queue allocation gate annotation persists on running pods after allocation, while the reserved-task snapshot only contains pending tasks. As a result, restoring reservations during SimulateRemoveTaskFn could incorrectly reintroduce an allocated victim into the reserved accounting.

I've removed the simulate-path reservation mutations and kept the reservation snapshot immutable during simulation. I also replaced the previous helper-based test with a session-level regression test that exercises the actual PrePredicateFn → SimulateRemoveTaskFn → SimulateAllocatableFn path and reproduces the issue.

The updated test fails with the buggy behavior and passes with the fix.

I've pushed the changes. Could you please take another look?

pkg/scheduler/plugins/capacity/capacity.go
devzizu: @WHOIM1205 I believe this is not yet resolved, can you please check?

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1571,62 +1589,102 @@ func (cp *capacityPlugin) isLeafQueue(queueID api.QueueID) bool {
...
+// after the queue attrs and their ancestor links are built.
+func (cp *capacityPlugin) aggregateReservedSubtree() {
+	for leafID, tasks := range cp.queueGateReservedTasks {
+		for _, task := range tasks {
+			addReservedSubtree(cp.queueOpts, leafID, task.Resreq)
+		}
+	}
+}
+
+func addReservedSubtree(queueOpts map[api.QueueID]*queueAttr, leafID api.QueueID, res *api.Resource) {
```
devzizu: nit: `addReservedSubtree`/`subReservedSubtree` are the only package-level functions here and everything else is a method on `*capacityPlugin`. Since `queueOpts` is already on the receiver, consider making them methods (`cp.addReservedSubtree(queueID, res)`) to drop the param and stay consistent with the rest of the file.

WDYT?

WHOIM1205:
Thanks for the review @devzizu 

Good catch on the Clone path. I cleaned up the reservedSubtree clone to follow the same pattern as the other resource fields.

I also converted addReservedSubtree/subReservedSubtree into capacityPlugin methods and updated the call sites for consistency with the rest of the file.

The branch has been updated. Could you please take another look when you have a chance?

devzizu:
> Thanks for the review @devzizu 
> 
> Good catch on the Clone path. I cleaned up the reservedSubtree clone to follow the same pattern as the other resource fields.
> 
> I also converted addReservedSubtree/subReservedSubtree into capacityPlugin methods and updated the call sites for consistency with the rest of the file.
> 
> The branch has been updated. Could you please take another look when you have a chance?

The implementation looks good to me. Let's wait for @hajnalmt review. Thank you

devzizu:
@JesseStutler @hajnalmt if you have a bit of time, can you please take a look at this issue?

JesseStutler:
> @JesseStutler @hajnalmt if you have a bit of time, can you please take a look at this issue?

Yes of course!
/cc

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1636,12 +1694,24 @@ func (cp *capacityPlugin) queueAllocatableWithReserved(attr *queueAttr, candidat
...
+		if _, ok := reservedTasks[candidate.UID]; ok {
+			if candidate.Resreq.LessEqual(reserved, api.Zero) {
+				reserved.Sub(candidate.Resreq)
+			} else {
+				// Skip exclusion on the (snapshot-prevented) underflow rather than panic in Resource.Sub.
+				klog.Warningf("[capacity] queue <%v>: reserved subtree <%v> does not cover reserved candidate <%v> request <%v>; skipping candidate exclusion",
+					queue.Name, reserved, candidate.Name, candidate.Resreq)
+			}
+		}
+	} else if queueGateReserved := cp.queueGateReservedTasks[queue.UID]; queueGateReserved != nil {
```
JesseStutler, line 1712: Hierarchy uses `reservedSubtree` while flat queues use`queueGateReservedTasks`, I think we should unify them, WDYT? @devzizu @WHOIM1205  They both deal with "reserved" calculations, yet the logic and fields differ. It makes other contributors hard to read

JesseStutler:
And needs to resolve the conflict

pkg/scheduler/plugins/capacity/capacity.go
devzizu: Agreed, and I think we can unify and the flat case is just a subtree with no children. My preference would be to keep `queueGateReservedTasks` as the single source of truth and make the check subtree-aware, rather than maintaining a separate aggregate, but we should also be careful performance-wise.

PR #5411, opened 2026-06-09
feature(svc plugin): support network_policy_strategy

### LINKED ISSUE #5410: feat(volcano svc plugin) support network_policy_strategy of the NetworkPolicy

Sometimes the user needs to specify the `network_policy_strategy` of the `NetworkPolicy` created by `svc` job plugin.

Add --network-policy-strategy arg to the `svc` plugin. I'd like to do this work.

### Additional context

_No response_

### Documentation Updates

kingeasternsun:
#### What type of PR is this?

/kind feature

This PR adds a new optional parameter `network-policy-strategy` to the `svc` job plugin for OVN-Kubernetes integration.

When users specify `--network-policy-strategy`, the `NetworkPolicy` created by the svc plugin will include the annotation `ovn.kubernetes.io/network_policy_strategy` with the configured value. If the parameter is not specified, behavior is unchanged (no annotation is added).

Changes include:
- **svc plugin**: parse `--network-policy-strategy` and set the OVN annotation on created `NetworkPolicy` objects
- **Admission validation**: reject invalid values via ValidatingAdmissionPolicy (VAP) and the job validating webhook (when VAP is disabled)
- **Documentation**: update svc plugin user guide and example
- **Tests**: unit tests for svc plugin flag parsing and webhook validation

Allowed values: `allow`, `allow-related`, `allow-stateless`, `drop`, `reject`.

Example usage:

```yaml
plugins:
  svc: ["--network-policy-strategy=allow-related"]
```

Fixes #5410 

- AI tools (Cursor) were used to assist with implementation, tests, and documentation. All changes have been reviewed and verified locally.
- `network-policy-strategy` is optional; omitting it does not trigger validation errors and does not add the annotation.
- VAP validation only checks the `--network-policy-strategy=<value>` format; the webhook additionally supports `--network-policy-strategy <value>`.

```release-note
Add optional `network-policy-strategy` parameter to the svc job plugin. When set, the created NetworkPolicy includes the `ovn.kubernetes.io/network_policy_strategy` annotation. Valid values are `allow`, `allow-related`, `allow-stateless`, `drop`, and `reject`.
```

### DIFF
--- a/pkg/controllers/job/plugins/svc/const.go
+++ b/pkg/controllers/job/plugins/svc/const.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Volcano Authors.
+Copyright 2026 The Volcano Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,6 +17,9 @@ limitations under the License.
 package svc
 
 const (
+	// NetworkPolicyStrategyAnnotation is the OVN annotation key for network policy strategy.
+	NetworkPolicyStrategyAnnotation = "ovn.kubernetes.io/network_policy_strategy"
+
 	// ConfigMapTaskHostFmt key in config map
 	ConfigMapTaskHostFmt = "%s.host"
 	// EnvTaskHostFmt is the key for host list in environment
--- a/docs/user-guide/how_to_use_svc_plugin.md
+++ b/docs/user-guide/how_to_use_svc_plugin.md
@@ -26,12 +26,16 @@ host files under the directory `/etc/volcano/`.
 * A headless service whose name is the same with job will be created.
 * If `disable-network-policy` is set to be false, a `NetworkPolicy` object with the type `Ingress` will be created for
 the job.
+* If `network-policy-strategy` is specified, the created `NetworkPolicy` will include annotation
+`ovn.kubernetes.io/network_policy_strategy` with the specified value. This annotation is used by OVN-Kubernetes to
+control network policy enforcement strategy.
 
 ## Arguments
-| ID  | Name                          | Value           | Default Value | Required | Description                                          | Example                                       |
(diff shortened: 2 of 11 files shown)


pkg/controllers/job/plugins/svc/const.go
```diff
@@ -17,6 +17,9 @@ limitations under the License.
 package svc
 
 const (
+	// NetworkPolicyStrategyAnnotation is the OVN annotation key for network policy strategy.
+	NetworkPolicyStrategyAnnotation = "ovn.kubernetes.io/network_policy_strategy"
```
hzxuzhonghu, line 21: I think we should not make volcano svc plugin aware of ovn-kubernetes. The whole thing should be agonostic, but first please let us understand your use case, you wan to use ovn-kubernetes achive what features?

kingeasternsun: @hzxuzhonghu 
Our use case is not OVN-Kubernetes-specific. The issue is that after enabling the Service plugin, the default NetworkPolicy created by Volcano restricts network access for the job pods.

Although pods within the same job can communicate with each other, many training workloads also need outbound access to external resources such as datasets, object storage, model registries, or other services. The current default policy may block such traffic and cause the workload to fail.

What we are looking for is a way to control or customize the generated NetworkPolicy, rather than introducing OVN-Kubernetes-specific logic into the Service plugin.

kingeasternsun: @hzxuzhonghu Alternatively, I can implement a more generic solution that allows annotations and labels to be configured on the generated NetworkPolicy via plugin parameters.

hzxuzhonghu: Got to understand the use case. Can you use `--disableNetworkPolicy` to remove the networkpolicy?

PR #5414, opened 2026-06-09
check gpuDevice nil in node_info

DSFans2014:

If `vgpu.NewGPUDevices(ni.Name, node)` returns nil, calling `GetIgnoredDevices` will cause a panic

```release-note

```


archlitchi (commented):
please adjust the UT for node_info accordingly

archlitchi (approved):
/lgtm

PR #5416, opened 2026-06-09
Clear device-related annotations during pod release

### LINKED ISSUE #5335: vgpu: stale annotations from dry-run rollback cause GPU over-subscription in subGroup scheduling
### Description

When using subGroupPolicy in a PodGroup ( kthena's 2P1D disaggregated inference scenario), if the scheduler performs a dry-run and then triggers a Discard (e.g., due to unsatisfied networkTopology constraints), RecoverOperations does not clean up the device-assignment annotations (such as `volcano.sh/vgpu-devices-to-allocate`) that were written to the Pod during the dry-run. When the PodGroup is re-scheduled, the alreadyAssignedOnNode function detects these stale annotations and skips checkNodeGPUSharingPredicateAndScore capacity validation entirely. This causes multiple Pods requesting cores=100 to be assigned to the same GPU, resulting in memory over-subscription and eventual OOM failures.

1. The cluster has only one node equipped with 4x NVIDIA A800 80GB GPUs (Memory=81920MiB), all idle — no Pods are configured or running on any GPU.
2. Create a 2P1D inference service using kthena. Each Pod has the following resource limits, with gang scheduling (subGroupPolicy) and networkTopology

```yaml
apiVersion: workload.serving.volcano.sh/v1alpha1
kind: ModelServing
metadata:
  name: qwen3-pd
  namespace: kthena-system
spec:
  recoveryPolicy: ServingGroupRecreate
  replicas: 1
  schedulerName: volcano
  template:
    gangPolicy:
      minRoleReplicas:
        prefill: 1
        decode: 1
    networkTopology:
      groupPolicy:
        highestTierAllowed: 2
        mode: hard
    restartGracePeriodSeconds: 60
    roles:
      - name: prefill
        replicas: 2
        entryTemplate:
          metadata:
            annotations:
              volcano.sh/vgpu-mode: hami-core
          spec:
            containers:
              - name: prefill
                image: ghcr.io/volcano-sh/vllm-openai:v0.10.0-cu128-nixl-v0.4.1-lmcache-0.3.2
                resources:
                  limits:
                    volcano.sh/vgpu-cores: 100
                    volcano.sh/vgpu-memory: 8192
                    volcano.sh/vgpu-number: 1
              ...
        workerReplicas: 0
      - name: decode
        replicas: 1
        entryTemplate:
          metadata:
            annotations:
              volcano.sh/vgpu-mode: hami-core
          spec:
            containers:
              - name: decode
                image: ghcr.io/volcano-sh/vllm-openai:v0.10.0-cu128-nixl-v0.4.1-lmcache-0.3.2
                resources:
                  limits:
                    volcano.sh/vgpu-cores: 100
                    volcano.sh/vgpu-memory: 8192
                    volcano.sh/vgpu-number: 1
                 ...
        workerReplicas: 0  
```

3. The kthena controller automatically creates a PodGroup with the following spec:

```yaml
kind: PodGroup
metadata:
  annotations:
    scheduling.k8s.io/group-name: qwen3-pd-0
  labels:
    modelserving.volcano.sh/group-name: qwen3-pd-0
    modelserving.volcano.sh/name: qwen3-pd
  name: qwen3-pd-0
  namespace: kthena-system
  ownerReferences:
  - apiVersion: workload.serving.volcano.sh/v1alpha1
    controller: true
    kind: ModelServing
    name: qwen3-pd
    uid: xxxx
spec:
  minMember: 2
  minResources: {}
  networkTopology:
    highestTierAllowed: 2
    mode: hard
  queue: default
  subGroupPolicy:
  - labelSelector:
      matchLabels:
        modelserving.volcano.sh/name: qwen3-pd
        modelserving.volcano.sh/role: prefill
      matchLabelKeys:
      - modelserving.volcano.sh/role-id
      minSubGroups: 1
      name: prefill
      subGroupSize: 1
  - labelSelector:
      matchLabels:
        modelserving.volcano.sh/name: qwen3-pd
        modelserving.volcano.sh/role: decode
      matchLabelKeys:
      - modelserving.volcano.sh/role-id
      minSubGroups: 1
      name: decode
      subGroupSize: 1
```

All 3 Pods (2 prefill + 1 decode) are correctly scheduled to the GPU node, but all 3 are assigned to the same GPU card. (Confirmed by checking the `volcano.sh/vgpu-ids-new` annotation on each Pod.)

Each Pod ends up with insufficient GPU memory to start vllm, causing failures. (I know this can be worked around by adjusting vllm's gpu_memory_utilization parameter, but that's beside the point.)

4687fce2dc41d3e5f56d38f34db559fe9f725ff0

_No response_

archlitchi:
/kind bug

fix https://github.com/volcano-sh/volcano/issues/5335

This applies to all sharable devices(NV, Ascend, etc..), so we need a unified fix instead of fixing them one by one.

### DIFF
--- a/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
+++ b/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
@@ -278,9 +278,31 @@ func (gs *GPUDevices) HasDeviceRequest(pod *v1.Pod) bool {
 }
 
 func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) error {
+	if gs == nil || pod == nil || pod.Annotations == nil {
+		return nil
+	}
 	// Release is required for rollback paths (e.g. UnPipeline) where NodeInfo
 	// does not invoke subResource for Pipelined tasks.
 	gs.SubResource(pod)
+
+	if pod.Annotations[DeviceBindPhase] == "success" {
+		return nil
+	}
+
+	keys := []string{
+		AssignedNodeAnnotations,          // volcano.sh/vgpu-node
+		AssignedIDsAnnotations,           // volcano.sh/vgpu-ids-new
+		AssignedIDsToAllocateAnnotations, // volcano.sh/devices-to-allocate
+		AssignedTimeAnnotations,          // volcano.sh/vgpu-time
+		BindTimeAnnotations,              // volcano.sh/bind-time
+		DeviceBindPhase,                  // volcano.sh/bind-phase
+	}
+	if err := devices.RemovePodAnnotations(kubeClient, pod, keys); err != nil {
+		return err
+	}
+	for _, k := range keys {
+		delete(pod.Annotations, k)
+	}
 	return nil
 }
 
--- a/pkg/scheduler/api/devices/ascend/hami/device_info.go
+++ b/pkg/scheduler/api/devices/ascend/hami/device_info.go
@@ -335,6 +335,33 @@ func (ads *AscendDevices) Allocate(kubeClient kubernetes.Interface, pod *v1.Pod)
 }
 
 func (ads *AscendDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) error {
+	if ads == nil || pod == nil || pod.Annotations == nil {
+		return nil
+	}
+	ads.SubResource(pod)
+	if pod.Annotations[util.DeviceBindPhase] == DeviceBindSuccess {
+		return nil
+	}
+	keys := []string{
+		util.AssignedNodeAnnotations,
+		util.AssignedTimeAnnotations,
+		util.DeviceBindPhase,
+		util.BindTimeAnnotations,
+		"predicate-time",
+	}
+	for commonWord := range devices.InRequestDevices {
+		keys = append(keys,
+			devices.InRequestDevices[commonWord],
+			devices.SupportDevices[commonWord],
+			fmt.Sprintf("huawei.com/%s", commonWord),
+		)
+	}
+	if err := devices.RemovePodAnnotations(kubeClient, pod, keys); err != nil {
+		return err
+	}
+	for _, k := range keys {
+		delete(pod.Annotations, k)
+	}
 	return nil
 }
 
--- a/pkg/scheduler/api/devices/ascend/hami/device_info_test.go
+++ b/pkg/scheduler/api/devices/ascend/hami/device_info_test.go
@@ -17,17 +17,20 @@ limitations under the License.
 package hami
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
 	"volcano.sh/volcano/pkg/scheduler/api/devices"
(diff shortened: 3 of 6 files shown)


pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
```diff
@@ -278,9 +278,31 @@ func (gs *GPUDevices) HasDeviceRequest(pod *v1.Pod) bool {
...
+
+	keys := []string{
+		AssignedNodeAnnotations,          // volcano.sh/vgpu-node
+		AssignedIDsAnnotations,           // volcano.sh/vgpu-ids-new
+		AssignedIDsToAllocateAnnotations, // volcano.sh/devices-to-allocate
+		AssignedTimeAnnotations,          // volcano.sh/vgpu-time
+		BindTimeAnnotations,              // volcano.sh/bind-time
+		DeviceBindPhase,                  // volcano.sh/bind-phase
+	}
+	if err := devices.RemovePodAnnotations(kubeClient, pod, keys); err != nil {
```
nishitnshah, line 300: **Efficiency:** `RemovePodAnnotations` always issues a PATCH for all keys, even when the pod has none of them. For a vGPU pod (`HasDeviceRequest` is resource-based, so always true), any deallocation where the pod has a non-nil annotations map, `bind-phase != "success"`, and none of the 6 device keys present results in an apiserver round-trip of harmless `null`-deletes for nothing.

Guarding on "at least one target key present" avoids the wasted call. Cleanest to put it inside `RemovePodAnnotations` so it covers all device `Release` paths (vgpu / hami / vnpu) at once:

```go
func RemovePodAnnotations(kubeClient kubernetes.Interface, pod *v1.Pod, keys []string) error {
	if kubeClient == nil || pod == nil {
		return nil
	}
	// Skip the apiserver call if the pod carries none of the target keys.
	present := false
	for _, k := range keys {
		if _, ok := pod.Annotations[k]; ok {
			present = true
			break
		}
	}
	if !present {
		return nil
	}
	// ... existing patch logic
}
```

Note the common speculative-rollback case (`unallocate`/`unPipeline` after `Allocate` wrote the keys) still patches as expected — the keys are present there. This only eliminates the no-op PATCH when no device annotations exist.

We should avoid unnecessary patches to avoid getting into apiserver throttling issues in future.

nishitnshah (changes requested):

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
archlitchi: added

archlitchi:
CC @JesseStutler

nishitnshah (approved):

JesseStutler:
/approve
/lgtm
Do we need to backport? @archlitchi

nishitnshah:
This needs to be cherry picked in release-1.14 as well. @JesseStutler @archlitchi

JesseStutler:
> This needs to be cherry picked in release-1.14 as well. @JesseStutler @archlitchi

Let me allow copilot to solve the conflict and take a look whether he can solve it

PR #5422, opened 2026-06-10
Fix unbounded job.Status.Conditions growth

### LINKED ISSUE #5421: job.Status.Conditions grows unbounded, eventually causing UpdateStatus to fail with etcd object-size exceeded
Was debugging a VCJob stuck in a permanent reconcile loop and noticed the conditions slice had over 2000 entries. Traced it back to the job controller, every phase transition unconditionally appends a new condition with no cap...

```
// pkg/controllers/job/job_controller_actions.go
// same pattern in killPods (L235), syncJob (L426, L612), killTarget (L970)
job.Status.Conditions = append(job.Status.Conditions, jobCondition)

```

There's no dedup, no max-length check, nothing. newCondition just stamps the current phase and time and hands it back. For a job that uses a RestartJob policy and retries a lot this becomes a real problem, the object grows until UpdateStatus returns:
`etcdserver: request is too large`

At that point the controller can't write status updates at all. The job stays stuck, kubectl describe stops showing new events, and the only way out is to delete the job manually.

Rough fix: cap the slice to the last N entries before the update call, or at minimum skip the append when the new condition's phase equals the previous one's. Something like:
const maxJobConditions = 32

```
func appendJobCondition(conditions []batch.JobCondition, c batch.JobCondition) []batch.JobCondition {
    if n := len(conditions); n > 0 && conditions[n-1].Status == c.Status {
        return conditions
    }
    conditions = append(conditions, c)
    if len(conditions) > maxJobConditions {
        conditions = conditions[len(conditions)-maxJobConditions:]
    }
    return conditions
}
```
Affected versions: master, v1.10.x (likely older too, the append pattern has been there a while)

avinxshKD:
#### What type of PR is this?
/kind bug

This PR fixes an issue where `job.Status.Conditions` grows unbounded during frequent job phase transitions (especially for jobs with `RestartJob` policies that retry often). Previously, every phase transition unconditionally appended a new condition without deduplication or a maximum length check. 

For heavily retried jobs, this slice could easily accumulate thousands of entries, eventually causing the `UpdateStatus` call to permanently fail with `etcdserver: request is too large`. At this point, the job becomes unrecoverable without manual deletion.

This fix introduces an `appendJobCondition` helper that:
1. Skips appending if the new condition's phase is identical to the most recent one (deduplicating redundant consecutive states).
2. Enforces a hard cap of `maxJobConditions = 32`, truncating the oldest entries to ensure the etcd object size remains bounded and safe.

Fixes #5421

The `maxJobConditions` is set to 32, which is more than enough to capture the recent historical context of a job's state transitions without risking etcd size limits.

```release-note
Fixed an issue where frequently restarting jobs could exceed etcd object size limits by capping `job.Status.Conditions` to a maximum of 32 entries and deduplicating consecutive identical phases.

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -1077,7 +1077,9 @@ func appendJobCondition(conditions []batch.JobCondition, c batch.JobCondition) [
 	}
 	conditions = append(conditions, c)
 	if len(conditions) > maxJobConditions {
-		conditions = conditions[len(conditions)-maxJobConditions:]
+		newConditions := make([]batch.JobCondition, maxJobConditions)
+		copy(newConditions, conditions[len(conditions)-maxJobConditions:])
+		return newConditions
 	}
 	return conditions
 }
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -1073,6 +1073,7 @@ const maxJobConditions = 32
 // only the most recent maxJobConditions entries are kept.
 func appendJobCondition(conditions []batch.JobCondition, c batch.JobCondition) []batch.JobCondition {
 	if n := len(conditions); n > 0 && conditions[n-1].Status == c.Status {
+		conditions[n-1] = c
 		return conditions
 	}
 	conditions = append(conditions, c)
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -232,7 +232,7 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 		if updateStatus(&job.Status) {
 			job.Status.State.LastTransitionTime = metav1.Now()
 			jobCondition := newCondition(job.Status.State.Phase, &job.Status.State.LastTransitionTime)
-			job.Status.Conditions = append(job.Status.Conditions, jobCondition)
+			job.Status.Conditions = appendJobCondition(job.Status.Conditions, jobCondition)
 		}
 	}
 
@@ -423,7 +423,7 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 		}
 		job.Status.State.LastTransitionTime = metav1.Now()
 		jobCondition = newCondition(job.Status.State.Phase, &job.Status.State.LastTransitionTime)
-		job.Status.Conditions = append(job.Status.Conditions, jobCondition)
+		job.Status.Conditions = appendJobCondition(job.Status.Conditions, jobCondition)
 		newJob, err := cc.vcClient.BatchV1alpha1().Jobs(job.Namespace).UpdateStatus(context.TODO(), job, metav1.UpdateOptions{})
 		if err != nil {
 			klog.Errorf("Failed to update status of Job %v/%v: %v",
@@ -609,7 +609,7 @@ func (cc *jobcontroller) syncJob(jobInfo *apis.JobInfo, updateStatus state.Updat
 	job.Status = newStatus
 	job.Status.State.LastTransitionTime = metav1.Now()
 	jobCondition = newCondition(job.Status.State.Phase, &job.Status.State.LastTransitionTime)
-	job.Status.Conditions = append(job.Status.Conditions, jobCondition)
+	job.Status.Conditions = appendJobCondition(job.Status.Conditions, jobCondition)
 	newJob, err := cc.vcClient.BatchV1alpha1().Jobs(job.Namespace).UpdateStatus(context.TODO(), job, metav1.UpdateOptions{})
 	if err != nil {
 		klog.Errorf("Failed to update status of Job %v/%v: %v",
@@ -967,7 +967,7 @@ func (cc *jobcontroller) initJobStatus(job *batch.Job) (*batch.Job, error) {
 	job.Status.State.LastTransitionTime = metav1.Now()
 	job.Status.MinAvailable = job.Spec.MinAvailable
 	jobCondition := newCondition(job.Status.State.Phase, &job.Status.State.LastTransitionTime)
-	job.Status.Conditions = append(job.Status.Conditions, jobCondition)
+	job.Status.Conditions = appendJobCondition(job.Status.Conditions, jobCondition)
 	newJob, err := cc.vcClient.BatchV1alpha1().Jobs(job.Namespace).UpdateStatus(context.TODO(), job, metav1.UpdateOptions{})
 	if err != nil {
 		klog.Errorf("Failed to update status of Job %v/%v: %v",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/controllers/job/job_controller_actions.go [hunk 1]
```diff
@@ -1062,6 +1062,26 @@ func newCondition(status batch.JobPhase, lastTransitionTime *metav1.Time) batch.
...
+// slice would exceed maxJobConditions the oldest entries are discarded so that
+// only the most recent maxJobConditions entries are kept.
+func appendJobCondition(conditions []batch.JobCondition, c batch.JobCondition) []batch.JobCondition {
+	if n := len(conditions); n > 0 && conditions[n-1].Status == c.Status {
+		return conditions
+	}
+	conditions = append(conditions, c)
+	if len(conditions) > maxJobConditions {
+		conditions = conditions[len(conditions)-maxJobConditions:]
+	}
```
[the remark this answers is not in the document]
avinxshKD: Good catch. Replaced the sub-slicing with a fresh allocation and copy to prevent pinning the old pointers. Addressed in the latest commit :p

avinxshKD:
@JesseStutler Just a quick note on this one: the core fix is very minimal. The bulk of the  lines added here are just the new unit tests in job_controller_actions_test.go to ensure the truncation and deduplication work exactly as expected.

Happy to make any adjustments to the cap size or logic if needed, thanks

pkg/controllers/job/job_controller_actions.go
```diff
@@ -1062,6 +1062,28 @@ func newCondition(status batch.JobPhase, lastTransitionTime *metav1.Time) batch.
 	}
 }
 
+// maxJobConditions is the upper bound on the number of conditions retained in
+// job.Status.Conditions.  Keeping the slice bounded prevents the etcd object
+// from growing without limit on long-running or frequently-restarted jobs.
+const maxJobConditions = 32
```
hzxuzhonghu: do we still need this limit?

pkg/controllers/job/job_controller_actions.go
```diff
@@ -1062,6 +1062,28 @@ func newCondition(status batch.JobPhase, lastTransitionTime *metav1.Time) batch.
...
+
+// appendJobCondition appends c to conditions, but skips the append when the
+// incoming phase matches the most-recent entry (no-op transition).  If the
+// slice would exceed maxJobConditions the oldest entries are discarded so that
+// only the most recent maxJobConditions entries are kept.
+func appendJobCondition(conditions []batch.JobCondition, c batch.JobCondition) []batch.JobCondition {
+	if n := len(conditions); n > 0 && conditions[n-1].Status == c.Status {
+		return conditions
```
hzxuzhonghu: I think we need to use the latest tansition timestamp

pkg/controllers/job/job_controller_actions.go
avinxshKD: Hey Yes, we def still need the hard limit. The deduplication only catches consecutive identical phases. If a job is stuck in a failing restart loop, it continuously cycles through Pending -> Running -> Restarting -> Pending.... Because the phases are different each time, the deduplication won't catch them, and the slice will still grow unbounded until etcd rejects the object. The limit acts as a hard safety net for those scenarios.

pkg/controllers/job/job_controller_actions.go
avinxshKD: Good catch, I've updated the logic to overwrite the last element (conditions[n-1] = c) so we correctly capture the latest LastTransitionTime while keeping the slice length from growing. Thanks :p

pkg/controllers/job/job_controller_actions.go
hzxuzhonghu: I mean， we could not have unlimited condition set

avinxshKD: Ah I see what you mean, yes exactly. We're on the same page. The hard limit prevents it from ever becoming unlimited.

avinxshKD:
@hzxuzhonghu lgtm, happy to make any further changes if needed thanks.

pkg/controllers/job/job_controller_actions.go
hzxuzhonghu: NOpe, we are not on sasme page. I mean the system implement have already gurangteed it could not exceed 32

avinxshKD: Oh I see what you mean. You're thinking of the MaxRetry limit. It's true that with the default MaxRetry: 3, the job only generates around ~15 conditions before failing permanently, so it wouldn't naturally exceed 32,

However, MaxRetry is user-configurable up to MaxInt32. If a user configures a very high MaxRetry (which is exactly what happened in our cluster for a long-running batch job), the job loops hundreds or thousands of times. Because the phases keep alternating (Pending -> Running -> Restarting -> Pending), the deduplication doesn't catch it, and the conditions array grows into the thousands until the controller completely crashes with etcdserver: request is too large.

Adding this explicit cap of 32 just guarantees the array is bounded regardless of how high the user sets MaxRetry, preventing the etcd crash entirely.

hzxuzhonghu: please remove， it can never run into this case > 32 with eliminating duplicate condition

pkg/controllers/job/job_controller_actions.go
```diff
@@ -1062,6 +1062,29 @@ func newCondition(status batch.JobPhase, lastTransitionTime *metav1.Time) batch.
...
+// job.Status.Conditions.  Keeping the slice bounded prevents the etcd object
+// from growing without limit on long-running or frequently-restarted jobs.
+const maxJobConditions = 32
+
+// appendJobCondition appends c to conditions, but skips the append when the
+// incoming phase matches the most-recent entry (no-op transition).  If the
+// slice would exceed maxJobConditions the oldest entries are discarded so that
+// only the most recent maxJobConditions entries are kept.
+func appendJobCondition(conditions []batch.JobCondition, c batch.JobCondition) []batch.JobCondition {
+	if n := len(conditions); n > 0 && conditions[n-1].Status == c.Status {
```
hzxuzhonghu: THis is not right.  You should use a loop and match delete the original condition and append the new condition at last

pkg/controllers/job/job_controller_actions.go
avinxshKD: Done

pkg/controllers/job/job_controller_actions.go
avinxshKD: I've rewritten the logic exactly as requested, it now loops through to filter out any existing condition with the matching phase, and appends the new one at the end. 

Tests are updated and passingg

avinxshKD:
@hzxuzhonghu can you pls take a look when free, thanks

hzxuzhonghu (approved):

JesseStutler:
/lgtm
Thanks

PR #5436, opened 2026-06-11
[Enhancement] Add grafana_dashboard_enable to conditionally publish Grafana dashboards

### LINKED ISSUE #4987: Publish Grafana dashboards or only deploy Grafana dashboards

Right when set `metrics_enable` to true

https://github.com/volcano-sh/volcano/blob/62ebc877b109bbb73ab08a59027cb9f09372e9f4/installer/helm/chart/volcano/values.yaml#L18

it will deploy a new Grafana based on
https://github.com/volcano-sh/volcano/blob/62ebc877b109bbb73ab08a59027cb9f09372e9f4/installer/helm/chart/volcano/templates/grafana.yaml#L71-L74

### Potential solution 1

Add another variable so that can only publish Grafana dashboards without Grafana instance.

### Potential solution 2

Change `metrics_enable` behavior to only publish Grafana dashboards without Grafana instance.
But this would be a "breaking change".

https://github.com/volcano-sh/volcano/blob/62ebc877b109bbb73ab08a59027cb9f09372e9f4/installer/helm/chart/volcano/values.yaml#L18

### Potential solution 3

It would be great to publish [Volcano dashboards](https://github.com/volcano-sh/volcano/blob/62ebc877b109bbb73ab08a59027cb9f09372e9f4/installer/helm/chart/volcano/templates/grafana.yaml#L60-L69) at

https://grafana.com/grafana/dashboards/?search=volcano

### Additional context

_No response_

### Documentation Updates

shivansh-gohem:
What type of PR is this?
/kind feature

This PR addresses issue #4987.
Currently, when `metrics_enable` is set to `true`, it deploys a new Grafana instance as well as publishing the dashboards. There is no way to deploy just the Grafana dashboards without the Grafana instance.

This PR adds a new variable `grafana_dashboard_enable` to `values.yaml` under `custom`. When set to `true`, it will publish the Grafana dashboard config maps. `metrics_enable` continues to deploy the Grafana instance itself.

Which issue(s) this PR fixes:
Fixes #4987


hajnalmt:
/area deploy
/area metrics

Thanks for working on this. My main concern is that this feels like a narrow design direction for what is really a broader monitoring/metrics chart API problem. After looking at the rendered chart output, I do not think we should solve this only with a Grafana-dashboard-specific flag.

`custom.metrics_enable` is currently a very broad switch: it gates Prometheus, kube-state-metrics, and Grafana resources. This PR adds `custom.grafana_dashboard_enable`, but that only creates an extra escape hatch for the Grafana dashboard ConfigMap. It does not make the metrics stack configurable in a consistent way, and it leaves Prometheus / kube-state-metrics completely out of the picture.

That also makes the new value semantics confusing: `grafana_dashboard_enable=false` does not disable dashboards when `metrics_enable=true`; it only allows rendering the dashboard ConfigMap when `metrics_enable=false`.

Could we step back and define the monitoring values more coherently instead of adding a one-off Grafana dashboard flag? It may make sense to deprecate the broad `custom.metrics_enable` flag and introduce a new external/root monitoring key with granular handling for Prometheus, kube-state-metrics, Grafana, and dashboard ConfigMaps for externally managed Grafana. I would prefer not to merge a Grafana-only flag until that design is clear.

installer/helm/chart/volcano/templates/grafana.yaml
```diff
@@ -46,7 +46,12 @@ data:
       options:
         path: /var/lib/grafana/dashboards 
         foldersFromFilesStructure: true
+{{- end }}
+
+{{- if or .Values.custom.metrics_enable .Values.custom.grafana_dashboard_enable }}
```
hajnalmt: This is an escape for the Config only if I understand your change and it really looks confusing to me. I understand your intent was to solve this directly, but this feels shallow.

metrics_enable=false, grafana_dashboard_enable=false
→ no Grafana resources

metrics_enable=true, grafana_dashboard_enable=false
→ grafana-datasources
→ grafana-<release>-dashboard-config
→ grafana-<release>-dashboard
→ Grafana Service
→ Grafana Deployment

metrics_enable=false, grafana_dashboard_enable=true (this is what you wanted to achieve)
→ only grafana-<release>-dashboard

metrics_enable=true, grafana_dashboard_enable=true
→ same full Grafana path as metrics_enable=true

shivansh-gohem:
Hi @hajnalmt, that makes total sense. I agree that a unified monitoring block is much cleaner than scattering flags under `custom`. 

How does this proposed structure for the root of `values.yaml` look to you?

```yaml
monitoring:
  prometheus:
    enable: false
  kubeStateMetrics:
    enable: false
  grafana:
    enable: false
    dashboards:
      enable: false
```

**Implementation Details:**
1. We will deprecate `custom.metrics_enable` (add comments indicating it's deprecated).
2. For backward compatibility, the templates will fallback to `custom.metrics_enable`. For example, Prometheus will render if `monitoring.prometheus.enable == true` OR `custom.metrics_enable == true`.
3. `grafana-datasources`, `grafana-dashboard-config`, and the Grafana Deployment/Service will be gated by `monitoring.grafana.enable` (or the legacy flag).
4. The dashboard JSONs (`grafana-dashboard` ConfigMap) will be gated by `monitoring.grafana.dashboards.enable` OR `monitoring.grafana.enable` (or the legacy flag), allowing them to be published independently for externally managed Grafana instances.

Does this API design look good to you for us to move forward with?

shivansh-gohem:
@hajnalmt can you please have a look!!

hajnalmt:
Thanks, this is much closer to the direction I had in mind.

One thing I would still like to change is the root key. Instead of introducing `monitoring`, I think the new API should be under a more general `external` / integration-style root.

The chart already has broad root keys like `basic` and `custom`, so `external` would fit that style better for me. It is also more extensible: today it can cover externally managed Grafana dashboards, but later it can be extended for other externally managed integrations without overloading a monitoring-specific key.

Something like this is closer to what I would like to achieve:

```yaml
external:
  prometheus:
    enabled: false
  kubeStateMetrics:
    enabled: false
  grafana:
    enabled: false
    dashboards:
      enabled: false
```

This keeps the API in one general external/integration namespace instead of splitting part of the design under `monitoring` and part of it under external Grafana semantics. `external.grafana.dashboards.enabled` can mean the chart only publishes the Volcano dashboard ConfigMaps for a Grafana instance managed outside this chart, while still leaving room to add other external integration settings later.

I am not strict on the exact key name (`external`, `integrations`, `externalServices`, etc.), but I would prefer this to be a general extensible root rather than `monitoring`. We should also keep `custom.metrics_enable` as a backward-compatible legacy path, with comments explaining that the new granular keys are preferred.

shivansh-gohem:
@hajnalmt sorry for being late!!

hajnalmt (commented):
Thanks for the changes! I am fine with this overall.
Can you add a note too about the deprecation?
There is a neat trick to warn users about the deprecation, you can use haskey
```
{{- if hasKey .Values "metrics_enable" }}

WARNING: `metrics_enable` is deprecated.

Please migrate:

    ....

{{- end }}
```
Also can you update the PR description and the title please?

PR #5441, opened 2026-06-13
scheduler/api: suppress MaxFloat64 sentinel values in Resource.String()

### LINKED ISSUE #5148: Noisy log outputs on realCapability and capability due to maxFloat and maxInt

A69SHUBHAM:
## What this PR does

When a queue has no explicit capability configured, Volcano uses `math.MaxFloat64` internally as a sentinel value representing unlimited resources.

Currently, `Resource.String()` formats these values using `%0.2f`, which results in extremely large numeric values appearing in scheduler logs. This makes log output difficult to read and obscures the actual meaning of the value.

This PR updates `Resource.String()` to render `math.MaxFloat64` as `<unlimited>` while preserving the existing formatting for normal resource values.

## Changes

* Add a helper to format resource values for display.
* Render `math.MaxFloat64` as `<unlimited>`.
* Update `Resource.String()` to use the helper for CPU, memory, and scalar resources.
* Add regression tests covering:

  * Normal resource values
  * Unlimited CPU and memory values
  * Mixed finite/unlimited values
  * Unlimited scalar resources

## Example

Before:

cpu 179769313486231570000000000000000000000000000000000000...

After:

cpu <unlimited>

## Testing

```bash
go test ./pkg/scheduler/api/... -run TestResourceString -v
go test ./pkg/scheduler/api/... -count=1
```

Fixes #5148

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -44,6 +44,10 @@ const (
 
 const (
 	minResource float64 = 0.1
+
+	// unlimitedResourceStr is printed by Resource.String() when a field holds
+	// math.MaxFloat64, which is used as a sentinel meaning "no limit".
+	unlimitedResourceStr = "<unlimited>"
 )
 
 // DimensionDefaultValue means default value for black resource dimension
@@ -168,7 +172,7 @@ func (r *Resource) Clone() *Resource {
 
 // String returns resource details in string format
 func (r *Resource) String() string {
-	str := fmt.Sprintf("cpu %0.2f, memory %0.2f", r.MilliCPU, r.Memory)
+	str := fmt.Sprintf("cpu %s, memory %s", formatResourceValue(r.MilliCPU), formatResourceValue(r.Memory))
 	// Sort scalar resource names to ensure consistent string output
 	var resourceNames []string
 	for rName := range r.ScalarResources {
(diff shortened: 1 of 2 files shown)


Aman-Cool:
Hey @A69SHUBHAM, Thanks for the PR :) 

This handles cpu and memory, but I don't think it fully closes #5148 yet.

There are actually two different sentinels in that log line. cpu/memory are `math.MaxFloat64` (the `1.79e308` values), which this catches. But the scalar resources (gpu, pods, ephemeral-storage, hugepages) print `9223372036854775808.00`, which is `float64(math.MaxInt64)` ; set on the root queue in `capacity.go`:

```go
// capacity.go ~L1337
infinityResource.SetScalar(k, math.MaxInt64)
```

Since `formatResourceValue` only compares against `MaxFloat64`, those scalars still come out as the long number, and they're most of the noisy line in the issue. End result is something like:

```
cpu <unlimited>, memory <unlimited>, nvidia.com/gpu 9223372036854775808.00, pods 9223372036854775808.00
```

Could you also check for `float64(math.MaxInt64)` in the helper?

```go
if v == math.MaxFloat64 || v == float64(math.MaxInt64) {
    return unlimitedResourceStr
}
```

Also the scalar test case uses `MaxFloat64`, but in practice scalars are never set to that; so it passes without actually covering the case from the issue. Worth switching it to `math.MaxInt64`.

One thing for later (not this PR): having two different "unlimited" sentinels is a bit of a trap. A small `IsUnlimited(v)` helper next to where they're defined would save the next person from hitting this again.

hajnalmt (commented):
Thanks for the PR!

From the 2 PRs I prefer: https://github.com/volcano-sh/volcano/pull/5276 this already solves what you suggest @Aman-Cool  Can you take a look on this too? A second opinion would be really helpful.

A69SHUBHAM:
Pushed cf04936 to address the scalar resource gap @Aman-Cool flagged — formatResourceValue (via a new IsUnlimited helper) now also catches float64(math.MaxInt64), not just math.MaxFloat64, so nvidia.com/gpu, pods, etc. correctly print as <unlimited> instead of the raw 9223372036854775808.00 value. Test coverage updated to actually exercise that case, plus a mixed-sentinel test.
@hajnalmt happy to take a look at #5276 too if a comparison would help.

Aman-Cool:
@hajnalmt, Took a look.., agreed, #5276 is the better approach, it already handles both sentinels. I left my detailed thoughts over there. 

Thanks @A69SHUBHAM for the work here and for turning the MaxInt64 feedback around so quickly :) since #5276 covers the same ground, it probably makes sense to consolidate on that one.

hajnalmt (commented):
/close 

Thank you for the PR @A69SHUBHAM and for your hard work on this!

I see you addressed the review points, I still decided to go with https://github.com/volcano-sh/volcano/pull/5276 in the end. That change was more close to my taste.

Thank you once more!

PR #5442, opened 2026-06-14
fix: move queue inqueue accounting from JobEnqueueable predicate to J…

WHOIM1205:
## Summary

This PR fixes incorrect `inqueue` resource accounting in the `proportion` and `capacity` plugins by moving accounting updates from `JobEnqueueableFn` to `JobEnqueuedFn`.

Previously, both plugins updated `inqueue` resources during the enqueue eligibility check. If a later plugin rejected the job, the enqueue operation was aborted but the resource reservation remained, resulting in stale `inqueue` accounting and potential false queue-capacity rejections.

This change aligns both plugins with the pattern already used by the `overcommit` plugin, where resource accounting is only committed after the job is successfully enqueued.

## Root Cause

`JobEnqueueableFn` is a predicate phase and can be overruled by a later plugin returning `Reject`.

However, `proportion` and `capacity` were performing `inqueue` accounting as a side effect during this phase, causing phantom reservations when:
1. A plugin returned `Permit`
2. `inqueue` resources were updated
3. A later plugin returned `Reject`
4. The job was never enqueued

As a result, queue accounting became inflated for the remainder of the scheduling session.

### Proportion Plugin
- Removed `attr.inqueue.Add(...)` from `JobEnqueueableFn`
- Added a `JobEnqueuedFn` hook to update `inqueue` resources only after successful enqueue

### Capacity Plugin
- Removed queue and ancestor `inqueue` updates from `JobEnqueueableFn`
- Added a `JobEnqueuedFn` hook to perform leaf queue, ancestor queue, and DRA `inqueue` accounting after successful enqueue

## Impact

- Prevents phantom `inqueue` reservations
- Eliminates false `queue resource quota insufficient` rejections caused by stale accounting
- Makes enqueue accounting independent of plugin ordering
- Preserves existing behavior for successfully enqueued jobs

## Testing

- `go build ./pkg/scheduler/plugins/proportion/... ./pkg/scheduler/plugins/capacity/...`
- `go test ./pkg/scheduler/plugins/proportion/...`
- `go test ./pkg/scheduler/plugins/capacity/...`
- `go test ./pkg/scheduler/actions/enqueue/...`
- `go test ./pkg/scheduler/framework/...`

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -775,6 +775,26 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 			return util.Reject
 		}
 
+		return util.Permit
+	})
+
+	// JobEnqueuedFn is the commit hook: it runs only after the aggregate JobEnqueueable
+	// verdict is Permit (a later plugin's Reject short-circuits JobEnqueueable before this
+	// is called). Updating inqueue here, rather than inside the JobEnqueueableFn predicate,
+	// avoids leaving a phantom inqueue reservation on the queue (and every ancestor queue)
+	// when another enqueue plugin rejects the job after capacity has already voted Permit.
+	ssn.AddJobEnqueuedFn(cp.Name(), func(obj interface{}) {
+		job := obj.(*api.JobInfo)
+		attr := cp.queueOpts[job.Queue]
+		if attr == nil || attr.realCapability == nil {
+			return
+		}
+		// Mirror the JobEnqueueableFn early return: nothing is reserved when the job has
+		// neither MinResources nor DRA min resources.
+		if job.PodGroup.Spec.MinResources == nil && !(cp.dynamicResourceAllocationEnable && attr.dra != nil && job.GetMinDRAResources() != nil) {
+			return
+		}
+
 		// job enqueued
 		deductedResources := job.DeductSchGatedResources(job.GetMinResources())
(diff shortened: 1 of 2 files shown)


kunalworldwide (commented):
The bug is subtle and the fix is well-motivated: moving `attr.inqueue.Add()` from `JobEnqueueableFn` to `JobEnqueuedFn` ensures the inqueue reservation is only committed when the job actually gets enqueued. 

Before this fix, if the capacity plugin voted `Permit` and added the inqueue reservation but a later plugin (like proportion) voted `Reject`, the reservation leaked — the queue's inqueue count was inflated by a job that never actually entered the queue, blocking other jobs from being admitted.

The `JobEnqueuedFn` hook runs only after all `JobEnqueueableFn` predicates pass, so it's the right place for the side effect. Both the `capacity` and `proportion` plugins are fixed consistently.

The comment explaining the design is clear and matches the code structure. LGTM.

JesseStutler:
https://github.com/volcano-sh/volcano/pull/5339 already followed

WHOIM1205:
Thanks for the review! @kunalworldwide The requested changes have been addressed and all checks are passing. Please let me know if there's anything else I should update.

hajnalmt (commented):
/close

Thanks for the PR!
I am closing this because https://github.com/volcano-sh/volcano/pull/5339 addresses the issue in every plugin properly not just proportion and capacity and the code there is much more mature (contains UTs etc...).

PR #5457, opened 2026-06-16
fix: add JSON tags for scheduler configuration

3th4novo:
#### What type of PR is this?

/kind bug

This PR adds JSON tags for scheduler configuration structs so scheduler configuration fields can be correctly serialized/deserialized with JSON field names.

It updates scheduler configuration options, tiers, plugin options, action configurations, and metrics configuration to include JSON mappings.

Fixes #

Please pay attention to scheduler configuration compatibility, especially the fields whose types were changed to support more flexible configuration input.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/conf/scheduler_conf.go
+++ b/pkg/scheduler/conf/scheduler_conf.go
@@ -27,81 +27,81 @@ var EnabledActionMap map[string]bool
 // SchedulerConfiguration defines the configuration of scheduler.
 type SchedulerConfiguration struct {
 	// Actions defines the actions list of scheduler in order
-	Actions string `yaml:"actions"`
+	Actions interface{} `yaml:"actions" json:"actions,omitempty"`
 	// Tiers defines plugins in different tiers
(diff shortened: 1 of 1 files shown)


JesseStutler:
The image building looks like failed

pkg/scheduler/conf/scheduler_conf.go
JesseStutler, line 1: But you didn't call `json.Unmarshal` to unmarshal it to the SchedulerConfiguration, so what's the point of adding JSON tags here?

3th4novo: The main purpose here is to let upstream services to modify and parse volcano configmaps. In those scenarios, we might conduct unmarshal actions.

JesseStutler: I still didn't get it, you mean the upstream service will convert json format configuration to yaml configuration for the scheduler?

PR #5462, opened 2026-06-17
support hami-vnpu-core for Ascend device

DSFans2014:

[HAMi-vnpu-core](https://github.com/Project-HAMi/hami-vnpu-core) is an Ascend NPU in-container resource controller written in Rust. It implements user-space interception via libvnpu.so (interceptor) and Limiter (manager). Two environment variables are used to declare resource quotas: `NPU_MEM_QUOTA` for memory limits and `NPU_PRIORITY` for scheduling priority. This design integrates that capability into HAMi scheduling to support Ascend NPU memory virtualization and compute time-slice soft partitioning. This feature has already been supported in HAMi v2.9.0. For more detailed information, please refer [here](https://project-hami.io/blog/hami-v2-9-0-release).

This PR enables this soft-partitioning capability support for Ascend devices in the Volcano scheduler. The main changes are as follows:
- add `PostStart` injection in the webhook
- add support for `hami-vnpu-core` in Ascend's device_info

Fixes #

```release-note

```

### DIFF
--- a/cmd/webhook-manager/app/server.go
+++ b/cmd/webhook-manager/app/server.go
@@ -34,6 +34,7 @@ import (
 	informers "volcano.sh/apis/pkg/client/informers/externalversions"
 	"volcano.sh/volcano/cmd/webhook-manager/app/options"
 	"volcano.sh/volcano/pkg/kube"
+	devconfig "volcano.sh/volcano/pkg/scheduler/api/devices/config"
 	"volcano.sh/volcano/pkg/signals"
 	commonutil "volcano.sh/volcano/pkg/util"
 	wkconfig "volcano.sh/volcano/pkg/webhooks/config"
@@ -68,6 +69,9 @@ func Run(config *options.Config) error {
 		klog.Errorf("loadAdmissionConf failed.")
 	} else {
 		klog.V(2).Infof("loadAdmissionConf:%v", admissionConf)
+		if admissionConf.DeviceMutatorConf.Enable {
+			devconfig.InitDevicesConfig(admissionConf.DeviceMutatorConf.KnownGeometriesCMName, admissionConf.DeviceMutatorConf.KnownGeometriesCMNamespace)
+		}
 	}
 
 	vClient := getVolcanoClient(restConfig)
--- a/cmd/webhook-manager/app/server.go
+++ b/cmd/webhook-manager/app/server.go
@@ -67,7 +67,7 @@ func Run(config *options.Config) error {
 	if admissionConf == nil {
 		klog.Errorf("loadAdmissionConf failed.")
 	} else {
-		klog.V(2).Infof("loadAdmissionConf:%v", admissionConf.ResGroupsConfig)
+		klog.V(2).Infof("loadAdmissionConf:%v", admissionConf)
 	}
 
 	vClient := getVolcanoClient(restConfig)
--- a/docs/user-guide/how_to_use_vnpu.md
+++ b/docs/user-guide/how_to_use_vnpu.md
@@ -25,7 +25,7 @@ with support for more chip types to come
 
 **Description**:
 
-This mode is developed by a third-party community 'HAMi', which is the developer of [volcano-vgpu](./how_to_use_volcano_vgpu.md) feature. It supports vNPU feature for both Ascend 310 and Ascend 910. It also supports managing heterogeneous Ascend cluster(Cluster with multiple Ascend types, i.e. 910A,910B2,910B3,310P)
+This mode is developed by a third-party community 'HAMi', which is the developer of [volcano-vgpu](./how_to_use_volcano_vgpu.md) feature. It supports vNPU feature for both Ascend 310 and Ascend 910. It also supports managing heterogeneous Ascend cluster(Cluster with multiple Ascend types, i.e. 910A,910B2,910B3,310P). Starting with version 1.16, we support soft device partitioning through [hami-vnpu-core](https://github.com/Project-HAMi/hami-vnpu-core). This feature is known as the `hami-core` mode.
 
 **Use case**:
 
@@ -43,7 +43,7 @@ To enable vNPU scheduling, the following components must be set up based on the 
 **Prerequisites**:
 
 Kubernetes >= 1.16  
-Volcano >= 1.14  
+Volcano >= 1.14 (1.16 for `hami-core` mode)  
 [ascend-docker-runtime](https://gitcode.com/Ascend/mind-cluster/tree/master/component/ascend-docker-runtime) (for HAMi Mode)
 
 ### Install Volcano:
@@ -115,19 +115,21 @@ data:
 
 #### HAMi mode
 
-##### Label the Node with `ascend=on`
+##### Template vNPU mode
+
+###### Label the Node with `ascend=on`
 
 ```
 kubectl label node {ascend-node} ascend=on
 ```
 
-##### Deploy `hami-scheduler-device` ConfigMap
+###### Deploy `hami-scheduler-device` ConfigMap
 
 ```
 kubectl apply -f https://raw.githubusercontent.com/Project-HAMi/ascend-device-plugin/refs/heads/main/ascend-device-configmap.yaml
 ```
 
-##### Deploy ascend-device-plugin
+###### Deploy ascend-device-plugin
 
 ```
 kubectl apply -f https://raw.githubusercontent.com/Project-HAMi/ascend-device-plugin/refs/heads/main/ascend-device-plugin.yaml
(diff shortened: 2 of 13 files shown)


JesseStutler:
/cc
Cool, will take a look in recent days

hajnalmt (commented):
/area scheduling
/area webhooks
/area deploy
/lgtm

Thanks for the PR, this looks good to me. The admission path and scheduler-side Ascend allocation changes line up with the HAMi v2.9.0 hami-vnpu-core flow, and the focused tests cover the key cases.
I don't have any additions, leaving place for others to review eg /cc artlichi can you check this too?

pkg/scheduler/api/devices/ascend/hami/device_info.go
```diff
@@ -49,6 +49,10 @@ const (
 	spreadPolicy      = "spread"
 	binpackMultiplier = 100
 	spreadMultiplier  = 100
+
+	VNPUModeAnnotation         = "huawei.com/vnpu-mode"
+	VNPUModeHamiCore           = "hami-core"
+	VNPUNodeSelectorAnnotation = "hami-vnpu-core"
```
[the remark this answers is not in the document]
hajnalmt: This makes sense for me to outsource this, but I wouldn't block on it. The duplication in the test which slightly bothers me.

hajnalmt (commented):
@DSFans2014
I wanted to ask if this works on any Ascend NPUs right? I have really little domain knowledge here, but from the code itself what I understand is that this works on any NPU thats compatible with the Ascend [CANN Runtime / RTS API](https://github.com/Project-HAMi/hami-vnpu-core/blob/ddd0ae003800c884c7a6f718dfad5e9d0dc3393e/crates/limiter/src/externed_api.rs#L1-L12), but the prerequisite in the README says [Ascend 910B](https://github.com/Project-HAMi/hami-vnpu-core#prerequisites).

You have used Ascend910A and Ascend310P in the tests so this is the case probably, just wanted to ask. Also please consider updating the User guide and the website regarding this (fine in a separate PR), and maybe outsourcing the hami annotations that are duplicated now to some common place like `pkg/scheduler/api/devices/ascend/hami/constants.go`.

JesseStutler:
@DSFans2014 Yes I think add a user guide will be better, and list supported Ascend devices will be better

DSFans2014:
> @DSFans2014 I wanted to ask if this works on any Ascend NPUs right? I have really little domain knowledge here, but from the code itself what I understand is that this works on any NPU thats compatible with the Ascend [CANN Runtime / RTS API](https://github.com/Project-HAMi/hami-vnpu-core/blob/ddd0ae003800c884c7a6f718dfad5e9d0dc3393e/crates/limiter/src/externed_api.rs#L1-L12), but the prerequisite in the README says [Ascend 910B](https://github.com/Project-HAMi/hami-vnpu-core#prerequisites).
> 
> You have used Ascend910A and Ascend310P in the tests so this is the case probably, just wanted to ask. Also please consider updating the User guide and the website regarding this (fine in a separate PR), and maybe outsourcing the hami annotations that are duplicated now to some common place like `pkg/scheduler/api/devices/ascend/hami/constants.go`.

- I've tested `hami-vnpu-core` on both 310P3 and 910B3, and everything worked fine. I understand that `hami-vnpu-core` is designed to be compatible with all Ascend NPUs. Could you explain why the [Prerequisites](https://github.com/Project-HAMi/hami-vnpu-core/blob/ddd0ae003800c884c7a6f718dfad5e9d0dc3393e/README.md?plain=1#L18) section exclusively lists the 910B model? @archlitchi  
- `constants.go` and docs have been added @hajnalmt @JesseStutler
- Additionally, the [vnpu-related documentation](https://github.com/volcano-sh/website/pull/430) hasn't been merged into the website yet.  Could you please review and confirm whether this should be merged? If there are any further revisions required, please let me know. @JesseStutler

docs/user-guide/how_to_use_vnpu.md
```diff
@@ -25,7 +25,7 @@ with support for more chip types to come
 
 **Description**:
 
-This mode is developed by a third-party community 'HAMi', which is the developer of [volcano-vgpu](./how_to_use_volcano_vgpu.md) feature. It supports vNPU feature for both Ascend 310 and Ascend 910. It also supports managing heterogeneous Ascend cluster(Cluster with multiple Ascend types, i.e. 910A,910B2,910B3,310P)
+This mode is developed by a third-party community 'HAMi', which is the developer of [volcano-vgpu](./how_to_use_volcano_vgpu.md) feature. It supports vNPU feature for both Ascend 310 and Ascend 910. It also supports managing heterogeneous Ascend cluster(Cluster with multiple Ascend types, i.e. 910A,910B2,910B3,310P). Starting with version 1.16, we support soft device partitioning through [hami-vnpu-core](https://github.com/Project-HAMi/hami-vnpu-core). This feature is known as the `hami-core` mode.
```
JesseStutler, line 28: Sorry for the confusing @DSFans2014 , doesn't hami's ascend-device-plugin already support vNPU feature and integrate with volcano in v1.14? What's new in hami vnpu core? Why we need to modify the scheduler logic to support the hami-vnpu-core? Is the code previously introduced in the scheduler insufficient for coverage?

DSFans2014: Although `hami-vnpu-core` is already supported in `ascend-device-plugin` v1.3.0, the scheduler requires additional work to enable the soft partitioning feature:
- Pod Injection: Inject a postStart hook to run the `limiter`, which applies the vNPU resource limits (cores/memory) to the container.
- Device Discovery: The scheduler need to discover which nodes have `hami-vnpu-core` enabled, as it can be configured on a per-node basis.
- Resource Allocation: The scheduler needs to support the huawei.com/xxx-core resource type which is used to limit npu utilization through time-slicing.

DSFans2014: > What's new in hami vnpu core? Why we need to modify the scheduler logic to support the hami-vnpu-core?

The previous vnpu feature only supported fixed partitioning based on templates, similar to NVIDIA GPU MIG. By comparison, `hami-vnpu-core` introduces support for flexible partitioning, allowing arbitrary resource allocation without being restricted to fixed template sizes.

JesseStutler: Cool, I got it, this release note is generated by AI:
```
Ascend NPU now supports **soft-partitioning** (compute time-slicing and arbitrary memory quotas)
via the new `hami-vnpu-core` mode, in addition to the existing template-based hard partitioning.

Pods can opt into soft slicing by setting the annotation `huawei.com/vnpu-mode: hami-core`,
and can additionally request the new `huawei.com/<chip>-core` resource to declare a
percentage-based compute share. The webhook injects a `postStart` hook to launch the
`hami-vnpu-core` limiter inside the container, and the scheduler discovers nodes that have
`hami-vnpu-core` enabled (either via the global `vnpus.hamiVnpuCore` switch or per-node
configuration). Pods without the annotation continue to use the existing
`ASCEND_VNPU_SPECS` template-based path with no behavior change.

Requires Ascend driver >= 25.5 and the chip's `device-share` mode enabled
(currently ARM-only). See https://project-hami.io/blog/hami-v2-9-0-release for details.
```
@DSFans2014  Any suggestion or opinion? I'll use this explanation to explain what's new for vnpu scheduling in v1.16

JesseStutler:
/lgtm
Let @archlitchi to approve it

docs/user-guide/how_to_use_vnpu.md
```diff
@@ -299,3 +432,38 @@ The supported Ascend chips and their `ResourceNames` are shown in the following
...
+|-------|-------|-------|
+| 910A | huawei.com/Ascend910A | huawei.com/Ascend910A-memory | huawei.com/Ascend910A-core |
+| 910B2 | huawei.com/Ascend910B2 | huawei.com/Ascend910B2-memory | huawei.com/Ascend910B2-core |
+| 910B3 | huawei.com/Ascend910B3 | huawei.com/Ascend910B3-memory | huawei.com/Ascend910B3-core |
+| 910B4 | huawei.com/Ascend910B4 | huawei.com/Ascend910B4-memory | huawei.com/Ascend910B4-core |
+| 910B4-1 | huawei.com/Ascend910B4-1 | huawei.com/Ascend910B4-1-memory | huawei.com/Ascend910B4-1-core |
+| 910C | huawei.com/Ascend910C | huawei.com/Ascend910C-memory | huawei.com/Ascend910C-core |
+| 310P3 | huawei.com/Ascend310P | huawei.com/Ascend310P-memory | huawei.com/Ascend310P-core |
+
+**Note**: If the pod's annotations do not specify `hami-core`, the device will be allocated in the normal hami mode even if the `hami-core` feature is enabled in the configuration file.
```
archlitchi: in the normal hami mode -> in the template vNPU mode

docs/user-guide/how_to_use_vnpu.md
```diff
@@ -158,6 +219,77 @@ data:
 
   **Note:** You may notice that, 'volcano-vgpu' has its own GeometriesCMName and GeometriesCMNamespace, which means if you want to use both vNPU and vGPU in a same volcano cluster, you need to merge the configMap from both sides and set it here.
 
+##### Update Admission Config
```
archlitchi: We need to automate this step, either move these part to helm chart, or put them in ascend-device-plugin. it's not a very good practice to let user execute such complicated operations during setup

docs/user-guide/how_to_use_vnpu.md
DSFans2014: It looks good to me

cmd/webhook-manager/app/server.go
```diff
@@ -67,7 +68,10 @@ func Run(config *options.Config) error {
 	if admissionConf == nil {
 		klog.Errorf("loadAdmissionConf failed.")
 	} else {
-		klog.V(2).Infof("loadAdmissionConf:%v", admissionConf.ResGroupsConfig)
+		klog.V(2).Infof("loadAdmissionConf:%v", admissionConf)
+		if admissionConf.DeviceMutatorConf.Enable {
+			devconfig.InitDevicesConfig(admissionConf.DeviceMutatorConf.KnownGeometriesCMName, admissionConf.DeviceMutatorConf.KnownGeometriesCMNamespace)
+		}
```
archlitchi, line 74: can we move the 'patch webhook' procedure to here? like adding the following:
``` 
if !strings.Contains(config.EnabledAdmission, "/pods/mutate") {
        if config.EnabledAdmission != "" {
            config.EnabledAdmission += ",/pods/mutate"
        } else {
            config.EnabledAdmission = "/pods/mutate"
        }
    }
```
instead of update these fields:
```
    --set custom.enabled_admissions='/jobs/mutate\,/jobs/validate\,/podgroups/validate\,/queues/mutate\,/queues/validate\,/hypernodes/validate\,/cronjobs/validate\,/pods/mutate'
```

docs/user-guide/how_to_use_vnpu.md
```diff
@@ -135,7 +135,68 @@ kubectl apply -f https://raw.githubusercontent.com/Project-HAMi/ascend-device-pl
```
archlitchi, line 137: We need to notify users to Install Volcano by using helm and set parameter 'custom.device_mutator_enable=true' here

docs/user-guide/how_to_use_vnpu.md
```diff
@@ -158,6 +219,13 @@ data:
 
   **Note:** You may notice that, 'volcano-vgpu' has its own GeometriesCMName and GeometriesCMNamespace, which means if you want to use both vNPU and vGPU in a same volcano cluster, you need to merge the configMap from both sides and set it here.
 
+##### Upgrade volcano-adminssion
```
archlitchi: Since we notify user to install volcano with device_mutator_enable=true and move the /pod/mutate to server.go, this field is no longer needed

cmd/webhook-manager/app/server.go
DSFans2014: Helm checks if `/pods/mutate` is in `enabled_admissions` to decide whether to deploy the corresponding `MutatingWebhookConfiguration`.
https://github.com/volcano-sh/volcano/blob/edde60ced3618de338863c59c5f2cbf48d01b754/installer/helm/chart/volcano/templates/webhooks.yaml#L3-L13

docs/user-guide/how_to_use_vnpu.md
DSFans2014: I've split the deployments of `hami-core` and `hami`, since the `hami-core` mode not only needs to modify variables in the `ascend-device-configmap`, but also requires the deployment of the `ascend-device-node-configmap`.  Do you mean we should put them together in one section and then cover each one separately?

archlitchi: ok, but still, you need to notify user to install volcano by using helm in your 'hami-core' section

DSFans2014: In hami mode, I've split it into two modes: Template vNPU mode and `hami-core` mode.
At the beginning of the `Installation` section, we cover how to install Volcano, and the later parts describe the `Scheduler Config` that needs to be configured.
https://github.com/volcano-sh/volcano/blob/6b698d455496782714cbd88aec06c0d75e0c37d1/docs/user-guide/how_to_use_vnpu.md?plain=1#L49-L53

archlitchi: My point is, volcano can be deployed either by using 'yaml'(https://github.com/volcano-sh/volcano?tab=readme-ov-file#install-with-yaml-files) or by using 'helm'. If they choose install volcano by using 'YAML', you won't be able to upgrade volcano-admission according to this document, since there's no chart for you to upgrade

DSFans2014: I got, if the user has already deployed via a Deployment YAML, we should ask them to either use Helm to generate a YAML and deploy it, or uninstall the current deployment first and then install with Helm.

archlitchi (approved):
/lgtm

PR #5464, opened 2026-06-17
docs(owners): add hajnalmt as approver

hajnalmt:
Adds `hajnalmt` as an approver in the OWNERS files. The approver membership request: https://github.com/volcano-sh/community/issues/150


JesseStutler:
/approve
/lgtm
Welcome aboard, mate! I think mate can absolutely handle all the current modules' reviews now

PR #5470, opened 2026-06-18
avoid idle gpu memory underflow in getDevicesIdleGPUMemory

netliomax25-code:
#### What type of PR is this?

/kind bug

1. getDevicesIdleGPUMemory reports a card's idle memory as allMemory - usedMemory, and both values are uint.
2. usedMemory is the sum of the pods accounted to that card by getUsedGPUMemory, and AddResource records a pod against whatever card its volcano.sh/gpu-index annotation names without checking capacity, so a pod bound directly to a node (or several pods pinned to the same card) can push the accounted usage past the card's memory.
3. Once usedMemory exceeds allMemory the unsigned subtraction wraps to a near-max value, so predicateGPUbyMemory treats a full card as almost empty and keeps offering it to other pods, placing more memory on it than the card has.

Clamped the idle value to zero once usage reaches capacity so the result can no longer wrap. Added a regression test that drives the accounted usage above the card capacity and checks the card reports zero idle memory and is not offered by the predicate.

N/A

The clamp lives in the accounting helper, so every caller of getDevicesIdleGPUMemory is covered rather than the current predicate path alone.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/api/devices/nvidia/gpushare/share.go
+++ b/pkg/scheduler/api/devices/nvidia/gpushare/share.go
@@ -34,7 +34,15 @@ func getDevicesIdleGPUMemory(gs *GPUDevices) map[int]uint {
 	res := map[int]uint{}
 	for id, allMemory := range devicesAllGPUMemory {
 		if usedMemory, found := devicesUsedGPUMemory[id]; found {
-			res[id] = allMemory - usedMemory
+			// usedMemory is summed from the pods accounted to this card, which can
+			// exceed the card capacity (for example a pod that sets volcano.sh/gpu-index
+			// itself). Clamp instead of subtracting so the unsigned result does not wrap
+			// around and make a full card look almost empty.
+			if usedMemory >= allMemory {
+				res[id] = 0
+			} else {
+				res[id] = allMemory - usedMemory
+			}
 		} else {
 			res[id] = allMemory
 		}
--- a/pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
+++ b/pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2023 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gpushare
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func gpuMemPod(uid string, mem int64) *v1.Pod {
+	return &v1.Pod{
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							VolcanoGPUResource: *resource.NewQuantity(mem, resource.DecimalSI),
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func TestGetDevicesIdleGPUMemoryOverSubscribed(t *testing.T) {
+	// A card with 1000 memory whose accounted usage exceeds its capacity, e.g.
+	// because a pod set volcano.sh/gpu-index itself and was bound directly to the
+	// node, must report 0 idle memory rather than a wrapped-around uint that makes
+	// the full card look almost empty.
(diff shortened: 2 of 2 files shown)


pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
```diff
@@ -0,0 +1,71 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func TestGetDevicesIdleGPUMemoryOverSubscribed(t *testing.T) {
+	// A card with 1000 memory whose accounted usage exceeds its capacity, e.g.
```
hajnalmt: You can remove these overexplaining comments too.

pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright 2023 The Volcano Authors.
```
hajnalmt: 2026 😄

pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
```diff
@@ -0,0 +1,71 @@
...
+							VolcanoGPUResource: *resource.NewQuantity(mem, resource.DecimalSI),
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func TestGetDevicesIdleGPUMemoryOverSubscribed(t *testing.T) {
```
hajnalmt: I don't like the naming of this test, because it asserts on 2 different things and second here is not inferred:

1. Oversubsribed cards reports zero.
2. Predicate doesn't offer oversubscride card.

I would rename it at least to `TestGPUShare_OverSubscribedCardDoesNotUnderflow` but the cleanest would be two distict tests:

`TestGetDevicesIdleGPUMemory_OverSubscribedCardReportsZero`
`TestPredicateGPUbyMemory_OverSubscribedCardNotOffered`

Following the repo conventions.

pkg/scheduler/api/devices/nvidia/gpushare/share.go
```diff
@@ -34,7 +34,15 @@ func getDevicesIdleGPUMemory(gs *GPUDevices) map[int]uint {
 	res := map[int]uint{}
 	for id, allMemory := range devicesAllGPUMemory {
 		if usedMemory, found := devicesUsedGPUMemory[id]; found {
-			res[id] = allMemory - usedMemory
+			// usedMemory is summed from the pods accounted to this card, which can
```
hajnalmt, line 37: This comment helps understanding I am fine with it.

hajnalmt (changes requested):
/area scheduling

Thanks for the PR! Clamping here is the right path I think too, nice catch.

The bots are right and I had minor comments too, can you fix these please?

pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
netliomax25-code: Split it into the two tests you suggested: TestGetDevicesIdleGPUMemory_OverSubscribedCardReportsZero and TestPredicateGPUbyMemory_OverSubscribedCardNotOffered, with a shared helper for the oversubscribed setup.

pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
netliomax25-code: Done, dropped the comment block.

pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
netliomax25-code: Ha, fixed to 2026.

pkg/scheduler/api/devices/nvidia/gpushare/share_test.go
```diff
@@ -0,0 +1,71 @@
...
+					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							VolcanoGPUResource: *resource.NewQuantity(mem, resource.DecimalSI),
+						},
+					},
+				},
+			},
+		},
+	}
+}
```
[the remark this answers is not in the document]
netliomax25-code: Dropped the unused uid param and updated the call sites to pass only the memory.

netliomax25-code:
Pushed the review fixes:

1. Dropped the unused `uid` param from `gpuMemPod` and updated both call sites.
2. Bumped the copyright year to 2026.
3. Removed the overexplaining comment block.
4. Split the regression into the two tests you suggested, `TestGetDevicesIdleGPUMemory_OverSubscribedCardReportsZero` and `TestPredicateGPUbyMemory_OverSubscribedCardNotOffered`, sharing a small setup helper.

Left the share.go comment as-is since you were fine with it. go test, gofmt and go vet are clean.

hajnalmt (commented):
/lgtm

Thanks for the changes! Giving other reviewers the chance to take a look on this for approve.

JesseStutler:
/approve
I have no doubt about the codes, but I'm wondering that do you really meet such bugs or you just scan the code base and think here should be enhanced? Not so familiar with the part of vGPU codes but why scheduler would allow previous codes usedMemory to exceed to allocatable memory?

Issue #5468, opened 2026-06-18
Helm v4 install gets stuck in pending-install after pre-install hook completes


JesseStutler:
## Description

Installing Volcano 1.15.0 with Helm v4.2.1 gets stuck in `pending-install`.

The pre-install hook Job `volcano-admission-init` appears to complete successfully, and the image is pulled quickly, but the Helm release remains in `pending-install`. The main Volcano resources such as CRDs, ClusterRoles, Deployments, and Services are not created.

This does not look like an image pull or registry network issue. The hook Pod pulled `docker.io/volcanosh/vc-webhook-manager:v1.15.0` successfully in about 3.3 seconds.

The same install works after switching to Helm v3.21.1, which suggests a possible compatibility issue with Helm v4 hook/install behavior.

## Steps to reproduce the issue

1. Use Helm v4.2.1.
2. Install Volcano 1.15.0 chart:

   ```bash
   helm install volcano <volcano-chart> \
     -n volcano-system \
     --create-namespace
   ```

3. Check the Helm release:

   ```bash
   helm status volcano -n volcano-system
   helm history volcano -n volcano-system
   ```

4. Check hook events:

   ```bash
   kubectl get events -n volcano-system --sort-by=.lastTimestamp
   kubectl get job,pod,sa,role,rolebinding,secret -n volcano-system
   ```

5. Inspect the Helm release secret:

   ```bash
   kubectl get secret sh.helm.release.v1.volcano.v1 -n volcano-system \
     -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d | jq '.info, .hooks[]? | {name: .name, kind: .kind, last_run: .last_run}'
   ```

## Describe the results you received and expected

Actual result:

- Helm release remains in `pending-install`.
- `helm status` shows:

  ```text
  STATUS: pending-install
  DESCRIPTION: Initial install underway
  ```

- The pre-install hook resources are processed slowly.
- The hook Job completes according to Kubernetes events:

  ```text
  Pulled image "docker.io/volcanosh/vc-webhook-manager:v1.15.0" in 3.335s
  Job completed
  ```

- However, the Helm release secret still records the hook Job as `Running`:

  ```json
  {
    "name": "volcano-admission-init",
    "kind": "Job",
    "last_run": {
      "started_at": "2026-06-17T09:39:36.258228927Z",
      "phase": "Running"
    }
  }
  ```

- Main Volcano resources are not created.

Expected result:

[...]

PR #5475, opened 2026-06-19
fix(helm): prevent pending-install hang on Helm v4 by removing hook-s…

vanshika2720:
## Summary
Fix Helm v4 pending-install hang caused by pre-install hook Job cleanup.

This PR mitigates issue #5468, where Volcano installations using Helm v4.x on Kubernetes v1.36.x can become stuck in `pending-install` after the `volcano-admission-init` pre-install hook Job completes successfully.

## Root Cause
Helm v4 introduced a new kstatus-based `StatusWatcher`. When the `admission-init` Job finishes, it emits a `SuccessCriteriaMet=True` condition just before the `Complete=True` condition. Due to `hook-succeeded` being in the Helm delete policy, Helm immediately deletes the Job before its own internal StatusWatcher has had a chance to fully process the terminal `Complete` state. This race condition causes Helm to stall and leave the release in a `pending-install` state.

Upstream Helm behavior tracking:
* helm/helm#31786
* helm/helm#32224
* helm/helm#32214

## Changes
* Removed `hook-succeeded` from the `helm.sh/hook-delete-policy` annotations for the `volcano-admission-init` Job, Role, and RoleBinding.
* Added `ttlSecondsAfterFinished` to the Job spec, defaulting to `3600` seconds (1 hour), configurable via `custom.admission_init_ttl` in `values.yaml`.

## Why the fix works
By relying on the native Kubernetes TTL controller (`ttlSecondsAfterFinished`) instead of Helm's immediate deletion (`hook-succeeded`), the Job remains in the cluster long enough for the Helm v4 watcher to successfully observe the `Complete` condition. A 3600-second TTL also ensures that if the pre-install hook fails, operators have a 1-hour window to retrieve logs and troubleshoot the failure.

## Release Notes
```release-note
Fix: Resolved a bug where Helm v4 installations would hang indefinitely in the `pending-install` state due to a race condition with the `volcano-admission-init` pre-install hook deletion.

### DIFF
--- a/installer/helm/chart/volcano/values.yaml
+++ b/installer/helm/chart/volcano/values.yaml
@@ -25,6 +25,7 @@ custom:
   metrics_enable: false
   admission_enable: true
   admission_replicas: 1
+  admission_init_ttl: 3600
   controller_enable: true
   controller_replicas: 1
   controller_metrics_enable: true
--- a/installer/helm/chart/volcano/templates/admission-init.yaml
+++ b/installer/helm/chart/volcano/templates/admission-init.yaml
@@ -69,7 +69,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-delete-policy": before-hook-creation
     "helm.sh/hook-weight": "5" # set a higher weight to reserve buffers.
   labels:
     app: volcano-admission-init
@@ -78,6 +78,7 @@ metadata:
     {{- end }}
 spec:
   backoffLimit: 3
+  ttlSecondsAfterFinished: {{ .Values.custom.admission_init_ttl | default 3600 }}
   template:
     spec:
       {{- if $admission_tolerations }}
--- a/installer/volcano-development-vap-map.yaml
+++ b/installer/volcano-development-vap-map.yaml
@@ -234,12 +234,13 @@ metadata:
   namespace: volcano-system
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-delete-policy": before-hook-creation
     "helm.sh/hook-weight": "5" # set a higher weight to reserve buffers.
   labels:
     app: volcano-admission-init
 spec:
   backoffLimit: 3
+  ttlSecondsAfterFinished: 3600
   template:
     spec:
       securityContext:
--- a/installer/volcano-development-vap.yaml
+++ b/installer/volcano-development-vap.yaml
@@ -234,12 +234,13 @@ metadata:
   namespace: volcano-system
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-delete-policy": before-hook-creation
     "helm.sh/hook-weight": "5" # set a higher weight to reserve buffers.
   labels:
     app: volcano-admission-init
 spec:
   backoffLimit: 3
+  ttlSecondsAfterFinished: 3600
   template:
     spec:
       securityContext:
--- a/installer/volcano-development.yaml
+++ b/installer/volcano-development.yaml
@@ -234,12 +234,13 @@ metadata:
   namespace: volcano-system
   annotations:
     "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-delete-policy": before-hook-creation
     "helm.sh/hook-weight": "5" # set a higher weight to reserve buffers.
   labels:
     app: volcano-admission-init
 spec:
   backoffLimit: 3
+  ttlSecondsAfterFinished: 3600
   template:
     spec:
       securityContext:

### CODE REVIEW

File: installer/helm/chart/volcano/values.yaml [hunk 1]
```diff
@@ -25,6 +25,7 @@ custom:
   metrics_enable: false
   admission_enable: true
   admission_replicas: 1
+  admission_init_ttl: 60
```
hajnalmt: I would set this one hour at least, so 3600 (reasoning in the main comment).

hajnalmt (commented):
/kind bug
/area deploy

/cc @JesseStutler
He will be the final approver for this as he raised the issue. 👍

Thanks for digging into this and putting together a clean, well-scoped fix. The reasinging behind the issue was excellent, and it's really good to know that the root cause lives in Helm's behavior rather than our chart. I did some upstream digging to back up the review, and it turns out this maps to a whole cluster of **Helm v4 kstatus status-watcher regressions**. Sharing them here so we have the full picture (asked an agent to do the this summary and align my thoughts regarding this):

#### Upstream Helm issues

**Direct match for #5468:**
- [helm/helm#32224](https://github.com/helm/helm/issues/32224) — *"[Helm v4.2.1] pre-install hook Jobs not created, release stuck in pending-install"* (OPEN). Same symptom, same Helm version (v4.2.1), reproduced with the ingress-nginx pre-install hook Jobs. It's a regression from v4.2.0 → v4.2.1.

**Watcher deletion race (affects `before-hook-creation`):**
- [helm/helm#32214](https://github.com/helm/helm/issues/32214) — *"`--wait` hangs for the full timeout on fast resource deletions in v4.2.1 (regression from v4.2.0)"* (CLOSED). Root-caused to [helm/helm#32081](https://github.com/helm/helm/pull/32081) (merged, shipped in v4.2.1).
- [helm/helm#31766](https://github.com/helm/helm/issues/31766) — *"`helm uninstall --wait` with kstatus watcher times out on resources deletion"* (OPEN).

**The one that concerns me about `ttlSecondsAfterFinished`:**
- [helm/helm#31786](https://github.com/helm/helm/issues/31786) — *"Regression from Helm 3: Waiting for hooks gets stuck and fails on timeout for **Jobs with TTL**"* (OPEN). The Helm 4 kstatus watcher gets stuck on `NotFound` precisely because the TTL controller wipes the Job after it finishes. ingress-nginx works around this by setting TTL to `0`.

The upstream fix for the TTL case is **not merged yet**:
- [helm/helm#31851](https://github.com/helm/helm/pull/31851) — *"handle Jobs with TTL in status watcher"* (OPEN)
- [helm/helm#31848](https://github.com/helm/helm/pull/31848), [#31849](https://github.com/helm/helm/pull/31849), [#31850](https://github.com/helm/helm/pull/31850) — earlier attempts, all CLOSED unmerged

#### A couple of thoughts on the approach

The `hook-delete-policy` change is spot on, so this is really just about tuning the second half:

1. **`ttlSecondsAfterFinished` may re-trigger the hang.** Per [helm/helm#31786](https://github.com/helm/helm/issues/31786), a hook Job *with* TTL is exactly the shape that makes the Helm 4 watcher get stuck on `NotFound`, and the fix is still unmerged. We are risking heretrading one hang trigger for another depending on the Helm version. But if Jesse verifies that this solves the issue, we can disregard this I think.
2. **TTL reaps failed init Jobs too.** `ttlSecondsAfterFinished` applies to both `Complete` and `Failed` Jobs, so a failed `volcano-admission-init` (after `backoffLimit: 3`) gets garbage-collected after 60s along with its pod logs. We are losing the diagnostics needed to debug the failure, you won't be able execute a `kubectl logs` in 60 seconds,  maybe raising it to 3600s is a better choice.

#### Suggestion (non-blocking)

- Keep `helm.sh/hook-delete-policy: before-hook-creation` (the real mitigation — the next install/upgrade reaps the stale Job).
- Consider **dropping `ttlSecondsAfterFinished`** or raising the default to a higher value (or at minimum document the [#31786](https://github.com/helm/helm/issues/31786) interaction and keep failed Jobs around for debugging).
- Reference [helm/helm#32224](https://github.com/helm/helm/issues/32224), [#32214](https://github.com/helm/helm/issues/32214), and [#31786](https://github.com/helm/helm/issues/31786) in the PR description so this is clearly a temporary chart-side workaround for upstream Helm regressions that we can revert once Helm fixes the watcher.



I want to add two not code related hing too. It seems that you generated the PR description with AI help. Its okay, a big chunk of this message was AI generated too, but please mention this in the description following our gudelines: https://github.com/volcano-sh/volcano/blob/master/contributing.md#ai-guidance . Also please add a Release Notes section at the end since its a user-facing change. It helps our release process work. 

Really nice catch overall. Thanks once more. 🙌

installer/helm/chart/volcano/values.yaml
vanshika2720: @hajnalmt Done!

vanshika2720:
@hajnalmt Thanks for the detailed review and for investigating the upstream Helm issues.

I've addressed the points you raised:

Increased the default admission_init_ttl from 60 to 3600 seconds to provide a more reasonable debugging window for failed admission-init Jobs.
Added references to the relevant upstream Helm issues in the PR description for additional context.
Added a Release Notes section.

The core mitigation remains unchanged: removing hook-succeeded from the hook Job lifecycle while relying on Kubernetes cleanup instead of immediate Helm-managed deletion.

I appreciate the detailed analysis and references. They were very helpful in refining the approach.

hajnalmt (commented):
/ok-to-test

Thanks let's see Jesse's opinion!

JesseStutler:
/approve
Cool, using ttl to clean is a good direction and can solve the race
/cc @hajnalmt Allow mate to merge it if he agrees with this direction

hajnalmt (commented):
Thanks I have tried to verify this but I wasn't able to reproduce the issue properly even. I have a WSL env with cgroup v1, and kind with k8s 1.36.1 needs cgroup v2, so I can't start a v1.36.1 cluster currently locally. 

What I see that even with the fix you have here helm is not installing volcano for me with 4.2.1. (with the --wait flag)

```
mkdir -p /tmp/helm421
curl -fsSL https://get.helm.sh/helm-v4.2.1-linux-amd64.tar.gz -o /tmp/helm421/helm.tar.gz
tar -xzf /tmp/helm421/helm.tar.gz -C /tmp/helm421 --strip-components=1 linux-amd64/helm
/tmp/helm421/helm version
```

The install stops at Serviceaccount creation:
```
/tmp/helm421/helm install volcano installer/helm/chart/volcano   --namespace volcano-system --create-namespace   --wait --timeout 10m   --debug
level=DEBUG msg="Original chart version" version=""
level=DEBUG msg="Chart path" path=/home/uih20178/Github/hajnalmt-volcano/installer/helm/chart/volcano
level=INFO msg="found symbolic link in path. Contents of linked file included and used" path=/home/uih20178/Github/hajnalmt-volcano/installer/helm/chart/volcano/charts/jobflow/crd/v1 resolved=/home/uih20178/Github/hajnalmt-volcano/installer/helm/chart/volcano/charts/jobflow/crd/bases
level=INFO msg="found symbolic link in path. Contents of linked file included and used" path=/home/uih20178/Github/hajnalmt-volcano/installer/helm/chart/volcano/crd/v1 resolved=/home/uih20178/Github/hajnalmt-volcano/installer/helm/chart/volcano/crd/bases
level=DEBUG msg="getting release history" name=volcano
level=DEBUG msg="number of dependencies in the chart" chart=volcano dependencies=1
level=DEBUG msg="number of dependencies in the chart" chart=jobflow dependencies=0
level=DEBUG msg="creating resource(s)" resources=1
level=DEBUG msg="using server-side apply for resource creation" forceConflicts=false dryRun=false fieldValidationDirective=Strict
level=DEBUG msg="Created resource via patch" namespace="" name=volcano-system gvk="/v1, Kind=Namespace"
level=DEBUG msg="creating release" key=sh.helm.release.v1.volcano.v1
level=DEBUG msg="starting delete resource" namespace=volcano-system name=volcano-admission-init kind=ServiceAccount
level=DEBUG msg="ignoring delete failure" namespace=volcano-system name=volcano-admission-init kind=ServiceAccount error="serviceaccounts \"volcano-admission-init\" not found"
```

But this symptom is slightly different from what Jesse saw and described in https://github.com/volcano-sh/volcano/issues/5468
because the init job never runs it hangs on the ServiceAccount before-hook-creation delete-wait waiting for resources to be deleted on a NotFound. What we see here is the regression in that came from the flag https://github.com/helm/helm/issues/32214 wait flag. Without the --wait flag this passed for me fine, I think Jesse did not use it though.

This status watcher is indeed fixed because with 4.2.2 and everything passes:
```
mkdir -p /tmp/helm422
curl -fsSL https://get.helm.sh/helm-v4.2.2-linux-amd64.tar.gz -o /tmp/helm422/helm.tar.gz
tar -xzf /tmp/helm422/helm.tar.gz -C /tmp/helm422 --strip-components=1 linux-amd64/helm
/tmp/helm422/helm version
...
```

Maybe its the Kubernetes version thats not right for me so I can't reproduce this properly @vanshika2720  Do you have any logs about the issue with the debug flag and have you verified if this is solved?

vanshika2720:
@hajnalmt Thanks for the review.

I tested the validation path on a Kubernetes v1.36 cluster with Volcano installed from this branch.

Observations:

* The HyperNode validating webhook is registered and receives requests (`admitting hypernode -- CREATE` appears in admission logs).
* `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` resources are not installed.
* The HyperNode CRD contains the new `x-kubernetes-validations` rules under `spec.members[].selector`.

However, I was able to create the following HyperNode successfully:

```yaml
apiVersion: topology.volcano.sh/v1alpha1
kind: HyperNode
metadata:
  name: invalid-hn
spec:
  tier: 1
  members:
    - type: Node
```

This object omits `selector` entirely.

From the generated CRD, the CEL validations appear to be attached to the `selector` object itself, but `selector` is not marked as required. As a result, the CEL rules are not evaluated when `selector` is absent.

Was this behavior intended as part of the validation migration, or should a member without a selector still be considered invalid?

If the latter, there may be a validation gap after moving the checks out of the webhook.

hajnalmt (commented):
Hello,
I think if the chart installs then we can disregard a hypernode bug (or at least handle in it a separate ticket, I am not sure if I understand the bug you describe correctly anyway). The question is that do you have any logs --debug enabled in the helm command without this fix in the PR? So we can verify that the bug actually disappeared.

vanshika2720:
@hajnalmt I reproduced the issue on Helm v4.2.1 + Kubernetes v1.36.x.

**Before (master branch)** 
Screenshot 2026-06-24 at 2 59 44 AM



`helm status volcano -n volcano-system` remains stuck in:

```text
STATUS: pending-install
DESCRIPTION: Initial install underway
```

**After (fix/helm-v4-pending-install)** 
Screenshot 2026-06-24 at 10 34 09 PM



The `volcano-admission-init` pre-install hook reaches:

```text
SuccessCriteriaMet
Complete
```

and the installation proceeds normally instead of remaining in `pending-install`.

For reference, the hook cleanup configuration involved is shown in 
Screenshot 2026-06-24 at 9 52 49 PM

hajnalmt (approved):
/lgtm

Thank you so much! This is enough for me.

PR #5469, opened 2026-06-18
chore(ci): pin GitHub Actions to full commit SHAs

FAUST-BENCHOU:
#### Which issue(s) this PR fixes:

Fixes https://github.com/volcano-sh/volcano/issues/5465

#### Does this PR introduce a user-facing change?

```release-note

```

Issue #5366, opened 2026-06-03
Improve CLOMonitor score for Volcano


JesseStutler:
## What is the problem you're trying to solve

Volcano's CLOMonitor score is currently below 95:

https://clomonitor.io/projects/cncf/volcano

Several checks are failing around security, release metadata, and documentation discovery. Some required documents already exist, but may not be detected because the file names or README links do not match CLOMonitor's expected patterns.

## Describe the solution you'd like

Track and fix the failed CLOMonitor checks.

Security and release tasks:

- [x] Add `SECURITY-INSIGHTS.yml` for the repo. Assigned to @JesseStutler in #5367.
- [x] Publish SBOM files with Volcano releases. Assigned to @JesseStutler in #5367.
- [ ] Add or link a dependency management policy from `SECURITY-INSIGHTS.yml`. This should be handled in the community repository together with the other security documents. 
    - **Blocked**: https://github.com/volcano-sh/community/pull/97 is going to add a security team in the community repo, after https://github.com/volcano-sh/community/pull/97 and #5367 merged, we can continue add the dependency management policy in the SECURITY-INSIGHTS.yml
- [x] Add signatures or provenance for release assets. Sub-issue: https://github.com/volcano-sh/volcano/issues/5376. Assigned to @xuezhaojun https://github.com/volcano-sh/volcano/pull/5384
- [x] Review GitHub Actions token permissions and use read-only permissions by default where possible. Assigned to @kubeboiii , ongoing PR: https://github.com/volcano-sh/volcano/pull/5369

Documentation and metadata tasks:

- [x] Fix license scanning detection. Volcano already has a FOSSA workflow, but CLOMonitor does not detect license scanning yet. We need to add a badge in the README: Assigned to @JesseStutler, ongoing pr: https://github.com/volcano-sh/volcano/pull/5420
- [x] Follow the CLOMonitor expectations, we need to rename the current `contribute.md` -> `contributing.md`, and needs to check whether there are other places to refer `contribute.md`, should all changed to `contributing.md`. Assigned to @safiya2610  , ongoing PR: https://github.com/volcano-sh/volcano/pull/5390
- [ ] Add summary table in the cncf/landscape
- [x] Website analytics should be added for the analytics check. Assigned to @kubeboiii, pr: https://github.com/volcano-sh/website/pull/517

## Additional context

`SECURITY.md` already exists and the security policy check passes, so this is not about rewriting the security policy from scratch.

[...]

PR #5481, opened 2026-06-21
chore(security): pin Dockerfile base images to SHA256 hashes

shivansh-gohem:
What type of PR is this?
/kind feature

This PR improves the OpenSSF Scorecard metrics by pinning the Docker base images to explicit SHA256 hashes. This will increase the Pinned-Dependencies metric score for Volcano to a perfect 10/10.

Which issue(s) this PR fixes:
Part of #5366 (improve CLOMonitor score).

Special notes for reviewer:
GitHub Actions were recently pinned in PR #5469. This PR covers the Docker images and missing script dependencies (like `pip` and `go install`) that were missed. Rebased onto `master` with golang builder tags bumped to `1.25.0`.

cc: @hajnalmt

### DIFF
--- a/example/custom-plugin/Dockerfile
+++ b/example/custom-plugin/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.24.0 AS builder
+FROM golang:1.25.0@sha256:5502b0e56fca23feba76dbc5387ba59c593c02ccc2f0f7355871ea9a0852cebe AS builder
 
 WORKDIR /go/src/volcano.sh/
 
@@ -28,7 +28,7 @@ RUN cd volcano && CC=/usr/local/musl/bin/musl-gcc CGO_ENABLED=1 \
 RUN cd volcano && SUPPORT_PLUGINS=yes make vc-scheduler
 
 # Build vc scheduler image with plugin
-FROM alpine:latest
+FROM alpine:3.24.1@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b
 COPY --from=builder /go/src/volcano.sh/volcano/_output/bin/vc-scheduler /vc-scheduler
 COPY --from=builder /go/src/volcano.sh/volcano/example/custom-plugin/magic.so /plugins/magic.so
 ENTRYPOINT ["/vc-scheduler"]
--- a/example/integrations/tensorflow/benchmark/Dockerfile
+++ b/example/integrations/tensorflow/benchmark/Dockerfile
@@ -3,11 +3,11 @@
 # original image is: gcr.io/kubeflow/tf-benchmarks-cpu:v20171202-bdab599-dirty-284af3,
 # the image needs an update to use the latest tf-benchmark logic
 # ref => https://github.com/tensorflow/benchmarks/tree/master/scripts/tf_cnn_benchmarks.
-FROM python:2.7
+FROM python:2.7@sha256:cfa62318c459b1fde9e0841c619906d15ada5910d625176e24bf692cf8a2601d
 MAINTAINER volcano <volcano-sh@googlegroups.com>
 RUN  apt-get update --fix-missing \
 && apt-get install -y git \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/*
-RUN pip install tf-nightly-gpu \
+RUN pip install --no-deps --require-hashes tensorflow-gpu==1.15.0 --hash=sha256:829c90021ec0fa33d74c2bcbff4e1e365fc63e875d2f04b60451c5abfeac8382 \
 && git clone https://github.com/tensorflow/benchmarks.git /opt/tf-benchmarks
--- a/hack/generate-internal-groups.sh
+++ b/hack/generate-internal-groups.sh
@@ -51,7 +51,14 @@ shift 5
   # To support running this script from anywhere, first cd into this directory,
   # and then install with forced module mode on and fully qualified name.
   cd "$(dirname "${0}")"
-  GO111MODULE=on GOOS=${OS} go install k8s.io/code-generator/cmd/{defaulter-gen@release-1.29,conversion-gen@release-1.29,client-gen@release-1.29,lister-gen@release-1.29,informer-gen@release-1.29,deepcopy-gen@release-1.29,openapi-gen@release-1.29}
+  GO111MODULE=on GOOS=${OS} go install \
+    k8s.io/code-generator/cmd/defaulter-gen@3f91291add4375b42d0f8c4d6d041bb6ed440d97 \
+    k8s.io/code-generator/cmd/conversion-gen@3f91291add4375b42d0f8c4d6d041bb6ed440d97 \
+    k8s.io/code-generator/cmd/client-gen@3f91291add4375b42d0f8c4d6d041bb6ed440d97 \
+    k8s.io/code-generator/cmd/lister-gen@3f91291add4375b42d0f8c4d6d041bb6ed440d97 \
+    k8s.io/code-generator/cmd/informer-gen@3f91291add4375b42d0f8c4d6d041bb6ed440d97 \
+    k8s.io/code-generator/cmd/deepcopy-gen@3f91291add4375b42d0f8c4d6d041bb6ed440d97 \
+    k8s.io/code-generator/cmd/openapi-gen@3f91291add4375b42d0f8c4d6d041bb6ed440d97
 )
 
 function codegen::join() { local IFS="$1"; shift; echo "$*"; }
--- a/hack/update-base-images.sh
+++ b/hack/update-base-images.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The Volcano Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# This script finds Dockerfiles in the repository and updates specific base images 
+# (e.g., alpine, openeuler) to their latest semantic version tag and corresponding digest.
+# It resolves 'latest' to a semver tag to avoid using mutable 'latest' references.
+# It requires `docker`, `curl`, and `jq` to be installed.
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${SCRIPT_ROOT}"
+
+echo "Scanning for Dockerfiles to update pinned base images..."
+
+# Helper function to get the semver tag that 'latest' currently points to
+get_latest_semver_tag() {
+    local api_repo=$1
+    local tag_regex=$2
+    local docker_repo=$3
+    
+    # Get index digest of latest tag
+    local digest=$(docker buildx imagetools inspect "${docker_repo}:latest" | grep '^Digest:' | awk '{print $2}' || true)
+    
+    if [[ -z "$digest" ]]; then
+        echo "Error: Could not fetch digest for ${docker_repo}:latest" >&2
+        return 1
+    fi
+    
+    # Query docker hub API to find tags pointing to this digest
+    local tags=$(curl -s "https://registry.hub.docker.com/v2/repositories/${api_repo}/tags/?page_size=100" | jq -r ".results[] | select(.digest == \"$digest\") | .name")
+    
+    # Filter by regex and get the highest version
+    local semver_tag=$(echo "$tags" | grep -E "$tag_regex" | sort -V | tail -n1)
+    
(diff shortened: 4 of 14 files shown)


installer/dockerfile/agent/Dockerfile
```diff
@@ -24,13 +24,13 @@ RUN go mod download
 ADD . volcano
 RUN cd volcano && make vc-agent
 
-FROM openeuler/openeuler:${OPEN_EULER_IMAGE_TAG} AS repo
+FROM openeuler/openeuler:22.03-lts-sp2@sha256:e8c56d3579a10fedd43c5e123db1143b6dcbe4cb59a6e45d6b0b2cf9b13a0c3b AS repo
```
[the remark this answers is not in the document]
JesseStutler: @shivansh-gohem Please not to hard code it

example/custom-plugin/Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN cd volcano && CC=/usr/local/musl/bin/musl-gcc CGO_ENABLED=1 \
 RUN cd volcano && SUPPORT_PLUGINS=yes make vc-scheduler
 
 # Build vc scheduler image with plugin
-FROM alpine:latest
```
JesseStutler, line 31: I saw that https://scorecard.dev/viewer/?uri=github.com/volcano-sh/volcano should be pinned to the latest + sha256 for the alpine image, if you pin the alpine to the 3.24.0 + sha256, I'm worrying that we will forget to update it in the later releases

shivansh-gohem:
Hi @JesseStutler, thanks for the review! Just to confirm our plan for the next push:
1. For Alpine, we will use `alpine:latest@sha256:...` across all Dockerfiles as you suggested, to avoid stale tags.
2. For OpenEuler, we will restore the parameterization using `ARG OPEN_EULER_IMAGE_TAG=22.03-lts-sp2@sha256:...` and `FROM openeuler/openeuler:${OPEN_EULER_IMAGE_TAG}` to avoid hardcoding.

Does this sound good to you?

example/custom-plugin/Dockerfile
hajnalmt: Using `alpine:latest@sha256:<digest>` would still be immutable, because Docker pulls the digest rather than whatever latest points to later. However, it also means we would not automatically follow newer Alpine releases when latest moves, so the tag could become misleading/stale. The concern would not disappear, it would just be harder to track.

We can add automation to refresh the pinned image before releases, and that could be a good first issue too. I think using an explicit version tag with a digest, e.g. `alpine:<version>@sha256:<digest>`, is clearer.

installer/dockerfile/agent/Dockerfile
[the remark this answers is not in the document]
hajnalmt: @shivansh-gohem will scorecard pass if we use an argument for this?

example/custom-plugin/Dockerfile
JesseStutler: Yes true, add a digest on the latest will still be stale and need to update, but when should an automation be triggered and how to notice that an image is old?

shivansh-gohem:
Ah, that's a great point @hajnalmt, I see what you mean about `latest` hiding the underlying version. Since you both agree on that, I will update the PR to use explicit version tags like `alpine:3.24.0@sha256:...` everywhere.

I will also restore the OpenEuler parameterization as requested by @JesseStutler. 

I'll push these fixes shortly!

example/custom-plugin/Dockerfile
hajnalmt: We can run a script before releases that updates these, so we don't need a notification. We should have some process description for this maybe even an agent can handle. Any process that's comfortable for you is fine I think.

JesseStutler: I suggest we should contain such a script in this PR @shivansh-gohem

shivansh-gohem:
Actually, before I push the script, I wanted to double-check: is a bash script using `docker pull` and `docker inspect` to scan and update the digests in the Dockerfiles exactly what you had in mind, or did you prefer a different mechanism/tooling for this?

shivansh-gohem:
dear @hajnalmt @JesseStutler I have pushed the `hack/update-base-images.sh` script to the PR ! It uses `docker pull` and `docker inspect` to automatically fetch the latest digests for our pinned versions. Let me know if you'd like any adjustments to it!

hack/generate-internal-groups.sh
```diff
@@ -18,6 +18,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+GOPATH="${GOPATH:-$(go env GOPATH)}"
```
hajnalmt: I thought you will remove this change in the new PR. Even if its a correct change it doesn't belong in this PR.

hack/update-gencode.sh
```diff
@@ -34,6 +34,6 @@ bash hack/generate-groups.sh "deepcopy,client,informer,lister" \
 bash hack/generate-internal-groups.sh "deepcopy,conversion" \
   volcano.sh/apis/pkg/apis/ volcano.sh/apis/pkg/apis volcano.sh/apis/pkg/apis\
   "scheduling:v1beta1"   \
-  --output-base "$(dirname "${BASH_SOURCE[0]}")/../../.." \
+  --output-base "${SCRIPT_ROOT}/_tmp" \
```
hajnalmt: This is the other change which does not belong in this PR, and I actually think it's incorrect (we can argue about this in an other PR). The main thing is that you are touching Kubernetes upgrade workflow here.

What I think happened here and in the previous PR, is that your agent went rouge running these scripts, realized it's not working, so it tried to patch it around without the proper context.

example/custom-plugin/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.24.0 AS builder
+FROM golang:1.24.0@sha256:3f7444391c51a11a039bf0359ee81cc64e663c17d787ad0e637a4de1a3f62a71 AS builder
```
hajnalmt: Can you bump this golang 1.25 if you are here?

staging/src/volcano.sh/apis/pkg/client/clientset/versioned/fake/clientset_generated.go
```diff
@@ -116,6 +112,10 @@ func (c *Clientset) IsWatchListSemanticsUnSupported() bool {
 // It's backed by a very simple object tracker that processes creates, updates and deletions as-is,
 // without applying any validations and/or defaults. It shouldn't be considered a replacement
 // for a real clientset and is mostly useful in simple unit tests.
+//
+// Compared to NewSimpleClientset, the Clientset returned here supports field tracking and thus
+// server-side apply. Beware though that support in that for CRDs is missing
```
hajnalmt: This is the other change. These are not belonging in this PR.

hack/update-base-images.sh
```diff
@@ -0,0 +1,71 @@
...
+# to the latest digest for their respective version tags.
+# It requires `docker` to be installed and running.
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${SCRIPT_ROOT}"
+
+echo "Scanning for Dockerfiles to update pinned base images..."
+
+# Find all Dockerfiles and scan for image:tag@sha256:digest
+find . -type f -name "*Dockerfile*" -print0 | while IFS= read -r -d '' file; do
```
hajnalmt: This will happily go through the vendor and _output folder too 😥 git ls-files is a lot more safer scope for this.

hack/update-base-images.sh
```diff
@@ -0,0 +1,71 @@
...
+        if [[ -z "$repo_digest" ]]; then
+            echo "  Warning: Could not get RepoDigest for ${image_tag}. Skipping."
+            continue
+        fi
+
+        new_sha=$(echo "$repo_digest" | cut -d'@' -f2)
+        
+        if [[ -n "$new_sha" && "$old_sha" != "$new_sha" ]]; then
+            echo "  Updating $old_sha -> $new_sha"
+            sed -i "s|$old_sha|$new_sha|g" "$file"
```
hajnalmt: We shouldn't use `latest@<digest>`, its antipattern and misleading. What this script does is that it checks the tag fetches the digest for that tag and changes the digest. Which works for the latest tag but that's not what we want. The script shall have a view images specified (openeauler and alpine I think) fetch the digest behind the latest tag, fetch the semver tag based on the digest, and update the image with <semver_tag>@<digest>

hack/update-base-images.sh
```diff
@@ -0,0 +1,71 @@
...
+    fi
+
+    for full_match in $matches; do
+        image_tag=$(echo "$full_match" | cut -d'@' -f1)
+        old_sha=$(echo "$full_match" | cut -d'@' -f2)
+        
+        echo "Checking ${image_tag} in ${file}..."
+        
+        # Pull the latest image for the given tag
+        if ! docker pull -q "$image_tag" > /dev/null; then
```
hajnalmt: Pulling the image is not making any sense for me. You could have used `docker buildx imagetools inspect` directly.

hajnalmt (changes requested):
Thanks for the update. I think this thread became quite confusing, I added my comments below. I tried to be elaborative, but I didn't check first that a lot of changes, which are not belonging here were carried away here from the previous PR again.

shivansh-gohem:
Thanks @hajnalmt, I've pushed an update to this pr !!! 
- Reverted the unintended code generation and script changes (`hack/generate-internal-groups.sh`, etc.) to keep the PR focused.
- Bumped golang to `1.25.0` in the custom plugin Dockerfile.
- Rewrote `hack/update-base-images.sh` to use `git ls-files` and `docker buildx imagetools inspect`. It now correctly resolves the semver tags for `alpine` and `openeuler` instead of using `latest`. 
 i hope now it may sound better !!!!!!

shivansh-gohem:
@hajnalmt have a review !!

pkg/scheduler/api/resource_info.go
```diff
@@ -168,27 +168,17 @@ func (r *Resource) Clone() *Resource {
...
-		}
-		if val == float64(math.MaxInt64) {
-			return "maxInt"
-		}
-		return fmt.Sprintf("%0.2f", val)
-	}
-
-	var sb strings.Builder
-	fmt.Fprintf(&sb, "cpu %s, memory %s", format(r.MilliCPU), format(r.Memory))
+	str := fmt.Sprintf("cpu %0.2f, memory %0.2f", r.MilliCPU, r.Memory)
```
hajnalmt, line 171: Please don't revert changes on master already in your commits. Maybe a merge went bad on your branches, but I don't see a merge commit in your history, so I am not sure.

hack/update-base-images.sh
```diff
@@ -0,0 +1,88 @@
...
+echo "Latest alpine is alpine:${alpine_tag}@${alpine_digest}"
+
+# 2. Update OpenEuler
+echo "Resolving latest openeuler tag..."
+openeuler_tag=$(get_latest_semver_tag "openeuler/openeuler" '^[0-9]+\.[0-9]+-lts-sp[0-9]+$' "openeuler/openeuler")
+openeuler_digest=$(docker buildx imagetools inspect "openeuler/openeuler:${openeuler_tag}" | grep '^Digest:' | awk '{print $2}')
+echo "Latest openeuler is openeuler/openeuler:${openeuler_tag}@${openeuler_digest}"
+
+# Update all Dockerfiles tracked by git
+dockerfiles=$(git ls-files "*Dockerfile*")
```
hajnalmt, line 72: This is the way thanks!

hajnalmt (changes requested):
Thanks! This is a lot better, the script is good now from my side.

The problem is that you have reverted changes that I didn't ask for so scorecard is 9/10 again. I asked to revert the side effects of running the codegen script not the actual version pinning changes, sorry if I didn't phrase it properly, I marked every line which should be reverted, I didn't mean file level reverts.

Also please squash your commits, I can't follow whats happening here properly, and your change includes now random revert changes in `resource_info.go` from master.

shivansh-gohem:
@hajnalmt have a review!!!

example/integrations/tensorflow/benchmark/Dockerfile
```diff
@@ -3,11 +3,11 @@
...
 # ref => https://github.com/tensorflow/benchmarks/tree/master/scripts/tf_cnn_benchmarks.
-FROM python:2.7
+FROM python:2.7@sha256:cfa62318c459b1fde9e0841c619906d15ada5910d625176e24bf692cf8a2601d
 MAINTAINER volcano <volcano-sh@googlegroups.com>
 RUN  apt-get update --fix-missing \
 && apt-get install -y git \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/*
-RUN pip install tf-nightly-gpu \
+RUN pip install --no-deps --require-hashes tensorflow-gpu==1.15.0 --hash=sha256:829c90021ec0fa33d74c2bcbff4e1e365fc63e875d2f04b60451c5abfeac8382 \
```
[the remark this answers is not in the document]
hajnalmt: This is is an example, I think its fine. Never intended to be buildable.

hajnalmt (commented):
/lgtm

Thanks for the effort. This will be fine from my side.
Validated the score card on the head and its good finally. Letting @JesseStutler to have a final review/approve on this.
`Aggregate score: 10.0 / 10`

JesseStutler:
/approve
Thanks to @hajnalmt mate's detailed help!

shivansh-gohem:
heyy @hajnalmt @JesseStutler thanks for all your support throughout!!  i will be contributing more to help the community !

PR #5482, opened 2026-06-21
fix(scheduler): use milli-units for scalar resources in GetInqueueRes…

WHOIM1205:
## Description

Fix incorrect scalar resource accounting in `GetInqueueResource()` caused by a unit mismatch between scalar resources stored in `api.Resource` and values derived from `PodGroup.MinResources`.

Scalar resources (e.g. GPUs, hugepages, and extended resources) are represented in milli-units throughout Volcano's resource model. However, `GetInqueueResource()` used `resource.Quantity.Value()` when calculating scalar inqueue reservations while subtracting allocations already stored in milli-units, resulting in inconsistent accounting and incorrect queue admission decisions.

## Root Cause

The scalar resource branch in `GetInqueueResource()` used:

```go
float64(rQuantity.Value())
```

while:

```go
allocated.ScalarResources
```

stores scalar resources in milli-units.

This caused:

- Scalar reservations to be reported 1000x smaller than expected
- Incorrect subtraction for partially allocated jobs
- GPU and extended-resource dimensions to disappear from inqueue accounting
- Queue capacity checks to evaluate against incomplete resource usage

## Fix

Replace `Value()` with `MilliValue()` for scalar resources:

```go
if allocatedMount, ok := allocated.ScalarResources[rName]; !ok {
    inqueue.ScalarResources[rName] = float64(rQuantity.MilliValue())
} else {
    reservedScalarRes := float64(rQuantity.MilliValue()) - allocatedMount
    if reservedScalarRes > 0 {
        inqueue.ScalarResources[rName] = reservedScalarRes
    }
}
```

This keeps scalar resource accounting consistent with the rest of the scheduler resource model and aligns inqueue reservations with allocated, capability, and min-resource calculations.

### Before

- GPU and extended-resource inqueue reservations were significantly undercounted
- Capacity and proportion plugins could over-admit GPU workloads
- Large numbers of unschedulable pods could be created
- Queue quota enforcement for scalar resources could be bypassed

### After

- Scalar resource accounting is consistent across scheduler components
- Queue admission correctly reserves GPU and extended resources
- Capacity and overcommit protections behave as intended
- Excess pending pod creation is prevented

## Testing

Added regression tests covering scalar resource accounting in:

```text
pkg/scheduler/plugins/util/util_test.go
```

Verified:

- GPU reservations are reported in milli-units
- Partially allocated GPU resources are subtracted correctly
- Existing capacity, proportion, overcommit, and util plugin tests continue to pass

## Type of Change

- [x] Bug fix
- [x] Added regression tests
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/pkg/scheduler/plugins/util/util.go
+++ b/pkg/scheduler/plugins/util/util.go
@@ -126,10 +126,15 @@ func GetInqueueResource(job *api.JobInfo, allocated *api.Resource) *api.Resource
 			if inqueue.ScalarResources == nil {
 				inqueue.ScalarResources = make(map[v1.ResourceName]float64)
 			}
+			// Scalar resources are stored in milli-units throughout api.Resource
+			// (see NewResource, which uses MilliValue() for scalar resources), so
+			// allocated.ScalarResources is in milli-units. Use MilliValue() here as
+			// well to keep units consistent; otherwise scalar (e.g. GPU/hugepages)
+			// inqueue accounting is off by 1000x.
 			if allocatedMount, ok := allocated.ScalarResources[rName]; !ok {
-				inqueue.ScalarResources[rName] = float64(rQuantity.Value())
+				inqueue.ScalarResources[rName] = float64(rQuantity.MilliValue())
 			} else {
-				reservedScalarRes := float64(rQuantity.Value()) - allocatedMount
+				reservedScalarRes := float64(rQuantity.MilliValue()) - allocatedMount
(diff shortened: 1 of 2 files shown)


hajnalmt (commented):
Hello,

Why have you resubmitted https://github.com/volcano-sh/volcano/pull/5391 ?
I think the fix tere is much more precise, and there weren't any review updates from others so we can merge that if it's alright.

PR #5490, opened 2026-06-23
benchmark: isolate kwok nodes and secure scheduler pprof

JesseStutler:
#### What type of PR is this?

/kind feature
/kind cleanup

This PR updates the benchmark environment and scheduler diagnostics used for large KWOK-based scheduling tests.

For KWOK benchmarks, it keeps system DaemonSets off KWOK nodes so kindnet and kube-proxy do not create thousands of extra system pods on simulated nodes. The benchmark node creation flow now taints KWOK nodes and patches the system DaemonSets to exclude nodes labeled `type=kwok`.

For scheduler diagnostics, it hardens the optional pprof endpoint:

- keeps pprof disabled by default
- serves pprof on a dedicated address instead of the metrics listener
- defaults the pprof listener to `127.0.0.1:6060`
- validates that pprof is bound only to literal loopback addresses
- uses a private pprof mux instead of the default HTTP mux
- wires the pprof address through the Helm chart

This keeps `/metrics` behavior unchanged while allowing controlled local profiling through `kubectl port-forward` or an SSH tunnel.

N/A

pprof can expose sensitive runtime data and can add load while profiles are collected. The endpoint is therefore opt-in and loopback-only by default. Remote access should be done through an authenticated tunnel such as `kubectl port-forward`.

Tests run:

```text
GOCACHE=/tmp/go-build go test ./cmd/scheduler/app/options
GOCACHE=/tmp/go-build go test ./cmd/scheduler/app
JOBS=50 REPLICAS=16 MIN_AVAILABLE=16 bash scripts/run-tests.sh gang --template=cases/case-template.yaml
```

```release-note
Added an optional loopback-only scheduler pprof endpoint controlled by --enable-pprof and --pprof-address. The endpoint remains disabled by default.
```


JesseStutler:
Wrongfully opened PR, will open a new one

PR #5496, opened 2026-06-24
[Enhancement] Optimize network-topology-aware plugin and document scoring extensibility

### LINKED ISSUE #5082: [Enhancement] Optimize PrioritizeNodes concurrency and regulate BatchNodeOrderFn usage usage

During large-scale benchmark profiling (e.g., thousands of pods and nodes), we observed that the node scoring phase (`PrioritizeNodes`) can become a scheduling bottleneck. Specifically, there are two distinct architectural issues regarding performance and the current plugin Extension Points:

1. **Lock Contention in mapFn (`NodeMapFn` / `NodeOrderMapFn`)**:
In `pkg/scheduler/util/scheduler_helper.go -> PrioritizeNodes`, the framework spins up concurrent workers via `workqueue.ParallelizeUntil` to evaluate node scores. However, inside the evaluation closure, it employs a single global `workerLock.Lock()` to safely write back to `pluginNodeScoreMap` and `nodeOrderScoreMap`. For clusters with thousands of nodes, evaluating scores becomes heavily bottlenecked by this lock contention, which essentially serializes a significant portion of the score aggregation.

2. **Misuse and Lack of Guidelines for `BatchNodeOrderFn`**:
The `BatchNodeOrderFn` extension point is extremely powerful as it exposes the entire node list to plugins (needed for I/O batching or topology gathering). However, unlike `NodeOrderFn` which is parallelized by the scheduling framework, it relies wholly on the Plugin writers to implement their own concurrency control. We have observed that plugin developers are often confused between `NodeOrderFn` and `BatchNodeOrderFn`, sometimes placing CPU-heavy computation in `BatchNodeOrderFn` using simple serial `for` loops, resulting in disastrous performance degradation. We lack standard documentation and constraints to guide the usage of these overlapping extension points. Note: `NodeMapFn` and `NodeReduceFn` cannot be simply removed yet as they are heavily relied upon by out-of-tree and commercial implementations.

Since removing `NodeMapFn` and `NodeReduceFn` breaks backward compatibility for commercial distributions and out-of-tree plugins, we should keep them but optimize the underlying implementation:

1. **Lock-Free Map Execution in `PrioritizeNodes`**:
We should eliminate the `sync.Mutex` (`workerLock`) inside the `ParallelizeUntil` loop. This can be achieved by pre-allocating an array/slice of the same length as the `nodes` slice (since the index is guaranteed unique per goroutine) to collect `mapScores` and `orderScore`. The aggregation into the global maps (`pluginNodeScoreMap` and `nodeOrderScoreMap`) can be done in a quick serial loop after `ParallelizeUntil` finishes, or reduced via a fast lock-free mechanism.

2. **Concurrent Execution of Independent Score Phases**:
`BatchNodeOrderFn` and `NodeOrderMapFn -> NodeOrderReduceFn` evaluations are fundamentally independent sources of scoring. Instead of running them sequentially within `PrioritizeNodes`, it is feasible to dispatch the `BatchNodeOrderFn` evaluation into its own goroutine while the `ParallelizeUntil` processes the map/reduce workloads.

3. **Standardize and Document Plugin Guidelines**:
Introduce explicit developer documentation inside the codebase and on `website` denoting best practices:
- Directing CPU-bound scalar assignments strictly to `NodeOrderFn` to benefit from framework-level parallelism.
- Restricting `BatchNodeOrderFn` primarily to plugins requiring aggregated state (PreScore semantics) or I/O consolidation (Extender webhooks), while explicitly mandating plugin developers to utilize `workqueue.ParallelizeUntil` if they iterate through nodes within this function.

### Additional context

This optimization was identified while inspecting metrics for `.NodeOrderFn` latencies during high-throughput gang scheduling performance tests scenarios. 

### Documentation Updates

shivansh-gohem:
#### What type of PR is this?
/kind feature

This PR implements tasks 3 & 4 of issue #5082:
1. Refactors the `network-topology-aware` plugin to use `workqueue.ParallelizeUntil` instead of serial loops for node iteration in `batchNodeOrderFn`. This eliminates scheduling bottlenecks while preventing data races by using pre-allocated intermediate slices.
2. Introduces explicit developer documentation detailing standard best practices for Node Scoring Extensibility points (`NodeOrderFn` vs `BatchNodeOrderFn`) inside the codebase.

Fixes #5082

#### Special notes for reviewer:
This branch specifically handles the `network-topology-aware` plugin parallelization and the developer documentation as discussed in the issue comments.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
+++ b/pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
@@ -36,7 +36,13 @@ import (
 
 const (
 	// PluginName indicates name of volcano scheduler plugin.
-	PluginName            = "network-topology-aware"
+	PluginName = "network-topology-aware"
+
+	// batchNodeOrderFnWorkers is the number of concurrent goroutines used during node scoring
+	batchNodeOrderFnWorkers = 16
+)
+
+const (
 	FullScore             = 1.0
 	ZeroScore             = 0.0
 	NetworkTopologyWeight = "weight"
@@ -500,7 +506,7 @@ func (nta *networkTopologyAwarePlugin) batchNodeOrderFnForNormalPods(ssn *framew
 	}
 
 	nodeScoresList := make([]float64, len(nodes))
-	workqueue.ParallelizeUntil(context.TODO(), 16, len(nodes), func(index int) {
+	workqueue.ParallelizeUntil(context.TODO(), batchNodeOrderFnWorkers, len(nodes), func(index int) {
 		node := nodes[index]
 		totalScore := 0.0
 		for tier := nta.hyperNodesTier.minTier; tier <= nta.hyperNodesTier.maxTier; tier++ {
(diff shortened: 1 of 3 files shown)


shivansh-gohem:
@JesseStutler  please have a look at this pr !!

pkg/scheduler/plugins/network-topology-aware/network_topology_aware.go
```diff
@@ -563,25 +577,50 @@ func (nta *networkTopologyAwarePlugin) batchNodeOrderFnForNetworkAwarePods(ssn *
 	if allocatedHyperNode == "" {
 		return nodeScores, nil
 	}
+
+	// Pre-build node to hypernode map to avoid O(N) lookup inside the parallel loops.
+	nodeToHyperNode := make(map[string]string)
```
LiZhenCheng9527: Is it built every time this function is called?
I recommend building it once when the session opens and reusing it afterward.

shivansh-gohem:
@LiZhenCheng9527 Good catch! I have updated the code to build `nodeToHyperNode` once during `OnSessionOpen` so that it can be reused across all subsequent batch scoring calls for the session, instead of rebuilding it every time the function is called. Thanks for the suggestion!

PR #5501, opened 2026-06-24
PR for unnecessary DeepCopy and Clone operations 

### LINKED ISSUE #5173: [Performance] Performance impact caused by excessive and unnecessary `DeepCopy` and `Clone` operations
### Description

Inspired by #5168, Currently, the Agent-Scheduler needs to take a snapshot of the node when scheduling each pod to fix the resource view. However, if the cluster is large and there are many pods, `UpdateSnapshot` will trigger a large number of node clones, including both node deep copies and pod deep copies (introduced from https://github.com/volcano-sh/volcano/pull/5086). In many cases, deep copies are unnecessary, as they only read values ​​that will not be modified. The same applies to the Batch Scheduler. The Batch Scheduler triggers a snapshot every time a session is started, including both node clones and job clones. Both node and job clones will add up to twice the number of pod deep copies, which will have a performance impact.

1.
2.
3.

We should refactor existing Clone and DeepCopy code in the codebase. If it is indeed necessary to modify the cached Node or Pod during the scheduling process, then DeepCopy or Clone should be performed as needed. When making a snapshot, we should minimize such operations and only copy the data required for scheduling, while shallow copying the rest is sufficient.

> Please note that this is not a good first issue. Do not assign it casually. Even if you remove an existing DeepCopy or Clone, you should still confirm that the code involving resource modifications is correct.

master

_No response_

safiya2610:
Fixes #5173

Replaced deep snapshot cloning with ShallowClone()
Introduced NodeInfo.ShallowClone() and JobInfo.ShallowClone() so scheduler snapshots no longer deep-copy all tasks by default.
Optimized scheduler cache snapshot creation
Updated the scheduler cache to use ShallowClone() instead of creating fully deep-cloned NodeInfo and JobInfo objects.
Optimized Kubernetes snapshot utilities
Modified snapshot update logic to reuse shallow-cloned NodeInfo objects instead of repeatedly deep-copying task data.
Reduced cloning during candidate selection
In preemption and reclaim paths, avoided cloning every candidate task. Tasks are cloned only when they are actually passed to mutating operations.
Moved cloning to mutation points
Instead of cloning tasks eagerly, tasks are cloned only immediately before mutating APIs such as Allocate(), Pipeline(), and Evict().
Removed redundant deep copies
Eliminated unnecessary node.DeepCopy() calls in the agent scheduler cache where the data was not modified.
Preserved correctness with targeted deep copies
When shallow-cloned tasks must be modified (e.g., Allocate() updates Pod.Spec.NodeName), only the Pod is deep-copied instead of deep-copying the entire task or snapshot.
Overall optimization

Previously, the scheduler performed deep cloning for many objects during snapshot creation, even though most snapshots were read-only. Your changes make snapshots lightweight by sharing immutable data through ShallowClone(), while performing deep copies only at the exact points where mutations occur. This significantly reduces memory allocations and CPU overhead without compromising correctness.

### DIFF
--- a/pkg/agentscheduler/cache/cache.go
+++ b/pkg/agentscheduler/cache/cache.go
@@ -1000,7 +1000,7 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 			continue
 		}
 
-		snapshot.Nodes[value.info.Name] = value.info.Clone()
+		snapshot.Nodes[value.info.Name] = value.info.ShallowClone()
 	}
 	klog.V(3).InfoS("SnapShot for scheduling", "NodeNum", len(snapshot.Nodes))
 	return snapshot
--- a/pkg/agentscheduler/cache/event_handlers.go
+++ b/pkg/agentscheduler/cache/event_handlers.go
@@ -458,8 +458,7 @@ func (sc *SchedulerCache) SyncNode(nodeName string) error {
 	if !sc.nodeCanAddCache(node) {
 		return nil
 	}
-	nodeCopy := node.DeepCopy()
-	return sc.AddOrUpdateNode(nodeCopy)
+	return sc.AddOrUpdateNode(node)
 }
 
 func (sc *SchedulerCache) nodeCanAddCache(node *v1.Node) bool {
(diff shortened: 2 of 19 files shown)


safiya2610:
@JesseStutler Hello, Can you Please review this PR once, and suggest any changes required.

PR #5502, opened 2026-06-24
fix(capacity): copy parent ancestors to avoid slice aliasing in hiera…

WHOIM1205:
## Summary

Fix a slice-aliasing issue in hierarchical queue ancestor construction that could corrupt ancestor chains in deep queue hierarchies, leading to incorrect quota enforcement and resource accounting.

## Problem

`updateAncestors()` built a child queue's ancestor chain using:

```go
append(parent.ancestors, parentInfo.UID)
```

When the parent ancestor slice had spare capacity, Go reused the same backing array. As a result, sibling subtrees could unintentionally share memory, allowing later updates to overwrite ancestor entries belonging to other queues.

This could cause:

- Incorrect hierarchical quota validation
- Wrong enqueue admission decisions
- Resource accounting being attributed to the wrong ancestor queue
- Potential quota bypass or starvation scenarios in deep hierarchical queue trees

## Root Cause

The ancestor list stored in `queueOpts` was reused via `append()`, which allowed sibling queues to alias the same backing array.

Since hierarchical capacity and enqueue checks rely on ancestor chains, corrupted ancestry could propagate into scheduling decisions.

### Ancestor Construction

- Replace direct `append()` usage in `updateAncestors()` with explicit slice allocation and copy.
- Ensure every queue receives an independent ancestor chain.

### Defensive Hardening

Avoid mutating stored ancestor slices in:

- `checkQueueAllocatableHierarchically()`
- `checkJobEnqueueableHierarchically()`

Both paths now construct temporary lists using fresh slices before appending queue IDs.

## Testing

Added regression test:

### `Test_updateAncestors_noSliceAliasing`

The test:

1. Creates a deep hierarchical queue tree with multiple sibling branches.
2. Builds ancestor chains using the scheduler logic.
3. Verifies every queue receives the expected ancestor path.
4. Ensures sibling branches cannot overwrite each other's ancestry.

The test reproduces the bug on the old implementation and passes with the fix.

## Verification

- ✅ Regression test passes
- ✅ Regression test fails on the pre-fix implementation
- ✅ Full capacity plugin test suite passes
- ✅ `go vet` passes successfully

## Impact

This change guarantees that ancestor chains remain isolated and immutable across sibling subtrees, restoring correct hierarchical quota enforcement, enqueue validation, and resource accounting for deep queue hierarchies while preserving existing behavior for shallow hierarchies.

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -1480,7 +1480,16 @@ func (cp *capacityPlugin) updateAncestors(queue *api.QueueInfo, ssn *framework.S
 	}
 
 	cp.queueOpts[parentInfo.UID].children[queue.UID] = cp.queueOpts[queue.UID]
-	cp.queueOpts[queue.UID].ancestors = append(cp.queueOpts[parentInfo.UID].ancestors, parentInfo.UID)
+	// Build a fresh backing array instead of append-ing onto the parent's
+	// ancestors slice. append would reuse the parent's backing array whenever it
+	// has spare capacity, causing sibling subtrees' ancestors slices to alias the
+	// same memory and overwrite each other in deep hierarchies, which corrupts the
+	// ancestor chain used for hierarchical quota enforcement.
+	parentAncestors := cp.queueOpts[parentInfo.UID].ancestors
+	ancestors := make([]api.QueueID, len(parentAncestors)+1)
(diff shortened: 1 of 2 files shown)


WHOIM1205:
Hi @hajnalmt , this PR includes a fix for the hierarchical queue ancestor slice-aliasing issue along with a regression test. All tests are passing. I'd appreciate a review when you have time. Thanks!

PR #5505, opened 2026-06-25
fix(predicates): register PreFilter plugins to skip no-op filters

JesseStutler:
#### What type of PR is this?

/kind bug

Some predicates plugins already have upstream `PreFilter` logic that can mark their `Filter` as skipped when the pod has nothing for that plugin to check. Volcano registered these plugins as `Filter` plugins, but not all of their `PreFilter` hooks were wired into the predicates plugin.

This PR registers these upstream `PreFilter` hooks so Volcano can avoid no-op per-node filter calls:

- `CSILimits`: skip `Filter` when the pod has no PVC, generic ephemeral volume, or inline-migratable in-tree volume. This matters because the service account admission plugin injects a projected token volume, so `len(pod.Spec.Volumes) == 0` is usually false even when there is no CSI-backed volume.
- `NodeAffinity`: skip `Filter` when the pod does not have node affinity terms that need per-node evaluation.
- `VolumeZone`: skip `Filter` when the pod has no persistent volumes that need zone checks.

Fixes 

```release-note
NONE
```


hzxuzhonghu (approved):
/lgtm

PR #5510, opened 2026-06-25
fix(nodeorder): log correct plugin weight in NodeOrderFn

### LINKED ISSUE #5509: nodeorder: NodeOrderFn debug log reports incorrect plugin weight
### Description

While reviewing the nodeorder plugin, I noticed that the debug log inside `NodeOrderFn()` reports `imageLocalityWeight` for every score plugin.

Although the scheduler correctly computes scores using each plugin's configured `p.weight`, the log message always prints `pp.weight.imageLocalityWeight`.

As a result, V(5) debug logs can be misleading when different score plugins are configured with different weights.

### Expected behavior

The debug log should display the weight associated with the current plugin (`p.weight`).

### Current behavior

All score plugins report the ImageLocality weight in the debug output.

### Possible fix

Replace the logged weight with `p.weight` inside `NodeOrderFn()`.

1. Configure the nodeorder plugin with different weights for multiple score plugins (for example, NodeAffinity and ImageLocality).
2. Run the scheduler with log verbosity `-v=5`.
3. Schedule a pod so `NodeOrderFn()` is executed.
4. Observe the debug logs emitted by `NodeOrderFn()`.

**Current behavior**

The debug log always reports `imageLocalityWeight` as the plugin weight, regardless of which score plugin produced the score.

**Expected behavior**

The debug log should report the weight of the current score plugin (`p.weight`) so that the logged weight matches the value actually used in score calculation.

master

_No response_

A69SHUBHAM:
## What this PR does

The debug log inside `NodeOrderFn()` reports `imageLocalityWeight` for every score plugin, even though the scheduler computes the score using each plugin's own configured weight (`p.weight`).

This PR updates the debug log to report the actual plugin weight used during score calculation, making V(5) scheduler logs consistent with the scheduler's behavior.

### Why is this change needed?

When multiple score plugins are configured with different weights, the current debug output can be misleading during troubleshooting because it always reports the ImageLocality weight regardless of which plugin generated the score.

### Does this PR introduce any user-facing changes?

No.

This change only corrects debug logging and does not affect scheduler behavior or score calculation.

Fixes #5509

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -323,7 +323,10 @@ func (pp *NodeOrderPlugin) NodeOrderFn(task *api.TaskInfo, node *api.NodeInfo, k
 
 		// Score from each plugin is multiplied by its configured weight before being added to the total.
 		nodeScore += float64(score) * float64(p.weight)
-		klog.V(5).Infof("Node: %s, task<%s/%s> %s weight %d, score: %f", node.Name, task.Namespace, task.Name, name, p.weight, float64(score)*float64(p.weight))
+		// Score from each plugin is multiplied by its configured weight before being added to the total.
+		weightedScore := float64(score) * float64(p.weight)
+		nodeScore += weightedScore
+		klog.V(5).Infof("Node: %s, task<%s/%s> plugin %s weight %d, raw score: %d, weighted score: %f", node.Name, task.Namespace, task.Name, name, p.weight, score, weightedScore)
 	}
 	klog.V(4).Infof("Nodeorder Total Score for task<%s/%s> on node %s is: %f", task.Namespace, task.Name, node.Name, nodeScore)
 	return nodeScore, nil
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -322,8 +322,6 @@ func (pp *NodeOrderPlugin) NodeOrderFn(task *api.TaskInfo, node *api.NodeInfo, k
 		}
 
 		// Score from each plugin is multiplied by its configured weight before being added to the total.
-		nodeScore += float64(score) * float64(p.weight)
-		// Score from each plugin is multiplied by its configured weight before being added to the total.
 		weightedScore := float64(score) * float64(p.weight)
 		nodeScore += weightedScore
 		klog.V(5).Infof("Node: %s, task<%s/%s> plugin %s weight %d, raw score: %d, weighted score: %f", node.Name, task.Namespace, task.Name, name, p.weight, score, weightedScore)
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -321,9 +321,9 @@ func (pp *NodeOrderPlugin) NodeOrderFn(task *api.TaskInfo, node *api.NodeInfo, k
 			return 0, status.AsError()
 		}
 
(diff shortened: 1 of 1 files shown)


A69SHUBHAM:
/cc @hajnalmt 

Would appreciate a review when you have time. Thanks!

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -321,9 +321,9 @@ func (pp *NodeOrderPlugin) NodeOrderFn(task *api.TaskInfo, node *api.NodeInfo, k
 			return 0, status.AsError()
 		}
 
-		// If imageLocalityWeight is provided, host.Score is multiplied with weight, if not, host.Score is added to total score.
+		// Score from each plugin is multiplied by its configured weight before being added to the total.
 		nodeScore += float64(score) * float64(p.weight)
-		klog.V(5).Infof("Node: %s, task<%s/%s> %s weight %d, score: %f", node.Name, task.Namespace, task.Name, name, pp.weight.imageLocalityWeight, float64(score)*float64(p.weight))
+		klog.V(5).Infof("Node: %s, task<%s/%s> %s weight %d, score: %f", node.Name, task.Namespace, task.Name, name, p.weight, float64(score)*float64(p.weight))
```
hajnalmt: Yeah, this is correct, thaks for the switch from `pp.weight.imageLocalityWeight` to `p.weight`, I bumped into this in a debug session too. This loop iterates over all `NodeOrderScorePlugins`, not only image locality. `p.weight` is the effective configured weight for the current plugin.

One small suggestion: could we make the log clearer by separating the raw plugin score from the weighted score added to `nodeScore`?

```suggestion
		// Score from each plugin is multiplied by its configured weight before being added to the total.
		weightedScore := float64(score) * float64(p.weight)
		nodeScore += weightedScore
		klog.V(5).Infof("Node: %s, task<%s/%s> plugin %s weight %d, raw score: %d, weighted score: %f", node.Name, task.Namespace, task.Name, name, p.weight, score, weightedScore)
```
This would improve debugging in general.

hajnalmt (commented):
/kind bug
/area scheduling
/approve

Thanks, this fix is correct. I left one small line comment to make the log output clearer by separating raw score from weighted score. Once that suggestion is applied, I will add `lgtm` too.

pkg/scheduler/plugins/nodeorder/nodeorder.go
A69SHUBHAM: Thanks for the review! I've addressed the suggestion and pushed the changes.

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -321,9 +321,12 @@ func (pp *NodeOrderPlugin) NodeOrderFn(task *api.TaskInfo, node *api.NodeInfo, k
 			return 0, status.AsError()
 		}
 
-		// If imageLocalityWeight is provided, host.Score is multiplied with weight, if not, host.Score is added to total score.
+		// Score from each plugin is multiplied by its configured weight before being added to the total.
 		nodeScore += float64(score) * float64(p.weight)
```
hajnalmt: You need to delete this line and the previous! Now accumulaion won't work correctly because nodeScore is doubled!

hajnalmt (changes requested):
/hold

Thanks but the copy paste made nodeScore doubled now. Please correct it!

pkg/scheduler/plugins/nodeorder/nodeorder.go
A69SHUBHAM: Thanks! I've removed the duplicate accumulation and pushed the update.

hajnalmt (approved):
/unhold

Thanks, lets if it passes CI, if yes I will merge it.

hajnalmt (approved):
/lgtm

PR #5516, opened 2026-06-26
fix(networkqos): match cni plugin by type

### LINKED ISSUE #5515: network-qos CNI config update can add duplicate plugin entries
`network-qos` CNI config handling looks like it only detects an existing plugin by `name`.

That works for the config Volcano writes itself, but CNI plugin entries are normally identified by `type`, and `name` is not required for every plugin entry. If the conflist already has a `network-qos` plugin with only `type: network-qos`, `prepare` appends another plugin instead of updating the existing one.

Delete has the same problem. A type-only `network-qos` entry is not removed.

Steps to reproduce:

1. Use a CNI conflist with this plugin entry:

```
   {
     "type": "network-qos",
     "args": {
       "colocation": "true"
     }
   }

```
2. Run network-qos prepare so it updates the conflist.

Expected:
Existing `network-qos` plugin is updated.

Actual:
A second `network-qos` plugin is appended.

This can leave duplicate CNI plugins in the chain, which is pretty risky for a host-level network config file.

Fix should probably match the plugin by `name` or `type`, and add tests for add/update/delete when `name` is missing.

avinxshKD:
#### What type of PR is this?

/kind bug

`network-qos` CNI config update/delete only matched an existing plugin by `name`.

Some CNI entries may only have `type`, so an existing `type: network-qos` entry could be missed. In that case prepare appends another `network-qos` plugin instead of updating the existing one, and delete does not remove the old entry.

This matches by `name` or `type` and adds test coverage for type-only entries.

Fixes #5515

NONE

```release-note
NONE

### DIFF
--- a/pkg/networkqos/cni/config.go
+++ b/pkg/networkqos/cni/config.go
@@ -76,17 +76,7 @@ func (p *ConfHandler) AddOrUpdateCniPluginToConfList(configPath, name string, pl
 			return fmt.Errorf("failed to get cni from cni conflist")
 		}
 
-		pluginName, ok := existedPlugin["name"]
-		if !ok {
-			continue
-		}
-
-		pluginNameInStr, ok := pluginName.(string)
-		if !ok {
-			continue
-		}
-
-		if pluginNameInStr == name {
+		if isCNIPluginMatched(existedPlugin, name) {
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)


avinxshKD:
Hey @JesseStutler @hzxuzhonghu pls take a look when you get time. This keeps the network-qos CNI config handling idempotent when an existing plugin entry is identified by type instead of name, and covers both update and delete paths.
Thanks

avinxshKD:
ptal @JesseStutler @hajnalmt when get chance

hzxuzhonghu (approved):

avinxshKD:
ptal @JesseStutler when free

PR #5521, opened 2026-06-27
Avoid skipping nodes prematurely when no reclaim candidates remain.

### LINKED ISSUE #5520: reclaim action: reclaimForTask skips nodes with sufficient FutureIdle when reclaimees is empty, causing unnecessary evictions and gang scheduling failures
### Summary

In `reclaimForTask` ([`reclaim.go:204-207`](https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/actions/reclaim/reclaim.go#L204-L207)), when a node has no `Running+Preemptable` tasks from other queues (`reclaimees == []`), the node is skipped unconditionally — even if the node's `FutureIdle()` (idle + releasing resources) is already sufficient to host the pending task without any further eviction.

This is inconsistent with:
- `allocate.go` `allocateResourcesForTask` ([`allocate.go:801-812`](https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/actions/allocate/allocate.go#L801-L812)), which explicitly pipelines a task onto a node when `task.InitResreq.LessEqual(node.FutureIdle(), api.Zero)`.
- The comment in `ValidateVictims` ([`scheduler_helper.go:314-317`](https://github.com/volcano-sh/volcano/blob/master/pkg/scheduler/util/scheduler_helper.go#L314-L317)), which explicitly states that an empty victims list is valid when `FutureIdle` is already sufficient.
- This is not another FutureIdle accounting issue.

FutureIdle is already computed correctly after recent fixes.

The remaining problem is that reclaimForTask never reaches the FutureIdle check when reclaimees becomes empty within the same scheduling session, causing reusable FutureIdle resources to be ignored.

### Root Cause

`Statement.Evict()` immediately transitions a task's status from `Running` to `Releasing` (and calls `node.UpdateTask()` in-place). When `reclaimForTask` processes subsequent tasks of the same job within the same scheduling session, it filters reclaimees by `taskOnNode.Status != api.Running`, so the already-evicted task (now `Releasing`) is excluded. This causes `reclaimees == []` on that node, and the node is skipped entirely — even though `FutureIdle = Idle + Releasing` is sufficient.

```go
// reclaim.go:188-207 — the problematic filter + early-exit
for _, taskOnNode := range n.Tasks {
    if taskOnNode.Status != api.Running || !taskOnNode.Preemptable {
        continue  // Releasing tasks are excluded here
    }
    ...
}

if len(reclaimees) == 0 {
    klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
    continue  // Node skipped even if FutureIdle is sufficient!
}
```

**Cluster setup:**
- 3 worker nodes, each with 4 GPUs (12 GPUs total)
- `k8s-master` has a taint (not schedulable for user workloads)

**Queue configuration:**
```yaml
# prod-queue: weight=4, reclaimable=true
kind: Queue
metadata:
  name: prod-queue
spec:
  weight: 4
  reclaimable: true
---
# dev-queue: weight=1, reclaimable=true
kind: Queue
metadata:
  name: dev-queue
spec:
  weight: 1
  reclaimable: true
```

**Step 1 — Submit a greedy job that occupies most GPUs:**
```yaml
kind: Job
metadata:
  name: dev-greedy-job
spec:
  minAvailable: 1
  schedulerName: volcano
  queue: dev-queue
  tasks:
    - replicas: 3
      name: greedy
      template:
        spec:
          containers:
            - name: greedy
              image: busybox:latest
              command: ["sleep", "3600"]
              resources:
                limits:
                  nvidia.com/gpu: 3   # 1 task per node, 3 GPU each → 9/12 GPUs used
          restartPolicy: Never
```

After this job runs, each node has: `idle=1 GPU`, `used=3 GPU` (by `dev-greedy-job`).

**Step 2 — Submit a prod job that requires reclaim:**
```yaml
kind: Job
metadata:
  name: prod-job
spec:
  minAvailable: 6
  schedulerName: volcano
  queue: prod-queue
  tasks:
    - replicas: 6
      name: prod-task
      template:
        spec:
          containers:
            - name: prod-task
              image: busybox:latest
              command: ["sleep", "3600"]
              resources:
                limits:
                  nvidia.com/gpu: 1
          restartPolicy: Never
```

**Observed:** All 6 pods remain `Pending` indefinitely.

### Expected Behavior

The reclaim action should successfully pipeline all 6 `prod-tasks` and commit the statement, allowing the gang-scheduled job to start.

**Correct scheduling trace (expected):**

| Task | Node | Action | Reason |
|---|---|---|---|
| prod-task-0 | node1 | pipeline (no eviction) | node1 idle=1 GPU ≥ 1 GPU |
| prod-task-1 | node1 | evict greedy-1, pipeline | node1 idle=0, FutureIdle=3 after eviction |
| prod-task-2 | node1 | pipeline (no eviction) | node1 FutureIdle=2 GPU (releasing=3, used=1) |
| prod-task-3 | node2 | pipeline (no eviction) | node2 idle=1 GPU ≥ 1 GPU |
| prod-task-4 | node2 | evict greedy-0, pipeline | node2 idle=0, FutureIdle=3 after eviction |
| prod-task-5 | node2 | pipeline (no eviction) | node2 FutureIdle=2 GPU |

Total evictions: 2 (greedy-1, greedy-0). `dev-queue allocated` = 9-6 = 3 GPU < deserved=6 GPU. `JobPipelined` = true → `stmt.Commit()`.

### Actual Behavior

**Actual scheduling trace:**

| Task | Node | Action | Reason |
|---|---|---|---|
| prod-task-3 | node1 | evict greedy-1, pipeline | node1 idle=0, FutureIdle=4 after eviction |
| prod-task-4 | node1 | **SKIP** (reclaimees=[]) | greedy-1 is now `Releasing`, not `Running` |
| prod-task-4 | node2 | evict greedy-0, pipeline | `DeallocateFunc` fires: dev-queue allocated=9-3=**6=deserved** |
| prod-task-5 | node1 | SKIP (reclaimees=[]) | greedy-1 is `Releasing` |
| prod-task-5 | node2 | SKIP (reclaimees=[]) | greedy-0 is `Releasing` |
| prod-task-5 | node3 | **FAIL** | dev-queue allocated(3) ≤ deserved(6) → greedy-2 not a victim |

After processing all 3 tasks: only 2 pipelined. `WaitingTaskNum(2) < minAvailable(6)` → `JobPipelined=false` → `stmt.Discard()`. This repeats every scheduling cycle.

**Key log evidence:**
```
reclaim.go:232] Try to reclaim Task <default/dev-greedy-job-greedy-1> for Tasks <default/prod-job-prod-task-3>
statement.go:171] After pipelined Task <prod-task-3> to Node <k8s-node1>: idle <nvidia.com/gpu 1000.00>, releasing <nvidia.com/gpu 3000.00>

# prod-task-4 on k8s-node1: reclaimees=[] because greedy-1 is Releasing → node skipped silently (V4 log)
reclaim.go:232] Try to reclaim Task <default/dev-greedy-job-greedy-0> for Tasks <default/prod-job-prod-task-4>

reclaim.go:211] No validated victims on Node <k8s-node3>: not enough resources: requested <nvidia.com/gpu 1000.00>, but future idle <nvidia.com/gpu 0.00>
reclaim.go:135] No preemptor task in job <default/prod-job-...>.
statement.go:366] Discarding operations ...
```

v1.15.0

### Proposed Fix

In `reclaimForTask`, before skipping a node due to `reclaimees == []`, check whether `n.FutureIdle()` already satisfies the task's resource request. If so, pipeline the task directly (with `evictionOccurred=false`), consistent with the behavior in `allocate.go`:

```go
if len(reclaimees) == 0 {
    // Even with no reclaimees, the node may have sufficient FutureIdle
    // (e.g., from a prior eviction in this same session). Pipeline directly
    // without evicting anything, matching allocate.go's behavior.
    if task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
        nodeStmt := framework.NewStatement(ssn)
        if err := nodeStmt.Pipeline(task, n.Name, false); err != nil {
            nodeStmt.Discard()
            continue
        }
        stmt.Merge(nodeStmt)
        return
    }
    klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
    continue
}
```

### Related Commits

Two prior fixes addressed related (but distinct) issues in `reclaimForTask`:

- [`3e738650`](https://github.com/volcano-sh/volcano/commit/3e7386508afceee31b1de997e0bbc241ab5e1d13) — *"fix: reclaimed resource add node futureIdle resource first"*: Fixed `availableResources` initialization to use `n.FutureIdle()` instead of zero, preventing over-eviction when the node already has idle capacity. Does **not** address the `reclaimees==[]` skip.

- [`98f47f5d`](https://github.com/volcano-sh/volcano/commit/98f47f5d39ef3bda04429088699d50dd5b1c958b) — *"fix(reclaim): isolate per-node evictions to prevent spurious victim commits"*: Introduced per-node `nodeStmt` and moved the "resources satisfied?" guard to the top of the victim loop. Does **not** address the `reclaimees==[]` skip.

The current issue is the remaining gap: when `reclaimees==[]` but `FutureIdle` is sufficient, the node should not be skipped.

### Environment

- Volcano version: master @ f865e399 (2026-06-25)  
- Affected file: pkg/scheduler/actions/reclaim/reclaim.go  
- Bug introduced by: cdaaa977 (2025-12-16)

vanshika2720:
## Description

Within a scheduling session, `Statement.Evict()` immediately transitions a task from `Running` to `Releasing`. When `reclaimForTask` later processes another task from the same gang on the same node, the previously evicted task is no longer considered a reclaim candidate because only `Running` and `Preemptable` tasks are collected into `reclaimees`.

When `reclaimees` becomes empty, `reclaimForTask` unconditionally skips the node before checking whether `FutureIdle()` already satisfies the pending task. As a result, resources released by earlier evictions in the same scheduling session are ignored, preventing additional tasks from being pipelined. For gang jobs, this can cause `JobPipelined()` to return false, triggering `stmt.Discard()` and rolling back all previous reclaim operations, resulting in repeated scheduling failures.

This PR avoids skipping nodes prematurely when reclaimees is empty. Instead of introducing a separate pipeline path, it falls through to the existing reclaim logic whenever the task depends on releasing resources (i.e. it fits FutureIdle() but not Idle()). This reuses the existing ValidateVictims → reclaim → pipeline flow, preserves existing behavior for idle resources, and avoids bypassing allocation checks.

A regression test has also been added to verify that subsequent gang tasks can be successfully pipelined using `FutureIdle()` after a prior eviction in the same scheduling session.

Fixes #5520

---

#### What type of PR is this?

/kind bug

* Fixes an early-exit path in `reclaimForTask` that skipped nodes with an empty `reclaimees` list even when `FutureIdle()` already had sufficient resources.
* Reuses `FutureIdle()` resources created by prior evictions within the same scheduling session.
* Prevents repeated reclaim/discard scheduling cycles for affected gang jobs.
* Aligns reclaim behavior with the existing logic in `allocate.go` and the documented behavior of `ValidateVictims`.
* Adds a regression test covering the scenario where a second gang task pipelines via `FutureIdle()` after an earlier eviction.

Fixes #5520

The change only affects the `reclaimees == 0` path. Existing reclaim behavior is unchanged when additional victims are required or when `FutureIdle()` is insufficient. The added regression test reproduces the reported gang scheduling failure and verifies that only a single eviction is needed while the remaining task is pipelined using `FutureIdle()`.

```release-note
Fix reclaim scheduling so nodes with sufficient FutureIdle resources are no longer skipped when no additional reclaim candidates remain, preventing repeated reclaim/discard scheduling cycles and improving gang scheduling success.
```

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -202,6 +202,17 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		}
 
 		if len(reclaimees) == 0 {
+			if task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
+				nodeStmt := framework.NewStatement(ssn)
+				if err := nodeStmt.Pipeline(task, n.Name, false); err != nil {
+					klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>",
+						task.Namespace, task.Name, n.Name)
+					nodeStmt.Discard()
+					continue
+				}
+				stmt.Merge(nodeStmt)
+				return
+			}
 			klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
 			continue
 		}
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -202,19 +202,16 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		}
 
 		if len(reclaimees) == 0 {
-			if task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
-				nodeStmt := framework.NewStatement(ssn)
-				if err := nodeStmt.Pipeline(task, n.Name, false); err != nil {
-					klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>",
-						task.Namespace, task.Name, n.Name)
-					nodeStmt.Discard()
-					continue
-				}
-				stmt.Merge(nodeStmt)
-				return
+			// A task can only be pipelined here without any reclaimees if resources
+			// are actually being released on this node (FutureIdle > Idle), e.g. by
+			// an eviction earlier in this same reclaim pass. If the task already fits
+			// in the node's current Idle resources, it doesn't depend on reclaiming
+			// anything, so leave it for the allocate action to handle instead of
+			// bypassing its AllocatableFn checks here.
+			if task.InitResreq.LessEqual(n.Idle, api.Zero) {
+				klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
+				continue
 			}
-			klog.V(4).Infof("No reclaimees on Node <%s>.", n.Name)
-			continue
 		}
 
 		victims := ssn.Reclaimable(task, reclaimees)
@@ -256,8 +253,8 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		}
 
 		if err := nodeStmt.Pipeline(task, n.Name, evictionOccurred); err != nil {
-			klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>",
-				task.Namespace, task.Name, n.Name)
+			klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>: %v",
+				task.Namespace, task.Name, n.Name, err)
 			nodeStmt.Discard()
 			continue
 		}
--- a/pkg/scheduler/actions/reclaim/reclaim_test.go
+++ b/pkg/scheduler/actions/reclaim/reclaim_test.go
@@ -292,6 +292,57 @@ func TestReclaim(t *testing.T) {
 			ExpectEvicted:  []string{"c1/victim-pod"},
 		},
 		{
+			// Regression: reclaimForTask skipped nodes with sufficient FutureIdle when
+			// reclaimees was empty, causing unnecessary gang scheduling failures.
+			//
+			// Within a single scheduling session, Statement.Evict() transitions a task
+			// from Running to Releasing immediately. On the next reclaimForTask call
+			// (for a second pending gang task), the same node's already-evicted task
+			// is Releasing and therefore excluded from reclaimees. The unconditional
+			// "continue" on empty reclaimees prevented reaching the FutureIdle check,
+			// so the task was never pipelined and stmt.Commit() was never called.
+			//
+			// Setup:
+			//   • n1 has 2 CPU total, fully used by victim (2 CPU, Preemptable).
+			//   • pg-preemptor (minMember=2) needs two 1-CPU tasks pipelined.
+			//   • First reclaimForTask: evicts victim (now Releasing), pipelines task1.
+			//     FutureIdle = 0 idle + 2 releasing - 1 pipelined = 1 CPU.
+			//   • Second reclaimForTask: reclaimees=[] (victim is Releasing, not Running),
+			//     but FutureIdle (1 CPU) >= task2 (1 CPU) — must pipeline without evicting.
+			//
+			// Without the fix: second task skips node, JobPipelined returns false,
+			// stmt.Discard() rolls back the eviction — 0 evictions committed.
+			// With the fix: second task pipelines via FutureIdle, JobPipelined=true,
+			// stmt.Commit() — exactly 1 eviction committed.
+			Name: "pipeline second task via FutureIdle when reclaimees empty after prior eviction",
+			Plugins: map[string]framework.PluginBuilder{
+				conformance.PluginName: conformance.New,
+				gang.PluginName:        gang.New,
+				proportion.PluginName:  proportion.New,
+			},
+			PodGroups: []*schedulingv1beta1.PodGroup{
+				util.BuildPodGroupWithPrio("pg-victim", "c1", "q1", 0, nil, schedulingv1beta1.PodGroupRunning, "low-priority"),
+				util.BuildPodGroupWithPrio("pg-preemptor", "c1", "q2", 2, nil, schedulingv1beta1.PodGroupInqueue, "low-priority"),
+			},
+			Pods: []*v1.Pod{
+				// victim occupies all 2 CPU on n1; both preemptor tasks need 1 CPU each.
+				util.BuildPod("c1", "victim", "n1", v1.PodRunning, api.BuildResourceList("2", "2G"), "pg-victim", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				util.BuildPod("c1", "preemptor-task1", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg-preemptor", make(map[string]string), make(map[string]string)),
+				util.BuildPod("c1", "preemptor-task2", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg-preemptor", make(map[string]string), make(map[string]string)),
+			},
+			Nodes: []*v1.Node{
+				util.BuildNode("n1", api.BuildResourceList("2", "2G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+			},
+			Queues: []*schedulingv1beta1.Queue{
+				// q1 deserves 1 CPU (weight 1 of 10), uses 2 CPU -> overused -> victim is reclaimable.
+				util.BuildQueue("q1", 1, nil),
+				// q2 deserves 9 CPU (weight 9 of 10), uses 0 -> starving -> preemptor reclaims.
+				util.BuildQueue("q2", 9, nil),
+			},
+			ExpectEvictNum: 1,
+			ExpectEvicted:  []string{"c1/victim"},
+		},
+		{
 			// Regression: the original victim loop placed the "resources satisfied"
 			// check AFTER each eviction rather than before it.  This caused at least
 			// one victim to be spuriously evicted even when the node's FutureIdle was
--- a/pkg/scheduler/actions/reclaim/reclaim_test.go
+++ b/pkg/scheduler/actions/reclaim/reclaim_test.go
@@ -292,28 +292,16 @@ func TestReclaim(t *testing.T) {
 			ExpectEvicted:  []string{"c1/victim-pod"},
 		},
 		{
-			// Regression: reclaimForTask skipped nodes with sufficient FutureIdle when
-			// reclaimees was empty, causing unnecessary gang scheduling failures.
+			// Regression (#5520): a gang's second task saw reclaimees=[] on n1 because
+			// the first task's eviction had already moved the victim to Releasing. The
+			// unconditional "continue" on empty reclaimees stopped before reaching the
+			// FutureIdle check, so the second task never pipelined, JobPipelined stayed
+			// false, and stmt.Discard() rolled back the first eviction, leaving the gang
+			// Pending with reclaim repeating the same evict/discard cycle each session.
 			//
-			// Within a single scheduling session, Statement.Evict() transitions a task
-			// from Running to Releasing immediately. On the next reclaimForTask call
-			// (for a second pending gang task), the same node's already-evicted task
-			// is Releasing and therefore excluded from reclaimees. The unconditional
-			// "continue" on empty reclaimees prevented reaching the FutureIdle check,
-			// so the task was never pipelined and stmt.Commit() was never called.
-			//
-			// Setup:
-			//   • n1 has 2 CPU total, fully used by victim (2 CPU, Preemptable).
-			//   • pg-preemptor (minMember=2) needs two 1-CPU tasks pipelined.
-			//   • First reclaimForTask: evicts victim (now Releasing), pipelines task1.
-			//     FutureIdle = 0 idle + 2 releasing - 1 pipelined = 1 CPU.
-			//   • Second reclaimForTask: reclaimees=[] (victim is Releasing, not Running),
-			//     but FutureIdle (1 CPU) >= task2 (1 CPU) — must pipeline without evicting.
-			//
-			// Without the fix: second task skips node, JobPipelined returns false,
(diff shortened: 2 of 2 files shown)


vanshika2720:
@hajnalmt @JesseStutler Please review this PR.
@wangyang0616

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -202,6 +202,17 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		}
 
 		if len(reclaimees) == 0 {
+			if task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
```
Aman-Cool: This condition is the problem: on a node with nothing releasing, `FutureIdle()` == `Idle`, so this pipelines tasks onto plain idle capacity — that's allocate's job, and it's what breaks the capacity hierarchy tests (they run reclaim before allocate). Needs `&& !task.InitResreq.LessEqual(n.Idle, api.Zero)` so the path only triggers when the task genuinely depends on releasing resources. I ran the suite with that guard added — capacity tests and your new regression test both pass. And rather than this whole block, consider just inverting the guard and falling through: the existing victims-loop path below already handles empty victims correctly end to end (ValidateVictims even documents it), which would also get you the error in the pipeline-failure log for free — as written this `klog.Errorf` drops the `err` it's reporting on, and the `return` here quietly diverges from the `break` the sibling path uses.

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -291,6 +291,57 @@ func TestReclaim(t *testing.T) {
...
+				// victim occupies all 2 CPU on n1; both preemptor tasks need 1 CPU each.
+				util.BuildPod("c1", "victim", "n1", v1.PodRunning, api.BuildResourceList("2", "2G"), "pg-victim", map[string]string{schedulingv1beta1.PodPreemptable: "true"}, make(map[string]string)),
+				util.BuildPod("c1", "preemptor-task1", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg-preemptor", make(map[string]string), make(map[string]string)),
+				util.BuildPod("c1", "preemptor-task2", "", v1.PodPending, api.BuildResourceList("1", "1G"), "pg-preemptor", make(map[string]string), make(map[string]string)),
+			},
+			Nodes: []*v1.Node{
+				util.BuildNode("n1", api.BuildResourceList("2", "2G", []api.ScalarResource{{Name: "pods", Value: "10"}}...), make(map[string]string)),
+			},
+			Queues: []*schedulingv1beta1.Queue{
+				// q1 deserves 1 CPU (weight 1 of 10), uses 2 CPU -> overused -> victim is reclaimable.
```
Aman-Cool: The math in this comment doesn't work out — the cluster has 2 CPU total, so with weights 1:9 proportion gives q1 0.2 CPU and q2 1.8 CPU deserved, not "1 CPU" and "9 CPU". The test still passes because `Preemptive` checks each task individually against deserved (pipelined tasks aren't added to `attr.allocated`), but anyone using this comment to understand the scenario will be misled. Could you fix the numbers, and generally trim the narrative block above? Twenty-plus lines is a lot for one table case — the existing regression tests in this file get it done in about five.

Aman-Cool (commented):
@vanshika2720, First off — the underlying bug is real. I verified the whole premise from #5520: reclaimees only collects `Running` preemptable tasks, `Statement.Evict` flips the victim to `Releasing` immediately, so the next gang task in the same session finds an empty list and bails before ever reaching a FutureIdle check. The comment inside `ValidateVictims` documents this exact scenario as supposed-to-work, so the early `continue` was defeating a contract that already existed. The issue's GPU trace is consistent with all of that.

But I have to hold this, because the CI failure is not flake.., it's caused by this change, and I can show it. I checked out master, ran `Test_capacityPlugin_OnSessionOpenWithHierarchy` (passes), applied this diff, ran again... fails exactly the way CI does, cases 2, 8 and 9. The mechanism: that test runs actions as `enqueue, reclaim, allocate`, so reclaim sees pending gang tasks before allocate does; and on a node with nothing releasing, `FutureIdle()` is just `Idle`, so the new path pipelines both tasks onto a plain idle node. Allocate only picks up Pending tasks, so nothing ever binds. Since the actions list is user-configurable we can't dismiss that ordering, and even in the normal allocate-first order there's a quieter version: a task that allocate declined because of an `AllocatableFn` gate (an ancestor capability cap in the capacity plugin, say) can still pass reclaim's weaker `Preemptive` check and get pipelined onto idle resources, side-stepping the reason allocate said no.

I noticed the code here is the proposed-fix block from #5520 taken as-is and the flaw is in that proposal, not in your transcription. The issue says it's "consistent with allocate.go", but allocate only falls back to FutureIdle pipelining *after* trying and failing to place the task on idle resources; lifting the FutureIdle check without that context is what opens the hole. The fix I'd suggest is one condition: only take this path when the task fits `FutureIdle()` but does **not** fit `n.Idle` — i.e. when it genuinely depends on releasing capacity. I ran your branch with exactly that guard added: the three capacity cases go green and your regression test still passes. It also still fully covers the issue's scenario — in their trace, allocate has already consumed all idle by the time reclaim runs, so every relevant node is `idle=0, releasing=3` and the guard is satisfied. (Nice side effect: their cluster then needs one eviction, not two.)

On the shape of the change, I'd rather not grow a second hand-rolled pipeline/merge block in this function. If the empty-reclaimees check just guards-and-falls-through instead of continuing, the existing machinery below does everything — `ssn.Reclaimable` on an empty list returns nothing, `ValidateVictims` applies its documented empty-victims rule, the victim loop no-ops, `evictionOccurred` stays false naturally, and the existing Pipeline/Merge/Discard path runs with its logging intact. That's also exactly how preempt is structured (it never had this early-continue), so the two actions would converge instead of drifting further apart. Fewer lines than the current diff, one pipeline path to maintain.

Two asks before this lands. First: did you get a chance to reproduce #5520's scenario yourself and run your patched build against it, or was validation the unit test? The issue author's logs establish the bug, but since the fix came from the issue text rather than from debugging, I'd feel much better seeing scheduler logs from a patched run of that repro — evict greedy-1, then the second and third prod-tasks pipelining via FutureIdle in the same session, ending in `Commit` instead of `Discarding operations`. Second, small but worth doing: the issue title and this PR's release note say the bug caused "unnecessary evictions"; it can't have. Evictions are deferred to `Commit`, and in the failure scenario the statement is `Discard`ed, so no pod is ever actually evicted; the user-visible symptom is a gang job stuck Pending forever while the scheduler repeats the evict-fail-discard dance every session (which is what the issue's own log excerpt shows). Could you fix the release note so we don't document a failure mode that doesn't exist?

Happy to re-review quickly once the guard is in... the fix underneath is worth landing.

vanshika2720:
@Aman-Cool I updated the implementation based on your suggestions.

Changes made:

Added the additional guard so the FutureIdle path is only taken when the task fits FutureIdle() but does not already fit Idle(), preventing reclaim from bypassing normal allocation on plain idle resources.
Removed the duplicate pipeline path and reused the existing reclaim flow instead, so empty reclaimees now falls through into the existing ValidateVictims → reclaim → pipeline logic.
Fixed the dropped pipeline error logging.
Corrected the CPU math in the regression test comments and shortened the explanatory block.

I also re-ran the regression test together with the capacity hierarchy tests. The new regression test still passes, and the previously failing capacity plugin cases now pass as well.

I wasn't able to reproduce the original cluster scenario directly. Validation was done through the regression test together with the capacity hierarchy tests. I also verified the behavioral difference by reverting the fix and confirming the capacity plugin regressions reappeared, then restoring the fix and confirming both the regression test and the capacity tests passed.

Aman-Cool:
@vanshika2720, Thanks for turning this around so quickly... and for the revert/restore check, that's exactly the kind of discipline I like to see. I pulled the updated branch onto current master myself and ran the capacity, reclaim and preempt suites: all green, matching what you reported. Code-wise I'm happy now; the fall-through shape reads like it was always meant to be there, and leaving the `FutureIdle` judgment to `ValidateVictims` means reclaim and preempt finally agree on what an empty victim list means.

One thing I want to pin down before this merges, though. When you say you weren't able to reproduce the original cluster scenario.., did you try and the bug didn't show up, or were you blocked from setting it up in the first place? It matters more than it might seem: if you ran the scenario and couldn't trigger the bug, that's a signal the diagnosis in #5520 is incomplete somewhere, and I'd want to understand that before we ship a fix for it. If it was a setup problem, I suspect the two usual culprits: the repro as written asks for `nvidia.com/gpu`, which you don't actually need; the whole scenario translates directly to CPU requests (say, 3 kind worker nodes, the greedy job requesting most of each node's CPU, prod job reclaiming); and reclaim isn't in the default action list, so nothing happens unless the scheduler config explicitly includes it, something like `actions: "enqueue, allocate, reclaim, backfill"`. With those two adjustments a kind cluster reproduces this class of bug fine.

If you get a chance to run that.., unpatched scheduler showing the repeating `Discarding operations ...` with the job stuck `Pending`, then your build showing the eviction commit and the gang coming up... paste the two log snippets here and I'll happily add my lgtm on the spot. Alternatively, @yanqianling, since you have the original environment: could you run this patched build against your workload and confirm it unsticks the prod job? Either one closes the loop for me.

@hajnalmt WDYT on this???

vanshika2720:
@JesseStutler @hajnalmt
@hwdef

PR #5523, opened 2026-06-27
fix(scheduler): stamp job priority onto snapshot clone, not the original

### LINKED ISSUE #5522: [Bug] SchedulerCache Snapshot() mutates cached JobInfo while constructing scheduling snapshots
### Description

What happened?

SchedulerCache.Snapshot() is expected to build an isolated scheduling snapshot from the scheduler cache without modifying the underlying cache state.

However, during snapshot construction, it updates the Priority field on the cached JobInfo before cloning it.

Current flow:

value.Priority = resolvedPriority
clone := value.Clone()

Instead of applying the resolved priority to the cloned object, the cached JobInfo stored in SchedulerCache is modified every time a scheduling session creates a snapshot.

This introduces an unintended side effect in what should be a read-only snapshot operation.

Why is this a problem?

Snapshot() should not mutate objects owned by the scheduler cache.

Although the assigned priority is deterministic (PriorityClass.Value or defaultPriority), mutating the cached object has several drawbacks:

breaks the expectation that snapshot generation is read-only;
introduces hidden side effects into scheduler cache state;
couples cache mutation with snapshot creation;
makes future snapshot refactoring (incremental snapshots, reduced lock scope, copy-on-write designs) more difficult.
Root Cause

Inside SchedulerCache.Snapshot(), the cached JobInfo is modified before cloning:

value.Priority = resolvedPriority
clone := value.Clone()

As a result, the mutation is applied to the original object stored in SchedulerCache, not just to the scheduling snapshot.
Proposed Solution

Clone the JobInfo first and apply the resolved priority only to the cloned object.

Conceptually:

clone := value.Clone()
clone.Priority = resolvedPriority

The priority lookup (defaultPriority and PriorityClasses) can be captured before spawning snapshot goroutines so the clone remains independent of the live cache.

Benefits
Makes Snapshot() side-effect free.
Preserves scheduler cache immutability during snapshot construction.
Simplifies future snapshot optimizations.
Does not change scheduling behavior.
Does not introduce API changes.
Validation

The proposed fix can be validated with unit tests that verify:

snapshot jobs receive the expected priority;
cached JobInfo objects remain unchanged after Snapshot();
the implementation passes the Go race detector (go test -race).
Additional Context

While investigating SchedulerCache lock contention (#5036), I initially explored reducing the lock scope around snapshot creation.

After tracing the ownership of NodeInfo, JobInfo, and QueueInfo, moving deep-copy operations outside the cache lock appears unsafe because informer callbacks can concurrently modify these objects.

However, this investigation revealed the unintended mutation of cached JobInfo instances during snapshot creation, which can be fixed independently without changing the existing locking semantics.

Create a SchedulerCache containing a JobInfo with Priority == 0.
Configure a matching PriorityClass (or use defaultPriority).
Invoke SchedulerCache.Snapshot().
Inspect both the returned snapshot and the original cached JobInfo.

Current behavior

The snapshot contains the expected priority.
The original cached JobInfo is also modified.

Expected behavior

The snapshot should contain the resolved priority.
The cached JobInfo stored in SchedulerCache should remain unchanged after Snapshot() completes.

master

_No response_

vanshika2720:
Snapshot() currently writes `value.Priority` directly to the live `JobInfo` objects stored in `sc.Jobs` before cloning them. As a result, every scheduling cycle mutates the scheduler cache, even though `Snapshot()` is conceptually a read-only operation. This makes the live cache state depend on whether `Snapshot()` has been executed.

This PR fixes the issue by:

* capturing `sc.defaultPriority` and a shallow copy of `sc.PriorityClasses` under the cache mutex before spawning the job-cloning goroutines;
* cloning each `JobInfo` first via `value.Clone()`;
* applying the resolved priority to the cloned object instead of the cached one.

The existing synchronization semantics remain unchanged. The cloning goroutines still execute while `sc.Mutex` is held (`wg.Wait()` remains inside the critical section), so this change does not introduce any new concurrency risks.

Tests added:

* `TestSnapshot_JobPriorityAppliedToClone` verifies that the snapshot receives the resolved `PriorityClass` value while the original cached `JobInfo` remains unchanged.
* `TestSnapshot_DefaultPriorityUsedWhenNoClass` verifies fallback to `sc.defaultPriority` when no matching `PriorityClass` exists, while ensuring the cached object is not modified.

#### What type of PR is this?

/kind bug

This PR removes an unintended side effect in `SchedulerCache.Snapshot()` by ensuring priority resolution only affects the snapshot copy. It preserves cache immutability during snapshot generation without changing scheduling behavior or locking semantics.

Fixes #5522 

* This PR intentionally does **not** modify the existing lock scope or snapshot concurrency model.
* The change is limited to preventing mutation of cached `JobInfo` objects during snapshot creation.
* Existing scheduling behavior remains unchanged.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1532,27 +1532,36 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 		snapshot.Queues[value.UID] = value.Clone()
 	}
 
+	// Capture priority data under the lock so that each cloneJob goroutine
+	// reads from local variables instead of the shared cache fields, and so
+	// that priority is stamped onto the cloned snapshot object rather than
+	// mutating the original JobInfo stored in the cache.
+	defaultPriority := sc.defaultPriority
+	priorityClassValues := make(map[string]int32, len(sc.PriorityClasses))
+	for name, pc := range sc.PriorityClasses {
+		priorityClassValues[name] = pc.Value
+	}
+
 	var cloneJobLock sync.Mutex
 	var wg sync.WaitGroup
 
 	cloneJob := func(value *schedulingapi.JobInfo) {
 		defer wg.Done()
-		if value.PodGroup != nil {
-			value.Priority = sc.defaultPriority
+		clonedJob := value.Clone()
+		if clonedJob.PodGroup != nil {
+			clonedJob.Priority = defaultPriority
 
-			priName := value.PodGroup.Spec.PriorityClassName
-			if priorityClass, found := sc.PriorityClasses[priName]; found {
-				value.Priority = priorityClass.Value
+			priName := clonedJob.PodGroup.Spec.PriorityClassName
+			if pv, found := priorityClassValues[priName]; found {
+				clonedJob.Priority = pv
 			}
 
 			klog.V(4).Infof("The priority of job <%s/%s> is <%s/%d>",
-				value.Namespace, value.Name, priName, value.Priority)
+				clonedJob.Namespace, clonedJob.Name, priName, clonedJob.Priority)
 		}
 
-		clonedJob := value.Clone()
-
 		cloneJobLock.Lock()
-		snapshot.Jobs[value.UID] = clonedJob
+		snapshot.Jobs[clonedJob.UID] = clonedJob
 		cloneJobLock.Unlock()
 	}
 
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1532,16 +1532,6 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 		snapshot.Queues[value.UID] = value.Clone()
 	}
 
-	// Capture priority data under the lock so that each cloneJob goroutine
-	// reads from local variables instead of the shared cache fields, and so
-	// that priority is stamped onto the cloned snapshot object rather than
-	// mutating the original JobInfo stored in the cache.
-	defaultPriority := sc.defaultPriority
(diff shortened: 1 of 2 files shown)


vanshika2720:
@hajnalmt @JesseStutler Please review this PR.

vanshika2720:
@hajnalmt @JesseStutler @hwdef Please review this PR.

pkg/scheduler/cache/cache.go
```diff
@@ -1532,27 +1532,36 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
...
 
+	// Capture priority data under the lock so that each cloneJob goroutine
+	// reads from local variables instead of the shared cache fields, and so
+	// that priority is stamped onto the cloned snapshot object rather than
+	// mutating the original JobInfo stored in the cache.
+	defaultPriority := sc.defaultPriority
+	priorityClassValues := make(map[string]int32, len(sc.PriorityClasses))
+	for name, pc := range sc.PriorityClasses {
+		priorityClassValues[name] = pc.Value
+	}
```
hwdef: This is not necessary; we only need a core fix.

vanshika2720: @hwdef I've removed the unnecessary priority pre-copy block and kept the change focused on the core fix. Snapshot() already holds sc.Mutex for its entire execution (including the clone goroutines via wg.Wait()), so reading sc.defaultPriority and sc.PriorityClasses directly is already race-free. The remaining change only clones the JobInfo first and applies the resolved priority to the clone, leaving the cached object unchanged.

vanshika2720:
@hajnalmt @JesseStutler

PR #5526, opened 2026-06-28
webhooks: validate root child queue resources

avinxshKD:
#### What type of PR is this?

/kind bug

Queue admission skipped resource validation for direct children of the `root` queue.

This makes root children follow the same hierarchy checks as other child queues when the root queue exists, so their capability/deserved/guarantee values are validated against root limits.

Fixes #

Added a regression test for root child capability and guarantee validation.

```release-note
NONE

### DIFF
--- a/pkg/webhooks/admission/queues/validate/validate_queue.go
+++ b/pkg/webhooks/admission/queues/validate/validate_queue.go
@@ -329,44 +329,28 @@ func validateQueueDeleting(queueName string) error {
 	return nil
 }
 
-// needsValidateHierarchicalQueue determines if hierarchy resource validation is necessary
-// Returns true only if:
-// - Queue is not the root queue itself AND
-// - Either it's a CREATE operation OR resources (capability/deserved/guarantee) changed on UPDATE OR parent changed on UPDATE
+// needsValidateHierarchicalQueue determines if hierarchy resource validation is necessary.
 func needsValidateHierarchicalQueue(queue, oldQueue *schedulingv1beta1.Queue, operation admissionv1.Operation) bool {
-	// Only the root queue itself should skip validation completely
-	if queue.Name == "root" {
-		return false
-	}
-
-	// For CREATE operations, always validate
 	if operation == admissionv1.Create {
-		return true
+		return queue.Name != "root"
 	}
 
-	// For UPDATE operations, only validate if resources or parent changed
 	if operation == admissionv1.Update && oldQueue != nil {
-		// Check if parent changed
 		if queue.Spec.Parent != oldQueue.Spec.Parent {
 			return true
 		}
-		// Check if capability changed
 		if !equality.Semantic.DeepEqual(queue.Spec.Capability, oldQueue.Spec.Capability) {
 			return true
 		}
-		// Check if deserved changed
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by avinxshKD** [comment]
@hwdef @JesseStutler Pls take a look when get chance.

**Comment by hwdef** [comment]
Please consider the case where the `spec.parent == ""`.

**Comment by avinxshKD** [comment]
Hey @hwdef, updated the validation to treat an empty parent as root for non-root queues, and adjusted the parent index lookup so root child sums include queues with spec.parent unset as well. take a look when get chance.

**Comment by avinxshKD** [comment]
ptal @JesseStutler @hwdef when get chance

File: pkg/webhooks/admission/queues/validate/validate_queue.go [hunk 1]
```diff
@@ -441,7 +441,10 @@ func getSingleResource(r *api.Resource, name v1.ResourceName) float64 {
 // Recursively searches for the ancestor queue that recently defines the capability
 func findNearestAncestorCapability(q *schedulingv1beta1.Queue, rname v1.ResourceName) (float64, bool) {
 	parent := q.Spec.Parent
-	for parent != "" && parent != "root" {
+	if parent == "" && q.Name != "root" {
+		parent = "root"
+	}
+	for parent != "" {
```
hzxuzhonghu, line 432: Normalizing only the initial queue leaves two bad termination cases. If an intermediate top-level queue has an empty parent, this walk stops before checking root, so descendants can exceed root capability. If `root.spec.parent` is `root` while root protection is disabled, the loop never terminates. Normalize each ancestor's empty parent and stop at root or track visited queues.

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -529,28 +532,34 @@ func validateQueueDepth(queue *schedulingv1beta1.Queue) error {
...
 // - Sum of children's guarantee/deserved <= parent's limit
 func validateHierarchicalQueueResources(queue *schedulingv1beta1.Queue) error {
+	parentName := queue.Spec.Parent
+	if parentName == "" && queue.Name != "root" {
+		parentName = "root"
+	}
+
 	// If this is a child queue, validate against parent
-	if queue.Spec.Parent != "" && queue.Spec.Parent != "root" {
+	if parentName != "" {
```
hzxuzhonghu, line 534: This adds root checks when a child changes, but `needsValidateHierarchicalQueue` still returns false for `queue.Name == "root"`. With root protection disabled, lowering root capability below an existing child is therefore admitted. Root resource updates need to run the child constraints too.

hzxuzhonghu (changes requested):
Root traversal still has correctness and termination gaps.

pkg/webhooks/admission/queues/validate/validate_queue.go
avinxshKD: fixed. the traversal now normalizes empty parents at each step and tracks visited queues so it won’t stop early or loop on root

pkg/webhooks/admission/queues/validate/validate_queue.go
avinxshKD: done this also, root updates that change resource fields now run the hierarchy resource validation too, so lowering root limits is checked against existing children

PR #5529, opened 2026-06-28
feat(scheduler): add timeout-based inqueue downgrade for stale jobs

### LINKED ISSUE #5330: [Feature] support high-priority Pending tasks to preempt low-priority Inqueue tasks within the same queue
### 1.1 Problem Statement

In the Volcano scheduler, tasks transition from Pending to Inqueue state through the Enqueue Action. The enqueue check evaluates queue quota (Queue Capability) using the formula:

```
(realCapability - inqueue - allocated + elastic) >= job.MinResources
```

where **inqueue** represents the sum of resource requests from all enqueued but not yet scheduled tasks.

The following problems exist:

1. **Inqueue tasks are not guaranteed to be scheduled**: After enqueuing, tasks may fail in the Allocate Action due to insufficient node resources, PodAffinity constraints, or other reasons.
2. **Discard does not release queue quota**: When Allocate fails, `Statement.Discard()` only rolls back task-level resource allocation (unallocate), without releasing the inqueue quota consumed.
3. **Inqueue state is not automatically downgraded**: When a Session closes, `jobStatus()` only preserves the Inqueue state when `allocated=0` and the phase is not Inqueue. It will not revert Inqueue back to Pending.

These three issues mean that: **an Inqueue task that cannot be scheduled continues to occupy queue quota, blocking subsequent tasks from enqueuing**.

### 1.2 Objective

When queue quota is insufficient for new Pending tasks, allow high-priority Pending tasks to preempt low-priority Inqueue tasks within the same queue, releasing quota so the preemptor can enqueue.

### 2.1 Approach

Add a second phase (Phase 2: Inqueue Preemption) to the Enqueue Action, executed after the Phase 1 normal enqueue loop:

```
Phase 1: Normal enqueue loop (existing logic unchanged)
    ↓
Phase 2: Inqueue Preemption (new)
    - Process remaining Pending tasks in queue-priority then job-priority order
    - Reuse Phase 1 priority queues; no additional sorting needed
    - For each Pending task, find lower-priority Inqueue candidates in the same queue
    - Select victims by priority first, then enqueue duration (lower priority, longer enqueued = higher priority to preempt)
    - Execute preemption: release victim's inqueue quota → revert victim to Pending → enqueue preemptor
```

### 2.2 Trigger Condition

The queue's remaining quota cannot satisfy a Pending task's MinResources requirement:
- After Phase 1's normal enqueue loop, there are still Pending tasks that failed to enqueue.

### 2.3 Flow Diagram

```
┌─────────────────────────────────────────┐
│         Phase 1: Normal Enqueue         │
│  Iterate queues by priority, enqueue    │
│  jobs one by one. Failed jobs stay in   │
│  jobsMap.                               │
└────────────────┬────────────────────────┘
                 │
                 ▼
┌─────────────────────────────────────────┐
│         Phase 2: Inqueue Preemption     │
│                                         │
│  Process pending jobs in queue-priority │
│  then job-priority order, reusing the   │
│  priority queues from Phase 1.          │
└────────────────┬────────────────────────┘
                 │
                 ▼
┌─────────────────────────────────────────┐
│  For each pending job (ordered):        │
│                                         │
│  1. Find Inqueue candidates in same     │
│     queue                               │
│  2. Filter: strictly lower priority     │
│  3. selectVictims: greedy by priority   │
│     first, then enqueue duration        │
│  4. Execute preemption                  │
│  5. Enqueue preemptor                   │
└─────────────────────────────────────────┘
```

### 3.1 Preemption Candidate Conditions

An Inqueue task can be preempted if all of the following hold:

| Condition | Description |
|-----------|-------------|
| Same queue | `victim.Queue == preemptor.Queue` |
| Inqueue state | `victim.PodGroup.Status.Phase == PodGroupInqueue` |
| Lower priority | `victim.Priority < preemptor.Priority` (strictly less) |

**No protection period is designed**: Inqueue tasks are processed continuously within the same Session. There is no cross-session time window, so a protection period mechanism is unnecessary.

### 3.2 Preemptor Processing Order

Preemptors are processed in a two-level ordering:

1. **Queue priority**: Reuses the queue priority queue built in Phase 1 (sorted by `QueueOrderFn`). Pending tasks in higher-priority queues get preemption opportunities first.
2. **Job priority**: Within the same queue, reuses the Job priority queue built in Phase 1 (sorted by `JobOrderFn`). Higher-priority Jobs preempt first.

This design avoids the overhead of collecting Pending Jobs into a flat slice and re-sorting. The existing priority queue data structures from Phase 1 are reused directly.

### 3.3 Victim Selection Algorithm

`selectVictims(preemptor, candidates)` uses a greedy strategy with two-level sorting:

1. **Primary sort — Priority**: Ascending by Job priority. Lower priority jobs are more likely to be preempted.
2. **Secondary sort — Enqueue duration**: Within the same priority, descending by enqueue duration (`now - EnqueueTimestamp`). Longer enqueued jobs are more likely to be preempted.
3. **Greedy selection**: Select victims in order, accumulating their MinResources until `victims.MinResources >= preemptor.MinResources`.
4. **Resource constraint**: If selecting all candidates still cannot satisfy the preemptor's resource requirements, return empty (no preemption).

**Rationale**: Priority is the core dimension of scheduling fairness — lower-priority tasks should be sacrificed first. Within the same priority, tasks enqueued longer are harder to schedule (repeatedly failed to allocate across rounds), so evicting them first minimizes impact on overall scheduling efficiency.

### 3.4 Quota Release and Updates

Both the Proportion plugin and the Capacity plugin register two callbacks to keep inqueue quota tracking consistent during preemption:

| Callback | Trigger | Operation |
|----------|---------|------------|
| `AddJobPreemptedFn` | Job preempted from inqueue | `attr.inqueue.Sub(job.GetMinResources())` |
| `AddJobEnqueuedFn` | Job enqueued | `attr.inqueue.Add(job.GetMinResources())` |

**Note**: `JobEnqueueableFn` no longer calls `attr.inqueue.Add()`; inqueue management is handled uniformly by `JobEnqueuedFn`, avoiding double counting. This pattern is consistent across both the Proportion and Capacity plugins.

### 3.5 Preemption Execution Flow

```
for each victim in victims:
    ssn.JobPreempted(victim)           // Notify plugin (release inqueue quota)
    victim.Phase = Pending             // Revert state
    ssn.Jobs[victim.UID] = victim      // Update Session

ssn.JobEnqueued(preemptor)             // Notify plugin (add inqueue quota)
preemptor.Phase = Inqueue              // Set Inqueue state
preemptor.EnqueueTimestamp = Now()     // Record enqueue time
ssn.Jobs[preemptor.UID] = preemptor    // Update Session
```

### 4.1 JobInfo New Field

**File**: `pkg/scheduler/api/job_info.go`

```go
type JobInfo struct {
    // ... existing fields ...
    ScheduleStartTimestamp metav1.Time
    // EnqueueTimestamp records when the job entered Inqueue phase.
    EnqueueTimestamp metav1.Time
    Preemptable bool
    // ...
}
```

**Purpose**: Records when a job entered the Inqueue phase, used for victim scoring.

### 4.2 Session New Field

**File**: `pkg/scheduler/framework/session.go`

```go
type Session struct {
    // ... existing fields ...
    jobPreemptedFns   map[string]api.JobPreemptedFn
    // ...
}
```

**File**: `pkg/scheduler/api/types.go`

```go
// JobPreemptedFn is the func declaration used to call after a job is preempted from inqueue.
type JobPreemptedFn func(interface{})
```

### 4.3 scoredCandidate Struct

**File**: `pkg/scheduler/actions/enqueue/enqueue.go`

```go
type scoredCandidate struct {
    job         *api.JobInfo   // Candidate job
    priority    int32          // Job priority (primary sort key)
    enqueueTime time.Time      // Enqueue time (secondary sort key)
}
```

Sorting uses `sort.Slice` with the comparison rule: ascending by priority (lower priority preempted first), and within the same priority, descending by enqueue time (longer enqueued preempted first).

## 5. Code Change Summary

| File | Change |
|------|--------|
| `pkg/scheduler/api/job_info.go` | Add `EnqueueTimestamp` field |
| `pkg/scheduler/api/types.go` | Add `JobPreemptedFn` type definition |
| `pkg/scheduler/framework/session.go` | Add `jobPreemptedFns` map to Session; initialize in `openSession()` |
| `pkg/scheduler/framework/session_plugins.go` | Add `AddJobPreemptedFn()` registration method and `JobPreempted()` invocation method |
| `pkg/scheduler/actions/enqueue/enqueue.go` | Record `EnqueueTimestamp` in Phase 1; add `tryPreemptForPendingJobs()`, `selectVictims()`, `sortScoredCandidates()`; Phase 2 reuses Phase 1 priority queues for queue-priority → job-priority ordering; use `sort.Slice` for candidate sorting |
| `pkg/scheduler/plugins/proportion/proportion.go` | Register `AddJobPreemptedFn` (release inqueue), register `AddJobEnqueuedFn` (add inqueue), remove `inqueue.Add()` from `JobEnqueueableFn` |
| `pkg/scheduler/plugins/capacity/capacity.go` | Register `AddJobPreemptedFn` (release inqueue), register `AddJobEnqueuedFn` (add inqueue), remove `inqueue.Add()` from `JobEnqueueableFn` |

## 6. Key Parameters Summary

| Parameter | Value | Description |
|-----------|-------|-------------|
| Trigger condition | Queue remaining quota < Pending task requirement | Tasks remain unenqueued after Phase 1 |
| Candidate scope | Same queue, Inqueue state, lower priority | Strict priority comparison |
| Processing order | Queue priority → Job priority | Reuses Phase 1 priority queues |
| Victim selection | Lower priority → longer enqueue time → higher preemption priority | Two-level sort: ascending by priority, then descending by duration |
| Resource constraint | Sum of victims' MinResources >= preemptor's MinResources | Otherwise no preemption |
| Protection period | Not needed | Continuous processing within same Session |
| Preemption count tracking | Not needed | Not implemented |
| SLA wait time | Not needed | Not implemented |
| Resource type matching | Not needed | Compared by total MinResources |

### 7.1 Priority Inversion Risk

The current design only performs preemption within the same queue. Different queues do not preempt each other, which avoids cross-queue priority inversion.

### 7.2 Resource Fragmentation

The victims' MinResources sum satisfies the preemptor's requirement, but what is released is inqueue quota rather than actual physical resources. The preemptor will still need sufficient available resources on nodes during the Allocate phase.

### 7.3 Preemption Cascading

This risk is eliminated by the priority-first victim selection strategy. Since victims are always chosen from the lowest-priority Inqueue jobs, all remaining Inqueue jobs after preemption have priority ≥ the victims' priority. The preemption condition requires `victim.Priority < preemptor.Priority` (strictly less), so victims cannot preempt any remaining Inqueue jobs — no cascading effect can occur.

### Documentation Updates

shivansh-gohem:
## What type of PR is this?

/kind feature

This PR implements a **timeout-based downgrade mechanism** for jobs stuck in the `Inqueue` state. Currently, once a job enters `Inqueue`, it holds queue quota indefinitely—even if it can never be scheduled (e.g., due to resource fragmentation or node affinity mismatches). This blocks higher-priority `Pending` jobs from enqueuing.

### Problem (Issue #5330)

The `Enqueue` action checks queue quota using:
```
(realCapability - inqueue - allocated + elastic) >= job.MinResources
```
Once low-priority jobs are enqueued, their `MinResources` are counted in `inqueue`, reducing the available quota. Higher-priority jobs arriving later find insufficient quota and remain stuck in `Pending`—even though they deserve the resources more.

### Solution: Timeout-Based Downgrade

Instead of greedy immediate preemption (which adds complexity and scheduler overhead), this PR introduces a **simple, safe timeout mechanism**:

1. **Track enqueue time**: Add `EnqueueTimestamp` to `JobInfo`, set when a job transitions to `Inqueue`.
2. **Evict stale jobs**: Before the normal enqueue loop, scan all `Inqueue` jobs. If a job has been `Inqueue` longer than the timeout (default: **5 minutes**), downgrade it back to `Pending`.
3. **Release quota**: Notify plugins (proportion, capacity) to subtract the job's `MinResources` from the queue's `inqueue` counter, freeing quota for other jobs.
4. **Re-enqueue fairly**: The downgraded job re-enters the normal priority-ordered enqueue competition, where higher-priority jobs now have a fair chance.

### Per-Job Timeout Override

Users can override the default 5-minute timeout per-job using the annotation:
```yaml
annotations:
  volcano.sh/inqueue-timeout: "600"  # 10 minutes, in seconds
```

Fixes #5330

## Changes

| File | Change |
|------|--------|
| `pkg/scheduler/api/job_info.go` | Add `EnqueueTimestamp` field to `JobInfo`, copy in `Clone()` |
| `pkg/scheduler/api/well_known_labels.go` | Add `JobInqueueTimeoutAnnotation` constant |
| `pkg/scheduler/api/types.go` | Add `JobInqueueEvictedFn` callback type |
| `pkg/scheduler/cache/cache.go` | Sync `EnqueueTimestamp` in `updateJobInfo()` |
| `pkg/scheduler/framework/session.go` | Add `jobInqueueEvictedFns` map to `Session` |
| `pkg/scheduler/framework/session_plugins.go` | Add `AddJobInqueueEvictedFn()` and `JobInqueueEvicted()` |
| `pkg/scheduler/actions/enqueue/enqueue.go` | Add Phase 1 eviction loop + `getInqueueTimeout()` helper |
| `pkg/scheduler/plugins/proportion/proportion.go` | Register eviction callback to release `inqueue` quota |
| `pkg/scheduler/plugins/capacity/capacity.go` | Register eviction callback to release `inqueue` quota (with hierarchy support) |

## Testing

- [x] `go build ./pkg/scheduler/...` — **PASS**
- [x] `go vet ./pkg/scheduler/api/... ./pkg/scheduler/actions/enqueue/... ./pkg/scheduler/framework/... ./pkg/scheduler/plugins/proportion/... ./pkg/scheduler/plugins/capacity/...` — **PASS** (clean)
- [x] `go test ./pkg/scheduler/actions/enqueue/...` — **5/5 PASS**
- [x] `go test ./pkg/scheduler/plugins/proportion/...` — **7/7 PASS**
- [x] `go test ./pkg/scheduler/framework/...` — **PASS**

Signed-off-by: Shivansh Sahu

### DIFF
--- a/pkg/scheduler/actions/enqueue/enqueue.go
+++ b/pkg/scheduler/actions/enqueue/enqueue.go
@@ -17,6 +17,7 @@ limitations under the License.
 package enqueue
 
 import (
+	"strconv"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -27,6 +28,12 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/util"
+)
+
+const (
+	// DefaultInqueueTimeout is the default duration a job is allowed to stay in Inqueue state
+	// before being downgraded back to Pending.
+	DefaultInqueueTimeout = 5 * time.Minute
 )
 
 type Action struct{}
@@ -45,6 +52,10 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 	klog.V(5).Infof("Enter Enqueue ...")
 	defer klog.V(5).Infof("Leaving Enqueue ...")
 
+	// Phase 1: Evict timed-out Inqueue jobs back to Pending to release quota.
+	enqueue.evictTimedOutInqueueJobs(ssn)
+
+	// Phase 2: Normal enqueue logic.
 	queues := util.NewPriorityQueue(ssn.QueueOrderFn)
 	queueSet := sets.NewString()
 	jobsMap := map[api.QueueID]*util.PriorityQueue{}
@@ -96,11 +107,76 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 			ssn.JobEnqueued(job)
 			job.PodGroup.Status.Phase = scheduling.PodGroupInqueue
 			ssn.Jobs[job.UID] = job
+
+			// Record the enqueue timestamp for timeout tracking.
+			if job.EnqueueTimestamp.IsZero() {
+				job.EnqueueTimestamp = metav1.Time{Time: time.Now()}
+			}
 		}
 
 		// Added Queue back until no job in Queue.
 		queues.Push(queue)
 	}
+}
+
+// evictTimedOutInqueueJobs iterates over all jobs in the Inqueue state and downgrades
+// those that have exceeded their inqueue timeout back to Pending, releasing quota.
+func (enqueue *Action) evictTimedOutInqueueJobs(ssn *framework.Session) {
+	for _, job := range ssn.Jobs {
+		if job.PodGroup == nil || job.PodGroup.Status.Phase != scheduling.PodGroupInqueue {
+			continue
+		}
+
+		// Initialize the enqueue timestamp if it's not set (e.g. for jobs that were
+		// already Inqueue before this feature was introduced).
+		if job.EnqueueTimestamp.IsZero() {
+			job.EnqueueTimestamp = metav1.Time{Time: time.Now()}
+			ssn.Jobs[job.UID] = job
+			ssn.MarkJobDirty(job.UID)
+			continue
+		}
+
+		timeout := getInqueueTimeout(job)
+		elapsed := time.Since(job.EnqueueTimestamp.Time)
+		if elapsed < timeout {
+			continue
+		}
+
+		klog.V(3).Infof("Job <%s/%s> exceeded inqueue timeout (%v > %v), downgrading to Pending",
+			job.Namespace, job.Name, elapsed, timeout)
+
+		// Downgrade to Pending.
+		job.PodGroup.Status.Phase = scheduling.PodGroupPending
+		job.EnqueueTimestamp = metav1.Time{} // reset timestamp
+
+		// Notify plugins to release the inqueue quota.
+		ssn.JobInqueueEvicted(job)
+
+		ssn.Jobs[job.UID] = job
+	}
+}
+
+// getInqueueTimeout returns the inqueue timeout for a job. It checks for an annotation
+// override first, and falls back to DefaultInqueueTimeout.
+func getInqueueTimeout(job *api.JobInfo) time.Duration {
+	if job.PodGroup == nil {
+		return DefaultInqueueTimeout
+	}
+	annotations := job.PodGroup.GetAnnotations()
+	if annotations == nil {
+		return DefaultInqueueTimeout
+	}
+	timeoutStr, found := annotations[api.JobInqueueTimeoutAnnotation]
+	if !found {
+		return DefaultInqueueTimeout
+	}
+	timeoutSeconds, err := strconv.Atoi(timeoutStr)
+	if err != nil || timeoutSeconds <= 0 {
+		klog.Warningf("Invalid inqueue-timeout annotation value %q for job <%s/%s>, using default %v",
+			timeoutStr, job.Namespace, job.Name, DefaultInqueueTimeout)
+		return DefaultInqueueTimeout
+	}
+	return time.Duration(timeoutSeconds) * time.Second
 }
 
 func (enqueue *Action) UnInitialize() {}
--- a/pkg/scheduler/actions/enqueue/enqueue.go
+++ b/pkg/scheduler/actions/enqueue/enqueue.go
@@ -109,9 +109,7 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
 			ssn.Jobs[job.UID] = job
 
 			// Record the enqueue timestamp for timeout tracking.
-			if job.EnqueueTimestamp.IsZero() {
-				job.EnqueueTimestamp = metav1.Time{Time: time.Now()}
-			}
+			job.EnqueueTimestamp = metav1.Now()
 		}
 
 		// Added Queue back until no job in Queue.
@@ -130,8 +128,7 @@ func (enqueue *Action) evictTimedOutInqueueJobs(ssn *framework.Session) {
 		// Initialize the enqueue timestamp if it's not set (e.g. for jobs that were
 		// already Inqueue before this feature was introduced).
 		if job.EnqueueTimestamp.IsZero() {
-			job.EnqueueTimestamp = metav1.Time{Time: time.Now()}
-			ssn.Jobs[job.UID] = job
+			job.EnqueueTimestamp = metav1.Now()
 			ssn.MarkJobDirty(job.UID)
 			continue
 		}
@@ -151,8 +148,6 @@ func (enqueue *Action) evictTimedOutInqueueJobs(ssn *framework.Session) {
 
 		// Notify plugins to release the inqueue quota.
 		ssn.JobInqueueEvicted(job)
-
-		ssn.Jobs[job.UID] = job
 	}
 }
 
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -452,6 +452,23 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		attr.inqueue.Add(job.DeductSchGatedResources(job.GetMinResources()))
 	})
 
+	ssn.AddJobInqueueEvictedFn(pp.Name(), func(obj interface{}) {
+		job, ok := obj.(*api.JobInfo)
+		if !ok || job == nil || job.PodGroup == nil {
+			return
+		}
+		if job.PodGroup.Spec.MinResources == nil {
+			return
+		}
+		attr := pp.queueOpts[job.Queue]
+		if attr == nil {
+			return
+		}
+		attr.inqueue.Sub(job.DeductSchGatedResources(job.GetMinResources()))
+		klog.V(4).Infof("%s:%s/%s inqueue quota released after timeout eviction",
+			pp.Name(), job.Namespace, job.Name)
+	})
+
 	ssn.AddSimulateAddTaskFn(pp.Name(), func(ctx context.Context, cycleState fwk.CycleState, taskToSchedule *api.TaskInfo, taskToAdd *api.TaskInfo, nodeInfo *api.NodeInfo) error {
 		state, err := getProportionState(cycleState)
 		if err != nil {
--- a/pkg/scheduler/plugins/proportion/proportion.go
+++ b/pkg/scheduler/plugins/proportion/proportion.go
@@ -464,7 +464,7 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
 		if attr == nil {
 			return
 		}
-		attr.inqueue.Sub(job.DeductSchGatedResources(job.GetMinResources()))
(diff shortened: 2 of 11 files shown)


shivansh-gohem:
@hzxuzhonghu @LiZhenCheng9527 @JesseStutler have a review

shivansh-gohem:
@kube-gopher @YaoZengzeng please have a look at this

pkg/scheduler/plugins/proportion/proportion.go
```diff
@@ -452,6 +452,23 @@ func (pp *proportionPlugin) OnSessionOpen(ssn *framework.Session) {
...
+			return
+		}
+		if job.PodGroup.Spec.MinResources == nil {
+			return
+		}
+		attr := pp.queueOpts[job.Queue]
+		if attr == nil {
+			return
+		}
+		attr.inqueue.Sub(job.DeductSchGatedResources(job.GetMinResources()))
```
hzxuzhonghu: At session initialization, an Inqueue job reserves `GetInqueueResource(job, job.Allocated)`, which is min resources minus what is already allocated. This callback subtracts the full `MinResources`. A partially allocated gang can therefore underflow `Resource.Sub` and panic the scheduler. Use the same remaining-resource calculation here; the capacity callback has the same mismatch.

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -96,11 +107,76 @@ func (enqueue *Action) Execute(ssn *framework.Session) {
...
+
+		klog.V(3).Infof("Job <%s/%s> exceeded inqueue timeout (%v > %v), downgrading to Pending",
+			job.Namespace, job.Name, elapsed, timeout)
+
+		// Downgrade to Pending.
+		job.PodGroup.Status.Phase = scheduling.PodGroupPending
+		job.EnqueueTimestamp = metav1.Time{} // reset timestamp
+
+		// Notify plugins to release the inqueue quota.
+		ssn.JobInqueueEvicted(job)
```
hzxuzhonghu, line 150: The default overcommit plugin also tracks `inqueueResource`, but it has no eviction callback. After this downgrade its stale reservation can reject the higher-priority pending jobs this feature is meant to admit; if the timed-out job re-enqueues in this pass, `JobEnqueued` adds the reservation again. Release overcommit accounting on this transition as well.

hzxuzhonghu (changes requested):
Timeout eviction can corrupt inqueue accounting.

shivansh-gohem:
@hzxuzhonghu @JesseStutler I've pushed the fixes to address all the review comments! The tests are fully green and the resource underflow issues in `proportion`, `capacity`, and `overcommit` have been resolved. PTAL when you have a moment.

shivansh-gohem:
@hzxuzhonghu i have adressed what uh have asked

pkg/scheduler/actions/enqueue/enqueue.go
```diff
@@ -29,6 +30,12 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
+const (
+	// DefaultInqueueTimeout is the default duration a job is allowed to stay in Inqueue state
+	// before being downgraded back to Pending.
+	DefaultInqueueTimeout = 5 * time.Minute
```
LiZhenCheng9527, line 36: Why is the default value set to 5 minutes? Do all other schedulers use 5 minutes as well?

LiZhenCheng9527 (commented):
Please fix AI review comments

shivansh-gohem:
@LiZhenCheng9527 Good question! 

I set it to 5 minutes to strike a balance between giving a job enough time to schedule and preventing it from hogging quota indefinitely. 

In a real cluster, if a job needs to trigger the Cluster Autoscaler to spin up new nodes, or pull very large images, that process typically takes about 2 to 4 minutes. If the timeout was shorter (e.g. 1 minute), jobs might get prematurely evicted just because a node is still spinning up. 

Also, 5 minutes is a fairly standard precedent in the ecosystem for similar features—for example, **Kueue** uses a `5m` default timeout for its `waitForPodsReady` mechanism for exactly the same reasons. 

Users can also override this default per-job using the `volcano.sh/inqueue-timeout` annotation if they have exceptionally slow node provisioning. I can add a brief comment above the constant in the code to document this rationale if you think it would be helpful!

Issue #5443, opened 2026-06-14
`PodGroup` admission webhook concatenates multiple `validation errors` without a separator


DevMhrn:
### Description

When a PodGroup fails more than one validation check at the same time (for example, a non-existent queue and an invalid `networkTopology` spec), the admission webhook returns a rejection message that joins each validator's error string without any separator. The two messages get smashed together mid-sentence, making the rejection hard to read.

The concatenation happens in `pkg/webhooks/admission/podgroups/validate/validate_podgroup.go` around lines 95-96, where `validatePodGroup` appends each helper's return value to `errMsg` using plain `+=:`

```go
errMsg += checkQueueState(pg.Spec.Queue)
errMsg += validateNetworkTopology(pg.Spec.NetworkTopology, pg.Spec.SubGroupPolicy)
```
The inner helper `validateNetworkTopology` already uses `strings.Join(..., " ")` for its own errors, but there is no equivalent join at this outer combine step.

### Steps to reproduce the issue

1. On a cluster with Volcano installed, apply a `PodGroup` whose `spec.queue` references a queue that does not exist, and which sets both `highestTierAllowed` and `highestTierName` under `spec.networkTopology`:
```yaml
apiVersion: scheduling.volcano.sh/v1beta1
kind: PodGroup
metadata:
  name: bad-pg
  namespace: default
spec:
  minMember: 1
  queue: doesnotexist
  networkTopology:
    mode: hard
    highestTierAllowed: 2
    highestTierName: spine


```
2. Observe the webhook rejection message from `kubectl apply`.
3. For comparison, apply the same PodGroup with only the bad queue (no `networkTopology`), then again with only the bad `networkTopology` (no `queue`), and observe that each individual validator produces a clean, well-formed message.

### Describe the results you received and expected

Received (literal `kubectl apply` output)
```error
Error from server: error when creating "STDIN": admission webhook "validatepodgroup.volcano.sh" denied the request: unable to find queue: queue.scheduling.volcano.sh "doesnotexist" not foundmust not specify 'highestTierAllowed' and 'highestTierName' in networkTopology simultaneously.
```

The `not foundmust not specify` portion is two distinct error messages concatenated with no delimiter. Each validator alone produces a clean message, confirmed by running the single-error reproducers in step 3:

 - Queue-only error: `unable to find queue: queue.scheduling.volcano.sh "doesnotexist" not found`
[...]

PR #5533, opened 2026-06-29
webhook: separate cronjob validation errors with "; "

### LINKED ISSUE #5489: CronJob validation webhook concatenates multiple errors without a separator
### Description

While reading through the admission webhooks I noticed that the CronJob validation webhook can return a rejection message where two separate errors are glued together with no space between them.

In `pkg/webhooks/admission/cronjobs/validate/admit_cronjob.go`, `validateCronJobCreate` builds the message by appending the results of two helpers:

```go
msg := validateCronjobSpec(&cronjob.Spec, cronjob.Namespace)
msg += validateCronJobName(cronjob.Name)
```

The helpers don't agree on spacing. Most messages in `validateCronjobSpec` start with a leading space, but the invalid-schedule one doesn't:

```go
msg += "schedule is not a valid cron expression"   // no leading space
```

and `validateCronJobName` also returns messages without a leading space (e.g. `cronJob name must be no more than 52 characters ...`). So when a CronJob fails more than one check at the same time, the words run straight into each other.

This is the same kind of problem that was reported for the PodGroup webhook in #5443, just in the CronJob webhook, which that issue doesn't cover.

1. Create a CronJob that fails two checks at once, for example an invalid schedule together with a name longer than 52 characters:

```yaml
kind: CronJob
metadata:
  name: this-name-is-deliberately-far-too-long-to-pass-the-cronjob-name-check
spec:
  schedule: "not-a-valid-cron"
  # ... rest of a valid cronJobSpec
```

2. Apply it and read the rejection message returned by the webhook.

Received (the two errors are concatenated with no separator):

```
schedule is not a valid cron expressioncronJob name must be no more than 52 characters to accommodate job name suffix (max 63 total)
```

Expected (errors reported as separate, readable items):

```
schedule is not a valid cron expression; cronJob name must be no more than 52 characters to accommodate job name suffix (max 63 total)
```

The Job webhook (`pkg/webhooks/admission/jobs/validate/admit_job.go`) leans on the same fragile pattern, where every message has to carry its own leading space / trailing `;`. I'm happy to tidy that up in the same change or a follow-up, whichever maintainers prefer.

master (the code is also present in v1.14.0)

This is purely in the admission webhook logic, so it isn't environment specific.

anxkhn:
#### What type of PR is this?

/kind bug

The CronJob create/update validators built their rejection message by appending
each validator's output with `+=` and no separator:

```go
msg := validateCronjobSpec(&cronjob.Spec, cronjob.Namespace)
msg += validateCronJobName(cronjob.Name)
```

So when a CronJob failed more than one check at once (for example an invalid
schedule *and* a name over 52 characters), the messages ran together mid-sentence
and the rejection was hard to read:

```
schedule is not a valid cron expressioncronJob name must be no more than 52 characters ...
```

This collects each non-empty message into a slice and joins them with `"; "`, so
the errors are reported as distinct, readable items. An empty slice still joins to
`""`, so valid CronJobs and single-error paths are unchanged. This is the CronJob
counterpart of the PodGroup fix in #5443 / #5487; it scopes only the CronJob
webhook (the optional `admit_job.go` cleanup mentioned in the issue is left out).

Fixes #5489

I used an AI assistant to help draft this change; I reviewed and tested every line
and can explain why each change was made, and I'll respond to review comments
without AI tools, per contributing.md#ai-guidance. The fix mirrors the in-flight
PodGroup fix (#5487): `validateCronjobSpec`'s own output is left untouched, so the
existing leading-space test assertions still hold. Added a focused test that fails
both the schedule and name checks at once and asserts the `"; "` separator.

@aeron-gh - you mentioned on the issue you might start on this; it sat unclaimed for
a week so I went ahead with the fix and a regression test. Happy to defer or close
if you'd rather take it.

```release-note
NONE
```

### DIFF
--- a/pkg/webhooks/admission/cronjobs/validate/admit_cronjob.go
+++ b/pkg/webhooks/admission/cronjobs/validate/admit_cronjob.go
@@ -105,21 +105,41 @@ func AdmitCronjobs(ar admissionv1.AdmissionReview) *admissionv1.AdmissionRespons
 	return &reviewResponse
 }
 func validateCronJobCreate(cronjob *v1alpha1.CronJob, reviewResponse *admissionv1.AdmissionResponse) string {
-	msg := validateCronjobSpec(&cronjob.Spec, cronjob.Namespace)
-	msg += validateCronJobName(cronjob.Name)
+	msg := joinCronJobErrs(cronjob)
 	if msg != "" {
 		reviewResponse.Allowed = false
 	}
 	return msg
 }
 func validateCronJobUpdate(new *v1alpha1.CronJob) error {
-	msg := validateCronjobSpec(&new.Spec, new.Namespace)
-	msg += validateCronJobName(new.Name)
-
+	msg := joinCronJobErrs(new)
 	if msg != "" {
 		return errors.New(msg)
 	}
 	return nil
+}
+
+// joinCronJobErrs collects the non-empty messages from each CronJob validator.
+// When only one check fails, its message is returned verbatim so single-error
+// responses stay byte-for-byte identical to the previous behaviour. When more
+// than one check fails at once, the messages are trimmed and joined with "; "
+// so they are reported as distinct, readable items instead of running together
+// mid-sentence.
+func joinCronJobErrs(cronjob *v1alpha1.CronJob) string {
+	var errs []string
+	if msg := validateCronjobSpec(&cronjob.Spec, cronjob.Namespace); msg != "" {
+		errs = append(errs, msg)
+	}
+	if msg := validateCronJobName(cronjob.Name); msg != "" {
+		errs = append(errs, msg)
+	}
+	if len(errs) == 1 {
+		return errs[0]
+	}
+	for i := range errs {
+		errs[i] = strings.TrimSpace(errs[i])
+	}
+	return strings.Join(errs, "; ")
 }
 func validateCronjobSpec(spec *v1alpha1.CronJobSpec, nameSpace string) string {
 	var msg string
--- a/pkg/webhooks/admission/cronjobs/validate/admit_cronjob_test.go
+++ b/pkg/webhooks/admission/cronjobs/validate/admit_cronjob_test.go
@@ -412,3 +412,90 @@ func TestValidateCronJobName(t *testing.T) {
 		})
 	}
 }
+
+// TestJoinCronJobErrsSeparatesMultipleErrors checks that when a CronJob fails
+// more than one validation at once (here an invalid schedule and a too-long
+// name), the messages are reported as distinct items joined by "; " rather than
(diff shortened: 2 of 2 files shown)


anxkhn:
Thanks both.

@gemini-code-assist `joinCronJobErrs` now takes the `*v1alpha1.CronJob` directly and the `validateCronJobCreate`/`validateCronJobUpdate` callers pass the object, so the data clump is gone.

@Copilot good catch on the formatting. `joinCronJobErrs` no longer trims unconditionally: when only one validator fails its message is returned verbatim, so single-error responses (including the UPDATE path via `util.ToAdmissionResponse`, and the no-error path) are byte-for-byte identical to before. The `TrimSpace` + `; ` join now applies only when more than one check fails. Added `TestJoinCronJobErrsSingleErrorUnchanged`, which asserts the spec-only message keeps its leading space and that no separator is introduced (it fails against the previous implementation).

hwdef:
/lgtm
Thanks

pkg/webhooks/admission/cronjobs/validate/admit_cronjob.go
```diff
@@ -105,22 +105,42 @@ func AdmitCronjobs(ar admissionv1.AdmissionReview) *admissionv1.AdmissionRespons
 	return &reviewResponse
 }
 func validateCronJobCreate(cronjob *v1alpha1.CronJob, reviewResponse *admissionv1.AdmissionResponse) string {
-	msg := validateCronjobSpec(&cronjob.Spec, cronjob.Namespace)
-	msg += validateCronJobName(cronjob.Name)
+	msg := joinCronJobErrs(cronjob)
```
hajnalmt, line 108: I don't like this coding style.

`JoinCronJobErrs` is not just joining errors, it validates too. You are hiding functionality in a very vaguely named function now.

Passing the whole pointer is really out of scope too for a function named like this.

hajnalmt (changes requested):
/area webhooks

Thanks for the PR, the issue is valid props for looking into it. After a search grep it looks like this is the last pattern where we have this, next to podgroup which will be fixed in https://github.com/volcano-sh/volcano/pull/5487 . Unfortunately the PR itself needs some improvements, it handles function naming and responsibility quite badly. Please rethink it.

aeron-gh:
i opened this issue (#5489) and had said I'd work on it. since @hajnalmt has asked for a rethink, i'd be glad to take it from here I've already implemented the same collect-and-join pattern in #5487, so I can mirror it cleanly for the CronJob webhook. thanks for kicking it off, @anxkhn.

hajnalmt (commented):
/close

Thank you for the PR @anxkhn !
#5610 was a lot closer to my taste. I hope this does not discourage you from contributing further!

PR #5539, opened 2026-06-29
fix(scheduler): propagate job anchor when both job and subJob use soft topology

### LINKED ISSUE #5514: Job-level soft HyperNode affinity ineffective when SubJob also uses soft topology
### Description

When both Job-level and SubJob-level HyperNode soft topology are configured on a PodGroup, Job-level soft affinity often appears ineffective: pods within a SubJob may cluster, but pods across SubJobs tend to spread across sibling HyperNodes (e.g. master on s0, worker’s first pod on s1).

[#5206](https://github.com/volcano-sh/volcano/pull/5206) converts soft constraints to hard and reuses the hard gradient path, which helps Job soft only or Job soft + SubJob hard; the issue remains when Job soft and SubJob soft are configured together.

**HyperNode: **
s2 (root, tier-2)
├── s0 (tier-1):  n0, n1   （each 8 GPU）
└── s1 (tier-1):  n2       （8 GPU）

**vcjob:**
```
spec:
  networkTopology:
    mode: soft
    highestTierAllowed: 2
  subGroupPolicy:
    - name: master
      matchLabels: ["volcano.sh/task-spec"]
      subGroupSize: 1
      networkTopology:
        mode: soft
        highestTierAllowed: 1
    - name: worker
      matchLabels: ["volcano.sh/task-spec"]
      subGroupSize: 2
      networkTopology:
        mode: soft
        highestTierAllowed: 1
```

- Prefer lower tiers: lower LCA HyperNode tier is better;
- Minimize sibling spread at the same tier: pods should stay on the same HyperNode when possible (e.g. all on s0).

**Example：**
`s2(root) → s0(n0,n1), s1(n2)`
Job and SubJobs use mode: soft; 1 master pod + 2 worker pods, 4 GPU each.

Pod | Expected | Common actual
-- | -- | --
p1 (master) | land on s0 or s1 | n0 @ s0
p2 (worker, first pod) | same as master on s0 | may land on n2 @ s1
Final placement | same tier-1 HyperNode | Job LCA may widen to s2

Typical trigger: Job dry-run on tier-1 (s0/s1) cannot fit all SubJobs, falls back to root s2, then the worker SubJob independently picks between s0 and s1.

master

**Root cause (two layers)**
1. HyperNode layer: SubJob hyperNodeGradientFn uses only subJob.AllocatedHyperNode, not job.AllocatedHyperNode; under root, each SubJob dry-runs / binpacks independently and does not follow sibling SubJob placement.

2. Node layer: task.JobAllocatedHyperNode = subJob.AllocatedHyperNode; for a SubJob’s first pod, the anchor is empty → topology score is 0, with no fallback to job.AllocatedHyperNode.

#5206 adds the hard gradient path for soft, but does not fix cross-SubJob anchor propagation or node scoring.

**Suggested follow-ups**
When the SubJob anchor is empty, use job.AllocatedHyperNode for gradient searchRoot and node scoring;
Add Job soft + SubJob soft allocate UT (existing cases mostly cover Job soft + SubJob hard).

vanshika2720:
## What type of PR is this?

/kind bug

When both a PodGroup and its SubGroups are configured with **soft** network topology, successive SubJob allocations within the same HyperNode dry-run do not share placement information.

Previously, the first successfully allocated SubJob updated only `subJob.AllocatedHyperNode`. The corresponding `job.AllocatedHyperNode` was updated later during `UpdateDecisionToJob()`, after the dry-run had completed. As a result, the next SubJob started with an empty topology anchor, causing the first task's node scoring (`batchNodeOrderFnForNetworkAwarePods`) to treat all candidate nodes equally instead of preferring the HyperNode selected by previously allocated SubJobs.

This PR preserves the job-level topology anchor throughout the dry-run so that later SubJobs can make topology-aware placement decisions.

### Changes

**allocate.go**

* Propagate each successfully allocated SubJob's HyperNode to `job.AllocatedHyperNode` during the per-HyperNode dry-run using `GetLCAHyperNode()`.
* In `allocateResourcesForTasks()`, fall back to `job.AllocatedHyperNode` when `subJob.AllocatedHyperNode` is empty, ensuring the first task of a SubJob always has a topology anchor for node scoring.

**recorder.go**

* Extend `SnapshotSubJobStatus()` and `RecoverSubJobStatus()` to snapshot and restore `job.AllocatedHyperNode` together with each SubJob's `AllocatedHyperNode`, ensuring different dry-run attempts remain fully isolated.

Fixes #5514

The original issue reproducer does not expose this behavior because, after the first SubJob is allocated, the remaining candidate nodes exactly match the remaining pod count, leaving no placement choice.

This change targets scenarios where the remaining candidate pool is larger than the remaining SubJob size. In those cases, preserving the job-level topology anchor allows subsequent SubJobs to continue preferring the same HyperNode instead of treating all candidate nodes equally.

```release-note
Improve soft network topology scheduling when both Job-level and SubJob-level topology policies are configured by preserving the job-level HyperNode anchor across SubJob allocations during scheduling. This allows subsequent SubJobs to make topology-aware placement decisions within the same scheduling cycle.
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -401,6 +401,12 @@ func (alloc *Action) allocateForJob(job *api.JobInfo, jobWorksheet *JobWorksheet
 				if stmt != nil && len(stmt.Operations()) > 0 {
 					stmtList = append(stmtList, stmt)
 					subJobsAllocationScore += allocationScore
+					// Propagate this subJob's anchor to the job so subsequent subJobs use it
+					// for node scoring. RecoverSubJobStatus will reset job.AllocatedHyperNode
+					// after each hyperNode dry-run, keeping iterations isolated.
+					if subJob.AllocatedHyperNode != "" {
+						job.AllocatedHyperNode = ssn.HyperNodes.GetLCAHyperNode(job.AllocatedHyperNode, subJob.AllocatedHyperNode)
+					}
 					// push back when subJob is ready and remain pending task
 					if !subJobWorksheet.Empty() {
 						jobWorksheetCopy.subJobs.Push(subJob)
@@ -738,6 +744,9 @@ func (alloc *Action) allocateResourcesForTasks(subJob *api.SubJobInfo, tasks *ut
 	ph := util.NewPredicateHelper()
 
 	allocatedHyperNode := subJob.AllocatedHyperNode
+	if allocatedHyperNode == "" {
+		allocatedHyperNode = job.AllocatedHyperNode
+	}
 
 	for !tasks.Empty() {
 		task := tasks.Pop().(*api.TaskInfo)
--- a/pkg/scheduler/actions/allocate/allocate_test.go
+++ b/pkg/scheduler/actions/allocate/allocate_test.go
@@ -5547,6 +5547,86 @@ func TestAllocateWithPartitionPolicyNetworkTopology(t *testing.T) {
 			ExpectBindsNum:   2,
 			MinimalBindCheck: true,
 		},
+		{
+			// Regression test for issue #5514: when both job and subGroup use soft topology the
+			// job-level anchor (AllocatedHyperNode) must be propagated to subsequent subJob
+			// gradient lookups so that workers are scored relative to where the master landed.
+			//
+			// Topology: s0(n1,n2) and s1(n3,n4) under root s2.  Each node holds exactly one
+			// 4-CPU pod.  The master subJob (size 1) and worker subJob (size 3) both set
+			// highestTierAllowed=1 (soft), so the job-level soft constraint (tier 2) drives
+			// allocation to s2 after Tier-1 candidates fail capacity checks.  Under s2 the
+			// anchor propagated from the master placement guides worker node scoring.
+			Name: "podgroup soft network topology and subGroup soft network topology, job falls to root tier, can allocate all pods",
+			PodGroups: []*schedulingv1.PodGroup{
+				util.BuildPodGroupWithSubGroupPolicy("pg1", "c1", "", "q1", 4, nil, schedulingv1.PodGroupInqueue, "soft", 2,
+					[]schedulingv1.SubGroupPolicySpec{
+						util.BuildSubGroupPolicyWithSubGroupSize("master", []string{"volcano.sh/task-spec"}, "soft", 1, 1),
+						util.BuildSubGroupPolicyWithSubGroupSize("worker", []string{"volcano.sh/task-instance"}, "soft", 1, 3),
+					}),
+			},
+			Pods: []*v1.Pod{
+				util.BuildPod("c1", "p1", "", v1.PodPending, api.BuildResourceList("4", "8G"), "pg1", map[string]string{"volcano.sh/task-spec": "master"}, nil),
+				util.BuildPod("c1", "p2", "", v1.PodPending, api.BuildResourceList("4", "8G"), "pg1", map[string]string{"volcano.sh/task-instance": "worker"}, nil),
+				util.BuildPod("c1", "p3", "", v1.PodPending, api.BuildResourceList("4", "8G"), "pg1", map[string]string{"volcano.sh/task-instance": "worker"}, nil),
+				util.BuildPod("c1", "p4", "", v1.PodPending, api.BuildResourceList("4", "8G"), "pg1", map[string]string{"volcano.sh/task-instance": "worker"}, nil),
+			},
+			Nodes: []*v1.Node{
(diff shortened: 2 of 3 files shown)


vanshika2720:
@JesseStutler
@hajnalmt
@wangyang0616
@hwdef

VanderChen:
**The Job anchor is not preserved across subsequent allocateForJob() calls**
This change allows a later SubJob to use the Job anchor for node scoring when multiple SubJobs are evaluated inside the same outer HyperNode dry-run. However, it does not cover the case where the first SubJob makes the Job ready while another SubJob remains in the worksheet.

For example, consider two independent tier-2 domains:
```text
tier2-A
  ├── tier1-A1
  └── tier1-A2

tier2-B
  ├── tier1-B1
  └── tier1-B2
```

The Job has soft topology with `highestTierAllowed: 2`, and two SubJobs each have soft topology with `highestTierAllowed: 1`. Each SubJob contains two Pods. Suppose the Job’s `minAvailable` is satisfied after the first SubJob, while `minSubGroups` does not require the second one—for example, `minAvailable: 2` and `minSubGroups: 0`.

After the first SubJob is placed on `A1`, `JobReady()` becomes true and `allocateForJob()` stops processing additional SubJobs. The resulting Statement is committed, the Job anchor is updated to `A1`, and, because the worksheet still contains the second SubJob, the Job is pushed back into the queue. The second SubJob is therefore handled by a subsequent `allocateForJob()` call, potentially within the same allocate action.

In that second call, `A2` and `B1` may be evaluated as separate outer HyperNode candidates. While trying `B1`, this patch passes the existing Job anchor `A1` into node scoring, but the dry-run contains only nodes from `B1`; there are no nodes from `A2` to compare against. As a result, the remote `B1` solution remains feasible.

The outer selection still calls `selectBestHyperNodeForJob()` using only `subJobsAllocationScores`. It does not compare each solution’s final Job LCA with the existing Job anchor. Therefore, `B1` may win because of a higher ordinary score or a tie, even though `A2` is feasible. The final Job LCA then rises above the configured preferred tier.

The regression test added in this PR does not exercise this path. It uses `minAvailable: 4`, so the one-Pod master does not make the Job ready and both SubJobs are processed in the same `allocateForJob()` call. It also has only one tier-2 parent, and after the master is placed, the three worker Pods exactly match the three remaining node slots. Binding all four Pods therefore does not prove that the cross-SubJob Job-level preference is preserved.

Could we add a regression that:

1. uses two independent tier-2 domains;
2. makes the first SubJob satisfy `JobReady()` while leaving the second SubJob in the worksheet;
3. gives a remote outer candidate a higher ordinary score than the local candidate; and
4. asserts that the final Job LCA remains within the original tier-2 domain?
 
Supporting this case likely requires carrying each outer solution’s final Job placement into the outer-candidate comparison, rather than using the Job anchor only as a node-scoring hint.

VanderChen:
**Keep the scoring anchor separate from SubJob placement state**

Additionally, I am concerned that this initialization conflates two different concepts:

```go
allocatedHyperNode := subJob.AllocatedHyperNode
if allocatedHyperNode == "" {
    allocatedHyperNode = job.AllocatedHyperNode
}
```

On current master, where only network-topology-aware scheduling consumes this state, using the Job anchor as a node-scoring hint is reasonable. However, the same variable is subsequently updated from selected nodes and written back to `subJob.AllocatedHyperNode`. This means the aggregate Job LCA becomes part of the SubJob’s own placement LCA.

This becomes problematic for the SubGroup affinity/anti-affinity scheduling proposed in issue #5347 and doc PR #5349 . For example:

```text
supernode-S
  ├── cabinet-A: prefill-0
  └── cabinet-B: prefill-1
```

The Job anchor is `supernode-S`. A new `decode-0` SubJob may need affinity with `prefill-0` in cabinet A and separation from `prefill-1` in cabinet B. The aggregate Job anchor cannot express that relationship. If `decode-0` starts with `supernode-S` and then selects a node in cabinet A, its recorded placement remains:

```text
LCA(supernode-S, cabinet-A) = supernode-S
```

We therefore lose the precise SubJob placement needed for peer-specific affinity. This can collapse distinct allocation options into the same parent HyperNode and may over-expand occupancy during fallback or restart recovery.

Could we keep two separate variables instead?

```go
subJobPlacement := subJob.AllocatedHyperNode

nodeScoreAnchor := subJobPlacement
if nodeScoreAnchor == "" {
    nodeScoreAnchor = job.AllocatedHyperNode
}
task.JobAllocatedHyperNode = nodeScoreAnchor

// Only the current SubJob's selected nodes update its placement.
subJobPlacement = getNewAllocatedHyperNode(
    ssn, bestNode.Name, subJobPlacement,
)
```

The Job anchor would then remain an NTA scoring hint, while `subJob.AllocatedHyperNode` continues to represent only that SubJob’s actual placement. Future SubGroup affinity should use the exact occupied domains of all peers matching the affinity term, rather than the aggregate Job anchor.

PR #5542, opened 2026-06-29
fix(predicates): check skip set in predicateByStablefilter

### LINKED ISSUE #5536: Add skip check in `predicateByStablefilter`
### Description

After #5505, plugins registered as stable filter (e.g. `NodeAffinity`) also have `PreFilter` that can return `Skip`. However, `predicateByStablefilter` does not call `handleSkipPredicatePlugin(state, name)` before executing `Filter`, so the skip optimization doesn't take effect in the stable filter path.

1. Schedule a pod with no node affinity terms.
2. `PreFilter` returns `Skip` and adds `NodeAffinity` to skip set.
3. `predicateByStablefilter` still calls `NodeAffinity.Filter()` for every node.

**Actual:** stable filter path ignores the skip set, `Filter()` still gets called.

**Expected:** `predicateByStablefilter` should check `handleSkipPredicatePlugin(state, name)` and skip accordingly, same as the regular filter path.

latest

Related: https://github.com/volcano-sh/volcano/pull/5505
Discussion: https://github.com/volcano-sh/volcano/pull/5535#discussion_r3490236147

vanshika2720:
## Description

Stable filter plugins registered via `addStableFilterPlugin` were not consulting the skip set populated during the `PreFilter` phase. As a result, plugins such as `NodeAffinity` would still execute `Filter()` for every node even when their `PreFilter()` had already returned `framework.Skip`, preventing the skip optimization introduced in #5505 from taking effect in the stable filter path.

This PR adds the same `handleSkipPredicatePlugin` check used by the regular filter path before invoking stable filter plugins. If a plugin is present in the skip set, its `Filter()` is skipped, making the behavior of the stable filter path consistent with the regular filter path.

A regression test has also been added to verify that a stable filter plugin whose name is present in the skip set does not have its `Filter()` invoked, while plugins that are not skipped continue to execute normally.

Fixes #5536

---

#### What type of PR is this?

/kind bug

* Adds the missing skip check to `predicateByStablefilter`.
* Makes stable filter plugins honor the skip set populated during `PreFilter`.
* Aligns the stable filter path with the existing behavior of the regular filter path.
* Adds a regression test to ensure skipped plugins do not execute `Filter()`, while non-skipped plugins continue to execute normally.

Fixes #5536

This change only affects the stable filter execution path when a plugin has already been marked as skipped during `PreFilter`. Scheduling behavior for plugins that are not skipped remains unchanged. The regression test covers both the skipped and non-skipped cases to ensure the optimization works as intended.

```release-note
Fix stable filter plugins to honor the skip set populated during PreFilter, avoiding unnecessary Filter() executions for skipped plugins.
```

### DIFF
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -682,6 +682,9 @@ func (pp *PredicatesPlugin) Predicate(task *api.TaskInfo, node *api.NodeInfo, st
 			if !exists {
 				continue
 			}
+			if handleSkipPredicatePlugin(state, name) {
+				continue
+			}
 			status := plugin.Filter(context.TODO(), state, task.Pod, nodeInfo)
 			filterStatus := api.ConvertPredicateStatus(status)
 			if filterStatus.Code != api.Success {
--- a/pkg/scheduler/plugins/predicates/predicates_test.go
+++ b/pkg/scheduler/plugins/predicates/predicates_test.go
@@ -26,6 +26,7 @@ import (
 	schedulingv1 "k8s.io/api/scheduling/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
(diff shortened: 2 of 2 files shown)


vanshika2720:
@JesseStutler 
@hajnalmt

JesseStutler:
Thank you so much
/approve
/lgtm

PR #5554, opened 2026-06-30
Fix agent scheduler config hot reload not affecting running workers

### LINKED ISSUE #5553: Agent scheduler config hot reload does not affect running workers
### Description

The agent scheduler logs that scheduler configuration has been reloaded after its ConfigMap is updated, but the running scheduling workers continue using the old framework/plugin configuration. The new configuration only takes effect after the agent scheduler Pod is restarted.

1. Deploy the agent scheduler with the default agent scheduler config:

   ```yaml
   actions: "allocate"
   tiers:
   - plugins:
     - name: predicates
     - name: nodeorder
   ```

2. Create a Pod scheduled by `agent-scheduler` with a `nodeSelector` that does not match any node:

   ```yaml
   apiVersion: v1
   kind: Pod
   metadata:
     name: reload-before
     namespace: volcano-reload-test
   spec:
     nodeSelector:
       volcano.sh/reload-test: impossible-before
     containers:
     - name: main
       image: busybox:1.36
       command: ["sh", "-c", "sleep 3600"]
   ```

3. Confirm that the Pod remains Pending because predicates reject it:

   ```text
   Warning  FailedScheduling  agent-scheduler  0/1 nodes are unavailable: 1 node(s) didn't match Pod's node affinity/selector.
   ```

4. Update the agent scheduler ConfigMap to disable the predicates plugin's predicate extension:

   ```yaml
   actions: "allocate"
   tiers:
   - plugins:
     - name: predicates
       enablePredicate: false
     - name: nodeorder
   ```

5. Wait for the projected ConfigMap volume to update. The agent scheduler logs show that config reload was triggered:

   ```text
   Finished loading scheduler config. Final state: actions=[allocate], plugins=[predicates nodeorder]
   Finished loading scheduler config. Final state: actions=[allocate], plugins=[predicates nodeorder]
   Finished loading scheduler config. Final state: actions=[allocate], plugins=[predicates nodeorder]
   ```

6. Create another Pod with the same kind of non-matching `nodeSelector`:

   ```yaml
   apiVersion: v1
   kind: Pod
   metadata:
     name: reload-after
     namespace: volcano-reload-test
   spec:
     nodeSelector:
       volcano.sh/reload-test: impossible-after
     containers:
     - name: main
       image: busybox:1.36
       command: ["sh", "-c", "sleep 3600"]
   ```

## Actual Result

The new Pod is still rejected by predicates after the config reload:

```text
Warning  FailedScheduling  agent-scheduler  0/1 nodes are unavailable: 1 node(s) didn't match Pod's node affinity/selector.
```

Agent scheduler logs after the reload also show predicate failures for the new Pod:

```text
Predicates failed: task volcano-reload-test/reload-after on node orbstack fit failed: node(s) didn't match Pod's node affinity/selector
"Predicate failed" err="0/1 nodes are unavailable: 1 node(s) didn't match Pod's node affinity/selector." task="volcano-reload-test/reload-after"
```

After restarting the agent scheduler Pod without changing the ConfigMap, the same disabled-predicate configuration takes effect. The previously Pending Pods are bound to the node, and then kubelet reports `NodeAffinity`, which confirms that the scheduler no longer rejected them during scheduling.

Observed after restart:

```text
NAME              PHASE    REASON         NODE       SCHEDULER
reload-after      Failed   NodeAffinity   orbstack   agent-scheduler
reload-before     Failed   NodeAffinity   orbstack   agent-scheduler
```

## Expected Result

After the agent scheduler reloads the updated scheduler configuration, newly scheduled Pods should use the reloaded plugin configuration without requiring an agent scheduler Pod restart.

v1.15.0

_No response_

togettoyou:
#### What type of PR is this?

/kind bug

This PR fixes agent scheduler configuration hot reload so that updated scheduler configuration is applied to already running workers.

Previously, `watchSchedulerConf` could reload the scheduler config and log a successful reload, but existing workers kept using the framework created during scheduler startup. As a result, changes to plugin configuration, such as disabling `predicates.enablePredicate`, did not affect newly scheduled Pods until the agent scheduler Pod was restarted.

This PR updates the agent scheduler to:

- Track running workers in the scheduler.
- Rebuild each worker's framework after scheduler config reload succeeds.
- Re-run action initialization when applying a new config.
- Avoid holding the worker framework lock while waiting on `SchedulingQueue.Pop`.
- Preserve default config fallback behavior and action initialization.

Fixes #5553

```release-note
NONE
```

### DIFF
--- a/pkg/agentscheduler/framework/plugins.go
+++ b/pkg/agentscheduler/framework/plugins.go
@@ -221,14 +221,23 @@ func (f *Framework) NodeOrderReduceFn(task *api.TaskInfo, pluginNodeScoreMap map
 }
 
 // Action management
-var actionMap = map[string]Action{}
+type ActionBuilder func() Action
+
+var actionBuilders = map[string]ActionBuilder{}
 
 // RegisterAction register action
 func RegisterAction(act Action) {
+	RegisterActionBuilder(act.Name(), func() Action {
+		return act
+	})
+}
+
+// RegisterActionBuilder registers the action builder.
+func RegisterActionBuilder(name string, ab ActionBuilder) {
 	pluginMutex.Lock()
 	defer pluginMutex.Unlock()
 
-	actionMap[act.Name()] = act
+	actionBuilders[name] = ab
 }
 
 // GetAction get the action by name
@@ -236,6 +245,9 @@ func GetAction(name string) (Action, bool) {
 	pluginMutex.RLock()
 	defer pluginMutex.RUnlock()
 
-	act, found := actionMap[name]
-	return act, found
+	ab, found := actionBuilders[name]
+	if !found {
+		return nil, false
+	}
+	return ab(), true
 }
--- a/pkg/agentscheduler/scheduler.go
+++ b/pkg/agentscheduler/scheduler.go
@@ -63,9 +63,12 @@ type Scheduler struct {
 	disableDefaultConf bool
 	workerCount        int
 	shardingMode       string
+	workers            []*Worker
+	configApplied      bool
 }
 
 type Worker struct {
+	mutex     sync.RWMutex
 	framework *framework.Framework
 	index     int
 }
@@ -99,19 +102,14 @@ func NewAgentScheduler(config *rest.Config, opt *options.ServerOption) (*Schedul
 // initializes the cache, and begins the scheduling process.
 func (sched *Scheduler) Run(stopCh <-chan struct{}) {
 	sched.loadSchedulerConf()
-	go sched.watchSchedulerConf(stopCh)
 	// Start cache for policy.
-	sched.cache.SetMetricsConf(sched.metricsConf)
+	sched.cache.SetMetricsConf(sched.getMetricsConf())
 	sched.cache.Run(stopCh)
-
-	for _, action := range sched.actions {
-		action.OnActionInit(sched.configurations)
-	}
 
 	klog.V(2).Infof("Scheduler completes Initialization and start to run %d workers", sched.workerCount)
 	for i := range sched.workerCount {
-		worker := &Worker{index: i}
-		worker.framework = framework.NewFramework(sched.actions, sched.tiers, sched.cache, sched.configurations)
+		worker := sched.newWorker(i)
+		sched.addWorker(worker)
 		go func() {
 			for {
 				select {
@@ -123,6 +121,7 @@ func (sched *Scheduler) Run(stopCh <-chan struct{}) {
 			}
 		}()
 	}
+	go sched.watchSchedulerConf(stopCh)
 	if options.ServerOpts.EnableCacheDumper {
 		sched.dumper.ListenForSignal(stopCh)
 	}
@@ -146,14 +145,22 @@ func (worker *Worker) runOnce() {
 		return
 	}
 
+	worker.mutex.RLock()
+	defer worker.mutex.RUnlock()
+	if worker.framework == nil {
+		klog.Errorf("Framework is not initialized in worker %d", worker.index)
+		return
+	}
+	fwk := worker.framework
+
 	scheduleStartTime := time.Now()
 
 	// Update snapshot from cache before scheduling
-	snapshot := worker.framework.GetSnapshot()
+	snapshot := fwk.GetSnapshot()
 	snapshotStart := time.Now()
-	if err := worker.framework.Cache.UpdateSnapshot(snapshot); err != nil {
+	if err := fwk.Cache.UpdateSnapshot(snapshot); err != nil {
 		klog.Errorf("Failed to update snapshot in worker %d: %v, skip this scheduling cycle", worker.index, err)
-		queue := worker.framework.Cache.SchedulingQueue()
+		queue := fwk.Cache.SchedulingQueue()
 		if err := queue.AddUnschedulableIfNotPresent(klog.Background(), schedCtx.QueuedPodInfo, queue.SchedulingCycle()); err != nil {
 			klog.ErrorS(err, "Failed to add pod back to scheduling queue",
 				"pod", klog.KObj(schedCtx.QueuedPodInfo.Pod))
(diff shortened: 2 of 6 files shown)


togettoyou:
@JesseStutler 

Sorry, I found two additional issues while re-testing the hot reload behavior.

### Binder registry is not refreshed after reload

Reproduce steps:

1. Deploy `agent-scheduler` with `predicates` enabled:

```yaml
actions: "allocate"
tiers:
- plugins:
  - name: predicates
  - name: nodeorder
```

2. Hot reload the config and remove `predicates`:

```yaml
actions: "allocate"
tiers:
- plugins:
  - name: nodeorder
```

3. Wait until the reload log shows:

```text
Finished loading scheduler config. Final state: actions=[allocate], plugins=[nodeorder]
```

4. Create a pod with a PVC so `VolumeBinding` `PreBind` is needed.

The `agent-scheduler` panics and restarts:

```text
I0701 04:38:36.346149       1 cache.go:1049] Snapshot updated: generation=4, total nodes num=1, updated nodes num=1
I0701 04:38:36.346226       1 allocate.go:91] There are <1> nodes for task <volcano-p1-test/p1-test-pod>
panic: interface conversion: cache.BindContextExtension is nil, not *predicates.BindContextExtension

goroutine 223 [running]:
volcano.sh/volcano/pkg/scheduler/plugins/predicates.(*PredicatesPlugin).PreBind(...)
        /go/src/volcano.sh/volcano/pkg/scheduler/plugins/predicates/predicates.go:858
volcano.sh/volcano/pkg/agentscheduler/plugins/predicates.(*predicatesPlugin).PreBind(...)
        /go/src/volcano.sh/volcano/pkg/agentscheduler/plugins/predicates/predicates.go:92
volcano.sh/volcano/pkg/agentscheduler/cache.(*SchedulerCache).executePreBind(...)
        /go/src/volcano.sh/volcano/pkg/agentscheduler/cache/cache.go:917
volcano.sh/volcano/pkg/agentscheduler/cache.(*SchedulerCache).BindTask.func1(...)
        /go/src/volcano.sh/volcano/pkg/agentscheduler/cache/cache.go:971
```

This happens because the old `predicates` `PreBinder` remains registered in the cache-level binder registry, while the new framework no longer creates the `predicates` bind context extension.

### Action arguments are not reset after reload

Reproduce steps:

1. Start with:

```yaml
actions: "allocate"
configurations:
- name: allocate
  arguments:
    candidateNodeCount: 10
    predicateErrorCacheEnable: false
tiers:
- plugins:
  - name: predicates
  - name: nodeorder
```

2. Hot reload and remove the `allocate` arguments:

```yaml
actions: "allocate"
tiers:
- plugins:
  - name: predicates
  - name: nodeorder
```

3. The in-memory `allocate` action still keeps:

```text
candidateNodeCount=10
predicateErrorCacheEnable=false
```

instead of restoring the defaults.

I will submit another fix for these shortly.

togettoyou:
Updated the PR to fix the issues mentioned above.

I also verified the related hot reload scenarios locally in Kubernetes.

pkg/agentscheduler/scheduler.go
```diff
@@ -249,18 +302,60 @@ func (sched *Scheduler) loadSchedulerConf() {
...
+
+	if applied || len(actions) == 0 {
 		return
 	}
+	sched.applySchedulerConf(actions, tiers, configurations, metricsConf)
+}
+
+func (sched *Scheduler) applySchedulerConf(actions []framework.Action, tiers []conf.Tier, configurations []conf.Configuration, metricsConf map[string]string) {
+	workers := sched.getWorkers()
+	for _, worker := range workers {
```
JesseStutler: Don't claim all of the locks once, please put lock claiming and unlock into one wrap func and wait all of the previous operation have done. Unless it will easily meet deadlock which should be scary

JesseStutler:
> @JesseStutler
> 
> Sorry, I found two additional issues while re-testing the hot reload behavior.
> 
> ### Binder registry is not refreshed after reload
> Reproduce steps:
> 
> 1. Deploy `agent-scheduler` with `predicates` enabled:
> 
> ```yaml
> actions: "allocate"
> tiers:
> - plugins:
>   - name: predicates
>   - name: nodeorder
> ```
> 
> 2. Hot reload the config and remove `predicates`:
> 
> ```yaml
> actions: "allocate"
> tiers:
> - plugins:
>   - name: nodeorder
> ```
> 
> 3. Wait until the reload log shows:
> 
> ```
> Finished loading scheduler config. Final state: actions=[allocate], plugins=[nodeorder]
> ```
> 
> 4. Create a pod with a PVC so `VolumeBinding` `PreBind` is needed.
> 
> The `agent-scheduler` panics and restarts:
> 
> ```
> I0701 04:38:36.346149       1 cache.go:1049] Snapshot updated: generation=4, total nodes num=1, updated nodes num=1
> I0701 04:38:36.346226       1 allocate.go:91] There are <1> nodes for task <volcano-p1-test/p1-test-pod>
> panic: interface conversion: cache.BindContextExtension is nil, not *predicates.BindContextExtension
> 
> goroutine 223 [running]:
> volcano.sh/volcano/pkg/scheduler/plugins/predicates.(*PredicatesPlugin).PreBind(...)
>         /go/src/volcano.sh/volcano/pkg/scheduler/plugins/predicates/predicates.go:858
> volcano.sh/volcano/pkg/agentscheduler/plugins/predicates.(*predicatesPlugin).PreBind(...)
>         /go/src/volcano.sh/volcano/pkg/agentscheduler/plugins/predicates/predicates.go:92
> volcano.sh/volcano/pkg/agentscheduler/cache.(*SchedulerCache).executePreBind(...)
>         /go/src/volcano.sh/volcano/pkg/agentscheduler/cache/cache.go:917
> volcano.sh/volcano/pkg/agentscheduler/cache.(*SchedulerCache).BindTask.func1(...)
>         /go/src/volcano.sh/volcano/pkg/agentscheduler/cache/cache.go:971
> ```
> 
> This happens because the old `predicates` `PreBinder` remains registered in the cache-level binder registry, while the new framework no longer creates the `predicates` bind context extension.
> 
> ### Action arguments are not reset after reload
> Reproduce steps:
> 
> 1. Start with:
> 
> ```yaml
> actions: "allocate"
> configurations:
> - name: allocate
>   arguments:
>     candidateNodeCount: 10
>     predicateErrorCacheEnable: false
> tiers:
> - plugins:
>   - name: predicates
>   - name: nodeorder
> ```
> 
> 2. Hot reload and remove the `allocate` arguments:
> 
> ```yaml
> actions: "allocate"
> tiers:
> - plugins:
>   - name: predicates
>   - name: nodeorder
> ```
> 
> 3. The in-memory `allocate` action still keeps:
> 
> ```
> candidateNodeCount=10
> predicateErrorCacheEnable=false
> ```
> 
> instead of restoring the defaults.
> 
> I will submit another fix for these shortly.

@togettoyou Sorry I didn't get your point why you mean "while the new framework no longer creates the predicates bind context extension.", also, I don't think `Prebinders` should be added to `BindContext`. Why doesn't your change support reloading at the cache level?

togettoyou:
@JesseStutler Thanks for the review. I think I may have mixed two different levels of hot reload here.

The `remove predicates` case is probably a bit extreme. In real usage, `predicates` is normally expected to stay enabled, so this may not be a case we really need to support.

What I was trying to fix is this mismatch:

`BindContext.Extensions` is created from the current worker framework:

```go
for _, plugin := range fwk.Plugins {
	if handler, ok := plugin.(framework.BindContextHandler); ok {
		state := fwk.CurrentCycleState
		handler.SetupBindContextExtension(state, bindContext)
	}
}
```

but `PreBind` was later picked from the cache-level binder registry:

```go
preBinders := sc.binderRegistry.getRegisteredPreBinders()
successfulPreBindContexts := sc.executePreBinds(cancelCtx, bindContexts, preBinders)
```

So if the plugin set changes during reload, these two may come from different plugin sets. Removing `predicates` just makes the problem easy to reproduce: the new framework does not create the predicates extension anymore, but the old predicates PreBinder may still be called from the cache registry.

But I agree this may be outside the expected hot-reload scope. Do we expect agent-scheduler to support changing the plugin list at runtime, or only changing arguments for the existing actions/plugins?

If changing the plugin list is not expected to be supported, I can drop the `PreBinders` in `BindContext` change and keep this PR smaller. The action argument reset issue still seems worth keeping, because removing action arguments after reload currently leaves the old in-memory values instead of restoring defaults.

pkg/agentscheduler/scheduler.go
togettoyou: Yes, that concern makes sense. I used all worker locks here mainly because actions are currently shared instances from the action registry.

For example, `UnmarshalSchedulerConf()` gets the same action object from the registry:

```go
}
```

and reload calls `OnActionInit()` on that shared object:

```go
alloc.enablePredicateErrorCache = true
alloc.candidateNodeCount = DefaultCandidateNodeCount
alloc.parseArguments(configurations)
```

while a worker may read the same fields during scheduling:

```go
ph.PredicateNodes(..., alloc.enablePredicateErrorCache, ...)
util.SelectBestNodes(..., alloc.candidateNodeCount, ...)
```

So if we simply update workers one by one, reload may mutate the shared action while another worker is still reading it, which can introduce a data race.

I agree holding all worker locks at once is not ideal. I will try to rework this part.

togettoyou:
Updated the PR. I removed the `PreBinders` in `BindContext` change, kept the action argument reset fix, and reworked the worker update path to avoid holding all worker locks at once.

pkg/agentscheduler/framework/plugins.go
```diff
@@ -221,21 +221,33 @@ func (f *Framework) NodeOrderReduceFn(task *api.TaskInfo, pluginNodeScoreMap map
...
 	pluginMutex.RLock()
 	defer pluginMutex.RUnlock()
 
-	act, found := actionMap[name]
-	return act, found
+	ab, found := actionBuilders[name]
+	if !found {
+		return nil, false
+	}
+	return ab(), true
```
togettoyou, line 252: I changed action registration to use an `ActionBuilder`, so each config load gets fresh action instances. Then worker frameworks can be updated one by one without holding all worker locks at once.

PR #5556, opened 2026-06-30
perf: remove unnecessary Clone() in NodeInfo.SetNode

### LINKED ISSUE #5555: [PERF] Avoid unnecessary NodeInfo.Clone() dry-run in SetNode() on every node update

pkg/scheduler/api/node_info.go:SetNode() performs a full NodeInfo.Clone() on every node update as a defensive "dry-run" before applying changes. The cloned object is discarded immediately after validation.

NodeInfo.Clone() is expensive because it reconstructs the entire scheduler state by recreating all TaskInfo objects and deep-copying structures such as NumaInfo, ResourceUsage, NumaSchedulerInfo, and GPU device state.

This clone is executed for every node update event—including periodic heartbeats, CSI updates, and label changes—even when these updates do not require rebuilding the scheduler state.

In large clusters (for example, 2,000 nodes with dozens of pods per node), this results in unnecessary CPU usage, memory allocations, and increased garbage collection pressure. Heap profiles are expected to show NodeInfo.Clone() as a major allocation hotspot during frequent node synchronization.

Avoid performing the defensive NodeInfo.Clone() during every SetNode() call.

Since readiness is determined from the incoming node object, the readiness check can be performed before mutating the existing NodeInfo. A possible implementation is:

func (ni *NodeInfo) SetNode(node *v1.Node) {
    ni.setNodeState(node)
    if !ni.Ready() {
        return
    }

    ni.setNode(node)
    ni.setNodeState(node)
}

This removes the unnecessary deep-copy while preserving the existing readiness validation.

Before adopting this optimization, setNode() should be audited to ensure it does not leave NodeInfo in an inconsistent state if execution is interrupted, since the current clone guarantees atomicity.

### Additional context

Affected file: pkg/scheduler/api/node_info.go
Call path: UpdateNode → SyncNode → AddOrUpdateNode → NodeInfo.SetNode()
Expected benefits:
Reduced CPU usage during node synchronization
Lower heap allocations and GC pressure
Improved scheduler scalability in large clusters with frequent node updates
Suggested validation:
Benchmark SetNode() before and after the change (ns/op, allocs/op, B/op)
Compare heap profiles using pprof
Measure GC statistics using runtime.ReadMemStats
Verify scheduler behavior under large-scale node update workloads (e.g., ~2,000 nodes)

This optimization targets a performance hotspot without changing the scheduler's external behavior, provided setNode() is confirmed to be safe without the defensive clone.

vanshika2720:
## What type of PR is this?

/kind cleanup

`NodeInfo.SetNode()` currently performs a defensive `NodeInfo.Clone()` before applying every node update. The cloned object is used only for a readiness check and is discarded immediately afterward.

After reviewing the implementation, the clone was found to be unnecessary. `setNodeState()` only marks a node as `NotReady` when the input node is `nil`; for any non-nil node it always sets the state to `Ready`. As a result, the readiness check on the cloned `NodeInfo` always succeeds, making the clone redundant.

This PR removes the unnecessary clone while preserving the existing nil-node guard before calling `setNode()`. Since `setNode()` rebuilds the node's resource accounting from the current `Tasks` map, the scheduler behavior remains unchanged while avoiding an expensive deep copy on every node update.

### Changes

- Remove the redundant `NodeInfo.Clone()` call from `SetNode()`
- Preserve the existing nil-node safety check
- Remove the unused dry-run logic
- Reduce unnecessary CPU usage and memory allocations during frequent node updates

Fixes #5555 

## Special notes for your reviewer

- `setNode()` was reviewed to ensure it remains safe without the defensive clone.
- Resource accounting is recomputed from the existing `Tasks` map, making the operation idempotent.
- This change preserves the existing behavior while eliminating unnecessary allocations on node heartbeats, informer resyncs, and other node update events.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -340,26 +340,14 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
 
 // SetNode sets kubernetes node object to nodeInfo object
 func (ni *NodeInfo) SetNode(node *v1.Node) {
-	ni.setNodeState(node)
-	if !ni.Ready() {
+	if node == nil {
+		ni.setNodeState(nil)
 		klog.Warningf("Failed to set node info for %s, phase: %s, reason: %s",
 			ni.Name, ni.State.Phase, ni.State.Reason)
 		return
 	}
-
-	// Dry run, make sure all fields other than `State` are in the original state.
-	copy := ni.Clone()
-	copy.setNode(node)
-	copy.setNodeState(node)
-	if !copy.Ready() {
-		klog.Warningf("SetNode makes node %s not ready, phase: %s, reason: %s",
-			copy.Name, copy.State.Phase, copy.State.Reason)
-		// Set state of node to !Ready, left other fields untouched
-		ni.State = copy.State
-		return
-	}
-
 	ni.setNode(node)
+	ni.setNodeState(node)
 }
 
 // setNodeOthersResource initialize sharable devices
--- a/pkg/scheduler/api/node_info.go
+++ b/pkg/scheduler/api/node_info.go
@@ -340,14 +340,13 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
 
 // SetNode sets kubernetes node object to nodeInfo object
 func (ni *NodeInfo) SetNode(node *v1.Node) {
-	if node == nil {
-		ni.setNodeState(nil)
+	ni.setNodeState(node)
+	if !ni.Ready() {
 		klog.Warningf("Failed to set node info for %s, phase: %s, reason: %s",
 			ni.Name, ni.State.Phase, ni.State.Reason)
 		return
 	}
 	ni.setNode(node)
-	ni.setNodeState(node)
 }
 
 // setNodeOthersResource initialize sharable devices
--- a/pkg/scheduler/api/node_info_test.go
+++ b/pkg/scheduler/api/node_info_test.go
@@ -315,6 +315,29 @@ func BenchmarkSetNode(b *testing.B) {
 	}
 }
 
+// BenchmarkNodeInfoClone measures the per-call cost of Clone() to show what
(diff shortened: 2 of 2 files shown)


vanshika2720:
@JesseStutler
@hajnalmt
@hwdef

JesseStutler:
@vanshika2720 So what's the result after you optimise? For example, the time it takes for `openSession` to create a snapshot—I am referring to actual testing, not unit tests.

pkg/scheduler/api/node_info.go
```diff
@@ -332,26 +332,14 @@ func (ni *NodeInfo) setNodeState(node *v1.Node) {
 
 // SetNode sets kubernetes node object to nodeInfo object
 func (ni *NodeInfo) SetNode(node *v1.Node) {
-	ni.setNodeState(node)
-	if !ni.Ready() {
+	if node == nil {
```
JesseStutler: It doesn't seem necessary to make changes here, `setNodeState` inside itself handles `nil` values.

JesseStutler:
I can accept, but better to see the actual performance improvement results. And besides I have a comment please check

vanshika2720:
@JesseStutler Thanks for the review. I've addressed the requested code change by restoring the original setNodeState(node) + !Ready() flow, since setNodeState already handles nil.

I also added BenchmarkNodeInfoClone using the same setup as BenchmarkSetNode so the cost of the removed clone can be compared directly. On my machine the benchmarks show:

- BenchmarkSetNode: ~3.1 µs/op, 1,296 B/op, 18 allocs/op
- BenchmarkNodeInfoClone: ~12.1 µs/op, 22,712 B/op, 239 allocs/op

I haven't yet collected an end-to-end measurement of openSession snapshot creation time. If that's a requirement for this PR, I'll set up a reproducible benchmark for that as well.

JesseStutler:
/approve
Although merged, @vanshika2720 better to test in a real cluster to measure what are the changes in performance before and after

PR #5558, opened 2026-07-01
fix(vnpu): correct spread scoring logic in `CalScore`

DSFans2014:

There's a problem with the spread scoring in vNPU's `CalScore`.  The spread policy is designed to prefer  idle devices, so those devices should get a higher score.

Fixes #

```release-note

```

### DIFF
--- a/pkg/scheduler/api/devices/ascend/hami/device_info.go
+++ b/pkg/scheduler/api/devices/ascend/hami/device_info.go
@@ -730,12 +730,15 @@ func (dev *AscendDevice) GetResourceNames() devices.ResourceNames {
 }
 
 func CalScore(schedulePolicy string, dev_usage *devices.DeviceUsage, dev_info *devices.DeviceInfo) float64 {
+	if dev_info.Devmem == 0 {
+		return 0
+	}
 	var score float64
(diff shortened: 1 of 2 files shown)


pkg/scheduler/api/devices/ascend/hami/device_info.go
```diff
@@ -730,12 +730,15 @@ func (dev *AscendDevice) GetResourceNames() devices.ResourceNames {
 }
 
 func CalScore(schedulePolicy string, dev_usage *devices.DeviceUsage, dev_info *devices.DeviceInfo) float64 {
+	if dev_info.Devmem == 0 {
```
archlitchi, line 733: which scenario can this branch be triggered, dev_info.Devmem is the device memory limit, so it shouldn't be 0 in most cases?

DSFans2014: Yes, it is usually non-zero.  This just is a defensive measure to prevent division by zero.
```
score = binpackMultiplier * (float64(dev_usage.Usedmem) / float64(dev_info.Devmem))
```

archlitchi (approved):
/lgtm

PR #5559, opened 2026-07-01
webhooks: reject malformed JDB percentages

avinxshKD:
#### What type of PR is this?

/kind bug

Reject malformed JDB percentage annotations in the pod validating webhook.

Values like `50%%` were accepted because the validator removed all `%` chars before parsing. Now percentages must use the normal trailing `%` form.

Fixes: None

NONE

```release-note
Reject malformed JDB percentage annotations in the pod validating webhook.

### DIFF
--- a/pkg/webhooks/admission/pods/validate/admit_pod.go
+++ b/pkg/webhooks/admission/pods/validate/admit_pod.go
@@ -147,13 +147,19 @@ func validateIntPercentageStr(key, value string) error {
 		}
 		return nil
 	case intstr.String:
-		s := strings.Replace(tmp.StrVal, "%", "", -1)
+		if !strings.Contains(tmp.StrVal, "%") {
+			return fmt.Errorf("invalid value <%q> for %v, it must be a positive integer or a percentage between 1%% and 99%%", tmp.StrVal, key)
+		}
+		if strings.Count(tmp.StrVal, "%") != 1 || !strings.HasSuffix(tmp.StrVal, "%") {
+			return fmt.Errorf("invalid value <%q> for %v, it must be a valid percentage between 1%% and 99%%", tmp.StrVal, key)
+		}
+		s := strings.TrimSuffix(tmp.StrVal, "%")
 		v, err := strconv.Atoi(s)
 		if err != nil {
 			return fmt.Errorf("invalid value %v for %v", err, key)
 		}
 		if v <= 0 || v >= 100 {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by avinxshKD** [comment]
Tiny webhook validation fix. This keeps malformed values like `50%%` from being treated as valid percentages. Pls take a look.

File: pkg/webhooks/admission/pods/validate/admit_pod_test.go [hunk 1]
```diff
@@ -132,3 +132,9 @@ func TestValidatePod(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateIntPercentageStr(t *testing.T) {
```
hajnalmt, line 136: This test is really scarce can we extend it please with positive cases and cases like 5%0 and %50, which were failing the previous behaviour too?

pkg/webhooks/admission/pods/validate/admit_pod.go
```diff
@@ -147,7 +147,14 @@ func validateIntPercentageStr(key, value string) error {
 		}
 		return nil
 	case intstr.String:
-		s := strings.Replace(tmp.StrVal, "%", "", -1)
+		if !strings.Contains(tmp.StrVal, "%") {
+			_, err := strconv.Atoi(tmp.StrVal)
```
hajnalmt: Why is Atoi here? To get an error message?

I think just returning the value  is a lot cleaner here (this is just an input after all):
```
if !strings.Contains(tmp.StrVal, "%") {
    return fmt.Errorf("invalid value <%q> for %v, it must be a positive integer or a percentage between 1%% ~ 99%%", tmp.StrVal, key)
}
```

Your message format will be really broken this way here.
  - On failure ("abc"): `invalid value strconv.Atoi: parsing "abc": invalid syntax for <key>`
- On success ("3000000000" overflowing int32 but valid as host int): `invalid value <nil> for <key>`.

The letter is a  reachable path because `intstr.Parse` types it as String with no %, then Atoi parses it fine.

I know its just cosmetics, but volcano's error messages are so hard to track now so it would be cool to be better in these.

hajnalmt (commented):
/area webhook

Thanks, this is a quite little bug, but its an improvement. Can you add a wider test coverage if we touch this? This will reject typos like 5%0 et cetera, which is the right path i think. Also check my remark please.

avinxshKD:
@hajnalmt updated, ptal. kept the parser change local, removed the extra Atoi path for non-percent strings, and added positive/negative cases for the percentage validator

avinxshKD:
@hajnalmt @JesseStutler this has been stale for too long, can u pls take a look when get chance :p

PR #5573, opened 2026-07-03
fix(vnpu): correct node lock key

DSFans2014:

- correct the key of node lock in vnpu.  device-plugin uses `hami.io/mutex.lock` as the node lock key refer [here](https://github.com/Project-HAMi/HAMi/blob/430b458c75c37092b2ea48c8b17bd6d1cfcf45f4/pkg/util/nodelock/nodelock.go#L39)
- remove the `ReleaseNodeLock`.  the  lock should be released in device-plugin

Fixes #

```release-note

```


archlitchi (approved):
/lgtm

Issue #5251, opened 2026-04-28
[LFX 2026 Term 2] Support Namespace-scoped Queue


JesseStutler:
### What is the problem you're trying to solve

Currently, Volcano's Queue resource is cluster-scoped. On multi-tenant platforms, tenants are namespace-scoped and cannot create or manage cluster-wide resources directly. This means only cluster admins can configure hierarchical Queues, making it difficult for tenants to use Volcano's advanced scheduling features independently. As a result, platform teams typically build custom abstractions with their own namespace CRDs and controllers to translate namespace configurations into cluster Queues, creating duplicated work and inconsistent validation. Moreover, any misconfiguration in these wrappers or changes in Volcano's validation logic introduce fragility and potential scheduling issues.

### Describe the solution you'd like

Introduce a new namespace-scoped CRD `NamespaceQueue` that inherits the capabilities of the existing cluster-scoped `Queue`, with the key difference being scope. Specifically:

- A `NamespaceQueue` has the same spec fields and semantics as a cluster `Queue` (e.g., `guarantee`, `deserved`, `capability`), but lives in a namespace.
- `NamespaceQueue` supports hierarchical queues: its parent can be another `NamespaceQueue` in the same namespace, or a cluster `Queue`.
-  PodGroups can be submitted to a `NamespaceQueue`, queue resolution is namespace-first with fallback to cluster `Queue`.
- `NamespaceQueue` maintains its own `.status` (allocated resources, conditions, readiness), giving tenants visibility without cluster-level access.
- Fully backward compatible: clusters not using this feature behave exactly as today.

For the detailed design, see: https://docs.google.com/document/d/19I98jpBHxlo3OLIKtchGvRu-jBQKC9JpDjO0TeV1CdY/edit?tab=t.0

### Documentation Updates

- [x] This feature requires design or user documentation changes.
- [x] If documentation changes are required, I will ensure the relevant documents are updated and published to the Volcano official website (https://volcano.sh) via the [volcano-sh/website](https://github.com/volcano-sh/website) repository.

JesseStutler:
We will consider it as the LFX Term 2 Project first, please do not randomly assign it first without LFX selection

JesseStutler:
Also /cc @devzizu @hajnalmt

Aman-Cool:
[...]

PR #5579, opened 2026-07-03
docs: NamespaceQueue design proposal (#5251)

gitGurugu:
#### What type of PR is this?

/kind feature

This PR adds a design proposal for a namespace-scoped `NamespaceQueue` CRD,
letting tenants author and manage queue hierarchies within their own namespace
without cluster-admin intervention.

The proposal derives `NamespaceQueue` from the cluster-scoped `Queue` (reusing
`QueueSpec`, so guarantee/deserved/capability semantics are unchanged). It uses a
namespace-qualified `QueueID` and namespace-first `PodGroup.spec.queue`
resolution, so the `capacity` and `proportion` plugins need no behavioral change.
The whole feature sits behind a default-off `NamespaceQueue` feature gate, so
clusters that do not opt in are unchanged.

This is a **draft** opened to align on the design (see the `Open questions`
section) before implementation. It contains the design doc plus the feature-gate
registration only — no CRD/controller/scheduler code yet.

Related to #5251

- This is a design-proposal draft PR; I'd especially appreciate feedback on the
  `Open questions for review` section (parent disambiguation, mid-flight
  re-resolution, status struct, RBAC packaging).
- AI disclosure: I used an AI coding assistant to help draft and structure the
  proposal document. I have reviewed and understand every part of it, and will
  respond to review comments myself.

```release-note
NONE

### DIFF
--- a/docs/design/namespace-queue.md
+++ b/docs/design/namespace-queue.md
@@ -62,8 +62,9 @@ namespaces may reference the same cluster `Queue`, and a namespace may host
 - Replacing or wrapping the cluster-scoped `Queue`. `NamespaceQueue` is
   **additive**; the current model is unchanged for clusters that do not opt in.
 - Introducing new tenant-isolation semantics.
-- Changing the meaning of `guarantee`, `deserved`, or `capability`. This proposal
-  does **not** alter how resource fields are defined or interpreted today.
+- Changing the meaning of `guarantee`, `capability`, or `weight`. The fields
+  `NamespaceQueueSpec` carries over from `Queue` keep their current definition and
+  interpretation; this proposal does not alter existing resource semantics.
 
 ## User Stories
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/design/namespace-queue.md [hunk 1]
```diff
@@ -0,0 +1,284 @@
...
+// +kubebuilder:resource:scope=Namespaced,shortName=nq
+
+// NamespaceQueue is the namespace-scoped counterpart of Queue.
+type NamespaceQueue struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    // Spec reuses QueueSpec; spec.parent references either a NamespaceQueue in
+    // the same namespace or a cluster Queue (resolution is namespace-first).
+    Spec   QueueSpec            `json:"spec,omitempty"`
```
JesseStutler: So expand the spec and introduce the fields within it

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,284 @@
...
+tree, so it cannot disturb other tenants' scheduling.
+
+### Story 3
+
+As an existing user, I want my current cluster `Queue`s and workloads to behave
+exactly as before if I don't enable this feature.
+
+## Design detail
+
+### The `NamespaceQueue` CRD
```
JesseStutler, line 10: We remember that we have discussed there are also some fields should added in the existing queue crd like allowedNamespaces list, should also introduced here

JesseStutler (commented):
@gitGurugu At a glance, it looks too AI-generated and simplistic; more details need to be shown

hajnalmt:
/cc @devzizu @hajnalmt 

Super! I want to check this tomorrow.

docs/design/namespace-queue.md
gitGurugu: Yes, I have already incorporated this into the updated design. I added the `allowedNamespaces` field to the Queue CRD and introduced the corresponding validation logic in the controller.The controller will check this field when a NamespaceQueue references a cluster-level Queue, and only allow the binding if the tenant Namespace is authorized.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+
+    Spec   NamespaceQueueSpec   `json:"spec,omitempty"`
+    Status NamespaceQueueStatus `json:"status,omitempty"`
+}
+
+type NamespaceQueueSpec struct {
+    // Parent points to either a local NamespaceQueue in the same Namespace
+    // or an authorized cluster-level Queue.
+    // +kubebuilder:validation:Required
+    Parent string `json:"parent"`
```
devzizu, line 33: I believe here we need to be more clear about how we want this string to be formatted and what are the validation rules. We actually settled this in a previous meeting, let's use the `cluster/<name>` prefix convention for a cluster parent, default an unset parent to `cluster/default`, and **reject root** as an explicit parent. Please document that resolution rule explicitly here, including the default.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+		}
+	}
+	return false
+}
+```
+
+Step 3: Translate and Clone Queue
+
+Design Intent:
+Convert the validated `NamespaceQueue` into a cluster-level Volcano `Queue`.
```
devzizu, line 316: I was not considering that a NSQ would generate a Volcano cluster-scoped Queue. Actually, I was considering that a NSQ is a separate object with its identity within the Volcano project types, but I guess that was not a clear design path established since the beginning and we can discuss the pros and cons of it. It does simplify things if we consider Queues instead of NSQs, but it could generate name collisions, weakens observability of the namespace objects, require a status round-trip, etc.

Let's align on the path forward cc @hajnalmt @JesseStutler

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+```go
+func (r *NamespaceQueueReconciler) validateNamespaceQueue(ctx context.Context, nsq *platformv1alpha1.NamespaceQueue) error {
+	currentNS := nsq.Namespace
+	parentName := nsq.Spec.Parent
+
+	// 1. Check if parent is a local NSQ within the same namespace (Sub-tree validation)
+	parentNSQ := &platformv1alpha1.NamespaceQueue{}
+	err := r.Get(ctx, client.ObjectKey{Namespace: currentNS, Name: parentName}, parentNSQ)
+	if err == nil {
+		if r.isResourceExceeded(nsq.Spec.Capability, parentNSQ.Spec.Capability) {
```
devzizu, line 262: I believe we are simplifying a bit here, leaving room for some bugs, for example: (1) only compares `Capability`, but the hierarchy invariant also covers deserved/guarantee; (2) treating "resource type absent in parent" as a violation will reject legitimate configs where the parent doesn't pin every resource type; (3) this is per-queue only and it doesn't account for the sum of siblings.

Can you please improve this section with those considerations?

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+Use the Job namespace as the tenant context, then rewrite the local queue name to its mapped physical Queue name according to the platform’s queue mapping rule.
+
+
+Detailed Mutation Logic:
+
+- The webhook reads `req.Namespace` as the tenant context.
+
+- It then reads `job.Spec.Queue` as the local queue name. If this field is empty, the request is rejected because tenants must explicitly specify a valid `NamespaceQueue`.
+
+- If `job.Spec.Queue` is provided, the webhook rewrites it to the physical queue name using the rule: `[Namespace]-[LocalQueueName]`.
```
devzizu, line 454: Rewriting the user's `spec.queue` in place will fight GitOps controllers with continuous drifts. It also only covers `Job`, but we have `PodGroup`s or the `scheduling.volcano.sh/queue-name` pod annotation.

Can we review this part?

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+```
+
+Step 3: Translate and Clone Queue
+
+Design Intent:
+Convert the validated `NamespaceQueue` into a cluster-level Volcano `Queue`.
+
+Implementation Details:
+
+- The controller generates a globally unique physical Queue name using the format `[Namespace]-[NSQ-Name]` to avoid naming conflicts.
```
devzizu, line 320: Be careful here with this claim: as it's not collision-free: namespace a-b/queue c collides with namespace a/queue b-c. If we ever needed a synthetic name we'd need a reserved separator or hash.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+    Weight int32 `json:"weight,omitempty"`
+
+}
+```
+
+```go
+type NamespaceQueueStatus struct {
+
+  // +optional
+  Phase string `json:"phase,omitempty" protobuf:"bytes,1,opt,name=phase"`
```
devzizu, line 65: Phase as a free-form string ("Admitted"/"Rejected") duplicates `State` and isn't machine-friendly. Prefer standard `metav1.Conditions` (e.g., a `Ready`/`Validated` condition with reason) plus the existing counters, matching how `QueueStatus` works today.

devzizu (commented):
Thanks for pushing this forward in this first draft @gitGurugu! Before we go deeper on details, I want to flag a direction concern, because I think we still have a few design decisions to align. Specially the need of considering two objects (a generation model), which is a permanent consistency concern.

Separately, I think the doc is currently code-first and should focus a bit more on the rationale / design choices. I would recommend adding Goals / Non-goals, and Alternatives Considered.

docs/design/namespace-queue.md
hajnalmt: ++ devzizu

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+Design Intent:
+Use NSQ events as the entry point of the control loop and determine whether the object still exists or is being deleted.
+
+Logic Flow:
+
+- The controller receives an NSQ reconcile request and fetches the NSQ object by namespace and name.
+
+- If the NSQ no longer exists, it means the object has already been deleted from the cluster. The controller exits safely without further processing.
+
+- If the NSQ is being deleted, Kubernetes garbage collection will clean up the corresponding physical Queue through the previously set OwnerReference. Therefore, no extra cleanup logic is required in the controller.
```
hajnalmt, line 162: NamespaceQueue shall not have owner references and it shouldn't be shadowed behind a cluster queue.

In the schedulers internal accounting and everywhere internally you can refer to a namespaced queue with 'namespace/queue_name' this will avoid collisions in general because a cluster queue cannot have '/' in its name.

Garbage collection is not an issue this way.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+			Parent:          fmt.Sprintf("%s-%s", nsq.Namespace, nsq.Spec.Parent), //有问题
+		},
+	}
+}
+```
+
+Step 4: Reconcile Queue and Sync Status
+
+Design Intent:
+Keep the physical Volcano `Queue` consistent with the tenant `NamespaceQueue`, and expose the backend queue status back to tenants.
```
hajnalmt, line 357: I don't understand why you say physical queue 🗡️nothing is "physical" here.

docs/design/namespace-queue.md
hajnalmt: This is not what we discussed on the last sync. We don't need this syntetic name and controller generation at all. NamespaceQueue is a separate resource, everywhere we count with internally next to cluster-wide Queues we can just use a `<namespace>/<queue-name>` form and name collisions are avoided as cluster queues cannot have a "/" character in their name.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+
+Reconciliation Mechanism:
+
+- The controller submits the physical `Queue` to the API Server.
+
+- If the physical `Queue` does not exist, the controller creates it.
+
+- If it already exists, the controller updates its `Spec` with the latest computed configuration, so tenant quota changes can take effect in the backend scheduler.
+
+- By declaring `Owns(&volcano.Queue{})`, the controller can detect when a physical `Queue` is accidentally deleted and recreate it automatically.
```
hajnalmt, line 367: Lets strip this ownership out.

hajnalmt (commented):
Let's discuss this on the meeting tomorrow.

docs/design/namespace-queue.md
gitGurugu: I means this is the true queue

docs/design/namespace-queue.md
gitGurugu: I got it.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+		return ctrl.Result{}, nil
+	} else if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Scenario B: Backend Queue exists -> Update (Enforce Eventual Consistency)
+	logger.Info("Updating backend physical Volcano Queue spec", "queueName", backendQueueName)
+	expectedQueue := r.buildVolcanoQueue(nsq, backendQueueName)
+	
+	foundQueue.Spec = expectedQueue.Spec
```
hzxuzhonghu, line 219: This update is unconditional, and the owned Queue is watched by this controller. Every successful reconciliation writes the Queue again and immediately schedules another reconciliation; `syncQueueStatusToNSQ` also writes status unconditionally. Compare desired and current values and patch only when something changed.

hzxuzhonghu (commented):
Several design gaps need to be resolved before implementation.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+Design Intent:
+Convert the validated `NamespaceQueue` into a cluster-level Volcano `Queue`.
+
+Implementation Details:
+
+- The controller generates a globally unique physical Queue name using the format `[Namespace]-[NSQ-Name]` to avoid naming conflicts.
+
+- The controller copies scheduling fields from `NSQ.Spec` into Volcano `QueueSpec`, including `Weight`, `Capability`, `Deserved`, `Guarantee`, `Priority`, and `DequeueStrategy`.
+
+- The controller also handles the parent queue mapping. If the parent is another `NamespaceQueue` in the same Namespace, it rewrites the parent name to the corresponding physical Queue name.
```
hzxuzhonghu, line 324: Rewriting the parent does not make the hierarchy effective for the proportion plugin: proportion never reads `Queue.spec.parent` and gives every generated child a cluster-wide weighted share. Multiple NamespaceQueues can therefore exceed their parent tenant's share. The design needs hierarchical proportion accounting or an explicit restriction to capacity.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+	nsq.Status.Running = q.Status.Running
+	nsq.Status.Inqueue = q.Status.Inqueue
+	nsq.Status.Completed = q.Status.Completed
+	nsq.Status.Reservation = platformv1alpha1.Reservation(q.Status.Reservation)
+	nsq.Status.Allocated = q.Status.Allocated
+
+	return r.Status().Update(ctx, nsq)
+}
+
+func (r *NamespaceQueueReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
hzxuzhonghu, line 390: Validity depends on the parent NamespaceQueue and on a cluster Queue's `allowedNamespaces`, but this setup watches only the object itself and its owned backend Queue. Parent fixes or revocations will not reconcile children, leaving rejected children stuck or unauthorized queues active. Add indexed watches from both dependencies to their children.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+
+Design Intent:
+Use the Job namespace as the tenant context, then rewrite the local queue name to its mapped physical Queue name according to the platform’s queue mapping rule.
+
+
+Detailed Mutation Logic:
+
+- The webhook reads `req.Namespace` as the tenant context.
+
+- It then reads `job.Spec.Queue` as the local queue name. If this field is empty, the request is rejected because tenants must explicitly specify a valid `NamespaceQueue`.
```
hzxuzhonghu, line 452: This changes existing queue semantics when the webhook is enabled. An omitted queue currently defaults to `default`, and a bare name currently refers to a cluster-scoped Queue; this rule rejects the first and rewrites the second. Please define a backward-compatible resolution rule, such as explicit disambiguation or cluster fallback.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,501 @@
...
+	localQueueName := job.Spec.Queue
+
+	if localQueueName == "" {
+		return admission.Denied(fmt.Sprintf(
+			"multi-tenancy policy violation: you must explicitly specify a 'spec.queue' in namespace '%s'", 
+			tenantNamespace,
+		))
+	}
+
+	globalQueueName := fmt.Sprintf("%s-%s", tenantNamespace, localQueueName)
```
hzxuzhonghu, line 469: The handler also runs on UPDATE, so this transformation is not idempotent. A stored value such as `ns-queue-a` becomes `ns-ns-queue-a` on the next Job update. Resolve only on creation or preserve enough information to recognize an already-resolved queue.

docs/design/namespace-queue.md
gitGurugu: Thanks for catching this. You are right—this unconditional update, together with watching the Queue and unconditionally updating the NSQ status, can cause repeated reconciliations. This PR is based on an earlier version of the design and does contain this issue.The latest design has been moved to #5629: https://github.com/volcano-sh/volcano/pull/5629. I’ll update or close this outdated PR to avoid further confusion.

PR #5584, opened 2026-07-05
scheduler: check Elasticsearch response status before decoding metrics

anxkhn:
#### What type of PR is this?

/kind bug

`NodeMetricsAvg` only checked the transport error returned by `es.Search`. The
go-elasticsearch client returns a nil transport error for HTTP 4xx/5xx responses,
so Elasticsearch error bodies were decoded as successful metrics with zero CPU
and memory values.

This change returns an error for HTTP failures and includes the node name, status,
and a response body bounded to `maxBodySize`. Transport and response decode errors
now include the node name as well. Tests cover the status, Elasticsearch reason,
node context, and response body bound.

Fixes #

The admission webhook E2E failure on the previous revision was unrelated. It
failed because `test-queue-open` already existed in two admission tests. This PR
only changes the scheduler Elasticsearch metrics client and its unit tests.

Verified with `make verify`, `make lint`,
`go vet ./pkg/scheduler/metrics/source/...`, and
`go test ./pkg/scheduler/metrics/...`.

Disclosure per the AI Guidance in `contributing.md`: I used an AI assistant while
preparing this change. I reviewed and verified every line myself, understand why
each change is here, and can explain and defend it in review.

```release-note
scheduler: Elasticsearch metrics errors now include bounded response details and node context instead of reporting zero utilization.
```

### DIFF
--- a/pkg/scheduler/metrics/source/metrics_client_elasticsearch.go
+++ b/pkg/scheduler/metrics/source/metrics_client_elasticsearch.go
@@ -22,6 +22,8 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"io"
 	"net/http"
 	"time"
 
@@ -146,9 +148,16 @@ func (e *ElasticsearchMetricsClient) NodeMetricsAvg(ctx context.Context, nodeNam
 		e.es.Search.WithBody(&buf),
 	)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("elasticsearch search request failed for node %s: %w", nodeName, err)
 	}
 	defer res.Body.Close()
+	if res.IsError() {
+		body, err := io.ReadAll(io.LimitReader(res.Body, maxBodySize))
+		if err != nil {
+			return nil, fmt.Errorf("elasticsearch search request failed for node %s: %s, failed to read response body: %w", nodeName, res.Status(), err)
+		}
+		return nil, fmt.Errorf("elasticsearch search request failed for node %s: %s: %s", nodeName, res.Status(), body)
+	}
 	var r struct {
 		Aggregations struct {
 			CPU struct {
@@ -161,7 +170,7 @@ func (e *ElasticsearchMetricsClient) NodeMetricsAvg(ctx context.Context, nodeNam
 	}
 	res.Body = http.MaxBytesReader(nil, res.Body, maxBodySize)
 	if err := json.NewDecoder(res.Body).Decode(&r); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to decode Elasticsearch metrics response for node %s: %w", nodeName, err)
 	}
 	// The data obtained from Elasticsearch is in decimals and needs to be multiplied by 100.
 	nodeMetrics.CPU = r.Aggregations.CPU.Value * 100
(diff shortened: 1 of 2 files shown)


pkg/scheduler/metrics/source/metrics_client_elasticsearch.go
```diff
@@ -149,6 +150,9 @@ func (e *ElasticsearchMetricsClient) NodeMetricsAvg(ctx context.Context, nodeNam
 		return nil, err
 	}
 	defer res.Body.Close()
+	if res.IsError() {
+		return nil, fmt.Errorf("elasticsearch search request failed for node %s: %s", nodeName, res.Status())
+	}
```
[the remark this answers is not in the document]
hajnalmt: Geminis suggestion is useful here because `res.String()` returns the body too which is I think crucial in these messages. It will have log lines like this instead:
```
{"error":{"reason":"all shards failed"},"status":500}
```

pkg/scheduler/metrics/source/metrics_client_elasticsearch_test.go
```diff
@@ -87,3 +87,44 @@ func TestElasticsearchMetricsClientNodeMetricsAvg_MaxBodySizeExceeded(t *testing
 		}
 	}
 }
+
+func TestElasticsearchMetricsClientNodeMetricsAvg_ErrorResponse(t *testing.T) {
```
hajnalmt, line 89: This test is fine if we can improve the message we can extend it though!

pkg/scheduler/metrics/source/metrics_client_elasticsearch.go
```diff
@@ -149,6 +150,9 @@ func (e *ElasticsearchMetricsClient) NodeMetricsAvg(ctx context.Context, nodeNam
 		return nil, err
 	}
 	defer res.Body.Close()
+	if res.IsError() {
```
hajnalmt, line 154: This is a good fix thanks, but I would improve this at several points if we touch the code, because I miss the node name in these messages and the error context (as Gemini suggested below).

Maybe we can read up to maxBodySize from the error response body and include it in the returned error? The unit test could also assert that the returned error contains the Elasticsearch reason, e.g. all shards failed, not only 500.

```
 if res.IsError() {
  body, readErr := io.ReadAll(io.LimitReader(res.Body, maxBodySize))
  if readErr != nil {
   return nil, fmt.Errorf("elasticsearch search request failed for node %s: %s, and failed to read response body: %v", nodeName, res.Status(), readErr)
  }
  return nil, fmt.Errorf("elasticsearch search request failed for node %s: %s: %s", nodeName, res.Status(), string(body))
 }
```

hajnalmt (commented):
/kind bug
/area scheduling

Good catch from Gemini in principle: `res.Status()` confirms the HTTP failure, but it hides the Elasticsearch error body, which often contains the useful reason such as missing index, query error, authorization failure, or shard failure.

I would prefer including the response body in the returned error, but bounded similarly to the successful decode path so an unexpectedly large error response is not read fully into memory. Something like reading up to `maxBodySize` before returning the error would make debugging better while keeping the existing size-safety intent.

pkg/scheduler/metrics/source/metrics_client_elasticsearch.go
anxkhn: Implemented, including bounded body details, node context, and coverage.

pkg/scheduler/metrics/source/metrics_client_elasticsearch_test.go
anxkhn: Extended the assertions and added bounded-body coverage.

PR #5585, opened 2026-07-06
fix: use queue UID consistently as map key for preemption/reclamation (#5479)

### LINKED ISSUE #5479: gangPreempt, preempt, and reclaim actions silently fail due to Queue Name vs UID map key mismatch
### Description
There is a bug in Volcano's preemption (preempt, gangPreempt) and reclamation (reclaim) actions.  

The Problem: When the scheduler saves jobs in its internal map, it uses the Queue Name (e.g., "default") as the key. But when it tries to find those jobs later in the loop, it searches using the Queue UID (e.g., "a9013f5c-...").  

bug: Because a Name (string) and a Unique ID (UUID string) are completely different, the lookup always returns nothing (nil).  

The Impact: The scheduler silently skips all preemption and reclamation logic without throwing any errors. High-priority jobs get stuck in the queue forever, even if they should easily kick out low-priority jobs.

### The bug spans multiple files:

pkg/scheduler/actions/gangpreempt/gangpreempt.go: Stores via job.Queue but retrieves via queue.UID.

pkg/scheduler/actions/preempt/preempt.go: Stores preemptorsMap and underRequestByQueue via job.Queue but retrieves via queue.UID.

pkg/scheduler/actions/reclaim/reclaim.go: Stores via job.Queue but retrieves via queue.UID.

Deploy Volcano with the capacity plugin configured in the scheduler.
Create a single queue with 100 CPUs capacity and fill it completely with a low-priority workload (Priority=10).
Submit a high-priority batch workload (Priority=1000) requiring 50 CPUs to the same queue.
Enable either gangPreempt or the legacy preempt action.
Observe that the high-priority job remains in a starving/pending state indefinitely. The scheduler logs will cycle endlessly without ever triggering any preemption or evicting any low-priority pods.

Expected results: When a queue is saturated by low-priority tasks and a starving high-priority task arrives, the preempt or gangPreempt action should successfully fetch the high-priority job from preemptorsMap[queue.UID] and trigger preemption of the low-priority tasks.

Received results: The lookup preemptorsMap[queue.UID] returns nil because the map was populated using job.Queue (Queue Name) instead of queue.UID (Queue UUID). Because of this mismatch, the inner loop instantly hits a break statement, skipping all preemption and reclamation logic silently.

Volcano v1.15.0

_No response_

vivek41-glitch:
#### What type of PR is this?
/kind bug

This PR fixes a critical bug in Volcano's preemption and reclamation logic where map keys were inconsistent. The scheduler was storing jobs using Queue Name (job.Queue) but retrieving them using Queue UID (queue.UID). This caused the lookup to return nil, causing preemption to silently fail. High-priority jobs would remain pending even when low-priority jobs should have been preempted.

This fix changes the storage logic in gangpreempt, preempt, and reclaim actions to consistently use job.QueueUID as the map key, matching the retrieval logic which already correctly uses queue.UID.

Fixes #5479

#### Special notes for  reviewer:
This is a focused fix that only changes 3 files (gangpreempt, preempt, reclaim). The changes are minimal and address the root cause of the issue. The Windows build failure (unix.Umask) is pre-existing and not related to this PR.

```release-note
Fixed preemption and reclamation actions that were silently failing due to queue name vs UID map key mismatch. High-priority jobs can now properly preempt low-priority jobs.
Signed-off-by: viveksantoshkumardubey@gmail.com

### DIFF
--- a/pkg/scheduler/actions/gangpreempt/gangpreempt.go
+++ b/pkg/scheduler/actions/gangpreempt/gangpreempt.go
@@ -104,10 +104,11 @@ func (gp *Action) Execute(ssn *framework.Session) {
 		if !ssn.JobStarving(job) {
 			continue
 		}
-		if _, found := preemptorsMap[job.Queue]; !found {
-			preemptorsMap[job.Queue] = util.NewPriorityQueue(ssn.JobOrderFn)
+		queue := ssn.Queues[job.Queue]
+		if _, found := preemptorsMap[queue.UID]; !found {
+			preemptorsMap[queue.UID] = util.NewPriorityQueue(ssn.JobOrderFn)
 		}
-		preemptorsMap[job.Queue].Push(job)
+		preemptorsMap[queue.UID].Push(job)
 	}
 
 	for !queues.Empty() {
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -136,11 +136,12 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 			continue
 		}
 
-		if _, found := preemptorsMap[job.Queue]; !found {
-			preemptorsMap[job.Queue] = util.NewPriorityQueue(ssn.JobOrderFn)
+		queue := ssn.Queues[job.Queue]
+		if _, found := preemptorsMap[queue.UID]; !found {
+			preemptorsMap[queue.UID] = util.NewPriorityQueue(ssn.JobOrderFn)
 		}
-		preemptorsMap[job.Queue].Push(job)
-		underRequestByQueue[job.Queue] = append(underRequestByQueue[job.Queue], job)
+		preemptorsMap[queue.UID].Push(job)
+		underRequestByQueue[queue.UID] = append(underRequestByQueue[queue.UID], job)
 		preemptorTasks[job.UID] = util.NewPriorityQueue(ssn.TaskOrderFn)
 		for _, task := range job.TaskStatusIndex[api.Pending] {
 			if task.SchGated {
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -88,10 +88,11 @@ func (ra *Action) Execute(ssn *framework.Session) {
 		}
 
 		if ssn.JobStarving(job) {
-			if _, found := preemptorsMap[job.Queue]; !found {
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by vivek41-glitch** [comment]
@shinytang6 @william-wang 

Fixes #5479 - preemption/reclamation map key mismatch. Changed storage to use `queue.UID` to match retrieval.

PTAL when you have time. Thanks!

**Comment by vivek41-glitch** [comment]
@holdenk @caniszczyk @hajnalmt

File: pkg/scheduler/actions/gangpreempt/gangpreempt.go [hunk 1]
```diff
@@ -104,10 +104,11 @@ func (gp *Action) Execute(ssn *framework.Session) {
 		if !ssn.JobStarving(job) {
 			continue
 		}
-		if _, found := preemptorsMap[job.Queue]; !found {
-			preemptorsMap[job.Queue] = util.NewPriorityQueue(ssn.JobOrderFn)
+		queue := ssn.Queues[job.Queue]
+		if _, found := preemptorsMap[queue.UID]; !found {
```
hajnalmt, line 108: `job.Queue` and `queue.UID`  are the same. This PR does nothing.

hajnalmt (changes requested):
/hold

Thank you for the PR!
I think it does nothing. `job.Queue` and `queue.UID`  are the same.

Evidence:
- QueueInfo.UID = QueueID(queue.Name): https://github.com/volcano-sh/volcano/blob/09f06d9c363f6ca7c84dbf4aa045e74063625a87/pkg/scheduler/api/queue_info.go#L56
- ji.Queue = QueueID(pg.Spec.Queue): https://github.com/volcano-sh/volcano/blob/09f06d9c363f6ca7c84dbf4aa045e74063625a87/pkg/scheduler/api/job_info.go#L457
- snapshot.Queues[value.UID] = value.Clone(): https://github.com/volcano-sh/volcano/blob/09f06d9c363f6ca7c84dbf4aa045e74063625a87/pkg/scheduler/cache/cache.go#L1529

Please raise your objection, if I am not right. Otherwise close the PR.

vivek41-glitch:
@hajnalmt thanks for review ,and i did verify and yes u r right , this pr is worthless . i was mislead by issue description mismatch between "Queue Name" and "Queue UID" as different strings. and actually did'nt verify properly , that cause everyone wasting time . no worries 
 so, The real question is :
 1.If the map keys are the same, why does `preemptorsMap[queue.UID]` return `nil`?
 2.Is there a race condition where `queue.UID` is empty?
3.Can you share debug logs or a reproduction setup?
 happy to closing  this pr , and ready to investigate in deep ,and finding actual bug .

PR #5588, opened 2026-07-06
fix(vgpu): sync TaskInfo annotations during bind lifecycle

### LINKED ISSUE #5532: deviceshare/vGPU: non-transactional allocation annotations can drift from the bound node and corrupt device-plugin allocation candidates
### Description

In the Volcano vGPU scheduling path, `deviceshare` / vGPU scheduler passes allocation results to the node-local device plugin through Pod annotations. The current annotation and node-lock lifecycle is not transactional enough. Under high concurrency, scheduling rollback, node-lock disabled, node-lock timeout, or lock acquisition failure, the scheduler-side state can drift from the final Kubernetes binding.

This can lead to:

- `spec.nodeName` and `volcano.sh/vgpu-node` disagreeing;
- stale or inconsistent `volcano.sh/devices-to-allocate`, `volcano.sh/vgpu-ids-new`, and `volcano.sh/bind-phase`;
- node-local device-plugin building a wrong candidate set and selecting the wrong Pod UID.

The downstream `volcano-vgpu-device-plugin` may then mount a wrong vGPU cache directory into `/tmp/vgpu`. If that backing directory is later removed, the running container gets a stale `S_DEAD` mount and libvgpu fails to create/open its shared-memory cache file with `ENOENT`.

## Reproduction / diagnostics

1. Submit many half-GPU vGPU Pods onto 8-GPU 3090 nodes, up to about 16 Pods per node.
2. Run:
[scan-drift.sh](https://github.com/106umao/106umao/commit/bcf477458767da6b584aa8f5a2f4546e6b21d3e0)
   ```bash
   bash scan-drift.sh --all --gpu 3090
   ```

3. Inspect:
   - `HEALTHY` / `BORROW_ALIVE` / `BORROW_ORPHAN`;
   - same real-node and claimed-node candidate pools;
   - run-count vs claim-count mismatch;
   - whether `hamivgpu` node locks are present.

### `vgpu-node` is a second node truth

Kubernetes `spec.nodeName` is the final bound node. `volcano.sh/vgpu-node` is a scheduler-written annotation. They should be consistent, but rollback, lock failure, patch delay, or incomplete cleanup can make them drift.

Once drifted, the device plugin can include wrong Pods or exclude the real Pod when building candidates.

### vGPU `Release()` is a no-op

`hpc-volcano/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go` currently contains:

```go
func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) error {
    return nil
}
```

Scheduling rollback and deallocation therefore may not compensate annotations, locks, or device accounting.

### Timestamp-only node lock cannot carry Pod identity

The current lock only stores a timestamp. It can throttle same-node allocation but cannot tell the device plugin which Pod is being processed. HAMi upstream has evolved its lock format to carry Pod namespace/name and uses it to identify the allocation owner.

### Node-level locking has scalability costs

HAMi community issues discuss node-lock performance bottlenecks under high concurrency: immediate lock failures, scheduler backoff, API server patch pressure, retry storms, and 5-minute waits. Enabling node lock is a mitigation, not a complete fix.

expected spec.nodename always equals vgpu-node

v1.13.1

## Environment

Sanitized production-like scale:

- 98 NVIDIA 3090 nodes, 8 GPUs per node;
- half-GPU vGPU workloads are common;
- peak platform workload can submit around `16 * 98` Pods concurrently;
- components:
  - Volcano scheduler `v1.15.0`, with `deviceshare` vGPU enabled;
  - `volcano-vgpu-device-plugin` `v1.12.0`;
  - libvgpu/HAMi-core in workload containers;
- current production deployment did not enable the timestamp-only `hamivgpu` node lock. Enabling it is expected to reduce drift, but the HAMi community has documented significant performance, queuing, backoff, and API server pressure issues with node-level locking.

## Background / mechanism

The following annotations are used in the current vGPU scheduling and binding path:

| Annotation | Writer | Purpose |
|---|---|---|
| `volcano.sh/vgpu-node` | scheduler vGPU `Allocate` | scheduler-side vGPU allocation node |
| `volcano.sh/vgpu-time` | scheduler vGPU `Allocate` | ordering key used by device-plugin candidate selection |
| `volcano.sh/vgpu-ids-new` | scheduler vGPU `Allocate` | full device allocation result |
| `volcano.sh/devices-to-allocate` | scheduler vGPU `Allocate` | queue consumed by device-plugin |
| `volcano.sh/bind-phase` | scheduler/device-plugin | `allocating` / `success` / `failed` state |
| `volcano.sh/bind-time` | scheduler vGPU `Allocate` | binding start timestamp |
| `hamivgpu` node annotation | scheduler/device-plugin | timestamp-only node lock in this implementation |

Relevant source files:

- `hpc-volcano/pkg/scheduler/plugins/deviceshare/deviceshare.go`: registers deviceshare predicate and node-order functions;
- `hpc-volcano/pkg/scheduler/plugins/predicates/predicates.go`: calls `devices.Allocate(...)` and `devices.Release(...)` in allocation/deallocation event handlers;
- `hpc-volcano/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go`: writes vGPU allocation annotations; current `Release()` is a no-op;
- `hpc-volcano/pkg/scheduler/plugins/util/nodelock/nodelock.go`: timestamp-only node lock with a 5-minute timeout.

## Evidence

The original workload logs show libvgpu/HAMi-core failing to create/open the shared-memory cache file at runtime. The following preserves the important error shape with PID and cache UUID sanitized:

```text
[HAMI-core ERROR (pid=<worker-pid> thread=<thread-id> multiprocess_memory_limit.c:1092)]:
Fail to open shrreg /tmp/vgpu/<cache-uuid>.cache: errno=2

[HAMI-core ERROR (pid=<worker-pid> thread=<thread-id> multiprocess_memory_limit.c:1097)]:
Fail to init shrreg /tmp/vgpu/<cache-uuid>.cache: errno=9

[HAMI-core ERROR (pid=<worker-pid> thread=<thread-id> multiprocess_memory_limit.c:1101)]:
Fail to write shrreg /tmp/vgpu/<cache-uuid>.cache: errno=9

[HAMI-core ERROR (pid=<worker-pid> thread=<thread-id> multiprocess_memory_limit.c:1104)]:
Fail to reseek shrreg /tmp/vgpu/<cache-uuid>.cache: errno=9

[HAMI-core ERROR (pid=<worker-pid> thread=<thread-id> multiprocess_memory_limit.c:1114)]:
Fail to lock shrreg /tmp/vgpu/<cache-uuid>.cache: errno=9
```

`errno=2` means the container's `/tmp/vgpu` backing directory is missing or stale (`S_DEAD`). The follow-up `errno=9` lines are cascading EBADF errors from an older libvgpu continuing with an invalid fd after `open` failed.

Sanitized diagnostic command:

```bash
bash scan-drift.sh --all --gpu 3090 --jobs 2
```

Scan scope:

- about 5k Pods in the cluster;
- 190 nodes;
- 98 matched 3090 nodes;
- 800 vGPU containers in the target workload namespace;
- all 800 containers scanned.

Classification:

| Class | Count | Meaning |
|---|---:|---|
| `HEALTHY` | 495 | `/tmp/vgpu` backing UID equals the current Pod UID |
| `BORROW_ALIVE` | 213 | container borrows a still-alive Pod's cache directory |
| `BORROW_ORPHAN` | 89 | container borrows a deleted Pod's orphan cache directory |
| `DEAD` | 0 | borrowed directory already removed |
| `SELF_DEAD` | 0 | own backing directory already removed |
| `NOMOUNT` | 0 | container is reachable but has no `/tmp/vgpu` mount |
| `EXECFAIL` | 3 | `kubectl exec` failed |

Among 797 inspectable containers, 302 borrowed another Pod's cache directory, about 37.9%. 89 containers borrowed orphan directories from deleted Pods, about 11.2%.

Sanitized candidate-pool evidence:

```text
same real node + same claimed node:
  count=16, borrowed=8, orphan=4, realNode=gpu-node-011, claimedNode=gpu-node-011
  count=16, borrowed=6, orphan=3, realNode=gpu-node-135, claimedNode=gpu-node-135
  count=16, borrowed=16, orphan=5, realNode=gpu-node-010, claimedNode=gpu-node-010
```

Sanitized run/claim mismatch:

```text
gpu-node-025  running=2  claimed=10
gpu-node-019  running=0  claimed=6
gpu-node-038  running=4  claimed=0
```

This shows that the scheduler-side claim space and the kubelet runtime space can diverge significantly.

# Proposed fix plan

1. **Implement vGPU Release / rollback compensation**
   - Clean up or rollback `vgpu-node`, `devices-to-allocate`, `bind-phase`, node lock, and device accounting on failed or discarded scheduling operations.

2. **Upgrade node lock semantics**
   - Store `timestamp + namespace + podName + podUID`;
   - add owner check on release;
   - support safe dangling-lock cleanup.

3. **Reduce or remove `vgpu-node` from identity matching**
   - `spec.nodeName` is the final runtime node;
   - `vgpu-node` may remain as diagnostic information, but should not be the sole device-plugin identity index.

4. **Add high-concurrency tests**
   - 16 half-GPU Pods per 8-GPU node;
   - around `16 * 98` Pod-level concurrency;
   - track scheduling latency, node-lock waits, API patch latency, and `BORROW_*` ratio.

## Related issue

A companion issue is needed in `volcano-vgpu-device-plugin`: the device plugin currently guesses Pod identity with `getOldestPod` and unconditionally removes vGPU cache directories in `Allocate`. The plugin also needs deviceID-aware matching, ambiguous-candidate rejection, and live mount protection.

106umao:
Keep TaskInfo's vGPU allocation annotations aligned with the pod copy used in device allocation and cleanup paths. This prevents stale bind metadata from being propagated while preserving committed annotations after successful bind.

#### What type of PR is this?

/kind bug

This PR fixes stale vGPU annotation propagation across scheduler allocate/bind/rollback paths.

Problem chain:
- `Allocate` patches annotations.
- API server update is asynchronous.
- Informer may still serve an older pod object.
- Binding/cleanup can then read stale annotations from `TaskInfo.Pod`, propagating outdated bind metadata.

What is fixed:
- Keep `TaskInfo.Pod` annotations synchronized with the allocate-time pod copy after successful `devices.Allocate`.
- Preserve committed annotations when `bind-phase=success`.
- Remove speculative annotations only on rollback paths.

Architectural note:
- The long-term direction should be cache-only and patchless behavior in scheduler-cycle `Allocate`, aligned with default scheduler principles.
- Scheduler-cycle allocate should only reserve optimistically in cache.
- API-server persistence should happen in bind-cycle through a dedicated `bindAllocatedAnnotation` step.
- This PR is a cache-consistency fix; it does not yet remove allocate-time patch calls.

Fixes #5532

This PR intentionally focuses on annotation consistency only.
Node lock release behavior is split into a separate PR to keep review scope small.

```release-note
TaskInfo now includes PodAnnotations for scheduler-owned pod annotation updates during scheduling. Plugins that need to modify and persist pod annotations should call MergePodAnnotations; the merged annotations are carried by the bind request.

### DIFF
--- a/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
+++ b/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
@@ -277,18 +277,20 @@ func (gs *GPUDevices) HasDeviceRequest(pod *v1.Pod) bool {
 	return false
 }
 
-func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) error {
-	if gs == nil || pod == nil || pod.Annotations == nil {
-		return nil
+func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) (*devices.DeviceReservation, error) {
+	if gs == nil || pod == nil {
+		return nil, nil
 	}
-	// Release is required for rollback paths (e.g. UnPipeline) where NodeInfo
-	// does not invoke subResource for Pipelined tasks.
+	// Roll back the in-memory device ledger for UnPipeline/UnAllocate paths.
 	gs.SubResource(pod)
 
-	if pod.Annotations[DeviceBindPhase] == "success" {
-		return nil
+	// Once binding succeeds, vgpu-device-plugin owns these annotations.
+	// The scheduler must not clean them again.
+	if pod.Annotations != nil && pod.Annotations[DeviceBindPhase] == "success" {
+		return nil, nil
 	}
 
+	// Clean allocation annotations persisted to the apiserver.
 	keys := []string{
 		AssignedNodeAnnotations,          // volcano.sh/vgpu-node
 		AssignedIDsAnnotations,           // volcano.sh/vgpu-ids-new
@@ -298,12 +300,12 @@ func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) erro
 		DeviceBindPhase,                  // volcano.sh/bind-phase
 	}
 	if err := devices.RemovePodAnnotations(kubeClient, pod, keys); err != nil {
-		return err
+		return nil, err
 	}
-	for _, k := range keys {
-		delete(pod.Annotations, k)
-	}
-	return nil
+	return &devices.DeviceReservation{
+		DeviceType:  DeviceName,
+		Annotations: devices.AnnotationKeyMap(keys),
+	}, nil
 }
 
 func (gs *GPUDevices) FilterNode(pod *v1.Pod, schedulePolicy string) (int, string, error) {
@@ -320,54 +322,43 @@ func (gs *GPUDevices) FilterNode(pod *v1.Pod, schedulePolicy string) (int, strin
 	return devices.Success, "", nil
 }
 
-func (gs *GPUDevices) Allocate(kubeClient kubernetes.Interface, pod *v1.Pod) error {
-	if VGPUEnable {
-		klog.V(4).Infoln("hami-vgpu DeviceSharing:Into AllocateToPod", pod.Name)
-		if alreadyAssignedOnNode(pod, gs.Name) {
-			klog.V(4).InfoS("hami-vgpu DeviceSharing: skip duplicate AllocateToPod",
-				"pod", pod.Name, "namespace", pod.Namespace, "node", gs.Name)
-			return nil
-		}
-		fit, device, _, err := checkNodeGPUSharingPredicateAndScore(pod, gs, false, SchedulePolicy)
-		if err != nil || !fit {
-			klog.ErrorS(err, "Failed to allocate vgpu task", "pod", pod.Name)
-			return err
-		}
-		if NodeLockEnable {
-			nodelock.UseClient(kubeClient)
-			err = nodelock.LockNode(gs.Name, DeviceName)
-			if err != nil {
-				return errors.Errorf("node %s locked for %s hamivgpu lockname %s", gs.Name, pod.Name, err.Error())
-			}
-		}
-
-		annotations := make(map[string]string)
-		annotations[AssignedNodeAnnotations] = gs.Name
-		annotations[AssignedTimeAnnotations] = strconv.FormatInt(time.Now().Unix(), 10)
-		annotations[AssignedIDsAnnotations] = encodePodDevices(device)
-		annotations[AssignedIDsToAllocateAnnotations] = annotations[AssignedIDsAnnotations]
-
-		annotations[DeviceBindPhase] = "allocating"
-		annotations[BindTimeAnnotations] = strconv.FormatInt(time.Now().Unix(), 10)
-		// Keep in-memory pod object in sync so rollback paths (UnPipeline ->
-		// Deallocate -> Release) can see allocated IDs before apiserver watch
-		// catches up.
-		if pod.Annotations == nil {
-			pod.Annotations = map[string]string{}
-		}
-		for k, v := range annotations {
-			pod.Annotations[k] = v
-		}
-		// To avoid that the pod allocated info updating latency, add it first
-		gs.addToPodMap(annotations, pod)
-		err = patchPodAnnotations(kubeClient, pod, annotations)
-		if err != nil {
-			return err
-		}
-
-		klog.V(3).Infoln("DeviceSharing:Allocate Success")
+func (gs *GPUDevices) Allocate(kubeClient kubernetes.Interface, pod *v1.Pod) (*devices.DeviceReservation, error) {
+	if !VGPUEnable {
+		return nil, nil
 	}
-	return nil
+	klog.V(4).Infoln("hami-vgpu DeviceSharing:Into AllocateToPod", pod.Name)
+	fit, device, _, err := checkNodeGPUSharingPredicateAndScore(pod, gs, false, SchedulePolicy)
+	if err != nil || !fit {
+		klog.ErrorS(err, "Failed to allocate vgpu task", "pod", pod.Name)
+		return nil, err
+	}
+	if NodeLockEnable {
+		nodelock.UseClient(kubeClient)
+		err = nodelock.LockNode(gs.Name, DeviceName)
+		if err != nil {
+			return nil, errors.Errorf("node %s locked for %s hamivgpu lockname %s", gs.Name, pod.Name, err.Error())
+		}
+	}
+
+	annotations := make(map[string]string)
+	annotations[AssignedNodeAnnotations] = gs.Name
+	annotations[AssignedTimeAnnotations] = strconv.FormatInt(time.Now().Unix(), 10)
+	annotations[AssignedIDsAnnotations] = encodePodDevices(device)
+	annotations[AssignedIDsToAllocateAnnotations] = annotations[AssignedIDsAnnotations]
+	annotations[DeviceBindPhase] = "allocating"
+	annotations[BindTimeAnnotations] = strconv.FormatInt(time.Now().Unix(), 10)
+
+	// Update the in-memory device ledger without mutating shared pod annotations.
+	// Passing annotations directly avoids concurrent reads/writes on the shared Pod map.
+	gs.addToPodMap(annotations, pod)
+
+	// Return allocation annotations so predicates can carry the same values on
+	// the binding request. The apiserver merges Binding annotations into the Pod
+	// atomically with node assignment.
+	return &devices.DeviceReservation{
+		DeviceType:  DeviceName,
+		Annotations: annotations,
+	}, nil
 }
 
 // DeepCopy returns a deep copy of GPUDevices for use in dry-run simulation.
@@ -415,13 +406,4 @@ func (gs *GPUDevices) DeepCopy() interface{} {
 		cp.Device[id] = newDev
 	}
 	return cp
-}
-
-func alreadyAssignedOnNode(pod *v1.Pod, nodeName string) bool {
-	if pod == nil || pod.Annotations == nil || nodeName == "" {
-		return false
-	}
-
-	return pod.Annotations[AssignedNodeAnnotations] == nodeName &&
-		pod.Annotations[AssignedIDsAnnotations] != ""
 }
--- a/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
+++ b/pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
@@ -290,7 +290,6 @@ func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) (*de
 		return nil, nil
 	}
 
-	// Clean allocation annotations persisted to the apiserver.
 	keys := []string{
 		AssignedNodeAnnotations,          // volcano.sh/vgpu-node
 		AssignedIDsAnnotations,           // volcano.sh/vgpu-ids-new
@@ -298,9 +297,6 @@ func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) (*de
 		AssignedTimeAnnotations,          // volcano.sh/vgpu-time
 		BindTimeAnnotations,              // volcano.sh/bind-time
 		DeviceBindPhase,                  // volcano.sh/bind-phase
-	}
-	if err := devices.RemovePodAnnotations(kubeClient, pod, keys); err != nil {
-		return nil, err
 	}
 	return &devices.DeviceReservation{
 		DeviceType:  DeviceName,
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -232,9 +232,10 @@ func (db *DefaultBinder) Bind(kubeClient kubernetes.Interface, tasks []*scheduli
 	for _, task := range tasks {
 		p := task.Pod
 		startTime := time.Now()
+		// The apiserver overlays Binding annotations onto the Pod during bind.
 		if err := db.kubeclient.CoreV1().Pods(p.Namespace).Bind(context.TODO(),
 			&v1.Binding{
-				ObjectMeta: metav1.ObjectMeta{Namespace: p.Namespace, Name: p.Name, UID: p.UID, Annotations: p.Annotations},
+				ObjectMeta: metav1.ObjectMeta{Namespace: p.Namespace, Name: p.Name, UID: p.UID, Annotations: task.PodAnnotations},
 				Target: v1.ObjectReference{
 					Kind: "Node",
 					Name: task.NodeName,
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -888,9 +888,11 @@ func (pp *PredicatesPlugin) PreBindRollBack(ctx context.Context, bindCtx *cache.
 }
 
 func (pp *PredicatesPlugin) SetupBindContextExtension(state *k8sframework.CycleState, bindCtx *cache.BindContext) {
-	if pp.needsPreBind(bindCtx.TaskInfo) {
-		bindCtx.Extensions[pp.Name()] = &BindContextExtension{State: state}
+	if !pp.needsPreBind(bindCtx.TaskInfo) {
+		return
 	}
+
+	bindCtx.Extensions[pp.Name()] = &BindContextExtension{State: state}
 }
 
 func handleSkipPredicatePlugin(state fwk.CycleState, pluginName string) bool {
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -238,11 +238,14 @@ func (pp *PredicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 						continue
 					}
 
-					err := devices.Allocate(ssn.KubeClient(), pod)
+					reservation, err := devices.Allocate(ssn.KubeClient(), pod)
 					if err != nil {
 						klog.Errorf("AllocateToPod failed %s", err.Error())
 						event.Err = err
 						return
+					}
+					if reservation != nil && len(reservation.Annotations) > 0 {
+						event.Task.MergePodAnnotations(reservation.Annotations)
 					}
 				} else {
 					klog.Warningf("Devices %s assertion conversion failed, skip", val)
@@ -284,10 +287,15 @@ func (pp *PredicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 					}
 
 					// deallocate pod gpu id
-					err := devices.Release(ssn.KubeClient(), pod)
+					reservation, err := devices.Release(ssn.KubeClient(), pod)
 					if err != nil {
 						klog.Errorf("Device %s release failed for pod %s/%s, err:%s", val, pod.Namespace, pod.Name, err.Error())
 						return
+					}
+					if reservation != nil {
+						if keys := reservation.AnnotationKeys(); len(keys) > 0 {
+							event.Task.DeletePodAnnotations(keys...)
+						}
 					}
 				} else {
 					klog.Warningf("Devices %s assertion conversion failed, skip", val)
@@ -880,11 +888,9 @@ func (pp *PredicatesPlugin) PreBindRollBack(ctx context.Context, bindCtx *cache.
 }
 
 func (pp *PredicatesPlugin) SetupBindContextExtension(state *k8sframework.CycleState, bindCtx *cache.BindContext) {
-	if !pp.needsPreBind(bindCtx.TaskInfo) {
-		return
+	if pp.needsPreBind(bindCtx.TaskInfo) {
+		bindCtx.Extensions[pp.Name()] = &BindContextExtension{State: state}
 	}
-
-	bindCtx.Extensions[pp.Name()] = &BindContextExtension{State: state}
 }
 
 func handleSkipPredicatePlugin(state fwk.CycleState, pluginName string) bool {
--- a/pkg/scheduler/api/devices/ascend/hami/device_info.go
+++ b/pkg/scheduler/api/devices/ascend/hami/device_info.go
@@ -315,18 +315,18 @@ func (ads *AscendDevices) ScoreNode(pod *v1.Pod, policy string) float64 {
 	return score
 }
 
-func (ads *AscendDevices) Allocate(kubeClient kubernetes.Interface, pod *v1.Pod) error {
+func (ads *AscendDevices) Allocate(kubeClient kubernetes.Interface, pod *v1.Pod) (*devices.DeviceReservation, error) {
 	klog.V(4).Infof("Allocate device %s to Pod %s", ads.Type, pod.Name)
 	if NodeLockEnable {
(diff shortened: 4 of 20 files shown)

### CODE REVIEW

**Comment by JesseStutler** [comment]
@106umao The code verify CI failed, needs to follow the go fmt

**Comment by 106umao** [comment]
@JesseStutler 
The current change leads to unsafe concurrent read/write operations on the map, causing the process to restart 
frequently. Are there any other recommended approaches to achieve the desired outcome?

**Comment by 106umao** [comment]
@JesseStutler 

I noticed that Volcano's `DefaultBinder.Bind` includes the full Pod annotations in the `Binding` object:

```go
    Namespace:   p.Namespace,
    Name:        p.Name,
    UID:         p.UID,
    Annotations: p.Annotations,
}
```

Could you help clarify why pod.Annotations are passed during binding?

Kubernetes apiserver merges binding.Annotations back into the Pod in the binding subresource path, but this does not protect the annotations with the latest Pod resourceVersion. If TaskInfo.Pod is stale, the binding request may write old annotations back to the Pod and overwrite newer annotation state.

This seems risky for plugins that use annotations as scheduling/allocation state. Would it be safer for Volcano to follow Kubernetes default scheduler behavior and only set Namespace, Name, and UID in the Binding.ObjectMeta?

JesseStutler:
> @JesseStutler
> 
> I noticed that Volcano's `DefaultBinder.Bind` includes the full Pod annotations in the `Binding` object:
> 
> ```go
> ObjectMeta: metav1.ObjectMeta{
>     Namespace:   p.Namespace,
>     Name:        p.Name,
>     UID:         p.UID,
>     Annotations: p.Annotations,
> }
> ```
> 
> Could you help clarify why pod.Annotations are passed during binding?
> 
> Kubernetes apiserver merges binding.Annotations back into the Pod in the binding subresource path, but this does not protect the annotations with the latest Pod resourceVersion. If TaskInfo.Pod is stale, the binding request may write old annotations back to the Pod and overwrite newer annotation state.
> 
> This seems risky for plugins that use annotations as scheduling/allocation state. Would it be safer for Volcano to follow Kubernetes default scheduler behavior and only set Namespace, Name, and UID in the Binding.ObjectMeta?

This has existed since the volcano was just open-sourced before I maintained it. kube-scheduler doesn't need to update pod annotations because it doesn't have custom plugins. However, what you mentioned does count as a risk item, especially when multiple components need to update pod annotations simultaneously. But so far, we haven't encountered this kind of issue. Needs more evidence that we trully meet them, and we need to be careful to refactor the current logic since it may also be a risk to migrate since currently lots of plugins may depend on this annotation updating logic

JesseStutler:
> @JesseStutler The current change leads to unsafe concurrent read/write operations on the map, causing the process to restart frequently. Are there any other recommended approaches to achieve the desired outcome?

What do you mean, I didn't get your point of the context

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -245,6 +284,9 @@ func (pp *PredicatesPlugin) OnSessionOpen(ssn *framework.Session) {
 						event.Err = err
 						return
 					}
+					if val == "hamivgpu" {
```
JesseStutler: I don't understand why there needs to be an annotation copy from one pod to another, both of which are in the session lifecycle? Are their annotations different? But devices.Allocate didn't get the pod from the api-server

Also could you explain that why deviceshare related codes @archlitchi  needs to directly update in the AllocateFunc process instead of during binding?

106umao:
> > @JesseStutler The current change leads to unsafe concurrent read/write operations on the map, causing the process to restart frequently. Are there any other recommended approaches to achieve the desired outcome?
> 
> What do you mean, I didn't get your point of the context

Too many places are mutating the v1.Pod.Annotations map concurrently, which can cause panics. We should move this state to a concurrency-safe map and consolidate its usage.

pkg/scheduler/plugins/predicates/predicates.go
106umao: Cause we could not directly modify the v1.Pod instance that DefaultBinder.Bind() depends on. I have now abandoned that approach. Please review the latest code.

106umao:
There is also a nodelock issue: it should also be removed from device.Allocate and reworked into a reserve/unreserve + prebind/prebind-rollback flow. But this PR does not address that yet.

BTW, I think the Device interface should ideally be designed more like a predicate plugin, so it can extend each scheduling phase explicitly. That is a separate discussion for later.

JesseStutler:
> > > @JesseStutler The current change leads to unsafe concurrent read/write operations on the map, causing the process to restart frequently. Are there any other recommended approaches to achieve the desired outcome?
> > 
> > 
> > What do you mean, I didn't get your point of the context
> 
> Too many places are mutating the v1.Pod.Annotations map concurrently, which can cause panics. We should move this state to a concurrency-safe map and consolidate its usage.

Could you link out? Thanks, but currently only cache and session are in the two different go routines I think, don't know if I missed something for like vgpu logic

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
```diff
@@ -320,54 +322,56 @@ func (gs *GPUDevices) FilterNode(pod *v1.Pod, schedulePolicy string) (int, strin
...
+	annotations[AssignedIDsToAllocateAnnotations] = annotations[AssignedIDsAnnotations]
+	annotations[DeviceBindPhase] = "allocating"
+	annotations[BindTimeAnnotations] = strconv.FormatInt(time.Now().Unix(), 10)
+
+	// Update the in-memory device ledger without mutating shared pod annotations.
+	// Passing annotations directly avoids concurrent reads/writes on the shared Pod map.
+	gs.addToPodMap(annotations, pod)
+
+	// Persist allocation annotations for vgpu-device-plugin on the kubelet side.
+	if err = patchPodAnnotations(kubeClient, pod, annotations); err != nil {
```
JesseStutler: I understand your point, but when the core functionality of vgpu was merged, I wasn't tracking it. I think calling `pod update` directly from `AllocateFunc` might not be a good approach, as it could block the session. Also, due to the existence of "gang," it's possible that the Allocate update for annotation results could be released and then rolled back, which I actually prefer not to happen here. I'd rather just update in the cache instead of making API calls. However, we can wait for @archlitchi to confirm

JesseStutler:
> There is also a nodelock issue: it should also be removed from device.Allocate and reworked into a reserve/unreserve + prebind/prebind-rollback flow. But this PR does not address that yet.
> 
> BTW, I think the Device interface should ideally be designed more like a predicate plugin, so it can extend each scheduling phase explicitly. That is a separate discussion for later.

@106umao Maybe we can have a discussion with @DSFans2014 @archlitchi together on the community meeting or private meeting to align the requirements, currently looks like vgpu still needs some improvement, also @nishitnshah Also have some fixes and requirements on it

106umao:
> > > > @JesseStutler The current change leads to unsafe concurrent read/write operations on the map, causing the process to restart frequently. Are there any other recommended approaches to achieve the desired outcome?当前的变化会导致对映射表进行不安全的同时读写操作，从而导致进程频繁重启。有没有其他推荐的方法可以达到预期的效果呢？
> > > 
> > > 
> > > What do you mean, I didn't get your point of the context你的意思是，我没有理解你所提到的上下文中的要点吗？
> > 
> > 
> > Too many places are mutating the v1.Pod.Annotations map concurrently, which can cause panics. We should move this state to a concurrency-safe map and consolidate its usage.有太多地方同时在对 v1.Pod.Annotations 映射进行修改操作，这可能会引发恐慌情况。我们应该将这种状态存储机制迁移到一种具备并发安全性的映射结构中，并整合其使用方式。
> 
> Could you link out? Thanks, but currently only cache and session are in the two different go routines I think, don't know if I missed something for like vgpu logic你能提供一下链接吗？谢谢，不过目前只有缓存和会话信息存在于这两个不同的程序中。不知道是否遗漏了与 VGP 逻辑相关的部分。

It might have been my mistake; in yesterday's commit, I tried modifying `v1.Pod.Annotations` directly inside `AllocateFunc`, which caused a panic. So, in the latest commit, I added a `BindAnnotation` field to `TaskInfo`.

106umao:
> > > > > @JesseStutler The current change leads to unsafe concurrent read/write operations on the map, causing the process to restart frequently. Are there any other recommended approaches to achieve the desired outcome?当前的变化会导致对映射表进行不安全的同时读写操作，从而导致进程频繁重启。有没有其他推荐的方法可以达到预期的效果呢？当前的变化会导致对映射表进行不安全的同时读写操作，进而使进程频繁重启。有没有其他推荐的方法可以达到预期的效果呢？
> > > > 
> > > > 
> > > > What do you mean, I didn't get your point of the context你的意思是，我没有理解你所提到的上下文中的要点吗？
> > > 
> > > 
> > > Too many places are mutating the v1.Pod.Annotations map concurrently, which can cause panics. We should move this state to a concurrency-safe map and consolidate its usage.有太多地方同时在对 v1.Pod.Annotations 映射进行修改操作，这可能会引发恐慌情况。我们应该将这种状态存储机制迁移到一种具备并发安全性的映射结构中，并整合其使用方式。
> > 
> > 
> > Could you link out? Thanks, but currently only cache and session are in the two different go routines I think, don't know if I missed something for like vgpu logic你能提供一下链接吗？谢谢，不过目前只有缓存和会话信息存在于这两个不同的程序中。不知道是否遗漏了与 VGP 逻辑相关的部分。能请您提供一下链接吗？谢谢。不过目前只有缓存和会话信息存在于这两个不同的程序中。不知道是否遗漏了与 VGP 逻辑相关的部分。
> 
> It might have been my mistake; in yesterday's commit, I tried modifying `v1.Pod.Annotations` directly inside `AllocateFunc`, which caused a panic. So, in the latest commit, I added a `BindAnnotation` field to `TaskInfo`.也许是我的错误；在昨天的提交中，我试图直接在 `AllocateFunc` 内部修改 `v1.Pod.Annotations` ，这导致了问题。因此，在最新的提交中，我添加了 `BindAnnotation` 字段到 `TaskInfo` 中。

I am new to the Volcano project, so there are still many aspects of its internal implementation mechanisms that I do not fully understand. I ask for your understanding should I happen to make any elementary mistakes.

106umao:
PTAL @JesseStutler

pkg/scheduler/cache/cache.go
```diff
@@ -131,7 +132,7 @@ type SchedulerCache struct {
 	quotaInformer              infov1.ResourceQuotaInformer
 	csiNodeInformer            storagev1.CSINodeInformer
 	csiDriverInformer          storagev1.CSIDriverInformer
-	csiStorageCapacityInformer storagev1.CSIStorageCapacityInformer
+	csiStorageCapacityInformer storagev1beta1.CSIStorageCapacityInformer
```
JesseStutler: No need to revert back to storagev1beta1

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
JesseStutler: /cc @106umao As we discussed before, I think we shouldn't call api server directly at Allocate stage, only patch the pod's annotation in the cache and wail until Bind API call to update it

pkg/scheduler/cache/cache.go
```diff
@@ -232,9 +233,10 @@ func (db *DefaultBinder) Bind(kubeClient kubernetes.Interface, tasks []*scheduli
 	for _, task := range tasks {
 		p := task.Pod
 		startTime := time.Now()
+		task.LoadOrStoreBindAnnotations(p.Annotations)
 		if err := db.kubeclient.CoreV1().Pods(p.Namespace).Bind(context.TODO(),
 			&v1.Binding{
-				ObjectMeta: metav1.ObjectMeta{Namespace: p.Namespace, Name: p.Name, UID: p.UID, Annotations: p.Annotations},
+				ObjectMeta: metav1.ObjectMeta{Namespace: p.Namespace, Name: p.Name, UID: p.UID, Annotations: task.ListBindAnnotations()},
```
JesseStutler: I still feel like the downstream fork users when they need to update to the latest version, they didn't know about such change and may need to other work to implement it. Before introducing TaskInfo.BindAnnotations, could you please provide the panic stack or race-detector report showing the concurrent reader and writer on the same Pod.Annotations map? After invesigation, I didn't find any other reader/writer during Binding stage.

JesseStutler (commented):
Also there are lots of CI failed, please resolve it before requesting review. If it's not introduced by your PR, you can just ping me, but I feel like there might be some risk in your PR then it cause so many CIs failed, please check

pkg/scheduler/cache/cache.go
106umao: Currently, concurrent read/write operations do not exist, but when we try to modify `TaskInfo.pod.Annotations` in `predicates.AllocateFunc`, it will cause a panic. Additionally, I've tested this; when `cache.Binding` is called, a panic will occur when serializing the map and other grouting operations are executed concurrently, because the underlying implementation of serializing the map is actually a traversal of the key-value pairs.

106umao: I'm sorry for making this silly mistake and wasting everyone's time. I accidentally rolled back the code while resolving the conflict.

106umao: 2026-07-07T10:05:25.613292608+08:00
k8s.io/client-go/tools/cache.FilteringResourceEventHandler.OnUpdate({0x78f495ba4e0?, {0x2d1f8a0?, 0x78f495a83f0?}}, {0x2b7e2a0, 0x78f65f8b688}, {0x2b7e2a0, 0x78f628aed88})
    $GOMODCACHE/k8s.io/client-go@v0.33.2/tools/cache/controller.go:329 +0xdc
k8s.io/client-go/tools/cache.(*processorListener).run.func1(0x78f7837bf5f, 0x78f493fc320, {0x2880080?, 0x78f6f02fc0?})
    $GOMODCACHE/k8s.io/client-go@v0.33.2/tools/cache/shared_informer.go:1075 +0x149
k8s.io/client-go/tools/cache.(*processorListener).run(0x78f493f320)
    $GOMODCACHE/k8s.io/client-go@v0.33.2/tools/cache/shared_informer.go:1087 +0x3c
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1()
    $GOMODCACHE/k8s.io/apimachinery@v0.33.2/pkg/util/wait/wait.go:72 +0x4c
created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start in goroutine 322
    $GOMODCACHE/k8s.io/apimachinery@v0.33.2/pkg/util/wait/wait.go:70 +0x73

2026-07-07T10:16:43.524412977+08:00
k8s.io/client-go/tools/cache.FilteringResourceEventHandler.OnUpdate({0x1e8068f84140?, {0x2d1f8a0?, 0x1e8068f88108?}}, {0x2b7e2a0, 0x1e808871ad88}, {0x2b7e2a0, 0x1e8082ad2908})
    $GOMODCACHE/k8s.io/client-go@v0.33.2/tools/cache/controller.go:329 +0xdc
k8s.io/client-go/tools/cache.(*processorListener).run.func1(0x1e806bbadf5f, 0x1e8068fb40a0, {0x2880080?, 0x1e8087fe1d60?})
    $GOMODCACHE/k8s.io/client-go@v0.33.2/tools/cache/shared_informer.go:1075 +0x149
k8s.io/client-go/tools/cache.(*processorListener).run(0x1e8068fb40a0)
    $GOMODCACHE/k8s.io/client-go@v0.33.2/tools/cache/shared_informer.go:1087 +0x3c
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1()
    $GOMODCACHE/k8s.io/apimachinery@v0.33.2/pkg/util/wait/wait.go:72 +0x4c
created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start in goroutine 281
    $GOMODCACHE/k8s.io/apimachinery@v0.33.2/pkg/util/wait/wait.go:70 +0x73

106umao: This section now implements merging old annotations into `.BindAnnotation`, which theoretically shouldn't affect downstream processes.

JesseStutler: Got it, from the trace stack, the panic looks like happened in the event handler, I think it's because the session currently didn't deep copy the Pod(but we found that deep copy the whole pod may cause latency during snapshot, so I don't prefer to deep copy the whole pod), therefore in the AllocateFunc the session directly modify the Pod's annotation and then cause panic, because the AllocateFunc directly call the API update and then the inforger got watch event.

So we have to prevent and remove the directly API call during AllocateFunc like I mentioned here: https://github.com/volcano-sh/volcano/pull/5588/changes#r3575861007

I think we can indeed define a new field called `Annotations` in the TaskInfo, but I don't think we need sync.Map or lock, during the `Clone` for the TaskInfo, the sesssion will call maps.Clone for the map, and then the vgpu plugin/ numa aware plugin only change this map(not directly API call in the session), then finally the cache send the bind API call and carry on the annotations map(there will be no concurrent read and write then)

106umao:
@JesseStutler PTAL

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
```diff
@@ -298,12 +300,12 @@ func (gs *GPUDevices) Release(kubeClient kubernetes.Interface, pod *v1.Pod) erro
 		DeviceBindPhase,                  // volcano.sh/bind-phase
 	}
 	if err := devices.RemovePodAnnotations(kubeClient, pod, keys); err != nil {
```
JesseStutler: The API call during Release stage to remove the pod annotations should all be removed and only operate in the cache, it needs to be aligned with Allocate.

pkg/scheduler/plugins/predicates/predicates.go
```diff
@@ -880,11 +888,9 @@ func (pp *PredicatesPlugin) PreBindRollBack(ctx context.Context, bindCtx *cache.
 }
 
 func (pp *PredicatesPlugin) SetupBindContextExtension(state *k8sframework.CycleState, bindCtx *cache.BindContext) {
```
JesseStutler, line 890: It looks like there's nothing that needs changing here?Keeping the code as-is simply aligns it with the PR description.

JesseStutler:
Please also update the release note that there is a new field called `PodAnnotations` in the `TaskInfo`, which is used to hold pod annotations that need to be modified during scheduling. If developers need to modify pod annotations and persist them, they need to call `MergePodAnnotations` to merge them into this annotations. @106umao

pkg/scheduler/api/devices/nvidia/vgpu/device_info.go
106umao: done

pkg/scheduler/plugins/predicates/predicates.go
106umao: done

JesseStutler:
/approve
Thanks
From the code, it lgtm @106umao Please make sure that current code change still doesn't influence the functionality and vGPU/NPU can be scheduled normally. 
/cc @archlitchi Please take a review

archlitchi (approved):
/lgtm

PR #5590, opened 2026-07-06
feat: add NamespaceQueue support

gitGurugu:
What type of PR is this?

/kind feature

This PR introduces NamespaceQueue support in Volcano, providing namespace-scoped queue management on top of the existing cluster-scoped Queue API.

NamespaceQueue allows users to define queue configuration within a namespace and enables workloads to reference queues in a namespace-scoped manner.

This PR includes the complete NamespaceQueue implementation, including:

- Adds the `NamespaceQueue` API type and CRD.
- Adds generated DeepCopy methods, clientsets, informers, and listers for `NamespaceQueue`.
- Adds NamespaceQueue lifecycle management and controller integration.
- Adds NamespaceQueue informer/cache integration into the scheduler.
- Adds namespace-scoped queue resolution for workloads.
- Integrates NamespaceQueue with the existing Volcano queue and scheduling workflow.
- Adds validation and related handling required for NamespaceQueue.
- Adds unit tests and generated manifests for the new API.

With this change, users can manage queue configuration at the namespace level while Volcano internally integrates the NamespaceQueue with the existing scheduling and queue management mechanisms.

Which issue(s) this PR fixes:

Related to #5251

This PR introduces the complete NamespaceQueue functionality across the API, controller, and scheduler layers.

The implementation mainly consists of:

1. NamespaceQueue API and CRD definitions.
2. Generated Kubernetes client/informer/lister code.
3. NamespaceQueue controller and lifecycle handling.
4. Scheduler cache and informer integration.
5. Namespace-aware queue resolution and scheduling integration.
6. Validation and related test coverage.

Most generated files in this PR are produced by the repository code-generation scripts. The primary manually maintained changes are concentrated in the NamespaceQueue API definitions, controller logic, scheduler integration, and queue resolution logic.

```release-note
Add NamespaceQueue support to enable namespace-scoped queue management and scheduling in Volcano.

### CODE REVIEW

**Comment by hajnalmt** [comment]
/cc @hajnalmt

^^

File: staging/src/volcano.sh/apis/pkg/apis/scheduling/v1beta1/types.go [hunk 1]
```diff
@@ -367,6 +367,73 @@ type Queue struct {
...
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Parent string `json:"parent" protobuf:"bytes,1,opt,name=parent"`
+
+	// +optional
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Weight int32 `json:"weight,omitempty" protobuf:"bytes,2,opt,name=weight"`
+
```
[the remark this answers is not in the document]
devzizu: We can ignore this as this field was proposed to be dropped: https://github.com/volcano-sh/volcano/pull/5629#discussion_r3638056278

installer/helm/chart/volcano/templates/scheduling_v1beta1_namespacequeue.yaml
```diff
@@ -0,0 +1 @@
+{{- tpl ($.Files.Get (printf "crd/%s/scheduling.volcano.sh_namespacequeues.yaml" (include "crd_version" .))) . }}
```
devzizu, line 1: One thing worth a quick note: RBAC for NSQs isn't in this PR, and this seems to be how previous CRDs were introduced like `nodeshards` and `jobflows`, so RBAC tends to land with the PR that adds the controller or scheduler code actually using the resource.

Please confirm this can be shipped without RBAC-related manifests @JesseStutler @hajnalmt, but I think it's ok since there is no informer setup yet that could break

devzizu (commented):
Also `installer/volcano-development.yaml` wasn't regenerated, so it's still missing the new CRD. This is the file the README tells users to `kubectl apply -f` directly for a standard install, and it currently bundles every other CRD (`queues`, `podgroups`, `hypernodes`, etc). I think `make update-development-yaml` does the trick, can you please apply it?

devzizu:
@gitGurugu please sign-off all commits `git rebase HEAD~5 --signoff` (or check DCO run recommendation: https://github.com/volcano-sh/volcano/pull/5590/checks?check_run_id=89962882518)

volcano.sh/apis/pkg/client/versioned/clientset.go
```diff
@@ -0,0 +1,146 @@
...
+*/
+
+// Code generated by client-gen. DO NOT EDIT.
+
+package versioned
+
+import (
+	batchv1alpha1 "clientset/versioned/typed/batch/v1alpha1"
+	busv1alpha1 "clientset/versioned/typed/bus/v1alpha1"
+	schedulingv1beta1 "clientset/versioned/typed/scheduling/v1beta1"
```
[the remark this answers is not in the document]
devzizu: I believe this has been fixed already in 65df374d3e7579c42e1c768e9e87ad05cb311a4e

volcano.sh/apis/pkg/client/versioned/clientset.go
```diff
@@ -0,0 +1,146 @@
...
+	fmt "fmt"
```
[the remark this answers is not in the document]
devzizu: I believe this has been fixed already in https://github.com/volcano-sh/volcano/commit/65df374d3e7579c42e1c768e9e87ad05cb311a4e

volcano.sh/apis/pkg/client/versioned/typed/scheduling/v1beta1/scheduling_client.go
```diff
@@ -0,0 +1,111 @@
...
+
+package v1beta1
+
+import (
+	scheme "clientset/versioned/scheme"
+	http "net/http"
+
+	rest "k8s.io/client-go/rest"
+	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+)
```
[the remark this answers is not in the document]
devzizu: I believe this has been fixed already in https://github.com/volcano-sh/volcano/commit/65df374d3e7579c42e1c768e9e87ad05cb311a4e

installer/helm/chart/volcano/templates/scheduling_v1beta1_namespacequeue.yaml
devzizu: Actually, `nodeshards` introduces the CRDs, RBAC, and actual implementation directly in the same PR (https://github.com/volcano-sh/volcano/pull/4804). I think It would depend on how maintainers prefer it, let's wait for other reviews.

hajnalmt (commented):
Super, this looks good so far.

Please squash the PR into a single commit and let's continue with the implementation here! The design is looking overall good to me.

installer/helm/chart/volcano/templates/scheduling_v1beta1_namespacequeue.yaml
hajnalmt: We should add this into the same PR, both the implementation and the RBAC.

staging/src/volcano.sh/apis/pkg/apis/scheduling/v1beta1/types.go
```diff
@@ -503,6 +605,11 @@ type QueueSpec struct {
 	// +kubebuilder:default:=traverse
 	// +kubebuilder:validation:Enum=fifo;traverse
 	DequeueStrategy DequeueStrategy `json:"dequeueStrategy,omitempty" protobuf:"bytes,11,opt,name=dequeueStrategy"`
+
+	// AllowedNamespaces lists the namespaces whose NamespaceQueues are permitted
```
JesseStutler: The comment should update now for the [*] @gitGurugu , just the same as in your design doc

JesseStutler:
@gitGurugu You can go ahead with actual namespace queue implementation on this pr and layering commits, we review them actual implementation by different commits

pkg/controllers/namespacequeue/namespacequeue_controller_handler.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+		nq, ok = tombstone.Obj.(*schedulingv1beta1.NamespaceQueue)
+		if !ok {
+			klog.Errorf("tombstone object is not a NamespaceQueue: %#v", tombstone.Obj)
+			return
+		}
+	}
+
+	// 当前对象已经删除，sync 时会因为 NotFound 直接结束。
```
JesseStutler: Before I start the review: @gitGurugu, please remove or translate it to English comments

staging/src/volcano.sh/apis/pkg/apis/scheduling/v1beta1/types.go
```diff
@@ -369,6 +369,113 @@ type Queue struct {
...
+	// "cluster/<name>" refers to a cluster-scoped Queue.
+	// An unprefixed value refers to a NamespaceQueue in the same namespace.
+	// +optional
+	// +kubebuilder:default:="cluster/default"
+	// +kubebuilder:validation:MaxLength=261
+	// +kubebuilder:validation:Pattern=`^(cluster/)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Parent string `json:"parent,omitempty" protobuf:"bytes,1,opt,name=parent"`
+
+	// +optional
+	Capability v1.ResourceList `json:"capability,omitempty" protobuf:"bytes,3,opt,name=capability"`
```
JesseStutler: The protobuf sequence number is incorrect, it's not consecutive.

pkg/controllers/namespacequeue/namespacequeue_controller.go
```diff
@@ -0,0 +1,209 @@
+/*
+Copyright 2019 The Volcano Authors.
```
JesseStutler: All new files should change to 2026

pkg/controllers/namespacequeue/namespacequeue_controller_handler.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+	}
+
+	c.enqueueNamespaceQueuesForParent(parentTarget{
+		scope: clusterParentScope,
+		name:  queue.Name,
+	})
+}
+
+func (c *namespaceQueueController) enqueueNamespaceQueuesForParent(
```
JesseStutler: The variable naming is too messy, making it difficult to read, and comments need to be added.



```suggestion
func (c *namespaceQueueController) enqueueDescendantNamespaceQueues(
	rootRef queueReference,
) {
	allNamespaceQueues, err := c.namespaceQueueLister.List(labels.Everything())
	if err != nil {
		klog.Errorf("failed to list NamespaceQueues: %v", err)
		return
	}

	visitedQueueRefs := map[queueReference]struct{}{rootRef: {}}
	frontier := []queueReference{rootRef}

	for len(frontier) > 0 {
		currentParentRef := frontier[0]
		frontier = frontier[1:]

		for _, namespaceQueue := range allNamespaceQueues {
			resolvedParentRef, err := resolveNamespaceQueueParent(namespaceQueue)
			if err != nil || resolvedParentRef != currentParentRef {
				continue
			}

			childQueueRef := queueReference{
				scope:     namespaceQueueReferenceScope,
				namespace: namespaceQueue.Namespace,
				name:      namespaceQueue.Name,
			}

			if _, alreadyVisited := visitedQueueRefs[childQueueRef]; alreadyVisited {
				continue
			}

			visitedQueueRefs[childQueueRef] = struct{}{}
			c.enqueueNamespaceQueue(namespaceQueue)
			frontier = append(frontier, childQueueRef)
		}
	}
}
```

Please rename these concepts, for example:

- parentTarget → queueReference
- enqueueNamespaceQueuesForParent → enqueueDescendantNamespaceQueues
- namespaceQueueParentTargets → namespaceQueueReferences
- parents → frontier
- current → currentParentRef
- target → resolvedParentRef
- child → childQueueRef

Also, the current traversal scans every NamespaceQueue once for every discovered descendant, resulting in O(N×D), or O(N²) in the worst case. Please build a parentRef → direct children adjacency map in one pass and then traverse it, reducing the operation to O(N+D). The visited check should happen before enqueueing so cyclic references do not cause already-visited queues to be enqueued again. E.g. Like:
```
func (c *namespaceQueueController) getDirectChildNamespaceQueues(
	parentRef controllerutil.ResolvedQueueReference,
) ([]*schedulingv1beta1.NamespaceQueue, error) {
	parentKey := queueReferenceKey(parentRef)
	if parentKey == "" {
		return nil, fmt.Errorf(
			"cannot query children for invalid queue reference: %#v",
			parentRef,
		)
	}

	objects, err := c.namespaceQueueInformer.
		Informer().
		GetIndexer().
		ByIndex(namespaceQueueParentIndexName, parentKey)
	if err != nil {
		return nil, fmt.Errorf(
			"failed to query NamespaceQueue parent index for %q: %w",
			parentKey,
			err,
		)
	}

	children := make(
		[]*schedulingv1beta1.NamespaceQueue,
		0,
		len(objects),
	)

	for _, object := range objects {
		child, ok := object.(*schedulingv1beta1.NamespaceQueue)
		if !ok {
			continue
		}

		children = append(children, child)
	}

	return children, nil
}
```
To directly get the children namespacequeues

pkg/controllers/namespacequeue/namespacequeue_controller_action.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		return err
+	}
+	status.Unknown = counters.Unknown
+	status.Pending = counters.Pending
+	status.Running = counters.Running
+	status.Inqueue = counters.Inqueue
+	status.Completed = counters.Completed
+
+	if status.State == "" {
```
JesseStutler: It seems too simple. I only see state changed to “open” here. But I don’t see any other changes for the state of the namespace queue. Some of the logic related to queue controllers can be reused here (the state trainsition of queue controllers can be extracted to utils).

JesseStutler (commented):
Submit some comment on controller first, will continue review scheduler in recent days

JesseStutler (commented):
I focused this pass on functional availability, upgrade compatibility, and minimizing impact on existing scheduler behavior. The comments below are the main blocking concerns. Will keep continue giving other review comments

hack/run-e2e-kind.sh
```diff
@@ -403,7 +407,8 @@ custom:
     node-role.kubernetes.io/control-plane: ""
   scheduler_feature_gates: ${FEATURE_GATES}
   admission_feature_gates: ${FEATURE_GATES}
-  enabled_admissions: "/pods/mutate,/queues/mutate,/podgroups/mutate,/jobs/mutate,/jobs/validate,/jobflows/validate,/pods/validate,/queues/validate,/podgroups/validate,/hypernodes/validate,/cronjobs/validate"
+  namespace_queue_enable: ${NAMESPACE_QUEUE_ENABLE}
```
JesseStutler, line 415: The NamespaceQueue E2E suite enables the feature here but still loads `config/volcano-scheduler-ci.conf`, which configures `proportion`. It never switches to `capacity` with `enableHierarchy: true`. Therefore these tests only prove that workloads can reference a NamespaceQueue as a flat queue; they do not exercise the supported hierarchical scheduling semantics or verify ancestor resource constraints. Run this suite with `capacity` and `enableHierarchy: true`, and add an E2E case that verifies a parent Queue/NamespaceQueue resource limit is enforced.

installer/helm/chart/volcano/templates/scheduler.yaml
```diff
@@ -5,6 +5,7 @@
 {{ $scheduler_main_csc := or .Values.custom.scheduler_main_csc .Values.custom.default_csc }}
 {{ $scheduler_ns := or .Values.custom.scheduler_ns .Values.custom.default_ns }}
 {{ $scheduler_name := .Values.custom.scheduler_name }}
+{{ $scheduler_feature_gates := include "volcano.featureGates" (dict "gates" .Values.custom.scheduler_feature_gates "namespaceQueueEnabled" .Values.custom.namespace_queue_enable) }}
```
JesseStutler, line 8: Enabling NamespaceQueue through the chart only adds the feature gate; the default scheduler configuration still uses `proportion`. The design states that NamespaceQueue is supported only by the hierarchical capacity path, so the chart currently exposes an enabled but unsupported configuration. When NamespaceQueue is enabled, install an NQ-compatible `capacity` configuration with `enableHierarchy: true`, or fail configuration validation unless the user supplies one. This must not silently change scheduler behavior when the feature is disabled.

pkg/scheduler/cache/cache.go
```diff
@@ -483,6 +542,10 @@ func newDefaultAndRootQueue(vcClient vcclient.Interface, defaultQueue string) {
 				Weight:      1,
 			},
 		}
+		if name == vcv1beta1.DefaultQueue &&
+			utilfeature.DefaultFeatureGate.Enabled(features.NamespaceQueue) {
+			newQueue.Spec.AllowedNamespaces = []string{"*"}
```
JesseStutler, line 543: This creates different behavior for fresh installations and upgrades. A newly created `default` Queue gets `allowedNamespaces: ["*"]`, but an existing default Queue is returned above without migration. After upgrade, a NamespaceQueue with the default parent `cluster/default` is therefore unauthorized in every namespace, while the same object works on a fresh installation. Do not silently grant wildcard access to existing Queues, but make authorization behavior consistent and provide an explicit administrator opt-in/migration path with an upgrade test. The default NamespaceQueue parent must not be usable only on fresh installations.

pkg/scheduler/cache/cache.go
```diff
@@ -1527,7 +1611,22 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
...
+			continue
+		}
+
 		snapshot.Queues[value.UID] = value.Clone()
+		if _, found := sc.AccountingQueues[value.UID]; !found {
+			snapshot.AccountingQueues[value.UID] = value.Clone()
+		}
+	}
+	for _, value := range sc.AccountingQueues {
+		snapshot.AccountingQueues[value.UID] = value.Clone()
```
JesseStutler, line 1625: Retaining inactive NamespaceQueues for allocation draining is necessary, but the current accounting model changes the hot path for every existing cluster. All cluster Queues are mirrored into `AccountingQueues`, active Queues are cloned into both snapshot maps, and all Jobs are included in `AccountingJobs`, including when the NamespaceQueue feature is disabled. Keep the existing Queue/Job snapshot path unchanged when the feature is disabled, and make accounting data supplemental: retain only inactive NamespaceQueues, the jobs that still require draining, and any ancestors required for propagation. Please also cover the multi-queue allocation/memory cost; the existing single-queue allocation benchmark does not exercise this overhead.

pkg/scheduler/framework/session.go
```diff
@@ -498,60 +508,84 @@ func addNodeSharableDeviceUsage(ssn *Session, task *api.TaskInfo) {
...
 }
 
-// updateQueueStatus updates allocated field in queue status on session close.
+// updateQueueStatus aggregates allocated resources through active and inactive
+// queue ancestors, then updates only scheduler-owned status fields on session
+// close. Accounting views are required because a closing NamespaceQueue is no
+// longer schedulable but must still release its recorded allocation.
 func updateQueueStatus(ssn *Session) {
 	rootQueue := api.QueueID("root")
+	queues := ssn.AccountingQueues
```
JesseStutler, line 512: This replaces the existing queue-status accounting data source for all Queues, although the additional behavior is only required for NamespaceQueue hierarchy and inactive-queue draining. Preserve the existing Cluster Queue status path when NamespaceQueue is disabled, and isolate the new logic to merging supplemental NamespaceQueue jobs/queues and propagating their allocations to ancestors. Add regression coverage demonstrating equivalent Cluster Queue allocated/DRA status behavior with the feature disabled; NamespaceQueue-specific tests alone do not protect the existing path from this rewrite.

pkg/util/namespace_queue.go
```diff
@@ -0,0 +1,260 @@
...
+		return schedulingv1beta1.QueueStateClosing
+	default:
+		return schedulingv1beta1.QueueStateUnknown
+	}
+}
+
+// IsNamespaceQueueWorkloadDrained reports whether no active PodGroup remains.
+func IsNamespaceQueueWorkloadDrained(status schedulingv1beta1.NamespaceQueueStatus) bool {
+	return status.Unknown == 0 &&
+		status.Pending == 0 &&
```
JesseStutler, line 152: Pending/Inqueue PodGroups block draining here, but once the NamespaceQueue enters Closing it is made not ready and removed from the scheduler's active Queue/Job view. Those workloads can no longer be scheduled, so they cannot naturally leave Pending/Inqueue; a normal queue close or deletion can remain Closing forever and keep the finalizer indefinitely. Define a terminating drain policy that can converge: either allow already admitted workloads to finish while rejecting new admission, or explicitly cancel/remove non-running workloads. Do not count states that the close operation itself makes unable to progress as permanent drain blockers.

pkg/controllers/framework/doc.go
JesseStutler, line 1: Why is there a separate doc.go in this package?

pkg/scheduler/cache/cache.go
```diff
@@ -1550,7 +1649,10 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 		clonedJob := value.Clone()
 
 		cloneJobLock.Lock()
-		snapshot.Jobs[value.UID] = clonedJob
+		snapshot.AccountingJobs[value.UID] = clonedJob
+		if _, active := snapshot.Queues[value.Queue]; active {
```
JesseStutler, line 1649: This makes inactive-queue jobs accounting-only, but `closeSession` still constructs `JobUpdater` exclusively from `ssn.Jobs`. Once a NamespaceQueue enters Closing, its running jobs disappear from that map, so their PodGroup status is no longer reconciled. When the pods finish naturally, the PodGroup can remain Running; the controller then keeps `workloadDrained=false`, leaving the NamespaceQueue permanently Closing and preventing finalizer removal. Accounting-only jobs must continue through PodGroup status reconciliation without being exposed to scheduling actions. Add an E2E that closes a NamespaceQueue with a short-running job and verifies it reaches Closed after the job finishes naturally, without explicitly deleting the Job.

pkg/controllers/namespacequeue/namespacequeue_controller_validation.go
```diff
@@ -0,0 +1,194 @@
...
+				return commonutil.NamespaceQueueReasonDuplicateClusterAttachment,
+					fmt.Sprintf("namespace %q already attaches NamespaceQueue %q to Queue %q", queue.Namespace, child.Name, parent.Name), nil
+			}
+		}
+	}
+
+	if err := c.validateCapabilityAgainstAncestors(queue); err != nil {
+		return commonutil.NamespaceQueueReasonParentConstraintViolation, err.Error(), nil
+	}
+	if err := c.validateChildrenAggregate(parent); err != nil {
```
JesseStutler: The indexed descendant traversal is now linear, but this call still makes a parent change quadratic overall. `enqueueNamespaceQueueTree(parent)` enqueues every descendant, and each direct child independently scans the same sibling set again in `validateChildrenAggregate(parent)`. With N NamespaceQueues attached to one cluster Queue, a single add/resize triggers O(N) reconciles performing O(N) repeated aggregate work; the 10-QPS workqueue further delays convergence. Compute/reuse the aggregate once per affected parent, or only revalidate queues whose result can change, and add a scale benchmark for this path.

JesseStutler (commented):
Review continue: the two comments below are blocking lifecycle-convergence and scale-path issues.

pkg/controllers/framework/interface.go
```diff
@@ -21,44 +21,80 @@ import (
...
+	// Initialize wires clients, informers, and controller-specific handlers.
 	Initialize(opt *ControllerOption) error
-	// Run run the controller
+	// Run starts the controller and stops when stopCh is closed.
 	Run(stopCh <-chan struct{})
 }
 
+// FeatureGatedController associates a controller with a feature gate. The
+// controller manager skips initialization while the gate is disabled.
+type FeatureGatedController interface {
```
JesseStutler: Introducing `FeatureGatedController` and cleanup-only mode for NamespaceQueue unnecessarily complicates and changes the shared controller framework. Use a regular feature-gate check to control whether the NamespaceQueue controller starts. Disabling the gate while NamespaceQueues still exist is an exceptional operation; document that users must clean up these resources first instead of adding a generic cleanup framework.

installer/helm/chart/volcano/templates/namespacequeue_editor_role.yaml
JesseStutler, line 1: Currently there are no corresponding binding for the namespace-editor-role and viewer-role, I think there are no need to add these two roles, if admins needs grant permissions, they need to define by themseleves and bind

installer/helm/chart/volcano/templates/agent_scheduler.yaml
```diff
@@ -5,6 +5,7 @@
 {{ $agent_scheduler_main_csc := or .Values.custom.agent_scheduler_main_csc .Values.custom.default_csc }}
 {{ $agent_scheduler_ns := or .Values.custom.agent_scheduler_ns .Values.custom.default_ns }}
 {{ $agent_scheduler_name := .Values.custom.agent_scheduler_name }}
+{{ $scheduler_feature_gates := include "volcano.featureGates" (dict "gates" .Values.custom.scheduler_feature_gates "namespaceQueueEnabled" .Values.custom.namespace_queue_enable) }}
```
JesseStutler: Currently there is no queue concept in the agent scheduler, there is no need to define namespace queue feature gate for the agent scheduler

hack/generate-yaml.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #
-# Copyright 2021 The Volcano Authors.
+# Copyright 2026 The Volcano Authors.
```
JesseStutler: There is no need to change the copyright year

staging/src/volcano.sh/apis/pkg/apis/scheduling/types.go
```diff
@@ -333,6 +333,96 @@ type Queue struct {
...
+	// +optional
+	Priority int32 `json:"priority,omitempty" protobuf:"varint,6,opt,name=priority"`
+
+	// +optional
+	DequeueStrategy DequeueStrategy `json:"dequeueStrategy,omitempty" protobuf:"bytes,7,opt,name=dequeueStrategy"`
+
+	// State specifies the desired lifecycle state of this NamespaceQueue.
+	// Closing is an observed state and cannot be requested directly.
+	// +optional
+	State QueueState `json:"state,omitempty" protobuf:"bytes,8,opt,name=state"`
```
JesseStutler: @gitGurugu Why is there a duplicate `State` field in both spec and status? The semantic is ambiguous, if you really need the state in the spec, it should be `desiredState` instead

pkg/controllers/namespacequeue/doc.go
JesseStutler, line 1: Same as this comment: https://github.com/volcano-sh/volcano/pull/5590/changes/BASE..1d19785d6683251a1d12ce6efb427be147a038c9#r3851224133

pkg/controllers/metrics/namespacequeue.go
JesseStutler, line 1: Metrics like this look like they offer very low returns, users usually don't care about such reconcile metrics.

pkg/controllers/namespacequeue/namespacequeue_controller_action.go
```diff
@@ -0,0 +1,630 @@
...
+		},
+	)
+}
+
+// updateNamespaceQueueStatus updates only controller-owned status fields and
+// preserves scheduler-owned Allocated and Reservation across conflict retries.
+func (c *namespaceQueueController) updateNamespaceQueueStatus(
+	nq *schedulingv1beta1.NamespaceQueue,
+) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
JesseStutler, line 422: `NamespaceQueue.status` now has two concurrent writers: this controller updates lifecycle/counters/conditions, while the scheduler updates `Allocated`/`Reservation` with another `Get + UpdateStatus` retry loop. They can conflict even though they own different fields, and retrying this whole function repeats PodGroup counting, parent-chain validation, and sibling aggregation on every conflict. Split reconciliation from persistence and make both writers patch/apply only their owned status fields with separate field managers. If `UpdateStatus` is retained, limit `RetryOnConflict` to re-GET, merge the controller-owned fields, recompute only lifecycle fields that depend on the latest `Allocated`/`Reservation`, and update; record events after the write succeeds.

pkg/controllers/namespacequeue/namespacequeue_controller_action.go
```diff
@@ -0,0 +1,630 @@
...
+// reconcileNamespaceQueue applies finalizer, status, and deletion invariants
+// for one NamespaceQueue. Finalizer removal is delayed until child queues and
+// active workloads are drained. Scheduler-owned runtime resources are also
+// required while the scheduler feature is active.
+func (c *namespaceQueueController) reconcileNamespaceQueue(nq *schedulingv1beta1.NamespaceQueue) error {
+	if c.cleanupOnly && nq.DeletionTimestamp == nil {
+		return nil
+	}
+	if nq.DeletionTimestamp == nil {
+		if err := c.ensureNamespaceQueueFinalizer(nq); err != nil {
```
JesseStutler: The delete webhook already rejects a NamespaceQueue that has children or is not Closed and drained, while this finalizer performs no cleanup and repeats the same checks after deletion is admitted. Remove the finalizer and the deletion-specific/cleanup-only controller path; keep the explicit Close/drain flow followed by webhook-validated Delete, and cover that unsafe deletion is rejected while a Closed/drained queue is deleted directly. A finalizer is only needed if Delete itself starts asynchronous draining, in which case the webhook must allow that deletion instead of requiring it to be drained first. Do not keep both models.

pkg/scheduler/api/queue_info.go
```diff
@@ -23,72 +23,271 @@ limitations under the License.
...
+	NamespaceQueueScope QueueScope = "namespace"
+
+	// DefaultQueueWeight is used when NamespaceQueue has no independent weight field.
+	DefaultQueueWeight int32 = 1
+)
+
+// QueueInfo is the normalized scheduler view of a cluster Queue or
+// NamespaceQueue. Queue and NamespaceQueue are mutually exclusive source
+// objects; plugins should use normalized fields instead of inspecting them.
 type QueueInfo struct {
```
JesseStutler, line 54: NamespaceQueue should not require this broad rewrite of `QueueInfo` and the existing scheduler paths. This changes the Cluster Queue data representation and forces corresponding changes in cache/framework as well as proportion, DRF, nodegroup, and extender, although NamespaceQueue is only supported by capacity with hierarchy enabled. That materially increases the functional and performance regression risk for existing configurations and makes this feature much harder to review safely.

Please preserve the existing Cluster Queue `QueueInfo` and plugin paths, and introduce only the minimal NamespaceQueue representation/adapter needed by capacity hierarchy. In particular, avoid keeping a full source object while duplicating the same fields in a normalized form, which creates two sources of truth and additional snapshot-copy cost. If a unified scheduler queue abstraction is independently necessary, make it a separate refactor with explicit compatibility and performance validation instead of coupling it to NamespaceQueue.

pkg/scheduler/cache/cache.go
```diff
@@ -1550,7 +1669,10 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
...
```
JesseStutler, line 1649: This is unnecessarily complicated for the requirement being solved. The only special case is that a non-schedulable NamespaceQueue must stop admitting new jobs while its existing workloads remain visible for drain and status accounting. That does not justify duplicating every Queue and Job into parallel `AccountingQueues`/`AccountingJobs` views or changing the existing Cluster Queue path.

Keep the existing `Queues`/`Jobs` scheduling path unchanged. Store only the NamespaceQueues and jobs excluded from scheduling in explicitly scoped collections such as `NonSchedulableNamespaceQueues` and `NonSchedulableNamespaceQueueJobs`, used only by the NamespaceQueue drain/status path. Active objects remain in the existing maps, so there is no second complete copy and the semantics are clear. Also log when a job is excluded because its NamespaceQueue or parent chain is not schedulable; the current missing-queue log does not cover this case.

pkg/controllers/namespacequeue/namespacequeue_controller_validation.go
gitGurugu: I will cache the aggregate validation result per parentRef and invalidate it when the relevant Queue or NamespaceQueue spec changes:
```
type parentAggregateResult struct {
    validationError     error
    attachmentNamesByNS map[string][]string
}

type namespaceQueueController struct {
    aggregateMu    sync.Mutex
    aggregateCache map[ResolvedQueueReference]parentAggregateResult
}
```
```
func (c *namespaceQueueController) validateChildrenAggregate(
    parentRef ResolvedQueueReference,
) error {
    c.aggregateMu.Lock()
    defer c.aggregateMu.Unlock()

    if result, ok := c.aggregateCache[parentRef]; ok {
        return result.validationError
    }

    result, err := c.computeParentAggregate(parentRef)
    if err != nil {
        return err
    }
    c.aggregateCache[parentRef] = result
    return result.validationError
}
```
Descendant re-enqueueing will remain unchanged because readiness and inherited constraints still need to converge for every descendant. The optimization only reuses the parent aggregate calculation.
This reduces the aggregate path from O(N²) to O(N) per affected parent. I will also add cache invalidation tests and a benchmark for 100, 1000, and 5000 child queues.

hack/generate-yaml.sh
gitGurugu: I have rewind it

pkg/controllers/framework/interface.go
gitGurugu: Yes，I have delete it.

installer/helm/chart/volcano/templates/namespacequeue_editor_role.yaml
gitGurugu: ok

hack/run-e2e-kind.sh
gitGurugu: I have solved it.

installer/helm/chart/volcano/templates/scheduler.yaml
gitGurugu: solved it！

pkg/scheduler/cache/cache.go
gitGurugu: I got it.

gitGurugu: Indeed, permissions are escalated—the scheduler overrides the admin's YAML settings.

hack/generate-yaml.sh
gitGurugu: solve it

staging/src/volcano.sh/apis/pkg/apis/scheduling/types.go
gitGurugu: Does this approach make sense? I have renamed spec.state to spec.desiredState and introduced a separate QueueDesiredState type that only allows Open and Closed.

JesseStutler: Yes, but you don't need to define a seperate `QueueDesiredState` type, the original QueueState is enough and you can add Enum kubebuilder validation for the only Open/Closed type

pkg/controllers/framework/doc.go
gitGurugu: I remove it.

pkg/controllers/namespacequeue/doc.go
gitGurugu: I remove it

pkg/controllers/metrics/namespacequeue.go
gitGurugu: I redesign a set of metrics for `NamespaceQueue` to expose queue admission status, workload backlog and resource‑usage information:

1. `volcano_namespacequeue_ready{queue_name}`: Indicate whether this NamespaceQueue is able to accept incoming workloads.
2. `volcano_namespacequeue_blocked{queue_name, reason}`: If `volcano_namespacequeue_ready` is false, this metric exposes why the queue cannot accept workloads.
3. Re‑use existing `volcano_queue_pod_group_{pending,inqueue,running,unknown,completed}_count` metrics to reflect workload states and backlog of NamespaceQueue.
4. Reuse canonical scheduler queue resource metrics with `queue_name` label for resource requested/used statistics; avoid introducing duplicate NamespaceQueue‑specific resource metrics.

JesseStutler:
@gitGurugu Please resolve the CI failure, currently the namespace queue CI is failed

pkg/util/queue_reference.go
```diff
@@ -0,0 +1,133 @@
...
+// reference. An unqualified workload reference selects a cluster Queue.
+func ResolveWorkloadQueueReference(
+	workloadNamespace, reference, defaultQueue string,
+) (ResolvedQueueReference, error) {
+	if reference == "" {
+		reference = defaultQueue
+	}
+
+	if !strings.Contains(reference, "/") {
+		if reference == "" {
```
JesseStutler, line 65: Will never goto this branch

pkg/controllers/job/job_controller_util.go
```diff
@@ -149,7 +150,10 @@ func createJobPod(job *batch.Job, template *v1.PodTemplateSpec, ix int, jobForwa
 	pod.Labels[batch.JobNameKey] = job.Name
 	pod.Labels[batch.TaskSpecKey] = tsKey
 	pod.Labels[batch.JobNamespaceKey] = job.Namespace
-	pod.Labels[batch.QueueNameKey] = job.Spec.Queue
+	delete(pod.Labels, batch.QueueNameKey)
```
JesseStutler, line 153: Why need to add a delete behavior? Also if you need to validate the queue name, you should do it in the `getJobForwardingQueue` instead

pkg/controllers/namespacequeue/namespacequeue_controller_action.go
```diff
@@ -0,0 +1,551 @@
...
+		if depth > c.maxNamespaceQueueDepth {
+			parentConditionFailure(&result, commonutil.NamespaceQueueReasonHierarchyDepthExceeded, fmt.Sprintf(
+				"NamespaceQueue hierarchy depth %d exceeds maximum depth %d",
+				depth,
+				c.maxNamespaceQueueDepth,
+			))
+			return result, nil
+		}
+
+		parentAuthorized := apiMeta.FindStatusCondition(
```
JesseStutler, line 245: nit: package name should use lower case

pkg/controllers/namespacequeue/namespacequeue_controller_handler.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+}
+
+func (c *namespaceQueueController) enqueueNamespaceQueueTrees(
+	includeRoots bool,
+	roots ...controllerutil.ResolvedQueueReference,
+) {
+	visited := make(map[controllerutil.ResolvedQueueReference]struct{}, len(roots))
+	frontier := make([]controllerutil.ResolvedQueueReference, 0, len(roots))
+	for _, root := range roots {
```
JesseStutler, line 439: nit: the name `root/roots` may not be clear for me and is ambiguous with the original `root` queue concept, `startRefs` may be more clear for me

docs/design/namespace-queue.md
JesseStutler, line 1: Better add a user-guide doc to teach users how to use namespace queue (but I will not block it in this PR if time is limit, can raise it in later PR)

pkg/controllers/metrics/namespacequeue.go
JesseStutler: I still feel these two metrics are of somewhat limited utility. Why wouldn't a user simply query the `condition` to check if a specific queue is ready? Is there actually a scenario where one needs to know which queues within a namespace are blocked or ready? I’d also like to hear @devzizu 's opinion—what kind of metrics do you find most useful when working with namespace queues?

pkg/scheduler/api/queue_info_namespacequeue_test.go
JesseStutler, line 1: nit: better rename this file to `queue_info_test.go`, previously we lacked a unit test file for the `queue_info.go`, and suggest to use table-driven style to write unit tests, it's more extensible and can write more cases

pkg/scheduler/api/queue_info.go
```diff
@@ -59,21 +73,92 @@ func NewQueueInfo(queue *scheduling.Queue) *QueueInfo {
...
+	}
+	parent, err := commonutil.ResolveNamespaceQueueParentReference(namespaceQueue.Namespace, namespaceQueue.Spec.Parent)
+	if err != nil {
+		return nil, err
+	}
+	parentName := parent.Name
+	if parent.Scope == commonutil.NamespaceQueueReferenceScope {
+		parentName = string(NamespaceQueueID(parent.Namespace, parent.Name))
+	}
+	queue := &scheduling.Queue{
```
JesseStutler, line 99: 👍 It's a clever idea, this way, most of the scheduler's implementation won't need to be changed. If differences between namespace queues and cluster queues arise in the future, we can then consider a new common struct.

pkg/scheduler/cache/cache.go
```diff
@@ -375,6 +376,21 @@ func (su *defaultStatusUpdater) UpdatePodGroup(pg *schedulingapi.PodGroup) (*sch
...
+			Resource: queue.Queue.Status.Reservation.Resource.DeepCopy(),
+		}
+		newNamespaceQueue := &vcv1beta1.NamespaceQueue{}
+		if err := schedulingscheme.Scheme.Convert(updated, newNamespaceQueue, nil); err != nil {
+			return err
+		}
+		_, err := su.vcclient.SchedulingV1beta1().NamespaceQueues(updated.Namespace).
+			UpdateStatus(context.TODO(), newNamespaceQueue, metav1.UpdateOptions{})
+		return err
+	}
```
JesseStutler, line 393: For `Deepcopy`, we need to be careful, excessive use can lead to performance degradation. I have noticed that here: https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/scheduler/framework/session.go#L562, the `updateQueueStatus` has already modified the queue ref's status, and in the `NewNamespaceQueueInfo` you have already deepcopy once, in the session, we will not modify the namespacequeue's spec, so it's not so necessary to DeepCopy all the fields I think. If I'm wrong, please correct me, suggested code:

```suggestion
	if queue != nil && queue.NamespaceQueue != nil {
	newNamespaceQueue := &vcv1beta1.NamespaceQueue{}
	if err := schedulingscheme.Scheme.Convert(
		queue.NamespaceQueue, newNamespaceQueue, nil,
	); err != nil {
		return err
	}

	newNamespaceQueue.Status.Allocated = queue.Queue.Status.Allocated
	newNamespaceQueue.Status.Reservation = vcv1beta1.Reservation{
		Nodes:    queue.Queue.Status.Reservation.Nodes,
		Resource: queue.Queue.Status.Reservation.Resource,
	}

	_, err := su.vcclient.SchedulingV1beta1().
		NamespaceQueues(newNamespaceQueue.Namespace).
		UpdateStatus(context.TODO(), newNamespaceQueue, metav1.UpdateOptions{})
	return err
}
```

And better add some comments to explain branches for namespace queue update and cluster queue update

installer/volcano-development.yaml
```diff
@@ -142,13 +148,14 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
         - args:
-            - --enabled-admission=/jobs/mutate,/jobs/validate,/podgroups/validate,/queues/mutate,/queues/validate,/hypernodes/validate,/cronjobs/validate
+            - --enabled-admission=/jobs/mutate,/jobs/validate,/namespacequeues/validate,/podgroups/mutate,/podgroups/validate,/queues/mutate,/queues/validate,/hypernodes/validate,/cronjobs/validate
```
JesseStutler, line 151: Better not to enable namespace queue validation by default in next version, since the feature gate for namespace queue is disabled by default

pkg/webhooks/router/queue_reference.go
JesseStutler, line 1: nit: this util funcs should be better to put under util rather than in `router`

pkg/webhooks/admission/cronjobs/validate/admit_cronjob.go
```diff
@@ -106,18 +111,45 @@ func AdmitCronjobs(ar admissionv1.AdmissionReview) *admissionv1.AdmissionRespons
 }
 func validateCronJobCreate(cronjob *v1alpha1.CronJob, reviewResponse *admissionv1.AdmissionResponse) string {
 	msg := validateCronJob(cronjob)
+	if queueMsg := validateCronJobQueue(cronjob); queueMsg != "" {
```
JesseStutler, line 114: Better use `strings.Builder`

pkg/webhooks/admission/pods/validate/admit_pod.go
JesseStutler, line 1: We don't need pod-level queue validation, podgroup-level for queue is enough @gitGurugu

installer/helm/chart/volcano/values.yaml
```diff
@@ -68,7 +69,7 @@ custom:
   scheduler_task_updater_worker_num: ~
   scheduler_percentage_nodes_to_find: ~
   agent_scheduler_worker_count: 1
-  enabled_admissions: "/jobs/mutate,/jobs/validate,/podgroups/validate,/queues/mutate,/queues/validate,/hypernodes/validate,/cronjobs/validate"
+  enabled_admissions: "/jobs/mutate,/jobs/validate,/namespacequeues/validate,/podgroups/mutate,/podgroups/validate,/queues/mutate,/queues/validate,/hypernodes/validate,/cronjobs/validate"
```
JesseStutler, line 72: Better not to enable namespace queue validation by default in next version, since the feature gate for namespace queue is disabled by default, if you can use go template to automatically inject the `/namespacequeues/validate` into the `enabled_admissions` according to the feature gate's value, is better

installer/helm/chart/volcano/config/volcano-scheduler-namespacequeue-ci.conf
JesseStutler, line 1: Do these two new files have any connection to `namespacequeue`? Presumably, only `capacity` requires a hierarchy, but the scheduler configuration can be modified before the e2e tests run. You can refer to how `gangevict` initialize their scheduler conf

test/e2e/namespacequeue/namespacequeue_test.go
JesseStutler, line 1: Better also add a case for validation that job/podgroup can't submit to the `NotReady` namespace queue

JesseStutler (commented):
Overall the implementation is good from my side, but there are some small comments that still needs to be solved, please check @gitGurugu especially we don't need pod level's queue validation, webhooks impact performance, so we need to be quite careful.

installer/helm/chart/volcano/values.yaml
```diff
@@ -293,6 +294,9 @@ custom:
   # Specify feature gates for components
   scheduler_feature_gates: ~
   admission_feature_gates: ~
+  controller_feature_gates: ~
+  namespace_queue_enable: false
+  namespace_queue_max_depth: 5
```
hajnalmt, line 299: Why do we expose this in the helm chart if we have a validated proper default on the application side? I don't see the benefit of this. It is confusing, for `queue-max-depth` we don't have it, we shouldn't have it for `namespace-queue-max-depth`.

installer/helm/chart/volcano/templates/_helpers.tpl
```diff
@@ -10,6 +10,15 @@ bases
 {{- end -}}
 {{- end -}}
 
+{{/* Validate and return the NamespaceQueue hierarchy depth limit. */}}
```
hajnalmt, line 13: We have application level validation we don't need this helper and validation in general.

installer/helm/chart/volcano/templates/webhooks.yaml
```diff
@@ -330,6 +330,7 @@ webhooks:
           - v1
         operations:
           - CREATE
+          - UPDATE
```
hajnalmt, line 333: This can have potential perform impacts, don't forget to remove it from here too.

docs/design/namespace-queue.md
```diff
@@ -78,6 +78,13 @@ type NamespaceQueue struct {
 
 `NamespaceQueueSpec` defines the desired state of a `NamespaceQueue`.
 
+The controller reports the observed lifecycle in `status.state`. A queue is
```
hajnalmt, line 81: Please remove the multiline wrap here. The other parts are not wrapped.

docs/design/metrics.md
```diff
@@ -82,6 +80,12 @@ This metrics describe internal state of volcano.
...
 |------------------------------------------------------|-----------------|-------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
 | `controller_job_to_pod_creation_latency_milliseconds` | Histogram      | None                                                                                      | Latency from VCJob creation to pod created in milliseconds                                   |
+| `namespacequeue_ready`                               | Gauge           | `queue_name`=&lt;namespace&gt;/&lt;name&gt;                                              | Whether a NamespaceQueue is ready for scheduling                                              |
+| `namespacequeue_blocked`                             | Gauge           | `queue_name`=&lt;namespace&gt;/&lt;name&gt;, `reason`=&lt;reason&gt;                       | Whether a NamespaceQueue is blocked, by controller condition reason                          |
+
+The existing `queue_pod_group_*_count` metrics also use the canonical
```
hajnalmt, line 86: This shall be a single line not 3, AI-s like to wrap the lines, but it makes reading harder. Also this is not consistent with the rest of the file.

pkg/controllers/metrics/namespacequeue.go
hajnalmt: I prefer not to add this too, because I don't see any use-case I would use it. We can handle this separately in an improvement/feature request if there is a need.

pkg/controllers/metrics/queue.go
```diff
@@ -100,10 +100,28 @@ func DeleteQueueMetrics(queueName string) {
...
 }
 
+// UpdateQueueMetrics updates queue workload metrics from the latest status.
 func UpdateQueueMetrics(queueName string, queueStatus *v1beta1.QueueStatus) {
-	UpdateQueuePodGroupPendingCount(queueName, queueStatus.Pending)
-	UpdateQueuePodGroupRunningCount(queueName, queueStatus.Running)
-	UpdateQueuePodGroupUnknownCount(queueName, queueStatus.Unknown)
-	UpdateQueuePodGroupInqueueCount(queueName, queueStatus.Inqueue)
-	UpdateQueuePodGroupCompletedCount(queueName, queueStatus.Completed)
+	UpdateQueuePodGroupCounts(
```
hajnalmt, line 105: Why did you refactor this? I don't see any value in adding one more function call that does the same.

pkg/controllers/namespacequeue/namespacequeue_controller_validation.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := commonutil.ValidateQueueResourceRelations(
+		resources.capability, resources.guarantee, resources.deserved,
+	); err != nil {
+		return commonutil.NamespaceQueueReasonParentConstraintViolation, err.Error(), nil
+	}
+
+	if err := c.validateCapabilityAgainstAncestors(namespaceQueue); err != nil {
+		return commonutil.NamespaceQueueReasonParentConstraintViolation, err.Error(), nil
+	}
+	if err := c.validateChildrenAggregate(parentRef); err != nil {
```
hajnalmt, line 65: This can still blow up right? You haven't fixed Jesse's comment yet:
https://github.com/volcano-sh/volcano/pull/5590#discussion_r3851297002

pkg/webhooks/admission/podgroups/validate/validate_podgroup.go
```diff
@@ -103,19 +111,19 @@ func validatePodGroup(pg *schedulingv1beta1.PodGroup) string {
...
-
-	if queue.Status.State != schedulingv1beta1.QueueStateOpen {
-		return fmt.Sprintf("can only submit PodGroup to queue with state `Open`, queue `%s` status is `%s`. ",
-			queue.Name, queue.Status.State)
+	if err := router.ValidateWorkloadQueueReference(
+		workloadNamespace,
+		queueReference,
+		schedulingv1beta1.DefaultQueue,
+		config,
+		router.QueueReferenceValidationOptions{},
```
hajnalmt, line 124: Why do we have different behaviour here and in Job/CronJob?
`QueueReferenceValidationOptions` ? Don't  we want podgroups to be submitted to leaf queues only as well?

pkg/webhooks/router/queue_reference.go
```diff
@@ -0,0 +1,201 @@
...
+	NamespaceQueueParentIndexName = "namespaceQueueParent"
+)
+
+// ResolvedQueueReference identifies the queue resource selected by a workload reference.
+type ResolvedQueueReference = commonutil.ResolvedQueueReference
+
+// QueueReferenceValidationOptions controls compatibility-sensitive workload
+// queue validation. Job validation requires a cluster Queue leaf; other
+// workload paths retain the existing cluster Queue behavior.
+type QueueReferenceValidationOptions struct {
```
hajnalmt, line 52: I start to understand this finally, but please help me if I am correct, that this is purely for backward compatibility right? In the logic below we won't verify that namespaceQueues are admitted to namespace leaf queues neither anything else:

Job → cluster Queue        leaf check
PodGroup → cluster Queue   no leaf check
Job → NamespaceQueue       no leaf check
PodGroup → NamespaceQueue  no leaf check

I feel that there should be a better way to do this by defining one leaf-eligibility policy over the combined hierarchy and enforce it consistently, with the scheduler remaining authoritative when topology changes after admission? I won't block on this but this is an admission gap.

pkg/webhooks/router/queue_reference.go
```diff
@@ -0,0 +1,201 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to get child queues for queue %s: %w", queue.Name, err)
+			}
+			if len(children) > 0 {
+				return fmt.Errorf("can only submit workload to leaf queue, queue `%s` has %d child queues",
+					queue.Name, len(children))
+			}
+		}
+
+	case NamespaceQueueReferenceScope:
```
hajnalmt, line 122: On a second thought, at least the leaf `NamespaceQueue` check should be added, and I block on this because the capacity builds a hierarchy containing `NamespaceQueue` too. A Job submitted directly to a cluster Queue with a `NamespaceQueue `child passes admission, but `buildHierarchicalQueueAttrs` then returns false for the whole session because that Job's queue is non-leaf. 

The NamespaceQueue branch shoudl not acccept jobs on `NamespaceQueue` parents. Can we make admission and the scheduler agree, cover adding a child to an occupied parent, we should add a test for this too.

test/e2e/admission/podgroup_validation_test.go
```diff
@@ -49,16 +49,20 @@ var _ = ginkgo.Describe("PodGroup Validating E2E Test", func() {
...
-		time.Sleep(100 * time.Millisecond)
-
-		// Get the latest version of the queue after potential mutations
-		updatedQueue, err := testCtx.Vcclient.SchedulingV1beta1().Queues().Get(context.TODO(), createdQueue.Name, metav1.GetOptions{})
-		gomega.Expect(err).NotTo(gomega.HaveOccurred())
-
-		// Update queue status to Open (since status is typically set by controller)
-		updatedQueue.Status.State = schedulingv1beta1.QueueStateOpen
-		_, err = testCtx.Vcclient.SchedulingV1beta1().Queues().UpdateStatus(context.TODO(), updatedQueue, metav1.UpdateOptions{})
+		// Update the latest Queue version because the queue controller may update status concurrently.
```
hajnalmt, line 52: Is there a reason for this test change? It seems so unrelated to the feature.

test/e2e/namespacequeue/namespacequeue_test.go
hajnalmt: ++

test/e2e/namespacequeue/namespacequeue_test.go
```diff
@@ -0,0 +1,1711 @@
...
+	DeferCleanup(func() {
+		fixture.cleanupJobs()
+		fixture.cleanupPodGroupsAndPods()
+		fixture.cleanupNamespaceQueues()
+		e2eutil.CleanupTestContext(ctx)
+	})
+	return fixture
+}
+
+func restartVolcanoComponent(labelSelector string) error {
```
hajnalmt, line 818: Everything below here belongs under a new `e2e/util/namespacequeue.go` file.
or into an appropriate file under util.

test/e2e/namespacequeue/namespacequeue_test.go
```diff
@@ -0,0 +1,1711 @@
...
+	requestCtx, cancel := context.WithTimeout(context.Background(), apiRequestTimeout)
+	pods, err := e2eutil.KubeClient.CoreV1().Pods(metav1.NamespaceAll).List(
```
hajnalmt, line 820: In the tests we are working on fixed namespaces, why are you querying all?

test/e2e/namespacequeue/namespacequeue_test.go
```diff
@@ -0,0 +1,1711 @@
...
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("child NamespaceQueues"))
+
+		Expect(fixture.deleteNamespaceQueueEventually(childName)).NotTo(HaveOccurred())
+		Expect(waitNamespaceQueueDeleted(fixture.ctx, childName)).NotTo(HaveOccurred())
+		Expect(fixture.deleteNamespaceQueueEventually(parentName)).NotTo(HaveOccurred())
+		Expect(waitNamespaceQueueDeleted(fixture.ctx, parentName)).NotTo(HaveOccurred())
+	})
+
+	It("recovers NamespaceQueue reconciliation after controller restart", func() {
```
hajnalmt, line 555: I am not sure we should cover tests like this in the e2e test suites. I am not particularly against it, but feels overkill for me. I would love opinions from others.

test/e2e/admission/queue_validation_test.go
```diff
@@ -40,7 +40,7 @@ var _ = ginkgo.Describe("Queue Validating E2E Test", func() {
 
 		queue := &schedulingv1beta1.Queue{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "test-queue-open",
+				Name: "test-queue-open-validation",
```
hajnalmt, line 43: Same question here. I would prefer PRs especially big ones like this to be focused.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -1073,7 +1069,7 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
 			queue := ssn.Queues[job.Queue]
 			attr := &queueAttr{
 				queueID: queue.UID,
-				name:    queue.Name,
+				name:    string(queue.UID),
```
hajnalmt, line 1072: Not in scope here, let this stay as this is. But after your PR gets merged I prefer to throw this field out fully from `queueAttr`. 

It was confusing before, now its straight out misleading, because `Queueinfo.Name` is not the same as `queueAttr.name` anymore.

pkg/scheduler/api/queue_info.go
hajnalmt: Yeah this is genius. 👍

hajnalmt (commented):
Thank you so much for the great work you did on this @gitGurugu ! It's so good to see this wrapping up. I had some comments too please find them below, but I am overall really satisfied with the work you have put in this.

docs/design/namespace-queue.md
devzizu: Yup, we should definitely add this, probably in a followup PR.

staging/src/volcano.sh/apis/pkg/apis/scheduling/v1beta1/types.go
```diff
@@ -369,6 +369,131 @@ type Queue struct {
...
+	// +optional
+	Spec NamespaceQueueSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status reports the observed NamespaceQueue lifecycle and workload state.
+	// +optional
+	Status NamespaceQueueStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// NamespaceQueueSpec represents the desired behavior of NamespaceQueue.
+type NamespaceQueueSpec struct {
```
devzizu, line 399: As per a [previous comment](https://github.com/volcano-sh/volcano/pull/5590#discussion_r3852657175), I thought that we'll be adding a `desiredState` field to the NS Queue spec, but It looks like it was dropped. As such, It seems that we have no way to close/drain the queue in this current version. I see two options here:

1. Reuse the `Command` path. Add a Command informer to the NamespaceQueue controller, extend vcctl queue (or add vcctl namespacequeue), which should be consistent with cluster Queue.
2. Add `spec.desiredState`. Which is simpler, declarative, GitOps-friendly, but diverges from how cluster Queue works.

I would prefer option 2 as it looks more appropriate, simple, gitops friendly, etc (in contrast, Command has other issues, like it seems that it requires default namespace right now). What do you guys prefer here? @hajnalmt @JesseStutler @gitGurugu

pkg/scheduler/api/queue_info.go
```diff
@@ -59,21 +73,92 @@ func NewQueueInfo(queue *scheduling.Queue) *QueueInfo {
...
+			Capability:  namespaceQueue.Spec.Capability.DeepCopy(),
+			Reclaimable: cloneBool(namespaceQueue.Spec.Reclaimable),
+			Guarantee: scheduling.Guarantee{
+				Resource: namespaceQueue.Spec.Guarantee.Resource.DeepCopy(),
+			},
+			Deserved:        namespaceQueue.Spec.Deserved.DeepCopy(),
+			Priority:        namespaceQueue.Spec.Priority,
+			DequeueStrategy: namespaceQueue.Spec.DequeueStrategy,
+		},
+		Status: scheduling.QueueStatus{
```
devzizu, line 118: Every failure the controller detects (`NamespaceNotAllowed`, `ParentNotReady`, etc.) sets `Ready=False` but leaves State at `Open`. So if an admin revokes a namespace from a Queue's `allowedNamespaces`, the webhook blocks new PodGroups but jobs already in that `NamespaceQueue` keep scheduling indefinitely.

My suggestion is to consider adding:

```go
state := namespaceQueue.Status.State
if !commonutil.IsNamespaceQueueSchedulable(
      namespaceQueue.Generation, string(state), namespaceQueue.Status.Conditions) {
      state = scheduling.QueueStateClosed
}
```

This could be added before you instantiate the queue "`queue := &scheduling.Queue{`", and `namespaceQueue.Status.State,` below could become `state, `.

pkg/webhooks/admission/namespacequeues/validate/validate_namespacequeue.go
```diff
@@ -0,0 +1,375 @@
...
+		old, err := schema.DecodeNamespaceQueue(ar.Request.OldObject, ar.Request.Resource)
+		if err != nil {
+			return util.ToAdmissionResponse(err)
+		}
+		if equality.Semantic.DeepEqual(old.Spec, currentNamespaceQueue.Spec) {
+			return &admissionv1.AdmissionResponse{Allowed: true}
+		}
+	}
+
+	if !utilfeature.DefaultFeatureGate.Enabled(features.NamespaceQueue) {
```
devzizu, line 101: This gate check is above the operation switch, so it rejects DELETE too, which makes disabling the feature gate unrecoverable. Deleting an object that already exists shouldn't require the feature to be enabled, so I suggest moving the gate check inside the switch so it only guards CREATE/UPDATE.

staging/src/volcano.sh/apis/pkg/apis/scheduling/v1beta1/types.go
```diff
@@ -369,6 +369,131 @@ type Queue struct {
...
+	// Parent can be a cluster Queue or another NamespaceQueue in the same namespace.
+	// "cluster/<name>" refers to a cluster-scoped Queue.
+	// An unprefixed value refers to a NamespaceQueue in the same namespace.
+	// +optional
+	// +kubebuilder:default:="cluster/default"
+	// +kubebuilder:validation:MaxLength=261
+	// +kubebuilder:validation:Pattern=`^(cluster/)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Parent string `json:"parent,omitempty" protobuf:"bytes,1,opt,name=parent"`
```
devzizu, line 407: parent defaults to `cluster/default`, but nothing ever populates `AllowedNamespaces` on the `default` Queue. So on a fresh install the first minimal `NamespaceQueue` is rejected.

How should we teach users to update the default queue? Scheduler flaag? Helm-bundled default queue? or ship a helm post-hook that patches it?

pkg/webhooks/util/namespace_queue.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	commonutil "volcano.sh/volcano/pkg/util"
+)
+
+// IsNamespaceQueueClosedAndDrained reports whether a NamespaceQueue can be
+// safely detached from its parent or deleted.
+func IsNamespaceQueueClosedAndDrained(namespaceQueue *schedulingv1beta1.NamespaceQueue) bool {
+	return commonutil.IsNamespaceQueueClosedAndDrained(namespaceQueue)
+}
```
devzizu, line 28: This whole file is a function with only one caller. Maybe drop it entirely?

pkg/webhooks/admission/queues/validate/validate_queue.go
```diff
@@ -129,6 +140,49 @@ func AdmitQueues(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse
...
+
+	sort.Strings(blocked)
+	return fmt.Errorf(
+		"cannot revoke NamespaceQueue authorization from Queue %q; update or close and drain these NamespaceQueues first: %s",
+		newQueue.Name,
+		strings.Join(blocked, ", "),
+	)
+}
+
+func isNamespaceAllowedByQueue(queue *schedulingv1beta1.Queue, namespace string) bool {
```
devzizu, line 176: This is very similar to commonutil.IsNamespaceAllowedByQueue, can we simplify using only one?

pkg/util/queue_resources.go
```diff
@@ -0,0 +1,79 @@
...
+		if found && deservedQuantity.Cmp(capabilityQuantity) > 0 {
+			return fmt.Errorf("deserved[%s] must not exceed capability[%s]", resourceName, resourceName)
+		}
+	}
+
+	return nil
+}
+
+// AddResourceList adds source quantities to destination in place.
+func AddResourceList(destination, source corev1.ResourceList) {
```
devzizu, line 61: Can you verify if we don't have already a function for this? (e.g., resourcequota.go, job_controller_util.go, etc.)

pkg/webhooks/admission/namespacequeues/validate/validate_namespacequeue.go
```diff
@@ -0,0 +1,375 @@
...
+var config = &router.AdmissionServiceConfig{}
+
+// AdmitNamespaceQueues validates NamespaceQueue create, update, and delete requests.
+func AdmitNamespaceQueues(ar admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	if ar.Request == nil {
+		if !utilfeature.DefaultFeatureGate.Enabled(features.NamespaceQueue) {
+			return util.ToAdmissionResponse(fmt.Errorf("NamespaceQueue feature is disabled"))
+		}
+		return util.ToAdmissionResponse(fmt.Errorf("admission request must not be nil"))
+	}
```
devzizu, line 75: Both paths are the same inside this nil branch, maybe we could simplify and just return since request is nil

pkg/controllers/namespacequeue/namespacequeue_controller_action.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+
+	return c.removeNamespaceQueueFinalizer(currentQueue)
+}
+
+func hasFinalizer(namespaceQueue *schedulingv1beta1.NamespaceQueue, finalizer string) bool {
+	if namespaceQueue == nil {
+		return false
+	}
+	return containsFinalizer(namespaceQueue.Finalizers, finalizer)
```
devzizu, line 71: Having this helper is not needed, we could just use  `slices.Contains(nq.Finalizers, namespaceQueueFinalizer)`. Wdyt?

So we can collapse hasFinalizer and containsFinalizer into the above suggestion?

devzizu (commented):
Thanks a lot for all the work on this @gitGurugu, it's a big feature and the effort really shows. I've left a few comments below, hope they're helpful!

PR #5592, opened 2026-07-06
perf: replace O(n) NodeList scans with O(1) index map in AddOrUpdateNode/RemoveNode

### LINKED ISSUE #5591: [PERF] Replace O(n) NodeList membership scans with O(1) lookups in scheduler cache

pkg/scheduler/cache/event_handlers.go currently performs O(n) linear scans over NodeList in both AddOrUpdateNode() and RemoveNode() while holding the global scheduler cache mutex (sc.Mutex.Lock()).

In AddOrUpdateNode(), the code iterates through NodeList to determine whether a node already exists before appending it. Similarly, RemoveNode() performs a linear scan to locate the node before removing it from the slice.

Since the scheduler cache already maintains a Nodes map keyed by node name, these membership checks could be performed in O(1). As cluster size increases (e.g., thousands of nodes), repeatedly scanning NodeList under the global mutex increases lock hold time and can contribute to higher scheduler latency during frequent node add/remove events.

Replace the O(n) membership check in AddOrUpdateNode() with an O(1) lookup using the existing Nodes map before appending to NodeList.

For RemoveNode(), maintain an auxiliary index (for example, nodeListIndex map[string]int) to enable O(1) lookup and removal from NodeList. The index can be updated whenever nodes are added or removed.

This change would:

Eliminate unnecessary linear scans while holding the global mutex.
Reduce lock contention during node add/remove operations.
Improve scalability for large clusters with frequent node events.
Preserve the existing scheduler behavior while improving performance.

### Additional context

The current implementation performs O(n) work under the scheduler cache mutex for every node add/remove operation. During large-scale cluster expansion or environments with frequent node churn, this can increase mutex contention with other scheduler operations such as snapshot creation and pod event handling.

Since the cache already maintains a Nodes map, reusing it (or maintaining a lightweight index for NodeList) provides a more scalable implementation with minimal behavioral impact.

I would be happy to work on implementing this optimization and submit a pull request if the proposal is accepted.

vanshika2720:
#### What type of PR is this?

/kind feature

This PR improves the performance of node add/remove operations in the scheduler cache by eliminating O(n) `NodeList` scans while holding the global scheduler cache mutex.

Previously:
- `AddOrUpdateNode()` performed a linear scan over `NodeList` to determine whether a node already existed before appending it.
- `RemoveNode()` performed another linear scan to locate the node and removed it using append-splice, resulting in O(n) lookup and O(n) slice copying.

This PR introduces a `nodeListIndex` (`map[string]int`) that is maintained alongside `NodeList` to provide efficient lookups.

The implementation includes:
- Using `nodeListIndex` for O(1) membership checks in `AddOrUpdateNode()`.
- Using `nodeListIndex` together with swap-delete in `RemoveNode()` to eliminate both the linear scan and slice shifting.
- Lazy initialization of `nodeListIndex` from an existing `NodeList` to maintain compatibility with tests that construct `SchedulerCache` directly.
- Keeping `nodeListIndex` synchronized whenever nodes are added or removed.

Since `NodeList` ordering is not relied upon by the scheduler, swap-delete preserves the existing behavior while improving performance and reducing mutex hold time.

Fixes #5591 

- This change is an internal performance optimization and does not modify scheduler behavior or public APIs.
- Existing tests were updated to validate `nodeListIndex` consistency.
- Added test coverage for duplicate add, remove, and re-add scenarios.
- Added benchmarks covering node add/remove operations at 100, 500, and 2000 node scale.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -149,6 +149,7 @@ type SchedulerCache struct {
 	NodeShards           map[string]*schedulingapi.NodeShardInfo
 	PriorityClasses      map[string]*schedulingv1.PriorityClass
 	NodeList             []string
+	nodeListIndex        map[string]int
 	defaultPriorityClass *schedulingv1.PriorityClass
 	defaultPriority      int32
 	CSINodesStatus       map[string]*schedulingapi.CSINodeStatusInfo
@@ -568,6 +569,7 @@ func newSchedulerCache(config *rest.Config, schedulerNames []string, defaultQueu
 		NodeShards:          make(map[string]*schedulingapi.NodeShardInfo),
 
 		NodeList:            []string{},
+		nodeListIndex:       make(map[string]int),
 		nodeWorkers:         nodeWorkers,
 		resourceSyncTimeout: resourceSyncTimeout,
 	}
--- a/pkg/scheduler/cache/cache_mock.go
+++ b/pkg/scheduler/cache/cache_mock.go
@@ -142,6 +142,7 @@ func newMockSchedulerCache(schedulerName string) *SchedulerCache {
 		NodeShards:             make(map[string]*schedulingapi.NodeShardInfo),
 
 		NodeList:       []string{},
+		nodeListIndex:  make(map[string]int),
 		binderRegistry: NewBinderRegistry(),
 		resyncPeriod:   0,
 	}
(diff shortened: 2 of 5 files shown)


vanshika2720:
@JesseStutler
@hajnalmt
@hwdef

PR #5596, opened 2026-07-07
cache: recover from PVC-informer race in addPod

gucardona:
## What this PR does

Fixes a scheduler regression introduced in v1.15.0 that permanently strands pods referencing an OnDemand PVC whose ADD event has not yet reached the scheduler's PVC informer at the moment the pod's ADD reaches the pod informer.

**Symptom.** Any pod created together with its OnDemand PVC in quick succession — for example a Spark executor pod created by the Spark k8s scheduler-backend with a `claimName=OnDemand` executor volume — triggers the race under even light load (16 concurrent workloads is sufficient). Driver: `Running`. Executor: `Pending` forever. PVC: `Pending` with `WaitForPodScheduled`. No `FailedScheduling` event, because the executor never enters the scheduler's consideration set.

The full diagnosis, failure chain, and version-scope bisect are in the commit message.

## Regression origin

#5058 (`capacity plugin support DRA`, commit `5ba79f316`).

Only v1.15.0 is affected. Every v1.14.x tag through v1.14.3 (released after v1.15.0) is clean; v1.15.0-alpha.0 is also clean. Verified with `git merge-base --is-ancestor 5ba79f316 <tag>`.

## Reproduction

Any pod that references a PVC that is created immediately after the pod on the apiserver. Spark 4.0 with the following executor conf reproduces it consistently at ≥16 concurrent workloads:

```
spark.kubernetes.executor.volumes.persistentVolumeClaim.spark-local-dir-1.options.claimName=OnDemand
```

Observed on a live scheduler at `-v=5`: pod ADD at time `T`, PVC-not-found log line at `T + 41µs`. The PVC lands on the apiserver in the same second (pod `rv+3` typical), but the PVC informer has not synced yet.

## Fix

Mirror the existing DRA-not-ready branch in `addPod` for PVC-not-found errors — add the incomplete task to `sc.Jobs`, enqueue for resync, and return `nil`. The retry loop then finds the task by UID and calls `syncTask`, which does a fresh apiserver `Get` for the pod. By that time the PVC informer has caught up and `NewTaskInfo` succeeds.

`isPVCNotFoundError` matches on the stable `"looking up PVC "` prefix emitted by `getPodCSIVolumes`. The wrapped error uses `fmt.Errorf` with `%v`, so `errors.Is` against a sentinel is not viable without also changing the wrapping.

## Test

- `TestAddPodWithUnresolvedPVCCachesTaskForResync` mirrors the existing DRA counterpart `TestAddPodWithUnresolvedResourceClaimTemplateCachesTaskForResync`. Seeds a pod that references a PVC not yet in the informer lister, calls `addPod`, and asserts the task landed in `sc.Jobs` and one entry sits in `errTasks`.
- `TestIsPVCNotFoundError` covers the helper's positive and negative cases.
- Extends `initMockInformers` to wire `sc.pvcInformer` (previously unset, since no existing test needed it).

Full package test suite passes:

```
$ go test ./pkg/scheduler/cache/...
ok  	volcano.sh/volcano/pkg/scheduler/cache	3.671s
```

## Runtime evidence from the internal environment where I first hit this

At `-v=5` on a live cluster with 16 concurrent Spark workloads using OnDemand PVCs:

- Every affected pod produced exactly one `event_handlers.go:275` failure — never more.
- Zero `Retry to resync task` V(5) lines.
- `cache.go:1274 "the length of errTasks is 0"` fires microseconds after each failure — the retry queue is drained and left empty by the same `Forget()` that just processed it.
- Pod `resourceVersion` never advances after creation.
- With this patch applied, `PVC for pod ... not yet in informer cache, add task and retry` fires once per affected pod, followed by `Successfully synced task <...>` ~100ms later. All 1000 test workloads reached terminal state.

## Related Slack discussion

Diagnosed and pre-reviewed by @JesseStutler on volcano-sh Slack: https://cloud-native.slack.com/archives/C011GJDQS0N/p1783368819159279


kunalworldwide (commented):
This is a subtle but important race fix. Treating a missing PVC during `addPod` as a reason to requeue the pod rather than failing scheduling permanently prevents OnDemand PVC pods from being stranded when the PVC informer lags behind the pod informer.

The new regression test is valuable — it reproduces the exact ordering where the pod ADD arrives before the PVC ADD. One suggestion: the test could also assert that the pod is eventually scheduled once the PVC ADD event is delivered, to verify the recovery path end-to-end.

LGTM otherwise.

gucardona:
> This is a subtle but important race fix. Treating a missing PVC during `addPod` as a reason to requeue the pod rather than failing scheduling permanently prevents OnDemand PVC pods from being stranded when the PVC informer lags behind the pod informer.
> 
> The new regression test is valuable — it reproduces the exact ordering where the pod ADD arrives before the PVC ADD. One suggestion: the test could also assert that the pod is eventually scheduled once the PVC ADD event is delivered, to verify the recovery path end-to-end.
> 
> LGTM otherwise.

Thanks for the review! Extended the test in a20a9b201 to seed the PVC in the fake client after the initial addPod, wait for the informer to observe it, drive one processResyncTask() iteration, and assert errTasks drains. That verifies the whole recovery arc, not just that the task was cached for resync.

pkg/scheduler/cache/event_handlers_pvc_race_test.go
JesseStutler, line 1: We don't neet to create a specific test file for pvc race, just put inside `event_handlers_test.go` file is enough

pkg/scheduler/cache/event_handlers.go
```diff
@@ -272,6 +277,16 @@ func (sc *SchedulerCache) addPod(pod *v1.Pod) error {
 			sc.resyncTask(pi)
 			return nil
 		}
+		// Recover from a Pod ADD that races ahead of its PVC ADD on
+		// the informers. Same pattern as the DRA branch above.
+		if errors.IsNotFound(err) {
```
JesseStutler: Should PVCs define custom errors—similar to DRA—instead of using a generic "not found" error? If we only simply use generic `isNotFound` error here, it is inevitable that pods will acquire more associated resources in the future and will jump into here when the pod needs to use other associated resources

pkg/scheduler/cache/event_handlers_pvc_race_test.go
```diff
@@ -0,0 +1,121 @@
...
+		return err == nil, nil
+	})
+	assert.NoError(t, err, "PVC never appeared on the informer lister")
+
+	// Drive one iteration of the resync worker. This pops the taskKey,
+	// calls syncTask (fresh kubeClient Get → NewTaskInfo → PVC lookup
+	// now succeeds), and Forgets the taskKey on success.
+	sc.processResyncTask()
+
+	assert.Equal(t, 0, sc.errTasks.Len(), "task must be removed from errTasks once the PVC has synced and syncTask succeeds")
```
JesseStutler: I don't think this unit test is correct, even if the PVC doesn't arrive, the resync process will retry. We should verify that the Pod has successfully become a `taskInfo` object and been added to the cache, and that the `taskInfo` contains all the necessary information.

pkg/scheduler/cache/event_handlers_pvc_race_test.go
gucardona: Done, moved it into event_handlers_test.go and removed the separate file.

pkg/scheduler/cache/event_handlers.go
gucardona: Good point. I added a pendingPVCError type mirroring pendingDRAResourceClaimError and wrap the PVC lookup miss with it in getPodCSIVolumes, then check isPendingPVCError in addPod instead of the generic isNotFound. That keeps the retry path scoped to the PVC case so future resource lookups in NewTaskInfo that return NotFound for other reasons will not fall into it.

pkg/scheduler/cache/event_handlers_pvc_race_test.go
gucardona: Agreed, reworked the test. It now asserts that addPod succeeds, the task is added to the cache as a proper TaskInfo carrying its pod reference (name/namespace), and that it is enqueued for resync — instead of driving the resync loop and checking the queue drains. Let me know if you want it to assert additional TaskInfo fields.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -127,10 +127,16 @@ func (sc *SchedulerCache) getPodCSIVolumes(pod *v1.Pod) (map[v1.ResourceName]int
...
-			// run without it. Bail out immediately.
-			return volumes, fmt.Errorf("looking up PVC %s/%s: %v", pod.Namespace, pvcName, err)
+			// The PVC is required to proceed with scheduling of a new
+			// pod because it cannot run without it. A PVC ADD event can
+			// reach the scheduler's PVC informer after the pod's ADD
+			// reaches the pod informer (e.g. a pod created together with
+			// an OnDemand PVC). Wrap the error in pendingPVCError so
+			// addPod can add the task to the cache and re-queue it for
+			// resync, rather than dropping the pod permanently.
+			return volumes, &pendingPVCError{
```
JesseStutler: The get pvc may cause other errors, should we also always addTask back instead?

JesseStutler:
Also please resolve the conflicts @gucardona

pkg/scheduler/cache/event_handlers.go
gucardona: Good point. addPod now always adds the task back and resyncs on any NewTaskInfo error, not just the PVC case, so a transient PVC Get error (or any other error from NewTaskInfo) no longer strands the pod either. Dropped the PVC-specific error type since it is no longer needed to distinguish the case.

pkg/scheduler/cache/event_handlers.go
```diff
@@ -271,9 +275,19 @@ func (sc *SchedulerCache) addPod(pod *v1.Pod) error {
...
 		}
-		klog.Errorf("generate taskInfo for pod(%s) failed: %v", pod.Name, err)
+		// NewTaskInfo can fail transiently, e.g. a referenced PVC has not
+		// yet reached the scheduler's informer when the pod ADD is
+		// processed. Always add the task to the cache and enqueue it for
+		// resync so syncTask re-fetches the pod and rebuilds the task once
+		// the dependency syncs, rather than dropping the pod: resyncTask on
+		// a task that was never added to sc.Jobs would be Forgotten by
+		// processResyncTask and lost.
+		klog.V(4).Infof("generate taskInfo for pod <%s/%s> failed, add task to cache and retry: %v", pod.Namespace, pod.Name, err)
```
JesseStutler: That doesn't seem quite right. I mean, shouldn't we only add the task back if we encounter a "PVC not found" error, rather than adding it back for *any* error?

gucardona: You are right, reverted that. getPodCSIVolumes now only wraps the error in pendingPVCError when the PVC lookup is not-found (errors.IsNotFound), and addPod adds the task back and resyncs only for that case. Any other error from NewTaskInfo falls through to the original drop path.

JesseStutler:
/approve
/lgtm
Thanks

JesseStutler:
@gucardona Could you help manually cherry-pick to release-1.15 since there are conflicts?

PR #5597, opened 2026-07-07
webhooks: validate podgroup jdb annotations

avinxshKD:
#### What type of PR is this?

/kind bug

PodGroup admission now validates JDB annotations the same way pod admission does.

This rejects conflicting volcano.sh/jdb-min-available/ volcano.sh/jdb-max-unavailable annotations and invalid JDB values before the scheduler sees them.

Fixes #

NONE

```release-note
PodGroup admission now rejects invalid JDB annotations.

### DIFF
--- a/pkg/webhooks/admission/podgroups/validate/validate_podgroup.go
+++ b/pkg/webhooks/admission/podgroups/validate/validate_podgroup.go
@@ -18,11 +18,13 @@ package validate
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	admissionv1 "k8s.io/api/admission/v1"
 	whv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/klog/v2"
 
 	schedulingv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by avinxshKD** [comment]
Found this while checking PodGroup admission against pod admission.

Pods already reject conflicting/invalid JDB annotations, but PodGroups were accepting them even though the scheduler reads JDB budget from PodGroup annotations too. Kept the fix local to PodGroup validation and added small regression cases.

PTAL @hzxuzhonghu @huone1 @merryzhou

**Comment by Aman-Cool** [comment]
@avinxshKD, thanks for this pr, can you check the failing CI check?

**Comment by avinxshKD** [comment]
failure looks unrelated to this
This patch only touches PodGroup admission validation, so it should not affect CronJob cleanup/history behavior. I’ll rerun the failed check.

**Comment by avinxshKD** [comment]
ptal @hajnalmt @JesseStutler when get chance

File: pkg/webhooks/admission/podgroups/validate/validate_podgroup.go [hunk 1]
```diff
@@ -94,10 +96,58 @@ func validatePodGroup(pg *schedulingv1beta1.PodGroup) string {
 
 	errMsg += checkQueueState(pg.Spec.Queue)
 	errMsg += validateNetworkTopology(pg.Spec.NetworkTopology, pg.Spec.SubGroupPolicy)
+	if err := validateJDBAnnotations(pg.Annotations); err != nil {
```
hzxuzhonghu, line 103: This validation only runs from the CREATE branch, and the webhook registration only includes CREATE. PodGroup annotations are mutable and `SetPodGroup` rebuilds `JobInfo.Budget` on updates, so a valid PodGroup can later be patched with `jdb-min-available: "0"` or both annotations and the scheduler will consume it. Please register UPDATE and validate the new object there too.

hzxuzhonghu (changes requested):
The new validation is bypassable on updates.

pkg/webhooks/admission/podgroups/validate/validate_podgroup.go
avinxshKD: done, updated the webhook to include UPDATE and run the same PodGroup validation on the new object for create/update. Added a regression test for updating a PodGroup with an invalid JDB annotation

avinxshKD:
failure look unrelated to this

PR #5610, opened 2026-07-08
fix(webhook): separate CronJob validation errors with "; "

### LINKED ISSUE #5489: CronJob validation webhook concatenates multiple errors without a separator

aeron-gh:
/kind bug

When a CronJob fails more than one check at once (say a bad schedule and a name over the length limit), the webhook joined the messages with no separator, so the rejection came out as one run-on line. This trims each validator message and joins the non-empty ones with "; " so each error reads on its own.

It's the CronJob version of the PodGroup fix in #5487 (same collect-and-join, with the leading/trailing whitespace trimmed for a clean result). Going by @hajnalmt's grep on that PR, this was the last spot left with the pattern. It's one `validateCronJob` helper shared by the create and update paths, plus a couple of tests.

Scoped to the CronJob webhook. The `admit_job.go` cleanup the issue mentions uses a different convention (its own leading-space / trailing-`;` style), so it's better kept as a separate change.

Thanks @anxkhn for starting this off in #5533. I opened this since it's issue (#5489) and I already had the same pattern from #5487.

Fixes #5489

```release-note
NONE
```

### DIFF
--- a/pkg/webhooks/admission/cronjobs/validate/admit_cronjob_test.go
+++ b/pkg/webhooks/admission/cronjobs/validate/admit_cronjob_test.go
@@ -412,3 +412,54 @@ func TestValidateCronJobName(t *testing.T) {
 		})
 	}
 }
+
+// When several checks fail at once, the messages should be separated by "; ".
+func TestValidateCronJobSeparatesMultipleErrors(t *testing.T) {
+	cronjob := &v1alpha1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      strings.Repeat("a", 53),
+			Namespace: metav1.NamespaceDefault,
+		},
+		Spec: v1alpha1.CronJobSpec{
+			Schedule:          "error schedule",
+			ConcurrencyPolicy: v1alpha1.AllowConcurrent,
+			JobTemplate: v1alpha1.JobTemplateSpec{
+				Spec: getJobTemplate(),
+			},
+		},
+	}
+
+	msg := validateCronJob(cronjob)
+
+	for _, want := range []string{"schedule is not a valid cron expression", "; ", "must be no more than 52 characters"} {
+		if !strings.Contains(msg, want) {
+			t.Errorf("expected %q in combined message, got %q", want, msg)
+		}
+	}
+}
+
+// A single failing check should come back on its own, with no separator.
+func TestValidateCronJobSingleErrorHasNoSeparator(t *testing.T) {
+	cronjob := &v1alpha1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "valid-cronjob",
+			Namespace: metav1.NamespaceDefault,
+		},
+		Spec: v1alpha1.CronJobSpec{
+			Schedule:          "",
+			ConcurrencyPolicy: v1alpha1.AllowConcurrent,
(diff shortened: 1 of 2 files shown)


aeron-gh:
e2e-shard / E2E Agent Scheduler Soft looks like an unrelated flake The rest is green. Re-running to confirm.

pkg/webhooks/admission/cronjobs/validate/admit_cronjob_test.go
```diff
@@ -412,3 +412,54 @@ func TestValidateCronJobName(t *testing.T) {
 		})
 	}
 }
+
+// When several checks fail at once, the messages should be separated by "; ".
+func TestValidateCronJobSeparatesMultipleErrors(t *testing.T) {
+	cronjob := &v1alpha1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      strings.Repeat("a", 53),
```
hajnalmt, line 420: I really like these funny touches in our tests 😄

hajnalmt (approved):
/area webhooks
/approve
/lgtm

pkg/webhooks/admission/cronjobs/validate/admit_cronjob_test.go
aeron-gh: thanks @hajnalmt! Glad the tests landed well 😄 Appreciate the quick review.

PR #5614, opened 2026-07-08
feat(scheduler): use SSA for queue-status allocated field

gucardona:
#### What type of PR is this?

/kind bug

Switches the scheduler's `Queue` status writer from a full-object `UpdateStatus` to a Server-Side Apply that owns only `.status.allocated` under the FieldManager `"vc-scheduler"`.

The queue-controller has already been using SSA under FieldManager `"queue-controller"` for `.status.state` (see [`pkg/controllers/queue/queue_controller_action.go`](https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/queue/queue_controller_action.go)). On the scheduler side, `UpdateStatus` sends the full `Queue` object.

The scheduler only writes when `.status.allocated` actually changes (session.go guards the call with a DeepEqual). The problem is the write itself: the full-object `UpdateStatus` carries the snapshot's `resourceVersion`, so when the queue-controller has written `.status.state` since that snapshot, the scheduler's update comes back 409 Conflict and retries. On busy queues that is steady 409 + retry noise, not a silent overwrite of `.state`.

After this change the two writers touch disjoint field sets under distinct managers, so there is no shared field to conflict on:

| Field manager | Owns |
|---|---|
| `"vc-scheduler"` | `.status.allocated` |
| `"queue-controller"` | `.status.state` |

`Force: true` is set on the scheduler-side apply so a v1.15 upgrade from a prior scheduler version that still owned `.status.allocated` under the legacy `UpdateStatus` manager cleanly transfers ownership.

No API changes, no new flags. Existing tests remain green (`go test ./pkg/scheduler/... ./pkg/controllers/queue/...`).

Fixes #

- RBAC: the scheduler ClusterRole now grants `patch` on `queues/status` (kept `update` for upgrade skew). Server-Side Apply is sent as a PATCH, so without it `ApplyStatus` 403s in a real cluster. Installer manifests regenerated with `make generate-yaml`.
- The controller-side `ApplyStatus` calls in `queue_controller_action.go` are **intentionally left alone**. They already work, and since the scheduler now owns a disjoint field, there is no conflict for `Force: true` to resolve on the controller side.
- The scheduler change is confined to one function in `pkg/scheduler/cache/cache.go` (`defaultStatusUpdater.UpdateQueueStatus`).
- Symptom this fixes in the wild: on high-throughput deployments the scheduler's status write raced the controller's `.state` write and came back 409, retrying on the next cycle. After the change the two managers own disjoint fields, so those conflicts go away.

```release-note
The scheduler now uses Server-Side Apply to write `Queue.status.allocated` under the FieldManager `"vc-scheduler"`, taking ownership of only that field instead of writing the whole object. This removes the 409 conflict-and-retry churn between the scheduler and the queue-controller (which owns `Queue.status.state`) on busy queues. The scheduler ClusterRole now also grants `patch` on `queues/status`, required because Server-Side Apply is sent as a PATCH.
```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -62,6 +62,7 @@ import (
 	"volcano.sh/apis/pkg/apis/scheduling"
 	schedulingscheme "volcano.sh/apis/pkg/apis/scheduling/scheme"
 	vcv1beta1 "volcano.sh/apis/pkg/apis/scheduling/v1beta1"
+	v1beta1apply "volcano.sh/apis/pkg/client/applyconfiguration/scheduling/v1beta1"
 	vcclient "volcano.sh/apis/pkg/client/clientset/versioned"
 	"volcano.sh/apis/pkg/client/clientset/versioned/scheme"
 	vcinformer "volcano.sh/apis/pkg/client/informers/externalversions"
@@ -75,6 +76,7 @@ import (
 	schedulingapi "volcano.sh/volcano/pkg/scheduler/api"
 	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/metrics/source"
+	schedulerutil "volcano.sh/volcano/pkg/scheduler/util"
 	schedulercache "volcano.sh/volcano/pkg/schedulercommon/cache"
 	"volcano.sh/volcano/pkg/util"
 )
@@ -383,8 +385,15 @@ func (su *defaultStatusUpdater) UpdateQueueStatus(queue *schedulingapi.QueueInfo
 		return err
 	}
 
-	_, err := su.vcclient.SchedulingV1beta1().Queues().UpdateStatus(context.TODO(), newQueue, metav1.UpdateOptions{})
-	if err != nil {
+	// Use Server-Side Apply and own only .status.allocated with the
+	// scheduler's FieldManager. The queue-controller owns .status.state
+	// via its own FieldManager (see pkg/controllers/queue/…).
+	// UpdateStatus with a full-object body used to stomp the state field
+	// on every apply, forcing the queue controller to re-write it and
+	// producing a flap between the two writers.
+	queueStatusApply := v1beta1apply.QueueStatus().WithAllocated(newQueue.Status.Allocated)
+	queueApply := v1beta1apply.Queue(newQueue.Name).WithStatus(queueStatusApply)
+	if _, err := su.vcclient.SchedulingV1beta1().Queues().ApplyStatus(context.TODO(), queueApply, metav1.ApplyOptions{FieldManager: schedulerutil.DefaultComponentName, Force: true}); err != nil {
 		klog.Errorf("error occurred in updating Queue <%s>: %s", newQueue.Name, err.Error())
 		return err
 	}
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -379,22 +379,20 @@ func (su *defaultStatusUpdater) UpdatePodGroup(pg *schedulingapi.PodGroup) (*sch
 
 // UpdateQueueStatus will update the status of queue
 func (su *defaultStatusUpdater) UpdateQueueStatus(queue *schedulingapi.QueueInfo) error {
-	newQueue := &vcv1beta1.Queue{}
-	if err := schedulingscheme.Scheme.Convert(queue.Queue, newQueue, nil); err != nil {
-		klog.Errorf("error occurred in converting scheduling.Queue to v1beta1.Queue: %s", err.Error())
-		return err
-	}
-
 	// Use Server-Side Apply and own only .status.allocated with the
 	// scheduler's FieldManager. The queue-controller owns .status.state
 	// via its own FieldManager (see pkg/controllers/queue/…).
 	// UpdateStatus with a full-object body used to stomp the state field
 	// on every apply, forcing the queue controller to re-write it and
 	// producing a flap between the two writers.
-	queueStatusApply := v1beta1apply.QueueStatus().WithAllocated(newQueue.Status.Allocated)
-	queueApply := v1beta1apply.Queue(newQueue.Name).WithStatus(queueStatusApply)
+	//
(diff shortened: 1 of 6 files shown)


Aman-Cool (commented):
@gucardona, I really like the direction here.., went and checked the controller side to be sure, and yeah, it only ever applies `.status.state` under its own field manager, so the disjoint-ownership picture in your table holds up.

But I think I found why this pattern worked for the controller and won't for the scheduler. While tracing how the controller's `ApplyStatus` gets away with it, I noticed its ClusterRole has `patch` on `queues/status`... the scheduler's role only has `update` (helm template, scheduler.yaml around line 93). SSA goes out as a PATCH, so every apply from the scheduler should come back 403 in a real cluster.

The scary part is that CI stays completely green through this; `updateQueueStatus` in session.go just logs the error and keeps scheduling from cache, and none of the e2es actually look at `status.allocated`. So allocated would quietly go stale in prod and nobody would notice until some dashboard reads it. Can you add `patch` to the scheduler role (I'd keep `update` too, for upgrade skew where an old scheduler is still running) and re-run `make generate-yaml` so the installer yamls stay in sync? The verify-generated-yaml gate will complain otherwise.

One smaller thing on the description... I tried to reproduce the flapping story in my head and I don't think it happens quite the way it's written. session.go already skips the write when allocated hasn't changed (there's a DeepEqual guard right before the update call), and since the old full-object update carries the snapshot's resourceVersion, a concurrent state write from the controller makes the scheduler's update fail with a 409 rather than silently stomping `.state`. So the real win here is killing those 409s and the retry noise, plus the clean ownership split; which is honestly still a good reason for the change, just worth saying it that way in the release note so we don't document a failure mode we don't have.

pkg/scheduler/cache/cache.go
```diff
@@ -383,8 +385,15 @@ func (su *defaultStatusUpdater) UpdateQueueStatus(queue *schedulingapi.QueueInfo
...
-	if err != nil {
+	// Use Server-Side Apply and own only .status.allocated with the
+	// scheduler's FieldManager. The queue-controller owns .status.state
+	// via its own FieldManager (see pkg/controllers/queue/…).
+	// UpdateStatus with a full-object body used to stomp the state field
+	// on every apply, forcing the queue controller to re-write it and
+	// producing a flap between the two writers.
+	queueStatusApply := v1beta1apply.QueueStatus().WithAllocated(newQueue.Status.Allocated)
+	queueApply := v1beta1apply.Queue(newQueue.Name).WithStatus(queueStatusApply)
+	if _, err := su.vcclient.SchedulingV1beta1().Queues().ApplyStatus(context.TODO(), queueApply, metav1.ApplyOptions{FieldManager: schedulerutil.DefaultComponentName, Force: true}); err != nil {
```
Aman-Cool, line 394: Since this changes what actually goes over the wire for the status write, could we get a small unit test pinning it? Fake clientset, call `UpdateQueueStatus`, then check the recorded action is a single apply-patch containing `allocated` and nothing else under the `vc-scheduler` manager... that way if someone ever reverts this to a full-object write, a test catches it instead of a busy cluster.

pkg/scheduler/cache/cache.go
```diff
@@ -383,8 +385,15 @@ func (su *defaultStatusUpdater) UpdateQueueStatus(queue *schedulingapi.QueueInfo
...
 
-	_, err := su.vcclient.SchedulingV1beta1().Queues().UpdateStatus(context.TODO(), newQueue, metav1.UpdateOptions{})
-	if err != nil {
+	// Use Server-Side Apply and own only .status.allocated with the
+	// scheduler's FieldManager. The queue-controller owns .status.state
+	// via its own FieldManager (see pkg/controllers/queue/…).
+	// UpdateStatus with a full-object body used to stomp the state field
+	// on every apply, forcing the queue controller to re-write it and
+	// producing a flap between the two writers.
+	queueStatusApply := v1beta1apply.QueueStatus().WithAllocated(newQueue.Status.Allocated)
```
Aman-Cool: +1 to the bot here, I double-checked and it's right for once — the internal `QueueStatus.Allocated` is already a `v1.ResourceList`, so you can build the apply straight from `queue.Name` and `queue.Queue.Status.Allocated` and drop the whole Convert dance along with its error branch. Less code, one less way to fail.

pkg/scheduler/cache/cache.go
```diff
@@ -383,8 +385,15 @@ func (su *defaultStatusUpdater) UpdateQueueStatus(queue *schedulingapi.QueueInfo
 		return err
 	}
```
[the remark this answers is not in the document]
gucardona: Good catch, dropped the conversion. The apply is now built straight from queue.Name and queue.Queue.Status.Allocated (which is already a v1.ResourceList), so the Convert call and its error branch are gone.

pkg/scheduler/cache/cache.go
gucardona: Done, removed the Convert dance and its error branch. Building the apply directly from queue.Name and queue.Queue.Status.Allocated as you and the bot suggested.

pkg/scheduler/cache/cache.go
gucardona: Added TestUpdateQueueStatusUsesServerSideApply. It uses a fake clientset, calls UpdateQueueStatus, and asserts a single apply patch on the status subresource under the vc-scheduler field manager whose payload contains allocated and not state. So a revert to a full-object write trips the test.

Aman-Cool (commented):
Please resolve the merge conflicts before requesting a review🙂

pkg/scheduler/cache/cache.go
gucardona: Already added in an earlier push, TestUpdateQueueStatusUsesServerSideApply. Fake clientset, asserts a single apply patch on the status subresource under vc-scheduler with allocated set and no state. A revert to a full-object write trips it.

pkg/scheduler/cache/cache.go
gucardona: Done in the earlier push, the Convert dance and its error branch are gone. Apply is built straight from queue.Name and queue.Queue.Status.Allocated.

PR #5618, opened 2026-07-09
fix(scheduler): add cleanup metrics for job and queue metrics

### LINKED ISSUE #4745: volcano-scheduler memory leak problem

fengruotj:
Author:    tanjie.master <tanjiemaster@gmail.com>
Date:      Thu Jul 9 21:40:24 2026 +0800

#### What type of PR is this?
/kind feature

Queue metrics have the same issue.

Image

``` 
func DeleteQueueMetrics(queueName string) {
	queueAllocatedMilliCPU.DeleteLabelValues(queueName)
	queueAllocatedMemory.DeleteLabelValues(queueName)
	queueRequestMilliCPU.DeleteLabelValues(queueName)
	queueRequestMemory.DeleteLabelValues(queueName)
	queueDeservedMilliCPU.DeleteLabelValues(queueName)
	queueDeservedMemory.DeleteLabelValues(queueName)
	queueShare.DeleteLabelValues(queueName)
	queueWeight.DeleteLabelValues(queueName)
	queueOverused.DeleteLabelValues(queueName)
	queueCapacityMilliCPU.DeleteLabelValues(queueName)
	queueCapacityMemory.DeleteLabelValues(queueName)
	queueRealCapacityMilliCPU.DeleteLabelValues(queueName)
	queueRealCapacityMemory.DeleteLabelValues(queueName)
	queueInqueueMilliCPU.DeleteLabelValues(queueName)
	queueInqueueMemory.DeleteLabelValues(queueName)
	partialLabelMap := map[string]string{"queue_name": queueName}
	queueAllocatedScalarResource.DeletePartialMatch(partialLabelMap)
	queueRequestScalarResource.DeletePartialMatch(partialLabelMap)
	queueDeservedScalarResource.DeletePartialMatch(partialLabelMap)
	queueCapacityScalarResource.DeletePartialMatch(partialLabelMap)
	queueRealCapacityScalarResource.DeletePartialMatch(partialLabelMap)
	queueInqueueScalarResource.DeletePartialMatch(partialLabelMap)
}
```

**Added a delayed cleanup metrics mechanism to the scheduler’s Metrics.**

volcano-scheduler stores Kubernetes objects in a local cache during scheduling. When scheduling begins, it takes a snapshot of this cache and executes scheduling logic based on configured plugins, during which metrics are recorded. If the snapshot and the local cache become inconsistent, metrics can leak. The delayed cleanup metrics feature prevents this.

@hzxuzhonghu  @hajnalmt @JesseStutler 

NONE

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -1586,6 +1586,43 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 	return snapshot
 }
 
+// SnapshotIndex returns a lightweight snapshot containing the object IDs
+// currently available for scheduling.
+func (sc *SchedulerCache) SnapshotIndex() *schedulingapi.ClusterInfo {
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
+
+	snapshot := &schedulingapi.ClusterInfo{
+		Nodes:  make(map[string]*schedulingapi.NodeInfo),
+		Jobs:   make(map[schedulingapi.JobID]*schedulingapi.JobInfo),
+		Queues: make(map[schedulingapi.QueueID]*schedulingapi.QueueInfo),
+	}
+
+	for _, value := range sc.Nodes {
+		if !value.Ready() {
+			continue
+		}
+
+		snapshot.Nodes[value.Name] = nil
+	}
+
+	for _, queue := range sc.Queues {
+		snapshot.Queues[queue.UID] = nil
+	}
+
+	for _, job := range sc.Jobs {
+		if job.PodGroup == nil {
+			continue
+		}
+		if _, found := snapshot.Queues[job.Queue]; !found {
+			continue
+		}
+		snapshot.Jobs[job.UID] = nil
+	}
+
+	return snapshot
+}
+
 func (sc *SchedulerCache) SharedDRAManager() fwk.SharedDRAManager {
 	return sc.sharedDRAManager
 }
--- a/go.mod
+++ b/go.mod
@@ -171,7 +171,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/selinux v1.13.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_model v0.6.2 // indirect
+	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/procfs v0.19.2 // indirect
 	github.com/stoewer/go-strcase v1.3.0 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
--- a/pkg/scheduler/cache/cache_test.go
+++ b/pkg/scheduler/cache/cache_test.go
@@ -685,3 +685,35 @@ func TestWaitForHandlerSync_InitialEventAsyncHandlerTracker_CompletesAfterDone(t
 	assert.Less(t, elapsed, 2*time.Second, "WaitForHandlerSync should return after all pending objects are processed")
 	assert.GreaterOrEqual(t, elapsed, 100*time.Millisecond, "WaitForHandlerSync should wait until Done is called")
 }
+
+func TestSnapshotIndex(t *testing.T) {
+	activeQueue := &api.QueueInfo{UID: "active-queue", Name: "active-queue"}
+	activeJob := api.NewJobInfo("ns/active-job")
+	activeJob.Queue = activeQueue.UID
+	activeJob.PodGroup = &api.PodGroup{}
+	jobWithoutPodGroup := api.NewJobInfo("ns/no-podgroup")
+	jobWithoutPodGroup.Queue = activeQueue.UID
+	jobWithMissingQueue := api.NewJobInfo("ns/missing-queue")
+	jobWithMissingQueue.Queue = "missing-queue"
+	jobWithMissingQueue.PodGroup = &api.PodGroup{}
+
+	sc := &SchedulerCache{
+		Queues: map[api.QueueID]*api.QueueInfo{
+			activeQueue.UID: activeQueue,
+		},
+		Jobs: map[api.JobID]*api.JobInfo{
+			activeJob.UID:           activeJob,
+			jobWithoutPodGroup.UID:  jobWithoutPodGroup,
+			jobWithMissingQueue.UID: jobWithMissingQueue,
+		},
+	}
+
+	index := sc.SnapshotIndex()
+
+	if len(index.Queues) != 1 || index.Queues[activeQueue.UID] != nil {
+		t.Fatalf("expected only active queue ID with nil value, got %#v", index.Queues)
+	}
+	if len(index.Jobs) != 1 || index.Jobs[activeJob.UID] != nil {
+		t.Fatalf("expected only schedulable job ID with nil value, got %#v", index.Jobs)
+	}
+}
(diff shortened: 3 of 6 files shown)


pkg/scheduler/cache/cache.go
```diff
@@ -1823,3 +1849,77 @@ func (sc *SchedulerCache) OnSessionClose() {
...
+		}
+	case QueueMetricsDeletion:
+		if _, found := sc.Queues[schedulingapi.QueueID(key.queueName)]; found {
+			klog.V(4).Infof("Skip deleting metrics for active Queue <%s>", key.queueName)
+			return
+		}
+	}
+
+	sc.deleteMetricsNow(key)
+}
```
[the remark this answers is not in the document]
fengruotj: The previous code had the same locking behavior as well.

pkg/scheduler/cache/cache.go
```diff
@@ -1823,3 +1849,77 @@ func (sc *SchedulerCache) OnSessionClose() {
...
+	})
+}
+
+func (sc *SchedulerCache) enqueueMetricsDeletion(key metricsDeletionKey) {
+	if sc.metricsDeletionQueue == nil {
+		sc.deleteMetricsNow(key)
+		return
+	}
+	sc.metricsDeletionQueue.AddAfter(key, sc.metricsDeletionDelay)
+}
```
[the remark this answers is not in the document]
fengruotj: Already fixed.

pkg/scheduler/cache/cache.go
```diff
@@ -864,6 +881,8 @@ func (sc *SchedulerCache) Run(stopCh <-chan struct{}) {
 
 	// Cleanup jobs.
 	go wait.Until(sc.processCleanupJob, 0, stopCh)
+	// Delay clean metrics
+	go wait.Until(sc.processMetricsDeletion, 0, stopCh)
```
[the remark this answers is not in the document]
fengruotj: The previous code had the same behavior as well.

pkg/scheduler/cache/cache.go
```diff
@@ -1999,3 +2026,72 @@ func (sc *SchedulerCache) buildTaskDRAInfo(pod *v1.Pod) (map[string]*schedulinga
...
+}
+
+func (sc *SchedulerCache) deleteMetricsIfAbsent(key metricsDeletionKey) {
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
+
+	switch key.deletionType {
+	case JobMetricsDeletion:
+		jobID := schedulingapi.JobID(key.jobID)
+		if _, found := sc.Jobs[jobID]; found {
```
kube-gopher: When `namespace-a/foo` is deleted while `namespaceb/foo` remains active, this lookup only checks the deleted job's namespaced ID, so the delayed pass proceeds. `DeleteJobMetrics` then deletes `unscheduleTaskCount` and `jobRetryCount` using only jobName, resetting metrics that the active `namespace-b/foo` accumulated since the initial deletion

kube-gopher (commented):
It is recommended to take into account scenarios where active jobs share these metric labels without namespace qualification

pkg/scheduler/cache/cache.go
fengruotj: Thanks for the review — that’s a good point. Perhaps I should delay the metric deletion directly instead. Because the queues and jobs have already been deleted earlier in the processing flow.

I think JobID is the globally unique identifier for a job. You can find the relevant definitions/usages here:

* pkg/scheduler/api/job_info.go:163
* pkg/scheduler/cache/event_handlers.go:70

pkg/scheduler/cache/cache.go
```diff
@@ -1999,3 +2026,72 @@ func (sc *SchedulerCache) buildTaskDRAInfo(pod *v1.Pod) (map[string]*schedulinga
...
+	metrics.DeleteQueueMetrics(queueName)
+
+	sc.enqueueMetricsDeletion(metricsDeletionKey{
+		deletionType: QueueMetricsDeletion,
+		queueName:    queueName,
+	})
+}
+
+func (sc *SchedulerCache) enqueueMetricsDeletion(key metricsDeletionKey) {
+	sc.metricsDeletionQueue.AddAfter(key, sc.metricsDeletionDelay)
```
hzxuzhonghu: `AddAfter` keeps the earlier deadline when an equal queue key is already waiting. If a queue is deleted, recreated, and deleted again within this delay, the second deletion still fires at the first deadline; a scheduling cycle for the second incarnation can then write metrics after the only cleanup has run. The same one-shot race exists for any scheduling cycle that outlives the delay. Please make cleanup generation-aware or sweep inactive series instead of relying on a single fixed timer.

hzxuzhonghu (changes requested):
The delayed cleanup still has a stale-writer race.

fengruotj:
> The delayed cleanup still has a stale-writer race.

I confirmed the race. `AddAfter` keeps the earliest `readyAt` for an equal key, so re-enqueuing the same queue cannot extend the cleanup deadline.

I removed the `AddAfter`-based one-shot cleanup and replaced it with an end-of-session inactive-series sweep. Queue deletion still removes metrics immediately. After all scheduling actions and every plugin’s `OnSessionClose` hook have completed, `CloseSession` compares the session snapshot with a fresh cache index and removes metrics for jobs and queues that are no longer active.

This ties cleanup to the metrics writer lifecycle instead of a fixed timer. The delete/recreate/delete case is no longer coalesced onto the first deadline, and scheduling cycles of any duration are cleaned up after their final session-level metric write. A recreated queue with the same name remains preserved while active because queue metrics are keyed only by `queue_name`.

I also added regression tests covering both the delete/recreate/delete late-writer case and the active same-name recreation case.

pkg/scheduler/cache/cache.go
```diff
@@ -1586,6 +1586,43 @@ func (sc *SchedulerCache) Snapshot() *schedulingapi.ClusterInfo {
 	return snapshot
 }
 
+// SnapshotIndex returns a lightweight snapshot containing the object IDs
+// currently available for scheduling.
+func (sc *SchedulerCache) SnapshotIndex() *schedulingapi.ClusterInfo {
```
JesseStutler, line 1591: I don't prefer holding the cache lock and doing the whole snapshot again during CloseSession only for cleaning the dirty metrics. Could we do like other ways? Like when the delete event handler is triggered, we could record these inflight deleting jobs and don't clean the jobs' metrics immediately, until the old session finished, then we could clean the metrics in the inflight deleting jobs map. That way, the amount of iteration is reduced. WDYT? Or are there other better approaches? But for now, I don't prefer current way... /cc @hajnalmt @hzxuzhonghu @wangyang0616

fengruotj: I think the main challenge with this approach is that the entries in the inflight deleting jobs map may not appear during `CloseSession`.

When `runOnce` has not been triggered, the deleted jobs or queues may not be included in the snapshot cache created by `OpenSession`. The metrics leak only occurs under this specific timing condition, so relying on `CloseSession` to clean up the inflight entries may still miss some cases.

JesseStutler: > "When runOnce has not been triggered, the deleted jobs or queues may not be included in the snapshot cache created by OpenSession"

Sorry I'm not quite sure what you're saying. Doesn't this mean that metrics wouldn't be recreated? I think you mean that in my proposal, if there's no snapshot, then no one would clean up the inflight pending map, right? Maybe we can add a flag in the cache to indicate whether a session is active? If it's an active session, we enter the inflight deletion pending map. If not, we proceed with deletion directly

PR #5619, opened 2026-07-09
fix: restore orphan cronjob active jobs

avinxshKD:
#### What type of PR is this?

/kind bug

CronJob reconciliation already detects owned running Jobs that are missing from `status.active`, but it only recorded an event.

This restores those Jobs back into the active list during reconciliation. That keeps `Forbid` and `Replace` concurrency policy checks from ignoring an already-running owned Job.

Fixes #

Added regression coverage for `Forbid` and `Replace` when an owned running Job is missing from `status.active`.

```release-note
Fix Volcano CronJob concurrency handling when a running owned Job is missing from status.active.
```


avinxshKD:
@hajnalmt @JesseStutler PTAL, restores missing active refs for running CronJob-owned Jobs so concurrency policies do not make decisions from stale status

PR #5621, opened 2026-07-09
fix: prevent int64 overflow in DRA device count quota accounting

### LINKED ISSUE #5620: int64 overflow in DRA device count quota accounting admits jobs that should be rejected
**What happened:**

The capacity plugin's DRA quota accounting sums user-controlled device counts on raw `int64` with no overflow guard, so a job whose device count (alone multiplied by gang size, or accumulated with the queue's existing `inqueue` total) exceeds `math.MaxInt64` wraps to a negative value and is wrongly admitted.

A ResourceClaim's `spec.devices.requests[].exactly.count` is an `int64` that the apiserver accepts up to `math.MaxInt64` (validation only rejects `count <= 0`). The gang size (`minMember` / TaskMinAvailable) is user-controlled too.

Three sites accumulate this without guarding overflow:

- `JobInfo.GetMinDRAResources` (`pkg/scheduler/api/job_info.go`): `result[deviceClass].Count += request.Count * int64(times)` — both the product and the running sum can overflow.
- `DRAResource.Add` (`job_info.go`): `d.Count += other.Count` — accumulates into a queue's `inqueue`/`allocated` totals.
- `checkDRAAllocatable` (`pkg/scheduler/plugins/capacity/capacity.go`): `allocatedCount + inqueueCount + request.Count > capability.Count` — a raw sum.

When the sum wraps negative, `negative > capability` is false, so `checkDRAAllocatable` returns allocatable and admits a job that should be rejected. The overflowed value is then committed into the queue's `inqueue`/`allocated` DRA accounting, corrupting quota decisions for other jobs in the queue for the rest of the scheduling session. (`DRAResource.Sub` already clamps at zero; `Add` and the comparison do not guard overflow.)

**What you expected to happen:**

An oversized device count should be rejected by the capacity check (it exceeds the configured capability), and the per-queue DRA accounting should never wrap negative.

**How to reproduce it:**

Enable the capacity plugin with a DRA quota (`deviceclass.count/<class>`) and `DynamicResourceAllocation` (GA, on by default). Submit a PodGroup/Job with `minMember: 2` whose pods reference a ResourceClaim with `count: 9223372036854775807`, or submit two jobs each requesting `count: 9223372036854775807` into the same queue. The overflowed count wraps negative and the job is admitted despite exceeding the DeviceClass quota.

**Environment:**

- Volcano version: master (058a1bc)
- Kubernetes: DynamicResourceAllocation feature gate enabled

thc1006:
#### What type of PR is this?

/kind bug

The capacity plugin's DRA quota accounting sums user-controlled device counts on raw `int64` with no overflow guard, so an oversized count (alone multiplied by the gang size, or accumulated with the queue's existing `inqueue` total) wraps negative and is wrongly admitted by `checkDRAAllocatable`, poisoning per-queue DRA accounting for the scheduling session.

A ResourceClaim device count is an `int64` the apiserver accepts up to `math.MaxInt64` (it only rejects `count <= 0`), and the gang size (`minMember`) is user-controlled as well. This adds saturating arithmetic (`SaturatingAdd` / `SaturatingMul`, clamping at `math.MaxInt64`) at the three accumulation/comparison sites:

- `GetMinDRAResources`: `request.Count * int64(times)` and the running sum.
- `DRAResource.Add`: the count accumulation into `inqueue`/`allocated`.
- `checkDRAAllocatable`: the `allocated + inqueue + request` comparison.

An oversized request now stays a large positive value and is correctly rejected by the quota check. `DRAResource.Sub` already clamps at zero; this brings `Add` and the comparison in line.

Fixes #5620

AI tools (Claude Code) were used to help prepare this change, disclosed here per the AI Guidance in contributing.md. As the author I reviewed and understand every line: I traced the overflow from a user-submitted ResourceClaim + PodGroup through `GetMinDRAResources` / `DRAResource.Add` into `checkDRAAllocatable`, and added unit tests for the saturating helpers plus regression tests for `DRAResource.Add` and `checkDRAAllocatable`. The two regression tests fail on the pre-fix code (the overflowed sum admits the job) and pass with the fix.

The change is intentionally minimal and does not touch the classic non-DRA resource paths, which use `resource.Quantity` (big.Int) and are unaffected by this `int64` overflow.

```release-note
Fixed an int64 overflow in the capacity plugin's DRA device-count quota accounting that could admit a job exceeding its DeviceClass quota and corrupt per-queue DRA accounting.
```

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -1414,7 +1414,7 @@ func (d *DRAResource) Add(other *DRAResource) {
 	if other == nil {
 		return
 	}
-	d.Count += other.Count
+	d.Count = SaturatingAdd(d.Count, other.Count)
 	if other.Capacity != nil {
 		if d.Capacity == nil {
 			d.Capacity = make(map[string]resource.Quantity)
@@ -1469,7 +1469,7 @@ func (ji *JobInfo) GetMinDRAResources() map[string]*DRAResource {
 				}
 			}
 
-			result[deviceClass].Count += request.Count * int64(times)
+			result[deviceClass].Count = SaturatingAdd(result[deviceClass].Count, SaturatingMul(request.Count, int64(times)))
 			for dim, cap := range request.Capacity {
 				totalCap := cap.DeepCopy()
 				for i := int32(0); i < times-1; i++ {
--- a/pkg/scheduler/api/saturating.go
+++ b/pkg/scheduler/api/saturating.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import "math"
+
+// SaturatingAdd returns a + b, clamped to [math.MinInt64, math.MaxInt64] on
+// overflow instead of wrapping around.
+//
+// DRA device counts (ResourceClaim ...Exactly.Count) and gang sizes are
+// user-controlled and effectively unbounded (the apiserver only rejects
+// count <= 0), so accumulating them with a plain += can wrap a large positive
+// sum into a negative value. A negative count then defeats the "requested >
+// capability" quota check in checkDRAAllocatable, admitting a job that should
+// be rejected. Saturating keeps the running total a large positive value so
+// the quota check still rejects it.
+func SaturatingAdd(a, b int64) int64 {
+	s := a + b
+	if a > 0 && b > 0 && s < 0 {
+		return math.MaxInt64
+	}
+	if a < 0 && b < 0 && s >= 0 {
+		return math.MinInt64
+	}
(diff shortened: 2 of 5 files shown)


pkg/scheduler/plugins/capacity/capacity_dra_overflow_test.go
JesseStutler, line 1: no need to define a new seperate test file I think, just put it in capacity_test.go is enough

JesseStutler:
Besides the unit test file overall is lgtm

pkg/scheduler/plugins/capacity/capacity_dra_overflow_test.go
thc1006: Thanks, done. Moved it into capacity_test.go and removed the separate file.

  One thing I noticed while doing that: the guard can still be bypassed from
  upstream. addDRAResource (cache.go) sums the per-request counts with a plain
  +=, so a pod with two same-deviceClass requests each MaxInt64 wraps the
  aggregate negative, and checkDRAAllocatable then admits it (a negative sum
  isn't > capability). SaturatingAdd can't recover a value that's already
  negative.

  I have a small fix ready with tests: reject negative counts in
  checkDRAAllocatable, and optionally SaturatingAdd in addDRAResource too. Want
  it in this PR, or should I keep this one focused and open a follow-up?

  Separately, that same function loops `for i := 1; i < count`, which would spin
  ~2^63 times on such a count. Probably worth its own issue; I can file it if you
  agree.

JesseStutler:
@thc1006 Since there are conflicts could you manually help us to cherry-pick it to release-1.15? Thanks!

thc1006:
> @thc1006 Since there are conflicts could you manually help us to cherry-pick it to release-1.15? Thanks!

Roger!

PR #5625, opened 2026-07-10
fix: close int64-overflow bypass in DRA device count accounting

thc1006:
#### What type of PR is this?
/kind bug

Follow-up to #5621. That PR added saturating arithmetic at three sites, but the DRA quota guard can still be bypassed from upstream.

`pkg/scheduler/cache/cache.go` `addDRAResource` accumulates the per-request device count with a plain `+=`. A pod whose ResourceClaims request the same DeviceClass with counts summing past `math.MaxInt64` (each `count` is individually valid; the apiserver only rejects `count <= 0`) wraps the aggregate to a negative value. That negative flows into `checkDRAAllocatable` as `request.Count`, and `SaturatingAdd(0, negative)` faithfully preserves the negative, so `sum > capability` is false and the request is wrongly admitted.

This PR adds two guards:
- `addDRAResource` accumulates with `SaturatingAdd`, so the aggregate can no longer wrap negative (it saturates to `MaxInt64`, which the quota check rejects for any queue capability below `MaxInt64`, i.e. every real quota).
- `checkDRAAllocatable` rejects a negative `allocated`/`inqueue` running total on a quota-governed DeviceClass. `getDRADelta` adjusts those totals with an unclamped subtract, so releasing an allocation an ancestor never balanced can drive one negative, and `SaturatingAdd` preserves it past the quota check.

Both regression tests (`TestAddDRAResource_saturates`, `TestCheckDRAAllocatable_negativeDeltaRejected`) fail on the pre-fix code and pass with the change.

This PR intentionally does **not** change the unbounded `for i := 1; i < count` loop in the same `addDRAResource` (an unrelated DoS concern); I'm reporting that separately so this stays focused.

Follow-up to #5621.

I found this while addressing review feedback on #5621. AI assistance was used for the analysis and for drafting this change; I have reviewed and understand it and can explain and defend every part, and I will respond to review comments myself.

```release-note
Fixed an int64 overflow in DRA device-count accounting that could bypass the capacity plugin's DRA quota check when a pod requested the same DeviceClass with counts summing past MaxInt64.
```

### DIFF
--- a/pkg/scheduler/cache/cache_test.go
+++ b/pkg/scheduler/cache/cache_test.go
@@ -23,6 +23,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"math"
 	"reflect"
 	"sync"
 	"testing"
@@ -853,4 +854,24 @@ func TestAddUnassignedNumaPods_NilNodeValueSkips(t *testing.T) {
 		t.Fatalf("AddUnassignedNumaPods returned error: %v", err)
 	}
 	// No panic is the success condition.
+}
+
+// addDRAResource accumulates a user-controlled device count. A pod with two
+// same-deviceClass requests each near math.MaxInt64 (both individually valid:
+// the apiserver only rejects count <= 0) must not wrap the aggregate Count to a
+// negative value, which would later bypass the queue's DRA quota check.
+func TestAddDRAResource_saturates(t *testing.T) {
+	const dc = "gpu.example.com"
+	const maxInt64 = math.MaxInt64
+
+	m := make(map[string]*api.DRAResource)
+	addDRAResource(m, dc, maxInt64, nil)
+	addDRAResource(m, dc, maxInt64, nil)
+
+	if got := m[dc].Count; got < 0 {
+		t.Fatalf("addDRAResource wrapped to a negative aggregate Count (%d) for two MaxInt64 requests; want saturated non-negative", got)
+	}
+	if got := m[dc].Count; got != maxInt64 {
+		t.Fatalf("addDRAResource Count = %d for two MaxInt64 requests; want saturated to MaxInt64 (%d)", got, maxInt64)
+	}
 }
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -273,6 +273,20 @@ func checkDRAAllocatable(dra *draQuotaAttr, taskDRA map[string]*api.DRAResource,
 		klog.V(5).Infof("checkDRAAllocatable: deviceClass=%s, allocated=%d, inqueue=%d, request=%d, capability=%d",
 			deviceClass, allocatedCount, inqueueCount, request.Count, capability.Count)
 
+		// A device count is always non-negative; a negative value here can only be
+		// the result of an upstream int64 overflow wrapping around. Reject it
+		// rather than let a poisoned negative pass the comparison below (a negative
+		// sum is not greater than the capability, so it would be wrongly admitted).
+		if request.Count < 0 || allocatedCount < 0 || inqueueCount < 0 {
+			klog.V(3).Infof("checkDRAAllocatable: negative device count for %s (allocated=%d, inqueue=%d, requested=%d); rejecting as int64 overflow",
+				deviceClass, allocatedCount, inqueueCount, request.Count)
+			return false
+		}
+
+		// For any positive capability below MaxInt64 the quota decision is exact: a
+		// sum that overflows saturates to MaxInt64, which still exceeds that
+		// capability and is rejected. A capability of MaxInt64 is effectively
+		// non-binding, since no realizable allocation reaches that many devices.
 		if capability.Count > 0 && api.SaturatingAdd(api.SaturatingAdd(allocatedCount, inqueueCount), request.Count) > capability.Count {
 			klog.V(3).Infof("checkDRAAllocatable: count exceeded for %s: allocated=%d, inqueue=%d, requested=%d, capability=%d",
 				deviceClass, allocatedCount, inqueueCount, request.Count, capability.Count)
--- a/pkg/scheduler/plugins/capacity/capacity_test.go
+++ b/pkg/scheduler/plugins/capacity/capacity_test.go
@@ -2108,3 +2108,42 @@ func TestCheckDRAAllocatable_overflowRejected(t *testing.T) {
 		t.Fatalf("checkDRAAllocatable rejected a valid request (4 <= 8); want admitted")
 	}
 }
+
+// A negative device count can only result from an upstream int64 overflow
+// wrapping around (e.g. a pod whose two same-deviceClass requests, each near
+// math.MaxInt64, are summed in the cache's addDRAResource). Such a poisoned
+// negative must never be admitted: saturating the accumulation is not enough
+// because SaturatingAdd faithfully preserves an already-negative input.
+func TestCheckDRAAllocatable_negativeCountRejected(t *testing.T) {
+	const dc = "gpu.example.com"
+
+	// Negative in the request itself.
+	negRequest := &draQuotaAttr{
+		capability: map[string]*api.DRAResource{dc: {Count: 8}},
+		allocated:  map[string]*api.DRAResource{},
+		inqueue:    map[string]*api.DRAResource{},
+	}
+	if checkDRAAllocatable(negRequest, map[string]*api.DRAResource{dc: {Count: -2}}, false, true) {
+		t.Fatalf("checkDRAAllocatable admitted a negative request.Count (-2) against capability 8; want rejected")
+	}
+
+	// Negative already poisoning the queue's inqueue total.
+	poisonedInqueue := &draQuotaAttr{
+		capability: map[string]*api.DRAResource{dc: {Count: 8}},
+		allocated:  map[string]*api.DRAResource{},
+		inqueue:    map[string]*api.DRAResource{dc: {Count: math.MinInt64}},
+	}
+	if checkDRAAllocatable(poisonedInqueue, map[string]*api.DRAResource{dc: {Count: 4}}, false, true) {
+		t.Fatalf("checkDRAAllocatable admitted a request against a poisoned negative inqueue (MinInt64); want rejected")
+	}
+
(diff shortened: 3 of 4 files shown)


pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -273,6 +273,16 @@ func checkDRAAllocatable(dra *draQuotaAttr, taskDRA map[string]*api.DRAResource,
 		klog.V(5).Infof("checkDRAAllocatable: deviceClass=%s, allocated=%d, inqueue=%d, request=%d, capability=%d",
 			deviceClass, allocatedCount, inqueueCount, request.Count, capability.Count)
```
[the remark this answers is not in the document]
thc1006: taskDRA values are non-nil by construction. they're built by addDRAResource / GetMinDRAResources, which always allocate and the existing debug log and comparison already dereference request.Count, so a nil check here would be unreachable. I'd prefer to keep this PR focused on the overflow, but I'm happy to add the defensive guard if you'd rather have it.

pkg/scheduler/cache/cache_test.go
```diff
@@ -685,3 +685,23 @@ func TestWaitForHandlerSync_InitialEventAsyncHandlerTracker_CompletesAfterDone(t
...
 	assert.GreaterOrEqual(t, elapsed, 100*time.Millisecond, "WaitForHandlerSync should wait until Done is called")
 }
+
+// addDRAResource accumulates a user-controlled device count. A pod with two
+// same-deviceClass requests each near math.MaxInt64 (both individually valid:
+// the apiserver only rejects count <= 0) must not wrap the aggregate Count to a
+// negative value, which would later bypass the queue's DRA quota check.
+func TestAddDRAResource_saturates(t *testing.T) {
+	const dc = "gpu.example.com"
+	const maxInt64 = int64(9223372036854775807)
```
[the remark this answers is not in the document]
thc1006: I didn t expect gemini can really work
Done. Switched to math.MaxInt64 and added the math import.

thc1006:
@hajnalmt @dafu-wu @shinytang6 — PTAL when available.

This PR has been open since 2026-07-10 and contains regression tests for the remaining DRA count wraparound and negative-count quota-check bypass.

I notified the Volcano PST on 2026-07-12, but I have not yet received a security disposition or Fix Lead assignment. A technical review here would help determine whether the change is ready for merge and release-1.15 backport consideration.

I am keeping the additional non-public sibling finding out of this PR.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -273,6 +273,20 @@ func checkDRAAllocatable(dra *draQuotaAttr, taskDRA map[string]*api.DRAResource,
...
 
+		// A device count is always non-negative; a negative value here can only be
+		// the result of an upstream int64 overflow wrapping around. Reject it
+		// rather than let a poisoned negative pass the comparison below (a negative
+		// sum is not greater than the capability, so it would be wrongly admitted).
+		if request.Count < 0 || allocatedCount < 0 || inqueueCount < 0 {
```
JesseStutler: I think the count can be negative since there is a validation in the apiserver: https://github.com/kubernetes/kubernetes/blob/756939600b9a7180fc2df6550a4585b638875e67/pkg/apis/resource/validation/validation.go#L268-L278 ?

thc1006: You're right about request.Count. It's rebuilt per call from the job's requests via GetMinDRAResources (SaturatingMul/SaturatingAdd), and validateDeviceAllocationMode already rejects ExactCount with count <= 0, so it's always >= 0 here. I'll drop that arm and its test case. 

I'll keep the allocatedCount/inqueueCount checks and fix my comment, which is stale. Those are the queue's running totals, adjusted by getDRADelta with a raw new - old subtract; when a saturated allocation is removed the delta is applied through updateDRAAllocated -> mergeTaskDRA -> SaturatingAdd, which doesn't clamp at zero, so a total can go negative. At line 290 SaturatingAdd preserves that negative, and a negative sum isn't greater than the capability, so it would be silently admitted. So the reject is guarding a reachable case, just not the overflow-wrap my comment claims (addDRAResource saturates now). Reworded comment and updated test incoming.

pkg/scheduler/plugins/capacity/capacity.go
```diff
@@ -273,6 +273,23 @@ func checkDRAAllocatable(dra *draQuotaAttr, taskDRA map[string]*api.DRAResource,
...
 			deviceClass, allocatedCount, inqueueCount, request.Count, capability.Count)
 
+		// The queue's running allocated/inqueue totals are adjusted by getDRADelta,
+		// which subtracts old from new without clamping, so removing a saturated
+		// allocation can drive a total below zero. SaturatingAdd below preserves a
+		// negative, so a poisoned total would slip under the capability and be
+		// admitted. request.Count is not checked here: GetMinDRAResources rebuilds it
+		// with SaturatingMul/SaturatingAdd and the apiserver rejects a non-positive
+		// count, so it is always non-negative.
+		if allocatedCount < 0 || inqueueCount < 0 {
```
JesseStutler: Why is there still negative checking? I didn't get the background, could you show your real configuration?

thc1006: Good question, and the old test wasn't a fair answer to it. It set the negative by hand, which doesn't show how a real total gets there, so I've replaced it with one that drives the actual path.

The total that can go negative is the queue's running allocated/inqueue, not request.Count. getDRADelta adjusts those with a raw subtract (`res.Count -= oldRes.Count`, no clamp), and the delta is propagated to ancestor queues through SaturatingAdd, which only guards overflow and preserves a negative. The trigger is a hierarchy where a child allocates a DeviceClass while the parent's capability doesn't list it yet: the add delta is filtered out of the parent, the class is later added to the parent's capability, and the child's release then applies a delta the parent never balanced, so the parent's allocated drops below zero. It also covers an upgrade still carrying a saturated total from before the addDRAResource fix.

The new TestCheckDRAAllocatable_negativeDeltaRejected builds that with getDRADelta and updateDRAAllocated, with no hand-set MinInt64, and checks both that the parent's allocated really goes negative and that checkDRAAllocatable then rejects. It fails without the guard and passes with it.

I did look at clamping the total at zero in updateDRAAllocated instead of rejecting here, to match DRAResource.Sub. I backed off it: a corrupted total floored to zero still under-counts real usage, so the quota check would admit against zero and the bypass survives in a milder form. Rejecting a total that has gone negative is the quota-safe failure mode, and it logs the corruption so an operator can see it. Reaching this at all needs an admin capability change on a parent queue, or a pre-fix residue, rather than a tenant action, so I read it as accounting-integrity defense-in-depth rather than a tenant-facing bypass.

pkg/scheduler/plugins/capacity/capacity_test.go
```diff
@@ -2108,3 +2108,37 @@ func TestCheckDRAAllocatable_overflowRejected(t *testing.T) {
...
+	}
+
+	// Parent got only this release delta; the matching add was filtered out earlier
+	// while dc was absent from its capability.
+	parent := &draQuotaAttr{
+		capability: map[string]*api.DRAResource{dc: {Count: 8}},
+		allocated:  map[string]*api.DRAResource{},
+		inqueue:    map[string]*api.DRAResource{},
+	}
+	updateDRAAllocated(parent, delta)
```
JesseStutler: It should write an e2e testing to test it, rather than directly call seperate functions, real user will not use like this way

thc1006: Thanks, that’s a fair point. I’m happy to move this wherever you think it fits best.

The test does not set the negative value directly. It calls `getDRADelta`, then `updateDRAAllocated`, and finally `checkDRAAllocatable`, so it exercises the actual accounting path.

I kept it at the function level because this state cannot arise through a normal user or admin flow. It requires a parent queue whose capability does not include a `DeviceClass` already allocated by a child, followed by that class being added to the parent’s capability and the child later releasing its allocation. The parent then receives a release delta for an allocation it never accounted for in the first place.

An e2e test would need to reproduce that exact reconfiguration while the workload is running. That would be relatively heavy and potentially flaky, while still testing a state caused only by the misconfiguration rather than a normal user path.

I can also move this to a higher-level test and exercise it through the plugin’s enqueue path, similar to `TestJobEnqueueableChecksDRAWithoutMinResources`. Alternatively, since this is defense-in-depth for a misconfiguration rather than the overflow issue this PR is primarily addressing, I’m also fine splitting it into a follow-up so the `SaturatingAdd` fix can land independently. Either direction works for me.

JesseStutler: If you call several different functions, it's deviating from the scheduler's standard invocation path. Therefore, either remove the `<0` check or ensure you reproduce the specific scenario and provide the e2e test that prove you can solve it

thc1006: Removed the <0 check and its test. You asked earlier for a real configuration that reaches it and I couldn't produce one. It guards a parent-queue accounting asymmetry that a user request has no way to drive, and an e2e would need that same unreachable path, so dropping the check is the honest call here. The overflow a user can reach is handled by the SaturatingAdd in addDRAResource, and I've kept that with its test.

If the accounting asymmetry is worth tracking on its own, I can open a separate issue for it.

JesseStutler:
lgtm
Hi @thc1006 Please resolve the conflicts and push again, thanks

thc1006:
Thanks @JesseStutler. Rebased onto master at 65f4d18b and pushed.

The conflict was only in `cache_test.go`: master's new `updateJobInfo` and `taskUnschedulable` tests and mine had both been appended at the end of the file. I kept both, so the change itself is untouched, still the same three files at +25/-1.

thc1006:
CI came back with one red check, `e2e-spark`. Master is failing it the same way right now: the buildx builder cannot pull `moby/buildkit:buildx-stable-1`, so the Spark image never builds and none of the Volcano code under test runs. #5976 tracks it and #5977 pins the image. The other 30 checks are green, including the remaining 22 e2e jobs.

JesseStutler:
Never mind, we can fix the spark failure in later separate PRs

PR #5628, opened 2026-07-10
example/extender: limit HTTP request body size

yash-pokhriyal:
## What does this PR do?

Adds a maximum request body size limit for the extender example HTTP server using `http.MaxBytesReader`.

## Why is this change needed?

Although this is an example server, limiting request body size is a good security practice that helps prevent excessive memory usage from oversized requests.

## Does this introduce breaking changes?

No.

### DIFF
--- a/example/extender/extender.go
+++ b/example/extender/extender.go
@@ -36,11 +36,7 @@ func onSessionOpen(w http.ResponseWriter, r *http.Request) {
 
 	content, err := io.ReadAll(r.Body)
 	if err != nil {
-		if _, ok := err.(*http.MaxBytesError); ok {
-			http.Error(w, "request body too large", http.StatusRequestEntityTooLarge)
-		} else {
-			http.Error(w, "failed to read request body", http.StatusBadRequest)
-		}
+		http.Error(w, "failed to read request body", http.StatusInternalServerError)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


example/extender/extender.go
```diff
@@ -58,14 +66,19 @@ func onSessionClose(w http.ResponseWriter, r *http.Request) {
 }
 
 func predicate(w http.ResponseWriter, r *http.Request) {
+	r.Body = http.MaxBytesReader(w, r.Body, maxRequestBody)
+	defer r.Body.Close()
+
 	content, err := io.ReadAll(r.Body)
 	if err != nil {
-		w.WriteHeader(http.StatusInternalServerError)
+		if _, ok := err.(*http.MaxBytesError); ok {
```
hajnalmt: This is not too idiomatic, but works now because the error is not unwrapped.

Importing `errors` and using:
```
var maxErr *http.MaxBytesError
if errors.As(err, &maxErr) {
```
would have been the way to go, but this is not blocker for me, I am fine without this change.

example/extender/extender.go
```diff
@@ -58,14 +66,19 @@ func onSessionClose(w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, "request body too large", http.StatusRequestEntityTooLarge)
+		} else {
+			http.Error(w, "failed to read request body", http.StatusBadRequest)
```
hajnalmt: `http.StatusInternalServerError` was more accurate here please change it back.

hajnalmt (commented):
/kind documentation

Thanks for the PR, this is low risk. Can you change the return code back to 500, we shouldn't have 400 here. Remarks are below.

example/extender/extender.go
yash-pokhriyal: Thanks for the review! I've addressed the requested feedback by changing the error handling back to http.StatusInternalServerError while keeping the request body size limit in place. Thank you for the suggestion!

PR #5629, opened 2026-07-10
docs:add namespaceQueue design

gitGurugu:
What type of PR is this?
/kind feature

This PR adds a design proposal for a namespace-scoped NamespaceQueue CRD,
letting tenants author and manage queue hierarchies within their own namespace
without cluster-admin intervention.

The proposal derives NamespaceQueue from the cluster-scoped Queue (reusing
QueueSpec, so guarantee/deserved/capability semantics are unchanged). It uses a
namespace-qualified QueueID and namespace-first PodGroup.spec.queue
resolution, so the capacity and proportion plugins need no behavioral change.
The whole feature sits behind a default-off NamespaceQueue feature gate, so
clusters that do not opt in are unchanged.

This is a draft opened to align on the design (see the Open questions
section) before implementation. It contains the design doc plus the feature-gate
registration only — no CRD/controller/scheduler code yet.

Which issue(s) this PR fixes:
Related to https://github.com/volcano-sh/volcano/issues/5251

This is a design-proposal draft PR; I'd especially appreciate feedback on the
Open questions for review section (parent disambiguation, mid-flight
re-resolution, status struct, RBAC packaging).
AI disclosure: I used an AI coding assistant to help draft and structure the
proposal document. I have reviewed and understand every part of it, and will
respond to review comments myself.

### DIFF
--- a/docs/design/namespace-queue.md
+++ b/docs/design/namespace-queue.md
@@ -1,0 +1,547 @@
+# Namespace-scoped Queues in Volcano
+
+[gitGurugu](https://github.com/gitGurugu); Jul 10, 2026
+
+Tracking issue: [#5251](https://github.com/volcano-sh/volcano/issues/5251)
+
+## 1. Motivation
+
+Volcano Queue is cluster-scoped, so creating or modifying a queue normally requires cluster-level permissions. In a multi-tenant cluster, tenants typically manage resources within their own namespaces. Requiring cluster administrators to create or modify queues limits tenant self-service, while granting tenants access to cluster-scoped Queues weakens namespace isolation and complicates resource ownership.
+
+`NamespaceQueue` addresses this gap by providing a namespaced queue resource that tenants can manage within their own namespaces. It supports namespace-local queue hierarchies and resource governance while allowing cluster administrators to control how those hierarchies attach to cluster-scoped Queues. Workloads can reference a NamespaceQueue explicitly, and existing cluster Queue behavior remains unchanged.
+
+## 2. Scope
+
+In Scope:
+
+- introduce the namespaced `NamespaceQueue` CRD and its resource-governance fields;
+- support namespace-local hierarchies and attachment to authorized cluster-scoped Queues;
+- allow workloads to reference NamespaceQueues through the common scheduler model; and
+- preserve existing Queue workload references and scheduling behavior while extending `QueueSpec` with optional namespace authorization.
+
+Out of Scope:
+
+- cross-namespace NamespaceQueue parent references;
+- materializing NamespaceQueues as cluster-scoped Queues or creating ownership relationships between them;
+- changing the Kubernetes namespace and RBAC model; and
+- NamespaceQueue support in scheduling plugins beyond the initial capacity plugin.
+
+## 3. Detailed Design
+
+### 3.1 Implementation Overview
+
+NamespaceQueue integrates with the API, admission, controller, and scheduler components. Their responsibilities are defined as follows:
+
+| Component | Responsibility |
+| --- | --- |
+| API and CRD | Define the `NamespaceQueue` resource, schema, object-local validation, defaulting, and status subresource. |
+| Admission Webhook | Perform request-time validation of workload queue references and best-effort validation of `NamespaceQueue` parent authorization. It does not maintain dynamic hierarchy state or determine scheduling readiness. |
+| `NamespaceQueue` controller | Resolve parent references, reconcile hierarchy and authorization, manage lifecycle and status, and requeue affected subtrees. |
+| Scheduler | Watch `Queue` and `NamespaceQueue` resources, build the unified `QueueInfo` cache, and perform the final scheduling-readiness check. |
+| Existing workload controllers | Preserve raw queue references when creating or updating `PodGroup` and `Pod` objects. They do not create shadow `Queue` resources. |
+
+### 3.2 API and CRD
+
+#### 3.2.1 Resource Definition
+
+`NamespaceQueue` is a namespaced custom resource used by Volcano scheduling. It contains the desired queue configuration in `spec` and the observed state in `status`. The `status` subresource is updated independently from the desired configuration.
+
+```go
+// NamespaceQueue is a namespaced queue resource used by Volcano scheduling.
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=namespacequeues,scope=Namespaced,shortName=nq
+// +kubebuilder:printcolumn:name="PARENT",type=string,JSONPath=`.spec.parent`
+// +kubebuilder:printcolumn:name="STATE",type=string,JSONPath=`.status.state`
+// +kubebuilder:printcolumn:name="AGE",type=date,JSONPath=`.metadata.creationTimestamp`
+type NamespaceQueue struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Standard object metadata.
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Desired queue configuration.
+	// +optional
+	Spec NamespaceQueueSpec `json:"spec,omitempty"`
+
+	// Observed queue state.
+	// +optional
+	Status NamespaceQueueStatus `json:"status,omitempty"`
+}
+```
+
+
+#### 3.2.2 NamespaceQueueSpec
+
+`NamespaceQueueSpec` defines the desired state of a `NamespaceQueue`.
+
+
+```go
+// NamespaceQueueSpec defines the desired state of a NamespaceQueue.
+type NamespaceQueueSpec struct {
+
+	// Capability specifies the maximum resource allocation for this queue.
+	//
+	// +optional
+	Capability v1.ResourceList `json:"capability,omitempty"`
+
+	// Reclaimable indicates whether unused resources can be reclaimed.
+	//
+	// +optional
+	Reclaimable *bool `json:"reclaimable,omitempty"`
+
+	// Guarantee specifies the resources reserved for this queue.
+	//
+	// +optional
+	Guarantee Guarantee `json:"guarantee,omitempty"`
+
+	// Parent identifies a NamespaceQueue in the same namespace or a
+	// cluster-scoped Queue. The reference format is defined in Section 3.4.1.
+	//
+	// +optional
+	// +kubebuilder:default:=cluster/default
+	Parent string `json:"parent,omitempty"`
+
+	// Deserved specifies the expected resource allocation. Excess resources
+	// may be shared and reclaimed when required.
+	//
+	// +optional
+	Deserved v1.ResourceList `json:"deserved,omitempty"`
+
+	// Priority specifies the scheduling priority of workloads in this queue.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	Priority int32 `json:"priority,omitempty"`
+
+	// DequeueStrategy specifies how workloads are selected from this queue.
+	//
+	// +optional
+	// +kubebuilder:default:=traverse
+	// +kubebuilder:validation:Enum=fifo;traverse
+	DequeueStrategy DequeueStrategy `json:"dequeueStrategy,omitempty"`
+}
+```
+
+The resource fields define the NamespaceQueue's resource partition. API-level and hierarchy-level validation rules are described in Sections 3.2 and 3.4.3.
+
+#### 3.2.3 Basic Example
+
+The following example creates a `NamespaceQueue` named `training` in the `team-a` namespace. It uses the cluster-scoped Queue `research` as its parent. The parent Queue must exist and include `team-a` in its `spec.allowedNamespaces`.
+
+The `status` field is omitted because it is maintained by the `NamespaceQueue` controller. The `dequeueStrategy` field is shown explicitly; when omitted, it defaults to `traverse`.
+
+```yaml
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: NamespaceQueue
+metadata:
+  name: training
+  namespace: team-a
+spec:
+  parent: cluster/research
+  capability:
+    cpu: "100"
+    memory: "200Gi"
+  deserved:
+    cpu: "60"
+    memory: "120Gi"
+  guarantee:
+    resource:
+      cpu: "20"
+      memory: "40Gi"
+  reclaimable: true
+  dequeueStrategy: traverse
+```
+
+
+#### 3.2.4 API Validation and Defaulting
+
+API validation is limited to object-local invariants. These invariants are enforced by the CRD structural schema and CEL where possible. Request-time reference validation is performed by the admission webhook, while validation that depends on parent or sibling state is performed by the `NamespaceQueue` controller.
+
+The following rules apply during API admission:
+
+- `spec.parent` must use the supported parent-reference format;
+- resource quantities must be valid Kubernetes quantities;
+- for each resource for which the corresponding values are specified, `guarantee <= deserved <= capability`;
+- `spec.dequeueStrategy` must be `fifo` or `traverse`;
+- `spec.priority` must be non-negative; and
+- cross-namespace parent references are not supported.
+
+The API applies the following defaults to omitted fields:
+
+- `spec.parent` defaults to `cluster/default`;
+- `spec.dequeueStrategy` defaults to `traverse`.
+
+Defaulting does not grant authorization, establish parent existence, or make a `NamespaceQueue` ready for scheduling. Parent existence, authorization, hierarchy constraints, and scheduling readiness are evaluated by the `NamespaceQueue` controller as described in Section 3.4.
+
+#### 3.2.5 Cluster Queue Authorization
+
+Cluster administrators control which namespaces may attach NamespaceQueue hierarchies to a cluster-scoped Queue. The cluster-scoped `QueueSpec` includes the following optional field:
+
+```go
+type QueueSpec struct {
+	// Existing scheduling fields are omitted.
+
+	// AllowedNamespaces lists the namespaces whose NamespaceQueues may use
+	// this Queue as their cluster-scoped parent. Each entry must be a valid
+	// Kubernetes namespace name or the literal "*". The literal "*" allows
+	// all namespaces and must be the only entry when present.
+	//
+	// +optional
+	AllowedNamespaces []string `json:"allowedNamespaces,omitempty"`
+}
+```
+
+An empty or omitted list denies NamespaceQueue attachment. A literal `*` allows NamespaceQueues from every namespace and must not be combined with namespace names. Each non-wildcard entry must be a valid Kubernetes namespace name. Regular-expression and other pattern matching are out of scope for this design. The cluster-scoped `default` Queue is initialized with `allowedNamespaces: ["*"]`, so it permits NamespaceQueue attachment from all namespaces by default.
+
+For example:
+
+```yaml
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: research
+spec:
+  allowedNamespaces:
+    - team-a
+    - team-b
+```
+
+A NamespaceQueue in `team-a` may use `cluster/research` as its parent, while a NamespaceQueue in `team-c` may not unless `team-c` is listed in `research.spec.allowedNamespaces`. A Queue with `allowedNamespaces: ["*"]`, including the initialized `cluster/default` Queue, permits attachment from any namespace.
+
+A NamespaceQueue has a cluster-scoped effective parent when its `spec.parent` resolves to `cluster/<queue-name>`. An update to `Queue.spec.allowedNamespaces` must be rejected if the resulting value would make an existing NamespaceQueue with that effective parent unauthorized. An administrator must first update each affected NamespaceQueue to use another valid parent, or close it and allow its admitted workloads to complete, before removing the namespace from `allowedNamespaces`.
+
+#### 3.2.6 NamespaceQueueStatus
+
+`NamespaceQueueStatus` represents the observed state of a `NamespaceQueue`. It includes:
+
+- the queue lifecycle `state`;
+- `status.conditions`, including the `Authorized` and `Ready` condition types;
+- workload counters derived from associated `PodGroup` resources; and
+- runtime resource fields such as `allocated` and `reservation`.
+
+The `NamespaceQueue` controller owns the lifecycle state, conditions, and workload counters. The Scheduler owns runtime allocation and reservation fields. Both components update their owned fields through the `status` subresource. The condition semantics and status transitions are defined in Section 3.4.2.
+
+### 3.3 Admission Webhook
+
+The admission webhook performs synchronous validation for API requests. It is responsible for:
+
+- validating queue references in `Job`, `PodGroup`, and supported queue annotations against the shared queue-reference grammar defined in Section 3.6;
+- honoring existing workload queue defaulting and annotation precedence rules; an omitted workload queue annotation is not a validation error;
+- validating the effective `NamespaceQueue.spec.parent` after API defaulting; an omitted or empty parent resolves to `cluster/default`;
+- performing a best-effort check that the referenced cluster-scoped `Queue` exists and that the `NamespaceQueue` namespace is allowed by the Queue's `spec.allowedNamespaces`;
+- rejecting an update to a cluster-scoped Queue's `spec.allowedNamespaces` if the resulting value would make an existing NamespaceQueue with that effective parent unauthorized;
+- rejecting malformed or unsupported queue references and references that fail authorization checks; and
+- performing an initial deletion-safety check for `NamespaceQueue` deletion requests.
+
+Admission validation applies to individual API requests. The webhook does not maintain parent hierarchy state, reconcile subtrees, manage lifecycle transitions, or continuously reconcile existing objects. The `NamespaceQueue` controller remains responsible for parent resolution, dynamic authorization, readiness, lifecycle, and deletion safety, as described in Section 3.4. The Scheduler performs the final scheduling eligibility check.
+
+The detailed workload reference contract is defined in Section 3.6.
+
+### 3.4 NamespaceQueue Controller
+
+The NamespaceQueue Controller resolves NamespaceQueue parents, evaluates authorization and hierarchy constraints, updates lifecycle and status, and reconciles affected subtrees when related NamespaceQueues, cluster-scoped Queues, or PodGroups change. Existing cluster Queue Controller behavior remains unchanged.
+
+
+#### 3.4.1 Parent Resolution
+
+The reference forms are:
+
+| Reference | Meaning |
+| --- | --- |
+| `<name>` | NamespaceQueue `<name>` in the current namespace |
+| `cluster/<name>` | Cluster-scoped Queue `<name>` |
+| omitted or empty | `cluster/default` |
+| `cluster/root` | Invalid: the scheduler-managed root queue cannot serve as a tenant parent. |
+
+The CRD default applies when `spec.parent` is omitted. During parent resolution, an omitted or explicitly empty `spec.parent` resolves to `cluster/default`. The admission webhook, `NamespaceQueue` controller, and Scheduler use the same effective parent resolution rules.
+
+The NamespaceQueue Controller resolves each API reference to a controller-internal target identifying the referenced Kubernetes object. The target must exist and satisfy authorization, hierarchy, and resource constraints; the `cluster/` prefix identifies a cluster-scoped parent and is removed before the controller looks up the Queue by name.
+
+The target can be represented internally as:
+
+```go
+type QueueScope string
+
+const (
+	ClusterQueueScope   QueueScope = "cluster"
+	NamespaceQueueScope QueueScope = "namespace"
+)
+
+type QueueTarget struct {
+	Scope     QueueScope
+	Namespace string     // empty for a cluster-scoped Queue
+	Name      string
+}
+```
+
+The target scope determines which lister is used:
+
+```txt
+Cluster target
+    -> cluster Queue lister, keyed by name
+
+Namespace target
+    -> NamespaceQueue lister, keyed by namespace and name
+```
+
+The resolver may be implemented as follows:
+
+```go
+func ResolveNamespaceQueueParent(nq *NamespaceQueue) (QueueTarget, error) {
+	parent := nq.Spec.Parent
+	if parent == "" {
+		parent = "cluster/default"
+	}
+
+	if strings.HasPrefix(parent, "cluster/") {
+		name := strings.TrimPrefix(parent, "cluster/")
+		if name == "" || name == "root" {
+			return QueueTarget{}, fmt.Errorf("invalid cluster NamespaceQueue parent")
+		}
+		return QueueTarget{Scope: ClusterQueueScope, Name: name}, nil
+	}
+
+	return QueueTarget{
+		Scope:     NamespaceQueueScope,
+		Namespace: nq.Namespace,
+		Name:      parent,
+	}, nil
+}
+```
+
+After resolving the target, the controller looks up the parent object, validates its authorization and hierarchy state, and records the resulting parent-child relationship. A lightweight index is sufficient for reconciliation:
+
+```go
+childrenByParent map[QueueTarget]map[QueueTarget]struct{}
+```
+
+This index is used to find NamespaceQueue descendants and requeue them after a parent or sibling change.
+
+#### 3.4.2 Authorization and Continuous Reconciliation
+
+At most one NamespaceQueue subtree per namespace may attach to a given cluster-scoped Queue. Additional NamespaceQueues below that attachment point only subdivide the namespace's assigned share.
+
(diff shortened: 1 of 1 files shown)


docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1502 @@
...
+```yaml
+metadata:
+  namespace: team-a
+spec:
+  queue: training
+```
+
+The persisted workload object continues to store training. The scheduler resolves this value into a canonical QueueID internally.
+
+#### Namespace-first Resolution
```
devzizu: This resolution model doesn't seem to match what we agreed in the last community meeting. We settled on **explicit prefixing** for workload queue references, not namespace-first auto-resolution:

  - `spec.queue: training` should resolve to a cluster Queue `training`.
  - `spec.queue: namespace/training` should resolve to the `NamespaceQueue` `training` in the current namespace.

> Note: the same reasoning should apply to the existing queue annotation that goes to workload objects: Pods, PodGroups, etc.

The main reason a bare name must not implicitly resolve to a NamespaceQueue is  **backward compatibility**. With namespace-first, an existing workload using `spec.queue: training` silently starts pointing at a newly created `team-a/training` NamespaceQueue.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1502 @@
...
+The following hierarchy is invalid:
+```
+
+This restriction applies only to the system cluster queue referenced as cluster/root.
+
+#### Cluster Queue Authorization
+
+Cluster administrators control which namespaces may attach NamespaceQueue hierarchies to a cluster-scoped Queue.
+
+The cluster-scoped QueueSpec is extended with an allowedNamespaces field:
```
devzizu: This section only describes **authorization** as a **create-time gate**. `allowedNamespaces` is a mutable admin field, so we also need to define what happens on **revocation** of an already-admitted, running subtree.

Please state that authorization is re-evaluated on every `allowedNamespaces` change, not only at creation, and that revocation **never force-evicts running workloads**. The affected subtree could transition to a Closing/inert state, but here I'm not entirely sure what we should do next.

I think Volcano's behavior on parent deletion is preventing during admission time (while parent has at least a child), and there is no "drain process" for the queue.

WDYT? @JesseStutler @hajnalmt

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1502 @@
...
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Weight int32 `json:"weight,omitempty" protobuf:"varint,1,opt,name=weight"`
+
+	// Capability defines the maximum amount of resources that this queue
+	// is allowed to use.
+	//
+	// +optional
+	Capability v1.ResourceList `json:"capability,omitempty" protobuf:"varint,2,opt,name=capability"`
```
devzizu: Let's not hand-write protobuf types here, we can exclude them from the design. In this case, for Capability, it should be `bytes` instead of `varint` (check the [Queue resource](https://github.com/volcano-sh/apis/blob/master/pkg/apis/scheduling/v1beta1/types.go#L467), for example).

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1502 @@
...
+	// may be reclaimed by other queues.
+	//
+	// +optional
+	Reclaimable *bool `json:"reclaimable,omitempty" protobuf:"bytes,3,opt,name=reclaimable"`
+
+	// ExtendClusters defines the remote clusters to which workloads in this
+	// queue may be dispatched.
+	//
+	// +optional
+	ExtendClusters []Cluster `json:"extendClusters,omitempty" protobuf:"bytes,4,opt,rep,name=extendClusters"`
```
devzizu: Is this a tenant-settable field? This looks like a cross-cluster capability a namespace tenant arguably shouldn't grant themselves, but up for discussion.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1502 @@
...
+	//
+	// +optional
+	Deserved v1.ResourceList `json:"deserved,omitempty" protobuf:"bytes,9,opt,name=deserved"`
+
+	// Priority defines the scheduling priority of this queue. Higher values
+	// are scheduled earlier and considered later during reclamation.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	Priority int32 `json:"priority,omitempty" protobuf:"varint,10,opt,name=priority"`
```
devzizu: I'm questioning how we should interpret this field...`capability`/`deserved`/`guarantee` are validated against the parent, but priority and weight aren't, and they're global scheduling-order/share levers. Under a shared admin-owned cluster parent, a tenant could set an arbitrarily high priority (or inflate weight) on their `NamespaceQueue` and jump ahead of sibling tenants, or even other queues, within that partition. That partially defeats the "admin controls the partition" model.

Should we reconsider this, specially for proportion plugin?

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1502 @@
...
+                  └── missing ──→ workload remains unschedulable
+```
+
+
+### NamespaceQueue Status and Events
+
+NamespaceQueue maintains its own status and does not depend on a generated cluster-scoped Queue.
+
+
+The scheduler owns the authoritative runtime information for each queue through the internal QueueInfo representation. This information is used to update the corresponding NamespaceQueue.status.
```
devzizu: The design covers status but not metrics, so can we add a section about it? Volcano's queue metrics are labeled only by `queue_name`, e.g. `volcano_queue_allocated_milli_cpu{queue_name="training"}`. With NSQ, a cluster Queue `training` and `NamespaceQueue` `team-a/training` may collide on the same series (if we emit Name).

Should we introduce `queue_scope` in metrics?

devzizu (commented):
Thank you @gitGurugu, nice progress! This new design is much more aligned with what we agreed. I have a few questions that I believe we need to sort out before implementation.

docs/design/namespace-queue.md
gitGurugu: Thanks for pointing this out. Mentor，You are right.I will update the design

gitGurugu: Hi，mentor，I have updated the design accordingly.

docs/design/namespace-queue.md
gitGurugu: Thanks for pointing this out. I agree that allowedNamespaces should not be treated only as a create-time authorization gate.

My proposed revocation semantics are:
Authorization is re-evaluated whenever Queue.spec.allowedNamespaces changes.
If a namespace is no longer authorized, the entire NamespaceQueue subtree attached from that namespace becomes non-admitting.
Running workloads are not forcefully evicted and are allowed to complete normally.
New workloads targeting the affected subtree are rejected, and existing pending workloads are not scheduled.
If the subtree still has running workloads, it enters a Closing state. After all running workloads finish, it transitions to a Closed or inactive state.
The NamespaceQueue objects are preserved rather than automatically deleted.
If authorization is restored, the subtree returns to Open, and previously pending workloads may resume scheduling.

The controller would update an Authorized=False condition with a reason such as NamespaceNotAllowed, while admission and scheduler-side checks would enforce the non-admitting behavior.

I will update the design to describe these semantics. Does the Open -> Closing -> Closed lifecycle sound reasonable, or would you prefer a simpler inactive state for the revoked subtree?

docs/design/namespace-queue.md
gitGurugu: Thanks for pointing this out. I will remove the hand-written protobuf tags from the design document and rely on the generated definitions in the implementation.

docs/design/namespace-queue.md
gitGurugu: Good point.Cross-cluster configuration will remain controlled by the cluster-scoped parent Queue and will be inherited by the NamespaceQueue hierarchy.

docs/design/namespace-queue.md
gitGurugu: Good point. I agree that Priority and Weight should not be directly controlled by namespace tenants, since they affect global scheduling order and resource shares rather than only limiting resources within the tenant subtree.

For the initial design, I will remove both fields from NamespaceQueueSpec. NamespaceQueues will use a fixed default weight, while priority remains controlled by the cluster-scoped parent Queue. This prevents a tenant from increasing its share or scheduling precedence relative to sibling tenants.

To avoid share multiplication through multiple attachment points, the design will also permit at most one root NamespaceQueue for each (namespace, parent Queue) pair. Any hierarchy below that root only divides the share already assigned to that namespace.

If per-namespace weights are needed in the future, they should be configured by the cluster administrator on the cluster-scoped Queue side rather than by the tenant-owned NamespaceQueue.

I will update the design accordingly, especially the proportion plugin section.

docs/design/namespace-queue.md
gitGurugu: I will add an observability section covering queue metrics.

I agree that emitting only metadata.name as queue_name would cause collisions. However, queue_scope alone would not be sufficient, because NamespaceQueues with the same name may also exist in different namespaces.

I propose using the following labels for all queue metrics:

queue_scope: cluster or namespace
queue_namespace: empty for a cluster-scoped Queue, and the namespace for a NamespaceQueue
queue_name: the object name

For example:

{queue_scope="cluster", queue_namespace="", queue_name="training"}
{queue_scope="namespace", queue_namespace="team-a", queue_name="training"}

The metric names will remain unchanged. The scheduler will derive these labels from the normalized QueueInfo, so both Queue types follow the same metrics path. This avoids collisions between cluster-scoped Queues and NamespaceQueues, as well as between identically named NamespaceQueues in different namespaces.

I will also document the compatibility impact of adding the new labels to the existing queue metric series.

docs/design/namespace-queue.md
devzizu: As a summary, this is how I currently see the problem:

- `weight` is only dangerous on the proportion plugin, where every queue competes for the [whole cluster by weight](https://github.com/volcano-sh/volcano/blob/v1.15.0/pkg/scheduler/plugins/proportion/proportion.go#L200-L230), so a tenant could inflate it and cross into other people's share. The `capacity` plugin (where the [hierarchical-queue feature lives](https://github.com/volcano-sh/volcano/blob/v1.15.0/pkg/scheduler/plugins/capacity/capacity.go#L1196)) doesn't use `weight` at all and partitions by `deserved`/`capability`/`guarantee`, so a child is already bounded by its parent there.

- `priority` is worse, and it's also dangerous in `capacity`. The queue ordering function [compares priority globally as the first tiebreaker](https://github.com/volcano-sh/volcano/blob/v1.15.0/pkg/scheduler/plugins/capacity/capacity.go#L1348-L1367), before any hierarchy logic runs. So a tenant setting a high priority on their queue jumps ahead of everyone regardless of where they sit in the tree.

So, we have a few options:

1. Scope NSQ to the `capacity` plugin **only**: NSQ is inherently hierarchical, and hierarchy is already capacity-only. Then, `weight` is a non-issue since capacity does not use it, and we drop `priority` (fixing it to a neutral value).
2. Remove `weight` and `priority` from the NSQ CRD: Simplest to implement and definitely safe, but it removes any per-queue lever tenants could use under DRF/proportion.

@hajnalmt @JesseStutler do you see any other way around this issue?

docs/design/namespace-queue.md
devzizu: If we add `queue_namespace` than this is essentially a breaking change for metrics right? Because any existing dashboards that have `queue_name` may match both cluster or namespace scoped queues.

Should we separate the metric series family? For example, consider adding `volcano_namespacequeue...*`? or prefix queue name with <namespace/name>?

cc @hajnalmt @JesseStutler

docs/design/namespace-queue.md
devzizu: > Running workloads are not forcefully evicted and are allowed to complete normally.
> New workloads targeting the affected subtree are rejected, and existing pending workloads are not scheduled.

Agree with the "let running finish, reject new" behavior. One clarification: the **scheduler** must be the authoritative enforcer (it re-derives authorization from the live `allowedNamespaces` + tree each session), because a workload admitted *before* revocation is already in the queue and the webhook can't retroactively touch it. The admission webhook is a best-effort fast-fail for new workloads, and worth noting the small race window between the admin's edit and enforcement.

> If the subtree still has running workloads, it enters a Closing state. After all running workloads finish, it transitions to a `Closed` or inactive state.
>
> Does the Open -> Closing -> Closed lifecycle sound reasonable, or would you prefer a simpler inactive state for the revoked subtree?

This is the part I'd change, I would not encode revocation in `State`. In Volcano, queue `State` (`Open`/`Closing`/`Closed`) is the queue's *own* lifecycle, driven by the controller and explicit open/close (`pkg/controllers/queue/state`). Overloading it with "parent revoked" conflates, I think. Namely, you can no longer distinguish "admin closed this queue" from "authorization was revoked," and restore becomes ambiguous if both happened. I would keep authorization as the `Authorized` condition you proposed and **derive** the effect: a queue participates only if `State==Open && Authorized==True`.

I would like to get the view of the other mentors here, specially because this is a different behavior that Volcano may not be implementing today and we should be careful proceeding with this. One bigger thing to review is this drain/non-admitting path is **inconsistent with how Volcano handles parent deletion today**, which is *prevention at admission* (you can't delete a parent that still has children). We'd now have two similar situations behaving differently. We should pick one model deliberately: 

1. prevent revocation while an NSQ is still attached.
2. allow + non-admitting/drain (this proposal). 

@JesseStutler @hajnalmt, any thoughts here?

docs/design/namespace-queue.md
JesseStutler: I think we can remove the weight field and only support the NSQ first, the priority may affect the VictimQueueOrder in the reclaim action. But I think even "deserved/capability" also has some issues. @devzizu , how do your platform controls nsq can create what value of deserved/capability? Or is it just a check to ensure it doesn't exceed the upper cluster scope queue?

devzizu: Yes, as discussed in the community meeting, tenants are free to set their values, as such, it's just a check iit doesn't exceed the upper cluster scope queue.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,2220 @@
...
+full annotation together with the Pod namespace and scope label. Existing
+Pods without the scope label retain the current bare-name label behavior.
+
+The `volcano.sh/queue-scope` label is proposed to make same-name queues
+unambiguous for Pod queries. Its addition and the corresponding CLI filtering
+behavior require confirmation.
+
+#### Admission Webhook Contract
+
+The admission webhook and the scheduler must use the same queue-reference
```
gitGurugu: Thank you for reviewing this design. Could you please confirm whether a new workload referencing a missing or `Ready=False` NamespaceQueue should be rejected by the admission webhook, or accepted and kept pending until the queue becomes ready?

Our current proposal is to reject it when the webhook has sufficient informer state, while keeping the scheduler as the authoritative enforcement point.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,2220 @@
...
+```
+
+The queue label is a compatibility and query field, not the scheduler's
+source of truth. The scheduler resolves the raw reference from the workload
+API or annotation and uses the canonical `QueueID`. Consumers that need to
+distinguish `Queue/training` from `NamespaceQueue team-a/training` must use the
+full annotation together with the Pod namespace and scope label. Existing
+Pods without the scope label retain the current bare-name label behavior.
+
+The `volcano.sh/queue-scope` label is proposed to make same-name queues
```
gitGurugu: Could you please confirm whether adding the proposed `volcano.sh/queue-scope` label is acceptable?

The complete queue reference would remain in the annotation, while the existing label would store only the local queue name. The additional scope label would distinguish a cluster Queue from a NamespaceQueue with the same name.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,2220 @@
...
+not implicitly select a NamespaceQueue with the same name. A workload must use
+the explicit `namespace/default` reference to select the NamespaceQueue in its
+own namespace.
+
+If `NamespaceQueue.spec.parent` is omitted, it independently defaults to the
+cluster-scoped parent reference `cluster/default`. This parent default does not
+change the queue selected by workloads and does not make the NamespaceQueue an
+implicit default for the namespace.
+
+Existing Volcano admission behavior may rewrite a PodGroup whose queue is
```
gitGurugu: Could you please confirm whether the existing behavior of rewriting a PodGroup's `default` queue from the namespace-level queue annotation should be preserved?

Current preference is to retain it for backward compatibility, while allowing users to explicitly select a NamespaceQueue with `namespace/default`.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,2220 @@
...
+  workloads may then be scheduled again.
+
+The scheduler treats both `Closing` and `Closed` as non-admitting states. The
+controller retains pending workload objects and does not mutate them into a
+failed or cancelled state solely because the NamespaceQueue was closed.
+
+<!--
+Pending design decision: behavior of already-pending workloads during close.
+
+Option A (recommended): preserve pending workloads and pause scheduling. This
```
gitGurugu: Could you please confirm whether the proposed drain-and-preserve behavior is appropriate?

Under this model, new workloads are rejected, running workloads are allowed to finish, and already-admitted pending workloads remain stored but are not scheduled until the NamespaceQueue is reopened.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,2220 @@
...
+
+#### Deletion
+
+A NamespaceQueue cannot be deleted while it has child NamespaceQueues or
+allocated workloads. The admission webhook rejects such deletion requests,
+following the existing cluster Queue deletion behavior. The controller also
+rechecks deletion safety and must not remove the NamespaceQueue from its
+active hierarchy while it still has children or allocated workloads.
+
+The current implementation proposal uses a controller finalizer for this
```
gitGurugu: Could you please confirm whether controller finalizer protection is acceptable here?

The webhook would provide fast validation, while the finalizer would prevent deletion races when children or resource allocations appear after the webhook check. This finalizer would protect the NamespaceQueue itself and would not create ownership of another Queue resource.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,2220 @@
...
+- Job queue-reference resolution;
+- PodGroup queue-reference resolution;
+- existing queue annotation and default-queue compatibility.
+
+The webhook must have access to the NamespaceQueue and cluster Queue informer
+caches used by the shared resolver. Webhook failure behavior must preserve
+existing cluster Queue compatibility while preventing invalid NamespaceQueue
+references from entering scheduling.
+
+#### Rollout and Compatibility
```
gitGurugu: Could you please advise whether NamespaceQueue support should be introduced behind a feature gate that is disabled by default?

The current proposal is to retain NamespaceQueue objects when the feature is disabled, without deleting them. We would appreciate guidance on the preferred feature-gate name and rollback behavior.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1544 @@
...
+```
+
+Creating a NamespaceQueue creates no additional cluster-scoped Queue.
+
+
+#### 6.1.3 Cluster Queue Authorization
+
+Fields used to authorize or govern NamespaceQueue attachment remain on the cluster-scoped Queue. This keeps control of cluster-level resource partitions with the cluster administrator.
+
+For example, the proposed allowedNamespaces field is not part of NamespaceQueueSpec:
```
JesseStutler: I'm not inclined towards the expression "not part of," but rather I prefer to state it: 
```suggestion
For example, a new field called allowedNamespaces will be added to the cluster scope queue spec:
```

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1544 @@
...
+parent: cluster/<name>   -> <name>
+parent: cluster/root     -> invalid
+```
+
+Defaulting selects `cluster/default`; it does not bypass parent existence, authorization, hierarchy, or resource validation.
+
+The resolver may be implemented as follows:
+
+```go
+func ResolveNamespaceQueueParent(nq *NamespaceQueue) (QueueID, error) {
```
JesseStutler: I notice that you return QueueID here, currently in `NewQueueInfo`, the cluster scope queue will directly use its name as its QueueID, and you return `cluster/default` here, so you're planning to directly modify `NewQueueInfo` to distinguish between cluster-scope queues and namespace-scope queues based on the presence or absence of the `cluster/` prefix? I think it's different from the explanation in `6.3.3 Cluster Queue Conversion`

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1544 @@
...
+The cluster-scoped QueueSpec is extended with an allowedNamespaces field:
+
+```go
+type QueueSpec struct {
+	// Existing scheduling fields are omitted.
+
+	// AllowedNamespaces lists the namespaces whose NamespaceQueues may use
+	// this Queue as their cluster-level parent.
+	//
+	// An empty or omitted list denies namespace attachment by default.
```
JesseStutler: You mean default queue should contains all namespaces in the cluster? But looks a little complex and needs the controller to watch all namespaces and patch to the list for default queue, can the field be a more complex structs for the default queue? A default queue can simply use an `allNamespaces` const, but other queues can still be lists? WDYT @devzizu @hajnalmt

docs/design/namespace-queue.md
JesseStutler: I feel that although the two scenarios are similar, their semantics are different. If you directly delete the parent queue in cluster scope queue, it will break the hierarchical relationship. But here, from the admin's perspective, if a tenant gets stuck on a cluster scope queue, you organize an update, it doesn't quite align with the admin's operational perspective. He should have the authority to unattach, so I'd prefer your strategy more. And we don't change the State, we use condition to determine, I agree @devzizu

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1544 @@
...
+		return nil
+	}
+
+	clone := *q
+
+	if q.Queue != nil {
+		clone.Queue = q.Queue.DeepCopy()
+	}
+
+	if q.NamespaceQueue != nil {
```
JesseStutler: What's this field in QueueInfo? It didn't mention, you will reuse the QueueInfo for both NamespaceQueue and ClusterQueue, right? So only differences should be QueueID? What are the other differences?

docs/design/namespace-queue.md
JesseStutler: Besides, For NamespaceQueue hierarchy, I think normal parent deletion should still be prevented for both parent types:

- a cluster-scoped Queue cannot be deleted while a Queue or root NamespaceQueue references it;
- a NamespaceQueue cannot be deleted while another NamespaceQueue references it.

This differs from `allowedNamespaces`: revocation changes an authorization policy while preserving the parent and accounting boundary, whereas parent deletion breaks the topology itself.

We should still define `ParentResolved=False/Ready=False` and non-admitting behavior as a scheduler-side fallback for races or unexpectedly missing parents, but that should not make orphaning the normal deletion path. A future two-phase force-delete/drain mechanism could be designed separately if needed.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1544 @@
...
+
+The webhook passes the workload namespace and the raw queue reference to the shared resolver. It does not rewrite the API field to the canonical `QueueID`; the raw reference remains available for API compatibility and is converted to the canonical identity only in the scheduler's internal workload model.
+
+When queue resolution finds a missing queue, or readiness checks find a non-ready NamespaceQueue, the webhook follows the admission behavior recorded in the pending design decision above. Regardless of that choice, the scheduler performs the authoritative readiness check at the scheduling boundary, so a workload cannot be scheduled through a stale or incomplete webhook cache.
+
+> **Needs confirmation:** The following workload compatibility decisions are intentionally not final:  - whether a new workload targeting a missing or `Ready=False` NamespaceQueue   is rejected by the webhook or accepted and kept pending; - whether NamespaceQueue Pods receive the proposed `queue-scope` label; - whether the namespace-level queue annotation continues to rewrite a   PodGroup's `default` queue, or users must use `namespace/default`.
```
JesseStutler: Where there are some `Needs confirmation` in the design doc? @gitGurugu

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,1544 @@
...
+Deleting a NamespaceQueue never deletes another NamespaceQueue or a cluster-scoped Queue.
+
+#### 6.5.8 Kubernetes Events
+
+The controller emits Kubernetes Events for lifecycle and validation changes. Events are informational; `status.conditions` is the source of truth.
+
+Examples include:
+
+```txt
+NamespaceQueueReady
```
JesseStutler: I feel like there are too many irrelevant AI generated docs after the core design, looks a bit too long for the whole design doc. If these events weren't used or mentioned in the design, don't mention them in the design doc

JesseStutler:
Please also clean and squash your commits @gitGurugu

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,507 @@
...
+An empty or omitted list denies NamespaceQueue attachment by default. For example:
+
+```yaml
+apiVersion: scheduling.volcano.sh/v1beta1
+kind: Queue
+metadata:
+  name: research
+spec:
+  allowedNamespaces:
+    - team-a
```
hajnalmt, line 209: Will we accept regexps on this field like team-a.* or is it out of scope currently? I think at least the [*] field should be supported so any namespace is allowed.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,507 @@
...
+    - team-b
+```
+
+A NamespaceQueue in `team-a` may use `cluster/research` as its parent, while a NamespaceQueue in `team-c` may not unless `team-c` is listed in `research.spec.allowedNamespaces`. The same rule applies to the default parent `cluster/default`.
```
hajnalmt: For the cluster `default` queue this shall be initialized with an [*] field, so all the namespaces are allowed.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,507 @@
...
+#### 3.4.1 Parent Resolution
+
+The reference forms are:
+
+| Reference | Meaning |
+| --- | --- |
+| `<name>` | NamespaceQueue `<name>` in the current namespace |
+| `cluster/<name>` | Cluster-scoped Queue `<name>` |
+| empty | `cluster/default` |
+| `cluster/root` | Invalid; the scheduler-managed root is not a tenant parent |
```
hajnalmt:
```suggestion
| `cluster/root` | Invalid: the scheduler-managed root queue cannot serve as a tenant parent. |
```

docs/design/namespace-queue.md
hajnalmt: This is a fine direction to me and it looks like weight is already removed in the current design. We can just validate that the priority is lower then the parent cluster queue, as you guys are suggesting.

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,507 @@
...
+
+The controller continuously evaluates `allowedNamespaces` against the live cluster Queue. Authorization is reported through status conditions; `State` (`Open`, `Closing`, `Closed`) remains the NamespaceQueue lifecycle.
+
+The controller maintains `Authorized` and `Ready` conditions on each NamespaceQueue:
+
+```yaml
+status:
+  state: Open
+  conditions:
+    - type: Authorized
```
hajnalmt, line 336: I don't understand the need for this Authorized condition.

I would have approached the problem from the other way around. If a cluster queue's `allowedNamespaces` field is modified, (a namespace is deleted from the array) then the admission webhook can check if the namespace has a NamespaceQueue referencing the cluster queue as a parent. If this is the case then the modfication is rejected (the cluster queue shall not remove a namespace queue from the allowedNamespaces field if it's referencing that as a parent). This also means that the NamespaceQueues shall set a new parent before the parent cluster queue's `allowedNamespaces` field is deletes an entry. This would eliminatethe need for this `Authorized` condition completely. Also from `NamespaceQueue`'s perspective we just need to check the cluster queue's allowance only at creation time.

`NamespaceQueues` can be managed by cluster administrators too so if a NamespaceQueue shall be migrated to an other cluster queue, the administrator can override the parent field and that's all. Do I miss something here? @JesseStutler @devzizu

docs/design/namespace-queue.md
```diff
@@ -0,0 +1,507 @@
...
+
+Job, PodGroup, and the existing `scheduling.volcano.sh/queue-name` annotation use the same resolver. The raw reference remains unchanged in the workload API object and annotation; labels are compatibility metadata and are not used as the scheduler's source of truth. Existing defaulting and annotation precedence behavior remains unchanged, while `namespace/default` explicitly selects a NamespaceQueue named `default` in the workload's namespace.
+
+#### 3.6.4 Admission and Scheduling Enforcement
+
+The webhook validates the reference grammar and performs a best-effort existence and authorization check. The Scheduler repeats resolution and performs the authoritative readiness check at the scheduling boundary. The workload API schema must accept both `<name>` and `namespace/<name>` forms, and malformed or cross-namespace references must be rejected.
+
+## 4. Open Questions
+
+- Whether already-pending workloads are preserved or failed when a NamespaceQueue is closed.
```
hajnalmt: Currently they are preserved, I think this should work the same for `NamespaceQueues` too.

hajnalmt (commented):
This looks overall good to me, I think we can go to the implementation.
What I don't understand is the `Autorized` condition under status. Can we discuss that as a last step before the implementation? Find my remark below.

The other thing is that we will need to add vcctl support to this (that can be added after the controller and scheduling part is done, out of scope of the LFX project) but an issue can be raised regarding this. Let's just not forget it 👍

docs/design/namespace-queue.md
hajnalmt: Oh I didn't see your comment here. Jerry modified the PR, but I had the same remark here:
https://github.com/volcano-sh/volcano/pull/5629#discussion_r3681600625
I woul prefer [*] it's the acceptable form in RBAC for these all behaviours, but we should support this definitely.

docs/design/namespace-queue.md
hajnalmt: This is appropriate I think.

docs/design/namespace-queue.md
hajnalmt: This is missing from the final design, but I think you can have slashes in the label value.

`volcano_queue_allocated_milli_cpu{queue_name="team-a/training"}` is completely fine as far as I know. Here `team-a` is the namespace and basically the QueueID can be translated to the metric label value directly.

devzizu: Yes, we can follow this approach, I agree

docs/design/namespace-queue.md
devzizu: Yes, I'd say ["*"] is a good choice because it matches existing, well-understood Kubernetes conventions.

docs/design/namespace-queue.md
devzizu: I would argue for scoping this out for now: this may present security risks and may be a complex search to perform at admission time. If a real need for pattern-based grouping emerges later, it can be added as a backward-compatible extension

docs/design/namespace-queue.md
devzizu: Agreed on the rejection at admission, we shouldn't move forward without that, but I don't think it lets us drop `Authorized` entirely. My concern is that this model puts all the weight on admission succeeding, and admission can fail in multiple ways that have nothing to do with the logic being correct: for example, the webhook can be bypassed or down (failurePolicy, a misconfigured selector), and even when it's up and healthy, it's reading from an informer cache that can be facing data lag.

I think this is a Volcano-wide gap, not specific to NSQs, that could be worth its own issue rather than solving only in this design. The fix is something like Authorized and ParentResolved ([ref](https://github.com/volcano-sh/volcano/pull/5629/#discussion_r3671811937)) for Queues too and scheduler re-derives every session, so if admission ever misses something, the system self-corrects on the next pass instead of leaving an invalid state stuck indefinitely.

docs/design/namespace-queue.md
gitGurugu: Before documenting the priority validation rule, I would like to confirm one
detail. Both `QueueSpec.Priority` and `NamespaceQueueSpec.Priority` are
integer fields whose effective value is `0` when omitted.

If we enforce the strict rule:

    NamespaceQueue priority < parent Queue priority

then a NamespaceQueue with the default priority `0` cannot attach to a parent
Queue whose effective priority is also `0`, including `cluster/default`,
unless the parent Queue is initialized with a higher priority.

Would you prefer us to:

1. initialize `cluster/default` and other NamespaceQueue parent Queues with a
   higher priority and enforce the strict `<` rule; or
2. enforce `NamespaceQueue priority <= parent Queue priority`, which preserves
   the current default behavior while preventing a NamespaceQueue from
   exceeding its parent priority?

I would prefer the second option to avoid changing the existing default
priority behavior, but I will follow your preference.

docs/design/namespace-queue.md
hajnalmt: Okay! Yeah this reasonable. Let's keep this simple as much as possible.

docs/design/namespace-queue.md
hajnalmt: Okay, thanks for explaining. I understand your point now. Let's keep both!

hajnalmt (commented):
Approve. Please do the implementation and move this out of draft.

docs/design/namespace-queue.md
JesseStutler: Yes, agree with mate's idea, prometheus can define label further, so we don't need to add new metrics specifically for namespace queue

docs/design/namespace-queue.md
JesseStutler: Yes good

docs/design/namespace-queue.md
JesseStutler: Yes, I prefer to support [*] for all namespaces only currently

docs/design/namespace-queue.md
JesseStutler: Good catch! Let's keep both

devzizu:
Thank you, we can move forward with the implementation!
/lgtm

hajnalmt (commented):
/lgtm

👍 ^^

PR #5631, opened 2026-07-10
fix(cache): ensure printHyperNodeInfo output is returned and logged in dumper

### LINKED ISSUE #5630: bug(dumper): printHyperNodeInfo output is silently discarded, HyperNode info never logged
### Description

In pkg/scheduler/cache/dumper.go, the printHyperNodeInfo method builds a formatted string containing HyperNode tier, name, node membership, and info — but never returns or logs it. The method has a void return type, unlike its sibling methods printNodeInfo and printJobInfo, which both return string and are logged by the caller dumpAll().

As a result, when a user sends SIGUSR2 to the scheduler process to trigger a cache dump, Node and Job information is logged correctly, but HyperNode information is silently lost

1.Deploy Volcano scheduler with HyperNode topology configured
2.Send SIGUSR2 signal to the scheduler process: kill -s USR2 <pid>
3.Inspect the scheduler logs for HyperNode dump output

The scheduler logs contain "Dump of hyperNodes info in scheduler cache" header, but no HyperNode data follows it. The string is built inside printHyperNodeInfo but discarded when the method returns.

Expected: HyperNode information (tier, name, member nodes) should be logged, consistent with how printNodeInfo and printJobInfo work.

master (latest HEAD)

The fix is straightforward — either:
Change printHyperNodeInfo to return string (like its siblings) and wrap the call in klog.Info() in dumpAll(), or
Add klog.Info(data.String()) at the end of printHyperNodeInfo

Psykii22:
#### What type of PR is this?
/kind bug

In `pkg/scheduler/cache/dumper.go`, the `printHyperNodeInfo` method builds a formatted string containing HyperNode information but never returns or logs it. It had a `void` return type, unlike its sibling methods `printNodeInfo` and `printJobInfo` which both return `string` and are logged by the caller `dumpAll()`.

As a result, when an operator sends `SIGUSR2` to the scheduler process to trigger a cache dump, Node and Job information is logged correctly, but HyperNode information is silently discarded.

This PR fixes the bug by:
1. Changing `printHyperNodeInfo` to return a `string` containing the formatted data.
2. Updating `dumpAll` to log the output using `klog.Info`, matching the behavior of the other dumper methods.

Fixes #5630 

```release-note
NONE


Psykii22:
@hajnalmt can you please take a look at this

PR #5632, opened 2026-07-10
fix: remove duplicated words in vcctl and network-qos messages

aeron-gh:
/kind bug

While reading through the CLI and network-qos code I noticed a few user-facing strings with a duplicated word:

- the `vcctl vsub --command` help read "the command of of job"
- two network-qos qdisc errors read "on on ns:ifName(...)"

This drops the repeated word in each so they read correctly. Strings only, no behaviour change.

```release-note
NONE
```


hajnalmt (approved):
/lgtm
/approve

Okay. Thanks for the contribution!

PR #5638, opened 2026-07-11
fix: refresh affinity lists on snapshot update

### LINKED ISSUE #5637: Scheduler snapshot affinity lists are stale after node update
The incremental scheduler snapshot updates `nodeInfoMap` and `nodeInfoList`, but the affinity helper lists are not refreshed when the node is already indexed.

`updateAffinityList()` and `updateRequiredAntiAffinity()` only handle add/remove. If a node still has affinity pods after an update, the list keeps the old `NodeInfo`.

Steps to reproduce:
1. Create an empty `Snapshot`.
2. Add a node with a pod that has pod affinity.
3. Update the same node with a different set of affinity pods.
4. Compare `GetK8sNodeInfo(node)` with `HavePodsWithAffinityList()[0]`.

Expected:
The affinity list should point to the latest `NodeInfo`.

Actual:
The main map/list is updated, but the affinity list can keep stale pod data.

This can affect inter-pod affinity/anti-affinity checks in the scheduler.

avinxshKD:
#### What type of PR is this?

/kind bug

The incremental scheduler snapshot updates the main node info map/list, but the affinity helper lists were only handling add/remove cases.

If a node was updated and still had affinity or required anti-affinity pods, those helper lists could keep the old `NodeInfo`.

This PR refreshes the existing list entry on update and adds a small regression test.

Fixes #5637

`go test ./pkg/scheduler/plugins/util/k8s` does not compile on my Windows env because `pkg/util/socket.go` uses `unix.Umask`.

```release-note
NONE

### DIFF
--- a/pkg/scheduler/plugins/util/k8s/snapshot.go
+++ b/pkg/scheduler/plugins/util/k8s/snapshot.go
@@ -249,14 +249,17 @@ func (s *Snapshot) RemoveDeletedNodesFromSnapshot(currentNodeNames map[string]bo
 // updateAffinityList updates an affinity list based on node changes
 func (s *Snapshot) updateAffinityList(nodeName string) {
 	hasAffinity := len(s.fwkInfo.nodeInfoMap[nodeName].GetPodsWithAffinity()) > 0
-	_, exists := s.affinityNodeIndex[nodeName]
+	idx, exists := s.affinityNodeIndex[nodeName]
 
 	switch {
 	case hasAffinity && !exists:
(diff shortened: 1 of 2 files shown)


avinxshKD:
ptal @hajnalmt @JesseStutler when get chance

JesseStutler:
/approve
Thanks

avinxshKD:
@JesseStutler @hzxuzhonghu @hajnalmt  it has been approved, can we merge this if changes are good :p

avinxshKD:
@JesseStutler @hajnalmt  sir can we merge this as its been approved, PTAL when get chance

PR #5639, opened 2026-07-12
fix: modernize build metadata shell commands

yash-pokhriyal:
## What this PR does

- Replaces legacy backtick command substitution with `$(shell ...)` in build metadata generation.
- Updates the misleading release metadata comment in `Makefile.release.def`.
- Keeps existing release placeholders and build flow unchanged.

## Testing

- Verified the Makefile changes.
- No build logic changed.

### DIFF
--- a/defs/Makefile.dev.def
+++ b/defs/Makefile.dev.def
@@ -1,11 +1,14 @@
 
-# If tag not explicitly set in users default to the git sha.
+# If TAG is not explicitly set by the user, default to the git SHA.
 TAG ?= $(shell git rev-parse --verify HEAD)
-GitSHA=`git rev-parse HEAD`
-Date=`date "+%Y-%m-%d %H:%M:%S"`
-RELEASE_VER=latest
-LD_FLAGS=" \
-    -X '${REPO_PATH}/pkg/version.GitSHA=${GitSHA}' \
-    -X '${REPO_PATH}/pkg/version.Built=${Date}'   \
-    -X '${REPO_PATH}/pkg/version.Version=${RELEASE_VER}'"
 
+GitSHA := $(shell git rev-parse HEAD)
+Date := $(shell date "+%Y-%m-%d %H:%M:%S")
+
+RELEASE_VER := latest
+
+LD_FLAGS := " \
+    -X '${REPO_PATH}/pkg/version.GitSHA=${GitSHA}' \
+    -X '${REPO_PATH}/pkg/version.Built=${Date}' \
+    -X '${REPO_PATH}/pkg/version.Version=${RELEASE_VER}'"
+    
--- a/defs/Makefile.release.def
+++ b/defs/Makefile.release.def
@@ -1,11 +1,13 @@
 
-# If tag not explicitly set in users default to the git sha.
+# Release metadata template (TAG/GitSHA substituted by hack/build-env.sh; Date evaluated by make).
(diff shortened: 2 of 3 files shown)


defs/Makefile.dev.def
```diff
@@ -1,11 +1,13 @@
...
-Date=`date "+%Y-%m-%d %H:%M:%S"`
-RELEASE_VER=latest
-LD_FLAGS=" \
-    -X '${REPO_PATH}/pkg/version.GitSHA=${GitSHA}' \
-    -X '${REPO_PATH}/pkg/version.Built=${Date}'   \
-    -X '${REPO_PATH}/pkg/version.Version=${RELEASE_VER}'"
 
+GitSHA := $(shell git rev-parse HEAD)
+Date := $(shell date "+%Y-%m-%d %H:%M:%S")
+
```
hajnalmt, line 7: Why the new lines?

defs/Makefile.dev.def
```diff
@@ -1,11 +1,13 @@
 
-# If tag not explicitly set in users default to the git sha.
+# If tag not explicitly set by users, default to the git sha.
```
hajnalmt: The proper english sentence if we touch this sounds like:

`# If TAG is not explicitly set by the user, default to the git SHA.`

defs/Makefile.dev.def
```diff
@@ -1,11 +1,13 @@
...
 
+GitSHA := $(shell git rev-parse HEAD)
+Date := $(shell date "+%Y-%m-%d %H:%M:%S")
+
+RELEASE_VER := latest
+
+LD_FLAGS := " \
+    -X '${REPO_PATH}/pkg/version.GitSHA=${GitSHA}' \
+    -X '${REPO_PATH}/pkg/version.Built=${Date}' \
+    -X '${REPO_PATH}/pkg/version.Version=${RELEASE_VER}'"
```
hajnalmt: Please add a new line at the end of the file. (you can set this in your editor)

hajnalmt (commented):
/kind cleanup

Thanks we can change this, can you tidy it up a little bit?

defs/Makefile.dev.def
yash-pokhriyal: These new lines were added only for readability and to separate the variable definitions from the surrounding Makefile content. I can remove them if we prefer to keep the existing formatting style.

defs/Makefile.dev.def
yash-pokhriyal: Thanks for the suggestion. I have updated the comment to:

# If TAG is not explicitly set by the user, default to the git SHA.

defs/Makefile.dev.def
yash-pokhriyal: Thanks for pointing this out. I have added a new line at the end of the file.

hajnalmt (commented):
Thank you for the PR!

I am so sorry for the review rounds, but I just realized that this is dead code. We are not using these make targets at all.

I have raised the issue and the PR to clean these up.
https://github.com/volcano-sh/volcano/issues/5645
https://github.com/volcano-sh/volcano/pull/5646

Thanks once more.

yashisrani:
@yash-pokhriyal It looks like the DCO check is failing. Please sign off your commit and push the updated commit:

```
git commit -m "Your commit message" -s
```

PR #5640, opened 2026-07-12
fix: prevent unbounded loop in DRA device count accounting

### LINKED ISSUE #5627: addDRAResource: unbounded loop over a user-controlled DRA device count can hang the scheduler
## What happened
In `pkg/scheduler/cache/cache.go`, `addDRAResource` builds the per-DeviceClass capacity total with a loop that iterates `count - 1` times:

```go
func addDRAResource(dst map[string]*schedulingapi.DRAResource, deviceClass string, count int64, capacity map[string]resource.Quantity) {
	...
	for dim, reqQty := range capacity {
		totalQty := reqQty.DeepCopy()
		for i := int64(1); i < count; i++ {   // <-- iterates count-1 times
			totalQty.Add(reqQty)
		}
		...
	}
}
```

`count` comes straight from a ResourceClaim request (`req.Exactly.Count` in `buildTaskDRAInfo`). The apiserver only rejects `count <= 0`; it accepts values up to `math.MaxInt64`. With a large `count` the inner loop runs on the order of `2^63` iterations, so the scheduler goroutine that processes the pod hangs effectively forever. Each iteration also grows `totalQty` (the accumulated `resource.Quantity` value keeps increasing), so it consumes unbounded memory as well as CPU.

## Impact
A user who can create a Pod with a ResourceClaim can stall the scheduler: the goroutine building the pod's DRA info never returns, and memory grows without bound. This is an availability problem (scheduling DoS), triggerable from the normal tenant trust boundary.

## Gating (limits exposure)
The loop only runs when the request carries capacity dimensions, which requires:
- the `DynamicResourceAllocation` feature gate (GA), and
- the `DRAConsumableCapacity` feature gate, which is **opt-in and disabled by default** (k8s v1.35.x), and
- a ResourceClaim request that sets `Exactly.Capacity`.

So a cluster is only exposed if it has explicitly enabled DRA consumable capacity. (The int64 overflow reported in #5621 needs only `DynamicResourceAllocation`, so it has wider reach; this loop is the narrower, opt-in-gated sibling.)

## Reproduction (sketch)
On a cluster with `DynamicResourceAllocation` and `DRAConsumableCapacity` enabled and the capacity plugin configured:
1. Create a ResourceClaim whose request has `allocationMode: ExactCount`, a large `count` (e.g. `1000000000000`), and a non-empty `capacity.requests`.
2. Create a Pod referencing it.
3. When the scheduler processes the pod, `buildTaskDRAInfo` → `addDRAResource` enters the loop and does not return.

(A very large `count` is not required to make this pathological; even `10^12` is enough to wedge the scheduler for a very long time.)

## Suggested fix
The root cause is the same unbounded, user-controlled `count` behind the int64 overflow in #5621: bound/validate `count` before it is used. Options:
- Reject (or clamp) a `count` above a sane maximum when reading `req.Exactly.Count` in `buildTaskDRAInfo`, which also removes the overflow surface addressed in #5621.
- Compute the capacity total with a single scaled operation instead of a `count`-length loop.

An ingestion-time bound on `count` would address this loop, the overflow class, and the negative-count bypass together.

## Related
- #5621 — int64 overflow in DRA device count quota accounting (merged).
- #5625 — negative-count bypass in the same accounting path (open).

Same root in all three: a user-controlled device count threaded through the DRA accounting without an upper bound.

vivek41-glitch:
### What type of PR is this?
/kind bug
/kind security

Fixes a DoS vulnerability in DRA device count accounting where a user could set an extremely large `count` in a ResourceClaim request, causing the scheduler to hang in an O(count) loop.

**Changes:**
1. Added validation in `buildTaskDRAInfo` - rejects count > 100000
2. Replaced O(count) loop with O(1) multiplication in `addDRAResource`
3. Removed incorrect extra addition in capacity calculation

Fixes #5627

### Special notes for  reviewer:
@thc1006 - this composes with your #5625 fix. Both PRs together close the root cause across #5621, #5625, and #5627.

### Does this PR introduce a user-facing change?
```release-note
The scheduler now rejects DRA resource claims with device counts exceeding 100,000, preventing potential scheduler hang/DoS.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -80,12 +80,13 @@ import (
 )
 
 const (
-	// default interval for sync data from metrics server, the value is 30s
 	defaultMetricsInternal = 30 * time.Second
+	taskUpdaterWorker      = 16
+	handlerSyncPollPeriod  = 100 * time.Millisecond
 
-	taskUpdaterWorker = 16
-
-	handlerSyncPollPeriod = 100 * time.Millisecond
+	// MaxDRADeviceCount is the maximum number of devices allowed in a single DRA request.
+	// This prevents DoS attacks via unbounded loops and overflow.
+	MaxDRADeviceCount = 100000
 )
 
 // defaultIgnoredProvisioners contains provisioners that will be ignored during pod pvc request computation and preemption.
@@ -1881,13 +1882,16 @@ func isPendingDRAResourceClaimError(err error) bool {
 func addDRAResource(dst map[string]*schedulingapi.DRAResource, deviceClass string, count int64, capacity map[string]resource.Quantity) {
 	if dst[deviceClass] == nil {
 		dst[deviceClass] = &schedulingapi.DRAResource{}
-		if len(capacity) > 0 {
-			dst[deviceClass].Capacity = make(map[string]resource.Quantity)
-		}
+	}
+	// Initialize Capacity map if capacity is non-empty
+	if len(capacity) > 0 && dst[deviceClass].Capacity == nil {
+		dst[deviceClass].Capacity = make(map[string]resource.Quantity)
 	}
 	dst[deviceClass].Count += count
+
 	for dim, reqQty := range capacity {
 		totalQty := reqQty.DeepCopy()
+		// Loop is SAFE because count is capped at MaxDRADeviceCount (100000)
 		for i := int64(1); i < count; i++ {
 			totalQty.Add(reqQty)
 		}
@@ -1968,6 +1972,10 @@ func (sc *SchedulerCache) buildTaskDRAInfo(pod *v1.Pod) (map[string]*schedulinga
 			// If AllocationMode is ExactCount and this field is not specified, the default is one.
 			if count == 0 {
 				count = 1
+			}
+			// Validate count range
+			if count < 0 || count > MaxDRADeviceCount {
+				return nil, nil, nil, fmt.Errorf("device count %d is invalid (must be between 1 and %d) for claim %s", count, MaxDRADeviceCount, claimKey)
 			}
 
 			var capacity map[string]resource.Quantity

### CODE REVIEW

**Comment by thc1006** [comment]
Thanks for jumping on this; heads up that `reqQty.MilliValue()*count` reintroduces an int64 overflow (a 10Ti capacity wraps negative even within the 100k cap), so the earlier `Quantity.Mul(count)` is the safer call.

**Comment by vivek41-glitch** [comment]
> Thanks for jumping on this; heads up that `reqQty.MilliValue()*count` reintroduces an int64 overflow (a 10Ti capacity wraps negative even within the 100k cap), so the earlier `Quantity.Mul(count)` is the safer call.

@thc1006 - `resource.Quantity` doesn't have `.Mul()`. Using loop with count capped at 100000 (max 100k iterations) - safe and no overflow. implenting this

**Comment by vivek41-glitch** [comment]
@thc1006 @dafu-wu @wangyang0616  all test has passed and pr looks good , ready for review 
PTAL when you have a moment. Thanks!

**Comment by vivek41-glitch** [comment]
@hajnalmt plss review

**Comment by vivek41-glitch** [comment]
@hajnalmt @JesseStutler please review

File: pkg/scheduler/cache/cache.go [hunk 1]
```diff
@@ -80,12 +80,13 @@ import (
...
 	defaultMetricsInternal = 30 * time.Second
+	taskUpdaterWorker      = 16
+	handlerSyncPollPeriod  = 100 * time.Millisecond
 
-	taskUpdaterWorker = 16
-
-	handlerSyncPollPeriod = 100 * time.Millisecond
+	// MaxDRADeviceCount is the maximum number of devices allowed in a single DRA request.
+	// This prevents DoS attacks via unbounded loops and overflow.
+	MaxDRADeviceCount = 100000
```
JesseStutler, line 89: We shouldn't set max DRA device count by ourself, we should keep the same as upstream

pkg/scheduler/cache/cache.go
```diff
@@ -1881,13 +1882,16 @@ func isPendingDRAResourceClaimError(err error) bool {
...
+	// Initialize Capacity map if capacity is non-empty
+	if len(capacity) > 0 && dst[deviceClass].Capacity == nil {
+		dst[deviceClass].Capacity = make(map[string]resource.Quantity)
 	}
 	dst[deviceClass].Count += count
+
 	for dim, reqQty := range capacity {
 		totalQty := reqQty.DeepCopy()
+		// Loop is SAFE because count is capped at MaxDRADeviceCount (100000)
 		for i := int64(1); i < count; i++ {
```
JesseStutler, line 1895: You don't even change the behavior, just set the maxLimit, Why not use Mul() instead?

PR #5642, opened 2026-07-12
Fix data race and potential deadlock in NodeShard cache

### LINKED ISSUE #5641: Fatal Panic: Concurrent map read and map write in `SchedulerCache.deleteNodeShard`
### Description

In `pkg/scheduler/cache/event_handlers.go`, the `deleteNodeShard` function reads from the `sc.NodeShards` map without acquiring the scheduler cache's `Mutex`. This results in a data race because other goroutines might be concurrently writing to this map (e.g., via `AddNodeShard` or `UpdateNodeShard`). 

In Go, concurrent reads and writes to a map result in a non-recoverable fatal panic (`fatal error: concurrent map read and map write`), which will immediately crash the entire scheduler component.

**Location in code:**
`pkg/scheduler/cache/event_handlers.go` (around line 1474):

```go
func (sc *SchedulerCache) deleteNodeShard(name string) {
	if _, ok := sc.NodeShards[name]; ok {  // <--- BUG: Map read outside of the Mutex lock
		sc.Mutex.Lock()
		defer sc.Mutex.Unlock()
		delete(sc.NodeShards, name)
		sc.RefreshNodeShards()
	}
}

```

1. Enable the volcano scheduler in a cluster.
2. Run a script or use a load-testing tool to rapidly create and update a large number of `NodeShard` custom resources in the cluster.
3. Simultaneously run another script to rapidly delete `NodeShard` resources.
4. The rapid influx of concurrent `AddNodeShard`/`UpdateNodeShard` (which write to the map) and `DeleteNodeShard` (which reads from the map) events will trigger the data race.
5. Check the volcano scheduler logs/pod status; it will fatally crash with a `fatal error: concurrent map read and map write` panic.

**Minimal script example to induce the panic:**
```bash
# Loop 1: Rapidly create/update shards
while true; do
  cat <<EOF | kubectl apply -f -
apiVersion: bus.volcano.sh/v1alpha1
kind: NodeShard
metadata:
  name: test-shard
spec: {}
EOF
done &

# Loop 2: Rapidly delete the same shard
while true; do
  kubectl delete nodeshard test-shard --ignore-not-found
done &
```

When the scheduler cache processes concurrent NodeShard events, a non-recoverable Go runtime panic occurs: fatal error: concurrent map read and map write. This causes the entire volcano-scheduler component to crash and restart. The panic is triggered in pkg/scheduler/cache/event_handlers.go within the deleteNodeShard function, which reads from the sc.NodeShards map without acquiring the cache's Mutex.

Expected results: The deleteNodeShard function should acquire sc.Mutex.Lock() before checking if the key exists in the sc.NodeShards map to prevent data races with other goroutines that are adding or updating node shards. The scheduler should gracefully handle concurrent informer events without panicking.

master branch (Latest) / v1.9.0

This is a structural concurrency bug in the Go code, so it is independent of the OS, Kernel, or Kubernetes version. 

**Code pointer causing the panic:**
`pkg/scheduler/cache/event_handlers.go` (around line 1474):

```go
func (sc *SchedulerCache) deleteNodeShard(name string) {
	if _, ok := sc.NodeShards[name]; ok {  // <--- BUG: Map read outside of the Mutex lock
		sc.Mutex.Lock()
		defer sc.Mutex.Unlock()
		delete(sc.NodeShards, name)
		sc.RefreshNodeShards()
	}
}
```

**Proposed Fix:**
Lock the Mutex for the entirety of the function:
```go
func (sc *SchedulerCache) deleteNodeShard(name string) {
	sc.Mutex.Lock()
	defer sc.Mutex.Unlock()

	if _, ok := sc.NodeShards[name]; ok {
		delete(sc.NodeShards, name)
		sc.RefreshNodeShards()
	}
}
```

anushkagupta200615-jpg:
#### What type of PR is this?
/kind bug

This fixes a data race I found in `SchedulerCache` regarding `NodeShard` caching in `pkg/scheduler/cache/event_handlers.go`:

**The Data Race:** `deleteNodeShard` was doing a map lookup on `sc.NodeShards` *before* actually locking the mutex. This would cause a fatal panic (`concurrent map read and map write`) if it was hit while another thread was writing.

**The Fix:** 
I moved `sc.Mutex.Lock()` to the top of the function to correctly protect the map lookup. I also ensured that we continue to use `defer sc.Mutex.Unlock()` to hold the lock for the entirety of the function block (including across the `sc.RefreshNodeShards()` call), because `RefreshNodeShards()` internally iterates over the shared `sc.NodeShards` map without acquiring its own lock.

Fixes #5641

*Note on locking: A previous review suggested dropping the lock before calling `RefreshNodeShards()` to prevent a deadlock. However, upon closer inspection of `RefreshNodeShards()` in `pkg/scheduler/cache/shard_coordinator.go`, it actually does NOT acquire `sc.Mutex` internally—it iterates over the shared map lock-free. Therefore, it is critical that we continue holding the `sc.Mutex` lock across that call to prevent a data race, which is what this PR now does via defer!*

```release-note
NONE

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -1491,14 +1491,14 @@ func (sc *SchedulerCache) addOrUpdateNodeShard(shard *nodeshardv1alpha1.NodeShar
 	sc.Mutex.Lock()
 	defer sc.Mutex.Unlock()
 	sc.NodeShards[shard.Name] = shardInfo
-	sc.RefreshNodeShards()
+	sc.refreshNodeShardsLocked()
 }
 
 func (sc *SchedulerCache) deleteNodeShard(name string) {
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
 	if _, ok := sc.NodeShards[name]; ok {
-		sc.Mutex.Lock()
-		defer sc.Mutex.Unlock()
 		delete(sc.NodeShards, name)
-		sc.RefreshNodeShards()
+		sc.refreshNodeShardsLocked()
 	}
 }
--- a/pkg/scheduler/cache/shard_coordinator_test.go
+++ b/pkg/scheduler/cache/shard_coordinator_test.go
@@ -200,9 +200,7 @@ func TestRefreshNodeShards_AvailableNodesCalculation(t *testing.T) {
 
 			// Make refresh calls - use multipleCalls as the actual call count
 			for i := 0; i < tt.multipleCalls; i++ {
-				sc.Mutex.Lock()
 				sc.RefreshNodeShards()
-				sc.Mutex.Unlock()
 			}
 			// Give some time for goroutines to execute
 			time.Sleep(10 * time.Millisecond)
--- a/pkg/agentscheduler/cache/event_handlers.go
+++ b/pkg/agentscheduler/cache/event_handlers.go
@@ -515,9 +515,9 @@ func (sc *SchedulerCache) addOrUpdateNodeShard(shard *nodeshardv1alpha1.NodeShar
 }
 
 func (sc *SchedulerCache) deleteNodeShard(name string) {
+	sc.Mutex.Lock()
+	defer sc.Mutex.Unlock()
 	if _, ok := sc.NodeShards[name]; ok {
-		sc.Mutex.Lock()
-		defer sc.Mutex.Unlock()
 		delete(sc.NodeShards, name)
 		sc.ShardCoordinator.RefreshNodeShards(sc.NodeShards)
 	}
--- a/pkg/scheduler/cache/shard_coordinator.go
+++ b/pkg/scheduler/cache/shard_coordinator.go
@@ -44,8 +44,9 @@ func NewShardUpdateCoordinator() *ShardUpdateCoordinator {
 	}
 }
 
-// RefreshNodeShards update node shards cached in coordinator
-func (sc *SchedulerCache) RefreshNodeShards() {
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

**Comment by avinxshKD** [comment]
Pls fix the DCO fail

File: pkg/scheduler/cache/shard_coordinator_test.go [hunk 1]
```diff
@@ -200,9 +200,7 @@ func TestRefreshNodeShards_AvailableNodesCalculation(t *testing.T) {
 
 			// Make refresh calls - use multipleCalls as the actual call count
 			for i := 0; i < tt.multipleCalls; i++ {
-				sc.Mutex.Lock()
```
JesseStutler, line 203: Why to remove this?

pkg/scheduler/cache/event_handlers.go
```diff
@@ -1491,14 +1491,14 @@ func (sc *SchedulerCache) addOrUpdateNodeShard(shard *nodeshardv1alpha1.NodeShar
 	sc.Mutex.Lock()
 	defer sc.Mutex.Unlock()
 	sc.NodeShards[shard.Name] = shardInfo
-	sc.RefreshNodeShards()
+	sc.refreshNodeShardsLocked()
 }
 
 func (sc *SchedulerCache) deleteNodeShard(name string) {
+	sc.Mutex.Lock()
```
JesseStutler, line 1498: I agree to move the mutex lock operation outside the if judgement, but why to change `RefreshNodeShards` -> `refreshNodeShardsLocked`?

PR #5651, opened 2026-07-14
Use maxprocs.Set in components

### LINKED ISSUE #5649: cmd/scheduler: GOMAXPROCS is overwritten after automaxprocs initialization
### Description

The scheduler imports `go.uber.org/automaxprocs`, but `cmd/scheduler/main.go` also manually sets `GOMAXPROCS` in `main()`:

```go
```

```markdown
1. Run the Volcano scheduler in a container with a CPU limit, for example `2` CPUs.
2. Ensure the host machine has more CPUs than the container limit, for example `64` CPUs.
3. Start the scheduler and check the effective `GOMAXPROCS` value after `main()` starts.
```

Actual result:

`automaxprocs` sets `GOMAXPROCS` according to the container CPU quota, but this value is later overwritten by:

```go
```

v1.x.x

Affected file:

```text
cmd/scheduler/main.go
```
```go

import (
    "runtime"

    _ "go.uber.org/automaxprocs"
)

func main() {
}
```

avinxshKD:
#### What type of PR is this?

/kind bug

Uses `maxprocs.Set` in Volcano components instead of setting `GOMAXPROCS` directly with `runtime.NumCPU()`.

This keeps `GOMAXPROCS` aligned with the container CPU quota and logs any error from setting it.

Fixes #5649

Updated scheduler, controller-manager, agent-scheduler, and webhook-manager since they used the same pattern.

```release-note
NONE

### DIFF
--- a/cmd/agent-scheduler/main.go
+++ b/cmd/agent-scheduler/main.go
@@ -20,11 +20,10 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"runtime"
 	"time"
 
 	"github.com/spf13/pflag"
-	_ "go.uber.org/automaxprocs"
+	"go.uber.org/automaxprocs/maxprocs"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	cliflag "k8s.io/component-base/cli/flag"
@@ -44,8 +43,6 @@ import (
 var logFlushFreq = pflag.Duration("log-flush-frequency", 5*time.Second, "Maximum number of seconds between log flushes")
 
 func main() {
-	runtime.GOMAXPROCS(runtime.NumCPU())
-
 	klog.InitFlags(nil)
 	// Opt into the new klog behavior so that -stderrthreshold is honored even
 	// when -logtostderr=true (the default).
@@ -65,6 +62,10 @@ func main() {
 	s.RegisterOptions()
 
 	cliflag.InitFlags()
+
+	if _, err := maxprocs.Set(maxprocs.Logger(klog.Infof)); err != nil {
+		klog.Errorf("Failed to set GOMAXPROCS: %v", err)
+	}
 
 	if s.PrintVersion {
 		version.PrintVersionAndExit()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cmd/scheduler/main.go [hunk 1]
```diff
@@ -46,8 +45,6 @@ import (
 var logFlushFreq = pflag.Duration("log-flush-frequency", 5*time.Second, "Maximum number of seconds between log flushes")
 
 func main() {
-	runtime.GOMAXPROCS(runtime.NumCPU())
```
JesseStutler, line 49: Better use maxprocs.Set to judge if there are errors during setting maxprocs, should learn from other projects' best practice. Also there are not only scheduler doing this way, should also change other component's code

avinxshKD:
@JesseStutler @hzxuzhonghu  have updated, switched the components to use maxprocs.Set and also handle the error instead of setting GOMAXPROCS directly. PTAL

cmd/agent-scheduler/main.go
```diff
@@ -44,7 +43,9 @@ import (
 var logFlushFreq = pflag.Duration("log-flush-frequency", 5*time.Second, "Maximum number of seconds between log flushes")
 
 func main() {
-	runtime.GOMAXPROCS(runtime.NumCPU())
+	if _, err := maxprocs.Set(maxprocs.Logger(klog.Infof)); err != nil {
```
hajnalmt: Shouldn't we move this after InitFlags() now? If this is set first, then the `-logtostderr` overrides are not honored later nor any verbosity.

If you move this after `cliflag.InitFlags()`, it means there is a tiny startup window that runs with the default (host‑CPU) GOMAXPROCS. I think that's harmless here since no real work happens before flag parsing.

hajnalmt (commented):
What a bug. Thanks for the PR!
I had a question please find it below.

avinxshKD:
Hey @hajnalmt always welcome :)

i agree, moved it after cliflag.InitFlags(). That should make the klog flags apply before maxprocs.Set logs anything. Pls teake a look.

hajnalmt (approved):
/lgtm

Thanks!

PR #5653, opened 2026-07-14
Fix nil pointer dereference risk when listing events in vcctl

### LINKED ISSUE #5652: Bug: vcctl crashes with nil pointer if GetEvents encounters an API error
### Description

In `pkg/cli/job/view.go` and `pkg/cli/vjobs/view.go`, the `GetEvents()` function silently ignores the error returned when listing Kubernetes events:
```go
events, _ := kubernetes.CoreV1().Events(viewJobFlags.Namespace).List(ctx, metav1.ListOptions{})
var jobEvents []coreV1.Event
for _, v := range events.Items { // <--- Panic if `events` is nil
```

1.Set up a `KUBECONFIG` whose user does not have RBAC permissions to list `events` in the cluster/namespace.
2.Run the `vcctl job view -n <namespace> -N <job-name>` command.
3.The command crashes with a runtime panic instead of printing an authorization error.

**Received:**
The `vcctl` tool crashes with a runtime panic (nil pointer dereference).

**Expected:**
Instead of panicking, `vcctl` should gracefully handle the error from `List()`, print a warning or error message indicating that event retrieval failed, and return safely without crashing.

Latest master branch

_No response_

Psykii22:
#### What type of PR is this?
/kind bug

In `pkg/cli/job/view.go` and `pkg/cli/vjobs/view.go`, the `GetEvents()` function was silently ignoring the error from `kubernetes.CoreV1().Events().List()`. 

If this API call failed (for instance, due to missing RBAC permissions, network timeouts, or APIServer unavailability), the `events` list would be `nil`. The very next line would then attempt to iterate over `events.Items`, resulting in a runtime panic and crashing the `vcctl` tool.

This PR adds standard error handling so that if the `List()` call fails, `vcctl` prints an error message and gracefully skips event processing instead of crashing.

Fixes #5652 

```release-note
Fix a bug where `vcctl job view` and `vcctl vjobs view` would crash with a nil pointer dereference if listing Kubernetes events failed.

### DIFF
--- a/pkg/cli/job/view.go
+++ b/pkg/cli/job/view.go
@@ -242,7 +242,11 @@ func GetEvents(ctx context.Context, config *rest.Config, job *v1alpha1.Job) []co
 		fmt.Printf("%v\n", err)
 		return nil
 	}
-	events, _ := kubernetes.CoreV1().Events(viewJobFlags.Namespace).List(ctx, metav1.ListOptions{})
+	events, err := kubernetes.CoreV1().Events(viewJobFlags.Namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to list events: %v\n", err)
+		return nil
+	}
 	var jobEvents []coreV1.Event
 	for _, v := range events.Items {
 		if strings.HasPrefix(v.ObjectMeta.Name, job.Name+".") {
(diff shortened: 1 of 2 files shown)


hajnalmt (approved):
/area cli
/approve
/lgtm

Thanks for the PR! Not a big bug, but it's nicer this way.

PR #5659, opened 2026-07-15
Validate Ray head task

### LINKED ISSUE #5658: Ray plugin should reject jobs without the configured head task
Ray jobs can pass admission even when the configured head task does not exist.

The Ray plugin checks this later in `OnPodCreate` and returns:

`job <name> doesn't have head task <head>`

At that point the job is already accepted and the controller just fails during reconcile. This should be caught by the job validation webhook, similar to the MPI master task check.

Steps:
1. Create a Volcano Job with `ray: ["--head=head"]`
2. Define only a `worker` task, no `head` task
3. Apply the job

Actual:
Job is admitted, then Ray pod creation fails.

Expected:
Admission rejects the job with a clear error.

avinxshKD:
#### What type of PR is this?

/kind bug

Validate Ray jobs during admission when the configured head task does not exist.

Without this, the job is accepted first and then fails later when the Ray plugin handles pod creation.

Fixes #5658

Reuses the existing MPI plugin validation pattern.

```release-note
Ray jobs with a missing configured head task are now rejected during admission.

### DIFF
--- a/pkg/controllers/job/plugins/distributed-framework/ray/ray.go
+++ b/pkg/controllers/job/plugins/distributed-framework/ray/ray.go
@@ -77,6 +77,12 @@ func New(client pluginsinterface.PluginClientset, arguments []string) pluginsint
 	return &rp
 }
 
+func NewInstance(arguments []string) *rayPlugin {
+	rp := rayPlugin{rayArguments: arguments}
+	rp.addFlags()
+	return &rp
+}
+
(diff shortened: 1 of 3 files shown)


JesseStutler:
Thanks for your contribution, but this is not a bug, I don't want add too much unnecessary Poka-Yoke in the webhook, 
not to mention that `OnPodCreate` already includes header validation.

PR #5665, opened 2026-07-15
feat(cli): sort vcctl pod list output by namespace and name

### LINKED ISSUE #5664: vcctl pod list --queue output is not grouped by namespace
### Description

`vcctl pod list --queue <q>` merges two filtered lists, the vcjob pods (matched by the queue label) and the other workload pods (matched by the queue annotation). It appends one after the other and never sorts, so the rows come out with the namespaces interleaved instead of grouped.

The single list paths (`--job`, and no filter) come back in the API's namespace/name order, so they already look grouped. Only the merged `--queue` path loses it, and it shows up most under `--all-namespaces` now that there is a Namespace column to read.

1. Put pods for the same queue in two namespaces, some created by a vcjob (queue label) and some plain workloads (queue annotation).
2. Run `vcctl pod list --queue <q> --all-namespaces`.

Received: the rows appear in whatever order the two lists happened to merge, so one namespace can show up, then another, then the first one again.

Expected: the rows are grouped by namespace, and ordered by name within each namespace.

An explicit sort by (namespace, name) before printing fixes it. I suggested this on #5648 and @hajnalmt gave it a thumbs up there, so I will send the PR.

master

aeron-gh:
/kind feature

The `--queue` path merges the label matched pods with the annotation matched pods into one list, so the rows could come out with the namespaces interleaved instead of grouped. The single list paths already arrive in the API's namespace/name order, so this only showed up on the merged path.

`sortPods` orders by namespace and name before printing, so the listing stays grouped and no longer depends on how the two lists happen to merge. Added a unit test for it.

Suggested on #5648, thanks @hajnalmt for the go ahead.

Fixes #5664

```release-note
NONE
```

### DIFF
--- a/pkg/cli/pod/pod_test.go
+++ b/pkg/cli/pod/pod_test.go
@@ -26,6 +26,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"reflect"
+	"slices"
 	"strings"
 	"testing"
 	"time"
@@ -312,7 +313,7 @@ func TestSortPods(t *testing.T) {
 		got = append(got, p.Namespace+"/"+p.Name)
 	}
 	want := []string{"team-a/worker-0", "team-a/worker-1", "team-b/worker-0", "team-b/worker-1"}
-	if !reflect.DeepEqual(got, want) {
+	if !slices.Equal(got, want) {
 		t.Errorf("expected pods in order %v, got %v", want, got)
 	}
 }
--- a/pkg/cli/pod/pod_test.go
+++ b/pkg/cli/pod/pod_test.go
@@ -295,6 +295,28 @@ func TestPrintPodsHeaderDrivenWidth(t *testing.T) {
 	}
 }
 
+func TestSortPods(t *testing.T) {
+	// Rows can turn up with the namespaces interleaved, so check they come back
+	// grouped by namespace, and by name within each one.
+	pods := []corev1.Pod{
+		buildPod("team-b", "worker-1", nil, nil),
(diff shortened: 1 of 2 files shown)


pkg/cli/pod/pod_test.go
```diff
@@ -295,6 +295,28 @@ func TestPrintPodsHeaderDrivenWidth(t *testing.T) {
...
+	}
+
+	sortPods(pods)
+
+	var got []string
+	for _, p := range pods {
+		got = append(got, p.Namespace+"/"+p.Name)
+	}
+	want := []string{"team-a/worker-0", "team-a/worker-1", "team-b/worker-0", "team-b/worker-1"}
+	if !reflect.DeepEqual(got, want) {
```
hajnalmt: nit: I like to avoid the reflect machinery in modern go, especially in tests. `got` and `want` are both `[]string`, so `slices.Equal(got, want)` would express this more directly than `reflect.DeepEqual`. This file would need to import `slices`; `reflect` is still used elsewhere in the file for the existing error comparison so this was easier I guess.

hajnalmt (approved):
/area ci
/approve 

Thanks for the PR. Leaving others a chance to take a look/let in 😊 if no objection, then merging it soon.

pkg/cli/pod/pod_test.go
aeron-gh: Good call
switched it to `slices.Equal`. Kept `reflect` only for the error comparison in TestListPod. 
Thanks!

PR #5668, opened 2026-07-16
agent-scheduler: validate scheduler-worker-count is greater than zero

### LINKED ISSUE #5666: agent-scheduler accepts scheduler-worker-count=0 and starts with no workers
### Description

The `vc-agent-scheduler` accepts `--scheduler-worker-count=0` without validation. It starts successfully and reports `Running`, but launches no scheduling workers. Pods assigned to `agent-scheduler` remain indefinitely `Pending`.
The option should reject zero during startup because at least one worker is required for scheduling.

1. Configure the development manifest with:

```yaml
- --scheduler-worker-count=0
```

2. Build and load the image.
3. Apply the manifest.
4. Confirm the scheduler is running:

```bash
kubectl -n volcano-system get pods | grep agent-scheduler
```

5. Check the scheduler logs:

```bash
kubectl -n volcano-system logs <agent-scheduler-pod>
```

The logs show:

```text
Scheduler completes Initialization and start to run 0 workers
```

Image

Actual result:

- The scheduler pod starts successfully with status `Running`.
- Logs report that it started `0 workers`.
- Pods assigned to `agent-scheduler` remain indefinitely `Pending`.
- No scheduling events are generated.

Expected result:

- Startup validation rejects `--scheduler-worker-count=0`.
- The process exits with a clear error, such as:

```text
scheduler-worker-count must be greater than 0
```

Alternatively, if zero is intended to disable scheduling, this behavior should be explicitly documented and the component should expose an unhealthy status instead of appearing operational.

Branch: master, Commit: 3019e2ba734bb5bb02c1342c9233bdd71869c2dc

The flag is described as the number of worker threads scheduling in parallel, and the default deployment configuration uses `1`. The current option validation does not reject zero, while the scheduler starts exactly the configured number of workers. A validation check and unit test for `ScheduleWorkerCount == 0` would prevent this silent failure.

miantalha45:
#### What type of PR is this?
/kind bug

This PR validates `--scheduler-worker-count` during agent-scheduler startup.

Previously, setting the value to `0` was accepted. The scheduler started successfully but launched no scheduling workers, leaving pods assigned to `agent-scheduler` in the `Pending` state indefinitely.

This change rejects zero with a clear validation error:

```text
scheduler-worker-count must be greater than 0
```

A regression test was added to verify the behavior.

Fixes #5666 

- The default worker count remains unchanged at `1`.
- Positive worker counts continue to work as before.
- The targeted options package tests pass.

Yes. Invalid configurations using `--scheduler-worker-count=0` now fail during startup instead of silently disabling scheduling.

```release-note
agent-scheduler now rejects scheduler-worker-count values of 0 during startup.
```

### DIFF
--- a/cmd/agent-scheduler/app/options/options.go
+++ b/cmd/agent-scheduler/app/options/options.go
@@ -128,6 +128,9 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 func (s *ServerOption) CheckOptionOrDie() error {
 	s.ServerOption.ShardingMode = s.ShardingMode
 	s.ServerOption.ShardName = s.ShardName
+	if s.ScheduleWorkerCount == 0 {
+		return fmt.Errorf("scheduler-worker-count must be greater than 0")
+	}
 	return componentbaseconfigvalidation.ValidateLeaderElectionConfiguration(&s.LeaderElection, field.NewPath("leaderElection")).ToAggregate()
 }
 
--- a/cmd/agent-scheduler/app/options/options.go
+++ b/cmd/agent-scheduler/app/options/options.go
@@ -128,7 +128,12 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 func (s *ServerOption) CheckOptionOrDie() error {
 	s.ServerOption.ShardingMode = s.ShardingMode
 	s.ServerOption.ShardName = s.ShardName
-	if s.ScheduleWorkerCount == 0 {
+	return s.validateOptions()
+}
+
+// validateOptions validates agent scheduler-specific options.
+func (s *ServerOption) validateOptions() error {
+	if s.ScheduleWorkerCount <= 0 {
 		return fmt.Errorf("scheduler-worker-count must be greater than 0")
 	}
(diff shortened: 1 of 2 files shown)


cmd/agent-scheduler/app/options/options.go
```diff
@@ -128,6 +128,9 @@ func (s *ServerOption) AddFlags(fs *pflag.FlagSet) {
 func (s *ServerOption) CheckOptionOrDie() error {
 	s.ServerOption.ShardingMode = s.ShardingMode
 	s.ServerOption.ShardName = s.ShardName
+	if s.ScheduleWorkerCount == 0 {
```
JesseStutler: Perhaps `<=0` is better, also could you wrap a common func and put arg validation inside, then we can add more args validation in the common func in the future

miantalha45: Sure, that will be much better to check `<=0` . also let me make it a plug-able for validation.

miantalha45:
@JesseStutler done. have a look on it now

JesseStutler:
/approve
/lgtm
Thanks

Issue #4442, opened 2025-07-09
When Vcjob Terminated, Jobflow always pending


dongjiang1989:
### Description

bad case:
```yaml
Name:         gpu-task-avocado-1e0bd46462
Namespace:    xxxx
Labels:       <none>
Annotations:  <none>
API Version:  flow.volcano.sh/v1alpha1
Kind:         JobFlow
Spec:
  Flows:
    Name:  template-avocado
  Job Retain Policy:   retain
Status:
  Conditions:
    gpu-task-avocado-1e0bd46462-template-avocado:
      Create Time:       2025-07-08T08:21:04Z
      Phase:             Terminated
      Running Duration:  2m45s
  Job Status List:
    End Timestamp:  2025-07-08T08:23:49Z
    Name:           gpu-task-avocado-1e0bd46462-template-avocado
    Running Histories:
      End Timestamp:    2025-07-08T08:23:49Z
      Start Timestamp:  2025-07-08T08:21:04Z
      State:            Pending
      End Timestamp:    2025-07-08T08:23:49Z
      Start Timestamp:  2025-07-08T08:23:49Z
      State:            Terminating
      Start Timestamp:  2025-07-08T08:23:49Z
      State:            Terminated
    Start Timestamp:    2025-07-08T08:21:04Z
    State:              Terminated
  State:
    Phase:  Pending
  Terminated Jobs:
    gpu-task-avocado-1e0bd46462-template-avocado
Events:  <none>
```

### Steps to reproduce the issue

update jobflow status

### Describe the results you received and expected

update jobflow status: 

### What version of Volcano are you using?

v1.11

PR #5675, opened 2026-07-17
fix(jobflow): settle the running state to Failed when a job is Terminated

### LINKED ISSUE #5672: JobFlow stays in Running when a job is Terminated, and the failed metric is skipped on that path
/kind bug

### Description

The jobflow controller decides the terminal phase in two states, `pending.go` and `running.go`, and the two have drifted apart.

`pending.go` treats a job that is `Failed` or `Terminated` as a failure:

```go
} else if len(status.FailedJobs) > 0 || len(status.TerminatedJobs) > 0 {
	UpdateJobFlowFailed(p.jobFlow.Namespace)
	status.State.Phase = Failed
}
```

`running.go` only looks at `FailedJobs`:

```go
} else if len(status.FailedJobs) > 0 {
	status.State.Phase = Failed
}
```

Two things fall out of that gap:

1. Once a jobflow is in `Running`, a job that ends up in the `Terminated` phase lands in `TerminatedJobs`, not `FailedJobs`. `running.go` then matches neither the `Succeed` branch (not all jobs completed) nor the `Failed` branch (no failed jobs), so the jobflow stays in `Running` and never reaches a terminal phase.

2. This `Failed` branch does not call `UpdateJobFlowFailed`, even though the `Succeed` branch right above it calls `UpdateJobFlowSucceed`, and the `pending.go` failed branch calls `UpdateJobFlowFailed`. So `volcano_jobflow_failed_phase_count` is not incremented when a jobflow fails from `Running`, which is the usual path since a job normally has to run before it fails.

It looks like the running side just never got the two changes the pending side did. #4442 added the `Terminated` handling to `pending.go` only, and the later metrics change wired `UpdateJobFlowFailed` into `pending.go` only. This is essentially the `Running` twin of #4442.

1. Create a jobflow whose jobs start running, so the jobflow moves to `Running`.
2. Get one of its jobs into the `Terminated` phase.
3. Watch the jobflow status.

Received: the jobflow stays in `Running` and never settles, and `volcano_jobflow_failed_phase_count` is not incremented.

Expected: the jobflow reaches `Failed`, the same as when a job is `Terminated` while the jobflow is still `Pending`, and the failed metric goes up.

master

aeron-gh:
/kind bug

The running state only looked at `FailedJobs`, so once a jobflow reached `Running`, a job that ended up `Terminated` left it stuck in `Running`, and that branch never incremented `volcano_jobflow_failed_phase_count`. The pending state already handles both, this brings the running state in line with it.

#4442 added the `Terminated` handling to the pending state only, this is the running twin of it.

Added a running state test, the `state` package had none.

Fixes #5672

```release-note
Fixed a bug where a JobFlow could stay in the Running phase forever when one of its jobs ended up Terminated.
```

### DIFF
--- a/pkg/controllers/jobflow/state/running.go
+++ b/pkg/controllers/jobflow/state/running.go
@@ -29,7 +29,8 @@ func (p *runningState) Execute(action v1alpha1.Action) error {
 			if len(status.CompletedJobs) == allJobList {
 				UpdateJobFlowSucceed(p.jobFlow.Namespace)
 				status.State.Phase = v1alpha1.Succeed
-			} else if len(status.FailedJobs) > 0 { // TODO(dongjiang199) Modify it when the if condition judgment is implemented
+			} else if len(status.FailedJobs) > 0 || len(status.TerminatedJobs) > 0 { // TODO(dongjiang199) Modify it when the if condition judgment is implemented
+				UpdateJobFlowFailed(p.jobFlow.Namespace)
 				status.State.Phase = v1alpha1.Failed
 			}
 		})
--- a/pkg/controllers/jobflow/state/running_test.go
+++ b/pkg/controllers/jobflow/state/running_test.go
@@ -31,6 +31,9 @@ func TestRunningStateExecute(t *testing.T) {
 		return nil
 	}
 
+	// Execute only runs while the jobflow sits in Running, so every case starts there.
+	running := v1alpha1.State{Phase: v1alpha1.Running}
+
 	tests := []struct {
 		name   string
 		status v1alpha1.JobFlowStatus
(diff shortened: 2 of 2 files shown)


aeron-gh:
@hwdef @hzxuzhonghu 
ping on this PR whenever you have a moment. Thanks!

aeron-gh:
/cc @JesseStutler  @hwdef 
 it only needs an lgtm. would you mind taking a look
when you get a chance? It also closes out https://github.com/volcano-sh/volcano/issues/5672

Thanks!

aeron-gh:
cc @JesseStutler 

Got this to happen on a real cluster, so here is the run.

Two JobTemplates in a JobFlow, the second depending on the first. The second one ends up Terminated through a lifecycle policy the webhook accepts as-is:

```yaml
  policies:
    - event: PodFailed
      action: TerminateJob
```

The first job completes, which puts the JobFlow into Running, then the second one terminates. Nineteen minutes later it is still sitting there:

```
$ kubectl -n repro5675 get jobflow flow-term -o jsonpath='{.status}'
phase          = Running
runningJobs    =
completedJobs  = ["flow-term-tpl-ok"]
failedJobs     =
terminatedJobs = ["flow-term-tpl-term"]

$ kubectl -n repro5675 get vcjob
flow-term-tpl-ok     Completed
flow-term-tpl-term   Terminated

$ kubectl -n repro5675 get pods
flow-term-tpl-ok-worker-0     Succeeded
flow-term-tpl-term-worker-0   Failed
```

Both jobs are terminal, both pods are done, `runningJobs` is empty, and the JobFlow still reports Running. It stayed that way across a controller restart as well.

Which matches what the running state does today: `failedJobs` is empty so the `len(status.FailedJobs) > 0` branch never fires, `terminatedJobs` is not looked at, so neither branch matches and the phase is left as it was. The pending state already covers both.

PR #5678, opened 2026-07-17
refactor(scheduler): make enabled actions session-scoped instead of a global map

### LINKED ISSUE #5674: conf.EnabledActionMap is a mutable package-level global written without synchronization (latent data race)
### Description

`conf.EnabledActionMap` is a package-level mutable map that is recreated on every scheduling cycle inside `pkg/scheduler/scheduler.go`:

```go
conf.EnabledActionMap = make(map[string]bool)
for _, action := range actions {
    conf.EnabledActionMap[action.Name()] = true
}
```

The map is later read by scheduler actions (e.g., `pkg/scheduler/actions/allocate/allocate.go`) and is accessible from any goroutine that references it.

Although the current scheduler execution path appears mostly serialized, this design introduces a latent data race because Go maps are not safe for concurrent access. Any future action parallelization, asynchronous plugin execution, or overlapping scheduling cycles could lead to race detector failures or a `concurrent map read and map write` panic.

This issue makes the scheduler more difficult to reason about and creates hidden constraints for future performance work.

This is primarily an architectural issue and may not reproduce under normal execution today.

1. Inspect the writer in `pkg/scheduler/scheduler.go`:

   ```go
   conf.EnabledActionMap = make(map[string]bool)
   for _, action := range actions {
       conf.EnabledActionMap[action.Name()] = true
   }
   ```

2. Inspect the global declaration in `pkg/scheduler/conf/scheduler_conf.go`:

   ```go
   var EnabledActionMap map[string]bool
   ```

3. Inspect read sites (for example `pkg/scheduler/actions/allocate/allocate.go`) and run:

   ```bash
   go test -race ./pkg/scheduler/...
   ```

4. Introduce concurrent access (e.g., parallel action execution or asynchronous goroutines referencing the map) and observe race detector warnings or potential concurrent map access panics.

### Current behavior

* A mutable package-level map is recreated on every scheduling cycle.
* The map can be read from multiple packages without synchronization.
* The implementation relies on an implicit assumption that no concurrent readers or writers will exist.
* Future parallelism could introduce `concurrent map read and map write` panics or race detector failures.

### Expected behavior

* Enabled actions should be scoped to the scheduling session instead of stored in global mutable state.
* Each session should own its own immutable copy of the enabled action set.
* Scheduler actions should access enabled actions through the session object (e.g., `ssn.EnabledActions`) rather than through `conf.EnabledActionMap`.

## Proposed fix

Move the enabled action map into the scheduling session:

```go
enabledActions := make(map[string]bool, len(actions))
for _, action := range actions {
    enabledActions[action.Name()] = true
}

ssn := framework.OpenSession(pc.cache, plugins, configurations)
ssn.EnabledActions = enabledActions
```

Then replace all usages of:

```go
conf.EnabledActionMap
```

with:

```go
ssn.EnabledActions
```

and remove the package-level global.

## What version of Volcano are you using?

* Observed on: `master` (HEAD as of 2026-07-17)
* Potentially affects all supported versions that include `conf.EnabledActionMap`.

Observed on: `master` (HEAD as of 2026-07-17) * Potentially affects all supported versions that include `conf.EnabledActionMap`. 

### Affected files

* `pkg/scheduler/scheduler.go`
* `pkg/scheduler/conf/scheduler_conf.go`
* `pkg/scheduler/framework/session.go`
* `pkg/scheduler/actions/allocate/allocate.go`

### Environment

* Kubernetes version: N/A (design-level issue)
* OS: N/A
* Kernel version: N/A

### Validation

```bash
go test ./pkg/scheduler/...
go test -race ./pkg/scheduler/...
```

## Additional context

This is a low-risk refactor that removes shared mutable scheduler state and improves correctness, maintainability, and future scheduler parallelization efforts without changing scheduling semantics.

vanshika2720:
## What type of PR is this?

/kind cleanup

This PR removes the package-level mutable global `conf.EnabledActionMap` and moves the enabled action set into the scheduler session.

Currently, the scheduler executes scheduling sessions serially, so this change does not fix a user-visible concurrency issue. Instead, it scopes the enabled action state to the object that owns it (`framework.Session`), making ownership explicit, improving test isolation, and removing unnecessary mutable global state.

This preserves the existing scheduling behavior while making the design easier to reason about and less dependent on process-wide state.

Fixes #5674

### Changes

- Added `EnabledActions map[string]bool` to `framework.Session`.
- Removed the package-level `conf.EnabledActionMap`.
- Built the enabled action map locally in `scheduler.runOnce()`.
- Assigned the map to `ssn.EnabledActions` after opening the session.
- Replaced `conf.EnabledActionMap` lookups with `ssn.EnabledActions`.
- Updated the scheduler test helper to initialize `EnabledActions` on the session.
- Updated capacity and proportion plugin tests to use session-scoped state instead of swapping a global variable.

### Verification

- `grep -R "EnabledActionMap" .` returns zero matches.
- `go test ./pkg/scheduler/...` passes.
- `go test -race ./pkg/scheduler/...` passes.
- `go build ./pkg/scheduler/...` succeeds.
- `gofmt` clean on all modified files.

### Behavior

- No scheduling logic changes.
- No action ordering changes.
- No plugin behavior changes.
- Moves enabled action state from global scope to session scope.
- Improves state ownership and test isolation by removing mutable global state.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -152,7 +152,7 @@ func (alloc *Action) buildAllocateContext() *allocateContext {
 	for _, job := range ssn.Jobs {
 		// If not config enqueue action, change Pending pg into Inqueue state to avoid blocking job scheduling.
 		if job.IsPending() {
-			if conf.EnabledActionMap["enqueue"] {
+			if ssn.EnabledActions["enqueue"] {
 				klog.V(4).Infof("Job <%s/%s> Queue <%s> skip allocate, reason: job status is pending.",
 					job.Namespace, job.Name, job.Queue)
 				continue
--- a/pkg/scheduler/conf/scheduler_conf.go
+++ b/pkg/scheduler/conf/scheduler_conf.go
@@ -21,9 +21,6 @@ limitations under the License.
 
 package conf
 
-// EnabledActionMap check if a action exist in scheduler configmap. If not exist the value is false.
-var EnabledActionMap map[string]bool
-
 // SchedulerConfiguration defines the configuration of scheduler.
 type SchedulerConfiguration struct {
 	// Actions defines the actions list of scheduler in order
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -99,6 +99,10 @@ type Session struct {
 	Tiers          []conf.Tier
 	Configurations []conf.Configuration
 	NodeList       []*api.NodeInfo
+	// EnabledActions records which actions are configured to run in this scheduling
+	// session. It is populated once by the caller after the session is opened and
+	// must not be mutated afterward.
+	EnabledActions map[string]bool
 	// HyperNodes stores the HyperNodeInfo of each HyperNode
 	HyperNodes           api.HyperNodeInfoMap
 	HyperNodeTierNameMap api.HyperNodeTierNameMap
@@ -186,6 +190,7 @@ func openSession(cache cache.Cache) *Session {
 		CSINodesStatus: map[string]*api.CSINodeStatusInfo{},
 		RevocableNodes: map[string]*api.NodeInfo{},
 		Queues:         map[api.QueueID]*api.QueueInfo{},
+		EnabledActions: map[string]bool{},
 
 		plugins:                       map[string]Plugin{},
 		jobOrderFns:                   map[string]api.CompareFn{},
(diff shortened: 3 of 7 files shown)


vanshika2720:
@hajnalmt 
@hwdef 
@JesseStutler

hwdef:
This bug seems to be a very marginal scenario.

vanshika2720:
@hwdef You're right that this isn't a user-visible issue in the current scheduler, and my original description overstated it as a bug. My intention was to remove mutable package-level state whose lifetime is really session-scoped. The change doesn't alter scheduling behavior it simply makes ownership explicit, improves test isolation, and avoids relying on shared global state. I'll rebase the branch onto the latest master and update the PR description to better reflect that this is a refactoring rather than a fix for an actively reproducible bug.

PR #5679, opened 2026-07-18
fix(cronjob): prune Terminated jobs with the failed history limit

### LINKED ISSUE #5655: CronJob controller never cleans up Terminated jobs: history limits skip them
### Description

`isJobFinished` in `pkg/controllers/cronjob/cronjob_controller_handler.go` treats three phases as finished:

```go
func isJobFinished(job *batchv1.Job) (bool, batchv1.JobPhase) {
	if job.Status.State.Phase == batchv1.Completed || job.Status.State.Phase == batchv1.Failed || job.Status.State.Phase == batchv1.Terminated {
		return true, job.Status.State.Phase
	}
	return false, ""
}
```

But `processFinishedJobs`, which consumes that return value, only buckets two of them:

```go
switch phase {
case batchv1.Completed:
    ...
    successfulJobs = append(successfulJobs, job)
case batchv1.Failed:
    failedJobs = append(failedJobs, job)
}
```

There is no `Terminated` case and no `default`, so a job that finishes in `Terminated` falls through. It is correctly removed from the CronJob's active list, but it never lands in `successfulJobs` or `failedJobs`. Since `removeOldestJobs` is only ever called with those two slices, a `Terminated` job is never counted against `successfulJobsHistoryLimit` or `failedJobsHistoryLimit`, and never gets deleted.

`Terminated` is a reachable terminal phase (`pkg/controllers/job/state/terminating.go` sets `status.State.Phase = vcbatch.Terminated`), and the rest of the codebase consistently treats it as a peer of `Completed`/`Failed`:

- `pkg/controllers/job/state/factory.go`: `case vcbatch.Terminated, vcbatch.Completed, vcbatch.Failed:` returns the finished state
- the garbage collector has its own `isJobFinished` that also includes `Terminated`

so the CronJob history-limit switch looks like the odd one out.

1. Create a CronJob with a small history limit, for example `failedJobsHistoryLimit: 1`.
2. Have its jobs end up in the `Terminated` phase, that is a job that gets terminated rather than completing or failing (for example through a policy action that terminates the job).
3. Let the CronJob fire a few times, then list the jobs in the namespace.

Received: the `Terminated` jobs keep piling up. They are gone from `status.active`, but they are never deleted no matter what `successfulJobsHistoryLimit` / `failedJobsHistoryLimit` are set to.

Expected: `Terminated` jobs get pruned like the other finished jobs, so the number kept is bounded by the configured history limit.

The only thing that reclaims them today is the garbage collector, and that only kicks in when `ttlSecondsAfterFinished` is set on the job template, which is optional. Without it they stay around forever.

This looks like an oversight rather than intent: `Terminated` was added to `isJobFinished` but not to the switch that consumes it, and there is no test covering the `Terminated` phase in the cronjob controller tests.

The natural fix looks like bucketing `Terminated` with the failed jobs, since it is documented as "the job is finished unexpected", so that `failedJobsHistoryLimit` applies to it:

```go
case batchv1.Failed, batchv1.Terminated:
    failedJobs = append(failedJobs, job)
```

Happy to send a PR with that plus a test if that is the behavior you want, or to do it differently if you would rather keep `Terminated` in its own bucket.

master

aeron-gh:
#### What type of PR is this?

/kind bug

`isJobFinished` counts `Terminated` as finished, but the `switch` in `processFinishedJobs` only has cases for `Completed` and `Failed`. A Terminated job matches neither, so it never lands in `successfulJobs` or `failedJobs`, never reaches `removeOldestJobs`, and is never pruned by `successfulJobsHistoryLimit` or `failedJobsHistoryLimit`. They pile up for the lifetime of the CronJob, and the only thing that clears them today is the opt-in `ttlSecondsAfterFinished`.

This adds `Terminated` to the failed case so it ages out with the failed jobs.

Fixes #5655

On the bucketing: `failedJobs` is the only option that does not need a new API field. A Terminated job is not a success, and putting it in `successfulJobs` would also move `LastSuccessfulTime`. The jobflow controller already treats Terminated the same way in `pending.go` (`FailedJobs > 0 || TerminatedJobs > 0` gives Failed), which came out of #4442.

The test asserts both halves: the Terminated job gets pruned, and `LastSuccessfulTime` stays nil. It is the first Terminated coverage in the cronjob tests.

```release-note
Fixed a bug where CronJob history limits never pruned jobs that ended in the Terminated phase, causing them to accumulate indefinitely.
```

### DIFF
--- a/pkg/controllers/cronjob/cronjob_controller_handler.go
+++ b/pkg/controllers/cronjob/cronjob_controller_handler.go
@@ -207,7 +207,8 @@ func (cc *cronjobcontroller) processFinishedJobs(cronJob *batchv1.CronJob, jobsB
 					updateStatus = true
 				}
 				successfulJobs = append(successfulJobs, job)
-			case batchv1.Failed:
+			case batchv1.Failed, batchv1.Terminated:
+				// A terminated job did not succeed, so it is pruned with the failed ones
 				failedJobs = append(failedJobs, job)
 			}
 		}
--- a/pkg/controllers/cronjob/cronjob_controller_test.go
+++ b/pkg/controllers/cronjob/cronjob_controller_test.go
@@ -766,6 +766,36 @@ func TestSyncCronJob(t *testing.T) {
 				expectDeleteJobNum:     1,
 			},
 		},
+		{
+			name: "not first run, terminated job, failedJobsHistoryLimit is 0",
+			now:  justTen().Add(7 * time.Minute),
+			cronjob: cjSpec{
+				concurrency:            "Allow",
+				failedJobsHistoryLimit: ptr.To[int32](0),
+			},
+			jobs: []jobSpec{
+				{
+					status:      batchv1.Terminated,
+					time:        fiveMinutesAfterTen(),
+					inLister:    true,
+					inClient:    false,
+					active:      false,
+					processTime: time.Minute,
+				},
+			},
+			lastScheduleTime: &metav1.Time{Time: fiveMinutesAfterTen()},
+			expect: testExpect{
+				expectRequeueAfter:     durationPtr(3*time.Minute + nextScheduleDelta),
+				expectUpdateStatus:     true,
+				expectError:            false,
+				expectLastScheduleTime: &metav1.Time{Time: fiveMinutesAfterTen()},
+				expectInActiveNum:      0,
+				expectEventsNum:        1,
+				expectLastSuccessTime:  nil,
+				expectCreateJobNum:     0,
+				expectDeleteJobNum:     1,
+			},
+		},
 		//TODO: 增加默认值
 		// {
 		// 	name:        "noy first run, successfulJobHistoryLimit is nil",


hwdef:
/approve
Thanks

aeron-gh:
/cc @hajnalmt 
hwdef already approved it, so it only needs an lgtm. Would you mind taking a look
when you get a chance? It also closes out #5655.


Thanks!

aeron-gh:
https://github.com/volcano-sh/volcano/issues/5655#issuecomment-5025363218

aeron-gh:
Ran this one on a real cluster and just left it going.

A CronJob on a one minute schedule whose jobs all end Terminated, again through a policy the webhook accepts:

```yaml
spec:
  schedule: "* * * * *"
  concurrencyPolicy: Allow
  successfulJobsHistoryLimit: 1
  failedJobsHistoryLimit: 1
  jobTemplate:
    spec:
      policies:
        - event: PodFailed
          action: TerminateJob
```

With those two limits at most two finished jobs should ever be kept. Counting them each minute:

```
minute 1:  2 jobs
minute 4:  5 jobs
minute 6:  7 jobs
minute 8:  9 jobs
```

and leaving it running a while longer:

```
$ kubectl -n repro5679 get vcjob --no-headers | wc -l
21

$ kubectl -n repro5679 get vcjob -o jsonpath='{range .items[*]}{.status.state.phase}{"\n"}{end}' | sort | uniq -c
     21 Terminated
```

One more job retained every minute, nothing ever pruned. The CronJob is otherwise fine, `lastScheduleTime` keeps advancing, it is only the cleanup that never happens.

That lines up with the switch in `processFinishedJobs` having only the `Completed` and `Failed` cases while `isJobFinished` also returns true for `Terminated`, so those jobs never reach `removeOldestJobs`.

aeron-gh:
@JesseStutler ptal whenever u have time!

aeron-gh:
@JesseStutler @hwdef @hajnalmt 
PTAL

PR #5680, opened 2026-07-18
Treat terminated jobs as failed for CronJob history cleanup and add u…

### LINKED ISSUE #5655: CronJob controller never cleans up Terminated jobs: history limits skip them

meekailkhan:
#### What type of PR is this?
bug fix

this pr added a case for appending  terminated job in failed job for cleanup cron job history and write unit test
Fixes #5655

NONE
```release-note
N/A
```

### DIFF
--- a/pkg/controllers/cronjob/cronjob_controller_handler.go
+++ b/pkg/controllers/cronjob/cronjob_controller_handler.go
@@ -209,6 +209,8 @@ func (cc *cronjobcontroller) processFinishedJobs(cronJob *batchv1.CronJob, jobsB
 				successfulJobs = append(successfulJobs, job)
 			case batchv1.Failed:
 				failedJobs = append(failedJobs, job)
+			case batchv1.Terminated:
+				failedJobs = append(failedJobs, job)
 			}
 		}
 	}
--- a/pkg/controllers/cronjob/cronjob_controller_handler_test.go
+++ b/pkg/controllers/cronjob/cronjob_controller_handler_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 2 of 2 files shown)


aeron-gh:
Just flagging that this is the same fix as #5655, which I opened and am assigned to. 
I've already got a PR up for it at #5679. Happy to let a maintainer decide which one to go with.

PR #5685, opened 2026-07-19
Fix missing session state rollback on RecoverOperations failure

### LINKED ISSUE #5684: RecoverOperations failure may leave partially replayed session state unrolled back
### Description

RecoverOperations() replays a previously saved sequence of scheduling operations onto a Statement. If an error occurs after some operations have already been replayed, the caller currently returns (or retries) without explicitly calling Discard() to roll back the successfully replayed operations.

This behavior exists in:

- pkg/scheduler/actions/allocate/allocate.go
- pkg/scheduler/actions/gangpreempt/gangpreempt.go
- pkg/scheduler/actions/gangreclaim/gangreclaim.go

As a result, partially replayed operations may remain applied after RecoverOperations() returns an error.

If RecoverOperations() fails after partially replaying operations, the caller should roll back the successfully replayed operations before returning or retrying.

1. Create a recovery plan that contains multiple operations, where at least one operation succeeds before a later operation fails (for example, an Evict followed by a failing Pipeline or Allocate).

2. Invoke RecoverOperations() with this plan.

3. Observe that the caller exits the error path without explicitly calling Discard() to roll back the operations that were successfully replayed before the failure.

Received:
If RecoverOperations() fails after partially replaying operations, the caller returns (or retries) without explicitly rolling back the successfully replayed operations.

Expected:
The caller should invoke Discard() on the error path so that any partially replayed operations are rolled back before returning or retrying.

master

_No response_

A69SHUBHAM:
#### What type of PR is this?

/kind bug

This PR adds rollback on the error paths following `RecoverOperations()`.

`RecoverOperations()` replays operations sequentially. If it returns an error after partially replaying operations, the caller currently returns (or retries) without explicitly discarding the successfully replayed operations.

This change calls `Discard()` before those error exits so that any partially replayed operations are rolled back before returning or retrying.

A regression test has been added to verify the rollback behavior.

Fixes #5684 

`Discard()` only reverses operations that have already been successfully recorded in the statement. Since operations are appended during successful replay, calling `Discard()` on the error path rolls back only the operations that were applied before the failure and does not affect the successful execution path.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -446,6 +446,7 @@ func (alloc *Action) allocateForJob(job *api.JobInfo, jobWorksheet *JobWorksheet
 		finalStmt := framework.NewStatement(ssn)
 		if err = finalStmt.RecoverOperations(bestStmt); err != nil {
 			klog.ErrorS(err, "Failed to recover operations", "job", job.UID, "hyperNode", bestHyperNode)
+			finalStmt.Discard()
 			return nil
 		}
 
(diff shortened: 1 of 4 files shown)


A69SHUBHAM:
@hajnalmt @JesseStutler can you review it ?

PR #5686, opened 2026-07-19
webhooks: validate PodGroup updates

avinxshKD:
#### What type of PR is this?

/kind bug

PodGroup validation rejects invalid networkTopology config, but the webhook only ran on create.

This allows a PodGroup to be updated with both highestTierAllowed and highestTierName set. This PR makes the PodGroup validating webhook handle updates too.

NONE

NONE

```release-note
PodGroup validating webhook now rejects invalid networkTopology updates.


avinxshKD:
@hajnalmt @JesseStutler found this while checking PodGroup admission: create rejects conflicting networkTopology, but update was never sent to the validator. PTAL

PR #5689, opened 2026-07-19
fix(hypernode): don't mutate the informer cache or clobber the status on update

### LINKED ISSUE #5683: HyperNode controller mutates the informer cache and uses a stale resourceVersion in UpdateHyperNode
/kind bug

### Description

`UpdateHyperNode` in `pkg/controllers/hypernode/utils/utils.go` has two problems in how it handles the HyperNode object.

**1. It mutates the informer cache.**

`lister.Get()` returns a pointer to the shared object in the informer cache, so it must not be modified. `UpdateHyperNode` writes straight into it:

```go
current, err := lister.Get(updated.Name)
...
current.Spec = updated.Spec
current.Status = updated.Status
current.Labels[k] = v
```

This corrupts the cache for every other reader (the scheduler and other controllers read HyperNodes from the same informer). The sibling code in `hypernode_handler.go` already knows this and deep copies first:

```go
// Create a deep copy to avoid modifying cache objects
hyperNodeCopy := hyperNode.DeepCopy()
```

**2. It uses a stale resourceVersion for UpdateStatus.**

HyperNode has a status subresource (`+kubebuilder:subresource:status`), so both `Update` (spec and metadata) and `UpdateStatus` (status) are needed. `Update` returns the object with a bumped resourceVersion, but the code discards it and reuses the old `current` for `UpdateStatus`:

```go
_, err = ...Update(ctx, current, ...)        // bumps the resourceVersion, return discarded
_, err = ...UpdateStatus(ctx, current, ...)  // current still has the old resourceVersion
```

Because `current` still carries the pre-Update resourceVersion, `UpdateStatus` fails the optimistic concurrency check. The call is wrapped in `RetryOnConflict`, so it retries and eventually succeeds once the informer cache catches up, but every spec change now goes through a conflict, a retry, and one or more redundant `Update` writes.

This second one came in with #4498, which fixed the original "spec is never updated" report in #4497 by adding the `Update` call, but it threaded the resourceVersion wrong.

1. Trigger a HyperNode spec change so the controller reconciles it (for example, remove a member as in #4497).
2. Watch the controller: the spec update goes through, but the status update logs a conflict and retries until the cache catches up.

Received: the cache object is mutated in place, and the status update conflicts on a stale resourceVersion and retries.

Expected: the update works on a copy of the cache object, and the status write uses the resourceVersion returned by `Update`, so there is no conflict.

The fix is small: deep copy `current` before changing it, and carry the status onto the object returned by `Update` before calling `UpdateStatus`. Happy to send the PR.

master

aeron-gh:
#### What type of PR is this?

/kind bug

`UpdateHyperNode` has two problems.

1. It mutates the object returned by the lister, which is a shared cache object, so writing to `current.Spec` and `current.Labels` corrupts the informer cache for every other reader. The sibling code in `hypernode_handler.go` already deep copies for this reason.

2. It hands a stale resourceVersion to the status write. `Update` returns the object with the bumped version, but that return value is discarded and `UpdateStatus` is given the pre-update `current`, so every real spec change fails the optimistic concurrency check and goes through a conflict and a retry.

This deep copies the cache object before changing it, and carries the status onto the object `Update` returns before calling `UpdateStatus`.

Two tests, the package had none before: one fails if the cached object is mutated, the other stands in for the apiserver and rejects a status write that still carries the old resourceVersion.

Fixes #5683

An earlier revision of this PR dropped the status handling altogether. That was wrong, as pointed out in review, since a discoverer can set its own status. It is restored and the change is now limited to the two bugs above.

```release-note
Fixed the HyperNode controller mutating the informer cache and using a stale resourceVersion when updating HyperNode status.
```

### DIFF
--- a/pkg/controllers/hypernode/utils/utils.go
+++ b/pkg/controllers/hypernode/utils/utils.go
@@ -46,6 +46,8 @@ func UpdateHyperNode(vcClient vcclientset.Interface, lister v1alpha1.HyperNodeLi
 			return err
 		}
 
+		// The lister returns a shared cache object, so work on a copy.
+		current = current.DeepCopy()
 		current.Spec = updated.Spec
 		current.Status = updated.Status
 
@@ -63,11 +65,15 @@ func UpdateHyperNode(vcClient vcclientset.Interface, lister v1alpha1.HyperNodeLi
 			current.Annotations[k] = v
 		}
 
-		_, err = vcClient.TopologyV1alpha1().HyperNodes().Update(context.Background(), current, metav1.UpdateOptions{})
+		applied, err := vcClient.TopologyV1alpha1().HyperNodes().Update(context.Background(), current, metav1.UpdateOptions{})
 		if err != nil {
 			return err
 		}
-		_, err = vcClient.TopologyV1alpha1().HyperNodes().UpdateStatus(context.Background(), current, metav1.UpdateOptions{})
+
+		// Update bumped the resourceVersion, so carry the status onto what it
+		// returned. Reusing current here sends a stale version and conflicts.
+		applied.Status = current.Status
+		_, err = vcClient.TopologyV1alpha1().HyperNodes().UpdateStatus(context.Background(), applied, metav1.UpdateOptions{})
 		return err
 	})
 }
--- a/pkg/controllers/hypernode/utils/utils_test.go
+++ b/pkg/controllers/hypernode/utils/utils_test.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"fmt"
+	"testing"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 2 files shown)


aeron-gh:
@hwdef 
Gentle ping on this
Whenever u have some time, looking for approved
Thanks!

hwdef:
> @hwdef Gentle ping on this Whenever u have some time, looking for approved Thanks!

hi, each reviewer only can add one label. every PR need two reviewer to review.

aeron-gh:
@DSFans2014 @archlitchi 
Hi maintainers, whenever it's convenient, could you please take a look at this PR? I'd be grateful for any feedback. 
Thanks for your time!

pkg/controllers/hypernode/utils/utils.go
```diff
@@ -64,10 +67,6 @@ func UpdateHyperNode(vcClient vcclientset.Interface, lister v1alpha1.HyperNodeLi
 		}
 
 		_, err = vcClient.TopologyV1alpha1().HyperNodes().Update(context.Background(), current, metav1.UpdateOptions{})
-		if err != nil {
```
JesseStutler, line 67: The deletion / "The status is owned by syncHyperNodeStatus" is not correct, the `UpdateHyperNode` is called by topology auto discoverer, the discoverer can update their own status

aeron-gh: @JesseStutler 

yeah fair, i shouldn't have dropped that. i took the bot's suggestion on it and it moved me off what i'd originally written in the issue, which was to keep UpdateStatus and just hand it the object Update gives back. what threw me is that nothing in tree sets a status on the discovered node today so it looked dead, but that's not a reason to remove the path.

restoring it. that leaves the two bits i think are actually broken, the lister object gets written to in place so the informer cache is corrupted for other readers, and UpdateStatus reuses the resourceVersion from before Update so every real spec change goes through a conflict and a retry. carrying the status onto what Update hands back sorts the second one out anyway.

ill push that shortly, shout if you'd rather it was scoped differently

aeron-gh:
@JesseStutler @hwdef 
ci is green i updated the pr
ptal thank you!

PR #5691, opened 2026-07-20
queue-controller: keep closed-by-parent annotation in sync on manual re-close

### LINKED ISSUE #5681: [Question/Bug] Manual close on a closed-by-parent child queue is overridden when parent queue is reopened
### Description

I noticed a behavior regarding the state synchronization between parent and child queues, and I wanted to get a sanity check on whether this is the intended design or a bug.
When a parent queue is closed, the controller automatically closes the child queue and tags it with the annotation `volcano.sh/closed-by-parent: "true"`. 
However, if an admin manually closes the child queue *while the parent is closed* (using `vcctl queue operate --name  <child-queue> --action close`), the controller does not update the `closed-by-parent` annotation because the queue state is already `Closed`. 
When the parent queue is later reopened, the controller sees `closed-by-parent: "true"` on the child queue and automatically reopens it, overriding the explicit manual close command.

1. Create a parent queue and a child queue pointing to it (both in `Open` state).
2.  Close the parent queue:
   ```bash
vcctl queue operate --name <parent-queue> --action close
   ```
3.  Close the child queue:
   ```bash
    vcctl queue operate --name <child-queue> --action close
```
4. Reopen the parent queue:
```bash
vcctl queue operate --name <parent-queue> --action open
```
5. Check child queue status:
```bash
kubectl get queue <child-queue> -o yaml   #childqueue should be open 
```

**Current behaviour:**  When the parent queue is reopened, the child queue automatically transitions back to the `Open` state. The annotation `volcano.sh/closed-by-parent: "true"` is never updated during the manual closing of child queue, which causes the controller to believe the queue was only closed by the parent, thus auto-reopening it.

**Expected behaviour:** The child queue should remain in the `Closed` state after the parent queue is reopened. When the admin explicitly closes the child-queue, the annotation `volcano.sh/closed-by-parent` should be updated to `"false" (or deleted)`, indicating that the queue is now manually closed and should not be subject to automatic parent-conjunction reopening.

version: "1.15.0"

_No response_

pujitha24:
What type of PR is this?
/kind bug

When a parent queue closes, the controller cascades the close to its child
queues and tags each child with the annotation `volcano.sh/closed-by-parent:
"true"`, so that reopening the parent auto-reopens the child. If an admin
then explicitly closes that same child queue directly (e.g. `vcctl queue
operate --action close`) while it is already Closed/Closing, the request was
previously routed to a code path that never touched the annotation. So when
the parent is later reopened, the controller still sees
`closed-by-parent: "true"` and silently auto-reopens the child, overriding
the admin's explicit close. This is a pure queue open/close state-machine
bug: the child queue ends up unexpectedly reopened; it is not a crash or
data loss.

Approach:
- Added `state.ClearClosedByParentAnnotation`, wired to a new
  `queuecontroller.clearClosedByParentAnnotation` that patches the
  `closed-by-parent` annotation to `"false"`.
- `closedState.Execute`/`closingState.Execute` (pkg/controllers/queue/state)
  now call it when a `CloseQueueAction` re-targets an already
  Closed/Closing queue.
- To avoid misfiring on a duplicate/retried automatic cascade request that
  happens to reach an already-Closed/Closing queue (the workqueue dedupes
  `*apis.Request` by pointer identity, not value, so this can occur when the
  parent's own close is retried after a failed status update, or on
  concurrent resyncs), the annotation is only cleared when the request's
  `Event` is `busv1alpha1.CommandIssuedEvent` - the signal already set by
  `handleCommand` exclusively for user-issued `vcctl`/Command-CR actions,
  never by the automatic parent<->child cascade paths
  (`closeHierarchicalQueue`, `syncHierarchicalQueue`). `state.NewState` now
  takes the triggering event so this signal is available where needed.

Which issue(s) this PR fixes:
Fixes #5681

AI tools were used to help investigate, implement, and adversarially review
this change (multiple rounds of self-review before submission); I have
reviewed and understand the diff.

```release-note
Fixed a bug where manually closing a child queue while its parent queue was
already closed did not clear the `volcano.sh/closed-by-parent` annotation,
causing the child queue to be unexpectedly auto-reopened when the parent
queue was later reopened.
```

Validation:
go build ./pkg/controllers/queue/...
go test ./pkg/controllers/queue/... -v
All existing and new tests pass.

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/controllers/queue/queue_controller.go
+++ b/pkg/controllers/queue/queue_controller.go
@@ -162,6 +162,7 @@ func (c *queuecontroller) Initialize(opt *framework.ControllerOption) error {
 	queuestate.SyncQueue = c.syncQueue
 	queuestate.OpenQueue = c.openQueue
 	queuestate.CloseQueue = c.closeQueue
+	queuestate.ClearClosedByParentAnnotation = c.clearClosedByParentAnnotation
 
 	c.syncHandler = c.handleQueue
 	c.syncCommandHandler = c.handleCommand
@@ -235,7 +236,7 @@ func (c *queuecontroller) handleQueue(req *apis.Request) error {
 		return fmt.Errorf("get queue %s failed for %v", req.QueueName, err)
 	}
 
-	queueState := queuestate.NewState(queue)
+	queueState := queuestate.NewState(queue, req.Event)
 	if queueState == nil {
 		return fmt.Errorf("queue %s state %s is invalid", queue.Name, queue.Status.State)
 	}
--- a/pkg/controllers/queue/queue_controller_action.go
+++ b/pkg/controllers/queue/queue_controller_action.go
@@ -174,9 +174,9 @@ func (c *queuecontroller) closeQueue(queue *schedulingv1beta1.Queue, updateState
 
 // clearClosedByParentAnnotation marks the queue as explicitly closed, so that syncHierarchicalQueue
 // no longer treats it as merely closed-by-parent and reopens it when the parent queue reopens.
-func (c *queuecontroller) clearClosedByParentAnnotation(queue *schedulingv1beta1.Queue) error {
-	_, err := c.updateQueueAnnotation(queue, ClosedByParentAnnotationKey, ClosedByParentAnnotationFalseValue)
-	return err
+// It returns the patched queue so callers can replace their stale in-memory reference.
+func (c *queuecontroller) clearClosedByParentAnnotation(queue *schedulingv1beta1.Queue) (*schedulingv1beta1.Queue, error) {
+	return c.updateQueueAnnotation(queue, ClosedByParentAnnotationKey, ClosedByParentAnnotationFalseValue)
 }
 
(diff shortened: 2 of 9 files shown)


pujitha24:
Good catch, both from gemini-code-assist and Copilot — `clearClosedByParentAnnotation` was discarding the patched queue, so `cs.queue` in `closedState`/`closingState` stayed stale after the annotation was cleared, which could let `syncHierarchicalQueue` see the old `closed-by-parent: "true"` value and re-open the queue in the same reconcile.

I've updated `clearClosedByParentAnnotation` and the `ClearClosedByParentAnnotation` func var to return `(*schedulingv1beta1.Queue, error)`, and `closed.go`/`closing.go` now assign the returned queue back to `cs.queue` before calling `SyncQueue`. Updated the mocks/assertions in the three affected test files accordingly. `go build` and `go test` for `pkg/controllers/queue/...` pass.

PR #5706, opened 2026-07-20
docs: require evidence for bug reports

JesseStutler:
#### What type of PR is this?

/kind documentation

Some recent `kind/bug` issues are based only on source scans. They describe unlikely edge cases or ask for extra validation without showing a failure that users can reach. Some are followed by implementation PRs before maintainers agree that the bug is valid.

This takes review time away from production bugs and useful enhancements.

This PR updates the reporting rules:

1. Cluster bugs should include the workload, configuration, and relevant logs, events, or stack trace.
2. Race, panic, and deadlock reports should include runtime evidence, a reproducer, or the complete concurrent path.
3. Performance reports should include a profile or representative benchmark.
4. Validation and hardening reports should explain why the input occurs in normal use and what failure it causes.
5. Source analysis should use links to an exact commit. A copied code block is not enough.
6. Reports that have not been reproduced should say so and explain how a user can reach the failure.
7. Bug-fix PRs should link to an issue with the required evidence. If there is no issue, the PR should include the affected version, the user path to the failure, and the evidence required for that type of report.

A suggested fix is optional. AI output, patch plans, and tests that only call internal functions do not replace evidence.

The bug issue form now requires reproduction steps and an evidence section. The PR template includes a shorter version of the same rule. The AI guidance now also applies to issue reports.

Maintainers may close an issue or PR if it does not follow the template or if requested information is not provided.

NONE

Drafted with AI assistance and revised by the author.

NONE

```release-note
NONE
```

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -70,6 +70,29 @@ Issues should be filed under the appropriate Volcano sub-repository.
 
 Please follow the prompted submission guidelines while opening an issue.
 
+Every bug report must include evidence. State the affected version and
+configuration. Provide the evidence that matches the type of report:
+
+1. For a bug observed in a cluster, include the workload or configuration.
+   Include the relevant logs, events, or stack trace.
+2. For a race, panic, or deadlock, include a race report, panic stack, or
+   reproducer. A complete concurrent execution path is also acceptable.
+3. For a performance problem, include a profile or representative benchmark.
+   State the scale used for the test.
+4. For validation or hardening, explain why the input is valid or expected in
+   normal use. Describe the failure it causes. Malformed input or an artificial
+   call to an internal function is not enough if the system rejects it safely.
+
+If the report includes source code analysis, use permalinks that point to the
+exact commit. Do not replace these links with copied code blocks. If the problem
+has not been reproduced, state that clearly. Use the permalinks to explain the
+complete path from a user action or cluster event to the failure.
+
+A suggested fix is optional. A patch plan does not replace evidence. A unit test
+that calls an internal function does not prove that users can reach the bug.
+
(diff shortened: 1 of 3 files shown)


contributing.md
```diff
@@ -94,15 +119,41 @@ The bot may also make some helpful suggestions for commands to run in your PR to
...
-[Large AI generated](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#large-or-automatic-edits) PRs and AI generated commit messages are discouraged.
+Using AI tools to help prepare an issue or PR is acceptable. The author remains
+responsible for every claim and change.
+
+For an AI-assisted issue, verify the affected version, code references,
+production path, and supporting evidence before submission. AI output and
+static analysis are starting points. They do not prove that a bug exists. Bug
+reports must follow the requirements in [File an Issue](#file-an-issue).
+
+For an AI-assisted PR, review every change and run the appropriate tests before
```
hajnalmt, line 134: Maybe we can add something like:
PRs not following our PR templates and issues not following our Issue templates can be closed automatically by the reviewers.

hajnalmt (commented):
+++ Jesse. This bothers me too. 👍

contributing.md
JesseStutler: Yes thanks, I have added the rule in L122

hajnalmt: Thanks!

hajnalmt (approved):
++ Thanks, awesome!
Let's see if other suggestions are coming in too.

hwdef:
I agree. We should take this opportunity to alleviate the pressure on reviewers and thereby improve the quality of the review process.

/approve

PR #5707, opened 2026-07-20
Reject JobFlow updates in admission

### LINKED ISSUE #5723: JobFlow spec updates are accepted but not applied by controller
### Description

JobFlow accepts UPDATE requests, but normal spec changes are not handled by the controller.

This was found by checking the JobFlow admission and controller code paths. The webhook registers JobFlow validation for CREATE and UPDATE, so an updated JobFlow can pass admission. In the controller, the JobFlow update path is explicitly reserved for future use and does not apply normal JobFlow process changes.

That leaves users with an accepted JobFlow spec update that does not affect the running workflow.

1. Create a JobFlow, for example from `example/jobflow/JobFlow.yaml`.
2. Change `spec.flows` or one of the `spec.flows[*].dependsOn` entries.
3. Apply the manifest again.

Permalinks are pinned to `d34c571c`.

The JobFlow validating webhook is registered for both `CREATE` and `UPDATE`:

https://github.com/volcano-sh/volcano/blob/d34c571c378b4c87865297ceb475d55610d174b1/pkg/webhooks/admission/jobflows/validate/validate_jobflow.go#L53

The handler currently validates `CREATE` and `UPDATE` through the same path:

https://github.com/volcano-sh/volcano/blob/d34c571c378b4c87865297ceb475d55610d174b1/pkg/webhooks/admission/jobflows/validate/validate_jobflow.go#L84

The JobFlow controller registers an update handler:

https://github.com/volcano-sh/volcano/blob/d34c571c378b4c87865297ceb475d55610d174b1/pkg/controllers/jobflow/jobflow_controller.go#L101

But the update handler says JobFlow update is reserved for future use and current updates do not affect the JobFlow process:

https://github.com/volcano-sh/volcano/blob/d34c571c378b4c87865297ceb475d55610d174b1/pkg/controllers/jobflow/jobflow_controller_handler.go#L68

So a JobFlow spec update can pass admission, while the controller does not apply that changed workflow behavior.

Actual:
The JobFlow update is accepted.

Expected:
Unsupported JobFlow spec updates should be rejected by admission, or the controller should actually reconcile them.

master

_No response_

avinxshKD:
#### What type of PR is this?

/kind bug

JobFlow updates are currently accepted by admission, but the controller does not apply updated JobFlow spec changes.

The JobFlow design doc says update is not supported and should be blocked by webhook. This PR makes the webhook reject JobFlow `UPDATE` requests and adds the same check to the JobFlow validating admission policy.

Fixes #5723

`UPDATE` is still registered in the webhook rule so the API server sends update requests to admission. The handler rejects them there.

```release-note
JobFlow spec updates are now rejected by admission because JobFlow update is not supported.

### DIFF
--- a/installer/volcano-development-vap.yaml
+++ b/installer/volcano-development-vap.yaml
@@ -11382,7 +11382,12 @@ spec:
     - name: flowNames
       expression: "variables.flows.map(flow, flow.name)"
   validations:
-    # 1. Valid Dependency References - Dependencies must reference existing flows
+    # 1. JobFlow spec updates are not supported
+    - expression: request.operation != "UPDATE" || object.spec == oldObject.spec
+      message: "jobflow spec updates are not supported"
+      reason: Invalid
+
+    # 2. Valid Dependency References - Dependencies must reference existing flows
     - expression: |
         variables.flows.all(flow,
           !has(flow.dependsOn) ||
@@ -11392,7 +11397,7 @@ spec:
       message: "jobflow Flow is not DAG: vertex is not defined"
       reason: Invalid
 
-    # 2. No Self-Dependencies - A flow cannot depend on itself
+    # 3. No Self-Dependencies - A flow cannot depend on itself
     - expression: |
         variables.flows.all(flow,
           !has(flow.dependsOn) ||
--- a/pkg/webhooks/admission/jobflows/validate/validate_jobflow.go
+++ b/pkg/webhooks/admission/jobflows/validate/validate_jobflow.go
@@ -22,6 +22,7 @@ import (
 
 	admissionv1 "k8s.io/api/admission/v1"
 	whv1 "k8s.io/api/admissionregistration/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: pkg/webhooks/admission/jobflows/validate/validate_jobflow.go [hunk 1]
```diff
@@ -78,10 +78,10 @@ func AdmitJobFlows(ar admissionv1.AdmissionReview) *admissionv1.AdmissionRespons
...
 
 	switch ar.Request.Operation {
-	case admissionv1.Create, admissionv1.Update:
+	case admissionv1.Create:
 		msg = validateJobFlowDAG(jobFlow, &reviewResponse)
 	default:
-		err := OperationNotCreateOrUpdate
+		err := OperationNotCreate
 		return util.ToAdmissionResponse(err)
 	}
```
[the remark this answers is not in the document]
avinxshKD: i moved the operation check before decoding and simplified the update test since UPDATE now returns before reading the object.

avinxshKD:
Ptal @JesseStutler @hwdef @FAUST-BENCHOU

installer/volcano-development-vap.yaml
```diff
@@ -11382,7 +11382,12 @@ spec:
     - name: flowNames
       expression: "variables.flows.map(flow, flow.name)"
   validations:
-    # 1. Valid Dependency References - Dependencies must reference existing flows
+    # 1. JobFlow updates are not supported
```
hwdef: I think we should only refuse to update the `spec`, while still allowing users to update the `metadata`.

avinxshKD: Hey agreed, i updated this to only reject JobFlow spec changes on UPDATE. Metadata-only updates are allowed now.

Also updated the webhook/VAP checks and tests for both paths. PTAL :p

hwdef (commented):
/lgtm

avinxshKD:
@hwdef thanks for the quick review, much appreciated :).

cc @JesseStutler

avinxshKD:
same here sir ptal when get chance @JesseStutler

avinxshKD:
@JesseStutler sir as its labelled as lgtm,ptal if we can merge this :).

PR #5722, opened 2026-07-21
vnpu: remove poststart hook

DSFans2014:
#### What type of PR is this?

/cleanup

#5462 has added support for Ascend soft partitioning via `hami-vnpu-core`. Since `ascend-device-plugin` v1.4.0, the `limiter` process no longer needs to be started separately.  Therefore, the posthook is no longer required.

Fixes #

```release-note

```


archlitchi (approved):
/lgtm

PR #5729, opened 2026-07-22
fix(scheduler): clear job metrics on podgroup delete

### LINKED ISSUE #4705: Volcano metrics related to the job were not properly recycled
### Description

I am conducting stress tests on Volcano. I simulated thousands of nodes using Kwok and created thousands of Pods. However, after I deleted all the Pods, the job-related metrics in the Volcano scheduler did not reset to zero.

Image

1. Create a batch of Pods
2. Quickly delete this batch of Pods
3. Observe that metrics such as `unschedule_task_count` are not cleared

Although I deleted 4096 pending Pods, there are still over 3,000 Pods remaining in the metrics.  

v1.13.0

I examined the logic for cleaning up the metrics:

```go
func (sc *SchedulerCache) processCleanupJob() {
	job, shutdown := sc.DeletedJobs.Get()
	if shutdown {
		return
	}

	defer sc.DeletedJobs.Done(job)

	sc.Mutex.Lock()
	defer sc.Mutex.Unlock()

	if schedulingapi.JobTerminated(job) {
		oldJob, found := sc.Jobs[job.UID]
		if !found {
			klog.V(3).Infof("Failed to find Job <%v:%v/%v>, ignore it", job.UID, job.Namespace, job.Name)
			sc.DeletedJobs.Forget(job)
			return
		}
		newPgVersion := oldJob.PgUID
		oldPgVersion := job.PgUID
		klog.V(5).Infof("Just add pguid:%v, try to delete pguid:%v", newPgVersion, oldPgVersion)
		if oldPgVersion == newPgVersion {
			delete(sc.Jobs, job.UID)
			metrics.DeleteJobMetrics(job.Name, string(job.Queue), job.Namespace)
			klog.V(3).Infof("Job <%v:%v/%v> was deleted.", job.UID, job.Namespace, job.Name)
		}
		sc.DeletedJobs.Forget(job)
	} else {
		// Retry
		sc.retryDeleteJob(job)
	}
}
```

bhuvan-somisetty:
Job related metrics in the scheduler cache are only cleared through processCleanupJob, and that path only fires once JobTerminated is true, meaning the job has no PodGroup and no tasks left. When a PodGroup is deleted while the job still has tasks lingering in the cache (a burst of pod deletions under load, pods removed out of order, etc), the job never reaches that state through the normal path and its metrics just sit there. unschedule_task_count, job_share, job_retry_counts and the e2e scheduling timing gauges stick around indefinitely unless the scheduler restarts, since restart is the only thing that resets the in memory registry.

deletePodGroup already unsets the job's PodGroup and queues it for cleanup, so it already has the definitive signal that this PodGroup is gone for good. This adds a direct call to metrics.DeleteJobMetrics right there, so metrics get cleared immediately on the PodGroup delete event instead of waiting on the async queue to eventually converge. processCleanupJob is untouched, this is purely additive and doesn't change its behavior or the sc.Jobs bookkeeping.

Tested with a new unit test in pkg/scheduler/metrics covering DeleteJobMetrics directly (sets each metric to a nonzero value, deletes, confirms all six are reset). Also cross compiled pkg/scheduler/cache and pkg/scheduler/metrics for linux and ran golangci-lint against both packages, both clean.

Fixes #4705

### DIFF
--- a/pkg/scheduler/cache/event_handlers.go
+++ b/pkg/scheduler/cache/event_handlers.go
@@ -893,6 +893,13 @@ func (sc *SchedulerCache) deletePodGroup(id schedulingapi.JobID) error {
 
 	sc.deleteJob(job)
 
+	// The PodGroup is gone for good at this point, so its metrics are stale
+	// regardless of whether the job still has tasks lingering in the cache.
+	// Clean them up here instead of relying solely on processCleanupJob,
+	// which only fires once the job has no tasks left and can otherwise
+	// leave metrics behind indefinitely for jobs that never reach that state.
+	metrics.DeleteJobMetrics(job.Name, string(job.Queue), job.Namespace)
+
 	return nil
 }
 
--- a/pkg/scheduler/cache/event_handlers_test.go
+++ b/pkg/scheduler/cache/event_handlers_test.go
@@ -21,7 +21,9 @@ import (
 	"log"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


bhuvan-somisetty:
@JesseStutler @hajnalmt would appreciate a look here when you have a moment, thanks!

bhuvan-somisetty:
Good catch, updated the test to assert pre-delete values for all six metrics (including the e2e gauges and jobRetryCount) with a fixed timestamp instead of time.Now() for determinism.

bhuvan-somisetty:
Added TestSchedulerCache_DeletePodGroupV1beta1_ClearsJobMetrics covering exactly that. Since the metric vars in pkg/scheduler/metrics are private, the test reads the default prometheus gatherer directly and checks that the specific label-matched series for e2e_job_scheduling_duration and unschedule_task_count are actually removed after DeletePodGroupV1beta1, not just present with a zero value.

bhuvan-somisetty:
Good point, WithLabelValues was recreating a fresh zero-value series on read regardless of whether the delete actually happened, so the post-delete checks couldn't catch a real regression. Switched to testutil.CollectAndCount on each vec, matching the pattern already used in queue_scalar_test.go, so it now checks the series is actually gone rather than reading a value that gets silently recreated.

PR #5730, opened 2026-07-22
fix: populate job.Status.Succeeded for retained pods on job completion

### LINKED ISSUE #4616: job.status.succeeded Is Not Correctly Populated.

bhuvan-somisetty:
#### What type of PR is this?

/kind bug

When `killPods` is invoked (such as during job completion), pods retained based on `podRetainPhase` are excluded from `podsToKill`. Previously, these retained pods were skipped during status calculation in `killPods`, leaving `job.Status.Succeeded` and `job.Status.Failed` unpopulated (remaining 0).

This PR updates `killPods` to classify and count retained pods that are not targeted for deletion, ensuring that `job.Status.Succeeded`, `job.Status.Failed`, and task status counts are accurately updated upon job completion. Added unit tests in `job_controller_actions_test.go` to verify this behavior.

Fixes #4616

NONE

```release-note
Fix an issue where job.Status.Succeeded and job.Status.Failed were not populated upon job completion.
```

### DIFF
--- a/pkg/controllers/job/job_controller_actions.go
+++ b/pkg/controllers/job/job_controller_actions.go
@@ -103,6 +103,9 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 	var total int
 
 	podsToKill := make(map[string]*v1.Pod)
+	// retainedPods holds pods that are kept alive (not sent to kill) and need
+	// their phase counted into the job status.
+	retainedPods := make(map[string]*v1.Pod)
 
 	if target != nil {
 		switch target.Type {
@@ -172,6 +175,9 @@ func (cc *jobcontroller) killPods(jobInfo *apis.JobInfo, podRetainPhase state.Ph
 
 				if !retain {
 					podsToKill[pod.Name] = pod
+				} else {
+					// Pod is retained; record it so we can update job status below.
+					retainedPods[pod.Name] = pod
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Updated the commit message to remove issue keywords/hashtags as required by the commit lint check, and updated unit tests to include task-level status assertions.

bhuvan-somisetty:
Fixed the two points raised in review: removed unrelated scheduler metrics cleanup code to keep the PR focused on the job status issue, and added assertions for both Status.Failed and TaskStatusCount in the unit test to cover all status counters updated by the fix.

bhuvan-somisetty:
Fixed the double-counting issue raised in review: previously the retained-pod loop re-scanned all jobInfo.Pods and recounted terminating pods that were already incremented in the initial target==nil pass. Now retained pods are tracked explicitly during the first scan via a retainedPods map, so the status classification only runs on that set — no double-counting of terminating pods.

bhuvan-somisetty:
Hi @JesseStutler @hajnalmt @hwdef @archlitchi - could you please take a look at this PR when you have a moment? All CI checks are passing and review feedback has been addressed. Thanks!

PR #5733, opened 2026-07-23
test: add kube-scheduler plugin extension compatibility guardrail

### LINKED ISSUE #5329: Support DRA Score/NormalizeScore and add guardrails for kube-scheduler plugin extension changes
## What is the problem you're trying to solve

When DRA support was introduced in Volcano, the reused kube-scheduler DRA plugin only needed to be adapted for extension points such as `PreFilter`, `Filter`, `Reserve`, and `PreBind`.

In Kubernetes v1.34, the DRA plugin added scoring support through `Score` and `NormalizeScore`. These extension points are not wired into Volcano yet, so DRA scoring behavior may be different from kube-scheduler after upgrading the Kubernetes dependency.

There is also a broader maintenance problem. Some Volcano plugins reuse kube-scheduler in-tree plugins, but Volcano does not always support every kube-scheduler extension point directly. For example, extension points like `PreEnqueue` or `Sign` may not map cleanly to Volcano's scheduling flow. When a reused kube-scheduler plugin adds a new extension point in a future Kubernetes release, it is easy to miss that during dependency upgrades.

## Describe the solution you'd like

First, support the DRA `Score` and `NormalizeScore` path in Volcano's predicates plugin. DRA should be registered as a score plugin when enabled, and the score helper should call the plugin's own `ScoreExtensions()` so `NormalizeScore` can run when provided.

Second, add an automated guardrail for kube-scheduler plugin extension point changes. The idea is to maintain an explicit decision list for each kube-scheduler plugin reused by Volcano:

- extension points adapted by Volcano
- extension points intentionally ignored because they do not fit Volcano's scheduling flow

A test or verify script can instantiate or inspect the reused kube-scheduler plugin and detect which kube-scheduler framework interfaces it currently implements. If the plugin implements an extension point that is neither adapted nor intentionally ignored, the check should fail and ask maintainers to make an explicit decision.

For example, after a Kubernetes dependency upgrade, if DRA starts implementing a new extension point, CI should report something like:

```text
DynamicResources implements Score, but Volcano has not marked it as adapted or intentionally ignored.
Please decide whether this extension point should be supported in Volcano.
```

This would help catch behavior changes during Kubernetes version upgrades without forcing Volcano to implement every kube-scheduler extension point.

## Additional context

This is mainly about plugins imported from kube-scheduler, such as DynamicResources/DRA and VolumeBinding.

For extension points that Volcano does not intend to support, the goal is not to wire them in automatically. The goal is to make the decision explicit and visible during dependency upgrades.

miantalha45:
#### What type of PR is this?

/kind feature

Adds a test-only compatibility manifest for upstream kube-scheduler plugin types that Volcano directly creates and uses.

For each reused plugin, the manifest explicitly records every supported scheduler extension point as either:

- adapted by Volcano, or
- intentionally ignored, with a reason.

The test fails when:

- an upstream plugin implements a known extension point that Volcano has not classified, or
- the manifest declares an extension point that the upstream plugin no longer implements.

This prevents Kubernetes dependency upgrades from silently changing upstream plugin behavior without an explicit Volcano compatibility decision.

This is the second PR for #5329. The first PR adds DRA `Score` and `NormalizeScore` support.

Fixes #5329

The guardrail covers all upstream kube-scheduler plugin implementations directly instantiated by Volcano’s `predicates` and `nodeorder` plugins.

It intentionally excludes upstream helper infrastructure such as the DRA manager, CSI manager, and VolumeBinder because they are not scheduler extension plugins.

Validated with:

```bash
go test ./pkg/scheduler/plugins ./pkg/scheduler/plugins/predicates ./pkg/scheduler/plugins/nodeorder

### DIFF
--- a/pkg/scheduler/plugins/upstream_extensions_test.go
+++ b/pkg/scheduler/plugins/upstream_extensions_test.go
@@ -1,0 +1,275 @@
+/*
+Copyright 2026 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugins
+
+import (
+	"context"
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	fwk "k8s.io/kube-scheduler/framework"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/imagelocality"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/interpodaffinity"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodeaffinity"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodeports"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/noderesources"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodeunschedulable"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodevolumelimits"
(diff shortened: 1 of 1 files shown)


miantalha45:
@JesseStutler Score/NormalizeScore implementation is done as mentioned in issue #5329, now what about guard-rails implementation? As you mentioned earlier, we have to discuss about it.

JesseStutler:
Honestly, I think this will be somewhat customized. Some extensions are originally unnecessary for Volcano, so I think I'll lower the priority of this

miantalha45:
Ok @JesseStutler , As you wish.

PR #5735, opened 2026-07-23
Fix PodGroup tombstone deletion

### LINKED ISSUE #5734: Queue can stay in Closing after a PodGroup tombstone delete
### Description

Found through source review.

When a PodGroup deletion is delivered as DeletedFinalStateUnknown, the queue controller unwraps the PodGroup but calculates the cache key from the tombstone object. The key lookup fails and the error is ignored, leaving the PodGroup in the controller cache.

This can leave a Queue in `Closing` until another event triggers reconciliation or the controller restarts.

1. Add a PodGroup to the queue controller cache as `default/pg1`.
2. Pass its deletion as a tombstone:

   ```go
   c.deletePodGroup(cache.DeletedFinalStateUnknown{
       Key: "default/pg1",
       Obj: pg,
   })

3. Check c.podGroups[pg.Spec.Queue].
    The default/pg1 entry is still present.
    A focused regression test fails with:

```
--- FAIL: TestDeletePodGroupTombstone
    podgroup "default/pg1" was not removed after tombstone delete
```

This has not been reproduced on a live cluster. It was reproduced with a focused controller test and verified against the normal informer deletion path.

Client-go can deliver DeletedFinalStateUnknown when a watch disconnects, misses a delete event, and later detects the deletion during relist:

https://github.com/kubernetes/client-go/blob/v0.36.1/tools/cache/controller.go#L275-L278

Volcano unwraps the tombstone, but then calls `MetaNamespaceKeyFunc` on the original wrapper and ignores the returned error:

https://github.com/volcano-sh/volcano/blob/4872fc71617e21c311bb2b917f4e4ad78cad30fd/pkg/controllers/queue/queue_controller_handler.go#L108-L137

The handler still enqueues the Queue. During sync, the stale key is removed after the PodGroup lookup returns `NotFound`, but the old `podGroups` slice is passed to the state transition:

https://github.com/volcano-sh/volcano/blob/4872fc71617e21c311bb2b917f4e4ad78cad30fd/pkg/controllers/queue/queue_controller_action.go#L52-L99

The closing state sees the stale non-empty list and keeps the Queue in `Closing`:

https://github.com/volcano-sh/volcano/blob/4872fc71617e21c311bb2b917f4e4ad78cad30fd/pkg/controllers/queue/state/closing.go#L37-L65

Using DeletionHandlingMetaNamespaceKeyFunc removes the correct key and makes the regression test pass

Actual: the deleted PodGroup remains in the queue controller cache. A Queue waiting for that PodGroup to disappear can remain in `Closing`.

Expected: tombstone deletion should remove the `namespace/name` key and allow the Queue to transition to `Closed`.

master at 4872fc71617e21c311bb2b917f4e4ad78cad30fd

Have reproduced with Go 1.26.4 on Windows/amd64.

The likely fix is to use cache.DeletionHandlingMetaNamespaceKeyFunc in deletePodGroup.

avinxshKD:
#### What type of PR is this?

/kind bug

Handles PodGroup tombstone deletes with `DeletionHandlingMetaNamespaceKeyFunc` so the queue cache removes the correct key.

Fixes #5734

Codex was used to help prepare this change.

Prompt: Fix #5734 with the smallest code change, reuse the existing helper, add regression coverage, and run the queue-controller tests.

Volcano burns bright; stale queues close right.

Tested with:

`
go test ./pkg/controllers/queue -count=1`

`
Fix Queue reconciliation after a PodGroup tombstone deletion.`

### DIFF
--- a/pkg/controllers/queue/queue_controller_handler.go
+++ b/pkg/controllers/queue/queue_controller_handler.go
@@ -121,7 +121,7 @@ func (c *queuecontroller) deletePodGroup(obj interface{}) {
 		}
 	}
 
-	key, _ := cache.MetaNamespaceKeyFunc(obj)
+	key, _ := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 
 	c.pgMutex.Lock()
 	defer c.pgMutex.Unlock()
--- a/pkg/controllers/queue/queue_controller_test.go
+++ b/pkg/controllers/queue/queue_controller_test.go
@@ -188,6 +188,12 @@ func TestDeletePodGroup(t *testing.T) {
 		if _, ok := c.podGroups[testcase.podGroup.Spec.Queue][key]; ok != testcase.ExpectValue {
 			t.Errorf("case %d (%s): expected: %v, got %v ", i, testcase.Name, testcase.ExpectValue, ok)
 		}
+
+		c.podGroups[testcase.podGroup.Spec.Queue][key] = struct{}{}
+		c.deletePodGroup(cache.DeletedFinalStateUnknown{Key: key, Obj: testcase.podGroup})
+		if _, ok := c.podGroups[testcase.podGroup.Spec.Queue][key]; ok != testcase.ExpectValue {
+			t.Errorf("case %d (%s) tombstone: expected: %v, got %v ", i, testcase.Name, testcase.ExpectValue, ok)
+		}
 	}
 }
 


avinxshKD:
ptal when get chance @JesseStutler @hwdef @hajnalmt

hwdef (commented):
/lgtm

Issue #5740, opened 2026-07-24
Webhook validation mutates old object as side effect


karthik120710:
## Description: 
During job update validation (validateJobUpdate), both new and old objects are normalized by clearing VolumeClaimName when VolumeClaim is set, so a subsequent DeepEqual comparison correctly ignores the auto-filled field. Mutating old is unnecessary — only new needs normalization before the comparison. Currently, the webhook handler decodes fresh copies from the admission request JSON, so the mutation is technically harmless. But this violates the principle that validating webhooks should not mutate their inputs and could corrupt the API server's copy if the call chain ever changes.

PR #5741, opened 2026-07-24
fix : remove unnecessary mutation of old job object in webhook validation

karthik120710:
## Description: 
During job update validation in validateJobUpdate, both new and old job objects were being normalized by clearing VolumeClaimName when VolumeClaim is set. Mutating the old object is unnecessary since only the new object needs normalization for the DeepEqual comparison. This violates the principle that validating webhooks should not mutate their inputs and could potentially corrupt the API server's copy if the calling convention changes. This fix removes the unnecessary mutation of the old object, keeping only the new object normalization.
#5740


hwdef:
/ok-to-test

After the controller writes the generated `VolumeClaimName` back to the Job, both `old` and `new` contain that name. Clearing it only from `new` makes `DeepEqual` fail and rejects valid updates.
Please deep-copy both specs and clear `VolumeClaimName` from both copies before comparing them.

PR #5747, opened 2026-07-26
fix: handle node tombstones in label discovery

### LINKED ISSUE #5745: Label-based HyperNode discovery misses node deletions delivered as tombstones
### Description

Found through source review and reproduced at the informer handler.

Label-based discovery does not handle `DeletedFinalStateUnknown` in `DeleteNode`. If a Node deletion is delivered as a tombstone, the object is rejected and topology discovery is not queued.

This can leave the deleted Node in generated HyperNodes until another relevant Node or HyperNode event triggers reconciliation.

1. Enable label-based HyperNode discovery and add a Node with the configured topology labels.
2. Wait for the corresponding HyperNode to be generated.
3. Deliver the Node delete event as `cache.DeletedFinalStateUnknown`. This is the normal client-go path when a delete is missed during a watch disconnect.
4. Check the discovery queue and generated HyperNode.

A minimal handler reproducer is:

```

discoverer.DeleteNode(cache.DeletedFinalStateUnknown{
    Key: "node-1",
    Obj: labeledNode,
})
```

The Node informer registers `DeleteNode` directly as its delete handler:

https://github.com/volcano-sh/volcano/blob/8a2c34cf264552faac0e1914c50041baf7d2270a/pkg/controllers/hypernode/discovery/label/label.go#L91-L98

`DeleteNode` passes the event object directly to `getNodeNetworkTopologyLabels`:

https://github.com/volcano-sh/volcano/blob/8a2c34cf264552faac0e1914c50041baf7d2270a/pkg/controllers/hypernode/discovery/label/label.go#L310-L316

That helper only accepts `*v1.Node`. A tombstone returns an empty label map, so `enqueue()` is skipped:

https://github.com/volcano-sh/volcano/blob/8a2c34cf264552faac0e1914c50041baf7d2270a/pkg/controllers/hypernode/discovery/label/label.go#L337-L348

Client-go documents that delete handlers receive `DeletedFinalStateUnknown` when a delete event is missed during a watch disconnect:

https://pkg.go.dev/k8s.io/client-go/tools/cache#DeletedFinalStateUnknown

The neighboring `DeleteHyperNode` handler already unwraps the same tombstone type:

https://github.com/volcano-sh/volcano/blob/8a2c34cf264552faac0e1914c50041baf7d2270a/pkg/controllers/hypernode/discovery/label/label.go#L318-L335

The handler was introduced with label-based discovery in:

https://github.com/volcano-sh/volcano/commit/3346a39bd381bdbbac66f546b11aa1819c1fa532

Focused reproducer output:

```
Cannot convert to *v1.Node
expected node tombstone to trigger topology discovery, queue length is 0
```

Actual: the tombstone is rejected, no discovery is queued, and stale Node membership can remain in the generated HyperNodes.

Expected: `DeleteNode` should unwrap the Node from `DeletedFinalStateUnknown` and enqueue topology discovery.

master (8a2c34cf264552faac0e1914c50041baf7d2270a)

Reproduced with a focused Go test using client-go v0.36.1. This was not observed against a release image in a live cluster.

avinxshKD:
#### What type of PR is this?

/kind bug

Handles Node delete events delivered as `DeletedFinalStateUnknown` in label-based HyperNode discovery.

Previously, the tombstone wrapper was passed to a helper that only accepts `*v1.Node`. This returned an empty label map and skipped topology reconciliation, potentially leaving stale Node membership in generated HyperNodes.

The change follows the existing `DeleteHyperNode` tombstone handling and adds a focused regression test.

Fixes #5745

Tests run:

```text
go test ./pkg/controllers/hypernode/discovery/label -run TestDeleteNodeWithTombstone -count=1
go test ./pkg/controllers/hypernode/...
git diff --check
```

AI assistance disclosure: Codex was used for source review, drafting the initial fix and test, and pre-submit review. I reviewed every change and ran the tests listed above on my own.

```release-note
Fix label-based HyperNode discovery so Node deletions delivered as informer tombstones trigger topology reconciliation.
```

### DIFF
--- a/pkg/controllers/hypernode/discovery/label/label.go
+++ b/pkg/controllers/hypernode/discovery/label/label.go
@@ -309,7 +309,21 @@ func (l *labelDiscoverer) UpdateNode(oldObj, newObj interface{}) {
 
 // DeleteNode Reconstruct the hyperNode when the node changes.
 func (l *labelDiscoverer) DeleteNode(obj interface{}) {
-	labelMap := l.getNodeNetworkTopologyLabels(obj)
+	node, ok := obj.(*v1.Node)
+	if !ok {
+		// If we reached here it means the Node was deleted but its final state is unrecorded.
+		tombstones, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			klog.Errorf("Couldn't get object from tombstone %#v", obj)
+			return
+		}
+		node, ok = tombstones.Obj.(*v1.Node)
+		if !ok {
+			klog.Errorf("Tombstone contained object that is not a Node: %#v", obj)
+			return
+		}
+	}
+	labelMap := l.getNodeNetworkTopologyLabels(node)
 	if len(labelMap) > 0 {
 		l.enqueue()
 	}
--- a/pkg/controllers/hypernode/discovery/label/label_test.go
+++ b/pkg/controllers/hypernode/discovery/label/label_test.go
@@ -26,6 +26,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
 	topologyv1alpha1 "volcano.sh/apis/pkg/apis/topology/v1alpha1"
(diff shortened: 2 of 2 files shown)


avinxshKD:
@hajnalmt Pls take a look when you get a chance. This fixes the missed Node tombstone path in label-based HyperNode discovery :p.


/cc @JesseStutler @hwdef

avinxshKD:
@JesseStutler @hajnalmt  as its labelled as lgtm, ptal if we can merge this :).

PR #5773, opened 2026-07-29
job controller: skip stale cache cleanup for recreated jobs

### LINKED ISSUE #3968: retryDeleteJob causes restart job to be unscheduled
### Description

image

1. first delete Podgroup and delete Pod triggered `retryDeleteJob` called a few times and was penalized with a delay of a few seconds
2. at this point the `deleteJob` was called directly out of the queue, removing the job from the jobcache
3. At this point, the job retry or re-create a job with the same UID.
4. Then after a few seconds the old job is queued up from the ratelimit queue, and at this time to determine the JobTerminated = true, it will be deleted from the job cache again!

Then the job will always exist and never be scheduled.

Expect not to have to delete the cache repeatedly for this scenario

volcano 1.9

_No response_

bhuvan-somisetty:
#### What type of PR is this?

/kind bug

The job cache removes a job by pushing it onto a rate-limited work queue
and retrying with backoff (`retryDeleteJob`) until all of its pods are
gone. If a job is deleted and quickly recreated under the same
namespace/name, an earlier retry scheduled for the old generation can
still fire after the new job has already been re-added to the cache.

`processCleanupJob` only looked the cache entry up by key before
deleting it:

```go
if jobTerminated(job) {
    jc.deletedJobs.Forget(job)
    key := keyFn(job.Namespace, job.Name)
    delete(jc.jobs, key)
    ...
}
```

Because the lookup is by key and not by the specific `JobInfo` the
retry was scheduled for, a late-firing retry belonging to the old,
already-cleaned-up generation ends up deleting whatever is currently
cached under that key, even if it's the live, recreated job. Once that
happens the job's cache entry is gone and it never gets scheduled
again.

This PR checks that the queued `JobInfo` is still the entry currently
cached for that key before removing it. If the cache has moved on to a
newer generation, the stale retry is simply forgotten instead of
evicting the live entry.

Added a unit test that reproduces the race directly: a stale,
already-terminated `JobInfo` is queued for cleanup while a different,
live `JobInfo` occupies the same key (simulating the recreate), and
asserts the live entry survives `processCleanupJob`.

Fixes #3968

The fix is scoped to `processCleanupJob` in `pkg/controllers/cache/cache.go`
and doesn't change the cache's external behavior for the normal
(non-racing) delete path.

```release-note
Fix a race in the job controller cache where a stale, rate-limited cleanup retry from a deleted job could evict the cache entry of a job recreated with the same name, leaving it stuck and never scheduled.
```

### DIFF
--- a/pkg/controllers/cache/cache.go
+++ b/pkg/controllers/cache/cache.go
@@ -371,9 +371,20 @@ func (jc *jobCache) processCleanupJob() bool {
 	jc.Mutex.Lock()
 	defer jc.Mutex.Unlock()
 
+	key := keyFn(job.Namespace, job.Name)
+	// A rate-limited cleanup task can outlive the JobInfo it was scheduled
+	// for: if the job is deleted and recreated under the same key while a
+	// previous, already-scheduled retry is still pending, that stale retry
+	// must not be allowed to evict the cache entry of the recreated job.
+	// Only proceed if this task's JobInfo is still the one currently cached
+	// for this key.
+	if current, found := jc.jobs[key]; !found || current != job {
+		jc.deletedJobs.Forget(job)
+		return true
+	}
+
 	if jobTerminated(job) {
 		jc.deletedJobs.Forget(job)
-		key := keyFn(job.Namespace, job.Name)
 		delete(jc.jobs, key)
 		klog.V(3).Infof("Job <%s> was deleted.", key)
 	} else {
--- a/pkg/controllers/cache/cache_test.go
+++ b/pkg/controllers/cache/cache_test.go
@@ -394,6 +394,50 @@ func TestJobCache_Delete(t *testing.T) {
 	}
 }
 
+func TestJobCache_ProcessCleanupJob_SkipsStaleGeneration(t *testing.T) {
+	namespace := "test"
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
@JesseStutler @hajnalmt @hwdef could one of you take a look when you get a chance? This fixes a real job-cache race (#3968) where a stale rate-limited cleanup retry can end up deleting the cache entry of a job that was recreated under the same name, so it never gets scheduled again. Added a unit test that reproduces it directly. Happy to adjust anything.

hwdef:
/lgtm
I'm not sure if this is a real bug, or if it might occur in real-world scenarios.

bhuvan-somisetty:
Thanks for the lgtm @hwdef!

It does happen in practice whenever a job is deleted and quickly recreated with the same name (think CI pipelines resubmitting jobs, or an operator restarting a failed job), the old cleanup retry can still be in-flight. Since processCleanupJob retries until all pods are gone, and pod termination can take a few seconds, the stale retry fires after the new job is already in the cache and wipes it out by key leaving the job invisible to the controller and never scheduled.

The unit test in the PR reproduces this directly.

PR #5774, opened 2026-07-29
fix(predicates): enforce ReadWriteOncePod volume restrictions

### LINKED ISSUE #5767: predicates: ReadWriteOncePod PVC exclusivity is not enforced
### Description

Found during source review.

Volcano does not register the Kubernetes `VolumeRestrictions` plugin. A second pod using an in-use `ReadWriteOncePod` PVC can therefore pass predicates and be bound, leaving CSI/kubelet to reject the mount later.

This breaks the expected cluster-wide single-pod guarantee for RWOP volumes.

1. Create a CSI PVC with `accessModes: [ReadWriteOncePod]`.
2. Start a pod with `schedulerName: volcano` that mounts the PVC.
3. Once it is running, create another Volcano-scheduled pod using the same PVC.
4. Check the second pod's `spec.nodeName` and events.

Checked at `d40caf9d3cb605df0babbb84e0221dd2c2e0b0a9`.

- Scheduling enters the predicates plugin here: https://github.com/volcano-sh/volcano/blob/d40caf9d3cb605df0babbb84e0221dd2c2e0b0a9/pkg/scheduler/plugins/predicates/predicates.go#L313-L326
- Only registered PreFilter/Filter plugins are executed: https://github.com/volcano-sh/volcano/blob/d40caf9d3cb605df0babbb84e0221dd2c2e0b0a9/pkg/scheduler/plugins/predicates/predicates.go#L446-L467 and https://github.com/volcano-sh/volcano/blob/d40caf9d3cb605df0babbb84e0221dd2c2e0b0a9/pkg/scheduler/plugins/predicates/predicates.go#L731-L748
- The plugin list includes other volume checks but not `VolumeRestrictions`: https://github.com/volcano-sh/volcano/blob/d40caf9d3cb605df0babbb84e0221dd2c2e0b0a9/pkg/scheduler/plugins/predicates/predicates.go#L513-L637
- The storage lister method required by that plugin is also unimplemented: https://github.com/volcano-sh/volcano/blob/d40caf9d3cb605df0babbb84e0221dd2c2e0b0a9/pkg/scheduler/plugins/util/k8s/snapshot.go#L418-L420

Kubernetes performs the RWOP conflict check in `VolumeRestrictions`:
https://github.com/kubernetes/kubernetes/blob/v1.36.1/pkg/scheduler/framework/plugins/volumerestrictions/volume_restrictions.go#L173-L245

Expected: the second pod remains pending with a `FailedScheduling` event because the RWOP PVC is already in use.

Actual from the current code path: there is no RWOP conflict check, so the pod can pass predicates and be bound. Any failure happens later during CSI attach/mount.

master @ d40caf9d3cb605df0babbb84e0221dd2c2e0b0a9

Applies to Kubernetes 1.29+ with a CSI driver supporting `ReadWriteOncePod`.

avinxshKD:
#### What type of PR is this?

/kind bug

Registers the Kubernetes `VolumeRestrictions` plugin in predicates and keeps its state updated during preemption simulation. This prevents a second pod from being scheduled with an in-use `ReadWriteOncePod` PVC.

Fixes #5767

Tested the affected predicate and snapshot packages on Linux.

AI tooling was used during implementation. I reviewed the complete diff and verified the tests locally on my own.

```release-note
Enforce ReadWriteOncePod PVC exclusivity during predicate checks.

### DIFF
--- a/pkg/scheduler/plugins/predicates/predicates.go
+++ b/pkg/scheduler/plugins/predicates/predicates.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/podtopologyspread"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/tainttoleration"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumebinding"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumerestrictions"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumezone"
 
 	"volcano.sh/volcano/pkg/scheduler/api"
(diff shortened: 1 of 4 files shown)


avinxshKD:
@hwdef ptal, this registers VolumeRestrictions and keeps its state correct during preemption simulation. also added regression coverage for both normal scheduling and preemption. PTAL

hwdef:
> @hwdef ptal, this registers VolumeRestrictions and keeps its state correct during preemption simulation. also added regression coverage for both normal scheduling and preemption. PTAL

Sorry, I don't know much about this area.

PR #5775, opened 2026-07-29
fix(scheduler): discard partially recovered operations when RecoverOperations fails

### LINKED ISSUE #5362: Missing Statement.Discard() on RecoverOperations failure in allocateForJob/allocateForSubJob causes session state corruption
### Description

In `pkg/scheduler/actions/allocate/allocate.go`, both `allocateForJob` and `allocateForSubJob` fail to call `finalStmt.Discard()` when `RecoverOperations` returns an error. This leaves partially-applied operations (Evict, Pipeline, Allocate) orphaned in session state, corrupting node resource accounting and task statuses for the rest of the scheduling cycle.
  
  `RecoverOperations` applies saved operations one-by-one. If it fails on the Nth operation (e.g., a plugin's `AllocateFunc` event handler returns an error), operations 1 through N-1 have already mutated session state (node.AddTask, job.UpdateTaskStatus, event handler callbacks) and been appended to `finalStmt.operations`. The function then returns `nil` without calling `Discard()`, so those mutations are never rolled back.
  
  **In `allocateForJob`:**
  
  finalStmt := framework.NewStatement(ssn)
  if err = finalStmt.RecoverOperations(bestStmt); err != nil {
      klog.ErrorS(err, "Failed to recover operations", "job", job.UID, "hyperNode", bestHyperNode)
      return nil  // BUG: no finalStmt.Discard() — partial ops orphaned in session
  }

 **In `allocateForSubJob`:**
  
  finalStmt := framework.NewStatement(ssn)
  if err = finalStmt.RecoverOperations(bestStmt); err != nil {
      klog.ErrorS(err, "Failed to recover operations", "subJob", subJob.UID, "hyperNode", bestHyperNode)
      return nil, 0  // BUG: no finalStmt.Discard()
  }

1. Enable topology-aware scheduling (HyperNodes) with plugins that register AllocateFunc event handlers (e.g., DRA + capacity)
2. Submit a job with hard network topology or SubJobPolicy (triggers the allocateForJob/allocateForSubJob path)
3. If a plugin's AllocateFunc errors during RecoverOperations (e.g., node removed between dry-run and recovery, or DRA claim goes stale), the partial session mutations are never rolled back
4. Remaining jobs in the same scheduling cycle see corrupted node Idle/Used values

Expected:
  
- When RecoverOperations fails, all partially-applied operations should be rolled back via finalStmt.Discard(), leaving session state clean for subsequent scheduling decisions.
  
Actual:
  
- Partially-applied operations (Evict/Pipeline/Allocate that succeeded before the failing operation) are left in session state. Nodes show reduced Idle resources, tasks are in Pipelined/Allocated status, but no statement reference exists to commit or discard them. This corrupts scheduling decisions for all remaining jobs in the current cycle.

v1.15.0

_No response_

SrikarChittemsetty:
### What type of PR is this?

/kind bug

### What this PR does / why we need it

`allocateForJob` and `allocateForSubJob` in `pkg/scheduler/actions/allocate/allocate.go` return early when `RecoverOperations` fails, without discarding the statement:

```go
finalStmt := framework.NewStatement(ssn)
if err = finalStmt.RecoverOperations(bestStmt); err != nil {
    klog.ErrorS(err, "Failed to recover operations", ...)
    return nil   // no finalStmt.Discard()
}
```

`RecoverOperations` replays saved operations one at a time through `Evict` / `Pipeline` / `Allocate`. Each one mutates session state (`node.AddTask`, `job.UpdateTaskStatus`, plugin event handlers) and is appended to the recovering statement's operations. So when it fails on the Nth operation, operations 1..N-1 are already applied — and returning without `Discard()` leaves them orphaned, with no statement reference left to commit or roll them back. Nodes show reduced `Idle`, tasks sit in `Allocated`/`Pipelined`, and every remaining job in that scheduling cycle schedules against corrupted accounting.

This adds `finalStmt.Discard()` on both paths, which is how every other early return in this file releases a statement (lines 429, 501, 633, 865).

### Which issue(s) this PR fixes

Fixes #5362

### Testing

Added `TestRecoverOperations_PartialFailureLeavesOperationsApplied` to `pkg/scheduler/framework`. It builds a plan whose second operation targets a node absent from the session, so replay fails after the first operation has been applied, and asserts that:

- `RecoverOperations` returns an error,
- the successfully replayed operation is still recorded on the statement,
- the node's `Idle` resources remain consumed,
- and only `Discard()` restores the node's resources and the task's `Pending` status.

That is the invariant the callers depend on, and it is what makes the missing `Discard()` a correctness bug rather than a cosmetic one.

`pkg/scheduler/framework` and `pkg/scheduler/actions/allocate` suites both pass; `gofmt` and `go vet` are clean.

### Note for reviewers

While writing the test I noticed that `RecoverOperations` resolves an `Allocate` op's node with `node := s.ssn.Nodes[op.task.NodeName]` and passes the result straight to `Allocate`, which dereferences `nodeInfo.Name`, so an unguarded map miss would segfault. I raised that separately in #5778/#5779 and have since closed both: it is not reachable through the current call graph, because `SaveOperations` stores `op.task.Clone()` and the replayed op keeps the `NodeName` it recorded. Noting it here only so the earlier cross-reference is not left dangling — it has no bearing on this PR.

Disclosure: this change was prepared with the assistance of an AI coding agent. The analysis, fix, and test were verified against the source and the existing suites before submitting.

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -446,6 +446,11 @@ func (alloc *Action) allocateForJob(job *api.JobInfo, jobWorksheet *JobWorksheet
 		finalStmt := framework.NewStatement(ssn)
 		if err = finalStmt.RecoverOperations(bestStmt); err != nil {
 			klog.ErrorS(err, "Failed to recover operations", "job", job.UID, "hyperNode", bestHyperNode)
+			// RecoverOperations replays operations one by one, so a failure part way
+			// through leaves the earlier ones applied to the session and recorded on
+			// finalStmt. Roll them back, otherwise they stay orphaned and corrupt node
+			// accounting and task statuses for the rest of the scheduling cycle.
+			finalStmt.Discard()
 			return nil
 		}
 
@@ -519,6 +524,9 @@ func (alloc *Action) allocateForSubJob(subJob *api.SubJobInfo, subJobWorksheet *
 		finalStmt := framework.NewStatement(ssn)
 		if err = finalStmt.RecoverOperations(bestStmt); err != nil {
 			klog.ErrorS(err, "Failed to recover operations", "subJob", subJob.UID, "hyperNode", bestHyperNode)
+			// See allocateForJob: partially applied operations must be rolled back or
+			// they remain orphaned in the session.
+			finalStmt.Discard()
 			return nil, 0
 		}
 		newAllocatedHyperNode := ssn.HyperNodes.GetLCAHyperNode(subJob.AllocatedHyperNode, bestHyperNode)
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -502,6 +502,11 @@ func (alloc *Action) allocateForJob(job *api.JobInfo, jobWorksheet *JobWorksheet
 		finalStmt := framework.NewStatement(ssn)
 		if err = finalStmt.RecoverOperations(bestStmt); err != nil {
 			klog.ErrorS(err, "Failed to recover operations", "job", job.UID, "hyperNode", bestHyperNode)
+			// RecoverOperations replays operations one by one, so a failure part way
+			// through leaves the earlier ones applied to the session and recorded on
+			// finalStmt. Roll them back, otherwise they stay orphaned and corrupt node
+			// accounting and task statuses for the rest of the scheduling cycle.
+			finalStmt.Discard()
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


SrikarChittemsetty:
Gentle ping — this has been open since late July with DCO green and no conflicts.

It addresses #5362 (reported by @ritup-klz): when `RecoverOperations` fails partway through, `allocateForJob`/`allocateForSubJob` return without discarding the partially applied statement, leaving the session with operations that were applied but never committed or rolled back. The change discards the statement on that path and adds a regression test.

@k82cn @dafu-wu when you get a chance. Happy to rebase if it has drifted.

PR #5777, opened 2026-07-29
fix(inqueue): add dequeue action to fix inqueue block

### LINKED ISSUE #5006: Inqueue PodGroups occupy queue resources even when pods cannot be scheduled

When a PodGroup is moved to `Inqueue` state, it immediately reserves resources from the queue's capacity. However, if the pods cannot be scheduled due to reasons like:
- Node affinity/anti-affinity constraints
- Insufficient resources per node (e.g., requesting 8 GPUs but max 4 GPUs/node)
- ...

The PodGroup will stay in `Inqueue` state indefinitely, blocking queue resources that could be used by other jobs.（Resource waste）

Implement an Inqueue timeout mechanism with the following behavior:

1. Queue-level default timeout (e.g., spec.inqueueTimeout: 10m
2. Record timestamp when PodGroup enters Inqueue state
3. Timeout action: When timeout is reached and no pods have been successfully scheduled, transition PodGroup back to Pending state
4.  Release reserved queue resources

### Additional context

_No response_

### Documentation Updates

Abingcbc:
#### What type of PR is this?

fix

An `Inqueue` Job can remain stuck indefinitely when it cannot make scheduling progress, for example because of insufficient node resources or affinity constraints. Although no task is successfully scheduled, the Job continues consuming the Queue's inqueue quota and can prevent other Pending Jobs from being admitted.

This PR introduces an optional `dequeue` scheduler action to address this problem.

The new action:

- Finds Jobs whose PodGroup is `Inqueue`, has Tasks, belongs to an existing Queue, and is neither `JobReady` nor `JobPipelined`.
- Releases partially reserved resources:
  - Unpipelines `Pipelined` Tasks.
  - Unallocates `Allocated` Tasks.
  - Restores Node resources, invokes deallocate handlers, and clears Task `NodeName`.
- Moves the PodGroup from `Inqueue` back to `Pending`.
- Adds the Job to a scheduler-local unschedulable Job cache for five minutes.
- Makes the `enqueue` action skip cached Jobs, preventing them from immediately reclaiming the Queue quota they just released.
- Automatically allows the Job to participate in enqueue again after the five-minute TTL expires.
- Removes cache entries when Jobs are deleted. Repeated additions do not extend an existing TTL.

Jobs remain part of scheduler snapshots while cached so that status updates and allocated-resource accounting continue to work correctly. Only admission by the `enqueue` action is skipped.

Fixes https://github.com/volcano-sh/volcano/issues/5006

We need to strike a balance between resource utilization and starvation of large tasks.

The action is opt-in and must be configured after all actions that can allocate, pipeline, preempt, or reclaim Tasks:

```yaml
actions: "enqueue, allocate, backfill, dequeue"
```

### DIFF
--- a/docs/user-guide/how_to_configure_scheduler.md
+++ b/docs/user-guide/how_to_configure_scheduler.md
@@ -26,7 +26,7 @@ requirement to decide how to combine these functions. That's why `tier` is requi
 ## Actions
 * `Action` implements the main logic of scheduling. 
 * Volcano allow users to make self-defined actions.
-* Volcano provides 7 built-in actions until April 2022. The details are as follows.
+* Volcano provides the following built-in actions.
 
 | ID  | Name     | Required | Description                                                                                                                                                                                                                                                           |
 |-----|----------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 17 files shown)


Abingcbc:
All E2E cases are passed
image

avinxshKD:
Pls rebase this branch onto master and remove the merge commit

JesseStutler:
This is an interesting direction for dequeue action, will take a look later, @Abingcbc Thanks for your contribution. We might need to reach an agreement, I also have some work to do here.

PR #5779, opened 2026-07-29
fix(scheduler): report an error when a recovered Allocate op's node is gone

### LINKED ISSUE #5778: scheduler: RecoverOperations panics when a saved Allocate operation's node has left the session
**What happened**:

`Statement.RecoverOperations` (`pkg/scheduler/framework/statement.go`) resolves an `Allocate` operation's node by name and passes the result straight to `Allocate`:

```go
case Allocate:
    node := s.ssn.Nodes[op.task.NodeName]
    err := s.Allocate(op.task, node)
```

A map lookup for a missing key yields `nil`, and `Allocate` dereferences it immediately:

```go
func (s *Statement) Allocate(task *api.TaskInfo, nodeInfo *api.NodeInfo) (err error) {
    ...
    hostname := nodeInfo.Name   // nil dereference
```

So if a node leaves the session between the operations being saved (`SaveOperations`) and replayed (`RecoverOperations`), the scheduler panics instead of returning an error.

The deferred rollback inside `Allocate` also dereferences `nodeInfo.Name` in its log line, so the panic can surface from there too.

`RecoverOperations` reports an error for the missing node, letting the caller discard the partially applied statement — the same way it already handles a failed `Pipeline` or `Allocate`.

**How to reproduce it**:

Replay a saved `Allocate` operation whose `task.NodeName` is not present in `ssn.Nodes`:

```go
ssn := &Session{Jobs: ..., Nodes: map[string]*api.NodeInfo{}}  // node absent
plan := &Statement{ssn: ssn, operations: []operation{{name: Allocate, task: task}}}
NewStatement(ssn).RecoverOperations(plan)
```

```
```

Note the sibling `Pipeline` branch is unaffected — it passes `op.task.NodeName` as a string and `Pipeline` reports a clean error when the node is absent.

**Anything else we need to know?**

Found while writing a regression test for #5362. I have a fix ready and will open a PR referencing this issue.

SrikarChittemsetty:
### What type of PR is this?

/kind bug

### What this PR does / why we need it

`Statement.RecoverOperations` resolved an `Allocate` operation's node with a bare map lookup and handed the result to `Allocate`:

```go
case Allocate:
    node := s.ssn.Nodes[op.task.NodeName]
    err := s.Allocate(op.task, node)
```

A missing key yields `nil`, and `Allocate` dereferences it on entry (`hostname := nodeInfo.Name`). So a node that left the session between `SaveOperations` and `RecoverOperations` panicked the scheduler rather than returning an error. The deferred rollback inside `Allocate` dereferences `nodeInfo.Name` in its log line too, so the panic can surface from there as well.

This checks the lookup and returns an error naming the missing node, letting the caller discard the partially applied statement — the same handling already applied to a failed `Pipeline` or `Allocate`. The sibling `Pipeline` branch needs no change: it passes `op.task.NodeName` as a string and `Pipeline` already reports a clean error when the node is absent.

### Which issue(s) this PR fixes

Fixes #5778

### Testing

Added `TestRecoverOperations_MissingNodeReturnsError`, which replays a saved `Allocate` operation against a session with no nodes. Against the previous behaviour it panics:

```
--- FAIL: TestRecoverOperations_MissingNodeReturnsError
```

With the fix it returns an error naming the node. The full `pkg/scheduler/framework` suite passes and `gofmt` is clean.

### Note

Found while writing the regression test for #5362 (PR #5775) — the two are independent and this one stands alone.

Disclosure: this change was prepared with the assistance of an AI coding agent. The analysis, fix, and test were verified against the source and the existing suite before submitting.

### DIFF
--- a/pkg/scheduler/framework/statement.go
+++ b/pkg/scheduler/framework/statement.go
@@ -470,10 +470,14 @@ func (s *Statement) RecoverOperations(stmt *Statement) error {
 				return err
 			}
 		case Allocate:
-			// The node may have left the session between the operations being
-			// saved and replayed. Allocate dereferences the node, so recovering
-			// against a missing one panics the scheduler; report it instead so the
-			// caller can discard the partially applied statement.
+			// This map read is the only unchecked one in the function, and Allocate
+			// dereferences the node immediately (hostname := nodeInfo.Name), so a
+			// miss here segfaults the scheduler instead of returning an error. The
+			// current callers cannot produce a miss — SaveOperations stores cloned
+			// tasks, so a replayed Allocate always carries the NodeName it recorded
+			// — but keep the invariant local rather than resting it on Clone()
+			// semantics in another package, and report it so the caller can discard
+			// the partially applied statement.
 			node, found := s.ssn.Nodes[op.task.NodeName]
 			if !found {
 				err := fmt.Errorf("failed to find node <%s> in session <%v> when recovering allocation of task <%v/%v>",
--- a/pkg/scheduler/framework/statement.go
+++ b/pkg/scheduler/framework/statement.go
@@ -470,7 +470,17 @@ func (s *Statement) RecoverOperations(stmt *Statement) error {
 				return err
 			}
 		case Allocate:
-			node := s.ssn.Nodes[op.task.NodeName]
+			// The node may have left the session between the operations being
+			// saved and replayed. Allocate dereferences the node, so recovering
+			// against a missing one panics the scheduler; report it instead so the
+			// caller can discard the partially applied statement.
+			node, found := s.ssn.Nodes[op.task.NodeName]
+			if !found {
+				err := fmt.Errorf("failed to find node <%s> in session <%v> when recovering allocation of task <%v/%v>",
+					op.task.NodeName, s.ssn.UID, op.task.Namespace, op.task.Name)
+				klog.Errorf("%v", err)
+				return err
+			}
 			err := s.Allocate(op.task, node)
 			if err != nil {
 				klog.Errorf("Failed to allocate task <%v/%v>: %v", op.task.Namespace, op.task.Name, err)
--- a/pkg/scheduler/framework/statement_recover_test.go
+++ b/pkg/scheduler/framework/statement_recover_test.go
@@ -77,3 +77,51 @@ func TestRecoverOperations_PipelinePreservesEvictionFlag(t *testing.T) {
 		assert.True(t, recoveredTask.EvictionOccurred)
 	}
 }
+
+// TestRecoverOperations_MissingNodeReturnsError covers replaying a saved Allocate
+// operation whose node is no longer in the session. RecoverOperations resolves the
+// node by name and hands the result straight to Allocate, which dereferences it,
+// so a node that left the session between save and replay panicked the scheduler
+// rather than surfacing an error the caller could act on.
+func TestRecoverOperations_MissingNodeReturnsError(t *testing.T) {
+	jobID := api.JobID("ns/job-missing-node")
+	task := &api.TaskInfo{
+		UID:        "t1",
+		Job:        jobID,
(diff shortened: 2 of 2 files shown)


SrikarChittemsetty:
The `e2e-sequence` failure here is unrelated to this change. The only failing spec is:

```
[FAIL] Queue Job Status Transition [It] Transform from running to unknown should succeed
33 Passed | 1 Failed | 1 Pending | 1 Skipped
```

That's a queue status-transition e2e; this PR only touches `pkg/scheduler/framework/statement.go` (the `Allocate` op recovery path) and its unit test, which don't interact with queue status transitions. Looks like a flake — could a maintainer re-run the `e2e-sequence` job when you get a chance? Happy to rebase if that's easier.

SrikarChittemsetty:
@JesseStutler — following up on `priority/awaiting-more-evidence`. I went back to substantiate the failure mode, and I want to report what I found honestly, including the part that doesn't support this PR as I originally wrote it.

**What the guard prevents.** Reverting `statement.go` to master and running the new test does produce a real segfault, not a returned error:

```
--- FAIL: TestRecoverOperations_MissingNodeReturnsError
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0]

volcano.sh/volcano/pkg/scheduler/framework.(*Statement).Allocate(...)
	pkg/scheduler/framework/statement.go:266
volcano.sh/volcano/pkg/scheduler/framework.(*Statement).RecoverOperations(...)
	pkg/scheduler/framework/statement.go:474
```

`RecoverOperations` reads `s.ssn.Nodes[op.task.NodeName]` and passes the result straight to `Allocate`, whose third line is `hostname := nodeInfo.Name`. A map miss yields a nil `*NodeInfo`, so it dereferences nil — and the deferred rollback dereferences `nodeInfo.Name` again, so it can't even unwind.

**What I could not establish: that it is reachable.** My description said the node "may have left the session between the operations being saved and replayed." I can't support that claim:

1. `ssn.Nodes` is a private snapshot. `Snapshot()` builds `make(map[string]*NodeInfo)` and fills it with `value.Clone()`, so deleting a node from the cache mid-session (`event_handlers.go:622`) does not remove it from a live session.
2. Nothing outside tests deletes from `ssn.Nodes`.
3. I thought I had a path through `Discard()`: it calls `unallocate()`, which ends with `task.NodeName = ""`, and `allocate.go` discards every dry-run statement *before* replaying the best one — which would make the replay look up `ssn.Nodes[""]`. But `SaveOperations` stores `op.task.Clone()`, and `Clone()` copies `NodeName`, so the backed-up operations keep their node name. That path does not fire.
4. A recorded `Allocate` op also implies the node was present when it was recorded, since `Statement.Allocate` errors and rolls back otherwise — which `TestAllocateAutoRollbackOnError` already covers.

So this is hardening against a nil `*NodeInfo` that the current call graph doesn't produce, not a bug with a reproducer. Your label was right, and I shouldn't have framed it as a live crash. I've pushed 50d0d5b to correct the code comment, the test comment, and the framing.

**Where that leaves it, your call.** The argument I'd still make for keeping it: `Statement.Allocate` `found`-checks every other lookup it does, including `ssn.Jobs` and `ssn.Nodes`; this is the one unchecked map read on the path into it, and the safety of that currently rests on `Clone()` semantics in another package rather than on anything local. Eight lines turns a scheduler segfault into a returned error the caller already knows how to handle.

But that's a judgement about whether the hardening is worth carrying, not evidence, and I don't want to keep a PR open under a premise I can't back. Happy to close it if you'd rather not carry it — just say the word and I'll close it myself.

SrikarChittemsetty:
Closing this per the evidence bar you set on #5778 — I can't provide a real-world log, panic trace, or triggering configuration, because I found this by reading the code rather than by hitting it. Analysis of why it isn't reachable through the current call graph is in my previous comment, for whoever reads this next.

Thanks for the review time.

PR #5783, opened 2026-07-29
scheduler/capacity: guard AddUnifiedEvictableFn against nil job/queue attrs

### LINKED ISSUE #5782: capacity plugin: AddUnifiedEvictableFn's GangReclaim path can panic with nil pointer dereference (missing job/queueOpts guards present elsewhere in the file)
## What happened

In `pkg/scheduler/plugins/capacity/capacity.go`, the `AddUnifiedEvictableFn` callback registered in `OnSessionOpen` (lines 607–646) looks up the job and queue attributes for each gang-reclaim candidate without checking for `nil`:

```go
for _, reclaimee := range candidates {
    job := ssn.Jobs[reclaimee.Job]
    attr := cp.queueOpts[job.Queue]
    if _, found := allocations[job.Queue]; !found {
        allocations[job.Queue] = attr.allocated.Clone()
    }
    ...
```

`ssn.Jobs[reclaimee.Job]` can return `nil` if the task's job/PodGroup was deleted between when `candidates` was built and when this callback runs. `cp.queueOpts[job.Queue]` can likewise return `nil` if the queue isn't tracked in `queueOpts`. Either case dereferences a nil pointer (`job.Queue` or `attr.allocated`) and panics.

The sibling callback in the same file, `AddReclaimableFn` (lines 459–478), handles exactly this scenario correctly:

```go
job := ssn.Jobs[reclaimee.Job]
if job == nil {
    klog.Warningf("[capacity] Skip reclaimee <%s/%s>: job <%s> not found in session (orphaned task from deleted PodGroup)",
        reclaimee.Namespace, reclaimee.Name, reclaimee.Job)
    continue
}
attr := cp.queueOpts[job.Queue]
if attr == nil {
    klog.Warningf("[capacity] Skip reclaimee <%s/%s>: queue <%s> not found in queueOpts",
        reclaimee.Namespace, reclaimee.Name, job.Queue)
    continue
}
```

## Impact

A crash in the scheduler's gang-reclaim eviction filtering during a live scheduling cycle whenever an orphaned/deleted-job task shows up among gang-reclaim candidates — e.g. a job or PodGroup deleted mid-cycle, or a queue removed from `queueOpts` between snapshot and this callback firing.

## Suggested fix

Add the same two nil guards used in `AddReclaimableFn` (skip with a `klog.Warningf` and `continue`) before dereferencing `job.Queue` and `attr.allocated`/`attr.guarantee`/`attr.deserved` in the `AddUnifiedEvictableFn` loop.

## Environment

Reproduced by code inspection against `master` (current as of the v1.14 line); not tied to a specific reported version.

bhuvan-somisetty:
#### What type of PR is this?

/kind bug

The GangReclaim branch of `AddUnifiedEvictableFn` in the capacity plugin looks up the reclaimee's job and queue attribute and dereferences them immediately, with no nil check:

```go
job := ssn.Jobs[reclaimee.Job]
attr := cp.queueOpts[job.Queue]
```

If a candidate's job was already removed from the session (job/PodGroup deleted mid-cycle) or its queue isn't tracked in `queueOpts`, this panics the scheduler instead of just skipping the candidate.

The sibling function `AddReclaimableFn`, a few dozen lines above in the same file, already guards against exactly this with a nil check and a `continue`. This PR applies the same pattern to `AddUnifiedEvictableFn` so a stale/orphaned candidate is skipped with a warning log instead of crashing the scheduling cycle.

Added a regression test that exercises both nil-guard paths directly against the plugin's registered `UnifiedEvictableFn` (a task whose job is missing from the session, and a task whose job exists but whose queue is missing from `queueOpts`), confirming neither panics and both are excluded from the returned victims.

Fixes #5782

None.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -622,7 +622,17 @@ func (cp *capacityPlugin) OnSessionOpen(ssn *framework.Session) {
 		allocations := map[api.QueueID]*api.Resource{}
 		for _, reclaimee := range candidates {
 			job := ssn.Jobs[reclaimee.Job]
+			if job == nil {
+				klog.Warningf("[capacity] Skip reclaimee <%s/%s>: job <%s> not found in session (orphaned task from deleted PodGroup)",
+					reclaimee.Namespace, reclaimee.Name, reclaimee.Job)
+				continue
+			}
 			attr := cp.queueOpts[job.Queue]
+			if attr == nil {
+				klog.Warningf("[capacity] Skip reclaimee <%s/%s>: queue <%s> not found in queueOpts",
+					reclaimee.Namespace, reclaimee.Name, job.Queue)
+				continue
+			}
 			if _, found := allocations[job.Queue]; !found {
 				allocations[job.Queue] = attr.allocated.Clone()
 			}
--- a/pkg/scheduler/plugins/capacity/capacity_test.go
+++ b/pkg/scheduler/plugins/capacity/capacity_test.go
@@ -1435,6 +1435,105 @@ func Test_buildHierarchicalQueueAttrs_nilSafety(t *testing.T) {
 	}
 }
 
+// Test_capacityPlugin_UnifiedEvictableFn_NilSafety is a regression test for
+// https://github.com/volcano-sh/volcano/issues/5782: the GangReclaim branch of
+// AddUnifiedEvictableFn dereferenced the reclaimee's job/queue attributes without
+// the nil checks that the sibling AddReclaimableFn already applies, so a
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
@avinxshKD good catch, thanks — updated the commit message subject to `capacity: ...` to match the repo convention and added the `Fixes #5782` footer. Force-pushed the fix.

bhuvan-somisetty:
yeah, agreed. closing this out — couldn't find a real path that reaches it, so not worth keeping open on a hunch. thanks for pushing on it, good reminder to actually trace the call sites instead of trusting a unit test that sets up the state directly. will reopen with real yamls/repro if this ever actually shows up in a live cluster.

PR #5784, opened 2026-07-29
job scheduler: fix preemption stuck Pending for fractional vGPU requests (vgpu-memory-percentage)

### LINKED ISSUE #4863: Vgpu can't execute preemption

CodeBuildder:
#### What type of PR is this?

/kind bug

A pod requesting a fractional vGPU slice (volcano.sh/vgpu-memory-percentage) stays Pending forever during preemption, even when evicting a lower priority pod would free up plenty of room on the card. Whole GPU requests using volcano.sh/vgpu-number only work fine.

normalPreempt decides whether an eviction freed up enough room by comparing the pod's requested resources against node.FutureIdle() using plain arithmetic. That comparison is only trustworthy for resources that actually have a capacity number in node.Status.Allocatable, like cpu, memory, or volcano.sh/vgpu-number. volcano.sh/vgpu-memory-percentage never does: it's a pure request-side modifier ("50% of whatever card this pod lands on" has no coherent node-wide total to advertise), and there's nowhere in this codebase, or in HAMi's device-plugin, that ever writes it into node capacity. So for a pod asking for it, that comparison can never come back true, no matter how much the device plugin's own ledger actually freed up after eviction, and the pod just stays stuck.

(Earlier draft of this PR also named volcano.sh/vgpu-cores as untracked. That was wrong: NewGPUDevices in device_info.go requires vgpu-cores, along with vgpu-memory, to be present and nonzero in node.Status.Allocatable, or it refuses to initialize GPU tracking for the node at all. So on any working vGPU node, vgpu-cores genuinely is tracked. The fix and the test now reflect that correction, one test case is kept as a synthetic regression guard for the general mechanism, and the realistic one confirms the bug via vgpu-memory-percentage specifically.)

deviceshare's own predicate already gets this right, it reads the real per-device usage and knows exactly what fits. It just never gets asked, normalPreempt makes its decision using only the generic arithmetic.

There's a second, earlier instance of the same problem in ValidateVictims (shared by both preempt and reclaim), which runs before any eviction happens at all, as a quick check on whether the selected victims would even be worth evicting. Same arithmetic, same blind spot, and it runs early enough that it blocks the fix in normalPreempt from ever getting a chance to matter.

The fix keeps the cheap arithmetic check as the fast path, since it's correct and free for anything with a real capacity number behind it. When it fails, it looks at exactly which resource dimensions it's unhappy about. If every one of them is something node.Allocatable has no entry for at all, the arithmetic genuinely can't answer the question, so it falls back to the real predicate instead of treating unknowable as no. If even one of the failing dimensions is something the node actually tracks, it trusts the arithmetic as is and doesn't second guess it. The fix doesn't hardcode either resource name anywhere, it reacts to whatever's actually missing from Allocatable at runtime.

This was deliberate, not just "always defer to the predicate on any failure": tried that first, and it broke two existing BestEffort preemption tests. If no predicate plugin is registered to check a given resource at all, ssn.PredicateFn just returns success by default since there's nothing to check against. A blanket fallback would quietly turn off resource enforcement for anything not covered by whatever plugins happen to be configured, which is a scarier failure mode than the one being fixed.

ValidateVictims gets the same treatment, except since it runs before any real eviction, it simulates the eviction on a throwaway clone of the node before asking the predicate, so nothing on the real node is touched unless the eviction actually goes ahead.

Fixes #4863

Used Claude Code (AI assistant) to investigate this issue, arrive at the root cause, and draft the fix and test. I reviewed every change, reasoned through the tradeoffs of a couple of alternative approaches before settling on this one, and ran the full pkg/scheduler/... suite myself before opening and updating this PR. I also went back and corrected an earlier claim in this PR (see below) after double-checking it against the actual code rather than taking it at face value. Happy to walk through any part of the reasoning or code in review.

Test coverage in TestPreemptFractionalVGPU:
- a synthetic vgpu-cores scenario (kept as a general regression guard for the fallback mechanism, its node doesn't declare vgpu-cores in Allocatable so it isn't proven to reflect production)
- a realistic vgpu-memory-percentage scenario, where the node properly declares vgpu-number/vgpu-cores/vgpu-memory in Allocatable matching NewGPUDevices's real gate, and the bug still reproduces purely from vgpu-memory-percentage being untracked
- a whole-GPU vgpu-number-only scenario as a regression guard confirming that path is untouched

All three fail against unpatched code and pass with the fix.

```release-note
Fix a bug where a pod requesting a fractional vGPU (volcano.sh/vgpu-memory-percentage) could stay Pending indefinitely during preemption, even when evicting a lower priority pod would have freed enough capacity. Whole-GPU requests (volcano.sh/vgpu-number) were not affected.
```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -342,14 +342,16 @@ func (pmpt *Action) preempt(
 //
 // It can never be satisfied, however, for resources a device plugin tracks
 // entirely through its own predicate/ledger instead of node.Status.Allocatable
-// - e.g. volcano.sh/vgpu-cores and volcano.sh/vgpu-memory-percentage, which by
-// design are never advertised as node capacity (see
-// docs/user-guide/how_to_use_volcano_vgpu.md) because "50% of one card" isn't
-// a poolable node-wide quantity. When every failing dimension is one of
-// these - node.Allocatable has no capacity number for it at all - the
-// comparison is structurally unable to answer the question, so fall back to
-// the real per-plugin predicate (which deviceshare uses for the correct,
-// device-aware fit check) instead of treating "unknowable" as "no".
+// - confirmed for volcano.sh/vgpu-memory-percentage, which is a pure
+// request-side modifier ("50% of whatever card this pod lands on" has no
+// coherent node-wide capacity total to advertise) and is never referenced
+// anywhere as something written into Allocatable. volcano.sh/vgpu-cores may or
+// may not have the same gap depending on how the device plugin advertises it;
+// this fallback covers that case too if so, without assuming it either way.
+// When every failing dimension is one node.Allocatable has no capacity number
+// for at all, the comparison is structurally unable to answer the question,
+// so fall back to the real per-plugin predicate (which deviceshare uses for
+// the correct, device-aware fit check) instead of treating "unknowable" as "no".
 //
 // Note: this fallback is only as good as the registered predicate plugins.
 // A resource dimension that's neither in node.Allocatable NOR checked by any
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -334,6 +334,57 @@ func (pmpt *Action) preempt(
 	return preemptSuccess, err
 }
 
+// fitsAfterEviction reports whether preemptor now fits on node. The generic
+// scalar-resource comparison is authoritative whenever every dimension it
+// flags as insufficient is one actually present in node.Allocatable (cpu,
+// memory, volcano.sh/vgpu-number, ...) - those have a real capacity number,
+// so the comparison can be trusted as-is, including a genuine "no" answer.
+//
+// It can never be satisfied, however, for resources a device plugin tracks
+// entirely through its own predicate/ledger instead of node.Status.Allocatable
+// - e.g. volcano.sh/vgpu-cores and volcano.sh/vgpu-memory-percentage, which by
+// design are never advertised as node capacity (see
+// docs/user-guide/how_to_use_volcano_vgpu.md) because "50% of one card" isn't
+// a poolable node-wide quantity. When every failing dimension is one of
+// these - node.Allocatable has no capacity number for it at all - the
+// comparison is structurally unable to answer the question, so fall back to
+// the real per-plugin predicate (which deviceshare uses for the correct,
+// device-aware fit check) instead of treating "unknowable" as "no".
+//
+// Note: this fallback is only as good as the registered predicate plugins.
+// A resource dimension that's neither in node.Allocatable NOR checked by any
+// registered predicate can't be judged by either mechanism - ssn.PredicateFn
+// simply has nothing to say about it and reports success. That matches
+// production configurations (the predicates plugin covers cpu/memory/pods;
+// deviceshare covers GPU shares), but a test harness that omits every plugin
+// covering a given dimension will see this fallback as permissive for it.
+func fitsAfterEviction(ssn *framework.Session, preemptor *api.TaskInfo, node *api.NodeInfo) bool {
+	ok, failing := preemptor.InitResreq.LessEqualWithResourcesName(node.FutureIdle(), api.Zero)
+	if ok {
+		return true
+	}
+	if !allUntrackedByAllocatable(failing, node) {
+		return false
+	}
+	return ssn.PredicateFn(preemptor, node) == nil
+}
+
+// allUntrackedByAllocatable reports whether every named resource dimension is
+// absent from node.Allocatable's tracked capacity. cpu/memory are always
+// considered tracked (they're dedicated Resource fields, not scalar entries,
+// and always have a real, if zero, capacity number).
+func allUntrackedByAllocatable(resourceNames []string, node *api.NodeInfo) bool {
+	for _, name := range resourceNames {
+		if name == "cpu" || name == "memory" {
+			return false
+		}
+		if _, present := node.Allocatable.ScalarResources[v1.ResourceName(name)]; present {
+			return false
+		}
+	}
+	return true
+}
+
 func (pmpt *Action) normalPreempt(
 	ssn *framework.Session,
 	stmt *framework.Statement,
@@ -369,7 +420,7 @@ func (pmpt *Action) normalPreempt(
 		victims := ssn.Preemptable(preemptor, preemptees)
 		metrics.UpdatePreemptionVictimsCount(len(victims))
 
-		if err := util.ValidateVictims(preemptor, node, victims); err != nil {
+		if err := util.ValidateVictims(preemptor, node, victims, ssn.PredicateFn); err != nil {
 			klog.V(3).Infof("No validated victims on Node <%s>: %v", node.Name, err)
 			continue
 		}
@@ -395,7 +446,7 @@ func (pmpt *Action) normalPreempt(
 			// so if current queue is not allocatable(the queue will be overused when consider current preemptor's requests)
 			// or current idle resource is not enough for preemptor, it need to continue preempting
 			// otherwise, break out
-			if ssn.Allocatable(currentQueue, preemptor) && preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
+			if ssn.Allocatable(currentQueue, preemptor) && fitsAfterEviction(ssn, preemptor, node) {
 				break
 			}
 			preemptee := victimsQueue.Pop().(*api.TaskInfo)
@@ -415,7 +466,7 @@ func (pmpt *Action) normalPreempt(
 			preempted, preemptor.Namespace, preemptor.Name, preemptor.InitResreq)
 
 		// If preemptor's queue is not allocatable, it means preemptor cannot be allocated. So no need care about the node idle resource
-		if ssn.Allocatable(currentQueue, preemptor) && preemptor.InitResreq.LessEqual(node.FutureIdle(), api.Zero) {
+		if ssn.Allocatable(currentQueue, preemptor) && fitsAfterEviction(ssn, preemptor, node) {
 			if err := nodeStmt.Pipeline(preemptor, node.Name, evictionOccurred); err != nil {
 				klog.Errorf("Failed to pipeline Task <%s/%s> on Node <%s>",
 					preemptor.Namespace, preemptor.Name, node.Name)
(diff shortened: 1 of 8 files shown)


CodeBuildder:
@merryzhou @wangyang0616 You're both already down as reviewers here - just making sure it didn't get buried. Fixes preemption getting stuck Pending for fractional vGPU requests (#4863). No rush, happy to adjust anything.

CodeBuildder:
@miantalha45 you caught a good one on #5808 so tagging you here too if you have time. Fixes preemption getting permanently stuck Pending for fractional vGPU requests (volcano.sh/vgpu-memory-percentage) - #4863. Rebased and cleaned up the commits since it last got attention, still passing the full pkg/scheduler/... suite.

While reviewing this one against the rest of the codebase I found the same root cause (a generic scalar-resource check with no fallback to the real predicate for resources untracked in node.Status.Allocatable) in two more spots that this PR doesn't touch: the allocate action's node predicate (#5857), and topology-aware preemption's victim selection (#5858). Kept this PR scoped to what it's linked to (#4863) rather than pulling those in, filed them separately with the repro/analysis. Planning to send fixes for those as their own PRs once I get to them.

CodeBuildder:
Hey, just noticed this has a conflict with master now, probably picked up after some unrelated merges landed since I opened it. Happy to rebase if that'd help move it along. Still fixes the same thing, preemption getting stuck Pending for fractional vGPU requests (#4863).

cc @hajnalmt, appreciate you taking a look when you get a chance.

CodeBuildder:
Rebased this on top of master, and it needed more than a rebase. #5843 and #5855 landed in the meantime and changed `preemptorFitsOnNode` and added `reclaimerFitsOnNode`, both now rechecking the real predicate after each tentative eviction (for the HAMi Ascend fragmentation case). That's the right fix for that bug, but the recheck is still gated behind the same untrustworthy arithmetic this PR was written to fix, so the original bug (#4863) was still there, just one layer deeper.

Reworked the fix to sit on top of that: both functions now fall back to the real predicate specifically when the arithmetic shortfall is confined to a resource dimension `node.Allocatable` has no entry for at all, instead of rejecting outright. `ValidateVictims` gets the same treatment since it runs before either loop and has the identical gap.

Added a regression test for each action. Confirmed both fail against unpatched master (no eviction happens, task stays pending) and pass with the fix. Full `pkg/scheduler/...` suite still green.

One thing I ran into while writing the reclaim test: `pkg/scheduler/plugins/proportion` has the same untracked-resource gap independently, in its own `Allocatable`, `Preemptive`, and `Reclaimable` checks (all built on the same `queueAllocatable` arithmetic). Left that alone here since it's a different plugin, but wanted to flag it since it means a real cluster with proportion's queue-fairness checks enabled could still see this class of bug even after this PR merges. Might be worth its own issue.

CodeBuildder:
Pushed a follow-up commit, first one was missing the DCO sign-off, fixed now.

cc @hajnalmt, you've been the one landing most of the scheduler fixes in this exact area lately (#5843, #5855), would appreciate a look when you get a chance.

CodeBuildder:
Found something important while extending this fix to the allocate action (separate issue, coming next): fixing normalPreempt and reclaim alone wasn't actually enough. A pipelined preemptor still has to pass through NodeInfo.AddTask's Binding-status check before it actually binds to the API server, and that check has the exact same unguarded arithmetic, with no predicate recheck and no fallback. It's the final common gate every task passes through before binding, across every action, so without fixing it too, a preemptor could get pipelined correctly by the earlier fix and then silently never actually bind.

Added the same fallback there, plus two direct unit tests on NodeInfo.AddTask (one confirming the untracked-resource case now passes, one confirming a genuine tracked-resource shortfall still correctly fails). Moved the shared fallback helper into pkg/scheduler/api since pkg/scheduler/api can't import pkg/scheduler/util where it originally lived.

Full pkg/scheduler/... suite still green.

hajnalmt (commented):
Thanks, but aren't you just misusing the API? (I know the documentation is bad, so it's not fully on you 😀)

Based on the code you need define either `volcano.sh/vgpu-number` or `volcano.sh/vgpu-memory`  always and there is no such case that `volcano.sh/vgpu-memory-percentage` is defined only.

Proof in ExtractResourceRequest:
https://github.com/volcano-sh/volcano/blob/v1.15.2/pkg/scheduler/api/devices/util.go#L213-L218

And this was a delibirate design choice, so if you want 2 cards each of them allocated in 30 percent you can define:
```
volcano.sh/vgpu-number: 2
volcano.sh/vgpu-memory-percentage: 30
```
In the end this will allocate 30% of the memory but on 2 cards! Respectively you should add `volcano.sh/vgpu-number: 1` always for a single percentage-wise memory allocation.

CodeBuildder:
You're right, and thanks for catching it — checked `ExtractResourceRequest` myself, percentage genuinely can't be requested standalone; the percentage read only happens after `vgpu-number`/`vgpu-memory` is already found.

The bug itself is still real, just not the way the test showed it: it's not that percentage gets requested alone, it's that it's normally requested alongside `vgpu-number` (fine, tracked), and the current gate rejects the whole pod the moment *any single* dimension among several is untracked — even with everything else on the same request perfectly satisfiable. Confirmed this directly: a pod requesting `vgpu-number: 1` + `vgpu-cores: 50` + `vgpu-memory-percentage: 50` together, on a fully idle 20GB node, still gets wrongly rejected by current code.

Pushed a fix to all three test fixtures — each pod now requests `vgpu-number` alongside the percentage, and each node's Allocatable tracks `vgpu-number` like a real HAMi node would, leaving `vgpu-memory-percentage` as the one deliberately-untracked dimension actually under test. Re-verified fail→pass in both directions with the corrected fixtures.

**Why this fix and not something bigger:** two other options I considered and passed on —
- *Swap the resource name for a purely synthetic one* (e.g. an invented `example.com/untracked-resource`) instead of `vgpu-memory-percentage`, to sidestep the "is this realistic" question entirely. Rejected — it would disconnect the test from the actual motivating bug report (#4863) and make it read as an abstract mechanism check rather than a fix for a real, reported failure.
- *Rewrite the test to go through the real HAMi deviceshare/vgpu predicate* end to end instead of the lightweight stand-in predicate plugin already in the fixture. Rejected as too big for this fix — it would also start exercising HAMi's own percentage-to-memory conversion logic, which is a separate concern from the one bug this test isolates (the generic scalar-gate rejecting a request before the real predicate even runs). That's worth having as its own, separate integration test, not a change to this one.

Adding `vgpu-number` alongside the percentage was the smallest change that fully answers your objection without touching what the test actually verifies or losing the tie to the real issue. Appreciate you pushing on this instead of letting it slide.

PR #5796, opened 2026-07-31
Fix PodGroup falling into Inqueue while pods are terminating

### LINKED ISSUE #5794: PodGroup falls into Inqueue while Pods are terminating (Releasing)
### Description

When a Pod managed by Volcano has `metadata.deletionTimestamp` set but its containers are still running (Kubernetes `Terminating`, Volcano task status `Releasing`), the associated PodGroup incorrectly transitions from `Running` to `Pending` / `Inqueue`.

This was observed in a real cluster. The impact is:

1. PodGroup phase no longer reflects that the job is still finishing / releasing resources.
2. With `actions: "allocate, backfill"` (no `enqueue`), allocate rewrites `Pending` → `Inqueue`.
3. With the `gang` plugin enabled, PodGroup gets a misleading `Unschedulable` / `NotEnoughResources` condition such as `1 Releasing, 1 minAvailable`, even though no new scheduling is needed.

Typical production path: long graceful termination (ignore SIGTERM, large `terminationGracePeriodSeconds`, or a multi-container Pod where the main container hangs). Short-lived Terminating windows may briefly pass through `Inqueue`; sticky Terminating makes PodGroup stay in `Inqueue` until the Pod is fully gone.

Here is my scheduler config:
```yaml
actions: "allocate, backfill"
tiers:
- plugins:
  - name: binpack
  - name: priority
  - name: gang
    enablePreemptable: false
  - name: conformance
- plugins:
  - name: drf
    enablePreemptable: false
  - name: proportion
  - name: nodeorder
  - name: predicates
```

1. Create  following PodGroup + Pod

```yaml
kind: PodGroup
metadata:
  name: pg-sticky-releasing
  namespace: volcano-repro
spec:
  minMember: 1
  queue: default
---
apiVersion: v1
kind: Pod
metadata:
  name: sticky-releasing
  namespace: volcano-repro
  annotations:
    scheduling.k8s.io/group-name: pg-sticky-releasing
spec:
  schedulerName: volcano
  restartPolicy: Never
  terminationGracePeriodSeconds: 3600
  containers:
  - name: main
    image: busybox:1.36
    command: ["sh", "-c", "trap '' TERM; echo started; sleep 999999"]
    resources:
      requests:
        cpu: "50m"
        memory: "64Mi"
      limits:
        cpu: "50m"
        memory: "64Mi"
```

2. Wait until both are Running,:

```text
pod/sticky-releasing                         1/1  Running   ...
podgroup.scheduling.volcano.sh/pg-sticky-releasing   Running   1   1   ...
```

3. Delete the Pod without waiting:

```bash
kubectl -n volcano-repro delete pod sticky-releasing --wait=false
```

4. Watch Pod / PodGroup for ~20s while the Pod stays Terminating.

Observed status sequence from our reproduction:

```text
B t=1s  pod=Running ready=true del=yes pg=Running
B t=2s  pod=Running ready=true del=yes pg=Inqueue
B t=3s  pod=Running ready=true del=yes pg=Inqueue
...
B t=20s pod=Running ready=true del=yes pg=Inqueue
```

Final objects:

```text
NAME               READY   STATUS        RESTARTS   AGE   IP               NODE
sticky-releasing   1/1     Terminating   0          31s   10.244.169.185   k8s-node2
```

PodGroup status:

```yaml
status:
  conditions:
  - type: Scheduled
    status: "True"
    reason: tasks in gang are ready to be scheduled
  - type: Unschedulable
    status: "True"
    reason: NotEnoughResources
    message: '1/1 tasks in gang unschedulable: pod group is not ready, 1 Releasing, 1 minAvailable'
  phase: Inqueue
```

It can be seen that the podgroup has entered the Inqueue state, as if it were still waiting to be scheduled.

```text
NAME               READY   STATUS        AGE   IP               NODE
sticky-releasing   1/1     Terminating   31s   10.244.169.185   k8s-node2
```

PodGroup flipped within ~2s and stayed in `Inqueue` for the full 20s watch window:

```text
B t=1s  pod=Running ready=true del=yes pg=Running
B t=2s  pod=Running ready=true del=yes pg=Inqueue
...
B t=20s pod=Running ready=true del=yes pg=Inqueue
```

PodGroup status written by the scheduler/gang:

```yaml
status:
  phase: Inqueue
  conditions:
  - type: Unschedulable
    status: "True"
    reason: NotEnoughResources
    message: '1/1 tasks in gang unschedulable: pod group is not ready, 1 Releasing, 1 minAvailable'
```

- While the job still has `Releasing` tasks and no pending work, PodGroup should remain **`Running`** (Because the container has not actually exited yet.).
- Gang plugin should not mark such a `Terminating` job as unschedulable for “not enough resources”.

master

_No response_

halcyon-r:
#### What type of PR is this?

/kind bug

  When a scheduled Pod is gracefully deleted, it may remain running on its node with a `deletionTimestamp` while Volcano represents it as a `Releasing` task.
  Volcano did not count these tasks as scheduled when calculating PodGroup status. As a result:
  - A single-Pod or fully terminating PodGroup could transition from `Running` to
    `Pending` and then `Inqueue`.
  - A partially terminating multi-Pod PodGroup could transition from `Running` to
    `Unknown`.
  - The gang plugin could repeatedly report misleading
    `Unschedulable` / `NotEnoughResources` conditions, even though the terminating
    Pods were still occupying their nodes and no new placement was required.

  This PR treats a `Releasing` task that is still assigned to a node as an already-scheduled PodGroup member. The PodGroup therefore remains `Running` during the graceful termination window when its scheduled members still satisfy `minMember`. A `Releasing` task without a node assignment is not counted, because it may have disappeared.

  The tests cover:
  - Graceful termination of a single-Pod PodGroup.
  - Partial graceful termination of a multi-Pod PodGroup.
  - Avoiding new misleading `NotEnoughResources` conditions during termination.
  - Excluding never-scheduled terminating Pods.
  - Preserving `Unknown` after a required member is actually gone.

AI disclosure: AI tools were used to help draft/debug parts of this change; the design, cluster reproduction, and final code were reviewed and validated by the author.

Fixes #5794 

`@miantalha45` is currently assigned on #5794; I left a note on the issue and opened this PR with the fix + tests to avoid duplicated work. Happy to align if you already have other thoughts ongoing changes.

```release-note
Fix PodGroups incorrectly falling back to Pending, Inqueue, or Unknown while scheduled member Pods are gracefully terminating, and avoid misleading NotEnoughResources conditions during that termination window.
```

### DIFF
--- a/pkg/scheduler/api/job_info.go
+++ b/pkg/scheduler/api/job_info.go
@@ -1259,6 +1259,32 @@ func (ji *JobInfo) IsStarving() bool {
 	return ji.WaitingTaskNum()+ji.ReadyTaskNum() < ji.MinAvailable
 }
 
+// ScheduledTaskNum returns the number of tasks in a status that indicates they
+// have already been scheduled.
+func (ji *JobInfo) ScheduledTaskNum() int32 {
+	count := 0
+	for status, tasks := range ji.TaskStatusIndex {
+		if ScheduledStatus(status) {
+			count += len(tasks)
+		}
+	}
+	return int32(count)
+}
+
+// ReleasingScheduledTaskNum returns the number of Releasing tasks that were
+// already scheduled to a node (pods with deletionTimestamp whose containers may
+// still be running and occupying node resources). Releasing tasks that were
+// never scheduled (deleted while pending) are not counted.
+func (ji *JobInfo) ReleasingScheduledTaskNum() int32 {
+	count := 0
+	for _, task := range ji.TaskStatusIndex[Releasing] {
+		if task.NodeName != "" {
+			count++
+		}
+	}
+	return int32(count)
+}
+
 // IsPending returns whether job is in pending status
 func (ji *JobInfo) IsPending() bool {
 	return ji.PodGroup == nil ||
--- a/pkg/scheduler/plugins/gang/gang.go
+++ b/pkg/scheduler/plugins/gang/gang.go
@@ -235,6 +235,9 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 			continue
 		}
 		if !job.IsReady() {
+			if shouldSkipUnschedulableForReleasing(job) {
+				continue
+			}
 			schedulableTaskNum := func() (num int32) {
 				for _, task := range job.TaskStatusIndex[api.Pending] {
 					ctx := task.GetTransactionContext()
@@ -294,4 +297,25 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 	}
 
 	metrics.UpdateUnscheduleJobCount(unScheduleJobCount)
+}
+
+// shouldSkipUnschedulableForReleasing reports whether an unready job has enough
+// effective scheduled tasks to satisfy minAvailable. Releasing tasks still
+// assigned to nodes count because their Pods continue to occupy gang slots.
+// Pending or Pipelined tasks prevent this shortcut because they still require
+// scheduling.
+func shouldSkipUnschedulableForReleasing(job *api.JobInfo) bool {
+	scheduledReleasing := job.ReleasingScheduledTaskNum()
+	if scheduledReleasing == 0 {
+		return false
+	}
+
+	// Pending or Pipelined tasks still require the normal gang condition so they
+	// remain visible as work that needs to be scheduled.
+	if len(job.TaskStatusIndex[api.Pending]) > 0 || len(job.TaskStatusIndex[api.Pipelined]) > 0 {
+		return false
+	}
+
+	effectiveScheduled := job.ScheduledTaskNum() + scheduledReleasing
+	return effectiveScheduled >= job.MinAvailable
 }
--- a/test/e2e/jobseq/queue_job_status.go
+++ b/test/e2e/jobseq/queue_job_status.go
@@ -20,15 +20,11 @@ import (
 	"context"
 	"fmt"
 	"strconv"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/watch"
-	"k8s.io/client-go/tools/cache"
-	watchtools "k8s.io/client-go/tools/watch"
 
 	"volcano.sh/apis/pkg/apis/scheduling/v1beta1"
 	e2eutil "volcano.sh/volcano/test/e2e/util"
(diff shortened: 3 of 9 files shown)


pkg/scheduler/api/job_info.go
```diff
@@ -1259,6 +1259,30 @@ func (ji *JobInfo) IsStarving() bool {
...
+	}
+	for status, tasks := range ji.TaskStatusIndex {
+		if len(tasks) == 0 {
+			continue
+		}
+		switch status {
+		case Succeeded, Failed, Releasing:
+			continue
+		default:
+			return false
```
JesseStutler: I noticed that non-`Succeeded/Failed/Releasing` here returns false here, and it return false here. This can solve your podgroup single pod issue, but what if the podgroup has multiple pods? Won't it still have reverting? @halcyon-r

halcyon-r: I think you're right. Indeed, if not all pods in a pod group are terminating, this will falls back to the old logic. Let me rethink the solution.

halcyon-r: Hey @JesseStutler . There's a question I'd like to discuss with you. Based on my testing, in scenarios with multiple pods within a single PodGroup, when some pods are in Releasing while the rest are still Running, the PodGroup transitions to an `Unknown` state. Is this in line with the design?

halcyon-r: The existing E2E test appears to treat Unknown as the expected behavior when only part of a PodGroup is deleted: [queue_job_status.go](https://github.com/volcano-sh/volcano/blob/master/test/e2e/jobseq/queue_job_status.go#L190-L225)

```go
By("Delete some of pod which will case pod group status transform from running to unknown.")
podDeleteNum := 0

err = e2eutil.WaitPodPhaseRunningMoreThanNum(testCtx, podNamespace, 2)
Expect(err).NotTo(HaveOccurred(), "Failed waiting for pods")

clusterPods, err := testCtx.Kubeclient.CoreV1().Pods(podNamespace).List(context.TODO(), metav1.ListOptions{})
for _, pod := range clusterPods.Items {
	if pod.Status.Phase == corev1.PodRunning {
		err = testCtx.Kubeclient.CoreV1().Pods(podNamespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
		Expect(err).NotTo(HaveOccurred(), "Failed to delete pod %s", pod.Name)
		podDeleteNum = podDeleteNum + 1
	}
	if podDeleteNum >= int(rep/2) {
		break
	}
}

By("Verify queue have pod groups unknown")
w := &cache.ListWatch{
	WatchFunc: func(options metav1.ListOptions) (i watch.Interface, e error) {
		return testCtx.Vcclient.SchedulingV1beta1().PodGroups(podNamespace).Watch(context.TODO(), options)
	},
}
wctx, cancel := watchtools.ContextWithOptionalTimeout(context.Background(), 5*time.Minute)
defer cancel()

_, err = watchtools.Until(wctx, clusterPods.ResourceVersion, w, func(event watch.Event) (bool, error) {
	switch t := event.Object.(type) {
	case *v1beta1.PodGroup:
		if t.Status.Phase == v1beta1.PodGroupUnknown {
			return true, nil
		}
	}
	return false, nil
})
```
That certainly aligns with the semantics of "gang" scheduling, but I feel it fundamentally overlooks the window where a pod might get stuck in the "Terminating" state—assuming instead that a pod disappears immediately upon deletion.

halcyon-r:
Adjustments were made based on the multi-pod PodGroup. For a pod in the `Terminating` state, the associated `PodGroup` is not set to `Unknown` but remains in the `Running` state; the status is switched to `Unknown` only after the pod has actually been deleted and the pod count falls below `minMember`.
The various status change scenarios are summarized as follows:
| Scenario | Result |
|---|---|
| All required members are Running | `Running` |
| Some members are Running and some are scheduled `Releasing`, with `effectiveScheduled >= minMember` | Remain `Running` |
| All active members are scheduled `Releasing`, with `effectiveScheduled >= minMember` | Remain `Running` until they reach a terminal state or are removed |
| Scheduled `Releasing` and completed tasks together satisfy `minMember` | Remain `Running` while any scheduled `Releasing` task is not completed |
| Scheduled `Releasing` tasks satisfy `minAvailable`, with no `Pending` or `Pipelined` tasks | Do not add a new gang `NotEnoughResources` condition |
| A deleted Pending Pod becomes `Releasing` but has no `NodeName` | It is not counted; normal `Pending/Inqueue` behavior is preserved |
| A replacement task is `Pending` or `Pipelined` | Keep the normal gang condition and allocation path; the replacement remains schedulable |
| A required member is completely gone, the remaining effective scheduled count is below `minMember`, and another member is still Running | `Unknown` after the gang is reported unschedulable |
| The effective scheduled count is below `minMember` and no member remains Running | `Pending/Inqueue`, not `Unknown` |
| Missing members are successfully scheduled again | `Unknown -> Running` |
| Tracked `Succeeded/Failed` tasks satisfy `minMember` and all scheduled tasks are completed | `Completed` |
| All Pod objects are removed without enough tracked terminal tasks | `Pending/Inqueue`, or the PodGroup is deleted by its owner |

pkg/scheduler/api/job_info.go
JesseStutler: > Hey @JesseStutler . There's a question I'd like to discuss with you. Based on my testing, in scenarios with multiple pods within a single PodGroup, when some pods are in Releasing while the rest are still Running, the PodGroup transitions to an `Unknown` state. Is this in line with the design?

I feel like this podgroup should not transition to Unknown state

JesseStutler: After looking at the original PodGroup status design, I’m not convinced that we should transition to Unknown at all.

Unknown currently means that the PodGroup previously reached minMember, but later dropped below it while some members remain running and the missing members cannot be scheduled. That is not actually an unknown state; it is a running but degraded/undersized PodGroup.

I think minMember should primarily be treated as the gang admission barrier. Once the PodGroup has started, dropping below minMember should not invalidate the fact that its remaining members are still running. The phase could remain Running, while the shortage and recovery failure are represented by the running counters and an Unschedulable/degraded condition.

Therefore, I’m wondering whether this PR should preserve Running not only during the Releasing window, but also after a member is gone, as long as other members are still running. We can retain the Unknown enum for API compatibility while no longer transitioning PodGroups into it.

pkg/scheduler/plugins/gang/gang.go
```diff
@@ -295,3 +298,15 @@ func (gp *gangPlugin) OnSessionClose(ssn *framework.Session) {
 
 	metrics.UpdateUnscheduleJobCount(unScheduleJobCount)
 }
+
+// shouldSkipConditionUpdateForScheduledReleasingTasks returns true when the phase's
+// scheduled slots, including completed tasks and already-scheduled Releasing
+// tasks, satisfy minAvailable without any pending work to schedule.
+func shouldSkipConditionUpdateForScheduledReleasingTasks(job *api.JobInfo) bool {
```
JesseStutler: This could use some more wrapping and comments to improve readability. Although I think it makes sense logically, there are too many `&&` here, making it hard for others to understand. Especially with the `!job.IsReady()` mixed in, others will find it strange what this function does.

test/e2e/jobseq/queue_job_status.go
```diff
@@ -151,79 +147,54 @@ var _ = Describe("Queue Job Status Transition", func() {
 		Expect(err).NotTo(HaveOccurred(), "Error waiting for queue Pending")
 	})
 
-	It("Transform from running to unknown should succeed", func() {
-		By("Prepare 2 job")
-
-		podNamespace = testCtx.Namespace
-		slot := e2eutil.HalfCPU
-		rep = e2eutil.ClusterSize(testCtx, slot)
+	It("Transform from running to unknown after a member is gone should succeed", func() {
```
JesseStutler, line 150: Currently I accept keeping the unknown phase, but I think deleting some pods shouldn't lead the podgroup to unknown phase , instead, some refined conditions should be added. And the podgroup should keep in Running phase. I think "unknown" was actually a very strange design from the birth of volcano, but I won't block this PR; it just needs optimization in the future

halcyon-r: Yes, I completely agree. This PR should only handle issues caused by tasks stuck in releasing. As for the semantic issues of job status, more detailed design improvements should be needed in the future.

pkg/scheduler/plugins/gang/gang.go
halcyon-r: Thanks for the suggestion. I have reorganized the judgment logic and comments that gang plugins should skip.

JesseStutler:
/approve
/lgtm
Thanks!

Issue #5803, opened 2026-07-31
Only trigger handler when `networkTopologyDiscovery` config changes


karthik120710:
### Description

The ConfigMap event handler triggers work on every update:

```go
// TODO: Only trigger handler when networkTopologyDiscovery config changed.
hn.configMapInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
    AddFunc:    hn.addConfigMap,
    UpdateFunc: hn.updateConfigMap,
    DeleteFunc: hn.deleteConfigMap,
})
```

`updateConfigMap` (line 80) enqueues the ConfigMap on **any** update, even when the `networkTopologyDiscovery` content is unchanged or unrelated data changed.

### Fix

1. In `updateConfigMap`, compare the old and new ConfigMap `Data` (specifically the `networkTopologyDiscovery` section).
2. Only call `enqueueConfigMap` when the relevant config actually changed.
3. Add unit tests for: unrelated change (no trigger), relevant change (trigger), and ConfigMap creation/deletion.

PR #5804, opened 2026-07-31
Only trigger hyperNode controller when network topology discovery config changes

karthik120710:
## Description:
The hyperNode controller's ConfigMap informer enqueued the ConfigMap on every update, even when the networkTopologyDiscovery content was unchanged or only unrelated keys in the ConfigMap Data changed. This caused the discovery manager to be triggered unnecessarily on every ConfigMap resync/update.
## This PR makes the handler change-aware:
- Extract ParseConfig in pkg/controllers/hypernode/config/loader.go so the YAML parsing logic is shared between the loader and change detection.
- Rewrite updateConfigMap in pkg/controllers/hypernode/configmap_handler.go to only enqueue when the config actually changed. A new configChanged helper:
  - Triggers when the volcano-controller.conf key is added or removed
  - Skips when the key is absent or byte-identical
  - Compares the parsed NetworkTopologyConfig so whitespace-only or unrelated-data changes are ignored
  - Falls back to triggering on any difference if parsing fails (conservative)
- Update unit and integration tests to cover: unrelated change (no trigger), unrelated change with unchanged config (no trigger), relevant config change (trigger), config key added/removed (trigger), and invalid objects (no trigger).
No behavior change when the config is actually modified — this only avoids spurious wake-ups of the discovery manager.

#5803

### DIFF
--- a/pkg/controllers/hypernode/config/loader.go
+++ b/pkg/controllers/hypernode/config/loader.go
@@ -70,8 +70,13 @@ func (l *loader) LoadConfig() (*api.NetworkTopologyConfig, error) {
 		return nil, fmt.Errorf("config key not found in ConfigMap: %s", l.configKey)
 	}
 
+	return ParseConfig(configYaml)
+}
+
+// ParseConfig parses the network topology config from YAML.
+func ParseConfig(configYaml string) (*api.NetworkTopologyConfig, error) {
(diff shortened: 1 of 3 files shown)


karthik120710:
@hzxuzhonghu @hwdef can u review my pr

PR #5808, opened 2026-07-31
fix(tdm): avoid panic on non-string plugin arguments

### LINKED ISSUE #5806: tdm plugin panics and crash-loops the entire scheduler on non-string plugin argument
### Description

`pkg/scheduler/plugins/tdm/tdm.go`'s `New()` function does unchecked type assertions on plugin arguments read from the scheduler configmap:

```go
revocableZone[strings.Replace(k, revocableZoneLabelPrefix, "", 1)] = v.(string)   // line 71
...
if d, err := time.ParseDuration(period.(string)); err == nil {                     // line 76
```

`framework.Arguments` is `map[string]interface{}` populated from YAML, so any `tdm.revocable-zone.*` or `tdm.evict.period` value that isn't quoted as a string (e.g. a bare number) causes `interface conversion: interface {} is int, not string`. The panic is recovered once by the top-level session handler and then **repanics**, killing the `volcano-scheduler` process, which restarts into the same bad config and crash-loops indefinitely — taking down scheduling cluster-wide, not just TDM-scheduled workloads.

This is inconsistent with the rest of the codebase: `pkg/scheduler/framework/arguments.go` provides `Arguments.GetString()` (and `GetInt`/`GetFloat64`/`GetBool`) specifically to do this safely via comma-ok assertions with a `klog.Warningf` fallback. Every other plugin I checked uses these helpers; `tdm.go` bypasses them with raw assertions.

1. Deploy Volcano normally (tested on v1.15.1).
2. Edit `volcano-scheduler-configmap` in `volcano-system` to add the `tdm` plugin with a non-string argument:
   ```yaml
   tiers:
   - plugins:
     - name: priority
     - name: gang
     - name: conformance
     - name: tdm
       arguments:
         tdm.evict.period: 60      # bare number instead of "60s"
   - plugins: [...]
   ```
3. `kubectl rollout restart deployment volcano-scheduler -n volcano-system`
4. `kubectl get pods -n volcano-system` and `kubectl logs -n volcano-system deploy/volcano-scheduler`

Actual result:
```
E... iface.go:275] "Observed a panic" panic="interface conversion: interface {} is int, not string"
panic: interface conversion: interface {} is int, not string [recovered, repanicked]
  /go/src/volcano.sh/volcano/pkg/scheduler/plugins/tdm/tdm.go:76 +0x1f0
```
`volcano-scheduler` enters `CrashLoopBackOff` immediately and stays there until the configmap is manually corrected.

Expected result: a malformed `tdm` argument should be logged as a warning and ignored (falling back to defaults), matching every other plugin's argument-parsing behavior, not crash the scheduler process.

`v1.15.1` (Helm chart `volcano-sh/volcano` version `1.15.1`, image `volcanosh/vc-scheduler:v1.15.1`)

Reproduced on a local cluster: Kubernetes v1.34.8, single node, OrbStack container runtime, darwin/arm64 host. Not host/OS-specific — pure Go type-assertion panic, will reproduce identically anywhere.

CodeBuildder:
#### What type of PR is this?
/kind bug

`tdm`'s `New()` used unchecked type assertions (`v.(string)`) on plugin arguments read from the scheduler configmap. Since `framework.Arguments` is `map[string]interface{}` decoded from YAML, a non-string value for `tdm.evict.period` or any `tdm.revocable-zone.*` key (e.g. a bare, unquoted number) causes an unrecovered `interface conversion` panic. The panic is recovered once by the top-level session handler and then repanics, crash-looping the entire `volcano-scheduler` process — not just TDM-scheduled workloads.

This PR switches both assertions to comma-ok checks with a `klog.Warningf` fallback (falling back to the default value), matching the existing `Arguments.GetString()` convention already used by every other plugin's argument parsing (`pkg/scheduler/framework/arguments.go`).

Added `Test_New_MalformedArguments`, which reproduces the exact crash scenario from the issue and confirms `New()` now degrades gracefully instead of panicking, plus a well-formed-arguments case to confirm no behavior change for valid config.

Fixes #5806

Verified live: reproduced the crash against a real v1.15.1 deployment (`tdm.evict.period: 60` in the scheduler configmap immediately crash-loops `volcano-scheduler`), then confirmed this fix resolves it in the same setup.

```release-note
Fixed a bug where a malformed (non-string) `tdm` plugin argument in the scheduler configuration would crash-loop the entire volcano-scheduler process instead of being rejected gracefully.
```

### DIFF
--- a/pkg/scheduler/plugins/tdm/tdm.go
+++ b/pkg/scheduler/plugins/tdm/tdm.go
@@ -67,13 +67,13 @@ func New(args framework.Arguments) framework.Plugin {
 	evictPeriod := time.Minute
 
 	for k, v := range args {
-		if strings.Contains(k, revocableZoneLabelPrefix) {
+		if strings.HasPrefix(k, revocableZoneLabelPrefix) {
 			value, ok := v.(string)
 			if !ok {
 				klog.Warningf("Could not parse argument: %v for key %s to string, skip this revocable zone", v, k)
 				continue
 			}
-			revocableZone[strings.Replace(k, revocableZoneLabelPrefix, "", 1)] = value
+			revocableZone[strings.TrimPrefix(k, revocableZoneLabelPrefix)] = value
 		}
 	}
 
--- a/pkg/scheduler/plugins/tdm/tdm.go
+++ b/pkg/scheduler/plugins/tdm/tdm.go
@@ -68,13 +68,22 @@ func New(args framework.Arguments) framework.Plugin {
 
 	for k, v := range args {
 		if strings.Contains(k, revocableZoneLabelPrefix) {
-			revocableZone[strings.Replace(k, revocableZoneLabelPrefix, "", 1)] = v.(string)
+			value, ok := v.(string)
+			if !ok {
+				klog.Warningf("Could not parse argument: %v for key %s to string, skip this revocable zone", v, k)
+				continue
+			}
+			revocableZone[strings.Replace(k, revocableZoneLabelPrefix, "", 1)] = value
 		}
 	}
 
-	if period, ok := args[evictPeriodLabel]; ok {
-		if d, err := time.ParseDuration(period.(string)); err == nil {
+	var periodStr string
+	args.GetString(&periodStr, evictPeriodLabel)
+	if periodStr != "" {
+		if d, err := time.ParseDuration(periodStr); err == nil {
 			evictPeriod = d
+		} else {
+			klog.Warningf("Could not parse argument %q for key %s to duration, using default %v", periodStr, evictPeriodLabel, evictPeriod)
 		}
 	}
 
--- a/pkg/scheduler/plugins/tdm/tdm_test.go
+++ b/pkg/scheduler/plugins/tdm/tdm_test.go
@@ -104,6 +104,57 @@ func Test_parseRevocableZone(t *testing.T) {
 	}
 }
 
+func Test_New_MalformedArguments(t *testing.T) {
+	tests := []struct {
+		name             string
+		args             framework.Arguments
+		wantEvictPeriod  time.Duration
+		wantRevocableLen int
+	}{
+		{
+			// Reproduces the crash from https://github.com/volcano-sh/volcano/issues/5806:
+			// a bare (unquoted) number in the scheduler configmap used to panic with
+			// "interface conversion: interface {} is int, not string" instead of being
+			// rejected gracefully.
+			name: "non-string evict period falls back to default",
+			args: framework.Arguments{
+				evictPeriodLabel: 60,
+			},
+			wantEvictPeriod:  time.Minute,
+			wantRevocableLen: 0,
+		},
+		{
+			name: "non-string revocable zone value is skipped",
+			args: framework.Arguments{
+				revocableZoneLabelPrefix + "rz1": 1000,
+			},
+			wantEvictPeriod:  time.Minute,
+			wantRevocableLen: 0,
+		},
+		{
+			name: "well-formed arguments still parse correctly",
+			args: framework.Arguments{
+				evictPeriodLabel:                 "2m",
+				revocableZoneLabelPrefix + "rz1": "10:00-21:00",
+			},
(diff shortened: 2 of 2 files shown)


CodeBuildder:
fixes the crash in #5806 by replacing the raw type assertions with the existing GetString helper, matching the pattern other plugins already use. Tests cover the malformed cases and confirm well-formed args still work. CC @merryzhou @wangyang0616.

CodeBuildder:
@merryzhou @shinytang6 Different PR from the vGPU one - this fixes a panic in the tdm plugin when a plugin argument isn't a string (#5806). Take a look whenever it's convenient.

pkg/scheduler/plugins/tdm/tdm.go
```diff
@@ -68,13 +68,22 @@ func New(args framework.Arguments) framework.Plugin {
 
 	for k, v := range args {
 		if strings.Contains(k, revocableZoneLabelPrefix) {
```
miantalha45: could we use `strings.HasPrefix` and `strings.TrimPrefix` here instead of Contains and Replace?
Contains also accepts keys where `revocableZoneLabelPrefix` appears in the middle of the key, which is not a valid revocable-zone argument. `HasPrefix` makes the accepted configuration shape explicit and avoids deriving an unintended zone name.

pkg/scheduler/plugins/tdm/tdm_test.go
```diff
@@ -104,6 +104,57 @@ func Test_parseRevocableZone(t *testing.T) {
...
+		},
+		{
+			name: "well-formed arguments still parse correctly",
+			args: framework.Arguments{
+				evictPeriodLabel:                 "2m",
+				revocableZoneLabelPrefix + "rz1": "10:00-21:00",
+			},
+			wantEvictPeriod:  2 * time.Minute,
+			wantRevocableLen: 1,
+		},
```
miantalha45, line 142: Also add this case to cover the prefix correction:
```go
{
    name: "revocable zone prefix must be at the start of the key",
    args: framework.Arguments{
        "other-" + revocableZoneLabelPrefix + "rz1": "10:00-21:00",
    },
    wantEvictPeriod:  time.Minute,
    wantRevocableLen: 0,
},
```

pkg/scheduler/plugins/tdm/tdm.go
CodeBuildder: Good catch, you're right. HasPrefix/TrimPrefix is the correct check here, Contains would happily match the prefix wherever it showed up in the key. Fixed in c47a86347, and I added the test case you suggested below too.

pkg/scheduler/plugins/tdm/tdm_test.go
CodeBuildder: Added, thanks for writing it out.

miantalha45:
@CodeBuildder Please check the CI failure.

CodeBuildder:
@miantalha45 could you take a look at it now?

miantalha45 (approved):
/lgtm
Cc @JesseStutler

CodeBuildder:
Rebased onto master to get rid of the merge commit the do-not-merge bot flagged, all good now. One thing I can't get past myself: the workflow runs are stuck on "pending" approval since this is my first PR here. @miantalha45 @merryzhou @shinytang6 could one of you approve the run whenever you get a chance? https://github.com/volcano-sh/volcano/actions/runs/31622114365

CodeBuildder:
This already has an lgtm from @miantalha45, just needs an approver to land it. Also looked into the CI failure from before, it's the MPI plugin test (mpi_plugin.go:81) failing, nothing to do with the tdm change here, just flaky.

cc @hajnalmt, would appreciate a look.

CodeBuildder:
Rebased this to drop a merge commit that had snuck in (the bot flagged it as do-not-merge), history's linear now, same two commits otherwise.

cc @hajnalmt, since you've been the one landing most of the scheduler fixes lately, if you get a chance to take a look now that the lgtm's already on it.

PR #5809, opened 2026-08-01
fix(webhooks): detect hierarchical queue cycles and bound capability walks

### LINKED ISSUE #5807: Hierarchical queue cycle bypasses admission validation and hangs the admission webhook indefinitely (CPU-leak DoS)
### Description

`validateHierarchicalQueue` (`pkg/webhooks/admission/queues/validate/validate_queue.go:373`) is supposed to reject cyclic queue hierarchies, but its ancestor-chain walk reads the queue being updated via `config.QueueLister.Get()` — which returns the **stale, pre-update** cached copy when the walk revisits it. Re-parenting an existing queue onto one of its own descendants is accepted, because the walk sees the queue's *old* parent chain (which validly reached `root`) instead of the new cyclic one.

Once a cycle exists, [`findNearestAncestorCapability`](https://github.com/volcano-sh/volcano/blob/v1.15.1/pkg/webhooks/admission/queues/validate/validate_queue.go#L442) (a `for` loop with no cycle detection or bound) spins forever on the next capability-changing update to any queue in the cycle. [`findSubtreeMaxCapability`](https://github.com/volcano-sh/volcano/blob/v1.15.1/pkg/webhooks/admission/queues/validate/validate_queue.go#L464) has the same gap but recursively, and would stack-overflow-crash the process under a slightly different trigger shape.

Impact:
- The stuck admission goroutine pins ~100%+ CPU **permanently** (confirmed via `docker stats`: sustained 117-118% CPU, flat ~23-26MB memory) and never returns — no context-cancellation check in the loop. The pod stays `1/1 Running/Ready` throughout, so standard health checks never catch it.
- No resource limits are set on `volcano-admission` by the default Helm chart, so repeating the trigger a few times exhausts the pod's available CPU, degrading or blocking *all* queue/job/pod admission cluster-wide.
- Compounding effect: `validateQueueDeleting` walks children the same way, so once the cycle exists, **neither queue in the cycle can be deleted** — each reports the other as a child, permanently, until one is re-parented back to something acyclic.

```bash
kubectl apply -f - <<'EOF'
kind: Queue
metadata: {name: cyc-a}
spec: {parent: root, reclaimable: true}
---
kind: Queue
metadata: {name: cyc-b}
spec: {parent: cyc-a, reclaimable: true}
EOF

# Re-parent cyc-a onto its own child -- accepted, should be rejected:
kubectl patch queue cyc-a --type=merge -p '{"spec":{"parent":"cyc-b"}}'

# Any capability update on either queue now hangs the webhook:
kubectl patch queue cyc-b --type=merge -p '{"spec":{"capability":{"cpu":"4"}}}'
```

Actual result:
- Step 2 (`kubectl patch queue cyc-a ...`) succeeds with no error, silently creating a real cycle (`cyc-a`→`cyc-b`→`cyc-a`) in etcd.
- Step 3 (`kubectl patch queue cyc-b ...`) hangs: `Error from server (InternalError): ... failed calling webhook "validatequeue.volcano.sh": ... context deadline exceeded`.
- `kubectl logs -n volcano-system deploy/volcano-admission` shows `Admitting UPDATE queue cyc-b.` logged and then nothing — no completion, ever.
- The admission pod's CPU usage jumps to and holds at ~117-118%, indefinitely, until manually restarted (`kubectl rollout restart deployment volcano-admission -n volcano-system`).

Expected result: step 2 should be rejected at admission with a clear cycle error (e.g. `queue cyc-a: setting parent to cyc-b would create a cycle`), and the capability-walk helpers should be defensively bounded regardless, as defense in depth against a cycle reaching the cluster through any other path.

`v1.15.1` (Helm chart `volcano-sh/volcano` version `1.15.1`, image `volcanosh/vc-webhook-manager:v1.15.1`)

Reproduced on a local cluster: Kubernetes v1.34.8, single node, OrbStack container runtime, darwin/arm64 host. Cleanup after reproducing requires patching one queue's `spec.parent` back to `root` (or another acyclic value) before either queue can be deleted — plain `kubectl delete` fails on both with "has 1 child queues" as long as the cycle stands.

CodeBuildder:
#### What type of PR is this?
/kind bug

`validateQueueDepth`'s ancestor-chain walk fetches every queue in the chain via `config.QueueLister`, including the queue currently being validated itself if the walk ever revisits it. On UPDATE, that lookup returns the **stale, pre-update** copy of the queue (the informer cache hasn't observed the in-flight change yet), so re-parenting an existing queue onto one of its own descendants was silently accepted: the walk escaped through the queue's old, valid parent chain to `root` instead of detecting the new cycle.

Once such a cycle exists, `findNearestAncestorCapability` (a `for` loop with no cycle detection or bound) spins forever on the next capability-changing update to any queue in the cycle, and `findSubtreeMaxCapability` has the same gap recursively. Live-verified impact: the stuck admission goroutine pins ~100%+ CPU permanently (no context-cancellation check), the pod stays `Running/Ready` throughout so health checks never catch it, and — since no resource limits are set on `volcano-admission` by the default Helm chart — repeating the trigger degrades or blocks all queue/job/pod admission cluster-wide.

This PR:
1. Adds a same-name check during `validateQueueDepth`'s walk: since only one queue's parent changes per admission request, any newly introduced cycle must pass back through that queue's own name, regardless of what the stale cache says about the rest of the chain. This closes the exploit at the source.
2. As defense in depth against a cycle reaching the cluster through any other path (e.g. one already present from before this fix), bounds `findNearestAncestorCapability` and `findSubtreeMaxCapability` with a visited-set, so a corrupted hierarchy fails fast (logs a warning, returns "not found"/0) instead of spinning forever.

Fixes #5807

- Verified live end-to-end, not just via unit tests: built `vc-webhook-manager` from this branch, deployed it to a real cluster, and re-ran the exact repro from the issue. Before the fix: `kubectl patch queue cyc-a ... parent: cyc-b` succeeds silently. After the fix: it's rejected immediately with `queue cyc-a: setting parent to cyc-b would create a cycle in the queue hierarchy`.
- Added `TestValidateQueueDepthCycleDetection`, which reproduces the stale-lister exploit directly (constructs the in-flight "new" object the same way `AdmitQueues` does, against a lister that still holds the pre-update state) and also confirms a legitimate, non-cyclic parent change is still accepted.
- Added `TestFindNearestAncestorCapabilityCycleSafety` / `TestFindSubtreeMaxCapabilityCycleSafety`, which set up a pre-existing cyclic pair directly in the fake lister and assert both helpers return within a bounded timeout instead of hanging. Confirmed these tests actually catch the regression: reverting just `validate_queue.go` while keeping the new tests causes the suite to fail/hang with a full goroutine dump, exactly the failure mode the fix prevents.
- `validateQueueDeleting`'s "can't delete a queue with children" behavior is left as-is; once a cycle can no longer be created, its side effect described in the issue (both queues in an existing cycle being permanently undeletable) can't newly occur either.

```release-note
Fixed a bug where re-parenting a queue onto one of its own descendants was silently accepted instead of rejected, which could later hang the admission webhook indefinitely (pinning its CPU) and make the queues involved permanently undeletable.
```

### DIFF
--- a/pkg/webhooks/admission/queues/validate/validate_queue.go
+++ b/pkg/webhooks/admission/queues/validate/validate_queue.go
@@ -437,8 +437,19 @@ func getSingleResource(r *api.Resource, name v1.ResourceName) float64 {
 
 // Recursively searches for the ancestor queue that recently defines the capability
 func findNearestAncestorCapability(q *schedulingv1beta1.Queue, rname v1.ResourceName) (float64, bool) {
+	// visited guards against a corrupted (cyclic) queue hierarchy reaching this walk by any
+	// path other than the admission check in validateQueueDepth, e.g. a cycle already present
+	// in the cluster from before that check existed. Without it, a cycle here would spin
+	// forever, permanently pinning the admission webhook's CPU.
+	visited := map[string]bool{q.Name: true}
 	parent := q.Spec.Parent
 	for parent != "" && parent != "root" {
+		if visited[parent] {
+			klog.Warningf("Detected a cycle in the queue hierarchy while resolving ancestor capability for queue %s at %s, aborting walk", q.Name, parent)
+			return 0, false
+		}
+		visited[parent] = true
+
 		pq, err := config.QueueLister.Get(parent)
 		if err != nil {
 			return 0, false
@@ -459,6 +470,21 @@ func findNearestAncestorCapability(q *schedulingv1beta1.Queue, rname v1.Resource
 
 // Recursively searches for the maximum capability value of a descendant queue
 func findSubtreeMaxCapability(q *schedulingv1beta1.Queue, rname v1.ResourceName) float64 {
+	return findSubtreeMaxCapabilityVisited(q, rname, map[string]bool{})
+}
+
+// findSubtreeMaxCapabilityVisited does the actual recursive walk. visited guards against a
+// corrupted (cyclic) queue hierarchy causing unbounded recursion; see findNearestAncestorCapability
+// for why this can't rely solely on the admission-time cycle check. Since every queue has exactly
+// one parent, the child graph is a proper tree/forest, so a single visited set shared across the
+// whole walk is safe: a legitimate hierarchy can never revisit the same queue name twice.
+func findSubtreeMaxCapabilityVisited(q *schedulingv1beta1.Queue, rname v1.ResourceName, visited map[string]bool) float64 {
+	if visited[q.Name] {
+		klog.Warningf("Detected a cycle in the queue hierarchy while resolving subtree max capability at queue %s, aborting walk", q.Name)
+		return 0
+	}
+	visited[q.Name] = true
+
 	if q.Spec.Capability != nil {
 		res := api.NewResource(q.Spec.Capability)
 		v := getSingleResource(res, rname)
@@ -473,7 +499,7 @@ func findSubtreeMaxCapability(q *schedulingv1beta1.Queue, rname v1.ResourceName)
 		klog.V(5).Infof("Failed to get child queues for queue %s: %v", q.Name, err)
 	} else {
 		for _, cq := range children {
-			v := findSubtreeMaxCapability(cq, rname)
+			v := findSubtreeMaxCapabilityVisited(cq, rname, visited)
 			if v > maxV {
 				maxV = v
 			}
@@ -502,6 +528,16 @@ func validateQueueDepth(queue *schedulingv1beta1.Queue) error {
 	parent := queue.Spec.Parent
 
 	for parent != "" && parent != "root" {
+		// If the walk up the ancestor chain ever reaches this queue's own name again,
+		// setting Spec.Parent to queue.Spec.Parent would introduce a cycle. This must be
+		// checked by name rather than relying solely on the depth bound below: when queue
+		// is itself being updated, config.QueueLister still holds its pre-update (stale)
+		// value the moment the walk revisits it, so a cycle can otherwise walk straight
+		// through queue's old, valid parent chain and reach "root" undetected.
+		if parent == queue.Name {
+			return fmt.Errorf("queue %s: setting parent to %s would create a cycle in the queue hierarchy",
+				queue.Name, queue.Spec.Parent)
+		}
 		depth++
 		if depth > config.MaxQueueDepth {
 			return fmt.Errorf("queue %s exceeds the maximum allowed depth of %d", queue.Name, config.MaxQueueDepth)
--- a/pkg/webhooks/admission/queues/validate/validate_queue.go
+++ b/pkg/webhooks/admission/queues/validate/validate_queue.go
@@ -320,6 +320,18 @@ func validateQueueDeleting(queueName string) error {
 	}
 
 	if len(childQueueNames) > 0 {
+		// A queue that's part of a hierarchy cycle predating the check in validateQueueDepth
+		// (e.g. from before this fix was deployed) will always appear to have children, since
+		// each queue in the cycle genuinely does list the next one as a child. That's a real,
+		// if confusing, consequence of the underlying corruption rather than a bug in this
+		// check itself, so give a diagnosis and a concrete recovery step instead of leaving the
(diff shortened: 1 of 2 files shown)


CodeBuildder:
Fixes the DoS in #5807 by adding cycle detection to validateQueueDepth's ancestor walk (previously fooled by stale lister data on UPDATE, letting a re-parent-onto-descendant slip through) and bounding the two capability-walk helpers with a visited-set as defense in depth. Verified live end-to-end against a real cluster in addition to the added unit tests — before the fix the exploit patch succeeds silently, after it's rejected immediately. CC @Thor-wl @merryzhou @wangyang0616

CodeBuildder:
Thanks for the thorough review — pushed a follow-up commit addressing all of it:

**Must-address #1 (deletion path):** Confirmed the diff didn't touch deletion, and that's intentional — `validateQueueDeleting`'s "has child queues" is correct given a cycle (each queue in a cycle genuinely does list the other as a child), so I didn't try to make deletion silently work around it. What I added instead: `validateQueueDeleting` now detects when that's actually a cycle and says so explicitly, with a concrete recovery step (re-parent one queue to an acyclic value first, then delete), instead of leaving an operator stuck on a generic message with no way to tell a real "delete children first" case apart from hierarchy corruption. Covered by `TestValidateQueueDeletingCycleMessage`. Note this is for the pre-existing-cycle case (upgrading a cluster that already has one from before this fix) — going forward, cycles can't be created at all, so this situation shouldn't newly occur.

**Must-address #2 (silent fallback):** Agreed this was the sharper risk of the two. `findNearestAncestorCapability` and `findSubtreeMaxCapability` now return an error when the walk/recursion exceeds `config.MaxQueueDepth` instead of silently degrading to "no capability found" behind a warning log. Both callers propagate it, so a cycle reaching either helper by any path now rejects the admission request outright rather than risking a queue's hierarchy validation being silently wrong for as long as the corruption persists.

**Question #3 (two mechanisms):** You're right, no reason for two. Replaced both visited-maps with the same `config.MaxQueueDepth` depth-counter `validateQueueDepth` already uses — cheaper (no per-call map allocation) and consistent. Renamed `findSubtreeMaxCapabilityVisited` → `findSubtreeMaxCapabilityAtDepth` to match, which also resolves the nit on naming (#6).

**Question #4 (self-check scope):** Confirmed by writing the test rather than just reasoning about it: `TestFindNearestAncestorCapabilityCycleNotInvolvingSelf` sets up a pre-existing x↔y cycle plus a third queue `q` newly parented onto `x`. The self-name check can't catch this (q was never part of the x/y cycle) — it's the pre-existing depth bound that does. Added an explicit comment in `validateQueueDepth` calling this out so it doesn't read as accidental.

**Nit #5 (mixed branches):** Added `TestFindSubtreeMaxCapabilityMixedBranches` — cyclic branch evaluated first (deliberately), then confirms an unrelated valid sibling branch still resolves correctly and isn't affected. This also incidentally surfaced a real test-isolation bug while writing it: `config.QueueInformer` is process-global across this file's tests, and an earlier test's stale informer was silently shadowing my fresh `config.QueueLister` in the new tests via `GetQueuesByParent`'s informer-index preference. Fixed by following the existing `setupQueueInformerWithIndex` pattern already used elsewhere in the file.

Also re-verified live end-to-end against a real cluster after these changes (rebuilt the image, redeployed, re-ran the exploit) — the cycle rejection still works correctly. Didn't live-test the new deletion-message wording specifically, since doing that now requires deliberately bypassing the (working) webhook to manufacture a pre-existing cycle first, which felt like the wrong kind of risk for a diagnostics-only message change on a live cluster; that one's covered by the unit test only.

Full diff still green on unit tests + the original e2e suite. CC @Thor-wl @merryzhou @wangyang0616

CodeBuildder:
@kingeasternsun @merryzhou Third and unrelated one from me, merryzhou - no pressure on any of them, just flagging: this one closes a gap where a hierarchical queue cycle can bypass admission validation (#5807).

CodeBuildder:
Bumping this one, CI's been green a while and it hasn't had a review yet. Closes a real DoS (#5807) so didn't want it to get lost.

cc @hajnalmt, appreciate a look when you can.

PR #5810, opened 2026-08-01
feat(overcommit): add queue-scoped admission limits

### LINKED ISSUE #5687: Add opt-in queue-scoped overcommit admission controls

The overcommit plugin currently uses one cluster-wide `overcommit-factor` to decide whether a job can enter the `Inqueue` phase. This applies the same admission and burst policy to every queue.

Some clusters have queues with different workload characteristics. For example, a production queue may require predictable admission while a batch queue can tolerate a larger backlog of admitted-but-waiting jobs. There is currently no supported way to configure different overcommit admission limits per queue.

Allowing queues to independently override the cluster-wide factor without safeguards would be unsafe: queues could collectively admit more work than the administrator's cluster-wide policy allows.

Add an opt-in, feature-gated queue-scoped overcommit model. The default behavior must remain the existing cluster-wide `overcommit-factor`.

Add an optional `spec.overcommitFactor` field to the Queue API. Enable the behavior only when a `QueueScopedOvercommit` feature gate is enabled. otherwise the field is rejected. A queue without this field continues to use the current cluster-wide behavior.

When the feature gate is enabled, a queue could define an overcommit admission factor. The scheduler would enforce both of the following checks for every job entering `Inqueue`:

1. The existing global check: `used resources + global inqueue resources + job minimum resources` must not exceed `cluster resources * global overcommit factor`.
2. A queue-level check: `queue allocated resources + queue inqueue resources + job minimum resources` must not exceed the queue's effective admission budget.

The queue-level budget should be bounded by the queue's existing resource policy:

```
min(queue realCapability, effective deserved resources * effective queue factor)
```

`effective deserved resources` is `max(queue deserved, queue guarantee)`, consistent with the Capacity plugin. `realCapability` uses Volcano's existing effective-capability calculation, so an omitted configured capability still receives the current derived limit.

The effective queue factor is bounded by the cluster administrator's configuration:

```
effective queue factor = min(
  Queue.spec.overcommitFactor,
  max-queue-overcommit-factor,
  global overcommit-factor,
)
```

`max-queue-overcommit-factor` is a scheduler-controlled plugin argument and must not be greater than the global `overcommit-factor`. This prevents a queue from bypassing the cluster-wide admission policy.

A queue may set `spec.overcommitFactor` only when it has a non-empty `spec.deserved`. Resource dimensions omitted from `deserved` do not receive queue-level overcommit. An omitted `spec.capability` uses the existing derived `realCapability`. no additional capability field is required.

For hierarchical queues, the scheduler tracks inqueue resources for the leaf queue and propagates them to every ancestor. It checks every ancestor that has `spec.overcommitFactor` configured. The factor changes only `Inqueue` admission: it does not modify `capability`, `deserved`, `guarantee`, allocation, or reclaim behavior. Existing guarantees remain protected, and reclaim continues to use the Capacity plugin's current rules.

The implementation maintains global, queue, and ancestor inqueue counters incrementally. it does not rescan all queues or jobs for each enqueue attempt.

No new field-level RBAC mechanism is proposed initially. The new field inherits the existing permission to update the cluster-scoped Queue resource. Deployments should grant Queue write access only to cluster administrators. A separate Queue subresource can be considered later if field-level delegation is required.

Maintainers, please review this proposed API and behavior and let me know whether it is acceptable or what should change before implementation.

### Additional context

The current overcommit implementation maintains a single global `inqueueResource` and computes one idle budget from the cluster's total resources. A queue-scoped design should preserve this global check as the authoritative protection against cluster-wide over-admission.

The queue hierarchy is normally limited to five levels by the webhook manager, so an incremental queue-plus-ancestor check should add bounded overhead. The runtime cost should be proportional to queue depth and the small number of requested resource dimensions, rather than the total number of queues or jobs.

This proposal follows issue #4530, which was marked deprecated after the root-queue handling changes in #4973. It is not a request to reintroduce the old root-queue behavior.

### Documentation Updates

miantalha45:
#### What type of PR is this?

/kind feature

Adds the alpha `QueueScopedOvercommit` feature. It gives administrators an optional Queue-level limit for PodGroups entering `Inqueue`.

- Adds the `volcano.sh/overcommit-factor` Queue annotation.
- Keeps the existing global `overcommit-factor` as the cluster-wide admission limit.
- Adds `max-queue-overcommit-factor` as an administrator-controlled cap for Queue factors.
- Validates Queue annotations through the Queue admission webhook when the feature gate is enabled.
- Tracks allocated and Inqueue resources for each Queue and annotated ancestors during a scheduling session.
- Checks annotated ancestors when hierarchical Queues are enabled.
- Adds unit tests, feature-gated E2E coverage, and user documentation.

Fixes #5687

- The feature is disabled by default behind `QueueScopedOvercommit`.
- Enable the gate on both `vc-scheduler` and `vc-webhook-manager`.
- The E2E suite runs with:

  ```bash
  make e2e-test-queue-overcommit
  ```

#### Live Cluster Verification

Tested on a local Kind cluster using locally built scheduler and webhook-manager images.

Test configuration:

- Global `overcommit-factor`: `2.0`
- `max-queue-overcommit-factor`: `1.5`
- Queue `deserved.cpu`: `2`
- Queue annotation: `volcano.sh/overcommit-factor: "1.5"`
- Effective Queue CPU admission budget: `3`

Results:

1. A first PodGroup requesting `2 CPU` entered `Inqueue`.
2. A second PodGroup requesting `2 CPU` remained `Pending`.
3. The scheduler recorded:

   ```text
   queue overcommit admission budget insufficient for queue-overcommit-limited: insufficient cpu
   ```

**Screenshots:**

#### deployed feature gate and scheduler overcommit configuration:
image

#### accepted PodGroup is Inqueue and rejected PodGroup is Pending
image

#### rejected PodGroup event showing insufficient Queue admission budget
```yaml
queue overcommit admission budget insufficient for queue-overcommit-limited: insufficient cpu
```
image

```release-note
Adds the alpha QueueScopedOvercommit feature gate. When enabled, administrators can configure an additional Queue-level Inqueue admission limit with the volcano.sh/overcommit-factor annotation.
```


miantalha45:
/cc @devzizu 
Can you please have a look on this PR?

miantalha45:
Hi @JesseStutler @hajnalmt,

I tested this feature on a local Kind cluster and added the live verification steps and screenshots to the PR description.

The first PodGroup entered `Inqueue`. The second PodGroup remained `Pending` with the expected Queue admission-budget event.

PR #5817, opened 2026-08-02
fix: wire event recorder into scheduler framework

### LINKED ISSUE #5812: Scheduler can panic on DRA devices with binding conditions
### Description

found while reviewing the Kubernetes 1.36 DRA integration.

Volcano registers the upstream DynamicResources PreBind plugin, but its scheduler framework returns a nil event recorder. When a selected DRA device has pending binding conditions, the upstream plugin uses that recorder and hits a nil pointer panic.

This terminates the scheduler process, not just the current scheduling cycle.

1. Run Volcano master on Kubernetes 1.36 with DRA enabled.
2. Advertise a DRA device with `bindingConditions: ["Ready"]`.
3. Create a matching DeviceClass, ResourceClaim and pod using `schedulerName: volcano`.
4. Leave the allocated device's `Ready` condition unset so PreBind has to wait.
5. Volcano enters DynamicResources PreBind and calls the nil event recorder.

I found this through source review and have not run the full cluster reproducer yet.

Execution path:

1. Volcano registers DynamicResources as a PreBind plugin:
   https://github.com/volcano-sh/volcano/blob/a4b775ad2c2d456ca85233f35f305e16707126b3/pkg/scheduler/plugins/predicates/predicates.go#L621-L654

2. PredicatesPlugin invokes the upstream PreBind implementation:
   https://github.com/volcano-sh/volcano/blob/a4b775ad2c2d456ca85233f35f305e16707126b3/pkg/scheduler/plugins/predicates/predicates.go#L886-L906

3. PreBind runs from the bind worker without panic recovery:
   https://github.com/volcano-sh/volcano/blob/a4b775ad2c2d456ca85233f35f305e16707126b3/pkg/scheduler/cache/cache.go#L1409-L1446
   https://github.com/volcano-sh/volcano/blob/a4b775ad2c2d456ca85233f35f305e16707126b3/pkg/scheduler/cache/cache.go#L1464-L1473

4. Upstream DRA uses the event recorder when binding conditions are pending:
   https://github.com/kubernetes/kubernetes/blob/v1.36.1/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go#L1371-L1386

5. Volcano's framework adapter returns nil:
   https://github.com/volcano-sh/volcano/blob/a4b775ad2c2d456ca85233f35f305e16707126b3/pkg/scheduler/plugins/util/k8s/framework.go#L135-L138

There is no matching open issue or fix that I could find. #5733 checks extension-point compatibility but does not wire an event recorder.

Expected: PreBind emits a pending-binding event and waits until the device is ready or the binding timeout expires.

Actual: `EventRecorder().WithLogger(...)` dereferences nil and panics the scheduler.

master at a4b775ad2c2d456ca85233f35f305e16707126b3

Kubernetes dependency: v1.36.1

`DRADeviceBindingConditions` and `DRAResourceClaimDeviceStatus` are enabled by default in Kubernetes 1.36.

Likely fix is to pass a real EventRecorderLogger into the Kubernetes framework adapter and cover the binding-condition PreBind path with a regression test.

avinxshKD:
#### What type of PR is this?

/kind bug

Passes the scheduler event recorder into the Kubernetes framework adapter.

The DynamicResources PreBind plugin emits an event while DRA binding conditions are pending. The adapter currently returns a nil recorder, causing the bind worker to panic and terminate the scheduler.

Also adds a regression test covering the recorder path through the predicates plugin.

Fixes #5812

Tested TestFrameworkEventRecorder on Linux.

AI assistance was used during implementation. I reviewed the final diff and ran the focused regression test on my own.

```release-note
Fix scheduler panic when DRA device binding conditions are pending.

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -140,7 +140,7 @@ type SchedulerCache struct {
 	StatusUpdater  StatusUpdater
 	PodGroupBinder BatchBinder
 
-	Recorder record.EventRecorder
+	Recorder record.EventRecorderLogger
 
 	Jobs                 map[schedulingapi.JobID]*schedulingapi.JobInfo
 	Nodes                map[string]*schedulingapi.NodeInfo
@@ -1072,7 +1072,7 @@ func (sc *SchedulerCache) AddUnassignedNumaPods(allocatedSets map[schedulingapi.
 }
 
 // EventRecorder returns the Event Recorder
-func (sc *SchedulerCache) EventRecorder() record.EventRecorder {
+func (sc *SchedulerCache) EventRecorder() record.EventRecorderLogger {
 	return sc.Recorder
 }
 
--- a/pkg/scheduler/cache/cache_mock.go
+++ b/pkg/scheduler/cache/cache_mock.go
@@ -47,7 +47,7 @@ func NewCustomMockSchedulerCache(schedulerName string,
 	evictor Evictor,
 	statusUpdater StatusUpdater,
 	PodGroupBinder BatchBinder,
-	recorder record.EventRecorder,
+	recorder record.EventRecorderLogger,
 ) *SchedulerCache {
 	options.Default() // init options first because cache may reference server opts in options
 	msc := newMockSchedulerCache(schedulerName)
--- a/pkg/scheduler/cache/interface.go
+++ b/pkg/scheduler/cache/interface.go
@@ -89,7 +89,7 @@ type Cache interface {
 	GetMetricsConf() map[string]string
 
 	// EventRecorder returns the event recorder
-	EventRecorder() record.EventRecorder
+	EventRecorder() record.EventRecorderLogger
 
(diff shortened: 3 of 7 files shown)


avinxshKD:
@hwdef @hajnalmt ptal this is high impact because the panic happens in the bind worker and terminates the scheduler process, rather than failing only the affected pod. The regression test covers the full event recorder path used by DRA PreBind.

/cc @JesseStutler

miantalha45:
@avinxshKD Please address to maintainer comment https://github.com/volcano-sh/volcano/issues/5812#issuecomment-5161451270, and Reproduce the issue on a real cluster.

avinxshKD:
@miantalha45 you don't need to worry pal. I already communicated to him regarding. 
All the best.

miantalha45:
> @miantalha45 you don't need to worry pal. I already communicated to him regarding. All the best.

ok mate. thanks for clarification, i am just reviewing some PRs as per maintainer direction.

PR #5818, opened 2026-08-02
scheduler: remove redundant preemptor requeue

### LINKED ISSUE #5776: Possible redundant preemptor re-queue after a successful preemption round?

I'm new to Volcano Scheduler and I'm trying to get a overall understand of schedule workflow. 

While reading the Execute function in pkg/scheduler/actions/preempt/preempt.go, I noticed that a preemptor job is pushed back into the priority queue after a successful assignment:

```go
for {
	preemptors := preemptorsMap[queue.UID]

	// If no preemptors, no preemption.
	if preemptors == nil || preemptors.Empty() {
		klog.V(4).Infof("No preemptors in Queue <%s>, break.", queue.Name)
		break
	}

	preemptorJob := preemptors.Pop().(*api.JobInfo)

	stmt := framework.NewStatement(ssn)
	var assigned bool
	var err error
	for {
		// If job is not request more resource, then stop preempting.
		if !ssn.JobStarving(preemptorJob) {
			break
		}

		// If not preemptor tasks, next job.
		if preemptorTasks[preemptorJob.UID].Empty() {
			klog.V(3).Infof("No preemptor task in job <%s/%s>.",
				preemptorJob.Namespace, preemptorJob.Name)
			break
		}

		preemptor := preemptorTasks[preemptorJob.UID].Pop().(*api.TaskInfo)

		assigned, err = pmpt.preempt(ssn, stmt, preemptor, func(task *api.TaskInfo) bool {
			// Ignore non running task.
			if !api.PreemptableStatus(task.Status) {
				return false
			}
			// BestEffort pod is not supported to preempt unBestEffort pod.
			if preemptor.BestEffort && !task.BestEffort {
				return false
			}
			if !task.Preemptable {
				return false
			}
			job, found := ssn.Jobs[task.Job]
			if !found {
				return false
			}
			// Preempt other jobs within queue
			return job.Queue == preemptorJob.Queue && preemptor.Job != task.Job
		}, ph)
		if err != nil {
			klog.V(3).Infof("Preemptor <%s/%s> failed to preempt Task , err: %s", preemptor.Namespace, preemptor.Name, err)
		}
	}

	// Commit changes only if job is pipelined, otherwise try next job.
	if ssn.JobPipelined(preemptorJob) {
		hasEvictions := stmt.HasEvictions()
		stmt.Commit()
		if hasEvictions {
			metrics.RegisterEvictionTransaction(pmpt.Name())
		}
	} else {
		stmt.Discard()
		continue
	}

	if assigned {
		preemptors.Push(preemptorJob)    // <- here
	}
}
``` 

I'm having trouble understanding the design intent behind this re-queueing.

Before reaching this code, the inner loop must have stopped for one of the following reasons:

``` go
if !ssn.JobStarving(preemptorJob) {
    break
}
```

``` go
if preemptorTasks[preemptorJob.UID].Empty() {
    break
}
```

If the preemptor task queue is already empty, it does not appear to be replenished before the job is popped again. If the job is no longer starving, I am not sure what could make it starving again before its next pop.

Therefore, it seems that when the same job is popped again, it will immediately hit one of these conditions and exit without performing another meaningful preemption operation.

Could someone help explain the intended scenario for pushing the preemptor job back into the queue? Is there a plugin interaction or state transition that allows the job to perform additional preemption work when it is popped again?

_No response_

DarkestSky:
#### What type of PR is this?

/kind cleanup

The inner between-job preemption loop already processes a preemptor job
until either the job is no longer starving or its preemptor task queue is
empty.

The preemptor task queue is not replenished during the action. Therefore,
pushing the job back into the priority queue after a successful assignment
causes the next iteration to immediately reach one of the same exit
conditions without performing additional preemption work.

This PR removes the redundant requeue and the now-unused `assigned`
variable. It does not change the expected scheduling behavior.

Fixes #5776 

No new test case is added because this cleanup intentionally does not
change observable scheduling behavior.

The existing normal and topology-aware preemption test cases already
exercise the successful assignment path affected by this change and verify
the resulting victim evictions. All existing tests continue to pass after
the redundant requeue is removed.

Tests performed:

- `make lint`
- `make verify`
- `make unit-test`

AI disclosure: I used OpenAI Codex to help inspect the scheduler control
flow and execute the repository checks. I reviewed the final change and
test results and take responsibility for the contribution.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -178,8 +178,6 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 			preemptorJob := preemptors.Pop().(*api.JobInfo)
 
 			stmt := framework.NewStatement(ssn)
-			var assigned bool
-			var err error
 			for {
 				// If job is not request more resource, then stop preempting.
 				if !ssn.JobStarving(preemptorJob) {
@@ -195,7 +193,7 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 
 				preemptor := preemptorTasks[preemptorJob.UID].Pop().(*api.TaskInfo)
 
-				assigned, err = pmpt.preempt(ssn, stmt, preemptor, func(task *api.TaskInfo) bool {
+				_, err := pmpt.preempt(ssn, stmt, preemptor, func(task *api.TaskInfo) bool {
 					// Ignore non running task.
 					if !api.PreemptableStatus(task.Status) {
 						return false
@@ -229,10 +227,6 @@ func (pmpt *Action) Execute(ssn *framework.Session) {
 			} else {
 				stmt.Discard()
 				continue
-			}
-
-			if assigned {
-				preemptors.Push(preemptorJob)
 			}
 		}
 


hajnalmt (commented):
/area scheduling
/ok-to-test

I think you are right, this looks correct to me.

hajnalmt (approved):
/lgtm
/approve

Thanks for the PR!

PR #5821, opened 2026-08-03
agent: fix divide-by-zero and lock contention in oversubscription reporting

### LINKED ISSUE #5820: bug: Three arithmetic defects in Agent oversubscription resource calculation corrupt reported capacity on colocation nodes
## Description

The volcano-agent's oversubscription resource calculation pipeline contains three arithmetic defects that are closely related and live in the same subsystem (`pkg/agent/oversubscription/` and `pkg/agent/events/probes/noderesources/`). Together they cause incorrect oversubscription capacity to be advertised to the scheduler on colocation nodes, leading to wrong workload eviction thresholds, scheduler over- or under-allocation of offline workloads, and in one case an unbounded kube-apiserver PATCH loop.

This report was found through source code review. The execution paths and input conditions that trigger each defect are fully traced below.

---

### Defect 1 — Float64 divide-by-zero in `ShouldUpdateNodeOverSubscription` produces `+Inf` / `NaN` on first enable

`GetNodeStatusOverSubscription` (called to populate `current`) explicitly initialises both resources to `0` before reading node annotations — see [resource.go#L86-L100](https://github.com/volcano-sh/volcano/blob/5c60bf80627297e46cb51300e7cb60e89f2dee2c/pkg/agent/utils/node/resource.go#L86-L100). On the very first enable cycle, or after `DeleteNodeOverSoldStatus` clears the annotations, `current[res]` is `0`. `ShouldUpdateNodeOverSubscription` then executes:

```go
if float64(delta)/float64(current[res]) > overSubscriptionChangeStep {
```

See [policy.go#L153](https://github.com/volcano-sh/volcano/blob/5c60bf80627297e46cb51300e7cb60e89f2dee2c/pkg/agent/oversubscription/policy/policy.go#L153).

In Go, dividing by `float64(0)`:
- Returns `+Inf` when `delta > 0` → condition is always true → `UpdateOverSubscription` fires every 10 s, ignoring the 10% change guard, flooding kube-apiserver with redundant PATCH requests.
- Returns `NaN` when `delta == 0` → condition evaluates to `false` → the very first oversubscription publish is silently skipped even though the node just transitioned to active.

The 10% throttle guard (`overSubscriptionChangeStep = 0.1`) is completely bypassed in both paths when the denominator is zero.

**Affected call sites:** [extend.go#L74-L76](https://github.com/volcano-sh/volcano/blob/5c60bf80627297e46cb51300e7cb60e89f2dee2c/pkg/agent/oversubscription/policy/extend/extend.go#L74-L76) and [resource_reporter.go#L145-L150](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/handlers/oversubscription/resource_reporter.go#L145-L150).

---

### Defect 2 — Reversed exponential weights in `computeOverSubRes` make stale data dominate

The sliding window (`queueSize = 10`) applies weights 1, 2, 4, 8, 16, 32, 64, 128, 256, 512 to successive usage snapshots. The queue is a FIFO where index `0` is the oldest entry — see [internal_queue.go#L34-L41](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/oversubscription/queue/internal_queue.go#L34-L41).

The loop iterates oldest-first and doubles `initWeight` each time — see [resource_calculator.go#L133-L139](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/probes/noderesources/resource_calculator.go#L133-L139):

```go
initWeight := int64(1)
for _, usage := range historicalUsages {  // data[0] = oldest, data[9] = newest
    totalWeight += initWeight
    overSubRes[res] = overSubRes[res] + usage[res]*initWeight
    initWeight = initWeight * 2
}
// Result: oldest sample has weight 1, newest has weight 512
```

This is the **opposite** of an exponential moving average. The most-recent sample (just collected) receives weight 1, while the oldest sample (up to 90 seconds stale) receives weight 512. A usage spike that cleared 10 s ago continues to dominate the reported oversubscription capacity for the next ~80 s, causing the scheduler to under-allocate offline workloads for a full sliding-window cycle.

---

### Defect 3 — `cfgLock` should be `sync.RWMutex`, not `sync.Mutex`

`historicalUsageCalculator.cfgLock` is declared as a plain `sync.Mutex` — see [resource_calculator.go#L49](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/probes/noderesources/resource_calculator.go#L49). `getOverSubscriptionTypes`, a read-only function invoked every 10 s by `preProcess`, acquires an exclusive lock unnecessarily — see [resource_calculator.go#L148-L149](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/probes/noderesources/resource_calculator.go#L148-L149). Under ColocationConfig churn this serialises the read path against `RefreshCfg` writes, blocking oversubscription event publishing for up to the `wait.Until` period (10 s).

---

## Steps to reproduce the issue

**Defect 1 — kube-apiserver PATCH flood:**
1. Deploy volcano-agent on a colocation node with `OverSubscriptionTypes: cpu,memory`.
2. Remove the oversubscription annotations: `kubectl annotate node <node> volcano.sh/oversubscription-cpu- volcano.sh/oversubscription-memory-`
3. Watch agent logs — `UpdateOverSubscription` fires every 10 s regardless of actual usage change.

**Defect 1 — silent first publish miss:**
1. Enable oversubscription on a node that currently reports zero actual CPU/memory usage.
2. The first `ShouldUpdateNodeOverSubscription` call returns `false` (NaN comparison) and the first publish is dropped.

**Defect 2 — stale-data-dominated capacity:**
1. Enable oversubscription on a colocation node.
2. Spike CPU/memory usage for 10 s, then drop back to idle.
3. Observe that the reported oversubscription capacity (visible via node annotations) remains suppressed for ~80 s after the spike clears — because the high-usage snapshots hold weights 2–512 in the queue, while the current low-usage snapshot has weight 1.

---

## Evidence and production path

**Defect 1 — complete call path (every 10 s):**

```
resource_reporter.go:Handle()
  → shouldPatchOverSubscription()
      → r.reportTimes % reSyncPeriod == 0   [resource_reporter.go#L147]  -- false on most ticks
      → ShouldUpdateOverSubscription()       [extend.go#L74-L76]
          → GetNodeStatusOverSubscription()  [resource.go#L85-L101]
              resources[CPU]    = 0   // annotation absent on first enable
              resources[Memory] = 0
          → ShouldUpdateNodeOverSubscription(current={0,0}, new={X,Y})
              delta = X - 0 = X
              float64(X) / float64(0) = +Inf   [policy.go#L153]  BUG
              +Inf > 0.1 → true
  → UpdateOverSubscription()  // fires every 10 s, should be gated
```

Source permalinks:
- [policy.go#L145-L158](https://github.com/volcano-sh/volcano/blob/5c60bf80627297e46cb51300e7cb60e89f2dee2c/pkg/agent/oversubscription/policy/policy.go#L145-L158)
- [resource.go#L85-L101](https://github.com/volcano-sh/volcano/blob/5c60bf80627297e46cb51300e7cb60e89f2dee2c/pkg/agent/utils/node/resource.go#L85-L101)
- [extend.go#L74-L76](https://github.com/volcano-sh/volcano/blob/5c60bf80627297e46cb51300e7cb60e89f2dee2c/pkg/agent/oversubscription/policy/extend/extend.go#L74-L76)
- [resource_reporter.go#L145-L150](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/handlers/oversubscription/resource_reporter.go#L145-L150)

**Defect 2 — weight order proof:**
- Queue FIFO: [internal_queue.go#L34-L57](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/oversubscription/queue/internal_queue.go#L34-L57) — `Enqueue` appends to the end, `GetAll` returns `data[0..n]` oldest-first.
- Loop: [resource_calculator.go#L125-L144](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/probes/noderesources/resource_calculator.go#L125-L144) — `initWeight` starts at 1 and doubles each iteration. Oldest sample (index 0) = weight 1; newest sample (index 9) = weight 512.

**Defect 3:**
- [resource_calculator.go#L47-L54](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/probes/noderesources/resource_calculator.go#L47-L54) — `cfgLock sync.Mutex` declaration.
- [resource_calculator.go#L147-L153](https://github.com/volcano-sh/volcano/blob/cd4cfc9c2113920f9309668c61318260b481c287/pkg/agent/events/probes/noderesources/resource_calculator.go#L147-L153) — `getOverSubscriptionTypes` takes exclusive lock on read path.

---

## Describe the results you received and expected

**Defect 1:**
- Received: `UpdateOverSubscription` fires every 10 s when `current[res] == 0`, ignoring the 10% change guard; or the first publish is silently dropped when both `current` and `new` are zero.
- Expected: A zero-denominator guard treats the initial state as "always publish once" without dividing. Subsequent ticks use the normal 10% gate.

**Defect 2:**
- Received: Oversubscription capacity is dominated by the oldest 90-second-stale samples (weight 512). A spike that cleared 10 s ago suppresses scheduler allocation for another ~80 s.
- Expected: The most-recent sample has the highest weight (512), and the oldest has the lowest weight (1), so the agent reacts quickly to current utilisation.

**Defect 3:**
- Received: `getOverSubscriptionTypes` (read-only, called every 10 s) acquires an exclusive mutex, blocking concurrently with `RefreshCfg` writes.
- Expected: `cfgLock` is a `sync.RWMutex`; read paths use `RLock`/`RUnlock` and can run concurrently with each other.

---

## What version of Volcano are you using?

v1.15.1 (latest). Affected code introduced via commits [`5c60bf806`](https://github.com/volcano-sh/volcano/commit/5c60bf80627297e46cb51300e7cb60e89f2dee2c) and [`cd4cfc9c2`](https://github.com/volcano-sh/volcano/commit/cd4cfc9c2113920f9309668c61318260b481c287).

---

## Any other relevant information

All three defects are in the same subsystem and can be fixed together in a single PR:

1. **`policy.go`** — Add `if current[res] == 0 { update = true; break }` before the division to safely handle the initial-enable case.
2. **`resource_calculator.go`** — Iterate the queue newest-first (reverse the slice before the loop, or decrement the index) so the most-recent sample receives weight 512 and the oldest receives weight 1.
3. **`resource_calculator.go`** — Change `cfgLock sync.Mutex` to `cfgLock sync.RWMutex`; update `getOverSubscriptionTypes` to use `RLock`/`RUnlock`.
4. Add unit tests covering: zero-denominator guard, correct weight order with a full 10-entry queue, and concurrent `RefreshCfg` + `preProcess` under `-race`.

bhuvan-somisetty:
#### What type of PR is this?

/kind bug

Fixes two defects in the volcano-agent oversubscription reporting pipeline:

- **`ShouldUpdateNodeOverSubscription` divide-by-zero** (`pkg/agent/oversubscription/policy/policy.go`): the change-step comparison divided `delta` by `current[res]` without checking for a zero denominator. `current[res]` is `0` on the very first oversubscription enable cycle, or right after `DeleteNodeOverSoldStatus` clears the node annotations. `float64(delta)/float64(0)` yields `+Inf` when `delta > 0` (bypassing the 10% change guard and triggering an update — and a kube-apiserver PATCH — on every 10s tick) or `NaN` when `delta == 0` (silently dropping the very first publish). Fixed by special-casing a zero `current[res]`: update once if the new value differs, otherwise skip, without ever dividing by zero.

- **`cfgLock` was an exclusive `sync.Mutex`** (`pkg/agent/events/probes/noderesources/resource_calculator.go`): `getOverSubscriptionTypes` is a read-only lookup invoked every 10s by `preProcess`, but it took an exclusive lock, serializing it against `RefreshCfg` writes (and against itself) for no reason. Changed `cfgLock` to `sync.RWMutex`, with `getOverSubscriptionTypes` using `RLock`/`RUnlock` so concurrent reads no longer block each other.

A third item from the linked issue (weight ordering in `computeOverSubRes`) was investigated but turned out to already be correct: `historicalUsages` is oldest-first (`queue.SqQueue.Enqueue` appends to the tail, `GetAll` returns oldest-first), so the existing forward iteration with a doubling weight already gives the newest sample the highest weight (512 for a full 10-entry window) and the oldest sample the lowest (1) — i.e. it already behaves like a recency-weighted average. No code change was needed there; a regression test was added instead to lock in that behavior and guard against reintroducing the problem the issue describes.

Fixes #5820

Test plan:
- `go test ./pkg/agent/oversubscription/... ./pkg/agent/events/probes/noderesources/... -race -v` — all tests pass (run under WSL, since this dependency tree doesn't build natively on Windows).
- New tests added:
  - `TestShouldUpdateNodeOverSubscription` — covers the zero-denominator guard (first-enable with zero delta, first-enable with non-zero delta, normal below/above-threshold cases, and a mixed zero/non-zero-resource case).
  - `Test_computeOverSubRes_weightOrder` — asserts a stale spike does not dominate the weighted result.
  - `Test_historicalUsageCalculator_cfgLock_concurrentReads` — exercises `RefreshCfg` and `getOverSubscriptionTypes` concurrently under `-race` to guard against reintroducing exclusive locking on the read path.
- `go vet ./pkg/agent/oversubscription/... ./pkg/agent/events/probes/noderesources/...` and `gofmt -l` on the changed files — clean.

```release-note
Fix a divide-by-zero in the agent's oversubscription change-step check that could flood the API server with PATCH requests or silently drop the first oversubscription publish, and fix unnecessary lock contention between concurrent reads of oversubscription resource types.
```

### DIFF
--- a/pkg/agent/events/probes/noderesources/resource_calculator.go
+++ b/pkg/agent/events/probes/noderesources/resource_calculator.go
@@ -46,7 +46,7 @@ func init() {
 // historicalUsageCalculator is used to calculate the overSubscription
 type historicalUsageCalculator struct {
 	sync.Mutex
-	cfgLock sync.Mutex
+	cfgLock sync.RWMutex
 	policy.Interface
 	eventQueueFactory *framework.EventQueueFactory
 	queue             *queue.SqQueue
@@ -131,6 +131,11 @@ func (r *historicalUsageCalculator) computeOverSubRes() apis.Resource {
 	overSubRes := make(apis.Resource)
 	totalWeight := int64(0)
 	initWeight := int64(1)
+	// historicalUsages is oldest-first (queue.GetAll returns FIFO order), so
+	// iterating forward while doubling the weight each step already gives the
+	// newest sample the highest weight (512 for a full 10-entry window) and
+	// the oldest sample the lowest weight (1) -- the correct EMA-like recency
+	// weighting.
 	for _, usage := range historicalUsages {
 		totalWeight += initWeight
 		for _, res := range apis.OverSubscriptionResourceTypes {
@@ -145,8 +150,8 @@ func (r *historicalUsageCalculator) computeOverSubRes() apis.Resource {
 }
 
 func (r *historicalUsageCalculator) getOverSubscriptionTypes(node *v1.Node) map[v1.ResourceName]bool {
-	r.cfgLock.Lock()
-	defer r.cfgLock.Unlock()
+	r.cfgLock.RLock()
+	defer r.cfgLock.RUnlock()
 	ret := make(map[v1.ResourceName]bool)
 	for _, item := range r.resourceTypes.List() {
 		ret[v1.ResourceName(item)] = true
--- a/pkg/agent/events/probes/noderesources/resource_calculator_test.go
+++ b/pkg/agent/events/probes/noderesources/resource_calculator_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"sync"
 	"testing"
 
 	. "github.com/agiledragon/gomonkey/v2"
@@ -282,4 +283,53 @@ func Test_historicalUsageCalculator_RefreshCfg(t *testing.T) {
 			assert.Equal(t, tt.expectedResourceType, r.resourceTypes)
 		})
 	}
+}
+
+// Test_computeOverSubRes_weightOrder verifies that the most-recent sample in
+// the queue (the tail, since Enqueue appends and GetAll returns oldest-first)
+// receives the highest weight, so the calculator reacts quickly to current
+// utilisation instead of being dominated by stale samples.
+func Test_computeOverSubRes_weightOrder(t *testing.T) {
+	sqQueue := queue.NewSqQueue()
+	// Oldest sample first: a stale usage spike that must NOT dominate the result.
+	sqQueue.Enqueue(apis.Resource{v1.ResourceCPU: 10000, v1.ResourceMemory: 10000})
+	for i := 0; i < 8; i++ {
+		sqQueue.Enqueue(apis.Resource{v1.ResourceCPU: 0, v1.ResourceMemory: 0})
+	}
+	// Newest sample last: current low usage after the spike cleared.
+	sqQueue.Enqueue(apis.Resource{v1.ResourceCPU: 100, v1.ResourceMemory: 100})
+
+	r := &historicalUsageCalculator{queue: sqQueue}
+	res := r.computeOverSubRes()
+
+	// The newest sample carries weight 512 out of totalWeight 1023, while the
+	// stale spike (oldest) carries weight 1, so the weighted result must stay
+	// close to the newest sample and far away from the stale spike.
+	assert.Less(t, res[v1.ResourceCPU], int64(200), "stale spike must not dominate the weighted result")
+	assert.Less(t, res[v1.ResourceMemory], int64(200), "stale spike must not dominate the weighted result")
+}
+
+// Test_historicalUsageCalculator_cfgLock_concurrentReads exercises RefreshCfg
+// (writer) concurrently with getOverSubscriptionTypes (reader) to guard
+// against reintroducing an exclusive sync.Mutex on the read path. Run with
+// -race to catch data races.
+func Test_historicalUsageCalculator_cfgLock_concurrentReads(t *testing.T) {
+	r := &historicalUsageCalculator{resourceTypes: sets.NewString("cpu")}
+	node := &v1.Node{}
+
+	var wg sync.WaitGroup
+	for i := 0; i < 50; i++ {
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			_ = r.RefreshCfg(&api.ColocationConfig{
+				OverSubscriptionConfig: &api.OverSubscription{OverSubscriptionTypes: utilpointer.String("cpu,memory")},
+			})
+		}()
+		go func() {
+			defer wg.Done()
+			_ = r.getOverSubscriptionTypes(node)
+		}()
+	}
+	wg.Wait()
 }
--- a/pkg/agent/oversubscription/policy/policy.go
+++ b/pkg/agent/oversubscription/policy/policy.go
@@ -150,6 +150,14 @@ func ShouldUpdateNodeOverSubscription(current, new apis.Resource) bool {
 			delta = -delta
 		}
 
+		if current[res] == 0 {
+			if delta != 0 {
+				update = true
+				break
+			}
+			continue
+		}
+
 		if float64(delta)/float64(current[res]) > overSubscriptionChangeStep {
 			update = true
 			break
--- a/pkg/agent/oversubscription/policy/policy_test.go
+++ b/pkg/agent/oversubscription/policy/policy_test.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2024 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package policy
+
+import (
+	"testing"
+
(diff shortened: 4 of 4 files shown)


bhuvan-somisetty:
@JesseStutler could you take a look when you get a chance? Small, focused fix in the agent oversubscription path, tests included. Thanks!

PR #5826, opened 2026-08-07
scheduler: expose queue workload state metrics

### LINKED ISSUE #5825: scheduler: expose per-queue PodGroup and task state metrics at session start
/kind feature

### What is the problem you are trying to solve

Volcano exposes scheduler action/session latency metrics and queue resource metrics, but the scheduler does not expose the workload state present in each queue at the beginning of a scheduling session.

The controller-manager already exposes phase-specific PodGroup gauges such as `volcano_queue_pod_group_pending_count`, but those metrics are updated on controller reconciliation and do not include scheduler TaskStatus counts. As a result, operators cannot correlate scheduler input pressure, especially pending task backlog, with per-action and end-to-end scheduling latency.

### Describe the solution you would like

Add these scheduler gauges:

- `volcano_queue_pod_group_count{queue_name,phase}`
- `volcano_queue_task_count{queue_name,status}`

Update them once per scheduling session immediately after the session snapshot is opened and before plugins or scheduling actions mutate the session.

Proposed semantics:

- Count PodGroups and tasks by their status at the beginning of the latest scheduling session.
- Attribute each job and its tasks only to `job.Queue`; do not roll child-queue counts into parent queues.
- Publish all known PodGroup phases and task statuses for every existing queue on every session, including explicit zero values, so stale series are reset.
- Remove both metric families for a queue when that queue is deleted.
- Use bounded, lowercase phase/status label values.
- Do not add a session identifier label; Prometheus users can correlate these gauges with scheduling latency over matching time windows without introducing unbounded cardinality.

Known PodGroup phase labels would include `pending`, `inqueue`, `running`, `completed`, and `unknown`. Task status labels would include `pending`, `allocated`, `pipelined`, `binding`, `bound`, `running`, `releasing`, `succeeded`, `failed`, and `unknown`.

### Additional context

This is related to #2493 and #3293, but differs in both source and timing: those issues concern controller/plugin PodGroup gauges, while this proposal describes the scheduler session input snapshot and also covers task statuses.

### Documentation Updates

- [x] This feature requires metrics documentation changes.
- [ ] This feature requires separate Volcano website documentation changes.

ruanwenjun:
#### What type of PR is this?

/kind feature

This PR adds a scheduler gauge that describes the task state in each queue at the beginning of the latest scheduling session:

- `volcano_queue_task_count{queue_name,status}`

The metric is collected while the session snapshot is opened and before plugins or scheduling actions mutate it. Counts use direct `job.Queue` membership without hierarchical rollup, and every known task status is explicitly updated on each session, including zero values. Unexpected task statuses are normalized to `unknown`, and queue deletion removes the metric through the existing scheduler metrics cleanup path.

Task aggregation is folded into the existing `openSession` job traversal, so it does not add another full job iteration. Existing controller-manager gauges remain the source for per-queue PodGroup phase counts, avoiding duplicate scheduler metrics.

This allows operators to correlate pending task backlog with action and end-to-end scheduling latency.

Validation:

- `go test ./pkg/scheduler/... -count=1`
- `go vet ./pkg/scheduler/framework ./pkg/scheduler/metrics`
- `make verify`
- `golangci-lint run ./pkg/scheduler/framework ./pkg/scheduler/metrics`

Fixes #5825

This PR was prepared with AI assistance.

```release-note
Add volcano_queue_task_count scheduler metrics for per-queue task state at the beginning of each scheduling session.
```

### DIFF
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -61,7 +61,7 @@ This metrics describe internal state of volcano.
 | `queue_pod_group_running_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Running PodGroups in this queue |
 | `queue_pod_group_unknown_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Unknown PodGroups in this queue |
 | `queue_pod_group_count`                | Gauge           | `queue_name`=&lt;queue_name&gt;, `phase`=&lt;phase&gt;             | Number of directly assigned PodGroups by phase at the beginning of the latest scheduling session |
-| `queue_task_count`                     | Gauge           | `queue_name`=&lt;queue_name&gt;, `status`=&lt;status&gt;           | Number of tasks in directly assigned PodGroups by status at the beginning of the latest scheduling session |
+| `queue_task_count`                     | Gauge           | `queue_name`=&lt;queue_name&gt;, `status`=&lt;status&gt;           | Number of tasks by status in jobs directly assigned to this queue at the beginning of the latest scheduling session |
 | `namespace_share`                      | Gauge           | `namespace_name`=&lt;namespace_name&gt;                           | Deserved CPU count for one namespace          |
 | `namespace_weight`                     | Gauge           | `namespace_name`=&lt;namespace_name&gt;                           | Weight for one namespace                      |
 | `job_share`                            | Gauge           | `job_id`=&lt;job_id&gt;, `job_ns`=&lt;job_ns&gt;                  | Share for one job                             |
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -60,6 +60,7 @@ This metrics describe internal state of volcano.
 | `queue_pod_group_pending_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Pending PodGroups in this queue |
 | `queue_pod_group_running_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Running PodGroups in this queue |
 | `queue_pod_group_unknown_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Unknown PodGroups in this queue |
+| `queue_session_start_task_count`       | Gauge           | `queue_name`=&lt;queue_name&gt;, `status`=&lt;status&gt;           | Number of tasks by status in jobs directly assigned to this queue at the beginning of the latest scheduling session |
 | `namespace_share`                      | Gauge           | `namespace_name`=&lt;namespace_name&gt;                           | Deserved CPU count for one namespace          |
 | `namespace_weight`                     | Gauge           | `namespace_name`=&lt;namespace_name&gt;                           | Weight for one namespace                      |
 | `job_share`                            | Gauge           | `job_id`=&lt;job_id&gt;, `job_ns`=&lt;job_ns&gt;                  | Share for one job                             |
--- a/docs/design/metrics.md
+++ b/docs/design/metrics.md
@@ -60,6 +60,8 @@ This metrics describe internal state of volcano.
 | `queue_pod_group_pending_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Pending PodGroups in this queue |
 | `queue_pod_group_running_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Running PodGroups in this queue |
 | `queue_pod_group_unknown_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Unknown PodGroups in this queue |
+| `queue_pod_group_count`                | Gauge           | `queue_name`=&lt;queue_name&gt;, `phase`=&lt;phase&gt;             | Number of directly assigned PodGroups by phase at the beginning of the latest scheduling session |
+| `queue_task_count`                     | Gauge           | `queue_name`=&lt;queue_name&gt;, `status`=&lt;status&gt;           | Number of tasks in directly assigned PodGroups by status at the beginning of the latest scheduling session |
 | `namespace_share`                      | Gauge           | `namespace_name`=&lt;namespace_name&gt;                           | Deserved CPU count for one namespace          |
 | `namespace_weight`                     | Gauge           | `namespace_name`=&lt;namespace_name&gt;                           | Weight for one namespace                      |
 | `job_share`                            | Gauge           | `job_id`=&lt;job_id&gt;, `job_ns`=&lt;job_ns&gt;                  | Share for one job                             |
--- a/pkg/scheduler/framework/session.go
+++ b/pkg/scheduler/framework/session.go
@@ -51,6 +51,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/cache"
 	"volcano.sh/volcano/pkg/scheduler/conf"
 	"volcano.sh/volcano/pkg/scheduler/gate"
+	"volcano.sh/volcano/pkg/scheduler/metrics"
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/scheduler/framework/session.go
```diff
@@ -283,6 +312,36 @@ func openSession(cache cache.Cache) *Session {
...
+		if podGroupCounts[job.Queue] == nil {
+			podGroupCounts[job.Queue] = make(map[scheduling.PodGroupPhase]int)
+		}
+		if taskCounts[job.Queue] == nil {
+			taskCounts[job.Queue] = make(map[api.TaskStatus]int)
+		}
+
+		if job.PodGroup != nil {
+			podGroupCounts[job.Queue][job.PodGroup.Status.Phase]++
+		}
```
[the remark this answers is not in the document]
ruanwenjun: Fixed in a0232192a. Unexpected or empty PodGroup phases are now normalized to PodGroupUnknown before counting, and TestUpdateQueueStateMetrics covers an empty phase.

docs/design/metrics.md
```diff
@@ -60,6 +60,8 @@ This metrics describe internal state of volcano.
 | `queue_pod_group_pending_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Pending PodGroups in this queue |
 | `queue_pod_group_running_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Running PodGroups in this queue |
 | `queue_pod_group_unknown_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Unknown PodGroups in this queue |
+| `queue_pod_group_count`                | Gauge           | `queue_name`=&lt;queue_name&gt;, `phase`=&lt;phase&gt;             | Number of directly assigned PodGroups by phase at the beginning of the latest scheduling session |
+| `queue_task_count`                     | Gauge           | `queue_name`=&lt;queue_name&gt;, `status`=&lt;status&gt;           | Number of tasks in directly assigned PodGroups by status at the beginning of the latest scheduling session |
```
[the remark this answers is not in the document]
ruanwenjun: Fixed in a0232192a. The metric Help text and documentation now describe tasks by status in jobs directly assigned to the queue, matching the job.Queue attribution.

ruanwenjun:
@hwdef @JesseStutler PTAL

docs/design/metrics.md
```diff
@@ -60,6 +60,8 @@ This metrics describe internal state of volcano.
 | `queue_pod_group_pending_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Pending PodGroups in this queue |
 | `queue_pod_group_running_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Running PodGroups in this queue |
 | `queue_pod_group_unknown_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Unknown PodGroups in this queue |
+| `queue_pod_group_count`                | Gauge           | `queue_name`=&lt;queue_name&gt;, `phase`=&lt;phase&gt;             | Number of directly assigned PodGroups by phase at the beginning of the latest scheduling session |
```
JesseStutler: Could current 
```
| `queue_pod_group_inqueue_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Inqueue PodGroups in this queue |
| `queue_pod_group_pending_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Pending PodGroups in this queue |
| `queue_pod_group_running_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Running PodGroups in this queue |
| `queue_pod_group_unknown_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Unknown PodGroups in this queue |
```
meet your requirement?

pkg/scheduler/framework/session.go
```diff
@@ -62,6 +63,34 @@ const (
 	ClusterTopHyperNode = "<cluster-top-hypernode>"
 )
 
+var (
```
JesseStutler: Should not put these global var in the session.go, should put them inside like metrics, besides, Is it necessary to define such a table-driven style map? Can we directly use `toLower` to convert the phase to the corresponding string?

pkg/scheduler/framework/session.go
```diff
@@ -283,6 +312,43 @@ func openSession(cache cache.Cache) *Session {
...
+			phase := job.PodGroup.Status.Phase
+			switch phase {
+			case scheduling.PodGroupPending, scheduling.PodGroupInqueue, scheduling.PodGroupRunning,
+				scheduling.PodGroupCompleted, scheduling.PodGroupUnknown:
+			default:
+				phase = scheduling.PodGroupUnknown
+			}
+			podGroupCounts[job.Queue][phase]++
+		}
+		for status, tasks := range job.TaskStatusIndex {
```
JesseStutler: I'm worried about adding a new iteration for only update metrics will cause performance downgrade, could you clarify the necessity of these metrics? Can they be replaced with other metrics or combinations thereof? When adding metrics, we need to consider their impact on performance, and they must be common, not customized. If replacement is not possible, can we record the data in existing traversals elsewhere, instead of during `openSession`(like in the closeSession there are already some iteration and will update the metrics)?

docs/design/metrics.md
ruanwenjun: Yes, the podgroup count  already exist at controller manager, it don't need to add at scheduler again.

ruanwenjun: The scheduler-side `volcano_queue_pod_group_count` metric, documentation, cleanup, and tests have been removed. The existing controller-manager phase-specific gauges remain the source for per-queue PodGroup counts.

pkg/scheduler/framework/session.go
ruanwenjun: Yes, the PodGroup phase table was removed together with the duplicate metric. Task aggregation now derives labels with `strings.ToLower(status.String())`; the bounded task-status label list lives in the metrics package only so absent statuses can still be published explicitly as zero. Unexpected or combined statuses are normalized to `unknown` by `TaskStatus.String()`.

pkg/scheduler/framework/session.go
ruanwenjun: Goot catch, I have moved the code under existing traversals.

The task status metrics for each queue are very meaningful to us; we can monitor the number of pending tasks in each queue.

hwdef:
/lgtm
Thanks!

pkg/scheduler/framework/session.go
```diff
@@ -228,13 +229,23 @@ func openSession(cache cache.Cache) *Session {
...
 	ssn.Jobs = snapshot.Jobs
 	for _, job := range ssn.Jobs {
 		if job.PodGroup != nil {
 			ssn.PodGroupOldState.Status[job.UID] = *job.PodGroup.Status.DeepCopy()
 			ssn.PodGroupOldState.Annotations[job.UID] = maps.Clone(job.PodGroup.GetAnnotations())
 		}
+		if taskCountsByQueue[job.Queue] == nil {
+			taskCountsByQueue[job.Queue] = make(map[string]int)
+		}
+		for status, tasks := range job.TaskStatusIndex {
```
JesseStutler, line 243: I'm wondering why these metric is recorded in openSession rather than in closeSession? Some status in previous session will not carry into this session

JesseStutler:
Also, I suggest not hold commit like `address xxx review feedback`, please squash and clean your commits into 1~2 once you resolve all the comments

pkg/scheduler/framework/session.go
ruanwenjun: I'm more interested in tracking changes to pending tasks.; if I put it in `closeSession`, I might miss some pending tasks.

ruanwenjun:
> Also, I suggest not hold commit like `address xxx review feedback`, please squash and clean your commits into 1~2 once you resolve all the comments

Thanks for your suggestion, I squash the commit into one.

JesseStutler:
@ruanwenjun Looks like after your squash, the codes revert back to previous code

pkg/scheduler/framework/session.go
JesseStutler: So if you only care about pending tasks, why not tracked in controller? Or you only care about when each session start, how many tasks are currently pending?

ruanwenjun:
> @ruanwenjun Looks like after your squash, the codes revert back to previous code

resolved

pkg/scheduler/framework/session.go
ruanwenjun: Yes, the best approach would be to track this metric in the controller. However, the controller currently only observes PodGroups, and PodGroups do not record the number of pending tasks. Therefore, the controller has no way to determine how many tasks are pending in each queue.

https://github.com/volcano-sh/volcano/blob/5a15213d8ebbf10bf2fdfd9c62af669d679d7446/staging/src/volcano.sh/apis/pkg/apis/scheduling/types.go#L279-L298

docs/design/metrics.md
```diff
@@ -60,6 +60,7 @@ This metrics describe internal state of volcano.
 | `queue_pod_group_pending_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Pending PodGroups in this queue |
 | `queue_pod_group_running_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Running PodGroups in this queue |
 | `queue_pod_group_unknown_count`        | Gauge           | `queue_name`=&lt;queue_name&gt;                                   | The number of Unknown PodGroups in this queue |
+| `queue_task_count`                     | Gauge           | `queue_name`=&lt;queue_name&gt;, `status`=&lt;status&gt;           | Number of tasks by status in jobs directly assigned to this queue at the beginning of the latest scheduling session |
```
JesseStutler: If really need it, I would prefer more precise name such as `queue_session_start_task_count`

ruanwenjun: Agree with you, done.

JesseStutler:
/approve
Wait CI get done

PR #5845, opened 2026-08-10
fix(scheduler): avoid int64 overflow for large scalar resources

### LINKED ISSUE #5844: Large ephemeral-storage or extended-resource capability on a Queue admits nothing
### Description

Setting a very large `ephemeral-storage` or extended-resource value in a Queue's `spec.capability`, which is a common way to express "effectively unlimited", makes that dimension of the queue admit nothing at all. A 1Gi request stops fitting a 100P capability.

The cause is that `NewResource` converts those dimensions with `MilliValue()`, which multiplies by 1000 and leaves `int64` at a raw value of roughly 9.2e15. `MilliCPU` and `Memory` are protected against the resulting non-positive figure a few lines later in the capacity plugin, but scalar resources are not, so the capability for that dimension stays at 0 and every request is compared against 0.

Found through source review and reproduced against the scheduler's own resource API. I have not reproduced it on a live cluster, so the cluster-side symptom is inferred from the code path below rather than observed.

1. Create a Queue whose `spec.capability` sets `ephemeral-storage` to a value at or above 10P, for example to mean "no practical limit".
2. Submit a job to that queue whose pods request a small amount of `ephemeral-storage`, for example 1Gi.
3. Observe the queue behaving as if it has no `ephemeral-storage` capacity at all.

The same applies to any extended resource, since those take the same `MilliValue()` path.

Path from the user action to the failure, at commit df3f9840:

1. The Queue's `spec.capability` is converted by the capacity plugin:
   https://github.com/volcano-sh/volcano/blob/df3f9840/pkg/scheduler/plugins/capacity/capacity.go#L1083
   The proportion plugin does the same:
   https://github.com/volcano-sh/volcano/blob/df3f9840/pkg/scheduler/plugins/proportion/proportion.go#L121

2. `NewResource` converts `ephemeral-storage` and every scalar resource with `MilliValue()`:
   https://github.com/volcano-sh/volcano/blob/df3f9840/pkg/scheduler/api/resource_info.go#L99
   https://github.com/volcano-sh/volcano/blob/df3f9840/pkg/scheduler/api/resource_info.go#L115
   `Memory` uses `Value()` and is far less exposed:
   https://github.com/volcano-sh/volcano/blob/df3f9840/pkg/scheduler/api/resource_info.go#L94

3. `MilliValue()` is the quantity multiplied by 1000, so it leaves `int64` once the raw value passes about 9.2e15.

4. The capacity plugin then repairs a non-positive `MilliCPU` and `Memory` by treating them as unlimited, but does nothing for scalar resources:
   https://github.com/volcano-sh/volcano/blob/df3f9840/pkg/scheduler/plugins/capacity/capacity.go#L1084-L1088
   So `MilliCPU` recovers by accident and `ephemeral-storage` is left at 0.

5. That 0 is then what requests are compared against.

Running this against the scheduler's own API shows the boundary. It calls `NewResource` and `LessEqual`, which is why it is offered as a demonstration of the conversion rather than as proof of the cluster symptom on its own:

```go
for _, s := range []string{"1P", "9P", "10P", "100P"} {
    cap := NewResource(v1.ResourceList{v1.ResourceEphemeralStorage: resource.MustParse(s)})
    req := NewResource(v1.ResourceList{v1.ResourceEphemeralStorage: resource.MustParse("1Gi")})
    t.Logf("capability=%-5s stored=%-24.0f 1Gi fits? %v", s,
        cap.ScalarResources[v1.ResourceEphemeralStorage], req.LessEqual(cap, Zero))
}
```

```
capability=1P    stored=1000000000000000000   1Gi fits? true
capability=9P    stored=9000000000000000000   1Gi fits? true
capability=10P   stored=0                     1Gi fits? false
capability=100P  stored=0                     1Gi fits? false
```

Separately, `Resource` holds every dimension as `float64`, so above 2^53 consecutive integers stop being distinguishable. `float64(Value())` maps 9007199254740992 and 9007199254740993 onto the same number, which makes a `used < requested` comparison answer false where it should answer true. That one needs quantities far larger than the overflow above, so it is a smaller concern in practice.

Received: with `capability` at 10P or more for `ephemeral-storage`, the stored capability for that dimension is 0 and a 1Gi request does not fit.

Expected: a larger capability never admits less than a smaller one. Either the conversion keeps the value, or the value is rejected at validation, or it is treated as unlimited in the same way `MilliCPU` and `Memory` already are.

master at df3f9840

This fails closed rather than open, so it turns away work that should be admitted rather than admitting work it should not. I would rather not guess at the fix: clamping the scalar path the way `MilliCPU` and `Memory` are clamped would hide the overflow rather than remove it, and moving `Resource` to fixed point is a much larger change. Happy to work on whichever direction you prefer.

khanskr:
#### What type of PR is this?

/kind bug

`NewResource` converts scalar resources (`ephemeral-storage` and extended resources) using `Quantity.MilliValue()`, which overflows `int64` for very large quantities such as `10P`. The overflowed value is stored as 0, so a Queue whose `spec.capability` sets such a large value on that dimension admits nothing, not even a small 1Gi request. The conversion now falls back to a float64 path for quantities too large for int64, preserving the value.

For quantities small enough to fit in int64 the behavior is unchanged.

Fixes #5844

The PR description was generated with the help of an AI tool.

NONE

### DIFF
--- a/pkg/scheduler/api/resource_info.go
+++ b/pkg/scheduler/api/resource_info.go
@@ -83,6 +83,20 @@ func InfiniteResource() *Resource {
 	}
 }
 
+// quantityToMilliFloat64 converts a resource.Quantity to milli-units as float64.
+// Quantity.MilliValue() overflows int64 for very large quantities (e.g. "10P"
+// ephemeral-storage), which would otherwise be stored as a broken, non-positive
+// value and make the resource appear unavailable. This conversion falls back to
+// a float64 path for such quantities so the value is preserved.
+func quantityToMilliFloat64(q resource.Quantity) float64 {
+	// Fast path: exact int64 milli value when it cannot overflow.
+	if v, ok := q.AsInt64(); ok && v <= math.MaxInt64/1000 && v >= math.MinInt64/1000 {
+		return float64(q.MilliValue())
+	}
+	// Fall back to a float64 representation for values too large for int64.
+	return q.AsApproximateFloat64() * 1000
+}
+
 // NewResource creates a new resource object from resource list
 func NewResource(rl v1.ResourceList) *Resource {
 	r := EmptyResource()
@@ -96,7 +110,7 @@ func NewResource(rl v1.ResourceList) *Resource {
 			r.MaxTaskNum += int(rQuant.Value())
 			r.AddScalar(rName, float64(rQuant.Value()))
 		case v1.ResourceEphemeralStorage:
-			r.AddScalar(rName, float64(rQuant.MilliValue()))
+			r.AddScalar(rName, quantityToMilliFloat64(rQuant))
 		default:
 			if IsCountQuota(rName) {
 				continue
@@ -112,7 +126,7 @@ func NewResource(rl v1.ResourceList) *Resource {
 					return true
 				})
 				if !ignore {
-					r.AddScalar(rName, float64(rQuant.MilliValue()))
+					r.AddScalar(rName, quantityToMilliFloat64(rQuant))
 				} else {
 					klog.V(4).Infof("Ignoring resource %s", rName.String())
 				}
--- a/pkg/scheduler/api/resource_info_test.go
+++ b/pkg/scheduler/api/resource_info_test.go
@@ -66,6 +66,33 @@ func TestNewResource(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


miantalha45:
@khanskr please wait for the maintainer direction as mentioned in issue #5844 before submitting this PR.

PR #5851, opened 2026-08-11
capacity: fix ancestor queue losing track of shared ResourceClaims

### LINKED ISSUE #5847: capacity: ancestor queues lose track of shared ResourceClaims, in both directions
### Description

A ResourceClaim shared by several tasks is charged once to a queue, and `queueAttr.resourceClaimRefs` is what keeps that from being charged again. Ancestor queues never get those references, so their DRA accounting drifts in both directions as soon as one claim is shared.

The two paths that reach an ancestor disagree. `buildHierarchicalQueueAttrs` charges the leaf through the refcounted helper and the ancestor through a plain delta (`pkg/scheduler/plugins/capacity/capacity.go`, leaf at L1261, ancestor at L1316):

```go
if cp.dynamicResourceAllocationEnable && attr.dra != nil && t.DRAResreq != nil {
    addTaskDRAAllocated(attr, t)          // maintains attr.resourceClaimRefs
}
...
updateDRAAllocated(ancestorAttr.dra, filteredDelta)   // no references recorded
```

`AllocateFn` and `DeallocateFn` then treat the ancestor as if it had them, at L853 and L883, and so do the event handlers at L951 and L989.

Since the ancestor's map is empty, `removeTaskDRAAllocated` reads a zero for every claim key and takes the last-reference branch (L403):

```go
refCount := attr.resourceClaimRefs[claimKey]
if refCount <= 1 {
    subtractDRAAllocated(attr.dra, claimReq)
```

so every release subtracts, and `addTaskDRAAllocated` charges again for a claim the ancestor is already carrying.

Against `df3f9840`, drop this in `pkg/scheduler/plugins/capacity/` and run `go test ./pkg/scheduler/plugins/capacity/`:

```go
package capacity

import (
	"testing"

)

func TestAncestorSharedClaim(t *testing.T) {
	sharedTask := func(name string) *api.TaskInfo {
		return &api.TaskInfo{
			Name:              name,
			Resreq:            api.EmptyResource(),
			ResourceClaimKeys: []string{"ns1/shared"},
			ResourceClaimDRAResreq: map[string]map[string]*api.DRAResource{
				"ns1/shared": {"gpu.com": {Count: 1}},
			},
			DRAResreq: map[string]*api.DRAResource{"gpu.com": {Count: 1}},
		}
	}
	newAttr := func(charged int64) *queueAttr {
		return &queueAttr{
			allocated:      api.EmptyResource(),
			realCapability: api.InfiniteResource(),
			dra: &draQuotaAttr{
				allocated:  map[string]*api.DRAResource{"gpu.com": {Count: charged}},
				capability: map[string]*api.DRAResource{"gpu.com": {Count: 4}},
			},
			resourceClaimRefs: make(map[string]int),
		}
	}

	taskA, taskB := sharedTask("a"), sharedTask("b")

	// Session start: the leaf holds two references to the one claim and
	// charges it once, the ancestor is charged the same once by delta.
	leaf := newAttr(0)
	addTaskDRAAllocated(leaf, taskA)
	addTaskDRAAllocated(leaf, taskB)
	ancestor := newAttr(leaf.dra.allocated["gpu.com"].Count)

	// One task is released. The claim is still held by the other.
	removeTaskDRAAllocated(leaf, taskA)
	removeTaskDRAAllocated(ancestor, taskA)

	leafCount := leaf.dra.allocated["gpu.com"].Count
	ancestorCount := ancestor.dra.allocated["gpu.com"].Count
	t.Logf("after releasing one of two tasks: leaf %d, ancestor %d", leafCount, ancestorCount)
	if ancestorCount < leafCount {
		t.Errorf("ancestor charges %d for a claim its child still charges %d for", ancestorCount, leafCount)
	}

	// The same gap the other way round: a task admitted during the session on
	// a claim the ancestor was already charged for at build.
	admitted := newAttr(1)
	addTaskDRAAllocated(admitted, taskB)
	if got := admitted.dra.allocated["gpu.com"].Count; got != 1 {
		t.Errorf("ancestor charges %d for one shared claim", got)
	}
}
```

```
--- FAIL: TestAncestorSharedClaim (0.00s)
    after releasing one of two tasks: leaf 1, ancestor 0
    ancestor charges 0 for a claim its child still charges 1 for
    ancestor charges 2 for one shared claim
```

Both halves need only a hierarchical queue, `dynamicResourceAllocationEnable`, and one ResourceClaim referenced by two tasks in the same leaf queue.

The release half needs one of those tasks to be evicted or to have its allocation rolled back during a session, which is ordinary preempt and reclaim traffic. The leaf keeps charging the claim because its second reference is still there, the ancestor drops to zero, and the parent quota then reads as free for a device that is still in use. That is the over-admission direction.

The admit half needs the second task to be scheduled after the session was built rather than before it, so the build charges the claim once by delta and `AllocateFn` charges it again. The parent then reads as fuller than it is and rejects work it has room for.

Neither shows up as a negative number, because `DRAResource.Sub` clamps at zero (`pkg/scheduler/api/job_info.go:1500`). The ancestor just drifts, and it stays drifted until the next `OpenSession` rebuilds the attributes.

I have not driven this from a live cluster. The reproduction above is at the level of the accounting helpers, using the state the build path leaves on an ancestor.

An ancestor should charge a shared claim for as long as any descendant charges it, and should charge it once. What happens instead is that the ancestor releases the claim on the first task that goes away, and charges it a second time for a task admitted later in the session.

main at `df3f9840`, reproduced with go 1.26.5.

The narrow fix is to seed the ancestor's references in the build path, so the two paths agree on what a reference means. The wider one is to stop keeping a separate refcount per ancestor at all and derive ancestor totals from their children, since the delta in the build path is already doing exactly that. I would rather not guess which you prefer, since the second changes where the accounting lives.

I am happy to put together a patch with tests once you have said which direction you want.

The refcount arrived in `5ba79f31` and the ancestor delta in `6275b05c`, so the two have been out of step since DRA queue accounting landed.

Anirudh0465:
#### What type of PR is this?
/kind bug

Previously, ancestor queues were charged using `updateDRAAllocated` during the build phase. This skipped updating `resourceClaimRefs` and left it empty. Because of this, the ancestor queue would incorrectly drop its charge on the first release and then double-charge when admitting the same shared claim.

This PR fixes that by using `addTaskDRAAllocated` directly for ancestor queues in the build path. This correctly seeds the per-task refcounts and keeps the build logic perfectly consistent with the live-update path.

**Test Command:**
`go test ./pkg/scheduler/plugins/capacity/ -v -run TestAncestorSharedClaim -count=1`

**Results:**
=== RUN   TestAncestorSharedClaim
I0812 02:34:16.065976  121163 feature_gate.go:362] "Warning: setting GA feature gate. It will be removed in a future release." featureGate="DynamicResourceAllocation" value=true
PASS: TestAncestorSharedClaim (0.00s)
PASS
ok      volcano.sh/volcano/pkg/scheduler/plugins/capacity       0.019s

Fixes #5847

Following up on the review feedback:
* Removed the unreachable `ancestors` propagation block from the non-hierarchical `buildQueueAttrs`.
* Dropped the custom filtered helper to keep the build and live-update paths fully consistent.
* Rewrote the regression test to construct a real hierarchical session with two leaf queues under one root. It now explicitly verifies all 5 requested invariants on the caller-visible state.

```release-note
Fix shared ResourceClaim refcount drift on ancestor queues during capacity allocation

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -409,61 +409,6 @@ func removeTaskDRAAllocated(attr *queueAttr, task *api.TaskInfo) {
 	}
 }
 
-// addTaskDRAAllocatedForAncestor is like addTaskDRAAllocated but only charges
-// DeviceClasses that are present in the ancestor's capability map. It must be
-// used instead of updateDRAAllocated in the build path so that ancestor queues
-// receive correct resourceClaimRefs and agree with the refcount-aware helpers
-// called by AllocateFn/DeallocateFn and the event handlers.
-//
-// Without this, ancestor.resourceClaimRefs stays empty: removeTaskDRAAllocated
-// always reads refCount=0 and subtracts (over-admission), and
-// addTaskDRAAllocated always reads refCount=0 and charges again (under-admission).
-// See https://github.com/volcano-sh/volcano/issues/5847.
-func addTaskDRAAllocatedForAncestor(attr *queueAttr, task *api.TaskInfo) {
-	if attr == nil || attr.dra == nil || task == nil {
-		return
-	}
-	capability := attr.dra.capability
-	if len(task.ResourceClaimDRAResreq) == 0 {
-		// Non-shared claims: filter task.DRAResreq by ancestor capability.
-		if task.DRAResreq != nil {
-			filtered := make(map[string]*api.DRAResource)
-			for dc, res := range task.DRAResreq {
-				if _, ok := capability[dc]; ok {
-					filtered[dc] = res
-				}
-			}
-			if len(filtered) > 0 {
-				updateDRAAllocated(attr.dra, filtered)
-			}
-		}
-		return
-	}
-	// Shared claims: same refcount logic as addTaskDRAAllocated, but only charge
-	// DeviceClasses that exist in the ancestor's capability.
-	if attr.resourceClaimRefs == nil {
-		attr.resourceClaimRefs = make(map[string]int)
-	}
-	for _, claimKey := range task.ResourceClaimKeys {
-		claimReq := task.ResourceClaimDRAResreq[claimKey]
-		if len(claimReq) == 0 {
-			continue
-		}
-		filtered := make(map[string]*api.DRAResource)
-		for dc, res := range claimReq {
-			if _, ok := capability[dc]; ok {
-				filtered[dc] = res
-			}
-		}
-		if len(filtered) == 0 {
-			continue
-		}
-		if attr.resourceClaimRefs[claimKey] == 0 {
-			updateDRAAllocated(attr.dra, filtered)
-		}
-		attr.resourceClaimRefs[claimKey]++
-	}
-}
 
 // New return capacityPlugin action
 func New(arguments framework.Arguments) framework.Plugin {
@@ -1168,16 +1113,6 @@ func (cp *capacityPlugin) buildQueueAttrs(ssn *framework.Session) {
 					attr.request.Add(t.Resreq)
 					if cp.dynamicResourceAllocationEnable && attr.dra != nil && t.DRAResreq != nil {
 						addTaskDRAAllocated(attr, t)
-						// Propagate to ancestors using the refcount-aware helper so that
-						// ancestorAttr.resourceClaimRefs is seeded for shared claims.
-						// This fixes issue #5847 where the old delta-based update left
-						// resourceClaimRefs empty, causing accounting drift in both directions.
-						for _, ancestor := range attr.ancestors {
-							ancestorAttr := cp.queueOpts[ancestor]
-							if ancestorAttr.dra != nil {
-								addTaskDRAAllocatedForAncestor(ancestorAttr, t)
-							}
-						}
 					}
 				}
 			} else if status == api.Pending {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by thc1006 — COMMENTED** [review]
Thanks for putting this together. Seeding the ancestor refcounts per task is a plausible direction, but the current test does not exercise the bug or the new production path.

`TestAncestorSharedClaim` constructs the ancestor like this:

```go
ancestor := newAttr(0)
addTaskDRAAllocated(ancestor, taskA)
addTaskDRAAllocated(ancestor, taskB)
```

That directly creates the expected post-fix state using the existing refcount-aware helper. The test never calls `buildHierarchicalQueueAttrs` or `addTaskDRAAllocatedForAncestor`, so it would pass against the old implementation as well. It would also continue passing if the new production wiring were removed.

Could you replace this with a regression test that drives the hierarchical build path, or an equivalent test of its actual caller-visible state? At minimum, please verify:

* after building two tasks that share one claim, the leaf and every applicable ancestor have refcount 2 and allocated count 1;
* removing one task leaves refcount 1 and allocated count 1;
* removing the final task removes the reference and reduces allocated count to 0;
* admitting another task for an already-counted claim does not charge it again;
* the test fails on the base revision and passes with this change.

It would also be valuable to cover two leaf queues under the same ancestor sharing the same claim.

One more consistency question: the new build helper filters DeviceClasses using the ancestor capability, while the simulate/event allocation paths still call the unfiltered `addTaskDRAAllocated` for ancestors. Please either use consistent filtering across build and live-update paths or add a focused test demonstrating why the different states are intentional and safe.

The additional propagation block in non-hierarchical `buildQueueAttrs` also appears unreachable because that path does not populate ancestors; please remove it unless there is a case that requires it.

Finally, please restore the repository PR-template sections, include the test command/results, and format the commit message with a blank line after the subject.

thc1006:
have a nice day or night. I am about to sleep other question I'll reply when I wake up ;)

Anirudh0465:
@thc1006 thanks for the thorough review!! Your points were spot on.

I've pushed an update that addresses everything you brought up:

1. Regression Test: I completely rewrote the test to construct a real hierarchical session with two leaf queues under a single root queue. It now drives the actual buildHierarchicalQueueAttrs path and explicitly verifies all 5 invariants you listed (build, remove 1st, remove 2nd, admit 1st, admit 2nd) by checking the caller-visible state. It correctly fails on the base branch and passes here.
2. Consistency: about t he filtering discrepancy, I've removed the custom addTaskDRAAllocatedForAncestor helper entirely.        The build path now just calls the unfiltered addTaskDRAAllocated(ancestor, task), perfectly matching the live-update behavior for consistency.
3. Unreachble Code: Removed the dead ancestors propagation block in buildQueueAttrs.
4. Formatting: I've updated the pr description with the template, included the test command and results, and formatted the commit message appropriately.

Let me know if the updated test and approach look good.

PR #5854, opened 2026-08-12
scheduler/framework: restore pod node name on unallocate

### LINKED ISSUE #5853: scheduler/framework: Pod.Spec.NodeName is not restored after allocation rollback
### What happened

`Statement.Allocate()` sets `task.Pod.Spec.NodeName` to the selected
node. When the allocation is later discarded/rolled back,
`Statement.unallocate()` resets `task.NodeName` but does not reset
`task.Pod.Spec.NodeName`.

This leaves the TaskInfo in an inconsistent state after the allocation
has been rolled back.

### Actual behavior

After `Discard()`:

```text
task.NodeName          == ""
task.Pod.Spec.NodeName == "n1"

1. Create a pending TaskInfo with an empty Pod.Spec.NodeName.
2. Create a Statement for the scheduler session.
3. Call Statement.Allocate(task, node).
4. Call Statement.Discard() to roll back the allocation.
5. Inspect task.NodeName and task.Pod.Spec.NodeName.

Observed:

task.NodeName          == ""
task.Pod.Spec.NodeName == "n1"
```
The following minimal Go test reproduces the issue:
```
func TestAllocateUnallocateSpecNodeName(t *testing.T) {
	ssn, _, task, node := newTestSession(t)
	stmt := NewStatement(ssn)

	originalSpecNodeName := task.Pod.Spec.NodeName
	if err := stmt.Allocate(task, node); err != nil {
		t.Fatalf("Allocate failed: %v", err)
	}
	if task.Pod.Spec.NodeName != node.Name {
		t.Fatalf("expected Spec.NodeName %s, got %s", node.Name, task.Pod.Spec.NodeName)
	}

	stmt.Discard()

	if task.Pod.Spec.NodeName != originalSpecNodeName {
		t.Errorf("expected Spec.NodeName to be restored to %q after Discard, got %q",
			originalSpecNodeName, task.Pod.Spec.NodeName)
	}
}
```

The issue is reproducible through both the framework and a normal scheduler path.

During speculative hyperNode allocation, `allocateResourcesForTasks()` calls `Statement.Allocate()`, which sets `task.Pod.Spec.NodeName` to the selected node. The allocation is then saved and discarded via `stmt.Discard()`, which calls `unallocate()`.

`unallocate()` resets `task.NodeName` but does not reset `task.Pod.Spec.NodeName`, leaving:

task.NodeName          == ""
task.Pod.Spec.NodeName == "<selected node>"

This is reproduced by `TestAllocateUnallocateSpecNodeName`, which fails before the fix with:

expected Spec.NodeName to be restored to "" after Discard, got "n1"

The fix restores `task.Pod.Spec.NodeName` during unallocation, and the regression test passes.

This is confirmed as an internal state inconsistency. A user-visible scheduling failure has not been reproduced; `PodLister.GetPod()` currently masks the mismatch for downstream consumers.

Actual result:

After Statement.Discard() rolls back the allocation, task.NodeName is
restored to "", but task.Pod.Spec.NodeName remains set to the node used
during the discarded allocation.

For example:

task.NodeName          == ""
task.Pod.Spec.NodeName == "n1"

The reproducing test fails with:

expected Spec.NodeName to be restored to "", got "n1"

Expected result:

After the allocation is discarded, both task.NodeName and
task.Pod.Spec.NodeName should be restored to their original values.

v1.16.0-alpha.1

_No response_

Roaimkhan:
#### What type of PR is this?
Bug fix
Restore Pod.Spec.NodeName when rolling back a task allocation.

Statement.Allocate sets Pod.Spec.NodeName, but Statement.unallocate did
not restore it when an allocation was discarded. This left task.NodeName
and Pod.Spec.NodeName inconsistent after rollback.

This PR restores the Pod node name during unallocation and adds a
regression test covering the rollback behavior.

Fixes #5853

```release-note
None
```


Roaimkhan:
Hi @JesseStutler 

It's been a while since this PR was opened! Could you please take a look?

PR #5843, opened 2026-08-10
fix(deviceshare): support HAMi Ascend normal preemption

miantalha45:
## Summary
- Return `Unschedulable` only for insufficient HAMi Ascend capacity.
- Preserve `Error` for malformed inventory and other device failures.
- Recheck full predicate feasibility after each tentative victim eviction in normal preemption.
- Add HAMi error-classification and normal-preemption regression tests.

## Validation
- `go test ./pkg/scheduler/actions/preempt ./pkg/scheduler/api/devices/ascend/hami`
- `git diff --check`
- Kind validation with two physical Ascend devices split into 14 vNPUs:
  - normal preemption succeeds
  - topology-aware preemption succeeds
  - both evict one `vir04` low-priority workload and schedule the high-priority workload.
  

Fixes #5828

PR #5855, opened 2026-08-12
fix(reclaim): recheck device feasibility after tentative evictions

### LINKED ISSUE #5852: Reclaim action stops after aggregate resources fit without rechecking device predicate feasibility
**Origin:** Identified during the HAMi Ascend preemption investigation in #5828. This report covers the separate `reclaim` action (cross-queue) path. The corresponding `normalPreempt` path is addressed by #5843.

**Description**

The cross-queue `reclaim` action selects victims using only generic Node resource accounting. After each tentative eviction, it checks whether the reclaiming task’s `InitResreq` fits `NodeInfo.FutureIdle()`. It does not re-run `ssn.PredicateFn()` against the Node’s simulated state before deciding that reclaiming is complete.

This is incorrect for resources whose feasibility cannot be determined from aggregate quantities alone, including HAMi Ascend vNPU allocations. Generic accounting can report sufficient NPU/core/memory totals while DeviceShare cannot select a valid concrete device allocation due to device-level memory, geometry, or fragmentation constraints.

The action then reaches `Statement.Pipeline()`, where device allocation fails. The statement is correctly discarded, so victims are not committed, but reclaim does not continue evaluating further victims that could make the device allocation feasible. The reclaiming task remains pending.

This is the cross-queue equivalent of the normal-preemption feasibility gap addressed by #5843.

**Evidence and production path**

On current `master` (`6575302cc650137929678ae010fa129a3553b630`):

1. `reclaimForTask()` initially uses `PredicateForPreemptAction`, which intentionally retains Nodes with ordinary `Unschedulable` predicate failures as possible preemption/reclaim candidates. https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/actions/reclaim/reclaim.go#L180

2. The reclaim victim loop stops solely when the task fits the aggregate `availableResources` value. https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/actions/reclaim/reclaim.go#L220

3. `availableResources` is derived from `n.FutureIdle()` and updated with each victim’s generic `Resreq`, no predicate/device feasibility check occurs after `nodeStmt.Evict()`
https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/actions/reclaim/reclaim.go#L225

4. `Pipeline()` is attempted only after the loop stops. A device allocation failure discards the per-Node statement, but the action does not select additional victims for that Node https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/actions/reclaim/reclaim.go#L247

5. PR #5843 adds the missing aggregate-plus-predicate feasibility re-check to `normalPreempt()`. Reclaim currently lacks the equivalent behavior.

**Steps to reproduce the issue**

1. Configure `actions: "enqueue, allocate, reclaim, ..."` with reclaimable queues.
2. Run a pending task in Queue A that requests a device-backed resource.
3. Run lower-priority reclaimable tasks in Queue B on the only eligible Node.
4. Arrange device state so that:
   - generic `FutureIdle()` becomes sufficient after reclaiming one victim;
   - the device predicate remains unschedulable after that victim;
   - reclaiming another valid victim would make a concrete device allocation feasible.
5. Observe that reclaim stops after the first victim according to aggregate resources, `Pipeline()` fails device allocation, the statement is discarded, and the pending task remains unscheduled.

**Expected result**

Reclaim should continue selecting valid victims until both conditions hold:

1. Generic Node resources are sufficient.
2. `ssn.PredicateFn(task, node)` confirms that the task fits the Node’s tentative post-eviction device state.

If no valid device allocation can be formed, the complete per-Node reclaim attempt must be discarded without evicting any victim.

miantalha45:
#### What type of PR is this?
/kind bug

## What this PR does

Fixes the cross-queue `reclaim` path when aggregate Node resources appear sufficient but a device-aware predicate, such as HAMi Ascend DeviceShare, cannot form a valid concrete allocation.

Reclaim now re-evaluates task feasibility after each tentative victim eviction:

- generic `FutureIdle()` resource availability; and
- scheduler `PredicateFn()` results, including device-backed predicates.

The per-Node `Statement` remains transactional: unsuccessful reclaim attempts are discarded and do not commit victim evictions.

## Why

`reclaim` previously stopped selecting victims based only on aggregate resources. This can be incorrect for device resources where physical-device availability, memory, geometry, or fragmentation cannot be represented by scalar Node totals.

This issue was identified while investigating the HAMi Ascend report in #5828. It is the cross-queue counterpart to the normal-preemption predicate recheck in #5843.

## Tests

Added a regression test where:

1. Generic resources become sufficient after the first reclaim victim.
2. A simulated device predicate remains unschedulable.
3. Reclaim selects the second valid victim.
4. The predicate becomes feasible and the reclaimer is pipelined.

## Validation

- `go test ./pkg/scheduler/actions/reclaim`
- `git diff --check`

Fixes #5852

### DIFF
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -218,11 +218,8 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		}
 
 		victimsQueue := ssn.BuildVictimsPriorityQueue(victims, task)
-		resreq := task.InitResreq.Clone()
 		reclaimed := api.EmptyResource()
-
-		// The reclaimed resources should be added to the remaining available resources of the nodes to avoid over-reclaiming.
-		availableResources := n.FutureIdle()
+		reclaimerFits := reclaimerFitsOnNode(ssn, task, n)
 
 		// Use a per-node statement so that evictions are isolated to this node.
 		// Only merge into the caller's stmt if Pipeline succeeds; otherwise discard
@@ -230,7 +227,7 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		nodeStmt := framework.NewStatement(ssn)
 		evictionOccurred := false
 		for !victimsQueue.Empty() {
-			if resreq.LessEqual(availableResources, api.Zero) {
+			if reclaimerFits {
 				break
 			}
 			reclaimee := victimsQueue.Pop().(*api.TaskInfo)
@@ -238,13 +235,13 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 				reclaimee.Namespace, reclaimee.Name, task.Namespace, task.Name)
 			nodeStmt.Evict(reclaimee, "reclaim")
 			reclaimed.Add(reclaimee.Resreq)
-			availableResources.Add(reclaimee.Resreq)
 			evictionOccurred = true
+			reclaimerFits = reclaimerFitsOnNode(ssn, task, n)
 		}
 
-		klog.V(3).Infof("Reclaimed <%v> for task <%s/%s> requested <%v>, and Node <%s> availableResources <%v>.", reclaimed, task.Namespace, task.Name, task.InitResreq, n.Name, availableResources)
+		klog.V(3).Infof("Reclaimed <%v> for task <%s/%s> requested <%v>.", reclaimed, task.Namespace, task.Name, task.InitResreq)
 
-		if !resreq.LessEqual(availableResources, api.Zero) {
+		if !reclaimerFits {
 			nodeStmt.Discard()
 			continue
 		}
@@ -258,6 +255,12 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		stmt.Merge(nodeStmt)
 		break
 	}
+}
+
+// reclaimerFitsOnNode verifies aggregate resources and plugin predicates after tentative evictions.
+func reclaimerFitsOnNode(ssn *framework.Session, task *api.TaskInfo, node *api.NodeInfo) bool {
+	return task.InitResreq.LessEqual(node.FutureIdle(), api.Zero) &&
+		ssn.PredicateFn(task, node) == nil
 }
 
 func (ra *Action) UnInitialize() {
--- a/pkg/scheduler/actions/reclaim/reclaim.go
+++ b/pkg/scheduler/actions/reclaim/reclaim.go
@@ -218,8 +218,11 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
 		}
 
 		victimsQueue := ssn.BuildVictimsPriorityQueue(victims, task)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -44,6 +44,40 @@ func init() {
 	options.Default()
 }
 
+const deviceFitAfterReclaimPluginName = "device-fit-after-reclaim"
```
hajnalmt, line 47: I would have moved this into a separate file possibly with the test too.

hajnalmt (approved):
/area scheduling
/approve

Thanks for the PR, nice find. Looks good from my side.

pkg/scheduler/actions/reclaim/reclaim_test.go
miantalha45: I have moved these into a separate file.

pkg/scheduler/actions/reclaim/reclaim_device_feasibility_test.go
JesseStutler, line 1: There is no need for a separate unit test file @miantalha45 Just define it in the reclaim_test.go is fine, or should add the copyright header

miantalha45: Ok, then let me move these tests to reclaim_test.go file.

miantalha45: i removed the separate test-file commit and restored the regression test in `reclaim_test.go`. The PR now contains one commit. please check now

pkg/scheduler/actions/reclaim/reclaim.go
```diff
@@ -218,33 +218,30 @@ func (ra *Action) reclaimForTask(ssn *framework.Session, stmt *framework.Stateme
...
 				break
 			}
 			reclaimee := victimsQueue.Pop().(*api.TaskInfo)
 			klog.V(3).Infof("Try to reclaim Task <%s/%s> for Tasks <%s/%s>",
 				reclaimee.Namespace, reclaimee.Name, task.Namespace, task.Name)
 			nodeStmt.Evict(reclaimee, "reclaim")
 			reclaimed.Add(reclaimee.Resreq)
-			availableResources.Add(reclaimee.Resreq)
 			evictionOccurred = true
+			reclaimerFits = reclaimerFitsOnNode(ssn, task, n)
```
JesseStutler: The logic is not correct, it's not the same as previous, since the `availableResources` will add reclaimee's resource request once the reclaimee be evicted(in the statement)

miantalha45: @JesseStutler I restored the existing `availableResources` accounting and update it after every tentative eviction. the predicate recheck now runs alongside this preserved resource check.

JesseStutler:
/lgtm
Thanks！

miantalha45:
@JesseStutler Working on `release 1.14` conflict fix.

PR #5859, opened 2026-08-13
test/e2e: exclude terminating pods from WaitTaskPhase readiness count

### LINKED ISSUE #5856: [E2E Test] Pod not rescheduled to same hypernode after being killed in network topology test
## Issue Summary
The E2E test "Case 2.3: pod of job will be rescheduled to same hypernode when be killed" is failing in the hypernode network topology tests.

## Test Details
- **Test File:** test/e2e/hypernode/network_topology_test.go
- **Failing Line:** Line 480
- **Test Case:** Case 2.3: pod of job will be rescheduled to same hypernode when be killed

## Failure Details
The test fails during the verification step that checks if a pod is scheduled to the same hypernode after being killed and rescheduled. The expected behavior is that when a pod of a job is killed, it should be rescheduled to the same hypernode, but this is not happening.

## Steps to Reproduce
1. Run the hypernode E2E test suite: `E2E_TYPE=HYPERNODE ./hack/run-e2e-kind.sh`
2. The test fails at the verification step after pod recreation

## Expected Behavior
When a pod is killed and recreated, it should be scheduled back to the same hypernode it was previously running on.

## Actual Behavior
The pod is being scheduled to a different hypernode instead of the same one.

## Impact
This indicates a potential issue with the hypernode affinity or scheduling logic in the network topology implementation.

vanshika2720:
#### What type of PR is this?

/kind bug

Fixes intermittent failures in the network topology E2E tests for hypernode-reschedule scenarios.

The scheduler correctly rebinds the replacement task to the same hyperNode after the original Pod is killed. The flakiness was caused by the E2E `WaitTaskPhase` helper, which considered any Pod matching the requested phase.

A killed Pod can temporarily retain its previous phase while its `DeletionTimestamp` is set but before it is fully removed. `WaitTaskPhase` could therefore return successfully based on the stale Pod instead of waiting for the newly created replacement Pod to reach the expected phase.

This change makes the wait logic ignore Pods that are being deleted, ensuring that the phase check is performed against an active Pod. This prevents the following assertions from observing unsettled replacement state.

Fixes #5856

The scheduler behaviour was verified independently: the replacement task is correctly rebound to the same hyperNode.

The failure was isolated to the E2E polling helper and its handling of Pods with a `DeletionTimestamp`.

The fix is scoped to E2E test synchronization and does not change scheduler or production scheduling behaviour.

NONE

```release-note
```

### DIFF
--- a/test/e2e/util/job.go
+++ b/test/e2e/util/job.go
@@ -309,6 +309,15 @@ func WaitTaskPhase(ctx *TestContext, job *batchv1alpha1.Job, phase []v1.PodPhase
 			if !metav1.IsControlledBy(&pod, job) {
 				continue
 			}
+			// Terminating pods (e.g. the old pod being replaced after a kill)
+			// can still report a stale matching phase; excluding them avoids
+			// counting them toward readyTaskNum before their replacement has
+			// actually reached the target phase.
+			if pod.DeletionTimestamp != nil {
+				podKey := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+				podNotReadyCache[podKey] = &pod
+				continue
+			}
 
 			podKey := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
 			podReady := false


vanshika2720:
@LiZhenCheng9527 @YaoZengzeng @kube-gopher

PR #5884, opened 2026-08-21
fix: set Pod.Spec.NodeName when pipelining tasks for DRA support

zhengxingmao:
Statement.Pipeline assigns a task to a node in the preemption/reclaim path but never sets task.Pod.Spec.NodeName. With Dynamic Resource Allocation (DRA) enabled, the kube-scheduler dynamicresources Reserve plugin uses pod.Spec.NodeName as the lookup key into the per-node allocation state written by Filter (state.nodeAllocations). When Pipeline leaves it empty, Reserve fails with "claim allocation not found for node" and jobs that consume ResourceClaims get stuck in Pending.

Statement.Allocate already sets task.Pod.Spec.NodeName; Pipeline now mirrors it so both scheduling paths behave identically.

Adds unit tests asserting that Pipeline sets Pod.Spec.NodeName (and that Discard keeps task state consistent), plus an assertion in the existing successful-pipeline test.

Fixes #

```release-note

```

### DIFF
--- a/pkg/scheduler/framework/statement.go
+++ b/pkg/scheduler/framework/statement.go
@@ -170,6 +170,7 @@ func (s *Statement) Pipeline(task *api.TaskInfo, hostname string, evictionOccurr
 	}
 
 	task.NodeName = hostname
+	task.Pod.Spec.NodeName = hostname
 	task.EvictionOccurred = evictionOccurred
 
 	if node, found := s.ssn.Nodes[hostname]; found {
--- a/pkg/scheduler/framework/statement_test.go
+++ b/pkg/scheduler/framework/statement_test.go
@@ -256,6 +256,13 @@ func TestPipelineAutoRollbackOnError(t *testing.T) {
 		if task.NodeName != node.Name {
 			t.Errorf("expected task on node %s, got %s", node.Name, task.NodeName)
 		}
+		// Pipeline must set the pod's NodeName, just like Allocate does.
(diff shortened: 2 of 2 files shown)


JesseStutler:
Thanks, `the kube-scheduler dynamicresources Reserve plugin uses pod.Spec.NodeName`, Could you please provide the code reference? Ideally, it would be best if there were an existing issue to track this, as the current information is too limited, provide more like why Reserve plugin needs this would be better

JesseStutler:
Hi, any feedback? @zhengxingmao 

> Thanks, `the kube-scheduler dynamicresources Reserve plugin uses pod.Spec.NodeName`, Could you please provide the code reference? Ideally, it would be best if there were an existing issue to track this, as the current information is too limited, provide more like why Reserve plugin needs this would be better

zhengxingmao:
https://chat.deepseek.com/share/vy3ls4pxhx9ypht2sx

zhengxingmao:
先参考下上面的链接，稍后会补充。

zhengxingmao:
先参考下上面的链接，稍后会补充。

JesseStutler:
> https://chat.deepseek.com/share/vy3ls4pxhx9ypht2sx

I got it, but I have to say that linking to a AI chat is not a good behavior :) it runs counter to the collaborative spirit of the open-source community and fails to be convincing to others.

Issue #5886, opened 2026-08-21
[Flaky] HyperNode(controller manager mode) "discovers, updates, and removes label topology while reconciling status" case always failed


JesseStutler:
There's a flaky test in hypernode(controller manager mode) always failed, the logs shows:

```
• [FAILED] [121.040 seconds]
HyperNode controller runtime [It] discovers, updates, and removes label topology while reconciling status
/home/runner/work/volcano/volcano/test/e2e/hypernode/controller_runtime_test.go:96

  [FAILED] Timed out after 120.000s.
  Expected success, but got an error:
      <*errors.errorString | 0x3e2ecf75f390>: 
      found 2 discovered HyperNodes, want 1
      {
          s: "found 2 discovered HyperNodes, want 1",
      }
  In [It] at: /home/runner/work/volcano/volcano/test/e2e/hypernode/controller_runtime_test.go:103 @ 08/21/26 06:13:54.388
------------------------------
SS••••••••••••••••••••••••••••S

Summarizing 1 Failure:
  [FAIL] HyperNode controller runtime [It] discovers, updates, and removes label topology while reconciling status
  /home/runner/work/volcano/volcano/test/e2e/hypernode/controller_runtime_test.go:103

Ran 30 of 33 Specs in 472.892 seconds
FAIL! -- 29 Passed | 1 Failed | 0 Pending | 3 Skipped
--- FAIL: TestE2E (472.89s)
FAIL
```

Example link: https://github.com/volcano-sh/volcano/actions/runs/32452259970/job/96684978268?pr=5873

miantalha45:
Hi @JesseStutler , as now a days, I am diving deeply in HyperNodes (#5752), I would love to inspect this flaky test case and fix it.

JesseStutler:
> Hi [@JesseStutler](https://github.com/JesseStutler) , as now a days, I am diving deeply in HyperNodes ([#5752](https://github.com/volcano-sh/volcano/pull/5752)), I would love to inspect this flaky test case and fix it.

Yes if you would like to, but I think @wangyang0616 may fix it since it's raise by him

miantalha45:
Thanks @JesseStutler. i will let @wangyang0616 take the lead. If it's not being worked on, I would be happy to pick it up.

wangyang0616:
[...]

PR #5894, opened 2026-08-24
fix(hypernode): prevent duplicate HyperNodes caused by stale informer cache

### LINKED ISSUE #5887: Label discovery may create duplicate HyperNodes when informer cache is stale
### Description

Label-based topology discovery may create multiple HyperNodes for the same logical topology when consecutive Node updates are processed before the HyperNode informer cache observes the previous reconciliation result.

This produces an incorrect topology tree and may cause the scheduler to observe the same Node through multiple logical HyperNodes. The problem affects the shared HyperNode implementation used by both controller-manager and standalone modes.

This functional bug was identified while investigating the flaky E2E test reported in #5886.

1. Enable label-based HyperNode discovery with a topology label such as volcano.sh/e2e-hypernode-rack.
2. Update two Nodes in quick succession with the same topology label value.
3. List label-discovered HyperNodes.
4. Depending on informer timing, two differently named HyperNodes may be created for the same label value.

The issue was reproduced in the HyperNode E2E workflow:
https://github.com/volcano-sh/volcano/actions/runs/32452259970/job/96684978268

The controller logs show the following sequence:
- The first discovery round generated and created hypernode-e2e-rack-topology-tier1-t4db4.
- Approximately 9 ms later, a second discovery round generated and created hypernode-e2e-rack-topology-tier1-6hwz7 for the same rack-a topology.
- The informer Add event for the first HyperNode arrived only after the second Create decision had already been made.
- Both resources remained until another Node-label event triggered a later discovery.

**Received:** Multiple HyperNodes with different generated names represent the same topology label value. They do not automatically converge without another discovery event.
**Expected:** Discovery must be idempotent. A logical topology should map to one HyperNode even when the informer cache temporarily lags behind API writes.

master

This is not a regression introduced by #5137. The affected cache-dependent name resolution and Create-or-Update decision already existed on master. PR #5137 changed informer ownership and introduced E2E coverage that made the existing race visible.

_No response_

wangyang0616:
#### What type of PR is this?
Fixes duplicate HyperNodes created by label discovery when a completed API write has not yet been observed by the informer cache.

The fix:

- confirms informer cache misses using a single live API snapshot per discovery round;
- reuses existing HyperNode names instead of generating duplicates;
- handles stale-cache `AlreadyExists` responses with an idempotent update;
- prevents overwriting same-name HyperNodes owned by another discovery source;
- adds the required standalone controller RBAC permission.

bugfix

Fixes #5887
relate #5886 

```release-note
None
```

### DIFF
--- a/docs/design/hypernode-controller-standalone.md
+++ b/docs/design/hypernode-controller-standalone.md
@@ -187,7 +187,7 @@ The standalone process initializes only the clients and informers that HyperNode
 | Resource | Access | Purpose |
 | --- | --- | --- |
 | Node | List/Watch | Read topology labels and calculate the node count represented by each HyperNode |
-| HyperNode | List/Watch and reconciliation writes | Maintain topology resources and their status |
+| HyperNode | Get/List/Watch and reconciliation writes | Maintain topology resources and their status |
 | Controller ConfigMap | List/Watch the installation's configuration object through a `metadata.name` field selector | Load and update topology discovery configuration |
 | UFM Secret | Get by name only when referenced; no Watch | Retrieve credentials for the external discovery system |
 | Leader-election Lease | Get/Create/Update as required by leader election | Provide high availability for the standalone controller |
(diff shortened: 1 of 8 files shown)


miantalha45:
Hi @wangyang0616, I went through this PR and the corresponding issue #5887. The fix looks good to me, and I don’t see any further changes needed.

PR #5897, opened 2026-08-24
Fix conformance plugin bypassing lower tier reclaimable evaluation

### LINKED ISSUE #5896: reclaim: `conformance` in a tier above `capacity`/`proportion` silently  disables deserved-based victim filtering (documented capacity-plugin  config is affected)
### Description

When `conformance` is configured in a **higher tier** than `capacity` or `proportion`, Volcano can skip the plugin that protects each queue's `deserved` capacity.

The reason is that `Session.Reclaimable` stops at the **first tier that returns a victim list**.

`conformance` usually returns all non-critical pods as reclaimable, so with the documented layout:

```yaml
tiers:
- plugins:
  - name: priority
  - name: conformance

- plugins:
  - name: capacity
```

the flow becomes:

```text
conformance returns victims
        ↓
Volcano stops
        ↓
capacity.ReclaimableFn is never called
```

That means the check that prevents reclaiming a queue below its `deserved` share is bypassed.

Three queues on a 64-unit cluster:

```text
Queue A: deserved 32, using 16
Queue B: deserved 16, using 24
Queue C: deserved 16, using 24
```

Scale A from 16 → 32.

Expected:

```text
Evict 8 from B
Evict 8 from C
Total: 16 evictions
```

Observed:

* 18 evictions
* C was pushed to `15`, below its deserved `16`
* C's replacement pod then became a reclaimer
* C reclaimed a pod from A, causing an eviction cascade

So an under-deserved queue can end up evicting another under-deserved queue.

Moving `conformance` into the **same tier** as `capacity` fixes the issue:

```yaml
- plugins:
  - name: conformance
  - name: capacity
```

With that change:

* exactly 16 evictions
* 8 from B, 8 from C
* 0 from A
* neither B nor C went below `deserved`

Ideally, `conformance` should `Abstain` when it has nothing to filter, allowing lower tiers such as `capacity` or `proportion` to make the reclaim decision.

At minimum, the documentation should place `conformance` in the same tier as the quota plugin and warn that a permissive `ReclaimableFn` in a higher tier can bypass lower-tier victim filtering.

## Environment

* Volcano v1.15.1
* Kubernetes v1.33 on EKS
* Reproduced with Deployments and KubeRay `RayService`

v1.15.1

_No response_

Paramveersingh-S:
This PR fixes an issue where the `conformance` plugin, when placed in a higher tier than `capacity` or `proportion` (which is the default in Volcano), silently disables deserved-based victim filtering. 

Currently, the `conformance` plugin unconditionally returns `util.Permit` with its list of filtered victims, even when it hasn't actually filtered out any critical pods (`len(victims) == len(candidates)`). When this happens, Volcano's `Session.Reclaimable` and `Session.Preemptable` functions interpret the non-nil victim list as a definitive tier decision, and exit early without evaluating lower tiers. This causes `capacity` and `proportion` plugins in Tier 2 to be completely bypassed, allowing under-deserved queues to evict pods from other under-deserved queues.

This PR updates `evictableFn` and `UnifiedEvictableFn` in the `conformance` plugin to return `util.Abstain` when it does not filter any candidates. This allows the tier system to safely fall through to the next tier so quota-enforcing plugins can do their job.

Fixes #5896

### Special notes for your reviewer:
This change ensures that `conformance` can safely stay in Tier 1 to protect critical pods without inadvertently disabling queue limits when no critical pods are at risk.

### DIFF
--- a/pkg/scheduler/plugins/conformance/conformance.go
+++ b/pkg/scheduler/plugins/conformance/conformance.go
@@ -59,6 +59,10 @@ func (pp *conformancePlugin) OnSessionOpen(ssn *framework.Session) {
 			victims = append(victims, evictee)
 		}
 
+		if len(victims) == len(evictees) {
+			return nil, util.Abstain
+		}
+
 		return victims, util.Permit
 	}
 
@@ -75,6 +79,9 @@ func (pp *conformancePlugin) OnSessionOpen(ssn *framework.Session) {
 				continue
 			}
 			victims = append(victims, evictee)
+		}
+		if len(victims) == len(candidates) {
+			return nil, util.Abstain
 		}
 		return victims, util.Permit
 	})


JesseStutler:
It's not the solution, and should wait for confirmation, not directly throw your PR

PR #5905, opened 2026-08-26
job scheduler: fix allocate action rejecting tasks requesting untracked resources

### LINKED ISSUE #5857: allocate action rejects any task requesting a resource untracked in node.Status.Allocatable, even into an idle node
#### What type of PR is this?

/kind bug

#### What happened:

A task requesting a resource that is never advertised in `node.Status.Allocatable` (confirmed case: `volcano.sh/vgpu-memory-percentage`, a HAMi fractional-vGPU request-side modifier that has no node-wide capacity total to advertise) can never be placed by the `allocate` action, on any node, at any cluster load - including a completely empty, idle node with plenty of real capacity for it.

`allocate.Action.predicate` (the function wired in as the `allocate` action's node filter) does this:

```go
// pkg/scheduler/actions/allocate/allocate.go:985
func (alloc *Action) predicate(task *api.TaskInfo, node *api.NodeInfo) error {
	var statusSets api.StatusSets
	if ok, resources := task.InitResreq.LessEqualWithResourcesName(node.FutureIdle(), api.Zero); !ok {
		statusSets = append(statusSets, &api.Status{Code: api.Unschedulable, Reason: api.WrapInsufficientResourceReason(resources)})
		return api.NewFitErrWithStatus(task, node, statusSets...)
	}
}
```
https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/actions/allocate/allocate.go#L985-L993

`LessEqualWithResourcesName` treats a scalar resource dimension that's absent from `node.FutureIdle()` (built purely from `node.Status.Allocatable`, see `NewNodeInfo`/`SetNode` at https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/api/node_info.go#L403-L407) as capacity `0`. So for any task requesting a nonzero amount of a resource that's structurally never written into `Allocatable`, `ok` is always `false`, and this function returns a hard rejection before it ever reaches `alloc.session.PredicateForAllocateAction` (the actual per-plugin predicate, e.g. deviceshare/vgpu, which does correctly track this via its own device ledger). The real predicate is simply never consulted.

Net effect: a pod requesting `volcano.sh/vgpu-memory-percentage` is rejected by every node during the `allocate` action's filtering step, unconditionally, regardless of whether the node has room. It can only ever get scheduled by falling through to a later action (`preempt`) that doesn't share this specific gate - and if that pod isn't eligible to preempt (lower priority, `preempt` action not configured in the tier list, etc.), it stays Pending forever even on a cluster with zero contention.

A task requesting `volcano.sh/vgpu-memory-percentage` (or any other resource a device plugin tracks entirely through its own ledger instead of `node.Status.Allocatable`) should be allocatable to an idle node with real free capacity for it, the same way a whole-GPU `volcano.sh/vgpu-number` request already is.

#### How to reproduce it:

Reproduced with an isolated unit test built on the existing `pkg/scheduler/uthelper` test harness (not a live cluster): a single pending pod requesting `volcano.sh/vgpu-number`, `volcano.sh/vgpu-cores`, and `volcano.sh/vgpu-memory-percentage` (via `addLimit`-style container resource requests/limits), one empty node whose `Allocatable` declares `vgpu-number`/`vgpu-cores`/`vgpu-memory` (matching `NewGPUDevices`'s real gate in `device_info.go`) but never `vgpu-memory-percentage` (matching HAMi's actual behavior - grepping this repo and Project-HAMi/HAMi turns up nowhere it's ever written into node capacity), `deviceshare` plugin enabled with `VGPUEnable: true`, `gang`/`priority` plugins enabled, no other pods on the cluster.

- Same pod requesting only `vgpu-number` + `vgpu-cores` (both tracked): binds cleanly on the first attempt.
- Same pod, same node, same everything, with `vgpu-memory-percentage` added to the request: `allocate` never binds it - `alloc.predicate` rejects the only node before `PredicateForAllocateAction`/deviceshare's real predicate is ever called.

This test file was a scratch reproduction and was not committed anywhere; happy to write and submit a proper regression test in `pkg/scheduler/actions/allocate/allocate_test.go` if that's useful alongside a fix.

#### Environment:

- Volcano version: current `master` @ `6575302cc650137929678ae010fa129a3553b630`
- Scheduler config: `deviceshare` plugin with `VGPUEnable: true` (HAMi-style fractional vGPU via `volcano.sh/vgpu-memory-percentage`)

Found while reviewing #5784, which fixes the same root-cause gap (a generic scalar-resource arithmetic check with no fallback to the real predicate for resources untracked in `Allocatable`) specifically in the `preempt` action's `normalPreempt`/`ValidateVictims`. This is the identical bug pattern in a different action, out of scope for that PR since it isn't linked to #4863 and isn't covered by that PR's evidence. Analysis and reproduction done with the assistance of Claude Code (AI); I reviewed the code paths and confirmed the reproduction myself before filing.

CodeBuildder:
#### What type of PR is this?

/kind bug

Fixes the allocate action rejecting any task that requests a resource dimension the node never advertises in `node.Status.Allocatable` at all, even into an otherwise completely idle node. `volcano.sh/vgpu-memory-percentage` from HAMi's fractional vGPU support is the concrete case: it's tracked entirely by the device plugin's own ledger, never written into `Allocatable`, so a task requesting it could never be placed by allocate on any node, at any cluster load, and could only ever get scheduled through preemption.

There are three separate occurrences of the same gap in this action:

1. `alloc.predicate` runs a generic scalar-resource arithmetic check first and returns immediately on failure, so `PredicateForAllocateAction` (the real device-aware predicate) never runs if that check fails.
2. `prioritizeNodes` has the identical gap independently: it buckets nodes into idle/future-idle candidates using the same raw arithmetic before scoring, so a node that only fails on an untracked dimension gets dropped before it's ever scored, even if the predicate already accepted it.
3. `allocateResourcesForTask`, which makes the actual bind-or-pipeline decision once a node is chosen, has the same gap a third time, with no predicate recheck at all as a backstop.

All three get the same fallback already used for the preempt and reclaim actions in #5784: when the arithmetic shortfall is confined to dimensions `node.Allocatable` has no entry for at all, treat that as inconclusive instead of a hard no and let the real predicate decide.

Fixes #5857

Depends on #5784 — this branch is built on top of it since it needs the `AllUntrackedByAllocatable` helper `#5784` adds to `pkg/scheduler/api`. The diff here will look larger than the actual change until #5784 merges and this rebases down to just the allocate-specific commit.

Used Claude Code (AI assistant) to investigate this issue and draft the fix and test, following the same pattern already reviewed in #5784. Verified the new regression test fails against the code before this change and passes after. Full `pkg/scheduler/...` suite is green.

```release-note
Fix a bug where the allocate action rejected any task requesting a resource untracked in node.Status.Allocatable (such as a fractional vGPU request), even on an otherwise idle node, forcing such tasks to rely on preemption instead of normal scheduling.
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -947,14 +947,23 @@ func (alloc *Action) prioritizeNodes(ssn *framework.Session, task *api.TaskInfo,
 	var futureIdleCandidateNodes []*api.NodeInfo
 	var idleCandidateNodesInOtherShards []*api.NodeInfo
 	var futureIdleCandidateNodesInOtherShards []*api.NodeInfo
+	// fitsResourcePool reports whether task fits in pool, treating a shortfall confined to
+	// resource dimensions node.Allocatable has no entry for at all as fitting. Every node reaching
+	// this loop already passed alloc.predicate, which applies the identical fallback, so a node
+	// filtered out here purely by an untracked dimension would otherwise be dropped from every
+	// candidate bucket and never scored, even though the real predicate already accepted it.
+	fitsResourcePool := func(pool *api.Resource, node *api.NodeInfo) bool {
+		ok, resources := task.InitResreq.LessEqualWithResourcesName(pool, api.Zero)
+		return ok || api.AllUntrackedByAllocatable(resources, node)
+	}
 	for _, n := range predicateNodes {
-		if task.InitResreq.LessEqual(n.Idle, api.Zero) {
+		if fitsResourcePool(n.Idle, n) {
 			if shardingMode == commonutil.SoftShardingMode && !ssn.NodesInShard.Has(n.Name) {
 				idleCandidateNodesInOtherShards = append(idleCandidateNodesInOtherShards, n)
 			} else {
 				idleCandidateNodes = append(idleCandidateNodes, n)
 			}
-		} else if task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
+		} else if fitsResourcePool(n.FutureIdle(), n) {
 			if shardingMode == commonutil.SoftShardingMode && !ssn.NodesInShard.Has(n.Name) {
 				futureIdleCandidateNodesInOtherShards = append(futureIdleCandidateNodesInOtherShards, n)
 			} else {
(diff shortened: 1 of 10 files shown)


CodeBuildder:
cc @hajnalmt, since you've been the one landing most of the fixes in this exact area lately. Heads up this one depends on #5784 landing first (needs the helper it adds).

CodeBuildder:
Same correction as #5784 (a reviewer there pointed out vgpu-memory-percentage is never requested alone in practice — vgpu-number always has to be set alongside it). This fixture had the same issue; fixed it the same way, re-verified fail→pass with the corrected version. Details in the commit.

PR #5906, opened 2026-08-26
job scheduler: fix topology-aware preemption over-evicting for untracked resources

### LINKED ISSUE #5858: topology-aware preemption's SelectVictimsOnNode has the same untracked-resource gap as #4863, causes over-eviction
#### What type of PR is this?

/kind bug

#### What happened:

`SelectVictimsOnNode` (used by `topologyAwarePreempt`, gated behind the opt-in `enableTopologyAwarePreemption` argument) decides both when to stop evicting victims and whether to reprieve an already-selected victim using a raw scalar-resource comparison with no fallback for resources that are never advertised in `node.Status.Allocatable` - the same class of gap as #5806/#4863, just in a different function:

```go
// pkg/scheduler/actions/preempt/preempt.go:777 (loop: decide whether to stop evicting)
if ssn.SimulateAllocatableFn(ctx, state, currentQueue, preemptor) && preemptor.InitResreq.LessEqual(nodeInfo.FutureIdle(), api.Zero) {
    if err := ssn.SimulatePredicateFn(ctx, state, preemptor, nodeInfo); err == nil {
        ...
        break
    }
}
```
```go
// pkg/scheduler/actions/preempt/preempt.go:813 (reprievePod: decide whether a victim can be spared)
var fits bool
if ssn.SimulateAllocatableFn(ctx, state, currentQueue, preemptor) && preemptor.InitResreq.LessEqual(nodeInfo.FutureIdle(), api.Zero) {
    err := ssn.SimulatePredicateFn(ctx, state, preemptor, nodeInfo)
    fits = err == nil
}
```
https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/actions/preempt/preempt.go#L768-L826

In both spots, `preemptor.InitResreq.LessEqual(nodeInfo.FutureIdle(), api.Zero)` is a plain arithmetic comparison with no fallback: for a resource dimension absent from `Allocatable` (e.g. `volcano.sh/vgpu-memory-percentage`), it can never be satisfied, so the `&&` short-circuits before `SimulatePredicateFn` (the real, device-aware predicate) is ever consulted.

Concretely, for a preemptor requesting such a resource:
- Line 777's `break` condition can never trigger, so the loop keeps popping and evicting every remaining preemptable task on the node instead of stopping once enough is actually free - unnecessary over-eviction of low-priority pods.
- Line 813's `fits` stays `false` unconditionally (the `&&` never reaches `SimulatePredicateFn`), so `reprievePod` never spares a single already-evicted victim - every potential victim ends up in the final `victims` list even when most weren't needed.

This is the same root cause PR #5784 fixes in `normalPreempt`/`ValidateVictims`, but that PR's fix doesn't reach this function's own two gates - it only updates the `ValidateVictims` call at the top of `SelectVictimsOnNode` (line ~760 in that PR), a few lines above where this bug lives.

When topology-aware preemption is enabled, victim selection for a fractional-vGPU-style preemptor should stop once the real predicate says the preemptor fits, the same way #5784 makes `normalPreempt` do - not evict every eligible pod on the node regardless of how much was actually needed.

#### How to reproduce it:

Not reproduced against a live cluster. This is a code-path analysis: `LessEqual`'s behavior for a scalar dimension missing from the comparison target (defaults to `0` under `api.Zero`) is directly verifiable at https://github.com/volcano-sh/volcano/blob/6575302cc650137929678ae010fa129a3553b630/pkg/scheduler/api/resource_info.go#L439-L470, and `node.Status.Allocatable` never carries `volcano.sh/vgpu-memory-percentage` per the analysis in #5784 and issue #5857. I did not build a `topologyAwarePreempt`-specific reproduction (it requires driving `DryRunPreemption`'s parallel candidate-selection path, not just a single-node scenario), and none of the tests added by #5784 exercise `EnableTopologyAwarePreemptionKey: true` - I checked, all of `TestPreemptFractionalVGPU`'s subtests set it to `false`, so this gap isn't covered by that PR's test suite either. Flagging this as a code-inspection finding with a suggested fix direction, not a confirmed live repro.

Suggested fix direction: apply the same `LessEqualWithResourcesName` + untracked-resource fallback (already extracted to `util.AllUntrackedByAllocatable` as of #5784) to both gates here, using `ssn.SimulatePredicateFn` instead of `ssn.PredicateFn` since this path is CycleState-based.

#### Environment:

- Volcano version: current `master` @ `6575302cc650137929678ae010fa129a3553b630`
- Relevant config: `enableTopologyAwarePreemption: true` on the `preempt` action, `deviceshare` plugin with `VGPUEnable: true`

Found while reviewing #5784 alongside #5857 (the `allocate` action instance of the same gap). Analysis done with the assistance of Claude Code (AI); I reviewed the code myself and traced every line referenced above before filing, but as noted this one is not backed by a runtime reproduction the way #5857 is.

CodeBuildder:
#### What type of PR is this?

/kind bug

Fixes topology-aware preemption's `SelectVictimsOnNode` over-evicting victims when the preemptor requests a resource dimension the node never advertises in `node.Status.Allocatable` at all, such as `volcano.sh/vgpu-memory-percentage` from HAMi's fractional vGPU support.

`SelectVictimsOnNode` decides when to stop evicting candidate victims, and which of the evicted victims to reprieve afterward, using the same generic scalar-resource arithmetic as `normalPreempt`: does the preemptor now fit given `node.FutureIdle()` after the simulated evictions so far. Both of the two call sites gate a real, device-aware predicate call behind that arithmetic. For an untracked resource, the arithmetic can never come back true, no matter how much room the real predicate would actually allow once enough has been evicted, so it keeps evicting every remaining candidate instead of stopping once the real predicate is satisfied, and the reprieve pass that would normally give back unnecessary evictions can't recover any of them either. Net effect: unnecessary over-eviction and disruption of running workloads.

Both call sites get the same fallback already used for the preempt and reclaim actions in #5784: when the arithmetic shortfall is confined to dimensions `node.Allocatable` has no entry for at all, treat that as inconclusive instead of a hard no and let the real predicate decide.

Fixes #5858

Depends on #5784 — this branch is built on top of it since it needs the `AllUntrackedByAllocatable` helper `#5784` adds to `pkg/scheduler/api`. The diff here will look larger than the actual change until #5784 merges and this rebases down to just this commit.

Used Claude Code (AI assistant) to investigate this issue and draft the fix and test, following the same pattern already reviewed in #5784. The regression test has three evictable victims at increasing priority and a device predicate satisfied after just one eviction; confirmed it fails against the code before this change (evicts an extra victim it didn't need to) and passes after (evicts only the lowest-priority one). Full `pkg/scheduler/...` suite is green.

```release-note
Fix a bug where topology-aware preemption over-evicted running pods when the preempting task requested a resource untracked in node.Status.Allocatable (such as a fractional vGPU request), disrupting more workloads than necessary.
```

### DIFF
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -758,6 +758,22 @@ func SelectVictimsOnNode(
 		return nil
 	}
 
+	// fitsAfterSimulatedEviction reports whether preemptor fits on nodeInfo given the current
+	// simulated eviction state. The generic scalar-resource comparison is only trustworthy for
+	// resources nodeInfo.Allocatable actually carries a capacity number for. A resource a device
+	// plugin tracks entirely through its own ledger instead (like volcano.sh/vgpu-memory-percentage,
+	// a pure request-side modifier with no coherent node-wide total to advertise) can never pass
+	// it, no matter how much room the real predicate would allow, so a shortfall confined to
+	// untracked dimensions is treated as inconclusive rather than a hard no; the caller still
+	// consults SimulatePredicateFn afterward to make the actual determination.
+	fitsAfterSimulatedEviction := func() bool {
+		if !ssn.SimulateAllocatableFn(ctx, state, currentQueue, preemptor) {
+			return false
+		}
+		ok, resources := preemptor.InitResreq.LessEqualWithResourcesName(nodeInfo.FutureIdle(), api.Zero)
+		return ok || api.AllUntrackedByAllocatable(resources, nodeInfo)
+	}
+
 	var preemptees []*api.TaskInfo
 	for _, task := range nodeInfo.Tasks {
 		if filter == nil {
@@ -788,7 +804,7 @@ func SelectVictimsOnNode(
 			return nil, api.AsStatus(err)
 		}
 
-		if ssn.SimulateAllocatableFn(ctx, state, currentQueue, preemptor) && preemptor.InitResreq.LessEqual(nodeInfo.FutureIdle(), api.Zero) {
+		if fitsAfterSimulatedEviction() {
 			if err := ssn.SimulatePredicateFn(ctx, state, preemptor, nodeInfo); err == nil {
 				klog.V(3).Infof("Pod %v/%v can be scheduled on node %v after preempt %v/%v, stop evicting more pods", preemptor.Namespace, preemptor.Name, nodeInfo.Name, task.Namespace, task.Name)
 				break
@@ -824,7 +840,7 @@ func SelectVictimsOnNode(
 		}
 
 		var fits bool
-		if ssn.SimulateAllocatableFn(ctx, state, currentQueue, preemptor) && preemptor.InitResreq.LessEqual(nodeInfo.FutureIdle(), api.Zero) {
+		if fitsAfterSimulatedEviction() {
 			err := ssn.SimulatePredicateFn(ctx, state, preemptor, nodeInfo)
 			fits = err == nil
 		}
--- a/pkg/scheduler/actions/preempt/preempt.go
+++ b/pkg/scheduler/actions/preempt/preempt.go
@@ -429,12 +429,20 @@ func (pmpt *Action) normalPreempt(
 // 2. The cluster has no free resources, and the queue is not allocatable.
 // 3. The cluster has no free resources, but the queue is allocatable.
 // 4. The node has sufficient aggregate resources, but PredicateFn fails because vNPU or vGPU resources are fragmented.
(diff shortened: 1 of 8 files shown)


CodeBuildder:
cc @hajnalmt, since you've been the one landing most of the fixes in this exact area lately. Heads up this one depends on #5784 landing first too (needs the same helper it adds).

CodeBuildder:
Same correction as #5784 — vgpu-memory-percentage is never requested alone in practice, vgpu-number always has to be set alongside it. Both fixtures in this branch (the inherited one and this PR's own topology-aware one) had the issue; fixed both, re-verified fail→pass with the corrected fixtures.

PR #5912, opened 2026-08-27
ci: run validation workflows on branch pushes

ranxi2001:
#### What type of PR is this?

/kind cleanup

`Code Verify` and `Licenses Lint` currently run for pull requests and `master` pushes, so contributors cannot run the repository's lightweight validation on a fork topic branch before opening a PR.

This change runs those lightweight workflows on branch pushes while excluding `dependabot/**`. Full E2E push validation remains enabled for `master` and becomes opt-in for `e2e/**` branches. Pull-request triggers, workflow jobs, permissions, and release, publication, and synchronization workflows are unchanged.

None.

This follows the branch-push validation pattern from [karmada-io/karmada#4762](https://github.com/karmada-io/karmada/pull/4762). `agent/**` intentionally remains eligible because human contributor branches use that prefix, including [#5826](https://github.com/volcano-sh/volcano/pull/5826). The invalid empty `tags:` filters are removed; tag publishing remains owned by the release workflows.

Contributors can push a candidate commit to an `e2e/**` branch in their fork to run the full suite. Ordinary WIP and experimental branches do not trigger E2E push validation.

Validation:

- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/code_verify.yaml .github/workflows/e2e.yaml .github/workflows/licenses_lint.yaml`
- `git diff --check`
- Fork push at `86bac2b69`: [Code Verify](https://github.com/ranxi2001/volcano/actions/runs/33059312413), [Licenses Lint](https://github.com/ranxi2001/volcano/actions/runs/33059312419), and [E2E Tests](https://github.com/ranxi2001/volcano/actions/runs/33059312721): passed
- Fork push at `0d9ddc938`: [Licenses Lint](https://github.com/ranxi2001/volcano/actions/runs/33156581849): passed; [Code Verify](https://github.com/ranxi2001/volcano/actions/runs/33156581887): in progress; E2E did not trigger for the ordinary topic branch

AI assistance: OpenAI Codex was used to audit the workflow scope, prepare the initial patch, and draft this PR text. The final three-file diff and validation evidence remain the contributor's responsibility.

Rhyme: Volcano branches rise and glow; checks run before reviews can flow.

Prompt: "Prepare a focused Volcano CI change that runs lightweight validation on topic-branch pushes and keeps full E2E push validation opt-in through `e2e/**` branches, without changing release or secret-bearing workflows."

```release-note
NONE
```

### DIFF
--- a/.github/workflows/code_verify.yaml
+++ b/.github/workflows/code_verify.yaml
@@ -2,9 +2,9 @@ name: Code Verify
 
 on:
   push:
-    branches:
-      - master
-    tags:
+    branches-ignore:
+      - "dependabot/**"
+      - "copilot/**"
   pull_request:
 
 permissions: read-all
--- a/.github/workflows/code_verify.yaml
+++ b/.github/workflows/code_verify.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches-ignore:
       - "dependabot/**"
-      - "copilot/**"
   pull_request:
 
 permissions: read-all
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -8,7 +8,6 @@ on:
   push:
     branches-ignore:
       - "dependabot/**"
-      - "copilot/**"
   pull_request:
 
 permissions: read-all
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -6,9 +6,9 @@ concurrency:
 
 on:
   push:
-    branches:
-      - master
-    tags:
+    branches-ignore:
+      - "dependabot/**"
+      - "copilot/**"
   pull_request:
 
 permissions: read-all
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -6,8 +6,9 @@ concurrency:
 
 on:
   push:
-    branches-ignore:
-      - "dependabot/**"
+    branches:
+      - master
+      - "e2e/**"
   pull_request:
 
 permissions: read-all
--- a/.github/workflows/licenses_lint.yaml
+++ b/.github/workflows/licenses_lint.yaml
@@ -2,9 +2,9 @@ name: Licenses Lint
 
 on:
   push:
-    branches:
-      - master
-    tags:
+    branches-ignore:
+      - "dependabot/**"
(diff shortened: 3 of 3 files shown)


.github/workflows/code_verify.yaml
```diff
@@ -2,9 +2,9 @@ name: Code Verify
 
 on:
   push:
-    branches:
-      - master
-    tags:
+    branches-ignore:
+      - "dependabot/**"
+      - "copilot/**"
```
JesseStutler: I don't get why to ignore `copilot/xxx` like branch?

ranxi2001: I initially thought Copilot would submit directly via PR, so there was no need to rely on push-triggered CI to observe things — but it actually creates branches independently, so those branches shouldn't be restricted from triggering CI. I've already removed the `copilot/**` entry from `branches-ignore` in the push CI config.

.github/workflows/e2e.yaml
JesseStutler, line 1: Enabling branch-push validation for lightweight checks makes sense, but this change also runs the full E2E suite on every push to every non-Dependabot branch, including WIP or experimental branches. After a PR is opened, contributors may also get push CI in their fork and pull-request CI upstream.

ranxi2001: How about switching to an allowlist instead? E2E push CI would only run on `master` and `e2e/**` branches (plus PRs) — so contributors can push an `e2e/`-prefixed branch to run the full suite themselves, while lightweight checks still run on every push. This would also avoid the duplicate CI.

```
on:
  push:
    branches:
      - master
      - "e2e/**"
    tags:
  pull_request:
  permissions: read-all
```

ranxi2001: How about switching to an allowlist instead? E2E push CI would only run on `master` and `e2e/**` branches (plus PRs) — so contributors can push an `e2e/`-prefixed branch to run the full suite themselves, while lightweight checks still run on every push. This would also avoid the duplicate CI.

```
on:
  push:
    branches:
      - master
      - "e2e/**"
```

JesseStutler: Yes I prefer this

PR #5913, opened 2026-08-27
fix(capacity): add IgnoreClusterResourceLimit parameter for realCapability calculation

### LINKED ISSUE #5911: :bug:  The capacity plugin may incorrectly prevent a job from being scheduled
### Description

The capacity plugin may incorrectly prevent a job from being scheduled when a node's allocatable resources decrease after a device failure, while the existing workload on that node is still running.

Assume a cluster with four nodes (A, B, C, and D) and **32 GPUs in total**. The queue is configured with a capacity of **32 GPUs**.

1. Submit a distributed `VCJob` requesting **16 GPUs (2 × 8 GPUs)**.

   The job is successfully scheduled as follows:

   ```text
   Node A: 8 GPUs allocated
   Node B: 8 GPUs allocated
   Total:  16 GPUs allocated
   ```

2. After the first job is running, one GPU on Node A becomes unhealthy/unavailable.

   The resource reported by Node A is reduced from 8 GPUs to 7 GPUs. However, the existing Pod on Node A remains running and continues to consume the resources from the original allocation.

   The relevant resource accounting becomes effectively:

   ```text
   Queue capacity:      32 GPUs
   totalResource:       31 GPUs
   realCapacity:        31 GPUs
   attr.allocated:      16 GPUs
   ```

3. Submit another distributed `VCJob` requesting **16 GPUs**.

4. During the `capacity` plugin check, the new candidate has:

   ```text
   attr.allocated = 16 GPUs
   candidate.Resreq = 16 GPUs
   realCapacity = 31 GPUs
   ```

   The capacity check therefore compares the existing allocation plus the candidate request against `realCapacity`:

   ```text
   16 + 16 < 31
   ```

   The new job is rejected by the `capacity` plugin.

> **Note:** This issue has been observed in a production cluster. A minimal standalone reproducer has not yet been prepared.

The failure is triggered by the following sequence of normal cluster events:

```text
VCJob #1 submitted
        |
        v
16 GPUs successfully allocated across Node A and Node B
        |
        v
One GPU on Node A becomes unavailable
        |
        v
Node A allocatable GPU count decreases from 8 to 7
        |
        v
Existing Pod remains running and attr.allocated remains 16 GPUs
        |
        v
capacity plugin recalculates totalResource / realCapacity as 31 GPUs
        |
        v
VCJob #2 requests 16 GPUs
        |
        v
capacity check rejects the candidate
```

The suspected failure is caused by `realCapacity` being derived from the current node resource capacity, while `attr.allocated` still accounts for resources already allocated to running workloads. After a device failure, `realCapacity` can therefore become smaller than the queue's configured capacity even though the affected workload is still running.

The resulting capacity check can reject a new job because the resource accounting combines:

* the reduced `realCapacity` (31 GPUs), and
* the existing allocation (16 GPUs), which still includes the GPU that became unavailable.

No panic, race, or deadlock is involved in this issue.

### Actual result

After one GPU becomes unavailable on a node hosting an already-running workload, `realCapacity` decreases from 32 to 31 GPUs. When a subsequent 16-GPU job is submitted, the `capacity` plugin rejects it because the existing 16-GPU allocation plus the new 16-GPU request exceeds the recalculated `realCapacity` of 31 GPUs.

This can leave the queue unable to schedule a workload even though the queue is configured with a 32-GPU capacity and the resource loss occurred after an existing workload had already been successfully scheduled.

### Expected result

The `capacity` plugin should account consistently for resources that are already allocated to running workloads when node capacity decreases due to a device failure.

A reduction in node allocatable resources should not cause `realCapacity` to be used in a way that makes already-running allocations effectively count against a reduced capacity for a second time and incorrectly prevents subsequent jobs from being scheduled.

The expected behavior is for the scheduler to distinguish between:

* the queue's configured capacity;
* resources currently allocatable on nodes; and
* resources already allocated to running workloads.

This should prevent a transient or externally caused reduction in node capacity from resulting in incorrect queue-level scheduling rejection.

1.14.2

I'd like to fix it by ：

Add an additional configuration parameter to the `capacity` plugin. The parameter should default to `false`, so existing configurations remain unchanged when the parameter is not explicitly configured.

When this parameter is set to `true`, the `capacity` plugin should skip the following calculation when computing the `QueueAttr`:

```go
realCapability.MinDimensionResource(attr.capability, api.Infinity)
```

Instead, `attr.realCapability` should be assigned directly from `attr.capability`:

```go
attr.realCapability = attr.capability
```

This provides an opt-in behavior that prevents `realCapability` from being reduced based on the current node-level resource availability, while preserving the existing behavior by default.

safiya2610:
### What type of PR is this?
/kind bug
/kind feature

Fixes #5911. When a node's allocatable resources decrease (e.g., due to device/GPU hardware failure) while existing workloads on that node continue running, the capacity plugin previously derived `realCapability` by computing `min(cp.totalResource, attr.capability)`. Since `attr.allocated` already accounted for the running workloads, the reduced `realCapability` caused subsequent jobs within the queue's configured capacity to be incorrectly rejected.

This PR introduces an opt-in parameter `capacity.IgnoreClusterResourceLimit` (default `false`) for the capacity plugin. When enabled (`true`), `attr.realCapability` is directly assigned from `attr.capability` without being capped by transient/degraded node resource totals, preventing queue-level rejections while preserving the existing behavior by default.

Fixes #5911

### Special notes for your reviewer:
- Backward-compatible: defaults to `false`.
- Supports both `capacity.IgnoreClusterResourceLimit` and unprefixed `ignoreClusterResourceLimit`.
- Added unit tests in `capacity_test.go` covering GPU, CPU, and Memory degradation scenarios.
- All existing unit tests pass.

### DIFF
--- a/pkg/scheduler/plugins/capacity/capacity.go
+++ b/pkg/scheduler/plugins/capacity/capacity.go
@@ -54,6 +54,9 @@ const (
 	// DRAConsumableCapacityEnable is the key for enabling DRA consumable capacity quota in the capacity plugin
 	DRAConsumableCapacityEnable = "capacity.DRAConsumableCapacityEnable"
 
+	// IgnoreClusterResourceLimit is the key for skipping reducing queue's realCapability by cluster total resource in capacity plugin
+	IgnoreClusterResourceLimit = "capacity.IgnoreClusterResourceLimit"
+
 	DeviceClassCountPrefix = "deviceclass/"
 	DeviceClassCapacitySep = ".deviceclass/"
 )
@@ -76,6 +79,8 @@ type capacityPlugin struct {
 	dynamicResourceAllocationEnable bool
 	// draConsumableCapacityEnable controls whether Capacity dimensions inside DRA are enforced
 	draConsumableCapacityEnable bool
+	// ignoreClusterResourceLimit controls whether queue realCapability is limited by cluster total resource
+	ignoreClusterResourceLimit bool
 }
 
 type queueAttr struct {
@@ -414,9 +419,20 @@ func New(arguments framework.Arguments) framework.Plugin {
 	// Default to k8s feature gate values, allow override via plugin arguments
 	dynamicResourceAllocationEnable := utilfeature.DefaultFeatureGate.Enabled(kubefeatures.DynamicResourceAllocation)
 	draConsumableCapacityEnable := utilfeature.DefaultFeatureGate.Enabled(kubefeatures.DRAConsumableCapacity)
+	ignoreClusterResourceLimit := false
 
 	arguments.GetBool(&dynamicResourceAllocationEnable, DynamicResourceAllocationEnable)
 	arguments.GetBool(&draConsumableCapacityEnable, DRAConsumableCapacityEnable)
+	arguments.GetBool(&ignoreClusterResourceLimit, IgnoreClusterResourceLimit)
+	if !ignoreClusterResourceLimit {
+		arguments.GetBool(&ignoreClusterResourceLimit, "ignoreClusterResourceLimit")
+	}
+	if !ignoreClusterResourceLimit {
+		arguments.GetBool(&ignoreClusterResourceLimit, "capacity.StrictCapability")
+	}
+	if !ignoreClusterResourceLimit {
+		arguments.GetBool(&ignoreClusterResourceLimit, "strictCapability")
+	}
 
 	return &capacityPlugin{
 		totalResource:                   api.EmptyResource(),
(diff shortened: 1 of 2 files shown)


JesseStutler:
@safiya2610 The #5911 still needs more discussion, currently I don't prefer to add more and more flags on the plugin, #5911’s main cocern is that we can't distinguish which unhealthy gpu card belongs to(allocated pod? or free gpu cards), so it's the root cause. If we can distinguish, we can use more better way

Issue #5494, opened 2026-06-24
[Umbrella] Track Volcano performance and scalability work


JesseStutler:
### What is the problem you're trying to solve

We need to keep improving Volcano scheduling performance for large-scale clusters and high-throughput workloads.

This issue is used to track performance problems found during benchmark, profiling and production feedback. The list below will be updated continuously as new bottlenecks or fixes are found.

### Describe the solution you'd like

Use this issue as the parent tracker. Detailed discussion and code changes should still happen in each child issue or PR.

#### Batch Scheduler

- [ ] #5082: Optimize `PrioritizeNodes` process, for example, when scoring, we can use lock-free to score. Partially assigned @JesseStutler 
  - Following PR: #5491

- [ ] #5036: Reduce `SchedulerCache` lock contention. Currently, the `SchedulerCache` has a global lock. In batch creation and batch scheduling scenarios, the lock will slow down the scheduling speed. Because this lock is used during both event handler and session snapshots. Assigned to @songmingming0827 

- [ ] #5173: Reduce unnecessary `DeepCopy` and `Clone`. https://github.com/volcano-sh/volcano/pull/5177 already solved core issue, but other unnecessary and time-consuming deepcopy/clone still need to take a look. This task allows contributors who are interested in it to participate. /help

- [ ] #5236: Reduce API server and etcd pressure from repeated status updates.

Reduce unnecessary snapshots and unnecessary scheduling retries: 
- [ ] https://github.com/volcano-sh/volcano/issues/5551. Assigned to @JesseStutler 
- [ ] https://github.com/volcano-sh/volcano/issues/5537. Assigned to @JesseStutler 

#### Agent scheduler and sharding

- [ ] #5168: Agent-scheduler snapshot performance. Some issues have been resolved, but AgentScheduler's snapshot still has a certain time consumption, requiring analysis in conjunction with our benchmark framework. This task allows contributors who are interested in it to participate. /help

#### Admission

- [ ] #4127: Reduce unnecessary webhooks and avoid registering them by default. Some webhook logic has been migrated, and we are trying to reduce the amount of logic in webhooks as much as possible, replacing it with VAP/MAP and KubeBuilder annotations. However, some logic still exists in the webbook, requiring analysis of the more time-consuming stages in the webhook. For example, we need to combine pprof and analyze the impact of webhooks on performance in batch creation and scheduling scenarios.

### Additional context

[...]

PR #5915, opened 2026-08-28
Perf/prioritize nodes lockfree order

songmingming0827:
#### What type of PR is this?
/kind feature
/area performance
`PrioritizeNodes` already uses `workqueue.ParallelizeUntil` with 16 workers to score nodes in parallel. However, in the original implementation, each worker had to acquire the same `sync.Mutex` after scoring a node before writing the results to `pluginNodeScoreMap` and `nodeOrderScoreMap`. In large clusters, this shared lock causes contention on the node scoring hot path and limits the benefits of parallel scoring.

This PR separates node order scoring from the legacy node map phase:

- `NodeOrderFn` continues to run in parallel. Each worker writes its `orderScore` to an independent index in a preallocated `[]float64`. Errors are only logged, and the corresponding score remains 0, so no shared lock is required.
- After `workqueue.ParallelizeUntil` completes, `PrioritizeNodes` calls `NodeOrderMapFn` serially for every node, in input order. `NodeOrderMapFn` invokes the registered `NodeMapFn` callbacks and returns their per-plugin map scores.
- After collecting the map scores, the existing reduce phase and batch scoring phase continue unchanged.
- When no `NodeMapFn` is registered, `NodeOrderMapFn` returns immediately to avoid unnecessary plugin tier traversal.
- The final aggregation reads each `orderScore` directly by node index, without constructing or looking up `nodeOrderScoreMap`.

This removes the shared `sync.Mutex` from the parallel node order scoring path without executing legacy `NodeMapFn` callbacks concurrently.

As @hzxuzhonghu pointed out during the review of #5491, `NodeMapFn` callbacks do not provide concurrency-safety guarantees and therefore must not be invoked concurrently. This PR addresses that concern directly: only `NodeOrderFn` runs in parallel, while `NodeMapFn` runs serially after the parallel phase completes. Therefore, multiple workers can no longer execute map callbacks concurrently in the `PrioritizeNodes` call path.

The same execution model is applied to both the Volcano scheduler and the agent scheduler.

Partially addresses #5082.
Partially addresses #5494.
Builds on #5491 and addresses the concurrency safety review feedback raised there.

Compared with #5491, this version preserves lock-free parallel node order scoring while moving legacy `NodeMapFn` callbacks out of the parallel section and executing them serially, addressing the concurrency safety concern raised during review.

AI disclosure: ai assisted with code migration, testing, and PR preparation. I reviewed all changes and completed the tests and benchmarks listed below.
#### Tests
The following unit tests and race test passed:
```bash
go test ./pkg/scheduler/util \
  ./pkg/scheduler/framework \
  ./pkg/agentscheduler/framework \
  ./pkg/scheduler/actions/allocate \
  ./pkg/scheduler/actions/backfill \
  ./pkg/scheduler/actions/preempt \
  ./pkg/scheduler/actions/utils \
  ./pkg/agentscheduler/actions/allocate

go test -race ./pkg/scheduler/util \
  -run 'TestPrioritizeNodes(NoRace)?$' \
  -count=1
```
The tests cover the following behavior:

- Combined results from batch scores, order scores, and reduce scores.
- Independent error handling for `NodeOrderFn` and `NodeMapFn`: a failure in one does not suppress the scoring result from the other.
- Behavior when the batch phase or reduce phase returns an error.
- Result integrity when `NodeOrderFn` runs in parallel across a large number of nodes.
- Serial execution of `NodeMapFn` in input node order.

#### Performance test

#5491 demonstrated that removing the worker lock from `PrioritizeNodes` can improve scheduling latency and throughput in large-node scenarios. Because this PR changes the execution model of the map phase, the results from #5491 cannot be directly applied to this version. The updated benchmark results are as follows:

- Environment:

  - VM: single machine with 32 vCPUs and 64 GB of memory

  - KWOK nodes: 1,000

  - Test scenario: 50 jobs, 16 replicas per job, 800 pods in total

  - Pod binding and creation timestamps were read directly from `audit.log` instead of being estimated using Prometheus histograms.

  - Pod latency = binding timestamp - creation timestamp

    P50/P90/P99 were calculated per run using the nearest-rank method. The table reports the averages across five runs.

    Binding window = timestamp of the last binding - timestamp of the first binding

    Throughput = total number of pods / binding window

| Version | Runs | Raw P50 avg | Raw P90 avg | Raw P99 avg | Throughput avg | Binding window avg |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| v1.15.1 baseline | 5 | 346.98 ms | 536.64 ms | 605.27 ms | 625.38 pods/s | 1.280s |
| v1.15.1 with this PR | 5 | 322.05 ms | 498.25 ms | 572.17 ms | 679.20 pods/s | 1.178s |
| Improvement | - | 7.19% lower | 7.15% lower | 5.47% lower | 8.61% higher | 7.95% lower |

- The differences from the results in #5491 are expected because the benchmarks used different plugins (this test only enabled `predicate`, `priority`, and `capacity`), different scheduling cycle settings (the scheduling interval was set to 0 to reduce benchmark variance and eliminate phase effects caused by a fixed scheduling interval), and different Volcano versions. Both the v1.15.1 baseline and v1.15.1 with this PR were tested in the same environment.

```release-note
Node order scoring and node map scoring are now decoupled. As a result, `PrioritizeNodes` accepts `NodeOrderFn` separately, and `NodeOrderMapFn` returns only per-plugin map scores. Errors in either phase no longer suppress scores from the other phase.
```

### DIFF
--- a/pkg/agentscheduler/framework/plugins.go
+++ b/pkg/agentscheduler/framework/plugins.go
@@ -169,32 +169,27 @@ func (f *Framework) BatchNodeOrderFn(task *api.TaskInfo, nodes []*api.NodeInfo) 
 	return priorityScore, nil
 }
 
-// NodeOrderMapFn invoke node order function of the plugins
-func (f *Framework) NodeOrderMapFn(task *api.TaskInfo, node *api.NodeInfo) (map[string]float64, float64, error) {
+// NodeOrderMapFn invokes node map functions of the plugins.
+func (f *Framework) NodeOrderMapFn(task *api.TaskInfo, node *api.NodeInfo) (map[string]float64, error) {
+	if len(f.NodeMapFns) == 0 {
+		return nil, nil
+	}
 	nodeScoreMap := map[string]float64{}
-	var priorityScore float64
 	for _, tier := range f.Tiers {
 		for _, plugin := range tier.Plugins {
 			if !isEnabled(plugin.EnabledNodeOrder) {
 				continue
 			}
-			if pfn, found := f.NodeOrderFns[plugin.Name]; found {
-				score, err := pfn(task, node)
-				if err != nil {
-					return nodeScoreMap, priorityScore, err
-				}
-				priorityScore += score
-			}
 			if pfn, found := f.NodeMapFns[plugin.Name]; found {
 				score, err := pfn(task, node)
 				if err != nil {
-					return nodeScoreMap, priorityScore, err
+					return nodeScoreMap, err
 				}
 				nodeScoreMap[plugin.Name] = score
 			}
 		}
 	}
-	return nodeScoreMap, priorityScore, nil
+	return nodeScoreMap, nil
 }
 
 // NodeOrderReduceFn invoke node order function of the plugins
--- a/pkg/scheduler/util/scheduler_helper.go
+++ b/pkg/scheduler/util/scheduler_helper.go
@@ -27,7 +27,6 @@ import (
 	"math"
 	"math/rand"
 	"sort"
-	"sync"
 	"sync/atomic"
 	"time"
 
@@ -50,6 +49,13 @@ const (
 )
 
 var lastProcessedNodeIndex atomic.Int64
+
+// nodeScoreResult holds the per-node scoring output collected by parallel workers.
+type nodeScoreResult struct {
+	pluginScores map[string]float64
+	orderScore   float64
+	err          error
+}
 
 // CalculateNumOfFeasibleNodesToFind returns the number of feasible nodes that once found,
 // the scheduler stops its search for more feasible nodes.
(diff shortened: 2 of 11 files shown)


songmingming0827:
@JesseStutler Please help me take a look at the pr I submitted

JesseStutler:
I think https://github.com/volcano-sh/volcano/pull/5491/changes#r3489805065 is probably wrong, concurrent reads across multiple goroutines are safe, however, a lock-free approach cannot avoid concurrent writes occurring within `nodeorder` or `nodemap`. I guess it's @hzxuzhonghu 's opinion. But we can't prevent such user's behavior, although I don't suggest `nodeorder` or `nodemap` to operate write of maps

pkg/agentscheduler/framework/plugins.go
```diff
@@ -169,32 +169,27 @@ func (f *Framework) BatchNodeOrderFn(task *api.TaskInfo, nodes []*api.NodeInfo)
 	return priorityScore, nil
 }
 
-// NodeOrderMapFn invoke node order function of the plugins
-func (f *Framework) NodeOrderMapFn(task *api.TaskInfo, node *api.NodeInfo) (map[string]float64, float64, error) {
+// NodeOrderMapFn invokes node map functions of the plugins.
+func (f *Framework) NodeOrderMapFn(task *api.TaskInfo, node *api.NodeInfo) (map[string]float64, error) {
```
JesseStutler, line 173: I don't suggest modifying existing extension points, it will cause compabiliity problem when users upgrade

pkg/scheduler/util/scheduler_helper.go
```diff
@@ -75,35 +74,42 @@ func CalculateNumOfFeasibleNodesToFind(numAllNodes int32) (numNodes int32) {
...
+		// Lock-free write to a pre-allocated slice at a unique index.
+		nodeOrderScores[index] = orderScore
 	}
 	scoreStart := time.Now()
 	workqueue.ParallelizeUntil(context.TODO(), 16, len(nodes), scoreNode)
+
+	pluginNodeScoreMap := map[string]fwk.NodeScoreList{}
+	// NodeMapFn is a legacy extension point without a concurrency-safety contract,
+	// so collect its per-plugin scores serially before running the reduce phase.
+	for index := range nodes {
```
JesseStutler, line 98: But nodemap is serial now when iterate nodes

pkg/agentscheduler/framework/plugins.go
songmingming0827: You're right. We should take existing users and upgrade compatibility into account. Would it be acceptable to document this change explicitly and provide migration guidance? I still believe separating the existing extension points is necessary. In the current design, NodeOrderMapFn combines the NodeOrderFn and NodeMapFn scoring paths, even though they return different types of results and require different processing. Keeping them coupled in the long term makes the implementation and extension-point semantics unnecessarily complex.

pkg/scheduler/util/scheduler_helper.go
songmingming0827: I'll consider following your suggestion in [#5082](https://github.com/volcano-sh/volcano/issues/5082) by adding guidelines for BatchNodeOrderFn and documenting the concurrency expectations for node-scoring extension points. The goal is to guide plugin authors toward concurrency-safe implementations and appropriate use of parallel scoring. With a clear contract in place, we could make node iteration parallel here again while also fixing the previous bug.

songmingming0827:
@JesseStutler If you think the approach I outlined above is reasonable and worth pursuing, I’m happy to continue working on this performance issue and revise the implementation accordingly.

PR #5926, opened 2026-09-01
job scheduler: fix proportion plugin blocking untracked resources on queue checks

### LINKED ISSUE #5907: proportion plugin's Allocatable/Preemptive/Reclaimable checks have the same untracked-resource gap as #4863
#### What happened:

The default `proportion` plugin has the same untracked-resource gap as #4863/#5857/#5858, independently, in three of its own checks: `Allocatable`, `Preemptive`, and `Reclaimable`. All three end up going through `queueAllocatable`'s (or an equivalent) generic scalar-resource arithmetic, which compares a queue's current usage against its deserved share using plain `Resource.LessEqual`. A resource dimension a device plugin tracks entirely through its own ledger, like `volcano.sh/vgpu-memory-percentage` from HAMi's fractional vGPU support, never appears in a queue's `deserved`/`allocated` totals at all, so any comparison involving it can behave incorrectly, in this case blocking the operation outright rather than deferring to a real per-task check.

Concretely, in `pkg/scheduler/plugins/proportion/proportion.go`:

- `AddAllocatableFn` → `queueAllocatable` (used by `ssn.Allocatable`)
- `AddPreemptiveFn` → also `queueAllocatable` (used by `ssn.Preemptive`)
- `AddReclaimableFn` (used by `ssn.Reclaimable`)

I found this concretely while adding regression tests for #5784 (preempt/reclaim) and a new allocate-action fix (#5857): a task requesting `volcano.sh/vgpu-memory-percentage` got blocked by `ssn.Allocatable`/`ssn.Preemptive` even after the node-level predicate/arithmetic fixes in those PRs, purely because the queue's own fairness accounting couldn't reason about the untracked dimension. My regression tests for those PRs had to explicitly disable `proportion`'s `Allocatable`/`Preemptive`/`Reclaimable` checks to isolate the fix under test, which means a real cluster running the default `proportion` plugin (as almost everyone does) could still see this class of bug in the queue-quota layer even after #5784/#5857/#5858 land.

A task requesting a resource that isn't tracked in queue-level `deserved`/`allocated` accounting shouldn't be blocked by that accounting alone; the check should either ignore the untracked dimension or defer to a mechanism that can actually evaluate it, consistent with the fallback pattern used in #5784/#5857/#5858 for the node-level equivalent of this problem.

#### How to reproduce it:

1. Deploy HAMi with fractional vGPU support (`volcano.sh/vgpu-memory-percentage`) and the default `proportion` plugin.
2. Submit a pod requesting `volcano.sh/vgpu-memory-percentage` into a queue that's already near or at its deserved cpu/memory share.
3. Observe that `ssn.Allocatable`/`ssn.Preemptive` can reject the task based on `queueAllocatable`'s arithmetic, independent of whether the underlying node/device actually has room.

#### Environment:

- Volcano scheduler, `pkg/scheduler/plugins/proportion`
- Related: #4863, #5784, #5857, #5858

CodeBuildder:
#### What type of PR is this?

/kind bug

Fixes the default `proportion` plugin (queue fairness/quota enforcement) blocking tasks and misjudging reclaim victims purely because they touch a resource dimension a queue's `deserved` share has no entry for at all, such as `volcano.sh/vgpu-memory-percentage` from HAMi's fractional vGPU support. This is the same class of bug already fixed in the allocate, preempt, and reclaim actions (#5784, #5905, #5906), independently, in three of `proportion`'s own checks:

1. `queueAllocatable` (shared by `AddAllocatableFn` and `AddPreemptiveFn`, used by `ssn.Allocatable` and `ssn.Preemptive`): compares a queue's future usage against its deserved share. A resource dimension `deserved` has no entry for at all always reads as a violation, since `deserved` defaults to zero capacity for it, so a queue with plenty of cpu/memory headroom still gets marked not-allocatable for a task requesting that resource.
2. `AddSimulateAllocatableFn` (used by `ssn.SimulateAllocatableFn`, called from topology-aware preemption's `SelectVictimsOnNode`): the identical arithmetic, duplicated for the simulated dry-run path.
3. `AddReclaimableFn` (used by `ssn.Reclaimable`): the same gap in the other direction. It decides whether a queue is still over its deserved share to keep offering more reclaim victims. A victim task using an untracked resource makes its queue look permanently over-deserved on that dimension alone, regardless of actual cpu/memory usage, so every task in that queue stays reclaimable indefinitely.

All three get the same fallback already used in #5784/#5905/#5906: when the arithmetic violation is confined to dimensions the baseline has no entry for at all, treat that as inconclusive rather than a real violation.

Fixes #5907

Depends on #5784 and #5905 — this branch is built on top of both since it needs `AllUntrackedByAllocatable`/the allocate-action fix. The diff here will look larger than the actual change until those merge and this rebases down to just this commit.

The fallback helper needed a queue-resource variant, since `deserved`/`capability` are plain `Resource` values, not a `NodeInfo`. Added `AllUntrackedByResource` next to the existing `AllUntrackedByAllocatable` in `pkg/scheduler/api`, with the latter now a thin wrapper over the former so existing callers (already reviewed in #5784/#5905/#5906) are unaffected.

Used Claude Code (AI assistant) to investigate this issue and draft the fix and test, following the pattern already reviewed in #5784. Each of the three regression tests is run through the actual `allocate`/`reclaim` actions rather than just the plugin function in isolation, so they exercise the real end-to-end path; confirmed each fails against the code before this change and passes after. Full `pkg/scheduler/...` suite is green.

Beyond the unit tests, also verified live against a real Kubernetes cluster (OrbStack) running the full Volcano stack (CRDs, controller-manager, admission webhook): built the stock v1.15.1 scheduler and this fix as separate binaries, ran the stock one first against a queue with `deserved` cpu/memory set but no `vgpu-memory-percentage` entry, and confirmed a pod requesting that resource sat Pending indefinitely with the scheduler repeatedly logging the queue as overused. Swapped in the fixed binary against the same queue and pod spec, no other changes, and the pod was scheduled, bound, and ran successfully within seconds.

```release-note
Fix a bug where the proportion plugin's queue fairness checks (Allocatable, Preemptive, Reclaimable) could block or misjudge tasks requesting a resource untracked in the queue's deserved share (such as a fractional vGPU request), even when the queue had ample cpu/memory headroom.
```

### DIFF
--- a/pkg/scheduler/actions/allocate/allocate.go
+++ b/pkg/scheduler/actions/allocate/allocate.go
@@ -947,14 +947,23 @@ func (alloc *Action) prioritizeNodes(ssn *framework.Session, task *api.TaskInfo,
 	var futureIdleCandidateNodes []*api.NodeInfo
 	var idleCandidateNodesInOtherShards []*api.NodeInfo
 	var futureIdleCandidateNodesInOtherShards []*api.NodeInfo
+	// fitsResourcePool reports whether task fits in pool, treating a shortfall confined to
+	// resource dimensions node.Allocatable has no entry for at all as fitting. Every node reaching
+	// this loop already passed alloc.predicate, which applies the identical fallback, so a node
+	// filtered out here purely by an untracked dimension would otherwise be dropped from every
+	// candidate bucket and never scored, even though the real predicate already accepted it.
+	fitsResourcePool := func(pool *api.Resource, node *api.NodeInfo) bool {
+		ok, resources := task.InitResreq.LessEqualWithResourcesName(pool, api.Zero)
+		return ok || api.AllUntrackedByAllocatable(resources, node)
+	}
 	for _, n := range predicateNodes {
-		if task.InitResreq.LessEqual(n.Idle, api.Zero) {
+		if fitsResourcePool(n.Idle, n) {
 			if shardingMode == commonutil.SoftShardingMode && !ssn.NodesInShard.Has(n.Name) {
 				idleCandidateNodesInOtherShards = append(idleCandidateNodesInOtherShards, n)
 			} else {
 				idleCandidateNodes = append(idleCandidateNodes, n)
 			}
-		} else if task.InitResreq.LessEqual(n.FutureIdle(), api.Zero) {
+		} else if fitsResourcePool(n.FutureIdle(), n) {
 			if shardingMode == commonutil.SoftShardingMode && !ssn.NodesInShard.Has(n.Name) {
 				futureIdleCandidateNodesInOtherShards = append(futureIdleCandidateNodesInOtherShards, n)
 			} else {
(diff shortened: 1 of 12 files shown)


CodeBuildder:
cc @hajnalmt, since you've been the one landing most of the fixes in this exact area lately. Heads up this one depends on #5784 and #5905 too.

CodeBuildder:
Quick note on the approach: we only skip the budget check for a resource when a queue's deserved/capability has no entry for it at all, never when it's explicitly set, even to zero. cpu and memory are never eligible for this, they're always checked for real. The reason it's ever missing isn't an admin oversight: deserved is calculated from node allocatable, and volcano.sh/vgpu-memory-percentage can't have a real node-wide number since it's a percentage of one specific physical GPU decided at scheduling time. So when it's genuinely absent, muting that one dimension and letting the device plugin's own per-card check decide is the correct read, not a workaround.

cc @hajnalmt, would appreciate your take on this approach and a look at the fix.

CodeBuildder:
Same correction as #5784 — vgpu-memory-percentage is never requested alone in practice, vgpu-number always has to be set alongside it. This branch inherited the same issue from #5784/#5905, plus this PR's own new proportion tests had it too. Fixed all of them, re-verified fail→pass.

PR #5927, opened 2026-09-01
fix(agent-scheduler): skip not-ready nodes in UpdateSnapshot

### LINKED ISSUE #5924: agent-scheduler: nil pointer panic in incremental UpdateSnapshot when placeholder NodeInfo has Node == nil
### Description

`volcano-agent-scheduler` panics with a nil pointer dereference during incremental snapshot update:

```text
UpdateSnapshot → AddOrUpdateNodes → addOrUpdateNode
```

When a Pod event arrives for a node that is not yet (or never will be) in the agent-scheduler cache, `addTask` creates a **placeholder** `NodeInfo` via `NewNodeInfo(nil)` (`Node == nil`, `State = NotReady`). That placeholder increments `Generation` and is moved to the head of the generation doubly-linked list.

The next scheduling cycle’s `UpdateSnapshot` collects nodes **only by Generation**, without filtering `Node == nil` / `!Ready()`. Then `addOrUpdateNode` unconditionally reads `volcanoNodeInfo.Node.Name` and SIGSEGVs.

This is an **agent-scheduler-specific** bug: the same package’s full `Snapshot()` already skips non-Ready nodes, but the incremental `UpdateSnapshot` path missed that guard.

Verified still present on `release-1.14`, `release-1.15`, and current `master`.

### Panic stack (production)

```text
W... node_info.go:363] received argument of nil node, no need to set other resources for
W... node_info.go:258] the argument node is null.
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=...]

goroutine ... [running]:
volcano.sh/volcano/pkg/scheduler/plugins/util/k8s.(*Snapshot).addOrUpdateNode(...)
        pkg/scheduler/plugins/util/k8s/snapshot.go:136
volcano.sh/volcano/pkg/scheduler/plugins/util/k8s.(*Snapshot).AddOrUpdateNodes(...)
        pkg/scheduler/plugins/util/k8s/snapshot.go:128
volcano.sh/volcano/pkg/agentscheduler/cache.(*SchedulerCache).UpdateSnapshot(...)
        pkg/agentscheduler/cache/cache.go:1039
volcano.sh/volcano/pkg/agentscheduler.(*Worker).runOnce(...)
        pkg/agentscheduler/scheduler.go:151
```

`addr=0x20` matches accessing `v1.Node.ObjectMeta.Name` when `*v1.Node` is nil. The two warnings are emitted by `Clone()` → `NewNodeInfo(nil)`, confirming the placeholder.

### Why the placeholder exists

```go
// pkg/agentscheduler/cache/event_handlers.go
if _, found := sc.Nodes[pi.NodeName]; !found {
    newNode := newNodeInfoListItem(schedulingapi.NewNodeInfo(nil)) // Node == nil
    sc.Nodes[pi.NodeName] = newNode
    sc.Nodes[pi.NodeName].info.Name = pi.NodeName
}
// ... Generation++, moveNodeToHead(...)
```

Common cases where the real Node event never arrives (placeholder stays forever):

1. Node labels do not match agent-scheduler `--node-selector` (`nodeCanAddCache` filters it out)
2. Node object already deleted while Pods remain
3. Startup informer race (Pod event before Node event) — explains restart storms at boot

Note: agent-scheduler’s Pod informer watches **all** assigned Pods (`spec.nodeName != ""`), not only Pods it schedules. Any Pod on a cache-external node can create a placeholder.

### Buggy incremental path (no Ready / nil filter)

```go
// pkg/agentscheduler/cache/cache.go — UpdateSnapshot
for node := sc.headNode; node != nil; node = node.next {
    if node.info.Generation <= snapshotGeneration {
        break
    }
    nodesToUpdate = append(nodesToUpdate, node.info) // no Ready()/Node==nil check
}
snapshot.AddOrUpdateNodes(nodesToUpdate)
```

### Crash site

```go
// pkg/scheduler/plugins/util/k8s/snapshot.go
func (s *Snapshot) addOrUpdateNode(nodeInfo *api.NodeInfo) {
    volcanoNodeInfo := nodeInfo.Clone()
    nodeName := volcanoNodeInfo.Node.Name // panic when Node == nil
```

### Contrast: full Snapshot() already filters

```go
for _, value := range sc.Nodes {
    if !value.info.Ready() {
        continue
    }
    snapshot.Nodes[value.info.Name] = value.info.Clone()
}
```

### Related defect

`RemoveNode` deletes from `sc.Nodes` / `NodeList` but does **not** unlink the node from the generation doubly-linked list or reset `headNode`. Stale list items can remain and confuse generation advancement.

### Trigger conditions

Both are required (this is why the process can run for days, then crash):

1. A Pod appears whose node is not in agent-scheduler cache → nil placeholder
2. There are pending tasks so `runOnce` actually calls `UpdateSnapshot` (if `schedCtx == nil`, it returns early)

### Permalinks (`release-1.14` @ `496b6ebff0a2df2d03a3a3071f1f6156c6034c83`, matches v1.14.2 stack lines)

- [UpdateSnapshot collect loop — **missing** Ready/nil filter (bug)](https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/cache.go#L1028-L1039)
- [addOrUpdateNode — panic on `Node.Name`](https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/scheduler/plugins/util/k8s/snapshot.go#L133-L136)
- [addTask placeholder `NewNodeInfo(nil)`](https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/event_handlers.go#L49-L63)
- [RemoveNode without list unlink](https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/event_handlers.go#L370-L404)
- [Full `Snapshot()` Ready() filter — **contrast, already correct**](https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/cache.go#L996-L1001)

## Suggested fix

| Location | Change |
| --- | --- |
| `UpdateSnapshot` | Skip `node.info.Node == nil` (or `!node.info.Ready()`) when collecting `nodesToUpdate` |
| `addOrUpdateNode` | Guard nil `Node` at entry |
| `RemoveNode` | Unlink from doubly-linked list + reset `headNode` |

1. Deploy `volcano-agent-scheduler` with `--scheduler-name=agent-scheduler` (default). No `--node-selector` change is required.
2. Create an already-bound Pod whose `spec.nodeName` points to a **non-existent** node (this creates a `NewNodeInfo(nil)` placeholder in cache):

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: as-repro-ph-1
  namespace: default
spec:
  nodeName: as-repro-fake-node-does-not-exist
  containers:
  - name: pause
    image: registry.k8s.io/pause:3.9
```

3. Create a pending Pod for agent-scheduler so `runOnce` calls `UpdateSnapshot` (use a never-matching `nodeSelector` to keep it pending):

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: as-repro-pend-1
  namespace: default
spec:
  nodeSelector:
    as-repro-never-match: "true"
  containers:
  - name: pause
    image: registry.k8s.io/pause:3.9
```

4. Check agent-scheduler logs / restart count. `spec.schedulerName` must exactly equal `--scheduler-name`.

### Cluster evidence (panic)

Observed on a running cluster (`volcano-agent-scheduler`, exit code 2). From `kubectl logs --previous`:

```text
W0831 07:32:11.251123 node_info.go:363] received argument of nil node, no need to set other resources for
W0831 07:32:11.251133 node_info.go:258] the argument node is null.
I0831 07:32:11.251158 panic.go:792] End scheduling in worker 0 ...
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x16d3985]

goroutine 402 [running]:
volcano.sh/volcano/pkg/scheduler/plugins/util/k8s.(*Snapshot).addOrUpdateNode(...)
        pkg/scheduler/plugins/util/k8s/snapshot.go:136 +0x45
volcano.sh/volcano/pkg/scheduler/plugins/util/k8s.(*Snapshot).AddOrUpdateNodes(...)
        pkg/scheduler/plugins/util/k8s/snapshot.go:128
volcano.sh/volcano/pkg/agentscheduler/cache.(*SchedulerCache).UpdateSnapshot(...)
        pkg/agentscheduler/cache/cache.go:1039 +0x966
volcano.sh/volcano/pkg/agentscheduler.(*Worker).runOnce(...)
        pkg/agentscheduler/scheduler.go:151 +0x20a
```

The same panic was reproduced with the steps above: restart count increased (e.g. 11 → 13), exit code 2, and logs again showed `nil node` / `the argument node is null` while scheduling the pending Pod.

### Production / user path to failure

```text
1. Assigned Pod event (spec.nodeName != "") for a node not in agent-scheduler cache
   (non-existent nodeName, --node-selector mismatch, deleted node, or Pod-before-Node race)
        │
        ▼
2. AddPodToCache → addTask
   if node missing: NewNodeInfo(nil) placeholder (Node == nil, NotReady),
   Generation++, moveNodeToHead
        │
        ▼
3. Pending Pod with schedulerName == --scheduler-name enters scheduling queue
        │
        ▼
4. Worker.runOnce → Pop succeeds → UpdateSnapshot
   generation list collects the placeholder (no Ready()/Node==nil filter)
        │
        ▼
5. AddOrUpdateNodes → addOrUpdateNode
   Clone() logs nil-node warnings → volcanoNodeInfo.Node.Name → SIGSEGV → exit 2
```

### Permalinks (`release-1.14` @ `496b6ebff0a2df2d03a3a3071f1f6156c6034c83`, matches v1.14.2 stack lines)

- Placeholder creation: https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/event_handlers.go#L49-L63
- Incremental collect without Ready/nil filter: https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/cache.go#L1028-L1039
- Panic site (`Node.Name`): https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/scheduler/plugins/util/k8s/snapshot.go#L133-L136
- Full Snapshot Ready filter (contrast): https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/cache.go#L996-L1001
- runOnce calls UpdateSnapshot only when a task is popped: https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/scheduler.go#L137-L153
- RemoveNode without list unlink: https://github.com/volcano-sh/volcano/blob/496b6ebff0a2df2d03a3a3071f1f6156c6034c83/pkg/agentscheduler/cache/event_handlers.go#L370-L404

**Received**

- Nil-node warnings, then panic at `snapshot.go:136` (`addOrUpdateNode`), exit code 2, kubelet restarts the Pod
- Stack: `addOrUpdateNode` → `AddOrUpdateNodes` → `UpdateSnapshot` → `Worker.runOnce`
- Reproduced stably with the steps above (restart count increased; leaving both Pods in place keeps crashing after restart)
- Also observed in a long-running cluster: early restart storm, then stable for days, then crash again when Pods landed on cache-external nodes

**Expected**

- Placeholder / `Node == nil` / NotReady nodes are skipped in incremental `UpdateSnapshot` (same as full `Snapshot()` `Ready()` filter)
- `addOrUpdateNode` guards nil `Node`
- `RemoveNode` unlinks the generation list item and resets `headNode`

v1.14.2 (volcano-agent-scheduler)

- Confirm with: `kubectl logs -n volcano-system <agent-scheduler-pod> --previous`
- `spec.schedulerName` must exactly match `--scheduler-name`; otherwise the pending Pod never enters the queue and `UpdateSnapshot` is not called
- Recent community nil-panic fixes (e.g. `String()` / PodGroup annotation / backfill best-node) do not cover this path

xiezhayang:
Volcano schedules pods with fire and might,
Through ashes of panic we bring back the light;
Not-ready placeholders no longer ignite—
Skip them in snapshots, and workers take flight.

#### What type of PR is this?

/kind bug

Agent-scheduler's incremental `UpdateSnapshot` walks the node generation list and feeds candidates into `Snapshot.AddOrUpdateNodes` **without** filtering on `Ready()`.

When a Pod is bound to a **non-existent** `nodeName`, cache creates a placeholder via `NewNodeInfo(nil)` (`Node == nil`, NotReady, Generation bumped). That placeholder is later included in the incremental update; `addOrUpdateNode` then dereferences `volcanoNodeInfo.Node.Name` and panics:

```
volcano.sh/volcano/pkg/scheduler/plugins/util/k8s.(*Snapshot).addOrUpdateNode
    .../snapshot.go:136
volcano.sh/volcano/pkg/agentscheduler/cache.(*SchedulerCache).UpdateSnapshot
    .../cache.go:1039
```

Full `Snapshot()` already skips `!Ready()` nodes. This PR aligns the incremental path: skip not-ready nodes before `AddOrUpdateNodes` (same style as the full snapshot).

**Verified:**
- Unfixed build + repro (unschedulable pending + Pod with missing `nodeName`) → panic as above
- With this change → `skip updating not-ready node <name> in snapshot`, then `Snapshot updated`, no panic

Fixes #5924

- Primary fix is only in agent-scheduler `UpdateSnapshot` (Ready check before append), consistent with full `Snapshot()`.
- Please cherry-pick to `release-1.15` / `release-1.14` after merge if maintainers agree (bug exists on those lines as well).
- Internal test also needed `--csi-storage=false` on clusters without `CSIStorageCapacity` API; unrelated to this nil-node fix but can block `WaitForCacheSync` and hide the scheduling path.

```release-note
NONE
```


pkg/agentscheduler/cache/cache.go
```diff
@@ -1033,6 +1033,10 @@ func (sc *SchedulerCache) UpdateSnapshot(snapshot *k8sutil.Snapshot) error {
 				node.info.Name, node.info.Generation, snapshotGeneration)
 			break
 		}
+		if !node.info.Ready() {
+			klog.V(5).Infof("skip updating not-ready node %s in snapshot", node.info.Name)
+			continue
+		}
```
miantalha45, line 1039: It would be good to add a regression test case for this panic.

pkg/agentscheduler/cache/cache.go
```diff
@@ -1033,6 +1033,10 @@ func (sc *SchedulerCache) UpdateSnapshot(snapshot *k8sutil.Snapshot) error {
 				node.info.Name, node.info.Generation, snapshotGeneration)
 			break
 		}
+		if !node.info.Ready() {
```
JesseStutler, line 1036: Didn't get the point, but https://github.com/volcano-sh/volcano/issues/5924 's panic trace shows that https://github.com/volcano-sh/volcano/blob/92d6e4cdcc5cb6cb6a10ae9a3025104a0066adba/pkg/scheduler/plugins/util/k8s/snapshot.go#L136 happened the panic, so it means `volcanoNodeInfo.Node` is nil and we can't de-reference it. But the `Ready` func only judge whether the phase of node is Ready

xiezhayang: Thanks @JesseStutler.

For this panic path, Node == nil and !Ready() are the same thing: NewNodeInfo(nil) → setNodeState(nil) sets Phase: NotReady / UnInitialized. So skipping !Ready() keeps the placeholder out of AddOrUpdateNodes, and Node.Name is never dereferenced.

This also matches volcano-scheduler: its Snapshot() filters with !Ready() the same way, and does not add a separate nil-Node check afterward. Agent-scheduler’s full Snapshot() already follows that pattern; this PR only applies the same filter to the incremental UpdateSnapshot path that was missing it.

JesseStutler: Yes you're right, thanks for your reminding

pkg/agentscheduler/cache/cache_snapshot_test.go
```diff
@@ -0,0 +1,89 @@
...
+	schedulingapi "volcano.sh/volcano/pkg/scheduler/api"
+	k8sutil "volcano.sh/volcano/pkg/scheduler/plugins/util/k8s"
+	"volcano.sh/volcano/pkg/scheduler/util"
+)
+
+// TestUpdateSnapshotSkipsNotReadyPlaceholder ensures a placeholder NodeInfo created by addTask
+// for a missing nodeName (NewNodeInfo(nil) → NotReady / UnInitialized) is not published into the
+// incremental scheduling snapshot. Without the Ready() filter, AddOrUpdateNodes would panic
+// on volcanoNodeInfo.Node.Name (see #5924).
+func TestUpdateSnapshotSkipsNotReadyPlaceholder(t *testing.T) {
```
JesseStutler: Suggestions:
1. Better use table-driven style to write unit test
2. The comment looks too AI generated, I prefer not to clarify that this ut is used to test a previous panic issue, but focus on what does this func do
3. Better to rename the func, we could add unit test for the `UpdateSnapshot` func, I think previously we missed to add unit test for the `UpdateSnapshot`, and then you can add arguments for `node==nil` case

xiezhayang: Got it, thanks for the review! So I take it we need a unit test for UpdateSnapshot instead of just covering the panic path. I will push it soon.

xiezhayang: Hi, @JesseStutler ,I've updated the tests based on your suggestions. Let me know if any further changes are needed.

JesseStutler: Thanks, it looks better

PR #5928, opened 2026-09-01
perf(nodeorder): run score plugins in a single node traversal

songmingming0827:
#### What type of PR is this?

/kind feature

/area performance

Before this change, `BatchNodeOrderFn` runs the scoring phase roughly as follows:

(Relevant code: `pkg/scheduler/plugins/predicates/predicates.go` and `pkg/scheduler/plugins/nodeorder/nodeorder.go`)

```text
for each registered plugin {
    run PreScore
    if PreScore returns Skip {
        continue
    }

    CalculatePluginScore {
        score all candidate nodes with 16 workers
        store the score of each node in nodeScoreList

        if the plugin implements ScoreExtensions {
            run NormalizeScore to update nodeScoreList in place
        }

        apply the plugin weight to nodeScoreList and return the per-node scores
    }

    add the plugin's scores to the total score of each node
}
```

With `N` score plugins and `M` candidate nodes, the scheduler still needs to execute `N * M` `Score` calls. Using a separate `NodeScoreList` for each plugin is also necessary because each plugin may implement `NormalizeScore`. The main issue is how these calls are organized:

1. Each plugin starts a separate parallel node traversal, repeatedly creating workers, contexts, and error channels.
2. The same set of `NodeInfo` objects is traversed again for every plugin, increasing memory access and reducing CPU cache locality.
3. Although each plugin scores nodes with 16 workers, plugins are processed one at a time. The next plugin cannot start scoring until the current plugin completes `NormalizeScore`.

This PR scores all candidate nodes in a single 16-worker traversal. Each worker runs all active score plugins for its current node:

(`CalculatePluginScore` no longer describes the new multi-plugin behavior, so it is replaced by `RunScorePlugins`.)

```text
build the PreScore and Score plugin lists

RunScorePlugins {
    run PreScore plugins serially and record skipped plugins
    build activePlugins

    score all candidate nodes with 16 workers {
        for each plugin in activePlugins {
            store the plugin's raw score for the current node in
            activePlugins[pluginIndex].scores[nodeIndex]
        }
    }

    run NormalizeScore for active plugins in parallel
    validate scores, apply plugin weights, and aggregate each node's total score in parallel
}
```

Compared with the previous implementation, this PR:

1. Score N*M times as well, but only starts  one parallel node traversal for all score plugins.
2. Processes all plugins for the same `NodeInfo` within one worker, improving CPU cache locality.
3. Normalizes plugins in parallel and aggregates weighted totals across nodes in parallel.

This reduces repeated parallelization and synchronization overhead and improves scheduling performance in large clusters.

N/A

AI tools were used to assist with the implementation and tests. I reviewed all generated changes and completed the tests and benchmark documented below.

#### Tests

The following unit and race tests passed. They cover the shared scoring helper, the volcano-scheduler `nodeorder` and `predicates` paths, and the agent-scheduler `predicates` adapter:

```bash
go test ./pkg/scheduler/plugins/util/nodescore \
  ./pkg/scheduler/plugins/nodeorder \
  ./pkg/scheduler/plugins/predicates \
  ./pkg/agentscheduler/plugins/predicates

go test -race ./pkg/scheduler/plugins/util/nodescore \
  ./pkg/scheduler/plugins/nodeorder \
  ./pkg/scheduler/plugins/predicates \
  ./pkg/agentscheduler/plugins/predicates
```

The tests cover the following behavior:

- All `PreScore` calls complete before `Score` starts, and all `Score` calls complete before `NormalizeScore` starts.
- When `PreScore` returns `Skip`, only the corresponding plugin's `Score` and `NormalizeScore` phases are skipped.
- Raw scores are normalized, weighted, and aggregated correctly. If any phase fails, no scores are returned and all partially computed results are discarded.
- The fused `nodeorder` and `predicates` paths produce the same results as the previous per-plugin execution, including when `LeastAllocated` and `MostAllocated` are enabled together.
- The agent-scheduler runs `PreScore` and `Score` only for candidate nodes.

#### Performance benchmark

The benchmark compares the original implementation with this PR. Each version was run seven times. The runs with the highest and lowest throughput were excluded, and the remaining five runs were averaged.

- Environment:

  - Machine: one VM with 32 vCPUs and 64 GiB of memory
  - Nodes: 1,000 KWOK nodes
  - Workload: 20 jobs, 500 replicas per job, 10,000 pods in total
  - Scheduler: agent-scheduler with 4 workers
  - Scheduling cycle interval: `0`, to avoid phase effects from a fixed scheduling interval
  - Volcano version: `master`

- Metrics:

  - Pod latency = binding timestamp - creation timestamp
  - P50/P90/P99 are calculated for each run using the nearest-rank method; the table reports the average of the five retained runs
  - Binding window = timestamp of the last binding - timestamp of the first binding
  - Throughput = total number of pods / binding window
  - Pod creation and binding timestamps are read directly from `audit.log` instead of being estimated from Prometheus histograms

| Version | Runs | Raw P50 avg | Raw P90 avg | Raw P99 avg | Throughput avg | Binding window avg |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| master baseline | 5 | 2265.296 ms | 3704.348 ms | 3961.418 ms | 762.772 pods/s | 13.1102 s |
| master with this PR | 5 | 1115.628 ms | 1554.856 ms | 1637.436 ms | 936.232 pods/s | 10.6820 s |
| Improvement | - | 50.75% lower | 58.03% lower | 58.67% lower | 22.74% higher | 18.52% lower |

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/plugins/nodeorder/nodeorder.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder.go
@@ -363,26 +363,25 @@ func (pp *NodeOrderPlugin) NodeOrderFn(task *api.TaskInfo, node *api.NodeInfo, k
 // BatchNodeOrderFn scores all candidate nodes for the given task in one call.
 func (pp *NodeOrderPlugin) BatchNodeOrderFn(task *api.TaskInfo, nodes []*api.NodeInfo, nodeMap map[string]fwk.NodeInfo, state *k8sframework.CycleState) (map[string]float64, error) {
 	nodeInfos := nodescore.NodeInfosForCandidateNodes(nodes, nodeMap)
-	nodeScores := make(map[string]float64, len(nodes))
-
+	preScorePlugins := make([]nodescore.PreScorePluginSpec, 0, len(pp.ScorePlugins))
+	scorePlugins := make([]nodescore.ScorePluginSpec, 0, len(pp.ScorePlugins))
 	for name, entry := range pp.ScorePlugins {
+		scorePlugins = append(scorePlugins, nodescore.ScorePluginSpec{
+			Name:   name,
+			Plugin: entry.plugin,
+			Weight: entry.weight,
+		})
 		if preScorePlugin, ok := entry.plugin.(fwk.PreScorePlugin); ok {
-			skipScore, err := nodescore.RunPreScorePlugin(preScorePlugin, state, task.Pod, nodeInfos)
-			if err != nil {
-				return nil, err
-			}
-			if skipScore {
-				continue
-			}
+			preScorePlugins = append(preScorePlugins, nodescore.PreScorePluginSpec{
+				Name:   name,
+				Plugin: preScorePlugin,
+			})
 		}
+	}
 
-		scores, err := nodescore.CalculatePluginScore(name, entry.plugin, state, task.Pod, nodeInfos, entry.weight)
-		if err != nil {
-			return nil, err
-		}
-		for n, s := range scores {
-			nodeScores[n] += s
-		}
+	nodeScores, err := nodescore.RunScorePlugins(context.TODO(), preScorePlugins, scorePlugins, state, task.Pod, nodeInfos)
+	if err != nil {
+		return nil, err
 	}
 
 	klog.V(4).Infof("Batch Total Score for task %s/%s is: %v", task.Namespace, task.Name, nodeScores)
--- a/pkg/scheduler/plugins/nodeorder/nodeorder_test.go
+++ b/pkg/scheduler/plugins/nodeorder/nodeorder_test.go
@@ -17,7 +17,9 @@ limitations under the License.
 package nodeorder
 
 import (
+	"context"
 	"os"
+	"reflect"
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
@@ -25,6 +27,7 @@ import (
 	"k8s.io/client-go/informers"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 	k8sframework "k8s.io/kube-scheduler/framework"
+	schedframework "k8s.io/kubernetes/pkg/scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/imagelocality"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/interpodaffinity"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/nodeaffinity"
@@ -41,6 +44,7 @@ import (
 	"volcano.sh/volcano/pkg/scheduler/framework"
 	"volcano.sh/volcano/pkg/scheduler/plugins/gang"
 	"volcano.sh/volcano/pkg/scheduler/plugins/util/k8s"
+	"volcano.sh/volcano/pkg/scheduler/plugins/util/nodescore"
 	"volcano.sh/volcano/pkg/scheduler/uthelper"
 	"volcano.sh/volcano/pkg/scheduler/util"
 )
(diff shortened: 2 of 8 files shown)


songmingming0827:
@JesseStutler Please help me take a look at the pr I submitted

pkg/scheduler/plugins/nodeorder/nodeorder.go
```diff
@@ -363,26 +363,25 @@ func (pp *NodeOrderPlugin) NodeOrderFn(task *api.TaskInfo, node *api.NodeInfo, k
 // BatchNodeOrderFn scores all candidate nodes for the given task in one call.
 func (pp *NodeOrderPlugin) BatchNodeOrderFn(task *api.TaskInfo, nodes []*api.NodeInfo, nodeMap map[string]fwk.NodeInfo, state *k8sframework.CycleState) (map[string]float64, error) {
 	nodeInfos := nodescore.NodeInfosForCandidateNodes(nodes, nodeMap)
-	nodeScores := make(map[string]float64, len(nodes))
-
+	preScorePlugins := make([]nodescore.PreScorePluginSpec, 0, len(pp.ScorePlugins))
```
JesseStutler, line 366: I don't get why we need to define a new `PreScorePluginSpec` and `ScorePluginSpec` slice? Looks like the `xxxScorePluginSpec` only contains the interface and name, but `pp.ScorePlugins` already have the name and interface

pkg/scheduler/plugins/nodeorder/nodeorder_test.go
```diff
@@ -443,3 +447,133 @@ func TestInitPlugin(t *testing.T) {
 		})
 	}
 }
+
+func TestBatchNodeOrderFnWithMultipleScorePlugins(t *testing.T) {
```
JesseStutler: Suggest rewriting unit tests to table-driven styles

pkg/scheduler/plugins/nodeorder/nodeorder_test.go
```diff
@@ -443,3 +447,133 @@ func TestInitPlugin(t *testing.T) {
...
+	legacyScores := runScorePluginsSequentially(t, pp.ScorePlugins, pod, nodeInfos)
+	if !reflect.DeepEqual(fusedScores, legacyScores) {
+		t.Fatalf("Least/Most fused scores differ from sequential scores: fused=%v sequential=%v", fusedScores, legacyScores)
+	}
+	if fusedScores["node-a"] == fusedScores["node-b"] {
+		t.Fatalf("expected asymmetric weights to produce distinct node scores, got %v", fusedScores)
+	}
+}
+
+func runScorePluginsSequentially(t *testing.T, plugins map[string]scorePluginEntry, pod *v1.Pod, nodeInfos []k8sframework.NodeInfo) map[string]float64 {
```
JesseStutler: There is no need to manually construct unit tests in this way; focus on testing the functions actually used by the scheduler, rather than attempting to reproduce the issue by opening up the func like this.

JesseStutler:
I like your refactor, would like to push forward, please cc my review comments @songmingming0827

pkg/scheduler/plugins/nodeorder/nodeorder.go
songmingming0827: I introduced these two Spec slices to provide a common input format for the two callers: nodeorder stores the Score plugin and its weight in the same map entry and obtains optional PreScore support through a type assertion, while predicates stores PreScore plugins, Score plugins, weights, and execution order separately. This keeps RunScorePlugins independent of each caller’s internal storage layout.

songmingming0827:
@JesseStutler I've addressed the test comments and explained the rationale for the two Spec slices. Could you please take another look? Thanks!

songmingming0827:
@JesseStutler I've addressed the review comments: clarified the plugin entry types and simplified the tests into table-driven, behavior-focused cases without manually reconstructing the scoring flow. All relevant unit and race tests pass. Could you please take another look? Thanks!

pkg/scheduler/plugins/util/nodescore/score_helper.go
```diff
@@ -45,79 +65,129 @@ func NodeInfosForCandidateNodes(nodes []*api.NodeInfo, nodeMap map[string]fwk.No
...
-	// note that, in such case, size of errCh should be no less than parallelization number
-	workerNum := 16
-	errCh := make(chan error, workerNum)
-	parallelizeContext, parallelizeCancel := context.WithCancel(context.TODO())
-	workqueue.ParallelizeUntil(parallelizeContext, workerNum, len(nodeInfos), func(index int) {
-		nodeInfo := nodeInfos[index]
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
-		s, status := plugin.Score(ctx, cycleState, pod, nodeInfo)
+	if err := ctx.Err(); err != nil {
```
JesseStutler: nit: All these 
```
if err := ctx.Err(); err != nil {
		return nil, err
	}
``` 
will not meet error since you only use `context.Todo()`

pkg/scheduler/plugins/util/nodescore/score_helper.go
```diff
@@ -77,47 +81,101 @@ func CalculatePluginScore(
...
-			return nil, fmt.Errorf("%s returns an invalid score %v for node %s", pluginName, nodeScore.Score, nodeScore.Name)
-		}
-		nodeScore.Score *= int64(weight)
-		nodeScoreList[i] = nodeScore
-		nodeScores[nodeScore.Name] = float64(nodeScore.Score)
+	for index, nodeInfo := range nodeInfos {
+		nodeScores[nodeInfo.Node().Name] = nodeTotalScores[index]
 	}
-
-	klog.V(4).Infof("%s Score for task %s/%s is: %v", pluginName, pod.Namespace, pod.Name, nodeScores)
```
JesseStutler, line 121: We can keep this log to allow admin know the score

JesseStutler:
/approve
/lgtm
Thanks!

PR #5935, opened 2026-09-01
feat: support optional cert-manager admission TLS

### LINKED ISSUE #5936: Improve admission certificate lifecycle management with cert-manager
## Background

Volcano admission TLS certificates are currently generated and managed by chart/operator hooks. This makes certificate ownership and renewal behavior difficult to reason about, and ordinary reconciles should not rotate a still-valid certificate.

## Proposal

- Make cert-manager integration an explicit, optional chart setting.
- When enabled, let cert-manager own the Issuer/Certificate lifecycle and renew certificates before expiry.
- Let cert-manager cainjector populate the CA bundle for every admission webhook configuration, including the CronJob validating webhook.
- Watch the mounted TLS files and hot-reload the admission server after cert-manager renews the Secret.
- Keep the existing self-managed certificate path for clusters without cert-manager.
- Ensure normal Helm/operator reconciles do not rotate certificates unnecessarily.
- Add documentation and install/upgrade/renewal coverage for both modes.

## Related PR

See #5935 for the initial implementation.

fyuan1316:
## Summary

- Add an opt-in cert-manager path for the admission TLS Secret and CA bundle.
- Reload serving certificates with controller-runtime certwatcher when the mounted certificate files change.
- Keep the existing self-signed certificate path and make its install hook safe for upgrades.

## Details

- `custom.certManager.enabled=false` keeps the existing self-managed certificate generation.
- `custom.certManager.enabled=true` creates a self-signed Issuer and Certificate; cert-manager cainjector owns webhook CA bundles.
- The admission manager only updates webhook CA bundles when `--manage-webhook-ca-bundle=true`.
- The cronjob validating webhook is included in cert-manager CA injection.
- Self-managed admission hook dependencies are retained while the pre-install Job runs, preventing premature ServiceAccount deletion.

## Validation

- `go test ./cmd/webhook-manager/...`
- `helm lint installer/helm/chart/volcano`
- Rendered chart with cert-manager enabled and disabled.
- Helm-installed chart with self-managed certificates: install, upgrade, Queue, PodGroup, CronJob, and Volcano scheduler smoke checks.
- OLM Operator deployment was validated separately with cert-manager-managed certificates.

Fixes the certificate ownership and reload behavior without changing the default self-managed mode.

Closes #5936

### DIFF
--- a/cmd/webhook-manager/app/options/options.go
+++ b/cmd/webhook-manager/app/options/options.go
@@ -57,6 +57,9 @@ type Config struct {
 	GracefulShutdownTime time.Duration
 
 	EnableHealthz bool
+	// ManageWebhookCABundle controls whether the webhook manager writes CA data
+	// into webhook configurations. cert-manager mode delegates this to cainjector.
+	ManageWebhookCABundle bool
 	// HealthzBindAddress is the IP address and port for the health check server to serve on
 	// defaulting to :11251
 	HealthzBindAddress string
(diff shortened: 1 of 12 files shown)


JesseStutler:
@fyuan1316 Hi please check and fix the CI failures

fyuan1316:
@JesseStutler 
Hi, thanks for the review. The CI error has been fixed. PTAL when you get a chance.

PR #5947, opened 2026-09-04
fix(rescheduling): stop mistyped scheduler config from panicking the plugin

omlahore:
## What this fixes

The rescheduling plugin reads four values out of its scheduler configuration with unchecked type assertions. YAML does not guarantee the type of a value, so a config that is merely mistyped panics the scheduler rather than being rejected.

**`parseArguments`** asserted `arguments["interval"]` and `arguments["metricsPeriod"]` straight to `string`:

```go
if intervalArg, ok := arguments["interval"]; ok {
    intervalStr = intervalArg.(string)
}
```

The `ok` there checks that the key is present, not that the value is a string. So this config panics:

```yaml
- plugins:
  - name: rescheduling
    arguments:
      interval: 5        # bare number, decodes to int
```

`framework.Arguments` already has `GetString`, which does the comma-ok, logs a warning and leaves the target untouched. Both call sites now use it. `deviceshare.go` (lines 111, 123, 125) and `warmup.go` already read their arguments through these helpers, so this brings rescheduling in line with them.

**`LowNodeUtilizationConf.parse`** had a subtler version of the same thing. It already asserts the container and logs when that fails:

```go
lowConfigs, ok := lowThresholdsConfigs.(map[interface{}]interface{})
if !ok {
    klog.Warningln("Assert lowThresholdsConfigs to map error, abort the configuration parse.")
    return
}
config := make(map[string]int)
for k, v := range lowConfigs {
    config[k.(string)] = v.(int)   // element types never checked
}
```

So the intent to validate is there, it just stops at the container. A quoted number is the easy way to hit it, because `cpu: "20"` decodes to a string:

```yaml
thresholds:
  cpu: "20"
  memory: 30
```

Element conversion moves into `toStringIntMap`, which skips and logs a bad entry instead of panicking, matching what the container check already did. Both threshold loops were identical, so they share it.

## Behaviour

Unchanged for valid config. `interval` and `metricsPeriod` already fell back to their defaults when parsing failed, and that path now actually gets reached instead of being pre-empted by a panic. Bad threshold entries are dropped with a warning naming the key and the type received, and the good entries in the same map still apply.

## Tests

This adds the package's first test file. Five tests covering a non-string `interval`, a non-string `metricsPeriod`, valid values still applying, a threshold map mixing a quoted and an unquoted number, and a non-string key.

I checked that they fail without the change. Reverting just the two source files and keeping the tests gives:

```
--- FAIL: TestParseArgumentsNonStringInterval (0.00s)
panic: interface conversion: interface {} is int, not string [recovered, repanicked]
```

`go build ./pkg/...`, `go vet` and the package tests all pass.

## Not in this PR

`pkg/scheduler/plugins/tdm/tdm.go:71` has the same unchecked element assertion on a label map. Different plugin and a different input path, so I have left it out rather than widening this. Say the word and I will send it separately.

### DIFF
--- a/pkg/scheduler/plugins/rescheduling/low_node_utilization.go
+++ b/pkg/scheduler/plugins/rescheduling/low_node_utilization.go
@@ -65,11 +65,7 @@ func (lnuc *LowNodeUtilizationConf) parse(configs map[string]interface{}) {
 			klog.Warningln("Assert lowThresholdsConfigs to map error, abort the configuration parse.")
 			return
 		}
-		config := make(map[string]int)
-		for k, v := range lowConfigs {
-			config[k.(string)] = v.(int)
-		}
-		parseThreshold(config, lnuc, "Thresholds")
+		parseThreshold(toStringIntMap(lowConfigs, "thresholds"), lnuc, "Thresholds")
 	}
 	targetThresholdsConfigs, ok := configs["targetThresholds"]
 	if ok {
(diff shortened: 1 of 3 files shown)


omlahore:
Correction to my own description: I wrote that `extender.go` already reads its arguments through `GetString`. It does not. It uses a discarded comma-ok, `arguments[ExtenderURLPrefix].(string)` at lines 132-145, which is safe but is not the helper. The actual precedents for `GetString` are `deviceshare.go` at 111, 123 and 125, and `warmup.go`. I have fixed the description. The change itself is unaffected.

PR #5950, opened 2026-09-04
docs: fix capacity plugin tier configuration and add reclaim regression test

### LINKED ISSUE #5896: reclaim: `conformance` in a tier above `capacity`/`proportion` silently  disables deserved-based victim filtering (documented capacity-plugin  config is affected)

devaniketh:
#### What type of PR is this?
/kind documentation

When `conformance` is placed in an earlier tier than `capacity` or `proportion`, `Session.Reclaimable` and `Session.Preemptable` short-circuit at the first tier that returns non-nil victim candidates (`pkg/scheduler/framework/session_plugins.go`). Because `conformance` considers all non-critical pods as candidates, placing it in a higher tier causes victim selection to halt before `capacity` can filter candidates against queue deserved quotas. As a result, queues can have pods reclaimed even when their allocation is within their deserved capacity.

The examples in `docs/user-guide/how_to_use_capacity_plugin.md` previously configured `conformance` in Tier 1 and `capacity` in Tier 2, exposing users to this behavior.

This PR:
1. Updates `how_to_use_capacity_plugin.md` to place `conformance` in Tier 2 alongside `capacity`.
2. Adds a note explaining why victim filtering plugins (`conformance`) and quota plugins (`capacity`/`proportion`) must be configured in the same tier.
3. Adds a regression unit test `TestReclaimTiers_ConformanceAndCapacity` in `pkg/scheduler/actions/reclaim/reclaim_test.go` covering both the same-tier (desired) and multi-tier (short-circuit) behaviors.

Fixes #5896

Volcano rhyme:
> *When workloads surge and clusters strain,*
> *Volcano yields the batch domain.*
> *Keep tiers aligned so quotas stand,*
> *And fair reclaim rules all the land.*

```release-note
Update capacity plugin documentation to place conformance and capacity in the same tier to ensure queue deserved quota is respected during reclaim.

### DIFF
--- a/docs/user-guide/how_to_use_capacity_plugin.md
+++ b/docs/user-guide/how_to_use_capacity_plugin.md
@@ -39,8 +39,8 @@ data:
       - name: priority
       - name: gang
         enablePreemptable: false
-      - name: conformance
     - plugins:
+      - name: conformance
       - name: drf
         enablePreemptable: false
       - name: predicates
@@ -48,6 +48,10 @@ data:
       - name: nodeorder
       - name: binpack
 ```
+
+> [!NOTE]
+> When using eviction actions like `reclaim` or `preempt`, victim eligibility filter plugins such as `conformance` should be placed in the **same tier** as quota plugins (`capacity` or `proportion`).
+> In Volcano, `Session.Reclaimable` and `Session.Preemptable` evaluate plugin tiers in sequence and short-circuit at the first tier that returns a non-nil candidate list. Placing `conformance` in a higher tier by itself causes it to return all non-critical pods, preventing lower-tier plugins like `capacity` from enforcing queue deserved quotas.
 
 ## Configure ancestor reclaim level
 
@@ -74,8 +78,8 @@ data:
       - name: priority
       - name: gang
         enablePreemptable: false
-      - name: conformance
     - plugins:
+      - name: conformance
       - name: drf
         enablePreemptable: false
       - name: predicates
--- a/pkg/scheduler/actions/reclaim/reclaim_test.go
+++ b/pkg/scheduler/actions/reclaim/reclaim_test.go
@@ -476,3 +476,88 @@ func TestReclaimRechecksPredicateAfterEviction(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestReclaimTiers_ConformanceAndCapacity(t *testing.T) {
+	trueValue := true
+	plugins := map[string]framework.PluginBuilder{
+		conformance.PluginName: conformance.New,
+		gang.PluginName:        gang.New,
+		capacity.PluginName:    capacity.New,
+	}
+
+	buildTest := func() uthelper.TestCommonStruct {
+		return uthelper.TestCommonStruct{
+			Plugins: plugins,
+			PodGroups: []*schedulingv1beta1.PodGroup{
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: docs/user-guide/how_to_use_capacity_plugin.md [hunk 1]
```diff
@@ -39,8 +39,8 @@ data:
       - name: priority
       - name: gang
```
hajnalmt, line 40: I would add a same tier only example too here maybe just to promote gang plugin's guard on reclaim if its configured in the same tier as capacity plugin. (I deliberately set `enableRecaimable` on it). 

This is what we use on our cluster:
```
          - name: conformance
          - name: priority
          - name: gang
            enablePreemptable: false  # This allows preemption without gang constraints
            enableReclaimable: false  # This allows reclaiming without gang constraints
          - name: predicates
          - name: capacity
          - name: nodeorder
```
Would love to hear some thoughts about this from others. @JesseStutler @devzizu

pkg/scheduler/actions/reclaim/reclaim_test.go
```diff
@@ -476,3 +476,88 @@ func TestReclaimRechecksPredicateAfterEviction(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestReclaimTiers_ConformanceAndCapacity(t *testing.T) {
```
hajnalmt, line 480: With this test we are wiring down the unexpected behaviour. I am ok with it, just wanted to write it down.

hajnalmt (approved):
/approve
/ok-to-test

Thanks, let's see the tests.

PR #5953, opened 2026-09-07
scheduler: support priority-first gang victim ordering

### LINKED ISSUE #5945: [Feature] Support configurable safe-first and priority-first victim ordering for gang-aware preemption and reclamation

The gang-aware preemption and reclamation actions split each victim workload into two bundle types:

- A **Safe** bundle contains tasks that can be evicted without making the victim workload lose gang readiness.
- A **Whole** bundle contains tasks whose eviction may make the victim workload fall below its gang availability requirements.

The current comparator always uses bundle type as its first ordering key. As a result, every Safe bundle is considered before any Whole bundle, regardless of the priorities of the victim workloads. This behavior is currently hard-coded in both `SortBundlesForPreempt` and `SortBundlesForReclaim`:

- https://github.com/volcano-sh/volcano/blob/4ca2f8251b8a70d66b6c6471d20847ff576422ef/pkg/scheduler/actions/utils/bundle.go#L248-L267
- https://github.com/volcano-sh/volcano/blob/4ca2f8251b8a70d66b6c6471d20847ff576422ef/pkg/scheduler/actions/utils/bundle.go#L269-L292

The existing order is effectively:

```text
victim queue
  -> all Safe bundles, ordered by the remaining policy keys
  -> all Whole bundles, ordered by the remaining policy keys
```

This is a reasonable general-purpose policy because it minimizes gang disruption. However, it cannot express strict workload-priority protection. In such a policy, the scheduler should be able to exhaust a lower-priority workload, including its Whole bundle when necessary, before evicting even elastic tasks from a higher-priority workload.

For example, given two victim workloads with different priorities, users may need this order:

```text
lower-priority workload.Safe
lower-priority workload.Whole, if still necessary
higher-priority workload.Safe
higher-priority workload.Whole, if still necessary
```

This request does not introduce a new workload class, label, annotation, or private priority model. It only changes how Volcano composes its existing victim queue order, `JobInfo.Priority`, Safe/Whole bundle type, and ROI.

There is also an ordering concern specific to `gangReclaim`: victim queue fairness should remain the outermost decision. The current implementation compares Safe/Whole before calling the victim queue comparator, so a Safe bundle from a less-preferred victim queue may be considered before a Whole bundle from the queue selected by `VictimQueueOrderFn`.

Add a `victimOrderPolicy` argument to both `gangpreempt` and `gangreclaim`.

```yaml
actions:
  - name: gangreclaim
    arguments:
      victimOrderPolicy: safe-first
```

Support the following values.

#### `safe-first`

Keep the current disruption-minimizing behavior and make it the default for backward compatibility.

The comparison keys should be:

```text
VictimQueueOrderFn
-> bundle type (Safe before Whole)
-> victim workload priority (lower priority first)
-> ROI
-> deterministic tie-break
```

For `gangPreempt`, all eligible victims are already restricted to the preemptor's queue, so the victim queue key is effectively constant.

#### `priority-first`

Prefer strict workload-priority isolation.

The comparison keys should be:

```text
VictimQueueOrderFn
-> victim workload priority (lower priority first)
-> bundle type within the workload (Safe before Whole)
-> ROI
-> deterministic tie-break
```

The policy should use Volcano's existing `JobInfo.Priority`; no new priority API or workload classification is required. Equal priorities should fall back to the existing ordering and deterministic tie-break behavior.

Under both policies:

- Safe must remain before Whole within the same workload.
- A Whole bundle must only be selectable when `allowWholeBundle` is enabled.
- Selection must stop as soon as the target demand can be satisfied and placement simulation succeeds.
- The same policy must be applied in both sorting passes: before `UnifiedEvictableFn` filtering and after allowed tasks are rebuilt into valid bundles.
- `UnifiedEvictableFn`, bundle integrity, reclaimability, and preemption eligibility semantics must remain unchanged.

Possible acceptance criteria:

1. With no new argument configured, victim ordering remains backward compatible with `safe-first`.
2. In `safe-first`, no Whole bundle is ordered before a Safe bundle from another workload in the same victim queue.
3. In `priority-first`, a lower-priority workload's Whole bundle may be ordered before a higher-priority workload's Safe bundle.
4. Safe is always ordered before Whole for bundles belonging to the same workload.
5. Within a victim queue, lower `JobInfo.Priority` values are considered first.
6. `gangReclaim` applies `VictimQueueOrderFn` before the queue-local Safe/Whole and workload-priority policy.
7. No Whole bundle is selected when `allowWholeBundle=false`.
8. Once sufficient resources have been selected and placement simulation succeeds, no additional bundle is selected.
9. The pre-filter and post-filter sorting passes use the same configured policy and remain deterministic.

### Additional context

The gang-aware eviction design currently describes Safe bundles as lower-cost opportunities and places them before Whole bundles. Keeping `safe-first` as the default preserves that behavior, while `priority-first` provides an explicit opt-in for environments where workload priority is a stronger protection boundary than elasticity.

This should be implemented as a generic Volcano scheduling option. It should not contain environment-specific workload tiers or policy names.

### Contribution

If the community agrees with this direction and would like to move it forward, I would be happy to work on the implementation and add the corresponding end-to-end tests.

OneSizeFitsQuorum:
#### What type of PR is this?

/kind feature

Add `victimOrderPolicy` to `gangpreempt` and `gangreclaim`, using existing `JobInfo.Priority`:

| Policy | Order within each victim queue |
|---|---|
| `safe-first` (default) | Safe/Whole → workload priority |
| `priority-first` | workload priority → Safe/Whole |

Lower workload priorities are evicted first; equal priorities prefer Safe across workloads. Preemption then uses its existing victim job comparator; both actions finish with ROI and deterministic tie-breaks. Configure each action independently; invalid values warn and fall back to `safe-first`.

For both policies, `gangreclaim` compares `VictimQueueOrderFn` first, grouping tied queues by UID before queue-local ordering. A preferred victim queue's Whole can therefore precede another queue's Safe. `gangpreempt` only considers its own queue. Ordering applies within each candidate domain.

Supporting fixes keep selection consistent with the policy:

- Skip disabled Whole bundles before filtering; replay accepted prefixes so rejected Whole trials do not consume later Safe candidates' allowance.
- Select Safe tasks incrementally, preserving task priority within each workload; keep Whole bundles atomic and try zero-victim placement when resources suffice.
- Recover each nomination plan in a temporary statement; roll back the entire attempt on failure before retrying.

`allowWholeBundle` remains the separate Whole permission. Existing gang definitions, pending-task collection, capacity eligibility, and action sequencing are unchanged.

Fixes #5945

Default gang ordering changes: victim queues precede Safe/Whole during reclamation, and workload priority is explicit within each bundle type. Legacy task-level ordering is unchanged.

Local validation: 10 scheduler packages pass; ordering, filtering, selection, and recovery regressions pass 20 repetitions. Paired action-level tests within a single victim queue use identical inputs to verify different victims under each policy and successful target pipelining, for both preemption and reclamation. Coverage also includes Safe task-priority preservation, real capacity filtering under both Whole settings, and failed-plan rollback/retry. The full Gang Eviction cluster E2E suite passed **31/31 specs** on `212de4ef6`, including GP-P1, GR-P1, and GR-P2 ([CI logs](https://github.com/volcano-sh/volcano/actions/runs/34327068121/job/102390410669)). The subsequent rebase onto current master removes the merge commit without changing the merged code tree; CI for the rebased SHA must be checked separately.

Prefix replay has quadratic cumulative work when all Safe candidates are accepted. A local microbenchmark with two pass-through plugins measured approximately 14–15 ms and 15.7 MB allocated for 512 candidates.

AI assisted implementation, tests, documentation, and review; the author remains responsible. Prompt: "Keep gang victim ordering queue-local and limit supporting fixes to filtering, incremental selection, and failed-plan recovery."

```release-note
Gang-aware eviction supports victimOrderPolicy: safe-first (default) or priority-first within victim queues. Both prioritize victim queue ordering during reclamation; default ordering changes accordingly.
```

### DIFF
--- a/docs/design/evictablefn-evolution-for-gang-eviction.md
+++ b/docs/design/evictablefn-evolution-for-gang-eviction.md
@@ -59,14 +59,14 @@ type EvictionContext struct {
 	Job    *JobInfo
 	Task   *TaskInfo
 
-	TargetTasks []*TaskInfo
+	PendingTasks []*TaskInfo
 	HyperNode string
 }
 
 type UnifiedEvictableFn func(ctx *EvictionContext, candidates []*TaskInfo) ([]*TaskInfo, int)
 ```
 
-`TargetTasks` is the exact pending gang target used by entitlement checks and placement simulation, not every pending task of the initiating job. Gang actions replay cumulative candidate prefixes against an unchanged session snapshot: filters must account for the entire input without persisting trial state. A rejected Whole trial is omitted from subsequent prefixes, while accepted victims remain included so quota limits cannot reset between workloads.
+Gang actions can replay cumulative candidate prefixes. Each invocation must evaluate the complete input against the unchanged session snapshot without persisting trial accounting. Rejected trials are omitted from subsequent prefixes.
 
 #### Pros
 
--- a/docs/design/evictablefn-evolution-for-gang-eviction.md
+++ b/docs/design/evictablefn-evolution-for-gang-eviction.md
@@ -59,12 +59,14 @@ type EvictionContext struct {
 	Job    *JobInfo
 	Task   *TaskInfo
 
-	PendingTasks []*TaskInfo
+	TargetTasks []*TaskInfo
 	HyperNode string
 }
 
 type UnifiedEvictableFn func(ctx *EvictionContext, candidates []*TaskInfo) ([]*TaskInfo, int)
 ```
+
+`TargetTasks` is the exact pending gang target used by entitlement checks and placement simulation, not every pending task of the initiating job. Gang actions replay cumulative candidate prefixes against an unchanged session snapshot: filters must account for the entire input without persisting trial state. A rejected Whole trial is omitted from subsequent prefixes, while accepted victims remain included so quota limits cannot reset between workloads.
 
 #### Pros
 
--- a/docs/design/gang-aware-eviction-design.md
+++ b/docs/design/gang-aware-eviction-design.md
@@ -117,7 +117,7 @@ Bundle ordering is performed in two passes for each HyperNode.
 
 In the first pass, the scheduler sorts raw bundles, skips Whole when disabled, and evaluates candidates through `UnifiedEvictable`. Every trial contains the already accepted prefix plus one Safe task or an entire Whole bundle. A trial is accepted only if all its tasks remain eligible. Rejected trials do not enter the next prefix, so partially approved Whole bundles cannot consume later Safe candidates' allowance.
 
-Each filter invocation evaluates an unchanged session snapshot. Replaying the prefix preserves cumulative quota checks across jobs and queues without persisting rejected trials. The capacity plugin shares its leaf and configured ancestor reclaim checks with legacy reclaim, while preserving gang actions' input order. This prioritizes correctness: replay can evaluate a quadratic number of candidate entries; plugins must not mutate session state during filtering.
+Each filter invocation evaluates an unchanged session snapshot. Replaying the prefix preserves the existing plugins' cumulative quota checks without persisting rejected trials. Replay can evaluate a quadratic number of candidate entries; plugins must not mutate session state during filtering. Plugin-result intersection preserves order and is linear per call.
 
 In the second pass, the scheduler re-sorts the rebuilt bundle list before final selection. This second sort is required because plugin filtering can remove or shrink bundles, which changes their effective value and relative priority.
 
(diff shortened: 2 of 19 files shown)


OneSizeFitsQuorum:
Actual gang eviction E2E validation completed successfully in the PR CI:

- Job: https://github.com/volcano-sh/volcano/actions/runs/34100365784/job/101676123343
- Suite result: **30/30 specs passed**, 0 failed, in 1186.450 seconds.
- `GP-P1` passed in 31.161 seconds. It filled a 4-slot cluster with low- and medium-priority victim workloads, submitted a 2-task high-priority preemptor, and verified the result was low-priority victim `Ready=0`, medium-priority victim `Ready=2`.
- `GR-P1` passed in 35.167 seconds. It placed the same victim layout in an overused queue, submitted a 2-task gang in an underused queue, and verified the result was low-priority victim `Ready=0`, medium-priority victim `Ready=2`.

This demonstrates that `priority-first` uses the lower-priority workload's Safe and Whole bundles before evicting the higher-priority workload's Safe bundle for both gangpreempt and gangreclaim.

OneSizeFitsQuorum:
The first GR-P2 CI attempt exposed a test-fixture issue rather than a scheduler-ordering failure. The scenario requested `NodesNumLimit: 6`, but this Kind job has four schedulable workers; the placeholder helper returned without creating placeholders, so the reclaimer was placed by normal `allocate` and `gangreclaim` never ran. Scheduler logs confirmed direct bindings with idle CPU. Commit `bf8f659f7` fixes the fixture to use four workers with two slots each and fills all eight slots with the two-task tight gang plus a six-task/minAvailable-four elastic gang. This preserves the intended discriminator (lower-priority queue Whole=2 versus higher-priority queue Safe=2) while forcing the reclaimer through gangreclaim. The corrected E2E package compiles successfully, and the updated 31-spec gang eviction suite passed in PR CI.

OneSizeFitsQuorum:
Local focused E2E validation is now complete for the corrected `GR-P2` fixture.

- Environment: five-node Kind cluster (one control plane, four workers), Volcano images built from `bf8f659f793459f015fe39957e1b5078cde9e174`
- Selection: 1 of 31 gangevict specs
- Result: 1 passed, 0 failed, 30 skipped
- Duration: 181.820 seconds for the spec
- Observed behavior: the lower-priority victim queue's Whole bundle was reclaimed before the higher-priority victim queue's Safe bundle

This confirms that `VictimQueueOrderFn` remains the outer ordering boundary and `victimOrderPolicy` only orders bundles within a victim queue. The corrected PR CI gangevict job has also passed.

PR #5955, opened 2026-09-07
queue: transition closed queue back to closing on resync if podgroups remain

magic-peach:
#### What type of PR is this?

/kind bug

The default (resync) branch of closedState.Execute in pkg/controllers/queue/state/closed.go hardcodes status.State back to Closed whenever the queue's spec state is Closed, regardless of whether podgroups still exist for that queue.

The other three state handlers (open, closing, unknown) all check len(podGroupList) in this same situation and transition to Closing when podgroups remain, only reporting Closed once the podgroup list is actually empty. closedState is the only one of the four that skips this check.

There is no filed issue for this; the evidence is the inconsistency between closed.go and the other three state files (open.go, closing.go, unknown.go), which all share the same pattern.

None.

The fix makes closed.go match the existing len(podGroupList) == 0 check already present in open.go, closing.go, and unknown.go. Added a test case for a closed queue with a remaining podgroup, confirmed it fails on unmodified code and passes with the fix.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/queue/state/closed.go
+++ b/pkg/controllers/queue/state/closed.go
@@ -44,7 +44,11 @@ func (cs *closedState) Execute(action v1alpha1.Action) error {
 			}
 
 			if specState == v1beta1.QueueStateClosed {
-				status.State = v1beta1.QueueStateClosed
+				if len(podGroupList) == 0 {
+					status.State = v1beta1.QueueStateClosed
+					return
+				}
+				status.State = v1beta1.QueueStateClosing
 				return
 			}
 
--- a/pkg/controllers/queue/state/closed_test.go
+++ b/pkg/controllers/queue/state/closed_test.go
@@ -137,6 +137,15 @@ func TestClosedState_SyncQueueAction(t *testing.T) {
 			expectedState: schedulingv1beta1.QueueStateClosed,
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/controllers/queue/state/closed.go
```diff
@@ -44,7 +44,11 @@ func (cs *closedState) Execute(action v1alpha1.Action) error {
 			}
 
 			if specState == v1beta1.QueueStateClosed {
-				status.State = v1beta1.QueueStateClosed
+				if len(podGroupList) == 0 {
```
hajnalmt, line 47: Okay if I understand you correctly this can happen if the informer is out of sync or something.

hajnalmt (approved):
/approve
/area controllers
/ok-to-test

Thank you for the PR! I agree with you, this is a simple consistency improvement here: `Closed` should represent the terminal state where no PodGroups remain, so moving back to `Closing` when sync observes remaining PodGroups is more accurate. 

Although, I must mention that the impact of this should be negligible or non-existent, for this to happen the informer should be out of sync (which is possible but unlikely) and mostly it will converge to be Closed again. The severity is low enough to not cherry-pick this to previous branches.

hajnalmt (approved):
/lgtm

This will help on: https://github.com/volcano-sh/volcano/issues/5960 too, wont solve it but its a good first step.

Thank you once more!

PR #5957, opened 2026-09-07
fix(cronjob): refetch a conflicting job in the CronJob's namespace

omlahore:
#### What type of PR is this?

/kind bug

`getJobFromTemplate` builds the Job without a Namespace ([cronjob_controller_util.go#L267-L274](https://github.com/volcano-sh/volcano/blob/606c628d364db8de18c045e947d5255ef058aa26/pkg/controllers/cronjob/cronjob_controller_util.go#L267-L274)), so `jobTemplate.Namespace` is the empty string.

`createJob` creates with the CronJob's namespace but refetches the conflicting Job with the template's ([cronjob_controller_handler.go#L388-L396](https://github.com/volcano-sh/volcano/blob/606c628d364db8de18c045e947d5255ef058aa26/pkg/controllers/cronjob/cronjob_controller_handler.go#L388-L396)):

```go
job, err := cc.jobClient.CreateJobClient(cc.vcClient, cronJob.Namespace, jobTemplate)
...
case errors.IsAlreadyExists(err):
    existingJob, err := cc.jobClient.GetJobClient(cc.vcClient, jobTemplate.Namespace, jobTemplate.Name)
```

So the AlreadyExists path always looks in the empty namespace, never finds the job, and takes the NotFound branch.

None filed. Evidence is below.

#### Bug evidence

Affected version: master at 606c628d364db8de18c045e947d5255ef058aa26.

User path: a CronJob fires, the Job for that scheduled time already exists (a retried sync, or a duplicate event after an apiserver hiccup). Create returns AlreadyExists, the controller refetches to decide whether it owns the existing Job, and instead concludes the Job vanished.

Adding a test that drives `realJobClient` against a namespace-aware fake clientset, on unmodified master:

```
W cronjob_controller_handler.go:399] Job cj-29813040 disappeared after creation conflict, retrying
--- FAIL: TestCreateJobRefetchesConflictingJobInCronJobNamespace
    createJob returned jobs.batch.volcano.sh "cj-29813040" not found, want the existing job
```

The controller reports that a Job disappeared while that Job is sitting in the CronJob's namespace. With the fix the same test passes.

This is invisible to the package's existing tests because `fakeJobClient.GetJobClient` ignores its namespace argument and returns a stored job ([injection.go#L122-L129](https://github.com/volcano-sh/volcano/blob/606c628d364db8de18c045e947d5255ef058aa26/pkg/controllers/cronjob/injection.go#L122-L129)), so the namespace can never be wrong under the mock.

`go test ./pkg/controllers/cronjob/`, `go vet` and `gofmt` are clean.

The new test uses `realJobClient` rather than the package's `fakeJobClient`, because the mock is what hides this class of bug.

```release-note
Fixed a CronJob controller bug where an already-existing Job was looked up in the wrong namespace after a creation conflict.
```

### DIFF
--- a/pkg/controllers/cronjob/cronjob_conflict_namespace_test.go
+++ b/pkg/controllers/cronjob/cronjob_conflict_namespace_test.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Copyright 2025 The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cronjob
+
+import (
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	batchv1 "volcano.sh/apis/pkg/apis/batch/v1alpha1"
+	volcanoclient "volcano.sh/apis/pkg/client/clientset/versioned/fake"
+)
+
+// getJobFromTemplate leaves ObjectMeta.Namespace empty, so the AlreadyExists
+// branch of createJob must refetch with the CronJob's namespace.
+func TestCreateJobRefetchesConflictingJobInCronJobNamespace(t *testing.T) {
+	scheduledTime := time.Date(2026, 9, 7, 12, 0, 0, 0, time.UTC)
+	cronJob := &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{Name: "cj", Namespace: "team-a", UID: "cj-uid"},
+	}
+
+	// The job the previous reconcile already created, in the CronJob's namespace.
+	existing := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            getJobName(cronJob, scheduledTime),
+			Namespace:       cronJob.Namespace,
+			OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(cronJob, controllerKind)},
+		},
+	}
+
+	vcClient := volcanoclient.NewSimpleClientset(existing)
+	cc := newFakeController()
+	cc.vcClient = vcClient
+	cc.jobClient = &realJobClient{}
+
+	got, err := cc.createJob(cronJob, scheduledTime)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controllers/cronjob/cronjob_conflict_namespace_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	batchv1 "volcano.sh/apis/pkg/apis/batch/v1alpha1"
+	volcanoclient "volcano.sh/apis/pkg/client/clientset/versioned/fake"
+)
+
+// getJobFromTemplate leaves ObjectMeta.Namespace empty, so the AlreadyExists
```
hajnalmt: Remove the comment please.

hajnalmt (commented):
Could we also set Namespace: cj.Namespace in getJobFromTemplate()? I’d prefer to keep your change here as well: the generated Job should carry its intended namespace, and the conflict lookup should explicitly use the same namespace as the create request. Maybe a dedicated helper-level assertion for the namespace can be added too.
 
https://github.com/volcano-sh/volcano/blob/606c628d364db8de18c045e947d5255ef058aa26/pkg/controllers/cronjob/cronjob_controller_util.go#L267-L274

It would have made a lot more sense to me, honestly I would fix it at two places, so this won't come up again.

pkg/controllers/cronjob/cronjob_conflict_namespace_test.go
```diff
@@ -0,0 +1,59 @@
...
+// branch of createJob must refetch with the CronJob's namespace.
+func TestCreateJobRefetchesConflictingJobInCronJobNamespace(t *testing.T) {
```
hajnalmt: I'd prefer TestCreateJobRefetchesConflictingJob, with the namespace check retained as one of its assertions.

omlahore:
I guess you're right, it makes more sense to fix it in both places so it can't come back. I set Namespace from the CronJob in getJobFromTemplate now and kept the refetch change as well. There's a namespace check in TestGetJobFromTemplate too, and it fails without the template change.

pkg/controllers/cronjob/cronjob_conflict_namespace_test.go
omlahore: I removed it, and the other one inside the test too.

pkg/controllers/cronjob/cronjob_conflict_namespace_test.go
omlahore: It's called TestCreateJobRefetchesConflictingJob now, and I kept the namespace check in it.

hajnalmt (changes requested):
/area controllers
/approve

Thanks for the update. You addressed all of my feedback 😊 Even the test rename and the comment cleanup look good too. I will merge it in the upcoming week if there is no objection from others. We should also consider cherry-picking it to the release branches.

hajnalmt (approved):

PR #5959, opened 2026-09-08
webhooks: report all invalid subGroupPolicy entries, not just the first

magic-peach:
#### What type of PR is this?
/kind bug

validateNetworkTopology broke out of the policy loop as soon as it found one subGroupPolicy with both highestTierAllowed and highestTierName set. A PodGroup with more than one bad policy still gets rejected either way, but the rejection message only names the first offending policy, so a user fixing it resubmits and gets rejected again for the next one. Removed the break so every bad policy is reported in the same pass.

Fixes #

This PR was prepared with AI assistance (Claude Code), reviewed and tested by me before submitting. Verified with `go test ./pkg/webhooks/admission/podgroups/validate/...`, all cases pass including a new one covering two bad policies reported together.

```release-note
Admission webhook now reports all invalid subGroupPolicy entries in a PodGroup's networkTopology, not just the first one found.
```


hajnalmt (approved):
/approve
/remove-kind bug
/area webhooks

Thanks for the PR! This is not a bug too, because these PodGroups were already rejected, but users now see all offending subgroup policies in a single response. I can accept it as an improvement like https://github.com/volcano-sh/volcano/pull/5962 merging it if no objection from others comes up.

PR #5962, opened 2026-09-10
webhooks: report all invalid hierarchical weight entries, not just the first

magic-peach:
#### What type of PR is this?

/kind bug

`validateHierarchicalAttributes` in the queue admission webhook returned as soon as it found one bad weight in the `volcano.sh/hierarchy-weights` annotation, or one queue conflicting in the sub path check. A queue with more than one bad weight still gets rejected either way, but the rejection message only names the first problem, so a user fixing it resubmits and gets rejected again for the next one. This is the same class of bug already fixed for `subGroupPolicy` in the podgroup webhook (PR #5959).

No issue filed, this is self-discovered while reading the same file after that fix. Evidence: creating a Queue with `volcano.sh/hierarchy: root/sci` and `volcano.sh/hierarchy-weights: bad/-1` (two invalid weights, one non-numeric and one non-positive) is rejected with an error naming only the first bad weight (`bad`); after fixing that one and resubmitting, it is rejected again for the second (`-1`).

Collected all bad weights and all sub path conflicts into the same error list instead of returning on the first one.

Fixes #

This PR was written with AI assistance (Claude Code), under my direction and review. I wrote and ran the tests, and reviewed and understand the change.

Added `TestValidateHierarchicalAttributesReportsAllBadWeights`, confirmed it fails against the pre-fix code (reports 1 error instead of 2) and passes against the fix. Full `pkg/webhooks/admission/queues/validate` package test suite still passes, `go vet` and `gofmt` clean.

```release-note
Queue admission now reports every invalid `volcano.sh/hierarchy-weights` entry and sub path conflict at once instead of only the first one.
```

### DIFF
--- a/pkg/webhooks/admission/queues/validate/validate_queue.go
+++ b/pkg/webhooks/admission/queues/validate/validate_queue.go
@@ -164,13 +164,14 @@ func validateHierarchicalAttributes(queue *schedulingv1beta1.Queue, fldPath *fie
 		for _, weight := range weights {
 			weightFloat, err := strconv.ParseFloat(weight, 64)
 			if err != nil {
-				return append(errs, field.Invalid(fldPath, hierarchicalWeights,
+				errs = append(errs, field.Invalid(fldPath, hierarchicalWeights,
 					fmt.Sprintf("%s in the %s is invalid number: %v",
 						weight, hierarchicalWeights, err,
 					)))
+				continue
 			}
 			if weightFloat <= 0 {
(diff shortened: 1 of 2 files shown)


hajnalmt (approved):
/remove-kind bug
/kind cleanup
/area webhooks
/approve

Thanks for the PR. This is definitely not a bug, because, invalid HDRF hierarchy weights were already rejected, but I can accept it as an improvement for drf plugin users. We should probably consolidate the hierarchical queue validation and usage on the long run.. Merging it if no objection comes from others.

Issue #5751, opened 2026-07-27
[LFX 2026 Term 3] Support Generic xPU Topology-Aware Scheduling


JesseStutler:
### What is the problem you're trying to solve

Volcano currently schedules accelerator resources mainly by their aggregate quantity on a Node. This is not sufficient for communication-intensive workloads because the same number of free xPUs can provide very different communication performance depending on their physical interconnect topology.

For example, a Huawei 16-NPU server may contain two independent 8-NPU HCCS domains. If six devices are free in one domain and two are free in the other, the Node has eight free NPUs in total but cannot satisfy an 8-NPU request that requires one connected domain. A similar problem can occur on GPU systems containing separate NVLink-connected groups or NVSwitch fabrics. Other xPU interconnects, such as MetaX sGPU and MetaXLink, have similar scheduling requirements.

The topology boundary may also extend beyond one Kubernetes Node. Systems such as GB200 NVL72 connect GPUs from multiple compute trays through NVLink switch trays into one rack-scale NVLink fabric. In contrast, an ordinary multi-node GPU cluster may have NVLink only inside each Node and use InfiniBand or RoCE between Nodes. The scheduler must distinguish these cases and place a gang according to the actual physical topology.

Relying only on kubelet-side `GetPreferredAllocation` is too late. By that point the Node has already been selected, so the scheduler cannot reject a Node whose free devices are fragmented across topology domains, coordinate device choices across a PodGroup, or prevent concurrent scheduling decisions from selecting the same devices. The feature must therefore make physical xPU topology and device availability visible during scheduling.

Topology discovery and allocation mechanisms differ across vendors and Kubernetes versions. Existing deployments primarily use Device Plugins, while newer environments may use DRA `ResourceSlice` and `ResourceClaim`. Other integrations may expose information through Node annotations or dedicated CRDs. Coupling the scheduling algorithm to one of these sources would make the feature difficult to adopt and maintain.

### Describe the solution you'd like

Introduce a generic scheduler-side xPU topology-aware scheduling framework and plugin. The scheduler should consume one canonical topology model, independent of whether topology data comes from Node annotations, a CRD, a Device Plugin companion component, or DRA.

The initial implementation can use Node annotations or mock topology objects with KWOK. Device Plugin and DRA integrations can be added later without changing the scheduling policy or workload API.

#### Case 1: Multiple device domains in one Node

[...]

PR #5965, opened 2026-09-11
docs: add generic xpu topology-aware scheduling design

miantalha45:
#### What type of PR is this?

/kind documentation

Adds a design proposal for generic xPU topology-aware scheduling in Volcano.

It defines the proposed topology provider, canonical topology cache, workload API, scheduling behavior, adapter contract, readiness and freshness handling, security model, compatibility boundaries, rollout plan, and validation strategy.

AI assistance was used to help structure and edit this design document. I reviewed the final content and take responsibility for it.

Related to #5751 

```release-note
NONE

### DIFF
--- a/docs/design/generic-xpu-topology-aware-scheduling.md
+++ b/docs/design/generic-xpu-topology-aware-scheduling.md
@@ -1,0 +1,1210 @@
+# Generic xPU Topology-Aware Scheduling
+
+**Status:** Proposal
+
+**Author:** Talha Amjad ([@miantalha45](https://github.com/miantalha45))
+
+## Table of Contents
+
+- [Abstract](#abstract)
+- [Problem](#problem)
+- [Goals](#goals)
+- [Non-goals](#non-goals)
+- [Architecture](#architecture)
+  - [Existing Volcano Building Blocks](#existing-volcano-building-blocks)
+  - [Proposed Architecture](#proposed-architecture)
+- [Design Details](#design-details)
+  - [Provider and Topology Ingestion](#provider-and-topology-ingestion)
+    - [Provider Lifecycle and Source Boundary](#provider-lifecycle-and-source-boundary)
+      - [Provider Readiness Contract](#provider-readiness-contract)
+      - [Initial Node Annotation Contract](#initial-node-annotation-contract)
+      - [Annotation Writer Authorization and Validation](#annotation-writer-authorization-and-validation)
+    - [Canonical Topology Model](#canonical-topology-model)
+    - [Topology Cache and Session Snapshot](#topology-cache-and-session-snapshot)
+    - [Availability and Aggregate Resource Safety](#availability-and-aggregate-resource-safety)
+  - [Workload Policy](#workload-policy)
+    - [Proposed API Types and Semantics](#proposed-api-types-and-semantics)
+    - [Admission Webhook Validation](#admission-webhook-validation)
+  - [Providers and Allocation Adapters](#providers-and-allocation-adapters)
+    - [Allocation Adapter Contract](#allocation-adapter-contract)
+    - [Resource Ownership and DeviceShare Coexistence](#resource-ownership-and-deviceshare-coexistence)
+  - [Scheduling and Gang Reservation](#scheduling-and-gang-reservation)
+    - [Per-Task Filter and Score](#per-task-filter-and-score)
+    - [Plugin Responsibilities and Boundaries](#plugin-responsibilities-and-boundaries)
+    - [Scheduling Pipeline and Existing Component Integration](#scheduling-pipeline-and-existing-component-integration)
+    - [Proposed Plugin and GangPlan Contracts](#proposed-plugin-and-gangplan-contracts)
+    - [Gang Planning Flow](#gang-planning-flow)
+    - [Transaction Participant Contract](#transaction-participant-contract)
+  - [Topology Cases](#topology-cases)
+  - [Lifecycle and Recovery](#lifecycle-and-recovery)
+    - [Failure and Retry Semantics](#failure-and-retry-semantics)
+  - [Performance and Observability](#performance-and-observability)
+  - [Compatibility and Security](#compatibility-and-security)
+- [Delivery and Review](#delivery-and-review)
+  - [Implementation Map](#implementation-map)
+    - [Alpha Scope Boundary](#alpha-scope-boundary)
+  - [Rollout and Implementation Phases](#rollout-and-implementation-phases)
+  - [Validation Plan](#validation-plan)
+  - [Open Questions for Review](#open-questions-for-review)
+  - [References](#references)
+
+## Abstract
+
+Volcano currently schedules accelerators primarily as aggregate Node resources. That works only when all free devices on a Node are interchangeable. Communication-heavy AI workloads instead depend on the physical interconnect between devices: eight free devices may be unusable for an eight-device Pod when they are split across two independent device domains.
+
+This proposal adds an optional, vendor-neutral xPU topology layer to Volcano. It separates topology discovery, scheduling policy, and runtime allocation enforcement:
+
+```mermaid
+flowchart LR
+    Sources[Topology sources] --> Cache[Canonical topology cache]
+    Cache --> Plugin[xPU scheduler plugin]
+    Plugin --> Adapter[Allocation adapter]
+    Adapter --> Bind[Kubernetes bind or DRA]
+
+    classDef source fill:#e8f1ff,stroke:#2563eb,color:#172554
+    classDef scheduler fill:#eafaf1,stroke:#16a34a,color:#14532d
+    classDef runtime fill:#fff7e6,stroke:#d97706,color:#78350f
+    class Sources source
+    class Cache,Plugin scheduler
+    class Adapter,Bind runtime
+```
+
+The initial implementation is deliberately narrow: full devices, Node-local domains, an annotation or mock provider, and feature-gated behavior. The same model can later support topology CRDs, Device Plugin companions, DRA `ResourceSlice`s, and declared cross-Node fabrics without changing the workload API or scheduling policy.
+
+## Problem
+
+Aggregate accelerator counts do not describe communication topology. Consider a 16-NPU Node with two independent eight-NPU HCCS domains:
+
+```text
+Node A
+├── domain A0: 6 free NPUs
+└── domain A1: 2 free NPUs
+
+aggregate free NPUs: 8
+largest connected free domain: 6
+```
+
+An eight-NPU Pod that requires one connected domain cannot run on this Node, despite the aggregate resource check succeeding. The same issue occurs with separate NVLink-connected GPU groups, NVSwitch domains, MetaXLink domains, and similar xPU interconnects.
+
+Volcano must also distinguish ordinary multi-Node clusters from rack-scale fabrics:
+
+```text
+Ordinary multi-Node cluster              Declared rack-scale xPU fabric
+---------------------------              ------------------------------
+Node A: local domain                     Node A ─┐
+Node B: local domain                             ├─ one physical fabric
+Node C: local domain                     Node C ─┘
+
+Network links connect Nodes.             The provider declares a real xPU fabric.
+Local domains stay independent.          Pods may share one fabric across Nodes.
+```
+
+Free devices on ordinary Nodes must not be treated as one shared xPU domain. A Device Plugin's kubelet-side `GetPreferredAllocation` is insufficient because it runs after Node selection and cannot reject fragmented Nodes, coordinate a PodGroup, or reserve IDs across concurrent scheduling attempts.
+
+## Goals
+
+1. Make physical device domains and availability visible before binding.
+2. Support required and preferred local-domain and fabric affinity.
+3. Prefer compact allocations that avoid avoidable device fragmentation.
+4. Plan and reserve all devices for a gang unit before any Pod in that plan binds.
+5. Keep provider parsing independent from scheduling policy and enforcement.
+6. Reconcile device health, allocation, reservation, release, and restart state.
+7. Preserve existing scheduling behavior when the feature is disabled.
+8. Keep the scheduling path indexed and bounded, do not scan all cluster devices for every scheduling attempt.
+
+## Non-goals
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/design/generic-xpu-topology-aware-scheduling.md [hunk 1]
```diff
@@ -0,0 +1,1210 @@
...
+3. Prefer compact allocations that avoid avoidable device fragmentation.
+4. Plan and reserve all devices for a gang unit before any Pod in that plan binds.
+5. Keep provider parsing independent from scheduling policy and enforcement.
+6. Reconcile device health, allocation, reservation, release, and restart state.
+7. Preserve existing scheduling behavior when the feature is disabled.
+8. Keep the scheduling path indexed and bounded, do not scan all cluster devices for every scheduling attempt.
+
+## Non-goals
+
+The first milestone does not implement vendor discovery, device configuration, NCCL ring construction, MIG/vGPU/fractional-device allocation, xPU-aware preemption, or SubGroup-level xPU policies. It does not replace DRA, Device Plugins, `deviceshare`, HyperNodes, or aggregate Kubernetes resource accounting. It also does not infer topology from device indices, PCI addresses, or aggregate capacity: providers must publish explicit topology facts.
```
devzizu, line 117: The proposal mentions: "_Updates and rollback must cover allocation, release, **preemption**, health changes, and failed scheduling cycles._".

Are we excluding the **preemption** action in the non-goals section for the first version?

The main thing that worries me is **cache correctness under preemption**: the preempt action evicts device-holding pods whether or not the plugin is aware of it. If the reservation state doesn't observe eviction, you leak devices until the scheduler restarts.

On the other hand, I'm not sure if you are deferring instead "Topology-aware victim selection", which I think it's OK for now. Either option should be clear here, if you can please improve the sentence.

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+
+Kubernetes does not provide multi-Pod atomic binding. This design guarantees atomic **planning and reservation before binding**, not rollback of Pods that Kubernetes has already bound successfully.
```
devzizu, line 119: > "Kubernetes does not provide multi-Pod atomic binding"

Can you clarify what do you mean with this sentence?

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+    Affinity TopologyAffinityMode `json:"affinity,omitempty"`
+}
+```
+
+For one requirement, exactly one selector is set:
+
+- `extendedResourceName` identifies a full-device Kubernetes extended resource, such as `nvidia.com/gpu`.
+- `claimName` identifies the Pod `resourceClaims` alias for a future DRA-backed workload. It is not accepted in alpha.
+
+`DeviceDomainAffinity` and fabric `Affinity` accept `Required` or `Preferred`. `Required` is a hard filter and requires an allocation adapter that can enforce the selected device IDs: no matching healthy, enforceable topology means no placement. `Preferred` is a score only: a normal placement remains valid if no matching domain or exact-ID enforcement is available. `AllocationStrategy` initially accepts `Compact` or an omitted value. `Compact` is a preference, never a hard constraint.
```
devzizu, line 651: Should we keep the vocabulary `hard`/`soft` as used in the [network topology](https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/docs/design/Network%20Topology%20Aware%20Scheduling.md?plain=1#L530)?

Example of the resulting user experience:

```
# E.g.: A PodGroup that constrains both
spec:
  networkTopology:
    mode: hard                      # one vocabulary
  xpuTopology:
    deviceDomainAffinity: Required  # same concept with a different word
```

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+
+At `OpenSession`, the scheduler captures a consistent cluster and topology view. Every topology Node fact records its Node UID and `resourceVersion`. A bounded session-view coordinator captures the normal `SchedulerCache` snapshot and a topology snapshot without holding both cache locks at once, then compares the topology snapshot's Node identity map with the corresponding Nodes in the cluster snapshot. It retries at most two times on a mismatch. This comparison is indexed by Node name and does not scan individual devices.
```
devzizu, line 504: "_retries at most two times_" - can you please clarify the reasoning?

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+
+If the bounded capture cannot produce a consistent view, the session continues without topology preference data. A `Required` request receives `XPUTopologyDataNotReady` for that scheduling cycle, while a `Preferred` request treats topology as unavailable preference data. Predicate and score callbacks use the accepted topology revision for their entire scheduling session, while reservation operations revalidate selected IDs and Node identity against the live ledger before succeeding. This gives stable planning input without allowing two sessions to reserve the same device.
+
+```go
+type TopologySessionView struct {
```
devzizu, line 509: Could you clarify why `TopologySnapshot` doesn't just live in `ClusterInfo`?

That's how network topology already does it, i.e., ClusterInfo carries HyperNodes, ..., and `HyperNodesReadyToSchedule`, all filled inside `Snapshot()` under the cache lock ([ref](https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/pkg/scheduler/actions/allocate/allocate.go#L195)).

That's the same readiness semantics you're specifying for `XPUTopologyDataNotReady`, and because the topology is captured inside the same `Snapshot()` call as the nodes, there's no second lock and no need for a retry mechanism + budget.

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+        "observedAt": "2026-08-04T12:00:00Z",
+        "devices": [
+          {
+            "id": "nvidia://GPU-4c2e",
+            "resourceName": "nvidia.com/gpu",
+            "localDomain": "node-a/nvlink-0",
+            "health": "Healthy"
+          }
+        ],
+        "localDomains": [
```
devzizu, line 332: The annotation examples seems to be building domain IDs from the Node names, but you mentioned as well that "Node name is diagnostic data only", and readiness is verified with "provider ID + Node UID".

Node names are reusable, i.e., a replaced node commonly returns with the same name and a new UID. After replacement, the tombstone kept for the old node's live devices and the new node's fresh domain carry an identical `DeviceDomainID`, and DeviceDomain has `NodeName` but no `NodeUID`, so the object itself can't distinguish them.

Also, it seems that DeviceDomainID and FabricDomainID are never defined, can you take a look?

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+
+Defaults are conservative: omitted local-domain affinity means no same-domain constraint, omitted fabric means no fabric constraint and omitted compact strategy adds no xPU-specific packing preference. Workloads never name device, domain, fabric, or vendor identifiers.
+
+The webhook validates feature-gate use, one resource selector per requirement, and recognized values. It cannot prove that a runtime adapter can enforce selected IDs, the scheduler rejects a Required request with an explicit reason when no compatible adapter is active.
+
+#### Proposed API Types and Semantics
+
+`xpuTopology` is one optional field on `JobSpec` and its scheduler-canonical `PodGroupSpec` equivalent. It does not replace existing `networkTopology`, `subGroupPolicy`, Pod `affinity`, or Pod `topologySpreadConstraints`. Those fields continue to express network placement, group membership, and ordinary Kubernetes Pod placement. `xpuTopology` expresses only accelerator-domain intent.
+
+The names below are proposed for API review. They make the ownership and validation surface concrete. The alpha implementation must not expose a different implicit annotation API for workload intent.
```
devzizu, line 622: If we assume this in alpha, then Deployment, StatefulSet, and bare-Pod users, who simply set `schedulerName: volcano` have no way to use this feature at all. Should we consider integrating this API with annotations? (topic discussed in the community meeting)

devzizu, line 622: If we assume this in alpha, then Deployment, StatefulSet, and bare-Pod users, who simply set schedulerName: volcano have no way to use this feature at all. Should we consider integrating this API with annotations? (topic discussed in the community meeting)

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+| A confirmation deadline expires after bind handoff | Report the missing confirmation and stop new Required plans from using those IDs. | Mark IDs `Unreconciled` until authoritative reconciliation. |
+| Scheduler restarts with exact adapter reservations | Reload provider facts, recover active adapter reservations and allocations, then reconstruct safe ledger state before enabling Required placement. | Block the adapter-owned resource when recovery fails or leaves an unknown reservation. |
+| A Pod is already bound when another gang member fails | Kubernetes cannot unbind it atomically. | Reconcile the bound allocation. Release only unconfirmed IDs. |
+
+An xPU topology failure is distinct from a normal Node predicate failure. The former means the topology snapshot cannot satisfy the declared device-domain rule, the latter means an otherwise topology-valid Node failed a standard check such as CPU, taints, volumes, ports, or DRA predicate behavior. Events and metrics should preserve that distinction for users and operators.
```
devzizu, line 1080: "_Events and metrics should preserve that distinction for users and operators._"

I would also recommend adding a condition (with a `Reason` field) so that `kubectl describe podgroup` helps users find the reason behind scheduling failures.

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1210 @@
...
+The webhook makes invalid intent fail at admission instead of becoming an ambiguous runtime placement failure:
+
+1. Reject a non-empty `xpuTopology` policy unless `XPUTopologyAwareScheduling` is enabled.
+2. Require at least one resource requirement and exactly one of `extendedResourceName` or `claimName` in each requirement.
+3. Require a valid extended resource name. Reject CPU and memory because the first scope is full accelerator devices, not CPU/NUMA topology.
+4. Accept only `Required` or `Preferred` affinity and `Compact` as the initial allocation strategy.
+5. Reject duplicate requirements targeting the same resource selector in one policy.
+6. Reject `claimName` in alpha. A later DRA-capable version will validate the alias at admission and resolve it against each Task Pod at runtime.
+7. Reject a policy that sets `fabric` with anything other than one resource requirement in alpha.
+8. Treat `xpuTopology` as immutable on both the Job and its scheduler-canonical PodGroup after creation. Users create a new Job to change topology intent instead of changing the policy below active reservations or allocations.
```
devzizu, line 668: Why this needs to be immutable after creation?

docs/design/generic-xpu-topology-aware-scheduling.md
devzizu, line 651: "_AllocationStrategy` initially accepts `Compact` or an omitted value._"

If it's empty what does it mean? Should we add `Spread` as a new option aside `Compact`?

devzizu (commented):
Thanks for this @miantalha45‚it's a serious piece of work, and the level of detail is well beyond what I expected at this stage. I've done a first pass and left comments inline, but I still have some sections to go through in the next few days.

docs/design/generic-xpu-topology-aware-scheduling.md
miantalha45: existing SchedulerCache and ClusterInfo.Snapshot() path adopted.

docs/design/generic-xpu-topology-aware-scheduling.md
miantalha45: good point. I clarified this is an alpha safety boundary, not an inherent API requirement. Mutable policy would need coordinated versioning and reservation/replan semantics. until that exists, accepting updates could invalidate active plans or create mixed gang intent.

docs/design/generic-xpu-topology-aware-scheduling.md
miantalha45: adopted

miantalha45:
thank you so much @devzizu for this detailed review. I have addressed all your comments and also made a lot of changes in document. Kindly have a look on it again now.

miantalha45:
Hi @JesseStutler @hajnalmt @devzizu @wangyang0616 
I have added the proposed subgroup xPU topology design, including policy API, scheduling flow, and reservation semantics. the document is ready for review from my side. Please share any feedback before I begin implementation.

docs/design/generic-xpu-topology-aware-scheduling.md
```diff
@@ -0,0 +1,1569 @@
...
+
+The alpha policy applies uniformly to its scheduling unit: the complete PodGroup for `spec.xpuTopology`, or one matching SubJob for `subGroupPolicy[].xpuTopology`. Every Task admitted under that policy must resolve every declared extended-resource requirement to exactly one regular container with an integral whole-device request. A Task that does not request that resource is not silently skipped, it makes its policy unit ineligible with `XPUTopologyUnsupportedPodRequest`. A parent `xpuTopology` policy and any subgroup `xpuTopology` policy are mutually exclusive in alpha. This avoids ambiguous fabric scope and inheritance. Heterogeneous accelerator Task templates, init-container accelerator requests, per-container policy targeting, and hierarchical parent-plus-subgroup xPU composition are deferred because they require explicit inheritance and allocation-lifecycle semantics. Exact API names and versions require API review. The following examples are illustrative.
```
devzizu: > A Task that does not request that resource is not silently skipped, it makes its policy unit ineligible with `XPUTopologyUnsupportedPodRequest`

This rule worries me, for example, take our own [TF example](https://github.com/volcano-sh/volcano/blob/master/example/kubecon-2019-china/tf-sample/tf-example.yaml). Following the rules in this paragraph: the unit is the complete PodGroup, so it includes `ps`. `ps` doesn't request `nvidia.com/gpu`, so it "makes its policy unit ineligible" rather than being skipped. As a result, one gets `XPUTopologyUnsupportedPodRequest` and the entire job never schedules, including the workers.

That said, I don't think "silently skip Tasks that don't request the resource" is the right fix either, and I assume that's what this rule is defending against.

So the goal is right, I'd just move where we fail closed: Apply the parent policy only to taskss that request the declared resource, and have the webhook reject a policy whose resource is requested by zero Tasks in the PodGroup. WDYT?

PR #5968, opened 2026-09-11
fix: propagate hot-reloaded configuration to agent scheduler workers

### LINKED ISSUE #5967: Agent Scheduler hot-reloaded configuration is not propagated to existing worker Frameworks
## Description

The Agent Scheduler's configuration watcher reloads `actions`, `tiers`, and `configurations` when the scheduler configuration file changes, but the new configuration is not propagated to the worker `Framework` instances that perform scheduling.

`pkg/agentscheduler/scheduler.go` describes the Agent Scheduler as supporting dynamic configuration management with file watching and hot-reload. `watchSchedulerConf()` calls `loadSchedulerConf()` after a configuration file change, and `loadSchedulerConf()` replaces `sched.actions`, `sched.tiers`, and `sched.configurations`.

However, workers are created only once during `Scheduler.Run()`:

`worker.framework = framework.NewFramework(sched.actions, sched.tiers, sched.cache, sched.configurations)`

The scheduling loop subsequently executes:

`for _, action := range worker.framework.Actions`

There is no corresponding update of the existing worker frameworks after `loadSchedulerConf()` completes.

As a result, after a successful configuration reload, the scheduler-level configuration can differ from the configuration actually used by running workers.

For example, if the scheduler starts with:

```yaml
actions:
  - name: allocate
```

and the configuration is changed at runtime to enable another action:

```yaml
actions:
  - name: allocate
  - name: backfill
```

the file watcher reloads the new configuration, but existing workers continue executing the action list stored in their already-created Framework instances.

The reverse case is also problematic: if an action/plugin is removed from the configuration, existing workers can continue using the previous Framework and therefore continue executing the removed behavior.

This also applies to plugin tiers/configurations because these values are supplied to `framework.NewFramework()` only when the worker Framework is created.

### Relevant source path

1. `Scheduler.Run()` loads the initial configuration and creates each worker's Framework from `sched.actions`, `sched.tiers`, and `sched.configurations`.

2. `watchSchedulerConf()` detects configuration changes and calls `loadSchedulerConf()`.

3. `loadSchedulerConf()` replaces the scheduler's configuration fields.

4. `Worker.runOnce()` continues to execute `worker.framework.Actions`, which belongs to the Framework created before the reload.

There is therefore no propagation path from the newly loaded scheduler configuration to existing worker Frameworks.

### Expected behavior

If Agent Scheduler configuration hot-reload is supported, a successfully reloaded configuration should be reflected by subsequent scheduling cycles.

In particular:

* newly enabled actions/plugins should become effective;
* removed actions/plugins should stop being used;
* updated plugin/action configuration should be applied consistently;
* worker Framework/plugin lifecycle should remain valid when the runtime configuration changes.

If some parts of the scheduler configuration are intentionally restart-only, the reload path should explicitly reject or document those changes rather than reporting the configuration as successfully reloaded.

### Actual behavior

The configuration watcher reports/loads the new configuration into the `Scheduler`, but existing workers retain the Framework created with the previous configuration.

Consequently, the configuration visible in the scheduler state can differ from the configuration actually used for scheduling.

### Steps to reproduce

1. Start Agent Scheduler with a scheduler configuration containing a known action set, for example:

```yaml
actions:
  - name: allocate
```

2. Run Agent Scheduler with multiple worker threads so that the worker Frameworks are created and remain active.

3. Modify the scheduler configuration while Agent Scheduler is running to add another valid action/plugin configuration, for example:

```yaml
actions:
  - name: allocate
  - name: <another-valid-action>
```

4. Wait for the configuration watcher to process the file change.

5. Submit workload that would exercise the newly configured action.

6. Compare the configuration reported by the scheduler after reload with the actions actually executed by the existing workers.

The reverse test should also be performed:

1. Start with two valid actions.
2. Remove one action from the configuration.
3. Wait for the reload event.
4. Verify whether existing workers continue executing the removed action.

**Received:**

The configuration reload updates `Scheduler.actions`, `Scheduler.tiers`, and `Scheduler.configurations`, but existing worker Frameworks continue to hold the configuration from the time they were created.

**Expected:**

After a successful hot reload, subsequent scheduling cycles should use the newly loaded configuration, or the scheduler should clearly require a restart for configuration changes that cannot safely be applied at runtime.

`master` (current development branch).

This report is based on source inspection of the current Agent Scheduler implementation. I have not yet reproduced the behavior in a live Kubernetes cluster.

The issue is particularly relevant because the Agent Scheduler source explicitly describes its configuration handling as providing "dynamic configuration management with file watching and hot-reload capabilities."

The normal Volcano Scheduler takes the currently loaded actions/plugins/configurations under the scheduler mutex inside each scheduling cycle, whereas the Agent Scheduler captures these values when each worker Framework is created. This makes the two scheduler implementations behave differently with respect to configuration reload.

A correct fix should also account for action/plugin initialization semantics. `Action.OnActionInit()` is documented as initialization performed once when the framework is created, so simply replacing the scheduler's configuration slices is not sufficient to safely make newly loaded actions/plugins active.

This report was prepared with AI assistance; the source paths, configuration lifecycle, and existing issue/PR surface were reviewed manually. The live-cluster reproduction should be verified before treating the behavior as conclusively reproduced.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

Fixes a bug in the Agent Scheduler where configuration changes detected by the hot-reload watcher were updated on the scheduler but were not propagated to existing worker Frameworks.

As a result, workers could continue scheduling with stale actions, tiers, and plugin configurations after a successful configuration reload. Newly added actions/configuration were not picked up, while removed actions could remain active.

This PR:

* Introduces an immutable configuration snapshot for atomic publication of scheduler configuration.
* Creates isolated action instances for reloadable configurations.
* Tracks configuration versions so workers can detect changes.
* Refreshes worker Frameworks at scheduling-cycle boundaries, ensuring a cycle always uses one consistent configuration version.
* Preserves the previous valid configuration when a reload fails.
* Adds regression and concurrent hot-reload tests.

Fixes #5967

## How has this been tested?

The following tests and checks were run successfully:

```text
go test ./pkg/agentscheduler/...
go test -race ./pkg/agentscheduler/...
go vet ./pkg/agentscheduler/...
git diff --check
gofmt -l pkg/agentscheduler
```

Additional regression coverage verifies:

* configuration changes are propagated to existing workers;
* actions can be added and removed through hot-reload;
* action/plugin configuration changes are applied;
* invalid configuration reloads preserve the previous valid configuration;
* concurrent scheduling and configuration reloads do not introduce data races.

Configuration changes are applied only at scheduling-cycle boundaries. An in-progress scheduling cycle continues using its existing Framework, while the next cycle observes the latest successfully loaded configuration.

The implementation also preserves the existing action registration path for backward compatibility while using fresh action instances for reloadable configurations.

### DIFF
--- a/pkg/agentscheduler/scheduler.go
+++ b/pkg/agentscheduler/scheduler.go
@@ -45,6 +45,13 @@ import (
 	agentmetrics "volcano.sh/volcano/pkg/agentscheduler/metrics"
 )
 
+type configSnapshot struct {
+	actions        []framework.Action
+	tiers          []conf.Tier
+	configurations []conf.Configuration
+	version        uint64
+}
+
 // Scheduler represents a "Volcano Agent Scheduler".
 // Scheduler watches for new unscheduled pods.
 // It attempts to find nodes that can accommodate these pods and writes the binding information back to the API server.
@@ -63,11 +70,34 @@ type Scheduler struct {
 	disableDefaultConf bool
 	workerCount        int
 	shardingMode       string
+	confSnapshot       *configSnapshot
+}
+
+func (sched *Scheduler) getConfSnapshot() *configSnapshot {
+	sched.mutex.Lock()
+	defer sched.mutex.Unlock()
+	return sched.confSnapshot
 }
 
 type Worker struct {
-	framework *framework.Framework
-	index     int
+	framework   *framework.Framework
+	index       int
+	sched       *Scheduler
+	confVersion uint64
+}
+
+func (worker *Worker) updateFrameworkIfNeeded() {
+	if worker.sched == nil {
+		return
+	}
+
+	snap := worker.sched.getConfSnapshot()
+	if snap == nil || snap.version == worker.confVersion {
+		return
+	}
+
+	worker.framework = framework.NewFramework(snap.actions, snap.tiers, worker.sched.cache, snap.configurations)
+	worker.confVersion = snap.version
 }
 
 // NewAgentScheduler returns a Scheduler
@@ -104,14 +134,15 @@ func (sched *Scheduler) Run(stopCh <-chan struct{}) {
 	sched.cache.SetMetricsConf(sched.metricsConf)
 	sched.cache.Run(stopCh)
 
-	for _, action := range sched.actions {
-		action.OnActionInit(sched.configurations)
-	}
-
+	snap := sched.getConfSnapshot()
 	klog.V(2).Infof("Scheduler completes Initialization and start to run %d workers", sched.workerCount)
 	for i := range sched.workerCount {
-		worker := &Worker{index: i}
-		worker.framework = framework.NewFramework(sched.actions, sched.tiers, sched.cache, sched.configurations)
+		worker := &Worker{
+			index:       i,
+			sched:       sched,
+			confVersion: snap.version,
+		}
+		worker.framework = framework.NewFramework(snap.actions, snap.tiers, sched.cache, snap.configurations)
 		go func() {
 			for {
 				select {
@@ -145,6 +176,8 @@ func (worker *Worker) runOnce() {
 		klog.Warningf("No task to schedule")
 		return
 	}
+
+	worker.updateFrameworkIfNeeded()
 
 	scheduleStartTime := time.Now()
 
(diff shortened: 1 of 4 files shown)


kingeasternsun:
Thans for your PR.

pkg/agentscheduler/scheduler.go
```diff
@@ -63,11 +70,34 @@ type Scheduler struct {
 	disableDefaultConf bool
 	workerCount        int
 	shardingMode       string
+	confSnapshot       *configSnapshot
```
kingeasternsun, line 73: 建议非强制, 使用 atomic.Pointer
```
type Scheduler struct {
	...
	confSnapshot atomic.Pointer[configSnapshot]  
}

func (sched *Scheduler) getConfSnapshot() *configSnapshot {
	return sched.confSnapshot.Load()              // 无锁，单条指令
}
```

然后有了原子指针后，worker 甚至不再需要 `confVersion` 计数器——直接比较指针身份：

```
type Worker struct {
	framework *framework.Framework
	index     int
	sched     *Scheduler
	conf      *configSnapshot   // 记录“当前 framework 是用哪份快照建的”
}

func (worker *Worker) updateFrameworkIfNeeded() {
	snap := worker.sched.getConfSnapshot()
	if snap == worker.conf {          // 指针不同 ⇒ 一定换过配置
		return
	}
	worker.framework = framework.NewFramework(snap.actions, snap.tiers, worker.sched.cache, snap.configurations)
	worker.conf = snap                // 新 framework 与新快照绑定
}

```

kingeasternsun (commented):
建议非强制。

AyushSrivastava1818:
@kingeasternsun Thanks for the suggestion..I'll switch confSnapshot to atomic.pointer and use snapshot pointer identity for worker framework refresh detection....i will update the tests accordingly and rerun the related tests

PR #5971, opened 2026-09-13
cli: return an error when jobflow and jobtemplate file ops fail

### LINKED ISSUE #5970: vcctl: jobflow and jobtemplate `-f` commands exit 0 when the API call fails
### Description

`vcctl jobflow create -f`, `jobflow delete -f`, `jobtemplate create -f` and
`jobtemplate delete -f` print the API error and then return nil. The command
exits 0 even when nothing was created or deleted, so anything chaining on the
exit status keeps going:

```bash
vcctl jobflow create -f flows.yaml && run-pipeline.sh
```

The same commands exit non-zero when you pass `-N` instead of `-f`. Only the
file branch swallows the error.

I found this by reading the code, then reproduced it with the CLI built from
master. Not seen on a live cluster.

1. Build the CLI: `go build -o vcctl ./cmd/cli`
2. Put any JobFlow in `jobflow.yaml`.
3. Point it at an address where nothing is listening, so the create fails:

```console
$ ./vcctl jobflow create -f jobflow.yaml -s http://127.0.0.1:1; echo $?
Failed to create JobFlow: Post "http://127.0.0.1:1/apis/flow.volcano.sh/v1alpha1/namespaces/default/jobflows": dial tcp 127.0.0.1:1: connectex: No connection could be made because the target machine actively refused it.
0
```

Any failing write does it. I also ran it against a server returning 403, which
is what RBAC denial looks like, with the same result. Swapping `-f` for `-N`
prints the same error and exits non-zero.

Permalinks pinned to d8984501e4ad9cd7c5a920d1dd2a28b9df48b681.

`CheckError` exits only when the error is non-nil:
https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/cmd/cli/util/util.go#L27-L38

`CreateJobFlow` prints the create error inside the loop and ends with
`return nil`, so `CheckError` always receives nil:
https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/pkg/cli/jobflow/create.go#L83-L91

Same shape in [jobflow/delete.go#L92-L100](https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/pkg/cli/jobflow/delete.go#L92-L100),
[jobtemplate/create.go#L83-L92](https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/pkg/cli/jobtemplate/create.go#L83-L92)
and [jobtemplate/delete.go#L92-L100](https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/pkg/cli/jobtemplate/delete.go#L92-L100).
Each keeps a `createdCount` or `deletedCount` that is incremented and never
read, which looks like the counts were meant to be reported.

The `-N` branch returns the error normally, which is why it exits non-zero:
https://github.com/volcano-sh/volcano/blob/d8984501e4ad9cd7c5a920d1dd2a28b9df48b681/pkg/cli/jobflow/delete.go#L112-L119

Received: the failure is printed and the command exits 0.

Expected: a non-zero exit when an object in the file could not be created or
deleted, matching the `-N` branch and the other vcctl subcommands.

v1.15.2 and master at d8984501e4ad9cd7c5a920d1dd2a28b9df48b681. Same code in
both.

CLI only, no cluster needed to hit it. Built with go1.26.0, run on Windows,
which is why the transcript shows the Windows wording for connection refused.

iamtanishqjain:
#### What type of PR is this?

/kind bug

The `-f` branch of `vcctl jobflow create`, `jobflow delete`, `jobtemplate
create` and `jobtemplate delete` printed each API error and then returned nil,
so the command exited 0 even when nothing was created. The issue has the
reproduction and the path from the command to the failure.

Each of the four now collects its failures and returns them with
`utilerrors.NewAggregate`, so `CheckError` exits non-zero. That is what the
`-N` branch of the same commands already did.

The per-item `Failed to ...` print is gone, because `CheckError` prints the
aggregate and keeping both printed the same error twice. Successes still print
as they happen.

Below, both binaries run against the same apiserver, which accepts JobFlow `a`
and rejects JobFlow `b`. The shell chains on the exit status, so before the
change the `DEPLOYED` step ran even though `ci/b` was never created.

vcctl exit status, before and after

The failing case is a new row in the existing `TestCreateJobFlow`,
`TestDeleteJobFlow`, `TestCreateJobTemplate` and `TestDeleteJobTemplate`
tables. All four fail on master and pass here. `CreateTestServer` in
`pkg/cli/util` gained a `CreateTestServerWithStatus` variant so a test can
drive a rejecting apiserver; `CreateTestServer` now calls it with 200 and
behaves as before.

`make lint` and `go test ./pkg/cli/...` are clean.

Fixes #5970

Two judgement calls, happy to change either.

`delete -f` now exits non-zero when an object is already gone, so a teardown
script that runs it twice fails on the second run. I matched `kubectl delete
-f` here, but I can skip NotFound if you would rather delete stayed idempotent.

I also removed `createdCount` and `deletedCount`. They were incremented and
never read, and they sat in the lines this change rewrites.

I used an AI tool while preparing this change, as contributing.md asks me to
disclose. I checked the affected versions, the code references and the
production path myself, and ran the tests and the linter before submitting.

```release-note
`vcctl jobflow` and `vcctl jobtemplate` now exit with a non-zero status when a create or delete driven by `-f` fails.
```


iamtanishqjain:
Small CLI fix. The `-f` form of `vcctl jobflow`/`jobtemplate` create and delete printed the API error but still exited 0, so scripts chaining on the exit status carried on after a failed run.

One question in the notes above: whether `delete -f` should keep failing on NotFound like `kubectl delete -f`, or skip it so teardown stays idempotent. Happy either way.

/cc @JesseStutler

iamtanishqjain:
@JesseStutler PTAL when you get a chance.

The short version: `vcctl jobflow create -f` and `jobtemplate create -f`, and both deletes, printed the API error and then returned nil, so the command exited 0 on failure and scripts chaining on it carried on. The `-N` form of the same commands already returned the error.

One open question in the description, whether `delete -f` should keep failing on NotFound like `kubectl delete -f`, or skip it so teardown stays idempotent. Happy either way.

hajnalmt (approved):
/ok-to-test
/area cli
/approve

Thanks let's see the CI, if it passes we can let it in.

iamtanishqjain:
Thanks @hajnalmt. CI came back green: E2E, Code Verify, CodeQL and Licenses Lint all passed. The one job that is not green is `build` from the FOSSA workflow, which skips on fork PRs because the secret is not exposed to them.

Could you drop an `/lgtm` when you get a moment? The `/approve` is already on it, so that is the last label it needs before tide can take it.

hajnalmt (approved):
/lgtm

Thanks!

PR #5978, opened 2026-09-15
fix: guard PodGroup deletion against stale scale-to-zero events

### LINKED ISSUE #5972: :bug: StatefulSet PodGroup Can Be Incorrectly Deleted During Rapid Scale-to-Zero/Scale-Up
### Description

In Volcano 1.14, when a `StatefulSet` is scheduled by Volcano, the `pg-controller` automatically creates a `PodGroup` for its Pods. The PodGroup name is derived from the StatefulSet UID.

A race condition can occur when a StatefulSet is rapidly scaled from `replicas=1` to `replicas=0` and then back to `replicas=1`.

In this scenario, an outdated `StatefulSet` update event with `replicas=0` can be processed after the new Pod and its PodGroup have already been created. The outdated event then deletes the newly created PodGroup.

The Pod still retains the `scheduling.k8s.io/group-name` annotation. When the StatefulSet is subsequently observed with `replicas=1`, Volcano detects that the Pod already has a `group-name` annotation and skips PodGroup creation. As a result, the Pod references a PodGroup that no longer exists.

This can leave the Pod stuck in an abnormal scheduling state until the Pod is manually deleted and recreated.

1. StatefulSet is running with replicas=1.

2. StatefulSet is scaled to replicas=0.

3. Before the old replicas=0 update event is fully processed,
   the StatefulSet is quickly scaled back to replicas=1.

4. A new Pod is created.

5. Volcano receives the Pod add event and creates the PodGroup.

6. Volcano patches the Pod with:
   scheduling.k8s.io/group-name: podgroup-<statefulset-uid>

7. The previously queued replicas=0 StatefulSet update event is processed.

8. pg-controller sees replicas=0 and deletes:
   podgroup-<statefulset-uid>

9. The StatefulSet replicas=1 event is processed.

10. Volcano sees that the Pod already has a group-name annotation
    and skips PodGroup creation.

11. The Pod references a PodGroup that does not exist.

Image

We observed this behavior in a production environment on September 7, 2026.

The relevant operations occurred within approximately 20 ms:

15:27:13.766  PodGroup creation succeeded
15:27:13.776  Pod annotation was updated with group-name
15:27:13.781  PodGroup deletion returned HTTP 200
15:27:13.783  Another PodGroup deletion returned HTTP 404
15:27:13.784  StatefulSet reconciliation found an existing group-name
              annotation and skipped PodGroup creation

This sequence is consistent with an outdated replicas=0 StatefulSet event being processed after the new PodGroup had already been created.

The PodGroup deletion returning 200 is particularly important: it indicates that the deletion actually removed the newly created PodGroup rather than merely attempting to delete an already missing object.

The workload recovered after the affected Pod was manually deleted. The StatefulSet then created a new Pod, Volcano recreated the PodGroup, and the workload became Ready.

When a StatefulSet is scaled back to `replicas > 0`, the PodGroup required by its Pods should remain available or be automatically recreated.

An outdated `replicas=0` event should not be able to delete a PodGroup that belongs to the current StatefulSet state.

1.14

Proposed Fix

Re-check the current StatefulSet before deleting the PodGroup

Before deleting the PodGroup for a replicas=0 event, retrieve the current StatefulSet and verify that it is still scaled to zero.

For example:
```go
if *sts.Spec.Replicas == 0 {
    latest, err := pg.kubeClient.AppsV1().StatefulSets(sts.Namespace).Get(
        context.TODO(), sts.Name, metav1.GetOptions{})
    if err == nil && latest.Spec.Replicas != nil && *latest.Spec.Replicas > 0 {
        return
    }

    pgName := batchv1alpha1.PodgroupNamePrefix + string(sts.UID)
    err = pg.vcClient.SchedulingV1beta1().PodGroups(sts.Namespace).Delete(
        context.TODO(), pgName, metav1.DeleteOptions{})
}
```

This prevents an outdated replicas=0 event from deleting a PodGroup after the StatefulSet has already been scaled back up.

An implementation using the latest object from the informer lister could also be considered if consistency requirements allow it.

shashankvarma499:
#### What type of PR is this?

Bug fix

A stale `replicas=0` StatefulSet or ReplicaSet event queued behind a rapid
scale-up deletes the PodGroup that was just recreated for the current
replicas. The Pod still carries the `scheduling.k8s.io/group-name`
annotation, so the later `replicas > 0` reconciliation skips recreating the
PodGroup, leaving the Pod referencing a PodGroup that no longer exists.

This PR re-checks the current workload before deleting the PodGroup. It
consults the informer lister first, and when that is inconclusive it queries
the API server directly, deleting only when the workload is still scaled to
zero. The same guard is applied to both the StatefulSet and ReplicaSet paths.

Fixes #5972

The podgroup controller enqueues Pod events through its workqueue, but the
StatefulSet/ReplicaSet handlers act inline on add/update events, which is the
existing architecture for these paths. I kept this change scoped to the race
fix rather than migrating those handlers onto the workqueue. I am happy to
look at that migration separately if you would like.

```release-note
Fixed a race where a stale scale-to-zero event could delete a StatefulSet or
ReplicaSet PodGroup after the workload had already been scaled back up,
leaving its Pod stuck referencing a missing PodGroup.
```


kingeasternsun:
@shashankvarma499 Thanks for your PR, and I agree with your fix method, but the `E2E Tests` needs to be fixed firstly.

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -80,6 +80,14 @@ func (pg *pgcontroller) addReplicaSet(obj interface{}) {
 
 	if *rs.Spec.Replicas == 0 {
 		pgName := batchv1alpha1.PodgroupNamePrefix + string(rs.UID)
+		// Re-check the current ReplicaSet before deleting: a stale
+		// replicas=0 event queued behind a rapid scale-up must not delete
+		// the PodGroup that was just recreated for the current replicas.
+		if !pg.replicaSetStillScaledToZero(rs) {
```
JesseStutler, line 86: Better return error outside, otherwise inside the `replicaSetStillScaledToZero` met error but not related with scale down, the klog still log `no longer scaled to zero` will confuse

JesseStutler:
The 2 CI failures are not related with this pr, I have rerun it, if the spark still failed, please rebase to the latest code, and please also help review it @kingeasternsun

pkg/controllers/podgroup/pg_controller_handler.go
```diff
@@ -186,6 +202,64 @@ func (pg *pgcontroller) updateStatefulSet(oldObj, newObj interface{}) {
 	pg.addStatefulSet(newObj)
 }
 
+// replicaSetStillScaledToZero reports whether the ReplicaSet is still scaled
+// to zero, re-checking the current object rather than trusting the event that
+// triggered this handler. A stale replicas=0 event (queued behind a rapid
+// scale-up) would otherwise delete a PodGroup that has just been recreated.
+func (pg *pgcontroller) replicaSetStillScaledToZero(rs *appsv1.ReplicaSet) bool {
```
hajnalmt, line 209: We should return errs in these functions Jesse is right above.

```
func (pg *pgcontroller) replicaSetStillScaledToZero(rs *appsv1.ReplicaSet) (bool, error)...
.....
// The informer cache can lag the API server. Verify directly.
	latest, err := pg.kubeClient.AppsV1().ReplicaSets(rs.Namespace).Get(context.TODO(), rs.Name, metav1.GetOptions{})
	if err != nil {
		// If the ReplicaSet no longer exists, its PodGroup is orphaned and
		// safe to delete (it is otherwise GC'd via the Pod owner reference).
		if apierrors.IsNotFound(err) {
			return true, nil
		}
		// Surface the error so the caller can log it correctly instead of
		// mislabeling it as a scale-up.
		return false, err
	}
		// Surface the error so the caller can log it correctly instead of
		// mislabeling it as a scale-up.
		return false, err
}
return latest.Spec.Replicas != nil && *latest.Spec.Replicas == 0, nil
 ```

File: pkg/controllers/podgroup/pg_controller_handler.go [hunk 3]
```diff
@@ -186,6 +202,64 @@ func (pg *pgcontroller) updateStatefulSet(oldObj, newObj interface{}) {
 	pg.addStatefulSet(newObj)
 }
 
+// replicaSetStillScaledToZero reports whether the ReplicaSet is still scaled
```
hajnalmt, line 205: nit: The comment is just unnecessary, because you explain it on the call side too. I would keep it there only.

hajnalmt (commented):
/hold

Thanks for fixing this! Please fix Jesse's comment and rebase so the CI can pass. Also please mention the AI usage in the PR description too 😊

JesseStutler:
@shashankvarma499 Hi, we're going to release v1.16 version soon, please take a look

PR #5993, opened 2026-09-16
fix: ignore stale Pod events in JobInfo cache

### LINKED ISSUE #5986: Stale Pod informer events can corrupt JobInfo Pod cache after Pod recreation
### Description

Volcano can apply a stale Pod informer event to the cached Pod state when an older Pod is replaced by a newer Pod with the same name.

`JobInfo.Pods` is indexed by Pod name. If a Pod is deleted and another Pod with the same name is created, the two Kubernetes objects have different UIDs. A delayed informer event for the old Pod can therefore be interpreted as an event for the current Pod.

This can cause the controller cache to become inconsistent with the actual Pod state.

The stale-event behavior was observed during a real Minikube cluster workload where a Volcano Job was repeatedly deleted and recreated with the same name.

1. Install Volcano v1.15.2 on a Minikube cluster.
2. Create a Volcano Job with a single task and Pod.
3. Repeatedly delete and recreate the same Volcano Job name with a short delay between operations.
4. Allow the previous Pod to generate informer events while the replacement Job/Pod is already present.
5. Inspect the Volcano controller logs and Job/Pod cache behavior.

A minimal workload used for the reproduction was a Volcano Job named `volcano-stress-test` with one worker Pod.

The Job was repeatedly created, allowed to start, deleted, and recreated with the same name.

This issue was reproduced through an actual Minikube cluster workload, not only through a unit test.

Environment:

* Volcano v1.15.2
* Kubernetes v1.35.1
* Minikube
* Docker runtime
* 1 control-plane node with approximately 4 allocatable CPUs

The workload repeatedly deleted and recreated a Volcano Job with the same name:

`default/volcano-stress-test`

During the rapid replacement sequence, Volcano received events belonging to the previous Job after a new Job with the same name was already present in the cache.

The controller logged:

```text
The req belongs to job(default/volcano-stress-test) and job uid is <old-uid>, but the uid of job(default/volcano-stress-test) is <new-uid> in cache, perform SyncJob action
```

The same test also produced stale Pod-related requests after the replacement Job was already cached.

The relevant execution path is:

```text
Kubernetes Pod informer event
        ↓
job_controller_handler.go
        ↓
jobCache.UpdatePod() / jobCache.DeletePod()
        ↓
JobInfo.UpdatePod() / JobInfo.DeletePod()
        ↓
JobInfo.Pods[taskName][podName]
        ↓
Pod is identified by name, without verifying Pod UID
```

The request-level controller subsequently contains a Job UID check in `applyPolicies()`. However, that check happens after the Pod informer handler has already updated the cache. Therefore, the later Job UID check does not prevent the stale Pod event from mutating `JobInfo.Pods`.

The cache-level behavior was also reproduced deterministically with two Kubernetes Pod objects having the same namespace/name but different UIDs. Before the fix:

* a stale DELETE event removed the current Pod from the cache;
* a stale UPDATE event replaced the current Pod with the stale Pod.

Regression tests were added for both cases:

* `TestJobCache_DeletePodIgnoresStalePodUID`
* `TestJobCache_UpdatePodIgnoresStalePodUID`

These tests demonstrate the exact cache corruption once the stale informer event reaches `JobInfo`, while the Minikube workload demonstrates that stale same-name object events can occur through the actual controller/informer path.

The expected cache invariant is that an informer event for Pod UID A must not modify or delete the cached Pod object when the cached object has Pod UID B.

**Actual result:**

When an older Pod and a newer Pod have the same name but different UIDs, a stale event for the older Pod can operate on the cache entry belonging to the newer Pod.

A stale DELETE can remove the current Pod from `JobInfo.Pods`.

A stale UPDATE can replace the current Pod object and state in `JobInfo.Pods`.

**Expected result:**

The cached Pod should only be modified or deleted when the UID of the informer event matches the UID of the currently cached Pod.

An event belonging to an older Pod with the same name should be ignored.

v1.15.2

Kubernetes: v1.35.1

Runtime: Docker

Environment: Minikube on Ubuntu

The local reproduction used rapid same-name Volcano Job deletion and recreation.

The stale Job UID was already detected by the controller's request-processing logic, but the Pod cache mutation occurs earlier in the informer handling path.

The proposed change is limited to the Pod cache mutation layer: before updating or deleting a cached Pod, compare the incoming Pod UID with the cached Pod UID and ignore the event when they differ.

Validation after the change:

```text
go test ./pkg/controllers/cache ./pkg/controllers/apis ./pkg/controllers/job -count=1
```

Faisal77666:
## What type of PR is this?

/kind bug

## What does this PR do / why do we need it?

This PR prevents stale Pod informer events from corrupting the `JobInfo` Pod cache when an older Pod is replaced by a newer Pod with the same name.

`JobInfo.Pods` is indexed by Pod name, while Kubernetes Pod objects are uniquely identified by their UID. When a Pod is deleted and a new Pod is created with the same name, the two Pod objects have different UIDs. A delayed informer event for the older Pod can therefore be interpreted as an event for the current Pod unless the Pod UID is checked.

Previously:

* A stale **DELETE** event could remove the current Pod from `JobInfo.Pods`.
* A stale **UPDATE** event could replace the current Pod with the stale Pod object.

This can leave the Volcano controller's cached Pod state inconsistent with the actual cluster state.

The fix compares the incoming Pod UID with the UID of the currently cached Pod and ignores the event when the UIDs differ.

### 1. Cluster reproduction

The issue was first observed during a real Volcano workload running on a local Minikube cluster.

The same Volcano Job was repeatedly deleted and recreated with the same name:

```text
default/volcano-stress-test
```

Each recreation received a new Kubernetes Job UID.

During the reproduction, Volcano received requests belonging to an older Job UID while a newer Job with the same name was already present in the cache.

**Reproduction environment:**

* Minikube: `v1.38.1`
* Kubernetes server: `v1.35.1`
* Volcano: `v1.15.2`
* Volcano installed using Helm
* Volcano controller, scheduler, and admission components were running

The controller logs showed the stale UID condition:

```text
The req belongs to job(default/volcano-stress-test) and job uid is 9f23c6e7-be44-4821-bad7-2140b81c2f63, but the uid of job(default/volcano-stress-test) is 36acddbe-e591-4126-b7e9-4e5e0207d5aa in cache, perform SyncJob action
```

Screenshot From 2026-09-21 14-26-30

The same stale-UID condition was observed more than once during the workload.

### 2. Regression reproduction

After identifying the stale-object behavior from the Minikube reproduction, regression tests were added for stale Pod UPDATE and DELETE events.

Before the fix, both tests failed:

```text
--- FAIL: TestJobCache_DeletePodIgnoresStalePodUID
    cache_test.go:640: current pod was incorrectly removed by stale delete event

--- FAIL: TestJobCache_UpdatePodIgnoresStalePodUID
    cache_test.go:701: stale update replaced current pod: expected UID "new-pod-uid", got "old-pod-uid"
FAIL
```

Screenshot From 2026-09-21 14-14-34

## Root Cause

The Pod cache is organized by Job name, task name, and Pod name. Pod UID was not previously checked when processing Pod UPDATE and DELETE events.

When an old Pod and a replacement Pod have the same name:

1. The old Pod is deleted.
2. A new Pod with the same name is created and receives a new UID.
3. The new Pod is stored in `JobInfo.Pods`.
4. A delayed informer event for the old Pod arrives.
5. The cache previously treated the event as belonging to the cached Pod because the Pod name matched.

As a result, a stale DELETE could remove the current Pod, while a stale UPDATE could replace the current Pod object.

## Implementation

This PR adds Pod UID validation to the cache layer:

* Added UID validation in `JobInfo.UpdatePod()`.
* Added UID validation in `JobInfo.DeletePod()`.
* Stale events are ignored when the incoming Pod UID differs from the cached Pod UID.
* Events for the currently cached Pod continue to behave as before.
* Added regression tests for both stale UPDATE and stale DELETE events.

The change is intentionally limited to the `JobInfo` Pod-cache update and delete paths.

## Which issue(s) does this PR fix?

Fixes #5986

## Validation

Targeted tests:

```bash
go test ./pkg/controllers/cache ./pkg/controllers/apis ./pkg/controllers/job -count=1
```

No direct API or configuration change is introduced.

This change prevents stale informer events from modifying the cached state of a newer Pod while preserving the existing behavior for matching Pod UIDs.


Faisal77666:
@JesseStutler @hajnalmt  PTAL

PR #5996, opened 2026-09-17
fix(scheduler): latch PodGroup Completed phase to prevent quota re-reservation after Pods deleted

### LINKED ISSUE #4195: Is it possible to block podgroups from transitioning from "Completed" to "Inqueue"?

Is it possible to prevent a PodGroup from transitioning back to the "InQueue" state once it has reached the "Completed" state?

## Use Case
- I have a workflow manager responsible for creating the PodGroup
- This workflow manager creates the first pod associated with the PodGroup, which is the primary controller that manages tasks
- This "master" pod launches several other pods associated with the PodGroup
- The success/failure of the entire job is determined solely by this first "master" pod
- Ideally, I would set the ownership of the PodGroup to the first pod, but there's a circular dependency (need the PodGroup before creating the "master" pod)
- The main issue occurs when the first pod is cleaned up after completion while the associated PodGroup remains
- This orphaned PodGroup then transitions back to "InQueue" state and continues holding queue resources
Currently, we delete the PodGroup after completion, but this approach isn't robust enough

I'm sorry if this was asked before or is in the documentation. Tried to find some info, with no success. Thanks!

_No response_

yuanhuaiwang:
## Bug Description

A PodGroup that reached the `Completed` phase is downgraded back to `Pending` and then re-admitted to `Inqueue` once its member Pods are deleted (by TTL, GC, or the workload controller). The resurrected PodGroup reserves its `MinResources` queue quota again through the capacity/proportion plugins, blocking other pending jobs — even though it can never run again.

Fixes #4195 (see also the root-cause analysis in [#4195 comment](https://github.com/volcano-sh/volcano/issues/4195#issuecomment-5711339934))

## Root Cause

`getPodGroupPhase()` recomputes the phase **from scratch every session** based solely on `TaskStatusIndex`:

1. While member Pods exist, the phase correctly reaches `Completed` (`scheduled >= MinMember && scheduled == completed`).
2. When the finished Pods are deleted, `deletePod → deleteTask` removes the tasks from `JobInfo`. The job stays in the scheduler cache because `JobTerminated()` requires `PodGroup == nil` (the PodGroup still exists).
3. Next session, `getPodGroupPhase()` recomputes with 0 tasks: `scheduled(0) < MinMember` → falls through to `if phase != Inqueue → Pending` — **`Completed` is downgraded to `Pending`**.
4. The following session, the enqueue action sees `job.IsPending()` → re-enqueues → sets `phase = Inqueue`.
5. The capacity/proportion plugins then count it again: `phase == Inqueue → attr.inqueue.Add(minResources)` — the finished job's `MinResources` reserves queue quota again.

There is no terminal-state latch, so `Completed` survives at most 1–2 sessions after the Pods are gone.

## Fix

Latch the terminal phase in `getPodGroupPhase()`: when the current phase is `Completed` and all member tasks are gone, return `Completed` instead of recomputing. This is strictly additive — it only affects the state where every task has been removed (Pods can never come back), so live-scheduling behavior is unchanged.

Reproduced with the exact sequence from #4195 (PodGroup + Pod with `restartPolicy: Never`, Pod completes → `Completed`; delete the Pod → `Inqueue`).

## How to Verify

1. Create a PodGroup and a member Pod as in #4195; wait for `Completed`.
2. Delete the Pod; before this fix the PodGroup flips `Completed → Pending → Inqueue` within ~2 sessions and reserves queue quota again; with this patch it stays `Completed` and no longer counts towards `inqueue`.
3. `go test ./pkg/scheduler/framework/ -run TestGetPodGroupPhase -count=1`

## Test Plan

- [x] Added regression cases to `TestGetPodGroupPhase`: completed group with no tasks left stays `Completed`; completed group with a remaining running task recomputes to `Running`; completed group with a new pending task recomputes to `Pending`
- [x] Existing tests pass: `pkg/scheduler/framework`, `pkg/scheduler/cache`, all `pkg/scheduler/actions/...`, all `pkg/scheduler/plugins/...` (the only failure, `TestCapacityPluginDRA`, is a pre-existing flake reproducible on unpatched master)
- [x] `go build ./pkg/scheduler/...` and `go vet ./pkg/scheduler/framework/` clean

## Risk Assessment

Low — the latch only triggers when the PodGroup is `Completed` **and** `len(jobInfo.Tasks) == 0`, a state that today immediately downgrades to `Pending` and re-enters the scheduling queue. No path where a job still has member tasks is affected, and the recompute logic for all other states is untouched.


yuanhuaiwang:
@kingeasternsun please take a look

pkg/scheduler/framework/session.go
```diff
@@ -590,6 +590,17 @@ func closeSession(ssn *Session) {
...
 
 func getPodGroupPhase(jobInfo *api.JobInfo, unschedulable bool) scheduling.PodGroupPhase {
+	// Latch the terminal phase: once a PodGroup reached Completed and all its
+	// member tasks have been removed afterwards (e.g. TTL/GC deleted the
+	// finished Pods), there is nothing left to recompute from. Without this
+	// latch the phase is downgraded back to Pending/Inqueue on the next
+	// session, so the enqueue action re-admits the dead group and the
+	// capacity/proportion plugins reserve its MinResources queue quota again.
+	if jobInfo.PodGroup.Status.Phase == scheduling.PodGroupCompleted &&
+		len(jobInfo.Tasks) == 0 {
```
JesseStutler, line 600: This judgement can still not prevent podgroup rollback from completed to pending, since the speed of pod deletion is uncertain It is highly likely to span multiple sessions, consequently, some pods might remain in the "Completed" state without being fully deleted.

So I suggest we can do like:
```suggestion
	if jobInfo.PodGroup.Status.Phase == scheduling.PodGroupCompleted {
		allCompleted := true
		for _, task := range jobInfo.Tasks {
			if !api.CompletedStatus(task.Status) {
				allCompleted = false
				break
			}
		}
		if allCompleted {
			return scheduling.PodGroupCompleted
		}
 	}
```

yuanhuaiwang: Good catch, agreed. Applied in 6d0575e: the latch now keeps `Completed` whenever every remaining task is in a terminal state (`api.CompletedStatus`), so a partially deleted group no longer rolls back to Pending/Inqueue mid-deletion. The empty-task case is covered since the loop passes vacuously. Regression tests added for the partially deleted (remaining Succeeded task below minMember) and remaining Failed task cases; `go test ./pkg/scheduler/framework/ ./pkg/scheduler/cache/ ./pkg/scheduler/actions/... ./pkg/scheduler/plugins/...` all pass.

JesseStutler:
@yuanhuaiwang Hi thanks for your contribution, but I have to notice you that should follow our AI guidance: https://github.com/volcano-sh/volcano/blob/master/contributing.md#ai-guidance, please do not use agent to directly discuss with reviewers

pkg/scheduler/framework/session.go
kingeasternsun: @JesseStutler @yuanhuaiwang 
Hi, I have some  questions:

When `jobInfo.PodGroup.Status.Phase == scheduling.PodGroupCompleted`, why not just return `scheduling.PodGroupCompleted` directly, instead of additionally checking whether there are any non-completed tasks remaining?

Does this PR handle the following scenario correctly?

```
 minMember=2, 5 replicas, 2 Succeeded + 3 Pending → scheduled=2, completed=2 →  Completed
 ```
 This group is a "legitimate Completed", but it still has 3 Pending pods attached.
 
 Once the TTL controller deletes the terminal **Succeeded/Failed** pods , the PodGroup is left with only 3 Pending pods. On the next scheduling session, `getPodGroupPhase` runs again and falls through to the normal recompute path: `scheduled = 0` (none of the 3 Pending tasks are counted) `< minMember = 2`, so it reaches the final check where `phase != Inqueue` and returns `PodGroupPending`.

kingeasternsun:
Another concern I have: if a StatefulSet's replicas scale down from 1 to 0 and then back up to 1 very quickly, the new pod will still inherit the old PodGroup from before — could this cause any issues?
@yuanhuaiwang @JesseStutler

JesseStutler:
> Another concern I have: if a StatefulSet's replicas scale down from 1 to 0 and then back up to 1 very quickly, the new pod will still inherit the old PodGroup from before — could this cause any issues? @yuanhuaiwang @JesseStutler

https://github.com/volcano-sh/volcano/pull/5344 Perhaps this PR can meet your requirement, and I have already cherry-picked it to v1.14 and v1.15, will release patch version later

yuanhuaiwang:
> @yuanhuaiwang Hi thanks for your contribution, but I have to notice you that should follow our AI guidance: https://github.com/volcano-sh/volcano/blob/master/contributing.md#ai-guidance, please do not use agent to directly discuss with reviewers

OK, thanks a lot !

JesseStutler:
/approve
Please /cc @kingeasternsun

PR #5997, opened 2026-09-17
fix: retry pods with unsynced ResourceClaims

gitGurugu:
A Pod ADD can reach the scheduler before its directly referenced ResourceClaim has been delivered to the ResourceClaim informer. The cache currently treats that miss as permanent, drops the task, and later cannot resync it.

Treat AssumeCache NotFound and nil results as transient for direct ResourceClaim references, matching the existing ResourceClaimTemplate path. The task stays in the scheduler cache and is retried after informer convergence; other cache errors still fail normally.

Validation:
- `go test ./pkg/scheduler/cache ./pkg/scheduler/plugins/capacity`
- `go test -race ./pkg/scheduler/cache -run 'TestAddPodWithUnresolvedResourceClaim(Template)?CachesTaskForResync' -count=5 -timeout=5m`
- `go test -race ./pkg/scheduler/plugins/capacity -run '^TestCapacityPluginDRA$/Case_1' -count=1 -timeout=5m`


JesseStutler:
Please provide the issue detail and reproduce steps, is it AI scanned or real bugs?

PR #5999, opened 2026-09-17
feat(scheduler): add checkpoint-aware task victim ordering

### LINKED ISSUE #5998: [Feature] Checkpoint-aware task victim ordering for preemption and reclamation

When `preempt`/`reclaim` select victims among tasks of equal priority, task victim ordering falls back to `CompareTask` (pod creation time). For long-running batch/ML workloads this ignores the one thing that actually determines the cost of eviction: **how much progress is lost**.

A task that just checkpointed can be evicted and resumed with almost no lost compute. A task that has never checkpointed loses all of its work. Today the scheduler cannot distinguish them, so preemption may evict the most expensive victim while sparing a cheap one, wasting cluster compute even when the eviction count is "minimal".

Creation time is only a weak proxy for this; two equal-priority tasks created together can have very different checkpoint progress.

Let the victim path optionally order tasks by **checkpoint recency**, so the least-progress-loss victim is preferred:

- Introduce a task victim-ordering extension point used **only** on the eviction path (`preempt`/`reclaim`), leaving `allocate`/`backfill` ordering unchanged.
- Add an opt-in `checkpoint` ordering plugin that reads a task's last-checkpoint time from a pod annotation (e.g. `volcano.sh/last-checkpoint-time`), with no new core API.
- Ordering keys (highest to lowest precedence): `victim queue -> priority -> checkpoint recency -> creation time -> UID`. Priority stays dominant; checkpoint only breaks ties among equal-priority tasks.
- **Most-recently-checkpointed is evicted first**; a task with no checkpoint is treated as maximum-cost and evicted **last**.
- Default behavior is unchanged: with the plugin disabled, task victim ordering is identical to today (`priority -> creation time`).

The workload supplies the checkpoint signal via annotation; the scheduler stays application-agnostic.

### Additional context

This reuses Volcano's existing tiered plugin ordering, so operators control the key's position via plugin order in config. It adds progress-loss as an ordering signal without changing how victims are otherwise selected.

mohitvinchoo:
#### What type of PR is this?

/kind feature

When the `preempt`/`reclaim` actions select victims among tasks of equal priority, ordering falls back to pod creation time. For long-running batch/ML workloads this ignores the factor that actually determines the cost of an eviction: how much progress is lost. A task that just checkpointed can be evicted and resumed with almost no lost compute, while a task that never checkpointed loses all of its work.

This PR lets the victim path optionally order tasks by checkpoint recency:

- Adds a `VictimOrderFn` plugin extension point consumed **only** on the eviction path (`BuildVictimsPriorityQueue`, used by `preempt`/`reclaim`). `allocate` and `backfill` continue to use `TaskOrderFn` and are unaffected. `VictimOrderFn` walks the enabled victim-order plugins and falls back to `TaskOrderFn`, so with no victim-order plugin enabled the eviction ordering is identical to the previous `!TaskOrderFn` behavior — a no-op by default.
- The `priority` plugin registers the same comparator on `VictimOrderFn`, so priority stays dominant on the victim path.
- Adds an opt-in `checkpoint` plugin that reads a task's last-checkpoint time from the pod annotation `volcano.sh/last-checkpoint-time` (RFC3339, UTC; key configurable via `checkpointTimeKey`). The most-recently-checkpointed task is evicted **first** (least work lost); a task with a missing or unparseable annotation is treated as never checkpointed and evicted **last**.

With `priority` listed before `checkpoint`, the victim order is:
`victim queue -> priority -> checkpoint recency -> creation time -> UID`. Priority stays dominant; checkpoint only breaks ties among equal-priority tasks.

An annotation (not a label) is used deliberately: the value changes on every checkpoint and is never used for object selection. Labels are indexed by the API server for querying, so mutating them frequently causes unnecessary etcd write overhead; annotations are the correct carrier for a frequently-updated, non-selectable value. The workload (or a sidecar) updates the annotation after each successful checkpoint, so the scheduler stays application-agnostic.

Design doc: `docs/design/checkpoint-victim-order.md`.

Fixes #5998

- Backward compatible: `enableVictimOrder` defaults to `true` (like `enableTaskOrder`), and `VictimOrderFn` falls back to `TaskOrderFn`, so existing configs behave exactly as before until the `checkpoint` plugin is enabled.
- `VictimOrderFn` falls back to `TaskOrderFn` whenever `VictimCompareFns(l, r)` returns 0 for that specific pair (a per-comparison decision), which happens in three situations:
  1. No `victimOrderFn` is registered at all — no plugin called `AddVictimOrderFn`.
  2. All plugins that have one are disabled — every `victimOrderFn` plugin has `enableVictimOrder: false`.
  3. All enabled `victimOrderFn`s return 0 (tie/abstain) for this pair — e.g. equal priority and equal/both-missing checkpoint. This is the one that's easy to miss, and it is what lets `creation-time` act as the final tie-break through the `TaskOrderFn` fallback.
- Only `BuildVictimsPriorityQueue` (preempt + reclaim) switches from `TaskOrderFn` to `VictimOrderFn`; the preemptor-ordering queues are unchanged.
- Unit tests cover: checkpoint recency evicts most-recent first, never-checkpointed evicted last, invalid annotation treated as never checkpointed, and priority dominating checkpoint.
- AI disclosure: this PR was prepared with AI assistance in accordance with the project's AI guidance.

### Validation on a local kind cluster

Scheduler config enabled `preempt`/`reclaim` with plugins `priority` then `checkpoint`. To rule out any accidental agreement between checkpoint recency and pod creation order, **checkpoint times were assigned in the inverse order of pod creation**: the earliest-created filler has the newest checkpoint. Legacy ordering evicts the newest-created task; checkpoint ordering evicts the most-recently-checkpointed one — so the two rules predict opposite victims.

**Most-recently-checkpointed task is evicted first, and disabling the plugin restores legacy ordering (preempt).** Five equal-priority, preemptable fillers in one PodGroup; one higher-priority preemptor needing exactly one slot. The same inputs were run with the `checkpoint` plugin enabled and disabled.

| filler | created  | checkpoint     | checkpoint ON | checkpoint OFF |
|--------|----------|----------------|---------------|----------------|
| p1     | oldest   | 00:05 (newest) | **evicted**   | retained       |
| p2–p4  | middle   | 00:04–00:02    | retained      | retained       |
| p5     | newest   | 00:01 (oldest) | retained      | **evicted**    |

Same setup, opposite victim: ON evicts the most-recently-checkpointed (p1); OFF falls back to legacy creation-time ordering and evicts the newest-created (p5). Exactly one eviction in each case — confirming both the checkpoint ordering and that the plugin is a no-op when disabled.

**A never-checkpointed task is evicted last (preempt).** Two equal-priority fillers; preemptor needs one slot.

| pod         | created | checkpoint | evicted?    |
|-------------|---------|------------|-------------|
| filler-ckpt | oldest  | 00:05      | **evicted** |
| filler-none | newest  | (none)     | retained    |

Legacy would evict `filler-none` (newest-created); checkpoint protects it (zero sentinel = evict last) and evicts the checkpointed one instead.

**The same ordering applies on the reclaim path.** An over-share queue runs five fillers with inverted checkpoints; another queue submits a reclaimer. Reclaim evicted `qa-p1` (checkpoint 00:05, earliest created):

```
checkpoint.go:103] Checkpoint VictimOrder: <default/qa-p1> checkpoint 2025-01-01 00:05:00 +0000 UTC, <default/qa-p2> checkpoint 2025-01-01 00:04:00 +0000 UTC
cache.go:269] Evicting pod default/qa-p1, because of reclaim
```

Priority dominance over checkpoint is covered by the unit test `TestPriorityDominatesCheckpoint`.

```release-note
Added an opt-in `checkpoint` scheduler plugin and a `VictimOrderFn` extension point that let `preempt`/`reclaim` order victims by checkpoint recency (via the `volcano.sh/last-checkpoint-time` pod annotation), evicting the most-recently-checkpointed task first to minimize lost compute. Default behavior is unchanged.
```


pkg/scheduler/plugins/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,118 @@
...
+
+// DefaultCheckpointTimeKey is the pod annotation key the workload updates after
+// every successful checkpoint. The value must be an RFC3339 timestamp (UTC).
+//
+// Annotations, not labels, are used deliberately: the value changes on every
+// checkpoint and is never used for object selection. Labels are indexed by the
+// API server for querying, so mutating them frequently adds unnecessary etcd
+// write overhead; annotations are the correct carrier for a frequently-updated,
+// non-selectable value.
+const DefaultCheckpointTimeKey = "volcano.sh/last-checkpoint-time"
```
JesseStutler, line 39: I don't see any controller updated this custom annotation onto pod, it means users can't directly use this plugin if they don't have such checkpoint aware controller. And I’d like to know roughly when you update this annotation. Also, it feels overly customized to me, is there an upstream standard for checkpoints?

pkg/scheduler/plugins/checkpoint/checkpoint.go
```diff
@@ -0,0 +1,118 @@
...
+func (cp *checkpointPlugin) OnSessionOpen(ssn *framework.Session) {
+	// victimOrderFn orders victim tasks by checkpoint recency. The returned int
+	// follows the keep-order convention shared by task/victim order plugins:
+	// a negative value means l should be preserved (evicted later).
+	//
+	// A more recent checkpoint means less work is lost on eviction, so such a
+	// task should be evicted first (kept last) -> return positive. Tasks with an
+	// older (or zero/never) checkpoint are preserved -> return negative. Equal
+	// checkpoint times defer to the next victim-order key.
+	victimOrderFn := func(l, r interface{}) int {
```
JesseStutler, line 96: I’d like to know what kind of higher-priority tasks cause these training tasks to be preempted? Do these training tasks perform periodic checkpointing and annotation updates? My concern is that, even with a preemption order in place, the training progress made between the last checkpoint and the moment of preemption would still be lost; ideally, I would like the system to complete a checkpoint before the preemption actually occurs.

mohitvinchoo: > I’d like to know what kind of higher-priority tasks cause these training tasks to be preempted?

Usually it comes from consolidating workloads to keep utilization high. Bigger shared clusters tend to run a wider variety of jobs, and training jobs themselves can have different priority levels, so it is possible for a high priority training job to preempt a low priority one. An even more common case is cross queue reclaim, where one queue takes back capacity another queue was borrowing.

> Do these training tasks perform periodic checkpointing and annotation updates? 

Yes, periodic checkpointing is very common for these jobs. Since companies usually already have their own internal storage, they tend to build their own checkpointing solution around it, and updating an annotation on each save fits naturally into that.

> My concern is that, even with a preemption order in place, the training progress made between the last checkpoint and the moment of preemption would still be lost; ideally, I would like the system to complete a checkpoint before the preemption actually occurs.

You are 100% right. Completing a checkpoint before the eviction happens is a stronger approach and I would like to see it too. It is just a good bit heavier, because the application has to:

- handle a termination signal, for example through a preStop hook
- reach the next safe point (at least finishing the current iteration or step)
- finish writing the checkpoint within the grace period, which can be tight for large models

pkg/scheduler/plugins/checkpoint/checkpoint.go
mohitvinchoo: > I don't see any controller updated

Right, Volcano is only managing the workload here. There is no reliable way for the scheduler itself to know when a workload has saved its progress, so it intentionally stays out of that and just consumes a signal.

> if they don't have such checkpoint aware controller.

Correct. In most of the industry, checkpoints are written to an internal artifact store or object storage after meaningful progress is made, so if the workload gets disrupted it can resume from the last saved checkpoint instead of starting over.

> And I’d like to know roughly when you update this annotation.

Ideally a sidecar, a training operator, or anything that talks to the API server updates the annotation as soon as a checkpoint is saved.

> Also, it feels overly customized to me, is there an upstream standard for checkpoints?

As far as I know there is no widely adopted upstream or OSS standard for application level checkpoint tracking. Almost every company rolls their own.
**That is also the opportunity here: if Volcano exposes a small, generic hook for it, it can be one of the first OSS schedulers smart enough to recognize that the most recently checkpointed workloads are the cheapest to preempt, since they lose the least progress.**

mohitvinchoo:
Discussed offline!

PR #6007, opened 2026-09-21
fix(fairshare): gate vcjobs by PodGroup MinResources when no pods exist

### LINKED ISSUE #6006: fairshare: enableEnqueueGate never gates Volcano Jobs
### Description

The fairshare plugin's `fairshare.enableEnqueueGate` option is documented as "blocks namespaces at/above their calculated share from entering the scheduling pipeline" For `vcjob` workloads it never does.

`JobEnqueueableFn` computes the job's resource from `job.Tasks`. For a Volcano Job the job controller does not create any pods until the PodGroup phase is no longer `Pending`, so at enqueue time `job.Tasks` is empty, `jobRes` is 0, and the reject condition `running >= share && jobRes > 0` can never be true. Every job gets admitted, regardless of the namespace's share. The same root cause also makes pending vcjob demand invisible to the max-min share calculation (`namespaceDemand` is built from Pending tasks), so a namespace that already holds the whole queue ends up with `share == running`.

The gate only works for workloads whose pods exist before admission (e.g. training-operator, Spark operator). With vcjobs it silently does nothing  no log, no event, and no test covers the gate

Found through source review and reproduced with a scheduler framework test (real `enqueue` action + fairshare plugin) not yet reproduced on a live cluster

1. Scheduler config: actions `enqueue, allocate, backfill`; plugin `fairshare` with `fairshare.targetQueues: "q1"`, `fairshare.resourceKey: "cpu"`, `fairshare.enableEnqueueGate: "true"`.
2. Queue `q1` with capability `cpu: 4`; a single node with 4 allocatable CPU.
3. In namespace `ns-a`, create 4 vcjobs requesting 1 CPU each so they run and saturate the queue.
4. Create one more 1-CPU vcjob in `ns-a` and one 1-CPU vcjob in `ns-b`.
5. `kubectl get podgroup -A`

Expected: `ns-a`'s 5th PodGroup stays `Pending` (ns-a is at 100% of the queue); `ns-b`'s goes `Inqueue`.
Actual: both PodGroups go `Inqueue` immediately.

All links are at master commit 8708aca.

1. User creates a vcjob → job controller creates the PodGroup with `MinResources` set:
   https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/controllers/job/job_controller_actions.go#L825

2. The controller does NOT create pods while the PodGroup is `Pending` (`syncTask` stays false):
   https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/controllers/job/job_controller_actions.go#L410-L412
   (the controller's own tests set `PodGroupInqueue` for that reason: https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/controllers/job/job_controller_actions_test.go#L298)

3. Scheduler `enqueue` action calls `JobEnqueueable` on the Pending PodGroup:
   https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/scheduler/actions/enqueue/enqueue.go#L95

4. fairshare computes `jobRes` from `job.Tasks` (empty at this point) and can only reject when `jobRes > 0`:
   https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/scheduler/plugins/fairshare/fairshare.go#L391-L396
   https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/scheduler/plugins/fairshare/fairshare.go#L604-L610

5. Pending demand is likewise read from Pending *tasks*, so pre-admission vcjob demand never enters the share calculation:
   https://github.com/volcano-sh/volcano/blob/8708aca48a4c6cb167a3b509aee972d912f2ab4e/pkg/scheduler/plugins/fairshare/fairshare.go#L289-L293

For comparison, every other enqueue-gating plugin uses `job.GetMinResources()` for exactly this reason (proportion L425, capacity L798, overcommit L98).

Suggested fix (optional): use `job.GetMinResources().Get(resourceKey)` for `jobRes` and for pending demand when the job has no tasks, mirroring the other plugins, and add unit tests for the gate (currently none exist).

Expected: with `fairshare.enableEnqueueGate: "true"`, the 5th vcjob in `ns-a` stays `Pending` because `ns-a` already holds 100% of queue `q1`, while `ns-b`'s job goes `Inqueue`

Received: both PodGroups go `Inqueue` immediately. The gate never rejects because `jobRes` is 0 for any vcjob (no pods exist yet), and `ns-b`'s pending demand is invisible to the share calculation (`demand=map[]`, `share[ns-a]=4000 == running`). The plugin behaves as if the gate were disabled ordering only, no admission control  with no log or event indicating this

master @ 8708aca (fairshare plugin merged in #5161, 2026-09-15; not yet in a tagged release)

_No response_

CapThunder19:
## What type of PR is this?

kind / bug

`fairshare.enableEnqueueGate` never gated Volcano Jobs. `JobEnqueueableFn` computed the job's resource from `job.Tasks`, but the job controller does not create pods for a vcjob until its PodGroup leaves `Pending`, so at enqueue time `jobRes` was always 0 and the reject condition `running >= share && jobRes > 0` could never be true. Pending demand was also read from tasks only, so a namespace with no pods yet got no share at all

This PR falls back to `PodGroup.Spec.MinResources` for both `jobTotalResource` and pending demand when a job has no tasks the same signal `proportion`, `capacity` and `overcommit` already use at enqueue time. Behaviour for jobs that do have tasks is unchanged

Tests added:
- `TestJobTotalResource_*` sums tasks when present, falls back to MinResources when not, 0 when neither
- `TestEnqueueGate` runs the real `enqueue` action with the gate enabled, in two variants: pending jobs with pods (operator-style) and without pods (vcjob). Both expect the namespace holding the whole queue to be held back and the other namespace admitted. The vcjob variant fails on master (`want: Pending, got: Inqueue`) and passes with this change

Fixes #6006

```release-note
fairshare: `enableEnqueueGate` now works for Volcano Jobs; jobs without pods are gated by PodGroup MinResources.

### CODE REVIEW

File: pkg/scheduler/plugins/fairshare/fairshare_enqueue_test.go [hunk 1]
```diff
@@ -0,0 +1,151 @@
...
+	job.SetPodGroup(&api.PodGroup{PodGroup: scheduling.PodGroup{
+		Spec: scheduling.PodGroupSpec{MinResources: ptrResourceList(api.BuildResourceList("10", "1G"))},
+	}})
+
+	if got := jobTotalResource(job, v1.ResourceCPU); got != 3000 {
+		t.Errorf("jobTotalResource() = %v, want 3000 (sum of tasks, not MinResources)", got)
+	}
+}
+
+func TestJobTotalResource_FallsBackToMinResourcesWithoutTasks(t *testing.T) {
```
JesseStutler: nit: Better re-write the test func name to write a common use func, and use table-driven styles to write

pkg/scheduler/plugins/fairshare/fairshare.go
```diff
@@ -291,6 +291,9 @@ func (fsp *fairSharePlugin) OnSessionOpen(ssn *framework.Session) {
 				qs.namespaceDemand[namespace] += taskResource(task, qs.resourceKey)
 			}
 		}
+		if len(job.Tasks) == 0 && job.IsPending() {
```
JesseStutler, line 294: Also should add test case for this `namespaceDemand` change

CapThunder19:
@JesseStutler pls review it, any changes needed

